Issue #3618, opened 2018-10-12
Is NOX buildable with python?


VictorEijkhout:
```
/admin/rpms/BUILD/trilinos-git/packages/PyTrilinos/src/NOX.__init__.i:130: Error: Unable to find 'NOX_Version.H'
packages/PyTrilinos/src/CMakeFiles/PyTrilinos_NOX___init__.dir/build.make:64: recipe for target 'packages/PyTrilinos/src/NOX.__init__PYTHON_wrap.cpp' failed
make[2]: *** [packages/PyTrilinos/src/NOX.__init__PYTHON_wrap.cpp] Error 1
[trilinos.log.gz](https://github.com/trilinos/Trilinos/files/2474856/trilinos.log.gz)

```

wfspotz:
@VictorEijkhout It should be buildable on `develop`. What branch are you on?

VictorEijkhout:
On Oct 12, 2018, at 6:39 PM, Bill Spotz <notifications@github.com<mailto:notifications@github.com>> wrote:

What branch are you on?

I have no idea. How can I tell?

Btw, is there a recent stable release? I’d vastly prefer that to tinkering with a development repo.

Victor.

mhoemmen:
@VictorEijkhout wrote:

> I have no idea. How can I tell?

```
$ git status
On branch ${BRANCH_NAME}
Your branch is up-to-date with 'origin/${BRANCH_NAME}'.

nothing to commit, working tree clean
```

> Btw, is there a recent stable release? I’d vastly prefer that to tinkering with a development repo.

We have a 12.14 release branch tag; you can see it in the list of branches on GitHub.

Trilinos developers plan to discuss and reevaluate the major and minor release cycle.

VictorEijkhout:
It looks like Domi/Tpetra is still not fixed:

/tmp/trilinos-build/packages/PyTrilinos/src/DomiPYTHON_wrap.cpp(4981): error: class "Domi::MDMap" has no member "getTpetraMap"
      return self->template getTpetraMap< int,
                            ^

/tmp/trilinos-build/packages/PyTrilinos/src/DomiPYTHON_wrap.cpp(4981): error: type name is not allowed
      return self->template getTpetraMap< int,
                                          ^

/tmp/trilinos-build/packages/PyTrilinos/src/DomiPYTHON_wrap.cpp(4982): error: type name is not allowed
                                          long long,
                                          ^

[...]

Issue #3456, opened 2018-09-17
PyTrilinos: Conflicting types in Teuchos_BLAS_wrapper.hpp


wfspotz:
@trilinos/pytrilinos 
@trilinos/teuchos 
@trilinos/kokkos-kernels 

## Expectations
I expect to build PyTrilinos without compilation errors

## Current Behavior
I have a wrapper file that `#include`s the header `MLAPI_MultiVector.h` which gives the compilation error

    /Development/Trilinos/packages/teuchos/numerics/src/Teuchos_BLAS_wrappers.hpp:173:13: error: conflicting types for 'daxpy_'
    void PREFIX DAXPY_F77(const int* n, const double* alpha, const double x[], const int* incx, double y[], const int* incy);
                ^
    /Development/Trilinos/packages/teuchos/numerics/src/Teuchos_BLAS_wrappers.hpp:78:21: note: expanded from macro 'DAXPY_F77'
    #define DAXPY_F77   F77_BLAS_MANGLE(daxpy,DAXPY)
                        ^
    /Development/Trilinos/MPI/packages/teuchos/core/src/Teuchos_config.h:10:37: note: expanded from macro 'F77_BLAS_MANGLE'
     #define F77_BLAS_MANGLE(name,NAME) name ## _
                                        ^
    <scratch space>:23:1: note: expanded from here
    daxpy_
    ^
    /Development/Trilinos/packages/kokkos-kernels/src/impl/tpls/KokkosBlas1_axpby_tpl_spec_decl.hpp:49:17: note: previous declaration is here
    extern "C" void daxpy_( const int* N, const double* alpha,

## Definition of Done
I can get the PyTrilinos package to build and all of the PyTrilinos tests to pass.

## Possible Solution
I'm not sure why this conflicting type declaration is occurring, but it is clearly related to macro expansion. Based on `git blame` of `Teuchos_BLAS_wrapper.h` I'm hoping either @jwillenbring or @hkthorn might have some idea what the problem could be. PyTrilinos can present some unique configuration issues.

## Steps to Reproduce
If it gets to this, I can help someone set up their environment to build PyTrilinos.

ibaned:
I think @trilinos/kokkos-kernels should also take a look at this, since their header provides the second of the two conflicting definitions.

mhoemmen:
kokkos-kernels didn't hide the macro & extern definitions :(

wfspotz:
[...]

PR #3731, opened 2018-10-24
Domi: Fix config bug

wfspotz:
@trilinos/domi 

## Description
In `domi/src/CMakeLists.txt`, I needed to change `Domi_ENABLE_Tpetra` to `Domi_ENABLE_TpetraCore` to ensure that the `HAVE_TPETRA` macro was set properly in `Domi_config.h`.

## Motivation and Context
This manifest in PyTrilinos, where Tpetra for `PyTrilinos.Domi` was enabled, but not Tpetra for Domi.

## Related Issues

* Addresses #3618, #3687

## How Has This Been Tested?
I tested this by running the Domi tests and the PyTrilinos tests. For PyTrilinos, I had to disable ML in order to work around #3456.  The error really doesn't show up in Domi tests: Tpetra was always improperly disabled and all tests passed.

## Checklist

### DIFF
--- a/packages/domi/src/CMakeLists.txt
+++ b/packages/domi/src/CMakeLists.txt
@@ -51,7 +51,7 @@ IF(Domi_ENABLE_Epetra)
   SET(HAVE_EPETRA ON)
 ENDIF()
 
-IF(Domi_ENABLE_Tpetra)
+IF(Domi_ENABLE_TpetraCore)
   SET(HAVE_TPETRA ON)
 ENDIF()
 


wfspotz:
@trilinos/tpetra why would I get an

    undefined reference to `Tpetra::MultiVector<long, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >::getNumVectors() const'

It compiles fine on my MacBook Pro.

jhux2:
@wfspotz I labelled this as WIP, just until you can resolve the Tpetra issue.  I hope this is ok.

kddevin:
@wfspotz scalar type "long" is not one of the default instantiations.  Are you using ETI?
The default ETI scalar types are int, long long, and double.

wfspotz:
@kddevin When I test locally, I haven't been using ETI, but the automated testing probably does. That would explain it.

wfspotz:
@mhoemmen Can you review this?  It applies to #4210 as well.

packages/domi/test/MDArray/MDArrayRCP_UnitTests.cpp
```diff
@@ -578,8 +580,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( MDArrayRCP, toString3D, T )
 
 UNIT_TEST_GROUP(int)
 #if 1
-UNIT_TEST_GROUP(long)
-UNIT_TEST_GROUP(float)
+UNIT_TEST_GROUP(long_long_type)
```
mhoemmen: Do these integers correspond to Tpetra's GlobalOrdinal template parameter?  If so, then you'll want to use Tpetra's macros to decide whether those types are enabled.  Even if ETI is OFF, users are not allowed to use disabled types.  See `TpetraCore_config.h` (generated in the build directory, in `packages/tpetra/core/src/` for the macros; they have the form `HAVE_TPETRA_INST_INT_LONG` (replace `LONG` with e.g., `LONG_LONG` or `INT`, for the GlobalOrdinal type).

packages/domi/test/MDVector/MDVector_UnitTests.cpp
```diff
@@ -871,21 +873,15 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( MDVector, augmentedConstruction, Sca )
 #endif
 
 #ifdef HAVE_TPETRA
-  Teuchos::RCP< Tpetra::Vector< Sca, int >      > tv1  =
-    mdv1.template getTpetraVectorView< int >();
-  Teuchos::RCP< Tpetra::MultiVector< Sca, int > > tmv2 =
-    mdv2.template getTpetraMultiVectorView< int >();
-  Teuchos::RCP< Tpetra::MultiVector< Sca, int > > tmv3 =
-    mdv3.template getTpetraMultiVectorView< int >();
+  Teuchos::RCP< Tpetra::Vector< Sca, int >      > tv1  = mdv1.template getTpetraVectorView< int >();
```
mhoemmen: `int` is the default (and only permitted) value of `LocalOrdinal`, so you don't need to specify `int` as the second template argument of `Tpetra::Vector` and `Tpetra::MultiVector`.

packages/domi/test/MDArray/MDArrayRCP_UnitTests.cpp
wfspotz: No, these are `SCALAR` types.  The ordinal types are determined at compile time and default to `int` for multidimensional indexes and `long long` for 1D flattened dimensions.

packages/domi/test/MDVector/MDVector_UnitTests.cpp
wfspotz: I have changed these tests to use the simplified template arguments.

wfspotz:
I am going to replace this pull request with a new one based on a new branch that cherry-picks the relevant commits and ignores all of the accidental merges.

packages/domi/test/MDArray/MDArrayRCP_UnitTests.cpp
mhoemmen: Note that even if ETI is OFF, you must first check the macros before using a particular integer type as a GlobalOrdinal type in Tpetra.

PR #3735, opened 2018-10-25
ATDM: Add mention of needing to use bash -l for non-bash users

bartlettroscoe:
Non-bash developers like @csiefer2 have been gotten tripped up by this.  Hopefully this helps.

### DIFF
--- a/cmake/std/atdm/README.md
+++ b/cmake/std/atdm/README.md
@@ -28,7 +28,8 @@ build locally as described below.
 After [cloning the Trilinos git
 repo](https://github.com/trilinos/Trilinos/wiki/VC-%7C-Initial-Git-Setup) on
 one of the supported ATDM machines, a local configure of Trilinos enabling a
-few packages is performed as:
+few packages is performed using the bash shell (or opening a new bash shell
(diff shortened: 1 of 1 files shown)


cmake/std/atdm/README.md
```diff
@@ -675,6 +676,11 @@ ATDM Trilinos configuration and testing scripts should work with just about
 any version of git 2.0+.  Another approach would be for the user to
 (temporarily) edit their `~/.gitconfig` file to address the problems.
 
+Another problem occurs for developers who don't use bash but instead need to
+switch into a bash shell.  In this case, one must use `bash -l` so that the
+users `.bash_profile` file will get sourced.  Without this, one can't load the
```
jmgate: user's

bartlettroscoe: @jmgate, I fixed in an amended commit and force-pushed.  Therefore, I dismissed your "request changes" review.

bartlettroscoe:
@csiefer2, this change was meant for non-bash users like yourself.  Can you please see if this would have been hopeful to you to avoid the `bash -l` problem?

PR #3658, opened 2018-10-17
Fixes condition estimate computation to address issue #3338

hkthorn:
The issue #3338 is the failure of the Tpetra pseudo-block CG test on an important test platform.

@trilinos/belos 

## Description

There are two issues, one the _STEQR provided by the LAPACK libraries
on that platform is seg faulting and, two, the general condition estimate
computation performed by single-vector/pseudo-block CG is generally wrong.
The condition estimate computation is now implemented for the single-vector
CG kernel, when it wasn't before. Furthermore, it is corrected for the
pseudo-block CG kernel, where it was wrongly storing information when
there was more than one right-hand side.  The LAPACK routine being used to
perform the condition estimation computation is _PTEQR, which has no issues
on the test platform of concern to #3338.

Tests have been fixed for the Tpetra pseudo-block CG, so that they output
information, including the condition estimation.  Tests have been augmented
to perform condition estimation for Epetra and perform an unpreconditioned
pseudo-block CG solve, using Epetra.

## Motivation and Context

The condition estimation computation was wrong, in almost every circumstance.
Now it is right.

## How Has This Been Tested?

Locally on OS X using GCC7.

## Checklist

- [x] My commit messages mention the appropriate GitHub issue numbers.
- [x] My code follows the code style of the affected package(s).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] No new compiler warnings were introduced.
- [ ] These changes break backwards compatibility.

Issue #3542, opened 2018-10-02
Stokhos tests failing in Trilinos-atdm-white-ride-cuda-9.2-debug-pt build


bartlettroscoe:
CC: @trilinos/stokhos, @rppawlo (Trilinos Embedded Nonlinear Analysis Product Area Lead)

## Description

The Stokhos package has 66 failing tests in the build `Trilinos-atdm-white-ride-cuda-9.2-debug-pt ` on 'white' and 'ride' as shown [here](https://testing.sandia.gov/cdash-dev-view/viewTest.php?onlyfailed&buildid=3998261) which shows the failing tests:

* Stokhos_AdaptivityToolsUnitTest_MPI_1	
* Stokhos_AlgebraicExpansionUnitTest_MPI_1	
* Stokhos_BasisInteractionGraphUnitTest_MPI_1	
* Stokhos_division_example_MPI_1	
* Stokhos_DivisionOperatorUnitTest_MPI_1	
* Stokhos_ExponentialRandomFieldUnitTest_MPI_1	
* Stokhos_GramSchmidtUnitTest_MPI_1	
* Stokhos_hermite_example_MPI_1	
* Stokhos_HermiteBasisUnitTest_MPI_1	
* Stokhos_InterlacedMapUnitTest_MPI_2	
* Stokhos_InterlacedOpUnitTest_MPI_2	
* Stokhos_JacobiBasisUnitTest_MPI_1	
* Stokhos_KokkosArrayKernelsUnitTest_Cuda_MPI_1	
* Stokhos_KokkosArrayKernelsUnitTest_Serial_MPI_1	
* Stokhos_KokkosCrsMatrixUQPCEUnitTest_Cuda_MPI_1	
* Stokhos_KokkosCrsMatrixUQPCEUnitTest_Serial_MPI_1	
* Stokhos_KokkosViewUQPCEUnitTest_Cuda_MPI_1	
* Stokhos_KokkosViewUQPCEUnitTest_Serial_MPI_1	
* Stokhos_LanczosUnitTest_MPI_1	
* Stokhos_LegendreBasisUnitTest_MPI_1	
* Stokhos_LexicographicTreeBasisUnitTest_MPI_1	
* Stokhos_Linear2D_Diffusion_CG_AGS_MPI_2	
* Stokhos_Linear2D_Diffusion_GMRES_AGS_MPI_2	
* Stokhos_Linear2D_Diffusion_GMRES_AJ_MPI_2	
[...]

PR #3741, opened 2018-10-26
Replace STEQR with PTEQR in Stokhos for issue #3542

etphipp:
The LAPACK function STEQR doesn't appear to work on IBM Power systems.  In #3658 a solution was found by replacing STEQR with PTEQR, which seems to not have problems on this platform.  This PR replaces usage of STEQR with PTEQR in Stokhos.  However this isn't completely trivial because PTEQR is for symmetric, positive-definite systems where STEQR is for symmetric, indefinite systems (and therefore is a completely different algorithm, which is probably why it works when STEQR doesn't).  So to use PTEQR I had to compute a shift to make the system positive definite.  This caused a few tests to fail due to numerical differences with very tight tolerances, so I had to loosen some of those tolerances to get things to pass.  After these changes, all of the Stokhos tests pass on white.  For issue #3542.

### DIFF
--- a/packages/stokhos/src/Stokhos_RecurrenceBasisImp.hpp
+++ b/packages/stokhos/src/Stokhos_RecurrenceBasisImp.hpp
@@ -375,30 +375,59 @@ getQuadPoints(ordinal_type quad_order,
       normalizeRecurrenceCoefficients(a, b, c, d);
   }
 
-  // With normalized coefficients, A is symmetric and tri-diagonal, with
-  // diagonal = a, and off-diagonal = b, starting at b[1]
-  Teuchos::SerialDenseMatrix<ordinal_type,value_type> eig_vectors(num_points,
-                                                                  num_points);
-  Teuchos::Array<value_type> workspace(2*num_points);
-  ordinal_type info_flag;
-  Teuchos::LAPACK<ordinal_type,value_type> my_lapack;
-
-  // compute the eigenvalues (stored in a) and right eigenvectors.
-  if (num_points == 1)
-    my_lapack.STEQR('I', num_points, &a[0], &b[0], eig_vectors.values(),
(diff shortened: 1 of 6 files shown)


packages/stokhos/src/Stokhos_RecurrenceBasisImp.hpp
```diff
@@ -375,30 +375,59 @@ getQuadPoints(ordinal_type quad_order,
...
+    }
+    value_type shift = 1.0 + max_a + 2.0*max_b;
+
+    // Shift diagonal
+    for (ordinal_type n = 0; n<num_points; n++)
+      a[n] += shift;
+
+    // compute the eigenvalues (stored in a) and right eigenvectors.
+    my_lapack.PTEQR('I', num_points, &a[0], &b[1], eig_vectors.values(),
+                    num_points, &workspace[0], &info_flag);
```
mhoemmen: Consider going back later and fixing this to use `workspace.data()` (if `std::vector`) or `workspace.getRawPtr()` (if `Teuchos::Array*`).

packages/stokhos/test/UnitTest/Stokhos_HermiteBasisUnitTest.cpp
```diff
@@ -61,7 +61,7 @@ namespace HermiteBasisUnitTest {
     OrdinalType p;
     Stokhos::HermiteBasis<OrdinalType,ValueType> basis;
     
-    UnitTestSetup() : rtol(1e-12), atol(1e-7), p(10), basis(p) {}
+    UnitTestSetup() : rtol(1e-12), atol(1e-5), p(10), basis(p) {}
```
mhoemmen: Consider doing a bit of numerical analysis later (not for this PR) to get these tolerances.

Issue #3785, opened 2018-10-31
Panzer Orientation Error 


bathmatt:
It looks like the orientation is stored as a workset local var, however, we are using MPI rank cell index.  This doesn't show up unless you use multi blocks.

Two blocks  gives this solution, blocks split down the middle

![visit0000](https://user-images.githubusercontent.com/14926624/47810242-6d6ae900-dd08-11e8-9f4e-9cf304e83982.png)

with the fix in the related PR you get

![visit0001](https://user-images.githubusercontent.com/14926624/47810314-a7d48600-dd08-11e8-9012-f00c4daa00a2.png)

which is the correct results..  I ran all the tests in panzer, they pass.

@trilinos/<teamName>

## Your Environment

- **Relevant repo SHA1s:**  
- **Relevant configure flags or configure script:**  
- **Operating system and version:**  
- **Compiler and TPL versions:**  

## Related Issues

* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of

bathmatt:
@rppawlo let me know if you can show this in a panzer test

bathmatt:
Before my changes or after

rppawlo:
> Before my changes or after

Sorry - missed your reply. I compared before and after your changes for CurlLaplacian and MixedPoisson. Multiple orders on hex, tri and quad with different mpi distributions. Couldn't trigger a failure or a difference with your changes. Were you using one of the panzer inline mesh generators for the problem that exhibited this or was it an exodus file? If inline, could you send the values to reconstruct that mesh exactly?

rppawlo:
Fix is in. New tests for multiblock are in as well. Closing.

PR #3786, opened 2018-10-31
Went to all workset local index into orientation array

bathmatt:
See #3785

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_IntrepidOrientation.hpp
+++ b/packages/panzer/disc-fe/src/Panzer_IntrepidOrientation.hpp
@@ -91,20 +91,27 @@ namespace panzer {
     
     const auto fp = NodalFieldPattern(cellTopo);
     connMgr.buildConnectivity(fp);
+
+    // Count and pre-alloc orientations
+    int total_elems = 0;
+    for (int i=0;i<numElementBlocks;++i) {
+      total_elems += connMgr.getElementBlock(elementBlockIds.at(i)).size();
+    }
     
+    orientation.resize(total_elems);
     // Loop over element blocks
     for (int i=0;i<numElementBlocks;++i) {
       // get elements in a block
       const auto &elementBlock = connMgr.getElementBlock(elementBlockIds.at(i));
       
       const int numElementsPerBlock = elementBlock.size();
-      
+
       // construct orientation information
       for (int c=0;c<numElementsPerBlock;++c) {
         const int localCellId = elementBlock.at(c);
         Kokkos::View<const GlobalOrdinal*, Kokkos::DefaultHostExecutionSpace>
           nodes(connMgr.getConnectivity(localCellId), numVertexPerCell);
-        orientation.push_back(Intrepid2::Orientation::getOrientation(cellTopo, nodes));
+        orientation[localCellId] = (Intrepid2::Orientation::getOrientation(cellTopo, nodes));
       }
     }
   }


bathmatt:
```
100% tests passed, 0 tests failed out of 159

Label Time Summary:
Panzer    = 773.96 sec (159 tests)

Total Test time (real) = 113.65 sec

```

eric-c-cyr:
I don't think this is right. The change proposed doesn't make a whole lot of sense. For instance in Panzer_WorksetContainer.cpp:298 orientations_ is defined over all cells on the processor. While the loop is clearly over a workset.

```
          ortsPerBlock.clear();
          for (int k=0;k<worksets[i].num_cells;++k) {
            ortsPerBlock.push_back((*orientations_)[details.cell_local_ids[k]]);
          }
```

bathmatt:
So you think what is written is correct, I changed it to push back just [k], the old code is listed below

> I don't think this is right. The change proposed doesn't make a whole lot of sense. For instance in Panzer_WorksetContainer.cpp:298 orientations_ is defined over all cells on the processor. While the loop is clearly over a workset.
> 
> ```
>           ortsPerBlock.clear();
>           for (int k=0;k<worksets[i].num_cells;++k) {
>             ortsPerBlock.push_back((*orientations_)[details.cell_local_ids[k]]);
>           }
> ```

bathmatt:
Don't merge, this has a bug on big problems...  Probably problems where the workset size is bigger than the block size...

rppawlo:
I'm working on converting CurlLaplacian and/or MixedPoisson to multiblock to see if we can replicate there. Have a meeting with @eric-c-cyr and Nate on Fluids/GPU. After that we will stay on and talk more about this. @bathmatt - do you want to call in? If so, will forward invite.

eric-c-cyr:
@bathmatt, did the fix you push solve the problem for your example?

eric-c-cyr:
@bathmatt, did the fix you push solve the problem for your example?

eric-c-cyr:
@bathmatt, did the fix you push solve the problem for your example?

bathmatt:
> @bathmatt, did the fix you push solve the problem for your example?

It fixed it for small problems, not for large problems.  
Let me try to tri meshes, it may be since all the meshes look the same.

bathmatt:
visit0002

Triangles also work with my fixes, trilinos/develop gives
visit0003

bathmatt:
Sure, i can call in @rppawlo

bathmatt:
I can't figure out what magic fills the setup data or what the setup data is even,   sd.orientations_ 
```
template<typename EvalT, typename Traits>
void
BasisValues_Evaluator<EvalT, Traits>::
postRegistrationSetup(
  typename Traits::SetupData  sd,
  PHX::FieldManager<Traits>&  fm)
{
  int space_dim = basis->dimension();

  orientations = sd.orientations_;
```

eric-c-cyr:
How big is the problem you are running, I assume there are issues in serial?

eric-c-cyr:
Or is this in parallel?

eric-c-cyr:
Or is this in parallel?

bathmatt:
> How big is the problem you are running, I assume there are issues in serial?

the test problem 30 elements, serial

The real problem 150k elements, parallel, 

There is a bug in what I've done. still tracking that down..  but I think WorksetContainer::orientations_ is a global structure,

Trying to figure out where the local ones are really used.

eric-c-cyr:
In BasisValues2 I suspect

eric-c-cyr:
In BasisValues2 I suspect

bathmatt:
> In BasisValues2 I suspect

I looked at the apply there, it takes in a copy from workset container of workset local orientations.  

```
      for(std::map<unsigned,Workset>::iterator itr=worksets.begin(); 
          itr!=worksets.end();++itr) { 
                // break out of the workset loop
        if(itr->second.num_cells<=0) continue;
        
        for(std::size_t j=0;j<itr->second.numDetails();j++) {  
          WorksetDetails & details = itr->second(j);
  
          ortsPerBlock.clear();
          for (int k=0;k<itr->second.num_cells;++k) {
            ortsPerBlock.push_back((*orientations_)[details.cell_local_ids[k]]);
          }
  
          for(const auto & id : needs.getIntegrators()) {
            // apply orientations for this basis
            details.getBasisValues(bd,id).applyOrientations(ortsPerBlock);
          }
        }
      }
    } // end for w

```

Then it passes that after a couple more copies into intrepid, 

```
        ots::modifyBasisByOrientation(drv_grad_basis, 
                                      drv_grad_basis_tmp, 
                                      drv_orts,
                                      intrepid_basis);
```

bathmatt:
Please ignore the previous changes, I just force pushed something that fixes it and I understand why it was broken so the fix I trust.

eric-c-cyr:
That seems pretty straight forward. Thanks for finding it.

PR #3799, opened 2018-11-05
Ifpack2: change ILUT level-of-fill to double, update documentation (I…

atoth1:
…ssue #3782)

@trilinos/Ifpack2 @mhoemmen 

## Description

In ILUT changed the storage of level-of-fill from int to double, updated documentation to reflect actual behavior, and added a unit test.

## Motivation and Context

Previously the documentation for this parameter given in setParameters() indicated that the parameter "fact: ilut level-of-fill" represented a max number of additional nonzero entries per row of the L and U factors, when it was actually being used as a fill fraction. As such, this is more appropriately represented by a double. The parameter was previously allowed to be provided as a double in input, but this value was cast to an int for storage and cast back to a double for computations. This truncation is unnecessary.

## Checklist

- [X] My commit messages mention the appropriate GitHub issue numbers.
- [X] My code follows the code style of the affected package(s).
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [X] No new compiler warnings were introduced.

## Additional Information
Unsure if this is consider breaking backwards compatibility. This could impact results for users providing a non-integer double value for this parameter, but in this case the previous behavior wasn't doing what was expected.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_ILUT_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_ILUT_def.hpp
@@ -118,7 +118,7 @@ ILUT<MatrixType>::ILUT (const Teuchos::RCP<const row_matrix_type>& A) :
   Athresh_ (Teuchos::ScalarTraits<magnitude_type>::zero ()),
   Rthresh_ (Teuchos::ScalarTraits<magnitude_type>::one ()),
   RelaxValue_ (Teuchos::ScalarTraits<magnitude_type>::zero ()),
-  LevelOfFill_ (1),
+  LevelOfFill_ (1.0),
   DropTolerance_ (ilutDefaultDropTolerance<scalar_type> ()),
   InitializeTime_ (0.0),
   ComputeTime_ (0.0),
@@ -152,7 +152,7 @@ void ILUT<MatrixType>::setParameters (const Teuchos::ParameterList& params)
   using Teuchos::Exceptions::InvalidParameterType;
 
   // Default values of the various parameters.
-  int fillLevel = 1;
+  double fillLevel = 1.0;
   magnitude_type absThresh = STM::zero ();
   magnitude_type relThresh = STM::one ();
   magnitude_type relaxValue = STM::zero ();
@@ -160,8 +160,8 @@ void ILUT<MatrixType>::setParameters (const Teuchos::ParameterList& params)
 
   bool gotFillLevel = false;
   try {
-    // Try getting the fill level as an int.
-    fillLevel = params.get<int> ("fact: ilut level-of-fill");
+    // Try getting the fill level as an double.
+    fillLevel = params.get<double> ("fact: ilut level-of-fill");
     gotFillLevel = true;
   }
   catch (InvalidParameterName&) {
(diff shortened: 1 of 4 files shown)


packages/ifpack2/src/Ifpack2_ILUT_def.hpp
```diff
@@ -152,16 +152,16 @@ void ILUT<MatrixType>::setParameters (const Teuchos::ParameterList& params)
...
   // Default values of the various parameters.
-  int fillLevel = 1;
+  double fillLevel = 1.0;
   magnitude_type absThresh = STM::zero ();
   magnitude_type relThresh = STM::one ();
   magnitude_type relaxValue = STM::zero ();
   magnitude_type dropTol = ilutDefaultDropTolerance<scalar_type> ();
 
   bool gotFillLevel = false;
   try {
```
mhoemmen: For later (not this PR): Let's get rid of the "try-catch as control logic" and replace with `params.isType<double>("fact: ilut level-of-fill")` etc.

packages/ifpack2/src/Ifpack2_ILUT_def.hpp
```diff
@@ -177,27 +177,28 @@ void ILUT<MatrixType>::setParameters (const Teuchos::ParameterList& params)
...
-      fillLevel = as<int> (params.get<double> ("fact: ilut level-of-fill"));
+      fillLevel = static_cast<double> (params.get<int> ("fact: ilut level-of-fill"));
     }
     catch (InvalidParameterType&) {}
   }
   // If none of the above attempts succeed, accept the default value.
 
   TEUCHOS_TEST_FOR_EXCEPTION(
-    fillLevel <= 0, std::runtime_error,
+    fillLevel < 1.0, std::runtime_error,
```
mhoemmen: Note to other readers: this is the one way in which this change could break backwards compatibility.  Users might have set fill level to zero before, expecting that ILUT's fill level behaves like the "k" in ILU(k).  ILUT's fill level does _not_ behave like the "k" in ILU(k).  I am in favor of stopping users from doing this wrong thing, but this could possibly mean that users get throws when they didn't before.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestAdditiveSchwarz.cpp
```diff
@@ -1012,7 +1012,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(Ifpack2AdditiveSchwarz, ILU_Overlap, Scalar, L
   params.set ("inner preconditioner name", "RILUK");
   {
     Teuchos::ParameterList innerParams;
-    innerParams.set ("fact: ilut level-of-fill", 0.0);
+    innerParams.set ("fact: iluk level-of-fill", 0.0);
```
mhoemmen: Wow, nice catch!

atoth1:
@mhoemmen Could you merge this when you get the chance. I don't have permissions to do it.

mhoemmen:
@atoth1 I'm in pretty intense meetings all week but just did it :-D

PR #3803, opened 2018-11-05
Framework: Make Kokkos-Kernels build in PR testing...

william76:
@trilinos/framework 	

# DO NOT MERGE THIS INTO TRILINOS!

This is a trivial edit to a file in Kokkos-Kernels to make the
PR testing system build the Kokkos-Kernels package to see if the
pthreads error from nightly clean builds is replicated.

PR #3814, opened 2018-11-06
Revert "Framework: Make Kokkos-Kernels build in PR testing..."

william76:
Reverts trilinos/Trilinos#3803 because people don't read.

### DIFF
--- a/packages/kokkos-kernels/src/graph/KokkosGraph_graph_color_d2.hpp
+++ b/packages/kokkos-kernels/src/graph/KokkosGraph_graph_color_d2.hpp
@@ -115,6 +115,3 @@ void graph_color_d2(KernelHandle *handle,
 
 #endif //_KOKKOS_GRAPH_COLOR_HPP
 
-
-// This is a trivial edit to force the PR tester to build KokkosKernels.  
-// DO NOT MERGE THIS PR INTO TRILINOS!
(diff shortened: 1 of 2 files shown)


crtrott:
To elaborate: besides the additional comment about Not Merging in in the code, this is the change the Kokkos team recommended in the github issue to be done for now. We merged this morning a new Kokkos and KokkosKernels release into Trilinos, and this issue here is potentially masking problems with that new release. I talked to Siva about this and as the code owners for Kokkos we decided this is the right thing to do for now. Issuing another pull request to just have the comment not be in the code felt rather excessive, in particular since it is not clear that the testing will be finished before nightlies go off. So we decided to merge.

william76:
@crtrott 	

ok, fine. I'll cancel this revert. But... In the future, please do not just add automerge tags, etc. to Framework team pull requests if you aren't part of the discussion we've had in our meetings about what our goals are in that PR without a conversation with us about it first. 

There is more going on with that PR than simply implementing the change we discussed in the other issue ticket, which is why I wanted to have an opportunity to review the output before doing anything.  That is exactly why I made notes in many places to tell people to NOT merge that PR.

It has happened a few times before where non framework folks have merged our PR's without consulting us first... I find it greatly annoying since if I'm wanting to merge immediately after testing is done without reviewing the results, I will set the AT:AUTOMERGE flag myself.

PR #3815, opened 2018-11-06
Framework: Removing dummy comment for Kokkos-Kernels Testing

william76:
Removing a dummy comment added in that got merged in by mistake.

I will merge this to remove the dummy comment that shouldn't have been merged in.  
Before I merge it in, I want to look at the testing results in the auto-tester system before I merge it, so do not set AT:AUTOMERGE on this.

@trilinos/framework


william76:
@prwolfe 	
@jwillenbring 	
This can be merged in, can I get an approval when you have a chance :)

william76:
FYI @srajama1 Just letting you know we removed that dummy comment from the Kokkos-Kernels instance that's in Trilinos.

PR #3829, opened 2018-11-08
NOX: Compile with petsc>=3.7 too

finkandreas:
`PetscGetOptions*` changed the signature in PETSc-3.7.
One pure virtual function was missing in the implementation of `Petsc::Vector.`


mhoemmen:
@trilinos/nox

Issue #3845, opened 2018-11-10
Sacado test Fad_KokkosTests_NoViewSpec_XXX.cpp CUDA build failures on all CUDA builds starting 11/10/2018


bartlettroscoe:
CC: @trilinos/sacado, @rppawlo  (Trilinos Nonlinear Solvers Product Lead)
 
## Next Action Status

Enable of `Sacado_ENABLE_HIERARCHICAL_DFAD=ON` in PR #3800 merged on 10/9/2018 triggered this build failure.  PR #3845 (once merged) should fix the failures for all builds ...

## Description

As shown in [this query](https://testing.sandia.gov/cdash-dev-view/index.php?project=Trilinos&date=2018-11-10&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=cuda&field2=subprojects&compare2=93&value2=Sacado&field3=testspassed&compare3=43&value3=0) the Sacado test suite has two build errors in every Sacado build that builds the Sacado test suite.

The ATDM builds involved (that have submitted to as of 9:20 AM MT on 11/10/2018) include:

* hansen | Trilinos-atdm-hansen-shiller-cuda-8.0-debug
* hansen | Trilinos-atdm-hansen-shiller-cuda-8.0-opt 
* hansen | Trilinos-atdm-hansen-shiller-cuda-9.0-debug
* hansen | Trilinos-atdm-hansen-shiller-cuda-9.0-opt 
* waterman | Trilinos-atdm-waterman-cuda-9.2-debug
* waterman | Trilinos-atdm-waterman-cuda-9.2-opt
* white | Trilinos-atdm-white-ride-cuda-9.2-debug
* white | Trilinos-atdm-white-ride-cuda-9.2-opt
* waterman | Trilinos-atdm-waterman-cuda-9.2-release-debug
* ride | Trilinos-atdm-white-ride-cuda-9.2-debug
* ride | Trilinos-atdm-white-ride-cuda-9.2-debug-pt
* white | Trilinos-atdm-white-ride-cuda-9.2-debug-pt
* ride | Trilinos-atdm-white-ride-cuda-9.2-opt

Some of the non-ATDM Trilinos builds showing this build error include:

* ascic141 | Linux-gcc-7.3.0-Sierra_MPI_release_DEV_ETI_SERIAL-ON_OPENMP-OFF_PTHREAD-OFF_CUDA-OFF_COMPLEX-ON
[...]

PR #3800, opened 2018-11-05
Set Sacado_ENABLE_HIERARCHICAL_DFAD=ON for all CUDA builds (#2446, #3579, #2751)

bartlettroscoe:
#2446, #3579, #2751, #3833 

Not ready to merge yet!

PR #3848, opened 2018-11-12
Sacado:  Fix build errors reported in issue #3845.

etphipp:
With Sacado_ENABLE_HIERARCHICAL_DFAD=ON, as was turned on in PR #3800 in
ATDM builds, a build error (#3845) was seen in the unit tests that disable the
View specializations for Fad, since `is_dynrankview_fad_contiguous<>` is
used without being defined.  I moved the default definition of this
trait class so that it is always defined, which fixed the build errors in my
CUDA build.

### DIFF
--- a/packages/sacado/src/Kokkos_DynRankView_Fad.hpp
+++ b/packages/sacado/src/Kokkos_DynRankView_Fad.hpp
@@ -58,6 +58,12 @@ struct MirrorDRVType;
 
 }
 
+template <typename view_type>
+struct is_dynrankview_fad { static const bool value = false; };
+
+template <typename view_type>
+struct is_dynrankview_fad_contiguous { static const bool value = false; };
+
 }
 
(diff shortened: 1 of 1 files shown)


etphipp:
Just to clarify, the symbols aren't in the global scope.  They were, and still are, in the Kokkos namespace.  I just moved them to be outside of the region that is only parsed if a certain #define is set.

bartlettroscoe:
> Just to clarify, the symbols aren't in the global scope. They were, and still are, in the Kokkos namespace. I just moved them to be outside of the region that is only parsed if a certain #define is set.

@etphipp, okay, thanks.  I see that now looking at the full file [here](https://github.com/trilinos/Trilinos/blob/00ddec1905fbed93b8e34c03334169fdcbd082c1/packages/sacado/src/Kokkos_DynRankView_Fad.hpp).  (In my C++ code I usually put `} // namespace Kokkos` to make this more obvious.)

PR #3868, opened 2018-11-13
Teuchos: Fix `mergeCounterNamesPair`

cgcgcg:
@trilinos/teuchos 

## Description
The binary tree reduce to get the union/intersection of the timer names on all MPI ranks accumulates into the variable `globalNames`, yet the variable `localNames` is sent. This lead to missing timers, since their names never made it to rank 0. This would happen if a timer only existed on ranks that are 2 or more hops away from rank 0. 

In the example from issue #3850, the missing timers from level 4 live only on rank 11. 
Rank 0 gets information from ranks 1, 2, 4, 8, and 16.
Rank 8 gets information from ranks 9, 10 and 12.
Rank 10 gets information from rank 11.

This addresses issue #3850.


packages/teuchos/comm/src/Teuchos_PerformanceMonitorBase.cpp
```diff
@@ -324,7 +324,6 @@ namespace Teuchos {
                            const int myRank,
                            const int left,
                            const int mid,
-                           const Array<std::string>& localNames,
                            Array<std::string>& globalNames,
```
mhoemmen: Please update the documentation if you plan to remove this parameter.

cgcgcg: Done. Thanks for catching that!

packages/teuchos/comm/test/TimeMonitor/TimeMonitor_Issue3868.cpp
```diff
@@ -0,0 +1,43 @@
...
+  using Teuchos::RCP;
+  using Teuchos::REDUCE_MAX;
+  using Teuchos::reduceAll;
+
+  RCP<const Comm<int> > comm = Teuchos::DefaultComm<int>::getComm ();
+
+  const int numProcs = comm->getSize ();
+  TEST_ASSERT( numProcs > 1 );
+  if (numProcs < 4) {
+    out << "This test requires at least 4 MPI processes." << std::endl;
```
mhoemmen: In the future, we should just make the test fail if it has the wrong number of processes.  I'm happy to have the test for now :D

packages/teuchos/comm/test/TimeMonitor/TimeMonitor_Issue3868.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+  if (myRank == 3) {
+    Teuchos::TimeMonitor timer(*Teuchos::TimeMonitor::getNewTimer(std::string("myTimer")));
+  }
+
+  std::ostringstream out1;
+  Teuchos::TimeMonitor::summarize(out1,false,true,false,Teuchos::Union,"",true);
+  int test = (out1.str().find("myTimer") != std::string::npos);
+
+  int gblTest = false; // output argument
```
mhoemmen: In the future, prefer `int glbTest = 0;`.

PR #3873, opened 2018-11-14
Seacas nov sync

gsjaardema:
@trilinos/seacas 

## Description
Periodic update from SEACAS github repository.

Primarily internal only refactoring.
 * Fix issue with exodiff element centroid matching code
 * Clean up nem_slice internals -- use more std::vector instead of raw arrays/mem pointers
 * Exodus add `ex_print_config` function
 * Exodus clean up `ex_err_fn()` uses [prints filename in error message]
 * Exodus better fortran doxygen output
 * IOSS: new `catch` release
 * IOSS: internal refactoring; remove unneeded function paramater
 * IOSS: cgns - clean up file handling so flush can be implemented more cleanly
 * IOSS: better exodus qa/info handling in parallel output

## Checklist


packages/seacas/applications/exodiff/ED_Version.h
```diff
@@ -30,10 +30,10 @@
...
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
-#ifndef SEACAS_Version_h
-#define SEACAS_Version_h
+#ifndef ED_Version_h
+#define ED_Version_h
 
-static std::string version("2.92");
-static std::string verdate("2018-06-27");
+static std::string version("2.94");
```
mhoemmen: I'm not a fan of `static std::string`, though I suppose it's OK.  Eventual C++ plan is to "`constexpr` _ALL_ THE THINGS!!!!1111!1!!1!11ONE!" but what's wrong with `const char version[] = "2.94";` for now?

packages/seacas/libraries/aprepro_lib/apr_builtin.cc
```diff
@@ -72,27 +73,11 @@ namespace SEAMS {
...
-      }
-    }
-    return hashval;
-  }
 
   std::vector<std::string> &get_tokenized_strings(const char *string, const char *delm)
   {
     // key is address of string + hash of delimiter
-    size_t key = hash(string) + hash(delm);
+    size_t key = std::hash<std::string>{}(string) + std::hash<std::string>{}(delm);
```
mhoemmen: It's more polite to use a hash combiner that promises not to overflow, but I suppose it doesn't matter a whole lot.

gsjaardema: Unsigned will wrap on overflow, so isn't undefined behavior as it would be with signed overflow.  Not trying to be impolite, but don't see that this is an issue here...

mhoemmen: I did approve the PR so don't worry :)

Issue #3879, opened 2018-11-14
Intrepid2_unit-test_Projection_Serial_Test_InterpolationProjection_HEX_MPI_1 failing on ATDM builds


fryeguy52:
CC: @trilinos/Intrepid2, @mperego (Trilinos Discretizations Product Lead), @bartlettroscoe, @fryeguy52 
 
## Next Action Status

PR #3882 that loosened tolerance merged on 11/15/2018 but this did not fix the failures in the `release-debug` builds on 'waterman'.  PR #3907 that loosed the tolerance some more was merged on 11/20/2018 to 'develop' that will also print out the numerical values to see why it is failing.  Next: Watch for results on 11/21 and beyond to see if these pass on all the builds (including the two `release-debug` builds on 'waterman').

## Description
As shown in [this query](https://testing.sandia.gov/cdash-dev-view/queryTests.php?project=Trilinos&filtercount=5&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-&field2=testname&compare2=65&value2=Intrepid2_unit-test_Projection_Serial_Test_InterpolationProjection_HEX_MPI_1&field3=status&compare3=62&value3=Passed&field4=buildstarttime&compare4=84&value4=2018-11-14&field5=buildstarttime&compare5=83&value5=2018-11-13) the test:
* Intrepid2_unit-test_Projection_Serial_Test_InterpolationProjection_HEX_MPI_1

is failing in the builds:
* Trilinos-atdm-serrano-intel-opt-openmp
* Trilinos-atdm-waterman-cuda-9.2-release-debug
* Trilinos-atdm-waterman-gnu-release-debug-openmp
* Trilinos-atdm-hansen-shiller-intel-opt-serial

This looks like a new test added recently.

## Current Status on CDash

The current status of these tests/builds for the current testing day can be found [here](https://testing.sandia.gov/cdash-dev-view/queryTests.php?project=Trilinos&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-&field2=testname&compare2=65&value2=Intrepid2_unit-test_Projection_Serial_Test_InterpolationProjection_HEX_MPI_1&field3=status&compare3=62&value3=Passed)

## Steps to Reproduce

One should be able to reproduce this failure on waterman as described in:
* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md
[...]

PR #3882, opened 2018-11-15
Intrepid2:  support first-order bases in interpolation class

mperego:
Intrepid2: Modified Lagrangian Interpolation to work also with low order bases
               Low order bases are used by Panzer (thanks to @egphill for pointing this out)
               Increased tolerance for test InterpolationProjection_HEX to address Issue #3879

@trilinos/intrepid2 

## Motivation and Context

Related to Issue #3879

## Checklist

### DIFF
--- a/packages/intrepid2/src/Projection/Intrepid2_LagrangianInterpolationDef.hpp
+++ b/packages/intrepid2/src/Projection/Intrepid2_LagrangianInterpolationDef.hpp
@@ -79,6 +79,9 @@ LagrangianInterpolation<SpT>::getDofCoordsAndCoeffs(
   bool isBasisHCURL = name.find("HCURL") != std::string::npos;
   bool isBasisHDIV = name.find("HDIV") != std::string::npos;
   bool isBasisTriOrTet = (name.find("TRI") != std::string::npos) || (name.find("TET") != std::string::npos);
+  bool isBasisHEXI1 = (name.find("HEX_I1") != std::string::npos);
+  bool isBasisTETI1 = (name.find("TET_I1") != std::string::npos);
+  bool isBasisI1 = (name.find("I1") != std::string::npos);
 
   ordinal_type numEdges = (basis->getDofCount(1, 0) > 0) ? topo.getEdgeCount() : 0;
   ordinal_type numFaces = (basis->getDofCount(2, 0) > 0) ? topo.getFaceCount() : 0;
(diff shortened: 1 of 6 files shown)


packages/intrepid2/src/Projection/Intrepid2_LagrangianInterpolationDef.hpp
```diff
@@ -185,7 +188,8 @@ LagrangianInterpolation<SpT>::getDofCoordsAndCoeffs(
       }
     }
 
-    auto edgeScale = isBasisTriOrTet ? 2.0 :1.0;
+    //depending on how the reference basis is defined, the edges are scaled differently
+    auto edgeScale = (isBasisTriOrTet||isBasisI1) ? 2.0 :1.0;
```
egphill: The logic here is if(triangle or tet or first order hex or first order tet) edgeScale = 2. Why is that?

packages/intrepid2/src/Projection/Intrepid2_LagrangianInterpolationDef.hpp
```diff
@@ -87,39 +90,39 @@ LagrangianInterpolation<SpT>::getDofCoordsAndCoeffs(
   Teuchos::RCP<Basis<SpT,scalarType,scalarType> > edgeBasis;
 
 
-  if (name == "Intrepid2_HGRAD_QUAD_Cn_FEM") {
+  if ((name == "Intrepid2_HGRAD_QUAD_Cn_FEM") || (name == "Intrepid2_HGRAD_QUAD_C1_FEM")) {
     edgeBasis = Teuchos::rcp(new Basis_HGRAD_LINE_Cn_FEM<SpT,scalarType,scalarType>(degree, pointType) );
-  } else if (name == "Intrepid2_HGRAD_HEX_Cn_FEM") {
+  } else if ((name == "Intrepid2_HGRAD_HEX_Cn_FEM") || (name == "Intrepid2_HGRAD_HEX_C1_FEM")){
     edgeBasis = Teuchos::rcp(new Basis_HGRAD_LINE_Cn_FEM<SpT,scalarType,scalarType>(degree, pointType) );
```
egphill: Why not use Basis_HGRAD_LINE_C1_FEM if name=="Intrepid2_HGRAD_HEX_C1_FEM"? Same for other first order bases.

packages/intrepid2/src/Projection/Intrepid2_LagrangianInterpolationDef.hpp
mperego: It really depends on how the basis are defined. We should uniform that at some point but we would break backward compatibility if somebody uses the bases in a weird way.
The convention used to build the first order basis are different the the ones used for high order basis, so, e.g. HGRAD_HEX_C1 is different then HGRAD_HEX_Cn with degree=1.

As for the difference between TRI/TET and QUAD/HEX, it also depends on the fact that the "reference" LINE elemet goes from -1 to 1, and all the edges in the reference HEX/QUAD do the same, however the TET/TRI are built considering a reference edge of size 1, which is were the factor 2 arises..

packages/intrepid2/src/Projection/Intrepid2_LagrangianInterpolationDef.hpp
mperego: It is possible, but I decided to simplify the code and not create more cases... This is only run once and it is independent on the batch side, so performance should not be an issue. Also we are considering refactorng all the basis, so that we can extract the sides/edges bases from the basis, so that we can get rid of all these conditionals

egphill: Ok. That makes sense. But given your other comment that "HGRAD_HEX_C1 is different than HGRAD_HEX_Cn with degree=1", is it really ok?

mperego: They are the same for LINE elements.

packages/intrepid2/src/Projection/Intrepid2_LagrangianInterpolationDef.hpp
mperego: B.t.w. This is partially the reason why @kyungjoo-kim does not like to use interpolations. In fact they heavily depend on the conventions  used in the bases and orientations. But they are fast. Alternatively you could use the projections (e.g. the l2 projections):
https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Projection/Intrepid2_ProjectionToolsDefL2.hpp

packages/intrepid2/src/Projection/Intrepid2_LagrangianInterpolationDef.hpp
egphill: Oh, I see. Ok.

Issue #3823, opened 2018-11-07
Tpetra: MultiVector unit test failure with complex turned on


lucbv:
@trilinos/tpetra @crtrott 

## Expectations
I am seeing test failures in MueLu's unit-test with complex turned on this morning.
My guess is that the failure is not completely (at all?) in MueLu as no commit was made to it since yesterday's nightly builds.

## Current Behavior
The unit-test: `TpetraCore_MultiVector_UnitTests_MPI_4` fails when complex is turned on with `SHA1 23c8989` but it works fine without complex unabled and with complex unabled and `SHA1 bf01cda`. Actually I narrowed it down to working commit `SHA1 0c5d4cf` and not working commit `SHA1 eb41c7f` which points at the kokkos promotion.

## Motivation and Context
I think this might be the reason for failures I am seeing in MueLu with complex enabled. I would like to resolve these...

## Definition of Done
We need to have that test pass with complex unabled

## Steps to Reproduce
Build with the indicated SHA1 and the following options:
```
            -D Trilinos_ENABLE_COMPLEX_DOUBLE=ON \
            -D Trilinos_ENABLE_Tpetra=ON \
            -D   Tpetra_ENABLE_TESTS=ON \
```

ndellingwood:
Added PR #3838 which addresses the compilation error, thanks to @crtrott for taking a look at it with me. Tpetra compiles without error but there are failing tests in the MultiVector unit test that need to be addressed.

mhoemmen:
Wow, we never actually fixed this.  @trilinos/tpetra @kddevin 

Does kokkos-kernels not test complex?
https://github.com/kokkos/kokkos-kernels/issues/360

mhoemmen:
I know the likely cause and how to fix this:

https://github.com/kokkos/kokkos-kernels/issues/360#issuecomment-455251898

mhoemmen:
@ndellingwood Does kokkos-kernels actually test the value of the complex dot product?

ndellingwood:
@mhoemmen kokkos-kernels tests the value of the complex dot product. Two views are initialized with random values, and the dot result is compared with the serialized result computed on the host. There is an error tolerance used in the comparison since sum is not associative for floating point types in a parallel reduction.

mhoemmen:
@ndellingwood Should we close this issue then? https://github.com/kokkos/kokkos-kernels/issues/360  I mean the promotion broke Trilinos, so something is wrong.

[...]

PR #3886, opened 2018-11-15
RefMaxwell & MiniEM updates

cgcgcg:
@trilinos/muelu 

## Description
This PR adds the option to perform the two sub-solves of RefMaxwell on disjoint sub-communicators, i.e. in parallel. I added to MueLu:
- Changes to MueLu's partition placement to set up the disjoint sub-communicators.
- Option to skip rebalancing a nullspace. (We rebalance coordinates and then use `CreateXpetraPreconditioner`)
- Fixes to `CreateXpetraPreconditioner` for when it is called on a sub-communicator.
- Finer solve timers in RefMaxwell.

Changes to MiniEM:
- Add 'Krylov only' option for the Schur complement solve. This is for comparisons at low CFL.
- Different conductivities

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/3886)


cgcgcg:
In order to fix these test failures, I need to rebase the gold files. However, the kokkos promotion broke all complex builds (Issue #3823), so I cannot do that currently.

PR #3917, opened 2018-11-21
Panzer: allow HVol elements

egphill:
Adapted CurlLaplacianExample into a mixed version that uses both HCurl and HDiv elements in 3D or HCurl and HVol elements in 2D.

Exposed HVol basis type and defined its basis values. Modified L2Projection to allow HVol elements.

@trilinos/<teamName>

## Description

Previously HVol elements were allowed in Panzer only through the Const basis option. This was restricted to zeroth order and did not use the HVOL basis value transformation for defining the basis. This change keeps the Const basis but adds the option to use arbitrary order HVol elements on 2D meshes.

## Motivation and Context

The z component of an HDiv DOF should use an HVol basis in 2D. This is necessary for arbitrary order electromagnetics in 2D.

## How Has This Been Tested?

Added the test example MixedCurlLaplacianExample. This is the same test as CurlLaplacianExample except it's in mixed form such that it uses both HCurl elements for the primary DOF and HVol elements for the curl of that DOF.

## Checklist

Most everything in the MixedCurlLaplacianExample directory is copied directly from CurlLaplacianExample. The changes of interest there are in Example_CurlLaplacianEquationSet_impl.hpp, main.cpp, and CMakeLists.txt.

### DIFF
--- a/packages/panzer/adapters-stk/example/CMakeLists.txt
+++ b/packages/panzer/adapters-stk/example/CMakeLists.txt
@@ -1,6 +1,7 @@
 ADD_SUBDIRECTORY(square_mesh)
 ADD_SUBDIRECTORY(PoissonExample)
 ADD_SUBDIRECTORY(CurlLaplacianExample)
+ADD_SUBDIRECTORY(MixedCurlLaplacianExample)
 ADD_SUBDIRECTORY(MixedPoissonExample)
 ADD_SUBDIRECTORY(PoissonInterfaceExample)
 ADD_SUBDIRECTORY(assembly_engine)
--- a/packages/panzer/adapters-stk/example/MixedCurlLaplacianExample/CMakeLists.txt
+++ b/packages/panzer/adapters-stk/example/MixedCurlLaplacianExample/CMakeLists.txt
@@ -1,0 +1,229 @@
+
+INCLUDE_DIRECTORIES(${CMAKE_CURRENT_SOURCE_DIR})
+
+SET(ASSEMBLY_EXAMPLE_SOURCES
+  main.cpp
+  )
+
+TRIBITS_ADD_EXECUTABLE(
+  MixedCurlLaplacianExample
+  SOURCES ${ASSEMBLY_EXAMPLE_SOURCES}
+  )
+
+TRIBITS_ADD_ADVANCED_TEST(
+  MixedCurlLaplacianExample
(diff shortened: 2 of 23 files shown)


egphill:
@mperego I want to keep Const for backward compatibility first of all. But in general, it might be useful. It's a piecewise constant basis that isn't scaled by the volume of the cell, so the basis is just one on every element.

packages/panzer/adapters-stk/example/MixedCurlLaplacianExample/Debugging.nb
```diff
@@ -0,0 +1,1741 @@
+(* Content-type: application/vnd.wolfram.mathematica *)
```
rppawlo: did you mean to commit this file?

mperego:
I see. 
Is Panzer using HVol for fields that are divergence of HDiv fields (in 2d and 3d) and curls of Hcurl fields in 2d?
I just want to make sure that we don't use the const basis when we should use the HVol basis instead.

egphill:
@mperego Now that we have HVol, we should be using it how you say. Historically, application codes have used Const for curls of HCurl fields in 2D, and it's been working at lowest order.

packages/panzer/adapters-stk/example/MixedCurlLaplacianExample/Debugging.nb
egphill: I can remove this file.

egphill:
@rppawlo I did not add an HVol projection test. Would you like me to?

rppawlo:
If it's quick, we probably should. I have a 2D projection test sitting on a branch so we could wait until that is merged in and add it there. If you wan to push now that's fine too.

egphill:
@rppawlo Is the 2D projection test you're talking about a 2D version of the ToNodal test in the L2Projection tests? If so, I think it'd be easiest just to include the HVol test there.

rppawlo:
> Is the 2D projection test you're talking about a 2D version of the ToNodal test in the L2Projection tests? If so, I think it'd be easiest just to include the HVol test there.

yes. I'll get that pushed next week and add it there.

packages/panzer/adapters-stk/example/MixedCurlLaplacianExample/Debugging.nb
egphill: Ok, I removed it. This needs to be re-approved.

PR #3921, opened 2018-11-22
Intrepid2: replace deep_copy with clone in FunctionSpaceTools

mperego:
deep_copy cannot handle the case where the input and output views have
  different types like double and FAD. Replacing deep_copy with clone in 
  function HGRADtransformVALUE. This should fix a runtime bug in Albany.

@trilinos/intrepid2 

## Checklist

### DIFF
--- a/packages/intrepid2/src/Discretization/FunctionSpaceTools/Intrepid2_FunctionSpaceToolsDef.hpp
+++ b/packages/intrepid2/src/Discretization/FunctionSpaceTools/Intrepid2_FunctionSpaceToolsDef.hpp
@@ -68,7 +68,7 @@ namespace Intrepid2 {
       }
     }
 #endif
-      Kokkos::deep_copy(output, input);
+      RealSpaceTools<SpT>::clone(output, input);
     }
     else
       ArrayTools<SpT>::cloneFields(output, input);


packages/intrepid2/src/Discretization/FunctionSpaceTools/Intrepid2_FunctionSpaceToolsDef.hpp
```diff
@@ -68,7 +68,7 @@ namespace Intrepid2 {
       }
     }
 #endif
-      Kokkos::deep_copy(output, input);
+      RealSpaceTools<SpT>::clone(output, input);
```
kyungjoo-kim: Before you change, we need to check @crtrott if this is a kokkos issue or not. As far as I know, if there is data type mismatch, it calls view mapping which uses an assignment operator. As fad and double has compatible assignment, it should be okay but I am not sure if kokkos supports it or not.

kyungjoo-kim: @etphipp Is it true that deep_copy between double and fad is not allowed ?

etphipp: Not directly.  You can assign a Fad view to a double-view using `array_type` and then do the deep_copy though.  It would work something like this:
```
typedef decltype(output) output_type;
typedef decltype(input) input_type;
typename output_type::array_type output_array = output;
typename input_type::array_type input_array = input;
deep_copy(output_array, input_array);
```
That should work for double or Fad.

mperego: @etphipp If I try that, I get
`/ascldap/users/mperego/Workspace/trilinos/builds/devel-shared-tmp/install/include/Kokkos_DynRankView.hpp: In instantiation of ‘class Kokkos::DynRankView<double*, Kokkos::LayoutRight, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<0u> >’:
/ascldap/users/mperego/Workspace/trilinos/builds/devel-shared-tmp/install/include/Intrepid2_FunctionSpaceToolsDef.hpp:74:40:   required from ‘static void Intrepid2::FunctionSpaceTools<ExecSpaceType>::HGRADtransformVALUE(Kokkos::DynRankView<DT, DP ...>, Kokkos::DynRankView<cubWeightValueType, cubWeightProperties ...>) [with outputValueType = Sacado::Fad::DFad<double>; outputProperties = {Kokkos::Serial}; inputValueType = double; inputProperties = {Kokkos::Serial}; ExecSpaceType = Kokkos::Serial]’
/ascldap/users/mperego/Workspace/trilinos/builds/devel-shared-tmp/install/include/Kokkos_DynRankView.hpp:420:3: error: static assertion failed: Cannot template DynRankView with array or pointer datatype - must be pod
   static_assert( !std::is_array<DataType>::value && !std::is_pointer<DataType>::value , "Cannot template DynRankView with array or pointer datatype - must be pod" );`

where line `/Intrepid2_FunctionSpaceToolsDef.hpp:74` is
`/Intrepid2_FunctionSpaceToolsDef.hpp:74`.

It seems that array_type is not correctly specified for DynRankViews.

etphipp: Darn, looks like you are correct.  I always forget you are using DynRankView.  I'll have to look into how this was done with DynRankView.

mperego: Thanks for looking into that. In the meanwhile I think it is better to push this fix so that Albany tests pass. @etphipp @kyungjoo-kim can you approve this PR?

Issue #2426, opened 2018-03-20
Framework: PR testing should enable complex by default


mhoemmen:
PR (pull request) testing should enable complex by default.  This is because at least two major internal customers need complex arithmetic to work.

I know it takes longer to build but it seems like PR testing should test the most used build configurations.

@trilinos/framework @prwolfe @rrdrake @rppawlo

bartlettroscoe:
As the Intel build dominates the PR builds, the GCC 4.8.4 build could have complex enabled and it would not impact the total wall-clock time for PR testing.

kddevin:
Can addition of complex scalar types to at least some of the PR builds be made a high priority?
We have a growing number of users of complex scalars, and are seeing repeated issues from not including this in PR testing.
Thanks!
@jwillenbring @william76 @bartlettroscoe @trilinos/framework

jmgate:
This just bit Charon again earlier this week.

jwillenbring:
@kddevin 

@ZUUL42 - let's talk about this Monday at the standup. Please bring it up as a parking lot.

@mhoemmen - could you list exactly the option(s) this will require? Do all packages use the same complex option?

bartlettroscoe:
Adding `Trilinos_ENABLE_COMPLEX_DOUBLE=ON` should be done for one of the GCC PR builds but not all of the builds (and not the Intel build which is the current bottle-neck in build times).  We need to have at least one PR build with complex disabled because having the code work with complex enabled does not guarantee that it will work with complex disabled and there are critical customers (e.g. SPARC and EMPIRE) that don't enable complex.

So I would say add `Trilinos_ENABLE_COMPLEX_DOUBLE=ON` to either the gcc-4.8.4 or the gcc-7.2.0 build, but not both.

Agreed?

mhoemmen:
- Salinas developers are the only ones who need complex.  They claim that they likely only use `complex<double>`, not `complex<float>`.  @crdohrm can clarify exactly what Trilinos packages Salinas uses with complex numbers.
  - @theguruat12 is currently testing the Sierra build with various Trilinos complex options.
  - @kddevin A newer GCC version would likely build faster with complex enabled.
  - @bartlettroscoe Yes, please at least enable `complex<double>` in one build.  I will report if Sierra needs more than that.

Thanks all!

jwillenbring:
@mhoemmen @kddevin 

[...]

PR #3940, opened 2018-11-27
Belos SingleRedIter: Fix type bug

cgcgcg:
@trilinos/belos

## Description
Fix recently introduced bug in BelosCGSingleRedIter that affects complex builds.


cgcgcg:
@mhoemmen I agree :-) Could we enable complex for one of the faster autotester builds?

mhoemmen:
@cgcgcg wrote:
> Could we enable complex for one of the faster autotester builds?

Hey @bartlettroscoe @trilinos/framework , here is more support for #2426.

Issue #3787, opened 2018-10-31
Belos: BiCGSTAB is broken


mhoemmen:
@trilinos/belos @vbrunini @jclause 

We noticed that Belos' BiCGSTAB implementation was failing to converge for a small (32 x 32) unpreconditioned nonsymmetric (but not badly so) linear system.  The linear system has real eigenvalues and RCOND about 1.0e9.  Both Belos' GMRES and AztecOO's BiCGSTAB handle it just fine.

I wrote my own Tpetra-only BiCGSTAB implementation based on the "Templates" book, in particular the [Matlab code here](http://www.netlib.org/templates/matlab/bicgstab.m).  It converged just fine, in a number of iterations comparable to Belos' GMRES.

## Motivation and Context

Many users of our application specify BiCGSTAB as a default solver in their input decks.  Its memory use does not depend on the number of iterations / restart length, but it can handle nonsymmetric linear systems.  This makes it a good compromise between CG and GMRES.

We would like BiCGSTAB to work to achieve our goal of replacing use of AztecOO with the Tpetra-based solver stack.

## Possible Solution

  1. As a temporary fix, use the framework in `belos/tpetra/src/solvers` to inject my Tpetra-only solver into Belos::SolverFactory.
  2. Fix `Belos::BiCGStab{Iter,SolMgr}`.  (It's not clear which class is causing the problem.)  It may be easier just to use the known working Tpetra-only implementation as a guide to rewrite the existing implementation.

hkthorn:
@mhoemmen Can you provide the small linear system that illustrated this issue?  I've been looking at the BiCGStab implementation and I want to reproduce what you are seeing.

mhoemmen:
@hkthorn I'll send it to you.  I was out all last week in all-day meetings.

amklinv:
That's unfortunate.  I'm curious to find out what's causing the issue.

mhoemmen:
I added some code to a branch to help with debugging this issue.

0. Receive the bicgstab-A.mm and bicgstab-b.mm files I'll send you.
1. Check out [the "Debug-3787" branch on my Trilinos fork](https://github.com/mhoemmen/Trilinos/tree/Debug-3787).
2. Enable Belos and Ifpack2, and enable Ifpack2 examples.
3. Build Trilinos.
4. Change to the `packages/ifpack2/example/` subdirectory of the build directory, and run the following (with 1 MPI process):
```
[...]

PR #3949, opened 2018-11-28
Ifpack2: Expand "relaxation w/ equilibration" example

mhoemmen:
@trilinos/ifpack2

As part of #3787, I expanded Ifpack2's "relaxation with equilibration" example, into a tool that I found useful for comparing solvers and preconditioners.  I added the following features:

  - Add hand-rolled BiCGSTAB that imitates AztecOO's implementation of BiCGSTAB
  - Add hand-rolled BiCGSTAB that imitates van der Vorst's unpreconditioned BiCGSTAB
  - Add AztecOO + Epetra solver to example (converts from Tpetra to Epetra)
  - Add option to compare with dense (LAPACK) solvers.  We compare with two solvers: an SVD-based least-squares solver, and LU factorization with partial pivoting and equilibration.
  - Add eigenvalue display to dense (LAPACK) solver option
  - Show explicit residual norm for each (iterative) solver
  - Add printSolution option

I think this could be useful for other tasks, so I'm adding it permanently to Ifpack2.

## Related Issues

* Related to #3787


mhoemmen:
Replaced by https://github.com/trilinos/Trilinos/pull/3953 .  Please add reviews there.  Thanks!

PR #3952, opened 2018-11-28
mm_TAFC: Commit to new branch to avoid PR issues

mhoemmen:
I'm making this PR for @DrBooom .


packages/muelu/test/maxwell/Maxwell3D.cpp
```diff
@@ -432,7 +430,7 @@ int MainWrappers<double,LocalOrdinal,GlobalOrdinal,Node>::main_(Teuchos::Command
     tm = Teuchos::null;
 
     if (solverName == "Belos") {
-      tm = rcp(new TimeMonitor(*TimeMonitor::getNewTimer("Maxwell: 2 - Build Belos solver etc")));
+      auto tm2 = TimeMonitor::getNewTimer("Maxwell: 2 - Build Belos solver etc");
```
csiefer2: What does this actually do?  Does this next timers or no?

mhoemmen: @DrBooom explained that this fixed "segfault on exit/return" issues, likely by ensuring that TimeMonitor instances don't persist beyond where they should.

If we think this is broken, let's all get together in a hackathon and fix it together and push it.

csiefer2: In the original code, each timer timed a disjoint region (modulo the usual issues).  These timers  *look* like they're overlapping.  That isn't correct behavior.

mhoemmen: @csiefer2 OK, I agree that this actually would change timer scopes.  This could be relevant to @DrBooom 's timings, though I think he collected timings before making this change.  @DrBooom had to do _something_ here, else the code would crash.  I propose that for this code _only_, we fix it to use TimeMonitor in the correct way: for example:
```c++
{ 
  TimeMonitor tm1 (...);
  thing_to_time_1 ();
}
{
  TimeMonitor tm2 (...);
  thing_to_time_2 ();
}
```
etc.  Does that sound like a good plan?  @DrBooom I will be happy to help you with this.

DrBooom: I've fixed this issue by simply assigning the old timer to null before the creation of the new timer. Is that acceptable?

DrBooom:
I've fixed the scoping of the timers by adding assignment to Teuchos::null before the new timer is created.

Tested, passed and pushed.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -8200,6 +8200,8 @@ namespace Tpetra {
         bool mismatch = source_vals != target_vals;
         immRedErr = MPI_Iallreduce(&mismatch,&reduced_mismatch,1,MPI::BOOL,MPI_LOR,rawComm,&rawRequest);
```
mhoemmen: Trilinos does not test any more with `MPI_VERSION < 3`, so I'm OK with leaving this `MPI_Iallreduce` in place.  Another option would be to use the existing `Tpetra::Details::iallreduce` function, that works regardless of `MPI_VERSION`.  (It uses `MPI_Iallreduce` for `MPI_VERSION >= 3`, and does something else otherwise that is still nonblocking.)  Let's leave that for later, though.

DrBooom:
WRT Jonathan's comments on timers in Maxwell3D, all three cases the timer in question goes out of scope. I'll be pedantic and assign them to null, but it will not change the functionality of the timers.

DrBooom:
Actually, one of those timers should have been set to null tm4 = null before tm5;

packages/muelu/test/maxwell/Maxwell3D.cpp
```diff
@@ -527,8 +524,8 @@ int MainWrappers<double,LocalOrdinal,GlobalOrdinal,Node>::main_(Teuchos::Command
       Teuchos::RCP<Thyra::LinearOpWithSolveBase<Scalar> > thyraInverseA = Thyra::linearOpWithSolve(*solverFactory, thyraA);
 
       comm->barrier();
-      tm = Teuchos::null;
-      tm = rcp(new TimeMonitor(*TimeMonitor::getNewTimer("Maxwell: 3 - Solve")));
+
```
csiefer2: Hate demands you add a tm4= Teuchos::null here.

csiefer2: Just push the PR and fix it later though.

DrBooom: I just caught that one, and it's been pushed.

I'm waiting on serial builds to come back clean...

mhoemmen:
Weirdly, the PR tests report as passing, but the link to Dashboard hasn't been updated. @trilinos/framework  I'm guessing it's OK, since the old link just shows the same build errors that @DrBooom has since fixed.  However, I'd like to see it run through one more cycle just to make sure.  Then I'll push it :-D

  - New link: https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-3952-test&field2=buildstarttime&compare2=84&value2=NOW
  - Old link: https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-3952-test&field2=buildstarttime&compare2=84&value2=NOW

mhoemmen:
I think it's OK.  I don't want to wait too long -- I was hoping the retest would start sooner.  Thanks @DrBooom !!! :D

PR #3953, opened 2018-11-28
Ifpack2: Expand "relaxation w/ equilibration" example

mhoemmen:
@trilinos/ifpack2

As part of #3787, I expanded Ifpack2's "relaxation with equilibration" example, into a tool that I found useful for comparing solvers and preconditioners.  I added the following features:

  - Add hand-rolled BiCGSTAB that imitates AztecOO's implementation of BiCGSTAB
  - Add hand-rolled BiCGSTAB that imitates van der Vorst's unpreconditioned BiCGSTAB
  - Add AztecOO + Epetra solver to example (converts from Tpetra to Epetra).  This introduces an optional AztecOO and Epetra dependency to Ifpack2.  I tested the build with those packages disabled; it works fine.  (The executable prints a useful error message if users try to run AztecOO, but the package is disabled.)
  - Add option to compare with dense (LAPACK) solvers.  We compare with two solvers: an SVD-based least-squares solver, and LU factorization with partial pivoting and equilibration.
  - Add eigenvalue display to dense (LAPACK) solver option
  - Show explicit residual norm for each (iterative) solver
  - Add printSolution option

I think this could be useful for other tasks, so I'm adding it permanently to Ifpack2.

My future plans are to spin off some of these tools into separate new features.  For example, I would like to move the Tpetra-to-Epetra and sparse-to-dense conversion functions to Tpetra.

## Related Issues

* Related to #3787


mhoemmen:
Huh, that's legit, I need to include `<memory>`.

mhoemmen:
Could y'all please re-review?  I fixed a minor build error that doesn't show up in Clang builds.  Thanks!

mhoemmen:
Let's try this again.

mhoemmen:
Thanks all! :D

PR #3974, opened 2018-12-02
Tpetra: Fix #3973 (build warnings in MultiVector unit test)

### LINKED ISSUE #3973: Tpetra: Build warnings in test suggest incorrectness
@trilinos/tpetra 

When I build current Trilinos develop with Clang on Mac, I get many errors in `MultiVector_UnitTests.cpp` that look like this:
```
.../Trilinos/packages/tpetra/core/test/MultiVector/MultiVector_UnitTests.cpp:2180:69: warning: expression result unused [-Wunused-value]
            os << ", B.modified_device: " << B_dv.need_sync_host ()?1:0;
                                                                    ^
.../Trilinos/packages/tpetra/core/test/MultiVector/MultiVector_UnitTests.cpp:2176:68: warning: operator '?:' has lower precedence than
      '<<'; '<<' will be evaluated first [-Wparentheses]
            os << ": A.modified_host: " << A_dv.need_sync_device ()?1:0;
```
That latter warning suggests that code that looks like the above might not be doing what people think.  In particular, I suspect that it does the following:

  1. Evaluates `os << ": A.modified_host: " << A_dv.need_sync_device ()`, which returns `std::ostream&`.
  2. Evaluates `(std::ostream::operator bool()) ? 1 : 0`, which should always return 1 if the stream is valid.

This is certainly not what the author intended.  This should be easy to fix: Just parenthesize the trinary operator invocations.

mhoemmen:
@trilinos/tpetra 

## Related Issues

* Closes #3973

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
@@ -321,8 +321,8 @@ namespace Tpetra {
     ///   null, any missing parameters will be filled in with their
     ///   default values.
     CrsGraph (const Teuchos::RCP<const map_type>& rowMap,
-              size_t maxNumEntriesPerRow,
-              ProfileType pftype = DynamicProfile,
+              const size_t maxNumEntriesPerRow,
+              const ProfileType pftype = DynamicProfile,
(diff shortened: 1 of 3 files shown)


mhoemmen:
Thanks @bartlettroscoe ! :D

Issue #3946, opened 2018-11-28
MueLu: Scalar=float build errors


mhoemmen:
@trilinos/muelu 
Reported by @davydden in [this comment](https://github.com/trilinos/Trilinos/issues/3572#issuecomment-442327278).  Here are the reported build errors:

```
    8158    In file included from /Users/davydden/spack/var/spack/stage/trilinos-develop-p6xmhguuqtomai5fy4a36qmnn4wpqzfq/Trilinos/spack-build/packages/muelu/src/Utils/ExplicitInstantiation/MueLu_Zoltan2Interface.cpp:55:
  >> 8159    /Users/davydden/spack/var/spack/stage/trilinos-develop-p6xmhguuqtomai5fy4a36qmnn4wpqzfq/Trilinos/packages/muelu/src/Rebalancing/MueLu_Zoltan2Interface_def.hpp:197:24: error: no matching constructor for initialization of 'InputAdapterType' (aka 'XpetraMultiVe
             ctorAdapter<MultiVector<float, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> > >')
     8160          InputAdapterType adapter(coords, weights, strides);
     8161                           ^       ~~~~~~~~~~~~~~~~~~~~~~~~
     8162    /Users/davydden/spack/var/spack/stage/trilinos-develop-p6xmhguuqtomai5fy4a36qmnn4wpqzfq/Trilinos/packages/muelu/src/Utils/MueLu_ETI_4arg.hpp:32:45: note: in instantiation of member function 'MueLu::Zoltan2Interface<float, int, int, Kokkos::Compat::KokkosDevi
             ceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >::Build' requested here
     8163      TPETRA_INSTANTIATE_SLGN_NO_ORDINAL_SCALAR(MUELU_ETI_GROUP)
     8164                                                ^
     8165    /Users/davydden/spack/var/spack/stage/trilinos-develop-p6xmhguuqtomai5fy4a36qmnn4wpqzfq/Trilinos/packages/zoltan2/src/input/Zoltan2_XpetraMultiVectorAdapter.hpp:119:3: note: candidate constructor not viable: no known conversion from 'std::vector<const double
              *>' to 'std::vector<const scalar_t *> &' (aka 'vector<const float *> &') for 2nd argument
```

[...]

PR #3977, opened 2018-12-03
MueLu: Fix some float build issues

cgcgcg:
@trilinos/muelu 

## Description
As pointed out in issue #3946, MueLu does not build with `Scalar=float` or `Scalar=complex<float>`. 
I do not think that MueLu has supported using float in the past.
With the above changes, the build goes through as long as 
```
-D MueLu_ENABLE_TESTS=OFF 
-D MueLu_ENABLE_EXAMPLES=OFF
```
I would be surprised if MueLu actually worked though. The reason is that we have a lot of spots were coordinates are hard-coded to double. For an actual fix, we would need to go through and replace all occurrences with `ScalarTraits<Scalar>::magnitudeType`.

### DIFF
--- a/packages/muelu/adapters/xpetra/MueLu_CreateXpetraPreconditioner.hpp
+++ b/packages/muelu/adapters/xpetra/MueLu_CreateXpetraPreconditioner.hpp
@@ -42,7 +42,7 @@ namespace MueLu {
   Teuchos::RCP<MueLu::Hierarchy<Scalar,LocalOrdinal,GlobalOrdinal,Node> >
   CreateXpetraPreconditioner(Teuchos::RCP<Xpetra::Matrix<Scalar,LocalOrdinal,GlobalOrdinal,Node> > op,
                              const Teuchos::ParameterList& inParamList,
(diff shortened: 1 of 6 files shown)


mhoemmen:
> I do not think that MueLu has supported using float in the past.

`Scalar=float` certainly did build in the past.  Does the MueLu team no longer want to support builds with `Scalar=float` enabled?

cgcgcg:
@mhoemmen I don't think that there are any good reasons for not supporting float. Given how widespread the use of `MultiVector<double,...>` for coordinates is, I have doubts that this ever worked in the past..

jhux2:
I don't know of any apps actually *using* float.

cgcgcg:
@lucbv Yep, that's the pattern that should be used. However, `grep -ri 'multivector\s*<double'` gives me 141 hits in MueLu..

lucbv:
@cgcgcg I am OK with us spending an afternoon and purging these 141 hits, it's not really hard and it would not take that long... plus 40 hits come from `example/` and `tests/` and these are not as critical for a first pass.

cgcgcg:
@lucbv I am trying out the nuclear option of grep+sed. Will let you know how well that goes..

PR #3979, opened 2018-12-03
Tpetra: Fix #3976 (build errors due to uint)

### LINKED ISSUE #3976: Tpetra:  Use of "uint" in multiple files causing build failures on Mac OS X, GNU 8.2.x
@trilinos/Tpetra

Something on Mac OS X has removed the definition of "uint" from the GNU 8.2 compiler suite and now I'm getting errors during the compilation of tpetra. I don't believe "uint" is portable and should be replaced with the full "unsigned int", although in many cases it looks like it's not used correctly in tpetra. For instance, core/src/Tpetra_Import_Util2.hpp:235
```
auto NumRecvs                   = Distor.getNumReceives();
.
.
.
for(uint i=0,j=0;i<NumRecvs;i++){
````
In this case "uint i" should be replaced with "auto i" to be correct.

## Your Environment

- **Relevant repo SHA1s:**  a6e9a58e3fe76a40c8a237c97d52e22d201012f3
- **Relevant configure flags or configure script:**  
- **Operating system and version:**  Mac OS X 10.12.6
- **Compiler and TPL versions:**  GNU 8.2.0

### PR DESCRIPTION
**Pull request by mhoemmen**
@trilinos/tpetra

## Related Issues

* Closes #3976

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
@@ -86,9 +86,9 @@ namespace { // (anonymous)
 
   std::shared_ptr< ::Tpetra::Details::CommRequest>
   iallreduceIntRaw (const int& localValue,
-		    int& globalValue,
-		    const ::Teuchos::EReductionType op,
-		    const Teuchos::Comm<int>& comm)
+                    int& globalValue,
+                    const ::Teuchos::EReductionType op,
+                    const Teuchos::Comm<int>& comm)
   {
     Kokkos::View<const int*, Kokkos::HostSpace> localView (&localValue);
     Kokkos::View<int*, Kokkos::HostSpace> globalView (&globalValue);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp [hunk 1]
```diff
@@ -8937,32 +8937,37 @@ namespace Tpetra {
...
-                    int col=colind[j];
-                    if(IsOwned[col] && SentTo[col]!=pid){
-                        SentTo[col]    = pid;
-                        GO gid = this->getColMap()->getGlobalElement(col); // *this is sourcematrix
-                        usrtg.push_back(std::pair<int,GO>(pid,gid));
-                    }
-            }
-        }
+	{
+	  const auto& colMap = * (this->getColMap ()); // *this is sourcematrix
```
mhoemmen: Calling `getColMap()` invokes RCP's constructor, and thus touches the reference count.  This costs something in tight loops.

PR #3985, opened 2018-12-03
Tpetra: beginFill/endFill helper functions

csiefer2:
Auto-PR for SHA f2c49f1

### DIFF
--- a/packages/tpetra/core/example/Finite-Element-Assembly/fem_assembly_InsertGlobalIndices_DP.hpp
+++ b/packages/tpetra/core/example/Finite-Element-Assembly/fem_assembly_InsertGlobalIndices_DP.hpp
@@ -53,6 +53,7 @@
 #include <Teuchos_CommandLineProcessor.hpp>
 #include <Teuchos_RCP.hpp>
 #include <Teuchos_FancyOStream.hpp>
+#include <Tpetra_Assembly_Helpers.hpp>
 
 #include "fem_assembly_typedefs.hpp"
 #include "fem_assembly_MeshDatabase.hpp"
@@ -236,7 +237,7 @@ int executeInsertGlobalIndicesDP_(const comm_ptr_t& comm, const struct CmdLineOp
   Teuchos::Array<Scalar> column_scalar_values(4);         // scalar values for each column
 
   // Loop over elements
-  rhs->beginFill();
+  Tpetra::beginFill(*rhs);
   for(size_t element_gidx=0; element_gidx<mesh.getNumOwnedElements(); element_gidx++)
   {
     // Get the contributions for the current element
@@ -277,7 +278,7 @@ int executeInsertGlobalIndicesDP_(const comm_ptr_t& comm, const struct CmdLineOp
   {
     // Global assemble the RHS
     TimeMonitor timer(*TimeMonitor::getNewTimer("5) GlobalAssemble (RHS)"));
-    rhs->endFill();
+    Tpetra::endFill(*rhs);
   }
 
 
--- a/packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
+++ b/packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
@@ -45,21 +45,21 @@
 namespace Tpetra {
 
 template<typename T>
-void beginFill(T t) { t.beginFill();}
+void beginFill(T& t) { t.beginFill();}
 
 template<typename T>
-void endFill(T t) { t.endFill();}
+void endFill(T& t) { t.endFill();}
 
 //! Helper function to begin fill on multiple FE{MultiVector|CrsMatrix} objects at once
 template<typename T, typename... Args>
-void beginFill(T t, Args... args) {
+void beginFill(T& t, Args... args) {
   t.beginFill();
   beginFill(args...);
 }
 
 //! Helper function to begin end on multiple FE{MultiVector|CrsMatrix} objects at once
 template<typename T, typename... Args>
-void endFill(T t, Args... args) {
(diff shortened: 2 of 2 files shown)


packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+template<typename T>
+void beginFill(T t) { t.beginFill();}
+
+template<typename T>
+void endFill(T t) { t.endFill();}
+
+//! Helper function to begin fill on multiple FE{MultiVector|CrsMatrix} objects at once
+template<typename T, typename... Args>
+void beginFill(T t, Args... args) {
```
mhoemmen: Pass `T` by reference here.  Also, suppose that we wanted two objects -- say CrsMatrix and Vector -- to share communication during global assembly.  I think that could work here, by overloading `beginFill` and `endFill` as follows:
```c++
template<class SC, class LO, class GO, class NT, class... Args>
void
beginFill (CrsMatrix<SC, LO, GO, NT>& A, MultiVector<SC, LO, GO, NT>& B, Args... args)
{
  special_combined_begin_fill (A, B);
  beginFill (args...);
}
```
Would that be how you might do this?

packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
```diff
@@ -0,0 +1,69 @@
...
+// ************************************************************************
+// @HEADER
+
+#ifndef TPETRA_ASSEMBLY_HELPERS_DECL_HPP
+#define TPETRA_ASSEMBLY_HELPERS_DECL_HPP
+
+namespace Tpetra {
+
+template<typename T>
+void beginFill(T t) { t.beginFill();}
```
mhoemmen: Pass `T` by nonconst reference, not value: e.g., template<class T> void beginFill (T& t) { t.beginFill (); }`

packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+template<typename T>
+void endFill(T t) { t.endFill();}
```
mhoemmen: Pass `T` by reference here.

packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
```diff
@@ -0,0 +1,69 @@
...
+//! Helper function to begin fill on multiple FE{MultiVector|CrsMatrix} objects at once
+template<typename T, typename... Args>
+void beginFill(T t, Args... args) {
+  t.beginFill();
+  beginFill(args...);
+}
+
+//! Helper function to begin end on multiple FE{MultiVector|CrsMatrix} objects at once
+template<typename T, typename... Args>
+void endFill(T t, Args... args) {
```
mhoemmen: Pass `T` by reference here.  Also, see above.

packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
mhoemmen: Bikeshed on the function names: "fill" might suggest something like `std::fill` or `Kokkos::deep_copy(array, value)`.  Would it make sense to say "assembly" instead, like `beginGlobalAssembly` / `endGlobalAssembly`?

packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
csiefer2: Fixing the reference issue.

As for "combo comm", that's kind of what I'm thinking.  We'd need some sorcery for that, but it would be all template specialization and would not be visible to use code.

packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
csiefer2: Fixed the reference issue.

I have no problem w/ changing the naming scheme on the FE* family, so long as we all agree on what it ought to be.

packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
csiefer2: Done

packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
csiefer2: Done

packages/tpetra/core/example/Finite-Element-Assembly/fem_assembly_InsertGlobalIndices_DP.hpp
```diff
@@ -277,7 +278,7 @@ int executeInsertGlobalIndicesDP_(const comm_ptr_t& comm, const struct CmdLineOp
   {
     // Global assemble the RHS
     TimeMonitor timer(*TimeMonitor::getNewTimer("5) GlobalAssemble (RHS)"));
-    rhs->endFill();
```
mhoemmen: @csiefer2 Is the intent that users should always call the beginFill and endFill nonmember functions, and never the instance methods?  If so, we could make the one-argument overloads of the nonmember functions friends (defined inline), and make the instance methods private.

csiefer2: There's no real harm in calling the members (unless we add code for comm batching).  It just generates more user code.

PR #3986, opened 2018-12-04
Muelu: Fix build issues with Scalar=float and Scalar=complex<float>

cgcgcg:
@trilinos/muelu 

## Description
I grep+sed replaced instances `object<double,...>` by `object<typename Teuchos::ScalarType<Scalar>::magnitudeType,...>` and cleaned up afterwards. Builds using `Scalar=float` or `Scalar=complex<float>` seem to be compiling ok now. There are still test errors due to a variety of reasons, e.g. tolerances that are too tight for `float`, and other packages like Ifpack2 asking for `double`-valued coordinates.


cgcgcg:
I am not going to hit the merge button yet in order to give people a chance to complain..

lucbv:
@cgcgcg I am OK with this PR, I think in the long run we might want to really create somthing like `typedef typename Real = Teuchos::ScalarType<Scalar>::magnitudeType` and stick it in some generic header, maybe the one defining the short names?

PR #3993, opened 2018-12-04
MueLu: Fixing compiler error on trappist

csiefer2:
Auto-PR for SHA 281cac4

### DIFF
--- a/packages/muelu/test/maxwell/Maxwell3D.cpp
+++ b/packages/muelu/test/maxwell/Maxwell3D.cpp
@@ -220,7 +220,7 @@ int MainWrappers<Scalar,LocalOrdinal,GlobalOrdinal,Node>::main_(Teuchos::Command
     tm = Teuchos::null;
 
     if (solverName == "Belos") {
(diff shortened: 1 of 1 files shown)


packages/muelu/test/maxwell/Maxwell3D.cpp
```diff
@@ -220,7 +220,7 @@ int MainWrappers<Scalar,LocalOrdinal,GlobalOrdinal,Node>::main_(Teuchos::Command
     tm = Teuchos::null;
 
     if (solverName == "Belos") {
-        auto tm2 = TimeMonitor(*TimeMonitor::getNewTimer("Maxwell: 2 - Build Belos solver etc"));
+      auto tm2  = TimeMonitor::getNewTimer("Maxwell: 2 - Build Belos solver etc");
```
mhoemmen: I don't think this would work like you think -- getNewTimer doesn't start the retrieved Teuchos::Time object, it just gets a pointer to it.  The result is that you wouldn't end up timing the thing.  You would just get a zero time.

csiefer2: That call syntax is used in other places in the same file.  Not saying it works.  But it does compile.

PR #3996, opened 2018-12-05
Tpetra: Hopefully fix #3968

### LINKED ISSUE #3968: Tpetra: Possible OpenMPI bug; don't enable CUDA-aware MPI support
@trilinos/tpetra @trilinos/framework 

When debugging #3500, @rppawlo found a possible bug in OpenMPI 2.1.2 with CUDA-aware MPI enabled.  When doing a Tpetra::Export between two MultiVectors with AbsMax CombineMode, the target MultiVector sometimes doesn't get the right result.  Roger found that on unpack, the receive buffer doesn't always have the right values in it.  It's not consistently reproducible -- Roger sometimes has to run the test three times to get this to happen -- but it does happen fairly often.  Roger thinks that MPI might not have finished writing to the buffer, on a process (Proc 0) that has no sends but some receives, even though Process 0 should have done an `MPI_Waitall` on the receives (they are `MPI_Irecv`, always).  The results are _always_ correct when running with CUDA-aware MPI disabled (that's a Tpetra option, not an MPI option).

Next steps:
  - [x] Make sure this isn't an issue with `Kokkos::atomic_assign` in AbxMax
  - [ ] Test with different OpenMPI versions
  - [ ] Write a short reproducer, to make sure this isn't a Tpetra issue (start perhaps with @crtrott 's CG solve)

We recommend disabling Tpetra's CUDA-aware MPI option for now.

mhoemmen:
@trilinos/tpetra @rppawlo 

## Description

Hopeful fix for #3968.  I was planning to push just the test first, so I could see it fail on the Dashboard and thus justify the fix.  However, I accidentally included the fix with the test.  We can still exercise the test manually by disabling the fix.  We know that the existing code was wrong, so the fix is still a good idea.

## Motivation and Context

Panzer needs Tpetra to fix #3968 so that global unknown numbering works.

## Related Issues

* Related to #3968


packages/tpetra/core/test/ImportExport/Issue_3968.cpp
```diff
@@ -0,0 +1,205 @@
...
+    const GO expectedValue = baseValue () + incrValue;
+    TEST_EQUALITY( X_lcl_1d_h(0), expectedValue );
+
+    int lclSuccess = success ? 1 : 0;
+    int gblSuccess = 0; // output argument
+    using Teuchos::outArg;
+    using Teuchos::REDUCE_MIN;
+    using Teuchos::reduceAll;
+
+    reduceAll (*comm, REDUCE_MIN, lclSuccess, outArg (gblSuccess));
```
mhoemmen: доверять, но проверить

packages/tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp
```diff
@@ -736,23 +736,23 @@ outOfBounds (const IntegerType x, const IntegerType exclusiveUpperBound)
...
+    static Scalar1 apply(const Scalar1& val1, const Scalar2& val2) {
+      const auto val1_abs = Kokkos::ArithTraits<Scalar1>::abs (val1);
+      const auto val2_abs = Kokkos::ArithTraits<Scalar2>::abs (val2);
+      return val1_abs > val2_abs ? Scalar1 (val1_abs) : Scalar1 (val2_abs);
+    }
+  };
+
   template<class ExecutionSpace>
   struct AbsMaxOp {
-    // ETP:  Is this really what we want?  This seems very odd if
```
DrBooom: Can this comment be condensed down to 1-2 lines.

rppawlo:
@mhoemmen - I missed this PR. I put up the AbsMax fix in #3999 Is this the sme code that was posted on #3968 ? It didn't compile on waterman without some minor changes.

mhoemmen:
@rppawlo If you've tested #3999 I would prefer that -- thanks!

mhoemmen:
Superseded by PR #4024 .

PR #3999, opened 2018-12-05
Tpetra: fix race condition in transfers with AbsMax for CUDA_AWARE_MPI

rppawlo:
#3968

Changes to each package are in separate commits.

Thanks to Mark Hoemmen for the tpetra fix!

### DIFF
--- a/packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
+++ b/packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
@@ -124,10 +124,10 @@ public:
                                     const std::vector<std::pair<int,LocalOrdinal> > & pid_and_lid) const
   {
     // always choose index of pair with smallest pid
-    auto numLids = pid_and_lid.size();
-    decltype(numLids) idx = 0;
-    auto minpid = pid_and_lid[0].first;
-    decltype(minpid) minidx = 0;
+    const std::size_t numLids = pid_and_lid.size();
+    std::size_t idx = 0;
+    int minpid = pid_and_lid[0].first;
+    std::size_t minidx = 0;
     for (idx = 0; idx < numLids; ++idx) {
       if (pid_and_lid[idx].first < minpid) {
         minpid = pid_and_lid[idx].first;
@@ -569,17 +569,22 @@ void DOFManager<LO,GO>::buildGlobalUnknowns(const Teuchos::RCP<const FieldPatter
     HashTable isOwned, remainingOwned;
 
     // owned_ is made up of owned_ids.: This doesn't work for high order
-    Teuchos::ArrayRCP<const GO> nvals = non_overlap_mv->get1dView();
-    Teuchos::ArrayRCP<const GO> tagged_vals = tagged_non_overlap_mv->get1dView();
+    non_overlap_mv->sync_host();
+    tagged_non_overlap_mv->sync_host();
+    PHX::Device::fence();
+    auto nvals = non_overlap_mv->template getLocalView<PHX::Device>();
+    auto tagged_vals = tagged_non_overlap_mv->template getLocalView<PHX::Device>();
     TEUCHOS_ASSERT(nvals.size()==tagged_vals.size());
-    for (int j = 0; j < nvals.size(); ++j) {
-      if (nvals[j] != -1) {
-        for(GO offset=0;offset<tagged_vals[j];offset++)
-          isOwned.insert(nvals[j]+offset);
-      }
-      else {
-        // sanity check
-        TEUCHOS_ASSERT(tagged_vals[j]==0)
+    for (size_t i = 0; i < nvals.extent(1); ++i) {
+      for (size_t j = 0; j < nvals.extent(0); ++j) {
+	if (nvals(j,i) != -1) {
+	  for(GO offset=0;offset<tagged_vals(j,i);++offset)
+	    isOwned.insert(nvals(j,i)+offset);
+	}
+	else {
+	  // sanity check
+	  TEUCHOS_ASSERT(tagged_vals(j,i)==0);
+	}
       }
     }
     remainingOwned = isOwned;
@@ -776,7 +781,6 @@ DOFManager<LO,GO>::buildGlobalUnknowns_GUN(const Tpetra::MultiVector<GO,LO,GO,pa
     non_overlap_mv = rcp(new MultiVector(*tagged_non_overlap_mv,Teuchos::Copy));
   }
 
-
  /* 10. Compute the local sum using Kokkos.
    */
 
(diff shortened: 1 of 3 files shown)


packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
```diff
@@ -805,7 +810,7 @@ DOFManager<LO,GO>::buildGlobalUnknowns_GUN(const Tpetra::MultiVector<GO,LO,GO,pa
 
     // do a prefix sum
     GO scanResult = 0;
-    Teuchos::scan<int, GO> (*getComm(), Teuchos::REDUCE_SUM, static_cast<size_t> (localsum), Teuchos::outArg (scanResult));
+    Teuchos::scan<int, GO> (*getComm(), Teuchos::REDUCE_SUM, static_cast<int> (localsum), Teuchos::outArg (scanResult));
```
mhoemmen: This is only helpful for removing warnings if `GO` is signed.  You really want to cast `localsum` to `GO` since that's the type of the scan.  It will get cast from `int` to `GO` here anyway.

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
```diff
@@ -930,7 +938,10 @@ DOFManager<LO,GO>::buildTaggedMultiVector(const ElementBlockAccess & ownedAccess
 
     // temporary working vector to fill each row in tagged array
     std::vector<int> working(overlap_mv->getNumVectors());
-    ArrayRCP<ArrayRCP<GO> > edittwoview = overlap_mv->get2dViewNonConst();
+    auto dual_view = overlap_mv->getDualView();
```
mhoemmen: I would prefer that you not call `getDualView`.  Please call the functions that get the Kokkos::View of the appropriate memory space directly.  This is fine for now, but I do plan on getting rid of `getDualView` at some point, since it exposes an implementation detail that constrains Tpetra to have allocations on both sides at all times (it can't do lazy allocation).

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
```diff
@@ -930,7 +938,10 @@ DOFManager<LO,GO>::buildTaggedMultiVector(const ElementBlockAccess & ownedAccess
...
+    dual_view.sync_host();
+    PHX::Device::fence();
+    auto edittwoview_host = dual_view.view_host();
     for (size_t b = 0; b < blockOrder_.size(); ++b) {
       // there has to be a field pattern assocaited with the block
```
mhoemmen: "associated"

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
```diff
@@ -1304,7 +1322,10 @@ fillGIDsFromOverlappedMV(const ElementBlockAccess & access,
   using Teuchos::ArrayRCP;
 
   //To generate elementGIDs we need to go through all of the local elements.
-  ArrayRCP<ArrayRCP<const GO> > twoview = overlap_mv.get2dView();
+  auto dual_view = overlap_mv.getDualView();
```
mhoemmen: See above.  Get Kokkos::View objects from the Tpetra::MultiVector directly; don't access the DualView.

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
```diff
@@ -569,17 +569,22 @@ void DOFManager<LO,GO>::buildGlobalUnknowns(const Teuchos::RCP<const FieldPatter
...
     TEUCHOS_ASSERT(nvals.size()==tagged_vals.size());
-    for (int j = 0; j < nvals.size(); ++j) {
-      if (nvals[j] != -1) {
-        for(GO offset=0;offset<tagged_vals[j];offset++)
-          isOwned.insert(nvals[j]+offset);
-      }
-      else {
-        // sanity check
-        TEUCHOS_ASSERT(tagged_vals[j]==0)
+    for (size_t i = 0; i < nvals.extent(1); ++i) {
```
mhoemmen: I think Kokkos::View has a `size_type` typedef that would be the best thing to use here.

rppawlo: Kokkos used to map the return type to a NODE dependency but when they moved to extent() it got hard coded to size_t. Here's the code from the View object:
```
  template< typename iType >
  KOKKOS_INLINE_FUNCTION constexpr
  typename std::enable_if< std::is_integral<iType>::value , size_t >::type
  extent( const iType & r ) const
    { return m_map.extent(r); }
```

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
rppawlo: thanks! will change

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
rppawlo: ok. Will fix that as well.

rppawlo:
@mhoemmen - I'm going to remove the automerge for now since I need to push the changes you requested.

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
mhoemmen: Ah, OK, sorry about that!

rppawlo:
@mhoemmen - changes have been pushed. Could you review and reapprove?

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
```diff
@@ -789,13 +789,9 @@ DOFManager<LO,GO>::buildGlobalUnknowns_GUN(const Tpetra::MultiVector<GO,LO,GO,pa
...
     PANZER_DOFMGR_FUNC_TIME_MONITOR("panzer::DOFManager::buildGlobalUnknowns_GUN::line_07-09 local_count");
-
-    typedef typename Tpetra::MultiVector<GO,Node> MV;
-    typedef typename MV::dual_view_type::t_dev KV;
-    typedef typename MV::dual_view_type::t_dev::memory_space DMS;
-    KV values = non_overlap_mv->template getLocalView<DMS>();
+    auto values = non_overlap_mv->getLocalViewDevice();
     auto mv_size = values.extent(0);
-    Kokkos::parallel_reduce(mv_size,panzer::dof_functors::SumRank2<GO,KV>(values),localsum);
+    Kokkos::parallel_reduce(mv_size,panzer::dof_functors::SumRank2<GO,decltype(values)>(values),localsum);
```
mhoemmen: In the future, please consider using an explicit RangePolicy instead of putting the dimension as the first argument of `Kokkos::parallel_reduce`.  This would let you execute on the `execution_space` corresponding to Tpetra's data, rather than on Kokkos' default execution space.  Tpetra's default execution space could differ from Kokkos' default execution space.  (Some people really want Serial for Tpetra but CUDA for everything else.)

PR #4001, opened 2018-12-05
Tpetra: Updating AssemblyHelper Refeence fu

csiefer2:
Auto-PR for SHA bdf0454

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
+++ b/packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
@@ -39,8 +39,9 @@
 // ************************************************************************
 // @HEADER
 
-#ifndef TPETRA_ASSEMBLY_HELPERS_DECL_HPP
-#define TPETRA_ASSEMBLY_HELPERS_DECL_HPP
+#ifndef TPETRA_ASSEMBLY_HELPERS_HPP
+#define TPETRA_ASSEMBLY_HELPERS_HPP
+#include <utility>
 
 namespace Tpetra {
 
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_Assembly_Helpers.hpp
```diff
@@ -52,18 +53,18 @@ void endFill(T& t) { t.endFill();}
 
 //! Helper function to begin fill on multiple FE{MultiVector|CrsMatrix} objects at once
 template<typename T, typename... Args>
-void beginFill(T& t, Args... args) {
+void beginFill(T& t, Args&&... args) {
   t.beginFill();
-  beginFill(args...);
+  beginFill(std::forward<Args>(args)...);
```
mhoemmen: yeah I missed that one it helps in case you have a `&&` kind of reference

csiefer2: @dsunder has skills.

mhoemmen: u gotta remind him of that he forgets sometimes

dsunder: The following code will work even better.  The problem with the cool argument recursion is that the number of symbols in the symbol table starts to explode, and since compile time is affected by the number of symbols our code gets slower and slower to compile.  The following code will only generate two additional symbols and will not create additional scopes.

```c++
namespace Tpetra {

namespace Impl {
// Helper function to transform a c++11 pack 
// since parameter expansion only happens in 
// limited contexts
template <typename... Args>
inline transform_pack(Args &&... args) {}
} // namespace Impl

template <typename... Args>
void beginFill(Args &&... args)
{
  //use the comma operator to transform a potentially void function
  // into a argument to allow proper parameter expansion for c++11
  Impl::transform_pack( (args.beginFill(),1)... );
}

template <typename... Args>
void endFill(Args &&... args)
{
  //use the comma operator to transform a potentially void function
  // into a argument to allow proper parameter expansion for c++11
  Impl::transform_pack( (args.endFill(),1)... );
}

} //namespace Tpetra
```

PR #4005, opened 2018-12-05
Phalanx default option for multiple evaluator field registration

### LINKED ISSUE #3995: Phalanx: turn PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD off by default
@trilinos/Phalanx

## Expectations

I _think_ more likely than not one wants to see an error if two evaluators evaluate the same field. I am ok with keeping the feature (CMakeLists.txt explain it is for backward compatibility), but should not be the default. It is more costly to debug an app that wrongly has two evaluators evaluating the same field, than it is for another app to add the configure option to allow this. The former, can take hours of debugging (I did this mistake _twice_); the latter takes one failed run (with meaningful error message) to adjust the installation.

## Current Behavior

The flag is on by default, so PHX allows multiple evaluators to evaluate the same field, in which case, the first is used, and the others tossed. If one does _not_ want this feature, the option must be explicitly turned off at configure time.

Note: this option was introduced in 2015, and left ON by default for bwd compatibility in existing codes. I believe 3 years is a reasonable time for codes to either adapt, or add a config option to their trilinos configuration.

## Motivation and Context

It is not unlikely to fall in the scenario where 2 evaluators evaluate the same field. Perhaps you used to have `OldEvaluator` to compute `F`, but now you wrote `VeryCoolNewEvaluator` to evaluate `F`. So you add it to your problem, but forget to delete the previous one. Or worse yet: two evaluators _happen_ to evaluate the same field because of a particular clash in your app parameters, in which case, you probably need to rethink your app flow (so you would _love_ to see an error message).

## Possible Solution

Simply change the default to OFF. Or remove the option altogether. It may be that it was meant to achieve the same thing that _contributed_ fields now allow. So _perhaps_ this is no longer needed.

## Steps to Reproduce

Take your favorite problem, and register the same evaluator twice, and see if you get an error. Yes, in that case everything would be fine, since the two evaluators are the same, but it still exposes the problem: no errors with the default PHX configuration.

rppawlo:
Closes #3995 

If panzer is not enabled, phalanx by default will not allow multiple evaluators to be registered that evaluate the same field. We have to allow multiple registration when panzer is enabled due to a design issue.

### DIFF
--- a/packages/phalanx/CMakeLists.txt
+++ b/packages/phalanx/CMakeLists.txt
@@ -15,7 +15,7 @@ TRIBITS_ADD_SHOW_DEPRECATED_WARNINGS_OPTION()
 TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD
   PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD
   "Enable to use backwards compatible mode where multiple evaluators can be registered that evalaute the same field. Phalanx will use the first evaluator it finds. Default to OFF but unless Panzer is enabled. For backwards compatibility with Panzer users, we default this to ON."
-  ${PROJECT_NAME}_ENABLE_Panzer )
+  "${${PROJECT_NAME}_ENABLE_Panzer}" )
 MESSAGE(STATUS "Allow Multiple Evaluator Registration for Same Field: ${${PACKAGE_NAME}_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD}")
 
 TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_EXPLICIT_TEMPLATE_INSTANTIATION
--- a/packages/phalanx/CMakeLists.txt
+++ b/packages/phalanx/CMakeLists.txt
@@ -14,12 +14,13 @@ TRIBITS_ADD_SHOW_DEPRECATED_WARNINGS_OPTION()
 
 TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD
   PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD
-  "Enable to use backwards compatible mode where multiple evaluators can be registered that evalaute the same field. Phalanx will use the first evaluator it finds. This should default to off but for backwards compatibility with current codes, we default this to ON."
-  ON )
+  "Enable to use backwards compatible mode where multiple evaluators can be registered that evalaute the same field. Phalanx will use the first evaluator it finds. Default to OFF but unless Panzer is enabled. For backwards compatibility with Panzer users, we default this to ON."
+  ${PROJECT_NAME}_ENABLE_Panzer )
+MESSAGE(STATUS "Allow Multiple Evaluator Registration for Same Field: ${${PACKAGE_NAME}_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD}")
 
 TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_EXPLICIT_TEMPLATE_INSTANTIATION
   PHX_ETI
-  "Build with explicit template instation enabled."
+  "Build with explicit template instation enabled in the examples."
   OFF )
 
 # If not explicitly defined by the user, pick a default node type that
@@ -98,6 +99,9 @@ TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_TEUCHOS_TIME_MONITOR
   "Build with Teuchos TimeMonitors wrapped around each evaluator."
   "${${PROJECT_NAME}_ENABLE_TEUCHOS_TIME_MONITOR}" )
 
+# Use only for CUDA builds
+MESSAGE(STATUS "Sacado Hierarchic DFAD support: ${Sacado_ENABLE_HIERARCHICAL_DFAD}")
+
 # Experimental device dag support
 TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_DEVICE_DAG
   PHX_ENABLE_DEVICE_DAG
@@ -124,8 +128,6 @@ IF(${PACKAGE_NAME}_ENABLE_KOKKOS_AMT)
   MESSAGE(FATAL_ERROR "Phalanx configure ERROR: Kokkos AMT support is enabled but is not implemented for Pthreads execution space. Please change your execution space or disable AMT support in Phalanx.")
 ENDIF()
 ENDIF(${PACKAGE_NAME}_ENABLE_KOKKOS_AMT)
-
-MESSAGE(STATUS "Sacado Hierarchic DFAD support: ${Sacado_ENABLE_HIERARCHICAL_DFAD}")
 
 # For use by Phalanx and Panzer unit tests
 GLOBAL_SET(PHALANX_UNIT_TEST_MAIN


packages/phalanx/CMakeLists.txt
```diff
@@ -14,12 +14,13 @@ TRIBITS_ADD_SHOW_DEPRECATED_WARNINGS_OPTION()
 
 TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD
   PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD
-  "Enable to use backwards compatible mode where multiple evaluators can be registered that evalaute the same field. Phalanx will use the first evaluator it finds. This should default to off but for backwards compatibility with current codes, we default this to ON."
-  ON )
+  "Enable to use backwards compatible mode where multiple evaluators can be registered that evalaute the same field. Phalanx will use the first evaluator it finds. Default to OFF but unless Panzer is enabled. For backwards compatibility with Panzer users, we default this to ON."
+  ${PROJECT_NAME}_ENABLE_Panzer )
```
bartlettroscoe: @rppawlo, 

the above always evaluates to `TRUE` which is not what you want.

This needs to be changed to:

```
"${${PROJECT_NAME}_ENABLE_Panzer}"
```

with the quotes for the reason described in:

* https://tribits.org/doc/TribitsDevelopersGuide.html#cmake-language-overview-and-gotchas

(see `EMPTY_VAR`)

The issue is that TriBITS "ENABLE" vars can be `ON` or `OFF` or empty "".    For the latter case, you need the quotes.

bartlettroscoe: @rppawlo, let me pull the branch and check this real quick.

bartlettroscoe:
@rppawlo, 

using the configure script:

```
#!/bin/bash

cmake \
-GNinja \
-DTrilinos_TRIBITS_DIR:STRING=TriBITS/tribits \
-DTPL_ENABLE_MPI=ON \
-DCMAKE_BUILD_TYPE=RELEASE \
-DTrilinos_ENABLE_DEBUG=ON \
-DBUILD_SHARED_LIBS=ON \
-DTrilinos_ENABLE_SECONDARY_TESTED_CODE=OFF \
-DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/MpiReleaseDebugSharedPtSettings.cmake,cmake/std/BasicCiTestingSettings.cmake,cmake/std/sems/SEMSDevEnv.cmake \
-DTrilinos_ENABLE_DEBUG_SYMBOLS=ON \
-DTrilinos_ENABLE_EXPLICIT_INSTANTIATION=ON \
-DTrilinos_ENABLE_CONFIGURE_TIMING=ON \
-DTrilinos_ENABLE_TESTS=ON \
-DTrilinos_TRACE_ADD_TEST=ON \
-DCTEST_BUILD_FLAGS=-j16 \
-DCTEST_PARALLEL_LEVEL=16 \
"$@" \
../../../Trilinos
```

I did:

```
$ . ~/Trilinos.base/Trilinos/cmake/load_sems_dev_env.sh

$ ./do-configure -DTrilinos_ENABLE_Phalanx=ON &> configure.out
```

The configure output showed:

```
Final set of enabled packages:  Kokkos Teuchos KokkosKernels Sacado Shards Phalanx 6
```

And the CMakeCache.txt file showed:

```
$ grep _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD CMakeCache.txt 
Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=Trilinos_ENABLE_Panzer
PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:INTERNAL=ON
```

That is not correct, is it?

bartlettroscoe:
@rppawlo, with the commit d4bda03, I just pushed, I get:

```
$ rm -r CMake* ; ./do-configure -DTrilinos_ENABLE_Phalanx=ON &> configure.out ; grep _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD CMakeCache.txt
Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=
PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:INTERNAL=OFF

$ rm -r CMake* ; ./do-configure -DTrilinos_ENABLE_Phalanx=ON -DTrilinos_ENABLE_Panzer=OFF &> configure.out ; grep _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD CMakeCache.txt
Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=OFF
PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:INTERNAL=OFF

$ rm -r CMake* ; ./do-configure -DTrilinos_ENABLE_Phalanx=ON -DTrilinos_ENABLE_Panzer=ON &> configure.out ; grep _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD CMakeCache.txt
Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=ON
PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:INTERNAL=ON
```

That looks correct, no?  NOTE: Empty "" in CMake means `OFF`.

What are you doing and seeing?

rppawlo:
@bartlettroscoe - can you reapprove this? It passed testing but still needs approval to merge.

I'm digging into my configure now...

bartlettroscoe:
And to show the right macro is defined in the configured header `Phalanx_config.hpp`:

```
$ rm -r CMake* ; ./do-configure -DTrilinos_ENABLE_Phalanx=ON &> configure.out ; grep _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD CMakeCache.txt ; grep  _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD packages/phalanx/src/Phalanx_config.hpp 
Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=
PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:INTERNAL=OFF
/* #undef PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD */

$ rm -r CMake* ; ./do-configure -DTrilinos_ENABLE_Phalanx=ON -DTrilinos_ENABLE_Panzer=OFF &> configure.out ; grep _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD CMakeCache.txt ; grep  _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD packages/phalanx/src/Phalanx_config.hpp 
Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=OFF
PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:INTERNAL=OFF
/* #undef PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD */

$ rm -r CMake* ; ./do-configure -DTrilinos_ENABLE_Phalanx=ON -DTrilinos_ENABLE_Panzer=ON &> configure.out ; grep _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD CMakeCache.txt ; grep  _ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD packages/phalanx/src/Phalanx_config.hpp 
Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:BOOL=ON
PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD:INTERNAL=ON
#define PHX_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD

```

This shows that having the CMake var `Phalanx_ALLOW_MULTIPLE_EVALUATORS_FOR_SAME_FIELD` be empty "" is just fine because empty is false in CMake.

I will go ahead and approve the PR.

bartlettroscoe:
@rppawlo, to explain what you likely were dealing with, see:

* https://tribits.org/doc/TribitsDevelopersGuide.html#cmake-language-overview-and-gotchas

which states:

> The handling of variables is one area where CMake is radically different from most other languages. First, a variable that is not defined simply returns nothing. What is surprising to most people about this is that it does not even return an empty string that would register as an array element! 

and:

> This is a common error that people make when they call CMake functions (built-in or TriBITS-defined) involving variables that might be undefined or empty. 

I think the latter is what bit you.

PR #4013, opened 2018-12-06
Fixed some issues with panzer not working for multi blocks

bathmatt:
There is a bug if panzer runs on multi blocks where all the blocks are not on all the MPI ranks.  This fixes the two that I found... Not sure if they are complete, doubt they are..

@rppawlo @eric-c-cyr can one of you look this over and accept?


rppawlo:
Need to add a unit test for this :)

I will open another ticket for that.

bathmatt:
It is hard to reproduce with the inline mesher...   but I agree, if we can merge this asap I won't have to backport the patch to our branch..

PR #4024, opened 2018-12-10
Tpetra: Add unit test for #3968

mhoemmen:
@trilinos/tpetra @rppawlo 

This was originally part of PR #3996.  @rppawlo pushed an improved fix to #3968 first in PR #3999, which got merged.  I thus deleted PR #3996. This commit includes just the unit test that was part of PR #3996, but not part of PR #3999.

## Related Issues

* Follows #3999 
* Related to #3968

### DIFF
--- a/packages/tpetra/core/test/ImportExport/Issue_3968.cpp
+++ b/packages/tpetra/core/test/ImportExport/Issue_3968.cpp
@@ -61,13 +61,15 @@ namespace { // (anonymous)
   using LO = Tpetra::Map<>::local_ordinal_type;
 
   // Try to get a 64-bit GlobalOrdinal type; that should be more
-  // likely to manifest atomic update issues.
+  // likely to manifest atomic update issues.  If we can't, though,
+  // then a 32-bit type is fine.  long long is guaranteed to be at
+  // least 64 bits, so prefer that over long or unsigned long.
 #if defined(HAVE_TPETRA_INST_INT_LONG_LONG)
-  using GO = long long;
-#elif defined(HAVE_TPETRA_INST_INT_LONG) && sizeof(long) >= 8
-  using GO = long;
-#elif defined(HAVE_TPETRA_INST_INT_UNSIGNED_LONG) && sizeof(unsigned long) >= 8
-  using GO = unsigned long;
+  using GO = long long; // always at least 64 bits
+#elif defined(HAVE_TPETRA_INST_INT_LONG)
+  using GO = long; // may be 64 or 32 bits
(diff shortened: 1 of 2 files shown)


mhoemmen:
I just fixed the build issue.  Let's try this again.

packages/tpetra/core/test/ImportExport/Issue_3968.cpp
```diff
@@ -0,0 +1,206 @@
...
+  // least 64 bits, so prefer that over long or unsigned long.
+#if defined(HAVE_TPETRA_INST_INT_LONG_LONG)
+  using GO = long long; // always at least 64 bits
+#elif defined(HAVE_TPETRA_INST_INT_LONG)
+  using GO = long; // may be 64 or 32 bits
+#elif defined(HAVE_TPETRA_INST_INT_UNSIGNED_LONG)
+  using GO = unsigned long; // may be 64 or 32 bits
+#else
+  using GO = Tpetra::Map<>::global_ordinal_type;
+#endif
```
rppawlo: Does this mean that the Tpetra default favors int over long long? I would have expected long long.

packages/tpetra/core/test/ImportExport/Issue_3968.cpp
```diff
@@ -0,0 +1,206 @@
...
+    std::vector<GO> gblInds (lclNumInds);
+    std::iota (std::begin (gblInds), std::end (gblInds), GO (0));
+    const GO indexBase = 0;
+    return Teuchos::rcp (new map_type (flag, gblInds.data (), lclNumInds, indexBase, comm));
+  }
+
+  // "Base value" should be some value, not default, not negative (in
+  // case GO is unsigned, which is not default but possible).  Pick
+  // baseValue() > comm->getSize().
+  constexpr GO baseValue () {
```
rppawlo: Maybe add:
TEUCHOS_ASSERT(baseValue > comm->getSize());

mhoemmen: True -- no need for `constexpr` here.

packages/tpetra/core/test/ImportExport/Issue_3968.cpp
mhoemmen: @rppawlo Fun fact: Default GlobalOrdinal type is `int`, even though `long long` is enabled by default.  See https://github.com/trilinos/Trilinos/issues/4029 .

mhoemmen:
Thanks @rppawlo ! :-D

mhoemmen:
Thanks @rppawlo ! :D

PR #4026, opened 2018-12-11
Panzer: use subviews for Intrepid2 methods

rppawlo:
This PR converts all calls to Intrepid2 to use subviews. The workset concept is used to control memory usage. The last workset may not be full. That means the actual number of cells is not always equal to workset size. Intrepid2 always uses the extent (workset size) for their loops leading to extra operations on dummy cells. In the past, we just iterated on the full extent, but some applications now enable FPE trapping by default. Switching to subviews fixes the extent correctly.


rppawlo:
@trilinos/framework - I just got the following error from the autotester. Any suggestions?

```
ERROR: Unable to find matching environment for job: Trilinos_pullrequest_gcc_7.2.0

Error code was: 42

Build step 'Execute shell' marked build as failure

Finished: FAILURE
```

PR #4040, opened 2018-12-12
Set CMAKE_SKIP_INSTALL_RPATH=ON for static builds on 'mutrino' (TRIL-241)

bartlettroscoe:
CC: @bathmatt, @jmgate, @krcb 

This is needed to get around an install error on 'mutrino' (see [TRIL-241](https://sems-atlassian-son.sandia.gov/jira/browse/TRIL-241)).

NOTE: This needed a FORCE set for the cache var since CMake seems to set the
default as 'NO'!


bartlettroscoe:
@trilinos/framework,

I added the `AT: RETEST` label 3 hours ago but no testing has started yet.  Is the PR tester just really backed up or is something else wrong here?

prwolfe:
The autotester instance for this job is currently busy running PR# 4020. I expect this one will pick up when that one completes (looks like it is about finished with it's build).  We did have a lot of jobs launch after CEE replaced the SSL certificate for Jenkins this morning.

Issue #4050, opened 2018-12-13
Panzer_IntegrationValues2 build failures on ATDM cuda 9.2 builds


fryeguy52:
CC: @trilinos/panzer, @mperego (Trilinos Discretizations Product Lead), @bartlettroscoe, @fryeguy52 
 
## Next Action Status

PR #4051 with fix merged to 'develop' on 12/13/2018.  Next: Watch for cleaned-up builds starting 12/14/2018 ...
 
## Description
 
As shown in [this query](https://testing.sandia.gov/cdash/index.php?project=Trilinos&date=2018-12-13&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=6&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-&field2=buildname&compare2=62&value2=Trilinos-atdm-white-ride-cuda-9.2-release-debug-pt&field3=subprojects&compare3=93&value3=Panzer&field4=builderrors&compare4=43&value4=0&field5=buildstarttime&compare5=83&value5=2018-12-12&field6=buildstarttime&compare6=84&value6=2018-12-14) there is a build failure while building `packages/panzer/disc-fe/src/CMakeFiles/panzer-disc-fe.dir/Panzer_IntegrationValues2.cpp.o` in the following builds:

* Trilinos-atdm-waterman-cuda-9.2-debug
* Trilinos-atdm-waterman-cuda-9.2-opt
* Trilinos-atdm-waterman-cuda-9.2-release-debug
* Trilinos-atdm-white-ride-cuda-9.2-debug
* Trilinos-atdm-white-ride-cuda-9.2-opt
* Trilinos-atdm-white-ride-cuda-9.2-release-debug

here is part of the output from the failures:
```
/home/jenkins/white/workspace/Trilinos-atdm-white-ride-cuda-9.2-release-debug/SRC_AND_BUILD/Trilinos/packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp(1140): error: no instance of function template "std::make_pair" matches the argument list
            argument types are: (int, Kokkos::Cuda::size_type)
          detected during instantiation of "void panzer::IntegrationValues2<Scalar>::getCubatureCV(const PHX::MDField<Scalar, panzer::Cell, panzer::BASIS, panzer::Dim, void, void, void, void, void> &, int) [with Scalar=double]" 
(275): here
```

## Current Status on CDash

[...]

PR #4051, opened 2018-12-13
Panzer: cuda fix

rppawlo:
Addresses #4050.

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
+++ b/packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
@@ -1121,6 +1121,7 @@ getCubatureCV(const PHX::MDField<Scalar,Cell,NODE,Dim>& in_node_coordinates,
   }
 
   size_type num_cells = in_num_cells < 0 ? in_node_coordinates.extent(0) : (size_type) in_num_cells;
+  std::pair<int,int> cell_range(0,num_cells);
   {
     size_type num_nodes = in_node_coordinates.extent(1);
     size_type num_dims = in_node_coordinates.extent(2);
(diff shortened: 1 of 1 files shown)


rppawlo:
@crtrott - the changes here are required for cuda, but the original code built and passed fine for other node types. It seems that cuda requires subview index pairs to be int.  I'm a little surprised by this restriction. Is this what you intended or is something using the wrong type underneath?

packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
```diff
@@ -1121,6 +1121,7 @@ getCubatureCV(const PHX::MDField<Scalar,Cell,NODE,Dim>& in_node_coordinates,
   }
 
   size_type num_cells = in_num_cells < 0 ? in_node_coordinates.extent(0) : (size_type) in_num_cells;
+  std::pair<int,int> cell_range(0,num_cells);
```
mperego: Wouldn't it be better to use Kokkos::pair instead?

rppawlo: > Wouldn't it be better to use Kokkos::pair instead?

Not according to the kokkos wiki subview page:

In the above example and those that follow in this chapter, we assume that Kokkos::View, Kokkos::subview, Kokkos::ALL, std::make_pair, and std::pair have been imported into the working C++ namespace.

rppawlo: I should comment that all subviews are being created on host, outside of kokkos kernels. Maybe you need the kokkos version of pair inside device kernels?

mperego: @rppawlo I see, that makes sense, thanks. I guess on device it would be safer to use Kokkos::pair.

Issue #3585, opened 2018-10-10
Test Anasazi_Epetra_OrthoManagerGenTester_0_MPI_4 appears to be randomly failing in many builds including CI, PR, and ATDM builds


bartlettroscoe:
CC: @trilinos/framework, @trilinos/anasazi, @srajama1 (Trilinos Linear Solver Product Area Lead)

## Next Action Status

PR #4052 merged to 'develop' on 12/18/2018 but still failing after that.  Next: Try to fix again?

## Description

It would seem that the test `Anasazi_Epetra_OrthoManagerGenTester_0_MPI_4` is very occasionally randomly failing in various builds.  As shown in [this query](https://testing.sandia.gov/cdash-dev-view/queryTests.php?project=Trilinos&date=2018-10-09&filtercount=4&showfilters=1&filtercombine=and&field1=testname&compare1=61&value1=Anasazi_Epetra_OrthoManagerGenTester_0_MPI_4&field2=status&compare2=61&value2=failed&field3=details&compare3=64&value3=timeout&field4=buildstarttime&compare4=83&value4=2018-07-01), this test failed 10 times since 7/1/2018 in the builds:

* `Linux-GCC-4.8.4-MPI_RELEASE_DEBUG_SHARED_PT_OPENMP_CI` (post-push CI build): 1 time (today)
* `PR-XXXX-test-Trilinos_pullrequest_gcc_4.9.3-YYYY` (standard PR build): 4 times
* `PR-XXXX-test-Trilinos_pullrequest_gcc_4.8.4-YYYY` (standard PR build): 1 time
* `Trilinos-atdm-chama-intel-debug-openmp` (standard ATDM build): 1 time
* `Trilinos-atdm-rhel6-gnu-opt-openmp` (standard ATDM build): 2 times
* `Trilinos-atdm-waterman-cuda-9.2-debug` (standard ATDM build): 1 time

In each of these 10 failures in the last 3 months, such as the CI failure today shown [here](https://testing.sandia.gov/cdash-dev-view/testDetails.php?test=56264374&build=4031303), it shows failures like:

```
projectAndNormalizeGen() returned rank 5
   || <S,S> - I || after  : 2.65912e-11
  1|| S_in - X1*C1 - X2*C2 - S_out*B || : 1.70776e-09
         vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv         tolerance exceeded! test failed!
```

[...]

PR #4052, opened 2018-12-13
Reducing subspace sizes due to biorthogonal basis construction for #3585

hkthorn:
@trilinos/anasazi 

## Description

This test, unlike the two other orthogonalization tests, generates a biorthogonal
basis.  When all the dimensions are summed up, the orthogonal constraints in this
test are twice the size of the other two tests.  Using random vectors, it would be
more likely that a degenerate basis could be constructed by this test.  Since the
objective of the test to make sure that the orthogonalization methods in Anasazi can
construct a biorthogonal basis, the size of the basis is not as much the point.
So, the dimension of each component of this test have been halved because of the
biorthogonality.  This seems to eliminate random failures that were observed on the
CEE, and are the subject of the bug #3585 .

## Motivation and Context

This should address the random failures in #3585.

## How Has This Been Tested?

On CEE, where the test failure was more often observed and could be studied.

## Checklist

### DIFF
--- a/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
+++ b/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
@@ -107,9 +107,9 @@ int main(int argc, char *argv[])
   bool debug = false;
   bool useMass = true;
   int dim = 100;
-  int sizeS  = 5;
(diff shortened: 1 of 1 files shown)


mhoemmen:
Thanks @hkthorn ! :-D

PR #4058, opened 2018-12-14
Panzer: fix clang warnings

rppawlo:
fixing clang warnings requested by empire

### DIFF
--- a/packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.hpp
+++ b/packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.hpp
@@ -46,7 +46,7 @@
 namespace panzer
 {
 template <typename LO, typename GO>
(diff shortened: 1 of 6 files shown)


packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.hpp
```diff
@@ -46,7 +46,7 @@
 namespace panzer
 {
 template <typename LO, typename GO>
-class LocalMeshInfo;
+struct LocalMeshInfo;
```
bartlettroscoe: Is the actually type a 'struct' and not a 'class'?

rppawlo: yes - clang picked up on the fact that the object was a "struct" but some files had forward declarations using "class". I just cleaned up all the forward declares.

PR #4070, opened 2018-12-17
Tpetra: Import/Export operations for StaticProfile CrsGraph

tjfulle:
@trilinos/tpetra 

PR enables import/export operations for CrsGraph with StaticProfile


packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -1195,6 +1206,19 @@ namespace Tpetra {
                     const Teuchos::ArrayView<const LocalOrdinal>& permuteToLIDs,
                     const Teuchos::ArrayView<const LocalOrdinal>& permuteFromLIDs) override;
 
+    void
+    resizeForIncomingDataImpl (const Kokkos::UnorderedMap<LocalOrdinal, size_t>& padding);
+
+    void
+    resizeForIncomingData (const RowGraph<LocalOrdinal, GlobalOrdinal, Node>& source,
```
mhoemmen: These two overloads of `resizeForIncomingData` look like they have very different purposes.  Should they perhaps have different names, e.g., `resizeForCopyAndPermute` and `resizeForUnpackAndCombine`?

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -1315,7 +1347,8 @@ namespace Tpetra {
     ///   only, and should never be called by user code.
     void
     setAllIndices (const typename local_graph_type::row_map_type& rowPointers,
-                   const typename local_graph_type::entries_type::non_const_type& columnIndices);
+                   const typename local_graph_type::entries_type::non_const_type& columnIndices,
+                   const bool overwrite=false);
```
mhoemmen: If we let users set `overwrite` and thus bypass the check, then it seems like we could just get rid of the check and keep the interface the same.  It's not obvious from the interface that the check exists.  Most "users" are actually MueLu developers, so they might not mind just getting rid of the check.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -1195,6 +1206,19 @@ namespace Tpetra {
                     const Teuchos::ArrayView<const LocalOrdinal>& permuteToLIDs,
                     const Teuchos::ArrayView<const LocalOrdinal>& permuteFromLIDs) override;
 
+    void
+    resizeForIncomingDataImpl (const Kokkos::UnorderedMap<LocalOrdinal, size_t>& padding);
```
mhoemmen: This declaration uses the default `Kokkos::Device` for `UnorderedMap`.  I'm guessing that you would instead want to use the same Device that CrsGraph is using.  If so, you should specify the type as `Kokkos::UnorderedMap<LocalOrdinal, size_t, device_type>`.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5647,6 +5657,122 @@ namespace Tpetra {
     }
   }
 
+  template <class LocalOrdinal, class GlobalOrdinal, class Node>
+  void
+  CrsGraph<LocalOrdinal, GlobalOrdinal, Node>::
+  resizeForIncomingDataImpl(const Kokkos::UnorderedMap<LocalOrdinal, size_t>& padding)
```
mhoemmen: See note above.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5647,6 +5657,122 @@ namespace Tpetra {
...
+
+    // Making copies here because k_rowPtrs_ has a const type. Otherwise, we
+    // would use it directly.
+    indices_type indices("indices", this->k_gblInds1D_.extent(0));
+    Kokkos::deep_copy(indices, this->k_gblInds1D_);
+
+    row_ptrs_type row_ptrs_beg("row_ptrs_beg", this->k_rowPtrs_.extent(0));
+    Kokkos::deep_copy(row_ptrs_beg, this->k_rowPtrs_);
+
+    const auto N = (row_ptrs_beg.extent(0) == 0 ? 0 : row_ptrs_beg.extent(0) - 1);
```
mhoemmen: The two branches of the ternary operator may have different types here.  I would recommend explicitly casting each to e.g., `size_t`.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5647,6 +5657,122 @@ namespace Tpetra {
...
+      refill_num_row_entries = true;
+      auto num_row_entries = this->k_numRowEntries_;
+      Kokkos::parallel_for("Fill end row pointers", range_policy(0, N),
+        KOKKOS_LAMBDA(const size_t i){
+          row_ptrs_end(i) = row_ptrs_beg(i) + num_row_entries(i);
+        }
+      );
+
+    } else {
+      // mfh If packed storage, don't need row_ptrs_end to be separate allocation;
```
mhoemmen: Are we saving the note in the comment for a later optimization?

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5647,6 +5657,122 @@ namespace Tpetra {
...
+    using GO = GlobalOrdinal;
+
+    // Resize row pointers and indices to accommodate incoming data
+    const map_type& src_row_map = *(source.getRowMap());
+    using padding_type = Kokkos::UnorderedMap<LocalOrdinal, size_t>;
+    padding_type padding(numSameIDs+permuteFromLIDs.size());
+    for (LO tgtid=0; tgtid<static_cast<LO>(numSameIDs); ++tgtid) {
+      const GO srcgid = src_row_map.getGlobalElement(tgtid);
+      auto how_much_padding = source.getNumEntriesInGlobalRow(srcgid);
+      padding.insert(tgtid, how_much_padding);
```
mhoemmen: 1. You should fence before (and possibly also after) this host code that accesses the (UVM) Views in `padding`.

2. `Kokkos::UnorderedMap` permits `insert` to fail, so you should check the return value and try again if it does fail.  (It's not likely outside of a parallel loop, but the interface permits it.)

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5687,15 +5813,36 @@ namespace Tpetra {
...
+                     Distributor& /* distor */) const
   {
     typedef LocalOrdinal LO;
     typedef GlobalOrdinal GO;
     typedef typename Kokkos::View<size_t*,
       device_type>::HostMirror::execution_space host_execution_space;
     typedef typename device_type::execution_space device_execution_space;
     const char tfecfFuncName[] = "pack: ";
-    constexpr bool debug = false;
+    const bool debug = ::Tpetra::Details::Behavior::debug("CrsGraph::pack");
```
mhoemmen: Thanks for changing this :-)

packages/tpetra/core/src/Tpetra_Details_Behavior.cpp
```diff
@@ -15,6 +15,7 @@ namespace Details {
 
 namespace BehaviorDetails {
 std::map<std::string, std::map<std::string, bool> > namedVariableMap_;
+bool verboseDisabled_ = false;
```
mhoemmen: Did you mean to include these changes here?  They look like they belong to some earlier PR.

packages/tpetra/core/src/Tpetra_Details_resizeRowPtrs.hpp
```diff
@@ -0,0 +1,192 @@
...
+///   of Tpetra.  The file itself or its contents may disappear or
+///   change at any time.
+
+namespace Tpetra {
+namespace Details {
+
+namespace pad_csr_details {
+
+template<class ViewType>
+ViewType empty_view(const std::string& name, const size_t& size) {
```
mhoemmen: It's not an "empty" View; it's an uninitialized View.

packages/tpetra/core/src/Tpetra_Details_resizeRowPtrs.hpp
```diff
@@ -0,0 +1,192 @@
...
+  ViewType v(Kokkos::view_alloc(name, Kokkos::WithoutInitializing), size);
```
mhoemmen: `return ViewType (Kokkos::view_alloc (name, Kokkos::WithoutInitializing), size));`

packages/tpetra/core/src/Tpetra_Details_resizeRowPtrs.hpp
```diff
@@ -0,0 +1,192 @@
...
+// Determine if row_ptr and indices arrays need to be resized to accommodate
+// new entries
+template<class RowPtr, class Indices, class Padding>
+void
+pad_csr_arrays(RowPtr& row_ptr_beg, RowPtr& row_ptr_end, Indices& indices, const Padding& padding)
+{
+  using range = Kokkos::pair<typename RowPtr::value_type, typename RowPtr::value_type>;
+
+  // Determine if the indices array is large enough
+  auto num_row = row_ptr_beg.size() - 1;
```
mhoemmen: Could `row_ptr_beg` ever be length 0, e.g., because there are no entries in the matrix/graph on the calling process?

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
```diff
@@ -136,78 +137,69 @@ unpackRow(typename Kokkos::View<GO*,Device,Kokkos::MemoryUnmanaged>& gids_out,
...
+  using LO = LocalOrdinal;
+  using GO = typename IndicesView::value_type;
+  using packet_type = Packet;
+  using row_ptrs_type = RowView;
+  using indices_type = IndicesView;
+  using buffer_device_type = BufferDevice;
 
-  typedef typename local_map_type::local_ordinal_type LO;
-  typedef typename local_map_type::global_ordinal_type GO;
   // Kokkos::parallel_reduce fails to compile if named device_type and typedef
```
mhoemmen: Is this comment still relevant?  I'm guessing not.  If not, now would be a good chance to remove it :-)

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
```diff
@@ -289,14 +279,15 @@ class UnpackAndCombineFunctor {
...
     }
 
-    // Column indices come in as global indices, in case the
-    // source object's column Map differs from the target object's
-    // (this's) column Map, and must be converted local index values
+    auto import_lid = import_lids(i);
     for (size_t k = 0; k < num_ent; ++k) {
-      lids_out(k) = local_col_map.getLocalElement(gids_out(k));
+      indices(row_ptrs_end(import_lid)) = gids_out(k);
+      // this is OK; don't need atomic, since LIDs to pack don't have repeats.
```
mhoemmen: Good catch! :-D

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
```diff
@@ -911,42 +878,89 @@ unpackCrsGraphAndCombine(
...
 
-  auto local_graph = sourceGraph.getLocalGraph();
-  auto local_col_map = sourceGraph.getColMap()->getLocalMap();
+  // mfh Could fix the use of protected data by adding a method
+  // to CrsGraph that returns the "local, globally indexed graph".
+  // However, this function needs to change those arrays anyway,
+  // so we might as well make this function a friend of CrsGraph.
+
+  //FIXME (tjf 11 Oct 2018) need to use non-protected data
+  indices_type indices("indices", graph.k_gblInds1D_.extent(0));
```
mhoemmen: Wait, we should just fix the code or make this a `friend` function if we need to get to these data.  We shouldn't have to copy them here.  This function is an implementation detail of CrsGraph, so we should feel free to make it a friend of CrsGraph.  We already do that for other functions relating to pack and unpack.

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
```diff
@@ -911,42 +878,89 @@ unpackCrsGraphAndCombine(
...
+  Kokkos::deep_copy(indices, graph.k_gblInds1D_);
+
+  row_ptrs_type row_ptrs_beg("row_ptrs_beg", graph.k_rowPtrs_.extent(0));
+  Kokkos::deep_copy(row_ptrs_beg, graph.k_rowPtrs_);
+
+  const auto N = (row_ptrs_beg.extent(0) == 0 ? 0 : row_ptrs_beg.extent(0) - 1); // TODO fix auto
```
mhoemmen: Yes; please see above comment.

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
```diff
@@ -911,42 +878,89 @@ unpackCrsGraphAndCombine(
...
+    refill_num_row_entries = true;
+    auto num_row_entries = graph.k_numRowEntries_;
+    Kokkos::parallel_for("Fill end row pointers", range_policy(0, N),
+        KOKKOS_LAMBDA(const size_t i){
+          row_ptrs_end(i) = row_ptrs_beg(i) + num_row_entries(i);
+      });
+
+  } else {
+    // mfh If packed storage, don't need row_ptrs_end to be separate allocation;
+    // could just have it alias row_ptrs_beg+1.
```
mhoemmen: Are we leaving this comment for later?

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5647,6 +5657,122 @@ namespace Tpetra {
...
+    for (LO tgtid=0; tgtid<static_cast<LO>(numSameIDs); ++tgtid) {
+      const GO srcgid = src_row_map.getGlobalElement(tgtid);
+      auto how_much_padding = source.getNumEntriesInGlobalRow(srcgid);
+      padding.insert(tgtid, how_much_padding);
+    }
+    for (LO i=0; i<permuteToLIDs.size(); ++i) {
+      const LO tgtid = permuteToLIDs[i];
+      const GO srcgid = src_row_map.getGlobalElement(permuteFromLIDs[i]);
+      auto how_much_padding = source.getNumEntriesInGlobalRow(srcgid);
+      padding.insert(tgtid, how_much_padding);
```
mhoemmen: Kokkos::UnorderedMap permits insert to fail, so you should check the return value and try again if it does fail. (It's not likely outside of a parallel loop, but the interface permits it.)

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5647,6 +5657,122 @@ namespace Tpetra {
...
+  CrsGraph<LocalOrdinal, GlobalOrdinal, Node>::
+  resizeForIncomingData (const Teuchos::ArrayView<const LocalOrdinal> &importLIDs,
+                         const Teuchos::ArrayView<size_t> &numPacketsPerLID)
+  {
+    // Creating padding for each new incoming index
+    using padding_type = Kokkos::UnorderedMap<LocalOrdinal, size_t>;
+    padding_type padding(importLIDs.size());
+    auto numEnt = static_cast<size_t>(importLIDs.size());
+    for (size_t i=0; i<numEnt; i++) {
+      padding.insert(importLIDs[i], numPacketsPerLID[i]);
```

mhoemmen:
Thanks @tjfulle ! :-D  There's one bit of code relating to disabling verbose output, that looks like it got pulled in from a different PR.  Is that what happened, or is that OK?

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
tjfulle: Perhaps?  Or I should just take the overwrite parameter out.  It was from an earlier incarnation where I used setAllIndices, I changed in favor of writing the more general resizeForIncomingData.  I’ll have to double check, but I don’t believe it is used anymore

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
tjfulle: Ok, if that is preferred.  They both just generate the padding map for different inputs.  I can change their names.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
tjfulle: 👍

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: Which note?

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: Doh!  Good catch!!

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: That was the intent.  I think some of this will change with the new interface, but FECrsGraph needs this functionality now :)

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: 1. 👍
2. 👍

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: On my list :)

packages/tpetra/core/src/Tpetra_Details_Behavior.cpp
tjfulle: No, this was the change I put in to test the StaticProfile test (where I wanted verbose output for only a part of the test).

packages/tpetra/core/src/Tpetra_Details_resizeRowPtrs.hpp
tjfulle: 👍

packages/tpetra/core/src/Tpetra_Details_resizeRowPtrs.hpp
tjfulle: Yes, in fact I had a test that did just that (inadvertently) and this function hung.  I fixed the test - I did not fix the function, yet

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
tjfulle: 👍

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
tjfulle: Yes, this function will change once the new interface is implemented

tjfulle:
@mhoemmen, the code to programmatically turn on and off verbose output was put in while I worked on this PR, I could pull it out as standalone if you prefer

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: -> not in the _decl file

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
tjfulle: actually, their purposes are identical - compute the necessary padding and call resizeForIncomingDataImpl.  The different inputs are used to compute the padding in different ways.

packages/tpetra/core/src/Tpetra_Details_Behavior.cpp
tjfulle: To be clear.  Yes, I meant to include these changes.  It was in this work that I wrote this stuff.

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
tjfulle: this function is already a friend of CrsGraph, just need to remove the comment

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
mhoemmen: It's so tempting to merge `copyAndPermute` and `unpackAndCombine` in CrsGraph and CrsMatrix....

I would prefer a rename if you happen to be changing other things -- the names should make it more clear which one to call when.  Thanks!

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
mhoemmen: That would be even better; thanks!

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
mhoemmen: Cool; thanks!

mhoemmen:
@tjfulle wrote:
> the code to programmatically turn on and off verbose output was put in while I worked on this PR, I could pull it out as standalone if you prefer

It's OK, I was just making sure something didn't get mixed up.  Thanks!

william76:
@tjfulle looks like it passed its test... that one error you had was probably just some kind of noise (?) .  I think occasionally there's some odd one-off error that pops up which just goes away on its own... blame them on cosmic rays or something /shrug.  Anyways, looks like that error just went away on its own. :)

tjfulle:
@mhoemmen - ping

I know you're busy and went to @csiefer2 first, but didn't catch him in time before he left for the holiday.  If you have time to look over the responses to your feedback that would be awesome.  If not, I'll get csiefer2 to do it after the shutdown :)

packages/tpetra/core/src/Tpetra_Details_unpackCrsGraphAndCombine_def.hpp
```diff
@@ -970,15 +1006,14 @@ unpackCrsGraphAndCombineNew(
...
-  auto local_col_map = sourceGraph.getColMap()->getLocalMap();
-  typedef decltype(local_col_map) local_map_type;
-
   // Now do the actual unpack!
-  UnpackAndCombineCrsGraphImpl::unpackAndCombine<
-    packet_type,local_graph_type,local_map_type,buffer_device_type>(
-      local_graph, local_col_map, imports_d, num_packets_per_lid_d,
-      import_lids_d, combineMode, false, atomic);
+  // TJF: Should be grabbed from the Graph
+  indices_type indices;
```
mhoemmen: Is there something missing here or is it OK?

tjfulle: i don't think so.  probably a place holder comment from the first iterations.  I'll kill the comment on the next round of work that will need to be done with the new interface.

tjfulle:
thanks @mhoemmen!

PR #4072, opened 2018-12-17
Epetra: add method to get pointer to Distributor in Import and Export classes

jhux2:
Auto-PR for SHA ca1e86e

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/4072)

### DIFF
--- a/packages/epetra/src/Epetra_Export.h
+++ b/packages/epetra/src/Epetra_Export.h
@@ -279,6 +279,8 @@ and entries 5 and 6 on PEs 1 and 2.
 
   Epetra_Distributor & Distributor() const {return(*Distor_);};
 
+  const Epetra_Distributor * DistributorPtr() const {return(Distor_);}
+
   //! @name Print object to an output stream
   //@{
   virtual void Print(std::ostream & os) const;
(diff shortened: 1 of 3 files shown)


jhux2:
@trilinos/framework This PR keeps getting retested.  Based on the output, it's not clear if there's a build failure or something else going on.

jwillenbring:
@jhux2 @trilinos/framework 

It appears that the first time the testing failed it was due to the PR being modified while it was being tested.

Subsequent failures have been due to heavy load on the PR system. I took action to reduce this load this morning. I will look into this further. It is possible we have orphaned jobs or some other issue here.

jhux2:
@jwillenbring Thank you.

jwillenbring:
@jhux2 @trilinos/framework 

OK, so despite measures to reduce the load, the PR load is still too heavy on the system. I spoke with @ZUUL42 and he is going to modify the set of jobs to turn off the GCC 4.9.3 MPI test now that the GCC 7.2 test has been functional for a while. This will reduce the number of jobs per run from 5 back down to 4 and should get us to a place where we have enough executors to prevent the timeouts. Further, @prwolfe is investigating more resources so we can run more tests simultaneously.

jhux2:
@trilinos/framework The autotester is still failing in a way that's not obvious to me.

PR #4083, opened 2018-12-19
Tpetra: Fix #4080

### LINKED ISSUE #4080: Stratimikos: Amesos2LinearOpWithSolveFactory adapter may cause link errors if Teuchos_ENABLE_FLOAT=ON but Tpetra_INST_FLOAT=OFF
@trilinos/stratimikos @trilinos/thyra @rrdrake 

Stratimikos [uses the macro](https://github.com/trilinos/Trilinos/blob/6f4bf237f689a547430c4183c41459d11946a60a/packages/stratimikos/adapters/amesos2/src/Thyra_Amesos2LinearOpWithSolve.cpp#L53) `TEUCHOS_CLASS_TEMPLATE_INSTANT_REAL_SCALAR_TYPES` to do explicit template instantiation (ETI) for its Amesos2 adapter, `Thyra::Amesos2LinearOpWithSolveFactory`.  [This macro depends on the](https://github.com/trilinos/Trilinos/blob/6f4bf237f689a547430c4183c41459d11946a60a/packages/teuchos/core/src/Teuchos_ExplicitInstantiationHelpers.hpp#L66) `HAVE_TEUCHOS_FLOAT` macro, which in turn depends on the `Teuchos_ENABLE_FLOAT` CMake option.  However, whether it's valid to use Tpetra objects with `Scalar=float` depends on a different CMake option, `Tpetra_INST_FLOAT`.

`Tpetra_INST_FLOAT` and `Teuchos_ENABLE_FLOAT` both take their default values from the common CMake option `Trilinos_ENABLE_FLOAT`.  However, #4059 shows that it's possible to override those defaults and set `Teuchos_ENABLE_FLOAT=ON` but `Tpetra_INST_FLOAT=OFF`.  This will result in link errors in packages that use `TEUCHOS_CLASS_TEMPLATE_INSTANT_REAL_SCALAR_TYPES` to instantiate code that uses Tpetra objects.

## Possible Solution

At configure time, forbid setting `Teuchos_ENABLE_FLOAT=ON` but `Tpetra_INST_FLOAT=OFF`.  This is most easily done in Tpetra, since Tpetra has a required dependence on Teuchos.

One could fix this in other ways, but I prefer solutions that avoid combinatorial explosion of build options.

## Related Issues

* Follows #4059 
* Related to #4059, #2009

mhoemmen:
@trilinos/tpetra

Forbid inconsistent CMake options about `Scalar=float`, `std::complex<double>`, and `std::complex<float>`, that could cause build errors downstream (in Thyra or Stratimikos).

## Related Issues

* Closes #4080 
* Related to #4059

### DIFF
--- a/packages/tpetra/CMakeLists.txt
+++ b/packages/tpetra/CMakeLists.txt
@@ -825,6 +825,7 @@ GLOBAL_SET(HAVE_TPETRA_FLOAT ${Tpetra_INST_FLOAT})
 # linking when ETI is ON, with downstream packages like Thyra, which
 # get used in surprising places, e.g., in Ifpack2).
 IF (Tpetra_INST_FLOAT)
+
   IF (NOT DEFINED Teuchos_ENABLE_FLOAT)
     MESSAGE(FATAL_ERROR "Tpetra: Tpetra_INST_FLOAT is ON (meaning that you want to instantiate and/or test Tpetra with Scalar = float), but Teuchos_ENABLE_FLOAT is not defined.  Tpetra needs Teuchos to support float.  Please set Teuchos_ENABLE_FLOAT:BOOL=ON, reconfigure, and rebuild.")
   ELSEIF (NOT Teuchos_ENABLE_FLOAT)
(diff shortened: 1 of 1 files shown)


mhoemmen:
Thanks @bartlettroscoe ! :D

Issue #4065, opened 2018-12-17
Framework: Remove unused parameters from Pull Request Jenkins Jobs


william76:
@trilinos/framework 

### Issue
Last week I ran into an issue where the various Jenkins jobs that are run by the autotester for Pull Request testing had different stets of parameters.  This was confusing so I made a pass through to synchronize all of them with the union of all the parameters from the different jobs. So now we have a bunch of parameters that aren't used in these jenkins jobs.

### Backstory
As discussed in our standup, the backstory on the extra parameters is that we initially pulled these jobs from the `parameterized build` and thought we might use them but then we decided against it because using these parameters might be difficult for developers to use to try and replicate builds that failed. 

### Done Criteria
- [x] **Trilinos_pullrequest_gcc_4.8.4**: Remove parameters: `COMPILER_MODULE`, `MPI_MODULE`, `JENKINS_BUILD_TYPE`, `JENKINS_COMM_TYPE`, `JENKINS_JOB_TYPE`, `JENKINS_DO_COMPLEX`
- [x] **Trilinos_pullrequest_gcc_4.9.3**: Remove parameters: `COMPILER_MODULE`, `MPI_MODULE`, `JENKINS_BUILD_TYPE`, `JENKINS_COMM_TYPE`, `JENKINS_JOB_TYPE`, `JENKINS_DO_COMPLEX`
- [x] **Trilinos_pullrequest_gcc_4.9.3_SERIAL**: Remove parameters: `COMPILER_MODULE`, `MPI_MODULE`, `JENKINS_BUILD_TYPE`, `JENKINS_COMM_TYPE`, `JENKINS_JOB_TYPE`, `JENKINS_DO_COMPLEX`
- [x] **Trilinos_pullrequest_gcc_7.3.0**: Remove parameters: `COMPILER_MODULE`, `MPI_MODULE`, `JENKINS_BUILD_TYPE`, `JENKINS_COMM_TYPE`, `JENKINS_JOB_TYPE`, `JENKINS_DO_COMPLEX`
- [x] **Trilinos_pullrequest_intel_17.0.1**: Remove parameters: `COMPILER_MODULE`, `MPI_MODULE`, `JENKINS_BUILD_TYPE`, `JENKINS_COMM_TYPE`, `JENKINS_JOB_TYPE`, `JENKINS_DO_COMPLEX`
- [x] Verify that all Jenkins pull request jobs are consistent in their parameters.
[...]

PR #4089, opened 2018-12-19
Tpetra: Clean up CrsGraph::Swap testing

william76:
@trilinos/tpetra

There were some `#if 0` blocks in the test which needed to be changed to use `Tpetra::Details::Behavior::verbose()` for consistency.

This PR is convenient as well for testing changes to PR jobs on Jenkins for @trilinos/framework for #4065 =) .

### DIFF
--- a/packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
+++ b/packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
@@ -125,28 +125,25 @@ TEUCHOS_STATIC_SETUP()
 // gbl_num_columns: Max # of columns in the matrix-representation of the graph.
 //                  This should be >= the highest value of v from all edges (u,v) in edges.
 //                  Note: u and v are 0-indexed, so if the highest v is 11, then this should be 12.
-template<class LO, class GO, class Node, class comm_type>
-Teuchos::RCP<Tpetra::CrsGraph<LO, GO, Node>>
-generate_crsgraph(Teuchos::RCP<comm_type>                      comm,
-                  const std::vector<std::pair<GO, GO>>&     gbl_edges,
-                  const std::vector<std::pair<GO, size_t>>& gbl_row_owners,
-                  const size_t                              gbl_num_columns,
-                  const bool                                do_fillComplete=true)
+template<class LO, class GO, class Node>
+Teuchos::RCP<Tpetra::CrsGraph<LO, GO, Node> >
+generate_crsgraph(Teuchos::RCP<Teuchos::Comm<int> >&      comm,
+                  const std::vector<std::pair<GO, GO> >&  gbl_edges,
+                  const std::vector<std::pair<GO, int> >& gbl_row_owners,
+                  const size_t                            gbl_num_columns,
+                  const bool                              do_fillComplete=true)
 {
     using Teuchos::Comm;
 
-    typedef Tpetra::CrsGraph<LO, GO, Node> graph_type;
-    typedef Tpetra::Map<LO, GO, Node>      map_type;
-
-    typedef typename std::map<GO, size_t> map_rows_type;      // row_id, num_entries
-
-
-    typedef typename std::vector<GO>        vec_go_type;
-    typedef typename std::map<GO, vec_go_type> map_row_to_cols_type;
+    using graph_type           = Tpetra::CrsGraph<LO,GO,Node>;      // Tpetra CrsGraph type
+    using map_type             = Tpetra::Map<LO,GO,Node>;           // Tpetra Map type
+    using map_rows_type        = std::map<GO,int>;                  // map rows to pid's
+    using vec_go_type          = std::vector<GO>;                   // vector of GlobalOrdinals
+    using map_row_to_cols_type = std::map<GO, vec_go_type>;         // Map rows to columns
 
     const bool verbose = Tpetra::Details::Behavior::verbose();
 
-    const size_t comm_rank = comm->getRank();
+    const int comm_rank = (size_t)comm->getRank();
 
     map_row_to_cols_type gbl_rows;
     for(auto& e: gbl_edges)
@@ -221,8 +218,8 @@ generate_crsgraph(Teuchos::RCP<comm_type>                      comm,
     for(auto& r: gbl_rows)
     {
         const GO     irow    = r.first;
-        const size_t row_pid = gbl_row2pid.find(irow)->second;
-        if(row_pid == comm_rank)
+        const int row_pid = gbl_row2pid.find(irow)->second;
+        if(comm_rank == row_pid)
         {
             num_ent_per_row[ idx++ ] = r.second.size();
         }
@@ -243,8 +240,8 @@ generate_crsgraph(Teuchos::RCP<comm_type>                      comm,
     for(auto& r: gbl_rows)
     {
         const GO     irow = r.first;
-        const size_t pid  = gbl_row2pid.find(irow)->second;
-        if(pid == comm_rank)
+        const int pid  = gbl_row2pid.find(irow)->second;
+        if(comm_rank == pid)
         {
             std::vector<GO> gbl_inds;
             for(auto& v: r.second)
@@ -293,18 +290,17 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(CrsGraph, Swap, LO, GO, Node)
     using Teuchos::outArg;
     using Teuchos::RCP;
 
-    typedef Teuchos::Comm<int>             comm_type;
-    typedef Tpetra::CrsGraph<LO, GO, Node> graph_type;
+    using graph_type      = Tpetra::CrsGraph<LO,GO,Node>;       // Tpetra CrsGraph type
+    using pair_edge_type  = std::pair<GO,GO>;                   // Edge typs, (u,v) using GlobalOrdinal type
+    using pair_owner_type = std::pair<GO,int>;                  // For row owners, pairs are (rowid, comm rank)
 
-    typedef typename std::pair<GO, GO>     pair_edge_type;
-    typedef typename std::pair<GO, size_t> pair_owner_type;
-
-    typedef typename std::vector<pair_edge_type>  vec_edges_type;
-    typedef typename std::vector<pair_owner_type> vec_owners_type;
+    using vec_edges_type  = std::vector<pair_edge_type>;        // For vectors of edges
+    using vec_owners_type = std::vector<pair_owner_type>;       // For vectors of owners
 
     bool verbose = Tpetra::Details::Behavior::verbose();
 
     auto initialComm = getDefaultComm();
+
     TEUCHOS_TEST_FOR_EXCEPTION(initialComm->getSize() < 2, std::runtime_error, "This test requires at least two processors.");
 
     // Set up a communicator that has exactly two processors in it for the actual test.
@@ -335,12 +331,12 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(CrsGraph, Swap, LO, GO, Node)
          }
 
         out << ">>> create graph_a" << std::endl;
-        RCP<graph_type> graph_a = generate_crsgraph<LO, GO, Node, comm_type>(comm, vec_edges, vec_owners, 12);
-        //graph_a->describe(out, Teuchos::VERB_DEFAULT);
+        RCP<graph_type> graph_a = generate_crsgraph<LO, GO, Node>(comm, vec_edges, vec_owners, 12);
+        if(verbose) graph_a->describe(out, Teuchos::VERB_DEFAULT);
 
         out << ">>> create graph_b" << std::endl;
-        RCP<graph_type> graph_b = generate_crsgraph<LO, GO, Node, comm_type>(comm, vec_edges, vec_owners, 12);
-        //graph_b->describe(out, Teuchos::VERB_DEFAULT);
+        RCP<graph_type> graph_b = generate_crsgraph<LO, GO, Node>(comm, vec_edges, vec_owners, 12);
+        if(verbose) graph_b->describe(out, Teuchos::VERB_DEFAULT);
 
         vec_edges.clear();
         vec_owners.clear();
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -160,20 +162,21 @@ generate_crsgraph(Teuchos::RCP<comm_t>                      comm,
...
 
     // Print out some debugging information on what's in the graph
-    #if 0
-    std::cout << "p=0 | gbl_num_rows: " << gbl_rows.size() << std::endl;
-    for(auto& p: gbl_rows)
+    if(verbose)
     {
-        std::cout << "p=0 | gbl_row     : " << p.first << " (" << p.second.size() << ") ";
-        for(auto& j: p.second)
+        std::cout << "p=0 | gbl_num_rows: " << gbl_rows.size() << std::endl;
```
mhoemmen: For future reference (not for this PR), when printing debugging messages nondeterministically over multiple MPI processes to cout or cerr, good practice is to assemble each line (including the endline character) from each process as a string.  This tends to make messages from different processes less mixed up.  For example:
```c++
std::ostringstream os;
os << "(Proc " << myRank << ") gbl_num_rows: " << gbl_rows.size () << std::endl;
std::cout << os.str ();
```

mhoemmen: At some point, it would make sense to turn that into a macro:
```c++
#define TPETRA_PRINTLN_LOCAL( out, myRank, stuff ) \
  do { \
    std::ostringstream tpl_out; \
    tpl_out << "(Proc " << myRank << ") " << stuff << std::endl; \
    out << tpl_out.str (); \
  } while (false);
```
Alas, C++ lacks the equivalent of Common Lisp's `GENSYM`, so I can't guarantee that `tpl_out` doesn't shadow any other names in scope.  This should still work, though.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -200,18 +202,18 @@ generate_crsgraph(Teuchos::RCP<comm_t>                      comm,
...
+                  << "global_ids.data(), " << lcl_num_rows << ", 0, comm)" << std::endl;
     }
-    #endif
 
     // Create the Row Map
-    #if 0
-    std::cout << "p=" << comm_rank << " | " << "row_map = map_t(" << gbl_num_rows << ", " << "global_ids.data(), "
-              << lcl_num_rows << ", 0, comm)" << std::endl;
-    #endif
     RCP<const map_t> row_map(new map_t(gbl_num_rows, global_ids.data(), lcl_num_rows, 0, comm));
```
mhoemmen: For future reference, it's nicer to follow Tpetra's existing convention when naming typedefs.  I consider `foo_t` to be reserved by system headers.  `foo_type` (snake case) is good for public typedefs.  Tpetra makes exceptions for internal (nonpublic) typedefs that it uses almost everywhere, like `LO`, `GO`, and `MV`.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -226,13 +228,15 @@ generate_crsgraph(Teuchos::RCP<comm_t>                      comm,
...
+        std::cout << "p=" << comm_rank << " | lcl_num_rows = " << lcl_num_rows << std::endl;
+        for(int i=0; i<lcl_num_rows; i++)
+        {
+            std::cout << "p=" << comm_rank << " | " << "num_ent_per_row[" << i << "] = " 
+                      << num_ent_per_row[i] << std::endl;
+        }
     }
-    #endif
 
     RCP<graph_t> output_graph(new graph_t(row_map, num_ent_per_row, Tpetra::StaticProfile));
```
mhoemmen: See above note on typedef naming conventions.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
william76: @mhoemmen I've never seen a style guide for Trilinos... does such a document exist that captures these conventions or is something like `foo_t` being reserved by system headers just a general C++ convention?  If it's the latter, do you have any links to something that lays these out? Where possible, I do prefer to avoid breaking the conventions ;)

william76:
@mhoemmen I updated the typedefs in the test file and pushed the change. :)

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
mhoemmen: There's no style guide, but if you read Tpetra classes and tests you'll get a general impression.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -125,55 +125,58 @@ TEUCHOS_STATIC_SETUP()
 // gbl_num_columns: Max # of columns in the matrix-representation of the graph.
 //                  This should be >= the highest value of v from all edges (u,v) in edges.
 //                  Note: u and v are 0-indexed, so if the highest v is 11, then this should be 12.
-template<class LO, class GO, class Node, class comm_t>
+template<class LO, class GO, class Node, class comm_type>
```
mhoemmen: 1. You don't need to template on `comm_type` here.  It's always `const Teuchos::Comm<int>`.
2. Typedefs are snake case, template parameters are Capitalized or CamelCase.  See `std::map` for examples: https://en.cppreference.com/w/cpp/container/map

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -125,55 +125,58 @@ TEUCHOS_STATIC_SETUP()
...
                   const std::vector<std::pair<GO, GO>>&     gbl_edges,
                   const std::vector<std::pair<GO, size_t>>& gbl_row_owners,
                   const size_t                              gbl_num_columns,
                   const bool                                do_fillComplete=true)
 {
     using Teuchos::Comm;
 
-    typedef Tpetra::CrsGraph<LO, GO, Node> graph_t;
-    typedef Tpetra::Map<LO, GO, Node>      map_t;
+    typedef Tpetra::CrsGraph<LO, GO, Node> graph_type;
```
mhoemmen: For future reference, prefer `using` syntax as the One Syntax for type aliases.  For example:
```c++
using graph_type = Tpetra::CrsGraph<LO, GO, Node>;
template<class GO>
using my_graph_type = Tpetra::CrsGraph<int, GO>;
```
We prefer `using` syntax because it makes type aliases read just like variable assignment.  "Define `new_name_type` to be Something."  I say "for future reference," because plenty of Tpetra code uses C++98 `typedef`.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -125,55 +125,58 @@ TEUCHOS_STATIC_SETUP()
...
 
 
-    typedef typename std::vector<GO>        vec_go_t;
-    typedef typename std::map<GO, vec_go_t> map_row_to_cols_t;
+    typedef typename std::vector<GO>        vec_go_type;
+    typedef typename std::map<GO, vec_go_type> map_row_to_cols_type;
+
+    const bool verbose = Tpetra::Details::Behavior::verbose();
 
     const size_t comm_rank = comm->getRank();
```
mhoemmen: MPI process ranks are always `int`.  This could cause a signed/unsigned conversion warning.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -330,27 +335,27 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(CrsGraph, Swap, LO, GO, Node)
...
         //graph_b->describe(out, Teuchos::VERB_DEFAULT);
 
         vec_edges.clear();
         vec_owners.clear();
 
-        vec_edges  = {pair_edge_t(0, 0), pair_edge_t(0, 11), pair_edge_t(1, 7), pair_edge_t(1, 8),
-                      pair_edge_t(3, 1), pair_edge_t(7, 5), pair_edge_t(10, 4) };
-        vec_owners = {pair_owner_t(0, 0), pair_owner_t(1, 0), pair_owner_t(3, 1), pair_owner_t(7, 1),
-                      pair_owner_t(10, 1)};
+        vec_edges  = {pair_edge_type(0, 0), pair_edge_type(0, 11), pair_edge_type(1, 7), pair_edge_type(1, 8),
```
mhoemmen: I like these typedefs btw :-D

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
william76: @mhoemmen I changed it up so that I can use `int` for `comm_rank`

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
william76: Updated to use the `using` syntax.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
william76: template on `comm_type` removed.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
william76: Is the `#define TPETRA_PRINTLN_LOCAL` a thing that already exists, or are you suggesting creating this macro in the file to use?  Either way is doable, but if there's already a macro or function defined that does this then I'm happy to use it :)

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -125,28 +125,25 @@ TEUCHOS_STATIC_SETUP()
...
+    using graph_type           = Tpetra::CrsGraph<LO,GO,Node>;      // Tpetra CrsGraph type
+    using map_type             = Tpetra::Map<LO,GO,Node>;           // Tpetra Map type
+    using map_rows_type        = std::map<GO,int>;                  // map rows to pid's
+    using vec_go_type          = std::vector<GO>;                   // vector of GlobalOrdinals
+    using map_row_to_cols_type = std::map<GO, vec_go_type>;         // Map rows to columns
 
     const bool verbose = Tpetra::Details::Behavior::verbose();
 
-    const size_t comm_rank = comm->getRank();
+    const int comm_rank = (size_t)comm->getRank();
```
mhoemmen: Don't cast to `size_t`.  It's not a `size_t` and it never was :-)

william76: gah, missed that when I was changing everything to int.  Will fix and push the update :p

william76: BTW, thanks for looking at this... I greatly appreciate your attention to detail.  Its so easy to miss a thing like this when the compiler doesn't give me any warnings on it locally and my eyes just don't want to see the miscast since that was something I tossed in going down the path of using `size_t` before I just fixed my use of `comm_rank` overall from `size_t` to `int`. :)

Also, I'm sure the autotester application is hating on me right now, I keep letting it kick off my jobs then it keeps having to kill the tests b/c I'm updating my PR. haha.  Normally that's a WIP thing, but I keep thinking 'oh, this is a minor fix' so it's totes good. :p Anyway, this one is the last tweak I think  <knock on wood>.  The build cluster for the tests isn't heavily loaded today so it's not a big deal so its good

william76:
ok, I think I got the little tweaks and style fixes sorted out... unless something is glaringly wrong, I don't think I'll update this PR again before it's merged on in.  

Requesting a review from someone on the  @trilinos/tpetra team :)

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
mhoemmen: @william76 I'm being a bit of a pain about "attention to detail" but it's so much easier to read the code and be sure that it's correct when it has a consistent style, so thanks :-D

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -125,55 +123,55 @@ TEUCHOS_STATIC_SETUP()
...
-template<class LO, class GO, class Node, class comm_t>
-Teuchos::RCP<Tpetra::CrsGraph<LO, GO, Node>>
-generate_crsgraph(Teuchos::RCP<comm_t>                      comm,
-                  const std::vector<std::pair<GO, GO>>&     gbl_edges,
-                  const std::vector<std::pair<GO, size_t>>& gbl_row_owners,
-                  const size_t                              gbl_num_columns,
-                  const bool                                do_fillComplete=true)
+template<class LO, class GO, class Node>
+Teuchos::RCP<Tpetra::CrsGraph<LO, GO, Node> >
+generate_crsgraph(Teuchos::RCP<Teuchos::Comm<int> >&      comm,
```
mhoemmen: Argh i forgot one thing: this should be `Teuchos::RCP<const Teuchos::Comm<int>>`.  But if it compiles and passes the test, don't worry about it for now.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
mhoemmen: @william76 I'm just suggesting this as a later thing to think about, after this PR is done :-D  Thanks though!

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
william76: @mhoemmen just to clarify, the only change here would be to make this a const, right?

That's easy enough to do and probably best to just do today before I forget about it over the break.  I can throw that in and shoot off a new PR for it.

william76: @mhoemmen I made this change and put it into PR #4120

PR #4097, opened 2018-12-20
Update kokkos integration testing script

ndellingwood:
Update the configure options used to enable complex double types


ndellingwood:
@mhoemmen This PR updates the kokkos integration testing script to enable complex types as you suggested in kokkos/kokkos#1951.

PR #4135, opened 2019-01-03
Modify logic in block Gmres solver to address bug #4128

hkthorn:
@trilinos/belos 

## Description
The flexible Gmres variant is accessed through the block Gmres solver manager.
The current logic in that solver will throw an error if there is no right
preconditioner specified in the linear problem.  If the right preconditioner is not specified, then this
solver manager could just employ an unpreconditioned block Gmres iteration.
Furthermore, the solver manager did not effectively handle the situation when the linear
problem has a left preconditioner.  That is being addressed in this commit.  A modified Tpetra test covers the change from flexible to unpreconditioned Gmres when a right preconditioned is not supplied.

## Motivation and Context
A user experience with this solver found it unhelpful that the solver threw an error and there is a natural resolution that the solver can take to address the issue.

## How Has This Been Tested?

OSX, Serial, GCC 7.x with a modified Tpetra test that is also being included in the pull request.

## Checklist

### DIFF
--- a/packages/belos/src/BelosBlockGmresSolMgr.hpp
+++ b/packages/belos/src/BelosBlockGmresSolMgr.hpp
@@ -813,6 +813,18 @@ bool BlockGmresSolMgr<ScalarType,MV,OP>::checkStatusTest() {
   // Basic test checks maximum iterations and native residual.
   maxIterTest_ = Teuchos::rcp( new StatusTestMaxIters<ScalarType,MV,OP>( maxIters_ ) );
 
+  // Perform sanity checking for flexible Gmres here.
+  // NOTE:  If the user requests that the solver manager use flexible GMRES, but there is no right preconditioner, don't use flexible GMRES.
+  //        Throw an error is the user provided a left preconditioner, as that is inconsistent with flexible GMRES.
+  if (isFlexible_ && Teuchos::is_null(problem_->getRightPrec())) {
+    isFlexible_ = false;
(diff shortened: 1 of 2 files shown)


packages/belos/src/BelosBlockGmresSolMgr.hpp
```diff
@@ -813,6 +813,18 @@ bool BlockGmresSolMgr<ScalarType,MV,OP>::checkStatusTest() {
   // Basic test checks maximum iterations and native residual.
   maxIterTest_ = Teuchos::rcp( new StatusTestMaxIters<ScalarType,MV,OP>( maxIters_ ) );
 
+  // Perform sanity checking for flexible Gmres here.
+  // NOTE:  If the user requests that the solver manager use flexible GMRES, but there is no right preconditioner, don't use flexible GMRES.
+  //        Throw an error is the user provided a left preconditioner, as that is inconsistent with flexible GMRES.
+  if (isFlexible_ && Teuchos::is_null(problem_->getRightPrec())) {
+    isFlexible_ = false;
```
mhoemmen: What if users later introduce a right preconditioner?  The above code would permanently change `isFlexible_`, which users might find surprising.  We might instead want a "we are actually going to do Flexible GMRES" flag that is the disjunction of "the user asked for Flexible GMRES" (what `isFlexible_` currently means) and "we have a right preconditioner."

PR #4140, opened 2019-01-04
Cleanup panzer Tpetra use

### LINKED ISSUE #4136: Panzer: address concerns in #4134
Address changes suggested by @mhoemmen in PR #4134

rppawlo:
Closes #4136

### DIFF
--- a/packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.cpp
+++ b/packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.cpp
@@ -40,6 +40,7 @@
 // ***********************************************************************
 // @HEADER
 
+#include "Panzer_NodeType.hpp"
 #include "Panzer_STK_LocalMeshUtilities.hpp"
 #include "Panzer_STK_Interface.hpp"
 #include "Panzer_STK_SetupUtilities.hpp"
@@ -182,7 +183,7 @@ buildCellToNodes(panzer::ConnManager<LO,GO> & conn, Kokkos::View<GO**> & globals
 }
 
 template <typename LO,typename GO>
-Teuchos::RCP<const Tpetra::Map<LO,GO> >
+Teuchos::RCP<const Tpetra::Map<LO,GO,panzer::TpetraNodeType> >
 buildNodeMap(const Teuchos::RCP<const Teuchos::Comm<int> > & comm,
                     Kokkos::View<const GO**> cells_to_nodes)
 {
@@ -201,7 +202,7 @@ buildNodeMap(const Teuchos::RCP<const Teuchos::Comm<int> > & comm,
 
    */
 
-  typedef Tpetra::Map<LO,GO> map_type;
+  typedef Tpetra::Map<LO,GO,panzer::TpetraNodeType> map_type;
 
   // get locally unique global ids
   std::set<GO> global_nodes;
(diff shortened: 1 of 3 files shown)


packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.cpp
```diff
@@ -326,7 +326,7 @@ buildGhostedCellOneRing(const Teuchos::RCP<const Teuchos::Comm<int> > & comm,
 {
 
   PANZER_FUNC_TIME_MONITOR("panzer_stk::buildGhostedCellOneRing");
-  typedef Tpetra::CrsMatrix<int,int,GO> crs_type;
+  typedef Tpetra::CrsMatrix<int,int,GO,panzer::TpetraNodeType> crs_type;
```
mhoemmen: In some places, you template on LO, but here, you use `LO=int`.  We haven't tested with `LO != int` for years, but you may want to fix this inconsistency at some point (not necessarily now).

packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.cpp
```diff
@@ -411,13 +411,13 @@ buildGhostedCellOneRing(const Teuchos::RCP<const Teuchos::Comm<int> > & comm,
   */
 template <typename GO>
 Kokkos::DynRankView<double,PHX::Device>
-buildGhostedVertices(const Tpetra::Import<int,GO> & importer,
+buildGhostedVertices(const Tpetra::Import<int,GO,panzer::TpetraNodeType> & importer,
```
mhoemmen: See above about `LO=int`.

packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.cpp
```diff
@@ -411,13 +411,13 @@ buildGhostedCellOneRing(const Teuchos::RCP<const Teuchos::Comm<int> > & comm,
...
                      Kokkos::DynRankView<const double,PHX::Device> owned_vertices)
 {
   using Teuchos::RCP;
   using Teuchos::rcp;
 
-  typedef Tpetra::MultiVector<double,int,GO> mvec_type;
+  typedef Tpetra::MultiVector<double,int,GO,panzer::TpetraNodeType> mvec_type;
```
mhoemmen: See above about `LO = int`.

packages/panzer/disc-fe/src/Panzer_L2Projection_impl.hpp
```diff
@@ -261,7 +261,8 @@ namespace panzer {
 
     {
       PANZER_FUNC_TIME_MONITOR("Exporting of mass matrix");
-      ghostedMatrix->fillComplete();
+      auto ghostedMap = factory.getGhostedMap(0);
+      ghostedMatrix->fillComplete(ghostedMap,ghostedMap);
```
mhoemmen: My guess is that `ghostedMap` is not one to one.  The domain Map and range Map (two arguments to fillComplete) _must_ be one to one.  My guess is that you want to use the row Map from the `ownedMatrix` as the domain and range Maps.

packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.cpp
rppawlo: yeah - we were supporting LO as a template parameter back in 2015/2016. Then about 2 years ago we decided LO should always be int in panzer. So newer code just uses int directly  and older code still has LO as a template parameter. We should fix this, just hasn't been high on the priority list.

packages/panzer/disc-fe/src/Panzer_L2Projection_impl.hpp
rppawlo: @mhoemmen - this is fixed. thanks for catching!

eric-c-cyr:
Looking at this, I think that the key point is that there is only one set of LIDs for each cell. So the fact that the column and row maps are chosen the way there are is to support this requirement. I was always fuzzy about some of the functionality in the Petra world in terms of fillComplete. So I was usually conservative and more explicit then needed. But I believe the use of consistent LIDs is critical unless you want to carry around row LIDs and column LIDs for each cell for each matrix block.

mhoemmen:
It's OK to choose the row and column Maps you want, but you have to pick one-to-one domain and range Maps.  However that works is your choice.

rppawlo:
Ok - I just pushed fixes for the correct maps. Can one of you approve this again?

Issue #4145, opened 2019-01-07
Teuchos stacked timer build failure in all ATDM builds


fryeguy52:
CC: @trilinos/teuchos, @jwillenbring (Trilinos Framework Product Lead), @bartlettroscoe, @fryeguy52
 
## Description
 
As shown in [this query](https://testing.sandia.gov/cdash/index.php?project=Trilinos&date=&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=Trilinos-atdm-) the test `TeuchosComm_stacked_timer_MPI_2` is not being run in all ATDM builds due to a build failure

`Error building packages/teuchos/comm/test/StackedTimer/CMakeFiles/TeuchosComm_stacked_timer.dir/stacked_timer.cpp.o `

```
/scratch/rabartl/Trilinos.base/NightlyBuilds/Trilinos-atdm-cee-rhel6-clang-5.0.1-openmpi-1.10.2-serial-static-opt/SRC_AND_BUILD/Trilinos/packages/teuchos/comm/test/StackedTimer/stacked_timer.cpp:357:3: error: use of undeclared identifier 'Kokkos'
  Kokkos::initialize(argc,argv);
  ^
/scratch/rabartl/Trilinos.base/NightlyBuilds/Trilinos-atdm-cee-rhel6-clang-5.0.1-openmpi-1.10.2-serial-static-opt/SRC_AND_BUILD/Trilinos/packages/teuchos/comm/test/StackedTimer/stacked_timer.cpp:365:7: error: use of undeclared identifier 'Kokkos'
  if (Kokkos::is_initialized())
      ^
/scratch/rabartl/Trilinos.base/NightlyBuilds/Trilinos-atdm-cee-rhel6-clang-5.0.1-openmpi-1.10.2-serial-static-opt/SRC_AND_BUILD/Trilinos/packages/teuchos/comm/test/StackedTimer/stacked_timer.cpp:366:5: error: use of undeclared identifier 'Kokkos'
    Kokkos::finalize_all();
    ^
3 errors generated.
```

The new commits on the day this started failing are:
```
65c3205:  Teuchos: Fix StackedTimer reporting
Author: Roger Pawlowski <rppawlo@sandia.gov>
Date:   Thu Jan 3 19:16:17 2019 -0700

M	packages/teuchos/comm/src/Teuchos_StackedTimer.hpp
M	packages/teuchos/comm/test/StackedTimer/stacked_timer.cpp

[...]

PR #4146, opened 2019-01-07
Teuchos: fix for #4145

rppawlo:
Addresses #4145.

### DIFF
--- a/packages/teuchos/comm/test/StackedTimer/stacked_timer.cpp
+++ b/packages/teuchos/comm/test/StackedTimer/stacked_timer.cpp
@@ -354,7 +354,9 @@ int main( int argc, char* argv[] )
   // Note that the dtor for GlobalMPISession will call
   // Kokkos::finalize_all().
   Teuchos::GlobalMPISession mpiSession(&argc, &argv);
+#if defined(HAVE_TEUCHOS_KOKKOS_PROFILING) && defined(HAVE_TEUCHOSCORE_KOKKOSCORE)
   Kokkos::initialize(argc,argv);
+#endif
   {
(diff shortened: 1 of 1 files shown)


rppawlo:
@mhoemmen - I'm just being consistent with how Teuchos::Time implemented support for kokkos profiling. According to git blame it was you that added both macros ;)

```
6892e170 packages/teuchos/core/src/Teuchos_Time.cpp (Mark Hoemmen          2018-02-21 16:43:02 -0700  81) #if defined(HAVE_TEUCHOS_KOKKOS_PROFILING) && defined(HAVE_TEUCHOSCORE_KOKKOSCORE)
```

rppawlo:
@mhoemmen - I take that back. You were the last to modify that line but @cgcgcg initially added it. It looks like we needed to enable kokkos yet disable the profiling for some reason.

rppawlo:
@bartlettroscoe - it looks like profiling is disabled in all atdm builds:
```
S1002730:atdm rppawlo$ find . -name '*' | xargs grep -is PROF
./ATDMDevEnvSettings.cmake:ATDM_SET_CACHE(Kokkos_ENABLE_Profiling OFF CACHE BOOL)
```
This flag should probably be added to one of the PR testing builds.

mhoemmen:
@rppawlo wrote:
> It looks like we needed to enable kokkos yet disable the profiling for some reason.

Let's find out that reason -- Kokkos profiling is supposed to be extremely low overhead, and Tpetra uses Kokkos profiling regions quite a bit, so I think y'all could benefit from turning it on.  If it's broken, we would like to fix it :)

PR #4151, opened 2019-01-08
Stacked timer unit testing support

rppawlo:
Some applications are writing unit testing scripts that rely on StackedTimer data. This commit adds queries and overrides of Timer values to help support unit testing of the StackedTimer for both Trilinos and within the applications.

Also cleaned up some inconsistent naming conventions.

### DIFF
--- a/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
+++ b/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
@@ -37,9 +37,9 @@ StackedTimer::LevelTimer::report(std::ostream &os) {
 
 }
 
-BaseTimer*
-StackedTimer::LevelTimer::findBaseTimer(const std::string &name) {
-  BaseTimer* t = NULL;
+const BaseTimer*
+StackedTimer::LevelTimer::findBaseTimer(const std::string &name) const {
+  const BaseTimer* t = NULL;
   if (get_full_name() == name) {
     return this;
   }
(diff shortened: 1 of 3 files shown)


packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -37,6 +37,22 @@ StackedTimer::LevelTimer::report(std::ostream &os) {
 
 }
 
+const BaseTimer*
+StackedTimer::LevelTimer::findBaseTimer(const std::string &name) const {
+  const BaseTimer* t = NULL;
```
mhoemmen: FYI for the future, we can and should use `nullptr` instead of `NULL`.

packages/teuchos/comm/src/Teuchos_StackedTimer.hpp
```diff
@@ -384,6 +397,13 @@ class StackedTimer
       */
      void report(std::ostream &os);
 
+    /**
+     * Return pointer to the BaseTimer corresponding to a given string
+     * @param name input string to search for
+     * @return pointer to BaseTimer (NULL if none found)
```
mhoemmen: Is this the same method as the one below?  The method implementation below throws if it finds `NULL`.

rppawlo: Thanks @mhoemmen ! The automerge pushed it so I will make these changes in another pr tomorrow. Just need to correct the doxygen.

rppawlo: The doxygen comment about returning null is correct. There are two implementations of findBaseTimer - the top level StackeTimer method that does throw an exception and the internal LevelTimer method that returns nullptr. The StackedTimer will throw if you don't find the timer, but as the we search through each sublevel of the timer tree, we don't want to throw but just return that it was not found in this particular subtree.

PR #4157, opened 2019-01-09
Trilinos: Code cleanup

rppawlo:
This set of commits covers the following:
1. Fix compiler warnings for unused parameter in Ifpack
2. Fix (most) shadow warnings in PIRO
3. Fix all shadow warnings in Panzer
4. Fix ptr use in Teuchos StackedTimer as suggested in #4151

### DIFF
--- a/packages/panzer/adapters-stk/src/Panzer_STK_ModelEvaluatorFactory_impl.hpp
+++ b/packages/panzer/adapters-stk/src/Panzer_STK_ModelEvaluatorFactory_impl.hpp
@@ -308,7 +308,7 @@ namespace panzer_stk {
     }
     m_is_transient = is_transient;
 
-    bool useDiscreteAdjoint = p.get<bool>("Use Discrete Adjoint");
+    useDiscreteAdjoint = p.get<bool>("Use Discrete Adjoint");
 
     ////////////////////////////////////////////////////////////////////////////////////////
     ////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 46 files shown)


packages/panzer/adapters-stk/src/Panzer_STK_ModelEvaluatorFactory_impl.hpp
```diff
@@ -602,8 +602,8 @@ namespace panzer_stk {
     // see if field coordinates are required, if so reset the workset container
     // and set the coordinates to be associated with a field in the mesh
     useDynamicCoordinates_ = false;
-    for(std::size_t p=0;p<physicsBlocks.size();p++) {
-      if(physicsBlocks[p]->getCoordinateDOFs().size()>0) {
+    for(std::size_t pb=0;pb<physicsBlocks.size();pb++) {
```
mhoemmen: For the next cleanup pass: If `physicsBlocks` is a standard container, prefer a loop of the form `for (auto&& block : physicsBlocks)`.

packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_Interface.cpp
```diff
@@ -123,7 +123,7 @@ void STK_Interface::addSideset(const std::string & name,const CellTopologyData *
    TEUCHOS_ASSERT(dimension_!=0);
 
    stk::mesh::Part * sideset = metaData_->get_part(name);
-   if(sideset==NULL)
+   if(sideset==nullptr)
```
mhoemmen: Thanks :D

rppawlo:
looks like gcc 4.8 compiler is getting confused with template type deduction and nullptr. Just pushed an attempted fix.

PR #4158, opened 2019-01-09
Fixed it so you use a numeric_limits call rather than a stupid math f…

bathmatt:
If you enable FPE checking too early teuchos throws an exception, this fixes that..

### DIFF
--- a/packages/teuchos/core/src/Teuchos_ScalarTraits.cpp
+++ b/packages/teuchos/core/src/Teuchos_ScalarTraits.cpp
@@ -41,6 +41,7 @@
 
 #include "Teuchos_ScalarTraits.hpp"
 #include "Teuchos_Assert.hpp"
+#include <limits>
 
 // Define this to throw exceptions when any Teuchos::ScalarTraits function
 // encounters a NaN or an Inf.
@@ -119,7 +120,7 @@ bool Teuchos::operator&&(const qd_real &a, const qd_real &b) {
 #ifndef __sun
 // This is an intentional computation of NaN.
 namespace Teuchos {
-const float  flt_nan = +returnFloatZero()/returnFloatZero();
-const double dbl_nan = +returnDoubleZero()/returnDoubleZero();
+  const float  flt_nan = std::numeric_limits<float>::signaling_NaN();
+  const double dbl_nan = std::numeric_limits<double>::signaling_NaN();
 }
 #endif


packages/teuchos/core/src/Teuchos_ScalarTraits.cpp
```diff
@@ -119,7 +120,7 @@ bool Teuchos::operator&&(const qd_real &a, const qd_real &b) {
 #ifndef __sun
```
mhoemmen: I'm pretty sure we could purge the work-arounds for broken old Sun compilers, but we don't have to do that in this commit.

packages/teuchos/core/src/Teuchos_ScalarTraits.cpp
```diff
@@ -119,7 +120,7 @@ bool Teuchos::operator&&(const qd_real &a, const qd_real &b) {
...
 // This is an intentional computation of NaN.
 namespace Teuchos {
-const float  flt_nan = +returnFloatZero()/returnFloatZero();
-const double dbl_nan = +returnDoubleZero()/returnDoubleZero();
+  const float  flt_nan = std::numeric_limits<float>::signaling_NaN();
```
mhoemmen: @bathmatt wrote:
> If you enable FPE checking too early teuchos throws an exception, this fixes that..

It looks like this creates a signaling NaN.  Wouldn't you want a quiet NaN (`std::numeric_limits<float>::quiet_NaN()`) instead?

bathmatt: Well, if one uses it should it signal?  I'm fine with quiet_NaN since we don't use it anywhere,   But if I was to use flt_nan as a default to make sure my code died if I used uninitialized memory I'd be angry that it wsa a quiet one.

bathmatt: Also, it was a signaling nan before.

packages/teuchos/core/src/Teuchos_ScalarTraits.cpp
bathmatt: There's several of them.

```
dir-teuchos (gcc) 261 $ find . -name \*pp -exec grep -l __sun {} \;
./core/test/MemoryManagement/ArrayRCP_test.cpp
./core/src/Teuchos_ArrayView.hpp
./core/src/Teuchos_ScalarTraits.hpp
./core/src/Teuchos_ArrayViewDecl.hpp
./core/src/Teuchos_ScalarTraits.cpp
./remainder/src/Teuchos_FloatingPointTrap.cpp

```

packages/teuchos/core/src/Teuchos_ScalarTraits.cpp
mhoemmen: @bathmatt Try it and see if it passes the tests?

rppawlo:
Albany is showing new failures due to FPEs today. Suspect it is this commit.

https://github.com/SNLComputation/Albany/issues/422

PR #4168, opened 2019-01-10
Introducing coordinateType into Teuchos::ScalarTraits and disabling ensemble reduction by default

### LINKED ISSUE #4150: Using magnitudeType for coordinates in MueLu
@trilinos/muelu @kliegeois 

MueLu developers,

I have a student, @kliegeois, working on a modification of one of Stokhos' scalar types that changes Teuchos::ScalarTraits::magnitudeType to something other than a builtin type, and we are running into many issues in MueLu, where magnitudeType is feely converted to/from double.  The issue seems to be that the interface for mesh coordinates takes vectors of magnitudeType, but the implementation often assigns these values to double types.  We started converting most of this code to use magnitudeType throughout, but this is becoming untenable and we are wondering what the right path forward is.  

It seems to me that using magnitudeType for coordinates is something of an abuse of what magnitudeType is supposed to represent, since it is supposed to represent magnitudes (i.e., real, non-negative numbers).  What would you think of introducing a new type trait for coordinates that would compute the coordinate type from the scalar type, but be divorced from magnitudeType?  This could be done using Teuchos::ScalarTraits (e.g., Teuchos::ScalarTraits::coordinateType), or MueLu could introduce its own trait, something like:
```
// Default implementation where CoordinateTraits<Scalar>::type == Scalar
template <class Scalar>
struct CoordinateTraits {
  typedef Scalar type;
};

// Specialization for std::complex<>
template <class T> {
struct CoordinateTraits< std::complex<T> > {
  typedef T type;
};
```
Then Stokhos would then provide its own specialization where CoordinateTraits<>::type is a builtin type.

To me, this seems like the right way to do this.  If someone wanted to use, e.g., extended precision types, then magnitudeType would be extended precision (and therefore not safely convertible to double), but it seems very unlikely that anyone would want the coordinate type to be extended precision.

kliegeois:
@trilinos/muelu @trilinos/stokhos

## Description

The main change consists in introducing a coordinateType into the Teuchos::ScalarTraits and use it in the related files in MueLu.
The usage of such coordinateType prevents the cast from magnitudeType to double as discussed in the issue #4150.
In order to prevent further commits from introducing cast from magnitudeType to double, the ensemble reduction in Stokhos has been switched to off by default.
This implies that the magnitudeType of MP_vector are now MP_vector by default and there is no valid cast from MP_vector to double.

Setting the ensemble reduction to off raised build errors that have been corrected in commits of this pull request.

## Motivation and Context

* Closes #4150

## How Has This Been Tested?

It has been tested with ctest with one Intel and one clang build.

## Checklist


packages/belos/src/BelosTypes.cpp
```diff
@@ -189,5 +189,14 @@ namespace Belos {
     }
   }
 
+  // Initialize DefaultSolverParameters.  Has to be done this way because
```
mhoemmen: Could you be more specific about these linking errors?  I haven't seen those with other static constexpr stuff in Belos.

packages/stokhos/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_Experimental
 TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_Ensemble_Reduction
   HAVE_STOKHOS_ENSEMBLE_REDUCT
   "Enable reduction across across ensemble components."
-  ON )
+  OFF )
```
mhoemmen: I don't see how this is relevant to this PR.

packages/teuchos/core/src/Teuchos_ScalarTraitsDecl.hpp
```diff
@@ -95,6 +95,8 @@ struct ScalarTraits
   typedef T halfPrecision;
   //! Typedef for double precision
   typedef T doublePrecision;
+  //! Typedef for coordinates
+  typedef T coordinateType;
```
mhoemmen: I don't see any specializations here for complex.  Note that PR testing may not enable complex -- you'll have to do that manually.

packages/belos/src/BelosTypes.cpp
etphipp: See [here](https://stackoverflow.com/questions/26196095/static-constexpr-odr-used-or-not).  In short, if a static constexpr is used as an argument to a function, it may be odr-used (depending on the compiler and optimization), in which case the constant must have a definition (and `static constexpr double blah = ...;` is not a definition).

packages/stokhos/CMakeLists.txt
etphipp: It's relevant because this whole PR is about making MueLu work when magnitudeType != double, and this change makes that the default case for Stokhos, to ensure it doesn't break in the future.

packages/teuchos/core/src/Teuchos_ScalarTraitsDecl.hpp
etphipp: Line 1135.

packages/belos/src/BelosTypes.cpp
etphipp: Btw, this is a problem with C++11 that was fixed in C++17.

mhoemmen: @etphipp I'm glad C++17 fixed that :-D

packages/stokhos/CMakeLists.txt
mhoemmen: @etphipp thanks for explaining!

packages/teuchos/core/src/Teuchos_ScalarTraitsDecl.hpp
mhoemmen: Ah, OK, my bad!

Issue #4169, opened 2019-01-10
Tempus: FPEs caused by use of std::numeric_limits::max()


ccober6:
EMPIRE identified FPEs in Tempus when using std::numeric_limits::max() to initialize variables with a default value.

@trilinos/tempus 

## Expectations

Should be able to run without FPEs.

## Possible Solution

Replace std::numeric_limits::max() with "reasonable" large default values.

PR #4171, opened 2019-01-11
Tempus: Fixed some FPEs.

ccober6:
Replaced the usage std::numeric_limits::max() for the default values
as it can cause FPEs.  Rebaselined the Integrator tests as it checks
the default values.

All tests pass on MacOS.

#4169 

@trilinos/tempus 

## Description

Replaced std::numeric_limits<Scalar>::max() with 1.0e+99, and std::numeric_limits<int>::max() with 1000000000.

## Motivation and Context

Avoid FPEs.

## How Has This Been Tested?

All test pass on MacOS.

## Checklist


krcb:
Looking at this now; need to build with an application code to see if the FPE's are resolved, so it might take a little time.

krcb:
This looks like it addresses the EMPIRE issue.

PR #4180, opened 2019-01-12
Made nan's quiet to hide FPEs

bathmatt:
make nan's quiet

### DIFF
--- a/packages/teuchos/core/src/Teuchos_ScalarTraits.cpp
+++ b/packages/teuchos/core/src/Teuchos_ScalarTraits.cpp
@@ -105,7 +105,7 @@ bool Teuchos::operator&&(const qd_real &a, const qd_real &b) {
 #ifndef __sun
 // This is an intentional computation of NaN.
 namespace Teuchos {
-  const float  flt_nan = std::numeric_limits<float>::signaling_NaN();
-  const double dbl_nan = std::numeric_limits<double>::signaling_NaN();
+  const float  flt_nan = std::numeric_limits<float>::quiet_NaN();
+  const double dbl_nan = std::numeric_limits<double>::quiet_NaN();
 }
 #endif


mhoemmen:
This also brings `Teuchos::ScalarTraits`' behavior in line with that of `Kokkos::ArithTraits`.  See https://github.com/kokkos/kokkos-kernels/issues/35.

bathmatt:
I still don't understand the point of using a quiet_nan vs a signaling nan, if the point of setting something to NaN is that you catch errors, wouldn't you want to catch those errors when they start?  Quiet nan seems useful for a flag, but since you can do math and the quiet bit propagates to all the other nans it seems somewhat pointless.  

I do understand that you can throw nan's in algorithms in a harmless way, then you might wanna use quiet, but I think that caching the FPE handling flags, handling the case where you go NaN and then resetting them after is a better way to handle this case.

krcb:
I agree with @bathmatt. Switching to the signaling NaN allowed an application to catch a range of errors in a Trilinos package that were generating NaNs quietly before. It seems that when an algorithm wants to use NaNs in a quiet way, this should be done explicitly. However, it appears though that the Teuchos::ScalarTraits::nan() implementation is being used both when an algorithm wants a harmless NaN and to indicate there's something wrong (e.g. in the Teuchos::ScalarTraits::squareroot() implementation). How is downstream application supposed to distinguish between these two cases?

mhoemmen:
> How is downstream application supposed to distinguish between these two cases?

There's no way to distinguish.  You just have to know that an algorithm might generate harmless NaNs, and turn off signaling while that algorithm runs.

I like IEEE 754 and I want users to have the option to enable signaling NaNs, but I think changing Teuchos (nearly the most upstream package in Trilinos) without a lot of application testing was premature.

krcb:
@mhoemmen Would a possible compromise be to add a compile time option to switch between quiet and signaling NaNs? This option would default to the quiet NaN case and and then an application could decide whether or not to turn it on.

mhoemmen:
> Would a possible compromise be to add a compile time option to switch between quiet and signaling NaNs?

I'm not a fan of adding Still More Compile-Time Options.  No one will ever test the signaling one, then someone will try it and it won't compile, exactly at the time when you needed it to.

If you decide to do it, note that the option would need to control both Teuchos::ScalarTraits and Kokkos::ArithTraits.  Right now, you've only been looking at ScalarTraits.

PR #4198, opened 2019-01-15
Thyra: add function to reset the ModelEvaluatorDefaultBase lazy initialization flag

rppawlo:
This change allows Drekar to run correctly with Trilinos_ENABLE_DEBUG=ON (in particular IMEX and adjoint problems).

Details on why this is needed are in the new unit test.


rppawlo:
@bartlettroscoe - Trilinos_ENABLE_DEBUG=ON should already work for EMPIRE (and Panzer) builds. Both Matt and I use them regularly without any problems. The automated builds of empire trilinos always do a debug-debug version with that flag enabled as far as I know (unless something changed recently). This push is for Drekar for some corner cases. This finally allows Drekar's test suite to pass all tests with Trilinos_ENABLE_DEBUG=ON.

rppawlo:
> Is that worth trying this with ATDM after this PR is merged?

yes

Issue #2464, opened 2018-03-27
Set up a CUDA build for an auto PR build


bartlettroscoe:
CC: @trilinos/framework, @mhoemmen, @rppawlo, @ibaned, @crtrott, @nmhamster 

## Description

This Issue is to scope out and track efforts to set up a CUDA build of Trilinos to be used as an auto PR build as described in https://github.com/trilinos/Trilinos/issues/2317#issuecomment-376551457.

For this build it was agreed to use that ATDM build on `white` that is currently running and submitting to CDash.  Questions about how to extend this build to be used as an auto PR build include:

* Should all of the PT package tests be built and run or just the ones that the current ATDM build of Trilinos builds and runs?  (Things can be set up to either way.)
* Does the machine 'white' have enough computing capacity in order to handle the load of builds needed for Trilinos PR testing?
* Are the Jenkins jobs running the builds using the `bsub` command robust enough to be a reliable PR build?

## Tasks:

1. Clean up the existing CUDA build on `white` until it is 100% clean [Done]
1. Set up an all-at-once nightly build that enables all PT package that submits to CDash "Specialized" [Done]
1. Clean up the all-at-once nightly build for all PT packages (disable whatever should be disabled) ...
1. ???

## Related Issues:

* Part of: #2317

bartlettroscoe:
FYI: We asked @nmhamster about using the `rhel7F` nodes on `white` for auto PR builds and he said we could try this.

Note that the target for this build should be the `Trilinos-atdm-white-ride-cuda-debug` build and not the `Trilinos-atdm-white-ride-cuda-opt` build due to the large number of segfaulting tests on the latter build described in #2454.  I will focus on cleaning up the `cuda-debug` build as there are just a few failing tests at this point.

Other than setting up the Jenkins job and cleaning up any Trilinos failures with this setup, my biggest concern is the stability of the Jenkins jobs on 'white'.  For example, if you look at the history of the nightly build `Trilinos-atdm-white-ride-cuda-debug` on `white` shown at:

* https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-white-ride-cuda-debug&field2=site&compare2=63&value2=white&field3=buildstarttime&compare3=84&value3=now

[...]

PR #4216, opened 2019-01-18
Add infrastructure for CUDA PR build

jwillenbring:
@trilinos/framework 

## Description

This PR adds/modifies files in preparation of adding a dev->master CUDA build and then a PR CUDA build.

## Motivation and Context

We need to protect against breaking the CUDA build at PR time. This is in response to #2464.

## How Has This Been Tested?

@ZUUL42 set up Jenkins infrastructure for this build also, and we have a [large subset of Trilinos tests passing](https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&parentid=4425385). We have disabled temporarily what is failing so that we can get a PR build in place that will protect what is passing while we work on the remaining issues.


jwillenbring:
We cannot work on this PR until we refine the set of commits this PR includes.

PR #4217, opened 2019-01-18
Add infrastructure for CUDA PR build

jwillenbring:
@trilinos/framework 

## Description

This PR adds/modifies files in preparation of adding a dev->master CUDA build and then a PR CUDA build.

## Motivation and Context

We need to protect against breaking the CUDA build at PR time. This is in response to #2464.

## How Has This Been Tested?

@ZUUL42 set up Jenkins infrastructure for this build also, and we have a [large subset of Trilinos tests passing](https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&parentid=4425385). We have disabled temporarily what is failing so that we can get a PR build in place that will protect what is passing while we work on the remaining issues.


cmake/std/PullRequestLinuxDriver-Test.sh
```diff
@@ -150,13 +150,16 @@ echo -e ""
...
 
-source /projects/sems/modulefiles/utils/sems-modules-init.sh
-
 declare -i ierror=0
 #Have to keep loading git
-module load sems-git/2.10.1
-
-
+regex=".*(_cuda_).*"
+if [[ ! ${JOB_BASE_NAME:?} =~ ${regex} ]]; then
```
william76: I checked this regex test in a script and it works... but I'm not sure we should make an assumption that if the job has `_cuda_` in its name that it's on Ride.  This might lead to bugs down the line if we create a CUDA job later on for a machine that isn't Ride... I'd rather see something like:

```
if job_name doesn't have CUDA:
    load non-cuda environment
else if machine is Ride
    load git module for Ride
else
    throw a meaningful error and notify DevOPs
fi
```

@prwolfe @jwillenbring your thougs?

ZUUL42: There might be a better way to word it, but, as things currently are, I was trying more so to warn someone if they are to create a CUDA job, they should put it on Ride... as it is currently the only one that is configured to run the rest of the CUDA build script that is to follow.
If we change that, we'll change/remove the note.
But it doesn't _need_ to be there, I don't have an issue with removing it.

prwolfe: Agreed - Thanks @ZUUL42

jwillenbring:
As discussed at the @trilinos/framework meeting this morning, I fired off another PR build (not the CUDA build) to make sure we didn't break anything for other cases. [The build came back clean](https://testing.sandia.gov/cdash/index.php?project=Trilinos&parentid=4437523), but we will wait for @ZUUL42 to make the other change and test again before merging.

cmake/std/PullRequestLinuxDriver-Test.sh
ZUUL42: Committed the additional check.

cmake/std/PullRequestLinuxCudaDriver.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -l
+
+if [ "${BSUB_CTEST_TIME_LIMIT}" == "" ] ; then
```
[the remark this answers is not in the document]
jwillenbring: @bartlettroscoe @ZUUL42 @william76 

> Is BSUB_CTEST_TIME_LIMIT a system defined environment variable or is this a new Jenkins parameter that's being introduced?
If this doesn't exist, can it break things?

I am not certain, but I think it is just an environment variable that can be defined by a user. 
@bartlettroscoe may be able to confirm or correct this.
If it is not defined, it is set to a default value, so it shouldn't break things. The variable is specific to this build as far as current PR tests go (because ride uses bsub).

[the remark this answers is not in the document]
bartlettroscoe: This was just an env var that allowed specialization in the individual driver scripts.  This came from the file:

* Trilinos/cmake/ctest/drivers/atdm/ride/local-driver.sh

which has:

```
#!/bin/bash -l

if [ "${BSUB_CTEST_TIME_LIMIT}" == "" ] ; then
  export BSUB_CTEST_TIME_LIMIT=12:00
fi

...

source $WORKSPACE/Trilinos/cmake/std/atdm/load-env.sh $JOB_NAME

set -x

bsub -x -Is -q $ATDM_CONFIG_QUEUE -n 16 -J $JOB_NAME -W $BSUB_CTEST_TIME_LIMIT \
  $WORKSPACE/Trilinos/cmake/ctest/drivers/atdm/ctest-s-driver.sh
```

Any questions about this?

william76:
@ZUUL42 looks good to me.

Issue #4221, opened 2019-01-21
SPARC source build error from Tpetra_Experimental_BlockCrsMatrix_decl.hpp -Werror for shadowing warning with Clang 5.0.1 build


bartlettroscoe:
CC: @trilinos/tpetra, @kddevin  (Trilinos Data Services Product Lead), @mhoemmen, @bartlettroscoe, @fryeguy52, @micahahoward, @sebrowne 

## Next Action Status

<status-and-or-first-action>

## Description

As of the Trilinos 'develop' version 882d842b73dc475291b1e2945a9ba532577e2f1d, the Clang 5.0.1 build of SPARC (using the `cee-rhel6_clang-5.0.1_openmpi-1.10.2_serial_static_opt` build of Trilinos) is broken which uses -Werror in the build showing the build error:

```
In file included from /scratch/rabartl/SPARC.base/sparc/Trilinos/cee-rhel6_clang-5.0.1_openmpi-1.10.2_serial_static_opt/include/Xpetra_TpetraBlockCrsMatrix.hpp:53:
In file included from /scratch/rabartl/SPARC.base/sparc/Trilinos/cee-rhel6_clang-5.0.1_openmpi-1.10.2_serial_static_opt/include/Tpetra_Experimental_BlockCrsMatrix.hpp:1:
/scratch/rabartl/SPARC.base/sparc/Trilinos/cee-rhel6_clang-5.0.1_openmpi-1.10.2_serial_static_opt/include/Tpetra_Experimental_BlockCrsMatrix_decl.hpp:757:3: error: 'Tpetra::Experimental::BlockCrsMatrix<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >::packAndPrepareNew' hides overloaded virtual function [-Werror,-Woverloaded-virtual]
  packAndPrepareNew (const SrcDistObject& sourceObj,
  ^
../include/linsys/MatrixTpetraBlockCrs.h:71:25: note: in instantiation of template class 'Tpetra::Experimental::BlockCrsMatrix<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >' requested here
      return tpetra_mat_->getNodeNumRows() * block_size_;
                        ^
[...]

PR #4226, opened 2019-01-21
Tpetra - fix warning error from mismatched virtual functions

kyungjoo-kim:
## Description

Fixiing mismatched virtual functions definitions. This could be a real error but somehow it works with other compilers and PR tests.  
## Related Issues

#4221 

## How Has This Been Tested?

As the error shows up from SRN machines with a clang compiler and this is not detected from PR tests, I cannot attempt to reproduce the error myself. From the error message, I matched the function signatures and hope that it will solve the problem.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Experimental_BlockCrsMatrix_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_Experimental_BlockCrsMatrix_decl.hpp
@@ -756,14 +756,14 @@ protected:
   virtual void
   packAndPrepareNew (const SrcDistObject& sourceObj,
                      const Kokkos::DualView<const local_ordinal_type*, device_type>& exportLIDs,
-                     Kokkos::DualView<impl_scalar_type*, buffer_device_type>& exports,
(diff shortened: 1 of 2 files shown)


kyungjoo-kim:
Sorry. I did a mistake to put markLoalErrorAndGetStream() to get a stream without a guard. When I test on my workstation, I probably do not set debug mode in tpetra.

mhoemmen:
Thanks @kyungjoo-kim !  If you need help, just let me know.

> When I test on my workstation, I probably do not set debug mode in tpetra.

It should be on by default in a debug build (`CMAKE_BUILD_TYPE=DEBUG`).

kyungjoo-kim:
I did ``CMAKE_BUILD_TYPE=DEBUG`` but it does not enable debug macro. ``Trilinos_ENABLE_DEBUG:BOOL=ON`` is the right flag which I did not enable before.

mhoemmen:
Thanks @kyungjoo-kim ! :D

mhoemmen:
FYI @theguruat12 @rrdrake @micahahoward (Kyungjoo fixed a thing that might have blocked promotion)

PR #4230, opened 2019-01-21
WIP: Panzer - Changes to workset interface

### LINKED ISSUE #4205: Panzer: Cleanup of Workset interface
## Motivation and Context

Currently the workset is more or less a massive storage container for everything we require, and much of it is stored in structure form (direct access to class members). The goal of this update is to add a function interface to access these calls, and to strip everything out of worksets that isn't required.

The main reason for doing so is to allow for on-demand allocation and filling of geometry/topological data arrays as required by the derived code. Currently the workset allocates and fills *everything* that could possibly be required by the user, which takes lots of memory, and a bit of compute time. By making these calls on-demand, the total allocations made will be greatly reduced for most applications. 

Making the interface on-demand also cleans up the workset construction by not requiring the user to know what Basis/Cubature/Geometry/Topology is required by the derived FEM/FV/FD code during workset construction.

@trilinos/panzer 

## What do we keep/remove/redesign

Ideally, worksets would be a simple container that would look like:

```c++

class Workset
{
public:

  using Scalar = double;
  using Index = int;

  Index
  numDimensions() const;

  Index
  subcellIndex() const;

  Kokkos::View<const Scalar***>
  cellVertices() const;

  const  std::string &
  blockID() const;

  Kokkos::View<const Index*>
  localCellIndexes() const;

  const panzer::SubcellConnectivity & 
  getSubcellConnectivity(const int subcell_dimension) const;

  const panzer::IntegrationValues2<Scalar> & 
  getIntegrationValues(const panzer::IntegrationDescriptor & description) const;

  const panzer::IntegrationRule & 
  getIntegrationRule(const panzer::IntegrationDescriptor & description) const;

  panzer::BasisValues2<Scalar> & 
  getBasisValues(const panzer::BasisDescriptor & basis_description, 
                           const panzer::IntegrationDescriptor & integration_description) const;

  const panzer::BasisValues2<Scalar> & 
  getBasisValues(const panzer::BasisDescriptor & basis_description, 
                           const panzer::IntegrationDescriptor & integration_description) const;

  const panzer::BasisValues2<Scalar> & 
  getBasisValues(const panzer::BasisDescriptor & basis_description, 
                           const panzer::PointDescriptor & point_description) const;

  const panzer::PointValues2<Scalar> & 
  getPointValues(const panzer::PointDescriptor & point_description) const;

  const panzer::PureBasis & 
  getBasis(const panzer::BasisDescriptor & description) const;

  Index
  numCells() const;

  Index 
  numOwnedCells() const;

  Index 
  numGhostCells() const;

  Index 
  numVirtualCells() const;

  std::ostream & 
  operator<<(std::ostream& os, const panzer::Workset& w);

protected:
  ...
};

```

Naturally, there are a few issues with this. Having 'const' everywhere is required since worksets are passed around as const, but it will make things tricky when adding an on-demand feature. For now we can get away with having a shared pointer to an underlying class that does all the mutable stuff... assuming this doesn't cause issues with CUDA.

It also leaves the question of what to do with the rest of the interface. As far as I can tell, the above is missing the following (most of which probably shouldn't be in the workset):

```c++
Teuchos::RCP< std::vector<std::string> > basis_names;
void setNumberOfCells(int o_cells,int g_cells,int v_cells);
void setIdentifier(std::size_t identifier) { identifier_ = identifier; }
std::size_t getIdentifier() const { return identifier_; }

double alpha;
double beta;
double time;
double step_size;
double stage_number;
std::vector<double> gather_seeds; // generic gather seeds
bool evaluate_transient_terms;
Teuchos::RCP<WorksetDetails> other;

class WorksetDetailsAccessor;
```

I also don't know why there are multiple 'WorksetDetails' available in a workset.

If @eric-c-cyr and @rppawlo could run me though the reasoning behind the additional workset stuff, I would be grateful.

## Definition of Done

The updates to workset are straightforward. Updating Drekar, Charon, and the various EMPIRE codes will be a bit tricky so we will break it into a set of phases. Updating Charon and Drekar may require some help from @jmgate. Updates to Empire will probably go smoother with @CamelliaDPG.

### Phase 1

Workset interface changes:

- [ ] Merge Workset and WorksetDetails into a single class.
- [ ] Clean up and minimize interface based after discussing things with Eric and Roger.
- [ ] Clean up workset construction interface (currently there are multiple conflicting ways to do this)
- [ ] Deprecate the old interface
- [ ] Deprecate old construction scheme
- [ ] Update Drekar
- [ ] Update Charon
- [ ] Update EMPIRE

### Phase 2

Workset on-demand changes:

- [ ] Modify workset so that requesting things using `Descriptors` will spawn new BasisValues/PointValues/IntegrationRules/etc.
- [ ] Remove 'old' deprecated interface
- [ ] Remove alternative construction schemes

Note that we will allow the worksets to be fully allocated/filled if the user requests this using the construction interface (i.e. WorksetNeeds is passed to the construction scheme).

### Phase 3

Update BasisValues2 class

- [ ] Split `BasisValues2` into two classes: `BasisValues` and `BasisIntegrationValues`
- [ ] Add function interface to new classes
- [ ] Implement on-demand calculations of various datasets
- [ ] Deprecate old interface
- [ ] Update Drekar
- [ ] Update Charon
- [ ] Update EMPIRE

### Phase 4

Update IntegrationValues2 class

- [ ] Rename to `IntegrationValues`
- [ ] Add function interface to new classes
- [ ] Implement on-demand calculations of various datasets
- [ ] Deprecate old interface
- [ ] Update Drekar
- [ ] Update Charon
- [ ] Update EMPIRE

### Phase 5

Update PointValues2 class

- [ ] Rename to `PointValues`
- [ ] Add function interface to new classes
- [ ] Implement on-demand calculations of various datasets
- [ ] Deprecate old interface
- [ ] Update Drekar
- [ ] Update Charon
- [ ] Update EMPIRE

### Phase 6

Update subcell connectivity interface

- [ ] Add node/edge/face subcell connectivity to topological mesh analysis in Panzer (lots of questions here)
- [ ] Add node/edge/face subcell connectivities to workset (on-demand)
- [ ] Deprecate old interface
- [ ] Update Drekar
- [ ] Update Charon
- [ ] Update EMPIRE

### Phase 7

Cleanup

- [ ] Once everyone stops yelling at me, we can go ahead and remove the deprecated interface

seamill:
This is a WIP-PR to discuss changes to the workset interface - for review purposes only.

Closes #4205. 

@trilinos/panzer 

## Description
See #4205.

## Motivation and Context
See #4205.


seamill:
@rppawlo and @eric-c-cyr: This was a quick mod I made to the workset last week. I'd like to hear your thoughts on what is really needed. We have a quite a few things left to discuss concerning various coefficients and values stored in workset.

I'm wondering if we should have some kind of abstract/templated set/get interface that allows us to register objects in the workset that can be pulled out at a later time. Things like 'TimeStepper', 'FADDescriptor', and whatever else we may need to define the items in Workset. I only suggest this because, as it stands, it is very hard for a derived code to add anything to a workset without modifying Panzer directly.

seamill:
I added an auxiliary storage system to the workset. It is modifiable from outside of panzer, which will make the transition easier, and gives us a place to put all of our unknown quantities.

seamill:
@rppawlo @eric-c-cyr I made a couple of changes to the `WorksetContainer` to streamline workset construction and what the WorksetFactory is actually responsible for. 

I'm still a little lost on how to handle orientations. From what what done in EMPIRE-Fluid, all we need to generate orientations is a DOFManager (BlockedDOFManager), which requires:

- Connectivity Manager (just takes the STK_Interface which we have already)
- MPI Comm (already have it)
- Field Patterns (a problem if we don't have Needs being passed in)
- A couple of calls to `setOrientationsRequired(true)`, and `buildGlobalUnknowns()`.

As far as I can tell that is all that is needed to get orientations constructed. My question to you guys is this:

Can we do this on demand? If we set `initializeNeeds(needs)` in the workset (as opposed to the `WorksetContainer`), can we build a DOFManager on the fly and generate the required orientations? i.e. I assume we need the FieldPatterns to be set in the DOFManager before the orientations will be properly setup. 

Since the DOFManager has an MPIComm, I assume generating orientations is a collective call, and is therefore impossible to do things on a workset by workset basis.

What are your thoughts?

rppawlo:
@seamill - This is something I was hoping to clean up as part of the refactor. I like the build on demand idea. My preference is to allow two options here. If the user has a (Block)DOFManager available, they can call a function to set the orientation data using that DOFManager. If there is not one available, the user sets the building blocks required for for an internal temporary DOFManager.

setOrientationRequirements(RCP<DOFManager> );
setOrientationRequirements(RCP<ConnManager>,...);

Since orientations are really tied to a basis, the worksets really should not be dependent on a particular DOFManager. For example, a code coupling where each application could share the same worksets, but are solving for different equations/discretizations on the same mesh (thus have different DOF managers). Orientation only really requires a basis descriptor (we can generate the FieldPattern from a descriptor - a separate improvement needed).  Since DOFManagers, are expensive and memory intensive to construct, if the user has one available, we should use it in the orientation computation if it makes sense (has the right bases registered). Let's talk more offline next week about this. I'd like to discuss some extensions to DOFManager as well that would work here. We have the normal Thursday meeting, but if that is too long a delay, feel free to schedule something earlier in the week.

william76:
@seamill closed due to inactivity.

seamill:
That is fine, I'll reopen when I can get back to Panzer work.

PR #4231, opened 2019-01-21
Domi: Fix config bug with clean branch

wfspotz:
@trilinos/domi 

## Description
In `domi/src/CMakeLists.txt`, I needed to change `Domi_ENABLE_Tpetra` to `Domi_ENABLE_TpetraCore` to ensure that the `HAVE_TPETRA` macro was set properly in `Domi_config.h`.  This pull request is similar to #3731 except is based on a clean feature branch that cherry-picked the pertinent commits and ignored the commits in #3731 from an accidental merge.

## Motivation and Context
This manifest in PyTrilinos, where Tpetra for `PyTrilinos.Domi` was enabled, but not Tpetra for Domi.

## Related Issues

* Addresses #3618, #3687
* Replaces #3731 

## How Has This Been Tested?
I tested this by running the Domi tests and the PyTrilinos tests. For PyTrilinos, I had to disable ML in order to work around #3456.  The error really doesn't show up in Domi tests: Tpetra was always improperly disabled and all tests passed.

## Checklist


wfspotz:
@mhoemmen The compile error here is

    undefined reference to `Tpetra::MultiVector<long long, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >::getNumVectors() const'

when compiling `MDVector_UnitTests.cpp`.  The compilation error is protected by `#ifdef HAVE_TPETRA`, and it wouldn't crash if `HAVE_TPETRA` wasn't defined.  I also checked `Domi_config.h` and it has `#define HAVE_TPETRA` in it.

This line compiles just fine on my MacBook Pro.  Any ideas what might be the problem on the testing platform?

wfspotz:
@mhoemmen Yes, those other places are pure Domi.  The tests that failed to build were calling the Tpetra converters.

We could try to apply the same ETI rules to Domi (actually, apply ETI _to_ Domi) to make the two packages more similar that way, but that would be a separate issue.

Issue #4211, opened 2019-01-17
Zoltan2 coordinateModelPartBox should store doubles, not scalar_t


kddevin:
@trilinos/zoltan2 

## Expectations 
coordinateModelPartBox is used for storing bounding box info for the parts in a geometric partition.  These boxes can be searched, then, for pointAssign and boxAssign functionality for, say, contact detection or particle methods.  One would expect the union of the boxes to cover the entire geometric domain given to the partitioner.  

## Current Behavior

Currently, coordinateModelPartBox stores bounding box info using the user's scalar_t.  However, if the user provides scalar_t = int, the box boundaries could be truncated so that a complete cover of the domain is not achieved.  Plus, there likely will be compiler issues in having scalar_t = int that would have to be resolved.

## Motivation and Context
Trying to use scalar_t=int with geometric partitioning causes problems in MJ and in use of the coordinateModelPartBox.  For example, one might give scalar_t=int to partition the (i,j) indices of a matrix.  

In truth, there is no reason to use the users' scalar_t in coordinateModelPartBox.  The data type is a coordinate, which we can define to be double.  We might allow the constructor to be templated on other types, but we should store doubles.  Users should not specify the type of values that we are computing.  (See #1312)

## Definition of Done
Remove template parameters to coordinateModelPartBox; define coordinate type to be stored as double; update code and interface.

PR #1533, opened 2017-07-25
Zoltan2 directory

MicheldeMessieres:
Note this PR is in progress and not ready yet for merging - setting this up to create commenting opportunities.

In the current state the unit test for the Kokkos directory demonstrates working Add and Replace modes for variable sized arrays. Aggregate mode will be added soon.

To get this working I implemented some code for the usage case which was previously marked as unsupported: _Non-blocked, variable-sized recvs not supported_. Some critical areas for discussion are marked with the text NEW METHOD. Even though this works I suspect this may not cover all usage cases or be the best approach and would like to discuss that further. In particular I have the do_wait method filling the receive buffer out of order (because the directory handles resorting anyways) but this is not consistent with the original implementation.

PR #4232, opened 2019-01-21
Zoltan2:  Enable MJ to work with int coordinates

### LINKED ISSUE #1312: Zoltan2 partitioning metrics return scalar_t; should return double
Zoltan2's partitioning metrics class EvaluatePartition returns, among other things, imbalance values.  Imbalance is inherently a float or double.  While scalar_t is often float or double, it could be int.  When scalar_t is int, returning imbalance as an int is inaccurate. 

It appears that currently, all metrics are stored as scalar_t.  We could store them as double instead.
@trilinos/zoltan2

### LINKED ISSUE #4212: Zoltan2:  MJ does not work with integer coordinates
@trilinos/zoltan2   @MicheldeMessieres 

## Expectations
MJ should function when users provide scalar_t = int coordinates (e.g., (i.,j) indices of matrix entries) to be partitioned.

## Current Behavior

MJ requires scalar_t = float or double.  MJ uses the users' scalar_t as the data type for cuts and ratios.  When scalar_t = int, MJ segfaults with division by zero or hangs, due to truncation of differences of coordinates, etc.  

## Possible Solution

Possible solutions:
-  throw an error if users provide scalar_t = int; require users to copy their int data to float or double before calling MJ
-  allow users to provide scalar_t = int, but define an internal coordinate type as double and use that throughout the code; copy the user's int coordinates into the internal coordinate type and proceed
-  do a massive overhall of MJ to use scalar_t in some places and internal coordinate type where needed.  

I prefer the second solution over the first, but it will require more work; it will likely require changes related to #4211 as well.  The third solution would cause many code diffs, which would complicated the eventual merge of @MicheldeMessieres' work on GPU-enabled MJ.

## Steps to Reproduce

A test is already written:   zoltan2/test/partition/mj_int_coordinates.cpp.

## Related Issues

#4211 #4207

kddevin:
@trilinos/zoltan2 

## Description
Enables MJ partitioner to work with integer coordinates:  #4212 
- MJ requires floating point calculations for determining cuts
- MJ previously used the User's scalar_t for cuts, which was problematic when scalar_t was int
- Fix defines mj_scalar_t to be the user's scalar_t if it is a floating point type; otherwise, mj_scalar_t is float
- If mj_scalar_t differs from the user's scalar_t, coordinates are copied to mj_scalar_t arrays.

Also, removed templated types from coordinateModelPartBoxes.  #4211 
- coordinateModelPartBoxes store cut information computed by Zoltan2; their storage should not depend on the user's data types
-  changed storage in coordinateModelPartBoxes to be double
-  these changes were needed to enable fixing #4212 
-  removing the template parameters breaks backward compatibility

## Motivation and Context

I want to partition integer coordinates.

## Related Issues

* Closes #4212, #4211
* Still need to fix #1312 (as for scalar_t = int, it always returns imbalance = 1), but waiting until #1533 is merged to avoid conflicts 

## How Has This Been Tested?

new test zoltan2/test/partition/mj_int_coordinates.cpp
mac and linux

## Checklist

- [ x] My commit messages mention the appropriate GitHub issue numbers.
- [ x] These changes break backwards compatibility.:  removing template parameters breaks backward compatibility of functions returning coordinateModelPartBox

### DIFF
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -99,9 +99,9 @@
 #define ZOLTAN2_ABS(x) ((x) >= 0 ? (x) : -(x))
 //imbalance calculation. Wreal / Wexpected - 1
 #define imbalanceOf(Wachieved, totalW, expectedRatio) \
-        double(Wachieved) / double((totalW) * (expectedRatio)) - 1
+        (Wachieved) / ((totalW) * (expectedRatio)) - 1
 #define imbalanceOf2(Wachieved, wExpected) \
-        double(Wachieved) / double(wExpected) - 1
+        (Wachieved) / (wExpected) - 1
 
 
 #define ZOLTAN2_ALGMULTIJAGGED_SWAP(a,b,temp) temp=(a);(a)=(b);(b)=temp;
@@ -6626,7 +6626,8 @@ public:
                         minimum_migration_imbalance(0.30),
                         mj_keep_part_boxes(false), num_threads(1), mj_run_as_rcb(false),mj_premigration_option(0), min_coord_per_rank_for_premigration(32000),
                         comXAdj_(), comAdj_(), coordinate_ArrayRCP_holder (NULL)
-    {}
+    {std::cout << "KDDKDD adapter_scalar_t " << typeid(adapter_scalar_t).name() << " mj_scalar_t " << typeid(mj_scalar_t).name() << std::endl;}
+
     ~Zoltan2_AlgMJ(){
       if (coordinate_ArrayRCP_holder != NULL){
         delete [] this->coordinate_ArrayRCP_holder;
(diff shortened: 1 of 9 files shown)


packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -99,9 +99,9 @@
...
 //imbalance calculation. Wreal / Wexpected - 1
 #define imbalanceOf(Wachieved, totalW, expectedRatio) \
-        double(Wachieved) / double((totalW) * (expectedRatio)) - 1
+        (Wachieved) / ((totalW) * (expectedRatio)) - 1
 #define imbalanceOf2(Wachieved, wExpected) \
-        double(Wachieved) / double(wExpected) - 1
+        (Wachieved) / (wExpected) - 1
 
 
 #define ZOLTAN2_ALGMULTIJAGGED_SWAP(a,b,temp) temp=(a);(a)=(b);(b)=temp;
```
mhoemmen: Please just use `std::swap` here.  If `a` or `b` are expressions with side effects, this macro will invoke the side effects multiple times.

packages/zoltan2/src/util/Zoltan2_StridedData.hpp
```diff
@@ -158,6 +158,44 @@ class StridedData {
...
+// because cannot specialize member function of templated class.
+template<typename scalar_t, typename T>
+static void getInputArrayHelper(
+  ArrayRCP<const T> &target,
+  const ArrayRCP<const scalar_t> &src)
+{
+  // Create a copy of desired type T
+  // From logic in getInputArray, we know stride == 1.
+  size_t n = src.size();
+  T *tmp = new T [n];
```
mhoemmen: You could just wrap `tmp` in an `ArrayRCP` from the beginning, using the `ArrayRCP` constructor that takes a raw pointer and `true` (for "owning")

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: I'll take care of this when I merge some ongoing MJ work.

PR #4233, opened 2019-01-22
Tpetra,Teuchos: Fix #4225

### LINKED ISSUE #4225: Tpetra: ImportExport2_UnitTest & FixedHashTable tests fail in debug mode; Teuchos::ArrayView ctor issue
@trilinos/tpetra @trilinos/teuchos @bartlettroscoe @ibaned 

Some Tpetra tests have been failing in debug mode.  For example, FixedHashTable's test fails.  `Teuchos::ArrayView`'s constructor that takes a pointer and a length throws, because the length is zero but the pointer is nonnull.  When I stop `ArrayView`'s constructor from throwing, the test passes just fine.

This is a bit of a problem, because `Kokkos::View` permits a View with a zero length to have a nonnull pointer.  I can see the value of the test, but it is a bit troublesome to have to do an extra test to convert from `Kokkos::View` to `Teuchos::ArrayView`.  Furthermore, `std::vector` behaves just like `Kokkos::View`, in that `.data()` need not return `nullptr` if the vector's length is zero.

My plan was to go through Tpetra, find all instances of that constructor, and fix them.  I plan to modify the relevant conversion functions in the TeuchosKokkosCompat subpackage, and only use those.  However, it seems like this could lead to trouble when attempting obvious conversions from `std::vector` to `Teuchos::ArrayView`.  Should we consider removing this debug test?

See e.g., https://github.com/trilinos/Trilinos/issues/184 .

## Expectations

Conversions from zero-length `std::vector` or `Kokkos::View` to `Teuchos::ArrayView` should not require an extra branch just to make the pointer null if the length is zero.

mhoemmen:
@trilinos/tpetra @trilinos/teuchos

Tpetra's ImportExport2_UnitTests and FixedHashTable tests now pass.
Tpetra::Details::getArrayViewFromDualView and a TeuchosKokkosCompat
function now uses the work-around required by Teuchos::ArrayView's
constructor in a debug build.  ImportExport2_UnitTests no longer
redefines that function.

I also added TPETRA_VERBOSE output to CrsMatrix communication
functions and CrsMatrix's transferAndFillComplete, and replaced a
HAVE_TPETRA_DEBUG region in transferAndFillComplete with a run-time -
controlled debug mode check.

## Related Issues

* Closes #4225

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_packCrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_packCrsMatrix_def.hpp
@@ -45,6 +45,7 @@
 #include "TpetraCore_config.h"
 #include "Teuchos_Array.hpp"
 #include "Teuchos_ArrayView.hpp"
+#include "Tpetra_Details_Behavior.hpp"
 #include "Tpetra_Details_castAwayConstDualView.hpp"
 #include "Tpetra_Details_createMirrorView.hpp"
 #include "Tpetra_Details_getEntryOnHost.hpp"
@@ -52,6 +53,8 @@
 #include "Tpetra_Details_PackTraits.hpp"
 #include "Tpetra_CrsMatrix_decl.hpp"
 #include <memory>
+#include <sstream>
+#include <stdexcept>
 #include <string>
 
 /// \file Tpetra_Details_packCrsMatrix.hpp
@@ -1043,6 +1046,29 @@ packCrsMatrixWithOwningPIDs (const CrsMatrix<ST, LO, GO, NT>& sourceMatrix,
 
   typename local_matrix_type::device_type outputDevice;
 
+  const bool verbose = ::Tpetra::Details::Behavior::verbose ();
+  std::unique_ptr<std::string> prefix;
+  if (verbose) {
+    const int myRank = [&] () {
+      auto map = sourceMatrix.getMap ();
+      if (map.get () == nullptr) {
+        return -1;
+      }
+      auto comm = map->getComm ();
+      if (comm.get () == nullptr) {
+        return -2;
+      }
+      return comm->getRank ();
+    } ();
+    std::ostringstream os;
+    os << "Proc " << myRank << ": packCrsMatrixWithOwningPIDs: ";
+    prefix = std::unique_ptr<std::string> (new std::string (os.str ()));
+
+    std::ostringstream os2;
+    os2 << *prefix << "start" << std::endl;
+    std::cerr << os2.str ();
+  }
+
   // Convert all Teuchos::Array to Kokkos::View
 
   // This is an output array, so we don't have to copy to device here.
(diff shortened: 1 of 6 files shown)


packages/tpetra/core/src/Tpetra_Details_packCrsMatrix_def.hpp
```diff
@@ -1068,15 +1094,62 @@ packCrsMatrixWithOwningPIDs (const CrsMatrix<ST, LO, GO, NT>& sourceMatrix,
                                             sourcePIDs.size (), true,
                                             "export_pids");
   constexpr bool pack_pids = true;
-  PackCrsMatrixImpl::packCrsMatrix(
-      sourceMatrix, exports_dv, num_packets_per_lid_d, export_lids_d,
-      export_pids_d, constantNumPackets, pack_pids, distor);
+  try {
```
bartlettroscoe: @mhoemmen, is a try/catch really needed here?  Is the function `packCrsMatrixWithOwningPIDs()` not allowed to throw?  It seems like this might be a poor usage of exception handling (but I would have to know more about this code).

mhoemmen: @bartlettroscoe This is the debug-mode branch of the code.  Normal release-mode execution doesn't take this code path.  (Tpetra's "debug mode" is enabled by default in a debug build and disabled by default in a release build, but users can control this at run time with the `TPETRA_DEBUG` environment variable.)

The issue is with `Teuchos::ArrayView` throwing ;-) and it's really hard to find all possible uses of such a fundamental class.  I'm just trying to be brutally conservative so that I can catch potential hangs in the code.  

In general, if a function like `packCrsMatrix` throws, the code would probably hang in an MPI run with multiple processes.

mhoemmen:
@bartlettroscoe No worries -- I appreciate your comments.  I'm not a huge fan of Tpetra's extreme debug-mode checking -- it's hard to maintain and can easily go stale -- but it has saved us a number of times when we otherwise would have had a miserable experience with a parallel debugger.  Debuggers WILL betray you at the worst time, but good old `printf` (usually) never lets you down :-D

mhoemmen:
Thanks @tjfulle ! :D

rppawlo:
Stupid question - Is ArrayView being used in tpetra just for backwards compatibility? If we have access to the underlying kokkos views should we be looking to phase out ArrayView use in tpetra in favor of kokkos view interfaces? It will probably be a lot of work on downstream packages, but just wondering.

packages/tpetra/core/src/Tpetra_Details_packCrsMatrix_def.hpp
bartlettroscoe: @mhoemmen, if this is a common idiom in Tpetra, then perhpas should be a macro that encapsulates this and removes a lot of this duplication and boilerplate code.   You can put hooks into macros calls that allow stopping a debugger.  Just look at `TEUCHOS_TEST_FOR_EXCEPTION()` functions.

mhoemmen:
@rppawlo wrote:
> Is ArrayView being used in tpetra just for backwards compatibility?

1. Trilinos hasn't had a release in three years, so we have no policy on breaking backwards compatibility,  other than "is this urgent enough to justify making >= 4 apps angry?".
2. `Teuchos::ArrayView` fulfills a useful role that won't exist in the Standard until C++20, with `std::span`.

It should be expected that low-level memory management needs evolve over time and that a library supporting these needs is transitional.  I spent a lot of time over the past year or so  revising a proposal for the C++ Standard Library that would move towards making `Kokkos::View` unnecessary.

mhoemmen:
I am delaying a tiny bit just to make sure that Teuchos::ArrayView's constructor is the reason for the test failures.

bartlettroscoe:
@mhoemmen said:

> `Teuchos::ArrayView` fulfills a useful role that won't exist in the Standard until C++20, with `std::span`.

I have not been keeping up with future C++ standards so it is exciting to see them implement something like `Teuchos::ArrayView`.  Now we can just hope that Clang and/or GCC implement optional debug-mode runtime checking with the types of debug-mode runtime checking as exists for the Teuchos Memory Management classes to catch things like dangling references (like when created from `std::vector` or from checked iterators).  But given the lack of optional debug-mode runtime checking for `std::shared_ptr`, I am not hopeful.

mhoemmen:
@bartlettroscoe Seriously, the best way to do that would be to start talking to Standard Library implementers.  I know they have some checks like that, but if you would like to write up an experience report on the value of checks that most vendors don't implement, and/or suggest changes to the Standard that would make debug-mode checks easier, we would be happy to show it to the right people.

mhoemmen:
I verified that commenting out the check in two of ArrayView's constructors makes Tpetra's tests pass (except for outstanding, unrelated issues with complex data types).  This suggests that it's safe to merge this PR.

mhoemmen:
@bartlettroscoe wrote:
> Looks like they missed an opportunity with `std::span`.

The debugging situation isn't hopeless.  You'll notice that `span::iterator` and `span::const_iterator` have [implementation-defined types](http://eel.is/c++draft/span.overview).  Standard Library implementers already exploit iterator type freedom to add debugging features to other classes.  The main issue for you would be the lack of provided conversions from `vector` or `*_ptr<T[]>` to `span`, that would permit flow of tracking information.  This could be an opportunity for you to contribute to the Standard :-) .

In general, people argued about whether `span` should be extensible, or a simple fundamental concept.  Hopefully the debate got documented well in the papers and minutes -- send me an e-mail if you'd like to learn more.  Folks have been more open to the idea of `mdspan` having extensibility.

PR #3317, opened 2018-08-17
Tpetra:  Remove Unused Parameter Warnings

jmgate:
@trilinos/tpetra 

## Description
Building Trilinos with gcc-7.2.0 with all warning on throws a ton of unused parameter warnings.  Commenting out the parameter name in the function definition does away with them.

## Motivation and Context
Just working toward a clean build.

PR #4239, opened 2019-01-22
Tpetra:  Remove Unused Parameter Warnings

jmgate:
@trilinos/tpetra 

## Description
Comment out parameter names in function definitions to avoid unused parameter warnings.  In cases where their use is hidden behind an `#ifdef`, `(void)` the parameters in the `#else` block.

This is a second attempt at #3317&mdash;for some reason GitHub wouldn't let me simply reopen that one after a force push to the branch.

## Motivation and Context
Just hoping for a clean build.

## How Has This Been Tested?
This works on my RHEL7 machine with gcc-7.3.0.


jmgate:
@mhoemmen, this is ready for a re-approval.

jmgate:
BTW, this reduces the Tpetra warning count on my build by ~3000.

mhoemmen:
@kddevin Let's talk about this today -- thanks!

mhoemmen:
We decided in the @trilinos/tpetra meeting that we want this now -- thanks!!!

Issue #4237, opened 2019-01-22
kokkos-kernels: now requires fortran to configure


rppawlo:
A change to kokkos-kernels last week now makes the package require fortran. This has brought down testing suite for Drekar that use the type2 stack but does not enable fortran. We are trying to avoid requiring users to have to use the kitware hacked ninja to get fortran support. Was this dependency change intended?

rppawlo:
@mhoemmen - #4228 also broke all empire builds since they disable fortran as well.

@jmgate

jmgate:
Yup.  If a fix is imminent, that'd be great.  Otherwise I'll look into backing out that merge commit in our fork.

mhoemmen:
Huh, well, I guess PR testing doesn't exercise this case.  Let me revert the changes for now.  It should be easy for me to fix.

rppawlo:
Thanks @mhoemmen !

Issue #4235, opened 2019-01-22
Albany builld broken due to missing KokkosKernels_FortranMangling.h header


ikalash:
Here is the error:
```
/home/ikalash/nightlyAlbanyTests/Results/Trilinos/build/install/include/KokkosBlas1_dot_tpl_spec_decl.hpp:50:10: fatal error: KokkosKernels_FortranMangling.h: No such file or directory
```
You can see more details on the Albany dashboard, e.g., http://cdash.sandia.gov/CDash-2-3-0/viewBuildError.php?buildid=80609

ikalash:
@trilinos/kokkos-kernels

spdomin:
Same with Nalu.. 

I opened an issue under: https://github.com/kokkos/kokkos-kernels/issues/372

rppawlo:
also related: #4237

mhoemmen:
Huh, well, I guess PR testing doesn't exercise this case. Let me revert the changes for now. It should be easy for me to fix.

mhoemmen:
Other kokkos-kernels developers: These changes were a patch to Trilinos. I was about to add them to the kokkos-kernels repo as well, but these build errors show that we also need to test with Fortran disabled. Thus, these changes are not in the kokkos-kernels repo; they are only in the Trilinos repo, in develop. I have a Trilinos PR open (#4241) to revert the changes.

I think I know what to do to fix this (extra header file with "if Fortran enabled" macro to decide whether to include the Fortran mangling header file that CMake generates).

ikalash:
Just to follow up: this issue has been resolved within Albany.

mhoemmen:
Thanks @ikalash ! :D

PR #4228, opened 2019-01-21
kokkos-kernels: Fix BLAS mangling

mhoemmen:
@trilinos/tpetra @kyungjoo-kim 

## Description

Patch kokkos-kernels to fix BLAS mangling.  We'll also add these changes to kokkos-kernels.

## Motivation and Context

Sierra needs Trilinos to build with XL.

## Related Issues

* Related to https://github.com/kokkos/kokkos-kernels/issues/369

Issue #4238, opened 2019-01-22
Disable Fortran in one of the Trilinos PR builds


bartlettroscoe:
CC: @trilinos/framework, @mhoemmen, @rppawlo, @bathmatt, @jmgate, @fryeguy52 

There are several projects that disable Fortran when building Trilinos.  Currently, the EMPIRE ATDM Trilinos configuration disables Fortran.  (But that will chance once we fully merge the SPARC and EMPIRE ATDM Trilinos configurations since the SPARC configuration enables Fortran.)  The Drekar developers disable Fortran in their configurations (which is not likely to change).  Also, many people who build on Macs disable Fortran (since Macs don't come with a Fortran compiler by default).  

Therefore, it is important for Trilinos to support a Fortran-less configuration and build.  Therefore, at least one of the Trilinos PR builds needs to set `Trilinos_ENABLE_Fortran=OFF`.

bartlettroscoe:
CC: @fryeguy52 

@mhoemmen, I added this since it looks like PR #4228 that you merged yesterday broke all of the EMPIRE ATDM Trilinos builds shown [here](https://testing.sandia.gov/cdash/index.php?project=Trilinos&date=2019-01-22&filtercount=1&showfilters=1&field1=buildname&compare1=65&value1=Trilinos-atdm-).

bartlettroscoe:
@trilinos/framework, it seems like a good candidate Trilinos PR build to set `Trilinos_ENABLE_Fortran=OFF` would be the build `Trilinos_pullrequest_gcc_4.9.3_SERIAL`.  It seems like someone who would want to an MPI-less build of Trilinos might also want to not have to have a working version of Fortran either.

mhoemmen:
Thanks @bartlettroscoe ; this would be super helpful :D

mhoemmen:
> It seems like someone who would want to an MPI-less build of Trilinos might also want to not have to have a working version of Fortran either.

That's a good idea.  Most of the Fortran issues relate to TPLs and binding to non-MPI functions.

jwillenbring:
@bartlettroscoe @mhoemmen 

This seems like a reasonable thing to do. It seems unlikely that we would have many cases where a serial no fortran build would pass, when a serial build with fortran would fail. I have fired off a build to help us look at the current situation more closely.

bartlettroscoe:
> It seems unlikely that we would have many cases where a serial no fortran build would pass, when a serial build with fortran would fail.

[...]

PR #4241, opened 2019-01-22
Revert PR #4228

mhoemmen:
This reverts commit 24c2eb2f9c622f856a0477a0ef8751b4145ca7b1, reversing
changes made to 7b60b19b90f7cc1066da428b7d8ad18813675a87.

It turns out that this breaks builds if Fortran is not enabled.  CMake
apparently permits the configuration to proceed, but the build fails
with a missing header file.  This revert should fix the following issues:

https://github.com/kokkos/kokkos-kernels/issues/372
https://github.com/trilinos/Trilinos/issues/4237
https://github.com/trilinos/Trilinos/issues/4235

@trilinos/kokkos-kernels 

## Related Issues

* Closes https://github.com/kokkos/kokkos-kernels/issues/372, #4235, #4237

### DIFF
--- a/packages/kokkos-kernels/Makefile.kokkos-kernels
+++ b/packages/kokkos-kernels/Makefile.kokkos-kernels
@@ -315,46 +315,6 @@ ifneq ($(KOKKOSKERNELS_INTERNAL_NEW_CONFIG), 0)
   tmp := $(shell cp KokkosKernels_config.tmp KokkosKernels_config.h)
 endif
 
-#==== User-settable options for Fortran mangling macros  =================================================
-
-KOKKOSKERNELS_FORTRAN_GLOBAL ?= name\#\#_
-KOKKOSKERNELS_FORTRAN_GLOBAL_ ?= name\#\#_
-KOKKOSKERNELS_FORTRAN_MODULE ?= __\#\#mod_name\#\#_MOD_\#\#name
-KOKKOSKERNELS_FORTRAN_MODULE_ ?= __\#\#mod_name\#\#_MOD_\#\#name
-
-#==== Create header file for Fortran mangling macros  =================================================
-
-$(shell echo "#ifndef KOKKOSKERNELS_FORTRAN_HEADER_INCLUDED" > KokkosKernels_FortranMangling.tmp)
-$(shell echo "#define KOKKOSKERNELS_FORTRAN_HEADER_INCLUDED" >> KokkosKernels_FortranMangling.tmp)
-$(shell echo "" >> KokkosKernels_FortranMangling.tmp)
-$(shell echo "/* Mangling for Fortran global symbols without underscores. */" >> KokkosKernels_FortranMangling.tmp)
-$(shell echo "#define KOKKOSKERNELS_FORTRAN_GLOBAL(name,NAME) $(KOKKOSKERNELS_FORTRAN_GLOBAL)" >> KokkosKernels_FortranMangling.tmp)
-$(shell echo "" >> KokkosKernels_FortranMangling.tmp)
-$(shell echo "/* Mangling for Fortran global symbols with underscores. */" >> KokkosKernels_FortranMangling.tmp)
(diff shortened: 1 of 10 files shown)


mhoemmen:
Thanks @jmgate !  I'm astonished that CMake didn't give a configure-time error for this case.

rppawlo:
@mhoemmen - a quick comment - the drekar builds were breaking during configure with a good message if Trilinos_ENABLE_Fortran=OFF was in the configure script. Not sure how it is getting further in the other app builds:
```
=======================
KokkosKernels ETI Types
   Devices:  <OpenMP,HostSpace>;<Serial,HostSpace>
   Scalars:  double
   Ordinals: int
   Offsets:  int;size_t

KokkosKernels TPLs
   BLAS
=======================

CMake Error at /ascldap/users/rppawlo/tpls/cmake/3.12.3/share/cmake-3.12/Modules/FortranCInterface.cmake:205 (message):
  FortranCInterface requires the Fortran language to be enabled.
  packages/kokkos-kernels/src/CMakeLists.txt:1 (INCLUDE)


-- Configuring incomplete, errors occurred!
```

ibaned:
Hmm it doesn't look like we do disable Fortran in the Albany build, but the mangling header is still not found

mayrmt:
I disable Fortran to be able to use the non-Kitware-modified ninja.

mhoemmen:
@ibaned wrote:
> Hmm it doesn't look like we do disable Fortran in the Albany build, but the mangling header is still not found

Do you get any interesting configure-time messages from CMake about this?  I was hoping to fix this issue by protecting inclusion of the header file with an "if Fortran enabled" macro, but if enabling Fortran doesn't fix this, then I'll need to think of another solution.

ibaned:
I do see lines that suggest the mangling scheme is being discovered. Maybe its as simple as forgetting to call CMake's `install()` on the generated header file?

bartlettroscoe:
@mhoemmen, @rppawlo, and others,

FYI: I created #4238 to request that the @trilinos/framework team change one of the PR builds to have `Trilinos_ENABLE_Fortran=OFF` so that we can catch issues like this before the merge to 'develop'.  Please comment in #4238.

ibaned:
@bartlettroscoe although this will catch CMake checks inside Trilinos, it can't detect a non-installed header

bartlettroscoe:
> @bartlettroscoe although this will catch CMake checks inside Trilinos, it can't detect a non-installed header

@ibaned, right, but the ATDM Trilinos builds all failed at configure time so at least we would have caught that.

We will need to add automated installation testing to one of the PR builds at some point.  That is very do-able if we brush off the TriBITS feature for installation testing.

rppawlo:
> I'm curious -- is there a particular reason y'all disable Fortran? Of course Trilinos needs to work with Fortran disabled -- I'm just wondering if you have a specific reason beyond "we don't need it."

ninja is critical for productivity - especially on cuda. Most of the app teams are now using it exclusively. I don't think it is reasonable to require every trilinos user to  have to use the hacked kitware version of ninja if they use Trilinos. It's also a pain on mac since it doesn't come with a stock fortran compiler. Maybe that will change with flang(?) in the future.

bartlettroscoe:
> hacked kitware version of ninja

I think the term "extended" version of ninja is more accurate :-)  If Google would give a damn about Fortran this would have been in the official ninja version years ago.  This says more about Fortran than about Google I suspect.

> Maybe that will change with flang(?) in the future.

Just found out at the ECP meeting that they are starting from scratch again with a new LLVM Fortran compiler called 'F18' which is planned to  become the new Flang in the future.  I am hearing this may be the last Fortran front-end ever written.

mhoemmen:
@rppawlo , cool, Ninja is a plenty good reason :-D  

@ibaned wrote:
> I do see lines that suggest the mangling scheme is being discovered. Maybe its as simple as forgetting to call CMake's `install()` on the generated header file?

Hm, that could be, if the generated header goes into a directory that doesn't get glob-installed.  I'll check.  Thanks!

bartlettroscoe:
Might not be a bad idea to glob and install all header files in the build directory as a general idiom (but Kitware would disagree but this is what happens if you don't glob).

Issue #3541, opened 2018-10-02
ShyLU_DD tests build failure in targeted CUDA PR bulid Trilinos-atdm-white-ride-cuda-9.2-release-debug-pt starting 10/2/2018


bartlettroscoe:
CC: @trilinos/shylu, @srajama1 (Trilinos Linear Solvers Product Area Lead), @fryeguy52, @roeverf, @searhein 

## Description

Starting today, there are two build errors for the ShyLU_DD package in the build `Trilinos-atdm-white-ride-cuda-9.2-release-debug-pt ` on 'white' and 'ride' as shown [here](https://testing.sandia.gov/cdash-dev-view/viewBuildError.php?buildid=3998259) which shows build errors starting with:

```
/home/jenkins/white/workspace/Trilinos-atdm-white-ride-cuda-9.2-release-debug-pt/SRC_AND_BUILD/Trilinos/packages/shylu/shylu_dd/frosch/test/Thyra_Tpetra/main.cpp(104): error: "EpetraNode" is ambiguous
```

ShyLU_DD was building just fine yesterday in this build as shown [here](https://testing.sandia.gov/cdash-dev-view/viewBuildError.php?type=0&buildid=3995410).

Looking at the commits pulled today shown [here](https://testing.sandia.gov/cdash-dev-view/viewNotes.php?buildid=3998221#!#note6), it seems likely this might have been caused by one of the commits from @roeverf  to the ShyLU_DD package merged in the PR #3472 merged to 'develop' by @searhein on 10/1/2018 as shown [here](https://github.com/trilinos/Trilinos/pull/3472#event-1876847991).

This is an important build because we are targeting this build on 'white' and 'ride' as a Trilinos PR testing build (see #2464 ).

## Current Status on CDash

The current status of `ShyLU_DD` in this build and tests over the last few days can be seen in [this CDash query](https://testing.sandia.gov/cdash-dev-view/index.php?project=Trilinos&date=2019-01-22&filtercount=4&showfilters=1&filtercombine=and&field1=subprojects&compare1=93&value1=ShyLU_DD&field2=buildname&compare2=61&value2=Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-release-debug-pt&field3=site&compare3=61&value3=ride&field4=buildstarttime&compare4=83&value4=1%20week%20ago).

## Steps to reproduce

[...]

PR #4248, opened 2019-01-23
Another check for Tpetra_DefaultNode.

searhein:
@trilinos/shylu 

I have added another check for Tpetra_DefaultNode to resolve the EpetraNode errors reported in #3541.


srajama1:
Wow the automerge already did its job, still might be ok to check it with a tpetra developer. I will check the failing builds tomorrow morning.

searhein:
@srajama1 This PR contained only a very slight modification, i.e., a condition for enabling some of the tests in FROSch depending on Tpetra_DefaultNode. Therefore, it should not be critical.
Since introducing this condition already resolved part of the errors mentioned in #3541, I now also force it for the remaining Epetra tests. 
I enabled the automerge to be sure that the PR was merged right after approval. I hope this was ok.

Issue #4244, opened 2019-01-23
Zoltan2 test build errors in all ATDM Trilinos CUDA builds starting 1/22/2019


bartlettroscoe:
CC: @trilinos/zoltan2, @kddevin (Trilinos Data Services Product Lead), @bartlettroscoe, @fryeguy52

<Checklist>
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

The build of several Zoltan2 test object files and executables are broken in the ATDM Trilinos CUDA builds starting 1/22/2018 as shown in the builds:

* [Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-release-debug-pt](https://testing.sandia.gov/cdash-dev-view/index.php?project=Trilinos&parentid=4443057)
* [Trilinos-atdm-white-ride-cuda-10.0-gnu-7.4.0-debug-exp](https://testing.sandia.gov/cdash-dev-view/index.php?project=Trilinos&parentid=4445140)

(NOTE: It is likely broken in all of the other CUDA builds as well but we can't tell because of a configure failure in all of that ATDM Trilinos CUDA builds on 1/22/2019 as described in #4242.  The only reason that the build `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-release-debug-pt` did not fail at configure time was that it has Fortran enabled and the only reason that the build `Trilinos-atdm-white-ride-cuda-10.0-gnu-7.4.0-debug-exp` which disables Fortran did not have configure failures is that reverting PR #4238 was merged before this build was run.)

There were also 2 new failing Zoltan2 tests as shown [here](https://testing.sandia.gov/cdash-dev-view/viewTest.php?onlyfailed&buildid=4442262) for the tests:

* Zoltan2_directoryTest_Kokkos_MPI_4
* Zoltan2_directoryTest_KokkosSimple_MPI_4

and 33 not-run tests (as you would expect given the build failures).

The new failing tests showed CUDA errors and segfaults like shown for [Zoltan2_directoryTest_Kokkos_MPI_4](https://testing.sandia.gov/cdash-dev-view/testDetails.php?test=64892618&build=4442262) showing:

```
...
 Passed: Vector_User_Multiple_GID contiguous int mode: Aggregate
Testing totalIds: 1
  Passed: Single_User_Single_GID contiguous int mode: Replace
[...]

Issue #4246, opened 2019-01-23
Zoltan2 missing file leading to Albany build failures


ikalash:
Looks like Trilinos broke the Albany nightly builds again.  Here is the error:
```
/home/ikalash/nightlyAlbanyTests/Results/Trilinos/build/install/include/Zoltan2_Directory_Impl.hpp:62:10: fatal error: murmur3.c: No such file or directory
```
Please see our dashboard for more info, e.g., http://cdash.sandia.gov/CDash-2-3-0/viewBuildError.php?buildid=80648 .

ikalash:
@trilinos/zoltan2

mhoemmen:
Looks like this broke ATDM too: https://github.com/trilinos/Trilinos/issues/4244

ikalash:
This has been resolved in Albany - thank you!

PR #4252, opened 2019-01-23
Fix CUDA and Albany builds from prior PR #1533

MicheldeMessieres:
@trilinos/zoltan2

## Description
Repairs issues from PR #1533 for directory work.
Resolved CUDA tests and Albany link failure. 

## Motivation and Context
Restore CUDA builds and Albany build.

## How Has This Been Tested?
On white:
 source $TRILINOS_DIR/cmake/std/atdm/load-env.sh cuda-9.2-gnu-7.2.0-release-debug
$ cmake \
  -GNinja \
  -DTrilinos_ENABLE_TESTS=ON -DTrilinos_ENABLE_Zoltan2=ON \
  $TRILINOS_DIR
$ bsub -x -Is -q rhel7F -n 16 ctest -j16

On clang parallel and serial.

## Checklist

- [x] No new compiler warnings were introduced. Except CUDA.

### DIFF
--- a/packages/zoltan2/src/directory/Zoltan2_Directory.hpp
+++ b/packages/zoltan2/src/directory/Zoltan2_Directory.hpp
@@ -58,7 +58,8 @@
 #include <mpi.h>
 #endif
 
-#include <Kokkos_UnorderedMap.hpp> // unordered map stores the local nodes
+// #include <Kokkos_UnorderedMap.hpp> // unordered map stores the local nodes
+#include <unordered_map>
 
 namespace Zoltan2 {
 
@@ -194,11 +195,21 @@ class Zoltan2_Directory {
 
       // fill
       size_t cnt = 0;
+
+      for(auto itr = node_map.begin(); itr != node_map.end(); ++itr) {
+        if(itr->second.free == 0) {
+          local_gids[cnt++] = itr->first;
+        }
+      }
+
+      // TODO: Restore Kokkos unordered_map
+      /*
       for(size_t n = 0; n < node_map.capacity(); ++n) {
         if(node_map.value_at(n).free == 0) {
           local_gids[cnt++] = node_map.key_at(n);
         }
       }
+      */
 
       if(cnt != node_map.size()) {
         throw std::logic_error("Unexpected counts. Internal error with the"
@@ -250,8 +261,10 @@ class Zoltan2_Directory {
 
     // originally the nodes are stored in a hash but in the new Kokkos mode
     // they are stored using Kokkos::UnorderedMap
-    typedef Kokkos::UnorderedMap<gid_t,
+    typedef std::unordered_map<gid_t,
       Zoltan2_Directory_Node<gid_t,lid_t,user_t>> node_map_t;
+//    typedef Kokkos::UnorderedMap<gid_t,
+//      Zoltan2_Directory_Node<gid_t,lid_t,user_t>> node_map_t;
     node_map_t node_map;
 
     // this method exists so constructor and copy constructor don't duplicate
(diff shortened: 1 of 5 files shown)


MicheldeMessieres:
This PR will resolve #4244 and #4246

packages/zoltan2/src/directory/Zoltan2_Directory.hpp
```diff
@@ -194,11 +195,21 @@ class Zoltan2_Directory {
...
       // fill
       size_t cnt = 0;
+
+      for(auto itr = node_map.begin(); itr != node_map.end(); ++itr) {
+        if(itr->second.free == 0) {
+          local_gids[cnt++] = itr->first;
+        }
+      }
+
+      // TODO: Restore Kokkos unordered_map
```
mhoemmen: You could have just used a HostSpace version of `Kokkos::UnorderedMap`, but just using a standard library class should be fine too.

packages/zoltan2/src/directory/Zoltan2_Directory.hpp
```diff
@@ -250,8 +261,10 @@ class Zoltan2_Directory {
 
     // originally the nodes are stored in a hash but in the new Kokkos mode
```
mhoemmen: Comment doesn't apply any more, but I want this fix to get in, so OK.

packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
```diff
@@ -50,16 +50,52 @@
...
-// otherwise. All this should eventually go away and either use Kokkos methods
-// or implement the hash here so we don't have the links to zoltan
-static inline void MurmurHash3_x86_32 (
-  const void * key, int len, uint32_t seed, void * out);
-#include "murmur3.c"
+// Copied from murmurc.3
+// TODO: Will be removed as Kokkos form develops
+#define ZOLTAN2_ROTL32(x,r) (uint32_t) \
+  (((uint32_t)(x) << (int8_t)(r)) | ((uint32_t)(x) >> (32 - (int8_t)(r))))
+static inline void Zoltan2_MurmurHash3_x86_32 ( const void * key, int len,
```
mhoemmen: The idiomatic way to have a `static` (internal linkage) function in C++ is to put it in an anonymous namespace, like this:
```c++
namespace { // (anonymous)
  void foo () {
    whatever_code ();
  }
} // namespace (anonymous)
```

packages/zoltan2/test/directory/CMakeLists.txt
```diff
@@ -16,9 +16,11 @@ TRIBITS_ADD_EXECUTABLE_AND_TEST(
...
 # result of findUniqueGids but using the new
 # Zoltan2Directory class.
-TRIBITS_ADD_EXECUTABLE_AND_TEST(
-  directoryTest_findUniqueGids.cpp
-  SOURCES directoryTest_findUniqueGids.cpp
-  COMM serial mpi
-  FAIL_REGULAR_EXPRESSION "FAIL"
-)
+
+# Temporarily disabling this to resolve cuda build issues.
```
mhoemmen: Shouldn't the above changes have fixed the build errors?

packages/zoltan2/src/directory/Zoltan2_Directory.hpp
MicheldeMessieres: @mhoemmen Thanks for reviewing this. Did not think of doing the host space which would have been easier. We are planning to do further development to use kokkos so eventually I'll be converting this back once I resolve some std::vector issues.

packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
MicheldeMessieres: I will fix this.

packages/zoltan2/test/directory/CMakeLists.txt
MicheldeMessieres: This was just one of the issues - but was a new test with some complications so I just disabled. However if i use your suggestion of Host space for the unordered map I should be able to restore this as well.

MicheldeMessieres:
@mhoemmen Thanks for the assistance on the review. I will make a follow up PR to implement your suggested changes.

mhoemmen:
@MicheldeMessieres Thanks!  Does the test actually build and pass with CUDA?

MicheldeMessieres:
> @MicheldeMessieres Thanks! Does the test actually build and pass with CUDA?

@mhoemmen To confirm tests were building and passing with CUDA which I tested on white. The PR fix is now merged.

MicheldeMessieres:
@kddevin @mhoemmen  I will set up another PR today to restore the new test we added and implement Mark's suggestions. Using the Kokkos unordered map with HostSpace also worked fine for the CUDA white build.

Issue #4249, opened 2019-01-23
Teuchos: Kokkos::Compat::persistingView must not call Teuchos::arcp with nonnull pointer and zero length


mhoemmen:
@trilinos/teuchos @trilinos/tpetra @trilinos/kokkos 

`Kokkos::Compat::persistingView` passes the `Kokkos::View`'s pointer directly to `Teuchos::arcp` (nonmember "constructor" for `Teuchos::ArrayRCP`).  `Kokkos::View`, just like `std::vector`, permits a nonnull pointer with a zero size.  `Teuchos::ArrayRCP` does not; its constructor throws when bounds checking is enabled (a configure-time option).  This causes some Tpetra tests to fail.

## Expectations

It should be harmless to convert a zero-size `Kokkos::View` to a `Teuchos::ArrayRCP`.

## Current Behavior

Some Tpetra tests fail, because `Teuchos::arcp` throws.  When I fix `persistingView` as suggested below, the tests pass.

## Possible Solution

Here are two possible solutions:

  1. Fix `persistingView` so that it gives `Teuchos::arcp` a null pointer if the `Kokkos::View`'s size is zero.
  2. Fix `Teuchos::arcp`, analogously to `Teuchos::ArrayView` (see #4234).

For now I prefer Solution (1).  See #4234 discussion, esp. the [comment here](https://github.com/trilinos/Trilinos/issues/4234#issuecomment-456920360). 

## Related Issues

* Related to #4234

Issue #4234, opened 2019-01-22
Teuchos::ArrayView constructor should set its pointer to null if input pointer is nonnull but length is zero


mhoemmen:
@trilinos/teuchos @trilinos/tpetra @kddevin @jhux2 @csiefer2 

See #4225 discussion.

## Expectations

It should be trivial to create a Teuchos::ArrayView from an std::vector or (host) Kokkos::View.  Users should not have to test whether the pointer is null if the length is zero.

## Related Issues

* Related to #4225

bartlettroscoe:
@mhoemmen, I guess I am not apposed to changing the pre-conditions and post-conditions for `Teuchos::ArrayView::ArrayView(T* ptr, Ordinal size)` to allow `ptr!=nullptr` when `size==0` and then to discard `ptr` in this case.  Technically that use case was invalid so if we make it valid, I don't think we can break backward compatibility (unless some code relies on the exception to be thrown but it only does so in a debug-mode build so that would be stupid code).

We would just need to update the unit tests to cover this case.  (Hopefully existing unit tests would fail if `Teuchos::ArrayView::ArrayView(T* ptr, Ordinal size)` where changed but no tests were updated.)

But we also need conversion functions (explicit or implicit) to and from `Teuchos::ArrayView` and `Kokkos::View`.   (That is the Teuchos way.  See "Figure 3. Conversions between array memory management types" in [Teuchos Memory Management Classes](https://trilinos.org/docs/dev/packages/teuchos/doc/html/TeuchosMemoryManagementSAND.pdf).)

Would there be any harm in just allowing implicit conversions from `Kokkos::View` to `Teuchos::ArrayView`?  We currently allow implicit conversions from `std::vector` and `Teuchos::ArrayRCP` to `Teuchos::ArrayView`.  If we allowed for implicit conversions from `Kokkos::View` to `Teuchos::ArrayView` in addition to this change, then that would make it even easier, no?  The only danger in allowing for the implicit conversion is that it opens the set of problems described in in section "5.7.3 Implicit type conversion problems and shortcomings" in [Teuchos Memory Management Classes](https://trilinos.org/docs/dev/packages/teuchos/doc/html/TeuchosMemoryManagementSAND.pdf).

mhoemmen:
Thanks @bartlettroscoe !  

[...]

PR #4253, opened 2019-01-23
Fix #4250, #4234, #4249, and #3823

### LINKED ISSUE #4250: TeuchosKokkosCompat: Test fails to build
@trilinos/teuchos @trilinos/kokkos @trilinos/tpetra 

A test in TeuchosKokkosCompat fails to build.  I get a few errors like the following:
```
.../Trilinos/packages/teuchos/kokkoscompat/test/linkTest.cpp: In member function ‘virtual void LinkTeuchosAndKokkos_NoInteraction_UnitTest::runUnitTestImpl(Teuchos::FancyOStream&, bool&) const’:
.../Trilinos/packages/teuchos/kokkoscompat/test/linkTest.cpp:109:3: error: ‘initialize’ is not a member of ‘TestDevice {aka Kokkos::OpenMP}’
   TestDevice::initialize();
   ^
.../Trilinos/packages/teuchos/kokkoscompat/test/linkTest.cpp:126:3: error: ‘finalize’ is not a member of ‘TestDevice {aka Kokkos::OpenMP}’
   TestDevice::finalize();
   ^
```

## Related Issues
* Blocks #4249, #4234

mhoemmen:
@trilinos/tpetra @trilinos/kokkos @trilinos/teuchos 

This fixes a few Teuchos issues, and one Tpetra issue.  The Tpetra fix is actually a work-around for kokkos-kernels issues.  @kyungjoo-kim might have some fixes there; see e.g., https://github.com/kokkos/kokkos-kernels/pull/373.

This may be "results impacting," because it affects how Tpetra computes dot products for complex Scalar types.  In particular, the work-around always uses `complex<double>` as the accumulator type.  This should improve accuracy for the `Scalar=complex<float>` case.

## Related Issues

* Closes #4250, #4234, #4249, #3823

### DIFF
--- a/packages/teuchos/core/src/Teuchos_ArrayView.hpp
+++ b/packages/teuchos/core/src/Teuchos_ArrayView.hpp
@@ -72,10 +72,20 @@ ArrayView<const T>::ArrayView( ENull )
 
 template<class T> inline
 ArrayView<T>::ArrayView( T* p, size_type size_in, const ERCPNodeLookup rcpNodeLookup )
-  :ptr_(p), size_(size_in)
+  :ptr_(size_in == 0 ? nullptr : p), size_(size_in)
 {
 #ifdef HAVE_TEUCHOS_ARRAY_BOUNDSCHECK
-  TEUCHOS_TEST_FOR_EXCEPT( p != 0 && size_in <= 0 );
+  // We comment out the one check below, as part of the fix for #4234:
+  //
+  // https://github.com/trilinos/Trilinos/issues/4234
+  //
+  // This permits conversion from std::vector or Kokkos::View, using
+  // ArrayView(x.data(), x.size(), RCP_DISABLE_NODE_LOOKUP).  The
+  // other part of the fix is that we make sure ptr_ is null if
+  // size_in is zero.
+  //
+  //TEUCHOS_TEST_FOR_EXCEPT( p != 0 && size_in <= 0 );
+
   TEUCHOS_TEST_FOR_EXCEPT( p == 0 && size_in != 0 );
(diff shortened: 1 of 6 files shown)


mhoemmen:
@lucbv You're the one who reported complex scalar types being broken in Tpetra::MultiVector, right?

lucbv:
@mhoemmen, yes I did report that fairly early.

ndellingwood:
Also adding reference to kokkos/kokkos-kernels#374, it looks like when that PR is complete it should fix the underlying BLAS tpl issue with zdot by providing configure-time (CMake) or makefile options to use the proper BLAS interface (some BLAS installs expect the dot result as the first argument to zdot and the routine has void return type, others provide the result via return type, etc).

@mhoemmen once the BLAS TPL issue is resolved in kokkos-kernels is the plan for Tpetra to go back to the `KokkosBlas::dot` routine or to continue using the routine your added in this PR? Just checking if a follow-on patch to Trilinos should be issued or if changes to Tpetra need to be made during the next promotion testing, or if you'd prefer no changes.

mhoemmen:
@ndellingwood wrote:
> once the BLAS TPL issue is resolved in kokkos-kernels is the plan for Tpetra to go back to the `KokkosBlas::dot` routine or to continue using the routine your added in this PR?

The plan is for Tpetra to go back to the `KokkosBlas::dot` routine, once it works correctly.  The point of the work-around is to get Tpetra tests passing for complex Scalar types.

I do think it would make sense, especially for `float` and `complex<float>`, to bypass the BLAS' dot products and roll our own dot products that use an extended-precision accumulator.  However, I think the existing `KokkosBlas::dot` interface suffices for that.

Issue #4258, opened 2019-01-24
Remove deprecated inheritance of SerialDense matrices from Object


rhoope:
Teuchos: Mike Eldred is using SerialDenseMatrix objects for his work in Dakota and found relatively large memory bloat associated with inherited string data from the Object base class, eg labels.  He also found side benefits associated with not inherting the Object operator<< method.

This is a follow on to Issue #4183 and essentially seeks to pull the trigger on the deprecation from 2014 in the Object class, cf Teuchos_Object.hpp:

> (mfh 23 Nov 2014) Pretty much just ignore the above
         description of this class.  Very few classes in Teuchos inherit
         from Object.  Such inheritance should be treated as deprecated
         legacy behavior.  It's not 1990 and C++ is not Java 1.0; we
         don't need a common base class for everything.

@trilinos/teuchos

Labels: performance

## Motivation and Context

Here is Mike's summary:

>     This patch removes Teuchos::Object from inheritance chains in all Teuchos::
    SerialDense* classes.  This has two primary effects:
    > eliminates overhead from the Teuchos::Object data attributes, especially
      the std::string label_.
    > removes the operator<< (std::ostream& os, const Teuchos::Object& obj) that
      has caused problems for us defining our own operators for Teuchos types.
    
    *** Important note: the latter change may induce requirements on client code,
    *** if they were reliant on this ostream operator.  For Dakota, this is a good
    *** thing as it provides the opportunity to replace an inconsistently formatted
    *** operator (that has previously been in the way) with a tailored one.

## Definition of Done

The advertised deprecated inheritance of SerialDense matrices is removed.

## Your Environment

- **Relevant repo SHA1s:**  
- **Relevant configure flags or configure script:**  
- **Operating system and version:**  
- **Compiler and TPL versions:**  

## Related Issues

* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Additional Information

I'm going to issue a pull request with at least two commits with the first containing changes needed to Teuchos proper followed by a second with the changes to client code throughout Trilinos identified using the GCC 4.8.4 PR testing setup reflected in cmake/std/PullRequestLinuxGCC4.8.4TestingSettings.cmake with Teuchos enabled and HDF5 disabled.

jhux2:
@trilinos/teuchos

mhoemmen:
Yay for modern C++ idioms!

PR #4259, opened 2019-01-24
Issue4258 remove teuchos sdmatrix inheritance

rhoope:
@trilinos/<teamName> (Several affected)

## Description

See Issue #4258 for details

## Motivation and Context

Closes Issue #4258 

## How Has This Been Tested?

Yes, using GCC 4.8.4 PR setup with Teuchos enabledand HDF5 disabled (includes all forward and optional packages affected).

## Checklist

- [x ] My commit messages mention the appropriate GitHub issue numbers.
- [x ] My code follows the code style of the affected package(s).
- [? ] My change requires a change to the documentation.
- [ x] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [ X] These changes break backwards compatibility.

### DIFF
--- a/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
+++ b/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
@@ -285,9 +285,10 @@ int main(int argc, char *argv[])
       MyOM->stream(Warnings) << "   ||<Y1,X1> - <X1,Y1>||     : " << err << endl;
       // test s.p.d. of <Y1,X1>: try to compute a cholesky
       lapack.POTRF('U',yTx1.numCols(),yTx1.values(),yTx1.stride(),&info);
-      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
-          "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " << info
-          << "\nyTx1: \n" << yTx1);
+      std::ostringstream msg("New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == ");
+      msg << info << "\nyTx1: \n";
+      yTx1.print(msg);
+      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error, msg);
 
       // X2 ortho to Y1
       MVT::MvRandom(*X2);
@@ -345,9 +346,10 @@ int main(int argc, char *argv[])
       MyOM->stream(Warnings) << "   ||<Y2,X2> - <X2,Y2>||     : " << err << endl;
       // test s.p.d. of <Y2,X2>: try to compute a cholesky
       lapack.POTRF('U',yTx2.numCols(),yTx2.values(),yTx2.stride(),&info);
-      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
-          "New <Y2,X2> not s.p.d.: couldn't computed Cholesky: info == " << info
-          << "\nyTx2: \n" << yTx2);
+      msg.str("New <Y2,X2> not s.p.d.: couldn't computed Cholesky: info == ");
+      msg << info << "\nyTx2: \n";
+      yTx2.print(msg);
+      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error, msg);
     }
     MyOM->stream(Warnings) << endl;
 
--- a/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
+++ b/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
@@ -288,7 +288,7 @@ int main(int argc, char *argv[])
       std::ostringstream msg("New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == ");
       msg << info << "\nyTx1: \n";
       yTx1.print(msg);
-      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error, msg);
+      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error, msg.str());
 
       // X2 ortho to Y1
       MVT::MvRandom(*X2);
@@ -349,7 +349,7 @@ int main(int argc, char *argv[])
       msg.str("New <Y2,X2> not s.p.d.: couldn't computed Cholesky: info == ");
       msg << info << "\nyTx2: \n";
       yTx2.print(msg);
-      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error, msg);
+      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error, msg.str());
     }
     MyOM->stream(Warnings) << endl;
 
--- a/packages/anasazi/epetra/test/OrthoManager/cxx_mattest.cpp
+++ b/packages/anasazi/epetra/test/OrthoManager/cxx_mattest.cpp
@@ -661,13 +661,13 @@ int testProjectAndNormalizeMat(RCP<MatOrthoManager<ST,MV,OP> > OM,
           if (err > ATOL*TOL) {
             sout << "S\n" << *S << endl;
             sout << "S_out\n" << *S_outs[o] << endl;
-            sout << "B_out\n" << *B_outs[o] << endl;
+            sout << "B_out\n"; B_outs[o]->print(sout); sout << endl;
             if (C_outs[o].size() > 0) {
               sout << "X1\n" << *X1 << endl;
-              sout << "C_out[1]\n" << *C_outs[o][0] << endl;
+              sout << "C_out[1]\n"; C_outs[o][0]->print(sout); sout << endl;
               if (C_outs[o].size() > 1) {
                 sout << "X22n" << *X2 << endl;
-                sout << "C_out[2]\n" << *C_outs[o][1] << endl;
+                sout << "C_out[2]\n"; C_outs[o][1]->print(sout); sout << endl;
               }
             }
           }
--- a/packages/anasazi/src/AnasaziBlockDavidsonSolMgr.hpp
+++ b/packages/anasazi/src/AnasaziBlockDavidsonSolMgr.hpp
@@ -679,8 +679,9 @@ BlockDavidsonSolMgr<ScalarType,MV,OP>::solve() {
 #       ifdef TEUCHOS_DEBUG
           {
             std::stringstream os;
-            os << "KK before HEEV...\n"
-              << *state.KK << "\n";
+            os << "KK before HEEV...\n";
+            state.KK->print(os);
+            os << "\n";
             *out << os.str();
           }
 #       endif
@@ -695,8 +696,9 @@ BlockDavidsonSolMgr<ScalarType,MV,OP>::solve() {
             std::stringstream os;
             *out << "Ritz values from heev(KK):\n";
             for (unsigned int i=0; i<theta.size(); i++) *out << theta[i] << " ";
-            os << "\nRitz vectors from heev(KK):\n" 
-                                                 << S << "\n";
+            os << "\nRitz vectors from heev(KK):\n";
+            S.print(os);
+            os << "\n";
             *out << os.str();
           }
 #       endif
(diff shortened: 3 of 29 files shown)


mhoemmen:
@trilinos/teuchos

packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
```diff
@@ -285,9 +285,10 @@ int main(int argc, char *argv[])
       MyOM->stream(Warnings) << "   ||<Y1,X1> - <X1,Y1>||     : " << err << endl;
       // test s.p.d. of <Y1,X1>: try to compute a cholesky
       lapack.POTRF('U',yTx1.numCols(),yTx1.values(),yTx1.stride(),&info);
-      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
-          "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " << info
-          << "\nyTx1: \n" << yTx1);
+      std::ostringstream msg("New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == ");
```
mhoemmen: Hm, I'd rather not build the `std::ostringstream` each time if we only use it when `info` is nonzero.  How about this?
```c++
if (info != 0) {
  // build the std::ostringstream
  yTx1.print (msg);
  TEUCHOS_TEST_FOR_EXCEPTION(true, std::runtime_error, msg); 
}
```

packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
```diff
@@ -345,9 +346,10 @@ int main(int argc, char *argv[])
       MyOM->stream(Warnings) << "   ||<Y2,X2> - <X2,Y2>||     : " << err << endl;
       // test s.p.d. of <Y2,X2>: try to compute a cholesky
       lapack.POTRF('U',yTx2.numCols(),yTx2.values(),yTx2.stride(),&info);
-      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
-          "New <Y2,X2> not s.p.d.: couldn't computed Cholesky: info == " << info
-          << "\nyTx2: \n" << yTx2);
+      msg.str("New <Y2,X2> not s.p.d.: couldn't computed Cholesky: info == ");
```
mhoemmen: See above.

packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
mhoemmen: I realize now that the above code is in a test, so I'm less worried about it.

packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
rhoope: I believe the suggestion to implement the operator<< is exactly what Mike did in Dakota.  If I added one for Teuchos::SerialDenseMatrix<Scalar>&, that would remove the need to touch the other Trilinos packages but in a way that has proven overly prescriptive for Dakota.  Is there a more attractive approach, eg a virtual operator<< ?

mhoemmen: @rhoope You could just do an `@`-mention of this GitHub account name here.  That should notify him.

mhoemmen: @rhoope wrote:
> Is there a more attractive approach, e.g., a virtual `operator<<` ?

Is the issue that you would like Dakota to be able to override any existing print behavior of `Teuchos::SerialDenseMatrix`, but still be able to use `operator<<` everywhere?

It's hard to do that when using an operator.  Argument-dependent lookup means that you can't use the same input types (`std::ostream&, const SDM&`) and still get a different `operator<<`.  However, you could achieve this effect in Dakota with a little wrapper that changes SDM into a different type.  This is analogous to ["function advising"](https://www.emacswiki.org/emacs/AdvisingFunctions) in some other languages.  Here is an example:
```c++
// Implementation detail; not for users
template<class OrdinalType, class ScalarType>
struct SerialDenseMatrixOutputWrapper {
  SerialDenseMatrix<OrdinalType, ScalarType>& A;
};

// this is for Dakota developers
template<class OrdinalType, class ScalarType>
SerialDenseMatrixOutputWrapper<OrdinalType, ScalarType>
wrapSDM (SerialDenseMatrix<OrdinalType, ScalarType>& A)
{
  return {A};
}

// Dakota would define this
template<class OrdinalType, class ScalarType>
std::ostream& operator<< (std::ostream& out, const SerialDenseMatrixOutputWrapper<OrdinalType, ScalarType>& A)
{
  // ... print A however Dakota likes to print it ...
  return out;
}
```
You would use the class this way:
```c++
SerialDenseMatrix<int, double> A ( /* whatever goes here */ );
// ... do stuff with A ...

std::cout << wrapSDM (A) << std::endl;
```

rhoope: Thanks @mhoemmen for this feedback.  I'm going to engage Mike to have him consider any implications to the larger Dakota context I'm not aware of.

mhoemmen: @rhoope Thanks!  I will mark this PR as a work in progress (WIP) for now, so that it doesn't repeat PR testing.

mseldre: Hi Mark,
First a little context.  The primary issue was to shed the overhead (especially the string_ label) from the Object base class.  We had a user report that 70% of the Dakota memory in a large study was being consumed by the "Teuchos_BlahBlah" labels on every instance of Vector,Matrix,SymMatrix.  Simply dropping the inheritance from Object nicely accomplishes that, and we noticed your comments in the base about the need to deprecate and retire this legacy code.  
As a bonus, I noticed that we could also shed the operator<<(ostream&, Object&) definition that has caused clashes for us in the past.  These bind to the virtual print() function, and we have no issue with providing a virtual print() function -- rather the issue was clashes between Teuchos::operator<<(ostream&, Object&) and corresponding Dakota::operator<<(ostream&, SerialDense*) that we would like to provide to conveniently support our own formatting for Vector,Matrix,SymMatrix.  This clash was historical and may have been due to "using namespace Teuchos" pollution somewhere in the header dependencies -- not sure.  But not having to worry about that clash anymore simplifies our code and allows us to retire a number of workarounds.
So my recommendation would be to delete Object and allow Teuchos clients to provide their own customized insertion/extraction operators.  The default one provided by Object is probably useful for debugging output, but is otherwise rather verbose and unformatted.
So Russell's solution for the tests was to replace uses of operator<< with uses of print(); alternatively, Anasazi could define an operator<< in their namespace if useful to them outside of debug output for unit tests.
I believe that this convention would mirror that of the STL, in that they don't provide default insertion/extraction operators for their data types and allow clients to define their own.
Does that help?
--Mike

mhoemmen: @mseldre wrote:
> Simply dropping the inheritance from Object nicely accomplishes that, and we noticed your comments in the base about the need to deprecate and retire this legacy code.

I very much appreciate your contributions here -- thanks!

> This clash was historical and may have been due to "using namespace Teuchos" pollution somewhere in the header dependencies -- not sure.

Argument-dependent lookup (ADL) would normally pick up `Teuchos::operator<<`, since it is in the same namespace as one of its arguments.  ADL would not normally pick up an overloaded `Dakota::operator<<`, even inside code in the Dakota namespace.  This is true even without `using namespace Teuchos;` etc. 

> So my recommendation would be to delete Object and allow Teuchos clients to provide their own customized insertion/extraction operators.

I'm OK with this personally, but it does break backwards compatibility for downstream applications.  It's very likely that you won't be able to test all possible combinations of build options even within Trilinos, so this may break some Trilinos packages.  It may also break downstream apps.

Trilinos hasn't had an official minor release (!) since 12.12, around three years ago (!!!).  This means that all bets are off in terms of deprecation and removal policy.  (Feel free to express your opinions about that to @maherou.  I have no control over it.)  Lack of policy mainly just means that you'll have to manage the angry-grams from various apps if their builds break.  If you're OK with that, we can merge this and file a separate issue where you will be listed as the point of contact for explaining how to work around any build issues that arise.  It's no big deal, but I just want to make sure we document why we did this, that it does break backwards compatibility, and how to fix it.  Is that OK?

mseldre: Thanks for the pointer about ADL -- that explains what I was observing and, IMO, also strengthens the argument for not providing these operators.

I can't speak to internal Trilinos policy for deprecation or release management, but seems there needs to be some mechanism to move forward.  From the Dakota side, we would of course like to see our local changes to Teuchos propagate upstream so that users can build with either our version or their own custom version of Trilinos, with minimal/no inconsistencies.  I am fine with delaying this patch until the appropriate time when deprecations can be announced and managed -- in the meantime, we can get by with our local version.  In terms of being the POC for Trilinos angry-grams, yes if that's necessary in the short term. 

On a semi-related note, I was also eyeing the CompObject data that appears to go unused by SerialDense*.  I added some (redundant) protection around the computation of FLOP increments since the flop accumulator that accepts them seems to always be NULL based on ctor initializers. It would be easy to manage ctor paths with and without meaningful FLOP counters, but it seems to currently be an afterthought. Am I missing context?

mhoemmen: @mseldre Thanks for the reply!  In brief, I would say, go for it! :-D  (Ditto for the CompObject changes, though please in a separate PR :-) .)  If you're OK with getting a few questions about how to fix build errors, I'll be happy to approve this PR.

Long answer: @ikalash, @rrdrake, @theguruat12, @micahahoward, @rppawlo, or @trilinos/framework may wish to chime in, but I would summarize the de facto "Trilinos versioning" policy for many internal applications as follows:

  1. App either forks Trilinos, or picks a Trilinos "master" branch commit.
  2. App defines its own acceptance testing for Trilinos.
  3. When an app wants to update its Trilinos, it uses acceptance tests to decide when to merge / pull.  Many apps facilitate this by testing the app nightly against current master Trilinos.

This implies that apps define their own "Trilinos versions."  Apps actually use this model.  That, along with open access to Trilinos' repo on GitHub, likely explains the lack of enthusiasm for official Trilinos releases.  I've gotten about as many complains about deprecation warnings ("hey, we can't build warning-free") as I've gotten about actual interface changes, so it's likely less effort just to break the build than it is to do an official deprecation (how long? who knows, since no official releases!).

What's missing from this model is a way for an app to work correctly with multiple Trilinos versions at once.  A few packages have added this feature themselves.  For example, Kokkos has a CMake option and macro to help manage deprecation.  Tpetra just recently added one.  It may make sense for Teuchos to add this as well, but Teuchos is a big utility package without a single focus, so it may be hard to enforce the deprecation model and decide when to remove deprecated stuff.

Anyway, I hope the long explanation helps!

mseldre: Thanks for the explanation.  Approving the PR sounds good to me.  Happy to help with suggestions/examples of supplying ostream insertion operators for Vector,Matrix,SymMatrix.

mhoemmen:
@trilinos/rol @trilinos/anasazi @trilinos/belos

mhoemmen:
This breaks Anasazi’s build. Not sure why your texting didn’t pick this up — maybe your script didn’t enable Anasazi.

mhoemmen:
Y'all have fun! 👍

bartlettroscoe:
CC: @rhoope, @mhoemmen, @mseldre 

FYI: This broke SPARC.  See #4330. 

Can we find a way to do what Dakota wants without breaking backward compatibility?

mseldre:
Hi Ross,
Can I ask what you were using the provided ostream insertion operator for?  This operator provides a lot of debug-level output that seems most appropriate for unit testing, and is "in the way" in terms of providing a better operator due to operator precedence rules.  When a similar issue arose with Anasazi (@rhoope), the better solution was to move forward and either remove the dependence on this operator or provide a better one.

PR #4266, opened 2019-01-25
Zoltan2: Restore Zoltan2 test and Kokkos unordered map

MicheldeMessieres:
@trilinos/zoltan2

## Motivation and Context
This is a follow up PR to clean up a few issues from PR #4252.

It restores a missing zoltan2 test.
It restored the directory to use Kokkos unordered map (now host space) anticipating we'll be implementing parallel forms in the future.
It cleans up some of the directory hash code to avoid a global static method definition.

## How Has This Been Tested?
Tested on white cuda.
Tested on mac clang parallel and serial.

## Checklist

### DIFF
--- a/packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
+++ b/packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
@@ -50,53 +50,6 @@
 #include "Zoltan2_Directory.hpp"
 #include "Zoltan2_Directory_Comm.hpp"
 
-// Copied from murmurc.3
-// TODO: Will be removed as Kokkos form develops
-#define ZOLTAN2_ROTL32(x,r) (uint32_t) \
-  (((uint32_t)(x) << (int8_t)(r)) | ((uint32_t)(x) >> (32 - (int8_t)(r))))
-static inline void Zoltan2_MurmurHash3_x86_32 ( const void * key, int len,
-  uint32_t seed, void * out )
-{
-  const uint8_t * data = (const uint8_t*)key;
-  const int nblocks = len / 4;
-  int i;
-  uint32_t h1 = seed;
-  uint32_t c1 = 0xcc9e2d51;
-  uint32_t c2 = 0x1b873593;
-  const uint32_t * blocks = (const uint32_t *)(data + nblocks*4);
-  for(i = -nblocks; i; i++)
-  {
-    uint32_t k1 = blocks[i];
-    k1 *= c1;
-    k1 = ZOLTAN2_ROTL32(k1,15);
-    k1 *= c2;
-    h1 ^= k1;
-    h1 = ZOLTAN2_ROTL32(h1,13);
-    h1 = h1*5+0xe6546b64;
-  }
-  const uint8_t * tail = (const uint8_t*)(data + nblocks*4);
-  uint32_t k1 = 0;
-  switch(len & 3)
-  {
-    case 3: k1 ^= tail[2] << 16;
-    case 2: k1 ^= tail[1] << 8;
-    case 1: k1 ^= tail[0];
-            k1 *= c1; k1 = ZOLTAN2_ROTL32(k1,15); k1 *= c2; h1 ^= k1;
-  };
-  h1 ^= len;
-  h1 ^= h1 >> 16;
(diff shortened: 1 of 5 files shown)


packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
```diff
@@ -595,8 +538,7 @@ int Zoltan2_Directory<gid_t,lid_t,user_t>::update_local(
   node.owner = owner;
   node.errcheck = owner;
 
-  if(!node_map.insert({*gid, node}).second) {
-  // if(node_map.insert(*gid, node).failed()) { // TODO Switch to Kokkos map
+  if(node_map.insert(*gid, node).failed()) {
     // Need more nodes (that's the assumption here if we have failure)
```
mhoemmen: `failed()` isn't the only possible non-successful result; see `freed_existing` in `Kokkos_UnorderedMap.hpp`.

packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
```diff
@@ -901,13 +842,11 @@ int Zoltan2_Directory<gid_t,lid_t,user_t>::find_local(
...
-  // if(node_map.valid_at(node_index))
+  size_t node_index = node_map.find(*gid);
+  if(node_map.valid_at(node_index))
   {
     const Zoltan2_Directory_Node<gid_t,lid_t,user_t> & node =
-      node_map.at(*gid);
-    //  node_map.value_at(node_index); // TODO: Switch to Kokkos map
+      node_map.value_at(node_index);
     /* matching global ID found! Return gid's information */
     if(lid) {
```
mhoemmen: Not a fan of a pointer being called `lid`.  That confused me until I saw the next line.

packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
```diff
@@ -1084,15 +1023,15 @@ int Zoltan2_Directory<gid_t,lid_t,user_t>::remove(
   // eliminating remove_local and just calling here but will keep for now to
   // keep symmetry with other modes.
   if(nrec > 0) {
-    // node_map.begin_erase(); // TODO Switch to Kokkos Map
+    node_map.begin_erase();
     for (int i = 0; i < nrec; i++)  {
       msg_t *ptr = reinterpret_cast<msg_t*>(&(rbuff[i*remove_msg_size]));
```
mhoemmen: Does this obey ANSI strict aliasing rules?  (Remember the `memcpy` line of code I showed a little while ago?)

MicheldeMessieres: This one doesn't give me the warning when I compile. It's a ptr cast versus the actual ptr write you showed me before.

I tried the following cases just to see how this works.
Only the 3rd case (matching the original memcpy case you suggested) gives the warning.

```
  // no strict-alias warning
  {
    char * char_ptr[10];
    int * ptr = reinterpret_cast<int *>(char_ptr);
  }

  // no strict-alias warning
  {
    char * char_ptr = new char[10];
    int * ptr = reinterpret_cast<int *>(char_ptr);
  }

  // strict-alias warning
  {
    char * char_ptr[10];
    int * ptr;
    *ptr = *(reinterpret_cast<int *>(char_ptr));
  }

  // no strict-alias warning
  {
    char * char_ptr = new char[10];
    int * ptr;
    *ptr = *(reinterpret_cast<int *>(char_ptr));
  }
```

packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
MicheldeMessieres: Agreed. Some code is still replicated from zoltan and as I continue to work on it, I need to convert the original c code to be more c++ like. We plan to return to this to add parallel forms. I will improve it.

packages/zoltan2/src/directory/Zoltan2_Directory_Impl.hpp
MicheldeMessieres: I think freed_existing() is only relevant for parallel or threaded cases? I can add this when we develop parallel forms.

PR #4292, opened 2019-01-29
Tpetra crs matrix swap

william76:
@trilinos/tpetra 

CrsMatrix::Swap() implementation

FYI @csiefer2


packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -682,20 +683,51 @@ namespace Tpetra {
...
+    Teuchos::RCP<Teuchos::FancyOStream> pOut;
+    if(0==myrank)
+        pOut = Teuchos::getFancyOStream(Teuchos::rcpFromRef(std::cout));
+    else
+        pOut = Teuchos::getFancyOStream(Teuchos::rcp(new Teuchos::oblackholestream()));
+    Teuchos::FancyOStream& out = *pOut;
+
+    out << "-------------------------------------------------------------------------" << std::endl;    
+
+    this->staticGraph_->getRangeMap()->describe(out, Teuchos::VERB_EXTREME);
```
mhoemmen: Please don't check in unprotected debug print code.

mhoemmen:
@william76 Sorry, I was occupied today with a massive Tpetra fix for GPU + MPI.

william76:
@mhoemmen no worries!  Thanks for having a look :)

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -2126,12 +2129,24 @@ namespace Tpetra {
...
     //! Throw an exception if the internal state is not consistent.
     void checkInternalState () const;
 
+    /// \brief Swaps the data from *this with the data and maps from graph.
+    ///
+    /// \param graph [in/out] a crsGraph
+    void swap(CrsGraph<LocalOrdinal, GlobalOrdinal, Node> & graph);
+
+    // Friend the tester for CrsGraph::swap
+    template <class LO, class GO, class N> friend class Tpetra::crsGraph_Swap_Tester;
```
kddevin: Question to the experts about syntax:
should this be
`friend class Tpetra::crsGraph_Swap_Tester<LocalOrdinal, GlobalOrdinal, Node>;`
(without the template specification at the beginning of the line)?

packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp
```diff
@@ -4788,6 +4786,18 @@ namespace Tpetra {
     Kokkos::View<impl_scalar_type*, execution_space, Kokkos::MemoryUnmanaged>
     getRowViewNonConst (const RowInfo& rowInfo) const;
 
+
+  protected:
+
+    // Friend the tester for CrsMatrix::swap
+    template <class LO, class GO, class N> friend class Tpetra::crsMatrix_Swap_Tester;
```
kddevin: Questions to the experts about syntax:

Should this be 
`friend class Tpetra::crsMatrix_Swap_Tester<LocalOrdinal, GlobalOrdinal, Node>;`
(without the template specification at the beginning of the line)?

And why are ordinal and node templates needed, but scalar template is not?  Will this test break if I build Trilinos with only float scalars (not double scalars)?

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -87,6 +87,17 @@ namespace Tpetra {
              const Teuchos::RCP<typename OutputCrsGraphType::node_type> nodeOut,
              const Teuchos::RCP<Teuchos::ParameterList>& params = Teuchos::null);
     };  // class CrsGraphCopier
+
+    template<class LO, class GO, class NT>
+    void
+    unpackCrsGraphAndCombine(
```
mhoemmen: What's going on here?  Was this declaration not here already?

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
william76: Fixed.

packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp
william76: Fixed the friend and set up the CrsMatrix test to use the SLGN test for the 4 template parameters.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
william76: @mhoemmen that got picked up when I merged develop into my branch... I had a merge conflict so that's why it's being picked up... I just picked what came in from the `develop` branch.

william76:
merged develop into branch to resolve conflicts.

william76:
@csiefer2 @kddevin 
When you get a chance, please have a look and check the review box if things look good... feel free to add the automerge flag if you're cool with it merging on in if the testing passes :)

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -44,348 +44,434 @@
...
-    clp.setOption("filedir", &filedir, "Directory of expected input files.");
-    clp.addOutputSetupOptions(true);
-    clp.setOption("test-mpi",
-                  "test-serial",
-                  &Tpetra::TestingUtilities::testMpi,
-                  "Test MPI (if available) or force test of serial.  In a serial build,"
-                  " this option is ignored and a serial comm is always used.");
-    clp.setOption("error-tol-slack", &errorTolSlack, "Slack off of machine epsilon used to check test results");
-}
+        TEUCHOS_TEST_FOR_EXCEPTION(initialComm->getSize() < 2, std::runtime_error, "This test requires at least two processors.");
```
lucbv: This should really be avoided, throwing on `comm-getSize()` leads to failing tests, you really want you test to handle all cases, after all are users forbidden to call this method on serial matrices?

packages/tpetra/core/test/CrsMatrix/CMakeLists.txt
```diff
@@ -47,6 +47,16 @@ TRIBITS_ADD_EXECUTABLE_AND_TEST(
   STANDARD_PASS_OUTPUT
   )
 
+TRIBITS_ADD_EXECUTABLE_AND_TEST(
+  CrsMatrix_UnitTests_Swap
+  SOURCES
+    CrsMatrix_UnitTests_Swap
+    ${TEUCHOS_STD_UNIT_TEST_MAIN}
+  ARGS ${ARGS}
+  COMM serial mpi
```
lucbv: If you want this problem to run only with 2 ranks you should specify it here, also if you want to avoid the case with a single rank in the communicator then specifying `COMM serial` seems wrong...

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
kddevin: I agree with @lucbv .  swap() should work for serial builds and MPI builds run on one rank, so we should test those cases.  Is the two-rank case just a legacy from an import test or similar?  

Are there any problems we might catch on more than two processors that we wouldn't see on only two?  ("No" is an acceptable answer if you think it is true.)

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
```diff
@@ -44,348 +44,434 @@
...
-using Tpetra::TestingUtilities::getDefaultComm;
-typedef Tpetra::global_size_t GST;
+namespace Tpetra {
 
-double      errorTolSlack = 1e+1;
-std::string filedir;
+template<class LocalOrdinal, class GlobalOrdinal, class Node>
+class crsGraph_Swap_Tester
+{
+    using Scalar                   = int;                                                   // Not used here, kept for consistency.
```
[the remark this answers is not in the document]
william76: It is used, but just for the structs I'm passing into the graph generator in the test so warnings won't get generated. It's kept here mostly so that I can just use the same thing for generating my graphs in both this test and the CrsMatrix swap test.  
In the CrsMatrix Swap test, the scalar parameter is coming in from Trilinos so I don't have this line in that test. For here, it's ok because I'm just using it to keep the struct happy.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -73,9 +73,11 @@
 #include "Teuchos_FancyOStream.hpp"
 #include "Teuchos_RCP.hpp"
 #include "Teuchos_SerialDenseMatrix.hpp" // unused here, could delete
+#include "Teuchos_FancyOStream.hpp"
```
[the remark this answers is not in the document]
william76: I think it's an artifact from testing stuff. I'll remove it and check in and update.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests_Swap.cpp
william76: @kddevin @lucbv 
There's nothing in Swap that's tied to any particular configuration of processors, so swap() itself should work fine since we're generally using std::swap() in most places. 

Generating the matrices and graphs for 2 processes is pretty much the result of me being fairly new to generating CrsGraphs and CrsMatrices and there aren't a lot of small(ish) self-contained examples that made much sense to me prior to this which can just generate a simple graph, so I used [this example][1] to generate the graphs, which followed on to the matrix stuff by copy and pasting.

I think it'd be good to rework this to work for 1..N processors... but if we make that a blocker on this PR I can kill it, but it's highly unlikely that I'll be getting that done until mid to late April at a minimum since I have to switch over to other projects for a while.  Can we put in an issue to rework the Swap tests to support 1..N processors so that this can get in for the FE* stuff?

[1]: https://github.com/trilinos/Trilinos/wiki/Tpetra::CrsGraph-creation,-from-a-graph-perspective

packages/tpetra/core/test/CrsMatrix/CMakeLists.txt
william76: @lucbv 
Ah, ok. I've modified the CMakeLists.txt here to just do mpi and to set it to 2 processes since that's what the test is set up for for now. In the future if we can make the test work for 1..N processes then we can update the CMakeLists.txt accordingly.
Running a quick local build before pushing this change up.

william76:
@kddevin it passed the PR testing, should we go ahead and push and put an issue in to update the tests later to support 1..N processors?  I think that's the only thing that's really left on this one...?

kddevin:
@william76 That's fine with me.  Thanks.

william76:
@kddevin ok. If you're good for this to merge, can I get the review checkbox marked :)

william76:
someone checked in a change to Tpetra_CrsMatrix_Swap to remove some warnings that changed lines that I had commented out so... yay, merge conflict.  Retesting it.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_UnitTests_Swap.cpp
```diff
@@ -0,0 +1,861 @@
...
+
+            out << std::endl << ">>> Compare matrix_b == matrix_d" << std::endl;
+            TEST_EQUALITY(compare_crsmatrix(comm, out, *matrix_b, *matrix_d),
+                          true);      // matrix_b and matrix_d shoudl be the same
+
+            //
+            // Swap matrix b and c again
+            //
+            out << ">>> swap matrix_b and matrix_c" << std::endl << std::endl;
+            matrix_c->swap(*matrix_b);
```
csiefer2: .

PR #4301, opened 2019-01-30
MueLu: Adding experimental node comm support

csiefer2:
Auto-PR for SHA 009a8bf

### DIFF
--- a/packages/muelu/src/Utils/MueLu_Utilities.cpp
+++ b/packages/muelu/src/Utils/MueLu_Utilities.cpp
@@ -277,7 +277,7 @@ bool IsParamValidVariable(const std::string& name)
        // Sort!
        std::sort(addressList.begin(),addressList.end());
 
-       // Find the number of node-buddies I have
+       // Find which node I'm on (and stop when I've done that)
        int numNodes = 0;
        for(int i=0, prev=addressList[0]; i<numRanks && prev != myaddr; i++) {
          if(prev != addressList[i]) {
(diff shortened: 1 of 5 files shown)


csiefer2:
Checked on serrano w/ the debugging stuff on and it does what we want.

packages/muelu/src/Utils/MueLu_Utilities.cpp
```diff
@@ -248,4 +256,43 @@ bool IsParamValidVariable(const std::string& name)
...
+       int myaddr = (int) htonl(inet_network(inet_ntoa(*(struct in_addr *)host->h_addr)));
+       
+       // All-to-all exchange of address integers
+       std::vector<int> addressList(numRanks);
+       Teuchos::gatherAll(*baseComm,1,&myaddr,numRanks,&addressList[0]);
+
+       // Sort!
+       std::sort(addressList.begin(),addressList.end());
+
+       // Find the number of node-buddies I have
```
jhux2: This comment misled me.  The code below is identifying the node ID of the current MPI rank, counting up from zero, isn't it?  All MPI ranks on a node should calculate the same node ID.

csiefer2: Oh, yeah, that comment is not correct.

csiefer2:
> Wait, woah, I thought several MPI implementations (OpenMPI and MPICH at least) came with built-in "node communicators" that are just named constants you can use. Wouldn't that be `MPI_COMM_NETWORK`? Maybe that was a thing that went away, but I see some Zoltan code that refers to `MPI_COMM_NODE`, `MPI_COMM_SOCKET`, and `MPI_COMM_CACHE`.

I think those are version specific.  I do not believe there's a portable way of getting those from MPI directly.

PR #4302, opened 2019-01-30
Belos CGSingleReduce: merge allreduces

cgcgcg:
@trilinos/belos 

## Description
The CG single reduce iteration performs two allreduces per iteration: one for the inner products and one for the convergence detection.
The proposed changes allow to merge these into a single allreduce per iteration. The options for that are:
- Use the preconditioner norm $||P^{1/2}r||$ for convergence detection which is already computed anyways in the inner products. (This makes specifying a good tolerance potentially tricky.)
- Compute the 2-norm at the same time as the inner products. (The implementation is not perfect, as it incurs more computation in its current form.)

The default behavior has not been changed.


mhoemmen:
Also, @iyamazaki wrote a Tpetra-specific implementation of single-reduce CG.  You can get it from `Belos::SolverFactory` using the solver name `"TPETRA CG SINGLE REDUCE"`.  It would be very interesting to do a performance comparison.  We do need some Tpetra-specific solvers, since MultiVecTraits doesn't and can't express the necessary linear algebra operations.  However, it would be nicer to have a generic implementation of single-reduce CG.

packages/belos/src/BelosBlockCGSolMgr.hpp
```diff
@@ -516,6 +521,9 @@ setParameters (const Teuchos::RCP<Teuchos::ParameterList> &params)
   // Check if the user is requesting the single-reduction version of CG (only for blocksize == 1)
   if (params->isParameter("Use Single Reduction")) {
     useSingleReduction_ = params->get("Use Single Reduction", useSingleReduction_default_);
+    if (useSingleReduction_)
```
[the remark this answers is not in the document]
cgcgcg: Yes, that's correct.

packages/belos/src/BelosBlockCGSolMgr.hpp
```diff
@@ -370,6 +370,8 @@ namespace Belos {
     static constexpr int verbosity_default_ = Belos::Errors;
     static constexpr int outputStyle_default_ = Belos::General;
     static constexpr int outputFreq_default_ = -1;
+    static constexpr const char * resNorm_default_ = "TwoNorm";
+    static constexpr bool foldConvergenceDetectionIntoAllreduce_default_ = false;
```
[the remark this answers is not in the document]
cgcgcg: I did not change the default norm that is used for convergence detection, i.e. the 2-norm. Switching this to true saves one allreduce, but incurs more local computation. My guess is that this only pays off once the allreduce is expensive enough, so not for small communicator size.

hkthorn:
@cgcgcg I am fine with the changes.  I need for there to be a test added to ensure that this modification to the solver works and continues working.

packages/belos/src/BelosCGSingleRedIter.hpp
```diff
@@ -398,6 +427,22 @@ class CGSingleRedIter : virtual public CGIteration<ScalarType,MV,OP> {
   }
 
 
+  //////////////////////////////////////////////////////////////////////////////////////////////////
+  // Get the norms of the residuals native to the solver.
+  template <class ScalarType, class MV, class OP>
+  Teuchos::RCP<const MV>
+  CGSingleRedIter<ScalarType,MV,OP>::getNativeResiduals( std::vector<MagnitudeType> *norms ) const {
```
mhoemmen: It's technically possible to call this method with `norms == nullptr`, though I suppose the "iter" classes are just implementation details of the "SolMgr" classes.

packages/belos/src/BelosCGSingleRedIter.hpp
```diff
@@ -441,71 +496,145 @@ class CGSingleRedIter : virtual public CGIteration<ScalarType,MV,OP> {
     ////////////////////////////////////////////////////////////////
     // Iterate until the status test tells us to stop.
     //
-    while (1) {
-      
-      // Update the solution vector x := x + alpha * P_
-      //
-      MVT::MvAddMv( one, *cur_soln_vec, alpha, *P_, *cur_soln_vec );
-      lp_->updateSolution();
+    if (foldConvergenceDetectionIntoAllreduce_) {
```
mhoemmen: Not a huge fan of the massive code duplication here.

cgcgcg: I can certainly pull the if-statement into the loop. I think it's easier to read that way, since there are a couple of changes between the two paths..

hkthorn:
@cgcgcg Thanks for the information, it is very useful to know.  There are pluses and minuses to every variation of the communication reducing variants.

jwillenbring:
@cgcgcg 

I applied the retest flag to this PR because we added a new PR test this afternoon that enables CUDA. Given the status of this PR, I hope this will not negatively impact your effort.

PR #4313, opened 2019-02-01
Job weights for cloud nodes

prwolfe:
## Description

This attempts to set the parallel_level on-the-fly based on the amount
of memory available, the number of jobs that will fit on a single node,
and the minimum amount of available memory per core in use up to
this point (1.7G)

## Motivation and Context

Monitoring while starting up nodes in the CEE cloud space has shown that we
under-utilize all but 4 of the current nodes and would under-utilize the cloud
nodes as well. This should help a bit.

## How Has This Been Tested?

Testing to this point was mainly ssh to each machine and paste in the shell
commands and evaluate the results.  Since these changes are in the -test
portion of the driver I expect we will avoid the bootstrapping issue and PR
testing will produce what I need.


prwolfe:
I noticed a failure early in the PR testing. Turns out it's due to scratch being full on 142.  I am checking all the node now.

PR #4320, opened 2019-02-05
Zoltan2:  fix divide by zero error

kddevin:
## Description, Motivation and Context

@seheracer reported divide-by-zero errors when partitioning with MJ.
The error occurred in migration between recursion levels.
 
The error occurred in MJ after it computed some slices of the domain, as it was determining how many processors to assign to each slice for the next recursion level.  In cases where some coordinates have large weights and others small weights, the number of coordinates assigned to a slice might be very small (i.e., the slice has just a few heavy objects).  

MJ assigns processors to slices proportionally; slices with many coordinates get more processors.  In this case, the number of processors chosen to use to further partition this slice with few coordinates could be zero, as the value was rounded from a double to an int.  I added a guard to ensure that at least one processor was assigned to each slice.

@trilinos/zoltan2 

## How Has This Been Tested?

All existing tests pass in linux build.
@seheracer 's test outside Trilinos also succeeds.

New Zoltan2 test sacer4320 uses the Zoltan2 driver to reproduce the problem.
The input files were generated by Zoltan2::VectorAdapter::generateFiles() run from @seheracer's test program.
The test needs to run on eight MPI ranks.  Building Trilinos with
`-D MPI_EXEC_MAX_NUMPROCS:STRING=11 \`
will ensure that this test and other more interesting Zoltan and Zoltan2 tests are run.
(Zoltan and Zoltan2 never conformed with the Trilinos' "four ranks is sufficient" standard.)

With this PR, the sacer4320 test passes; without this PR, it fails. 

## Checklist

- [x ] My code follows the code style of the affected package(s).
- [ x] I have added tests to cover my changes.


mhoemmen:
See https://github.com/trilinos/Trilinos/issues/3276#issuecomment-462067485 for PR test failures.

PR #4321, opened 2019-02-05
Sacado:  Fix loop for hierarchical.

etphipp:
One loop in GeneralFad wasn't correct for Cuda with hierarchical
parallelism.  This would only affect computations with a passive scalar
(i.e., variable declared as a Fad but had a zero-length derivative
array).


etphipp:
@rppawlo I don't think this will affect any of the experiments you've been doing, but it might.

rppawlo:
Thanks @etphipp ! Will run a quick test now.

PR #4324, opened 2019-02-05
ShyLU:Build Repeated Map

roeverf:
@trilinos/shylu 

## Description
This PR adds a function to FROSch_Tools_decl/def.hpp calculating a repeated map from the dual graph and an element node list.
 
The graph ist partitioned using Zoltan2. The new column-map is used to obtain the repeated map from the element node list.

The function is used to calculate a repeated map for the coarse level. The calculation of the repeated map can be switched on and off  in the parameterlist.
For the coarse level: The required graph is build using the getConnectivityEntities function from the DDInterface class . The element node list is built from the initial row map of the coarse matrix.

So far the calculation of the repeated map for the coarse matrix is only possible if the underlying is Tpetra.

## How Has This Been Tested?
All prior existing ShyLu-DD/FROSch test were passed. For specific tests concerning the repeated map for the coarse space new parameter lists were added to the Thyra_Xpetra_Laplace  test switching on the calculation of the repeated map. Iteration counts are equal to the existing tests solving the same problem.
All test were run on a Mac OS X, compiled with MPICH version 3.2.1.

## Checklist


lucbv:
@roeverf can you please do a rebase and squash your commits into logical ones?
Ideally they would compile and run, but let's not shoot for the moon ; )

PR #4293, opened 2019-01-29
Disable Tpetra test for CUDA PR build

jwillenbring:
@trilinos/framework 
@trilinos/tpetra

## Description

I am disabling one additional Tpetra test that started failing so that we can baseline a set of passing tests to begin to clean up the remaining failures so that other build/test issues don't backslide.

## Motivation and Context

After getting the CUDA PR infrastructure changes through to master, one test started failing when we turned the CUDA testing on at the dev->master level

## Related Issues

Issue #2464 

## How Has This Been Tested?

I did not test this change, but rather am basing the specific disable on this mornings dev->master [test results](https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&parentid=4469798&filtercount=2&showfilters=1&field1=buildstarttime&compare1=84&value1=NOW&filtercombine=and)

Issue #4257, opened 2019-01-24
TpetraCore_BlockCrsMatrix_MPI_4 failing in ATDM cuda builds


fryeguy52:
CC: @trilinos/tpetra, @kddevin (Trilinos Data Services Product Lead), @bartlettroscoe, @fryeguy52
 
## Next Action Status

With the merge of PR #4307 on to 'develop' on 2/4/2018, the test `TpetraCore_BlockCrsMatrix_MPI_4` seems to be passing in all of the ATDM Trilinos builds on 2/5/2018.  Next: Get PR #4326 merged which re-enables this test in the Trilinos CUDA PR build ...

## Description
 
As shown in [this query](https://testing.sandia.gov/cdash-dev-view/queryTests.php?project=Trilinos&filtercount=5&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm&field2=testname&compare2=61&value2=TpetraCore_BlockCrsMatrix_MPI_4&field3=status&compare3=62&value3=Passed&field4=buildstarttime&compare4=83&value4=2019-01-24&field5=buildstarttime&compare5=84&value5=2019-01-25) the test:
* TpetraCore_BlockCrsMatrix_MPI_4

is failing in the builds:
* Trilinos-atdm-waterman-cuda-9.2-debug
* Trilinos-atdm-waterman-cuda-9.2-opt
* Trilinos-atdm-waterman-cuda-9.2-release-debug
* Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-debug
* Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-release
* Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-release-debug

It is failing with the following output:
```
p=0: *** Caught standard std::exception of type 'std::logic_error' :
 
  /home/jenkins/white/workspace/Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-debug/SRC_AND_BUILD/Trilinos/packages/tpetra/core/src/Tpetra_Experimental_BlockCrsMatrix_def.hpp:2825:
  
  Throw number = 1
  
  Throw test that evaluated to true: numBytesOut != numBytes
  
  unpackRow: numBytesOut = 4 != numBytes = 156.
 [FAILED]  (0.0877 sec) BlockCrsMatrix_double_int_int_Kokkos_Compat_KokkosCudaWrapperNode_write_UnitTest
[...]

PR #4326, opened 2019-02-05
Revert "Disable Tpetra test for CUDA PR build"

bartlettroscoe:
CC: @trilinos/framework 

Reverts trilinos/Trilinos#4293

As per https://github.com/trilinos/Trilinos/issues/4257#issuecomment-460704262, this test should now be passing in all of the CUDA builds.

As this test is protecting ATDM, I added the `ATDM` label.  It is important that the PR builds protect the ATDM builds to the greatest extent they can.


mhoemmen:
Ross, please feel free to mark this automerge if you think it's ready -- thanks! :D

PR #4322, opened 2019-02-05
WIP: Tpetra: Fix DistObject LID array DualView semantics

mhoemmen:
@trilinos/tpetra This is a work in progress; please do NOT merge!

PR #4328, opened 2019-02-05
Tpetra::DistObject::doTransferNew: Don't recreate DualViews for LID arrays each call

mhoemmen:
@trilinos/tpetra 

1. Export and Import now store `{permuteTo, permuteFrom, remote, export}LIDs` arrays as `Kokkos::DualView`, rather than as `Teuchos::Array`.

2. `DistObject::doTransferNew` no longer allocates new `Kokkos::DualView` from host `Teuchos::ArrayView` for the aforementioned four LIDs arrays.  This should avoid a major overhead that @crtrott and I noticed last Friday in the Tpetra CG solve benchmark on 2 GPUs on waterman.

3. Export and Import had duplicated code (for data access) and state (e.g., `out_` and `verbose_`).  I factored out duplication into their common base class, Transfer.

4. I made verbose debugging output (`TPETRA_VERBOSE`) more consistent.

5. I fixed CUDA build errors in https://github.com/trilinos/Trilinos/pull/4322 and rebased on top of current develop, in order to get @kyungjoo-kim 's BlockCrs fixes.  The above changes address part of Kyungjoo's concerns about DistObject taking `const DualView<const T>&` but sync'ing it.  (We still need to change `const DualView&` to `DualView&` in the case where the DistObject subclass will certainly _want_ to sync, but now the four LID arrays mentioned above will never need sync'ing.)

I still want to test downstream, but this is more ready for review now.


mhoemmen:
I'm still working on this w.r.t. BlockCrs, but it's almost done.

mhoemmen:
It's weird, I see the same error in BlockCrsMatrix as #4257.  Not sure what's going on.  This is an opt static build like what SPARC uses.  I will experiment more.

mhoemmen:
Huh, guess I need to add some includes.

mhoemmen:
Also, same deal with BlockCrs in the (new!) CUDA PR test:
```
 
  .../Trilinos/packages/tpetra/core/src/Tpetra_Experimental_BlockCrsMatrix_def.hpp:2833:
  
  Throw number = 1
  
  Throw test that evaluated to true: numBytesOut != numBytes
  
  unpackRow: numBytesOut = 4 != numBytes = 156.
```

mhoemmen:
@trilinos/framework The CUDA PR build is posting run-time warnings:
```
Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
                                  without setting CUDA_MANAGED_FORCE_DEVICE_ALLOC=1 or 
                                  setting CUDA_VISIBLE_DEVICES.
                                  This could on multi GPU systems lead to severe performance
                                  penalties.
```

kyungjoo-kim:
@mhoemmen I am not sure how the error comes back after it is fixed. The previous PR #4307 fixes the case where ``this->import_`` is not synced. Probably when you merge the PR, something is not synced correctly and it does not include the line 1908, DistObject_def.hpp.

mhoemmen:
@kyungjoo-kim Thanks for the hint -- I will look carefully at the sync state of `imports_`.  When you say, "does not include the line 1908, DistObject_def.hpp," does that mean that my changes somehow clobbered yours?

kyungjoo-kim:
@mhoemmen I mean the files modified in the PR #4307 
https://github.com/trilinos/Trilinos/pull/4307/files

mhoemmen:
@kyungjoo-kim wrote:
> How do other tests pass without host sync of imports?  Does it mean that other classes actual unpack from devices and only block crs unpacks the buffer from host?

Some classes actually do unpack on device.  Classes that don't will sync to host manually if needed.

It's a bit more complicated than that.  For example, CrsMatrix can only unpack on device if the data live on device.  (They don't if it's DynamicProfile.)  MultiVector actually runs the unpack kernel where `imports` was last modified, either on device or on host.  CooMatrix always unpacks on host.

mhoemmen:
@kyungjoo-kim btw I just verified that the BlockCrs test fails exactly because `imports` wasn't sync'd to host.  Now I know what to do :-)

mhoemmen:
@kyungjoo-kim wrote:
> yes... that is why I like the way to sync all "input" dualviews before they are used. Then, the derived classes have a choice about which buffer to be used.

That makes perfect sense for `permuteToLIDs`, `permuteFromLIDs`, `remoteLIDs`, and `exportLIDs`, since they depend only on the Export / Import object.  That's how I implemented this PR.  However, for the actual pack and unpack buffers, it would be nice to avoid the extra sync to host if the class knows how to unpack on device.

I agree with your comments that `const DualView<const T*, ...>&` is totally the wrong type to convey the idea that the subclass is responsible for sync'ing.  `DualView<const T*, ...>` forbids syncs at run time, so it looks like `DualView<T*, ...>` is the only way we can express the desired behavior:

  - Subclass of DistObject must sync where needed
  - Don't reallocate (since caller won't see it)

There's no way to forbid the subclass of DistObject from modifying the entries of `imports`, but there's actually no reason to forbid this anyway, so `DualView<T*, ...>` is actually what we want.

mhoemmen:
@trilinos/tpetra This PR is ready for review now.  @kyungjoo-kim brought up a useful point, that the DualView arguments to some of the DistObject methods have the wrong type, given the expectation that DistObject subclasses (not DistObject itself) must sync e.g., `imports` and `numImportPacketsPerLID` as needed.  I could include that in this PR if y'all like.

mhoemmen:
Issue #4344 talks about "DualView not having the right type" in detail.

jwillenbring:
I ended the most recent PR tests manually to fix the testing configuration, and we will rerun after it is fixed.

mhoemmen:
@trilinos/framework Weird, it looks like PR tests didn't pick up the latest commits.  I'll try a retest.

mhoemmen:
@trilinos/framework I noticed that when I click on the above "CDash Test Results for PR# 4328" link, I see old PR tests (with build failures) as well as new ones (without).  I'm attaching a screenshot.  Later pages of PR tests have the tests that passed.

trilinos-dashboard-screenshot

mhoemmen:
@vbrunini says this was quite helpful.

Issue #2749, opened 2018-05-14
Teuchos: Add .data() method to Array*, for std::vector and Kokkos::View compatibility


mhoemmen:
`std::vector` and `Kokkos::View` both have a `.data()` instance method, that returns a raw pointer to the object's data.  Teuchos memory management classes have a `.getRawPtr()` method that returns the same thing.  It would be nice if `Teuchos::Array`, `Teuchos::ArrayRCP`, and `Teuchos::ArrayView` also had a `.data()` method.  That would make them compile-time interchangeable with a lot of code that uses `std::vector`.

@trilinos/teuchos

bartlettroscoe:
Sounds good to me.  We just need to add some unit tests.  It would not take much.  You basically have `::data()` call `::getRawPtr()` and then add a few unit tests for each class for the major use cases (is null, is non-null, etc.).

mhoemmen:
@bartlettroscoe Thanks!  Yup, that's the idea.

PR #4331, opened 2019-02-06
zoltan2:  fixed ArrayView usage in graph metrics' directory calls

kddevin:
@trilinos/zoltan2 

## Description
When some processors have no IDs to partition or no graph edges, the graph metrics calculation core-dumped or caused a hang.  The culprit was attempts to compute the address of element 0 of an ArrayView of length zero.  I added tests for zero-length arrays before the attempted accesses.

## Motivation and Context
Test Zoltan2_zoltanCompare.exe failed in the metric computation for input graph nograph (which doesn't have edges).  These changes allow Zoltan2_zoltanCompare.exe to succeed.

## Related Issues
* Related to   #1533 

## How Has This Been Tested?
New tests are added that exercise the graph metrics in these conditions:
chacoGraphMetricTestNoGraph.xml -- tests graph metric computation on graph with no edges
chacoGraphMetricTestNoGraphNoDistribute.xml -- tests on graph with no edges and all input on rank 0
chacoGraphMetricTestNoDistribute.xml -- tests on graph with all input on rank 0
chacoGraphMetricTest.xml -- tests on graph with edge weights

## Checklist

- [x ] I have added tests to cover my changes.

### DIFF
--- a/packages/zoltan2/src/util/Zoltan2_GraphMetricsUtility.hpp
+++ b/packages/zoltan2/src/util/Zoltan2_GraphMetricsUtility.hpp
@@ -165,10 +165,19 @@ void globalWeightedByPart(
     typedef Zoltan2_Directory_Simple<part_t,lno_t,gno_t> directory_t;
     int debug_level = 0;
     directory_t directory(comm, bUseLocalIDs, debug_level);
-    directory.update(localNumVertices, &Ids[0], NULL, &parts[0],
-      NULL, directory_t::Update_Mode::Replace);
-    directory.find(localNumEdges, &edgeIds[0], NULL, &e_parts[0],
-      NULL, NULL, false);
+    if (localNumVertices)
+      directory.update(localNumVertices, &Ids[0], NULL, &parts[0],
+        NULL, directory_t::Update_Mode::Replace);
+    else
+      directory.update(localNumVertices, NULL, NULL, NULL,
+        NULL, directory_t::Update_Mode::Replace);
+
+    if (localNumEdges)
+      directory.find(localNumEdges, &edgeIds[0], NULL, &e_parts[0],
+        NULL, NULL, false);
+    else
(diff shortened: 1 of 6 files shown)


mhoemmen:
`Teuchos::ArrayView` has a `.getRawPtr()` method.  See also https://github.com/trilinos/Trilinos/issues/2749.

kddevin:
@mhoemmen From #4234, is it now safe to assume getRawPtr() returns NULL if a zero-length ArrayView is created?  The documentation says it returns NULL only if unsized, not if the specified size is zero.
https://trilinos.org/docs/dev/packages/teuchos/doc/html/classTeuchos_1_1ArrayView.html#a690d98de2e5ad95cfd11bb0a321485bd

mhoemmen:
@kddevin wrote:
> From #4234, is it now safe to assume getRawPtr() returns NULL if a zero-length ArrayView is created?

If you create an ArrayView from a raw pointer and a length, yes.  I would prefer that code not assume this, however, because it is _not_ true for `std::vector` or other C++ Standard Library classes.  Begin and end iterators always make a valid range, but a valid empty range is `[p, p)` for any possibly nonnull pointer `p`.

kddevin:
@mhoemmen Thanks.  It is probably best, then, to leave the code as-is.

mhoemmen:
@MicheldeMessieres Yup, Intel license server is down.  Framework is working on it.

mhoemmen:
@MicheldeMessieres wrote:
> But having to special case the 0 handling is also awkward.

I did fix `Teuchos::ArrayView` so that you can create it from an `std::vector` directly, without needing you to make the pointer null if the length is null.  (ArrayView now does that check for you.  I retained the existing behavior, that the pointer is always null if the length is null.  C++ Standard Library classes like `std::vector` and `std::span` do _not_ promise this, though, so be careful.)

PR #4332, opened 2019-02-06
Teuchos::{Array,ArrayView} Fix #2749

### LINKED ISSUE #2749: Teuchos: Add .data() method to Array*, for std::vector and Kokkos::View compatibility

mhoemmen:
@trilinos/teuchos 

`Teuchos::Array` and `Teuchos::ArrayView` now have a `.data()` method, for
compatibility with `std::vector` and other C++ Standard Library classes
(as of C++11), as well as `Kokkos::View`.

I took the liberty to use `std::vector::data` in the implementation of
`Teuchos::Array` instead of `&vec[0]`, and to replace a few instances of `0`
(in a pointer context) with `nullptr`.  Note that the Teuchos Memory 
Management Classes always return `nullptr` if the array length is zero;
`std::vector` does _not_ promise that.

By analogy with `std::shared_ptr<T[]>`, I did _not_ add a `.data()` method
to `Teuchos::ArrayRCP`.  The whole point of the new methods is to
increase compatibility with the C++ Standard Library.  Here is a table
of correspondences between Teuchos Memory Management Classes and C++
Standard Library classes:

Teuchos class      | Standard Library class
----------------- | --------
`Array`     | `std::vector`
`ArrayView` | `std::span` (C++20 draft)
`ArrayRCP`  | `std::shared_ptr<T[]>`

## Related Issues

* Closes #2749

### DIFF
--- a/packages/teuchos/core/src/Teuchos_Array.hpp
+++ b/packages/teuchos/core/src/Teuchos_Array.hpp
@@ -407,8 +407,18 @@ public:
   /** \brief Return a raw pointer to beginning of array or NULL if unsized. */
   inline T* getRawPtr();
 
+  /// \brief Return a raw pointer to beginning of array.
+  ///
+  /// Same semantics as \c getRawPtr (which see).
+  inline T* data();
+
   /** \brief Return a const raw pointer to beginning of array or NULL if unsized. */
   inline const T* getRawPtr() const;
+
+  /// \brief Return a const raw pointer to beginning of array.
+  ///
+  /// Same semantics as \c getRawPtr (which see).
+  inline const T* data() const;
 
   //@}
   /** \name Conversions to and from std::vector. */
@@ -1364,16 +1374,26 @@ bool Array<T>::hasBoundsChecking()
 template<typename T> inline
 T* Array<T>::getRawPtr()
 {
-  return ( size() ? &(*this)[0] : 0 );
+  return ( size() ? &(*this)[0] : nullptr );
 }
 
+template<typename T> inline
+T* Array<T>::data()
+{
+  return ( size() ? &(*this)[0] : nullptr );
+}
 
 template<typename T> inline
 const T* Array<T>::getRawPtr() const
 {
-  return ( size() ? &(*this)[0] : 0 );
+  return ( size() ? &(*this)[0] : nullptr );
 }
 
+template<typename T> inline
+const T* Array<T>::data() const
+{
+  return ( size() ? &(*this)[0] : nullptr );
+}
 
 // Conversions to and from std::vector
 
--- a/packages/teuchos/core/src/Teuchos_ArrayView.hpp
+++ b/packages/teuchos/core/src/Teuchos_ArrayView.hpp
@@ -140,7 +140,7 @@ template<class T> inline
 ArrayView<T>::ArrayView(
   std::vector<typename ConstTypeTraits<T>::NonConstType>& vec
   )
-  : ptr_( vec.empty() ? 0 : &vec[0] ), size_(vec.size())
+  : ptr_( vec.empty() ? 0 : vec.data() ), size_(vec.size())
 {
   setUpIterators();
 }
@@ -149,7 +149,7 @@ template<class T> inline
 ArrayView<const T>::ArrayView(
   std::vector<typename ConstTypeTraits<T>::NonConstType>& vec
   )
-  : ptr_( vec.empty() ? 0 : &vec[0] ), size_(vec.size())
+  : ptr_( vec.empty() ? 0 : vec.data() ), size_(vec.size())
 {
   setUpIterators();
 }
@@ -159,7 +159,7 @@ template<class T> inline
 ArrayView<T>::ArrayView(
   const std::vector<typename ConstTypeTraits<T>::NonConstType>& vec
   )
-  : ptr_( vec.empty() ? 0 : &vec[0] ), size_(vec.size())
+  : ptr_( vec.empty() ? 0 : vec.data() ), size_(vec.size())
 {
   setUpIterators();
 }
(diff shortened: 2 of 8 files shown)


packages/teuchos/core/src/Teuchos_Array.hpp
```diff
@@ -407,9 +407,19 @@ class Array
   /** \brief Return a raw pointer to beginning of array or NULL if unsized. */
   inline T* getRawPtr();
 
+  /// \brief Return a raw pointer to beginning of array.
```
bartlettroscoe: @mhoemmen, does Doxygen render this comment correctly using `///`?

packages/teuchos/core/src/Teuchos_ArrayView.hpp
```diff
@@ -315,13 +315,26 @@ T* ArrayView<T>::getRawPtr() const
   return ptr_;
 }
 
+template<class T> inline
+T* ArrayView<T>::data() const
+{
+  debug_assert_valid_ptr();
```
bartlettroscoe: To avoid duplication, why not just inline call `this->getRawPtr()`?

packages/teuchos/core/src/Teuchos_ArrayView.hpp
```diff
@@ -315,13 +315,26 @@ T* ArrayView<T>::getRawPtr() const
...
 const T* ArrayView<const T>::getRawPtr() const
 {
   debug_assert_valid_ptr();
   return ptr_;
 }
 
+template<class T> inline
+const T* ArrayView<const T>::data() const
+{
+  debug_assert_valid_ptr();
```
bartlettroscoe: Same deal here. To void duplication, why not just call `this->getRawPtr()`?

packages/teuchos/core/src/Teuchos_Array.hpp
mhoemmen: @bartlettroscoe I've had success doing that with Tpetra.  Thanks for asking, though!

packages/teuchos/core/src/Teuchos_ArrayView.hpp
mhoemmen: @bartlettroscoe I wanted to keep the function call stack short so the compiler has a chance at inlining :-)

packages/teuchos/core/src/Teuchos_ArrayView.hpp
mhoemmen: See above.

mhoemmen:
Thanks Ross!

packages/teuchos/core/src/Teuchos_ArrayView.hpp
bartlettroscoe: @mhoemmen said:

> I wanted to keep the function call stack short so the compiler has a chance at inlining :-)

Okay, I can accept that.  Performance and  maintainability are often at odds. This is pretty minor duplication.  But it is fragile to someone removing the call to `debug_assert_valid_ptr()` and there are no unit tests that ensure that `data()` throws in debug mode for null or invalid pointers (like dangling references).  So technically, there is now a lack of testing.

packages/teuchos/core/test/MemoryManagement/ArrayView_UnitTests.cpp
```diff
@@ -334,6 +350,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( ArrayView, danglingView_rcp_std_vector, T )
   }
 #ifdef TEUCHOS_DEBUG
   TEST_THROW(av.getRawPtr(), DanglingReferenceError);
+  TEST_THROW(av.data(), DanglingReferenceError);
```
mhoemmen: @bartlettroscoe Here's the unit test that `.data()` does the same debug-mode dangling reference testing that `.getRawPtr()` does.

packages/teuchos/core/src/Teuchos_ArrayView.hpp
mhoemmen: @bartlettroscoe Not true -- line 353 of `teuchos/core/test/MemoryManagement/ArrayView_UnitTests.cpp` tests that `.data()` throws in debug mode on dangling reference.

packages/teuchos/core/test/MemoryManagement/ArrayView_UnitTests.cpp
bartlettroscoe: Okay, I missed that.  Thanks for adding that!   I think we are 100% good then.

packages/teuchos/core/src/Teuchos_ArrayView.hpp
bartlettroscoe: Thanks!  I missed that looking over the tests.

packages/teuchos/core/test/MemoryManagement/ArrayView_UnitTests.cpp
bartlettroscoe: Okay, there is the dangling reference check

bartlettroscoe: But there are two `data()` functions (const and a non-const function).  Do both get tested?

mhoemmen: @bartlettroscoe I think only one gets tested.  If you like, I can delay this PR and push another commit that fixes this.

bartlettroscoe: > I think only one gets tested. If you like, I can delay this PR and push another commit that fixes this.

@mhoemmen, no need for delay.  I think this missing test did not show in the 100% coverage because all lines of code are covered but that does not technically guarantee full coverage of behavior. 

Thanks for adding these tests.

Cross fingers the PR tester will not fail again.

PR #4349, opened 2019-02-08
New scripts for getting jenkins to clean up (somewhat)

prwolfe:
## Description
These scripts will do a few things to improve jenkins
space usage.

'clean_workspace --force %s' will clean the build products
  from the %s directory but not remove the clones. I expect
   this to be useful after a job runs

'clean_workspace %s' will lok up the last time this workspace
   was cleaned, compare that to the last time a clean request
   was issued, and call make clean in the passed diretory if
   needed. We can place this at the start of the job and have it
   always check.

'clean_all_jobs' will just reset the date that the clean_workspace
   jobs use. I expect we will create a jenkins job to be launched
   by hand that does this.

Still needed is a way to remove all workspaces that are not
   actively being built - clones and all. I have seen a few
   groovy scripts that will do this, but those must be run as
   the jenkins adminstrator.

## How Has This Been Tested?
See unit tests in the commit - still needs to be hand tested.

### DIFF
--- a/commonTools/framework/clean_workspace/Modules.py
+++ b/commonTools/framework/clean_workspace/Modules.py
@@ -1,0 +1,205 @@
+#!/usr/bin/env python
+# -*- mode: python; py-indent-offset: 4; py-continuation-offset: 4 -*-
+# pylint: disable=wrong-import-position
+# pylint: disable=relative-import
+"""Fuctions for managing Environment Modules in Sierra python scripts."""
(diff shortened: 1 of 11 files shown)


prwolfe:
Jim - I did try to take all this into account. We can set it up so incremental builds always happen and the workspaces are only cleaned when we request it. I will be testing that before I add this to all the jobs of course.

PR #4356, opened 2019-02-08
cmake/std/PullRequestLinuxDriver-Test.sh: use environment to get job …

prwolfe:
…weight

Previously hard coded to 29 - still defaults to that if not set

I am using this in my testing and it seems to work fine.  Merging
it will not cause any change if the jobs are not modified.

## Motivation and Context
This will allow us to set some jobs at a higher weight and still utilize those core for the build.


prwolfe:
I would like to get this merged so I can work on it Friday.

PR #4365, opened 2019-02-11
ShyLU/FROSch: use Muelu in Laplace problem

### LINKED ISSUE #4208: ShyLu/FROSch: improve run-time error in case of missing packages
@trilinos/shylu 
@searhein 

## Expectations
Provide a more meaningful error message in case that a certain (optional) Trilinos package has been chosen as subdomain solver, but is not included into the build configuration.

## Current Behavior
I'm attempting to use MueLu as coarse solver for FROSch's GDSW preconditioner. This requires the cmake option `-D ShyLU_DDFROSch_ENABLE_MueLu:BOOL=ON`. However, without this option the subdomain solver is not recognized and the error message is just `FROSCH_ASSERT(false,"SolverType unknown...");`. 

Taking MueLu as an example as implemented [here](https://github.com/trilinos/Trilinos/blob/develop/packages/shylu/shylu_dd/frosch/src/Tools/FROSch_SubdomainSolver_def.hpp#L123): 
```
#ifdef HAVE_SHYLU_DDFROSCH_MUELU
    } else if (!ParameterList_->get("SolverType","Amesos").compare("MueLu")) {
        // ... calling MueLu setup here ...
    }
#endif
    else {
        FROSCH_ASSERT(false,"SolverType unknown...");
    }
```

The check, whether `"MueLu"` has been set as `"SolverType"` in the parameter list is already guarded by `#ifdef HAVE_SHYLU_DDFROSCH_MUELU`, so we can't tell the user that MueLu is required, but not enabled by the build configuration.

## Motivation and Context
Telling the user that a package is required, but has not been enabled in the build configuration helps to figure out, why an error occurs and points to possible solutions.

## Definition of Done
- [ ] Improve error message such that it contains details about missing packages.

## Possible Solution
We could move the `#ifdef HAVE_SHYLU_DDFROSCH_MUELU` inside the parameter check. This detects that MueLu has been asked for, but is not enabled. We can throw an error then that contains much more details. Maybe something along the lines of:
```
    } else if (!ParameterList_->get("SolverType","Amesos").compare("MueLu")) {
    #ifdef HAVE_SHYLU_DDFROSCH_MUELU
        // ... calling MueLu setup here ...
    #else
        FROSCH_ASSERT(false, "SolverType is chosen as MueLu, but MueLu is not enabled to work with FROSch. Modify your build configuration to enable MueLu inside ShyLU_DDFROSch!
    #endif
    }
    else {
        FROSCH_ASSERT(false,"SolverType unknown...");
    }
```

## Steps to Reproduce
1. Disable MueLu inside FROSch by setting `-D ShyLU_DDFROSch_ENABLE_MueLu:BOOL=OFF` in the configure script
1. Set the coarse solver's `"SolverType"` to `"MueLu"`
1. Run any test based on `ShyLU_DDFROSch_thyra_xpetra_laplace.exe`.

## Additional Information
I described the issue for MueLu here, but a similar strategy could be applied to all other package dependencies, namely Amesos, Amesos2, and Belos.

mayrmt:
@trilinos/shylu 

## Description
- Laplace problem: Adding tests to use MueLu as a coarse level solver within FROSch's GDSW
- Adding a more thorough error check to address #4208 
- Stokes problem: read file name of HDF5-file from the command line

## Motivation and Context
We want to try MueLu as coarse level solver for FROSch's GDSW. This starts with Laplace, but we aim at Stokes (or similar) problems in the end.

## Related Issues

* Closes #4208  

## How Has This Been Tested?
I added an Epetra and a Tpetra test for the Laplace problem. Both pass locally.

## Checklist

### DIFF
--- a/packages/shylu/shylu_dd/frosch/src/Tools/FROSch_SubdomainSolver_decl.hpp
+++ b/packages/shylu/shylu_dd/frosch/src/Tools/FROSch_SubdomainSolver_decl.hpp
@@ -135,40 +135,94 @@ namespace FROSch {
         typedef Teuchos::RCP<MueLu::Hierarchy<SC,LO,GO,NO> > MueLuHierarchyPtr;
 #endif
 
+        /*!
+        \brief Constructor
+
+        @param k Matrix
+        @param parameterList Parameter list
+        @param blockCoarseSize
+        */
         SubdomainSolver(CrsMatrixPtr k,
                         ParameterListPtr parameterList,
                         GOVecPtr blockCoarseSize=Teuchos::null);
 
+        //! Destructor
         virtual ~SubdomainSolver();
 
+        //! Initialize member variables
         virtual int initialize();
 
+        /*!
+        \brief Compute/setup this operator/solver
+
+        \pre Routine initialize() has been called and #isInitialized_ is set to \c true.
+
+        @return Integer error code
+        */
         virtual int compute();
 
-        // Y = alpha * A^mode * X + beta * Y
+        /*! \brief Apply subdomain solver to input \c x
+         *
+         * y = alpha * A^mode * X + beta * Y
+         *
+         * \param[in] x Input vector
+         * \param[out] y result vector
+         * \param[in] mode
+         */
+
+        /*!
+        \brief Computes the operator-multivector application.
+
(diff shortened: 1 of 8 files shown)


mayrmt:
Adding the `AT:WIP` label since I first have to look into the test failures more closely.

PR #4372, opened 2019-02-11
Change name of branches from 'atdm-develop-nightly' to 'atdm-nightly' (TRIL-260)

bartlettroscoe:
CC: @fryeguy52, @mhoemmen 

## Description

The reasons for making this name change are:

* Avoid these branches coming up in GitHub PR creation when looking to change
  from 'master' to 'develop'.

* Allow for more flexibility in case in the future we want to update from
  'master' instead of 'develop'.

* Shorten the names of the branches because, why not.

## Motivation and Context

See above.


bartlettroscoe:
Thanks Mark!

Issue #4334, opened 2019-02-06
Albany build failing in MueLu: AggregationStructuredAlgorithm_kokkos


ibaned:
@trilinos/muelu 

## Current Behavior

```
.../Trilinos/packages/muelu/src/Graph/StructuredAggregation/MueLu_StructuredAggregationFactory_kokkos_def.hpp:221:39: error: unknown type name 'AggregationStructuredAlgorithm_kokkos'; did you mean 'AggregationStructuredAlgorithm'?

      myStructuredAlgorithm = rcp(new AggregationStructuredAlgorithm_kokkos());
                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                      AggregationStructuredAlgorithm
/.../Trilinos/packages/muelu/src/Interface/../Headers/MueLu_UseShortNamesOrdinal.hpp:36:80: note: 'AggregationStructuredAlgorithm' declared here
typedef MueLu::AggregationStructuredAlgorithm<LocalOrdinal,GlobalOrdinal,Node> AggregationStructuredAlgorithm;
                                                                               ^
```

lucbv:
@ibaned I will look into it, somehow this did not get picked up by the PR auto-tester.
Are you building using any special configuration?

ibaned:
Here is the full CMake command:
```
"/usr/bin/cmake" \
"-DBUILD_SHARED_LIBS:BOOL=ON" \
"-DCMAKE_BUILD_TYPE:STRING=RELEASE" \
"-DCMAKE_CXX_COMPILER:FILEPATH=/usr/lib64/openmpi/bin/mpicxx" \
"-DCMAKE_C_COMPILER:FILEPATH=/usr/lib64/openmpi/bin/mpicc" \
"-DCMAKE_Fortran_COMPILER:FILEPATH=/usr/lib64/openmpi/bin/mpif90" \
"-DCMAKE_INSTALL_PREFIX:PATH=/home/lcm/LCM/trilinos-install-serial-clang-release" \
"-DCMAKE_VERBOSE_MAKEFILE:BOOL=OFF" \
"-DTPL_ENABLE_MPI:BOOL=ON" \
"-DTPL_ENABLE_BinUtils:BOOL=OFF" \
"-DTPL_MPI_INCLUDE_DIRS:STRING=/usr/include/openmpi-x86_64" \
[...]

PR #4376, opened 2019-02-12
Tpetra: Fix #4344 & ETI-OFF build errors

### LINKED ISSUE #4344: Tpetra::DistObject: DualView arguments have wrong types
@trilinos/tpetra @kyungjoo-kim 

This concerns the following DistObject virtual methods that its subclasses must implement:

  - copyAndPermuteNew
  - packAndPrepareNew
  - unpackAndCombineNew

As of PR #4328, we have clarified how `DistObject::doTransferNew` expects these methods to behave, with respect to their DualView parameters.  Export and Import setup is responsible for ensuring that `permuteToLIDs`, `permuteFromLIDs`, `exportLIDs`, and `remoteLIDs` are sync'd on both host and device.  Thus, the above methods need not sync any of these four DualViews.  As a result, their type `const DualView<const LO*, buffer_device_type>&` is correct.  (`const DualView&` means you can't sync them; `DualView<const LO*, ...>` means you can't modify their entries.)

PR #4328 further clarifies that `packAndPrepareNew` may fill its `exports` and `numPacketsPerLID` DualView parameters wherever the subclass likes, either on host or device.  It is not responsible for sync'ing either of these two DualViews to any particular place; `DistObject::doTransferNew` must sync them wherever it needs them to be.  Similarly, `unpackAndCombineNew` may unpack received data (`imports` and `numPacketsPerLIDs`) wherever it wants, either on host or device, but is responsible for sync'ing them wherever it needs them.  `DistObject::doTransferNew` is not responsible for sync'ing them.

I justify the previous paragraph as follows:

  - Avoid unnecessary syncs to host, if the subclass knows how to unpack on device
  - Separation of concerns between communication and pack / unpack

This imposes the following requirements on `exports`, `imports`, and `numPacketsPerLIDs`:

  1. `DistObject::doTransferNew` allocates the `numPacketsPerLIDs` argument when it calls `packAndPrepareNew`, but `packAndPrepareNew` must allocate the `exports` output argument.  Thus, subclasses cannot reallocate `numPacketsPerLID`.
  2. `DistObject::doTransferNew` allocates the `imports` and `numPacketsPerLIDs` arguments when it calls `unpackAndCombineNew`; subclasses cannot reallocate them.
  3. Subclasses need to be able to sync any of these arrays wherever they like.
  4. `packAndPrepareNew` implementations access `exports` and `numPacketsPerLIDs` in write-only fashion.
  5. `unpackAndPrepareNew` implementations technically only need read-only access to `imports` and `numPacketsPerLIDs`.  However, `DistObject::doTransferNew` only accesses them in write-only fashion before calling `unpackAndPrepareNew`, and doesn't access them afterwards.  Thus, `unpackAndPrepareNew` implementations could freely write to these arrays, e.g., to use them as scratch space.

This means that the correct type of `imports` and `numPacketsPerLIDs` is `DualView<T*, ...>`.

  - Can't have `DualView<const T*, ...>`, because that would forbid sync'ing.
  - Can't have `const DualView&` or `const DualView`, because that would forbid sync'ing.
  - Can't have `DualView&`, because that would expose DistObject's internal DualViews to reallocation by subclasses.  Passing by value means that even if the subclass reallocates, the caller won't see that.  It's just like passing `double*` into a function, instead of `double*&`.

Currently, these arrays have type `const DualView<const T*, ...>&`.  This makes implementations do hacks in order to meet the above four requirements (esp. (2)).

I propose changing the types of `imports`, and `numPacketsPerLID` (as parameters to `packAndPrepareNew` and `unpackAndCombineNew`) from `const DualView<const T*, ...>&`, to `DualView<T*, ...>`.

mhoemmen:
@trilinos/tpetra @ikalash 

1. Fix #4344, and document DistObject to clarify what DualViews can and should be sync'd
2. Fix build errors that show up only with ETI off: https://github.com/trilinos/Trilinos/issues/4334#issuecomment-462417954

## Related Issues

* Closes #4344

### DIFF
--- a/packages/ifpack2/example/RelaxationWithEquilibration.cpp
+++ b/packages/ifpack2/example/RelaxationWithEquilibration.cpp
@@ -87,7 +87,7 @@ tpetraToEpetraComm (const Teuchos::Comm<int>& tpetraComm)
 template<class LO, class GO, class NT>
 Epetra_Map
 tpetraToEpetraMap (const Tpetra::Map<LO, GO, NT>& map_t,
-		   const Epetra_Comm& comm_e)
+                   const Epetra_Comm& comm_e)
 {
   const int gblNumInds = static_cast<int> (map_t.getGlobalNumElements ());
   const int lclNumInds = static_cast<int> (map_t.getNodeNumElements ());
@@ -133,17 +133,10 @@ getRankSafelyFromTpetraDistObject (const TpetraDistObjectType& X)
 template<class LO, class GO, class NT>
 std::pair<int, std::string>
 deep_copy (Epetra_Vector& X_e,
-	   const Tpetra::Vector<double, LO, GO, NT>& X_t)
+           const Tpetra::Vector<double, LO, GO, NT>& X_t)
 {
-  using vec_type = Tpetra::Vector<double, LO, GO, NT>;
-  using device_type = typename vec_type::device_type;
-  using memory_space = typename device_type::memory_space;
-  using host_execution_space =
-    typename Kokkos::View<double*, Kokkos::LayoutLeft,
-			  device_type>::HostMirror::execution_space;
-  using host_memory_space = Kokkos::HostSpace;
-  using host_device_type = Kokkos::Device<host_execution_space, host_memory_space>;
-  using host_view_type = Kokkos::View<double*, Kokkos::LayoutLeft, host_device_type>;
+  using host_view_type = Kokkos::View<double*, Kokkos::LayoutLeft,
+    Kokkos::HostSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged> >;
 
   int lclErrCode = 0;
 
@@ -154,7 +147,7 @@ deep_copy (Epetra_Vector& X_e,
     std::ostringstream errStrm;
     errStrm << "Proc " << myRank << ": deep_copy(Vector): X_t.getLocalLength()"
       " = " << X_t.getLocalLength () << " != map_e.NumMyElements() = "
-	    << lclNumRows << "." << std::endl;
+            << lclNumRows << "." << std::endl;
     return {lclErrCode, errStrm.str ()};
   }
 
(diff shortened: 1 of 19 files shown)


mhoemmen:
This is my bad -- I'll look more closely at this one.

mhoemmen:
I need to add a new commit to the PR to fix some issues, but otherwise this is almost done.

mhoemmen:
@trilinos/tpetra This should be ready now.

mhoemmen:
@tjfulle wrote:
> why change passing `DualView` by reference to passing by value in many places? (just curious)

That's the most correct expression of the semantics.

  - DistObject, the caller, does not permit callees to reallocate.  Hence, it does not pass by nonconst reference.
  - DistObject, the caller, _does_ permit callees to sync.  Hence, it cannot pass by const or const reference.  Furthermore, the DualView cannot be a view of const; it must be `DualView<T*, ...>`.  (DualView doesn't currently have a way to express "can't sync, but can't modify the entries."  Furthermore, in that case with DistObject, it's actually harmless for the callee to modify the entries.)

This only leaves passing the DualView by _nonconst_ _value_.

I'm imitating Ross' guide to the Teuchos Memory Management Classes, in particular the four pages of idioms.  The point is that the way that you pass the DualView exactly documents what callees can and cannot do with the DualView.  Readers can go through this elimination argument themselves to see what they can and cannot do.

mhoemmen:
@tjfulle btw thanks for the review! :D

mhoemmen:
@trilinos/framework I'm seeing stale test failures again :(

mhoemmen:
@trilinos/framework This is fixing a build error with ETI off.  Would you consider a manual merge?

mhoemmen:
@ikalash says this is blocking Albany's nightly testing against develop.

mhoemmen:
It got through, yay!  @ikalash , this should fix Albany build errors with ETI off using Trilinos develop.

I don't recommend ever turning ETI off, and I think enabling `GO=unsigned long` at configure time would make DTK work with ETI on, but we still want to support the ETI off build.

ikalash:
Thanks Mark!  I’ll look at whether we can get away with ETI on with DTK.

mhoemmen:
@ikalash You'll almost certainly need to set `Tpetra_INST_INT_UNSIGNED_LONG:BOOL=ON`.

ikalash:
@mhoemmen : I tried adding that option and it lets us build with code with ETI on + DTK.  This was not the case a few months ago - not sure what changed.  This is great news b/c we don't need to worry about a non-ETI build now, plus our code will build much faster.

mhoemmen:
@ikalash Awesome! :-D  It does mean that you might be building more of Trilinos with `GO = unsigned long` than you were before, so if you encounter any interesting errors, please let us know.

ikalash:
Ok, will do!

Issue #2008, opened 2017-11-21
Sacado: Derivative of std::pow produces floating point exception


xamhn:
@trilinos/sacado

## Expectations

For power-functions with variable exponent and variable base the partial derivative wrt to the base should also work for negative values of the evaluation point, i.e. e.g. df/dx | x=-1 should work. 

## Current Behavior

Given a function f = x^y. Then df/dx for x <= 0 produces a floating point exception. This is possibly because for a derivative df/d**y** the function is rewritten as f = exp(y*ln x) whose derivative df/d**x** obviously produces a floating point exception for x <= 0.

## Steps to Reproduce

Snippet 1 (with error):
```
  int num_deriv = 2;
  double base_d = -1.0;
  double expo_d = 2.0;
  Sacado::Fad::DFad<double> base(num_deriv,0,base_d);
  Sacado::Fad::DFad<double> expo(num_deriv,1,expo_d);
  Sacado::Fad::DFad<double> result;

  result = std::pow(base,expo);

  double val = result.val();
  std::cout << "value: " << val << std::endl;

  double deriv = result.dx(0);
  std::cout << "derivative: " << deriv << std::endl;
```
Produces:
```
[ 0]: sigfpe_handler(int)  (baci.cpp:50 (discriminator 6))
[ 1]: 
[ 2]: log
[ 3]: Sacado::Fad::Expr<Sacado::Fad::PowerOp<Sacado::Fad::Expr<Sacado::Fad::GeneralFad<double, Sacado::Fad::DynamicStorage<double, double> >, Sacado::Fad::ExprSpecDefault>, Sacado::Fad::Expr<Sacado::Fad::GeneralFad<double, Sacado::Fad::DynamicStorage<double, double> >, Sacado::Fad::ExprSpecDefault> >, Sacado::Fad::ExprSpecDefault>::fastAccessDx(int) const  (Sacado_Fad_Ops.hpp:643)
[...]

PR #4380, opened 2019-02-12
Sacado:  Add specializations so pow() only uses if_then_else for simd types

etphipp:
Related to issue #2008.  Recently Sacado's overloads of std::pow() were
changed to use if_then_else() to support use of Sacado with STK's SIMD
type, instead of the ternary operator to check when the base is equal to
0.  However when using if_then_else(), both branches are always
evaluated, and so a FPE can be generated even though the value isn't
used.  This changes the implementation by adding two specializations of
the PowerOp expression template, one for SIMD and one for everything
else, where only the SIMD one uses if_then_else().  This requires any
SIMD type that wants to pick up that implementation specialize
Sacado::IsSimdType<>.

@bartgol @vbrunini

### DIFF
--- a/packages/sacado/test/TestSuite/FadUnitTests.hpp
+++ b/packages/sacado/test/TestSuite/FadUnitTests.hpp
@@ -182,6 +182,8 @@ class FadOpsUnitTest : public CppUnit::TestFixture {
   CPPUNIT_TEST(testTimesLR);
   CPPUNIT_TEST(testDivideLR);
 
+  CPPUNIT_TEST(testPowConstB);
+
   CPPUNIT_TEST_SUITE_END();
 
 public:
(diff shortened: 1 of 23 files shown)


packages/sacado/test/TestSuite/FadUnitTests.hpp
```diff
@@ -318,7 +377,7 @@ class FadOpsUnitTest : public CppUnit::TestFixture {
 template <class FadType, class ScalarType>
 FadOpsUnitTest<FadType,ScalarType>::
 FadOpsUnitTest() :
-  urand(), n(5), tol_a(1.0e-15), tol_r(1.0e-14) {}
+  urand(), n(5), tol_a(1.0e-15), tol_r(1.0e-12) {}
```
rppawlo: Do the tests only work for double types? Should this be something like:
```
tol = std::numeric_limits<ScalarType>::epsilon() * 100.0;
```

etphipp: The tests are only instantiated on double.  Something like that probably would be better, although it is unclear what the factor (100 in this case) should be in general.

bartgol: Perhaps a `Teuchos::ScalarTraits<ScalarType>::magnitude_type`? That's if ScalarType is supported by Teuchos' scalar traits, of course. Though one could "easily" extend some of those traits to a SIMD type too.

PR #4387, opened 2019-02-12
Ifpack2: Fix #4343.

### LINKED ISSUE #4343: Ifpack2: BlockTriDi test does (bool > 1); probably meant just to eval the bool
@trilinos/ifpack2 

NVCC warned about the following line evaluating a `bool > 1` expression, that's always false. https://github.com/trilinos/Trilinos/blob/e476b66bcdd0fb2efa6018a4d5ff10f5bb02d2b9/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestBlockTriDiContainerUtil.hpp#L249

ambrad:
@trilinos/ifpack2 

## Description

Activate a test that was inactive due to a bug in the unit test.

## How Has This Been Tested?

I verified the test is now active and meaningfully tests norm-based termination.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -631,6 +631,7 @@ namespace Ifpack2 {
 #ifdef HAVE_IFPACK2_MPI
         for (local_ordinal_type i=0,iend=pids.recv.extent(0);i<iend;++i)
           MPI_Cancel(&reqs.recv[i]);
+        waitall(reqs.recv.size(), reqs.recv.data());
         waitall(reqs.send.size(), reqs.send.data());
 #endif
       }
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -629,8 +629,6 @@ namespace Ifpack2 {
 
       void cancel () {
 #ifdef HAVE_IFPACK2_MPI
-        for (local_ordinal_type i=0,iend=pids.recv.extent(0);i<iend;++i)
-          MPI_Cancel(&reqs.recv[i]);
         waitall(reqs.recv.size(), reqs.recv.data());
         waitall(reqs.send.size(), reqs.send.data());
 #endif
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -631,6 +631,7 @@ namespace Ifpack2 {
 #ifdef HAVE_IFPACK2_MPI
         for (local_ordinal_type i=0,iend=pids.recv.extent(0);i<iend;++i)
           MPI_Cancel(&reqs.recv[i]);
+        waitall(reqs.send.size(), reqs.send.data());
 #endif
       }
 
--- a/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestBlockTriDiContainerUtil.hpp
+++ b/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestBlockTriDiContainerUtil.hpp
@@ -286,12 +286,23 @@ struct BlockTriDiContainerTester {
           TEST_BR_BTDC_FAIL("FAIL: test_BR_BTDC (A = D + R) " << details << " rd " << rd);
         else
           TEST_BR_BTDC_SUCCESS("SUCCESS: test_BR_BTDC (A = D + R) " << details << " rd " << rd);
-        if ( ! T_bare.is_null()) {
+        { // Advanced options only the bare object supports.
+          auto T_bare_advanced = T_bare;
+          if (T_bare_advanced.is_null()) {
+            T_bare_advanced = make_BTDC(sb, sbp, A, overlap_comm, nonuniform_lines, jacobi,
+                                        seq_method);
+            T_bare_advanced->initialize();
+            T_bare_advanced->compute(T_bare_advanced->createDefaultComputeParameters());
+          }
           if ( ! seq_method) { // Test norm-based termination.
-            const int nits = apply(*B, *X_solve, true);
+            auto input = T_bare_advanced->createDefaultApplyParameters();
+            input.zeroStartingSolution = zero_starting;
+            input.maxNumSweeps = num_sweeps;
+            input.tolerance = tol;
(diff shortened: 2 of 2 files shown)


ambrad:
I'm looking into the comm failures. I haven't reproduced, but I pushed a commit that is logically correct according to my analysis and I believe at least targets the problem.

kyungjoo-kim:
I don't think you need to worry too much about it. It is just unlucky happening. Repeated retesting would make the PR pass.

ambrad:
No, I think it must mean something. The BlockTriDi test has never failed before, right? Or else we would get messages about it. I think reactivating the overlap-comm norm-based convergence test must be triggering a latent bug.

kyungjoo-kim:
From the PR tests, we see one GCC build and CUDA build are okay. If all host builds are failed, I agree that something might be wrong. However, the one successful GCC build make me hope that it is a merely temporal network issue.  ; ). 

I am glad if you find out the latent bug (if exist). However, I also worry that it might be difficult for you to track the problem after my modifications from your original version.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -631,6 +631,7 @@ namespace Ifpack2 {
 #ifdef HAVE_IFPACK2_MPI
         for (local_ordinal_type i=0,iend=pids.recv.extent(0);i<iend;++i)
           MPI_Cancel(&reqs.recv[i]);
+        waitall(reqs.send.size(), reqs.send.data());
```
mhoemmen: That's right, you have to wait on cancelled messages.

ambrad: Is it? Here I'm waiting on the sends, not the receives.

I believe MPI_Cancel, MPI_Waitany, and MPI_Waitall have the same semantics w.r.t. termination of a request. But I definitely could be wrong. Did some googling and still think a wait isn't needed on a cancelled request.

The motivation for waiting on the sends is that (1) that should have been done and I missed it; (2) I think the sends are not done when the next test, using the same BlockTriDi object (a good thing) starts, after the norm-based termination test. The motivation not to cancel the sends is that MPI_Cancel warns against it.

For clarity, all of this concerns the overlapped-comm code path, which in practice isn't being used.

kyungjoo-kim: I just confirmed that there is a comm issue in the overlapped-comm code path where norm based termination is on. I agree that it is not important and it is not used in practice. I am fine if we decide not to fix now.

ambrad: I'd like to fix it. I _think_ the above commit should. Since you have a reproducer, can you check if inserting the waitall line above fixes it? I couldn't reproduce it, so I don't know for sure that it fixes the problem.

kyungjoo-kim:
These test failures show some consistency here. Cuda and Serial build are okay and the others are not. The difference is that both Cuda and Serial has serial host space and others use openmp host space. Any clue from this ?

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: @ambrad MPI standard says: "It is still necessary to call `MPI_REQUEST_FREE`, `MPI_WAIT` or `MPI_TEST` (or any of the derived operations) with the cancelled request as argument after the call to `MPI_CANCEL`."

mhoemmen: @ambrad `MPI_WAIT` "behaves as a local function" (i.e., "is guaranteed to return, irrespective of the activities of other processes") if called on a cancelled thing.  Also, "if `MPI_TEST` is repeatedly called in a busy wait loop for a cancelled communication, then `MPI_TEST` will eventually be successful."

ambrad:
@kyungjoo-kim, you have a good point and might be right. But I want to finish up fixing the treatment of MPI_Cancel before I consider other possibilities. My guess right now is that the association with OpenMP is a red herring. By the way, don't let this PR block yours. You can merge your PR with the norm-based-termination, overlapped-comm test disabled, and then I'll merge this PR once I've nailed that issue.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
ambrad: A few things.

First, I'm speculating the OpenMP builds may be failing b/c there's processor imbalance, which makes MPI incorrectness exibit itself with greater likelihood.

Second, after finding a better source to read about MPI_Cancel, I think I want to remove it completely, and simply do
```
        waitall(reqs.recv.size(), reqs.recv.data());
        waitall(reqs.send.size(), reqs.send.data());
```
This will let the final round of comm complete, but nothing will be done with the data. The more I read about MPI_Cancel, the more I think it's not going to provide any speedup to cancel the receives. In any case, the speedup is minor, a fraction of one iteration at most.

Third, after reviewing the unit tests, I decided I want to test the advanced features in the general test (i.e., `{"general", false, false}, // Test the general case.`) in every parameterization of BlockTriDi. Currently, these do not get tested if BlockRelaxation wraps BlockTriDi, and this happens whenever the feature set to be tested falls within BlockRelaxation's feature set. So I'm adjusting the unit tester to _always_ test the advanced features. The unit tests are fast, so while this adds ~50% (very rough estimate), this is to a ~1-2s overall test time, which I view as insignificant.

ambrad:
@mhoemmen sorry about the churn on this PR. I want to keep the overlapped comm code path alive, even if it's not being used in practice right now, and I definitely want norm-based termination to be tested in the unit tests. Putting those two wants together has complicated this PR a bit.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: @ambrad That makes sense -- the MPI Standard is full of warnings about how `MPI_Cancel` isn't fast, etc.

kyungjoo-kim:
@ambrad I am always surprised at how quickly you can catch up all the codes. For me, it probably takes a few days to just make my brain warm to start work.

ambrad:
@kyungjoo-kim everything is working fine now. At your convenience, let me know how you would like to incorporate these changes. I can simply merge the PR, you could merge the PR into your branch, or you could copy the changes. Any approach is fine with me.

kyungjoo-kim:
@ambrad You can merge it.

ambrad:
@kyungjoo-kim ok, will do. I think if you "Approve" this PR, it will automerge now that you put the label on it (thanks).

Issue #4353, opened 2019-02-08
Ifpack2_unit_tests_MPI_4 randomly failing on ATDM waterman build


fryeguy52:
CC: @trilinos/ifpack2, @srajama1 (Trilinos Linear Solvers Product Lead), @bartlettroscoe, @fryeguy52
 
<Checklist>
<???: Add label "ATDM">
<???: Add label "bug"?>
<???: Add label for affected packages (e.g. "MueLu", "Tpetra", "Kokkos", etc.)>
<???: Add milestone "Initial cleanup of new ATDM builds of Trilinos" or "Keep promoted ATDM builds of Trilinos clean">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
 
## Next Action Status
<status-and-or-first-action>
 
## Description
 
As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercombine=and&filtercombine=&filtercombine=and&filtercombine=and&filtercount=5&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-waterman-cuda-9.2-opt&field2=testname&compare2=61&value2=Ifpack2_unit_tests_MPI_4&field3=site&compare3=61&value3=waterman&field4=buildstarttime&compare4=84&value4=2019-02-08T00%3A00%3A00&field5=buildstarttime&compare5=83&value5=2018-12-27T00%3A00%3A00) the test:
* Ifpack2_unit_tests_MPI_4

is randomly failing in the buils:
* Trilinos-atdm-waterman-cuda-9.2-opt

It has failed roughly 6 times in the last month.  Here are some examples of the output when it fails:

```
Error, relErr(Y.get1dView ()[9932],Z.get1dView ()[9932]) = relErr(29832,0) = 1 <= tol = 2.22045e-12: failed!
```

```
p=0 | The following tests FAILED:
p=0 |     48. Ifpack2OverlappingRowMatrix_default_scalar_type_default_local_ordinal_type_default_global_ordinal_type_Test0_UnitTest ... 
p=0 | 
p=0 | Total Time: 6.49 sec
p=0 | 
p=1 | Summary: total = 82, run = 82, passed = 81, failed = 1
p=1 | 
p=1 | End Result: TEST FAILED
```

## Current Status on CDash

[...]

PR #4393, opened 2019-02-13
Ifpack2::OverlappingRowMatrix::apply: Fix #4392

### LINKED ISSUE #4392: Ifpack2::OverlappingRowMatrix::apply is unnecessarily sequential
@trilinos/ifpack2 

It uses the old ArrayRCP host access interface of `Tpetra::MultiVector`, when it could just use `Tpetra::CrsMatrix::localApply` and get thread / CUDA parallelism for free.  This could also be related to #4353.

This will require making `Tpetra::CrsMatrix::localApply` public, but I'm OK with that; it's a useful function for implementing block operators efficiently.

## Related Issues
* Related to #4353

mhoemmen:
@trilinos/tpetra 

Fix #4392 by rewriting `Ifpack2::OverlappingRowMatrix::apply` to call `Tpetra::CrsMatrix::apply` directly for the two CrsMatrix objects in OverlappingRowMatrix, instead of rerolling `CrsMatrix::apply` in a slow and sequential way.  Fix potential issues with `Tpetra::MultiVector`'s offset view constructor that came up as a result of this fix.  This may also address #4353.

## Related Issues

* Closes #4392 
* Related to #4353

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -67,6 +67,7 @@
 #include "Kokkos_Random.hpp"
 #include "Kokkos_ArithTraits.hpp"
 #include <memory>
+#include <sstream>
 
 #ifdef HAVE_TPETRA_INST_FLOAT128
 namespace Kokkos {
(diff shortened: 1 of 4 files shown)


packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3295,34 +3296,64 @@ namespace Tpetra {
...
-#ifdef HAVE_TPETRA_DEBUG
-    const size_t strideBefore =
-      X.isConstantStride () ? X.getStride () : static_cast<size_t> (0);
-    const size_t lclNumRowsBefore = X.getLocalLength ();
-    const size_t numColsBefore = X.getNumVectors ();
-    const impl_scalar_type* hostPtrBefore =
-      X.getDualView().view_host().data ();
-#endif // HAVE_TPETRA_DEBUG
+    if (debug) {
+      reduceAll<int, int> (*comm, REDUCE_MIN, lclGood, outArg (gblGood));
```
bartlettroscoe: Since `blGood` is really an in/out arg, this might be better to use [`Teuchos::inoutArg()`](https://trilinos.org/docs/dev/packages/teuchos/doc/html/classTeuchos_1_1Ptr.html#acb9418d4e46ac7b589f8de2ab89bc56e) to make this more self-documenting.  Just a suggestion.

mhoemmen: @bartlettroscoe Actually, it's totally an output argument; its value on input doesn't matter.  `lclGood` is the input value here; their minimum over all processes in `*comm` is written to `gblGood`.

bartlettroscoe: @mhoemmen, ah, excellent!  It has been a long time since I looked at `reduceAll()`.  I thought that it took the initial value and reduced them all and put the result back in that object.  (Am I thinking about `broadcast()`?)  Now I see the `iclGood` argument.  Thanks!

jhux2:
___
*[packages/ifpack2/src/Ifpack2_OverlappingRowMatrix_decl.hpp, line 348 at r1](https://reviewable.io/reviews/trilinos/trilinos/4393#-LYc5iE1BxWHx-Pj-ILo:-LYc5iE2DAarP5_iYw0N:b-w7m4vz) ([raw file](https://github.com/trilinos/trilinos/blob/c2e203e1338c8689d0b8da471eb72e8a7a1e03c7/packages/ifpack2/src/Ifpack2_OverlappingRowMatrix_decl.hpp#L348)):*
> ```cpp
>   typedef Tpetra::Vector<scalar_type, local_ordinal_type, global_ordinal_type, node_type> vector_type;
> 
>   //! The input matrix to the constructor.
> ```

This comment is incorrect.

mhoemmen:
@jhux2 wrote:
> Why is `localApply` now public? Just curious.

The commit messages (and the Ifpack2 code) explain.  `Ifpack2::OverlappingRowMatrix::apply` wants to handle the Import operations itself.

mhoemmen:
> This comment is incorrect.

In what way?  `A_` is the result of dynamic-casting the input matrix `A`.

jhux2:
> In what way? `A_` is the result of dynamic-casting the input matrix `A`.

Yes.  A is a  `row_matrix`, although it must actually be a CrsMatrix.

mhoemmen:
> Yes. A is a `row_matrix`, although it must actually be a CrsMatrix.

That was true before -- I just moved the test up.  Thanks!

Issue #4038, opened 2018-12-12
Xpetra/Epetra: link error in panzer test for intel debug static


bathmatt:
When I try to link packages/panzer/adapters-stk/test/solver/PanzerAdaptersSTK_solver.exe I get a missing tpetra symbol for intel debug static, but intel opt static works..  No idea why.  This is with a RHEL7 machine using all sems and Ross' new build scripts.  

Here is the error message I see, no idea why the xpetra epetra map is looking for something tpetra??  

```
packages/xpetra/src/libxpetra.a(Xpetra_EpetraMap.cpp.o): In function `Tpetra::Details::LocalMap<int, long, Kokkos::Device<Kokkos::OpenMP, Kokkos::HostSpace> >::LocalMap()':

```

Here is my link line

```
[...]

Issue #4325, opened 2019-02-05
MueLu: why using `linAlgebra=Epetra` with unit_tests_kokkos?


lucbv:
@trilinos/muelu

## Expectations
It should be possible to use all of `kokkos` and `kokkos-kernels` in functions marked `_kokkos` in MueLu. Otherwise what is the point of the kokkos refactor?

## Current Behavior
`unit_tests_kokkos.exe` is called with both `linAlgebra=Tpetra` and `linAlgebra=Epetra`

## Motivation and Context
Running these kokkos specific unit tests with `linAlgebra=Epetra` pretty much guarantees a problematic behavior and a failing test...

## Definition of Done
MueLu team members need to discuss the need for the `linAlgebra=Epetra` test of the kokkos refactor path in MueLu, if there is a legitimate reason for this behavior we then also need to come up with a reasonable way to still allow kokkos refactor work to happen while this restriction is applied.

csiefer2:
I have no problem with banning Epetra matrices w/ Kokkos refactor... as long as the non-Kokkos code path still exists.

The problem comes when we try to eliminate the non-Kokkos code path AND still support Epetra.

jhux2:
@lucbv and I have discussed this offline.  For the life of me I can't remember what the rationale (if any) there was for testing Kokkos factories with Epetra.  I am in favor of dropping Epetra from these tests.

lucbv:
@jhux2 @csiefer2 I have created a PR to remove these tests (finally), I think that if we really wanted to run these type of tests we should do it in builds with `Tpetra=OFF` which of course would not work...
The only thing that was potentially tested here is that `Xpetra` can create Tpetra objects with an `EpetraNode`.

lucbv:
The merge of PR #5302 fixed this so I am closing the issue

PR #4400, opened 2019-02-14
Xpetra: fixed dependency on Tpetra instantations

brian-kelley:
Satisfy dependency of Xpetra EpetraMap on
Tpetra::Details::FixedHashTable with int/long and
int/longlong as LO/GO. These dependencies always exist
as long as Tpetra is enabled, but regardless of whether
these LO/GO combinations are enabled in Tpetra ETI.

@trilinos/xpetra 

## Description

Xpetra_EpetraMap.cpp now includes the def of
FixedHashTable so that these instantiations can be made, even if int/long
and/or int/longlong haven't been enabled in ETI. The include is only done if
Tpetra and Xpetra kokkos refactor are both enabled, because Xpetra::EpetraMap
only uses FixedHashTable in this case.

## Motivation and Context

This fixes an "undefined symbol" error during the final link of Albany (or any other app executable). Without this code change, the only workaround is to enable both TPETRA_INST_INT_LONG and TPETRA_INST_INT_LONG_LONG in configuration (huge code size increase). Only the instantiation of FixedHashTable is needed, not all of Tpetra.

* Related to #4038 

## How Has This Been Tested?

Tested by successfully building Trilinos/Albany without TPETRA_INST_INT_LONG.

## Checklist

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Cuda.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Cuda.cpp
@@ -79,6 +79,14 @@ namespace Details {
 
   TPETRA_INSTANTIATE_L( TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT_CUDA_INT )
 
+  // Make sure that KeyType = long, ValueType = int and
+  // KeyType = long long, ValueType = int both get instantiated -
+  // whenever Tpetra is enabled, Xpetra::Epetra_Map depends on both
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, cuda_device_type)
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, cuda_uvm_device_type)
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long long, int, cuda_device_type)
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long long, int, cuda_uvm_device_type)
+
   // FIXME (mfh 26 Sep 2015) Once it becomes possible to disable LO =
   // int, add an instantiation here for {KeyType = LO, ValueType =
   // LO}.  However, this case has likely already been covered above,
--- a/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_OpenMP.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_OpenMP.cpp
@@ -71,6 +71,12 @@ namespace Details {
 
   TPETRA_INSTANTIATE_L( TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT_OPENMP_INT )
 
+  // Make sure that KeyType = long, ValueType = int and
+  // KeyType = long long, ValueType = int both get instantiated -
+  // whenever Tpetra is enabled, Xpetra::Epetra_Map depends on both
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, openmp_device_type)
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long long, int, openmp_device_type)
+
   // FIXME (mfh 26 Sep 2015) Once it becomes possible to disable LO =
   // int, add an instantiation here for {KeyType = LO, ValueType =
   // LO}.  However, this case has likely already been covered above,
--- a/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Serial.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Serial.cpp
@@ -71,6 +71,12 @@ namespace Details {
 
   TPETRA_INSTANTIATE_L( TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT_SERIAL_INT )
 
+  // Make sure that KeyType = long, ValueType = int and
+  // KeyType = long long, ValueType = int both get instantiated -
+  // whenever Tpetra is enabled, Xpetra::Epetra_Map depends on both
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, serial_device_type)
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long long, int, serial_device_type)
+
   // FIXME (mfh 26 Sep 2015) Once it becomes possible to disable LO =
   // int, add an instantiation here for {KeyType = LO, ValueType =
   // LO}.  However, this case has likely already been covered above,
--- a/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Threads.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Threads.cpp
@@ -71,6 +71,12 @@ namespace Details {
 
   TPETRA_INSTANTIATE_L( TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT_PTHREAD_INT )
 
+  // Make sure that KeyType = long, ValueType = int and
+  // KeyType = long long, ValueType = int both get instantiated -
+  // whenever Tpetra is enabled, Xpetra::Epetra_Map depends on both
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, threads_device_type)
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long long, int, threads_device_type)
+
   // FIXME (mfh 26 Sep 2015) Once it becomes possible to disable LO =
   // int, add an instantiation here for {KeyType = LO, ValueType =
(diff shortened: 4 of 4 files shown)


brian-kelley:
@csiefer2 Thanks! I'll merge when the testing finishes.

packages/xpetra/src/Map/Xpetra_EpetraMap.cpp
```diff
@@ -49,6 +49,12 @@
 
 #include "Xpetra_EpetraMap.hpp"
 
+#ifdef HAVE_XPETRA_KOKKOS_REFACTOR
+#ifdef HAVE_XPETRA_TPETRA
+#include "Tpetra_Details_FixedHashTable_def.hpp"
```
mhoemmen: :( This is kinda ungood.  If you need another instantiation, add another instantiation.  I can show you how.

mhoemmen:
> Satisfy dependency of Xpetra EpetraMap on
> Tpetra::Details::FixedHashTable with int/long and
> int/longlong as LO/GO. These dependencies always exist
> as long as Tpetra is enabled, but regardless of whether
> these LO/GO combinations are enabled in Tpetra ETI.

We can add ETI just for FixedHashTable.  But why does Xpetra enable _both_ `GO=long` and `GO=long long` for Epetra?  That makes no sense at all.  I could see `long long` because of Epetra64, but why `long`?

packages/xpetra/src/Map/Xpetra_EpetraMap.cpp
mhoemmen: Also, this could cause link errors due to multiple definitions, depending on Tpetra's ETI settings.

jhux2: I agree with Mark.  I also don't like the fact that there's a dependence on Tpetra in the Epetra path.

brian-kelley:
@mhoemmen Can you take another look? I think I did what you have in mind.

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Cuda.cpp
```diff
@@ -79,6 +79,14 @@ namespace Details {
 
   TPETRA_INSTANTIATE_L( TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT_CUDA_INT )
 
+  // Make sure that KeyType = long, ValueType = int and
+  // KeyType = long long, ValueType = int both get instantiated -
+  // whenever Tpetra is enabled, Xpetra::Epetra_Map depends on both
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, cuda_device_type)
```
mhoemmen: 1. You actually want `int, long`, not the other way around.  The macro (see the bottom of `Tpetra_Details_FixedHashTable_def.hpp`) takes arguments `LO, GO, DEVICE`.  It reverses `LO` and `GO` as template arguments to `FixedHashTable`.

2. These macro invocations don't belong inside `#ifndef HAVE_TPETRA_INST_INT_INT ... #endif`.   You should actually put them inside `#ifndef HAVE_TPETRA_INST_INT_LONG ... #endif` resp. `#ifndef HAVE_TPETRA_INST_INT_LONG_LONG ... #endif`.  Otherwise, you'll get link errors due to duplicate instantiations.  (Are you somehow testing with ETI OFF?)
 
3. I know it's not your doing, but it seems really broken that Xpetra needs the `long` instantiation for Epetra.

4. Why are we instantiating both for `cuda_uvm_device_type` and for `cuda_device_type`?  Do we really need both?

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_OpenMP.cpp
```diff
@@ -71,6 +71,12 @@ namespace Details {
 
   TPETRA_INSTANTIATE_L( TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT_OPENMP_INT )
 
+  // Make sure that KeyType = long, ValueType = int and
+  // KeyType = long long, ValueType = int both get instantiated -
+  // whenever Tpetra is enabled, Xpetra::Epetra_Map depends on both
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, openmp_device_type)
```
mhoemmen: See above.

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Serial.cpp
```diff
@@ -71,6 +71,12 @@ namespace Details {
 
   TPETRA_INSTANTIATE_L( TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT_SERIAL_INT )
 
+  // Make sure that KeyType = long, ValueType = int and
+  // KeyType = long long, ValueType = int both get instantiated -
+  // whenever Tpetra is enabled, Xpetra::Epetra_Map depends on both
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, serial_device_type)
```
mhoemmen: See above.

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_Threads.cpp
```diff
@@ -71,6 +71,12 @@ namespace Details {
 
   TPETRA_INSTANTIATE_L( TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT_PTHREAD_INT )
 
+  // Make sure that KeyType = long, ValueType = int and
+  // KeyType = long long, ValueType = int both get instantiated -
+  // whenever Tpetra is enabled, Xpetra::Epetra_Map depends on both
+  TPETRA_DETAILS_FIXEDHASHTABLE_INSTANT(long, int, threads_device_type)
```
mhoemmen: See above.

lucbv:
@brian-kelley @jhux2 @mhoemmen (I'm broadcasting to get input from y'all).

This issue started because a `local_map_type` is used in `MueLu_TentativePFactory_kokkos` (see commit db7d4cbb by @aprokop ). The overall goal is to loop over global elements in a map and check if they are locally owned.
This raises a few questions (at least in my mind):

1) This code exists because of the kokkos refactor work in Xpetra/MueLu, why do we try to have Epetra implement kokkos refactor (see #4325)? Should we not check in Xpetra if `lib==UseEpetra` and then just throw instead of reimplementing Tpetra in Xpetra_Epetra?

2) is the code in `MueLu_TentativePFactory_kokkos` really necessary? Based on the fact that we are doing uncoupled aggregation in that particular spot we should **never** be dealing with off process nodes, so we should not have to check if nodes/dofs are local...

3) it seems to me that as long as we are not ready to use Cuda aware communication the map will always need to do synchronize host and device, why don't we just wrap the host data into a view and then deep_copy to device?

My recommendation is to remove `Xpetra::EpetraMapT::getLocalMap` from Xpetra, that will fix the ETI issue, then if some wants to use `getLocalMap` in kokkos refactor code, then should feel free to recast the `Xpetra::Map` into `Xpetra::TpetraMap` or even better to wrap the underlying map data into a Kokkos::View and then ship it to device if needed...

lucbv:
@brian-kelley @mhoemmen @jhux2 after calming down a bit... I think that we could fix the dependency issue by creating a `Xpetra::LocalMap` pure virutal class that is then implemented in `Xpetra::EpetraLocalMap` and `Xpetra::TpetraLocalMap`?
This way the return type of `Xpetra::Map::getLocalMap` would be `Xpetra::LocalMap` and the Tpetra headers will only be pulled in `Xpetra::TpetraLocalMap`.
This is annoying because it forces us to create a new class, but then again Tpetra did create a new class for the local map...

mhoemmen:
@lucbv I'm in meetings all week, but just wanted to say that Xpetra should do what Xpetra thinks best -- it doesn't have to be constrained by Tpetra's interface.  Furthermore, `Tpetra::Details::LocalMap` is a very simple class; if you don't want a dependency on Tpetra, you could just replicate it as a concrete class.  The Epetra and Tpetra versions of it could have exactly the same implementation, since they both need to do the same things.

mhoemmen:
I'm with @lucbv -- it makes sense to have a separate Xpetra class that does what you want.  If you design it carefully, I think you could even make it work on device, despite virtual methods.  (Have an outer wrapper class with nonvirtual methods, that uses an inner Xpetra class with inner virtual methods.  The outer class' copy constructor would invoke the inner class' constructor using placement `new`, so that the virtual method calls would work on device.  Users wouldn't need to know about the placement `new` stuff.)

The point of Tpetra's `LocalMap` is that the class is totally safe for use on the GPU.  It exposes only the methods that you're allowed to call on device.  While the corresponding Map methods are safe to call on device, Map's constructor is _not_.  Launching a device kernel with a Map would require invoking Map's copy constructor.  This would likely cause many "can't call host function from host device function" warnings.

jhux2:
All, see PR #4863.

jhux2:
Since #4863 has been merged, should this be close?

PR #4402, opened 2019-02-14
MueLu: Change MultiJagged defaults

cgcgcg:
@trilinos/muelu

## Description
Change the default `mj_premigration_coordinate_count` used in Zoltan2's MJ partitioning.
Instead of Zoltan2's default value, use `"repartition: target rows per proc"`.

### DIFF
--- a/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
+++ b/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
@@ -372,6 +372,7 @@ namespace MueLu {
         coarseLevel.Set("R",R11_);
         coarseLevel.Set("Coordinates",CoordsH_);
         coarseLevel.Set("number of partitions", numProcsAH);
+        coarseLevel.Set("repartition: heuristic target rows per process", 1000);
 
         coarseLevel.setlib(AH_->getDomainMap()->lib());
         fineLevel.setlib(AH_->getDomainMap()->lib());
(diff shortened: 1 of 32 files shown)


packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
```diff
@@ -372,6 +372,7 @@ namespace MueLu {
         coarseLevel.Set("R",R11_);
         coarseLevel.Set("Coordinates",CoordsH_);
         coarseLevel.Set("number of partitions", numProcsAH);
+        coarseLevel.Set("repartition: heuristic target rows per process", 1000);
```
jhux2: @cgcgcg I thought the idea was that this should be the same as `repartition: target rows per processor`?

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
```diff
@@ -536,6 +537,7 @@ namespace MueLu {
         if (doRebalancing) {
 
           coarseLevel.Set("number of partitions", numProcsA22);
+          coarseLevel.Set("repartition: heuristic target rows per process", 1000);
```
jhux2: Same question here.

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
cgcgcg: Yes, but that variable gets transformer depending on the number of threads. We should probably rename this one though.

jhux2: I see.  If someone sets `repartition: min rows per thread` to a positive value, then this would in turn modify `repartition: target rows per proc`.  I don't object to your change, but this does make the number of options rather daunting.

cgcgcg: We are not exposing this parameter to the user, so I think this is ok. If the user does not want the default, he/she needs to set `mj_premigration_coordinate_count`, same as before.

PR #4411, opened 2019-02-15
Issue 4283: refactor muelu to not use dynamic profile when declaring crsgraph

GeoffDanielson:
@trilinos/muelu 
@trilinos/tpetra 

## Description
This removes Tpetra::DynamicProfile from muelu

## Motivation and Context
Deprecation of DynamicProfile in Tpetra's effect on downstream packages

## How Has This Been Tested?
Since this is a change to tested code, the affected unit tests have been rerun to ensure that they are no longer using the deprecated code path.  Tests conducted on linux 2.6.32.

## Checklist

- [ ] These changes break backwards compatibility.once created. Most areas in this message are

### DIFF
--- a/packages/muelu/src/Graph/MueLu_CoalesceDropFactory_def.hpp
+++ b/packages/muelu/src/Graph/MueLu_CoalesceDropFactory_def.hpp
@@ -828,7 +828,7 @@ namespace MueLu {
       GetOStream(Statistics1) << "CoalesceDropFactory: nodeMap " << nodeMap->getNodeNumElements() << "/" << nodeMap->getGlobalNumElements() << " elements" << std::endl;
 
       // 3) create graph of amalgamated matrix
-      RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::DynamicProfile);
+      RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::StaticProfile);
 
       LO numRows = A->getRowMap()->getNodeNumElements();
       LO numNodes = nodeMap->getNodeNumElements();
--- a/packages/muelu/src/Graph/MueLu_CoalesceDropFactory_def.hpp
+++ b/packages/muelu/src/Graph/MueLu_CoalesceDropFactory_def.hpp
@@ -828,7 +828,7 @@ namespace MueLu {
       GetOStream(Statistics1) << "CoalesceDropFactory: nodeMap " << nodeMap->getNodeNumElements() << "/" << nodeMap->getGlobalNumElements() << " elements" << std::endl;
 
       // 3) create graph of amalgamated matrix
-      RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::StaticProfile);
+      RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, A->getNodeMaxNumRowEntries()*blockdim, Xpetra::StaticProfile);
 
       LO numRows = A->getRowMap()->getNodeNumElements();
       LO numNodes = nodeMap->getNodeNumElements();
--- a/packages/muelu/src/Graph/StructuredAggregation/MueLu_AggregationStructuredAlgorithm_def.hpp
+++ b/packages/muelu/src/Graph/StructuredAggregation/MueLu_AggregationStructuredAlgorithm_def.hpp
@@ -249,7 +249,7 @@ namespace MueLu {
     myGraph = CrsGraphFactory::Build(rowMap,
                                      colMap,
                                      nnzOnRow,
-                                     Xpetra::DynamicProfile);
+                                     Xpetra::StaticProfile);
 
     *out << "Fill CrsGraph" << std::endl;
     LO rowIdx = 0;
--- a/packages/muelu/src/Rebalancing/MueLu_IsorropiaInterface_def.hpp
+++ b/packages/muelu/src/Rebalancing/MueLu_IsorropiaInterface_def.hpp
@@ -134,7 +134,7 @@ namespace MueLu {
     
 
     // 3) create graph of amalgamated matrix
-    RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::DynamicProfile);
+    RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::StaticProfile);
 
     // 4) do amalgamation. generate graph of amalgamated matrix
     for(LO row=0; row<Teuchos::as<LO>(A->getRowMap()->getNodeNumElements()); row++) {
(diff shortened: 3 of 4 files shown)


packages/muelu/src/Graph/MueLu_CoalesceDropFactory_def.hpp
```diff
@@ -828,7 +828,7 @@ namespace MueLu {
       GetOStream(Statistics1) << "CoalesceDropFactory: nodeMap " << nodeMap->getNodeNumElements() << "/" << nodeMap->getGlobalNumElements() << " elements" << std::endl;
 
       // 3) create graph of amalgamated matrix
-      RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::DynamicProfile);
+      RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::StaticProfile);
```
jhux2: @GeoffDanielson Ugh.  I suspect `10` is arbitrary, and that this code relies on having the matrix expand its row storage as needed.  Did you run the MueLu unit tests to see if this branch is ever hit?

GeoffDanielson: I did run the muelu unit tests and the limit doesn't get hit.  I suspect you're right, but I'm not sure whether it'd be more correct to fix it with the proper upper limit or just leave as is.

mhoemmen:
@tjfulle Is that CrsGraph and CrsMatrix storage expansion thing on Export all done?  (Too many things to keep track of ;-P )

tjfulle:
@mhoemmen - for graph, yes.  Export/import in to matrix uses the underlying local matrix suminto and/or replace.  I haven’t touched that.  Sorted and merged inserts are next on my list and that may help for this issue if failure is due to repeats being inserted.

packages/muelu/src/Graph/MueLu_CoalesceDropFactory_def.hpp
jhux2: If the matrix is an Xpetra::BlockedCrsMatrix, then the method `getNodeMaxNumRowEntries` is available.  I am fine with this change.

packages/muelu/src/Graph/StructuredAggregation/MueLu_AggregationStructuredAlgorithm_def.hpp
```diff
@@ -249,7 +249,7 @@ namespace MueLu {
     myGraph = CrsGraphFactory::Build(rowMap,
                                      colMap,
                                      nnzOnRow,
-                                     Xpetra::DynamicProfile);
+                                     Xpetra::StaticProfile);
```
jhux2: @lucbv should really comment on this.

packages/muelu/src/Rebalancing/MueLu_IsorropiaInterface_def.hpp
```diff
@@ -134,7 +134,7 @@ namespace MueLu {
     
 
     // 3) create graph of amalgamated matrix
-    RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::DynamicProfile);
+    RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::StaticProfile);
```
jhux2: Ok, for the same reason as for `MueLu::CoalesceDropFactory_def.hpp`.

packages/muelu/src/Graph/StructuredAggregation/MueLu_AggregationStructuredAlgorithm_def.hpp
lucbv: My feeling is: this should be fine, the graph constructor actually allocates the exact amount of memory it needs since nnzOnRow is specified.
Of course I would still want this example to be tested to double check correctness.

mhoemmen: Thanks @lucbv :-)

packages/muelu/src/Graph/MueLu_CoalesceDropFactory_def.hpp
```diff
@@ -828,7 +828,7 @@ namespace MueLu {
       GetOStream(Statistics1) << "CoalesceDropFactory: nodeMap " << nodeMap->getNodeNumElements() << "/" << nodeMap->getGlobalNumElements() << " elements" << std::endl;
 
       // 3) create graph of amalgamated matrix
-      RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::DynamicProfile);
+      RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, A->getNodeMaxNumRowEntries()*blockdim, Xpetra::StaticProfile);
```
mhoemmen: This code might run fine, but it could blow out memory if one row happens to have many more entries than other rows.  Once we make this change, we may also want to duplicate the loop below, once to get a number of entries in each row (it looks like the loop is amenable to that), and once again to fill the CrsGraph.

packages/muelu/src/Rebalancing/MueLu_IsorropiaInterface_def.hpp
```diff
@@ -134,7 +134,7 @@ namespace MueLu {
     
 
     // 3) create graph of amalgamated matrix
-    RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, 10, Xpetra::DynamicProfile);
+    RCP<CrsGraph> crsGraph = CrsGraphFactory::Build(nodeMap, A->getNodeMaxNumRowEntries()*blockdim, Xpetra::StaticProfile);
```
mhoemmen: See note above.  Can MueLu actually call Isorropia with a Tpetra matrix?

packages/muelu/src/Graph/MueLu_CoalesceDropFactory_def.hpp
jhux2: @mhoemmen That's a good suggestion, I'll push an update.

packages/muelu/src/Rebalancing/MueLu_IsorropiaInterface_def.hpp
jhux2: No, MueLu will throw.  See 

https://github.com/trilinos/Trilinos/blob/5c9a2ec8019b310b5b74a22e23fcbea11486f781/packages/muelu/src/Rebalancing/MueLu_IsorropiaInterface_def.hpp#L216-L223

mhoemmen: @jhux2 Thanks for the clarification!  It's probably better for even the Epetra-based code to use StaticProfile, as long as it has good bounds for each row.

mhoemmen:
@jhux2 were you planning on pushing an update first?

jhux2:
@mhoemmen @GeoffDanielson I'm ok with merging this as is.   Adding more precise graph sizes will involve changes to Xpetra, which can be a separate PR.

mhoemmen:
@jhux2 Thanks!

PR #4413, opened 2019-02-15
Tpetra::Distributor: Fix #4412 (don't fill indicesFrom_)

### LINKED ISSUE #4412: Tpetra::Distributor: Don't fill indicesFrom_
@trilinos/tpetra @jjellio @kddevin @vbrunini @crtrott 

`Tpetra::Distributor` would always fill `indicesFrom_` with 0, 1, ..., `totalReceiveLength_-1`.  This was not only a waste of time, it would unnecessarily make all reverse-mode communication take the slow path.  (Reversing a Distributor swaps `indicesTo_` and `indicesFrom_`.  A nonempty `indicesTo_` tells Distributor to take the slow path.  If `indicesFrom_` is just 0, 1, ..., `totalReceiveLength_-1`, then Distributor does not need to take the slow path.)

Note that `Epetra_MpiDistributor` comments out the code that fills `indices_from_`, and claims that doing so fixes reverse communication.

@vbrunini found that applying my patch for fixing this made Aria's post-solve communication of the solution vector of the ViewFactor system a lot faster on 2 GPUs.

## Related Issues
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of

mhoemmen:
@trilinos/tpetra 

`Tpetra::Distributor` would always fill `indicesFrom_` with 0, 1, ..., `totalReceiveLength_-1`. This was not only a waste of time, it would unnecessarily make all reverse-mode communication take the slow path. (Reversing a Distributor swaps `indicesTo_` and `indicesFrom_`. A nonempty `indicesTo_` tells Distributor to take the slow path. If `indicesFrom_` is just 0, 1, ..., `totalReceiveLength_-1`, then Distributor does not need to take the slow path.)

Note that `Epetra_MpiDistributor` comments out the code that fills `indices_from_`, and claims that doing so fixes reverse communication.

@vbrunini found that applying my patch for fixing this made Aria's post-solve communication of the solution vector of the ViewFactor system a lot faster on 2 GPUs.

## Related Issues

* Closes #4412 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of

### DIFF
--- a/packages/teuchos/core/src/Teuchos_RCPNode.hpp
+++ b/packages/teuchos/core/src/Teuchos_RCPNode.hpp
@@ -132,7 +132,6 @@ public:
       // Should never get here!
 #ifdef TEUCHOS_DEBUG
       TEUCHOS_TEST_FOR_EXCEPT(true);
-      return "";
 #else
       return "";
 #endif
--- a/packages/tpetra/core/src/Tpetra_Distributor.cpp
+++ b/packages/tpetra/core/src/Tpetra_Distributor.cpp
@@ -45,10 +45,10 @@
 #include <numeric>
 
 namespace { // (anonymous)
-  
+
   Teuchos::RCP<Teuchos::FancyOStream>
   makeValidVerboseStream (const Teuchos::RCP<Teuchos::FancyOStream>& out =
-			  Teuchos::null)
+                          Teuchos::null)
   {
     if (out.is_null ()) {
       return Teuchos::getFancyOStream (Teuchos::rcpFromRef (std::cerr));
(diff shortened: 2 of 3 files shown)


mhoemmen:
@vbrunini reports that applying this patch may have changed Chaparral test problem results a bit, with slightly different solver iteration counts and slightly different solutions.  I've marked this patch as "results impacting" just in case that's a Tpetra issue.  The only way that solve depends on reverse-mode communication is in scattering out the solution vector back to the application's distribution of degrees of freedom, using reverse-mode Import (with an Export object).  @vbrunini , is that with the `REPLACE` `CombineMode`?

mhoemmen:
@jjellio This is likely relevant to your current interests.

vbrunini:
Yes, it is REPLACE combine mode.

mhoemmen:
@vbrunini wrote:
> Yes, it is REPLACE combine mode.

Did your results include the patch to MultiVector to stop doing `Kokkos::atomic_assign`?

vbrunini:
False alarm. I first noticed the differences while testing your patches, but I've now gone back and reverted those and still see issues so the cause must be something prior to this change (and not in Trilinos).

mhoemmen:
Thanks @vbrunini !  I'm glad because it would be Very Weird otherwise :-D

PR #4425, opened 2019-02-18
Parameterized pr cdash track

william76:
@trilinos/framework 

Updating PR test driver script to use the `PULLREQUEST_CDASH_TRACK` parameter, if it exists.

I'd like to let this run through tests before we merge the gitlab update to the autotester to add the parameter in, and then rerun the tests once that sees them.

@jwillenbring fyi.

### DIFF
--- a/cmake/std/PullRequestLinuxDriver-Test.sh
+++ b/cmake/std/PullRequestLinuxDriver-Test.sh
@@ -302,7 +302,7 @@ CDASH_TRACK="Pull Request"
 echo -e ">>> PULLREQUEST_CDASH_TRACK = ${PULLREQUEST_CDASH_TRACK}"
 
 # If PULLREQUEST_CDASH_TRACK is set and nonzero
-if [ ! -z ${FOO} ]; then
+if [ ! -z ${PULLREQUEST_CDASH_TRACK} ]; then
     CDASH_TRACK=${PULLREQUEST_CDASH_TRACK:?}
     echo -e ">>> PULLREQUEST_CDASH_TRACK is set. Setting CDASH_TRACK=${PULLREQUEST_CDASH_TRACK:?}"
 else
--- a/cmake/std/PullRequestLinuxDriver-Test.sh
+++ b/cmake/std/PullRequestLinuxDriver-Test.sh
@@ -295,7 +295,20 @@ else
     echo -e "MPI type = sems-${SEMS_MPI_NAME:?}/${SEMS_MPI_VERSION:?}"
 fi
 
+# Set the Pull Request CDash Track. 
+# Defaults to "Pull Request" if the parameter PULLREQUEST_CDASH_TRACK does not exist or is empty
 CDASH_TRACK="Pull Request"
+
+echo -e ">>> PULLREQUEST_CDASH_TRACK = ${PULLREQUEST_CDASH_TRACK}"
+
+# If PULLREQUEST_CDASH_TRACK is set and nonzero
+if [ ! -z ${FOO} ]; then
+    CDASH_TRACK=${PULLREQUEST_CDASH_TRACK:?}
+    echo -e ">>> PULLREQUEST_CDASH_TRACK is set. Setting CDASH_TRACK=${PULLREQUEST_CDASH_TRACK:?}"
+else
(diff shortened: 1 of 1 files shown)


william76:
@prwolfe I don't think so...

Prior to submitting the PR, I tested it with this:

```bash
#!/usr/bin/env bash

function test_envvar_FOO_exists() {
    if [ ! -z ${FOO} ]; then
        echo -e "- FOO is set with value '${FOO}'"
    else
        echo -e "- FOO is either unset or empty"
    fi
}

echo -e "Unset environment var: FOO"
unset FOO
test_envvar_FOO_exists

echo -e ""
echo -e "Set environment var: FOO=\"\""
export FOO=""
test_envvar_FOO_exists

echo -e ""
echo -e "Set environment var: FOO=\"theValue\""
export FOO="theValue"
test_envvar_FOO_exists

unset FOO
```

Which gives me this output:

```
0 $ ./test-envvar-exists.sh
Unset environment var: FOO
- FOO is either unset or empty

Set environment var: FOO=""
- FOO is either unset or empty

Set environment var: FOO="theValue"
- FOO is set with value 'theValue'
```

So I followed that style and did this:

https://github.com/trilinos/Trilinos/blob/466443b711c0acd7f997c4f8c7ee61b2ad4c8fab/cmake/std/PullRequestLinuxDriver-Test.sh#L302-L310

Which gave me this output in my jenkins simulator on a RHEL7 machine:

```
 14) atdm-env
 15) atdm-ninja_fortran/1.7.2
MPI type = sems-openmpi/1.10.1
>>> PULLREQUEST_CDASH_TRACK = Clean
>>> PULLREQUEST_CDASH_TRACK is set. Setting CDASH_TRACK=Clean
CDash Track = Clean
pwd: /ascldap/users/wcmclen/dev/jsim/jsim/packages/trilinos-pr-gcc-7.2.0/workspace
+ env TRILINOS_DIR=/home/wcmclen/dev/jsim/jsim/packages/trilinos-pr-gcc-7.2.0/workspace/Trilinos TRILINOS_SCRIPTS_DIR=/home/wcmclen/dev/jsim/jsim/packages/trilinos-pr-gcc-7.2.0/workspace/Trilinos /home/wcmclen/dev/jsim/jsim/packages/trilinos-pr-gcc-7.2.0/workspace/Trilinos/commonTools/framework/get-changed-trilinos-packages.sh origin/develop HEAD packageEnables.cmake
```

I tested out a couple different variations of the if -z statement:

| Variation   |  Works? |
| -------------------------- | -------------------------|
| `if [ ! -z ${FOO} ]; then`       | YES  |
| `if [ ! -z "${FOO}" ]; then`    | YES  |
| `if [[ ! -z ${FOO} ]]; then`    | YES  | 
| `if [[ ! -z "${FOO}" ]]; then` | YES  |

Do you know what the difference is on `if` statements in bash between the `[[ ... ]]` and the `[ ... ]` notations?  It seems to work fine with or without quotes around the `${VAR}`. 

I can add them in if you think it'd be an issue, but the testing passed last night using it... but I wanted to run a clean test first and then do a retest after merging in the [PR over on our gitlab repository][1].

[1]: https://gitlab.sandia.gov/trilinos-project/autotester_configs/merge_requests/3

prwolfe:
Probably me remembering older versions of bash - I expected a complaint about -z has no operand (basically the empty string does not show up without the quotes.) Probably fixed by now.

the "test", "[]", and "[[]]" constructs all do the same thing.  The only difference is "test" and "[]" sort the results differently than "[[]]" and I would have to look up which one does what.

william76:
I know I've bumped into some things where the quotes matter... but it might be older bash versions /shrug... I've gotten into the habit of writing a small test script whenever I'm doing something that's messing with `if` statements in bash because there always seems to be some odd thing that I miss on my first go, such as needing the double bracket when I give it a single bracket or needing quotations on some things, etc... Bash tends to be universal but man it can be finicky sometimes :p 

Oh, I just remembered, my edits on the gitlab repo only affect dev->master PR's.  I'll go ahead and merge this one in and once someone hits the merge button on my merge-request over there to add the parameter into the dev->master config for the autotester I'll kick off a dev->master run and see if it does the right thing.  (I don't have merge powers on that repo... /shrug, it's probably for the best :p )

PR #4428, opened 2019-02-19
Ifpack: Ifpack_Hypre.cpp improved test for vector X and Y sameness

jblahos:
@trilinos/Ifpack

## Description
An additional check was added to test the sameness of vectors X and Y used in Ifpack_Hypre.cpp. The previous version only checked for equality of the "outer pointers" but not the "inner" ones.

## Motivation and Context

The X and Y vectors were sometimes wrongly considered not being the same because the outer pointers were different but the inner ones were the same. So both vectors could in the end pointed to the same data array and the code wouldn't recognise that.

## How Has This Been Tested?

I tested this in my code that uses Ifpack_Hypre as a preconditioner for newton solver in a loca/nox continuation project. The preconditioner was failing before, the linear system returned error message that it wasn't solved. After this change, the linear solve works and I am obtaining the same results as when using other preconditioners.

## Checklist

I assume that no compiler warnings were introduced and all tests would pass since the change is minor but I didn't verify that.


packages/ifpack/src/Ifpack_Hypre.cpp
```diff
@@ -299,7 +299,7 @@ int Ifpack_Hypre::ApplyInverse(const Epetra_MultiVector& X, Epetra_MultiVector&
   bool SameVectors = false;
   int NumVectors = X.NumVectors();
   if (NumVectors != Y.NumVectors()) IFPACK_CHK_ERR(-1);  // X and Y must have same number of vectors
-  if(X.Pointers() == Y.Pointers()){
+  if(X.Pointers() == Y.Pointers() || (NumVectors == 1 && X[0] == Y[0])){
```
mhoemmen: Is it possible to construct a noncontiguous column view of another MultiVector?  If so, what you actually want is to test whether each column pointer is the same.  Otherwise, I think this is OK.

jblahos: I thought that X[0] == Y[0] does that, checks for equality of column pointers. But I am not that familiar with Trilinos, the solution I made was actually proposed by Mr Heroux after I emailed him about the issue.

Issue #4430, opened 2019-02-19
Try 2: Remove deprecated inheritance of SerialDense matrices from Object


rhoope:
This is a new issue for the idea captured in Issue #4258 but using an approach that doesn't break backward compatibility (cf Issue #4330) and attempts to align with current guidance for deprecating code as in Issue #4269).

@trilinos/teuchos

## Motivation and Context

Details are provided in Issue #4258.
## Your Environment

- **Relevant repo SHA1s:**  
- **Relevant configure flags or configure script:**  
- **Operating system and version:**  
- **Compiler and TPL versions:**  

## Related Issues

* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to Issues #4258 and #4330
* Part of 
* Composed of Issue #4258

mhoemmen:
@trilinos/teuchos

rhoope:
Thanks much @mhoemmen !

mhoemmen:
@rhoope Thanks for all your hard work! :D

PR #4432, opened 2019-02-19
Issue #4430 : Try 2 to Remove deprecated inheritance of SerialDense m…

rhoope:
…atrices from Object

 - remove inheritance from Teuchos::Object of Teuchos::SerialDenseMatrix
   (which in turn includes Teuchos::SerialDenseVector),
   Teuchos::SerialBandDenseMatrix and Teuchos::SerialSymDenseMatrix
 - add operator<< for each class protected by TEUCHOS_HIDE_DEPRECATED_CODE
   to preserve backward compatibility and allow Dakota to explicitly disable

I tested that these changes allow existing client code in Trilinos to
compile, and all Teuchos tests/examples pass under GCC 4.8.4

@trilinos/teuchos 

## Checklist

### DIFF
--- a/packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
+++ b/packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
@@ -130,7 +130,7 @@ the object.  Specifically:
 namespace Teuchos {
 
 template<typename OrdinalType, typename ScalarType>
-class SerialBandDenseMatrix : public CompObject, public Object, public BLAS<OrdinalType, ScalarType>
+class SerialBandDenseMatrix : public CompObject, public BLAS<OrdinalType, ScalarType>
 {
 public:
 
@@ -421,7 +421,7 @@ public:
 
   //! @name I/O methods.
   //@{
-  //! Print method.  Defines the behavior of the std::ostream << operator inherited from the Object class.
+  //! Print method.  Defines the behavior of the std::ostream << operator
   virtual void print(std::ostream& os) const;
 
   //@}
@@ -448,7 +448,6 @@ protected:
 template<typename OrdinalType, typename ScalarType>
 SerialBandDenseMatrix<OrdinalType, ScalarType>::SerialBandDenseMatrix ()
   : CompObject (),
-    Object(),
     BLAS<OrdinalType,ScalarType>(),
     numRows_ (0),
     numCols_ (0),
@@ -467,7 +466,6 @@ SerialBandDenseMatrix (OrdinalType numRows_in,
                        OrdinalType ku_in,
                        bool zeroOut)
   : CompObject (),
-    Object(),
     BLAS<OrdinalType,ScalarType>(),
     numRows_ (numRows_in),
     numCols_ (numCols_in),
@@ -493,7 +491,6 @@ SerialBandDenseMatrix (DataAccess CV,
                        OrdinalType kl_in,
                        OrdinalType ku_in)
   : CompObject (),
-    Object(),
     BLAS<OrdinalType,ScalarType>(),
     numRows_ (numRows_in),
     numCols_ (numCols_in),
@@ -515,7 +512,6 @@ template<typename OrdinalType, typename ScalarType>
 SerialBandDenseMatrix<OrdinalType, ScalarType>::
 SerialBandDenseMatrix (const SerialBandDenseMatrix<OrdinalType, ScalarType> &Source, ETransp trans)
   : CompObject (),
-    Object(),
     BLAS<OrdinalType,ScalarType>(),
     numRows_ (0),
     numCols_ (0),
@@ -574,7 +570,6 @@ SerialBandDenseMatrix (DataAccess CV,
                        OrdinalType numCols_in,
                        OrdinalType startCol)
   : CompObject (),
-    Object(),
     BLAS<OrdinalType,ScalarType>(),
     numRows_ (numRows_in),
     numCols_ (numCols_in),
@@ -1074,6 +1069,16 @@ void SerialBandDenseMatrix<OrdinalType, ScalarType>::copyMat(
     }
   }
 }
+
+#ifndef TEUCHOS_HIDE_DEPRECATED_CODE
+/// \brief Print the given SerialBandDenseMatrix to the given output stream.
+template<typename OrdinalType, typename ScalarType>
+std::ostream& operator<< (std::ostream& os, const Teuchos::SerialBandDenseMatrix<OrdinalType, ScalarType>& obj)
+{
+  obj.print (os);
+  return os;
+}
+#endif
 
 } // namespace Teuchos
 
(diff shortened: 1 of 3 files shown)


kddevin:
@kddevin @hkthorn

packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
```diff
@@ -1075,6 +1070,16 @@ void SerialBandDenseMatrix<OrdinalType, ScalarType>::copyMat(
   }
 }
 
+#ifndef TEUCHOS_HIDE_DEPRECATED_CODE
```
mhoemmen: For consistency with Kokkos, should we instead call the macro `TEUCHOS_ENABLE_DEPRECATED_CODE`, do an `#ifdef` test, and have the macro be defined by default?

bartlettroscoe: @mhoemmen, actually, this is consistent with the original standard documented at:

* https://tribits.org/doc/TribitsDevelopersGuide.html#hiding-deprecated-code-to-certify-and-facilitate-later-removal

By requiring a user to define `TEUCHOS_ENABLE_DEPRECATED_CODE` you are essentially breaking backward compatibility.  (Of course the configured header file should set this define correctly but requiring a macro to be defined to hide deprecated seems to better avoid breaks in backward compatibility.)

packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
```diff
@@ -1075,6 +1070,16 @@ void SerialBandDenseMatrix<OrdinalType, ScalarType>::copyMat(
...
+/// \brief Print the given SerialBandDenseMatrix to the given output stream.
+template<typename OrdinalType, typename ScalarType>
+std::ostream& operator<< (std::ostream& os, const Teuchos::SerialBandDenseMatrix<OrdinalType, ScalarType>& obj)
```
bartlettroscoe: @rhoope, is function is not actually deprecated.  Technically speaking, to deprecate, you need to apply the macro `TEUCHOS_DEPRECATED` as described at:

* https://tribits.org/doc/TribitsDevelopersGuide.html#deprecating-c-c-classes-structs-functions-typdefs-etc

Otherwise, users will not know this function is deprecated and therefore will not know to stop using it.

rhoope: @bartlettroscoe I can add the macro to alert clients to the deprecation of the code.  What I'm unclear on is the idea of adding new headers; won't this break backward compatibility by requiring clients to include the new files?

bartlettroscoe: @rhoope, the problem is that if you add the `TEUCHOS_DEPRECATED` macro then you need to provide a way for existing users to remove the usage of the deprecated code.  How can they do this?  We don't expect them to have to change to raw `obj.print()` do we?  That is not good.

For now we can leave it as it as is without adding the `TEUCHOS_DEPRECATED` macro and therefore the deprecated warning will not appear.  But we are creating a time bomb that will eventually go off when Trilinos does its next major release and follows the process at:

* https://tribits.org/doc/TribitsDevelopersGuide.html#physically-removing-deprecated-code

to actually remove the deprecated code.  And for any customers that are configuring Trilinos to remove deprecated code, we need to provide them a smooth path to upgrade.  That path, I think, should be to provide `<class_header_name>OStream.hpp` and let them include that.  But we can add that in a future PR when needed.

I will go ahead and approve the PR.

rhoope:
Thanks @bartlettroscoe .  I'll be happy to assist with the fallout as time gets closer to the next Trilinos release.

packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
mseldre: With respect to users removing dependence on this deprecated code, I see two options:

- clients supply their own operator<<(ostream&) within their native namespaces, if desired, for the container types they employ.  They could choose to have this trivially invoke obj.print() in a minimalist solution, or implement a prettier print than what's provided (especially if the output is intended for something other than debug dumps/exception handling).  This is what we have done for Dakota.

- The other alternative would be for Teuchos to provide an operator<<(ostream&) in the global namespace, in order to provide a default implementation than can be overridden. Neither you nor Mark seemed keen on this idea, which is fine with me as I personally prefer the previous option.  The only reason to consider this is if we want to reduce the deprecation burden.

packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
mhoemmen: @bartlettroscoe Hm, Kokkos has already chosen the `ENABLE` path; maybe we should work something out with Christian so that it's consistent across packages.

bartlettroscoe: @mhoemmen, I don't think it really matters what macro names are used as long as the package configures its `<Package>_config.h` file to define its macros correctly according to the cmake configure options defined at:

* https://tribits.org/doc/TribitsBuildReference.html#disabling-deprecated-code

We could look in the git repo histories but I am pretty sure the stuff documented at:

* https://tribits.org/doc/TribitsDevelopersGuide.html#regulated-backward-compatibility-and-deprecated-code

existed before Kokkos added deprecated support.

packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
bartlettroscoe: @mseldre and @rhoope, I think the best approach is to new `<class_header_name>OStream.hpp` files and define ostream operator ifdefed as:

```
#ifdef TEUCHOS_HIDE_DEPRECATED_CODE
namespace Teuchos {
// \brief Print the given SerialBandDenseMatrix to the given output stream.
template<typename OrdinalType, typename ScalarType>
std::ostream& operator<< (std::ostream& os, const Teuchos::SerialBandDenseMatrix<OrdinalType, ScalarType>& obj)
{
  obj.print (os);
  return os;
}
} // namespace Teuchos
#endif // TEUCHOS_HIDE_DEPRECATED_CODE
```

Then customers that are building with deprecated code disable just need to include `<class_header_name>OStream.hpp` and they will compile with or without `TEUCHOS_HIDE_DEPRECATED_CODE` defined.  Then we the deprecated code is removed, they will continue to compile just fine.

mhoemmen:
Sorry, I'm in meetings all week; it might take me a couple days to get to this.

mhoemmen:
@ikalash FYI, please let us know if Albany has any build issues as a result of this change.  Hopefully not :-)

ikalash:
Will do.  I hope not too - finally today our dashboard was back to normal!

Issue #3647, opened 2018-10-16
Amesos2: Linear system segfaults w/ SuperLU, SuperLU_DIST, & UMFPACK, but not with KLU2


mhoemmen:
@trilinos/amesos2 

@vbrunini found a Tpetra linear system on a single MPI process, that makes Amesos2 segfault when using `SuperLU`, `SuperLU_DIST`, or `UMFPACK`.  However, `KLU2` does not segfault.

I have put the linear system through R&A and will post it as a pull request.  I just want to make this issue now so I can put the issue number in the PR.

@vbrunini 's guess is that Amesos2 does not like matrices that have "gaps" in their row Map.  That is the case for the matrix in the test that I will post.

srajama1:
KLU2 takes a special path for single rank cases compared to other codes. The bug might be in the code that we don't use for KLU2. @ndellingwood wanted to propagate these KLU2 changes to other solvers anyway. This might be the time to do it.

vbrunini:
Since you mentioned a special code path for single rank cases I just tried running the problem this matrix is from on 2 ranks and in that case KLU2 also segfaults.

srajama1:
Thanks for confirming it. That helps. We will resolve it. Is this blocking you from making progress ? I am asking to understand the timeline. We are busy with a Kokkos integration into Trilinos, we can look at this after that is done. Is that reasonable ?

vbrunini:
Yeah no problem.

srajama1:
@ndellingwood is looking at this

vbrunini:
Any update on this?

ndellingwood:
@vbrunini working on it this week, sorry for the delay I've been a bit overloaded the past few weeks but aiming to put in a PR with a fix this week.
Brief update - some previous work to handle non-contiguous GIDs was intended to resolve cases like this, but this case exposed some shortcomings I'm working to resolve. The single-thread run on KLU2 happened to run to completion but the results are not correct (confirmed by testing the matrix Mark provided with a different RHS and checking the residual). When setting a parameter to expect non-contiguous GIDs the code seg-faults, which exposed the existing issue. I've created a much smaller (5x5) single node reproducer that I'm using to work through the issue, will modify it to use the provided matrix and rhs as a test in Amesos2 once resolved.

vbrunini:
Great, thanks. Once you've got something working I can test it out against more of our test suite to make sure there isn't anything else unusual we are doing with Maps that isn't captured by this one test matrix.

[...]

PR #3649, opened 2018-10-17
WIP: Amesos2: Add "example" to test #3647

mhoemmen:
@trilinos/amesos2

## Description

Add "example" for testing #3647.  The example does not include the linear system files.  For the files and instructions, see #3647.  This PR should be harmless to merge, but I've nevertheless marked it as a WIP, with the intent that Amesos2 developers use it to develop a true nightly test.

## Related Issues

* Part of #3647 

## How Has This Been Tested?

Linux, GCC, OpenMPI.

PR #4454, opened 2019-02-21
Amesos2: Fix single-proc case for non-contiguous GIDs

ndellingwood:
Potential fix for issue #3647


ndellingwood:
@vbrunini this PR fixed the single proc case with KLU2 I used while working on the issue, labeled WIP until I cleanup the test. If you'd like to try it before merge please pull this PR; your code will require setting a parameter in the parameter list to indicate to Amesos2 that the row ids are "gapped" or non-contiguous, here is sample syntax:

```
std::string solverName = "KLU2";
Teuchos::ParameterList amesos2_params("Amesos2");
amesos2_params.sublist(solverName).set("IsContiguous", false, "Are GIDs Contiguous");
```

ndellingwood:
@vbrunini here's some updates on this PR:

In the recent commit I pushed an example `Amesos2_GappedMtxGIDs-1proc` (and small example matrices) to the `Trilinos/packages/amesos2/example` directory. The example uses some parts from Mark's PR #3649 to allow for reading in matrix market files with gapped row ids. By default, for now, the example uses a small matrix (1-proc) test case though you can pass the 'badMatrix.mm', 'badRhs.mm', and 'rowMap.mm' files as arguments and it's hard-coded to do the right things to test with those matrices. Later I'll push the 'badMatrix' etc files when I'm confident I have them marked correctly for release. 

A couple comments about the example:

To run with non-default matrices, you can execute something like this:
`./Amesos2_GappedMtxGIDs-1proc.exe  --matrixFilename="<DATADIRPATH>/badMatrix.mm" --rhsFilename="<DATADIRPATH>/badRhs.mm" --mapFilename="<DATADIRPATH>/rowMap.mm"`

The example is hard-coded to detect "badMatrix" in the `matrixFilename` string and set some internal variables to handle this case (reading matrix market files and maps with gapped row ids may require extra info depending on how the map was written e.g. index base 1 vs 0, knowing how many lines are in the header file). 

By default a parameter list is created for the solver that sets "IsContiguous" to "false", indicating to Amesos2 that the matrix contains gaps in the row ids and to accommodate for this. This can be disabled with the command line argument `--isContiguous`. To add this functionality to your codes, add lines like the following:

```
    Teuchos::ParameterList amesos2_params("Amesos2");
    amesos2_params.sublist(solverName).set("IsContiguous", false, "Are GIDs Contiguous");
    solver->setParameters( Teuchos::rcpFromRef(amesos2_params) );
```

It looks like it's important the `solverName` matches the solver closely which we need to ease (there is more leniency regarding caps when creating the solver). For the solvers listed in the issue, `solverName` should be:
`std::string solverName="KLU2";`
`std::string solverName="SuperLU";`
`std::string solverName="SuperLU_DIST";`
`std::string solverName="Umfpack";`

I tested this example with KLU2 and SuperLU 4.3 using the badMatrix etc. files and I get correct results (within tolerance via residual check); I had issues building SuperLU_DIST TPL locally and need to test on a machine with SEMS modules next.

Prior to this example, I wrote some separate driver code to read the badMatrix, badRhs and rowMap files into corresponding Tpetra data structures, convert them to have locally contiguous GIDs and then write to matrix market files. I was then able to read those files into Octave and compute a solution as a sanity check that the matrix itself was solvable, and the results also agreed well with the solvers used in Amesos2. If it's useful I can clean up the driver codes or files and send them to you. 

Hopefully the updates from this PR allow your code to work with KLU2 or SuperLU (I'll post updates once I finish testing with SLUDIST); if the changes here don't resolve the issues you are seeing it may help if we can follow up offline to figure out what extra info beyond the data files I'll need to reproduce and resolve.

ndellingwood:
@mhoemmen I used some of your matrix reader code for gapped-row-id cases from your example PR #3649 in the example I added to this PR, thanks for providing that (very helpful!) and let me know how I should properly credit you.

ndellingwood:
@srajama1 if you get a break during the conference can you review?

ndellingwood:
@srajama1 it fixes the problem for the matrix, rhs, and map provided.

ndellingwood:
@srajama1 we'll continue to followup if the provided files aren't a sufficient proxy for the problems reported in #3647, so it should stay open for now.

PR #4465, opened 2019-02-21
MueLu/Ifpack2: Improving interface to BlockRelaxation

csiefer2:
Auto-PR for SHA 7c6ec9e

### DIFF
--- a/packages/muelu/src/Smoothers/MueLu_Ifpack2Smoother_def.hpp
+++ b/packages/muelu/src/Smoothers/MueLu_Ifpack2Smoother_def.hpp
@@ -136,7 +136,18 @@ namespace MueLu {
     }
     else if (type_ == "BLOCK RELAXATION" ||
              type_ == "BLOCK_RELAXATION" ||
-             type_ == "BLOCKRELAXATION")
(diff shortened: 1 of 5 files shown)


packages/muelu/src/Smoothers/MueLu_Ifpack2Smoother_def.hpp
```diff
@@ -136,7 +136,18 @@ namespace MueLu {
     }
     else if (type_ == "BLOCK RELAXATION" ||
              type_ == "BLOCK_RELAXATION" ||
-             type_ == "BLOCKRELAXATION")
+             type_ == "BLOCKRELAXATION" ||
```
mhoemmen: Y'all know that Ifpack2::Factory does this, right?

csiefer2: It does, but not quite for the reason we're doing it here, which is to figure out if we need to give it coordinates.

mhoemmen: ah ok.

PR #4467, opened 2019-02-21
Tpetra MV: remove unneeded MPI buffer

cgcgcg:
@trilinos/tpetra 

## Description
In `MultiVector::reduce`, instead of using two distinct buffers for a `MPI_Allreduce`, use `MPI_IN_PLACE` and save one allocation per call.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -4364,11 +4364,25 @@ namespace Tpetra {
     // pointers.
     dev_view_type tgtBuf_dev;
     host_view_type tgtBuf_host;
+    const bool commIsInterComm = ::Tpetra::Details::isInterComm (comm);
+    // If comm is an intercomm, then we may not alias input and
+    // output buffers, so we have to make a copy of the local
+    // sum.
     if (useHostVersion) {
-      tgtBuf_host = host_view_type ("tgtBuf", lclNumRows, numCols);
+      if (commIsInterComm || !contig) {
+        tgtBuf_host = host_view_type (Kokkos::ViewAllocateWithoutInitializing ("tgtBuf"), lclNumRows, numCols);
+      }
+      else {
+        tgtBuf_host = srcBuf_host;
+      }
     }
     else {
-      tgtBuf_dev = dev_view_type ("tgtBuf", lclNumRows, numCols);
+      if (commIsInterComm || !contig) {
+        tgtBuf_dev = dev_view_type (Kokkos::ViewAllocateWithoutInitializing ("tgtBuf"), lclNumRows, numCols);
+      }
+      else {
+        tgtBuf_dev = srcBuf_dev;
+      }
     }
 
     const int reduceCount = static_cast<int> (totalAllocSize);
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -4357,66 +4357,46 @@ namespace Tpetra {
...
     if (useHostVersion) {
-      TEUCHOS_TEST_FOR_EXCEPTION
-        (static_cast<size_t> (tgtBuf_host.size ()) < totalAllocSize,
-         std::logic_error, "Tpetra::MultiVector::reduce: tgtBuf_host.size() = "
-         << tgtBuf_host.size () << " < lclNumRows*numCols = " << totalAllocSize
-         << ".  Please report this bug to the Tpetra developers.");
       reduceAll<int, impl_scalar_type> (comm, REDUCE_SUM, reduceCount,
                                         srcBuf_host.data (),
-                                        tgtBuf_host.data ());
+                                        srcBuf_host.data ());
```
mhoemmen: This is only correct if the MPI communicator is _not_ an intercommunicator.  Tpetra has a predicate for that: https://github.com/trilinos/Trilinos/blob/69f8850c25bdead51b38866ae34d4fa40947b67f/packages/tpetra/core/src/Tpetra_Details_isInterComm.hpp#L82

```c++
if (Details::isInterComm (comm)) {
  // intercomms are rare; don't prematurely optimize
  // 
  // allocate temp tgtBuf_host 
  // reduceAll from srcBuf_host to tgtBuf_host
  // copy back to srcBuf_host
}
else {
  reduceAll<int, impl_scalar_type> (comm, REDUCE_SUM, reduceCount, srcBuf_host.data (), srcBuf_host.data ());
}
```

cgcgcg: That's probably safer, but shouldn't we always be in that case? This is used in `MV::multiply`, and uses the communicator of the result C.

mhoemmen: @cgcgcg It's always legal to create a Map with an Comm that wraps an intercommunicator.  Intercommunicators behave like intracommunicators (the usual ones, including `MPI_COMM_WORLD`), except for this "`MPI_IN_PLACE` doesn't work" property.

cgcgcg: @mhoemmen Sounds good. I'll change it.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -4397,7 +4407,8 @@ namespace Tpetra {
     if (useHostVersion) {
       this->modify_host ();
       if (contig || isConstantStride ()) {
```
mhoemmen: Not quite sure why we're checking both `contig` and `isConstantStride()` here, but that's a separate issue, unrelated to this PR.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -4397,7 +4407,8 @@ namespace Tpetra {
...
-        Kokkos::deep_copy (srcView_host, tgtBuf_host);
+        if (commIsInterComm)
```
mhoemmen: For future work, please follow Tpetra coding style; e.g., always use `{ ... }` to set off `if` statements.

cgcgcg: Fixed.

mhoemmen:
@trilinos/framework Uh oh: "g++: internal compiler error: Killed (program cc1plus)"

cgcgcg:
@mhoemmen Thanks for re-approving!

PR #4491, opened 2019-02-25
Tempus: Infrastructure for the initial conditions.

### LINKED ISSUE #3715: Tempus: Add Stepper Initial Conditions
Adding a Stepper initial conditions has several benefits:
 * Ensure that the initial SolutionState is consistent, i.e., x, xDot, and xDotDot satisfy the governing equation at the initial time.
 * Currently explicit Steppers are not consistent until the next time step fills in the currentState.  This change will make the SolutionState consistent within the time step.
 * For explicit Steppers with a consistent SolutionState, the residual can be determined and compared to residuals from implicit Steppers.  This is SPARC's main driver for this change.
 * For DAEs, making the initial condition consistent is critical to obtain the correct solution. 

@trilinos/tempus 

## Expectations
 * Changes for this issue should not effect solutions.

## Definition of Done
- [x] First convert Forward Euler.
- [x] Convert all other explicit Steppers, ERK, Leapfrog, ...
- [x] Convert all implicit Steppers
- [x] Passes all tests.

ccober6:
@trilinos/tempus 

## Description

In IntegratorBasic::initializeSolutionHistory() and
IntegratorBasic::setSolutionHistory(), setInitialConditions() calls
the Stepper to get the first SolutionState set to the initial
conditions and make sure the initial conditions are "consistent",
i.e., x, xDot and xDotDot satisfy the governing equations.

 * Change name of IntegratorBasic::setInitialState() to
   IntegratorBasic::initializeSolutionHistory() and also
   in all derived classes and tests.

 * Added consistency parameter "Initial Condition Consistency" which
   indicates which type of consistency should be applied to
   the initial conditions (ICs):
     'None'   - Do nothing to the ICs provided in the SolutionHistory.
     'Zero'   - Set the derivative of the SolutionState to zero in the
                SolutionHistory provided, e.g., xDot^0 = 0, or
                xDotDot^0 = 0.
     'App'    - Use the application's ICs, e.g., getNominalValues().
     'Consistent' - Make the initial conditions for x and xDot
                consistent with the governing equations, e.g.,
                xDot = f(x,t), and f(x, xDot, t) = 0.  For implicit
                ODEs, this requires a solve of f(x, xDot, t) = 0 for
                xDot, and another Jacobian and residual may be
                needed, e.g., boundary conditions on xDot may need
                to replace boundary conditions on x.
   For explicit ODEs, the default is 'Consistent', because it is one
   additional evaluation.  For implicit ODEs, the default is 'None',
   because the application very often knows its IC and is easier and
   cheaper than the solve for 'Consistent'.

 * Added consistency check parameter "Initial Condition Consistency Check"
   which will check if the initial condition, x and xDot, is consistent
   with the governing equations, xDot = f(x,t), or f(x, xDot, t) = 0.
   The default value is true.

 * Started the migration to new "ModelEvaluator" interface by changing
   the solve and evaluate interface.  For implicit steppers, the solve
   call is solveImplicitODE(x, xDot, time, p), which closely matches the
   mathematics, f(x, xDot, time, p).  "p" are the parameters needed by
   the models, and currently includes :
     - TimeDerivative object which defines the time derivative for the stepper
     - Time step size
     - Stage number for Runge-Kutta methods
     - Alpha, the derviative of d(xDot)/d(x)
     - Beta, the derviative of d(x)/d(x) (often = 1)
     - EvaluationType, an enum to indicate how to evaluate the model
         . EVALUATE_RESIDUAL      - Evaluate residual for the implicit ODE
         . SOLVE_FOR_X            - Solve for x and determine xDot from x.
         . SOLVE_FOR_XDOT_CONST_X - Solve for xDot keeping x constant (for ICs).
             Note that this makes Alpha = d(xDot)/d(xDot) = 1 and
             Beta = d(x)/d(xDot) = 0.
   The implicit interface also has a residual evaluation call,
   evaluateImplicitODE(f, x, xDot, time, p), to evaluate f(x, xDot, t, p) = r
   for the initial condition consistency check.  For explicit steppers,
   the evaluate call is evaluateExplicitODE(xDot, x, time), which closely
   matches the mathematics, xDot = f(x, t).

 * Added new methods, getUseFSAL() and setUseFSAL(), to indicate when
   the Stepper can use the First-Step-As-Last (FSAL) principle, where the
   last function evaluation, f(x^{n-1},t^{n-1}) [a.k.a. xDot^{n-1}],
   can be used for the first function evaluation, f(x^n,t^n).  Often the
   Stepper can reuse xDot^{n-1}, but an exception is with operator
   splitting when xDot^{n-1} is invalid due to other Steppers changing
   the solution.

 * With "Use FSAL"=true, the time levels have consistent x, xDot and xDotDot,
   so isSynced=true for the state.  For "Use FSAL"=false, x, xDot and xDotDot
   are not necessarily at the same time level, so isSynced=false.

 * Added new methods to get/set x, xDot, and/or xDotDot to the Stepper,
   (getStepperX(), getStepperXDot() and getStepperXDotDot())
   which will return an RCP to either SolutionState memory or Stepper
   allocated memory. This reduces the memory needs for the solution in
   SolutionState while provides the Stepper with minimum required memory.

Miscellaneous Mods:

 * Create StepperExplicit base class.  Base class many of the functions
   from explicit steppers.
 * Added a default constructor for SolutionState, which does not
   set the solution vectors, x, xdot and xdotdot, which should be set
   via setX(), setXDot(), and/or setXDotDot() prior to being added
   to SolutionHistory.
 * Removed unused constructors for SolutionState.
 * Removed experimental and unused SolutionState::swapSolutonData().
 * Added set methods for x, xDot and xDotDot in SolutionState, which
   can be used with the default constructor.
 * SolutionState::copySolutionData() now clones x, xDot and xDotDot
   when those vectors are null.
 * Added new method to set the solution status based on the solver
   return status, e.g., workingState->setSolutionStatus(sStatus).

## Motivation and Context

This change adds the ability to set and make consistent the initial conditions (x, xDot and xDotDot) and use the FSAL principle, which can reduce the computational costs.

## Related Issues

* Closes #3715 

## How Has This Been Tested?
Passed all tests.

 * Explicit Steppers
   - General defaults unless noted
       . "Initial Condition Consistency" = "Consistent"
       . "Initial Condition Consistency Check" = true
   - ForwardEuler:
       . Default ("Use FSAL"=true) reproduces previous solutions, and
         "Use FSAL"=false works and is tested with StepperOperatorSplitting.
   - ExplicitRK:
       . Default ("Use FSAL"=false) reproduces previous solutions, and
         "Use FSAL"=true works and is tested with Bogacki-Shampine 3(2) method.
   - Leapfrog:
       . Default ("Use FSAL"=false) reproduces previous solutions, and
         "Use FSAL"=true will also work (i.e., no-op), but issue a warning
         that it will have no affect.
   - NewmarkExplicitAForm:
       . Default ("Use FSAL"=true) reproduces previous solutions, and
         "Use FSAL"=false works and is tested with SinCos, BallParabolic
         and HarmonicOscillar problems.

 * Implicit Stepper
   - General defaults unless noted
       . "Use FSAL"=false
       . "Initial Condition Consistency" = "None"
       . "Initial Condition Consistency Check" = false
   - BDF2:
   - BackwardEuler:
       . "Initial Condition Consistency Check" = true for CDR and
         van der Pol test.
   - DIRK:
   - HHTAlpha:
       . Not fully updated because it does not test beyond Newmark.
         Likely will be replaced by Generalize-Alpha method stepper.
   - IMEX_RK_Partition:
   - IMEX_RK:
   - NewmarkImplicitAForm:
       . Default ("Use FSAL"=true) reproduces previous solutions.
   - NewmarkImplicitDForm:
       . Not fully updated because it does not have any tests.
         Likely will be replaced by Generalize-Alpha method stepper.
   - Trapezoidal:
       . Required ("Use FSAL"=true) reproduces previous solutions.

 * OperatorSplit Stepper
   - In most cases, subSteppers cannot use xDotOld, thus the default is
     "Use FSAL"=false.  In some cases, the xDotOld can be used and save
     compute cycles.  The user can set this for each subStepper.

 * StaggeredForwardSensitivity
   - Not fully updated because constructor needs reworking.

 * CDR test needed modification to handle solving for xDot.  Solving
   for consistent initial conditions (i.e., solving for xDot) can
   change the Jacobian, especially wrt boundary conditions (BCs).
   Often BCs are in terms of the solution vector, x.  If the boundary
   condition are included in the Jacobian, solving for xDot could
   create a singular matrix, as the boundary value, x, is not a
   function of xDot.

 * Fix Piro Test due to Tempus changes. Piro_TempusSolver_UnitTests.cpp

## Checklist


ccober6:
Pinging interested individuals @tcfisher @ikalash @egphill @etphipp

sconde:
> Pinging interested individuals @tcfisher @ikalash @egphill @etphipp

and @rppawlo

sconde:
`Use FSAL` has to be defined in all the PL. We should be using the default value. Almost all the Drekar/Tempus tests are failing because of this issue. I suspect that other App might run into this same problem

packages/tempus/src/Tempus_StepperDIRK_decl.hpp
```diff
@@ -79,6 +79,10 @@ namespace Tempus {
  *   - end for
  *   - \f$x_n \leftarrow x_{n-1} + \Delta t\,\sum_{i=1}^{s}b_i\,\dot{X}_i\f$
  *   - Solve \f$f(\dot{x}_n,x_n,t_n)=0\f$ for \f$\dot{x}_n\f$ [Optional]
+ *
```
sconde: I'm not sure I understand this comment for DIRK stepper

packages/tempus/src/Tempus_StepperExplicit_decl.hpp
```diff
@@ -0,0 +1,137 @@
...
+    virtual bool isImplicit()         const {return false;}
+    virtual bool isExplicitImplicit() const
+      {return isExplicit() and isImplicit();}
+    virtual bool isOneStepMethod()    const {return true;}
+    virtual bool isMultiStepMethod()  const {return !isOneStepMethod();}
+
+    virtual bool getEmbedded() const
+      { return stepperPL_->get<bool>("Use Embedded", false); }
+    virtual void setUseFSAL(bool a) {stepperPL_->set<bool>("Use FSAL", a);}
+    virtual bool getUseFSAL() const {return stepperPL_->get<bool>("Use FSAL");}
```
sconde: should return a default value of `false`

packages/tempus/src/Tempus_StepperExplicit_decl.hpp
```diff
@@ -0,0 +1,137 @@
...
+    virtual void setICConsistency(std::string s)
+      {stepperPL_->set<std::string>("Initial Condition Consistency", s);}
+    virtual std::string getICConsistency() const
+      {return stepperPL_->get<std::string>("Initial Condition Consistency");}
```
sconde: should return a default value of `None`

packages/tempus/src/Tempus_StepperExplicit_decl.hpp
```diff
@@ -0,0 +1,137 @@
...
+    virtual void setICConsistencyCheck(bool c)
+      {stepperPL_->set<bool>("Initial Condition Consistency Check", c);}
+    virtual bool getICConsistencyCheck() const
+      {return stepperPL_->get<bool>("Initial Condition Consistency Check");}
```
sconde: should return a default value of `false`

packages/tempus/src/Tempus_StepperExplicit_impl.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+template<class Scalar>
+void StepperExplicit<Scalar>::setNonConstModel(
+  const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& appModel)
+{
+  this->setModel(appModel);
+}
+
+template<class Scalar>
+void StepperExplicit<Scalar>::setSolver(std::string solverName)
```
sconde: Does this mean that explicit stepper can now overlook the solver PL if it exists?

packages/tempus/src/Tempus_StepperForwardEuler_decl.hpp
```diff
@@ -54,54 +68,30 @@ class StepperForwardEuler : virtual public Tempus::Stepper<Scalar>
...
-      getModel(){return appModel_;}
-
-    virtual void setSolver(std::string solverName);
-    virtual void setSolver(
-      Teuchos::RCP<Teuchos::ParameterList> solverPL=Teuchos::null);
-    virtual void setSolver(
-        Teuchos::RCP<Thyra::NonlinearSolverBase<Scalar> > solver);
-    virtual Teuchos::RCP<Thyra::NonlinearSolverBase<Scalar> > getSolver() const
-      { return Teuchos::null; }
+    //virtual Teuchos::RCP<const Thyra::ModelEvaluator<Scalar> >
```
sconde: Is this needed anymore?

packages/tempus/test/Leapfrog/Tempus_Leapfrog_SinCos.xml
```diff
@@ -37,7 +37,7 @@
 
     <ParameterList name="Default Stepper">
       <Parameter name="Stepper Type"           type="string" value="Leapfrog"/>
-      <Parameter name="Is xDotDot Initialized" type="bool"   value="false"/>
+      <Parameter name="Use FSAL"       type="bool" value="true"/>
```
sconde: How is this change consistent?

packages/tempus/src/Tempus_StepperDIRK_decl.hpp
ccober6: In many of the stepper algorithms, I mention the possibility of updating xDot at the end of the step, though in most instances we do not have that code implemented.  I have been debating on removing these notes until that code is implemented and working.  I think I will remove them to reduce confusion.

packages/tempus/src/Tempus_StepperExplicit_decl.hpp
ccober6: Got it!

packages/tempus/src/Tempus_StepperExplicit_decl.hpp
ccober6: Got it!

packages/tempus/src/Tempus_StepperExplicit_decl.hpp
ccober6: I went with true, because it is just a residual evaluation and could catch some errors.

packages/tempus/src/Tempus_StepperExplicit_impl.hpp
ccober6: I think the answer is yes.  If not we should make it work that way.

packages/tempus/src/Tempus_StepperForwardEuler_decl.hpp
ccober6: Yeah, I went through and removed many functions that were covered by the StepperExplicit base class.  Good catch!

ccober6:
Ok, I think got all the changes in with 134b49e2c0c832904502e2ef3f97094cc6ad5e20, and I pushed it to the tempus-add-stepper-initial-conditions-3715 branch.  Please take a look and let me know if you see anything else.

Thanks for the help!

packages/tempus/src/Tempus_StepperExplicit_impl.hpp
sconde: I just tested this in Drekar. Tempus did not ignore the solver name or the solver PL. 

```
435: p=1 | *********** Caught Exception std::exception: Begin Error Report ***********¬
435: p=1 | Error, the parameter {name="Default Solver",type="ParameterList",value="..."}¬
435: p=1 | in the parameter (sub)list "Drekar Parameters->Tempus->Euler Example"¬
435: p=1 | was not found in the list of valid parameters!¬
435: p=1 |·¬
435: p=1 | The valid parameters and types are:¬
435: p=1 |   {¬
435: p=1 |     "Use FSAL" : bool = 0¬
435: p=1 |     "Initial Condition Consistency" : string = None¬
435: p=1 |     "Initial Condition Consistency Check" : bool = 1¬
435: p=1 |     "Use Embedded" : bool = 0¬
435: p=1 |     "Description" : string = RK Explicit 2 Stage 2nd order by Runge¬
435: p=1 | Also known as Explicit Midpoint¬
435: p=1 | Solving Ordinary Differential Equations I:¬
435: p=1 | Nonstiff Problems, 2nd Revised Edition¬
435: p=1 | E. Hairer, S.P. Norsett, G. Wanner¬
435: p=1 | Table 1.1, pg 135¬
435: p=1 | c = [  0  1/2 ]'¬
435: p=1 | A = [  0      ]¬
435: p=1 |     [ 1/2  0  ]¬
435: p=1 | b = [  0   1  ]'¬
435: p=1 |·¬
435: p=1 |     "Stepper Type" : string = RK Explicit 2 Stage 2nd order by Runge¬
435: p=1 |   }¬
435: p=1 |·¬
435: p=1 |·¬
435: p=1 | Throw number = 1¬
```

It's not an issue with this PR but I think it's somethign we should definitely consider

sconde:
@ccober6 thank you

packages/tempus/src/Tempus_StepperExplicit_impl.hpp
ccober6: Yeah, let talk this through and create another issue.

sconde:
Good with Drekar

ccober6:
OK @sconde, I just did a merge from some other commits, and it is ready for another review.   

You can use "git diff 585794d..5cd3e42 tempus/src/." to see the differences.  This includes the changes from the merge with "Tempus: Add ability to get alpha and beta from Steppers. #4635", but I could not refine it any better.  :(

Appreciate you reviewing this one last time (hopefully!).

PR #4519, opened 2019-02-28
[WIP] Tpetra: CrsGraph::insertGlobalIndicesImpl does a sorted/merged insert…

tjfulle:
… for StaticProfile graph.

Issues: #4278.

@mhoemmen this is my idea for doing a merged and sorted insert.  There is still much more to do, but it’s a start!  Your feedback is welcome before I sink more time in to the rest of CraGraph and then matrix.

@trilinos/tpetra


mhoemmen:
Awesome work @tjfulle !  I want to take a closer look at this after I give my talk this morning.

tjfulle:
@mhoemmen, there is quite a change in philosophy from the previous way, if you want to talk on the phone instead of trying to figure it out, feel free to call.  I’ve got other stuff to do, so no hurry :)

tjfulle:
@mhoemmen I put in a lot of clarifying comments and one question for you in `Tpetra_CrsGraph_def.hpp`.  Search for `@mfh`

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -0,0 +1,401 @@
+// @HEADER
```
tjfulle: Ugh, in a previous commit, this file showed up as a rename of `Tpetra_Details_padCrsArrays` (which it is).  But, this latest commit added a lot of comments, perhaps fooling git in to thinking it is not a rename but a new file?  At any rate, the diff is bigger than it need to be because much of the code is from the old file.

tjfulle:
Just added unit tests for all new functionality

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -0,0 +1,404 @@
...
+    return 0;
+
+  // Rather than loop through the two arrays of indices and looking for their
+  // difference and then checking if their is enough capacity in indices to
+  // accommodate the new elements (not already in indices, excluding
+  // duplicates), we go ahead and merge the two arrays in to one. After the
+  // arrays are merged, we can check if the length of the merged array exceeds
+  // the capacity of the indices array. If it does, return -1 to indicate the
+  // error. If it does not, however, we can just copy the merged elements back
+  // in to \c indices.
```
mhoemmen: OK, I think I see the value here of allocating a new array to return from `ind_merge` -- the idea is that you want to know whether there is enough space before destructively modifying the graph.  I'm OK with that.

packages/tpetra/core/test/Utils/TpetraUtils_crsUtils.cpp
```diff
@@ -0,0 +1,309 @@
...
+  using execution_space = typename device_type::execution_space;
+  using ordinal_type = size_t;
+  using view_type = Kokkos::View<ordinal_type*, device_type>;
+  using size_type = typename view_type::size_type;
+
+  ordinal_type num_row = 4;
+  ordinal_type num_indices_per_row = 5;
+  ordinal_type num_indices = num_indices_per_row * num_row;
+  auto row_ptrs_beg = view_type("beg", num_row+1);
+  // this assumes UVM
```
mhoemmen: You should insert a fence here.  View's constructor will fill on device, asynchronously, so your subsequent writes on host may conflict.

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -0,0 +1,404 @@
...
+/// \param [in] indices1 - the first array of indices
+/// \param [in] n1 - the length of indices1
+/// \param [in] indices2 - the second array of indices
+/// \param [in] n2 - the length of indices2
+///
+/// \return merged - The sorted and unique elements
+///
+template <class Ordinal>
+Teuchos::Array<Ordinal>
+ind_merge(
```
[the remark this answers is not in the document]
tjfulle: The user’s may not necessarily give a sorted array, so we’d have to create sorted copy anyway.  That’s why I implemented this `argsort` strategy - we can do the same work we’d do to count duplicates and extra capacity and then determine if we can insert all the new indices - without destroying the original graph if we don’t have the capacity

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -2099,243 +2099,69 @@ namespace Tpetra {
...
-        return numNewToInsert; // all done!
-      } // if the graph is StaticProfile
-      else { // DynamicProfile
+    if (this->getProfileType () == StaticProfile) {
+      auto numEntries = rowInfo.numEntries;
+      auto numAllocated = rowInfo.allocSize;
+      int numInserted = 0;
+      // @mfh: Is this right? We should just try to insert, even if k_gblInds1D_
+      // has length 0 (eg, empty graph). In that case, insertCrsIndices below
+      // will return -1 indicating not enough room for the new elements and we
```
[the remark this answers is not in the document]
tjfulle: I think this check was originally in to differentiate static from dynamic graph but, as we discussed over email, I think this is the wrong thing to do anyway.

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
tjfulle: Right.  To use any of the `std` sort or merge procedures we’d have to allocate a sorted copy of the user’s input array, so why not just create a new super merged array - which allows us to leave the original graph unmodified if there is not capacity for the new stuff, eh?

packages/tpetra/core/test/Utils/TpetraUtils_crsUtils.cpp
mhoemmen: Also, consider allocating the View without initializing, since you're going to fill it anyway (and it's a device View).

tjfulle:
@mhoemmen, I've modified `setAllIndices` to sort indices.  Besides `insertGlobalIndicesImpl` are there other procedures that need to be modified to guarantee that indices are sorted?

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -0,0 +1,376 @@
...
+///   sorted. The intended use case is the first range contains new elements to
+///   merge in to the existing CRS indices contained in the second range - which
+///   should already be sorted.
+///
+/// \param first1, last1 - first range of elements to merge (need not be sorted)
+/// \param first2, last2 - second range of elements to merge (must be sorted)
+///
+/// \return last - Output iterator to element past the last unique element in
+///   the merged range.
+///
```
tjfulle: @mhoemmen, this new version of `ind_merge` uses `inplace_merge`.  This version seems much cleaner than before.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -2099,243 +2099,69 @@ namespace Tpetra {
...
-             ") = " << chkNewNumEntries << " != newNumEntries = " << newNumEntries
-             << ".  Please report this bug to the Tpetra developers.");
-        }
-        return numNewToInsert; // all done!
-      } // if the graph is StaticProfile
-      else { // DynamicProfile
+    if (this->getProfileType () == StaticProfile) {
+      auto numEntries = rowInfo.numEntries;
+      auto numAllocated = rowInfo.allocSize;
+      int numInserted = 0;
```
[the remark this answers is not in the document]
tjfulle: `LO` may not be known by `insertCrsIndices`, but I changed the return type to `make_signed<typename T::value_type>::type`.  it may be `make_signed<GO>::type`.  But, the return from `insertGlobalIndicesImpl` is `size_t`, so whatever `insertCrsIndices` returns, it must be converted to `size_t` anyway.

mhoemmen:
@tjfulle wrote:
> @mhoemmen, I've modified `setAllIndices` to sort indices.

@trilinos/muelu uses `setAllIndices` a lot, and it's likely that MueLu presorts.  @csiefer2 @jhux2 et al., should we add a flag to `setAllIndices` to let you specify that the indices are already presorted?

jhux2:
@mhoemmen wrote:
> @tjfulle trilinos/muelu uses `setAllIndices` a lot, and it's likely that MueLu presorts. @csiefer2 @jhux2 et al., should we add a flag to `setAllIndices` to let you specify that the indices are already presorted?

You're right, MueLu uses this quite a bit.  Yes, adding such a flag is a good idea.

mhoemmen:
> You're right, MueLu uses this quite a bit. Yes, adding such a flag is a good idea.

Would you prefer that setAllIndices just assumed that the indices are presorted?  This is really a routine for MueLu, which seems to prefer multistage initialization of matrices instead of just calling a constructor that takes all the data.

tjfulle:
I may put this on hold.  @mhoemmen and I are discussing a different approach that does not require indices be sorted and can be abstracted to both global and local indices.

mhoemmen:
The question about `setAllIndices` is still valid.  On the other hand, Tpetra has to copy the `Teuchos::ArrayRCP` inputs anyway, since there's no nice way to make `Kokkos::View` use & track `ArrayRCP` allocations.

tjfulle:
@mhoemmen, the latest push implements ideas from our email conversation.  Specifically `Details::insertCrsIndices` now 

- looks to see if an incoming index is already in the array of indices
- if it is, record the relative offset to the existing index
- if it is not, insert the new index at the end of the array of indices (if there is capacity)
- calls an optional callback function with the offset.

I modified the behavior of the previous incarnation of `insertCrsIndices` which was templated on the ordinal type (so that it can be called with local or global indices).  Untemplating the function and implementing separate local and global ordinal versions won't be a problem.

tjfulle:
I also added a test of the callback functionality.

tjfulle:
Made obsolete by #4733

PR #4527, opened 2019-02-28
Amesos2: Fix another code path for gapped row id case

ndellingwood:
Partially addresses issue #3647 reported by @vbrunini

### DIFF
--- a/packages/amesos2/src/Amesos2_Umfpack_def.hpp
+++ b/packages/amesos2/src/Amesos2_Umfpack_def.hpp
@@ -63,6 +63,7 @@ Umfpack<Matrix,Vector>::Umfpack(
   , nzvals_()                   // initialize to empty arrays
   , rowind_()
   , colptr_()
+  , is_contiguous_(true)
 {
   data_.Symbolic = 0;
   data_.Numeric = 0;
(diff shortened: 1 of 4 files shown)


ndellingwood:
Spoke to @vbrunini , this fixed issues with SuperLU and KLU2; next step is to update the UMFPACK interface to support the non-contiguous GID parameter setting. Marking this as WIP until the UMFPACK interface update is pushed in.

ndellingwood:
@vbrunini I pushed 2 more commits, one adds non-contiguous GID support to the Umfpack interface. I tested the GappedMtxGIDs-1proc example and it now runs with correct results using Umfpack as the solver. I also updated the GappedMtxGIDs-1proc example to rerun the solve with a new RHS but reuse the previous solve's symbolic and numeric results to add coverage for the missed case in my previous PR.

packages/amesos2/src/Amesos2_Umfpack_def.hpp
```diff
@@ -250,6 +275,8 @@ Umfpack<Matrix,Vector>::getValidParameters_impl() const
   if( is_null(valid_params) ){
     Teuchos::RCP<Teuchos::ParameterList> pl = Teuchos::parameterList();
 
+    pl->set("IsContiguous", true, "Whether GIDs contiguous");
```
srajama1: nit pick -> "GIDs are contiguous"

ndellingwood:
@srajama1 this PR completes the majority of the issues from #3647, could you review when you get a chance? @vbrunini confirmed the fix with SuperLU. I'm having issues building and testing with SuperLU_Dist, so if there are problems there I will take care of it in a follow up PR, as the SLU_Dist interface may require additional updates as well.

packages/amesos2/src/Amesos2_Umfpack_def.hpp
ndellingwood: Good catch, looks like auto-tester already started so I'll clean this up while addressing more issues. Thanks for quick review!!

Issue #3978, opened 2018-12-03
amesos2 klu2_internal.h defining NDEBUG


crdrumm:
@trilinos/amesos2

## Current Behavior

In klu2_internal.h the following lines define NDEBUG which disables assertions in SCEPTRE unit tests

/* make sure debugging and printing is turned off */

#ifndef NDEBUG
#define NDEBUG
#endif

/* To enable debugging and assertions, uncomment this line:
 #undef NDEBUG
 */

To be more specific, we have the following line in our code:

#include "Amesos2.hpp"

That file has the following chain of #includes:
#include "Amesos2_Factory.hpp"
#include "Amesos2_KLU2.hpp"
#include "Amesos2_KLU2_decl.hpp"
#include "Amesos2_KLU2_FunctionMap.hpp"
#include "klu2_defaults.hpp"
#include "klu2_internal.h"

In particular, the file Amesos2_KLU2_FunctionMap.hpp has the following section:

/* External definitions of the KLU2 functions
 *
 */
 // TODO
namespace KLU2 {
#include "klu2_defaults.hpp"
#include "klu2_analyze.hpp"
#include "klu2_factor.hpp"
#include "klu2_solve.hpp"
#include "klu2_tsolve.hpp"
#include "klu2_free_symbolic.hpp"
#include "klu2_free_numeric.hpp"
} // end namespace KLU2

The namespace does nothing to protect against preprocessor commands, which I suspect was the intention of the KLU2 namespace.

## Motivation and Context
assertions have been unexpectedly disabled, so tests compiled with -g erroneously passed

## Your Environment

version 12.12.1

  -D Trilinos_VERBOSE_CONFIGURE:BOOL=ON \
  -D CMAKE_VERBOSE_MAKEFILE:BOOL=TRUE \
  -D CMAKE_BUILD_TYPE:STRING=RELEASE \
  -D CMAKE_CXX_FLAGS="-O3" \
  -D TPL_ENABLE_BinUtils:BOOL=OFF \
  -D TPL_ENABLE_MPI:BOOL=ON \
  -D MPI_BASE_DIR:PATH=$MPIHOME \
  -D TPL_ENABLE_BLAS:BOOL=ON \
  -D TPL_ENABLE_LAPACK:BOOL=ON \
  -D LAPACK_LIBRARY_NAMES:STRING="mkl_core;mkl_sequential;mkl_intel_lp64" \
[...]

PR #4528, opened 2019-02-28
Amesos2: Rename KLU2 NDEBUG Macro

ndellingwood:
Address issue #3978


ndellingwood:
@srajama1 I renamed the macros causing the problems reported in #3978 as a fast fix. If we'd like to introduce KLU2-specific debug macro options to enable within Amesos2 I would prefer we do that in a follow up PR.

Issue #4457, opened 2019-02-21
Building Sacado with relocatable device code enabled


jwillenbring:
@trilinos/sacado 
CC: @trilinos/framework @bartlettroscoe @rppawlo @alanw0 

## Expectations

Multiple major Trilinos customers now need to build Trilinos with relocatable device code enabled for CUDA builds. Per @rppawlo this can take a lot of effort to get working. 

## Current Behavior

I fired off a build with rdc enabled to help gauge our current status and encountered [these issues](https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&parentid=4589862) with Sacado (note that the system caps errors at 50).

## Definition of Done

I had two objectives in filing this ticket. 

1. Post the current Sacado issues I am seeing so that they can be evaluated and possibly fixed.

1. Start a broader discussion about the effort that might be required to get rdc working across Trilinos.

## Steps to Reproduce

I used the jwillenbring/trilinos rdc_test branch and the standard PR CUDA build on ride to produce the errors. A fix can be tested by submitting a PR, or I would be willing to test a branch that someone points me to.

rppawlo:
RDC is not the most stable thing. Long term we will probably need two cuda builds for the PR testing - one with RDC enabled and one disabled. Failures are usually build errors with RDC, so maybe we get away with building but not running the tests for one of the two builds?

etphipp:
This is a duplicate of issue #2891, and this has nothing to do with Sacado per se.  It is caused by libteuchoscore.a, which defines some static symbols, appearing twice on the link line.  Reading through the comments on that issue, it is unclear if it was ever determined if this is a bug in CMake, Tribits, or nvcc/nvcc_wrapper.

bartlettroscoe:
FYI: I will go ahead and set up some nightly "Specialized" ATDM Trilinos CUDA 9.2 RDC builds on 'white'/'ride' and 'waterman'.  If you want to get updates on this, add yourself as a watcher to  [TRIL-262](https://sems-atlassian-son.sandia.gov/jira/browse/TRIL-262).

vbrunini:
nvcc_wrapper does have a "--remove-duplicate-link-files" option that helps a lot working around these issues.

etphipp:
PR #4546 fixed the dependencies for Sacado and I verified it eliminates the duplication of libteuchoscore.a on the link line, so I am closing this issue.

Issue #2891, opened 2018-06-05
Sacado: CUDA build errors in test with RDC ON


mhoemmen:
@trilinos/sacado 

With a static CUDA build and with relocatable device code (RDC) enabled, I get the following build errors:
```
/tmp/tmpxft_00001079_00000000-2_Sacado_Fad_Kokkos_CommTests_Serial_dlink.reg.c: At global scope:
.../nvidia/cuda_9.0.176/bin/crt/link.stub:92:118: error: redefinition of ‘const unsigned char def_module_id_str_48_tmpxft_00007161_00000000_6_Teuchos_Utils_cpp1_ii_dbeaaeb9 []’
 static const unsigned char __DEFSTRNAME(id) [] __attribute__((aligned(1))) __attribute__((section ("__nv_module_id"))) = "def " __TO_STRING(id); \
                                                                                                                      ^
/tmp/tmpxft_00001079_00000000-2_Sacado_Fad_Kokkos_CommTests_Serial_dlink.reg.c:145:1: note: in expansion of macro ‘DEFINE_REGISTER_FUNC’
 DEFINE_REGISTER_FUNC(_48_tmpxft_00007161_00000000_6_Teuchos_Utils_cpp1_ii_dbeaaeb9)
 ^
```
Have y'all ever seen those?  Here are my relevant build options:
```
-D CMAKE_BUILD_TYPE=RELEASE
-D BUILD_SHARED_LIBS=OFF
-D Trilinos_ENABLE_EXPLICIT_INSTANTIATION=ON
-D Trilinos_ENABLE_OpenMP=OFF
-D TPL_ENABLE_Pthread=OFF
-D Trilinos_ENABLE_Sacado=ON
-D Trilinos_ENABLE_Stokhos=OFF
-D Kokkos_ENABLE_Serial=ON
-D Kokkos_ENABLE_OpenMP=OFF
-D Kokkos_ENABLE_Pthread=OFF
-D Kokkos_ENABLE_Cuda=ON
-D Kokkos_ENABLE_Cuda_UVM=ON
-D Kokkos_ENABLE_Cuda_Lambda=ON
-D Kokkos_ENABLE_Cuda_Relocatable_Device_Code=ON
-D TPL_ENABLE_CUDA=ON
-D KOKKOS_ARCH="Pascal60"
```
I'm using CUDA 9.0.176, GCC 4.9.3, OpenMPI 1.10.2, and CMake 3.8.2.

[...]

Issue #1263, opened 2017-04-25
Sacado: change dependency to Teuchos sub-packages


ibaned:
Please refer to TriBITSPub/TriBITS#188 for the details on why it is bad to list just `Teuchos` as a dependency of `Sacado`. This issue is to request that Sacado developers change their dependencies to list teuchos sub-packages like `TeuchosCore`, and change the `#ifdef`s in the code to use `HAVE_SACADO_TEUCHOSCORE` instead of `HAVE_SACADO_TEUCHOS`, for example.

@trilinos/sacado

bartlettroscoe:
Looking at:

```
$ cat packages/sacado/cmake/Dependencies.cmake 
SET(LIB_REQUIRED_DEP_PACKAGES)
SET(LIB_OPTIONAL_DEP_PACKAGES Teuchos KokkosCore TeuchosKokkosComm)
SET(TEST_REQUIRED_DEP_PACKAGES)
SET(TEST_OPTIONAL_DEP_PACKAGES TeuchosCore KokkosContainers)
SET(LIB_REQUIRED_DEP_TPLS)
SET(LIB_OPTIONAL_DEP_TPLS)
SET(TEST_REQUIRED_DEP_TPLS)
SET(TEST_OPTIONAL_DEP_TPLS ADOLC ADIC CppUnit)
```

and

```
$ find packages/sacado -name "CMakeLists.txt" -exec grep -nH ENABLE_Teuchos {} \;
packages/sacado/test/UnitTests/CMakeLists.txt:10:IF (Sacado_ENABLE_Teuchos)
packages/sacado/test/performance/CMakeLists.txt:6:IF (${PACKAGE_NAME}_ENABLE_Teuchos AND NOT Kokkos_ENABLE_Cuda)
packages/sacado/test/performance/CMakeLists.txt:75:IF (Sacado_ENABLE_Teuchos AND Sacado_ENABLE_KokkosCore)
packages/sacado/test/performance/fenl_assembly_view/CMakeLists.txt:7:    Sacado_ENABLE_Teuchos)
packages/sacado/test/performance/fenl_assembly/CMakeLists.txt:7:    Sacado_ENABLE_Teuchos)
packages/sacado/test/TestSuite/CMakeLists.txt:16:IF (${PACKAGE_NAME}_ENABLE_Teuchos)
packages/sacado/example/CMakeLists.txt:126:IF(Sacado_ENABLE_Teuchos)
[...]

PR #4546, opened 2019-03-05
Properly handle Teuchos sub-package dependencies for Sacado

etphipp:
This fixes Sacado's dependencies on Teuchos to properly only depend on
the needed Teuchos sub-packages.  This involved a fair amount of code
reorganization because several optional dependencies were combined in
one file.  There is some danger in this in that I didn't test every
possible combination of Teuchos sub-packages.  However I did test no
packages, all packages, and enabling each package individually.

This should resolve the duplicate symbol errors observed in issues #2891
and #4457 by removing the duplicate link of teuchos-core.  This was also
requested in issue #1263.


etphipp:
Yes that would simplify things somewhat and I thought about doing something like that.  There aren't that many dependencies, so I just tried to do it properly.

PR #4549, opened 2019-03-05
Fix for Ifpack2 that causes SPARC slowness. 

kyungjoo-kim:
## Description

Last a few commits that I made for Ifpack2 cause slowness in SPARC. The reason for the slowness is surprisingly from some template instanciations. I cannot precisely explain why. Previously we use macros for blockwise specialization and I changed it to template implementation for doing the same thing. After changing it back to macro expressions, it recovers the previous performance. Sigh Sigh Sigh....  Mystery.... 

## How Has This Been Tested?

This is tested on Apollo and @vbrunini confirmed that it is fixed on SPARC.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -1930,6 +1930,11 @@ namespace Ifpack2 {
 #else
       typedef KokkosBatched::Experimental::Algo::Level3::Blocked multi_vector_algo_type;
 #endif
+      static int recommended_team_size(const int blksize, 
+				       const int vector_length, 
+				       const int internal_vector_length) { 
+	return 1; 
+      }
     };
     
 #if defined(KOKKOS_ENABLE_CUDA) 
@@ -1938,6 +1943,27 @@ namespace Ifpack2 {
       typedef KokkosBatched::Experimental::Mode::Team mode_type;
       typedef KokkosBatched::Experimental::Algo::Level2::Unblocked single_vector_algo_type;
       typedef KokkosBatched::Experimental::Algo::Level3::Unblocked multi_vector_algo_type;
+      static int recommended_team_size(const int blksize, 
+				       const int vector_length, 
+				       const int internal_vector_length) { 
+	const int vector_size = vector_length/internal_vector_length;
+    	const int total_team_size = blksize <= 8 ? 32 : blksize <= 16 ? 64 : 128;
+	return internal_vector_length == 1 ? 32/vector_size : total_team_size/vector_size; 
+      }
+    };
+    template<>
+    struct SolveTridiagsDefaultModeAndAlgo<Kokkos::CudaUVMSpace> {
+      typedef KokkosBatched::Experimental::Mode::Team mode_type;
+      typedef KokkosBatched::Experimental::Algo::Level2::Unblocked single_vector_algo_type;
+      typedef KokkosBatched::Experimental::Algo::Level3::Unblocked multi_vector_algo_type;
+      static int recommended_team_size(const int blksize, 
+    				       const int vector_length, 
+    				       const int internal_vector_length) { 
+    	const int vector_size = vector_length/internal_vector_length;
+    	const int total_team_size = blksize <= 8 ? 32 : blksize <= 16 ? 64 : 128;
+	// to match the old code behavior when internal vector length is 1
+	return internal_vector_length == 1 ? 32/vector_size : total_team_size/vector_size; 
+      }
     };
 #endif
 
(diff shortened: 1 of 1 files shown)


packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1592,41 +1624,60 @@ namespace Ifpack2 {
...
 	const local_ordinal_type packidx = member.league_rank();
 
 	const local_ordinal_type partidx = packptr(packidx);
 	const local_ordinal_type npacks = packptr(packidx+1) - partidx;
 	const local_ordinal_type i0 = pack_td_ptr(partidx);
 	const local_ordinal_type nrows = partptr(partidx+1) - partptr(partidx);
+	Kokkos::parallel_for
+	  (Kokkos::ThreadVectorRange(member, vector_length_value), [&](const local_ordinal_type &v) {
+	    if (v < npacks) extract(member, partidx+v, v);
+	    factorize(member, i0, nrows, v, scalar_values);
```
srajama1: Does one need a team barrier here ? Just making sure.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1592,41 +1624,60 @@ namespace Ifpack2 {
...
 	Kokkos::parallel_for
 	  (Kokkos::ThreadVectorRange(member, vector_loop_size), [&](const local_ordinal_type &v) {
             const local_ordinal_type vbeg = v*internal_vector_length;
             const local_ordinal_type vend = vbeg + internal_vector_length;
             for (local_ordinal_type vv=vbeg;vv<vend;++vv)
-              if (vv < npacks) extract(member, partidx+vv, vv);
-	    member.team_barrier();
-	    factorize(member, i0, nrows, v);
+	    if (vv < npacks) extract(member, partidx+vv, vv);
+	    factorize(member, i0, nrows, v, internal_vector_values);
```
srajama1: See above. Does one need a team barrier here ? Just making sure.

kyungjoo-kim: In practice, considering the thread mapping to matrices, it is safe to remove the barrier. In the LU factorization, the first element is mapped to the same thread and after first row is processed, it has a team barrier.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1592,41 +1624,60 @@ namespace Ifpack2 {
...
 	  });
       }
 
       void run() {
 #if defined(KOKKOS_ENABLE_CUDA) && defined(IFPACK2_BLOCKTRIDICONTAINER_ENABLE_PROFILE)
         cudaProfilerStart();
```
mhoemmen: We should think about wrapping these CUDA profiling API calls, so that we don't have so many macros.  Also, if we have CUDA enabled, couldn't we decide at run time whether or not to enable CUDA profiling?

kyungjoo-kim: That should be possible to check that in runtime using some environmental variables.

mhoemmen: @kyungjoo-kim We could use the existing `Tpetra::Details::Behavior` facility -- there's nothing specific about that to Tpetra.

kyungjoo-kim: When I profile performance, I only need to see one particular kernel at a time. Even though I put kernel names, the profiler shows too many kernels if I use the common variable. Potentially, it also shows cuda kernels from other packages, which is too much information.

mhoemmen: @kyungjoo-kim `Behavior` can decide behavior based on a comma-delimited list of strings, not just a Boolean setting.  We already use this feature to restrict what classes or functions should print verbose debugging output.  We could do that for profiling as well.

kyungjoo-kim: Oh... really ? Then, it will be fine. I will try to use that. Thanks.

mhoemmen: @kyungjoo-kim We would need to create a new environment variable and corresponding function.  Behavior tries not to do too much at run time, which is why it doesn't offer run-time look-up by environment variable.

kyungjoo-kim:
@mhoemmen Could you help me break ifpack2 code for seprate ETI ? Now whatever I change the code, it slow down the performance. I looked at the code, the eti mechanism is quite sophisticated.

mhoemmen:
> @mhoemmen Could you help me break ifpack2 code for seprate ETI ? Now whatever I change the code, it slow down the performance. I looked at the code, the eti mechanism is quite sophisticated.

I'm heading over there right now for the Tpetra meeting; I can show you how it works then.

kyungjoo-kim:
Something wrong.... I will work again.

kyungjoo-kim:
Fixed the problem on host architectures.

kyungjoo-kim:
@trilinos/rol @trilinos/framework This PR test fails from ride (cuda build) as one of ROL tests (Nonlinear Problem test). How can I merge this PR ? The test failure seems to be due to time out (10min).

kyungjoo-kim:
@dridzal Could you look at the error from ROL in the PR test ? This PR does not include anything about ROL. From the error message, I am not sure if this is a simple time out error or a real test failure.

jwillenbring:
@kyungjoo-kim 

This timeout has been popping up with other PRs also. We have not yet determined the best way to deal with it, but have been discussing it.

mhoemmen:
@micahahoward

kyungjoo-kim:
@jwillenbring 

How can I merge this PR then ? Do I just re-test again until it passes ? or do we have someone with a superpower who can merge the PR ?

kyungjoo-kim:
Lucky !!!

Issue #4515, opened 2019-02-27
Stokhos: Remove Build Warnings


ZUUL42:
@trilinos/stokhos 

## Expectations

Stokhos builds should not emit any warnings that will be promoted to errors once Werror is set in the GCC 7.2.0 automated build.
## Current Behavior

Currently Stokhos has a number of warnings that need to be handled before we can set Werror for all packages.
A recent test build was performed with -Werror set for Stokhos. The CDash report can be found [here](https://testing-vm.sandia.gov/cdash/viewBuildError.php?buildid=4624610).
## Motivation and Context

Issue #3178 is working toward turning Warnings as Errors on for all packages.

etphipp:
PR #4552 should have fixed the stokhos-related warnings/errors.  Some of them were coming from @trilinos/muelu.  Is there an updated cdash build where I can see if they are fixed?

ZUUL42:
I kicked off a build for 7.2.0 with Werror set for Stokhos.
The CDash results are coming together [here](https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=3&showfilters=1&filtercombine=and&field1=site&compare1=61&value1=ascic143&field2=buildname&compare2=61&value2=PR-UNKNOWN-test-Trilinos_pullrequest_gcc_7.2.0-673&field3=buildstamp&compare3=61&value3=20190306-1727-Pull).
CDash has been having some issues today, but it is building and I'll keep an eye on it directly.

ZUUL42:
@etphipp, looks like there's an unused-variable left.

lucbv:
@etphipp I think that there is no point in you starting this until the MueLu PR is merged in. Until that is done you will have warning coming out of the MueLu code...

@ZUUL42 you can see that the build is picking up the warning I am fixing right now. I am still waiting for [this build](https://testing.sandia.gov/cdash/buildSummary.php?buildid=4667508) to coming back without warnings and with all tests passing. Once that works, I will flag PR #4543 with `AT: AUTOMERGE` and when the PR does merge we can reattempt Stokhos

etphipp:
Right, I new the MueLu error would still be there.  But I wanted to make sure the other Stokhos errors were fixed, which it looks like they are.  So once they are fixed in MueLu, Stokhos should be fixed too.

PR #4552, opened 2019-03-05
Stokhos:  Fix a few warnings reported in issue #4515

etphipp:
Involves fixing a warning from the mask traits that was a true error.

For issue #4515


kliegeois:
It looks good!

PR #4554, opened 2019-03-06
Set up 'complex' builds on 'sems-rhel7' to support Gemma and EMPIRE (TRIL-255)

bartlettroscoe:
CC: @fryeguy52, @jmgate 

This sets up new ATDM Trilinos builds for the 'sems-rhel7' env that help to protect the ATDM code Gemma (see [TRIL-255](https://sems-atlassian-son.sandia.gov/jira/browse/TRIL-255)).   All of these enable 'complex' support.  In particular, there are cuda+complex builds that currently fail in Muelu that will need to be fixed.  These builds will run on 'ascicgpu14' run from jenkins-srn.sandia.gov.

These builds also help to support EMPIRE that has moved to use the 'sems-rhel7' env but we had no automated testing for the builds with the 'sems-rhel7' env posting to CDash. 

I did a good bit of testing as documented in [TRIL-264](https://sems-atlassian-son.sandia.gov/jira/browse/TRIL-264?focusedCommentId=26200&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26200).  Since these are mostly just setting up new builds these should not break existing builds.  The only risk is are the changes to the `ATDMDevEnvSettings.cmake` file with the addition of the optional `ATDM_CONFIG_CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS` var but I tested builds that did not set that so I am pretty confident that this will not break builds in other envs.


cmake/std/atdm/sems-rhel7/environment.sh
```diff
@@ -181,6 +181,29 @@ module load sems-zlib/1.2.8/base
...
+  export ATDM_CONFIG_USE_NINJA=OFF
+  export ATDM_CONFIG_CMAKE_CXX_USE_RESPONSE_FILE_FOR_OBJECTS=OFF
+fi
+# NOTE: The reason for the above logic is that 'nvcc' can't handle *.rsp
+# response files that CMake switches to when the command-lines become too long
+# and there is no way to turn off the usage of response files with the CMake
+# Ninja generator as of CMake 3.14.0.  The problem is that when CUDA and
+# complex are enabled and shared libs are used, 'nvcc' is used to create the
+# kokkoskernels shared lib which has a ton of object files that trigger
+# internal the CMake logic to condense these down into a single
```
mhoemmen: "internal the"

bartlettroscoe: Thanks for catching this!  I added a new PR to fix this and other typos in that long paragraph in PR #4558.

PR #4555, opened 2019-03-06
Tpetra: Fix #4542 (add computeRowOneNorms)

### LINKED ISSUE #240: Add  InvRowSums method to Tpetra
It looks like Epetra has an InvRowSums method, but not Tpetra.  Would it be possible to add the InvRowSums method to Tpetra?  It will faciliate some scaling implementations in Albany.

### LINKED ISSUE #4542: Tpetra: Add computeRowOneNorms function
@trilinos/tpetra 

@vbrunini requests that we add `Tpetra::computeRowOneNorms`.  His app has the option to scale linear systems only on the left side, only by row norms.  The existing `Tpetra::computeRowAndColumnOneNorms` function already has an option not to compute column norms (set `assumeSymmetric=true`), but it still communicates the row norms to the column Map distribution for symmetric scaling.  @vbrunini doesn't even want to communicate the row norms, since he doesn't plan to right-scale with them.

mhoemmen:
@trilinos/tpetra 

Per request by @vbrunini, add Tpetra:::computeRowOneNorms function.  @vbrunini wants to use it just for left-side scaling of a matrix.  @ikalash may appreciate this, as it closes #240 more directly than the current `Tpetra::computeRowAndColumnOneNorms` function, without additional computation or communication (if you only want to scale on the left).

## Related Issues

* Closes #4542 
* Related to #240

### DIFF
--- a/packages/tpetra/core/src/Tpetra_computeRowAndColumnOneNorms_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_computeRowAndColumnOneNorms_decl.hpp
@@ -52,11 +52,26 @@
 
 namespace Tpetra {
 
+/// \brief Compute global row one-norms ("row sums") of the input
+///   sparse matrix A, in a way suitable for one-sided (left only)
+///   equilibration.
+///
+/// \note This function is collective over A's communicator, and may
+///   need to communicate, depending on A's Maps.
+///
+/// \param A [in] The input sparse matrix A.
+///
+/// \return Input to leftAndOrRightScaleCrsMatrix (which see).  The
+///   result is only safe to use for left scaling, not for right
+///   scaling.
+template<class SC, class LO, class GO, class NT>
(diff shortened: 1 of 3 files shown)


mhoemmen:
I turned off automerge so that I can get @vbrunini 's feedback before merging.

mhoemmen:
@trilinos/framework uh oh :(

mhoemmen:
Victor likes it; I'm bestowing an autoblessing.

mhoemmen:
@trilinos/tpetra @trilinos/muelu y'all might want this.

mhoemmen:
Thanks @lucbv ! :-D  I would prefer to optimize the "scale both rows and columns" algorithm first, since it's more generally useful.  Also, a good multilevel parallelization should imitate Christian's sparse matrix-vector multiply, by grouping up rows into chunks and doing dynamic load balancing.

Issue #3566, opened 2018-10-04
Amesos2 + Superludist 5.3.0: compilation issues with float


davydden:
PackageName:  Amesos2
@trilinos/amesos2

I see a lot of compilation issues with Amesos2 when enabling
```
-DTpetra_INST_COMPLEX_FLOAT=ON
-DTeuchos_ENABLE_FLOAT=ON
```

## Expectations

Trilinos should build with parameters outline above.

## Current Behavior
Compiler errors:

```
/spack/stage/trilinos-12.12.1-i5lgamlpcrzplkscjaus6oe3wfv7ps6j/Trilinos-trilinos-release-12-12-1/packages/amesos2/src/Amesos2_Superludist_decl.hpp:115:30: error: no type named 'type' in 'Amesos2::TypeMap<Amesos2::Superludist, float>'
  typedef typename type_map::type                                  slu_type;
          ~~~~~~~~~~~~~~~~~~~^~~~
/spack/stage/trilinos-12.12.1-i5lgamlpcrzplkscjaus6oe3wfv7ps6j/Trilinos-trilinos-release-12-12-1/packages/amesos2/src/Amesos2_Superludist.cpp:62:30: note: in instantiation of template class 'Amesos2::Superludist<Tpetra::CrsMatrix<float, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>, false>, Tpetra::MultiVector<float, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>, false> >' requested here
  AMESOS2_SOLVER_TPETRA_INST(Superludist,float,int,int);
                             ^
In file included from /spack/stage/trilinos-12.12.1-i5lgamlpcrzplkscjaus6oe3wfv7ps6j/Trilinos-trilinos-release-12-12-1/packages/amesos2/src/Amesos2_Superludist.cpp:47:
/spack/stage/trilinos-12.12.1-i5lgamlpcrzplkscjaus6oe3wfv7ps6j/Trilinos-trilinos-release-12-12-1/packages/amesos2/src/Amesos2_Superludist_decl.hpp:116:30: error: no type named 'magnitude_type' in 'Amesos2::TypeMap<Amesos2::Superludist, float>'
  typedef typename type_map::magnitude_type                  magnitude_type;
          ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
[...]

PR #4567, opened 2019-03-07
Issue 3566

ndellingwood:
@trilinos/amesos2

## Description
    Amesos2: Fix for superludist when float is enabled
    
    Remove float and complex<float> from ETI since superludist only
    implements double and complex<double>
    Should address issue #3566

## Tested

Tested on MacOSX HighSierra, gcc/6.5, mpich, superludist 5.2.2


ndellingwood:
@srajama1 I'm guessing most people compiling with ETI use double or complex double, not float which allowed this to sneak through.

PR #4572, opened 2019-03-07
This disables the ROL_NonlinearProblemTest_MPI_4 test for cuda testing

prwolfe:
While this runs in about 10 seconds on CPU machines it is running
anywhere from 400-600 on CUDA equipment.  This needs to be properly
triaged and repaired before re-enabling it.

@trilinos/framework 
@trilinos/rol 

## Description
Disabled this test until the ROL team can address it. This will allow other teams to get pull requests tested in a reasonable time frame.

## How Has This Been Tested?
Since this is just a test disable I expect that the PR testing is the best way to confirm it works as expected.


prwolfe:
I inspected the PR builds and can confirm this turns off the test for the CUDA build only. Merging.

PR #4582, opened 2019-03-08
New Stokhos ETI system that splits instantiations into multiple files and libraries

etphipp:
This is a complete rework of the Stokhos ETI system.  It makes several
changes:
  * Each scalar type instantiation is put into a separate file for
    compilation.
  * Each set of instantiations for a given package (Tpetra, MueLu, ...)
    for each scalar type is put into a separate library.  This will keep
    the libraries much smaller to fix linker issues when the libraries
    get too big (related to issue #4502)
  * You can now set what ensemble sizes you want to instantiate with via
    the cmake option Stokhos_ETI_Ensemble_Sizes, e.g.,
       -D Stokhos_ETI_Ensemble_Sizes="4,8,16,32"
  * By default only an ensemble size of 16 is used.  This dramatically
    reduces the number of default instantiations, which will speed up
    build times and reduce library sizes.
  * ETI for MueLu in Stokhos now uses the same class list as the MueLu
    package, so they will no longer need to run a separate script for
    Stokhos when adding new classes.

This PR also includes some changes to get some MueLu files
to compile that previously weren't being instantiated with Stokhos.

It also eliminates a lot of old files that aren't needed anymore, and cleans up some warnings.

@trilinos/muelu


jhux2:
>     * Each set of instantiations for a given package (Tpetra, MueLu, ...)
>       for each scalar type is put into a separate library.  This will keep
>       the libraries much smaller to fix linker issues when the libraries
>       get too big (related to issue #4502)

@etphipp This is a good idea.  We might be able to do something similar for MueLu, e.g., for different node types.

etphipp:
@kliegeois wrote:
> It looks good.
> I think that the option -D Stokhos_ETI_Ensemble_Sizes="4,8,16,32" is a good idea.
> However, I am wondering, how does Stokhos behave if you ask -D Stokhos_ETI_Ensemble_Sizes="8,16,24,32" for instance?
> It seems that the `PROCESS_ENSEMBLE_ETI` will behave correctly but there will be no `Stokhos_Ensemble_24_Index` nor `HAVE_STOKHOS_ENSEMBLE_SIZE_24` will there be any consequence of that except the fact that Stokhos_config.h will not include information related to the fact Stokhos has an ETI of ensemble of size 24?

The purpose of those macros was not to create one for every enabled ensemble size.  Rather for code that uses specific ensemble sizes, to check whether that ensemble size was enabled.  Currently this is only needed in TrilinosCouplings/FENL, which can only run with specific ensemble sizes (namely 4, 8, 16, and 32).  Ideally one would like FENL to use any ensemble sizes that were enabled, but that would require some serious meta programming that I haven't had the desire to write.

etphipp:
@kliegeois, I guess I could have CMake write out something like an mpl::vector with the list of ensemble sizes.  Do you need this for your code?

kliegeois:
@etphipp  wrote:
> @kliegeois wrote:
> 
> > It looks good.
> > I think that the option -D Stokhos_ETI_Ensemble_Sizes="4,8,16,32" is a good idea.
> > However, I am wondering, how does Stokhos behave if you ask -D Stokhos_ETI_Ensemble_Sizes="8,16,24,32" for instance?
> > It seems that the `PROCESS_ENSEMBLE_ETI` will behave correctly but there will be no `Stokhos_Ensemble_24_Index` nor `HAVE_STOKHOS_ENSEMBLE_SIZE_24` will there be any consequence of that except the fact that Stokhos_config.h will not include information related to the fact Stokhos has an ETI of ensemble of size 24?
> 
> The purpose of those macros was not to create one for every enabled ensemble size. Rather for code that uses specific ensemble sizes, to check whether that ensemble size was enabled. Currently this is only needed in TrilinosCouplings/FENL, which can only run with specific ensemble sizes (namely 4, 8, 16, and 32). Ideally one would like FENL to use any ensemble sizes that were enabled, but that would require some serious meta programming that I haven't had the desire to write.

Thanks for the clarification.

kliegeois:
@etphipp wrote:
> @kliegeois, I guess I could have CMake write out something like an mpl::vector with the list of ensemble sizes. Do you need this for your code?

In my code, I am currently using a hard coded set of ensemble sizes and it is sufficient, as I know for which ensemble sizes I had built Stokhos.
I was just wondering if we could find a way such that Stokhos_config.h includes the values of all enabled ensemble sizes in a current build such that a code that build against Stokhos can find this information and do the ETI for those. 

Therefore, to answer the question, my code does not require it now, but I think that it could be an enhancement. 

What do you think of that?

etphipp:
I could create a CPP macro for each enabled ensemble size, not just the 4,8,16,32.  But it isn't clear to me how you would use this for ETI in a 3rd party code (I suppose the code could try every possible ensemble size, and just check which are enabled, but that seems very inefficient).  I could also write out the ensemble sizes as a string, but I don't know how you would use that either.  I have heard you can do (either horizontal or vertical) iteration with the preprocessor, so in principle you could write some CPP code that would parse this string and iterate over the ensemble sizes, but I don't know how to do it.

kliegeois:
@etphipp I think that the string with all ensemble sizes should be fine. I will look on how to use it in preprocessing. If it is fine for you. We can add it later if my test works.

etphipp:
> @etphipp I think that the string with all ensemble sizes should be fine. I will look on how to use it in preprocessing. If it is fine for you. We can add it later if my test works.

OK.  Once this PR merges (which should be soon) I will add another one that writes the list of ensembles as a string.  I might also try and write out an mpl::vector<> and change FENL to iterate over the enabled ensemble sizes.

mhoemmen:
@etphipp wrote:
> I guess I could have CMake write out something like an mpl::vector with the list of ensemble sizes. Do you need this for your code?

With C++14, we can use `std::integer_sequence`: https://en.cppreference.com/w/cpp/utility/integer_sequence

etphipp:
Actually adding a list of ensemble sizes and iterating through them at compile-time was fairly easy.  See PR #4598

PR #4585, opened 2019-03-08
Amesos2: Fix for superludist when float is enabled

ndellingwood:
Remove float and complex<float> from ETI since suplerludist only
implements double and complex<double>

### DIFF
--- a/packages/amesos2/src/Amesos2_Superludist.cpp
+++ b/packages/amesos2/src/Amesos2_Superludist.cpp
@@ -58,14 +58,8 @@ namespace Amesos2 {
 #endif
 
 #ifdef HAVE_TPETRA_INST_INT_INT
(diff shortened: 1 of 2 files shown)


ndellingwood:
Addresses issue #3566, fix needed for this branch also.
Adding @davydden who first reported the issue.

ndellingwood:
Adding @jwillenbring and @mhoemmen

packages/amesos2/src/Amesos2_Superludist.cpp
```diff
@@ -58,60 +58,36 @@ namespace Amesos2 {
 #endif
 
 #ifdef HAVE_TPETRA_INST_INT_INT
```
mhoemmen: At some point, Amesos2 needs to revise how it does ETI.  Right now, it's super brittle.

Issue #4584, opened 2019-03-08
Typo error in commit #4388 


kliegeois:
There is a typo error in commit #4388 that prevents from compiling packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp if __KOKKOSBATCHED_INTEL_MKL_COMPACT_BATCHED__ is defined. 

Lines 2035-2039:
```
#if defined(__KOKKOSBATCHED_INTEL_MKL_COMPACT_BATCHED__) 
      typedef KokkosBatched::Experimental::Algo::Level3::CompactMKL algo_type;
#else
      typedef KokkosBatched::Experimental::Algo::Level3::Blocked multi_vector_algo_type;
#endif
```
should be replaced by
```
#if defined(__KOKKOSBATCHED_INTEL_MKL_COMPACT_BATCHED__) 
      typedef KokkosBatched::Experimental::Algo::Level3::CompactMKL multi_vector_algo_type;
#else
      typedef KokkosBatched::Experimental::Algo::Level3::Blocked multi_vector_algo_type;
#endif
```

@trilinos/ifpack2 @kyungjoo-kim

kyungjoo-kim:
@kliegeois Sorry. I will submit a PR soon.

kliegeois:
@kyungjoo-kim  No problem, thanks!

kyungjoo-kim:
this should be fixed now. you can close this issue after you confirm that the problem is gone.

kliegeois:
Thanks a lot.
I close the issue.

PR #4388, opened 2019-02-13
Ifpack2 - BlockTriDiags team policy only.

kyungjoo-kim:
## Description

Previously, the code separates RangePolicy for host architecture from TeamPolicy for gpu architectures. This makes code difficult to read and we can achieve a single version of code with proper meta programming. This PR includes the modification for the team policy only version.

## Motivation and Context

While I am working on the code, it is too much headache working on two codes for two architectures. 

## How Has This Been Tested?

Tested from white and bowman.

PR #4587, opened 2019-03-08
Ifpack2 - BlockTridiagContainer improvement and fix typo for intel compact blas 

kyungjoo-kim:
## Description

To improve the line solver performance of SPARC, I made some changes especially for the problem sizes of interest (approximately 4 GB usage of device memory). 
- In general, it uses a larger team size. As the target problem size is smaller, most line kernels cannot fill the whole gpu unit. Increasing team size, it can bring up more concurrency.
- ExtractAndFactorize: now extract routine from tpetra block crs matrix has different loop order to reduce memory transactions. 
- ComputeResidual: as we use Tpetra BlockCrs format, we cannot really expect high degree of interleaved memory access. So, I moved the parallel loop one-level up and coasen the parallelism with atomic add.
- BlockJacobi: block jacobi is used when a line has a unit length. Previously, a numeric phase factorize and a solve phase applies forward/backward solves. Now when a line has a unit length, the numeric phase invert diagonals and the solve phase just apply gemv.  
- KokkosBatched files: I also include batched header files that requires for this update. 
- @kliegeois This also include the fix for the typo of compact mkl.

## Related Issues

#4388 #4584 

## How Has This Been Tested?

Tested on Kokkos-dev-2 and bowman.


packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -55,6 +55,8 @@
 #include <KokkosBatched_Copy_Impl.hpp>
 #include <KokkosBatched_AddRadial_Decl.hpp>
 #include <KokkosBatched_AddRadial_Impl.hpp>
+#include <KokkosBatched_SetIdentity_Decl.hpp>
```
mhoemmen: There's a pattern here of including two header files.  I wonder if it would make sense for kokkos-kernels to generate header files like `KokkosBatched_SetIdentity.hpp`, that include both the Decl and Impl header files.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -74,8 +76,8 @@
 #include <memory>
 
 // need to interface this into cmake variable (or only use this flag when it is necessary)
```
mhoemmen: For the future, it would be good to wrap up the profiling API in an interface that hides how it's implemented.  That would minimize exposure of Ifpack2 code to macros and macrology.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -107,6 +109,12 @@ namespace Ifpack2 {
     template <typename ViewType>
     using ConstUnmanaged = Const<Unmanaged<ViewType> >;
```
mhoemmen: These modifying aliases look like things we could promote to live in kokkos-kernels or even Kokkos at some point.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1432,17 +1448,27 @@ namespace Ifpack2 {
     };
     
 #if defined(KOKKOS_ENABLE_CUDA) 
+    int ExtractAndFactorizeRecommendedCudaTeamSize(const int blksize,
```
mhoemmen: This function is a candidate for `constexpr` with C++14: https://en.cppreference.com/w/cpp/language/constexpr

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1594,37 +1619,56 @@ namespace Ifpack2 {
...
-            const size_type Aj = A_rowptr(lclrow(ri0 + tr)) + A_colindsub(kfs + j);
-            const impl_scalar_type* block = &A_values(Aj*blocksize_square);
-            const size_type pi = kps + j;
-            Kokkos::parallel_for
-              (Kokkos::TeamThreadRange(member,blocksize_square), [&](const local_ordinal_type &ij) {
-                const local_ordinal_type ii = ij/blocksize;
-                const local_ordinal_type jj = ij%blocksize;
-                scalar_values(pi, ii, jj, v) = block[ii*blocksize + jj];
-              });
+              const local_ordinal_type &partidxbeg,
```
mhoemmen: It's better to pass by value than by reference here, since the values have smaller type.  That won't matter if inlining works, but it's good to get into the habit of passing by value where it makes sense.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1594,37 +1619,56 @@ namespace Ifpack2 {
...
         }
       }
 
-      template<typename AAViewType>
+      template<typename AAViewType,
+	       typename WWViewType>
       KOKKOS_INLINE_FUNCTION 
       void 
       factorize(const member_type &member, 
 		const local_ordinal_type &i0,
```
mhoemmen: See above (pass `local_ordinal_type` by value).

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1666,57 +1708,51 @@ namespace Ifpack2 {
                    default_mode_type,KB::Algo::LU::Unblocked>
               ::invoke(member, A, tiny);
           }
-        }
+        } else {
+	  // for block jacobi invert a matrix here
+	  auto W = Kokkos::subview(WW, Kokkos::ALL(), Kokkos::ALL(), v);
+	  KB::Copy<member_type,KB::Trans::NoTranspose,default_mode_type>
```
mhoemmen: Future revisions should consider a "team BLAS" - like free function interface here, that calls `KB::Copy::invoke` etc.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1726,13 +1762,16 @@ namespace Ifpack2 {
         cudaProfilerStart();
 #endif
 	const local_ordinal_type blocksize = internal_vector_values.extent(1);
-	const local_ordinal_type team_size = ExtractAndFactorizeTridiagsDefaultModeAndAlgo<typename execution_space::memory_space>::
+	const local_ordinal_type team_size = 
+	  ExtractAndFactorizeTridiagsDefaultModeAndAlgo<typename execution_space::memory_space>::
```
mhoemmen: It's a bit funny that this class takes a memory space, rather than an execution space.  Could it give different results for `CudaSpace` vs. `CudaUVMSpace`?  If not, then it should take an execution space (e.g., `Cuda`).

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2045,6 +2084,20 @@ namespace Ifpack2 {
     };
     
 #if defined(KOKKOS_ENABLE_CUDA) 
+    int SolveTridiagsRecommendedCudaTeamSize(const int blksize,
```
mhoemmen: See above; this function is a candidate for `constexpr` in C++14.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2142,14 +2192,16 @@ namespace Ifpack2 {
       ///
       /// cuda team vectorization
       ///
+      template<typename WWViewType>
       KOKKOS_INLINE_FUNCTION 
       void 
       solveSingleVector(const member_type &member, 
                         const local_ordinal_type &blocksize,
```
mhoemmen: See above; pass `local_ordinal_type` by value.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2250,17 +2303,34 @@ namespace Ifpack2 {
...
+             X, xs0);
+	}
       }
-
+      
+      template<typename WWViewType>
       KOKKOS_INLINE_FUNCTION 
       void 
       solveMultiVector(const member_type &member, 
                        const local_ordinal_type &blocksize,
```
mhoemmen: See above.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2269,56 +2339,67 @@ namespace Ifpack2 {
...
-          KB::Trsm<member_type,
-                   KB::Side::Left,KB::Uplo::Upper,KB::Trans::NoTranspose,KB::Diag::NonUnit,
-                   default_mode_type,default_algo_type>
-            ::invoke(member, one, A, X2);
-          X1.assign_data( X2.data() );
-        }
+
+	if (nrows > 1) {
+	  // solve Lx = x
+	  KB::Trsm<member_type,
```
mhoemmen: See above (consider free functions).

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2409,6 +2501,18 @@ namespace Ifpack2 {
     ///
     /// compute local residula vector y = b - R x 
     ///
+    int ComputeResidualVectorRecommendedCudaVectorSize(const int blksize,
```
mhoemmen: See above; this is a candidate for `constexpr` in C++14.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2492,7 +2596,7 @@ namespace Ifpack2 {
 
       inline
       void 
-      serialGemv(const local_ordinal_type &blocksize,
+      SerialGemv(const local_ordinal_type &blocksize,
                  const impl_scalar_type * const __restrict__ AA,
```
mhoemmen: `__restrict__` is not a standard C++ keyword.  This may cause build warnings that break warnings-as-errors builds, depending on their settings.  It would be best to use a macro that only invokes the `__restrict__` keyword if you know that the compiler supports it.  Does Kokkos have such a macro?

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2492,7 +2596,7 @@ namespace Ifpack2 {
 
       inline
       void 
-      serialGemv(const local_ordinal_type &blocksize,
+      SerialGemv(const local_ordinal_type &blocksize,
```
mhoemmen: In future revisions, we should factor out these BLAS functions into kokkos-kernels.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: I was ambitious that sometime I can do eti of these files (too much ambitious). I can just create a single header file to include all files. something like e.g. KokkosBatched_FunctorLevelFunctions.hpp

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: I agree.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: We have TeamCopy and SerialCopy. This Copy function has selective mode "Serial/Team". Here if the execution space is host space then I use serial function and if execution space space is cuda, it calls team interface accordingly. This is my way to pursue performance for both KNL and GPU architectures.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: I put template specialization for HostSpace, CudaSpace and CudaUVMSpace. I don't want to put specialization based on the execution space of OpenMP and Serial specialization.  One reason is that sometimes OpenMP or Serial space is not enabled and I need to put another macro to wrap the specialization. HostSpace is always defined. When Cuda is enabled, I just take care of CudaSpace and CudaUVMSpace only.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: Yes sometime but not right now. I put this another implementation separately from kokkoskernels because of different performance characteristics. This functions are used to block sparse matrix vector multiplication where input matrix is packed and output matrix is tpetra multivector format. I tested a few variations and what I found is to use atomic add and parallelize over columns in a row. Also, for performance reason, here we specialize blocksize with a template parameter and I give such information explicitly. With kokkos interface, it is ideal to use view interface. Then, templatized blocksize porbably won't be effective.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: I think that they have one. However, I use this keyword in kokkoskernels and often find out it helps vectorization. At this moment, it does not generate warnings from there.

srajama1:
I need to do review this one carefully, let me do it when I get a block of time. That said I agree with the comment above that Ifpack2 shouldn't have BLAS calls. It is better to move it to Kokkos Kernels.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: @kyungjoo-kim We could just use the keyword through a macro, defined as `__restrict__` if the compiler supports that, and defined as nothing otherwise.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: @kyungjoo-kim I see, these are specialized for the block sparse mat-vec use case.  It might be better to move them into a separate file so they can be tested separately, but it sounds like they wouldn't be more generally useful outside of the block sparse mat-vec case.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: done.

PR #4592, opened 2019-03-09
Fix CUDA PR build to enable SEACAS, STK, and 103 extra Panzer tests (#2464)

bartlettroscoe:
This fixes the Trilinos CUDA PR build to match the build `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-release-debug-pt` is was copied from (see #2464).  This now enables SEACAS and STK and all of their tests (except the ones we know already fail).  It also increases the number Panzer tests run from 60 to 163.

For details see the commits for details.  I will add a comment to #2464 with details of the diffing process and for the local testing that was done.

NOTE: After this PR is merged, we can work on enabling other TPLs like METIS, ParMETIS, SuperLUDist and others.  We will need to do this to protect SPARC's usage since it enables these extra TPLs.  But at least now we have a solid foundation for that.


bartlettroscoe:
@mhoemmen, thanks for the approval!  I had to disable a few tests due to the PR testing running with a too-high of a ctest parallel level so it discarded your review.  Can you please review again?

Issue #4593, opened 2019-03-10
PyTrilinos: Implement new infrastructure for nested sub-packages


wfspotz:
@trilinos/pytrilinos 

## Expectations
All packages that have nested sub-packages should use the same infrastructure, and we need to avoid swig-generated base classes that include the `__init__` "module", because this is not valid python syntax.  Also, there are import issues that are different for Python 2 and Python 3, and the resulting solution should work for both.

## Current Behavior
Import issues keep cropping up, and some "solutions" create base classes with `__init__` in the module path.

## Possible Solution
* Use the relatively new `moduleimport` option for the `%module` SWIG directive to better control importing required modules.
* Change `__init__.py` modules from swig-generated, which includes a compiled extension, to pure python.  Use `__init__.py.in` files to generate these files in

    `packages/PyTrilinos/src/PyTrilinos`

* Change `Package.__init__.i` to `Package.i` and have it generate a python file `Package/Base.py` which is imported by the `__init__.py` file from the previous bullet.  Use `globals().update(Base.__dict__)` to create references from everything in `Base` to the `__init__.py` file.  Certain special variables, such as `__file__` will need to be preserved.

wfspotz:
PR #4594 implemented the nested package refactor, but I have discovered a bug: when PyTrilinos is installed, the `__init__.py` files are installed into the wrong place.  This needs to be fixed for the Teuchos, Isorropia, NOX, and LOCA python wrappers for PyTrilinos.

wfspotz:
Checklist:
- [ ] Teuchos
- [ ] Isorropia
- [ ] NOX
- [ ] NOX.Epetra
- [ ] NOX.PETSc
- [ ] LOCA
- [ ] LOCA.Epetra
- [ ] LOCA.Hopf
- [ ] LOCA.Pitchfork
- [ ] LOCA.TurningPoint

PR #4594, opened 2019-03-10
PyTrilinos: Refactor infrastructure of nested packages

wfspotz:
@trilinos/pytrilinos 

## Description
Refactor the infrastructure of nested packages to improve behavior between Python 2 and Python 3, and simplify by using the new-ish `moduleimport` option of the `%module` SWIG directive.

## Motivation and Context
Various reports have popped up about import errors, and it was getting harder to support both Python 2 and 3.

## Related Issues

* Related to #4593 

## How Has This Been Tested?
I have tested a fully-configured PyTrilinos built using both Python 2 and Python 3.

## Checklist


packages/PyTrilinos/src/Teuchos_Comm.i
```diff
@@ -610,40 +610,17 @@ else:
...
-////////////////////////////////////////////////////////////////////////////////
+#endif
 
-/////////////////////////////////////////////
-// Teuchos.DefaultComm support without MPI //
-/////////////////////////////////////////////
+/////////////////////////////////
+// Teuchos.DefaultComm support //
+/////////////////////////////////
+#define MPI_VERSION 2
```
mhoemmen: Are you sure you want to do this?  Trilinos only tests now with MPI implementations that have `MPI_VERSION` at least 3.

wfspotz: This was from a while ago... I think it was a problem where SWIG needed it to be defined, and extracting just `MPI_VERSION` from `mpi.h` was too much of a hassle.  I can probably just change it to `3` if that is what Trilinos uses.

mhoemmen: @wfspotz It is definitely 3 is all of our tests.  Not sure if this matters much for PyTrilinos, though, so I'll just approve this PR for now.

wfspotz: Nice catch.  This was from a while ago ... SWIG needs `MPI_VERSION` to be defined, but having it parse all of `mpi.h` to get it is not really practical.  It should just be changed to 3.

I discovered a number of cosmetic things in PyTrilinos that needed to be updated as well: the Developers Guide has changed because of the new infrastructure, URLs in docstrings needed to be changed, an unused script from the old infrastructure needed to be deleted, etc.  So my update will include `#define MPI_VERSION 3` as well as a bunch of other documentation-related things.

wfspotz:
I'm not sure why no packages were enable by the autotester....

PR #4596, opened 2019-03-11
Update stk with recent development.

alanw0:
1. Fix Nvidia/Cuda build warnings in stk-topology.
2. Fix failing stk_ngp unit test.
3. Add stk-balance cmake files.

@trilinos/<teamName>

## Checklist


mhoemmen:
@alanw0 The build errors suggest that you are using `GlobalOrdinal=long`.  This is not necessarily enabled in all of Trilinos' tests (and shouldn't always be enabled).  Do you test the appropriate CMake options and macros before using `GlobalOrdinal=long`?

alanw0:
@mhoemmen I didn't knowingly specify GlobalOrdinal, or anything else. Doesn't the PR-testing system specify all the builds? Or are you suggesting there is some weird GlobalOrdinal usage in the stk_balance source code? In that case, I don't know. It's only been tested in sierra and using nalu-wind's build, and it worked with that...

mhoemmen:
@alanw0 wrote:
> Or are you suggesting there is some weird GlobalOrdinal usage in the stk_balance source code?

There are a lot of files to read, but I'm guessing you instantiate Zoltan2 objects at some point, with some GlobalOrdinal type.

> It's only been tested in sierra and using nalu-wind's build, and it worked with that...

Both products enable `GlobalOrdinal=long`.

mhoemmen:
(@maherou @kddevin Here is another example of why we need to get rid of the GlobalOrdinal template parameter)

alanw0:
@mhoemmen yes stk_balance certainly does instantiate zoltan2 objects. What are the "always-available" types we can choose from for those instantiations?

mhoemmen:
@alanw0 wrote:
> What are the "always-available" types we can choose from for those instantiations?

`Tpetra::Map<>::global_ordinal_type` is the default GlobalOrdinal type.  That always works.  If you need a 64-bit GlobalOrdinal type for some tests, you can test the following CMake options:

  - Tpetra_INST_INT_LONG_LONG (preferred, since `long long` always has >= 64 bits, while `long` only has 32 bits on some platforms)
  - Tpetra_INST_INT_LONG

and their corresponding macros:

  - `HAVE_TPETRA_INST_INT_LONG_LONG`
  - `HAVE_TPETRA_INST_INT_LONG`

PR #4598, opened 2019-03-11
Stokhos: Add compile-time list of enable ensemble sizes.

etphipp:
This adds a compile-time list of enabled ETI ensembles sizes (using
Sacado::mpl::vector), which is used by TrilinosCouplings/FENL to run
with any enabled ensemble size.  This replaces the hard-code values that
were used before.  Also moved all ensemble-size related values to a new
header.  Also write out enabled ensemble sizes as a string for those
that might want it.

Changes have been made in the past that broke these codes, but were
never caught because they didn't run any tests.  Now fenl_ensemble and
fenl_pce can each run a simple test with tribits as well.


packages/stokhos/CMakeLists.txt
```diff
@@ -72,22 +72,19 @@ IF(Stokhos_ENABLE_Ensemble_Scalar_Type)
...
-    SET(HAVE_STOKHOS_ENSEMBLE_SIZE_16 1)
-  ENDIF()
-  IF(NOT (Stokhos_Ensemble_32_Index EQUAL -1))
-    SET(HAVE_STOKHOS_ENSEMBLE_SIZE_32 1)
-  ENDIF()
+
+  # Make list of enabled ensemble sizes suitable for iteration/searching
+  SET(STOKHOS_MPL_ENSEMBLE_SIZES "typedef Sacado::mpl::vector< ")
+  FOREACH(ENSEMBLE_SIZE ${Stokhos_ETI_Ensemble_Sizes})
+    STRING(APPEND STOKHOS_MPL_ENSEMBLE_SIZES "Sacado::mpl::integral_c<unsigned,${ENSEMBLE_SIZE}>, ")
```
mhoemmen: Once we have C++14, we can make this an alias to `std::integral_constant`.  Yay for deleting code! :-D

PR #4601, opened 2019-03-11
Teuchos: Fix #4600

### LINKED ISSUE #4600: Teuchos: Some CMake checks happen at run time unconditionally; this hinders cross compilation
@trilinos/teuchos @trilinos/framework @theguruat12 @vbrunini @rrdrake 

@theguruat12 and I found that `teuchos/CMakeLists.txt` does some checks that involve running code.  That's normally OK, except that it's not OK for cross compilation.  For example, when building KNL-specific binaries on a login node that's not KNL, a lot of tests (like `HAVE_TEUCHOS_BLASFLOAT`) will incorrectly report back false.  This will break the build if `Scalar=float` is enabled, which @vbrunini needs.

@theguruat12 and I have prototyped a fix that seems to work.  It follows the preexisting Trilinos idiom of letting users set CMake options in advance, if they want to avoid CMake tests that run code.  I will submit a PR soon that implements this fix.

mhoemmen:
@trilinos/teuchos @trilinos/framework @theguruat12 @rrdrake 

Make it possible to cross-compile Teuchos and downstream packages with `Scalar=float` enabled.

## Related Issues

* Closes #4600

### DIFF
--- a/packages/teuchos/CMakeLists.txt
+++ b/packages/teuchos/CMakeLists.txt
@@ -14,7 +14,7 @@ IF (TRIBITS_PROCESSING_PACKAGE)
 #
 
 #
-# A) Forward delcare the package so that certain options are also defined for
+# A) Forward declare the package so that certain options are also defined for
 # subpackages
 #
 
@@ -30,8 +30,16 @@ TRIBITS_ADD_DEBUG_OPTION()
 
 TRIBITS_ADD_SHOW_DEPRECATED_WARNINGS_OPTION()
 
-INCLUDE(CheckForAbiDemangle)
-CHECK_FOR_ABI_DEMANGLE(HAVE_GCC_ABI_DEMANGLE)
+#
+# The check needs to run code, so give cross-compiling users a bypass.
+#
+IF (DEFINED HAVE_GCC_ABI_DEMANGLE)
+  MESSAGE(STATUS "You set HAVE_GCC_ABI_DEMANGLE=${HAVE_GCC_ABI_DEMANGLE}; this suggests you are cross compiling.  If you're not cross compiling, best practice is not to set this; Trilinos will figure it out for you.")
+ELSE()
+  INCLUDE(CheckForAbiDemangle)
+  CHECK_FOR_ABI_DEMANGLE(HAVE_GCC_ABI_DEMANGLE)
+ENDIF()
+ASSERT_DEFINED(HAVE_GCC_ABI_DEMANGLE)
 
 ASSERT_DEFINED(TPL_ENABLE_BinUtils CMAKE_CXX_COMPILER_ID)
 IF(TPL_ENABLE_BinUtils)
@@ -48,7 +56,7 @@ IF(TPL_ENABLE_BinUtils)
     "Enable Teuchos to show stack track for GCC (or Intel) on segfaults and uncaught exceptions."
     ${${PACKAGE_NAME}_ENABLE_STACKTRACE_DEFAULT}
     )
-  
+
   IF (${PACKAGE_NAME}_ENABLE_STACKTRACE)
     SET(HAVE_TEUCHOS_LINK ON)
     SET(HAVE_TEUCHOS_BFD ON)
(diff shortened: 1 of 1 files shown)


packages/teuchos/CMakeLists.txt
```diff
@@ -226,19 +248,37 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
...
-    ELSE()
-      CHECK_CXX_COMPLEX_BLAS_PROBLEM_CAN_BE_FIXED(HAVE_FIXABLE_COMPLEX_BLAS_PROBLEM)
+    IF ((NOT (DEFINED HAVE_VECLIB_COMPLEX_BLAS)) AND
+        (NOT (DEFINED HAVE_FIXABLE_COMPLEX_BLAS_PROBLEM)))
+      MESSAGE(FATAL_ERROR "IF HAVE_COMPLEX_BLAS_PROBLEM is defined and TRUE, then you must define at least one of HAVE_VECLIB_COMPLEX_BLAS or HAVE_FIXABLE_COMPLEX_BLAS_PROBLEM.  This is most likely to happen if you are cross compiling and setting these options yourself, rather than relying on CMake to set the options.")
     ENDIF()
   ENDIF()
-  GLOBAL_SET(HAVE_COMPLEX_BLAS 
+
+  GLOBAL_SET(HAVE_COMPLEX_BLAS
```
bartlettroscoe: I am not familiar with this type of CMake logic.  Does this work?  Can you assign a CMake variable to a boolen expression?

mhoemmen:
@trilinos/framework @trilinos/rol the dreaded "rol failure" strikes again

packages/teuchos/CMakeLists.txt
```diff
@@ -79,44 +87,58 @@ ENDIF()
...
+  # The check needs to run code, so give cross-compiling users a bypass.
+  #
+  IF (DEFINED HAVE_TEUCHOS_BLASFLOAT)
+    MESSAGE(STATUS "You set HAVE_TEUCHOS_BLASFLOAT=${HAVE_TEUCHOS_BLASFLOAT}; this suggests you are cross compiling.  If you're not cross compiling, best practice is not to set this; Trilinos will figure it out for you.")
+  ELSE()
+    INCLUDE(CheckBlasFloat)
+    CHECK_BLAS_FLOAT(HAVE_TEUCHOS_BLASFLOAT)
     IF(NOT HAVE_TEUCHOS_BLASFLOAT)
-      INCLUDE(CheckBlasFloatReturnsDouble)
-      CHECK_BLAS_FLOAT_DOUBLE_RETURN(HAVE_TEUCHOS_BLASFLOAT_DOUBLE_RETURN)
```
jhux2: Following up on @bartlettroscoe's comment, I can't find this in the cmake [documentation](https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/VariablesListsStrings).

packages/teuchos/CMakeLists.txt
```diff
@@ -30,8 +30,16 @@ TRIBITS_ADD_DEBUG_OPTION()
 
 TRIBITS_ADD_SHOW_DEPRECATED_WARNINGS_OPTION()
 
-INCLUDE(CheckForAbiDemangle)
-CHECK_FOR_ABI_DEMANGLE(HAVE_GCC_ABI_DEMANGLE)
+#
+# The check needs to run code, so give cross-compiling users a bypass.
+#
+IF (DEFINED HAVE_GCC_ABI_DEMANGLE)
```
jhux2: @mhoemmen Would this be enabled by a cmake option?

packages/teuchos/CMakeLists.txt
rppawlo: cross compiling with cmake is typically handled with toolchain files. Not sure if that works in this case though. 

https://cmake.org/cmake/help/v3.6/manual/cmake-toolchains.7.html

bartlettroscoe: @jhux2, the function `CHECK_BLAS_FLOAT_DOUBLE_RETURN()` is our own custom function but it calls the standard CMake function `CHECK_CXX_SOURCE_RUNS()` which works with cache vars as described above.   That is how this is supposed to work.  You define your own configure-time checks that calls `CHECK_CXX_SOURCE_RUNS()` and then you can provide the value of the cmake cache var (shown as `HAVE_TEUCHOS_BLASFLOAT_DOUBLE_RETURN` in this case).  Make sense?

mhoemmen: @jhux2 Those custom checks live in `Trilinos/packages/teuchos/cmake`.

packages/teuchos/CMakeLists.txt
mhoemmen: @jhux2 Yes.  The idea is that if you want to bypass the check, you just set the corresponding CMake option (in this case, `HAVE_GCC_ABI_DEMANGLE`) to the value that you want.  I learned this approach from @etphipp , back when he had to cross-compile for what I think was KNC.

packages/teuchos/CMakeLists.txt
mhoemmen: @bartlettroscoe Not sure; that's what was there before.  I'm pretty sure that setting these to FALSE works, since that's what was happening in the cross-compilation case.

mhoemmen:
@bartlettroscoe wrote:
> Therefore, I think you can remove all of this extra logic and have it still work the way you want (in a more CMakey way).

Ah, that's nicer; thanks!  I still think it would be helpful to print the status messages, though.  Typical user / developer dialog:

  - User: Trilinos is broken!
  - Developer: May I see your CMake options?
  - User: *sends enormous list of options*
  - Developer: Oh, you're setting `${OPTION}`.  CMake's output literally tells you not to do that unless `${REASON}`.

What do you think?  Is there a way we could get the status messages (e.g., "You are skipping `${CHECK}` by setting `${OPTION}`.  Usually you only want to do that if you are cross compiling.") without the redundant checks?

bartlettroscoe:
@mhoemmen asked:

> What do you think? 

Giving the user nice error messages is good but it takes more cmake code to maintain this and you need to manually test the various use cases manually to make sure the work.  My opinion is that developers and users need to stop setting so many Trilinos configure options unless they really know what they are doing.  I think we just give them some general guidance on the types of vars they need to set when cross compiling.  It would be really nice if CMake could spit out a template file with all of the vars that get set with try compile commands so that you know which you need to set.  I think having one or two up-to-date examples of toolchain files for Trilinos should do the trick.

sebrowne:
I can try the build again tomorrow and see if the behavior expected by @bartlettroscoe is indeed what I see, I thought I had already tried that but will confirm.

packages/teuchos/CMakeLists.txt
jjellio: The problem with `SOURCE_RUNS` is that it won't work with cross compiling.  If this can be coerced into a `try_compile` test, then you skirt the problem of compiling for a different architecture than the build system is running on.

mhoemmen:
@bartlettroscoe wrote:
> My opinion is that developers and users need to stop setting so many Trilinos configure options unless they really know what they are doing.

AMEN

> It would be really nice if CMake could spit out a template file with all of the vars that get set with try compile commands so that you know which you need to set.

That would be super awesome.

ndellingwood:
The failing test was due to a the failing Ifpack2 test fixed in PR #4608 (merged into develop), so I'll add a retest to this PR and it should pass.

mhoemmen:
@rrdrake @theguruat12

PR #4606, opened 2019-03-12
Belos Stratimikos adapter: Use solver factory

cgcgcg:
@trilinos/belos @trilinos/stratimikos 

## Description
Use the Belos solver factory in the Stratimikos adapter.

@mhoemmen Could you check if this is correct? I tried matching the pattern for the Tpetra solver manager, but I am not sure if I got the whole DII stuff right..

### DIFF
--- a/packages/stratimikos/adapters/belos/src/Belos_Details_Thyra_registerSolverFactory.cpp
+++ b/packages/stratimikos/adapters/belos/src/Belos_Details_Thyra_registerSolverFactory.cpp
@@ -1,0 +1,136 @@
+//@HEADER
+// ************************************************************************
+//
+//                 Belos: Block Linear Solvers Package
+//                  Copyright 2004 Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the Corporation nor the names of the
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL SANDIA CORPORATION OR THE
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact Michael A. Heroux (maherou@sandia.gov)
+//
+// ************************************************************************
+//@HEADER
+
+
+/// \file Belos_Details_Thyra_registerSolverFactory
+/// \brief Implement Injection and Inversion (DII) for Thyra
+
+#include "Thyra_BelosSolverFactory.hpp"
+
+#include "BelosBiCGStabSolMgr.hpp"
+#include "BelosBlockCGSolMgr.hpp"
+#include "BelosBlockGmresSolMgr.hpp"
+#include "BelosFixedPointSolMgr.hpp"
+#include "BelosGCRODRSolMgr.hpp"
+#include "BelosGmresPolySolMgr.hpp"
+#include "BelosLSQRSolMgr.hpp"
+#include "BelosMinresSolMgr.hpp"
+#include "BelosPCPGSolMgr.hpp"
+#include "BelosPseudoBlockCGSolMgr.hpp"
+#include "BelosPseudoBlockGmresSolMgr.hpp"
+#include "BelosPseudoBlockTFQMRSolMgr.hpp"
+#include "BelosRCGSolMgr.hpp"
+#include "BelosTFQMRSolMgr.hpp"
+
+#include "Teuchos_ExplicitInstantiationHelpers.hpp"
+
+namespace Belos {
+namespace Details {
+namespace Thyra {
+
+void registerSolverFactory() {
+
+  #define BELOS_LCL_CALL_FOR_MANAGER(manager,name,SC)   \
+    Impl::registerSolverSubclassForTypes<                           \
+      manager<SC,::Thyra::MultiVectorBase<SC>,             \
+              ::Thyra::LinearOpBase<SC> >, SC,         \
+              ::Thyra::MultiVectorBase<SC>, \
+      ::Thyra::LinearOpBase<SC> > (name);
+
+  #define LCL_CALL( SC ) BELOS_LCL_CALL_FOR_MANAGER(BiCGStabSolMgr, "BICGSTAB", SC)
+  TEUCHOS_MACRO_TEMPLATE_INSTANT_REAL_SCALAR_TYPES(LCL_CALL);
+
+  #undef LCL_CALL
+  #define LCL_CALL( SC ) BELOS_LCL_CALL_FOR_MANAGER(BlockCGSolMgr, "BLOCK CG", SC)
+  TEUCHOS_MACRO_TEMPLATE_INSTANT_REAL_SCALAR_TYPES(LCL_CALL);
+
+  #undef LCL_CALL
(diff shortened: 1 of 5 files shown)


packages/stratimikos/adapters/belos/src/Belos_Details_Thyra_registerSolverFactory.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+  #define BELOS_LCL_CALL_FOR_MANAGER(manager,name,SC)   \
+    Impl::registerSolverSubclassForTypes<                           \
+      manager<SC,::Thyra::MultiVectorBase<SC>,             \
+              ::Thyra::LinearOpBase<SC> >, SC,         \
+              ::Thyra::MultiVectorBase<SC>, \
+      ::Thyra::LinearOpBase<SC> > (name);
+
+  #define LCL_CALL( SC ) BELOS_LCL_CALL_FOR_MANAGER(BiCGStabSolMgr, "BICGSTAB", SC)
+  LCL_CALL(double);
```
mhoemmen: Do we only register for `double` automatically, or should we also do so for all the Scalar types that Thyra enables at configure time?  I'm not sure which -- if I remember right, I think Stratimikos only registers things for `Scalar=double`.  If so, then this is fine.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -77,28 +68,10 @@ namespace Thyra {
 template<class Scalar>
 const std::string BelosLinearOpWithSolveFactory<Scalar>::SolverType_name = "Solver Type";
 template<class Scalar>
-const std::string BelosLinearOpWithSolveFactory<Scalar>::SolverType_default = "Pseudo Block GMRES";
+const std::string BelosLinearOpWithSolveFactory<Scalar>::SolverType_default = "PSEUDO BLOCK GMRES";
```
mhoemmen: Does this actually matter?  I don't mind the change, but I think `Belos::SolverFactory` capitalizes and otherwise canonical-izes the input name.  Also, this is not consistent with a default set below.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -110,8 +83,8 @@ const std::string LeftPreconditionerIfUnspecified_name = "Left Preconditioner If
 
 template<class Scalar>
 BelosLinearOpWithSolveFactory<Scalar>::BelosLinearOpWithSolveFactory()
-  :solverType_(SOLVER_TYPE_PSEUDO_BLOCK_GMRES),
-   convergenceTestFrequency_(1)
+  : solverName_("Pseudo Block GMRES"),
```
mhoemmen: See above note on `SolverType_default`.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -121,7 +94,7 @@ template<class Scalar>
 BelosLinearOpWithSolveFactory<Scalar>::BelosLinearOpWithSolveFactory(
   const RCP<PreconditionerFactoryBase<Scalar> > &precFactory
   )
-  :solverType_(SOLVER_TYPE_PSEUDO_BLOCK_GMRES)
+  : solverName_("Pseudo Block GMRES")
```
mhoemmen: See above note on `SolverType_default`.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -299,10 +272,20 @@ void BelosLinearOpWithSolveFactory<Scalar>::setParameterList(
   )
 {
   TEUCHOS_TEST_FOR_EXCEPT(paramList.get()==NULL);
-  paramList->validateParametersAndSetDefaults(*this->getValidParameters(), 1);
+  std::string solverName = paramList->get<std::string>(SolverType_name, SolverType_default);
+  std::string solverNameUC = Belos::Impl::upperCase (solverName);
```
mhoemmen: It looks like you need Belos' canonical solver name here.  Belos does more than just capitalize the input name; it also has a mapping from "alias(es)" to a "canonical name."  If you want access to that here, let me know.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -367,87 +350,12 @@ BelosLinearOpWithSolveFactory<Scalar>::generateAndGetValidParameters()
...
-        "Block CG",
-        "Pseudo Block CG",
-        "Pseudo Block Stochastic CG",
-        "GCRODR",
-        "RCG",
-        "MINRES",
-        "TFQMR"
-        ),
-      tuple<std::string>(
-        "Block GMRES solver for nonsymmetric linear systems.  It can also solve "
```
mhoemmen: I don't think these documentation strings exist anywhere else.  Do we want to save the strings somewhere, say in `Belos::SolverFactory` itself?

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -459,60 +367,21 @@ Teuchos::ValidatorXMLConverterDB::addConverter(
...
+
+    Belos::ThyraSolverFactory<Scalar> factory;
+    Teuchos::Array<std::string> supportedSolvers = factory.supportedSolverNames();
+    typedef Belos::SolverManager<Scalar,MV_t,LO_t> IterativeSolver_t;
+    for (size_t i = 0; i < supportedSolvers.size(); i++) {
+      try {
+        RCP<IterativeSolver_t> iterativeSolver = factory.create ( supportedSolvers[i], Teuchos::null );
+        solverTypesSL.sublist(supportedSolvers[i]) = *(iterativeSolver->getValidParameters());
+      } catch (std::invalid_argument) {
+        // pass
```
mhoemmen: Hm, this is using try-catch as control flow.  I guess there's no way around it, though.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -691,193 +560,23 @@ void BelosLinearOpWithSolveFactory<Scalar>::initializeOpImpl(
...
-      }
-      break;
-    }
-
-    default:
-    {
-      TEUCHOS_TEST_FOR_EXCEPT(true);
-    }
+  Teuchos::ParameterList &solverTypesPL = paramList_->sublist(SolverTypes_name);
+  Teuchos::ParameterList &gmresPL = solverTypesPL.sublist(solverName_);
```
mhoemmen: Why is this called `gmresPL`?  I don't think it's specific to GMRES.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -691,193 +560,23 @@ void BelosLinearOpWithSolveFactory<Scalar>::initializeOpImpl(
   // Generate the parameter list.
   //
 
+  Belos::ThyraSolverFactory<Scalar> factory;
   typedef Belos::SolverManager<Scalar,MV_t,LO_t> IterativeSolver_t;
   RCP<IterativeSolver_t> iterativeSolver = Teuchos::null;
   RCP<Teuchos::ParameterList> solverPL = Teuchos::rcp( new Teuchos::ParameterList() );
```
mhoemmen: `auto solverPL = Teuchos::parameterList ();`

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
```diff
@@ -691,193 +560,23 @@ void BelosLinearOpWithSolveFactory<Scalar>::initializeOpImpl(
...
-      break;
-    }
-
-    default:
-    {
-      TEUCHOS_TEST_FOR_EXCEPT(true);
-    }
+  Teuchos::ParameterList &solverTypesPL = paramList_->sublist(SolverTypes_name);
+  Teuchos::ParameterList &gmresPL = solverTypesPL.sublist(solverName_);
+  solverPL = Teuchos::rcp( &gmresPL, false );
```
mhoemmen: Why not just `solverPL = gmresPL`?  Why the nonowning RCP?

packages/stratimikos/adapters/belos/src/Belos_Details_Thyra_registerSolverFactory.cpp
cgcgcg: I thought it was only `double`, but it's in fact all real-valued scalar types. Fixed.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
cgcgcg: This probably does not matter. Fixed the issue with different defaults given below.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
cgcgcg: This should be ok, since I am actually getting valid lists for all available solver names, even if they are just aliases.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
cgcgcg: Fixed.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
cgcgcg: Fixed.

packages/stratimikos/adapters/belos/src/Thyra_BelosLinearOpWithSolveFactory_def.hpp
cgcgcg: I could move them to `Belos_Details_EBelosSolverType.cpp`..

cgcgcg:
@mhoemmen Thank you very much for reviewing!

Issue #3570, opened 2018-10-06
Amesos2: undefined symbols for complex numbers


davydden:
## Expectations
Trilinos should build.

## Current Behavior
linker errors:
```
Undefined symbols for architecture x86_64:
  "Amesos2::KLU2<Tpetra::CrsMatrix<std::__1::complex<double>, int, unsigned long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>, false>, Tpetra::MultiVector<std::__1::complex<double>, int, unsigned long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>, false> >::KLU2(Teuchos::RCP<Tpetra::CrsMatrix<std::__1::complex<double>, int, unsigned long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>, false> const>, Teuchos::RCP<Tpetra::MultiVector<std::__1::complex<double>, int, unsigned long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>, false> >, Teuchos::RCP<Tpetra::MultiVector<std::__1::complex<double>, int, unsigned long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>, false> const>)", referenced from:
[...]

PR #4609, opened 2019-03-12
Amesos2 linsolfact

ndellingwood:
@trilinos/amesos2 

Addresses issue #3570
Updates LinearSolverFactory to include umfpack as a solver; ETI fixes for missed instantiations by using Tpetra macro, thanks @mhoemmen for assistance.

PR #4608, opened 2019-03-12
Ifpack2 - for the testing with kokkos bound check

kyungjoo-kim:
## Description
Ifpack2 blodk tridiag solver fails when kokkos bound check is enabled. I put an additional index guard in extract routine. 

## Related Issues
#4597 

## How Has This Been Tested?
Tested on white P100.

@jjellio @jhux2

PR #4610, opened 2019-03-12
Amesos2: Fix ETI issues and non-contiguous GID single rank issue

ndellingwood:
Addresses issue #3570, thanks @mhoemmen for assistance.
Addresses issue #3647.


ndellingwood:
PR mirrors fixes to Trilinos develop branch in #4609 and past PR with fixes for issue #3647.

packages/amesos2/src/Amesos2_Umfpack.cpp
```diff
@@ -58,332 +57,14 @@ namespace Amesos2 {
...
-#include "Kokkos_DefaultNode.hpp"
-#include "TpetraCore_ETIHelperMacros.h"
-
-#define AMESOS2_UMFPACK_LOCAL_INSTANT(S,LO,GO,N)                        \
-  template class Amesos2::Umfpack<Tpetra::CrsMatrix<S, LO, GO, N>,      \
-                                  Tpetra::MultiVector<S, LO, GO,  N> >;
+} //namespace Amesos2
 
+#define AMESOS2_UMFPACK_LOCAL_INSTANT(S,LO,GO,N) \
+  template class Amesos2::Umfpack<Tpetra::CrsMatrix<S, LO, GO, N>, \
```
mhoemmen: Note that in a complex arithmetic build, this requires that UMFPACK was built with complex arithmetic enabled.  @theguruat12 and @rrdrake know about this.

ndellingwood: @mhoemmen there is a TPL check to make sure the UMFPACK version supports complex types [see here](https://github.com/trilinos/Trilinos/blob/adbf724c2e749af6103daac1b041e578c3ce2525/cmake/TPLs/FindTPLUMFPACK.cmake#L57-L93), does this address your concern or do we need to add warning messages into Amesos2 as well?

ndellingwood:
Adding WIP label until #4614 is addressed which will also allow #4608 to complete and merge.

packages/amesos2/src/Amesos2_Umfpack.cpp
mhoemmen: @ndellingwood I think that's OK -- the issue didn't come up before for Sierra because now they are switching to build Trilinos with CMake.  They should encounter this test now.  Thanks!

Issue #4614, opened 2019-03-13
PR test timeouts preventing PRs from passing


ndellingwood:
@trilinos/framework

The following test timeouts have resulted in several PRs failing and requiring re-test:

```
PanzerAdaptersSTK_PoissonInterfaceExample_2d_diffsideids_MPI_1
ROL_example_PDE-OPT_0ld_adv-diff-react_example_02_MPI_4
ROL_example_PDE-OPT_navier-stokes_example_01_MPI_4
```

Of immediate consequence is that this is preventing PR #4608 from completing and merging in, which will result in any Cuda PR testing that involves Ifpack2 from passing. 

The tests timing out above, along with #4608 in some cases which is also unable to complete PR testing due to these tests timing out is preventing PR completion of PRs #4609, #4603, #4604, #4601, #4597, ...

rppawlo:
Did something change in the configure for these builds? The last commit to panzer was almost 4 weeks ago.

ndellingwood:
@rppawlo I believe in one of the builds kokkos debugging was recently enabled, which exposed the error in Ifpack2 that #4608 will fix once it makes it through without a timeout. Kokkos debugging will increase the length of runtime, which I'm guessing is impacting the Panzer test. I'm not certain how long the ROL tests have been timing out and if they're correlated with the kokkos debug option or not.

jhux2:
@bartlettroscoe

bartlettroscoe:
The @trilinos/framework team just needs to turn down the parallel test level from 29 to 8 as per https://github.com/trilinos/Trilinos/issues/2464#issuecomment-471221881.  I told @jwillenbring about this over the phone on Monday.

kyungjoo-kim:
@bartlettroscoe Your commit to reduce parallel test level from 29 to 8 is merged about 3 days ago and we still have failed tests owing to timeout.

bartlettroscoe:
@kyungjoo-kim 

> @bartlettroscoe Your commit to reduce parallel test level from 29 to 8 is merged about 3 days ago and we still have failed tests owing to timeout.

What commit?  It is impossible for me to reduce the parallel test level for 29 to 8.  Only the @trilinos/framework team can do that.

kyungjoo-kim:
@bartlettroscoe while I am editing the previous comment, you responded. I mean 
[...]

PR #4615, opened 2019-03-13
Disable tests timing out in PR autotester

ndellingwood:
Temporarily address #4614 to get PR tests passing.

Not adding AUTOMERGE label, will let @trilinos/framework make the call or close this PR with their preferred solution.


kyungjoo-kim:
@ndellingwood You also probably need to add Ifpack2 BlockTridiag test as it is a failure due to debugging check.

jwillenbring:
It might be worth holding off on merging this to see if lowering the parallel level to 9 resolves the issue. That change was merged late this morning.

ndellingwood:
@jwillenbring thanks! I figured I'd get something in so testing could start just in case it's needed to get the PRs moving again, hopefully the parallel level change works so this PR is unnecessary.

ndellingwood:
Talked to @kyungjoo-kim , marking this WIP until we see if PR #4608 passes with the changed parallel level in testing; if that PR still fails due to the timeouts, will update this PR to disable the failing Ifpack2 test temporarily to break the PR testing deadlock (otherwise this PR will fail as well). @jhux2

ndellingwood:
Closing this PR, unnecessary after change of parallel level for autotesting.

PR #4618, opened 2019-03-14
ThyraTpetraAdapters: Added getter for Tpetra::Map in Thyra::TpetraVectorSpace

### LINKED ISSUE #4616: Thyra_TpetraVectorSpace to expose getter for underlying Tpetra object
@trilinos/tpetra
@trilinos/thyra

## Expectations

I want to be able to extract the underlying Tpetra map from a vector space

## Current Behavior

You can't.

## Motivation and Context

The Tpetra adapter offer the possibility to extract the Tpetra object for Vector, MultiVector, and LinearOp. However, such capability does not exist for VectorSpace. This is unfortunate. For (Multi)Vector, one can survive without this capability (at least on host), since there are methods in parent classes that allow extraction of local data or subviews. However, there is no Thyra equivalent of Global ID and Local ID, so one _needs_ to get the underlying map from Thyra, in order to access that information.

## Possible Solution

Simply add a method like this (copied and adapted from other adapters):

```
  /** \brief Get the embedded non-const Tpetra::Vector. */
  RCP<const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> >
  getConstTpetraMap() const;
```

I can set up a quick PR for this, if the proposed addition is deemed reasonable. I can also add a quick check in the tpetra adapters tests.

bartgol:
Fixes #4616

@trilinos/thyra

## Description

This PR adds a getter to Thyra::TpetraVectorSpace, to retrieve the underlying Tpetra::Map.

## Motivation and Context

All other Thyra::TpetraXYZ wrappers allow this behavior. While for (Multi)Vector such behavior is bypassable (via methods of base classes returing local data), the Map's capabilities are not accessible via base classes.

## Related Issues

* Closes 4616

## How Has This Been Tested?

I added a check in `TpetraThyraWrappers_UnitTests.cpp`, to check that the map wrapped in a newly created Thyra::TpetraVectorSpace was indeed the same as the original Tpetra::Map

## Checklist

### DIFF
--- a/packages/thyra/adapters/tpetra/src/Thyra_TpetraThyraWrappers_decl.hpp
+++ b/packages/thyra/adapters/tpetra/src/Thyra_TpetraThyraWrappers_decl.hpp
@@ -176,7 +176,10 @@ template <class Scalar = Tpetra::Vector<>::scalar_type,
 class TpetraOperatorVectorExtraction {
 public:
 
-  // ToDo: Get a Tpetra::Map from a Thyra::VectorSpaceBase?
+  /** \brief Get a const Tpetra::Map from a const Thyra::VectorSpaceBase object.
+   */
+  static RCP<const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> >
+  getTpetraMap(const RCP<const VectorSpaceBase<Scalar> > &vs);
 
   /** \brief Get a non-const Tpetra::Vector from a non-const
    * Thyra::VectorBase object.
--- a/packages/thyra/adapters/tpetra/src/Thyra_TpetraThyraWrappers_def.hpp
+++ b/packages/thyra/adapters/tpetra/src/Thyra_TpetraThyraWrappers_def.hpp
@@ -214,6 +214,15 @@ Thyra::createConstLinearOp(
 
 namespace Thyra {
 
+template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
+RCP<const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> >
+TpetraOperatorVectorExtraction<Scalar,LocalOrdinal,GlobalOrdinal,Node>::
+getTpetraMap(const RCP<const VectorSpaceBase<Scalar> > &vs)
+{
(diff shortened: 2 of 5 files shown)


bartgol:
@bartlettroscoe Ah, I never used that class. In my code Tpetra is just one of the possible backends, and in some cases I want to allow failure in the conversion, so that I can move to the next backend and try again. But I will add a getter to that class right away, yes. Thanks.

@mhoemmen I think the tpetra adapters do not offer free function at all. That would be another big discussion, since the Epetra adapters could use a refactor that makes the design similar to the Tpetra ones (namely, have a Thyra::EpetraXYZ that inherits from Thyra::SpmdXYZ, I tried that in the past (the branch is still up somewhere), but due to time bandwidth I abandoned). So I don't think we want to replicate the Epetra behavior, especially since Tpetra has the nice Extraction class that @bartlettroscoe mentioned.

Thank you all for the feedback.

bartgol:
> @bartgol,
> 
> Looks good to me. I am surprised this function was not already there. 

Speak o fthe devil, here's what I just found in the `TpetraOperatorVectorExtraction` class:

```
class TpetraOperatorVectorExtraction {
public:

  // ToDo: Get a Tpetra::Map from a Thyra::VectorSpaceBase?
```
So yeah, I guess this addition was already planned in the past...

bartgol:
The message says the autotester failed, but when I open the [cdash link ](https://testing-vm.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-4618-test&field2=buildstarttime&compare2=84&value2=NOW) I don't see any red. Am I missing something about how the autotester work?

mhoemmen:
@trilinos/framework yo

bartgol:
Also, from the autotester message:

>  Note: Testing will normally be attempted again in approx. 2 Hrs 30 Mins. If a change to the PR source branch occurs, the testing will be attempted again on next available autotester run.

That was 4 days ago.

PR #4620, opened 2019-03-14
Tpetra MV: Use specialized KokkosBlas::dot in multiply()

cgcgcg:
@trilinos/tpetra 

## Description
Use a specialized KokkosKernels function in `multiply` when either one of the two factors has only one column.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -4127,17 +4127,38 @@ namespace Tpetra {
       const LO C_lclNumRows = C_tmp->getLocalLength ();
       const LO C_numVecs = C_tmp->getNumVectors ();
       auto C_lcl = C_tmp->template getLocalView<device_type> ();
-      auto C_sub = Kokkos::subview (C_lcl,
-                                    std::make_pair (LO (0), C_lclNumRows),
-                                    std::make_pair (LO (0), C_numVecs));
       const char ctransA = (transA == Teuchos::NO_TRANS ? 'N' :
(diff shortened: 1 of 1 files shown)


cgcgcg:
Hm, does Stokhos not have a `dot` with a matching signature?

mhoemmen:
@cgcgcg wrote:
> Hm, does Stokhos not have a `dot` with a matching signature?

Ah, that's how it works -- Stokhos' PCE specialization of Tpetra has `dot` return `dot_type`, which happens to be `double` and not `scalar_type`.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -4127,17 +4127,38 @@ namespace Tpetra {
...
-                                    std::make_pair (LO (0), C_numVecs));
       const char ctransA = (transA == Teuchos::NO_TRANS ? 'N' :
                             (transA == Teuchos::TRANS ? 'T' : 'C'));
       const char ctransB = (transB == Teuchos::NO_TRANS ? 'N' :
                             (transB == Teuchos::TRANS ? 'T' : 'C'));
       const impl_scalar_type alpha_IST (alpha);
 
-      ::Tpetra::Details::ProfilingRegion regionGemm ("Tpetra::MV::multiply-call-gemm");
-      KokkosBlas::gemm(&ctransA, &ctransB, alpha_IST, A_sub, B_sub, beta_local, C_sub);
+      // The first two cases are specializations for inner products between a vector and a multivector
```
mhoemmen: `Tpetra::idot` has a specialization for this case.  We should perhaps lift that out into a single function.

cgcgcg:
Replaced by PR #5400

PR #4403, opened 2019-02-14
MueLu/Xpetra: Add Matrix Labels / Make Visible in Tpetra

csiefer2:
As per the request of @jjellio

PR #4624, opened 2019-03-14
Xpetra: Apply labels to wrapper and underlying object

cgcgcg:
@trilinos/xpetra 

## Description
`setObjectLabel` for matrix objects now applies the label to both the wrapper and the underlying matrix object. 

@pwxy


jhux2:
@cgcgcg Will it be a problem that two objects (wrapper and underlying object) both have the same label?

cgcgcg:
@jhux2 We'll see :-) Originally, the label was only set on the wrapper. That changed with PR #4403 and labels were only applied to the underlying object..

PR #4629, opened 2019-03-15
Teuchos: VerboseObject fix filestreams

### LINKED ISSUE #4731: Teuchos: VerboseObject fix filestreams
Currently, VerboseObject provides an option to allow output to be redirected to a file. Unfortunately, this feature results in every MPI process opening/writing the file, which is not typically what an OStream is used to accomplish. For file output, this implementation leads to extreme performance degradation with large parallel jobs.

@trilinos/teuchos @bartlettroscoe @mhoemmen 

## Expectations
VerboseObject should write to a single file (and be as performant at stdout.

## Current Behavior
Currently, we open multiple files and typically write exactly the same data to the same file. This gives the illusion of working as intended. Since the file will actually be correct at the end of the run. The issue is that every MPI process is writing the data, and this is not intended.

## Motivation and Context
ATDM apps would like to use this functionality. We uncovered/identified this issue by running with VerboseObject output to files with >64 nodes.

## Possible Solution
This fix is straightforward, and is addressed in PR#4629

(This is an issue to map to the PR... I should create the issues first in the future ... woops!)

jjellio:
@trilinos/teuchos 

## Description

Currently, VerboseObject provides an option to allow output to be redirected to a file. Unfortunately, this feature results in every MPI process opening/writing the file, which is not typically what an OStream is used to accomplish. For file output, this implementation leads to extreme performance degradation with large parallel jobs.

This patch accomplishes two things:

1) It restricts file opening/writing to a single MPI process.
The implementation of this is problematic, because VerboseObject's sublist reader has no access to any Comm. Ideally, the interface should be refactored to allow a comm to be passed, so that the logic
can use the com to enforce which rank writes output.

For now, the implementation guards the use of Teuchos::GlobalMPISession to obtain the process' global rank, and uses this to restrict `std::ofstream` construction.

2) This patch provides a means to configure a large file buffer for the `ofstream`.
Implementing this is problematic as well. The question is where and how to store the allocated buffer. Extending the FancyOStream class to add storage for a buffer would imply that other OStreams could support buffer modifications (which is a much broader impact than configuring a file stream buffer - it isn't clear that any arbitrary OStream can support this.)

To handle construction and deallocation of the buffer, a static `std::vector` of `std::vector<char>` is stored in the associated source files' object code. This is not ideal, but it should result in the memory being deallocated at program exit (not FancyOStream deconstruction though).

Moreover, this implementation is not thread safe w.r.t. to creating this OStream object. Concurrency could be guarded using C++11's concurrency features (or possibly Kokkos if using Kokkos inside this package is allowed).

I am open to suggestions to handle any of the above issues better.  `GlobalMPISession` is guarded by the Teuchos macro, `HAVE_TEUCHOS_MPI`, and there is an implicit assumption that MPI has been initialized and that calling `GlobalMPISession::getRank` is safe.

I suppose if this patch isn't accepted, then this code path should really be eradicated, because it presented a very hard to identify scaling problem, simply due to a change in Stratimikos parameters in an XML file.

## How Has This Been Tested?
Full sems/std testing suite (release/gcc-7.3)

## Checklist

- [ ] I have added tests to cover my changes. - It is not obvious how to test this. The code has been verified by using the system tool 'strace' to ensure single process opening/writing of the file

closes #4731

### DIFF
--- a/packages/teuchos/parameterlist/src/Teuchos_VerboseObjectParameterListHelpers.cpp
+++ b/packages/teuchos/parameterlist/src/Teuchos_VerboseObjectParameterListHelpers.cpp
@@ -117,19 +117,81 @@ void Teuchos::readVerboseObjectSublist(
   *verbLevel = VerbosityLevel_validator->getIntegralValue(
     voSublist,VerbosityLevel_name,VerbosityLevel_default
     );
+  // the default file string is nothing
   if (outputFileStr==OutputFile_default) {
     *oStream = null;
   }
+  // if a file is specified then output to an fstream
   else {
-    RCP<std::ofstream>
-      oFileStream = rcp(new std::ofstream(outputFileStr.c_str()));
-    TEUCHOS_TEST_FOR_EXCEPTION_PURE_MSG(
-      oFileStream->eof(), Exceptions::InvalidParameterValue,
-      "Error, the file \"" << outputFileStr << "\n given by the parameter\n"
-      "\'" << OutputFile_name << "\' in the sublist\n"
-      "\'" << voSublist.name() << "\' count not be opened for output!"
-      );
-    *oStream = fancyOStream(rcp_implicit_cast<std::ostream>(oFileStream));
+
+    // JJE: 14 March 2019
+    //  A fix for file output of an VerboseObject.
+    //
+    // This step is very important. With filestreams it does not make
+    // sense for multiple MPI ranks to open the same file.  Nor,
+    // does it seem inline with the OStream model that each stream
+    // represent a unique file. Perhaps, if that functionality is desired
+    // then the file name could be suffixed with the MPI Rank.
+    //
+    // A fundamental flaw with this implementation is that we have no knowledge
+    // of a communicator on which this OStream belongs. That makes the idea
+    // of using a rank ambiguous.
+    //
+    // The code below was added, and uses COMM_WORLD, because as-is
+    // this functionality was fundamentally broken. Without restricting
+    // the stream to a single rank, two severe consquences follow:
+    //   1) Each MPI process opens the file, which is not scalable;
+    //   2) Moreover, each MPI process *writes* to the file. Which
+    //      can give the illusion that things are working, if each
+    //      process writes exactly the same information (e.g., solver
+    //      convergence information for a bulk synchronous solve).
+    //      This introduces a terrible scalability problem, as the
+    //      filesystem is then tasked with coping with concurrent writes
+    //      to the same shared file, which is should make you cry a little.
+    //
+    // The resolution, is two fold:
+    //   1st, construct the ostream as a regular wrapper around cout
+    //   2nd, restrict the file creation and opening to a single process
+    //   Finally, map all ostreams except the fstream one to
+    //    a blackhole. Ensuring each rank has a functional stream
+    //    but that only one actually emits data to disk
+    //
+
+    // this could be a BlackHole, but calling setOutputToRootOnly does slightly
+    // more than simply blackhole output, it also disabled buffering across processes
+    *oStream = Teuchos::fancyOStream(Teuchos::rcpFromRef(std::cout));
+
+    // Until we resolve OS streams that are communicator aware, use rank 0
+    const int outputFileMPIRank = 0;
+
+    #if defined(HAVE_TEUCHOS_MPI)
+    const int rank = Teuchos::GlobalMPISession::getRank();
+    #else
+    const int rank = outputFileMPIRank;
+    #endif
+
+    if ( rank  == outputFileMPIRank) {
+      RCP<std::ofstream> oFileStream = rcp(new std::ofstream());
+      // If, in the future we decide to alter buffers, then
+      // change the fstream's buffer before calling open()
+
+      // open the fstream only on a single MPI process
+      oFileStream->open(outputFileStr);
+
+      TEUCHOS_TEST_FOR_EXCEPTION_PURE_MSG(
+        oFileStream->eof(), Exceptions::InvalidParameterValue,
+        "Error, the file \"" << outputFileStr << "\n given by the parameter\n"
+        "\'" << OutputFile_name << "\' in the sublist\n"
+        "\'" << voSublist.name() << "\' count not be opened for output!"
+        );
+      // wrap the fstream inside fancyOStream
+      *oStream = fancyOStream(rcp_implicit_cast<std::ostream>(oFileStream));
+    }
+
+    #if defined(HAVE_TEUCHOS_MPI)
+    // ensure that only one stream actually emits data
+    (*oStream)->setOutputToRootOnly(outputFileMPIRank);
+    #endif
   }
 #ifdef TEUCHOS_DEBUG
   voSublist.validateParameters(*getValidVerboseObjectSublist());


packages/teuchos/parameterlist/src/Teuchos_VerboseObjectParameterListHelpers.cpp
```diff
@@ -117,19 +137,94 @@ void Teuchos::readVerboseObjectSublist(
...
+        "Error, the file \"" << outputFileStr << "\n given by the parameter\n"
+        "\'" << OutputFile_name << "\' in the sublist\n"
+        "\'" << voSublist.name() << "\' count not be opened for output!"
+        );
+      // wrap the fstream inside fancyOStream
+      *oStream = fancyOStream(rcp_implicit_cast<std::ostream>(oFileStream));
+    }
+
+    #if defined(HAVE_TEUCHOS_MPI)
+    // ensure that only one stream actually emits data
```
mhoemmen: Doesn't this render ineffective all the changes above?

jjellio: No, that step is necessary. If you follow the logic, the initial OStream, is actually std::cout. The ensures each process has an ostream.

The final tidbit, simply blackholes everyone that didn't go through the if condition (the rank allowed output is the rank that goes into the if.

Unless, I botched the guards...

jjellio:
@bartlettroscoe 

This is Empire/Mini-EM.  The complexity, comes because this bug arises by using Stratimikos + Thyra/(Belos or whatever their XML deck tells stratimikos to use).  They build a Belos Status Test, which uses whatever OSTream the class has (by inheriting from VerboseObject). So, to go a route where they build their own FStream output, would entail patching that into Stratimikos, which would be unusual, since that packages doesn't really define Parameters, they simply expose other packages.

You *could* modify the higher level code, but it would entail hunting *all* classes that use or inherit from VerboseObject.  Infact, adding a communicator to the interface would entail this. This would likely touch all interfaces in Stratimikos, and possibly Thyra as well, because the communicator parameter would need to be propagated to the point where the VerboseObject sublist is parsed (Thyra), but communicators are available in the classes that subclass (stratimikos).

The above is true, I believe.... The hunt to isolate was a bit confusing.  The astounding discovery, is that because the code is bulk synchronous, the 'write' system calls all happen in at the same time. Every write, actually has the exact same data size. So these calls get serialized and they actually overwrite each other perfectly. Which gives the illusion the file is correct.  You can verify this behavior, by prepending the writing process' rank to the output (in Stratimikos' interface). You get a logfile with convergence history written by various ranks (which every rank happened to be the last to get the data written.).   What EMPIRE saw, was that large scale runs  were taking hours to finish what should have been a simple 1 minute solve.

The feature was turned on, because developers wanted to understand how the solver(s) behave, and it is nice when you can place that data into separate files.

As for the review criticisms:

I am calling the static function `Teuchos::GlobalMPISession::getRank()`, that is really the best I can do. I also provide a user configurable way to choose who writes. But the problem, is that those Ranks are global COMM_WORLD ranks. To make it ideal, the rank's would need to be set using a comm.

jjellio:
I'll cross post the performance data using this patch.


Time in seconds for Mini-EM, 64 nodes, flat MPI on KNL (using the patched code).

|		|	stdout	|	Vanilla	|	64kb	|
|	---	|	---	|	---	|	---	|
|		|	73.4477	|	73.4046	|	73.2069	|
|		|	73.0025	|	73.455	|	73.0733	|
|		|	73.4146	|	73.1872	|	72.8896	|
|		|	73.1576	|	73.7567	|	73.1381	|
|		|	72.9953	|	75.3206	|	73.2944	|
|		|	73.2635	|	73.1576	|	73.1555	|
|		|	73.125	|	73.5493	|	73.1448	|
|		|	72.9246	|	73.0669	|	73.2628	|
|		|	74.527	|	72.8133	|	73.0058	|
|		|	73.0017	|	73.271	|	73.3367	|
|	**Mean**	|	**73.28595**	|	**73.49822**	|	**73.15079**	|
|	**Std**	|	**0.447436568**	|	**0.657317223**|	**0.128975598**	|
|	Median	|	73.1413	|	73.3378	|	73.15015	|

Without the patch, the runtime on 64 nodes (a small job) using flat MPI was about `300 sec`. On the large scale runs, the descrpancy in runtime vs. stdout scaled with the number of MPI procs. (over an hour delta... yikes)

jhux2:
@bartlettroscoe For the time being, EMPIRE has reverted its source code changes that revealed this bug.  So we have some time ensure that this is fixed properly.

jjellio:
I ran some longer runs over the weekend. Specifically to see if the buffer option is worth the effort of pushing into a library (it is simple to code)

To evaluate the benefit of configurable buffers, you really need to write a sufficient amount of data.  The short runs I used in development write several KB of data, which isn't representative of how this option would be used. The expected  usecase is a long run that studies convergence over time. To see what that would create in terms of file IO, I ran 10k time steps (1000x more than in development).

When using `stdio`, this generates 35MB of out
When using separate solver files this generates 4 files:
`stdout = 2MB`, `solver1 = 16MB`, `solver2 = 5.5`, `solver3= 11MB` (34.5MB of total IO)

The table below shows 4 samples, with each run taking about 1.5 hours.
The 99% CI intervals point out that the means are actually identical (the variation is within the overlap of the each respective samples' CI).
From the above, we could argue that buffering isn't worth the trouble. Using a larger buffer is probably a good thing, since it costs nothing and doesn't appear to have any negative side effects. The reality is that going to disk is expensive and noisy, and buffering can't hide that.

I'd say, refactor the code to do the right thing with files (which it does), and drop the buffering stuff. This is a low priority thing, and the gain isn't worth the extra effort IMO.

|		|	64kb	|	stdout	|	van	|
|	---	|	---	|	---	|	---	|
|		|	5172.62	|	5170.74	|	5194.19	|
|		|	5191.94	|	5202.39	|	5200.03	|
|		|	5198.5	|	5207.71	|	5204.51	|
|		|	5207.06	|	5210.13	|	5207.67	|
|	Mean	|	5192.53	|	5197.7425	|	5201.6	|
|	Std	|	12.68382829	|	15.83931402	|	5.066705044	|
|	Median	|	5195.22	|	5205.05	|	5202.27	|
|	99% CI	|	37.04254536	|	46.25799837	|	14.79708233	|

bartlettroscoe:
@jjellio said:

> I'd say, refactor the code to do the right thing with files (which it does), and drop the buffering stuff. This is a low priority thing, and the gain isn't worth the extra effort IMO.

Sounds good to me.  What is left should be very minimal.  Can you remove the buffer-related stuff and hard push the branch?

Longer term, I think we do want to allow each process to print output at which time we should write a specialized `std::ostream` DECORATOR subclass that has a `Teuchos::Comm<int>` object and will gather output from each process (at some interval and when flushing occurs) to the root process to print there.

jjellio:
@bartlettroscoe

I will clean it up. Fixing the file IO without anything fancy will allow this to be used, without really any penalty over stdout. Which the apps should enjoy.

And Yes, I think at some point creating an OStream that is communicator aware is the way to go. This would allow things like MueLu to avoid the problems of 'losing rank 0' and it makes a 'rank' for filtering output meaningful.

jjellio:
@bartlettroscoe @mhoemmen 

I've removed the buffer stuff + squashed the commits.  I also opened an issue so this can be automagically closed like other cool git kids.

It should be easier to review, as the changes are effectively around the if statement for fstream usage.

mhoemmen:
@jjellio wrote:
> I also opened an issue so this can be automagically closed like other cool git kids.

The commit message that fixes the issue needs to say `Fix #NNNN` or something like it, where `NNNN` is the issue number.  That's what does the automagic.

packages/teuchos/parameterlist/src/Teuchos_VerboseObjectParameterListHelpers.cpp
```diff
@@ -117,19 +117,81 @@ void Teuchos::readVerboseObjectSublist(
...
+
+    // this could be a BlackHole, but calling setOutputToRootOnly does slightly
+    // more than simply blackhole output, it also disabled buffering across processes
+    *oStream = Teuchos::fancyOStream(Teuchos::rcpFromRef(std::cout));
+
+    // Until we resolve OS streams that are communicator aware, use rank 0
+    const int outputFileMPIRank = 0;
+
+    #if defined(HAVE_TEUCHOS_MPI)
+    const int rank = Teuchos::GlobalMPISession::getRank();
```
jhux2: @jjellio What happens if process 0 in MPI_COMM_WORLD is no longer part of the current MPI communicator?  No output?

jjellio: @jhu2 This is a broader problem.  OStreams (and VerboseObject) are communicator-dumb. We discussed this above, the fix for that is more general and invasive.  Effectively, we need OStreams to become linked to a communicator.

Currently, when you blackhole a communicator (setOutputToRootOnly) this is actually using MPI_COMM_WORLD (from Teuchos GlobalMPISession ).  So existing output stuff is just as vulnerable to this as the fstream business. (Stratimikos is setting output to root only FYI).

I think this headache plagues MueLu more than anyone else.

As-is, this seems to work, since most (all?) codes like to use `setOutputToRootOnly` and guru like the MueLu devs have worked hard to make sure Rank 0 doesn't disappear.

bartlettroscoe: @jjellio and @jhux2,

I actually think this could be pretty easy to fix.  You just add a virtual function to this interface with a default implementation:

```
  virtual bool getVerbObjCommRank() const { return Teuchos::GlobalMPISession::getRank(); }
```

then subclasses that know they are parallel objects and have a `Teuchos::Comm<int>`  object (call it `comm_`) can define this as:

```
  virtual bool getVerbObjCommRank() const override { return comm_->getRank(); }
```

What do you think?

NOTE: We don't want to overload virtual functions, hence the longer name `getVerbObjCommRank()`.

jjellio: Clever!

Yes, that would default to what I have now.

I am not sure what Stratimikos/ThyraLinearOp expose in terms of communicators.

bartlettroscoe: @jjellio said:

> I am not sure what Stratimikos/ThyraLinearOp expose in terms of communicators.

That is true.  Since Thyra is an ANA (abstract numerical algorithm) interface, the objects don't return comms because that allows the client to assume a runtime model for the machine.  But given the practicality of comms in cases like this, we might consider adding a virtual function for returning an `RCP<const Teuchos::Comm<int>>` object with a default `null` implementation.

But that can happen later. For now, hopefully this PR will make things a little better.

But again, note that the fallback that gives complete control by the client is to create your own `Teuchos::FancyOStream` object wrapping any `std::ostream` subclass you want and then you can make it do anything you want.  It is just this ParameterList construction case that is the challenge.

packages/teuchos/parameterlist/src/Teuchos_VerboseObjectParameterListHelpers.cpp
```diff
@@ -117,19 +117,81 @@ void Teuchos::readVerboseObjectSublist(
...
+    //   1st, construct the ostream as a regular wrapper around cout
+    //   2nd, restrict the file creation and opening to a single process
+    //   Finally, map all ostreams except the fstream one to
+    //    a blackhole. Ensuring each rank has a functional stream
+    //    but that only one actually emits data to disk
+    //
+
+    // this could be a BlackHole, but calling setOutputToRootOnly does slightly
+    // more than simply blackhole output, it also disabled buffering across processes
+    *oStream = Teuchos::fancyOStream(Teuchos::rcpFromRef(std::cout));
```
bartlettroscoe: This is safe since `std::cout` is guaranteed to live past all possible references, even after `main()` ends.

packages/teuchos/parameterlist/src/Teuchos_VerboseObjectParameterListHelpers.cpp
```diff
@@ -117,19 +117,81 @@ void Teuchos::readVerboseObjectSublist(
...
+
+    // Until we resolve OS streams that are communicator aware, use rank 0
+    const int outputFileMPIRank = 0;
```
bartlettroscoe: Does every MPI_Comm object not guaranteed to have a proc with rank 0?  If not, then how do you pick the "root" rank?

bartlettroscoe:
FYI: Just added `AT: RETEST`.  Looks like the last PR GCC 4.8.4 build randomly failed (see https://github.com/trilinos/Trilinos/issues/3276#issuecomment-477162374).

packages/teuchos/parameterlist/src/Teuchos_VerboseObjectParameterListHelpers.cpp
mhoemmen: @bartlettroscoe Ranks are always 0, 1, ... size-1.

PR #4638, opened 2019-03-18
Ifpack2 - change set_scratch_size method according to KOKKOS_DEPRECATED_CODE

kyungjoo-kim:
## Description

Change ``set_scratch_size`` method according to ``KOKKOS_DEPRECATED_CODE`` requested from SPARC.


packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1769,11 +1769,19 @@ namespace Ifpack2 {
...
 
+#if defined(KOKKOS_ENABLE_DEPRECATED_CODE)
+	const Kokkos::TeamPolicy<execution_space,ExtractAndFactorizeTag>
+	  policy(packptr.extent(0)-1, team_size, vector_loop_size); 
+	Kokkos::parallel_for("ExtractAndFactorize::TeamPolicy::run<ExtractAndFactorizeTag>", 
+                             policy.set_scratch_size(0,Kokkos::PerTeam(per_team_scratch)), *this);
+#else
 	Kokkos::TeamPolicy<execution_space,ExtractAndFactorizeTag>
 	  policy(packptr.extent(0)-1, team_size, vector_loop_size); 
+        policy.set_scratch_size(0,Kokkos::PerTeam(per_team_scratch));
```
mhoemmen: Does this not build if deprecated code is enabled?  I saw your comment, though -- it looks like you prefer the constructor form and want to keep it around.

kyungjoo-kim: It is not a matter of my preference. It does not build when kokkos deprecated code is OFF. I don't know why. Only difference is that one returns an object and the other returns reference of this object. Do you have a clue ?

mhoemmen: @kyungjoo-kim It's a bit weird to declare a named variable and call a method on it (`set_scratch_size`) in the same declaration.  I think the build issue is that if deprecated code is OFF, `set_scratch_size` returns a reference, whereas it returns a value if deprecated code is ON.  In the former case, that would make the type of the variable declaration (`TeamPolicy`, not `TeamPolicy&`) incorrect.  If you separate out the variable declaration from the `set_scratch_size` call, the code should compile in both cases.

kyungjoo-kim:
Oh..... I forgot that I use the same branch for developing ifpack2. I accidently push into the branch work in progress.

PR #4640, opened 2019-03-18
Belos,Tpetra: Fix #4626, #4639, #4633

### LINKED ISSUE #4626: Belos::MultiVecTraits: Tpetra specialization is slow on GPU, because it creates local MV & thus does extra CUDA allocations
@trilinos/belos @trilinos/tpetra @cgcgcg 

The Tpetra specialization of `Belos::MultiVecTraits` has two methods that need to create temporary "local" MultiVector instances.  This creation does a new DualView allocation each time.  This makes running on the GPU slow, especially for classical Gram-Schmidt in GMRES, or in general whenever doing `X^T * Y` or `X^H * Y` for either X or Y having multiple columns. 

I am working on a fix.  The idea is to maintain a static DualView "pool" that the local MultiVector can view.  One must be careful not to let any DualView instance persist past `Kokkos::finalize`.

### LINKED ISSUE #4633: Tpetra::MultiVector constructor that takes a device View uses create_mirror instead of create_mirror_view
@trilinos/tpetra @vbrunini 

```
```

### LINKED ISSUE #4639: Tpetra::MultiVector::reduce broken if getStride() > getLocalLength()

## Current Behavior

## Motivation and Context

## Possible Solution

I have a fix ready.

## Steps to Reproduce

## Related Issues

* Related to #4626, #4633

mhoemmen:
@trilinos/belos @trilinos/tpetra @cgcgcg @vbrunini 

## Description

  - Fix #4639 (a Tpetra bug)
  - Fix #4633 (a Tpetra not quite bug but weird wrinkle) 
  - Fix #4626 

The solution for #4626 needs an optimization (always allocate contiguous DualViews, by making them nonowning from a 1-D memory pool), but at least should avoid the performance issue of repeatedly allocating and freeing CUDA UVM allocations.

## Related Issues

* Closes #4626, #4639, #4633


mhoemmen:
Please don't merge this yet -- I realize the Belos code won't pass the SERIAL PR test.  Just wanted to see what people think.

mhoemmen:
Superseded by https://github.com/trilinos/Trilinos/pull/4648 .  Thanks all!

PR #4643, opened 2019-03-18
Reduce test wallclock time for cee-rhel6 gnu-7.2.0 build (TRIL-243)

bartlettroscoe:
Since this is not an OpenMP build, removed all of the OpenMP env vars and set
--bind-to none.  This reduced runtime from 45 minutes to 10 minutes.

### DIFF
--- a/cmake/std/atdm/cee-rhel6/environment.sh
+++ b/cmake/std/atdm/cee-rhel6/environment.sh
@@ -47,14 +47,6 @@ export ATDM_CONFIG_BUILD_COUNT=$ATDM_CONFIG_MAX_NUM_CORES_TO_USE
 
 module purge
 
-if [[ "$ATDM_CONFIG_NODE_TYPE" == "OPENMP" ]] ; then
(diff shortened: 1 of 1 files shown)


cmake/std/atdm/cee-rhel6/environment.sh
```diff
@@ -80,6 +72,9 @@ if [[ "$ATDM_CONFIG_COMPILER" == "CLANG-5.0.1_OPENMPI-1.10.2" ]]; then
   export ATDM_CONFIG_MKL_ROOT=${CBLAS_ROOT}
 elif [[ "$ATDM_CONFIG_COMPILER" == "GNU-7.2.0_OPENMPI-1.10.2" ]] ; then
   module load sparc-dev/gcc-7.2.0_openmpi-1.10.2
+  unset OMP_NUM_THREADS  # SPARC module sets these and we must unset!
```
mhoemmen: Yeah, if you do `ctest -j N` with `N > 1` and run with threads, binding threads to anything would be Quite Ungood.

PR #4644, opened 2019-03-18
Fix EMPIRE CUDA Build Failure

jmgate:
# DO NOT MERGE

Only using this for review.

### DIFF
--- a/packages/kokkos/bin/nvcc_wrapper
+++ b/packages/kokkos/bin/nvcc_wrapper
@@ -191,8 +191,12 @@ do
     if [ $stdcxx_applied -eq 1 ]; then
        echo "nvcc_wrapper - *warning* you have set multiple optimization flags (-std=c++1* or --std=c++1*), only the first is used because nvcc can only accept a single std setting"
     else
-       shared_args="$shared_args $1"
-       stdcxx_applied=1
+      if [ "$1" == "--std=c++1z" ] | [ "$1" == "-std=c++1z" ]; then
+        shared_args="$shared_args --std=c++14"
+      else
+        shared_args="$shared_args $1"
+      fi
+      stdcxx_applied=1
     fi
     ;;
 


jmgate:
Just to be sure, @bathmatt, this is the patch you want merged into `EM-Plasma/Trilinos:develop`?

packages/kokkos/bin/nvcc_wrapper
```diff
@@ -191,8 +191,12 @@ do
     if [ $stdcxx_applied -eq 1 ]; then
        echo "nvcc_wrapper - *warning* you have set multiple optimization flags (-std=c++1* or --std=c++1*), only the first is used because nvcc can only accept a single std setting"
     else
-       shared_args="$shared_args $1"
-       stdcxx_applied=1
+      if [ "$1" == "--std=c++1z" ] | [ "$1" == "-std=c++1z" ]; then
+        shared_args="$shared_args --std=c++14"
```
mhoemmen: Wouldn't the right mapping be `--std=c++17`?

jmgate:
Sorry, @trilinos-autotester.  I applied the "AT:  WIP" label when creating the issue, but that didn't seem to stick for some reason.  GitHub glitch perhaps.  It also didn't take my reviewer or assignee either.

jmgate:
Paging @bathmatt.  Can you confirm this is the change you want merged into our `develop`?  See @mhoemmen's comment above.

packages/kokkos/bin/nvcc_wrapper
ndellingwood: @mhoemmen I'm guessing this is a workaround to this issue: kokkos/kokkos#2035

mhoemmen: It looks like @crtrott et al. are discussing it, so I think we should defer these changes until that discussion gets resolved.

jmgate:
To be clear, folks, I created this PR so I could double-check with @bathmatt that these were the changes he wanted merged into *EMPIRE's fork* of Trilinos.  We have no intention of merging this as is into Trilinos itself&mdash;this is a temporary stop-gap to get our CUDA builds up and running again.  Closing&hellip;

bathmatt:
Sorry, just saw this, yes that is the change we need

PR #4648, opened 2019-03-18
Belos,Tpetra: Fix #4626, #4639, #4633

### LINKED ISSUE #4626: Belos::MultiVecTraits: Tpetra specialization is slow on GPU, because it creates local MV & thus does extra CUDA allocations

### LINKED ISSUE #4633: Tpetra::MultiVector constructor that takes a device View uses create_mirror instead of create_mirror_view

### LINKED ISSUE #4639: Tpetra::MultiVector::reduce broken if getStride() > getLocalLength()

mhoemmen:
@trilinos/belos @trilinos/tpetra @cgcgcg @vbrunini 

## Description

This PR supersedes #4640.

  - Fix #4639 (an actual correctness bug in `Tpetra::MultiVector::reduce` that hindered solving #4626)
  - Fix #4633 (a strange thing in Tpetra::MultiVector that could lead to bugs, especially with FEMultiVector)
  - Fix #4626 (a Belos performance issue, especially on GPUs)

## Related Issues

* Closes #4626, #4639, #4633


mhoemmen:
There's something really quite weird about the test results, especially with the Intel build.  CUDA build passes, though!  I'll mark this as WIP, but please feel free to review.

mhoemmen:
I fixed the build warnings and added an extra check for what I suspect is an Intel compiler bug.  If the tests still fail in the Intel build, then I may need to change DualViewPool's implementation to use a common CUDA UVM memory pool for MultiVectors of all types (so that initialization of the memory pool happens in a nontemplated function), rather than a memory pool per (Scalar, Device) combination.

mhoemmen:
@trilinos/framework PR testing didn't get my new commits.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3677,77 +3692,71 @@ namespace Tpetra {
...
-          std::logic_error, ": srcColView and dstColView_dev have different "
-          "dimensions.  Please report this bug to the Tpetra developers.");
-        Kokkos::deep_copy (dstColView, srcColView_dev);
+	Kokkos::deep_copy (A_view, srcView_dev);
+      }
+    }
+    else {
+      for (size_t j = 0; j < numCols; ++j) {
+	const size_t srcCol = this->whichVectors_[j];
+	auto dstColView = Kokkos::subview (A_view, rowRange, j);
```
mhoemmen: @trilinos/framework @GeoffDanielson @william76 @bartlettroscoe @jwillenbring @prwolfe  You can literally see that I've removed the line of code that shows up in PR testing as an unused variable warning.

mhoemmen:
@bartlettroscoe @jwillenbring @william76 @GeoffDanielson @prwolfe @rrdrake @theguruat12 @trilinos/framework @vbrunini @jclause 

PR testing fails to pick up new commits, mysteriously runs out of memory, etc.

mhoemmen:
@kddevin @trilinos/tpetra It's pretty hard to refactor Tpetra and make the deadline for backwards compatibility changes, if we can't get PRs through.

mhoemmen:
Here is the commit in question that fixes the warning, that nevertheless still shows up in PR tests: https://github.com/trilinos/Trilinos/pull/4648/commits/83ad1c2fa5920697d0692fde2c2127fcbe0a5145

prwolfe:
Come see me when you get in – this is worse than you realized.

prwolfe:
It looks like all PR builds that test stratimikos have been failing since Mar 19, 2019 18:53 UTC (10:53 MDT) This includes one I have in that did not touch any source at all. #4659

bartlettroscoe:
@prwolfe said:

> It looks like all PR builds that test stratimikos have been failing since Mar 19, 2019 18:53 UTC (10:53 MDT) This includes one I have in that did not touch any source at all. #4659

The post-push CI build clearly showed this last night as described in the below email I just sent out.

Are you and the other @trilinos/framework team members getting those CDash emails?  It is important to get those and react quickly to those because that means that eveyrone's PR builds will fail and no one will be able to merge anything (that flows through those packages).

----


From: Bartlett, Roscoe A 
Sent: Wednesday, March 20, 2019 9:30 AM
To: 'trilinos-developers@trilinos.org' <trilinos-developers@trilinos.org>
Subject: Trilinos 'develop' broken?

Hello Trilinos Developers,

FYI: The post-push CI build shown at:

https://testing-vm.sandia.gov/cdash-dev-view/index.php?project=Trilinos&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Linux-GCC-4.8.4-MPI_RELEASE_DEBUG_SHARED_PT_OPENMP_CI&field2=groupname&compare2=61&value2=Continuous&field3=buildstarttime&compare3=83&value3=4%20weeks%20ago

shows that somehow two Stratimikos tests are now failing in the Trilinos ‘develop’ branch in the GCC 4.8.4 OpenMP build (which should match the Trilinos GCC 4.8.4 OpenMP PR build). I think that means that almost all of the Trilinos PRs that change code upstream from Stratimikos will also be broken.

One of the commits from one of the PRs merged to ‘develop’ shown at:

   https://testing-vm.sandia.gov/cdash-dev-view/viewNotes.php?buildid=4746581#!#note0 

must have triggered this. 

Someone should do a root-cause analysis to see how this could have occurred.

Cheers,

-Ross

cgcgcg:
Hm, my PR for the Belos Stratimikos adapter was merged yesterday: #4606

prwolfe:
I get them, but just triaging my email could easily be a full time job. I frequently see issues in other ways or in this case from another e-mail first.

cgcgcg:
I believe I found the culprit PR #4650 where we changed the some defaults for PseudoBlockGMRES.
This does not explain, however, how this managed to get past the autotester..

cgcgcg:
My attempt at a fix: PR #4677

jhux2:
___
*[packages/tpetra/core/src/Tpetra_MultiVector_def.hpp, line 1815 at r2](https://reviewable.io/reviews/trilinos/trilinos/4648#-LaRuyuY2uC4O6YTlyk_:-LaRuyuY2uC4O6YTlyka:bfu5wjw) ([raw file](https://github.com/trilinos/trilinos/blob/6e83ef970a7e2433fd9815624eb382bd098b8a78/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp#L1815)):*
> ```cpp
>       auto map = x.getMap ();
>       Teuchos::RCP<const Teuchos::Comm<int> > comm =
> 	map.is_null () ? Teuchos::null : map->getComm ();
> ```

Minor quibble -- is this a tab?  reviewable shows a number of red ">>" marks, which I think mean tabs.

mhoemmen:
> Minor quibble -- is this a tab? reviewable shows a number of red ">>" marks, which I think mean tabs.

Oof, could be.  I will review my `.emacs` files to make sure they autoconvert tabs to spaces.

mhoemmen:
I tested with Intel 17.0.1, with OpenMP both on and off, and could not replicate the test failures seen earlier.  Thus, I'm just marking this as RETEST and hoping that the previous fixes will make it all work.

mhoemmen:
@trilinos/tpetra @kddevin I suspect Intel has a bug in its handling of `static` initialization, since the assert fails only in the Intel build.  (It even passes the CUDA build, and it certainly passes my Clang build.)  Regardless, I will need to work around the Intel bug, so I'm marking this PR as WIP.
```
 
  /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/Trilinos/packages/belos/tpetra/src/BelosMultiVecTraits_Tpetra.hpp:157:
 
  Throw number = 1
  
  Throw test that evaluated to true: !(d_view_ptr_ != nullptr)
 ```

mhoemmen:
@trilinos/tpetra @kddevin 

I'll roll the dice once more on this one.  If the Intel compiler persists in being buggy, I have a work-around in progress.

mhoemmen:
@trilinos/tpetra The same `!(d_view_ptr_ != nullptr)` nonsense that's clearly impossible.  I have a rewrite in progress that should work around this Intel compiler bug.

mhoemmen:
PR #4747 and PR #4734 (plus follow-ons to the latter) have superseded this PR.  See explanation in #4747.

mhoemmen:
See also branch https://github.com/mhoemmen/Trilinos/tree/TpetraStaticView3 , which contains the aforementioned salvaged and updated code that fixes #4626.

mhoemmen:
Update: See branch https://github.com/mhoemmen/Trilinos/tree/TpetraStaticView4.

mhoemmen:
PR #4765 supersedes this PR.  Thanks all!

PR #4650, opened 2019-03-19
Belos: Fix #994 (make ICGS default orthogonalizer for GMRES)

### LINKED ISSUE #994: Belos: Change GMRES default orthogonalizer from DGKS to ICGS 2-pass
@trilinos/belos @jjellio @hkthorn

mhoemmen:
@trilinos/belos @jjellio @hkthorn @vbrunini @iyamazaki 

## Description

Change `Belos::PseudoBlockGmresSolMgr`'s default orthogonalizer from DGKS (modified Gram-Schmidt with careful checks) to ICGS (classical Gram-Schmidt, repeated unconditionally).
  
## Motivation and Context

1. Get better performance with default settings
2. Make Belos behave like AztecOO out of the box

## Related Issues

* Closes #994

### DIFF
--- a/packages/belos/src/BelosPseudoBlockGmresSolMgr.hpp
+++ b/packages/belos/src/BelosPseudoBlockGmresSolMgr.hpp
@@ -208,8 +208,9 @@ namespace Belos {
      *   bigger than the "Block Size" parameter.
      * - "Orthogonalization" (\c std::string): The desired
      *   orthogonalization method.  Currently accepted values are
-     *   "DGKS", "ICGS", and "IMGS".  Please refer to Belos'
-     *   documentation for more details.
+     *   "DGKS", "ICGS", "IMGS", and optionally "TSQR" (depending on
+     *   build settings).  Please refer to Belos' documentation for
+     *   more details.
      *
      * For an explanation of "implicit" vs. "explicit" residuals,
      * please see the documentation of isLOADetected().  The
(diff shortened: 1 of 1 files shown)


jjellio:
I know in the past, when this came up. The idea was to provide a 'fast' set of parameters, but it seems that paradigm just doesn't work well =\

I seem to remember that DGKS doesn't performance substantially different from ICGS2. Infact, don't both perform similarly when loss of orthogonality is not detected? If the apps really want performance at the cost of accuracy, then doing ICGS1 (when they can) is the way to go for performance.

mhoemmen:
Thanks @cgcgcg !  I would like @hkthorn to comment, since this change may affect Anasazi users who are using Belos as the inner solver.

hkthorn:
@cgcgcg @mhoemmen As long as ICGS performs two passes, I agree to this change.

hkthorn:
@cgcgcg @mhoemmen This change should be made to all the solvers that require orthogonalization for commonality.  Otherwise, users will be confused why one version of GMRES uses this orthogonalization and others do not.

mhoemmen:
@hkthorn Would you prefer to make that change in the same PR or could we wait until a separate one?  I've been trying to merge as soon as tests pass since PR testing has been a disaster lately.

hkthorn:
@mhoemmen A separate one is fine, I just wanted to get my thought out there before it got lost.

mhoemmen:
@hkthorn Thanks!

mhoemmen:
@hkthorn I filed a separate issue for changing the other GMRES-like solvers' default: https://github.com/trilinos/Trilinos/issues/4657

PR #4653, opened 2019-03-19
Remove build cee-rhel6_gnu-4.9.3_openmpi-1.10.2_serial_shared_opt from nighlty testing (TRIL-243)

bartlettroscoe:
SPARC does not need this build and since the gnu-7.2.0 build's test suite was
made to run fast (see [TRIL-269](https://sems-atlassian-son.sandia.gov/jira/browse/TRIL-269)), we don't need to compare against the old
gnu-4.9.3 build anymore (see [TRIL-243](https://sems-atlassian-son.sandia.gov/jira/browse/TRIL-243))

Note that I did not remove the env for the GNU 4.9.3 build in case someone
wants to test it locally.

NOTE: The automated builds that submit to CDash will automatically get updated with this one change because a local cron job with a script the reads this file will get called.


bartlettroscoe:
Thanks @mhoemmen!

PR #4658, opened 2019-03-19
Panzer: clarify valid use of a ProjectToFace algorithm

rppawlo:
Clarify a use case in panzer by renaming an input flag

### DIFF
--- a/packages/panzer/disc-fe/src/evaluators/Panzer_ProjectToFaces_impl.hpp
+++ b/packages/panzer/disc-fe/src/evaluators/Panzer_ProjectToFaces_impl.hpp
@@ -60,10 +60,13 @@
 
 #include "Teuchos_FancyOStream.hpp"
 
+#include <cstring>
+
 template<typename EvalT,typename Traits>
 panzer::ProjectToFaces<EvalT, Traits>::
-ProjectToFaces(
-  const Teuchos::ParameterList& p)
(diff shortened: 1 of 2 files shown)


packages/panzer/disc-fe/src/evaluators/Panzer_ProjectToFaces_impl.hpp
```diff
@@ -106,6 +111,9 @@ ProjectToFaces(
     this->addEvaluatedField(gatherFieldNormals);
 
   } else {
+    TEUCHOS_ASSERT(quad_degree == 1); // One pt quadrature for fast method
+    TEUCHOS_ASSERT(std::strstr(basis->getCellTopology()->getBaseName(),"Hexahedron") != nullptr);
```
jmgate: What is this line doing?

rppawlo: Enforcing that it can only be used for hex topology. There are multiple Hexahedron topologies that can be returned from shards: "Hexehedron_8", "Hexahedron_27", ...

jmgate: Thanks&mdash;I saw `strstr` and my mind went to `stringstream`.

PR #4670, opened 2019-03-20
Zoltan2:  fix #4669 (crash in MPI_Allreduce reported by MiniEM)

### LINKED ISSUE #4669: Zoltan2 crashes in MPI_Allreduce for MiniEM test
@trilinos/zoltan2 
@jjellio @cgcgcg @jhux2 

The MiniEM team reports crashes in MPI_Allreduce in Zoltan2 when using MultiJagged partitioning.  They provide instructions for reproducing the problem in 
 https://gitlab-ex.sandia.gov/jhu/TrilinosSolverPerformance/issues/25

Similar issues were also reported by external user Robert Starr long ago, but could not be replicated in our environment.

## Steps to Reproduce

See
 https://gitlab-ex.sandia.gov/jhu/TrilinosSolverPerformance/issues/25

## Your Environment

Develop branch; linux or mac

kddevin:
@trilinos/zoltan2 @jjellio @jhux2 @cgcgcg 

## Description

MiniEM reported a crash in MPI_Allreduce when using Multijagged partitioning
https://gitlab-ex.sandia.gov/jhu/TrilinosSolverPerformance/issues/25

The problem was that processors within a subcommunicator were making different decisions about whether or not to migrate data during partitioning.  When some processors decided to migrate while others did not, the processors got out of sync with their calls to MPI_Allreduce.  Some Allreduces continued to execute without error (although, perhaps, with the wrong result), as long as the argument types agreed.  When the argument types finally didn't agree (several Allreduces later), the Allreduce crashed.  Thus, tracking this error was difficult, as the source of the error was not near the crash.

The changed code ensures that all processors within a communicator make the same decision about whether or not to migrate, as it uses the global number of coordinates within the communicator, rather than the local number on a processor.

Later, we may (or may not) choose to change the default value of MIN_WORK_LAST_DIM to have similar behavior to before this change.  The current value seems reasonable for now.

As a side note, I observed that after migration, the number of coordinates per processor could be quite imbalanced.  This is a separate (but important) issue that will be addressed later.  

## Motivation and Context
https://gitlab-ex.sandia.gov/jhu/TrilinosSolverPerformance/issues/25

## How Has This Been Tested?

New test miniem4669 added replicates the miniEM data using zoltan2's driver.
Also, tested with miniem in panzer:
 mpiexec -n 64 ./PanzerMiniEM_BlockPrec.exe --x-elements=64 --y-elements=64 --z-elements=64

## Checklist

- [ x] My commit messages mention the appropriate GitHub issue numbers.
- [ x] I have added tests to cover my changes.


jjellio:
@srajama1 I can report back (in probably 2 weeks), to say how this impacts the larger runs that ptlin does.

srajama1:
@jjellio : Thanks that will be be very useful.

kddevin:
Thanks for your comments, @srajama1 .
Would you prefer a solution that did Allreduce of the local decisions to get agreement within a communicator?  I can do that instead if you like.

srajama1:
@kddevin I was wondering about doing another all reduce. Honestly, I don't know if that is better. This is all a heuristic any way.

kddevin:
@trilinos/framework   What can I do to resolve these failing tests?  As far as I can tell, the Stratimikos tests that are failing do not call  Zoltan2, so the changes in this PR cannot be causing the failures.  Is this a known problem?

mhoemmen:
@kddevin I would just wait it out -- @cgcgcg committed something to Stratimikos to try to get things on track.

mhoemmen:
@trilinos/framework MueLu dies again:
```
collect2: fatal error: ld terminated with signal 9 [Killed]
compilation terminated.
```

PR #4677, opened 2019-03-20
Stratimikos tests: Set tolerance to avoid test failures

cgcgcg:
@trilinos/stratimikos 

## Description
Fix test failures in Stratimikos tests

### DIFF
--- a/packages/stratimikos/test/_FourByFour.belos.ifpack.xml
+++ b/packages/stratimikos/test/_FourByFour.belos.ifpack.xml
@@ -2,6 +2,16 @@
   <Parameter name="Matrix File" type="string" value="FourByFour.mtx"/>
   <ParameterList name="Linear Solver Builder">
     <Parameter name="Linear Solver Type" type="string" value="Belos"/>
+    <ParameterList name="Linear Solver Types">
+      <ParameterList name="Belos">
+        <Parameter name="Solver Type" type="string" value="Pseudo Block GMRES"/>
+        <ParameterList name="Solver Types">
+          <ParameterList name="Pseudo Block GMRES">
+            <Parameter name="Convergence Tolerance" type="double" value="1.0e-14"/>
+          </ParameterList>
+        </ParameterList>
+      </ParameterList>
+    </ParameterList>
     <Parameter name="Preconditioner Type" type="string" value="Ifpack"/>
   </ParameterList>
   <ParameterList name="LinearOpWithSolveTester"/>
--- a/packages/stratimikos/test/_FourByFour.belos.ml.xml
+++ b/packages/stratimikos/test/_FourByFour.belos.ml.xml
@@ -2,6 +2,16 @@
   <Parameter name="Matrix File" type="string" value="FourByFour.mtx"/>
   <ParameterList name="Linear Solver Builder">
     <Parameter name="Linear Solver Type" type="string" value="Belos"/>
+    <ParameterList name="Linear Solver Types">
+      <ParameterList name="Belos">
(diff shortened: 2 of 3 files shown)


bartlettroscoe:
FYI: @trilinos/framework

cgcgcg:
@bartlettroscoe That's a mystery to me. The changes of the defaults were in Belos. Did Stratimikos not get tested?

cgcgcg:
Before this change, the test would try to solve a 4x4 system with tolerance set to -1, so it would iterate for 1000 iterations. I could see how this could lead to instability that manifests itself differently from system to system..

packages/stratimikos/test/_FourByFour.belos.ifpack.xml
```diff
@@ -2,6 +2,16 @@
   <Parameter name="Matrix File" type="string" value="FourByFour.mtx"/>
   <ParameterList name="Linear Solver Builder">
     <Parameter name="Linear Solver Type" type="string" value="Belos"/>
+    <ParameterList name="Linear Solver Types">
+      <ParameterList name="Belos">
+        <Parameter name="Solver Type" type="string" value="Pseudo Block GMRES"/>
+        <ParameterList name="Solver Types">
+          <ParameterList name="Pseudo Block GMRES">
+            <Parameter name="Convergence Tolerance" type="double" value="1.0e-16"/>
```
mhoemmen: I'm glad the tolerance isn't -1 any more. Does it really need to be less than machine precision, though?

cgcgcg: Probably not. But this way the test will more or less do what it did before, i.e. solve to machine precision. Once this unblocks other PRs, we should probably check what we are actually testing here..

mhoemmen: I only mention it because if we're going to get the PR through anyway, we should at least make it some small integer factor times machine precision.

cgcgcg: Fair enough, let's make sure this will actually work.

bartlettroscoe:
@cgcgcg said:

> Before this change, the test would try to solve a 4x4 system with tolerance set to -1, so it would iterate for 1000 iterations.

How is it that the default tolerance is -1?  That seems like a defect that should be fixed rather than have to update these inputs. I know that when I added these tests it did not iterate for 1000s of iterations.  There must be some problem with lack of translation or asserting values or something.

jwillenbring:
> @bartlettroscoe That's a mystery to me. The changes of the defaults were in Belos. Did Stratimikos not get tested?

First, I want to thank everyone for looking at this and getting a PR issued, especially @cgcgcg.

I dug into the cause of this and it looks like an additive assumption of branches issue with #4606 and #4650.  I looked at the target branches for both tests and although they were different, both were before either commit was applied to develop. The builds fired up less than 2.5 hours apart. This is the kind of thing we will see from time to time, unless we completely serialize our commits, which has very limited scalability.

srajama1:
@jwillenbring : I don't understand why you thing this is due to additive commits. They look orthogonal to me.

bartlettroscoe:
Can we revert #4606 for now and fix it so that it restores correct behavior for default solve tolerance instead of taking this approach?   See analysis in https://github.com/trilinos/Trilinos/issues/4678#issuecomment-475017379.

For now I will remove 'AT: AUTOMERGE'.

cgcgcg:
@bartlettroscoe Sure, if you can get that faster through the autotester, then go for it!

Issue #4678, opened 2019-03-20
Stratimikos and Rythmos tests failing on many ATDM builds


fryeguy52:
CC: @trilinos/stratimikos, @srajama1 (Trilinos Linear Solvers Product Lead), @rppawlo (Trilinos Nonlinear Solvers Product Lead), @bartlettroscoe, @fryeguy52
 
<Checklist>
<???: Add label "ATDM">
<???: Add label "bug"?>
<???: Add label for affected packages (e.g. "MueLu", "Tpetra", "Kokkos", etc.)>
<???: Add milestone "Initial cleanup of new ATDM builds of Trilinos" or "Keep promoted ATDM builds of Trilinos clean">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
 
## Next Action Status
<status-and-or-first-action>
 
## Description
 
As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2019-03-20&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=5&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=ATDM&field2=status&compare2=61&value2=Failed&field3=testname&compare3=62&value3=Ifpack2_BlockTriDiContainerUnitAndPerfTests_MPI_4&field4=buildstarttime&compare4=83&value4=2019-03-20&field5=buildstarttime&compare5=84&value5=2019-03-21) the tests:
* Stratimikos_test_single_stratimikos_solver_driver_belos_np_MPI_1
* Stratimikos_test_single_stratimikos_solver_driver_belos_ml_MPI_1
* Stratimikos_test_single_stratimikos_solver_driver_belos_ifpack_MPI_1
* Rythmos_timeDiscretizedBackwardEuler_amesos_MPI_1

are failing in many ATDM builds.

[new commits when these started failing](https://testing.sandia.gov/cdash/viewNotes.php?buildid=4754139#!#note4)

## Current Status on CDash

[...]

PR #4684, opened 2019-03-20
Revert "Belos Stratimikos adapter: Use solver factory"

bartlettroscoe:
CC: @cgcgcg , @trilinos/framework 

Reverts trilinos/Trilinos#4606

Reverting for now as per https://github.com/trilinos/Trilinos/pull/4677#issuecomment-475018795.  This can then be carefully fixed offline to restore correct behavior for default tolerances.

This should fix the Trilinos PR builds and a bunch of ATDM Triilnos builds (#4678).


bartlettroscoe:
@trilinos/framework 

Can you kill all of the doomed PR builds and let this one run to fix the PR builds for changes upstream from Stratimikos?

PR #4695, opened 2019-03-21
Ifpack2 BlockTriDiContainer - various improvements on the line solver.

kyungjoo-kim:
## Description

This PR includes improvements on the Ifpack2 BlockTriDiContainer. Previously, 
- the block tridiagonal solve is separated from conversion of multivector between packed and flat formats. 
- the conversion routine is fused with norm calculation. 
- this inefficiently use team parallelism.

Now,
- the block tridiagonal solve is fused with multivector conversion routine.
- norm caculation is done on a tpetra multi vector with a range policy.  

## Motivation and Context

This solves slowness in computing norms for stopping the fixed pointer iterations. 

## How Has This Been Tested?

Kokkos-dev-2.


packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1829,77 +1834,20 @@ namespace Ifpack2 {
...
-
-      KOKKOS_INLINE_FUNCTION
-      void 
-      join(volatile magnitude_type *dst, const volatile magnitude_type *src) const {
-        for (int i=0;i<value_count;++i) 
-          dst[i] += src[i];
-      }          
-
       MultiVectorConverter() = delete;
       MultiVectorConverter(const MultiVectorConverter &b) = default;
```
mhoemmen: Is there any particular reason you `default` the copy constructor, but not the assignment operator?  If you don't actually need this, it's better to leave it off, since it suppresses implicit declaration of a move constructor and move assignment operator.

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-five

If the default operations are fine here, you could just go with the Rule of Zero: 

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-zero

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2442,17 +2422,22 @@ namespace Ifpack2 {
...
 	  (Kokkos::ThreadVectorRange(member, vector_loop_size),[&](const int &v) {
 	    solveMultiVector(member, blocksize, i0, r0, nrows, v, WW);
+	    copyToFlatMultiVector(member, partidx, npacks, pri0, v, blocksize, num_vectors);
 	  });
       }      
 
-      void run() {
+      void run(const impl_scalar_type_2d_view &Y,
+	       const impl_scalar_type_2d_view &Z) {
 #if defined(KOKKOS_ENABLE_CUDA) && defined(IFPACK2_BLOCKTRIDICONTAINER_ENABLE_PROFILE)
```
mhoemmen: It would be nice if we could combine all these different profiling and timing interface calls into a single function that doesn't have macros wrapped all around it.  How about this:
```c++
namespace Ifpack2 {
namespace Details {
  void startProfiling (const char label[]);
  void stopProfiling ();

  class ProfilingScope {
  public:
    ProfilingScope (const char label[]) { startProfiling (label); }
    ~ProfilingScope () { stopProfiling (); }
  };
}
}
```
The implementation of `startProfiling` could look like this:
```
void startProfiling (const char label[]) {
#ifdef WHATEVER_CUDA_STUFF
  cudaProfilerStart ();
#endif 
#ifdef WHATEVER_WE_LOVE_MACROS_I_GUESS
  TEUCHOS_FUNC_TIME_MONITOR(label);
#endif 
}
```

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3210,6 +3213,75 @@ namespace Ifpack2 {
       }
     }; 
 
+
+    template<typename MatrixType>
+    struct ReduceMultiVector {
```
mhoemmen: I thought the point was to fuse the norm with another kernel -- it looks like this is just a norm by itself.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1771,9 +1770,15 @@ namespace Ifpack2 {
 
 	Kokkos::TeamPolicy<execution_space,ExtractAndFactorizeTag>
 	  policy(packptr.extent(0)-1, team_size, vector_loop_size); 
+#if defined(KOKKOS_ENABLE_DEPRECATED_CODE)
 	Kokkos::parallel_for("ExtractAndFactorize::TeamPolicy::run<ExtractAndFactorizeTag>", 
 			     policy.set_scratch_size(0,Kokkos::PerTeam(per_team_scratch)), *this);
-	
+#else
+        policy.set_scratch_size(0,Kokkos::PerTeam(per_team_scratch));
```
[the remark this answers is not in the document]
kyungjoo-kim: No it does not work. With ``set_scratch_size`` with deprecated code on, this returns an object with a scratch space and it does not change itself. Basically I do not really like this change and rather wait until kokkos propose a better solution as described in kokkos issue 2037.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2477,7 +2463,24 @@ namespace Ifpack2 {
 	    ("SolveTridiags::TeamPolicy::run<MultiVector>",		\
 	     policy.set_scratch_size(0,Kokkos::PerTeam(per_team_scratch)), *this); \
 	} break
-	
+#else	
+#define BLOCKTRIDICONTAINER_DETAILS_SOLVETRIDIAGS(B)			\
```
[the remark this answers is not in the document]
kyungjoo-kim: Ironically, I previously changed into a function and runtime increases 2x #4549. It has to be a macro and I do not know a better solution for this and it is smart not to spend time for this when a solution works.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3210,6 +3213,75 @@ namespace Ifpack2 {
...
+	this->z = zz;
+
+	const local_ordinal_type nrows = zz.extent(0);
+	const local_ordinal_type ncols = zz.extent(1);
+
+	for (local_ordinal_type j=0;j<ncols;++j) {
+	  this->colidx = j;
+	  impl_scalar_type reduced_value(0);
+	  Kokkos::RangePolicy<execution_space,SingleVectorTag> policy(0, nrows);
+	  Kokkos::parallel_reduce("ReduceMultiVector::SingleVetor",
```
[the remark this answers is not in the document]
kyungjoo-kim: will fix this.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3210,6 +3213,75 @@ namespace Ifpack2 {
...
+				  policy, *this, reduced_value);
+	  /// I don't know why we compute norms element-wisely...
+	  /// SPARC needs to change the interface (compute a norm per rhs)
```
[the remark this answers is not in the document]
kyungjoo-kim: You are right. I can just call norm function. Good point!! My comment of this is actually about the previous version of computing norms. It computes norms element-wisely in a small block vector. will fix it.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3210,6 +3213,75 @@ namespace Ifpack2 {
...
+      {}
+      
+      struct SingleVectorTag {};
+      
+      KOKKOS_INLINE_FUNCTION
+      void
+      operator()(const SingleVectorTag &, 
+		 const local_ordinal_type &i, 
+		 impl_scalar_type &update) const {
+	update += z(i,colidx);
```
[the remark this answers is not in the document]
kyungjoo-kim: the vector is already computed as abs values. abs(diff) is computed with the solve phase.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_def.hpp
```diff
@@ -115,6 +115,7 @@ namespace Ifpack2 {
     impl_->overlap_communication_and_computation = overlapCommAndComp;
     
     impl_->Z = typename impl_type::tpetra_multivector_type();
+    impl_->W = typename impl_type::tpetra_multivector_type();
```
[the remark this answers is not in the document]
kyungjoo-kim: this function is an internal function to initialize variables. Your point is right but I did this  because I have an experience that the default constructor in some compilers does not really initialize its member variables. It is just a safe way to do it myself and does not harm performance.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1829,77 +1834,20 @@ namespace Ifpack2 {
       // packed multivector output (or input)
       Unmanaged<vector_type_3d_view> packed_multivector;
       Unmanaged<impl_scalar_type_2d_view> scalar_multivector;
-      impl_scalar_type damping_factor;
 
       template<typename TagType>
       KOKKOS_INLINE_FUNCTION
       void copy_multivectors(const local_ordinal_type &j,
```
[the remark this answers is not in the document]
kyungjoo-kim: Using reference for int is not my style but I do not want to add unnecessary modification. This code is getting picky. Some minor changes (I thought) sometimes cause slow performance. When such changes are mixed with other major changes (I thought), it is kind of difficult to figure out to recover the performance. An example is changing a macro into a template function causes slowness. As the commit also include another major changes, I have spend another day to first look at the major changes. As you know, the increasing Trilinos compiling time is annoying when you need to debug with trials and errors. So, when it does not bother me much, I don't do cosmetic changes.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: Good point. I just delete it because the class always need part interface. Probably when I create the class, I just want to prevent accidental mistake myself. I will just delete the deleted constructor. I will fix.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: Good idea. I will fix this way.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: My comment may confuse you. Previously the code fuse norm computation and multi vector conversion. In this commit, I first compute diff of values in a multi vector with a conversion. Then, I reduce it. So, here I can say that I fuse solve and diff computation. This results in better usage of team threads.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -2185,11 +2080,91 @@ namespace Ifpack2 {
...
+		      for (local_ordinal_type col=0;col<num_vectors;++col) {
+			impl_scalar_type &y = Y_scalar_multivector(row,col);
+			const impl_scalar_type yd = X_internal_vector_values(pri, i, col, v)[vi] - y;
+			y  += df*yd;
+			
+			if (compute_diff) {
+			  impl_scalar_type &z = Z_scalar_multivector(row,col);
+			  const magnitude_type abs_yd = Kokkos::ArithTraits<impl_scalar_type>::abs(yd); 
+			  z = abs_yd*abs_yd;
+			}
```
kyungjoo-kim: @mhoemmen The norm computation is right. here I compute square of diff value.

kyungjoo-kim:
@mhoemmen Thanks. You pointed several good points.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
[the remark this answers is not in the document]
mhoemmen: Cool, thanks for explaining!

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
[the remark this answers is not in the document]
mhoemmen: ok

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: (FYI the idea is that the implementation of these functions would go in a .cpp file, so that we wouldn't have to recompile Ifpack2's solvers if we change what Ifpack2 does for profiling.)

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: ok

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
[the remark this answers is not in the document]
mhoemmen: ok

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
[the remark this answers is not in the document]
mhoemmen: ok

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_def.hpp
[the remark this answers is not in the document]
mhoemmen: ok

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: BTW, teuchos time monitor cannot be inside of the function. The macro put an object to get the scope of timer. If we put the timer with the function, it only measure time for the startProfiling function.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: ok

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: @kyungjoo-kim We could use the `Teuchos::Time` object directly:
```c++
namespace Ifpack2 {
namespace Details {

class ProfilingScope {
public:
  ProfilingScope ();
  ~ProfilingScope ();
private:
#ifdef WHATEVER_WE_LOVE_MACROS_I_GUESS
  Teuchos::RCP<Teuchos::Time> timer_;
#endif 
};

void ProfilingScope::startProfiling () {
  // ...
#ifdef WHATEVER_WE_LOVE_MACROS_I_GUESS
  // If the timer exists, return it, else make a new one.
  timer_ = Teuchos::TimeMonitor::getNewCounter (label);
  timer->start ();
#endif 
}

void ProfilingScope::startProfiling () {
  // ...
#ifdef WHATEVER_WE_LOVE_MACROS_I_GUESS
  timer_->stop ();
  timer_ = Teuchos::null;
#endif 
}
```

kyungjoo-kim: when we define timer in this way, Is this also included in the summarize report ? and stacked timer ?

mhoemmen: @kyungjoo-kim The macro also calls `Teuchos::TimeMonitor::getNewCounter` (look in `teuchos/comm/src/Teuchos_TimeMonitor.hpp` for the macro definition).  If that works with stacked timer, then the above code should be fine.

kyungjoo-kim: This teuchos timer does not accumulate time over multiple iterations.

kyungjoo-kim:
@mhoemmen the profiler turns out not easy. At least I replace multiple macro ifdef guards into a single macro.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -84,6 +85,11 @@
 #include "cuda_profiler_api.h"
 #endif
 
+// I am not 100% sure about the mpi 3 on cuda
```
mhoemmen: @kyungjoo-kim Are you asking whether MPI supports nonblocking collectives with CUDA buffers?  It should work but it might be slow.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3208,31 +3207,29 @@ namespace Ifpack2 {
...
+	// this is expensive when I use cublas with uvm memory...
+	// no clue yet.
+	const auto norm2 = KokkosBlas::nrm2(rz);
+#else
+        const Kokkos::RangePolicy<execution_space> policy(0,rz.extent(0));
+	magnitude_type norm2;
+	Kokkos::parallel_reduce
+	  (" ReduceMultiVector",
+	   policy, KOKKOS_LAMBDA(const local_ordinal_type &i, magnitude_type &update) {
+	    const magnitude_type val = Kokkos::ArithTraits<impl_scalar_type>::abs(rz(i));
```
mhoemmen: @kyungjoo-kim Did you mean to change to a two-norm here?

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3208,31 +3207,29 @@ namespace Ifpack2 {
...
-	    const auto norm = rv_host(j)/magnitude_type(blocksize);
-	    for (local_ordinal_type i=0;i<blocksize;++i)
-	      vals[j*blocksize+i] = norm;
-	  }
-	}
+	impl_scalar_type_1d_view rz(zz.data(), nrows*ncols);
+#if 0
+	// this is expensive when I use cublas with uvm memory...
+	// no clue yet.
+	const auto norm2 = KokkosBlas::nrm2(rz);
```
mhoemmen: Given that your hand-rolled kernel is faster and that `rz` has rank 1, two possibilities come to mind:

  1. `KokkosBlas::nrm2` is calling cuBLAS, and cuBLAS is slow.
  2. `KokkosBlas::nrm2` is not calling cuBLAS, and the `final` kernel that takes the square root (on device!) may be slow.  See `V_Nrm2_Functor`, in `kokkos-kernels/src/blas/impl/KokkosBlas1_nrm2_impl.hpp`.  Just passing `take_sqrt=false` into `KokkosBlas::nrm2` won't help, since the `final` kernel will always run (`take_sqrt` is a run-time `bool` in the kernel).

jjellio: Are you sure nrm2 is actually calling cuBLAS?

I've been looking at the CuBLAS/GEMM thing - (trying to figure out exactly what is happening on Cuda...)

GEMM (with y'all patches!) looks great: 4-5x improvement in total time!
```

        |   |       |   |   |   |   |-> 6.77e+01 sec 5.5% 5.2% 0.3% 28.5% 47990 Belos::MVT::MvTransMv [region]
        |   |       |   |   |   |   |   |-> 3.65e+01 sec 3.0% 0.0% 2.9% 5.2% 47990 Tpetra::MV::multiply [region]
        |   |       |   |   |   |   |   |   |-> 2.42e+01 sec 2.0% 0.0% 1.7% 4.5% 47990 Tpetra::MV::multiply-call-gemm [region]
        |   |       |   |   |   |   |   |   |   |-> 2.31e+01 sec 1.9% 0.0% 1.7% 100.0% 47990 KokkosBlas::gemm[TPL_BLAS,double] [region]
        |   |       |   |   |   |   |   |   |-> 1.04e+01 sec 0.8% 0.0% 13.0% 100.0% 47990 Tpetra::MV::reduce [region]
```

Notice GEMM is marked TPL_BLAS... norm2 is not.
```
 |   |   |   |   |   |-> 1.18e+01 sec 1.0% 32.5% 11.2% 58.1% 47990 Tpetra::MV::norm2 (Teuchos::ArrayView) [region]
        |   |       |   |   |   |   |       |-> 4.44e+00 sec 0.4% 75.0% 5.4% 25.0% 47990 KokkosBlas::nrm2[ETI] [region]
        |   |       |   |   |   |   |       |   |-> 3.33e+00 sec 0.3% 100.0% 6.5% ------ 47990 KokkosBlas::Nrm2::S0 [reduce]
```

kyungjoo-kim: For my own testing, I changed kokkoskernels to eti uvm space. Without cublas, it shows identical performance and with cublas, it is slower. If it is faster, I would include the modification on the kokkoskernels in this PR. As it is slower in my test, I did not include it to prevent any side effect of other slowness it might cause.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3208,31 +3207,29 @@ namespace Ifpack2 {
...
+        const Kokkos::RangePolicy<execution_space> policy(0,rz.extent(0));
+	magnitude_type norm2;
+	Kokkos::parallel_reduce
+	  (" ReduceMultiVector",
+	   policy, KOKKOS_LAMBDA(const local_ordinal_type &i, magnitude_type &update) {
+	    const magnitude_type val = Kokkos::ArithTraits<impl_scalar_type>::abs(rz(i));
+	    update += val*val;
+	  }, norm2);
+#endif	
+	const auto norm = norm2*norm2/magnitude_type(ncols);
```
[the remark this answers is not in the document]
kyungjoo-kim: you're right. need to fix this.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3294,18 +3288,22 @@ namespace Ifpack2 {
...
+	auto send_data = work_.data() + n;
+	auto recv_data = work_.data();
         if (collective_) {
-          std::copy(work_.begin(), work_.begin() + n, work_.begin() + n);
+          std::copy(recv_data, send_data, send_data);
 #ifdef HAVE_IFPACK2_MPI
-# if MPI_VERSION >= 3
-          MPI_Iallreduce(work_.data() + n, work_.data(), n,
+# if defined(IFPACK2_BLOCKTRIDICONTAINER_USE_MPI_3) 
+          MPI_Iallreduce(send_data, recv_data, n,
```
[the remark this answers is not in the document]
kyungjoo-kim: My worry is not about the functionality. This reduce for norm is done on host. The buffer lives in the host. I was wondering if mpi can cause the code is hanging in somewhere I did not put a timer. I need to find out where 37% missing time in applyInverseJacobi.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3260,7 +3257,6 @@ namespace Ifpack2 {
       NormManager(const NormManager &b) = default;
```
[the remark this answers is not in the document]
kyungjoo-kim: I put the default from my experience a few years ago. At that time, the compiler generated default constructors do not really initialize the member values. I hope that those problems are gone and this PR tests may filter it out if it still exist. I am going to remove all default constructors.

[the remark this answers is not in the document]
kyungjoo-kim: Okay. It does not work without explicit expression of the default constructor.
```
[ 93%] Building CXX object packages/ifpack2/src/CMakeFiles/ifpack2.dir/Ifpack2_BlockTriDiContainer_Cuda.cpp.o
/ascldap/users/kyukim/Work/lib/trilinos/kyukim/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_def.hpp(89): error: the default constructor of "Ifpack2::BlockTriDiContainerDetails::ImplObject<Tpetra::RowMatrix<double, int, int, Kokkos_Compat_KokkosCudaWrapperNode>>" cannot be referenced -- it is a deleted function
          detected during instantiation of "void Ifpack2::BlockTriDiContainer<MatrixType, Ifpack2::BlockTriDiContainerDetails::ImplSimdTag>::initInternal(const Teuchos::RCP<const Ifpack2::BlockTriDiContainer<MatrixType, Ifpack2::BlockTriDiContainerDetails::ImplSimdTag>::row_matrix_type> &, const Teuchos::Array<Teuchos::Array<Ifpack2::BlockTriDiContainer<MatrixType, Ifpack2::BlockTriDiContainerDetails::ImplSimdTag>::local_ordinal_type>> &, const Teuchos::RCP<const Ifpack2::BlockTriDiContainer<MatrixType, Ifpack2::BlockTriDiContainerDetails::ImplSimdTag>::import_type> &, __nv_bool, __nv_bool) [with MatrixType=Tpetra::RowMatrix<double, int, int, Kokkos_Compat_KokkosCudaWrapperNode>]" 
(162): here

/ascldap/users/kyukim/Work/lib/trilinos/kyukim/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_def.hpp(147): error: no instance of constructor "Ifpack2::BlockTriDiContainerDetails::NormManager<MatrixType>::NormManager [with MatrixType=Tpetra::RowMatrix<double, int, int, Kokkos_Compat_KokkosCudaWrapperNode>]" matches the argument list
          detected during instantiation of "void Ifpack2::BlockTriDiContainer<MatrixType, Ifpack2::BlockTriDiContainerDetails::ImplSimdTag>::clearInternal() [with MatrixType=Tpetra::RowMatrix<double, int, int, Kokkos_Compat_KokkosCudaWrapperNode>]" 
(247): here
```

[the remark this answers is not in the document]
mhoemmen: The compiler might delete an implicitly declared default constructor, under the conditions listed here:

https://en.cppreference.com/w/cpp/language/default_constructor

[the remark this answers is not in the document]
kyungjoo-kim: Yeah.... it is possible that the compiler delete it but let's keep this way using explicit default constructors.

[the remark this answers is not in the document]
mhoemmen: ok

kyungjoo-kim:
This PR is done and next improvement will be included in a new PR.

kyungjoo-kim:
Anyone who want approve this ?

srajama1:
Looking at it now

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3204,38 +3235,60 @@ namespace Ifpack2 {
...
+#if 0
+      const auto norm2 = KokkosBlas::nrm1(zz);
+#else
+      impl_scalar_type norm2(0);
+      Kokkos::parallel_reduce
+        ("ReduceMultiVector::Device",
+         Kokkos::RangePolicy<typename impl_type::execution_space>(0,zz.extent(0)),
+         KOKKOS_LAMBDA(const local_ordinal_type &i, impl_scalar_type &update) {
+          update += zz(i);
+        }, norm2);
```
mhoemmen: Is this supposed to be the norm or the square of the norm?

kyungjoo-kim: it is supposed to be square of the norm. sqrt is applied after mpi all reduce.

srajama1: @kyungjoo-kim What is the difference between the above if and else part ? Why is this here instead of in KokkosBlas ?

srajama1: Also we should call cuSparse through Kokkos Blas, when available ...

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3291,76 +3336,44 @@ namespace Ifpack2 {
...
+        if (sweep <= 0) return false;	
+	
+	IFPACK2_BLOCKTRIDICONTAINER_TIMER("BlockTriDi::NormManager::CheckDone");
+	
         TEUCHOS_ASSERT(sweep >= 1);
         if ( ! force && (sweep - 1) % sweep_step_) return false;
         if (collective_) {
 #ifdef HAVE_IFPACK2_MPI
-# if MPI_VERSION >= 3
+# if defined(IFPACK2_BLOCKTRIDICONTAINER_USE_MPI_3)
```
srajama1: Is there a global flag for MPI_3 that we can use ? Defining one for each solver will make our already unmanageable cmake scripts worse

mhoemmen: It would be better to use `Tpetra_Details_iallreduce.hpp` here; it has a fall-back path (that's still nonblocking) for the `MPI_VERSION < 3` case.

We don't actually test Trilinos with OpenMPI < 1.10, so perhaps we don't have to worry about this until someone complains.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: When we do not use CUBLAS, it is not different. When CUBLAS is actually used, it makes difference (in this case, cublas is slower). To prevent slowness, I just put this reduction with kokkos. Once we identify the perf problem on the cublas, we can replace it with that.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: this is not a cmake flag. the macro is just replacement of MPI_VERSION >= 3. see the beginning of the code. i do this way because I want to separate the effect of mpi 3 under my control. previously, it just use mpi 3 as it uses mpi macro.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
srajama1: Please file an issue so we don't forget this ..

mhoemmen:
I don't want to be any more obstructionist than the PR testing was last week ;-)

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
srajama1: Strange, but ok ...

PR #4699, opened 2019-03-22
Tpetra: Can we remove the Experimental Guards for the FE Tests?

csiefer2:
I don't know, but the PR tester sure does!


csiefer2:
Failing with nothing posted to the testing server.  _Excellent_

jjellio:
I doubt anything can go through until PR#4685 is manually pushed

csiefer2:
We just keep on winning the autotester lottery... @trilinos/framework

csiefer2:
The Intel fails are addressed by the other PR.  Let this age and then I'll update the branch when that pushes

csiefer2:
Updated branch from other PR.  Let's see what fails tonight @tjfulle

mhoemmen:
Build error is interesting...
```
error: cannot convert ‘Teuchos::ScalarTraits<Kokkos::complex<double> >::magnitudeType {aka Kokkos::complex<double>}’ to ‘const double’ in initialization
     const ScalarMag err = relErr( a1[i], a2[i] );
                     ^~~
```
We don't actually have a specialization of `Teuchos::ScalarTraits` for `Kokkos::complex`.  You might want to do the absolute value of the difference yourself, rather than having some Teuchos macro do it.

csiefer2:
@mhoemmen  So the TEST_COMPARE_FLOATING_ARRAYS macro doesn't work for complex.

Lovely.

csiefer2:
Once more unto the breach, dear friends, once more...

packages/tpetra/core/test/FECrsGraph/FECrsGraph_UnitTests.cpp
```diff
@@ -338,11 +338,17 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL( FECrsGraph, Assemble1D_LocalIndex, LO, GO, No
 // INSTANTIATIONS
 //
 
+#ifdef HAVE_TPETRA_EXPERIMENTAL
```
mhoemmen: If experimental is disabled, you might get a lot of "unused classes" warnings in this .cpp file.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -98,7 +98,10 @@ bool compare_final_matrix_structure_impl(Teuchos::FancyOStream &out,Tpetra::CrsM
   TEST_COMPARE_ARRAYS(colind1,colind2);
   if (!success) {out<<"Compare: colind match failed"<<endl;return false;}
 
-  TEST_COMPARE_FLOATING_ARRAYS(values1,values2,tol);
+  // This is necessary to make sure that complex works (since Teuchos::ScalarTraits does not have a Kokkos::complex specialization)
+  auto values1_av = Teuchos::av_reinterpret_cast<Scalar>(Kokkos::Compat::getArrayView(values1));
```
mhoemmen: In the future, try this:
```c++
auto values1_h = Kokkos::create_mirror_view (values1);
auto values1_av = Teuchos::av_reinterpret_cast<Scalar> (Kokkos::Compat::getArrayView (values1_h));
```

packages/tpetra/core/test/FECrsGraph/FECrsGraph_UnitTests.cpp
csiefer2: Good catch!

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
csiefer2: Fixed

csiefer2:
MueLu link fail :(

mhoemmen:
Looks like a legit CUDA test failure in the FECrsGraph test:
```
3. FECrsGraph_int_int_Kokkos_Compat_KokkosCudaWrapperNode_Assemble1D_LocalIndex_UnitTest ... terminate called after throwing an instance of 'std::runtime_error'
  what():  cudaDeviceSynchronize() error( cudaErrorIllegalAddress): an illegal memory access was encountered /home/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_cuda_9.2/Trilinos/packages/kokkos/core/src/Cuda/Kokkos_Cuda_Impl.cpp:119

[ride12:51600] *** Process received signal ***
[ride12:51600] Signal: Aborted (6)
```

csiefer2:
@mhoemmen Legit Cuda error that debuggers are utterly unhelpful with.  Calling in @dsunder for a consult this afternoon...

mhoemmen:
I'll call this a WIP -- no worries, we'll figure it out :-)

csiefer2:
@dsunder was very helping in getting the debugging started.

Segfaulting in Kokkos::deep_copy().... I suspect this is an issue of garbage in / segfault out.

csiefer2:
See #4792

csiefer2:
Failure to compile ROL?  Retest!

csiefer2:
@trilinos/framework GCC 7.2.0 PR tester fails.

william76:
@csiefer2 reducing the PR concurrency should help with this.  The issue here is that when the system is getting oversubscribed, we can end up with multiple PR's getting unlucky and having 4 of their 5 jobs get slots to run on Jenkins with the 5th job stalling in the queue waiting for available executors and timing out.

I want to say that you don't need to add the RETEST flag... buuuut I'm not 100% sure there.  @jwillenbring to verify the behavior, when these sorts of test failures happen the autotester will restart the test a bit later without requiring the AT: RETEST flag, right?

PR #4701, opened 2019-03-22
Tpetra: Deprecate DynamicProfile

GeoffDanielson:
@trilinos/tpetra 

## Description
This uses the TPETRA_ENABLE_DEPRECATED_CODE macro to switch the default constructors in CrsGraph and CrsMatrix to use StaticProfile as default, and warn of profileTypes upcoming deprecation

## Motivation and Context
This brings us one step closer to static allocation for GPU usage.

## How Has This Been Tested?
This was tested by building the tpetra package and building its tests both with and without the TPETRA_ENABLE_DEPRECATED_CODE macro enabled.  

The non-GCC guarded portions (for Intel compilers) remains untested, because I don't have a machine with an intel compiler new enough.  However, GCC works on the pragma messages alone (I called out the pragma warning specific to the GCC compiler because warnings are more appropriate to the issue than simple messages).  It may be that the Intel compiler will support pragma warnings as well, but I figured I would stick with the universally supported macros.

## Checklist

## Additional Information
This introduces new compiler warnings, on purpose (the entire point was to introduce compiler warnings).


packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -339,7 +339,11 @@ namespace Tpetra {
     ///   default values.
     CrsGraph (const Teuchos::RCP<const map_type>& rowMap,
               const size_t maxNumEntriesPerRow,
-              const ProfileType pftype = DynamicProfile,
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
+	      ProfileType pftype = DynamicProfile,
```
mhoemmen: Could we factor out the default value and put it in a common place, ideally where the `ProfileType` enum gets defined, in `Tpetra_ConfigDefs.hpp`?  That would avoid possible typos with macros, or people adding constructors and forgetting to use the macros.  Also, now is our chance to make these `const`.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -251,6 +251,20 @@ namespace Tpetra {
     , haveGlobalConstants_ (false)
     , sortGhostsAssociatedWithEachProcessor_ (true)
   {
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
+      #if defined (__GNUC__)
+        #pragma GCC warning "Default profile with deprecated code is DynamicProfile, pftype will soon be unsupported."
```
mhoemmen: Please don't do this; this will break Sierra's build.  (It won't pass PR tests anyway, since we have a warnings as errors build.)  The right way to do this is to mark the `DynamicProfile` value as deprecated, using the `TPETRA_DEPRECATED` macro.  Check out this example:

https://godbolt.org/z/oJ_ZWk
```c++
#include <iostream>
// Don't actually do this -- Tpetra does it for you.
#define TPETRA_DEPRECATED __attribute ((deprecated))

enum EFoo {
    FOO_ZERO,
    FOO_ONE,
    // This is how you mark an enum value as deprecated.
    FOO_TWO TPETRA_DEPRECATED
};

int main() {
    // This line of code emits a deprecation warning.
    std::cout << FOO_TWO << std::endl;
    return 0;
}
```
Note that if we deprecate `DynamicProfile`, we'll need to change the rest of Tpetra so that it only ever uses the `StaticProfile` enum value.  (Instead of testing `pftype == DynamicProfile`, test `pftype == StaticProfile`.)

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
GeoffDanielson: Yep, we can do all those things.

mhoemmen:
@GeoffDanielson Thanks for working on this!  I'm taking a look now.

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
```diff
@@ -129,9 +129,19 @@ namespace Tpetra {
   /*! Allocation profile for matrix/graph entries */
   enum ProfileType {
     StaticProfile,  /*!< Single, static allocation (strict and more efficient) */
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
     DynamicProfile  /*!< Multiple, dynamic allocations (flexibile, but less efficient) */
```
mhoemmen: Don't we still want to mark the thing as deprecated, even if it's enabled?  Users have a way to disable deprecation warnings per package.  Furthermore, "enable deprecated code" implies that if it's _not_ enabled, then the code doesn't even exist.
```c++

enum ProfileType {
  StaticProfile
  , DynamicProfile TPETRA_DEPRECATED
#endif 
};
```

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
```diff
@@ -129,9 +129,19 @@ namespace Tpetra {
...
+#else
+    DynamicProfile TPETRA_DEPRECATED
+#endif
   };
 
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
mhoemmen: Do we want to expose this to users?  I'm thinking no -- users either get the default, or they specify the value explicitly.  If we don't want to expose this to users, then we should make this a constant in the `Tpetra::Details` namespace, not a macro.  If we _do_ want to expose it to users, then we should make it a constant in the `Tpetra` namespace, not a macro.

In general, prefer constants and other code things to macro / preprocessor things.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -251,6 +251,12 @@ namespace Tpetra {
     , haveGlobalConstants_ (false)
     , sortGhostsAssociatedWithEachProcessor_ (true)
   {
+#ifndef TPETRA_ENABLE_DEPRECATED_CODE
+#if (pftype == DynamicProfile)
```
mhoemmen: See above.  The enum value shouldn't even exist if `TPETRA_ENABLE_DEPRECATED_CODE` is not defined.

packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
```diff
@@ -818,9 +818,6 @@ namespace Tpetra {
 
         // Construct the CrsMatrix, using the row map, with the
         // constructor specifying the number of nonzeros for each row.
-        // Create with DynamicProfile, so that the fillComplete() can
-        // do first-touch reallocation (a NUMA (Non-Uniform Memory
-        // Access) optimization on multicore CPUs).
         RCP<sparse_matrix_type> A =
           rcp (new sparse_matrix_type (pRowMap, myNumEntriesPerRow (),
                                        DynamicProfile));
```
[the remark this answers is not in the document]
GeoffDanielson: I was kind of hoping that @tjfulle might have more insight on this one, since he did the lion's share of work on the matrix market stuff.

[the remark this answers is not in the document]
GeoffDanielson: Let me elaborate: MatrixMatrix tests addition, but the test is built in such a way that the addition of the two matrices requires one to be dynamically allocated, and I'm not sure how we want to test that situation.  It should fail, it does fail, but the test expects it to succeed.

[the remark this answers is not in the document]
mhoemmen: @GeoffDanielson We may need to fix addition in the same way that @tjfulle fixed Export / Import to a StaticProfile matrix.

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
GeoffDanielson: You're right, this code doesn't even do anything.  I had meant to cull it previously, but thanks for catching it.

packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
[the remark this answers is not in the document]
tjfulle: What is actually failing?  For the matrix market conversion to `StaticProfile`, I just counted number of entries and changed the profile type to static (using the  number of entries previously determined).  How does that relate to matrix-matrix add?

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
GeoffDanielson: TPETRA_DEPRECATED doesn't work on a member of an enum, unfortunately.

rppawlo: wonder if c++14 [deprecated] would work? Wish we could move trilinos to c++14...

packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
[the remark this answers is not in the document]
GeoffDanielson: @tjfulle the test that tickles it is in the MatrixMatrix unit tests.  When adding to a matrix that's created by readSparseFile, if the read matrix does not have all the correct non-zero entries allocated, the op will fail.

I am working on a workaround, maybe allocating a third matrix with the max numEntriesPerRow picked from A and B with B's entries copied in.

Let me know if that sounds horrible, I guess.

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
mhoemmen: @GeoffDanielson wrote:
> `TPETRA_DEPRECATED` doesn't work on a member of an enum, unfortunately.

What does the compiler say?

packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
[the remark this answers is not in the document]
GeoffDanielson: So, it was a little more involved than that, but I think I did the necessary.  Please look at the unit test for me (MatrixMatrix) @tjfulle and @mhoemmen and rain down verbal abuse.

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
GeoffDanielson: In file included from /home/gcdanie/Trilinos/packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp:45:0,
                 from /home/gcdanie/Trilinos/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp:45,
                 from /home/gcdanie/Trilinos/packages/tpetra/core/src/Epetra_TsqrMessenger.hpp:60,
                 from /home/gcdanie/Trilinos/packages/tpetra/core/src/Epetra_TsqrMessenger.cpp:57:
/home/gcdanie/Trilinos/Obj_vesper/packages/tpetra/core/src/TpetraCore_config.h:188:32: error: expected ‘}’ before ‘__attribute__’
 #    define TPETRA_DEPRECATED  __attribute__((__deprecated__))
                                ^
/home/gcdanie/Trilinos/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp:133:20: note: in expansion of macro ‘TPETRA_DEPRECATED’
     DynamicProfile TPETRA_DEPRECATED /*!< Multiple, dynamic allocations (flexibile, but less efficient) */

GeoffDanielson: Ugh, there's a way to do it, but it's grosssss.

```
  /*! Allocation profile for matrix/graph entries */
  enum ProfileType {
    StaticProfile
  };

  TPETRA_DEPRECATED const ProfileType DynamicProfile = ProfileType(StaticProfile+1);
#define TPETRA_DEFAULT_PROFILE_TYPE DynamicProfile
#else
#define TPETRA_DEFAULT_PROFILE_TYPE StaticProfile
#endif
```

mhoemmen: @GeoffDanielson What compiler version is that?

GeoffDanielson: GCC 4.9.3:

```
 gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/projects/sems/install/rhel6-x86_64/sems/compiler/gcc/4.9.3/base/libexec/gcc/x86_64-unknown-linux-gnu/4.9.3/lto-wrapper
Target: x86_64-unknown-linux-gnu
Configured with: ./configure --prefix=/projects/sems/install/rhel6-x86_64/sems/compiler/gcc/4.9.3/base --disable-multilib
Thread model: posix
gcc version 4.9.3 (GCC) 
```

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -1019,20 +1020,37 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(Tpetra_MatMat, operations_test,SC,LO, GO, NT)
...
+      n += A->getGlobalMaxNumRowEntries();
+
+      newOut << "construct row map" << endl;
+      RCP<const map_type> rm = B->getRowMap();
+      newOut << "construct new Sparse E" << endl;
+      RCP<Matrix_t> E = rcp (new Matrix_t(rm, n, Tpetra::StaticProfile));
+      newOut << "get one" << endl;
+      auto one = Teuchos::ScalarTraits<SC>::one();
+      newOut << "add B to new sparse matrix" << endl;
+      Tpetra::MatrixMatrix::Add(*B, BT, one, *E, one);
```
mhoemmen: We might actually need to fix Add to do the right thing here.

GeoffDanielson: I don't see a way for Add to work on non-DynamicProfiles without something like this, though maybe there's a more elegant way to do it inside MatrixMatrix than in the unit test.

Honestly, I'm surprised that more didn't break within Tpetra that got caught by the testing.

Suggestions on an approach to fix?

mhoemmen: The logical thing to do is to make Add do the same thing that Export / Import to a StaticProfile CrsGraph does (thanks to @tjfulle ), namely reallocate if needed.

GeoffDanielson: In this case "do the right thing" according to the docs is to make B big enough on input to the += to allow it to hold the whole input.

I'll look into this on Friday to see how to do that.

GeoffDanielson: So, it appears that I've made B big enough on input to the += to allow it to hold the whole input.

This way, we keep the test, and show a possible way to work around the issue.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -1076,20 +1075,33 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(Tpetra_MatMat, operations_test,SC,LO, GO, NT)
...
 
-      B = Reader<Matrix_t >::readSparseFile(B_file, comm, false);
+      B = Reader<Matrix_t >::readSparseFile(B_file, comm, true, false, false);
+      //declare E with enough entries to receive A + B
+      size_t n = 0;
+      for (size_t i = 0; i < B->getNodeNumRows(); i++) {
+	if (n < B->getNumEntriesInLocalRow(i))
+	  n = B->getNumEntriesInLocalRow(i);
+      }
+      n += A->getNodeMaxNumRowEntries();
```
mhoemmen: MueLu folks said this was OK for now.

packages/tpetra/core/src/Tpetra_RowMatrix_def.hpp
```diff
@@ -215,12 +215,14 @@ namespace Tpetra {
     }
     else { // the row Maps of A and B are not the same
       // Construct the result matrix C.
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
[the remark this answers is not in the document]
kddevin: This looks like a bug.  If we disable TPETRA_ENABLE_DEPRECATED_CODE, the "else" block is empty.  

Does the "else" condition never occur with TPETRA_ENABLE_DEPRECATED_CODE=OFF?  If so, shouldn't the entire block be in the ifdef guard?  

If the "else" condition would still occur, having an empty code block doesn't seem right. How does C get set?

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
```diff
@@ -128,10 +128,16 @@ namespace Tpetra {
 
   /*! Allocation profile for matrix/graph entries */
   enum ProfileType {
-    StaticProfile,  /*!< Single, static allocation (strict and more efficient) */
-    DynamicProfile  /*!< Multiple, dynamic allocations (flexibile, but less efficient) */
+    StaticProfile
   };
 
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
+  TPETRA_DEPRECATED const ProfileType DynamicProfile = ProfileType(StaticProfile+1);
```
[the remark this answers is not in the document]
kddevin: It will be gone in a month. :)

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
```diff
@@ -128,10 +128,16 @@ namespace Tpetra {
...
+#define TPETRA_DEFAULT_PROFILE_TYPE DynamicProfile
```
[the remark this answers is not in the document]
kddevin: It will be gone in a month. :)

packages/tpetra/core/src/Tpetra_RowMatrix_def.hpp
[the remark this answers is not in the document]
mhoemmen: @kddevin This is the case where A and B have different row Maps.  My guess is that this case never gets exercised in practice.  However, this should be easy to fix in terms of correctness: we would just Export both A and B to matrices with the same, nonoverlapping row Map, and then do the addition there (by calling `add` recursively).  I can do that if you like.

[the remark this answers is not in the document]
GeoffDanielson: This is a case that we specifically argued about previously - where the output matrix doesn't have enough entries per row to fulfill the operation.  It was my understanding that we failed on that case.

kddevin:
One test failed; I will take a look right now.

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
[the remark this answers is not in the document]
GeoffDanielson: It is so gross, I agree @mhoemmen

PR #4702, opened 2019-03-22
Improve cee-rhel6 intel-18.0.2 OpenMP test suite runtimes (#4251, #4260, #4262)

bartlettroscoe:
This uses MPI and OpenMP settings discovered by Brad King (Kitware) that massively improve the runtime of the ATDM Trilinos test suite for the build `Trilinos-atdm-cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt` bringing down the time from currently over 2 hours as shown [here](https://testing.sandia.gov/cdash-dev-view/index.php?project=Trilinos&parentid=4769282) to less than 9 minutes (see below).

This also seems to have fixed the timing out tests reported in #4251 #4260, and #4262. 

NOTE: The following tests FAILED:
    
    1477 - STKBalance_stk_balance_MPI_4 (Failed)
    1478 - STKBalance_stk_balance_m2n_MPI_4 (Failed)
   
But these tests also fail in the promoted cee-rhel6-intel-17.0.1 build so there is no reason to block the promotion of this build.

## How this was tested

I tested this on 'ceerws1113' where the nightly build `Trilinos-atdm-cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt` runs using:

```
$ cd /scratch/rabartl/Trilinos.base/BUILDS/ATDM/CEE-RHEL6/CHECKIN/

$  ./checkin-test-atdm-cee-rhel6.sh \
    cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt \
    --enable-all-packages=on --local-do-al
```

which gave the result:

```
  FAILED: Trilinos/cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt: passed=2207,notpassed=2
  
  Fri Mar 22 13:29:21 MDT 2019
  
  Enabled Packages: 
  Enabled all Packages
  Hostname: ceerws1113
  Source Dir: /scratch/rabartl/Trilinos.base/Trilinos/cmake/tribits/ci_support/../../..
  Build Dir: /scratch/rabartl/Trilinos.base/BUILDS/ATDM/CEE-RHEL6/CHECKIN/cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt
  
[...]

PR #4685, opened 2019-03-20
Switch from testing-vm to testing.sandia.gov

bartlettroscoe:
I noticed today that at some point I could not get to testing-vm.sandia.gov in
the browser but I could get to testing.sandia.gov.  Therefore, I figured this
might be the cause of the CDash upload problem I reported to Kitware today for
the build Trilinos-atdm-waterman-cuda-9.2-release-debug in:

* https://gitlab.kitware.com/snl/project-1/issues/89

I tested this locally on 'crf450' with:

```
$ env Trilinos_PACKAGES=Kokkos,Teuchos  \
  ./ctest-s-local-test-driver.sh sems-rhel6-gnu-7.2.0-openmp-release-debug
```

which showed, for example:

```
Submit files (using http)
   Send to track: Experimental
   Using HTTP submit method
   Drop site:http://testing.sandia.gov/cdash/submit.php?project=Trilinos
   Uploaded: /ascldap/users/rabartl/Trilinos.base/BUILDS/ATDM/CTEST_S/Trilinos-atdm-sems-rhel6-gnu-7.2.0-openmp-release-debug/SRC_AND_BUILD/BUILD/Testing/20190320-2350/Test.xml
   Submission successful
```

and that showed up on CDash at:

* https://testing.sandia.gov/cdash-dev-view/index.php?project=Trilinos&parentid=4759203

Issue #4262, opened 2019-01-24
MueLu tests timing out randomly on ATDM intel18 mpich build


fryeguy52:
CC: @trilinos/muelu, @srajama1 (Trilinos Linear Solvers Product Lead), @bartlettroscoe, @fryeguy52
  
## Next Action Status
<status-and-or-first-action>
 
## Description
 
As shown in the lonks below the tests:
* [MueLu_UnitTestsTpetra_MPI_4](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercombine=and&filtercombine=&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=4&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt&field2=buildstarttime&compare2=84&value2=2019-01-25&field3=buildstarttime&compare3=83&value3=2019-01-01&field4=testname&compare4=61&value4=MueLu_UnitTestsTpetra_MPI_4)
* [MueLu_UnitTestsBlockedTpetra_MPI_4](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercombine=and&filtercombine=&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=4&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt&field2=buildstarttime&compare2=84&value2=2019-01-25&field3=buildstarttime&compare3=83&value3=2019-01-01&field4=testname&compare4=61&value4=MueLu_UnitTestsBlockedTpetra_MPI_4)

are failing in the builds:
* Trilinos-atdm-cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt

## Current Status on CDash

[...]

Issue #4260, opened 2019-01-24
Belos tests timing out on ATDM intel-18 mpich build


fryeguy52:
CC: @trilinos/belos, @srajama1 (Trilinos Linear Solvers Product Lead), @bartlettroscoe, @fryeguy52
  
## Next Action Status
<status-and-or-first-action>
 
## Description
 
As shown in the links below the tests:
* [Belos_rcg_hb_MPI_4](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercombine=and&filtercombine=&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=4&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt&field2=buildstarttime&compare2=84&value2=2019-01-25&field3=buildstarttime&compare3=83&value3=2019-01-01&field4=testname&compare4=61&value4=Belos_rcg_hb_MPI_4)
* [Belos_gcrodr_hb_MPI_4](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercombine=and&filtercombine=&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=4&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt&field2=buildstarttime&compare2=84&value2=2019-01-25&field3=buildstarttime&compare3=83&value3=2019-01-01&field4=testname&compare4=61&value4=Belos_gcrodr_hb_MPI_4) 

are randomly timing out in the build:
* Trilinos-atdm-cee-rhel6_intel-18.0.2_mpich2-3.2_openmp_static_opt

## Current Status on CDash

[...]

PR #4706, opened 2019-03-25
Merge atdm-nightly to 'develop' to avoid future merge commits the other way

bartlettroscoe:
In order to get the ATDM Trilinos builds submitting results to CDash again, I had to manually merge the branch from the PR #4685 to the 'atdm-nightly' branch since the Trilinos PR tester is broken currently because it is trying to submit to testing-vm.sandia.gov which blocks all the PRs from getting merged.

I also merged in the branch from PR  #4702 in order to fix the intel-18.0.2 test runtimes to avoid timeouts in that build (see #4251, #4260, #4262).

It is important to merge 'atdm-nightly' to 'develop' so that future merges from 'develop' to 'nightly' will be non-fast-forward merges.  That is important so that the SHA1s shown on CDash will be SHA1s in the develop branch.  We don't want 'atdm-nightly' to be a release of Trilinos, we just want it to ensure that the same version of Trilinos is tested each testing day.

However, if the Trilinos PR tester continues to have problems we may need to update the 'atdm-nightly' branch more in the future (but we would like to avoid that) and then have the ATDM APPs pull from 'atdm-nightly' instead of 'develop' (but since that bypasses PR testing builds that would not be good to do in general).


bartlettroscoe:
Thanks @mhoemmen!

prwolfe:
This might have conflict due to the merge of #4691

bartlettroscoe:
@prwolfe said:

> This might have conflict due to the merge of #4691

If there is a conflict, I will merge 'develop' into a branch started from 'atdm-nightly' and then post a new PR.  As I said above, 'atdm-nightly' should be a fast-forward of 'develop' in general and we need a merge from 'atdm-nightly' to 'develop' to accomplish that going forward.

PR #4708, opened 2019-03-25
Tpetra: FECrsGraph cleanup 

csiefer2:
Auto-PR for SHA a031b0f

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -5661,63 +5661,72 @@ namespace Tpetra {
     using execution_space = typename device_type::execution_space;
     using row_ptrs_type = typename local_graph_type::row_map_type::non_const_type;
     using indices_type = t_GlobalOrdinal_1D;
+    using local_indices_type = t_LocalOrdinal_1D;
     using range_policy = Kokkos::RangePolicy<execution_space, Kokkos::IndexType<LocalOrdinal>>;
     using Tpetra::Details::padCrsArrays;
 
+    // Assume global indexing we don't have any indices yet
     if (! this->indicesAreAllocated()) {
       allocateIndices(GlobalIndices);
     }
-    TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(
-      ! this->isGloballyIndexed(), std::runtime_error,
-      ": must be globally indexed to resize!\n");
 
     // Making copies here because k_rowPtrs_ has a const type. Otherwise, we
     // would use it directly.
-    indices_type indices("indices", this->k_gblInds1D_.extent(0));
-    Kokkos::deep_copy(indices, this->k_gblInds1D_);
-
+    
     row_ptrs_type row_ptrs_beg("row_ptrs_beg", this->k_rowPtrs_.extent(0));
     Kokkos::deep_copy(row_ptrs_beg, this->k_rowPtrs_);
-
+    
     const size_t N = (row_ptrs_beg.extent(0) == 0 ? 0 : row_ptrs_beg.extent(0) - 1);
-    row_ptrs_type row_ptrs_end("row_ptrs_end", N);
-
+      row_ptrs_type row_ptrs_end("row_ptrs_end", N);
+    
     bool refill_num_row_entries = false;
     if (this->k_numRowEntries_.extent(0) > 0) {
       // Case 1: Unpacked storage
       refill_num_row_entries = true;
       auto num_row_entries = this->k_numRowEntries_;
       Kokkos::parallel_for("Fill end row pointers", range_policy(0, N),
-        KOKKOS_LAMBDA(const size_t i){
-          row_ptrs_end(i) = row_ptrs_beg(i) + num_row_entries(i);
-        }
-      );
-
+                           KOKKOS_LAMBDA(const size_t i){
+                             row_ptrs_end(i) = row_ptrs_beg(i) + num_row_entries(i);
+                           }
+                           );
+      
     } else {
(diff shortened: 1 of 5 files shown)


mhoemmen:
@csiefer2 Do the "weird recursions" involve nonmember `endFill` getting into a mysterious infinite loop with the member function `endFill`?  The Intel compiler is buggy sometimes.

csiefer2:
@mhoemmen We switched it to virtual because you brought up the issue previously.  Which, in retrospect, was a mistake.

mhoemmen:
@csiefer2 oops ;-P

csiefer2:
Nah, this is one of those, "We didn't think through the consequences and the tests were disabled"  things 🤦‍♂️

csiefer2:
Added code to mod CrsGraph so that self-transfer into a locallyIndexed matrix will work.

Note: CopyAndPermute does not work correctly into a locallyIndexed matrix, so non-self transfer still won't work.  But that never did.

@mhoemmen please re-bless at your leisure.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5657,67 +5657,75 @@ namespace Tpetra {
...
-      ! this->isGloballyIndexed(), std::runtime_error,
-      ": must be globally indexed to resize!\n");
 
     // Making copies here because k_rowPtrs_ has a const type. Otherwise, we
     // would use it directly.
-    indices_type indices("indices", this->k_gblInds1D_.extent(0));
-    Kokkos::deep_copy(indices, this->k_gblInds1D_);
-
+    
     row_ptrs_type row_ptrs_beg("row_ptrs_beg", this->k_rowPtrs_.extent(0));
```
mhoemmen: In the next commit, create `row_ptrs_beg` uninitialized.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5657,67 +5657,75 @@ namespace Tpetra {
...
-        }
-      );
+                           KOKKOS_LAMBDA(const size_t i){
+                             row_ptrs_end(i) = row_ptrs_beg(i+1);
+                           }
+                           );
+    }
+    
+    if(this->isGloballyIndexed()) {
+      indices_type indices("indices", this->k_gblInds1D_.extent(0));
```
mhoemmen: In the next commit, create `indices` uninitialized.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5657,67 +5657,75 @@ namespace Tpetra {
...
+      Kokkos::deep_copy(indices, this->k_gblInds1D_);
+      using padding_type = Kokkos::UnorderedMap<LocalOrdinal, size_t, device_type>;
+      padCrsArrays<row_ptrs_type,indices_type,padding_type>(row_ptrs_beg, row_ptrs_end, indices, padding);
```
mhoemmen: Are you paying extra for all these copies if you don't need to do any work?  I guess this code was here before and you're just adding the branch.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -6084,20 +6092,41 @@ namespace Tpetra {
...
       const LO numEnt = numPacketsPerLID[i];
       const GO* const gblColInds = (numEnt == 0) ? nullptr : &imports[importsOffset];
-      if (gblRow == Tpetra::Details::OrdinalTraits<GO>::invalid ()) {
-        // This row is not in the row Map on the calling process.
-        this->insertGlobalIndicesIntoNonownedRows (gblRow, gblColInds, numEnt);
-      }
+      if(!this->isLocallyIndexed()) {
+        if (gblRow == Tpetra::Details::OrdinalTraits<GO>::invalid ()) {
+          // This row is not in the row Map on the calling process.
+          this->insertGlobalIndicesIntoNonownedRows (gblRow, gblColInds, numEnt);
```
mhoemmen: Do we ever reach this case where the incoming row index is not in the row Map on the calling (in this case, receiving) process?  That doesn't really make sense -- if the Export or Import directed those rows to go to this process, then those rows should be in the target CrsGraph's row Map, no?

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -6084,20 +6092,41 @@ namespace Tpetra {
...
+        }
+      } 
       else {
-        this->insertGlobalIndicesFiltered (lclRow, gblColInds, numEnt);
+        for(LO j = 0; j < numEnt; j++)  {
+          lclColInds[j] = this->colMap_->getLocalElement(gblColInds[j]);
+        }
+        TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(gblRow == Tpetra::Details::OrdinalTraits<GO>::invalid (),
+                                              std::runtime_error,
+                                              "cannot insert into unowned rows if isLocallyIndexed().");
```

packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
```diff
@@ -252,16 +252,7 @@ namespace Tpetra {
...
     //@{
 
-    /// \brief Communicate nonlocal contributions to other processes.
-    ///
-    /// This method is called automatically by fillComplete().  Most
-    /// users do not need to call this themselves.
-    ///
-    /// This method must be called collectively (that is, like any MPI
-    /// collective) over all processes in the graph's communicator.
-    void globalAssemble (){endFill();}
```
mhoemmen: If you removed this due to an Intel compiler bug, you could try qualifying the method with `this->`.

PR #4716, opened 2019-03-25
Zoltan2:  Remove Unused Parameter Warnings

jmgate:
@trilinos/zoltan2 

## Description
`-Wunused-parameter` warnings.

## Motivation and Context
Clean build.

### DIFF
--- a/packages/zoltan2/src/algorithms/order/Zoltan2_AlgAMD.hpp
+++ b/packages/zoltan2/src/algorithms/order/Zoltan2_AlgAMD.hpp
@@ -128,7 +128,7 @@ class AlgAMD : public Algorithm<Adapter>
     { }
 
     int globalOrder(
-      const RCP<GlobalOrderingSolution<typename Adapter::gno_t> > &solution) {
+      const RCP<GlobalOrderingSolution<typename Adapter::gno_t> > &/* solution */) {
         throw std::logic_error("AlgAMD does not yet support global ordering.");
     }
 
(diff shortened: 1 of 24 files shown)


packages/zoltan2/src/algorithms/order/Zoltan2_AlgAMD.hpp
```diff
@@ -128,14 +128,15 @@ class AlgAMD : public Algorithm<Adapter>
...
-      const RCP<GlobalOrderingSolution<typename Adapter::gno_t> > &solution) {
+      const RCP<GlobalOrderingSolution<typename Adapter::gno_t> > &/* solution */) {
         throw std::logic_error("AlgAMD does not yet support global ordering.");
     }
 
     int localOrder(
       const RCP<LocalOrderingSolution<typename Adapter::lno_t> > &solution)
     {
 #ifndef HAVE_ZOLTAN2_AMD
+      (void)solution;
```
kddevin: An odd line like this one should have a comment saying why it was added (e.g., "// needed to avoid unused-parameter warnings for function argument solution").  Otherwise, someone is likely to say someday, "That's weird" and remove it.

jmgate: Addressed in the latest commit.

jmgate:
@kddevin, any other changes you'd like made to it?

PR #4719, opened 2019-03-25
MueLu: Add MagmaSparse support to matvec driver

### LINKED ISSUE #4718: MueLu: Add MagmaSparse support to Matvec driver
Enable evaluation and testing of SpMVs using MAGMA Sparse.

@trilinos/muelu 

## Expectations
Matvec driver should be able to compare Tpetra vs Magama Sparse for ScalarType=Double and Ordinals = INT or LONG LONG INT

## Possible Solution
Accept my PR that provides this.

jjellio:
@trilinos/muelu 

## Description
This PR adds a MAGMA Sparse TPL to Trilinos and optionally enables that functionality in MueLu.
The MatvevDriver provides testing for correctness, but this testing is not enabled by default, since this TPL's use depends on how it was built (e.g., Cuda vs MIC).

The test currently provides Cuda functionality + testing.  TODO is to ensure it works seamlessly with many core.

Notes:
This moves the matvec driver out of the EXPERIMENTAL guard (for evaluations we really don't want to be enabling Experimental IMO).

## Motivation and Context
Allow testing of SpMV using various underlying implementations.  This driver intentionally works at the Xpetra/Tpetra level, so that matrices are constructed in a  manner similar to how MueLu does.

## How Has This Been Tested?
The code has correctness testing, which is defined by comparing against SpMVs performed by Tpetra.

## Checklist
Closes #4718


jhux2:
___
*[packages/muelu/cmake/Dependencies.cmake, line 9 at r1](https://reviewable.io/reviews/trilinos/trilinos/4719#-LarNqySAUJJdtnnvuGv:-LarNqyT1U3WcrsdVNhX:b7txjez) ([raw file](https://github.com/trilinos/trilinos/blob/1b2ab3e6889edbe12ba2eb52793539fb2cebd12c/packages/muelu/cmake/Dependencies.cmake#L9)):*
> ```CMake
> SET(TEST_OPTIONAL_DEP_PACKAGES AztecOO Galeri Belos Pamgen)
> SET(LIB_REQUIRED_DEP_TPLS BLAS LAPACK)
> SET(LIB_OPTIONAL_DEP_TPLS Boost MATLAB AmgX CGAL ViennaCL MKL Avatar CUSPARSE MAGMASparse)
> ```

Shouldn't this be an optional test dependency?

jhux2:
___
*[packages/muelu/cmake/Dependencies.cmake, line 9 at r1](https://reviewable.io/reviews/trilinos/trilinos/4719#-LarNqySAUJJdtnnvuGv:-LarOQn_CKE2LzJB3tT4:b-7hwan5) ([raw file](https://github.com/trilinos/trilinos/blob/1b2ab3e6889edbe12ba2eb52793539fb2cebd12c/packages/muelu/cmake/Dependencies.cmake#L9)):*
<details><summary><i>Previously, jhux2 (Jonathan Hu) wrote…</i></summary><blockquote>

Shouldn't this be an optional test dependency?
</blockquote></details>

Also, we've not done a good job of this, but could you also add this to the table in MueLu user's guide on page 19?

jjellio:
@jhu2 I didn't specify a test, because it wasn't obvious how to filter out tests by enabled execution spaces.  Currently, this is tested with Cuda.  Without out Cuda, it should build the original KK v Tpetra code. I haven't done the stuff to figure out Magma (on non-cuda).  The code could function as a test though... For Magma and CuSparse TPL usage.

As for documentation: What would I document? That you can access the TPL if you want? The same would be true for CuSPARSE. I haven't provided any TPL specific functionality in the library.

packages/muelu/test/scaling/CMakeLists.txt
```diff
@@ -415,22 +415,17 @@ IF (${PACKAGE_NAME}_ENABLE_Experimental AND ${PACKAGE_NAME}_ENABLE_Tpetra)
...
-      ARGS "--linAlgebra=Tpetra"
-      NUM_MPI_PROCS 1
-      COMM serial mpi
-      )
   ENDIF()
 ENDIF()
 
 
+# remove this from Experimental it requires Galeri + Tpetra
+IF (${PACKAGE_NAME}_ENABLE_Tpetra)
```
mhoemmen: Galeri is optional (surprisingly! see `Dependencies.cmake` above), so you'll also need to protect this with `${PACKAGE_NAME}_ENABLE_Galeri`.

packages/muelu/test/scaling/MatvecKernelDriver.cpp
```diff
@@ -100,46 +100,34 @@ void print_crs_graph(std::string name, const V1 rowptr, const V2 colind) {
...
+template<typename Scalar, typename LocalOrdinal, typename GlobalOrdinal, typename Node>
+class MagmaSparse_SpmV_Pack {
+  typedef Tpetra::CrsMatrix<Scalar,LocalOrdinal,GlobalOrdinal,Node> crs_matrix_type;
+  typedef Tpetra::MultiVector<Scalar,LocalOrdinal,GlobalOrdinal,Node> vector_type;
+
+public:
+  MagmaSparse_SpmV_Pack (const crs_matrix_type& A,
+                      const vector_type& X,
+                            vector_type& Y)
+  {}
```
mhoemmen: It would be nicer for this class not to compile at all if Scalar is not one of the supported types.

packages/muelu/test/scaling/MatvecKernelDriver.cpp
```diff
@@ -100,46 +100,34 @@ void print_crs_graph(std::string name, const V1 rowptr, const V2 colind) {
...
 
+ ~MagmaSparse_SpmV_Pack() {};
```
mhoemmen: Prefer `= default;` to `{};`.

packages/muelu/test/scaling/MatvecKernelDriver.cpp
```diff
@@ -100,46 +100,34 @@ void print_crs_graph(std::string name, const V1 rowptr, const V2 colind) {
...
-  default:
-    return (std::string("ERROR: cuda_error got a return code it doesn't not understand") + std::to_string(code));
-  }
-}
+#if defined(HAVE_MUELU_MAGMASPARSE) && defined(HAVE_MUELU_TPETRA)
+#include <magma_v2.h>
+#include <magmasparse.h>
+template<typename Scalar, typename LocalOrdinal, typename GlobalOrdinal, typename Node>
+class MagmaSparse_SpmV_Pack {
+  typedef Tpetra::CrsMatrix<Scalar,LocalOrdinal,GlobalOrdinal,Node> crs_matrix_type;
```
mhoemmen: Prefer C++11 `using crs_matrix_type = Tpetra::CrsMatrix<...>;`.

cmake/TPLs/FindTPLMAGMASparse.cmake
```diff
@@ -0,0 +1,61 @@
...
+# INFORMATION, APPARATUS, PRODUCT, OR PROCESS DISCLOSED, OR REPRESENTS
+# THAT ITS USE WOULD NOT INFRINGE PRIVATELY OWNED RIGHTS.
+#
+# ************************************************************************
+# @HEADER
+
+
+TRIBITS_TPL_FIND_INCLUDE_DIRS_AND_LIBRARIES( MAGMASparse
+  REQUIRED_HEADERS magmasparse.h
+  REQUIRED_LIBS_NAMES magma_sparse magma
```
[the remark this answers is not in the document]
jjellio: Actually, I started with MagmaSparse only... but they do depend.  MagmaSparse requires Magma.

This TPL is a WIP.  We could make a MAGMA TPL, that provides an identifier that states whether it provides sparse. I would need to look at NetCDF or HDF5, since those packages provide various attributes along with  libraries/headers..

We should talk about my scheme for supporting multiple linear algebra things.

packages/muelu/test/scaling/CMakeLists.txt
jjellio: Unless I screwed up, this is inside the larger Galeri statement.

packages/muelu/test/scaling/MatvecKernelDriver.cpp
jjellio: Will do... I've added your changes in from the CuSparse round, but due to PR latency it seems more efficient to push them in the next round.

packages/muelu/test/scaling/MatvecKernelDriver.cpp
jjellio: I'm not sure. This gets complicated, because we instantiate the driver itself for multiple ETI types.  The general idea is that the code will compile, but that it simply won't provide the runtime option.

What we really need, is a better way to express the different instantiations (it would speed up build times).

mhoemmen: Christian's way of doing it in kokkos-kernels makes sense.  You have a struct with a static method that says "try using `${TPL}` to solve the problem."  If `${TPL}` can't do it -- e.g., because it doesn't support the Scalar type -- the method returns false, and the caller has to try something else.

packages/muelu/test/scaling/MatvecKernelDriver.cpp
mhoemmen: ok

cmake/TPLs/FindTPLMAGMASparse.cmake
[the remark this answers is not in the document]
mhoemmen: > This TPL is a WIP.

Shall we mark it as such then? or do you mean "WIP" in the sense that you plan future work?

> We should talk about my scheme for supporting multiple linear algebra things.

Let's do that at some point.

jhux2:
> @jhu2 I didn't specify a test, because it wasn't obvious how to filter out tests by enabled execution spaces. Currently, this is tested with Cuda. Without out Cuda, it should build the original KK v Tpetra code. I haven't done the stuff to figure out Magma (on non-cuda). The code could function as a test though... For Magma and CuSparse TPL usage.

No, I mean something different.  I believe Magma should be in `TEST_OPTIONAL_DEP_PACKAGES`.  As it is now, doesn't cmake think this is an optional *library* dependency?

> 
> As for documentation: What would I document? That you can access the TPL if you want? The same would be true for CuSPARSE. I haven't provided any TPL specific functionality in the library.

Fair point.  I guess the only thing one might document would be that one can do comparisons with Magma.

jjellio:
> No, I mean something different. I believe Magma should be in TEST_OPTIONAL_DEP_PACKAGES. As it is now, doesn't cmake think this is an optional library dependency?

I think, I maybe perhaps can possibly understand what you mean. 

With`LIB_OPTIONAL_DEP_TPLS`,  this assumes the library depends on the TPL. With `TEST_OPTIONAL_DEP_PACKAGES` That would mean a 'test' depends on the TPL, but not the library itself?  I can fiddle with it and see if it works. If that is the case, then CUSPARSE needs to get moved there as well.

cmake/TPLs/FindTPLMAGMASparse.cmake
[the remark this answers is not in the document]
jjellio: WIP = work in progress = meaning the details of in CMake for the TPL need to be worked on - but that is in the direction of better multi-blas support.

jhux2:
> > No, I mean something different. I believe Magma should be in TEST_OPTIONAL_DEP_PACKAGES. As it is now, doesn't cmake think this is an optional library dependency?
> 
> I think, I maybe perhaps can possibly understand what you mean.
> 
> With`LIB_OPTIONAL_DEP_TPLS`, this assumes the library depends on the TPL. With `TEST_OPTIONAL_DEP_PACKAGES` That would mean a 'test' depends on the TPL, but not the library itself? I can fiddle with it and see if it works. If that is the case, then CUSPARSE needs to get moved there as well.

Yes, this is what I mean.

jjellio:
You have increased my TRIBITS knowledge by 1!

PR #4732, opened 2019-03-27
Another step towards correct cmake files for stk...

alanw0:
In this case, user-facing executables were being added
as tests, and the unit-test executable wasn't being added
at all.

Also fix a case where stk-balance was calling a couple of
deprecated exodus functions.

Also restrict the unit-tests that are run, since we don't yet
have sufficient guards in place for insisting on what packages
and tpls need to have been enabled...

@trilinos/<teamName>

## Checklist


alanw0:
@bartlettroscoe I'm not sure why they failed for certain platforms. If the set of enables differs, then that could be an explanation. I don't yet have the guards in place to have stk_balance insist that the set of enables matches what is being done in the sierra build. But the things I'm fixing in this PR are clear problems: previously the unit-tests weren't being built; also the user-facing executables stk_balance and stk_balance_m2n were being declared as tests and they shouldn't be. So this is a step in the right direction even though the journey isn't complete yet.

mhoemmen:
@trilinos/stk

packages/stk/stk_balance/test_utils/MeshFixtureMxNRebalance.hpp
```diff
@@ -79,7 +79,7 @@ class MeshFixtureMxNRebalance : public stk::unit_test_util::MeshFixture
     void create_field_on_entire_mesh(const std::string& fieldName)
     {
         targetDecompField = &get_meta().declare_field<stk::mesh::Field<double> >(stk::topology::ELEMENT_RANK, fieldName, 1);
-        stk::mesh::put_field_on_mesh(*targetDecompField, get_meta().universal_part(), nullptr);
+        stk::mesh::put_field_on_mesh(*targetDecompField, get_meta().universal_part(), (double*)nullptr);
```
mhoemmen: It looks like `put_field_on_mesh` is templated on the type of the third argument, so `nullptr` won't suffice.  Still, I would prefer `static_cast<double*>` to the C-style cast.

packages/stk/stk_unit_tests/stk_balance/UnitTestLoadBalanceMultipleCriteria.cpp
```diff
@@ -60,9 +60,9 @@ class MultipleCriteria : public stk::unit_test_util::MeshFixture
             stk::mesh::Field<double> &criteria2 = get_meta().declare_field<stk::mesh::Field<double>>(stk::topology::ELEM_RANK, "criteria2");
             stk::mesh::Field<double> &procId = get_meta().declare_field<stk::mesh::Field<double>>(stk::topology::ELEM_RANK, "proc_id");
 
-            stk::mesh::put_field_on_mesh(criteria1, get_meta().universal_part(), nullptr);
-            stk::mesh::put_field_on_mesh(criteria2, get_meta().universal_part(), nullptr);
-            stk::mesh::put_field_on_mesh(procId, get_meta().universal_part(), nullptr);
+            stk::mesh::put_field_on_mesh(criteria1, get_meta().universal_part(), (double*)nullptr);
```
mhoemmen: See comment above on C-style casts.

alanw0: I agree Mark, static_cast is better. I've pushed a new commit to switch these to static_cast.

Issue #4281, opened 2019-01-28
Tpetra: sort and merge values: CrsMatrix::insertLocalValues


tjfulle:
## What

Sort and merge new values in `CrsMatrix::insertLocalValues`

## Why

`CrsMatrix::insertLocalValues` does not sort/merge existing values. Instead, indices/values are appended to their row of the matrix, regardless if they are already present in the row.  During `fillComplete`, rows are sorted/merged so that duplicate entries are "compressed" out.  For `StaticProfile` matrix, this can lead to running out of capacity due to duplicate entries.

@trilinos/tpetra 

## Related Issues
#4278, #4279, #4280

tjfulle:
This is done in #4733.  indices are not sorted/merged, but are unique.

mhoemmen:
Closed by PR #4733.

Issue #4280, opened 2019-01-28
Tpetra: sort and merge values: CrsMatrix::insertGlobalValues


tjfulle:
## What

Sort and merge new values in `CrsMatrix::insertGlobalValues`

## Why

`CrsMatrix::insertGlobalValues` does not sort/merge existing values. Instead, indices/values are appended to their row of the matrix, regardless if they are already present in the row.  During `fillComplete`, rows are sorted/merged so that duplicate entries are "compressed" out.  For `StaticProfile` matrix, this can lead to running out of capacity due to duplicate entries.

@trilinos/tpetra 

## Related Issues
#4278, #4279

mhoemmen:
Closed by PR #4733.

tjfulle:
Thanks @mhoemen!

Issue #4279, opened 2019-01-28
Tpetra: sort and merge on insert: CrsGraph::insertLocalIndices


tjfulle:
## What

Sort and merge new indices in `CrsGraph::insertLocalIndices`

## Why

`CrsGraph` does not sort/merge new indices. Instead, indices are appended to their row of the graph, regardless if they are already present in the row.  During `fillComplete`, rows are sorted/merged so that duplicate entries are "compressed" out.  For `StaticProfile` graph, this can lead to running out of capacity due to duplicate entries.

@trilinos/tpetra 

## Related Issues
#4278

mhoemmen:
Closed by PR #4733.

PR #4733, opened 2019-03-27
Tpetra: code clean up between CrsMatrix and CrsGraph

### LINKED ISSUE #4278: Tpetra: sort and merge on insert: CrsGraph::insertGlobalIndices
## What

Sort and merge new indices in `CrsGraph::insertGlobalIndices`

## Why

`CrsGraph` does not sort/merge new indices. Instead, indices are appended to their row of the graph, regardless if they are already present in the row.  During `fillComplete`, rows are sorted/merged so that duplicate entries are "compressed" out.  For `StaticProfile` graph, this can lead to running out of capacity due to duplicate entries.

@trilinos/tpetra

tjfulle:
@trilinos/tpetra

## Description
Implements mechanisms to more closely couple CrsMatrix to the underlying graph.  Moves CrsMatrix away from the business of modifying its graph's structure directly.  Instead, CrsGraph is left to do any modifications and uses callback functions to communicate the changes back to CrsMatrix.

## Motivation and Context
Code clean up related to deprecation of DynamicProfile

## Related Issues

* Makes #4519 obsolete
* Closes #4278, #4279, #4280, #4281 

## How Has This Been Tested?
Tpetra tests pass locally


packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -1838,11 +1838,40 @@ namespace Tpetra {
     size_t
     insertGlobalIndicesImpl (const RowInfo& rowInfo,
                              const GlobalOrdinal inputGblColInds[],
-                             const size_t numInputInds);
+                             const size_t numInputInds,
+                             std::function<void(const size_t, const size_t, const size_t)> fun =
+                                 std::function<void(const size_t, const size_t, const size_t)>());
 
     void
```
mhoemmen: Hm, how come this one returns `void` instead of `size_t`?  (Don't have to change for now.)

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -2106,283 +2108,111 @@ namespace Tpetra {
...
-             << ".  Please report this bug to the Tpetra developers.");
-        }
-        return numNewToInsert; // all done!
-      } // if the graph is StaticProfile
-      else { // DynamicProfile
+    if (this->getProfileType () == StaticProfile) {
+      auto numEntries = rowInfo.numEntries;
+      using inp_view_type = View<const GO*, execution_space, MemoryUnmanaged>;
+      inp_view_type inputInds(inputGblColInds, numInputInds);
+      auto numInserted = Details::insertCrsIndices(lclRow, k_rowPtrs_,
```
mhoemmen: It's a bit annoying that this function returns signed, but the caller returns unsigned.  I'm not too bothered by that, though.  `std::optional` and ilk could help here, once they arrive.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -2407,6 +2237,90 @@ namespace Tpetra {
...
+
+    size_t numFound = 0;
+    LocalOrdinal lclRow = rowInfo.localRow;
+    if (this->isLocallyIndexed())
+    {
+      if (this->colMap_.is_null())
+        return invalidCount;
+      auto map =
+        [&](GlobalOrdinal const gblInd){
+          return this->colMap_->getLocalElement(gblInd);
```
mhoemmen: You might want to make the lambda lighter-weight:
```c++
const auto& colMap = * (this->colMap_);
auto map = [&] (const GO gblInd) { return colMap.getLocalElement (gblInd); };
```
Small lambdas (in this case, it only holds a reference) might have a small object optimization.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -3413,6 +3327,18 @@ namespace Tpetra {
     indicesAreAllocated_ = true;
     indicesAreLocal_     = true;
     pftype_              = StaticProfile; // if the profile wasn't static before, it sure is now.
+
+    // Column indices sorted by row
```
mhoemmen: 1. Were there actual cases where they were _not_ sorted by row on input?
2. MueLu usually wants to give Tpetra sorted arrays here -- should we add an option not to sort? 

@csiefer2

tjfulle: That is a vestige of the last PR where indices are sorted and I didn’t mean for it to sneak through.  I’ll remove the changes so setAplIndices’ behavior is not changed

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
tjfulle: Which, the std::function callback?  If so, why would it return anything but void?  The caller never uses the return value

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: The signed return value was for the function to return a negative number to indicate 1 or more indices were invalid.  I can change it to be unsigned and return OrdinalTraits<size_t>::invalid() for that case instead

tjfulle: Changed in latest commit

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: Changed in latest commit

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: Reverted to previous behavior in latest commit

tjfulle:
Removing `[WIP]` designation to run the PR tests.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
mhoemmen: @tjfulle I mean the outer method, `insertLocalIndicesImpl`.  `insertGlobalIndicesImpl` returns `size_t`, but `insertLocalIndicesImpl` returns `void`.

tjfulle: @mhoemmen Ah, I see.  I don’t know why, I’d have to redirect that question back to you!

mhoemmen: @tjfulle I don't know why either ;-)  We should fix that at some point.

mhoemmen:
@tjfulle Only the "SERIAL" (non-MPI) test failed; is that an issue with the test?
```
 Next-door-neighbor graph test
      myrow_gbl.size() = 1 == 3 : FAILED ==> /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_gcc_4.9.3_SERIAL/Trilinos/packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests0.cpp:1045
      Calling globalAssemble()
      ngraph.getNumEntriesInLocalRow(0) = 1 == ngraph.getNumAllocatedEntriesInLocalRow(0) = 3 : FAILED ==> /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_gcc_4.9.3_SERIAL/Trilinos/packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests0.cpp:1051
```

tjfulle:
@mhoemmen, I’ll take a look.  I had to fix several tests that expected the old behavior of appending duplicate indices.  This could be a case where multiple duplicates were inserted, artificially increasing the size of the row.  I’ll take a look

tjfulle:
yep, it was a case of the test expected duplicate entries to be inserted.  Fixed in 6b1f72a

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -1943,37 +1943,46 @@ namespace Tpetra {
...
-      Teuchos::av_reinterpret_cast<const IST> (values);
-    this->insertIndicesAndValues (graph, rowInfo, indsView, valsView,
-                                  valsIn, LocalIndices, LocalIndices);
+
+    if (this->getProfileType() == StaticProfile)
+    {
+      Teuchos::ArrayView<IST> valsView = this->getViewNonConst(rowInfo);
+      auto fun = [&](size_t const k, size_t const /*start*/, size_t const offset) {
+                   valsView[offset] += values[k]; };
+      graph.insertLocalIndicesImpl(lclRow, indices, fun);
```
mhoemmen: Since it's necessary for the graph to take `fun` as an `std::function`, and since the graph won't keep the `std::function`, we could create the `std::function` explicitly by reference.  This would trigger small object optimizations for `std::function`.  That is, it would end up just storing a single reference, instead of various ArrayView stuff (that would likely bring it over the small object optimization limit).  Here's how to do it:
```c++
// Prefer std::cref.  You can't use it here, since
// ArrayView<IST>::operator[] is nonconst.
std::function<void(const size_t, const size_t, const size_t)> fun2 (std::ref (fun));
// You can't inline the declaration of fun inside fun2 above,
// since that would make the lambda const and would require
// std::ref(const T&&), which is deleted.
graph.insertLocalIndicesImpl (lclRow, indices, fun2);
```

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -2015,42 +2027,48 @@ namespace Tpetra {
...
-      Teuchos::Array<IST>& curVals = this->values2D_[rowInfo.localRow];
-      // Teuchos::ArrayRCP::resize automatically copies over values on
-      // reallocation.
-      graph.gblInds2D_[rowInfo.localRow].resize (newNumEnt);
-      curVals.resize (newNumEnt);
-      rowInfo.allocSize = newNumEnt; // reassign for updated allocSize
+    if (this->getProfileType() == StaticProfile)
+    {
+      Teuchos::ArrayView<IST> valsView = this->getViewNonConst(rowInfo);
+      auto fun = [&](size_t const k, size_t const /*start*/, size_t const offset) {
```
mhoemmen: See note above.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -2352,9 +2370,20 @@ namespace Tpetra {
...
                           const LocalOrdinal numElts) const
   {
+    if (graph.getProfileType() == StaticProfile)
+    {
+      Teuchos::ArrayView<const LocalOrdinal> indsT(inds, numElts);
+      auto fun =
+        [&](size_t const k, size_t const /*start*/, size_t const offset) {
+          rowVals[offset] = newVals[k];
+        };
+      return graph.findLocalIndices(rowInfo, indsT, fun);
```
mhoemmen: See note above on `std::function` small object optimization.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -2475,6 +2504,18 @@ namespace Tpetra {
                            const impl_scalar_type newVals[],
                            const LocalOrdinal numElts) const
   {
+    if (graph.getProfileType() == StaticProfile)
+    {
+      Teuchos::ArrayView<const GlobalOrdinal> indsT(inds, numElts);
+      auto fun =
```
mhoemmen: See note above on `std::function` small object optimization.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -2600,6 +2641,21 @@ namespace Tpetra {
                            const LocalOrdinal numElts,
                            const bool atomic) const
   {
+    if (graph.getProfileType() == StaticProfile)
+    {
+      Teuchos::ArrayView<const GlobalOrdinal> indsT(inds, numElts);
+      auto fun =
```
mhoemmen: See note above on `std::function` small object optimization.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -2407,6 +2228,86 @@ namespace Tpetra {
...
+
+    size_t numFound = 0;
+    LocalOrdinal lclRow = rowInfo.localRow;
+    if (this->isLocallyIndexed())
+    {
+      if (this->colMap_.is_null())
+        return invalidCount;
+      const auto& colMap = *(this->colMap_);
+      auto map = [&](GO const gblInd){return colMap.getLocalElement(gblInd);};
+      numFound = Details::findCrsIndices(lclRow, k_rowPtrs_,
```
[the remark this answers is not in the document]
tjfulle: Hmmm, `Details::findCrsIndices` is templated on the `map` type.  Are you seeing a different version of the code?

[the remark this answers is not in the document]
tjfulle: `Details::insertCrsIndices` is *not* templated on the `map` type since it leads to an ambiguous function overload set since the callback function is optional.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
tjfulle: done.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
tjfulle: Done.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
tjfulle: Done.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
tjfulle: Done.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
tjfulle: Done.

tjfulle:
@trilinos/framework @william76 how do I fire off the pre-merge inspection?

prwolfe:
> @trilinos/framework @william76 how do I fire off the pre-merge inspection?

That is automated. The PR tester won't do them when it is running a build but will get to it.

PR #4734, opened 2019-03-27
Tpetra: Start working around Intel compiler issues noticed in PR #4648

mhoemmen:
@trilinos/tpetra 

## Description

Start working around Intel compiler issues that hindered PR #4648.  Add functions in the `Tpetra::Details` namespace for getting Views of statically allocated memory.  (We'll need this for #4626.)  Overload `Tpetra::deep_copy` for `Teuchos::SerialDenseMatrix`.  (We'll also need this for #4626.)

This PR is less ambitious than PR #4648, since I want to make sure that the Intel compiler work-around gets in before we pile more fixes on top.

## Related Issues

* Is blocked by #4747 
* Related to #4626, #4648

## How Has This Been Tested?

With CUDA.

PR #4740, opened 2019-03-27
Tpetra: Fix #4729

### LINKED ISSUE #4729: Tpetra:  std::call_once does not work well with gcc serial builds
@trilinos/tpetra @egboman 

Recall from #3057:
Use of std::call_once with gcc in serial builds requires TPL_ENABLE_PThread=ON.

This requirement is not obvious to the user; indeed, only because I had been bitten by it once before (#3033) did I know that info was available in a github issue.  Most users will not guess that Pthreads is required for a serial build (nor should they).

Use of std::call_once has been reintroduced in Tpetra_Details_Behavior.cpp.
Is it necessary?  Can it be removed?

## Expectations

Serial configurations should not need to specify TPL Pthreads to work correctly with Tpetra.

## Current Behavior

std::call_once crashes in serial Trilinos builds without TPL_ENABLE_Pthread=ON.

## Motivation and Context

@egboman is trying to run serial coloring algorithms in Zoltan2 on his Linux workstation.  Tpetra CrsMatrix construction fails because std::call_once fails in fetching debug environment variables in CrsGraph constructor.

## Definition of Done

I hope:  std::call_once removed from Tpetra; I hate being bitten twice by the same bug.

## Steps to Reproduce

build serial Trilinos with Zoltan2 and TPL_ENABLE_Pthread=OFF
make test

## Your Environment
On @egboman's workstation

module purge
module load sems-env
module load sems-cmake
module load sems-gcc/4.9.3

cmake \
  -D CMAKE_BUILD_TYPE:STRING=DEBUG \
  -D BUILD_SHARED_LIBS:BOOL=ON \
  -D Trilinos_ENABLE_Zoltan2:BOOL=ON \
  -D Zoltan2_ENABLE_Experimental:BOOL=ON \
  -D TPL_BLAS_LIBRARIES="/usr/lib64/libblas.so.3" \
  -D TPL_LAPACK_LIBRARIES="/usr/lib64/liblapack.so.3" \
..

92bf4f0867cd171f89afcfd0a2266d33fd77de39

## Related Issues

Same as #3057 
Related to #3033

mhoemmen:
@trilinos/tpetra 

## Description

  - Remove use of C++11 thread stuff from `Tpetra::Details::Behavior`.
  - Fix Tpetra build warnings observed on the Dashboard

## Motivation and Context

1. GCC annoyingly requires linking with libpthread in order for that stuff to work.  GCC apparently generates code that crashes, if Trilinos was not built with libpthread support.  I consider that a GCC bug, but we have to work around.
2. `std::call_once` etc. only help with thread safety of code that uses `std::thread` threads.  They don't help with other thread models, like OpenMP (or even Pthreads).  This reduces the value of this feature.

## Related Issues

* Closes #4729

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_Behavior.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_Behavior.cpp
@@ -1,10 +1,8 @@
 #include "Tpetra_Details_Behavior.hpp"
 #include "TpetraCore_config.h"
 #include <algorithm> // std::transform
-#include <atomic> // std::atomic_thread_fence, std::memory_order_release
 #include <cstdlib> // std::getenv
 #include <cctype> // std::toupper
-#include <mutex> // std::call_once, std::once_flag
 #include <string>
 #include <map>
 #include <vector>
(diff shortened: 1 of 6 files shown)


mhoemmen:
@trilinos/framework I submitted https://github.com/trilinos/Trilinos/pull/4734 12 hours ago, and the builds haven't started yet.  I submitted this PR an hour ago, and the builds just started.  What gives?

mhoemmen:
@prwolfe explained just now that PR testing distributes PRs to servers in round-robin fashion, and PR #4734 happened to get assigned to a heavily loaded server.  Framework has plans to fix the code that does this distribution, so that this is less likely to happen.

prwolfe:
the PR autotester instance 4 is currently running for PR 4629. I expect that to run at least another hour. At that point it will pick up 4539, 4649, 4699, or this one depending on state. Let just say this instance is busy.

4539 has already passed and is waiting for review
4649 has failed twice so it should not run
4699 has a retest flag on it so it will be considered

I'm not sure which will run next, but I think it will be the lowest numbered, so 4699 first, then 4734

egboman:
All my Zoltan2 tests pass now. Thanks, @mhoemmen !

PR #4741, opened 2019-03-27
clean_workspace: correct flag and make command

prwolfe:
the --force-clean flag had two conflicting names
and the final   make command (make -C %s clean)
failed in an infinite recursion.  I just modified
it to cd to the directory first and then run 'make clean'

@trilinos/framework 

## How Has This Been Tested?
tested with the unit tests and I created a directory, built and used both the date and force options to clean it.


prwolfe:
Not bonkers is a pretty high bar ;_>

PR #4745, opened 2019-03-27
Fix some more cmake issues in stk_balance and stk_util.

alanw0:
This is attempting to fix the error Albany is seeing.

@trilinos/<teamName>

## Checklist


packages/stk/stk_balance/test_utils/BalanceTestUtilities.cpp
```diff
@@ -1,6 +1,6 @@
 #include "BalanceTestUtilities.hpp"
-#include <gtest/gtest.h>
 #include <iomanip>
+#include <unistd.h>
```
mhoemmen: That's a POSIX header file -- do you actually need POSIX, non-C++ functions?

alanw0: Unfortunately this file uses the 'unlink' function.

mhoemmen:
My current web browser won't let me reply in line: If you need `unlink` then including that header file makes sense :-)  Once the Filesystem additions to the C++ Standard Library land, we won't need nonstandard functions like that any more.

PR #4747, opened 2019-03-27
Tpetra: Fix #4633 & #4639

### LINKED ISSUE #4633: Tpetra::MultiVector constructor that takes a device View uses create_mirror instead of create_mirror_view

### LINKED ISSUE #4639: Tpetra::MultiVector::reduce broken if getStride() > getLocalLength()

mhoemmen:
@trilinos/tpetra @trilinos/belos @vbrunini 

## Description

This PR supersedes PR #4648.  I've extracted out the parts of #4648 that are actual bug fixes rather than optimizations.  Here they are:

  - Fix #4639 (an actual correctness bug in `Tpetra::MultiVector::reduce` that hindered solving #4626)
  - Fix #4633 (a strange thing in `Tpetra::MultiVector` that could lead to bugs, especially with `FEMultiVector`)
  - Clean up a few things in Belos

The part of PR #4648 I did _not_ include here is the fix for #4626.  That triggers a likely Intel compiler bug in PR tests, which is why I'm not including it.  I have a multiple-step work-around in progress; see PR #4734 for the start of that.

## Related Issues

* Closes #4639, #4633 
* Blocks #4626 
* Related to #4640, #4648, #4734 

## How Has This Been Tested?

With and without CUDA.

### DIFF
--- a/packages/belos/tpetra/src/solvers/Belos_Tpetra_Cg.hpp
+++ b/packages/belos/tpetra/src/solvers/Belos_Tpetra_Cg.hpp
@@ -15,15 +15,13 @@ private:
   using base_type = Krylov<SC, MV, OP>;
 
 public:
-  Cg () :
-    base_type::Krylov ()
-  {}
+  Cg () = default;
 
   Cg (const Teuchos::RCP<const OP>& A) :
     base_type::Krylov (A)
   {}
 
-  virtual ~Cg () {}
+  virtual ~Cg () = default;
 
 protected:
   using vec_type = typename base_type::vec_type;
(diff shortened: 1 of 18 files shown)


mhoemmen:
@trilinos/framework @prwolfe `tr-test-1.novalocal` strikes again, killing the MueLu build :-(

mhoemmen:
@vbrunini @micahahoward @jclause I think this is the third week of trying to get this stuff through PR testing -- the linker keeps dying in the middle of MueLu :( .  FYI, this means that the stuff that makes both Aria and SPARC go faster on multiple GPUs, that we've tested in both, hasn't gotten in yet.

prwolfe:
Well, that is happening everywhere of course.

mhoemmen:
@prwolfe Is this is a problem w/ the one test machine perhaps?

jhux2:
Lgtm

mhoemmen:
PR #4765 has the rest of the fix for #4626.

PR #4748, opened 2019-03-28
Tpetra: Fix #4680

### LINKED ISSUE #4680: Tpetra::CrsGraph, CrsMatrix: Constructor that takes an array of max row counts, takes it by ArrayRCP, but doesn't actually keep the reference
@trilinos/tpetra @vbrunini 

Both `Tpetra::CrsGraph` and `Tpetra::CrsMatrix` have constructors that take an array of the maximum number of entries in each row.  The constructors take that array by `Teuchos::ArrayRCP<const size_t>`.  Idiomatically, taking an ArrayRCP of const suggests that the class keeps a reference to the array, so that users shouldn't modify it.  This is not what happens.  Tpetra actually deep-copies the array.

We should thus change the constructors so they take either a raw array, or `Teuchos::ArrayView<const size_t>`.  Once we have C++20, we can use `std::span`, which exactly expresses the desired idiom in native C++.

I found this by looking through Chaparral (@vbrunini of course has already fixed it to use StaticProfile, because he is just effective like that ;-) ).  I saw that it was using a nonowning ArrayRCP there, and got worried, but then discovered that it didn't matter, since CrsGraph (and thus CrsMatrix) just deep-copies it.

mhoemmen:
@trilinos/tpetra 

## Description

Deprecate `Tpetra::CrsGraph` and `Tpetra::CrsMatrix` constructors that take `ArrayRCP<const size_t>` for the maximum number of entries per row.  Replace them with `ArrayView<const size_t>` constructors.

## Motivation and Context

Deprecation deadline for the next Trilinos release is coming soon.  Also, I can't get any other PRs through PR testing, so I might as well try something.

## Related Issues

* Closes #4680

### DIFF
--- a/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
+++ b/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
@@ -822,7 +822,7 @@ namespace Tpetra {
         // do first-touch reallocation (a NUMA (Non-Uniform Memory
         // Access) optimization on multicore CPUs).
         RCP<sparse_matrix_type> A =
-          rcp (new sparse_matrix_type (pRowMap, myNumEntriesPerRow,
+          rcp (new sparse_matrix_type (pRowMap, myNumEntriesPerRow (),
                                        DynamicProfile));
 
         // List of the global indices of my rows.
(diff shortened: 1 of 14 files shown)


egboman:
OK

PR #4753, opened 2019-03-28
Tempus fix fpes

ikalash:
This PR addressed the following Albany issue https://github.com/SNLComputation/Albany/issues/461 by fixing floating point exceptions in some of the Newmark steppers.  These were only showing up in a debug build with FPE checking on. 

All Tempus tests pass, as do all Albany Tempus tests with the fix.

### DIFF
--- a/packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
+++ b/packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
@@ -196,7 +196,9 @@ void StepperNewmarkExplicitAForm<Scalar>::setInitialConditions(
     auto f       = initialState->getX()->clone_v();
     this->evaluateExplicitODE(f, x, xDot, initialState->getTime());
     Thyra::Vp_StV(f.ptr(), Scalar(-1.0), *(xDotDot));
-    Scalar reldiff = Thyra::norm(*f)/Thyra::norm(*xDotDot);
+    Scalar reldiff = Thyra::norm(*f);
+    //The following logic is to prevent FPEs  
+    if (Thyra::norm(*xDotDot) > 1.0e-12) reldiff /= Thyra::norm(*xDotDot); 
 
     Scalar eps = Scalar(100.0)*std::abs(Teuchos::ScalarTraits<Scalar>::eps());
     if (reldiff > eps) {
@@ -314,7 +316,7 @@ std::string StepperNewmarkExplicitAForm<Scalar>::description() const
 template<class Scalar>
 void StepperNewmarkExplicitAForm<Scalar>::describe(
    Teuchos::FancyOStream               &out,
-   const Teuchos::EVerbosityLevel      /* verbLevel */) const
+   const Teuchos::EVerbosityLevel      verbLevel) const
 {
   out << description() << "::describe:" << std::endl
       << "appModel_ = " << this->appModel_->description() << std::endl;
--- a/packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
+++ b/packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
@@ -197,11 +197,11 @@ void StepperNewmarkExplicitAForm<Scalar>::setInitialConditions(
     this->evaluateExplicitODE(f, x, xDot, initialState->getTime());
     Thyra::Vp_StV(f.ptr(), Scalar(-1.0), *(xDotDot));
     Scalar reldiff = Thyra::norm(*f);
+    Scalar normxDotDot = Thyra::norm(*xDotDot); 
     //The following logic is to prevent FPEs  
-    if (Thyra::norm(*xDotDot) > 1.0e-12)
-      reldiff /= Thyra::norm(*xDotDot); 
-
     Scalar eps = Scalar(100.0)*std::abs(Teuchos::ScalarTraits<Scalar>::eps());
+    if (normxDotDot > eps*reldiff) reldiff /= normxDotDot;  
+
     if (reldiff > eps) {
       RCP<Teuchos::FancyOStream> out = this->getOStream();
       Teuchos::OSTab ostab(out,1,"StepperForwardEuler::setInitialConditions()");
--- a/packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
+++ b/packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
@@ -198,7 +198,8 @@ void StepperNewmarkExplicitAForm<Scalar>::setInitialConditions(
     Thyra::Vp_StV(f.ptr(), Scalar(-1.0), *(xDotDot));
     Scalar reldiff = Thyra::norm(*f);
     //The following logic is to prevent FPEs  
-    if (Thyra::norm(*xDotDot) > 1.0e-12) reldiff /= Thyra::norm(*xDotDot); 
+    if (Thyra::norm(*xDotDot) > 1.0e-12)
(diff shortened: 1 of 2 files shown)


packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
```diff
@@ -314,7 +316,7 @@ std::string StepperNewmarkExplicitAForm<Scalar>::description() const
 template<class Scalar>
 void StepperNewmarkExplicitAForm<Scalar>::describe(
    Teuchos::FancyOStream               &out,
-   const Teuchos::EVerbosityLevel      /* verbLevel */) const
+   const Teuchos::EVerbosityLevel      verbLevel) const
```
jmgate: Replace `verbLevel` with `/* verbLevel */`.

packages/tempus/src/Tempus_StepperNewmarkImplicitAForm_impl.hpp
```diff
@@ -424,7 +427,7 @@ std::string StepperNewmarkImplicitAForm<Scalar>::description() const
 template<class Scalar>
 void StepperNewmarkImplicitAForm<Scalar>::describe(
    Teuchos::FancyOStream               &out,
-   const Teuchos::EVerbosityLevel      /* verbLevel */) const
+   const Teuchos::EVerbosityLevel      verbLevel) const
```
jmgate: Replace `verbLevel` with `/* verbLevel */`.

ikalash:
@jmgate : sure thing, just made the change.  Can you please re-review?

packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
```diff
@@ -196,7 +196,9 @@ void StepperNewmarkExplicitAForm<Scalar>::setInitialConditions(
     auto f       = initialState->getX()->clone_v();
     this->evaluateExplicitODE(f, x, xDot, initialState->getTime());
     Thyra::Vp_StV(f.ptr(), Scalar(-1.0), *(xDotDot));
-    Scalar reldiff = Thyra::norm(*f)/Thyra::norm(*xDotDot);
+    Scalar reldiff = Thyra::norm(*f);
+    //The following logic is to prevent FPEs  
+    if (Thyra::norm(*xDotDot) > 1.0e-12) reldiff /= Thyra::norm(*xDotDot);
```
ccober6: So I like the compare, but we should probably used a relative compare 
```
if (Thyra::norm(*x) > eps*reldiff) reldiff /= Thyra::norm(*x); 
```
Also I tried to be consistent and use x instead of xDotDot across the Steppers.  that means the output below this needs changing.

packages/tempus/src/Tempus_StepperNewmarkImplicitAForm_impl.hpp
```diff
@@ -290,8 +290,11 @@ void StepperNewmarkImplicitAForm<Scalar>::setInitialConditions(
     appInArgs.set_t        (initialState->getTime()    );
 
     this->wrapperModel_->getAppModel()->evalModel(appInArgs, appOutArgs);
-
-    Scalar reldiff = Thyra::norm(*f)/Thyra::norm(*x);
+ 
+    Scalar reldiff = Thyra::norm(*f);
+    //The following logic prevents FPEs  
+    if (Thyra::norm(*x) > 1.0e-12) reldiff /= Thyra::norm(*x);
```
ccober6: Same as above.

packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
```diff
@@ -196,7 +196,10 @@ void StepperNewmarkExplicitAForm<Scalar>::setInitialConditions(
     auto f       = initialState->getX()->clone_v();
     this->evaluateExplicitODE(f, x, xDot, initialState->getTime());
     Thyra::Vp_StV(f.ptr(), Scalar(-1.0), *(xDotDot));
-    Scalar reldiff = Thyra::norm(*f)/Thyra::norm(*xDotDot);
+    Scalar reldiff = Thyra::norm(*f);
+    //The following logic is to prevent FPEs  
+    if (Thyra::norm(*xDotDot) > 1.0e-12)
+      reldiff /= Thyra::norm(*xDotDot);
```
jmgate: Reposting @ccober6's suggestion:

> So I like the compare, but we should probably use a relative compare 
> ```
> if (Thyra::norm(*x) > eps*reldiff) reldiff /= Thyra::norm(*x); 
> ```
> Also I tried to be consistent and use x instead of xDotDot across the Steppers.  That means the output below this needs changing.

I modified these lines because some compilers would have warned on misleading indentation.

packages/tempus/src/Tempus_StepperNewmarkImplicitAForm_impl.hpp
```diff
@@ -290,8 +290,12 @@ void StepperNewmarkImplicitAForm<Scalar>::setInitialConditions(
     appInArgs.set_t        (initialState->getTime()    );
 
     this->wrapperModel_->getAppModel()->evalModel(appInArgs, appOutArgs);
-
-    Scalar reldiff = Thyra::norm(*f)/Thyra::norm(*x);
+ 
+    Scalar reldiff = Thyra::norm(*f);
+    //The following logic prevents FPEs  
+    if (Thyra::norm(*x) > 1.0e-12)
+      reldiff /= Thyra::norm(*x);
```
jmgate: From @ccober6:

> Same as above.

packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
ikalash: @ccober6 : I like the relative compare.  I'll go ahead and make the change.  When you say: 

Also I tried to be consistent and use x instead of xDotDot across the Steppers. That means the output below this needs changing.

I think you're referring to this code: https://github.com/trilinos/Trilinos/blob/master/packages/tempus/src/Tempus_StepperNewmarkImplicitAForm_impl.hpp#L296-L307 , is that right?  In particular, 
```
 << "  ||f(x,xDot,xDotDot,t)||/||x|| = " << reldiff << std::endl
```
wouldn't be accurate if `||x|| = 0`, so for that case the print statement needs modification.  Is that what you meant?

ccober6: Yeah.  I will need to go and clean the other steppers.  (Output can be so tedious.  :) )

ikalash: @ccober6 : I see, thanks for clarifying.  If I make the change to the output for the 2 steppers involved in this PR, are you OK merging it in, and having the cleaning up of the output for the other steppers be part of another PR?  I'd like to merge this in sooner rather than later b/c w/o it, there are some failing tests in the Albany nightlies.

ikalash: @ccober6 : I just pushed a fix to do the relative compare as you suggest.  I decided not to change the output at this point - I thought looking at the code and what you said, that if you do a general cleanup of the output that is printed to the screen for all the steppers in a separate PR, it probably makes sense to do this one at that time so that they all look consistent.  I'd like to have this PR pushed to fix the Albany tests failures in the meantime - let me know if you're OK with that.

ccober6: Yes, I was thinking the same thing.

Issue #4770, opened 2019-03-31
Deprecation in Tpetra::CrsMatrix constructor triggers warning in Xpetra that crashes the SPARC build


bartlettroscoe:
CC: @trilinos/xpetra, @trilinos/tpetra , @kddevin  (Trilinos Data Services Product Lead), @bartlettroscoe, @fryeguy52, @mhoemmen 

## Next Action Status

<status-and-or-first-action>

## Description

The commit 24cef7553e09f35f08a8c93bd3c7f8dab5aeb3ea merged in PR #4748 deprecated a constructor for `Tpetra::CrsMatrix` and triggered a deprecated warning when SPARC builds showing the error:

```
/projects/atdm_devops/trilinos_installs/2019-03-29/cee-rhel6_clang-5.0.1_openmpi-1.10.2_serial_static_opt/include/Xpetra_TpetraCrsMatrix.hpp:109:31: error: 'CrsMatrix' is deprecated [-Werror,-Wdeprecated-declarations]
     : mtx_(Teuchos::rcp(new Tpetra::CrsMatrix (toTpetra(rowMap), NumEntriesPerRowToAlloc, toTpetra(pftype), params))) {  }
```

This deprecated warning (elevated to an error in SPARC because they set `-Werror`) is triggered inside of Trilinos headers but is not shown in the Xpetra test suite shown [here](https://testing.sandia.gov/cdash-dev-view/index.php?project=Trilinos&parentid=4809252).  Does this show a gap in the automated tests for Xpetra?

## Current Status on CDash

The current status can only be seen in the SPARC Trilinos Integration builds shown over the last few days [here](http://compsim-dashboard.sandia.gov/cdash/index.php?project=SPARC&date=2019-03-30&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=66&value1=-trildev&field2=buildstarttime&compare2=83&value2=5%20days%20ago).

## Steps to Reproduce

Since the native test suite does not show a problem, one can only reproduce this by building SPARC itself.  See the instructions for doing that [here](https://snl-wiki.sandia.gov/display/CoodinatedDevOpsATDM/Building+ATDM+APPs+Against+Local+Installs+of+Trilinos#BuildingATDMAPPsAgainstLocalInstallsofTrilinos-BuildingagainsttheSPARCTrilinos'master'versionusingthenativeSPARCTrilinosconfigurationscripts).

[...]

PR #4772, opened 2019-03-31
Remove usage of deprecated Tpetra::CrsMatrix constructor (#4770)

bartlettroscoe:
CC: @trilinos/tpetra 

This allows SPARC to build again without requiring any changes to SPARC.

Should not this Xpetra function be deprecated as well?

### DIFF
--- a/packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix.hpp
+++ b/packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix.hpp
@@ -106,7 +106,7 @@ namespace Xpetra {
 
     //! Constructor specifying (possibly different) number of entries in each row.
     TpetraCrsMatrix(const Teuchos::RCP< const Map< LocalOrdinal, GlobalOrdinal, Node > > &rowMap, const ArrayRCP< const size_t > &NumEntriesPerRowToAlloc, ProfileType pftype=DynamicProfile, const Teuchos::RCP< Teuchos::ParameterList > &params=Teuchos::null)
-      : mtx_(Teuchos::rcp(new Tpetra::CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> (toTpetra(rowMap), NumEntriesPerRowToAlloc, toTpetra(pftype), params))) {  }
+      : mtx_(Teuchos::rcp(new Tpetra::CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> (toTpetra(rowMap), NumEntriesPerRowToAlloc(), toTpetra(pftype), params))) {  }
 
     //! Constructor specifying column Map and fixed number of entries for each row.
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
FYI: I will try adding `-Wno-deprecated-declarations` and see if that fixes SPARC without having to apply this patch.

mhoemmen:
@bartlettroscoe wrote:
> Should not this Xpetra function be deprecated as well?

That's up to the @trilinos/muelu team.  Xpetra doesn't necessarily promise the same semantics as Tpetra.

bartlettroscoe:
@mhoemmen write:

> That's up to the @trilinos/muelu team. Xpetra doesn't necessarily promise the same semantics as Tpetra.

But if Tpetra deprecates something, then Xpetra must do the same deprecation or users will be stuck with deprecated warnings and no way to turn them off (other than turn off all deprecated warnings).

mhoemmen:
@bartlettroscoe wrote:
> But if Tpetra deprecates something, then Xpetra must do the same deprecation or users will be stuck with deprecated warnings and no way to turn them off (other than turn off all deprecated warnings).

It should be OK, since `Xpetra::CrsMatrix != Tpetra::CrsMatrix`, as long as Xpetra doesn't call deprecated functions (thanks for the fix! :-) ).

bartlettroscoe:
@mhoemmen,

I removed `AT: AUTOMERGE` and set `AT: WIP` because I think a better solution is to set `-Wno-error=deprecated-declaration` in the SPARC builds that have `-Werror` set.  I am testing this now.

But longer-term, we need to figure out why the native Xpetra build did not show these deprecated warnings.  Are all the Xpetra wrapper functions not getting tested?

mhoemmen:
@bartlettroscoe wrote:
> But longer-term, we need to figure out why the native Xpetra build did not show these deprecated warnings. Are all the Xpetra wrapper functions not getting tested?

It's possible.  Xpetra mainly gets tested through MueLu.  There are some Xpetra unit tests but it's possible that they don't have complete coverage.

bartlettroscoe:
@trilinos/xpetra, @trilinos/muelu, @trilinos/tpetra ,

Do we still want to merge this little fix in the PR or do we want to make a more complete fix of the Xpetra interface as well to match the updated non-deprecated Tpetra interface?  Does it make sense for the Xpetra and Tpetra interfaces to diverge?

mhoemmen:
@jhux2 (sometimes it helps to name specific folks for notifications to kick in)

bartlettroscoe:
FYI: I added `-Wno-error=deprecated-declaration` to the SPARC builds so this no longer results in a build error.  But it would still be nice to merge this PR so that SPARC developers don't see a deprecated warning they can't do anything about.

Therefore, I removed `AT: WIP` and added `AT: AUTOMERGE` to get this merged.

lucbv:
@bartlettroscoe it is possible that the test coverage in Xpetra is not complete however these warnings did show up in MueLu.
Actually there are a few more spots in Xpetra that need to be cleaned up. I will take care of those in a separate PR.

PR #4781, opened 2019-04-01
Muelu: Allow to change distributor options

cgcgcg:

## Description
This PR adds the option to set a new sublist "matvec params" in the xml interface. The content of the sublist will be applied to the underlying Tpetra::Distributors for the solve phase. In particular, this allows to choose different communication options apart from the default Irecv-Send pattern.


mhoemmen:
@cgcgcg I wouldn't mind if we got rid of all the Distributor options except whether to use persistent communication (per your PR https://github.com/trilinos/Trilinos/pull/4782).  Ready sends are never a good idea, and the implementation knows when it can use `MPI_Isend`.

Issue #3580, opened 2018-10-09
Tpetra::Distributor: Fix "slow path" so we can use MPI_Isend


mhoemmen:
@trilinos/tpetra @jjellio @csiefer2 

Fix the "slow path" of `Distributor::doPosts`, so we can use nonblocking sends (`MPI_Isend`).  The "slow path" kicks in when the data to send are not neatly grouped in contiguous chunks per process.  It permutes the data into contiguous-by-target-process-rank chunks for sending.  Currently, the slow path uses the same send buffer for all the messages.  This means that it cannot use nonblocking sends.

We must fix both the "three-argument" (all messages have the same size) and "four-argument" (different messages may have different sizes) overloads of `doPosts`, and both the `Teuchos::ArrayRCP` and `Kokkos::View` versions of each.

## Motivation and Context

This is part of the overall effort to improve MPI+CUDA performance and make Tpetra's boundary exchange and sparse matrix-vector multiply communication nonblocking.

## Definition of Done

- [x] Fix 3-argument `Teuchos::ArrayRCP` overload of `doPosts`
- [x] Fix 3-argument `Kokkos::View` overload of `doPosts`
- [x] Fix 4-argument `Teuchos::ArrayRCP` overload of `doPosts`
- [x] Fix 4-argument `Kokkos::View` overload of `doPosts`

## Related Issues

* Part of #383

mhoemmen:
PR #3581 fixes all four cases.  In that PR, Distributor now _only_ uses `MPI_Isend`.

mhoemmen:
PR #3581 broke one test _way_ downstream in MueLu.  This suggests that we need much better unit testing in Distributor.

mhoemmen:
@trilinos/tpetra @jjellio Current plan: First, _avoid_ the slow path.  

#4412 (fixed in develop by PR #4413) prevents Distributor from unnecessarily engaging the slow path for reverse-mode communication.  (Reverse-mode sends should _always_ be contiguous by process rank, since forward-mode receives are always contiguous by process rank.)

[...]

Issue #767, opened 2016-11-02
Tpetra: Overlap communication and computation


mhoemmen:
@trilinos/tpetra 

Epic for tracking all stories relating to overlapping communication and computation in Tpetra.

Issue #608, opened 2016-09-12
Tpetra::Distributor: Split out execution of comm pattern into separate class 


mhoemmen:
@trilinos/tpetra 

Tpetra::Distributor currently combines setting up a communication pattern, with executing that pattern.  Methods for executing the communication pattern are templated on Packet type.

We might like to have different Distributor back-ends, in order to support communication protocols other than MPI 2-sided.  (For example, we could have an MPI 1-sided implementation, or a PGAS implementation, or a Kokkos-wrapping-PGAS implementation.)  This implies a base class with subclasses for the different implementations.

In order to make execution of a communication plan happen through virtual methods, we would need to template the class on Packet, not the methods.  (Templated methods can't be virtual.)  However, the setup code does _not_ depend on the Packet type.  We would not want to build all that setup code redundantly for all Packet types (there are a lot of them!).

This suggests splitting the setup code into a separate class from the execution code.

Issue #383, opened 2016-05-23
Tpetra::Distributor: Make doPosts nonblocking


mhoemmen:
@trilinos/tpetra 

Epic: #767.

Tpetra::Distributor implements the MPI communication that happens in an Export or Import.  It uses MPI 2-sided point-to-point communication.  Its `doPosts` method starts the receives and sends, and its `doWaits` method waits on them (`MPI_Waitall`).  Receives are nonblocking (`MPI_Irecv`) and sends may be either blocking (various options, but only `MPI_Send` is used in practice) or nonblocking (`MPI_Isend`).  However, sends default to blocking, and this is the only completely correct path.  This is because of the so-called "slow path" of doPosts.

The "slow path" comes about when the indices in a send to a particular process aren't contiguous (i.e., are interrupted by data ~~from~~ [meant for]* other process(es)).  The current implementation thus requires an intermediate pack buffer in that case.  It allocates the extra buffer on the spot.  In order to avoid holding on to that memory, the implementation forces blocking sends in that case (it throws `std::logic_error` otherwise).

Two fixes come to mind:
1. Keep the extra buffer.  Keep it in the returned CommRequest so it doesn't get deallocated.
2. Pre-permute the data during packing (`DistObject::packAndPrepare`) so the slow path never gets invoked in practice.

The first fix is easier, but may be ultimately less performant.

The "slow path" occurs in both the 3-argument (fixed # packets per index, used by Vector and MultiVector) and 4-argument (variable # packets per index, used by CrsGraph, CrsMatrix, etc.) versions of doPosts.  The 3-argument version matters most for solver performance, but it's easy to do both at the same time.

[*edit by jhux2 23-Jan-2019]

mhoemmen:
Some ideas for how to work on this:

1. Change "slow path" to use a single, sufficiently long buffer.
2. Only use the Isend paths.  Purge all other send options and the "barrier between" option.
3. Make `requests_` an `std::vector<MPI_Request>`, only existing in MPI builds.  Make Distributor keep the buffer (including the "single, sufficiently long" buffer for the slow path mentioned above).  Only reallocate that buffer if needed.
[...]

PR #4782, opened 2019-04-01
WIP Tpetra: persistent MPI requests

cgcgcg:
@trilinos/tpetra 

## Description
This PR adds the option to use persistent MPI communication in Tpetra. The defaults communication option is not changed. I tested this using MueLu, where this can be switched on for the solve phase only. There are probably quite a lot of cases where this will not work correctly, so I appreciate feedback!

### DIFF
--- a/packages/teuchos/comm/src/Teuchos_Comm.hpp
+++ b/packages/teuchos/comm/src/Teuchos_Comm.hpp
@@ -149,6 +149,12 @@ public:
 
   /// Wait on this request (a blocking operation).
   virtual RCP<CommStatus<OrdinalType> > wait () = 0;
+
+  //! Start this request.
+  virtual void start () = 0;
+
+  //! Free this request.
+  virtual void free () = 0;
 };
 
 /// \class Comm
@@ -606,6 +612,54 @@ public:
             const int sourceRank,
             const int tag) const = 0;
 
+
+  /** \brief Persistent send.
+   *
+   * \param sendBuffer [in] The data buffer to be sent.
+   *
+   * \param destRank [in] The rank of the process to receive the data.
+   *
+   * <b>Preconditions:</b><ul>
+   * <li><tt>0 <= destRank && destRank < this->getSize()</tt>
+   * <li><tt>destRank != this->getRank()</tt>
+   * </ul>
+   */
+  virtual RCP<CommRequest<Ordinal> > sendInit(
+    const ArrayView<const char> &sendBuffer,
+    const int destRank
+    ) const = 0;
+
+  //! Variant of isend() that takes a tag.
+  virtual RCP<CommRequest<Ordinal> >
+  sendInit (const ArrayView<const char> &sendBuffer,
+         const int destRank,
+         const int tag) const = 0;
+
+  /** \brief Persistent receive.
+   *
+   * \param recvBuffer [out] The location for storing the received data.
+   *
+   * \param sourceRank [in] The rank of the process to receive the data from.
+   * If <tt>sourceRank < 0</tt> then data will be received from any process.
+   *
+   * <b>Preconditions:</b><ul>
+   * <li>[<tt>sourceRank >= 0] <tt>sourceRank < this->getSize()</tt>
+   * <li><tt>sourceRank != this->getRank()</tt>
+   * </ul>
+   *
+   * \return Returns the senders rank.
+   */
+  virtual RCP<CommRequest<Ordinal> > receiveInit(
+    const ArrayView<char> &recvBuffer,
+    const int sourceRank
+    ) const = 0;
+
+  //! Variant of ireceive that takes a tag.
+  virtual RCP<CommRequest<Ordinal> >
+  receiveInit (const ArrayView<char> &recvBuffer,
+            const int sourceRank,
+            const int tag) const = 0;
+
   /** \brief Wait on a set of communication requests.
    *
    * <b>Preconditions:</b><ul>
@@ -617,7 +671,8 @@ public:
    * </ul>
    */
   virtual void waitAll(
-    const ArrayView<RCP<CommRequest<Ordinal> > > &requests
+    const ArrayView<RCP<CommRequest<Ordinal> > > &requests,
+    bool releaseRequests=true
     ) const = 0;
 
   /// \brief Wait on communication requests, and return their statuses.
@@ -638,7 +693,8 @@ public:
   ///   requests.
   virtual void
   waitAll (const ArrayView<RCP<CommRequest<Ordinal> > >& requests,
-           const ArrayView<RCP<CommStatus<Ordinal> > >& statuses) const = 0;
+           const ArrayView<RCP<CommStatus<Ordinal> > >& statuses,
+           bool releaseRequests=true) const = 0;
 
   /// \brief Wait on a single communication request, and return its status.
   ///
@@ -665,6 +721,34 @@ public:
   /// associated with the CommRequest object has completed.
   virtual RCP<CommStatus<Ordinal> >
   wait (const Ptr<RCP<CommRequest<Ordinal> > >& request) const = 0;
+
+  /// \brief Start a single communication request.
+  ///
+  /// \param request [in/out] On input: request is not null, and
+  /// <tt>*request</tt> is either null (in which case this function
+  /// does nothing and returns null) or an RCP of a valid CommRequest
+  /// instance representing an outstanding communication request.
+  virtual void
+  start (const Ptr<RCP<CommRequest<Ordinal> > >& request) const = 0;
+
+  /** \brief Start on a set of communication requests.
+   *
+   * <b>Preconditions:</b><ul>
+   * <li> <tt>requests.size() > 0</tt>
+   * </ul>
+   *
+   */
+  virtual void
+  startAll(const ArrayView<RCP<CommRequest<Ordinal> > > &requests) const = 0;
+
+  /// \brief Free a single communication request.
+  ///
+  /// \param request [in/out] On input: request is not null, and
+  /// <tt>*request</tt> is either null (in which case this function
+  /// does nothing and returns null) or an RCP of a valid CommRequest
+  /// instance representing an communication request.
+  virtual void
+  free (const Ptr<RCP<CommRequest<Ordinal> > >& request) const = 0;
 
   //@}
 
--- a/packages/teuchos/comm/src/Teuchos_DefaultMpiComm.hpp
+++ b/packages/teuchos/comm/src/Teuchos_DefaultMpiComm.hpp
@@ -210,6 +210,11 @@ public:
     return tmp_rawMpiRequest;
   }
 
+  MPI_Request getRawMpiRequest()
+  {
+    return rawMpiRequest_;
+  }
+
   //! Whether the raw MPI_Request is <tt>MPI_REQUEST_NULL</tt>.
   bool isNull() const {
     return rawMpiRequest_ == MPI_REQUEST_NULL;
@@ -268,6 +273,25 @@ public:
         "error: " << mpiErrorCodeToString (err));
       return mpiCommStatus<OrdinalType> (status);
     }
+  }
+
+  /// \brief Start the communication request.
+  void start () {
+    int err = MPI_Start (&rawMpiRequest_);
+    TEUCHOS_TEST_FOR_EXCEPTION(
+      err != MPI_SUCCESS, std::runtime_error,
+      "Teuchos: MPI_Start failed with the following error: "
+      << mpiErrorCodeToString (err));
+  }
+
+  /// \brief Free the communication request.
+  void free () {
+    int err = MPI_Request_free (&rawMpiRequest_);
+    TEUCHOS_TEST_FOR_EXCEPTION(
+      err != MPI_SUCCESS, std::runtime_error,
+      "Teuchos: MPI_Start failed with the following error: "
+      << mpiErrorCodeToString (err));
+    rawMpiRequest_ = MPI_REQUEST_NULL;
   }
 
   //! Destructor; cancels the request if it is still pending.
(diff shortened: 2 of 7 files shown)


packages/teuchos/comm/src/Teuchos_DefaultMpiComm.hpp
```diff
@@ -261,6 +266,15 @@ class MpiCommRequestBase : public CommRequest<OrdinalType> {
     }
   }
 
+  /// \brief Start the communication request.
+  void start () {
```
mhoemmen: This method should be virtual, and it should be added to the base class `CommRequest`.  We could use `std::function` and lambda capture to implement `start` for `SerialComm`.  Compare to how Tpetra implements `Tpetra::Details::iallreduce` when `MPI_VERSION < 3`.

packages/tpetra/core/src/Tpetra_Distributor.cpp
```diff
@@ -554,9 +566,32 @@ namespace Tpetra {
       std::ostringstream os;
       os << myRank << ": doWaits: # reqs = "
          << requests_.size () << endl;
+      os << myRank << ": doWaits: # persistentReqs = "
+         << persistentRequests_.size () << endl;
       *out_ << os.str ();
     }
 
+    int count = persistentRequests_.size();
```
mhoemmen: 1. Use of raw MPI needs to be protected by the `HAVE_TPETRACORE_MPI` macro.
2. Just call `Teuchos::waitAll`; it does exactly what lines 578-586 do.
3. We should talk about better encapsulations (e.g., a "CommRequests" interface with an `MpiCommRequests` base class that has an array of raw `MPI_Request` in it).

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -69,6 +69,12 @@
 #include <sstream>
 #include <type_traits>
 
+#ifdef HAVE_TEUCHOS_MPI
+#  include "Teuchos_Comm.hpp"
+#  include "Teuchos_Details_MpiTypeTraits.hpp"
```
mhoemmen: Please include `Tpetra_Details_MpiTypeTraits.hpp` instead.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -69,6 +69,12 @@
 #include <sstream>
 #include <type_traits>
 
+#ifdef HAVE_TEUCHOS_MPI
```
mhoemmen: The right macro is `HAVE_TPETRACORE_MPI`.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -261,7 +268,21 @@ namespace Tpetra {
     ///
     /// \pre No outstanding communication requests.
     ///   (We could check, but see GitHub Issue #1303.)
-    virtual ~Distributor () = default;
+    ~Distributor () {
+      for (int i = 0; i<persistentRequests_.size(); i++) {
+        Teuchos::RCP<Teuchos::CommRequest<int> > request = persistentRequests_[i];
+        Teuchos::RCP<Teuchos::MpiCommRequestBase<int> > mpiRequest =
+          Teuchos::rcp_dynamic_cast<Teuchos::MpiCommRequestBase<int> > (request);
+        MPI_Request mR = mpiRequest->releaseRawMpiRequest();
```
mhoemmen: Let's talk about better encapsulation strategies.  I suspect that you don't have to do anything here; CommRequest's virtual destructor will dispatch to MpiCommRequest's destructor, which will cancel and free the outstanding requests.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -261,7 +268,21 @@ namespace Tpetra {
...
+        const int err = MPI_Request_free(&mR);
+        if (err != MPI_SUCCESS) {
+          std::ostringstream os;
+          os << "Teuchos::MpiComm::waitAll: MPI_Request_free() failed with error \""
+             << Teuchos::mpiErrorCodeToString (err) << "\".";
+          TEUCHOS_TEST_FOR_EXCEPTION(true, std::runtime_error, os.str());
```
mhoemmen: Destructors shouldn't throw.  Best practice would be to ignore any MPI errors at this point.  How is user code supposed to respond to them?

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -1339,6 +1368,100 @@ namespace Tpetra {
     if (indicesTo_.empty ()) {
       // Data are already blocked (laid out) by process, so we don't
       // need a separate send buffer (besides the exports array).
+
+      if (sendType == Details::DISTRIBUTOR_PERSISTENT) {
+
+        if (persistentRequests_.size() == 0) {
+
+          Teuchos::RCP<const Teuchos::MpiComm<int> > tmpic = Teuchos::rcp_dynamic_cast<const Teuchos::MpiComm<int> >(comm_);
```
mhoemmen: See above.  I think we can encapsulate this in a way that doesn't expose MPI.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -1339,6 +1368,100 @@ namespace Tpetra {
...
+
+              Packet t;
+              MPI_Datatype rawType = Teuchos::Details::MpiTypeTraits<Packet>::getType (t);
+              Packet* rawRecvBuf = recvBuf.getRawPtr ();
+              const int count = as<int> (recvBuf.size ());
+              MPI_Request rawRequest = MPI_REQUEST_NULL;
+              MPI_Recv_init (rawRecvBuf, count, rawType, procsFrom_[i], tag, rawComm, &rawRequest);
+
+              Teuchos::RCP<Teuchos::CommRequest<int> > req = Teuchos::Details::mpiCommRequest(rawRequest, Teuchos::arcp_reinterpret_cast<const char>(recvBuf));
+              Teuchos::set_extra_data( recvBuf, "buffer", inOutArg(req) );
```
mhoemmen: I think this can be part of the encapsulation as well.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -1339,6 +1368,100 @@ namespace Tpetra {
...
+        Array<MPI_Request> rawMpiRequests (count, MPI_REQUEST_NULL);
+        int k = 0;
+        for (int i = 0; i<persistentRequests_.size(); i++) {
+          Teuchos::RCP<Teuchos::CommRequest<int> > request = persistentRequests_[i];
+          Teuchos::RCP<Teuchos::MpiCommRequestBase<int> > mpiRequest =
+            Teuchos::rcp_dynamic_cast<Teuchos::MpiCommRequestBase<int> > (request);
+          rawMpiRequests[k] = mpiRequest->getRawMpiRequest();
+          k++;
+        }
+        const int err = MPI_Startall (count, rawMpiRequests.getRawPtr());
```
mhoemmen: Why wouldn't we start each send as it's processed?  Why do we start them all at once?

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -1339,6 +1368,100 @@ namespace Tpetra {
...
+        if (err != MPI_SUCCESS) {
+          std::ostringstream os;
+          os << "Teuchos::MpiComm::waitAll: MPI_Startall() failed with error \""
```
mhoemmen: The error message looks wrongly labeled.  More importantly, see notes above about encapsulation.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -1419,6 +1545,13 @@ namespace Tpetra {
         "The \"send buffer\" code path doesn't currently work with "
         "nonblocking sends.");
 
+      // FIXME (CAG 01 April 2019) This does not work for Persistent.
```
mhoemmen: Why not?  Is this because persistent sends are nonblocking, and thus share the same issues as the `MPI_Isend` code path?

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -1722,6 +1855,107 @@ namespace Tpetra {
 
       // Data are already blocked (laid out) by process, so we don't
       // need a separate send buffer (besides the exports array).
+
+      if (sendType == Details::DISTRIBUTOR_PERSISTENT) {
+
+        if (persistentRequests_.size() == 0) {
+
+          Teuchos::RCP<const Teuchos::MpiComm<int> > tmpic = Teuchos::rcp_dynamic_cast<const Teuchos::MpiComm<int> >(comm_);
```
mhoemmen: See notes above on encapsulation.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -2306,6 +2549,104 @@ namespace Tpetra {
         *out_ << os.str ();
       }
 
+      if (sendType == Details::DISTRIBUTOR_PERSISTENT) {
+
+        if (persistentRequests_.size() == 0) {
+
+          typedef typename ExpView::non_const_value_type Packet;
```
mhoemmen: Prefer C++11 `using` statements:
```c++
using packet_type = typename ExpView::non_const_value_type;
```

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -2306,6 +2549,104 @@ namespace Tpetra {
...
+          // typedef typename ExpView::array_layout Layout;
+          // typedef typename ExpView::device_type Device;
+          // typedef typename ExpView::memory_traits Mem;
+
+          Teuchos::RCP<const Teuchos::MpiComm<int> > tmpic = Teuchos::rcp_dynamic_cast<const Teuchos::MpiComm<int> >(comm_);
```
mhoemmen: See notes above on encapsulation.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -2306,6 +2549,104 @@ namespace Tpetra {
...
+            if (procsFrom_[i] != myRank) {
+              imports_view_type recvBuf =
+                subview_offset (imports, curBufOffset, curBufLen);
+
+              Packet t;
+              MPI_Datatype rawType = Teuchos::Details::MpiTypeTraits<Packet>::getType (t);
+              Packet* rawRecvBuf = recvBuf.data ();
+              const int count = as<int> (recvBuf.size ());
+              MPI_Request rawRequest = MPI_REQUEST_NULL;
+              MPI_Recv_init (rawRecvBuf, count, rawType, procsFrom_[i], tag, rawComm, &rawRequest);
```
mhoemmen: We have Teuchos comm wrappers for `Kokkos::View`; we should extend those to support `MPI_Recv_init` etc.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -2306,6 +2549,104 @@ namespace Tpetra {
...
+
+              Teuchos::RCP<Teuchos::CommRequest<int> > req = Teuchos::mpiCommRequest<int>(rawRequest, count);
+              // Teuchos::set_extra_data( globalRecvBuf, "buffer", inOutArg(req) );
```
mhoemmen: The encapsulation should be able to take a buffer.  Remember that you can always wrap a `Kokkos::View` in the custom deleter for a `Teuchos::ArrayRCP`.  We have helper functions for that.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -2306,6 +2549,104 @@ namespace Tpetra {
...
+        Array<MPI_Request> rawMpiRequests (count, MPI_REQUEST_NULL);
+        int k = 0;
+        for (int i = 0; i<persistentRequests_.size(); i++) {
+          Teuchos::RCP<Teuchos::CommRequest<int> > request = persistentRequests_[i];
+          Teuchos::RCP<Teuchos::MpiCommRequestBase<int> > mpiRequest =
+            Teuchos::rcp_dynamic_cast<Teuchos::MpiCommRequestBase<int> > (request);
+          rawMpiRequests[k] = mpiRequest->getRawMpiRequest();
+          k++;
+        }
+        const int err = MPI_Startall (count, rawMpiRequests.getRawPtr());
```
mhoemmen: See comment above.  Why do we use `MPI_Startall` instead of starting each message separately?

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -2723,6 +3074,112 @@ namespace Tpetra {
 
       // Data are already blocked (laid out) by process, so we don't
       // need a separate send buffer (besides the exports array).
+
```
mhoemmen: See notes above.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -2723,6 +3074,112 @@ namespace Tpetra {
...
+        }
+
+        if (verbose_) {
+          std::ostringstream os;
+          os << "Proc " << myProcID
+             << ": doPosts(4 args, Kokkos, fast): Starting persistent requests" << endl;
+          *out_ << os.str ();
+        }
+        int count = persistentRequests_.size();
+        Array<MPI_Request> rawMpiRequests (count, MPI_REQUEST_NULL);
```
mhoemmen: See notes above.  We have a `Teuchos::waitAll`; a `Teuchos::startAll` would be the next logical step.

packages/tpetra/core/src/Tpetra_Distributor.cpp
cgcgcg: As far as I can see, 'waitAll' actually frees the requests. I can add an optional parameter to take care of that. And add a 'startAll' function.

packages/tpetra/core/src/Tpetra_Distributor.hpp
cgcgcg: It seemed cleaner to me, and maybe there could be some MPI optimization benefits?

mhoemmen:
@cgcgcg Thanks for working on this! :-D  I think it's a great idea.  See also the following Tpetra issues:

  - #3580
  - #383 
  - #608 
  - #767

packages/tpetra/core/src/Tpetra_Distributor.hpp
cgcgcg: Yep.

packages/teuchos/comm/src/Teuchos_Comm.hpp
```diff
@@ -662,6 +718,34 @@ class Comm : virtual public Describable {
...
+  startAll(const ArrayView<RCP<CommRequest<Ordinal> > > &requests) const = 0;
+
+  /// \brief Free a single communication request.
+  ///
+  /// \param request [in/out] On input: request is not null, and
+  /// <tt>*request</tt> is either null (in which case this function
+  /// does nothing and returns null) or an RCP of a valid CommRequest
+  /// instance representing an communication request.
+  virtual void
+  free (const Ptr<RCP<CommRequest<Ordinal> > >& request) const = 0;
```
mhoemmen: Do you actually need an explicit `free` method?  CommRequest objects know how to free themselves already.

packages/tpetra/core/src/Tpetra_Distributor.hpp
mhoemmen: Ah, sorry, never mind.  (The idea is that you have the requests all set up already, so of course you want to start them all at once.)

packages/teuchos/comm/src/Teuchos_Comm.hpp
cgcgcg: The `MpiCommRequestBase` has a destructor that calls `MPI_Cancel`, not `MPI_Request_free`. I think that is is reasonable in the case of persistent requests to ask the user to explicitly free them..

kddevin:
@trilinos/tpetra @cgcgcg 
We would like to know how this performs before it is merged to develop.  Is that feasible?

cgcgcg:
@kddevin I am running MiniEM on Cori, but I am getting a lot of noise. Some runs look very good, other runs show no improvement over standard Send-Ireceive. I'll keep investigating. We should definitely not make this the default before it has been thoroughly tested!

packages/teuchos/comm/src/Teuchos_Comm.hpp
mhoemmen: @cgcgcg Then we should fix the destructor :-)

mhoemmen:
@cgcgcg This actually passed -- how do you feel about this?  Do you want to keep working on it?

cgcgcg:
@mhoemmen I'm not surprised, we did not change the default behavior, so it should pass. I want to get some timings before pushing it in..

mhoemmen:
@cgcgcg Cool -- thanks for the status report!  I'll mark this WIP then.

mhoemmen:
@cgcgcg did you ever end up building a nice abstraction for this or did you just hack it in?  only curious ; - )

cgcgcg:
I mirrored Tpetra's current design :-P

I'll have to go over this PR a bit more and clean up, but at last it was kinda useful: #9159

mhoemmen:
@cgcgcg haha you totally hacked it in, ah well ; - P

mhoemmen:
lol why did you copy my FIXME comments from six years ago? ; - )

mhoemmen:
All of the four issues mentioned above were "closed to due lack of activity."  None of them were actually implemented.

PR #4800, opened 2019-04-03
Change PR testing from 2.1G to 3G per core

prwolfe:
This is hopefully temporary as we know it will slow down
the PR testing - particulary in the moderate case where
smaller subsets are being built. This is needed due to
the large size of some MueLu objects and libraries when
building all of Trilinos. Later efforts should place a
ulimit on compile processes to keep this kind of
expansion in check.

## Related Issues

* Related to #4696


william76:
update: On our call earlier, I reduced the swim lanes for PR's from 5 to 4.  Waiting on the currently running autotester drivers to finish before restarting instances 0-3.

Issue #4821, opened 2019-04-05
Tempus: Add method to set Tableaus.


ccober6:
@trilinos/tempus 

## Motivation and Context
To better setup RK methods, it is nice to set the Butcher Tableau directly instead of through a ParameterList.

## Solution
Add set method for the tableaus.

PR #4824, opened 2019-04-05
Tempus: Add method to set Tableaus.

ccober6:
To better setup RK methods, it is nice to set the Butcher Tableau
directly instead of through a ParameterList.  Added set method for
the tableaus for both ExplicitRK and DIRK.  The IMEX and IMEX
partitioned already have these methods.

Thanks to Kris Beckwith for pointing out these missing methods.

 #4821

@trilinos/tempus 

## Checklist

### DIFF
--- a/packages/tempus/src/Tempus_StepperDIRK_decl.hpp
+++ b/packages/tempus/src/Tempus_StepperDIRK_decl.hpp
@@ -122,6 +122,9 @@ public:
     virtual void setTableau(
       Teuchos::RCP<Teuchos::ParameterList> pList = Teuchos::null);
 
+    virtual void setTableau(
+      Teuchos::RCP<const RKButcherTableau<Scalar> > DIRK_ButcherTableau);
+
     /// Initialize during construction and after changing input parameters.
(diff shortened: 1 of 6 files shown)


krcb:
@ccober6 This looks good. One question: is it worth making the setTableau methods that create instances of the ERKButcherTableau utilize the new setTableau method (which accepts an ERKButcherTableau)? This would help to ensure that the new method is covered in testing.
One other thing that I wasn't clear on is whether there are an implications to replacing the tableau after the ERK instance is created (e.g. related to change the number of stages)? Maybe we should create a test that demonstrates this?

ccober6:
Morning @krcb.  Yeah I also thought about running all the setTableaus through the new one after committing.  I will do this and restart this.

On the second point, the general procedure is to create a stepper, make any changes to the stepper (e.g., setTableau) and finally call initialize when changes are complete.  This procedure is currently tested, and should cover is this basic usage.

ccober6:
@krcb, here are the changes, can you please review them?  Thanks!

Issue #4196, opened 2019-01-15
Trilinos (Tpetra & downstream) no longer build with Scalar=complex<float>


mhoemmen:
@trilinos/tpetra @trilinos/amesos2

I found this out when testing Amesos2.  See e.g., #4194.

## Current Behavior

Tpetra and downstream classes do not build with `Scalar=complex<float>`.  I get errors of the form discussed here: https://github.com/kokkos/kokkos/issues/1964 .  The case `Scalar=Kokkos::complex<double>` still works, because it uses a different overload of these Kokkos functions that does not attempt to put the output type in a union with `unsigned long long int`.

## Motivation and Context

Sierra currently enables `complex<float>` in its Trilinos build.  This error also showed up in #4194, since Amesos2 was incorrectly attempting to build with `Scalar=complex<float>`, even though it was not enabled in Tpetra.

## Possible Solution

I am working on a work-around, namely to overload `Kokkos::Impl::atomic_fetch_oper` and `Kokkos::Impl::atomic_oper_fetch` specifically for `Kokkos::complex<float>`.

## Steps to Reproduce

Set the CMake option `Trilinos_ENABLE_COMPLEX_FLOAT:BOOL=ON`.

## Related Issues
* Related to #4194

mhoemmen:
- I polled Salinas developers; they report that they almost certainly do not need `complex<float>`.
  - Salinas does actually _use_ `complex<double>`, but @crdohrm could say whether Salinas uses any Trilinos features that depend on this being enabled in Trilinos.
  - @theguruat12 is currently testing Sierra with various combinations of Trilinos complex options.

mhoemmen:
@dsunder is working on a fix in Kokkos.  See discussion here: https://github.com/kokkos/kokkos/issues/1964 .  For now, I plan on just submitting fixes to Tpetra's tests (e.g., `42.0*z` doesn't compile if `z` is `complex<float>`; that breaks some tests).  I don't want code to compile unless I know that it works, so I'd rather not submit the compilation work-around for now.

crdohrm:
Hi guys, Salinas uses Tpetra with complex matrices for frequency domain analyses.

crdohrm:
Oops, hit the wrong button. Will Reopen and comment.

mhoemmen:
@crdohrm wrote:
> Hi guys, Salinas uses Tpetra with complex matrices for frequency domain analyses.

`complex<double>`, `complex<float>`, or both?

[...]

Issue #3574, opened 2018-10-07
Ifpack2 + Kokkos + complex float : compilation error


davydden:
## Expectations

Trilinos `trilinos-release-12-14-branch` builds  with ETI and complex and float types

## Current Behavior
```
8051    In file included from /Users/davydden/spack/var/spack/stage/trilinos-12.14-hpmbucm6pphwdcf3p6hhqcac2cm5qts6/Trilinos/spack-build/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_Serial.cpp:50:
     8052    In file included from /Users/davydden/spack/var/spack/stage/trilinos-12.14-hpmbucm6pphwdcf3p6hhqcac2cm5qts6/Trilinos/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_def.hpp:52:
  >> 8053    /Users/davydden/spack/var/spack/stage/trilinos-12.14-hpmbucm6pphwdcf3p6hhqcac2cm5qts6/Trilinos/packages/kokkos-kernels/src/batched/KokkosBatched_Util.hpp:195:7: error: static_assert failed "KokkosKernels:: Invalid SIMD<> type."
     8054          static_assert( std::is_same<T,bool>::value                     ||
     8055          ^              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     8056    /Users/davydden/spack/var/spack/stage/trilinos-12.14-hpmbucm6pphwdcf3p6hhqcac2cm5qts6/Trilinos/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp:1568:28: note: in instantiation of template class 'KokkosBatched::Experimental::SIMD<Kokkos::complex<floa
             t> >' requested here
     8057                B.assign_data( &vector_values(i0+1,0,0) );
     8058                               ^
     8059    /Users/davydden/spack/var/spack/stage/trilinos-12.14-hpmbucm6pphwdcf3p6hhqcac2cm5qts6/Trilinos/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp:1651:9: note: in instantiation of member function 'Ifpack2::BlockTriDiContainerDetails::ExtractAndFactori
[...]

PR #4828, opened 2019-04-08
Tpetra: Fix #4827; make more Tpetra classes follow Rule of Five

### LINKED ISSUE #4827: Tpetra::Details::copyConvert doesn't build for conversions between Kokkos::complex<float> and double
@trilinos/tpetra 

This breaks `Tpetra::deep_copy` between `MultiVector<std::complex<float>, ...>` and `MultiVector<double, ...>`.

mhoemmen:
@trilinos/tpetra 

## Description

1. Fix #4827.
2. Make more Tpetra classes follow the Rule of Five (C++ Core Guidelines C.21).
3. Clean up FEMultiVector code, syntax, and documentation.

## Related Issues

* Closes #4827 
* Related to #4196, #3946, #3574

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp
@@ -652,7 +652,8 @@ namespace Tpetra {
     explicit CrsMatrix (const Teuchos::RCP<const crs_graph_type>& graph,
                         const Teuchos::RCP<Teuchos::ParameterList>& params = Teuchos::null);
 
-    /// \brief Constructor specifying a previously constructed graph and entries array
+    /// \brief Constructor specifying a previously constructed graph
+    ///   and entries array
     ///
     /// Calling this constructor fixes the graph structure of the
     /// sparse matrix.  We say in this case that the matrix has a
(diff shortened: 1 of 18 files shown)


mhoemmen:
Looks legit -- will need to fix something for Stokhos.

mhoemmen:
```
.../Trilinos/packages/tpetra/core/src/Tpetra_Details_copyConvert.hpp:247:43: error: comparison between ‘enum Kokkos::View<Sacado::UQ::PCE<Stokhos::DynamicStorage<int, double, Kokkos::Serial> >*, Kokkos::LayoutRight, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<0> >::<unnamed>’ and ‘enum Kokkos::View<const Sacado::UQ::PCE<Stokhos::DynamicStorage<int, double, Kokkos::Serial> >*, Kokkos::LayoutRight, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<0> >::<unnamed>’ [-Werror=enum-compare]
       static_assert (OutputViewType::Rank == InputViewType::Rank,
                      ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
```
@etphipp It's interesting that `Rank` is an `int` in all the other `Kokkos::View` specializations.  Should the Sacado etc. specializations behave the same way?

mhoemmen:
Hm, this is interesting: https://testing.sandia.gov/cdash/testDetails.php?test=74057758&build=4851178 .  I can `Kokkos::deep_copy` from LayoutLeft to LayoutRight on the same memory space, but not across memory spaces:
```
 OutputDeviceType: HostSpace
 InputDeviceType: CudaSpace
  OutputArrayLayout: Left
  InputArrayLayout: Left
   OutputValueType: int
   InputValueType: int
   OutputValueType: long long
   InputValueType: int
   OutputValueType: int
   InputValueType: long long
   OutputValueType: Kokkos::complex<float>
   InputValueType: double
   OutputValueType: double
   InputValueType: Kokkos::complex<float>
  OutputArrayLayout: Right
  InputArrayLayout: Left
   OutputValueType: int
   InputValueType: int

p=0: *** Caught standard std::exception of type 'std::runtime_error' :
 
  Error: Kokkos::deep_copy with no available copy mechanism: x to y
```

etphipp:
> @etphipp It's interesting that `Rank` is an `int` in all the other `Kokkos::View` specializations. Should the Sacado etc. specializations behave the same way? I don't think this will be hard for me to work around in Tpetra, so there's no urgency.

I'm curious why you say this. From [here](https://github.com/trilinos/Trilinos/blob/ce172471817489c5eb1bdddce8a1948887a6c03f/packages/kokkos/core/src/Kokkos_View.hpp#L610), `Kokkos::View::Rank` is always an enum:
```
  enum { Rank = map_type::Rank };
```
Furthermore, Sacado/Stokhos don't actually specialize `Kokkos::View`, but instead an internal class called `Kokkos::ViewMapping` (which becomes `map_type` in the above).

mhoemmen:
@trilinos/tpetra this should be ready for review now.

packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp
```diff
@@ -1048,12 +1049,58 @@ namespace Tpetra {
...
+    /// \brief Destructor (virtual for memory safety of derived classes).
+    ///
+    /// \note To Tpetra developers: See the C++ Core Guidelines C.21
+    ///   ("If you define or <tt>=delete</tt> any default operation,
+    ///   define or <tt>=delete</tt> them all"), in particular the
+    ///   AbstractBase example, for why this destructor declaration
+    ///   implies that we need the above four <tt>=delete</tt> (or
+    ///   <tt>=default</tt>) declarations for copy construction, move
+    ///   construction, copy assignment, and move assignment.
+    virtual ~CrsMatrix () = default;
```
lucbv: It does seem odd to have the default destructor and not the default constructor... but I understand the historical argument.

mhoemmen: @lucbv https://en.cppreference.com/w/cpp/language/default_constructor#Implicitly-defined_default_constructor :
> If the implicitly-declared default constructor is not defined as deleted, it is defined (that is, a function body is generated and compiled) by the compiler if odr-used, and it has exactly the same effect as a user-defined constructor with empty body and empty initializer list. That is, it calls the default constructors of the bases and of the non-static members of this class.

mhoemmen: @lucbv The issue with CrsMatrix is that the copy constructor and copy assignment operator were always declared private and not implemented.  This is equivalent to C++11 `=delete`.  I have no idea what motivated random pre-2010 Tpetra developers to declare them that way, but I feel like fixing that should be a separate PR ;-) .

mhoemmen:
@trilinos/tpetra The above is interesting, because it suggests two things:

1. We weren't using `Kokkos::deep_copy` before, when we could have.
2. We might have been doing the wrong thing before.

@hkthorn @trilinos/anasazi this could actually be a bug.

csiefer2:
"Rule of Five" sounds like something out of Tolkien...

mhoemmen:
> "Rule of Five" sounds like something out of Tolkien...

Best is Rule of Zero i mean they didn't even read a scroll of Detect Curse first

mhoemmen:
I'm working on https://github.com/trilinos/Trilinos/pull/4871 today, but I also want to fix this.

mhoemmen:
@lucbv Thanks for your review!  This PR is superseded by https://github.com/trilinos/Trilinos/pull/4886.  Could you please review that one?  Thanks! :-D

Issue #4836, opened 2019-04-09
Tempus: Improve Interval Output Setup


ccober6:
In the output interval and the screen output interval, all the intervals are placed in a vector along with the specified output times or indices, and then sorted.  This allowed all the output times and indices to be in one vector and searched.  However when the number of time steps reaches millions, the sort is far too expensive.  This will rework the interval specification so that it is not part of the vector or sorting.

@trilinos/tempus 

## Expectations
Work the same as previously but without the costly sorting.

PR #4837, opened 2019-04-09
Tempus: improve interval output setup

ccober6:
In the output interval and the screen output interval, all the
intervals are placed in a vector along with the specified output
times or indices, and then sorted. This allowed all the output times
and indices to be in one vector and searched. However when the
number of time steps reaches millions, the sort is far too expensive.
This reworks the interval specification so that it is not part
of the vector or sorting.

 #4836

@trilinos/tempus 

## How Has This Been Tested?
Passes all tests on Mac.

## Checklist

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
@@ -498,6 +498,11 @@ void IntegratorBasic<Scalar>::startTimeStep()
     wsmd->setOutputScreen(false);
   else
     wsmd->setOutputScreen(true);
+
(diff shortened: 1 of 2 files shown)


ccober6:
@krcb I believe this is the source of your slowdown as I found it running some of Tempus' tests.  I did not check against sod_shock_limited_tempus, but can you try it out?

krcb:
@ccober6 This is going to take a little bit of time for me to get tested with the appropriate code base. Even if this *isn't* the source of the slowdown that I'm seeing, I think the changes here are a good improvement, so I'm going to approve this so that we can get it moving through the pipeline.

PR #4846, opened 2019-04-09
Xpetra:  Remove Unused Parameter Warnings

jmgate:
@trilinos/xpetra 

## Description
`-Wunused-parameter` warnings.

## Motivation and Context
Clean build.

### DIFF
--- a/packages/xpetra/src/BlockedMultiVector/Xpetra_BlockedMultiVector.hpp
+++ b/packages/xpetra/src/BlockedMultiVector/Xpetra_BlockedMultiVector.hpp
@@ -269,22 +269,22 @@ namespace Xpetra {
     //@{
 
     //! Replace value, using global (row) index.
-    virtual void replaceGlobalValue(GlobalOrdinal globalRow, size_t vectorIndex, const Scalar &value) {
+    virtual void replaceGlobalValue(GlobalOrdinal /* globalRow */, size_t /* vectorIndex */, const Scalar &/* value */) {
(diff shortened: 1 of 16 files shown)


jmgate:
@jhux2, I'm not setting the "AT: AUTOMERGE" label because a handful of these files are marked as automatically generated.  Should I look into modifying whatever it is that generates them, or would that be something you'd prefer to do?

csiefer2:
@jmgate The auto-generation stuff hasn't worked for years.  Don't worry about it :)

packages/xpetra/src/BlockedMultiVector/Xpetra_BlockedMultiVector.hpp
```diff
@@ -742,6 +742,7 @@ namespace Xpetra {
       XPETRA_MONITOR("BlockedMultiVector::getMultiVector(r,bThyraMode)");
       TEUCHOS_TEST_FOR_EXCEPTION(r > map_->getNumMaps(), std::out_of_range, "Error, r = " << r << " is too big. The BlockedMultiVector only contains " << map_->getNumMaps() << " partial blocks.");
       XPETRA_TEST_FOR_EXCEPTION(map_->getThyraMode() != bThyraMode, Xpetra::Exceptions::RuntimeError, "BlockedMultiVector::getMultiVector: inconsistent Thyra mode");
+      (void)bThyraMode; // avoid unused parameter warning when HAVE_XPETRA_DEBUG isn't defined
```
jmgate: Note that I'm not thrilled about this solution here, but the `#ifdef` that hides the usage of `bThyraMode` is buried within `XPETRA_TEST_FOR_EXCEPTION`, so I'm not sure what else to do.

PR #4848, opened 2019-04-09
Tpetra: Deprecate MultiVectorFiller

csiefer2:
Auto-PR for SHA 0447543

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp
@@ -1,5 +1,6 @@
-#include <Tpetra_MultiVectorFiller.hpp>
+#include "TpetraCore_config.h"
 #ifdef TPETRA_ENABLE_DEPRECATED_CODE
+#include "Tpetra_MultiVectorFiller.hpp"
 namespace Tpetra {
   namespace Test {
 
--- a/packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp
@@ -1,5 +1,5 @@
 #include <Tpetra_MultiVectorFiller.hpp>
-
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
 namespace Tpetra {
   namespace Test {
 
@@ -68,3 +68,4 @@ namespace Tpetra {
 
   } // namespace Test
 } // namespace Tpetra
+#endif // TPETRA_ENABLE_DEPRECATED_CODE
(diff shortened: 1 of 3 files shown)


packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp
```diff
@@ -1,5 +1,5 @@
 #include <Tpetra_MultiVectorFiller.hpp>
```
mhoemmen: We should actually consider the header file itself deprecated, as well as its contents.  (We could put a `#warning` in the header file, so users who include it see the warning, even if they don't use its contents.)  This means that the .cpp file would still exist, but would look like this:
```c++
#include "TpetraCore_config.h"
#include "Tpetra_MultiVectorFiller.hpp"
// ... rest of contents of file ...
#endif 
```

packages/tpetra/core/src/Tpetra_MultiVectorFiller.hpp
```diff
@@ -68,9 +69,12 @@ namespace Details {
   // newEntries into allEntries and make the results unique.  (This is
   // cheaper than sorting the whole array.)
   //
+  // \warning DO NOT CALL THIS METHOD!  This method is DEPRECATED
```
mhoemmen: If we deprecate the helper methods, then MultiVectorFiller itself will generate spurious build warnings.  Best practice is only to deprecate user-facing stuff.

packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp
csiefer2: For the header:
```

#ifndef __Tpetra_MultiVectorFiller_hpp
#define __Tpetra_MultiVectorFiller_hpp

#include "Tpetra_MultiVector.hpp"
#warning "... this header is deprecated."
...
#else
 

#error "... this header is deprecated"
#endif 
'''

Comment (csiefer2) [reply packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp]: Is that what you want?

Comment (mhoemmen) [reply packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp]: @csiefer2 Hm, I realized that it may not be possible to turn off `#warning` warnings, or at least that this would require a different build option than the option to turn off deprecated warnings.  Perhaps we should then omit the `#warning` pragma.  Also, the first header file to include should be `"TpetraCore_config.h"`.  It is low cost to include that and nothing else, if deprecated code is off.

Comment (csiefer2) [reply packages/tpetra/core/src/Tpetra_MultiVectorFiller.cpp]: So

```
#ifndef __Tpetra_MultiVectorFiller_hpp
#define __Tpetra_MultiVectorFiller_hpp

#include "Tpetra_MultiVector.hpp"
...
#else
#error "... this header is deprecated"
#endif 
'''

packages/tpetra/core/src/Tpetra_MultiVectorFiller.hpp
csiefer2: Easy enough to fix.

PR #4863, opened 2019-04-11
Xpetra: fix issue #4038.

jhux2:
Auto-PR for SHA 66aff72

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/4863)

### DIFF
--- a/packages/xpetra/src/Map/Xpetra_EpetraMap.hpp
+++ b/packages/xpetra/src/Map/Xpetra_EpetraMap.hpp
@@ -249,9 +249,7 @@ namespace Xpetra {
     using local_map_type = typename Map<LocalOrdinal, GlobalOrdinal, Node>::local_map_type;
     /// \brief Get the local Map for Kokkos kernels.
     local_map_type getLocalMap () const {
-      TEUCHOS_TEST_FOR_EXCEPTION(true, Xpetra::Exceptions::RuntimeError,
-        "Xpetra::EpetraMap only available for GO=int or GO=long long with EpetraNode (Serial or OpenMP depending on configuration)");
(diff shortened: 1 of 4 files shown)


packages/xpetra/src/Map/Xpetra_EpetraMap.hpp
```diff
@@ -249,9 +249,7 @@ namespace Xpetra {
     using local_map_type = typename Map<LocalOrdinal, GlobalOrdinal, Node>::local_map_type;
     /// \brief Get the local Map for Kokkos kernels.
     local_map_type getLocalMap () const {
-      TEUCHOS_TEST_FOR_EXCEPTION(true, Xpetra::Exceptions::RuntimeError,
-        "Xpetra::EpetraMap only available for GO=int or GO=long long with EpetraNode (Serial or OpenMP depending on configuration)");
-      TEUCHOS_UNREACHABLE_RETURN(local_map_type());
+      throw std::runtime_error("Xpetra::EpetraMap::getLocalMap is not implemented.");
```
bartlettroscoe: Why not using TEUCHOS_TEST_FOR_EXCPETION()?  That allows for setting a breakpoint before the thrown even with debuggers that don't allow you to set the breakpoint on throws.  Also, note that when binutils support is enabled in Teuchos, this can produce a readable stacktrace outside of a debugger.

jhux2: I could, but this is a code path that we don't plan to support, and I don't see any of the ATDM codes ever hitting this.

jhux2:
@trilinos/framework Any ideas why this is failing?

PR #4865, opened 2019-04-11
MiniEM: Fix discrete gradient

cgcgcg:
@trilinos/panzer 

## Description
The discrete gradient matrix used in MiniEM was not correctly built.


dridzal:
@cgcgcg , would this have affected performance on unstructured meshes with large variations in cell size (1:100 ratios)?

cgcgcg:
@dridzal Possibly :-) We found this because we started running meshes with bigger ratios. It only affected the Tpetra code path, however.

dridzal:
@cgcgcg , yes, I recall problems when running in the Tpetra code path.  Since then, we've been using the Epetra code path, but mostly on uniform meshes.  I'll check both code paths again with large-ratio meshes.  I have a few examples with very challenging geometries.

Issue #4728, opened 2019-03-26
MueLu: deprecate code in preparation for next release


jhux2:
@trilinos/muelu 

If there are features that we want to deprecate in MueLu, we should do so before the next minor release, which is around April 15.

lucbv:
@jhux2 @csiefer2 @cgcgcg @rstumin 
I would like to deprecate a bunch of the: CreateX/T/EpetraPreconditioner() interfaces in order to keep only:
```
CreateXpetraPreconditioner(Teuchos::RCP<Xpetra::Matrix<Scalar,LocalOrdinal,GlobalOrdinal,Node> > op, const Teuchos::ParameterList& inParamList)
CreateXpetraPreconditioner(Teuchos::RCP<Xpetra::Matrix<Scalar,LocalOrdinal,GlobalOrdinal,Node> > op, const std::string& xmlFileName)
```

We would then ask users to put their coordinates and nullspace on the parameter list directly even if that is the only thing that is on said list.

Any objections or comments?

cgcgcg:
@lucbv RefMaxwell is currently passing coordinates as argument, but that's simple to fix.

lucbv:
@cgcgcg this would just be a deprecation, you would have time to adapt to the new interface.

jhux2:
> I would like to deprecate a bunch of the: CreateX/T/EpetraPreconditioner() interfaces

I approve!

lucbv:
@jhux2 I think this can be closed now?

PR #4869, opened 2019-04-12
MueLu: deprecate adapters

### LINKED ISSUE #4868: MueLu: deprecate some adapters
@trilinos/muelu 

## Expectations
The interface to create a preconditioner using MueLu should be simple and uniform for all users.

## Current Behavior
nullspace multivector and coordinates multivectors can be passed optionally to the preconditioner constructor. These inputs could actually be passed just as easily on the input parameter list as it is the case of other user defined parameters. This makes it awkward to pass them with additional user data.

## Motivation and Context
With the addition of the "user data" sublist to the MueLu parameter list, users have a unique and general way to provide data to MueLu. It would make sense to use this as the only mean of passing data to MueLu instead of maintaining multiple interfaces.

## Definition of Done
A single interface to MueLu exists and it allows user to pass any kind of data necessary.

## Possible Solution
Deprecate the special interfaces used to pass nullspace and coordinates.

## Related Issues
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of #4728 
* Composed of

lucbv:
@trilinos/muelu 

## Description
Deprecating all CreateT/E/XpetraConstructor except for one accepting an operator and a parameter list

## Motivation and Context
This will give more uniformity and reduce maintenance

## Related Issues

* Closes #4868
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of #4728
* Composed of 

## How Has This Been Tested?
So far this is a work in progress but I plan on testing it with a local build of MueLu, all tests and examples should pass and compile without warnings.

## Checklist

### DIFF
--- a/packages/muelu/CMakeLists.txt
+++ b/packages/muelu/CMakeLists.txt
@@ -57,6 +57,33 @@ ENDIF()
 TRIBITS_ADD_DEBUG_OPTION()
 TRIBITS_ADD_SHOW_DEPRECATED_WARNINGS_OPTION()
 
+TRIBITS_ADD_OPTION_AND_DEFINE(
+  ${PACKAGE_NAME}_ENABLE_DEPRECATED_CODE
+  HAVE_${PACKAGE_NAME_UC}_DEPRECATED_CODE
+  "Whether MueLu enables deprecated code (that is, anything marked with the MUELU_DEPRECATED macro) at compile time.  Default is ON (deprecated code enabled).  If OFF, then deprecated code does not exist."
+  YES
+)
+IF (HAVE_${PACKAGE_NAME_UC}_DEPRECATED_CODE)
+  MESSAGE(STATUS "MueLu: Enabling deprecated code")
+ELSE ()
+  MESSAGE(STATUS "MueLu: Disabling deprecated code")
+ENDIF ()
+
+IF (NOT ${PACKAGE_NAME}_ENABLE_DEPRECATED_CODE AND ${PACKAGE_NAME}_ENABLE_DEPRECATED_TESTS)
+  MESSAGE(FATAL_ERROR "MueLu: You cannot enable deprecated tests when deprecated codes is disabled!")
+ENDIF ()
+
+TRIBITS_ADD_OPTION_AND_DEFINE(
+  ${PACKAGE_NAME}_ENABLE_DEPRECATED_TESTS
+  HAVE_${PACKAGE_NAME_UC}_DEPRECATED_TESTS
+  "Whether MueLu compiles and run tests (or areas of a test) that contain deprecated code (that is, anything marked with the MUELU_DEPRECATED macro).  Default is OFF (deprecated tests disabled).  If ON, then deprecated tests are compiled and tested."
+  NO
+)
+
+IF (HAVE_${PACKAGE_NAME_UC}_DEPRECATED_TESTS)
+  MESSAGE(STATUS "MueLu: Enabling deprecated tests")
+ENDIF ()
+
 IF (${PACKAGE_NAME}_ENABLE_Epetra AND NOT ${PACKAGE_NAME}_ENABLE_EpetraExt)
   MESSAGE(FATAL_ERROR "You have enabled Epetra, but not EpetraExt. MueLu requires that either both are enabled, or both are disabled. Please either disable Epetra, or enable EpetraExt.")
 ENDIF()
--- a/packages/muelu/test/unit_tests/Adapters/AmgxOperatorAdapter.cpp
+++ b/packages/muelu/test/unit_tests/Adapters/AmgxOperatorAdapter.cpp
@@ -92,35 +92,35 @@ namespace MueLuTests {
       RCP<const Teuchos::Comm<int> > comm = TestHelpers::Parameters::getDefaultComm();
       int nx;
       //disable amgx test in parallel
-      if(comm->getSize() > 1) 
+      if(comm->getSize() > 1)
 	return;
       else
 	nx = 91;
 
       //matrix
-      RCP<Matrix> Op = TestHelpers::TestFactory<Scalar, LocalOrdinal, GlobalOrdinal, Node>::Build2DPoisson(nx, -1, Xpetra::UseTpetra); 
+      RCP<Matrix> Op = TestHelpers::TestFactory<Scalar, LocalOrdinal, GlobalOrdinal, Node>::Build2DPoisson(nx, -1, Xpetra::UseTpetra);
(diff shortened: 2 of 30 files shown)


packages/muelu/test/unit_tests/Adapters/AmgxOperatorAdapter.cpp
```diff
@@ -92,35 +92,35 @@ namespace MueLuTests {
...
+
       RCP<AMGXOperator> aH = Teuchos::rcp_dynamic_cast<AMGXOperator>(tH);
       TEST_EQUALITY(aH->sizeA()==nx*nx/comm->getSize(), true);
-   
+
       RCP<MultiVector> RHS = MultiVectorFactory::Build(Op->getRowMap(), 1);
       RCP<MultiVector> X   = MultiVectorFactory::Build(Op->getRowMap(), 1);
 
       //RHS=1, zero initial guess
       RHS->putScalar( (double) 1.0);
```
mhoemmen: You don't need to cast to `double` here.  Check out the following example: https://godbolt.org/z/x1huHU
```c++
#include <complex>
#include <type_traits>

int main () {
    std::complex<float> x ((double) 1.0, (double) 2.0);
    std::complex<float> y (1.0, 2.0);
    static_assert (std::is_same<decltype (1.0), double>::value, "uh oh");
    return 0;
}
```
Also, prefer C++ - style casts: either `double (1.0)` or `static_cast<double> (1.0)`.  (You don't need the cast here, but prefer these styles of cast in general.)

jhux2:
> Both Kokkos and Tpetra have a CMake option and a macro that actually disables deprecated code. For example:
> 
> ```c++
> #ifdef TPETRA_ENABLE_DEPRECATED_CODE
> TPETRA_DEPRECATED void some_deprecated_function ();
> #endif // TPETRA_ENABLE_DEPRECATED_CODE
> ```
> 
> Would MueLu like something like that too?

@mhoemmen Maybe I've missed it, but isn't this what @lucbv has set up in muelu/CMakeLists.txt?

packages/muelu/CMakeLists.txt
```diff
@@ -57,6 +57,19 @@ ENDIF()
 TRIBITS_ADD_DEBUG_OPTION()
 TRIBITS_ADD_SHOW_DEPRECATED_WARNINGS_OPTION()
 
+TRIBITS_ADD_OPTION_AND_DEFINE(
+  ${PACKAGE_NAME}_ENABLE_DEPRECATED_CODE
+  HAVE_${PACKAGE_NAME_UC}_DEPRECATED_CODE
```
mhoemmen: Thanks for adding this!  Now might be the time to think about making this consistent with Kokkos and Tpetra, and also putting the package name first: `MUELU_ENABLE_DEPRECATED_CODE`.

lucbv: @mhoemmen @jhux2 
I do not know what the historical reason is for choosing `HAVE_${PACKAGE_NAME_UC}_DEPRECATED_CODE` over the Tpetra convention but it is consistent in MueLu at this point so unless we make the decision to rename all our preprocessor variables it would not make sense to change just this one...

@mhoemmen maybe a GitHub issue can be opened to discuss why MueLu uses this naming scheme. @cgcgcg hypothesized that it creates a distinction between CMake variables and actual preprocessor variables.

lucbv:
My guess is the auto-tester won't like our unit-test that checks the deprecated interfaces still work...

packages/muelu/CMakeLists.txt
mhoemmen: I doubt there was a reason other than "that's what Trilinos always did."  MueLu has never bothered much with consistency with other Trilinos packages (see e.g., early debates about method name capitalization) so it's all good ;-)

jhux2: > I do not know what the historical reason is for choosing `HAVE_${PACKAGE_NAME_UC}_DEPRECATED_CODE` over the Tpetra convention but it is consistent in MueLu at this point so unless we make the decision to rename all our preprocessor variables it would not make sense to change just this one...

@lucbv Afaik, this is consistent with what is done throughout Trilinos (see Belos, Epetra, Stokhos, ...).   My understanding is that cmake variables/options start with the package name, preprocessor macros start with `HAVE`.

lucbv:
That or you can also put it on a `level X`, where X is a number, list.

lucbv:
@trilinos/framework I do not see errors on the dashboard corresponding to this build. Can you check if something was not reported for some reason?

prwolfe:
Both this failure in build 3108 and build 3102 for the intel 17.0.1 compiler used in #4896 had the same issue.

> CMake Error at /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/TFW_testing_single_configure_prototype/simple_testing.cmake:118 (ctest_empty_binary_directory):
>   ctest_empty_binary_directory problem removing the binary directory:
>   /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/pull_request_test
> 

This one I have never seen before.  Perhaps @bartlettroscoe has some insight?

bartlettroscoe:
CC: @prwolfe, @lucbv 

> ctest_empty_binary_directory problem removing the binary directory:
/scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/pull_request_test

This usually means that that directory does not contain the CMakeCache.txt file which CTest uses as a sign that you are not pointing to a CMake build directory.  In that past I have had to blow away the entire build directory and then it works.

prwolfe:
> CC: @prwolfe, @lucbv
> 
> > ctest_empty_binary_directory problem removing the binary directory:
> > /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/pull_request_test
> 
> This usually means that that directory does not contain the CMakeCache.txt file which CTest uses as a sign that you are not pointing to a CMake build directory. In that past I have had to blow away the entire build directory and then it works.

So that really means the previous build failed.  I will look and see if I can reset the error after this happens as it is not relevant to this build.

lucbv:
@prwolfe @bartlettroscoe
OK, I will then just wait and see what happens with the next round of testing.
Thanks for looking at it.

mhoemmen:
@prwolfe Thanks for explaining!

prwolfe:
> @prwolfe @bartlettroscoe
> OK, I will then just wait and see what happens with the next round of testing.
> Thanks for looking at it.

That will need a retest flag - I am adding it.

PR #4871, opened 2019-04-12
Tpetra::MultiVector: Fix #4870 (prefer running norms on device)

### LINKED ISSUE #4870: Tpetra::MultiVector::norm*: Prefer running on device
@trilinos/tpetra 

The `Tpetra::MultiVector::norm*` methods currently run where the MultiVector's data were most recently updated.  Reports from @vbrunini running SPARC with vectors with 1.8 million local rows show that this can cause a slowdown, if running on host.  We need to fix these methods so they prefer running on device unless the vector has too few rows to justify running on device.

mhoemmen:
@trilinos/tpetra @vbrunini 

## Description

  - Prefer running MultiVector norms on device, even if it is currently sync'd to host, unless the vector has too few local rows.
  - Factor out the implementation of `Tpetra::MultiVector::norm*` into a separate function, `Tpetra::Details::normImpl`.  I do ETI for that function, when ETI is enabled.  This required adding new ETI macros for iterating over the `(Scalar, Node)` type combination.

## Motivation and Context

  - SPARC performance
  - Build time

## Related Issues

* Closes #4870


mhoemmen:
Argh, I need to do ETI for Stokhos.  I might change the ETI a bit so that it works more like how Christian Trott did it in kokkos-kernels.

mhoemmen:
Superseded by PR https://github.com/trilinos/Trilinos/pull/4881, which has a work-around for the Stokhos link errors experienced in this PR.

PR #4877, opened 2019-04-12
Tpetra: Add deprecations for global ordinal types

william76:
Adding the modifications to the Global Ordinal type handling per the notes from our 2019-04-09 meeting:

### CMakeLists.txt modifications
- IF `DEPRECATED_CODE` is enabled:
  - IF there are > 1 global ordinals set:
    - [x]  Emit a CMake warning that that Tpetra requires only one global ordinal to be set
  - IF there are 0 global ordinal set:
    - [x] Emit a CMake error that Tpetra requires a a global ordinal to be set. 
        Note: This is inconsistent with the meeting notes _but_ the behavior in `Tpetra_Details_DefaultTypes.hpp` will result in a preprocessor error if there is no GO type set so it makes more sense to catch this error at configuration.  I don't think this breaks backwards compatibility because the build would have always failed?
IF `DEPRECATED_CODE` is not enabled:
  - IF there are > 1 global ordinals set:
    - [x]  Emit a CMake error that that Tpetra requires only one global ordinal to be set
  - IF there are 0 global ordinal set:
    - [x] Emit a CMake error that Tpetra requires a a global ordinal to be set.
  
### `Tpetra_Details_DefaultTypes.hpp` modifications
- [x] Change the order of the preprocessor 'switch' statement to move `long long` to the first position to prioritize it over int.  Keep the fail-over `#error` directive if there is no GO set.

Making this a WIP so that @kddevin can see my changes.  There are scaffolding printouts still in the `CMakeLists.txt` file that I'll want to remove before removing the WIP flag.

@trilinos/tpetra


william76:
Quick check done:

| Enable Deprecated Code | GO=int | GO=long long | Result |
| :---: | :---: | :---: | ----- |
| ON | ON | ON    | CMake warning |
| ON | ON | OFF   | No CMake warning |
| ON | OFF | ON   | No CMake warning |
| OFF | ON | ON   | CMake Error |
| OFF | ON | OFF  | No CMake warning |
| OFF | OFF | ON  | No CMake warning |
| ON | OFF | OFF  | CMake Error |
| OFF | OFF | OFF | CMake Error |

Note: I didn't test setting `Tpetra_INST_INT_LONG` because it doesn't really matter once the # of enabled global ordinals is > 1 w/rt to what CMake is doing.

mhoemmen:
@william76 Would the current Sierra use case (GlobalOrdinal=long enabled) work without warnings?

william76:
@mhoemmen I noted in my comment that it should work with `long` is set, I never check anything dealing with the type in the CMake file, just the # of global ordinals set.

https://github.com/trilinos/Trilinos/blob/8f2f2d66a8cb0d860728e663447aecd43cc0268c/packages/tpetra/core/src/CMakeLists.txt#L318-L335

And the header just reorders the preprocessor switch to make `long long` the preferred type, but it's kind of moot in a way since CMake will fail if you pick > 1 once the deprecated stuff is removed:

https://github.com/trilinos/Trilinos/blob/8f2f2d66a8cb0d860728e663447aecd43cc0268c/packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp#L76-L106

I ran a test with GO set to long and it configured just fine.

william76:
@mhoemmen ok, I'll move it over to `pacakges/tpetra/CMakelists.txt`

william76:
@mhoemmen I moved the logic over to `packages/tpetra/CMakeLists.txt`

william76:
@mhoemmen @kddevin 
I checked in the cleaned up version.

william76:
ugh, moving things into `packages/tpetra/CMakeLists.txt` fails somehow, but `packages/tpetra/core/CMakeLists.txt` is fine. I'll have to look at it this afternoon.  So, Still WIP

mhoemmen:
@william76 Be sure to put this after all the logic for picking GlobalOrdinal types is done.

william76:
@mhoemmen I got everything moved over into `packages/tpetra/CMakeLists.txt` and have tested it using the above configurations.  Looks like it's doing the right thing. 

CMake output also notes the validation check:
```
--
-- Tpetra: Validate global ordinal setting ...
-- Tpetra: global ordinal setting is OK
--
```
If an error or warning happens, the "is OK" line is replaced by the appropriate warning or error.

packages/tpetra/CMakeLists.txt
```diff
@@ -1159,6 +1160,39 @@ GLOBAL_SET(${PACKAGE_NAME}_ETI_GORDS   ${${PACKAGE_NAME}_ETI_GORDS})
...
+# Verify that a valid GORDS configuration is set.
+#
+MESSAGE(STATUS "")
+MESSAGE(STATUS "Tpetra: Validate global ordinal setting ...")
+SET(lenTpetraCore_ETI_GORDS 0 INTERNAL)
+LIST(LENGTH ${PACKAGE_NAME}_ETI_GORDS lenTpetraCore_ETI_GORDS)
+
+IF(Tpetra_ENABLE_DEPRECATED_CODE)
+    IF(lenTpetraCore_ETI_GORDS EQUAL 0)
+        MESSAGE(FATAL_ERROR "Tpetra requires a global ordinal type, but none were specified. "
```
mhoemmen: This check makes lines 714-727 (see above) redundant and unnecessary.  You could actually delete that check above in favor of your new check here.  The redundant check above won't break anything, so it's OK to address that in a future PR.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
```diff
@@ -73,20 +73,39 @@ namespace DefaultTypes {
...
-#elif defined(HAVE_TPETRA_INST_INT_LONG)
-  typedef long global_ordinal_type;
-#elif defined(HAVE_TPETRA_INST_INT_UNSIGNED_LONG)
-  typedef unsigned long global_ordinal_type;
-#elif defined(HAVE_TPETRA_INST_INT_UNSIGNED)
-  typedef unsigned global_ordinal_type;
-#else
-#  error "Tpetra: No global ordinal types in the set {int, long long, long, unsigned long, unsigned} have been enabled."
+#if defined(TPETRA_ENABLE_DEPRECATED_CODE)
+    // NEW style picks `int` for Global Ordinal as the first choice.
```
mhoemmen: Not sure what "NEW style" means -- wouldn't this be old style?  Anyway, the comment is redundant (just reading the code should explain which type is picked first), but harmless.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
```diff
@@ -73,20 +73,39 @@ namespace DefaultTypes {
...
+      typedef long global_ordinal_type;
+    #elif defined(HAVE_TPETRA_INST_INT_UNSIGNED_LONG)
+      typedef unsigned long global_ordinal_type;
+    #elif defined(HAVE_TPETRA_INST_INT_UNSIGNED)
+      typedef unsigned global_ordinal_type;
+    #else
+        #error "Tpetra: No global ordinal types in the set {int, long long, long, unsigned long, unsigned} have been enabled."
+    #endif
+#else  // TPETRA_ENABLE_DEPRECATED_CODE IS NOT DEFINED
+    // NEW style picks `long long` for Global Ordinal as the first choice.
```
mhoemmen: The order doesn't actually matter here, since exactly one of these macros is allowed to be defined.

william76: True with the changes to CMake that enforce only 1 allowable GORD to be set, but reordering that switch statement was specifically requested by Karen and the notes from our 4/9 meeting saying to modify the header to "pick long long" as the default was interpreted to mean putting long long first in the if/elseif/else chain.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
william76: meh, sure.  what I get working on my day off in a sleep deprived state :p

william76: we can shoot the comment in the head now or wait till after the major release happens and we nuke all the deprecated stuff.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
mhoemmen: @william76 It's cool, I don't mind :)

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
mhoemmen: @william76 It's totally cool -- we can fix it later (since we'll be touching all these soon again anyway).

PR #4879, opened 2019-04-12
Use Multiple Issue Templates

jmgate:
@trilinos/framework 

## Description
GitHub now supports multiple issue templates.  This commit slims down
our templates in a big way and break them up into four separate ones,
depending on what the issue is.

## Motivation and Context
Our issue template was initially quite large because it was trying to play many roles at once.  Now we can split things up so the templates are easier to use.

## How Has This Been Tested?
I've used multiple templates like this in another GitHub project.

## Screenshots
When creating a new issue, you'll be greeted with this page.
Screen Shot 2019-04-12 at 11 41 43 AM

## Additional Information
Note that I also ripped out any instructions that were apparently only good for Trilinos developers (e.g., @mentioning teams, adding labels, etc.).


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+name: Bug Report
+about: Help us fix something that's broken.
+title: "PackageName:  General Summary of the Bug"
+labels: "type: bug"
+assignees: ''
+
+---
+
+## Bug Report
+@trilinos/PackageName
```
mhoemmen: Should we make this something obviously invalid, like `@trilinos/<PackageName>`?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Description
+What went wrong?  What should have happened?  Do you have an idea what might
```
mhoemmen: I don't think we need any text here.  Most readers will understand what "Description" means.  Also, the first thing most people do when they fill out an issue is delete most of the template -- it's better if we keep it as short as possible, so that readers don't delete something they should fill out.

.github/ISSUE_TEMPLATE/documentation.md
```diff
@@ -0,0 +1,16 @@
...
+name: Documentation
+about: Help us improve our documentation.
+title: "PackageName:  General Summary of the Documentation Needed"
+labels: "impacting: documentation"
+assignees: ''
+
+---
+
+## Documentation
+@trilinos/PackageName
```
mhoemmen: See above.

.github/ISSUE_TEMPLATE/documentation.md
```diff
@@ -0,0 +1,16 @@
...
+
+What would you like to see in the documentation?  How does our current
```
mhoemmen: See above.

.github/ISSUE_TEMPLATE/enhancement.md
```diff
@@ -0,0 +1,17 @@
...
+name: Enhancement
+about: Help us fix something that's broken.
+title: "PackageName:  General Summary of the Enhancement"
+labels: "type: enhancement"
+assignees: ''
+
+---
+
+## Enhancement
+@trilinos/PackageName
```
mhoemmen: See above.

.github/ISSUE_TEMPLATE/enhancement.md
```diff
@@ -0,0 +1,17 @@
...
+
+What feature/addition is needed?  How does this differ from what we currently
```
mhoemmen: See above.

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,15 @@
+---
+name: Question
+about: What's something you need help with?
+title: "PackageName:  General Summary of the Question"
+labels: "type: question"
+assignees: ''
+
+---
+
+## Question
```
mhoemmen: Not even sure if we need the header here.  The whole thing is a question.

jmgate: Yes, the though was that each template have a level two header indicating its type.  Seems that's more visible than the label off to the right, but I could be swayed.

.github/ISSUE_TEMPLATE/bug_report.md
jmgate: It seems we should be giving some guidance, otherwise there's not much point in having the template.  Particularly I find people often leave out the broader context of what they're trying to do.

.github/ISSUE_TEMPLATE/bug_report.md
jmgate: Made the switch.

.github/ISSUE_TEMPLATE/bug_report.md
mhoemmen: I don't mean to block your work -- it's just that every time we open an issue, we delete almost everything that we see.  Maybe we're not the target audience, though ;-)

jmgate:
@jwillenbring, @william76, @prwolfe, anyone care to weigh in on this one?

.github/ISSUE_TEMPLATE/bug_report.md
jmgate: No worries.  I think since this is already a significant reduction in the template text, we'll probably leave things as is, unless someone on the framework team wants to chime in.

Issue #4883, opened 2019-04-13
Tempus: Add Ability to Integrate Auxiliary Variables


ccober6:
@trilinos/tempus 

## Expectations
Have the ability to time integrate auxiliary variables.

There is a difference between the PhysicsState, _p_, and an auxiliary variable, _y_.  The PhysicState is physics data that does not need time integration.  It is just data at each time step that is needed to evaluate _f(x,xDot,t,p)_.  An auxiliary variable, _y_, however does need time integration, but has been separated from the solution variable, _x_, because it is a simple update or has a different time-integration.

To start with we will assume that the solution, _x_, and the auxiliary variables, _y_, are integrated with the same Stepper, but is a simple update from the SolutionState (e.g., _x__{n-1}, _x__{n}, _xDot__{n-1}, _xDot__{n}, _y__{n-1}, _yDot__{n-1}, and RK stage values) and other values available from the TimeDerivative class.

## Motivation and Context
The ability to segregate the solution, _x_, and auxiliary variables, _y_, can be a substantial computational savings because of the reduction in size of the solution vector from {_x_, _y_} to just {_x_} and simple update of {_y_}.

## Definition of Done
- [ ] Add auxiliary variables, _y_, to SolutionState.
- [ ] Incorporate _y_ to TimeDerivative class.
- [ ] Develop means for application to provide update function to Stepper.  (This may be a prototype for a new ModelEvaluator).
- [ ] Have TempusWrapperModelEvaluator use _y_ and compute _yDot_.
- [ ] Either implement the new auxiliary variables in every Stepper, or have the Stepper error out if auxiliary variables are used.
- [ ] Add test to cover this new capability.

PR #4884, opened 2019-04-14
Tempus: Add Ability to Integrate Auxiliary Variables #4883

ccober6:
@trilinos/tempus 

## Description
See #4883 

## Checklist

### DIFF
--- a/packages/tempus/src/Tempus_AuxModel.hpp
+++ b/packages/tempus/src/Tempus_AuxModel.hpp
@@ -1,0 +1,40 @@
+// @HEADER
+// ****************************************************************************
+//                Tempus: Copyright (2017) Sandia Corporation
+//
+// Distributed under BSD 3-clause license (See accompanying file Copyright.txt)
+// ****************************************************************************
+// @HEADER
+
+#ifndef Tempus_AuxModel_hpp
+#define Tempus_AuxModel_hpp
+
+
+#include "Tempus_ODEParameters.hpp"
+
+
+namespace Tempus {
+
+
+/** \brief A model to compute auxiliary variables.
+ *
+ *  This could grow into the Tempus version of ModelEvaluator.
+ */
+template <typename Scalar>
+class AuxModel
+{
+public:
+
+  virtual void evaluateAuxModel(
+  const Teuchos::RCP<Thyra::VectorBase<Scalar> > & y,
+  const Teuchos::RCP<Thyra::VectorBase<Scalar> > & yDot,
+  const Teuchos::RCP<Thyra::VectorBase<Scalar> > & x,
+  const Teuchos::RCP<Thyra::VectorBase<Scalar> > & xDot,
+  const Scalar time,
+  const Teuchos::RCP<ODEParameters<Scalar> > & p ) = 0;
+
+};
+
+} // namespace Tempus
+
+#endif // Tempus_AuxModel_hpp
--- a/packages/tempus/src/Tempus_ODEParameters.hpp
+++ b/packages/tempus/src/Tempus_ODEParameters.hpp
@@ -22,11 +22,8 @@ class ODEParameters
 {
   public:
     /// Constructor
-    ODEParameters()
-      : timeDer_(Teuchos::null), timeStepSize_(Scalar(0.0)), stageNumber_(0),
-        alpha_(Scalar(0.0)), beta_(Scalar(0.0)),
-        solutionHistory_(Teuchos::null), evaluationType_(SOLVE_FOR_X)
-    {}
+    ODEParameters() {}
+
     /// Constructor
     ODEParameters(Teuchos::RCP<TimeDerivative<Scalar> > timeDer,
                   Scalar timeStepSize, Scalar alpha, Scalar beta,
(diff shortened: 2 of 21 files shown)


packages/tempus/src/Tempus_AuxModel.hpp
```diff
@@ -0,0 +1,40 @@
...
+ *
+ *  This could grow into the Tempus version of ModelEvaluator.
+ */
+template <typename Scalar>
+class AuxModel
+{
+public:
+
+  virtual void evaluateAuxModel(
+  const Teuchos::RCP<Thyra::VectorBase<Scalar> > & y,
```
mhoemmen: Will AuxModel implementations be allowed to keep RCPs to any of these input arguments?  If not, then you should pass the arguments by const reference (if they are always nonnull) or raw pointer (if they may be nonnull) or `Teuchos::Ptr` (if they may be nonnull).

Passing an `RCP<T>` (or `const RCP<T>&`) into an instance of a class suggests that you intend to share ownership.  I'm guessing that's not what you intend here.  Implementations of this method will likely just use the objects during the method and not keep RCPs to them.

packages/tempus/src/Tempus_ODEParameters.hpp
```diff
@@ -0,0 +1,50 @@
...
+        alpha_(alpha), beta_(beta), solutionHistory_(sh),
+        evaluationType_(evaluationType)
+    {}
+
+    Teuchos::RCP<TimeDerivative<Scalar> > timeDer_;
+    Scalar                                timeStepSize_;
+    int                                   stageNumber_;
+    Scalar                                alpha_;  // Only valid for Implicit
+    Scalar                                beta_;   // Only valid for Implicit
+    const Teuchos::RCP<SolutionHistory<Scalar> >& solutionHistory_;
```
mhoemmen: It's not a good idea to keep an RCP in a class instance by const reference.  That loses the entire value of passing things around by RCP, since it doesn't prevent `solutionHistory_` from falling out of scope and being deallocated outside of this class' control.  Just keep the RCP by value:
```c++
Teuchos::RCP<SolutionHistory<Scalar>> solutionHistory_;
```
If you don't want to share ownership of the input SolutionHistory with ODEParameters, then you could store SolutionHistory as a (raw) pointer or (raw) reference.

packages/tempus/src/Tempus_ODEParameters.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+
+template<class Scalar>
+class ODEParameters
+{
+  public:
+    /// Constructor
+    ODEParameters()
+      : timeDer_(Teuchos::null), timeStepSize_(Scalar(0.0)), stageNumber_(0),
+        alpha_(Scalar(0.0)), beta_(Scalar(0.0)),
```
mhoemmen: You could just set default values for the fields where you declare them.  For example:
```c++
Scalar timeStepSize_ (0.0);
int stageNumber_ (0);
```
The default value of an RCP is always as if you assigned `Teuchos::null` to it, so you don't need to specify the default there.

packages/tempus/src/Tempus_Stepper_decl.hpp
```diff
@@ -79,6 +80,12 @@ class Stepper
       const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& appModel) = 0;
     virtual Teuchos::RCP<const Thyra::ModelEvaluator<Scalar> > getModel() = 0;
 
+    virtual void setAuxModel(
+      const Teuchos::RCP<const AuxModel<Scalar> >& auxModel);
+    virtual void setNonConstAuxModel(
```
mhoemmen: I can see why you have a separate `setNonConstAuxModel`; it's because the compiler isn't smart enough both to cast to the base class and to cast from nonconst to const at the same time.  I'm not sure if this would increase usability much over just using the existing `rcp_const_cast` function, though.

packages/tempus/src/Tempus_AuxModel.hpp
ccober6: Hi Mark,  thanks for the "pointers" as I am not very savvy about these things. :)

But I has following the ModelEvaluator::evalModelImpl() example on this.  Is there a reason it is done there?

packages/tempus/src/Tempus_ODEParameters.hpp
ccober6: Ugh.  Yes, I should storage by RCP by value.

packages/tempus/src/Tempus_ODEParameters.hpp
ccober6: Yeah, old habits.  :)

packages/tempus/src/Tempus_Stepper_decl.hpp
ccober6: Not sure where I picked up this piece of code, but wouldn't we need 

`RCP<const A> ca_ptr = rcp_implicit_cast<const A>(a_ptr);// To const`

to insert the const, and get rid of `setNonConstAuxModel()`?

packages/tempus/src/Tempus_AuxModel.hpp
mhoemmen: I'm kind of intruding here -- I was just looking through GitHub notifications while waiting on a long build.  I'm not sure about the history of ModelEvaluator.  If you don't need interface compatibility with ModelEvaluator, then I think you should feel free to do the right thing.  If you _do_ need interface compatibility with ModelEvaluator, then you're stuck, I guess.

packages/tempus/src/Tempus_Stepper_decl.hpp
mhoemmen: The following should work:
```c++
RCP<AuxModelSubclass<Scalar>> a_ptr = whatever ();
// implicitly casts the result of rcp_implicit_cast to const AuxModel<Scalar>
thing.setAuxModel (rcp_implicit_cast<AuxModel<Scalar>> (a_ptr));
```

william76:
@trilinos/tempus This PR has been closed due to inactivity.

PR #4839, opened 2019-04-09
Tpetra: deprecate node args (phase one)

kddevin:
@trilinos/tpetra

## Description
We are deprecating use of Node as an argument to functions.
These commits deprecate node arguments in readSparse*, readHBMatrix, generateMatrix, create*MapWithNode, and generate_miniFE_*.  Tests calling these functions have been updated.

More work remains to be done, but getting it all into one PR may be onerous.  I'd like to see how these changes do in PR testing.  Then I can add other work to this PR, or create a phase-two PR.
## Motivation and Context

Users should not create Node instances.

## How Has This Been Tested?

Tested with and without TPETRA_ENABLE_DEPRECATE_NODE on mac with clang.
Tested with Tpetra build and MueLu build with all optional dependencies on.
10 tests (of 287) fail in all builds:
         37 - Belos_gcrodr_hb_MPI_4 (Failed)
         66 - Belos_BlockGmresPoly_Epetra_File_Ex_0_MPI_4 (Failed)
         67 - Belos_BlockGmresPoly_Epetra_File_Ex_1_MPI_4 (Failed)
         85 - Belos_CustomSolverFactory_MPI_4 (Failed)
        175 - MueLu_MasterList_MPI_1 (Failed)
        201 - MueLu_UnitTestsBlockedTpetra_MPI_1 (Failed)
        226 - MueLu_CreateOperatorEpetra_MPI_1 (Failed)
        227 - MueLu_CreateOperatorEpetra_MPI_4 (Failed)
        232 - MueLu_CreateOperatorTpetra_MPI_1 (Failed)
        233 - MueLu_CreateOperatorTpetra_MPI_4 (Failed)

## Checklist

- [ ] My commit messages mention the appropriate GitHub issue numbers.
- [ x] My code follows the code style of the affected package(s).
- [ x] My change requires a change to the documentation.  
- [x ] I have updated the documentation accordingly. (DOXYGEN IN ifdef BLOCKS, TOO)
- [ ] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.  (SEE 10 TESTS ABOVE)
- [ ] No new compiler warnings were introduced.  (MANY DEPRECATION WARNINGS INTRODUCED)
- [ x] These changes break backwards compatibility. (THAT'S THE POINT)

PR #4888, opened 2019-04-15
Tpetra:  deprecate node args (phase two):  remove node args in map

kddevin:
@trilinos/tpetra @trilinos/xpetra @trilinos/muelu @trilinos/stokhos 

## Description
Tpetra is deprecating use of Node as an argument in functions.  
This PR removes the Node argument from map creation, and removes other interfaces that take a Node argument.
Still remaining:  removal of Node from clone functions and TSQR.

## Motivation and Context
Node is made obsolete by Kokkos::ExecutionSpace and Kokkos::MemorySpace.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #4839 
* Part of 
* Composed of 

## How Has This Been Tested?

Builds of Tpetra with and without deprecated code.
Builds with MueLu, Stokhos, Nox, Thyra, Anasazi and Belos, with and without deprecated code.
See #4839 for details.

## Checklist

SOME TESTS FAIL, BUT THEY ALSO FAIL IN DEVELOP BRANCH.
DEPRECATION WARNINGS ARE INTRODUCED
THAT'S THE POINT.


packages/tpetra/core/test/Map/Map_UnitTests.cpp
```diff
@@ -426,7 +426,7 @@ namespace {
     RCP<N2> n2 (new N2);
 
     // create a contiguous uniform distributed map with numLocal entries per node
-    RCP<const Map1> map1 = createUniformContigMapWithNode<LO, GO, N1> (numGlobal, comm, n1);
+    RCP<const Map1> map1 = createUniformContigMapWithNode<LO, GO, N1> (numGlobal, comm);
```
lucbv: @kddevin should a test using the deprecated function signature be kept to ensure it works until the deprecated function signature is actually removed from the code?
We are having a similar discussion for some MueLu deprecation...

packages/tpetra/core/test/ImportExport2/ImportExport2_UnitTests.cpp
```diff
@@ -2026,7 +2026,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL( Import_Util, UnpackAndCombineWithOwningPIDs,
   RCP<const MapType> MapSource, MapTarget;
 
   const bool debug = ::Tpetra::Details::Behavior::debug ();
-  const bool verbose = ::Tpetra::Details::Behavior::verbose ();
+  verbose = ::Tpetra::Details::Behavior::verbose ();
```
mhoemmen: This doesn't quite make sense -- did the unit test capture a `verbose` variable from outside?

packages/tpetra/core/test/Map/Map_UnitTests.cpp
mhoemmen: I don't think we're obligated to keep supporting deprecated functions, but it certainly wouldn't hurt to leave a test in place, as long as it doesn't generate deprecated warnings if deprecated code is disabled.

kddevin: I admit I was not consistent.  In some places, I kept the old interface (e.g., https://github.com/trilinos/Trilinos/blob/e000e39a04969848a2e462def7dbde9e83f822ff/packages/tpetra/core/test/TpetraUtils_MatrixGenerator.hpp#L226 and https://github.com/trilinos/Trilinos/blob/e000e39a04969848a2e462def7dbde9e83f822ff/packages/tpetra/core/test/TpetraUtils_MatrixGenerator.hpp#L176 ).  In other places, I just removed calls to the old interface.

packages/tpetra/core/test/ImportExport2/ImportExport2_UnitTests.cpp
kddevin: Removes a shadow warning from https://github.com/trilinos/Trilinos/blob/e000e39a04969848a2e462def7dbde9e83f822ff/packages/tpetra/core/test/ImportExport2/ImportExport2_UnitTests.cpp#L117

mhoemmen: @kddevin Thanks for explaining!  I would have somewhat preferred getting rid of the outer `verbose` variable, but the way you did it is fine.  Thanks!

etphipp:
This generates build errors if Tpetra deprecated code is enabled:
```
/home/etphipp/Trilinos/Trilinos/packages/xpetra/src/Map/Xpetra_TpetraMap.hpp:643:55: error: 
      missing default argument on parameter
               const Teuchos::RCP< Node > & /* node */) 
                                                      ^
```
Looking at the code, you have a function argument with no default value after an argument with a default value:
```
    TPETRA_DEPRECATED
    TpetraMap (global_size_t numGlobalElements,
               GlobalOrdinal indexBase,
               const Teuchos::RCP< const Teuchos::Comm< int > > &comm,
               LocalGlobal lg=GloballyDistributed,
               const Teuchos::RCP< Node > & /* node */)
      : TpetraMap(numGlobalElements, indexBase, comm, lg)
    {}
#endif // TPETRA_ENABLE_DEPRECATED_CODE
```
Can the default value for `lg` just be removed?

kddevin:
Yes, "=GloballyDistributed" should be removed.  I will do it.  Sorry we missed that one (and that PR testing did not catch the problem).

PR #4885, opened 2019-04-14
Tpetra::MultiVector: Fix #4870 (prefer running norms on device)

mhoemmen:
@trilinos/tpetra @trilinos/stokhos @vbrunini @etphipp 

Superseded by https://github.com/trilinos/Trilinos/pull/4899.  Supersedes PR #4881 and PR #4871.

## Description

  - Prefer running `Tpetra::MultiVector` norms on device, even if it is currently sync'd to host, unless the vector has too few local rows.
  - Factor out the implementation of `Tpetra::MultiVector::norm*` into a separate function, `Tpetra::Details::normImpl`.  I do ETI for that function, when ETI is enabled.  This required adding new ETI macros for iterating over the `(Scalar, Node)` type combination.
  - Fix Stokhos to work around lack of ETI for `Tpetra::Details::normImpl`.

To elaborate the last point: Tpetra's ETI macro for `Tpetra::Details::normImpl` only takes 2 template parameters: `Scalar` and `Node`.  This meant that I couldn't use Stokhos' ETI system directly (see `TPETRA_ETI_CLASSES` in `stokhos/src/CMakeLists.txt`).  I had to take different approaches for PCE types vs. `MP::Vector` types.  My solution for PCE types looks more permanent (read "is less of a hack") than my solution for `MP::Vector` types; it would be great if @etphipp could comment.

## Motivation and Context

  - SPARC performance
  - Build time

## Related Issues

* Closes #4870

PR #4881, opened 2019-04-13
Tpetra::MultiVector: Fix #4870 (prefer running norms on device)

mhoemmen:
@trilinos/tpetra @trilinos/stokhos 

Supersedes PR #4871.

## Description

  - Prefer running MultiVector norms on device, even if it is currently sync'd to host, unless the vector has too few local rows.
  - Factor out the implementation of `Tpetra::MultiVector::norm*` into a separate function, `Tpetra::Details::normImpl`.  I do ETI for that function, when ETI is enabled.  This required adding new ETI macros for iterating over the `(Scalar, Node)` type combination.
  - Fix Stokhos to work around lack of ETI for `Tpetra::Details::normImpl`.  Tpetra's ETI macro for `Tpetra::Details::normImpl` only takes 2 template parameters: `Scalar` and `Node`.  This meant that I couldn't use Stokhos' ETI system (see `TPETRA_ETI_CLASSES` in `stokhos/src/CMakeLists.txt`).  I hacked in a header file include in Stokhos to make this build.

## Motivation and Context

  - SPARC performance
  - Build time

## Related Issues

* Closes #4870

PR #4899, opened 2019-04-15
Tpetra,Stokhos: Fix #4870 (prefer running MV norms on device)

### LINKED ISSUE #4870: Tpetra::MultiVector::norm*: Prefer running on device

mhoemmen:
@trilinos/tpetra @trilinos/stokhos @vbrunini @etphipp 

Supersedes PR #4881, PR #4871, and PR #4885.

## Description

  1. Make `Tpetra::MultiVector` prefer computing the local (to each MPI process) parts of norms (`norm1`, `norm2`, `normInf`) on device.  Do so even if the MultiVector is currently sync'd to host, unless the MultiVector has too few local rows.
  2. Factor out the implementation of `Tpetra::MultiVector::norm*` into a separate function, `Tpetra::Details::normImpl`.  Do ETI for that function, when ETI is enabled.
  3. Make Stokhos do ETI for `Tpetra::Details::normImpl`, for its `UQ::PCE` and `MP::Vector` Scalar types.

To elaborate the last point: Tpetra's ETI macro for `Tpetra::Details::normImpl` only takes 2 template parameters: `Scalar` and `Node`.  This meant that I couldn't use Stokhos' ETI system directly (see `TPETRA_ETI_CLASSES` in `stokhos/src/CMakeLists.txt`).  I had to take different approaches for PCE types vs. `MP::Vector` types.  My solution for PCE types looks more permanent (read "is less of a hack") than my solution for `MP::Vector` types; it would be great if @etphipp could comment.

## Motivation and Context

  - SPARC performance
  - Build time

## Related Issues

* Closes #4870


packages/stokhos/src/sacado/kokkos/vector/tpetra/Tpetra_Details_normImpl_MP_Vector_Cuda.cpp
```diff
@@ -0,0 +1,70 @@
...
+// Tpetra::Details::normImpl only takes two arguments: (SCALAR, NODE).
+
+#include "Stokhos_Sacado_Kokkos_MP_Vector.hpp"
+#include "Kokkos_ArithTraits_MP_Vector.hpp"
+#include "Tpetra_Details_normImpl_decl.hpp"
+#include "Tpetra_Details_normImpl_def.hpp"
+#include "TpetraCore_ETIHelperMacros.h"
+
+#define TPETRA_DETAILS_NORMIMPL_INSTANT_N( NODE ) \
+  using scalar_type_##NODE = Sacado::MP::Vector<Stokhos::StaticFixedStorage<int, double, 16, NODE::execution_space> >; \
```
etphipp: This should work in the sense that the autotester should pass, but this isn't in the spirit of how ETI for MP::Vector is supposed to work.  The issue is hard-coding the 16 for the ensemble size.  That is supposed to be a configure argument (with the default = 16).  So if the user chose anything else, it won't work.

mhoemmen: @etphipp I wasn't pleased with hard-coding 16 ;-) .  How do I fix this?

etphipp: @mhoemmen, I don't think it is hard to fix, but it is probably just easier if I just do it.  How about I pull down your PR and add some commits that use the ETI logic to put in the proper ensemble size.  Can you not merge the PR until I've done that? (I can start working on it now, might be done by the end of the day).

mhoemmen: @etphipp Thanks!  You should have permission just to push the fix to this PR.  I've removed the "automerge" tag, but won't use the WIP tag, just so that we get PR test results.

etphipp: OK, I just pushed a commit that removes the hard-coding of the ensemble size.  Unfortunately I had to basically duplicate all of Stokhos' ETI macros for the case when there is no LO, GO.  But it seems to work.

packages/stokhos/src/CMakeLists.txt
```diff
@@ -804,6 +807,15 @@ IF (Stokhos_ENABLE_Sacado)
           "${STOKHOS_DEPLIBS}"
           DEPLIBS_OUT)
         LIST(APPEND STOKHOS_TPETRA_DEPLIBS ${DEPLIBS_OUT})
+       PROCESS_ENSEMBLE_ETI(
+          "${Stokhos_ETI_Ensemble_Sizes}"
+          "${TPETRA_SD_ETI_CLASSES}"
```
mhoemmen: This is much better -- I was thinking on the way home about setting up a table of ETI classes, that lists a subset of `{S, LO, GO, N}` over which to instantiate each.

packages/stokhos/src/sacado/kokkos/pce/tpetra/Stokhos_Tpetra_ETI_Helpers_UQ_PCE_DEVICE.hpp
```diff
@@ -46,23 +46,43 @@
 #define INSTANTIATE_UQ_PCE_STORAGE(INSTMACRO, STORAGE, LO, GO, N)      \
   INSTMACRO( Sacado::UQ::PCE<STORAGE>, LO, GO, N )
 
+#define INSTANTIATE_UQ_PCE_STORAGE_SD(INSTMACRO, STORAGE, N)      \
```
mhoemmen: "SD" is the right name -- Node has all the typedefs that `Kokkos::Device` has, so it's harmless use to Node where a `Kokkos::Device` would do.  We will _eventually_ get rid of Node ;-) .

packages/stokhos/src/sacado/kokkos/pce/tpetra/Stokhos_Tpetra_ETI_Helpers_UQ_PCE_DEVICE.hpp
```diff
@@ -46,23 +46,43 @@
...
+  INSTANTIATE_UQ_PCE_S_D_SD(INSTMACRO, DFN_ ## N, N)
+
 #define INSTANTIATE_UQ_PCE(INSTMACRO, LO, GO, N) \
   INSTANTIATE_UQ_PCE_S(INSTMACRO, LO, GO, N)
 
+#define INSTANTIATE_UQ_PCE_SD(INSTMACRO, N) \
+  INSTANTIATE_UQ_PCE_S_SD(INSTMACRO, N)
+
 #define INSTANTIATE_TPETRA_UQ_PCE_N(INSTMACRO, N)  \
   INSTANTIATE_UQ_PCE_S(INSTMACRO, int, int, N)
```
mhoemmen: @etphipp Does `int, int` here refer to LO, GO?  This might be the reason for https://github.com/trilinos/Trilinos/issues/4902, then.

packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_ETI_Helpers_MP_Vector_ENSEMBLE_SIZE_DEVICE.hpp
```diff
@@ -46,23 +46,43 @@
...
 
+#define INSTANTIATE_MP_VECTOR_S_D_SD(INSTMACRO, D, N) \
+  INSTANTIATE_MP_VECTOR_SFS_SLD_SD(INSTMACRO, double, int, D, N)
+
 #define INSTANTIATE_MP_VECTOR_S(INSTMACRO, LO, GO, N) \
   typedef Stokhos::DeviceForNode<N>::type DFN_CPU_ ## LO ## _ ## GO ## _ ## N; \
   INSTANTIATE_MP_VECTOR_S_D(INSTMACRO, DFN_CPU_ ## LO ## _ ## GO ## _ ## N, LO, GO, N)
 
+#define INSTANTIATE_MP_VECTOR_S_SD(INSTMACRO, N) \
+  typedef Stokhos::DeviceForNode<N>::type DFN_CPU_ ## N; \
```
mhoemmen: I don't think we actually need `DeviceForNode` any more, since all the Node types remaining are post-Kokkos-refactor types that have `device_type`, `execution_space`, and `memory_space` typedefs.  This is not relevant to the current PR, though.

packages/stokhos/src/sacado/kokkos/pce/tpetra/Tpetra_Details_normImpl_UQ_PCE_Cuda.cpp
```diff
@@ -1,69 +0,0 @@
...
-
-#include "Stokhos_Tpetra_UQ_PCE.hpp"
-#include "Tpetra_Details_normImpl_decl.hpp"
-#include "Tpetra_Details_normImpl_def.hpp"
-#include "TpetraCore_ETIHelperMacros.h"
-
-#define TPETRA_DETAILS_NORMIMPL_INSTANT_N( NODE ) \
-  using scalar_type_##NODE = Sacado::UQ::PCE<Stokhos::DynamicStorage<int, double, NODE::execution_space> >; \
-  TPETRA_DETAILS_NORMIMPL_INSTANT( scalar_type_##NODE, NODE ) \
-  using host_device_type_##NODE = Kokkos::View<double*, NODE::device_type>::HostMirror::device_type; \
```
mhoemmen: @etphipp I had to add this extra instantiation to the `_Cuda.cpp` file, since `normImpl` needs to instantiate for both `device_type` and its host mirror device type.  For the Cuda case, the latter is `Device<DefaultHostExecutionSpace, CudaUVMSpace>`.

packages/stokhos/src/sacado/kokkos/pce/tpetra/Stokhos_Tpetra_ETI_Helpers_UQ_PCE_DEVICE.hpp
etphipp: Yes, I think the first one is LO and the second was is GO.  The fix might be as simple as replacing them with `Tpetra::Details::DefaultTypes::local_ordinal_type` and `Tpetra::Details::DefaultTypes::global_ordinal_type`.

packages/stokhos/src/sacado/kokkos/pce/tpetra/Tpetra_Details_normImpl_UQ_PCE_Cuda.cpp
etphipp: OK, I missed that.  It would be nice if we could somehow include that in the instantiation macro in Tpetra, but I don't see a way to do that and avoid a duplicate instantiation for host device types.  I will modify the Stokhos ETI logic to change the macro for CUDA.

mhoemmen: @etphipp Thanks!  Sorry to make extra trouble for you.  There's no way to make an explicit instantiation conditional on a constexpr, alas.  Would it help for Tpetra to make a macro just for CUDA?  Another idea would be to make ETI for Devices separate from ETI for Nodes, and add `Device<DefaultHostExecutionSpace, CudaUVMSpace>` to the list if CUDA is enabled.

mhoemmen:
@etphipp I'm marking this as WIP to avoid clogging the PR testing pipeline.  Please feel free to unmark once you've pushed a commit.  Thanks!

etphipp:
I believe I fixed the HostMirror instantiation issues, so I removed the WIP label.

etphipp:
> @etphipp I can't actually approve this since it's my own PR -- I'm not sure if you can either -- but these changes look great; thanks! :D

You’re right, I can’t review it.  Can someone else?

PR #4912, opened 2019-04-16
Tpetra: Deprecate Clone

csiefer2:
This only does Tpetra.  But it is a beginning.


packages/tpetra/core/test/CrsGraph/CrsGraph_ReindexColumns.cpp
```diff
@@ -155,21 +155,7 @@ namespace {
...
-	Teuchos::RCP<Node> node; // can be null; only for type deduction
-        graph2 = graph.clone (node, clonePlist);
-      } catch (std::exception& e) {
-        std::ostringstream err2;
-        err2 << "Proc " << myRank << ": CrsGraph::clone threw an exception: "
-             << e.what () << endl;
-        cerr << err2.str ();
-      }
-    }
+    RCP<graph_type> graph2 = rcp(new graph_type(graph));
```
mhoemmen: That's... not a deep copy.  All the Tpetra constructors shallow-copy.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_NodeConversion.cpp
```diff
@@ -198,8 +198,9 @@ namespace {
     TEUCHOS_UNIT_TEST_TEMPLATE_1_INSTANT( CrsMatrix, NodeConversion, N2 )
 
   TPETRA_ETI_MANGLING_TYPEDEFS()
-
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
mhoemmen: We also need to protect the declaration of the test with this macro.  Otherwise, the compiler may emit "you declared a class without using it" warnings, and may also emit "you're using deprecated code" warnings.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_ReindexColumns.cpp
```diff
@@ -156,21 +156,7 @@ namespace {
...
-	RCP<Node> node; // can be null; only need it for type deduction
-        graph2 = graph.clone (node, clonePlist);
-      } catch (std::exception& e) {
-        std::ostringstream err2;
-        err2 << "Proc " << myRank << ": CrsMatrix::clone threw an exception: "
-             << e.what () << endl;
-        cerr << err2.str ();
-      }
-    }
+    RCP<graph_type> graph2 = rcp(new graph_type(graph));
```
mhoemmen: See above; this is not a deep copy.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrixKernels_UnitTests.cpp
```diff
@@ -166,7 +166,16 @@ getIdentityMatrixWithMap (Teuchos::FancyOStream& out,
   return identityMatrix;
 }
 
-
+template<class Matrix_t>
+RCP<Matrix_t>
+copyMatrix(RCP<Matrix_t> &A) {
```
mhoemmen: If you don't intend on the RCP being an output argument, then pass it either by value `RCP<Matrix_t>` or by const reference `const RCP<Matrix_t>&`.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -166,7 +166,16 @@ getIdentityMatrixWithMap (Teuchos::FancyOStream& out,
   return identityMatrix;
 }
 
-
+template<class Matrix_t>
+RCP<Matrix_t>
+copyMatrix(RCP<Matrix_t> &A) {
```
mhoemmen: See above.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrixKernels_UnitTests.cpp
mhoemmen: More importantly, if you need to deep-copy a CrsMatrix a lot, then we should add a new CrsMatrix deep-copy constructor like `MultiVector<> Y (X, Teuchos::Copy)`.

csiefer2: As discussed.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_NodeConversion.cpp
csiefer2: We routinely declare unused test macros and the compiler never complains...

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -764,6 +764,7 @@ namespace Tpetra {
       Kokkos::resize(newvals,vals.extent(0));
       Kokkos::deep_copy(newvals,vals);
       k_values1D_ = newvals;   
+      this->fillComplete(source.getDomainMap(),source.getRangeMap());
```
mhoemmen: 1. Getting `vals` will only work if the input matrix is static profile.
2. The input CrsMatrix may not necessarily be fill complete.
```c++
if (source.isFillComplete ()) {
  this->fillComplete (source.getDomainMap (), source.getRangeMap ());
}
```

csiefer2: Fixed.

PR #4918, opened 2019-04-16
Tpetra: Adding Dirichlet condition helper routine

csiefer2:
Tests: CrsMatrix/applyDirichlet; RHEL6


packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp
```diff
@@ -5106,6 +5106,35 @@ namespace Tpetra {
...
+
+    Kokkos::parallel_for (range_type (0, rowIndices.extent(0)), KOKKOS_LAMBDA(const LocalOrdinal i) {
+        GlobalOrdinal row_gid = rmap.getGlobalElement(rowIndices(i));
+        for(size_t j=rowptr(i); j < rowptr(i+1); j++) 
+          if (cmap.getGlobalElement(colind(j)) == row_gid)
+            values(j) = one;
+          else
+            values(j)=zero;
+      });
+    A.fillComplete(A.getDomainMap(),A.getRangeMap());
```
mhoemmen: I would very much prefer to let the user call fillComplete.

csiefer2: This would probably be best handled with a "callFillComplete" argument to allow it to be used either way.

mhoemmen: @csiefer2 or you could just let the user call fillComplete when they are ready... just sayin' ;-)

I want to finish some other things today but I'll be happy to contribute patches to this.

csiefer2: I mean, sure, you could.  But a convenience routine (which is what this is) should really have maximal convenience :P

mhoemmen:
I'm working on a patch now.

mhoemmen:
@csiefer2 Check this out -- I put it into a separate file for better encapsulation and so that users who don't use it won't have to pay for its build.  I didn't do ETI for it, though we could later.  Also, I added error checks.

csiefer2:
Cool!  Looks good to me!

mhoemmen:
Build errors look legit.  I'll need to reevaluate this.

csiefer2:
Bump.  Do not close.

csiefer2:
Figured I'd resurrect this before @william76 purges all of the old PRs :)

william76:
reaper-death

csiefer2:
Finally got the last lingering cuda issues...  "const" == evil

PR #4922, opened 2019-04-16
Tpetra: Fix default Global Ordinals

william76:
@trilinos/tpetra 

#4907 turned into a mess with that merge conflict in terms of what GitHub thought was in the changelog for that PR.  Easier to just kill it and create a new clean PR on a fresh branch since I was only touching 2 files.

I don't plan to run all 18 of my test cases from before, but I'll run some representative cases to make sure the merge with @mhoemmen updates didn't break anything.

### DIFF
--- a/packages/tpetra/CMakeLists.txt
+++ b/packages/tpetra/CMakeLists.txt
@@ -603,50 +603,78 @@ LIST (APPEND ${PACKAGE_NAME}_ETI_LORDS "int")
 # than global_size_t (e.g., using the wrong
 # Teuchos::OrdinalTraits<>::invalid().
 
-# GO = int is enabled by default, but only because of Bug 6358.
-GLOBAL_SET(HAVE_TPETRA_INST_INT_INT_DEFAULT ON)
-
-# GO = long long is enabled by default.  However, if the user enabled
-# GO = long EXPLICITLY, we disable GO = long long by default, because
-# we don't want to enable multiple "long" GO types by default.
-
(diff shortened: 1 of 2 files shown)


william76:
Replaced the warning message in the CMakeLists.txt file that warned when the user overrides all the default GlobalOrdinal types without enabling something to replace with a new message at the place where the FATAL\_ERROR is invoked.  Going forward, if the user disables the default without turning on something else they'll see this:

```
-- Tpetra: ERROR Tpetra requires a GlobalOrdinal type, but none were specified.
-- This can happens when the default is overridden by user options without enabling
-- another GlobalOrdinal type in its place. Please check your configuration to ensure
-- that the default is not being specifically disabled without enabling one of the others:
--    Tpetra_INST_INT_INT
--    Tpetra_INST_INT_LONG
--    Tpetra_INST_INT_LONG_LONG (Default, but was overridden)
--    Tpetra_INST_INT_UNSIGNED
--    Tpetra_INST_INT_UNSIGNED_LONG
CMake Error at packages/tpetra/CMakeLists.txt:1204 (MESSAGE):
  Tpetra: Tpetra requires a global ordinal type, but none were specified.
  See above message for options.
```

william76:
@mhoemmen @kddevin I ran through my set of configure checks on this one and it looks like it's behaving as I expect it to.  Please have a look when you get a chance and review :)

Screen Shot 2019-04-16 at 5 44 28 PM

PR #4926, opened 2019-04-17
Fix issue 4925, undefined symbols in static build

jhux2:
Auto-PR for SHA cd624b0

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/4926)

### DIFF
--- a/packages/muelu/src/CMakeLists.txt
+++ b/packages/muelu/src/CMakeLists.txt
@@ -460,19 +460,18 @@ ENDIF()
 #
 
 TRIBITS_ADD_LIBRARY(
-  muelu
-  HEADERS ${HEADERS}
-  SOURCES ${SOURCES}
-  )
-
-TRIBITS_ADD_LIBRARY(
   muelu_lgn
   HEADERS ${HEADERS}
   SOURCES ${SOURCES_LGN}
   )
 
-target_link_libraries(muelu muelu_lgn)
-target_link_libraries(muelu_lgn muelu)
+TRIBITS_ADD_LIBRARY(
+  muelu
+  HEADERS ${HEADERS}
+  SOURCES ${SOURCES}
+  DEPLIBS muelu_lgn
+  )
+
 
 # touch CMakeLists.txt because a new file was created in Utils/ExplicitInstantiation of Utils/ForwardDeclaration
 # touch CMakeLists.txt because a new file was created in Utils/ExplicitInstantiation of Utils/ForwardDeclaration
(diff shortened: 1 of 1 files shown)


bathmatt:
@bartlettroscoe question, do we have any static lib testing?  A bug snuck in that failed due to lib order errors, this is still failing, but . . . It would have been nice to catch it a bit early

packages/muelu/src/CMakeLists.txt
```diff
@@ -471,6 +471,9 @@ TRIBITS_ADD_LIBRARY(
   SOURCES ${SOURCES_LGN}
   )
 
+target_link_libraries(muelu muelu_lgn)
```
bartlettroscoe: FYI: This will break the debian builds of Trilinos which prefix all of the Trilinos library names.  This will also break any future attempt to get installation testing and backward compatibility testing working again where one configures the test sand examples in Trilinos against a pre-installed version of Trilinos and therefore in that case TRIBITS_ADD_LIBRARY() is a no-op and these `target_link_libraries()` command will crash.

In a follow-up PR, please switch to using the [DEPLIBS argument in TRIBITS_ADD_LIBRARY()](https://tribits.org/doc/TribitsDevelopersGuide.html#formal-arguments-tribits-add-library) as that will work in all cases.

bathmatt:
If this is going to be a long merge process, we may want to back out the library split of muelu?

bartlettroscoe:
@bathmatt asked:

> question, do we have any static lib testing? A bug snuck in that failed due to lib order errors, this is still failing, but . . . It would have been nice to catch it a bit early

There are currently no PR builds that do static builds.  See my comment to the @trilinos/framework team in https://github.com/trilinos/Trilinos/issues/4929#issuecomment-484074590. 

But of course our post-merge ATDM Trilinos builds caught this.  I wish we could find a way for the ATDM Trilinos builds could protect EMPIRE.  The git-git workflow would basically allow this I think.

lucbv:
@bartlettroscoe could a static build be part of the promotion to master process to protect the master branch for ATDM customers?

bartlettroscoe:
@lucbv asked:

> could a static build be part of the promotion to master process to protect the master branch for ATDM customers?

I can't comment on that.  I don't control that process.

jhux2:
@bartlettroscoe I switch to using DEPLIBS, please review.  This works in a local clang static build, and a local clang shared build is almost done.

jhux2:
Ok, my clang shared build completes.  I'll retest.

@mhoemmen Thanks for the quick review.

jhux2:
> @jhux2, actually, this is not even configuring. See:
> 
> * https://testing.sandia.gov/cdash/viewConfigure.php?buildid=4906085

Correct, but I've just pushed a commit that switches to DEPLIBS, which works for me in both static and shared.  Please have a look at the change.

bathmatt:
Both shared and static libs work on my desktop

bartlettroscoe:
FYI: PR testing already shows that configure passed.

Issue #2689, opened 2018-05-08
 Add install hooks for ATDM Trilinos configuration and ctest -S driver


bartlettroscoe:
**CC:** @fryeguy52

## Description

In order for the ATDM projects to adopt the new [ATDM Trilinos Configuration and Jenkins Drivers](https://snl-wiki.sandia.gov/display/CoodinatedDevOpsATDM/ATDM+Builds+of+Trilinos) (that submit to CDash as a byproduct), these scripts must support the installation of Trilinos and using Trilinos from that install.

For this to occur:

* There must be a single that can be sourced sitting in the installation directory that will load the correct env to use that installed version of Trilinos. **[Done]**
* Trilinos must be installeable using that ATDM Trilinos configuration from a Jenkins job 

## Possible solutions

First, a new env var `ATDM_CONFIG_TRIL_CMAKE_INSTALL_PREFIX` could be added that would be read in the `ATDMDevEnvSettings.cmake` file and would set the CMake var `CMAKE_INSTALL_PREFIX`.  That would give you the correct configuration of Trilinos and would know where to install.

As for doing the installation itself, one simple solution would be to provide a `atdm/install-prebuilt-trilinos.sh` script that will run `make install` on the built Trilinos version.  This script would need to be explicitly called by some Jenkins driver script.  But any failures would not be posted to CDash.

If we wanted to allow for installation errors to get reported up to CDash, then a `CTEST_DO_INSTALL` option could be added to the [TRIBITS_CTEST_DRIVER()](https://tribits.org/doc/TribitsDevelopersGuide.html#tribits-ctest-driver) function and would do a `cmake_build( ... )` command that would run the `install` target (and in parallel) after the project build was complete.  That would also allow logic for not installing if there was a build failure and, again, would allow for any install failures to get reported to CDash.

Also, to help to make installation testing for downstream customer codes stronger, the smart-jenkins-driver.sh script could be updated to move the source and build directories out of the way to catch mistakes were the installed configuration files point into the source or build trees.  This could be accomplished by adding the steps:

* Run the smart-jenkins-driver.sh script under `$WORKSPACE/Trilinos`
* If  `$WORKSPACE/moved/SRC_AND_BUILD/` already exists, the move it back to  `$WORKSPACE/SRC_AND_BUILD/`
* After the build, installation and testing are completed, move `$WORKSPACE/SRC_AND_BUILD/` to `$WORKSPACE/moved/SRC_AND_BUILD/`.

[...]

PR #4944, opened 2019-04-18
Make installs more robust handle directory permissions (#2689)

bartlettroscoe:
This PR sets up the ATDM Trilinos build of Trilinos to automatically make the install directories world-readable and it makes the installs performed by `tribits_ctest_driver()` more robust.  In the past, if there was even a single build error such as in Panzer or Piro, it would stop the install of Trilinos.  That broke the SPARC Trilinos Integration builds where SPARC does not even depend on Piro or Panzer.  With this new setup , if packages that SPARC does not use fail to build, it will not stop the other packages SPARC does use from getting correctly built and installed.  This therefore decouples the SPARC and EMPIRE packages in the testing and installation process.  Therefore, we can get away with a single set of builds for SPARC, EMPIRE, and Gemma.

## How this was tested:

There is strong testing for this in TriBITS but the one Trilinos-specific commit https://github.com/trilinos/Trilinos/commit/11fc4d30af47e3642b8653aa4e1a3b972bf0a6d1 was tested locally just to make sure the variable `Trilinos_MAKE_INSTALL_WORLD_READABLE=ON` is getting set correctly.  I set the `umask` to `u=rwx,g=,o=` and then did the `make install` and it correctly set the permissions on the created dirs to group and world readable.


bartlettroscoe:
Thanks @rppawlo !

PR #4947, opened 2019-04-18
Adding the unit tests for the clean_workspace tool

prwolfe:
As we discussed getting these to always run is vital.
I will be working on python 2/3 coexistence next.

@trilinos/framework 

## Description
Just adding the make spec to get these python tests running.

## Motivation and Context
All testing should happen at PR time as well as by hand.

## How Has This Been Tested?
Built as if in a PR test and ran the python tests by hand.


prwolfe:
This is not likely to go anytime soon - it looks like the mock package is not installed in the sems python 2 packaging and there is a known bug in the modules python script that keeps modules from loading under python 3.  I will petition to get mock installed.

prwolfe:
@bartlettroscoe - I have been looking, but I can't figure out what caused these to be reported as Zoltan2 tests. Can I buy or beg a clue off of you?

bartlettroscoe:
@prwolfe asked:

>  I have been looking, but I can't figure out what caused these to be reported as Zoltan2 tests. Can I buy or beg a clue off of you?

It is because the outer ctest -S script did not define TrilinosFrameworkTests as a proper CDash Subproject and therefore CDash seems to assign it to some random subproject (as far as I can tell).

I am guessing that list of PT packages defining CDash SubProjects is not being automatically assembled automatically like it is in the [`tribits_ctest_driver()`](https://tribits.org/doc/TribitsDevelopersGuide.html#tribits-ctest-driver) function?

Would this be a good time to consider switching the the Trilinos PR tester to use `tribits_ctest_driver()` like the rest of Trilinos testing?  That function now has very strong automated testing and it is getting extended regularly.

cmake/std/PullRequestLinuxGCC7.2.0TestingSettings.cmake
```diff
@@ -20,6 +20,10 @@ set (Trilinos_ENABLE_COMPLEX_DOUBLE ON CACHE BOOL "Set by default for PR testing
 # Disable just one Teko sub-unit test that fails with openmpi 1.10 (#2712)
 set (Teko_DISABLE_LSCSTABALIZED_TPETRA_ALPAH_INV_D ON CACHE BOOL "Temporarily disabled in PR testing")
 
+# Disable the test of Modules in python for python 3 - this build only.
+# re-enable when python 3 and modules decide to play nice (likley RHEL8)
+set (TrilinosFrameworkTests_Modules_UnitTests_DISABLE ON CACHE BOOL "Temporarily disabled in PR testing")
```
bartlettroscoe: @prwolfe can we get these tests working before merging this PR?  It would be nice to protect this stuff in PR testing if possible.

prwolfe: The Teko test has been disabled for some time. The Modules.py test will not run under python3 until the modules package fixes their python interface.  I could mock that out if it would import, but it won't.

I have a similar issue with your python usage in the get_changed_trilinos_packages, but that is due to an issue with the sems installation - I have a ticket open with them, Do you expect the rest of your code to pass testing under python3 or is this an unknown quantity?

bartlettroscoe: @prwolfe, we can fix the code to run with python 3 as well.  Why is python 2 being used in some PR builds and python 3 being used in others?

prwolfe: We are trying to get ahead of this issue - python 3 will be the default in RHEL 8 so I was attempting to add tests in 3. Not urgent by any means.

prwolfe:
took off the WIP flag so that I can test this, but there is no corresponding job in the autotester or Jenkins so do not merge at this point.

prwolfe:
@william76 - this is the PR I mentioned in todays standup. I think it will now have everything needed to set up Jenkins Jobs for the python testing and I will then add a PR to change the auto tester to call those.

cmake/std/sems/PullRequestPython3TestingEnv.sh
```diff
@@ -0,0 +1,48 @@
...
+# SEMS does not yet supply a recent enough version of CMake
+# for the single configure/build/test capability. We are also
+# using a custom version of Ninja (with Fortran support not
+# available in main-line Ninja) to significantly speed up
+# compile and link times.
+module load atdm-env
+module load atdm-ninja_fortran/1.7.2
+
+# we will have implicitly gotten the sems python from
+# the boost module above for whaever reason - reset it
```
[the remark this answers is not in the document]
prwolfe: If it's in use, yes. I would have left selection of the python version to use to the developer...

prwolfe:
Ok, I got this to the point that the python tests come up and run but do not pass. I would like to merge it the way it is and then do a second PR to work out those failures and enable these as a regular part of PR testing.

PR #4972, opened 2019-04-19
Amesos2,Tpetra: Fix #3752 (remove HAVE_STD_NEW_COUNT_SYNTAX macro)

### LINKED ISSUE #3752: Amesos2,Tpetra: Remove HAVE_STD_NEW_COUNT_SYNTAX macro
@trilinos/amesos2 @trilinos/tpetra 
```
$ git grep HAVE_STD_NEW_COUNT_SYNTAX
packages/amesos2/cmake/Amesos2_config.h.in:#cmakedefine HAVE_STD_NEW_COUNT_SYNTAX
packages/tpetra/core/cmake/TpetraCore_config.h.in:#cmakedefine HAVE_STD_NEW_COUNT_SYNTAX
```
Nothing in Trilinos uses the `HAVE_STD_NEW_COUNT_SYNTAX` macro, in either package.  No CMake code in Trilinos defines this macro.  Furthermore, since both packages enable exactly the same macro, the macro could only possibly be defined if exactly one of the two packages is enabled.  Otherwise, a build error would result.  However, the Dashboard tests and users regularly build with both packages enabled.  This suggests that neither package actually ever defines this macro.

`git blame` on the Tpetra file shows that the macro definition is almost 10 years old:
```
551d96784e4 packages/tpetra/cmake/Tpetra_config.h.in          (Roscoe A. Bartlett   2008-11-18 05:23:16 +0000   5) /* define if new form of std::count is supported */
c642d67612f packages/tpetra/cmake/Tpetra_config.h.in          (Roscoe A. Bartlett   2008-11-24 03:44:00 +0000   6) #cmakedefine HAVE_STD_NEW_COUNT_SYNTAX
```
My guess is that Ross added this macro for a legitimate purpose (likely relating to a particular Standard Template Library implementation) long ago.  Someone else must have removed the CMake option that would have enabled this code, once the option was no longer useful.  However, whoever did that never changed `TpetraCore_config.h.in` to remove the macro corresponding to that CMake option.

The Amesos2 file may have been a copy-and-paste directly from Tpetra.  It never caused a build error, since the option must never have been enabled as of first introduction of Amesos2 into Trilinos, in 2011.  `git blame` shows that this part of the Amesos2 file has not changed since Eric Bavier checked in it, in 2010.  (The fact that Eric changed the surrounding lines of the file in 2011 supports the copy-and-paste theory.)

All of this strongly suggests that we can safely remove the macro.

mhoemmen:
@trilinos/amesos2 @trilinos/tpetra

## Description

Remove definition of `HAVE_STD_NEW_COUNT_SYNTAX` macro from Amesos2 and Tpetra.

## Related Issues

* Closes #3752

### DIFF
--- a/packages/amesos2/cmake/Amesos2_config.h.in
+++ b/packages/amesos2/cmake/Amesos2_config.h.in
@@ -54,9 +54,6 @@
  * Config Options *
  ******************/
 
-/* define if new form of std::count is supported */
-#cmakedefine HAVE_STD_NEW_COUNT_SYNTAX
-
 /* Define if want to build amesos2-debug */
 #cmakedefine HAVE_AMESOS2_DEBUG
 
--- a/packages/tpetra/core/cmake/TpetraCore_config.h.in
+++ b/packages/tpetra/core/cmake/TpetraCore_config.h.in
@@ -5,9 +5,6 @@
 /* Define whether Tpetra enables deprecated code at compile time. */
 #cmakedefine TPETRA_ENABLE_DEPRECATED_CODE
 
-/* define if new form of std::count is supported */
-#cmakedefine HAVE_STD_NEW_COUNT_SYNTAX
-
 /* Define if want to build with epetra enabled */
 #cmakedefine HAVE_TPETRACORE_EPETRA
 #ifdef HAVE_TPETRACORE_EPETRA


mhoemmen:
Thanks @csiefer2 ! :D

PR #4977, opened 2019-04-22
Reduce parallel build jobs for Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-rdc-release-debug-pt bulids

bartlettroscoe:
The build `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-rdc-release-debug-pt` appears to be crashing the nodes on 'white' and 'ride' lately (see [ATDV-155](https://sems-atlassian-srn.sandia.gov/browse/ATDV-155)).  This reduces the overall parallel level from 32 to 20 and the link level from 16 to 10 in hopes this will avoid crashing the nodes on 'white' and 'ride'.  If this is not enough of a change, I will reduce this even more.  Even at 20 and 10, this will make this build take upwards of likely 7 hours, just for the build!


cmake/std/atdm/ride/environment.sh
```diff
@@ -131,10 +131,20 @@ export ATDM_CONFIG_USE_NINJA=ON
...
+  # When CUDA+RDC is enabled, it uses more RAM bulding with more build
+  # processes crashes nodes on 'white' and 'ride (see #4502)
+  if [[ "${ATDM_CONFIG_PT_PACKAGES}" == "ON" ]] ; then
+    # Building with all PT packages crashes nodes on 'white' and 'ride' for
+    # the same parallel levels that work for just the ATDM packages below.
+    # See ATDV-155.
+    export ATDM_CONFIG_BUILD_COUNT=20
+    export ATDM_CONFIG_PARALLEL_LINK_JOBS_LIMIT=10
+  else
+    # We seem to be able to use more buld processes when just ATDM packages
```
mhoemmen: "build"

bartlettroscoe: Thanks for catching @mhoemmen!  Just pushed a fix.

Issue #3234, opened 2018-08-03
Amesos2 fails to build with Tpetra_INST_INT_INT=OFF & Amesos2_ENABLE_Epetra=OFF


mhoemmen:
@micahahoward reports:
```
CUDA is still failing for me.  I think this is because we have Tpetra_INST_INT_INT turned off now.  See this:
.../Trilinos/.../libamesos2.a(Amesos2_Superludist.cpp.o): In function `Amesos2::ConcreteMatrixAdapter<Tpetra::CrsMatrix<double, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> > >::get_impl(Teuchos::Ptr<Tpetra::Map<int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> > const>, Amesos2::EDistribution) const'
```
Amesos2 used to build (and should build) just fine a week or two ago, with `Tpetra_INST_INT_INT=OFF` and `Amesos2_ENABLE_Epetra=OFF`.  (See #2548 discussion.)  

@trilinos/amesos2 @trilinos/tpetra @tcfisher 

## Expectations

All Trilinos packages that use Tpetra should respect what Tpetra enables and disables.

## Motivation and Context

SPARC needs Amesos2 to build with these options set, because otherwise the linker fails (static libraries get too large when we set `Tpetra_INST_INT_INT=ON`).

## Related Issues
* Blocks #2548

mhoemmen:
Fixed in develop by PR #3254 being merged.

PR #4994, opened 2019-04-23
Amesos2: Fix #4992

### LINKED ISSUE #4992: Amesos2 does not support Epetra with Tpetra_INST_INT_INT=OFF
## Bug Report
@trilinos/amesos2 @rppawlo @kddevin 

### Description

Amesos2 does not support Epetra with `Tpetra_INST_INT_INT=OFF`.

### Steps to Reproduce

1. Set `Tpetra_INST_INT_INT=OFF` and `Amesos2_ENABLE_Epetra=ON`
2. Watch Amesos2 complain and fail to implement the feature

This is related to #3234.  That issue was about Amesos2 failing to catch its failure to implement this feature.

mhoemmen:
@trilinos/amesos2

## Description

- Permit Amesos2_ENABLE_Epetra=ON with Tpetra_INST_INT_INT=OFF.
- Fix Amesos2's build for that case.

This took me about an hour to do and required only trivial index conversions.

## Motivation and Context

@rppawlo needs this for Empire.  Tpetra developers want to deprecate and remove template parameters like GlobalOrdinal.  It's trivial to do index conversions, especially for a package like Amesos2, where the solves will take much longer than any index conversions that need to happen.

## Related Issues

* Closes #4992 
* Related to #3234

### DIFF
--- a/packages/amesos2/src/Amesos2_PardisoMKL.cpp
+++ b/packages/amesos2/src/Amesos2_PardisoMKL.cpp
@@ -53,10 +53,8 @@
 
 namespace Amesos2 {
 
-#ifdef HAVE_TPETRA_INST_INT_INT
 #ifdef HAVE_AMESOS2_EPETRA
   AMESOS2_SOLVER_EPETRA_INST(PardisoMKL);
-#endif
 #endif
 
 }
@@ -307,7 +305,7 @@ TPETRA_ETI_MANGLING_TYPEDEFS()
   #endif
 #endif
 #ifdef HAVE_TPETRA_INST_COMPLEX_DOUBLE
-    #ifdef HAVE_TPETRA_INST_INT_INT 
+    #ifdef HAVE_TPETRA_INST_INT_INT
      AMESOS2_PARDISOMKL_LOCAL_INSTANT(std::complex<double>, int, int, NODETYPE)
     #endif
     #ifdef HAVE_TPETRA_INST_INT_LONG
--- a/packages/amesos2/src/Amesos2_Util.cpp
+++ b/packages/amesos2/src/Amesos2_Util.cpp
@@ -50,16 +50,16 @@
 #  include <Teuchos_DefaultMpiComm.hpp>
 #  ifdef HAVE_AMESOS2_EPETRA
 #    include <Teuchos_OpaqueWrapper.hpp>
-#  endif
-#endif
+#  endif // HAVE_AMESOS2_EPETRA
+#endif // HAVE_MPI
 
 #ifdef HAVE_AMESOS2_EPETRA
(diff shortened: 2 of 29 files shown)


packages/amesos2/src/Amesos2_PardisoMKL.cpp
```diff
@@ -307,7 +305,7 @@ TPETRA_ETI_MANGLING_TYPEDEFS()
   #endif
 #endif
 #ifdef HAVE_TPETRA_INST_COMPLEX_DOUBLE
-    #ifdef HAVE_TPETRA_INST_INT_INT
```
ndellingwood: @mhoemmen looks like you missed removing the corresponding `#endif` here

packages/amesos2/src/Amesos2_Util.cpp
```diff
@@ -112,12 +109,11 @@ Amesos2::Util::to_epetra_comm(Teuchos::RCP<const Teuchos::Comm<int> > c)
     Teuchos::RCP<const Epetra_SerialComm> serialComm = rcp(new Epetra_SerialComm());
     return serialComm;
   }
-#endif	// HAVE_MPI
+#endif  // HAVE_MPI
```
ndellingwood: @mhoemmen were there missing `#endif`s in this file? Asking since the number of `#ifdef` removed does not match removed `#endif`s

packages/amesos2/src/Amesos2_PardisoMKL.cpp
mhoemmen: Thanks @ndellingwood !  I'll work on this.

mhoemmen: @ndellingwood I think this one looks fine -- thanks for checking, though!

packages/amesos2/src/Amesos2_Util.cpp
mhoemmen: @ndellingwood In my changed version, the counts of `#ifdef`s and `#endif`s match.  It's possible that the code was not correct before.

mhoemmen:
@ndellingwood I think my changes are OK -- I pushed a new commit that adds some comments to make ifdef / endif pairs more clear in that one file.  Could you please rereview?  Thanks!

mhoemmen:
Failure is because I marked this as WIP, which killed the builds.

packages/amesos2/src/Amesos2_Util.cpp
ndellingwood: Oh goodness, thanks for clarifying!

mhoemmen:
Thanks @ndellingwood ! :D

PR #4996, opened 2019-04-24
Xpetra: start of ETI system

jhux2:
This is the beginning of an ETI system for Xpetra.

@trilinos/xpetra @jjellio

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/4996)


csiefer2:
There's some weird stuff in this branch (changes to SEACAS?)

csiefer2:
Added #4998 which should be easier to see the relevant changes than this PR.

packages/muelu/src/Utils/MueLu_TimeMonitor.cpp
```diff
@@ -0,0 +1,112 @@
...
+        // TODO: there is no function to register a timer in Teuchos::TimeMonitor after the creation of the timer. But would be useful...
+        timer_ = Teuchos::TimeMonitor::getNewTimer("MueLu: " + msg);
+      } else {
+        timer_ = rcp(new Teuchos::Time("MueLu: " + msg));
+      }
+
+      // Start the timer (this is what is done by Teuchos::TimeMonitor)
+      timer_->start();
+      timer_->incrementNumCalls();
+#ifdef HAVE_TEUCHOS_ADD_TIME_MONITOR_TO_STACKED_TIMER
```
mhoemmen: This really feels like it could be a run-time option, given the amount of overhead already in here.  However, that's a separate issue from this PR.

packages/muelu/src/Utils/MueLu_TimeMonitor.cpp
```diff
@@ -0,0 +1,112 @@
...
+      const auto stackedTimer = Teuchos::TimeMonitor::getStackedTimer();
+      if (nonnull(stackedTimer))
+        stackedTimer->start(timer_->name());
+#endif
+    }
+  } //TimeMonitor::TimeMonitor()
+
+  TimeMonitor::TimeMonitor() {}
```
mhoemmen: Prefer replacing trivial default constructor definitions with `TimeMonitor() = default;` in the class declaration.  It would also make sense to forbid copy and move construction and copy and move assignment, using `=delete`.  (That would prevent handling TimeMonitor by RCP, but I see that as a good thing ;-) .)

packages/muelu/src/Utils/MueLu_TimeMonitor.hpp
```diff
@@ -183,6 +130,9 @@ namespace MueLu {
     RCP<MutuallyExclusiveTime<TagName> > timer_; // keep a reference on the timer to print stats if RuntimeTimings=ON //TODO:use base class instead
   };
 
+  extern template class MutuallyExclusiveTimeMonitor<FactoryBase>;
```
mhoemmen: A+++

packages/seacas/applications/mapvar/mapvar.f
```diff
@@ -430,7 +430,7 @@ PROGRAM MAPVAR
      *      ' handle interpolating a mesh with nodal variables',
      *      /,9x,'and multiple element blocks. Shared nodes',
      *      ' might be incorrect.',
-     *      /,9x'See https://github.com/gsjaardema/seacas/packages'
+     *      /,9x,'See https://github.com/gsjaardema/seacas/packages'
```
mhoemmen: I recognize these changes -- it looks like you've brought in stuff from another PR.  Did you mean to do that?

packages/tempus/src/Tempus_TimeStepControl_decl.hpp
```diff
@@ -117,6 +117,8 @@ class TimeStepControl
       { return tscPL_->get<int>   ("Maximum Order"); }
     virtual std::string getStepType() const
       { return tscPL_->get<std::string>("Integrator Step Type"); }
+    virtual bool getOutputExactly() const
```
mhoemmen: Did you mean to include Tempus changes?  It looks like something got mixed up with a rebase here.

packages/xpetra/src/CMakeLists.txt
```diff
@@ -16,9 +16,88 @@ TRIBITS_CONFIGURE_FILE(${PACKAGE_NAME}_config.hpp)
...
+      string(REPLACE "]" ")" CONDITION_NAME ${CONDITION_NAME})
+      string(REPLACE "." " " CONDITION_NAME ${CONDITION_NAME})
+      string(REPLACE "-" "" CONDITION_NAME ${CONDITION_NAME})
+      string(REPLACE "?" "#" CONDITION_NAME ${CONDITION_NAME})
+      string(STRIP CONDITION_NAME ${CONDITION_NAME})
+      #MESSAGE("New condition name = ${CONDITION_NAME}")
+      set (CONDITION_NAME_END "#endif")
+    ENDIF()
+    
+    string(REPLACE "::" "_" CLASS_FILE_NAME "${CLASS_NAME}")
```
mhoemmen: I wonder if we should move the "mangle class name for use in macro" function (rewritten in a few places, perhaps with differences) up into Teuchos, where different packages could use it.

packages/xpetra/src/CMakeLists.txt
```diff
@@ -37,6 +116,9 @@ IF (${PACKAGE_NAME}_ENABLE_Epetra)
     )
 ENDIF()
 
+MESSAGE("JHU: CMAKE_CURRENT_SOURCE_DIR=${CMAKE_CURRENT_SOURCE_DIR}")
+TRILINOS_CREATE_CLIENT_TEMPLATE_HEADERS(${CMAKE_CURRENT_SOURCE_DIR}/CrsMatrix NOSIERRABJAM)
```
mhoemmen: Sierra now uses CMake, not bjam, to set up Trilinos' build.  That could be relevant to this `NOSIERRABJAM` option, depending on what it means.

packages/xpetra/src/CMakeLists.txt
```diff
@@ -37,6 +116,9 @@ IF (${PACKAGE_NAME}_ENABLE_Epetra)
     )
 ENDIF()
 
+MESSAGE("JHU: CMAKE_CURRENT_SOURCE_DIR=${CMAKE_CURRENT_SOURCE_DIR}")
```
mhoemmen: I think there's a way to mark a PR as a "draft."

packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix_decl.hpp
```diff
@@ -102,102 +102,49 @@ namespace Xpetra {
...
-      bool restrictComm=false;
-      if(!params.is_null()) restrictComm = params->get("Restrict Communicator",restrictComm);
-      if(restrictComm && mtx_->getRowMap().is_null()) mtx_=Teuchos::null;
-    }
+                    const Teuchos::RCP<Teuchos::ParameterList>& params);
 
     //! Constructor for a fused export
     TpetraCrsMatrix(const Teuchos::RCP<const CrsMatrix<Scalar,LocalOrdinal,GlobalOrdinal,Node> >& sourceMatrix,
                     const Export<LocalOrdinal,GlobalOrdinal,Node> & RowExporter,
                     const Teuchos::RCP<const Export<LocalOrdinal,GlobalOrdinal,Node> > DomainExporter,
```
mhoemmen: I know this is preexisting, but did you mean to pass the RCP by const value rather than by const reference (like the other RCPs)?  It shouldn't matter for performance, but it's not consistent and that could introduce build errors later.

packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix_decl.hpp
```diff
@@ -224,8 +171,7 @@ namespace Xpetra {
     TpetraCrsMatrix (const Teuchos::RCP<const Map<LocalOrdinal,GlobalOrdinal,Node> >& rowMap,
         const Teuchos::RCP<const Map<LocalOrdinal,GlobalOrdinal,Node> >& colMap,
         const local_matrix_type& lclMatrix,
-        const Teuchos::RCP<Teuchos::ParameterList>& params = null)
-    : mtx_(Teuchos::rcp(new Tpetra::CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node>(toTpetra(rowMap), toTpetra(colMap), lclMatrix, params))) {  }
+        const Teuchos::RCP<Teuchos::ParameterList>& params = null);
```
mhoemmen: This is a great opportunity to purge header file includes (e.g., for `toTpetra`) from the decl header.

packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix_decl.hpp
```diff
@@ -234,315 +180,229 @@ namespace Xpetra {
...
         const Teuchos::RCP<const Map<LocalOrdinal,GlobalOrdinal,Node> >& rangeMap = Teuchos::null,
-        const Teuchos::RCP<Teuchos::ParameterList>& params = null)
-    : mtx_(Teuchos::rcp(new Tpetra::CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node>(lclMatrix, toTpetra(rowMap), toTpetra(colMap), toTpetra(domainMap), toTpetra(rangeMap), params))) {  }
+        const Teuchos::RCP<Teuchos::ParameterList>& params = null);
 #endif
 #endif
 
     //! Destructor.
-    virtual ~TpetraCrsMatrix() {  }
+    virtual ~TpetraCrsMatrix();
```
mhoemmen: If the destructor body is empty, you could (and should) just write this:
```c++
virtual ~TpetraCrsMatrix() = default;
```

packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix_def.hpp
```diff
@@ -0,0 +1,473 @@
...
+//                    Andrey Prokopenko (aprokop@sandia.gov)
+//                    Ray Tuminaro      (rstumin@sandia.gov)
+//
+// ***********************************************************************
+//
+// @HEADER
+#ifndef XPETRA_TPETRACRSMATRIX_DEF_HPP
+#define XPETRA_TPETRACRSMATRIX_DEF_HPP
+
+#include "Xpetra_TpetraCrsMatrix_decl.hpp"
```
mhoemmen: The `_def.hpp` headers should not need to include the `_decl.hpp` headers, for the following reasons:

  1. Automatically generated `.hpp` header files will do the right thing, whether or not ETI is enabled.  If ETI is ON, they include only the corresponding decl file.  If ETI is OFF, they include first the decl file, then the def file.
  2. This means that if ETI is OFF, users will double-include the def file.  This has a nonzero cost (the C preprocessor generally must reread the file).

Removing the include means that your automatically generated .cpp ETI files must include both the decl and the def header files.  Right now, they just include the def header.

packages/xpetra/src/Utils/ClassList/LO-GO-NO.classList
```diff
@@ -0,0 +1,19 @@
+##BlockMap
```
lucbv: Is the double hash symbol there on purpose?

csiefer2:
@jhux2  Well since the PR tester woofed badly, I'll sneak in the branch cleanup.

csiefer2:
All of the crazy seacase stuff should be gone from the diffs now.

csiefer2:
I have a sinking suspicion that Stokhos needs to have ETI for Xpetra for this to work :(

@etphipp

etphipp:
Yep.

csiefer2:
@etphipp Any chance we can get your help to stub that out?

etphipp:
Yes, I can help with that.  It will probably be a day or two before I can get to it though.

I assume Xpetra is using a similar system to MueLu?  If so, this shouldn't be too hard.

csiefer2:
@etphipp I believe that @jhux2 copied MueLu's ETI system, yes.

csiefer2:
@jhux2 Is there a reason why the partial specialization stuff is still in the decl and not in the def?

jhux2:
@csiefer2 did a count of symbols in `libmuelu.a`.  Here are the top offenders:
```
   7727  Xpetra::BlockedVector
   7225  Xpetra::BlockedMultiVector
   6657  Xpetra::TpetraMultiVector
   5100  Xpetra::EpetraCrsMatrixT
   4944  Xpetra::BlockedMap
   4874  Xpetra::TpetraMap
   4705  Xpetra::EpetraMapT
   4585  Xpetra::CrsMatrix
   2601  Xpetra::TpetraVector
   2545  Xpetra::EpetraMultiVectorT
   2378  Xpetra::StridedMap
   2144  Xpetra::TpetraImport
   2144  Xpetra::EpetraImportT
   1521  Xpetra::BlockedCrsMatrix
   1046  Xpetra::Matrix
```

jhux2:
> Is there a reason why the partial specialization stuff is still in the decl and not in the def?

No good one.  I wanted to push something that has a chance of working.

packages/xpetra/src/Utils/ClassList/LO-GO-NO.classList
jhux2: @lucbv A few classes were commented out already. I first commented out everything (hence the double hash), then uncommented the one class I was ETI'ing.

packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix_def.hpp
jhux2: Huh -- doesn't the *_def.hpp need to know about the class declaration?

packages/xpetra/src/CMakeLists.txt
jhux2: Woops, this was a debugging message, it should be deleted.

packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix_def.hpp
mhoemmen: @jhux2 Nothing ever includes the `${FOO}_def.hpp` file by itself, without `${FOO}_decl.hpp` first having been included.  Here are the possible scenarios:

1. With ETI OFF, user includes `${FOO}.hpp`, which includes first `${FOO}_decl.hpp`, then `${FOO}_def.hpp`.
2. With ETI ON, user includes `${FOO}.hpp`, which includes just `${FOO}_decl.hpp`.
3. User includes just `${FOO}_decl.hpp` for the class declaration.  At some later point, user may include `${FOO}.hpp`.  See above.
4. With ETI ON, the ETI `.cpp` file includes first `${FOO}_decl.hpp`, then `${FOO}_def.hpp`.

jjellio: @mhoemmen You are assuming CMake generated `Foo.hpp`, and that the source is partitioned into `Foo_decl.hpp` and `Foo_def.hpp`. Seems your comment is really rooted in how CMake enables/disables ETI.

How does MueLu/Xpetra handle this?  When I grepped the source, I saw lots of "Foo.hpp", which seems to indicate those files are not generated by CMake.

etphipp:
Actually it was pretty easy to add the ETI to Stokhos, and I just pushed a commit for that.  Based on my testing, that resolves the linker issues.  It reads the same cmake file with the list of xpetra classes to instantiate as xpetra, so Stokhos will pick those up automatically when classes are added.

csiefer2:
Awesome!  Thanks @etphipp

packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix_decl.hpp
jhux2: Looking through MueLu's library, I find many destructor weak symbols, e.g., `MueLu::Monitor::~Monitor()`.  I can get rid of these by declaring a nontrivial destructor:
```
// hpp
~Monitor();
// cpp
Monitor::~Monitor() {}
```

Four such destructors are about 3 MB in the MueLu library.

Question:  what are the downsides (if any) to doing this?

jjellio: Jonathan, isn't that a question about marking destructors virtual?

https://stackoverflow.com/questions/7403883/derived-class-with-non-virtual-destructor


Just doing:
```
// hpp
~Monitor() = default;
```
 Should be the same as ` defining w/ {}`

jhux2: @jjellio It appears they are not the same.

If I do
```
// hpp
~FactoryMonitor() = default;
```
then I see many symbols in `libmuelu.a` of the form
```
W MueLu::FactoryMonitor::~FactoryMonitor()
```

If I do
```
// hpp
~FactoryMonitor();
// cpp
FactoryMonitor::~FactoryMonitor();
```
then in `libmuelu.a`, I see only
```
U MueLu::FactoryMonitor::~FactoryMonitor()
```

More importantly, in the latter case, the library is 1.1 MB smaller.

jjellio: Same as what we talked about  in email.

```
~FactoryMonitor() = default;
```
is placing the definition within the class decl, so it is inlined.

```
FactoryMonitor::~FactoryMonitor();
```

Declares the destructor, but you better provide a def in the def file. (I don't think you can define it using `default`, the only way would be to do `FactoryMonitor::~FactoryMonitor() {}` in the def, unless I am wrong)

jjellio: Btw, you could try:

in Foo_def.hpp
```
FactoryMonitor::~FactoryMonitor() = default;
```
Foo_decl.hpp
```
~FactoryMonitor();
```

jhux2: Ah, you are right.  However, back to my question.  Why would I choose to use the default ctor, which appears multiple times as a weak symbol in the library and thus increases the library's size, instead of placing it in the definition?

jjellio: default is just sugar for `{}` as I understand it.

What if you put in the def file,
```
FactoryMonitor::~FactoryMonitor() = default;
```

jjellio: I just don't know if you can use 'default' outside a class declaration ... you could always write
Foo_def.hpp
```
FactoryMonitor::~FactoryMonitor() = {}
```

jjellio: Oh, and since this is a Base Class, you I think you have to define the destructor virtual.

jhux2: > Oh, and since this is a Base Class, you I think you have to define the destructor virtual.

Yep.

csiefer2:
Almost got TpetraMultiVector working...

packages/xpetra/src/CrsMatrix/Xpetra_TpetraCrsMatrix_def.hpp
mhoemmen: @jjellio Good point -- I was assuming that `${FOO}.hpp` includes just the decl if ETI is ON, and includes first the decl and then the def if ETI is OFF.  As long as MueLu's `${FOO}.hpp` header files look like that, what I wrote above should still apply.

PR #5001, opened 2019-04-24
SEACAS: cgns - fix zgc uniquification

gsjaardema:
@trilinos/seacas 

## Description

Fix for SPARC 
 * The ZoneGridConnectivity instances need to be unique for a zone.  In the file-per-processor case, there may be multiple ZGC with the same name that arise due to processor decomposition.  The previous code assumed a maximum of 26 instances and made them unique by appending a 'A' ... 'Z' to the name until found a unique name.  Failure was not detected.

New code tries 'A' .. 'Z' and if not unique, then tries 'AA' 'AB',...'ZZ' and if still not unique it will stop and output a meaningful error message. 

Also added some debugging output for the non-file-per-processor output to make sure that zgc consolidation is working correctly

## How Has This Been Tested?

Tested on actual mesh that was causing issues with original code.  A representation of this mesh is now in test suite (which will be in later commit)

## Checklist

### DIFF
--- a/packages/seacas/libraries/ioss/src/cgns/Iocgns_Utils.C
+++ b/packages/seacas/libraries/ioss/src/cgns/Iocgns_Utils.C
@@ -606,6 +606,13 @@ namespace {
       assert(off_data / count == INT_PER_ZGC);
       assert(off_name % count == 0 && off_name / count == BYTE_PER_NAME);
 
+#if IOSS_DEBUG_OUTPUT
+      std::cerr << "ZGC_CONSOLIDATE: Before consolidation: (" << zgc.size() << ")\n";
+      for (const auto &z : zgc) {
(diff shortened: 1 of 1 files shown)


gsjaardema:
@bartlettroscoe I think SPARC prefers getting SEACAS from Trilinos.  For this fix, I provided them with patches to their Trilinos to make sure we fixed the bug and once it was confirmed, then I created this PR from the patches that they tried locally.

bartlettroscoe:
@gsjaardema said:

> For this fix, I provided them with patches to their Trilinos to make sure we fixed the bug and once it was confirmed, then I created this PR from the patches that they tried locally.

Sounds reasonable as long as these patches also get back into the main SEACAS github repo that gets snapshotted back into Trilinos.

Issue #4976, opened 2019-04-21
ifpack - add an interface for a hypre set function taking int ** arguments to Ifpack_Hypre


jthano:
## Enhancement
@trilinos/ifpack

Ifpack_Hypre does not currently include an interface for a hypre parameter set function taking an int** argument. The only such set function in hypre is HYPRE_BoomerAMGSetGridRelaxPoints and there was a comment declaring this set function obsolete. However, this function will no longer be obsolete as it is being used for AIR AMG which can be found in the AIR branch of the development hypre repo. The AIR branch will be merged into master in the near future.

I have already added an interface for int** functions in my fork of Trilinos. I am creating this issue because the contribution guidelines state an issue should be created first and then referred to in a pull request.

mhoemmen:
@trilinos/ifpack FYI, PR testing does not enable the Hypre TPL, so whatever testing you need will have to happen on your end.  PR tests will only ensure that the Hypre TPL is truly optional, that is, that Trilinos builds and passes tests with Hypre disabled.

jthano:
Okay I have been using the modified ifpack_Hypre code in my fork for a while and have tested the added functionality.  What are the steps I should complete before submitting a PR. Do I need to email a signed contribution form? Or is that not required for a single contribution. Would it be acceptable to submit a PR for this change against the primary development branch?

mhoemmen:
@jwillenbring or @trilinos/framework could comment on the contribution form issue.  Besides that, we require that PRs be submitted against the `develop` branch.  Best practice is for you to create a branch in your own fork of Trilinos, and submit the PR from that branch.  Thanks!

jwillenbring:
> @jwillenbring or @trilinos/framework could comment on the contribution form issue. Besides that, we require that PRs be submitted against the develop branch.

@jthano How extensive is the modification? It sounds modest at a high level.

jthano:
The modification is small, but not trivial. For every location in Ifpack_Hypre.cpp and Ifpack_Hypre.h related to the interface with hypre set functions, I added support for a new interface to a function with the signature of int (*)(HYPRE_Solver, int**). I made sure all comments for the new code match the style of the already existing code.

jthano:
I can go ahead and submit a PR so that the code changes can be easily viewed

PR #5003, opened 2019-04-24
Ifpack hypre set function addition

jthano:
This PR is to address issue #4976 and adds support of a set function in hypre taking an int ** argument.

The code has been tested by compiling the updated code and using the new interfaces to set values in a hypre solver object with a pointer to the HYPRE_BoomerAMGSetGridRelaxPoints function.

NOTE: I see that the commit history is cluttered. I think this is because my branch was created based on the Trilinos master branch. I think I need to make my changes based on the develop branch that I have submitted the PR against. I'll work on doing this

@trilinos/ifpack

### DIFF
--- a/packages/seacas/libraries/ioss/src/cgns/Iocgns_Utils.C
+++ b/packages/seacas/libraries/ioss/src/cgns/Iocgns_Utils.C
@@ -606,6 +606,13 @@ namespace {
       assert(off_data / count == INT_PER_ZGC);
       assert(off_name % count == 0 && off_name / count == BYTE_PER_NAME);
 
+#if IOSS_DEBUG_OUTPUT
+      std::cerr << "ZGC_CONSOLIDATE: Before consolidation: (" << zgc.size() << ")\n";
+      for (const auto &z : zgc) {
+        std::cerr << "\tOZ " << z.m_ownerZone << z << "\n";
+      }
+#endif
+
       // Consolidate down to the minimum set that has the union of all ranges.
       for (size_t i = 0; i < zgc.size(); i++) {
         if (zgc[i].m_ownerZone > 0 && zgc[i].m_donorZone > 0) {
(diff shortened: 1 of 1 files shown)


mhoemmen:
@trilinos/ifpack

packages/ifpack/src/Ifpack_Hypre.h
```diff
@@ -140,36 +140,47 @@ class FunctionParameter{
...
+            return double_func_(solver, double_param1_);
+          } else if(option_ == 2){
+            return double_int_func_(solver, double_param1_, int_param1_);
+          } else if (option_ == 3){
+            return int_int_func_(solver, int_param1_, int_param2_);
+          } else if (option_ == 4){
+            return int_star_func_(solver, int_star_param_);
+          } else if (option_ == 5){
+            return double_star_func_(solver, double_star_param_);
+          } else {
```
mhoemmen: It would be nicer to test `option_ == 6`, and to have an `else` that catches unsupported `option_` values.  However, I would accept this PR without that.

packages/ifpack/src/Ifpack_Hypre.h
```diff
@@ -326,6 +339,18 @@ class Ifpack_Hypre: public Ifpack_Preconditioner {
    */
     int SetParameter(Hypre_Chooser chooser, int (*pt2Func)(HYPRE_Solver, int*), int* parameter);
 
+    //! Set a parameter that takes an int**.
+    /*!
+    \param chooser (In) -A Hypre_Chooser enumerated type set to Solver or Preconditioner, whatever the parameter is setting for.
```
mhoemmen: I'm not sure if this is correct Doxygen syntax, but it's consistent with what Ifpack does.

packages/ifpack/src/Ifpack_Hypre.h
jthano: I thought I needed to rebase on devlop branch and just did that. But I see that this PR was already approved before I force updated by local branch

jthano:
I forced push to rebase by changes on the develop branch instead of master because there were a number of commits in master that were not in develop. My branch was original created from master. But I did not realize the PR had already been approved before I force pushed. I have now reset head to where it was when the PR was approved. Should the PR be reopened so that I can rebase my branch off of develop?

jthano:
@mhoemmen I apologize for killing this PR with my bad force update. Can I just submit a new PR? It would look exactly the same as this did. I'm not sure there is a way to fix this one.

mhoemmen:
@jthano No need to apologize!  "Approved" does not mean "merged."  It just means that if the PR passes tests, then it can be merged.

mhoemmen:
@jthano It's up to you -- you could either rebase and force-push to your branch, or you could open a new PR.

jthano:
Note see PR #5012

PR #5012, opened 2019-04-25
Ifpack: Fix #4976: add interface for hypre set function taking an int** argument. This…

### LINKED ISSUE #4976: ifpack - add an interface for a hypre set function taking int ** arguments to Ifpack_Hypre

jthano:
… is added for HYPRE_BoomerAMGSetGridRelaxPoints which is used primarily for AIR AMG
@trilinos/ifpack

The code has been tested by compiling the updated code and using the new interfaces to set values in a hypre solver object with a pointer to the HYPRE_BoomerAMGSetGridRelaxPoints function.

This PR addresses issue #4976. It is also related to PR #5003 which included the same code changes. The previous PR was for a branch based on master instead of the develop branch.

Closes #4976.

### DIFF
--- a/packages/ifpack/src/Ifpack_Hypre.cpp
+++ b/packages/ifpack/src/Ifpack_Hypre.cpp
@@ -252,6 +252,13 @@ int Ifpack_Hypre::SetParameter(Hypre_Chooser chooser, int (*pt2Func)(HYPRE_Solve
 } //SetParameter() - int* function pointer
 
 //==============================================================================
+int Ifpack_Hypre::SetParameter(Hypre_Chooser chooser, int (*pt2Func)(HYPRE_Solver, int**), int** parameter){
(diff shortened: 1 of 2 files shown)


mhoemmen:
@jthano I took the liberty of changing your title and adding the words "Closes #4976" to your description.  This will automatically close the issue once the fix reaches the master branch.

PR #4924, opened 2019-04-16
Tpetra / Xpetra: Deprecating the Clones

csiefer2:
Let the PR tester speak!  Hear its electronic voice!

Issue #4900, opened 2019-04-15
Tpetra/Xpetra/MueLu: Purge the Clones


csiefer2:
See discussion in #4893

mhoemmen:
@csiefer2 Do we want to try doing this for the minor release?  @kddevin thinks that it's too late, but perhaps this is just a matter of deleting a bunch of code that barely gets tested and never actually gets used.

csiefer2:
I'm taking a swing at deprecating clone().  We'll see what happens when I compile w/ deprecated code off.

csiefer2:
#4912

kddevin:
@csiefer2 
With Tpetra_ENABLE_DEPRECATED_CODE=OFF, the develop branch produces the following compilation errors.  Do you have time to fix this soon (as it blocks builds with Tpetra_ENABLE_DEPRECATED_CODE=OFF)?  If not, let me know and I'll try.  Thanks.

```
[ 62%] Building CXX object packages/tpetra/core/test/CrsMatrix/CMakeFiles/TpetraCore_CrsMatrix_NodeConversion.dir/CrsMatrix_NodeConversion.cpp.o
/Users/kddevin/tmp/tmp/Trilinos/packages/tpetra/core/test/CrsMatrix/CrsMatrix_NodeConversion.cpp:133:35: error: 
      'clone' following the 'template' keyword does not refer to a template
      RCP<Mat2> A2 = A1->template clone<N2>(n2,plClone);
                         ~~~~~~~~ ^~~~~
/Users/kddevin/tmp/tmp/Trilinos/packages/tpetra/core/test/CrsMatrix/CrsMatrix_NodeConversion.cpp:133:40: error: 
      expected unqualified-id
      RCP<Mat2> A2 = A1->template clone<N2>(n2,plClone);
                                       ^
/Users//tmp/tmp/Trilinos/packages/tpetra/core/test/CrsMatrix/CrsMatrix_NodeConversion.cpp:133:41: error: 
      'N2' does not refer to a value
      RCP<Mat2> A2 = A1->template clone<N2>(n2,plClone);
                                        ^
/Users//tmp/tmp/Trilinos/packages/tpetra/core/test/CrsMatrix/CrsMatrix_NodeConversion.cpp:104:65: note: 
      declared here
  TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( CrsMatrix, NodeConversion, N2 )
                                                                ^
[...]

PR #5021, opened 2019-04-26
Tpetra deprecate node args get node

kddevin:
@trilinos/muelu :  please look at the two test failures called out in detail below; they fail only with Tpetra_ENABLE_DEPRECATED_CODE=OFF.

@trilinos/tpetra @trilinos/stokhos @trilinos/xpetra @trilinos/ifpack2 @trilinos/belos @trilinos/thyra @trilinos/trilinoscouplings 

## Description
Deprecating getNode() in Tpetra classes
Deprecations propagate to Xpetra and Ifpack2, and use cases in belos, trilinoscouplings, thyra and stokhos

## Motivation and Context
The Node object is deprecated in Trilinos.

## Related Issues
#4932 #4924 #4888 #4839 

## How Has This Been Tested?
On my workstation, with Tpetra deprecated code enabled:
```
cmake \
-D CMAKE_BUILD_TYPE:STRING="DEBUG" \
-D CMAKE_VERBOSE_MAKEFILE:BOOL=OFF \
\
-D MPI_BIN_DIR:PATH="/Users//InstalledSoftware/openmpi-1.8.3_clang/bin" \
-D TPL_ENABLE_MPI:BOOL=ON \
-D MPI_EXEC_MAX_NUMPROCS:STRING=11 \
\
-D TPL_ENABLE_BinUtils:BOOL=OFF \
-D TPL_ENABLE_Pthread:BOOL=OFF \
\
-D CMAKE_C_FLAGS:STRING="-Wall  -pedantic -Wno-unknown-pragmas -Wno-narrowing -Wno-inline -Wshadow -Wdeprecated-declarations -Wempty-body  -Wignored-qualifiers -Wmissing-field-initializers  -Wsign-compare  -Wtype-limits   -Wuninitialized -Winit-self -fstrict-aliasing -Wno-long-long" \
-D CMAKE_CXX_FLAGS:STRING="-Wall -pedantic -Wno-unknown-pragmas -Wno-narrowing -Wno-delete-non-virtual-dtor -Wno-inline -Wshadow -Wdeprecated-declarations -Wempty-body  -Wignored-qualifiers -Wmissing-field-initializers  -Wsign-compare  -Wtype-limits   -Wuninitialized -Winit-self -fstrict-aliasing" \
\
-D Trilinos_ENABLE_ALL_OPTIONAL_PACKAGES:BOOL=ON \
\
-D Trilinos_ENABLE_SHADOW_WARNINGS:BOOL=ON \
-D Trilinos_VERBOSE_CONFIGURE:BOOL=OFF \
-D Trilinos_ENABLE_Fortran:BOOL=OFF \
\
-D Trilinos_ENABLE_Stokhos:BOOL=ON \
-D Trilinos_ENABLE_Nox:BOOL=ON \
-D Trilinos_ENABLE_ROL:BOOL=ON \
-D ROL_ENABLE_EXAMPLES:BOOL=OFF \
-D Trilinos_ENABLE_MiniTensor:BOOL=OFF \
-D ROL_ENABLE_MiniTensor:BOOL=OFF \
-D Trilinos_ENABLE_Panzer:BOOL=ON \
-D Trilinos_ENABLE_PanzerAdaptersSTK:BOOL=OFF \
-D Trilinos_ENABLE_PanzerAdaptersIOSS:BOOL=OFF \
-D Trilinos_ENABLE_Thyra:BOOL=ON \
-D Trilinos_ENABLE_MueLu:BOOL=ON \
-D Trilinos_ENABLE_Anasazi:BOOL=ON \
-D Trilinos_ENABLE_TrilinosCouplings:BOOL=ON \
-D Trilinos_ENABLE_STKIO:BOOL=OFF \
-D Trilinos_ENABLE_STKUtil:BOOL=OFF \
\
-D Teuchos_ENABLE_STACKTRACE:BOOL=OFF \
-D Teuchos_ENABLE_LONG_LONG_INT:BOOL=ON \
..
```
Some tests fail, but these same tests fail with the Trilinos develop branch:
```
98% tests passed, 13 tests failed out of 617

The following tests FAILED:
        167 - Belos_CustomSolverFactory_MPI_4 (Failed)
        257 - MueLu_MasterList_MPI_1 (Failed)
        283 - MueLu_UnitTestsBlockedTpetra_MPI_1 (Failed)
        308 - MueLu_CreateOperatorEpetra_MPI_1 (Failed)
        309 - MueLu_CreateOperatorEpetra_MPI_4 (Failed)
        314 - MueLu_CreateOperatorTpetra_MPI_1 (Failed)
        315 - MueLu_CreateOperatorTpetra_MPI_4 (Failed)
        404 - Stokhos_SacadoTraitsUnitTest_MPI_1 (Failed)
        406 - Stokhos_SacadoPCEUnitTest_MPI_1 (Failed)
        407 - Stokhos_SacadoETPCEUnitTest_MPI_1 (Failed)
        408 - Stokhos_SacadoPCESerializationTests_MPI_1 (Failed)
        409 - Stokhos_SacadoPCECommTests_MPI_1 (Failed)
        445 - Stokhos_sacado_example_MPI_1 (Failed)
```
Also tested with Tpetra_ENABLE_DEPRECATED_CODE=OFF.
Again, the tests above fail, along with
```
        298 - MueLu_DriverEpetra_IsorropiaPoisson_MPI_4 (Timeout)
        299 - MueLu_DriverEpetra_CircNspDependency_MPI_4 (Timeout)
```
The error message for these two failures is
```
Laplace2D: Build (MueLu::RebalanceTransferFactory)
MueLu_Driver: preconditioner setup crashed w/ message:/Users/kddevin/code/Trilinos/packages/muelu/src/Interface/../MueCentral/MueLu_Level.hpp:196:

Throw number = 3

Throw test that evaluated to true: !IsKey(fac, ename)

"P" not found
MueLu_Driver: Not solving system due to crash in preconditioner setup
```
@trilinos/muelu : do you suspect a problem with the deprecations here?  Do I need to address this problem, or is it not a concern?


packages/ifpack2/src/Ifpack2_Details_OverlappingRowGraph_decl.hpp
```diff
@@ -118,8 +118,10 @@ class OverlappingRowGraph :
   //! The communicator over which the graph is distributed.
   virtual Teuchos::RCP<const Teuchos::Comm<int> > getComm () const;
 
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
   //! The graph's Node instance.
   virtual Teuchos::RCP<node_type> getNode () const;
```
mhoemmen: We should probably mark the function with `TPETRA_DEPRECATED` as well, though this doesn't matter as much for virtual methods where the base class' method has the marking.  Ditto for other instances in this PR.  I mention this in part because you do mark some functions with `TPETRA_DEPRECATED`, like `getOrCreateLocallyReplicatedTpetraVectorSpace` below.

packages/ifpack2/src/Ifpack2_Details_OverlappingRowGraph_def.hpp
```diff
@@ -88,12 +88,14 @@ OverlappingRowGraph<GraphType>::getComm () const
 }
   
 
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
 template<class GraphType>
 Teuchos::RCP<typename GraphType::node_type> 
 OverlappingRowGraph<GraphType>::getNode () const
 {
   return nonoverlappingGraph_->getNode();
```
mhoemmen: This calls a deprecated method, so it will emit a redundant deprecated warning.  Does any caller depend on the returned `RCP<Node>` being nonnull?  My guess is no, so we could just return `Teuchos::null` here and avoid the extra deprecated warning.  Ditto for other calls to `getNode` in this PR.

packages/ifpack2/src/Ifpack2_RILUK_decl.hpp
```diff
@@ -640,10 +640,9 @@ clone (const Teuchos::RCP<const NewMatrixType>& A_newnode) const
   RCP<ParameterList> plClone = Teuchos::parameterList ();
```
mhoemmen: This code occurs in the `clone` method, which we should actually deprecate as well.

packages/ifpack2/src/Ifpack2_RILUK_decl.hpp
```diff
@@ -640,10 +640,9 @@ clone (const Teuchos::RCP<const NewMatrixType>& A_newnode) const
...
   plClone = detail::setLocalSolveParams<NewMatrixType, new_node_type>::setParams (plClone);
 
-  RCP<new_node_type> new_node = A_newnode->getNode ();
-  new_riluk->L_ = L_->clone (new_node, plClone);
-  new_riluk->U_ = U_->clone (new_node, plClone);
-  new_riluk->D_ = D_->clone (new_node);
+  new_riluk->L_ = rcp(new crs_matrix_type(L_, Teuchos::Copy));
```
mhoemmen: The deep copy constructor `CrsMatrix(const CrsMatrix&, Teuchos::Copy)` currently only works if the input CrsMatrix is fill complete.  It would be easy to fix that, but please do be aware of that for now.  Thanks!

packages/thyra/adapters/tpetra/src/Thyra_TpetraThyraWrappers_def.hpp
```diff
@@ -152,7 +167,9 @@ Thyra::createMultiVector(
     getOrCreateTpetraVectorSpace(rangeSpace, tpetraMultiVector_in->getMap()),
     getOrCreateLocallyReplicatedTpetraVectorSpace<Scalar, LocalOrdinal, GlobalOrdinal, Node>(
       domainSpace, tpetraMultiVector_in->getMap()->getComm(),
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
       tpetraMultiVector_in->getMap()->getNode(),
```
mhoemmen: I don't think you need to call the deprecated overload any more, even if deprecated code is enabled.  Ditto for other instances below.

packages/tpetra/core/test/Tpetra_TestingUtilities.hpp
```diff
@@ -117,15 +117,18 @@ namespace Tpetra {
...
 
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
     /// \brief Get the default Kokkos(Classic) Node for Tpetra tests.
     /// \tparam Node The Kokkos(Classic) Node type.
     ///
     /// \warning This function is an implementation detail of Tpetra.
     ///   Users must not call this function or rely on its behavior.
     template <class Node>
+    TPETRA_DEPRECATED
     Teuchos::RCP<Node> getNode () {
```
mhoemmen: This function is an implementation detail of Tpetra's tests, so I don't think we need to deprecate it (though it doesn't hurt).  See also the notes above about calling deprecated methods (like Node's constructor, which should be deprecated if it's not already).

mhoemmen:
@trilinos/tpetra Regarding `delete`-ing Node's constructor: That should perhaps be the last thing we do, since it would depend on all these other changes.  We could deprecate it first, then `delete` it once doing so gets through the build.

mhoemmen:
I also realized that KokkosDeviceWrapperNode lives in Teuchos (`Trilinos/packages/teuchos/kokkoscompat/src`), so we can't use the Tpetra deprecation macros there.  We could use the Kokkos or Teuchos deprecation macros, or we could just let use of Node downstream in Tpetra be deprecated.  The latter might end up being more useful and easier (since Xpetra is full of Node stuff).

packages/ifpack2/src/Ifpack2_Details_OverlappingRowGraph_decl.hpp
kddevin: Done

packages/ifpack2/src/Ifpack2_Details_OverlappingRowGraph_def.hpp
kddevin: Done

packages/ifpack2/src/Ifpack2_RILUK_decl.hpp
kddevin: Deprecating Ifpack2's clone methods (which clone preconditioners) is beyond the scope of this PR.  It is more relevant to #4900.

packages/ifpack2/src/Ifpack2_RILUK_decl.hpp
kddevin: Noted; I think it is fine here.

packages/thyra/adapters/tpetra/src/Thyra_TpetraThyraWrappers_def.hpp
kddevin: Calling the deprecated overload is the only way to test that the overload is correct (or at least compiles).  We learned this the hard way with the xpetra deprecations.

packages/tpetra/core/test/Tpetra_TestingUtilities.hpp
kddevin: Now returns Teuchos::null
I'll go with "though it doesn't hurt" for the rest.

kddevin:
We certainly need to keep KokkosDeviceWrapperNode until the Node template argument is removed.  But we could, perhaps, deprecate the constructor and all of the class' other methods.  I will try it.

mhoemmen:
@kddevin If you need me to approve this, just let me know.

kddevin:
@mhoemmen I made the code changes as recommended.
My tests are still building, but you can look at the code.  Thanks.

packages/teuchos/kokkoscompat/src/KokkosCompat_ClassicNodeAPI_Wrapper.hpp
```diff
@@ -43,21 +44,28 @@ class KokkosDeviceWrapperNode {
   /// release.  This Node type is safe to use.
   static constexpr bool classic = false;
 
+#if defined(TPETRA_ENABLE_DEPRECATED_CODE) || !defined(TEUCHOS_HIDE_DEPRECATED_CODE)
```
mhoemmen: We shouldn't actually use the Tpetra macro here, since Teuchos does not depend on Tpetra.  However, the macro is likely harmless here.

kddevin: I agree, but if someone has TEUCHOS_HIDE_DEPRECATED_CODE=ON and TPETRA_ENABLE_DEPRECATED_CODE=ON, he'll get a bunch of compiler errors unless we test for TPETRA_ENABLE_DEPRECATED_CODE here.

mhoemmen: should be ok then

PR #5026, opened 2019-04-29
NOX: use similarity transform Jacobian when computing row sums

rppawlo:


packages/nox/src-thyra/NOX_PrePostOperator_RowSumScaling.C
```diff
@@ -104,9 +114,20 @@ computeScaling(const NOX::Solver::Generic& solver)
...
     thyra_group->getScaledJacobianOperator();
 
+  // If similarity transform is enabled, left scale matrix by
+  // D_{x}^{-1} before computing row sum to preserve spectral radius
+  // bound: row sum will then be of the system: D_{x}^{-1}JD_{x}
+  if (use_similarity_transform_jacobian_) {
+    const Teuchos::RCP< const ::Thyra::ScaledLinearOpBase<double> > scaled_jac =
+      Teuchos::rcp_dynamic_cast< const ::Thyra::ScaledLinearOpBase<double> >(jac, true);
+    auto& nonconst_scaled_jac = const_cast< ::Thyra::ScaledLinearOpBase<double>&>(*scaled_jac); 
+    nonconst_scaled_jac.scaleLeft(*(thyra_group->getInvRightWeightVector()));
```
mhoemmen: @bartlettroscoe @rppawlo @egphill Does Thyra actually use Tpetra's new-ish threaded functions for computing row sums and left scaling?

rppawlo: @mhoemmen - I just took a look - no, it does not. We will have to update the thyra-tpetra adapters. Thanks for mentioning this.

mhoemmen: Thanks @rppawlo !  Check out the equilibration example in `ifpack2/example`.

bartlettroscoe: Need to implement Kokkos-aware RTOps or this will turn into an interface/adapter mess.

rppawlo:
> This doesn't seem to enable the similarity transform without also doing a row sum scaling. Is that an option you would want to enable?

@egphill - good point. I think we always use row sum in Drekar so for our use case it is fine. Would you want this without row sum for fluids?

> Also, does the linear system end up being any different if you do the similarity transform plus row sum scaling versus just doing row sum scaling?

It is different.

etphipp:
@rppawlo, do you want me to go ahead and approve this?

rppawlo:
@etphipp - no - please disregard this PR. After talking with @egphill and John S. last night, we are going to hold off on committing until we work some things out. Thanks for checking!

william76:
@trilinos/nox 

Issue #5015, opened 2019-04-25
KokkosKernels promotion needs to update Ifpack2


jjwilke:
`KokkosKernels::Experimental` has apparently disappeared in the devel branch of KokkosKernels. Ifpack2 references `namespace Experimental` in several places.  This will make the next promotion difficult - particularly the new CMake updates for kokkos-kernels.

srajama1:
@kyungjoo-kim is aware of this. We will fix it during promotion.

ndellingwood:
There is the kokkos-promotion branch of Trilinos tracking changes needed in Trilinos during the next promotion with a couple fixes. @kyungjoo-kim or I can make the fixes on that branch.

ndellingwood:
@jjwilke is this blocking your cmake testing? If so we can get the fixes in quickly, I'm assuming this is the removal the batched blas stuff from the experimental namespace.

ndellingwood:
@trilinos/kokkos-kernels

kyungjoo-kim:
@jjwilke Could you check your cmake work after the PR #5032 ? If it works, would you please run your performance test of sparc to check that I did not create any side effects ?

ndellingwood:
This was resolved by #5032, closing.

PR #5032, opened 2019-04-29
Ifpack2 kokkos promotion preparation

kyungjoo-kim:
@trilinos/ifpack2 

## Description

kokkoskernels removes experimental namespace. this patch includes appropriate fixes. it also includes the latest ifpack2 improvement. one possible thing that could happen in sparc is that norm interface needs to be modified. @jjwilke (github does not complete your id automatically). would you change sparc for that ? another major change would be separating tpetra memory space from the internal memory space. The internal memory space is either host space or cuda space. Tpetra memoryspace is right now assumed to be uvm.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -103,7 +103,7 @@
 //#define IFPACK2_BLOCKTRIDICONTAINER_USE_CUDA_MEMORY_FOR_MPI
 
 // if defined, all views are allocated on cuda space intead of cuda uvm space
-// #define IFPACK2_BLOCKTRIDICONTAINER_USE_CUDA_SPACE
+#define IFPACK2_BLOCKTRIDICONTAINER_USE_CUDA_SPACE
 
 namespace Ifpack2 {
 
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -2088,19 +2088,6 @@ namespace KB = KokkosBatched::Experimental;
 	const local_ordinal_type per_team_scratch = internal_vector_scratch_type_3d_view::
 	  shmem_size(blocksize, blocksize, vector_loop_size);
 
-#if defined(KOKKOS_ENABLE_CUDA)
-        // when tpetra memory space is uvm, prefetch them to device or host accordingly
-        if (std::is_same<typename impl_type::node_memory_space,Kokkos::CudaUVMSpace>::value) {
-          if (std::is_same<execution_space,Kokkos::Cuda>::value) {
-            CUDA_SAFE_CALL(cudaMemPrefetchAsync(A_rowptr.data(), A_rowptr.span()*sizeof(a_rowptr_value_type), 0));
-            CUDA_SAFE_CALL(cudaMemPrefetchAsync(A_values.data(), A_values.span()*sizeof(impl_scalar_type   ), 0));
-          } else {
-            CUDA_SAFE_CALL(cudaMemPrefetchAsync(A_rowptr.data(), A_rowptr.span()*sizeof(a_rowptr_value_type), cudaCpuDeviceId));
-            CUDA_SAFE_CALL(cudaMemPrefetchAsync(A_values.data(), A_values.span()*sizeof(impl_scalar_type   ), cudaCpuDeviceId));
-          }
-        }
-#endif
-
 	Kokkos::TeamPolicy<execution_space,ExtractAndFactorizeTag>
 	  policy(packptr.extent(0)-1, team_size, vector_loop_size); 
 #if defined(KOKKOS_ENABLE_DEPRECATED_CODE)
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -94,8 +94,13 @@
 // this requires kokkos develop branch (do not use this yet)
 //#define IFPACK2_BLOCKTRIDICONTAINER_USE_CUDA_STREAM
 
-// if defined, this use pinned memory instead of device pointer, cuda aware mpi
+// ::: Experiments :::
+// define either pinned memory or cudamemory for mpi
+// if both macros are disabled, it will use tpetra memory space which is uvm space for cuda 
+// if defined, this use pinned memory instead of device pointer
+// by default, we enable pinned memory
 #define IFPACK2_BLOCKTRIDICONTAINER_USE_PINNED_MEMORY_FOR_MPI
+//#define IFPACK2_BLOCKTRIDICONTAINER_USE_CUDA_MEMORY_FOR_MPI
 
 // if defined, all views are allocated on cuda space intead of cuda uvm space
 // #define IFPACK2_BLOCKTRIDICONTAINER_USE_CUDA_SPACE
@@ -484,10 +489,16 @@ namespace KB = KokkosBatched::Experimental;
       using size_type_1d_view = typename impl_type::size_type_1d_view;
       using size_type_1d_view_host = Kokkos::View<size_type*,Kokkos::HostSpace>;
 
-#if defined(KOKKOS_ENABLE_CUDA) && defined(IFPACK2_BLOCKTRIDICONTAINER_USE_PINNED_MEMORY_FOR_MPI)
+#if defined(KOKKOS_ENABLE_CUDA) 
       using impl_scalar_type_1d_view = 
         typename std::conditional<std::is_same<execution_space,Kokkos::Cuda>::value,
+#  if defined(IFPACK2_BLOCKTRIDICONTAINER_USE_PINNED_MEMORY_FOR_MPI)
                                   Kokkos::View<impl_scalar_type*,Kokkos::CudaHostPinnedSpace>,
+#  elif defined(IFPACK2_BLOCKTRIDICONTAINER_USE_CUDA_MEMORY_FOR_MPI)
+                                  Kokkos::View<impl_scalar_type*,Kokkos::CudaSpace>,
+#  else                           // no experimental macros are defined
+                                  typename impl_type::impl_scalar_type_1d_view,
+#  endif                          /// when cuda enabled and exec space is not cuda
                                   typename impl_type::impl_scalar_type_1d_view>::type;
 #else
       using impl_scalar_type_1d_view = typename impl_type::impl_scalar_type_1d_view;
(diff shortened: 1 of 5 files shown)


kyungjoo-kim:
#5015

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -90,10 +90,26 @@
...
+#define IFPACK2_BLOCKTRIDICONTAINER_USE_PINNED_MEMORY_FOR_MPI
+
+// if defined, all views are allocated on cuda space intead of cuda uvm space
+// #define IFPACK2_BLOCKTRIDICONTAINER_USE_CUDA_SPACE
+
 namespace Ifpack2 {
 
   namespace BlockTriDiContainerDetails {
 
+    /// for the next promotion
```
mhoemmen: I do wish kokkos-kernels had promotion number macros :) .  Nothing we can do about that now, though.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -364,32 +397,58 @@ namespace Ifpack2 {
       /// use scalar_type for communication data type enum
       using scalar_type = typename impl_type::scalar_type;
 
-      template <typename T>
-      static int isend(const MPI_Comm comm, const T* buf, int count, int dest, int tag, MPI_Request* ireq) {
+      static int isend(const MPI_Comm comm, const char* buf, int count, int dest, int tag, MPI_Request* ireq) {
```
mhoemmen: For the future: Do we actually need this?  It looks like you just call `MPI_Isend` in some places, and in others, you call this `isend` function.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -364,32 +397,58 @@ namespace Ifpack2 {
...
         if (ireq == NULL) MPI_Request_free(&ureq);
         return ret;
 #else
         return 0;
 #endif
       }
 
-      template <typename T>
-      static int irecv(const MPI_Comm comm, T* buf, int count, int src, int tag, MPI_Request* ireq) {
+      static int irecv(const MPI_Comm comm, char* buf, int count, int src, int tag, MPI_Request* ireq) {
```
mhoemmen: See above note on `isend`.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -527,16 +607,246 @@ namespace Ifpack2 {
...
+                       const Teuchos::RCP<const tpetra_map_type>& tgt_map, 
+                       const local_ordinal_type blocksize_,
+                       const local_ordinal_type_1d_view dm2cm_) {
+        blocksize = blocksize_;
+        dm2cm = dm2cm_;
+        
+#ifdef HAVE_IFPACK2_MPI
+        const auto mpi_comm = Teuchos::rcp_dynamic_cast<const Teuchos::MpiComm<int> >(tgt_map->getComm());
+        TEUCHOS_ASSERT(!mpi_comm.is_null());
+        comm = *mpi_comm->getRawMpiComm();
```
mhoemmen: For future reference: We have a function already for extracting the raw `MPI_Comm`: https://github.com/trilinos/Trilinos/blob/master/packages/tpetra/core/src/Tpetra_Details_extractMpiCommFromTeuchos.hpp

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -527,16 +607,246 @@ namespace Ifpack2 {
...
+          // data_at_i.pid = pids.send[i];
+          // data_at_i.tag = 42;
+          // data_at_i.ireq = &reqs.send[i];
+          // data_at_i.r_val = 0;
+          // CUDA_SAFE_CALL(cudaStreamAddCallback(stream_at_i, callback_isend, (void*)&data_at_i, 0));
+        }
+
+        Kokkos::fence();
+#if defined (KOKKOS_ENABLE_OPENMP)
+#pragma omp parallel for
```
mhoemmen: This parallel loop requires `MPI_THREAD_MULTIPLE`.  I worry that we don't have adequate testing for the case where MPI does not provide `MPI_THREAD_MULTIPLE`.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -527,16 +607,246 @@ namespace Ifpack2 {
...
+        }
+#endif // HAVE_IFPACK2_MPI
+      }
+
+      void syncRecvViaCudaStream() {
+	IFPACK2_BLOCKTRIDICONTAINER_TIMER("BlockTriDi::AsyncableImport::SyncRecv");
+#ifdef HAVE_IFPACK2_MPI
+        // 0. wait for receive async.
+#if defined (KOKKOS_ENABLE_OPENMP)
+#pragma omp parallel for
```
mhoemmen: This parallel loop requires `MPI_THREAD_MULTIPLE`.  I worry that we don't have adequate testing for the case where MPI does not provide `MPI_THREAD_MULTIPLE`.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1144,15 +1398,16 @@ namespace Ifpack2 {
             Kokkos::parallel_for(Kokkos::ThreadVectorRange(member, vector_loop_size),[&](const int &v) {
                 Kokkos::parallel_for(Kokkos::TeamThreadRange(member,icount),[&](const local_ordinal_type &ii) {
                     const local_ordinal_type i = ibeg + ii*3;
-                    for (local_ordinal_type j=0;j<blocksize;++j)
-                      values(i,j,j,v) = 1;
+                    for (local_ordinal_type j0=0;j0<blocksize;++j0) 
+                      for (local_ordinal_type j1=0;j1<blocksize;++j1) 
+                        values(i,j0,j1,v) = (j0 == j1 ? 1 : 0);
```
mhoemmen: This will likely fail to compile in the complex arithmetic case, since there's no assignment defined from integers to `complex<double>`.  The easiest fix is just to use `1.0` and `0.0` instead of 1 and 0.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3299,28 +3570,33 @@ namespace Ifpack2 {
           comm_ = *mpi_comm->getRawMpiComm();
 #endif
         }
+        work_[0] = 0; 
+        work_[1] = 0; 
+        work_[2] = -1;
```
mhoemmen: If `work_` stores `impl_scalar_type`, then these assignments may not compile for the complex arithmetic case.  See note above.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: I just put the temporary macro in kokkoskernels when I need to include changes related to the promotion. After promotion is done, I remove the macro from kokkoskernels and also remove it from ifpack2. It is not systematic but works ; )

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: I don't think that I use send/recv other places. Communication happens in AsyncableImporter (Send/Recv) and NormManager (AllReduce).

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: No worries :)

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: It's OK, I'm not so worried about this.  It could lead to a multiple definition error if ETI is OFF, though.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: done.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: This is just a pipelining experiment using cuda streams. I was working on a few things and cannot decouple each other but I put a macro so that the code path is not visible at this moment. So, we don't need to worry about this. In this PR, I do not use cuda stream (but some implementation is included with ifdef guards) nor use cuda space (but some implementation is already made to distinct tpetra memory space from internal memory space).

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: This so far have been passed for complex tests but will change it using one and zero. done.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: work_ is magnitude type. It is anyway fixed as it is more clear and prevent future warnings.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: @kyungjoo-kim cool, just wanted to make sure -- thanks!

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: Resolved -- see above.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: Thanks!

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: Thanks!

kyungjoo-kim:
something went wrong.

kyungjoo-kim:
The error message is ‘’invalid device ordinal’’. @crtrott Any idea ?

mhoemmen:
@kyungjoo-kim wrote:
> The error message is ‘’invalid device ordinal’’. @crtrott Any idea ?

The internet suggests that `cudaSetDevice` was given an out-of-range GPU index.  For example, if one only has 1 GPU, `cudaSetDevice(0)` is valid but `cudaSetDevice(1)` will fail.

kyungjoo-kim:
Yes. That is what I read and I cannot believe that this kind of error indeed possibly happens. If that is the root of the error, then other nightly tests would also fail.

mhoemmen:
> Yes. That is what I read and I cannot believe that this kind of error indeed possibly happens. If that is the root of the error, then other nightly tests would also fail.

True.  It could be that some other error wasn't caught and cleared, so the next call to `cudaSetDevice` looks like it failed.

kyungjoo-kim:
The problem was that I use cuda prefetch function for uvm locality as an experiment. It works for P100 and V100 and I did not check it for K80 which is used for PR test. Anyway it is not really necessary and I removed it. This time, it will pass the pr test.

PR #5048, opened 2019-04-30
Ifpack2: Add update damping to AdditiveSchwarz

csiefer2:
Auto-PR for SHA 7a4b9a5

### DIFF
--- a/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp
@@ -233,6 +233,7 @@ AdditiveSchwarz (const Teuchos::RCP<const row_matrix_type>& A) :
   FilterSingletons_ (false),
   NumIterations_(1),
   ZeroStartingSolution_(true),
+  UpdateDamping_(Teuchos::ScalarTraits<scalar_type>::one()),
   NumInitialize_ (0),
   NumCompute_ (0),
   NumApply_ (0),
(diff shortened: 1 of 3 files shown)


packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp
```diff
@@ -966,6 +968,9 @@ setParameterList (const Teuchos::RCP<Teuchos::ParameterList>& plist)
   // singletons should help for PDE problems with Dirichlet BCs.
   FilterSingletons_ = plist->get ("schwarz: filter singletons", FilterSingletons_);
 
+  // Allow for damped Schwarz updates
+  UpdateDamping_ = Teuchos::as<scalar_type>(plist->get("schwarz: update damping",UpdateDamping_));
```
mhoemmen: Would it ever make sense for the damping parameter to have a nonzero imaginary part?  If not, then it would be nicer if we made this `magnitude_type` with a fall-back to `double`.

packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp
```diff
@@ -1041,6 +1046,7 @@ getValidParameters () const
     const bool filterSingletons     = false;
     const int  numIterations        = 1;
     const bool zeroStartingSolution = true;
+    const double updateDamping      = 1;
```
mhoemmen: You use `scalar_type` as the parameter's type above, but use `double` here.  This could cause failures in two cases:

  1. `scalar_type` is `complex<T>`
  2. `scalar_type` is `float`

csiefer2: Yeah, that's a MueLu-style intentional inconsistency.

Issue #4986, opened 2019-04-22
Xpetra: implement ETI system


jhux2:
In order to decrease downstream package's library sizes, such as MueLu's, Xpetra needs to allow for ETI (explicit template instantiation).

@trilinos/xpetra @cgcgcg @lucbv 

Related to #4984, #3137.

lucbv:
@jhux2 maybe @cgcgcg and I should be added to the xpetra team to get these notifications?

jhux2:
@lucbv Ask and you shall receive :).

csiefer2:
Yay! Team!

jhux2:
See PR #4996.

jhux2:
Latest count of weak symbols in libmuelu.a and libmuelu_lgn.a combined

```
   8469 Xpetra::BlockedVector
   7911 Xpetra::BlockedMultiVector
   7271 Xpetra::TpetraMultiVector
   5580 Xpetra::TpetraMap
   5460 Xpetra::BlockedMap
   5418 Xpetra::EpetraMapT
   5100 Xpetra::EpetraCrsMatrixT
   4585 Xpetra::CrsMatrixWrap
   2839 Xpetra::TpetraVector
   2797 Xpetra::EpetraMultiVectorT
   2398 Xpetra::TpetraImport
   2378 Xpetra::StridedMap
   2368 Xpetra::EpetraImportT
   1521 Xpetra::BlockedCrsMatrix
   1048 Xpetra::Matrix
```

Command to get the count:

```
nm --demangle libmuelu.a libmuelu_lgn.a | grep " [W] Xpetra::" > weaksymbols.txt
cat weaksymbols.txt | cut -f1 -d'<' | sort | cut -f3 -d' ' | uniq -c | sort -nr | head -n 15
```

william76:
Chris and I chatted abut this today. I'll take a crack at `Tpetra::Map` and will document the steps of what I did...

csiefer2:
@william76 You update the checklist above?

william76:
> @william76 You update the checklist above?

@csiefer2 yup.  TpetraBlockCrsMatrix is merged in so I checked it off.  TpetraExport is the next on my list.

william76:
With #5579 merged in I re-ran the weak symbol count that Jonathan ran earlier.  Here are the new numbers I'm getting:

```
   2636 Xpetra::EpetraMapT
   2208 Xpetra::StridedMap
[...]

PR #5052, opened 2019-05-01
Xpetra: ETI for CrsMatrixWrap (try 2) 

jhux2:
Partially address issue #4986.


packages/xpetra/sup/CMakeLists.txt
```diff
@@ -12,12 +12,63 @@ TRIBITS_CONFIGURE_FILE(${PACKAGE_NAME}_config.hpp)
...
+  SET(SOURCES ${XPETRA_ETI_CPP_SOURCES})
+  # grab any hand rolled ETI source files
+  APPEND_GLOB(SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/Utils/ExplicitInstantiation/*.cpp)
+  
+ENDIF() # if XPETRA_ENABLE_EXPLICIT_INSTANTIATION
+
 SET(HEADERS "")
 APPEND_GLOB(HEADERS */*.hpp)
-APPEND_GLOB(HEADERS */*.cpp)
+APPEND_GLOB(HEADERS */*.cpp) #JHU 25-Apr-2019 FIXME is this correct?
```
mhoemmen: Wouldn't you want `.cpp` files to go into `SOURCES`, so they can go into the list of files to compile into the library (see `TRIBITS_ADD_LIBRARY` invocation below)?

packages/xpetra/sup/CMakeLists.txt
```diff
@@ -12,12 +12,63 @@ TRIBITS_CONFIGURE_FILE(${PACKAGE_NAME}_config.hpp)
...
 
+# FIXME JHU is this still needed?
 SET(SOURCE mac.cpp) # Avoid empty library (MacOs 'ar' utility doesn't work on empty library).
```
mhoemmen: It shouldn't even be right -- I don't see where `SOURCE` gets used.

jjellio: `SET(SOURCE mac.cpp)` => `SET(SOURCES mac.cpp)`

jjellio:
Let's say you disable ETI.   What CMake stuff takes in the 'decl' and generates Foo.hpp (with include decl; include def as its contents)?  I don't understand how Trilinos handles the ETI 'off' case.  It seems in all cases, `Foo.hpp` should always be auto generated if there is a `Foo_decl.hpp`

mhoemmen:
@jjellio wrote:
> Let's say you disable ETI. What CMake stuff takes in the 'decl' and generates Foo.hpp (with include decl; include def as its contents)?

https://github.com/trilinos/Trilinos/blob/master/cmake/tribits/core/package_arch/TribitsCreateClientTemplateHeaders.cmake

jhux2:
@trilinos/framework @ZUUL42 @bartlettroscoe PR testing uses "ninja -w dupbuild=err".  ninja was issuing a warning along the lines of
```
ninja: warning: multiple rules generate /scratch/jhu/trilinos/Trilinos/packages/xpetra/src/Utils/ExplicitInstantiation/Xpetra_ETI_SC_LO_GO_NO.tmpl.  builds involving this target will not be correct; continuing anyway [-w dupbuild=warn]
```
but this isn't captured in the output logs.

It seems that the PR tester should stop immediately on such a ninja error and print it to a log.  Of course, I don't know how hard implementing this might be.

Catching this would have been really difficult without @william76's jsim tool!  Thanks, @william76!

bartlettroscoe:
> "ninja -w dupbuild=err"

Never seen that option.  Why out PR testing set that?

PR #5062, opened 2019-05-01
Panzer:  Create ZeroContributedField Evaluator

jmgate:
@trilinos/panzer 

## Description
Create an `Evaluator` for the sake of initializing a contributed field to zero.

## Motivation and Context
This will make it such that we can switch more `Evaluator`s over to `CONTRIBUTES`-style evaluation.  This is currently needed by work underway in Drekar, but it will also benefit Charon and perhaps others.

### DIFF
--- a/packages/panzer/disc-fe/src/evaluators/Panzer_ZeroContributedField.cpp
+++ b/packages/panzer/disc-fe/src/evaluators/Panzer_ZeroContributedField.cpp
@@ -1,0 +1,60 @@
+// @HEADER
+// ***********************************************************************
+//
+//           Panzer: A partial differential equation assembly
+//       engine for strongly coupled complex multiphysics systems
+//                 Copyright (2011) Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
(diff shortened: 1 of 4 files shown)


jmgate:
@rppawlo, is this the kind of thing we need?

jmgate:
Added a test in `disc-fe/test/core_tests/evaluators.cpp`.  Is that the right spot?  Looks like most of those just test the construction of the `Evalautor`.  Should there be more to it than that?

rppawlo:
Evaluator unit test should also check values. Those are some older tests that are not very good. Was thinking more along the lines of whats in Trilinos/packages/panzer/disc-fe/test/evaluator_tests. The painful part is building the workset. An empty one might work for your case. It also might be easier to just find an equation set or test that uses a sum evaluator and replace it with zero out plus contributed?

jmgate:
How's this look?  I replaced an `EVALUATES`-style one with a `ZeroContributedField` and a `CONTRIBUTES`-style one.  Seems like it does the trick.

jmgate:
Perhaps a conversation over Skype sometime, but I'd like to brainstorm the ideal way to unit test an `Evaluator` in general without jumping through all the hoops of mesh, workset, dependent fields, etc.  I'm thinking something along the lines of having everything within `evaluateFields()` broken up into private functions that you could then unit test through a `friend` interface or something along those lines.  Charon's going to begin making a push to increase unit testing everywhere, and we were talking about whipping up a generic unit test template for an `Evaluator`&mdash;just not sure what exactly that should look like.

jmgate:
The CUDA build failed due to a missing sems-env.  Retesting.

jmgate:
@trilinos/framework, the @trilinos-autotester could *really* use a re-design.  This was approved two hours ago, but we can't merge it yet because the autotester hasn't come back around to look at this PR again.

Issue #4330, opened 2019-02-06
Build failures in SPARC 'master' + Trilinos 'develop' starting 2/5/2019 due to break in Teuchos::SerialDenseMatrix


bartlettroscoe:
CC: @trilinos/teuchos , @jwillenbring (Trilinos Framework Product Lead), @bartlettroscoe, @fryeguy52, @sebrowne, @micahahoward

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this SPARC CDash query for 2019-02-05](http://compsim-dashboard.sandia.gov/cdash/index.php?project=SPARC&date=2019-02-05&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=-trildev), the version of Trilinos 'develop' for testing day 2019-02-05 has broken the new SPARC 'master' + Trilinos 'develop' builds (see [TRIL-243](https://sems-atlassian-son.sandia.gov/jira/browse/TRIL-243)) which detect defects in Trilinos 'develop' which break SPARC.  These build failures did not exist in those builds the previous day as shown in [this SPARC CDash query for 2019-02-04](http://compsim-dashboard.sandia.gov/cdash/index.php?project=SPARC&date=2019-02-04&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=-trildev).

The 'clang-5.0.1' build shows just one build failure [here](http://compsim-dashboard.sandia.gov/cdash/viewBuildError.php?buildid=92804) which shows:

```
sparc-ear99/src/solver-analysis/ROLInterface.C:313:14: error: invalid operands to binary expression ('std::ofstream' (aka 'basic_ofstream') and 'Teuchos::SerialDenseMatrix' (aka 'SerialDenseMatrix'))
      cov_out << cov;
     ~~~~~~~ ^  ~~~
```

That SPARC 'master' build was built against Trilinos 'develop' for testing day 2019-02-05 for the build `Trilinos-atdm-cee-rhel6_clang-5.0.1_openmpi-1.10.2_serial_static_opt` shown [here](https://testing.sandia.gov/cdash-dev-view/index.php?project=Trilinos&parentid=4494553). 

## Current Status on CDash

The current status of the SPARC 'master' + Trilinos 'develop' builds for the current testing day can be see [here|http://compsim-dashboard.sandia.gov/cdash/index.php?project=SPARC&date=2019-02-05&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=-trildev] (internal link).

[...]

PR #5064, opened 2019-05-01
Teuchos:   added deprecation warnings that were missing in #ifdef'ed deprecated code

kddevin:
@trilinos/teuchos 

## Description
Methods in Teuchos were #ifdef'ed as deprecated, but did not include TEUCHOS_DEPRECATED warnings.  Now they do.

This PR will need to be cherry-picked to the 12.6 release branch.

## Related Issues
#4430 #4330 #4258 

## How Has This Been Tested?
Compilation on mac clang.
Warnings appear as expected; no other behavior changes from this PR.

### DIFF
--- a/packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
+++ b/packages/teuchos/numerics/src/Teuchos_SerialBandDenseMatrix.hpp
@@ -1073,6 +1073,7 @@ void SerialBandDenseMatrix<OrdinalType, ScalarType>::copyMat(
 #ifndef TEUCHOS_HIDE_DEPRECATED_CODE
 /// \brief Print the given SerialBandDenseMatrix to the given output stream.
 template<typename OrdinalType, typename ScalarType>
+TEUCHOS_DEPRECATED
 std::ostream& operator<< (std::ostream& os, const Teuchos::SerialBandDenseMatrix<OrdinalType, ScalarType>& obj)
 {
   obj.print (os);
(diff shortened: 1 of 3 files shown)


kddevin:
@rhoope @mseldre

kddevin:
@trilinos/framework  @trilinos/tempus 

In two of the PR test builds, deprecation warnings are flagged as errors in Tempus, even though they are ignored in other packages.

Is there a way to ignore deprecated warnings for Tempus in the PR tests?

bartlettroscoe:
@kddevin, @trilinos/tempus,

Can we refactor Tempus (and intermediate headers in intermediate packages) to remove usage of deprecated code?  We should be able to do that at least in Trilinos, no?

I think when we deprecate something, we need to remove the usage of that in all of Trilinos if possible.

kddevin:
Thanks, @bartlettroscoe .
I am refactoring Trilinos packages to remove the deprecated use cases; that's how I found this problem.
Deprecation warnings shouldn't break PR testing, though.  One might add deprecation warnings at any time.  Other packages must be seeing lots of deprecation warnings, but they still are passing the PR testing.

bartlettroscoe:
@kddevin said:

> Deprecation warnings shouldn't break PR testing, though. One might add deprecation warnings at any time. Other packages must be seeing lots of deprecation warnings, but they still are passing the PR testing.

The issue is that if non-deprecated functions in header files (i.e. inline or template functions) call deprecated functions, then downstream customers will get deprecated warnings that they can't turn off.  For example, that happened in #4770.  If we are going to issue deprecated warnings that users see in their code, then they need to be able to refactor their code eliminate those deprecated warnings.  But if non-deprecated functions that users have to call is calling deprecated code, they can't make those deprecated warnings go away and therefore we defeat the purpose of deprecated warnings and users will ignore them.

The only way that I know for sure to ensure that users will not see deprecated warnings from header files that they can't make go away is to make sure that Trilinos packages build with no deprecated warnings. (at least in one PR build where deprecated code is hidden and disabled).

I know that we need to still support deprecated code until it is removed so we would need to have one build that turns off deprecate code (i.e. hides it) and then builds clean with `-Werror`. Perhaps we can do that with this Intel 17.0.1 PR build?  The other builds that have `-Werror` could also set `-Wno-error=deprecated-declarations` so deprecated warnings don't trigger errors so we still test deprecated code.

Does this make sense?  We need to work this out with the @trilinos/framework team.

kddevin:
Closing this PR; going to submit to release 12.6 branch.

mhoemmen:
Likely correction: 12.16 branch.

PR #5068, opened 2019-05-01
fix a stk-topology unit-test failure on gpu.

alanw0:
A couple of stk-topology unit-test functions were using new and delete on device.
Surprisingly this failed only on some GPUs, not all... Very difficult to track down,
manifested as memory corruption. Finally found it with cuda-memcheck which
provided a clue which led us to replace the dynamically allocated array with a
stack array.

@trilinos/<teamName>

## Checklist

### DIFF
--- a/packages/stk/stk_unit_tests/stk_topology/topology_test_utils.hpp
+++ b/packages/stk/stk_unit_tests/stk_topology/topology_test_utils.hpp
@@ -326,7 +326,8 @@ template <unsigned MAX_NODES>
 STK_INLINE_FUNCTION
 void check_side_nodes_ngp(stk::topology topology, unsigned gold_side_node_ordinals[][MAX_NODES])
 {
-  stk::mesh::Entity * allElemNodes = new stk::mesh::Entity[topology.num_nodes()];
+  constexpr unsigned MAX_NODES_PER_ELEM = 100;
+  stk::mesh::Entity allElemNodes[MAX_NODES_PER_ELEM];
   for (unsigned nodeOrdinal = 0; nodeOrdinal < topology.num_nodes(); ++nodeOrdinal) {
     allElemNodes[nodeOrdinal] = nodeOrdinal + 100;
   }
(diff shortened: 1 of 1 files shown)


mhoemmen:
@alanw0 Thanks!

packages/stk/stk_unit_tests/stk_topology/topology_test_utils.hpp
```diff
@@ -326,7 +326,8 @@ template <unsigned MAX_NODES>
 STK_INLINE_FUNCTION
 void check_side_nodes_ngp(stk::topology topology, unsigned gold_side_node_ordinals[][MAX_NODES])
 {
-  stk::mesh::Entity * allElemNodes = new stk::mesh::Entity[topology.num_nodes()];
+  constexpr unsigned MAX_NODES_PER_ELEM = 100;
```
bartlettroscoe: How do we know that the number of nodes per element is never over 100?  Is there no element in existence that would ever have over 100 nodes?

alanw0: For the element topologies defined in stk-topology, none have more than 27 nodes. But you make a good point, this is not future-proof. It's probably worth putting a check in place so that when a new developer adds some high-order element in the future, this would trip and let them know. We'll take on a task to make this more robust.

packages/stk/stk_unit_tests/stk_topology/topology_test_utils.hpp
```diff
@@ -322,12 +322,15 @@ inline void check_permutation_nodes(stk::topology topology, std::vector<std::vec
...
 
+constexpr unsigned MAX_NODES_PER_ELEM = 100;
+
 template <unsigned MAX_NODES>
 STK_INLINE_FUNCTION
 void check_side_nodes_ngp(stk::topology topology, unsigned gold_side_node_ordinals[][MAX_NODES])
 {
-  constexpr unsigned MAX_NODES_PER_ELEM = 100;
   stk::mesh::Entity allElemNodes[MAX_NODES_PER_ELEM];
+  NGP_EXPECT_TRUE(topology.num_nodes() < MAX_NODES_PER_ELEM);
```
bartlettroscoe: Excellent, thanks!

packages/stk/stk_unit_tests/stk_topology/topology_test_utils.hpp
```diff
@@ -322,11 +322,15 @@ inline void check_permutation_nodes(stk::topology topology, std::vector<std::vec
   }
 }
 
+constexpr unsigned MAX_NODES_PER_ELEM = 100;
```
bartlettroscoe: Excellent, one definition of this.

packages/stk/stk_unit_tests/stk_topology/topology_test_utils.hpp
bartlettroscoe: Just one suggestion.  You might consider an inlined function that checks this this and calls this in all three places.  Actually it looks like the code:

```
  stk::mesh::Entity allElemNodes[MAX_NODES_PER_ELEM];
  NGP_EXPECT_TRUE(topology.num_nodes() < MAX_NODES_PER_ELEM);
```

is 100% duplicated in these three places.   Would be easy to define a macro to reduce duplication like:

```
#define STK_DEFINE_ELEMENT_NODES_ARRAY(TOPOLOGY, ARRAY_NAME) \
  stk::mesh::Entity ARRAY_NAME[MAX_NODES_PER_ELEM]; \
  NGP_EXPECT_TRUE(TOPOLOGY.num_nodes() < MAX_NODES_PER_ELEM)
```

and then call it like:

```
  STK_DEFINE_ELEMENT_NODES_ARRAY(topology, allElemNodes);
```

Less duplication.  Less boilerplate.

PR #5077, opened 2019-05-02
Panzer: remove LO, GO template parameters from Connection Manager

rppawlo:
Stage 1 of removing multiple global ordinal types in panzer.

PLEASE DO NOT SET THIS TO AUTOMERGE!


packages/panzer/adapters-ioss/test/ioss_connmngr/tIOSSConnManager.cpp
```diff
@@ -354,7 +354,7 @@ TEUCHOS_UNIT_TEST(tIOSSConnManager, basic)
 
       // Check for correct connectivity for all local elements
       if (print) output << "Checking for correct connectivity for all local elements." << std::endl;
-      int * conn;
+      panzer::Ordinal64 * conn;
```
seamill: Should we have a `panzer::LocalOrdinal` and `panzer::GlobalOrdinal` defined, instead of `panzer::Ordinal64`? There is/was a hidden config call that changed Ordinal64 to a regular int, so maybe specializing the name would be helpful to the user. We could probably get away with deprecating the `panzer::Ordinal64` to avoid breaking Drekar/Charon/EMPIRE.

rppawlo: yes - I'm going to rename Ordinal64 to GlobalOrdinal, but not until the last phase of this process. Right now we have too many classes defining template parameters named LocalOrdinal and GlobalOrdinal. Want to keep the unique name until all template instances of GlobalOrdinal are removed and then use sed to rename.

packages/panzer/adapters-stk/src/Panzer_STK_SetupLOWSFactory.cpp
```diff
@@ -105,31 +105,35 @@ namespace panzer_stk {
       reqHandler_local = Teuchos::rcp(new Teko::RequestHandler);
     #endif
 
+    auto stk_conn_manager = Teuchos::rcp_dynamic_cast<panzer_stk::STKConnManager>(conn_manager,true);
+
 #ifndef PANZER_ORDINAL64_IS_INT
```
seamill: Do all of these if statements get removed after we update the DOFManager/GlobalIndexer?

rppawlo: yes!!!

packages/panzer/adapters-stk/test/field_manager_builder/field_manager_builder.cpp
```diff
@@ -143,18 +143,18 @@ namespace panzer {
...
-    const Teuchos::RCP<panzer::ConnManager<int,int> > conn_manager 
-           = Teuchos::rcp(new panzer_stk::STKConnManager<int>(mesh));
+    const Teuchos::RCP<panzer::ConnManager> conn_manager
+           = Teuchos::rcp(new panzer_stk::STKConnManager(mesh));
 
-    Teuchos::RCP<const panzer::UniqueGlobalIndexerFactory<int,int,int,int> > indexerFactory
+    Teuchos::RCP<const panzer::UniqueGlobalIndexerFactory<int,int> > indexerFactory
           = Teuchos::rcp(new panzer::DOFManagerFactory<int,int>);
-    const Teuchos::RCP<panzer::UniqueGlobalIndexer<int,int> > dofManager 
+    const Teuchos::RCP<panzer::UniqueGlobalIndexer<int,int> > dofManager
```
seamill: Is it easy to do a find replace and replace all these dofManager definitions with a 'const auto' to simplify the next update so we don't hit all these files again in the review?

packages/panzer/adapters-stk/test/local_mesh/local_mesh_utilities.cpp
```diff
@@ -81,14 +81,14 @@ TEUCHOS_UNIT_TEST(localMeshUtilities, basic)
   // Make sure if fails when you pass in a null mesh
   {
     panzer_stk::STK_Interface mesh;
-    panzer::LocalMeshInfo<int,int> mesh_info;
+    panzer::LocalMeshInfo<int,panzer::Ordinal64> mesh_info;
```
seamill: Will we be removing this templating at the same time as the DOFManager? I don't think any downstream codes directly access the LocalMeshInfo objects, so it should be pretty clean.

rppawlo: yes - I might do that before starting the DOF manager

packages/panzer/adapters-stk/test/field_manager_builder/field_manager_builder.cpp
rppawlo: Will have to look into this...

packages/panzer/adapters-stk/test/model_evaluator/model_evaluator_blocked_hessians.cpp
```diff
@@ -417,7 +417,7 @@ namespace panzer {
     {
       Teuchos::RCP<panzer::FunctionalResponse_Builder<int,int> > builder
```
seamill: Are these also LO/GO templates?

rppawlo: I suspect they are. Adding phase 4 to the plan...

packages/panzer/disc-fe/src/Panzer_FaceToElement_impl.hpp
```diff
@@ -115,7 +115,7 @@ initialize(panzer::ConnManager<LocalOrdinal,GlobalOrdinal> & conn)
     //const std::vector<GlobalOrdinal> &block_elems = conn.getElementBlock(block_ids[iblk]);
     const std::vector<LocalOrdinal> &block_elems = conn.getElementBlock(block_ids[iblk]);
     for (size_t i=0; i<block_elems.size(); ++i) {
-      const GlobalOrdinal * connectivity = conn.getConnectivity(block_elems[i]);
+      const panzer::Ordinal64 * connectivity = conn.getConnectivity(block_elems[i]);
```
seamill: You've used auto before. Two more places after this as well.

packages/panzer/dof-mgr/src/Panzer_DOFManager_impl.hpp
```diff
@@ -942,8 +942,8 @@ DOFManager<LO,GO>::buildTaggedMultiVector(const ElementBlockAccess & ownedAccess
       const std::vector<LO> & fieldIds= fa_fps_[b]->fieldIds();
       const std::vector<LO> & myElements = connMngr_->getElementBlock(blockOrder_[b]);
       for (size_t l = 0; l < myElements.size(); ++l) {
-        LO connSize = connMngr_->getConnectivitySize(myElements[l]);
-        const GO * elmtConn = connMngr_->getConnectivity(myElements[l]);
+        auto connSize = connMngr_->getConnectivitySize(myElements[l]);
+        const panzer::Ordinal64 * elmtConn = connMngr_->getConnectivity(myElements[l]);
```
seamill: auto

packages/panzer/adapters-stk/test/field_manager_builder/field_manager_builder.cpp
seamill: Nah, no worries, mainly grunt work. I just remember in my changes associated with the workset, all of these entries are copied and pasted from the same source, so some regex may help later on.

seamill:
This merge looks good, I'm looking forward to seeing how clean things will look after the next changes. My only comments were on things that look like they're already in the works.

rppawlo:
@seamill - I've fixed the auto things. Could you reapprove this?

Drekar updates are done, almost finished with empire, still have to do charon. Will not merge this until all apps have branches ready for updates for these changes.

seamill:
Do you want me to submit an approved review? It says it automatically merges with 1 approving review, and you mentioned you didn't want to automerge (I'm not sure how this new interface works after so long on gitlab...).

packages/panzer/adapters-ioss/src/Panzer_IOSSConnManager.cpp
```diff
@@ -40,21 +40,970 @@
...
+#include "Ioss_ElementBlock.h"
 
 namespace panzer_ioss {
 
-template class IOSSConnManager<int>;
+using Teuchos::RCP;
+using Teuchos::rcp;
+
+IOSSConnManager::IOSSConnManager(Ioss::DatabaseIO * iossMeshDB)
+   : iossMeshDB_(iossMeshDB), iossRegion_(), iossNodeBlocks_(), iossElementBlocks_(),
```
mhoemmen: You don't need to invoke default constructors of members (e.g., `iossRegion_()`) explicitly.

packages/panzer/adapters-ioss/src/Panzer_IOSSConnManager.cpp
```diff
@@ -40,21 +40,970 @@
...
+	 elementBlocks_(), neighborElementBlocks_(),
+	 numUniqueEdges_(0), numUniqueFaces_(0), edgeNodeToEdgeMap_(), faceNodeToFaceMap_(),
+	 elementEdgeNodes_(), elementFaceNodes_(),
+	 elmtLidToGid_(), elmtLidToConn_(), connSize_(),
+	 connectivity_(), ownedElementCount_(0), sidesetsToAssociate_(),
+	 sidesetYieldedAssociations_(), elmtToAssociatedElmts_(), placeholder_()
+{
+
+  std::string error_message;
+  int bad_count;
```
mhoemmen: `int bad_count = 0;` (else it won't have a default value).

packages/panzer/adapters-ioss/src/Panzer_IOSSConnManager.cpp
```diff
@@ -40,21 +40,970 @@
...
+
+  // Check for a non-null database
+  TEUCHOS_TEST_FOR_EXCEPTION(iossMeshDB_ == nullptr, std::logic_error, "Error, iossMeshDB is nullptr.");
+  TEUCHOS_TEST_FOR_EXCEPTION(iossMeshDB_ == NULL, std::logic_error, "Error, iossMeshDB is NULL.");
```
mhoemmen: Line 116 is superfluous (or did you find a case where something could equal `NULL` but not `nullptr`? please tell me you didn't ;-) ).

packages/panzer/adapters-ioss/src/Panzer_IOSSConnManager.cpp
```diff
@@ -40,21 +40,970 @@
...
+	   std::string filename = iossMeshDB_->get_filename();
+
+	   // Verify that multiple open databases are supported
+	   TEUCHOS_TEST_FOR_EXCEPTION(IossDatabaseTypeManager::supportsMultipleOpenDatabases(type), std::logic_error, "Error, " << type << "does not support multiple open databases.");
+
+	   // temporarily try this with a _copy file
+	   size_t dotPosition = filename.find(".");
+	   std::string filenameBase(filename, 0, dotPosition);
+	   std::string filenameExtension(filename, dotPosition+1);
+	   std::string filenameCopy = filenameBase + "_copy." + filenameExtension;
```
mhoemmen: It looks like you might want to use POSIX's temp file creation facility here.  There's a way to put temp files in any directory you want and to give them any prefix name you want.

packages/panzer/adapters-ioss/src/Panzer_IOSSConnManager.cpp
```diff
@@ -40,21 +40,970 @@
...
+  connSize_.clear();
+  connSize_.resize(ownedElementCount_,0);
+}
+
+void IOSSConnManager::buildEdgeFaceCornerNodeMapping() {
 
-#ifndef PANZER_ORDINAL64_IS_INT
-//template class IOSSConnManager<panzer::Ordinal64>;
-template class IOSSConnManager<int64_t>; // IOSS currently does not have a
-                                         // get_field_data<long long int>() defined.
```
mhoemmen: Fun fact: `int64_t` could be either `long` or `long long`.  (It's a typedef and doesn't have a required type.)

packages/panzer/adapters-ioss/src/Panzer_IOSSConnManager.cpp
```diff
@@ -40,21 +40,970 @@
...
+  GlobalOrdinal numFaces = 0;
+
+  if (dim > 1) {
+    numEdges = edgeNodeToEdgeMap_.size();
+    if (dim > 2) {
+      numFaces = faceNodeToFaceMap_.size();
+    }
+  }
+
+#ifdef HAVE_MPI
```
mhoemmen: You should be able to avoid the `ifdef / endif` here for everything but the actual communicator creation, just with a run-time `if (comm.getSize() > 1)`.  That might give you an optimization for the single process case.

The fewer ifdefs you have, the easier it is to avoid build errors and bugs.

packages/panzer/adapters-ioss/src/Panzer_IOSSConnManager.cpp
```diff
@@ -40,21 +40,970 @@
...
+		blockmaxNodeId = *(std::max_element(blockConnectivity->begin(), blockConnectivity->end()));
+	  }
+          if (blockmaxNodeId > localmaxNodeId)
+            localmaxNodeId = blockmaxNodeId;
+	}
+	if (blockmaxElementId > localmaxElementId)
+          localmaxElementId = blockmaxElementId;
+  }
+
+#ifdef HAVE_MPI
```
mhoemmen: See above.  Also, you should merge the two all-reduces into a single all-reduce of an array of length two.
```c++
std::array<GlobalOrdinal, 2> lclMaxes {localmaxNodeId, localmaxElementId};
std::array<GlobalOrdinal, 2> gblMaxes;
Teuchos::reduceAll<int, GlobalOrdinal> (comm, Teuchos::REDUCE_MAX, 2, lclMaxes.data (), gblMaxes.data ());
```
Panzer is really worried about strong scaling these days.  If that's a thing, then you need to avoid latency-bound MPI communication wherever you can.

packages/panzer/adapters-stk/src/Panzer_STKConnManager.cpp
```diff
@@ -40,17 +40,413 @@
...
+
+void STKConnManager::applyPeriodicBCs(const panzer::FieldPattern & fp, GlobalOrdinal nodeOffset, GlobalOrdinal edgeOffset,
+                                      GlobalOrdinal faceOffset, GlobalOrdinal /* cellOffset */)
+{
+   using Teuchos::RCP;
+   using Teuchos::rcp;
+
+#ifdef HAVE_EXTRA_TIMERS
+  using Teuchos::TimeMonitor;
+  RCP<Teuchos::TimeMonitor> tM = rcp(new TimeMonitor(*TimeMonitor::getNewTimer(std::string("panzer_stk::STKConnManager::applyPeriodicBCs"))));
```
mhoemmen: TimeMonitor is a scope guard.  Do not handle it by pointer.  Do this:
```c++
#ifdef HAVE_EXTRA_TIMERS
  using Teuchos::TimeMonitor;
  TimeMonitor tM (*TimeMonitor::getNewTimer ("panzer_stk::STKConnManager::applyPeriodicBCs"));
#endif 
```

rppawlo:
> Do you want me to submit an approved review? It says it automatically merges with 1 approving review, and you mentioned you didn't want to automerge (I'm not sure how this new interface works after so long on gitlab...).

@seamill - yes. please submit an approved review. The automerge feature only happens if the automerge label is added to the PR (it's not added in this case - comment was to make sure no one else added that flag). Merging (including manual) is completely blocked until someone approves the PR.

rppawlo:
@mhoemmen - thanks for your comments on the code. The code in question - the IOSS adapter - was experimental code that has been deprecated. It showed up in this PR because I used a sed script to remove the templates in all occurrences of the ConnManager base class. We have kept the code around just in case we decide to revive the work, but is currently disabled in the build.

mhoemmen:
@rppawlo Cool, no worries then :-)

Issue #5083, opened 2019-05-03
Panzer MiniEM: Install binary and inputs if built


jjellio:
## Enhancement
@trilinos/panzer @cgcgcg @egphill @jhux2 

Currently, the mini app Panzer Mini-EM is not installed, and this makes using the app tedious when Trilinos is installed (and the build tree deleted).

## Outcomes:
If MiniEM + Examples are turned on, then the PanzerMiniEM_BlockPrec.exe will be installed to $INSTALL_PREFIX/bin/PanzerMiniEM_BlockPrec.exe

The required inputs (*.xml) will be installed to:
$INSTALL_PREFIX/bin/PanzerMiniEM_InputFiles/

(PR will be added)

PR #5084, opened 2019-05-03
Install MiniEM

jjellio:
This adds a target to install MiniEM as well as the needed input files

Closes: #5083

@trilinos/panzer @egphill @cgcgcg @bartlettroscoe 

## Motivation and Context
Make performance tracking easier since most teams install Trilinos and need the apps to be usable from the install directory

## Related Issues
Closes: #5083

## How Has This Been Tested?
Tested on White

## Checklist

 @bartlettroscoe I could not find a way to get Tribits to install arbitrary files. In this case, we need to install the `xml` input files. They really don't make sense inside `lib/` or `include/`

If there is a better way to do this, please let me know. The motivation is to allow easier performance tracking, since the build tree is usually blown away.

### DIFF
--- a/packages/panzer/mini-em/example/BlockPrec/CMakeLists.txt
+++ b/packages/panzer/mini-em/example/BlockPrec/CMakeLists.txt
@@ -23,6 +23,25 @@ TRIBITS_COPY_FILES_TO_BINARY_DIR(CopyBlockPrecFiles
   DEST_DIR ${CMAKE_CURRENT_BINARY_DIR}
   )
 
+# rather than list all files like above, I glob all xmls
+FILE(GLOB ${PACKAGE_NAME}_InputFiles *.xml)
+
+# allow the user to override the install location and
+# provide a simple defintion for the default
+# EXAMPLE_INSTALL_DIR/Panzer/
+IF (NOT DEFINED ${PACKAGE_NAME}_INSTALL_PREFIX)
+  SET(${PACKAGE_NAME}_INSTALL_PREFIX "${${PROJECT_NAME}_INSTALL_EXAMPLE_DIR}/${PACKAGE_NAME}/")
+ELSE()
+  # inform the user where it will go, since they overrode the default
+  PRINT_VAR(${PACKAGE_NAME}_INSTALL_PREFIX)
+ENDIF()
+
+INSTALL(FILES ${${PACKAGE_NAME}_InputFiles}
+        DESTINATION ${${PACKAGE_NAME}_INSTALL_PREFIX}
+       )
+INSTALL(TARGETS "${PACKAGE_NAME}_BlockPrec"
+        COMPONENT ${PACKAGE_NAME} RUNTIME DESTINATION ${${PACKAGE_NAME}_INSTALL_PREFIX})
+
 #################################################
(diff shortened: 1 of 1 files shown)


packages/panzer/mini-em/example/BlockPrec/CMakeLists.txt
```diff
@@ -23,6 +25,18 @@ TRIBITS_COPY_FILES_TO_BINARY_DIR(CopyBlockPrecFiles
...
+# rather than list all files like above, I glob all xmls
+file(GLOB ${PACKAGE_NAME}_InputFiles *.xml)
+
+# TODO: ask for comments on how this should be done
+# Tribits does not seem to support installing arbitrary files
+#
+# This will place all xml globbed above into $INSTALL_PREFIX/bin/PanzerMiniEM_InputFiles
+#
+install(FILES ${${PACKAGE_NAME}_InputFiles}
+        DESTINATION bin/${PACKAGE_NAME}_InputFiles/
```
bartlettroscoe: @jjellio, actually, as mentioned in:

* https://tribits.org/doc/TribitsBuildReference.html#setting-the-install-prefix

you should not hard-code `bin` but should use `${Project>_INSTALL_RUNTIME_DIR}`.

I guess we need to add a short "HowTo install extra files" to the document:

* https://tribits.org/doc/TribitsDevelopersGuide.html#howtos

bartlettroscoe: Actually, it would be better, in my opinion, to install this stuff under `${${PROJECT_NAME}_INSTALL_EXAMPLE_DIR}/Panzer/`

jjellio: The `bin` business was to ensure it matched what Tribits chose. I think the best approach is to use `${${PROJECT_NAME}_INSTALL_EXAMPLE_DIR}/Panzer/`. This does mean I will CMake install the binary + files to that directory, rather than Tribits Install the binary, and CMake install the files. (the distinction being calling cmake's install directly for both, rather than using `INSTALLABLE` with Tribits)

I added `${PACKAGE_NAME}_INSTALL_PREFIX` as a new variable that can override the installation location, and this variable defaults to `${${PROJECT_NAME}_INSTALL_EXAMPLE_DIR}/${PACKAGE_NAME}/`

I am waiting on the build to run, then I can see if actually does what I think it should.

Thanks for the input, I couldn't find a clear guide for this (I actually looked at SEACAS because they install their binaries, they use Cmake install directly)

jjellio:
@bartlettroscoe @jhu2 

By default, if MiniEM is enabled, it will now be installed into:
`${CMAKE_INSTALL_PREFIX}/${${PROJECT_NAME}_INSTALL_EXAMPLE_DIR}`
which is
`${CMAKE_INSTALL_PREFIX}/example/PanzerMiniEM/`

Optionally, you can specify:
`-D PanzerMiniEM_INSTALL_PREFIX="..."`
Which will override the install location

I tested this on White with Shared/Static linking, and in the shared case it installs to either location above and sets the RPATH correctly (so I believe this will be functional for the ATDM guys)

I will mark Auto Merge now

PR #5085, opened 2019-05-03
Teuchos:  ostream deprecation -- corrections for 12.16 release branch

kddevin:
@trilinos/teuchos  @rhoope @mseldre 
 

## Description
The ostream operator of serial dense matrix was deprecated in release 12.6.
This PR adds the TEUCHOS_DEPRECATION flags and fixes upstream packages.

## Motivation and Context

#4430 #4330 

## How Has This Been Tested?
On Mac with TEUCHOS_HIDE_DEPRECATED_CODE=ON.

```
cmake \
-D CMAKE_BUILD_TYPE:STRING="DEBUG" \
-D CMAKE_VERBOSE_MAKEFILE:BOOL=OFF \
\
-D MPI_BIN_DIR:PATH="/Users/kddevin/InstalledSoftware/openmpi-1.8.3_clang/bin" \
-D TPL_ENABLE_MPI:BOOL=ON \
-D MPI_EXEC_MAX_NUMPROCS:STRING=11 \
\
-D TPL_ENABLE_BinUtils:BOOL=OFF \
-D TPL_ENABLE_Pthread:BOOL=OFF \
\
-D CMAKE_C_FLAGS:STRING="-Wall  -pedantic -Wno-unknown-pragmas -Wno-narrowing -Wno-inline -Wshadow -Wdeprecated-declarations -Wempty-body  -Wignored-qualifiers -Wmissing-field-initializers  -Wsign-compare  -Wtype-limits   -Wuninitialized -Winit-self -fstrict-aliasing -Wno-long-long" \
-D CMAKE_CXX_FLAGS:STRING="-DTPETRA_ENABLE_DEPRECATED_CODE -Wall -pedantic -Wno-unknown-pragmas -Wno-narrowing -Wno-delete-non-virtual-dtor -Wno-inline -Wshadow -Wdeprecated-declarations -Wempty-body  -Wignored-qualifiers -Wmissing-field-initializers  -Wsign-compare  -Wtype-limits   -Wuninitialized -Winit-self -fstrict-aliasing" \
\
-D Trilinos_ENABLE_ALL_OPTIONAL_PACKAGES:BOOL=ON \
\
-D Trilinos_ENABLE_SHADOW_WARNINGS:BOOL=ON \
-D Trilinos_VERBOSE_CONFIGURE:BOOL=OFF \
-D Trilinos_ENABLE_Fortran:BOOL=OFF \
\
-D Teuchos_HIDE_DEPRECATED_CODE:BOOL=ON \
\
-D Trilinos_ENABLE_Stokhos:BOOL=ON \
-D Trilinos_ENABLE_Nox:BOOL=ON \
-D Trilinos_ENABLE_ROL:BOOL=ON \
-D ROL_ENABLE_EXAMPLES:BOOL=OFF \
-D Trilinos_ENABLE_MiniTensor:BOOL=OFF \
-D ROL_ENABLE_MiniTensor:BOOL=OFF \
-D Trilinos_ENABLE_Panzer:BOOL=ON \
-D Trilinos_ENABLE_Tempus:BOOL=ON \
-D Trilinos_ENABLE_PanzerAdaptersSTK:BOOL=OFF \
-D Trilinos_ENABLE_PanzerAdaptersIOSS:BOOL=OFF \
-D Trilinos_ENABLE_Thyra:BOOL=ON \
-D Trilinos_ENABLE_MueLu:BOOL=ON \
-D Trilinos_ENABLE_Anasazi:BOOL=ON \
-D Trilinos_ENABLE_TrilinosCouplings:BOOL=ON \
-D Trilinos_ENABLE_STKIO:BOOL=OFF \
-D Trilinos_ENABLE_STKUtil:BOOL=OFF \
\
-D Teuchos_ENABLE_STACKTRACE:BOOL=OFF \
..
```

## Checklist

### DIFF
--- a/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
+++ b/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
@@ -285,9 +285,14 @@ int main(int argc, char *argv[])
       MyOM->stream(Warnings) << "   ||<Y1,X1> - <X1,Y1>||     : " << err << endl;
       // test s.p.d. of <Y1,X1>: try to compute a cholesky
       lapack.POTRF('U',yTx1.numCols(),yTx1.values(),yTx1.stride(),&info);
-      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
-          "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " << info
-          << "\nyTx1: \n" << yTx1);
+      if (info != 0) {
+        std::stringstream os;
+        os << "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " 
+           << info << "\nyTx1: \n";
+        yTx1.print(os);
+        os << "\n";
+        TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error, os.str());
+      }
 
       // X2 ortho to Y1
       MVT::MvRandom(*X2);
@@ -345,9 +350,14 @@ int main(int argc, char *argv[])
       MyOM->stream(Warnings) << "   ||<Y2,X2> - <X2,Y2>||     : " << err << endl;
       // test s.p.d. of <Y2,X2>: try to compute a cholesky
       lapack.POTRF('U',yTx2.numCols(),yTx2.values(),yTx2.stride(),&info);
-      TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
-          "New <Y2,X2> not s.p.d.: couldn't computed Cholesky: info == " << info
-          << "\nyTx2: \n" << yTx2);
+      if (info != 0) {
+        std::stringstream os;
+        os << "New <Y2,X2> not s.p.d.: couldn't computed Cholesky: info == " 
+           << info << "\nyTx2: \n";
+        yTx2.print(os);
+        os << "\n";
+        TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,os.str());
+      }
     }
     MyOM->stream(Warnings) << endl;
 
--- a/packages/anasazi/epetra/test/OrthoManager/cxx_mattest.cpp
+++ b/packages/anasazi/epetra/test/OrthoManager/cxx_mattest.cpp
@@ -661,13 +661,13 @@ int testProjectAndNormalizeMat(RCP<MatOrthoManager<ST,MV,OP> > OM,
           if (err > ATOL*TOL) {
             sout << "S\n" << *S << endl;
             sout << "S_out\n" << *S_outs[o] << endl;
-            sout << "B_out\n" << *B_outs[o] << endl;
+            sout << "B_out\n"; (*B_outs[o]).print(sout); sout << endl;
             if (C_outs[o].size() > 0) {
               sout << "X1\n" << *X1 << endl;
-              sout << "C_out[1]\n" << *C_outs[o][0] << endl;
+              sout << "C_out[1]\n"; (*C_outs[o][0]).print(sout); sout << endl;
               if (C_outs[o].size() > 1) {
                 sout << "X22n" << *X2 << endl;
-                sout << "C_out[2]\n" << *C_outs[o][1] << endl;
+                sout << "C_out[2]\n"; (*C_outs[o][1]).print(sout); sout << endl;
               }
             }
           }
--- a/packages/anasazi/src/AnasaziBlockDavidsonSolMgr.hpp
+++ b/packages/anasazi/src/AnasaziBlockDavidsonSolMgr.hpp
@@ -679,8 +679,9 @@ BlockDavidsonSolMgr<ScalarType,MV,OP>::solve() {
 #       ifdef TEUCHOS_DEBUG
           {
             std::stringstream os;
-            os << "KK before HEEV...\n"
-              << *state.KK << "\n";
+            os << "KK before HEEV...\n";
+            (*state.KK).print(os); 
+            os << "\n";
             *out << os.str();
           }
 #       endif
@@ -695,8 +696,9 @@ BlockDavidsonSolMgr<ScalarType,MV,OP>::solve() {
             std::stringstream os;
             *out << "Ritz values from heev(KK):\n";
             for (unsigned int i=0; i<theta.size(); i++) *out << theta[i] << " ";
-            os << "\nRitz vectors from heev(KK):\n" 
-                                                 << S << "\n";
+            os << "\nRitz vectors from heev(KK):\n";
+            S.print(os);
+            os << "\n";
             *out << os.str();
           }
 #       endif
@@ -715,8 +717,9 @@ BlockDavidsonSolMgr<ScalarType,MV,OP>::solve() {
             std::stringstream os;
             *out << "Ritz values from heev(KK) after sorting:\n";
             std::copy(theta.begin(), theta.end(), std::ostream_iterator<ScalarType>(*out, " "));
-            os << "\nRitz vectors from heev(KK) after sorting:\n" 
-              << S << "\n";
+            os << "\nRitz vectors from heev(KK) after sorting:\n";
+            S.print(os);
+            os << "\n";
             *out << os.str();
           }
 #       endif
@@ -726,8 +729,9 @@ BlockDavidsonSolMgr<ScalarType,MV,OP>::solve() {
 #       ifdef TEUCHOS_DEBUG
           {
             std::stringstream os;
-            os << "Significant primitive Ritz vectors:\n"
-              << Sr << "\n";
+            os << "Significant primitive Ritz vectors:\n";
+            Sr.print(os);
+            os << "\n";
             *out << os.str();
           }
 #       endif
@@ -756,8 +760,9 @@ BlockDavidsonSolMgr<ScalarType,MV,OP>::solve() {
 #       ifdef TEUCHOS_DEBUG
           {
             std::stringstream os;
-            os << "Sr'*KK*Sr:\n"
-              << newKK << "\n";
+            os << "Sr'*KK*Sr:\n";
+            newKK.print(os);
+            os << "\n";
             *out << os.str();
           }
 #       endif
--- a/packages/anasazi/src/AnasaziIRTR.hpp
+++ b/packages/anasazi/src/AnasaziIRTR.hpp
@@ -889,15 +889,21 @@ namespace Anasazi {
         MVT::MvTransMv(ONE,*this->delta_,*this->Adelta_,AA);
         MVT::MvTransMv(ONE,*this->delta_,*this->Bdelta_,BB);
       }
-      this->om_->stream(Debug) << "AA: " << std::endl << AA << std::endl;;
-      this->om_->stream(Debug) << "BB: " << std::endl << BB << std::endl;;
+      this->om_->stream(Debug) << "AA: " << std::endl; 
+      AA.print(this->om_->stream(Debug));
+      this->om_->stream(Debug) << std::endl;
+      this->om_->stream(Debug) << "BB: " << std::endl; 
+      BB.print(this->om_->stream(Debug));
+      this->om_->stream(Debug) << std::endl;;
       {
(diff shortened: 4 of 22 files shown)


kddevin:
Sure, @bartlettroscoe .  I would be happy for you to take a crack at addressing the deprecation and the issues in #4330 and #4430.  Thank you.

bartlettroscoe:
@kddevin said:

> I would be happy for you to take a crack at addressing the deprecation and the issues in #4330 and #4430. Thank you.

Can you contact me offline with a P/T to do this work?

mhoemmen:
@kddevin Did you mean the 12.16 branch?

kddevin:
@mhoemmen Yes, 12.16. Good catch

kddevin:
@bartlettroscoe Can you outline/describe what you propose to do?  What do you mean by stream manipulator?  Will it avoid the conflicts described in #4330 and #4430?
I can look for a P/T but it would help to have the outline first so I can estimate level of effort.

bartlettroscoe:
@kddevin asked:

> Can you outline/describe what you propose to do? What do you mean by stream manipulator? 

See https://github.com/trilinos/Trilinos/issues/4330#issuecomment-461042814.   Beyond that it would be easiest just to create a quick branch with the basics that shows how this works. 

We don't want customers to have to hack up their code like you see, for example in:

* https://github.com/trilinos/Trilinos/pull/5085/files#diff-d2234cb4c06f7578675f88b33338d66eR288

In that example, the original code looks like:

```
  TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
          "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " << info
          << "\nyTx1: \n" << yTx1);
```

and the refactored code looks like:

```
  TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
          "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " << info
          << "\nyTx1: \n" << print(yTx1));
```

Make sense?  If not, I will post a quick PR that demonstrates this.  (And the full refactoring would not be that hard.)

mseldre:
@bartlettroscoe , 
I agree with you that the Anasazi example is not as clean as it should be.  An alternative to your suggestion of having print() return an ostream would be to just add the necessary mapping to Anasazi so that the previous code will compile, e.g.

```
namespace MY_CLIENT {

template <typename OrdinalType, typename ScalarType>
std::ostream& operator<<(std::ostream& s, const Teuchos::SerialDenseMatrix<OrdinalType, ScalarType>& data)
{ data.print(s); return s; } // or, better, a customized pretty-print output handler

}
```

The point in all of this was to move such code out of the Teuchos namespace so that it does not take precedence over client-supplied operators (due to ADL). Now that Teuchos operators are going away, Anasazi and any other client now has the option to define their own, whether just rebinding with the Teuchos print() function or providing a customized pretty-print. This seems preferable to me and more readable in the code.

bartlettroscoe:
@mseldre, @kddevin, @trilinos/teuchos,

To be clear about what I am suggesting, it is to do the following:

In `Teuchos_SerialDensMatrix.hpp`:

```
namespace Teuchos {

...

#ifndef TEUCHOS_HIDE_DEPRECATED_CODE
/// \brief DEPRECATED!
template<typename OrdinalType, typename ScalarType>
TEUCHOS_DEPRECATED std::ostream&
operator<< (std::ostream& os, const SerialDenseMatrix<OrdinalType, ScalarType>& obj)
{
  obj.print (os);
  return os;
}
#endif

/// \brief Ostream manipulator for SerialDensMatrix
template<typename OrdinalType, typename ScalarType>
struct SerialDensMatrixPrinter {
public:
  const SerialDenseMatrix<OrdinalType,ScalarType> &obj;
  SerialDensMatrixPrinter(const SerialDenseMatrix<OrdinalType,ScalarType> &obj_in)
    : obj(obj_in)
  {}
};

/** \brief Output SerialDenseMatrix object through its stream manipulator.  */
std::ostream&
operator<<(std::ostream &out, const SerialDensMatrixPrinter<OrdinalType,ScalarType> printer)
{
    printer.obj.print(out);
    return out;
}

/** \brief Return SerialDenseMatrix ostream manipulator
  * 
  * Use as:
  *
  *    Teuchos::SerialDenseMatrix<int,double> mat;
  *    ...
  *    std::cout << "This is my matrix:\n" << print(mat) << "and I like it very much!\n";
 */
SerialDensMatrixPrinter<OrdinalType,ScalarType>
print(const const SerialDenseMatrix<OrdinalType,ScalarType> &obj)
{ return SerialDensMatrixPrinter<OrdinalType,ScalarType>(obj); }


} // namespace Teuchos
``` 

This is just a classic C++ ostream manipulator and almost identical to [Teuchos::describe()](https://trilinos.org/docs/dev/packages/teuchos/doc/html/Teuchos__Describable_8hpp_source.html#l00209).

This is massively better and less managing to code quality to downstream customer code that forcing users to directly use `Teuchos::SerialDensMatrix::print()` like the refactorings in this PR are doing.

Again, this simple set of code allows you to take client code like this:

```
TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
          "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " << info
          << "\nyTx1: \n" << yTx1);
```

and just add the `print()` call around the object the refactored code looks like:

```
  TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error,
          "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " << info
          << "\nyTx1: \n" << print(yTx1));
```

instead of the refactored mess:

```
      if (info != 0) {
        std::stringstream os;
        os << "New <Y1,X1> not s.p.d.: couldn't computed Cholesky: info == " 
           << info << "\nyTx1: \n";
        yTx1.print(os);
        os << "\n";
        TEUCHOS_TEST_FOR_EXCEPTION(info != 0,std::runtime_error, os.str());
      }
```

The former is a trivial refactoring that does not damage the readability or maintainability of the code at all (IMHO) while the latter damages both (IMHO).

Any questions about this?

mseldre:
Even better, you could do both:
- Have the Teuchos::SerialDense* print() functions return the ostream instead of void
- You then have the option of using print() inline or adding the one-liner to map from operator<< within the client namespace:
`std::ostream& operator<<(std::ostream& s, ...) { return data.print(s); }  // or a pretty-print handler`

bartlettroscoe:
@mseldre said:

> Even better, you could do both:

Can you be specific with a code example?  I think I know what you are suggesting and it does not work out the way one might thing (it creates jumbled output).

mseldre:
To be clearer (I hope), the replacement of operator<< with calls to print() was never intended (by me) in client code, as these are exactly the type of ugly work-arounds that we have been able to remove from Dakota, now that we have the ability to provide our own operators.  You seem focused on the Anasazi example, and I completely agree that this is not the right way to go.  @rhoope would have to comment on the history of that -- I was under the impression that this was more of a short-term patch to get things building, but perhaps it has persisted.

The last thing I want to do is induce the need for these types of work-arounds on someone else.  I think the best solution is for clients provide their own operator<< for vector/matrix types that they use, so that they can tailor their output as desired for their particular context.  If they like the debug-level output coming from Teuchos print() for their use cases, then they can provide the one-liner mapping for their namespace to use it (as suggested above).  But now they also have the option to provide customized definitions, such as a pretty print for actual end-user consumption.  This approach seems preferable to having to update every instance of matrix output syntax all over their code / unit tests, as in the patch you are suggesting.

So having said that, having the Teuchos vector/matrix print() functions return an ostream seems like a very useful intermediate step that provides the option of refactoring in the way you suggest, without precluding the ability to provide the one-liner that I suggest (avoiding the refactoring altogether). In terms of your SerialDenseMatrixPrinter layer, I'm not sure what value this adds relative to just modifying the return type of the already public print() function.

bartlettroscoe:
@mseldre said:

>  In terms of your SerialDenseMatrixPrinter layer, I'm not sure what value this adds relative to just modifying the return type of the already public print() function.

Can you give me a concrete example of what you are suggesting?

mseldre:
Not sure what kind of example you are looking for, but here are the operators from dakota/src/dakota_data_io.hpp, spanning insert and extract for multiple Teuchos types.  Nothing fancy here.  They replace Teuchos print() with locally-defined pretty-print operators, as we want them for end-user output (not unit test debug output) in our case.

```
namespace Dakota {

/// std::istream extraction operator for SerialDenseVector
template <typename OrdinalType, typename ScalarType>
std::istream& operator>>(std::istream& s, Teuchos::SerialDenseVector<OrdinalType, ScalarType>& data)
{ read_data(s, data); return s; }

/// std::ostream insertion operator for SerialDenseVector
template <typename OrdinalType, typename ScalarType>
std::ostream& operator<<(std::ostream& s, const Teuchos::SerialDenseVector<OrdinalType, ScalarType>& data)
{ write_data(s, data); return s; }

/// std::istream extraction operator for SerialDenseMatrix
template <typename OrdinalType, typename ScalarType>
std::istream& operator>>(std::istream& s, Teuchos::SerialDenseMatrix<OrdinalType, ScalarType>& data)
{ read_data(s, data); return s; }

/// std::ostream insertion operator for SerialDenseMatrix
template <typename OrdinalType, typename ScalarType>
std::ostream& operator<<(std::ostream& s, const Teuchos::SerialDenseMatrix<OrdinalType, ScalarType>& data)
{ write_data(s, data, true, true, true); return s; } // default matrix formatting options

/// std::istream extraction operator for SerialSymDenseMatrix
template <typename OrdinalType, typename ScalarType>
std::istream& operator>>(std::istream& s, Teuchos::SerialSymDenseMatrix<OrdinalType, ScalarType>& data)
{ read_data(s, data); return s; }

/// std::ostream insertion operator for SerialSymDenseMatrix
template <typename OrdinalType, typename ScalarType>
std::ostream& operator<<(std::ostream& s, const Teuchos::SerialSymDenseMatrix<OrdinalType, ScalarType>& data)
{ write_data(s, data, true, true, true); return s; } // default matrix formatting options

} // namespace Dakota

```
Then of course we just insert vectors and matrices into ostreams and ofstreams as per usual.  And our Dakota namespace operators should be out of the way for clients that link with Dakota. 

If you instead wanted debug-level I/O and print was modified to return an ostream, then this could become, e.g.

```
namespace Dakota {

/// std::ostream insertion operator for SerialDenseMatrix
template <typename OrdinalType, typename ScalarType>
std::ostream& operator<<(std::ostream& s,  const Teuchos::SerialDenseMatrix<OrdinalType, ScalarType>& data)
{ return data.print(s); }

} // namespace Dakota
```

where I have assumed the optional modification to Teuchos print() to return the ostream

```
namespace Teuchos {

template<typename OrdinalType, typename ScalarType>
std::ostream& SerialDenseMatrix<OrdinalType, ScalarType>::print(std::ostream& os) const
{
  <existing implementation>
  return os;
}

} // namespace Teuchos
```

There's no real difference in the debug-level case from what Teuchos was supplying previously, beyond the namespace locations -- we're only addressing the precedence of argument lookups for these operators, allowing for client customizations (or not).

bartlettroscoe:
@mseldre, okay, as long as you are only using the return `std::ostream&` from the `print()` member function like:

```
template <typename OrdinalType, typename ScalarType>
std::ostream& operator<<(std::ostream& s,  const Teuchos::SerialDenseMatrix<OrdinalType, ScalarType>& data)
{ return data.print(s); }
```

then that is fine.  What is not fine is stuff like:

```
std::cout << "This is my great matrix:\n" << matrix.print(std::cout) << "it is really great!";
```
That has undefined behavior.

If we want to let downstream customers define their own `operator<<(std::ostream&, ...)` and `operator>>(std::istream&,...)` functions, I am fine with that.  I just don't think we should force customers to write their own and some of the example refactorings in this PR demonstrate that mess that can be created without them.  So why did each customer Trilinos package not define their own `operator<<()` function in this PR?  That is what is being suggested right?  But what about downstream packages that include these other packages and inherit multiple `operator<<()` functions?   Depending on how namespaces are used, that could create ambiguous calls.

kddevin:
Revoking this PR; see #5374.

PR #5087, opened 2019-05-03
MueLu/regionMG: separate setup and apply of coarse solver

### LINKED ISSUE #5072: MueLu: separate setup and solve of regionMG coarse level solver
## Enhancement
@trilinos/muelu 

On the coarse level, we need to perform a direct solve on the composite system. Right now, this is done by creating a Amesos2 KLU solver in every visit to the coarsest level.

Steps to do:
- Move the setup phase of the Amesos2 KLU solver to the setup of the hierarchy and store the factorizations
- Only leave the apply phase of the direct solver in the V-Cycle

Overall, this aims at reducing unnecessary (and duplicate) cost.

mayrmt:
@trilinos/muelu 

## Description

Split setup and solve of the composite coarse level solver into two parts. Do the setup during hierarchy construction and just apply the solver during the V-cycle.

I also have removed commented and obsolete code from my previous unsuccessful attempts to create a coarse solver.

## Motivation and Context

The matrix does not change, it's sufficient to create the solver once and then just re-use multiple times. This should save unecessary setup costs.

## Related Issues

* Closes #5072 

## How Has This Been Tested?

Build and regionMG tests pass locally.

## Checklist

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
@@ -944,6 +944,20 @@ void MakeCoarseLevelMaps(const int maxRegPerProc, const int numLevels,
   }
 } // MakeCoarseLevelMaps
 
+/* Reconstruct coarse-level maps (assuming fully structured grids)
+ *
+ * We know the regional map on the coarse levels since they are just the
+ * row maps of the coarse level operators. Though, we need to re-construct
+ * the quasiRegional and composite maps ourselves.
+ *
+ * We ultimately are only interested in the composite map on the coarsest level.
+ * Intermediate levels are dealt with along the way, because we go through all
+ * levels recursively.
+ *
+ * Assumptions:
+ * - fully structured grid
+ * - only on region per proc and group
+ */
 template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
 void MakeCoarseLevelMaps2(const int maxRegPerGID,
                           Teuchos::ArrayView<LocalOrdinal>  compositeToRegionLIDsFinest,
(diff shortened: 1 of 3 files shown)


packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
```diff
@@ -1326,6 +1383,48 @@ void createRegionHierarchy(const int maxRegPerProc,
 } // createRegionHierarchy
 
 
+// Wrapper to be used from the Matlab-based driver
```
lucbv: That is not a bad idea, as you realize, I was using default values for some parameters to comply with the matlab driver but this looks better.

mayrmt: I had trouble to add another argument with default value. It always becomes difficult if there are multiple default values...

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
```diff
@@ -1314,6 +1367,10 @@ void createRegionHierarchy(const int maxRegPerProc,
                               regMatrices,
                               coarseCompOp);
 
+  std::cout << mapComp->getComm()->getRank() << " | MakeCoarseCompositeDirectSolver ..." << std::endl;
+
+  coarseSolver = MakeCompositeDirectSolver(coarseCompOp);
```
lucbv: I think we should go a step further and actually do the LU factorization here, calling the following methods:
```
coarseSolver->symbolicFactorization();
coarseSolver->numericFactorization();
```
You can see an example in Amesos2 [here](https://trilinos.org/docs/r12.12/packages/amesos2/doc/html/TwoPartSolve_8cpp-example.html).

mayrmt: I call exactly these two methods inside `MakeCompositeDirectSolver()` as the last step before returning the solver object.

lucbv: Yes I see that now! Great then it's good with me : )

PR #5092, opened 2019-05-03
Muelu ETI bugfix for Clang

william76:
PLEASE DO NOT MERGE THIS YET
==========================
This PR is still a work-in-progress but we'd like to let it go through a cycle of PR testing and see what happens =)

Description
========

I was getting link errors on `muelu_lgn` where it was failing under clang on OSX because symbols from files in `Utils/*.cpp` and `MueCentral/*.cpp` were not there.

This is because a dependency was put in making MueLu dependent on MueLu_LGN.

In this fix I just added `MueCentral/*.cpp` and `Utils/*.cpp` to `SOURCES_LGN` in the MueLu CMakeLists.txt file so it can have them.  

I'm able to compile MueLu now and have also tested these changes on a RHEL7 machine too with successful compilation.

@trilinos/muelu 
@jhux2


packages/muelu/src/CMakeLists.txt
```diff
@@ -356,6 +356,9 @@ APPEND_GLOB(SOURCES */*.cpp)
 LIST(REMOVE_ITEM SOURCES ${DIR}/Graph/MueLu_LinkedList.cpp)
 SET(SOURCES_LGN ${MUELU_ETI_CPP_SOURCES_LGN})
 APPEND_GLOB(SOURCES_LGN ${DIR}/Graph/MueLu_LinkedList.cpp)
+APPEND_GLOB(SOURCES_LGN "${DIR}/MueCentral/*.cpp")
+APPEND_GLOB(SOURCES_LGN "${DIR}/Utils/*.cpp")
```
jhux2: This is adding `*.cpp` files to SOURCES_LGN that have already been added to SOURCES.  The result is that the same object files will appear in both `libmuelu.a` and `libmuelu_lgn.a`.

william76:
@jhux2 closing this out for now. I still can't link `muelu_lgn` locally on my OSX machine but this fix isn't the right thing.

PR #5110, opened 2019-05-06
NOX:  Fix possible overflow when computing matrix size in NOX::LAPACK::Matrix

etphipp:
Reported by Alan Heirich.


rppawlo:
Thanks @etphipp !

Issue #3471, opened 2018-09-20
Tpetra: Consider adding vector element-wise division function


pelesh:
Please consider adding element-wise division function for Tpetra (multi)vector:
z(i) = x(i)/y(i) for all i
This function is needed for Tpetra interface to [SUNDIALS](http://github.com/LLNL/sundials), but I guess it may be useful elsewhere.

mhoemmen:
Here is an example of how to implement this: https://github.com/trilinos/Trilinos/blob/802b9e46b0ecbd1298e149bf63608d5791ef41d8/packages/ifpack2/example/RelaxationWithEquilibration.cpp#L416

mhoemmen:
@trilinos/tpetra This should not be an instance method; it should be a free function.

Issue #1424, opened 2017-06-14
I know how to get rid of CUDA_LAUNCH_BLOCKING=1 requirement


mhoemmen:
@trilinos/tpetra @trilinos/ifpack2 @trilinos/muelu @trilinos/amesos2 @trilinos/zoltan2 @crtrott @nmhamster 

The issue is that the mirror View of a CudaUVMSpace View is just the same View.  That's not unreasonable, since it _is_ accessible from host.  However, it means that if you write to the View from (host, device), then want to access it from (device, host), you'll need to fence (otherwise you'll get "bus errors," invalid data, etc.).

I tried this and it worked.  First, if you need to create mirror Views, insist on them being host Views:
```
Kokkos::View<double*> x ("x", 100);
auto x_h = Kokkos::create_mirror_view (Kokkos::HostSpace (), x);
Kokkos::deep_copy (x_h, x);
fill_on_host (x_h);
Kokkos::deep_copy (x, x_h);
```
No need for fences in the above code.  Second, insert fences only if you plan to access the UVM Views on host without copying.

I've CC'd everyone because this is not just some Tpetra thing.  Everything downstream of Tpetra needs to pay attention to how they access Kokkos::View data, or data that ultimately live in Kokkos Views.

The best way to fix this would be to turn off Kokkos' use of UVM by default, then fix the resulting build and test errors.  However, I don't see that happening any time soon.

kddevin:
Thanks, @mhoemmen .  I know that we originally scheduled this work for FY18, but I've heard some chatter about it here, too.  Should we look for a way to reprioritize it to this summer?  Maybe look for someone to help you do it?  What do you think?

mhoemmen:
@kddevin Thanks for responding! :-)  I've been experimenting with this in my current project.  I don't have a sense for priority, but it's frustrating to require `CUDA_LAUNCH_BLOCKING=1`.  I think we could fix a lot of this with some small changes to Kokkos, e.g., requiring `deep_copy` from or to a UVM View to fence, and making `host_mirror_space` of `CudaUVMSpace` be `HostSpace`.  Mainly, I would like some downstream packages to know about and take some ownership for this issue.

ibaned:
> The best way to fix this would be to turn off Kokkos' use of UVM by default, then fix the resulting build and test errors. However, I don't see that happening any time soon.

[...]

Issue #768, opened 2016-11-02
Tpetra: Implement "dual view" semantics


mhoemmen:
@trilinos/tpetra 

Epic for collecting all stories relating to "dual view" semantics in Tpetra.

kddevin:
#8591 begins to address this issue

PR #5115, opened 2019-05-06
Tpetra: "New RTI" (Reduction Transform Interface), Part 1 (for_each, transform)

mhoemmen:
@trilinos/tpetra 

## Description

Tpetra once had something called the Reduction Transform Interface (RTI).  It let users apply a function to each entry of one or two (or three, I think) `Tpetra::Vector`s, and then possibly end with a reduction over the output Vector.

This commit starts anew with RTI, by adding the following features:

  1. `withLocalAccess`: Access the local data of a `Tpetra::)Multi)Vector` in a given memory space -- a declarative replacement for DualView-style functions
  2. `for_each`: Works like `std::for_each`, but for `Tpetra::(Multi)Vector`
  3. `transform`: Works like unary or binary `std::transform`, but for `Tpetra::(Multi)Vector`

(2) and (3) use (1).  All of these have extension points so that they can be adapted for other objects, such as CrsGraph or CrsMatrix.

## Motivation and Context

Users want to apply functions entrywise to a Vector or MultiVector.  They want to get at the local data, and they want Tpetra to handle the sync and modify flag stuff for them.  Tpetra wants the power to do things like lazy allocation (see #333) that go beyond what `Kokkos::DualView` allows.  Tpetra wants to protect users from holding on to owning `Kokkos::View` that really belong to Tpetra objects, because that can cause bugs (e.g., if the mesh that produced the linear system has changed) and unnecessary memory usage.

## Next steps

  1. Add `transform_reduce` (the "R" in "RTI")
  2. Think about how to do kernel fusion with a mixture of Vector and CrsMatrix (or at least provide a "residual" interface)

## Related Issues

* Closes: #3471 
* Related to #333, #364, #381, #415, #768, #1424, #1896


packages/tpetra/core/test/MultiVector/ForEach.cpp
```diff
@@ -0,0 +1,423 @@
...
+    reduceAll<int, int> (*comm, REDUCE_MIN, lclSuccess, outArg (gblSuccess));
+    TEST_ASSERT( gblSuccess == 1 );
+    if (gblSuccess != 1) {
+      out << "Returning early" << endl;
+      return;
+    }
+
+    out << "X.sync_device(); X.putScalar(666.0);" << endl;
+    X.sync_device ();
+    X.putScalar (666.0);
```
csiefer2: I see what you did there...

mhoemmen: "It's a solar number.  You may call me 'Little Sunshine.'"

mhoemmen:
@nmhamster Check this out.  If `withLocalAccess` were the only way to get at a Tpetra object's local data, then we could totally detach Tpetra from MPI and let the parallel distribution change whenever it wants.

mhoemmen:
Weird Intel build errors:
```
/scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/Trilinos/packages/tpetra/core/src/Tpetra_for_each_MultiVector.hpp(234): error: a built-in binary operator applied to a scoped enumeration requires two operands of the same type
        static_assert (argsType != EMultiVectorForEachFuncArgs::ERROR,
```

mhoemmen:
@trilinos/tpetra developers reviewed this on Wednesday, 08 May 2019.  We decided the following:

1. withLocalAccess is fine as is.
2. Remove `value&, row index, column index` and `value&, row index` overloads of `for_each.

My new commits implement these changes.  Tests pass with CUDA 9.2 + GCC 7.2.0 and with Intel 17.0.1 + OpenMP (with `OMP_NUM_THREADS=4`).  I also redesigned GetNonowningLocalObject to make specialization for custom types easier, and added tests that exercise with withLocalAccess for a custom type (to make sure that specialization for custom types actually works).

mhoemmen:
@trilinos/framework Those are old PR test results.  Not sure why they haven't updated.

mhoemmen:
Not sure what's up w/ the SERIAL build.  I tested CUDA 9.2 and Intel 17.0.1 + OpenMP by hand.  I'll take another look.

mhoemmen:
@trilinos/framework Dashboard link is broken: "Error: CDash cannot connect to the database."

mhoemmen:
I set up a GCC 4.9.3 "serial" (MPI disabled) OpenMP build, and got a build error:
```
.../packages/tpetra/core/src/Tpetra_for_each_MultiVector.hpp:189:38: error: invalid type argument of unary ‘*’ (have ‘auto’)
                 decltype (readWrite (*X_j).on (memSpace))>;
```
This could be a compiler bug, since other compilers have no problem deducing the type of `X_j` here.  I will try to work around by assigning `*X_j` to `Tpetra::Vector&` explicitly.

mhoemmen:
@trilinos/tpetra My work-around for GCC 4.9.3 issues was successful.  Please review if you have a chance.  Thanks!

prwolfe:
> @trilinos/framework Dashboard link is broken: "Error: CDash cannot connect to the database." Here is the link: https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-5115-test&field2=buildstarttime&compare2=84&value2=NOW

Must have been transient as it works for me at this point.

mhoemmen:
@prwolfe wrote:
> Must have been transient as it works for me at this point.

Cool, glad it works now :-D

mhoemmen:
@csiefer2 U R THE MAN

Issue #5117, opened 2019-05-07
Tpetra: CrsGraph & CrsMatrix still use deprecated enum value DynamicProfile


mhoemmen:
## Bug Report
@trilinos/tpetra 

### Description

When I disable deprecated code in Tpetra, I get the following build errors:
```
In file included from .../Trilinos/packages/tpetra/core/src/Tpetra_CrsGraph_SerialWrapperNode.cpp:52:
.../Trilinos/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp:2582:59: error: use of undeclared identifier
      'DynamicProfile'; did you mean 'StaticProfile'?
    return rcp (new graph_type (map, maxNumEntriesPerRow, DynamicProfile, params));
                                                          ^~~~~~~~~~~~~~
                                                          StaticProfile
.../Trilinos/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp:131:5: note: 'StaticProfile' declared here
    StaticProfile
    ^
In file included from .../Trilinos/packages/tpetra/core/src/Tpetra_CrsGraph_SerialWrapperNode.cpp:54:
In file included from .../Trilinos/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp:66:
In file included from .../Trilinos/packages/tpetra/core/src/Tpetra_Import_Util2.hpp:64:
.../Trilinos/packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp:5039:43: error: use of undeclared identifier
      'DynamicProfile'; did you mean 'StaticProfile'?
                                          DynamicProfile, params));
                                          ^~~~~~~~~~~~~~
                                          StaticProfile
```
When I fix those and enable tests, I get the following build errors:
```
.../Trilinos/packages/tpetra/core/test/CrsGraph/CrsGraph_insertGlobalIndicesFiltered.cpp:137:16: error: no member
      named 'DynamicProfile' in namespace 'Tpetra'; did you mean 'StaticProfile'?
      {Tpetra::DynamicProfile, Tpetra::StaticProfile};
       ~~~~~~~~^~~~~~~~~~~~~~
               StaticProfile
.../Trilinos/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp:131:5: note: 'StaticProfile' declared here
    StaticProfile
    ^
...
.../Trilinos/packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests1.cpp:51:17: error: no member named
[...]

Issue #5042, opened 2019-04-30
Tpetra:  fix deprecated case in RowMatrix::add


kddevin:
## Bug Report
@trilinos/tpetra

relate to #4701 

### Description

One case of RowMatrix::add is not handled correctly in the deprecation of DynamicProfile.  This case is rarely, if ever, used.  we should either fix it and test it so it works, or throw an error to state we don't support it.  (I prefer the first.)

for details and a proposed solution, see https://github.com/trilinos/Trilinos/pull/4701/files/cfd349400d2faa8009822abf165b7c7cfcfa7e74#diff-c109d71d09ef95780e788d789a64f2d9

GeoffDanielson:
I have this fixed in the branch "5042_deprecated_row_add"

mhoemmen:
@GeoffDanielson Best practice is to let develop-to-master promotion close the issue automatically, because the commit that closes the issue has a phrase like `"fix #NNNN` in it (where `#NNNN` is the GitHub issue number to close).

PR #5126, opened 2019-05-07
#5042: add throw to row matrix add

GeoffDanielson:
@trilinos/tpetra 

## Description
This changes an untested code path to a throw.

## Motivation and Context
Without dynamic profile, the case of adding two row matrices is not as cut-and-dried as before.  The team reached the conclusion that we should throw instead of proceeding.

## Related Issues

* Closes 5042

## How Has This Been Tested?
Current unit tests do not fail, as this is/was an untested code path.  Testing the added throw made certain that no currently-run unit tests will fail.

## Checklist

- [na] I have added tests to cover my changes.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_RowMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_RowMatrix_def.hpp
@@ -223,7 +223,8 @@ namespace Tpetra {
                                       constructorSublist));
       }
 #else
-      TEUCHOS_TEST_FOR_EXCEPTION(!A_rowMap->isSameAs (*B_rowMap),
+      // true: !A_rowMap->isSameAs (*B_rowMap)
+      TEUCHOS_TEST_FOR_EXCEPTION(true,
 				 std::invalid_argument,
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_RowMatrix_def.hpp
```diff
@@ -222,6 +222,12 @@ namespace Tpetra {
         C = rcp (new crs_matrix_type (C_rowMap, 0, DynamicProfile,
                                       constructorSublist));
       }
+#else
+      TEUCHOS_TEST_FOR_EXCEPTION(!A_rowMap->isSameAs (*B_rowMap),
```
mhoemmen: We've already tested this condition above.  Just do this:
```c++
TEUCHOS_TEST_FOR_EXCEPTION(true, std::invalid_argument, "error message goes here");
```

GeoffDanielson: Fair point.

mhoemmen:
This is purely academic, though, until we fix #5117 etc.

Issue #5147, opened 2019-05-09
Framework: Cuda PR testing fails all tests (PR Testing Blocked)


csiefer2:
Something is seriously messed up with the Cuda testing as part of the autotesting system.  *All* cuda tests are failing.

All PRs will likely be blocked until this is fixed.

https://github.com/trilinos/Trilinos/pull/5144

csiefer2:
@trilinos/framework @jwillenbring @prwolfe @william76

william76:
#5124 as well...

@jwillenbring @prwolfe @ZUUL42 

I had a quick glance over at the [cuda jobs on Jenkins][1] and they're definitely all failing.  Have any of you guys had a chance to look at it?

Is it possible it's an additive problem?  @bartlettroscoe are you seeing cuda errors with ATDM testing, or is this just affecting our Trilinos PRs?

[1]: https://ascic-jenkins.sandia.gov/job/trilinos-folder/job/Trilinos_pullrequest_cuda_9.2/

bartlettroscoe:
Something as gone wrong on 'ride'.  This is not a problem with the code.  If you look at the [ATDM Trilinos CUDA builds today](https://testing.sandia.gov/cdash/index.php?project=Trilinos&date=2019-05-10&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-&field2=buildname&compare2=63&value2=cuda), it is clear the problem is with 'ride' and not Trilinos itself.

Someone needs to contact 'ride-help' and ask what changed.  I know there was an upgrade of jenkins-srn.sandia.gov that completed Wednesday 5/1/2019.  That seems like the first place to start.

william76:
@bartlettroscoe Thanks!
I just saw a new email thread that Paul kicked off with ride-help and Si has responded.  =)

prwolfe:
I saw failures in atom jobs as well. I put in a ticket with the ride admins, but it's possible that the wrapper script may be the common thread.

prwolfe:
Otherwise I am with Ross - it's not at our end.

nmhamster:
@prwolfe / @william76 / @bartlettroscoe - can you give me nodes you were running on please?

csiefer2:
Thanks for hunting this down!

prwolfe:
I am seeing this on most of them

13, 16, 7, 12, 11, 10

Which is why I had thought of software changes…

-----------------------------------------------------------------------
[...]

PR #5149, opened 2019-05-10
ROL Trilinos Sync 09-May-2019

dridzal:
@trilinos/rol 

## Description
Regular merge between the ROL-Trilinos repo and the main Trilinos repo, via a Trilinos fork.

## Motivation and Context
Dynamic optimization (PinT) additions, Fletcher's method, bug fixes, etc.

## How Has This Been Tested?
Trilinos check-in tests, performed in ROL-Trilinos and in the Trilinos fork.


mhoemmen:
CUDA PR test is currently broken for reasons likely unrelated to Trilinos: #5147.

dridzal:
@mhoemmen , thanks for letting us know.  Most of the PR errors seem to be due to that issue.  We may, however, a few real ones, but it's hard to tell.  Do you know if I need to set the retest label, or if that is automatic?

mhoemmen:
@dridzal You shouldn't have to set the retest label.  The PR tester should revisit all the open PRs.  Once #5147 gets fixed, you might need to wait a while for all the open PRs to get through.

prwolfe:
> @dridzal You shouldn't have to set the retest label. The PR tester should revisit all the open PRs. Once #5147 gets fixed, you might need to wait a while for all the open PRs to get through.

This is not correct. The attester will only run a PR twice in a given configuration (just to keep machines available). If a new commit is pushed or the retest label is applied it will start over.

dridzal:
@prwolfe , so if I want to try again with the **same** code, I should set the retest label?

prwolfe:
> @prwolfe , so if I want to try again with the **same** code, I should set the retest label?

Yep!

dridzal:
Several bugs have been confirmed, and the fixes will be part of a new PR.

Issue #3178, opened 2018-07-24
Pull request testing should set -Werror


mhoemmen:
@vbrunini asks whether pull request testing could set `-Werror`, so as to avoid issues like #3177.

@trilinos/framework @khpierson 

## Expectations

Trilinos -- at least the library, not necessarily tests and examples -- should build warning-free.

## Current Behavior

See #3177.  There is an issue that it's impossible to fix warnings in some packages.

## Motivation and Context

Sierra builds with warnings as errors, so they want Trilinos to build warning-free.

## Possible Solution

Exclude legacy packages like ML.  Fix warnings.  Add `-Werror` to at least one PR build.

## Related Issues
* Related to #3177

bartlettroscoe:
We have to clean up the warnings in the PR builds before you can set `-Werror`.  However, we could do this package-by-package.  There is a way to do that with TriBITS I think (and if not, we can make sure there is).

bartlettroscoe:
CC: @jwillenbring, @srajama1, @kddevin, @rppawlo, @mperego (The Trilinos Product Area Leads)

@trilinos/framework, can we make this the official issue for cleaning up the warnings in the PR builds, one package at a time and setting `-Werror` as we have been discussing for the last several months?  (I could not find any other Trilinos GitHub issue that was tracking this effort.)

Also, can we please add the following warning flags to the GCC and Intel Trilinos PR builds before we start setting -Werror?  These were specifically requested by the ATDM Gemma code and they really need to have zero Trilinos warnings.  They have indicated that this is a major impediment to using Trilinos and to improving their own software quality.


1. For gcc/clang:
     -Wall
     -Wextra

2. For intel compilers (icc):
     -Wall  
     -Warray-bounds
     -Wchar-subscripts  
     -Wcomment   
     -Wenum-compare 
     -Wformat 
     -Wuninitialized 
     -Wmaybe-uninitialized 
     -Wmain 
     -Wnarrowing 
     -Wnonnull 
     -Wparentheses 
     -Wpointer-sign 
     -Wreorder 
     -Wreturn-type 
     -Wsign-compare 
     -Wsequence-point 
     -Wtrigraphs 
     -Wunused-function 
     -Wunused-but-set-variable 
     -Wunused-variable 
     -Wwrite-strings

srajama1:
@bartlettroscoe Thanks for taking the lead on this. I like the idea of starting with minimal set of Trilinos packages and keep adding more to that list.

[...]

PR #5153, opened 2019-05-10
Fix warnings and set Werror for Triutils in GCC 7.2.0 build

ZUUL42:
@trilinos/framework
@trilinos/triutils 

## Description

Fixed warnings and set Werror for Triutils in the GC 7.2.0 PR build.
Wrapped variable in #ifdef DEBUG Werror sees as unused only by asserts and reversed NULL checks protecting strcpys.
## Motivation and Context

This completes #4511 and is work toward Issue #3178, eventually setting -Werror for all packages.

## Checklist

- [x] My commit messages mention the appropriate GitHub issue numbers.
- [x] My code follows the code style of the affected package(s).
- [ ] My change requires a change to the documentation.
- [x] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [x] All new and existing tests passed.
- [x] No new compiler warnings were introduced.
- [ ] These changes break backwards compatibility.

PR #5158, opened 2019-05-13
Fix warnings and set Werror for Isorropia in GCC 7.2.0 build

ZUUL42:
@trilinos/framework
@trilinos/isorropia

## Description

Fixed warnings and set Werror for Isorropia in the GC 7.2.0 PR build.
Wrapped variable in #ifdef DEBUG Werror sees as unused only by asserts.
## Motivation and Context

This completes #5105 and is work toward Issue #3178, eventually setting -Werror for all packages.

## Checklist

### DIFF
--- a/packages/isorropia/example/probe_tridiag.cpp
+++ b/packages/isorropia/example/probe_tridiag.cpp
@@ -77,7 +77,10 @@ int main(int argc, char *argv[])
 
     std::vector<int> NumNz(NumMyElements);
 
-    int i, ierr;
+    int i;
+#ifdef DEBUG
+    int ierr;
+#endif
     for (i=0; i<NumMyElements; i++)
     {
         if (MyGlobalElements[i]==0 ||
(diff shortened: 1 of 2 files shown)


packages/isorropia/example/probe_tridiag.cpp
```diff
@@ -77,7 +77,10 @@ int main(int argc, char *argv[])
 
     std::vector<int> NumNz(NumMyElements);
 
-    int i, ierr;
+    int i;
+#ifdef DEBUG
```
mhoemmen: Is it really `#ifdef DEBUG`?  I thought it was `NDEBUG`, like this:
```c++
#ifndef NDEBUG
int ierr;
#endif 
```

ZUUL42: I've seen examples of both through out the code. I think there's been _DEBUG as well. So, I've been searching the package and using whichever was previously used in that package.
Isorropia specifically didn't have an example of any of these. So, I used what I had used on the previous package I touched.
I can change it if needed, but as is, it has already passed the GCC 7.2.0 PR test with Werror on.

mhoemmen: Are there any debug PR test builds?

ZUUL42: Not currently.
@trilinos/triutils just brought this up as well.
So this will also be needed for #5153 & #5159.
I'll make the changes and file a PR for the one that got merged into Triutils.

ZUUL42:
> Did you test in a debug build? It could be that PR testing doesn't exercise debug builds.

I had not. And currently there isn't a PR debug build. So, I was just going off what builds in the 7.2.0 PR build with `-Werror` set.
I thought I'd be able to move the Werror Issue forward, clearing up some of these simpler warnings,  without diving too deep into any of the packages.

mhoemmen:
@ZUUL42 wrote:
> I had not. And currently there isn't a PR debug build.

Then you'll need to build debug yourself before submitting the PR, at least until PR tests get a debug build.

You're touching packages that haven't been actively developed in a long time, so the more builds we can test before merging (and potentially finding a bug long after the merge), the better.

Issue #5148, opened 2019-05-09
Ifpack2:  Build errors in BlockTriDiContainer with float enabled.


vbrunini:
## Bug Report

@trilinos/Ifpack2

### Description
```
/scratch/vebruni/src/Trilinos/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp:2117:7:   required from 'void Ifpack2::BlockTriDiContainerDetails::performNumericPhase(const Teuchos::RCP<const typename Ifpack2::BlockTriDiContainerDetails::ImplType<MatrixType>::tpetra_block_crs_matrix_type>&, const Ifpack2::BlockTriDiContainerDetails::PartInterface<MatrixType>&, Ifpack2::BlockTriDiContainerDetails::BlockTridiags<MatrixType>&, typename Ifpack2::BlockTriDiContainerDetails::ImplType<MatrixType>::magnitude_type) [with MatrixType = Tpetra::RowMatrix<float, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >; typename Ifpack2::BlockTriDiContainerDetails::ImplType<MatrixType>::tpetra_block_crs_matrix_type = Tpetra::Experimental::BlockCrsMatrix<float, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >; typename Ifpack2::BlockTriDiContainerDetails::ImplType<MatrixType>::magnitude_type = float]'
/scratch/vebruni/src/Trilinos/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_def.hpp:235:9:   required from 'void Ifpack2::BlockTriDiContainer<MatrixType, Ifpack2::BlockTriDiContainerDetails::ImplSimdTag>::compute() [with MatrixType = Tpetra::RowMatrix<float, int, int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >]'
packages/ifpack2/src/Ifpack2_BlockTriDiContainer_Serial.cpp:61:1:   required from here
/scratch/vebruni/src/Trilinos/packages/kokkos-kernels/src/batched/KokkosBatched_Scale_Internal.hpp:28:20: error: no match for 'operator*=' (operand types are 'KokkosBatched::Experimental::Vector<KokkosBatched::Experimental::SIMD<float>, 16>' and 'const double')
           A[i*as0] *= alpha;
           ~~~~~~~~~^~~~~~~~
```
[...]

PR #5160, opened 2019-05-13
Fix for #5148

kyungjoo-kim:
@trilinos/ifpack2 

## Description

type mismatch in the kernels.

## Related Issues

#5148


kyungjoo-kim:
I got wrong thing.

PR #5167, opened 2019-05-13
Ifpack2 develop

kyungjoo-kim:
@trilinos/ifpack2 

## Description

This PR chages the data type of block tridiagonal matrices. With a cmake option, it can use smaller data type i.e., if ``double`` is used as ``scalar_type``, it can use ``float`` for the block tridiagonals. The way to set it up on cmake is as follows:
```
-D Ifpack2_ENABLE_BlockTriDiContainer_Small_Scalar=ON 
```
This also needs some update from kokkoskernels. Thus, the default behavior is OFF but I confirmed that the code works. With a brief testing, it does not gain much yet. Maybe it suits better for the Volta architecture (as it shows lower memory efficiency). 

## Motivation and Context

This line solver seems to hit the bandwidth limit. To save bandwidth, using smaller precision type might help. As inverting block tridiagonals behaves like a preconditioner, lowering the precision might gain more performance than losing from providing a less precise preconditioner.


packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -292,6 +306,9 @@ namespace KB = KokkosBatched::Experimental;
       typedef typename Kokkos::Details::ArithTraits<scalar_type>::val_type impl_scalar_type;
       typedef typename Kokkos::ArithTraits<impl_scalar_type>::mag_type magnitude_type;
 
+      typedef typename BlockTridiagScalarType<impl_scalar_type>::type btdm_scalar_type;
```
mhoemmen: Note to self: This should work fine with complex arithmetic enabled, since `impl_scalar_type` would be `Kokkos::complex<T>` in that case, not `std::complex<T>`.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1776,7 +1794,7 @@ namespace KB = KokkosBatched::Experimental;
       const int vector_size = vector_length/internal_vector_length;
       int total_team_size(0);
       if      (blksize <=  5) total_team_size =  32;
-      else if (blksize <=  9) total_team_size =  64;
+      else if (blksize <=  9) total_team_size =  32; // 64
```
mhoemmen: Kinda curious where these magic numbers come from :-)

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -160,6 +165,15 @@ namespace KB = KokkosBatched::Experimental;
                                  typename ViewType::execution_space::scratch_memory_space,
                                  MemoryTraits<typename ViewType::memory_traits, Kokkos::Unmanaged> >;
 
+    ///
+    /// block tridiag scalar type
+    ///
+    template<typename T> struct BlockTridiagScalarType { typedef T type; };
+#if defined(IFPACK2_BLOCKTRIDICONTAINER_USE_SMALL_SCALAR_FOR_BLOCKTRIDIAG)
+    template<> struct BlockTridiagScalarType<double> { typedef float type; };    
+    //template<> struct SmallScalarType<Kokkos::complex<double> > { typedef Kokkos::complex<float> type; };
```
mhoemmen: I'm OK with leaving this out, since it would make the build trickier.

mhoemmen:
Please also note that PR testing will not enable this option by default, so we'll have to rely on manual testing to ensure correctness.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
kyungjoo-kim: On Cuda, I use vector length 16 and would like to use doubl2 (or float4). This gives blockDim.x is 8 (or 4) -- interleaved vector parallelism. Giving 32 total team size means that I set blockDim.y 32/8=4 (or 32/4=8) -- matrix parallelism. The maximum matrix parallelism would be limited by the blocksize (gemm can have blocksize^2 parallelism but trsm and lu would explore about blocksize parallelism -- limiting factor. So, setting blockDim.y = 4 would be good for blocksize 4. Then, I just empirically test a few team size multiplication of 32. If we have large enough workload to exploit the GPU, the 32 team size would suffice for all cases. As we underutilize the device, it would be good to give the large team size so that it reaches the algorithm parallelism limit.

kyungjoo-kim:
@mhoemmen correct. The PR test does not have this test. If SPARC find out this approach useful, then we can set this two scalar type approach as default.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
mhoemmen: Thanks @kyungjoo-kim ! :D

PR #5175, opened 2019-05-14
Fix ifdef DEBUGs to NDEBUG

ZUUL42:
@trilinos/triutils 

## Description

Correcting `ifdef DEBUG` to `NDEBUG` for recent changes in #5153.

## Checklist

### DIFF
--- a/packages/triutils/src/Trilinos_Util_ReadHpc2Epetra.cpp
+++ b/packages/triutils/src/Trilinos_Util_ReadHpc2Epetra.cpp
@@ -80,7 +80,7 @@ void Trilinos_Util_ReadHpc2Epetra_internal(
     exit(1);
   }
   int_type numGlobalEquations, total_nnz;
-#ifdef DEBUG
+#ifndef NDEBUG
   int cnt;
 #endif
   // mfh 24 Mar 2015: We use temporaries of the type corresponding to
@@ -91,12 +91,12 @@ void Trilinos_Util_ReadHpc2Epetra_internal(
   // the warning is easy to fix.
   if(sizeof(int) == sizeof(int_type)) {
     int numGlobalEquations_int, total_nnz_int;
-#ifdef DEBUG
+#ifndef NDEBUG
     cnt =
 #endif
     fscanf(in_file,"%d",&numGlobalEquations_int);
     assert(cnt > 0);
-#ifdef DEBUG
+#ifndef NDEBUG
     cnt =
 #endif
     fscanf(in_file,"%d",&total_nnz_int);
@@ -107,12 +107,12 @@ void Trilinos_Util_ReadHpc2Epetra_internal(
   }
   else if(sizeof(long long) == sizeof(int_type)) {
     long long numGlobalEquations_ll, total_nnz_ll;
-#ifdef DEBUG
+#ifndef NDEBUG
     cnt =
 #endif
     fscanf(in_file,"%lld",&numGlobalEquations_ll);
     assert(cnt > 0);
-#ifdef DEBUG
+#ifndef NDEBUG
     cnt =
 #endif
     fscanf(in_file,"%lld",&total_nnz_ll);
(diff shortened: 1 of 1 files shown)


packages/triutils/src/Trilinos_Util_ReadHpc2Epetra.cpp
```diff
@@ -80,7 +80,7 @@ void Trilinos_Util_ReadHpc2Epetra_internal(
     exit(1);
   }
   int_type numGlobalEquations, total_nnz;
-#ifdef DEBUG
+#ifdef NDEBUG
```
mhoemmen: It's `ifndef NDEBUG`.  `assert` only exists if `NDEBUG` is _not_ defined.

https://en.cppreference.com/w/c/error/assert

ZUUL42: Ahh!
I'm trying to burn through this a bit too fast today.

bathmatt:
Let me know via my SNL email what other patches you need reviewed, you mentioned yesterday that there were other MRs ??

packages/triutils/src/Trilinos_Util_ReadHpc2Epetra.cpp
```diff
@@ -91,12 +91,12 @@ void Trilinos_Util_ReadHpc2Epetra_internal(
   // the warning is easy to fix.
   if(sizeof(int) == sizeof(int_type)) {
     int numGlobalEquations_int, total_nnz_int;
-#ifdef DEBUG
+#ifndef NDEBUG
     cnt =
 #endif
     fscanf(in_file,"%d",&numGlobalEquations_int);
     assert(cnt > 0);
```
jmgate: Doesn't this line need to be inside one of the `#if` blocks?

mhoemmen: @jmgate It's fine -- `assert` has the following definition, according to https://en.cppreference.com/w/cpp/error/assert :
```c++
#ifdef NDEBUG
#define assert(condition) ((void)0)
#else
#define assert(condition) /*implementation defined*/
#endif
```

bartlettroscoe: The macro `assert()` is only defined if the macro `NDEBUG` is **not** defined.

bartlettroscoe:
We need to make one of the Trilinos PR builds set `CMAKE_BUILD_TYPE=DEBUG` which will not define `NDEBUG` and will catch this type of error in the future.  See #5184.

ZUUL42:
> We need to make one of the Trilinos PR builds set `CMAKE_BUILD_TYPE=DEBUG` which will not define `NDEBUG` and will catch this type of error in the future. See #5184.

We were just discussing the possibility and path forward for a PR debug build at the standup this morning. Nothing concrete yet.

packages/triutils/src/Trilinos_Util_ReadHpc2Epetra.cpp
jmgate: Does the fact that all other occurrences of `cnt` are wrapped in an `#if` not come into play here?  @mhoemmen, it sounds like you're saying line 98 gets replaced with `((void)0)` so it won't throw this error again?
```
error: ‘cnt’ was not declared in this scope
```

I'm asking Santa for a Trilinos without `#ifdef`s this year.

mhoemmen: @jmgate wrote:
> @mhoemmen, it sounds like you're saying line 98 gets replaced with `((void)0)` so it won't throw this error again?

Yes, that's correct.  `assert` is a macro.  If `NDEBUG` is defined, then the macro's argument does not appear in the compiled source code.  That is, in fact, why the compiler was warning about "unused variables."

> I'm asking Santa for a Trilinos without `#ifdef`s this year.

I ... don't have anything constructive to say about this ;-)

Issue #5135, opened 2019-05-08
MueLu: regionMG residual is 2x as large as expected at region interfaces


mayrmt:
## Bug Report
@trilinos/muelu 

### Description
Here's a brief summary provided by @rstumin: 
"In particular, I am doing a 2 region example on a 13 x 13 mesh .... so there is just one interface. Along this interface the composite residual is  2x of what I expect it to be. Specifically, I initialize x to be 0 ... so I am expecting the composite residual to be r = b - A x = b. I've attached rhs.mm and compres.mm. You can do a diff -w to see the differences. Basically, we take the original composite b and convert it to regB. We do the same for composite x ... converting it to regX. The problem is that in the case of regX we want to replicate the values on the interface. However, I think for regB we want to scale things by the number of regions sharing an interface. Instead we replicate B along the interface. This means that when we compute a regional residual and then sum along the interface to get back to a composite residual ... we get 2x along the interface."

### Steps to Reproduce

Run the `MueLu_StructuredRegionDriver.exe` and look at the composite residual. In particular, compare 
- the composite residual computed with composite vectors/matrices only
- the composite residual computed via the region framework and then transforming it back to the composite layout.

They are expected to be identical, though they are not.

### Possible solution

Introduce a scaling option into `compositeToRegional()`, such that one *can* scale interface values by 1/N with N being the number of adjacent regions.

lucbv:
@mayrmt let us see if @rstumin latest batch of tests using the Matlab code returns without error, if so I propose that we close this issue as it would seem that code is know working properly.

mayrmt:
@lucbv I agree.

lucbv:
@mayrmt I think it's time to close, feel free to reopen if needed!

PR #5178, opened 2019-05-15
MueLu: add inverse scaling of composite residual

mayrmt:
@trilinos/muelu 

## Description

When moving the residual from regional to composite, we need a scaling of interface values. This adds the scaling just prior to the convergence check in the outer fix-point iteration.

Maybe we need another scaling on the coarse level before entering the direct solver.

## Motivation and Context

As reported in #5135, the composite residual is twice as large at the interface as expected. Not good. This should address this.

## Related Issues

* Part of #5135

## How Has This Been Tested?

Builds locally. Compared composite residual as computed via r=b-Ax (with x=0) and computed via the region framework before entering the V-cycle. Now, they match.

## Checklist

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
@@ -82,6 +82,41 @@ using Teuchos::Array;
 using Teuchos::ArrayView;
 using Teuchos::ParameterList;
 
+/*! \brief Apply scaling to interface DOFs
+ *
+ * The vector scalingFactors contains the number of adjacent regions for every DOFs.
+ * In the interior of a region, this is always 1. On region interfaces, this is >1.
+ *
+ * We often need to scale interface entries by 1/N with N being the number of adjacent regions.
+ * This can be achieved by setting \c inverseScaling to \c true.
+ */
+template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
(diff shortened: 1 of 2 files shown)


mayrmt:
@rstumin I added the missing inverse scaling when restricting the residual. I could not reproduce the residual norms that you send me via email, however this might just be related to the initial guess. The driver uses a random initial guess for `X`, so I might solve a different problem than you. 

Anyway, let me know, if this is ok now when you compare to your Matlab implementation. If yes, I will create a routine `inverseScaling()` in `SetupRegionHierarchy_def.hpp`, that can be called from all places in the code where we need the inverse scaling.

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
```diff
@@ -1640,7 +1640,9 @@ void vCycle(const int l, ///< ID of current level
       coarseRegX[j] = VectorFactory::Build(regRowMaps[l+1][j], true);
       coarseRegB[j] = VectorFactory::Build(regRowMaps[l+1][j], true);
 
-      regRes[j]->elementWiseMultiply(SC_ONE, *regRes[j], *((regInterfaceScalings[l])[j]), SC_ZERO);
+      RCP<Vector> inverseInterfaceScaling = VectorFactory::Build(regInterfaceScalings[l][j]->getMap());
```
mhoemmen: This sequence would be a good candidate for kernel fusion later on.

mayrmt:
@lucbv I moved the inverse scaling of interface DOFs to its own function for convenience. Residual history is still the same, so we can proceed to merge this.

PR #5188, opened 2019-05-15
Piro: reporting final point to the wrapped ME at the end of NOX solve

bartgol:
@trilinos/nox 

## Description

Make Piro::NOXSolver set the final point into the underlying model. If the underlying model overrides `reportFinalPoint`, then other customers of the model can grab the final solution for their own use.

## Motivation and Context

I have two customers of a model ME: the NOXSolver, and another class, call it MyClass. Unfortunately, MyClass does not have access to the NOXSolver, so upon completion of the NOX solve, it can't get the solution from the NOX solver (in principle, it does not even know that the ME used to solve the model was Piro::NOXSolver either). With this modification, Piro::NOXSolver can set the solution in the model, so that MyClass can later access it (assuming that the underlying ME overrides `reportFinalPoint`, and can expose the reported final point, but that's not a concern of Piro::NOXSolver).

## Checklist

### DIFF
--- a/packages/piro/src/Piro_InvertMassMatrixDecorator_Def.hpp
+++ b/packages/piro/src/Piro_InvertMassMatrixDecorator_Def.hpp
@@ -312,13 +312,10 @@ template<typename Scalar>
 void
 Piro::InvertMassMatrixDecorator<Scalar>::reportFinalPoint(
 #endif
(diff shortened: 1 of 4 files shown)


bartgol:
The failure is due to an implementation detail of the test. Should be an easy fix. I'll push it tomorrow morning.

PR #5191, opened 2019-05-16
SEACAS: ioss/cgns - fix update_db_zone_property 

gsjaardema:
The `update_db_zone_property` function should only be called when the cgns database is closed and then reopened during an execution; not when appending to a database in a restart run.  In the first case, the region pointer is non-null since the region has already been constructed and we need to make sure the zone ordering is the same for the existing region and the newly-reopened cgns database.  In the restart case, we read the zones from the database so the order is correct by definition.

@trilinos/seacas 

## Checklist

### DIFF
--- a/packages/seacas/libraries/ioss/src/cgns/Iocgns_DatabaseIO.C
+++ b/packages/seacas/libraries/ioss/src/cgns/Iocgns_DatabaseIO.C
@@ -534,7 +534,7 @@ namespace Iocgns {
         set_int_byte_size_api(Ioss::USE_INT64_API);
       }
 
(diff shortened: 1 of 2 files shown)


gsjaardema:
@bartlettroscoe This is probably related to the SPAR-569 tests although it was not reported to me via that.  I am working on getting a test in place for this.

bartlettroscoe:
@gsjaardema, if no SEACAS tests were failing and you were not aware of the SPARC Trilinos Integration test failures, how did you get tipped off that this was a problem?

gsjaardema:
@bartlettroscoe Email from one of the SPARC developers.

PR #5197, opened 2019-05-17
Intrepid2 - cuda orientation tools

kyungjoo-kim:
@trilinos/intrepid2 

## Description

Cuda interface for orientation tools

## Motivation and Context

This feature is requested from @mperego 

## How Has This Been Tested?

New tests on cuda are established and passed on white.

### DIFF
--- a/packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefModifyBasis.hpp
+++ b/packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefModifyBasis.hpp
@@ -54,7 +54,7 @@
 #endif
 
 namespace Intrepid2 {
-  
+
   template<typename SpT>
   template<typename ptViewType>
   KOKKOS_INLINE_FUNCTION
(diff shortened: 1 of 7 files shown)


packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefModifyBasis.hpp
```diff
@@ -151,121 +291,40 @@ namespace Intrepid2 {
...
-                }
-            }
-          }
-        }
-        
-      }
+
+      const Kokkos::RangePolicy<SpT> policy(0, numCells);
+      typedef F_modifyBasisByOrientation
+        <decltype(orts),
```
mhoemmen: Best practice with the "implementation struct and templated function" pattern using Kokkos::View is to have the function "canonicalize" the struct's template arguments.  This avoids redundant specializations of the struct.

kyungjoo-kim: I originally want to use labmda but somehow it does not compile. I just make it as a functor. Here look and feel does not really matter. Thanks if you can approve the changes.

mhoemmen: Done! :-)

PR #5200, opened 2019-05-17
MueLu: region driver coordinates

lucbv:
@trilinos/muelu 

## Description
Fixing coordinates in the region driver, the logic around them was never properly put in place...

## Motivation and Context
This fixes a bug which prevents from computing the appropriate prolongator in region format


prwolfe:
All three of the machines we use for intel jobs were down starting at 3:15pm Friday and continuing until  11 am Saturday. It looks like all the aspic machines were down and the gcc jobs that did try to start were on the smaller cloud nodes. Everything seems to be up at this point.

prwolfe:
Sorry - wrong window, this note and close was meant for the bug report.

prwolfe:
Adding retest to check my hypothesis about the weekends failures.

PR #5217, opened 2019-05-21
Fix warnings and set Werror for Anasazi in GCC 7.2.0 build

ZUUL42:
@trilinos/framework 
@trilinos/anasazi 

## Description

Fixed warnings and set Werror for Anasazi in the GC 7.2.0 PR build.
Wrapped variables only used by asserts Werror sees as unused in TEUCHOS_ASSERT().
Removed a few sections of unreachable code sitting behind always-false if statements.
## Motivation and Context

This completes #4475 and is work toward Issue #3178, eventually setting -Werror for all packages.

## How Has This Been Tested?

This was tested in debug and release Jenkins builds.

## Checklist

### DIFF
--- a/packages/anasazi/epetra/example/BlockDavidson/BlockDavidsonEpetraEx.cpp
+++ b/packages/anasazi/epetra/example/BlockDavidson/BlockDavidsonEpetraEx.cpp
@@ -46,6 +46,7 @@
 #include "Epetra_CrsMatrix.h"
 #include "Teuchos_CommandLineProcessor.hpp"
 #include "Teuchos_StandardCatchMacros.hpp"
+#include "Teuchos_Assert.hpp"
 
 #ifdef HAVE_MPI
 #include "Epetra_MpiComm.h"
@@ -250,7 +251,7 @@ int main(int argc, char *argv[]) {
 
     // Finish up
     int info = A->FillComplete();
-    assert( info==0 );
+    TEUCHOS_ASSERT( info==0 );
     A->SetTracebackMode(1); // Shutdown Epetra Warning tracebacks
 
     // Create a identity matrix for the temporary mass matrix
@@ -261,11 +262,11 @@ int main(int argc, char *argv[]) {
       Indices[0] = i;
       NumEntries = 1;
       info = M->InsertGlobalValues(MyGlobalElements[i], NumEntries, &Values[0], &Indices[0]);
-      assert( info==0 );
+      TEUCHOS_ASSERT( info==0 );
     }
     // Finish up
     info = M->FillComplete();
-    assert( info==0 );
+    TEUCHOS_ASSERT( info==0 );
     M->SetTracebackMode(1); // Shutdown Epetra Warning tracebacks
 
     //************************************
--- a/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
+++ b/packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
@@ -637,14 +637,6 @@ int testProjectAndNormalizeGen(RCP<GenOrthoManager<ST,MV,OP> > OM,
       theMY = tuple(MY1);
       C = tuple( rcp(new SerialDenseMatrix<int,ST>(sizeX1,sizeS)) );
     }
-    else if ( (t && 3) == 2 ) {
-      // <X2,Y2>
-      theX = tuple(X2);
-      theY = tuple(Y2);
-      theMX = tuple(MX2);
-      theMY = tuple(MY2);
-      C = tuple( rcp(new SerialDenseMatrix<int,ST>(sizeX2,sizeS)) );
-    }
     else {
       // <X1,Y2> and <X2,Y2>, and the reverse.
       theX = tuple(X1,X2);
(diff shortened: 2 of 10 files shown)


packages/anasazi/epetra/example/BlockDavidson/BlockDavidsonEpetraEx.cpp
```diff
@@ -249,7 +249,10 @@ int main(int argc, char *argv[]) {
     }
 
     // Finish up
-    int info = A->FillComplete();
+#ifndef NDEBUG
```
mhoemmen: Anasazi has a required dependence on Teuchos.  You could do this instead:
```c++
#include "Teuchos_Assert.hpp"
// ...
int info = A->FillComplete();
TEUCHOS_ASSERT( info == 0 );
```

packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
```diff
@@ -625,19 +625,19 @@ int testProjectAndNormalizeGen(RCP<GenOrthoManager<ST,MV,OP> > OM,
     Array<RCP<const MV> > theX, theY, theMX, theMY;
     RCP<SerialDenseMatrix<int,ST> > B = rcp( new SerialDenseMatrix<int,ST>(sizeS,sizeS) );
     Array<RCP<SerialDenseMatrix<int,ST> > > C;
-    if ( (t && 3) == 0 ) {
+    if ( (t & 3) == 0 ) {
```
mhoemmen: We've tried to fix this before, and it broke the tests.  Do the tests pass for you?

ZUUL42: That's one thing I'm sorting out now. I swore I ran this through on my own as a GCC 7.2.0 debug build and they all passed just fine. Now that I look at it closer, apparently I was discounting these tests failing as a false positive from something else I was working on.
I could remove the changes, but then I won't be able to set Werror.

ZUUL42: Though the way it was, it doesn't look like it would ever enter anything except the `((t && x) == 0)` or `( (t && x) == 1 )` lines. It would never find its way into `( (t && x) == 2 )` or `( (t && x) == 3 )`. Those are the ones the Werrors are complaining about:
```
error: comparison of constant ‘2’ with boolean expression is always false [-Werror=bool-compare]
error: comparison of constant ‘3’ with boolean expression is always false [-Werror=bool-compare]
```
So it seems like the tests are sort of broken to start with, or at very least include extraneous `if` statements. I just don't know enough about what they are doing to make sure they get fixed properly without really digging in.
The other simple solution would be to remove the `==2) {...}` & `==3) {...}` segments entirely because they are never actually run anyway as things currently are.
That just seemed too simple before. So, I was taking a stab at what might have been intended given similar if statements above these.
But does just removing the unused `if` statements seem reasonable? If everything is testing correctly, that actually seems like the correct answer.

mhoemmen: @ZUUL42 wrote:
> I could remove the changes, but then I won't be able to set Werror.

It would make sense to separate out the `TEUCHOS_ASSERT` changes from the `&& 3` etc. changes.  @hkthorn has seen the latter before and might have some insight.

ZUUL42: Just to try it out, I created a branch and removed the 8 segments of what appear to be dead code the compiler is complaining about, everything `if ( (t && 3) == 2 ) {...}` and `if ( (t && 3) == 3 ) {...}`. See [commit 3196142](https://github.com/ZUUL42/Trilinos/commit/3196142d82af4c3855f3cb19fa72da7e9b23db0e).
I built it with the GCC 7.2.0 PR tester and everything builds without error and passes the tests. However, since that removes a good chunk of code, I wanted to check with @trilinos/anasazi before I push it to this branch.

mhoemmen: @ZUUL42 I just e-mailed @hkthorn to ask about this.

ZUUL42: Since it has been 4 weeks, I'd like to propose a path forward.
The compiler is tagging these expressions always false and thus everything inside is never used.
If I remove these sections, the functionality _should_ remain unchanged.
That will get this PR moving forward and allow me to set Werror for Anasazi.
Then if there is any reason to bring back any of the removed code, that can be done with the Werror flag in place.

ZUUL42:
Anasazi is the last remaining package before [Werror can be set globally](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=3&showfilters=1&filtercombine=and&field1=site&compare1=61&value1=tr-test-0.novalocal&field2=buildname&compare2=61&value2=PR-Werror_all-test-Trilinos_pullrequest_gcc_7.2.0-2149&field3=buildstamp&compare3=61&value3=20190726-1702-Experimental). So, I would like to move forward in some way on this one.
The remaining -Werror=bool-compare issues exist because there are segments of code that are unreachable behind if statements that will always be false. I would prefer not to guess as to the original intent. So, the most straightforward solution is to remove the problem segments. If that code needs to be there, it can always be added back later when the problem sections are corrected, but they haven't been running for some time now.
That segment removal is the difference in the latest push to Werror_Anasazi.
Of course this is still up for review, but I hope that is acceptable. I have verified that these changes do [build and pass all tests](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=3&showfilters=1&filtercombine=and&field1=site&compare1=61&value1=ascic158&field2=buildname&compare2=61&value2=PR-Werror_Anasazi_remove_segments-test-Trilinos_pullrequest_gcc_7.2.0-2150&field3=buildstamp&compare3=61&value3=20190726-2031-Experimental).

packages/anasazi/epetra/test/OrthoManager/cxx_gentest.cpp
```diff
@@ -733,72 +725,6 @@ int testProjectAndNormalizeGen(RCP<GenOrthoManager<ST,MV,OP> > OM,
...
-        // flip the inputs back
-        theX = tuple( theX[1], theX[0] );
-        theY = tuple( theY[1], theY[0] );
-        theMX = tuple( theMX[1], theMX[0] );
-        theMY = tuple( theMY[1], theMY[0] );
-      }
-
-
       // test all outputs for correctness
       for (unsigned int o=0; o<S_outs.size(); o++) {
```
mhoemmen: Proper type should be `size_t` here, but we can always fix that later.

Issue #4473, opened 2019-02-22
Belos: Remove build warnings


ZUUL42:
@trilinos/belos  

## Expectations

Belos builds should not emit any warnings that will be promoted to errors once Werror is set in the GCC 7.2.0 automated build.

## Current Behavior

Currently Belos has a number of warnings that need to be handled before we can set Werror for all packages.
A recent test build was performed with -Werror set for Belos. The CDash report can be found [here](https://testing.sandia.gov/cdash/viewBuildError.php?buildid=4589921).

## Motivation, Context and Related Issues

Issue #3178 is working toward turning Warnings as Errors on for all packages.

srajama1:
@trilinos/belos @hkthorn

ZUUL42:
If you'd like the PR to test with Werror set for Belos, pull the commit from branch [Werror_Belos](https://github.com/ZUUL42/Trilinos/tree/Werror_Belos).

PR #5218, opened 2019-05-21
Fix warnings and set Werror for Belos in GCC 7.2.0 build

ZUUL42:
@trilinos/framework 
@trilinos/belos 

## Description

Fixed warnings and set Werror for Belos in the GC 7.2.0 PR build.
Wrapped variables only used by asserts in #ifndef NDEBUG Werror sees as unused.
## Motivation and Context

This completes #4473 and is work toward Issue #3178, eventually setting -Werror for all packages.

## How Has This Been Tested?

This was tested in debug and release Jenkins builds.

## Checklist


packages/belos/epetra/example/GCRODR/PrecGCRODREpetraExFile.cpp
```diff
@@ -154,12 +154,18 @@ bool proc_verbose = false;
   Teuchos::RCP<Ifpack_CrsRiluk> ilukFactors;
   //
   ilukGraph = Teuchos::rcp(new Ifpack_IlukGraph(A->Graph(), Lfill, Overlap));
-  int info = ilukGraph->ConstructFilledGraph();
+#ifndef NDEBUG
```
mhoemmen: Belos has a required dependence on Teuchos, so you could do this:
```c++
#include "Teuchos_Assert.hpp"
// ...
int info = ilukGraph->ConstructFilledGraph();
TEUCHOS_ASSERT( info == 0 );
```
This approach would make the most sense for tests and examples, where we always want to check the assertion.  ("Tests" that only use `assert` aren't actually testing those assertions in any PR test right now.)

ZUUL42: OK. I'll switch them over.
"tests and examples"? Would using `TEUCHOS_ASSERT` also be preferred for `packages/belos/epetra/src/BelosEpetraUtils.cpp`?
Or shall I just leave that one be with the assert & macro lines?

mhoemmen: @ZUUL42 Thanks!  "Tests and examples" means "not library source code."  `TEUCHOS_ASSERT` throws an exception; that's different than what `assert` does, and we don't want to change the library's behavior.

> Would using `TEUCHOS_ASSERT` also be preferred for `packages/belos/epetra/src/BelosEpetraUtils.cpp`?

That's library source code; it would be best to leave that alone.

ZUUL42:
No problem. Thanks for the comments. Since it's not my code, I want to make sure to fix it not just to get my Werror in, but the best manner possible since I'm not familiar with the packages.

PR #5231, opened 2019-05-22
Ifpack2: Fixing compiler error w/ Cuda 10.1

csiefer2:
Auto-PR for SHA 5eba430

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_CrsArrays.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_CrsArrays.hpp
@@ -125,7 +125,6 @@ struct CrsArrayReader
     KOKKOS_INLINE_FUNCTION void operator()(const int row) const
     {
       LocalOrdinal offset = rowptrs(row);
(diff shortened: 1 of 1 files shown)


packages/ifpack2/src/Ifpack2_Details_CrsArrays.hpp
```diff
@@ -125,7 +125,6 @@ struct CrsArrayReader
     KOKKOS_INLINE_FUNCTION void operator()(const int row) const
     {
       LocalOrdinal offset = rowptrs(row);
-      std::cout << "Rowptr[" << row << "] = " << offset << '\n';
       size_t entries = rowptrs(row + 1) - offset;
       Teuchos::Array<Scalar> valsArray(entries);
```
mhoemmen: * Don't use `Teuchos::Array` in a device function.
* If you won't ever run this on CUDA, then don't use the `KOKKOS_INLINE_FUNCTION` macro -- that will avoid build errors and warnings.

PR #5246, opened 2019-05-24
Remove mandatory dependence on ParMETIS from stk-balance.

alanw0:
It should be an optional dependence, but it should be an optional
runtime dependence. Since we get the ParMETIS dependency transitively
through Zoltan2, we don't need to mention it at all in stk-balance
cmake files, but rather selectively enable/disable tests that need it.

Also, when a user runs the stk_balance executable it should produce
a semi-graceful message if they try to use a feature that needs
ParMETIS when ParMETIS hasn't been built into Zoltan2.

@trilinos/<teamName>

## Checklist


alanw0:
Thanks @mhoemmen , please do approve if you don't mind. Getting the cmake dependencies right will take a little longer.

alanw0:
@mhoemmen I just added one more commit in response to a nalu-wind build error that was caused by a bad stk_transfer API deprecation. I'll set the retest label and let it run through the builds again.

PR #5253, opened 2019-05-25
Ifpack2::Chebyshev: Fix #5249 (fuse kernels)

### LINKED ISSUE #5249: Ifpack2::Chebyshev: Fuse "scaled damped residual" kernel
## Enhancement
@trilinos/ifpack2 @csiefer2 @jhux2 

Ifpack2::Chebyshev depends on the following computational kernels:

  * First iteration:
    * (Untested case) either `W := (1/theta)*D_inv*B` and `X := 0 + W`; or,
    * (Tested case) if the initial guess is assumed nonzero, `W := (1/theta)*D_inv*(B-A*X)` and `X := X + W`.
  * Subsequent iterations: `W := alpha*D_inv*(B-A*X) + beta*W` and `X := X + W`.

Here, `D_inv` is the inverse of a diagonal matrix, stored as a `Tpetra::Vector`.  This makes the following operation a candidate for kernel fusion:

`W := alpha*D_inv*(B-A*X) + beta*W`

I call this kernel "compute a scaled damped residual vector."  Note that we can't fuse this with a subsequent `X := X + W`, because X as output has data-dependent loop dependencies with X as input (in the sparse matrix-vector multiply `A*X`).

Right now, Ifpack2 issues the following kernels:
  1. `Tpetra::deep_copy (V1, B);`
  2. `A.apply (X, V1, Teuchos::NO_TRANS, -1.0, 1.0);`
  3. `W.elementWiseMultiply (alpha, D_inv, V1, beta);`

If the matrix A has no Export object, then we can fuse all these kernels into one.  This will make each iteration take only two kernels instead of four, not counting pack and unpack kernels in the sparse matrix-vector multiply.  In addition, we can get rid of the temporary vector V1.

mhoemmen:
@trilinos/ifpack2 

## Description

1. Introduce a new "scaled damped residual" vector kernel.
2. Use this new kernel in `Ifpack2::Chebyshev`.

## Motivation and Context

* Reduce the number of parallel kernel launches
* Avoid extra passes over vectors

## Related Issues

* Closes #5249

### DIFF
--- a/packages/ifpack2/src/CMakeLists.txt
+++ b/packages/ifpack2/src/CMakeLists.txt
@@ -368,4 +368,5 @@ TRIBITS_ADD_LIBRARY(
 #
 # Make a trivial change here if you want CMake to run, due to changes
 # you make to files in Ifpack2.  Here is another such change.
+# Behold, I make another change.
 #
--- a/packages/ifpack2/src/CMakeLists.txt
+++ b/packages/ifpack2/src/CMakeLists.txt
@@ -225,6 +225,7 @@ IF(Ifpack2_ENABLE_EXPLICIT_INSTANTIATION)
     Details::Factory
     Details::LinearSolver
     Details::OneLevelFactory
+    Details::ScaledDampedResidual
     Details::TriDiSolver
     Experimental::RBILUK
     )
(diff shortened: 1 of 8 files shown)


mhoemmen:
Warnings as errors build (GCC 7.2.0) strikes again:
```
.../packages/ifpack2/src/Ifpack2_Details_ScaledDampedResidual_def.hpp:359:16: error: unused variable ‘numVecs’ [-Werror=unused-variable]
   const size_t numVecs = B.getNumVectors ();
                ^~~~~~~
```

mhoemmen:
@trilinos/ifpack2 @trilinos/muelu I fixed the build warning; please feel free to review.  Thanks!

mhoemmen:
@trilinos/tpetra

PR #5255, opened 2019-05-25
Fix post-push CI server to enable everything when global file is updated

bartlettroscoe:
This will allow the Trilinos post-push CI server to correct run a CI rebuild iteration when all PT packages need to be enabled.  This will give better statistics for rebuilds of Trilinos and timings to compare to the Trilinos PR builds.  For details, see tribitspub/tribits@35510d754843752c4bd4da3a4dffdfd54a0ff389.

There are excellent automated tests for other supported use cases.

This also brings in changes from TriBITS for the `cdash_analyze_and_report.py` tool (see tribitspub/tribits#288).


cmake/tribits/ci_support/CDashQueryAnalyzeReport.py
```diff
@@ -1054,15 +1055,20 @@ def dateFromBuildStartTime(buildStartTime):
 #
 # Inputs:
 #
-#   testHistoryLOD [in]: List of test dicts for the same test.  This list nore
-#   its elements are modified in this call. (The base list object is shallow
-#   copied before it is sorted.)
+#   testHistoryLOD [in]: List of test dicts for the same test.  This input
```
mhoemmen: "Neither this input list nor its elements are modified..."

bartlettroscoe:
Thanks @mhoemmen!

cmake/tribits/ci_support/CDashQueryAnalyzeReport.py
bartlettroscoe: Thanks @mhoemmen!  I fixed this in https://github.com/TriBITSPub/TriBITS/commit/af11d587ea4abfefbfc55635ce78b9cbda252f50.

Issue #5251, opened 2019-05-24
Xpetra, MueLu: CUDA compile error when Epetra is disabled


jhux2:
When MueLu and Xpetra are compiled with CUDA and Epetra disabled, the build fails.
```
/data/jhu/trilinos-test/OPENMPI-1.8.7_RELEASE_AMGX_CUDA-8.0.44/Trilinos/packages/xpetra/src/Operator/Xpetra_TpetraOperator.hpp(201): error: identifier "EpetraNode" is undefined
```

mhoemmen:
`OPENMPI-1.8.7_RELEASE` -- if the name is accurate, that's a super old version of OpenMPI.  It might even have `MPI_VERSION == 2`, though I'm not sure.

jhux2:
@mhoemmen This configuration uses the module  `kokkos-openmpi/1.8.7/cuda`.  Would you recommend `kokkos-openmpi/2.0.1/cuda` instead?

mhoemmen:
@jhux2 wrote:
> This configuration uses the module `kokkos-openmpi/1.8.7/cuda`. Would you recommend `kokkos-openmpi/2.0.1/cuda` instead?

I'm just saying that I don't know any customers who use an OpenMPI version that old.  I think people have been using 2.0.1, but since this is a specialized build anyway, I would suggest trying OpenMPI 4.x if available.

jhux2:
> I'm just saying that I don't know any customers who use an OpenMPI version that old. I think people have been using 2.0.1, but since this is a specialized build anyway, I would suggest trying as new of a version of OpenMPI as possible.

Thanks for the suggestion.  The compile works locally for me, so I'm moving geminga CUDA to OpenMPI 2.0.1.

jhux2:
Fixed.

PR #5266, opened 2019-05-28
MueLu: testing: update openmpi on geminga

jhux2:
Update OpenMPI version on geminga. See #5251.

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/5266)


jhux2:
Argh, is the PR broken again?  The testing for this PR should be a no-op.

Issue #5065, opened 2019-05-01
Tpetra:  Deprecated DynamicProfile used without deprecation guards


kddevin:
## Bug Report
@trilinos/tpetra 

### Description
DynamicProfile is a deprecated feature.
It is used in Tpetra_CrsMatrix_decl.hpp without deprecation guards.
https://github.com/trilinos/Trilinos/blob/ec7ddad27cbfca383fb4c70f8beeba6193739853/packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp#L5039

Thus, building with TPETRA_ENABLE_DEPRECATED_CODE=OFF fails.

Related to #4701

### Steps to Reproduce

Build Tpetra with TPETRA_ENABLE_DEPRECATED_CODE=OFF.
I am building with many other packages; I don't know if they are needed to see the error.
```
-D Tpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
-D Teuchos_HIDE_DEPRECATED_CODE:BOOL=ON \
\
-D Trilinos_ENABLE_Stokhos:BOOL=ON \
-D Trilinos_ENABLE_Nox:BOOL=ON \
-D Trilinos_ENABLE_ROL:BOOL=ON \
-D ROL_ENABLE_EXAMPLES:BOOL=OFF \
-D Trilinos_ENABLE_MiniTensor:BOOL=OFF \
-D ROL_ENABLE_MiniTensor:BOOL=OFF \
-D Trilinos_ENABLE_Panzer:BOOL=ON \
-D Trilinos_ENABLE_PanzerAdaptersSTK:BOOL=OFF \
-D Trilinos_ENABLE_PanzerAdaptersIOSS:BOOL=OFF \
-D Trilinos_ENABLE_Thyra:BOOL=ON \
-D Trilinos_ENABLE_MueLu:BOOL=ON \
-D Trilinos_ENABLE_Anasazi:BOOL=ON \
-D Trilinos_ENABLE_Belos:BOOL=ON \
-D Trilinos_ENABLE_TrilinosCouplings:BOOL=ON \
-D Trilinos_ENABLE_STKIO:BOOL=OFF \
-D Trilinos_ENABLE_STKUtil:BOOL=OFF \
```

kddevin:
This one may be difficult to fix, as it is used without max number of entries per row throughout the Trilinos stack.  We may need help from other package developers to fix it correctly.

[...]

PR #5268, opened 2019-05-28
Tpetra: Allow CrsMatrix with StaticProfile to resize during import/export

### LINKED ISSUE #5164: Tpetra: Allow CrsMatrix import/export to change its graph's structure
## Enhancement
@trilinos/tpetra 

PR #4147 allows import/export operations to  change the structure of a `CrsGraph`.  Importantly, the user cannot change the structure directly, the structure changes happen in internal `CrsGraph` methods.

@kddevin and @mhoemmen should `CrsMatrix` provide the same functionality?

The answer to whether or not `CrsMatrix` does provide the functionality will have implications on many tests that fail with deprecated code OFF.

tjfulle:
@trilinos/tpetra 

## Description
Allow CrsMatrix with StaticProfile to resize during import/export

## Motivation and Context
See #5164 

## Related Issues

* Closes #5164 

## How Has This Been Tested?
Tpetra tests pass locally

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -5845,7 +5845,7 @@ namespace Tpetra {
         this->getGlobalRowCopy(tgt_gid, tgt_row_inds(), check_row_length);
 
         size_t how_much_padding = 0;
-        for (const auto& src_row_ind : src_row_inds) {
+        for (auto src_row_ind : src_row_inds) {
           if (std::find(tgt_row_inds.begin(), tgt_row_inds.end(), src_row_ind) == tgt_row_inds.end()) {
             // The target row does not have space for
             how_much_padding++;
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -5792,8 +5792,7 @@ namespace Tpetra {
     computeCrsPaddingForSameIDs(padding, source, numSameIDs, false);
     computeCrsPaddingForPermutedIDs(padding, source, permuteToLIDs, permuteFromLIDs, false);
 
-    using execution_space = typename device_type::execution_space;
-    execution_space::fence ();  // Make sure device sees changes made by host
+    Kokkos::fence ();  // Make sure device sees changes made by host
     TEUCHOS_TEST_FOR_EXCEPTION
       (padding.failed_insert(), std::runtime_error,
        "failed to insert one or more indices in to padding map");
@@ -5811,17 +5810,14 @@ namespace Tpetra {
   {
     using LO = LocalOrdinal;
     using GO = GlobalOrdinal;
-    using execution_space = typename device_type::execution_space;
     const char tfecfFuncName[] = "computeCrsPaddingForSameIds: ";
 
-    execution_space::fence ();
+    Kokkos::fence ();
 
     using insert_result =
       typename Kokkos::UnorderedMap<LocalOrdinal, size_t, typename Node::device_type>::insert_result;
 
     // Compute extra capacity needed to accommodate incoming data
-    Teuchos::Array<GO> src_row_inds;
-    Teuchos::Array<GO> tgt_row_inds;
     const map_type& src_row_map = * (source.getRowMap ());
     for (LO tgt_lid = 0; tgt_lid < static_cast<LO> (numSameIDs); ++tgt_lid) {
       const GO src_gid = src_row_map.getGlobalElement(tgt_lid);
@@ -5837,12 +5833,14 @@ namespace Tpetra {
       }
       else {
         size_t check_row_length = 0;
-        src_row_inds.resize(num_src_entries);
-        source.getGlobalRowCopy(src_gid, src_row_inds(), check_row_length);
+        std::vector<GO> src_row_inds(num_src_entries);
+        Teuchos::ArrayView<GO> src_row_inds_view(src_row_inds.data(), src_row_inds.size());
+        source.getGlobalRowCopy(src_gid, src_row_inds_view, check_row_length);
 
         auto num_tgt_entries = this->getNumEntriesInGlobalRow(tgt_gid);
-        tgt_row_inds.resize(num_tgt_entries);
-        this->getGlobalRowCopy(tgt_gid, tgt_row_inds(), check_row_length);
+        std::vector<GO> tgt_row_inds(num_tgt_entries);
+        Teuchos::ArrayView<GO> tgt_row_inds_view(tgt_row_inds.data(), tgt_row_inds.size());
+        this->getGlobalRowCopy(tgt_gid, tgt_row_inds_view, check_row_length);
 
         size_t how_much_padding = 0;
         for (auto src_row_ind : src_row_inds) {
@@ -5874,12 +5872,9 @@ namespace Tpetra {
   {
     using LO = LocalOrdinal;
     using GO = GlobalOrdinal;
-    using execution_space = typename device_type::execution_space;
     const char tfecfFuncName[] = "computeCrsPaddingForPermutedIds: ";
-    execution_space::fence ();
+    Kokkos::fence ();
 
-    Teuchos::Array<GO> src_row_inds;
-    Teuchos::Array<GO> tgt_row_inds;
     const map_type& src_row_map = * (source.getRowMap ());
 
     using insert_result =
@@ -5901,16 +5896,18 @@ namespace Tpetra {
       }
       else {
         size_t check_row_length = 0;
-        src_row_inds.resize(num_src_entries);
-        source.getGlobalRowCopy(src_gid, src_row_inds(), check_row_length);
+        std::vector<GO> src_row_inds(num_src_entries);
+        Teuchos::ArrayView<GO> src_row_inds_view(src_row_inds.data(), src_row_inds.size());
+        source.getGlobalRowCopy(src_gid, src_row_inds_view, check_row_length);
 
         const GO tgt_gid = rowMap_->getGlobalElement (tgt_lid);
         auto num_tgt_entries = this->getNumEntriesInGlobalRow(tgt_gid);
-        tgt_row_inds.resize(num_tgt_entries);
-        this->getGlobalRowCopy(tgt_gid, tgt_row_inds(), check_row_length);
+        std::vector<GO> tgt_row_inds(num_tgt_entries);
+        Teuchos::ArrayView<GO> tgt_row_inds_view(tgt_row_inds.data(), tgt_row_inds.size());
+        this->getGlobalRowCopy(tgt_gid, tgt_row_inds_view, check_row_length);
 
         size_t how_much_padding = 0;
-        for (const auto& src_row_ind : src_row_inds) {
+        for (auto src_row_ind : src_row_inds) {
           if (std::find(tgt_row_inds.begin(), tgt_row_inds.end(), src_row_ind) == tgt_row_inds.end()) {
             // The target row does not have space for
             how_much_padding++;
@@ -5936,11 +5933,10 @@ namespace Tpetra {
                      buffer_device_type>& importLIDs,
                      Kokkos::DualView<size_t*, buffer_device_type> numPacketsPerLID) const
   {
-    using execution_space = typename device_type::execution_space;
     const char tfecfFuncName[] = "computeCrsPadding: ";
 
     // Creating padding for each new incoming index
-    execution_space::fence ();  // Make sure device sees changes made by host
+    Kokkos::fence ();  // Make sure device sees changes made by host
     using padding_type = Kokkos::UnorderedMap<local_ordinal_type, size_t, device_type>;
     padding_type padding (importLIDs.extent (0));
     auto numEnt = static_cast<size_t> (importLIDs.extent (0));
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -5609,9 +5609,9 @@ namespace Tpetra {
         os << *prefix << "Target is StaticProfile; do CRS padding" << endl;
         std::cerr << os.str ();
       }
-      auto padding = computeCrsPaddingNew (srcRowGraph, numSameIDs,
-                                           permuteToLIDs, permuteFromLIDs);
-      this->applyCrsPadding (padding);
+      auto padding =
+        computeCrsPadding (srcRowGraph, numSameIDs, permuteToLIDs, permuteFromLIDs);
+      this->applyCrsPadding(padding);
     }
     else if (debug) {
       std::ostringstream os;
@@ -5712,6 +5712,9 @@ namespace Tpetra {
     using range_policy = Kokkos::RangePolicy<execution_space, Kokkos::IndexType<LocalOrdinal>>;
     using Tpetra::Details::padCrsArrays;
 
+    if (padding.size() == 0)
+      return;
+
     // Assume global indexing we don't have any indices yet
     if (! this->indicesAreAllocated()) {
       allocateIndices(GlobalIndices);
@@ -5724,7 +5727,7 @@ namespace Tpetra {
     Kokkos::deep_copy(row_ptrs_beg, this->k_rowPtrs_);
 
     const size_t N = (row_ptrs_beg.extent(0) == 0 ? 0 : row_ptrs_beg.extent(0) - 1);
-      row_ptrs_type row_ptrs_end("row_ptrs_end", N);
+    row_ptrs_type row_ptrs_end("row_ptrs_end", N);
 
     bool refill_num_row_entries = false;
     if (this->k_numRowEntries_.extent(0) > 0) {
@@ -5763,7 +5766,6 @@ namespace Tpetra {
       this->k_lclInds1D_ = indices;
     }
 
-
     if (refill_num_row_entries) {
       auto num_row_entries = this->k_numRowEntries_;
       Kokkos::parallel_for("Fill num entries", range_policy(0, N),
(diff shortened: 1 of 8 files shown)


packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5609,9 +5609,9 @@ namespace Tpetra {
         os << *prefix << "Target is StaticProfile; do CRS padding" << endl;
         std::cerr << os.str ();
       }
-      auto padding = computeCrsPaddingNew (srcRowGraph, numSameIDs,
-                                           permuteToLIDs, permuteFromLIDs);
-      this->applyCrsPadding (padding);
+      auto padding =
```
mhoemmen: This is fine for now, but in the future, we should have a special case for when the target has a const graph.  If so, then we don't need to compute padding at all, since the target's structure can't change.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5724,7 +5727,7 @@ namespace Tpetra {
     Kokkos::deep_copy(row_ptrs_beg, this->k_rowPtrs_);
 
     const size_t N = (row_ptrs_beg.extent(0) == 0 ? 0 : row_ptrs_beg.extent(0) - 1);
-      row_ptrs_type row_ptrs_end("row_ptrs_end", N);
+    row_ptrs_type row_ptrs_end("row_ptrs_end", N);
```
mhoemmen: In the future, let's allocate this Kokkos::View without initializing, if that makes sense.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5779,137 +5781,167 @@ namespace Tpetra {
   Kokkos::UnorderedMap<LocalOrdinal, size_t, typename Node::device_type>
```
mhoemmen: In the future, when we get rid of template parameters, we'll want to get `device_type` from RowGraph, not from Node.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5779,137 +5781,167 @@ namespace Tpetra {
...
+      const GO src_gid = src_row_map.getGlobalElement(tgt_lid);
+      auto num_src_entries = source.getNumEntriesInGlobalRow(src_gid);
+
+      if (num_src_entries == 0)
+        continue;
+
+      insert_result result;
+      const GO tgt_gid = rowMap_->getGlobalElement(tgt_lid);
+      if (padAll) {
+        result = padding.insert(tgt_lid, num_src_entries);
```
mhoemmen: Right now, we just want things to compile and pass tests.  All the following comments are for the future.

I think you could use two 1-D `Kokkos::View`s to store padding information.  There are actually two sets of data over which to compute padding: sames, and permutes.  There's no reason why you need to store both in a single data structure, as long as the client can map from local row index to required padding.  The first data structure, for sames, could just be a 1-D `Kokkos::View`, since it's just indexed by `0 .. numSameIDs - 1`.  That would let you use a much smaller hash table for the second data structure.  Alternately, you could index the "permutes" padding data structure by the k in `permuteFromLIDs[k]`.  That would let you use a 1-D `Kokkos::View` as well.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5779,137 +5781,167 @@ namespace Tpetra {
...
+        size_t check_row_length = 0;
+        src_row_inds.resize(num_src_entries);
+        source.getGlobalRowCopy(src_gid, src_row_inds(), check_row_length);
+
+        auto num_tgt_entries = this->getNumEntriesInGlobalRow(tgt_gid);
+        tgt_row_inds.resize(num_tgt_entries);
+        this->getGlobalRowCopy(tgt_gid, tgt_row_inds(), check_row_length);
+
+        size_t how_much_padding = 0;
+        for (const auto& src_row_ind : src_row_inds) {
```
mhoemmen: References are bigger than values in this case, so you might as well just use `auto` instead of `const auto&` here.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -2434,6 +2436,7 @@ namespace Tpetra {
           // curOffset, the range is empty.
           const LO numIndInSeq = (endOffset - curOffset);
           if (numIndInSeq != 0) {
+            rowInfo = graph.getRowInfo(lclRow);  // KDD 5/19 Need fresh RowInfo in each loop iteration
```
mhoemmen: Not really.  `insertGlobalValuesImpl` actually updates its `rowInfo` argument (it's a read + write argument).  It should have the correct values on exit.  Calling `getRowInfo` can be a bit expensive, which is why the `insert*ValuesImpl` and other methods separate that out.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -6615,6 +6618,98 @@ namespace Tpetra {
...
+      refill_num_row_entries = true;
+      auto num_row_entries = myGraph_->k_numRowEntries_;
+      Kokkos::parallel_for("Fill end row pointers", range_policy(0, N),
+                           KOKKOS_LAMBDA(const size_t i){
+                             row_ptr_end(i) = row_ptr_beg(i) + num_row_entries(i);
+                           }
+                           );
+
+    } else {
+      // mfh If packed storage, don't need row_ptr_end to be separate allocation;
```
mhoemmen: What about this comment?

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -6615,6 +6618,98 @@ namespace Tpetra {
...
+                           );
+    }
+
+    using values_type = typename local_matrix_type::values_type;
+    values_type values("values", k_values1D_.size());
+    Kokkos::deep_copy(values, k_values1D_);
+
+    if(myGraph_->isGloballyIndexed()) {
+      using indices_type = typename crs_graph_type::t_GlobalOrdinal_1D;
+      indices_type indices("indices", myGraph_->k_gblInds1D_.extent(0));
```
mhoemmen: Do a view allocation without initializing here.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -6615,6 +6618,98 @@ namespace Tpetra {
...
+      padCrsArrays<row_ptrs_type,indices_type,values_type,padding_type>(row_ptr_beg, row_ptr_end, indices, values, padding);
+      myGraph_->k_gblInds1D_ = indices;
+      TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(
+          values.size() != indices.size(),
+          std::logic_error,
+          "After padding, values and indices should be same size");
+    }
+    else {
+      using indices_type = typename local_graph_type::entries_type::non_const_type;
+      indices_type indices("indices", myGraph_->k_lclInds1D_.extent(0));
```
mhoemmen: Do a view allocation without initializing here.

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -106,10 +109,13 @@ pad_crs_arrays(
...
     }, additional_size_needed);
 
-  if (additional_size_needed == 0) return;
+  if (additional_size_needed == 0)
+    return;
 
   // The indices array must be resized and the row_ptr arrays shuffled
   auto indices_new = uninitialized_view<Indices>("ind new", indices.size()+additional_size_needed);
+  auto values_new = uninitialized_view<Values>("val new", pad_values ? values.size()+additional_size_needed : 0);
+  Kokkos::deep_copy(values_new, Teuchos::ScalarTraits<typename Values::value_type>::zero());
```
mhoemmen: `Kokkos::View` allocation initializes with zero by default; you could save a line of code here by relying on that.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_StaticImportExport.cpp
```diff
@@ -0,0 +1,232 @@
...
+  auto A = rcp(new matrix_type(map, 3, Tpetra::StaticProfile));
+
+  // const size_t numMyElements = map->getNodeNumElements ();
+  // The list of global elements owned by this MPI process.
+  const Tpetra::global_size_t num_gbl_inds = map->getGlobalNumElements();
+  auto my_gbl_elems = map->getNodeElementList();
+  for (auto it : my_gbl_elems) {
+    const auto gbl_row = map->getGlobalElement(it);
+    if (gbl_row == 0)
+      A->insertGlobalValues(gbl_row, tuple(gbl_row, gbl_row+1), tuple(two, neg_one));
```
mhoemmen: Everybody's favorite matrix ;-P

packages/tpetra/core/test/CrsMatrix/CrsMatrix_StaticImportExport.cpp
```diff
@@ -0,0 +1,232 @@
...
+  }
+  B->fillComplete();
+
+  auto loc_num_errs = 0;
+
+  // The test below uses the host Tpetra::CrsMatrix interface to
+  // compare matrix values.  Thus, we need to do a fence before
+  // comparing matrix values, in order to ensure that changes made on
+  // device are visible on host.
+  execution_space::fence();
```
mhoemmen: `fence()` as a static method has been deprecated in Kokkos.  Do this instead:
```c++
execution_space ().fence ();
```

packages/tpetra/core/test/CrsMatrix/Tpetra_Test_CrsMatrix_WithGraph.hpp
```diff
@@ -890,21 +890,6 @@ inline void tupleToArray(Array<T> &arr, const tuple &tup)
       GO r = map->getMinGlobalIndex();
       TEST_THROW( matrix.insertGlobalValues( r, tuple(r+1), tuple(ST::one()) ), std::runtime_error );
     }
-    if (numImages > 1) {
```
mhoemmen: Do we still want this test here?

kddevin:
@tjfulle Does Tpetra test TpetraCore_CrsMatrix_NonlocalAfterResume pass for you with TPETRA_ENABLE_DEPRECATED_CODE=OFF?  With my branch and this PR merged into it, this test still hangs for me.  All other Tpetra tests are fine.  Some Teko tests still fail; I will look at them.

mhoemmen:
@kddevin Don't forget to avail yourself of the `TPETRA_VERBOSE=1` environment variable setting.

kddevin:
Good suggestion, @mhoemmen .  Thanks.  Stay tuned.

tjfulle:
@mhoemmen, as you noted, this is a stop gap for getting dynamic profile deprecated.  per our previous discussions, I think this whole area of Tpetra could use some serious clean up.

@kddevin, this PR is only to add matrix resizing during import/export.  I suspect this will help with #5065, but probably is not a complete fix.  Let’s chat tomorrow about syncing our efforts to finish #5065

tjfulle:
Arg - failed to build with complex - should be an easy fix.  I'll look at the other test failures too.

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -112,37 +112,39 @@ pad_crs_arrays(
...
 
+  using ptrs_value_type = typename RowPtr::non_const_value_type;
+  using inds_value_type = typename Indices::non_const_value_type;
+  using vals_value_type = typename Values::non_const_value_type;
+
   // The indices array must be resized and the row_ptr arrays shuffled
   auto indices_new = uninitialized_view<Indices>("ind new", indices.size()+additional_size_needed);
   auto values_new = uninitialized_view<Values>("val new", pad_values ? values.size()+additional_size_needed : 0);
-  Kokkos::deep_copy(values_new, Teuchos::ScalarTraits<typename Values::value_type>::zero());
+  Kokkos::deep_copy(values_new, vals_value_type(0.0));
```
mhoemmen: You could also use `Kokkos::ArithTraits<vals_value_type>::zero()`, though this is fine (even for integer or complex types).

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -112,37 +112,39 @@ pad_crs_arrays(
...
 
   // mfh: Not so fussy about this not being a kernel initially,
   // since we're adding a new feature upon which existing code does not rely,
   // namely Export/Import to a StaticProfile CrsGraph.  However, watch out
   // for fence()ing relating to UVM.
   auto this_row_beg = row_ptr_beg(0);
   auto this_row_end = row_ptr_end(0);
```
mhoemmen: For the future: `Kokkos::deep_copy` now supports copying from a 0-D device View to a value in host memory.  Thus, you could do this:
```c++
ptrs_value_type this_row_beg;
Kokkos::deep_copy (this_row_beg, Kokkos::subview (row_ptr_beg, 0));
```

mhoemmen:
@micahahoward FYI, this is the complicated issue I was talking about that we wanted to finish before trying host-pinned buffers (getting that to compile required touching all of this code which is currently in flux and really needs to be right).

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
tjfulle: I still have some work to do to build/pass tests on Cuda.  Is this preferred?  If so, I'll make the change now.

mhoemmen: @tjfulle I wouldn't worry about this for now.  You're accessing every row, so it will likely be faster just to let UVM do its thing.  I would prefer that we spent future effort on making this into a parallel kernel on device.  Thanks though!

tjfulle:
Hmmm, the PR tester does not seem to be building the most recent commit.

tjfulle:
Hmmm, passes on all platforms but Linux+intel17.  These 4 Zoltan2 tests fail with that combination:
- `Zoltan2_TpetraRowGraphInput_MPI_4`
- `Zoltan2_TpetraRowMatrixInput_MPI_4`
- `Zoltan2_XpetraCrsGraphInput_MPI_4`
- `Zoltan2_XpetraTraits_MPI_4`

They all report “PASS”, but dangling RCPs cause a test failure.

@kddevin, do these sound familiar?  Any ideas?  If not, I can get a debut build going with intel to try and track it down.

kddevin:
@tjfulle 
I have not seen failures in these Zoltan2 tests before.  
All of these tests do roughly the following:
- read a very simple matrix from a file onto four processors (using Tpetra's matrix-market reader)
- collect the matrix or graph onto one processor using Tpetra's Import.  I believe they call importAndFillComplete for the CrsMatrix (which is not reporting problems) and doImport for the CrsGraph (which is reporting problems).  I'd have to look a little more closely to see how we're testing RowMatrix.  

I'd start with XpetraCrsGraphInput.cpp.  If we fix one, we'll likely fix them all.  

Unfortunately, I will not have time to look at it today, but I am happy to work on it tomorrow if it can wait until then.

tjfulle:
@kddevin I'll take a look to see if I see anything.  If I run in to problems, I'll let you know.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
tjfulle: yep, when this is done, I have some ideas along the lines you mention above to get rid of `UnorderedMap` in favor of a `Kokkos::View`

tjfulle:
The error on intel comes down to this function: https://github.com/trilinos/Trilinos/pull/5268/files#diff-90fa5171ebff6b0e9646050da6ce8bd3R5869

I'll take a detailed look at how this function differs from similar code in the master branch to see why it errors and master doesn't.  `valgrind` got me in the ballpark, hopefully `idb` can get me the rest of the way.

tjfulle:
@mhoemmen, do you see anything obvious in `CrsGraph::computeCrsPaddingForPermutedIDs` that would lead to dangling `RCP`?

mhoemmen:
@tjfulle Nothing looks obviously wrong in this function.  You could add a debug-mode all-reduce check in DistObject to see whether all the MPI processes actually make it through the routine.  It's entirely possible that the Zoltan2 tests in question only check if Process 0 made it through.  You could also check whether the input Map / Comm is null; I'm not sure whether that could happen due to subcommunicator stuff.

tjfulle:
Ughhh, the sems installation of mpich loaded with the Intel17 pull request environment does not have debug support (options `-tv` or `-debug` not recognized).  I'll have to rebuild with openmpi (which was built with debug support)

mhoemmen:
> Ughhh, the sems installation of mpich loaded with the Intel17 pull request environment does not have debug support (options `-tv` or `-debug` not recognized). 

:-(  Can you get away with not building the TPLs at least?

tjfulle:
Sems has an Intel/openmpi stack, so I can just swapped the mpich and openmpi modules and rebuilt Trilinos.  I’m firing up totalview as we speak.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5837,12 +5833,14 @@ namespace Tpetra {
       }
       else {
         size_t check_row_length = 0;
-        src_row_inds.resize(num_src_entries);
-        source.getGlobalRowCopy(src_gid, src_row_inds(), check_row_length);
+        std::vector<GO> src_row_inds(num_src_entries);
```
mhoemmen: Tim and I talked on the phone about this.  As far as he can tell, some issue with Teuchos memory management classes' debug-mode checking caused the test failures.  That's why he replaced the `Teuchos::Array` with `std::vector` here.

@tjfulle , you could use exactly the same `resize` logic for `std::vector` as for `Teuchos::Array`; it should work fine.  In fact, you don't have to resize the vector if it already has enough space; just resize if you need more space, and always give `Teuchos::ArrayView`'s constructor `num_src_entries` instead of `src_row_inds.size()` as its size.

mhoemmen:
@tjfulle CONGRATS !!! :-D 

As another data point: I've noticed before that Intel 17 is a bit buggy where GCC and Clang are not.

PR #5290, opened 2019-06-01
Fe assembly example

mperego:
@trilinos/intrepid2
@trilinos/panzer
@trilinos/tpetra

## Description
This wants to be an example of assembling an high-order finite element matrix and right-hand-side corresponding to a Poisson problem. The idea is to provide a reasonable implementation that correctly uses tools such as the Intrepid2 orientation, the Tpetra FE CRS matrices and the Panzer Dof Manager. It's mainly targeting people outside the Sandia world that want to try out the Trilinos Discretization tools for FE. It doesn't need to be  the most performant option, but it shouldn't be a dumb implementation either.

I'd ask @kyungjoo-kim and @CamelliaDPG to look at the finite element assembly part and make sure that I'm not doing anything unreasonable. Similarly I'd ask @eric-c-cyr and @rppawlo to have a look at the use of the ConnManager and DofManager, and to @csiefer2 to look at the use of the FE CRS matrices. I'd like in the near future to assemble the fe_matrix on device, but it might be OK for now. Anyway, any suggestion is welcome.
I'm going to show this example at the Euro TUG on June 11, so if you  can give me feedback next week it would be great. Thanks!
Also thanks to @csiefer2 @eric-c-cyr and @kyungjoo-kim for helping already.

I still need to clean this up, add more comments, timings, possibly split this into multiple files, but I think that the main pieces are there and I hope the code is already easy to follow for the developers I mentioned.
Also at the moment I put this example in Panzer Dof Manager to avoid recursive dependences (Intrepid2 depending on Panzer depending on Intrepid2), but the idea is to create a Discretization Tools package that contains examples and tutorials on how to use the different discretization capabilities.

## Checklist


packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.hpp
```diff
@@ -0,0 +1,608 @@
...
+        for(ordinal_type col_idx=0; col_idx<basisCardinality; col_idx++)
+          column_scalar_values[col_idx] = cellMatHost(elem_id, node_id, col_idx);
+
+        fe_matrix->sumIntoLocalValues(local_row_id, column_local_ids, column_scalar_values);
+        rhs->sumIntoLocalValue(column_local_ids[node_id], 0, cellRhsHost(elem_id, node_id));
+      }
+    }
+
+    Tpetra::endFill(*fe_matrix);
+    Tpetra::endFill(*rhs);
```
csiefer2: You can actually do:
Tpetra::endFill(*fe_matrix, *rhs);

packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.hpp
```diff
@@ -0,0 +1,608 @@
...
+      }
+
+      li::getBasisCoeffs(basisCoeffsLI, funAtDofPoints, dofCoeffsPhys);
+    }
+
+    {
+      Tpetra::Vector<scalar_t> sol(domain_map);
+      auto basisCoeffsLIHost = Kokkos::create_mirror_view(basisCoeffsLI);
+      Kokkos::deep_copy(basisCoeffsLIHost,basisCoeffsLI);
+      for(ordinal_type elem_id=0; elem_id<numOwnedElems; elem_id++)
```
rppawlo: another fence

packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.hpp
```diff
@@ -0,0 +1,608 @@
...
+          global_ordinal_t gid = elementGIDs[elmtOffset[node_id]];
+          if(domain_map->isNodeGlobalElement(gid))
+            sol.replaceGlobalValue(gid, basisCoeffsLIHost(elem_id, node_id));
+        }
+      }
+
+      fe_matrix->apply(sol, *rhs, Teuchos::NO_TRANS, -1.0, 1.0);   // b - A x
+    }
+
+    *outStream << "Residual norm : " << rhs->getVector(0)->norm2() <<std::endl;
```
rppawlo: Another quick question - how would this test fail in ctest for an incorrectly assembled matrix/rhs? It looks like the check prints the residual norm to screen instead of checking a tolerance and throwing.

packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.hpp
```diff
@@ -0,0 +1,608 @@
...
+      auto physVertexesHost = Kokkos::create_mirror_view(physVertexes);
+      DynRankView ConstructWithLabel(refVertices, numNodesPerElem, dim);
+      Basis_HGRAD_HEX_C1_FEM<DeviceSpaceType,scalar_t,scalar_t> hexaLinearBasis;
+      hexaLinearBasis.getDofCoords(refVertices);
+      auto refVerticesHost = Kokkos::create_mirror_view(refVertices);   
+      Kokkos::deep_copy(refVerticesHost, refVertices);
+   
+      auto elemTriplet = connManager->getMyElementsTriplet();
+      double h[3] = {hx, hy, hz};
+
```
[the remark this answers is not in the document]
kyungjoo-kim: I am not sure if the fence is indeed required here if kokkos uses the mirror space as host space (which should be). If the device space is cuda or cuda uvm space, the deep copy triggers memory transfer which automatically puts the fence. @crtrott could you confirm this ? Now I am wondering if kokkos deep copy behaves like cuda memcpy between host and device and enforces synchronization or not.

packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.hpp
mperego: @rppawlo yeah, I plan to check if the error is within some tolerance. In principle one can change the order of the basis and number of elements as arguments of the executable. I can add a check only when the parameters are the default ones.

mperego: Thanks @rppawlo I'll add the fences.

packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.cpp
```diff
@@ -52,12 +52,12 @@
 
 int main(int argc, char *argv[]) {
 
-  Kokkos::initialize();
+  Kokkos::initialize(argc, argv);
```
mhoemmen: Try this instead; it means you can get rid of `finalize_all`:
```c++
Kokkos::ScopeGuard kokkosScope (argc, argv);
Teuchos::GlobalMPISession mpiScope (&argc, &argv);
// rest of your code ...
```

mperego:
@rppawlo I updated the branch to make it compatible with recent changes in Tpetra and Panzer.
I also changed the matrix assembly using parallel_for on device (following @csiefer2 example).
I have a couple of questions (@trilinos/tpetra):
In this test we sum into the local matrix one element at a time, instead of adding the element row:
https://github.com/mperego/Trilinos/blob/fe-assembly-example/packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.hpp#L585
How expensive is this? In order to add the element row, we would need to use teams to have scratch memory, right? 
How expensive is to go from the global Id to the local Id as done here?
https://github.com/mperego/Trilinos/blob/fe-assembly-example/packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.hpp#L576
Is it possible to directly compute the local Ids for the column map, without going through the global Ids?

mperego:
@rppawlo The issue was that I put  `#ifdef PANZER_HAVE_EPETRA` before `#include "PanzerDofMgr_config.hpp"`. 
When you have time can you look at the PR?

packages/panzer/dof-mgr/test/fe_assembly/test_fe_assembly_HEX.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "Kokkos_Core.hpp"
+
+
+int main(int argc, char *argv[]) {
+
+  Kokkos::ScopeGuard kokkosScope (argc, argv);
+  Teuchos::GlobalMPISession mpiScope (&argc, &argv);
+  
+  const int r_val = Intrepid2::Test::feAssemblyHex<double,PHX::Device>(argc, argv);
```
rppawlo: Why is this in the intrepid2 namespace? The code and test is in panzer? Could we change this? If this is code written by the user, we usually don't use the panzer namespace so users understand it is not part of the library. I would use something like panzer_test for the namespace.

mperego: good catch! What about Discretization::Example::feAssemblyHex? The idea eventually is to move this in a "Discretization" package.

rppawlo: I like that!

PR #5315, opened 2019-06-04
Tpetra: Fix #4627

### LINKED ISSUE #4627: Tpetra::BlockMultiVector needs a copy constructor
@trilinos/tpetra requested by @fnrizzi

`Tpetra::BlockMultiVector` has no copy constructor.  We should add it (`= default` should suffice), as well as the two-argument deep-copying "copy constructor" that MultiVector has.

mhoemmen:
@trilinos/tpetra @fnrizzi 

## Description

1. Add `= default` copy and move constructors and copy and move
   assignment operators to both BlockMultiVector and BlockVector.

2. Add two-argument "copy constructors" to both BlockMultiVector and
   BlockVector, just like MultiVector and Vector already have, that
   give users the option to deep copy or shallow copy.

3. Add tests for the two-argument "copy constructors."  They pass.

## Motivation and Context

SPARC wants this.

## Related Issues

* Closes #4627

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Experimental_BlockMultiVector_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_Experimental_BlockMultiVector_decl.hpp
@@ -217,6 +217,30 @@ public:
   //! \name Constructors
   //@{
 
+  /// \brief Default constructor.
+  ///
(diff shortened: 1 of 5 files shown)


mhoemmen:
Thanks @csiefer2 ! :D

PR #5326, opened 2019-06-05
Tpetra DynamicProfile deprecation:  some fixes for #5065

kddevin:
@trilinos/tpetra @trilinos/belos @trilinos/xpetra 

## Description
#5065 reports that Tpetra::DynamicProfile was still being used when Tpetra_ENABLE_DEPRECATED_CODE=OFF.
These changes begin to address tests that must have succeeded before with Tpetra_ENABLE_DEPRECATED_CODE=OFF because of this missed use of DynamicProfile.

Note that not all tests in Trilinos will pass yet with Tpetra_ENABLE_DEPRECATED_CODE=OFF.  In particular, refactoring of Ifpack's ILUT and ILUK is needed to support StaticProfile.   Other packages (e.g., MueLu and Teko) will need to be re-examined after Ifpack2 is fixed.

## Motivation and Context
#5065 #5117 

## How Has This Been Tested?
ATDM test scripts with and without Tpetra_ENABLE_DEPRECATED_CODE=OFF on linux workstation
The former passes; the latter compiles.


tjfulle:
Yeah!!  Did #5268 help resolve some of the failing Tpetra tests?

kddevin:
Thanks, @tjfulle 
Without deprecated code, the only Tpetra test failing for me is TpetraCore_CrsMatrix_NonlocalAfterResume_MPI_4
I recall @mhoemmen was seeing issues with the same test.
Belos, Anasazi and Thyra all pass.  I think Ifpack2 is next on my list; I hope that fixing Ifpack2 will also fix Teko and MueLu.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -687,9 +688,18 @@ mult_test_results multiply_reuse_test(
   computedC2->resumeFill();
   Tpetra::MatrixMatrix::Multiply(*As, AT, *Bs, BT, *computedC2, true/*call_FillCompleteOnResult*/);
 
+  // computedC1 = leftScaling * (op(A)*op(B)) * rightScaling
+  RCP<Matrix_t> computedC1 = rcp( new Matrix_t(map, computedC2->getGlobalMaxNumRowEntries()));
```
mhoemmen: I'm not sure how I feel about this.  Now we're tying these tests to a global quantity that could be quite large.  I guess we can always go back and tighten the bounds later.

packages/tpetra/core/test/ImportExport2/ImportExport2_UnitTests.cpp
```diff
@@ -2064,7 +2064,9 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL( Import_Util, UnpackAndCombineWithOwningPIDs,
     os << *prefix << "Build test matrix and source Map" << std::endl;
     std::cerr << os.str ();
   }
+std::cout << Comm->getRank() << " KDDKDD build A" << std::endl;
```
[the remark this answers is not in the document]
kddevin: Doh!  I'll remove that.  Thanks.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
kddevin: Understood.  This max value is not needed if Multiply fillCompletes, but this specific test is for the case without fillComplete.  So unless we do some sort of symbolic phase first to figure out the nonzero pattern, we have to use a max.    Indeed, this case is a bit artificial anyway, as we only have the max available because we've done the multiplication a different way already.  We could deprecate the non-fillCompleted version at some point.

packages/tpetra/core/test/CrsGraph/CrsGraph_UnitTests1.cpp
```diff
@@ -441,6 +442,7 @@ namespace { // (anonymous)
   ////
   TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL( CrsGraph, ActiveFill, LO, GO , Node )
```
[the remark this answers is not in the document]
kddevin: The test exercises the StaticProfile capabilitiy at lines 137-138; I assume @trilinos/muelu would have to use the StaticProfile capability in the future.  These changes remove only the DynamicProfile capability.

packages/teko/src/Tpetra/Teko_InterlacedTpetra.cpp
```diff
@@ -239,12 +239,12 @@ RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > buildSubBlock(int i,int j,const RCP<const T
    Tpetra::CrsMatrix<ST,LO,GO,NT> localA(rcpFromRef(gRowMap),0);
    localA.doImport(*A,import,Tpetra::INSERT);
 
-   RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > mat = Tpetra::createCrsMatrix<ST,LO,GO,NT>(rcpFromRef(rowMap),0);
-
    // get entry information
    LO numMyRows = rowMap.getNodeNumElements();
    LO maxNumEntries = A->getGlobalMaxNumRowEntries();
 
+   RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > mat = Tpetra::createCrsMatrix<ST,LO,GO,NT>(rcpFromRef(rowMap),maxNumEntries);
```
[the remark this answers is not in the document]
kddevin: OK; this one looks straightforward to do a count, then construct, then insert pattern.  I'll do that.

kddevin:
I addressed @mhoemmen 's concerns in the Teko test; it now counts the number of nonzeros per row and passes them to the constructor.
I also fixed the compilation error.
I retested with the ATDM test framework with
```
MPI_EXEC=`which mpiexec`
cmake \
  -GNinja \
  -DMPI_EXEC=${MPI_EXEC} \
  -DTrilinos_ENABLE_TESTS=ON \
  -DTrilinos_ENABLE_Tpetra=ON \
  -DTrilinos_ENABLE_Teko=ON \
  -DTrilinos_ENABLE_Belos=ON  \
  -DTrilinos_ENABLE_MueLu=ON  \
  -DTrilinos_ENABLE_Ifpack2=ON  \
  -DTrilinos_ENABLE_Anasazi=ON  \
  -DTrilinos_ENABLE_Thyra=ON  \
  -DTpetra_INST_INT_INT=ON \
  -DTpetra_INST_INT_LONG_LONG=OFF \
  -DDART_TESTING_TIMEOUT=90 \
.. |& tee OUTPUT.CMAKE

```
I'll need a fresh review.  Thanks!

packages/teko/src/Tpetra/Teko_InterlacedTpetra.cpp
```diff
@@ -243,12 +243,48 @@ RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > buildSubBlock(int i,int j,const RCP<const T
...
    LO maxNumEntries = A->getGlobalMaxNumRowEntries();
 
-   RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > mat = Tpetra::createCrsMatrix<ST,LO,GO,NT>(rcpFromRef(rowMap),maxNumEntries);
-
    // for extraction
    std::vector<GO> indices(maxNumEntries);
    std::vector<ST> values(maxNumEntries);
 
+   // for counting row sizes
+   std::vector<size_t> numEntriesPerRow(numMyRows,0);
```
mhoemmen: You don't actually need the zero here; `std::vector` invokes the default constructor of `size_t` for all the entries.

packages/teko/src/Tpetra/Teko_InterlacedTpetra.cpp
```diff
@@ -243,12 +243,48 @@ RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > buildSubBlock(int i,int j,const RCP<const T
...
+      for(size_t localCol=0;localCol<numEntries;localCol++) {
+         GO globalCol = indices[localCol];
+
+         // determinate which block this column ID is in
+         int block = globalCol / numGlobalVars;
+         
+         bool inFamily = true; 
+ 
+         // test the beginning of the block
+         inFamily &= (block*numGlobalVars+colBlockOffset <= globalCol);
```
mhoemmen: It looks like you might want `&&=` here instead of `&=`.

mhoemmen:
@kddevin , I think this is fine (thanks!!!) but I'm not sure if you want to merge yet, so I'll leave it.  It should be fine for a couple days.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
mhoemmen: @kddevin ~ 2015, we talked about another option to do thread-parallel graph creation: Add a new graph class that bounds the total number of entries per _process_, but does not fix the number of entries per _row_.  The class would use `Kokkos::UnorderedMap` inside.  (That might have been before UnorderedMap existed, so it was just a theory.)  That could be a helpful future option for users who only want to build a graph once.  Tpetra already has a CooMatrix that could be a start for such a class.

packages/teko/src/Tpetra/Teko_InterlacedTpetra.cpp
kddevin: This code was copied from the original insertion code (lines 314-315).  The code should be the same both in both the counting phase and insertion phase.

kddevin:
Thanks for the re-review, @mhoemmen

PR #5331, opened 2019-06-05
Fix std::complex<> interface to GEEV/GGEVX in Teuchos::LAPACK

hkthorn:
@trilinos/teuchos 

## Description

The generalized GEEV interface in Teuchos::LAPACK has been written
so that templated code can use a single call to LAPACK routines that
vary arguments based on real vs. complex arithmetic.  The GEEV and
GGEVX call both copy eigenvalues back from LAPACK only when the integer
(info) comes back zero.  Unfortunately, info is actually a pointer to
the integer.  So, when info is checked to be zero, the pointer is checked
to be zero.  It is not, so the values are not copied for calls to this
method that are complex-valued.  This commit fixes that error.

## Motivation and Context

GEEV/GGEVX would not return eigenvalues in complex arithmetic.  This affects users of those routines, including solvers like Belos' GCRODR.

## How Has This Been Tested?

MacOSX, GCC7, Serial

## Checklist

### DIFF
--- a/packages/belos/test/GCRODR/test_gcrodr_complex_hb.cpp
+++ b/packages/belos/test/GCRODR/test_gcrodr_complex_hb.cpp
@@ -157,12 +157,12 @@ int main(int argc, char *argv[]) {
     // *****************(can be user specified)******************
     int maxits = dim/blocksize; // maximum number of iterations to run
     int numBlocks = 100;
-    int numRecycledBlocks = 80;
+    int numRecycledBlocks = 20;
     int numIters1, numIters2, numIters3;
     ParameterList belosList;
     belosList.set( "Maximum Iterations", maxits );         // Maximum number of iterations allowed
     belosList.set( "Convergence Tolerance", tol );         // Relative convergence tolerance requested
-    belosList.set( "Verbosity", Belos::Errors + Belos::Warnings + Belos::StatusTestDetails);
+    belosList.set( "Verbosity", Belos::Errors + Belos::Warnings + Belos::StatusTestDetails + Belos::TimingDetails);
     belosList.set( "Num Blocks", numBlocks );
     belosList.set( "Num Recycled Blocks", numRecycledBlocks );
     // Construct the right-hand side and solution multivectors.
@@ -214,16 +214,13 @@ int main(int argc, char *argv[]) {
         norm_failure = true;
       }
     }
-    // Setup new rhs, perform solve
-    MVT::MvRandom( *soln );
-    OPT::Apply( *A, *soln, *rhs );
+    // Resolve linear system with same rhs and recycled space
     MVT::MvInit( *soln, zero );
     solver.reset(Belos::Problem);
     ret = solver.solve();
     numIters2=solver.getNumIters();
-    // Setup new rhs, perform solve
-    MVT::MvRandom( *soln );
-    OPT::Apply( *A, *soln, *rhs );
+
+    // Resolve linear system (again) with same rhs and recycled space
     MVT::MvInit( *soln, zero );
     solver.reset(Belos::Problem);
     ret = solver.solve();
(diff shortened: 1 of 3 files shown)


packages/belos/test/GCRODR/test_gcrodr_complex_hb.cpp
```diff
@@ -157,12 +157,12 @@ int main(int argc, char *argv[]) {
     // *****************(can be user specified)******************
     int maxits = dim/blocksize; // maximum number of iterations to run
     int numBlocks = 100;
-    int numRecycledBlocks = 80;
+    int numRecycledBlocks = 20;
```
bartlettroscoe: This is a non-backward compatible change, no?

packages/belos/test/GCRODR/test_gcrodr_complex_hb.cpp
```diff
@@ -234,7 +231,7 @@ int main(int argc, char *argv[]) {
     delete [] rowind;
     delete [] cvals;
     // Test for failures
-    if ( ret!=Belos::Converged || norm_failure || numIters1 < numIters2 || numIters1 < numIters3 ) {
+    if ( ret!=Belos::Converged || norm_failure || numIters1 < numIters2 || numIters2 < numIters3 ) {
```
bartlettroscoe: This is also a non-backward compatible change?

packages/teuchos/numerics/src/Teuchos_LAPACK.cpp
```diff
@@ -1331,7 +1331,7 @@ namespace Teuchos
     std::vector<std::complex<float> > w (n);
     std::complex<float>* w_rawPtr = (n == 0) ? NULL : &w[0];
     GEEV (JOBVL, JOBVR, n, A, lda, w_rawPtr, VL, ldvl, VR, ldvr, WORK, lwork, RWORK, info);
-    if (info == 0) {
+    if (*info == 0) {
```
bartlettroscoe: What there no unit test to catch this bug?  Is there no unit test to catch this now?

packages/belos/test/GCRODR/test_gcrodr_complex_hb.cpp
mhoemmen: @bartlettroscoe this is a test.

packages/belos/test/GCRODR/test_gcrodr_complex_hb.cpp
mhoemmen: This is a test.

packages/teuchos/numerics/src/Teuchos_LAPACK.cpp
mhoemmen: I agree that the state of `Teuchos::LAPACK` unit testing is poor, but I'm not sure if we have to fix that now.  On the other hand, if there's an easy way to write a regression test for this one issue, let's do it.

packages/belos/test/GCRODR/test_gcrodr_complex_hb.cpp
hkthorn: @bartlettroscoe @mhoemmen This test started failing after the correction to the GEEV/GGEVX interface.  That is because the assumptions made by the test are false and needed to be corrected.

packages/teuchos/numerics/src/Teuchos_LAPACK.cpp
bartlettroscoe: As long s there is a test that had to change to notice this, that is fine for now.

packages/belos/test/GCRODR/test_gcrodr_complex_hb.cpp
bartlettroscoe: That is good.  At least a test had to change (sorry I missed this was in the /test/ directory).  But there are not test cases for when `*info != 0`?

mhoemmen:
Thanks @hkthorn ! :D

PR #5332, opened 2019-06-05
Zoltan2: Convert MJ to use Kokkos and CUDA

MicheldeMessieres:
This PR converts Zoltan2 Multijagged data structures to Kokkos views and implements CUDA support.

Documentation is in progress.
Code is in a working state for CUDA, OpenMP, and Serial builds.

@trilinos/zoltan2 

## Description
Refactors all the MJ code to support Kokkos and implements CUDA algorithms.
Updates related tests for compatibility.
Added new tests for UVM off and on testing with CUDA.

## Motivation and Context
Multijagged needs to have CUDA capability.

## How Has This Been Tested?

white ATDM Cuda (parallel, serial), mac (parallel, OpenMP, serial).

## Checklist

- [ ] (in progress) I have updated the documentation accordingly.

### DIFF
--- a/packages/zoltan2/example/block/kokkosBlock.cpp
+++ b/packages/zoltan2/example/block/kokkosBlock.cpp
@@ -75,6 +75,7 @@ int main(int narg, char *arg[]) {
   typedef Map_t::local_ordinal_type localId_t;
   typedef Map_t::global_ordinal_type globalId_t;
   typedef Tpetra::Details::DefaultTypes::scalar_type scalar_t;
+  typedef Tpetra::Map<>::node_type node_t;
 
   ///////////////////////////////////////////////////////////////////////
   // Generate some input data.
@@ -82,7 +83,11 @@ int main(int narg, char *arg[]) {
   int localCount = 40 * (rank + 1);
   int totalCount = 20 * nprocs * (nprocs + 1);
   int targetCount = totalCount / nprocs;
-  Kokkos::View<globalId_t*> globalIds("globalIds", localCount);
+
+  Kokkos::View<globalId_t*, typename node_t::device_type>
+    globalIds(Kokkos::ViewAllocateWithoutInitializing("globalIds"), localCount);
+  typename decltype(globalIds)::HostMirror
+    host_globalIds = Kokkos::create_mirror_view(globalIds);
 
   if (rank == 0) {
     for (int i = 0, num = 40; i < nprocs ; i++, num += 40) {
@@ -96,18 +101,19 @@ int main(int narg, char *arg[]) {
   }
 
   for (int i = 0; i < localCount; i++) {
-    globalIds(i) = offset++;
+    host_globalIds(i) = offset++;
   }
+  Kokkos::deep_copy(globalIds, host_globalIds);
 
   ///////////////////////////////////////////////////////////////////////
   // Create a Zoltan2 input adapter with no weights
 
   typedef Zoltan2::BasicUserTypes<scalar_t, localId_t, globalId_t> myTypes;
   typedef Zoltan2::BasicKokkosIdentifierAdapter<myTypes> inputAdapter_t;
-  typedef typename Zoltan2::BasicKokkosIdentifierAdapter<myTypes>::weight_layout_t Layout;
 
   const int nWeights = 1;
-  Kokkos::View<scalar_t **, Layout> weights("weights", localCount, nWeights);
+  Kokkos::View<scalar_t **, typename node_t::device_type>
+    weights("weights", localCount, nWeights);
   for (int index = 0; index < localCount; index++) {
     weights(index, 0) = 1; // Error check relies on uniform weights
   }
@@ -141,22 +147,27 @@ int main(int narg, char *arg[]) {
   // Check and print the solution.
   // Count number of IDs assigned to each part; compare to targetCount
 
-  Kokkos::View<globalId_t *> ids;
+  Kokkos::View<const globalId_t *, typename node_t::device_type> ids;
   ia.getIDsKokkosView(ids);
 
-  Kokkos::View<int*> partCounts("partCounts", nprocs);
+  typename decltype(ids)::HostMirror host_ids =
+    Kokkos::create_mirror_view(ids);
+  Kokkos::deep_copy(host_ids, ids);
 
-  Kokkos::View<int*> globalPartCounts("globalPartCounts", nprocs);
+  Kokkos::View<int*, Kokkos::HostSpace> partCounts("partCounts", nprocs);
+
+  Kokkos::View<int*, Kokkos::HostSpace>
+    globalPartCounts("globalPartCounts", nprocs);
 
   for (size_t i = 0; i < ia.getLocalNumIDs(); i++) {
     int pp = problem->getSolution().getPartListView()[i];
-    std::cout << rank << " LID " << i << " GID " << ids(i)
+    std::cout << rank << " LID " << i << " GID " << host_ids(i)
               << " PART " << pp << std::endl;
     partCounts(pp)++;
   }
 
   Teuchos::reduceAll<int, int>(*comm, Teuchos::REDUCE_SUM, nprocs,
-      &partCounts(0), &globalPartCounts(0));
+      partCounts.data(), globalPartCounts.data());
 
   if (rank == 0) {
     int ierr = 0;
--- a/packages/zoltan2/example/geometric/rcb_C.cpp
+++ b/packages/zoltan2/example/geometric/rcb_C.cpp
@@ -61,14 +61,11 @@
 
 int main(int argc, char *argv[])
 {
-#ifdef HAVE_ZOLTAN2_MPI                   
-  MPI_Init(&argc, &argv);
-  int rank, nprocs;
-  MPI_Comm_size(MPI_COMM_WORLD, &nprocs);
-  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
-#else
-  int rank=0, nprocs=1;
-#endif
+  Tpetra::ScopeGuard tscope(&argc, &argv);
+  Teuchos::RCP<const Teuchos::Comm<int> > comm = Tpetra::getDefaultComm();
+
+  int rank = comm->getRank();
+  int nprocs = comm->getSize();
 
   // For convenience, we'll use the Tpetra defaults for local/global ID types
   // Users can substitute their preferred local/global ID types
@@ -428,10 +425,6 @@ int main(int argc, char *argv[])
   delete ia2;
   delete problem3;
   delete ia3;
-
-#ifdef HAVE_ZOLTAN2_MPI
-  MPI_Finalize();
-#endif
 
   if (rank == 0)
     std::cout << "PASS" << std::endl;
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -3875,10 +3875,11 @@ struct ReduceWeightsFunctor {
             // which loads a lot of coordinates on the same point, so without
             // this our cuts would all just sit at 0.
             part_t base_b = part;
+            scalar_t base_coord = cut_coordinates(base_b);
             part += 1;
             while(part < num_cuts) {
               b = cut_coordinates(part);
-              scalar_t delta = b - base_b;
+              scalar_t delta = b - base_coord;
               if(delta < 0) delta = -delta;
               if(delta < sEpsilon) {
                 // Note if on cut we set right/left closest to the cut itself
@@ -3900,7 +3901,7 @@ struct ReduceWeightsFunctor {
             part = base_b - 1;
             while(part >= 0) {
               b = cut_coordinates(part);
-              scalar_t delta = b - base_b;
+              scalar_t delta = b - base_coord;
               if(delta < 0) delta = -delta;
               if(delta < sEpsilon) {
                 // Note if on cut we set right/left closest to the cut itself
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -1595,15 +1595,7 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     local_part_xadj(0) = static_cast<mj_lno_t>(num_selected_coords);
   });
 
-  {
-    auto local_coordinate_permutations = this->coordinate_permutations;
-    auto local_initial_adjList_output_adjlist = initial_adjList_output_adjlist;
-    Kokkos::parallel_for(
-      Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_lno_t>
-        (0, num_total_coords), KOKKOS_LAMBDA (mj_lno_t i) {
-      local_coordinate_permutations(i) = local_initial_adjList_output_adjlist(i);
-    });
-  }
+  Kokkos::deep_copy(coordinate_permutations, initial_adjList_output_adjlist);
 
   mj_part_t current_num_parts = 1;
 
@@ -2085,15 +2077,7 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     this->new_part_xadj = Kokkos::View<mj_lno_t*, device_t>("empty");
   }
 
-  {
-    auto local_initial_adjList_output_adjlist = initial_adjList_output_adjlist;
-    auto local_coordinate_permutations = this->coordinate_permutations;
-    Kokkos::parallel_for(
-      Kokkos::RangePolicy<typename mj_node_t::execution_space,
-      mj_part_t> (0, num_total_coords), KOKKOS_LAMBDA (mj_part_t i) {
-      local_initial_adjList_output_adjlist(i) = local_coordinate_permutations(i);
-    });
-  }
+  Kokkos::deep_copy(initial_adjList_output_adjlist, coordinate_permutations);
 
   // Return output_xadj in CSR format
   output_xadj[0] = 0;
@@ -8235,16 +8219,18 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
                 output_part_index + output_array_shift) = part_size;
             });
 
-            auto local_new_coordinate_permutations =
-              this->new_coordinate_permutations;
-            auto local_coordinate_permutations =
-              this->coordinate_permutations;
-            Kokkos::parallel_for(
-              Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_lno_t>
-                (0, part_size), KOKKOS_LAMBDA (mj_lno_t n) {
-              local_new_coordinate_permutations(n+coordinate_begin) =
-                local_coordinate_permutations(n+coordinate_begin);
-            });
+            auto subview_new_coordinate_permutations =
+              Kokkos::subview(this->new_coordinate_permutations,
+                std::pair<mj_lno_t, mj_lno_t>(
+                  coordinate_begin,
+                  coordinate_begin + part_size));
+            auto subview_coordinate_permutations =
+              Kokkos::subview(this->coordinate_permutations,
+                std::pair<mj_lno_t, mj_lno_t>(
+                  coordinate_begin,
+                  coordinate_begin + part_size));
+            Kokkos::deep_copy(subview_new_coordinate_permutations,
+              subview_coordinate_permutations);
           }
           cut_shift += num_parts - 1;
           output_array_shift += num_parts;
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -1526,7 +1526,7 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
   // coordinates in MJ are LayoutLeft since Tpetra Multivector gives LayoutLeft
   Kokkos::View<mj_scalar_t **, Kokkos::LayoutLeft, device_t> &
     mj_coordinates_,
-  Kokkos::View<mj_lno_t *, device_t> & inital_adjList_output_adjlist,
+  Kokkos::View<mj_lno_t *, device_t> & initial_adjList_output_adjlist,
   mj_lno_t *output_xadj,
   int rd,
   const Kokkos::View<mj_part_t *, Kokkos::HostSpace> & part_no_array_,
@@ -1595,14 +1595,12 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     local_part_xadj(0) = static_cast<mj_lno_t>(num_selected_coords);
   });
 
-  auto local_coordinate_permutations =
-    this->coordinate_permutations;
-  auto local_inital_adjList_output_adjlist =
-    inital_adjList_output_adjlist;
+  auto local_coordinate_permutations = this->coordinate_permutations;
+  auto local_initial_adjList_output_adjlist = initial_adjList_output_adjlist;
   Kokkos::parallel_for(
     Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_lno_t>
       (0, num_total_coords), KOKKOS_LAMBDA (mj_lno_t i) {
-    local_coordinate_permutations(i) = local_inital_adjList_output_adjlist(i);
+    local_coordinate_permutations(i) = local_initial_adjList_output_adjlist(i);
   });
 
   mj_part_t current_num_parts = 1;
@@ -2087,7 +2085,7 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
   Kokkos::parallel_for(
     Kokkos::RangePolicy<typename mj_node_t::execution_space,
     mj_part_t> (0, num_total_coords), KOKKOS_LAMBDA (mj_part_t i) {
-    inital_adjList_output_adjlist(i) = local_coordinate_permutations(i);
+    local_initial_adjList_output_adjlist(i) = local_coordinate_permutations(i);
   });
 
   // Return output_xadj in CSR format
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -1995,16 +1995,18 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
                 = part_size;
             });
 
-            auto local_new_coordinate_permutations =
-              this->new_coordinate_permutations;
-            auto local_coordinate_permutations =
-              this->coordinate_permutations;
-            Kokkos::parallel_for(
-              Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_lno_t>
-              (0, part_size), KOKKOS_LAMBDA (mj_lno_t n) {
-              local_new_coordinate_permutations(n+coordinate_begin) =
-                local_coordinate_permutations(n+coordinate_begin);
-            });
+            auto subview_new_coordinate_permutations =
+              Kokkos::subview(this->new_coordinate_permutations,
+                std::pair<mj_lno_t, mj_lno_t>(
+                  coordinate_begin,
+                  coordinate_begin + part_size));
+            auto subview_coordinate_permutations =
+              Kokkos::subview(this->coordinate_permutations,
+                std::pair<mj_lno_t, mj_lno_t>(
+                  coordinate_begin,
+                  coordinate_begin + part_size));
+            Kokkos::deep_copy(subview_new_coordinate_permutations,
+              subview_coordinate_permutations);
           }
 
           cut_shift += num_parts - 1;
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -1123,7 +1123,7 @@ public:
     Kokkos::View<mj_lno_t *, device_t> &
       initial_selected_coords_output_permutation,
     mj_lno_t *output_xadj,
-    int recursion_depth,
+    int recursion_depth_,
     const Kokkos::View<mj_part_t *, Kokkos::HostSpace> & part_no_array,
     bool partition_along_longest_dim,
     int num_ranks_per_node,
@@ -1528,7 +1528,7 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     mj_coordinates_,
   Kokkos::View<mj_lno_t *, device_t> & initial_adjList_output_adjlist,
   mj_lno_t *output_xadj,
-  int rd,
+  int recursion_depth_,
   const Kokkos::View<mj_part_t *, Kokkos::HostSpace> & part_no_array_,
   bool partition_along_longest_dim,
   int num_ranks_per_node,
@@ -1550,7 +1550,7 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
   this->imbalance_tolerance = 0;
   this->num_global_parts = num_target_part;
   this->part_no_array = part_no_array_;
-  this->recursion_depth = rd;
+  this->recursion_depth = recursion_depth_;
 
   // If nonuniform first level partitioning, the requested num of parts and the
   // requested distribution of elements for each part
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -1595,13 +1595,15 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     local_part_xadj(0) = static_cast<mj_lno_t>(num_selected_coords);
   });
 
-  auto local_coordinate_permutations = this->coordinate_permutations;
-  auto local_initial_adjList_output_adjlist = initial_adjList_output_adjlist;
-  Kokkos::parallel_for(
-    Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_lno_t>
-      (0, num_total_coords), KOKKOS_LAMBDA (mj_lno_t i) {
-    local_coordinate_permutations(i) = local_initial_adjList_output_adjlist(i);
-  });
+  {
+    auto local_coordinate_permutations = this->coordinate_permutations;
+    auto local_initial_adjList_output_adjlist = initial_adjList_output_adjlist;
+    Kokkos::parallel_for(
+      Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_lno_t>
+        (0, num_total_coords), KOKKOS_LAMBDA (mj_lno_t i) {
+      local_coordinate_permutations(i) = local_initial_adjList_output_adjlist(i);
+    });
+  }
 
   mj_part_t current_num_parts = 1;
 
@@ -1633,10 +1635,6 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     view_total_reduction_size("view_total_reduction_size", 1);
 
   for(int rd = 0; rd < this->recursion_depth; ++rd) {
-
-    auto local_new_coordinate_permutations =
-      this->new_coordinate_permutations;
-
     // next_future_num_parts_in_parts will be as the size of outnumParts,
     // and this will hold how many more parts that each output part
     // should be divided. this array will also be used to determine the weight
@@ -2005,6 +2003,10 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
                 = part_size;
             });
 
+            auto local_new_coordinate_permutations =
+              this->new_coordinate_permutations;
+            auto local_coordinate_permutations =
+              this->coordinate_permutations;
             Kokkos::parallel_for(
               Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_lno_t>
               (0, part_size), KOKKOS_LAMBDA (mj_lno_t n) {
@@ -2030,6 +2032,8 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
             host_num_partitioning_in_current_dim(current_work_part + kk);
           auto local_new_part_xadj = this->new_part_xadj;
           auto local_mj_current_dim_coords = mj_current_dim_coords;
+          auto local_new_coordinate_permutations =
+            new_coordinate_permutations;
           Kokkos::parallel_for(
             Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_part_t> (
             0, num_parts), KOKKOS_LAMBDA (mj_part_t ii) {
@@ -2081,12 +2085,15 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     this->new_part_xadj = Kokkos::View<mj_lno_t*, device_t>("empty");
   }
 
-  local_coordinate_permutations = this->coordinate_permutations;
-  Kokkos::parallel_for(
-    Kokkos::RangePolicy<typename mj_node_t::execution_space,
-    mj_part_t> (0, num_total_coords), KOKKOS_LAMBDA (mj_part_t i) {
-    local_initial_adjList_output_adjlist(i) = local_coordinate_permutations(i);
-  });
+  {
+    auto local_initial_adjList_output_adjlist = initial_adjList_output_adjlist;
+    auto local_coordinate_permutations = this->coordinate_permutations;
+    Kokkos::parallel_for(
+      Kokkos::RangePolicy<typename mj_node_t::execution_space,
+      mj_part_t> (0, num_total_coords), KOKKOS_LAMBDA (mj_part_t i) {
+      local_initial_adjList_output_adjlist(i) = local_coordinate_permutations(i);
+    });
+  }
 
   // Return output_xadj in CSR format
   output_xadj[0] = 0;
--- a/packages/zoltan2/src/input/Zoltan2_Adapter.hpp
+++ b/packages/zoltan2/src/input/Zoltan2_Adapter.hpp
@@ -104,6 +104,7 @@ public:
   typedef typename InputTraits<User>::lno_t lno_t;
   typedef typename InputTraits<User>::gno_t gno_t;
   typedef typename InputTraits<User>::scalar_t scalar_t;
+  typedef typename InputTraits<User>::node_t node_t;
   typedef typename InputTraits<User>::part_t part_t;  
   typedef typename InputTraits<User>::offset_t offset_t;
 
@@ -116,23 +117,37 @@ public:
   virtual ~BaseAdapter() {};
 
   /*! \brief Provide a pointer to this process' identifiers.
-
       \param ids will on return point to the list of the global Ids for 
         this process.
    */
   virtual void getIDsView(const gno_t *&ids) const {
-    Kokkos::View<gno_t *> kokkosIds;
+    // If adapter does not define getIDsView, getIDsKokkosView is called.
+    // If adapter does not define getIDsKokkosView, getIDsView is called.
+    // Allows forward and backwards compatibility.
+    Kokkos::View<const gno_t *, typename node_t::device_type> kokkosIds;
     getIDsKokkosView(kokkosIds);
     ids = kokkosIds.data();
   }
 
-  /*! \brief Provide a pointer to this process' identifiers.
-
+  /*! \brief Provide a Kokkos view to this process' identifiers.
       \param ids will on return point to the list of the global Ids for 
         this process.
    */
-  virtual void getIDsKokkosView(Kokkos::View<gno_t *> &/* ids */) const {
-    Z2_THROW_NOT_IMPLEMENTED
+  virtual void getIDsKokkosView(Kokkos::View<const gno_t *,
+    typename node_t::device_type> &ids) const {
+    // If adapter does not define getIDsView, getIDsKokkosView is called.
+    // If adapter does not define getIDsKokkosView, getIDsView is called.
+    // Allows forward and backwards compatibility.
+    const gno_t * ptr_ids;
+    getIDsView(ptr_ids);
+    typedef Kokkos::View<gno_t *, typename node_t::device_type> view_t;
+    view_t non_const_ids = view_t("ptr_ids", getLocalNumIDs());
+    typename view_t::HostMirror host_ids = Kokkos::create_mirror_view(non_const_ids);
+    for(size_t i = 0; i < this->getLocalNumIDs(); ++i) {
+       host_ids(i) = ptr_ids[i];
+    }
+    Kokkos::deep_copy(non_const_ids, host_ids);
+    ids = non_const_ids;
   }
 
   ///*! \brief Provide pointer to a weight array with stride.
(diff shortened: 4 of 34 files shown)


packages/zoltan2/example/block/kokkosBlock.cpp
```diff
@@ -75,14 +75,15 @@ int main(int narg, char *arg[]) {
...
+  typedef Tpetra::Map<>::node_type node_t;
 
   ///////////////////////////////////////////////////////////////////////
   // Generate some input data.
 
   int localCount = 40 * (rank + 1);
   int totalCount = 20 * nprocs * (nprocs + 1);
   int targetCount = totalCount / nprocs;
-  Kokkos::View<globalId_t*> globalIds("globalIds", localCount);
+  Kokkos::View<globalId_t*, Kokkos::HostSpace> globalIds("globalIds", localCount);
```
mhoemmen: The typical pattern for tests is to create device Views, but use `create_mirror_view` to get host mirrors, fill the host mirrors, then copy back to device.

packages/zoltan2/example/block/kokkosBlock.cpp
```diff
@@ -141,7 +142,7 @@ int main(int narg, char *arg[]) {
   // Check and print the solution.
   // Count number of IDs assigned to each part; compare to targetCount
 
-  Kokkos::View<globalId_t *> ids;
+  Kokkos::View<const globalId_t *, Kokkos::HostSpace> ids;
```
mhoemmen: Comment on line 156 below: You're assuming UVM there.  Consider the host mirror approach I discussed above.

mhoemmen: Comment on line 160 above: Please never do `&array(0)` or `&array[0]` or anything like that, ever, with any array data structure that has a `.data()` method.
 
1. Kokkos::View, like std::vector, has a `.data()` method; use it.
2. `&x[0]` with a zero-length std::vector may throw if you are using a checked STL.
3. `&x(0)` with a zero-length Kokkos::View will throw if Kokkos debug checking is on.

mhoemmen: Line 165 also assumes UVM.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9172 @@
...
-     */
-    void fill_permutation_array(
-        mj_part_t output_num_parts,
-        mj_part_t num_parts);
-
-    /*! \brief Function checks if should do migration or not.
-     * \param current_num_parts is the number of parts in the process.
-     * \param output_part_begin_index is the number that will be used as beginning part number
-     * \param output_part_boxes is the array that holds the part boxes
-     * \param is_data_ever_migrated is the boolean
```
srajama1: TODO

srajama1: Another worrisome TODO

srajama1: I didn't think resize could be used this way for NULL. Need to check.

srajama1: Note to self : Think about this

srajama1: I think this ifdef should go away

packages/zoltan2/example/block/kokkosBlock.cpp
MicheldeMessieres: For MJ we decided to keep some things like IDs on HostSpace since we never use them in the kernels and this makes the current MPI communications more efficient. @kddevin I had converted this test to HostSpace for consistency with that change. This could probably use some further discussion when we go over these tests.

packages/zoltan2/example/block/kokkosBlock.cpp
MicheldeMessieres: Switched to use data() method. Switched the views to HostSpace for now. See comments on prior. Will discuss with Karen to figure out how we want to structure this test. I'm not sure we want to work with any kernels or conversions here.

packages/zoltan2/example/geometric/rcb_C.cpp
```diff
@@ -429,10 +426,6 @@ int main(int argc, char *argv[])
   delete problem3;
```
mhoemmen: Best practice is to use `unique_ptr` for objects that need to be managed by pointer, yet have local scope.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,21 +54,16 @@
...
-#include <Teuchos_StandardParameterEntryValidators.hpp>
-
+#include <Zoltan2_CoordinatePartitioningGraph.hpp>
+#include <Zoltan2_Util.hpp>
 #include <Tpetra_Distributor.hpp>
+#include <Teuchos_StandardParameterEntryValidators.hpp>
 #include <Teuchos_ParameterList.hpp>
-#include <Zoltan2_CoordinatePartitioningGraph.hpp>
+
 #include <new>          // ::operator new[]
```
mhoemmen: You don't need to include `<new>` just to use default `operator new`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
+// facilitate testing this on and off. When ZOLTAN2_MJ_USE_CUDA_KERNEL is
+// defined, the main kernel is run using a parallel_for -> parallel_for loop
+// with atomics, instead of parallel_reduce -> parallel_reduce which gives
+// better performance on host. This occurs because CUDA is more efficient with
+// atomics than host code. I understand ScatterView may be a mechanism to allow
+// a single code pipe-line which runs both reduction or atomic patterns but did
+// not investigate that yet.
+// TODO: Implement ScatterView and get rid of this macro.
+#ifdef KOKKOS_HAVE_CUDA
+#define ZOLTAN2_MJ_USE_CUDA_KERNEL // Atomic Atomic Loops
```
mhoemmen: This should be a run-time choice.  You might want the atomic updates - based algorithm for OpenMP, depending on the number of threads available at run time.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
 #endif
 
+// TODO: This macro will be removed after some further profiling.
+// This option is being maintained to evaluate the perfomance of using floats
+// versus doubles for the reduction arrays in the main kernel of MJ.
+#define ZOLTAN2_MJ_USE_FLOAT_ARRAY_FOR_KERNEL
+
 #define LEAST_SIGNIFICANCE 0.0001
```
mhoemmen: Please don't pollute the global namespace with non-namespaced macros in a public header file.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
 #define FUTURE_REDUCEALL_CUTOFF 1500000
+
 //if parts right before last dimension are estimated to have less than
 //MIN_WORK_LAST_DIM many coords, migration will be forced in earlier iterations.
 #define MIN_WORK_LAST_DIM 1000
 
-
-
-
 #define ZOLTAN2_ABS(x) ((x) >= 0 ? (x) : -(x))
```
mhoemmen: Remove this and use `std::abs`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
+
 //imbalance calculation. Wreal / Wexpected - 1
 #define imbalanceOf(Wachieved, totalW, expectedRatio) \
```
mhoemmen: See note about non-namespaced macros.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
         (Wachieved) / ((totalW) * (expectedRatio)) - 1
 #define imbalanceOf2(Wachieved, wExpected) \
         (Wachieved) / (wExpected) - 1
 
-
 #define ZOLTAN2_ALGMULTIJAGGED_SWAP(a,b,temp) temp=(a);(a)=(b);(b)=temp;
```
mhoemmen: Remove this and use `std::swap`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
 */
-
 template <typename Ordinal, typename T>
 class Zoltan2_BoxBoundaries  : public ValueTypeReductionOp<Ordinal,T>
 {
 private:
-    Ordinal size;
-    T _EPSILON;
+  Ordinal size;
+  T _EPSILON;
```
mhoemmen: Symbols starting with underscore are reserved by the system.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
-            throw  "cannot allocate memory";
-        }
-        return a;
-    }
-    else {
-        return NULL;
+T *allocMemory(size_t size) {
+  if(size > 0) {
+    T * a = new T[size];
+    if(a == NULL) {
```
mhoemmen: You don't need this.  C++98 requires that default `operator new` throw `bad_alloc` on failure to allocate.  https://en.cppreference.com/w/cpp/memory/new/operator_new

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
- *
  */
 template <typename T>
-void freeArray(T *&array){
-    if(array != NULL){
-        delete [] array;
-        array = NULL;
-    }
+void freeArray(T *&array) {
+  if(array != NULL) {
```
mhoemmen: You don't need this.  `delete` on a null pointer is harmless.  https://en.cppreference.com/w/cpp/language/delete

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
-            }
-            //if next value is smaller return false;
-            else //(this->val[i] > other.val[i])
-            {
-                return false;
-            }
-        }
-        //if they are totally equal.
-        return this->index > other.index;
+  bool operator>(const uMultiSortItem<IT,CT,WT>& other) const {
```
mhoemmen: `operator<` should suffice for sorting.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
-                istack[jstack-1]=i;
-                ir=j-1;
-            }
-            else
-            {
-                istack[jstack]=j-1;
-                istack[jstack-1]=l;
-                l=i;
-            }
+void uqsort(IT n, uSortItem<IT, WT> * arr) {
```
mhoemmen: You don't need this.  If you define `operator<` on `uSortItem`, then you can use `std::sort`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
-    bool operator<=(const uSignedSortItem<IT, WT, SIGN>& rhs){
-      return !(*this > rhs);}
-    bool operator>=(const uSignedSortItem<IT, WT, SIGN>& rhs){
-      return !(*this  < rhs);}
+    else {
+      /*if I am negative, the other is positive*/
+      return false;
+    }
+  }
+  bool operator<=(const uSignedSortItem<IT, WT, SIGN>& rhs) {
```
mhoemmen: If you use `std::sort`, you should only need `operator<`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -78,7013 +73,9090 @@
...
-     */
-    void fill_permutation_array(
-        mj_part_t output_num_parts,
-        mj_part_t num_parts);
-
-    /*! \brief Function checks if should do migration or not.
-     * \param current_num_parts is the number of parts in the process.
-     * \param output_part_begin_index is the number that will be used as beginning part number
-     * \param output_part_boxes is the array that holds the part boxes
-     * \param is_data_ever_migrated is the boolean
```
mhoemmen: If the type of the thing is `bool`, you don't need to say "it's a boolean value."

mhoemmen: Use `std::to_string`: https://en.cppreference.com/w/cpp/string/basic_string/to_string

packages/zoltan2/example/geometric/rcb_C.cpp
MicheldeMessieres: Will discuss with team - there are a bunch of places like this so perhaps better done in a separate job since this is not new code. @kddevin

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: @kddevin Recommended changes would simplify code a lot. But this code is not really changed with the PR, just cleaned up formatting, so we may want to consider as a separate PR.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: @kddevin Recommended changes would simplify code a lot. But this code is not really changed with the PR, just cleaned up formatting, so we may want to consider as a separate PR.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: @kddevin Recommended changes would simplify code a lot. But this code is not really changed with the PR, just cleaned up formatting, so we may want to consider as a separate PR.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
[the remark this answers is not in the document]
MicheldeMessieres: All views now so no management of delete.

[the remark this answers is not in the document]
MicheldeMessieres: We discussed and since the original code had mixed formats, we decided we won't fix all now and just document. I still have general documentation to fix everywhere.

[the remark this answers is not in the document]
MicheldeMessieres: We discussed and since the original code had mixed formats, we decided we won't fix all now and just document. I still have general documentation to fix everywhere.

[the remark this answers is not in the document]
MicheldeMessieres: We discussed and since the original code had mixed formats, we decided we won't fix all now and just document. I still have general documentation to fix everywhere.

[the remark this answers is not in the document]
MicheldeMessieres: Fixed - taskMapper now tested and running for UVM off. I forgot to remove this comment but just removed it for next update.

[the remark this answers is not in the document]
MicheldeMessieres: Currently specifying AUTO for threads, but the parameter controls teams. I'm not sure if you meant to use AUTO for teams as well as threads as I didn't think that was possible for team count. Need to discuss.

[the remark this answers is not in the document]
MicheldeMessieres: Switched everything to be default (no layout specified) except the coords stuff for compatibility with Tpetra multi vector which gives LayoutLeft.

MicheldeMessieres: Refactored and this was removed/changed.

MicheldeMessieres: Just using KOKKOS_HAVE_CUDA now so the macro is gone but as Mark suggested, perhaps we want no macro and a parameter to control the choice of parallel_reduce with no atomics, or parallel_for with atomics. Currently Cuda runs the parallel_for and OpenMP uses parallel_reduce. Need to discuss.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Done

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Macro removed and just using KOKKOS_HAVE_CUDA but will discuss with team having it as an option.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: changed all macros to be ZOLTAN2 prefix.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Done

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Changed all macros to be ZOLTAN2 prefixed.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Done

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Fixed

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Fixed. @kddevin I think just using new and delete[] would be a good change but I wasn't positive if there might be a logic problem if something allocated for size 0. Original would return null but now we have new type[0] returns a non null ptr in C++. I kept this change as a separate commit in case it was a problem. But tests are all ok.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Fixed. @kddevin I think just using new and delete[] would be a good change but I wasn't positive if there might be a logic problem if something allocated for size 0. Original would return null but now we have new type[0] returns a non null ptr in C++. I kept this change as a separate commit in case it was a problem. But tests are all ok.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Fixed. Note PR has a lot of doc stuff still to fix which I have commented but not done yet. Still TODO.

MicheldeMessieres: Fixed

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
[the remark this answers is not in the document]
MicheldeMessieres: Generally improved many of the single host-device communications - eliminated some by converting small arrays back to host only. Others fixed by some restructuring.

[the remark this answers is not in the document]
MicheldeMessieres: Yes. Is Ok.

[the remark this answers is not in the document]
MicheldeMessieres: Fixed all TODOs except 1 - will discuss. With new commits these comments now showing up in code and not synched here - so not certain which is which. But generally, they are done.

MicheldeMessieres: Fixed all TODOs except 1 - will discuss. With new commits these comments now showing up in code and not synched here - so not certain which is which. But generally, they are done.

MicheldeMessieres: Fixed all TODOs except 1 - will discuss. With new commits these comments now showing up in code and not synched here - so not certain which is which. But generally, they are done.

packages/zoltan2/example/geometric/rcb_C.cpp
mhoemmen: Naked `new` and `delete` is an antipattern specifically called out in the C++ Core Guidelines.  For example:

  - [P.3](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#p3-express-intent)
  - [R.1](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rr-raii)
  - [R.11 ("Avoid calling new and delete explicitly")](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rr-newdelete)

I'm not your boss and perhaps you didn't write this code, so I won't tell you what to do here.  I'm just pointing out the above.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,7037 +54,8852 @@
...
-#if defined(__cplusplus) && __cplusplus >= 201103L
 #include <unordered_map>
-#else
-#include <Teuchos_Hashtable.hpp>
-#endif // C++11 is enabled
 
 #ifdef ZOLTAN2_USEZOLTANCOMM
 #ifdef HAVE_ZOLTAN2_MPI
-#define ENABLE_ZOLTAN_MIGRATION
+#define ZOLTAN2_MJ_ENABLE_ZOLTAN_MIGRATION
```
mhoemmen: It would be nice if Zoltan2 didn't expose Zoltan headers to downstream code.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,7037 +54,8852 @@
...
-#define FUTURE_REDUCEALL_CUTOFF 1500000
+#define ZOLTAN2_MJ_FUTURE_REDUCEALL_CUTOFF 1500000
+
 //if parts right before last dimension are estimated to have less than
 //MIN_WORK_LAST_DIM many coords, migration will be forced in earlier iterations.
-#define MIN_WORK_LAST_DIM 1000
-
-
-
+#define ZOLTAN2_MJ_MIN_WORK_LAST_DIM 1000
```
mhoemmen: This is better than before.  Consider `constexpr` constants instead of macros in future revisions.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,7037 +54,8852 @@
...
 
-#define ZOLTAN2_ABS(x) ((x) >= 0 ? (x) : -(x))
 //imbalance calculation. Wreal / Wexpected - 1
-#define imbalanceOf(Wachieved, totalW, expectedRatio) \
-        (Wachieved) / ((totalW) * (expectedRatio)) - 1
-#define imbalanceOf2(Wachieved, wExpected) \
+#define ZOLTAN2_MJ_IMBALANCEOF2(Wachieved, wExpected) \
         (Wachieved) / (wExpected) - 1
```
mhoemmen: If the `wExpected` expression is an integer, this will truncate.  In general, prefer functions (templated if necessary) to macros.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,7037 +54,8852 @@
...
-    volatile WT _EPSILON;
-
-    uMultiSortItem(){
-        this->index = 0;
-        this->count = 0;
-        this->val = NULL;
-        this->_EPSILON = std::numeric_limits<WT>::epsilon() * 100;
-    }
-
+  // TODO: Why volatile?
```
mhoemmen: Would be nice to know why `volatile`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,7037 +54,8852 @@
...
+    this->epsilon = std::numeric_limits<WT>::epsilon() * 100;
+  }
 
-    uMultiSortItem( const uMultiSortItem<IT,CT,WT>& other ){
-        this->index = other.index;
-        this->count = other.count;
-        this->val = other.val;
-        this->_EPSILON = other._EPSILON;
-    }
+  uMultiSortItem( const uMultiSortItem<IT,CT,WT>& other ) {
```
mhoemmen: You don't need to provide any of these copy constructors or assignment operators.  The [Rule of Zero](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-zero) will provide correct defaults here.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
mhoemmen: I'm not your boss -- I just think that if it doesn't happen now, it probably will never happen.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
mhoemmen: This is a function that almost certainly will never get tested, and thus introduces technical debt on creation.  I'm not your boss, but I think now is a good time to clean it up.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,7037 +54,8852 @@
...
+        do i++; while (arr[i].val < aval);
+        do j--; while (arr[j].val > aval);
+        if(j < i) break;
+        std::swap(arr[i],arr[j]);
+      }
+      arr[l]=arr[j];
+      arr[j]=a;
+      jstack += 2;
+      if(jstack > NSTACK) {
+        std::cout << "uqsort: NSTACK too small in sort." << std::endl;
```
mhoemmen: We can't have every MPI process printing some message.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,7037 +54,8852 @@
...
+        exit(1);
```
mhoemmen: It would be better at least to call `MPI_Abort` or `std::terminate` so that we have some opportunity to print a stack trace etc.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -54,7037 +54,8852 @@
...
+        do i++; while (arr[i] < a);
+        do j--; while (arr[j] > a);
+        if(j < i) break;
+        std::swap(arr[i],arr[j]);
+      }
+      arr[l]=arr[j];
+      arr[j]=a;
+      jstack += 2;
+      if(jstack > NSTACK) {
+        std::cout << "uqsort: NSTACK too small in sort." << std::endl;
```
mhoemmen: See above.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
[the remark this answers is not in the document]
MicheldeMessieres: We discussed by phone. I think it's not possible to pass Kokkos::AUTO for the team count. I don't see that used and would not compile. So I believe we always need to pass a calculated team count. You plan to check this.

mhoemmen: @MicheldeMessieres You'll need to use `KOKKOS_ENABLE_CUDA`, not `KOKKOS_HAVE_CUDA`.  The latter no longer exists.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
mhoemmen: @MicheldeMessieres Be sure to use `KOKKOS_ENABLE_CUDA`, not `KOKKOS_HAVE_CUDA`.  The latter has long been deprecated and may not exist any more.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Removed operator>. I didn't realize it was unused. Thanks.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: I removed operator> and logic now uses just operator< but did not switch to the std::sort - not sure exactly what uqsort does.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: I removed operator> and logic now uses just operator< but did not switch to the std::sort - not sure exactly what uqsort does.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Switched to constexpr.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: switched to a function.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: switched these exit(1) calls to std::terminate

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
mhoemmen: @MicheldeMessieres wrote:
> not sure exactly what uqsort does

It looks like a non-recursive variant of quicksort.  More untested functions, joy ;-)

MicheldeMessieres: @kddevin uqsort is currently only tested through migration, which won't execute in our current test suite. So we have a bunch of methods related to migration not default tested. I think adding these should be straight forward. We should probably have some tests to use the Zoltan migration option as well.

mhoemmen: I'm not your boss, but it seems like adding unit tests for yet another sort routine is a waste of time, vs. just making it work with `std::sort`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
mhoemmen: or you could just call `std::sort` :-)

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -6460,8 +6460,18 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
       Kokkos::create_mirror_view(this->mj_coordinates);
     Kokkos::deep_copy(host_src_coordinates, this->mj_coordinates);
     for(int i = 0; i < this->coord_dim; ++i) {
-      auto sub_host_src_coordinates
-        = Kokkos::subview(host_src_coordinates, Kokkos::ALL, i);
+      Kokkos::View<mj_scalar_t*, Kokkos::LayoutLeft, device_t>
+        sub_host_src_coordinates;
+      if(host_src_coordinates.extent(0) == 0) {
+        // view could be size 0 if graph was not distributed - make an empty view
+        sub_host_src_coordinates =
```
mhoemmen: You don't need this; it's already an empty View (that's what the default constructor does).  
```c++
if (host_src_coordinates.extent(0) != 0) {
  sub_host_src_coordinates = 
    Kokkos::subview(host_src_coordinates, Kokkos::ALL, i);
}
```

MicheldeMessieres: Thanks. Resolved.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
kddevin: @MicheldeMessieres You are correct that we could use a migration test in the test suite.  It isn't required for this PR.  Perhaps we can create a separate issue so that we don't forget to do it.  My concern is that the test might be large for standard Trilinos testing; I'd have to look at the details of migration for that.

packages/zoltan2/src/input/Zoltan2_Adapter.hpp
```diff
@@ -116,22 +117,31 @@ template <typename User>
...
-  virtual void getIDsKokkosView(Kokkos::View<gno_t *> &/* ids */) const {
+  virtual void getIDsKokkosView(Kokkos::View<const gno_t *,
+    Kokkos::HostSpace> &ids) const {
+    Z2_THROW_NOT_IMPLEMENTED
+  }
+
+  /*! \brief Provide a pointer to the coordinates view.
+  */
+  inline void getCoordinatesKokkosView(
+    // coordinates in MJ are LayoutLeft since Tpetra Multivector gives LayoutLeft
```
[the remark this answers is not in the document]
mhoemmen: @kddevin I think a different interface could help.  For example (requires C++14 but easy to work around in legal C++11):
```c++
withCoordinates (
  [] (const auto& coords) { /* user's code that accesses coords */ },
  readOnly (zoltan2AdapterInstance).as (Kokkos::LayoutLeft ()).on (Kokkos::HostSpace ())
);
```
This says, "Within the given lambda or function, I want to access the coordinates (local to my MPI process) in read-only fashion, as a LayoutLeft Kokkos::View in host memory."  If the coordinates are stored as a LayoutLeft View in host memory, then I can just view them.  Otherwise, the implementation of `withLocalCoordinates` would make a temporary copy.  `readOnly` tells Zoltan2 that it need not copy the coordinates back to their original storage format after use. 

You could have different `with*` functions for different kinds of data: coordinates, weights, global indices, etc.  Alternately, you could specify the kind of data by tag:
```c++
withLocalDataAccess (
  [] (auto coords) { /* user's code that accesses coordinates */ },
  from (zoltan2AdapterInstance).readOnly (coordinates_tag).as (Kokkos::LayoutLeft ()).on (Kokkos::HostSpace ())
);

withLocalDataAccess (
  [] (auto weights) { /* user's code that accesses weights */ },
  from (zoltan2AdapterInstance).readWrite (weights_tag).as (Kokkos::LayoutRight ()).on (Kokkos::CudaSpace ())
);
```
With this tag-based approach, you could access multiple kinds of data at a time:
```c++
withLocalDataAccess (
  [] (auto coords, auto weights) { /* user's code */ },
  from (zoltan2AdapterInstance).readOnly (coordinates_tag).as (Kokkos::LayoutLeft ()).on (Kokkos::HostSpace (),
  from (zoltan2AdapterInstance).readWrite (weights_tag).as (Kokkos::LayoutRight ()).on (Kokkos::CudaSpace ()
);
```

packages/zoltan2/src/input/Zoltan2_Adapter.hpp
```diff
@@ -146,9 +156,9 @@ template <typename User>
   // */ 
   virtual void getWeightsView(const scalar_t *&wgt, int &stride,
                               int idx = 0) const {
-    Kokkos::View<scalar_t *> tempWeightsView;
-    getWeightsKokkosView(tempWeightsView, idx);
-    wgt = tempWeightsView.data();
+    Kokkos::View<scalar_t *, typename node_t::device_type> kokkos_wgts;
```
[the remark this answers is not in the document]
MicheldeMessieres: MJ will use the ids only for migration. So we had decided to make ids host only since otherwise we would have a copy to host, MPI calls, then a copy back to device.

packages/zoltan2/src/input/Zoltan2_Adapter.hpp
```diff
@@ -159,9 +169,16 @@ template <typename User>
...
   // */ 
-  virtual void getWeightsKokkosView(Kokkos::View<scalar_t *> &/* wgt */, 
-                              int /* idx */ = 0) const {
-    Z2_THROW_NOT_IMPLEMENTED
+  virtual void getWeightsKokkosView(Kokkos::View<scalar_t *,
+    typename node_t::device_type> &wgt, int idx) const {
+    throw std::logic_error("getWeightsKokkosView not implemented.");
+  }
+
+  // TODO: 2nd form returns the 2darray - still need to decide how this
```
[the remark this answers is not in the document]
MicheldeMessieres: I removed this TODO. Currently we have getWeightsKokkosView and a getWeightsKokkosView2D. Only two things depend on getWeightsKokkosView, the tests Zoltan2_BasicKokkosIdentifierInput and Zoltan2_kokkosBlock. Zoltan2_BasicKokkosIdentifierInput would easily be refactored to use the 2D form, which MJ also depends on. Zoltan2_kokkosBlock works through Zoltan2::IdentifierModel which calls getWeightsView, so depends on it then calling getWeightsKokkosView. Maybe the role of these tests could be rethought. I am wondering if we would like to remove the getWeightsKokkosView completely. Then the getWeightsView method here would throw instead of calling getWeightsKokkosView. It depends on whether we want to preserve Zoltan2_kokkosBlock in the present form.

packages/zoltan2/src/input/Zoltan2_VectorAdapter.hpp
```diff
@@ -141,6 +141,14 @@ template <typename User>
   virtual void getEntriesView(const scalar_t *&elements, int &stride,
                               int idx = 0) const = 0;
 
+  virtual void getEntriesKokkosView(
+    // coordinates in MJ are LayoutLeft since Tpetra Multivector gives LayoutLeft
+    Kokkos::View<scalar_t **, Kokkos::LayoutLeft,
+    typename node_t::device_type> & elements) const {
+    // TODO: Eventually this can become abstract if everything is using kokkos
```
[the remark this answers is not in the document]
MicheldeMessieres: I removed this TODO as it was more of a comment to myself.

packages/zoltan2/src/util/Zoltan2_ImbalanceMetricsUtility.hpp
```diff
@@ -156,14 +156,16 @@ template <typename scalar_t, typename lno_t, typename part_t>
                                              // [0,maxPartPlusOne)
 
   part_t globalSumSize = maxPartPlusOne * numMetrics;
-  scalar_t * sumBuf = new scalar_t [globalSumSize];
+
+  typedef scalar_t temp_t; // TODO remove this - added to resolve cuda warnings I didn't understand yet
```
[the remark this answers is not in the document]
MicheldeMessieres: I removed this. It causes a warning now but this is common in many spots:
warning: non-constant array new length must be specified without parentheses around the type-id [-Wvla].
Seems nvcc things there are parentheses around the type scalar_t but if I make a local typedef it avoids the warning.

packages/zoltan2/test/partition/fix4785.cpp
```diff
@@ -99,7 +99,18 @@ void test_no_weights(
   int &nFail
 )
 {
-  typedef Zoltan2::BasicUserTypes<myScalar_t, myLocalId_t, myGlobalId_t> myTypes;
+
+// Tried UVM off to see if it would speed things up - it did not.
+// TODO: Decide if we want to keep this or change it.
```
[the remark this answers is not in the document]
MicheldeMessieres: I set this back to the default node (UVM on).

packages/zoltan2/test/partition/fix4785.cpp
```diff
@@ -146,7 +157,17 @@ void test_weights(
   int &nFail
 )
 {
-  typedef Zoltan2::BasicUserTypes<myScalar_t, myLocalId_t, myGlobalId_t> myTypes;
+// Tried UVM off to see if it would speed things up - it did not.
+// TODO: Decide if we want to keep this or change it.
```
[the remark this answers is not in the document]
MicheldeMessieres: I set this back to the default node (UVM on).

packages/zoltan2/src/input/Zoltan2_Adapter.hpp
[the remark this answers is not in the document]
MicheldeMessieres: Currently we have MJ fixed to use Layout Left for the coordinates since we had decided we wanted to be able to use the MultiVector coordinates directly without a copy, which is currently the case. The only reason for LayoutLeft appearing in a few places in MJ is because of this. If we did decide to copy they could all go away, though we do take subviews of the coordinates since MJ likes to work with coords just on a single dimension and we have dimension as the 2nd parameter. I'm not sure how it would affect performance to change these things. Other adapters create a LayoutLeft coordinate view as well, but this is only for compatibility with MultiVector which currently determines MJ's layout type.

So we have this flow:
  MultiVector is LayoutLeft
  So we made MJ coords LayoutLeft
  So we made other adapter coords LayoutLeft

packages/zoltan2/src/util/Zoltan2_ImbalanceMetricsUtility.hpp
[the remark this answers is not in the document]
mhoemmen: Teuchos::SerialDenseMatrix has this issue too.  I don't think we need to worry about it, but I would prefer to see `std::unique_ptr` used here instead of explicit `new` and `delete`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -3569,7 +3567,7 @@ struct ReduceWeightsFunctor {
 
     Kokkos::parallel_reduce(
       Kokkos::TeamThreadRange(teamMember, begin, end),
-      [=] (const size_t ii, Zoltan2_MJArrayType<array_t>& threadSum) {
+      [=] (size_t ii, Zoltan2_MJArrayType<array_t>& threadSum) {
```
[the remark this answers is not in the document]
MicheldeMessieres: Working for OpenMP but not Cuda. It worked for Cuda until recently but some changes to Kokkos invalidated the internal handling of the array ptr. I've sent the Kokkos team some tests to demonstrate this issue. We had stopped using this for Cuda anyways due to performance issues so it was not blocking the current setup, but if we eventually did your suggestion of making the reduce path versus atomic path a parameter, then we'd need to resolve that Kokkos issue first.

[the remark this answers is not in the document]
mhoemmen: Does `Kokkos::Array` not work as an array type?  I'm a little bit uncomfortable with arrays of run-time length as reduction types, since intermediate reduction results take up CUDA shared memory.

bartlettroscoe:
FYI: This PR has merge conflicts so the PR tester can't run so the `AT: RETEST` label is doing nothing but spamming people every day.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -1595,15 +1595,15 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
...
-      (0, num_total_coords), KOKKOS_LAMBDA (mj_lno_t i) {
-    local_coordinate_permutations(i) = local_inital_adjList_output_adjlist(i);
-  });
+  {
+    auto local_coordinate_permutations = this->coordinate_permutations;
+    auto local_initial_adjList_output_adjlist = initial_adjList_output_adjlist;
+    Kokkos::parallel_for(
+      Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_lno_t>
+        (0, num_total_coords), KOKKOS_LAMBDA (mj_lno_t i) {
+      local_coordinate_permutations(i) = local_initial_adjList_output_adjlist(i);
```
mhoemmen: Why not just use `Kokkos::deep_copy`?  You're just copying a 1-D View into a 1-D View.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -2083,12 +2085,15 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
...
-    mj_part_t> (0, num_total_coords), KOKKOS_LAMBDA (mj_part_t i) {
-    inital_adjList_output_adjlist(i) = local_coordinate_permutations(i);
-  });
+  {
+    auto local_initial_adjList_output_adjlist = initial_adjList_output_adjlist;
+    auto local_coordinate_permutations = this->coordinate_permutations;
+    Kokkos::parallel_for(
+      Kokkos::RangePolicy<typename mj_node_t::execution_space,
+      mj_part_t> (0, num_total_coords), KOKKOS_LAMBDA (mj_part_t i) {
+      local_initial_adjList_output_adjlist(i) = local_coordinate_permutations(i);
```
mhoemmen: See above.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -2007,6 +2003,10 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
                 = part_size;
             });
 
+            auto local_new_coordinate_permutations =
+              this->new_coordinate_permutations;
+            auto local_coordinate_permutations =
+              this->coordinate_permutations;
```
mhoemmen: You don't need the `parallel_for` below; just use `Kokkos::deep_copy`.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Thanks for catching this oversight. Fixed.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Fixed.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Fixed.

Issue #5327, opened 2019-06-05
Unguarded SuperLU macro "Reduce" causes compilation error in Epetra_MultiVector.h


ndellingwood:
## Bug Report
@trilinos/stokhos @trilinos/muelu 

### Description
Attempting to build Trilinos VOTD develop branch on Waterman using the ATDM scripts resulted in build errors. The error itself pops up as an Epetra error in code that hasn't been touched in a long time while trying to compile a (ETI-generated?) muelu cpp file in stokhos.

Failure:
```
[5929/9067] Building CXX object packages/stokhos/src/CMakeFiles/stokhos_muelu_pce_cuda.dir/MueLu_AlgebraicPermutationStrategy_UQ_PCE_Cuda.cpp.o
FAILED: packages/stokhos/src/CMakeFiles/stokhos_muelu_pce_cuda.dir/MueLu_AlgebraicPermutationStrategy_UQ_PCE_Cuda.cpp.o

```

Error messages:
```
../../packages/epetra/src/Epetra_MultiVector.h(988): error: expected an identifier

../../packages/epetra/src/Epetra_MultiVector.h(988): error: expected a ")"

../../packages/epetra/src/Epetra_MultiVector.h(294): error: exception specification for virtual function "Epetra_MultiVector::~Epetra_MultiVector" is incompatible with that of overridden function "Epetra_BLAS::~Epetra_BLAS"

../../packages/epetra/src/Epetra_MultiVector.h(294): error: exception specification for virtual function "Epetra_MultiVector::~Epetra_MultiVector" is incompatible with that of overridden function "Epetra_CompObject::~Epetra_CompObject"

../../packages/epetra/src/Epetra_MultiVector.h(294): error: exception specification for virtual function "Epetra_MultiVector::~Epetra_MultiVector" is incompatible with that of overridden function "Epetra_DistObject::~Epetra_DistObject"

../../packages/epetra/src/Epetra_MultiVector.h(294): error: exception specification for virtual function "Epetra_MultiVector::~Epetra_MultiVector" is incompatible with that of overridden function "Epetra_SrcDistObject::~Epetra_SrcDistObject"

../../packages/epetra/src/Epetra_MultiVector.h(294): error: exception specification for virtual function "Epetra_MultiVector::~Epetra_MultiVector" is incompatible with that of overridden function "Epetra_Object::~Epetra_Object"

[...]

PR #5333, opened 2019-06-06
epetra: Fix for unguarded SuperLU macro

ndellingwood:
Fix for #5327

### DIFF
--- a/packages/epetra/src/Epetra_MultiVector.h
+++ b/packages/epetra/src/Epetra_MultiVector.h
@@ -985,7 +985,9 @@ class EPETRA_LIB_DLL_EXPORT Epetra_MultiVector: public Epetra_DistObject, public
   //@}
 
   // Expert-only function
+#ifdef Reduce
 #undef Reduce
+#endif
   int Reduce();
 
  protected:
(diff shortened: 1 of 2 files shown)


packages/epetra/src/Epetra_MultiVector.h
```diff
@@ -985,6 +985,7 @@ class EPETRA_LIB_DLL_EXPORT Epetra_MultiVector: public Epetra_DistObject, public
   //@}
 
   // Expert-only function
+#undef Reduce
```
mhoemmen: Can you `#undef` something that's not defined?  Perhaps we should do this:
```c++
#ifdef Reduce
#undef Reduce
#endif 
```

ndellingwood: yeah that's safer, I just copy/pasted in rush, thanks!

ndellingwood: Will update in a moment

ndellingwood: @mhoemmen updated, thanks!

ndellingwood:
Updated Epetra_IntMultiVector.h and Epetra_MultiVector.h with guards around `#undef Reduce` as well

jhux2:
@ndellingwood Thanks!

PR #5361, opened 2019-06-10
MueLu: fixing some issues in MakeQuasiRegionMatrix logic

lucbv:
@trilinos/muelu 

## Description
Two parts of the regionMG code are modified:
- `regionsPerGIDWithGhost` is now built using `rowMapPerGrp` instead of `AComp->getColMap()` which makes it independent of the discretization use to build `AComp`
- the construction of `quasiRegionMatrix` has been modified due to some logic error...

## Motivation and Context
This fixes a logic error which could potentially lead to bugs down the road when running more complex problems.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #5135 
* Part of 
* Composed of 

## How Has This Been Tested?
I have done a local build and local test of my modifications.

## Checklist

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionMatrix_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionMatrix_def.hpp
@@ -73,10 +73,10 @@ using Teuchos::Array;
  *
  */
 template<class LocalOrdinal, class GlobalOrdinal, class Node>
-std::vector<int> findCommonRegions(const GlobalOrdinal nodeA, ///< GID of first node
-                                   const GlobalOrdinal nodeB, ///< GID of second node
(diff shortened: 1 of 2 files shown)


lucbv:
@mayrmt, yes I think this will at least reduce our memory footprint but certainly also speed things up by removing the intermediate data.
There was also an issue with using the column map of the composite matrix. If you think of a 2D mesh with 4 ranks, assuming that rank 0 owns all its interfaces, it's column map will include indices that are not local to the region (these are the off process indices of interface node rows). This was creating a problem when I fixed regionsPerGIDWithGhosts. @rstumin and I have been pondering why things were working when regionsPerGIDWithGhosts was clearly wrong x)
I will merge this as is for now, I'll clean up the commented code once I'm back on Thursday.

PR #5366, opened 2019-06-11
Set python executable for cmake

prwolfe:
Not sure why it still uses /usr/bin even though this one is
first in the path. This allow the tests to run, but they still need to be added to the autotester list.

@trilinos/framework

## How Has This Been Tested?
I ran this by hand on my blade and will run this PR against the Jenkins jobs Trilinos_pullrequest_python_2 and Trilinos_pullrequest_python_3

### DIFF
--- a/cmake/std/sems/PullRequestPython3TestingEnv.sh
+++ b/cmake/std/sems/PullRequestPython3TestingEnv.sh
@@ -37,12 +37,9 @@ module load atdm-ninja_fortran/1.7.2
 # to one that has the proper sym-links from python -> python3
 module unload sems-python
 # module load sierra-python/3.6.3 - permissions do not allow this, but the execs are ok
-PATH=/projects/sierra/linux_rh7/install/Python/3.6..3/bin:${PATH}
+export PATH=/projects/sierra/linux_rh7/install/Python/3.6..3/bin:${PATH}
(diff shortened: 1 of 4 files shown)


prwolfe:
I had to add the paths explicitly as Trilinos does not have access to the Sierra modules, but does have access to the Sierra python installation.

cmake/std/sems/PullRequestPython3TestingEnv.sh
```diff
@@ -37,12 +37,9 @@ module load atdm-ninja_fortran/1.7.2
 # to one that has the proper sym-links from python -> python3
 module unload sems-python
 # module load sierra-python/3.6.3 - permissions do not allow this, but the execs are ok
-PATH=/projects/sierra/linux_rh7/install/Python/3.6..3/bin:${PATH}
+export PATH=/projects/sierra/linux_rh7/install/Python/3.6..3/bin:${PATH}
```
mhoemmen: @prwolfe Did you mean to have two dots here between 6 and 3?

prwolfe: No I did not - this will fail.

prwolfe: I wonder how I got it to run on my blade? Must have had both in place.

PR #5379, opened 2019-06-13
Delete clean directory

prwolfe:
@trilinos/framework 

## Description
change to do os.unlink instead of make clean

## Motivation and Context
Resolves issues with generated files that should no longer be present and makes support on Ride easier.

Marking as WIP until the current issues with 5228 clear


prwolfe:
Wrong branch - grrrr

PR #5385, opened 2019-06-14
Fix some types with doc for new ATDM spack-rhel env

bartlettroscoe:


bartlettroscoe:
@fryeguy52, you see if you see any other typos and fix them by adding new commits?  I think you can do that directly on GitHub.

PR #5228, opened 2019-05-22
Panzer dofmgr remove LO/GO template params 2

rppawlo:
DO NOT AUTOMERGE!

This commit removes all GO/LO template parameters from panzer. This allows for a single GO in tpetra and still supporting the type I and type II stacks in the same build.

PR #5388, opened 2019-06-15
MiniEM: Re-enable Epetra tests

cgcgcg:
@trilinos/panzer 

## Description
The MiniEM - Epetra tests were disabled in PR #5228. I made the necessary changes to re-enable them.

### DIFF
--- a/packages/panzer/mini-em/example/BlockPrec/CMakeLists.txt
+++ b/packages/panzer/mini-em/example/BlockPrec/CMakeLists.txt
@@ -63,23 +63,24 @@ TRIBITS_ADD_TEST(
    NUM_MPI_PROCS 4
    )
 
-IF(NOT Kokkos_ENABLE_Cuda AND NOT Kokkos_ENABLE_OpenMP AND MueLu_ENABLE_Epetra)
+IF(NOT Kokkos_ENABLE_Cuda AND NOT Kokkos_ENABLE_OpenMP AND MueLu_ENABLE_Epetra AND NOT HAVE_TPETRA_INST_INT_LONG_LONG AND HAVE_TPETRA_INST_INT_INT)
+ # We want GlobalOrdinal=int. Panzer picks long long, if it is available, and otherwise Tpetra's default.
 
-#  TRIBITS_ADD_TEST(
-#    BlockPrec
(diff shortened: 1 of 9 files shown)


packages/panzer/mini-em/example/BlockPrec/CMakeLists.txt
```diff
@@ -63,23 +63,23 @@ TRIBITS_ADD_TEST(
...
-#    )
-
-#  TRIBITS_ADD_TEST(
-#    BlockPrec
-#    NAME "MiniEM-BlockPrec_Augmentation_Epetra"
-#    ARGS "--solver=Augmentation --numTimeSteps=1 --linAlgebra=Epetra"
-#    COMM mpi
-#    NUM_MPI_PROCS 4
-#    )
+IF(NOT Kokkos_ENABLE_Cuda AND NOT Kokkos_ENABLE_OpenMP AND MueLu_ENABLE_Epetra AND NOT HAVE_TPETRA_INST_INT_LONG_LONG)
```
mhoemmen: Wouldn't you need to say `AND HAVE_TPETRA_INST_INT_INT` instead?  Tpetra has other options for GlobalOrdinal, like `long`.

cgcgcg: True. Is there a way to check from cmake what Tpetra's default GlobalOrdinal will be?

mhoemmen: > Is there a way to check from cmake what Tpetra's default GlobalOrdinal will be?

Alas no, not yet, but that will change :-) In this case, since you're running Epetra through MueLu (presumably), you really do need `GlobalOrdinal=int`.

cgcgcg:
@rppawlo I did. I still need to check for `NOT HAVE_TPETRA_INST_INT_LONG_LONG` since otherwise Panzer picks `long long` over `int`.

rppawlo:
Thanks! So does that mean that you will no longer be able to jump between muelu-epetra and muelu-tpetra at runtime? You will need two separate builds now?

cgcgcg:
@rppawlo You will, provided that Tpetra was built with `GlobalOrdinal=int`. The Epetra code path never worked for `GlobalOrdinal=long long`.

PR #5400, opened 2019-06-18
Thyra+Tpetra: Inner products on Cuda

cgcgcg:
@trilinos/thyra @trilinos/tpetra 

## Description
This deals with some of the performance issues seen on Cuda:
In every inner product call, Thyra created one temporary Tpetra map and one temporary Tpetra vector. These are now being cached.

### DIFF
--- a/packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_decl.hpp
+++ b/packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_decl.hpp
@@ -46,6 +46,7 @@
 
 #include "Thyra_SpmdVectorSpaceDefaultBase.hpp"
 #include "Tpetra_Map.hpp"
+#include "Tpetra_MultiVector.hpp"
 
 
 namespace Thyra {
@@ -105,6 +106,12 @@ protected:
   RCP<MultiVectorBase<Scalar> >
   createMembers(int numMembers) const;
 
+  RCP<MultiVectorBase<Scalar> >
+  createMembersView( const RTOpPack::SubMultiVectorView<Scalar> &raw_mv ) const;
+
+  RCP<const MultiVectorBase<Scalar> >
+  createMembersView( const RTOpPack::ConstSubMultiVectorView<Scalar> &raw_mv ) const;
+
   //@}
 
 public:
@@ -125,6 +132,8 @@ private:
   // Private data members
 
   RCP<const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> > tpetraMap_;
+  mutable RCP<TpetraVectorSpace<Scalar,LocalOrdinal,GlobalOrdinal,Node> > tpetraDomainSpace_;
+  mutable RCP<Tpetra::MultiVector<Scalar,LocalOrdinal,GlobalOrdinal,Node> > tpetraMV_;
   // The only reason Thyra needs this comm_ object is because Thyra
   // uses Ordinal as the Comm template parameter, while Tpetra uses
   // int.  Ordinal is some 64-bit type, which doesn't make any sense,
@@ -142,7 +151,7 @@ private:
 }; // end class TpetraVectorSpace
 
 
-/** \brief Nonmember consturctor that creats a serial vector space.
+/** \brief Nonmember constructor that creats a serial vector space.
  *
  * \relates TpetraVectorSpace
  */
--- a/packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_def.hpp
+++ b/packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_def.hpp
@@ -49,7 +49,7 @@
 #include "Thyra_TpetraVector.hpp"
 #include "Thyra_TpetraMultiVector.hpp"
 #include "Thyra_TpetraEuclideanScalarProd.hpp"
-
+#include "Tpetra_Details_StaticView.hpp"
 
 namespace Thyra {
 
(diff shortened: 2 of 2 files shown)


packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3996,18 +3997,10 @@ namespace Tpetra {
...
-      const char ctransA = (transA == Teuchos::NO_TRANS ? 'N' :
-                            (transA == Teuchos::TRANS ? 'T' : 'C'));
-      const char ctransB = (transB == Teuchos::NO_TRANS ? 'N' :
-                            (transB == Teuchos::TRANS ? 'T' : 'C'));
       const impl_scalar_type alpha_IST (alpha);
 
-      ProfilingRegion regionGemm ("Tpetra::MV::multiply-call-gemm");
-      KokkosBlas::gemm (&ctransA, &ctransB, alpha_IST, A_sub, B_sub,
-                        beta_local, C_sub);
+      typedef typename dual_view_type::t_dev XMV;
```
mhoemmen: Prefer `using XMV = typename dual_view_type::t_dev;` for new code.

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
```diff
@@ -0,0 +1,149 @@
...
+  const char ctransB = (transB == NO_TRANS ? 'N' :
+                        (transB == TRANS ? 'T' : 'C'));
+
+  const size_t zero = 0;
+
+  auto C_sub = subview (C_lcl,
+                        std::make_pair (zero, C_lcl.extent(0)),
+                        std::make_pair (zero, C_lcl.extent(1)));
+  ::Tpetra::Details::ProfilingRegion regionGemm ("Tpetra::MV::multiply-call-gemm");
+  KokkosBlas::gemm(&ctransA, &ctransB, alpha, A_lcl, B_lcl, beta, C_sub);
```
mhoemmen: 1. This is the case where `dot_type != value_type`.  How is it possible for `C_lcl` to have the same
type as `A_lcl` and `B_lcl`?

2. `XMV::value_type` could be either const or nonconst.  It's fine to use for `alpha` and `beta`, where constness doesn't matter, but just remember that if you want the nonconst type, you'll need to use `XMV::non_const_value_type`.

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+#include "Kokkos_ArithTraits.hpp"
+#include "KokkosBlas.hpp"
+#include "Tpetra_Details_Profiling.hpp"
+#include "Kokkos_InnerProductSpaceTraits.hpp"
+
+namespace Tpetra {
+namespace Details {
+
+template<class XMV,
```
mhoemmen: It's not clear to me why you need the `enable_if`, since both versions appear to invoke the same version of `KokkosBlas::gemm`.

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
```diff
@@ -0,0 +1,149 @@
...
+    KokkosBlas::dot(C_sub, A_lcl, B_lcl);
+    if (alpha != oneVT)
+      KokkosBlas::scal (C_sub, alpha, C_sub);
+  }
+  else {
+    auto C_sub = subview (C_lcl,
+                          std::make_pair (zero, C_lcl.extent(0)),
+                          std::make_pair (zero, C_lcl.extent(1)));
+    ::Tpetra::Details::ProfilingRegion regionGemm ("Tpetra::MV::multiply-call-gemm");
+    KokkosBlas::gemm(&ctransA, &ctransB, alpha, A_lcl, B_lcl, beta, C_sub);
```
mhoemmen: Do you see what I mean?  Both the "dot type same as value type" and "dot type not same as value type" branches invoke `KokkosBlas::gemm` with arguments of the same types.

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
cgcgcg: Maybe I mocked things up, cause I don't understand `enable_if` very well. What I want is that if `value_type==dot_type`, we enable the specialized `dot` for the case the one of the multivectors is actually a vector. Otherwise we always use `gemm`. So the `gemm` call is identical in both cases.

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
cgcgcg: `MV::multiply` multiplies two MVs into `this`. So `C_lcl` is the memory of a MV, not `dot_type`. Is that what you are getting at?

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
mhoemmen: @cgcgcg I took another look at Stokhos, and I think it's OK not to worry about the `dot_type != value_type` case.  Stokhos has specializations of `Belos::MultiVecTraits` that just extract out the MultiVectors' local data, flatten the Views into `dot_type`, and do the local multiply on the "flat views."

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
mhoemmen: Ah, never mind, sorry.

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
mhoemmen: Oh wait, I think I see what you mean.  Why not just have special cases that call `KokkosBlas::gemv`?  That would avoid the whole `dot_type != value_type` issue.

cgcgcg: @mhoemmen Let me give try that out..

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3996,18 +3997,39 @@ namespace Tpetra {
...
+      const char ctransA = (transA == NO_TRANS ? 'N' :
+                            (transA == TRANS ? 'T' : 'C'));
+      const char ctransB = (transB == NO_TRANS ? 'N' :
+                            (transB == TRANS ? 'T' : 'C'));
       const impl_scalar_type alpha_IST (alpha);
 
-      ProfilingRegion regionGemm ("Tpetra::MV::multiply-call-gemm");
-      KokkosBlas::gemm (&ctransA, &ctransB, alpha_IST, A_sub, B_sub,
-                        beta_local, C_sub);
+      const size_t zero = 0;
```
mhoemmen: This is harmless, but you don't need it.  `Kokkos::subview(X, ALL (), 0)` works perfectly fine.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3996,18 +3997,39 @@ namespace Tpetra {
...
+
+      const bool cond = (transA != NO_TRANS && transB == NO_TRANS);
```
mhoemmen: Do you actually need this `cond` special case?  `gemv` should work just fine regardless.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3996,18 +3997,39 @@ namespace Tpetra {
...
+      // The first two cases are specializations for inner products between a vector and a multivector
+      if (cond && A_sub.extent(1) == 1) {
+        auto A_sub_sub = subview (A_sub, ALL (), zero);
+        auto C_sub = subview (C_lcl, zero, ALL ());
+        ::Tpetra::Details::ProfilingRegion regionGemm ("Tpetra::MV::multiply-call-gemv");
+        KokkosBlas::gemv(&ctransA, alpha_IST, B_sub, A_sub_sub, beta_local, C_sub);
+
+        if (transA == CONJ_TRANS)
```
mhoemmen: Doesn't `gemv`'s `ctransA` argument take care of this?  Why do you need another conjugation?

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
mhoemmen: Never mind, I see why -- `gemv` can't transpose or conjugate-transpose B.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
cgcgcg: We want to compute A^H * B, where A is a vector. We compute B^H * A and then complex conjugate the result.

mhoemmen: @cgcgcg Why not just compute `B^T * A`?  That way you don't have to double-conjugate.

packages/tpetra/core/src/Tpetra_Details_lclMultiply.hpp
mhoemmen: I think we figured this out -- the issue was that one branch wanted to use `dot` to do a GEMV, but that only works if `dot_type == value_type`.  I think we should just use GEMV in that case (see my comments above).

packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_decl.hpp
```diff
@@ -46,6 +46,7 @@
 
 #include "Thyra_SpmdVectorSpaceDefaultBase.hpp"
 #include "Tpetra_Map.hpp"
+#include "Tpetra_MultiVector.hpp"
```
mhoemmen: It looks like you only need forward declarations for Map and MultiVector in this file.  You can get those from `Tpetra_Map_fwd.hpp` and `Tpetra_MultiVector_fwd.hpp`.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
cgcgcg: @mhoemmen That does not seem to be correct, no? Cause that way, A does not get conjugated.

packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_decl.hpp
cgcgcg: Ok, will do.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
mhoemmen: Could we only do the conjugation for complex scalar types?  I'd rather not launch another kernel in the common case.

cgcgcg: Shouldn't that already be taken care of by `transA == CONJ_TRANS`? Or should we check whether `CONJ_TRANS` has been passed, even though everything is real-valued?

mhoemmen: It's legal to pass `CONJ_TRANS` for real-valued types.  In that case, it means the same thing as `TRANS`.

bartlettroscoe: > It's legal to pass CONJ_TRANS for real-valued types. In that case, it means the same thing as TRANS.

That is correct.  If you think about it, a real number is just a complex number with the imaginary part being 0.

bartlettroscoe: BTW, Thyra has a function for this use case [`real_trans(transp)`](https://trilinos.org/docs/dev/packages/thyra/doc/html/group__Thyra__Op__Vec__fundamental__interfaces__code__grp.html#gae7d8a1c53cf64f6008b62eceb3b93299).  Makes it easy to write operators that are only for real numbers.

mhoemmen: @bartlettroscoe wrote:
> That is correct. If you think about it, a real number is just a complex number with the imaginary part being 0.

Fun fact: `std::conj(T)` works for real `T`, but always returns a complex number.

cgcgcg:
I removed the commit that allowed using `gemv` instead of `gemm`, since there appears to be a bug in the `gemv`. I'll open a new PR for that once that's fixed.

packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_def.hpp
```diff
@@ -114,6 +114,107 @@ TpetraVectorSpace<Scalar,LocalOrdinal,GlobalOrdinal,Node>::createMembers(int num
...
+  TEUCHOS_TEST_FOR_EXCEPT( raw_mv.subDim() != this->dim() );
+#endif
+  // Create a multi-vector
+  RCP< MultiVectorBase<Scalar> > mv;
+  if (!tpetraMap_->isDistributed()) {
+    if (tpetraMV_.is_null() || (tpetraMV_->getNumVectors() != (size_t)raw_mv.numSubCols())) {
+      using IST = typename Tpetra::MultiVector<Scalar,LocalOrdinal,GlobalOrdinal,Node>::impl_scalar_type;
+      using DT = typename Tpetra::MultiVector<Scalar,LocalOrdinal,GlobalOrdinal,Node>::device_type;
+      auto dv = ::Tpetra::Details::getStatic2dDualView<IST, DT> (tpetraMap_->getGlobalNumElements(), raw_mv.numSubCols());
+      tpetraMV_ = Teuchos::rcp(new Tpetra::MultiVector<Scalar,LocalOrdinal,GlobalOrdinal,Node>(tpetraMap_, dv));
```
mhoemmen: Suppose that you create two different locally replicated multivectors with this function, and the second multivector has more columns.  The second call to `getStatic2dDualView` may then reallocate.  This will invalidate the first multivector's storage.  Is that OK here?  If not, then we need to write a proper memory pool and not just use this static thing.

cgcgcg: It's only ok as long as we don't try to use two of these temporary MV objects at the same time. Currently, we are not doing that anywhere. I could add a status variable to the MV that is checked before a `CopyMultiVectorViewBack` object is built, and reset once the copy-out happens. That way, we could throw, in case anyone ever tries to do this.

mhoemmen: @cgcgcg wrote:
> I could add a status variable to the MV that is checked before a `CopyMultiVectorViewBack` object is built, and reset once the copy-out happens. That way, we could throw, in case anyone ever tries to do this.

That would be super useful -- if you have time, could you please do that?  Thanks! :-D

packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_def.hpp
```diff
@@ -114,6 +114,139 @@ TpetraVectorSpace<Scalar,LocalOrdinal,GlobalOrdinal,Node>::createMembers(int num
...
+  mv->acquireDetachedView(Range1D(),Range1D(),&smv);
+  RTOpPack::assign_entries<Scalar>(
+    Ptr<const RTOpPack::SubMultiVectorView<Scalar> >(Teuchos::outArg(smv)),
+    raw_mv
+    );
+  mv->commitDetachedView(&smv);
+  // Setup smart pointer to multi-vector to copy view back out just before multi-vector is destroyed
+  Teuchos::set_extra_data(
+    // We create a duplicate of the RCP, otherwise the ref count does not go to zero.
+    Teuchos::rcp(new CopyMultiVectorViewBack<Scalar,LocalOrdinal,GlobalOrdinal,Node>(Teuchos::rcpFromRef(*mv),raw_mv)),
```
mhoemmen: If you're calling `rcpFromRef` all the time, you could just have CopyMultiVectorViewBack take a reference.

mhoemmen:
I see a lot of errors like the following.  Does this mean that your check caught something?
```
p=0: *** Caught standard std::exception of type 'std::invalid_argument' :
 
  .../packages/teuchos/core/src/Teuchos_RCPNode.cpp:231:
  
  Throw number = 3
  
  Throw test that evaluated to true: (itr != extra_data_map_->end() && force_unique)
  
  Error, the type:name pair 'bool:inUse' already exists and force_unique==true!
```

cgcgcg:
@mhoemmen I don't think so, the check should give a different error message. The `set_extra_data` does not seem to work if the (key, type) pair already exists. It's weird, cause it does pass on my system :-/

packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_def.hpp
```diff
@@ -114,6 +114,139 @@ TpetraVectorSpace<Scalar,LocalOrdinal,GlobalOrdinal,Node>::createMembers(int num
 }
 
 
+template<class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
+class CopyMultiVectorViewBack {
+public:
+  CopyMultiVectorViewBack( RCP<MultiVectorBase<Scalar> > mv, const RTOpPack::SubMultiVectorView<Scalar>  &raw_mv )
```
mhoemmen: Now that I think about it, wouldn't it be bad to store `mv_` as an RCP?  CopyMultiVectorViewBack's destructor only kicks in whenever the RCP's reference count goes to zero.  You've avoided trouble below by creating the input RCP as nonowning (`rcpFromRef`), but in general this is unwise, I think.

packages/thyra/adapters/tpetra/src/Thyra_TpetraVectorSpace_def.hpp
```diff
@@ -114,6 +114,139 @@ TpetraVectorSpace<Scalar,LocalOrdinal,GlobalOrdinal,Node>::createMembers(int num
...
+#ifdef TEUCHOS_DEBUG
+  TEUCHOS_TEST_FOR_EXCEPT( raw_mv.subDim() != this->dim() );
+#endif
+  // Create a multi-vector
+  RCP< MultiVectorBase<Scalar> > mv;
+  if (!tpetraMap_->isDistributed()) {
+    if (tpetraMV_.is_null() || (tpetraMV_->getNumVectors() != size_t (raw_mv.numSubCols()))) {
+      if (!tpetraMV_.is_null())
+        // The MV is already allocated. If we are still using it, then very bad things can happen.
+        TEUCHOS_TEST_FOR_EXCEPTION(Teuchos::get_extra_data<bool>(tpetraMV_,"inUse"),
```
mhoemmen: We could just create a new one for now.  Caching is only an optimization.

mhoemmen:
@bartlettroscoe wrote:
> I think @mhoemmen had previously signed off on this PR (looking for a Tpetra point of view) so I will go ahead and approve this.

There are ways this could be improved, but I don't want to be obstructionist.  We can always go back and fix it.

mhoemmen:
I just thought of something -- if you're running on a single MPI process, every Tpetra::Map is "not distributed."  This means that this PR might be creating every single MultiVector to have the same storage.  That can't be good.

mhoemmen:
@cgcgcg I'm a little worried -- I put a throw in the destructor of CopyMultiVectorViewBack, but all of Thyra's tests passed.  I'll try Stratimikos and Thyra.

cgcgcg:
@mhoemmen We should be bugging out with an error message if several multivectors want to use the cached Tpetra MV at the same time. Please remember that this is only used in the case that we want to wrap a `SubMultiVectorView` into a `MultiVectorBase`. I got to this situation from Belos, where a `SerialDenseMatrix` is used this way.

PR #5402, opened 2019-06-18
MueLu: triple product defaults, Maxwell test

cgcgcg:
@trilinos/muelu 

## Description
- Use triple product for unsmoothed hierarchies, since it's always faster.
- Second attempt at disabling failing test with ML style list. 
  (Why do we have builds with ML enabled and Epetra disabled?)


cgcgcg:
@trilinos/framework Looks like the autotester is broken..

jwillenbring:
@cgcgcg 

Here is the output for the one [failing test](https://testing.sandia.gov/cdash/testDetails.php?test=81880923&build=5217220)

cgcgcg:
@jwillenbring Seems to have been an intermittent failure. As you can see, the autotester launched 3 times without reporting anything.

PR #5409, opened 2019-06-19
Tpetra::CrsMatrix: Fix #5407

### LINKED ISSUE #5407: Tpetra::CrsMatrix::sumIntoLocalValues is slow
## Bug Report
@trilinos/tpetra 

### Description

`Tpetra::CrsMatrix::sumIntoLocalValues` (and presumably all of the sumInto, replace, and transform methods) are much slower than they used to be.  See discussion [here](https://github.com/trilinos/Trilinos/pull/5403).

mhoemmen:
@trilinos/tpetra @tjfulle 

## Description

Quick fix for #5407.  I just want to see if all the tests pass, so that I can apply this fix quickly to the version of Trilinos in Sierra.  We'll do some more extensive performance testing later.  I also have a restructuring of `sumIntoLocalValues` in progress that should reduce overhead by avoiding `getRowInfo` and the various "get a view" methods, but I want to test that more carefully.

## Related Issues

* Closes #5407

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
@@ -2473,19 +2473,6 @@ namespace Tpetra {
                           const impl_scalar_type newVals[],
                           const LocalOrdinal numElts) const
   {
-    if (graph.getProfileType() == StaticProfile)
-    {
-      Teuchos::ArrayView<const LocalOrdinal> indsT(inds, numElts);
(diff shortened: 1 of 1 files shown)


mhoemmen:
My attempt to refactor `sumIntoLocalValues` caused some `Tpetra::FECrsMatrix` tests to fail, so I think I'll leave it for now and just go with the simple fix.

mhoemmen:
Thanks @kyungjoo-kim !

Issue #4385, opened 2019-02-12
Tpetra::BlockCrsMatrix: Deprecate "experimental" headers & namespace


mhoemmen:
@trilinos/tpetra @kddevin (as discussed in the weekly Tpetra meeting today)

`Tpetra::BlockCrsMatrix` is implemented in the `Tpetra_Experimental_BlockCrsMatrix*.hpp` headers and the `Tpetra::Experimental` namespace.  However, at least one production code uses BlockCrsMatrix.  This means that we should not call it "experimental" any more.  We should deprecate the "experimental" headers and namespace, and put this class into the `Tpetra` namespace.  (In fact, it's already there; see below.)

"De-experimental-ing" BlockCrsMatrix could happen in N steps:

  1. The `Tpetra_BlockCrsMatrix*.hpp` headers already exist, and they import BlockCrsMatrix into the `Tpetra` namespace.  Production codes should just include those header files for now.
  2. Tpetra could then put the implementation of BlockCrsMatrix in the existing `Tpetra_BlockCrsMatrix*.hpp` headers, make sure that Tpetra's CMake logic auto-generates `Tpetra_BlockCrsMatrix.cpp` (instead of `Tpetra_Experimental_BlockCrsMatrix.cpp`).  Each `Tpetra_Experimental_BlockCrsMatrix*.hpp` header would then just include the corresponding `Tpetra_BlockCrsMatrix*.hpp` header, and import the class into the `Tpetra::Experimental` namespace.
  3. It's possible to deprecate header files, so we can then mark the `Tpetra_Experimental_BlockCrsMatrix*.hpp` headers as deprecated.

@kyungjoo-kim brings up a good concern, that perhaps we should evaluate BlockCrsMatrix's performance and fix any major issues we find, before "releasing" it into the Tpetra namespace.  While the `Tpetra_BlockCrsMatrix*.hpp` headers do already import it into the Tpetra namespace, the above actions would constitute a more formal release.  The application that uses BlockCrsMatrix depends mainly on a specialized solver for performance, not on sparse matrix-vector multiply (`BlockCrsMatrix::apply`).

kyungjoo-kim:
This is done by PR #5417

PR #5417, opened 2019-06-19
Remove experimental from Tpetra block crs/multivector/vector.

kyungjoo-kim:
@trilinos/tpetra 

## Description

This PR removes all experimental name space from block crs/multivector/vector.

## Motivation and Context

#4385


kyungjoo-kim:
@mhoemmen What you described is the current statue of the code. Tpetra_Block* headers include Tpetra_Experimental_Block* headers. I think that we want to remove the experimental namespace completely. The name space "Experimental" means that we can remove it anytime we want. Even if some downstream codes actually use the Experimental namespace for block matrix operations, it should be modified in the downstream codes. For now, the only downstream package is SPARC and I don't think that they use the experimental headers in their code. Just in case, would you make sure of it as I don't have access the code ?

kddevin:
The "trilinos blessed" way to deprecate header files:  add something like the following to the header file, changing the claps message to a useful message.

```
#if !defined(TRILINOS_HIDE_DEPRECATED_HEADER_WARNINGS)
#ifdef __GNUC__
#  warning "The package Claps is deprecated in April 2019; it will be removed from Trilinos in May 2019."
#endif
#endif
```

mhoemmen:
@kyungjoo-kim wrote:
> What you described is the current status of the code.

Not exactly.  The current status of the code is as follows:

1. The actual declarations and definitions of the classes live in the `Tpetra::Experimental` namespace, in `Tpetra_Experimental_*` headers.
2. Each `Tpetra_*` header includes the corresponding `Tpetra_Experimental_*` header, and imports the corresponding class into the Tpetra namespace.

I had understood the end goal as follows:

1. Move the actual declarations and definitions of the classes into the Tpetra namespace, and into `Tpetra_*` headers.
2. Mark the `Tpetra_Experimental_*` headers as deprecated, using the procedure that Karen outlines above.
3. Have each `Tpetra_Experimental_*` header import the corresponding class into the `Tpetra::Experimental` namespace.

This is "deprecation without removal."

mhoemmen:
Removal breaks backwards compatibility.  It doesn't matter whether or not SPARC uses the headers or namespace; what matters is that the hypothetical user's code isn't broken.

mhoemmen:
SPARC does use the "Experimental" namespace and headers still.  I'm fixing that now, but that's not the point.  The point is that removing the headers breaks backwards compatibility.

kyungjoo-kim:
Okay. I understand the concern. I will just leave the experimental headers.

mhoemmen:
> I will just leave the experimental headers.

The experimental headers would also need to be changed, so that they just include the non-experimental headers and import their classes into the `Tpetra::Experimental` namespace.

I don't mean to get in the way -- I'm just worried about backwards compatibility.

kyungjoo-kim:
I left all experiemantal headers and BlockExpNamespace.cpp for testing.

mhoemmen:
SPAR-580 (reported by @bartlettroscoe ) suggests that we may have missed something with the BlockCrs writer.  We'll take a look.

mhoemmen:
SPAR-580 was easy to fix in SPARC.  Your changes were just fine :-D .

bartlettroscoe:
> SPAR-580 was easy to fix in SPARC. Your changes were just fine :-D .

But not strictly backward compatible (or SPARC was not including header files and using types correctly).  But I guess since SPARC was using "Experimental" types, there was no expectation for backwards compatibility anyway so this is all good..

kyungjoo-kim:
@mhoemmen Thank you for fixing the problem in SPARC. As I do not have an access to SPARC, I do not even get such a notice of the failure. In Tpetra BlockExpNamespace, I tested it compiled with the experimental namespace but that does not seem to be enough.

mhoemmen:
@kyungjoo-kim Don't worry about it :-)  I don't think you did anything wrong at all.  SPARC should have been including the new header files anyway; they've existed for a long time.  Doing that made the build work for both old and new versions of Trilinos.

bartlettroscoe:
@mhoemmen said:

>  @kyungjoo-kim Don't worry about it :-) I don't think you did anything wrong at all

Yes, I agree.  Given that SPARC was pointing to "Experimental" stuff in the first place, there is no expectation of maintaining backward compatibility.  Just my opinion.

PR #5419, opened 2019-06-19
Tpetra:  changes to reduce the number of deprecated warnings issued 

kddevin:
@trilinos/tpetra 
#5117 

This PR reduces the number of uses of the enum value DynamicProfile in Tpetra source code.  It will reduce the number of warnings issued by Tpetra during compilation.  

Note that not all deprecation warnings in Tpetra and the rest of the software stack can be eliminated.  With explicit instantiation and deprecated code enabled, deprecated code gets compiled and warnings get issued, even if the code is not called anywhere.  Also, because DynamicProfile deprecation changes both a default behavior and a keyword, it is difficult to remove warnings due to the default behavior change.  Finally, since DynamicProfile is currently the Tpetra default, we need to retain its use in testing; tests exercising DynamicProfile should not be removed just to silence warnings.

## How Has This Been Tested?
ATDM scripts on linux workstation
```
MPI_EXEC=`which mpiexec`


cmake \
  -GNinja \
  -DMPI_EXEC=${MPI_EXEC} \
  -DTrilinos_ENABLE_TESTS=ON -DTrilinos_ENABLE_Teuchos=ON \
  -DTrilinos_ENABLE_Tpetra=ON \
  -DTrilinos_ENABLE_Belos=ON \
  -DTrilinos_ENABLE_Tempus=ON \
.. |& tee OUTPUT.CMAKE
```

### DIFF
--- a/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
+++ b/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
@@ -133,9 +133,9 @@ namespace Tpetra {
 
 #ifdef TPETRA_ENABLE_DEPRECATED_CODE
   TPETRA_DEPRECATED const ProfileType DynamicProfile = ProfileType(StaticProfile+1);
-#define TPETRA_DEFAULT_PROFILE_TYPE DynamicProfile
+#define TPETRA_DEFAULT_PROFILE_TYPE Tpetra::ProfileType(Tpetra::StaticProfile+1)  // DynamicProfile
 #else
-#define TPETRA_DEFAULT_PROFILE_TYPE StaticProfile
+#define TPETRA_DEFAULT_PROFILE_TYPE Tpetra::StaticProfile
 #endif
 
   /*! Optimize storage option */
--- a/packages/tpetra/core/test/CrsMatrix/CrsMatrix_UnitTests.cpp
+++ b/packages/tpetra/core/test/CrsMatrix/CrsMatrix_UnitTests.cpp
@@ -113,13 +113,13 @@ namespace { // (anonymous)
     {
       RCPMap map  = createContigMapWithNode<LO,GO,Node>(INVALID,numLocal,comm);
       MV mv(map,1);
-      zero = rcp( new MAT(map,0,Tpetra::TPETRA_DEFAULT_PROFILE_TYPE) );
+      zero = rcp( new MAT(map,0,TPETRA_DEFAULT_PROFILE_TYPE) );
       TEST_THROW(zero->apply(mv,mv), std::runtime_error);
 #   if defined(HAVE_TPETRA_THROW_EFFICIENCY_WARNINGS)
       // throw exception because we required increased allocation
       TEST_THROW(zero->insertGlobalValues(map->getMinGlobalIndex(),tuple<GO>(0),tuple<Scalar>(ST::one())), std::runtime_error);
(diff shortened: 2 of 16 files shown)


packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
```diff
@@ -133,9 +133,9 @@ namespace Tpetra {
 
 #ifdef TPETRA_ENABLE_DEPRECATED_CODE
   TPETRA_DEPRECATED const ProfileType DynamicProfile = ProfileType(StaticProfile+1);
-#define TPETRA_DEFAULT_PROFILE_TYPE DynamicProfile
+#define TPETRA_DEFAULT_PROFILE_TYPE Tpetra::ProfileType(Tpetra::StaticProfile+1)  // DynamicProfile
```
mhoemmen: Adding 1 to an enum value might generate a warning.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_UnitTests.cpp
```diff
@@ -113,13 +113,13 @@ namespace { // (anonymous)
     {
       RCPMap map  = createContigMapWithNode<LO,GO,Node>(INVALID,numLocal,comm);
       MV mv(map,1);
-      zero = rcp( new MAT(map,0,Tpetra::TPETRA_DEFAULT_PROFILE_TYPE) );
+      zero = rcp( new MAT(map,0,TPETRA_DEFAULT_PROFILE_TYPE) );
```
mhoemmen: If you're using the default, you could just remove the argument and rely on the default value.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_UnitTests3.cpp
```diff
@@ -422,9 +422,6 @@ inline void tupleToArray(Array<T> &arr, const tuple &tup)
 
     // call fillComplete()
     out << "Call fillComplete on the matrix" << endl;
-#ifdef TPETRA_ENABLE_DEPRECATED_CODE
-    TEST_EQUALITY_CONST( A.getProfileType() == Tpetra::DynamicProfile, true );
```
mhoemmen: Why just delete this one, but test with `TPETRA_DEFAULT_PROFILE_TYPE` above?

packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
kddevin: Then line 135 will produce a warning as well.  :)
I did not see any warnings from these lines in my builds.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_UnitTests.cpp
kddevin: True; I was just doing minimal editing to make the compiler happy.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_UnitTests3.cpp
kddevin: This test seemed minimally important, and would go away with the deprecated code anyway.

kddevin:
Thanks for the review, @mhoemmen

PR #5420, opened 2019-06-19
Error out if trying to set Kokkos Pthreads backend (TRIL-272)

bartlettroscoe:
Kokkos is going to longer support the Pthreads backend (see [TRIL-272](https://sems-atlassian-son.sandia.gov/jira/browse/TRIL-272)).  So
this commit adds checks if the user tries to set the 'pthread' keyword and
errors out if they do.

I also added some error handling so that the atdm/utils/set_build_options.sh
can raise an error and abort and that will abort the atdm/load-env.sh script.


ibaned:
To clarify, Kokkos is not dropping all support for Pthreads, we are simply dropping responsibility for making sure that Trilinos works well on the Pthreads backend. Non-Trilinos-based users of Kokkos may continue to rely on it.

bartlettroscoe:
@ibaned said:

> To clarify, Kokkos is not dropping all support for Pthreads, we are simply dropping responsibility for making sure that Trilinos works well on the Pthreads backend. Non-Trilinos-based users of Kokkos may continue to rely on it.

Thanks. Given that this impacts just the ATDM configurations, we don't likely need to make that distinction here.

PR #5422, opened 2019-06-20
Phalanx: Users can now control layouts for any Field 

rppawlo:
Also has the start of view allocation tracking but it is not complete.


rppawlo:
@etphipp - had to fix a cuda issue. Could you reapprove?

etphipp:
Done.

PR #5426, opened 2019-06-21
Ifpack2:  revisions to support static profile 

kddevin:
@trilinos/ifpack2 

## Description
Tpetra is deprecating DynamicProfile.  These changes allow Ifpack2 to work with StaticProfile.
The changes update Containers and ILUT.  ILUK is not yet done and is not included in this pull request.

This code discovers the nonzeros needed in ILUT as before, but rather than insert them immediately, it stores them in Teuchos Arrays.  Once all nonzeros are discovered, it creates L and U with StaticProfile and the number of entries per row.  Then it inserts the nonzeros.

These changes also update several tests and test helpers in the Ifpack2 test directory.

#5065 

## How Has This Been Tested?

ATDM Script:  All Ifpack2 tests pass with Tpetra_ENABLE_DEPRECATED_CODE=ON
```
MPI_EXEC=`which mpiexec`
cmake \
  -GNinja \
  -DMPI_EXEC=${MPI_EXEC} \
  -DTrilinos_ENABLE_TESTS=ON  \
  -DTrilinos_ENABLE_Ifpack2=ON \
.. |& tee OUTPUT.CMAKE
```

Not all tests pass yet with Tpetra_ENABLE_DEPRECATED_CODE=OFF, but more pass than before.

## Checklist

### DIFF
--- a/packages/ifpack2/src/Ifpack2_ILUT_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_ILUT_def.hpp
@@ -467,13 +467,6 @@ void ILUT<MatrixType>::compute ()
     L_ = rcp (new crs_matrix_type (A_local_->getRowMap (), A_local_->getColMap (), 0));
     U_ = rcp (new crs_matrix_type (A_local_->getRowMap (), A_local_->getColMap (), 0));
 
-    // CGB: note, this caching approach may not be necessary anymore
-    // We will store ArrayView objects that are views of the rows of U, so that
-    // we don't have to repeatedly retrieve the view for each row. These will
-    // be populated row by row as the factorization proceeds.
-    Array<ArrayView<const local_ordinal_type> > Uindices (myNumRows);
-    Array<ArrayView<const scalar_type> >       Ucoefs (myNumRows);
-
     // If this macro is defined, files containing the L and U factors
     // will be written. DON'T CHECK IN THE CODE WITH THIS MACRO ENABLED!!!
     // #define IFPACK2_WRITE_FACTORS
@@ -499,8 +492,10 @@ void ILUT<MatrixType>::compute ()
 
     Array<scalar_type> InvDiagU (myNumRows, zero);
 
-    Array<local_ordinal_type> tmp_idx;
-    Array<scalar_type> tmpv;
+    Array<Array<local_ordinal_type> > L_tmp_idx(myNumRows);
+    Array<Array<scalar_type> > L_tmpv(myNumRows);
+    Array<Array<local_ordinal_type> > U_tmp_idx(myNumRows);
+    Array<Array<scalar_type> > U_tmpv(myNumRows);
 
     enum { UNUSED, ORIG, FILL };
     local_ordinal_type max_col = myNumRows;
@@ -612,8 +607,8 @@ void ILUT<MatrixType>::compute ()
 
         /* Reduce current row */
 
-        ArrayView<const local_ordinal_type>& ColIndicesU = Uindices[row_k];
-        ArrayView<const scalar_type>& ColValuesU = Ucoefs[row_k];
+        ArrayView<local_ordinal_type> ColIndicesU = U_tmp_idx[row_k]();
+        ArrayView<scalar_type> ColValuesU = U_tmpv[row_k]();
         size_type ColNnzU = ColIndicesU.size();
 
         for(size_type j=0; j<ColNnzU; ++j) {
@@ -645,16 +640,16 @@ void ILUT<MatrixType>::compute ()
       //   first, the original entries from the L section of A:
       for (size_type i = 0; i < ColIndicesA.size (); ++i) {
         if (ColIndicesA[i] < row_i) {
-          tmp_idx.push_back(ColIndicesA[i]);
-          tmpv.push_back(cur_row[ColIndicesA[i]]);
+          L_tmp_idx[row_i].push_back(ColIndicesA[i]);
+          L_tmpv[row_i].push_back(cur_row[ColIndicesA[i]]);
           pattern[ColIndicesA[i]] = UNUSED;
         }
       }
 
       //   next, the L entries resulting from fill:
       for (size_type j = 0; j < L_vals_heaplen; ++j) {
-        tmp_idx.push_back(L_vals_heap[j]);
-        tmpv.push_back(cur_row[L_vals_heap[j]]);
+        L_tmp_idx[row_i].push_back(L_vals_heap[j]);
+        L_tmpv[row_i].push_back(cur_row[L_vals_heap[j]]);
         pattern[L_vals_heap[j]] = UNUSED;
       }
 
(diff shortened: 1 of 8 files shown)


packages/ifpack2/src/Ifpack2_ILUT_def.hpp
```diff
@@ -499,8 +492,10 @@ void ILUT<MatrixType>::compute ()
 
     Array<scalar_type> InvDiagU (myNumRows, zero);
 
-    Array<local_ordinal_type> tmp_idx;
-    Array<scalar_type> tmpv;
+    Array<Array<local_ordinal_type> > L_tmp_idx(myNumRows);
```
mhoemmen: Given that ILUT already fills the matrix by rows, it seems like the next logical step would be to get rid of this intermediate data structure and do a count-allocate-fill thing.  Of course it's good to know whether these changes can make it through all the tests first.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestTriDiSolver.cpp
```diff
@@ -125,7 +125,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(TriDiSolver, LapackComparison, ScalarType, Loc
 
   // The matrix will be block diagonal, so we can use the row Map as
   // the column Map.
-  RCP<crs_matrix_type> A = rcp(new crs_matrix_type (rowMap, colMap, (size_t) 0));
+  RCP<crs_matrix_type> A = rcp(new crs_matrix_type (rowMap, colMap, (size_t) 3));
```
mhoemmen: Prefer C++ - style casts: `size_t (3)`.

packages/ifpack2/src/Ifpack2_ILUT_def.hpp
```diff
@@ -704,41 +700,62 @@ void ILUT<MatrixType>::compute ()
...
+    }
+
+    U_ = rcp (new crs_matrix_type (A_local_->getRowMap(), A_local_->getColMap(), 
+                                   nnzPerRow()), Tpetra::StaticProfile);
+
+    for (local_ordinal_type row_i = 0 ; row_i < myNumRows ; ++row_i) {
+      U_->insertLocalValues (row_i, U_tmp_idx[row_i](), U_tmpv[row_i]());
+    }
+
+    U_->fillComplete(); // FIXME (mfh 03 Apr 2013) Do we need domain and range Map?
```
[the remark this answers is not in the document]
mhoemmen: Fixing that will be _hard_ btw.  The associated Bugzilla issue discusses the problem.

kddevin:
Thanks, @mhoemmen 
I started down a count-allocate-fill path, but we need to store at least U_idx_tmp[num_rows] even for the counting phase (lines 610, 617).

I agree that the memory will increase temporarily, but how would the runtime slow down?  If we did two passes (one to count, and one to fill), I could see that the runtime would slow down.  But as it is, we are still doing only one pass.

In our separate email chain on this topic, you said, "What we could do is change the preconditioner to use a different data structure for creation, then copy into crs."  That is what I was doing in this PR.

mhoemmen:
@kddevin wrote:
> ... but how would the runtime slow down?

We would have to benchmark it.  It may even use less memory: DynamicProfile uses nearly the same representation that you're using, and if the StaticProfile upper bound on the number of entries in each row is exact, then `fillLocalGraphAndMatrix` won't copy and pack the matrix (since it's already packed).  That makes me not so worried about this.

packages/ifpack2/src/Ifpack2_ILUT_def.hpp
```diff
@@ -663,26 +658,27 @@ void ILUT<MatrixType>::compute ()
...
       }
 #endif
 
-      tmp_idx.clear();
-      tmpv.clear();
 
       // Pick out the diagonal element, store its reciprocal.
       if (cur_row[row_i] == zero) {
-        std::cerr << "Ifpack2::ILUT::Compute: zero pivot encountered! Replacing with rownorm and continuing...(You may need to set the parameter 'fact: absolute threshold'.)" << std::endl;
+        std::cerr << "Ifpack2::ILUT::Compute: zero pivot encountered! "
```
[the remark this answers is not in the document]
mhoemmen: There's already an issue open for this: https://github.com/trilinos/Trilinos/issues/2256

packages/ifpack2/src/Ifpack2_SparseContainer_decl.hpp
```diff
@@ -289,6 +289,27 @@ class SparseContainer : public Container<MatrixType> {
...
+  ///  \param Indices,            [in]  Storage for row indices of inputMatrix
+  ///  \param Values,             [in]  Storage for row values of inputMatrix
+  ///  \param Indices_insert,     [out] Indices to be inserted for this row
+  ///  \param Values_insert,      [out] Value associated with the index
+  ///  \param num_entries_found   [out] number of Indices to be inserted
+  void findRowIndicesAndCounts(
+    int blockIndex,
+    local_ordinal_type rowIndex,
+    const Teuchos::ArrayView<const local_ordinal_type> &localRows,
+    Teuchos::Array<local_ordinal_type> &Indices,
```
mhoemmen: "In" arguments shouldn't be passed in by nonconst reference.  It looks like you could just pass these in as `ArrayView`, since you don't need to resize them.

packages/ifpack2/src/Ifpack2_SparseContainer_def.hpp
```diff
@@ -141,13 +204,39 @@ void SparseContainer<MatrixType,InverseType>::initialize ()
   // local matrices to use for solves.
   diagBlocks_.reserve(this->numBlocks_);
   Inverses_.reserve(this->numBlocks_);
+
+  const size_t maxNumEntriesInRow = this->inputMatrix_->getNodeMaxNumRowEntries();
+  Teuchos::Array<scalar_type> Values;
+  Teuchos::Array<local_ordinal_type> Indices;
+  Teuchos::Array<InverseScalar> Values_insert;
+  Teuchos::Array<InverseGlobalOrdinal> Indices_insert;
+  Values.resize (maxNumEntriesInRow);
```
mhoemmen: The constructor of `Teuchos::Array`, just like `std::vector`, takes an integer size argument.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestRILUK.cpp
```diff
@@ -99,7 +99,7 @@ static Teuchos::RCP<Ifpack2::RILUK<Tpetra::RowMatrix<Scalar,LocalOrdinal,GlobalO
   //Construct a nondiagonal matrix.  It's not tridiagonal because of processor boundaries.
   const Scalar one = Teuchos::ScalarTraits<Scalar>::one();
   const Scalar two = one + one;
-  RCP<crs_matrix_type > A = Teuchos::rcp(new crs_matrix_type(map, 1));
+  RCP<crs_matrix_type > A = Teuchos::rcp(new crs_matrix_type(map, 3));
```
mhoemmen: Ifpack2's tests probably don't exercise actual fill, since they always use the same tridiagonal matrix.

packages/ifpack2/src/Ifpack2_ILUT_def.hpp
```diff
@@ -467,13 +467,6 @@ void ILUT<MatrixType>::compute ()
     L_ = rcp (new crs_matrix_type (A_local_->getRowMap (), A_local_->getColMap (), 0));
```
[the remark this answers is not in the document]
kddevin: Done.

packages/ifpack2/src/Ifpack2_ILUT_def.hpp
kddevin: See comments below.

packages/ifpack2/src/Ifpack2_SparseContainer_decl.hpp
kddevin: Done.

packages/ifpack2/src/Ifpack2_SparseContainer_def.hpp
kddevin: Done.

kddevin:
@mhoemmen I implemented the changes you requested, and will wait for re-testing.  Thanks for the review.

packages/ifpack2/src/Ifpack2_SparseContainer_def.hpp
```diff
@@ -124,6 +124,70 @@ void SparseContainer<MatrixType,InverseType>::setParameters(const Teuchos::Param
...
+    //    on-process row and column indices are the same.
+    // 2. All off-process indices in the column Map of the input
+    //    matrix occur after that initial set.
+    if(static_cast<size_t> (localCol) >= MatrixInNumRows)
+      continue;
+
+    // for local column IDs, look for each ID in the list
+    // of columns hosted by this object
+    InverseLocalOrdinal jj = INVALID;
+    for(local_ordinal_type kk = 0; kk < this->blockRows_[blockIndex]; kk++)
```
mhoemmen: In the future, we could consider using `std::find` for this, but not for now : - ) .

Issue #4757, opened 2019-03-28
Tpetra::DistObject: Evaluate host-pinned buffers for CUDA + MPI


mhoemmen:
@trilinos/tpetra 

`Tpetra::DistObject` currently offers two options for CUDA + MPI:

  1. Give CudaSpace buffers to MPI
  2. Give HostSpace buffers to MPI

(DistObject lets subclasses pack and unpack wherever they like, as long as they update the sync state correctly.)

We want to evaluate a third option: Give `CudaHostPinnedSpace` buffers to MPI.  Host-pinned memory behaves like CudaUVMSpace, in the sense that both host and device can access it.  However, MPI sees host-pinned memory as host memory.  This means two things:

  1. MPI need not be CUDA aware in order to access host-pinned memory, yet we can still pack and unpack on device.
  2. We don't have to worry about CUDA-aware MPI being slow.

The latter is important, since we've observed some "CUDA-aware MPI"  implementations being slow in practice.

Host-pinned memory has a high allocation cost.  It may make sense to start with the static View allocation functions in PR #4734.  We can't use those without further work, because each DistObject instance will need its own pack and unpack buffers.  This may call for a simple memory pool, and for changes to DistObject and/or subclasses so that they only hold buffer allocations while communication is active.

## Definition of Done
- [ ] Write a Tpetra test that prototypes use of CudaHostPinnedSpace for communication.
- [ ] Write a benchmark to compare performance of point-to-point communication with CudaHostPinnedSpace vs. CudaSpace communication buffers.
- [ ] If CudaHostPinnedSpace pays off, change DistObject to use it for communication buffers.
- [ ] Evaluate performance with a Tpetra benchmark.
- [ ] If performance is good, deploy solution in Tpetra.

## Related Issues

* Is blocked by #4734

mhoemmen:
I have some benchmark results for MPI pairwise messages between 2 GPUs (thus, 2 MPI processes total) on waterman.

Here's how I ran the benchmark:

  1. Check out the TpetraStaticView5 branch from https://github.com/mhoemmen/Trilinos.
  2. I used commit c37236bc026f78cfcb5f137928f5e156dfe8f6eb.
  3. Build Tpetra examples.  I used the SPARC waterman CMake options as a start.
  4. In the build directory, look in `packages/tpetra/core/example/advanced/Benchmarks/` for the `TpetraCore_commBuffers.exe` executable.
  5. `bsub -n 4 -Is -q rhel7W`
[...]

PR #5438, opened 2019-06-25
Tpetra: Fix build for buffer_memory_space = CudaHostPinnedSpace (but don't change it)

mhoemmen:
@trilinos/tpetra 

## Description

Currently, Tpetra uses `CudaSpace` as the Kokkos memory space for MPI communication buffers.  We want to try using `CudaHostPinnedSpace` instead, and compare performance of the two options.  (See #4757.)  Trivially changing the space (via the `buffer_memory_space` typedef that appears in a few places) does not build.  This PR fixes the build for that case.

NOTE: This PR does NOT change `buffer_memory_space` to `CudaHostPinnedSpace`.  It just makes sure that Tpetra can build with that change.  If you want to try `buffer_memory_space=CudaSpace`, just drop the last commit.

## Motivation and Context

ATDM performance on GPU clusters.

## Related Issues

* Part of #4757

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
@@ -61,48 +61,47 @@ namespace Details {
 namespace DefaultTypes {
   //! Default value of Scalar template parameter.
 #if defined(HAVE_TPETRA_INST_DOUBLE)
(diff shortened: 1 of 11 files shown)


mhoemmen:
CUDA build was perfectly fine, yay!

mhoemmen:
@trilinos/tpetra yo

PR #5445, opened 2019-06-27
Tpetra: Fix logic used in all Reduce

jjellio:
The prior logic had conditions based on whether the view was Cuda UVM or
not.  If UVM, you always end up creating 2 views and performance 2
copies. This unneeded.  Moreover, the logic for whether the input views
were contiguous was backwards (it trigged the code path for non-contig
cases when the views were actually contiguous!).

This fix enables a 2-3x speedup for small payload allreduces.

@trilinos/tpetra

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_allReduceView.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_allReduceView.hpp
@@ -54,13 +54,17 @@
 
 namespace { // (anonymous)
 
-// We can't assume C++14 yet, so we can't have templated constants.
+// helper for detecting views that have Cuda memory
+// Technically, UVM should not be here, but it seems we
+// treat UVM as CudaSpace, so I have left it. With Cuda >= 8 on Linux
+// UVM is handled by page faults in the Kernel, so it should always
+// be addressable.
 template<class ViewType>
-struct view_is_cuda_uvm {
+struct view_uses_cuda_spaces {
   static constexpr bool value =
 #ifdef KOKKOS_ENABLE_CUDA
-    std::is_same<typename ViewType::memory_space,
-                 Kokkos::CudaUVMSpace>::value;
+      std::is_same<typename ViewType::memory_space, Kokkos::CudaSpace>::value
+   || std::is_same<typename ViewType::memory_space, Kokkos::CudaUVMSpace>::value;
 #else
     false;
 #endif // KOKKOS_ENABLE_CUDA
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_Details_allReduceView.hpp
```diff
@@ -182,17 +182,11 @@ allReduceView (const OutputViewType& output,
     return;
   }
 
-  // We've had some experience that some MPI implementations do not
-  // perform well with UVM allocations.
-  const bool assumeMpiCanAccessBuffers =
-    ! view_is_cuda_uvm<OutputViewType>::value &&
-    ! view_is_cuda_uvm<InputViewType>::value &&
-    ::Tpetra::Details::Behavior::assumeMpiIsCudaAware ();
```
mhoemmen: What happens if we leave in `assumeMpiIsCudaAware` and take out the "not UVM" requirement?  Does it still work?  I'd rather not break the case where MPI really can't access CUDA buffers.

mhoemmen:
I'd rather that we not break the "MPI is not CUDA aware" case -- could we keep that perhaps?

jjellio:
@mhoemmen Are you saying that if using cuda, the user should expect every allReduce to entail 2 allocs + 2 copies in addition to the Reduction? That should be Opt-in, not mandated. That literally doubles or triples the cost. At best, we can cache the memory so you end up with 2 copies. Still that could easily mean two kernel launches per all reduce. That seems a bit extreme. My studies have shown a 10x improvement for Cuda using the patch attached.

mhoemmen:
@jjellio wrote:
> Are you saying that if using cuda, the user should expect every allReduce to entail 2 allocs + 2 copies in addition to the Reduction?

It sounded like you were saying that the main issue was whether the buffers are UVM.  I'm perfectly fine with letting MPI talk to UVM buffers.  My question is whether we could keep the `assumeMpiIsCudaAware` check.  If MPI was not built to be CUDA aware, then this PR could cause segfaults.

packages/tpetra/core/src/Tpetra_Details_allReduceView.hpp
jjellio: See final commit

jjellio:
New patch:

1.  Correct contig logic
2.  Remove UVM specific logic triggering the contig case
3.  Correct Cuda aware MPI logic

Tested on white, this allows about 2x faster reductions

packages/tpetra/core/src/Tpetra_Details_allReduceView.hpp
```diff
@@ -54,13 +54,17 @@
...
 
-// We can't assume C++14 yet, so we can't have templated constants.
+// helper for detecting views that have Cuda memory
+// Technically, UVM should not be here, but it seems we
+// treat UVM as CudaSpace, so I have left it. With Cuda >= 8 on Linux
+// UVM is handled by page faults in the Kernel, so it should always
+// be addressable.
 template<class ViewType>
-struct view_is_cuda_uvm {
+struct view_uses_cuda_spaces {
```
mhoemmen: We keep rewriting this "is a CUDA thing" class.  I filed a Kokkos [feature request](https://github.com/kokkos/kokkos/issues/2196).

packages/tpetra/core/src/Tpetra_Details_allReduceView.hpp
```diff
@@ -182,17 +186,29 @@ allReduceView (const OutputViewType& output,
     return;
   }
 
-  // We've had some experience that some MPI implementations do not
-  // perform well with UVM allocations.
-  const bool assumeMpiCanAccessBuffers =
-    ! view_is_cuda_uvm<OutputViewType>::value &&
-    ! view_is_cuda_uvm<InputViewType>::value &&
-    ::Tpetra::Details::Behavior::assumeMpiIsCudaAware ();
+  // we must esnure MPI can handle the pointers we pass it
```
mhoemmen: "ensure"

PR #5460, opened 2019-07-01
Expand contracts to help give stack traces

ibaned:
This is entirely a prototype designed to
help Trilinos developers debug spurious
failures. See kokkos/kokkos#2173

@bartlettroscoe @mhoemmen 

## Description

This is not considered ready for merging into Trilinos. I haven't tested Trilinos itself, just Kokkos with this change. It creates a Kokkos API for registering a contract violation handler, and then it treats CUDA API error code handling using the contract violation handler. A user may register a contract violation handler which calls Teuchos' stack trace functionality. This PR does not include a contract violation handler with good stack trace functionality, only the registration mechanism and the guarantee that most Kokkos CUDA failures will go through the violation handler.

## Motivation and Context

per kokkos/kokkos#2173, Trilinos developers were eager to have a way to get stack traces when certain tests fail at certain points in Kokkos. This branch is a starting point to having that happen.

## How Has This Been Tested?
Kokkos spot-check was run. Literally no Trilinos testing has been done.

### DIFF
--- a/packages/kokkos/core/src/impl/Kokkos_Error.cpp
+++ b/packages/kokkos/core/src/impl/Kokkos_Error.cpp
@@ -56,6 +56,28 @@
 
 namespace Kokkos {
 namespace Impl {
+static void (*violation_handler_pointer)(violation_info const&) = default_violation_handler;
+}
+
+void set_violation_handler(void (*new_handler)(violation_info const&)) {
+  violation_handler_pointer = new_handler;
+}
+
+namespace Impl {
+
+void default_violation_handler(violation_info const& info) {
+  fprintf(stderr, "contract violation in %s:%d\nfunction %s\n%s\n",
+      info.file_name, info.line, info.function_name, info.comment);
+  std::stringstream string_stream;
+  traceback_callstack(string_stream);
+  std::string s = string_stream.str();
+  fprintf(stderr, "%s\n", s.c_str());
+  std::abort();
+}
+
+void call_host_violation_handler(violation_info const& info) {
+  (*violation_handler_pointer)(info);
+}
 
 void host_abort( const char * const message )
 {
--- a/packages/kokkos/core/src/Cuda/Kokkos_Cuda_Error.hpp
+++ b/packages/kokkos/core/src/Cuda/Kokkos_Cuda_Error.hpp
@@ -51,15 +51,15 @@ namespace Kokkos { namespace Impl {
 
 void cuda_device_synchronize();
 
-void cuda_internal_error_throw( cudaError e , const char * name, const char * file = NULL, const int line = 0 );
+void cuda_internal_error_throw( cudaError e , const char * name, const char * file, char const* function, const int line );
 
-inline void cuda_internal_safe_call( cudaError e , const char * name, const char * file = NULL, const int line = 0)
+inline void cuda_internal_safe_call( cudaError e , const char * name, const char * file, char const* function, const int line)
 {
-  if ( cudaSuccess != e ) { cuda_internal_error_throw( e , name, file, line ); }
+  if ( cudaSuccess != e ) { cuda_internal_error_throw( e , name, file, function, line ); }
 }
 
 #define CUDA_SAFE_CALL( call )  \
-	Kokkos::Impl::cuda_internal_safe_call( call , #call, __FILE__, __LINE__ )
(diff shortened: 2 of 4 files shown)


ibaned:
Just lack of testing, and of higher-level approval by @crtrott

packages/kokkos/core/src/impl/Kokkos_Error.cpp
```diff
@@ -56,6 +56,28 @@
...
+static void (*violation_handler_pointer)(violation_info const&) = default_violation_handler;
+}
+
+void set_violation_handler(void (*new_handler)(violation_info const&)) {
+  violation_handler_pointer = new_handler;
+}
+
+namespace Impl {
+
+void default_violation_handler(violation_info const& info) {
```
bartlettroscoe: So to be safe, we should override this in Teuchos and have it call MPI_Abort()?

packages/kokkos/core/src/impl/Kokkos_Error.cpp
```diff
@@ -56,6 +56,28 @@
...
+  fprintf(stderr, "contract violation in %s:%d\nfunction %s\n%s\n",
+      info.file_name, info.line, info.function_name, info.comment);
+  std::stringstream string_stream;
+  traceback_callstack(string_stream);
+  std::string s = string_stream.str();
+  fprintf(stderr, "%s\n", s.c_str());
```
bartlettroscoe: Is this function run on the device or always on the host?  If so, whe using fprint() and not std::cerr?

packages/kokkos/core/src/impl/Kokkos_Error.cpp
ibaned: Yes, if one is concerned about `std::abort` or `std::terminate` causing an MPI hang, then `MPI_Abort` should be called. Since Kokkos is determined to ignore the existence of MPI, this must be done in a handler outside of Kokkos. I personally have never seen this issue happen, but won't argue against it.

packages/kokkos/core/src/impl/Kokkos_Error.cpp
ibaned: (1) This function always runs on the host, (2) I use `fprintf` when reporting errors in an MPI-parallel setting because it is usually better at keeping messages from different ranks from getting mixed up with one another. It is possible that a single `operator<<` call to `std::cerr` might give the same atomicity, but I did this out of habit.

mhoemmen: @ibaned wrote:
> It is possible that a single `operator<<` call to `std::cerr` might give the same atomicity, but I did this out of habit.

I have good luck first building a string (with `ostringstream`) and then writing it once to the `ostream`.  I'm perfectly fine with using `fprintf` here for this reason.

mhoemmen:
See [my comments here](https://github.com/kokkos/kokkos/pull/2199#issuecomment-507445002).

packages/kokkos/core/src/impl/Kokkos_Error.cpp
bartlettroscoe: Yup, that is the trick.  You build a `std::ostringstream` first then output the entire string in one `std::cerr <<` operator invocation.

mhoemmen:
@ibaned Check out this stack tracing thing: https://github.com/trilinos/Trilinos/pull/5493

I'll make a Kokkos PR -- that's a Trilinos one for now.

packages/kokkos/core/src/impl/Kokkos_Error.cpp
```diff
@@ -56,6 +56,28 @@
...
+  violation_handler_pointer = new_handler;
+}
+
+namespace Impl {
+
+void default_violation_handler(violation_info const& info) {
+  fprintf(stderr, "contract violation in %s:%d\nfunction %s\n%s\n",
+      info.file_name, info.line, info.function_name, info.comment);
+  std::stringstream string_stream;
+  traceback_callstack(string_stream);
```
mhoemmen: Regarding `traceback_callback`'s [current implementation](https://github.com/kokkos/kokkos/blob/d3a941925cbfb71785d8ea68259123ed52d3f9da/core/src/impl/Kokkos_Error.cpp#L153), I had thought that `dlopen` only works for shared object files.  Not sure if it works for statically build executables etc.

william76:
@ibaned 

PR #5466, opened 2019-07-02
MueLu region driver multiple dofs per node

### LINKED ISSUE #4304: MueLu: structured region driver
@trilinos/muelu 
@rstumin @mayrmt @pohm01 

## Expectations
This new driver constructs a problem on a perfectly structured grid and provides necessary data for HHG region matrix construction

## Motivation and Context
This driver will provide a testing ground for new HHG capabilities on a simple mesh.

lucbv:
@trilinos/muelu 

## Description
Adding support for system of equations in the structured region driver

## Motivation and Context
This allows to test the robustness of the region multigrid implementation.

## Related Issues

* Closes #4304 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## How Has This Been Tested?
This was tested locally and the tests are running fine.

## Checklist

### DIFF
--- a/packages/galeri/src-xpetra/Galeri_StencilProblems.hpp
+++ b/packages/galeri/src-xpetra/Galeri_StencilProblems.hpp
@@ -57,13 +57,28 @@ namespace Galeri {
 
   namespace Xpetra {
 
-    // =============================================  Laplace1D  =============================================
-    template <typename Scalar, typename LocalOrdinal, typename GlobalOrdinal, typename Map, typename Matrix, typename MultiVector>
-    class Laplace1DProblem : public Problem<Map,Matrix,MultiVector> {
+    // =============================================  Scalar Problem =========================================
+    template <typename Map, typename Matrix, typename MultiVector>
+    class ScalarProblem : public Problem<Map,Matrix,MultiVector> {
     public:
       using RealValuedMultiVector = typename Problem<Map,Matrix,MultiVector>::RealValuedMultiVector;
 
-      Laplace1DProblem(Teuchos::ParameterList& list, const Teuchos::RCP<const Map>& map) : Problem<Map,Matrix,MultiVector>(list, map) { }
+      ScalarProblem(Teuchos::ParameterList& list, const Teuchos::RCP<const Map>& map) : Problem<Map,Matrix,MultiVector>(list, map) { }
+      Teuchos::RCP<MultiVector> BuildNullspace();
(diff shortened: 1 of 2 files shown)


lucbv:
@mayrmt @rstumin @pohm01 
I am able to run the structured region driver with `matrixType=Elasticity2D`.
I think it makes sense to incorporate these changes.

mayrmt:
@lucbv When transferring the matrix from composite to region format, we use nullspace information to fix values on the diagonal (cf. `MakeRegionMatrices()`). Right now, the nullspace vector is hard-coded to an all-ones-vector (i.e. nullspace for a single DOF per node problem). It's not clear, if and how this has to be changed to also properly work for multiple DOFs per node (especially in the case of elasticity, where we also have rotational nullspace vectors).

I'm ok with merging this for now, since the issue of rational nullspace vectors might be more tricky and a research topic on its own. 

I have removed the automerge label, so that you can still decide, whether to proceed with the merge or now. As I said, I'm ok with merging.

lucbv:
@mayrmt 
I think that we should be able to ask the initial MueLu hierarchy to generate our nullspace as long as we pass it appropriate coordinates, which we do for the linear interpolation operator.
Although I have not verified this yet...

mayrmt:
> I think that we should be able to ask the initial MueLu hierarchy to generate our nullspace as long as we pass it appropriate coordinates, which we do for the linear interpolation operator.
> Although I have not verified this yet...

I think, we don't have a MueLu hierarchy yet, when creating the region matrices. We first need the region matrices and then can create a region hierarchy.

lucbv:
@mayrmt you are correct, the nullspace is actually computed by Galeri using the `BuildNullspace()` method of the `Pr` object.
So for the driver I propose that we use the Galeri infrastructure to create the composite nullspace and then the region nullspace can be generated with the same logic used to generate the region coordinates.

lucbv:
@mayrmt 
I am finishing a bit of work on the structured/unstructured logic for the driver, I'll get to this afterwards.

lucbv:
@mayrmt I had to do a little more infrastructure work to implement the nullspace correctly for scalar problems.
It all works fine and provides a simple interface for the construction of a nullspace within driver examples.
I also did the modification in the structured region driver to have it use the galeri nullspace as a starting point for the construction of a region nullspace.

PR #5473, opened 2019-07-03
Fix broken code inside HAVE_TPETRA_MM_TIMINGS

DrBooom:
Code inside #ifdef HAVE_TPETRA_MM_TIMINGS is broken, and will not compile. 
There is not issue number.

### DIFF
--- a/packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
@@ -1313,7 +1313,7 @@ namespace Tpetra {
 #ifdef HAVE_TPETRA_MMM_TIMINGS
       std::string prefix_mmm = std::string("TpetraExt ") + label + std::string(": ");
       using Teuchos::TimeMonitor;
-      Teuchos::TimeMonitor MM = rcp(new TimeMonitor (*TimeMonitor::getNewTimer(prefix_mmm + std::string("PTAP local transpose"))));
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
```diff
@@ -1313,7 +1313,7 @@ namespace Tpetra {
 #ifdef HAVE_TPETRA_MMM_TIMINGS
       std::string prefix_mmm = std::string("TpetraExt ") + label + std::string(": ");
       using Teuchos::TimeMonitor;
-      Teuchos::TimeMonitor MM = rcp(new TimeMonitor (*TimeMonitor::getNewTimer(prefix_mmm + std::string("PTAP local transpose"))));
+      Teuchos::TimeMonitor MM = *new TimeMonitor (*TimeMonitor::getNewTimer(prefix_mmm + std::string("PTAP local transpose")));
```
mhoemmen: This is probably not what you want to do.  Unless your compiler does C++17 copy elision, this may create two TimeMonitor instances.  It's probably harmless but it likely breaks downstream code that wants to assign an RCP to MM.

DrBooom: There are no downstream in scope instances of assignment to MM.

mhoemmen: Phone discussion concluded that this might not be harmless.  Default constructor of TimeMonitor has an invalid reference, that would get used in the destructor.

PR #5481, opened 2019-07-04
Teuchos: Fix #5480

### LINKED ISSUE #5480: Teuchos::TimeMonitor: Delete default constructor
## Enhancement
@trilinos/teuchos 

`Teuchos::TimeMonitor`'s constructor takes and holds a reference to a `Teuchos::Time`, and `TimeMonitor`'s destructor stops the timer by calling an instance method on the reference.  `TimeMonitor` has an implicit default constructor, and `~TimeMonitor` has no way to know that the object was default constructed.  Thus, users are allowed to default-construct `TimeMonitor` and get themselves in trouble. 

We can fix this by adding `TimeMonitor() = delete;`  to the declaration of `TimeMonitor`.

See [discussion here](https://github.com/trilinos/Trilinos/pull/5473#discussion_r300179859).

mhoemmen:
@trilinos/teuchos @DrBooom 

## Description

Delete default constructors of `Teuchos::PerformanceMonitorBase` and `Teuchos::TimeMonitor`.  Default construction is unsafe, since it creates an invalid reference that the destructor then uses as if it were valid.

## Motivation and Context

Prevent `~TimeMonitor` from calling instance methods on an invalid timer reference.  See [discussion here](https://github.com/trilinos/Trilinos/pull/5473#discussion_r300179859).

## Related Issues

* Closes #5480 
* Related to #5473

### DIFF
--- a/packages/teuchos/comm/src/Teuchos_PerformanceMonitorBase.hpp
+++ b/packages/teuchos/comm/src/Teuchos_PerformanceMonitorBase.hpp
@@ -34,18 +34,15 @@
 // NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
-// Questions? Contact Michael A. Heroux (maherou@sandia.gov)
-//
 // ***********************************************************************
 // @HEADER
 
 #ifndef TEUCHOS_PERFORMANCEMONITORBASE_H
 #define TEUCHOS_PERFORMANCEMONITORBASE_H
 
-/*! \file Teuchos_PerformanceMonitorBase.hpp
-  \brief Provides common capabilities for collecting and reporting
-  performance data across processors
-*/
+/// \file Teuchos_PerformanceMonitorBase.hpp
+/// \brief Common capabilities for collecting and reporting
+///   performance data collectively across MPI processes.
 
 #include "Teuchos_ConfigDefs.hpp"
 #include "Teuchos_Array.hpp"
@@ -157,7 +154,7 @@ namespace Teuchos
   class PerformanceMonitorBase
   {
   public:
-    /** \brief Construct with a counter. */
+    //! Construct with a counter.
     PerformanceMonitorBase(T& counter_in, bool reset=false)
       : counter_(counter_in), isRecursiveCall_(counter_.isRunning())
     {
@@ -165,11 +162,14 @@ namespace Teuchos
       counter_.incrementNumCalls ();
     }
 
+    //! Default constructor is deleted, since it would be unsafe.
+    PerformanceMonitorBase () = delete;
+
     /// \brief Destructor.
     ///
     /// The destructor for the base class does nothing.  We provide a
     /// virtual destructor for memory safety of derived classes.
-    virtual ~PerformanceMonitorBase() {}
+    virtual ~PerformanceMonitorBase() = default;
 
     /** \brief Create a new counter with the specified name and add it
      *   to a global set of counters of this type.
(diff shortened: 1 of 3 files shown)


packages/teuchos/comm/src/Teuchos_PerformanceMonitorBase.hpp
```diff
@@ -157,19 +154,22 @@ namespace Teuchos
...
 
+    //! Default constructor is deleted, since it would be unsafe.
+    PerformanceMonitorBase () = delete;
+
     /// \brief Destructor.
     ///
     /// The destructor for the base class does nothing.  We provide a
     /// virtual destructor for memory safety of derived classes.
-    virtual ~PerformanceMonitorBase() {}
+    virtual ~PerformanceMonitorBase() = default;
```
bartlettroscoe: Why delete this?  What is the harm in having a virtual destructor with an empty implementation?

packages/teuchos/comm/src/Teuchos_TimeMonitor.cpp
```diff
@@ -1488,7 +1488,7 @@ namespace Teuchos {
     stackedTimer_ = t;
   }
 
-  const Teuchos::RCP<Teuchos::StackedTimer>&
+  Teuchos::RCP<Teuchos::StackedTimer>
```
bartlettroscoe: Yes.  Thank you!  See "return smart pointers by value" in section "5.8.5 Idioms for returning objects from C++ functions" in [Teuchos MM document](https://trilinos.org/docs/dev/packages/teuchos/doc/html/TeuchosMemoryManagementSAND.pdf).

packages/teuchos/comm/src/Teuchos_TimeMonitor.hpp
```diff
@@ -192,8 +192,11 @@ class TEUCHOSCOMM_LIB_DLL_EXPORT TimeMonitor :
   ///   Default behavior is not to reset the timer.
   TimeMonitor (Time& timer, bool reset=false);
 
+  //! Default constructor is deleted, since it would be unsafe.
+  TimeMonitor () = delete;
+
   //! Destructor: stops the timer.
-  ~TimeMonitor();
+  ~TimeMonitor() override;
```
bartlettroscoe: Should this not also have 'virtual' to make this more clear?

packages/teuchos/comm/src/Teuchos_PerformanceMonitorBase.hpp
mhoemmen: `= default` declares intent.  I actually should have deleted the two lines of documentation above this, since "virtual default destructor" is a typical idiom for an abstract base class.

Using `= default` instead of `{}` matters more for non-virtual destructors, since it is necessary for the class to be [trivially destructible](https://en.cppreference.com/w/cpp/language/destructor#Trivial_destructor).

packages/teuchos/comm/src/Teuchos_TimeMonitor.cpp
mhoemmen: @bartlettroscoe you're welcome :-)

packages/teuchos/comm/src/Teuchos_TimeMonitor.hpp
mhoemmen: [C++ Core Guidelines C.128](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rh-override) says: "Virtual functions should specify exactly one of `virtual`, `override`, or `final`."

Rationale is that specifying both `virtual` and `override` is redundant.  See the example [https://en.cppreference.com/w/cpp/language/virtual](here): "Then this function in the class `Derived` is also virtual (whether or not the keyword `virtual` is used in its declaration) and overrides `Base::vf` (whether or not the word `override` is used in its declaration)."

I realize now that C.128 says that `TimeMonitor`'s destructor should be declared neither `virtual` nor `override`.  It's harmless to leave `override` in place, though.  We do that a lot, and I suspect some tools or compiler flags will warn if child class destructors lack decoration.

mhoemmen:
Thanks Ross!

packages/teuchos/comm/src/Teuchos_PerformanceMonitorBase.hpp
bartlettroscoe: Got it.  Thanks!

PR #5493, opened 2019-07-09
WIP Kokkos: Add stack trace functions

mhoemmen:
@trilinos/kokkos 

## Description

THIS IS A WORK IN PROGRESS; PLEASE DO NOT MERGE.

Add functions to (Trilinos' version of) Kokkos that save and print stack traces, and can register an `std::terminate` handler that prints the last saved stack trace and calls a user-specified abort function (which could be `MPI_Abort`).  The functions use portable POSIX functions, except for symbol demangling.  The latter should work with GCC and Clang.

## Motivation and Context

It's hard to debug random `std::terminate` calls at scale due to uncaught exceptions that happen in Kokkos' parallel kernels.

## Related Issues

* Part of https://github.com/kokkos/kokkos/issues/2173

## How Has This Been Tested?

Locally, with AppleClang.


packages/kokkos/core/src/impl/Kokkos_stacktrace.cpp
```diff
@@ -0,0 +1,265 @@
...
+    }
+  }
+
+  // Make another pass to get the column lengths.
+  // Only demangle the column of functions.
+  std::vector<size_t> max_col_lens;
+  for (auto&& entry : traceback) {
+    size_t col_count = 0;
+    for_each_token (entry, [&] (const std::string& s, bool) {
+                             const size_t cur_col_len =
```
bartlettroscoe: Wow, that is some crazy indentation.  What so much?

mhoemmen:
@bartlettroscoe wrote:
> does this work? 

Yes, it works on AppleClang.

> Does this work on the target systems with CUDA?

I haven't gotten that far yet ;-) .  The idea is to integrate with https://github.com/trilinos/Trilinos/pull/5460 .

packages/kokkos/core/src/impl/Kokkos_stacktrace.cpp
mhoemmen: Emacs wanted it that way, I guess?  I can fix that.

bartlettroscoe: Yes, default email tab indentation is insane.  I wrote a much more sane indentation scheme for emails.  You can find it at:

https://github.com/trilinos/Trilinos/tree/master/packages/thyra/emacs

Should be a 5-min setup.

ibaned:
Thanks for the good work @mhoemmen !

@bartlettroscoe I think we are unlikely to extract good stack traces from within a CUDA kernel in the near term. The work we're doing now is towards getting a stack trace up to the `parallel_for` that failed an internal assert.

mhoemmen:
> The work we're doing now is towards getting a stack trace up to the `parallel_for` that failed an internal assert.

This is exactly what we want -- to give users some clue which kernel had a bug.  C++11 doesn't promise anything more about exceptions in thread-parallel code (other than `std::thread`).

bartlettroscoe:
> The work we're doing now is towards getting a stack trace up to the parallel_for that failed an internal assert.

That should be a good start.  How deep are the call stacks under a parallel_for() invocation in Trilinos?

nmhamster:
@bartlettroscoe / @mhoemmen - I've seen some fairly deep stacks in the past (around 20-25 stack entries), although this was from an application perspective.

bartlettroscoe:
@nmhamster said:

> I've seen some fairly deep stacks in the past (around 20-25 stack entries), although this was from an application perspective.

20-25 levels **below** the `parallel_for()` call?

nmhamster:
@bartlettroscoe - no 20-25 from the application until the `parallel_for` call. Seen around 5-10 within the call. Typically a lot less and mostly because of inlining.

mhoemmen:
@nmhamster wrote:
> Typically a lot less and mostly because of inlining.

The View bounds check exceptions that stirred up this whole discussion generally only get enabled in debug builds, btw.

william76:
@trilinos/kokkos 

Issue #4468, opened 2019-02-22
Epetra: Remove build warnings


ZUUL42:
@trilinos/epetra 

## Expectations

Epetra builds should not emit any warnings that will be promoted to errors once Werror is set in the GCC 7.2.0 automated build.
## Current Behavior

Currently Epetra has a number of warnings that need to be handled before we can set Werror for all packages.
A recent test build was performed with -Werror set for Epetra. The CDash report can be found [here](https://testing.sandia.gov/cdash/viewBuildError.php?buildid=4976007).
## Motivation, Context and Related Issues

Issue #3178 is working toward turning Warnings as Errors on for all packages.

ZUUL42:
If you'd like the PR to test with Werror set for Epetra, pull the commit from branch [Werror_Epetra](https://github.com/ZUUL42/Trilinos/tree/Werror_Epetra).

PR #5507, opened 2019-07-10
Fix warning-errors and set Werror for Epetra in GCC 7.2.0 PR build

ZUUL42:
@trilinos/framework 
@trilinos/epetra 

## Description

Fixed warning-errors and set `-Werror` for Epetra in the GCC 7.2.0 PR build.
Wrapped variables Werror sees as unused that are only used by asserts in `#ifndef NDEBUG`.
Wrapped or changed some `int` as `unsigned int` to fix the `always false [-Werror=strict-overflow]` Werrors.

## Motivation and Context

This completes #4468 and is work toward Issue #3178, eventually setting `-Werror` globally.

## How Has This Been Tested?

This was tested sucessfully running the GCC 7.2.0 PR Jenkins job.

## Checklist

### DIFF
--- a/packages/epetra/src/Epetra_Util.h
+++ b/packages/epetra/src/Epetra_Util.h
@@ -348,7 +348,7 @@ int Epetra_Util_insert_empty_positions(T*& array, int& usedLength,
     return(-1);
   }
 
-  if ((usedLength+numPositions) < allocatedLength) {
+  if (((unsigned int)usedLength+(unsigned int)numPositions) < (unsigned int)allocatedLength) {
     for(int i=usedLength-1; i>=insertOffset; --i) {
       array[i+numPositions] = array[i];
     }
--- a/packages/epetra/test/VbrMatrix/cxx_main.cpp
+++ b/packages/epetra/test/VbrMatrix/cxx_main.cpp
@@ -2211,9 +2211,9 @@ int checkVbrMatrixOptimizedGraph(Epetra_Comm& comm, bool verbose)
   Epetra_SerialDenseMatrix C(3, 3);
   Epetra_SerialDenseMatrix L(3, 3);
   Epetra_SerialDenseMatrix R(3, 3);
-  EPETRA_TEST_ERR( ! 3 == C.LDA(), ierr );
-  EPETRA_TEST_ERR( ! 3 == L.LDA(), ierr );
-  EPETRA_TEST_ERR( ! 3 == R.LDA(), ierr );
+  EPETRA_TEST_ERR( (! 3) == C.LDA(), ierr );
+  EPETRA_TEST_ERR( (! 3) == L.LDA(), ierr );
+  EPETRA_TEST_ERR( (! 3) == R.LDA(), ierr );
   std::fill(C.A(), C.A()+9, -4.0);
   std::fill(L.A(), L.A()+9,  2.0);
   std::fill(R.A(), R.A()+9,  2.0);
@@ -2236,7 +2236,7 @@ int checkVbrMatrixOptimizedGraph(Epetra_Comm& comm, bool verbose)
         int bottom = gi + (gj - 1) * Gi;
         int top    = gi + (gj + 1) * Gi;
 
-        EPETRA_TEST_ERR( ! local == matrix->LCID(global), ierr );
+        EPETRA_TEST_ERR( (! local) == matrix->LCID(global), ierr );
 
         indices[ctr++] = local;
         if (gi > first)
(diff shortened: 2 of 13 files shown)


packages/epetra/src/Epetra_Util.h
```diff
@@ -348,7 +348,7 @@ int Epetra_Util_insert_empty_positions(T*& array, int& usedLength,
     return(-1);
   }
 
-  if ((usedLength+numPositions) < allocatedLength) {
+  if (((unsigned int)usedLength+(unsigned int)numPositions) < (unsigned int)allocatedLength) {
```
mhoemmen: I can guess why compilers warned on this, but I'm not a huge fan of the change.  It turns undefined behavior on overflow, into well defined but totally wrong behavior on overflow.  I'm not sure what change would be better, though.

packages/epetra/test/VbrMatrix/cxx_main.cpp
```diff
@@ -1740,7 +1740,7 @@ int checkMergeRedundantEntries(Epetra_Comm& comm, bool verbose)
   int localProc = comm.MyPID();
 
   int myFirstRow = localProc*3;
-  int myLastRow = myFirstRow+2;
+  unsigned int myLastRow = myFirstRow+2;
```
mhoemmen: Wait, woah, why is this `unsigned` here?

packages/epetra/test/VbrMatrix/cxx_main.cpp
```diff
@@ -1781,7 +1781,7 @@ int checkMergeRedundantEntries(Epetra_Comm& comm, bool verbose)
   //FillComplete, which internally calls MergeRedundantEntries, the
   //matrix should contain 1.0 in each entry.
 
-  for(int i=myFirstRow; i<=myLastRow; ++i) {
+  for(unsigned int i=myFirstRow; i<=myLastRow; ++i) {
```
mhoemmen: `unsigned` shouldn't be the right type for a global index in Epetra.

packages/epetra/test/VbrMatrix/cxx_main.cpp
ZUUL42: This took care of the Werror coming from lines 1758 & 1897:
`error: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Werror=strict-overflow]`

packages/epetra/src/Epetra_Util.h
ZUUL42: I'm very open to alternatives.

ZUUL42:
All of the instances of `unsigned` in this PR are for the same error type.
`error: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Werror=strict-overflow]`
It seemed the least intrusive way I found around those.
See [CDash report](https://testing.sandia.gov/cdash/viewBuildError.php?buildid=4976007)
If you know of a better fix, I am all ears.
I was just hoping this would be good or at least good enough for now and can be modified in the future to something more appropriate.

mhoemmen:
The proper type of global indices in Epetra is `int` (or `long long` with Epetra64).  It's a bit risky to mix signed and unsigned types with global indices.  Bugs tend to show up only at large scales.

The "`X + c < X` is always false" warning is interesting -- I see three distinct variables in that expression in the code.  Perhaps the compiler has deduced that two of the variables are the same.

packages/epetra/test/VbrMatrix/cxx_main.cpp
ZUUL42: The `unsigned int i` felt rather wrong, but seemed to do the trick. I suppose I was _hoping_ large scale edge cases were tested for.
I went back and found if I changed `int myLastRow` and `int col` from the `unsigned int` I set to `long long`. Then I could return to `int i`. That seems better.
I'm still looking for an alternative to the `if ((usedLength+numPositions) < allocatedLength)` `strict-overflow` one though.

mhoemmen: > I suppose I was _hoping_ large scale edge cases were tested for.

That's ... optimistic.

mhoemmen: > I went back and found if I changed `int myLastRow` and `int col` from the `unsigned int` I set to `long long`. Then I could return to `int i`. That seems better.

Local row indices have type `int` in Epetra.

mhoemmen: > I'm still looking for an alternative to the `if ((usedLength+numPositions) < allocatedLength)` strict-overflow one though.

That's a tricky one.  I wonder if the compiler has figured out that two of those variables are actually the same thing.

Issue #5510, opened 2019-07-11
Ifpack2: RILUK does not work with Tpetra::StaticProfile graph


tjfulle:
@trilinos/ifpack2 @trilinos/tpetra 

which is a problem, since `DynamicProfile` is deprecated...

I have a fix, just opening this issue for documentation.

PR #5511, opened 2019-07-11
Ifpack2: Heuristic to set upper size of graphs in RILUK.

tjfulle:
Needed to run with `Tpetra::StaticProfile` which will become default
after `Tpetra::DynamicProfile` is removed.

Closes: #5510

@trilinos/ifpack2 @trilinos/tpetra 

The heuristic is probably very bad.  Saad's "Iterative Methods for Sparse Linear Systems" says

> the amount of fill-in and computational work for obtaining the ILU(p) factorization is not predictable for p > 0. 

This statement is in relation to Algorithm 10.5 of the same text.  I'm not sure that is true of this algorithm.  I figured the heuristic would be faster/easier than counting.

All tests now pass

### DIFF
--- a/packages/ifpack2/src/Ifpack2_IlukGraph.hpp
+++ b/packages/ifpack2/src/Ifpack2_IlukGraph.hpp
@@ -260,22 +260,27 @@ void IlukGraph<GraphType>::initialize()
 
   constructOverlapGraph();
 
-  L_Graph_ = rcp (new crs_graph_type (OverlapGraph_->getRowMap (),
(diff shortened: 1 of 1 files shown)


mhoemmen:
Thanks for the quick fix btw :-)

tjfulle:
When I say heuristic, I really mean keep increasing the upper bound till the tests pass ;).  We should be able store all matrix values in crs arrays and defer allocating the actual matrix until after the arrays are full.  Then we can create the matrix with the maps and crs arrays directly.  Right?  I can merge this for now, but it definitely needs more thought.

PR #5516, opened 2019-07-12
Fixed lapack and blas lib locations.

bbraunj:
@trilinos/EM-Plasma

## Description

Changed `Trilinos/cmake/std/atdm/sems-rhel7/environment.sh` lines that include:

```
export ATDM_CONFIG_LAPACK_LIBS="-L${LAPACK_ROOT};-llapack"
export ATDM_CONFIG_BLAS_LIBS="-L${BLAS_ROOT}/lib;-lblas" 
```

to:

```
export ATDM_CONFIG_LAPACK_LIBS="/usr/lib64/liblapack.so.3"
export ATDM_CONFIG_BLAS_LIBS="/usr/lib64/libblas.so.3"
```

## Motivation and Context

Ran into an issue where Trilinos would not build successfully on a RHEL7 machine because it could not find the `lapack` and `blas` libraries.  This points to the right library files.

## How Has This Been Tested?

The configure command that was used:

```
cmake \
  -GNinja \
  -DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake,cmake/std/atdm/apps/empire/EMPIRETrilinosEnables.cmake \
  -DCMAKE_INSTALL_PREFIX=/home/josbrau/Documents/aaa/Trilinos_Install \
  /home/josbrau/Documents/aaa/Trilinos_Src
```

The build command that was used:

```
ninja -j 8
```
 

## Checklist

- [ ] My commit messages mention the appropriate GitHub issue numbers.
- [x] My code follows the code style of the affected package(s).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] No new compiler warnings were introduced.
- [ ] These changes break backwards compatibility.

PR #5517, opened 2019-07-12
Fix inaccurate comment in SEMS RHEL7 env script

mhoemmen:
@trilinos/framework 

## Description

Fix inaccurate comment.  It's really downstream solvers that need `CUDA_LAUNCH_BLOCKING=1`.  Tpetra runs in SPARC just fine without it.


cmake/std/atdm/sems-rhel7/environment.sh
```diff
@@ -190,7 +190,6 @@ elif [[ "$ATDM_CONFIG_COMPILER" == "CUDA-9.2" ]] ; then
   export LAPACK_ROOT=/usr/lib64/atlas
   export ATDM_CONFIG_LAPACK_LIBS="-L${LAPACK_ROOT};-llapack"
   export ATDM_CONFIG_BLAS_LIBS="-L${BLAS_ROOT}/lib;-lblas"
-  # Needed for some Tpetra UVM stuff
```
bartlettroscoe: That was the case the last time I checked.  I had to add this back specifically to make a bunch of Tpetra tests pass.  When did Tpetra remove this requirement?

If it is true that Tpetra does not need this anymore, then we should just remove this and test to make sure.

mhoemmen: @bartlettroscoe Most of _Tpetra_ runs fine without `CUDA_LAUNCH_BLOCKING=1`, but most downstream packages do _not_ run fine.  I merely object to the characterization of Tpetra as the source of all trouble ;-)  @kddevin doesn't need more stress in her life....

bartlettroscoe: Then let's change the comment to something like "some Trilinos tests require this to run correctly"

mhoemmen: Done; thanks!

bartlettroscoe:
@mhoemmen, looks like a conflict is stopping this from merging

prwolfe:
Conflicts with some input from #5516 - looks like a simple fix.

mhoemmen:
Thanks Ross!

PR #5518, opened 2019-07-12
Kokkos FECrsMatrix Assembly Examples

GeoffDanielson:
@trilinos/tpetra 

## Description
Add kokkos-ised example of FECrsMatrix assembly

## Motivation and Context
This is a first step towards a normative example for using kokkos and tpetra together.

## How Has This Been Tested?
This is adding two tests to FECrsMatrix Unit Tests.  These were tested on MPI and Cuda machines.

## Checklist


packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -145,7 +142,7 @@ class GraphPack {
   std::vector<std::vector<GO> > element2node;
 
   // NOTE: This is hardwired for 1D bar elements
-  typedef Kokkos::View<LO*[2], Kokkos::LayoutLeft, typename Node::device_type > k_element2node_type;
+  typedef Kokkos::View<GO*[2], Kokkos::LayoutLeft, typename Node::device_type > k_element2node_type;
```
mhoemmen: Prefer C++11 style `using new_name_type = OldName;`

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -227,6 +224,19 @@ std::vector<std::vector<Scalar> > generate_fem1d_element_values() {
   return mat;
 }
 
+template<class Scalar, class Node>
+Kokkos::View<Scalar*[2], Kokkos::LayoutLeft, typename Node::device_type > generate_fem1d_element_values_kokkos() {
```
mhoemmen: You know that it's a 2x2 matrix, so why not give it the type `Kokkos::View<Scalar[2][2], ...>`?

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -227,6 +224,19 @@ std::vector<std::vector<Scalar> > generate_fem1d_element_values() {
...
+  Kokkos::View<Scalar*[2], Kokkos::LayoutLeft, typename Node::device_type > mat;
+
+  Kokkos::resize(mat, 2);
```
mhoemmen: If you use the type `View<Scalar[2][2], ...>`, then you won't have to resize (in fact, you won't be able to, which is good because it's a 2 x 2 thing).

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -227,6 +224,19 @@ std::vector<std::vector<Scalar> > generate_fem1d_element_values() {
...
+  //plz don't abuse UVM
```
mhoemmen: Well, then don't abuse UVM!  You're welcome to fill on host, but then do this:
```c++
auto mat_h = Kokkos::create_mirror_view (mat);
mat_h(0,0) = 1.0;
// ... etc. ...
Kokkos::deep_copy (mat, mat_h);
return mat;
```

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -292,6 +302,81 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+  FEMAT mat1(graph); // Here we use graph as a FECrsGraph
+  CMAT mat2(graph);  // Here we use graph as a CrsGraph in OWNED mode
+  mat1.beginFill();
+
+  auto k_e2n = pack.k_element2node;
+  auto localMat = mat1.getLocalMatrix();
+  auto localMap = pack.overlapMap->getLocalMap();
+  //get local map too
+
+  Kokkos::parallel_for(Kokkos::RangePolicy<typename Node::execution_space>(0,k_e2n.extent(0)),
```
mhoemmen: 1. Label your kernels, please.  (The first argument should be a string with a human-readable name.)
2. Use typedefs to make this more readable.  For instance:
```c++
using exec_space = typename Node::execution_space;
using range _type= Kokkos::RangePolicy<exec_space, LO>;
Kokkos::parallel_for ("sum into matrix", range_type (0, ...), ...);
```

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
mhoemmen: Also, don't assign integers to Scalar.  That only works if Scalar is `double`, etc.  It breaks if Scalar is `complex`.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
GeoffDanielson: Done.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
GeoffDanielson: IDK why this is even typedef'd, it's not used anywhere else.  Gone.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -349,7 +352,8 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D_Kokkos, LO, GO, Scala
   auto localMap = pack.overlapMap->getLocalMap();
   //get local map too
 
-  Kokkos::parallel_for(Kokkos::RangePolicy<typename Node::execution_space>(0,k_e2n.extent(0)), 
+  Kokkos::parallel_for("assemble_1d",
+		       range_type (0,k_e2n.extent(0)), 
 		       KOKKOS_LAMBDA(const size_t& i) {
```
mhoemmen: Best practice is to pass the loop index (or any type that fits in 64 bits) by value, rather than by reference.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -145,8 +142,7 @@ class GraphPack {
   std::vector<std::vector<GO> > element2node;
 
   // NOTE: This is hardwired for 1D bar elements
-  typedef Kokkos::View<LO*[2], Kokkos::LayoutLeft, typename Node::device_type > k_element2node_type;
-  k_element2node_type k_element2node;
+  Kokkos::View<GO*[2], Kokkos::LayoutLeft, typename Node::device_type > k_element2node;
```
mhoemmen: Do we have a specific reason for using `LayoutLeft` here?  If not, let's drop it, to make the example shorter.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
GeoffDanielson: Yes, that's a good idea.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
GeoffDanielson: Thanks.  The kokkos examples are somewhat varied on that.  I've corrected it.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
GeoffDanielson: OK.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -224,16 +224,16 @@ std::vector<std::vector<Scalar> > generate_fem1d_element_values() {
 }
 
 template<class Scalar, class Node>
-Kokkos::View<Scalar*[2], Kokkos::LayoutLeft, typename Node::device_type > generate_fem1d_element_values_kokkos() {
-  Kokkos::View<Scalar*[2], Kokkos::LayoutLeft, typename Node::device_type > mat;
+Kokkos::View<Scalar[2][2], Kokkos::LayoutLeft, typename Node::device_type > generate_fem1d_element_values_kokkos() {
+  Kokkos::View<Scalar[2][2], Kokkos::LayoutLeft, typename Node::device_type> mat ("fem1d_element_values");
```
mhoemmen: Consider this:
```c++
using Kokkos::view_alloc;
using LL = Kokkos::LayoutLeft;
using DT = typename Node::device_type;

Kokkos::View<Scalar[2][2], LL, DT> mat
  (view_alloc ("fem1d_element_values", WithoutInitializing));
```

mhoemmen: This looks resolved.  Not sure why these little blocks need to be LayoutLeft, but there's nothing incorrect about that.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -292,13 +302,182 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+
+  Kokkos::parallel_for("assemble_1d",
+		       range_type (0,k_e2n.extent(0)), 
+		       KOKKOS_LAMBDA(const size_t i) {
+    size_t extent = k_e2n.extent(1);
+    for(size_t j=0; j < extent; j++) {
+      LO lid_j = localMap.getLocalElement(k_e2n(i, j));
+      for(size_t k=0; k < extent; k++) {
+        LO lid_k = localMap.getLocalElement(k_e2n(i, k));
+	fprintf(stderr, "converting to kokkos complex\n");
```
csiefer2: Can we get rid of these printfs?

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -292,13 +302,182 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+	Kokkos::complex<double> elem(kokkosValues(j, k));
+	fprintf(stderr, "converted to kokkos complex\n");
```
csiefer2: See 366

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -355,14 +534,177 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D_LocalIndex, LO, GO, S
   TPETRA_GLOBAL_SUCCESS_CHECK(out,comm,success)
 }
 
+  /*
```
csiefer2: If we don't need this, nuke it.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -301,15 +302,97 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
   TPETRA_GLOBAL_SUCCESS_CHECK(out,comm,success)
 }
 
-////
-TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D_Kokkos, LO, GO, Scalar, Node )
-{
+//// specializations for kokkos fecrsmatrix assembly
+//std::complex<double>
+template<class LO, class GO, class Scalar, class Node>
```
mhoemmen: 1. This code needs to be protected by the `HAVE_TPETRA_INST_COMPLEX_DOUBLE` macro, since `Scalar=std::complex<double>` can be enabled or disabled at configure time.

2. It's not clear to me why this specialization needs to exist, since there's nothing in it that depends on Scalar being `std::complex<double>`.  (See note below about `Kokkos::complex<double>`.)

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -301,15 +302,97 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+  Kokkos::parallel_for("assemble_1d",
+		       range_type (0,k_e2n.extent(0)), 
+		       KOKKOS_LAMBDA(const size_t i) {
+    size_t extent = k_e2n.extent(1);
+    for(size_t j=0; j < extent; j++) {
+      LO lid_j = localMap.getLocalElement(k_e2n(i, j));
+      for(size_t k=0; k < extent; k++) {
+        LO lid_k = localMap.getLocalElement(k_e2n(i, k));
+	fprintf(stderr, "converting to kokkos complex\n");
+	Kokkos::complex<double> elem(kokkosValues(j, k));
```
mhoemmen: [**[complex.numbers]**](https://eel.is/c++draft/complex.numbers) in C++11 specifies that you may freely `reinterpret_cast` between `std::complex<T>` and `T[2]`.  We use this to justify freely `reinterpret_cast`ing between `Kokkos::complex<T>` (or `impl_scalar_type`) and `std::complex<T>` (or `scalar_type`).  This means that you can write generic code on Scalar, as long as you respect that `impl_scalar_type` and `scalar_type` may not necessarily be the same type.  Thus, you should not need to write a complex-arithmetic specialization of a test, unless the test exercises something specific about complex numbers.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -68,6 +67,7 @@ using Teuchos::NO_TRANS;
 using Teuchos::CONJ_TRANS;
 using std::endl;
 typedef Tpetra::global_size_t GST;
+typedef std::complex<double> scd;
```
mhoemmen: Prefer C++11 `using` syntax over `typedef` for new code.  (This is new code.)

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -145,8 +143,7 @@ class GraphPack {
   std::vector<std::vector<GO> > element2node;
 
   // NOTE: This is hardwired for 1D bar elements
-  typedef Kokkos::View<LO*[2], Kokkos::LayoutLeft, typename Node::device_type > k_element2node_type;
-  k_element2node_type k_element2node;
+  Kokkos::View<GO*[2], typename Node::device_type > k_element2node;
```
mhoemmen: You don't need to leave a space before the closing angle bracket.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
mhoemmen: We lost the `WithoutInitializing` bit -- how come?

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -292,13 +302,92 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+  // get a comm
+  RCP<const Comm<int> > comm = getDefaultComm();
+  
+  // Generate a mesh
+  size_t numLocal = 10;
+  GraphPack<LO,GO,Node> pack;
+  generate_fem1d_graph(numLocal,comm,pack);
+
+  // Make the graph    
+  // FIXME: We should be able to get away with 3 for StaticProfile here, but we need 4 since duplicates are
```
mhoemmen: @tjfulle 's recent work may have fixed this.  It might be worth testing again.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -292,13 +302,92 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+  // not being handled correctly. 
+  RCP<FEG> graph = rcp(new FEG(pack.uniqueMap,pack.overlapMap,4));
+
+  graph->beginFill();
+  for(size_t i=0; i<(size_t)pack.element2node.size(); i++) {
```
mhoemmen: Prefer C++ - style casts.  Also, you use `pack.element2node.size()` and `pack.element2node[i].size()` a lot; consider naming them.
```c++
// use () instead of = to avoid signed/unsigned warnings.
const size_t outer_size (pack.element2node.size());
for (size_t i = 0; i < outer_size; ++i) {
  const size_t inner_size (pack.element2node[i].size());
  for (size_t j = 0; j < inner_size; ++j) {
    // ... etc. ...
  }
}
```

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -292,13 +302,92 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+  size_t numLocal = 10;
+  GraphPack<LO,GO,Node> pack;
+  generate_fem1d_graph(numLocal,comm,pack);
+
+  // Make the graph    
+  // FIXME: We should be able to get away with 3 for StaticProfile here, but we need 4 since duplicates are
+  // not being handled correctly. 
+  RCP<FEG> graph = rcp(new FEG(pack.uniqueMap,pack.overlapMap,4));
+
+  graph->beginFill();
```
mhoemmen: Prefer the nonmember function that Chris wrote:
```c++
beginFill(*graph);
// ... fill the graph ...
endFill(*graph);
```

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -292,13 +302,92 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+  mat1.beginFill();
+
+  auto k_e2n = pack.k_element2node;
+  auto localMat = mat1.getLocalMatrix();
+  auto localMap = pack.overlapMap->getLocalMap();
+  //get local map too
+
+  Kokkos::parallel_for("assemble_1d",
+		       range_type (0,k_e2n.extent(0)), 
+		       KOKKOS_LAMBDA(const size_t i) {
```
mhoemmen: `range_type` uses `LO` (`int`) as the index type, but you're using `size_t` here.  This may cause a signed/unsigned warning.  It's also deceptive for readers, because Kokkos will hand your functor `int`, even though it claims to take `size_t`.

packages/tpetra/core/test/FECrsMatrix/FECrsMatrix_UnitTests.cpp
```diff
@@ -292,13 +302,92 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( FECrsMatrix, Assemble1D, LO, GO, Scalar, Node
...
+    size_t extent = k_e2n.extent(1);
+    for(size_t j=0; j < extent; j++) {
+      LO lid_j = localMap.getLocalElement(k_e2n(i, j));
+      for(size_t k=0; k < extent; k++) {
+        LO lid_k = localMap.getLocalElement(k_e2n(i, k));
+	ImplScalarType tmp = kokkosValues(j, k);
+	localMat.sumIntoValues(lid_j, &lid_k, 1, &tmp, true, true);
```
mhoemmen: Prefer naming those `is_sorted` and `force_atomic` constants.  Also, whether or not you set `force_atomic` should be a function of whether the loop is parallel.
```c++
constexpr bool is_sorted = true;
const bool do_atomic = execution_space().concurrency(); // call this outside parallel region
// ...
localMat.sumIntoValues(lid_j, &lid_k, 1, &tmp, is_sorted, force_atomic);
```

PR #5519, opened 2019-07-12
Tpetra: Fix #4607 (sort rows in explicit transpose by default)

### LINKED ISSUE #4607: Tpetra::RowMatrixTransposer does not sort rows
@trilinos/tpetra @mhoemmen 

The CrsMatrix path of `Tpetra::RowMatrixTransposer::createTransposeLocal` does not sort rows in the `local_matrix_type`, yet it passes the `local_matrix_type` to CrsMatrix's constructor.  The latter assumes that the rows are sorted.  This causes problems downstream.  I have a small reproducer that I need to clean up, but @mhoemmen has a solution in mind.

mhoemmen:
@trilinos/tpetra @trilinos/muelu 

## Description

1. `Tpetra::RowMatrixTransposer::createTranspose*` methods now sort rows (by local column indices) of the result by default.  This fixes #4607.
2. As discussed [here](https://github.com/trilinos/Trilinos/issues/4607#issuecomment-472869688), the sparse matrix-matrix multiply routines do their own sorting.  Thus, we have changed those routines to tell RowMatrixTransposer not to sort rows.
3. Optimized local transpose: Fewer parallel kernel launches; don't initialize allocations unless needed.
4. Removed an unused code path in local transpose, where the input matrix is a RowMatrix but not a CrsMatrix.  In that case, createTransposeLocal now first converts the input to CrsMatrix.

## Related Issues

* Closes #4607

### DIFF
--- a/packages/muelu/src/Utils/MueLu_Utilities_decl.hpp
+++ b/packages/muelu/src/Utils/MueLu_Utilities_decl.hpp
@@ -753,7 +753,17 @@ namespace MueLu {
               // Compute the transpose A of the Tpetra matrix tpetraOp.
               RCP<Tpetra::CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> > A;
               Tpetra::RowMatrixTransposer<Scalar, LocalOrdinal, GlobalOrdinal, Node> transposer(rcpFromRef(tpetraOp),label);
-              A = transposer.createTranspose(params);
+
+              {
+                using Teuchos::ParameterList;
+                using Teuchos::rcp;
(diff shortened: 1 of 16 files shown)


mhoemmen:
@trilinos/tpetra @trilinos/muelu I'd really like a MueLu developer to review this -- I don't think I've changed default behavior for MueLu, but it would be great to know for sure.

mhoemmen:
Thanks @etphipp !  I want to make sure a @trilinos/muelu developer reviews this one, just to make sure.

mhoemmen:
Thanks @etphipp and @csiefer2 ! :-D

PR #5523, opened 2019-07-15
NOX: update inv right scale vector

sconde:
@trilinos/nox 

## Motivation and Context

The `Right Scaling Vector` was normally initialized and never updated. So it makes sense to only compute its reciprocal once and not update it. 

Now that we are updating the scaling vector dynamically in Drekar, the reciprocal must be updated at as well to ensure correctness.

## How Has This Been Tested?

@egphill @rppawlo apart from testing in Drekar, what do you recommend?

## Checklist

### DIFF
--- a/packages/nox/src-thyra/NOX_Thyra_Group.C
+++ b/packages/nox/src-thyra/NOX_Thyra_Group.C
@@ -75,6 +75,7 @@ Group(const NOX::Thyra::Vector& initial_guess,
       const Teuchos::RCP< const ::Thyra::ModelEvaluator<double> >& model,
       const Teuchos::RCP<const ::Thyra::VectorBase<double> >& weight_vector,
       const Teuchos::RCP<const ::Thyra::VectorBase<double> >& right_weight_vector,
+      const Teuchos::RCP<::Thyra::VectorBase<double> >& inv_right_weight_vector,
       const bool rightScalingFirst):
   model_(model),
   rightScalingFirst_(rightScalingFirst),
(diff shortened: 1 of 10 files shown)


rppawlo:
Before pushing, let me look into this. I think the original intent was to not compute the scaling vectors in NOX code. You set them here for nox to use, but the user updates them manually using an observer at the right time. When the update occurs is application dependent -  for example - whether you are doing a steady-state or transient problem.

rppawlo:
@sconde - I would prefer not to do this. This is adding extra computation to an assignment operator. This indicates that the inv_right_weight_vec is broken/out of sync in the source group if I understand things correctly. We should try to find where the right_weight_vec is updated and make sure the source inv_right_weight_vec already has the correct values. Does this make sense?

sconde:
> We should try to find where the right_weight_vec is updated and make sure the source inv_right_weight_vec already has the correct values. Does this make sense?

I think I understand what you mean. The `right_weight_vec` is updated through its RCP that is sent through the NOX sublist. I can update the code to pass in the inverse vector through the parameter list as well if this ok with everyone.

In Drekar, we just grab the RCP for the scaling vector through the solver sublist in the observer. So I can ensure that both the vector and its reciprocal are updated at the same time.

rppawlo:
@sconde - yes- that is exactly it. This PR would be recomputing the inv_right_scaling every time you use the assignment operator for the group, resulting much less efficient code. When the right_sclaing is updated, we should consistently udpate the inv_right_sclaing vec as well. Thanks for taking care of this!

sconde:
@rppawlo @egphill @seamill Is there a reason why the abstract vector is sent in instead of a `Thyra::VectorBase`?
Drekar is casting the `Thyra::VectorBase` to a `NOX::Abstract::Vector` and NOX is recasting that to a `VectorBase`.

`Teuchos::RCP< ::Thyra::VectorBase<double> > right_scaling_vector_;`
```
Teuchos::RCP<NOX::Abstract::Vector> abstract_vec = param_list_->get<RCP<NOX::Abstract::Vector> >("Right Scaling Vector");
right_scaling_vector_ = Teuchos::rcp_dynamic_cast<NOX::Thyra::Vector>(abstract_vec)->getThyraRCPVector();  
```

sconde:
@rppawlo I can use another review

rppawlo:
> @rppawlo @egphill @seamill Is there a reason why the abstract vector is sent in instead of a `Thyra::VectorBase`?

what object/function are you referring to?

sconde:
[Wednesday-2019-07-24-17_26-build_drekar-test.log](https://github.com/trilinos/Trilinos/files/3429113/Wednesday-2019-07-24-17_26-build_drekar-test.log)

sconde:
> looks good! I think you might need to fix some unit tests due to the change in the Group ctor for this to pass the PR tester.

NOX tests are now passing

sconde:
I have Panzer, NOX, and Drekar tests enabled. Should I test this against any other packages?

rppawlo:
@sconde - that looks good shouldn't have to test anything else. PR testing should start shortly.

Issue #4469, opened 2019-02-22
ROL: Remove build warnings


ZUUL42:
@trilinos/rol 

## Expectations

ROL builds should not emit any warnings that will be promoted to errors once Werror is set in the GCC 7.2.0 automated build.

## Current Behavior

Currently ROL has a number of warnings that need to be handled before we can set Werror for all packages.
A recent test build was performed with -Werror set for ROL. The CDash report can be found [here](https://testing.sandia.gov/cdash/viewBuildError.php?buildid=4983662).
UPDATE: [CDash](https://testing.sandia.gov/cdash/viewBuildError.php?buildid=5301681) after most recent ROL sync.

## Motivation, Context and Related Issues

Issue #3178 is working toward turning Warnings as Errors on for all packages.

ZUUL42:
If you'd like the PR to test with Werror set for ROL, pull the commit from branch [Werror_ROL](https://github.com/ZUUL42/Trilinos/tree/Werror_ROL).

ZUUL42:
I've made a [number minor changes](https://github.com/trilinos/Trilinos/compare/master...ZUUL42:Werror_ROL) and have got this issue down to one Werror. However, I'm not quite sure what might be the reasonable/prefered solution for this last one.
@dridzal @trilinos/rol @trilinos/tempus  Would you have any input?

[CDash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&parentid=5321033&filtercount=1&showfilters=1&field1=buildstamp&compare1=61&value1=20190718-1619-Experimental)
`...`
`/scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_gcc_7.2.0/Trilinos/packages/teuchos/core/src/Teuchos_RCPDecl.hpp:990:33: error: deleting object of polymorphic class type ‘ROL::PinTVectorCommunication_Tpetra<double>’ which has non-virtual destructor might cause undefined behavior [-Werror=delete-non-virtual-dtor]`
`   void free( T* ptr ) { if(ptr) delete ptr; }`

dridzal:
@ZUUL42 , my guess is that you would need to add a virtual destructor to the base class `PinTVectorCommunication`

`virtual ~PinTVectorCommunication() {}`

Let me know if it works.

PR #5554, opened 2019-07-19
Fix Warning-Errors and set Werror for ROL in GCC 7.2.0 PR build

ZUUL42:
@trilinos/framework 
@trilinos/rol 

## Description

Fixed warning-errors and set `-Werror` for ROL in the GCC 7.2.0 PR build.
Removed a number of unused variables.
Changed or cast a few variable types from `uint` to `long long` to avoid `Werror=sign-compare`.
Added a virtual dtor for PinTVectorCommunication.
Moved one varable so it would be initialized in the right order to quiet a `Werror=reorder`.

## Motivation and Context

This completes #4469 and is work toward Issue #3178, eventually setting `-Werror` globally.

## How Has This Been Tested?

This was tested [sucessfully running the GCC 7.2.0 PR Jenkins job](https://testing.sandia.gov/cdash/index.php?project=Trilinos&parentid=5322447&filtercount=1&showfilters=1&field1=buildstamp&compare1=61&value1=20190718-2144-Experimental&filtercombine=and).

## Checklist

### DIFF
--- a/packages/rol/adapters/mpi/src/function/ROL_PinTHierarchy.hpp
+++ b/packages/rol/adapters/mpi/src/function/ROL_PinTHierarchy.hpp
@@ -250,7 +250,7 @@ private:
     std::unordered_map<int,DefaultInt> procToIndex;
 
     // loop over all vectors you need to recieve, and recieve them
-    for(int i=0;i<commMap.sendToProc_.size();i++) {
+    for(size_t i=0;i<commMap.sendToProc_.size();i++) {
       int proc = commMap.sendToProc_[i];
  
       if(proc==myRank) continue;
@@ -286,7 +286,7 @@ private:
     std::unordered_map<int,DefaultInt> procToIndex;
 
     // loop over all vectors you need to recieve, and recieve them
-    for(int i=0;i<commMap.recvFromProc_.size();i++) {
+    for(size_t i=0;i<commMap.recvFromProc_.size();i++) {
       int proc = commMap.recvFromProc_[i];
  
       if(proc==myRank) continue;
@@ -536,7 +536,7 @@ public:
        // rebalance the coarse stamps
        ROL::Ptr<std::vector<ROL::TimeStamp<Real>>> coarseStamps 
          = ROL::makePtr<std::vector<ROL::TimeStamp<Real>>>();
-       bool value = ROL::PinT::exportToCoarseDistribution_TimeStamps(*stamps_[level],*coarseStamps,*comm,0);
+       ROL::PinT::exportToCoarseDistribution_TimeStamps(*stamps_[level],*coarseStamps,*comm,0);
 
        stamps_[level] = coarseStamps;
      }
@@ -564,7 +564,7 @@ public:
          MPI_Recv(&sizeA,1,MPI_INT,2*rank  ,level,mpiComm,MPI_STATUS_IGNORE);
          MPI_Recv(&sizeB,1,MPI_INT,2*rank+1,level,mpiComm,MPI_STATUS_IGNORE);
 
-         TEUCHOS_ASSERT(sizeA+sizeB==stamps_[level]->size());
+         TEUCHOS_ASSERT(static_cast<long long>(sizeA) + static_cast<long long>(sizeB) == static_cast<long long>(stamps_[level]->size()));
 
          for(int i=0;i<sizeA;i++) 
            commMap.recvFromProc_[i      ] = 2*rank;
@@ -603,7 +603,7 @@ public:
          MPI_Recv(&sizeA,1,MPI_INT,2*rank  ,level,mpiComm,MPI_STATUS_IGNORE);
          MPI_Recv(&sizeB,1,MPI_INT,2*rank+1,level,mpiComm,MPI_STATUS_IGNORE);
 
-         TEUCHOS_ASSERT(sizeA+sizeB==2*stamps_[level]->size());
+         TEUCHOS_ASSERT(static_cast<long long>(sizeA) + static_cast<long long>(sizeB) == 2 * static_cast<long long>(stamps_[level]->size()));
 
          for(int i=0;i<sizeA;i++) 
            commMap.recvFromProc_[i      ] = 2*rank;
@@ -935,8 +935,6 @@ public:
      if(levelIsActiveOnMyRank(inputLevel)) {
        int timeRank = pint_output->communicators().getTimeRank();
 
-       int fineSteps = pint_output->numOwnedSteps();
-
        // communicate points on the left of this interval
        pint_input->boundaryExchangeLeftToRight();
        auto u_fineStart = pint_input->getRemoteBufferPtr(0)->clone();
@@ -956,7 +954,6 @@ public:
 
        // for(int fne_i=0;fne_i<fineSteps;fne_i++) {
        int fne_i = 0;
-       std::pair<int,int> fneRange = pint_output->ownedStepRange();
        std::pair<int,int> crsRange = pint_input->ownedStepRange();
        for(int crs_i=0;crs_i<pint_input->numOwnedSteps();crs_i++) {
          int crsIndex = crs_i+crsRange.first;
@@ -973,7 +970,6 @@ public:
            int v_crs_index = -1; 
 
            int coarse_si_0 = crs_i-1;     // coarse step index
-           int coarse_si_1 = coarse_si_0+1;
 
            u_crs_index = 2*coarse_si_0;
            v_crs_index = 2*coarse_si_0+1;
@@ -1037,8 +1033,6 @@ public:
 
      // if the fine level is inactive on this proccessor skip the recieve step
      if(levelIsActiveOnMyRank(inputLevel-1)) {
-       int timeRank = pint_output->communicators().getTimeRank();
-
        recvAllFromProcs(*pint_output,
                          pint_output->communicators(),          
                          *pint_output->vectorCommunicationPtr(),
@@ -1098,7 +1092,6 @@ public:
    {
      // nothing to do in this case
      if(levelIsActiveOnMyRank(inputLevel)) {
-       int timeRank = pint_output->communicators().getTimeRank();
        int fineRank = pint_output->communicators().getTimeRank();
 
        // communicate points on the right of this interval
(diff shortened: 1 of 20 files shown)


ZUUL42:
> @ZUUL42 , the type `long long` is not used anywhere in ROL, so I'm not sure that your suggested type changes are correct.

I didn't look to see if it was used elsewhere in ROL. I was only looking at instances where a uint was being compared to an int and throwing a `Werror=sign-compare`. So, I wanted to change or cast the uint into a signed type. Since I'm not familiar with the package or how it's used, I wasn't sure if the value in the uint might overflow when cast/changed to an int, so I chose long long.
I'm open to changing those to something else. Just let me know what would be better.

dridzal:
In general, it is best to figure out what the source type is.  For example, in packages/rol/adapters/mpi/test/function/test_restrict_prolong.cpp, in

`for(long long k=0; k<localNt; ++k ) {`

the type should be `int` because `localNt` is an `int`.  In other cases, the type may not be a standard type.  For example, in packages/rol/example/tempus/example_parabolic_thyravec.hpp, in

for (long long i=0; i<u.subDim(); i++) {

the type is the potentially polymorphic `Teuchos_Ordinal` because the Thyra object `u` returns that as the type.  In other words, some detective work is necessary here.  Overall, fixing these type warnings may not be a trivial exercise.  Generally speaking, hard-typing something to `long long` in Trilinos is probably not ideal, considering that Tpetra and other packages are templated on multiple ordinal parameters.  @trilinos/tpetra @trilinos/thyra

packages/rol/adapters/mpi/src/function/ROL_PinTHierarchy.hpp
```diff
@@ -603,7 +603,7 @@ class PinTHierarchy {
          MPI_Recv(&sizeA,1,MPI_INT,2*rank  ,level,mpiComm,MPI_STATUS_IGNORE);
          MPI_Recv(&sizeB,1,MPI_INT,2*rank+1,level,mpiComm,MPI_STATUS_IGNORE);
 
-         TEUCHOS_ASSERT(sizeA+sizeB==2*stamps_[level]->size());
+         TEUCHOS_ASSERT(sizeA+sizeB==2*(long long)stamps_[level]->size());
```
bartlettroscoe: That is a C-style cast.  Might consider `static_cast<long long>(...)`?

packages/rol/adapters/mpi/test/function/test_hierarchy_coarse_rebalance.cpp
```diff
@@ -251,9 +251,9 @@ void testHiearchyCoarseTimeStamps(MPI_Comm comm, const ROL::Ptr<std::ostream> &
 
     Real tol = 1e-14;
 
-    ROL_TEST_EQUALITY(stamps_0->size(),localNt,false);     
-    ROL_TEST_EQUALITY(stamps_1->size(),localNt,false);     
-    ROL_TEST_EQUALITY(stamps_2->size(),localNt,false);     
+    ROL_TEST_EQUALITY((long long)stamps_0->size(),localNt,false);
```
bartlettroscoe: More C-style casts.

packages/rol/adapters/mpi/test/function/test_timestamp_coarsening.cpp
```diff
@@ -185,7 +185,7 @@ void run_TimeStamp_test(MPI_Comm comm, const ROL::Ptr<std::ostream> & outStream,
       throw std::logic_error(ss.str());
     }
 
-    if(coarseStamps.size()!=local_Nt*2) {
+    if((long long)coarseStamps.size()!=local_Nt*2) {
```
bartlettroscoe: Another C-style cast.

packages/rol/example/tempus/example_parabolic_thyravec.hpp
```diff
@@ -881,8 +881,8 @@ class Objective_ParabolicControl : public ROL::DynamicObjective<Real> {
     const Thyra::ConstDetachedVectorView<Real> x(x_vb);
     const Thyra::ConstDetachedVectorView<Real> y(y_vb);
     Real ip(0);
-    Real c = ((x.subDim()==nx_) ? four : two);
-    for (uint i=0; i<x.subDim(); i++) {
+    Real c = ((x.subDim()==(long long)nx_) ? four : two);
```
bartlettroscoe: C-style cast.

dridzal: @bartlettroscoe , sure, static casts are probably better, however the issue is that we don't have any explicit `long long` types in ROL.  So static casts with the correct types are the best solution.

ZUUL42:
I was thinking of what they were being compared to, but I guess I came to different conclusions. Also, since I don't work on these packages, I wanted to make the least invasive changes possible and wanted the chance of breaking something to be as small as possible. So ...

With that `k` & `localNt` example, I was thinking if `localNt` hits `INT_MAX`, `k` will need to count one past. It could with the former `size_t k`, but `int k` would overflow. Though I'm guessing that is highly unlikely, but that was purely a guess. I wasn't sure why or if it needed to be `size_t` in the first place. So I stayed with the safe choice.

`TEUCHOS_ASSERT(sizeA+sizeB==stamps_[level]->size());` and other `size()` cast to `long long`...
Since `sizeA` & `sizeB` are both `int`, they could potentially add up to higher than `INT_MAX`. So ...
`TEUCHOS_ASSERT(sizeA+sizeB==(long long)stamps_[level]->size());`
Though looking now, I see I could have used a `long int`, well assuming `size()` isn't going beyond `LONG_MAX`. Though if it's hitting `SIZE_MAX` this never would have worked and the whole thing needs to be rethought.

For the few `subDim()` cases, since _it is_ polymorphic I gave up looking for the largest existing case. I was again looking at the original value of what I was changing for a cue.
`i` was a `uint` & `nx_` was a `uint`. So, a `long long` seemed to handle all possible values for those `uint` and if that was good enough before, it's good enough for me too. Though again, that could be a `long int`. Guess I got stuck in `long long` mode with some other packages.

I can change the `localNt` `for` counters to `int`. Though I'm not sure which way you'd like to go on the ordinals and asserts. I was basing my choices on what was already in place.

dridzal:
@ZUUL42 , for the polymorphic types like `Teuchos_Ordinal`, can we just use that type (I mean `Teuchos_Ordinal`, not some standard type that's a "superset")?  I think that's the cleanest solution.  For the asserts, where there is a mismatch let's go with a readily available type, like `int` in the case of `localNt`.

bartlettroscoe:
> For the asserts, where there is a mismatch let's go with a readily available type, like `int` in the case of `localNt`.

That will overflow if the type actually turns out to be `long long`.  If you know the types are signed, then should be safe to up-convert both to `long long` for the comparison.

dridzal:
> That will overflow if the type actually turns out to be long long. If you know the types are signed, then should be safe to up-convert both to long long for the comparison.

@bartlettroscoe , would it be possible to get the index type over which we are iterating and declare the upper bound variable to be of that type?  Why insist on `long long`?

bartlettroscoe:
> would it be possible to get the index type over which we are iterating and declare the upper bound variable to be of that type? Why insist on `long long`?

Actually, this is what a promotion traits class is for.    And look, C++11 have such a thing:

* https://en.cppreference.com/w/cpp/types/common_type

That is what you want to use.

packages/rol/adapters/mpi/src/function/ROL_PinTHierarchy.hpp
mhoemmen: `long long (value)` is also legit.

packages/rol/example/tempus/example_parabolic_thyravec.hpp
mhoemmen: That's not a static cast; it's a C-style cast.  They [mean different things](https://en.cppreference.com/w/cpp/language/explicit_cast).

packages/rol/adapters/mpi/src/function/ROL_PinTHierarchy.hpp
bartlettroscoe: > `long long (value)` is also legit.

Unfortunately, I believe that that is equivalent to `(long long) value` which is a C-style cast.

A correct implicit conversion in C++11 is `long long {value}` I think.  If the conversion from `value` to `long long` is not allowed it will issue an error.  If it is not safe, it will issue a warning.

mhoemmen:
@ZUUL42 wrote:
> Since `sizeA` & `sizeB` are both `int`, they could potentially add up to higher than `INT_MAX`. So ...
> `TEUCHOS_ASSERT(sizeA+sizeB==(long long)stamps_[level]->size());`

That doesn't quite make sense to me.  Isn't the idea that you want to prevent overflow?  Overflowing `int` (and signed built-in integers in general) does not promote to `long long`; it's undefined behavior.  If you're worried about overflowing `int`, you should promote each term in the sum, like this:
```c++
TEUCHOS_ASSERT( static_cast<long long> (sizeA) + static_cast<long long> (sizeB) == static_cast<long long> (stamps_[level]->size ()) );
```

ZUUL42:
@mhoemmen I just put the cast there to take care of the unsigned property of size_t. Then I thought the initial suggestion was pushing toward casting it to an int instead of the larger long long.
But, ya, that would be fully expicit. Though I could have sworn it wouldn't overflow upon adding on one side of a comparitor.
`INT_MAX + INT_MAX == static_cast<long>(INT_MAX * 2)`
That would overflow and not be true?

mhoemmen:
@ZUUL42 The left-hand side definitely overflows.  [Check out this godbolt example](https://godbolt.org/z/k_HR-v):
```c++
#include <climits>
#include <cassert>
int main ()
{
    constexpr auto lhs = INT_MAX + INT_MAX;
    return 0;
}
```
It doesn't even compile, because constant expressions aren't allowed to overflow.

ZUUL42:
Based on all of the comments, I made a few changes.
If this works for everyone, please remove the WIP tag and we'll get this tested & merged.
If not, please let me know what should be changed and I'll get that in.
Thanks.

PR #5570, opened 2019-07-24
Intrepid2: add getFunctionSpace(), etc. to Basis

CamelliaDPG:
@trilinos/intrepid2 

## Description
This PR:
1. adds an `EFunctionSpace` field to Basis, along with a getter, `getFunctionSpace()`.  It also:
2. adds typedefs for the ExecutionSpace and output/point value types,
3. adds typedef aliases that follow C++ camel-case class naming conventions (e.g. the existing `ebasis_view_type` gets an alias `EBasisViewType`),
4. defines a container `fieldOrdinalPolynomialDegree_` in the base class that will be useful for hierarchical bases that will be defined in a future PR, as well as the following related methods:
     - `getFieldOrdinalsForDegree()`
     - `getPolynomialDegreeOfField()`
     - `getPolynomialDegreeOfFieldAsVector()`
     - `getPolynomialDegreeLength()`

## Motivation and Context
1. The `getFunctionSpace()` addition essentially addresses an oversight in the prior implementation.  The new field is initialized in all existing Basis subclasses, and a test for correctness of the value has been added to existing Basis tests.
2. The typedef for ExecutionSpace and output/point value types enables template logic in derived classes without requiring the user to redundantly specify those fields; this will be used in the forthcoming PR with hierarchical basis support.
3. The typedef aliases are mainly a matter of taste and/or style; they allow me to style the code in the forthcoming PR the way that I would like.
4. One use case for hierarchical basis functions involves eliminating higher-order modes in the basis representation (in the context of limiters); another involves using the embedding feature to map from a higher-order basis to a lower-order basis.  For both of these use cases, being able to reason in code about the basis members of a specified degree is crucial.

## Checklist

### DIFF
--- a/packages/intrepid2/perf-test/DynRankView/test_02.hpp
+++ b/packages/intrepid2/perf-test/DynRankView/test_02.hpp
@@ -60,11 +60,11 @@ namespace Intrepid2 {
     namespace Serial {
       
       // compute determinant for rank 2 array
-      template<typename outputViewType,
+      template<typename OutputViewType,
                typename inputViewType>
       KOKKOS_FORCEINLINE_FUNCTION
       void
-      clone( /**/  outputViewType output,
+      clone( /**/  OutputViewType output,
              const inputViewType  input ) {
         const ordinal_type iend = output.extent(0);
         const ordinal_type jend = output.extent(1);
--- a/packages/intrepid2/src/Cell/Intrepid2_CellTools.hpp
+++ b/packages/intrepid2/src/Cell/Intrepid2_CellTools.hpp
@@ -1509,11 +1509,11 @@ namespace Intrepid2 {
       \param  cellTopo          [in]  - cell topology with a reference cell required
   */
   template<typename jacobianViewType,
-           typename pointViewType,
+           typename PointViewType,
            typename worksetCellViewType>
   static void
   CellTools_setJacobianArgs( const jacobianViewType     jacobian,
-                             const pointViewType        points,
+                             const PointViewType        points,
                              const worksetCellViewType  worksetCell,
                              const shards::CellTopology cellTopo );
 
--- a/packages/intrepid2/src/Cell/Intrepid2_CellToolsDefValidateArguments.hpp
+++ b/packages/intrepid2/src/Cell/Intrepid2_CellToolsDefValidateArguments.hpp
@@ -63,11 +63,11 @@ namespace Intrepid2 {
   //============================================================================================//
 
   template<typename jacobianViewType, 
-           typename pointViewType,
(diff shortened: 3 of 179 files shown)


CamelliaDPG:
@mperego, thanks for the review!

> 1. The documentation of the added functions in Intrepid2_Basis.hpp is missing input and output parameters (see other functions in the same file).

Thanks for catching that.  I'll fix.

> 2. I'm not a fan of typedef aliases. If we agree to use your convention (which is fine with me) we should probably rename the types and then create deprecated aliases for the old name, which we'll get rid of eventually.

I'm happy to do that.  Would you like me to include the type renaming (throughout Intrepid2) with this PR, or is it OK to do that in a separate PR?

> 3. Do we really need `std::vector` overloads of functions like `getFieldOrdinalsForDegree`? If all you need are simple std::vector functionalities like push_back, you can probably use Kokkos::vector and then pass around the vector's view.

I hadn't been aware of [Kokkos::vector](https://github.com/kokkos/kokkos/wiki/Interoperability#1233-stdvector).  I just wanted to present a flexible API that allows things like calling with an argument of numbers inside braces (e.g., `getFieldOrdinalsForDegree({1,2,2})`), which Kokkos::View arguments do not support.  It seems likely that Kokkos::vector will support that, though; I'm happy to switch to using that.

> 4. There are modifications in MueLu files. Not sure how to fix it.. maybe rebase your branch against the develop branch
Yes, I'm not sure how those came in, and not yet sure how to fix, either.  I just did a fresh fetch into my fork's develop branch, and then a squash atop that in the PR branch in my fork, but it doesn't look like that's resolved it.  I'll work on it.

mperego:
Thanks @CamelliaDPG !
As for renaming the types, If you are willing to, I would do that in this PR. We should also fix them in the rest of Trilinos, to avoid warnings of deprecated code. I think we are about to have a minor release in Trilinos with tons of deprecated code (all the Tpetra changes), and it would be good if we can have this merged before the release.

@kyungjoo-kim, are you OK with the the naming convention for types that @CamelliaDPG wants to use? https://google.github.io/styleguide/cppguide.html#Type_Names

CamelliaDPG:
@mperego, assuming that this round of testing succeeds, it's ready for another look.  I've added deprecation warnings for the old typedefs, and resolved all warnings generated by that addition.  It looks like, within Trilinos, only Intrepid2 used the deprecated typedefs (before the fixes, I was getting warnings during Panzer's build, but the actual use was within Intrepid2).  I've also added some guards to throw a clean exception if methods intended for hierarchical bases are invoked on non-hierarchical bases.

Issue #2518, opened 2018-04-09
Question: Tpetra: Appropriate data structure for block data


ctanis:
I am getting terrible performance with Experimental::BlockCrsMatrix  and MultiVec (BlockCrsVector doesn't compile).   What is currently the best way to deal with multiple degrees of freedom per mesh point?

@trilinos/tpetra

## Expectations

Experimental::BlockCrsMatrix.apply should be faster in parallel, or I should be doing this in a different way.

## Current Behavior

Using MPI (np = 2), BlockCrsMatrix.apply is taking approximately 100 times longer than in serial (still getting the correct answer, I think).

This is applying a block matrix with 1 degree of freedom to a MultiVector (not block) with 1 vector.

Attempting to compile a simple example with BlockVector does not compile. 
[compilation_error.txt](https://github.com/trilinos/Trilinos/files/1891118/compilation_error.txt)

## Motivation and Context

I understand these are "experimental" data structures, so I would love to know the most appropriate way to go forward.

Also note:  I was using 1 degree of freedom per mesh point to make sure I was solving the problem correctly.  My imminent need is to have 4 or more per mesh point.  Potentially many more!

I'm going to attempt to add a 'question' label after submitting this.  My github interface is somehow missing the label and assignee dropdowns.

mhoemmen:
@ctanis What version of Trilinos are you using?  Also, are you using OpenMP?

mhoemmen:
@trilinos/tpetra

mhoemmen:
> My github interface is somehow missing the label and assignee dropdowns.

The Trilinos repository restricts assignment of labels and assignees to members.

kyungjoo-kim:
@ctanis We have an example code to measure performance of block crs matrix vector multiplication. 
After you compile Trilinos, you can find out an executable ``packages/tpetra/core/example/BlockCrs/TpetraCore_BlockCrsPerfTest.exe``. 
Testing the code with 
- 7 point 3D stencil on a 32x32x32 block mesh 
- my workstation xeon 2x8 intel sandy bridge
- OMP_NUM_THREADS=1 (mpi only).
I have the following result. Could you verify that you also get the same performance as I have seen below ? The code compares the block crs apply method against the point crs apply method. 

```
[...]

PR #5574, opened 2019-07-25
Small Improvement on BlockCrs 

kyungjoo-kim:
@trilinos/tpetra 

## Description

Fix for block crs multivector comm layer that perform deep copy for each block. This also include a bit of improvement for the block crs apply method.

## Motivation and Context

#2518  @ctanis This PR includes the performance benchmark of block crs apply method in tpetra. We also include a markdown summary of the performance you can reproduce.


packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
```diff
@@ -124,537 +124,541 @@ int main (int argc, char *argv[])
...
+
+        // Tpetra BlockCrsMatrix only has high level access functions
+        // To fill this on device, we need an access to the meta data of blocks
+        const auto rowptr_host = Kokkos::create_mirror_view(rowptr);
+        const auto colidx_host = Kokkos::create_mirror_view(colidx);
+
+        Kokkos::deep_copy(rowptr_host, rowptr);
+        Kokkos::deep_copy(colidx_host, colidx);
+
+        blocks = Kokkos::View<block_type*,exec_space>("blocks", rowptr_host(num_owned_elements));
```
mhoemmen: I guess this is OK, because `block_type` is explicitly `MemoryUnmanaged`.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
```diff
@@ -124,537 +124,541 @@ int main (int argc, char *argv[])
...
+        // getLocalBlock.  This means we must NOT use KOKKOS_LAMBDA,
+        // since that would build the lambda for both host AND device.
+        Kokkos::parallel_for
+          (Kokkos::RangePolicy<host_space, LO> (0, num_owned_elements),
+           [=] (const LO row) {
+            const auto beg = rowptr_host(row);
+            const auto end = rowptr_host(row+1);
+            typedef typename std::remove_const<decltype (beg) >::type offset_type;
+            for (offset_type loc = beg; loc < end; ++loc) {
+              blocks_host(loc) = A_bcrs->getLocalBlock(row, colidx(loc));
```
mhoemmen: I'm not sure if it's OK to capture RCP by value in a thread-parallel loop, even though it's on host.  I would recommend capturing `const BlockCrsMatrix&` instead.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
```diff
@@ -124,537 +124,541 @@ int main (int argc, char *argv[])
...
+        typename tpetra_crs_matrix_type::local_matrix_type
+          local_matrix("local_crs_matrix",
+                       num_owned_and_remote_elements*blocksize,
+                       crs_values,
+                       local_graph_type(crs_colidx, crs_rowptr));
 
-    if (verbose) {
-      A_crs->describe(*out, Teuchos::VERB_EXTREME);
-    }
+        A_crs = rcp(new tpetra_crs_matrix_type(row_crs_map,
```
mhoemmen: You could specify the domain and range Maps explicitly here.

packages/tpetra/core/src/Tpetra_BlockMultiVector_decl.hpp
```diff
@@ -138,6 +138,11 @@ namespace Tpetra {
 /// It doesn't make sense for BlockMultiVector to implement
 /// MultiVector, because the desired fill interfaces of the two
 /// classes are different.
+///
+/// BlockMultiVector does not have comm layer e.g., packAndPrepare,
```
mhoemmen: Hm, this wasn't what I meant.  What I was thinking is that we could implement BlockMultiVector's DistObject methods by dispatching to MultiVector methods on the underlying MultiVector.  That would also simplify BlockCrsMatrix -- BlockCrsMatrix could just trust BlockMultiVector to do the right thing, and BlockMultiVector would call MultiVector's methods.

kyungjoo-kim: I am not sure if that approach is feasible. I think that what you want, for example ``copyAndPermute``, is 
```
BlockMultiVector(src, numSameIDs, permuteToLIDs, permuteFromLIDS) {
  this->getMultiVectorView().copyAndPermute
    (src.getMultiVectorView(),
     numSamdIDs*blockSize_,
     // the following is problematic
     permuteToLIDs.withMagicToPopulateIndicies(),
     permuteFromLIDS.withMagicToPopulateIndicies());
 }
```
If this is what you intended, it is a bit problematic as it requires a new view object to populate flattened indicies. I can write the routine with some cost and we can recommend not to use block multivector importer. However, I do not think that this is good approach. That is why I completely remove the comm layer. Instead of deleting the methods, I can put the less-performant version with a warning message to make the block multivector comm operational.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
kyungjoo-kim: umm... you are right. it should be captured by reference.

packages/tpetra/core/src/Tpetra_BlockMultiVector_decl.hpp
mhoemmen: Ah, good point, I forgot about the indexing thing.  I'm OK with this.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
mhoemmen: @kyungjoo-kim This is the only part I'm really worried about, and it's only a big deal if the host execution space is OpenMP.  It's unwise to take RCP inside a parallel loop.

packages/tpetra/core/src/Tpetra_BlockMultiVector_decl.hpp
kyungjoo-kim: I would like to do so but I see some missing symbol errors in linking. I may need to provide some fake imple with throwing an error if it is used. I will also modify the test to use the point multi vector dist object interface.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
kyungjoo-kim: Now I have some curiociy. If using rcp is not a good idea in a parallel loop, why is kokkos view reference counting okay ? For the same reason, block_host will increase the view reference count multiple times. Does it use atomic ? Does this mean that using unmanaged views inside of a parallel loop is always better than using using managed views ? KOKKOS_LAMBDA basically capture by value. On GPUs, does this cause the kokkos reference count to explode ? Just curiocity.

mhoemmen: > If using rcp is not a good idea in a parallel loop, why is kokkos view reference counting okay ?

1. Kokkos switches off reference counting for Views inside parallel regions.
2. RCP's reference count is not thread safe by default.  (There's a nondefault build option to enable this, but it does have a run-time cost.)  `std::shared_ptr`'s reference count _is_ thread safe with respect to `std::thread`.

mhoemmen: > On GPUs, does this cause the kokkos reference count to explode ?

RCP's methods don't have device function markings, so they should not even compile in a device kernel.  Even if they did have device function markings, RCP's reference count lives in a pointer to host memory (a struct that has the reference count in it), so it's not accessible on device if you created the RCP on host.

kyungjoo-kim:
Ummm.... one PR test reports warning as an error with [-Werror=overloaded-virtual], which means that some function signature is different from the base class (DistObject). I don't know why the particular compiler complains. A fix is added and will be retested.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
mhoemmen: This should actually be fine -- @kyungjoo-kim captures the RCP by reference.  Better practice would be to capture a reference to the object, but this is not incorrect.

kyungjoo-kim:
@mhoemmen Thanks. Would you mind if I ask you to check if SPARC uses mesh-map based importer instead of a point importer ? I am wondering if SPARC integration test fails due to some interface changes made in this PR.

mhoemmen:
@kyungjoo-kim SPARC configures Trilinos with `Tpetra_BCRS_Point_Import=ON` only for GPU builds, not for CPU builds.

PR #5579, opened 2019-07-26
Xpetra: ETI for BlockedVector, BlockedMultiVector, MapExtractor, MultiVectorFactory

william76:
@trilinos/xpetra 
@trilinos/muelu 

Partially completes #4986 

The initial plan was to ETI just BlockedVector but that cascaded into more that needed ETI work.  In this PR I've got 
- [x] `Xpetra::BlockedVector`
- [x] `Xpetra::BlockedMultiVector`
- [x] `Xpetra::MapExtractor`
- [x] `Xpetra::MultiVectorFactory`

There are a couple dangling things on this PR that I am not sure about:

In BlockedMultiVector and in MapExtractor there are a couple calls that were being made in the pre-ETI code that looked something like `full->setMultiVector(...);` where `full` is a `MultiVector` and `MultiVector::setMultiVector()` doesn't exist.  There is a `bfull` var in those functions that is a `BlockedMultiVector` which does have that method.  I wasn't sure if this is just a typo and the intent is to use bfull or not so I just a `#if 0` guard around it and throw an exception.  The exception didn't get thrown in my testing so maybe these aren't ever actually getting called (and maybe never were hence why the typo has persisted in develop so long?).  These functions are here:

https://github.com/william76/Trilinos/blob/1868de2f95eade8016fd6c835c09bcc64090de36/packages/xpetra/src/BlockedMultiVector/Xpetra_BlockedMultiVector_def.hpp#L1217-L1223

https://github.com/william76/Trilinos/blob/1868de2f95eade8016fd6c835c09bcc64090de36/packages/xpetra/src/BlockedCrsMatrix/Xpetra_MapExtractor_def.hpp#L527-L533


mayrmt:
@william76 I don't have practical experience with these parts of the code, but here's my view of thinks: I suspect, that this is not a type, but rather a safety mechanism to deal with the case if the `Teuchos::rcp_dynamic_cast<Xpetra::BlockedMultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>>(full)` fails, i.e. if `full` is not a `BlockedMultiVector, but just a regular `MultiVector`. In that case, one does not have to insert into a specific block of the `BlockedMultiVector`, but can rather insert into the entire `MultiVector`. Assuming my explanation is true, then I still don't know, why there is the `block` argument in `full->setMultiVector(block, partial, bThyraMode);`.

william76:
> @william76 I don't have practical experience with these parts of the code, but here's my view of thinks: I suspect, that this is not a type, but rather a safety mechanism to deal with the case if the `Teuchos::rcp_dynamic_cast<Xpetra::BlockedMultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>>(full)` fails, i.e. if `full` is not a `BlockedMultiVector`, but just a regular `MultiVector`. In that case, one does not have to insert into a specific block of the `BlockedMultiVector`, but can rather insert into the entire `MultiVector`. Assuming my explanation is true, then I still don't know, why there is the `block` argument in `full->setMultiVector(block, partial, bThyraMode);`.

I'm suspicious that those functions haven't ever really been compiled since `MultiVector->setMultiVector()` doesn't exist but that call was in the original code prior to my ETI work... if a compiler had ever tried to compile down into that function it'd have returned an error.  That I put a throw there and none of my builds have triggered it seems to corroborate my thinking on this one.  

@csiefer2 @jhux2 any thoughts?

mayrmt:
I see. Well, then maybe leaving it as it is right now seems to be a viable option. If somebody comes across this error in the future, then there is at least a hint, what might have to be implemented.

PR #5413, opened 2019-06-19
WIP: do not commit. Automatic snapshot commit from seacas at d7789af

gsjaardema:
Origin repo remote tracking branch: 'origin/master'
Origin repo remote repo URL: 'origin = https://github.com/gsjaardema/seacas'

At commit:

commit d7789af0cdfa0555fa245f9930f4723dc90e368b
Author:  Greg Sjaardema <gdsjaar@sandia.gov>
Date:    Wed Jun 19 07:57:42 2019 -0600
Summary: EPU: Enable smart assert for release builds also

* Some clang-format changes
* Run epu with SMART_ASSERT enabled even in release
* Fix some uses/conflicts of C-library functions (abs, fabs, mergesort)
* Eliminate some unused variable warnings
* Some exodus warnings were using stderr directly instead of ex_err.
* Exodus summary output uses dynamic width now
* io_shell has `add_processor_id_field` option to make easier to visualize a
  decomposition (structured and unstructured)
* refactor 'create_path' to make clearer
* update to Catch-2.9.1
* CGNS: Remove the single-file read in parallel since new CGNS-develop
  library has a faster solution (compact storage).
* CGNS: detect if not using patched CGNS library
* Fix heartbeat option to make easier to override default options.
* Refine cgns_decomp output and options.
* cleanup some old copyright info
* fix some fortran warnings

NOTE: This commit will not build until Trilinos is pointing to a patched version of the
      CGNS library.  The patch is in the SEACAS github repository TPL/CGNS/CGNS-sandia.patch
      and is applied to the current CGNS/develop version

@trilinos/seacas 

## Checklist

- [x] My commit messages mention the appropriate GitHub issue numbers.
- [x] My code follows the code style of the affected package(s).
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] No new compiler warnings were introduced.
- [x] These changes break backwards compatibility.
         (Will not build with an unpatched CGNS library)

PR #5597, opened 2019-07-30
CGNS: Move CGNS patch detection compile-time error to runtime warning

gsjaardema:
The CGNS library is addressing parallel scalability issues and has started using
a compact storage method available in HDF5. This can speed up parallel execution
by around two orders of magnitude.  

Prior to this fix, the SEACAS/IOSS/CGNS implementation had its own workaround which
sped up parallel execution by two orders of magnitude at low to medium processor counts
and an order of magnitude at "larger" processor counts.  This workaround has been removed since the CGNS compact storage usage is a much better and maintainable solution.  This was comitted to CGNS in https://github.com/CGNS/CGNS/commit/6199af200f0692d8da8ab5f7f0825d4100d54cee

There is also another issue in that CGNS had mistakenly applied another optimization which provided a little more speedup, but made the resulting files incompatible with downstream clients using an older version of the HDF5 library.  By default, the CGNS library creates files which are readable by HDF5-1.8.X clients even though the CGNS-writing application may be using HDF5-1.10.X libraries.  This issue was fixed in https://github.com/CGNS/CGNS/commit/dfdd91da824798332b539cf74e513a70aa6ed76b

But, even with both of these fixes applied to the current CGNS/develop, it is still required that clients that use the IOSS library to read/write CGNS files be patched using the "CGNS-sandia.patch"  file at https://github.com/gsjaardema/seacas/blob/master/TPL/cgns/CGNS-sandia.patch.  This patch makes it possible to do both parallel and serial reading and writing of CGNS files during the same application execution.  This can be the reading of a file-per-processor combined with the writing of an N->1.  The changes provided in the patch are part of a PR for CGNS (https://github.com/CGNS/CGNS/pull/61), but have not yet been accepted and applied.

This patch was recently changed to add the definition of the symbol `CGNS_SANDIA_PARALLEL_MODS` in order to make it easier to determine whether the application is using a version of the library with the patch applied or not.  Note that if the patch is not applied, then the application may experience hangs. 

The issue we are potentially seeing in Trilinos and SPARC is that an older version of the patch
did not define the symbol, so it was not possible to determine whether the patch had been
applied. It is possible that the CGNS library being used has been patched so there won't be
hangs (but probably still slower execution at scale) but we cannot detect it.  This commit removes the compile-time error check for the symbol and replaces it with a run-time warning.

This is related to https://sems-atlassian-srn.sandia.gov/browse/SPAR-589, #5413 

@trilinos/seacas

## Checklist

### DIFF
--- a/packages/seacas/libraries/ioss/src/cgns/Iocgns_ParallelDatabaseIO.C
+++ b/packages/seacas/libraries/ioss/src/cgns/Iocgns_ParallelDatabaseIO.C
@@ -184,11 +184,13 @@ namespace Iocgns {
 #if !defined(CGNS_SANDIA_PARALLEL_MODS)
     fmt::print("WARNING: The CGNS library being used in this executable does *NOT* have the "
                "CGNS-sandia.patch applied.\n"
-               "         This can result in very slow execution at large processor counts and can "
-               "possibly create\n"
+               "         This can result in hangs in parallel exeuctions or, with older versions,\n"
+	       "         very slow execution at large processor counts and can possibly create\n"
                "         CGNS files which are not readable by applications linked with older HDF5 "
                "libraries.\n"
-               "         contact gdsjaar@sandia.gov for additional info.\n");
+	       "         Note that in some cases, an older version of the patch has been applied\n"
+	       "         which does not define the symbol currently being used to detect the patch.\n"
+               "         Contact gdsjaar@sandia.gov for additional info.\n");
 #endif
 
 #if IOSS_DEBUG_OUTPUT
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
CC: @fryeguy52, @mhoemmen 

@gsjaardema, just to be clear, this should fix the Trilinos build errors reported in [SPAR-589](https://sems-atlassian-srn.sandia.gov/browse/SPAR-589)?

gsjaardema:
@bartlettroscoe Yes this should fix https://sems-atlassian-srn.sandia.gov/browse/SPAR-589

mhoemmen:
Ross beat me to it :-D

Issue #5595, opened 2019-07-30
MueLu: tests  fail to build if Galeri is disabled


tjfulle:
## Bug Report
@trilinos/muelu 

### Description
Many MueLu tests fail to build if `Trilinos_ENABLE_Galeri:BOOL=OFF` is defined.

Related to #5590

jhux2:
@tjfulle I will take this.

jhux2:
@tjfulle This was a simple fix.  PR is in.

jhux2:
This was already fixed in head:
 https://github.com/trilinos/Trilinos/blob/0193ca84a98452528343e362b77a0f61ee757ac3/packages/muelu/example/basic/CMakeLists.txt#L18

I will leave this open until you verify it works for you.

tjfulle:
Thanks @jhux2!  I think it is working.  I'll leave it open a while longer until the other MueLu work related #5602 is finished.

cgcgcg:
Since we have made Galeri a required dependency for MueLu's tests, this is fixed.
https://github.com/trilinos/Trilinos/blob/6e5c6c9586ccd0b63855c5fd9789a386d85a6a9c/packages/muelu/cmake/Dependencies.cmake#L6

PR #5601, opened 2019-07-30
MueLu: add cmake guard if Galeri disabled

jhux2:
Fixes issue #5595

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/5601)


packages/muelu/example/basic/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ INCLUDE_DIRECTORIES(${CMAKE_CURRENT_SOURCE_DIR})
 #
 
 IF(HAVE_MUELU_INST_DOUBLE_INT_INT)
```
mhoemmen: Y'all realize that this test will never get enabled in the future, right?

lucbv: @mhoemmen we won't allow `int_int` only `long_int_int` and `long_long_int_int` ?

mhoemmen: That's correct, but this test only gets enabled for `int_int` right?

tjfulle:
@jhu, autotester reports conflicts in your changes.  Did someone else modify this file too?

jhux2:
This appears to be fixed in develop already.  I'm going to delete this PR.

Issue #5594, opened 2019-07-30
FEI: build fails with Epetra disabled


tjfulle:
@trilinos/fei 

 With `Trilinos_ENABLE_Epetra:BOOL=OFF` defined, I get the following build error:

```
packages/fei/support-Trilinos/fei_Include_Trilinos.hpp:84:10: fatal error: BelosEpetraAdapter.hpp: No such file or directory
 #include <BelosEpetraAdapter.hpp>
```

Full configure:

```
/projects/sems/install/rhel6-x86_64/sems/utility/cmake/3.12.2/bin/cmake \
  -G "Ninja" \
  -D Trilinos_CONFIGURE_OPTIONS_FILE:FILEPATH=$TRILINOS_DIR/cmake/std/PullRequestLinuxCommonTestingSettings.cmake \
  -D Trilinos_ENABLE_Tpetra:BOOL=ON \
  -D Trilinos_ENABLE_Epetra:BOOL=OFF \
  -D Trilinos_ENABLE_Moertel:BOOL=OFF \
  -D Trilinos_ENABLE_ShyLU_DDFROSch:BOOL=OFF \
  -D Trilinos_ENABLE_Stokhos:BOOL=OFF \
  -D ROL_ENABLE_EXAMPLES:BOOL=OFF \
  -D Tpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
  -D Xpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
  -D Trilinos_ENABLE_Galeri:BOOL=OFF \
  -D Rythmos_ENABLE_TESTS:BOOL=OFF \
  $TRILINOS_DIR
```

alanw0:
Hmm, this is a tough one. FEI isn't going to live into the Tpetra era, it really needs Epetra. Perhaps we need to add some cmake logic such that epetra is a mandatory dependency and FEI is disabled if Epetra is not available.

tjfulle:
@alanw0 that would be fine, and I'm happy to add such logic to CMake.  Does FEI have any Tpetra code?  If so, I'll have to be careful about how Tpetra is instantiated in the event that Tpetra+Epetra are enabled.  If not, I can just put a hard stop in CMake if Epetra is not enabled.

rppawlo:
Would this be the time to consider removing FEI from trilinos?

alanw0:
[...]

Issue #5602, opened 2019-07-30
Tpetra: Removal of deprecated code meta issue


tjfulle:
## Bug Report
@trilinos/tpetra 

### Description
Meta issue to track work to remove deprecated code from @trilinos/tpetra

The work (so far) concentrates on two configurations, each with deprecated code **disabled**:

1. Tpetra's global ordinal = `int` (`TPETRA_INST_INT_INT:BOOL=ON`)
2. Tpetra's global ordinal = the default value (`long long int`)

Each configuration has build and/or test problems outlined in the following comments.

*Related PRs*
- #5590 
- #5573 

*Related Issues*
- #5537 
- #5538 
- #5540 
- #5541 
- #5592 
- #5593 
- #5594 
- #5595 

*Supersedes*
- #5536

tjfulle:
# Status for `GlobalOrdinal=int`, `LocalOrdinal=int`

## Configuration and Build

With deprecated code disabled and using the default Tpetra `GlobalOrdinal`, the
build fails due to inconsistencies between Tpetra's global ordinal and Epetra.
For instance, Xpetra fails to build because it requires that Tpetra's global
ordinal be `int` when Epetra is enabled.  The solution is to disable Epetra
support in this and other packages, or set the global ordinal to `int`, as in
the configuration below.

Note, the configuration options file below requires that
`${TRILINOS_DIR}/cmake/std/sems/PullRequestGCC7.3.0TestingEnv.sh` be sourced.

```
#!/bin/sh
rm -rf CMake*
  cmake
  -G "Ninja" \
  -D Trilinos_ENABLE_Tpetra:BOOL=ON \
  -D Tpetra_INST_INT_INT:BOOL=ON \
  -D Trilinos_CONFIGURE_OPTIONS_FILE:FILEPATH=${TRILINOS_DIR}/cmake/std/PullRequestLinuxCommonTestingSettings.cmake \
  -DTpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
  -DXpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
  -D Trilinos_ENABLE_STK:BOOL=OFF \
  -D ROL_ENABLE_EXAMPLES:BOOL=OFF \
  ${TRILINOS_DIR}
```

## Failing tests

- [x] Amesos2_GappedMtxGIDs-1proc_MPI_1
- [ ] Amesos2_KLU2_UnitTests_MPI_2
[...]

PR #5603, opened 2019-07-30
FEI: make Epetra a required dependency

tjfulle:
Closes: #5594
Related: #5602


tjfulle:
@alanw0, not the most elegant solution, but I’m not inclined to spend any more time on a package scheduled for deletion.

PR #5606, opened 2019-07-31
Panzer delayed dump

bathmatt:
This takes advantage of the stk changes which allow delayed adding vars to dump.


bathmatt:
I've checked that this worked in EMPIRE for delayed initialization.  There will be more changes coming but I wanna push this through

bathmatt:
There is going to be a change on this, figured out a better way without the performance hit if you don't need it..

rppawlo:
ok. This was already merged into develop so I guess you will need a new PR.

PR #5612, opened 2019-07-31
Panzer delayed dump2

bathmatt:
I made it so you only turn on fields after finalize if you need to.  Should not change any exisiting performance.


rppawlo:
@bathmatt - did you mean to close this?

PR #5613, opened 2019-07-31
Turned off delated fields unless needed

bathmatt:
Made it so you only turn on delayed fields if needed


bathmatt:
@rppawlo this is a minor performance change for other panzer codes if you don't finalize the mesh up front.

rppawlo:
thanks @bathmatt !

PR #5633, opened 2019-08-02
Phalanx: remove clang warning

rppawlo:


rppawlo:
yes. gcc 7.2 somehow doesn't pick it up.

Issue #5617, opened 2019-07-31
ShyLU: Tests failures with Tpetra's deprecated code is disabled


tjfulle:
## Bug Report
@trilinos/shylu 

Part of: #5602 

### Description

With Tpetra's deprecated code disabled, 2 ShyLU tests fail for `GlobalOrdinal=int`:

- ShyLU_DDFROSch_test_thyra_xpetra_laplace_TLP_IPOUHarmonic_DIM2_DPN2_ORD0_TPETRA_MPI_4
- ShyLU_DDFROSch_test_thyra_xpetra_laplace_TLP_IPOUHarmonic_DIM2_DPN2_ORD1_TPETRA_MPI_4

See the full configuration for `GlobalOrdinal=int` at https://github.com/trilinos/Trilinos/issues/5602#issuecomment-516593697

tjfulle:
@kddevin can you assign the Tpetra team member that was going to look at this?

brian-kelley:
@kddevin @tjfulle @trilinos/xpetra I have a fix for these, it was just about changing an Xpetra::MatrixFactory::Build() call with StaticProfile and the exact # entries per row. However, I noticed a couple of things that should probably be changed:

-Xpetra::CrsMatrix does not support getNumEntriesInGlobalRow(), which would be really useful for changes like this one. I had to do getGlobalRowView again for every row to read off the row sizes. It's easy to implement getNumEntriesInGlobalRow() by calling it on the underlying matrix, but I'm wondering if there's a good reason this hasn't already been done.

-Xpetra::MatrixFactory::Build still uses Xpetra::DynamicProfile as the default "profile" parameter value, even if Tpetra the underlying library.

brian-kelley:
See #5637. I ended up adding "getNumEntriesInGlobalRow()" to Xpetra.

PR #5637, opened 2019-08-02
ShyLU: fixed test failures with deprecated off (#5617)

brian-kelley:
Fixed test failures:
...IPOUHarmonic_DIM2_DPN2_ORD0_TPETRA_MPI_4 and
...IPOUHarmonic_DIM2_DPN2_ORD1_TPETRA_MPI_4 in ShyLu_DDFROSch
by changing Xpetra matrix construction to use StaticProfile.

@trilinos/shylu 
@trilinos/xpetra 

## Description

At FROSch_CoarseOperator_def.hpp:235, changed the creation of Xpetra::Matrix to use StaticProfile and exact entry counts. This is effectively a deep copy from tmpCoarseMatrix to CoarseMatrix_, except with 1.0 inserted on the diagonal in rows which are empty in tmpCoarseMatrix.

As part of this, implemented
Xpetra::Matrix::getNumEntriesInGlobalRow(GO) because without it,
counting entries in each row of tmpCoarseMatrix requires an additional call to
getGlobalRowView() for every row. Like Xpetra::Matrix::getNumEntriesInLocalRow(LO),
this function is a trivial one-line wrapper for the same function on
the underlying matrix.
## Motivation and Context

Fixes the test failures: ShyLU_DDFROSch_test_thyra_xpetra_laplace_TLP_IPOUHarmonic_DIM2_DPN2_ORD0_TPETRA_MPI_4 and ShyLU_DDFROSch_test_thyra_xpetra_laplace_TLP_IPOUHarmonic_DIM2_DPN2_ORD1_TPETRA_MPI_4 when Tpetra_ENABLE_DEPRECATED_CODE=OFF.

## Checklist

- [x] My commit messages mention the appropriate GitHub issue numbers.
- [x] My code follows the code style of the affected package(s).
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] No new compiler warnings were introduced.
- [ ] These changes break backwards compatibility.

PR #5650, opened 2019-08-02
ShyLU: Fixed test link errors #5618

### LINKED ISSUE #5618: ShyLU: build failures with Tpetra's deprecated code disabled
## Bug Report
@trilinos/shylu 

Part of: #5602 

### Description

With Tpetra's deprecated code disabled, ShyLU fails to build with the configuration at `GlobalOrdinal=long long int` at: https://github.com/trilinos/Trilinos/issues/5602#issuecomment-516594843 (removing the lines disabling ShyLU, of course)

The build failure could be due to `GlobalOrdinal=long long int`, or Epetra being disabled?

brian-kelley:
Fixed link errors in several ShyLU_DDFROSch tests, where the global ordinal was
unknowingly hardcoded as int. Made Xpetra pick a reasonable default GlobalOrdinal.

@trilinos/xpetra 
@trilinos/shylu 

## Description

Building with Tpetra GO=long long (but not GO=int) caused the failure, because Xpetra structrues were not instantiated with *_int_int_node. The reason why GO was hardcoded to int in ShyLU is because ShyLU used the Xpetra "default types" to define Scalar, LO, GO and NO. In Xpetra_DefaultTypes.hpp, GO was hardcoded to int, regardless of what was enabled in CMake. So this commit adds logic in Xpetra to choose a reasonable default GlobalOrdinal based on what is enabled: if Tpetra is enabled, the default GO of Tpetra is chosen. Then if Epetra 64-bit GO is enabled, long long is used. Otherwise, int is used.

## Motivation and Context

This fixes a ShyLU build failure as part of the Tpetra deprecated code removal (#5602).

## How Has This Been Tested?

This was built with Tpetra GO=long long enabled, GO=int disabled, Epetra disabled, and deprecated code disabled. Most tests in FROSch failed to build before, but now build and run successfully (except for the two in #5617, which were fixed separately).

## Checklist

I didn't say that all existing tests passed, because some other fixes (see #5602) for tests in this configuration have not made it to develop branch yet. The tests affected by these changes do pass.

### DIFF
--- a/packages/xpetra/src/Headers/Xpetra_UseDefaultTypes.hpp
+++ b/packages/xpetra/src/Headers/Xpetra_UseDefaultTypes.hpp
@@ -54,7 +54,19 @@
 // Define default data types
 typedef double Scalar;
 typedef int LocalOrdinal;
+
+// Choose global ordinal based on enabled underlying libraries
+#ifdef HAVE_XPETRA_TPETRA
+typedef typename Tpetra::Map<>::global_ordinal_type GlobalOrdinal;
+//Have Epetra only; use Epetra's default GlobalOrdinal
+#else
+#ifdef EPETRA_NO_32BIT_GLOBAL_INDICES
+typedef long long GlobalOrdinal;
+#else
 typedef int GlobalOrdinal;
+#endif
+#endif
+
 typedef KokkosClassic::DefaultNode::DefaultNodeType Node;
 
 #endif // DOXYGEN_SHOULD_SKIP_THIS
--- a/packages/xpetra/src/Map/Xpetra_Map.hpp
+++ b/packages/xpetra/src/Map/Xpetra_Map.hpp
@@ -75,6 +75,21 @@ namespace Xpetra {
 # endif
 #endif
 
+  namespace Details
+  {
+    //Check for the default Tpetra GlobalOrdinal (long long)
+    #ifdef HAVE_XPETRA_TPETRA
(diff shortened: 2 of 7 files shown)


packages/xpetra/src/Headers/Xpetra_UseDefaultTypes.hpp
```diff
@@ -54,7 +54,19 @@
 // Define default data types
 typedef double Scalar;
 typedef int LocalOrdinal;
+
+// Choose global ordinal based on enabled underlying libraries
+#ifdef HAVE_XPETRA_TPETRA
+typedef typename Tpetra::Map<>::global_ordinal_type GlobalOrdinal;
```
mhoemmen: Brian and I talked about this in person today.  I suggested adding `#error` if Epetra and Tpetra are enabled but Tpetra's default ordinal type is not `int`.  That's already an error in Xpetra and it might already catch that case at configure time, but just in case, it would be good to add a test here.

packages/xpetra/src/Map/Xpetra_Map.hpp
```diff
@@ -75,14 +75,29 @@ namespace Xpetra {
 # endif
 #endif
 
+  namespace Details
+  {
+    //Check for the default Tpetra GlobalOrdinal (long long)
+    #ifdef HAVE_XPETRA_TPETRA
```
mhoemmen: See above.  Also, perhaps we should consolidate these two "test the macros" things in a single place.

searhein:
@brian-kelley Thanks for fixing this and #5637.

PR #5651, opened 2019-08-03
teko:  fixes to support staticProfile matrix construction in Tpetra #5602

kddevin:
#5602 
@trilinos/teko 

## Description
Changes to enable only static profile construction of Tpetra CrsMatrices.
Enables builds with Tpetra_ENABLE_DEPRECATED_CODE=OFF to succeed.

Changed one test to provide a maximum number of entries per row.  (Note that this test is rather odd in that it inserts into one matrix row twice, while not inserting at all into another matrix row.  But I digress.)

Changed the subblock constructor to count the number of nonzeros per row, then create the matrix, and then insert the nonzeros.

## Motivation and Context
DynamicProfile construction of Tpetra CrsGraph and CrsMAtrix is deprecated and will disappear soon.

Tested on linux with
Tpetra_ENABLE_DEPRECATED_CODE=OFF
Xpetra_ENABLE_DEPRECATED_CODE=OFF
Tpetra_INST_INT_INT=ON

### DIFF
--- a/packages/teko/src/Tpetra/Teko_BlockingTpetra.cpp
+++ b/packages/teko/src/Tpetra/Teko_BlockingTpetra.cpp
@@ -236,7 +236,6 @@ RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > buildSubBlock(int i,int j,const RCP<const T
    const RCP<Tpetra::Vector<GO,LO,GO,NT> > plocal2ContigGIDs = getSubBlockColumnGIDs(*A,subMaps[j]);
    Tpetra::Vector<GO,LO,GO,NT> & local2ContigGIDs = *plocal2ContigGIDs;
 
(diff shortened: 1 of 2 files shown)


packages/teko/src/Tpetra/Teko_BlockingTpetra.cpp
```diff
@@ -250,6 +249,34 @@ RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > buildSubBlock(int i,int j,const RCP<const T
...
+   // for counting
+   std::vector<size_t> nEntriesPerRow(numMyRows);
+
+   // insert each row into subblock
+   // Count the number of entries per row in the new matrix
+   for(LO localRow=0;localRow<numMyRows;localRow++) {
+      size_t numEntries = -1;
+      GO globalRow = gRowMap->getGlobalElement(localRow);
+      LO lid = A->getRowMap()->getLocalElement(globalRow);
+      TEUCHOS_ASSERT(lid>-1);
```
mhoemmen: In the future, it's nicer to use the flag value `Teuchos::OrdinalTraits<LO>::invalid()`.

Issue #5672, opened 2019-08-07
Xpetra: Compile error in BlockedMap


william76:
@trilinos/xpetra 

## Description
Kevin Cox sent us a report of a compilation he was getting that looks like it might be something from the Xpetra ETI work.  I put together a configuration that is as close as I can get to the one he reported but it does not replicate his error.  The biggest difference might be that he's on Geminga but I don't have an account there.  He's also loading AmgX and I don't have that in my environment.

### Error Message
Here's the error message he's seeing:
```
/home/kcox/Trilinos/packages/xpetra/src/BlockedMap/Xpetra_BlockedMap.hpp(537): error: incomplete type is not allowed
          detected during:
            instantiation of "void Xpetra::BlockedMap<LocalOrdinal, GlobalOrdinal, Node>::assign(const Xpetra::BlockedMap<LocalOrdinal, GlobalOrdinal, Node> &) [with LocalOrdinal=int, GlobalOrdinal=int, Node=Kokkos_Compat_KokkosCudaWrapperNode]"
/home/kcox/Trilinos/packages/teuchos/core/src/Teuchos_dyn_cast.hpp(175): here
            instantiation of "T_To &Teuchos::dyn_cast<T_To,T_From>(T_From &) [with T_To=const Xpetra::BlockedMap<int, int, Kokkos_Compat_KokkosCudaWrapperNode>, T_From=const Xpetra::Map<int, Xpetra::Details::DefaultGlobalOrdinal, KokkosClassic::DefaultNode::DefaultNodeType>]"
/home/kcox/Trilinos/packages/teuchos/core/src/Teuchos_RCP.hpp(768): here
            instantiation of "Teuchos::RCP<T2> Teuchos::rcp_dynamic_cast<T2,T1>(const Teuchos::RCP<T1> &, bool) [with T2=const Xpetra::BlockedMap<int, int, Kokkos_Compat_KokkosCudaWrapperNode>, T1=const Xpetra::Map<int, Xpetra::Details::DefaultGlobalOrdinal, KokkosClassic::DefaultNode::DefaultNodeType>]"
/home/kcox/Trilinos/packages/xpetra/src/MultiVector/Xpetra_MultiVectorFactory_def.hpp(68): here
[...]

PR #5673, opened 2019-08-07
Xpetra: Fix for #5672

william76:
@trilinos/xpetra 
This branch / PR is set up to work on the bug reported in #5672 

@csiefer2 per our conversation... I put this branch up and have this PR as a WIP.  I couldn't replicate the error Kevin reported on CEE but I made a couple small tweaks here that are my best guess.  I moved the include of `Xpetra_BlockedMap.hpp` from `Xpetra_MultiVectorFactory_decl.hpp` into `Xpetra_MultiVectorFactory_def.hpp` which will force the decl file to have been processed prior to linking in BlockedMap.  Since BlockedMap isn't actually referenced in the decl file this can't hurt anything. 

If that doesn't fix, we might also try including `Xpetra_MultiVectorFactory_decl.hpp` in `Xpetra_BlockedMap.hpp`... but knowing that will require replicating the error cited in the issue.


packages/xpetra/src/MultiVector/Xpetra_MultiVectorFactory_decl.hpp
```diff
@@ -58,7 +58,7 @@
 #include "Xpetra_EpetraIntMultiVector.hpp"
 #endif
 
-#include "Xpetra_BlockedMap.hpp"
+// #include "Xpetra_BlockedMap.hpp"
```
mhoemmen: Here are the rules for whether or not to include this header:

1. If you use the class by pointer, you only need a forward declaration.
2. If you use a typedef in the class, you need an actual class declaration.
3. If you call a method in the class, you need the class' definition.

It looks like `Xpetra::BlockedMap` does not show up in this file at all, so commenting it out is right.

packages/xpetra/src/MultiVector/Xpetra_MultiVectorFactory_def.hpp
```diff
@@ -64,8 +65,8 @@ Build(const Teuchos::RCP<const Xpetra::Map<LocalOrdinal, GlobalOrdinal, Node>>&
 {
     XPETRA_MONITOR("MultiVectorFactory::Build");
 
-    RCP<const BlockedMap<LocalOrdinal, GlobalOrdinal, Node>> bmap =
-        Teuchos::rcp_dynamic_cast<const BlockedMap<LocalOrdinal, GlobalOrdinal, Node>>(map);
+    RCP<const Xpetra::BlockedMap<LocalOrdinal, GlobalOrdinal, Node>> bmap =
```
mhoemmen: This looks like the only place that actually needs the definition of `Xpetra::BlockedMap` (since the target type [must be a complete type](https://en.cppreference.com/w/cpp/language/dynamic_cast)).

mhoemmen: Do people see the trouble that all these template parameters are causing us?  Some piece of code that should be trivial is costing hours and hours of multiple developers' time.  None of these three template parameters have ever added value for downstream solvers or applications.

@trilinos/framework @prwolfe @jwillenbring @bartlettroscoe @micahahoward @rrdrake I want people to see this.

mhoemmen: @trilinos/tpetra @kddevin

csiefer2:
Yeah, this doesn't fix the problem.  Will look at it in more detail.

william76:
sonofa... I hate it when github grabs other junk to "add" to a PR.  killing this PR

PR #5676, opened 2019-08-07
SEACAS: ioss - Fix bad previous change in SerializeIO

gsjaardema:
@trilinos/seacas 

## Description

A previous snapshot had changed the Ioss_SerializeIO class by removing a member variable which was private and unused.  This removal resulted in an empty if block which was removed; however, it was part of an if/ifelse/else block and the removal of the empty if block was incorrect since entering that block would result in *not* entering the other blocks.  

Added back the if block which fixes the behavior of Nalu which is hanging due to this change.

PR also includes some clang-format changes.

## Checklist

### DIFF
--- a/packages/seacas/libraries/ioss/src/Ioss_SerializeIO.C
+++ b/packages/seacas/libraries/ioss/src/Ioss_SerializeIO.C
@@ -76,10 +76,12 @@ namespace Ioss {
       }
     }
 
-    if (s_groupFactor > 0) {
+    if (m_activeFallThru) {
+    }
+    else if (s_groupFactor > 0) {
 #ifdef SEACAS_HAVE_MPI
       do {
-	MPI_Barrier(util.communicator());
+        MPI_Barrier(util.communicator());
       } while (++s_owner != s_groupRank);
 #endif
       m_databaseIO->openDatabase();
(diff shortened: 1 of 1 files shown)


packages/seacas/libraries/ioss/src/Ioss_SerializeIO.C
```diff
@@ -76,10 +76,12 @@ namespace Ioss {
...
     }
 
-    if (s_groupFactor > 0) {
+    if (m_activeFallThru) {
+    }
+    else if (s_groupFactor > 0) {
 #ifdef SEACAS_HAVE_MPI
       do {
-	MPI_Barrier(util.communicator());
+        MPI_Barrier(util.communicator());
```
mhoemmen: This code makes me quite nervous (how does it not deadlock?) but I don't know the context.

bartlettroscoe: I think I remember one of the Epetra print functions had like 4 MPI_Barrier() calls :-)

gsjaardema: The code is somewhat confusing and is best understood by running it in the debugger.  

The deadlock is avoided by:
 * there also being a MPI_Barrier call in the Ioss::SerializeIO destructor, and
 * all processors must call the Ioss::SerializeIO constructor synchronously.

The flow is that the ranks are split into groups of the specified size.  Assume 3 ranks of group size 1.  
 * First time through, rank 0 falls through and ranks 1, 2 sit at the barrier
 * rank 0 hits the destructor and then all 3 ranks are in the barrier, so they all go to next step
   - (rank 1,2 in constructor, rank 0 in destructor)
 * s_owner is now equal to m_groupRank on rank 1, so it falls out of the do while; 
    - rank 2 still in the constructor do while Barrier
    - rank 0 in the destructor do while Barrier
    - rank 1 does its work and calls destructor;
    - all ranks in Barrier, so they go to next step.
 * s_owner now equal to m_groupRank on rank 2, so if falls out of the do while;
    - ranks 0,1 in destructor do while at the Barrier
    - rank 2 does it work and calls destructor
    - all ranks are now in the destructor Barrier, so they go to next step
    - all ranks clear the Destructor and go to next step.

There are 'm_groupSize' calls of `MPI_Barrier()`.  Overall this works nicely if you have a file system that gets overwhelmed by all ranks calling the IO functions simultaneously.  You do have to ensure that you don't stack calls to SerializeIO or miss calling it on one or more ranks.

gsjaardema: @bartlettroscoe I now have a test of this in the SEACAS github repository which will be synched over in the next snapshot.

PR #5694, opened 2019-08-12
Adding option to tempus DIRK stepper for resetting initial guess.

crockatt:
@trilinos/tempus
Tagging drekar team for visibility: @sconde @johnshadid @rppawlo 

## Motivation and Context

The changes in this pull request are motivated by solver difficulties observed in Drekar. In certain situations, the nonlinear solve for a stage may become severely divergent. The resulting diverged solution is nevertheless returned and stored in the internal stage vector of the StepperDIRK class, and becomes the initial guess for the next nonlinear solve. When the timestep is cut and the step is attempted again, this diverged solution persists. Because this diverged solution is a very poor initial guess, subsequent solves will fail regardless of how far the timestep is reduced. The purpose of this pull request is to implement a mechanism to ensure that a reasonable initial guess is used for nonlinear solves after a timestep has failed.

## Description

Implemented option in StepperDIRK to reset the initial stage vector at the beginning of each timestep. If this option is enabled, the stage vector is set to the current solution (the initial condition for the timestep). If this option is set to false, the current behavior of the stepper class is recovered.

## How Has This Been Tested?

A test has been implemented in Drekar to verify that this change has the desired effect on solver robustness.

## Checklist

- [X] My code follows the code style of the affected package(s).
- [X] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [X] All new and existing tests passed.
- [X] No new compiler warnings were introduced.

### DIFF
--- a/packages/tempus/src/Tempus_StepperDIRK_decl.hpp
+++ b/packages/tempus/src/Tempus_StepperDIRK_decl.hpp
@@ -132,6 +132,12 @@ public:
     virtual void setInitialConditions (
       const Teuchos::RCP<SolutionHistory<Scalar> >& solutionHistory);
 
+    /// Set parameter so that the initial guess is reset at the beginning of each timestep.
+    virtual void setResetInitialGuess(bool reset_guess)
+      { this->stepperPL_->template set<bool>("Reset Initial Guess", reset_guess); }
+    virtual bool getResetInitialGuess() const
+      { return this->stepperPL_->template get<bool>("Reset Initial Guess", true); }
+
     /// Take the specified timestep, dt, and return true if successful.
     virtual void takeStep(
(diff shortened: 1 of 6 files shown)


crockatt:
Pull request has been updated following discussion with @ccober6. Resetting the initial guess at the beginning of each timestep is now the default behavior for DIRK steppers.

sconde:
@ccober6 what is your suggestion for testing?

@crockatt has a test for this in Drekar already.

ccober6:
@crockatt  and I talked about this, and tested an idea for a quick and easy Tempus test, but it did not pan out.  I think we may need a test similar to what is in Drekar for this, but not sure if we can make one of our ModelEvaluators go unstabel and recover very easily.  Do you have any thoughts @sconde?

PR #5701, opened 2019-08-13
Adding Gauss-Seidel to region MG

rstumin:
Auto-PR for SHA 1d4fc18

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
@@ -187,6 +187,101 @@ void jacobiIterate(RCP<Teuchos::ParameterList> smootherParams,
   return;
 } // jacobiIterate
 
+/*! \brief Do Gauss-Seidel smoothing
+ *
+ *  Perform Gauss-Seidel smoothing in the region layout using the true diagonal value
+ *  recovered from the splitted matrix.
+ */
+template<class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
+void GSIterate(RCP<Teuchos::ParameterList> smootherParams,
+                   std::vector<RCP<Xpetra::Vector<Scalar, LocalOrdinal, GlobalOrdinal, Node> > >& regX, // left-hand side (or solution)
+                   const std::vector<RCP<Xpetra::Vector<Scalar, LocalOrdinal, GlobalOrdinal, Node> > > regB, // right-hand side (or residual)
+                   const std::vector<RCP<Xpetra::Matrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> > > regionGrpMats, // matrices in true region layout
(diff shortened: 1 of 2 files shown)


packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
```diff
@@ -187,6 +177,91 @@ void jacobiIterate(RCP<Teuchos::ParameterList> smootherParams,
...
+  const Scalar SC_ZERO = Teuchos::ScalarTraits<Scalar>::zero();
+  const Scalar SC_ONE = Teuchos::ScalarTraits<Scalar>::one();
+  const int maxIter    = smootherParams->get<int>   ("smoother: sweeps");
+  const double damping = smootherParams->get<double>("smoother: damping");
+
+
+  std::vector<RCP<Vector> > regRes(maxRegPerProc);
+  createRegionalVector(regRes, maxRegPerProc, revisedRowMapPerGrp);
+
+  // extract diagonal from region matrices, recover true diagonal values, invert diagonal
```
lucbv: I think the extraction of the diagonal should go in a setup phase and simply be reused here instead of recomputing it on the fly.
That being said I don't mind if that change happens in another PR as it is not essential for a first cut.

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
```diff
@@ -208,7 +283,7 @@ void smootherSetup(RCP<Teuchos::ParameterList> params,
     jacobiSetup(params, maxRegPerProc, revisedRowMapPerGrp, regionGrpMats, regionInterfaceScaling);
     break;
   case 2:
-    std::cout << "Gauss-Seidel smoother not implemented yet no smoother is applied" << std::endl;
+    jacobiSetup(params, maxRegPerProc, revisedRowMapPerGrp, regionGrpMats, regionInterfaceScaling);
```
lucbv: Do we need the data computed for the Jacobi preconditioner to be computed in the Gauss-Seidel case?

mayrmt: @lucbv @rstumin Right now, it looks like the inverse of the true diagonal needed in Gauss-Seidel is computed during the apply-phase. If we keep it that way, then we don't have to call `jacobiSetup()` here. If we want to pre-compute the inverse of the true diagonal in the setup-phase, then we should 
- rename `jacobiSetup()` to `relaxationSmootherSetup()`
- call `relaxationSmootherSetup()` for both cases, Jacobi and Gauss-Seidel
- use the pre-comuted inverse of the true diagonal during `GSIterate()`

> This would also address @lucbv's [other comment to this PR](https://github.com/trilinos/Trilinos/pull/5701/files#r313500551).

I can have a look at this next week, maybe on Monday. Please let me know what you think.

PR #5707, opened 2019-08-14
Seacas pamgen fix

gsjaardema:
@trilinos/seacas 

## Motivation and Context

Fixes the pamgen DatabaseIO class to correctly handle the `compute_block_membership` function.  Also fixes a few other issues discovered while testing this issue:
* output_summary would deadlock in threadsafe code
* pamgen class was not being tested in Trilinos/SEACAS 

Added test which tests both Pamgen Database object and the io_info summary output.

This was reported by Alan Williams due to an issue reported by Roger Pawloski.

## Checklist


packages/seacas/libraries/ioss/src/pamgen/Iopg_DatabaseIO.h
```diff
@@ -101,8 +101,8 @@ namespace Iopg {
     std::string title() const { return databaseTitle; }
     int         maximum_symbol_length() const override { return 32; }
 
-    void compute_block_membership(Ioss::SideBlock *         efblock,
-                                  std::vector<std::string> &block_membership) const;
+    void compute_block_membership__(Ioss::SideBlock *         efblock,
```
mhoemmen: If `block_membership` is empty on entrance and you mean to create it, it's perfectly efficient just to return the `std::vector` by value.  Return value optimization will avoid reallocation and copying.

gsjaardema: It is not necessarily empty on entry.  Same vector will be passed in for each block in the model and additional names pushed back to end.  With new move semantics, it is probably ok to return by value now, but originally was written pre C++11.  Will look into modernizing...

mhoemmen: Thanks!

PR #5712, opened 2019-08-14
Fix #5711

### LINKED ISSUE #5711: CUDA and CUSPARSE TPLs are ST; make them PT
## Bug Report
@trilinos/framework 

### Description

`Trilinos/TPLsList.cmake` says that the CUDA and CUSPARSE (sic) TPLs (third-party libraries) are ST (Secondary Tested).  They should be PT (Primary Tested).  Reasons:

1. We have a CUDA PR test build, so CUDA should be considered just as important as MPI, the BLAS, and LAPACK (which are all PT).
2. CUSPARSE comes with CUDA automatically, but since it's ST, we're almost certainly not enabling it by default in our CUDA PR test build.  This is bad because NVIDIA has done a lot of work to implement kernels that we're likely not testing.

mhoemmen:
@trilinos/framework 

## Description

Make CUDA and CUSPARSE (sic) PT instead of ST.

## Motivation and Context

1. We have a CUDA PR test build, so CUDA should be considered just as important as MPI, the BLAS, and LAPACK (which are all PT).
2. CUSPARSE comes with CUDA automatically, but since it's ST, we're almost certainly not enabling it by default in our CUDA PR test build. This is bad because NVIDIA has done a lot of work to implement kernels that we're likely not testing.

## Related Issues

* Closes #5711

### DIFF
--- a/TPLsList.cmake
+++ b/TPLsList.cmake
@@ -57,8 +57,8 @@ TRIBITS_REPOSITORY_DEFINE_TPLS(
   MKL             "cmake/TPLs/"    EX
   yaml-cpp        "cmake/TPLs/"    EX
   Peano           "cmake/TPLs/"    EX
-  CUDA            "${${PROJECT_NAME}_TRIBITS_DIR}/core/std_tpls/"    ST
-  CUSPARSE        "cmake/TPLs/"    ST
+  CUDA            "${${PROJECT_NAME}_TRIBITS_DIR}/core/std_tpls/"    PT
+  CUSPARSE        "cmake/TPLs/"    PT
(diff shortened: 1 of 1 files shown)


mhoemmen:
Thanks Ross! :-D

@bartlettroscoe wrote:
> But making this change seems like the right thing to do to communicate the prominence of these TPLs to Trilinos testing.

Yes, that was the intent.  Mainly I wanted CUSPARSE to be enabled by default if CUDA is enabled.  There's no reason for us not to use CUSPARSE if it's available, and it's always available if CUDA is ON.

mhoemmen:
Thanks Ross! :-D

PR #5714, opened 2019-08-14
Amesos2: Initial setup for UVM off via Tacho

MicheldeMessieres:
These are changes to Amesos2 to use Kokkos::Views and remove UVM dependencies. Currently using the Amesos2 Tacho solver test as the development case.

@trilinos/amesos2

## Description
Switched Tacho solver from  DefaultHostExecutionSpace to DefaultExecutionSpace and converted Teuchos::Arrays to be Kokkos::Views. Started implementing changes so this will run on GPU without dependencies on UVM. API calls are being rewritten to take Kokkos::Views. Type conversions are now handled in Kokkos::parallel_for loops on the appropriate execution space.

## Motivation and Context

Change Amesos2 handling of ptrs to use Kokkos::Views and consider device/host. Remove UVM dependence.

## How Has This Been Tested?

Currently tested on Mac serial and parallel and Cuda serial with UVM on due to Tpetra requirements. In progress.

## Checklist


packages/amesos2/src/Amesos2_TpetraMultiVecAdapter_def.hpp
```diff
@@ -219,6 +219,133 @@ namespace Amesos2 {
...
+        if ( redist_mv.isConstantStride() ) {
+          // MDM-TODO - Determine if this is the best way
+          // WARNING - This is not tested currently by Tacho Amesos2_Solver_Test so not tested anywhere yet!
+          auto contig_local_view_2d = redist_mv.template getLocalView<typename KV::execution_space>();
+          Kokkos::parallel_for(
+            Kokkos::RangePolicy<typename KV::execution_space, size_t> (0, num_vecs * lda),
+            KOKKOS_LAMBDA (const size_t & n) {
+            auto i = n % num_vecs;
+            auto j = n / num_vecs;
+            kokkos_view(i,j) = contig_local_view_2d(i,j);
```
mhoemmen: You don't need to write a kernel here; just use `Kokkos::deep_copy`.

srajama1:
I would let @kyungjoo-kim and @ndellingwood comment on this.

packages/amesos2/src/Amesos2_TpetraMultiVecAdapter_def.hpp
```diff
@@ -559,11 +553,7 @@ namespace Amesos2 {
       // MDM-TODO Is this the right way to update MV: copying one element at a time?
       // WARNING - This is not tested currently by Tacho Amesos2_Solver_Test - tested manually by setting numVecs 1 in Solver_Test.cpp. Add param?
       auto mv_view_to_modify_2d = mv_->template getLocalView<typename KV::execution_space>();
```
mhoemmen: I don't see any `sync*` or `modify*` calls before you modify (on what could be host or device).  I would strongly recommend using `withLocalAccess` if you don't understand how that stuff works.  Thanks!

MicheldeMessieres:
@kyungjoo-kim @ndellingwood @srajama1 

These are notes to describe the current changes for setting up Amesos2 GPU solvers to handle UVM off in a stream-lined way.

Now if we run tests we will have the following output on a GPU build for Amesos2 Tacho Solver:

```
      Start 17: Amesos2_Tacho_Solver_Test_Default_To_Host_MPI_1
17/27 Test #17: Amesos2_Tacho_Solver_Test_Default_To_Host_MPI_1 ............................   Passed    1.73 sec
      Start 18: Amesos2_Tacho_Solver_Test_Host_To_Host_MPI_1
18/27 Test #18: Amesos2_Tacho_Solver_Test_Host_To_Host_MPI_1 ...............................   Passed    1.84 sec
      Start 19: Amesos2_Tacho_Solver_Test_CudaUVM_To_Host_MPI_1
19/27 Test #19: Amesos2_Tacho_Solver_Test_CudaUVM_To_Host_MPI_1 ............................   Passed    1.77 sec
      Start 20: Amesos2_Tacho_Solver_Test_Cuda_To_Host_MPI_1
20/27 Test #20: Amesos2_Tacho_Solver_Test_Cuda_To_Host_MPI_1 ...............................   Passed    1.78 sec
      Start 21: Amesos2_Tacho_Solver_Test_Default_To_Cuda_MPI_1
21/27 Test #21: Amesos2_Tacho_Solver_Test_Default_To_Cuda_MPI_1 ............................   Passed    1.78 sec
      Start 22: Amesos2_Tacho_Solver_Test_Host_To_Cuda_MPI_1
22/27 Test #22: Amesos2_Tacho_Solver_Test_Host_To_Cuda_MPI_1 ...............................   Passed    1.79 sec
      Start 23: Amesos2_Tacho_Solver_Test_CudaUVM_To_Cuda_MPI_1
23/27 Test #23: Amesos2_Tacho_Solver_Test_CudaUVM_To_Cuda_MPI_1 ............................   Passed    1.83 sec
      Start 24: Amesos2_Tacho_Solver_Test_Cuda_To_Cuda_MPI_1
24/27 Test #24: Amesos2_Tacho_Solver_Test_Cuda_To_Cuda_MPI_1 ...............................   Passed    1.78 sec
```

A copy manager handles data transfer between the memory spaces. If the memory spaces match, the view is just assigned, otherwise a deep_copy is performed. The copy manager will allocate the dst view size only if needed, so we no longer allocate at initialization to avoid an unnecessary allocation.

In the above tests, Host_To_Host and Cuda_To_Cuda will just assign while everything else uses deep_copy. I’ve left logs in place just to verify we see it doing the right thing.

Default means don’t do testing mode and use whatever comes ’naturally’, which is currently CudaUVMSpace since Tpetra requires it. If the source memory space is requested for a different testing mode, I copy the Tpetra data into that memory space, then send it to Amesos2.

Having the data from Tpetra intercepted and copied into the test memory space is a temporary measure so I could set up the testing framework. I would like to make a new adapter which is based on Kokkos CrsMatrix and can take any memory space. That would remove the need for the intermediate copy in the testing mode.

Half of the tests are copying to Host and the 2nd half copy to Cuda. This is because I’ve modified the Amesos2 TachoSolver so it builds both a Device and Host version for GPU builds, This was done using a macro define and perhaps better handled by templating, but I don’t think I can do it in a better way without manipulating SolverCore template assumptions, and I didn’t want to mess with that before discussion.

To get around the Tpetra UVM on constraint and get a testing framework, I implemented a new execution space called CudaUVMOff. This is a copy of the Cuda execution space except it always has memory space CudaSpace, not CudaUVMSpace. So Tacho and Amesos2 are being built using the CudaUVMOff (UVM off) execution space while Tpetra stays with Cuda execution space (UVM on).

To avoid the need for CudaUVMOff we would need to change how Tacho and TaskScheduler are templated. Perhaps they can be templated on a device which takes both an execution space and a memory space. Then I could just specify UVM off.

So CudaUVMOff execution space is not necessarily expected to be a permanent fixture. If we did want to keep it, we might do it more cleanly by having a CudaBase execution space and then derive Cuda and CudaUVMOff from that. The derived classes would set the typedefs including the memory space. However because of the way they self reference the execution space, I’m not sure what is involved to make that all work.

These tests would pass on non-GPU with more than 1 MPI rank, but fail for GPU. Right now I’ve left them all at MPI 1. It will crash in Tacho_Chol_External.hpp and I think it could be Tacho related - or perhaps MPI 1 is all we care about presently for Tacho. But presently there is some matrix code that needs completion which is only relevant to more than 1 rank - I am holding off on that until I have a testing foundation for it.

I’ve inserted asserts in spots I have not been actually testing if they are not currently part of the test suite.

I did not try Epetra yet so that will have build issues..

For writing data back to the MultiVector I’m not sure exactly what is best for Tpetra but if I make the simple adapter, then I expect we could just be solving directly to the adapters view and not have a copy. For Tpetra I need to work out some issues for that.

Next step is probably to make a simple non-Tpetra adapter which can handle all memory spaces to use for testing.

packages/amesos2/src/Amesos2_Kokkos_View_Copy_Assign.hpp
```diff
@@ -0,0 +1,180 @@
...
+*/
+
+#ifndef AMESOS2_KOKKOS_VIEW_COPY_ASSIGN_HPP
+#define AMESOS2_KOKKOS_VIEW_COPY_ASSIGN_HPP
+
+namespace Amesos2 {
+
+// allocate dst size if necessary - 2 methods handle 1d and 2d
+template<class dst_t, class src_t> // version for 1d view
+typename std::enable_if<std::is_same<typename dst_t::scalar_array_type,
```
mhoemmen: `Kokkos::View` has a `Rank` member.  It's `int` except for some weird specializations that use an enum, so you'll have to cast it to `int` generally:
```c++
template<class dst_t, class src_t> // version for 1d view
typename std::enable_if<static_cast<int>(dst_t::Rank) == 1>::type 
update_dst_size(dst_t& dst, const src_t& src) { /* ... */ }
```

packages/amesos2/src/Amesos2_Kokkos_View_Copy_Assign.hpp
```diff
@@ -0,0 +1,180 @@
...
+  implement_copy_or_assign(test_src, src);
+  // now we run the normal copy procedure but using the test as the src
+  implement_copy_or_assign(dst, test_src);
+}
+
+// this is the testing mode which will use the param string to determine
+// a new src memory space and copy the data to that before continuing with
+// the normal copy manager procedure.
+template<class dst_t, class src_t>
+void test_deep_copy_or_assign_view(dst_t & dst, const src_t & src) {
```
mhoemmen: Test code doesn't belong in headers that are exposed to users.

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -283,6 +293,26 @@ namespace Amesos2 {
 		     EStorage_Ordering ordering,
 		     no_special_impl nsi) const;
 
+    template<typename KV_S, typename KV_GO, typename KV_GS>
+    void help_getCrs_kokkos_view(KV_S & nzval,
+         KV_GO & colind,
+         KV_GS & rowptr,
+         global_size_t& nnz,
+         const Teuchos::Ptr<const Tpetra::Map<local_ordinal_t,global_ordinal_t,node_t> > rowmap,
```
mhoemmen: C++11 permits `>>` with no intermediate space.

packages/amesos2/src/Amesos2_MatrixAdapter_def.hpp
```diff
@@ -249,14 +265,50 @@ namespace Amesos2 {
 				     EStorage_Ordering ordering,
 				     no_special_impl nsi) const
   {
-  
+
     //Added void to remove parameter not used warning
     ((void)nsi);
```
mhoemmen: The proper idiom is not to name the parameter:
```c++
void my_function(int used_parameter, double /* unused_parameter */);
```
Some compilers will warn even if you cast to `void`, so the best thing to do is not to name the parameter in the function's definition.  If you have a separate declaration, you may name the parameter in the declaration.

packages/amesos2/src/Amesos2_MultiVecAdapter_decl.hpp
```diff
@@ -284,6 +302,28 @@ namespace Amesos2 {
               const size_t ldx);
     };
 
+    template <class MV, typename KV>
+    struct get_1d_copy_helper_kokkos_view {
+      static void
+      do_get (const Teuchos::Ptr<const MV>& mv,
```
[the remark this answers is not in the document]
mhoemmen: Oops, I realize this is one of those partial specialization of an implementation struct patterns.  Never mind :)

packages/amesos2/src/Amesos2_MatrixAdapter_def.hpp
```diff
@@ -77,6 +77,22 @@ namespace Amesos2 {
...
+  template<typename KV_S, typename KV_GO, typename KV_GS>
+  void
+  MatrixAdapter<Matrix>::getCrs_kokkos_view(KV_S & nzval,
+        KV_GO & colind,
+        KV_GS & rowptr,
+        typename MatrixAdapter<Matrix>::global_size_t& nnz,
+        const Teuchos::Ptr<const Tpetra::Map<local_ordinal_t, global_ordinal_t, node_t> > rowmap,
+        EStorage_Ordering ordering,
+        EDistribution distribution) const
+  {
```
ndellingwood: If this is the replacement call for `getCrs` for the new kokkos-view code paths, where does the input `row_map` come from? I think this is missing a call to `getDistributionMap` to create the map needed later to redistribute data to the rank 0 process, which is necessary for most of the direct solvers except superlu_dist (I think that solver hits a different code path). The issue won't be noticed running on a single MPI proc, but would show up when testing with more than one MPI process. See the `getCrs` function below for `getDistributionMap` usage.

MicheldeMessieres:
@kyungjoo-kim @ndellingwood @srajama1

Notes on the recent PR update which is still in progress.

Added a Kokkos adapter with node support.
This will currently test Serial, Cuda, and CudaUVMOff nodes using KokkosSparse::CrsMatrix.

Also improved the copy manager to handle type mismatches and allocations when required.

Now things become greatly simplified at the higher level, so for example Amesos2 TachoSolver symbolic is just:

```
  int status = 0;
  if ( this->root_ ) {
    if(do_optimization()) {
      this->matrixA_->returnRowPtr_kokkos_view(host_row_ptr_view_);
      this->matrixA_->returnColInd_kokkos_view(host_cols_view_);
    }
    data_.solver.analyze(this->globalNumCols_, host_row_ptr_view_, host_cols_view_);
  }
  return status;
```

And numeric is:

```
  int status = 0;
  if ( this->root_ ) {
    if(do_optimization()) {
     this->matrixA_->returnValues_kokkos_view(device_nzvals_view_);
    }
    data_.solver.factorize(device_nzvals_view_);
  }
  return status;
```

For the solve we used to do do_get for b, solve, then do_put for x.
But when the views can just be assigned, I think it makes more sense to have do_get for x and b and skip the do_put.
I need to add a method so the solver can know at the high level if the views will match.
Right now I am calling do_get and do_put for x.
If x will just be assigned, they both do nothing but assign the ptr.
If x will be copied, this currently adds a bit of extra overhead since we get the x values which we don’t actually need.
I plan to hold on this until I discuss with you to determine the design we want.

The copy manager performs the following tasks:

-    Allocates dst size if a deep_copy will be required - so assignment avoids allocation when not necessary.
-    Converts type mismatch and memory mismatch via deep_copy.
-    If type and memory both mismatch, deep_copy is called twice (arbitrarily doing type mismatch first).
-    If type and memory match it just assigns the view.

This copy manager could potentially be a kokkos package call?
Something like Kokkos::deep_copy_or_assign(View & dst, const View & src).

Also fixed some ETI build issues but I haven’t tested all solvers, so I need to repair other things like Basker just so they have the Kokkos version.
Or rework things so they don’t try that linkage at all but that might take a bit of restructuring.

Also fixes Epetra to work with new Kokkos view API calls.

The kokkos CrsMatrix is currently being loaded with the help of Tpetra calls.
So I’d like to eradicate that and make it completely Tpetra independent.
These are also things I’d like to improve after discussing further with you.

packages/amesos2/src/Amesos2_MatrixAdapter_def.hpp
MicheldeMessieres: Simplified the PR so just contains the Tacho serial setup currently.

packages/amesos2/src/Amesos2_Tacho_def.hpp
```diff
@@ -127,36 +127,53 @@ TachoSolver<Matrix,Vector>::symbolicFactorization_impl()
...
 
       if(std::is_same<ordinal_type, typename std::remove_pointer<col_ind_t>::type>::value) {
         // cast is so that things compile when this line is not used - same issues as above.
-        cols = ordinal_type_array((ordinal_type*)sp_colind, this->globalNumNonZeros_);
+        cols = device_ordinal_type_array((ordinal_type*)sp_colind, this->globalNumNonZeros_);
       }
       else {
-        cols = ordinal_type_array("c", this->globalNumNonZeros_);
-        for(global_size_type n = 0; n < this->globalNumNonZeros_; ++n) {
+        cols = device_ordinal_type_array("c", this->globalNumNonZeros_);
```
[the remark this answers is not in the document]
MicheldeMessieres: Resolved. This is now allocated in the new copy manager but uses uninitialized.

packages/amesos2/src/Amesos2_TpetraMultiVecAdapter_def.hpp
MicheldeMessieres: Changed to deep_copy.

packages/amesos2/src/Amesos2_MultiVecAdapter_def.hpp
```diff
@@ -134,6 +144,42 @@ namespace Amesos2{
...
+      mv->get1dCopy(mv_data, ldx, distribution_map, distribution);
+
+      // Now copy element by element in KV::execution_space
+      // Note this code is not currently tested in the default Tacho Amesos2_Solver_Test
+      // setup so is not tested anywhere. Manually tested by forcing diff_type_get_copy_kokkos_view
+      // to be always on.
+      Kokkos::parallel_for(
+        Kokkos::RangePolicy<typename KV::execution_space, int> (0, kokkos_view.size()),
+        KOKKOS_LAMBDA (const size_t & n) {
+        kokkos_view.data()[n] = Teuchos::as<typename KV::value_type>(mv_data.data()[n]);
```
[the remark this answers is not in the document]
MicheldeMessieres: Changed to deep_copy.

packages/amesos2/src/Amesos2_MultiVecAdapter_decl.hpp
```diff
@@ -241,6 +241,15 @@ namespace Amesos2 {
                         EDistribution distribution );
     };
 
+    template <typename MV, typename KV>
+    struct same_type_get_copy_kokkos_view {
+      static void apply(const Teuchos::Ptr<const MV>& mv,
```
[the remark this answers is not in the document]
MicheldeMessieres: This specific change no longer in PR but I did change other places which had this issue.

packages/amesos2/src/Amesos2_Kokkos_View_Copy_Assign.hpp
MicheldeMessieres: Implemented this change.

packages/amesos2/src/Amesos2_MultiVecAdapter_decl.hpp
```diff
@@ -312,6 +361,15 @@ namespace Amesos2 {
                         EDistribution distribution );
     };
 
+    template <typename MV, typename KV>
+    struct diff_type_data_put_kokkos_view {
+      static void apply(const Teuchos::Ptr<MV>& mv,
+                        KV& kokkos_data,
+                        const size_t& ldx,
```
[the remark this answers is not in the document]
MicheldeMessieres: This specific change no longer in the PR but I resolved the various places this appeared.

mhoemmen:
> To get around the Tpetra UVM on constraint and get a testing framework, I implemented a new execution space called CudaUVMOff. This is a copy of the Cuda execution space except it always has memory space CudaSpace, not CudaUVMSpace. So Tacho and Amesos2 are being built using the CudaUVMOff (UVM off) execution space while Tpetra stays with Cuda execution space (UVM on).

I just read that bit -- I'm not a fan of this "creating a new execution space" approach.  It's true that execution spaces have a `memory_space` typedef in them that defines their _default_ memory space, but you don't need to use the default memory space.  This is why `Kokkos::Device` and the corresponding `device_type` typedef exists.  All you need to do is consistently use `device_type = Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>` throughout.

MicheldeMessieres:
@mhoemmen TaskScheduler had some UVM on constraints making it difficult to just apply Device with UVM off. We decided to just leave Tacho UVM on until that is resolved. I've removed all the kokkos changes and now using Device to create UVM off input data for the solvers.

mhoemmen:
@MicheldeMessieres You can use any memory space you like for internal storage -- you don't have to take the input's memory space.

packages/amesos2/src/Amesos2_Tacho_decl.hpp
```diff
@@ -96,10 +98,20 @@ class TachoSolver : public SolverCore<Amesos2::TachoSolver, Matrix, Vector>
   typedef Tacho::ordinal_type                                  ordinal_type;
   typedef Tacho::size_type                                        size_type;
   typedef Kokkos::DefaultHostExecutionSpace                   HostSpaceType;
-  typedef Kokkos::TaskScheduler<HostSpaceType>                SchedulerType;
-  typedef Kokkos::View<size_type*,HostSpaceType>            size_type_array;
-  typedef Kokkos::View<ordinal_type*,HostSpaceType>      ordinal_type_array;
-  typedef Kokkos::View<tacho_type*, HostSpaceType>         value_type_array;
+
+  typedef Kokkos::DefaultExecutionSpace                      DeviceSpaceType;
+
```
kyungjoo-kim: Why do you name execution space as device space type ?

packages/amesos2/src/Amesos2_Tacho_decl.hpp
```diff
@@ -96,10 +98,20 @@ class TachoSolver : public SolverCore<Amesos2::TachoSolver, Matrix, Vector>
...
+
+  typedef Kokkos::TaskScheduler<DeviceSpaceType>              SchedulerType;
+
```
kyungjoo-kim: Task scheduler requires execution space as its template parameter.

packages/amesos2/src/Amesos2_Tacho_decl.hpp
```diff
@@ -96,10 +98,20 @@ class TachoSolver : public SolverCore<Amesos2::TachoSolver, Matrix, Vector>
...
+  typedef Kokkos::View<size_type*, DeviceSpaceType>       device_size_type_array;
+  typedef Kokkos::View<ordinal_type*, DeviceSpaceType> device_ordinal_type_array;
```
kyungjoo-kim: I would probably use "typename Tacho::UseThisDevice<exec_space>::device_type" with typedef. I lately include this fix as the scheduler requires uvm memory space but the other internal views do not require UVM. See, Tacho_Util.hpp

packages/amesos2/src/Amesos2_Tacho_decl.hpp
MicheldeMessieres: The naming was carry-over from the original code before the PR. I switched to exec_space_type_ and host_exec_space_type.

packages/amesos2/src/Amesos2_Tacho_decl.hpp
MicheldeMessieres: This will currently always be an exec space and naming has been updated.

packages/amesos2/src/Amesos2_Tacho_decl.hpp
MicheldeMessieres: Changed to now use Tacho::UseThisDevice.

packages/amesos2/src/Amesos2_Tacho_decl.hpp
```diff
@@ -204,43 +222,42 @@ class TachoSolver : public SolverCore<Amesos2::TachoSolver, Matrix, Vector>
...
+  host_ordinal_type_array host_cols_view_;
 };                              // End class Tacho
 
 
 // Specialize solver_traits struct for Tacho
 template <>
 struct solver_traits<TachoSolver> {
 #ifdef HAVE_TEUCHOS_COMPLEX
-  typedef Meta::make_list4<float,
+  typedef Meta::make_list6<float,
```
mhoemmen: We have C++11 now; we can use `std::tuple` here.

packages/amesos2/src/Amesos2_Tpetra_Impl.hpp
```diff
@@ -0,0 +1,376 @@
...
+#ifndef AMESOS2_TPETRA_IMPL_HPP
+#define AMESOS2_TPETRA_IMPL_HPP
+
+// MDM-DISCUSS Moved this Tpetra ETI into a dedicated file so we could share it
+// for different solvers. Decide if this is ok. As we add the other solvers we
+// can switch them to this file to avoid the duplication.
+
+#ifdef HAVE_TPETRA_INST_INT_INT
+#ifdef HAVE_TPETRA_INST_FLOAT
+  AMESOS2_SOLVER_TPETRA_INST(AMESOS2_TPETRA_IMPL_SOLVER_NAME,float,int,int);
```
mhoemmen: Doesn't Amesos2 have a macro now for doing instantiations?

MicheldeMessieres: @kyungjoo-kim I see now superlu was using TPETRA_INSTANTIATE_SLGN_NO_ORDINAL_SCALAR but the other solvers manually implement the ETI. I changed Tacho to also use TPETRA_INSTANTIATE_SLGN_NO_ORDINAL_SCALAR. Then I can remove the Amesos2_Tpetra_Impl.hpp file. If we want all solvers like this I could migrate them as I do the follow up PRs. If this is a problem I will just roll back.  @mhoemmen Not sure if this is what you meant. I don't know if there are any higher level macros I should be using in Amesos2.

packages/amesos2/src/Amesos2_Tacho_decl.hpp
MicheldeMessieres: @kyungjoo-kim @ndellingwood @srajama1 If we did a change for this I expect we'd want to do it as a separate PR to cover all the solvers at once. We need the compile check if the type is in the list. make_listx sets this up with a recursive head + tail variable implementation. If we used tuple I expect we need some extra machinery to scan the types. I'm not sure yet what the cleanest way to scan the tuple types in C++11 is. I can investigate if you want.

MicheldeMessieres:
@kyungjoo-kim I removed the WIP label but didn't get the auto tester to trigger. I've since made a change to ETI base on comments from @mhoemmen which I think is cleaner. So if we can approve again then I hope to see the auto test.

packages/amesos2/src/Amesos2_Tacho_decl.hpp
srajama1: @MicheldeMessieres You don't need to address this comment now. We can come back later. Just make sure to keep it in the todo.

srajama1:
@MicheldeMessieres : Looks like tests pass, can I merge this ?

MicheldeMessieres:
@srajama1 Yes, thanks. Ok to merge.

MicheldeMessieres:
@srajama1 I went ahead merged.

srajama1:
@MicheldeMessieres Thank you !

PR #5715, opened 2019-08-15
Tpetra:  temporary workaround for #5179

kddevin:
@trilinos/tpetra 

## Description
This is a temporary workaround for the most reproducible error of the errors in #5179 .
As we debug the issue, we will reverse this workaround.

## Motivation and Context
In response to user deadlines, we hope this workaround will enable users to make progress.

## How Has This Been Tested?
Tpetra tests on linux all pass
Using instructions for running MueLu_Driver.cpp in #5179, type 1 failures were eliminated.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
@@ -76,7 +76,14 @@ namespace FHT {
 // we would need experiments to find out.
 template<class ExecSpace>
 bool worthBuildingFixedHashTableInParallel () {
-  return ExecSpace::concurrency() > 1;
+//  KDD 8/19:  A temporary fix for #5179.  It appears some errors come from
+//  KDD 8/19:  computeOffsetsFromCounts; this temporary patch avoids the most
+//  KDD 8/19:  reproducible of the errors.  We'll reverse this as we debug
+//  KDD 8/19:  the problem, but this temporary fix may allow users to make 
+//  KDD 8/19:  progress toward their milestones.
+//  return ExecSpace::concurrency() > 1;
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
```diff
@@ -76,7 +76,14 @@ namespace FHT {
...
 template<class ExecSpace>
 bool worthBuildingFixedHashTableInParallel () {
-  return ExecSpace::concurrency() > 1;
+//  KDD 8/19:  A temporary fix for #5179.  It appears some errors come from
+//  KDD 8/19:  computeOffsetsFromCounts; this temporary patch avoids the most
+//  KDD 8/19:  reproducible of the errors.  We'll reverse this as we debug
+//  KDD 8/19:  the problem, but this temporary fix may allow users to make 
+//  KDD 8/19:  progress toward their milestones.
+//  return ExecSpace::concurrency() > 1;
+    return false;
```
mhoemmen: This makes me sad, but if it must be, it must be.

jhux2: @mhoemmen This is hopefully temporary.  The change buys us time to figure out the root.

jhux2: cause.

kddevin: Yup, intended to be temporary so we aren't a roadblock.

mhoemmen: Thanks for tracking this down!

PR #5717, opened 2019-08-15
Panzer: add sideset data to pamgen generated exodus output

rppawlo:
* deprecated explicit pamgen reader (combined with exodus reader now)
* added unit tests for pamgen meshes
* verification of sideset data in output meshes

note: I have not removed the old pamgen exodus reader until the apps switch their code over to the new factory. This way we can prevent CI issues in app/trilinos integration.

### DIFF
--- a/packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_ExodusReaderFactory.cpp
+++ b/packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_ExodusReaderFactory.cpp
@@ -61,6 +61,20 @@
 
 namespace panzer_stk {
 
+int getMeshDimension(const std::string & meshStr,
+                     stk::ParallelMachine parallelMach,
+                     const bool isExodus)
+{
+  stk::io::StkMeshIoBroker meshData(parallelMach);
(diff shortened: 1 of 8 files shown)


packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_ExodusReaderFactory.cpp
```diff
@@ -249,6 +271,11 @@ void STK_ExodusReaderFactory::setParameterList(const Teuchos::RCP<Teuchos::Param
 
    restartIndex_ = paramList->get<int>("Restart Index");
 
+   {
+     const auto fileType = paramList->get<std::string>("File Type");
+     isExodus_ = fileType == "Exodus" ? true : false;
```
jmgate: Is the `? true : false` needed?

rppawlo: yikes! how did that sneak in? I'll fix it.

rppawlo:
Thanks @jmgate ! Since the PR passed, I'll push this as is - want to make sure these changes get to empire tonight. I have another PR coming and will make the fix there.

PR #5720, opened 2019-08-15
MueLu Amesos2: Add nullspace fix

cgcgcg:
@trilinos/muelu 

## Description
When the near-nullspace is an actual nullspace, direct solvers will bug out. This change allows to add a rank one correction to the matrix to make it non-singular.

### DIFF
--- a/packages/muelu/src/Smoothers/MueLu_Amesos2Smoother_decl.hpp
+++ b/packages/muelu/src/Smoothers/MueLu_Amesos2Smoother_decl.hpp
@@ -91,6 +91,9 @@ namespace MueLu {
 
     //! Destructor
     virtual ~Amesos2Smoother();
+
+    RCP<const ParameterList> GetValidParameterList() const;
+
     //@}
 
     //! Input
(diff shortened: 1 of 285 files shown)


mhoemmen:
Could this be a more general thing?  Is there a reason why this would only work for Amesos2?

cgcgcg:
@mhoemmen I believe that Amesos2 gives us access to all our Tpetra compatible direct solvers. We could do the same for the Epetra stack if this turns out to be useful.

mhoemmen:
Why just direct solvers, though?  Why not do this for iterative solvers as well?

cgcgcg:
@mhoemmen The smoothers do not really care whether there is a zero eigenvalue, and the iterates would be identical (since the rhs should not have any component in the nullspace). Also, this fix is super expensive for larger matrices, since it makes the matrix dense.

mhoemmen:
@cgcgcg wrote:
> The smoothers do not really care whether there is a zero eigenvalue, and the iterates would be identical (since the rhs should not have any component in the nullspace). Also, this fix is super expensive for larger matrices, since it makes the matrix dense.

It just seems like this is a useful feature that has nothing to do with the linear solver method.  It hardly even belongs in MueLu.  I don't mean to get in your way, though :-) .

cgcgcg:
@mhoemmen I believe that we do have a projected CG in Belos that can be used to achieve a similar effect, my change essentially addresses the issue that the direct solvers bug out if we hand them a singular matrix.

Issue #5751, opened 2019-08-19
Zoltan2: Use ZOLTAN2_HAVE_EPETRA instead of Trilinos_ENABLE_EPETRA


tjfulle:
Zoltan2 uses Trilinos_ENABLE_Epetra instead of `ZOLTAN2_HAVE_EPETRA` to turn on and off Epetra support in Zoltan2 in some CMakeLists.txt files.

PR #5752, opened 2019-08-19
Zoltan2: use ZOLTAN2_HAVE_EPETRA.

tjfulle:
Closes: #5751
Part of: #5602


packages/zoltan2/CMakeLists.txt
```diff
@@ -48,9 +48,12 @@ ENDFOREACH()
 #
 # B) Set up package-specific options
 #
-
 IF (${PROJECT_NAME}_ENABLE_Epetra)
- GLOBAL_SET(HAVE_${PACKAGE_NAME_UC}_EPETRA ON)
+  IF (DEFINED Zoltan2_ENABLE_Epetra)
```
mhoemmen: Epetra is an optional library dependency of Zoltan2, so `Zoltan2_ENABLE_Epetra` should always be defined.

tjfulle:
thanks @mhoemmen !

tjfulle:
Hmmm, some versions of CMake did not like the last commit.

Issue #5592, opened 2019-07-30
Panzer: build fails with Epetra disabled


tjfulle:
## Bug Report
@trilinos/panzer 

When `Trilinos_ENABLE_Epetra:BOOL=OFF` is defined, I get the following build error:

```
packages/panzer/dof-mgr/src/Panzer_GlobalIndexer_EpetraUtilities.hpp:52:10: fatal error: Epetra_Vector.h: No such file or directory
 #include <Epetra_Vector.h>
```

Full configure:

```
/projects/sems/install/rhel6-x86_64/sems/utility/cmake/3.12.2/bin/cmake \
  -G "Ninja" \
  -D Trilinos_CONFIGURE_OPTIONS_FILE:FILEPATH=$TRILINOS_DIR/cmake/std/PullRequestLinuxCommonTestingSettings.cmake \
  -D Trilinos_ENABLE_Tpetra:BOOL=ON \
  -D Trilinos_ENABLE_Epetra:BOOL=OFF \
  -D Trilinos_ENABLE_Moertel:BOOL=OFF \
  -D Trilinos_ENABLE_ShyLU_DDFROSch:BOOL=OFF \
  -D Trilinos_ENABLE_Stokhos:BOOL=OFF \
  -D ROL_ENABLE_EXAMPLES:BOOL=OFF \
  -D Tpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
  -D Xpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
  -D Trilinos_ENABLE_Galeri:BOOL=OFF \
  -D Rythmos_ENABLE_TESTS:BOOL=OFF \
  -D Trilinos_ENABLE_FEI:BOOL=OFF \
  -D MueLu_ENABLE_TESTS:BOOL=OFF \
  $TRILINOS_DIR
```

rppawlo:
yes - panzer currently requires both the type 1 and type 2 stacks to be enabled. It's on the list of things to fix in FY20. What is the timeline/priority for this?

tjfulle:
@rppawlo, by type 1 and 2, are you referring to Epetra and Tpetra?  If so, I disabled Epetra because many packages assume that if Epetra is enabled then Tpetra’s GO=int - which is not true by default.  Those build failures will eventually be addressed in the PR.  So long as Panzer is ok with different GO in [ET]petra, or Trilinos is built with (non-default) GO=int in Tpetra, it should be fine.

[...]

PR #5756, opened 2019-08-19
Panzer: require Epetra to build

tjfulle:
Per the discussion in #5592, Panzer requires Epetra.

Closes: #5592
Part of: #5602


rppawlo:
@tjfulle - I think this should actually be added to the dof-mgr subpackage instead of core. If you prefer to push now, I can approve and we can change it later when we make epetra optional in panzer.

tjfulle:
Hey @rppawlo, I stuck it in core only because core is a required PT sub package and I wasn’t familiar with Panzer.  I have no preference where it goes, I’ll leave it up to you whether the user cannot enable Panzer if Epetra is disabled or just the dof-mgr subpackage.

PR #5799, opened 2019-08-27
This patch makes it so ctest will be slightly more efficient on GPUs

bathmatt:
by picking a device at random we should be able to up the ctest -j level on machines like waterman..

@crtrott thoughts?


jmgate:
@trilinos/kokkos

crtrott:
This looks ok. We can merge this in for now. We will figure out before the next merge what to do in Kokkos proper.

PR #5800, opened 2019-08-27
Python drivers

prwolfe:
These changes have been in develop for a bit (8/24/2019 at about 9:00 am) and will not merge into master via the attester due to bootstrapping issues in the top level sh script.  This will need to be careful considered and then for pushed to get dev->master testing back on track.


prwolfe:
After merging #5814 all the issues I found on develop should be resolved. I am going to force push this so dev -> master will run again.

PR #5805, opened 2019-08-28
ROL - change dimension to extent for kokkos deprecation

kyungjoo-kim:
@trilinos/rol , @trilinos/tpetra

## Description
With deprecation of tpetra and kokkos for downstream packages, ROL does not compile because it uses dimension. This PR changes it with extent.

## Motivation and Context
We are testing Tpetra downstream deprecation. The code does not compile.


kyungjoo-kim:
@tjfulle With this PR merged, the rocketman Tpetra test does not complain the build errors. Did you check if the test failures are the same as you experienced before ?

PR #5807, opened 2019-08-28
Amesos2 Klu2: Remove superfluous MPI_Bcast

cgcgcg:
@trilinos/amesos2

## Description
Remove a superfluous MPI_Bcast. The communicated error code is always zero, as far as I can tell.


cgcgcg:
@mhoemmen Build went through it seems. `ierr` is still returned by the function, so there should be no warning.

cgcgcg:
@mhoemmen No worries! Thanks for approving.

PR #5811, opened 2019-08-28
Tpetra: More permanent fixes for FixedHashTable issue in #5179

mhoemmen:
@trilinos/tpetra 

## Description

Please don't merge this quite just yet, per request by @kddevin .

Per request by @kddevin , here are the changes I made to FixedHashTable and computeOffsetsFromCounts, that prevent the crash observed in the MueLu driver, per discussion in #5179.   The fix makes the code cleaner and simpler.

## Related Issues

* Related to #5179 

## How Has This Been Tested?

On waterman, with the MueLu driver, and with Tpetra's tests.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
@@ -82,8 +82,9 @@ bool worthBuildingFixedHashTableInParallel () {
 //  KDD 8/19:  reproducible of the errors.  We'll reverse this as we debug
 //  KDD 8/19:  the problem, but this temporary fix may allow users to make
 //  KDD 8/19:  progress toward their milestones.
-  return ExecSpace::concurrency() > 1;
-//    return false;
+std::cout << "KDDKDD WorthIt " << (ExecSpace::concurrency() > 1) << std::endl;
+    return ExecSpace::concurrency() > 1;
+//  return false;
 //  KDD 8/19
 }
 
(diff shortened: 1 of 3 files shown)


mhoemmen:
The only test failure is an irrelevant STK SIMD test.  I think a fix for that is in progress.

The "build error" is a shadowing warning; I'll fix that.

jhux2:
@mhoemmen I'm starting to look at this PR.  I've only looked at `Tpetra_Details_FixedHashTable_def.hpp` so far. The changes there appear to be primarily for debug-mode checking (which is great, thank you).  Have I got that right, or are there changes that affect how/when the random failures occur?

mhoemmen:
@jhux2 wrote:
> Have I got that right, or are there changes that affect how/when the random failures occur?

The main change I made was to make the kernels use Kokkos::AnonymousSpace.  That made the random failures stop.  The extra debugging code was an afterthought.

mhoemmen:
@jhux2 wrote:
> Question: what motivated the introduction of `run` in the `ComputeOffsetsFromConstantCount` functor -- the use of Kokkos::AnonymousSpace, something else?

The idea is to avoid redundant instantations.  Here is the pattern.  Suppose that the interface we want is a nonmember function -- like `computeOffsetsFromConstantCount`.  Then: 

1. Implementation struct takes "canonical" template parameters, to avoid redundant specializations (e.g., `View<const T*>` vs. `View<T*>` for a read-only View).
2. Nonmember function first "canonicalizes" its template arguments, then calls the implementation struct.  This ensures that the nonmember function only instantiates the implementation struct for the minimum set of template arguments.
3. As much of the actual code as possible should go in the implementation struct, since the point is to build less redundant code.

(3) means that we put the actual `Kokkos::parallel_*` kernel launch in the implementation struct.

Issue #5390, opened 2019-06-16
ePetra: OMP statement incompatible with OMP 4.0 / GCC 9 data sharing rules


StefanBruens:
## Bug Report
@trilinos/ePetra

### Description
https://www.gnu.org/software/gcc/gcc-9/porting_to.html
> GCC releases before 9 were implementing an OpenMP 3.1 data sharing rule that const qualified variables without mutable member are predetermined shared, ... GCC 9 started implementing the OpenMP 4.0 and later behavior. When not using default clause or when using default(shared), this makes no difference, but if using default(none), previously the choice was not specify the const qualified variables on the construct at all, ...

### Steps to Reproduce
Try to build with GCC 9
```
[  299s] /home/abuild/rpmbuild/BUILD/Trilinos-trilinos-release-12-10-1/packages/epetra/src/Epetra_CrsGraph.cpp: In member function 'int Epetra_CrsGraph::OptimizeStorage()':
[  299s] /home/abuild/rpmbuild/BUILD/Trilinos-trilinos-release-12-10-1/packages/epetra/src/Epetra_CrsGraph.cpp:1949:7: error: 'numMyBlockRows' not specified in enclosing 'parallel'
[  299s]  1949 |       for(int i = 0; i < numMyBlockRows; i++) {
[  299s]       |       ^~~
[  299s] /home/abuild/rpmbuild/BUILD/Trilinos-trilinos-release-12-10-1/packages/epetra/src/Epetra_CrsGraph.cpp:1947:9: error: enclosing 'parallel'
[  299s]  1947 | #pragma omp parallel for default(none) shared(indexOffset,all_indices,indices)
[  299s]       |         ^~~
[  299s] /home/abuild/rpmbuild/BUILD/Trilinos-trilinos-release-12-10-1/packages/epetra/src/Epetra_CrsGraph.cpp:1989:6: error: 'numMyBlockRows' not specified in enclosing 'parallel'
```

rppawlo:
Just confirmed this - epetra with OpenMP enabled is broken for gcc 9.2, tested on both mac and linux machines. @kddevin - is there any plans to fix this? We have a temporary workaround in #5823 for my specific use case but it is not the correct long term fix.

ikalash:
Any idea when this issue might be resolved?  The Albany OpenMP nightly build is broken due to this issue (#5836).

rppawlo:
[...]

PR #5823, opened 2019-08-30
Epetra: allow users to disable OpenMP support in Epetra.

rppawlo:
The epetra use of OpenMP is broken for gcc 9 (see #5390 ). If we want to build trilinos with OpenMP support enabled (for kokkos and tpetra), we currently need to disable epetra. The work around to allow for epetra without openmp enabled would be to set:
```
-D EPETRA_HAVE_OMP:BOOL=OFF \
-D Epetra_HAVE_OMP:BOOL=OFF \
```
However these variables are ignored because the variables are set as normal instead of cached variables in epetra's CMakeLists.txt file. This PR changes the variables to cached so that users can override the default, allowing us to build trilinos with openmp support enabled and with epetra enabled.

NOTE: I think only the first variable listed above is used, but the second was kept around for backwards compatibility according to a comment in code.

### DIFF
--- a/packages/epetra/src/CMakeLists.txt
+++ b/packages/epetra/src/CMakeLists.txt
@@ -8,9 +8,11 @@ SET(EPETRA_MPI ${TPL_ENABLE_MPI})
 IF(CMAKE_SIZEOF_VOID_P GREATER 4)
   SET(EPETRA_ADDRESS64BIT ON)
 ENDIF()
-SET(Epetra_HAVE_OMP ${${PROJECT_NAME}_ENABLE_OpenMP})
-# This is the preferred macro to key on, but am keeping the old one for backward compatibility
-SET(EPETRA_HAVE_OMP ${${PROJECT_NAME}_ENABLE_OpenMP})
+
+TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_OpenMP
+  EPETRA_HAVE_OMP
+  "Enable OpenMP support in Epetra."
+  ${${PROJECT_NAME}_ENABLE_OpenMP})
 
 SET(HAVE_BLAS ON)
 SET(HAVE_LAPACK ON)


rppawlo:
ok - will change that. I just changed the actual variable used in the config file generation.

rppawlo:
@bartlettroscoe - is Epetra_ENABLE_OpenMP automatically set as a cache variable by tribits? If so then it is just the logic for the option in the config file that needs to change.

bartlettroscoe:
@rppawlo asked:

>  @bartlettroscoe - is Epetra_ENABLE_OpenMP automatically set as a cache variable by tribits? If so then it is just the logic for the option in the config file that needs to change.

It does not look like it.  Looking at [CMakeCache.txt](https://testing-dev.sandia.gov/cdash/viewNotes.php?buildid=4818898#!#note0) for an OpenMP build, we only see the Epetra cache vars:

```
Epetra_ENABLE_ABC:BOOL=OFF
Epetra_ENABLE_DEBUG:BOOL=OFF
Epetra_ENABLE_EXAMPLES:BOOL=OFF
Epetra_ENABLE_FATAL_MESSAGES:BOOL=OFF
Epetra_ENABLE_FORMAT_IO:BOOL=OFF
Epetra_ENABLE_Fortran:BOOL=ON
Epetra_ENABLE_TESTS:BOOL=OFF
Epetra_ENABLE_THREADS:BOOL=OFF
Epetra_ENABLE_WARNING_MESSAGES:BOOL=OFF
Epetra_HIDE_DEPRECATED_CODE:BOOL=OFF
Epetra_SHOW_DEPRECATED_WARNINGS:BOOL=ON
```

with:

```
Trilinos_ENABLE_OpenMP:BOOL=ON
```

Therefore, it does not look like we implemented a function called something like `TRIBITS_ADD_OpenMP_OPTION()` analogous to [TRIBITS_ADD_DEBUG_OPTION()](https://tribits.org/doc/TribitsDevelopersGuide.html#tribits-add-debug-option).

Since this is a special use case (to address legacy code), I think the code is very simple to add to `epetra/CMakeLists.txt`:

```
  ${PACKAGE_NAME}_ENABLE_OpenMP      # CMake Cache var name
  EPETRA_ENABLE_OMP      # C/C++ macro name
  "Enable support for OpenMP in Epetra (default given by ${PROJECT_NAME}_ENABLE_OpenMP)"
  ${${PROJECT_NAME}_ENABLE_OpenMP}
  )
# Backward compatibility define (that no one should be using)
SET(Epetra_HAVE_OMP ${EPETRA_ENABLE_OMP })
```

And then downstream packages should be reading global CMake cache var `Epetra_ENABLE_OpenMP`.  It is only internal Epetra C++ code that should be reading the C/C++ preprocessor `EPETRA_ENABLE_OMP`.

Does that seem reasonable?

BTW, who is doing an GCC 9.x build?  We don't have any automated builds for GCC 9.x yet do we?  I think we could a GCC 9.x build using Spack at this point if we wanted to.

rppawlo:
@bartlettroscoe - just force pushed a new fix. Can you review again? I also removed the deprecated variable since a search on all of Trilinos didn't show it being used anywhere.

rppawlo:
I don't know of any clusters using gcc 9, but my workstation was rebuilt yesterday so I upgraded my local tool set to 9.2 and started seeing this. The default gcc in mac brew is also 9.2.

PR #5831, opened 2019-08-30
Tempus: Missing Args in explicit Steppers InArgs.

ccober6:
* Added new parameters for explicit evaluation, xDot = f(x,t,p).
   Parameters for explicit steppers, p, are currently dt and stage number.

 * Cleaned up implicit parameters (e.g., included stage number in
   argument list).

 * StepperTrapezoidal was missing the new solveImplicitODE().

Passes all tests on my Mac.

@trilinos/tempus

### DIFF
--- a/packages/tempus/src/Tempus_StepperExplicit_decl.hpp
+++ b/packages/tempus/src/Tempus_StepperExplicit_decl.hpp
@@ -13,6 +13,24 @@
 #include "Tempus_Stepper.hpp"
 
 
+template<class Scalar>
+class ExplicitODEParameters
+{
+  public:
+    /// Constructor
+    ExplicitODEParameters()
+      : timeStepSize_(Scalar(0.0)), stageNumber_(0)
+    {}
+    /// Constructor
+    ExplicitODEParameters(Scalar timeStepSize, int stageNumber = 0)
+      : timeStepSize_(timeStepSize), stageNumber_(stageNumber)
+    {}
+
+    Scalar                                timeStepSize_;
+    int                                   stageNumber_;
+};
+
+
 namespace Tempus {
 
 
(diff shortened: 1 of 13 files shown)


ccober6:
@sconde @michael-a-hansen @ikalash can you check this branch against your codes?  The major change is for Stepper::evaluateExplicitODE(xDot, x, t, p).  It now has a parameter object, p, to pass through to the ModelEvaluator and supply dt and stage number.  This should only affect specialized Steppers you have made.  Let me know if you have questions.

sconde:
@ccober6 Why would a ME need the stage number of an RK scheme?

ccober6:
It shouldn't, but it was a request from John Shadid several years ago and is in the implicit MEs.  Also MEs should not need dt either, but applications seem to want it.  :)

ikalash:
Sure I can test it.  Will let you know when I’ve had the chance to do so.

sconde:
This builds and passes for Drekar.

ikalash:
Just tested Albany - everything passes.

ccober6:
Thanks @sconde and @ikalash!  Just waiting on @Michael-a-hansen, and I should be able to merge this.

packages/tempus/src/Tempus_StepperExplicit_decl.hpp
```diff
@@ -13,6 +13,24 @@
...
+    ExplicitODEParameters()
+      : timeStepSize_(Scalar(0.0)), stageNumber_(0)
+    {}
+    /// Constructor
+    ExplicitODEParameters(Scalar timeStepSize, int stageNumber = 0)
+      : timeStepSize_(timeStepSize), stageNumber_(stageNumber)
+    {}
+
+    Scalar                                timeStepSize_;
+    int                                   stageNumber_;
```
mhoemmen: Prefer setting default values like this, rather than in the constructor.
```c++
Scalar timeStepSize_ = 0.0;
int stageNumber_ = 0;
```
Then you can declare the default constructor like this:
```c++
ExplicitODEParameters() = default;
```

packages/tempus/src/Tempus_StepperExplicit_impl.hpp
```diff
@@ -108,8 +108,9 @@ void StepperExplicit<Scalar>::setInitialConditions(
         *out << "Warning -- Requested IC consistency of 'None' but\n"
              << "           initialState does not have an xDot.\n"
              << "           Setting a 'Consistent' xDot!\n" << std::endl;
+        auto p = Teuchos::rcp(new ExplicitODEParameters<Scalar>(0.0));
```
mhoemmen: The default constructor suffices here; `0.0` is not needed.

Issue #5838, opened 2019-09-02
MueLu: kokkos refactor of uncoupled aggregation and amalgamation process


lucbv:
## Enhancement
@trilinos/muelu 

This issue will track the progress made to integrate all changes from the branch [MueLu_uncoupled_aggregation_refactor](https://github.com/trilinos/Trilinos/tree/MueLu_uncoupled_aggregation_refactor). The aggregation refactor was done about a year ago and cannot be merged easily anymore so merging the previously mentioned branch will not be attempted.
Instead the changes in that branch will broken into smaller incremental changes with the hope that these incremental changes will be merged quickly and show a more logical progression.
As I tried to merge the old branch I also realized that some changes are potentially poorly designed and I am planning on modifying the work a bit to accommodate better design changes.
Here is a list of new features that are needed to complete the refactor work:

- [x] `MueLu::LWGraph` needs to new methods: `getRowPtrs()` and `getEntries()` to avoid awkward extraction of data from `graph_` see [here](https://github.com/trilinos/Trilinos/blob/21ec71443e69f66e2af1fad742bbdc8afb8411e7/packages/muelu/src/Graph/UncoupledAggregation/MueLu_AggregationPhase1Algorithm_kokkos_def.hpp#L280) and [here](https://github.com/trilinos/Trilinos/blob/21ec71443e69f66e2af1fad742bbdc8afb8411e7/packages/muelu/src/Graph/UncoupledAggregation/MueLu_AggregationPhase1Algorithm_kokkos_def.hpp#L329)
- [x] instead of modifying heavily the signature of `MueLu:: AggregationAlgorithmBase_kokkos::BuildAggregates(...)` I would like to update it to use a `Kokkos::View` instead of a `std::vector` for portability reasons and store pointers to additional data such as coloring information in `MueLu::Aggregates_kokkos`. This will avoid clashes with `StructuredAggregation` which does not require coloring information and already stores a pointer to a `MueLu::IndexManager_kokkos`
- [x] re-write the uncoupled aggregation phases using the work done in the MueLu_uncoupled_aggregation_refactor branch allowing for both non-deterministic and deterministic aggregation options
- [x] create `MueLu::AmalgamationInfo_kokkos` and `MueLu::AmalgamationFactory_kokkos` to set the infrastructure in MueLu for the related algorithms and data structured to be ported to kokkos.
[...]

PR #5841, opened 2019-09-03
MueLu: extending the interface of LWGraph to give direct access to rowPtrs and entries

lucbv:
@trilinos/muelu

## Description
I am simply adding two host / device function that give access to the row pointers and entries from the `Kokkos::StaticCrsGraph`.

## Motivation and Context
Accessing the underlying local data avoid having to use methods that transfer data to host which is not always the right thing to do.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of #5838
* Composed of 

## How Has This Been Tested?
I did  a local build and things seem OK so far.

## Checklist

### DIFF
--- a/packages/muelu/src/Graph/Containers/MueLu_LWGraph_kokkos_decl.hpp
+++ b/packages/muelu/src/Graph/Containers/MueLu_LWGraph_kokkos_decl.hpp
@@ -132,6 +132,16 @@ namespace MueLu {
       return graph_.row_map(GetNodeNumVertices());
     }
 
+    //! Return the row pointers of the local graph
+    KOKKOS_INLINE_FUNCTION typename local_graph_type::row_map_type getRowPtrs() const {
+      return graph_.row_map;
+    }
+
+    //! Return the list entries in the local graph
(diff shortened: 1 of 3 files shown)


lucbv:
I am wondering if this point we should just add a method that gives direct access to the `StaticCrsGraph` something like:

```
    //! Return the local graph
    KOKKOS_INLINE_FUNCTION local_graph_type getLocalGraph() const {
      return graph_;
    }
```

@jhux2 @csiefer2 what do you think?

jhux2:
> I am wondering if at this point we should just add a method that gives direct access to the `StaticCrsGraph` something like:

Yes, if you want something else about the graph besides the rowmap and entries.

mhoemmen:
@lucbv I am partial to the `withLocalAccess` style of accessing local data:
```c++
withLocalAccess ([&] (const local_graph_type& lclGraph) { /* ... */ }, gblGraph);
```

lucbv:
Adding a unit test for LWGraph that test the two new methods.

lucbv:
It turns out if you don't commit the source code for the test it does not work that well...

PR #5845, opened 2019-09-03
WIP: initial try at an ATDM build on doom

fryeguy52:
@bartlettroscoe 
## Description
Additions to the ATDM infrastructure to add support on doom. currently not working


cmake/std/atdm/utils/set_build_options.sh
```diff
@@ -89,6 +89,8 @@ elif [[ $ATDM_CONFIG_BUILD_NAME == *"cuda-9.0"* ]]; then
 elif [[ $ATDM_CONFIG_BUILD_NAME == *"cuda-9.2-gnu-7.2.0"* ]] \
   || [[ $ATDM_CONFIG_BUILD_NAME == *"cuda-9.2_gnu-7.2.0"* ]]; then
   export ATDM_CONFIG_COMPILER=CUDA-9.2_GNU-7.2.0
+elif [[ $ATDM_CONFIG_BUILD_NAME == *"cuda-9.2_gnu-6.3.1_openmpi-2.1.1"* ]]; then
```
bartlettroscoe: I think this is too specialized to put into this file.  Can you copy the example in cee-rhel6/ and use a custom_builds.sh script for this?

cmake/std/atdm/doom/environment.sh
```diff
@@ -0,0 +1,292 @@
...
+# <outout_file>.  The job name for the sbatch script is taken from the env var
+# 'ATDM_CONFIG_BUILD_NAME'.  This works for local builds since ATDM_CONFIG_BUILD_NAME.
+#
+# Note that you can pass in the script to run with arguments such as with
+# "<some-script> <arg1> <arg2>" and it will work.  But note that this has to
+# be bash script that 'sbatch' can copy and run form a temp location and it
+# still has to work.  So the script has to use absolute directory paths, not
+# relative paths or asume sym links, etc.
+#
+function atdm_run_script_on_compute_node {
```
bartlettroscoe: It would be good to find a way to not duplicate this function.  Could we put this in its own file and then reuse this across all system envs that use it?

cmake/std/atdm/utils/get_known_system_info.sh
```diff
@@ -66,6 +66,9 @@ elif [[ $ATDM_CONFIG_REAL_HOSTNAME == "ghost"* ]] \
   || [[ $ATDM_CONFIG_REAL_HOSTNAME =~ gho[0-9]+ ]] ; then
   ATDM_HOSTNAME=ghost
   ATDM_SYSTEM_NAME=serrano
+elif [[ $ATDM_CONFIG_REAL_HOSTNAME == "doom"* ]] ; then
```
bartlettroscoe: So this would change from 'doom' to 'cts1'.  (Let's see if we can get that to work.)

bartlettroscoe:
The current PR for this is #6071.  Closing this PR.

PR #5857, opened 2019-09-04
Tpetra::Import: Add debug checks to createRemoteOnlyImport

mhoemmen:
@trilinos/tpetra

Add extra debug checks and output to `Tpetra::Import::createRemoteOnlyImport`.  This is part of #5179 debugging.  I don't think anything is wrong with that function, but I suspect that its input may be wrong in some way.


jjellio:
Do you want me to merge/test before approving? And let this be a WIP?

PR #5861, opened 2019-09-05
Tpetra: Adding block-diagonal extraction routine

csiefer2:
Auto-PR for SHA a1cfb13

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
@@ -45,6 +45,7 @@
 #include "TpetraCore_config.h"
 #include "Tpetra_CrsMatrix.hpp"
 #include "Teuchos_RCP.hpp"
+#include "Tpetra_Details_Behavior.hpp"
 
 /// \file Tpetra_Details_extractBlockDiagonal.hpp
 /// \brief Functions that allow for the extraction of a block diagonal
@@ -61,6 +62,7 @@ template<class SparseMatrixType,
          class MultiVectorType>
 void extractBlockDiagonal(const SparseMatrixType& A, MultiVectorType & diagonal) {
   using local_map_type = typename SparseMatrixType::map_type::local_map_type;
+  using SC             = typename MultiVectorType::scalar_type;
   using LO             = typename SparseMatrixType::local_ordinal_type;
   using KCRS           = typename SparseMatrixType::local_matrix_type;
   using lno_view_t     = typename KCRS::StaticCrsGraphType::row_map_type::const_type;
@@ -70,27 +72,31 @@ void extractBlockDiagonal(const SparseMatrixType& A, MultiVectorType & diagonal)
   using range_type     = Kokkos::RangePolicy<typename SparseMatrixType::node_type::execution_space, size_t>;
 
   // Sanity checking: Map Compatibility (A's rowmap matches diagonal's map)
-  TEUCHOS_TEST_FOR_EXCEPTION(!A.getRowMap()->isSameAs(*diagonal.getMap()),
-    std::runtime_error, "Tpetra::Details::extractBlockDiagonal was given incompatible maps");
+  if (Tpetra::Details::Behavior::debug() == true) {
+    TEUCHOS_TEST_FOR_EXCEPTION(!A.getRowMap()->isSameAs(*diagonal.getMap()),
+       std::runtime_error, "Tpetra::Details::extractBlockDiagonal was given incompatible maps");
+  }
 
-  int numrows   = diagonal.getLocalLength();
-  int blocksize = diagonal.getNumVectors();
-  diagonal.putScalar(Teuchos::ScalarTraits<typename MultiVectorType::scalar_type>::zero());
+  LO numrows   = diagonal.getLocalLength();
+  LO blocksize = diagonal.getNumVectors();
+  SC ZERO = Teuchos::ScalarTraits<typename MultiVectorType::scalar_type>::zero();
 
   // Get Kokkos versions of objects
   local_map_type rowmap  = A.getRowMap()->getLocalMap();
   local_map_type colmap  = A.getRowMap()->getLocalMap();
   local_mv_type diag     = diagonal.getLocalViewDevice();
-  const KCRS & Amat      = A.getLocalMatrix();
+  const KCRS   Amat      = A.getLocalMatrix();
   lno_view_t Arowptr     = Amat.graph.row_map;
   lno_nnz_view_t Acolind = Amat.graph.entries;
   scalar_view_t Avals    = Amat.values;
 
-  Kokkos::parallel_for("Tpetra::extractBlockDiagonal",range_type(0,numrows),[=](const size_t i){
+  Kokkos::parallel_for("Tpetra::extractBlockDiagonal",range_type(0,numrows),KOKKOS_LAMBDA(const LO i){
       LO diag_col   = colmap.getLocalElement(rowmap.getGlobalElement(i));
       LO blockStart = diag_col - blocksize + 1;
       LO blockStop  = diag_col + blocksize - 1;
-      
+      for(LO k=0; k<blocksize; k++)
+        diag(i,k)=ZERO;
+
       for (size_t k = Arowptr(i); k < Arowptr(i+1); k++) {
         LO col = Acolind(k);
         if (blockStart <= col && col <= blockStop) {
--- a/packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
@@ -92,8 +92,8 @@ void extractBlockDiagonal(const SparseMatrixType& A, MultiVectorType & diagonal)
 
   Kokkos::parallel_for("Tpetra::extractBlockDiagonal",range_type(0,numrows),KOKKOS_LAMBDA(const LO i){
       LO diag_col   = colmap.getLocalElement(rowmap.getGlobalElement(i));
-      LO blockStart = diag_col - blocksize + 1;
-      LO blockStop  = diag_col + blocksize - 1;
+      LO blockStart = diag_col - (diag_col % blocksize);
+      LO blockStop  = blockStart + blocksize;
       for(LO k=0; k<blocksize; k++)
         diag(i,k)=ZERO;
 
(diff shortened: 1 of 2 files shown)


packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+  using local_map_type = typename SparseMatrixType::map_type::local_map_type;
+  using LO             = typename SparseMatrixType::local_ordinal_type;
+  using KCRS           = typename SparseMatrixType::local_matrix_type;
+  using lno_view_t     = typename KCRS::StaticCrsGraphType::row_map_type::const_type;
+  using lno_nnz_view_t = typename KCRS::StaticCrsGraphType::entries_type::const_type;
+  using scalar_view_t  = typename KCRS::values_type::const_type;
+  using local_mv_type  = typename MultiVectorType::dual_view_type::t_dev;
+  using range_type     = Kokkos::RangePolicy<typename SparseMatrixType::node_type::execution_space, size_t>;
+
+  // Sanity checking: Map Compatibility (A's rowmap matches diagonal's map)
```
mhoemmen: It would be nicer to do this check only in debug mode (`Tpetra::Details::Behavior::debug() == true`).

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+  TEUCHOS_TEST_FOR_EXCEPTION(!A.getRowMap()->isSameAs(*diagonal.getMap()),
+    std::runtime_error, "Tpetra::Details::extractBlockDiagonal was given incompatible maps");
+
+  int numrows   = diagonal.getLocalLength();
```
mhoemmen: This may cause a build warning (assignment of unsigned to signed).  You can avoid it like this:
```c++
int numrows (diagonal.getLocalLength ());
```
It would be better to use the local ordinal type:
```c++
LO numrows (diagonal.getLocalLength ());
```

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+  int blocksize = diagonal.getNumVectors();
+  diagonal.putScalar(Teuchos::ScalarTraits<typename MultiVectorType::scalar_type>::zero());
```
mhoemmen: This can be fused into the kernel below.

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+         class MultiVectorType>
+void extractBlockDiagonal(const SparseMatrixType& A, MultiVectorType & diagonal) {
+  using local_map_type = typename SparseMatrixType::map_type::local_map_type;
+  using LO             = typename SparseMatrixType::local_ordinal_type;
+  using KCRS           = typename SparseMatrixType::local_matrix_type;
+  using lno_view_t     = typename KCRS::StaticCrsGraphType::row_map_type::const_type;
+  using lno_nnz_view_t = typename KCRS::StaticCrsGraphType::entries_type::const_type;
+  using scalar_view_t  = typename KCRS::values_type::const_type;
+  using local_mv_type  = typename MultiVectorType::dual_view_type::t_dev;
+  using range_type     = Kokkos::RangePolicy<typename SparseMatrixType::node_type::execution_space, size_t>;
```
mhoemmen: 1. This means that you intend to run on device, if `execution_space` is `Cuda`.  See below.
2. The most appropriate index type is `LO`, not `size_t`, because you're iterating over local rows.

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+  int numrows   = diagonal.getLocalLength();
+  int blocksize = diagonal.getNumVectors();
+  diagonal.putScalar(Teuchos::ScalarTraits<typename MultiVectorType::scalar_type>::zero());
+
+  // Get Kokkos versions of objects
+  local_map_type rowmap  = A.getRowMap()->getLocalMap();
+  local_map_type colmap  = A.getRowMap()->getLocalMap();
+  local_mv_type diag     = diagonal.getLocalViewDevice();
+  const KCRS & Amat      = A.getLocalMatrix();
```
mhoemmen: Since you intend to run on device, best practice would be to make `Amat` a value, rather than a reference.  Capturing a reference for a device kernel is ungood.

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+  lno_view_t Arowptr     = Amat.graph.row_map;
+  lno_nnz_view_t Acolind = Amat.graph.entries;
+  scalar_view_t Avals    = Amat.values;
+
+  Kokkos::parallel_for("Tpetra::extractBlockDiagonal",range_type(0,numrows),[=](const size_t i){
```
mhoemmen: If you're intending to run on device, you'll want to mark the lambda with the appropriate macro.  I've also changed the index type (see note above).
```c++
Kokkos::parallel_for("Tpetra::extractBlockDiagonal", range_type(0, numrows), KOKKOS_LAMBDA (const LO i) { /* ... */ });
```

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+    std::runtime_error, "Tpetra::Details::extractBlockDiagonal was given incompatible maps");
+
+  int numrows   = diagonal.getLocalLength();
+  int blocksize = diagonal.getNumVectors();
+  diagonal.putScalar(Teuchos::ScalarTraits<typename MultiVectorType::scalar_type>::zero());
+
+  // Get Kokkos versions of objects
+  local_map_type rowmap  = A.getRowMap()->getLocalMap();
+  local_map_type colmap  = A.getRowMap()->getLocalMap();
+  local_mv_type diag     = diagonal.getLocalViewDevice();
```
mhoemmen: Consider using `withLocalAccess` to access the MultiVector.  There is an example in `tpetra/core/test/MultiVector`.

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+///   of Tpetra.  The file itself or its contents may disappear or
+///   change at any time.
+
+namespace Tpetra {
+namespace Details {
+
+
+template<class SparseMatrixType,
+         class MultiVectorType>
+void extractBlockDiagonal(const SparseMatrixType& A, MultiVectorType & diagonal) {
```
mhoemmen: Do you really want to extract into a MultiVector?  It seems like the lowest-level interface would extract into a rank-3 `Kokkos::View`, where the leftmost index is the local block row index.

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+  Kokkos::parallel_for("Tpetra::extractBlockDiagonal",range_type(0,numrows),[=](const size_t i){
+      LO diag_col   = colmap.getLocalElement(rowmap.getGlobalElement(i));
+      LO blockStart = diag_col - blocksize + 1;
+      LO blockStop  = diag_col + blocksize - 1;
+      
+      for (size_t k = Arowptr(i); k < Arowptr(i+1); k++) {
+        LO col = Acolind(k);
+        if (blockStart <= col && col <= blockStop) {
+          diag(i,col-blockStart) = Avals(k);
```
mhoemmen: I'm a little bit confused about the purpose of this kernel.  It looks like you're flattening out the blocks into a row of the MultiVector; is that right?

csiefer2:
> I'm a little bit confused about the purpose of this kernel. It looks like you're flattening out the blocks into a row of the MultiVector; is that right?

Correct.

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: Easily fixed

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: That guy actually returns a size_t.

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: Indeed.

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: Fixed

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: Fixed

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: Yes

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: Fixed

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: But, hey.  LO is smaller.

mhoemmen:
It looks like this is actually a band extraction routine.  Perhaps we should call it that instead?

packages/tpetra/core/src/Tpetra_Details_extractBlockDiagonal.hpp
csiefer2: > It looks like this is actually a band extraction routine. Perhaps we should call it that instead?

Which is isn't supposed to be.  Good catch!

csiefer2:
> It looks like this is actually a band extraction routine. Perhaps we should call it that instead?

Which it isn't supposed to be.  Good catch.

PR #5873, opened 2019-09-06
MueLu RefMaxwell: Apply nullspace fix when no BCs are detected

cgcgcg:
@trilinos/muelu

## Description
Automatically apply nullspace fix to avoid singular matrix in RefMaxwell when no boundary conditions were detected.


packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
```diff
@@ -786,6 +787,11 @@ namespace MueLu {
         if (!reuse) {
           ParameterList& userParamList = precList22_.sublist("user data");
           userParamList.set<RCP<RealValuedMultiVector> >("Coordinates", Coords_);
+          // If we detected no boundary conditions, the (2,2) problem is singular
+          if (BCrowcount_ == 0 &&
+              !precList22_.isSublist("coarse: params") &&
```
jhux2: Don't you want this sublist to exist?

cgcgcg: The logic is: if we didn't already set the parameter before, then set it here. I don't want to clobber the preexisting value.

jhux2: :vulcan_salute:

PR #5877, opened 2019-09-07
Epetra: fix GCC 9 OpenMP errors #5390

### LINKED ISSUE #5390: ePetra: OMP statement incompatible with OMP 4.0 / GCC 9 data sharing rules

brian-kelley:
Add variables to shared clause. This fixes build errors in GCC 9
which is more strict about variable usage in parallel sections.

@trilinos/epetra

## Motivation and Context

Fixes build on GCC 9 if Epetra_ENABLE_OpenMP=ON

## How Has This Been Tested?

Built on my workstation with GCC 9.1 and OpenMP enabled.

## Checklist

### DIFF
--- a/packages/epetra/src/Epetra_CrsGraph.cpp
+++ b/packages/epetra/src/Epetra_CrsGraph.cpp
@@ -1938,7 +1938,7 @@ int Epetra_CrsGraph::OptimizeStorage() {
 
     if (!(StaticProfile())) {
 #ifdef EPETRA_HAVE_OMP
-#pragma omp parallel for default(none) shared(indexOffset,all_indices,indices)
+#pragma omp parallel for default(none) shared(indexOffset,all_indices,indices,numMyBlockRows)
 #endif
       for(int i = 0; i < numMyBlockRows; i++) {
   int numColIndices = indexOffset[i+1] - indexOffset[i];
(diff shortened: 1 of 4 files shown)


brian-kelley:
Well, it looks like we have a catch-22 with the different GCC versions. 9+ requires constant local variables to be in the "shared" clause when there's "default(none)". But GCC 4 doesn't support explicitly shared const variables ([see this stackoverflow question](https://stackoverflow.com/questions/13199398/openmp-predetermined-shared-for-shared/13770822#comment30469273_13770822). default(none) is there to prevent variables being shared implicitly and causing data races, which seems much more important for preventing bugs than having "const" on the variables in question, so I'm just removing the const. Now it works on all compiler versions.

PR #5912, opened 2019-09-16
Update stk to Trilinos

alanw0:
Bring in ngp::Field versions of parallel_sum, copy_owned_to_shared, and
communicate_field_data.

Also fix some openmp compile issues reported by Roger Pawlowski.

@trilinos/<teamName>

## Checklist

### DIFF
--- a/packages/stk/stk_ngp/stk_ngp/NgpField.hpp
+++ b/packages/stk/stk_ngp/stk_ngp/NgpField.hpp
@@ -205,9 +205,9 @@ private:
 private:
 #endif
 
-    void swap_data(ConstStkFieldAdapter<T> &sf) { }
+    void swap(ConstStkFieldAdapter<T> &sf) { }
 
-    void swap_data(StkFieldAdapter<T> &sf) { }
+    void swap(StkFieldAdapter<T> &sf) { }
 
     bool need_sync_to_host() const { return false; }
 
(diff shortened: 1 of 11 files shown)


packages/stk/stk_ngp/stk_ngp/NgpField.hpp
```diff
@@ -379,6 +379,14 @@ class StaticField : public FieldBase
 
     const stk::mesh::BulkData& get_bulk() const { return *hostBulk; }
 
+    STK_FUNCTION
+    void swap_data(StaticField<T> &sf)
```
mhoemmen: Why not just call this "swap"?

alanw0: Yes, swap would be just as good here. Maybe there's even a better name. The intent is to do a "shallow" swap. I.e., in traditional sierra-framework fields it was just a pointer swap, not a copying of the data. It is for managing multi-state fields, e.g., an 'n' and 'n+1' state.

mhoemmen: @alanw0 It looks like this means the same thing as, say, `std::shared_ptr` swap:

https://en.cppreference.com/w/cpp/memory/shared_ptr/swap

If so, then I think `swap` is the right name for this function.

alanw0: done.

PR #5913, opened 2019-09-16
Moved from gettime to getpid for parallel jobs

bathmatt:
Realized that get time returns time in seconds, so mpirun -np 4 assigns to the same device...  Moving to getpid()


ndellingwood:
@bathmatt this is now merged.

ndellingwood:
Cross-referencing again to kokkos/kokkos#2254, apparently links cross-referencing other projects within a review don't pop up.

Issue #5914, opened 2019-09-16
MPI_EXEC_PRE_NUMPROCS_FLAGS:STRING not propagated to install cmake files


bathmatt:
## Enhancement
@trilinos/

It doesn't look like MPI_EXEC_PRE_NUMPROCS_FLAGS:STRING=--bind-to;none is being saved in any of the cmake install files.  

searching the install dir I see 
(DESKTOP)-dir-GNU-7.2.0-RELEASE-OPENMP-shared [GNU-7.2.0]$ 59 $ find . -type f -exec grep -I bind-to {} \;
export ATDM_CONFIG_MPI_PRE_FLAGS="--bind-to;none"

Something like Trilinos_MPI_EXEC_PRE_NUMPROCS_FLAGS

would be good

bartlettroscoe:
CC: @jfrye

@bathmatt,

I will take a look at this and see if this can be resolved.

bartlettroscoe:
@bathmatt,

Looks like this is already supported as the env var `ATDM_CONFIG_MPI_PRE_FLAGS`:

```
cmake/std/atdm/ATDMDevEnvSettings.cmake:255:ATDM_SET_CACHE(MPI_EXEC_PRE_NUMPROCS_FLAGS "$ENV{ATDM_CONFIG_MPI_PRE_FLAGS}" CACHE STRING)
```

And there is lots of examples of how this gets used in:

```
$ find cmake/std/atdm -type f -exec grep -nH ATDM_CONFIG_MPI_PRE_FLAGS {} \;
cmake/std/atdm/utils/unset_atdm_config_vars_environment.sh:43:unset ATDM_CONFIG_MPI_PRE_FLAGS
cmake/std/atdm/sems-rhel6/environment.sh:204:export ATDM_CONFIG_MPI_PRE_FLAGS="--bind-to;none"
cmake/std/atdm/cee-rhel6/environment.sh:94:  export ATDM_CONFIG_MPI_PRE_FLAGS="--bind-to;none"
cmake/std/atdm/cee-rhel6/environment.sh:107:  export ATDM_CONFIG_MPI_PRE_FLAGS="--bind-to;none"
cmake/std/atdm/examples/my-sems-rhel6-config/environment.sh:43:export ATDM_CONFIG_MPI_PRE_FLAGS="--bind-to;none"
[...]

PR #5916, opened 2019-09-16
Add Trilinos_MPI_EXEC_[PRE|POST]_NUMPROCS_FLAGS to TrilinosConfig.cmake (#5914)

bartlettroscoe:
Addresses #5914  (requested by EMPIRE).

This has strong unit testing in TriBITS so I am pretty confidence this is doing what it is supposed to do (and it is so simple that is is likely correct by inspection).


bartlettroscoe:
FYI: This changed a TriBITS *.cmake file so it will trigger testing of every Trilinos PT package on every platform.

PR #5920, opened 2019-09-17
Seacas sept snapshot

gsjaardema:
@trilinos/seacas 

## Description

    * Primarily spelling fixes caught by codespell
    * Update fmt:: library
    * Add comma-separated output to explore summary
    * Add some C++11 =default and =delete
    * Remove some leftover `std::numpunct` usages
    * Fix array overrun in exodus `ex_put_init_ext` name output
    * Clean up exodus format logic to unify serial and parallel
    * Skinner now skins block blocks instead of entire model [experimental]
    * Add `format_id_list` function which gathers ranges:
      -- 1-10, 13-15, 17, 18 instead of 1,2,3,4,5,6,7,8,9,10,13,14,15,17,18
      -- range and sequence separators can be specified
      -- Add use of this in a few error/warning message outputs.
    * Add Utst_utils.C to test some functions in Ioss::Utils class (`number_width` and `format_id_list`)
    * Add `Ioss::ParallelUtils::barrier()` to eliminate some `#ifdef SEACAS_HAVE_MPI` uses in code
    * Refactor Ioss::Property API to make safer to use. Eliminated defaulted argument and a never-used function
    * Better high-water memory and normal memory use routines.
    * Doxygen comment cleanup
    * Make field length optional -- can be obtained from GE that field is added to.
    * Fix size of embedded NodeBlock in an empty StructuredBlock: 0 not 1
    * Use RAII for MPI_Finalize
    * Improved `io_info` output for parallel runs.
    * Fix version handling in exodus-*.py.in
## Checklist


gsjaardema:
Sorry for the large amount of changes -- most of them are spelling fixes, but that doesn't make it any easier to review...

packages/seacas/applications/blot/cnvmax.f
```diff
@@ -48,7 +48,7 @@ SUBROUTINE CNVMAX (NUMNPF, VARNP, IN2ELB, NMIN, NMAX, FMIN, FMAX)
 C   --   Written by Amy Gilkey - revised 03/09/88
```
mhoemmen: Let no one forget that our software has a long lifetime :-)

packages/seacas/applications/conjoin/CJ_ExodusFile.h
```diff
@@ -43,6 +43,8 @@ namespace Excn {
   public:
     explicit ExodusFile(size_t which);
     ~ExodusFile();
+    ExodusFile(const ExodusFile &) = delete;
+    ExodusFile operator=(const ExodusFile &) = delete;
```
mhoemmen: The usual idiom is to disable move construction and assignment as well.

packages/seacas/libraries/aprepro_lib/apr_builtin.cc
```diff
@@ -179,7 +179,7 @@ namespace SEAMS {
     return (temp);
   }
 
-  // DO_HYPOT: calcluate sqrt(p^2 + q^2)
+  // DO_HYPOT: calculate sqrt(p^2 + q^2)
```
mhoemmen: C++ has `std::hypot` now.

packages/seacas/libraries/ioss/src/Ioss_Utils.C
```diff
@@ -41,9 +41,9 @@
 #include <cstdlib>
 #include <cstring>
 #include <ctime>
+#include <fmt/chrono.h>
```
mhoemmen: Are you including this for `std::chrono`?  That should live in `<chrono>`.

gsjaardema: No, this is from lib::fmt which is the prototype for the new pythonesque output library which will be included in C++20.

packages/seacas/libraries/aprepro_lib/apr_builtin.cc
gsjaardema: That's true... I will look into what can be replaced with C++11 math/numeric functions.

packages/seacas/applications/blot/cnvmax.f
gsjaardema: Yes, older than some of the people running it...

packages/seacas/applications/conjoin/CJ_ExodusFile.h
gsjaardema: Yes, I should add those everywhere...

gsjaardema:
There is a non-backward compatible change in this set -- Removed a defaulted parameter from the Ioss::Property constructor.  Stk_IO was using this parameter (at its defaulted value) so I removed those uses.

gsjaardema:
@alanw0 You might want to be aware of commit ad74712 that I included with this PR.  My PR removed the default parameter in the Ioss::Property constructor that defaulted to false.  The IossBridge file was using this parameter a few places (always with false) so I removed those. The parameter can never be `true`, so there was no reason for the parameter to exist.

alanw0:
Thanks Greg, I'll go ahead and remove those 'false' usages from sierra's copy of IossBridge.C

gsjaardema:
@mhoemmen Can you review one last time -- all tests and builds are now passing.  Thanks.

gsjaardema:
@mhoemmen Thanks.  For some reason it wasn't showing up till I sent the message...  Sorry for the unneeded message

Issue #5932, opened 2019-09-18
MueLu:  Can default GO macros depend on default types in Tpetra?


jhux2:
@trilinos/muelu @csiefer2 

As reported by @tjfulle:

"The situation:

```
cmake -DTrilinos_ENABLE_Tpetra=ON  -DTpetra_ENABLE_DEPRECATED_CODE=OFF [PR options to enable all downstream, tests etc.] -DMueLu_ENABLE_Epetra=OFF ${TRILINOS_DIR}
```

I would expect (hope?) that MueLu would inherit Tpetra’s default global ordinal, but with this config `muelu/src/Headers/MueLu_Details_DefaultTypes.hpp` set MueLu’s default global ordinal to int.  I *think* this is causing link errors in many examples/tests (though I can’t be sure because the linker spews hundreds of pages of template vomit).

Can the macros `HAVE_MUELU_DEFAULT_GO_*` be related to the default types in Tpetra, rather than preferring int?"

tjfulle:
Thanks @jhux2!  I forgot to add that this is with `-DTpetra_ENABLE_DEPRECATED_CODE=OFF`

jhux2:
@tjfulle Can you tell us what is reported during cmake configure in the MueLu section that starts with `MueLu ETI support enabled`?

tjfulle:
@jhux2, ya.  I'm in the middle of a build right now, give me a bit until it fails again :)

brian-kelley:
@tjfulle @jhux2 Yes, this is an issue with my changes from #5697. I'm working on a fix.

jhux2:
@trilinos/MUELU

brian-kelley:
The fix in #5933 seems to have worked, so I'm closing.

PR #5697, opened 2019-08-12
Xpetra/MueLu: remove hardcoded default GO=int (#5656)

brian-kelley:
Removed Xpetra Sc, LO, GO default template parameters entirely.
Left Node alone for now, but it will be removed soon
along with Tpetra deprecated code (see #5602)

Improved MueLu default types: it now chooses a default GO in
CMake that is instantiated if ETI is on
(preferring int over long long, and long long over long).

This fixes errors downstream when Epetra is off
and Tpetra is only instantiated on only INT_LONG_LONG and/or INT_LONG.
Default GO related compile and link errors were also fixed in
Domi, ROL, ShyLU_DDFROSch, and TrilinosCouplings.

@trilinos/muelu 
@trilinos/xpetra 
@trilinos/shylu 

## Description

Xpetra: Removed default Scalar, LO, GO, Node types from all types. @trilinos/xpetra: would it be
preferable to still have default types, with GO chosen based on Xpetra ETI configuration? It appears that
ShyLU_DDFROSch was the only package that directly used these defaults (in tests).

MueLu: Previously, the default template arguments for all classes were <double, int, int, KokkosClassic::DefaultNode::DefaultNodeType>. Now, the default GO is chosen
via CMake logic, like in Tpetra. If ETI is OFF, then int is still the default. Otherwise, GO is chosen based on what
is instantiated in MueLu ETI. If instantiated for multiple GOs, int > long long > long is the order of greatest to least preference. All four default types are now defined in muelu/src/Headers/MueLu_Details_DefaultTypes.hpp:
```
    namespace MueLu
    {
    typedef double DefaultScalar;
    typedef int DefaultLocalOrdinal;
    typedef <depends on CMake defs> DefaultGlobalOrdinal;
    typedef KokkosClassic::DefaultNode::DefaultNodeType DefaultLocalNode;
    }
```

This provides a single place to define the defaults for all of MueLu. This should also make it easier to remove the LO, GO and Node template parameters later. MueLu_Details_DefaultTypes.hpp is now
included in MueLu_ConfigDefs.hpp, so that "DefaultScalar" etc. can be used in each class decl as the default template arguments.

## Motivation and Context

Xpetra and MueLu default template parameters were incorrect if
Tpetra_INST_INT_INT=OFF and ETI for any package was on. Using
the default parameters anywhere would lead to undefined symbol errors.
In practice this only manifested itself in a few downstream tests and examples
in Domi, ROL, ShyLU, and TrilinosCouplings. Now, there are no build errors
[...]

PR #5933, opened 2019-09-18
MueLu: Fix CMake/preprocessor logic for default GO (#5932)

### LINKED ISSUE #5923: MueLu: update makepr.sh to use PR template
It would nice if `makepr.sh` would pull in the [PR template](https://github.com/trilinos/Trilinos/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

brian-kelley:
-Fixes what was supposed to be included in #5697
-setting global CMake variables about which
  integer type should be the default GO in MueLu
-this enables/disables #defines in MueLu_config.h
-#defines from MueLu_config.h are used to set the
  default GO typedef in MueLu_Details_DefaultTypes.hpp

@trilinos/muelu 

## Description

The changes that fixed the default GO selection were:

- fix misspelling of the MueLu ETI option in cmake: changed ``${${PACKAGE_NAME}_INST_DOUBLE_INT_LONG_LONG}`` to ``${${PACKAGE_NAME}_INST_DOUBLE_INT_LONGLONGINT}``
- include MueLu_config.hpp into MueLu_Details_DefaultTypes.hpp; without this, the preprocessor definitions were never seen by it
- use nested cmake "IF(...)" to set the boolean variables instead of trying to use boolean expressions as the right-hand side of GLOBAL_SET(). Added benefit of the IF()s making the intention clearer.
- add ``#cmakedefine HAVE_MUELU_DEFAULT_GO_LONGLONG`` and ``#cmakedefine HAVE_MUELU_DEFAULT_GO_LONG`` (thanks @csiefer2)

## Motivation and Context

Fixes linker errors caused by MueLu's default GO type disagreeing with what's enabled in MueLu ETI [see issue](https://github.com/trilinos/Trilinos/issues/5932)

## How Has This Been Tested?

Went through MueLu configuration with the following settings, and then checked that MueLu_config.hpp had #defined the right things:
``HAVE_MUELU_DEFAULT_GO_LONGLONG`` (use GO=long long), ``HAVE_MUELU_DEFAULT_GO_LONG`` (use GO=long), or neither (GO=int).

- all default: Tpetra_INST_INT_INT and Tpetra_INST_INT_LONG_LONG (default GO=int)
- Tpetra_INST_INT_INT (default GO=int)
- Tpetra_INST_INT_LONG (default GO=long)
- what will be default after Tpetra deprecated code removal: Tpetra_INST_INT_LONG_LONG (default GO=long long)
- Tpetra_INST_INT_LONG and Tpetra_INST_INT_LONG_LONG (default GO=long long)
- ETI off (default GO=int)

## Checklist

### DIFF
--- a/packages/muelu/CMakeLists.txt
+++ b/packages/muelu/CMakeLists.txt
@@ -265,8 +265,10 @@ GLOBAL_SET(HAVE_${PACKAGE_NAME_UC}_TEKO ${${PACKAGE_NAME}_ENABLE_Teko})
 # - MueLu_ENABLE_EXPLICIT_INSTANTIATION
 # - Enabling/disabling of scalar/ordinal types is controlled by Tpetra
 
-IF(${PACKAGE_NAME}_ENABLE_EXPLICIT_INSTANTIATION)
+GLOBAL_SET(HAVE_${PACKAGE_NAME_UC}_DEFAULT_GO_LONG OFF)
+GLOBAL_SET(HAVE_${PACKAGE_NAME_UC}_DEFAULT_GO_LONGLONG OFF)
 
+IF(${PACKAGE_NAME}_ENABLE_EXPLICIT_INSTANTIATION)
   # <double, int, int>
   # Examples and tests need at least this one, also if Epetra is enabled
   IF(Tpetra_INST_DOUBLE AND Tpetra_INST_INT_INT)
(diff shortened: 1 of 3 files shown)


packages/muelu/CMakeLists.txt
```diff
@@ -265,8 +265,10 @@ GLOBAL_SET(HAVE_${PACKAGE_NAME_UC}_TEKO ${${PACKAGE_NAME}_ENABLE_Teko})
 # - MueLu_ENABLE_EXPLICIT_INSTANTIATION
 # - Enabling/disabling of scalar/ordinal types is controlled by Tpetra
 
-IF(${PACKAGE_NAME}_ENABLE_EXPLICIT_INSTANTIATION)
+GLOBAL_SET(HAVE_${PACKAGE_NAME_UC}_DEFAULT_GO_LONG OFF)
+GLOBAL_SET(HAVE_${PACKAGE_NAME_UC}_DEFAULT_GO_LONGLONG OFF)
 
+IF(${PACKAGE_NAME}_ENABLE_EXPLICIT_INSTANTIATION)
```
mhoemmen: The set of enabled types should not depend on whether explicit instantiation is enabled.  For example, Belos uses the set of enabled types for SolverFactory setup, and many packages use the set of enabled types to decide what tests to enable.

brian-kelley: @mhoemmen This doesn't affect what's enabled or instantiated, just what is used to pick one type as the default template parameter. The user of MueLu_blah.hpp can still put anything they want as the GO template parameter.

brian-kelley: @mhoemmen This was all supposed to go into #5697 which removed the default SC,LO,GO from Xpetra entirely, and tried to change MueLu so that the default GO template parameter was chosen from the set of enabled GOs, since before both Xpetra and MueLu used int no matter what.

Issue #5614, opened 2019-07-31
MueLu: Failing tests with Tpetra deprecated code disabled


tjfulle:
## Bug Report
@trilinos/muelu 

Part of: #5602 
### Description

With Tpetra's deprecated code disabled, many MueLu tests fail for `GlobalOrdinal=[int,long long int]`

#### If `GlobalOrdinal=int`

See the full configuration at https://github.com/trilinos/Trilinos/issues/5602#issuecomment-516593697

-  [x] MueLu_BlockedTransfer_Tpetra_MPI_4: see issue #5629 
-  [x] MueLu_DriverEpetra_CircNspDependency_MPI_4: fix in PR #5631
-  [x] MueLu_DriverEpetra_IsorropiaPoisson_MPI_4: fix in PR #5631
-  [x] MueLu_ParameterListInterpreterTpetra_MPI_1
-  [x] MueLu_UnitTestsEpetra_MPI_1: see issue #5629 
-  [x] MueLu_UnitTestsIntrepid2Tpetra_MPI_1: assigned to @csiefer2 in #5625 
-  [x] MueLu_UnitTestsIntrepid2Tpetra_MPI_4: assigned to @csiefer2 in #5625 
-  [x] MueLu_UnitTestsTpetra_MPI_1: partial fix in #5620 and issue #5629
-  [x] MueLu_UnitTestsTpetra_MPI_4: partial fix in #5620 and issue #5629
-  [x] MueLu_VarDofDriver_MPI_1: fix in #5628
-  [x] MueLu_VarDofDriver_MPI_2: fix in #5628

#### If `GlobalOrdinal=long long int`

See the full configuration at: https://github.com/trilinos/Trilinos/issues/5602#issuecomment-516594843

 - [x]  MueLu_UnitTestsTpetra_MPI_1 (Failed): partial fix in #5620 and issue #5629
 - [x] MueLu_UnitTestsTpetra_MPI_4 (Failed): partial fix in #5620 and issue #5629 
 - [x] MueLu_UnitTestsIntrepid2Tpetra_MPI_1 (Failed): assigned to @csiefer2 in #5625 
 - [x] MueLu_UnitTestsIntrepid2Tpetra_MPI_4 (Failed): assigned to @csiefer2 in #5625 
 - [x] MueLu_BlockedTransfer_Tpetra_MPI_4 (Failed): see issue #5629 
 - [x] MueLu_VarDofDriver_MPI_1 (Failed): fix in #5628
[...]

PR #5935, opened 2019-09-18
Tpetra, MueLu: deprecated code fixes

brian-kelley:
@trilinos/muelu 
@trilinos/tpetra 

## Description

    MueLu: fix PgP, emin test failures when Tpetra deprecated is off.
    
    Tpetra: fix MatrixMatrix::Multiply timer scope issue.
    
    Also in MueLu, replaced one instance of deprecated Kokkos fence:
    exec_space::fence() ==> exec_space().fence().
    
    TpetraExt::MatrixMatrix::Multiply has two main code paths:
    a fast one that does fillComplete() and one that doesn't. The
    fast code path (mult_A_B_newmatrix()) for computing C=A*B also
    allocates C to exactly the right size, avoiding the guesswork of
    constructing C with enough entries per row.
    
    However, Xpetra's wrapper for Multiply doesn't request
    fillComplete unless optimizeStorage is also requested. So this commit
    passes "optimizeStorage = true" to Xpetra::Multiply in the
    PG and EMin transfers. This now uses the faster code path in Tpetra
    multiply, and if Tpetra deprecated is off, avoids MueLu test failures
    related to not enough entries being allocated per row.

## Motivation and Context

This change fixes these MueLu test failures (the last remaining failures when Tpetra/Kokkos deprecated off, and GO=long long - see #5614 ):
```
	198 - MueLu_UnitTestsTpetra_MPI_1 (Failed)
	199 - MueLu_UnitTestsTpetra_MPI_4 (Failed)
	226 - MueLu_ParameterListInterpreterTpetraHeavy_MPI_1 (Failed)
	227 - MueLu_ParameterListInterpreterTpetraHeavy_MPI_4 (Failed)
	228 - MueLu_BlockedTransfer_Tpetra_MPI_4 (Failed)
```

## How Has This Been Tested?

Built and ran tests for Tpetra and downstream with Serial, Cuda and OpenMP; with and without Tpetra/Kokkos deprecated code enabled. There are still some failures in Cuda with deprecated off, but I checked that these were not caused by my changes. To the best of my knowledge no new build errors or test failures were introduced.

## Checklist

### DIFF
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
@@ -108,6 +108,7 @@ void Multiply(
 {
   using Teuchos::null;
   using Teuchos::RCP;
+  using Teuchos::rcp;
   typedef Scalar                            SC;
   typedef LocalOrdinal                      LO;
   typedef GlobalOrdinal                     GO;
@@ -121,7 +122,9 @@ void Multiply(
 #ifdef HAVE_TPETRA_MMM_TIMINGS
   std::string prefix_mmm = std::string("TpetraExt ") + label + std::string(": ");
   using Teuchos::TimeMonitor;
-  TimeMonitor MM(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All Setup")));
+  //MM is used to time setup, and then multiply.
+  
+  RCP<TimeMonitor> MM = rcp(new TimeMonitor(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All Setup"))));
 #endif
 
   const std::string prefix = "TpetraExt::MatrixMatrix::Multiply(): ";
@@ -222,7 +225,7 @@ void Multiply(
 
 #ifdef HAVE_TPETRA_MMM_TIMINGS
   {
-  TimeMonitor MM2(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All I&X")));
+  TimeMonitor MM_importExtract(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All I&X")));
 #endif
 
   // Now import any needed remote rows and populate the Aview struct
@@ -243,9 +246,9 @@ void Multiply(
     MMdetails::import_and_extract_views(*Bprime, targetMap_B, Bview, Aprime->getGraph()->getImporter(), false, label, params);
 
 #ifdef HAVE_TPETRA_MMM_TIMINGS
-  }
-  TimeMonitor MM3(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All Multiply")));
-  {
+  } //stop MM_importExtract here
+  //stop the setup timer, and start the multiply timer
+  MM = rcp(new TimeMonitor(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All Multiply"))));
 #endif
 
   // Call the appropriate method to perform the actual multiplication.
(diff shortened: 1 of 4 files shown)


csiefer2:
This is a semantic change --- resumeFill() does *not* allow changes to a graph.  The Matrix w/o fillComplete() does.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -121,7 +122,9 @@ void Multiply(
 #ifdef HAVE_TPETRA_MMM_TIMINGS
   std::string prefix_mmm = std::string("TpetraExt ") + label + std::string(": ");
   using Teuchos::TimeMonitor;
-  TimeMonitor MM(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All Setup")));
+  //MM is used to time setup, and then multiply.
+  
+  RCP<TimeMonitor> MM = rcp(new TimeMonitor(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All Setup"))));
```
mhoemmen: Using `RCP` with a scope guard like TimeMonitor always makes me sad, especially since you use it correctly (without a scope guard) elsewhere in this commit.  (If the scope guard isn't appropriate for the situation, you could always call start and stop on the timer manually.)

brian-kelley: @mhoemmen If I want to use start() and stop(), do I use the Teuchos::Time directly, instead of TimeMonitor?

brian-kelley:
@csiefer2 That's what I was afraid of, but I thought about it and I think it's OK because I see only two possible use cases here:

- the user knows how much storage their matrix will need after the multiply and whatever else they will add to it later, so they preallocate that, and mult_A_B_reuse() will write into that storage, and then resumeFill() lets them insert more entries right away
- the user doesn't know how much storage A*B needs, they just want the product. So they pass in an empty matrix, and mult_A_B_newmatrix will allocate what is needed. The user can't insert new entries beyond that sparsity pattern, but they could add into the values that are there.

brian-kelley:
@csiefer2 Do you think I should make "doFillComplete" an option in the kernels?

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
mhoemmen: @brian-kelley wrote:
> If I want to use start() and stop(), do I use the Teuchos::Time directly, instead of TimeMonitor?

Oh wait -- I realize now that TimeMonitor has the stacked timer option, so you should use the TimeMonitor object.  Sorry :(

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -121,7 +122,9 @@ void Multiply(
 #ifdef HAVE_TPETRA_MMM_TIMINGS
   std::string prefix_mmm = std::string("TpetraExt ") + label + std::string(": ");
   using Teuchos::TimeMonitor;
-  TimeMonitor MM(*TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM All Setup")));
```
brian-kelley: I reverted all the semantic changes to Multiply(), but the timer-related changes should stay. "MMM All Setup" was timing the entire function body, but it should really just be timing everything up to the multiply.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -265,21 +268,19 @@ void Multiply(
     CrsWrapper_CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> crsmat(C);
 
     MMdetails::mult_A_B(Aview, Bview, crsmat, label,params);
+  }
 
 #ifdef HAVE_TPETRA_MMM_TIMINGS
-  }
```
brian-kelley: Also, having this closing brace inside a timing #ifdef didn't seem right, since when timing was enabled it matched with the "else {" on line 261, when I think it was supposed to match with line 248. This actually changes the semantics of the function depending on if timing is enabled.

brian-kelley:
@csiefer2 @mhoemmen Thanks for looking at this!

PR #5949, opened 2019-09-20
Ifpack2 vector length adjustment for ARM

kyungjoo-kim:
## Description
For ARM vectorizations, we may want to use 256bit vector length (4 doubles) even if the hardware supports 128bit (2 doubles). Using 128bit length, it does not seem to hide latency well. This is a very rough estimation with experiments. We do this so that we can set up a baseline performance of the Ifpack2 block tridiagonal solver (SPARC). I will submit a separate PR for KK.

FYI: As we do not have have a PR test with ARM, the PR test would (should) not complain in this PR.


nmhamster:
@kyungjoo-kim - there are 2 x 128b VPUs per core, so I think having 256b makes good sense.

PR #5951, opened 2019-09-20
Framework: Cleaning up the PR Driver Python Script

william76:
@trilinos/framework 

A little bit of cleanup to the Python PR Test script.

Mostly this is just cosmetic / aesthetic cleanup to make the indentations a bit cleaner using `textwrap.dedent()`.

I moved `config_map` which is a globally defined dictionary that was defined kind of in the middle of the script sandwiched in-between a couple of functions to be just before the `run()` function, which is the only place that dict is used.  The original location seemed a little random and could be tricky to notice that it was scoped globally within the script.

@prwolfe @jwillenbring Since I'll be out of town, feel free to review this and merge it in or turn on automerge if you're ok with this modification.

### DIFF
--- a/cmake/std/PullRequestLinuxDriverTest.py
+++ b/cmake/std/PullRequestLinuxDriverTest.py
@@ -30,6 +30,7 @@ import argparse
 import os
 import re
 import sys
+import textwrap
 
 import subprocess
 from multiprocessing import cpu_count
@@ -51,6 +52,9 @@ except ImportError:
 
 
 def parse_args():
+    """
+    Parse the command line arguments.
+    """
     parser = argparse.ArgumentParser(
         description='Parse the repo and build information')
     parser.add_argument('sourceRepo',
@@ -78,6 +82,9 @@ def parse_args():
 
 
 def print_input_variables(arguments):
+    """
+    Print out the input parameters to the console.
+    """
     print(
         "\n==========================================================================================",
         file=sys.stdout)
@@ -116,6 +123,10 @@ def print_input_variables(arguments):
 
 
 def confirmGitVersion():
+    """
+    Verify that the git version is at least 2.10.
+    If it is not then exit with an error.
+    """
     git_version_string = subprocess.check_output(['git', '--version'])
     git_version_number_string = git_version_string[git_version_string.rfind(' '):]
     major_git_version = int(git_version_number_string[:git_version_number_string.find('.')])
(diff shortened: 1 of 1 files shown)


william76:
Tests are failing due to a git version prolem:
```
======================================================================
FAIL: test_verifyGit_fails_with_old_version (__main__.Test_run)
Check to see that git is in path
----------------------------------------------------------------------
  File "/scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_python_2/Trilinos/commonTools/framework/../../cmake/std/unittests/TestPullRequestLinuxDriverTest.py", line 78, in test_verifyGit_fails_with_old_version
    PullRequestLinuxDriverTest.confirmGitVersion()
AssertionError: "Git version  should be 2.10 or better - Exiting!" does not match "Git version should be 2.10 or better - Exiting!"
```

@trilinos/framework

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -124,12 +135,14 @@ def confirmGitVersion():
 
     if major_git_version  <  2 or (major_git_version == 2 and
                                    minor_git_version < 10):
-        raise SystemExit("Git version  should be 2.10 or better - Exiting!")
```
prwolfe: moving this space is what caused the tests to fail. I'm sure that extra space is in the test as well ;->

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -467,19 +496,22 @@ def run():
...
     if "master" == arguments.targetBranch:
         if re.match(re_src_branchname, arguments.sourceBranch) is None:
-            sys.exit("""------------------------------------------------------------------------------------------
-NOTICE: Destination branch is trilinos/Trilnos::master
-ERROR : Source branch is NOT trilinos/Trilinos::master_merge_YYYYMMDD_HHMMSS
-      : This violates Trilinos policy, pull requests into the master branch are restricted.
-      : Perhaps you forgot to specify the develop branch as the target in your PR?
-------------------------------------------------------------------------------------------
-""")
+            sys.exit(textwrap.dedent("""\
```
prwolfe: I have never understood decent for this type of use. Ehh.

cmake/std/PullRequestLinuxDriverTest.py
william76: huh? how does an extra whitespace in the text string of the exit cause a failure?

cmake/std/PullRequestLinuxDriverTest.py
william76: I don't follow your comment...  the `\` after the `"""` means no newline and `dedent(<string>)` will de-indent the string by the line with the least amount of leading whitespace... this makes long multiline strings easier to read in code since they don't break the visual indendentation within the code. Mostly it's just for readability.

cmake/std/PullRequestLinuxDriverTest.py
prwolfe: It's not a smart comparison so the strings must match exactly.  Not elegant is it?

cmake/std/PullRequestLinuxDriverTest.py
prwolfe: sorry - not decent, dedent. This is a non-issue, just stylistic differences. And me fighting with autocorrect.

PR #5956, opened 2019-09-22
MueLu: create kokkos amalgamation factories

lucbv:
@trilinos/muelu 

## Description
The goal of this PR is to introduces two new factories: AmalgamationFactory_kokkos and AmalgamationInfo_kokkos. These are so far identical to their non-kokkos counter part.
However having these new factories creates some changes in the default behavior of MueLu and generated a few errors since xml files and tests did not specify the amalgamation factory to be used...

## Motivation and Context
This is another step forward in the refactor of SA-AMG which will enable to run the setup phase of MueLu to run on device or with OpenMP.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of #5838 
* Composed of 

## How Has This Been Tested?
The changes are covered in the unit_test_kokkos that already test this factory in Aggregates and aggregation. An additional test might be created when the actual underlying algorithm is changed to use Kokkos::View and on node parallelism. 

## Checklist

- [x] My commit messages mention the appropriate GitHub issue numbers.
- [x] My code follows the code style of the affected package(s).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] No new compiler warnings were introduced.
- [ ] These changes break backwards compatibility.

PR #5964, opened 2019-09-23
MueLu: fixing bug with kokkos refactor in FactoryManager, see issue #5961

### LINKED ISSUE #5961: MueLu: useKokkos_ is not set in MueLu_FactoryManager for easy xml
## Bug Report
@trilinos/muelu 

### Description
When using an xml with easy paramter list format the `useKokkos_` flag is not set correctly in the FactoryManager which means that if a factory is needed but not specified by the easy parameter list interpreter it may be set incorrectly by the FactoryManager.

lucbv:
@trilinos/muelu 

## Description
The `useKokkos_` attribute of FactoryManager is not set appropriately set when using an easy xml.
For instance it is set to true in `muelu/example/basic/Simple.cpp` when `simple.xml` clearly sets:
```
<Parameter        name="use kokkos refactor"                  type="bool"     value="false"/>
```

## Motivation and Context
This is a bug and as such should be fixed but it also blocks PR #5956 

## Related Issues

* Closes #5961
* Blocks #5956 and #5838
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## How Has This Been Tested?
A new unit-test is added to MueLu to make sure this parameter is consistently set between the parameter list interpreter and the factory manager.

## Checklist

### DIFF
--- a/packages/muelu/src/Interface/MueLu_HierarchyManager.hpp
+++ b/packages/muelu/src/Interface/MueLu_HierarchyManager.hpp
@@ -353,7 +353,7 @@ namespace MueLu {
             if (!M.is_null()) {
               Xpetra::IO<Scalar, LocalOrdinal, GlobalOrdinal, Node>::Write(fileName,* M);
             }
-          }	  
+          }
 	  else if (L->IsAvailable(name)) {
 	    // Try nofactory
             RCP<T> M = L->template Get< RCP<T> >(name);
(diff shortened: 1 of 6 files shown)


lucbv:
I am working on the unit-test now, this has been uncovered in #5956 but I think that it should not be mixed with it as it is clearly a separate issue.

lucbv:
@jhux2 @cgcgcg 
I have added a test in the first commit that fails but is fixed by the second commit.

PR #5968, opened 2019-09-23
fix find region bug

rstumin:
Auto-PR for SHA 9be5dba

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
@@ -313,7 +313,6 @@ void MakeCoarseLevelMaps2(const int maxRegPerGID,
 
   using MT = typename Teuchos::ScalarTraits<SC>::magnitudeType;
 
-  RCP<const Teuchos::Comm<int> > comm = regProlong[1][0]->getRowMap()->getComm();
   const GO GO_INV = Teuchos::OrdinalTraits<GO>::invalid();
   const int numLevels = regProlong.size();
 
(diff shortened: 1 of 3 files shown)


packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
```diff
@@ -1048,87 +1047,95 @@ void vCycle(const int l, ///< ID of current level
...
-      for (int j = 0; j < maxRegPerProc; j++) {
-        RCP<Vector> inverseInterfaceScaling = VectorFactory::Build(regInterfaceScalings[l][j]->getMap());
-        inverseInterfaceScaling->reciprocal(*regInterfaceScalings[l][j]);
-        fineRegB[j]->elementWiseMultiply(SC_ONE, *fineRegB[j], *inverseInterfaceScaling, SC_ZERO);
-      }
-
-      regionalToComposite(fineRegB, compRhs, maxRegPerProc, quasiRegRowMaps[l],
-                          regRowImporters[l], Xpetra::ADD);
+    const bool useCoarseSmoother = false;
+    if (useCoarseSmoother) {
```
lucbv: It looks like this code branch is never used since `useCoarseSmoother` is set to `false`, would it make more sense to have the value come from the command line with a default value `false` instead?

mayrmt: @lucbv I think you’re right. Still, this is the coarse level and we usually just Wanne use a direct solver, no smoother. This is still a weird logic here, but it doesn’t matter in practice. Let’s talk about it in our next meeting.

PR #6010, opened 2019-10-01
SuperLU 5.4.0 support

kyungjoo-kim:
## Description
SPARC wants to upgrade their TPLs. So far, super lu dist 4.2 is supported and tested in atdm nightly build tests. However, superlu dist github repository provides release version from 5 and there is no place to download version 4.2. Supporting version 4.2 seems to be a bit outdated. This PR includes temporary modifications that requires to support version 5 and greater. If we decide to discard superlu support of version 4.2, we can put a more permanent solution in the code. 

## Motivation and Context
SPARC reports memory issues when interfacing super lu dist. The problem is because super lu uses its own malloc and free with additional meta information but amesos2 interface uses standard free function to deallocate superlu buffers. This causes memory issues.

## How Has This Been Tested?
Jeff verified that this resolved SPARC issues. To test this PR, you need to add ``-DCMAKE_CXX_FLAGS="-DAMESOS2_ENABLE_SUPERLUDIST_5_4_0"`` in the configure script.


packages/amesos2/src/Amesos2_Superludist_FunctionMap.hpp
```diff
@@ -372,10 +372,14 @@ namespace Amesos2 {
     static void LUstructInit(SLUD::int_t m, SLUD::int_t n,
 			     type_map::LUstruct_t* lu)
     {
+#if defined(AMESOS2_ENABLE_SUPERLUDIST_5_4_0)
```
srajama1: This test may not be needed. The CMake 2 arg test will pass and the line number 378 will evaluate to true, right ?

packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
```diff
@@ -70,6 +70,11 @@ namespace SLUD {
 
 extern "C" {
 
+  /// use the same function with name space in the macro
+#if defined(AMESOS2_ENABLE_SUPERLUDIST_5_4_0)
+#define USER_FREE(addr) SLUD::superlu_free_dist(addr)
```
ndellingwood: @kyungjoo-kim should this be `SUPERLU_FREE` as below? I don't see `USER_FREE` appearing anywhere else in the PR or in VOTD Amesos2.

packages/amesos2/src/Amesos2_Superludist_FunctionMap.hpp
kyungjoo-kim: Unfortunately, it sometimes does not work. As we decide to support superlu dist version 5 and greater, the 2arg test won't be necessary.

ndellingwood: Is this set by users at configure time? Is it worth detecting the version at compile time via a test in `cmake/TPLs/FindTPLSuperLUDist.cmake` that would set an appropriate variable in `amesos2/cmake/Amesos2_config.h.in`? This isn't a request, just checking.

packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
kyungjoo-kim: It is a macro used in  SUPERLU_FREE which is defined in superlu dist. 
```
/// this gives users can over-ride memory functions.
#if !defined(USER_FREE)
#define USER_FREE(addr) superlu_free_dist(addr)
#endif
#define SUPERLU_FREE(addr) USER_FREE(addr) 
```
As Amesos2 has namespace, I define the same default function with the name space.

packages/amesos2/src/Amesos2_Superludist_FunctionMap.hpp
kyungjoo-kim: This was a temporary flag and the flag should come with cmake_cxx_flags. I did in this way because we need to agree between application developers and test team about what version of superlu dist should be supported and tested. I am going to remove the flag as it won't be necessary anymore.

packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
ndellingwood: Thanks for clarification!

Issue #6031, opened 2019-10-03
Implement proper move constructor for Teuchos::RCP


bartlettroscoe:
It would seem that C++11 compilers automatically create incorrect move constructors for the Teuchos::RCP class.  This is causing a major problem with EMPIRE (see https://cee-gitlab.sandia.gov/EM-Plasma/EMPIRE-PIC/issues/465).

The initial goal will be to fix the immediate problem for EMPIRE.

After Teuchos::RCP is fixed for EMPIRE, this story will also cover adding correct move constructors and assugment operators for all of the Teuchos Memory Management types (to avoid other cases like this).

bartlettroscoe:
FYI: PR #6034 has a clean and complete implementation of move functions for Teuchos::RCP.  I will let the branch and PR as-is and let it merge so EMPIRE can get that update.   I will create a new branch and new PR to fill in move operations for the rest of the Teuchos MM classes.

bartlettroscoe:
FYI: It was determined that this was not a bug in Teuchos::RCP or the compiler.  It was a misunderstanding about 

In any case, PR #6034 was just merged to 'develop' that implements a proper move functions for Teuchos::RCP.

I will now expand the scope of this issue to add move operations for all of the Teuchos MM classes.

bartlettroscoe:
This still needs to get done at some point.

PR #6034, opened 2019-10-03
Add move functions to Teuchos::RCP (#6031)

bartlettroscoe:
This is a WIP PR to develop C++11 move functions for Teuchos Memory Management classes (see #6031).

This branch can be merged at any time if needed, even if all of the move operations have not been added yet since they are finished for Teuchos::RCP.

### DIFF
--- a/packages/teuchos/core/src/Teuchos_RCP.hpp
+++ b/packages/teuchos/core/src/Teuchos_RCP.hpp
@@ -282,6 +282,15 @@ RCP<T>::RCP(const RCP<T>& r_ptr)
 
 
 template<class T>
+inline
+RCP<T>::RCP(RCP<T>&& r_ptr)
+  : ptr_(r_ptr.ptr_), node_(std::move(r_ptr.node_))
+{
+  r_ptr.ptr_ = 0;
+}
+
+
+template<class T>
 template<class T2>
 inline
 RCP<T>::RCP(const RCP<T2>& r_ptr)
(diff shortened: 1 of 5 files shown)


bartlettroscoe:
The Teuchos::RCP move operations on this branch should be good to go and have strong unit tests.  Therefore, I will let this PR merge to 'develop' and I will create a new branch to add move operations for the rest of the Teuchos Memory Management classes.

packages/teuchos/core/src/Teuchos_RCP.hpp
```diff
@@ -281,6 +281,15 @@ RCP<T>::RCP(const RCP<T>& r_ptr)
 {}
 
 
+template<class T>
+inline
+RCP<T>::RCP(RCP<T>&& r_ptr)
```
mhoemmen: @bartlettroscoe It looks like the reason you can't just write `RCP(RCP<T>&&) == default;` is because you want to set `r_ptr.ptr_` to `nullptr` after moving.  Is that right?

packages/teuchos/core/src/Teuchos_RCP.hpp
```diff
@@ -281,6 +281,15 @@ RCP<T>::RCP(const RCP<T>& r_ptr)
...
+  : ptr_(r_ptr.ptr_), node_(std::move(r_ptr.node_))
+{
+  r_ptr.ptr_ = 0;
```
mhoemmen: Correct C++11 idiom is to use `nullptr`, not `0`.

packages/teuchos/core/src/Teuchos_RCP.hpp
bartlettroscoe: Correct.  Think about what would happen if you used the default move constructor implementation.   I would leave the raw pointer set in the input RCP `r_ptr` which could have disastrous consequences  if someone were to try to use it.

packages/teuchos/core/src/Teuchos_RCP.hpp
bartlettroscoe: I think tried that and it did not compile.  Is there a header that needs to be included?

packages/teuchos/core/src/Teuchos_RCP.hpp
mhoemmen: True, though the only requirement for the result of a move is that it be left in an indeterminate but valid state.  I'm guessing the destructor would be unhappy if the pointer were nonnull, though.

packages/teuchos/core/src/Teuchos_RCP.hpp
mhoemmen: `nullptr` is a keyword, but its type is defined in `<cstddef>`.

PR #6038, opened 2019-10-04
Tempus rkstepperobserver protect dynamic cast

krcb:
@trilinos/tempus 

## Motivation

Checks that the passed in observer dynamic casts successfully to RKStepperObserver
*before* adding to the composite. Needed to protect a downstream application.


packages/tempus/src/Tempus_StepperDIRK_impl.hpp
```diff
@@ -97,8 +97,10 @@ void StepperDIRK<Scalar>::setObserver(
   if (( obs == Teuchos::null ) and (this->stepperObserver_->getSize() == 0) )
      obs = Teuchos::rcp(new StepperRKObserver<Scalar>());
 
-  this->stepperObserver_->addObserver(
-       Teuchos::rcp_dynamic_cast<StepperRKObserver<Scalar> > (obs, true) );
+  // Check that this casts to prevent a runtime error if it doesn't
+  if (Teuchos::rcp_dynamic_cast<StepperRKObserver<Scalar> > (obs) != Teuchos::null)
+    this->stepperObserver_->addObserver(
+         Teuchos::rcp_dynamic_cast<StepperRKObserver<Scalar> > (obs, true) );
```
rppawlo: This is dangerous in that the check allows you to silently ignore registered observers. I can see people trying to debug asking "why when I call addObserver does it silently not add it?"

krcb: I can add a warning to that effect; the code right now dies at runtime if you don't do this. Alternatively, we can remove it after the downstream app has done the promotion.

rppawlo: I would probably lean towards leaving this as is to get the app up and running in the short term. @ccober6 any preference? We should set up a meeting to discuss a long term design of observers. They seem to be the source of many complications.

ccober6: Yeah, I agree.  My branch that is on hold simplifies this setObserver(), and the primary reason I wanted to get my branch through.  FYI, the changes that @sconde put into EMPIRE-Hybrid will need to be reverted with my branch.  This is an unfortunate situation where Sidafa changes were not the path we wanted and the next commit changes it to something better, but currently we are in-between those changes.

krcb: Ok, so we're all agreed that we put this in and then improve it later?

ccober6: I would probably lean towards adding print statement noting that an observer was not added because of casting problem.  But otherwise yes let's make the change and improve later.

krcb: Warnings have been added.

sconde:
This doesnt look like it has the changed I pushed for the segfault

krcb:
> This doesnt look like it has the changed I pushed for the segfault

@sconde This is straight from develop.

sconde:
@krcb sorry. I see that now.

PR #6045, opened 2019-10-04
Xpetra: Fix default builds with deprecated OFF

brian-kelley:
Now disabling Epetra support in Xpetra+downstream if
Epetra was not explicitly enabled by user, and if int
is not a common global ordinal type between Epetra and Tpetra.

@trilinos/xpetra 
@trilinos/muelu 

## Motivation

This change is needed because a build without deprecated code, with Xpetra enabled and all other parameters left as the default would fail during configuration. Epetra would use 32-bit global indices, Tpetra would use 64-bit global indices only, and so Xpetra wouldn't be able to find a common type.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows #6019  (but actually fixes it this time)
* Precedes 
* Related to #5602 
* Part of 
* Composed of 

## Stakeholder Feedback

N/A, this is part of Tpetra's deprecated code removal effort
## Testing

N/A, it's a bulid system change only


packages/xpetra/CMakeLists.txt
```diff
@@ -149,12 +120,30 @@ IF (${PACKAGE_NAME}_ENABLE_Epetra)
           MESSAGE(FATAL_ERROR "If Xpetra Epetra support is enabled and Epetra 32-bit global indices are enabled, Xpetra requires that you enable Tpetra_INST_INT_INT")
         ELSE()
           # Epetra was enabled but only implicitly; give a warning but turn off Xpetra's Epetra support
-          MESSAGE(WARNING "32-bit global indices are enabled in Epetra but not Tpetra. Since Epetra was not enabled explicitly by user, disabling Epetra support in Xpetra.")
+          SET(WARN_MSG "32-bit global indices are enabled in Epetra but not Tpetra. Since Epetra was not enabled explicitly by user, **DISABLING EPETRA SUPPORT** in Xpetra and: ")
+          SET(${PACKAGE_NAME}_ENABLE_Epetra OFF PARENT_SCOPE)
           SET(${PACKAGE_NAME}_ENABLE_Epetra OFF)
```
bartlettroscoe: You might consider removing some duplication by using TriBITS utility function [DUAL_SCOPE_SET()](https://tribits.org/doc/TribitsDevelopersGuide.html#dual-scope-set).  Unfortunately this is a too common pattern.

brian-kelley: @bartlettroscoe  Thanks for the tip, just changed it

jhux2:
@brian-kelley @csiefer2 It's good to see the new SQA template in action.

brian-kelley:
@jhux2 Even though I put "N/A" for 2 out of 3 sections? :)

packages/xpetra/CMakeLists.txt
```diff
@@ -149,12 +120,27 @@ IF (${PACKAGE_NAME}_ENABLE_Epetra)
...
-          MESSAGE(WARNING "32-bit global indices are enabled in Epetra but not Tpetra. Since Epetra was not enabled explicitly by user, disabling Epetra support in Xpetra.")
-          SET(${PACKAGE_NAME}_ENABLE_Epetra OFF)
+          SET(WARN_MSG "32-bit global indices are enabled in Epetra but not Tpetra. Since Epetra was not enabled explicitly by user, **DISABLING EPETRA SUPPORT** in Xpetra and: ")
+          DUAL_SCOPE_SET(${PACKAGE_NAME}_ENABLE_Epetra OFF)
+          DUAL_SCOPE_SET(${PACKAGE_NAME}_ENABLE_EpetraExt OFF)
+          DUAL_SCOPE_SET(${PACKAGE_NAME}_Epetra_NO_32BIT_GLOBAL_INDICES ON)
+          GLOBAL_SET(${PACKAGE_NAME_UC}_EPETRA_NO_32BIT_GLOBAL_INDICES ON)
+          IF(${PROJECT_NAME}_ENABLE_Zoltan2)
+            SET(WARN_MSG "${WARN_MSG} Zoltan2")
+            SET(Zoltan2_ENABLE_Epetra OFF PARENT_SCOPE)
```
bartlettroscoe: I just remembered that you might want to print something like "NOTE: Setting Zoltan2_ENABLE_Epetra=OFF because Xpetra_ENBALE_Epetra=OFF" like in the email response I sent you on this yesterday.

Otherwise, users and developers may scratch their heads about how `Zoltan_ENABLE_Epetra` got set to `OFF` when the output clearly showed it was `ON`.

packages/xpetra/CMakeLists.txt
```diff
@@ -149,12 +120,27 @@ IF (${PACKAGE_NAME}_ENABLE_Epetra)
...
+          ENDIF()
+          IF(${PROJECT_NAME}_ENABLE_MueLu)
+            SET(WARN_MSG "${WARN_MSG} MueLu")
+            SET(MueLu_ENABLE_Epetra OFF PARENT_SCOPE)
```
bartlettroscoe: Likewise, you might want to print "NOTE: Setting MueLu_ENABLE_Epetra=OFF because Xpetra_ENBALE_Epetra=OFF" .

brian-kelley:
@bartlettroscoe OK, I have it printing this during Xpetra's configuration:

```
CMake Warning at packages/xpetra/CMakeLists.txt:123 (MESSAGE):
  32-bit global indices are enabled in Epetra but not Tpetra.  Since Epetra
  was not enabled explicitly by user, **DISABLING EPETRA SUPPORT** in Xpetra
  and its downstream.

-- Setting Zoltan2_ENABLE_Epetra=OFF because Xpetra_ENABLE_Epetra=OFF.
-- Setting MueLu_ENABLE_Epetra=OFF because Xpetra_ENABLE_Epetra=OFF.
```

brian-kelley:
OK, I just found out that parts of Zoltan2 actually do use Epetra directly, so I'm going to try to only disable the parts that actually depend on Xpetra.

brian-kelley:
It turns out that the entire library and all but 2 tests in Zoltan2 should actually stay enabled because they just use Epetra directly. Good thing I caught that, it's fixed now. I reverted all the other changes to zoltan2/CMakeLists.txt.

I did test out both ways to silence the main warning in Xpetra, and they work.

packages/xpetra/CMakeLists.txt
```diff
@@ -149,12 +120,23 @@ IF (${PACKAGE_NAME}_ENABLE_Epetra)
           MESSAGE(FATAL_ERROR "If Xpetra Epetra support is enabled and Epetra 32-bit global indices are enabled, Xpetra requires that you enable Tpetra_INST_INT_INT")
         ELSE()
           # Epetra was enabled but only implicitly; give a warning but turn off Xpetra's Epetra support
-          MESSAGE(WARNING "32-bit global indices are enabled in Epetra but not Tpetra. Since Epetra was not enabled explicitly by user, disabling Epetra support in Xpetra.")
-          SET(${PACKAGE_NAME}_ENABLE_Epetra OFF)
+          MESSAGE(WARNING "32-bit global indices are enabled in Epetra but not Tpetra. Since Epetra was not enabled explicitly by user, **DISABLING EPETRA SUPPORT** in Xpetra and its downstream."
```
jhux2: "and its downstream" ==>  "and its downstream dependencies"

packages/xpetra/CMakeLists.txt
```diff
@@ -149,12 +120,23 @@ IF (${PACKAGE_NAME}_ENABLE_Epetra)
...
+            " To make this warning go away, set -DTpetra_INST_INT_INT=ON to enable 32-bit global indices and Epetra support, OR set -DXpetra_ENABLE_Epetra=OFF if you don't need Epetra.")
```
mhoemmen: Thanks for adding the instructions!  I'm OK with this, as long as `Xpetra_ENABLE_Epetra=OFF` automatically sets `MueLu_ENABLE_Epetra=OFF`.

packages/xpetra/CMakeLists.txt
brian-kelley: Thanks, I made this message clearer.

packages/shylu/shylu_dd/frosch/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 TRIBITS_SUBPACKAGE(FROSch)
 
-IF (${PROJECT_NAME}_ENABLE_ShyLU_DDFROSch AND NOT (HAVE_XPETRA_EPETRA AND HAVE_AMESOS2_EPETRA))
+IF (${PROJECT_NAME}_ENABLE_ShyLU_DDFROSch AND NOT (Xpetra_ENABLE_Epetra AND Amesos2_ENABLE_Epetra))
   SET(HAVE_SHYLU_DDFROSCH_EPETRA OFF)
```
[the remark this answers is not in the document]
brian-kelley: @mhoemmen Can you clarify what should be printed? This isn't intended to be a semantic change - I was trying to make it match the convention to use ``Xyz_ENABLE_Abc`` instead of ``HAVE_XYZ_ABC`` in CMake logic.

packages/xpetra/CMakeLists.txt
brian-kelley: @mhoemmen Yes it does.

PR #6047, opened 2019-10-04
Tpetra deprecation: Part n of N

tjfulle:
@trilinos/tpetra 
@trilinos/muelu 
@trilinos/stratimikos 
@trilinos/piro 

## Motivation

Cleans up many uses of hard coded `GlobalOrdinal=int` in preparation of removal of Tpetra deprecated code

## Stakeholder Feedback

None

## Testing

N/A these changes should affect only the build


tjfulle:
@trilinos/shylu

packages/piro/src/Piro_InvertMassMatrixDecorator.hpp
```diff
@@ -55,10 +55,14 @@
...
+#include "Tpetra_Map.hpp"
+using default_lo = Tpetra::Map<>::local_ordinal_type;
+using default_go = Tpetra::Map<>::global_ordinal_type;
+
 namespace Piro {
 
 #ifdef ALBANY_BUILD
-template <typename Scalar, typename LocalOrdinal = int, typename GlobalOrdinal = LocalOrdinal,
+template <typename Scalar, typename LocalOrdinal = default_lo, typename GlobalOrdinal = default_go,
           typename Node = KokkosClassic::DefaultNode::DefaultNodeType>
```
mhoemmen: Did you also want to make the default Node type in Piro a type alias, as with MueLu above?

packages/piro/test/MockModelEval_A_Tpetra.cpp
```diff
@@ -88,7 +88,7 @@ MockModelEval_A_Tpetra::MockModelEval_A_Tpetra(const Teuchos::RCP<const Teuchos:
 
     //set up jacobian graph
     crs_graph = rcp(new Tpetra_CrsGraph(x_map, vecLength));
-    std::vector<int> indices(vecLength);
+    std::vector<typename Tpetra_CrsGraph::global_ordinal_type> indices(vecLength);
     for (int i=0; i<vecLength; i++) indices[i]=i;
```
mhoemmen: Best practice is to make the cast explicit.  We've talked before about using a tagged integer type for GO, but that's future work.

packages/piro/test/MockModelEval_A_Tpetra.cpp
```diff
@@ -88,7 +88,7 @@ MockModelEval_A_Tpetra::MockModelEval_A_Tpetra(const Teuchos::RCP<const Teuchos:
...
     const int nodeNumElements = x_map->getNodeNumElements();
```
mhoemmen: This might cause an unsigned-to-signed cast warning, since (inexplicably) `getNodeNumElements` returns `size_t`.  (Ask Chris Baker why; he might then ask some earlier postdoc why.)

packages/trilinoscouplings/examples/scaling/example_Poisson2D_p2_tpetra.cpp
```diff
@@ -159,8 +159,8 @@ using Teuchos::ParameterList;
 /*********************************************************/
 typedef double scalar_type;
 typedef Teuchos::ScalarTraits<scalar_type> ScalarTraits;
-typedef int local_ordinal_type;
-typedef int global_ordinal_type;
+using local_ordinal_type = Tpetra::Map<>::local_ordinal_type;
+using global_ordinal_type = Tpetra::Map<>::global_ordinal_type;
 typedef KokkosClassic::DefaultNode::DefaultNodeType NO;
```
mhoemmen: See above.  Why not just get this Node typedef from `Tpetra::Map`?

tjfulle: Thanks @mhoemmen !  The response to all your questions is the same - getting the code to build with deprecated code off was a game of wack-a-mole and I only hit the moles that taunted me 😂. After the tester is done I’ll change the node types you mentioned

mhoemmen: @tjfulle cool no worries :D

PR #6074, opened 2019-10-10
Tempus: Add Output Control to TimeStepControl.

ccober6:
* Added ability to control printing of timestep changes.
 * Added set function for "Output Index List" in TimeStepControl.
 * Added ability to set the observer on the Subcycling Integrator.
 * Added a no-op IntegratorObserver that does nothing, including
   printing headers and timestep during integrator->advanceTime().
   Primarily used for Subcycling integrator without printing, which
   is the default.
 * Added testing for setting the output on/off.

All tests pass on my Mac.

@trilinos/tempus

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorObserverNoOp_decl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorObserverNoOp_decl.hpp
@@ -1,0 +1,63 @@
+// @HEADER
+// ****************************************************************************
+//                Tempus: Copyright (2017) Sandia Corporation
+//
+// Distributed under BSD 3-clause license (See accompanying file Copyright.txt)
+// ****************************************************************************
+// @HEADER
+
+#ifndef Tempus_IntegratorObserverNoOp_decl_hpp
+#define Tempus_IntegratorObserverNoOp_decl_hpp
+
+#include "Tempus_IntegratorObserver.hpp"
+#include "Tempus_Integrator.hpp"
+#include "Teuchos_Time.hpp"
+
+namespace Tempus {
+
+/** \brief IntegratorObserverNoOp class for time integrators.
+ *  This basic class has simple no-op functions, as all basic
+ *  functionality should be handled through other methods.
+ */
+template<class Scalar>
+class IntegratorObserverNoOp
+  : virtual public Tempus::IntegratorObserver<Scalar>
+{
+public:
+
+  /// Constructor
+  IntegratorObserverNoOp();
+
+  /// Destructor
+  virtual ~IntegratorObserverNoOp();
+
+  /// \name Basic IntegratorObserver methods
+  //@{
+    /// Observe the beginning of the time integrator.
+    virtual void observeStartIntegrator(const Integrator<Scalar>& integrator) override;
+
+    /// Observe the beginning of the time step loop.
+    virtual void observeStartTimeStep(const Integrator<Scalar>& integrator) override;
+
+    /// Observe after the next time step size is selected.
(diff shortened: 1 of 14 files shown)


ccober6:
@sconde I will be on vacation, and only occasionally check email.  Hopefully there is no problem with this, and no conflict with EMPIRE (as I believe that this branch should be backward compatible).  If so can you start the automerge, and get this on develop.  If not we can cover things on Monday.  Thanks, Curt.

packages/tempus/test/Subcycling/Tempus_SubcyclingTest.cpp
```diff
@@ -241,6 +245,9 @@ TEUCHOS_UNIT_TEST(Subcycling, SinCosAdapt)
     stepper->setSubcyclingMaxFailures      (10);
     stepper->setSubcyclingMaxConsecFailures(5);
     stepper->setSubcyclingScreenOutputIndexInterval(1);
+    //stepper->setSubcyclingIntegratorObserver(
+    //  Teuchos::rcp(new Tempus::IntegratorObserverSubcycling<double>()));
+    //stepper->setSubcyclingPrintDtChanges   (true);
```
sconde: Are these comments needed anymore?

packages/tempus/test/Subcycling/Tempus_SubcyclingTest.cpp
```diff
@@ -410,6 +417,10 @@ TEUCHOS_UNIT_TEST(Subcycling, VanDerPolOperatorSplit)
     stepperSC->setSubcyclingMaxFailures      (10);
     stepperSC->setSubcyclingMaxConsecFailures(5);
     stepperSC->setSubcyclingScreenOutputIndexInterval(1);
+    //stepper->setSubcyclingIntegratorObserver(
+    //  Teuchos::rcp(new Tempus::IntegratorObserverSubcycling<double>()));
+    //stepperSC->setSubcyclingPrintDtChanges   (true);
+
```
sconde: Needed?

packages/tempus/src/Tempus_IntegratorObserverNoOp_decl.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Tempus {
+
+/** \brief IntegratorObserverNoOp class for time integrators.
+ *  This basic class has simple no-op functions, as all basic
+ *  functionality should be handled through other methods.
+ */
+template<class Scalar>
+class IntegratorObserverNoOp
+  : virtual public Tempus::IntegratorObserver<Scalar>
```
sconde: I think I understand what this is doing.

But shouldn't `IntegratorObserverBasic` really be the noop observer? 

And move the output to a `Tempus_IntegratorObseverOutput` or something?

packages/tempus/test/Subcycling/Tempus_SubcyclingTest.cpp
ccober6: I kept these for when we want to output the subcycling timestep information during debugging.

packages/tempus/test/Subcycling/Tempus_SubcyclingTest.cpp
ccober6: I kept these for when we want to output the subcycling timestep information during debugging.

packages/tempus/src/Tempus_IntegratorObserverNoOp_decl.hpp
ccober6: I am fine with that.  The only downside is it is an incompatible backward change, which could require us to make changes to downstream apps.  :(

PR #6076, opened 2019-10-10
KokkosKernels - trsm transpose does not solve the problem correctly.

kyungjoo-kim:
## Motivation
@csiefer2 uses trsm with transpose in muelu. The code does not produce a right solution (a bug). This PR fixed the problem.

## Testing
In this PR, @csiefer2 confirms that the code works and his problem is resolved. Later, I will add more tests in the kokkoskernels repo.


csiefer2:
This breaks a unit test, because the Tpetra code is wrong...

ndellingwood:
This PR never had a matching patch to kokkos-kernels. Issued with kokkos/kokkos-kernels#541

Issue #6005, opened 2019-10-01
MueLu/Belos: Preconditioning Poisson system with rank 1 nullspace


DABH:
## Question

I have a square, symmetric, positive-(semi)-definite matrix A which comes from a Poisson system with all Neumann boundary conditions and therefore has a rank one nullspace, whose basis is just the constant vector.  I'm able to use Belos with an unpreconditioned CG solve to solve my system without any issues.  However, when trying to add preconditioners from MueLu, the ability of CG to converge seems very sensitive to the preconditioner settings -- sometimes I can get a solution with a great speedup and other times I get errors about p^H A p < 0.  It tends to be much more of the latter for larger problem sizes.

I'm aware that due to the singularity of the matrix it's natural to expect issues like this, but I also believe there are several potential ways to work around them.

1) Are you aware of any preconditioner / preconditioner settings which are effective for a case like this with a trivial, know, rank 1 nullspace for an otherwise "nice" system?

2) I've seen some references in the MueLu user's guide / tutorials about nullspace, but I didn't find a complete example of how to set the nullspace for the preconditioner/solver, and I'm not sure where this info about the nullspace needs to be propagated.  For instance:
-- Do all layers of the hierarchy need to have the nullspace set?
-- Does the Belos solver need to be told about the nullspace, or only the preconditioners?
-- If I specify the nullspace from an XML file, how do I specify _what_ the nullspace is?

3) For other libraries like petsc I've found threads like
https://lists.mcs.anl.gov/pipermail/petsc-users/2008-August/003240.html
which describe options they have like `-pc_factor_shift_positive_definite` which help keep preconditioners positive definite (probably at the cost of some efficacy?).  Is there anything similar in MueLu where I can just "force" the levels to all be PD with some diagonal shifting?

4) On paper at least, if x sums to zero, then I can solve (A + oo^T)x = b (where o is an all ones vector) and get the same answer as Ax=b, but in this case (A + oo^T) is full rank.  Is there any way with Trilinos where I could solve this modified system in a way that does the rank-one update efficiently, i.e. computes Ax and o*(o^T x) separately in the CG iteration, etc.?  I'm not sure whether this would ultimately help, and the problem may be in the preconditioners not the system itself, but thought I'd ask your thoughts about it.

Thanks in advance for sharing any expertise you might have!

mayrmt:
@trilinos/muelu

mayrmt:
@DABH Regarding your second question about setting the null space:

[...]

PR #6085, opened 2019-10-11
Plumbing for BelosBlockCGSolMgr so that Assert Positive Definiteness flag gets passed to BelosCGIter

DABH:
@trilinos/belos

## Motivation

This was brought up as part of https://github.com/trilinos/Trilinos/issues/6005 .  I tried to do something like `belosList.set("Assert Positive Definiteness", false);` in my code while setting up a BelosBlockCGSolMgr but the value didn't get passed through to the underlying BelosCGIter.  AFACIT the plumbing was jut missing -- not sure if there was a good reason for that?  If not, this PR adds the necessary plumbing so the flag value is now correctly passed through.

https://github.com/trilinos/Trilinos/issues/6005


packages/belos/src/BelosBlockCGSolMgr.hpp
```diff
@@ -828,6 +838,8 @@ BlockCGSolMgr<ScalarType,MV,OP,true>::getValidParameters() const
       "The string to use as a prefix for the timer labels.");
     pl->set("Orthogonalization", static_cast<const char *>(orthoType_default_),
       "The type of orthogonalization to use: DGKS, ICGS, or IMGS.");
+    pl->set("Assert Positive Definiteness",static_cast<bool>(assertPositiveDefiniteness_default_),
```
mhoemmen: It's a `bool`; you don't need the `static_cast`.

DABH: I think you're right -- any idea why, in that block of code, there are several static casts of things to ints and bools?  Even for the static casts to const char*, those variables that are being casted seem to already be of those types.  Are any of these static casts here actually necessary?

mhoemmen: @DABH The casts are probably not necessary.  Belos is nearly 15 years old -- it's possible that some compilers we had to support back that had trouble deducing some of those types.  It's also likely that later developers went overboard with the casts.  I'll be happy to approve PRs that purge these; I just don't have time at the moment to clean this up.

mhoemmen: btw I'm also not convinced that ParameterList needs overloads for `const char value[]` and `char value[]`; the former should suffice.  [This example](https://godbolt.org/z/hXjTcB) compiles even with C++98.

PR #6087, opened 2019-10-12
MueLu: using kokkos for on node parallelism in Amalgamation process

### LINKED ISSUE #5838: MueLu: kokkos refactor of uncoupled aggregation and amalgamation process

lucbv:
@trilinos/muelu 

## Motivation
Final step in the refactor of the aggregation process of MueLu.
After that profiling will be needed to assess performance and non-trivial serial parts of the code.

## Related Issues

* Closes #5838
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback
See issue #5838 

## Testing
Local testing has been performed as well as testing on `while`.

### DIFF
--- a/packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationFactory_kokkos_decl.hpp
+++ b/packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationFactory_kokkos_decl.hpp
@@ -56,7 +56,7 @@
 #include "MueLu_Exceptions.hpp"
 
 #ifdef HAVE_MUELU_KOKKOS_REFACTOR
-#include "MueLu_AmalgamationInfo_kokkos_fwd.hpp"
+#include "MueLu_AmalgamationInfo_kokkos.hpp"
 
 namespace MueLu {
 
@@ -82,6 +82,9 @@ namespace MueLu {
 #include "MueLu_UseShortNames.hpp"
 
   public:
+
+    using execution_space = typename AmalgamationInfo_kokkos::execution_space;
+    using memory_space    = typename AmalgamationInfo_kokkos::memory_space;
 
     //! @name Constructors/Destructors.
     //@{
@@ -117,7 +120,7 @@ namespace MueLu {
     // @param A (const Matrix&): operator A (matrix) with striding information (if available)
     // @param amalgamatedMap (const Map&): amalgamated node based map
     // @param translation (Array<LO>&): array storing local node ids given local dof ids (needed in CoalesceDropFactory)
-    static void AmalgamateMap(const Map& sourceMap, const Matrix& A, RCP<const Map>& amalgamatedMap, Array<LO>& translation); // const;
+    static void AmalgamateMap(const Map& sourceMap, const Matrix& A, RCP<const Map>& amalgamatedMap, Kokkos::View<LO*, memory_space> translation); // const;
 
 
   }; //class SubBlockUnAmalgamationFactory_kokkos
--- a/packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_kokkos_decl.hpp
+++ b/packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_kokkos_decl.hpp
@@ -64,7 +64,7 @@
 
 #ifdef HAVE_MUELU_KOKKOS_REFACTOR
 #include "MueLu_AmalgamationInfo_kokkos_fwd.hpp"
-#include "MueLu_Aggregates_kokkos_fwd.hpp"
+#include "MueLu_Aggregates_kokkos.hpp"
 
 namespace MueLu {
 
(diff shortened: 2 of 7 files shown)


lucbv:
@mhoemmen fair comment, to be honest this has been there for a long time, I am suspecting that guards should have been implemented to switch between the two definitions.
I do hope though that no-one is relying on this node definition as I have no idea how consistent it is with the rest of the code?
In any case I have decided to keep the public definition as it is most likely the one being used at the moment.

lucbv:
@mhoemmen actually it is fine and somewhat necessary to have both definitions.
The first one just defines a short hand tpye: `node_type` and the second one creates an alias: "Node = node_type" that is being used internally for MueLu's short names header included just after the aliasing.

mhoemmen:
@lucbv my bad -- thanks for checking!

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationFactory_kokkos_decl.hpp
```diff
@@ -56,7 +56,7 @@
 #include "MueLu_Exceptions.hpp"
 
 #ifdef HAVE_MUELU_KOKKOS_REFACTOR
-#include "MueLu_AmalgamationInfo_kokkos_fwd.hpp"
+#include "MueLu_AmalgamationInfo_kokkos.hpp"
```
jhux2: Doesn't the file `MueLu_AmalgamationInfo_kokkos.hpp` include this file, `MueLu_AmalgamationFactory_kokkos_decl.hpp`?

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_kokkos_decl.hpp
```diff
@@ -86,8 +86,11 @@ namespace MueLu {
 
   public:
 
-    AmalgamationInfo_kokkos(RCP<Array<LO> > rowTranslation,
-                            RCP<Array<LO> > colTranslation,
+    using execution_space = typename LWGraph_kokkos::execution_space;
```
jhux2: Does this need to be public?

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationFactory_kokkos_decl.hpp
lucbv: No, `MueLu_AmalgamationFactory_kokkos_{decl,def}.hpp` may include `MueLu_AmalgamationInfo_kokkos.hpp` but not the otherway around. Actually I think that `MueLu_AmalgamationInfo_kokkos_{decl,def}.hpp` have no reference to the `AmalgamationFactory_kokkos`.

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_kokkos_decl.hpp
lucbv: Maybe not but it is a type definition not a variable so I feel there is no harm in making it public?

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationFactory_kokkos_decl.hpp
jhux2: My mistake.  I misread `MueLu_AmalgamationInfo_kokkos` as `MueLu_AmalgamationFactory_kokkos`.

jhux2:
@lucbv What would you like to do with this PR?

lucbv:
Yeah, I'm not sure what it's state is to be honest...
I would actually lean toward closing it and reworking the issue if this is still something we need.

Issue #5541, opened 2019-07-16
STK: compile time error that could be configure time


tjfulle:
## Bug Report

The following logic in `packages/stk/stk_balance/stk_balance/internal/balanceTypes.hpp` can be moved to CMake and be caught at configure time.

```
#if defined(HAVE_TPETRA_INST_INT_LONG_LONG)
typedef long long BalanceGlobalNumber;
#elif defined(HAVE_TPETRA_INST_INT_LONG)
typedef long BalanceGlobalNumber;
#else
#error "stk_balance requires that Tpetra instantiate either long or long-long."
#endif
```

alanw0:
Yes I agree with this, we will figure out how to put this check in a cmake file.
Thanks!

PR #6102, opened 2019-10-15
STK:  fix for #5541  Detect insufficient global ordinal type at configure rather than build time

kddevin:
@trilinos/stk @alanw0 @bartlettroscoe 

## Description
Some Cmake voodoo to detect STKBalance's insufficient global ordinal types (int, unsigned) at configure time rather than at build time.
If the user specifies Trilinos_ENABLE_STKBalance=ON with insufficient data types, a CMake error is issued.
If STKBalance is enabled only indirectly via enabling STK and the GOs are insufficient, a CMake warning is issued and STKBalance is disabled.

Thanks to @bartlettroscoe for instructions on making these changes and pointing me to the right file.

## Motivation

STKBalance requires global ordinal types long or long long.  It checks for these types at compile time.  But by then, users may have wasted a lot of time building with a global ordinal type that is no good.  It is better to check for sufficient GO types when configuring.
This change allows Trilinos to build with GO=INT only.

#5541 #5602 

## Stakeholder Feedback

@alanw0 Please let me know if you see a problem with this change.

## Testing

On CEE-LAN, 
with and without Trilinos_ENABLE_STKBalance, 
with Trilinos_ENABLE_STKBalance=ON and OFF,
with Tpetra_INST_INT_INT=ON and OFF and unspecified,
with Tpetra_INST_INT_UNSIGNED=ON and unspecified,
with Tpetra_INST_INT_LONG=ON and unspecified,
with Tpetra_INST_INT_LONG_LONG=ON and OFF and unspecified

### DIFF
--- a/cmake/RepositoryDependenciesSetup.cmake
+++ b/cmake/RepositoryDependenciesSetup.cmake
@@ -14,12 +14,13 @@ endif()
 
 #########################################################################
 # STKBalance does not work with GO=INT or GO=UNSIGNED
-# Warn and disable it if it was set as a dependence of STK
+# Note and disable it if it was set as a dependence of STK
 # Error out if it was explicitly requested by user
 
-IF ("${Tpetra_ENABLE_DEPRECATED_CODE}" STREQUAL "OFF")  # Remove this test when
-                                                        # Tpetra DEPRECATED
-                                                        # code is removed
+IF ((NOT ("${Tpetra_ENABLE_DEPRECATED_CODE}" STREQUAL "")) # Remove this test
+    AND                                                    # when removing
+    (NOT Tpetra_ENABLE_DEPRECATED_CODE))                   # Tpetra DEPRECATED
+                                                           # code 
   SET(KDD_ENABLE_DEPRECATED OFF)  # Remove this line when DEPRECATED is removed
   SET(KDD_INT_INT OFF)     # Current default
 ELSE() # Remove this "else" section when Tpetra DEPRECATED code is removed
@@ -65,16 +66,16 @@ ENDIF()
 IF ((NOT ${KDD_INT_LONG}) AND (NOT ${KDD_INT_LONG_LONG}))
   IF ("${${PROJECT_NAME}_ENABLE_STKBalance}" STREQUAL "")
     # STKBalance may be enabled but only implicitly (as a dependence of STK);
-    # give a warning but turn off STKBalance support
-    MESSAGE(WARNING "int global indices are enabled in Trilinos. "
-                    "Because STKBalance requires long or long long "
-                    "global indices, STKBalance will be disabled.  "
-                    "To make this warning go away, do not request "
-                    "int global indices in Trilinos (that is,  do not "
-                    "set Tpetra_INST_INT_INT=ON or "
-                    "Tpetra_INST_INT_UNSIGNED=ON)." )
+    # give a message but turn off STKBalance support
+    MESSAGE("NOTE:  int global indices are enabled in Trilinos. "
+            "Because STKBalance requires long or long long "
+            "global indices, STKBalance will be disabled.  "
+            "To make this warning go away, do not request "
+            "int global indices in Trilinos (that is,  do not "
+            "set Tpetra_INST_INT_INT=ON or "
+            "Tpetra_INST_INT_UNSIGNED=ON)." )
     SET(${PROJECT_NAME}_ENABLE_STKBalance OFF)
-  ELSEIF (${${PROJECT_NAME}_ENABLE_STKBalance})
+  ELSEIF (${PROJECT_NAME}_ENABLE_STKBalance)
     # STKBalance was explicitly enabled by the user, so error out
     MESSAGE(FATAL_ERROR 
             "STKBalance requires long or long long global indices, "
--- a/cmake/RepositoryDependenciesSetup.cmake
+++ b/cmake/RepositoryDependenciesSetup.cmake
@@ -64,7 +64,7 @@ ENDIF()
 
 IF ((NOT ${KDD_INT_LONG}) AND (NOT ${KDD_INT_LONG_LONG}))
   IF ("${${PROJECT_NAME}_ENABLE_STKBalance}" STREQUAL "")
-    # STKBalance was enabled but only implicitly (as a dependence of STK);
+    # STKBalance may be enabled but only implicitly (as a dependence of STK);
     # give a warning but turn off STKBalance support
     MESSAGE(WARNING "int global indices are enabled in Trilinos. "
                     "Because STKBalance requires long or long long "
@@ -74,7 +74,7 @@ IF ((NOT ${KDD_INT_LONG}) AND (NOT ${KDD_INT_LONG_LONG}))
                     "set Tpetra_INST_INT_INT=ON or "
                     "Tpetra_INST_INT_UNSIGNED=ON)." )
     SET(${PROJECT_NAME}_ENABLE_STKBalance OFF)
-  ELSE()
+  ELSEIF (${${PROJECT_NAME}_ENABLE_STKBalance})
     # STKBalance was explicitly enabled by the user, so error out
     MESSAGE(FATAL_ERROR 
             "STKBalance requires long or long long global indices, "
--- a/cmake/RepositoryDependenciesSetup.cmake
+++ b/cmake/RepositoryDependenciesSetup.cmake
@@ -11,3 +11,76 @@ if (TPL_ENABLE_CUDA AND NOT Kokkos_ENABLE_Cuda_Relocatable_Device_Code)
     message(FATAL_ERROR "ERROR: ${PROJECT_NAME}_ENABLE_ShyLU_NodeTacho=ON but TPL_ENABLE_CUDA='${TPL_ENABLE_CUDA}' AND Kokkos_ENABLE_Cuda_Relocatable_Device_Code='${Kokkos_ENABLE_Cuda_Relocatable_Device_Code}' which is not allowed!")
   endif()
 endif() 
+
+#########################################################################
+# STKBalance does not work with GO=INT or GO=UNSIGNED
+# Warn and disable it if it was set as a dependence of STK
+# Error out if it was explicitly requested by user
+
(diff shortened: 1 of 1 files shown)


cmake/RepositoryDependenciesSetup.cmake
```diff
@@ -10,4 +10,77 @@ if (TPL_ENABLE_CUDA AND NOT Kokkos_ENABLE_Cuda_Relocatable_Device_Code)
...
+
+#########################################################################
+# STKBalance does not work with GO=INT or GO=UNSIGNED
+# Warn and disable it if it was set as a dependence of STK
+# Error out if it was explicitly requested by user
+
+IF ("${Tpetra_ENABLE_DEPRECATED_CODE}" STREQUAL "OFF")  # Remove this test when
+                                                        # Tpetra DEPRECATED
+                                                        # code is removed
+  SET(KDD_ENABLE_DEPRECATED OFF)  # Remove this line when DEPRECATED is removed
```
bartlettroscoe: what does "KDD" stand for?

cmake/RepositoryDependenciesSetup.cmake
```diff
@@ -10,4 +10,77 @@ if (TPL_ENABLE_CUDA AND NOT Kokkos_ENABLE_Cuda_Relocatable_Device_Code)
...
-endif() 
\ No newline at end of file
+endif() 
+
+#########################################################################
+# STKBalance does not work with GO=INT or GO=UNSIGNED
+# Warn and disable it if it was set as a dependence of STK
+# Error out if it was explicitly requested by user
+
+IF ("${Tpetra_ENABLE_DEPRECATED_CODE}" STREQUAL "OFF")  # Remove this test when
```
bartlettroscoe: I think this needs to be:

```
IF ("${Tpetra_ENABLE_DEPRECATED_CODE}" NOT STREQUAL "" AND NOT Tpetra_ENABLE_DEPRECATED_CODE)
```

Otherwise, if the user sets `-DTpetra_ENABLE_DEPRECATED_CODE=FALSE` (or `NO` or `false`, or any of the string values that CMake recognizes as "false") then it will fail this logic (and confuse users).  See:

* https://tribits.org/doc/TribitsDevelopersGuide.html#how-to-check-for-and-tweak-tribits-enable-cache-variables

cmake/RepositoryDependenciesSetup.cmake
```diff
@@ -10,4 +10,77 @@ if (TPL_ENABLE_CUDA AND NOT Kokkos_ENABLE_Cuda_Relocatable_Device_Code)
...
+    # give a warning but turn off STKBalance support
+    MESSAGE(WARNING "int global indices are enabled in Trilinos. "
+                    "Because STKBalance requires long or long long "
+                    "global indices, STKBalance will be disabled.  "
+                    "To make this warning go away, do not request "
+                    "int global indices in Trilinos (that is,  do not "
+                    "set Tpetra_INST_INT_INT=ON or "
+                    "Tpetra_INST_INT_UNSIGNED=ON)." )
+    SET(${PROJECT_NAME}_ENABLE_STKBalance OFF)
+  ELSEIF (${${PROJECT_NAME}_ENABLE_STKBalance})
```
bartlettroscoe: This might issue a warning.  I think you want:

```
ELSEIF (${PROJECT_NAME}_ENABLE_STKBalance)
```

One should not dereference a variable name unless comparing to an explicit value.

cmake/RepositoryDependenciesSetup.cmake
```diff
@@ -10,4 +10,77 @@ if (TPL_ENABLE_CUDA AND NOT Kokkos_ENABLE_Cuda_Relocatable_Device_Code)
...
+
+IF(NOT ("${Tpetra_INST_INT_LONG_LONG}" STREQUAL ""))
+  SET(KDD_INT_LONG_LONG ${Tpetra_INST_INT_LONG_LONG})
+ENDIF()
+
+IF ((NOT ${KDD_INT_LONG}) AND (NOT ${KDD_INT_LONG_LONG}))
+  IF ("${${PROJECT_NAME}_ENABLE_STKBalance}" STREQUAL "")
+    # STKBalance may be enabled but only implicitly (as a dependence of STK);
+    # give a warning but turn off STKBalance support
+    MESSAGE(WARNING "int global indices are enabled in Trilinos. "
```
bartlettroscoe: Should this really be a `MESSAGE(WARNING ...)` instead of a `MESSAGE("NOTE: " ...)`?  I know that some projects like Charon2 really don't like CMake warnings unless they are real warnings.  But I guess they can make this go away by just explicitly setting `-DTrilinos_ENABLE_STKBalance=OFF`, right?

bartlettroscoe: Just to be clear, if I set `-DTpetra_INST_INT_INT=OFF`, then I should not get this warning?

alanw0:
@kddevin Looks like there is some discussion with @bartlettroscoe , I'm happy when he's happy since he knows more about the cmake issues. Once it's in I'll try to get it back to sierra so that it doesn't get overwritten with the next stk snapshot into trilinos.

bartlettroscoe:
@alanw0 said:

> Once it's in I'll try to get it back to sierra so that it doesn't get overwritten with the next stk snapshot into trilinos.

That will not be needed as no files under packages/stk/ are modified in this PR.

alanw0:
@bartlettroscoe ok, I obviously didn't look closely enough, just assumed it modified stk files since the PR had STK  in the title. Carry on. :-)

cmake/RepositoryDependenciesSetup.cmake
kddevin: KDD == my initials.  Lower risk of conflict with other variables than "TMP_" or similar. :)

kddevin:
@bartlettroscoe Thanks for your review.  I will make revisions this afternoon.

cmake/RepositoryDependenciesSetup.cmake
kddevin: Good!  Shall do.

cmake/RepositoryDependenciesSetup.cmake
kddevin: Right, with either Trilinos_ENABLE_STKBalance=OFF or Tpetra_INST_INT_INT=OFF, there would be no message.
Also, there's no message in the current default Trilinos configuration (with both GO=int and GO=long long) or the soon-to-be default configuration (with GO=long long) since long long GOs are available.  
That said, if you say "NOTE" is better than "WARNING," I am fine with that.  I'll follow your judgement on CMake etiquette.

cmake/RepositoryDependenciesSetup.cmake
kddevin: Got it. Thanks!

kddevin:
@bartlettroscoe I made all the changes you requested.  Thanks for the great review.  Would you please re-review?  Thanks!

cmake/RepositoryDependenciesSetup.cmake
```diff
@@ -65,16 +66,16 @@ ENDIF()
...
-    # give a warning but turn off STKBalance support
-    MESSAGE(WARNING "int global indices are enabled in Trilinos. "
-                    "Because STKBalance requires long or long long "
-                    "global indices, STKBalance will be disabled.  "
-                    "To make this warning go away, do not request "
-                    "int global indices in Trilinos (that is,  do not "
-                    "set Tpetra_INST_INT_INT=ON or "
-                    "Tpetra_INST_INT_UNSIGNED=ON)." )
+    # give a message but turn off STKBalance support
+    MESSAGE("NOTE:  int global indices are enabled in Trilinos. "
```
bartlettroscoe: I could go either way but I think "NOTE:" is a better choice in this case so thanks for making that change.

cmake/RepositoryDependenciesSetup.cmake
```diff
@@ -10,4 +10,78 @@ if (TPL_ENABLE_CUDA AND NOT Kokkos_ENABLE_Cuda_Relocatable_Device_Code)
...
-endif() 
\ No newline at end of file
+endif() 
+
+#########################################################################
+# STKBalance does not work with GO=INT or GO=UNSIGNED
+# Note and disable it if it was set as a dependence of STK
+# Error out if it was explicitly requested by user
+
+IF ((NOT ("${Tpetra_ENABLE_DEPRECATED_CODE}" STREQUAL "")) # Remove this test
```
mhoemmen: I think the usual idiom is this:
```
IF ((DEFINED Tpetra_ENABLE_DEPRECATED_CODE) AND (NOT Tpetra_ENABLE_DEPRECATED_CODE))
  # stuff ...
ENDIF()
```

bartlettroscoe: > I think the usual idiom is this:

@mhoemmen, that is actually not correct as explained in:

* https://tribits.org/doc/TribitsDevelopersGuide.html#how-to-check-for-and-tweak-tribits-enable-cache-variables

which says:

> First, note that most of these particular ENABLE variables are not BOOL cache variables but are actually STRING variables with the possible values of ON, OFF and empty "" (see the macro `SET_CACHE_ON_OFF_EMPTY()`). Therefore, just because the value of a `<XXX>_ENABLE_<YYY> `variable is defined (e.g. `IF (DEFINED <XXX>_ENABLE_<YYY>) ... ENDIF())` does not mean that it has been set to ON or OFF yet (or any non-empty values that evaluates to true or false in CMake). To see if an ENABLE variable is one of these variables, look in the CMakeCache.txt file for the type. If the type is STRING, then it is most likely this type of variable with a default value of empty "". However, if the cache type is `BOOL` then it is likely a standard bool variable that is not allowed to have a value of empty "".

mhoemmen: uh oh never mind then :)

kddevin:
@trilinos/framework @prwolfe   This PR seems to be stuck in testing that began yesterday.  The PR doesn't change any source code.  Is there something wrong with it?  Or is testing down?  Thanks!

kddevin:
Testing delayed due to #3276.  Stay tuned.

Issue #5619, opened 2019-07-31
Trilinos: Add support for scalable, advanced architecture performance testing


jjellio:
Performance tracking seems to come up frequently, with various groups all hand-rolling approaches.  Trilinos as a product would benefit from an extensible testing framework that supports performance testing. Since we utilize Cmake, Ctest, and Cdash in many places (e.g., projects beyond Trilinos have adopted this), I way forward for performance tracking would be to leverage workflows that projects are familiar with.

At the heart of this issue are a few basic requirements:

1. The ability for product owners to describe performance tests
2. The ability to run the performance test in suitable manner (e.g., multi-node, with proper process binding, thread settings, and device utilization)
3. The ability to save seemingly arbitrary data
4. The ability for diverse product teams to access and visualize that data how they see fit

The prior list seems a bit lofty, but **2** is actively being tackled in issue #5598. **3** is semi-solved - CDash is used daily by teams to assess whether builds completed and unit tests succeeded. Where CDash is lacking is 1) exposing app specific data; and 2) providing robust reporting tools, which is effectively **4**; and 3) gathering data that is suitable for performance analysis.

To that end, I think leveraging the tools and people already in place is scalable way forward to enable diverse performance tracking.

To explore this, I wrote some python scripts that slurp CDash builds and runs.  From that data, I can then populate a database to track all sorts of things.  Since CTest effectively stores the output of each run, I can then parse and retain application specific timings. (e.g., associate Teuchos Timers with the data already present in CDash).  With that data, I then setup a docker Grafana instance to visualize the data.

The above process worked.  I can visualize teuchos timer information for specific apps (all unit tests), over time for the past 8 months.

Incidentally, Irina (@ikalash) has a dashboard where she is using CTest to drive performance testing for Mini-EM, and she and Jerry are mining the CDash for some performance data.  So it seems the approach is semi-attractive (not perfect, but it meshes with workflows that we are accustomed to).

If we can performance testing as tractable as defining a `unit_test`, then customers such as ATDM which build and run on production machines can utilize the approach define and run performance tests. This lets apps define the tests, and ultimately view the results.  As a support-type person, this puts me in the position to focus on providing better performance tracking.

Where the above process is falling short is in two areas:
1. We cannot define a 'performance' tests.
[...]

PR #6107, opened 2019-10-16
Initial Tpetra stack nightly performance tests

brian-kelley:
Tpetra:        Added CGSolve strong scaling (MPI procs = 1, 4, 9, ...) tests
Kokkos/Zoltan: Fixed test categories, test exectuable names for perf tests
MueLu:         Perf tests for driver (Laplace3D setup+solve) and mat-mat

@trilinos/muelu 
@trilinos/tpetra 

## Motivation

This change is part of the new automated nightly performance testing effort. It tests the following performance-critical features of the Tpetra solver stack in a nightly rocketman build:

- Finite element assembly (MPI strong scaling test)
- CG solve (MPI strong scaling test)
- Matrix-matrix multiply
- Full MueLu setup + solve on a Laplace3D problem from Galeri. This one exercises MueLu smoothed aggregation, Ifpack2 Chebyshev smoothing and the Belos CG solver. 

All of these tests produce clean Teuchos timer summaries at the end (to stdout). Later, it will be possible to download the timings from CDash and plot them over time.

I also fixed some places in CMake (including in Kokkos and Zoltan) where the "PERFORMANCE" category was missing from ``TRIBITS_ADD_EXECUTABLE`` and ``TRIBITS_COPY_FILES`` calls. Now, the performance tests in those packages run cleanly (have the files they need) when the test category is set to PERFORMANCE.

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #5619
* Part of 
* Composed of 

## Stakeholder Feedback

Every single Tpetra stakeholder had performance as one of their FY20 requests. This will help us track improvements and catch regressions before they affect applications. Emphasis/Nevada explicitly asked for performance testing. See #5619 for a lot of discussion about the performance testing system in general.
## Testing

These are all changes to tests. I did a dry run of the nightly performance test on rocketman.


packages/zoltan/src/driver/CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ TRIBITS_ADD_EXECUTABLE(
   NOEXEPREFIX
   SOURCES ${COMMON_DRIVER_HDRS} ${ZDRIVE_SOURCES}
   COMM serial mpi
+  CATEGORIES BASIC NIGHTLY PERFORMANCE
```
kddevin: @brian-kelley Thanks for retrofitting zoltan; I did not realize there were issues with its tests.
Why are there three categories here but only two in zoltan/test/CMakeLists.txt:150  below?    
Are categories also needed for the zCPPdrive executable below (line 95)?

brian-kelley: Good catch. After some reading about tribits I think I understand that the categories say how expensive tests are to run: basic <= continuous < nightly < heavy, and then performance is completely on its own (needs to be added explicitly, and excludes basic if you don't also ask for basic). Since "nightly" is redundant after making it "basic", I just deleted that (and added basic to the zoltan/test file).

The default category (basic) will be fine for zCPPdrive because it doesn't get used by any performance tests.

jjellio:
This stuff sounds good, but we really won't know how this will work till post-TUG.

I think it is generally safe to say that CMake/Ctest *will* be a supported way forward. Even with uncertainty around how everything will work, ultimately having test programs that can dump relevant timing information is always required.

brian-kelley:
FYI, the Kokkos cmake changes can just live in Trilinos until they get clobbered by the next Kokkos develop->snapshot promotion. That promotion will overhaul Kokkos cmake, and it will no longer use Tribits (or the categories) to manage its tests.

PR #6122, opened 2019-10-18
Framework: Remove jquery.js from documentation files

william76:
@trilinos/framework 

Remove `jquery.js` because it fails certain security checks for exploits.

FYI @jwillenbring


william76:
@jwillenbring

PR #6131, opened 2019-10-21
PullRequestLinuxDriver.sh: remove ".sandia.gov" from the no_proxy env…

prwolfe:
…ironemt variable

Testing shows this works on the ascic machines as well
as the cloud nodes for both testing (although it is
currently showing "500 error" and for testing-dev. Lets
hope that it holds up for a while.

Why our networks are not more stable is beyond me...

This will have to be force pushed of course as no PR's are currently functional.


prwolfe:
Jim Glenn confirmed for me that this change is real and needed. He is unsure who made it or if it will remain consistent.

prwolfe:
Spoke with Jim W. and am pushing to try and get this re-started.

PR #6145, opened 2019-10-23
Tempus 6141 no getter functions for output interval and output indices in time step control

asteyer:
@trilinos/tempus 

This fixes issue 6141 by adding getter functions for output time interval and output indices in Tempus_TimeStepControl


packages/tempus/src/Tempus_TimeStepControl_decl.hpp
```diff
@@ -129,6 +129,11 @@ class TimeStepControl
       { return tscPL_->get<int>("Number of Time Steps"); }
     virtual Teuchos::RCP<TimeStepControlStrategyComposite<Scalar>>
        getTimeStepControlStrategy() const { return stepControlStrategy_;}
+    virtual int getOutputIndexInterval()
+      { return outputIndexInterval_;}
+    virtual Scalar setOutputTimeInterval()
```
ccober6: I think this should be getOutputTimeInterval().  Do you have tests to cover these additional functions?

asteyer: Not in tempus - where would be the best place to put a test for this in tempus?

ccober6: I just added a TimeStepControl unit test with #6144 (tempus/unit_test/Tempus_UnitTest_TimeStepControl.cpp) (merged this morning).  Add another test to it.

PR #6150, opened 2019-10-24
ATDM: Do rebuilds by default for all ATDM Trilinos builds

bartlettroscoe:
This will massively speed up the build times and will also test to see how robust the Trilinos CMake build system is to rebuilds.

## How was this tested?

```
$ ssh ceerws1113

$ cd /scratch/rabartl/Trilinos.base/BUILDS/ATDM/SEMS-RHEL6/CTEST_S

$  env Trilinos_PACKAGES=Kokkos
    ./ctest-s-local-test-driver.sh sems-rhel6-gnu-7.2.0-openmp-release-debug
```

The log file showed:

````
-- CTEST_START_WITH_EMPTY_BINARY_DIRECTORY='FALSE'
````

### DIFF
--- a/cmake/ctest/drivers/atdm/TrilinosCTestDriverCore.atdm.cmake
+++ b/cmake/ctest/drivers/atdm/TrilinosCTestDriverCore.atdm.cmake
@@ -52,7 +52,7 @@ MACRO(TRILINOS_SYSTEM_SPECIFIC_CTEST_DRIVER)
     IF ("$ENV{ATDM_CONFIG_BUILD_COUNT}" GREATER "0")
       SET(CTEST_BUILD_FLAGS "-j$ENV{ATDM_CONFIG_BUILD_COUNT} ")
     ELSE()
-      SET(CTEST_BUILD_FLAGS "")
+      SET(CTEST_BUILD_FLAGS "")  # Use all cores!
     ENDIF()
     SET(CTEST_BUILD_FLAGS "${CTEST_BUILD_FLAGS}-k 999999")
   ELSE()
@@ -154,6 +154,9 @@ MACRO(TRILINOS_SYSTEM_SPECIFIC_CTEST_DRIVER)
     # for the number of new warnings, errors, tests, etc.
     SET(CTEST_SITE "$ENV{ATDM_CONFIG_CDASH_HOSTNAME}")
   ENDIF()
+
+  # Let's be brave and do all rebuilds (massively speed up build times)
+  SET(CTEST_START_WITH_EMPTY_BINARY_DIRECTORY FALSE)
 
   # Don't process any extra repos
   SET(Trilinos_EXTRAREPOS_FILE NONE)


rppawlo:
With this change, how will you handle changes that require a wipe of the build directories? I'm thinking of when we refactor to remove autogenerated header files.

bartlettroscoe:
> With this change, how will you handle changes that require a wipe of the build directories? I'm thinking of when we refactor to remove autogenerated header files.

We will just have to temporarily set `CTEST_START_WITH_EMPTY_BINARY_DIRECTORY=TRUE` for the next testing day.

But I want to see if we can make Panzer and other package rebuilds robust to changes in autogenerated header files.   Would also want to understand why the autogenerated header files are getting removed because they are critical to avoid unnecessary rebuilds with Cmake.

rppawlo:
> Would also want to understand why the autogenerated header files are getting removed because they are critical to avoid unnecessary rebuilds with Cmake.

The autogenerated headers are only needed if you want to toggle between ETI enabled and disabled. If you pick one path (e.g. always do ETI) then everything rebuilds correctly since the header includes are hand coded to be correct. In Panzer's case, the instantiations are evaluation types (Residual, Jacobian, Hessian, ...) that we know at configure time. We do not need to ever turn ETI off. So as we touch classes, we remove the ETI on/off support and just always do ETI. The history here is that about 9 years ago there was a compiler on an HPC machine that seg faulted when attempting to compile ETI code. I'll give you one guess as to the compiler :) So at that time we had to have the ability to disable all ETI in Trilinos/Drekar just to compile on this machine. The compiler can now do ETI correctly so we don't need the toggle anymore.

bartlettroscoe:
@rppawlo said:

> The compiler can now do ETI correctly so we don't need the toggle anymore.

What if someone tries to use Panzer classes with a custom scalar type?  The reason we made Thyra, for example, support ETI and implicit TI was to allow for the option (and also address machines were ETI did not work, like in the case you mentioned).

The next time one of these Panzer ETI refactorings takes place, I will try to carefully examine what is causing the rebuild error and see if there is a way to automatically address this in a rebuild.

rppawlo:
Our template types are different than other packages. Adding a new evaluation type requires specializations for things like gather/scatter evaluators. If users really want a new scalar type we would add it to panzer using ETI. Given the limited manpower and complaints about build times, we aren't interested in supporting a no-ETI path. Compilers have been compliant on this front for a long time now.

The reason why it breaks is known. If both the file_decl.hpp and file_impl.hpp exist then ETI autogenerates a header in the bin directory. When we remove the no-ETI path, we move the code from file_decl.hpp into file.hpp to disable autogeneration. This is the same name as the autogenerated header. So if the compiler finds the old autogenerated header in bin instead of the header with the same name in the source directory it crashes.

bartlettroscoe:
> So if the compiler finds the old autogenerated header in bin instead of the header with the same name in the source directory it crashes.

Okay, thanks for explaining that.   I think we can resolve that by adding a cmake function that will automatically remove a file `<something>.h` in the build tree if the is a file of the same name in the source tree.

Please ping me on a PR the next time Panzer removes one or more of these generated generated `<basename>.h` headers and I will add CMake code to automatically remove an existing generated `<basename>.h` file at configure time (if a matching file exists in the source tree).

There is such a massive advantage to rebuilding that we should put some work in to ensure that Trilinos can be robustly rebuilt.  I think it will reduce the cost of our ATDM Trilinos builds by more than a factor of 2x on most days.

Issue #6149, opened 2019-10-23
MueLu, Xpetra, Tpetra:  Error in Add() with DEPRECATED code disabled


kddevin:
## Bug Report
@trilinos/muelu @trilinos/xpetra @trilinos/tpetra 

### Description
@rppawlo is seeing a DynamicProfile deprecation error from xpetra when running in his application.  @rppawlo says:

The error and stack trace is below. Looks like the MatrixMatrix::Add being used by MueLu RefMaxwell preconditioner is not allocating enough space in construction. This is in an EMPIRE-PIC unit test. What’s interesting is that this error happens on the test with one mpi process. With multiple mpi processes, the same exact test passes.

```


 

*********** Caught Exception: Begin Error Report ***********

/home/rppawlo/Trilinos/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp:2299:

 

Throw number = 1

 

Throw test that evaluated to true: (numInserted == Teuchos::OrdinalTraits<size_t>::invalid())

 

Tpetra::CrsGraph<int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> >::insertGlobalIndicesImpl: There is not enough capacity to insert indices in to row 24. The upper bound on the number of entries in this row must be increased to accommodate one or more of the new indices.

************ Caught Exception: End Error Report ************

 

 

(gdb) bt

#0  __cxxabiv1::__cxa_throw (obj=0x1227d650, tinfo=0xa0918d0 <typeinfo for std::runtime_error>,

    dest=0x17aec90 <_ZNSt13runtime_errorD1Ev@plt>) at ../../.././libstdc++-v3/libsupc++/eh_throw.cc:75

#1  0x0000000004c50eae in Tpetra::CrsGraph<int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> >::insertGlobalIndicesImpl(Tpetra::RowInfo const&, long long const*, unsigned long, std::function<void (unsigned long, unsigned long, unsigned long)>) ()

[...]

PR #6152, opened 2019-10-24
Xpetra MM add: Delegate allocation of new matrix to Tpetra

cgcgcg:
@trilinos/xpetra 

## Motivation
Tpetra can decide itself how big of a matrix should be allocated to hold the sum of two matrices.

This should take care of #6149.

### DIFF
--- a/packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
+++ b/packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
@@ -566,53 +566,8 @@ Note: this class is not in the Xpetra_UseShortNames.hpp
           throw Exceptions::Incompatible("TwoMatrixAdd: matrix row maps are not the same.");
 
 
-        if (C == Teuchos::null) {
-          size_t maxNzInA     = 0;
-          size_t maxNzInB     = 0;
-          size_t numLocalRows = 0;
-          if (A.isFillComplete() && B.isFillComplete()) {
-            maxNzInA     = A.getNodeMaxNumRowEntries();
-            maxNzInB     = B.getNodeMaxNumRowEntries();
-            numLocalRows = A.getNodeNumRows();
-          }
-          if (maxNzInA == 1 || maxNzInB == 1 || AHasFixedNnzPerRow) {
-            // first check if either A or B has at most 1 nonzero per row
-            // the case of both having at most 1 nz per row is handled by the ``else''
-            Teuchos::ArrayRCP<size_t> exactNnzPerRow(numLocalRows);
-
-            if ((maxNzInA == 1 && maxNzInB > 1) || AHasFixedNnzPerRow) {
-              for (size_t i = 0; i < numLocalRows; ++i)
-                exactNnzPerRow[i] = B.getNumEntriesInLocalRow(Teuchos::as<LO>(i)) + maxNzInA;
-
-            } else {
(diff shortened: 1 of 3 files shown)


packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
```diff
@@ -622,7 +577,13 @@ Note: this class is not in the Xpetra_UseShortNames.hpp
...
           RCP<Tpetra::CrsMatrix<SC,LO,GO,NO> >  tpC = Xpetra::Helpers<SC,LO,GO,NO>::Op2NonConstTpetraCrs(C);
 
-          Tpetra::MatrixMatrix::Add(tpA, transposeA, alpha, tpB, transposeB, beta, tpC);
+          RCP<ParameterList> params = rcp(new ParameterList(""));
+          // params->set("Call fillComplete",false);
+          Tpetra::MatrixMatrix::add(alpha, transposeA, tpA, beta, transposeB, tpB, *tpC,
+                                    Xpetra::toTpetra(transposeA ? A.getRangeMap() : A.getDomainMap()),
+                                    Xpetra::toTpetra(transposeA ? A.getDomainMap() : A.getRangeMap()),
+                                    params);
+          C->resumeFill();
```
cgcgcg: I'd rather not resume fill here, but tell Tpetra to not fill-complete the matrix. That bugs out, however.

brian-kelley: @cgcgcg You're right, Tpetra didn't behave correctly if "Call fillComplete" is set to false in the parameter list. The documentation for add() promises that params is interpreted the same way RowMatrix::add() does. This is an easy fix, can I just send you a patch to include in this PR?

packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
```diff
@@ -622,7 +577,13 @@ Note: this class is not in the Xpetra_UseShortNames.hpp
           const Tpetra::CrsMatrix<SC,LO,GO,NO>& tpB = Xpetra::Helpers<SC,LO,GO,NO>::Op2TpetraCrs(B);
           RCP<Tpetra::CrsMatrix<SC,LO,GO,NO> >  tpC = Xpetra::Helpers<SC,LO,GO,NO>::Op2NonConstTpetraCrs(C);
 
-          Tpetra::MatrixMatrix::Add(tpA, transposeA, alpha, tpB, transposeB, beta, tpC);
+          RCP<ParameterList> params = rcp(new ParameterList(""));
+          // params->set("Call fillComplete",false);
+          Tpetra::MatrixMatrix::add(alpha, transposeA, tpA, beta, transposeB, tpB, *tpC,
+                                    Xpetra::toTpetra(transposeA ? A.getRangeMap() : A.getDomainMap()),
```
brian-kelley: add() always uses the domain and range map of B (or B^T if transposeB) as the domain and range map for C (the documentation is very wrong here, it says it uses the rowmap of B for both). But since you want C to use A's domain/range map here, you could just pass in B before A.

packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
cgcgcg: Sure.

PR #6156, opened 2019-10-24
Adds option to use ICGS as orthogonalization method in BKS and LOBPCG

hkthorn:
@trilinos/anasazi 

## Motivation

The ICGS orthogonalization method has been an option in Anasazi, but has
not been available for some more commonly used eigensolvers, like BKS and
LOBPCG.  Similar to the study performed for Belos, it is of interest to
use this orthogonalization method to see if it improves performance on
more advanced architectures.

## Testing

Tests were modified to include ICGS orthogonalization.

Tests pass on:  OSX, serial, GCC 9.x

### DIFF
--- a/packages/anasazi/epetra/test/LOBPCG/cxx_main.cpp
+++ b/packages/anasazi/epetra/test/LOBPCG/cxx_main.cpp
@@ -78,6 +78,8 @@ int main(int argc, char *argv[])
 #endif
   MyPID = Comm.MyPID();
 
+  int nev = 5;
+  int blockSize = -1;
   bool testFailed;
   bool verbose = false;
   bool debug = false;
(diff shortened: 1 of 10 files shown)


packages/anasazi/epetra/test/LOBPCG/cxx_main.cpp
```diff
@@ -190,11 +196,13 @@ int main(int argc, char *argv[])
...
   MyPL.set( "Locking Tolerance", tol/10 );
   MyPL.set( "Full Ortho", fullOrtho );
+  MyPL.set( "Orthogonalization", ortho );
+
   if (testRecovery) {
     // initialize with bad P
     RCP<Anasazi::LOBPCGState<ScalarType,MV> > badstate = rcp( new Anasazi::LOBPCGState<ScalarType,MV>() );
-    Anasazi::SVQBOrthoManager<ScalarType,MV,OP> ortho(M);
-    ortho.normalize(*ivec,null);
+    Anasazi::SVQBOrthoManager<ScalarType,MV,OP> orthoM(M);
```
mhoemmen: If users can specify the orthogonalization method, then why is SVQBOrthoManager hard-coded here?  I'm guessing that this a special case and not the main orthogonalization loop in LOBPCG.

packages/anasazi/src/AnasaziBlockKrylovSchurSolMgr.hpp
```diff
@@ -388,7 +389,7 @@ BlockKrylovSchurSolMgr<ScalarType,MV,OP>::BlockKrylovSchurSolMgr(
 
   // which orthogonalization to use
   ortho_ = pl.get("Orthogonalization",ortho_);
-  if (ortho_ != "DGKS" && ortho_ != "SVQB") {
+  if (ortho_ != "DGKS" && ortho_ != "SVQB" && ortho_ != "ICGS") {
```
mhoemmen: Anasazi doesn't have "IMGS" (while Belos does), right?

hkthorn: Correct.  There isn't IMGS in Anasazi, only those three methods.  An orthogonalization factory would be preferable, but was never implemented in Anasazi.

packages/anasazi/epetra/test/LOBPCG/cxx_main.cpp
hkthorn: This is a special case where we are trying to create a bad subspace to make sure the recovery mechanism in the solver works.  The orthogonalization method is not critical to this part of the test.

packages/anasazi/src/AnasaziBlockKrylovSchurSolMgr.hpp
mhoemmen: Cool; thanks for explaining!

packages/anasazi/epetra/test/LOBPCG/cxx_main.cpp
mhoemmen: Cool; got it!

Issue #6159, opened 2019-10-25
ifpack2: fix creation of vectors and cache instead


bathmatt:
## Bug Report
@trilinos/ifpack2 

In the damped residual code you have
```
    RCP<vector_type> W_vec = W.getVectorNonConst (0);
    RCP<vector_type> B_vec = B.getVectorNonConst (0);
    RCP<vector_type> X_vec = X.getVectorNonConst (0);
 
```

which creates currently 18 launches every iteration,  find a way to not have launches for  these, cache or something.

### Description
What went wrong?  What should have happened?  Do you have an idea what might
fix things?  How has this bug affected you?  What are you trying to accomplish?
What specifically must we demonstrate to call this work complete?

### Steps to Reproduce
1. SHA1:  [insert here]
1. Configure script:  [attach here]
1. Configure log:  [attach here]
1. Build log:  [attach here]
1. Input deck:  [attach here]
1. Do this.
1. Do that.
1. Shake fist angrily at computer.
1. Run log:  [attach here]

PR #6160, opened 2019-10-25
Ifpack2 ScaledDampedResidual: Cache vectors

cgcgcg:
@trilinos/ifpack2 

## Motivation
Cache Vectors in Chebyshev. Should reduce view allocations.

#6159

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_ScaledDampedResidual_decl.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_ScaledDampedResidual_decl.hpp
@@ -111,6 +111,9 @@ private:
   std::unique_ptr<vector_type> X_colMap_;
   std::unique_ptr<multivector_type> V1_;
 
+  typename multivector_type::dual_view_type::t_host viewW_, viewB_, viewX_;
+  Teuchos::RCP<vector_type> W_vec_, B_vec_, X_vec_;
+
   // Do the Import, if needed, and return the column Map version of X.
   vector_type&
   importVector (vector_type& X_domMap);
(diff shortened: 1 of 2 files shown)


packages/ifpack2/src/Ifpack2_Details_ScaledDampedResidual_decl.hpp
```diff
@@ -111,6 +111,9 @@ class ScaledDampedResidual {
   std::unique_ptr<vector_type> X_colMap_;
   std::unique_ptr<multivector_type> V1_;
 
+  typename multivector_type::dual_view_type::t_host viewW_, viewB_, viewX_;
```
mhoemmen: @cgcgcg It's better not to store the Views.  (You were there for that part of my presentation Thursday morning, right? ;-) )  Here's why:

1. You only ever use the Views to check for aliasing, and you only ever use the raw pointer (result of `.data()`).

2. Storing the Views means you have to remember to update them if you ever need to resize or replace the Vectors.

Furthermore, you're storing rank-2 Views even though you're always using Vectors.

cgcgcg: @mhoemmen What's the correct way then to check whether two MultiVectors are identical? There is no `operator==`, right?

mhoemmen: @cgcgcg Don't store the Views.  Just get them temporarily where you need them.  Don't store them in the class.

stanmoore1:
Can we get resolution on this? I want to do some performance profiling of EMPIRE and it would be nice to have this optimization.

cgcgcg:
@stanmoore1 What is needed for resolution? It's merged into Trilinos, so you should be able to test it in EMPIRE.

stanmoore1:
Ah great I didn't see it was merged, thanks

PR #6171, opened 2019-10-28
Tpetra/Ifpack2/MueLu/Xpetra/FROsch: Residual function implementation

csiefer2:
Creates a Tpetra-level residual function with a placeholder implementation. Percolates this through MueLu's Residual function as well as Ifpack2's Jacobi, Richardson and Chebyshev.

This function makes no change in existing code behavior, but is merely a code refactor allowing the next PR to implement at Tpetra-level fused residual kernel.

Testing(Tpetra): CrsMatrix_gaussSeidel.
Testing(MueLu): As this changes the innards of the Residual function, all of the Hierarchy & ParameterList tests exercise this guy.


packages/tpetra/core/src/Tpetra_Details_residual.hpp
```diff
@@ -0,0 +1,294 @@
...
+              const MultiVector<SC,LO,GO,NO> & B_in,
+              MultiVector<SC,LO,GO,NO> & R_in) {
+  using Tpetra::Details::ProfilingRegion;
+  using Teuchos::RCP;
+  using Teuchos::rcp;
+  using Teuchos::rcp_const_cast;
+  using Teuchos::rcpFromRef;
+  
+  const CrsMatrix<SC,LO,GO,NO> * Apt  = dynamic_cast<const CrsMatrix<SC,LO,GO,NO>*>(&Aop);
+  if(!Apt) {
```
mhoemmen: You could just call the `CrsMatrix` overload here.  That means you'll only have to optimize one implementation.

packages/xpetra/sup/Matrix/Xpetra_Matrix.hpp
```diff
@@ -599,6 +599,11 @@ namespace Xpetra {
 #endif
     // ----------------------------------------------------------------------------------
 
+    //! Compute a residual R = B - (*this) * X
+    virtual void residual(const MultiVector< Scalar, LocalOrdinal, GlobalOrdinal, Node > & X,
+                          const MultiVector< Scalar, LocalOrdinal, GlobalOrdinal, Node > & B,
+                          MultiVector< Scalar, LocalOrdinal, GlobalOrdinal, Node > & R) const = 0;
+
     protected:
       Teuchos::Hashtable<viewLabel_t, RCP<MatrixView> > operatorViewTable_; // hashtable storing the operator views (keys = view names, values = views).
```
mhoemmen: Oh wow, we really need to get rid of this in favor of `std::unordered_map` (a C++11 feature).

packages/tpetra/core/src/Tpetra_Details_residual.hpp
```diff
@@ -0,0 +1,294 @@
...
+              const MultiVector<SC,LO,GO,NO> & X_in,
+              const MultiVector<SC,LO,GO,NO> & B_in,
+              MultiVector<SC,LO,GO,NO> & R_in) {
+  using Tpetra::Details::ProfilingRegion;
+  using Teuchos::RCP;
+  using Teuchos::rcp;
+  using Teuchos::rcp_const_cast;
+  using Teuchos::rcpFromRef;
+  
+  const CrsMatrix<SC,LO,GO,NO> * Apt  = dynamic_cast<const CrsMatrix<SC,LO,GO,NO>*>(&Aop);
```
mhoemmen: I would prefer the opposite code organization:

  * The overload (above) that takes CrsMatrix should have the optimized implementation.
  * The overload here that takes Operator should first attempt the `dynamic_cast`, if it succeeds, it should just invoke the CrsMatrix overload; otherwise, it should have a fall-back implementation that works for Operator.

This organization would mean that you only need to implement the optimized version once.

packages/xpetra/sup/Matrix/Xpetra_Matrix.hpp
csiefer2: To add to the Xpetra backlog :)

packages/tpetra/core/src/Tpetra_Details_residual.hpp
csiefer2: Not entirely following --- I only have one optimized implementation (OK, it isn't optimized, now, but it will be).

mhoemmen: Here's what I mean:

1. Overload of `residual` that takes `CrsMatrix` does the optimized thing.
2. Overload of `residual` that takes `Operator` first attempts the `dynamic_cast` to CrsMatrix.  If the cast succeeds, then it should just call the overload of `residual` that takes `CrsMatrix`.  If not, then it should do the generic Operator thing.

srajama1:
@csiefer2 What is the performance benefit that comes out of this ?

csiefer2:
> @csiefer2 What is the performance benefit that comes out of this ?

This PR?  Nothing.  This is prep work.

Once we get the kernel, this will cut the number of kernel launches in the smoothers & AMG hierarchy.

csiefer2:
The tests failed... because they all passed.  Wonderful.  Time for a retest @trilinos/framework

prwolfe:
> The tests failed... because they all passed. Wonderful. Time for a retest @trilinos/framework

The intel 17 build reported

"ERROR: Error fetching remote repo 'origin'
Archiving artifacts
Finished: FAILURE"

This is the same issue I fixed earlier today.

PR #6174, opened 2019-10-28
6141no getter functions for output interval and output indices in time step control

asteyer:
Tempus:  This closes issue 6141, no getter functions for output interval and output indices in time step control.  Adds getter functions for output interval and output indices and a unit test demonstrating their correctness.


ccober6:
@asteyer I fixed a warning in TEST_COMPARE.  It needed "==" instead of "=".

PR #6190, opened 2019-10-29
Tpetra/Xpetra: fix add kernels

### LINKED ISSUE #6149: MueLu, Xpetra, Tpetra:  Error in Add() with DEPRECATED code disabled

brian-kelley:
Xpetra: use fast (KokkosKernels) version of Tpetra::MatrixMatrix::add(). This fixes
StaticProfile rows overflowing (#6149). 

Tpetra::MatrixMatrix::add() now respects "Call fillComplete" =
false in the input params. Doing fillComplete is still the default.

Fix the Tpetra add() result column map
(result is always locally indexed, even if not fillComplete).
The colmap now has the correct index base, correct min/max GID bounds,
and puts remote entries after the local entries (so that all the locally owned LIDs are contiguous).

@trilinos/tpetra 
@trilinos/xpetra 

## Motivation

This PR does three things:
- Fixes ``Tpetra::MatrixMatrix::add()`` for several situations that led to an incorrect column map (for example, if the A and/or B colmap had a nonzero index base).
- Removes the slow code path in add() for non-contiguous domain maps (this was really the wrong check for avoiding the previous problem). Now this code path isn't needed.
- Made ``Xpetra::TwoMatrixAdd`` use this fast version (add) instead of Add. add uses KokkosKernels to do the actual addition, whereas Add runs in serial and uses getGlobalRowCopy() and insertGlobalEntries(). KokkosKernels sparse addition also computes the space required for C exactly, so this fixes #6149 and also means Xpetra no longer needs to estimate the nnz/row. 

## Related Issues

* Closes #6149
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #6152
* Part of 
* Composed of


packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+Teuchos::RCP<Map<LocalOrdinal, GlobalOrdinal, Node> > AddKernels<Scalar, LocalOrdinal, GlobalOrdinal, Node>::
+makeColMapAndConvertGids(GlobalOrdinal indexBase, GlobalOrdinal minCol, GlobalOrdinal maxCol,
+                   const typename MMdetails::AddKernels<Scalar, LocalOrdinal, GlobalOrdinal, Node>::global_col_inds_array& gids,
+                   typename MMdetails::AddKernels<Scalar, LocalOrdinal, GlobalOrdinal, Node>::col_inds_array& lids,
+                   const Teuchos::RCP<const Teuchos::Comm<int>>& comm)
+{
+  using Teuchos::RCP;
+  using Teuchos::rcp;
+  int myRank = comm->getRank();
+  std::cout << "Rank " << myRank << ": column map for some subset of global cols: " << minCol << " <= col <= " << maxCol << '\n';
```
lucbv: This should go before the PR gets merged in

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_decl.hpp
```diff
@@ -648,6 +542,105 @@ void setMaxNumEntriesPerRow(
 template<class CrsMatrixType>
 size_t C_estimate_nnz(CrsMatrixType & A, CrsMatrixType &B);
 
+template<typename Scalar, typename LocalOrdinal, typename GlobalOrdinal, typename Node>
+struct AddKernels
+{
+  typedef Tpetra::CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> crs_matrix_type;
```
mhoemmen: This is a copy-and-paste so don't worry for now, but prefer C++11 alias syntax `using new_type = old_type;` to `typedef new_type old_type;` for future code.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_decl.hpp
```diff
@@ -648,6 +542,105 @@ void setMaxNumEntriesPerRow(
...
+    const values_array& Bvals,
+    const row_ptrs_array_const& Browptrs,
+    const col_inds_array& Bcolinds,
+    const impl_scalar_type scalarB,
+    GlobalOrdinal numGlobalCols,
+    values_array& Cvals,
+    row_ptrs_array& Crowptrs,
+    col_inds_array& Ccolinds);
+
+  static Teuchos::RCP<map_type> makeColMapAndConvertGids(GlobalOrdinal indexBase, GlobalOrdinal minCol, GlobalOrdinal maxCol, const global_col_inds_array& gids, col_inds_array& lids, const Teuchos::RCP<const Teuchos::Comm<int>>& comm);
```
mhoemmen: `Tpetra::Map` is immutable, so if you return it by pointer, the pointer should always be a pointer-to-const (e.g., `RCP<const map_type>`).

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -622,11 +551,16 @@ add (const Scalar& alpha,
 {
   typedef Map<LocalOrdinal,GlobalOrdinal,Node>                     map_type;
   typedef CrsMatrix<Scalar,LocalOrdinal,GlobalOrdinal,Node>  crs_matrix_type;
+  //Check that A,B are fillComplete before getting B's column map
+  TEUCHOS_TEST_FOR_EXCEPTION
+    (! A.isFillComplete () || ! B.isFillComplete (), std::invalid_argument,
+     "TpetraExt::MatrixMatrix::add(): A and B must both be fill complete.");
   Teuchos::RCP<const map_type> CrowMap = transposeB ?  B.getDomainMap() : B.getRowMap();
```
mhoemmen: It's called the "row Map," and it's used as a row Map below, but it's either a domain or range Map.  Is that OK?

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -622,11 +551,16 @@ add (const Scalar& alpha,
...
-
   Teuchos::RCP<crs_matrix_type> C = rcp(new crs_matrix_type(CrowMap, 0));
-
-  add(alpha,transposeA,A,beta,transposeB,B,*C,domainMap,rangeMap,params);
+  //this version of add() always fill completes the result
+  if(!params.is_null())
```
mhoemmen: You shouldn't need to say `<bool>` here.  The following should suffice, since it can deduce the template argument from the method's second argument here.
```c++
params->set("Call fillComplete", true);
```

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  GlobalOrdinal nentries = gids.extent(0);
+  typedef Kokkos::Bitset<device_type> bitset_t;
+  typedef Kokkos::ConstBitset<device_type> const_bitset_t;
+  //Determine the set of GIDs in the column map using a dense bitset (covers entire possible range)
+  bitset_t presentGIDs(maxCol - minCol + 1);
+  Kokkos::parallel_for(range_type(0, nentries), ColMapFunctors::GatherPresentEntries<GlobalOrdinal, GView, bitset_t>(minCol, gids, presentGIDs));
+  execution_space().fence();
+  const_bitset_t constPresentGIDs(presentGIDs);
+  GView mapGIDs("Sum colmap GIDs", constPresentGIDs.count());
+  std::cout << "Rank " << myRank << ": determined there are " << mapGIDs.extent(0) << " unique global cols on this proc.\n";
```
lucbv: Same as above

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -837,9 +756,9 @@ add (const Scalar& alpha,
     MM = rcp(new TimeMonitor(*TimeMonitor::getNewTimer(prefix_mmm + std::string("building optimized column map"))));
 #endif
     CrowMap = Bprime->getRowMap();
-    //OLD: CcolMap = AddKern::makeColMapAndConvertGids(Aprime->getGlobalNumCols(), globalColinds, colinds, comm);
-    //Get C's column map as the union of Aprime and Bprime col maps
-    CcolMap = AddKern::makeColMapAndConvertGids(Aprime->getGlobalNumCols(), globalColinds, colinds, comm);
+    colinds = col_inds_array("C colinds", globalColinds.extent(0));
```
mhoemmen: It looks like this View is an output argument of the function call below, so consider not initializing it here:
```c++
using Kokkos::view_alloc;
colinds = col_inds_array(view_alloc("C colinds", WithoutInitializing), globalColinds.extent(0));
```

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -890,42 +809,43 @@ add (const Scalar& alpha,
   MM = Teuchos::null;
   MM = rcp(new TimeMonitor(*TimeMonitor::getNewTimer(prefix_mmm + std::string("Tpetra::Crs constructor"))));
 #endif
-  //      C = rcp(new crs_matrix_type(CrowMap, CcolMap, rowptrs, colinds, vals, params));
   C.replaceColMap(CcolMap);
```
mhoemmen: It's a bit goofy that we create an empty C, then call a bunch of methods on it. CrsMatrix has a constructor that takes everything a fill-complete matrix needs, including all four Maps, the Import and Export, and all the local data.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  };
+
+  template<typename GO, typename GView, typename const_bitset_t>
+  struct ListGIDs
+  {
+    ListGIDs(GO minGID_, GView& gidList_, const_bitset_t& present_)
+      : minGID(minGID_), gidList(gidList_), present(present_)
+    {}
+
+    KOKKOS_INLINE_FUNCTION void operator()(const GO i, GO& lcount, const bool& finalPass) const
```
mhoemmen: Pass `bool` input parameters by value, not by reference.  This code is correct, but you're taking up `sizeof(void*)` on the stack rather than `sizeof(bool)`.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
mhoemmen: This `std::cout` will print on all processes in the MPI communicator.  We'll need to take this out or protect it behind `Tpetra::Details::Behavior::verbose()`.  There are other unprotected `std::cout` uses in this PR.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  typedef typename Map<LocalOrdinal, GlobalOrdinal, Node>::local_map_type LocalMap;
+  GlobalOrdinal nentries = gids.extent(0);
+  typedef Kokkos::Bitset<device_type> bitset_t;
+  typedef Kokkos::ConstBitset<device_type> const_bitset_t;
+  //Determine the set of GIDs in the column map using a dense bitset (covers entire possible range)
+  bitset_t presentGIDs(maxCol - minCol + 1);
+  Kokkos::parallel_for(range_type(0, nentries), ColMapFunctors::GatherPresentEntries<GlobalOrdinal, GView, bitset_t>(minCol, gids, presentGIDs));
+  execution_space().fence();
+  const_bitset_t constPresentGIDs(presentGIDs);
+  GView mapGIDs("Sum colmap GIDs", constPresentGIDs.count());
```
mhoemmen: Create uninitialized if appropriate (see above).

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  typedef typename MMdetails::AddKernels<Scalar, LocalOrdinal, GlobalOrdinal, Node>::global_col_inds_array::non_const_type GView;
+  typedef typename MMdetails::AddKernels<Scalar, LocalOrdinal, GlobalOrdinal, Node>::col_inds_array::non_const_type LView;
+  typedef typename Map<LocalOrdinal, GlobalOrdinal, Node>::local_map_type LocalMap;
+  GlobalOrdinal nentries = gids.extent(0);
+  typedef Kokkos::Bitset<device_type> bitset_t;
+  typedef Kokkos::ConstBitset<device_type> const_bitset_t;
+  //Determine the set of GIDs in the column map using a dense bitset (covers entire possible range)
+  bitset_t presentGIDs(maxCol - minCol + 1);
+  Kokkos::parallel_for(range_type(0, nentries), ColMapFunctors::GatherPresentEntries<GlobalOrdinal, GView, bitset_t>(minCol, gids, presentGIDs));
+  execution_space().fence();
```
mhoemmen: The only reason you might need the fence is because of the `std::cout` uses, so once you take those out, you can take out the fences too.

packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
```diff
@@ -561,68 +561,19 @@ Note: this class is not in the Xpetra_UseShortNames.hpp
...
   #ifdef HAVE_XPETRA_TPETRA
           const Tpetra::CrsMatrix<SC,LO,GO,NO>& tpA = Xpetra::Helpers<SC,LO,GO,NO>::Op2TpetraCrs(A);
           const Tpetra::CrsMatrix<SC,LO,GO,NO>& tpB = Xpetra::Helpers<SC,LO,GO,NO>::Op2TpetraCrs(B);
           RCP<Tpetra::CrsMatrix<SC,LO,GO,NO> >  tpC = Xpetra::Helpers<SC,LO,GO,NO>::Op2NonConstTpetraCrs(C);
-
-          Tpetra::MatrixMatrix::Add(tpA, transposeA, alpha, tpB, transposeB, beta, tpC);
+          auto addParams = rcp(new Teuchos::ParameterList);
+          addParams->set<bool>("Call fillComplete", false);
+          //passing A after B means C will have the same domain/range map as A (or A^T if transposeA)
+          Tpetra::MatrixMatrix::add<SC,LO,GO,NO>(beta, transposeB, tpB, alpha, transposeA, tpA, *tpC, Teuchos::null, Teuchos::null, addParams);
```
cgcgcg: There are two more specializations of `TwoMatrixAdd` that also need to be changed..

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  std::cout << "Rank " << myRank << ": column map for some subset of global cols: " << minCol << " <= col <= " << maxCol << '\n';
+  typedef typename MMdetails::AddKernels<Scalar, LocalOrdinal, GlobalOrdinal, Node>::global_col_inds_array::non_const_type GView;
+  typedef typename MMdetails::AddKernels<Scalar, LocalOrdinal, GlobalOrdinal, Node>::col_inds_array::non_const_type LView;
+  typedef typename Map<LocalOrdinal, GlobalOrdinal, Node>::local_map_type LocalMap;
+  GlobalOrdinal nentries = gids.extent(0);
+  typedef Kokkos::Bitset<device_type> bitset_t;
+  typedef Kokkos::ConstBitset<device_type> const_bitset_t;
+  //Determine the set of GIDs in the column map using a dense bitset (covers entire possible range)
+  bitset_t presentGIDs(maxCol - minCol + 1);
+  Kokkos::parallel_for(range_type(0, nentries), ColMapFunctors::GatherPresentEntries<GlobalOrdinal, GView, bitset_t>(minCol, gids, presentGIDs));
```
mhoemmen: It would be better to lift out these Kokkos parallel loops into separate free functions.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  execution_space().fence();
+  const_bitset_t constPresentGIDs(presentGIDs);
+  GView mapGIDs("Sum colmap GIDs", constPresentGIDs.count());
+  std::cout << "Rank " << myRank << ": determined there are " << mapGIDs.extent(0) << " unique global cols on this proc.\n";
+  Kokkos::parallel_scan(range_type(0, maxCol - minCol + 1), ColMapFunctors::ListGIDs<GlobalOrdinal, GView, const_bitset_t>(minCol, mapGIDs, constPresentGIDs));
+  execution_space().fence();
```
mhoemmen: You probably need the fence here because Map's constructor needs to access the data on host, but perhaps we should move the fence into Map's constructor.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  //Construct the map
+  auto colMap = rcp(new Map<LocalOrdinal, GlobalOrdinal, Node>(
+        Teuchos::OrdinalTraits<Tpetra::global_size_t>::invalid(), //have Tpetra compute global #GIDs
+        mapGIDs, indexBase, comm));
+  //Get the device-friendly local map
+  LocalMap localColMap = colMap->getLocalMap();
+  //Use it to convert gids to lids
+  Kokkos::parallel_for(range_type(0, nentries), ColMapFunctors::ConvertGlobalToLocal<GlobalOrdinal, GView, LView, LocalMap>(gids, lids, localColMap));
+  execution_space().fence();
```
mhoemmen: We shouldn't sprinkle fences around without understanding why -- I don't think we need it here.  See above comment about adding a fence to Map's constructor.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::col_inds_array& Bcolinds,
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::impl_scalar_type scalarB,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::values_array& Cvals,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::row_ptrs_array& Crowptrs,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::col_inds_array& Ccolinds)
+{
+  using Teuchos::TimeMonitor;
+  TEUCHOS_TEST_FOR_EXCEPTION(Arowptrs.extent(0) != Browptrs.extent(0), std::runtime_error, "Can't add matrices with different numbers of rows.");
+  auto nrows = Arowptrs.extent(0) - 1;
+  Crowptrs = row_ptrs_array("C row ptrs", nrows + 1);
```
mhoemmen: Do we need to initialize this View?

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::row_ptrs_array_const& Browptrs,
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::col_inds_array& Bcolinds,
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::impl_scalar_type scalarB,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::values_array& Cvals,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::row_ptrs_array& Crowptrs,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::col_inds_array& Ccolinds)
+{
+  using Teuchos::TimeMonitor;
+  TEUCHOS_TEST_FOR_EXCEPTION(Arowptrs.extent(0) != Browptrs.extent(0), std::runtime_error, "Can't add matrices with different numbers of rows.");
+  auto nrows = Arowptrs.extent(0) - 1;
```
mhoemmen: What if `Arowptrs.extent(0)` is zero?  We allow that case for a matrix with zero rows.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::impl_scalar_type scalarB,
+  GO /* numGlobalCols */,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::values_array& Cvals,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::row_ptrs_array& Crowptrs,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::col_inds_array& Ccolinds)
+{
+  using Teuchos::TimeMonitor;
+  TEUCHOS_TEST_FOR_EXCEPTION(Arowptrs.extent(0) != Browptrs.extent(0), std::runtime_error, "Can't add matrices with different numbers of rows.");
+  auto nrows = Arowptrs.extent(0) - 1;
+  Crowptrs = row_ptrs_array("C row ptrs", nrows + 1);
```
mhoemmen: See above.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::col_inds_array& Bcolinds,
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::impl_scalar_type scalarB,
+  GO /* numGlobalCols */,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::values_array& Cvals,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::row_ptrs_array& Crowptrs,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::col_inds_array& Ccolinds)
+{
+  using Teuchos::TimeMonitor;
+  TEUCHOS_TEST_FOR_EXCEPTION(Arowptrs.extent(0) != Browptrs.extent(0), std::runtime_error, "Can't add matrices with different numbers of rows.");
+  auto nrows = Arowptrs.extent(0) - 1;
```
mhoemmen: See above.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+                         const LocalIndicesType& colindsOrig_,
+                         const GlobalIndicesType& colindsConverted_,
+                         const ColMapType& colmap_) :
+    minGlobal (minGlobal_),
+    colindsOrig (colindsOrig_),
+    colindsConverted (colindsConverted_),
+    colmap (colmap_)
+  {}
+  KOKKOS_INLINE_FUNCTION void
+  operator() (const size_t& i) const
```
mhoemmen: 1. Prefer signed integer loop indices.
2. Pass the loop index by value, not by reference.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3504,6 +3084,282 @@ merge_matrices(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview
...
+  const typename MMdetails::AddKernels<SC, LO, GO, NO>::local_map_type& BcolMap,
+  GO minGlobalCol,
+  GO /* numGlobalCols */,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::values_array& Cvals,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::row_ptrs_array& Crowptrs,
+  typename MMdetails::AddKernels<SC, LO, GO, NO>::global_col_inds_array& Ccolinds)
+{
+  using Teuchos::TimeMonitor;
+
+  const values_array& Avals = A.values;
```
mhoemmen: Be careful handling Views by reference, since somebody might insert code below that captures them in a device kernel.  Prefer handling them by value.

packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
brian-kelley: @cgcgcg Those versions of Xpetra::TwoMatrixAdd() can take a non fill-complete B, and they add values into B (rather than making a new C). Tpetra::MatrixMatrix::add() doesn't support this yet.

cgcgcg: How about these guys?
https://github.com/trilinos/Trilinos/blob/75333e4ed9915f29cfcd916dc01ec877e81b60fc/packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp#L1254-L1256
https://github.com/trilinos/Trilinos/blob/75333e4ed9915f29cfcd916dc01ec877e81b60fc/packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp#L1870-L1872

brian-kelley: Yep you're right.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
brian-kelley: colMap is fully constructed by now, but this parallel_for (ConvertGlobalToLocal) populates lids, which is the other output of this function. Shouldn't we make sure it can be immediately accessed on device by any caller? In theory, if the parallel_for took a long time, having lids be partially computed could cause problems in the expertStaticFillComplete.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
brian-kelley: I made the edits locally using the other syntax ``Kokkos::ViewAllocateWithoutInitializing("C colidnds")`` but then I scrolled up and saw this. Is ``view_alloc(..., WithoutInitializing)`` the preferred syntax?

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
mhoemmen: @brian-kelley It populates LIDs on device, so any subsequent device function on the same CUDA stream should be able to access the LIDs without an intervening fence.  Device kernels on the same stream are sequenced.

mhoemmen: Trilinos developers would like to remove both the `CUDA_LAUNCH_BLOCKING` and UVM assumptions at some point.  Right now, the fences are just extra work, because Trilinos requires `CUDA_LAUNCH_BLOCKING` to be set.  Removing the `CUDA_LAUNCH_BLOCKING` requirement will take lots of work to find existing issues.  I'd rather not hide any existing correctness issues by doing unnecessary fences.  It's like doing extra `MPI_Barrier`s to hide MPI correctness issues -- it doesn't necessarily make the code correct.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
brian-kelley: Good catch! Neither B's domain map nor col map are correct for C row map (if transposeB). I ended up copying the explicit transpose of B into this function, since the RowMapTransposer gives Bprime the correct rowmap.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_decl.hpp
```diff
@@ -638,7 +638,7 @@ struct AddKernels
     row_ptrs_array& Crowptrs,
     col_inds_array& Ccolinds);
 
-  static Teuchos::RCP<const map_type> makeColMapAndConvertGids(GlobalOrdinal indexBase, GlobalOrdinal minCol, GlobalOrdinal maxCol, const global_col_inds_array& gids, col_inds_array& lids, const Teuchos::RCP<const Teuchos::Comm<int>>& comm);
+  static Teuchos::RCP<const map_type> makeColMapAndConvertGids(GlobalOrdinal minCol, GlobalOrdinal maxCol, const global_col_inds_array& gids, col_inds_array& lids, const Teuchos::RCP<const Teuchos::Comm<int>>& comm);
```
mhoemmen: Ah, yes, I missed that one.  Thanks for fixing this!

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -1617,6 +1617,66 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(Tpetra_MatMat, add_zero_rows, SC, LO, GO, NT)
...
+{
+  using Teuchos::RCP;
+  using Teuchos::Array;
+  using crs_matrix_type = Tpetra::CrsMatrix<SC, LO, GO, NT>;
+  using map_type = Tpetra::Map<LO, GO, NT>;
+  RCP<const Teuchos::Comm<int> > comm = Tpetra::getDefaultComm();
+  LO nlocal = 5;
+  GO nrows = nlocal * comm->getSize();
+  GO indexBase = 4;
+  RCP<map_type> rowMap = rcp(new map_type(nrows, 0, comm));
```
mhoemmen: Prefer `RCP<const map_type>` to emphasize immutability of Maps.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -1558,7 +1558,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(Tpetra_MatMat, threaded_add_sorted, SC, LO, GO
   //now run the threaded addition on mats[0] and mats[1]
   ISC zero(0);
   ISC one(1);
-  Tpetra::MatrixMatrix::AddDetails::AddKernels<SC, LO, GO, NT>::addSorted(valsCRS[0], rowptrsCRS[0], colindsCRS[0], one, valsCRS[1], rowptrsCRS[1], colindsCRS[1], one, valsCRS[2], rowptrsCRS[2], colindsCRS[2]);
+  Tpetra::MMdetails::AddKernels<SC, LO, GO, NT>::addSorted(valsCRS[0], rowptrsCRS[0], colindsCRS[0], one, valsCRS[1], rowptrsCRS[1], colindsCRS[1], one, valsCRS[2], rowptrsCRS[2], colindsCRS[2]);
```
mhoemmen: We should just have a function that adds `KokkosSparse::CrsMatrix`.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -620,13 +549,30 @@ add (const Scalar& alpha,
...
+    RowMatrixTransposer<Scalar, LocalOrdinal, GlobalOrdinal, Node> transposer(Brcp);
+    RCP<ParameterList> transposeParams(new ParameterList);
+    transposeParams->set ("sort", false);
+    Brcp = transposer.createTranspose(transposeParams);
+  }
+  //Check that A,B are fillComplete before getting B's column map
+  TEUCHOS_TEST_FOR_EXCEPTION
+    (! A.isFillComplete () || ! Brcp->isFillComplete (), std::invalid_argument,
+     "TpetraExt::MatrixMatrix::add(): A and B must both be fill complete.");
+  Teuchos::RCP<crs_matrix_type> C = rcp(new crs_matrix_type(Brcp->getRowMap(), 0));
```
mhoemmen: You've already done `using Teuchos::RCP;` above, so you don't need to write `Teuchos::RCP` here.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -620,13 +549,30 @@ add (const Scalar& alpha,
...
+  //this version of add() always fill completes the result, no matter what is in params on input
+  if(!params.is_null())
+    params->set("Call fillComplete", true);
+  add(alpha, transposeA, A, beta, false, *Brcp, *C, domainMap, rangeMap, params);
```
mhoemmen: I'm guessing the point is that the changes to `add` below make this call work even if C is StaticProfile, as it would be with deprecated code in Tpetra disabled.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
mhoemmen: My comment still stands, but we can fix that later.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -3123,28 +3123,40 @@ namespace ColMapFunctors
...
-      }
-      if(present.test(i))
+      bool isRemote = localDomainMap.getLocalElement(i + minGID) == ::Tpetra::Details::OrdinalTraits<LO>::invalid();
+      if(present.test(i) && doingRemotes == isRemote)
       {
+        if(finalPass)
+        {
+          if(doingRemotes)
+          {
+            //remotes are inserted at the end
```
mhoemmen: You might actually want to use the existing Tpetra function `makeColMap` to make the column Map, rather than trying to reimplement it (but not correctly) here.  On the other hand, note that `makeColMap` needs to do MPI communication by default.  It would be great if you already had that data before entering your function.

brian-kelley: @mhoemmen What is the issue you see with this col map construction? I would be fine with using makeColMap, but if I go through every row in serial on host and do insertGlobalEntries() with the global colinds and values I have, the speed benefit from using the KokkosKernels spadd is mostly negated. Is there a way to construct a globally indexed, packed 1D storage CrsGraph? Then I would be done. I wouldn't even have to make a column map then if fill complete wasn't requested (which it isn't in Xpetra::TwoMatrixAdd, which seems to be the most common user of this add()).

brian-kelley:
@mhoemmen OK, I understand now that the col map remotes need to be grouped by proc ID. I would like to use makeColMap. I propose that we add new versions of ``setAllValues()``/``setAllIndices()`` that take a ``CrsGraph::t_GlobalOrdinal_1D`` (1D view of GO) instead of ``local_graph_type::entries_type`` (1D view of LO). It would be the same as the current ``setAllIndices()`` but it would set indicesAreLocal_ = false, and wouldn't construct the local graph.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
mhoemmen: @brian-kelley wrote:
> Do the remote GIDs need to be in a certain order?

When you let the CrsGraph or CrsMatrix construct its own column Map at fillComplete, then the remote GIDs on each process are sorted first by owning process rank in the domain Map.  The GID order affects the LIDs.  Some tests might just check the LIDs and/or column Map for equality directly.  I'm not saying your algorithm is incorrect, just that you might be producing a different column Map in a way that makes the tests grumpy.

> Is there a way to construct a globally indexed, packed 1D storage CrsGraph?

Do you mean "give the CrsGraph's constructor all the global indices at once, as packed 1-D storage"?  Not currently; CrsGraph doesn't have that constructor.

brian-kelley: @mhoemmen Thank you! That function would solve half the problem. The other half is the case where we don't have a domain/range map, and ``add()`` produces a non-fill complete C. Could I make a version ``makeCrsGraphFromGlobalIndices`` that just takes a ``StaticCrsGraph<GO>`` and row map? Could these just be static functions inside CrsGraph?

mhoemmen: @brian-kelley wrote:
> Could these just be static functions inside CrsGraph?

[C++ Core Guidelines C.4](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-member): "Make a function a member only if it needs direct access to the representation of a class."

> add is going to fillComplete and then resumeFill to let the user sum into the existing values - I think that's the best I can do.

What's the intent of "add and return a non-fill-complete matrix"?  Would users be surprised that the resulting matrix is locally indexed and has a column Map?

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -834,24 +798,37 @@ add (const Scalar& alpha,
...
+    C.replaceColMap(CcolMap);
+    col_inds_array localColinds("C colinds", globalColinds.extent(0));
+    Kokkos::parallel_for(Kokkos::RangePolicy<exec_space>(0, globalColinds.extent(0)),
+        ConvertGlobalToLocalFunctor<LocalOrdinal, GlobalOrdinal,
+                              col_inds_array, global_col_inds_array,
+                              typename map_type::local_map_type>
+        (localColinds, globalColinds, CcolMap->getLocalMap()));
+    C.setAllValues(rowptrs, localColinds, vals);
+    //Use a regular fillComplete so that the values and indices get sorted.
+    C.fillComplete(CDomainMap, CRangeMap, params);
```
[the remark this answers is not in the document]
brian-kelley: @mhoemmen There's no other reason than that's how it was before :) The other add() just returns a new matrix; that's the only interface that's actually necessary.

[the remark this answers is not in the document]
mhoemmen: @brian-kelley cool, sorry, not your fault :-) There may have been an intent at one point for add to have the ability to reuse existing storage in the output matrix.  That might explain why it was an output argument instead of a return value.

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
```diff
@@ -56,12 +56,272 @@
...
 #include <vector>
 
 namespace Tpetra {
 namespace Details {
 
+template <class LO, class GO, class NT>
+int
+makeColMapImpl (Teuchos::RCP<const Tpetra::Map<LO, GO, NT> >& colMap,
+            Teuchos::Array<int>& remotePIDs,
+            const Teuchos::RCP<const Tpetra::Map<LO, GO, NT> >& domMap,
```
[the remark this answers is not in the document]
brian-kelley: @mhoemmen Right,  this was just a copy-paste.

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
```diff
@@ -56,12 +56,272 @@
...
+#include "Kokkos_Bitset.hpp"
 #include <set>
 #include <vector>
 
 namespace Tpetra {
 namespace Details {
 
+template <class LO, class GO, class NT>
+int
+makeColMapImpl (Teuchos::RCP<const Tpetra::Map<LO, GO, NT> >& colMap,
```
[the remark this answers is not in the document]
brian-kelley: @mhoemmen Yes to the first, no to the second - it's only an output. If colMap is non-null on input to ``makeColMap()`` then it will be clobbered with the new map.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -554,6 +554,19 @@ add (const Scalar& alpha,
   using Teuchos::rcp;
   using Teuchos::ParameterList;
   typedef CrsMatrix<Scalar,LocalOrdinal,GlobalOrdinal,Node>  crs_matrix_type;
+  if(!params.is_null())
+  {
+    if(params->isParameter("Call fillComplete") && !params->get<bool>("Call fillComplete"))
+    {
+      if(A.getRowMap()->getComm()->getRank() == 0)
+      {
+        std::cout << "WARNING: Tpetra::MatrixMatrix::add(): 'Call fillComplete' was explicitly set to false,\n";
```
mhoemmen: I'm not a fan of printing these messages, since this use case is likely driven by MueLu, not by applications (whose stdout will now be filled with warning messages that they will report as a bug to Trilinos).  The `add` function should do the right thing, whatever that means.  Printing that the behavior changed won't help users.

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
```diff
@@ -532,7 +532,7 @@ struct ListGIDs
       }
       lcount++;
     }
-    if(i == (GO) present.size() - 1 && finalPass)
+    if((i == (GO) present.size() - 1) && finalPass)
```
mhoemmen: Nice catch!  Also, prefer C++ casts to C-style casts.  I would recommend the following, since it will work even for those rare users that have enabled unsigned GO:
```c++
if((i == static_cast<GO>(present.size() - 1)) && finalPass)
```

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
mhoemmen: Just to clarify -- if changing the behavior doesn't break anything, I'm OK with changing the behavior.  The issue I have is that applications normally would have no control over the behavior of sparse matrix-matrix multiply.  If you don't intend to support this case, I would rather that you threw an exception with an informative message.

brian-kelley: @mhoemmen OK, I will throw an exception then. Nothing in Trilinos sets "Call fillComplete" = false and then calls this version (which they don't and shouldn't). But for the apps, just like we have warnings for explicitly-enabled things in CMake getting disabled automatically, I don't want this to just continue silently.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -834,24 +798,37 @@ add (const Scalar& alpha,
...
+    col_inds_array localColinds("C colinds", globalColinds.extent(0));
+    Kokkos::parallel_for(Kokkos::RangePolicy<exec_space>(0, globalColinds.extent(0)),
+        ConvertGlobalToLocalFunctor<LocalOrdinal, GlobalOrdinal,
+                              col_inds_array, global_col_inds_array,
+                              typename map_type::local_map_type>
+        (localColinds, globalColinds, CcolMap->getLocalMap()));
+    C.setAllValues(rowptrs, localColinds, vals);
+    //Use a regular fillComplete so that the values and indices get sorted.
+    C.fillComplete(CDomainMap, CRangeMap, params);
+    if(!doFillComplete)
```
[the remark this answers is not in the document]
brian-kelley: @mhoemmen I talked with @csiefer2 about this and he said that if every test in Trilinos passes (esp. Xpetra, MueLu and Teko which actually use these kernels directly) then it will be OK. The function comments don't make a promise either way about the result state.

brian-kelley:
@mhoemmen I addressed most of your things (and marked those as resolved). There is more refactoring that can be done but for now I think we should merge this for EMPIRE.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -558,12 +558,8 @@ add (const Scalar& alpha,
   {
     if(params->isParameter("Call fillComplete") && !params->get<bool>("Call fillComplete"))
     {
-      if(A.getRowMap()->getComm()->getRank() == 0)
-      {
-        std::cout << "WARNING: Tpetra::MatrixMatrix::add(): 'Call fillComplete' was explicitly set to false,\n";
-        std::cout << "but this version of add() always calls fillComplete() on the sum (C). Use the other version\n";
-        std::cout << "of add() (that works on a newly constructed C) to get a non-fillComplete matrix.\n";
-      }
+      throw std::invalid_argument("Tpetra::MatrixMatrix::add(): this version of add() always calls fillComplete\n"
```
mhoemmen: "parameter list" (spelling).  Also, prefer the `TEUCHOS_TEST_FOR_EXCEPTION` macro as a point at which we can inject a breakpoint in case this exception should come up.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
mhoemmen: @brian-kelley Thanks Brian -- I don't mean to be a blocker; I just want to make sure that we can quickly find any issues with the semantics change.  I doubt there will be, but it's likely that we don't have good downstream testing of all the possibilities.

rppawlo:
@brian-kelley it looks like this PR has broken empire. This is for trilinos with deprecated code enabled. Info is in #6264

PR #6191, opened 2019-10-29
Tpetra: Implementing residual kernel

csiefer2:
Auto-PR for SHA ed2c660

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_residual.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_residual.hpp
@@ -58,6 +58,79 @@
 namespace Tpetra {
   namespace Details {
 
+
+template<class ExecutionSpace>
+int64_t
+residual_launch_parameters (int64_t numRows,
+                            int64_t nnz,
+                            int64_t rows_per_thread,
+                            int& team_size,
+                            int& vector_length)
+{
+  using execution_space = typename ExecutionSpace::execution_space;
+
+  int64_t rows_per_team;
+  int64_t nnz_per_row = nnz/numRows;
+
+  if (nnz_per_row < 1) {
+    nnz_per_row = 1;
+  }
+
+  if (vector_length < 1) {
+    vector_length = 1;
+    while (vector_length<32 && vector_length*6 < nnz_per_row) {
+      vector_length *= 2;
+    }
+  }
+
+  // Determine rows per thread
+  if (rows_per_thread < 1) {
+#ifdef KOKKOS_ENABLE_CUDA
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_Details_residual.hpp
```diff
@@ -58,6 +58,79 @@
 namespace Tpetra {
   namespace Details {
 
+
+template<class ExecutionSpace>
+int64_t
+residual_launch_parameters (int64_t numRows,
```
mhoemmen: I'm just as guilty of copy-pasting this magic, but it would still be nice if we had it in one place.

packages/tpetra/core/src/Tpetra_Details_residual.hpp
```diff
@@ -121,11 +195,102 @@ void localResidual(const CrsMatrix<SC,LO,GO,NO> &  A,
        std::runtime_error, "X, Y and R may not alias one another.");
   }
       
+#ifdef TPETRA_DETAILS_USE_REFERENCE_RESIDUAL
```
mhoemmen: Why do we need this macro here?  Is the point to do performance testing?

packages/tpetra/core/src/Tpetra_Details_residual.hpp
```diff
@@ -121,11 +195,102 @@ void localResidual(const CrsMatrix<SC,LO,GO,NO> &  A,
...
+  }
+
+  bool is_vector = (X_lcl.extent(1) == 1);
+
+  if(is_vector) {
+    // Vector case
+    // Kernel interior shamelessly horked from Ifpack2_Details_ScaledDampedResidual_def.hpp
+    Kokkos::parallel_for("residual-vector",policy,KOKKOS_LAMBDA(const team_member& dev) {
+        // NOTE: It looks like I should be able to get this data up above, but if I try to
+        // we get internal compiler errors.  Who knew that gcc tried to "gimplify"?
```
mhoemmen: Sometimes the internal compiler errors go away if you capture inside by value, instead of by reference.  Also, always pass the loop index by (const) value, not by const reference.

mhoemmen: On the other hand, if you capture inside by value, you might have bigger kernels that take up more GPU registers etc.

csiefer2: I know you're supposed to pass PODs by const value.  team_member should be treated the same way?

mhoemmen: `team_member` is a struct, so it's probably better to pass by const reference.

PR #6196, opened 2019-10-29
MueLu: adding region matrix unit-test

lucbv:
@trilinos/muelu 

## Motivation
The region multigrid algorithms are not currently protected against regression with unit-tests, this should fix that problem.
This add infrastructure to test the region multigrid work without modifying the current set of tests in a non-experimental build.
One new test is currently added and checks that the construction of a region matrix from a composite matrix is performed correctly.

## Stakeholder Feedback
This is internal development of experimental code, no external stakeholder

## Testing
This new test has been run locally and passes with 1 and 4 MPI ranks


lucbv:
@rstumin 
I have updated the unit-test to use a more general approach to create the region matrix.
This will make it easier to tests multiple matrix sizes, the reference solution still needs to be hard coded.
I fixed the issue with the relative path for the header inclusion so the code is a bit cleaner.

Issue #6200, opened 2019-10-30
Tpetra: GCC 5.3.0 internal compiler error


brian-kelley:
## Bug Report
@trilinos/tpetra 
@csiefer2 

### Description
GCC 5.3.0 (on trappist, rocketman and enigma) has an internal compiler error in 
Tpetra_Details_residual.hpp, line 216, even though the line looks very innocuous.
```
/home/nightlyTesting/OPENMPI-1.10.1_RELEASE_PERFORMANCE/Trilinos/packages/tpetra/coreTpetra_Details_residual.hpp:216:48: internal compiler error: in tsubst_copy, at cp/pt.c:13151
     residual_launch_parameters<execution_space>(A_lcl.numRows (), A_lcl.nnz (), rows_per_thread, team_size, vector_length);
                                                ^
0x65f242 tsubst_copy
	../../.cp/pt.c:13149
0x65f878 tsubst_copy_and_build(tree_node*, tree_node*, int, tree_node*, bool, bool)
	../../.cp/pt.c:15747
...etc
```

I think this started with commit ed2c6605. This affects most of the "Experimental" CDash builds on several different machines so hopefully it's easy to reproduce and find a workaround.

### Steps to Reproduce
1. SHA1:  1f4c2d is the one showing in the failing dashboard builds, but probably ed2c6605 or later will have it.
2. Build with GCC 5.3, it doesn't seem like the exact configuration or machine details matter.
3. Shake fist angrily at compiler.

csiefer2:
@mhoemmen Any ideas?

rppawlo:
Are you using nested lambdas? Gcc 5 and 6 have a nasty bug with nested lambdas.

csiefer2:
I am using nested lambdas, but the compiler woofs long before that.

mhoemmen:
@csiefer2 Sometimes it helps to make code simpler -- e.g., make a temporary variable for `A_lcl.numRows()` etc.

csiefer2:
@mhoemmen Since I'm mucking with it anyway, do you mind if I yank the launch parameters function out of Ifpack2's scaledDampedResidual and just have it call this?

csiefer2:
Buh?

Tpetra_Details_residual.hpp:219:48: error: no matching function for call to ‘residual_launch_parameters(const int64_t&, const int64_t&, const int64_t&, const int&, const int&)’
[...]

PR #6205, opened 2019-10-31
Tpetra: Because GCC 5.3.0 hates valid C++

csiefer2:
Issue #6200; Testing: n/a

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_residual.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_residual.hpp
@@ -211,9 +211,12 @@ void localResidual(const CrsMatrix<SC,LO,GO,NO> &  A,
   int team_size = -1;
   int vector_length = -1;
   int64_t rows_per_thread = -1;
+  
+  int64_t numLocalRows = A_lcl.numRows ();
+  int64_t myNnz = A_lcl.nnz();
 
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_Details_residual.hpp
```diff
@@ -211,9 +211,12 @@ void localResidual(const CrsMatrix<SC,LO,GO,NO> &  A,
   int team_size = -1;
   int vector_length = -1;
   int64_t rows_per_thread = -1;
+  
+  int64_t numLocalRows = A_lcl.numRows ();
+  int64_t myNnz = A_lcl.nnz();
 
-  const int64_t rows_per_team = 
-    residual_launch_parameters<execution_space>(A_lcl.numRows (), A_lcl.nnz (), rows_per_thread, team_size, vector_length);
+  int64_t rows_per_team =
```
mhoemmen: I have some vague recollection that `const` is related to this compiler bug somehow.

rppawlo: the nested lambda issue was fixed by removing const from any captured variables.

csiefer2:
@trilinos/framework Not posting to either testing or testing-dev

PR #6211, opened 2019-11-01
Tpetra: Sample MultiVector test using Serial and Cuda on different MPI ranks

csiefer2:
Testing: new MultiVector_MixedNode test

### DIFF
--- a/packages/tpetra/core/test/MultiVector/MultiVector_MixedNode.cpp
+++ b/packages/tpetra/core/test/MultiVector/MultiVector_MixedNode.cpp
@@ -1,0 +1,263 @@
+/*
+// @HEADER
+// ***********************************************************************
+//
+//          Tpetra: Templated Linear Algebra Services Package
+//                 Copyright (2008) Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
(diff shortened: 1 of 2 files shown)


brian-kelley:
@csiefer2 I think you need to add MultiVector_MixedNode.cpp

csiefer2:
Fixed

packages/tpetra/core/test/MultiVector/MultiVector_MixedNode.cpp
```diff
@@ -0,0 +1,263 @@
...
+        "Test MPI (if available) or force test of serial.  In a serial build,"
+        " this option is ignored and a serial comm is always used." );
+    clp.setOption(
+        "error-tol-slack", &errorTolSlack,
+        "Slack off of machine epsilon used to check test results" );
+  }
+
+  // no ScalarTraits<>::eps() for integer types
+  template <class Scalar>
+  typename Teuchos::ScalarTraits<Scalar>::magnitudeType testingTol() { return Teuchos::ScalarTraits<Scalar>::eps(); }
```
mhoemmen: You don't use these in the file, so they can go away.  That also makes the `TRILINOS_UNUSED_FUNCTION` logic unnecessary.

PR #6212, opened 2019-11-01
Intrepid2: Reduce view allocs and inits

### LINKED ISSUE #6172: Intrepid2, Remove wasteful initialization
## Enhancement
@trilinos/intrepid2

This type of code: https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp#L251

```c++
      // Create ordinalToTag
      ordinalToTag = OrdinalTypeView2D("ordinalToTag", basisCard, tagSize);
```

wastes a CUDA kernel launch because Kokkos views are initialized to 0 by default, and then the view is set to -1 in the next line: https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp#L254. 

```c++
      // Initialize with -1
      Kokkos::deep_copy( ordinalToTag, -1 );
```

These types of allocations should use `Kokkos::WithoutInitializing`

```c++
      ordinalToTag = OrdinalTypeView2D(Kokkos::view_alloc("ordinalToTag",Kokkos::WithoutInitializing), basisCard, tagSize);
```

stanmoore1:
@trilinos/intrepid2 

## Motivation
Reduce view allocs and inits in Intrepid2

## Related Issues
Closes #6172 

## Stakeholder Feedback
EMPIRE ATDM

## Testing

Tested with EMPIRE on GPUs

### DIFF
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
@@ -211,6 +211,7 @@ namespace Intrepid2 {
         \li     ordinalToTag_[DofOrd][3] = total number of DoFs associated with the subcell
     */
     OrdinalTypeArray2DHost ordinalToTag_;
+    OrdinalTypeArray2D ordinalToTag_device;
 
     /** \brief  DoF tag to ordinal lookup table.
 
@@ -224,6 +225,7 @@ namespace Intrepid2 {
         \li     tagToOrdinal_[subcDim][subcOrd][subcDofOrd] = Degree-of-freedom ordinal
     */
     OrdinalTypeArray3DHost tagToOrdinal_;
+    OrdinalTypeArray3D tagToOrdinal_device;
 
     /** \brief  Fills <var>ordinalToTag_</var> and <var>tagToOrdinal_</var> by basis-specific tag data
 
@@ -248,7 +250,7 @@ namespace Intrepid2 {
                             const ordinal_type       posScOrd,
                             const ordinal_type       posDfOrd ) {
       // Create ordinalToTag
-      ordinalToTag = OrdinalTypeView2D("ordinalToTag", basisCard, tagSize);
+      ordinalToTag = OrdinalTypeView2D(Kokkos::view_alloc("ordinalToTag",Kokkos::WithoutInitializing), basisCard, tagSize);
 
       // Initialize with -1
       Kokkos::deep_copy( ordinalToTag, -1 );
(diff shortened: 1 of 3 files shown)


stanmoore1:
@bathmatt

mperego:
Looks good to me. Thanks!
@kyungjoo-kim is that OK with you?

kyungjoo-kim:
This PR itself is fine and might be some improvement. A real problem is that this fix does not fix the root cause of the empire performance problem. By removing the data trasfer of small tag view (roughly the size O(10)), it is observed that 50% time improvement. This means that actual kernel time is very very short and the time cost of the function is almost equivalent to time for data trasfer of O(10). In my opinion, this overhead becomes dominent because the actual workload of the kernel function is too light weight. Although this PR itself is improvement for a very rare use case where cuda kernel workload is very very light weight, it is not the intended use case. Can we hold this a bit till someone (nate) look at the code and check how intrepid2 is used ?

stanmoore1:
>this overhead becomes dominent because the actual workload of the kernel function is too light weight

I think it is still bad form to waste CUDA kernel launches on unnecessary initialization.

stanmoore1:
I think this is sort of urgent, every time @bathmatt runs the code he pays a large overhead for initialization, reducing his turnaround time of trying things out.

kyungjoo-kim:
In  #6172, I commented about the potential problem of the empire code. I agree that removing unnecessary view initialization in the code is certainly improvement. However, if the performance problem is rooted from the empire code (misuse of the orientation function), then the member object "tag_device" introduced in this PR to avoid repeated view constructions in orientation and projection tools is unnecessary. I would avoid unnecessary complications in the basis object.

stanmoore1:
OK feel free to close this PR and submit your own.

stanmoore1:
@kyungjoo-kim you can also modify as you see fit, the "Allow edits from maintainers" is checked.

CamelliaDPG:
I've looked at the EMPIRE code; it is indeed invoking Intrepid2 in a suboptimal way, but it's going to take some work to refactor.  My current thinking is that for optimal performance in EMPIRE, we will need to offer a somewhat different interface in Intrepid2's OrientationTools.  I think the changes by @stanmoore1 are a good fix for the time being; they have the advantage of actually being ready to go.  So I suggest merging this, @mperego and @kyungjoo-kim.

kyungjoo-kim:
@stanmoore1 Would you email me the building script for trilinos and empire ? I can help fix the empire code so that it does call the intrepid2 functions in a right way.

CamelliaDPG:
@kyungjoo-kim, I'm already working on this in EMPIRE.  The particular use case (building an incidence matrix) is not well-supported by Intrepid2.  My current approach involves a small modification to the Intrepid2 functor, allowing the input to have no cell dimension, and the addition of an "expert" version of `modifyBasisByOrientation` that supplies the device copies of `ordinalToTag` and `tagToOrdinal`.

kyungjoo-kim:
@CamelliaDPG I am not sure how you are going to fix the interface problem. We can disbuss offline. FYI: Intrepid2 interface is designed to be a batch interface with workset arrays. This part is not going to be changed as almost all sandia application follows this logic and it does not significantly hurt if the workset size is appropriately determined. Intrepid2 also has Impl level (functor level) interface. This interface is only exposed upon users' requests but it does not provide runtime polymophism as the impl function is used on devices. The main problem in the empire code is that parallel-for is not used but a serial cell-for loop uses the batch interface with a single workset.

CamelliaDPG:
@kyungjoo-kim, yes, I'm aware of those things, and I'm refactoring the way that EMPIRE does things as well.  I'll send you an email with details, and when it comes time to do the PR for Intrepid2, I'll include you as a reviewer.

kyungjoo-kim:
I put WIP not to bother PR test system. @CamelliaDPG is working on the empire code and re-assess this PR afterwards.

CamelliaDPG:
This has been addressed by PR #6323, and changes to the calling application, so I'm closing this PR.

Issue #6163, opened 2019-10-25
Panzer/Tpetra:  Export from ghosted graph to empty graph reports insufficient capacity 


kddevin:
## Bug Report
@trilinos/tpetra @trilinos/panzer @rppawlo @tjfulle @bathmatt 
User's issue-tracking number 882
Resolution needed before merging tpetra-remove-deprecated branch.

### Description
Although all panzer tests pass with Tpetra_ENABLE_DEPRECATED_CODE=OFF, a panzer user is experiencing problems.  The user sees
```
p=0 | /home//Trilinos/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp:2299:
p=0 | 
p=0 | Throw number = 1
p=0 | 
p=0 | Throw test that evaluated to true: (numInserted == Teuchos::OrdinalTraits<size_t>::invalid())
p=0 | 
p=0 | Tpetra::CrsGraph<int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> >::insertGlobalIndicesImpl: There is not enough capacity to insert indices in to row 71. The upper bound on the number of entries in this row must be increased to accommodate one or more of the new indices.
p=0 | 

```
The stack trace is
```
#0  __cxxabiv1::__cxa_throw (obj=0xa2ba9b0, 
    tinfo=0x8e6c290 <typeinfo for std::runtime_error>, 
    dest=0x459d10 <_ZNSt13runtime_errorD1Ev@plt>)
    at ../../.././libstdc++-v3/libsupc++/eh_throw.cc:75
#1  0x0000000003b981de in Tpetra::CrsGraph<int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> >::insertGlobalIndicesImpl(Tpetra::RowInfo const&, long long const*, unsigned long, std::function<void (unsigned long, unsigned long, unsigned long)>) ()
#2  0x0000000003b9825d in Tpetra::CrsGraph<int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> >::insertGlobalIndicesImpl(int, long long const*, unsigned long) ()
[...]

PR #6236, opened 2019-11-06
Tpetra: possible fix for #6163

tjfulle:
Use std::map to temporarily store padding needed to resize CrsGraph and then pass to Kokkos::UnorderedMap.
Allows accumulation of new entries, since Kokkos::UnorderedMap does not allow reinsertion.

Broken Albany test shows as passed.

@trilinos/tpetra 

## Motivation
#6163

## Testing
Broken Albany test passes

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -2294,8 +2294,12 @@ namespace Tpetra {
       TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(
           numInserted == Teuchos::OrdinalTraits<size_t>::invalid(),
           std::runtime_error,
-          "There is not enough capacity to insert indices in to row " << lclRow <<
-          ". The upper bound on the number of entries in this row must be increased to "
+          "There is not enough capacity to insert " << inputInds.size()  << " " <<
+          "indices in to row " << lclRow <<
+          " on rank " << this->getComm()->getRank() << "." <<
+          "The current size of the indices array is " << this->k_gblInds1D_.size() << ".  " <<
+          "The current size of the row pointers array is " << this->k_rowPtrs_.size() << ".  " <<
+          "The upper bound on the number of entries in this row must be increased to "
           "accommodate one or more of the new indices.");
       this->k_numRowEntries_(lclRow) += numInserted;
       this->setLocallyModified();
@@ -5871,6 +5875,9 @@ namespace Tpetra {
     using GO = GlobalOrdinal;
     const char tfecfFuncName[] = "computeCrsPaddingForSameIds: ";
 
+    if (!numSameIDs)
+      return;
+
     Kokkos::fence ();
 
     using insert_result =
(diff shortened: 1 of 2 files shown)


packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -6009,12 +6013,20 @@ namespace Tpetra {
     // without unpacking the import/export buffer, we don't know how many of the
     // numPacketsPerLID[i] LIDs exist in the target. Below, it is assumed that
     // none do, and padding is requested for all.
-    for (size_t i = 0; i < numEnt; ++i) {
-      auto result = padding.insert (importLIDs_h[i], numPacketsPerLID_h[i]);
+    //
+    // Use tmp_padding since Kokkos::UnorderedMap does not allow re-insertion
+    std::map<local_ordinal_type, size_t> tmp_padding;
```
tjfulle: @mhoemmen - this is the only functional change to the code (the rest being some indentation and extra comments to debug/failure messages).

The case that Albany was exercising had repeated indices in the `importLIDs`, since `Kokkos::UnorderedMap` does not allow re-insertion, use `std::map` to properly accumulate the extra capacity needed and then pass to `Kokkos::UnorderedMap`

tjfulle:
@mperego this PR fixes the broken Albany test on my side
@rppawlo can you test this branch to see if it fixes your issues?

mperego:
> @mperego this PR fixes the broken Albany test on my side
> @rppawlo can you test this branch to see if it fixes your issues?

@tjfulle Yes, it seems it fixes all the Albany tests that were failing. Thanks!!

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5871,6 +5875,9 @@ namespace Tpetra {
     using GO = GlobalOrdinal;
     const char tfecfFuncName[] = "computeCrsPaddingForSameIds: ";
 
+    if (!numSameIDs)
```
mhoemmen: Prefer `numSameIDs == 0`.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
mhoemmen: I'm pretty sure that `Kokkos::UnorderedMap` allows reinsertion.  You just have to check the return value of `insert`; it should say whether the thing is in the map already.  On the other hand, if this works, I'll be happy to approve :-) .  We can always go back later and fix this by using `UnorderedMap::insert`'s return value.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -6009,12 +6013,20 @@ namespace Tpetra {
...
     // none do, and padding is requested for all.
-    for (size_t i = 0; i < numEnt; ++i) {
-      auto result = padding.insert (importLIDs_h[i], numPacketsPerLID_h[i]);
+    //
+    // Use tmp_padding since Kokkos::UnorderedMap does not allow re-insertion
+    std::map<local_ordinal_type, size_t> tmp_padding;
+    for (size_t i = 0; i < numEnt; ++i)
+      tmp_padding[importLIDs_h[i]] += numPacketsPerLID_h[i];
+
+    using padding_type = Kokkos::UnorderedMap<local_ordinal_type, size_t, device_type>;
```
mhoemmen: We're assuming UVM here.  That's fine for now, but if UnorderedMap's constructor initializes its data on device, then we might need to insert a `Kokkos::fence()` after `padding`'s constructor.  Otherwise, this code's correctness may depend on whether `CUDA_LAUNCH_BLOCKING` is set.

PR #6250, opened 2019-11-09
Update mapping of git users to remove duplicate and put in sandia.gov addresses

bartlettroscoe:
This gives a more accurate count of the number of commits being made by Sandia and non-Sandia developers.


bartlettroscoe:
Thanks @rppawlo!

PR #6251, opened 2019-11-10
Automatic snapshot commit from tribits at 83aecff

bartlettroscoe:
Main purpose is to bring in TriBITSPub/TriBITS#287 so that @jmgate  and EMPIRE can use this. 

This also brings in some other minor TriBITS changes but they don't impact Trilinos at this point.


bartlettroscoe:
Thanks @rppawlo!

PR #6254, opened 2019-11-11
STK snapshot as of 11/11/2019

alanw0:
In particular, reduce the occurrence of warnings about internal
sidesets.

@trilinos/<teamName>


packages/stk/stk_io/stk_io/SidesetUpdater.cpp
```diff
@@ -85,9 +85,11 @@ stk::mesh::Entity get_side(const stk::mesh::BulkData& bulk, const stk::mesh::Sid
     return stk::mesh::Entity();
 }
 
-bool is_side_internal(const stk::mesh::BulkData& bulk, stk::mesh::Entity side,
+std::pair<bool,bool> is_side_internal_and_modified(const stk::mesh::BulkData& bulk, stk::mesh::Entity side,
```
mhoemmen: General advice from the C++ Standard Committee these days is to avoid pairs and tuples in favor of structs with named fields.  e.g.,
```c++
struct InternalOrModified {
  bool isInternal;
  bool isModified;
};

InternalOrModified is_side_internal_or_modified(...) {
  // etc. ....
  return {false, true}; // etc.
}
```

alanw0: Thanks @mhoemmen I like that too. Didn't realize it was general advice these days. I certainly don't like the idiom of using std::tie with returned pairs etc.

mhoemmen: @alanw0 Structured binding will make this easier -- you won't need to use `std::tie`.

packages/stk/stk_balance/stk_balance/balance.hpp
```diff
@@ -42,12 +42,27 @@ namespace stk { namespace mesh { class BulkData; } }
...
 
 namespace stk
 {
 namespace balance
 {
 
+class StkBalanceSettings : public GraphCreationSettings
+{
+public:
+    StkBalanceSettings()
```
mhoemmen: It suffices, and is better, to say `StkBalanceSettings() = default;`.  It suffices because defaulted or implicitly defined default constructors [automatically call their base class default constructors](https://en.cppreference.com/w/cpp/language/default_constructor).  It's better for two reasons:

1. It declares intent: "My class has no invariants other than the base class' invariants."  (Defining an explicit constructor marks the class as having invariants.)
2. If a class would otherwise be trivially default constructible, defining the constructor explicitly without `= default` would make the class not trivially default constructible.

You want to collect those "trivially `$X` constructible" attributes because they can help with compiler optimization.  It's probably less useful for this class because it looks like it has virtual methods, but it's a good habit (esp. trivially copy constructible, because that more or less says the compiler can copy the bits directly).

PR #6257, opened 2019-11-12
ATDM: Fix detection of sems-rhel6 and sems-rhel7 when SEMS_SYSTEM is not set (ATDV-228)

bartlettroscoe:
Turns out that you have to actually define the SEMS modules in order to hvae
SEMS_SYSTEM set.  Therefore, this adds back logic to look for the file:

  /projects/sems/modulefiles/utils/get-platform

This should fix the 'sems-rhel7' builds on ascicgpu14 and other such machines.

## How was this tested

I disabled the sourcing of:

```
. /projects/sems/modulefiles/utils/sems-modules-init.sh
```

in my `.bashrc` file and I was able to reproduce the problem where it did not recognize support for the system 'sems-rhel7' on ascicgpu14.  After this patch, it did.


bartlettroscoe:
Thanks @rppawlo!  Hopefully this got merged in time to fix what Adam B. was seeing today.

bartlettroscoe:
FYI: We are now seeing 'sems-rhel7' builds showing up again on CDash this morning:

* https://testing.sandia.gov/cdash/index.php?project=Trilinos&date=2019-11-12&filtercount=1&showfilters=1&field1=buildname&compare1=65&value1=Trilinos-atdm-

So this is confirmation this should be fixed.

PR #6261, opened 2019-11-12
Change EMPIRE Trilinos Configuration

jmgate:
Migrate two Trilinos configuration flags from EMPIRE's build scripts to those in Trilinos itself.  These flags have been included in EMPIRE's build scripts for some time, but if Trilinos developers are going to exactly replicate an EMPIRE build of Trilinos, these need to be in Trilinos instead.

### DIFF
--- a/cmake/std/atdm/apps/empire/EMPIRETrilinosEnables.cmake
+++ b/cmake/std/atdm/apps/empire/EMPIRETrilinosEnables.cmake
@@ -10,6 +10,9 @@ ATDM_SET_CACHE(Gtest_SKIP_INSTALL TRUE CACHE BOOL)
 # #5341.  (Once we can get SPARC to have its own gtest, then we can move this
 # setting to ATDMDevEnvSettings.cmake. See SPAR-614.)
 
+ATDM_SET_CACHE(Trilinos_ENABLE_Gtest FALSE CACHE BOOL)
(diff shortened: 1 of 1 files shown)


jmgate:
@bartlettroscoe, @fryeguy52, just wanted to make sure this is the right spot to put these configure flags.

cmake/std/atdm/apps/empire/EMPIRETrilinosEnables.cmake
```diff
@@ -10,6 +10,9 @@ ATDM_SET_CACHE(Gtest_SKIP_INSTALL TRUE CACHE BOOL)
 # #5341.  (Once we can get SPARC to have its own gtest, then we can move this
 # setting to ATDMDevEnvSettings.cmake. See SPAR-614.)
 
+ATDM_SET_CACHE(Trilinos_ENABLE_Gtest FALSE CACHE BOOL)
+ATDM_SET_CACHE(Kokkos_ENABLE_Profiling TRUE CACHE BOOL)
```
bartlettroscoe: FYI: You don't need to set this since profiling is on by default in Trilinos.  You can see this is set in the ATDM Trilinos builds already by looking [here](https://testing.sandia.gov/cdash/viewNotes.php?buildid=5888339#!#note3) for example showing:

```
Kokkos_ENABLE_Profiling:BOOL=ON
```

It is fine to leave it in, but it is not necessary.

PR #6266, opened 2019-11-13
Partial6190 revert

bathmatt:
@trilinos/<teamName>

### DIFF
--- a/packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
+++ b/packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
@@ -83,13 +83,13 @@
 namespace Xpetra {
   namespace Details
   {
+#ifdef HAVE_XPETRA_TPETRA
     template <typename tcrs_matrix_type>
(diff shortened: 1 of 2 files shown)


bathmatt:
This appears to fix the EMPIRE stuff.

DO we want to merge this?  Someone tag auto merge and review please if we do

rppawlo:
@brian-kelley - do you want to hold off on merging this pr until we see if the disabling the kokkos refactor flag in empire works? I took off the automerge for now.

brian-kelley:
@rppawlo @cgcgcg  I'm fine with merging this ASAP.

But, I'm definitely curious whether disabling MueLu kokkos refactor (but keeping #6190) also fixes EMPIRE. I think the problem is that somewhere, MueLu is assuming that a non-fill complete matrix is globally indexed. That's the only semantic change to ``Xpetra::MatrixMatrix::TwoMatrixAdd`` in #6190. But I don't think that's a safe assumption to make in general. Even if it is, this failure case isn't caught by MueLu's testing.

PR #6270, opened 2019-11-13
Ifpack2: Fuse all Chebyshev kernels into one

cgcgcg:
@trilinos/ifpack2 

## Motivation
Fuse ScaledDampedResidual and update kernels into one Chebyshev kernel. This is a blatant rip-off of @mhoemmen's work.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_def.hpp
@@ -1,0 +1,565 @@
+/*
+//@HEADER
+// ***********************************************************************
+//
+//       Ifpack2: Templated Object-Oriented Algebraic Preconditioner Package
+//                 Copyright (2009) Sandia Corporation
+//
+// Under terms of Contract DE-AC04-94AL85000, there is a non-exclusive
+// license for use of this work by or on behalf of the U.S. Government.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the Corporation nor the names of the
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL SANDIA CORPORATION OR THE
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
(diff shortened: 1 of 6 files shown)


packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_def.hpp
```diff
@@ -0,0 +1,511 @@
...
+
+template<class ExecutionSpace>
+int64_t
+chebyshev_kernel_vector_launch_parameters (int64_t numRows,
+                                                 int64_t nnz,
+                                                 int64_t rows_per_thread,
+                                                 int& team_size,
+                                                 int& vector_length)
+{
+  using execution_space = typename ExecutionSpace::execution_space;
```
mhoemmen: Later TODO: Figure out if these parameters are actually good.  Also, I would have preferred returning a struct with named fields, to the current "return a value and have two output parameters" approach.

packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_def.hpp
```diff
@@ -0,0 +1,511 @@
...
+  if (vector_length < 1) {
+    vector_length = 1;
+    while (vector_length<32 && vector_length*6 < nnz_per_row) {
+      vector_length *= 2;
+    }
+  }
+
+  // Determine rows per thread
+  if (rows_per_thread < 1) {
+#ifdef KOKKOS_ENABLE_CUDA
```
mhoemmen: We have way too much `#ifdef KOKKOS_ENABLE_CUDA` in Trilinos -- not sure how to generalize that here but just something to keep in mind.

packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_def.hpp
```diff
@@ -0,0 +1,511 @@
...
+  const char kernel_label[] = "chebyshev_kernel_vector";
+  policy_type policy (1, 1);
+  if (team_size < 0) {
+    policy = policy_type (worksets, Kokkos::AUTO, vector_length);
+  }
+  else {
+    policy = policy_type (worksets, team_size, vector_length);
+  }
+
+  // Canonicalize template arguments to avoid redundant instantiations.
```
mhoemmen: We should actually canonicalize them to their nonowning variants (`MemoryTraits<Unmanaged>`).

packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_def.hpp
```diff
@@ -0,0 +1,511 @@
...
+template<class TpetraOperatorType>
+void
+ChebyshevKernel<TpetraOperatorType>::
+setMatrix (const Teuchos::RCP<const operator_type>& A)
+{
+  if (A_op_.get () != A.get ()) {
+    A_op_ = A;
+
+    // We'll (re)allocate these on demand.
+    X_colMap_ = std::unique_ptr<vector_type> (nullptr);
```
mhoemmen: I wish there were a way to get at the column-Map MultiVector that already lives in CrsMatrix.

cgcgcg:
@jhux2 Both in Ifpack2 and in MueLu. (Hence all the failing tests in the previous iterations of this change.)

packages/ifpack2/src/Ifpack2_Parameters.cpp
```diff
@@ -64,6 +67,17 @@ void getValidParameters(Teuchos::ParameterList& params)
   params.set("fact: iluk level-of-fill", (int)1);
   params.set("fact: iluk level-of-overlap", (int)0);
 
+  // Ifpack2_Chebyshev
+  params.set("chebyshev: max eigenvalue",STS::nan());
+  params.set("chebyshev: ratio eigenvalue",STS::nan());
+  params.set("chebyshev: min eigenvalue",(double)30.0);
```
mhoemmen: 1. Avoid C-style casts; use C++ casts (either `double(30.0)` or `static_cast<double>(30.0)` -- I prefer the latter for integer casts, which may be a source of bugs).
2. The type of the literal `30.0` is `double`.

packages/ifpack2/src/Ifpack2_Parameters.cpp
```diff
@@ -64,6 +67,17 @@ void getValidParameters(Teuchos::ParameterList& params)
...
+  params.set("chebyshev: degree",(int)1);
+  params.set("chebyshev: eigenvalue max iterations",(int)10);
+  params.set("chebyshev: assume matrix does not change",false);
+  // params.set("chebyshev: operator inv diagonal",Teuchos::null);
+  params.set("chebyshev: min diagonal value",STS::eps());
```
mhoemmen: Is the "min diagonal value" a magnitude (as `eps()` is) or a value?  That is, if `Scalar` is complex, should the min diagonal value be complex?

packages/ifpack2/src/Ifpack2_Parameters.cpp
cgcgcg: Have a look at the rest of the file. Just trying to be consistent here. (But we should change a lot of that once we rework that, I agree.)

packages/ifpack2/src/Ifpack2_Parameters.cpp
cgcgcg: The file assumes `Scalar=double`, so that's ok here.

Issue #6264, opened 2019-11-12
MueLu/Amesos2/Tpetra: new Failures in empire


rppawlo:
## Bug Report
@trilinos/amesos2 
Starting last night, empire tests are failing in builds against trilinos develop.

The hanging tests are reporting:
```
Set 1 -> Set 1 -> Set 1 -> Set 1 -> 
...

*********** Caught Exception: Begin Error Report ***********
/scratch/mbetten/Trilinos/Trilinos/packages/amesos2/src/Amesos2_KLU2_def.hpp:277:

Throw number = 1

Throw test that evaluated to true: info > 0

KLU2 numeric factorization failed
 ```

The stack trace:
```
    at ../../.././libstdc++-v3/libsupc++/eh_throw.cc:75
#1  0x00007fffcf0b9b8c in Amesos2::KLU2<Tpetra::CrsMatrix<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> >, Tpetra::MultiVector<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> > >::numericFactorization_impl (
    this=0x5ba82b0)
    at /scratch/mbetten/Trilinos/Trilinos/packages/amesos2/src/Amesos2_KLU2_def.hpp:276
#2  0x00007fffcf12a43d in Amesos2::SolverCore<Amesos2::KLU2, Tpetra::CrsMatrix<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> >, Tpetra::MultiVector<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> > >::numericFactorization (this=0x5ba82b0)
    at /scratch/mbetten/Trilinos/Trilinos/packages/amesos2/src/Amesos2_SolverCore_def.hpp:153
[...]

PR #6277, opened 2019-11-14
Xpetra: Re-enabled fast TwoMatrixAdd path

brian-kelley:
DO NOT MERGE until we know that this doesn't break EMPIRE.

@trilinos/xpetra 
@trilinos/tpetra

## Motivation

This change re-enables the fast codepath in ``Xpetra::MatrixMatrix::TwoMatrixAdd()`` that was originally merged in #6190, but which was reverted in #6266 because it appeared to break EMPIRE (see #6264). But now it seems that #6190 just exposed an existing bug in MueLu kokkos refactor uncoupled aggregation. So @rppawlo when it is confirmed that this patch (reverting the revert) doesn't break EMPIRE, we can put it back in develop.

## Related Issues
* Closes 
* Blocks 
* Is blocked by 
* Follows #6190 (which had this enabled originally), #6266 (which reverted this code path) 
* Precedes 
* Related to #6264
* Part of 
* Composed of 

## Stakeholder Feedback

See comments on #6264 for the discussion about the new EMPIRE failures from 11-12-19.

### DIFF
--- a/packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
+++ b/packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
@@ -110,7 +110,6 @@ namespace Xpetra {
       if(transposeA)
         Aprime = transposer_type(Aprime).createTranspose();
       //Decide whether the fast code path can be taken.
-      /*
       if(A.isFillComplete() && B.isFillComplete())
       {
         RCP<tcrs_matrix_type> C = rcp(new tcrs_matrix_type(Aprime->getRowMap(), 0));
@@ -122,7 +121,6 @@ namespace Xpetra {
       }
       else
       {
-      */
         //Slow case - one or both operands are non-fill complete.
         //TODO: deprecate this.
         //Need to compute the explicit transpose before add if transposeA and/or transposeB.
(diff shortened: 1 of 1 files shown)


packages/xpetra/sup/Utils/Xpetra_MatrixMatrix.hpp
```diff
@@ -152,7 +150,7 @@ namespace Xpetra {
             *Bprime, false, beta,
             C);
         return rcp(new CrsWrap(rcp_implicit_cast<CrsType>(rcp(new XTCrsType(C)))));
```
mhoemmen: This line of code is so weird, but you didn't write it :-)

brian-kelley: @mhoemmen Nope, that's just how Xpetra works :)
I did look all over the place for a helper function that does this (Tpetra::CrsMatrix to Xpetra::Matrix) but I don't think there is one. The inverse exists though (probably because it has a bunch of dynamic_casts that need to be checked at each step).

jhux2: :octopus:

brian-kelley:
@rppawlo Do you think this is OK to merge? It looks like the actual problem in #6264 (the divide-by-zero thing in KokkosKernels) was exposed by my add() changes in #6190, but #6190 itself was OK. If you agree, I'll merge this in.

rppawlo:
@brian-kelley - Let me run a quick check on empire tests.

rppawlo:
empire is passing. good to merge. Thanks for waiting!

brian-kelley:
@rppawlo Thanks!

Issue #6276, opened 2019-11-14
ATDM Build Scripts:  Error in get_known_system_info.sh


jmgate:
## Bug Report
@bartlettroscoe, @fryeguy52 

### Description
`get_known_system_info.sh` was modified in c2c3e244ea95d3f4396b3b3220f45e97216b1890 and 5188171ea2d68f6f4f9b9fec92280a6fa6c9a16d.  I'm not sure which of those is the culprit, but now when I use `load-env.sh`, I get the following error:
```
/path/to/Trilinos/cmake/std/atdm/utils/get_known_system_info.sh: line 118: \
systemNameTypeMatchedListHostNames[${hostnameMatchSystemName}]: bad array subscript
```

For full details of the run that led to this failure, [see this log](https://jenkins-srn.sandia.gov/user/jmgate/my-views/view/jmgate/job/EMPIRE-jmgate-build-test-Trilinos-cortado-clang-opt/ws/log/Build_Trilinos.html).

### Steps to Reproduce
See [this Jenkins job](https://jenkins-srn.sandia.gov/user/jmgate/my-views/view/jmgate/job/EMPIRE-jmgate-build-test-Trilinos-cortado-clang-opt/).  I've also run into the same problem trying to configure on my local RHEL7 machine with gnu.

bartlettroscoe:
@jmgate, what version of Trilinos is this?  What exact command did you run to generate this error?

jmgate:
2fc3a14830572701739f9c44cd246d1e1c426dee

jmgate:
`/path/to/Trilinos/cmake/std/atdm/load-env.sh EMPIRE-jmgate-build-test-Trilinos-cortado-clang-opt`

jmgate:
I'm seeing this problem on cortado, my local RHEL7 machine, and also on cee-build012.  Scot Swan mentioned there were some changes that went in for HPCs recently, but perhaps they inadvertently hosed RHEL machines.

jmgate:
<img width="1664" alt="Screen Shot 2019-11-14 at 10 42 18 AM" src="https://user-images.githubusercontent.com/20327215/68881995-805e1a80-06cb-11ea-9d32-5f9a5227b44d.png">

I'm getting a return code of 127 when sourcing `load-env.sh` via a Python script.  When I run the command manually in a terminal, no such problem.

This problem did not exist when last I tried this on Tuesday.

jmgate:
[...]

PR #6280, opened 2019-11-14
Fix bug in atdm/utils/get_know_system_info.sh (#6276)

bartlettroscoe:
This should fix the bug reported in #6276 for systems where the ATDM system is not selected from the hostname (e.g. 'sems-rhel7').  It was just a bug on my part.

I also got rid of some unnecessary code so the should should be more uniform now and easier to maintain.

### DIFF
--- a/cmake/std/atdm/utils/get_known_system_info.sh
+++ b/cmake/std/atdm/utils/get_known_system_info.sh
@@ -115,6 +115,8 @@ fi
 #echo "hostnameMatch ='${hostnameMatch}'"
 
 if [[ "${hostnameMatch}" != "" ]] ; then
+  # A matching system by hostname becomes the first preferred match (but not
+  # the only possible match)
   systemNameTypeMatchedList+=(${hostnameMatchSystemName})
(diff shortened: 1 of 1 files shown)


cmake/std/atdm/utils/get_known_system_info.sh
```diff
@@ -44,7 +44,7 @@ realHostname=`hostname`
 ATDM_KNOWN_SYSTEM_NAMES_LIST=(
   shiller
   ride
-  mutrino   # Will be repalced by 'cts1'
+  mutrino   # Will be repalced by 'ats1'
```
mhoemmen: spelling: "replaced"

mhoemmen:
Thanks Ross! :-D

cmake/std/atdm/utils/get_known_system_info.sh
bartlettroscoe: Thanks!  It will be "replaced" when 'ats1' replaces 'mutrino' :-)

PR #6284, opened 2019-11-15
Tpetra: Fix #6282 (in CrsMatrix unpack functor, assume matrix rows sorted)

### LINKED ISSUE #6282: Tpetra::CrsMatrix::unpackAndCombine fails to exploit local matrix having sorted rows
## Bug Report
@trilinos/tpetra 

### Description

@vbrunini et al. discovered that CrsMatrix doExport was slow and suffered from load imbalance when some processes had dense or nearly dense rows to export.  One would expect some load imbalance for this case, but the comparable FEI / Epetra - based assembly was much faster.

I dug a little bit and found the following call chain:
  * `CrsMatrix::unpackAndCombine` calls `CrsMatrix::unpackAndCombineImpl`.
  * `CrsMatrix::unpackAndCombineImpl` calls `unpackCrsMatrixAndCombine(New)` if `isStaticGraph()` is `true`.
  * `unpackCrsMatrixAndCombine(New)` call `unpackAndCombineIntoCrsMatrix`.
  * `unpackAndCombineIntoCrsMatrix` takes a local matrix, and uses `UnpackCrsMatrixAndCombineFunctor`.
  * `UnpackCrsMatrixAndCombineFunctor` calls `KokkosSparse::CrsMatrix::sumIntoValues` with `isSorted=false` (meaning that `sumIntoValues` is _not_ allowed to assume that the matrix's row has sorted local column indices).

However, if we've gotten this far, then the local matrix _must_ have sorted local column indices.  First, `isStaticGraph()` being `true` means that the graph is fill complete; that means the graph is not globally indexed, and has sorted local column indices on each process.  ("Not locally indexed" and "not globally indexed" is a valid state, meaning "empty on that process.  Thank Epetra for that.)  Second, if CrsMatrix has a local matrix, then the local matrix must have sorted local column indices.  (`CrsMatrix::fillComplete` sorts and merges rows before calling `fillLocalGraphAndMatrix`.)

Failure to assume that the matrix has sorted rows, means that `sumIntoValues` must be doing linear search through the matrix row, for at least the first received entry to unpack and sum in.  If the received column indices and the matrix row's column indices match exactly, then the "hint" path in `sumIntoValues` should kick in for subsequent received entries.  However, the hint could fail if the index sequences differ -- e.g., if the source and target matrices have different column Maps.  In that case, we'll be doing linear search through a dense row, times the number of input entries (also dense) to unpack and combine into that row.  That's N^2 work!

The easy fix should be to change `UnpackCrsMatrixAndCombineFunctor` so it assumes the matrix has sorted rows.  This affects lines 381 and 386 of `Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp`.

mhoemmen:
@trilinos/tpetra @vbrunini 

In the Tpetra::CrsMatrix unpack functor, assume that the local matrix's rows are sorted, when calling sumIntoValues and replaceValues.

## Related Issues

* Closes #6282 

## Stakeholder Feedback

See Sierra Thermal Fluids issue TF-1264.

## Testing

This should be exercised by existing Tpetra::CrsMatrix unpack / doExport / doImport tests.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
@@ -375,15 +375,17 @@ struct UnpackCrsMatrixAndCombineFunctor {
     const LO* const lids_raw = const_cast<const LO*> (lids_out.data ());
     const ST* const vals_raw = const_cast<const ST*> (vals_out.data ());
     LO num_modified = 0;
+
+    constexpr bool matrix_has_sorted_rows = true; // see #6282
     if (combine_mode == ADD) {
       num_modified +=
         local_matrix.sumIntoValues (import_lid, lids_raw, num_ent,
-                                    vals_raw, false, atomic);
+                                    vals_raw, matrix_has_sorted_rows, atomic);
     }
     else if (combine_mode == REPLACE) {
       num_modified +=
         local_matrix.replaceValues (import_lid, lids_raw, num_ent,
-                                    vals_raw, false, atomic);
+                                    vals_raw, matrix_has_sorted_rows, atomic);
     }
     else {
       dst = Kokkos::make_pair (4, i); // invalid combine mode


mhoemmen:
Thanks @csiefer2 ! :-D

PR #6286, opened 2019-11-15
MueLu: adding region vector unit-test.

pohm01:
@trilinos/muelu 

## Motivation

The region multigrid algorithms are not currently protected against regression with unit-tests, this should fix that problem.
This add infrastructure to test the region multigrid work without modifying the current set of tests in a non-experimental build.
One new test is currently added and checks that conversion between region vectors and composite vectors is performed correctly.

## Stakeholder Feedback

This is internal development of experimental code, no external stakeholder.

## Testing

This new test has been run locally and passes with 1 and 4 MPI ranks.


packages/muelu/test/unit_tests/RegionVector.cpp
```diff
@@ -0,0 +1,381 @@
...
+                                              10.0, 11.0, 12.0, 13.0, 14.0,
+                                              15.0, 16.0, 17.0, 18.0, 19.0,
+                                              20.0, 21.0, 22.0, 23.0, 24.0}));
+      myValues = compVec->getData(0);
+      for(int idx = 0; idx < compVec->getLocalLength(); ++idx) {
+        TEST_FLOATING_EQUALITY(myValues[idx], refValues[idx], 100*TMT::eps());
+      }
+
+    } else if(numRanks == 4) {
+      // All ranks will have the same number of rows/cols/entries
```
mayrmt: This comment is misleading. The global length should be the same on all ranks, but the local length differs. Maybe replace it with something along the lines of:
```suggestion
      // All ranks will have the same global length of the vector, but local length differs.
      // Due to region numbering and region interface duplications,
      // lower rank IDs have more local entries then higher rank IDs.
```

packages/muelu/test/unit_tests/RegionVector.cpp
```diff
@@ -0,0 +1,381 @@
...
+      } else if(myRank == 1) {
+        refValues.deepCopy(ArrayView<const SC>({1.1, 2.1, 4.1, 5.1, 8.4, 10.4}));
+
+      } else if(myRank == 2) {
+        refValues.deepCopy(ArrayView<const SC>({3.2, 4.2, 8.5, 6.2, 7.2, 14.5}));
+
+      } else if(myRank == 3) {
+        refValues.deepCopy(ArrayView<const SC>({4.3, 5.3, 7.3, 8.3}));
+      }
+      // Loop over region matrix data and compare it to ref data
```
mayrmt:
```suggestion
      // Loop over region vector data and compare it to reference data
```

packages/muelu/test/unit_tests/RegionVector.cpp
```diff
@@ -0,0 +1,381 @@
...
+      } else if(myRank == 1) {
+        refValues.deepCopy(ArrayView<const SC>({2, 1, 1, 2, 1, 1, 4, 2, 2}));
+
+      } else if(myRank == 2) {
+        refValues.deepCopy(ArrayView<const SC>({2, 2, 4, 1, 1, 2, 1, 1, 2}));
+
+      } else if(myRank == 3) {
+        refValues.deepCopy(ArrayView<const SC>({4, 2, 2, 2, 1, 1, 2, 1, 1}));
+      }
+      // Loop over region matrix data and compare it to ref data
```
mayrmt:
```suggestion
      // Loop over region vector data and compare it to reference data
```

packages/muelu/test/unit_tests/RegionVector.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    compositeToRegional(compVec, quasiRegVec, regVec, maxRegPerProc, rowMapPerGrp, revisedRowMapPerGrp, rowImportPerGrp);
+
+    //RCP<Teuchos::FancyOStream> outF = Teuchos::fancyOStream(Teuchos::rcpFromRef(std::cout));
+    //std::cout<<"compVec:"<<std::endl;
+    //compVec->describe(*outF, Teuchos::VERB_EXTREME);
+    //std::cout<<"Quasi:"<<std::endl;
+    //quasiRegVec[0]->describe(*outF, Teuchos::VERB_EXTREME);
+    //std::cout<<"Reg:"<<std::endl;
+    //regVec[0]->describe(*outF, Teuchos::VERB_EXTREME);
```
mayrmt: Can this be removed?

mayrmt:
@pohm01 Thanks for setting up these tests. I made some very minor comments regarding documentation. Since @lucbv knows more about setting up this type of tests, I let him push the final approval button. :wink:

packages/muelu/test/unit_tests/RegionVector.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+  { // test compositeToRegional
+    // Create a composite vector
+    RCP<Vector> compVec = VectorFactory::Build(dofMap);
+    TEUCHOS_ASSERT(!compVec.is_null());
+    compVec->putScalar(0.0);
+    const size_t localLength = compVec->getLocalLength();
+    for (size_t k = 0; k < localLength; ++k) {
+      compVec->replaceLocalValue(k, k + myRank/10.0);// (LO, Scalar) need to fix type of k
```
lucbv: To fix type of k you only need to use `static_cast<some_new_type>(myOldTypeVar)`

mayrmt: @pohm Do you have time to take care of this? If not, please let me know. Then, I will look at it in detail.

packages/muelu/test/unit_tests/RegionVector.cpp
mayrmt: I just removed the commented debug output.

mayrmt:
@lucbv Requesting another review...

PR #6291, opened 2019-11-18
Teuchos: added StackedTimer option to print a timer only if time > dr…

### LINKED ISSUE #6289: Teuchos: Have stacked time summary have a minimum report time
## Enhancement
@trilinos/tempus 

I'm seeing a ton of 0 cost things in the timing log file, I'd like to have a minimum value for run time to count

```
|   |   runEvaluators: 362.898 - 17.9087% [6000] {min=198.397, max=629.832, std dev=140.738} <12, 19, 8, 3, 0, 5, 0, 0, 13, 4>
|   |   |   BC1 - AfterSolve: 0.00091644 - 0.000252534% [1000] {min=0.00077158, max=0.00107357, std dev=7.58366e-05} <3, 8, 5,\

```

SO, this is about 1e-6 and could be ignored.  So adding an arg to the dump command

rppawlo:
Closes #6289

### DIFF
--- a/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
+++ b/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
@@ -192,6 +192,8 @@ StackedTimer::computeColumnWidthsForAligment(std::string prefix,
   double total_time = 0.0;
 
   for (int i=0; i<flat_names_.size(); ++i ) {
+    if (sum_[i]/active_[i] <= options.drop_time)
+      continue;
     if (printed[i])
       continue;
     int level = std::count(flat_names_[i].begin(), flat_names_[i].end(), '@');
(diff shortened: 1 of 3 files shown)


packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -192,6 +192,8 @@ StackedTimer::computeColumnWidthsForAligment(std::string prefix,
   double total_time = 0.0;
 
   for (int i=0; i<flat_names_.size(); ++i ) {
+    if (sum_[i]/active_[i] < options.drop_time)
```
mhoemmen: Wouldn't you want `<= 0`, if the drop time is zero?

rppawlo: I wanted to make sure that all timers print if drop tolerance is zero. It should never happen in practice but... Happy to change this if you want. I could then make the default -1.0 to force all timers to print.

rppawlo: fixed.

rppawlo:
@mhoemmen could you review again?

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
mhoemmen: @rppawlo Thanks! :-D

PR #6304, opened 2019-11-19
Ifpack2: Adding a unit test to demo reduced matvec for use in s-step methods

csiefer2:
Auto-PR for SHA 0ba6888

### DIFF
--- a/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
+++ b/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
@@ -66,6 +66,7 @@
 #include <Galeri_XpetraMatrixTypes.hpp>
 #endif
 
+#include <Tpetra_Details_residual.hpp>
 
 #include <Ifpack2_UnitTestHelpers.hpp>
 #include <Ifpack2_OverlappingRowMatrix.hpp>
@@ -107,6 +108,145 @@ typedef Tpetra::global_size_t GST;
     return; \
   } \
 } while (false)
+
+
+
+
+/***********************************************************************************/
+template<class MatrixClass, class MultiVectorClass>
+void localReducedMatvec(const MatrixClass & A_lcl,
+                        const MultiVectorClass & X_lcl,
+                        const int userNumRows,
+                        MultiVectorClass & Y_lcl) {
+  using Teuchos::NO_TRANS;
+
+  using execution_space = typename MatrixClass::execution_space;
+
+  if (A_lcl.numRows() == 0 || userNumRows ==0 || userNumRows > A_lcl.numRows()) {
+    return;
+  }
+
+  int team_size = -1;
+  int vector_length = -1;
+  int64_t rows_per_thread = -1;
+  
+  int64_t numLocalRows = userNumRows;
+  int64_t myNnz = A_lcl.nnz();
+
+  int64_t rows_per_team = 
+    Tpetra::Details::residual_launch_parameters<execution_space>(numLocalRows, myNnz, rows_per_thread, team_size, vector_length);
+  int64_t worksets = (X_lcl.extent (0) + rows_per_team - 1) / rows_per_team;
+
+  using policy_type = typename Kokkos::TeamPolicy<execution_space>;
+  using team_member = typename policy_type::member_type;
+
+  using residual_value_type = typename MultiVectorClass::non_const_value_type;
+  using KAT = Kokkos::ArithTraits<residual_value_type>;
+  using LO = int64_t;
+
+  policy_type policy (1, 1);
+  if (team_size < 0) {
+    policy = policy_type (worksets, Kokkos::AUTO, vector_length);
+  }
+  else {
+    policy = policy_type (worksets, team_size, vector_length);
+  }
+
+  bool is_vector = (X_lcl.extent(1) == 1);
+
+  if(is_vector) {
+    // Vector case
+    // Kernel interior shamelessly horked from Ifpack2_Details_ScaledDampedResidual_def.hpp
+    Kokkos::parallel_for("reduced-mv-vector",policy,KOKKOS_LAMBDA(const team_member& dev) {
+        Kokkos::parallel_for(Kokkos::TeamThreadRange (dev, 0, rows_per_team),[&] (const LO& loop) {
+            const LO lclRow = static_cast<LO> (dev.league_rank ()) * rows_per_team + loop;
+            
+            if (lclRow >= numLocalRows) {
+              return;
+            }
+
+            const auto A_row = A_lcl.rowConst(lclRow);
+            const LO row_length = static_cast<LO> (A_row.length);
+            residual_value_type A_x = KAT::zero ();          
+            
+            Kokkos::parallel_reduce(Kokkos::ThreadVectorRange (dev, row_length), [&] (const LO iEntry, residual_value_type& lsum) {
+                const auto A_val = A_row.value(iEntry);
+                lsum += A_val * X_lcl(A_row.colidx(iEntry),0);
+              }, A_x);
+            Y_lcl(lclRow,0) = A_x;
+            
+          });//end parallel_for TeamThreadRange
+      });//end parallel_for "residual-vector"
+  } else {
+    // MultiVector case
+    // Kernel interior shamelessly horked from Ifpack2_Details_ScaledDampedResidual_def.hpp
+    Kokkos::parallel_for("reduced-mv-multivector",policy,KOKKOS_LAMBDA(const team_member& dev) {
+        // NOTE: It looks like I should be able to get this data up above, but if I try to
+        // we get internal compiler errors.  Who knew that gcc tried to "gimplify"?
+        const LO numVectors = static_cast<LO>(X_lcl.extent(1));
+        Kokkos::parallel_for(Kokkos::TeamThreadRange (dev, 0, rows_per_team),[&] (const LO& loop) {
+            const LO lclRow = static_cast<LO> (dev.league_rank ()) * rows_per_team + loop;
+            
+            if (lclRow >= numLocalRows) {
+              return;
+            }
+            const auto A_row = A_lcl.rowConst(lclRow);
+            const LO row_length = static_cast<LO> (A_row.length);
+            for(LO v=0; v<numVectors; v++) {
+              residual_value_type A_x = KAT::zero ();          
+              
+              Kokkos::parallel_reduce(Kokkos::ThreadVectorRange (dev, row_length), [&] (const LO iEntry, residual_value_type& lsum) {
+                  const auto A_val = A_row.value(iEntry);
+                  lsum += A_val * X_lcl(A_row.colidx(iEntry),v);
+                }, A_x);
+              Y_lcl(lclRow,v) = A_x;
+              
+            }//end for numVectors
+          });//end parallel_for TeamThreadRange
+      });//end parallel_for "residual-multivector"
+  }// end else                                     
+}// end reducedMatvec
+
+
+    
+/***********************************************************************************/
+template<class OverlappedMatrixClass, class MultiVectorClass>
+void reducedMatvec(const OverlappedMatrixClass & A,
+                   const MultiVectorClass & X,
+                   const int userExtNumRows,
+                   MultiVectorClass & Y) {
+  using crs_matrix_type = Tpetra::CrsMatrix<typename OverlappedMatrixClass::scalar_type,
+    typename OverlappedMatrixClass::local_ordinal_type,
+    typename OverlappedMatrixClass::global_ordinal_type,
+    typename OverlappedMatrixClass::node_type>;
+
+  // Assumes that X& Y are sufficiently overlapped for this to work
+  RCP<const crs_matrix_type> undA = Teuchos::rcp_dynamic_cast<const crs_matrix_type>(A.getUnderlyingMatrix());
+  RCP<const crs_matrix_type> extA = Teuchos::rcp_dynamic_cast<const crs_matrix_type>(A.getExtMatrix());
+
+  auto undA_lcl = undA->getLocalMatrix ();
+  auto extA_lcl = extA->getLocalMatrix ();
+  auto X_lcl = X.getLocalViewDevice ();
+  auto Y_lcl = Y.getLocalViewDevice ();
+  
+  // Do the "Local part"
+  auto numLocalRows = undA_lcl.numRows();
+  localReducedMatvec(undA_lcl,X_lcl,numLocalRows,Y_lcl);
+  
+  // Now, do the "overlapped part"
+  if(userExtNumRows > 0) {
+    auto X_ext = Kokkos::subview(X_lcl,std::make_pair(0,numLocalRows+userExtNumRows),Kokkos::ALL());
+    auto Y_ext = Kokkos::subview(X_lcl,std::make_pair(numLocalRows,numLocalRows+userExtNumRows),Kokkos::ALL());
+    
+    localReducedMatvec(extA_lcl,X_ext,userExtNumRows,Y_ext);
+  }
+
+}
+
+
+
+
+
 
 
 TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(Ifpack2OverlappingRowMatrix, Test0, Scalar, LO, GO)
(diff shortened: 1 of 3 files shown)


packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -66,6 +66,7 @@
 #include <Galeri_XpetraMatrixTypes.hpp>
 #endif
 
+#include <Tpetra_Details_residual.hpp>
```
mhoemmen: In the future, use double quotes for any header that's not a system header.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -109,6 +110,145 @@ typedef Tpetra::global_size_t GST;
...
+    policy = policy_type (worksets, team_size, vector_length);
+  }
+
+  bool is_vector = (X_lcl.extent(1) == 1);
+
+  if(is_vector) {
+    // Vector case
+    // Kernel interior shamelessly horked from Ifpack2_Details_ScaledDampedResidual_def.hpp
+    Kokkos::parallel_for("reduced-mv-vector",policy,KOKKOS_LAMBDA(const team_member& dev) {
+        Kokkos::parallel_for(Kokkos::TeamThreadRange (dev, 0, rows_per_team),[&] (const LO& loop) {
```
mhoemmen: If it's no bigger than `void*`, pass it by value.  (`LO` is `int` so it's definitely no bigger than `void*`.)

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -109,6 +110,145 @@ typedef Tpetra::global_size_t GST;
...
+            const LO lclRow = static_cast<LO> (dev.league_rank ()) * rows_per_team + loop;
+            
+            if (lclRow >= numLocalRows) {
+              return;
+            }
+
+            const auto A_row = A_lcl.rowConst(lclRow);
+            const LO row_length = static_cast<LO> (A_row.length);
```
mhoemmen: This is harmless, but the type of `A_row.length` should actually be `LO` here (unless we intend to support more general local matrix types).

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -109,6 +110,145 @@ typedef Tpetra::global_size_t GST;
 } while (false)
 
 
+
+
+/***********************************************************************************/
+template<class MatrixClass, class MultiVectorClass>
```
mhoemmen: It's not your fault, but it's annoying that `KokkosSparse::CrsMatrix` doesn't work like `Kokkos::View`, with a parameter pack `...`.  That would let us write `KokkosSparse::CrsMatrix<T, ...>` instead of this excessively general `MatrixClass`.  On the other hand, we could probably be more specific for `MultiVectorClass` here.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -109,6 +110,145 @@ typedef Tpetra::global_size_t GST;
...
+    typename OverlappedMatrixClass::global_ordinal_type,
+    typename OverlappedMatrixClass::node_type>;
+
+  // Assumes that X& Y are sufficiently overlapped for this to work
+  RCP<const crs_matrix_type> undA = Teuchos::rcp_dynamic_cast<const crs_matrix_type>(A.getUnderlyingMatrix());
+  RCP<const crs_matrix_type> extA = Teuchos::rcp_dynamic_cast<const crs_matrix_type>(A.getExtMatrix());
+
+  auto undA_lcl = undA->getLocalMatrix ();
+  auto extA_lcl = extA->getLocalMatrix ();
+  auto X_lcl = X.getLocalViewDevice ();
```
mhoemmen: We should make sure that X and Y are correctly sync'd -- otherwise, we'll be assuming UVM implicitly.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -416,9 +556,82 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(Ifpack2OverlappingRowMatrix, getLocalDiag, Sca
     TEST_EQUALITY(ldGids[i],ovrmGids[i]);
 }
 
+
+TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(Ifpack2OverlappingRowMatrix, reducedMatvec, Scalar, LocalOrdinal, GlobalOrdinal)
```
mhoemmen: It should be perfectly harmless to assume default `LocalOrdinal` and `GlobalOrdinal` here.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -416,9 +556,82 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(Ifpack2OverlappingRowMatrix, getLocalDiag, Sca
...
+    printf("Halo Starts:");
+    for(size_t i=0; i< (size_t)hstarts.size(); i++)
+      printf("%d ",(int) hstarts[i]);
+    printf("\n");
+    
+    reducedMatvec(ovA,ovX,hstarts[2],temp1);
+    reducedMatvec(ovA,temp1,hstarts[1],temp2);
+    reducedMatvec(ovA,temp2,hstarts[0],ovY);
+
+    // This isn't a Kokkos::deep_copy() since the Kokkos view was complaining about invalid ranges
```
mhoemmen: Could you please clarify?

csiefer2: Kokkos::subview woofs in a static assert.  Something about it not liking the ranges.  Wasn't obvious why.

csiefer2:
I cooked up a smarter way of handling the overlap stuff this AM (basically suck the halo stuff into the reducedMatvec).  Also, I don't check the sizes of X & Y in a correct fashion (this code is functional but... generous n the sizing).

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
mhoemmen: Could you post the error message here?  Let's see if we can figure it out....

csiefer2:
@mhoemmen It also totally fails to run correctly on Cuda, so I'll be asking for your blessing again later :)

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
csiefer2: We're so inconsistent about that :)

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
csiefer2: Easy to fix

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
csiefer2: I'm just in the habit of static casting all "length" numbers because each of the various vcontainers we use returns a different type :)

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
csiefer2: Because gitlab doesn't have a pitchfork emoji  ⛏     :P

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
csiefer2: For this test that's not true --- everything should be happening on the device.  In general, we should definitely check sync states first.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
csiefer2: Code:
```
 auto ovY_lcl = ovY.getLocalViewDevice();
    auto Y_lcl = y_overlap.getLocalViewDevice();
    auto ovYsub = Kokkos::subview(ovY_lcl,std::make_pair(0,Y_lcl.extent(0)), Kokkos::ALL);
```

Error Message:
```
/ascldap/users/csiefer/Trilinos/sandbox/Trilinos/packages/kokkos/core/src/impl/Kokkos_ViewMapping.hpp: In instantiation of ‘struct Kokkos::Impl::is_integral_extent<0u, std::pair<int, long unsigned int>, Kokkos::Impl::ALL_t>’:
/ascldap/users/csiefer/Trilinos/sandbox/Trilinos/packages/kokkos/core/src/impl/Kokkos_ViewMapping.hpp:3230:5:   required from ‘class Kokkos::Impl::ViewMapping<void, Kokkos::ViewTraits<double**, Kokkos::LayoutLeft, Kokkos::Serial, void>, std::pair<int, long unsigned int>, Kokkos::Impl::ALL_t>’
/ascldap/users/csiefer/Trilinos/sandbox/Trilinos/packages/kokkos/core/src/Kokkos_View.hpp:2487:1:   required by substitution of ‘template<class D, class ... P, class ... Args> typename Kokkos::Impl::ViewMapping<void, Kokkos::ViewTraits<DataType, Properties ...>, Args ...>::type Kokkos::subview(const Kokkos::View<D, P ...>&, Args ...) [with D = double**; P = {Kokkos::LayoutLeft, Kokkos::Serial, void}; Args = {std::pair<int, long unsigned int>, Kokkos::Impl::ALL_t}]’
/ascldap/users/csiefer/Trilinos/sandbox/Trilinos/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp:617:89:   required from ‘void {anonymous}::Ifpack2OverlappingRowMatrix_reducedMatvec_UnitTest<Scalar, LocalOrdinal, GlobalOrdinal>::runUnitTestImpl(Teuchos::FancyOStream&, bool&) const [with Scalar = double; LocalOrdinal = int; GlobalOrdinal = int; Teuchos::FancyOStream = Teuchos::basic_FancyOStream<char>]’
/ascldap/users/csiefer/Trilinos/sandbox/Trilinos/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp:646:1:   required from here
/ascldap/users/csiefer/Trilinos/sandbox/Trilinos/packages/kokkos/core/src/impl/Kokkos_ViewMapping.hpp:315:3: error: static assertion failed: subview argument must be either integral or integral extent
   static_assert( value ||

```

csiefer2: Figured it out.  Kokkos::subview is deathly alergic to size_t's in their pair type...

csiefer2: std::make_pair(0, (any size_t variable)) ends in suffering.

mhoemmen: @csiefer2 Right -- `std::make_pair(0, size_t(k))` would deduce a return type of `std::pair<int, size_t>`.  `Kokkos::subview` doesn't seem to know what to do with that....

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -184,14 +184,14 @@ void localReducedMatvec(const MatrixClass & A_lcl,
...
             const LO lclRow = static_cast<LO> (dev.league_rank ()) * rows_per_team + loop;
             
             if (lclRow >= numLocalRows) {
               return;
             }
             const auto A_row = A_lcl.rowConst(lclRow);
-            const LO row_length = static_cast<LO> (A_row.length);
+            const LO row_length = A_row.length;
             for(LO v=0; v<numVectors; v++) {
               residual_value_type A_x = KAT::zero ();
```
mhoemmen: I'm realizing now that we really don't need `ArithTraits` for zero -- `{}` initialization works just fine for any reasonable arithmetic type:
```c++
residual_value_type A_x {};
```

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -599,25 +606,23 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(Ifpack2OverlappingRowMatrix, reducedMatvec, Sc
...
 
-    // This isn't a Kokkos::deep_copy() since the Kokkos view was complaining about invalid ranges
+    // And yes, that int cast is really necessary
     auto ovY_lcl = ovY.getLocalViewDevice();
     auto Y_lcl = y_overlap.getLocalViewDevice();
-    Kokkos::parallel_for("copy out",rowmap->getNodeNumElements(),KOKKOS_LAMBDA(const int i) {
-        for(int j=0; j < numVecs; j++)
-          Y_lcl(i,j) = ovY_lcl(i,j);
-      });
+    auto ovYsub = Kokkos::subview(ovY_lcl,std::make_pair(0,(int)Y_lcl.extent(0)), Kokkos::ALL);
```
mhoemmen: You could also do this:
```c++
std::pair<size_t, size_t>(0, Y_lcl.extent(0))
```

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
csiefer2: This is a strong argument for consolidating the "almost matvec" kernels in one place... KAT is used in the other two or three places these guys exist...

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -234,15 +234,16 @@ void reducedMatvec(const OverlappedMatrixClass & A,
...
   localReducedMatvec(undA_lcl,X_lcl,numLocalRows,Y_lcl);
+
   
   // Now, do the "overlapped part"
   if(overlapLevel > 0) {
     int yrange = hstarts[overlapLevel];
     auto Y_ext = Kokkos::subview(Y_lcl,std::make_pair(numLocalRows,numLocalRows+yrange),Kokkos::ALL());
     
-    int xlimit = numLocalRows + ( (overlapLevel == hstarts.size()-1) ? X_lcl.extent(0) : hstarts[overlapLevel+1] );
+    int xlimit = ( (overlapLevel == hstarts.size()-1) ? X_lcl.extent(0) : numLocalRows+hstarts[overlapLevel+1] );
```
mhoemmen: Hm, this might not even compile, or it might warn.  `numLocalRows` has type `size_t`, and `size_t` + `int` is `size_t`.  `X_lcl.extent(0)` might be `unsigned int` (for `CudaSpace`) or `size_t` (for `HostSpace`).  The two branches thus may have different types.  Also, both are unsigned, but you're assigning them to signed.

mhoemmen:
@csiefer2 Only the non-MPI ("SERIAL") test is failing, and the reason looks easy to fix:
```
p=0 | 53. Ifpack2OverlappingRowMatrix_default_scalar_type_default_local_ordinal_type_default_global_ordinal_type_reducedMatvec_UnitTest ... 
p=0 |  
p=0 |  p=0: *** Caught standard std::exception of type 'std::runtime_error' :
p=0 |  
p=0 |   /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_gcc_4.9.3_SERIAL/Trilinos/packages/ifpack2/src/Ifpack2_OverlappingRowMatrix_def.hpp:83:
p=0 |   
p=0 |   Throw number = 5
p=0 |   
p=0 |   Throw test that evaluated to true: A_->getComm()->getSize() == 1
p=0 |   
p=0 |   Ifpack2::OverlappingRowMatrix: Matrix must be distributed over more than one MPI process.
p=0 |  [FAILED]  (0.000289 sec) Ifpack2OverlappingRowMatrix_default_scalar_type_default_local_ordinal_type_default_global_ordinal_type_reducedMatvec_UnitTest
p=0 |  Location: /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_gcc_4.9.3_SERIAL/Trilinos/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp:569
```

PR #6306, opened 2019-11-19
Automatic snapshot commit from seacas at a34490f

gsjaardema:

At commit:

commit a34490f6a901f724f7544a486a1fc9d9464b9207
Author:  Greg Sjaardema <gdsjaar@sandia.gov>
Date:    Tue Nov 19 08:08:36 2019 -0700
Summary: SLICE: Fix incorrect equal cal

* Replace case_strcmp with more C++-like str_equal and substr_equal
* Use C++11 class member initialization syntax
* EPU: Check subcycles < proc count
* APREPRO: improve array interactions with other types
* EXODUS: Static analyzer fixes
* CATCH2: Update 2.9.2 -> 2.10.2
* FMT: Update to current version
* PYTHON: minor cleanups in exodus.py and exomerge.py

@trilinos/seacas


packages/seacas/applications/epu/EP_ExodusEntity.h
```diff
@@ -106,14 +97,14 @@ namespace Excn {
     size_t entity_count() const { return elementCount; }
 
     char                     elType[MAX_STR_LENGTH + 1]{};
-    std::string              name_;
+    std::string              name_{""};
```
mhoemmen: `std::string`'s default constructor initializes it to the empty string, so you don't need to do this here.

packages/seacas/applications/epu/EP_ExodusEntity.h
```diff
@@ -106,14 +97,14 @@ namespace Excn {
...
-    int64_t                  offset_;
-    int                      position_;
+    int64_t                  id{0};
+    int64_t                  elementCount{0};
+    int                      nodesPerElement{0};
+    int                      attributeCount{0};
+    int64_t                  offset_{0};
+    int                      position_{0};
 
     Block &operator=(const Block &other)
```
mhoemmen: If `copy_string` does what I think it does to an `std::string`, then you should be able to get rid of the body of this function:
```c++
Block& operator=(const Block&) = default;
```

packages/seacas/applications/epu/EP_ExodusEntity.h
```diff
@@ -193,39 +184,35 @@ namespace Excn {
   class CommunicationMap
   {
   public:
-    CommunicationMap() : id(0), entityCount(0), type('U') {}
```
mhoemmen: You shouldn't even need to state the default constructor here, due to the member initializers below.

packages/seacas/applications/epu/EP_ExodusEntity.h
```diff
@@ -193,39 +184,35 @@ namespace Excn {
...
 
   class CommunicationMetaData
   {
   public:
-    CommunicationMetaData()
-        : processorId(0), processorCount(0), globalNodes(0), globalElements(0), nodesInternal(0),
-          nodesBorder(0), nodesExternal(0), elementsInternal(0), elementsBorder(0)
-    {
-    }
+    CommunicationMetaData()                              = default;
```
mhoemmen: See above.

packages/seacas/applications/epu/EP_SystemInterface.C
```diff
@@ -507,6 +501,19 @@ bool Excn::SystemInterface::parse_options(int argc, char **argv)
   else {
     throw std::runtime_error("\nERROR: (EPU) basename not specified\n");
   }
+
+  // Check that subcycle count does not match processor count --
+  // in that case the existing files will be overwritten.
+  if (processorCount_ <= subcycle_) {
+    if (myRank_ == 0) {
+      fmt::print(stderr,
```
mhoemmen: This is going into the C++ Standard Library soon! :-D

packages/seacas/applications/exodiff/ED_SystemInterface.C
```diff
@@ -30,19 +30,13 @@ namespace {
                               Tolerance &def_tol, std::vector<std::string> &names,
                               std::vector<Tolerance> &toler, int max_names);
 
-  int case_strcmp(const std::string &s1, const std::string &s2)
+  bool str_equal(const std::string &s1, const std::string &s2)
```
mhoemmen: Could it pay to define this in one place?

packages/seacas/libraries/aprepro_lib/apr_builtin.cc
```diff
@@ -1064,6 +1064,7 @@ namespace SEAMS {
         }
       }
       assert(rows - rows_to_skip == (size_t)array_data->rows);
+      delete file;
```
mhoemmen: Consider using `std::unique_ptr` to manage raw pointers like this.

packages/seacas/libraries/aprepro_lib/apr_scanner.cc
```diff
@@ -2328,11 +2328,11 @@ yyFlexLexer::yyFlexLexer(std::istream *arg_yyin, std::ostream *arg_yyout)
   yy_start_stack_ptr = yy_start_stack_depth = 0;
   yy_start_stack                            = NULL;
```
mhoemmen: `nullptr` works here too :-)

packages/seacas/applications/epu/EP_ExodusEntity.h
gsjaardema: That is why it is being removed.... ;)

packages/seacas/applications/epu/EP_SystemInterface.C
gsjaardema: Yes!  I was glad to hear that.  Although I may be gone by the time we are allowed to use C++20...

packages/seacas/applications/exodiff/ED_SystemInterface.C
gsjaardema: Yes, should probably go in either IOSS or in Suplib.  Will look into that.

packages/seacas/libraries/aprepro_lib/apr_scanner.cc
gsjaardema: Auto-generated code.

packages/seacas/applications/epu/EP_ExodusEntity.h
gsjaardema: Will take a look.  Thanks.

packages/seacas/libraries/aprepro_lib/apr_builtin.cc
gsjaardema: Yes, I definitely need to go through and use many more unique_ptr and shared_ptr...

gsjaardema:
@mhoemmen Thanks for detailed review.

packages/seacas/applications/epu/EP_SystemInterface.C
mhoemmen: @gsjaardema It may not be quite so long -- we're thinking about C++14 finally.  C++17 is a bigger jump due to semantics changes, but 20 might not be so bad.

Issue #962, opened 2017-01-03
Tpetra: Avoid atomic ops for unpack if no duplicate LIDs


mhoemmen:
@trilinos/tpetra 
Epic: #796
This story matters to other epics besides #796.

Thread-parallel implementations of Tpetra::DistObject::unpackAndCombine(New) do not need to do atomic updates if the result LIDs have no duplicates (meaning that at most one thread will ever modify any one value of the destination DistObject at a time).  "Result LIDs have no duplicates" is a property of the Import / Export object, so the Import / Export object should remember this at construction time for reuse.

This matters because atomic updates have a run-time cost, even if there is no contention between threads.  Sparse matrix-vector multiply does an Import on the input MultiVector.  In the common case, the result LIDs in this Import should have no duplicates.  (This is because the column Map is constructed that way, if users let CrsGraph or CrsMatrix construct the column Map.)  Thus, atomic updates add unnecessary run time to this important kernel.

@jjellio and @tjfulle might be interested in this.  I would like to fix this for MultiVector in a way that neatly encapsulates its unpack kernels (and ideally also its pack kernels).  Fixing this issue should have the side effect of improving MPI-only performance, which is an important use case for many customers (who do not have hardware that requires MPI + threads for good performance).

mhoemmen:
The `getRemoteLIDs` method of Import and Export tells one the LIDs to receive from incoming processes.  These LIDs go into `MultiVector::unpackAndCombineNew` as the first argument `importLIDs`, and in turn go into unpack_array_multi_column as its `idx` (indices) argument.  That array is exactly the set of indices (MultiVector rows, in this case) where duplicates require atomic updates in a thread-parallel unpack kernel.  (See lines 797-800 of tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp.)

It should be easy enough, during Import / Export setup, to scan that array for duplicates.

Note that Stokhos depends on partial specializations of the pack and unpack kernels, and we will need to reimplement those specializations for this fix.  This means we need to enable Stokhos (explicitly! it may not be Primary Tested yet) when testing this fix.

mhoemmen:
[...]

PR #6316, opened 2019-11-20
Tpetra: Fix #6313  (and build warnings)

### LINKED ISSUE #6313: Tpetra::CrsMatrix unpack should only use atomic updates for ADD, and with more than one thread
## Bug Report
@trilinos/tpetra 

### Description

`Tpetra::CrsMatrix::unpackAndCombine*` currently decides whether to use atomic updates, based only on whether the matrix's `execution_space` is `Kokkos::Serial`.  This is too conservative.  It really only needs atomic updates under the following conditions:

1. If `execution_space::concurrency() > 1`, and
2. if the `CombineMode` is `ADD`.

Multiple threads attempting to unpack into the same row is possible, because

- thread parallelization for `unpackAndCombine*` is over import LIDs, not over target matrix rows; and
- multiple MPI processes might send the same process data for the same target matrix row.

However, the `REPLACE` `CombineMode` doesn't make sense if multiple threads could write to the same data, so we might as well not bother doing atomic updates in that case.  (`Tpetra::MultiVector::unpackAndCombine*` already exploits this optimization.)

This is related to #962, but that is a more general problem best fixed by having the Export or Import object report to the target DistObject (during a doExport or doImport) whether the list of import LIDs has duplicates.  No duplicates means no threads can collide, which would mean no need for atomic updates even with multiple threads unpacking concurrently.

mhoemmen:
@trilinos/tpetra 

## Motivation

Speed up unpack for dense CrsMatrix rows, corresponding to bulk nodes in Aria.

## Related Issues

* Closes #6313 
* Related to #962 

## Stakeholder Feedback

See TF-1264.

### DIFF
--- a/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixDenseRowUnpack.cpp
+++ b/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixDenseRowUnpack.cpp
@@ -146,6 +146,30 @@ getLclRowsToTest(const export_type& exporter)
   for(const LO incomingRow : incomingRows) {
     lclRowsToTest_h[incomingRow] = 1;
   }
+
+  const bool verbose = Tpetra::Details::Behavior::verbose();
+  if(verbose) {
+    using std::cerr;
+    using std::endl;
+
+    const auto comm = exporter.getSourceMap()->getComm();
+    const int myRank = comm->getRank();
+    if(myRank == 0) {
+      cerr << "lclRowsToTest:" << endl;
+    }
+    std::ostringstream os;
+    os << "Proc " << myRank << ": [";
+    const LO numLclRowsToTest(lclRowsToTest_h.extent(0));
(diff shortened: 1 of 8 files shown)


mhoemmen:
@vbrunini FYI

mhoemmen:
@csiefer2 Thanks for approving!  There's a minor build error I need to fix yet.

mhoemmen:
@csiefer2 I think I've fixed it :-)

mhoemmen:
@trilinos/tpetra @csiefer2 might I pester y'all for another review? thanks! :-D

mhoemmen:
@trilinos/tpetra @rrdrake @sebrowne also this fixes build warnings for Sierra :)

mhoemmen:
@trilinos/tpetra :)

mhoemmen:
Thanks @csiefer2 ! :-D

PR #6324, opened 2019-11-21
Stacked timer changes

cgcgcg:
@trilinos/teuchos 

## Motivation
- Add option to print rank with max time.
- Take min only over active ranks.

### DIFF
--- a/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
+++ b/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
@@ -142,8 +142,14 @@ StackedTimer::collectRemoteData(Teuchos::RCP<const Teuchos::Comm<int> > comm, co
   reduce(used.getRawPtr(), active_.getRawPtr(), num_names, REDUCE_SUM, 0, *comm);
 
   if (min_.size()) {
-    reduceAll(*comm, REDUCE_MIN, num_names, time.getRawPtr(), min_.getRawPtr());
     reduceAll(*comm, REDUCE_MAX, num_names, time.getRawPtr(), max_.getRawPtr());
+    for (int i=0;i<num_names;++i)
+      if (!used[i])
+        time[i] = max_[i];
+    reduceAll(*comm, REDUCE_MIN, num_names, time.getRawPtr(), min_.getRawPtr());
+    for (int i=0;i<num_names;++i)
+      if (!used[i])
+        time[i] = 0.;
     if (procmin_.size()) {
       Array<int> procmin(num_names);
       Array<int> procmax(num_names);
--- a/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
+++ b/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
@@ -94,11 +94,16 @@ StackedTimer::collectRemoteData(Teuchos::RCP<const Teuchos::Comm<int> > comm, co
   updates_.resize(num_names);
   active_.resize(num_names);
 
-  if (options.output_minmax || options.output_histogram) {
+  if (options.output_minmax || options.output_histogram || options.output_proc_minmax) {
     min_.resize(num_names);
     max_.resize(num_names);
     if ( options.output_minmax )
       sum_sq_.resize(num_names);
+  }
+
+  if (options.output_proc_minmax) {
+    procmin_.resize(num_names);
+    procmax_.resize(num_names);
   }
 
 
@@ -139,6 +144,23 @@ StackedTimer::collectRemoteData(Teuchos::RCP<const Teuchos::Comm<int> > comm, co
   if (min_.size()) {
     reduceAll(*comm, REDUCE_MIN, num_names, time.getRawPtr(), min_.getRawPtr());
     reduceAll(*comm, REDUCE_MAX, num_names, time.getRawPtr(), max_.getRawPtr());
+    if (procmin_.size()) {
+      Array<int> procmin(num_names);
+      Array<int> procmax(num_names);
+      int commRank = comm->getRank();
+      for (int i=0;i<num_names; ++i) {
+        if (used[i] && (min_[i]==time[i]))
+          procmin[i] = commRank;
+        else
+          procmin[i] = -1;
+        if (used[i] && (max_[i]==time[i]))
+          procmax[i] = commRank;
+        else
+          procmax[i] = -1;
+      }
+      reduceAll(*comm, REDUCE_MAX, num_names, procmin.getRawPtr(), procmin_.getRawPtr());
+      reduceAll(*comm, REDUCE_MAX, num_names, procmax.getRawPtr(), procmax_.getRawPtr());
+    }
   }
 
   if (options.output_histogram) {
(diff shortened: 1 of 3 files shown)


rppawlo:
Thanks for adding @cgcgcg ! Could you add a quick unit test for this? You can set the time on any timer (see stacked_timer.cpp unit test, line 385), so you artificially hack one rank take longer to run and then check that the correct process is reported?

packages/teuchos/comm/src/Teuchos_StackedTimer.hpp
```diff
@@ -589,13 +589,14 @@ class StackedTimer
...
     OutputOptions() : output_fraction(false), output_total_updates(false), output_histogram(false),
-                      output_minmax(false), num_histogram(10), max_levels(INT_MAX),
+                      output_minmax(false), output_procmax(false), num_histogram(10), max_levels(INT_MAX),
                       print_warnings(true), align_columns(false), print_names_before_values(true),
                       drop_time(-1.0) {}
     bool output_fraction;
     bool output_total_updates;
     bool output_histogram;
     bool output_minmax;
+    bool output_procmax;
```
mhoemmen: Prefer C++11 member initializers.  Then you could just write `OutputOptions() = default;` (or not write anything at all).

packages/teuchos/comm/src/Teuchos_StackedTimer.hpp
```diff
@@ -653,6 +656,7 @@ class StackedTimer
       total_updates_(0),
       min_(0),
       max_(0),
+      procmax_(0),
```
mhoemmen: Again, prefer member initializers to implementing a nondefault default constructor.

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -137,8 +142,31 @@ StackedTimer::collectRemoteData(Teuchos::RCP<const Teuchos::Comm<int> > comm, co
   reduce(used.getRawPtr(), active_.getRawPtr(), num_names, REDUCE_SUM, 0, *comm);
 
   if (min_.size()) {
-    reduceAll(*comm, REDUCE_MIN, num_names, time.getRawPtr(), min_.getRawPtr());
     reduceAll(*comm, REDUCE_MAX, num_names, time.getRawPtr(), max_.getRawPtr());
+    for (int i=0;i<num_names;++i)
+      if (!used[i])
+        time[i] = max_[i];
+    reduceAll(*comm, REDUCE_MIN, num_names, time.getRawPtr(), min_.getRawPtr());
```
mhoemmen: You could combine these two all-reduces in one, by the `max(-x) == -min(x)` trick.

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -137,8 +142,31 @@ StackedTimer::collectRemoteData(Teuchos::RCP<const Teuchos::Comm<int> > comm, co
...
+        if (used[i] && (min_[i]==time[i]))
+          procmin[i] = commRank;
+        else
+          procmin[i] = -1;
+        if (used[i] && (max_[i]==time[i]))
+          procmax[i] = commRank;
+        else
+          procmax[i] = -1;
+      }
+      reduceAll(*comm, REDUCE_MAX, num_names, procmin.getRawPtr(), procmin_.getRawPtr());
```
mhoemmen: See above.  I think Tpetra uses this trick somewhere.

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -255,6 +283,20 @@ StackedTimer::computeColumnWidthsForAligment(std::string prefix,
           os << "}";
         alignments_.max_ = std::max(alignments_.max_,os.str().size());
       }
+      if (procmin_.size()) {
+        std::ostringstream os;
+        os << ", proc min=" << procmin_[i];
+        if (active_[i] <= 1)
+          os << "}";
+        alignments_.procmin_ = std::min(alignments_.procmin_,os.str().size());
```
mhoemmen: The two arguments of `min` might have different type here; that could lead to warnings at best.

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -255,6 +283,20 @@ StackedTimer::computeColumnWidthsForAligment(std::string prefix,
...
+      }
+      if (procmax_.size()) {
+        std::ostringstream os;
+        os << ", proc max=" << procmax_[i];
+        if (active_[i] <= 1)
+          os << "}";
+        alignments_.procmax_ = std::max(alignments_.procmax_,os.str().size());
```
mhoemmen: See above.

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -255,6 +283,20 @@ StackedTimer::computeColumnWidthsForAligment(std::string prefix,
...
+      }
       if (active_[i]>1) {
         std::ostringstream os;
         os << ", std dev=" << sqrt(std::max<double>(sum_sq_[i]-sum_[i]*sum_[i]/active_[i],0.0)/(active_[i]-1));
```
mhoemmen: In some cases, the standard library forbids specifying the template arguments of a standard algorithm.  I think it's allowed for `max` etc., but it's poor form in my opinion.  (General best practice is to rely on template argument deduction for nonmember functions.)  It would be better to cast the first argument to `double`.  (`0.0` already has type `double`.)  Doing so would help avoid any truncation if `sum_[i]` and `active_[i]` both have integer types.

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -395,6 +437,24 @@ StackedTimer::printLevel (std::string prefix, int print_level, std::ostream &os,
...
+        tmp <<", proc max="<<procmax_[i];
+        if (active_[i] <= 1)
+          tmp << "}";
+        if (options.align_columns)
+          os << std::left << std::setw(alignments_.procmax_);
+        os << tmp.str();
+      }
       if (active_[i]>1) {
         std::ostringstream tmp;
         tmp << ", std dev="<<sqrt(std::max<double>(sum_sq_[i]-sum_[i]*sum_[i]/active_[i],0.0)/(active_[i]-1));
```
mhoemmen: See above.

packages/teuchos/comm/src/Teuchos_StackedTimer.hpp
```diff
@@ -589,13 +589,14 @@ class StackedTimer
    */
   struct OutputOptions {
     OutputOptions() : output_fraction(false), output_total_updates(false), output_histogram(false),
-                      output_minmax(false), num_histogram(10), max_levels(INT_MAX),
+                      output_minmax(false), output_proc_minmax(false), num_histogram(10), max_levels(INT_MAX),
```
mhoemmen: As mentioned previously, prefer member initializers; that would let you write `OutputOptions() = default;` here, and it could help make the class trivially default constructible.

PR #6326, opened 2019-11-21
MueLu: fixing the count of aggregated nodes in refactored Phase2a of aggregation, see #6325

### LINKED ISSUE #6325: MueLu: Phase2a of refactored aggregation wrongly counts aggregated nodes
## Bug Report
@trilinos/muelu 

### Description
There is a bug in refactored aggregation that leads to the wrong number of aggregated nodes after phase 2a. This sometimes leads to unaggregated nodes or to have a negative number of unaggregated nodes.

lucbv:
A piece of logic is wrong in the algorithms that counts
how many nodes have been aggregated in Phase2a of aggregation.
This commit fixes that by individually counting every node
that is being aggregated instead of counting all the nodes in
a newly formed aggregate, at once.

@trilinos/muelu 

## Motivation
All bugs shall be fixed, but this is also generating a failing test in Empire and blocking the adoption of the latest refactored setup of MueLu.

## Related Issues

* Closes #6325
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback
Empire: @cgcgcg will check that the failing test is fixed after this modification is applied.
ExaWind: @lucbv checked that ExaWind is not negatively affected by these changes.

## Testing
Local tests have been performed and @cgcgcg will apply locally these changes to test these with Empire

### DIFF
--- a/packages/muelu/src/Graph/UncoupledAggregation/MueLu_AggregationPhase2aAlgorithm_kokkos_def.hpp
+++ b/packages/muelu/src/Graph/UncoupledAggregation/MueLu_AggregationPhase2aAlgorithm_kokkos_def.hpp
@@ -136,14 +136,20 @@ namespace MueLu {
 
                                   // Loop over neighbors to count how many nodes could join
                                   // the new aggregate
+                                  // Note on 2019-11-22, LBV:
+                                  // The rootCandidate is not taken into account and in fact
+                                  // not aggregatesd later on. To change that we want to
+                                  // modify:
+                                  // if(aggSize < maxNodesPerAggregate)
+                                  // to:
+                                  // if(aggSize < maxNodesPerAggregate - 1)
                                   LO numNeighbors = 0;
                                   for(int j = 0; j < neighbors.length; ++j) {
                                     LO neigh = neighbors(j);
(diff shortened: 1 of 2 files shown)


packages/muelu/src/Graph/UncoupledAggregation/MueLu_AggregationPhase2aAlgorithm_kokkos_def.hpp
```diff
@@ -165,13 +165,13 @@ namespace MueLu {
                                         if(graph.isLocalNeighborVertex(neigh) &&
                                            aggStat(neigh) == READY &&
                                            aggSize < maxNodesPerAggregate) {
```
cgcgcg: Unrelated note: once `aggSize` hits `maxNodesPerAggregate` in the loop above, this condition is always false.

lucbv: Changed the logic so that a partial aggregate is now form as it was the case in the non-refactored code.

PR #6331, opened 2019-11-22
SEACAS: Bug fix since snapshot

gsjaardema:
* Conjoin -- fix time tolerance testing for inter-database times
* Aprepro -- fix eating of newline following endswitch

@trilinos/seacas


packages/seacas/libraries/aprepro_lib/test_standard.out
```diff
@@ -82,6 +82,7 @@ $ Test switch
 This should echo
 3.141592654
 
+
```
mhoemmen: Alas, we'll have to wait until C++20 lands to get `std::numbers::pi_v`: https://en.cppreference.com/w/cpp/numeric/constants

PR #6343, opened 2019-11-25
ShyLU/FROSch: New implementation of BuildRepeatedMap()

searhein:
@trilinos/shylu 

## Motivation

This PR contains a new implementation of BuildRepeatedMap() as needed for an application code. 
@chochmuth, can you please check if this solves our issue in the application?


packages/shylu/shylu_dd/frosch/src/CoarseSpaces/FROSch_CoarseSpace_def.hpp
```diff
@@ -115,7 +115,7 @@ namespace FROSch {
         reduceAll(*this->MpiComm_,REDUCE_MAX,int(buildUniqueMap),ptr(&buildUniqueMapMax));
         
         if (buildUniqueMapMax>0) {
-            if (this->MpiComm_->getRank()==0) std::cout << "FROSch::CoarseSpace : WARNING: We re-build a unique map of the AssembledBasisMap_." << std::endl;
+            if (this->MpiComm_->getRank()==0) std::cout << "FROSch::CoarseSpace : WARNING: We re-build a unique map of AssembledBasisMap_." << std::endl;
```
mhoemmen: 1. Warnings should go to `std:cerr`, not `std::cout`.
2. Warnings are best reserved for situations where the user did something wrong.  If the user didn't do anything wrong here, then the warning is just useless noise.  Users will report it as a bug and developers will have to waste time explaining that nothing is wrong.

packages/shylu/shylu_dd/frosch/src/CoarseSpaces/FROSch_CoarseSpace_def.hpp
```diff
@@ -150,10 +150,10 @@ namespace FROSch {
             }
         }
 
-        UnassembledBasesMaps_.resize(0);
-        UnassembledBasesMapsUnique_.resize(0);
-        UnassembledSubspaceBases_.resize(0);
-        Offsets_.resize(0);
+        UnassembledBasesMaps_.clear();
```
mhoemmen: This doesn't necessarily free anything.  The standard idiom for ensuring that all the elements of a `std::vector` (or `Teuchos::Array`) get freed is to swap the vector with an empty vector.

packages/shylu/shylu_dd/frosch/src/InterfaceSets/FROSch_DDInterface_def.hpp
```diff
@@ -687,12 +687,12 @@ namespace FROSch {
...
-                    XMapPtr rangeMap = MapFactory<LO,GO,NO>::Build(NodesMap_->lib(),-1,myPID(),0,NodesMap_->getComm());
+                    XMapPtr domainMap = MapFactory<LO,GO,NO>::Build(NodesMap_->lib(),-1,myPID(),0,NodesMap_->getComm());
 
-                    commMat->fillComplete(NodesMap_,rangeMap);
+                    commMat->fillComplete(domainMap,NodesMap_);
                     commMatTmp->doExport(*commMat,*commExporter,INSERT);
-                    commMatTmp->fillComplete(UniqueNodesMap_,rangeMap);
-                    commMat = MatrixFactory<SC,LO,GO,NO>::Build(NodesMap_,10);
+                    commMatTmp->fillComplete(domainMap,UniqueNodesMap_);
+                    commMat = MatrixFactory<SC,LO,GO,NO>::Build(NodesMap_,as<LO>(0));
```
mhoemmen: You don't need `as` here.  Just do this: `LO(0)`.

packages/shylu/shylu_dd/frosch/src/CoarseSpaces/FROSch_CoarseSpace_def.hpp
searhein: I just made another commit, which should take care of your three comments.

packages/shylu/shylu_dd/frosch/src/CoarseSpaces/FROSch_CoarseSpace_def.hpp
searhein: See above.

packages/shylu/shylu_dd/frosch/src/InterfaceSets/FROSch_DDInterface_def.hpp
searhein: See above.

packages/shylu/shylu_dd/frosch/src/Adapters/Thyra_FROSchFactory_def.hpp
```diff
@@ -422,7 +423,7 @@ namespace Thyra {
                     repeatedMap = rcp_dynamic_cast<ConstXMap>(xTpetraRepeatedMap);
                 } else {
 #ifdef HAVE_SHYLU_DDFROSCH_EPETRA
-                    if (comm->getRank()==0) std::cerr << "FROSch::FROSchFactory : WARNING: Cannot retrieve Epetra objects from ParameterList. Use Xpetra instead." << std::endl;
+                    FROSCH_WARNING("FROSch::FROSchFactory",comm->getRank()==0,"Cannot retrieve Epetra objects from ParameterList. Use Xpetra instead.");
```
mhoemmen: I like that you're using a macro here :-) .  Can the factory actually keep going and return a meaningful result if this happens?

packages/shylu/shylu_dd/frosch/src/Adapters/Thyra_FROSchLinearOp_def.hpp
```diff
@@ -56,7 +56,9 @@ namespace Thyra {
     // Constructors/initializers
     template <class SC, class LO, class GO, class NO>
     FROSchLinearOp<SC,LO,GO,NO>::FROSchLinearOp()
-    {}
+    {
```
mhoemmen: See above (you don't need to define the constructor explicitly if its body would be empty anyway; either let the compiler define it or use the `=default` syntax in the class declaration).

packages/shylu/shylu_dd/frosch/src/CoarseSpaces/FROSch_CoarseSpace_decl.hpp
```diff
@@ -138,19 +138,19 @@ namespace FROSch {
         CommPtr MpiComm_;
         CommPtr SerialComm_;
         
-        ConstXMapPtrVec UnassembledBasesMaps_;
-        ConstXMapPtrVec UnassembledBasesMapsUnique_;
+        ConstXMapPtrVec UnassembledBasesMaps_ = ConstXMapPtrVec(0);
```
mhoemmen: That _should_ be what the default constructor of this class does, so you don't need the member initializer here.  (You don't explicitly initialize `MpiComm_` above, and you don't need to, so why initialize this one explicitly?)

packages/shylu/shylu_dd/frosch/src/CoarseSpaces/FROSch_CoarseSpace_def.hpp
```diff
@@ -115,7 +106,7 @@ namespace FROSch {
         reduceAll(*this->MpiComm_,REDUCE_MAX,int(buildUniqueMap),ptr(&buildUniqueMapMax));
         
         if (buildUniqueMapMax>0) {
-            if (this->MpiComm_->getRank()==0) std::cout << "FROSch::CoarseSpace : WARNING: We re-build a unique map of the AssembledBasisMap_." << std::endl;
+            FROSCH_NOTIFICATION("FROSch::CoarseSpace",this->MpiComm_->getRank()==0,"We re-build a unique map of AssembledBasisMap_.");
```
mhoemmen: I _really_ like that this is a macro.  (It would be awesome if these notifications were off by default, but users could enable them at run time via an environment variable.)

packages/shylu/shylu_dd/frosch/src/CoarseSpaces/FROSch_CoarseSpace_def.hpp
```diff
@@ -176,12 +174,10 @@ namespace FROSch {
 
         LO iD;
         SC valueTmp;
-        GOVec indices;
-        SCVec values;
 
         for (UN i=0; i<AssembledBasis_->getLocalLength(); i++) {
-            indices.resize(0);
-            values.resize(0);
+            GOVec indices;
```
mhoemmen: The resize approach wasn't actually so bad.  `std::vector`, and therefore `Teuchos::Array`, do not necessarily free memory if you resize to zero.  Thus, resizing the existing vectors might have let you reuse storage.  I don't think that will affect performance much, though.

packages/shylu/shylu_dd/frosch/src/CoarseSpaces/FROSch_GDSWInterfacePartitionOfUnity_def.hpp
```diff
@@ -157,7 +147,7 @@ namespace FROSch {
         FROSCH_TIMER_START_LEVELID(sortInterfaceTime,"GDSWInterfacePartitionOfUnity::sortInterface");
         if (this->ParameterList_->get("Test Unconnected Interface",true)) {
             if (matrix.is_null()) {
-                if (this->Verbose_) std::cout << "FROSch::GDSWInterfacePartitionOfUnity : WARNING: divideUnconnectedEntities() cannot be performed without the matrix." << std::endl;
+                FROSCH_WARNING("FROSch::GDSWInterfacePartitionOfUnity",this->Verbose_,"divideUnconnectedEntities() cannot be performed without the matrix.");
```
mhoemmen: This looks like an error condition.  Can you safely continue and return a reasonable result after this point?

packages/shylu/shylu_dd/frosch/src/SchwarzOperators/FROSch_CoarseOperator_decl.hpp
```diff
@@ -167,16 +167,16 @@ namespace FROSch {
...
+        ConstXMapPtrVecPtr GatheringMaps_ = ConstXMapPtrVecPtr(0);
         XMapPtr CoarseSolveMap_;
 
         SubdomainSolverPtr CoarseSolver_;
 
         ParameterListPtr DistributionList_;
 
-        XExportPtrVecPtr CoarseSolveExporters_;
+        XExportPtrVecPtr CoarseSolveExporters_ = XExportPtrVecPtr(0);
 #ifdef FROSCH_COARSEOPERATOR_EXPORT_AND_IMPORT
```
mhoemmen: Was this issue (with reverse-mode Export not doing what you wanted) ever fixed?  If so, then we can get rid of this macro.

packages/shylu/shylu_dd/frosch/src/SchwarzOperators/FROSch_GDSWCoarseOperator_def.hpp
```diff
@@ -216,7 +215,7 @@ namespace FROSch {
         GOVecPtr dirichletBoundaryDofs = FindOneEntryOnlyRowsGlobal(this->K_->getCrsGraph(),nodesMap);
 #end
  */
-        if (this->Verbose_) std::cout << "FROSch::GDSWCoarseOperator : WARNING: We do not have the right map (repeatedMap) to use FindOneEntryOnlyRowsGlobal. A variant that uses the row map could be implemented?! => We use dirichletBoundaryDofs = null for now" << std::endl;
+        FROSCH_WARNING("FROSch::GDSWCoarseOperator",this->Verbose_,"We do not have the right map (repeatedMap) to use FindOneEntryOnlyRowsGlobal. A variant that uses the row map could be implemented?! => We use dirichletBoundaryDofs = null for now.");
```
mhoemmen: This looks like an error condition.

packages/shylu/shylu_dd/frosch/src/SchwarzOperators/FROSch_SchwarzOperator_def.hpp
```diff
@@ -53,31 +53,21 @@ namespace FROSch {
...
     MpiComm_ (k->getRangeMap()->getComm()),
-    SerialComm_ (),
     K_ (k),
     ParameterList_ (parameterList),
     Verbose_ (MpiComm_->getRank()==0),
-    IsInitialized_ (false),
-    IsComputed_ (false),
     LevelID_ (ParameterList_->get("Level ID",UN(1)))
     {
         FROSCH_ASSERT(getDomainMap()->isSameAs(*getRangeMap()),"SchwarzOperator assumes DomainMap==RangeMap");
```
mhoemmen: Note that `Map::isSameAs` may do an all-reduce in the worst case.

packages/shylu/shylu_dd/frosch/src/SchwarzPreconditioners/FROSch_TwoLevelBlockPreconditioner_def.hpp
```diff
@@ -110,26 +109,29 @@ namespace FROSch {
...
-//            XMapPtrVecPtr subMapVec = BuildSubMaps(tmpMap,blockMaxGIDVec);// Todo: Achtung, die UniqueMap könnte unsinnig verteilt sein. Falls es eine repeatedMap gibt, sollte dann die uniqueMap neu gebaut werden können. In diesem Fall, sollte man das aber basierend auf der repeatedNodesMap tun
-//            repeatedMapVec = BuildRepeatedSubMaps(this->K_,subMapVec);
-//
-//        }
+        FROSCH_ASSERT(!repeatedMapVec.is_null(),"repeatedMapVec.is_null() = true. Please provide the repeated maps vector. The maps itself can be null and will be constructed.");
+        for (UN i = 0; i < repeatedMapVec.size(); i++) {
+            if (repeatedMapVec[i].is_null()) {
+                FROSCH_ASSERT( i==0, "We can only construct a repeated map for a non block system");
+                repeatedMapVec[i] = BuildRepeatedMap( this->K_ );
+//                repeatedMapVec[i] = BuildRepeatedMap(this->K_, this->ParameterList_->get("Reduce approx repeated map",true) ); // Todo: Achtung, die UniqueMap könnte unsinnig verteilt sein. Falls es eine repeatedMap gibt, sollte dann die uniqueMap neu gebaut werden können. In diesem Fall, sollte man das aber basierend auf der repeatedNodesMap tun
```
mhoemmen: The situation you mention in the comment sounds like an error condition.  If so and if it's easy to detect, then you should just `FROSCH_ASSERT` here.

packages/shylu/shylu_dd/frosch/src/Tools/FROSch_Tools_decl.hpp
```diff
@@ -58,6 +58,14 @@
...
+#ifndef FROSCH_WARNING
+#define FROSCH_WARNING(CLASS,VERBOSE,OUTPUT) if (VERBOSE) std::cerr << CLASS << " : WARNING: " << OUTPUT << std::endl;
+#endif
+
+#ifndef FROSCH_NOTIFICATION
+#define FROSCH_NOTIFICATION(CLASS,VERBOSE,OUTPUT) if (VERBOSE) std::cout << CLASS << " : NOTIFICATION: " << OUTPUT << std::endl;
+#endif
+
 #ifndef FROSCH_TEST_OUTPUT
 #define FROSCH_TEST_OUTPUT(COMM,VERBOSE,OUTPUT) COMM->barrier(); COMM->barrier(); COMM->barrier(); if (VERBOSE) std::cout << OUTPUT << std::endl;
```
mhoemmen: If you want to print on all processes, consider using something like `Tpetra::Details::gathervPrint` (see `Trilinos/packages/tpetra/core/src/Tpetra_Details_gathervPrint.hpp`).  That tends to produce more consistent output, even if only a few processes end up printing something.

packages/shylu/shylu_dd/frosch/src/Tools/FROSch_Tools_def.hpp
```diff
@@ -474,6 +473,56 @@ namespace FROSch {
         return MapFactory<LO,GO,NO>::Build(tmpGraph->getRowMap()->lib(),-1,repeatedIndices(),0,graph->getRowMap()->getComm());
     }
 
+    template <class LO,class GO,class NO>
+    RCP<const Map<LO,GO,NO> > BuildRepeatedMapOld(RCP<const CrsGraph<LO,GO,NO> > graph)
```
mhoemmen: If it's "Old", why not get rid of it?  If you can't because some application is using it, consider marking it with a deprecation macro.  We can show you how to set that up if it doesn't already exist in ShyLU.

packages/shylu/shylu_dd/frosch/src/Tools/FROSch_Tools_def.hpp
```diff
@@ -877,6 +926,96 @@ namespace FROSch {
...
+        // Build node maps for all blocks
+        for (unsigned block=0; block<nmbBlocks; block++) {
+            
+            if (dofOrderingVec[block] == NodeWise) {
+                ArrayView< const GO > globalIndices = dofsMapsVecVec[block][0]->getNodeElementList();
+                Array<GO> globalIndicesNode( globalIndices );
+                GO offset = dofsMapsVecVec[block][0]->getMinAllGlobalIndex();
+                for (unsigned i=0; i<globalIndicesNode.size(); i++){
+                    // multiplier is not correct if isMergedPrior==true because we substract minAllGIDBlock. We have to adjust for this later
+                    // was ist wenn mergedPrior und blockOffset existiert, dann ist multiplier falsch.
```
mhoemmen: Is that an error condition?  If so, then consider `FROSCH_ASSERT`.

packages/shylu/shylu_dd/frosch/src/Tools/FROSch_Tools_def.hpp
```diff
@@ -1035,29 +1174,33 @@ namespace FROSch {
...
+                    de = tmpMultiVector_j->dot(*tmpMultiVector_j);
+                    tmpMultiVectorNonConst_i_itmp->update(-en/de,*tmpMultiVector_j,ScalarTraits<SC>::one());
                 }
-                norm = tmpMultiVector->getVector(i-itmp)->norm2();
+                norm = tmpMultiVectorNonConst_i_itmp->norm2();
                 if (norm<1.0e-10) {
                     arrayZero.push_back(i);
                     itmp++;
                 } else {
                     //tmpMultiVector->getVectorNonConst(i-itmp)->scale(1.0/norm);
```
mhoemmen: I'm not sure what's going on here, but if you really don't need to scale the multivector, then you should just delete the else branch and comment, to reassure readers that you didn't forget anything here :-) .

PR #6348, opened 2019-11-25
Sacado:  Fix ViewFad conversions for new design.

etphipp:
The new design implementation was allowing conversions from other AD
types to ViewFad, which is not correct.  In addition to being generally
unsafe, this was causing compiler errors in Panzer when the two return
types of the ternary operator differ, with one being ViewFad.

It tooks some time to figure out a way to do this that NVCC accepted.
It seems to not handle inherited constructors disabled through SFINAE
correctly.

@trilinos/<teamName>


etphipp:
> Thansk @etphipp ! Looks good to me. Did not see a unit test for this. Do you need one to protect?

There actually is a unit test (sacado/test/UnitTests/ConversionsTests.cpp), but because of another mistake it was passing with the new design when it shouldn't have.  The test expects different conversion behavior with regular Fad types and ViewFad, but I hadn't specialized IsViewFad<> for the ViewFad in the new design.  That is fixed too and the test passes properly.

Issue #6345, opened 2019-11-25
Tpetra: CrsGraph not enough capacity to insert error


stanmoore1:
## Bug Report
@trilinos/tpetra

### Description
Running EMPIRE on Astra with Trilinos from 11-22-2019, I get an error `Tpetra:CrsGraph not enough capacity to insert ... `. This has the same signature as the so-called "UVM" bug on Sierra. Using an older version of EMPIRE and Trilinos from 11-11-2019 does not fail.

@bathmatt can we get this labeled as a super critical L1 milestone blocker?

rppawlo:
@kddevin @trilinos/tpetra

stanmoore1:
I reproduced this on Stria and got a different error:
```
*********** Caught Exception: Begin Error Report ***********
../packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp:1175:

Throw number = 2

Throw test that evaluated to true: importLen < requiredImportLen

Tpetra::Details::DistributedNoncontiguousDirectory::getEntriesImpl: On Process 15: The 'imports' array must have length at least 1273674, but its actual length is 1202616.  numRecv: 636837, packetSize: 2, numEntries (# GIDs): 636837, numMissing: 0: distor.getTotalReceiveLength(): 601308.
Distributor description: "Tpetra::Distributor": {How initialized: By createFromRecvs, Parameters: {Send type: Send, Barrier between receives and sends: false, Use distinct tags: true, Debug: false}}.
Please report this bug to the Tpetra developers.
************ Caught Exception: End Error Report ***
```

stanmoore1:
I ran it again on Stria and got a different error:

```
*********** Caught Exception: Begin Error Report ***********
../packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp:878:

Throw number = 2

Throw test that evaluated to true: dirMapLid == LINVALID

Tpetra::Details::DistributedNoncontiguousDirectory<int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::OpenMP, Kokkos::HostSpace> > constructor: Incoming global index 15295108 does not have a corresponding local index in the Directory Map.  Please report this bug to the Tpetra developers.
************ Caught Exception: End Error Report ************
*********** Caught Exception: Begin Error Report ***********
../packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp:878:
```

mhoemmen:
[...]

PR #6353, opened 2019-11-26
tpetra:  work around one parallel_scan as an experiment on stria for #6345

kddevin:
@trilinos/tpetra @csiefer2 @stanmoore1

## Motivation

@stanmoore1 reports errors on stria akin to those we saw on GPU systems.
The errors on GPU systems were avoided if we serialized one parallel scan using some Kokkos magic.  This code serializes the same scan for experimentation on stria -- no Kokkos magic needed.

#6345 

## Stakeholder Feedback

@csiefer2 @stanmoore1  
To try this code, add -DTPETRA_AVOID_PARALLEL_SCAN  to your CMAKE_CXX_FLAGS in Cmake.

## Testing

Tested on CEE LAN, with -DTPETRA_AVOID_PARALLEL_SCAN in CMAKE_CXX_FLAGS.
All Tpetra tests pass.

## Additional Information

I do not know whether or not this PR needs to be pushed to the develop branch.  Since this branch is for experimentation only, it would probably be better to simply build with this branch.

It will be very interesting, indeed, if serializing this scan fixes the issues on stria.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_computeOffsets.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_computeOffsets.hpp
@@ -279,6 +279,32 @@ computeOffsetsFromCounts (const ExecutionSpace& execSpace,
        ": counts.size() = " << numCounts << " >= ptr.size() = " <<
        numOffsets << ".");
 
+#ifdef TPETRA_AVOID_PARALLEL_SCAN
+    // KDD 11/25/19  Temporary workaround to the one scan whose behavior
+    // makes a difference for GPU runs.  On the GPU platforms, serializing
+    // this scan (via Kokkos' magic) makes the code perform correctly.  
+    // This code serializes the scan, without the Kokkos magic, for debugging
+    // on stria.  #6345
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_Details_computeOffsets.hpp
```diff
@@ -279,6 +279,32 @@ computeOffsetsFromCounts (const ExecutionSpace& execSpace,
        ": counts.size() = " << numCounts << " >= ptr.size() = " <<
        numOffsets << ".");
 
+#ifdef TPETRA_AVOID_PARALLEL_SCAN
```
mhoemmen: This really could have been a run-time switch.  I should have thought of that during yesterday's meeting.  That would have helped speed up debugging for the EMPIRE folks.

kddevin:
We haven't decided yet whether this workaround should be merged with develop or not.
Keeping the PR for now, but marking it WIP.

csiefer2:
This is something which will be actually useful for what I'm doing.  I'm going to merge this and then we can make this a run-time option later.

PR #6363, opened 2019-11-26
Ifpack2: Return of the s-step PR

csiefer2:
Auto-PR for SHA fa71b5e

### DIFF
--- a/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
+++ b/packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
@@ -66,6 +66,7 @@
 #include <Galeri_XpetraMatrixTypes.hpp>
 #endif
 
+#include <Tpetra_Details_residual.hpp>
 
 #include <Ifpack2_UnitTestHelpers.hpp>
 #include <Ifpack2_OverlappingRowMatrix.hpp>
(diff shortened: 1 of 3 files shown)


packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -109,6 +110,153 @@ typedef Tpetra::global_size_t GST;
 } while (false)
 
 
+
+
+/***********************************************************************************/
+template<class MatrixClass, class MultiVectorClass>
+void localReducedMatvec(const MatrixClass & A_lcl,
+                        const MultiVectorClass & X_lcl,
+                        const int userNumRows,
```
mhoemmen: `userNumRows` should have the same type as `A_lcl.numRows()`.  If `MatrixClass` is a `KokkosSparse::CrsMatrix`, then you can get the type from the `ordinal_type` typedef in `KokkosSparse::CrsMatrix`.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -109,6 +110,153 @@ typedef Tpetra::global_size_t GST;
...
+  int64_t rows_per_thread = -1;
+  
+  int64_t numLocalRows = userNumRows;
+  int64_t myNnz = A_lcl.nnz();
+
+  int64_t rows_per_team = 
+    Tpetra::Details::residual_launch_parameters<execution_space>(numLocalRows, myNnz, rows_per_thread, team_size, vector_length);
+  int64_t worksets = (X_lcl.extent (0) + rows_per_team - 1) / rows_per_team;
+
+  using policy_type = typename Kokkos::TeamPolicy<execution_space>;
```
mhoemmen: If you use `int64_t` for `worksets`, then you should also use `int64_t` as the index type for `Kokkos::TeamPolicy`.  (I doubt you need `int64_t`, though, since you're limited by the type of `A_lcl.numRows()` anyway.)

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
```diff
@@ -416,9 +565,86 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(Ifpack2OverlappingRowMatrix, getLocalDiag, Sca
...
+#endif
+    //    printf("Before matvec A is (locally)%dx%d x is of size %d, ovX is ov size %d\n",(int)A->getNodeNumRows(),(int)A->getNodeNumCols(),
+    //           (int)x.getMap()->getNodeNumElements(),(int)ovX.getMap()->getNodeNumElements());
+    //    printf("ovA->getUnderlyingMatrix() is (locally) %dx%d\n",(int)ovA.getUnderlyingMatrix()->getNodeNumRows(),(int)ovA.getUnderlyingMatrix()->getNodeNumCols());
+
+    reducedMatvec(ovA,ovX,2,temp1);
+    reducedMatvec(ovA,temp1,1,temp2);
+    reducedMatvec(ovA,temp2,0,ovY);
+
+    // And yes, that int cast is really necessary
```
mhoemmen: You could also just do `std::pair<int, int>(0, Y_lcl.extent(0))`.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
csiefer2: Au contraire!  The point of this kernel is that the user can multiply *fewer* rows that the whole matrix.

csiefer2:
> Why is it that we need new kernels in the unit test? This just looks like a mat-vec.

This is "test code" which will (hopefully later) be recycled into a correct implementation of a s-step Chebyshev method.

packages/ifpack2/test/unit_tests/Ifpack2_UnitTestOverlappingRowMatrix.cpp
mhoemmen: @csiefer2 Can't you just create a "subview" of the matrix that only views a subset of its rows, then call the existing kernel?

Issue #6346, opened 2019-11-25
Updating PR tester env


lucbv:
## Enhancement
@bartlettroscoe 

How do you feel about switching `atdm-ninja_fortran/1.7.2` for `sems-ninja_fortran/1.8.2`, it would be both an update of the ninja version and would also allow the scripts to run using only sems modules instead of a combination of sems and atdm modules.

Also I noticed that @ZUUL42 has been performing some tests to build with more up-to-date compilers, would it make sense to also have a build that uses the new `sems-openmpi/4.0.2` which is a significant upgrade from `sems-openmpi/1.10.1`?

bartlettroscoe:
> How do you feel about switching atdm-ninja_fortran/1.7.2 for sems-ninja_fortran/1.8.2?

@lucbv, fine by me but I don't have any say in the Trilinos PR builds.  The @trilinos/framework team needs to okay this.  Otherwise, you can go ahead and try making the change and posting a Trilinos PR.  I think we already switched most of the ATDM Trilinos builds to `sems-ninja_fortran/1.8.2`.

> would it make sense to also have a build that uses the new sems-openmpi/4.0.2 which is a significant upgrade from sems-openmpi/1.10.1?

Also fine by me but not my say.

Note that we will be updating the 'cee-rhel6' builds to OpenMPI 4.0.1 as per https://sems-atlassian-srn.sandia.gov/browse/ATDV-237.

PR #6366, opened 2019-11-27
Framework: Some cleanup and tweaks on PR testing script

william76:
@trilinos/framework 

This is WIP - I want to see what the Python PR testing shows since I haven't been able to replicate it locally.

Mostly just some style and formatting cleanup to make some things more readable, but I also made a more generic memory information option that tries to use psutil, and if it's not available then we try to load `/proc/meminfo` which would only work on Linux machines.

### DIFF
--- a/cmake/std/unittests/TestPullRequestLinuxDriverTest.py
+++ b/cmake/std/unittests/TestPullRequestLinuxDriverTest.py
@@ -715,8 +715,7 @@ class testCompute_n(unittest.TestCase):
         we should still  run at least one job'''
         m_open =  mock.mock_open(read_data='''MemTotal 32965846 kB''')
         with self.m_environ, \
-            mock.patch('PullRequestLinuxDriverTest.open',
-                       m_open, create=True), \
+            mock.patch('PullRequestLinuxDriverTest.open', m_open, create=True), \
             mock.patch('PullRequestLinuxDriverTest.cpu_count', return_value=18):
             parallel_level = PullRequestLinuxDriverTest.compute_n()
         self.assertEqual(10, parallel_level)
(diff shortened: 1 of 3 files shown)


lucbv:
@william76 any chance you can also take a look at the `cmake/std/sems/PullRequestGCCX.Y.ZTestingEnv.sh` a few of them could be modified to take advantage of  `sems-cmake/3.12.2` and `sems-ninja_fortran/1.8.2` instead of loading ATDM modules?

I've been talking about this in issue #6346

cmake/std/unittests/TestPullRequestLinuxDriverTest.py
```diff
@@ -73,7 +73,7 @@ def test_verifyGit_fails_with_old_version(self):
             m_check_out.return_value='git version 1.10.1'
 
             bad_git_string = 'Git version  should be 2.10 or better - Exiting!'
-            if sys.version_info.major is not 3:
+            if(sys.version_info.major != 3):
```
prwolfe: Why did you move away from python style to a c style of comparison? I always struggle with this one myself.

william76: In Python using `is` or `is not` is an identity test and is checking that the operands are the exact same object... where `==` and `!=` are equality checks that are using the output of the operands' `__eq__` and `__cmp__` methods.  I generally think of using `is` for singletons like `None` comparisons to see if something is exactly that thing and for value checks I tend to use equality/inequality operators.

In this case, I think using `is` is fine since sys.version_info.major is an `int` ... buuuut at least under Python 3.8 you'll get a warning using `is` instead of `!=`.  For example:
```
>>> sys.version_info.major is not 3
<stdin>:1: SyntaxWarning: "is not" with a literal. Did you mean "!="?
False
>>> sys.version_info.major != 3
False
>>>
```

I added the change to eliminate the warning.

prwolfe:
Interesting that they added a warning. The main reason I started to use is/is not came from pep8 complaints. I think the c-style =/!= is more readable.

Thanks!

Paul

PR #6376, opened 2019-11-28
MueLu: cleanup function interfaces in regionMG

mayrmt:
@trilinos/muelu 
@lucbv 

## Motivation

First cut at cleaning up function interfaces. In particular, this PR removes function arguments
- that haven't been used anyway.
- that are kind of redundant, because their content can be easily extracted from other arguments.

## Stakeholder Feedback

This is experimental code. No applications use it right now.

## Testing

The set of regionMG tests runs and passes.

## Additional Information

For sure, there's more to do. In particular, we could/should think about getting rid of the concept of "groups"...

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/main.cpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/main.cpp
@@ -915,8 +915,7 @@ int main_(int argc, char *argv[]) {
 //        printRegionalObject<Vector>("regB 1", regB, myRank, *fos);
 
         compRes = VectorFactory::Build(mapComp, true);
-        regionalToComposite(regRes, compRes,
-                            rowImportPerGrp, Xpetra::ADD);
+        regionalToComposite(regRes, compRes, rowImportPerGrp);
         typename Teuchos::ScalarTraits<Scalar>::magnitudeType normRes = compRes->norm2();
 
         // Output current residual norm to screen (on proc 0 only)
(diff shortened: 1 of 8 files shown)


packages/muelu/research/mmayr/composite_to_regions/src/main.cpp
```diff
@@ -935,7 +934,7 @@ int main_(int argc, char *argv[]) {
       /////////////////////////////////////////////////////////////////////////
 
 //      printRegionalObject<Vector>("regB 2", regB, myRank, *fos);
-      vCycle(0, numLevels, maxCoarseIter, maxRegPerProc,
```
lucbv: After this change `const int maxCoarseIter = 100;` on line 891 becomes unused and needs to be removed (-Werror does not like it)

packages/muelu/research/regionMG/examples/structured/Driver_Structured_Regions.cpp
```diff
@@ -839,7 +838,7 @@ int main_(Teuchos::CommandLineProcessor &clp, Xpetra::UnderlyingLib& lib, int ar
         /////////////////////////////////////////////////////////////////////////
 
         //      printRegionalObject<Vector>("regB 2", regB, myRank, *fos);
-        vCycle(0, numLevels, maxCoarseIter, maxRegPerProc,
```
lucbv: Same as previous comment, after this change `const int maxCoarseIter = 100;` on line 775 becomes unused.

PR #6377, opened 2019-11-30
Zoltan2: Fix view allocations for deprecated off (#6374)

### LINKED ISSUE #6358: Nalu application diffs due to new Trilinos/develop
## Bug Report
@trilinos/\<Zoltan2\>

### Description
I am seeing four new Nalu nightly regression tests diffs on gcc 7.2 and 8.3 due to a new Trilinos develop build:

git bisect bad 143ac5f761f187b98f79173b1670cb40c86e357e
git bisect good 0141ed90626699ba8402f52af9c25553bf947d67

The diffs in Nalu also seemingly show up on Nalu-Wind.

Once I have more details, I will report back. I just wanted to capture this to avid multiple developers performing a Trilinos bisect.

Best,

### LINKED ISSUE #6374: Zoltan2: Hang in MultiJagged
## Bug Report
@trilinos/zoltan2 

### Description
Some EMPIRE + MueLu runs on lassen are hanging in the rebalancing using Zoltan2 MultiJagged. Reverting https://github.com/trilinos/Trilinos/pull/5332 fixes the issue. Since I am not getting any output from Zoltan2: what can I do to make it easier to debug this?

MicheldeMessieres:
Removed an unused view and repair view allocations to indicate sizes according to dimension.

This resolves #6358 which could have errors or hang when KOKKOS_ENABLE_DEPRECATED_CODE was off.

@trilinos/zoltan2 

## Motivation

Resolve #6374

## Testing

Tested by the multijagged testes in zoltan2.

### DIFF
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -503,9 +503,6 @@ private:
   // if the target parts are uniform
   Kokkos::View<bool *, Kokkos::HostSpace> mj_uniform_parts;
 
-  // target part weight sizes.
-  Kokkos::View<mj_scalar_t **, device_t> mj_part_sizes;
-
   // if the coordinates have uniform weights
   Kokkos::View<bool *, Kokkos::HostSpace> mj_uniform_weights;
 
@@ -1009,8 +1006,6 @@ public:
    * \param mj_uniform_weights: if weight index [i] has uniform weight or not.
    * \param mj_weights: the two dimensional array for weights
    * \param mj_uniform_parts: if the target partitioning aims uniform parts
-   * \param mj_part_sizes: if the target partitioning does not aim uniform
-   * parts, then weight of each part.
    * \param result_assigned_part_ids: Output - the result partids corresponding
    * to the coordinates given im result_mj_gnos.
    * \param result_mj_gnos: Output - the result coordinate global id's
@@ -1034,7 +1029,6 @@ public:
     Kokkos::View<bool*, Kokkos::HostSpace> & mj_uniform_weights,
     Kokkos::View<mj_scalar_t**, device_t> & mj_weights,
     Kokkos::View<bool*, Kokkos::HostSpace> & mj_uniform_parts,
-    Kokkos::View<mj_scalar_t**, device_t> & mj_part_sizes,
     Kokkos::View<mj_part_t*, device_t> & result_assigned_part_ids,
     Kokkos::View<mj_gno_t*, device_t> & result_mj_gnos);
 
@@ -1575,13 +1569,11 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
   this->mj_uniform_weights(0) = true;
 
   this->mj_weights = Kokkos::View<mj_scalar_t**, device_t>
-    ("weights", 1);
+    ("weights", 1, 1);
 
   this->mj_uniform_parts =
     Kokkos::View<bool*, Kokkos::HostSpace>("uniform parts", 1);
   this->mj_uniform_parts(0) = true;
-
-  this->mj_part_sizes = Kokkos::View<mj_scalar_t**, device_t> ("part sizes", 1);
 
   this->set_part_specifications();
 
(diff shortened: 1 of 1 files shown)


bathmatt:
I posted a comment above but it shows pending, anyone know why it's pending??

MicheldeMessieres:
@bathmatt I don't see your comment. But I forgot to switch the PR against develop and it ran on master. That was resolved 2 hours ago and I set AT: RETEST so hoping that is sufficient to get the rerun. The switch to develop may be related to any issues you were seeing.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -1575,14 +1569,12 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
   this->mj_uniform_weights(0) = true;
 
   this->mj_weights = Kokkos::View<mj_scalar_t**, device_t>
-    ("weights", 1);
+    ("weights", 1, 1);
```
mhoemmen: 1. I'm pretty sure that building with deprecated code off would have caught this; please do so in the future.  Thanks!
2. Will these objects always be 1 x 1?

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -2076,7 +2068,7 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     this->part_xadj = this->new_part_xadj;
     this->host_part_xadj = Kokkos::create_mirror_view(part_xadj);
     Kokkos::deep_copy(host_part_xadj, part_xadj); // keep in sync
-    this->new_part_xadj = Kokkos::View<mj_lno_t*, device_t>("empty");
+    this->new_part_xadj = Kokkos::View<mj_lno_t*, device_t>("empty", 0);
```
mhoemmen: You don't need to allocate empty Views.  Just let them be default constructed.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -2604,9 +2596,9 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
   this->new_coordinate_permutations = Kokkos::View<mj_lno_t*, device_t>(
     Kokkos::ViewAllocateWithoutInitializing("num_local_coords"),
     this->num_local_coords);
-  // TODO empty is ok for NULL replacement?
+
```
mhoemmen: I'm not quite sure what the question means, but note that `view.data()` need not necessarily return `nullptr` if the view's dimensions are all zero.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Yes, not building with deprecated off was the start of my troubles and these errors were caught by deprecated off tests. Will do in the future. The weights will not always be 1x1.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Deprecated off won't allow the constructor with label and no dimensions: view_t("empty"), but will allow the default constructor: view_t() or with size: view_t("empty",0). The latter would provide the label for a debug out of bounds error which seems potentially useful. Wondering now if there might be a utility to allowing the constructor with label only.

@kddevin if we change to default constructors I could do it on a subsequent pass as I expect there will be further updates. Whether we keep the label does not seem critical.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: This was an old comment to myself during the refactor from raw ptrs to views - I should have deleted it sometime ago. I think I have ironed out all those issues and should be referencing just size in the logic. Thanks.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
mhoemmen: It's alright -- Trilinos really should have PR tests with deprecated code both on and off.  Not sure if that's the case yet.  Do you need those weights to be allocated initially?  If not, you could just leave them default constructed.

PR #6378, opened 2019-12-01
Update cdash_analyze_and_report.py and gitdist-setup.sh

bartlettroscoe:
The main update from this snapshot is the updated gitdist-setup.sh script that automatically sets and alias to the gitdist file in the same directory.  Makes it super easy to get setup to use gitidst (just a single  source command).  This brings this into Trilinos.

The update to cdash_analyze_and_report.py is to support `--date=today` and `--date=yesterday` which are the most common use cases.

### DIFF
--- a/cmake/tribits/ci_support/cdash_build_testing_date.py
+++ b/cmake/tribits/ci_support/cdash_build_testing_date.py
@@ -74,8 +74,8 @@ def injectCmndLineOptionsInParser(clp, gitoliteRootDefault=""):
   
   clp.add_option(
     "--cdash-project-start-time", dest="cdashProjectStartTimeUtcStr", type="string", default="",
-    help="Starting time for the CDash testing day in 'HH:MM' in UTC."\
-      + " Check the CDash project settings for the testing date." )
+    help="Starting time for the CDash testing day in 'hh:mmm' in UTC."\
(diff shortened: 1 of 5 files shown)


cmake/tribits/ci_support/cdash_build_testing_date.py
```diff
@@ -74,8 +74,8 @@ def injectCmndLineOptionsInParser(clp, gitoliteRootDefault=""):
   
   clp.add_option(
     "--cdash-project-start-time", dest="cdashProjectStartTimeUtcStr", type="string", default="",
-    help="Starting time for the CDash testing day in 'HH:MM' in UTC."\
-      + " Check the CDash project settings for the testing date." )
+    help="Starting time for the CDash testing day in 'hh:mmm' in UTC."\
```
rppawlo: 'hh:mmm' -> 'hh:mm'

bartlettroscoe: @rppawlo, thanks for the review!  I just fixed this in the TriBITS 'mastesr' commit https://github.com/TriBITSPub/TriBITS/commit/aaa30da4526b07948d6efef987e676aa8cf85787.  The next time I snapshot TriBITS into Trilinos it will get updated there there.

PR #6380, opened 2019-12-02
Fixes to Amesos2_MUMPS interface

chochmuth:
@trilinos/amesos2 

## Motivation
I made some changes to the MUMPS interface to fix the following two issues:
1. Mumps data was not freed correct which resulted in memory leaks.
2. Input matrix was only converted to mumps compatible data when the internal variable was `MUMPS_MATRIX_LOAD=false`. As a consequence, the data was only correct if the solver was constructed for the first time. It was not possible to reuse the symbolic factorization of a previous matrix with the same pattern and only call the numeric factorization.

Furthermore, I added the error messages of the Amesos_MUMPS interface to the Amesos2_MUMPS interface.

## Testing
The previous four test still work. Additionally, I added a fifth test where the numeric factorization and the solve is preformed twice without resetting the solver object.

### DIFF
--- a/packages/amesos2/src/Amesos2_MUMPS_decl.hpp
+++ b/packages/amesos2/src/Amesos2_MUMPS_decl.hpp
@@ -205,6 +205,7 @@ private:
 
   bool MUMPS_MATRIX_LOAD;
   bool MUMPS_STRUCT;
+  mutable bool MUMPS_MATRIX_LOAD_PREORDERING;
 
   // The following Arrays are persisting storage arrays for A, X, and B
   /// Stores the values of the nonzero entries for MUMPS
--- a/packages/amesos2/src/Amesos2_MUMPS_def.hpp
+++ b/packages/amesos2/src/Amesos2_MUMPS_def.hpp
@@ -85,7 +85,8 @@ namespace Amesos2
     
     MUMPS_MATRIX_LOAD = false;
     MUMPS_STRUCT = false;
-    
+    MUMPS_MATRIX_LOAD_PREORDERING = false;
+      
     #ifdef HAVE_MPI
     using Teuchos::Comm;
     using Teuchos::MpiComm;
@@ -161,6 +162,11 @@ namespace Amesos2
         free(mumps_par.jcn);
         free(mumps_par.irn);
       }
+      mumps_par.job = -2;
+      if (this->rank_ < this->nprocs_) {
+          function_map::mumps_c(&(mumps_par));
+      }
+
   }
   
   template<class Matrix, class Vector>
@@ -290,7 +296,8 @@ namespace Amesos2
             as<size_t>(ld_rhs),
             CONTIGUOUS_AND_ROOTED);
     }
-    
+    // ch: see function loadA_impl()
+    MUMPS_MATRIX_LOAD_PREORDERING = false;
     return(ierr);
   }//end solve()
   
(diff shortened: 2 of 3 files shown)


packages/amesos2/src/Amesos2_MUMPS_decl.hpp
```diff
@@ -205,6 +205,7 @@ class MUMPS : public SolverCore<Amesos2::MUMPS, Matrix, Vector>
 
   bool MUMPS_MATRIX_LOAD;
   bool MUMPS_STRUCT;
+  mutable bool MUMPS_MATRIX_LOAD_PREORDERING;
```
mhoemmen: "const is a lie"

packages/amesos2/src/Amesos2_MUMPS_def.hpp
```diff
@@ -496,12 +502,25 @@ namespace Amesos2
...
+        std::cerr << "Amesos_Mumps : INFOG(1) = " << mumps_par.infog[0] << std::endl;
+        std::cerr << "Amesos_Mumps : INFOG(2) = " << mumps_par.infog[1] << std::endl;
+      }
+      if (mumps_par.info[0] != 0  && Wrong) {
+        std::cerr << "Amesos_Mumps : On process " << this->matrixA_->getComm()->getRank()
+        << ", INFO(1) = " << mumps_par.info[0] << std::endl;
+        std::cerr << "Amesos_Mumps : On process " << this->matrixA_->getComm()->getRank()
+        << ", INFO(2) = " << mumps_par.info[1] << std::endl;
+      }
+      if (this->rank_==0) {
```
mhoemmen: Amesos2 should throw on all processes, if it expects all processes to know about the error.  Otherwise, some processes may continue on, which could result in a hang.  Do you consider this error recoverable?

packages/amesos2/test/solvers/MUMPS_UnitTests.cpp
```diff
@@ -322,6 +322,79 @@ namespace {
     TEST_COMPARE_FLOATING_ARRAYS( xhatnorms, xnorms, 0.005 );
   }
 
+  TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL( MUMPS, SolveTwice, SCALAR, LO, GO )
```
mhoemmen: It's supposed to be possible to build Amesos2 without Tpetra support, so you might need to protect this unit test with an appropriate macro, depending on the CMakeLists.txt settings.

packages/amesos2/src/Amesos2_MUMPS_decl.hpp
chochmuth: @mhoemmen Thanks for pointing this out. 
To 1: I just added another subtest wich `TEUCHOS_UNIT_TEST_TEMPLATE_3_INSTANT( MUMPS, SolveTwice, SCALAR, LO, GO )` and not a whole separated test. I assume that the whole test was protected before.

packages/amesos2/src/Amesos2_MUMPS_def.hpp
chochmuth: You are right, the error is now broadcasted so every rank can throw.

packages/amesos2/src/Amesos2_MUMPS_decl.hpp
mhoemmen: @chochmuth You could try explicitly disabling Tpetra (`Trilinos_ENABLE_Tpetra:BOOL=OFF`) and enabling Epetra and Amesos2, just to see if it builds.  Amesos2 does claim to support this use case, so it's important to test.

packages/amesos2/test/solvers/MUMPS_UnitTests.cpp
mhoemmen: We discussed this more above.

srajama1:
> Amesos2 does claim to support this use case, so it's important to test.

It is the opposite. We use Tpetra always. We allow to turn off Epetra.

PR #6385, opened 2019-12-03
Zoltan2: Fix precision issues causing poor load balancing in some cas…

MicheldeMessieres:
…es (#6358)

@trilinos/zoltan2

## Motivation

Fix diffs seen in regression tests - issue #6358

## Testing

Tested in Zoltan2 multijagged tests though we need to expand those tests to better capture these issues.

### DIFF
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -3060,24 +3060,24 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     }
   }
 
-  Kokkos::View<mj_part_t*, device_t> device_cumulative(
+  Kokkos::View<mj_scalar_t*, device_t> device_cumulative(
     "device_cumulative", num_cuts);
   typename decltype(device_cumulative)::HostMirror
     host_cumulative = Kokkos::create_mirror_view(device_cumulative);
 
-  mj_part_t cumulative = 0;
+  mj_scalar_t cumulative = 0;
 
   if(bUniformTargetWeights) {
     // How many total future parts the part will be partitioned into.
     mj_scalar_t total_future_part_count_in_part =
-      mj_scalar_t((*future_num_part_in_parts)[concurrent_current_part]);
+      static_cast<mj_scalar_t>((*future_num_part_in_parts)[concurrent_current_part]);
 
     // How much each part should weigh in ideal case.
     mj_scalar_t unit_part_weight =
       global_weight / total_future_part_count_in_part;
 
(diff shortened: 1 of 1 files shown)


packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -3060,24 +3060,24 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
     }
   }
 
-  Kokkos::View<mj_part_t*, device_t> device_cumulative(
+  Kokkos::View<mj_scalar_t*, device_t> device_cumulative(
     "device_cumulative", num_cuts);
   typename decltype(device_cumulative)::HostMirror
```
mhoemmen: Consider for the future: If you're using `decltype` anyway, you might as well use `auto` here.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -3109,13 +3109,17 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
...
-  Kokkos::parallel_for(
-    Kokkos::RangePolicy<typename mj_node_t::execution_space, mj_part_t>
-      (0, num_cuts + 1),
-    KOKKOS_LAMBDA (mj_part_t i) {
-    current_target_part_weights(i) =
-      long(current_target_part_weights(i) + 0.5);
-  });
+  // Note need to discuss regarding DragonFly commits and determine if we
+  // would not simply check mj_uniform_weights here.
+  if (!bUniformPartsCheck || this->mj_uniform_weights[0]) {
```
johelli: This looks good to me for nonuniform partitioning. It may have been rounding unnecessarily previously.

kddevin:
@MicheldeMessieres 
If you build with 
-D MPI_EXEC_MAX_NUMPROCS:STRING=64 
you'll run a broader range of tests in Zoltan2.  
Test sacer4320 runs on eight ranks, instead of the PR tester's four ranks.  
It may exercise this issue.

MicheldeMessieres:
@kddevin Quite right. I confirmed sacer4320 on 8 ranks fails prior to the fix.

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Thanks. I did not ever consider that and it really cleans things up. I'll make those changes in a different PR.

mhoemmen:
@rrdrake @sebrowne FYI this merged into Trilinos develop.

Issue #6295, opened 2019-11-18
Create GCC 8.3 PR Testing Build


ZUUL42:
## Enhancement
@trilinos/framework

Look into creating a PR Testing build that uses GCC 8.3.
Currently we have a GCC 7.2 PR testing build.
GCC 8.3 would likely replace the 7.2 build.

ZUUL42:
A ticket has been submitted with SEMS to add a GCC 8.3 module.

Testing will look at how building with the same Werror flags the 7.2.0 build uses currently.

bartlettroscoe:
Who is the customer for this?

mhoemmen:
@bartlettroscoe Sierra would like a GCC 8.3 build, I think.  @rrdrake can comment more, but I think the reason is that GCC 8.3 has different (probably more) warnings.

rrdrake:
The warnings from a newer GCC version is one reason, but also important is Astra (the Vanguard series of ARM machines) uses GCC 8 headers by default.

ZUUL42:
[CDash report](https://testing.sandia.gov/cdash/index.php?project=Trilinos&parentid=5952912&filtercount=1&showfilters=1&field1=buildstamp&compare1=61&value1=20191123-0034-Experimental&filtercombine=and) from a run with GCC 8.3 @ Werror flags set as they have been in 7.2.0.
There are a number of Werror flags that are causing build errors.
@trilinos/muelu @trilinos/intrepid @trilinos/stk @trilinos/stokhos @trilinos/amesos2 @trilinos/zoltan2 @trilinos/ml @trilinos/panzer @trilinos/tpetra 
[gcc_8.3 branch](https://github.com/trilinos/Trilinos/compare/develop...ZUUL42:gcc_8.3)

etphipp:
In the stokhos case all of the warnings/errors are coming from boost:
```
/projects/sems/install/rhel7-x86_64/semsboost/1.63.0/gcc/8.3.0/base/include/boost/mpl/assert.hpp:188:21: error: unnecessary parentheses in declaration of ‘assert_arg’ [-Werror=parentheses]
 failed ************ (Pred::************
                     ^
/projects/sems/install/rhel7-x86_64/semsboost/1.63.0/gcc/8.3.0/base/include/boost/mpl/assert.hpp:193:21: error: unnecessary parentheses in declaration of ‘assert_not_arg’ [-Werror=parentheses]
 failed ************ (boost::mpl::not_<Pred>::************
[...]

PR #6394, opened 2019-12-04
Tpetra: Fix #6393

### LINKED ISSUE #6393: Tpetra: Purge use of memcpy with Kokkos::complex
## Bug Report
@trilinos/tpetra 

### Description

`Kokkos::complex<T>` [is not TriviallyCopyable](https://github.com/kokkos/kokkos/issues/2577), so it's not correct to use in `memcpy`.  GCC 8.3 [issues a warning](https://github.com/trilinos/Trilinos/issues/6295#issuecomment-558323602) due to Tpetra code that invokes `memcpy` with `Kokkos::complex<T>` pointers.  (Such pointers usually show up in generic Tpetra code as `impl_scalar_type*`.)

mhoemmen:
@trilinos/tpetra

Using `memcpy` with `Kokkos::complex<T>` pointers causes GCC 8.3 to issue warnings, since `Kokkos::complex<T>` is not `TriviallyCopyable`.  This commit removes use of `memcpy` with `Kokkos::complex<T>` pointers from Tpetra.

## Related Issues

* Closes https://github.com/trilinos/Trilinos/issues/6393
* Related to https://github.com/kokkos/kokkos/issues/2577
* Part of https://github.com/trilinos/Trilinos/issues/6295

## Testing info

Tested with GCC 4.8.4 + OpenMP on Linux.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
@@ -61,6 +61,7 @@
 #include "Tpetra_Details_gathervPrint.hpp"
 #include "Tpetra_Details_getDiagCopyWithoutOffsets.hpp"
 #include "Tpetra_Details_leftScaleLocalCrsMatrix.hpp"
+#include "Tpetra_Details_local_deep_copy.hpp"
 #include "Tpetra_Details_Profiling.hpp"
 #include "Tpetra_Details_rightScaleLocalCrsMatrix.hpp"
 #include "Tpetra_Details_ScalarViewTraits.hpp"
@@ -2072,9 +2073,7 @@ namespace Tpetra {
         (tgt_IST, numInserted);
       View<const IST*, HostSpace, MemoryTraits<Unmanaged>> src
         (src_IST, numInserted);
-      // NOTE (mfh 04 Dec 2019) Use local_deep_copy, not deep_copy,
-      // since this copy may be too small to justify parallelization.
-      using Kokkos::Experimental::local_deep_copy;
+      using Tpetra::Details::local_deep_copy;
       local_deep_copy (tgt, src);
     }
   }
--- a/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
@@ -2056,16 +2056,26 @@ namespace Tpetra {
                           const ELocalGlobal I)
   {
     const size_t oldNumEnt = rowInfo.numEntries;
-    const size_t numInserted = graph.insertIndices (rowInfo, newInds, lg, I);
-
-    // Use of memcpy here works around an issue with GCC >= 4.9.0,
-    // that probably relates to scalar_type vs. impl_scalar_type
-    // aliasing.  See history of Tpetra_CrsGraph_def.hpp for
-    // details; look for GCC_WORKAROUND macro definition.
+    const size_t numInserted =
+      graph.insertIndices (rowInfo, newInds, lg, I);
     if (numInserted > 0) {
+      using Kokkos::HostSpace;
+      using Kokkos::MemoryTraits;
+      using Kokkos::Unmanaged;
+      using Kokkos::View;
+      using IST = impl_scalar_type;
+
       const size_t startOffset = oldNumEnt;
-      memcpy (&oldRowVals[startOffset], &newRowVals[0],
-              numInserted * sizeof (impl_scalar_type));
+      IST* tgt_IST = oldRowVals.getRawPtr () + startOffset;
+      const IST* src_IST = newRowVals.getRawPtr ();
+      View<IST*, HostSpace, MemoryTraits<Unmanaged>> tgt
+        (tgt_IST, numInserted);
+      View<const IST*, HostSpace, MemoryTraits<Unmanaged>> src
+        (src_IST, numInserted);
+      // NOTE (mfh 04 Dec 2019) Use local_deep_copy, not deep_copy,
+      // since this copy may be too small to justify parallelization.
+      using Kokkos::Experimental::local_deep_copy;
+      local_deep_copy (tgt, src);
     }
   }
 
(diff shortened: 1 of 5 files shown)


packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -2056,16 +2056,26 @@ namespace Tpetra {
...
+      IST* tgt_IST = oldRowVals.getRawPtr () + startOffset;
+      const IST* src_IST = newRowVals.getRawPtr ();
+      View<IST*, HostSpace, MemoryTraits<Unmanaged>> tgt
+        (tgt_IST, numInserted);
+      View<const IST*, HostSpace, MemoryTraits<Unmanaged>> src
+        (src_IST, numInserted);
+      // NOTE (mfh 04 Dec 2019) Use local_deep_copy, not deep_copy,
+      // since this copy may be too small to justify parallelization.
+      using Kokkos::Experimental::local_deep_copy;
+      local_deep_copy (tgt, src);
```
kddevin: Why doesn't the reinterpret_cast trick work here?  Creating new views just to copy src to tgt is non-intuitive.

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -133,20 +132,26 @@ pad_crs_arrays(
...
     {
-      auto indices_old_subview = subview(indices, range(this_row_beg, this_row_beg+used_this_row));
-      auto indices_new_subview = subview(indices_new, range(row_ptr_beg(i), row_ptr_beg(i)+used_this_row));
-      // just call memcpy; it works fine on device if this becomes a kernel
-      memcpy(indices_new_subview.data(), indices_old_subview.data(), used_this_row * sizeof(inds_value_type));
+      auto indices_old_subview = subview(indices,
+        range(this_row_beg, this_row_beg+used_this_row));
+      auto indices_new_subview = subview(indices_new,
+        range(row_ptr_beg(i), row_ptr_beg(i)+used_this_row));
+      using Kokkos::Experimental::local_deep_copy;
```
kddevin: I'm not excited about relying on Kokkos::Experimental.  
What is Kokkos' plan for local_deep_copy?

kddevin:
@mhoemmen  Would you also please add details on how you tested this change (e.g., "compilation with gcc 8.3 on cee-lan")?  For software quality assessments, we want to record our testing info in each PR.  Thanks.

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
mhoemmen: @kddevin Christian tells me that it will go into the Kokkos namespace in the next Kokkos release.  There will also be team versions.

mhoemmen:
@kddevin wrote:
> Would you also please add details on how you tested this change (e.g., "compilation with gcc 8.3 on cee-lan")?

I edited the original description yesterday with testing info, but not sure if you got the update -- thanks!

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
kddevin: OK; I think this section should use the simplest approach (reinterpret_cast) instead of this approach (creating views for non-view data).  Just because Christian says we CAN use local_deep_copy doesn't mean we SHOULD use it.  :-)

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
kddevin: OK; I'll add a Kokkos issue requesting the move and link it back to this issue.

kddevin:
> @kddevin wrote:
> 
> > Would you also please add details on how you tested this change (e.g., "compilation with gcc 8.3 on cee-lan")?
> 
> I edited the original description yesterday with testing info, but not sure if you got the update -- thanks!

Probably should test with gcc8.3, since the original issue arose with gcc 8.3.
On cee lan, the following worked for me in #6397.
```
module load sierra-devel/gcc-8.3.0-openmpi-1.10.2
```

mhoemmen:
@kddevin I just pushed a new commit.  I added a `Tpetra::Details::local_deep_copy` function that can go away, once Kokkos' `local_deep_copy` moves from the `Kokkos::Experimental` namespace to `Kokkos`.  For now, `Tpetra::Details::local_deep_copy` replicates the subset of Kokkos' functionality that Tpetra needs.  I added this function for three reasons:

  1. It avoids use of `memcpy` for certain Scalar types (like Sacado's DFad types), thus helping move Tpetra slowly towards one of its original design goals of genericity.
  2. It prevents any possible issues of `Kokkos::complex` and `std::complex` having differing alignment requirements on different platforms.  (Note that `std::complex<double>` need only be aligned to 8 bytes, not 16 bytes.  On CUDA, `complex<T>` values must be accessed with sizeof(T)-byte alignment.  This is one reason why CUDA has its own complex type.)
  3. It encourages Tpetra developers to move to `Kokkos::local_deep_copy`, once it becomes available.

@trilinos/tpetra

mhoemmen:
Weird -- apparently GCC 4.8.4 is not a complete C++11 implementation.  https://en.cppreference.com/w/cpp/types/is_trivially_copyable

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
kddevin: I still don't understand why we need to create views here, just to copy values.  Why doesn't the reinterpret_cast trick work here?

packages/tpetra/core/src/Tpetra_Details_local_deep_copy.hpp
```diff
@@ -0,0 +1,117 @@
+// @HEADER
```
kddevin: How does this reimplementation differ from Kokkos::Experimental::local_deep_copy?
My comments about the Experimental namespace were not intended to create a reimplementation that we need to maintain, but, rather, to encourage Kokkos to promote local_deep_copy quickly.

mhoemmen:
@kddevin wrote:
> I still don't understand why we need to create views here, just to copy values. Why doesn't the reinterpret_cast trick work here?

It works, but it encourages bad behavior to use `memcpy` for types for which that may not be valid.  If you want to fix that in a separate PR, that's fine.

mhoemmen:
Approach was rejected.

kddevin:
Not rejected; I just had questions about it.   We can discuss in person if you prefer.

mhoemmen:
@kddevin wrote:
> Not rejected; I just had questions about it. We can discuss in person if you prefer.

It's OK; I'm not offended -- I just don't have time to work more on this, and I'd rather close the PR so it doesn't consume PR testing resources.

PR #6388, opened 2019-12-03
Tpetra/MueLu/Ifpack2/Xpetra: remove all deprecated code (including DynamicProfile)

brian-kelley:
@trilinos/tpetra 
@trilinos/muelu 
@trilinos/ifpack2 
@trilinos/xpetra 

## Motivation

Removing all deprecated code from the core of the Tpetra stack. Much of this is related to the removal of DynamicProfile from Tpetra.

The motivation for that is to reduce the number of code paths in CrsGraph and CrsMatrix, and to provide opportunities for performance improvements that can't be done with the Teuchos array-of-arrays storage format. In StaticProfile, entries are stored in a 1D Kokkos::View with fixed storage allocated per row, much like a fill-complete matrix.

## Related Issues

* Closes #5602
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to  
* Part of 
* Composed of 

## Stakeholder Feedback

EMPIRE has already been building Trilinos successfully with deprecated code disabled.

## Testing

This change touches tests in just about every package downstream of Tpetra, but since this PR is about removing code, there are no new features that need to be tested.

PR #6400, opened 2019-12-05
Zoltan2:  allow test using zoltan to pass with default long long global_ordinal type

kddevin:
@trilinos/zoltan2 

## Motivation
Trilinos' default global ordinal type will soon switch to long long.

Test zoltanCompare compares Zoltan2 results with Zoltan results for PHG, RCB, RIB, etc.
Since PHG hashes global ordinals, all global ordinals must be passed to the library in the same format.  If Zoltan2 uses global ordinal = long long, Zoltan must use NUM_GID_ENTRIES=2 and pass each GID in an array of two ZOLTAN_ID_TYPE.

Previously, this test assumed global ordinal = int.  These changes allow it to work with global ordinal = int or long long.

Changes were required only in the test file; no changes to the Zoltan or Zoltan2 libraries were needed.

#6388 #5602 

## Stakeholder Feedback

N/A

## Testing

Tested on cee lan with gcc 7.3 and gcc 8.3, with and without Tpetra_ENABLE_DEPRECATED_CODE (that is, with default global ordinal = int and long long, respectively)

### DIFF
--- a/packages/zoltan2/test/correctness/zoltanCompare.cpp
+++ b/packages/zoltan2/test/correctness/zoltanCompare.cpp
@@ -54,6 +54,7 @@
 #include <Zoltan2_PartitioningProblem.hpp>
 
 #include <Tpetra_MultiVector.hpp>
+#include <zoltan.h>
 
 using Teuchos::RCP;
 using Teuchos::rcp;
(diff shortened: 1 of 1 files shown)


packages/zoltan2/test/correctness/zoltanCompare.cpp
```diff
@@ -119,6 +120,15 @@ typedef Zoltan2::XpetraMultiVectorAdapter<tMVector_t> vectorAdapter_t;
 typedef Zoltan2::XpetraCrsMatrixAdapter<tMatrix_t,tMVector_t> matrixAdapter_t;
 typedef Zoltan2::EvaluatePartition<matrixAdapter_t> quality_t;
 
+// Number of ZOLTAN_ID in a zgno_t (for NUM_GID_ENTRIES)
+static constexpr int nGidEnt = sizeof(zgno_t) / sizeof(ZOLTAN_ID_TYPE);
+#define SET_ZID(n,a,b)                                            \
```
mhoemmen: For the future, best practice would be to capture the values of `n`, `a`, and `b` above the loop using `auto`:
```c++
const auto zoltan_zid_loop_n = n;
auto zoltan_zid_loop_a = a;
auto zoltan_zid_loop_b = b;
for (ZOLTAN_ID_LOOP = 0; ZOLTAN_ID_LOOP < n; ++ZOLTAN_ID_LOOP) {
  zoltan_zid_loop_a[ZOLTAN_ID_LOOP] = zoltan_zid_loop_b[ZOLTAN_ID_LOOP];
}
```

packages/zoltan2/test/correctness/zoltanCompare.cpp
```diff
@@ -135,14 +145,18 @@ static void zobjlist(void *data, int ngid, int nlid,
 {
   *ierr = ZOLTAN_OK;
   tMVector_t *vec = (tMVector_t *) data;
-  int n = vec->getLocalLength();
-  for (int i = 0; i < n; i++) {
-    gids[i] = vec->getMap()->getGlobalElement(i);
+  size_t n = vec->getLocalLength();
+
+  for (size_t i = 0; i < n; i++) {
+    zgno_t vgid = vec->getMap()->getGlobalElement(i);
```
mhoemmen: This may issue a warning, since it casts `i` from unsigned `size_t` to signed `LO`.

PR #6405, opened 2019-12-07
Tpetra: explicitly sort entries of add() result, make setAllIndices() expect sorted.

brian-kelley:
@trilinos/tpetra 
@cgcgcg 

## Motivation

TpetraExt::MatrixMatrix::add() was producing a fill-complete result matrix with unsorted local entries (invalid). It was doing setAllValues() followed by fillComplete(), but the fillComplete was not sorting indices because the graph's isSorted_ flag was still true.

- Fixed the actual problem in add() by sorting indices before calling setAllValues()
- Updated the documentation for CrsGraph::setAllIndices() and CrsMatrix::setAllValues() to say that the indices must be sorted on input, because they do not check for sorted
- setAllIndices() now sets isSorted_ and noRedundancies_ to true, since neither sorting nor merging will be done during fillComplete(). This was already happening in all the CrsGraph constructors that call setAllIndices().

## Stakeholder Feedback

Fixes a bug exposed by EMPIRE.
## Testing

add() is tested in Tpetra and Xpetra, and used in many MueLu tests.

### DIFF
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
@@ -824,15 +824,7 @@ add (const Scalar& alpha,
     //TODO: use KokkosKernels batched sort on device as soon as it's available
     //But now, have to sort on host (using UVM)
     exec_space().fence();
-    for(LO row = 0; row < numLocalRows; row++)
-    {
-      size_type rowBegin = rowptrs(row);
-      size_type rowEnd = rowptrs(row + 1);
-      lno_t* entriesBegin = &localColinds(rowBegin);
-      lno_t* entriesEnd = &localColinds(rowEnd);
-      typename AddKern::impl_scalar_type* valuesBegin = &vals(rowBegin);
-      Tpetra::sort2(entriesBegin, entriesEnd, valuesBegin);
-    }
+    Tpetra::Import_Util::sortCrsEntries(rowptrs, localColinds, vals);
     C.setAllValues(rowptrs, localColinds, vals);
     C.fillComplete(CDomainMap, CRangeMap, params);
     if(!doFillComplete)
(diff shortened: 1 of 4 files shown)


packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -817,6 +821,18 @@ add (const Scalar& alpha,
                               col_inds_array, global_col_inds_array,
                               typename map_type::local_map_type>
         (localColinds, globalColinds, CcolMap->getLocalMap()));
+    //TODO: use KokkosKernels batched sort on device as soon as it's available
```
lucbv: Why not use the function `sortCrsEntries` from `Tpetra_Import_Util2.hpp`?
It would avoid redundant code and potentially bugs...

brian-kelley: @lucbv Thanks, it does do exactly what I want. Fixed.

rppawlo:
@brian-kelley - is there a test or a debug mode check you could add that will catch this from breaking again in future refactors?

brian-kelley:
@rppawlo I'm adding checks to the MatrixMatrix tests that entries are sorted after fillComplete().
@mhoemmen, would it be too expensive to check that the input to CrsGraph::setAllIndices() is sorted, inside a ``#ifdef HAVE_TPETRA_DEBUG``? That could possibly catch other bugs down the road.

mhoemmen:
@brian-kelley wrote:
> @mhoemmen, would it be too expensive to check that the input to CrsGraph::setAllIndices() is sorted, inside a `#ifdef HAVE_TPETRA_DEBUG`?

It's not too expensive for debug mode -- we already do checks of comparable cost elsewhere in debug mode.  Please use the run-time test `Tpetra::Details::Behavior::debug()`, not the macro.  Thanks!

mhoemmen:
@brian-kelley Regarding the check: Best practice would be to do an all-reduce to make sure that all processes pass the check.  Otherwise, it's annoying to debug a throw on Process 4099 ;-) .  Consider using `Tpetra::Details::gathervPrint` to gather and print error messages from all processes.

brian-kelley:
Better test coverage and the HAVE_TPETRA_DEBUG checks that would have caught this are now in #6433.

PR #6410, opened 2019-12-09
Tpetra: document the requirements of column maps

brian-kelley:
At CrsMatrix/CrsGraph constructors and replaceColMap(),
document the GID ordering requirements of column maps:
-locals before remotes (determined by domain map)
-locals sorted
-remotes sorted first by proc and then by GID

Strongly recommending that Details::makeColMap() be
used instead of constructing it manually from raw GIDs.

Also, fix a typo in a makeColMap comment.

@trilinos/tpetra 

## Motivation

The requirements of column maps need to be codified, especially as more matrix/graph construction stuff gets rewritten using Kokkos. Having these comments here would have made #6190 a lot easier, for example.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows #6190  
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing

Just documentation (doxygen/comment) changes.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
@@ -1548,16 +1548,12 @@ namespace Tpetra {
     /// \param newColMap [in] New column Map.  Must be nonnull.
     ///   Within Tpetra, there are no particular restrictions on the column map.
     ///   However, if this graph will be used in Xpetra, Ifpack2, or MueLu,
-    ///   the column map's list of global indices on each process must be in a certain order:
-    ///   locally owned GIDs (sorted), followed by remote GIDs (in order of
-    ///   owning proc, and sorted within each proc).
+    ///   the column map's list of global indices must follow "Aztec ordering":
+    ///   locally owned GIDs (same order as in domain map), followed by remote GIDs
+    ///   (in order of owning proc, and sorted within each proc).
     ///
-    ///   If in doubt, it is strongly recommended to use Tpetra::Details::makeColMap()
-    ///   to create the column map. makeColMap() takes a domain map
-    ///   that determines which columns are locally owned,
-    ///   and the RowGraph whose globally indexed entries
-    ///   must be mapped to local indices. Alternately, it can take
-    ///   a domain map and a raw Kokkos::View of global indices to be mapped.
+    ///   It is strongly recommended to use \c Tpetra::Details::makeColMap()
+    ///   to create the column map. makeColMap() follows Aztec ordering by default.
     void replaceColMap (const Teuchos::RCP<const map_type>& newColMap);
 
     /// \brief Reindex the column indices in place, and replace the
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
@@ -1546,11 +1546,13 @@ namespace Tpetra {
     /// does their order change, as a result of calling this method.
     ///
     /// \param newColMap [in] New column Map.  Must be nonnull.
-    ///   The column map's list of global indices on each process must be in a certain order:
+    ///   Within Tpetra, there are no particular restrictions on the column map.
+    ///   However, if this graph will be used in Xpetra, Ifpack2, or MueLu,
+    ///   the column map's list of global indices on each process must be in a certain order:
     ///   locally owned GIDs (sorted), followed by remote GIDs (in order of
     ///   owning proc, and sorted within each proc).
     ///
-    ///   It is strongly recommended to use Tpetra::Details::makeColMap()
+    ///   If in doubt, it is strongly recommended to use Tpetra::Details::makeColMap()
     ///   to create the column map. makeColMap() takes a domain map
     ///   that determines which columns are locally owned,
     ///   and the RowGraph whose globally indexed entries
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
@@ -411,6 +411,7 @@ namespace Tpetra {
     /// \param rowMap [in] Distribution of rows of the graph.
     ///
     /// \param colMap [in] Distribution of columns of the graph.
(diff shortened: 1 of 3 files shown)


mhoemmen:
@brian-kelley Those are not actually requirements of CrsGraph or CrsMatrix.  Both of those classes work just fine with arbitrary column Maps.  We've exercised those cases in Nalu, for example.  The issue is that certain preconditioners have those requirements.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -411,6 +411,7 @@ namespace Tpetra {
     /// \param rowMap [in] Distribution of rows of the graph.
     ///
     /// \param colMap [in] Distribution of columns of the graph.
+    ///   See replaceColMap() for the requirements.
```
mhoemmen: This is not correct.  CrsGraph itself can work with any column Map.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -1540,6 +1546,16 @@ namespace Tpetra {
     /// does their order change, as a result of calling this method.
     ///
     /// \param newColMap [in] New column Map.  Must be nonnull.
+    ///   The column map's list of global indices on each process must be in a certain order:
```
mhoemmen: This is not correct.  CrsGraph itself can work with any column Map.

brian-kelley:
@mhoemmen OK, maybe this isn't the right place to put these comments/requirements. Basically, if somebody comes along later to do what I tried to do with the add (which is to build the column map from scratch for efficiency), they should know what the de facto requirements are for stuff to work downstream. It took several tries for me to get #6190 to work because I didn't understand what was causing MueLu/Xpetra/Ifpack2 tests to fail. So I would still like to codify them somewhere that's easy to find.

@jhux2 @csiefer2 @cgcgcg @lucbv Do you have any thoughts on this, about where column map ordering requirements could be documented?

mhoemmen:
@brian-kelley I think it's fine to put the requirements in CrsGraph's documentation, but they are not CrsGraph requirements.  We could say something like "some downstream packages, including preconditioners in Ifpack2 and MueLu, have the following requirements on column Maps ...."

brian-kelley:
@mhoemmen OK, thanks. I just made those changes. Could @trilinos/muelu weigh in? Are the new comments correct, and can we make the restrictions more specific/less conservative?

jhux2:
> I think it's fine to put the requirements in CrsGraph's documentation, but they are not CrsGraph requirements. We could say something like "some downstream packages, including preconditioners in Ifpack2 and MueLu, have the following requirements on column Maps ...."

This could go in the Ifpack2 and MueLu user guides and/or doxygen for the preconditioner adapters.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -1546,11 +1546,13 @@ namespace Tpetra {
     /// does their order change, as a result of calling this method.
     ///
     /// \param newColMap [in] New column Map.  Must be nonnull.
-    ///   The column map's list of global indices on each process must be in a certain order:
+    ///   Within Tpetra, there are no particular restrictions on the column map.
+    ///   However, if this graph will be used in Xpetra, Ifpack2, or MueLu,
+    ///   the column map's list of global indices on each process must be in a certain order:
     ///   locally owned GIDs (sorted), followed by remote GIDs (in order of
```
mhoemmen: Locally owned GIDs don't have to be sorted; they just need to be in the same order as in the domain Map.  The domain Map defines the "locally owned GIDs" for each process.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -1546,11 +1546,13 @@ namespace Tpetra {
...
     ///   owning proc, and sorted within each proc).
     ///
-    ///   It is strongly recommended to use Tpetra::Details::makeColMap()
+    ///   If in doubt, it is strongly recommended to use Tpetra::Details::makeColMap()
```
mhoemmen: I think it's fair to say "It is strongly recommended...."

packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp
```diff
@@ -2389,11 +2389,13 @@ namespace Tpetra {
     /// \brief Replace the matrix's column Map with the given Map.
     ///
     /// \param newColMap [in] New column Map.  Must be nonnull.
-    ///   The column map's list of global indices on each process must be in a certain order:
+    ///   Within Tpetra, there are no particular restrictions on the column map.
+    ///   However, if this matrix will be used in Xpetra, Ifpack2, or MueLu,
+    ///   the column map's list of global indices on each process must be in a certain order:
     ///   locally owned GIDs (sorted), followed by remote GIDs (in order of
```
mhoemmen: See above.

packages/tpetra/core/src/Tpetra_CrsMatrix_decl.hpp
```diff
@@ -2389,11 +2389,13 @@ namespace Tpetra {
...
     ///   owning proc, and sorted within each proc).
     ///
-    ///   It is strongly recommended to use Tpetra::Details::makeColMap()
+    ///   If in doubt, it is strongly recommended to use Tpetra::Details::makeColMap()
```
mhoemmen: See above.

jhux2:
@brian-kelley wrote:
> OK, thanks. I just made those changes. Could @trilinos/muelu weigh in? Are the new comments correct, and can we make the restrictions more specific/less conservative?

These comments looks good to me.

mhoemmen:
Locally owned GIDs don't have to be sorted in the column Map; they just need to be in the same order as in the domain Map. The domain Map defines the "locally owned GIDs" for each process.

jhux2:
> Locally owned GIDs don't have to be sorted in the column Map; they just need to be in the same order as in the domain Map.

@mhoemmen Good catch.

brian-kelley:
I removed "If in doubt", and said that this is called Aztec ordering. I ran Doxygen to make sure everything looks right.

Issue #6404, opened 2019-12-06
Xpetra:  General Summary of the Bug


ndellingwood:
## Bug Report
@trilinos/xpetra @trilinos/tpetra 

### Description
Compilation error when trying to build Trilinos with OpenMP+Serial enabled (no MPI) with gcc/6.5:

```
In file included from /Users/ndellin/Research/trilinos/Trilinos-stable/Build/Build-Pristine/packages/xpetra/src/Xpetra_TpetraVector.hpp:2:0,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/packages/xpetra/src/MultiVector/Xpetra_TpetraMultiVector_def.hpp:1028,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/Build/Build-Pristine/packages/xpetra/src/Xpetra_TpetraMultiVector.hpp:2,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/packages/xpetra/src/Vector/Xpetra_TpetraVector_decl.hpp:53,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/Build/Build-Pristine/packages/xpetra/src/Xpetra_TpetraVector.hpp:1,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/packages/xpetra/src/Vector/Xpetra_VectorFactory_decl.hpp:53,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/Build/Build-Pristine/packages/xpetra/src/Xpetra_VectorFactory.hpp:1,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/packages/xpetra/src/Vector/Xpetra_EpetraVectorFactory.cpp:46:
/Users/ndellin/Research/trilinos/Trilinos-stable/packages/xpetra/src/Vector/Xpetra_TpetraVector_def.hpp:57:106: error: invalid use of incomplete type 'class     Xpetra::TpetraVector<S, LO, GO, N>'
                                 bool                                                              zeroOut)
                                                                                                          ^
In file included from /Users/ndellin/Research/trilinos/Trilinos-stable/Build/Build-Pristine/packages/xpetra/src/Xpetra_TpetraMultiVector.hpp:1:0,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/packages/xpetra/src/Vector/Xpetra_TpetraVector_decl.hpp:53,
[...]

PR #6414, opened 2019-12-09
Xpetra: ETI - Cleaning up some circular references

william76:
@trilinos/xpetra 

Trying to clean up those incomplete type errors that have popped up. Generally, I think the issue that's happening is this:
- The ETI'd classes are in files `filename_decl.hpp` and `filename_def.hpp`
- The ETI system generates a `filename.hpp` that pretty much just does this:
  ```C++
  #include "filename_decl.hpp"
  #include "filename_def.hpp"
  ```
- The issue happens when inside some `filename_def.hpp` file there's something along its include chain that also includes `filename.hpp` file. This goes to the ETI generated header which will include `filename_def.hpp` and then include `filename_decl.hpp`.  The problem is that this load of `filename_def.hpp` is basically a noop since the guard has been seen already and it tries to load in the `filename_decl.hpp` file without ever fully processing the `_def` file.  

The solution seems to be to go through the include chain and do one of these:
1. Move the `#include "Xpetra_something.hpp"` line from the `_def` into the `_decl` file if it's not needed in the def.
2. If (1) can't be done, then change the `#include "Xpetra_something.hpp"` to `#include "Xpetra_something_decl.hpp"`

In this PR I am using (2) on several places.

Affects #6404 and maybe #6314 

FYI @ndellingwood


jhux2:
@william76 I believe #6323 is do to the same problem as here.  I'll try your patch to see if it fixes that issue.

> The solution seems to be to go through the include chain and do one of these:
    1) Move the #include "Xpetra_something.hpp" line from the _def into the _decl file if it's not needed in the def.
    2) If (1) can't be done, then change the #include "Xpetra_something.hpp" to #include "Xpetra_something_decl.hpp"

I don't think we should be doing 1, i.e., you don't want a complete class definition in a decl file unless it's absolutely needed.

ndellingwood:
Thanks @william76 ! Is this PR ready for testing? I can pull it and retest locally.

packages/xpetra/src/MultiVector/Xpetra_TpetraMultiVector_def.hpp
```diff
@@ -1025,7 +1025,7 @@ namespace Xpetra {
 // Following header file inculsion is needed for the dynamic_cast to TpetraVector in elementWiseMultiply (because we cannot dynamic_cast if target is not a complete type)
 // It is included here to avoid circular dependency between Vector and MultiVector
 // TODO: there is certainly a more elegant solution...
-#include "Xpetra_TpetraVector.hpp"
+#include "Xpetra_TpetraVector_decl.hpp"
```
jhux2: The comment on line 1025 indicates a full class definition is necessary.

william76:
@ndellingwood feel free to try it out and see. I was building it earlier and got some MueLu issue but it made it past all the Xpetra stuff.
I've also got a rebuild kicked off on a RHEL7 system and will see what happens there but I'm also interested in seeing what the PR testing turns up.

william76:
> @william76 I believe #6323 is do to the same problem as here. I'll try your patch to see if it fixes that issue.
> 
> > The solution seems to be to go through the include chain and do one of these:
> > 1) Move the #include "Xpetra_something.hpp" line from the _def into the _decl file if it's not needed in the def.
> > 2) If (1) can't be done, then change the #include "Xpetra_something.hpp" to #include "Xpetra_something_decl.hpp"
> 
> I don't think we should be doing 1, i.e., you don't want a complete class definition in a decl file unless it's absolutely needed.

@jhux2 yeah, I think this wasn't a problem early on because there was so little that had been ETI'd. Now that there's alot of stuff in ETI, this issue turns up more.
Assuming this gets fixed, I can try making a pass through all the Xpetra classes we've done ETI to and make them include just the declarations when it's something that's already been through ETI.

jhux2:
>     * The issue happens when inside some `filename_def.hpp` file there's something along its include chain that also includes `filename.hpp` file. This goes to the ETI generated header which will include `filename_def.hpp` and then include `filename_decl.hpp`.  The problem is that this load of `filename_def.hpp` is basically a noop since the guard has been seen already and it tries to load in the `filename_decl.hpp` file without ever fully processing the `_def` file.

@william76 If a macro is already defined, that implies that the file has already been included upstream.  So I'm not sure how this can happen.

william76:
@jhux2 
It happens because the _def file might look something like this, say I'm in `Xpetra_ClassA_decl.hpp`:
```c++
#ifndef __XPETRA_CLASSA_DECL_HPP__
#define __XPETRA_CLASSA_DECL_HPP__
#include "Xpetra_ClassB.hpp"
#include "Xpetra_ClassC.hpp"

// stuff...
#endif
```
Say there's some #include chain that happens from ClassB that somewhere calls `#include "Xpetra_ClassA.hpp"`, the pre-processor will try and load:
```C++
#include "Xpetra_ClassA_decl.hpp"
#include "Xpetra_ClassA_def.hpp"
```
It is in THAT include where the preprocessor loads the decl file and sees that the guard has already been seen so it'll do nothing for the decl file and will immediately then try and load the def file. But since the first ClassA file in the stack hasn't actually been processed, the preprocessor is still sitting on line 5 or so of it the class hasn't actually been declared yet so when it hits the def file we get some flavor of "I don't know what this type is" from the compiler.  In this case yeah, the macro being defined does indicate that the file itself is include upstream but that is not the same as saying the file has been fully processed upstream.

I think the fix is to just make sure inside Xpetra decl files we're not trying to load the actual Class.hpp file but rather the Class_decl.hpp file.

william76:
@ndellingwood @trilinos/muelu 

I re-ran a local build using Nathan's configure script with the ETI fixes I put in.  One thing to note is that Nathan's build doesn't turn on ETI... It gets through the Xpetra builds but then MueLu has a failure:
```
In file included from /ascldap/users/wcmclen/dev/trilinos-dev/build/xpetra-eti-issue-6314/packages/muelu/src/MueLu_RefMaxwell.hpp:1:0,
                 from /ascldap/users/wcmclen/dev/trilinos-dev/source/Trilinos/packages/muelu/adapters/aztecoo/MueLu_AztecEpetraOperator.cpp:8:
/ascldap/users/wcmclen/dev/trilinos-dev/source/Trilinos/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_decl.hpp:50:43: fatal error: MueLu_ExplicitInstantiation.hpp: No such file or directory
 #include "MueLu_ExplicitInstantiation.hpp"
                                           ^
compilation terminated.
make[2]: *** [packages/muelu/adapters/CMakeFiles/muelu-adapters.dir/aztecoo/MueLu_AztecEpetraOperator.cpp.o] Error 1
```

My configuration script, congfig and make logs:
- [do-configure.txt](https://github.com/trilinos/Trilinos/files/3942140/do-configure.txt)
- [__output.configure.log](https://github.com/trilinos/Trilinos/files/3942137/__output.configure.log)
- [__output.make.log](https://github.com/trilinos/Trilinos/files/3942138/__output.make.log)

@ndellingwood does this happen to you too?  I'm trying a build with ETI enabled to see if that error persists. I'm curious too to see what PR testing turns up.

ndellingwood:
@william76 thanks for all the updates. My build has resumed without problems, but very slow going on my laptop. I don't have MueLu enabled but can reconfigure and build to see if I run into the problem you mention.

william76:
@ndellingwood I'm using that same configure as you sent in for #6404 and the configure log shows that MueLu is enabled because Stokhos is enabled... so how are you not enabling MueLu?

packages/xpetra/src/MultiVector/Xpetra_TpetraMultiVector_def.hpp
william76: Done.

william76:
It compiled for me with ETI enabled.  Re-trying without ETI to see if that issue happens with the latest push update.

ndellingwood:
@william76 I didn't explicitly enable MueLu, didn't realize Stokhos automatically enabled it.

ndellingwood:
@william76 I am now seeing a compile error in MueLu, looks like the same one you are seeing:

```
In file included from /Users/ndellin/Research/trilinos/Trilinos-stable/Build/Build-Pristine/packages/muelu/src/MueLu_RefMaxwell.hpp:1:0,
                 from /Users/ndellin/Research/trilinos/Trilinos-stable/packages/muelu/adapters/aztecoo/MueLu_AztecEpetraOperator.cpp:8:
/Users/ndellin/Research/trilinos/Trilinos-stable/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_decl.hpp:50:43: fatal error:                      MueLu_ExplicitInstantiation.hpp: No such file or directory
 #include "MueLu_ExplicitInstantiation.hpp"
                                           ^     
compilation terminated.
make[2]: *** [packages/muelu/adapters/CMakeFiles/muelu-adapters.dir/aztecoo/MueLu_AztecEpetraOperator.cpp.o] Error 1
```

william76:
@ndellingwood  This is the same error I was seeing if you don't turn on ETI when I was doing some testing using your configuration.  I'll let this merge in and will file an issue for the @trilinos/muelu team to look at.

If you want to get past this in the meantime, you should be able to turn on `Trilinos_ENABLE_EXPLICIT_INSTANTIATION:BOOL=ON` and everything should build properly.  At this time, I don't think ETI is on by default but it probably should be. I don't know of anyone downstream that relies on ETI being off anymore.

ndellingwood:
@william76 it's not urgent, thanks for the quick fix!

Issue #6429, opened 2019-12-11
 TeuchosCore_Array_PerformanceTests_basic_MPI_1 rarely randomly failing with unit test ratio timing failures in ATDM CUDA debug builds on 'white' and 'ride' starting 2019-10-12


bartlettroscoe:
CC: @trilinos/teuhcos, @jwillenbring  (Trilinos <product-area-name> Product Lead), @bartlettroscoe, @fryeguy52

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-07-01&end=2019-12-02&filtercount=3&showfilters=1&filtercombine=and&field1=testname&compare1=61&value1=TeuchosCore_Array_PerformanceTests_basic_MPI_1&field2=status&compare2=61&value2=failed&field3=testoutput&compare3=97&value3=Ratio.*FAILED%20%3D%3D) the test:

* `TeuchosCore_Array_PerformanceTests_basic_MPI_1`

in the builds:

* `Trilinos-atdm-white-ride-cuda-10.1-gnu-7.2.0-release-debug` (white)
* `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-rdc-release-debug` (ride)
* `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-rdc-release-debug-pt` (ride)
* `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-release-debug` (ride)

started rarely randomly failing on testing day 2019-10-12 for a total of 16 faluires in taht 4 month period across all of those builds!

[...]

PR #6430, opened 2019-12-11
Teuchos: Change default of maxArrayIterRatio from 100 to 200 (#6429)

bartlettroscoe:
For the CUDA debug builds on 'white' and 'ride', this ratio has gotten as high
as 123 (see #6429).  Not sure what changes on 'white' and 'ride' to trigger these timing
failures, but this particular test is not designed to be performance test per
say in an arbitrary (ebug) build.  It is hard to believe the ratio of
non-optimized Teuchos Array iterator operations is more than 100x slower than
raw pointer operations on this machine but I guess anything is possible.

NOTE: This just changes the default.  It does not change the proper
PERFORMANCE test.

### DIFF
--- a/packages/teuchos/core/test/MemoryManagement/Array_Performance_UnitTests.cpp
+++ b/packages/teuchos/core/test/MemoryManagement/Array_Performance_UnitTests.cpp
@@ -60,7 +60,7 @@ using Teuchos::Ordinal;
 double relCpuSpeed = 1e-2;
 int maxArraySize = 10000;
 double maxArrayBracketRatio =100.0;
-double maxArrayIterRatio = 100.0;
+double maxArrayIterRatio = 200.0;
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
Thanks for fast approval @rppawlo!

mhoemmen:
@bartlettroscoe wrote:
>  It is hard to believe the ratio of
> non-optimized Teuchos Array iterator operations is more than 100x slower than
> raw pointer operations on this machine but I guess anything is possible.

Does "non-optimized" mean "debug checked"?

bartlettroscoe:
> Does "non-optimized" mean "debug checked"?

@mhoemmen, yes.  These 'debug' builds are with `CMAKE_BUILD_TYPE=DEBUG` and `Trilinos_ENABLE_DEBUG=ON`.  I should have mentioned that.

PR #6431, opened 2019-12-11
Expected to fix issue #6423

### LINKED ISSUE #6423: Zoltan2: MultiJagged issue on Spectrum MPI
## Bug Report
@trilinos/zoltan2 

### Description
When running on lassen, MultiJagged is bugging out with the following message:
```
[lassen258:148413] coll:ibm:allreduce: GPU awareness in PAMI requested. It is not safe to defer to another component.
[lassen258:148413] *** An error occurred in MPI_Allreduce
[lassen258:148413] *** reported by process [2154496201,11]
[lassen258:148413] *** on communicator MPI COMMUNICATOR 6 DUP FROM 5
[lassen258:148413] *** MPI_ERR_UNSUPPORTED_OPERATION: operation not supported
[lassen258:148413] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[lassen258:148413] ***    and potentially your MPI job)
```
It seems that there is an issue with certain things not being supported by Spectrum MPI:
https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#open-issues

MicheldeMessieres:
@trilinos/zoltan2

## Motivation
This is expected to fix #6423. We meant to have the reduceAll calls on host and this was still using device buffers.

## Testing
Zoltan2 has 17 tests which run this code.

### DIFF
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -2967,6 +2967,11 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t,
   // max for next concurrentPartCount elements, reduce sum for the last
   // concurrentPartCount elements.
   if(this->comm->getSize()  > 1) {
+    auto host_local_min_max_total =
+      Kokkos::create_mirror_view(local_min_max_total);
+    auto host_global_min_max_total =
+      Kokkos::create_mirror_view(global_min_max_total);
+    Kokkos::deep_copy(host_local_min_max_total, local_min_max_total);
     Teuchos::MultiJaggedCombinedMinMaxTotalReductionOp<int, mj_scalar_t>
       reductionOp(current_concurrent_num_parts,
         current_concurrent_num_parts, current_concurrent_num_parts);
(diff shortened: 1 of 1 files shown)


packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -2967,6 +2967,11 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t,
   // max for next concurrentPartCount elements, reduce sum for the last
   // concurrentPartCount elements.
   if(this->comm->getSize()  > 1) {
+    auto host_local_min_max_total =
```
mhoemmen: Those may be device (CudaUVMSpace) buffers.  Try this:
```c++
auto host_local_min_max_total =
  Kokkos::create_mirror_view(Kokkos::HostSpace(), local_min_max_total);
```

cgcgcg: With this change, i.e. explicit `HostSpace`, it works. There is a second reduction 
https://github.com/trilinos/Trilinos/blob/8e19d1eccea0b01957d562f82a2b3dfb05d5be5d/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp#L3340
that needs the same change. But then things seems to be running again on lassen :-)

MicheldeMessieres:
@mhoemmen Thanks for helping on this. @kddevin Thanks for finding the issue. @cgcgcg  Thanks for trying it. I will update the PR accordingly for the 2 reduceAll calls. I don't understand yet why other MPI calls might not have similar issues on your system but I will do those 2 problem spots just for this PR. Should have this up shortly.

cgcgcg:
@MicheldeMessieres The other spots might not be using a custom reduction operator. Maybe that's the difference?

packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
MicheldeMessieres: Changes implemented.

MicheldeMessieres:
@cgcgcg I updated the PR. Please spot check the changes to make sure I did the same thing you did.

PR #6432, opened 2019-12-11
MueLu: Missing test output

cgcgcg:
@trilinos/muelu

## Motivation
The parameter list interpreter test was missing output since only rank 0 was redirected to file.


cgcgcg:
@jhux2 No, there isn't. But a lot of MueLu builds are failing in this test.

Issue #6392, opened 2019-12-04
MueLu: Hanging in empire tests.


rppawlo:
## Bug Report
@trilinos/muelu @cgcgcg @csiefer2 @jmgate @bathmatt 

### Description
The update of EMPIRE's fork of Trilinos has been broken for some time. Using today's trilinos development branch, we are seeing a hang in muelu in some empire tests. Stack trace is below. On a 4 process problem, mpi processes 0, 2, and 3 hang here:
```
(gdb) bt
#0  0x00007f2a1f9685de in mca_btl_vader_component_progress () from /projects/sems/install/rhel7-x86_64/sems/compiler/clang/3.9.0/openmpi/1.10.1/lib/openmpi/mca_btl_vader.so
#1  0x00007f2a29a49dba in opal_progress () from /projects/sems/install/rhel7-x86_64/sems/compiler/clang/3.9.0/openmpi/1.10.1/lib/libopen-pal.so.13
#2  0x00007f2a2d1ef8d5 in ompi_request_default_wait_all () from /projects/sems/install/rhel7-x86_64/sems/compiler/clang/3.9.0/openmpi/1.10.1/lib/libmpi.so.12
#3  0x00007f2a2d21d1a1 in PMPI_Waitall () from /projects/sems/install/rhel7-x86_64/sems/compiler/clang/3.9.0/openmpi/1.10.1/lib/libmpi.so.12
#4  0x0000000001cdd117 in Teuchos::MpiComm<int>::waitAll(Teuchos::ArrayView<Teuchos::RCP<Teuchos::CommRequest<int> > > const&) const ()
#5  0x000000000561f93c in Tpetra::Distributor::doWaits() ()
[...]

PR #6433, opened 2019-12-11
Tpetra: setAllIndices verifies sorting in debug

brian-kelley:
Also improve test coverage for MatrixMatrix::add().
Both the HAVE_TPETRA_DEBUG check in CrsGraph::setAllIndices()
and the new check in MatrixMatrix unit tests would have caught #6392.

Deleted AddProfilingTest - it was really just a couple of tests
for MatrixMatrix::add, not a performance test. The tests have
been moved into MatrixMatrix_UnitTests.cpp. I can always create
a performance test for add() if needed.

@trilinos/tpetra 

## Motivation

- Clearly needed better testing for add() after #6392 
- Enforcing the CRS sorting precondition in CrsGraph::setAllIndices() may catch bugs in the future. Only done in debug mode.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #6392 #6405 
* Part of 
* Composed of 

## Stakeholder Feedback

@rppawlo suggested adding this check
## Testing

The changes are just to tests.


packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -3471,12 +3471,49 @@ namespace Tpetra {
     TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(
       ! hasColMap () || getColMap ().is_null (), std::runtime_error,
       "The graph must have a column Map before you may call this method.");
+    LocalOrdinal numLocalRows = this->getNodeNumRows ();
     TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(
-      static_cast<size_t> (rowPointers.size ()) != this->getNodeNumRows () + 1,
+      static_cast<LocalOrdinal> (rowPointers.size ()) != numLocalRows + 1,
```
mhoemmen: If the number of local rows is zero, I think Tpetra permits `rowPointers` to have size zero.  This code doesn't change the current requirements, though.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -3471,12 +3471,49 @@ namespace Tpetra {
...
       std::runtime_error, "rowPointers.size() = " << rowPointers.size () <<
-      " != this->getNodeNumRows()+1 = " << (this->getNodeNumRows () + 1) <<
+      " != this->getNodeNumRows()+1 = " << (numLocalRows + 1) <<
       ".");
 
+#ifdef HAVE_TPETRA_DEBUG
```
mhoemmen: Please use `Tpetra::Details::Behavior::debug()` for a run-time test.  This will let users decide at run time if they want to do debug checks, even in a release build.  On some platforms, it's impossible or infeasible to rebuild or to finish a debug build.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -1588,7 +1593,14 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(Tpetra_MatMat, threaded_add_sorted, SC, LO, GO
         std::string("Threaded addition of sorted Kokkos::CrsMatrix produced row ") + std::to_string(i) + " with the wrong number of entries.");
     for(size_t j = 0; j < Crowlen; j++)
     {
-      TEUCHOS_TEST_FOR_EXCEPTION(valsCRS[2](Crowstart + j) != correctVals[colindsCRS[2](Crowstart + j)], std::logic_error,
+      size_t Coffset = Crowstart + j;
+      if(j > 0)
+      {
+        //Check entries are sorted
+        TEUCHOS_TEST_FOR_EXCEPTION(colindsCRS[2](Coffset - 1) > colindsCRS[2](Coffset), std::logic_error,
+            std::string("Threaded addition of unsorted Kokkos::CrsMatrix: row ") + std::to_string(i) + " is not sorted");
```
mhoemmen: You don't need to cast to string here, and you can use `<<` instead of `+`.  The third macro argument gets pasted in like this:
```c++
std::ostringstream os;
os << THIRD_ARG;
```

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
mhoemmen:
```c++
#include "Tpetra_Details_Behavior.hpp"
// ...
const bool debug = Tpetra::Details::Behavior::debug();
if (debug) {
  // ... do expensive debug checks ...
}
```

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -1782,14 +1728,262 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(Tpetra_MatMat, threaded_add_unsorted, SC, LO,
...
+      GO col = Cinds[Cit];
+      SC val = Cvals[Cit];
+      SC goldVal = 0;
+      if(Ait < Aentries && Ainds[Ait] == col)
+        goldVal += Avals[Ait++];
+      if(Bit < Bentries && Binds[Bit] == col)
+        goldVal += Bvals[Bit++];
+      //Any scalar magnitude should implicitly convert to double
+      double err = KAT::abs(val - goldVal);
+      if(err > 1e-13)
```
mhoemmen: At some point, we should put a Scalar-specific, matrix-specific tolerance here.

brian-kelley:
Looks like the new debug check caught two places that don't sort the indices, so that's good.

PR #6439, opened 2019-12-12
Ifpack2: Relaxation MTGS/MTSGS enhancements

brian-kelley:
Add support for new features in KokkosKernels Gauss-Seidel:
multivector apply, and cluster-coloring GS (aka multicolor block GS).

Test the new features in Relaxation unit tests, and document the new
"relaxation: mtgs cluster size" parameter in user guide.

@trilinos/ifpack2 

## Motivation

The multivector support was specifically requested by @iyamazaki - it significantly speeds up the smoother apply() and reduces the number of kernel launches. The cluster-coloring Gauss-Seidel is a new feature in KokkosKernels that provides better convergence than point-coloring. There is still some work needed to make apply() performance match point coloring, but it does provide the numerical improvements.

## Stakeholder Feedback

@iyamazaki requested multivector support, and Exawind was interested in using the cluster GS.
## Testing

The new features are now tested in Ifpack2 unit tests: Ifpack2Relaxation_MTSGS, and Ifpack2Relaxation_ClusterMTSGS.

@ndellingwood, since I ripped out some but not all of the KokkosKernels changes that will appear in the 3.0 snapshot, this might create merge conflicts in that branch but I can help fix them.


packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
```diff
@@ -393,7 +392,7 @@ struct FillSymmetricLowerEdgesHashMap{
     in_lno_nnz_view_t adj_,
     hashmap_t hashmap_,
     out_lno_row_view_t pre_pps_,
-    bool lower_only_ = false
+    bool /* lower_only_ */ = false
```
srajama1: hmm .. what is this doing ?

ndellingwood:
> @ndellingwood, since I ripped out some but not all of the KokkosKernels changes that will appear in the 3.0 snapshot, this might create merge conflicts in that branch but I can help fix them.

@brian-kelley to make sure I understand, the concern is because in this PR you've re-introduced some of the updates from kokkos-kernels develop branch (cluster GS related), but they may not match exactly with what is in kokkos-kernels develop, correct? We shouldn't have to worry about a merge conflict because the snapshot will clobber the Trilinos copy with an updated master.

packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
brian-kelley: @srajama1 I just copied in the whole Utils.hpp since there were a couple of things the MTSGS needed from there (an array sum-reducer, and the updated team/vector size heuristics). This change removing an unused parameter warning:
```
commit d2032fdf9e1893a469285a184a52e018920e9ddc
Author: Jason M. Gates <jmgate@sandia.gov>
Date:   Tue Apr 23 16:54:50 2019 -0600

    Remove Unused Parameter Warnings
    
    `-Wunused-parameter` warnings.  If their use is hidden behind an #ifdef,
    use an #else block to avoid the warning appropriately.
```

brian-kelley: Based on the name of the functor, I guess "lower only" is redundant since that's always what it does.

packages/ifpack2/src/Ifpack2_Parameters.cpp
```diff
@@ -129,6 +129,7 @@ void getValidParameters(Teuchos::ParameterList& params)
   params.set("relaxation: l1 eta",(double)1.5);
   params.set("relaxation: banded container superdiagonals", -1);
   params.set("relaxation: banded container subdiagonals", -1);
+  params.set("relaxation: mtgs cluster size", (int)1);
```
mhoemmen: `1` has type `int` already, but the extra cast is harmless.

brian-kelley:
@ndellingwood Yes exactly, if it just clobbers packages/kokkos-kernels then everything will be fine.

packages/ifpack2/src/Ifpack2_Parameters.cpp
```diff
@@ -129,6 +129,7 @@ void getValidParameters(Teuchos::ParameterList& params)
   params.set("relaxation: l1 eta",(double)1.5);
```
mhoemmen: `1.5` has type `double` already, but the extra cast is harmless.

packages/ifpack2/src/Ifpack2_Relaxation_decl.hpp
```diff
@@ -782,6 +782,8 @@ class Relaxation :
   bool fixTinyDiagEntries_;
   //! Whether to spend extra effort and all-reduces checking diagonal entries.
   bool checkDiagEntries_;
+  //! For MTSGS, the cluster size (use point coloring if equal to 1)
+  int clusterSize_;
```
mhoemmen: At some point (not now), we should start using member initializers to set defaults.

packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
```diff
@@ -2165,45 +2175,37 @@ MTGaussSeidel (const Tpetra::MultiVector<scalar_type,local_ordinal_type,global_o
...
-        TEUCHOS_TEST_FOR_EXCEPTION(
-            true, std::invalid_argument,
-            prefix << "The 'direction' enum does not have any of its valid "
-            "values: Forward, Backward, or Symmetric.");
-      }
+    if (direction == Tpetra::Symmetric) {
+      KokkosSparse::Experimental::symmetric_gauss_seidel_apply
+      (mtKernelHandle_.getRawPtr(), A_->getNodeNumRows(), A_->getNodeNumCols(),
+          kcsr.graph.row_map, kcsr.graph.entries, kcsr.values,
+          X_colMap->template getLocalView<MyExecSpace> (),
```
mhoemmen: This should always be device, so you could just call the nontemplated method `getLocalViewDevice`.

packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
```diff
@@ -1916,8 +1900,56 @@ void init_view_withscalar(typename in_row_view_t::size_type num_elements, in_row
   MyExecSpace().fence();
 }
 
+//A sum-reduction scalar representing a fixed-size array.
```
mhoemmen: This feels like a thing that belongs in Kokkos.  It looks related to `Kokkos::array`.

brian-kelley: @mhoemmen It's similar in that it's a fixed-size array, but Kokkos::Array doesn't have operator+ built in which I need to define the reduction. I could replace the ``data`` member with a Kokkos::Array but I would still have to keep ``array_sum_reduce``.

brian-kelley: There is an opportunity to remove duplicate code though, since @lucbv made exactly the same thing for SPMV.

packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
srajama1: I would say delete the line :)

packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
srajama1: We should consider it at least as part of Kokkos Kernels Utils, if it doesn't belong in Kokkos. Could be a different commit though.

mhoemmen: > It's similar in that it's a fixed-size array, but Kokkos::Array doesn't have operator+ built in which I need to define the reduction.

I was just thinking we could use Kokkos::Array if it exists, but no worries :)

packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
srajama1: BTW, doesn't have to block this commit. It is just silly.

PR #6452, opened 2019-12-13
Teuchos: SyncTimeMonitor

cgcgcg:
@trilinos/teuchos @trilinos/muelu 

## Motivation
- Add a special TimeMonitor to Teuchos that calls MPI_Barrier before stopping,
- Use it in MueLu RefMaxwell.


cgcgcg:
@bartlettroscoe Sounds good! Will do next year.

mhoemmen:
Are you really sure you want to have a collective in a destructor?  That would make the destructor blocking.

cgcgcg:
@mhoemmen The destructor of the parent class stops the timer. Since we want to have the barrier before the timer stops, this seems like the way to go. Is there any issue with having a blocking destructor?

rppawlo:
> Is there any issue with having a blocking destructor?

If an exception is thrown on one process and the stack unwound, the collective in the dtor will cause the code to hang.

bartlettroscoe:
Don't know if a possible hang in a destructor is very different than regular C++ code.  For example, consider the function:

```
void outerFunc()
{
  innerFunc();
  comm_->barrier()  
}
```

If the function `innerFunc()` throws an exception on one rank but the other ranks don't and call `comm_->barrier()`, then the program will also hang.

Do we really have rational approach for addressing hangs in C++ programs where only a subset of processes throw exceptions?

mhoemmen:
@bartlettroscoe wrote:
> Do we really have rational approach for addressing hangs in C++ programs where only a subset of processes throw exceptions?

No ;-P  (Seriously, the only way to do that is to maintain separate local and global error states, and "go through the motions" locally until you get to a collective that permits exchange of error state.  It's hard to write and hard to test.)

mhoemmen:
Actually wait -- MueLu could piggyback error handling on top of this, no?  Just maintain the local error state until the timing region closes, then all-reduce instead of doing a barrier.  If you're not doing a barrier, you could just call `MPI_Abort` if there's a local error.

rppawlo:
If an exception can unroll all the way up it can at least terminate the other mpi processes as it exits. We just hit this bug last week with hdf5 calling a collective in the dtor that hung large runs in empire.

cgcgcg:
@mhoemmen @rppawlo Could you have a look if the added unit test covers the scenario that you were referring to?

packages/teuchos/comm/src/Teuchos_TimeMonitor.hpp
```diff
@@ -757,12 +757,45 @@ class TEUCHOSCOMM_LIB_DLL_EXPORT TimeMonitor :
...
+  //! Default constructor is deleted, since it would be unsafe.
+  SyncTimeMonitor () = delete;
+
+  //! Destructor: stops the timer.
+  ~SyncTimeMonitor() override;
+  //@}
+
+private:
+  // \brief Communicator on which barrier will be called.
+  Ptr<const Comm<int> > comm_;
```
mhoemmen: If you use a `Ptr` and not `const Comm<int>&`, then it should be possible for `comm_` to be null.  However, it looks like you always assume that `comm_` is nonnull.

Issue #6355, opened 2019-11-26
Teko: configuration error with deprecated code off and -D Tpetra_INST_INT_UNSIGNED_LONG:BOOL=ON 


ikalash:
I attempted to switch Albany to turn off Tpetra and Xpetra deprecated code in the Albany nightlies (with @mperego).  In some of the builds, I need to have the flag: 

```
-D Tpetra_INST_INT_UNSIGNED_LONG:BOOL=ON \
```

in order to avoid DTK-related compilation errors.  I also have Teko on, as it is used in some of the tests that use DTK.  Doing so produces the following error:

```
Processing enabled package: Teko (Libs)
CMake Error at packages/teko/CMakeLists.txt:25 (MESSAGE):
  Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long
  long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO =
  long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like
  to use other GO types, please contact the Teko developers.
```
Is it possible to have a build with the following 4 options simultaneously?

```
 -D Tpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
 -D Xpetra_ENABLE_DEPRECATED_CODE:BOOL=OFF \
 -D Trilinos_ENABLE_Teko:BOOL=ON \
 -D Tpetra_INST_INT_LONG_LONG=ON \
```
@trilinos/teko

eric-c-cyr:
Teko is probably hard coded to use one global ordinal type. I’m not really sure about changes propagated as part of the DEPRECATED refactor. I will do a quick check to see if there is something easy. Otherwise, I probably have no bandwidth for this until late January at the earliest.

Sent from my iPad

On Nov 25, 2019, at 11:21 PM, Irina K. Tezaur <notifications@github.com> wrote:

﻿

I attempted to switch Albany to turn off Tpetra and Xpetra deprecated code in the Albany nightlies (with @mperego<https://github.com/mperego>). In some of the builds, I need to have the flag:

-D Tpetra_INST_INT_UNSIGNED_LONG:BOOL=ON \


in order to avoid DTK-related compilation errors. I also have Teko on, as it is used in some of the tests that use DTK. Doing so produces the following error:

Processing enabled package: Teko (Libs)
CMake Error at packages/teko/CMakeLists.txt:25 (MESSAGE):
[...]

PR #6454, opened 2019-12-16
Teko, Tpetra: implementing proposed fixes in issue #6355 to make it p…

ikalash:
This PR makes it possible to compile Trilinos with Tpetra_INST_INT_UNSIGNED_LONG:BOOL=ON with the Tpetra/Tpetra deprecated code turned OFF with DataTransferKit (DTK) v2.  We need such a configuration for Albany.

@trilinos/teko 
@trilinos/tpetra

### DIFF
--- a/packages/teko/CMakeLists.txt
+++ b/packages/teko/CMakeLists.txt
@@ -10,53 +10,6 @@ TRIBITS_PACKAGE(Teko)
 # B) Set up package-specific options
 #
 
-# FIXME (mfh 02 Nov 2015, 04 Nov 2015) The right thing for Teko to do
-# would be to have logic to use any enabled Tpetra GlobalOrdinal type.
-# For now, we just error out if neither long long nor int are enabled.
-
-#IF (DEFINED Tpetra_INST_INT_LONG_LONG)
-#  IF (Tpetra_INST_INT_LONG_LONG)
-#    SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT ON)
-#  ELSE ()
-#    IF (DEFINED Tpetra_INST_INT_INT)
-#      IF (Tpetra_INST_INT_INT)
-#	SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT OFF)
-#      ELSE ()
-#	MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
-#      ENDIF () # Tpetra_INST_INT_INT
-#    ELSE () # NOT DEFINED Tpetra_INST_INT_INT
-#      MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
-#    ENDIF ()
-#  ENDIF () # Tpetra_INST_INT_LONG_LONG
-#ELSE () # NOT DEFINED Tpetra_INST_INT_LONG_LONG
-#  IF (NOT DEFINED Tpetra_INST_INT_INT)
-#    MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
-#  ELSEIF (NOT Tpetra_INST_INT_INT)
-#    MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
-#  ENDIF ()
-#  SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT OFF)
-#ENDIF ()
-
-#ASSERT_DEFINED(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT)
-#TRIBITS_ADD_OPTION_AND_DEFINE(
-#  ${PACKAGE_NAME}_USE_LONGLONG_GO
-#  HAVE_${PACKAGE_NAME}_USE_LONGLONG_GO
-#  "Use long long for global ordinal"
-#  ${${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT}
-#  )
-
-#ASSERT_DEFINED(${PACKAGE_NAME}_USE_LONGLONG_GO)
-#IF (${PACKAGE_NAME}_USE_LONGLONG_GO)
-#  IF (DEFINED Tpetra_INST_INT_LONG_LONG)
-#    IF (NOT Tpetra_INST_INT_LONG_LONG)
-#      MESSAGE(FATAL_ERROR "Teko: If GO=long long is disabled in Tpetra, you may not set ${PACKAGE_NAME}_USE_LONGLONG_GO = ON.  First, set Tpetra_INST_INT_LONG_LONG=ON to enable GO=long long in Tpetra.")
-#    ENDIF ()
-#  ELSE () # NOT DEFINED Tpetra_INST_INT_LONG_LONG
-#    MESSAGE(FATAL_ERROR "Teko: If GO=long long is disabled in Tpetra, you may not set ${PACKAGE_NAME}_USE_LONGLONG_GO = ON.  First, set Tpetra_INST_INT_LONG_LONG=ON to enable GO=long long in Tpetra.")      
-#  ENDIF ()
-#ENDIF ()
-
-
 TRIBITS_ADD_OPTION_AND_DEFINE(
   Ifpack2_Trilinos
   HAVE_${PACKAGE_NAME}_ENABLE_Ifpack2
--- a/packages/teko/CMakeLists.txt
+++ b/packages/teko/CMakeLists.txt
@@ -14,47 +14,47 @@ TRIBITS_PACKAGE(Teko)
 # would be to have logic to use any enabled Tpetra GlobalOrdinal type.
 # For now, we just error out if neither long long nor int are enabled.
 
-IF (DEFINED Tpetra_INST_INT_LONG_LONG)
-  IF (Tpetra_INST_INT_LONG_LONG)
-    SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT ON)
-  ELSE ()
-    IF (DEFINED Tpetra_INST_INT_INT)
-      IF (Tpetra_INST_INT_INT)
-	SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT OFF)
-      ELSE ()
-	MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
-      ENDIF () # Tpetra_INST_INT_INT
-    ELSE () # NOT DEFINED Tpetra_INST_INT_INT
-      MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
-    ENDIF ()
-  ENDIF () # Tpetra_INST_INT_LONG_LONG
-ELSE () # NOT DEFINED Tpetra_INST_INT_LONG_LONG
-  IF (NOT DEFINED Tpetra_INST_INT_INT)
-    MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
-  ELSEIF (NOT Tpetra_INST_INT_INT)
-    MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
-  ENDIF ()
-  SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT OFF)
-ENDIF ()
+#IF (DEFINED Tpetra_INST_INT_LONG_LONG)
+#  IF (Tpetra_INST_INT_LONG_LONG)
+#    SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT ON)
+#  ELSE ()
+#    IF (DEFINED Tpetra_INST_INT_INT)
+#      IF (Tpetra_INST_INT_INT)
+#	SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT OFF)
+#      ELSE ()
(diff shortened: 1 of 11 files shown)


packages/teko/CMakeLists.txt
```diff
@@ -14,47 +14,47 @@ TRIBITS_PACKAGE(Teko)
...
-ENDIF ()
+#IF (DEFINED Tpetra_INST_INT_LONG_LONG)
+#  IF (Tpetra_INST_INT_LONG_LONG)
+#    SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT ON)
+#  ELSE ()
+#    IF (DEFINED Tpetra_INST_INT_INT)
+#      IF (Tpetra_INST_INT_INT)
+#	SET(${PACKAGE_NAME}_USE_LONGLONG_GO_DEFAULT OFF)
+#      ELSE ()
+#	MESSAGE (FATAL_ERROR "Teko currently requires that Tpetra enable either GlobalOrdinal (GO) = long long or GO = int.  Please set Tpetra_INST_INT_LONG_LONG=ON to enable GO = long long, or Tpetra_INST_INT_INT=ON to enable GO = int.  If you would like to use other GO types, please contact the Teko developers.")
```
mhoemmen: We'll have to ask a @trilinos/teko developer whether Teko needs those defines, in case `Tpetra_INST_INT_LONG_LONG` or `Tpetra_INST_INT_INT` are set.  On the other hand, we plan to land the big "remove deprecated code" PR #6388 tomorrow.  That PR would make it so that Tpetra only ever has one `GO` type enabled.  Thus, it should be correct for Teko not to need these CMake options any more; in fact, there would be no way to use them once deprecated code is removed.

packages/teko/src/Teko_ConfigDefs.hpp
```diff
@@ -54,7 +54,13 @@
 namespace Teko {
 
 typedef double ST;
-typedef int LO;
+using LO = Tpetra::Map<>::local_ordinal_type;
```
mhoemmen: I think this change is fine, especially since we plan to land the big "remove deprecated code" PR #6388 tomorrow.  That PR would make it so that Tpetra only ever has one `GO` type enabled.  Thus, it would be correct for Teko to use Tpetra's default types.

packages/teko/src/Teko_Utilities.cpp
```diff
@@ -234,21 +234,21 @@ RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > buildGraphLaplacian(int dim,ST * coords,con
...
+   auto sz = stencil.getGlobalMaxNumRowEntries()+1; 
+   Teuchos::Array<ST> rowData(sz);
+   Teuchos::Array<GO> rowInd(sz);
 
    // loop over all the rows
    for(LO j=0;j< (LO) gl->getNodeNumRows();j++) {
       GO row = gl->getRowMap()->getGlobalElement(j);
       ST diagValue = 0.0;
-      GO diagInd = -1;
+      int diagInd = -1;
```
mhoemmen: This could cause 64-bit - to - 32-bit assignment issues.  It would be better to do this:
```c++
GO diagInd = Teuchos::OrdinalTraits<GO>::invalid();
```

packages/teko/tests/src/Tpetra/tTpetraThyraConverter.cpp
```diff
@@ -87,11 +87,11 @@ double compareTpetraMVToThyra(const Tpetra_MultiVector & eX,
                             int indexStart=-1,int indexEnd=-1); */
 double compareTpetraMVToThyra(const Tpetra::MultiVector<ST,LO,GO,NT> & eX,
                             const Teuchos::RCP<const Thyra::MultiVectorBase<ST> > & tX,
-                            int verbosity,std::ostream & os,GO indexStart=-1,GO indexEnd=-1);
+                            int verbosity,std::ostream & os,int indexStart=-1,GO indexEnd=-1);
```
mhoemmen: See above: use `Teuchos::OrdinalTraits<GO>::invalid()` instead of `-1`, and preserve `GO` here.

packages/teko/tests/unit_tests/tDiagonallyScaledPreconditioner.cpp
```diff
@@ -132,7 +132,8 @@ const RCP<Thyra::LinearOpBase<ST> > buildSystem(const Teuchos::RCP<const Teuchos
    RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > mat = Tpetra::createCrsMatrix<ST,LO,GO,NT>(map,3);
 
    ST values[] = { -1.0, 2.0, -1.0};
-   GO iTemp[] = {-1,0,1}, indices[3];
+   int iTemp[] = {-1,0,1};
```
mhoemmen: I would need a Teko developer to tell me whether this test would make sense with `GO=unsigned long`.

packages/teko/tests/unit_tests/tStratimikosFactory.cpp
```diff
@@ -256,7 +258,7 @@ const RCP<Tpetra::Operator<ST,LO,GO,NT> > buildStridedSystem(const Teuchos::RCP<
         vPtr = (n==0) ? valuesA : valuesB;
         iPtr = indices;
      
-        if(gid<numUnks) {
+        if(gid<(GO)numUnks) {
```
mhoemmen: Prefer C++ - style casts to this C-style cast.

ikalash:
Thanks for the review @mhoemmen .  I will implement the changes you suggest.  It seems that the auto-checking of my PR failed last night and I'm trying to see what went wrong.  It seems my PR is not there on the CDash site, however: https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-6454-test&field2=buildstarttime&compare2=84&value2=NOW .  Am I missing something?

packages/teko/CMakeLists.txt
ikalash: I agree - I had the same concern.  @eric-c-cyr , can you please have a look and comment?

packages/teko/src/Teko_ConfigDefs.hpp
ikalash: Great.  The removal of the deprecated StaticGraph code PR #6388 today is why I was hoping to get this PR in sooner rather than later....

mhoemmen:
@ikalash I didn't see the PR build there this morning -- weird.  @trilinos/framework might have more to say, but I think there have been some file system issues recently relating to the Dashboard.

packages/teko/src/Teko_Utilities.cpp
ikalash: I had that originally.  The problem is with the change, some of the Teko tests fail if I have ```Tpetra_INST_INT_UNSIGNED_LONG:BOOL=ON``` b/c there is logic assuming diagInd = -1, which it cannot be if the type of it is unsigned long.  I agree with you that with my change, there's a concern about a 64-bit -to - 32-bit assignment.  I think the question is, do we care if some Teko tests fail if ```Tpetra_INST_INT_UNSIGNED_LONG:BOOL=ON```  (as some Tpetra tests do).  If we are OK with them failing, we can keep the logic as it was originally.

packages/teko/tests/src/Tpetra/tTpetraThyraConverter.cpp
ikalash: Same comment as above.

packages/teko/tests/unit_tests/tStratimikosFactory.cpp
ikalash: Do you mean: ```reinterpret_cast<GO>(numUnks)``` instead of ```(GO)numUnks```?  I get a compilation error when I try this, hmmm...

packages/teko/tests/unit_tests/tDiagonallyScaledPreconditioner.cpp
ikalash: I agree - had the same concern.  If we're OK with some Teko tests failing if ```Tpetra_INST_INT_UNSIGNED_LONG:BOOL=ON```, I'd probably vote for reverting the logic to how it was before.

ikalash:
Hmmm, weird... I wonder if the auto-tester is misbehaving again...  Is there a way I can try the auto-test again?  I'd prefer to see on the CDash what failed prior to trying to reproduce the failures (esp. since the autotester is flaky, and sometimes things show up as "failed" even when the build is fine...).

packages/teko/src/Teko_Utilities.cpp
mhoemmen: @ikalash I don't think Albany uses Teko, does it?  It might be enough just to fix Tpetra.

packages/teko/tests/unit_tests/tStratimikosFactory.cpp
mhoemmen: `static_cast<GO>(numUnks)` or just `GO(numUnks)`.

packages/teko/src/Teko_Utilities.cpp
ikalash: @mhoemmen : it does use Teko actually, in the LCM package.  This usage will be expanded in the near future, as we switch the entire code to block data structures.  @eric-c-cyr seems to imply in his email that the changes should be OK, but I'd like him to confirm.  Also it appears auto-tester got relaunched, and I'm also doing a build now to reproduce one of the auto-tester builds - this might shed some light into whether or not there are potential issues.  I think the auto-tester build failed earlier b/c I accidentally indicated that I want to merge the PR into master, not develop.

packages/teko/tests/unit_tests/tStratimikosFactory.cpp
ikalash: I prefer the C-style syntax, I guess... But I'm ok with the change - just made it.

packages/teko/tests/unit_tests/tDiagonallyScaledPreconditioner.cpp
ikalash: @eric-c-cyr is OK with this change - in the unit tests, iTemp will never be so large that ``long`` will be needed.  Therefore keeping change.

packages/teko/tests/src/Tpetra/tTpetraThyraConverter.cpp
ikalash: @eric-c-cyr is OK with this change - in the unit tests, indexStart will never be so large that long will be needed. Therefore keeping change.

packages/teko/src/Teko_Utilities.cpp
ikalash: @mhoemmen : I just pushed a fix that should avoid Teko test failures and potential issues with 64-bit - to -32-bit conversions in the Laplacian graph routine (which @eric-c-cyr was concerned about) using ifdefs.  Let me know if you don't like the fix.

packages/teko/CMakeLists.txt
ikalash: I think we resolved this in our email chain with @eric-c-cyr ...

packages/teko/src/Teko_Utilities.cpp
ikalash: I just pushed a better fix involving invalid() and no ifdefs.  I think we can say the issue has been resolved.  There are no test failures with GO=unsigned long.

packages/teko/src/Teko_Utilities.cpp
```diff
@@ -238,11 +238,17 @@ RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > buildGraphLaplacian(int dim,ST * coords,con
    Teuchos::Array<ST> rowData(sz);
    Teuchos::Array<GO> rowInd(sz);
 
+#ifdef HAVE_TPETRA_INT_UNSIGNED_LONG
+   long diagInd;
```
mhoemmen: @ikalash You can do this without macros with `std::make_signed`.

eric-c-cyr: Thanks @mhoemmen I was going to suggest writing this, its nice to know the standard is somewhat sane.

ikalash: I re-did this w/o ifdefs - sorry about the confusion!  Here is the current commit showing the change: https://github.com/trilinos/Trilinos/pull/6454/commits/b8456c949203728608559390b59bebdca91e116d .

mhoemmen: @eric-c-cyr if it's not sane, please feel encouraged to write a paper and one of our C++ Standard Committee representatives will take it up at the next meeting :-)

packages/teko/src/Teko_Utilities.cpp
```diff
@@ -234,21 +234,21 @@ RCP<Tpetra::CrsMatrix<ST,LO,GO,NT> > buildGraphLaplacian(int dim,ST * coords,con
...
-      GO diagInd = -1;
+      GO diagInd = Teuchos::OrdinalTraits<GO>::invalid();
       size_t rowSz = 0;
 
       // extract a copy of this row...put it in rowData, rowIndicies
       stencil.getGlobalRowCopy(row,Teuchos::ArrayView<GO>(rowInd),Teuchos::ArrayView<ST>(rowData),rowSz);
- 
       // loop over elements of row
-      for(size_t i=0;i<rowSz;i++) {
+      for(GO i=0;i<rowSz;i++) {
```
mhoemmen: I forgot to mention that `rowSz` is a `size_t` here, so this may cause signed/unsigned comparison warnings.

ikalash: Good call... I don't know why I made this change now that I look at it.  I just changed it back.

ikalash:
@mhoemmen , @eric-c-cyr : ok, I'm done with all the changes you requested, I think.  Let me know if there's something else you'd like me to change/look at.  Otherwise I'll merge once the auto-tester completes.

ikalash:
So, autotesting just completed, and there are some test failures in Panzer on certain machines: https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-6454-test&field2=buildstarttime&compare2=84&value2=NOW .  I looked at the output and I don't see how my changes could have caused such issues.  In particular, all the errors seem to have to do with a bad cast from Tpetra to Epetra: 
```

 Entering Thyra::BelosLinearOpWithSolveFactory<double>::initializeOpImpl(...) ...
  Teko: "buildInverse" could not construct the inverse operator using "Teko::AutoClone<mini_em::FullMaxwellPreconditionerFactory_Augmentation, mini_em::FullMaxwellPreconditionerFactory_Augmentation>"
  
  *** THROWN EXCEPTION ***
  /scratch/trilinos/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/Trilinos/packages/teuchos/core/src/Teuchos_dyn_cast.cpp:61:
  
  Throw number = 1
  
  Throw test that evaluated to true: true
  
  dyn_cast<Thyra::EpetraLinearOp>(Thyra::LinearOpBase<double>) : Error, the object with the concrete type 'Thyra::TpetraLinearOp<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >' (passed in through the interface type 'Thyra::LinearOpBase<double>')  does not support the interface 'Thyra::EpetraLinearOp' and the dynamic cast failed!
  ************************
terminate called after throwing an instance of 'std::exception'
  what():  std::exception
```
@eric-c-cyr , @mhoemmen : thoughts on what could be going on?  I can start digging in, but I'm surprised my changes could lead to this kind of error so some leads would be helpful.   I would think the only possible culprit could be my changes to packages/teko/CMakeLists.txt .

mhoemmen:
@ikalash wrote:
> thoughts on what could be going on? I can start digging in, but I'm surprised my changes could lead to this kind of error so some leads would be helpful. I would think the only possible culprit could be my changes to packages/teko/CMakeLists.txt .

My guess is that taking out that CMake logic from Teko enabled Teko's Epetra support, or that turning off `GO=int` disabled some Epetra factory registration.  If you have MueLu enabled in your build and `GO=int` is disabled, then you'll want to set `Xpetra_ENABLE_Epetra=OFF` and `Amesos2_ENABLE_Epetra=OFF`; that might help.

ikalash:
So I think I am going to abandon this PR - I've been unsuccessful in reproducing the Panzer failures (had a build going on my RHEL7 machine to try to reproduce the nightlies, which died after almost a day...), and I actually have a better fix now to my problem.  Basically, I figured out how to modify DTK to work with a signed GO, which is in a fork of DTK I created.  I will just switch our DTK to use my fork of it with the necessary changes for now, I think.  @eric-c-cyr , @mhoemmen : if you're interested in this PR, I'm happy to reopen it, since all the code is there; but I'd need some help debugging the failing Teko tests, I think; or at least reproducing the problem.

mhoemmen:
@ikalash Thanks for working on this!  I think your new approach is a better idea.  I would worry about assuming that `GO=unsigned long` still works, without some more extensive testing.

ikalash:
@mhoemmen : agreed.  BTW, in working on this, I discovered that a bunch of things in Trilinos (tests at least for various packages, e.g. Tpetra) fail with ```GO=unsigned long```.  Maybe if that's the case more packages than just Teko should throw an error when trying to build with this GO type?  Curiously, it could be that just tests are broken, not core capabilities, as I had several Albany builds with ```GO=unsigned long``` until now, which ran just fine.  Albany uses a lot of Trilinos packages pretty extensively, which suggests the capabilities worked in the packages morked just fine with the GO type  (just the unit tests failed, probably due to something hard-coded to a value that is not defined when ```GO=unsigned long```, as was the case in Teko).

Issue #6361, opened 2019-11-26
Several MueLu_ParameterListXXX tests failing in all CUDA builds on 'white', 'ride', 'waterman', and 'vortex' starting 2019-11-23 and other non-CUDA builds starting 3019-12-14


bartlettroscoe:
CC: @trilinos/muelu, @srajama1 (Trilinos Linear Solvers Product Lead), @bartlettroscoe, @fryeguy52, @lucbv 

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-10-27&end=2019-11-26&filtercount=5&showfilters=1&filtercombine=and&field1=groupname&compare1=62&value1=Experimental&field2=buildname&compare2=65&value2=Trilinos-atdm-&field3=testname&compare3=65&value3=MueLu_ParameterList&field4=status&compare4=61&value4=failed&field5=details&compare5=62&value5=Completed%20(Timeout)) the tests:

* `MueLu_ParameterListInterpreterTpetra_MPI_1`
* `MueLu_ParameterListInterpreterTpetra_MPI_4`
* `MueLu_ParameterListInterpreterTpetraHeavy_MPI_4`

in the builds:

* `Trilinos-atdm-waterman_cuda-9.2_fpic_static_opt`
* `Trilinos-atdm-waterman_cuda-9.2_shared_opt`
* `Trilinos-atdm-waterman-cuda-9.2-debug`
* `Trilinos-atdm-waterman-cuda-9.2-opt`
* `Trilinos-atdm-waterman-cuda-9.2-rdc-release-debug`
* `Trilinos-atdm-waterman-cuda-9.2-release-debug`
[...]

PR #6457, opened 2019-12-16
Tpetra: Fix #6455

### LINKED ISSUE #6455: Teko_testdriver_tpetra_MPI_1 failing in ATDM Power/GPU CUDA builds on 'ride' and 'waterman' starting 2019-12-13
CC: @trilinos/teko, @srajama1  (Trilinos Linear Solvers Product Lead), @bartlettroscoe, @fryeguy52, @brian-kelley

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-11-10&end=2019-12-16&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=62&value1=Experimental&field2=buildname&compare2=65&value2=Trilinos-atdm-&field3=testname&compare3=61&value3=Teko_testdriver_tpetra_MPI_1&field4=status&compare4=61&value4=failed) the test:

* `Teko_testdriver_tpetra_MPI_1`

in the builds:

* `Trilinos-atdm-waterman-cuda-9.2-debug`
* `Trilinos-atdm-waterman-cuda-9.2-rdc-release-debug`
* `Trilinos-atdm-waterman-cuda-9.2-release-debug`
* `Trilinos-atdm-white-ride-cuda-10.1-gnu-7.2.0-release-debug`
* `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-debug`
* `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-rdc-release-debug`
* `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-release-debug`

failing every day starting on testing day 2019-12-13.

As shown in [this query](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2019-12-15&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=62&value1=Experimental&field2=buildname&compare2=65&value2=Trilinos-atdm-&field3=buildname&compare3=63&value3=cuda&field4=testname&compare4=61&value4=Teko_testdriver_tpetra_MPI_1), this test only failes in builds that have `Trilinos_ENABLE_DEBUG=ON` set (i.e. with `debug` or `dbg` in the name) and does not fail in any full `opt`/`release` builds.

As shown in [this query](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-11-10&end=2019-12-16&filtercount=5&showfilters=1&filtercombine=and&field1=groupname&compare1=62&value1=Experimental&field2=buildname&compare2=65&value2=Trilinos-atdm-&field3=testname&compare3=61&value3=Teko_testdriver_tpetra_MPI_1&field4=status&compare4=61&value4=failed&field5=testoutput&compare5=97&value5=Tpetra%3A%3ACrsMatrix.*%3A%3AsetAllValues%3A%20myGraph_-.setAllIndices..%20threw%20an%20exception%3A%20CrsGraph%3A%3AsetAllIndices..%3A%20provided%20columnIndices%20are%20not%20sorted%20within%20rows%20on%20at%20least%20one%20process) (click "Show Matching Output" in upper right) all of these failing tests showed:

```
ERROR, rank 0, CrsGraph::setAllIndices(): provided columnIndices are not sorted!
  what():  /home/jenkins/waterman/workspace/Trilinos-atdm-waterman-cuda-9.2-debug/SRC_AND_BUILD/Trilinos/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp:4117:

Throw number = 3

Throw test that evaluated to true: (true)

Tpetra::CrsMatrix<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaUVMSpace> >::setAllValues: myGraph_->setAllIndices() threw an exception: CrsGraph::setAllIndices(): provided columnIndices are not sorted within rows on at least one process.
[waterman1:143826] *** Process received signal ***
[waterman1:143826] Signal: Aborted (6)
[waterman1:143826] Signal code:  (-6)
```

Looking at the new commits pulled that day, for example [here](https://testing-dev.sandia.gov/cdash/viewNotes.php?buildid=5084925#!#note3), it seems very likely that this failure was triggered by the changes in the commit:

```
a7eee78:  Tpetra: setAllIndices verifies sorting in debug
Author: Brian Kelley <bmkelle@sandia.gov>
Date:   Wed Dec 11 10:25:12 2019 -0700

M	packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
M	packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
D	packages/tpetra/core/test/MatrixMatrix/AddProfilingTest.cpp
M	packages/tpetra/core/test/MatrixMatrix/CMakeLists.txt
M	packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```

from the PR #6433 by @brian-kelley.

## Current Status on CDash

* [Status of `Teko_testdriver_tpetra_MPI_1` in all CUDA builds for current testing day](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=62&value1=Experimental&field2=buildname&compare2=65&value2=Trilinos-atdm-&field3=buildname&compare3=63&value3=cuda&field4=testname&compare4=61&value4=Teko_testdriver_tpetra_MPI_1)

## Steps to Reproduce

One should be able to reproduce this failure the machines 'white' (SON) 'ride' (SRN), or 'waterman' (SRC) as described in:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md

The exact commands to reproduce this failure on 'ride' or 'white', for example, should be:

```
$ cd <some_build_dir>/

$ source $TRILINOS_DIR/cmake/std/atdm/load-env.sh \
    Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-debug

$ cmake \
 -GNinja \
 -DTrilinos_ENABLE_TESTS=ON -DTrilinos_ENABLE_Teko=ON \
 $TRILINOS_DIR

$ make NP=16

$ bsub -x -Is -q rhel7F -n 16 ctest -R Teko_testdriver_tpetra_MPI_1
```

brian-kelley:
@trilinos/tpetra 

## Motivation
For transposed operands, Tpetra::MatrixMatrix::add() didn't request the RowMatrixTransposer to sort rows because it can work with unsorted rows. But the transpose graph still reported ``isSorted() == true``. Semi-related to #6405 

## Testing

This was caught due to a test failure in Teko: #6455, though it actually has nothing to do with Teko or Cuda.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -731,7 +731,7 @@ namespace Tpetra {
             const Teuchos::RCP<const map_type>& colMap,
             const typename local_graph_type::row_map_type& rowPointers,
             const typename local_graph_type::entries_type::non_const_type& columnIndices,
-            const Teuchos::RCP<Teuchos::ParameterList>& /* params */) :
+            const Teuchos::RCP<Teuchos::ParameterList>& params) :
     dist_object_type (rowMap)
     , rowMap_(rowMap)
     , colMap_(colMap)
@@ -749,13 +749,16 @@ namespace Tpetra {
     , fillComplete_(false)
     , lowerTriangular_ (false)
     , upperTriangular_ (false)
-    , indicesAreSorted_(true)
     , noRedundancies_(true)
     , haveLocalConstants_ (false)
     , haveGlobalConstants_ (false)
     , sortGhostsAssociatedWithEachProcessor_(true)
   {
     staticAssertions ();
+    if(!params.is_null() && !params->get<bool>("sorted"))
+      indicesAreSorted_ = false;
+    else
+      indicesAreSorted_ = true;
     setAllIndices (rowPointers, columnIndices);
     checkInternalState ();
   }
@@ -766,7 +769,7 @@ namespace Tpetra {
             const Teuchos::RCP<const map_type>& colMap,
             const Teuchos::ArrayRCP<size_t>& rowPointers,
             const Teuchos::ArrayRCP<LocalOrdinal> & columnIndices,
-            const Teuchos::RCP<Teuchos::ParameterList>& /* params */) :
+            const Teuchos::RCP<Teuchos::ParameterList>& params) :
     dist_object_type (rowMap)
     , rowMap_ (rowMap)
     , colMap_ (colMap)
@@ -784,13 +787,16 @@ namespace Tpetra {
     , fillComplete_ (false)
     , lowerTriangular_ (false)
     , upperTriangular_ (false)
-    , indicesAreSorted_ (true)
     , noRedundancies_ (true)
     , haveLocalConstants_ (false)
     , haveGlobalConstants_ (false)
     , sortGhostsAssociatedWithEachProcessor_ (true)
   {
     staticAssertions ();
+    if(!params.is_null() && !params->get<bool>("sorted"))
+      indicesAreSorted_ = false;
+    else
+      indicesAreSorted_ = true;
     setAllIndices (rowPointers, columnIndices);
     checkInternalState ();
   }
@@ -835,7 +841,6 @@ namespace Tpetra {
     , fillComplete_ (false)
     , lowerTriangular_ (false)
     , upperTriangular_ (false)
-    , indicesAreSorted_ (true)
     , noRedundancies_ (true)
     , haveLocalConstants_ (false)
     , haveGlobalConstants_ (false)
@@ -854,6 +859,7 @@ namespace Tpetra {
       "number of rows.  The row Map claims " << rowMap->getNodeNumElements ()
       << " row(s), but the local graph claims " << k_local_graph_.numRows ()
       << " row(s).");
+
     // NOTE (mfh 17 Mar 2014) getNodeNumRows() returns
     // rowMap_->getNodeNumElements(), but it doesn't have to.
     // TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(
@@ -867,6 +873,11 @@ namespace Tpetra {
     TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(
       ! lclInds2D_.is_null () || ! gblInds2D_.is_null (), std::logic_error,
       ": cannot have 2D data structures allocated.");
+
+    if(!params.is_null() && !params->get<bool>("sorted"))
+      indicesAreSorted_ = false;
+    else
+      indicesAreSorted_ = true;
 
     setDomainRangeMaps (domainMap.is_null() ? rowMap_ : domainMap,
                         rangeMap .is_null() ? rowMap_ : rangeMap);
@@ -920,7 +931,6 @@ namespace Tpetra {
     fillComplete_ (false), // not yet, but see below
     lowerTriangular_ (false),
     upperTriangular_ (false),
-    indicesAreSorted_ (true),
     noRedundancies_ (true),
     haveLocalConstants_ (false),
     haveGlobalConstants_ (false),
@@ -936,6 +946,12 @@ namespace Tpetra {
 
     k_lclInds1D_ = lclGraph_.entries;
     k_rowPtrs_ = lclGraph_.row_map;
+
+    if(!params.is_null() && !params->get<bool>("sorted"))
+      indicesAreSorted_ = false;
+    else
+      indicesAreSorted_ = true;
+
     const bool callComputeGlobalConstants =
       params.get () == nullptr ||
       params->get ("compute global constants", true);
@@ -3487,7 +3503,7 @@ namespace Tpetra {
       }
     }
 
-    if (Tpetra::Details::Behavior::debug()) {
+    if (this->isSorted() && Tpetra::Details::Behavior::debug()) {
       // Verify that the local indices are actually sorted
       int notSorted = 0;
       using exec_space = typename local_graph_type::execution_space;
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -755,7 +755,7 @@ namespace Tpetra {
     , sortGhostsAssociatedWithEachProcessor_(true)
   {
     staticAssertions ();
-    if(!params.is_null() && !params->get<bool>("sorted"))
+    if(!params.is_null() && params->isParameter("sorted") && !params->get<bool>("sorted"))
       indicesAreSorted_ = false;
     else
       indicesAreSorted_ = true;
@@ -793,7 +793,7 @@ namespace Tpetra {
     , sortGhostsAssociatedWithEachProcessor_ (true)
   {
     staticAssertions ();
-    if(!params.is_null() && !params->get<bool>("sorted"))
+    if(!params.is_null() && params->isParameter("sorted") && !params->get<bool>("sorted"))
       indicesAreSorted_ = false;
     else
       indicesAreSorted_ = true;
@@ -874,7 +874,7 @@ namespace Tpetra {
       ! lclInds2D_.is_null () || ! gblInds2D_.is_null (), std::logic_error,
       ": cannot have 2D data structures allocated.");
 
-    if(!params.is_null() && !params->get<bool>("sorted"))
+    if(!params.is_null() && params->isParameter("sorted") && !params->get<bool>("sorted"))
       indicesAreSorted_ = false;
     else
       indicesAreSorted_ = true;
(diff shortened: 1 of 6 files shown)


mhoemmen:
@brian-kelley wrote:
> For transposed operands, Tpetra::MatrixMatrix::add() didn't request the RowMatrixTransposer to sort rows because it can work with unsorted rows. But the transpose graph still reported `isSorted() == true`.

Wouldn't this be a bug in `RowMatrixTransposer`?

brian-kelley:
@mhoemmen I guess so. The RowMatrixTransposer lets you ask to not sort indices, but then it calls a CrsGraph constructor that asks for (but doesn't verify) a sorted local graph.

mhoemmen:
@brian-kelley wrote:
> I guess so. The RowMatrixTransposer lets you ask to not sort indices, but then it calls a CrsGraph constructor that asks for (but doesn't verify) a sorted local graph.

There's some reason relating to MueLu setup why the RowMatrixTransposer lets you ask not to sort indices.  Should we pass that reason along to CrsGraph's constructor?

mhoemmen:
@brian-kelley If I remember right, it was @csiefer2 who added that option to RowMatrixTransposer as an optimization, because the matrix in question is some intermediate term that doesn't need sorted rows.

brian-kelley:
@mhoemmen Yeah, if I could add a "bool sorted" parameter to that CrsGraph constructor, that would  completely solve the problem. It could default to true. If we're doing that, it should also be added to the constructors that take individual CRS views and then call setAllIndices().

@bartlettroscoe I agree, I will add a test tomorrow.

mhoemmen:
@brian-kelley wrote:
> Yeah, if I could add a "bool sorted" parameter to that CrsGraph constructor, that would completely solve the problem.

CrsGraph does have constructors that take a ParameterList; that might be a place to add it.  It should default to `true`, since `add` is a public interface (apparently -- Teko is calling it) and thus should return a sensible CrsGraph by default.

brian-kelley:
@mhoemmen By "sensible", do you mean that the CrsGraph ctor should call sortCrsEntries if the parameter list says the input's not sorted?

mhoemmen:
@brian-kelley "Sensible" means that the result of `add` should be sorted by default, while still preserving the option that MueLu wants of having a nonsorted CrsGraph.  I don't like generally how MueLu interacts with Tpetra objects, but that's too big of an issue.

mhoemmen:
@bartlettroscoe I appreciate the commitment to testing, but this change actually makes testing easier by removing the need for test cases.  The only reason I haven't approved yet is because I want to hear from @trilinos/muelu about whether this will affect performance a lot.

brian-kelley:
@mhoemmen After I add support for "sorted" in the ParameterList, I still want to add a new test for add() in the transposed A/B operand cases, because this should have been caught by MatrixMatrix tests.

brian-kelley:
The result of add() having sorted rows was always intended to be an invariant (except when we realized it was broken because of the setAllValues/fillComplete logic, in #6405).

mhoemmen:
@brian-kelley Looks like MueLu is OK with this, so I approved.  @bartlettroscoe , sorry to dismiss your review -- I think this change actually reduces the need for tests, by eliminating a possible state of CrsGraph.

brian-kelley:
@mhoemmen This PR does fix the original issue, but if you'd like I could still add this change to the Graph/Matrix constructors that take local CRS + maps + paramlist:

```
if(!params->get("sorted"))
    isSorted_ = false;
else
    isSorted_ = true;
```

This is probably still worth doing since the RowMatrixTransposer still produces graphs that "lie" about whether they're sorted.

brian-kelley:
Also, it looks like the only failure in PR testing here was due to #6361, which is not related to this commit.

bartlettroscoe:
CC: @trilinos/framework 

> Also, it looks like the only failure in PR testing here was due to #6361, which is not related to this commit.

@brian-kelley, I fear that until someone addresses #6361, none of these PRs that trigger the testing of MueLu will be able to be merged.

brian-kelley:
@mhoemmen Actually, I need to do one more thing, you don't need to review yet.

packages/tpetra/core/src/Tpetra_RowMatrixTransposer_def.hpp
```diff
@@ -271,12 +273,18 @@ createTransposeLocal (const Teuchos::RCP<Teuchos::ParameterList>& params)
   RCP<const export_type> myExport = origImport.is_null () ?
     Teuchos::null : rcp (new export_type (*origImport));
 
+  RCP<Teuchos::ParameterList> graphParams = Teuchos::null;
+  if(!sort) {
```
mhoemmen: I think this is OK.

packages/tpetra/core/src/Tpetra_RowMatrixTransposer_def.hpp
```diff
@@ -271,12 +273,18 @@ createTransposeLocal (const Teuchos::RCP<Teuchos::ParameterList>& params)
   RCP<const export_type> myExport = origImport.is_null () ?
     Teuchos::null : rcp (new export_type (*origImport));
 
+  RCP<Teuchos::ParameterList> graphParams = Teuchos::null;
```
mhoemmen: `RCP`'s default value is null, so you don't need the assignment here.

brian-kelley:
@bartlettroscoe One new Tpetra test ("locally_unsorted") for MatrixMatrix::add() now exercises the case where the inputs are unsorted and have the same column map. There was never a problem with that kernel, but it needed test coverage. 

I can confirm that the other new test ("tranposed_b") would have caught the original bug, where the graph lied about being sorted.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -749,13 +749,16 @@ namespace Tpetra {
...
     , lowerTriangular_ (false)
     , upperTriangular_ (false)
-    , indicesAreSorted_(true)
     , noRedundancies_(true)
     , haveLocalConstants_ (false)
     , haveGlobalConstants_ (false)
     , sortGhostsAssociatedWithEachProcessor_(true)
   {
     staticAssertions ();
+    if(!params.is_null() && !params->get<bool>("sorted"))
```
mhoemmen: If `params` is not null but "sorted" isn't in it, then one-argument `get` will throw.  (Two-argument `get` sets a default value if the parameter doesn't exist.)  You could try this instead:
```c++
if (! params.is_null () && params->isParameter("sorted") &&
  ! params->get<bool>("sorted")) // etc.
```

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -784,13 +787,16 @@ namespace Tpetra {
...
     , lowerTriangular_ (false)
     , upperTriangular_ (false)
-    , indicesAreSorted_ (true)
     , noRedundancies_ (true)
     , haveLocalConstants_ (false)
     , haveGlobalConstants_ (false)
     , sortGhostsAssociatedWithEachProcessor_ (true)
   {
     staticAssertions ();
+    if(!params.is_null() && !params->get<bool>("sorted"))
```

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -936,6 +946,12 @@ namespace Tpetra {
 
     k_lclInds1D_ = lclGraph_.entries;
     k_rowPtrs_ = lclGraph_.row_map;
+
+    if(!params.is_null() && !params->get<bool>("sorted"))
```

brian-kelley:
@bartlettroscoe It looks a random failure in MueLu_ParameterListInterpreter (on intel 17, but not the other builds) was the only issue with the last PR test. If that doesn't get resolved soon could we just manually merge this PR? It fixes a bug and a test failure.

bartlettroscoe:
> @bartlettroscoe It looks a random failure in MueLu_ParameterListInterpreter (on intel 17, but not the other builds) was the only issue with the last PR test. If that doesn't get resolved soon could we just manually merge this PR? It fixes a bug and a test failure.

@brian-kelley, I don't have the permissions to do that.  I am no different than any other Trilinos developer.  Until someone resolves this failing test as part of #6361 (and reverting the PR #6432 would likely do this), I don't think anyone's PR is going to get merged.  You will need to take this up with @trilinos/framework.

PR #6458, opened 2019-12-17
Changes in Piro and ROL to enable the full-space optimization in Piro through SimOpt interface

mperego:
Changes:
ROL: added Thyra Adapters for Constraint and Objective SimOpt classes
ROL: removed non needed calls to update function in ConstraintSimOpt
Piro: enabled full-space optimization using ROL SimOpt interface
Piro: modified/added tests to test new capability
Piro: removed code related to Moocho, OptiPack and GlobiPack

@trilinos/rol 
## Testing
test:
 Piro_AnalysisDriverTpetra_MPI_4
 this includes 7 sub-tests checking different configurations of the new full-space capability as well as backward compatibility


packages/piro/src/Piro_NOXSolver_Def.hpp
```diff
@@ -161,12 +162,14 @@ void Piro::NOXSolver<Scalar>::evalModelImpl(
               std::ostringstream ss; ss << "Parameter Vector Index " << i;
               const int p_ind = rolParams.get(ss.str(), i);
               const auto paramNames = *this->getModel().get_p_names(p_ind);
-              this->observer->parameterChanged(paramNames[0]);
+              if(Teuchos::nonnull(this->observer))
+                this->observer->parameterChanged(paramNames[0]);
             }
             optimizationParams.template set<bool>("Optimization Variables Changed", false);
```
mhoemmen: You shouldn't need `template` here if `optimizationParams` is just a `Teuchos::ParameterList`, because its type does not depend upon a template parameter, and the type of the argument is known (`bool`).  Also, you don't need to say `bool` here; it's inferred from the input argument `false`.

packages/piro/src/Piro_NOXSolver_Def.hpp
```diff
@@ -200,11 +210,23 @@ void Piro::NOXSolver<Scalar>::evalModelImpl(
           "Nonlinear solver failed to converge");
     }
 
+    if(appParams->isSublist("Analysis")){
+      Teuchos::ParameterList& analysisParams = appParams->sublist("Analysis");
+      if(analysisParams.isSublist("Optimization Status")) {
+        analysisParams.sublist("Optimization Status").set<bool>("State Solve Converged", solve_status.solveStatus==Thyra::SOLVE_STATUS_CONVERGED);
+        analysisParams.sublist("Optimization Status").set<bool>("Compute State", false);
```
mhoemmen: You don't need `<bool>` here; the compiler will deduce the type from `false`.

packages/piro/src/Piro_NOXSolver_Def.hpp
```diff
@@ -222,459 +244,462 @@ void Piro::NOXSolver<Scalar>::evalModelImpl(
...
+              "adjoint sensitivities." << std::endl);
+          auto dgdx  = modelOutArgs.get_DgDx(j).getMultiVector();
+          if (dgdx != Teuchos::null) {
+            int num_cols = dgdx->domain()->dim();
+
+            // (2) Calculate xbar multivector from -(J^{-T}*dg/dx)
+
+            auto xbar = createMembers(dgdx->range(), num_cols);
+            {
+              const auto timer = Teuchos::rcp(new Teuchos::TimeMonitor(*Teuchos::TimeMonitor::getNewTimer("Piro::NOXSolver::evalModelImpl::calc xbar")));
```
mhoemmen: Please don't handle TimeMonitor by pointer if you can avoid it (which you easily can here).  TimeMonitor is a scope guard.  Its constructor starts the timer, and its destructor stops it.  You already create a new scope here.  Why not use the scope guard as a scope guard, directly?

packages/piro/test/Main_AnalysisDriver_Tpetra.cpp
```diff
@@ -155,19 +157,16 @@ int main(int argc, char *argv[]) {
...
-      // Thyra::ModelEvaluator is the base class of all Piro::Epetra solvers
       const RCP<Thyra::ModelEvaluatorDefaultBase<double>> piro = solverFactory.createSolver(piroParams, ModelWithSolve);
-      // END Builder
 
       // Call the analysis routine
       RCP<Thyra::VectorBase<double>> p;
       status = Piro::PerformAnalysis(*piro, analysisParams, p);
 
-      if (Teuchos::nonnull(p)) { //p might be null if the packages ROL, Moocho, OptiPack are not enabled
+      if (Teuchos::nonnull(p)) { //p might be null if the packages ROL is not enabled
```
mhoemmen: "if the package ROL is not enabled"

packages/piro/test/MockModelEval_A_Tpetra.cpp
```diff
@@ -320,13 +320,18 @@ void MockModelEval_A_Tpetra::evalModelImpl(
...
+    f_out->putScalar(0.0);
     for (int i=0; i<myVecLength; i++) {
       int gid = x_in->getMap()->getGlobalElement(i);
 
-      if (gid==0) // f_0 = (x_0)^2 - p_0
-       f_out->getDataNonConst()[i] = x[i] * x[i] -  p[0];
-      else // f_i = x_i^2 - (i+p_1)^2 (for i != 0)
-       f_out->getDataNonConst()[i] = x[i] * x[i] - (gid + p[1])*(gid + p[1]);
+      if (gid==0) {// f_0 = (x_0)^3 - p_0
+       f_out->getDataNonConst()[i] = x[i] * x[i] * x[i] -  p[0];
```
mhoemmen: Please lift the `getDataNonConst()` calls outside the loop.  Those are expensive, and will get VERY expensive once we get rid of the UVM assumption.

packages/piro/test/MockModelEval_A_Tpetra.cpp
```diff
@@ -349,8 +364,14 @@ void MockModelEval_A_Tpetra::evalModelImpl(
     dfdp_out->putScalar(0.0);
     for (int i=0; i<myVecLength; i++) {
       const int gid = x_in->getMap()->getGlobalElement(i);
-      if  (gid==0) dfdp_out->getVectorNonConst(0)->getDataNonConst()[i] = -1.0;
-      else         dfdp_out->getVectorNonConst(1)->getDataNonConst()[i] =  -2.0* (gid + p[1]);
+      if  (gid==0) {
+        dfdp_out->getVectorNonConst(0)->getDataNonConst()[i] = -1.0;
```
mhoemmen: Please don't call `getDataNonConst` in an inner loop.

packages/piro/test/MockModelEval_A_Tpetra.cpp
```diff
@@ -349,8 +364,14 @@ void MockModelEval_A_Tpetra::evalModelImpl(
     dfdp_out->putScalar(0.0);
```
mhoemmen: `putScalar` runs on device, but then you write to the vector on host.  This will be quite slow.

packages/piro/src/Piro_NOXSolver_Def.hpp
mperego: Thanks, I'll fix this. I remeber that I had to use `template` in a getter to avoid some mysterious build error on some machine. However here I don't need the bool, as you say.

packages/piro/src/Piro_NOXSolver_Def.hpp
mperego: OK!

packages/piro/src/Piro_NOXSolver_Def.hpp
mperego: Will do.

packages/piro/test/MockModelEval_A_Tpetra.cpp
mperego: Thanks. These are tiny vectors in a test. Anyway I'll fix that.

packages/piro/test/MockModelEval_A_Tpetra.cpp
mperego: This is really a tiny test.

mhoemmen: @mperego Developers have the habit of imitating what they see -- tests get copied and pasted a lot.

mperego: I understand. But I'm not willing to move this test on device.

mhoemmen: I'm not saying you need to do that.  It's just that `getDataNonConst` reserves the right to make a copy of all the Vector's device data, and copy it to host.  It doesn't do that right now because Vector uses UVM, but it could do that in the future.

mperego: @mhoemmen  I see. I moved the getDataNonConst outside of the loops.
Can you approve if you are fine with that?

mperego:
@rppawlo @mhoemmen can you approve this? Thanks!

mperego:
@rppawlo I think that the ROL team is taking care of that. 
Anyway, let me make sure that @dpkouri is fine with the modification in ROL

mperego:
@rppawlo I talked to Denis and Drew, they are going to take care of the changes. We also decided to wait for some changes in ROL to avoid re-computations, so I removed one of the commits. I think this can be safely merged.

mperego:
@rppawlo  Yes, those are changes in Thyra adapters, mainly additions, and Denis and Drew are fine with those and they will take care of pushing them back to the ROL repo.

mperego:
@rppawlo Would you mind approving this again? Thanks!

PR #6465, opened 2019-12-18
MueLu: extracting coarse coordinates from region hierarchy to create AMG coarse solver

### LINKED ISSUE #5503: MueLu: possibly use AMG V-cycle as coarse solver in region MG problems
## Enhancement
@trilinos/muelu 

Enable the region MG algorithm to use another (unstructured) AMG hierarchy with just one sweep through the V-cycle instead of applying a direct solver on the coarsest region MG level.

A particular use case is the following: region MG is applied to partially structured meshes. Regional coarsening has reduced the coarse size of each region to a single (or a few) "element", but there are many of these very small coarse regions now. From a global perspective, this looks like an unstructured mesh. Hence, AMG is the perfect candidate to further coarsen such a problem. 

## Definition of Done

- [x] Add command line arguments to en-/disable this feature
- [x] Compute nullspace of composite coarse level operator
- [x] Pass coordinates along to enable rebalancing
- [x] Setup the additional AMG hierarchy, starting from the composite coarse level operator
- [x] On the region MG coarse level: replace direct solver by one sweep through an AMG V-cycle
- [x] Add tests

## Possible Solution

The region MG hierarchies store should be able to provide regional nullspaces on the coarse level. They can be combined to form a composite coarse level nullspace to be used in the setup of the additional AMG hierarchy.

## Interested Parties
@lucbv @rstumin @pohm01

lucbv:
@trilinos/muelu 

## Motivation
The goal is to use MueLu as a coarse grid solver for region multigrid.
This requires us to extract coordinates from the last level of the region multigrid hierarchy and then bring them to composite format before passing them to a second MueLu hierarchy.
Once this is done, repartitioning can be used in the coarse AMG hierarchy.

## Related Issues

* Closes #5503
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback
No external stakeholder for this work

## Testing
Local tests have been performed, this code is guarded by experimental flags.


lucbv:
This is a work in progress, a rebase against develop need to be done and the construction of coordinates when solving a system of PDEs still need to be implemented.

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
```diff
@@ -280,41 +250,73 @@ void MakeCoarseLevelMaps2(const int maxRegPerGID,
...
+                                     compRowMap->getIndexBase(),
+                                     compRowMap->getComm());
+  }
+  compCoarseCoordinates = MultiVectorFactory::Build(compCoordMap, regCoarseCoordinates[0]->getNumVectors());
+  TEUCHOS_ASSERT(Teuchos::nonnull(compCoarseCoordinates));
+
+  // The following looks like regionalToComposite for Vector
+  // but it is a bit different since we do not want to add
+  // entries in the coordinate MultiVector as we would for
+  // a solution or residual vector.
```
mayrmt: We could add an argument `Xpetra::CombineMode combineMode` defaulting to `Xpetra::ADD` to `regionalToComposite()`. Then, we can call it here with a combine mode `Xpetra::INSERT`.

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
```diff
@@ -587,15 +590,23 @@ void createRegionHierarchy(const int maxRegPerProc,
...
+        if(mapComp->getComm()->getRank() == 0) level->print(std::cout, MueLu::Extreme);
+        regCoarseCoordinates[j] = level->Get<RCP<realvaluedmultivector_type> >("Coordinates2", MueLu::NoFactory::get());
+      }
+
+      regProlong[l][j]  = level->Get<RCP<Matrix> >("P", MueLu::NoFactory::get());
       regMatrices[l][j] = level->Get<RCP<Matrix> >("A", MueLu::NoFactory::get());
 
-      regRowMaps[l][j] = Teuchos::rcp_const_cast<Xpetra::Map<LO,GO,NO> >(regMatrices[l][j]->getRowMap()); // ToDo (mayr.mt) Should we rather copy?
-      regColMaps[l][j] = Teuchos::rcp_const_cast<Xpetra::Map<LO,GO,NO> >(regMatrices[l][j]->getColMap()); // ToDo (mayr.mt) Should we rather copy?
+      regRowMaps[l][j]  = Teuchos::rcp_const_cast<Xpetra::Map<LO,GO,NO> >(regMatrices[l][j]->getRowMap()); // ToDo (mayr.mt) Should we rather copy?
```
mayrmt:
```suggestion
      regRowMaps[l][j]  = Teuchos::rcp_const_cast<Xpetra::Map<LO,GO,NO> >(regMatrices[l][j]->getRowMap());
```

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
```diff
@@ -587,15 +590,23 @@ void createRegionHierarchy(const int maxRegPerProc,
...
+      regColMaps[l][j]  = Teuchos::rcp_const_cast<Xpetra::Map<LO,GO,NO> >(regMatrices[l][j]->getColMap()); // ToDo (mayr.mt) Should we rather copy?
```
mayrmt:
```suggestion
      regColMaps[l][j]  = Teuchos::rcp_const_cast<Xpetra::Map<LO,GO,NO> >(regMatrices[l][j]->getColMap());
```

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
mayrmt: We can get rid of this comment, as we now know, that we don't have to copy. Other than that, this change here (and in the next line) is just a change in white spaces.

packages/muelu/research/regionMG/examples/structured/amg_1dof_rebalance.xml
```diff
@@ -0,0 +1,54 @@
+<ParameterList name="MueLu">
+
+  <!--
+    For 3D elasticity, these are the recommended settings for MueLu.
+  -->
```
mayrmt: This comment does not apply here, as this file is not suitable for elasticity. We should remove it.

packages/muelu/src/Transfers/GeneralGeometric/MueLu_GeometricInterpolationPFactory_def.hpp
```diff
@@ -159,6 +160,10 @@ namespace MueLu {
       coarseCoordinates->replaceMap(coarseCoordsMap);
 
       Set(coarseLevel, "Coordinates", coarseCoordinates);
+
+      if(pL.get<bool>("keep coarse coords")) {
+        coarseLevel.Set<RCP<realvaluedmultivector_type> >("Coordinates2", coarseCoordinates, NoFactory::get());
```
mayrmt: @lucbv 👍 Doesn't look to complicated now. One could be wondering, why we weren't able to keep coordinates during one of our earlier attempts. 🤣

lucbv:
Rebased against develop after deprecated code purge, this worked better than I feared it would :D

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
lucbv: no we cannot, the main problem is that the current regionalToComposite does not work with MultiVectors, we would need an overload that does the exact same thing as the current Vector based version...

packages/muelu/research/regionMG/examples/structured/amg_1dof_rebalance.xml
lucbv: Done

packages/muelu/src/Transfers/GeneralGeometric/MueLu_GeometricInterpolationPFactory_def.hpp
lucbv: Yes once I looked at it afterwards I thought it was rather easy, of course once you have learned how things work everything looks easy. It did take some trial and error to arrive to that!

lucbv:
It's almost done but there is something fishy happening in the new test that I would like to understand before merging this!

lucbv:
@mayrmt 
This should allow us to run the Poisson problem without issue, there is still something a bit strange in the Elasticity code path though.
I have merged this since it is experimental and I don't want to risk having to resolve too many conflicts after the holidays.

PR #6473, opened 2019-12-19
Phalanx shared fields 

rppawlo:
Adds field memory recycling for kokkos views


rppawlo:
@etphipp - could you re-approve? Had to fix an issue with gcc 4.8 compiler.

rppawlo:
Thanks @mhoemmen !

rppawlo:
@ikalash - This PR adds a new capability to phalanx that significantly improves memory use. It is disabled by default. The changes have not caused any failures in in panzer/drekar/empire/charon. So it is probably safe to merge without testing albany. But if you wanted to test, I can hold off merging. What would you prefer?

ikalash:
This sounds like a great improvement to Phalanx!  Tagging @jewatkins who is likely interested in this.  If the new option is off by default, it shouldn't cause issues for Albany so I'm OK merging the PR.  How would we enable the option if we wanted to try it in Albany?

rppawlo:
To enable memory sharing, export the following variable into your environment:
```
export PHX_ENABLE_SHARED=1
```

I'm pretty sure enabling the memory sharing will break Albany. There are some short cuts for performance in the AD gather that need to be changed. That's why the feature currently defaults to off. Once we clean up the apps, we can change the default to on.

ikalash:
Ok, thanks, Roger.  @jewatkins it may be interesting to play with this option in Albany.  It will be good to learn more about it in a few weeks at the Albany User Meeting.

PR #6477, opened 2019-12-19
Fix misc. warnings

brian-kelley:
Fix some random warnings I've been seeing:
-using uninitialized var
-misleading indentation
-set-but-unused var
-catch polymorphic object (std::exception) by value

@trilinos/galeri 
@trilinos/tpetra 
@trilinos/isorropia 
@trilinos/epetraext 
## Motivation

No issue for these

### DIFF
--- a/packages/belos/src/BelosGmresPolyOp.hpp
+++ b/packages/belos/src/BelosGmresPolyOp.hpp
@@ -620,11 +620,11 @@ namespace Belos {
     try {
       gmres_iter->iterate();
     }
-    catch (GmresIterationOrthoFailure e) {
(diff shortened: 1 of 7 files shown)


kddevin:
Thanks, @brian-kelley  
Can you add which compiler generated the warnings?

brian-kelley:
@kddevin GCC 9.1.0, which is pretty new and seems to be more strict on warnings than other compilers.

brian-kelley:
@kddevin Thanks!

lucbv:
Does it makes sense to updates the warning as error list if we start doing this type of work?

Issue #6478, opened 2019-12-19
Teuchos:  stacked timer very slow on big timer dumps


bathmatt:
## Enhancement
@trilinos/teuchos 

The stacked timers are very slow on big timer dumps.  This is partly due to the findName algorithm, partly due to output.  We should explore this.

PR #6479, opened 2019-12-19
Fixed it so you don't have to go through all the names if it can't be…

bathmatt:
… at this point

@trilinos/<teamName>

### DIFF
--- a/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
+++ b/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
@@ -56,6 +56,12 @@ StackedTimer::LevelTimer::findBaseTimer(const std::string &name) const {
 BaseTimer::TimeInfo
 StackedTimer::LevelTimer::findTimer(const std::string &name, bool& found) {
   BaseTimer::TimeInfo t;
+  auto full_name = get_full_name();
+  auto level=0;
+  if (full_name.size() > name.size())
+    return t;
+  if ( strncmp(full_name.c_str(), name.c_str(), full_name.size()))
+    return t;
   if (get_full_name() == name) {
     t = BaseTimer::TimeInfo(this);
     found = true;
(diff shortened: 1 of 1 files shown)


packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -56,6 +56,12 @@ StackedTimer::LevelTimer::findBaseTimer(const std::string &name) const {
 BaseTimer::TimeInfo
 StackedTimer::LevelTimer::findTimer(const std::string &name, bool& found) {
   BaseTimer::TimeInfo t;
+  auto full_name = get_full_name();
+  auto level=0;
```
rppawlo: level looks to be unused.

bathmatt: you passed the test, very good...  (it was used before I realized my code could be much simpler)

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -56,6 +56,11 @@ StackedTimer::LevelTimer::findBaseTimer(const std::string &name) const {
 BaseTimer::TimeInfo
 StackedTimer::LevelTimer::findTimer(const std::string &name, bool& found) {
   BaseTimer::TimeInfo t;
+  auto full_name = get_full_name();
+  if (full_name.size() > name.size())
```
mhoemmen: I think this is fine in context of the `strncmp` call below.  On the other hand, now you're doing two string comparisons.  I would have written the code like this:
```c++
const std::string full_name = get_full_name ();
if (full_name != name) {
  return t;
}
t = BaseTimer::TimeInfo(this);
found = true;
```

mhoemmen:
Does this fix #6478 or is it just a part?

PR #6488, opened 2019-12-21
TpetraTSQR: Run on NVIDIA GPUs; ensure correct testing; refactor

mhoemmen:
@trilinos/tpetra @trilinos/belos @trilinos/anasazi @iyamazaki @ndellingwood 

## Motivation

1. Add TSQR back-end that uses cuSOLVER where appropriate to run on NVIDIA GPUs.
2. Use LAPACK's BLAS 3 optimizations for matrices with larger numbers of columns.
3. Make sure that tests are actually testing what we want to test.
4. Refactor implementation to simplify integration with Kokkos.
5. Remove dead code: e.g., the TBB TSQR implementation (that hasn't been tested since ~2010).

## Stakeholder Feedback

This is mostly funded by the Pressio project, who specifically requested GPU support and optimizations for matrices with more columns.  @fnrizzi may wish to comment.  ECP ExaWind may also have an interest in TSQR for CA-GMRES.  @iyamazaki may wish to comment.

## Context

I first added this TSQR implementation to Trilinos in 2010.  One learns something after a decade.  Trilinos has also changed a lot, in particular due to the introduction of Kokkos (what Trilinos developers would call Kokkos >= 2.0; 1.0 was Chris Baker's 2009 project, and 0.0 was a library of computational kernels circa 2004-5).  The lack of Kokkos at the time explains the odd mix of custom matrix views and raw pointers in the interface.  I haven't tried to fix all that here.


TPLsList.cmake
```diff
@@ -58,6 +58,8 @@ TRIBITS_REPOSITORY_DEFINE_TPLS(
   yaml-cpp        "cmake/TPLs/"    EX
   Peano           "cmake/TPLs/"    EX
   CUDA            "${${PROJECT_NAME}_TRIBITS_DIR}/core/std_tpls/"    PT
+  CUBLAS          "cmake/TPLs/"    PT
```
mhoemmen: Yes, I'm adding new TPLs.  On the other hand, these are trivial to enable when CUDA is enabled: just set `TPL_ENABLE_CUBLAS=ON` etc.  I will work on automating that.

packages/tpetra/tsqr/CMakeLists.txt
```diff
@@ -8,23 +8,38 @@ TRIBITS_SUBPACKAGE(TSQR)
 # Enabled by default (unless disabled explicitly at the command line)
 # if Teuchos is built with complex arithmetic support.
 TRIBITS_ADD_OPTION_AND_DEFINE(
-  KokkosTSQR_ENABLE_Complex
-  HAVE_KOKKOSTSQR_COMPLEX
+  ${PACKAGE_NAME}_ENABLE_Complex
+  HAVE_TPETRATSQR_COMPLEX
   "Enable complex arithmetic (std::complex<T>) support for TSQR.  This is currently ON if Teuchos_ENABLE_COMPLEX is ON.  The default behavior may change as we migrate TSQR to depend on new Kokkos.  New Kokkos does not currently support complex arithmetic, but this will change."
```
mhoemmen: Oops, the comment needs to change.

mhoemmen:
Summary of test failures:
  - CUDA 9.2 (2908): Intrepid2 test (timed out; red herring)
  - GCC 7.2.0 (3354): Build warnings (as errors)
  - GCC 4.9.3 SERIAL (3506): PASSED
  - Intel 17.0.1 (5077): packages/tpetra/tsqr/src/libkokkostsqr.so.12.17: undefined reference to `TSQR::Impl::Lapack<double>::~Lapack()' (probably a bug in Intel 17; I'll try working around by defining destructor explicitly in .cpp file)
  - GCC 4.8.4 (5252): Combine tests don't initialize Kokkos::OpenMP (parallel_reduce)

I'm working on fixes now.

mhoemmen:
I just pushed some fixes.  Let's try this again.  The "MpiAndKokkosScope" changes are to help diagnose an issue I'm seeing on one particular test computer, where I run `mpiexec -np 4` but `MPI_Comm_size` outputs 1 instead of 4.

mhoemmen:
Intel 17 build (5078) failed, again with the same link errors:
```
packages/tpetra/tsqr/src/libkokkostsqr.so.12.17: undefined reference to TSQR::Impl::Lapack<double>::~Lapack()
```
This was after my change that declared `~Lapack() override = default;`.  My guess is that Intel 17 has a bug.  I will try doing this: `~Lapack() override {}`.

mhoemmen:
I just pushed a work-around for the Intel 17 bug.  I also made the `Tpetra::MultiVector` specialization of `TsqrAdaptor` [sic] use device memory if the TSQR implementation wants it.  This should have no practical effect, since `Tpetra::MultiVector` currently uses a single UVM allocation for both "host" and "device" memory.  However, this will help us support future use cases, for instance if we remove the UVM assumption from `MultiVector` or need to support other architectures that have two separate memory spaces.

mhoemmen:
Just fixed a typo.

mhoemmen:
@alanw0 wrote:
> In stk/sierra we try to make sure that each commit is self-sufficient, meaning it builds and passes tests so that we preserve the ability to bisect. I can't tell if these commits maintain that, or if that is a goal here.

That was absolutely the goal.  Almost all commits have that property at least on the computer on which I was testing.  That's the main reason why there are so many commits -- I wanted to move always from a state of "passing tests" to another state of "passing tests."

mhoemmen:
@kddevin wrote:
> We'll eventually want to use KokkosKernels for all cublas-like operations, rather than call cublas directly. Newer platforms will not have cublas, and responsibility for on-node operations should lie in KokkosKernels. We'll make conversion to KokkosKernels a goal for later in the FY.

I'm just waiting for a complete kokkos-kernels BLAS.  I'll be happy to switch once it's ready and I know that it's actually calling TPLs.

mhoemmen:
@kddevin Also, the point of these changes is cuSOLVER, not cuBLAS.  Not all platforms will necessarily have an accelerator-based LAPACK.

mhoemmen:
Thanks all! :-D  @kddevin btw I figured out how to make TriBITS automatically detect CUBLAS and CUSOLVER -- by using `TRIBITS_TPL_TENTATIVELY_ENABLE(CUBLAS)` etc. in `packages/tpetra/tsqr/cmake/Dependencies.cmake`.  I'll put up a new PR with that change soon.

srajama1:
> @kddevin wrote:
> 
> > We'll eventually want to use KokkosKernels for all cublas-like operations, rather than call cublas directly. Newer platforms will not have cublas, and responsibility for on-node operations should lie in KokkosKernels. We'll make conversion to KokkosKernels a goal for later in the FY.
> 
> I'm just waiting for a complete kokkos-kernels BLAS. I'll be happy to switch once it's ready and I know that it's actually calling TPLs.

The BLAS/LAPACK calls are added on a need-only basis. We are happy to add the calls needed by Tpetra. Please file a GitHub issue with the calls needed.

kddevin:
@mhoemmen In the TSQR code, I see the following LAPACK calls:

- UNGQR  (in compute_explicit_Q_lwork and compute_explicit_Q)
- GEQRF  (in compute_QR_lwork and compute_QR)
- UNMQR  (in apply_Q_factor_lwork and apply_Q_factor)
- LARFG  (in factor_inner and factor_pair)
- GESVD  (in reveal_R_rank)
- POTRF  (in factor)
- LARNV  (in fill_buffer)

Are all of these needed in KokkosKernels?  I see only the following reimplemented for CUDA in Tsqr_Impl_CuSolver.cpp:

- apply_Q_factor, apply_Q_factor_lwork
- compute_QR, compute_QR_lwork
- compute_explicit_Q. compute_explicit_Q_lwork

Does that mean we would need to request only the following from KokkosKernels?

- UNGQR  (in compute_explicit_Q_lwork and compute_explicit_Q)
- GEQRF  (in compute_QR_lwork and compute_QR)
- UNMQR  (in apply_Q_factor_lwork and apply_Q_factor)

Thanks @mhoemmen .

kddevin:
Continuing to @mhoemmen 
Or is the NodeTsqr class the more appropriate level of abstraction for KokkosKernels (with CuSolverNodeTsqr being one implementation)?

mhoemmen:
@kddevin Are LAPACK functions actually in-scope for kokkos-kernels?  It seemed a bit too much to ask for LAPACK when there were BLAS functions left to implement.

kddevin:
Given the name of the class (NodeTsqr), should the capability be in KokkosKernels rather than in Tpetra?  The comments say the TSQR implementation is not Tpetra-specific and, indeed, no Tpetra data structures are used in the implementation.  Probably the entire capability should be moved into KokkosKernels or one of the solver packages (Belos, ShyLu, etc.).

srajama1:
> @kddevin Are LAPACK functions actually in-scope for kokkos-kernels? It seemed a bit too much to ask for LAPACK when there were BLAS functions left to implement.

Yes, we do support some LAPACK functionality at the team level. It is driven by user requests. It is going to take a long time to achieve complete support. 

Sorry, @mhoemmen I edited your comment by mistake instead of quoting it.

kddevin:
@srajama1 Is a NodeTsqr appropriate for KokkosKernels or a better fit for something like ShyLu?

mhoemmen:
@kddevin I don't really care where things live, as long as they get enabled and tested by default.  There's a risk putting stuff too far downstream in packages like ShyLU, that historically had (have?) components that weren't enabled or tested by default.  Also, TSQR is out of scope for ShyLU.

srajama1:
@kddevin I don't have particular preference for where TSQR lives. That is primarily the decision of the developer given everything else is equal.  However, ShyLU is not the right place (ShyLU_Node is mainly sparse factorizations and solvers needed by DD methods). Kokkos Kernels might be ok, Belos might be ok. Essentially, the questions are

1. Where are we going to put other variations like CholQR ?
2. Where would an user go looking for this kernel logically ?
3. What dependencies need to be satisfied ?

@iyamazaki might have some thoughts as well.

mhoemmen:
TSQR depends on MPI as well as Kokkos. This makes it out of scope for kokkos-kernels.

srajama1:
That leaves Belos or Tpetra then. May be a silly question, if it depends on MPI, why is it NodeTSQR ? I guess TSQR depends on MPI and NodeTSQR doesn't. I assumed Karen was asking about NodeTSQR.

kddevin:
There's NodeTSQR and DistTSQR.  Probably doesn't make sense to separate them.  I will file a KokkosKernels request for the needed support.

kddevin:
https://github.com/kokkos/kokkos-kernels/issues/567
Requested KokkosKernels support for cusolve calls.

PR #6495, opened 2019-12-23
zoltan:  minor cleanup

kddevin:
Removing unneeded #ifdef ZOLTAN_HIER; it was always defined anyway so checking it didn't make sense (and we never tested without it).
Removing unneeded Allreduces; we were overly cautious for a case that never gets exercised on any of our machines.

@trilinos/zoltan 

## Motivation

Minor cleanup and simplication.
Remove two allreduces that aren't needed.

## Testing

Tested on mac with clang.

### DIFF
--- a/packages/zoltan/src/zz/zz_struct.c
+++ b/packages/zoltan/src/zz/zz_struct.c
@@ -135,19 +135,6 @@ int proc;
 
   zz->ZTime = Zoltan_Timer_Create(ZOLTAN_TIMER_DEF);
 
-  /* Test that size_t is uniform on all processors */
-  if (communicator != MPI_COMM_NULL) {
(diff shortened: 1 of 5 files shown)


packages/zoltan/src/zz/zz_struct.c
```diff
@@ -135,19 +135,6 @@ int proc;
 
   zz->ZTime = Zoltan_Timer_Create(ZOLTAN_TIMER_DEF);
 
-  /* Test that size_t is uniform on all processors */
```
mhoemmen: I'm pretty sure that other Trilinos code sends and receives arrays of `size_t` without checking whether `size_t` has uniform size on different processes, so yes, I agree that this is an ill-tested case we need not support.

PR #6496, opened 2019-12-24
TSQR: Automatically detect CUBLAS & CUSOLVER TPLs; improve TPL handle wrappers

mhoemmen:
@trilinos/tpetra @trilinos/belos @trilinos/anasazi 
@iyamazaki @vqd8a @ndellingwood 

## Motivation

1. Automatically detect CUBLAS and CUSOLVER TPLs.  Now users don't have to set `TPL_ENABLE_CUBLAS=ON` etc. explicitly.
2. Improve cuBLAS and cuSOLVER handle wrappers so that the kokkos-kernels team can adopt the design pattern directly.

## Related Issues

* Superseded by https://github.com/trilinos/Trilinos/pull/6583
* Related to https://github.com/trilinos/Trilinos/pull/6488, https://github.com/kokkos/kokkos-kernels/pull/552, https://github.com/kokkos/kokkos-kernels/pull/554

## Testing

I improved tests for cuBLAS and cuSOLVER handle wrappers.  I also tested this with a CUDA build.  It does actually automatically detect the "CUBLAS" and "CUSOLVER" TPLs in that case.  I don't want to force those to be enabled in a CUDA build just yet, so the test doesn't require them if CUDA is enabled.


mhoemmen:
The autotester appears broken today; let's try again.

bartlettroscoe:
@mhoemmen, looks like the CMake logic is faulty for non-CUDA builds.  These TPLs should only be tentatively enabled for CUDA builds.

mhoemmen:
@bartlettroscoe wrote:
>  These TPLs should only be tentatively enabled for CUDA builds.

If a TPL is tentatively enabled, and the FindTPLWhatever.cmake file says that the TPL needs CUDA, then wouldn't the TPL be disabled?  Just checking -- that was how I understood it.

bartlettroscoe:
> If a TPL is tentatively enabled, and the FindTPLWhatever.cmake file says that the TPL needs CUDA, then wouldn't the TPL be disabled? Just checking -- that was how I understood it.

@mhoemmen, currently a TriBITS TPL does not have a dependence on other TPLs.  I am actually going to be working to address in TriBITSPub/TriBITS#63 (which I am supposed to be working on right now).

mhoemmen:
@bartlettroscoe I'm working on this now; will have a PR pretty soon.

mhoemmen:
Superseded by https://github.com/trilinos/Trilinos/pull/6583.  Thanks all!

Issue #6440, opened 2019-12-12
Zoltan2: Zoltan2_mjTest_geomgen_nomjparts_MPI_3 is randomly failing


MicheldeMessieres:
## Bug Report
@trilinos/zoltan2 

Zoltan2_mjTest_geomgen_nomjparts_MPI_3 is randomly failing.

https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-11-11&end=2019-12-11&filtercount=3&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Pull%20Request&field2=testname&compare2=61&value2=Zoltan2_mjTest_geomgen_nomjparts_MPI_3&field3=status&compare3=61&value3=failed

### Description
The new zoltan2 tests with PR #5332  run twice and compare results so they can catch indeterminate behavior. Several times this causes a test failure. One time there was a seg fault. One time there was a file error I am not sure about yet.

bartlettroscoe:
Info on random failures ...

It looks like one of these randomly failing tests `Zoltan2_mjTest_geomgen_nomjparts_MPI_3` is impacting the standard post CI build as shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-11-11&end=2019-12-11&filtercount=3&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Continuous&field2=buildname&compare2=61&value2=Linux-GCC-4.8.4-MPI_RELEASE_DEBUG_SHARED_PT_OPENMP_CI&field3=status&compare3=61&value3=failed) showing 2 failures in since 2019-12-04.  And it looks like this test has killed several Trilinos PR iterations (5 total) since 2019-11-21 as shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-11-11&end=2019-12-11&filtercount=3&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Pull%20Request&field2=testname&compare2=61&value2=Zoltan2_mjTest_geomgen_nomjparts_MPI_3&field3=status&compare3=61&value3=failed).

MicheldeMessieres:
@bartlettroscoe @kddevin Query shows this test has not had any failures in the past 2 months so I think we can close this issue:

[...]

PR #6514, opened 2020-01-02
Zoltan2: Fixes for #6440

MicheldeMessieres:
This will resolve indeterminate behavior in the Multijagged calculation.
This is expected to resolve most of the cases where 6440 failed. However in one case, cdash showed a segfault and it's not clear yet why that would occur or if it would be impacted by this fix. This will at least get the test running in a consistent manner.

@trilinos/zoltan2

## Motivation

Eliminates indeterminate behavior. Will fix most but perhaps not all intermittent failures reported in #6440.

### DIFF
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -4766,9 +4766,13 @@ mj_create_new_partitions(
     }
   }, total_on_cut);
 
-  Kokkos::View<mj_lno_t *, device_t> track_on_cuts(
-    "track_on_cuts", // would do WithoutInitialization but need last init to 0
-    total_on_cut + 1); // extra index to use for tracking
+  Kokkos::View<mj_lno_t *, device_t> track_on_cuts;
+  if(total_on_cut > 0) {
+    track_on_cuts = Kokkos::View<mj_lno_t *, device_t>(
+      "track_on_cuts", // would do WithoutInitialization but need last init to 0
+      total_on_cut + 1); // extra index to use for tracking
+  }
+
 
   // here we need to parallel reduce an array to count coords in each part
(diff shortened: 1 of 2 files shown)


packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -4829,6 +4830,12 @@ mj_create_new_partitions(
   delete [] reduce_array;
 #endif
 
+  // the last member is utility used for atomically inserting the values.
+  // Sorting here avoids potential indeterminancy in the partitioning results
+  auto track_on_cuts_sort = Kokkos::subview(track_on_cuts,
+    std::pair<mj_lno_t, mj_lno_t>(0, track_on_cuts.size() - 1));
```
mhoemmen: `Kokkos::subview` takes a half-exclusive range `(start, end+1)`.  Also, what if `track_on_cuts.size()` is zero?

MicheldeMessieres:
@mhoemmen Thanks. This array is always size 1 or greater and the last element is used for index tracking so not sorted. Though I could add a check to skip the sort when size = 1 and no sorting is necessary. However I've temporarily closed the PR as the bug just stopped replicating for me in develop.  I need to investigate this further.

MicheldeMessieres:
@kddevin PR fixes indeterminate result causing the MultiVector pass and BasicVectorAdapter pass to give different results on the cuts and fail the comparison part of the test.

I've also fixed another issue in the GeometricGenerator random number generation. The statics were not handled in a thread safe way and the original coordinate distribution was changing run to run for OpenMP. I think performance for GeometricGenerator is not very important since it's for tests only so I've simplified this and just removed the statics.

kddevin:
Thanks, @MicheldeMessieres.   
You are correct that performance of the GeometricGenerator does not matter.  If its use of OpenMP is problematic, it is OK to serialize it all.

PR #6526, opened 2020-01-06
MueLu: updating the nightly configuration from geminga's cuda build

lucbv:
@trilinos/muelu 

## Motivation
Enabling Tpetra_INST_INT_INT so that Tpetra builds at least one global ordinal type.
We might want to change that and enable Tpetra_INST_INT_LONG_LONG instead as it is Tpetra's default and will likely be what users do...


cmake/ctest/drivers/geminga/ctest_linux_nightly_mpi_release_muelu_kokkos_refactor_cuda_geminga.cmake
```diff
@@ -75,29 +75,30 @@ SET(Trilinos_PACKAGES MueLu Xpetra Amesos2)
...
-    "-DTrilinos_ENABLE_COMPLEX:BOOL=ON"
-    "-DTeuchos_ENABLE_COMPLEX:BOOL=OFF"
-    "-DTpetra_INST_COMPLEX_DOUBLE:BOOL=OFF"
-    "-DTpetra_INST_COMPLEX_FLOAT:BOOL=OFF"
-    "-DTpetra_INST_SERIAL=ON"
+  "-D Trilinos_ENABLE_EXPLICIT_INSTANTIATION:BOOL=ON"
+    "-D Tpetra_INST_INT_LONG_LONG:BOOL=OFF"
+    "-D Tpetra_INST_INT_INT:BOOL=ON"
+    "-D Trilinos_ENABLE_COMPLEX:BOOL=OFF"
+    "-D Teuchos_ENABLE_COMPLEX:BOOL=OFF"
```
mhoemmen: You don't need to set `Teuchos_ENABLE_COMPLEX`; the Trilinos flag should work.  Does it not?

lucbv: Yes that's right, I can also remove the other two lines:
```
    "-D Tpetra_INST_COMPLEX_DOUBLE:BOOL=OFF"
    "-D Tpetra_INST_COMPLEX_FLOAT:BOOL=OFF"
```
for the same reason.

mhoemmen: @lucbv yes please :-) There are also separate options `Trilinos_ENABLE_COMPLEX_DOUBLE` and `Trilinos_ENABLE_COMPLEX_FLOAT`, if you just want one but not the other.

PR #6531, opened 2020-01-06
MueLu ParameterList: Output fix (second attempt)

cgcgcg:
@trilinos/muelu 

## Motivation
* Catch missing output from ranks != 0
* Stop testing the interpreter for complex scalar type.


cgcgcg:
The first version of this PR introduced some random failures. Locally, I do not see them anymore, but I'll retest a couple of times to make sure.

Issue #6532, opened 2020-01-07
TeuchosCore_testTeuchosTestForTermination_X_MPI_4 tests randomly timing out/hanging in new cee-rhel6 openmpi-4.0.1 builds starting 2020-01-04


bartlettroscoe:
CC: @trilinos/teuchos, @bartlettroscoe

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

Likely due to a defect in this OpenMPI 4.0.1 install not correctly terminating MPI job when one of the ranks is terminated.  Next: Let run for a few weeks to get some more data then decide what to do.

## Description

As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-01-01&end=2020-01-06&filtercount=5&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-cee-rhel6&field2=buildname&compare2=63&value2=openmpi-4.0.1&field3=groupname&compare3=62&value3=Experimental&field4=status&compare4=62&value4=passed&field5=status&compare5=62&value5=notrun) the tests:

* `TeuchosCore_testTeuchosTestForTermination_0_MPI_4`
* `TeuchosCore_testTeuchosTestForTermination_2_MPI_4`
* `TeuchosCore_testTeuchosTestForTermination_3_MPI_4`

in the new builds:

* `Trilinos-atdm-cee-rhel6_clang-5.0.1_openmpi-4.0.1_serial_static_opt`
* `Trilinos-atdm-cee-rhel6_gnu-7.2.0_openmpi-4.0.1_serial_shared_opt`

are randomly timing out.

These are most likely hanging when one of the processes calls abort but it seems that OpenMPI 4.0.1 is not properly terminating the other processes.

## Current Status on CDash

[...]

PR #6533, opened 2020-01-07
Promote new cee-rhel6 openmpi-4.0.1 builds to ATDM group (ATDV-237)

bartlettroscoe:
I triaged the one randomly failing test in #6532.  I think it is appropriate to promote these builds given the other builds in the ATDM group are not 100% clean (and no one seems to care that much).

SPARC is using these builds in production so they need to be promoted.


bartlettroscoe:
Thanks @mhoemmen!

PR #6535, opened 2020-01-07
MueLu: Exposing aggregation algorithm via easy parameterlist

csiefer2:
Testing: RHEL6/OpenMP


cgcgcg:
@csiefer2 You can be the guinea pig for rebasing the gold files :-)

csiefer2:
@cgcgcg IKR?

@william76 Comments on naming?

william76:
@csiefer2 I don't really have any preferences.  VB_BIT was the vertex-based variant that used a bitvector to mark which colors were already taken (the forbidden array).  I kept the general naming convention from the D1 coloring with that and don't really have any energy about naming as long as it's reasonably descriptive of the variant...

PR #6536, opened 2020-01-07
make 'cts1' env the default on CTS-1 systems and remove 'serrano'

jeforster:
Make 'cts1' env the default on CTS-1 systems and remove the old 'serrano' env

See [ATDV-280](https://sems-atlassian-srn.sandia.gov/browse/ATDV-280)


cmake/std/atdm/utils/get_known_system_info.sh
```diff
@@ -101,19 +100,19 @@ elif [[ $realHostname == "waterman"* ]] ; then
   hostnameMatch=waterman
   hostnameMatchSystemName=waterman
 
-# Specifically named cts1 systems (currently maps to 'serrano' env)
+# Specifically named cts1 systems
```
bartlettroscoe: @jeforster, please just delete this entire block of code based on 'hostname'.  The 'cts1' env is already detected below by looking at `SNLSYSTEM`.

bartlettroscoe:
@jeforster, see, we are already adding support for 'cts1' at:

* https://github.com/trilinos/Trilinos/blob/7530661aaf79ecdaa41d0d926d8a8855b821f497/cmake/std/atdm/utils/get_known_system_info.sh#L143

We just need to remove all traces of the old 'serrano' env.

cmake/std/atdm/utils/get_known_system_info.sh
jeforster: @bartlettroscoe You are correct.  I had forgotten that we were also checking SNLSYSTEM.  I took this part out.

bartlettroscoe:
Hum, this PR was submitted 5 hours ago and the Trilinos PR tester has not even consider starting to run builds/tests yet.

PR #6539, opened 2020-01-07
Intrepid2: Allow using Kokkos with deprecated code

masterleinad:
@trilinos/intrepid2

## Motivation
Using `Intrepid2::Impl::CellTools::Serial::mapToReferenceFrame()` when `Kokkos` has deprecated code disabled fails. That functions calls `Intrepid2::Kernels::Serial::copy()` with a rank-1 `Kokkos::View` object which assumes a rank-2 `Kokkos::View` object. However, accessing entries in `Kokkos::View` objects via the call operator is only allowed for matching ranks. It seems that this function is only used in  `Intrepid2::Impl::CellTools::Serial::mapToReferenceFrame()`. Hence, it seems like modifying the copy function to take a rank-1 `Kokkos::View` looks like another possible solution.

## Testing
The change relaxes the constraints on using `Intrepid2::Kernels::Serial::copy()`. Hence, there should be no noticeable change.


mhoemmen:
@trilinos/intrepid2

masterleinad:
Thanks @mhoemmen!

PR #6544, opened 2020-01-08
Panzer: Adding domain EXTERNAL to DomainEvaluator 

seamill:
@trilinos/panzer 

## Motivation

We need this to simplify the handling of interior sideset boundary conditions in EMPIRE.

## Testing

Added test to ensure the proper indexes were selected with EXTERNAL domain selected


seamill:
@rppawlo If you don't mind reviewing this. We're trying to get interior boundary conditions running in EMPIRE and this should make it a bit easier. Let me know if you have a better name than EXTERNAL... I tried UN/NOTOWNED, but it didn't sound right.

seamill:
@rppawlo Do I need to add the retest flag to restart testing, or is that done automatically when the queue is empty?

PR #6551, opened 2020-01-08
Panzer: Bugfix for sideset definitions on processor boundaries

seamill:
@trilinos/panzer

## Motivation

This is a bug fix where the LocalMeshInfo objects were not properly loading sideset information on processor boundaries due to the side's being 'owned' by only one process. Now all connected processors will see the sideset (required by discontinuous FEM).

## Testing

Added test that checked that a shared sideset is accessible from two adjoining ranks.


seamill:
@rppawlo If you want to review this. It should only affect EMPIRE.

seamill:
@rppawlo Aye, I'll bring them down in size. Does 12K apiece work for you (currently 56K)? Seems a bit large for 16 cells, but oh well. The inline mesh method doesn't actually allow us to test this. I'll add the journal file.

seamill:
@rppawlo Ready for you to look again. The mesh is as small as I can make it. I can't seem to read cubit meshes into panzer in 1D or 2D without causing issues with topology types.

seamill:
@rppawlo Sorry, but I had to fix a warning that triggered a fail with one of the compilers. Could you review this again?

Issue #6554, opened 2020-01-09
Phalanx_mda_data_layout_MPI_1 test failing in build Trilinos-atdm-mutrino-intel-opt-openmp-KNL starting 2020-01-04


bartlettroscoe:
CC: @trilinos/phalanx , @mperego (Trilinos Discretizations Product Lead)

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-12-10&end=2020-01-08&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-mutrino-intel-opt-openmp-KNL&field2=testname&compare2=61&value2=Phalanx_mda_data_layout_MPI_1&field3=site&compare3=61&value3=mutrino) the test:

* `Phalanx_mda_data_layout_MPI_1`

in the build:

* `Trilinos-atdm-mutrino-intel-opt-openmp-KNL`

started failing on testing day 2020-01-06.

There are numerous failures (O(90) failures).  Some examples of the types of failures seen include:

```
[...]

PR #6562, opened 2020-01-10
Phalanx: fix data layout test #6554

rppawlo:
Fix for issue #6554

### DIFF
--- a/packages/phalanx/test/DataLayout/MDALayout.cpp
+++ b/packages/phalanx/test/DataLayout/MDALayout.cpp
@@ -66,6 +66,21 @@ PHX_EXTENT(Ordinal6)
 PHX_EXTENT(Ordinal7)
 PHX_EXTENT(Ordinal8)
 
+namespace PHX {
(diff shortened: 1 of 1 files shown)


rppawlo:
> So this is a namespace issue? Wonder why this was a problem with the KNL backend by not the HSW backend with the same Intel compiler?

@bartlettroscoe - No, this is a problem with name demangling on knl. The default print function is:
```
namespace PHX {
  /// Default print function for a user defined object, typically used for user defined MDField extents. Users can specialize this.
  template<typename T> std::string print()
  {return Teuchos::demangleName(typeid(T).name());}
}
```
The unit test was failing on a string compare that expected a demangled name. Apparently the intel compiler fails to support demangling for knl but the same compiler works fine for host. I could have fixed the test by constructing the gold comparison string with PHX::print(), but instead chose to write template specializations of print for the extent objects to fix the test. I intended the print command to be specialized so users can shorten the field names during debugging. This fix allowed me to add another example of that.

rppawlo:
@bartlettroscoe - Thanke for approving!

PR #6565, opened 2020-01-10
Restore old behavior of Trilinos_CXX_COMPILER_FLAGS

jjwilke:
ProjectCompilerPostConfig.cmake can no longer be used because we don't want any CMake to depend on raw Makefiles. To avoid breaking downstream projects that rely on the Trilinos_CXX_COMPILER_FLAGS variable, this restores the old behavior.


ndellingwood:
@jjwilke the recent kokkos and nvcc_wrapper changes will need to go into the release-candidate-3.0 branch of kokkos, and the Trilinos changes into the kokkos-promotion branch of Trilinos, I'll need to re-snapshot both kokkos and kokkos-kernels into the updated kokkos-promotion branch of Trilinos (I can add the Trilinos change in this PR to the kokkos-promotion branch prior to creating the updated release candidate).

ndellingwood:
Oop, looks like the Trilinos changes were just line changes probably left during testing.

jjwilke:
Closed in favor of PR #6569

PR #6568, opened 2020-01-13
Revert "EXPLORE: Comma-sep output in id ranges"

gsjaardema:
This reverts commit 35551398b71daa4b6014065c8a8a2207184f23e2.

Revert "EXPLORE: Use suplib convert for initial output"

This reverts commit 735faf0bf301336600dcb2b98cce5dc1a7f68bae.

Revert "SUPLIB: Add function to convert int to thousands-separated string"

This reverts commit 13330ac3e99b0a6bffd89fad63220bb8143ce04d.

@trilinos/seacas

## Motivation

A few months ago, added a `convert` function which would output fortran integers with thousands separated by commas (1,234,567 vs 1234567) to make it easier for users to 
visually parse large integers.   The function `convert` was a C++ function called by fortran
which works well on some platform/compilers, but is difficult on other platforms/compilers.

Based on this, it was decided to revert the capability until a more robust and portable solution 
is developed.  This commit reverts the capability.


gsjaardema:
Note that the commits mentioned in the message above refer to commits to the SEACAS github repository and not the Trilinos repository.  These commits (and others) were snapshotted into Trilinos some time after the SEACAS commits.

gsjaardema:
@bartlettroscoe Yes, reverted changes on SEACAS repository and then created patch with `git format-patch` moved over to Trilinos repository and applied via `git am`

packages/seacas/libraries/suplib/convert.C
```diff
@@ -1,19 +0,0 @@
-#include <fmt/format.h>
-
-extern "C" {
-#if defined(ADDC_)
-void convert_(char *str, int64_t *value, int64_t *length, int len)
```
mhoemmen: If you want to try this again at some point, best practice would be to use Fortran's `ISO_C_BINDING` to declare the function to have C binding.  That way, you wouldn't need to worry about underscores or how Fortran passes strings (some Fortran ABIs pass strings as a pointer and a length, etc.).

gsjaardema: Yes, at some point all of the C/Fortran interactions should be converted to `ISO_C_BINDING`, but the priority never makes it to the top...

The main issue with this change was use of C++ in a Fortran application...

PR #6571, opened 2020-01-13
Automatic snapshot commit from seacas at 191567ad4b

gsjaardema:

At commit:

commit 191567ad4bbf005c6128ac938fb0f5c3bb61b167
Date:    Mon Jan 13 09:13:03 2020 -0700
Summary: Revert "EXPLORE: Comma-sep output in id ranges"

* conjoin: eliminate mixed (stderr, stdout) output of similar data
* exodiff: Better diagnostic for sideset side member mismatch
* explore: fixed bug in sideset values output; if subsetted, was outputting wrong values
* aprepro: clang-format changes
* exodus: update element documentation for cubic elements
  -- initial support for szip compression (in progress in NetCDF)
  -- modify ex_put_init of nodal variables to eliminate duplicated code
* IOSS: move file path creation code to Utils
  -- Update property description to include setting via env variable
  -- Clean up face generator code to make a little clearer and more efficient
  -- Add cubic elements (edge4, hex32, hex64, quad12, quad16, tet16, tet40, tri9, tri13, wedge24, wedge52)
  -- Modify model summary output to reduce width (eliminate lots of "Number of")
  -- Catch 2.10.2 -> Catch 2.11.1
  -- Handle mismatch between CGNS and IOSS/Exodus connectivity for hex27
  -- lib::fmt 6.0.0 -> 6.1.2
  -- generated mesh -- fix sideset face topology logic
  -- skinner: work with cgns mesh, auto determine database type, better help and diagnostic output
  -- io_info: handle sideset|nodeset with no dist factors
  -- set locale based on environment
* exodus.py: fix a few bugs, add missing functions, better import of ctypes
* exomerge.py: fix some bugs in transition from python-2 to python-3

@trilinos/seacas


packages/seacas/libraries/ioss/src/fmt/format-inl.h
```diff
@@ -1061,7 +1079,9 @@ namespace internal {
   template <typename T>
   int format_float(T value, int precision, float_specs specs, buffer<char> &buf)
   {
+#if !(defined(__INTEL_COMPILER) && __INTEL_COMPILER < 1800)
     static_assert(!std::is_same<T, float>(), "");
```
mhoemmen: Huh, can you actually convert an instance of `is_same` to `bool`?  That doesn't look right.  I would write this:
```cpp
static_assert(! std::is_same<T, float>::value, "T is not float");
```

gsjaardema: Thanks.  Looks like `std::is_same<...>()` returns a bool as of C++17, but for C++14 or earlier, the `::value` is needed.  I will fix and let lib::fmt project know.

mhoemmen: @gsjaardema Huh, you're right!  The object instance is convertible to `bool` (which is what you're relying on) as of C++11, and it has `bool operator()` [as of C++14](https://en.cppreference.com/w/cpp/types/is_same).  I would prefer `std::is_same<U, `V>::value` for C++11 code or `std::is_same_v<U, V>` for C++17 code, but that's correct syntax.

gsjaardema:
Thanks @mhoemmen , sorry for the churn.  Will try to do better next time.  On the plus side, all of the changes to the lib::fmt library were accepted into their development version, so should be cleaner at next version update.

mhoemmen:
@gsjaardema No need to be sorry; we're a team! :-D

PR #6584, opened 2020-01-15
Tpetra::CrsGraph: Improve exception messages on insert failure

mhoemmen:
@trilinos/tpetra @tjfulle 

## Motivation

Help with Sierra SD debugging by improving exception messages on insert failure, to show information helpful for users, like the row's capacity.

## Stakeholder Feedback

Feature requested by Sierra.

## Testing

Tested locally, Linux, GCC 4.9.3, debug build.  These are error messages.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -1987,16 +1987,24 @@ namespace Tpetra {
     inp_view_type inputInds(inputGblColInds, numInputInds);
     size_t numInserted = Details::insertCrsIndices(lclRow, k_rowPtrs_,
       this->k_gblInds1D_, numEntries, inputInds, fun);
-    TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC(
-        numInserted == Teuchos::OrdinalTraits<size_t>::invalid(),
-        std::runtime_error,
-        "There is not enough capacity to insert " << inputInds.size()  << " " <<
-        "indices in to row " << lclRow <<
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -1987,16 +1987,24 @@ namespace Tpetra {
...
+
+    const bool insertFailed =
+      numInserted == Teuchos::OrdinalTraits<size_t>::invalid();
+    if(insertFailed) {
+      constexpr size_t ONE (1);
+      const int myRank = this->getComm()->getRank();
+      std::ostringstream os;
+      os << "On MPI Process " << myRank << ": Not enough capacity to "
+        "insert " << numInputInds
+         << " ind" << (numInputInds != ONE ? "ices" : "ex")
```
brian-kelley: You could just compare against "1U" instead of using a ONE constant, right?

brian-kelley: (you don't have to change it, I'm just curious)

mhoemmen: @brian-kelley That would fix signedness and any resulting warnings, but I prefer to compare against something known to be the same type.  Integer types have caused too much pain in the past ;-) .

mhoemmen:
@kddevin wrote:
> For completeness (for future software quality reviews), please add to your description how you tested the changes (which platform, which tests exercise them, etc.). Having this info in each PR will make future SQ reviews easier.

Done.

PR #6586, opened 2020-01-15
Tpetra: Making MMM_Timings play nicer with stacked timers

csiefer2:
Tests:  Code guarded by #define.  RHEL6.


packages/tpetra/core/ext/TpetraExt_MatrixMatrix_ExtraKernels_def.hpp
```diff
@@ -1256,7 +1256,7 @@ static inline void mult_R_A_P_newmatrix_LowThreadGustavsonKernel(CrsMatrixStruct
 
   #ifdef HAVE_TPETRA_MMM_TIMINGS
         MMsort.~TimeMonitor();
```
mhoemmen: Calling the destructor explicitly, let allowing the object to persist, is kinda not OK.  I have to check the standard to see if it might double-call the destructor; if it does, it would restart the timer, which would result in incorrect timings.

csiefer2: I didn't want to muck with more things than I had to.  But that probably needs to go, you are correct.

PR #6587, opened 2020-01-15
Tpetra: Fix export/import to local graph/matrix

tjfulle:
@mhoemmen reports that certain import/export operations fail during the copy and permute phase.  This PR is an attempt to fix the problem.

### DIFF
--- a/packages/tpetra/core/test/CrsGraph/CrsGraph_StaticImportExport.cpp
+++ b/packages/tpetra/core/test/CrsGraph/CrsGraph_StaticImportExport.cpp
@@ -208,8 +208,112 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(CrsGraph, ImportToStaticGraph, LO, GO, NT)
   }
 }
 
+
+TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(CrsGraph, ImportToStaticGraphLocal, LO, GO, NT)
+{
+  // Set up Tpetra typedefs.
+  using map_type = Tpetra::Map<LO, GO, NT>;
+  using graph_type = Tpetra::CrsGraph<LO, GO, NT>;
+  using device_type = typename NT::device_type;
+  using execution_space = typename device_type::execution_space;
+  const char prefix[] = "ImportToStaticGraphLocal: ";
+
+  Tpetra::Details::Behavior::disable_verbose_behavior();
+
+  auto comm = getDefaultComm();
+  const auto my_rank = comm->getRank();
+
+  auto loc_success = 1; // to be revised below
+  auto gbl_success = 0; // output argument
+
+  // The global number of rows in the matrix A to create.  We scale
+  // this relative to the number of (MPI) processes, so that no matter
+  // how many MPI processes you run, every process will have 10 rows.
+  const Tpetra::global_size_t num_gbl_inds = 10 * comm->getSize();
+  const GO idx_base = 0;
+
+  // Construct a Map that is global, but puts all the equations on MPI Proc 0.
+  RCP<const map_type> proc_zero_map;
+  {
+    const size_t num_loc_inds = (my_rank == 0) ? num_gbl_inds : 0;
+    proc_zero_map = rcp(new map_type(num_gbl_inds, num_loc_inds, idx_base, comm));
+  }
+
+  // Create a sparse graph using procZeroMap.
+  auto A = generate_crs_graph<LO,GO,NT>(proc_zero_map);
+  comm->barrier();
+
+  // We've created a sparse matrix that lives entirely on Process 0.
+  // Now we want to distribute it over all the processes.
+  //
+  // Redistribute the matrix.  Since both the source and target Maps
+  // are one-to-one, we could use either an Import or an Export.  If
+  // only the source Map were one-to-one, we would have to use an
+  // Import; if only the target Map were one-to-one, we would have to
+  // use an Export.  We do not allow redistribution using Import or
+  // Export if neither source nor target Map is one-to-one.
+  RCP<graph_type> B;
+  {
+    using export_type = Tpetra::Export<LO,GO,NT>;
+    export_type exporter(proc_zero_map, proc_zero_map);
+    comm->barrier();
+
+    Tpetra::Details::Behavior::enable_verbose_behavior();
+    // Make a new sparse graph whose row map is the global Map.
+    out << prefix << "Creating empty graph from global map.\n";
+    B = rcp(new graph_type(proc_zero_map, 0, Tpetra::StaticProfile));
+    out << prefix << "Empty graph from proc_zero_map map created.\n";
+
+    // Redistribute the data, NOT in place, from graph A (which lives
+    // entirely on Proc 0) to graph B (which is distributed evenly over
+    // the processes).
+    out << prefix << "Performing export operation\n";
+    B->doExport(*A, exporter, Tpetra::INSERT);
+    out << prefix << "Export operation done.\n";
+  }
+  B->fillComplete();
+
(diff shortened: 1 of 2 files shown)


tjfulle:
The first commit does not actually change any code, it just adds a new test that creates a graph/matrix on processor 0 and exports/imports it to another (empty) graph/matrix also on processor 0.  However, this test does not fail.

@mhoemmen, what kind of use case is showing the error?

tjfulle:
Should `copyAndPermute` resize the graph/matrix?  Resizing is *only* done in import/export right now.

packages/tpetra/core/test/CrsGraph/CrsGraph_StaticImportExport.cpp
```diff
@@ -208,8 +208,112 @@ TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(CrsGraph, ImportToStaticGraph, LO, GO, NT)
...
+    // Make a new sparse graph whose row map is the global Map.
+    out << prefix << "Creating empty graph from global map.\n";
+    B = rcp(new graph_type(proc_zero_map, 0, Tpetra::StaticProfile));
+    out << prefix << "Empty graph from proc_zero_map map created.\n";
+
+    // Redistribute the data, NOT in place, from graph A (which lives
+    // entirely on Proc 0) to graph B (which is distributed evenly over
+    // the processes).
+    out << prefix << "Performing export operation\n";
+    B->doExport(*A, exporter, Tpetra::INSERT);
```
mhoemmen: Sierra SD actually reported that this threw an exception.  I had them wrap the doExport call in a try-catch, and used an all-reduce to figure out whether doExport threw.

shardest: Our failing test within Sierra/SD is fairly small - is there a way for us to output the graph and maps so that we can either send them to you or construct a simpler test that exhibits the same behavior? I haven't looked at your code, but it seems that you are using only a single MPI rank, whereas our example had two.

mhoemmen: @shardest This example runs with 1-4 MPI processes.  (You have to look at the CMakeLists.txt file in this directory to see that.  The test source code doesn't usually specify the required number of processes.)

mhoemmen: @shardest wrote:
> ... [i]s there a way for us to output the graph and maps so that we can either send them to you or construct a simpler test that exhibits the same behavior?

Yes -- either way is good, but if you include `MatrixMarket_Tpetra.hpp`, there are a bunch of writers in there, including for Map and CrsGraph.  Thanks!

tjfulle:
I just pushed some more tests.  They are simple, but should exercise the requirement of resizing the objects (graph or matrix) before `copyAndPermute`.  The tests are simple:

- fill `A` on proc 0
- export `A` to `B` that lives on proc 0
- export `B` to `C` that is distributed
- export `C` to `D` that is distributed

Everything still passes.

tjfulle:
@shardest, I will create a Tpetra unit test with your matrix that is showing failure as soon as you can get it to me.

shardest:
We wrote the maps and graph from Salinas to files, which turned out to be quite small. We then constructed a test to read in those files and try to recreate the problem. It fails reading the graph file, apparently due to some inconsistency about 0/1-based indexing.

We added a ".txt" extension to the program because it wouldn't let us upload it with a .C extension.


[input_graph.txt](https://github.com/trilinos/Trilinos/files/4073034/input_graph.txt)
[root_map.txt](https://github.com/trilinos/Trilinos/files/4073035/root_map.txt)
[range_map.txt](https://github.com/trilinos/Trilinos/files/4073036/range_map.txt)
[domain_map.txt](https://github.com/trilinos/Trilinos/files/4073037/domain_map.txt)

[TpetraGraphTest.C.txt](https://github.com/trilinos/Trilinos/files/4073039/TpetraGraphTest.C.txt)

tjfulle:
That's great, thanks @shardest!  I'll use your files to cook up a Tpetra unit test.

tjfulle:
@shardest: update: I got a modified version of your test to run with the input files you attached and it passes.  I'll call you tomorrow (1/21/2020) to discuss the test implementation and see if it reflects what you need/see.

shardest:
@tjfulle - This is a good time to work on this together if possible - Clark is in town. We are in the SD team room, but we have meetings 9-10, 11-12, and 1:30-3.

mhoemmen:
@tjfulle @shardest cool, let me know if you want a hand!

tjfulle:
Thanks @mhoemmen.  I sent an email to @shardest earlier.  I'll call right now.

shardest:
Attached the output generated by this code Mark wrote:

//Tpetra::MatrixMarket::Writer<CrsMatrix>::writeSparseGraphFile("input_graph",*constraintNodeGraphSub, "constraintNodeGraphSub", "input graph");
    {
       std::ostringstream os;
       ThrowAssert(! constraintNodeGraphSub->getRowMap().is_null());
       const auto& rowMap = *(constraintNodeGraphSub->getRowMap());

       Teuchos::Array<GO> gblColInds;
       const LO lclNumRows (rowMap.getNodeNumElements());
       for (LO lclRow = 0; lclRow < lclNumRows; ++lclRow) {
           size_t numEnt (constraintNodeGraphSub->getNumEntriesInLocalRow(lclRow));
           if (size_t(gblColInds.size()) < numEnt) {
               gblColInds.resize(numEnt);
           }
           const GO gblRow = rowMap.getGlobalElement(lclRow);
           ThrowAssert(gblRow != Teuchos::OrdinalTraits<GO>::invalid());
           constraintNodeGraphSub->getGlobalRowCopy(gblRow, gblColInds(), numEnt);
           for (const GO gblCol : gblColInds) {
               os << gblRow << " " << gblCol << std::endl;
           }
       }
       std::ofstream outFile("input_graph");
       Tpetra::Details::gathervPrint(outFile, os.str(), *Comm);
    }


[input_graph.txt](https://github.com/trilinos/Trilinos/files/4093793/input_graph.txt)

mhoemmen:
@shardest We're a bit confused by this example.  It looks like you want to redistribute from a distributed graph whose row Map is what you are calling "range Map," to a "gathered graph" whose row Map is what you are calling "root Map."  However, the "range Map" has GIDs in it (e.g., 20, 21, etc.) that do not appear in the "root Map."  Thus, it seems like there is something wrong with the Maps.

mhoemmen:
@tjfulle , I just worked with @shardest this afternoon on this, and it looks like it was a bug in SD, not in Tpetra.  SD was constructing an Import from the "sub" graph to the "root" graph, but it was using the "sub" graph's range Map instead of its row Map.  Its row Map is quite likely overlapping.  When we changed the Import to use the "sub" graph's row Map instead, the test passed.

shardest:
[column_map.txt](https://github.com/trilinos/Trilinos/files/4106875/column_map.txt)
[row_map.txt](https://github.com/trilinos/Trilinos/files/4106876/row_map.txt)

In case you think it would be useful to generate a unit test based on this data, I've attached the other two maps associated with the sub graph.

tjfulle:
That's great, thanks @mhoemmen!  This is reason 873 to remove `DynamicProfile`!

I'll go ahead and merge the tests that I added in the PR since the do explicitly exercise `copyAndPermute`ing in to an initially empty matrix.

kddevin:
Nice!  Thanks, @tjfulle and @mhoemmen

mhoemmen:
@tjfulle Shall we merge this one?

PR #6590, opened 2020-01-16
Teuchos:  allow unit tests to proceed with unrecognized command line arguments

kddevin:
@trilinos/teuchos 

This change allows teuchos' unit test framework to run tests even when they see unrecognized command line arguments.  Unrecognized command line arguments now generate a warning, not an error.

## Motivation

Some platforms (e.g., doom) require command line arguments like --kokkos-ndevices=4 to be specified.  That is,
```
srun --mpi=pmi2 -u -N 1 -n 4 packages/tpetra/core/test/Map/TpetraCore_Map_UnitTests.exe --kokkos-ndevices=4
```
will work but 
```
srun --mpi=pmi2 -u -N 1 -n 4 packages/tpetra/core/test/Map/TpetraCore_Map_UnitTests.exe
```
does not run correctly (problems in MPI).

But Teuchos' unit test framework reported an error and failed when unrecognized command line arguments like --kokkos-ndevices=4 were encountered.

This change allows the test framework to run and report a warning, not an error, when such command line arguments are encountered.

## Testing

Tested in slurm scripts on doom (see examples above).
Also tested on cee-lan, ATDM modules; all tests passed.
```
source /home//code/Trilinos/cmake/std/atdm/load-env.sh  gnu-7.2.0-openmp-release-debug

cmake \
  -GNinja \
 -DMPI_EXEC=${MPI_EXEC} \
 -DTrilinos_ENABLE_TESTS=ON \
 -DTrilinos_ENABLE_Teuchos:BOOL=ON \
 -DTrilinos_ENABLE_ALL_OPTIONAL_PACKAGES:BOOL=ON \
 -DTrilinos_ENABLE_SECONDARY_TESTED_CODE:BOOL=ON \
 -DTPL_ENABLE_Matio=OFF \
 -DTPL_ENABLE_X11=OFF \
```

### DIFF
--- a/packages/teuchos/core/src/Teuchos_UnitTestRepository.cpp
+++ b/packages/teuchos/core/src/Teuchos_UnitTestRepository.cpp
@@ -187,7 +187,7 @@ public:
   int testCounter;
 
   InstanceData()
-    :clp(false),
+    :clp(false,false),
      showTestDetails(SHOW_TEST_DETAILS_TEST_NAMES),
 #if defined(HAVE_TEUCHOS_GLOBALLY_REDUCE_UNITTEST_RESULTS)
      globallyReduceUnitTestResult(true),


bartlettroscoe:
@kddevin, instead of  `--kokkos-ndevices=4` can you just set the env var `KOKKOS_NUM_DEVICES=4`?  That would avoid having to weaken the command-line argument checking.

mhoemmen:
@bartlettroscoe Is it possible to set an environment variable per test?  I think the problem is that people might want to have different tests, that set different Kokkos options per test.

bartlettroscoe:
CC: @kddevin 

@mhoemmen 

> Is it possible to set an environment variable per test?

Yes, see the `ENVIRONMENT <var0>=<value0> <var1>=<value1> ` argument in [`TRIBITS_ADD_TEST()`](https://tribits.org/doc/TribitsDevelopersGuide.html#formal-arguments-tribits-add-test).

But what is the harm in setting `KOKKOS_NUM_DEVICES=4` for all tests?  If a test does not use the GPU, then it gets ignored, no?

mhoemmen:
@bartlettroscoe wrote:
> But what is the harm in setting `KOKKOS_NUM_DEVICES=4` for all tests? If a test does not use the GPU, then it gets ignored, no?

What about nodes that have less than 4 GPUs?

kddevin:
Sure, I can set the environment variable for num devices, but other kokkos flags don't have documented environment variables.
https://github.com/kokkos/kokkos/wiki/Initialization

Using command line arguments is one way that Kokkos suggests running; do we want to exclude it from the testing framework?

Setting the environment variable in CMakeLists.txt doesn't help if I want to manually run the test as shown above.

mhoemmen:
@kddevin wrote:
> Sure, I can set the environment variable for num devices, but other kokkos flags don't have documented environment variables.

I just documented some of the flags on the Kokkos wiki yesterday (that section didn't exist before) ;-) .  I can document the rest of them if you like!

bartlettroscoe:
> Setting the environment variable in CMakeLists.txt doesn't help if I want to manually run the test as shown above.

@kddevin, can't you just do:


```
$ export KOKKOS_NUM_DEVICES=4
$ ctest
```

?

My concern with removing validation of commandline arguments is that it will make tests weaker by letting people think they are setting test arguments where they in fact are not.  This is like the old NOX PL that did not validate.

bartlettroscoe:
> What about nodes that have less than 4 GPUs?

@mhoemmen, when you run the Trilinos test suite, you always run on one node.

kddevin:
Yes, I can set the environment variable, but that isn't a general solution.  It seems wrong to support only one way of using Kokkos, especially without documenting that Teuchos testing supports only that one way. 

As @mhoemmen said, the environment variables weren't documented yesterday when I was looking for a solution.  The advertised way to use Kokkos was with command-line arguments.  (Thanks for the new documentation, @mhoemmen!   It is unfortunate that the environment variables have different names than the command-line arguments.)

If you don't want to approve the PR, close it.  I suspect someone will get bitten by this again in the future, though.

bartlettroscoe:
> If you don't want to approve the PR, close it. I suspect someone will get bitten by this again in the future, though.

@kddevin, it is just that I worry about weakening the testing system for everyone because of this one Kokkos issue.  For example, what if someone thinks they are adding a new argument `--num-elements=50` in the `CMakeLists.txt` file but it is really spelled `--num-ele`.  With the current implementation, the would fail in the PR testing build right away and they would be forced to fix it.   If we were to approve and merge this PR, that error would be silently (by PR testing standards) ignored.  I think that is a very bad thing.  I think we always want to bias correctness (check that things are correct) over robustness (just run whatever garbage is passed in).   This is computational science after all.  That is why, for example, years ago I put so much effort in Teuchos::ParameterList validation.  

I wonder if we should not open this up to more developers because this will impact all Trilinos packages and even all Trilinos uses that use the Teuchos Unit Test Harness?  This is not a change to take lightly (IMHO).

mhoemmen:
@bartlettroscoe wrote:
> when you run the Trilinos test suite, you always run on one node.

Right, but different nodes might have different numbers of GPUs.

mhoemmen:
@bartlettroscoe wrote:
> With the current implementation, the would fail in the PR testing build right away and they would be forced to fix it. If we were to approve and merge this PR, that error would be silently (by PR testing standards) ignored. I think that is a very bad thing.

I'm with you there.  I would prefer that we use the Kokkos environment variables, or even that we connect Teuchos with Kokkos in some way so that Teuchos knows about Kokkos' command-line options.  I'm not a fan of silently ignoring stuff.

kddevin:
I didn't intend for this to be a big design issue.  It isn't worth spending more time (or more people's time) on it.

mhoemmen:
@kddevin wrote:
>  It is unfortunate that the environment variables have different names than the command-line arguments

We can fix that, though!  Let's file Kokkos issues.  The nice thing about environment variables is that it's pretty easy to add new spellings and keep the old ones for a while.

PR #6593, opened 2020-01-16
Tpetra: Fix overflow for TpetraCore_MatrixMatrix_UnitTests

MicheldeMessieres:
TpetraCore_MatrixMatrix_UnitTests has a case which uses partially filled matrices (cdt_tce.mtx for example).

This causes C_estimate_nnz to overflow with A.getNodeNumEntries() less than B.getNodeNumRows().

Then Aest = 0 and Best = 0 overflow this line with result: 18446744073709551615

  size_t nnzperrow = (size_t)(sqrt((double)Aest) + sqrt((double)Best) - 1);

Then the following line overflows again to coincidentally bring the guess back to 1.

  nnzperrow *= nnzperrow;

The submitted changes will fix the overflow.

@trilinos/tpetra

## Motivation

Fixes an overflow which is probably not causing any defects.

## Testing

Overflow occurred for TpetraCore_MatrixMatrix_UnitTests.

### DIFF
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
@@ -1526,9 +1526,9 @@ template<class CrsMatrixType>
 size_t C_estimate_nnz(CrsMatrixType & A, CrsMatrixType &B){
   // Follows the NZ estimate in ML's ml_matmatmult.c
   size_t Aest = 100, Best=100;
-  if (A.getNodeNumEntries() > 0)
-    Aest = (A.getNodeNumRows() > 0)?  A.getNodeNumEntries()/A.getNodeNumRows() : 100;
-  if (B.getNodeNumEntries() > 0)
+  if (A.getNodeNumEntries() >= A.getNodeNumRows())
+    Aest = (A.getNodeNumRows() > 0) ? A.getNodeNumEntries()/A.getNodeNumRows() : 100;
+  if (B.getNodeNumEntries() >= B.getNodeNumRows())
     Best = (B.getNodeNumRows() > 0) ? B.getNodeNumEntries()/B.getNodeNumRows() : 100;
 
   size_t nnzperrow = (size_t)(sqrt((double)Aest) + sqrt((double)Best) - 1);


packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -1526,9 +1526,9 @@ template<class CrsMatrixType>
 size_t C_estimate_nnz(CrsMatrixType & A, CrsMatrixType &B){
   // Follows the NZ estimate in ML's ml_matmatmult.c
   size_t Aest = 100, Best=100;
-  if (A.getNodeNumEntries() > 0)
-    Aest = (A.getNodeNumRows() > 0)?  A.getNodeNumEntries()/A.getNodeNumRows() : 100;
-  if (B.getNodeNumEntries() > 0)
+  if (A.getNodeNumEntries() >= A.getNodeNumRows())
+    Aest = (A.getNodeNumRows() > 0) ? A.getNodeNumEntries()/A.getNodeNumRows() : 100;
```
mhoemmen: All these things return `size_t`, which is usually unsigned.  We should really fix that; I'm surprised that this trinary operator line doesn't issue a warning.

@csiefer2 and @brian-kelley , will this code still work, given that graphs and matrices are all StaticProfile now?

brian-kelley: I would expect a sign-compare warning on line 1530, but other than that it should still work.

From the definition of getNodeNumRows(), I don't think moving to StaticProfile changed the semantics of this. If the matrix is fill-active, it does a sum over k_numRowEntries_. Otherwise, it just returns the length of the allocated entries view.

MicheldeMessieres: @brian-kelley Can you clarify about the sign warning? I did not detect one and not sure why there should be one. Please let me know if you think that requires a change.

mhoemmen: @MicheldeMessieres
1. Line 1530 compares unsigned `size_t` to signed `int` (0 is an `int`).  That could cause a signed-to-unsigned comparison warning, though it's harmless in this case.
2. Line 1530 has a trinary operator with `size_t` for one result and `int` (100) for the second result.  It should just harmlessly convert the second result to `size_t`, but that's a signed to unsigned conversion and thus could cause a warning.

I don't think your changes should introduce any _new_ warnings, though.  @brian-kelley , would you agree?

brian-kelley: @mhoemmen That's what I was thinking, but I can build Tpetra with -Wall in GCC 9.1 and get no warnings. In this test program, only the second comparison produces a warning:
```
#include <iostream>
using namespace std;

int main()
{
  int i = 0;
  size_t j = 452U;
  if(j > 0)
    cout << "Bigger than 0.\n";
  if(j > i)
    cout << "Bigger than i.\n";
}
```
So I guess signed literals can be promoted to unsigned without a warning? I don't think the conditions for the warning are part of the standard, though.

brian-kelley: I don't get a warning from comparison with a literal in GCC 4.8, 7.4 or 9.1, IBM xl 2019 or Intel 18. So @MicheldeMessieres I would say don't worry about it.

mhoemmen: @brian-kelley I say it's fine for now.  That code needs more love anyway, but we should fix this bug first.

mhoemmen:
@MicheldeMessieres Thanks for the fix!!!

PR #6596, opened 2020-01-16
Zoltan2:  default test types

kddevin:
@trilinos/zoltan2 

## Motivation
Tpetra now builds with only global ID type at a time.
Thus, the complicated logic to determine the type of global ID to use in Zoltan2 tests is no longer needed.
These changes adopt the default types from Tpetra.
They also correct builds with default Tpetra types and no explicit instantiation.

## Testing

Linux with sems modules
```
module load sems-env
module load sems-cmake/3.10.3
module load sems-gcc/4.9.3
```

With gno=long long and gno=int
With explicit instantiation
non-ETI builds are experiencing an Xpetra error; these changes have no impact on that error.

### DIFF
--- a/packages/zoltan2/test/helpers/Zoltan2_TestHelpers.hpp
+++ b/packages/zoltan2/test/helpers/Zoltan2_TestHelpers.hpp
@@ -101,50 +101,23 @@ typedef Tpetra::Map<>::node_type znode_t;
 
 #include <TpetraCore_config.h>
 
-    typedef int zpart_t; // added this for consistency but needs further discussion
+typedef int zpart_t; // added this for consistency but needs further discussion
 
-#ifdef HAVE_TPETRA_EXPLICIT_INSTANTIATION
+typedef Tpetra::Map<>::local_ordinal_type zlno_t;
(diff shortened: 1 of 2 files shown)


packages/zoltan2/test/helpers/Zoltan2_TestHelpers.hpp
```diff
@@ -101,50 +101,23 @@ typedef Tpetra::Map<>::node_type znode_t;
...
-    typedef double zscalar_t;
-    typedef int zlno_t;
-    typedef int zgno_t;
-#   define HAVE_EPETRA_DATA_TYPES
-# endif  // TEST_STK_DATA_TYPES
-
-#endif // HAVE_TPETRA_EXPLICIT_INSTANTIATION
+typedef int zpart_t; // added this for consistency but needs further discussion
+
+typedef Tpetra::Map<>::local_ordinal_type zlno_t;
```
mhoemmen: Prefer the C++11 form `using zlno_t = Tpetra::Map<>::local_ordinal_type;` in new code.

Issue #6158, opened 2019-10-25
tpetra vector creation creates 6 modified flags causing a ton of launch overhead


bathmatt:
## Bug Report
@trilinos/tpetra 

### Description
I've figured out that this call here 

    RCP<vector_type> B_vec = B.getVectorNonConst (0);

creates 6 views of modified flags

```
|   |   |   |   |   |   |   |   |   |   |   |   |   compute: 5.93946 - 99.1017% [2148]
|   |   |   |   |   |   |   |   |   |   |   |   |   |   fused: 5.93083 - 99.8546% [2148]
|   |   |   |   |   |   |   |   |   |   |   |   |   |   |   Kokkos View Allocation - DualView::modified_flags: 0.00229072 - 0.0386239% [3866\

```

It should be 1.  Fixes in the non-deprecated branch make it 2, 

Currently a 0 length view takes 60us to create and ifpack2 makes 1000s of them per solve...  That is free money

There are 4500 calls in trilinos to this, that is a lot of places to reduce launch overhead of initializing these.  
### Steps to Reproduce
1. SHA1:  [insert here]
1. Configure script:  [attach here]
1. Configure log:  [attach here]
1. Build log:  [attach here]
1. Input deck:  [attach here]
1. Do this.
1. Do that.
1. Shake fist angrily at computer.
1. Run log:  [attach here]

kddevin:
The modified flags are in Kokkos::DualView.  There are three now, but when applications set   KOKKOS_ENABLE_DEPRECATED_CODE=OFF, there is one.  

getVectorNonConst(0) creates a new Tpetra::MultiVector containing the 0th vector of the "this" multivector (not a zero-length view).  New flags are created when subviews of the DualView are taken; the subview code is in Kokkos (Kokkos_DualView.hpp).  Subviews of both the host and device memory are also created in the DualView subview constructor.  Subviews of the import and export buffers are also created on host and device.

In Tpetra, perhaps we can add logic to test the number of vectors in a MultiVector, just reusing the views (new = old) instead of subviewing (new = subview(old, ...)) when the multivector has only one vector. @mhoemmen, what do you think?

 @cgcgcg

mhoemmen:
Hi @kddevin !  I'm looking at the implementations of `getVectorNonConst` and the DualView subview constructor, and I'm not seeing where the subviews create new flags.  (This matters because people don't always realize that subviews of a DualView share flags with the parent.) 

[...]

PR #6598, opened 2020-01-17
Tpetra:  use created subviews of communication buffers

kddevin:
@trilinos/tpetra 

## Motivation

When looking at #6158 , I noticed that the constructor of a vector from a multivector created subviews of the multivector's communication buffer, but never kept/used the subviews.  Communication routines allocate memory if needed, so the vectors worked fine.  But using the subviews should reduce the number of memory allocations.

I discussed these changes with @mhoemmen before implementing the changes.

## Testing

Tested on linux with Tpetra tests and downstream packages Belos, Ifpack2, MueLu, Zoltan2, and Teko.
```
cmake \
  -GNinja \
  -DMPI_EXEC=${MPI_EXEC} \
  -DTrilinos_ENABLE_TESTS=ON  \
  -DTrilinos_ENABLE_Tpetra=ON \
  -DTrilinos_ENABLE_MueLu=ON \
  -DTrilinos_ENABLE_Anasazi=ON \
  -DTrilinos_ENABLE_Belos=ON \
  -DTrilinos_ENABLE_Ifpack2=ON \
  -DTrilinos_ENABLE_Zoltan2=ON \
  -DTrilinos_ENABLE_Teko=ON \
.. |& tee OUTPUT.CMAKE
```
Results:
```
100% tests passed, 0 tests failed out of 558

Subproject Time Summary:
Belos      = 181.72 sec*proc (99 tests)
Ifpack2    =  60.77 sec*proc (44 tests)
MueLu      = 758.59 sec*proc (63 tests)
Teko       = 152.66 sec*proc (18 tests)
Tpetra     = 369.48 sec*proc (221 tests)
Zoltan2    = 238.28 sec*proc (113 tests)
```


packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3395,15 +3395,23 @@ namespace Tpetra {
     // be exactly what we need, so we won't have to resize them later.
     {
       const size_t newSize = X.imports_.extent (0) / numCols;
+      const size_t offset = j*newSize;
```
mhoemmen: `j` isn't necessarily the correct offset.  We should actually use `jj` (see above) here, in case `*this*` views a noncontiguous set of columns of another MultiVector.

Actually, if Y views a contiguous proper subset of columns of X, and Z views a noncontiguous proper subset of columns of Y, then `jj` is not even the right entry of `imports_`.  We can fix that in the "view contiguous subset of columns" constructors and methods by adjusting the offset of `imports_` and `exports_` accordingly.

kddevin: So I should switch j to jj?  Note that previously, the subviews took the first newSize entries of the buffer always.  If the buffers are strictly used for communication, that might be fine, but I thought indexing into them would be better.

kddevin:
An alternate approach would be to remove the subview construction and let the code operate as it does currently.  Would that be preferable to the changes in this PR?

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
mhoemmen: @kddevin wrote:
>  If the buffers are strictly used for communication, ....

That's correct.  They are used for nothing other than communication of packed data.

> So I should switch `j` to `jj`?

Yes.  I actually studied the example in my above comments, and I think this is a complete fix, given how `MultiVector::subView*` work. 

If we switch `j` to `jj`, that will fix the case where different single Vectors view distinct columns of one MultiVector.  In this case, your change will mean that different Vectors will view different parts of the buffer.  Thus, you'll be able to do concurrent asynchronous doExport or doImport correctly with the Vectors.

Consider the following, more complicated example:
```cpp
using MV = Tpetra::MultiVector<>;
MV X (map, 10);
Teuchos::Range1D Y_colRange (1, 5); // inclusive: 1, 2, 3, 4, 5
RCP<MV> Y = X.subViewNonConst (Y_colRange);

Teuchos::Array<size_t> whichVecsZ {{0, 2, 4}};
Teuchos::Array<size_t> whichVecsW {{0, 2, 4}};
RCP<MV> Z = X.subViewNonConst (whichVecsZ ());
RCP<MV> W = Y.subViewNonConst (whichVecsW ());
```
Y views columns 1, 2, 3, 4, and 5 of X.  Z views columns 0, 2, and 4 of X.  W views columns 0, 2, and 4 of Y -- thus, columns 1, 3, and 5 of X.  Users would expect it to be safe to do concurrent communication with Z and W.

In this case, they are right, because neither overload of `subViewNonConst` reuses their parent MultiVector's buffers.  Thus, Z and W each would start with zero-length `imports_` and `exports_` buffers, and would each reallocate them on demand.

Thus, this case is already correct, and your change doesn't break it.  If we decide to make `subView*` reuse communication buffers in the future, we only need to use the correct offset.  That will make the above example continue to be correct.

mhoemmen: The only way we could write a reliable test for this use case, though, would be if we had asynchronous doExport / doImport.  We don't have that right now.

Issue #6603, opened 2020-01-17
Tpetra::CrsMatrix::unpackAndCombine: Optimize GPU unpack for few long rows


mhoemmen:
## Bug Report
@trilinos/tpetra 

### Description

@vbrunini reports that `Tpetra::CrsMatrix::unpackAndCombine` is causing significant load imbalance for Aria's L2 Milestone problem, since shared bulk nodes mean that some MPI processes must unpack nearly dense rows at `doExport`.  The current `unpackAndCombine` implementation has one level of thread parallelism, over rows to unpack.  If there are a few rows with many entries, this is inefficient on GPUs.

### Suggested fix

- Introduce hierarchical parallelism for unpacking rows.
- Treat long rows in a separate pass from short rows when unpacking.  We have row length on both host and device, so we can preprocess on host and launch two separate kernels.  Task parallelism per thread block might be another option. 

### Steps to Reproduce

1. Start with Tpetra's [example / benchmark](https://github.com/trilinos/Trilinos/blob/master/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixDenseRowUnpack.cpp).

mhoemmen:
We could take either of two possible directions:

1. Sort the unpacked rows before merging them into the matrix; use sorted-sorted (both input and matrix row are sorted) merge.
2. Use hierarchical parallelism, over both rows and entries within a row.

I have been sketching out (2), but that might not help so much if it doesn't have a team sort.

brian-kelley:
@mhoemmen Team sort! #6646

mhoemmen:
Here is a sketch of some possible solutions.  At first, I was thinking of the case where the rows mostly have about the same number of entries, but that number of entries could be very short or very long.  For completeness, I'll describe a sketch of the solution to that problem first.  After that, I'll talk about the case where different rows have widely different numbers of entries.

When we refer to "long rows" in what follows, we mean the number of incoming (received) entries to unpack and combine with a local row of the target CrsMatrix, not the number of entries in the target local row.

[...]

PR #6604, opened 2020-01-17
Tpetra: Start work on #6603 (optimize CrsMatrix GPU unpack for long rows)

mhoemmen:
@trilinos/tpetra 

## Motivation

- `Tpetra::CrsMatrix::unpackAndCombine` is slow on GPUs for rows with many entries.
- We first need to define what a "row with many entries" is.
- Logical thing to do is to add a new function to `Tpetra::Details::Behavior`.
- Also, clean up some data structures in CrsGraph and CrsMatrix that are no longer needed.  This will help with the effort by reducing the number of cases to consider.

## Related Issues

* Part of #6603

## Stakeholder Feedback

Requested by @vbrunini based on Aria GPU L2 Milestone.

## Testing

Tested on Linux GCC 4.9.3.  Added new tests for existing and new Behavior methods.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
@@ -226,26 +226,32 @@ namespace Tpetra {
   {
     const char tfecfFuncName[] = "CrsMatrix(RCP<const Map>, RCP<const Map>, "
       "size_t, ProfileType[, RCP<ParameterList>]): ";
-
+    const char suffix[] =
+      "  Please report this bug to the Tpetra developers.";
 #ifdef HAVE_TPETRA_DEBUG
-    // An artifact of debugging something a while back.
-    TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC
-      (! staticGraph_.is_null (), std::logic_error,
-      "staticGraph_ is not null at the beginning of the constructor.  "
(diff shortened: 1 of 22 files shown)


mhoemmen:
@trilinos/tpetra @tjfulle Any takers for a review?  Thanks!

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -1390,108 +1392,112 @@ namespace Tpetra {
...
 
-#ifdef HAVE_TPETRA_DEBUG
+      if (debug) {
+        const char myPrefix[] =
+          "(StaticProfile \"Optimize Storage\"=false branch) ";
+        TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC
+          (k_ptrs_const.extent (0) == 0, std::logic_error, myPrefix
+           << "k_ptrs_const.extent(0) = 0.  This probably means "
+           "that k_rowPtrs_ was never allocated.");
+        if (k_ptrs_const.extent (0) != 0) {
```
brian-kelley: This condition is always true after throwing if it's false.

mhoemmen: Thanks @brian-kelley ! :-D  I have some other fixes in the pipe, so I'll add this.

PR #6615, opened 2020-01-20
spadd bugfix, cleanup for performance

brian-kelley:
- Making sure last entry of rowcount view is correct, if Crowmap on input
wasn't initialized (a bug affecting Tpetra!)
- Remove unnecessary view initializations (@csiefer2 measured this was taking up to 50% of the
    spadd time on vortex!)
- Remove unnecessary fences
- Still have fences at the end of user-facing functions,
  so that timers on host accurately reflect kernel times
  - Use idiomatic deep_copy to scalar for accessing views, rather than
  making a host mirror of a single-element subview
- Negligible impact, but avoid View reference count increments (pass-by-ref to functor ctor)

There will be a PR in kokkos-kernels develop as well, but these changes couldn't wait until the next snapshot after 3.0.

@trilinos/kokkos-kernels 

## Stakeholder Feedback

Code owner

## Testing

This code is tested by MatrixMatrix_UnitTests in Tpetra, and the sparse unit tests in KokkosKernels.

### DIFF
--- a/packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
+++ b/packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
@@ -113,12 +113,12 @@ namespace Experimental {
         Crowcounts(nrows) = 0;
       }
     }
+    ordinal_type nrows;
     const typename ARowPtrsT::const_type Arowptrs;
     const AColIndsT Acolinds;
     const typename BRowPtrsT::const_type Browptrs;
     const BColIndsT Bcolinds;
     CRowPtrsT Crowcounts;
-    ordinal_type nrows;
   };
 
   //get upper bound for C entries per row (assumes worst case, that entries in A and B on each row are disjoint)
@@ -382,6 +382,7 @@ namespace Experimental {
       if(i == nrows - 1)
         Crowcounts(nrows) = 0;
     }
+    ordinal_type nrows;
     const ArowptrsT Arowptrs;
     const BrowptrsT Browptrs;
     const CrowptrsT Crowptrs;
@@ -390,7 +391,6 @@ namespace Experimental {
     const CcolindsT ABperm;
     CcolindsT Apos;
     CcolindsT Bpos;
-    ordinal_type nrows;
   };
 
   //Symbolic: count entries in each row in C to produce rowmap
--- a/packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
+++ b/packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
@@ -69,15 +69,16 @@ namespace Experimental {
   struct SortedCountEntries
   {
     SortedCountEntries(
+        ordinal_type nrows_,
         const typename ARowPtrsT::const_type& Arowptrs_,
         const AColIndsT& Acolinds_,
         const typename BRowPtrsT::const_type& Browptrs_,
         const BColIndsT& Bcolinds_,
-        CRowPtrsT& Crowcounts_) :
+        const CRowPtrsT& Crowcounts_) :
+      nrows(nrows_),
       Arowptrs(Arowptrs_), Acolinds(Acolinds_),
       Browptrs(Browptrs_), Bcolinds(Bcolinds_),
-      Crowcounts(Crowcounts_),
-      nrows(Arowptrs.extent(0) - 1)
+      Crowcounts(Crowcounts_)
     {}
     KOKKOS_INLINE_FUNCTION void operator()(const ordinal_type i) const
     {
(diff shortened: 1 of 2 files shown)


ndellingwood:
@brian-kelley add me to the list for the kokkos-kernels PR so I can cherry-pick for the 3.0 release.

packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
```diff
@@ -69,15 +69,17 @@ namespace Experimental {
...
-        const typename ARowPtrsT::const_type Arowptrs_,
-        const AColIndsT Acolinds_,
-        const typename BRowPtrsT::const_type Browptrs_,
-        const BColIndsT Bcolinds_,
-        CRowPtrsT Crowcounts_) :
+        const typename ARowPtrsT::const_type& Arowptrs_,
+        const AColIndsT& Acolinds_,
+        const typename BRowPtrsT::const_type& Browptrs_,
+        const BColIndsT& Bcolinds_,
+        CRowPtrsT& Crowcounts_) :
```
mhoemmen: Why is this one a nonconst reference?  If these are all `Kokkos::View` objects and if you're not reallocating them, then you can pass them either by value or by const reference.

packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
```diff
@@ -69,15 +69,17 @@ namespace Experimental {
...
       Arowptrs(Arowptrs_), Acolinds(Acolinds_),
       Browptrs(Browptrs_), Bcolinds(Bcolinds_),
-      Crowcounts(Crowcounts_) {}
-    KOKKOS_INLINE_FUNCTION void operator()(const size_type i) const
+      Crowcounts(Crowcounts_),
+      nrows(Arowptrs.extent(0) - 1)
```
mhoemmen: `Arowptrs.extent(0)` is zero if the matrix has zero rows.

ndellingwood:
@brian-kelley once the test failure is worked out but before merging this, please put in a PR to kokkos-kernels with a testing spot-check (use a different testbed for host coverage than bowman for faster testing), so we avoid running into issues not caught by gaps in coverage between the two different collections of tests.

packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
```diff
@@ -182,125 +203,126 @@ namespace Experimental {
...
+    SortEntriesFunctor(const CrowptrsT& Crowptrs_, const CcolindsT& Ccolinds_, const CcolindsT& ABperm_) :
+      Crowptrs(Crowptrs_),
+      Ccolinds(Ccolinds_),
+      CcolindsAux("C colind aux", Ccolinds_.extent(0)),
+      ABperm(ABperm_),
+      ABpermAux("AB perm aux", ABperm_.extent(0))
+    {}
+
+    KOKKOS_INLINE_FUNCTION static void
+    radixSortKeysAndValues(ordinal_type* keys, ordinal_type* keysAux, ordinal_type* values, ordinal_type* valuesAux, size_type n)
```
mhoemmen: @brian-kelley @csiefer2 It looks like it wouldn't be too hard to convert this into a team-parallel sort.  I might do that at some point.  Do you see any obstacles to that?

packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
```diff
@@ -427,87 +433,72 @@ namespace Experimental {
...
     //easy for sorted, but for unsorted is easiest to just compute the whole sum
     auto addHandle = handle->get_spadd_handle();
-    auto nrows = a_rowmap.extent(0) - 1;
+    if(a_rowmap.extent(0) == 0 || a_rowmap.extent(0) == 1)
+    {
+      //Have 0 rows, so nothing to do except set #nnz to 0
+      addHandle->set_max_result_nnz(0);
+      //If c_rowmap has a single entry, it must be 0
+      if(c_rowmap.extent(0))
+        Kokkos::deep_copy(c_rowmap, (size_type) 0);
```
mhoemmen: Prefer C++ - style casts, and avoid implicit conversions of integers to `bool`:
```cpp
if (c_rowmap.extent(0) != 0) {
  Kokkos::deep_copy(c_rowmap, size_type(0));
}
```

packages/kokkos-kernels/src/sparse/KokkosSparse_spadd.hpp
brian-kelley: @mhoemmen It will be after the KokkosKernels snapshot. On GPU it will be parallel (team) bitonic sort within rows, and on CPU it's still serial radix sort within rows. This would also (probably) give a big speedup to Tpetra::sortCrsEntries().

brian-kelley: @mhoemmen If you meant parallelizing within the radix sort, I have thought about it and I don't think it's possible/easy without shared memory atomics, which are apparently no faster than global memory atomics.

mhoemmen: @brian-kelley Right now I have no team sort, so as long as a team sort is faster than a sequential sort, I would take it :-) .

PR #6617, opened 2020-01-21
Tpetra: Resolve CUDA_LAUNCH_BLOCKING requirements for serial

MicheldeMessieres:
These changes allow all Tpetra tests in serial to pass with
env variable CUDA_LAUNCH_BLOCKING=0, except for these utility
tests which specificaly check CUDA_LAUNCH_BLOCKING to be on.
They have not been changed and will still fail as expected:
- 	 TpetraCore_Core_initialize_where_tpetra_initializes_kokkos
- 	 TpetraCore_Core_ScopeGuard_where_tpetra_initializes_kokkos
- 	 TpetraCore_Core_initialize_where_user_initializes_kokkos
- 	 TpetraCore_Core_ScopeGuard_where_user_initializes_kokkos
- 	 TpetraCore_Map_no_warn_on_pre_finalize_destruction_issue_2372

One exception is TpetraCore_Transform which has an issue caused by the setup for  withLocalAccess. A planned fix is expected to fix the test to also work for CUDA_LAUNCH_BLOCKING=0 but that will be handled separately.

Parallel tests which fail with CUDA_LAUNCH_BLOCKING=0 have not been addressed yet.

@trilinos/tpetra

## Motivation

Resolve Tpetra to not required CUDA_LAUNCH_BLOCKING

## Testing

These were identified running Tpetra tests on white's Pascal node.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
+++ b/packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
@@ -325,8 +325,15 @@ namespace Tpetra {
              range_type range (execSpace, 0, lclNumRows);
 
              // MDM-TODO
-             // Need to resolve this fencine for the transform tests.
-             // Also may need host or device fence depending on user choice.
+             // Need to resolve this fencing for the transform tests.
+             // If data is on device and the transform calls for execSpace host
+             // we'll need a fence. However the transform to host call currently
+             // leaves need_sync_host() true which i need to discuss and verify
+             // if it's an error. Perhaps that should be resolved first.
+             // Also may need host or device fence. This default fence is on device
+             // which happens to work for the current test setup. I think we'll
+             // want to make a flipped version of the test to force the error here.
+             // Note PR 6617 discusses these current issues.
              Kokkos::fence();
 
              Kokkos::parallel_for (kernelLabel, range, g);
@@ -364,9 +371,7 @@ namespace Tpetra {
             using range_type = Kokkos::RangePolicy<ExecutionSpace, LO>;
             range_type range (execSpace, 0, lclNumRows);
 
-            // MDM-TODO
-            // Need to resolve this fencine for the transform tests.
-            // Also may need host or device fence depending on user choice.
+            // MDM-TODO - see similar note above for transform_vec_notSameObject
             Kokkos::fence();
 
             Kokkos::parallel_for (kernelLabel, range, g);
--- a/packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
+++ b/packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
@@ -324,17 +324,14 @@ namespace Tpetra {
              using range_type = Kokkos::RangePolicy<ExecutionSpace, LO>;
              range_type range (execSpace, 0, lclNumRows);
 
-             // MDM-TODO
-             // Need to resolve this fencing for the transform tests.
-             // If data is on device and the transform calls for execSpace host
-             // we'll need a fence. However the transform to host call currently
-             // leaves need_sync_host() true which i need to discuss and verify
-             // if it's an error. Perhaps that should be resolved first.
-             // Also may need host or device fence. This default fence is on device
-             // which happens to work for the current test setup. I think we'll
-             // want to make a flipped version of the test to force the error here.
-             // Note PR 6617 discusses these current issues.
-             Kokkos::fence();
+             // Note that the Transform.cpp test will not currently pass
+             // with CUDA_LAUNCH_BLOCKING=0. This is because execSpace is not
+             // considered by withLocalAccess so after requesting the transform
+             // with Host we don't get a sync and need_sync_host() is still true.
+             // That issue is regardless of whether CUDA_LAUNCH_BLOCKING is set.
+             // When withLocalAccess is fixed we should then have the proper fencing
+             // and the test will run correctly with CUDA_LAUNCH_BLOCKING=0 or =1.
+             // PR 6617 discusses these current issues.
 
              Kokkos::parallel_for (kernelLabel, range, g);
            },
@@ -371,8 +368,8 @@ namespace Tpetra {
             using range_type = Kokkos::RangePolicy<ExecutionSpace, LO>;
             range_type range (execSpace, 0, lclNumRows);
 
-            // MDM-TODO - see similar note above for transform_vec_notSameObject
-            Kokkos::fence();
+            // See note above for explanation why a pending fix to withLocalAccess
+            // will fix Transform.cpp test to pass with CUDA_LAUNCH_BLOCKING=0.
 
             Kokkos::parallel_for (kernelLabel, range, g);
           },
--- a/packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
+++ b/packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
@@ -323,6 +323,12 @@ namespace Tpetra {
              const LO lclNumRows = static_cast<LO> (input_lcl.extent (0));
              using range_type = Kokkos::RangePolicy<ExecutionSpace, LO>;
              range_type range (execSpace, 0, lclNumRows);
+
+             // MDM-TODO
+             // Need to resolve this fencine for the transform tests.
+             // Also may need host or device fence depending on user choice.
+             Kokkos::fence();
+
              Kokkos::parallel_for (kernelLabel, range, g);
            },
            readOnly (input).on (memSpace),
@@ -357,6 +363,12 @@ namespace Tpetra {
             const LO lclNumRows = static_cast<LO> (input_lcl.extent (0));
             using range_type = Kokkos::RangePolicy<ExecutionSpace, LO>;
             range_type range (execSpace, 0, lclNumRows);
+
+            // MDM-TODO
+            // Need to resolve this fencine for the transform tests.
+            // Also may need host or device fence depending on user choice.
+            Kokkos::fence();
+
             Kokkos::parallel_for (kernelLabel, range, g);
           },
           readOnly (input).on (memSpace),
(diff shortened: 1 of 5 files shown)


MicheldeMessieres:
@crtrott @mhoemmen @kddevin I just looked at serial tests so far for Tpetra. This PR will address those. I'll resolve parallel tests as a second pass.

packages/tpetra/core/test/MultiVector/Transform.cpp
```diff
@@ -138,6 +139,8 @@ namespace { // (anonymous)
     }
 
     Y.putScalar (418.0);
+    Kokkos::fence();
```
mhoemmen: It's interesting that we needed a fence here.  The last `transform` call should have marked the MultiVector as modified on device.  This means that `putScalar` should run on device. The `transform` that follows should notice that `Y` has been modified on device, and call `sync_host`.  It used to be the case that `Kokkos::DualView::sync_host` called `Kokkos::fence` even if the two memory spaces were the same.  Now it does nothing.  Surprise! ;-)

I would say, rather than putting a fence here, try changing the implementation of `Tpetra::MultiVector::sync_host` to look like this:
```cpp
if (need_sync_host ()) {
  execution_space ().fence ();
  view_.sync_host ();
}
```
That should fix this issue.  If it doesn't, then there's something wrong.

packages/tpetra/core/test/MultiVector/Transform.cpp
```diff
@@ -270,6 +273,7 @@ namespace { // (anonymous)
     constexpr double flagValue = -1.0;
     X.putScalar (flagValue);
     Y.putScalar (666.0);
+    Kokkos::fence();
```
mhoemmen: See above; we should be able to remove this fence.

packages/tpetra/core/test/MultiVector/Transform.cpp
```diff
@@ -298,6 +302,8 @@ namespace { // (anonymous)
     }
 
     Y.putScalar (418.0);
+    Kokkos::fence();
```
mhoemmen: See above; we should be able to remove this fence.

packages/tpetra/core/test/MultiVector/Transform.cpp
```diff
@@ -107,6 +107,7 @@ namespace { // (anonymous)
     constexpr double flagValue = -1.0;
     X.putScalar (flagValue);
     Y.putScalar (666.0);
+    Kokkos::fence();
```
mhoemmen: See below; we should be able to remove this fence.

packages/tpetra/core/test/MultiVector/Transform.cpp
MicheldeMessieres: @mhoemmen Thanks for the feedback. I verified putScalar was acting on device. Tried your recommended change but then it seems transform never actually calls sync_host or sync_device. You are expecting transform to do this, so perhaps we want to call it at the beginning of the transform for the requested exec space?

In this case transform calls transform_mv_notSameObject and I observe we'll need a fence before the parallel_for, shown below. But I expect we resolve this earlier, probably at the beginning of the transform.

```
        withLocalAccess
          ([=] (const input_view_type& input_lcl,
            using functor_type = MultiVectorUnaryTransformLoopBody<
            input_view_type, output_view_type, UnaryFunctionType, LO>;
            functor_type g (input_lcl, output_lcl, f);

            const LO lclNumRows = static_cast<LO> (input_lcl.extent (0));
            using range_type = Kokkos::RangePolicy<ExecutionSpace, LO>;
            range_type range (execSpace, 0, lclNumRows);
// we need the fence before this parallel_for
            Kokkos::parallel_for (kernelLabel, range, g);
          },
          readOnly (input).on (memSpace),
          writeOnly (output).on (memSpace));
```
Also it seems I made a mistake on two other fences - I could not reproduce their necessity. Only the putScalar calls before the transform for host execution space are necessary. The transform on device does not a fence since the prior putScalar was also on device.

packages/tpetra/core/test/MultiVector/Transform.cpp
MicheldeMessieres: This fence was apparently a mistake.

packages/tpetra/core/test/MultiVector/Transform.cpp
MicheldeMessieres: This fence was apparently a mistake.

packages/tpetra/core/test/MultiVector/Transform.cpp
mhoemmen: `withLocalAccess` should call the templated form of `Tpetra::MultiVector::sync`.  See lines 118-151 of [`Tpetra_withLocalAccess_MultiVector.hpp`](https://github.com/trilinos/Trilinos/blob/5bd5b68a3284b1bfacdafebcfd23484057fdf486/packages/tpetra/core/src/Tpetra_withLocalAccess_MultiVector.hpp#L143).  If it's not, then that might indicate that `Tpetra::MultiVector::need_sync` is not doing the right thing.

mhoemmen: As a temporary test, could you please try the following change to that file?  Replace lines 142-144 with the following:
```cpp
if (LA.G_.template need_sync<space>()) {
  if (is_host) {
    using dev_exec_space = typename dual_view_type::t_dev::execution_space;
    dev_exec_space().fence();
  }
  LA.G_.template sync<space>();
}
```
Alternately, we could just change `Tpetra::MultiVector::sync` always to do a fence, as above.

MicheldeMessieres: You had asked me to originally modify need_sync_host and need_sync_device, but I'm realizing now we also have need_sync which is called in this case. Do we want to modify the need_sync the same way, with a fence on the opposite of host or device depending on TargetDeviceType. 

```
    //! Whether this MultiVector needs synchronization to the given space.
    template<class TargetDeviceType>
    bool need_sync () const {
// I think we want to fence here similar to your original suggestions
// fence host if TargetDeviceType is device
// fence device if TargetDeviceType is host
      return view_.template need_sync<TargetDeviceType> ();
    }

    //! Whether this MultiVector needs synchronization to the host.
    bool need_sync_host () const;

    //! Whether this MultiVector needs synchronization to the device.
    bool need_sync_device () const;
```

mhoemmen: @MicheldeMessieres Wait, I didn't ask you to modify `need_sync*`.  That doesn't make sense -- the point of `need_sync*` is to check the flags, not to do work.

Here's what should actually happen:

1. `sync_host` should only fence if a fence is needed, i.e., if `need_sync_host()` is true.
2. `sync<Space>` should only fence if a fence is needed (i.e., if `need_sync<Space>()` is true), and only if sync'ing to host.

I'm all for getting this stuff over with, so whatever, just put the fences wherever, but please just put a big comment over them saying something like "THIS IS WEIRD WE SHOULDN'T NEED THIS SEE `#6617`" and I'll go back to it later.  I don't want to waste too much time when it could be some issue with DualView or even with `Kokkos::fence()` being a lie.

mhoemmen:
@MicheldeMessieres OK, here are two options:

1. Investigate further.  This could be an issue in the MultiVector specialization of withLocalAccess, or with `Kokkos::DualView::sync*`. 
2. Just put the fences in there and move on.  Put a big angry TODO comment above them pointing to this issue.

@kddevin is in charge

packages/tpetra/core/test/MultiVector/Transform.cpp
MicheldeMessieres: Sorry - I meant you asked me to modify sync_host and sync_device and then noticed we also had the templated sync in the decl. Then did my own brain transform and edited the wrong thing.

I updated the PR to remove fences from Transform.cpp and added fences with a TODO comment in the transform calls. This works but I need to investigate further.

The user choice of host (execSpace) is not passed into withLocalAccess so I don't think it can know about the choice and act on it. The problem is when host is requested in the transform call and data is on device (or vice versa I expect). I will educate myself on exactly what these range policies are doing and learn more about this setup. Then the proper design may be more apparent.

MicheldeMessieres:
In MultiVector/Transform.cpp line 143 has a transform to host:

`transform ("418 -> 419", Kokkos::DefaultHostExecutionSpace (), Y, X, KOKKOS_LAMBDA (const double& X_ij) { return X_ij + 1.0; });`

Afterwards, X.need_sync_host() is true - but I think perhaps it should be false. The test also has a comment that says X.sync_host() is not needed.

If X.need_sync_host() should be false, then I should resolve that first, unrelated to CUDA_LAUNCH_BLOCKING.

If there is a reason X.need_sync_host() should/could still be true even though the transform was to host, I will appreciate clarification how it is expected to work.

mhoemmen:
@MicheldeMessieres @kddevin I'm working on fixing `withLocalAccess` in a more permanent way.  I'll thus be happy to approve these changes and keep working on a better `withLocalAccess` fix.

MicheldeMessieres:
@mhoemmen @kddevin One further edit, I removed the fence in the transform calls and just left a note as per our conversation. My expectation is that the withLocalAccess modification will fix the sync state and then CUDA_LAUNCH_BLOCKING=0 will just work because the fence is already getting called. I updated the PR description as well and removed WIP. Will merge if it's ok.

packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
```diff
@@ -324,17 +324,14 @@ namespace Tpetra {
...
-             // If data is on device and the transform calls for execSpace host
-             // we'll need a fence. However the transform to host call currently
-             // leaves need_sync_host() true which i need to discuss and verify
-             // if it's an error. Perhaps that should be resolved first.
-             // Also may need host or device fence. This default fence is on device
-             // which happens to work for the current test setup. I think we'll
-             // want to make a flipped version of the test to force the error here.
-             // Note PR 6617 discusses these current issues.
-             Kokkos::fence();
+             // Note that the Transform.cpp test will not currently pass
```
mhoemmen: I was actually fine with the temporary fix, but this is OK too : - )

PR #6621, opened 2020-01-21
Fix const_cast in Tpetra, Amesos and Ifpack2

matt-frey:
@trilinos/<teamName>

## Motivation

Compiling Trilinos (i.e. explicitly Tpetra, Amesos2 and Ifpack2 ) for GPUs fails due to
some const_cast.

The problem is fixed by using & in const_cast with *this pointer.


matt-frey:
@mhoemmen Below an example error message in case of ```Tpetra::MultiVector```:
```bash
Trilinos/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp:1920:1: 
error: passing 'const Tpetra::MultiVector<double, int,
long long int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >'
as 'this' argument discards qualifiers [-fpermissive]
       const_cast<MV*>(this)->sync_device ();
```

mhoemmen:
@matt-frey Huh, that's weird.  `this` is a prvalue and it should be OK to `const_cast` it.  Anyway, thanks for reporting and fixing! :-D

PR #6625, opened 2020-01-21
Tpetra: MMM Modifications to avoid remotemap construction in serial

csiefer2:
Tests: Vortex, RHEL7

### DIFF
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
@@ -2889,9 +2889,11 @@ void import_and_extract_views(
   Aview.colMap       = A.getColMap();
   Aview.domainMap    = A.getDomainMap();
   Aview.importColMap = null;
+  RCP<const map_type> rowMap = A.getRowMap();
+  const int numProcs = rowMap->getComm()->getSize();
 
-  // Short circuit if the user swears there are no remotes
-  if (userAssertsThereAreNoRemotes)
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -2906,13 +2908,16 @@ void import_and_extract_views(
 
     // Mark each row in targetMap as local or remote, and go ahead and get a view
     // for the local rows
-    RCP<const map_type> rowMap = A.getRowMap(), remoteRowMap;
+    RCP<const map_type> remoteRowMap;
```
jhux2: where did `rowMap` go?

csiefer2: Up above...

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -2926,6 +2931,9 @@ void import_and_extract_views(
 
     } else if (prototypeImporter.is_null()) {
       // No prototype importer --- count the remotes the hard way
+#ifdef HAVE_TPETRA_MMM_TIMINGS
+      TimeMonitor MM2 = *TimeMonitor::getNewTimer(prefix_mmm + std::string("MMM I&X RemoteMap-Mode2"));
```
mhoemmen: I realized this compiles only because `TimeMonitor(Timer&)` is not marked `explicit`.  At least TimeMonitor isn't being handled by RCP here ;-) .

PR #6627, opened 2020-01-21
Kokkos: fix threadfence issue in scan

crtrott:
@Trilinos/tpetra @trilinos/kokkos 

This is a hot fix fixing a very rare race condition in the parallel_scan algorithm.


mhoemmen:
In theory, I'm not supposed to approve PRs that don't talk about testing.  However, I know that Christian tested this in the EMPIRE application, where the bug originally came up.

ibaned:
very glad you found this! always good for us to remember that even the longest-lived and most trusted piece of central code we write can still have bugs.

kddevin:
Victory!  Congratulations and thank you, @crtrott.

PR #6628, opened 2020-01-21
Tpetra: Correcting unecessary extraction of remotes for A matrix in Jacobi

csiefer2:
Tests: RHEL6, Cuda(vortex)


csiefer2:
This superceeds the previous PR.  This also makes Kokkos-ifying the remotes less important.

PR #6630, opened 2020-01-22
Added os.path.abspath to find the full path to the tested file

prwolfe:
Basically if I don't then running from
inside the unittest directory will fail.


mhoemmen:
@prwolfe Sorry to make more work for you, but Trilinos leadership team now insists that PRs not get approved unless they describe how the changes were tested.

prwolfe:
@mhoemmen It's ok - I want @william76 to see this more than anything.  I tested it by running the python test I am changing from every directory from the one it's in up to the Trilinos root.  Perhaps I should check it in a parallel directory, but the change is basically only fixing the fact that there is no path when I run from it's containing directory, so even a parallel directory would have something.

PR #6631, opened 2020-01-22
Fix cluster MTSGS crashing with remote neighbors/columnss

brian-kelley:
@trilinos/kokkos-kernels 

## Motivation

Make cluster MTSGS ignore remote columns  - fixes crash  

## Stakeholder Feedback

- Requests from @cgcgcg , @iyamazaki and @lucbv for using this with Exawind
- Code owner
## Testing

Tested in Ifpack2_UnitTestRelaxation

### DIFF
--- a/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
+++ b/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
@@ -529,7 +529,7 @@ namespace KokkosSparse{
       struct BuildCrossClusterMaskFunctor
       {
         BuildCrossClusterMaskFunctor(Rowmap& rowmap_, Colinds& colinds_, nnz_view_t& clusterOffsets_, nnz_view_t& clusterVerts_, nnz_view_t& vertClusters_, bitset_t& mask_)
-          : rowmap(rowmap_), colinds(colinds_), clusterOffsets(clusterOffsets_), clusterVerts(clusterVerts_), vertClusters(vertClusters_), mask(mask_)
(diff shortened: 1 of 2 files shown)


ndellingwood:
@brian-kelley please put in PR to kokkos-kernels and spot-check (skip bowman testing and xl on white since slow) before merging this so this is not clobbered by the promotion, and this will fill gaps in Trilinos PR testing coverage.

packages/kokkos-kernels/src/sparse/impl/KokkosSparse_cluster_gauss_seidel_impl.hpp
```diff
@@ -529,7 +529,7 @@ namespace KokkosSparse{
       struct BuildCrossClusterMaskFunctor
       {
         BuildCrossClusterMaskFunctor(Rowmap& rowmap_, Colinds& colinds_, nnz_view_t& clusterOffsets_, nnz_view_t& clusterVerts_, nnz_view_t& vertClusters_, bitset_t& mask_)
-          : rowmap(rowmap_), colinds(colinds_), clusterOffsets(clusterOffsets_), clusterVerts(clusterVerts_), vertClusters(vertClusters_), mask(mask_)
+          : numRows(rowmap_.extent(0) - 1), rowmap(rowmap_), colinds(colinds_), clusterOffsets(clusterOffsets_), clusterVerts(clusterVerts_), vertClusters(vertClusters_), mask(mask_)
```
mhoemmen: If the matrix has zero rows, `numRows` could be negative.  That doesn't break the code below, though.

brian-kelley: @mhoemmen Yes, every functor is a loop over rows (which is passed in, not determined from rowmap length) so they should be fine. In KokkosKernels I added a test that runs MTSGS and cluster MTSGS on zero-row matrices, with both 0-length and 1-length rowmaps. That ran clean in valgrind.

brian-kelley:
@lucbv I didn't even build Trilinos locally, I was relying on the PR tester :-1:

Issue #6632, opened 2020-01-22
Sacado:  Change implementation of derivatives of hyperbolic functions to avoid FPE


mperego:
## Enhancement
@trilinos/sacado 

Change the algorithm to compute the derivative of tanh and coth functions.

Current algorithm:
tanh(x)' = 1/cosh^2(x)
coth(x)' = -1/sinh^2(x)

Issue: for large x, sinh(x) and coth(x) can be Inf, triggering FPE errors.

Proposed algorithm:
tanh(x)' = 1-tanh^2(x)
coth(x)' = 1-coth^2(x)

Thanks.

mhoemmen:
@mperego I edited your post to get the group tag right.

etphipp:
Done in PR #6645.  Note Sacado currently doesn't implement coth(x), so I just did this for tanh(x).

etphipp:
PR #6645 has been merged, so closing.

PR #6645, opened 2020-01-24
Sacado:  Use more overflow-safe implementation of d(tanh(x))/dx.

etphipp:
See issue #6632.  Replace d(tanh(x))/dx = sech^2(x) with 1-tanh^2(x).


mhoemmen:
I'm under orders to find out how changes were tested before approving a PR, even if the changes are obviously a good thing :-) .  Is there a test that was failing before but now isn't?

etphipp:
> I'm under orders to find out how changes were tested before approving a PR, even if the changes are obviously a good thing :-) . Is there a test that was failing before but now isn't?

There wasn't a failing Sacado test, but there was a failing Albany test, as referenced in issue #6632.  I verified the Sacado tests still pass after this change.

prwolfe:
re-testing due to a full disk

PR #6647, opened 2020-01-24
Sacado:  Remove DMFad (both old and new design).

etphipp:
This was a poorly executed, highly experimental implementation of a
memory pool for Fad that no one should be using.  Furthermore it
generates a very strange warning with GCC 8.3 (see PR #6295) that I
cannot reproduce or see how to fix.  So there is no reason to keep it
and this commit removes it from Sacado.  Technically this is a break in
backwards compatibility, but I can think of no good reason for anyone to
use this.


prwolfe:
retesting due to a full disk

Issue #5472, opened 2019-07-03
ShyLU_Node: Intermittent test failure on master in GCC 4.8.4


prwolfe:
## Bug Report
@trilinos/shylu 
@trilinos/kokkos 

### Description
The test ShyLU_NodeTacho_Tacho_TestOpenMPDoubleTaskSchedulerMultiple_MPI_1 has been failing itermmitently since the 28th when run on gc 4.8.4 builds. We have had develop -> master promotions during this time so it's already there. the error is

> Kokkos contract violation:
>    Expected precondition `not m_wait_queue.is_consumed()` evaluated false.

So this may relate to the Kokkos team.

### Steps to Reproduce
Build and run the test using gcc 4.8.4 and the pull request instructions at https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors. Repeat until you get a failure (about 1 out of 3 so far.)

srajama1:
@kyungjoo-kim I am not able to find David Hollman in github mention, so adding Kyungjoo.

kyungjoo-kim:
@prwolfe Could you show me the complete error message output ? It uses gtest and we need to know which test triggers the failure. Does this failure only happen with a gcc compiler ? 

@dhollman I think that we saw this error before somehow and this should not be happened. I thought that we fixed the problem but we did not check openmp gcc version.  As kokkos promotion PR passes the PR test, this might be related to a runtime race condition. Do you have any idea of this failure ?

prwolfe:
@kyungjoo-kim - see https://testing.sandia.gov/cdash/testDetails.php?test=82838381&build=5260496 for the full output. If you don't have access it's

> Test output
> Kokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
>   In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
>   For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
>   For unit testing set OMP_PROC_BIND=false
>      DeviceSpace::  Kokkos::OpenMP thread_pool_topology[ 1 x 2 x 1 ]
>        HostSpace::  Kokkos::OpenMP thread_pool_topology[ 1 x 2 x 1 ]
> [==========] Running 58 tests from 7 test cases.
> [----------] Global test environment set-up.
> [----------] 2 tests from CrsMatrixBase
> [ RUN      ] CrsMatrixBase.constructor
[...]

Issue #3263, opened 2018-08-09
Test ShyLU_NodeTacho_Tacho_TestSerial_double_MPI_1 randomly failing in CI and PR GCC 4.8.4 + OpenMP builds


bartlettroscoe:
@trilinos/shylu, @trilinos/framework, @srajama1 (Trilinos Linear Solvers Product Lead)

## Expectations

A test should not fail unless a changes is made to break it.  A test should not randomly fail.

## Current Behavior

Looking at the four most recent failures of the test `ShyLU_NodeTacho_Tacho_TestSerial_double_MPI_1` in [this query](https://testing-vm.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=4&showfilters=1&filtercombine=and&field1=testname&compare1=61&value1=ShyLU_NodeTacho_Tacho_TestSerial_double_MPI_1&field2=status&compare2=62&value2=passed&field3=status&compare3=62&value3=notrun&field4=buildstarttime&compare4=84&value4=now) the test appears to be randomly failing in the GCC 4.8.4 OpenMPI builds.  In the most recent case, this test broke the auto PR GCC 4.8.4 + OpenMP build in PR #3260.  IN each of the last for failures of this test dating back to 6/28/2018, they all fail showing:

```
....
[ RUN      ] CrsMatrixBase.matrixmarket
unknown file: Failure
C++ exception with description "View bounds error of view ap ( 13 < 13 )
Traceback functionality not available
" thrown in the test body.
[  FAILED  ] CrsMatrixBase.matrixmarket (23 ms)
...
[  FAILED  ] 1 test, listed below:
[  FAILED  ] CrsMatrixBase.matrixmarket

 1 FAILED TEST
```

## Definition of Done

The test `ShyLU_NodeTacho_Tacho_TestSerial_double_MPI_1` is fixed to make it so that it does not randomly fail or is removed for CI and auto PR testing.

## Possible Solution

Fix it so that it does not randomly fail or remove it from CI and auto PR testing.

## Steps to Reproduce

See https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors.

## Your Environment

Standard SEMS GCC 4.8.4 auto PR build env (see above).

kyungjoo-kim:
[...]

PR #6649, opened 2020-01-24
Tacho - disable to use a recursive function

kyungjoo-kim:
## Motivation

On Cuda tasking, it seems to be dangerous to use a recursive function as kokkos tasking on cuda uses fixed size of stack size. With a recursive function, it causes illegal memory access error in a random place. This PR disable the recursive function on the CUDA active space.

### DIFF
--- a/packages/shylu/shylu_node/tacho/src/Tacho_TaskFunctor_FactorizeChol.hpp
+++ b/packages/shylu/shylu_node/tacho/src/Tacho_TaskFunctor_FactorizeChol.hpp
@@ -94,7 +94,13 @@ namespace Tacho {
 
         switch (_state) {
         case 0: { // tree parallelsim
-          if (_info.serial_thres_size > _s.max_decendant_supernode_size) {
+#if defined( KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST )
+          const bool recursive_function_work_on_exec_space = true;          
+#else
+          const bool recursive_function_work_on_exec_space = false;          
+#endif
+          if (recursive_function_work_on_exec_space && 
+              _info.serial_thres_size > _s.max_decendant_supernode_size) {
             r_val = factorize_internal(member, _s.max_decendant_schur_size, true);
             Kokkos::single(Kokkos::PerTeam(member), [&]() {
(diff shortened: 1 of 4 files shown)


kyungjoo-kim:
The previous gcc 4.8.4 error is a kind of a server error. Retest

prwolfe:
re-testing due to full disk

brian-kelley:
@kyungjoo-kim Do these changes fix the test ``ShyLU_NodeTacho_Tacho_TestSerialDoubleComplexTaskSchedulerMultiple``? After the full disk problem was fixed this morning, that test was still failing on GCC 4.8.4.

kyungjoo-kim:
GCC 4.8.4 test is not executed due to the same server failure. I do not recorgnize ``ShyLU_NodeTacho_Tacho_TestSerialDoubleComplexTaskSchedulerMultiple`` failure. What failure do you mean ? Anyway, this PR address the problem of recursive function calls on CUDA tasking, of which test is enabled when relocatable device code is ON.

brian-kelley:
@kyungjoo-kim This test started right after the server failure was resolved. This is the error output:
```
     DeviceSpace::         HostSpace::  [==========] Running 58 tests from 7 test cases.
[----------] Global test environment set-up.
[----------] 2 tests from CrsMatrixBase
[ RUN      ] CrsMatrixBase.constructor
[       OK ] CrsMatrixBase.constructor (2 ms)
[ RUN      ] CrsMatrixBase.matrixmarket
unknown file: Failure
C++ exception with description "View bounds error of view ap ( 13 < 13 )
" thrown in the test body.
[  FAILED  ] CrsMatrixBase.matrixmarket (31 ms)
[----------] 2 tests from CrsMatrixBase (33 ms total)
```

It's actually the same compiler and same test executable as #5472, but a different gtest case now that I look closer (I'm seeing ``CrsMatrixBase.matrixmarket`` fail, but the issue is about ``DenseByBlocks.trsm``). Do you think those could be related? @prwolfe

brian-kelley:
@kyungjoo-kim @prwolfe Sorry, this is totally unrelated to my PR test failure, which is #3263.

PR #6653, opened 2020-01-27
Resolve 6652 Intrepid2: Renaming variable to avoid shadow warning.

seamill:
@trilinos/<teamName>


seamill:
@CamelliaDPG Could you review this. I did this on my branch and things built and passed tests, but it probably this feature probably isn't tested in empire.

CamelliaDPG:
@seamill Looks good to me.  Thanks for the fix!

PR #6661, opened 2020-01-27
Tacho - external user interface example for collaboration

kyungjoo-kim:
## Motivation

Application solver interface example.


packages/shylu/shylu_node/tacho/example/Tacho_ExampleExternalInterface.cpp
```diff
@@ -0,0 +1,168 @@
+#include "Tacho_ExampleExternalInterface.hpp"
+#include "Tacho_CommandLineParser.hpp" 
+
+#if defined(TACHO_USE_INT_INT)
```
mhoemmen: If this is at all tied to `TPETRA_INST_INT_INT`, then this code may not build by default, and there may not yet be a PR build that exercises it.

kyungjoo-kim: Tacho uses size_type unsigned int while the application uses int for interfacing with other solvers. This is not related with Tpetra ordinal types. You are right. This is not tested in this PR test and it is intended. I tested functionality with other unit tests and examples. This is just a reproducer that I can communicate with application developers.

Issue #6655, opened 2020-01-27
Remove Deprecated Code


jeforster:
In Trilinos v12.16, many features of Trilinos were deprecated and should be removed before Trilinos v13 is released.  This issue was created to track the removal of these deprecated features.  The List of deprecations and their removal status will be updated here:

1. [Trilinos-wide next generation stack] Explicit instantiation of multiple different global ordinal types and multiple different local ordinal types Mitigation: Use default global ordinal type long long and local ordinal type int, or specify to CMake -DTpetra INST INT INT= to use global ordinal type int and local ordinal type int. **[DONE]**
2. [Tpetra] DynamicProfile CrsGraph and CrsMatrix construction: new default construction type is StaticProfile Mitigation: Use StaticProfile construction, specifying (at least) the maximum number of nonzeros per row that the graph/matrix will hold. **[DONE]**
3. [Teuchos] Use of KokkosClassic::DefaultNode and its synonyms Kokkos::Compat::Kokkos*WrapperNode as an object **[DONE]**
4. [Tpetra, Thyra, Xpetra, Ifpack2] All functions taking a KokkosClassic::DefaultNode object as an argument **[DONE]**
5. [Tpetra, Xpetra, Ifpack2, KokkosClassic] Methods getDefaultNode and getNode to return KokkosClassic::DefaultNode **[DONE]**
6. [Tpetra, Xpetra, MueLu, Ifpack2] clone methods for MultiVector, Vector, CrsGraph, CrsMatrix, Directory, Preconditioner **[IN PROGRESS]**
7. [Tpetra] getDualView method of Tpetra::MultiVector Mitigation: Use getLocalViewDevice or getLocalViewHost. Justification: Hides implementation details in Tpetra::MultiVector class. **[DONE]**
8. [Tpetra] Tpetra::MultiVector::normWeighted **[DONE]**
9. [Tpetra] Tpetra::MultiVectorFiller **[DONE]**
10. [Tpetra, Ifpack2] Methods getGlobalNumDiags, getNodeNumDiags, isLowerTriangular, isUpperTriangular Mitigation: Replicate mitigations implemented in Ifpack2; see https://github.com/trilinos/Trilinos/issues/2630. **[DONE]**
11. [Tpetra] getLocalDiagOffsets and getLocalDiagCopy overloads that return Teuchos::ArrayView Mitigation: Use overloads that return Kokkos::View **[DONE]**
[...]

PR #6670, opened 2020-01-28
Belos: removed deprecated Belos::toString()

jeforster:
@trilinos/belos

## Motivation
Removing deprecated Belos::toString() because the original name was a bad choice.  It has been replaced by Belos::convertStatusTypeToString()

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to
* Part of #6655 
* Composed of


jeforster:
Karen & Mark,

I was unaware of the rule, but it seems like a reasonable requirement.  I've updated the description to show the commands I used to build & test.

Issue #6675, opened 2020-01-28
Zoltan2:  RCM ordering might result in ineffective incomplete factorization preconditioner


jhux2:
@vbrunini has reported at issue when solving a small (268 DOF) steady incompressible Q2/Q1 fluid problem with an additional energy advection diffusion equation.  The linear system can be solved with GMRES preconditioned by ILU(0) in AztecOO, but will fail when using the same solver in Belos/Ifpack2.  

I've traced the difference in behavior to differences in how AztecOO and Zoltan2 generate the RCM ordering:

1) AztecOO chooses a root node based on the algorithm in George and Liu.  Zoltan2's default option is to do two BFS passes.
2) AztecOO uses a stable sort to order "unmarked" neighbors by increasing degree.  Zoltan2 uses std::sort, which is not guaranteed to be stable.

I've found that, at least for the Aria test problem, using a stable sort is necessary for ILU(0) to be effective.  The choice of root node has much less effect.

@trilinos/zoltan2 

Related issue: #6735.

mhoemmen:
Users might actually prefer exactly the same results as with AztecOO, but using a stable sort would be a good start.  @alanw0 tells me that the replacement of Aztec with Epetra and AztecOO led to grief, due to failure to solve some problems that could be solved before.  This suggests that we should try to use the same algorithms as before, unless we have better replacements.

srajama1:
This is an interesting problem ! I have never seen the root make that much of a difference (successfully solving and not solving) in practice. I would like to understand this problem better and keep it as an example. This goes against what I have said many times and have heard others says many times that two BFS does a decent job in finding a pseudo-peripheral node. I would like to know what is the difference in this case. Can you please send it to me ? We could write the stable sort and call it.

egboman:
I am surprised a stable sort is required as I never heard this issue mentioned anywhere. Nice job tracking it down. What do you mean by Belos/Ifpack2 (with Zoltan2) fails? The number of GMRES iterations is too high?

jhux2:
> I am surprised a stable sort is required as I never heard this issue mentioned anywhere.

I think the fact that maintaining the original ordering of the children with equal degrees results in an effective factorization is a total crap shoot.
 
> What do you mean by Belos/Ifpack2 (with Zoltan2) fails? The number of GMRES iterations is too high?

Yes, GMRES stagnates.

jhux2:
[...]

PR #6676, opened 2020-01-28
Address #6675

jhux2:
@trilinos/zoltan2 @vbrunini 

## Motivation

See issue #6675.

## Testing
I've verified that this change results in a matrix ordering that allows an Aria test problem to converge with GMRES/ILU(0) from Belos/Ifpack2.

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/6676)

### DIFF
--- a/packages/zoltan2/src/algorithms/order/Zoltan2_Sort.hpp
+++ b/packages/zoltan2/src/algorithms/order/Zoltan2_Sort.hpp
@@ -89,9 +89,9 @@ class SortPairs
     {
       // Sort in increasing (default) or decreasing order of value
       if (inc)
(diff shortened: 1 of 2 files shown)


jhux2:
It occurs to me that it may preferable to have the choice of sort be a ParameterList option.  Right now, this PR represents a change to the default RCM behavior.  @trilinos/zoltan2, please let me know what you would prefer.

srajama1:
I think the primary use case is for Ifpack2. As long as this improves or keeps the #iters for Ifpack2 use cases, this is only an improvement. I don't see why we need the unstable sort option.

mhoemmen:
This is the thing we should have done for the May 2013 milestone but didn't have time to do back then.

PR #6682, opened 2020-01-29
MueLu: adding thresholding of the interpolation values in GeometricInterpolationPFactory, see issue #6681

### LINKED ISSUE #6681: MueLu: sparser GeometricInterpolationP
## Enhancement
@trilinos/muelu @rstumin @mayrmt 

Adding an option to the GeometricInterpolationPFactory that generates sparser prolongation operators by filtering out the interpolation coefficients below 1e-6.
This generates much sparser P, R and Ac matrices while retaining good interpolation accuracy since the total number of interpolation coefficients per row is caped at 8.

lucbv:
@trilinos/muelu 

## Motivation
This is a first cut that performs reasonably well, the idea is to look at interpolation coefficient generated for each fine point.
If the coefficients are smaller than a given tolerance (1e-6), they are dropped.
This results in much sparser coarse grid as well as smaller P and R matrices.
If we want to keep this idea around we should rewritte the structured aggregation factory to obtain the correct graph directly instead of recomputing it in the prolongator factory.

## Related Issues

* Closes #6681 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing
The new algorithm has been tested locally, all MueLu examples and tests are passing.
Also performed a larger simulation using 3D elasticity with 100x100x100 nodes mesh which resulted in good performance.

### DIFF
--- a/packages/muelu/src/Graph/StructuredAggregation/MueLu_AggregationStructuredAlgorithm_def.hpp
+++ b/packages/muelu/src/Graph/StructuredAggregation/MueLu_AggregationStructuredAlgorithm_def.hpp
@@ -256,7 +256,7 @@ namespace MueLu {
     myGraph->fillComplete(domainMap, rowMap);
     *out << "Prolongator CrsGraph computed" << std::endl;
 
-  } // BuildAggregates()
+  } // BuildGraph()
 
 
   template <class LocalOrdinal, class GlobalOrdinal, class Node>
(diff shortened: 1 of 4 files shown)


lucbv:
@jhux2 it could be useful but I am not sure whether I want to keep it in its current form or not.
Ideally I would rather have this work happen at the aggregation level where one can decided to include less nodes.
This would avoid having to rebuild the graph/matrix in the prolongation factory... that's also why I am holding off on reproducing this in the kokkos counter part since it would like not scale well.

PR #6684, opened 2020-01-30
MueLu: Add cut-based dropping

cgcgcg:
@trilinos/muelu 

## Motivation
Add a cut-based dropping strategy implemented by @dsunder.

THIS IS EXPERIMENTAL!

To try it out:
```
<Parameter name="aggregation: drop scheme"             type="string" value="distance laplacian"/>
<Parameter name="aggregation: distance laplacian algo" type="string" value="scaled cut"/>
<Parameter name="aggregation: drop tol"                type="double" value="8.0"/>
```

### DIFF
--- a/packages/muelu/doc/UsersGuide/masterList.xml
+++ b/packages/muelu/doc/UsersGuide/masterList.xml
@@ -387,6 +387,15 @@
     </parameter>
 
     <parameter>
+      <name>aggregation: distance laplacian algo</name>
+      <type>string</type>
+      <default>"default"</default>
+      <description>Type of sub-algorithm for "distance laplacian" dropping. Possible values: "default", "unscaled cut", "scaled cut".</description>
+      <visible>false</visible>
+      <comment-ML>parameter not existing in ML</comment-ML>
+    </parameter>
+
+    <parameter>
       <name>aggregation: drop tol</name>
       <type>double</type>
       <default>0.0</default>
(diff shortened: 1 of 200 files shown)


packages/muelu/doc/UsersGuide/masterList.xml
```diff
@@ -386,6 +386,15 @@
       <comment-ML>parameter not existing in ML</comment-ML>
     </parameter>
 
+    <parameter>
+      <name>aggregation: distance laplacian algo</name>
+      <type>string</type>
+      <default>"default"</default>
+      <description>Type of sub-algorithm for "distance laplacian" dropping. Possible values: "default", "unscaled cut", "scaled cut".</description>
+      <visible>true</visible>
```
jhux2: Make the visibility `false` so that it doesn't show up in the manual.

cgcgcg: That was the intent :-/
Good catch!

packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
```diff
@@ -713,28 +766,158 @@ namespace MueLu {
             numTotal += indices.size();
 
             LO nnz = indices.size(), rownnz = 0;
+
+            LO rowNonUniqueMap = nonUniqueMap->getLocalElement(uniqueMap->getGlobalElement(row));
```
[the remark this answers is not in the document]
jhux2: As @cgcgcg I discussed earlier this week, local to global lookup is O(1), global to local requires a hash map lookup.

jhux2:
Any update on this?

cgcgcg:
@jhux2 I'm in the process of debugging with EMPIRE to see why rowmap != colmap on the local part.

jhux2:
> I'm in the process of debugging with EMPIRE to see why rowmap != colmap on the local part.

Thanks!  @bathmatt was asking.

cgcgcg:
@jhux2 This should be good to go now.

Issue #6648, opened 2020-01-24
Tpetra:  Increased memory usage with latest Trilinos


kddevin:
## Bug Report
@trilinos/tpetra 

### Description
From @crdohrm 
We've noticed a large increase in the memory high water mark for some of our tests. This morning we isolated it to some doImport operations on Tpetra CrsMatrices. For the case below, we have a fillCompleted CrsMatrix inputMatrix and would like to extract a subset of its rows and columns into outputMatrix (see function below).

 ```
getTpetraMatrix(RCP<const CrsMatrix> inputMatrix,
            RCP<const Map> RowMap,
            RCP<CrsMatrix> & outputMatrix) const
{
  RCP<const Map> RowMap1to1 = inputMatrix->getRowMap();
  Import Importer(RowMap1to1, RowMap);
  outputMatrix = rcp( new CrsMatrix(RowMap, RowMap, 0) );
  outputMatrix->doImport(*inputMatrix, Importer, Tpetra::INSERT);
  outputMatrix->fillComplete(RowMap1to1, RowMap1to1);
}
```

Some data for the existing version of Trilinos in Sierra follows.

Memory high water mark in MB (max over processes) for existing Trilinos in Sierra:
```
just before doImport: 688
just before fillComplete: 720
just after fillComplete: 822
```

The corresponding numbers using the new version of Trilinos:
```
just before doImport: 712
just before fillComplete: 1424
just after fillComplete: 1622
```
 
Anyway, we just wanted to know if you see anything obvious in the function above that could be done to avoid the large increase in memory. We tried constructing a CrsGraph first and then use that in the constructor for outputMatrix, but saw some errors during fillComplete of the CrsGraph (we think these errors may be related to using the same row and column map in the constructor).

Here are some more old/new Trilinos results for some other tests (notice Max memory high water):

old Trilinos:
```
PERFORMANCE TEST 1
Parallel Memory high water (MB):  Min: 180.3  Max: 360.1  Avg: 296.4  Sum: 2371.2

PERFORMANCE TEST 2
Parallel Memory high water (MB):  Min: 146.8  Max: 217.7  Avg: 183.2  Sum: 1465.7
 ```

new Trilinos:
```
PERFORMANCE TEST 1
Parallel Memory high water (MB):  Min: 241.4  Max: 847.1  Avg: 672.6  Sum: 5380.7

PERFORMANCE TEST 2
[...]

PR #6685, opened 2020-01-30
Tpetra::Crs{Graph,Matrix}: Add verbose debugging output for memory allocations 

### LINKED ISSUE #6663: Tpetra::CrsMatrix::unpackAndCombine, non-static-graph path, allocates too much memory
## Bug Report
@trilinos/tpetra  

### Description

### Steps to Reproduce

mhoemmen:
@trilinos/tpetra @tjfulle 

Add verbose debugging output to CrsGraph and CrsMatrix, to help diagnose and fix #6663 and #6648 in general.

Tested TpetraCore tests locally on my Mac laptop with Clang and OpenMPI, with `TPETRA_VERBOSE` both set and unset.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
@@ -1163,17 +1163,21 @@ namespace Tpetra {
        buffer_device_type>& permuteFromLIDs) override;
 
     void
-    applyCrsPadding (const Kokkos::UnorderedMap<local_ordinal_type, size_t, device_type>& padding);
+    applyCrsPadding(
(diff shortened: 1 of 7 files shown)


mhoemmen:
@tjfulle Hold on, Clang seems to include `<memory>` for me, but some GCCs don't.

mhoemmen:
Not sure why those tests failed -- those builds appeared to have passed.

mhoemmen:
@trilinos/framework All those are Moertel build errors, not related to Tpetra.

mhoemmen:
Thanks @jhux2 ! :-D  The extra output has already proven useful.

mhoemmen:
Preserving the branch for now, since it's useful for reproducing an experiment.

PR #6693, opened 2020-01-30
Tpetra: adding benchmark for CrsMatrix::apply, see issue #6692

### LINKED ISSUE #6692: Tpetra: add simple MatVec benchmark test
## Enhancement
@trilinos/tpetra

I would like to add a new test in tpetra that can be used to benchmark the performance of CarMatrix::apply.
Somehow we have one for BlockCrsMatrix but not one for the regular CrsMatrix?

lucbv:
@trilinos/tpetra 

## Motivation
This would just make life a little easier when I want to compare the performance of the matvec on different platforms. Also it seems somewhat surprising that we have a test for BlockCrsMatrix matvec but not for the regular CrsMatrix.

## Related Issues

* Closes #6692 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback
N/A

## Testing
I have a local build and test that pass fine on rhel7 with gcc/7.2.0 auto-tester config

### DIFF
--- a/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
+++ b/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
@@ -39,7 +39,7 @@
 // ************************************************************************
 // @HEADER
 
-// This benchmark exercises Tpetra::CrsMatrix's the apply() method.
+// This benchmark exercises Tpetra::CrsMatrix's apply() method.
 // Tpetra implements sparse matrix and dense vector data structures and
 // computational kernels for users and other Trilinos data structures.
 // Tpetra uses MPI (Message Passing Interface) for distributed-memory
--- a/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
+++ b/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
@@ -66,11 +66,11 @@ struct CmdLineOpts {
   // adequate precision.
   int numTrials;
   // Number of rows per MPI process (hence "local") in the graph;
-  // number of block rows per MPI process in the BlockCrsMatrix.
   int lclNumRows;
-  // Number of entries per row in the sparses graph; thus, number of
-  // blocks per block row of the BlockCrsMatrix.
+  // Number of entries per row in the sparses graph;
   int numEntPerRow;
+  // Bool that determines if a warm up apply is performed before timing
+  bool warmUp;
 };
 
 // Use a utility from the Teuchos package of Trilinos to set up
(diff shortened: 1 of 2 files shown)


packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
```diff
@@ -0,0 +1,380 @@
...
+      gblColInds[k] = curColInd;
+    }
+    G->insertGlobalIndices (gblInd, gblColInds ());
+  }
+  // Make the graph ready for use by CrsMatrix.
+  G->fillComplete ();
+  return G;
+}
+
+// Get a Tpetra::BlockCrsMatrix for use in benchmarks.
```
mhoemmen: It's a CrsMatrix, not a BlockCrsMatrix.  Also, we can return CrsMatrix by value now; it behaves somewhat like a Kokkos::View.

packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
```diff
@@ -0,0 +1,380 @@
...
+  using Teuchos::RCP;
+  using Teuchos::rcp;
+  using std::endl;
+  using matrix_type = Tpetra::CrsMatrix<>;
+  using device_type = matrix_type::device_type;
+  using SC  = matrix_type::impl_scalar_type;
+  using KAT = Kokkos::ArithTraits<SC>;
+  using LO  = Tpetra::Map<>::local_ordinal_type;
+
+  // mfh 02 Jun 2016: Prefer Kokkos::Serial to the HostMirror as the
```
mhoemmen: Is this still true?

packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
```diff
@@ -0,0 +1,380 @@
...
+    // norms increase or decrease a lot, that might trigger the slow
+    // case.
+    const SC X_val = static_cast<SC> (1.0) /
+      static_cast<SC> (opts.numEntPerRow);
+    X.putScalar (X_val);
+    Y.putScalar (0.0);
+
+    auto timer = TimeMonitor::getNewCounter ("Tpetra CrsMatrix apply (mat-vec)");
+    {
+      TimeMonitor timeMon (*timer);
```
mhoemmen: We should do an apply to "warm up" the UVM allocations, since the code fills the matrix on host.

packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
lucbv: I searched for "block" in the file and removed a bunch of other misleading occurrences.

packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
lucbv: This indeed looks like it might not be accurate anymore, I will remove this and use the default mirror space. If I observe anythin suspicious on CUDA I can revisit this later (I'll actually make a note in the issue so I don't forget about it).

packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
lucbv: Yes, that's a good idea, I'll add that.

lucbv:
@mhoemmen 
I have modified the test, I think it looks better now, let me know if you want more changes though.

packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
```diff
@@ -0,0 +1,380 @@
...
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact Michael A. Heroux (maherou@sandia.gov)
+//
+// ************************************************************************
+// @HEADER
+
+// This benchmark exercises Tpetra::CrsMatrix's the apply() method.
```
mhoemmen: Grammar: omit "the."

lucbv: Good catch, fixed it

PR #6694, opened 2020-01-30
MueLu: region MG Chebyshev updated to match ifpack2

pohm01:
@trilinos/muelu 

## Motivation

An update to the region MG Chebyshev relaxation so that it more closely resembles the implementation in Ifpack2.

mayrmt: added command line argument to control boost factor. Included this also into tests.

## Testing

Matches Ifpack2 on single region for unsmoothed aggregation.

mayrmt: ran `ctest -R MueLu_Structured_Region` and `ctest -R UnitTestsRegion` locally. All passed.

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
@@ -410,10 +410,10 @@ void chebyshevIterate ( RCP<Teuchos::ParameterList> params,
   // Get max number of regions per proc
   const int maxRegPerProc = regX.size();
 
-  const int maxIter      = params->get<int>   ("smoother: sweeps");
-  const Scalar eigRatio  = params->get<double>("smoother: eigRatio");
+  const int maxIter = params->get<int>   ("smoother: sweeps");
+  const Scalar eigRatio = params->get<double>("smoother: Chebyshev eigRatio");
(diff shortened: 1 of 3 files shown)


packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
```diff
@@ -413,16 +413,19 @@ void chebyshevIterate ( RCP<Teuchos::ParameterList> params,
   const int maxIter      = params->get<int>   ("smoother: sweeps");
   const Scalar eigRatio  = params->get<double>("smoother: eigRatio");
   const Scalar lambdaMax = params->get<Scalar>("chebyshev: lambda max");
-  const Scalar lambdaMin = lambdaMax / eigRatio;
+  const Scalar boostFactor = 1.1; // TODO: add as a param
```
mayrmt: @pohm01 You could read this parameter from the command line in the driver and then pass it thought the parameter list as entry `chebyshev: boost factor`.

mayrmt: Added the command line argument `--smootherChebyBoostFactor` to control this parameter from the command line.

mayrmt:
@pohm01 I added the command line argument to control the boost factor and took the liberty to update the PR description accordingly.

mhoemmen:
Yay code duplication?  I guess MueLu can't depend directly on Ifpack2, but maybe it would make sense for MueLu to have a single common interface to algorithms like Chebyshev.

PR #6704, opened 2020-01-31
IOSS: cgns - fix handling of periodic single-block models in parallel

gsjaardema:
@trilinos/seacas 

## Motivation

Added a `Descriptor_t` node under the `GridConnectivity1_to1_t` node during decomposition
if the GridConnectivity node is created as the result of parallel decomposition (connectivity from one processor to another).  This is needed for periodic models in which, for example, the K_max face is connected to the K_min face.  The name of the Descriptor node is 'Decomp' and the value is (currently) unused.

Modified fpp output of Iocgns_DatabaseIO to add this descriptor to all models.  Models without this descriptor that worked previous to the change will continue to work post-change; it is only required for periodic models in which a zone is adjacent to itself.

A test is being developed with a smaller model to ensure that the behavior does not regress; it will be added at a later time.

### DIFF
--- a/packages/seacas/libraries/ioss/src/main/CMakeLists.txt
+++ b/packages/seacas/libraries/ioss/src/main/CMakeLists.txt
@@ -380,6 +380,21 @@ TRIBITS_ADD_ADVANCED_TEST(structured_cgns_roundtrip_zgc
   COMM mpi serial
   )
 
+TRIBITS_ADD_ADVANCED_TEST(structured_cgns_periodic_roundtrip_fpp
+   TEST_0 EXEC io_shell ARGS --compose external ${CMAKE_CURRENT_SOURCE_DIR}/test/self-looping-grid.cgns self-looping-fpp.cgns
+     NUM_MPI_PROCS 1-4
+     NOEXEPREFIX NOEXESUFFIX
+   TEST_1 EXEC io_shell ARGS --external --compose external self-looping-fpp.cgns self-looping2.cgns
+     NUM_MPI_PROCS 1-4
+     NOEXEPREFIX NOEXESUFFIX
+   TEST_2 EXEC io_shell ARGS --external self-looping2.cgns self-looping3.cgns
+     NUM_MPI_PROCS 1-4
+     NOEXEPREFIX NOEXESUFFIX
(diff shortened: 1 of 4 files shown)


mhoemmen:
@gsjaardema Would you be OK waiting to merge until you have a test in place, or is this capability urgently needed in Trilinos?

gsjaardema:
@mhoemmen A test has been added.

packages/seacas/libraries/ioss/src/main/CMakeLists.txt
```diff
@@ -380,6 +380,21 @@ TRIBITS_ADD_ADVANCED_TEST(structured_cgns_roundtrip_zgc
   COMM mpi serial
   )
 
+TRIBITS_ADD_ADVANCED_TEST(structured_cgns_periodic_roundtrip_fpp
+   TEST_0 EXEC io_shell ARGS --compose external ${CMAKE_CURRENT_SOURCE_DIR}/test/self-looping-grid.cgns self-looping-fpp.cgns
+     NUM_MPI_PROCS 1-4
```
bartlettroscoe: Is a range of procs allows `1-4`?  There must not be a check for that in the CMake code.  I think that this will build a command of the form:

```
mpiexec -np 1-4 ...
```  

That is not what you want, is it?

gsjaardema: It has been working that way in other tests for a few years...  I thought I got the original example from other Trilinos code.

bartlettroscoe: Let's watch the output when this runs on CDash and see what happens.

gsjaardema:
No, it is only active if CGNS TPL is enabled.

packages/seacas/libraries/ioss/src/main/CMakeLists.txt
mhoemmen: @bartlettroscoe I always interpreted "1-4" as "the test may run on any number of MPI processes 1, 2, 3, or 4."  Tpetra has used that for as long as I've been working on Tpetra.

gsjaardema:
Thanks.

PR #6705, opened 2020-01-31
ATDM Fix issue with empty cache vars (CDOFA-72)

bartlettroscoe:
Should fix a problem with empty cache var set directly from empty env var reported by @bathmatt.

I tested locally on 'ews00232' with 'sems-rhel7' env and it showed:

```
$ grep Deprecated CMakeCache.txt 
Kokkos_ENABLE_Deprecated_Code:BOOL=ON
```

so hopefully this is okay now for EMPIRE.

But note that none of the ATDM Trilinos builds showed any problem with the code the way it was so not sure why EMPIRE was seeing a problem and in what context.


cmake/std/atdm/ATDMDevEnvSettings.cmake
```diff
@@ -57,6 +57,9 @@ ATDM_SET_ATDM_VAR_FROM_ENV_AND_DEFAULT(CUDA_RDC OFF)
 ATDM_SET_ATDM_VAR_FROM_ENV_AND_DEFAULT(FPIC OFF)
 ATDM_SET_ATDM_VAR_FROM_ENV_AND_DEFAULT(COMPLEX OFF)
 
+ATDM_SET_ATDM_VAR_FROM_ENV_AND_DEFAULT(Kokkos_ENABLE_Serial ON)
+ATDM_SET_ATDM_VAR_FROM_ENV_AND_DEFAULT(Kokkos_ENABLE_Deprecated_Code ON)
+
```
bathmatt: This should probably be off...

bartlettroscoe: That is just the default, which has not changed, yet.

bartlettroscoe:
> This fix doesn't fix our issues.

Can you please post an a GitHUb issue with more detail and exact reprodocuability instructions?

bathmatt:
Hey @rppawlo can you resolve this?

I'm using this in our buildscripts and you need a clean directory

dir-Trilinos []$ 59 $ ../BuildScripts/build_trilinos.py --compiler clang  --src-dir /home/mbetten/Trilinos/  --stage configure


Happens with clang, gnu and I'm assuming everything else

rppawlo:
@bathmatt - sure!

bartlettroscoe:
@bathmatt and @rppawlo, now that I know the source of the error with EMPIRE's Trilinos configuration, I know how to change this to make this pass for EMPIRE's usage.  Should I bother or just have EMPIRE developers deal with this in EMPIRE's scripts (as per @rppawlo's offline email)?

rppawlo:
@bartlettroscoe - either way is fine with me. If you think this might come up again with another atdm code, maybe it should go into trilinos. I was only avoiding the change to trilinos since I think the kokkos merge is imminent and if they did rework everything, this might cause conflicts. Let me know if you wan tot change in trilinos and I can cancel the empire PR.

bartlettroscoe:
@rppawlo, whatever is faster I guess is the best solution.

rppawlo:
ok - I'll push the change on the empire side.

bartlettroscoe:
Should be okay to merge.

PR #6712, opened 2020-01-31
Remove Kokkos related code from Intrepid

mperego:
@trilinos/intrepid 

## Motivation
Kokkos related code in Intrepid has been disabled/deprecated since long time.
Here we are purging all the Kokkos related code.

Related to #6655


mperego:
@CamelliaDPG Thanks. Would you mind approving again? For some obscure reasons Github asked me to resolve a conflict.. which ended up in a useless extra commit.

Issue #6691, opened 2020-01-30
Tpetra: BlankRowBugTest_MPI_2 Seg faulting on vortex


e10harvey:
## Question
@trilinos/tpetra 

We are trying to stabilize a new environment introduced in #6547 but are seeing many seg faults in multi-process Tpetra tests. The set of failing tests experiencing seg faults are not failing intermittently. What's causing the seg faults in BlankRowBugTest_MPI_2 on vortex? Is this an issue with how we're running the tests? See [CDASH](https://testing.sandia.gov/cdash/testDetails.php?test=100357586&build=6286330).

To reproduce, see: [CDOFA-94](https://sems-atlassian-srn.sandia.gov/browse/CDOFA-94?focusedCommentId=50902&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-50902).

As shown in CDASH and CDOFA-94, there are two unmapped addresses: `0x7fff34937e80`, which is read by rank 1, while `0x7fff14937a80` is read by rank 0.
  0. Does the base of these addresses first come in through rank 1 (via `m_impl_handle`) at `Kokkos::Impl::ViewDimension5`, `m_map = {m_impl_handle = 0x7fff34938080, <snip>`?
  1. Is this related to https://github.com/trilinos/Trilinos/blob/develop/packages/tpetra/core/src/Tpetra_Map_def.hpp#L2215?
  2. Is this branch (https://github.com/trilinos/Trilinos/blob/develop/packages/teuchos/comm/src/Teuchos_CommHelpers.cpp#L647) exercised by BlankRowBugTest_MPI_2 on other machines using spectrum MPI?

CC: @bartlettroscoe.

mhoemmen:
@e10harvey Rerun with environment variable `TPETRA_DEBUG` set to 1.

bartlettroscoe:
NOTE: We can't really support Trilinos on ATS-2 until we work this out as we have about 270 failing tests which is too much for us to even start to monitor with our current system.

I am sure there is something to do with how we are launching the job with LFS bsub or using mpiexec.  It might even be related to an issue that @jjellio reported a while back in #5855 about `-disable_gpu_hooks` ?  (But have not had time to investigate myself.)

mhoemmen:
@e10harvey One possibility is that someone is passing UVM memory down through Tpetra or through `Teuchos::*send`.

vbrunini:
[...]

PR #6718, opened 2020-02-03
Testing/Framework: Add jsrun wrapper

jjellio:
The short is that -disable_gpu_hooks is something that needs to be applied to single process runs (by default spectrum loads a shared library that hooks all kinds of things, but those hooks depend on MPI_Init being called to initialize things).

The goal of this effort, which probably should be moved to a specific issue, is that Trilinos needs to have atleast two code paths tested:

Tpetra w/cuda-aware, which requires -M -gpu, and
Tpetra w/out cuda-aware which requires not having -M -gpu (I've yet to hear what the flag is to specifically disable GPU aware MPI (there is only an option to enable it).
All of the above are runtime options. When you invoke jsrun without -M -gpu, you are getting an entirely different implementation of MPI.

The wrapper I've proposed would enable Trilinos to configure with a fake jsrun that magically handles the -M -gpu flag based on whether Tptera's cuda-aware ENV is set to 0 or 1. Also, the wrapper automatically disables the GPU hooks if the process is not MPI-parallel.

I played the game of Vortex unit testing over the summer of 2019. I seem to remember that I was able to get most tests passing, but it required substantial find/replace on the generated ctest files.

To test vortex with both cuda-aware and regular MPI, the testing harness will need to do something like:

configure w/wrapped jsrun (via ATDM ENV's)
build once
set TPETRA_ASSUME_CUDA_AWARE_MPI=0 (do the regular MPI case)
invoke ctest
post the results to a dashboard
do something to change the 'build_id' of the ctest file. e.g.,
Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg
to
Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_cuda-aware_static_dbg
set TPETRA_ASSUME_CUDA_AWARE_MPI=1
reset ctest (clear all pass/fail stuff)
invoke ctest
post the results to a dashboard, they will get attributed to the modified build_id mentioned in (6)
I'll need to add the logic to set -M -gpu to the wrapped jsrun, and the testing harness will need to figure out how to do 6-10.

It's worth noting, this same testing workflow could be implemented on White/Ride/Waterman. On those machines the MPI impl is effectively running with the openMPI variant of -M -gpu. So on those machines mpirun would need to be wrapped to disable cuda-aware MPI.

There is a 'test', but it relies on manual inspection
`Trilinos/cmake/std/atdm/ats2/trilinos_jsrun_test.sh`

@e10harvey 
Please mark this with appropriate labels.

### DIFF
--- a/cmake/std/atdm/ats2/trilinos_jsrun
+++ b/cmake/std/atdm/ats2/trilinos_jsrun
@@ -13,6 +13,11 @@ function debug_print {
   fi
 }
 
+if [[ "$TPETRA_ASSUME_CUDA_AWARE_MPI" != "0" ]] && [[ "$TPETRA_ASSUME_CUDA_AWARE_MPI" != "1" ]]; then
+  echo "WARNING, you have not set TPETRA_ASSUME_CUDA_AWARE_MPI=0 or 1, defaulting to TPETRA_ASSUME_CUDA_AWARE_MPI=0"
+  export TPETRA_ASSUME_CUDA_AWARE=0
+fi
+
 for i in "${!args[@]}";
 do
   arg=${args[${i}]}
@@ -68,7 +73,7 @@ if [ "$np_is_one" == "true" ]; then
       debug_print "arg=$arg, found -M, prepending -disable_gpu_hooks"
       ip1=$(($i+1));
       # single quotes matter, as the argument to -M needs to be all args originally quoted to it
-      args[${ip1}]="'-disable_gpu_hooks ${args[${ip1}]}'"
+      args[${ip1}]="-disable_gpu_hooks ${args[${ip1}]}"
       debug_print "updated arg = ${args[${ip1}]}"
       added_disable="true"
       break
@@ -97,7 +102,7 @@ if [[ "$TPETRA_ASSUME_CUDA_AWARE_MPI" == "1" ]] && [[ "$np_is_one" == "false" ]]
       debug_print "arg=$arg, found -M, prepending -gpu"
       ip1=$(($i+1));
       # single quotes matter, as the argument to -M needs to be all args originally quoted to it
-      args[${ip1}]="'-gpu ${args[${ip1}]}'"
+      args[${ip1}]="-gpu ${args[${ip1}]}"
       added_arg="true"
       break
     fi
@@ -112,8 +117,8 @@ if [[ "$TPETRA_ASSUME_CUDA_AWARE_MPI" == "1" ]] && [[ "$np_is_one" == "false" ]]
 fi
 
 if [ "$ECHO_CMD" == "1" ]; then
-  1>&2 echo "TPETRA_ASSUME_CUDA_AWARE=${TPETRA_ASSUME_CUDA_AWARE_MPI} BEFORE: jsrun ${orig_args[@]}"
-  1>&2 echo "TPETRA_ASSUME_CUDA_AWARE=${TPETRA_ASSUME_CUDA_AWARE_MPI} AFTER : jsrun ${args[@]}"
+  1>&2 echo "TPETRA_ASSUME_CUDA_AWARE_MPI=${TPETRA_ASSUME_CUDA_AWARE_MPI} BEFORE: jsrun " $(printf "'%s' " "${orig_args[@]}")
+  1>&2 echo "TPETRA_ASSUME_CUDA_AWARE_MPI=${TPETRA_ASSUME_CUDA_AWARE_MPI} AFTER : jsrun " $(printf "'%s' " "${args[@]}")
 fi
-eval jsrun ${args[@]}
+eval jsrun $(printf "'%s' " "${args[@]}")
 
--- a/cmake/std/atdm/ats2/trilinos_jsrun
+++ b/cmake/std/atdm/ats2/trilinos_jsrun
@@ -1,0 +1,119 @@
+#!/bin/bash
+
+DEBUG_SCRIPT="${DEBUG_SCRIPT-:0}"
+ECHO_CMD="${ECHO_CMD=:0}"
+np_is_one="false"
+orig_args=("$@")
+args=("$@")
+
+function debug_print {
+  if [ "$DEBUG_SCRIPT" == "1" ]; then
+    msg="$@"
+    1>&2 echo "$msg"
+  fi
+}
+
+for i in "${!args[@]}";
+do
+  arg=${args[${i}]}
+
+  debug_print "Processing arg=$arg"
+
+  # handle the case of seeing the np argument
+  # lets assume -p ## or --np=##
+  if [[ "$arg" =~ ^-p[0-9]+$ ]]; then
+    # np is just the 2nd arg split on 'p'
+    my_np=$(echo "$arg" | cut -f2 -d'p');
+    debug_print "  Case: -p##, with ##=$my_np"
+    if [ "$my_np" == "1" ]; then
+      np_is_one="true";
+      debug_print "  Case: -p##, with ##=$my_np and np=1"
+    fi
+    # we are finished
+    break;
+  # next, -p ##
+  elif [ "$arg" == "-p" ]; then
+    ip1=$(($i+1));
+    my_np=${args[${ip1}]}
+    debug_print "  Case: -p ##, with ##=$my_np"
+    if [ "$my_np" == "1" ]; then
+      np_is_one="true";
+      debug_print "  Case: -p ##, with ##=$my_np and np=1"
+    fi
+    # we are finished
+    break;
+  # next, --np=##
+  elif [[ "$arg" =~ ^--np=[0-9]+$ ]]; then
+    # np is just the 2nd arg split on '='
+    my_np=$(echo "$arg" | cut -f2 -d'=');
+    debug_print "  Case: --np=##, with ##=$my_np"
+    if [ "$my_np" == "1" ]; then
+      np_is_one="true";
+      debug_print "  Case: --np=##, with ##=$my_np and np=1"
+    fi
+    # we are finished
+    break;
+  fi
+done
+
+if [ "$np_is_one" == "true" ]; then
+  added_disable="false"
+  debug_print "NP=1 adding disable_gpu_hooks"
+  # look for an '-M' arg and append disable hooks to the next argument
+  for i in "${!args[@]}";
+  do
+    arg=${args[${i}]}
+    debug_print "arg=$arg"
+    if [ "$arg" == "-M" ]; then
+      debug_print "arg=$arg, found -M, prepending -disable_gpu_hooks"
+      ip1=$(($i+1));
+      # single quotes matter, as the argument to -M needs to be all args originally quoted to it
+      args[${ip1}]="'-disable_gpu_hooks ${args[${ip1}]}'"
+      debug_print "updated arg = ${args[${ip1}]}"
+      added_disable="true"
+      break
+    fi
+  done
+
+  # did not find -M, so we add it
+  if [ "$added_disable" == "false" ]; then
+    debug_print "NP=1, but we didn't find a -M argument, so adding -M and disabling the gpu_hooks"
+    # prepend the argument
+    args=("-M -disable_gpu_hooks" "${args[@]}")
+  fi
+fi
+
+# I think if np=1 and disable_gpu_hooks is passed, then you aren't supposed to use -gpu
+if [[ "$TPETRA_ASSUME_CUDA_AWARE_MPI" == "1" ]] && [[ "$np_is_one" == "false" ]]; then
+  debug_print "TPETRA_ASSUME_CUDA_AWARE_MPI=1 and np>1, adding -M -gpu"
+  added_arg="false"
+
+  # look for an '-M' arg and prepend an arg (this could be a function)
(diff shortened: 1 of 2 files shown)


e10harvey:
Related to #6691.

e10harvey:
@jjellio: Would you add support for '-np', please?
```bash
[eharvey@vortex59 Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg]$ "../Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-p" "1" "--rs_per_socket 4" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe"
TPETRA_ASSUME_CUDA_AWARE=0 BEFORE: jsrun -p 1 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe
TPETRA_ASSUME_CUDA_AWARE=0 AFTER : jsrun -M -disable_gpu_hooks -p 1 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe
Test cuBLAS and cuSOLVER handle creation
s.get () != nullptr = 1 == true : passed
Original x: 666: Converted x: 666
b.get () != nullptr = 1 == true : passed
C_h(0,0) = 7.00000000000000000e+00 == 7.00000000000000000e+00 : passed
C_h(0,0) = 3.10000000000000000e+01 == 3.10000000000000000e+01 : passed
s.get () != nullptr = 1 == true : passed
Original x: 666: Converted x: 666
b.get () != nullptr = 1 == true : passed
C_h(0,0) = 7.00000000e+00 == 7.00000000e+00 : passed
C_h(0,0) = 3.10000000e+01 == 3.10000000e+01 : passed

End Result: TEST PASSED
(reverse-i-search)`mpiexe': /usr/tce/packages/spectrum-mpi/spectrum-mpi-2019.06.24-gcc-7.3.1/bin/mpiexec -gpu "-np" "2" "-map-by" "socket:PE=4" "./packages/tpetra/core/test/BugTests/TpetraCore_BlankRowBugTest.exe"
[eharvey@vortex59 Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg]$ "../Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-p" "1" "--rs_per_socket 4" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe"
TPETRA_ASSUME_CUDA_AWARE=0 BEFORE: jsrun -p 1 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe
TPETRA_ASSUME_CUDA_AWARE=0 AFTER : jsrun -M -disable_gpu_hooks -p 1 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe
Test cuBLAS and cuSOLVER handle creation
s.get () != nullptr = 1 == true : passed
Original x: 666: Converted x: 666
b.get () != nullptr = 1 == true : passed
C_h(0,0) = 7.00000000000000000e+00 == 7.00000000000000000e+00 : passed
C_h(0,0) = 3.10000000000000000e+01 == 3.10000000000000000e+01 : passed
s.get () != nullptr = 1 == true : passed
Original x: 666: Converted x: 666
b.get () != nullptr = 1 == true : passed
C_h(0,0) = 7.00000000e+00 == 7.00000000e+00 : passed
C_h(0,0) = 3.10000000e+01 == 3.10000000e+01 : passed

End Result: TEST PASSED
[eharvey@vortex59 Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg]$ "../Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-np" "1" "--rs_per_socket 4" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe"
TPETRA_ASSUME_CUDA_AWARE=0 BEFORE: jsrun -np 1 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe
TPETRA_ASSUME_CUDA_AWARE=0 AFTER : jsrun -np 1 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/tsqr/test/TpetraTSQR_CuSolver.exe
Error, invalid argument: p

Usage: jsrun [OPTION...] [<command>]
Try `jsrun --help' or `jsrun --usage' for more information.
[eharvey@vortex59 Trilinos-atdm-vortex-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg]$ cd ../Trilinos/cmake/std/atdm/ats2
[eharvey@vortex59 ats2 (add_jsrun_wrapper)]$ git log-short
62fc99f31c (HEAD -> add_jsrun_wrapper, jjellio/add_jsrun_wrapper) Draft of jsrun wrapper to allow for automatic testing of cuda-aware and regular MPI code paths.
```

jjellio:
@e10harvey 
It shouldn't need any NP stuff.  There is a CMake variable `Trilinos_MPI_EXEC_NUMPROCS_FLAG` or maybe just `MPI_EXEC_NUMPROCS_FLAG`.   If it isn't exposed in the ENV interface, then an `ATDM_MPI_EXEC_NUMPROCS_FLAG` needs to be added and have it map to the Cmake one

cmake/std/atdm/ats2/trilinos_jsrun
```diff
@@ -0,0 +1,119 @@
...
+
+  # did not find -M, so we add it
+  if [ "$added_disable" == "false" ]; then
+    debug_print "NP=1, but we didn't find a -M argument, so adding -M and disabling the gpu_hooks"
+    # prepend the argument
+    args=("-M -disable_gpu_hooks" "${args[@]}")
+  fi
+fi
+
+# I think if np=1 and disable_gpu_hooks is passed, then you aren't supposed to use -gpu
```
e10harvey: @jjellio,

- Thinking about this more now, I am confused. I don't have a very good understanding of Trilinos, MPI, etc. I'm also not sure exactly how test commands are configured in Trilinos. For the ATDM build environment, it sets "mpiexec flags" and an "mpiexec" program for a given system, like vortex, for example. I believe these "mpiexec" options apply to all tests that are enabled at configure time and that individual tests or packages do not set their own "mpiexec" options. @bartlettroscoe, can you confirm?

- Why is `/home/projects/ppc64le-pwr9-nvidia/openmpi/2.1.2/gcc/7.2.0/cuda/9.2.88/bin/mpiexec` able to handle executing both mpi based tests and cuda+mpi based tests on waterman? In looking at the waterman Jenkins projects and test commands, waterman just runs a cuda aware mpiexec and doesn't parse mpi arguments and hack them. Does `/home/projects/ppc64le-pwr9-nvidia/openmpi/2.1.2/gcc/7.2.0/cuda/9.2.88/bin/mpiexec` look at the application binary and determine which HW devices that binary is targeting?

I am concerned that #6718 and #6691 are attempting to work around an issue that should be fixed in spectrum MPI.

I think there are actually 3 issues that should be addressed:
1. The way we currently run tests doesn't exercise both paths affected by setting `TPETRA_ASSUME_CUDA_AWARE_MPI` to either `0` or `1`.
2. Spectrum MPI needs to be updated to properly execute application binaries that target either GPU, HOST, or  HOST+GPU.
3. Until 2. is addressed, work around it with #6718 and #6691.

Thoughts?

jjellio: MPI on waterman/ride/white is defaulting to enable cuda-aware. On Vortex, you shouldn't use mpiexec, instead you need to use jsrun (similar to using srun on Mutrino).

To be honest, many of the issues we've had are because the MPI on waterman/white are *not* making it clear which code path they are using. This explanation is probably easier in person.

Tpetra supports a runtime option to decide whether to use device buffers w/MPI, or to only use host-side buffers with MPI. Tpetra configures the **default** setting for this runtime parameter at *build time*.   If CMake determined that MPI is cuda-aware, then Tpetra will assume it is cuda aware and pass device buffers.  You can change this behavior by exporting `TPETRA_ASSUME_CUDA_AWARE_MPI`, and at Tpetra::initialize it will change the default.

What stinks, is that running configure on the login node may report that MPI is not cuda aware (which was the case on LLNL Lassen).  Yet, on the compute nodes, MPI would report it is cuda-aware.  This leads to confusion over which implementation of MPI we are using.

The implementation of MPI is (cuda-aware or not), is basically like using two different MPI's from a performance/speed perspective.  The flaw in general is that these options are **runtime**, but they require independent code paths inside our own code.  This causes the problems we have right now. Tpetra must implement logic to copy devices buffers to the host if you run on a machine where you can't or don't want to use device buffers w/MPI (but you do want to use the device for compute). 

The end result is that users and developers never had to worry about MPI flags aside from performance tuning (binding things right, choosing your 'np', etc...). With Device specific MPI, we now have MPI implementation details leaking into logic inside libraries like Tpetra. Hence, people can choose (or not set MPI flags), and Tpetra tries to make the library functional <- because some platforms historically had devices but no MPI was allowed with device buffers.

Our own test beds made a switch to device-aware MPI several years ago, but unfortunately I don't think we didn't realize how impactful that was.  For years, our codes have been CI tested with a very specific code path. Now, the fact there are multiple code paths has been thrust into the spot light.  By default, Vortex *use* to report that MPI was not cuda-aware, a very different thing than White (where it would report aware).

Spectrum does support everything for (2).  See my comment in the WIP issue

e10harvey: > MPI on waterman/ride/white is defaulting to enable cuda-aware. On Vortex, you shouldn't use mpiexec, instead you need to use jsrun (similar to using srun on Mutrino).
> 
> To be honest, many of the issues we've had are because the MPI on waterman/white are _not_ making it clear which code path they are using. This explanation is probably easier in person.

How is `/home/projects/ppc64le-pwr9-nvidia/openmpi/2.1.2/gcc/7.2.0/cuda/9.2.88/bin/mpiexec` able to handle executing both mpi based tests and cuda+mpi based tests on waterman? I mean, there is no need to pass `-gpu` or `-disable_gpu_hooks`, `/home/projects/ppc64le-pwr9-nvidia/openmpi/2.1.2/gcc/7.2.0/cuda/9.2.88/bin/mpiexec` figures it out, right?

jjellio: No it doesn't figure anything out.

`-M -gpu` is enabling device pointers w/MPI.  Waterman default to this behavior.   With that feature enabled, you get a superset of support. MPI will accept both host and device pointers... so adding `-M -gpu` tends to 'fix' things.... but in reality is isn't fixing anything. It is masking the fact that your code is using MPI in an unpredictable way.  Since Tpetra/Trilinos supports the use case that you can use 'accelerators' independent of whether MPI supports their memory address, then it is important that the library enforce that it passes MPI buffers in a safe way.

The 'feature' of having device memory accessible to MPI can result in a 40x hit in terms of network performance, which is why apps may choose not to use it.  It has also been buggy. Hence, we want to test the case of using Cuda with MPI+Cuda buffers.

jjellio:
I've added the conditional to check for Tpetra's ENV being set, and if not it defaults to the non-device aware MPI.

I also fixed a bug that occurs when you have args that need to be quoted (have whitespace).

e10harvey:
@jjellio: Thanks for your help! I've put these changes into #6724.

PR #6719, opened 2020-02-03
ATDM: Add ATS-2 section to README

e10harvey:
## How were these changes tested?
By logging on to vortex and following the example shown in the changes introduced by this PR.

### DIFF
--- a/cmake/std/atdm/README.md
+++ b/cmake/std/atdm/README.md
@@ -617,6 +617,7 @@ example, skip the configure, skip the build, skip running tests, etc.
 * <a href="#spack-rhel-environment">Spack RHEL Environment</a>
 * <a href="#cee-rhel6-environment">CEE RHEL6 and RHEL7 Environment</a>
 * <a href="#waterman">waterman</a>
+* <a href="#vortex">ATS-2</a>
 
 
 ### ride/white
@@ -1034,6 +1035,39 @@ $ bsub -x -Is -n 20 \
 ```
 
 
+### vortex
+
+Once logged on to 'vortex' (SRN), one can either build and configure on the
+login node or the compute node. Make sure to setup SSH keys as described in
+`/opt/VORTEX_INTRO` before trying to build on a compute node. To get a list of
+compute nodes, run `bhosts`, then simply ssh to a given hostname. For example to
+configur, build and run the tests for the default `cuda-debug` build for
+`kokkos` (after cloning Trilinos on the `develop` branch), do:
+
+```bash
+$ cd <some_build_dir>/
+
+$ source $TRILINOS_DIR/cmake/std/atdm/load-env.sh cuda-debug
+
+$ cmake -G'Unix Makefiles' \
+  -DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake \
+  -DTrilinos_ENABLE_TESTS=ON \
+  -DTrilinos_ENABLE_Kokkos=ON \
+  $TRILINOS_DIR
+
+$ make NP=20
+
+$ ctest
+```
+
+**NOTES:**
+- Do NOT do `module purge` before loading the environment. Simply start off with
+  a clean default environment on vortex.
+- One can also `ssh` to a compute node (if you built on the login node) and
+run ctest from there.
+- To target the GPUs on vortex, one must run tests from the login node.
+
+
 ## Building and installing Trilinos for ATDM Applications
 
 See the following internal SNL wiki page for instructions on building and
@@ -1406,7 +1440,7 @@ they support are:
 
 * `spack-rhel/`: RHEL (and likely other Linux) systems with the SNL ATDM Spack modules installed.
 
-* `serrano/`: Supports SNL HPC CTS-1 machines 'serrano', 'eclipse', and
+* `cts1/`: Supports SNL HPC CTS-1 machines 'serrano', 'eclipse', and
   'ghost'.
 
 * `shiller/`: Supports GNU, Intel, and CUDA builds on both the SRN machine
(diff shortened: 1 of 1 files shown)


cmake/std/atdm/README.md
```diff
@@ -1034,6 +1035,39 @@ $ bsub -x -Is -n 20 \
...
+
+$ cmake -G'Unix Makefiles' \
+  -DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake \
+  -DTrilinos_ENABLE_TESTS=ON \
+  -DTrilinos_ENABLE_Kokkos=ON \
+  $TRILINOS_DIR
+
+$ make NP=20
+
+$ ctest
```
bartlettroscoe: This is not really true if you build on a login node, is it?  What is the exact command to run ctest on a compute node?

cmake/std/atdm/README.md
```diff
@@ -1034,6 +1035,39 @@ $ bsub -x -Is -n 20 \
...
+```
+
+**NOTES:**
+- Do NOT do `module purge` before loading the environment. Simply start off with
+  a clean default environment on vortex.
+- One can also `ssh` to a compute node (if you built on the login node) and
+run ctest from there.
```
bartlettroscoe: Is this really true? Is this only allowed if you first allocate the nodes first?

cmake/std/atdm/README.md
bartlettroscoe: Also, technically, should they not run `ctest -j4` because that is what the Jenkins build does, no?

cmake/std/atdm/README.md
```diff
@@ -1034,6 +1035,39 @@ $ bsub -x -Is -n 20 \
...
+
+$ source $TRILINOS_DIR/cmake/std/atdm/load-env.sh cuda-debug
+
+$ cmake -G'Unix Makefiles' \
+  -DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake \
+  -DTrilinos_ENABLE_TESTS=ON \
+  -DTrilinos_ENABLE_Kokkos=ON \
+  $TRILINOS_DIR
+
+$ make NP=20
```
bartlettroscoe: This is technically not correct because the Ninja generator is not being used.  This needs to be `make -j20`.  For details, see:

* https://tribits.org/doc/TribitsBuildReference.html#building-in-parallel-with-ninja

cmake/std/atdm/README.md
```diff
@@ -1034,6 +1035,39 @@ $ bsub -x -Is -n 20 \
 ```
 
 
+### vortex
+
+Once logged on to 'vortex' (SRN), one can either build and configure on the
+login node or the compute node. Make sure to setup SSH keys as described in
+`/opt/VORTEX_INTRO` before trying to build on a compute node. To get a list of
+compute nodes, run `bhosts`, then simply ssh to a given hostname. For example to
+configur, build and run the tests for the default `cuda-debug` build for
```
bartlettroscoe: Change 'configur' to 'configure'.

bartlettroscoe: If this is running from a compute node, please show the commands to move on to a compute node to make this clear.

cmake/std/atdm/README.md
e10harvey: > This is not really true if you build on a login node, is it?

Is is.

e10harvey: > Also, technically, should they not run `ctest -j4` because that is what the Jenkins build does, no?

I will check the Jenkins build and update accordingly.

cmake/std/atdm/README.md
e10harvey: Yes, this is true. No, you do not need to allocated nodes first.

cmake/std/atdm/README.md
e10harvey: I think we should wait to continue this work on this until #6724 is rejected or accepted. 

Currently, we run tests via raw `mpiexec` (you can do this by hopping on a login or compute node with ssh and running the `mpiexec` command or `ctest`). But, if we use the jsrun wrapper going forward, we will need to be within a compute node allocation from `bsub` before running tests.

bartlettroscoe:
@e10harvey, looking at:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/ctest/drivers/atdm/ats2/local-driver.sh
* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/ats2/environment.sh#L256
* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/common/define_atdm_run_script_on_local_node.sh

it looks like the configure and build are just running on the same node.  Are the jenkins slaves just allocating compute nodes and running there are or these actually login nodes?  Also, why not just simply:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/ctest/drivers/atdm/ats2/local-driver.sh

and make it just be:

```
source $WORKSPACE/Trilinos/cmake/ctest/drivers/atdm/ctest-s-driver.sh
```

like you see for other systems.

bartlettroscoe:
> I think we should wait to continue this work on this until #6724 is rejected or accepted.

@e10harvey, agreed.  I added the label `AT: WIP` top stop the PR tester until we are really to do final review and merge.

I am still a bit confused about what the rules are for where (i.e. login node vs. compute node) where you need to load env, configure, build, run ctest, run an individual test, etc.

e10harvey:
> Are the jenkins slaves just allocating compute nodes and running there are or these actually login nodes?

Judging by the cdash output, the jenkins slaves run on both compute and login nodes. Currently, our configure, build, and test just runs on the same node that the slaves is running on.

e10harvey:
> I am still a bit confused about what the rules are for where (i.e. login node vs. compute node) where you need to load env, configure, build, run ctest, run an individual test, etc.

AFAIK, you can do anything except run a test that targets GPUs on any node. If you're targeting GPUs, you must start that test from a login node.

e10harvey:
Closing. Moved commits to #6724.

PR #6722, opened 2020-02-03
ifpack2: Fix ApplyInverseSGS_BlockCrsMatrix import operations.

### LINKED ISSUE #5885: Ifpack2: TpetraBlockCrs Import changes need to be propagated
## Bug Report
@trilinos/tpetra
@trilinos/ifpack2

### Description
The recent change in tpetra to disallow the usage of import with the block vector format has not been propagated to Ifpack2 relaxation routines. Specifically in Sierra we are seeing an error in ApplyInverseSGS_BlockCrsMatrix. It appears the importer needs to be changed to use the point map. Until this happens Sierra won't be able to update to the latest Trilinos. Can someone please look into this?

vbrunini:
This appears to fix the issue for #5885.

@trilinos/Ifpack2

## Motivation
BlockMultiVector Import throws and says to use a point MultiVector import instead.

## Related Issues

* Closes #5885

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
@@ -1830,6 +1830,13 @@ ApplyInverseGS_BlockCrsMatrix (const block_crs_matrix_type& A,
                       static_cast<local_ordinal_type> (yBlock.getNumVectors ())));
     }
     yBlockCol = yBlockColumnPointMap_;
+    if (pointImporter_.is_null()) {
+      using map_type = Tpetra::Map<local_ordinal_type, global_ordinal_type, node_type>;
+      using import_type = Tpetra::Import<local_ordinal_type, global_ordinal_type, node_type>;
+      auto srcMap = rcp(new map_type(yBlock.getPointMap()));
+      auto tgtMap = rcp(new map_type(yBlockCol->getPointMap()));
+      pointImporter_ = rcp(new import_type(srcMap, tgtMap));
+    }
     performImport = true;
   }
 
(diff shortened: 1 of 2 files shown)


mhoemmen:
@trilinos/ifpack2

packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
```diff
@@ -1837,15 +1837,15 @@ ApplyInverseGS_BlockCrsMatrix (const block_crs_matrix_type& A,
     yBlockCol->putScalar (STS::zero ());
   }
   else if (performImport) {
-    yBlockCol->doImport (yBlock, *Importer_, Tpetra::INSERT);
+    yBlockCol->getMultiVectorView().doImport (yBlock.getMultiVectorView(), *Importer_, Tpetra::INSERT);
```
mhoemmen: The Import object needs to be a point Import, not a block Import.

mhoemmen: You can get the point Map of a BlockMultiVector by calling `getPointMap()` on it.  It returns a `Map` instance.  (I know, it's not an RCP.)  I would suggest keeping a `pointImporter_` object around in the class.  Try this:
```cpp
if (pointImporter_.is_null()) {
  auto srcMap = rcp(new map_type(src.getPointMap()));
  auto tgtMap = rcp(new map_type(tgt.getPointMap()));
  pointImporter_ = rcp(new import_type(srcMap, tgtMap));
}
```

mhoemmen:
Superseded by #6756 .  Thanks Victor! :-D

PR #6724, opened 2020-02-03
ATDM: Utilize new jsrun wrapper (CDOFA-94)

e10harvey:
This PR updates the test command invocation to use a new jsrun wrapper. 
## Summary of changes:
- mpiexec specified arguments: `-map-bs` `sockets:PE=4` was replaced with jsrun specific args: `--rs_per_socket 4`. 
- The ATDM MPI EXEC command has been specified to be the jsrun wrapper.
- The jsrun wrapper looks at the command line arguments and invokes jsrun with the correct options (-gpu, -disable_gpu_hooks, LD_PRELOAD=/path/to/libpami.so).
- Since jsrun must be invoked from a compute node, atdm_run_script_on_compute node has been changed to acquire a bsub allocation.
- To work around a JSM Remote server error that crops up intermittently, the jsrun wrapper runs tests via a while loop that will retry the test if the JSM Remote server error is encountered.
- The ats2 local-driver.sh now runs configure, build, and test on a compute node. And it runs the same set of tests (and binaries) with both `TPETRA_ASSUME_CUDA_AWARE_MPI=0` and `TPETRA_ASSUME_CUDA_AWARE_MPI=1`
- The ATDM README was updated to include an ats2 section.

## Testing:
Done via jenkins, see: [cdash](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2020-02-12&filtercount=4&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=-atdm-ats2&field2=groupname&compare2=61&value2=Experimental&field3=site&compare3=61&value3=vortex60&field4=status&compare4=61&value4=failed).

Related to #6691.


cmake/std/atdm/load-env.sh
```diff
@@ -22,11 +22,13 @@ function atdm_config_get_abs_dir_path() {
 export atdm_config_get_abs_dir_path
 
 # Get the base dir for the sourced script
-ATDM_SCRIPT_DIR=`echo $BASH_SOURCE | sed "s/\(.*\)\/.*\.sh/\1/g"`
-#echo "ATDM_SCRIPT_DIR = '$ATDM_SCRIPT_DIR'"
+ATDM_SCRIPT_DIR=$(readlink -f \
+                $(echo $BASH_SOURCE | sed "s/\(.*\)\/.*\.sh/\1/g"))
+echo "ATDM_SCRIPT_DIR = '$ATDM_SCRIPT_DIR'"
```
bartlettroscoe: Let's remove this `echo` in the final version.  It is fine for debugging.

cmake/std/atdm/ats2/environment.sh
```diff
@@ -253,7 +257,9 @@ export F90=mpifort
 
 # Define function atdm_run_script_on_compute_node
 unset atdm_run_script_on_compute_node
-source $ATDM_SCRIPT_DIR/common/define_atdm_run_script_on_local_node.sh
+
+# TODO: Will this work be able to find the cuda runtime for targeting GPUs?
+source $ATDM_SCRIPT_DIR/common/define_run_on_lsf_compute_node_func.sh
```
bartlettroscoe: Should this use `ATDM_CONFIG_SCRIPT_DIR` instead?

cmake/std/atdm/ats2/environment.sh
```diff
@@ -242,7 +242,11 @@ export MPICC=`which mpicc`
 export MPICXX=`which mpicxx`
 export MPIF90=`which mpif90`
 
-export ATDM_CONFIG_MPI_POST_FLAGS="-disable_gpu_hooks;-map-by;socket:PE=4"
+export ATDM_CONFIG_MPI_EXEC=${ATDM_SCRIPT_DIR}/ats2/trilinos_jsrun
+
+# TODO: Need -np support from the trilinos_jsrun wrapper.
+export ATDM_CONFIG_MPI_POST_FLAGS="--rs_per_socket 4"
+export ATDM_CONFIG_MPI_EXEC_NUMPROCS_FLAG="-p"
```
bartlettroscoe: To be safe, we should add `unset ATDM_CONFIG_MPI_EXEC_NUMPROCS_FLAG` to:

* Trilinos/cmake/std/atdm/utils/unset_atdm_config_vars_environment.sh

cmake/std/atdm/ats2/environment.sh
```diff
@@ -242,7 +242,11 @@ export MPICC=`which mpicc`
 export MPICXX=`which mpicxx`
 export MPIF90=`which mpif90`
 
-export ATDM_CONFIG_MPI_POST_FLAGS="-disable_gpu_hooks;-map-by;socket:PE=4"
+export ATDM_CONFIG_MPI_EXEC=${ATDM_SCRIPT_DIR}/ats2/trilinos_jsrun
```
bartlettroscoe: Use `ATDM_CONFIG_SCRIPT_DIR`?

cmake/std/atdm/ats2/environment.sh
e10harvey: It could but other environments are experiencing the same "relative path issue" so it seems to make more sense to just make `ATDM_SCRIPT_DIR` an absolute path as well, since, AFAIK, this does not change the semantics of ATDM_SCRIPT_DIR.

cmake/std/atdm/ats2/environment.sh
e10harvey: Same as above. What are the semantics of `ATDM_SCRIPT_DIR` and `ATDM_CONFIG_SCRIPT_DIR`?

cmake/std/atdm/ats2/environment.sh
e10harvey: @jjellio: Note these configuration options for MPI (or jsrun in this case). I've tested the changes introduced in this PR using your `trilinos_jsrun` wrapper. Based on the results of those tests, it looks like we're still missing the logic for handling GPU-based tests in the wrapper.

jjellio:
I believe the broader issue is that
```
export ATDM_CONFIG_MPI_POST_FLAGS="--rs_per_socket 4"
export ATDM_CONFIG_MPI_EXEC_NUMPROCS_FLAG="-p"
```

Is not enabling GPUs.

I'd change to:
```
export ATDM_CONFIG_MPI_POST_FLAGS="-a1;-r4;-c4;-g1;-brs"
```

Unpacked, that is saying:
4 resource sets
1 process per resource set
1 GPU per resource set
4 cores per resource set
bind things to the resource set


Once we can vet that tests work in an non-oversubscribed setting, then I would proceed to optimize the bindings to allow multiple processes to share a GPU.

e10harvey:
> I believe the broader issue is that
> 
> ```
> export ATDM_CONFIG_MPI_POST_FLAGS="--rs_per_socket 4"
> export ATDM_CONFIG_MPI_EXEC_NUMPROCS_FLAG="-p"
> ```
> 
> Is not enabling GPUs.
> 
> I'd change to:
> 
> ```
> export ATDM_CONFIG_MPI_POST_FLAGS="-a1;-r4;-c4;-g1;-brs"
> ```

@jjellio: This gives:
```bash
<snip>
TPETRA_ASSUME_CUDA_AWARE= BEFORE: jsrun -a1 -r4 -c4 -g1 -brs /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe
TPETRA_ASSUME_CUDA_AWARE= AFTER : jsrun -a1 -r4 -c4 -g1 -brs /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe
  what():  Kokkos::Cuda::initialize(1) FAILED : Device identifier out of range [0..1]

[vortex7:52028] *** Process received signal ***
[vortex7:52028] Signal: Aborted (6)
[vortex7:52028] Signal code:  (-6)
  what():  Kokkos::Cuda::initialize(2) FAILED : Device identifier out of range [0..1]

[vortex7:52029] *** Process received signal ***
[vortex7:52029] Signal: Aborted (6)
[vortex7:52029] Signal code:  (-6)
<snip>
```

instead of:
```bash
TPETRA_ASSUME_CUDA_AWARE= BEFORE: jsrun -p 4 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe
TPETRA_ASSUME_CUDA_AWARE= AFTER : jsrun -p 4 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe

Creating the sparse matrix
[vortex7:53346] *** Process received signal ***
[vortex7:53346] Signal: Segmentation fault (11)
[vortex7:53346] Signal code: Address not mapped (1)
[vortex7:53346] Failing at address: 0x200094937a80
[vortex7:53349] *** Process received signal ***
[vortex7:53349] Signal: Segmentation fault (11)
[vortex7:53349] Signal code: Address not mapped (1)
[vortex7:53349] Failing at address: 0x200094937a80
[vortex7:53347] *** Process received signal ***
[vortex7:53347] Signal: Segmentation fault (11)
[vortex7:53347] Signal code: Address not mapped (1)
[vortex7:53347] Failing at address: 0x200094937a88
[vortex7:53348] *** Process received signal ***
[vortex7:53348] Signal: Segmentation fault (11)
[vortex7:53348] Signal code: Address not mapped (1)
[vortex7:53348] Failing at address: 0x200094937a88
```

jjellio:
Please post the verbose `ctest` output and the output from `env`
I suspect you have an ENV somewhere that is controlling '--kokkos-ndevices', on Vortex/Lassen, etc.. you  don't set that.

e10harvey:
> Please post the verbose `ctest` output and the output from `env`
> I suspect you have an ENV somewhere that is controlling '--kokkos-ndevices', on Vortex/Lassen, etc.. you don't set that.

I ran the tests manually with trilinos_jsrun, like this:
```bash
"/vscratch1/eharvey/TRILINOS.base/Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-p" "4" "--rs_per_socket 4" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"
"/vscratch1/eharvey/TRILINOS.base/Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-a1" "-r4" "-c4" "-g1" "-brs" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"
```

```bash
$ env | grep -i kokkos
ATDM_CONFIG_KOKKOS_ARCH=Power9,Volta70
ATDM_CONFIG_Kokkos_ENABLE_Cuda_Lambda=ON
ATDM_CONFIG_Kokkos_ENABLE_Serial=ON
ATDM_CONFIG_Kokkos_ENABLE_Cuda=ON
KOKKOS_NUM_DEVICES=4
ATDM_CONFIG_Kokkos_ENABLE_Deprecated_Code=OFF
$ unset KOKKOS_NUM_DEVICES
$ "/vscratch1/eharvey/TRILINOS.base/Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-a1" "-r4" "-c4" "-g1" "-brs" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"
TPETRA_ASSUME_CUDA_AWARE= BEFORE: jsrun -a1 -r4 -c4 -g1 -brs /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe
TPETRA_ASSUME_CUDA_AWARE= AFTER : jsrun -a1 -r4 -c4 -g1 -brs /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe

Creating the sparse matrix
[vortex7:56008] *** Process received signal ***
[vortex7:56008] Signal: Segmentation fault (11)
[vortex7:56008] Signal code: Address not mapped (1)
[vortex7:56008] Failing at address: 0x200094937a80
[vortex7:56011] *** Process received signal ***
[vortex7:56011] Signal: Segmentation fault (11)
[vortex7:56011] Signal code: Address not mapped (1)
[vortex7:56011] Failing at address: 0x200094937a80
[vortex7:56010] *** Process received signal ***
[vortex7:56010] Signal: Segmentation fault (11)
[vortex7:56010] Signal code: Address not mapped (1)
[vortex7:56010] Failing at address: 0x200094937a88
[vortex7:56009] *** Process received signal ***
[vortex7:56009] Signal: Segmentation fault (11)
[vortex7:56009] Signal code: Address not mapped (1)
[vortex7:56009] Failing at address: 0x200094537a88
```

e10harvey:
@jjellio:
```bash
$ jsrun -M -gpu -p 4 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe

Creating the sparse matrix
Iteration 0:
- lambda = 2.51742
- ||A*q - lambda*q||_2 = 1.13181
Iteration 10:
- lambda = 3.94716
- ||A*q - lambda*q||_2 = 0.140141
Iteration 20:
- lambda = 3.98359
- ||A*q - lambda*q||_2 = 0.0433598
Iteration 30:
- lambda = 3.98878
- ||A*q - lambda*q||_2 = 0.022649
Iteration 40:
- lambda = 3.9905
- ||A*q - lambda*q||_2 = 0.0147021
Iteration 50:
- lambda = 3.99131
- ||A*q - lambda*q||_2 = 0.0108221
Iteration 60:
- lambda = 3.99178
- ||A*q - lambda*q||_2 = 0.00872345
Converged after 60 iterations

Estimated max eigenvalue: 3.99178

Increasing magnitude of A(0,0), solving again
Iteration 0:
- lambda = 2.52015
- ||A*q - lambda*q||_2 = 3.71146
Iteration 10:
- lambda = 20.0556
- ||A*q - lambda*q||_2 = 3.15223e-06
Converged after 10 iterations

Estimated max eigenvalue: 20.0556
End Result: TEST PASSED
```

So, how does trilinos_jsrun know which tests to add `-M -gpu` to? Does it need to inspected the application binary?

jjellio:
The problem is your Trilinos is configured with `TPETRA_ASSUME_CUDA_AWARE=1`. 

You aren't setting the ENV `TPETRA_ASSUME_CUDA_AWARE`, so the wrapper isn't doing anything w.r.t. adding `-M -gpu`.

The wrapper depends on the ENV to know what your intent is.

I would modify the wrapper to add a check at the start like:

```
if [[ "$TPETRA_ASSUME_CUDA_AWARE" != "0" ]] || [[ "$TPETRA_ASSUME_CUDA_AWARE" != "1" ]]; then
  echo "WARNING, you have not set TPETRA_ASSUME_CUDA_AWARE=0 or 1, defaulting to TPETRA_ASSUME_CUDA_AWARE=0"
  export TPETRA_ASSUME_CUDA_AWARE=0
fi
```

Then, it will ensure that jsrun and trilinos are on the same page in regards to which code to use.

jjellio:
FYI, inspecting the binary won't save you. These are runtime options. The library (Tpetra) is setting the default at build time.

One thing we could do, is look at adding a check to Tpetra::initialize that would attempt to use a device buffer, if it causes a segfault, then change the default to host.  That type of check would be expensive and platform specific though. (We'd need to use a signal handler and trap the segfault, which could interfere with user code that also wants to trap the signal).   The bottom line is that there is no function in MPI that can report what buffers MPI supports, and the MPI standards body seems to enjoy ignoring this important issue.

e10harvey:
> The problem is your Trilinos is configured with `TPETRA_ASSUME_CUDA_AWARE=1`.
> 
> You aren't setting the ENV `TPETRA_ASSUME_CUDA_AWARE`, so the wrapper isn't doing anything w.r.t. adding `-M -gpu`.
> 
> The wrapper depends on the ENV to know what your intent is.

I tried setting it, but I still get the same errors:
```bash
$ export TPETRA_ASSUME_CUDA_AWARE=0; "/vscratch1/eharvey/TRILINOS.base/Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-p" "4" "--rs_per_socket 4" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"

Creating the sparse matrix
[vortex5:67546] *** Process received signal ***
[vortex5:67546] Signal: Segmentation fault (11)
[vortex5:67546] Signal code: Address not mapped (1)
[vortex5:67546] Failing at address: 0x200094937a88
[vortex5:67545] *** Process received signal ***
[vortex5:67545] Signal: Segmentation fault (11)
[vortex5:67545] Signal code: Address not mapped (1)
[vortex5:67545] Failing at address: 0x200094937a88
[vortex5:67544] *** Process received signal ***
[vortex5:67544] Signal: Segmentation fault (11)
[vortex5:67544] Signal code: Address not mapped (1)
[vortex5:67544] Failing at address: 0x200094937a80
[vortex5:67547] *** Process received signal ***
[vortex5:67547] Signal: Segmentation fault (11)
[vortex5:67547] Signal code: Address not mapped (1)
[vortex5:67547] Failing at address: 0x200094937a80
$ echo $TPETRA_ASSUME_CUDA_AWARE
0
$ "/vscratch1/eharvey/TRILINOS.base/Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-a1" "-r4" "-c4" "-g1" "-brs" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"
  what():  Kokkos::Cuda::initialize(1) FAILED : Device identifier out of range [0..1]

[vortex5:68043] *** Process received signal ***
[vortex5:68043] Signal: Aborted (6)
[vortex5:68043] Signal code:  (-6)
```

> I would modify the wrapper to add a check at the start like:

@jjellio: Would you please update the wrapper accordingly in #6718? It looks like the wrapper still needs to add '-M -gpu' as well -- can you take care of that for me as well?

jjellio:
Can you open up permissions on `/vscratch1/eharvey/TRILINOS.base` ? I have to run, but it would easier for me to run the binary and see the behavior.

mhoemmen:
@jjellio wrote:
> One thing we could do, is look at adding a check to `Tpetra::initialize` that would attempt to use a device buffer, if it causes a segfault, then change the default to host.

Most applications don't actually call `Tpetra::initialize` or use Tpetra's scope guard (alas).

This test could actually be done lazily, when users first invoke some Tpetra function.  I'm not keen on catching SIGSEGV, though.  Note that there's no promise that SIGSEGV be synchronous -- it might show up "sometime later."

e10harvey:
> Can you open up permissions on `/vscratch1/eharvey/TRILINOS.base` ? I have to run, but it would easier for me to run the binary and see the behavior.

@jjellio: No problem, see `/vscratch1/forJames`. You might need the latest tip of Trilinos:develop to get the env setup with `source Trilinos/cmake/std/atdm/load-env.sh cuda-debug`.

jjellio:
Mark, I do **not** want to put code that traps signals into Trilinos.

The problem is really ensuring that users understand that MPI now has variable behavior and semantics based on flags set. AND, that defaults are different by platform, so behavior can shift because the installer changed the defaults

jjellio:
I am seeing success manually running.  I think you are just bumping into the hell that is cuda-aware MPI.

I misspelled the Tpetra ENV in my conditional above, so it wasn't setting the variable if it not set.

On a different note, the wrapper isn't preserving quotes the way it should.. ugh

Below I run by setting the Tpetra ENV, and based on that `-M -gpu` is getting added if needed.
I manually loaded modules (I rarely use the Trilinos 'environment' stuff).

```
[jjellio@vortex59 eharvey]$ ml

Currently Loaded Modules:
  1) StdEnv       3) python/2.7.16   5) cmake/3.14.5             7) spectrum-mpi/rolling-release
  2) git/2.20.0   4) gcc/7.3.1       6) lapack/3.8.0-gcc-4.9.3   8) cuda/10.1.243
```

```
[jjellio@vortex59 eharvey]$ ECHO_CMD=1 TPETRA_ASSUME_CUDA_AWARE_MPI=0 CUDA_LAUNCH_BLOCKING=1 ~/src/Trilinos-jje/cmake/std/atdm/ats2/trilinos_jsrun -p 4 -r4 -c4 -g1 "$b/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"
TPETRA_ASSUME_CUDA_AWARE_MPI=0 BEFORE: jsrun -p 4 -r4 -c4 -g1 /vscratch1/forJames/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe
TPETRA_ASSUME_CUDA_AWARE_MPI=0 AFTER : jsrun -p 4 -r4 -c4 -g1 /vscratch1/forJames/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe

Creating the sparse matrix
Iteration 0:
- lambda = 2.51742
- ||A*q - lambda*q||_2 = 1.13181
Iteration 10:
- lambda = 3.94716
- ||A*q - lambda*q||_2 = 0.140141
Iteration 20:
- lambda = 3.98359
- ||A*q - lambda*q||_2 = 0.0433598
Iteration 30:
- lambda = 3.98878
- ||A*q - lambda*q||_2 = 0.022649
Iteration 40:
- lambda = 3.9905
- ||A*q - lambda*q||_2 = 0.0147021
Iteration 50:
- lambda = 3.99131
- ||A*q - lambda*q||_2 = 0.0108221
Iteration 60:
- lambda = 3.99178
- ||A*q - lambda*q||_2 = 0.00872345
Converged after 60 iterations

Estimated max eigenvalue: 3.99178

Increasing magnitude of A(0,0), solving again
Iteration 0:
- lambda = 2.52015
- ||A*q - lambda*q||_2 = 3.71146
Iteration 10:
- lambda = 20.0556
- ||A*q - lambda*q||_2 = 3.15223e-06
Converged after 10 iterations

Estimated max eigenvalue: 20.0556
End Result: TEST PASSED
```

Now, with the Tpetra ENV = 1 (enable cuda-aware MPI)
```
[jjellio@vortex59 eharvey]$ ECHO_CMD=1 TPETRA_ASSUME_CUDA_AWARE_MPI=1 CUDA_LAUNCH_BLOCKING=1 ~/src/Trilinos-jje/cmake/std/atdm/ats2/trilinos_jsrun -p 4 -r4 -c4 -g1 "$b/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"
TPETRA_ASSUME_CUDA_AWARE_MPI=1 BEFORE: jsrun -p 4 -r4 -c4 -g1 /vscratch1/forJames/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe
TPETRA_ASSUME_CUDA_AWARE_MPI=1 AFTER : jsrun -M -gpu -p 4 -r4 -c4 -g1 /vscratch1/forJames/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe

Creating the sparse matrix
Iteration 0:
- lambda = 2.51742
- ||A*q - lambda*q||_2 = 1.13181
Iteration 10:
- lambda = 3.94716
- ||A*q - lambda*q||_2 = 0.140141
Iteration 20:
- lambda = 3.98359
- ||A*q - lambda*q||_2 = 0.0433598
Iteration 30:
- lambda = 3.98878
- ||A*q - lambda*q||_2 = 0.022649
Iteration 40:
- lambda = 3.9905
- ||A*q - lambda*q||_2 = 0.0147021
Iteration 50:
- lambda = 3.99131
- ||A*q - lambda*q||_2 = 0.0108221
Iteration 60:
- lambda = 3.99178
- ||A*q - lambda*q||_2 = 0.00872345
Converged after 60 iterations

Estimated max eigenvalue: 3.99178

Increasing magnitude of A(0,0), solving again
Iteration 0:
- lambda = 2.52015
- ||A*q - lambda*q||_2 = 3.71146
Iteration 10:
- lambda = 20.0556
- ||A*q - lambda*q||_2 = 3.15223e-06
Converged after 10 iterations

Estimated max eigenvalue: 20.0556
End Result: TEST PASSED
```

bartlettroscoe:
@jjellio and @e10harvey, should we not merge this branch/PR and the branch/PR #6718 into the same branch/PR?  Why are they separate?

cmake/std/atdm/ats2/environment.sh
bartlettroscoe: @e10harvey, I think it makes sense to make `ATDM_CONFIG_SCRIPT_DIR` the abs path and then use it consistently throughout and get rid of `ATDM_SCRIPT_DIR` unless there is a reason to keep `ATDM_SCRIPT_DIR`.

cmake/std/atdm/ats2/environment.sh
bartlettroscoe: > What are the semantics of ATDM_SCRIPT_DIR and ATDM_CONFIG_SCRIPT_DIR?

Let's make ` ATDM_CONFIG_SCRIPT_DIR` an absolute path and let's get rid of `ATDM_SCRIPT_DIR`.

bartlettroscoe:
> The bottom line is that there is no function in MPI that can report what buffers MPI supports, and the MPI standards body seems to enjoy ignoring this important issue.

@jjellio, this is information that we provide on a system-by-system based in the ATDM Trilinos configuration process.  We know what systems we are one, what MPI implementation, version, compiler, version, etc. we are using.  We are not guessing.

e10harvey:
> @jjellio and @e10harvey, should we not merge this branch/PR and the branch/PR #6718 into the same branch/PR? Why are they separate?

You're right. We should combine them. My mistake - I asked @jjellio to open a separate PR for the wrapper.

e10harvey:
@jjellio: Excellent, thanks! I am also seeing success running manually with the updated wrapper:
```bash
[eharvey@vortex59 Trilinos (add_jsrun_wrapper)]$ export KOKKOS_NUM_DEVICES=4
[eharvey@vortex59 Trilinos (add_jsrun_wrapper)]$ ECHO_CMD=1 TPETRA_ASSUME_CUDA_AWARE_MPI=1 "/vscratch1/eharvey/TRILINOS.base/Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-p" "4" "--rs_per_socket 4" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"
TPETRA_ASSUME_CUDA_AWARE=1 BEFORE: jsrun -p 4 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe
TPETRA_ASSUME_CUDA_AWARE=1 AFTER : jsrun -M -gpu -p 4 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe

Creating the sparse matrix
Iteration 0:
- lambda = 2.51742
- ||A*q - lambda*q||_2 = 1.13181
Iteration 10:
- lambda = 3.94716
- ||A*q - lambda*q||_2 = 0.140141
Iteration 20:
- lambda = 3.98359
- ||A*q - lambda*q||_2 = 0.0433598
Iteration 30:
- lambda = 3.98878
- ||A*q - lambda*q||_2 = 0.022649
Iteration 40:
- lambda = 3.9905
- ||A*q - lambda*q||_2 = 0.0147021
Iteration 50:
- lambda = 3.99131
- ||A*q - lambda*q||_2 = 0.0108221
Iteration 60:
- lambda = 3.99178
- ||A*q - lambda*q||_2 = 0.00872345
Converged after 60 iterations

Estimated max eigenvalue: 3.99178

Increasing magnitude of A(0,0), solving again
Iteration 0:
- lambda = 2.52015
- ||A*q - lambda*q||_2 = 3.71146
Iteration 10:
- lambda = 20.0556
- ||A*q - lambda*q||_2 = 3.15223e-06
Converged after 10 iterations

Estimated max eigenvalue: 20.0556
End Result: TEST PASSED
[eharvey@vortex59 Trilinos (add_jsrun_wrapper)]$ ECHO_CMD=1 TPETRA_ASSUME_CUDA_AWARE_MPI=0 "/vscratch1/eharvey/TRILINOS.base/Trilinos/cmake/std/atdm/ats2/trilinos_jsrun" "-p" "4" "--rs_per_socket 4" "/vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe"
TPETRA_ASSUME_CUDA_AWARE=0 BEFORE: jsrun -p 4 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe
TPETRA_ASSUME_CUDA_AWARE=0 AFTER : jsrun -p 4 --rs_per_socket 4 /vscratch1/eharvey/TRILINOS.base/ATS2/Trilinos-atdm-ats2-cuda-10.1.243-gcc-7.3.1-spmpi-2019.06.24_static_dbg/packages/tpetra/core/guide/src/Examples/SourceCode/TpetraCore_guide_power_method_1.exe

Creating the sparse matrix
Iteration 0:
- lambda = 2.51742
- ||A*q - lambda*q||_2 = 1.13181
Iteration 10:
- lambda = 3.94716
- ||A*q - lambda*q||_2 = 0.140141
Iteration 20:
- lambda = 3.98359
- ||A*q - lambda*q||_2 = 0.0433598
Iteration 30:
- lambda = 3.98878
- ||A*q - lambda*q||_2 = 0.022649
Iteration 40:
- lambda = 3.9905
- ||A*q - lambda*q||_2 = 0.0147021
Iteration 50:
- lambda = 3.99131
- ||A*q - lambda*q||_2 = 0.0108221
Iteration 60:
- lambda = 3.99178
- ||A*q - lambda*q||_2 = 0.00872345
Converged after 60 iterations

Estimated max eigenvalue: 3.99178

Increasing magnitude of A(0,0), solving again
Iteration 0:
- lambda = 2.52015
- ||A*q - lambda*q||_2 = 3.71146
Iteration 10:
- lambda = 20.0556
- ||A*q - lambda*q||_2 = 3.15223e-06
Converged after 10 iterations

Estimated max eigenvalue: 20.0556
End Result: TEST PASSED
```

I will cherry pick the commits from #6718 into this PR.

cmake/std/atdm/ats2/environment.sh
e10harvey: Sounds good. See #6754.

cmake/std/atdm/utils/unset_atdm_config_vars_environment.sh
```diff
@@ -54,3 +54,4 @@ unset ATDM_CONFIG_COMPLETED_ENV_SETUP
 unset ATDM_CONFIG_Kokkos_ENABLE_Serial
 unset ATDM_CONFIG_Kokkos_ENABLE_Cuda
 unset ATDM_CONFIG_Kokkos_ENABLE_Deprecated_Code
+unset ATDM_CONFIG_MPI_EXEC_NUMPROCS_FLAG
```
bartlettroscoe: Can you please add a newline here?

cmake/ctest/drivers/atdm/ats2/drivers/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt.sh
```diff
@@ -2,4 +2,11 @@
 if [ "${Trilinos_TRACK}" == "" ] ; then
   export Trilinos_TRACK=Experimental
 fi
+export TPETRA_ASSUME_CUDA_AWARE_MPI=0
 $WORKSPACE/Trilinos/cmake/ctest/drivers/atdm/ats2/local-driver.sh
+
+export TPETRA_ASSUME_CUDA_AWARE_MPI=1
+#TODO: Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_cuda-aware_static_dbg
```
e10harvey: @bartlettroscoe: DO you know how to change the cdash "Build Name" here while still reusing the contents of the workspace?

bartlettroscoe: > DO you know how to change the cdash "Build Name" here while still reusing the contents of the workspace?

Yes, as described in:
 
* https://tribits.org/doc/TribitsDevelopersGuide.html#tribits-ctest-driver

I think before the second call you want to set something like:

```
export CTEST_BUILD_NAME=${JOB_NAME}-cuda-aware
export CTEST_DO_UPDATE=FALSE
export CTEST_DO_CONFIGURE=FALSE
export CTEST_DO_BUILD=FALSE
```

cmake/ctest/drivers/atdm/ats2/drivers/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt.sh
```diff
@@ -2,4 +2,11 @@
 if [ "${Trilinos_TRACK}" == "" ] ; then
   export Trilinos_TRACK=Experimental
 fi
+export TPETRA_ASSUME_CUDA_AWARE_MPI=0
 $WORKSPACE/Trilinos/cmake/ctest/drivers/atdm/ats2/local-driver.sh
```
bartlettroscoe: Should we not also be configuring, building, and running on the compute node as well, even for the non-CUDA aware MPI mode?  That way we can use the entire node to build and run.

e10harvey: Right, I've done this on a per-project basis from jenkins -- checkout the "Execute Shell" section for the nightly ATS2 jobs.

jjellio:
> > The bottom line is that there is no function in MPI that can report what buffers MPI supports, and the MPI standards body seems to enjoy ignoring this important issue.
> 
> @jjellio, this is information that we provide on a system-by-system based in the ATDM Trilinos configuration process. We know what systems we are one, what MPI implementation, version, compiler, version, etc. we are using. We are not guessing.

Ross, my gripe is that developers cannot use the MPI Standard functions as a means to determine what MPI supports.  We could avoid a lot of this if MPI provided a set of functions to query what type of memory MPI can address. e.g., `MPI_Supported_Memory( MPI_Memory_type * supported_memory_types, int * count)`, and if the standard body would then define some common annotations, "MPI_DEVICE_VIRTUAL_MEMORY, MPI_HOST_VIRTUAL_MEMORY, MPI_HOST_MEMORY, MPI_DEVICE_MEMORY, etc..."  Then dudes and dudettes like Tpetra could program against MPI to figure out what is supported.

cmake/ctest/drivers/atdm/ats2/drivers/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt.sh
jjellio: I am curious, by setting these, and not doing a configure, how will the CTEST_BUILD_NAME propagate into the Ctest file? (Or is that ENV being used to override whatever Ctest finds in that file when it posts to a dashboard -- I don't know how the dashboard posting stuff works)

If this works as I think it is intended, it sounds like we aren't far from seeing this working.

bartlettroscoe: > I am curious, by setting these, and not doing a configure, how will the CTEST_BUILD_NAME propagate into the Ctest file? 

ctest -S is an independent program and has its own variables.

bartlettroscoe:
> Ross, my gripe is that developers cannot use the MPI Standard functions as a means to determine what MPI supports

That is an argument to make to the MPI standards committee and is something we can't solve here.

cmake/ctest/drivers/atdm/ats2/drivers/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt.sh
```diff
@@ -2,4 +2,16 @@
...
+
+export TPETRA_ASSUME_CUDA_AWARE_MPI=1
+export CTEST_BUILD_NAME=Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_cuda-aware_static_dbg
```
bartlettroscoe: Why is this hard coded?

cmake/ctest/drivers/atdm/ats2/drivers/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt.sh
```diff
@@ -2,4 +2,16 @@
...
+export CTEST_DO_UPDATES=FALSE
+#export CTEST_DO_NEW_START=FALSE
+export CTEST_START_WITH_EMPTY_BINARY_DIRECTORY=FALSE
+export CTEST_DO_CONFIGURE=FALSE
+export CTEST_DO_BUILD=FALSE
+atdm_run_script_on_compute_node \
```
bartlettroscoe: NOTE: The second call for CUDA-aware MPI makes no sense if you are not doing a CUDA build.  We need a way to handle that.  Therefore, we will need a way to handle that.  One way would be to source `local-driver.sh` instead of call it and with some other refactoring we have access to the `ATDM_CONFIG_` env vars generated from the parsed build name and then we can use this to set BUILD_NAME and decide if we want to run the second call.   Should not be too hard to do

cmake/ctest/drivers/atdm/ats2/drivers/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt.sh
e10harvey: For testing.

cmake/ctest/drivers/atdm/ats2/drivers/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt.sh
e10harvey: This driver is specifically for cuda builds, right?

bartlettroscoe: > This driver is specifically for cuda builds, right?

Got it.  Thanks for pointing that out.

I think we can refactor to move logic like this into Trilinos/cmake/ctest/drivers/atdm/ats2/local-driver.sh for all CUDA build to automatically run non-CUDA-aware and CUDA-aware MPI test suite.  But we can do that after we get this one build to work correctly.

Thanks!

e10harvey: No problem! Agreed :)

jjellio: It's worth pointing out:

It would be nice if we wanted to setup testing on different platform, then all we needed to do is provide a suitable launch wrapper. (On Ride/White/Waterman, such a wrapper would would disable cuda-aware rather than enable)

IMO, that would be a good goal to keep in mind.

If the logic can avoid 'cuda' specific stuff, that would be a good long term strategy. (i.e., there will probably be other accelerators out there).  Of course, by then, the vendors may have eliminated the concept of 'device-aware' and 'regular' MPI (hopefully they will provide options to turn things off, and then we could still do these types of tests)

bartlettroscoe:
@e10harvey, there is a merge conflict in:

* cmake/std/atdm/utils/unset_atdm_config_vars_environment.sh

that will need to be resolved before this can be merged.  I would recommend merging in 'develop' and then carefully documenting how the conflicts were resolved.

e10harvey:
> I would recommend merging in 'develop' and then carefully documenting how the conflicts were resolved.

Should this be documented in the commit message?

bartlettroscoe:
> Should this be documented in the commit message?

Yes.  Otherwise, it is very hard to determine how such conflicts got resolved (and otherwise it is very hard to see by just running `git diff -m -1` on that merge commit).

e10harvey:
> Yes. Otherwise, it is very hard to determine how such conflicts got resolved (and otherwise it is very hard to see by just running `git diff -m -1` on that merge commit).

OK. I see that the diff is not clear now! Thanks!

bartlettroscoe:
@e10harvey, here is the CDash query characterizing all of the failures we saw for the CUDA aware and non-CUDA aware builds you ran on 2020-02-07:

* https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2020-02-07&filtercount=8&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-ats2&field2=groupname&compare2=61&value2=Experimental&field3=site&compare3=61&value3=vortex60&field4=status&compare4=61&value4=failed&field5=testoutput&compare5=94&value5=CUDA%20Hook%20Library%3A%20Failed%20to%20find%20symbol%20mem_find_dreg_entries&field6=testoutput&compare6=94&value6=cat%3A%20cmd.out%3A%20No%20such%20file%20or%20directory&field7=testoutput&compare7=94&value7=rm%3A%20cannot%20remove%20%E2%80%98cmd.out%E2%80%99%3A%20No%20such%20file%20or%20directory&field8=testoutput&compare8=94&value8=Remote%20JSM%20server%20is%20not%20responding%20on%20host

There were 216 tests that failed.  That included a bunch of tests that failed showing no output at all.  If you filter out the tests that showed no output in addition with:

* https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2020-02-07&filtercount=9&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-ats2&field2=groupname&compare2=61&value2=Experimental&field3=site&compare3=61&value3=vortex60&field4=status&compare4=61&value4=failed&field5=testoutput&compare5=94&value5=CUDA%20Hook%20Library%3A%20Failed%20to%20find%20symbol%20mem_find_dreg_entries&field6=testoutput&compare6=94&value6=cat%3A%20cmd.out%3A%20No%20such%20file%20or%20directory&field7=testoutput&compare7=94&value7=rm%3A%20cannot%20remove%20%E2%80%98cmd.out%E2%80%99%3A%20No%20such%20file%20or%20directory&field8=testoutput&compare8=94&value8=Remote%20JSM%20server%20is%20not%20responding%20on%20host&field9=testoutput&compare9=97&value9=%5Ba-z%5D%2B

you get just 32 failing tests.  Many of those 32 failing tests we already know about and have dept with.

The test failures that we need to deal with contain:

* `CUDA Hook Library: Failed to find symbol mem_find_dreg_entries`
* `cat: cmd.out: No such file or directory`
* `rm: cannot remove ‘cmd.out’: No such file or directory`
* `Remote JSM server is not responding on host`

And otherwise, need to figure out why so many tests are failing with no output when we are expecting output and are getting zero 0 return status.

jjellio:
I am not familiar with `CUDA Hook Library: Failed to find symbol mem_find_dreg_entries`.

If we can provide a full run line and instructions to replicate, then I can forward this to the ATS-2 LLNL admins and ask for guidance to work around.

It has been a while since I originally did the `disable_cuda_hooks` work... it could be they've added more requirements to disabling it.

I'm a little concerned that to get proper GPU MPI support, we need to have `-g ##` on the JSRUN line (i.e., request a GPU in the allocation).  The run line used is just asking for `"-p" "4" "--rs_per_socket" "4"`.

For the issue above, 
`https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2020-02-07&filtercount=5&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-ats2&field2=groupname&compare2=61&value2=Experimental&field3=site&compare3=61&value3=vortex60&field4=status&compare4=61&value4=failed&field5=testoutput&compare5=95&value5=CUDA%20Hook%20Library%3A%20Failed%20to%20find%20symbol%20mem_find_dreg_entries`

This should be a cuda-aware run, so `-M -gpu` should be set, and since `-p` is `4`, it should not be disabling the Cuda hooks.

Would it be possible to have the Trilinos-jsrun wrapper echo the commands (e.g., set ENV `ECHO_CMD=1`) ?

If we bug this with LLNL, we will need to provide the real JSRUN line.

bartlettroscoe:
> I am not familiar with `CUDA Hook Library: Failed to find symbol mem_find_dreg_entries.`

Looks like @e10harvey got that fixed in 4665ced9c9104bea6fa3ea3f91f5c35a3519d6e3

We will get a fresh set of experimental test results and assess were we are.

jjellio:
FYI, the issue here could be unit tests that are making Memory allocations before calling MPI_Init.

For the `MueLu_ImportPerformance_Tpetra_4`, that code calls:
```

#ifdef HAVE_MUELU_KOKKOSCORE
    // jje: KokkosScope takes argc/argv by reference and I believe can returned modified (i.e., allocated memory)
    Kokkos::ScopeGuard KokkosScope(argc,argv);
#endif
   // jje MPI_Init is going to be called in this function 
    auto val =  Automatic_Test_ETI(argc,argv);
    return val;
}
```

I haven't vetted my hypothesis (and won't be able to in a short time frame).  If this pattern of calling KokkosScope before MPI_Init is the problem, we will need to write down guidence on how to Init Kokkos and MPI. (i.e., which one first).

@mhoemmen  @csiefer2 

I've CC'd Mark and Chris, they may be able to quickly help with this (I am out of town till Thursday)

bartlettroscoe:
Latest experimental runs on this branch shown here:

* https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&begin=2020-02-10&end=2020-02-11&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-ats2&field2=site&compare2=61&value2=vortex60

which were:

* [Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp](https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&parentid=5227674) (passed=1418, failed=764)
* [Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt_cuda-aware-mpi](https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&parentid=5228199) (passed=2175, failed=7)

looks promising for the CUDA-aware MPI tests.  The failing tests for that build:


Subproject | Name | Status | Time | Proc Time | Details | Labels | Summary | Processors
-- | -- | -- | -- | -- | -- | -- | -- | --
Kokkos | KokkosCore_UnitTest_PushFinalizeHook_terminate | Failed | 1s 60ms | 1s 60ms | Completed (Failed) | Kokkos | Unstable | 1
MueLu | MueLu_ParameterListInterpreterTpetra_MPI_1 | Failed | 40s 820ms | 40s 820ms | Completed (Failed) | MueLu | Unstable | 1
Teuchos | TeuchosCore_testTeuchosTestForTermination | Failed | 1m 1s 110ms | 1m 1s 110ms | Completed (Failed) | Teuchos | Unstable | 1
Teuchos | TeuchosCore_testTeuchosTestForTermination_0_MPI_4 | Failed | 1m 1s 350ms | 4m 5s 400ms | Completed (Failed) | Teuchos | Unstable | 4
Teuchos | TeuchosCore_testTeuchosTestForTermination_1_MPI_4 | Failed | 1m 1s 380ms | 4m 5s 520ms | Completed (Failed) | Teuchos | Unstable | 4
Teuchos | TeuchosCore_testTeuchosTestForTermination_2_MPI_4 | Failed | 1m 1s 370ms | 4m 5s 480ms | Completed (Failed) | Teuchos | Unstable | 4
Teuchos | TeuchosCore_testTeuchosTestForTermination_3_MPI_4 | Failed | 1m 1s 370ms | 4m 5s 480ms | Completed (Failed) | Teuchos | Unstable | 4

are all known issues except for `KokkosCore_UnitTest_PushFinalizeHook_terminate`.

For the 764 test failures for the build `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp` shown here:

* https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-02-10&end=2020-02-11&filtercount=4&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp&field2=site&compare2=65&value2=vortex60&field3=groupname&compare3=61&value3=Experimental&field4=status&compare4=61&value4=failed

if you exclude the tests showing:

* `Remote JSM server is not responding on host vortex`

in the query:

* https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-02-10&end=2020-02-11&filtercount=5&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp&field2=site&compare2=65&value2=vortex60&field3=groupname&compare3=61&value3=Experimental&field4=status&compare4=61&value4=failed&field5=testoutput&compare5=94&value5=Remote%20JSM%20server%20is%20not%20responding%20on%20host%20vortex

you get just 7 failing tests:

Site | Build Name | Test Name | Status | Time | Proc Time | Details | Build Time | Processors
-- | -- | -- | -- | -- | -- | -- | -- | --
vortex60 | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp | KokkosCore_UnitTest_Cuda_MPI_1 | Failed | 1m 9s 760ms | 1m 9s 760ms | Completed (Failed) | 2020-02-10T16:29:41 MST | 1
vortex60 | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp | KokkosCore_UnitTest_PushFinalizeHook_terminate | Failed | 1s 640ms | 1s 640ms | Completed (Failed) | 2020-02-10T16:29:41 MST | 1
vortex60 | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp | MueLu_ParameterListInterpreterTpetra_MPI_1 | Failed | 45s 330ms | 45s 330ms | Completed (Failed) | 2020-02-10T16:29:41 MST | 1
vortex60 | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp | ROL_example_dense-hessian_DenseHessian_MPI_1 | Failed | 740ms | 740ms | Completed (Failed) | 2020-02-10T16:29:41 MST | 1
vortex60 | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp | ROL_test_function_StdObjectiveCheck_MPI_1 | Failed | 370ms | 370ms | Completed (Failed) | 2020-02-10T16:29:41 MST | 1
vortex60 | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp | TeuchosCore_testTeuchosTestForTermination | Failed | 1m 1s 120ms | 1m 1s 120ms | Completed (Failed) | 2020-02-10T16:29:41 MST | 1
vortex60 | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp | TeuchosCore_TwoDArrayTests_MPI_1 | Failed | 660ms | 660ms | Completed (Failed) | 2020-02-10T16:29:41 MST | 1

So why does the non-CUDA-aware build not show any `Remote JSM server is not responding on host` failures but the CUDA-aware build does?  They are both using `jsrun` under the `trilinos_jsrun` wrapper script.

e10harvey:
Thank you for look at this and summarizing the results, @bartlettroscoe!
> So why does the non-CUDA-aware build not show any `Remote JSM server is not responding on host` failures but the CUDA-aware build does? They are both using `jsrun` under the `trilinos_jsrun` wrapper script.

@bartlettroscoe: AFAIK, this happens intermittently and it doesn't have to do with our CUDA-aware build. IIRC, it may have to do with the bsub allocation. I will update the retry loop to wait for a few seconds before retrying if `Remote JSM server is not responding on host` is encountered. There are folks working on this jsrun error and I've updated the ticket I have with them to let them know we're still experiencing this issue. @jjellio, do you know more about the status of the `Remote JSM server is not responding on host` issue?

Thanks @bartlettroscoe for confirming that
>     * `CUDA Hook Library: Failed to find symbol mem_find_dreg_entries`
> 
>     * `cat: cmd.out: No such file or directory`
> 
>     * `rm: cannot remove ‘cmd.out’: No such file or directory`
have been resolved by 4665ced

I will look into the failures you've summarized [above](https://github.com/trilinos/Trilinos/pull/6724#issuecomment-584731008)!

cmake/std/atdm/ats2/trilinos_jsrun
```diff
@@ -0,0 +1,158 @@
...
+  if [[ ! "$JSRUN_WRAPPER_NUM_RETRIES" == "" ]]; then
+    retry=${JSRUN_WRAPPER_NUM_RETRIES}
+  else
+    retry=3
+  fi
+  grep_status=0
+
+  # Retry jsrun command until retry is exhausted or JSM server error is not found
+  while [[ $grep_status -eq 0 && $retry -ne 0 ]]; do
+    out_file=$(${ATDM_CONFIG_SCRIPT_DIR}/ats2/eval_wrapper jsrun $(printf "'%s' " "${args[@]}"))
```
e10harvey: > Is this going to preserve quoted arguments?
>
> E.g.,jsrun ... bash -c 'echo $HOSTNAME'
>
> You probably need to use:
> eval $(printf "'%s' " "$@") &> $out_file

@jjellio: We've taken care of that here.

bartlettroscoe:
@e10harvey, from Kevin Buterbaugh just now:

> In the meantime, the only workaround is to resubmit any jobs which fail until they ultimately succeed.

So it sounds like we do want to do the re-tries to see if we can get around the `Remote JSM server is not responding on host` failures.

cmake/ctest/drivers/atdm/ats2/local-driver.sh
```diff
@@ -21,12 +21,12 @@ atdm_run_script_on_compute_node \
   ${LSF_CTEST_TIMEOUT}
 
 if [ "${Trilinos_CTEST_RUN_CUDA_AWARE_MPI}" == "1" ]; then
+  # Wait a little while before attempting another bsub allocation
+  # to try working around "'Error: Remote JSM server is not responding'"
+  sleep 5
   export TPETRA_ASSUME_CUDA_AWARE_MPI=1
   export CTEST_BUILD_NAME=${ATDM_CONFIG_BUILD_NAME}_cuda-aware-mpi
-  export CTEST_DO_UPDATES=FALSE
```
bartlettroscoe: @e10harvey, I think you want to avoid the updates as that will be confusing for CDash.  It is okay to do the configure and build though.

e10harvey: This export is redundant due to https://github.com/trilinos/Trilinos/pull/6724/files#diff-a95886e3c981a2798804e59c33701ac4L18.

bartlettroscoe: Right, that is correct.  That is calling the "test" driver script.  Thanks for pointing that out.

cmake/ctest/drivers/atdm/ats2/local-driver.sh
```diff
@@ -19,3 +19,16 @@ atdm_run_script_on_compute_node \
...
+  # Wait a little while before attempting another bsub allocation
+  # to try working around "'Error: Remote JSM server is not responding'"
+  sleep 5
+  export TPETRA_ASSUME_CUDA_AWARE_MPI=1
+  export CTEST_BUILD_NAME=${ATDM_CONFIG_BUILD_NAME}_cuda-aware-mpi
+  export CTEST_START_WITH_EMPTY_BINARY_DIRECTORY=FALSE
+  atdm_run_script_on_compute_node \
+    $WORKSPACE/Trilinos/cmake/ctest/drivers/atdm/ctest-s-driver-test.sh \
+    $PWD/ctest-s-driver-test.out
+    ${LSF_CTEST_TIMEOUT}
```
bartlettroscoe: Wait, there is actually one thing wrong with calling the script ctest-s-driver-test.sh in this case.  The script ctest-s-driver-test.sh sets `CTEST_DO_NEW_START=OFF` which assumes you are following up running tests on a previous build but that is not really what you are doing here.  What you are really doing is running a new CDash build in the same build directory as another build.  But since the build name is different, this is showing up as a different build on CDash but this seems a bit fragile here and we may need to ask Zack G. at Kitware if this is good idea or not.

e10harvey: Good point. We should confirm this behavior with Kitware. Would you do that, @bartlettroscoe?

> but this seems a bit fragile here
Do you suspect this is related to the Remote JSM server errors in the CUDA-aware build?

bartlettroscoe: > Do you suspect this is related to the Remote JSM server errors in the CUDA-aware build?

No, I am only worried about how this shows up on CDash, not how things run locally.

e10harvey:
Here is an initial triage of the remaining failing tests identified by @bartlettroscoe:

### Known failures in both CUDA-aware and non-CUDA-aware tests:
>MueLu_ParameterListInterpreterTpetra_MPI_1


>TeuchosCore_testTeuchosTestForTermination


### Unknown failures in both CUDA-aware and non-CUDA-aware tests:
> KokkosCore_UnitTest_PushFinalizeHook_terminate

This test is failing in both non-CUDA-aware and CUDA-aware builds due to a process exiting with signal 6. Also, this test fails the same way when run with either `jsrun` ([cdash-link](https://testing-dev.sandia.gov/cdash/testDetails.php?test=83808735&build=5228204)) or `mpiexec` ([cdash-link](https://testing.sandia.gov/cdash/testDetails.php?test=101157721&build=6385458)). However, when run with `mpiexec`, the test is marked as passing -- see cdash links in previous sentence.


@bartlettroscoe: what's the next triaging step here?
### Unknown failures in CUDA-aware tests:
>KokkosCore_UnitTest_Cuda_MPI_1

From [cdash](https://testing-dev.sandia.gov/cdash/testDetails.php?test=83996559&build=5227679).
```
<snip>
[ RUN      ] cuda.debug_pin_um_to_host
Time CudaSpace: 0.052340 CudaUVMSpace_1: 0.052490 CudaUVMSpace_2: 0.049860 CudaPinnedHostSpace: 0.096996 CudaUVMSpace_Pinned: 0.052756
/vscratch1/jenkins/vortex-slave/workspace/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt-exp/SRC_AND_BUILD/Trilinos/packages/kokkos/core/unit_test/cuda/TestCuda_DebugPinUVMSpace.cpp:127: Failure
Value of: passed
  Actual: false
Expected: true
[  FAILED  ] cuda.debug_pin_um_to_host (378 ms)
<snip>
```

@bartlettroscoe: what's the next triaging step here?
>ROL_example_dense-hessian_DenseHessian_MPI_1

This passes when run manually. This could be another issue with the `*cmd.out` files given that the output appears to be truncated. I will submit a fix.
>ROL_test_function_StdObjectiveCheck_MPI_1

This passes when run manually. This could be another issue with the `*cmd.out` files given that the output appears to be truncated. I will submit a fix.
>TeuchosCore_TwoDArrayTests_MPI_1

This passes when run manually. This could be another issue with the `*cmd.out` files given that the output appears to be truncated. I will submit a fix.

bartlettroscoe:
> what's the next triaging step here?

@e10harvey, once we are sure we have a small number of consistently failing tests (e.g. O(10) or less), we can go ahead and merge this to 'develop' and get this new setup running nightly off of the 'develop' branch with updates of Trilinos.  For the set of failing tests you list [above](https://github.com/trilinos/Trilinos/pull/6724#issuecomment-585296512), if you look over the existing set of [ATDM Trilinos GitHub issues](https://github.com/trilinos/Trilinos/issues?utf8=%E2%9C%93&q=label%3A%22client%3A+ATDM%22+label%3A%22type%3A+bug%22+-label%3A%22Disabled+Tests%22+-label%3AStalled+is%3Aopen+) you will see that some of these are already known failures.  And the others we can create new ATDM Trilinos GitHub issues for and get the Trilinos developers working on those (once we are pretty sure they are not system issues).

If we can get down the bulk of the failing tests and address the CUDA-aware MPI and non-CUDA-aware MPI test suite (which I think we have now), then we can merge this to 'develop', get this running nightly (likely posted to the 'Specialized' CDash gorup if there are just a handful of failing tests) and then we can triage the remaining failures from there using the process described [here](https://snl-wiki.sandia.gov/display/CoodinatedDevOpsATDM/Triaging+and+Addressing+ATDM+Trilinos+Failures?src=contextnavchildmode).

Once you think you have the bulk of the `Remote JSM server is not responding on host` errors addressed using the retry approach (seeing just  5 or less hopefully per run), can you please manually fire off another Jenkins-srn job and we can see what this looks like?

e10harvey:
> Once you think you have the bulk of the `Remote JSM server is not responding on host` errors addressed using the retry approach (seeing just 5 or less hopefully per run), can you please manually fire off another Jenkins-srn job and we can see what this looks like?

A new job has been started, as noted in CDOFA-94.

e10harvey:
The latest run just finished and can be seen at: https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2020-02-12&filtercount=4&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=-atdm-ats2&field2=groupname&compare2=61&value2=Experimental&field3=site&compare3=61&value3=vortex60&field4=status&compare4=61&value4=failed

e10harvey:
@bartlettroscoe: Thanks for pointing the return code issue (addressed in 3841ff6d7c74e1bf9a6ce77e16bec18cf71bb90b) out!

e10harvey:
@bartlettroscoe: Testing kokkos and Techos locally via ctest shows that the return code fix in 
3841ff6 is correct. I've updated the README as well, so, this PR is ready for review. Thanks to @bartlettroscoe, @jjellio, @mhoemmen, and anyone who I've missed for all your help!!

bartlettroscoe:
@e10harvey, there is a small risk that fixing the return code return from `trilinos_jsrun` may uncover more failing tests.  That is, test may have actually have been failing and returning a nonzero error code and a zero error code was the pass criteria for the test but was being ignored by ctest because `trilinos_jsrun` was ignoring the return code and always returning 0.  But given that the build `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt` is an 'Experimental' build (and this PR does not change that yet) and given that the current build on 'develop' has 270 failing tests, the risk is super low :-)

I will go ahead and review and approve if I don't see any major issues.  I may do some local testing just to check a few things.  Hopefully this can be merged tonight so that we can see the impact of this in the morning.

jjellio:
@bartlettroscoe @e10harvey 

It could be helpful to somehow include the 'real' `jsrun` line used.

The wrapper has a ENV `ECHO_CMD=1` that prints the before/after commands. Perhaps that part of the script could be changed to echo to STDOUT (Rather than STDERR), and then this info would get included in the output for failed tests?

Infact, if the ECHO is always used, then the ENV could just default to `1` inside the script.

This would be helpful to people trying to debug a failed test.

jjellio:
> > I am not familiar with `CUDA Hook Library: Failed to find symbol mem_find_dreg_entries.`
> 
> Looks like @e10harvey got that fixed in [4665ced](https://github.com/trilinos/Trilinos/commit/4665ced9c9104bea6fa3ea3f91f5c35a3519d6e3)
> 
> We will get a fresh set of experimental test results and assess were we are.

I am pretty sure this Cuda Hooks issue is caused by doing memory allocations before calling MPI_Init (what I said in my comment above this). I forward this problem to LLNL and the Sierra guru (John) said this is a usually caused by the allocs before MPI_Init.


IMO, I would delete the LD_PRELOAD and we need to move this to a new issue so that developers can address this in their unit tests. (Or mark that those unit tests are expected failures. E.g., Init Kokkos first, then MPI *should* fail)

bartlettroscoe:
> IMO, I would delete the LD_PRELOAD and we need to move this to a new issue so that developers can address this in their unit tests. (Or mark that those unit tests are expected failures. E.g., Init Kokkos first, then MPI should fail)

@jjellio, do you have time to help look into this?  Looks like @e10harvey needs to shift to other projects for a while (he has been burning hot on ATDM for a while now).

bartlettroscoe:
> Infact, if the ECHO is always used, then the ENV could just default to 1 inside the script.

@jjellio, sure thing. Can you post a PR for that?

jjellio:
I had a call with the Testbed admins today, and I have an open email with LLNL over this.

Per LLNL, this is most likely a result of allocs before MPI_Init.  Looking at one unit test:
```
./packages/tpetra/core/test/Core/TpetraCore_Core_initialize_where_tpetra_initializes_mpi_and_user_initializes_kokkos.exe
```

This test is doing a Kokkos initialize before MPI_init.  So, it should be an expected failure.

Modifying the wrapper to delete the LD_PRELOAD is straightforward.

IMO, package owners should look through the 'failures', and if they are due to Kokkos coming first, they should mark the test as expected failure and add a test that does it with MPI first. (unfortunately, changing the order can be harder than it would seem on the surface)

bartlettroscoe:
> IMO, package owners should look through the 'failures', and if they are due to Kokkos coming first, they should mark the test as expected failure and add a test that does it with MPI first. (unfortunately, changing the order can be harder than it would seem on the surface)

@jjellio, can we talk about this next week?  Can we set up a meeting?

jjellio:
Sure.

Deleting the LD_PRELOAD is going to cause some number of 'failures'.  I would view this as lower priority, since we know the real apps are doing this.

bartlettroscoe:
> Deleting the LD_PRELOAD is going to cause some number of 'failures'. I would view this as lower priority, since we know the real apps are doing this.

@jjellio, doing what?

jjellio:
They aren't doing memory allocs before MPI_Init, or their codes would have segfaulted.

PR #6732, opened 2020-02-03
Epetra: Removed Deprecated Epetra_MpiSmpComm Class

jeforster:
@trilinos/epetra

## Motivation
The Epetra_MpiSmpComm class was deprecated because it was unused and untested.  This PR removes it.

## Related Issues

* Part of #6655

## Testing
Configured Built and Tested using ATDM scripts on the cee-lan

```bash
mkdir -p $TRILINOS_DIR/build
cd $TRILINOS_DIR/build

source $TRILINOS_DIR/cmake/std/atdm/load-env.sh default

cmake \
  -GNinja \
  -DTrilinos_ENABLE_TESTS=ON \
  -DTrilinos_ENABLE_MueLu=ON \
  -DTrilinos_ENABLE_Belos=ON \
  -DTrilinos_ENABLE_Teuchos=ON \
  -DTrilinos_ENABLE_Epetra=ON \
  -DTrilinos_ENABLE_Panzer=ON \
  -DTrilinos_ENABLE_STK=ON \
  -DTrilinos_ENABLE_Percept=ON \
  -DTrilinos_ENABLE_Intrepid=ON \
  $TRILINOS_DIR

make NP=16

ctest -j16
```


packages/epetra/src/Epetra_MpiSmpCommData.h
```diff
@@ -54,7 +54,6 @@
 */
 
 class Epetra_MpiSmpCommData : public Epetra_Data {
```
kddevin: I suspect Epetra_MpiSmpCommData should also be removed, since Epetra_MpiSmpComm is being removed.
@maherou can you confirm?

kddevin: @jeforster On closer inspection, I don't see anyplace that Epetra_MpiSmpCommData is used in Trilinos.  Removing it along with Epetra_MpiSmpComm should be fine.  Would you please make that change?  Thanks!

jeforster: @kddevin Makes sense.  I will remove that now.

PR #6736, opened 2020-02-04
Zoltan2: fix 6735 (bug in RCM)

### LINKED ISSUE #6735: Zoltan2: reversal of Cuthill McKee ordering is incorrect
The reversal of the Cuthill McKee ordering in `Zoltan2_AlgRCM.hpp` is incorrect.  This can result in orderings that make the matrix bandwidth worse than the original bandwidth.

jhux2:
@trilinos/zoltan2 
## Motivation

This fixes issues reported by @vbrunini, specifically, GMRES/ILU(0) convergence failures for certain application test matrices.

## Related Issues

* Fixes #6735.
* Related to #6675.

## Testing

I think that the Zoltan2 test `ordering1.cpp` could be extended to check that after applying an ordering algorithm, the resulting bandwidth of the matrix matches some gold result.  A weaker check would be to see that an ordering like RCM improves the matrix bandwidth, which might have caught this bug.

I'll check whether one of the matrices I've been working with can be added as a new test input.

I've verified that this allows GMRES/ILU(0) to converge on the two test linear systems that were previously not converging.  It also drastically improved the bandwidth on matrices that I ran through `Zoltan2_Ordering1.exe`.

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/6736)

### DIFF
--- a/packages/zoltan2/src/algorithms/order/Zoltan2_AlgRCM.hpp
+++ b/packages/zoltan2/src/algorithms/order/Zoltan2_AlgRCM.hpp
@@ -116,6 +116,7 @@ class AlgRCM : public Algorithm<Adapter>
   
     // RCM constructs invPerm, not perm
     const ArrayRCP<lno_t> invPerm = solution->getPermutationRCP(true);
+    const ArrayRCP<lno_t> tmpPerm(invPerm.size()); //temporary array used in reversing order
   
     // Check if there are actually edges to reorder.
     // If there are not, then just use the natural ordering.
@@ -140,9 +141,9 @@ class AlgRCM : public Algorithm<Adapter>
     size_t count = 0; // CM label, reversed later
     size_t next = 0;  // next unmarked vertex
     Teuchos::Array<std::pair<gno_t, offset_t> >  children; // children and their degrees
-  
+
     while (count < nVtx) {
-  
+
       // Find suitable root vertex for this component.
       // First find an unmarked vertex, use to find root in next component.
       while ((next < nVtx) && (static_cast<Tpetra::global_size_t>(invPerm[next]) != INVALID)) next++;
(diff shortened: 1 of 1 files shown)


packages/zoltan2/src/algorithms/order/Zoltan2_AlgRCM.hpp
```diff
@@ -140,9 +141,9 @@ class AlgRCM : public Algorithm<Adapter>
...
     size_t next = 0;  // next unmarked vertex
     Teuchos::Array<std::pair<gno_t, offset_t> >  children; // children and their degrees
-  
+
     while (count < nVtx) {
-  
+
       // Find suitable root vertex for this component.
       // First find an unmarked vertex, use to find root in next component.
       while ((next < nVtx) && (static_cast<Tpetra::global_size_t>(invPerm[next]) != INVALID)) next++;
```
mhoemmen: I'm surprised this doesn't warn about extra parentheses, but that's how the code was....

jhux2:
I plan on giving the Zoltan2 team a matrix.  I think one of the existing tests could be adapted without too much trouble.  

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jhux2:
@srajama1 The inversion phase was there but incorrect. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kddevin:
Thanks, @jhux2 
If you provide a matrix, we will add it to our test suite.

packages/zoltan2/src/algorithms/order/Zoltan2_AlgRCM.hpp
egboman: Agreed, looks like a parentheses bug to me.

egboman:
Thanks, this was a bug. Better Zoltan2 tests might have caught it. I still don't like having "gold standard" answers to tests, as there are many implementation details in RCM that can cause slight differences. Probably better to check the bandwidth before and after.

PR #6737, opened 2020-02-04
NonlinML: Fix some scopes in nlnml_nonlinearlevel.cpp

masterleinad:
@trilinos/nonlinML

## Motivation
Avoid some compiler warnings. It looks pretty clear that the scopes were intended to be a little different.

## Testing
Compiling still works and avoids compiler warnings.


mhoemmen:
@trilinos/ml

Issue #6734, opened 2020-02-04
Non-CUDA Clang builds fail in Kokkos


jhux2:
I just did fresh pull of Trilinos develop, SHA 44f13cb14.  An ATDM build on the CEE lan that previously worked just fine is now failing with this error:

```
[2/2834] Building CXX object packages/kokkos/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_CPUDiscovery.cpp.o
FAILED: packages/kokkos/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_CPUDiscovery.cpp.o 
/projects/sems/install/rhel7-x86_64/sems/compiler/clang/7.0.1/openmpi/1.10.1/bin/mpicxx   -I. -Ipackages/kokkos/core/src -I/scratch/jhu/trilinos/Trilinos/packages/kokkos/core/src -Ipackages/kokkos -std=c++11 -x cuda   -O2 -g -DNDEBUG -MD -MT packages/kokkos/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_CPUDiscovery.cpp.o -MF packages/kokkos/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_CPUDiscovery.cpp.o.d -o packages/kokkos/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_CPUDiscovery.cpp.o -c /scratch/jhu/trilinos/Trilinos/packages/kokkos/core/src/impl/Kokkos_CPUDiscovery.cpp
clang-7: error: cannot find libdevice for sm_20. Provide path to different CUDA installation via --cuda-path, or pass -nocudalib to build without linking with libdevice.
clang-7: error: cannot find CUDA installation.  Provide its path via --cuda-path, or pass -nocudainc to build without CUDA includes.
clang-7: error: cannot find CUDA installation.  Provide its path via --cuda-path, or pass -nocudainc to build without CUDA includes.
```

This is not a CUDA build.

Modules:
```
Currently Loaded Modulefiles:
  1) sems-env                         5) sems-gcc/5.3.0                   9) sems-netcdf/4.4.1/exo_parallel
  2) sems-git/2.10.1                  6) sems-clang/7.0.1                10) sems-zlib/1.2.8/base
  3) sems-cmake/3.12.2                7) sems-openmpi/1.10.1             11) sems-boost/1.59.0/base
[...]

PR #6738, opened 2020-02-04
kokkos/cmake/kokkos_arch.cmake: Fix for clang + NO cuda

ndellingwood:
Address compilation issue reported in #6734

## Testing
Tested with similar configuration to that in #6734, but only enabling Kokkos. This resolves compilation errors.

## Configuration
```bash
export TRILINOS_DIR=<PATH>
source ${TRILINOS_DIR}/cmake/std/atdm/load-env.sh Trilinos-atdm-sems-rhel7-clang

ARGS=(
  -GNinja
  -DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake
  -DCMAKE_BUILD_TYPE=RELWITHDEBINFO
  -DTrilinos_ENABLE_TESTS=OFF
  -DTrilinos_ENABLE_Kokkos=ON
  -DKokkos_ENABLE_TESTS=ON
  -DKokkos_ENABLE_SERIAL=ON
)

cmake "${ARGS[@]}" ${TRILINOS_DIR}
```

Trilinos SHA (pre-pr commit): 44f13cb14c643b9f89ebaafe18268c8c48a1e6c3


ndellingwood:
This resolve the compilation errors reported in #6734; however I encountered configuration errors requiring that I modify the ATDM scripts for the sems-rhel7 environment to get a build going, though this could have been user error.

ndellingwood:
@jjwilke adding you here, I'm not allowed to add you as a reviewer.

ndellingwood:
PR's to kokkos need to be issued as well.

bartlettroscoe:
@bathmatt, @jmgate, @rppawlo, this PR will likely be needed to get clang builds of Trilinos + EMPIRE to work.

packages/kokkos/cmake/kokkos_arch.cmake
```diff
@@ -155,7 +155,7 @@ IF (KOKKOS_ENABLE_CUDA_CONSTEXPR)
   ENDIF()
 ENDIF()
 
-IF (KOKKOS_CXX_COMPILER_ID STREQUAL Clang)
+IF (KOKKOS_CXX_COMPILER_ID STREQUAL Clang AND KOKKOS_ENABLE_CUDA)
   SET(CUDA_ARCH_FLAG "--cuda-gpu-arch")
   GLOBAL_APPEND(KOKKOS_CUDA_OPTIONS -x cuda)
   IF (KOKKOS_ENABLE_CUDA)
```
bartlettroscoe: @ndellingwood, given the above addition of `AND KOKKOS_ENABLE_CUDA`, can't you remove this `IF (KOKKOS_ENABLE_CUDA)`?  It would just simplify the code.

bartlettroscoe:
FYI: Last PR testing iteration failed the intel build due to a problem with the Intel license server.  Hopefully that will be resolved next iteration.  I set the `AT: RETEST` label.  Fingers crossed.

jjwilke:
This is likely not the source of the issue. The change that really needs to be done is in the top-level CMakeLists.txt
````
IF (KOKKOS_ENABLE_CUDA)
  STRING(REPLACE ";" " " KOKKOSCORE_COMPILE_OPTIONS "${KOKKOS_COMPILE_OPTIONS}")
  STRING(REPLACE ";" " " KOKKOSCORE_CUDA_OPTIONS    "${KOKKOS_CUDA_OPTIONS}")
  FOREACH(CUDAFE_FLAG ${KOKKOS_CUDAFE_OPTIONS})
    SET(KOKKOSCORE_CUDAFE_OPTIONS "${KOKKOSCORE_CUDAFE_OPTIONS} -Xcudafe ${CUDAFE_FLAG}")
  ENDFOREACH()
ENDIF()
````

ndellingwood:
@jjwilke did you get a chance to test out that change (I can do so shortly)? and should the change I made in this PR stay (I don't see harm in an added safety check based on this issue slipping through, but I don't want to add unnecessary clunkiness to the revamped CMake system)

packages/kokkos/cmake/kokkos_arch.cmake
ndellingwood: Good point, I was hurrying to get this fix in the PR tester. Let me get @jjwilke feedback as well, he proposed a different change below and may want this removed though it seems like an appropriate safety check to add.

bartlettroscoe: @jjwilke and @ndellingwood, if the user is not asking for CUDA, why add **anything** for CUDA in CMake variables at all?  What good can come of that?  That seems like you are just asking for trouble.  If you have to add REMOVE statements, that is a sign of a fragile systems.

jjwilke:
There are no REMOVE calls.

jjwilke:
The flags do not get added if CUDA is not enabled in standalone. Yes, it would probably make more sense to bracket a lot of the logic inside one big `IF(KOKKOS_ENABLE_CUDA)` - but we have been avoiding as much as possible overhauling large swaths of CMake code.

I don't want to rely on `KOKKOS_CUDA_OPTIONS` even if CUDA is not enabled for now.

jjwilke:
Okay - thinking about this some more. We have the situation where we have backends other than CUDA that will end up needing CUDA flags (e.g. OPENMPTARGET) so we will have logic on CUDA flags regardless. It's definitely a todo item for 3.1 to clean up what's there as some of the logic can probably be brackets with an `IF(ENABLE)`

ndellingwood:
Follow up PR with additional changes suggested by @jjwilke issued #6744

PR #6746, opened 2020-02-04
Tacho - disable matrix market reader/writer test to improve PR test stability

kyungjoo-kim:
## Motivation

According to #3276, there is a random test failure from the matrix market interface in Tacho. The apparent error message is "view bounds error". The matrix market code is sequential and run on host. I don't understand why this happens randomly as the code and algorithm are deterministic. I was not able to reproduce this error on openmp nor cuda.  I would like to disable this test for better PR testing. The matrix market code in Tacho is to interface sparse matrices for evaluations, which won't affect the core functionalities of Tacho. 

## Related Issues

#3276

### DIFF
--- a/packages/shylu/shylu_node/tacho/unit-test/Tacho_TestCrsMatrixBase.hpp
+++ b/packages/shylu/shylu_node/tacho/unit-test/Tacho_TestCrsMatrixBase.hpp
@@ -92,39 +92,39 @@ TEST( CrsMatrixBase, constructor ) {
   TEST_END;
 }
 
-TEST( CrsMatrixBase, matrixmarket ) {
-  TEST_BEGIN;
-  std::string inputfilename  = MM_TEST_FILE + ".mtx";
-  std::string outputfilename = MM_TEST_FILE + "_read_output.mtx";      
+// TEST( CrsMatrixBase, matrixmarket ) {
+//   TEST_BEGIN;
+//   std::string inputfilename  = MM_TEST_FILE + ".mtx";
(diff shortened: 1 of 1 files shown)


packages/shylu/shylu_node/tacho/unit-test/Tacho_TestCrsMatrixBase.hpp
```diff
@@ -92,39 +92,39 @@ TEST( CrsMatrixBase, constructor ) {
...
-    EXPECT_EQ(Ah.RowPtrBegin(i), Bh.RowPtrBegin(i));
-    const ordinal_type jbeg = Ah.RowPtrBegin(i), jend = Ah.RowPtrEnd(i);
-    for (ordinal_type j=jbeg;j<jend;++j) {
-      EXPECT_EQ(Ah.Col(j), Bh.Col(j));      
-      EXPECT_EQ(Ah.Value(j), Bh.Value(j));      
-    }
-  }
-  TEST_END;
-}
+// TEST( CrsMatrixBase, matrixmarket ) {
```
ndellingwood: @kyungjoo-kim it looks like the matrix market reader is used in other tests as well, will this change result in the error occurring in a different test?

kyungjoo-kim: Yes. it is used in other tests but it does not fail. That is why I don't understand about this random test failure. If it fails, it should always fail. So, first we disable the problematic test first and see how the PR tests goes afterwards.

bartlettroscoe:
Thanks @kyungjoo-kim!

PR #6750, opened 2020-02-05
Zoltan2:  upgraded testing for issue #6735 and PR #6736

kddevin:
@trilinos/zoltan2 @egboman @srajama1 @jhux2 

## Motivation

Added check for bandwidth reduction for RCM ordering to better catch bugs in implementation.
Added matrix that first exposed the error (although with these checks, we might have caught the problem earlier).

@egboman @srajama1 :  is a non-increasing bandwidth a good enough metric for RCM?

#6735 
#6736 

## Stakeholder Feedback

@vbrunini confirmed that #6736 fixed the issues he was seeing.
@vbrunini also confirmed that Ashear.mtx is acceptable for inclusion in the Zoltan2 repo.
This PR puts his matrix into Zoltan2 testing and adds bandwidth reduction tests for RCM.

## Testing

Tested on mac with clang and openmpi
Confirmed that the new test criteria failed before merging #6736; all tests then passed with merge of #6736.


egboman:
Thanks, Karen! This is a good approach, though RCM is not guaranteed to reduce the bandwidth. I'll look at this when I get back.

PR #6755, opened 2020-02-05
Tacho - example for cusparse trisolve interface from tacho

kyungjoo-kim:
## Motivation
This PR include TriSolve interface to cuSparse. This is to evaluate cuSparse TriSolve performance without supernodal information and tasking overhead.


srajama1:
This is also a good time to try the trisolve @iyamazaki put together.

packages/shylu/shylu_node/tacho/src/Tacho_CuSparseTriSolve.hpp
```diff
@@ -0,0 +1,313 @@
+#ifndef __TACHO_CUSPARSE_TRISOLVE_HPP__
+#define __TACHO_CUSPARSE_TRISOLVE_HPP__
+
+#if defined (KOKKOS_ENABLE_CUDA)
+#include "Tacho_Util.hpp"
+#include "cusparse_v2.h"
```
mhoemmen: I'll note my objection to @crdohrm here that Trilinos is exposing TPL header files downstream.

srajama1: I think @kyungjoo-kim  made it clear he didn't mean to any down stream app or package to include this header. This is for his own testing. However, @kyungjoo-kim I thought I requested is to move this to a test directory or utils to make your intentions clear.

kyungjoo-kim: As I mentioned earlier, these tpl interface is for evaluating their cholesky and its solve performance. The reason I put this in the src directory is because I may want to support these options through the tacho solver interface. I have not decided it yet. If I want to put these TPL interface in the code, I will hide the tpl headers through ETI or a wrapper interface but this is a later work.

srajama1: I think we should not support cuSolver through Tacho interface. That is what Amesos2 is for. It will be very confusing when we have interfaces for the same solver in two places. @MicheldeMessieres we need your PR for cuSolver. It might be of to have it just for testing.

mhoemmen: @kyungjoo-kim I don't mean to take this out on you -- it's probably too late since those TPL header includes are all over kokkos-kernels -- but I just want to say that it's been a problem for Sierra in the past and will likely continue to be a problem.

srajama1: @mhoemmen Can file another issue for Kokkos Kernels ? If this is a concern for applications, we are happy to fix it.

srajama1: @kyungjoo-kim my concern regarding providing interface to TPLs from Tacho remains.

kyungjoo-kim: Right now, the code does not provide TPL interface from tacho. Let's not concern about a thing that does not happen yet.

srajama1: Ok @mhoemmen raised the main issue, he seems fine with it now.

mhoemmen: I'm just expressing a concern for the future.

PR #6768, opened 2020-02-06
tpetra: Add sync check in MV packAndPrepare

ndellingwood:
This PR adds a check for the sync state of 'exports', and synchronizes
if necessary. This prevents issues where the reallocDualViewIfNeeded
call does not result in a resize of 'exports', and thus not clearing the
modify flags before being set in anticipation of computation later in
the routine.

@trilinos/tpetra

## Motivation
Albany builds of Trilinos detected dual view errors, as reported in SNLComputation/Albany#555, following the kokkos 2.9.99 release.

These errors occurred in the `Tpetra::MultiVector::packAndPrepare` routine in cases where the `exports` variable is not resized during a call to `reallocDualViewIfNeeded` which also resulted in the modify flags not being reset. This causes problems later when setting the modify flags in preparation for computation later in the routine.

This PR add a check on the sync state of `exports` and performs a synchronization if neccessary before setting the `exports` modify flags for later computation.

## Stakeholder Feedback
Hopefully addresses SNLComputation/Albany#555

## Testing
Tested a host build using the configuration below on machine with sems modules available

```bash
module load sems-cmake/3.12.2 sems-gcc/7.2.0 sems-git sems-openmpi/4.0.2 sems-superlu/4.3 

export TRILINOS_PATH=<PATH_TO_TRILINOS>

INSTALL_LOCATION=<DESIRED_INSTALL_PATH>
 
cmake \
-D CMAKE_INSTALL_PREFIX:PATH="${INSTALL_LOCATION}" \
-D TPL_ENABLE_MPI:BOOL=ON \
-D CMAKE_CXX_COMPILER="`which mpicxx`" \
-D CMAKE_C_COMPILER="`which mpicc`" \
-D CMAKE_Fortran_COMPILER="`which mpifort`" \
-D KOKKOS_ARCH="SKX" \
  -D TPL_ENABLE_BLAS:STRING=ON \
  -D TPL_ENABLE_LAPACK:STRING=ON \
-D Trilinos_ENABLE_TESTS:BOOL=OFF \
-D Trilinos_ENABLE_EXAMPLES:BOOL=OFF \
\
  -DCMAKE_BUILD_TYPE=DEBUG \
  -DTrilinos_ENABLE_TESTS=OFF \
  -DTPL_ENABLE_SUPERLU=ON \
  -DTPL_ENABLE_Netcdf=OFF \
  -DTrilinos_ENABLE_Kokkos=ON \
    -DKokkos_ENABLE_SERIAL=ON \
    -DKokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON \
  -DTrilinos_ENABLE_Tpetra=ON \
    -DTpetra_ENABLE_TESTS=ON \
  -DTrilinos_ENABLE_Amesos2=ON \
  -DTrilinos_ENABLE_MueLu=ON \
    -DMueLu_ENABLE_TESTS=ON \
  -DTrilinos_ENABLE_Thyra=ON \
    -DThyra_ENABLE_TESTS=ON \
  -DTrilinos_ENABLE_Panzer=ON \
    -DPanzer_ENABLE_TESTS=ON \
${TRILINOS_PATH}
```

The key configure option to set to see the errors is
`    -DKokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON \`

With the changes above in this PR, test in MueLu that would fail with messages like 
`Kokkos::DualView::modify_device ERROR: Concurrent modification of host and device views in DualView "exports"` 
now pass (for example, `MueLu_UnitTestsTpetra_MPI_4`)

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -1362,9 +1362,25 @@ namespace Tpetra {
     // Resizing currently requires calling the constructor, which
     // clears out the 'modified' flags.
     if (packOnHost) {
+      // nde 06 Feb 2020: If 'exports' does not require resize
+      // when reallocDualViewIfNeeded is called, the modified flags 
+      // are not cleared out. This can result in host and device views
+      // being out-of-sync, resuling in an error in exports.modify_* calls.
+      // Adding a check of the sync state, and syncing if necessary, prevents
+      // this possible case.
+      if (exports.need_sync_host ())
+        exports.sync_host ();
       exports.modify_host ();
     }
     else {
+      // nde 06 Feb 2020: If 'exports' does not require resize
+      // when reallocDualViewIfNeeded is called, the modified flags 
+      // are not cleared out. This can result in host and device views
+      // being out-of-sync, resuling in an error in exports.modify_* calls.
+      // Adding a check of the sync state, and syncing if necessary, prevents
+      // this possible case.
+      if (exports.need_sync_device ())
+        exports.sync_device ();
       exports.modify_device ();
     }
 
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -1366,10 +1366,8 @@ namespace Tpetra {
       // when reallocDualViewIfNeeded is called, the modified flags 
       // are not cleared out. This can result in host and device views
       // being out-of-sync, resuling in an error in exports.modify_* calls.
-      // Adding a check of the sync state, and syncing if necessary, prevents
-      // this possible case.
-      if (exports.need_sync_host ())
-        exports.sync_host ();
+      // Clearing the sync flags prevents this possible case.
+      exports.clear_sync_state ();
       exports.modify_host ();
     }
     else {
@@ -1377,10 +1375,8 @@ namespace Tpetra {
       // when reallocDualViewIfNeeded is called, the modified flags 
       // are not cleared out. This can result in host and device views
       // being out-of-sync, resuling in an error in exports.modify_* calls.
-      // Adding a check of the sync state, and syncing if necessary, prevents
-      // this possible case.
-      if (exports.need_sync_device ())
-        exports.sync_device ();
+      // Clearing the sync flags prevents this possible case.
+      exports.clear_sync_state ();
       exports.modify_device ();
     }
 


ndellingwood:
Adding @bartgol @mperego @ikalash

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -1362,9 +1362,25 @@ namespace Tpetra {
     // Resizing currently requires calling the constructor, which
     // clears out the 'modified' flags.
     if (packOnHost) {
+      // nde 06 Feb 2020: If 'exports' does not require resize
```
mhoemmen: We could just call `clear_sync_flags` on `exports`, since we don't care about the DualView's contents on device.  This is a good use case for write-only access.

mperego:
@jwillenbring @bartlettroscoe should we add the option
```
-DKokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON \
```
to a debug PR build?

bartlettroscoe:
@trilinos/kokkos, why is `Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON` not automatically set when `Kokkos_ENABLE_DEBUG=ON` is set?

ndellingwood:
> @trilinos/kokkos, why is `Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON` not automatically set when `Kokkos_ENABLE_DEBUG=ON` is set?

@bartlettroscoe this is the case with 2.9.99, but it was not the case prior to the new release, so this issue probably has been sitting there.

ndellingwood:
@bartlettroscoe I should also point out in my configure above I explicitly enabled `Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON` rather than the full `Kokkos_ENABLE_DEBUG=ON` check because I only needed that option to reproduce the error.

ndellingwood:
> should we add the option
> 
> ```
> -DKokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON \
> ```
> 
> to a debug PR build?

@mperego this is now automatically enabled with `Kokkos_ENABLE_DEBUG` with the 2.9.99 release which is why the issue started showing up.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
ndellingwood: @mhoemmen just to be clear, the suggested change is to replace:
```c++
      if (exports.need_sync_host ())
        exports.sync_host ();
```
with
```c++
      exports.clear_sync_flags ();
```
?

mperego:
> > @trilinos/kokkos, why is `Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON` not automatically set when `Kokkos_ENABLE_DEBUG=ON` is set?
> 
> @bartlettroscoe this is the case with 2.9.99, but it was not the case prior to the new release, so this issue probably has been sitting there.

So why this issue was not caught by the PR tests? Don't we have a PR build that enables `Kokkos_ENABLE_DEBUG=ON` ?

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
ndellingwood: @mhoemmen addressed in recent commit

ndellingwood: Thanks to @brian-kelley for sanity checking the change for me

ndellingwood:
Autotester fail due to update to the PR I pushed to address @mhoemmen comments, adding RETEST

ndellingwood:
@mhoemmen @kddevin @brian-kelley ready for re-review please

PR #6769, opened 2020-02-06
Tpetra: adding more fine grained timing in MatVec Benchmark

lucbv:
@trilinos/tpetra 

## Motivation
Small incremental improvment that should help understand output from timers

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of

### DIFF
--- a/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
+++ b/packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
@@ -335,8 +335,18 @@ main (int argc, char* argv[])
     out << "Command-line options:" << endl;
     printCmdLineOpts (out, opts);
 
-    auto G = getTpetraGraph (comm, opts);
-    auto A = getTpetraCrsMatrix (out, G, opts);
+    auto timer = TimeMonitor::getNewCounter ("Tpetra CrsMatrix Benchmark: getGraph");
+    RCP<Tpetra::CrsGraph<> > G;
+    {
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/example/advanced/Benchmarks/CrsMatrixMatVec.cpp
```diff
@@ -346,23 +356,29 @@ main (int argc, char* argv[])
...
     // case.
-    const SC X_val = static_cast<SC> (1.0) /
-      static_cast<SC> (opts.numEntPerRow);
-    X.putScalar (X_val);
-    Y.putScalar (0.0);
+    timer = TimeMonitor::getNewCounter ("Tpetra CrsMatrix Benchmark: create vectors");
+    {
+      TimeMonitor timeMon (*timer);
+      const SC X_val = static_cast<SC> (1.0) /
+        static_cast<SC> (opts.numEntPerRow);
```
mhoemmen: This might fail if SC is a complex number, since there's no conversion defined from integers to `std::complex<T>`.  However, I'm pretty sure you mean SC to be `double` here, so this should be harmless.

lucbv: Yes that a good point, I think that something like `static_cast<SC>(1.0 / opts.numEntPerRow);` makes more sense and would be safer, I'll push that in a subsequent PR.

PR #6770, opened 2020-02-06
Belos: TpetraOperator

jennloe:
@trilinos/Belos

## Motivation

This pull request makes 3 changes in Belos/tpetra:

1. It gives a Belos::TpetraOperator wrapper class which allows a Belos solver to be used as a Tpetra::Operator.  (Useful for FGMRES, GMRES polynomial preconditioning, etc.)

2. It separates the Block GMRES test in belos/tpetra from the FGMRES test.   The FGMRES option is removed from the block Gmres test, and a new FGMRES test is added which uses Block GMRES as a preconditioner.  (This uses the new TpetraOperator class.)

3.  I added to the comments and explanation for the belos/tpetra/ hybrid GMRES test.  (I thought about adding another test using the polynomial preconditioner via the TpetraOperator interfaces but decided this would be too much duplication.)

## Testing
Tested with Intel compilers.


mhoemmen:
@trilinos/belos (let me see about adding you to the Belos team)

mhoemmen:
@trilinos/framework Could you please add @jennloe to the Trilinos organization?  (I can't add you to the Belos team until you're a part of the Trilinos organization.)

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+/*! \file BelosTpetraOperator.h
+    \brief This file provides a Tpetra::Operator interface so Belos can be integrated into
+     other codes as an abstract operator.
+*/
+
+#ifndef BELOS_TPETRA_OPERATOR_H
+#define BELOS_TPETRA_OPERATOR_H
+
+#include "Tpetra_MultiVector_decl.hpp"
```
mhoemmen: You actually call MultiVector methods (like putScalar) in this file, so you need to include `"Tpetra_MultiVector.hpp"` to get definitions.  Just a class declaration is not enough.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+///////////////////////////////////////////////////////////////
+//-------- class BelosTpetraOperator --------------------
+//
+// This class will allow a Belos solver to be called as an Tpetra_Operator.
+// Thus, it can use itself as a preconditioner if need be.  It can
+// also be used as the inner iteration of Anasazi :)
+//
+///////////////////////////////////////////////////////////////
+template<class Scalar = ::Tpetra::Details::DefaultTypes::scalar_type,
```
mhoemmen: You can actually get the default types from `Tpetra::Operator`.  That way, you won't have to depend on an implementation detail of Tpetra.
```cpp
template<class Scalar = Tpetra::Operator<>::scalar_type,
  class LocalOrdinal = Tpetra::Operator<>::local_ordinal_type,
  class GlobalOrdinal = Tpetra::Operator<>::global_ordinal_type,
  class Node = Tpetra::Operator<>::node_type>
```

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+         class LocalOrdinal = ::Tpetra::Details::DefaultTypes::local_ordinal_type,
+         class GlobalOrdinal = ::Tpetra::Details::DefaultTypes::global_ordinal_type,
+         class Node = ::Tpetra::Details::DefaultTypes::node_type>
+class TpetraOperator : public virtual Tpetra::Operator<Scalar, LocalOrdinal, GlobalOrdinal, Node> {
+public:
+  //! @name Typedefs
+  //@{
+  //! The Tpetra Multivector 
+  typedef Tpetra::MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node> Tpetra_MultiVector;
```
mhoemmen: Public typedefs are part of the public interface in your class, so don't make typedefs public unless you really mean for users to be able to depend on them.  It's perfectly fine to use private typedefs in the declarations of public methods.  For instance:
```cpp
private:
  using Tpetra_MultiVector =
    Tpetra::MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>; // see below
public:
  void apply(const Tpetra_MultiVector& X,
    Tpetra_MultiVector& Y, 
    Teuchos::ETransp mode = Teuchos::NO_TRANS,
    Scalar alpha = Teuchos::ScalarTraits<Scalar>::one(),
    Scalar beta = Teuchos::ScalarTraits<Scalar>::zero()) const override; // see below
```
Also, prefer C++11 `using` syntax (see above code) over `typedef`.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+  for (int i=0; i<(int)solver_name.length(); i++) {
+    Solver[i] = solver_name[i];
+  }   
+  Solver[solver_name.length()] = 0;
+
+  //  
+  // Create solver and solve problem.  This is inefficient, an instance of the solver should
+  // exist already and just be reset with a new RHS.
+  //  
+  if (solver == "BlockGmres") {
```
mhoemmen: `Belos::SolverFactory` already has "aliases" (like nicknames) for solvers.  It's best just to use SolverFactory's facility for aliases, and not try to do SolverFactory's work for it redundantly here.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+    solver = "Pseudo Block Gmres";
+  }
+  else if (solver == "BlockCG") {
+    solver = "Block CG";
+  }
+  else if (solver == "PseudoBlockCG") {
+    solver = "Pseudo Block CG";
+  }
+
+  plist_->remove( "Solver" );
```
mhoemmen: Why remove the parameter?

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+  //! Constructor
+  TpetraOperator( const Teuchos::RCP<LinearProblem<Scalar,Tpetra_MultiVector,Tpetra_Operator> >& lp, 
+		  const Teuchos::RCP<Teuchos::ParameterList>& plist,
+                  bool initSolnVec = false )  
+  : lp_(lp), 
+    plist_(plist),
+    initSolnVec_(initSolnVec)
+  {
+  // Get the solver's name from the parameter list, use block Gmres by default.
+  std::string solver = plist_->get("Solver", "BlockGmres");
```
mhoemmen: If you're going to pass the ParameterList to the solver, then it's best to have the solver name be stored separately from that ParameterList.  Otherwise, the solver's setParameters method might report an error, because it doesn't understand the "Solver" parameter.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+  
+  Belos::SolverFactory<Scalar,Tpetra_MultiVector,Tpetra_Operator> factory;
+  solver_ = factory.create( solver, plist );
+  solver_->setProblem( lp_ );
+
+}
+
+  
+  //! Destructor
+  virtual ~TpetraOperator() {}
```
mhoemmen: Prefer C++11 syntax for defining an empty virtual destructor:
```cpp
virtual ~TpetraOperator() = default;
```

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+  //@}
+  
+  //! @name Operator application methods
+  //@{ 
+  
+  //! Apply the operator.
+  void apply( const Tpetra_MultiVector &X, Tpetra_MultiVector &Y,
```
mhoemmen: It's best to use the `override` keyword to make it clear that this method overrides a virtual method in the base class.  That way, if you get the parameters wrong here, the compiler will report it as a compiler error.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+    lp_->setProblem( vec_Y, vec_X );
+    Belos::ReturnType ret = solver_->solve();
+  };
+  //@}
+  
+  //! @name Attribute access functions
+  //@{ 
+
+  //! Return the label of the operator.
+  const char* Label() const { return(&Solver[0]); };
```
mhoemmen: 1. `return Solver.data();`.  We have C++11 now.  If the `std::vector` is empty, `&Solver[0]` could throw in `operator[]` (so-called "checked" STL implementations can and do).
2. Method names should be lower case.
3. Do you mean to override a `label()` method in a base class?  If so, then I don't think the solver name is really the right "label."  A "label" is something a user can specify, and it can be unique, even for two objects that are otherwise the same.  If you really need this method (ask yourself first why you do), then consider renaming it to something like "solverName".

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,200 @@
...
+  
+  //! Return whether the operator supports applying its transpose or conjugate transpose.
+  bool hasTransposeApply() const { return(false); };
```
mhoemmen: 1. Use the `override` keyword (see above) for all the virtual methods inherited from `Operator`.
2. You don't need the parentheses around `false`.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
```diff
@@ -0,0 +1,271 @@
...
+#define HIDE_TPETRA_INOUT_IMPLEMENTATIONS
+#include <Tpetra_MatrixIO.hpp>
+
+#include <Teuchos_CommandLineProcessor.hpp>
+#include <Teuchos_ParameterList.hpp>
+#include <Teuchos_StandardCatchMacros.hpp>
+#include <Tpetra_Core.hpp>
+#include <Tpetra_CrsMatrix.hpp>
+
+using namespace Teuchos;
```
mhoemmen: Even though this is just a test, it's generally bad practice to import an entire namespace into the global namespace.  What if something in Teuchos that you didn't know about collides with something else?

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
```diff
@@ -0,0 +1,271 @@
...
+//
+#include "BelosConfigDefs.hpp"
+#include "BelosLinearProblem.hpp"
+#include "BelosTpetraAdapter.hpp"
+#include "BelosTpetraOperator.hpp"
+#include "BelosBlockGmresSolMgr.hpp"
+
+// I/O for Harwell-Boeing files
+#define HIDE_TPETRA_INOUT_IMPLEMENTATIONS
+#include <Tpetra_MatrixIO.hpp>
```
mhoemmen: Use double quotes for any header that's not a system header.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
```diff
@@ -0,0 +1,271 @@
...
+// NOTE: No preconditioner is used in this case.
+//
+#include "BelosConfigDefs.hpp"
+#include "BelosLinearProblem.hpp"
+#include "BelosTpetraAdapter.hpp"
+#include "BelosTpetraOperator.hpp"
+#include "BelosBlockGmresSolMgr.hpp"
+
+// I/O for Harwell-Boeing files
+#define HIDE_TPETRA_INOUT_IMPLEMENTATIONS
```
mhoemmen: 1. Prefer run-time options to compile-time options.
2. Will this option ever not be defined in practice?  If so, then you don't need the macro.
3. I don't actually see that you used this macro anywhere below.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
```diff
@@ -0,0 +1,271 @@
...
+using Tpetra::CrsMatrix;
+using Tpetra::MultiVector;
+using std::endl;
+using std::cout;
+using std::vector;
+using Teuchos::tuple;
+
+int main(int argc, char *argv[]) {
+
+  typedef double                           ST;
```
mhoemmen: Prefer C++11 `using` syntax to `typedef`.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
```diff
@@ -0,0 +1,271 @@
...
+    //
+    bool proc_verbose = false;
+    bool debug = false;
+    int frequency = -1;  // how often residuals are printed by solver
+    int numrhs = 1;      // total number of right-hand sides to solve for
+    int blocksize = 1;   // blocksize used by solver
+    int maxiters = -1;   // maximum number of iterations for solver to use
+    int length = 100;    // max subspace size
+    std::string filename("bcsstk14.hb");
+    MT tol = 1.0e-5;     // relative residual tolerance
```
mhoemmen: The default should depend on the Scalar type, e.g., via `Teuchos::ScalarTraits<Scalar>::eps()`.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
jennloe: This was @hkthorn 's doing, copied from the Epetra analogue of this file.  I'm not sure what the intent was.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
jennloe: ?? @hkthorn

packages/belos/tpetra/src/BelosTpetraOperator.hpp
jennloe: So then this would be the reason to delete the solver name from the parameter list....

packages/belos/tpetra/src/BelosTpetraOperator.hpp
jennloe: See below.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
jennloe: 95% of this file is copied from the existing test_bl_gmres_hb.cpp.  All of your comments apply to both files.  Should I make fixes in both places or leave it consistent with what was previously available?

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
jennloe: This would make convergence more dependent on scalar type, but I haven't seen such a convention used in any other Belos tests.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
jennloe: This does not override of a base class function.  Changed the name for clarity.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
jennloe: I do not actually know what this is.  Assumed it was needed because it was in the original version.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
mhoemmen: @jennloe If "Scalar" is always `double` then it doesn't matter so much.  If you intend for the test to be generic, then it matters.  In particular, `Scalar=float` does not get enabled by default in PR tests (even though Sierra depends on it! @trilinos/framework ), so if you intend to test that case, then the tolerance should be a function of the Scalar type.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
mhoemmen: @jennloe I've never seen this macro before.  Does it exist anywhere else in Trilinos?

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
mhoemmen: @jennloe It depends on how much time you have.  If you have time to fix existing Belos tests, please do feel free! :-)

packages/belos/tpetra/src/BelosTpetraOperator.hpp
mhoemmen: It's probably because solvers don't know what this "Solver" parameter is.  That suggests we should make the solver name separate from the ParameterList.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
jennloe: I'm going to create a separate issue for this.  It affects nearly every Belos test, and I am not ready to change it here.

jennloe: Or maybe I won't.  For these particular block GMRES tests, the scalar type is hard-coded as double.  It would need further work to generalize.  
There are some tests for block GMRES and CG that test both double and float.

jennloe: In other concerns- Is there a reason that Tpetra doesn't have tests for BiCGStab, TFQMR, or MinRes?  Or PseudoBlockGMRES?  We might be asking for trouble when Epetra gets phased out.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
jennloe: I'm not seeing it anywhere else in Trilinos.  But it is in 3 of 4 existing GMRES tests, 4 CG tests and a fixed point iteration test.

jennloe: @hkthorn Do you know what this is?

Removing it doesn't seem to cause problems on my end, so we will try taking it out.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
jennloe: Moving it to be a parameter to the constructor.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
hkthorn: @jennloe I have no idea where this came from, it is likely very old and unnecessary.  Feel free to remove.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
jennloe: Leaving this for now to maintain consistency with other files.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
hkthorn: @jennloe. Please feel free to more broadly correct the Belos Tpetra tests.

packages/belos/tpetra/test/BlockGmres/test_bl_fgmres_hb.cpp
hkthorn: @jennloe There is no good reason for there not being tests for the other Belos solvers and you bring up a good point that this code coverage is essential to ensuring correctness between Tpetra and Belos.  I say that you move forward with the focus of this particular pull request.  Addressing further necessary improvements to testing is beyond the scope of this work.

jennloe:
Okay, see if this looks better.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,205 @@
...
+    lp_->setProblem( vec_Y, vec_X );
+    Belos::ReturnType ret = solver_->solve();
+  };
+  //@}
+  
+  //! @name Attribute access functions
+  //@{ 
+
+  //! Return the name of the requested solver.
+  const char* getSolverName() const { return Solver.data(); };
```
mhoemmen: 1. Users would expect to be able to use this name in `Belos::SolverFactory`, to get an instance of the same solver used here.  However, users would be wrong -- you changed the original name they provided by adding some label stuff to it.  It would be better to give back either the original name, or the canonical name from `Belos::SolverFactory`.
2. You don't need the `std::vector<char>`.  Just return `solver_name_.c_str()`.
3. Why not return a `std::string`?

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,205 @@
...
+  
+  //! Return whether the operator supports applying its transpose or conjugate transpose.
+  bool hasTransposeApply() const override { return false; };
+  
+  //! Return the domain map for this operator.
+  Teuchos::RCP<const Tpetra_Map> getDomainMap() const override
+   { return (lp_->getOperator()->getDomainMap()); };
```
mhoemmen: You don't need the extra parentheses around the return value here.

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,205 @@
...
+#include "Teuchos_ParameterList.hpp"
+
+/*! \class Belos::TpetraOperator
+    \brief This class provides an interface to the Tpetra::Operator class, so Belos can be 
+    integrated into other codes as an abstract operator.
+*/
+
+namespace Belos {
+
+///////////////////////////////////////////////////////////////
```
mhoemmen: This documentation is redundant.  It's also wrong -- it's a `Belos::TpetraOperator`, not a `BelosTpetraOperator`, etc.  Why not just delete it and add text to the Doxygen documentation above?

jennloe:
@hkthorn What was the original intent of the Label function in the Epetra version of this?  @mhoemmen My first thought was that this might be used to label Teuchos timers, which was why I left the existing string which has more information than just the name of the solver.  Maybe we need two functions- one that returns the solver name and one that gives a more descriptive label??

Related- @hkthorn  Do you remember why the Epetra Label function returned std::vector<char> instead of a string?

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+  Teuchos::RCP<SolverManager<Scalar,Tpetra_MultiVector,Tpetra_Operator> > solver_;
+  Teuchos::RCP<LinearProblem<Scalar,Tpetra_MultiVector,Tpetra_Operator> > lp_;
+  Teuchos::RCP<Teuchos::ParameterList> plist_;
+
+  //! Name of solver to be passed into solver factory. 
+  std::vector<char> Solver;
+  std::string solver_name_;
+
+  //! True if TpetraOperator should initialize the solution vector to zero before
```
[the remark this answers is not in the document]
jennloe: Okay.  On a related note- I'm not sure if the Doxygen is actually being generated for files in this directory.  I can't find it online in the Belos Doxygen.  Where would I check for this?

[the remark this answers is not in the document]
hkthorn: https://docs.trilinos.org/dev/packages/belos/doc/html/index.html

packages/belos/tpetra/src/BelosTpetraOperator.hpp
hkthorn: @mhoemmen @jennloe The joys of inheritance!  Please eliminate the usage of char* in the Tpetra implementation of the operator.  The Label() function in Epetra returns const char*, which is the reason for character string manipulation including the addition of a null-termination character that was for Janus (a computer from 2003 with some bad juju compilers).

packages/belos/tpetra/src/BelosTpetraOperator.hpp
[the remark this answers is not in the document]
jennloe: Have looked through this.  Can't find Doxygen for any files in belos/tpetra.  What am I missing?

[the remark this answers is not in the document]
hkthorn: @jennloe I don't see Doxygen for belos/epetra either, just links to the belos/epetra examples.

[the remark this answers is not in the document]
mhoemmen: @hkthorn Do we need to add an `INPUT` line to `belos/doc/Doxyfile.options`?  It doesn't have one, but other packages do -- see e.g., `tpetra/doc/Doxyfile`.

mhoemmen:
1. Unused variable warning (see below)
2. CUDA build error: For whatever reason, that build can't deduce the conversion from `std::string` to `const char*`.  Best way to fix that would be either to change `getSolverName` to return `std::string`, or to change the implementation of `getSolverName` to return `solver_name_.c_str()`.

```
...BelosTpetraOperator.hpp:154:23: error: unused variable ‘ret’ [-Werror=unused-variable]
     Belos::ReturnType ret = solver_->solve();
                       ^~~
```

packages/belos/tpetra/src/BelosTpetraOperator.hpp
```diff
@@ -151,15 +151,15 @@ class TpetraOperator : public Tpetra::Operator<Scalar, LocalOrdinal, GlobalOrdin
       lp_->setInitResVec( vec_X );
     }
     lp_->setProblem( vec_Y, vec_X );
-    Belos::ReturnType ret = solver_->solve();
+    solver_->solve();
```
mhoemmen: Thanks for the fixes!  Later on, we could think about giving users a way to get the solve status, but that can be for later : - ) .

jennloe: Yes, I thought about it briefly, but no immediate solution presented itself.  I guess we could create another function that could query the result of the last solve... But if the main purpose is to use belos as a preconditioner, the user probably isn't especially concerned about whether it converged.

mhoemmen: @jennloe It could be useful for inner-outer iterations, to avoid recomputing the residual, but you might need to do that anyway.  We can always not implement optional features ;-) .

mhoemmen:
@jennloe Do you have merge permission?  If not, just let us know and we'll merge it.

jennloe:
@mhoemmen No, I'm blocked from merging.  Not an "authorized user."

mhoemmen:
@jennloe tagging @trilinos/framework to see about getting you added to the list of Trilinos developers, so we can add you to the Belos team.  Not sure who is in charge of this now -- maybe @jwillenbring ?

kddevin:
It's best to email @jwillenbring directly.

PR #5680, opened 2019-08-08
Added check for divide by 0

bathmatt:
@trilinos/<teamName>

## Checklist

- [ ] My commit messages mention the appropriate GitHub issue numbers.
- [ ] My code follows the code style of the affected package(s).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] No new compiler warnings were introduced.
- [ ] These changes break backwards compatibility.

PR #6772, opened 2020-02-06
Fix kk spgemm zero row

brian-kelley:
@trilinos/kokkos-kernels 

## Motivation

Fixes SPGEMM for input with zero rows. This fixes the same problem as #5680 did before it got clobbered by the promotion, but it fixes it for multiple SPGEMM algorithms. Tests this case now. See https://github.com/kokkos/kokkos-kernels/pull/599 for more info.

## Stakeholder Feedback

EMPIRE needed this fixed because it broke many tests
## Testing

Added an SPGEMM test for zero rows. This replicated the failure that was seen in EMPIRE tests, and this PR fixes it.

### DIFF
--- a/packages/kokkos-kernels/src/common/KokkosKernels_IOUtils.hpp
+++ b/packages/kokkos-kernels/src/common/KokkosKernels_IOUtils.hpp
@@ -78,7 +78,7 @@ void kk_sparseMatrix_generate(
 {
   rowPtr = new SizeType[nrows+1];
 
-  OrdinalType elements_per_row = nnz/nrows;
+  OrdinalType elements_per_row = nrows ? nnz/nrows : 0;
   srand(13721);
   rowPtr[0] = 0;
   for(int row=0;row<nrows;row++)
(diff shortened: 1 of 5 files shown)


ndellingwood:
Note #5680 never had a matching PR submitted to kokkos-kernels which resulted in the changes being clobbered. To prevent this in the future, PRs involving changes to kokkos-kernels now require approval from the code owners (and that approval requires referencing the matching PR to kokkos-kernels)

brian-kelley:
@ndellingwood Does the PR testing do complex? If not, I'll do a local build of it since the behavior is different (with alignment) than what I've already tested for the KK PR.

ndellingwood:
> @ndellingwood Does the PR testing do complex? If not, I'll do a local build of it since the behavior is different (with alignment) than what I've already tested for the KK PR.

@brian-kelley good question, I'm pretty sure complex is tested in a host build but not with cuda to trim down time on autotester builds, so a local cuda build with complex enabled is a great idea, thanks!

brian-kelley:
@ndellingwood With complex+cuda, kokkos-kernels and tpetra are clean.

Issue #6422, opened 2019-12-10
Zoltan2_fix4785_MPI_4 test in build Trilinos-atdm-white-ride-cuda-10.1-gnu-7.2.0-release-debug failing starting 2019-11-26


bartlettroscoe:
CC: @trilinos/zoltan2, @kddevin  (Trilinos Data Services Product Lead), @bartlettroscoe, @fryeguy52, @MicheldeMessieres 

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-11-11&end=2019-12-10&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Trilinos-atdm-white-ride-cuda-10.1-gnu-7.2.0-release-debug&field2=testname&compare2=61&value2=Zoltan2_fix4785_MPI_4&field3=site&compare3=61&value3=ride) the test:

* `Zoltan2_fix4785_MPI_4`

in the build:

* `Trilinos-atdm-white-ride-cuda-10.1-gnu-7.2.0-release-debug`

started failing on 2019-11-26.

[...]

PR #6778, opened 2020-02-07
Zoltan2 mj clean up

MicheldeMessieres:
@trilinos/zoltan2

## Motivation

@kddevin Apart from one change to make progress on #6422, this PR is not associated with any failures. It fixes various issues I had noticed while working on other MJ bugs over the past few weeks.

Changed doPostsAndWaits to get host ptrs and not uvm space ptrs. This was not causing any issue I am aware of but it seems it could cause an issue for some MPI implementations.

Changed deccltype declarations to use auto which cleans up code.

Removes a pointless #ifdef which can never do anything. This was leftover from an earlier refactor.

Removes a hack on one of the kernels to adjust team count for small test sizes. This was an issue a while back but it seems fine now. So perhaps it was caused by another bug (now fixed in MJ) or something was repaired in Kokkos.

Fixed an incorrect openmp thread count reading. For one kernel it was always using 1 thread for OpenMP regardless of actual availability. This will improve OpenMP performance.

Also I have made one seemingly pointless change which clears the cuda 10 pascal bug - #6422. Since tracking this bug has been very difficult, I think it may be useful to just see if this does clear for all the auto testing. There may be an issue with how I am handling ArrayView of subviews.

 ## Related Issues
Related to #6422 

## Testing

Tested on Cuda (white), OpenMP (white) and Mac parallel/serial builds.

### DIFF
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -4938,8 +4938,8 @@ mj_create_new_partitions(
 
 #else
 
-#ifdef KOKKOS_HAVE_OPENMP
-  const int num_threads = omp_get_num_threads();
+#ifdef KOKKOS_ENABLE_OPENMP
+  const int num_threads = Kokkos::OpenMP::impl_max_hardware_threads();
 #else
   const int num_threads = 1;
 #endif
--- a/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
+++ b/packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
@@ -5438,12 +5438,12 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t,
   }, rectilinear_cut_count);
 
   if(rectilinear_cut_count > 0) {
-    typename decltype(local_process_rectilinear_cut_weight)::HostMirror
-      host_local_process_rectilinear_cut_weight =
-      Kokkos::create_mirror_view(local_process_rectilinear_cut_weight);
-    typename decltype(local_global_rectilinear_cut_weight)::HostMirror
-      host_local_global_rectilinear_cut_weight =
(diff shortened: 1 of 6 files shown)


packages/zoltan2/src/algorithms/partition/Zoltan2_AlgMultiJagged.hpp
```diff
@@ -6559,18 +6521,22 @@ void AlgMJ<mj_scalar_t, mj_lno_t, mj_gno_t, mj_part_t, mj_node_t>::
...
-        host_dst_coordinates = Kokkos::create_mirror_view(dst_coordinates);
+      auto host_dst_coordinates = Kokkos::create_mirror_view(
+        Kokkos::HostSpace(), dst_coordinates);
       for(int i = 0; i < this->coord_dim; ++i) {
-        Kokkos::View<mj_scalar_t *, device_t> sub_host_src_coordinates
-          = Kokkos::subview(host_src_coordinates, Kokkos::ALL, i);
-        Kokkos::View<mj_scalar_t *, device_t> sub_host_dst_coordinates
+        Kokkos::View<mj_scalar_t*, Kokkos::Serial> sub_host_src_coordinates;
+        // view could be size 0 if graph was not distributed
+        if(host_src_coordinates.extent(0) != 0) {
```
kddevin: Is this check for extent != 0 needed?  There is not a similar check of the extent for sub_host_src_weights or below.

MicheldeMessieres: @kddevin I forgot I had included that change which was specific just to a chaco test. This change is in the #ifdef block for using Zoltan to migrate instead of Tpetra, which is off unless someone does a manual code change. So it's not currently used (or tested). When I did the original PR updating MJ to kokkos that code path was actually left broken for the chaco test and I didn't try it until later. So this was to fix that unused path which I have been maintaining. However if we want to keep that code I should probably add some testing for it.  I expect no one is using the Zoltan option as it was in a broken state for many weeks without any complaints.

I will review the situation to better answer your question. If you agree I could address further issues with this as a follow up PR since I don't think anyone is using that code right now. So it should be safe to allow the change. There is a similar change for the Tpetra path which has been in for a while.

PR #6780, opened 2020-02-07
Fix div by zero

brian-kelley:
@trilinos/kokkos-kernels 

## Motivation

In debug builds spgemm was still failing with FPE

## Stakeholder Feedback

Still breaking EMPIRE-PIC tests


srajama1:
Thanks, @brian-kelley , for the quick fix.

brian-kelley:
@srajama1 Me too. Still trying to understand why the release vs. debug behavior is different here - the divide was happening outside of a ``KOKKOSKERNELS_VERBOSE`` block and yet it passed my new test with zero rows/zero entries...

And the TransmissionLine-mpi-parallel test passed on release build this morning.

ndellingwood:
Please add reference to the matching kokkos-kernels PR when it goes in, thanks!

brian-kelley:
OK, I guess FPEs don't happen in a release build, you just get the undefined behavior. The complex alignment crap probably hid the failure in the spot-check debug builds I did yesterday.

brian-kelley:
@ndellingwood Matches https://github.com/kokkos/kokkos-kernels/pull/601

bathmatt:
so I asked above but the message is pending for some odd reason, why is ratio set to 0?  It will return true in this case maybe

brian-kelley:
@bathmatt It will always return true, but that just means it will use the compressed version of symbolic over zero rows, instead of the uncompressed version. Neither of them will do any actual work over zero rows so there's no performance difference.

brian-kelley:
@bathmatt I guess we could avoid the cost of 2 kernel launches (< 100 us) by intercepting the special zero-row/zero-entry case before even attempting compression, but that shouldn't ever be in the critical section because some other ranks will be doing a real product at the same time.

Issue #6781, opened 2020-02-07
Panzer but think it is really tpetra


bathmatt:
## Bug Report
@trilinos/tpetra

If you build panzer with debug-openmp and run the tests you will see a ton fail with an assert

```
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false
Kokkos::DualView::modify_host ERROR: Concurrent modification of host and device views in DualView "MV::DualView"
[s993289:11497] *** Process received signal ***
[s993289:11497] Signal: Aborted (6)
[s993289:11497] Signal code:  (-6)
Kokkos::DualView::modify_host ERROR: Concurrent modification of host and device views in DualView "MV::DualView"
[s993289:11498] *** Process received signal ***
[s993289:11498] Signal: Aborted (6)
[s993289:11498] Signal code:  (-6)

```

Assume this is caused by the checks added by @ndellingwood on the 6th...

### Steps to Reproduce

build develop with gnu-openmp on rhel7 (probably all machines)
run panzer test

(DESKTOP)-dir-gnu-debug-openmp-shared [GNU-7.2.0]$ 58 $ ctest -R PanzerDofMgr_scaling_test

see bug

bathmatt:
So, tpetra tests pass as do the muelu tests.  So, it looks more like a panzer bug that got exposed with the check

mhoemmen:
@bathmatt, @ndellingwood created a PR that might fix this: https://github.com/trilinos/Trilinos/pull/6768 .  It's in develop now.

The cause was that Kokkos' debug settings were not enabling DualView flag checks by default.  They do now.  @ndellingwood noticed the test failures and changed Tpetra to fix them.

ndellingwood:
@bathmatt I can try and help look into this later (I'm out today but had a chance to check email and github and saw this). What Trilinos branch and SHA is this occurring with and what configuration + system?
The error message you posted is occurring because a DualView's  device flags were marked modified by `modify_device`, and then the host was later marked modified by `modify_host` without a synchronization or clearing of flags in between (i.e. "Concurrent modification of host and device views in DualView")

A couple things that can help chase this down:
If it is occurring in Tpetra, setting these environment variables may help hone in on the trouble spot
`export TPETRA_DEBUG=1; export TPETRA_VERBOSE=1` 

[...]

PR #6782, opened 2020-02-07
panzer dof-mgr: add needed check and sync for MV dual view

ndellingwood:
Part of issue #6781

Tested that this resolved failing Panzer tests in host build with `-DKokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK=ON`


ndellingwood:
@trilinos/panzer

packages/panzer/dof-mgr/src/Panzer_DOFManager.cpp
```diff
@@ -795,6 +795,8 @@ DOFManager::buildGlobalUnknowns_GUN(const Tpetra::MultiVector<panzer::GlobalOrdi
   {
     PANZER_FUNC_TIME_MONITOR_DIFF("panzer::DOFManager::buildGlobalUnknowns_GUN::line_13-21 gid_assignment",GUN13_21);
     int which_id=0;
+    if (non_overlap_mv->need_sync_host())
```
mhoemmen: The access pattern below is read-and-write, so this is correct.

Issue #6790, opened 2020-02-10
Test NOX_Tpetra_1DFEM_MPI_4 random failure showing "Concurrent modification of host and device views in DualView" starting 2020-02-09 


bartlettroscoe:
CC: @trilinos/kokkos, @trilinos/tpetra, @trilinos/nox, @kddevin (Data Services Product Lead), @rppawlo (Nonlinear Solvers Product Lead)

As shown [here](https://testing-dev.sandia.gov/cdash/testDetails.php?test=83986517&build=5227386), the test:

*  NOX_Tpetra_1DFEM_MPI_4

in the build:

*  Trilinos-atdm-sems-rhel6-gnu-7.2.0-openmp-debug

failed with the error:

```
0. NOX_Tpetra_1DFEM_AnalyticJacobian_NoPrec_UnitTest ... Kokkos::DualView::modify_host ERROR: Concurrent modification of host and device views in DualView "MV::DualView"
[sems-srn-rhel6-slave-01:15867] *** Process received signal ***
[sems-srn-rhel6-slave-01:15867] Signal: Aborted (6)
[sems-srn-rhel6-slave-01:15867] Signal code:  (-6)
Kokkos::DualView::modify_host ERROR: Concurrent modification of host and device views in DualView "MV::DualView"
Kokkos::DualView::modify_host ERROR: Concurrent modification of host and device views in DualView "MV::DualView"
[sems-srn-rhel6-slave-01:15869] *** Process received signal ***
[sems-srn-rhel6-slave-01:15870] *** Process received signal ***
[sems-srn-rhel6-slave-01:15870] Signal: Aborted (6)
[sems-srn-rhel6-slave-01:15870] Signal code:  (-6)
[sems-srn-rhel6-slave-01:15869] Signal: Aborted (6)
[sems-srn-rhel6-slave-01:15869] Signal code:  (-6)
Kokkos::DualView::modify_host ERROR: Concurrent modification of host and device views in DualView "MV::DualView"
[sems-srn-rhel6-slave-01:15868] *** Process received signal ***
[sems-srn-rhel6-slave-01:15868] Signal: Aborted (6)
[sems-srn-rhel6-slave-01:15868] Signal code:  (-6)
[sems-srn-rhel6-slave-01:15867] [ 0] /lib64/libpthread.so.0[0x309a60f7e0]
[sems-srn-rhel6-slave-01:15870] [ 0] /lib64/libpthread.so.0[0x309a60f7e0]
[...]

PR #6793, opened 2020-02-10
nox: add missing check for sync in dual view

ndellingwood:
Address issue #6790

 Changes to be committed:
	modified:   test/tpetra/ME_Tpetra_1DFEM_def.hpp

@trilinos/nox 

## Motivation

Address issue #6790 

## Testing

Reproduced the failure in test `NOX_Tpetra_1DFEM_MPI_4` in a host build, confirmed it passes with these changes.

### DIFF
--- a/packages/nox/test/tpetra/ME_Tpetra_1DFEM_def.hpp
+++ b/packages/nox/test/tpetra/ME_Tpetra_1DFEM_def.hpp
@@ -327,12 +327,11 @@ evalModelImpl(const Thyra::ModelEvaluatorBase::InArgs<Scalar> &inArgs,
     // Create ghosted objects
     if (is_null(uPtr_))
       uPtr_ = Teuchos::rcp(new tpetra_vec(xGhostedMap_));
-      uPtr_->modify_host();
-      uPtr_->doImport(*(tpetra_extract::getConstTpetraVector(inArgs.get_x())), *importer_, Tpetra::REPLACE);
+
+    uPtr_->doImport(*(tpetra_extract::getConstTpetraVector(inArgs.get_x())), *importer_, Tpetra::REPLACE);
 
     if (is_null(xPtr_)) {
       xPtr_ = Teuchos::rcp(new tpetra_vec(xGhostedMap_));
-      xPtr_->modify_host();
       xPtr_->doImport(*nodeCoordinates_, *importer_, Tpetra::INSERT);
     }
 


packages/nox/test/tpetra/ME_Tpetra_1DFEM_def.hpp
```diff
@@ -325,13 +325,19 @@ evalModelImpl(const Thyra::ModelEvaluatorBase::InArgs<Scalar> &inArgs,
     typedef typename tpetra_vec::execution_space execution_space;
 
     // Create ghosted objects
+    // Should there be braces corresponding to this if-statement and the following lines of code (as in xPtr_ case below)?
```
ndellingwood: @rppawlo can you confirm if braces should wrap the code after the if-statement, similar to the `xPtr_` case below?

ndellingwood:
Marked WIP until confirmation regarding potential missing braces for an if-statement.

packages/nox/test/tpetra/ME_Tpetra_1DFEM_def.hpp
```diff
@@ -325,13 +325,19 @@ evalModelImpl(const Thyra::ModelEvaluatorBase::InArgs<Scalar> &inArgs,
...
     if (is_null(uPtr_))
       uPtr_ = Teuchos::rcp(new tpetra_vec(xGhostedMap_));
+
+      if (uPtr_->need_sync_host())
+        uPtr_->sync_host();
       uPtr_->modify_host();
```
mhoemmen: @rppawlo Why are you calling `modify_host` on the target of an Import?  It looks like this was intended to force MultiVector to pack and unpack on host.  That doesn't necessarily work like you think it does any more.  For example, the target of an Export or Import has the right to control its own sync state.

packages/nox/test/tpetra/ME_Tpetra_1DFEM_def.hpp
rppawlo: @ndellingwood - the uPtr code should NOT be wrapped in braces. Each iteration we will need a new import on the solution vector (uPtr). The xPtr holds nodal coords and is not changing from iteration to iteration so only needs to be called once. We can probably remove the modify_host() calls as @mhoemmen suggested.

packages/nox/test/tpetra/ME_Tpetra_1DFEM_def.hpp
rppawlo: @mhoemmen - not sure. This was written by someone who has left sandia, so I can't get a quick answer. Can we just remove the modify calls?

mhoemmen: @rppawlo We should remove them, unless downstream code implicitly depends on MultiVector being sync'd to host.  There's no good way to test for that, though.

ndellingwood: > We should remove them, unless downstream code implicitly depends on MultiVector being sync'd to host

@rppawlo @mhoemmen I'll hold off making any changes until getting a definite "yes" to remove the `modify_host` calls (and the sync checking I added in this PR)

packages/nox/test/tpetra/ME_Tpetra_1DFEM_def.hpp
ndellingwood: @rppawlo thanks for clarifcation.

packages/nox/test/tpetra/ME_Tpetra_1DFEM_def.hpp
rppawlo: yes - let's remove them.

ndellingwood:
@rppawlo @mhoemmen I removed the sync changes I introduced, removed the unnecessary `modify_host` calls, and changed indentation of the uPtr_ import call to prevent potential warnings about misleading indentation. Rebased the changes for a clean single commit

mhoemmen:
Thanks @ndellingwood ! :-D

PR #6798, opened 2020-02-11
kokkos-kernels: spgemm: fix scalar alignment

brian-kelley:
Permanent fix to scalar_t memory alignment in scratch views.
Affected KKMEM and SPEED algorithms.
Use new Utils function "alignPtr" that adds a few bytes to an
arbitrary non-void pointer (char*, nnz_lno_t*, etc.) to
correctly align to another type T*.

Mirror of https://github.com/kokkos/kokkos-kernels/pull/602

@trilinos/kokkos-kernels 

## Motivation

In SPGEMM numeric phase, fixes incorrect alignment of ``scalar_t`` arrays in scratch buffers. On CPU this usually just degrades performance but on CUDA alignment is required and misalignment causes kernels to crash. This was observed with ``scalar_t = Kokkos::complex<double>`` where the complex was 16 byte aligned, but in principle the error could happen with ``scalar_t=double`` too.

## Testing

This is tested in the SPGEMM unit tests in kokkos-kernels, and in KokkosKernels, this was failing for CUDA+complex enabled. However, this never happens when testing kokkos-kernels as a Trilinos package, because Kokkos::complex alignment is disabled in Trilinos. The purpose of this PR in Trilinos is just to prepare for when this workaround is removed.

### DIFF
--- a/packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
+++ b/packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
@@ -1523,6 +1523,7 @@ struct ReduceMaxRowFunctor{
       view_type rowmap_view_): rowmap_view(rowmap_view_),
           min_val(0)
   {}
+
   KOKKOS_INLINE_FUNCTION
   void operator()(const size_t &i, value_type &max_reduction) const {
     value_type val = rowmap_view(i+1) - rowmap_view(i) ;
@@ -1933,6 +1934,16 @@ struct array_sum_reduce
       data[i] += src.data[i];
   }
 };
+
+template<typename InPtr, typename T>
+KOKKOS_INLINE_FUNCTION T* alignPtr(InPtr p)
+{
+  //ugly but computationally free and the "right" way to do this in C++
+  std::uintptr_t ptrVal = reinterpret_cast<std::uintptr_t>(p);
+  //ptrVal + (align - 1) lands inside the next valid aligned scalar_t,
+  //and the mask produces the start of that scalar_t.
+  return reinterpret_cast<T*>((ptrVal + alignof(T) - 1) & (~(alignof(T) - 1)));
+}
 
 }
 }
--- a/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_impl_kkmem.hpp
+++ b/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_impl_kkmem.hpp
@@ -40,7 +40,10 @@
 // ************************************************************************
 //@HEADER
 */
+
 #define HASHSCALAR 107
+
+#include "KokkosKernels_Utils.hpp"
 
 namespace KokkosSparse{
 
@@ -58,7 +61,6 @@ struct KokkosSPGEMM
     b_lno_row_view_t_, b_lno_nnz_view_t_, b_scalar_nnz_view_t_>::
   PortableNumericCHASH
 {
-
   nnz_lno_t numrows;
 
   a_row_view_t row_mapA;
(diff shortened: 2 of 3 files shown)


packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_impl_kkmem.hpp
```diff
@@ -40,8 +40,11 @@
 // ************************************************************************
 //@HEADER
 */
+
 #define HASHSCALAR 107
```
mhoemmen: We need to fix that macro sometime.  Macros that are exposed in header files must have package namespacing.

packages/kokkos-kernels/src/common/KokkosKernels_Utils.hpp
```diff
@@ -1934,6 +1935,16 @@ struct array_sum_reduce
...
 };
 
+template<typename InPtr, typename T>
+KOKKOS_INLINE_FUNCTION T* alignPtr(InPtr p)
+{
+  //ugly but computationally free and the "right" way to do this in C++
+  std::uintptr_t ptrVal = reinterpret_cast<std::uintptr_t>(p);
+  //ptrVal + (align - 1) lands inside the next valid aligned scalar_t,
+  //and the mask produces the start of that scalar_t.
+  return reinterpret_cast<T*>((ptrVal + alignof(T) - 1) & (~(alignof(T) - 1)));
```
mhoemmen: Neat bit trick! :-D

packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_impl_kkmem.hpp
```diff
@@ -168,8 +170,10 @@ struct KokkosSPGEMM
 
     while (team_cuckoo_key_size * 2 < tmp_team_cuckoo_key_size) team_cuckoo_key_size = team_cuckoo_key_size * 2;
     team_cuckoo_hash_func = team_cuckoo_key_size - 1;
-    team_shmem_key_size = ((shared_memory_size - sizeof(nnz_lno_t) * 4) / unit_memory);
-    thread_shmem_key_size = ((thread_memory - sizeof(nnz_lno_t) * 4) / unit_memory);
+    //How many extra bytes are needed to align a scalar_t after an array of nnz_lno_t, in the worst case?
+    constexpr size_t scalarAlignPad = (alignof(scalar_t) > alignof(nnz_lno_t)) ? (alignof(scalar_t) - alignof(nnz_lno_t)) : 0;
```
mhoemmen: Would be nice to abstract this into a `constexpr` function at some point.

packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_impl_kkmem.hpp
```diff
@@ -492,7 +412,7 @@ struct KokkosSPGEMM
 
     hm2.keys = (nnz_lno_t *) (tmp);
     tmp += max_nnz;
-    hm2.values = (scalar_t *) (tmp);
+    hm2.values = KokkosKernels::Impl::alignPtr<volatile nnz_lno_t*, scalar_t>(tmp);
```
mhoemmen: Do volatile pointers have different alignment than non-volatile pointers?

brian-kelley: Not that I know of, but tmp is volatile so I didn't want to take chances with cv- conversion warnings or errors. I don't actually think the 'volatile' here is necessary, but that's a different issue.

brian-kelley: Like the macro you saw above, SPGEMM has a lot of little cleanup things that need to be done. Another one is that $view.data() gets taken as a raw pointer for the numeric functors, instead of just storing and accessing the view.

PR #6749, opened 2020-02-05
(WIP DO NOT MERGE) Tpetra: First pass at fixing #6663

mhoemmen:
@trilinos/tpetra @nate945 @shardest @crdohrm 

## Motivation

Our hypothesis is that double-counting incoming entries that could be merged into existing target matrix entries is the reason for the memory increase.  These changes will help test that.  They are also a first pass at fixing #6663.  This PR includes the following changes:

  1. Do not double-count incoming entries in `CrsMatrix::copyAndPermute` or `CrsMatrix::unpackAndCombine`.  (The way I did this may also benefit `CrsGraph::copyAndPermute` in the same way.)
  2. Remove redundant allocations in `CrsMatrix::applyCrsPadding`.
  3. Add verbose output for counting how much space (1) saved.

## Related Issues

* Part of #6663 

## Stakeholder Feedback

See #6663 discussion from Sierra SD developers.

## Testing

Tested TpetraCore on my Mac with Clang and OpenMPI.

PR #6806, opened 2020-02-11
(WIP; DO NOT MERGE) Tpetra: Fix #6663

### LINKED ISSUE #6663: Tpetra::CrsMatrix::unpackAndCombine, non-static-graph path, allocates too much memory

mhoemmen:
@trilinos/tpetra @tjfulle 

## Motivation

Tpetra::CrsGraph currently may use too much memory when it is the target of a doExport or doImport.  Likewise, Tpetra::CrsMatrix currently may use too much memory when it owns its graph (has a "non-static" CrsGraph) and is the target of a doExport or doImport.

Due to removal of DynamicProfile, graphs and matrices now only have the StaticProfile option, which stores data in a "row offsets begin, number of entries in row" 4-array (3-array for graphs) variant of compressed sparse row storage.  This means that `copyAndPermute` and `unpackAndCombine` must reallocate storage before inserting incoming new graph / matrix entries.

Before this PR, `copyAndPermute` and `unpackAndCombine` would allocate space for all the incoming entries in each row, without considering that they might merge with existing target graph / matrix entries in that row.  (They would "merge" because they have the same row and column indices.)  That meant that the target graph / matrix might allocate much more storage than it needs.

This PR fixes that issue, by considering merges when computing the new allocation sizes needed in `copyAndPermute` and `packAndPrepare`.

I ended up deleting `unpackAndCombineCrsGraph`.  Sorry, @tjfulle -- the function depended too heavily on the old implementation of space reallocation.  It also wasn't being used at all, though it was being tested.  (This could explain why the unit tests before weren't enough to catch errors -- CrsGraph wasn't actually calling those functions, but it gave the impression that we had plenty of unit tests for `unpackAndCombine`.)

## Related Issues

* Closes #6663 
* Replaces #6749
* Related to #6648 

## Stakeholder Feedback

See #6648 and #6663 for feedback and lots of diagnosis help from Sierra SD developers.

## Testing

* Added new Tpetra tests for new functionality
* Tested on Mac Clang + OpenMPI


mhoemmen:
Oops, gotta get rid of that "extended lambda" for the CUDA build....

tjfulle:
Thanks @mhoemmen!

I'm sitting down right now to go through this.  I'm planning to just build and run the tests locally and read through.  Anything in particular to look out for?

RE removal of `unpackAndCombineCrsGraph` - I had forgotten about the function.  I think it's just a straight copy of `unpackAndCombineCrsMatrix` (without values) and maybe had been intended to be modified further to unpack optional values - thereby giving a single graph/matrix interface.  But, I think the Stokhos types complicated the matter enough that  it didn't make sense (that's my recollection, at least).  I'm glad to see the old stuff go :)

tjfulle:
All Tpetra tests pass on rhel6/gcc7/openmpi2.  Building all downstream code now...

mhoemmen:
@tjfulle wrote:
> I'm sitting down right now to go through this. I'm planning to just build and run the tests locally and read through. Anything in particular to look out for?

Thanks Tim! :-D  The main thing is to check whether the strategy makes sense.  Is the code straightforward to read, maintain, and test?  Is there a natural extension to optimizations like only (re)allocating the target once?  (I aimed for that -- note how CrsPadding stores all the incoming data in a single data structure indexed by target local row index.  The idea is that import LIDs might overlap with sames or permutes, and there might be multiple imports to the same local row.)

My view about GPU-ization is that we should wait until we can show a performance bottleneck.  Right now, we have a memory use bottleneck, so we should address that first.  A more GPU-friendly approach would have just added overhead.  CrsPadding is not thread-safe to write, but `get_result` is thread safe (on host) to read, which I think I exploit at one point in a host `Kokkos::parallel_for`.

tjfulle:
On my blade, many tests timeout and a few rol tests fail.  @mhoemmen, do you think the timeout is related or maybe an issue with my blade?  I'll track down the rol failures - it seems par the course anytime these codes are touched!!

mhoemmen:
@tjfulle wrote:
>  @mhoemmen, do you think the timeout is related or maybe an issue with my blade?

I would be surprised if it were related -- that would mean some tests are doing a lot of CrsGraph doExport or doImport calls.  The one test failure is a build warning that I can easily fix.  Thanks for looking into the ROL test failures!

tjfulle:
The ROL tests don't appear to be Tpetra related.  They might be related to the dozen or more timeouts.  I'll look more tomorrow

mhoemmen:
Just fixed the build warning!  This will give me a chance to test in situ.

mhoemmen:
@tjfulle Hold on, I got SD to fail with this:
```
.../Tpetra_Details_CrsPadding.hpp:257:

Throw number = 1

Throw test that evaluated to true: !(newLen == newDiffNumEnt)
```

mhoemmen:
@tjfulle There was a bug in how I was computing the size of the set difference.  I'm testing a fix in SD now.

mhoemmen:
Just fixed the other build warning, and more importantly, fixed some correctness issues in CrsPadding that only SD was exercising.  If I can get the SD test to pass, I think it's OK to accept these changes for now, but I would like to add more tests to Tpetra at some point.  It's pretty scary that no Trilinos test failed when I had the set difference size computations wrong.  We need to fix that.

mhoemmen:
@tjfulle @trilinos/tpetra While this passes Trilinos tests, it causes SD to hang before it even gets to the code in question.  We need more tests.  I will try to diagnose the SD issue.

mhoemmen:
It's hanging on Process 2 somewhere after the start but before the end of `Tpetra::CrsGraph::computePaddingForCrsMatrixUnpack`.

tjfulle:
@mhoemmen hmmmm, maybe the timeouts in downstream tests aren't red herrings?  ROL seemed like a dead end, I'll look at a few of the MueLu timeouts

mhoemmen:
@tjfulle I'm adding a ton more verbose output to try to figure this out.  Right now, SD is my only hope for debugging.

mhoemmen:
@tjfulle Yay, I got through the part of SD that was failing!!! :-D  I had to remove a chunk of debugging code to make it get through.  That chunk looks like it was looping forever, even though it just calls a function that actually has unit tests :-) .  Thus, I'd like to investigate a bit more before I actually push this.

mhoemmen:
Superseded by https://github.com/trilinos/Trilinos/pull/6831 .

mhoemmen:
I just merged PR #6831, which fixes #6663.  Thus, I'm deleting this branch.

PR #6808, opened 2020-02-11
Tempus: Prototype of New Interface for BE.

### LINKED ISSUE #6711: Tempus:  Prototype Stepper Interface Manager
## Enhancement
@trilinos/tempus 

Try out design ideas around how to replace the StepperObserver.  After talking with application teams various ideas were proposed.  Will to implement some of them and see how it meets desired requirements.

ccober6:
Closes #6711 

@trilinos/tempus 

## Description
This is a new design to replace the current Observer design.  The basic idea is from discussions with @krcb and others which generalizes this interface.  The basic idea is to have a general interface on the Tempus side for the Steppers, while having various interfaces available to users on the application side.

To begin, we have duplicated the Backward Euler Stepper and created the BE Stepper to prototype the new interface.  In the BE Stepper, we replaced the StepperObserver with the StepperBEAppAction, which is pure virtual requiring implementation of the execute() function.  This is essentially a Strategy design pattern to allow multiple interfaces to the applications "actions" in the StepperBE.
Screen Shot 2020-02-11 at 9 59 59 AM

We have three derived "actions":
 * Modifier - has non-const access to the SolutionHistory and the StepperBE.  (This is essentially the old Observer.)
 * Observer - has const access to the SolutionHistory and the StepperBE.
 * ModifierX - has reduced access to the solution to reduce the chance of users inadvertently corrupting the stepper method.  This interface just allows changes to x and/or xDot, thus the name ModifierX. 

The base class versions of these derived "actions" take the execute() call and adapt it to its interface (an Adapter design).
 * StepperBEAppAction.execute() --> StepperBEModifierBase.execute() --> StepperBEModifier.modify()
 * StepperBEAppAction.execute() --> StepperBEObserverBase.execute() --> StepperBEObserver.observe()
 * StepperBEAppAction.execute() --> StepperBEModifierXBase.execute() --> StepperBEModifierX.modify()

As examples of what the application needs to implement, the "default" version show the interface that the application sees and would need to implement.  The default versions are simple "no-op" implementations.

## Further Extensions
 * Should implement a composite StepperBEAppAction, so users can apply multiple "actions".
 * After agreement on this design, we should prototype a StepperRKAppAction.  Note: each Stepper gets its own AppAction, and there is no base class across Steppers.
 * With two steppers with the AppAction, we should test the idea of reusing an AppAction for multiple steppers.  This would be done by creating an AppAction derived from both steppers AppActions.

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorForwardSensitivity_impl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorForwardSensitivity_impl.hpp
@@ -326,35 +326,35 @@ createSensitivityModelAndStepper(
 
 /// Non-member constructor
 template<class Scalar>
-Teuchos::RCP<Tempus::IntegratorForwardSensitivity<Scalar> >
+Teuchos::RCP<IntegratorForwardSensitivity<Scalar> >
 integratorForwardSensitivity(
   Teuchos::RCP<Teuchos::ParameterList>                     pList,
   const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >&      model)
 {
-  Teuchos::RCP<Tempus::IntegratorForwardSensitivity<Scalar> > integrator =
-    Teuchos::rcp(new Tempus::IntegratorForwardSensitivity<Scalar>(pList, model));
+  Teuchos::RCP<IntegratorForwardSensitivity<Scalar> > integrator =
+    Teuchos::rcp(new IntegratorForwardSensitivity<Scalar>(pList, model));
   return(integrator);
 }
 
 /// Non-member constructor
 template<class Scalar>
-Teuchos::RCP<Tempus::IntegratorForwardSensitivity<Scalar> >
+Teuchos::RCP<IntegratorForwardSensitivity<Scalar> >
 integratorForwardSensitivity(
   const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >&      model,
   std::string stepperType)
 {
-  Teuchos::RCP<Tempus::IntegratorForwardSensitivity<Scalar> > integrator =
-    Teuchos::rcp(new Tempus::IntegratorForwardSensitivity<Scalar>(model, stepperType));
+  Teuchos::RCP<IntegratorForwardSensitivity<Scalar> > integrator =
+    Teuchos::rcp(new IntegratorForwardSensitivity<Scalar>(model, stepperType));
   return(integrator);
 }
 
 /// Non-member constructor
 template<class Scalar>
-Teuchos::RCP<Tempus::IntegratorForwardSensitivity<Scalar> >
+Teuchos::RCP<IntegratorForwardSensitivity<Scalar> >
 integratorForwardSensitivity()
 {
-  Teuchos::RCP<Tempus::IntegratorForwardSensitivity<Scalar> > integrator =
-    Teuchos::rcp(new Tempus::IntegratorForwardSensitivity<Scalar>());
+  Teuchos::RCP<IntegratorForwardSensitivity<Scalar> > integrator =
+    Teuchos::rcp(new IntegratorForwardSensitivity<Scalar>());
   return(integrator);
 }
 
--- a/packages/tempus/src/Tempus_StepperBackwardEulerAppAction.hpp
+++ b/packages/tempus/src/Tempus_StepperBackwardEulerAppAction.hpp
@@ -1,0 +1,75 @@
+// @HEADER
+// ****************************************************************************
+//                Tempus: Copyright (2017) Sandia Corporation
+//
+// Distributed under BSD 3-clause license (See accompanying file Copyright.txt)
+// ****************************************************************************
+// @HEADER
+
+#ifndef Tempus_StepperBackwardEulerAppAction_hpp
+#define Tempus_StepperBackwardEulerAppAction_hpp
+
+#include "Tempus_config.hpp"
+#include "Tempus_SolutionHistory.hpp"
+#include "Tempus_StepperBackwardEuler.hpp"
+
+
+namespace Tempus {
+
+// Forward Declaration for recursive includes (this AppAction <--> Stepper)
+template<class Scalar> class StepperBackwardEuler;
+
+/** \brief Application Action for StepperBackwardEuler.
+ *
+ *  This class provides a means to apply various actions with the BackwardEuler time step.
+ *  The data available to this class is solution variables (through
+ *  SolutionHistory), and stepper data (through the Stepper).  It allows
+ *  the application to just observe this data (i.e., use but not change the
+ *  data) to change any of it (USER BEWARE!).
+ *
+ *  Below is the BackwardEuler algorithm and includes the locations where the
+ *  application can take actions (in italicized).
+ *
+ *  \f{algorithm}{
+ *  \renewcommand{\thealgorithm}{}
+ *  \caption{Backward Euler with application-action locations indicated.}
+ *  \begin{algorithmic}[1]
+ *    \State {\it appAction.execute(solutionHistory, stepper, BEGIN\_STEP)}
+ *    \State Compute the predictor (e.g., apply stepper to $x_n$).
+ *    \State {\it appAction.execute(solutionHistory, stepper, BEFORE\_SOLVE)}
+ *    \State Solve $\mathcal{F}_n(\dot{x}=(x_n-x_{n-1})/\Delta t_n, x_n, t_n)=0$ for $x_n$
+ *    \State {\it appAction.execute(solutionHistory, stepper, AFTER\_SOLVE)}
+ *    \State $\dot{x}_n \leftarrow (x_n-x_{n-1})/\Delta t_n$
+ *    \State {\it appAction.execute(solutionHistory, stepper, END\_STEP)}
+ *  \end{algorithmic}
+ *  \f}
+ */
+template<class Scalar>
+class StepperBackwardEulerAppAction
+{
+public:
+
+  /// Indicates the location of application action (see algorithm).
+  enum ACTION_LOCATION {
+    BEGIN_STEP,     ///< At the beginning of the step.
+    BEFORE_SOLVE,   ///< Before the implicit solve.
+    AFTER_SOLVE,    ///< After the implicit solve.
+    END_STEP        ///< At the end of the step.
+  };
+
+  /// Constructor
+  StepperBackwardEulerAppAction(){}
+
+  /// Destructor
+  virtual ~StepperBackwardEulerAppAction(){}
+
+  /// Execute application action for BackwardEuler Stepper.
+  virtual void execute(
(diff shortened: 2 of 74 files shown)


ccober6:
@krcb @eric-c-cyr @seamill @michael-a-hansen @ikalash @bartgol @mperego @asteyer @sconde @rppawlo 

Hi all,
I wanted you to take a look at the new design to replace the old Observer design, and provide feedback on it.  Hopefully it meets many of our needs.  I have provided an intro in the PR description, and you can see the details in the code changes.  Let me know what you think.
Thanks, Curt

michael-a-hansen:
Hi @ccober6 

Nice work, this looks neat. It sounds like it provides the complete existing capability, which allows a Tempus user to do _anything_ they want, along with two 'safer' interfaces that each limit the capability in some way and may be easier to use than `Modifier`. Is that accurate? Can the new `Modifier` do everything the old `Observer` could do? The ability for a user to "roll your own" components (e.g., steppers, controllers for us) into the overall time stepping process was important for us and I think the extensibility of `oldObserver` can't be lost. Stacking safer interfaces on top of it is great. Ideally that would reduce the learning curve in addition to making things safer.

Question: is the intention to provide only these three interfaces and, if a user needs some intermediate state between two of them, they have to jump to the next less safe interface?

eric-c-cyr:
@ccober This looks like what we I expected based on our conversations. I think this is a big step in the right direction. The fact that @michael-a-hansen gleaned its motivation just from looking att the code is a good sign. Thanks.

ccober6:
Morning Mike,

Your description of this is correct.  The new `Modifier` can do everything the old `Observer` could do, so there is no lost in extensibility.  Users can create their own `AppActions` (e.g., to make their own specialized interface) or derive off of three current `AppActions` (`Modifier`, `Observer` or `ModifierX`).

Right now we have these three interfaces, but we can add others to Tempus as applications find a need.  With this Strategy/Adapter pattern, it should be fairly easy to create what various teams might want.

ccober6:
Morning Eric,

Glad to hear we are moving in the right direction on this!  Please let us know if you have any suggested changes.

ccober:
Hi, you've added my email to this thread by mistake.  I'm Chris Cober, not
curtis ober.

Ccober@gmail.com

ccober6:
@krcb I am starting to work on the Explicit RK version of this.  Let me know if there is anything with the BE version that you would like see changed in the Explicit RK version.

ccober6:
@asteyer @krcb Just added you as reviewers.  These changes are checking out with EMPIRE and backward compatibility.  @krcb Can you look it over for meeting EMPIRE needs?

packages/tempus/src/Tempus_IntegratorForwardSensitivity_impl.hpp
```diff
@@ -326,35 +326,35 @@ createSensitivityModelAndStepper(
 
 /// Non-member constructor
 template<class Scalar>
-Teuchos::RCP<Tempus::IntegratorForwardSensitivity<Scalar> >
+Teuchos::RCP<IntegratorForwardSensitivity<Scalar> >
```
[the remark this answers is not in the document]
ccober6: So I thought this is OK and it passed tests.  Tests are in the Tempus_Test or Tempus_Unit_Test namespaces, so there are out side the Tempus namespace and I have never used "using namespace Tempus".  Is there a case that this will not work?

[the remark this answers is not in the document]
krcb: These objects are inside the Tempus namespace and this code is as well? I'm just puzzled about the original usage and the change; not implying that it won't work.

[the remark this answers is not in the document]
ccober6: This was cleanup.  Like you mention, these objects and code are inside the Tempus namespace.  I am not sure why the original author (Eric Phipps) did it this way.  I was hopefully removing a possible double namespace issue (i.e., Tempus::Tempus:: ...).

krcb:
@ccober6 When I try and build the application against this branch, I see the following:

```
<PATH_TO_TRILINOS>/install/trilinos-release/include/Tempus_StepperBackwardEuler_decl.hpp:17:10: fatal error: 'Tempus_StepperBackwardEulerAppAction.hpp' file not found
#include "Tempus_StepperBackwardEulerAppAction.hpp"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Seems like a missing header file from the install?

ccober6:
Tempus_StepperBackwardEulerAppAction.hpp is part of the branch?!  Not sure how it can not be there?  Are you using the 6711-Tempus-Prototype-Stepper-Interface-Manager branch?

ccober6:
Downstream applications builds fine in my tests.

krcb:
@ccober6 Well, this is strange. The first time I built and installed, this didn't install. The second time it appears to have done so.

ccober6:
@krcb I hate when that happens!  :)

packages/tempus/unit_test/Tempus_UnitTest_BackwardEuler.cpp
```diff
@@ -79,25 +83,346 @@ TEUCHOS_UNIT_TEST(BackwardEuler, Construction)
...
+  auto solutionHistory = Tempus::createSolutionHistoryME(model);
+
+  // Take one time step.
+  stepper->setInitialConditions(solutionHistory);
+  solutionHistory->initWorkingState();
+  double dt = 0.1;
+  solutionHistory->getWorkingState()->setTimeStep(dt);
+  stepper->takeStep(solutionHistory);
+
+  TEST_COMPARE(modifier->testBEGIN_STEP, ==, true);
```
krcb, line 202: @ccober6 Could you add some comments on what these tests on the modifier are checking? Is it that each of the ACTION_LOCATION's have been called?

packages/tempus/src/Tempus_StepperBackwardEulerAppAction.hpp
```diff
@@ -0,0 +1,75 @@
...
+  StepperBackwardEulerAppAction(){}
+
+  /// Destructor
+  virtual ~StepperBackwardEulerAppAction(){}
+
+  /// Execute application action for BackwardEuler Stepper.
+  virtual void execute(
+    Teuchos::RCP<SolutionHistory<Scalar> > sh,
+    Teuchos::RCP<StepperBackwardEuler<Scalar> > stepper,
+    const typename StepperBackwardEulerAppAction<Scalar>::ACTION_LOCATION actLoc) = 0;
```
krcb, line 70: @ccober6 To make sure that I understand, the user can only pass in a member of `StepperBackwardEulerAppAction<Scalar>::ACTION_LOCATION` here; all of which are executed in the context of this stepper. In addition, if the user tries to pass in a different type of `ACTION_LOCATION`, there is a compile time error?

packages/tempus/unit_test/Tempus_UnitTest_BackwardEuler.cpp
ccober6: @krcb Will do. Yes, just checking that each ACTION_LOCATION has been called.  I have also done this for the Explicit RK testing.

packages/tempus/src/Tempus_StepperBackwardEulerAppAction.hpp
ccober6: @krcb Yes. Each Stepper's AppAction is tied to its Stepper.  This way we should not need to cast the Stepper down to its derived type.  With that, we can type the ACTION_LOCATION to it derived type.  For the RK Steppers (Explicit, DIRK and IMEX), we added an RKBase class to allow reusing AppActions across all the RK Stepper (along with other things).  This is not complete yet, and mods are going into #7150 "as we speak".  With the second RK type (i.e., DIRK) going in, I am seeing things that should moved up the inheritance.

ccober6:
> @ccober6 A couple of more comments regarding AppAction and the tests. I do have a design concern here in addition to these comments: because AppAction is being specialized for each class of Stepper, I think this means that the downstream application is going to have to implement different AppActions for each Stepper type (all of which potentially could be performing the same operation).

So you are right, and this is the tension we have been having.  Trying to reuse AppActions across all Steppers is generally not possible as the takeStep() varies greatly across all the Steppers.  The attempt here is to tie the AppActions to its Stepper.  However for RK Steppers which have similar takeStep() functions, we will generalize it so that the RK AppActions can be reused across all RK Steppers.

I am hoping that this is the compromise that meets the most needs.  If a non-RK Stepper is needed and the AppAction needs to be reused, we should use the General RK to create a new RK Stepper which could reused the AppAction, or just explicitly create a new RK Stepper.  Thoughts?

packages/tempus/unit_test/Tempus_UnitTest_BackwardEuler.cpp
krcb: @ccober6 Ok, this makes sense. Based on [this comment](https://github.com/trilinos/Trilinos/pull/6808#discussion_r406841146), it's not possible to perform a runtime test of what happens if the wrong action is passed in, as this will result in a compile time error?

krcb:
> So you are right, and this is the tension we have been having. Trying to reuse AppActions across all Steppers is generally not possible as the takeStep() varies greatly across all the Steppers. The attempt here is to tie the AppActions to its Stepper. However for RK Steppers which have similar takeStep() functions, we will generalize it so that the RK AppActions can be reused across all RK Steppers.
> 
> I am hoping that this is the compromise that meets the most needs. If a non-RK Stepper is needed and the AppAction needs to be reused, we should use the General RK to create a new RK Stepper which could reused the AppAction, or just explicitly create a new RK Stepper. Thoughts?

@ccober6 I think [this comment](https://github.com/trilinos/Trilinos/pull/6808#discussion_r406841146) ameliorates a lot of my concern. If we can have a base AppAction that unifies together the different RK schemes, then I think that enables a sufficient level of code reuse.

ccober6:
@asteyer Can you re-review this?  I had to make changes to pass on Intel.

krcb:
@jmgate This is the issue we were just discussing.

krcb:
@ccober6 @asteyer Please hold off from approving/merging this until @jmgate can chime in.

jmgate:
You guys can merge whenever you like.  If we need to clean things manually on the EMPIRE side, that's fine.

ccober6:
@jmgate Sounds good!  Once I get approval from @asteyer, I will merge it.

jmgate:
@krcb, this merge means we need to clean our Trilinos build and install directories, yes?

ccober6:
@jmgate I did when I was testing, but I am not sure it is required.  I added some new files and changed some includes but have not removed any files yet.

krcb:
@jmgate I had to do do this, yes. Otherwise I got build errors.

jmgate:
Sounds good.  Thanks.

Issue #532, opened 2016-07-27
Tpetra::CrsMatrix: Implement 2-arg "copy constructor" like MultiVector's


mhoemmen:
@trilinos/tpetra Give Tpetra::CrsMatrix a two-argument "copy constructor" like MultiVector's.  It should take an optional second argument that tells it whether to make a deep copy.  If the graph is a const graph, it should be able just to copy the values.  

If users want a "true deep copy" of the graph as well, the fix for #531 will offer that.

mhoemmen:
Clarification: This was only relevant for `Tpetra::MultiVector` during the 2013-5 "Kokkos refactor" effort, when we had to support two Tpetra implementations, only one of which had view semantics (where constructors only shallow copied by default).  We don't need this any more.

PR #6813, opened 2020-02-11
Use Kokkos-kernels TRSM for CholQR

iyamazaki:
@trilinos/<teamName>

## Motivation
Use KokkosKernel TRSM for CholQR from KokkosKernel PR [#532](https://github.com/kokkos/kokkos-kernels/pull/532).

## Testing
Passed ctest in belos/tpetra/test on Blake.


packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSstep.hpp
```diff
@@ -333,8 +369,26 @@ class GmresSstep : public Gmres<SC, MV, OP>  {
...
+        int rank = 0;
+        {
+          Teuchos::TimeMonitor LocalTimer (*tsqrTimer);
+          rank = normalizeCholQR (iter, step, Q, G);
+          if (useCholQR2_) {
+            rank = normalizeCholQR (iter, step, Q, G2);
+            // merge R 
+            dense_matrix_type Rfix (Teuchos::View, G2, step+1, step+1, iter, 0);
+            dense_matrix_type Rold (Teuchos::View, G,  step+1, step+1, iter, 0);
+            blas.TRMM (Teuchos::LEFT_SIDE, Teuchos::UPPER_TRI,
```
mhoemmen: It looks like the issue is that you want to use `Belos::(Mat)OrthoManager`, but that interface always takes small dense matrices as host data in a `Teuchos::SerialDenseMatrix`.  This is why you're calling the BLAS on host here.  Is that right?

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSstep.hpp
```diff
@@ -48,32 +47,50 @@ class CholQR {
...
     int info = 0;
-    lapack.POTRF ('U', ncols, R.values (), R.stride(), &info);
-    if (info < 0) {
-      ncols = -info;
-      // FIXME (mfh 17 Sep 2018) Don't throw; report an error code.
-      throw std::runtime_error("Cholesky factorization failed");
+    R_mv.modify_host ();
+    {
+      auto R_h = R_mv.getLocalViewHost ();
+      lapack.POTRF ('U', ncols, R_h.data (), R_h.extent (0), &info);
```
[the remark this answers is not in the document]
iyamazaki: That sounds great. So, we'll be calling cuSolver through Kokkos-kernels?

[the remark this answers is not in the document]
mhoemmen: @iyamazaki It looks like kokkos-kernels wants to maintain the cuSOLVER interface, so as soon as they have that interface available, we can use it.  My main worry is that we're switching back and forth between device and host a lot.  If the multiply ends up being on host anyway (because `MPI_Allreduce` doesn't know how to handle CUDA buffers), and if the matrix to factor is small, then it might not be worth the kernel launch overhead to do Cholesky on device.  Thus, I think it would be better to do an experiment first, before we go through the trouble of wrapping cuSOLVER for kokkos-kernels.

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSstep.hpp
iyamazaki: I was doing it just for convenience (e.g., I convert R to H, and solve the small least-square problem on the host), but I think you are right. We have the option to use Belos OrthoManager, which keeps the small matrices on host. I'll look at if it makes more sense to keep them on device for the performance..

mhoemmen:
@iyamazaki The only reason I didn't approve this is because there's no test for the CholQR2 (reorthogonalization) case : - ) .  (It's Trilinos policy that a new feature requires a new test.)

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSstep.hpp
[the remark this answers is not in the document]
srajama1: I am going to take the bait on the cuSOLVER interface comment and respond. I will also include @kddevin as the Tpetra lead. Adding cuSOLVER to Tpetra was incorrect and I said that even when that PR was going in. It was ok to push it because some customer needed it. However, this goes against Trilinos principles of TPL gets interfaced at the right level and in one place. We have violated this basic principle multiple times and that has caused issues with mismatched TPLs. Doing it once for a customer as a **temporary** fix was ok with an assumption it will be fixed. We already have @e10harvey looking at the interfaces to clean that problem on the issue @kddevin filed for Kokkos Kernels. Recommending that pattern to others is just simply not right. 


@iyamazaki did the right thing by requesting TRSM interface and interfacing it right here.

@kddevin Let us talk about this in in the next product leaders meeting on what are Tpetra's plans for design choices with cuSOLVER. I am willing to do cleanup one hot fix, but we cannot do this if using cuSolver is a recommended in PRs.

packages/belos/tpetra/test/Native/CMakeLists.txt
```diff
@@ -104,3 +104,27 @@ TRIBITS_ADD_TEST(
...
+  NAME Tpetra_Native_GMRES_S_STEP_nonsymm_matrix_step5_cholqr
+  ARGS "--solver=\"TPETRA GMRES S-STEP\" --restartLength=10 --stepSize=5 --useCholQR"
+  COMM serial mpi
+  STANDARD_PASS_OUTPUT
+  )
+
+TRIBITS_ADD_TEST(
+  Tpetra_Native_solvers_nonsymm_matrix
+  NAME Tpetra_Native_GMRES_S_STEP_nonsymm_matrix_step5_cholqr2
+  ARGS "--solver=\"TPETRA GMRES S-STEP\" --restartLength=10 --stepSize=5 --useCholQR2"
```
mhoemmen: Thanks for the test! :-D  Does this mean that it _always_ reorthogonalizes?

iyamazaki: Thank *you* for the review. Yes, with CholQR2, it always reorthogonalize. No check on when to reorthogonalize is not implemented..

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresPipeline.hpp
```diff
@@ -276,8 +293,9 @@ class GmresPipeline : public Gmres<SC, MV, OP> {
           Teuchos::Range1D index_prev(0, iter+1);
           const MV Qprev  = * (Q.subView(index_prev));
 
+          auto v_iter = Kokkos::subview(vals, std::pair<int, int>(0, iter+2));
           vec_type W = * (V.getVectorNonConst (iter+1));
-          req = Tpetra::idot (vals, Qprev, W);
+          req = Tpetra::idot (v_iter, Qprev, W);
```
mhoemmen: Hi @iyamazaki !  It looks like your change has a build error:
```
.../packages/belos/tpetra/solvers/Belos_Tpetra_GmresPipeline.hpp:298:47: error: no matching function for call to ‘idot(Kokkos::View<double*, Kokkos::LayoutRight, Kokkos::Device<Kokkos::OpenMP, Kokkos::HostSpace>, Kokkos::MemoryTraits<0u> >&, const Tpetra::MultiVector<>&, BelosTpetra::Impl::GmresPipeline<double, Tpetra::MultiVector<>, Tpetra::Operator<> >::vec_type&)’
           req = Tpetra::idot (v_iter, Qprev, W);
```
The issue could be that `idot` expects a rank-0 `Kokkos::View` if its second and third arguments have type `Tpetra::Vector`.  That's not what you want here -- you want to compute `Qprev^* W`.  You can fix this by casting `W` to `const MV&`:
```cpp
req = Tpetra::idot(v_iter, Qprev, static_cast<const MV&>(W));
```
GCC 4.8.4 might not be as good at deducing template parameters as newer compilers, so sometimes we have to help it along.

iyamazaki: Thank you, @mhoemmen for the tip. I have not managed to reproduce the compilation errors, but pushed the change you have suggested.

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSstep.hpp
```diff
@@ -83,7 +84,8 @@ class CholQR {
     {
       auto R_h = R_mv.getLocalViewHost ();
       int ldr = int (R_h.extent (0));
-      lapack.POTRF ('U', ncols, R_h.data (), ldr, &info);
+      IST *Rimpl = reinterpret_cast<IST*> (R_h.data ());
```
mhoemmen: That's right -- that's how you do it : - ) .

iyamazaki: No.. It still did not work. I guess my IST is still Kokkos::complex<double>. So, I get "no known conversion to std::complex<double>". :(

mhoemmen:
@iyamazaki Are you ready to merge?

iyamazaki:
Yes. Thank you so much for the help with this PR, @mhoemmen !!

mhoemmen:
Thanks @iyamazaki for your patience and hard work! : - )

Issue #6260, opened 2019-11-12
Trilinos: Transition Trilinos to require c++14 standard


rppawlo:
Transition Trilinos to require c++14 standard rather than c++11. Many apps are already building with c++14 and Kokkos would benefit greatly from this. This ticket is to outline requirements and start discussions with stakeholders. 

Compiler support:
Aggregate table of compiler support: https://en.cppreference.com/w/cpp/compiler_support

Minimum compiler version that has full support for c++14:
- gcc: 5.0 (https://gcc.gnu.org/projects/cxx-status.html#cxx14)
- clang: 3.4 (https://clang.llvm.org/cxx_status.html)
- intel: 17.0 
   - supports everything except N3664: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3664.html
- ibm xl c/c++: 16.1 
- Visual Studio: VS 2017, 15.7 (https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=vs-2019)

From a SEMS standpoint, it looks like we are in good shape.

To do:
* [ ] Discussion with SART teams. Any blockers?
* [ ] Check on SEMS module support for all compilers
* [ ] Set up test builds 
* [ ] Switch PR testing to c++14 
* [ ] ?

@jwillenbring @bartlettroscoe @jmgate @crtrott @mwglass @maherou @prwolfe @trilinos/framework

mhoemmen:
The last check box is definitely "profit" -- it would be pleasant to have auto lambdas and not to have to reimplement `make_unique`.

mwglass:
Here’s a short list of where a correct C++11 program can break under a C++14 compliant compiler:

- Digit Separators

The digit separator was introduced so that one could, in a more readable manner, write numeric literals and split them up in a more natural way.

- Sized Deallocation

C++14 introduces the opportunity to declare a global overload of operator delete suitable for sized deallocation, something which wasn't possible in C++11.

- constexpr member-functions, no longer implicitly const

There are many changes to constexpr in C++14, but the only change that will change semantics between C++11, and C++14 is the constantness of a member-function marked as constexpr.

- Removal of std::gets

std::gets has been removed from the Standard Library because it is considered dangerous.

swbova:
[...]

PR #6814, opened 2020-02-11
cmake: Handle C++ standard selection and propagation via CMake features

bradking:
C++11 is now required by Trilinos. Enable support unconditionally and drop the `Trilinos_ENABLE_CXX11` option. Use CMake's standard settings for specifying the C++ standard and propagating it to dependents. Avoid hard-coding -std= flags, as doing so breaks CMake's model of the compiler's C++ standard modes.

See the driver for this in:

* https://gitlab.kitware.com/snl/project-1/issues/131

Cc: @bartlettroscoe


bradking:
See TriBITSPub/TriBITS#300 and kokkos/kokkos#2749 for the upstream TriBITS and kokkos changes currently included here.

cmake/CallbackSetupExtraOptions.cmake
```diff
@@ -3,29 +3,6 @@
 # TrilinosCreateClientTemplateHeaders.cmake
 SET(CMAKE_MODULE_PATH  ${CMAKE_MODULE_PATH} "${Trilinos_SOURCE_DIR}/cmake")
 
-MACRO(TRILINOS_DISABLE_PACKAGE_REQUIRING_CXX11  CXX11_PACKAGE_NAME_IN)
```
bartlettroscoe: Right, since C++11 has been required for a long time, this macro should have been deleted a long time ago.

cmake/tribits/core/package_arch/TribitsDefineStandardCompileVars.cmake
```diff
@@ -108,14 +108,8 @@ MACRO(TRIBITS_DEFINE_STANDARD_COMPILE_FLAGS_VARS  ENABLE_SHADOWING_WARNINGS)
   IF (${PROJECT_NAME}_CXX_STRONG_COMPILE_WARNING_FLAGS)
     SET(CXX_STRONG_COMPILE_WARNING_FLAGS ${${PROJECT_NAME}_CXX_STRONG_COMPILE_WARNING_FLAGS})
   ELSE()
-    IF (NOT ${PROJECT_NAME}_ENABLE_CXX11)
-       SET(CXX_98_OPTION " -std=c++98") # C++98 standard code
```
bartlettroscoe: Right, this was to enforce the C++98 standard for code we wanted to make sure did not allow to skip in C+11 features.  No need for that anymore.  In fact, this should have been removed a long time ago when Trilinos was made to require C++11.

cmake/tribits/core/package_arch/TribitsDefineStandardCompileVars.cmake
```diff
@@ -108,14 +108,8 @@ MACRO(TRIBITS_DEFINE_STANDARD_COMPILE_FLAGS_VARS  ENABLE_SHADOWING_WARNINGS)
...
-    ELSE()
-       SET(CXX_98_OPTION)
-    ENDIF()
     MULTILINE_SET(CXX_STRONG_COMPILE_WARNING_FLAGS
       ${COMMON_STRONG_COMPILE_WARNING_FLAGS}
-      "${CXX_98_OPTION}"
```
bartlettroscoe: Was dumb to add `-std=c++98` to warning flags.  That was just bad programming on my part :-(

cmake/tribits/core/package_arch/TribitsGlobalMacros.cmake
```diff
@@ -254,12 +254,9 @@ MACRO(TRIBITS_DEFINE_GLOBAL_OPTIONS_AND_DEFINE_EXTRA_REPOS)
...
 
-  IF ("${${PROJECT_NAME}_ENABLE_CXX11_DEFAULT}" STREQUAL "")
-    SET(${PROJECT_NAME}_ENABLE_CXX11_DEFAULT OFF)
-  ENDIF()
-  ADVANCED_OPTION(${PROJECT_NAME}_ENABLE_CXX11
-    "Enable the C++11 compiler options and related code (see ${PROJECT_NAME}_CXX11_FLAGS)"
-    ${${PROJECT_NAME}_ENABLE_CXX11_DEFAULT} )
+  # Hard-code a variable with the same name as a now-removed option that is always enabled.
+  # This can be removed after clients have been updated.
+  SET(${PROJECT_NAME}_ENABLE_CXX11 ON)
```
bartlettroscoe: I wonder if this is really needed.  With the Trilinos 13.0 release coming up, it might be better to just remove this.  This might be confusing when Trilinos requires C++14 (see #6260) to have a variable `Trilinos_ENABLE_CXX11` set to `ON`.

bartlettroscoe: On second thought, it is better to keep this in for now.  I am just wondering how we can get rid of this variable later in client code so they will not be confused when Trilinos requires C++14 support.

packages/stokhos/CMakeLists.txt
```diff
@@ -30,12 +30,8 @@ TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_Ensemble_Reduction
...
   OFF )
 
-TRIBITS_ADD_OPTION_AND_DEFINE(
-  Stokhos_ENABLE_CXX11
-  HAVE_STOKHOS_CXX11
-  "Enable C++11 support in Stokhos."
-  "${${PROJECT_NAME}_ENABLE_CXX11}"
-  )
+# FIXME: Remove this hard-coded setting.  We always have C++11 now.
+SET(HAVE_STOKHOS_CXX11 ON)
```
bartlettroscoe: Right, someone should refactor Stokhos to not require this var (which I would assume is turned into a macro in a configured header file).

packages/teuchos/core/src/CMakeLists.txt
```diff
@@ -26,9 +26,8 @@ IF (${PROJECT_NAME}_ENABLE_${PARENT_PACKAGE_NAME}ParameterList)
   SET(HAVE_TEUCHOS_PARAMETERLIST TRUE)
 ENDIF()
 
-# Have C++11?
-ASSERT_DEFINED(${PROJECT_NAME}_ENABLE_CXX11)
-SET(HAVE_TEUCHOSCORE_CXX11 ${${PROJECT_NAME}_ENABLE_CXX11})
+# FIXME: Remove this hard-coded setting.  We always have C++11 now.
+SET(HAVE_TEUCHOSCORE_CXX11 ON)
```
bartlettroscoe: Right, we should remove this and eliminate the need for a macro of this name in the configured header file.

packages/teuchos/core/test/MemoryManagement/CMakeLists.txt
```diff
@@ -4,15 +4,8 @@
...
-  (${PACKAGE_NAME}_ENABLE_Boost AND HAVE_TEUCHOSCORE_BOOST_IS_POLYMORPHIC)
-  OR ${PROJECT_NAME}_ENABLE_CXX11
-  )
-  SET(HAS_TEUCHOS_GET_BASE_OBJ_VOID_PTR  TRUE)
-ELSE()
-  SET(HAS_TEUCHOS_GET_BASE_OBJ_VOID_PTR  FALSE)
-ENDIF()
-
+# FIXME: Remove this hard-coded setting.  We always have C++11 now.
+SET(HAS_TEUCHOS_GET_BASE_OBJ_VOID_PTR  TRUE)
```
bartlettroscoe: Here is another var that should be removed and taken out of the configured header file now that C++11 is required.

bartlettroscoe:
This is part of a refactoring to allow EMPIRE to use CMake 3.15+ which changed behavior w.r.t. C++ standards.  See:

* https://gitlab.kitware.com/cmake/cmake/issues/19917
* https://cee-gitlab.sandia.gov/EM-Plasma/EMPIRE/issues/1110

But we should wait until SPARC and EMPIRE update Trilinos before merging this.  (They have not been able to update since the Kokkos 2.99 promotion on 2020/02/03 which broke Trilinos for them.)

bartlettroscoe:
@bradking, I tested out this branch with SPARC and I got the build error:

```
/sierra/sntools/SDK/compilers/gcc/4.9.3-RHEL6/lib/gcc/x86_64-unknown-linux-gnu/4.9.3/../../../../include/c++/4.9.3/bits/c++0x_warning.h:32:2: error: This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
#error This file requires compiler and library support for the \
 ^
```

This is with clang-5.0.1.

bartlettroscoe:
@bradking, it looks like the fix is to set:

```
set(CMAKE_CXX_STANDARD 11 CACHE STRING "Set the C++ standard (11, 14, 17, etc.)")
```

in the customer APP's CMakeLists.txt file.  That means this breaks backwards compatibility.

bradking:
Hmm.  Trilinos was pushing `-std=` flags on clients before and now it is not.  Instead we are relying on the `cxx_std_11` compile feature to propagate to client targets that link Trilinos targets in order to make sure the compiler runs in a mode that can handle C++11 constructs in Trilinos headers.  However, that will work *only for client targets that link Trilinos targets*.  Before the pushed `-std=` flag would apply to *all targets in the client* even if they do not link anything from Trilinos.

Is the failure in SPARC on a target that does not link Trilinos?

To retain compatibility in this case we can consider having `TrilinosConfig.cmake` set `CMAKE_CXX_STANDARD` on behalf of its clients, but with an option to turn it off.

bartlettroscoe:
> However, that will work only for client targets that link Trilinos targets

@bradking, yes, on more careful inspection, that does appear to be the case.   The object file and library in question does not appear to be getting liked against any Trilinos libraries.

I will be documenting details in:

* https://sems-atlassian-srn.sandia.gov/browse/ATDV-303

Adding:

```
set(CMAKE_CXX_STANDARD 11 CACHE STRING "Set the C++ standard (11, 14, 17, etc.)")
```

does seem to fix the problem and does seem like the right thing to do anyway.

bradking:
Yes, setting `CMAKE_CXX_STANDARD` in the project's code is correct and should work with older versions of Trilinos too.  One could argue that was an existing bug in the corresponding project since they were not telling CMake about the standard requirements of their code.

FYI, `CMAKE_CXX_STANDARD` does not necessarily have to be a cache entry.

bartlettroscoe:
I completed some pretty detailed testing with SPARC as documented in:

* https://sems-atlassian-srn.sandia.gov/browse/ATDV-305

and if SPARC merges the MR:

* https://cee-gitlab.sandia.gov/sparc/sparc-ear99/merge_requests/309

which explicitly sets `CMAKE_CXX_STANDARD=11`, then SPARC builds runs fine with clang-5.0.1, gnu-7.2.0, intel-19.0.3, and cuda/nvcc-10.1.

Next I will test EMPIRE with this branch and the MR:

* https://cee-gitlab.sandia.gov/EM-Plasma/EMPIRE/merge_requests/951

for a few different compilers including CUDA and with CMake 3.16.5 just to make sure.

I am also taking off the `AT: WIP` label to allow the Trilinos PR builds to run on this to see what that looks like.

bartlettroscoe:
I put the `AT: WIP` label to try to stop the PR tester from spamming each day.

bartlettroscoe:
Let's wait for the Kokkos 3.1 update get merged in PR #7172 and then we will fix up the Kokkos changes with that and work on merging this.

bradking:
@bartlettroscoe with the Kokkos 3.1 update merged, I've rebased this branch.

bartlettroscoe:
I have finished testing this against SPARC as described in [ATDV-348](https://sems-atlassian-srn.sandia.gov/browse/ATDV-348?focusedCommentId=58330&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-58330).

I will test against EMPIRE but I think this is ready to merge.

bartlettroscoe:
Once Trilinos PR testing comes back online (hopefully tomorrow at some point), the Trilinos PR tester will test this and it will merge (if testing passes).  I the meantime, I will get this running in ATDM Trilinos testing.

bartlettroscoe:
FYI: The matching Kokkos changes were just merged to Kokkos 'develop' in:

* https://github.com/kokkos/kokkos/pull/2749

Issue #1194, opened 2017-03-29
Tpetra::MultiVector: cublas{D,S}gemm interface change?


mhoemmen:
@trilinos/tpetra

@mwglass reported build errors in tpetra/core/src/kokkos_refactor/Kokkos_MV_GEMM.hpp, relating to a wrong first argument for cublasSgemm and cublasDgemm.  It looks like he is seeing a different cuBLAS header file than we are.  This could be an artifact of how he builds Trilinos, but it reminds me that we need to check whether there are interface changes coming for those cuBLAS functions that could break the build in the near future.

Subtasks:
- [x] Research interface changes and associated cuBLAS version numbers for these functions
- [x] If needed, implement a work-around in Tpetra
- [x] Discuss a better interface in KokkosKernels for GEMM functionality (#1195)

mwglass:
There is a cublas.h and a cublas_v2.h file in the cuda include dir. The v2 file has a cublasHandle_t argument add at the beginning of the arg list. We must be picking up that header somewhere.

mhoemmen:
Thanks @mwglass !  The cuBLAS manual says that "Starting with version 4.0, the cuBLAS Library provides a new updated API, in addition to the existing legacy API. . . . The new cuBLAS library API can be used by including the header file 'cublas_v2.h'." 

http://docs.nvidia.com/cuda/cublas/

The manual also says: "In general, new applications should not use the legacy cuBLAS API, and existing existing applications should convert to using the new API if it requires sophisticated and optimal stream parallelism or if it calls cuBLAS routines concurrently from multiple threads."

This suggests that Tpetra needs to switch to the new interface in cublas_v2.h.  Would that fix this issue for you?

mhoemmen:
@mwglass -- another approach would be for Tpetra to hide the 'cublas.h' header file include in a .cpp file.  This should be quicker to do than what I described in my previous comment.  I will try that first.  Later we can go back and use the new API.

I would prefer to defer using the new API, because it requires global initialization and finalization of a context object.  I'll have to be careful to see how that interacts with `Kokkos::Cuda::finalize`, in particular, whether it needs to be called before `Kokkos::Cuda::finalize`.  See https://github.com/kokkos/kokkos/issues/714 for a Kokkos feature request that could make this easier.

mwglass:
Thanks for the update. I will also see if I can track down how we are picking up the new api header file in our builds.

mhoemmen:
[...]

PR #6823, opened 2020-02-12
Tacho trisolve improvement

kyungjoo-kim:
## Motivation

Tasking is a bit expensive when task granularity is small when the overhead matters. This code solves a supernodal triangular system resulting from Tacho Cholesky factorization. As Tacho already perform symbolic factorization and supernodal analysis, implementation of level set parallel solve is straight forward and efficient.

@mhoemmen may raise a concern that I expose cublas headers. This code in this PR is intended to be used by downstream applications (so it is a bit different situation from the other example code that you saw in my previous PR). For blas fortran interface, I was able to hide the interface using a wrapper interface. However, I cannot hide cublas interface as it requires cuda and cublas infrastructure. To hide the details, I am going to do ETI and a separate PR for ETI will be submitted later.


mhoemmen:
@kyungjoo-kim wrote:
> However, I cannot hide cublas interface as it requires cuda and cublas infrastructure.

It's definitely possible to hide cuBLAS header files:

1. Wrap cuBLAS handle in an opaque wrapper, with a function to get a pointer to it.  If you don't actually call cuBLAS, then you only need [a forward declaration of the wrapper class, and a declaration of the function that returns a pointer to it](https://github.com/trilinos/Trilinos/blob/master/packages/tpetra/tsqr/src/Tsqr_Impl_CuBlasHandle_fwd.hpp).
2. Declare [wrappers for cuBLAS functions that take a pointer or reference to the opaque wrapper class](https://github.com/trilinos/Trilinos/blob/master/packages/tpetra/tsqr/src/Tsqr_Impl_CuBlas.hpp).
3. Actual calls to cuBLAS functions [happen in a .cpp file](https://github.com/trilinos/Trilinos/blob/master/packages/tpetra/tsqr/src/Tsqr_Impl_CuBlas.cpp).
4. Fall-back implementations for types that cuBLAS does not support don't need to call cuBLAS, so they don't need to include cuBLAS headers.

I don't mean to give you a hard time, so I won't block this PR.  However, I really think kokkos-kernels could fix this.  It's not hard; it just takes time.

kyungjoo-kim:
During PT tests, gnu compilers complain about template struct inside of struct. It needs to know it is template.

packages/shylu/shylu_node/tacho/src/Tacho_Util.hpp
```diff
@@ -30,6 +30,11 @@
 #include "mkl.h"
 #endif
 
+#if defined(KOKKOS_ENABLE_CUDA) 
+#include "cublas.h"
```
ndellingwood: @kyungjoo-kim is this safe to mix/match the two cublas version headers?

mhoemmen: @ndellingwood Exposing both header files to the same compilation unit caused a build error once in Sierra: https://github.com/trilinos/Trilinos/issues/1194 .

kyungjoo-kim: Good catch. This compiles because I include both headers sequentially. So, all the v2 functions are available and the same core function names are redefined with v2 version functions. As long as I use v2 APIs in the code, it works. I will remove the cublas.h and v2 (new) version is the right one that we are supposed to use.

kyungjoo-kim: @ndellingwood would you review again  ?

ndellingwood: From the comments here https://github.com/trilinos/Trilinos/commit/d47d1d09323fc3e7ab6feecdc673506503142976 it seems the removal of the "old" API (hiding away in cpp file) fixed compilation with applications using the newer cublas header; the recent commit removes the older header known to cause issues, @kyungjoo-kim comment says that a follow up PR will contain ETI changes to properly hide headers in cpp files.

PR #6824, opened 2020-02-12
Create PR test for GCC 8.3.0 w/WError flag set

ZUUL42:
@trilinos/framework 

## Motivation

Creating new PR testing build for GCC 8.3.0 that will eventually replace the GCC 7.2.0 build

Moves Issue #6295 forward.

## Testing

This will have no effect as is until the PR testing set script is updated.
It will currently build and pass all tests as set.


cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -193,6 +193,22 @@ def setBuildEnviron(arguments):
                      'sems-cmake/3.10.3',
                      'atdm-env',
                      'atdm-ninja_fortran/1.7.2'],
+                'Trilinos_pullrequest_gcc_8.3.0':
```
prwolfe: We should have tests added for this change in unit_tests/TestPullRequestLinuxDriverTest.py. Otherwise looks good

ZUUL42: Ahh. Right. Addition pushed.

PR #6831, opened 2020-02-14
Tpetra: Fix #6663

### LINKED ISSUE #6663: Tpetra::CrsMatrix::unpackAndCombine, non-static-graph path, allocates too much memory

mhoemmen:
@trilinos/tpetra @tjfulle 

## Motivation

Tpetra::CrsGraph currently may use too much memory when it is the target of a doExport or doImport.  Likewise, Tpetra::CrsMatrix currently may use too much memory when it owns its graph (has a "non-static" CrsGraph) and is the target of a doExport or doImport.

Due to removal of DynamicProfile, graphs and matrices now only have the StaticProfile option, which stores data in a "row offsets begin, number of entries in row" 4-array (3-array for graphs) variant of compressed sparse row storage.  This means that `copyAndPermute` and `unpackAndCombine` must reallocate storage before inserting incoming new graph / matrix entries.

Before this PR, `copyAndPermute` and `unpackAndCombine` would allocate space for all the incoming entries in each row, without considering that they might merge with existing target graph / matrix entries in that row.  (They would "merge" because they have the same row and column indices.)  That meant that the target graph / matrix might allocate much more storage than it needs.

This PR fixes that issue, by considering merges when computing the new allocation sizes needed in `copyAndPermute` and `packAndPrepare`.

## Related Issues

* Closes #6663 
* Replaces #6749, #6806 
* Related to #6648 

## Stakeholder Feedback

See #6648 and #6663 for feedback and lots of diagnosis help from Sierra SD developers.  I tested this in situ and the SD test in question successfully ran to completion with these changes (not counting some extra debug checks and verbose output).

## Testing

* Added new Tpetra tests for new functionality
* Tested on Mac Clang + OpenMPI
* Tested in SD (Clang debug + OpenMPI)


mhoemmen:
Just fixed a build warning.

mhoemmen:
@tjfulle Thanks!!! :-D

PR #6851, opened 2020-02-18
DO NOT MERGE Zoltan2: Hybrid MPI+Kokkos distributed distance-1 coloring algorithms

ibogle:
@trilinos/zoltan2 

## Motivation

This change introduces hybrid distributed distance-1 coloring implementations to Zoltan2. 

## Testing

Both included coloring algorithms are tested by an executable in the zoltan2/test/color/ directory. Test inputs include a small mesh-like input, the commanche_dual sparse matrix, an 8 vertex complete graph, and a graph with no edges. Old tests of the existing non-distributed coloring are untouched and pass. New tests and validation test our distributed coloring under normal circumstances, and check that our implementation can handle distributions where certain processes get zero vertices.


packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
```diff
@@ -0,0 +1,458 @@
+#ifndef _ZOLTAN2_ALGHYBRIDGMB_HPP_
+#define _ZOLTAN2_ALGHYBRIDGMB_HPP_
+
+#include <vector>
+#include <unordered_map>
+#include <iostream>
+#include <queue>
+#include <unistd.h>
```
mhoemmen: This is not a Standard C++ header file.  People like to include this file to get the definition of `NULL`.  However, we have C++11 now; use `nullptr` instead.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
```diff
@@ -0,0 +1,458 @@
...
+
+#include <Zoltan2_Algorithm.hpp>
```
mhoemmen: Use double quotes for headers that are not system headers.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
```diff
@@ -0,0 +1,458 @@
...
+//!        and Boman
+
+namespace Zoltan2 {
+
+template <typename Adapter>
+class AlgHybridGMB : public Algorithm<Adapter>
+{
+  public:
+  
+    typedef typename Adapter::lno_t lno_t;
```
mhoemmen: We have C++11 now; use `using`, not `typedef`.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
```diff
@@ -0,0 +1,458 @@
...
+                       Teuchos::RCP<femv_t> femv,
+                       bool recolor=false){
+      
+      //default values are taken from KokkosKernels_TestParameters.hpp
+      
+      int algorithm = pl->get<int>("Hybrid_algorithm",0);
+      int repeat = pl->get<int>("Hybrid_repeat",1); //probably not using this 
+                                                    //until we test performance
+      int chunk_size = pl->get<int>("Hybrid_chunk_size",-1); 
+      int shmemsize = pl->get<int>("Hybrid_shmemsize", 16128);
```
mhoemmen: Where do these magic constants come from?

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
```diff
@@ -0,0 +1,458 @@
...
+      //set the initial coloring of the kh.get_graph_coloring_handle() to be
+      //the data view from the femv.
+      Kokkos::View<int**, Kokkos::LayoutLeft> femvColors = femv->template getLocalView<MemorySpace>();
+      Kokkos::View<int*, Tpetra::Map<>::device_type >  sv = subview(femvColors, Kokkos::ALL, 0);
+      kh.get_graph_coloring_handle()->set_vertex_colors(sv);
+      
+      KokkosGraph::Experimental::graph_color_symbolic(&kh, nVtx, nVtx, 
+                                                      offset_view, adjs_view);
+      
+      Kokkos::fence();
```
mhoemmen: Why do you need a fence here?

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
```diff
@@ -0,0 +1,458 @@
...
+      } 
+      
+      //Create random numbers seeded on global IDs so that we don't
+      //need to communicate for consistency. These numbers determine
+      //which vertex gets recolored in the event of a conflict.
+      //taken directly from the Zoltan coloring implementation 
+      std::vector<int> rand(finalGIDs.size());
+      for(size_t i = 0; i < finalGIDs.size(); i++){
+        Zoltan_Srand((unsigned int) finalGIDs[i], NULL);
+        rand[i] = (int) (((double) Zoltan_Rand(NULL)/(double) ZOLTAN_RAND_MAX)*100000000);
```
mhoemmen: 1. Use C++ - style casts, not C-style casts.
2. Don't use `NULL`; use `nullptr` instead.
3. Consider factoring out this unpleasant expression into a separate function.
4. Don't use magic numbers like `100000000`.  Give them names at least.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
```diff
@@ -0,0 +1,458 @@
...
+                   Teuchos::ArrayRCP<int> colors, Teuchos::RCP<femv_t> femv,
+                   std::vector<gno_t> reorderGIDs,
+                   std::vector<int> rand,
+                   Teuchos::RCP<Teuchos::Time> timeInterior,
+                   Teuchos::RCP<Teuchos::Time> timeBoundary){
+     Teuchos::RCP<Teuchos::Time> 
+             timeBoundaryComm(Teuchos::TimeMonitor::getNewTimer("03 BOUNDARY-COMM")),
+             timeBoundaryComp(Teuchos::TimeMonitor::getNewTimer("04 BOUNDARY-COMP"));
+      
+      timeInterior->start();
```
mhoemmen: Use `Teuchos::TimeMonitor` instead of calling start/stop explicitly.

packages/zoltan2/test/color/ReadMatrixFromFile.hpp
```diff
@@ -0,0 +1,325 @@
...
+  broadcast (*pComm, rootRank, 1, &globalNumNonzeros);
+
+  //std::cout << globalNumRows << " " << globalNumNonzeros << std::endl;
+
+  global_ordinal_type *rowPtr;
+  global_ordinal_type *colInd;
+
+  if (myRank == rootRank) {
+
+    rowPtr = new global_ordinal_type[globalNumRows+1];
```
mhoemmen: Never use `new T[N]` without at least wrapping it in `unique_ptr` or `shared_ptr`.  You can use `unique_ptr` here:
```cpp
std::unique_ptr<global_ordinal_type> rowPtr;
if (myRank == rootRank) {
  rowPtr = std::unique_ptr<global_ordinal_type>(new global_ordinal_type[globalNumNonzeros]);
}
```

packages/zoltan2/test/color/coloring1.cpp
```diff
@@ -105,6 +111,43 @@ int validateColoring(RCP<SparseMatrix> A, int *color)
   return nconflicts;
 }
 
+int validateDistributedColoring(RCP<SparseMatrix> A, int *color, int rank){
```
mhoemmen: The input `A` can never be null, so just pass in a reference (`const SparseMatrix&`) instead of an RCP.

packages/zoltan2/test/color/coloring1.cpp
```diff
@@ -136,9 +179,9 @@ int checkBalance(zlno_t n, int *color)
     }
   }
 
-  std::cout << "Color size[0:2] = " << colorCount[0] << ", " << colorCount[1] << ", " << colorCount[2] << std::endl;
-  std::cout << "Largest color class = " << largest << " with " << colorCount[largest] << " vertices." << std::endl;
-  std::cout << "Smallest color class = " << smallest << " with " << colorCount[smallest] << " vertices." << std::endl;
+  //std::cout << "Color size[0:2] = " << colorCount[0] << ", " << colorCount[1] << ", " << colorCount[2] << std::endl;
```
mhoemmen: Delete these, please.

packages/zoltan2/test/color/coloring1.cpp
```diff
@@ -197,31 +255,97 @@ int main(int narg, char** arg)
...
-
   if (me == 0)
     std::cout << "NumRows     = " << Matrix->getGlobalNumRows() << std::endl
          << "NumNonzeros = " << Matrix->getGlobalNumEntries() << std::endl
          << "NumProcs = " << comm->getSize() << std::endl;
 
+  if (prepartition != "") {
+
+    // Compute new partition of matrix
+    SparseMatrixAdapter *zadapter;
```
mhoemmen: We have `std::unique_ptr`.  Don't use raw pointers; use `unique_ptr`.

packages/zoltan2/test/color/coloring1.cpp
```diff
@@ -231,19 +355,19 @@ int main(int narg, char** arg)
...
   {
   Zoltan2::ColoringProblem<SparseMatrixAdapter> problem(&adapter, &params);
-  std::cout << "Going to color" << std::endl;
+  if(comm->getRank()==0) std::cout << "Going to color" << std::endl;
   problem.solve();
 
   ////// Basic metric checking of the coloring solution
   size_t checkLength;
-  int *checkColoring;
+  int *checkColoring = NULL;
```
mhoemmen: More raw pointers.  Raw pointers are bad.

packages/zoltan2/test/color/coloring1.cpp
```diff
@@ -258,29 +382,45 @@ int main(int narg, char** arg)
...
+  
   } catch (std::exception &e){
       std::cout << "Exception caught in coloring" << std::endl;
       std::cout << e.what() << std::endl;
       std::cout << "FAIL" << std::endl;
       return 0;
   }
-
+  int numGlobalConflicts = 0;
+  Teuchos::reduceAll<int,int>(*comm, Teuchos::REDUCE_MAX,1,&testReturn,&numGlobalConflicts);
```
mhoemmen: This overload may have been deprecated.  Use this instead:
```cpp
Teuchos::reduceAll(*comm, Teuchos::REDUCE_MAX, testReturn, Teuchos::outArg(numGlobalConflicts));
```

packages/kokkos-kernels/src/graph/KokkosGraph_Distance1Color.hpp
```diff
@@ -74,8 +74,16 @@ void graph_color_symbolic(
...
-  color_view_type colors_out = color_view_type("Graph Colors", num_rows);
-
+  //printf("GraphColorHandle's colorView reference count: %d, and label: %s",gch->get_vertex_colors().use_count(),
+  //                                                                         gch->get_vertex_colors().label());
+  
+  color_view_type colors_out;// = color_view_type("Graph Colors", num_rows);
+  if(gch->get_vertex_colors().use_count() > 0){
+    colors_out = gch->get_vertex_colors();
+  } else {
+    colors_out = color_view_type("Graph Colors", num_rows);
```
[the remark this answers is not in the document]
ibogle: As far as I know, yes. This part of the file checks to see whether or not the view passed in (via gch->get_vertex_colors()) is initialized. If it is, we use that. If not, the view needs to be initialized. Previously, the view was always just initialized, which didn't allow for passing in a partial coloring.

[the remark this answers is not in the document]
mhoemmen: I mean, why not do this?
```cpp
using Kokkos::view_alloc;
colors_out = color_view_type(view_alloc("Graph Colors", WithoutInitializing), num_rows);
```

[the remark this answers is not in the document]
ibogle: The initialization is not something I added to this file, I just added the if/else surrounding the initialization. I'm not familiar enough with the KokkosKernels coloring codebase to say confidently that this will not break anything, so I don't think I should make this change.

packages/zoltan2/test/color/coloring1.cpp
ibogle: This variable is used to store the returned value of the Zoltan2::ColoringSolution class's getColors() function.

packages/zoltan2/test/color/coloring1.cpp
ibogle: I've been told that this call matches the usage across Zoltan2, and it is not deprecated.

packages/kokkos-kernels/src/graph/KokkosGraph_Distance1Color.hpp
```diff
@@ -74,8 +74,16 @@ void graph_color_symbolic(
 
   gch->set_tictoc(handle->get_verbose());
 
-  color_view_type colors_out = color_view_type("Graph Colors", num_rows);
-
+  //printf("GraphColorHandle's colorView reference count: %d, and label: %s",gch->get_vertex_colors().use_count(),
```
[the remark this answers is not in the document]
ibogle: Done.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
ibogle: The underlying cause of this include has been removed, so I have removed the include.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
ibogle: Done.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
ibogle: Switched all instances of typedef to using.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
ibogle: The defaults were taken from KokkosKernels_TestParameters.hpp, but it looks like the call succeeds without setting these parameters, so I will remove them.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
ibogle: I don't think I do. My GPU-based platform ran tests just fine without it, so I have removed this fence.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
ibogle: This random number generation has been replaced with calls to std::srand and std::rand. I think all of these issues are resolved.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
ibogle: I am not certain how to achieve the same effect using Teuchos::TimeMonitor without negatively impacting readability, so I just removed the timers.

packages/zoltan2/test/color/ReadMatrixFromFile.hpp
ibogle: Done.

packages/zoltan2/test/color/coloring1.cpp
ibogle: Done.

packages/zoltan2/test/color/coloring1.cpp
ibogle: Done.

packages/zoltan2/test/color/coloring1.cpp
ibogle: Done

packages/zoltan2/test/color/coloring1.cpp
mhoemmen: @ibogle wrote:
> This variable is used to store the returned value of the `Zoltan2::ColoringSolution` class's `getColors()` function.

Ah, so be it then.  If `getColors()` returns a new allocation that the caller is responsible for deallocating, then it should actually return `std::unique_ptr<int*>`, not `int*`.

packages/zoltan2/src/algorithms/color/Zoltan2_Distance1_2GhostLayer.hpp
```diff
@@ -0,0 +1,521 @@
...
+        ownedPlusGhosts[i+offsets.size()-1] = sendbuf[i];
+      }
+      
+      //communicate GIDs to owners
+      Teuchos::ArrayView<int> sendcounts_view = Teuchos::arrayViewFromVector(sendcounts);
+      Teuchos::ArrayView<gno_t> sendbuf_view = Teuchos::arrayViewFromVector(sendbuf);
+      Teuchos::ArrayRCP<gno_t>  recvbuf;
+      std::vector<int> recvcounts(comm->getSize(),0);
+      Teuchos::ArrayView<int> recvcounts_view = Teuchos::arrayViewFromVector(recvcounts);
+      Zoltan2::AlltoAllv<gno_t>(*comm, *env, sendbuf_view, sendcounts_view, recvbuf, recvcounts_view);
```
mhoemmen: Is this a slow reimplementation of Distributor?  Epetra and Tpetra have plenty of functions for doing things like "figuring out what process owns what GIDs."  `Zoltan2::AlltoAllv` will have every process send a message to every other process (!!!).

packages/zoltan2/test/color/ReadMatrixFromFile.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+    Teuchos::send (*pComm, myNumRows, rootRank);
+    
+    if (myNumRows != 0) {
+
+      Teuchos::send (*pComm, static_cast<int> (myNumRows), myRows.getRawPtr(), rootRank);      
+      Teuchos::receive (*pComm, rootRank, static_cast<int> (myNumRows), myNumEntriesPerRow.getRawPtr());
+
+      const local_ordinal_type myNumEntries =
+	std::accumulate (myNumEntriesPerRow.begin(),
```
mhoemmen: This is fine for now, but be aware that `std::accumulate`'s internal accumulation type is the type of the third argument, which is `int` here (the type of `0`).  That would be bad if you were summing up 64-bit integers, for example.

packages/zoltan2/test/color/ReadMatrixFromFile.hpp
```diff
@@ -0,0 +1,319 @@
+#ifndef __READMATRIXFROMFILE_HPP
+#define __READMATRIXFROMFILE_HPP
+#include<vector>
+template <typename global_ordinal_type, typename local_ordinal_type, typename scalar_type, typename map_type>
+void
+distribute (Teuchos::ArrayRCP<size_t>& myNumEntriesPerRow,
```
mhoemmen: We should think about having a common interface for binary and ASCII sparse matrix I/O, to avoid code duplication with Tpetra.

packages/zoltan2/test/temp/readSymmetricMatrixMarket.cpp
```diff
@@ -0,0 +1,115 @@
+
```
mhoemmen: Why does Zoltan2 have a "temp" directory?

packages/zoltan2/src/algorithms/color/Zoltan2_Distance1_2GhostLayer.hpp
ibogle: The purpose of this AlltoAllv is to figure out which owned vertices are ghosted on which processes. The ghost owners send ghosted GIDs back to their owners (which I determine the Tpetra way). The owners then send  degree and adjacency information back to the ghost owners. If there is a more efficient way of determining where owned vertices are ghosted from the owning process, that would be preferable over this.

packages/zoltan2/test/color/ReadMatrixFromFile.hpp
ibogle: I will make sure to check that that doesn't pose any potential issues.

packages/zoltan2/src/algorithms/color/Zoltan2_Distance1_2GhostLayer.hpp
mhoemmen: @ibogle You can do this with Import and Export on a CrsGraph whose rows are the GIDs, and whose (global) column indices are MPI process ranks. 

1. Construct a Map, `rowMapAll`, whose indices are all the GIDs (owned and ghosted) on each process.
2. Construct a Map, `rowMapOwned`, whose indices are the owned GIDs on each process.
3. Construct an Import object: `Import<...> importer(rowMapOwned, rowMapAll);`.
4. Construct a CrsGraph, `graphOwned`, whose row Map is `rowMapOwned`.
5. On each process p, insert p as a global column index once into each row of `graphOwned`.
6. Construct a CrsGraph, `graphAll`, whose row Map is `rowMapAll`.
7. Import to populate `graphAll`: `graphAll.doImport(graphOwned, importer, INSERT);`

Now global row `gid` of `graphAll` has as its (global) column indices, all the ranks of the MPI processes that either ghost or own `gid`.  At this point, you can populate other data structures and do a reverse-mode Export (with `importer`) to send information back to the owners.

kddevin: @ibogle Let's discuss this suggestion at our meeting next week before you dive in to this change.

packages/zoltan2/test/color/ReadMatrixFromFile.hpp
mhoemmen: @ibogle Canonical way would be to cast the initial value to `local_ordinal_type`.
```cpp
const local_ordinal_type result =
  std::accumulate(x.begin(), x.end(), local_ordinal_type(0));
```

packages/zoltan2/src/problems/Zoltan2_ColoringProblem.hpp
```diff
@@ -135,9 +135,17 @@ class ColoringProblem : public Problem<Adapter>
   {
     RCP<Teuchos::StringValidator> color_method_Validator = Teuchos::rcp(
       new Teuchos::StringValidator(
-        Teuchos::tuple<std::string>( "SerialGreedy" )));
+        Teuchos::tuple<std::string>( "SerialGreedy","Hybrid","2GL" )));
     pl.set("color_method", "SerialGreedy", "coloring algorithm",
      color_method_Validator);
+    pl.set("Hybrid_batch_size",-1,"Batch size for distributed coloring, default is all verts",
```
egboman: Is batch size here the same as superstep size in the old Zoltan coloring? Please add a comment in the code to clarify.

packages/zoltan2/src/problems/Zoltan2_ColoringProblem.hpp
```diff
@@ -201,6 +209,22 @@ void ColoringProblem<Adapter>::solve(bool newData)
       AlgSerialGreedy<Adapter> alg(this->graphModel_, this->params_,
                                    this->env_, this->comm_);
       alg.color(this->solution_);
+  } 
+  else if (method.compare("Hybrid") == 0)
+  {
+      //new stuff using Kokkos and Gebremedhin-Manne-Boman framework for hybrid architectures
+      //this->inputAdapter_ is the adapter passed to this problem
+      printf("Do the hybrid algorithm\n");
```
egboman: Is this a debug printf? The deafult behavior should be silent.

packages/zoltan2/src/problems/Zoltan2_ColoringProblem.hpp
```diff
@@ -201,6 +209,22 @@ void ColoringProblem<Adapter>::solve(bool newData)
...
+      //new stuff using Kokkos and Gebremedhin-Manne-Boman framework for hybrid architectures
+      //this->inputAdapter_ is the adapter passed to this problem
+      printf("Do the hybrid algorithm\n"); 
+      AlgHybridGMB<Adapter> alg(this->inputAdapter_, this->params_,
+                                this->env_, this->comm_);
+      alg.color(this->solution_);
+  } 
+  else if (method.compare("2GL") == 0)
+  {
+      printf("Do the two layer ghost method\n");
```
egboman: Is this a debug printf? The default behavior should be silent.

packages/zoltan2/test/color/coloring1.cpp
```diff
@@ -148,24 +184,39 @@ int main(int narg, char** arg)
...
   ////// Establish session.
   Tpetra::ScopeGuard tscope(&narg, &arg);
   Teuchos::RCP<const Teuchos::Comm<int> > comm = Tpetra::getDefaultComm();
   int me = comm->getRank();
 
+  if (me == 0) std::cout << "Starting everything" << std::endl;
   // Read run-time options.
   Teuchos::CommandLineProcessor cmdp (false, false);
+  cmdp.setOption("colorMethod", &colorAlg,
+                 "Coloring algorithms supported: SerialGreedy, Hybrid, 2GL");
```
egboman: I think 2GL is a variation of Hybrid, as it also uses hybrid MPI+Kokkos. Perhaps it should be called Hybrid2GL? (Optional, just a suggestion.)

kddevin:
It was marked "DO NOT MERGE" because I did a tentative approval to trigger PR testing but did not want the code merged until @egboman approved it.  
Those PR tests failed, so I revoked my approval.

packages/zoltan2/src/algorithms/color/Zoltan2_AlgHybridGMB.hpp
```diff
@@ -0,0 +1,76 @@
+#ifndef _ZOLTAN2_ALGHYBRIDGMB_HPP_
+#define _ZOLTAN2_ALGHYBRIDGMB_HPP_
+
+#include <Zoltan2_Algorithm.hpp>
+#include <Zoltan2_GraphModel.hpp>
+#include <Zoltan2_ColoringSolution.hpp>
+
+//////////////////////////////////////////////
+//! \file Zoltan2_AlgHybridGMB.hpp
+//! \brief A hybrid version of the framework proposed by Gebremedhin, Manne, and Boman
```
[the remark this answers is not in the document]
ibogle: I will make this change, each instance of GMB will change to GM.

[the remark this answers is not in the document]
ibogle: I also added a citation for the original paper.

packages/zoltan2/src/problems/Zoltan2_ColoringProblem.hpp
ibogle: I will clean up those arguments, as they are no longer used.

packages/zoltan2/src/problems/Zoltan2_ColoringProblem.hpp
ibogle: I have removed these printfs. They were primarily for debugging.

packages/zoltan2/test/color/coloring1.cpp
ibogle: This is a good suggestion. The current name is very long, and it doesn't indicate that it follows the Hybrid method fairly closely. I'll make this change.

ibogle:
I think I've fixed the build error from the previous tests, and I addressed the points that @egboman raised. Please let me know if any further changes are necessary.

Issue #6856, opened 2020-02-18
STK:  stk_simd and stk_math broken on GPUs


sayerhs:
## Bug Report
@trilinos/stk

### Description

We are seeing compilation errors in exawind/nalu-wind with the latest Trilinos develop branch on GPUs. [see CDash](https://my.cdash.org/viewBuildError.php?buildid=1776645). I believe it is because of the definition of `STK_MATH_FORCE_INLINE` https://github.com/trilinos/Trilinos/blob/747a3e780d399a5a5b80940c8cf089c7c0ff3559/packages/stk/stk_math/stk_math/StkMath.hpp#L43-L47 

### Steps to Reproduce
1. SHA1:  747a3e780d399a5a5b80940c8cf089c7c0ff3559

Cc: @alanw0

srajama1:
@sayerhs : Can we close this ?

spdomin:
Hi. I am reviewing recent Nalu and Nalu-Wind build/test issues. I see a common pattern or 1) commit, 2) break, 3) revert, without a subsequent push to resolve the initial push that was reverted. Top capture my statistics properly, was this ever re-commited?

alanw0:
The fix no doubt went in with a subsequent stk snapshot.

spdomin:
Good to know. I will exclude this from my data point.

PR #6857, opened 2020-02-19
Revert a change that appears to have broken stk_simd compilation for

alanw0:
cuda.

@trilinos/<teamName>


alanw0:
This should address #6856

PR #6864, opened 2020-02-19
Tempus fix imex initialization

### LINKED ISSUE #6862: Tempus:  IMEX Not Fully Constructed Through StepperFactory
## Bug Report
@trilinos/tempus 

### Description
When an IMEX stepper is constructed through the StepperFactory, it is not fully constructed so that the tableaus can be set later in the code.

Need to get the IMEX stepper fully constructed, an dneed to have some test to cover this.  Should survey other steppers and ensure that this is not happening with other steppers.

krcb:
@trilinos/tempus 

## Motivation

Addresses IMEX setup issues reported in #6862. @ccober6 you may want to see if this kind of approach needs to be applied in other places and what testing is needed.

## Related Issues

Closes #6862


ccober6:
@krcb Just pushed some changes.  Look them over, and see what you think.  I have several meetings this after, so I will likely touch base with you tomorrow morning.

krcb:
> @krcb Just pushed some changes. Look them over, and see what you think. I have several meetings this after, so I will likely touch base with you tomorrow morning.

@ccober6 Can we have versions of these tests that create the stepper with both a null parameter list and a null model?

ccober6:
Sure.  Would a test that does the following sequence cover your case?
* Constructs the IMEX Stepper with a null parameterlist and null model
* Sets the model
* Test if the IMEX Stepper is correctly initialized.

Thus the parameterlist is the default parameterlist, or would you like the parameterlist set also?  It would be a little harder to test the state of the IMEX Stepper when the model is not set, but doable.

ccober6:
@krcb @rppawlo Roger, can you review this?

krcb:
@ccober6 The test changes look good. I can't approve though, as I opened the PR. @rppawlo can you review?

packages/tempus/src/Tempus_StepperFactory.hpp
```diff
@@ -1459,56 +1496,58 @@ class StepperFactory
...
+      while ((pos != std::string::npos) || (lastPos != std::string::npos)) {
+        std::string token = str.substr(lastPos,pos-lastPos);
+        // Strip single quotes
+        std::string::size_type beg = token.find_first_of("'") + 1;
+        std::string::size_type end = token.find_last_of ("'");
+        stepperListStr.push_back(token.substr(beg,end-beg));
+
+        lastPos = str.find_first_not_of(delimiters, pos); // Skip delimiters
+        pos = str.find_first_of(delimiters, lastPos);     // Find next delimiter
+      }
```
rppawlo: Not saying we need to do this now, but I have similar code in panzer. One day we should move things like this into a teuchos string utils library.

packages/tempus/unit_test/Tempus_UnitTest_IMEX_RK.cpp
```diff
@@ -40,6 +40,8 @@ using Tempus::StepperExplicitRK;
 // Comment out any of the following tests to exclude from build/run.
 #define CONSTRUCTION
 #define STEPPERFACTORY_CONSTRUCTION
+#define STEPPERFACTORY_CONSTRUCTION_GENERAL_WO_PARAMETERLIST
+#define STEPPERFACTORY_CONSTRUCTION_GENERAL_WO_PARAMETERLIST_MODEL
```
rppawlo: Why do these defines exist? There is a runtime flag that you can use to selectively execute just a single test from an executable. Maybe this is more convenient?

packages/tempus/src/Tempus_StepperFactory.hpp
```diff
@@ -1459,56 +1496,58 @@ class StepperFactory
...
-    using Teuchos::RCP;
-    using Teuchos::ParameterList;
+    if (stepperPL != Teuchos::null) {
+      using Teuchos::RCP;
+      using Teuchos::ParameterList;
+
+      // Parse Stepper List String
+      std::vector<std::string> stepperListStr;
+      stepperListStr.clear();
+      std::string str = stepperPL->get<std::string>("Stepper List");
```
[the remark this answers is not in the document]
ccober6: No I do not think so.  I will remove it in another branch, as it does not hurt everything here.

packages/tempus/src/Tempus_StepperFactory.hpp
ccober6: Yeah, you shared it with me originally from panzer.  Putting it in teuchos would be good.

packages/tempus/unit_test/Tempus_UnitTest_IMEX_RK.cpp
ccober6: Ah, what does that command look like?  This original came from Irina, and I thought it was useful during debugging.

krcb:
@ccober6 @rppawlo looks like this timed out waiting for a job and so testing failed. Do we need to do anything to restart this?

ccober6:
I added the labels AUTOMERGE and RETEST.  Hopefully this will get it going again.

packages/tempus/unit_test/Tempus_UnitTest_IMEX_RK.cpp
rppawlo: test.exe --test=<my_test_name>

For example, I have a panzer unit test declared with:
```
  TEUCHOS_UNIT_TEST(global_data, builder)
  {
   ...
  }
```

You could run only that one with:
```
./PanzerDiscFE_global_data.exe --test=builder
```

PR #6867, opened 2020-02-19
MueLu: Add Belos & Stratimikos smoothers

cgcgcg:
@trilinos/muelu

## Motivation
Allow to use Belos and Stratimikos as smoother and coarse solver.

For this to work, I had to change Belos from being an optional test dependency to an optional dependency.

Since not all Belos solvers are exposed via Stratimikos, we want the BelosSmoother. 
Since Stratimikos allows to construct preconditioners, we want the StratimikosSmoother.


packages/muelu/src/Utils/ForwardDeclaration/MueLu_AMGXOperator_fwd.hpp
```diff
@@ -47,7 +47,7 @@
 #define MUELU_AMGXOPERATOR_FWD_HPP
 
 #include "MueLu_ConfigDefs.hpp"
-#if defined(HAVE_MUELU_EXPERIMENTAL) && defined(HAVE_MUELU_AMGX)
```
jhux2: @csiefer2 had the same change in his PR.  I guess this is script generated, but indicates that someone modified the file by hand at some point?

cgcgcg: Yup, looks that way. I ran the script to generate fwd declarations.

packages/muelu/src/Smoothers/MueLu_BelosSmoother_decl.hpp
```diff
@@ -0,0 +1,249 @@
...
+#ifndef _MSC_VER
+    // Avoid error C3772: invalid friend template declaration
+    template<class Scalar2, class LocalOrdinal2, class GlobalOrdinal2, class Node2>
+    friend class BelosSmoother;
+#endif
+
+    BelosSmoother(const std::string type, const Teuchos::ParameterList& paramList = Teuchos::ParameterList());
+
+    //! Destructor
+    virtual ~BelosSmoother() { }
```
mhoemmen:
```cpp
virtual ~BelosSmoother() = default;
```

packages/muelu/src/Smoothers/MueLu_BelosSmoother_decl.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+  private:
+
+    std::string                         type_;
+    // typedef Tpetra::Operator<Scalar>             OP;
+    // typedef Tpetra::MultiVector<Scalar>          MV;
+    typedef MultiVector          MV;
+    typedef Belos::OperatorT<MV> OP;
+
+    RCP<Belos::LinearProblem<Scalar, MV, OP> > belosProblem_;
```
mhoemmen: It would be better to dispatch to the Epetra or Tpetra specialization of Belos, instead of using Xpetra or this `Belos::OperatorT` thing.  Right now, you're forcing Trilinos to rebuild all of Belos.

cgcgcg: That should allow me to get all the solvers that are registered for Tpetra/Epetra respectively, right? Cause I noticed that some of the solvers were supposedly supported (via `supportedSolverNames`), but not via `isSupported`, which is somewhat odd behavior.

mhoemmen: @cgcgcg wrote:
> That should allow me to get all the solvers that are registered for Tpetra/Epetra respectively, right?  Cause I noticed that some of the solvers were supposedly supported (via `supportedSolverNames`), but not via `isSupported`, which is somewhat odd behavior.

Would you happen to have an example?  Both of these functions appear to check the custom factories, which would let you call things like the Tpetra-only communication-avoiding and pipelined solvers.

PR #6869, opened 2020-02-19
ATDM: ATS-2: Allow CUDA RDC to be enabled (CDOFA-72)

bartlettroscoe:
I think CUDA RDC got turned off when SPARC settings were copied over to this
file and I missed this in my review.  That was my fault.

We don't have any automated CUDA RDC builds yet so I don't know if CUDA RDC
builds will actually work on ATS-2 (with any CUDA or compiler) but there is no
reason not to let developers try.

## How was this tested:

```
$ hostname -f
vortex60.snl.gov

$ . cmake/std/atdm/load-env.sh cuda-rdc-opt
Hostname 'vortex60' matches known ATDM host 'vortex' and system 'ats2'
Setting compiler and build options for build-name 'cuda-rdc-opt'
Using ats2 compiler stack CUDA-10.1.243_GNU-7.3.1_SPMPI-2019.06.24 to build RELEASE code with Kokkos node type CUDA

$ set | grep ATDM_CONFIG_ | grep RDC
ATDM_CONFIG_CUDA_RDC=ON
```


bartlettroscoe:
@bathmatt specifically noted this was broken and asked for this.

PR #6874, opened 2020-02-20
MueLu Region cheby switched to use existing region smoother setup

pohm01:
The setup for region Chebyshev now uses the region smoother setup for the matrix diag.

@trilinos/muelu 

## Motivation
This fixes the interface scaling on the diagonal matrix used in the region Chebyshev smoother. It also reduces duplicate code in region smoothers.

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
@@ -318,7 +318,7 @@ powerMethod(RCP<Teuchos::ParameterList> params,
   // Get max number of regions per proc
   const int maxRegPerProc = regionGrpMats.size();
 
-  Teuchos::Array<RCP<Vector> > diagInv = params->get<Teuchos::Array<RCP<Vector> > >("chebyshev: inverse diagonal");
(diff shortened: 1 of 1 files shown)


packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
```diff
@@ -536,6 +518,7 @@ void smootherSetup(RCP<Teuchos::ParameterList> params,
   }
   case 3: // Chebyshev
   {
+    relaxationSmootherSetup(params, revisedRowMapPerGrp, regionGrpMats, rowImportPerGrp);
```
mayrmt: I guess, the method `relaxationSmootherSetup()` could/should be renamed to something like `smootherSetup()` or `precomputeInverseDiagonal()`, since this is all that it does. In particular, it is not specific to relaxation smoothers anymore.

Furthermore, the name of the parameter to store the inverse diagonal in the parameter list could/should be renamed as well, since it is not limited to the case of a "relaxation smoother".

mayrmt: Let's rename the parameter list entry to `smoother: inverse diagonal`.

mayrmt: Done.

mayrmt:
@pohm01 @lucbv @rstumin I did some renaming following Peter's changes. In my opinion, this is ready to merge.

PR #6878, opened 2020-02-20
Move to python3

prwolfe:
## Motivation
This moves us to rely on the sims packages as Sierra recently deleted som of their python installs. Due to some differences in the setup this means python 2 is no longer usable so all #! lines are changed to python3.

## Testing
I ran this through all the PR testing jobs by hand with the exception of the CUDA build which was down at the time (I will be watching the PR testing for that one.)


cmake/std/sems/PullRequestGCC4.8.4TestingEnv.sh
```diff
@@ -12,9 +12,9 @@ source /projects/sems/modulefiles/utils/sems-modules-init.sh
 
 module load sems-gcc/4.8.4
 module load sems-openmpi/1.10.1
-module load sems-python/2.7.9
 module load sems-git/2.10.1
 module load sems-boost/1.63.0/base
+module swap sems-python/3.5.2
```
ZUUL42: If another python module isn't already loaded, will `swap` do anything?

prwolfe: it will act as a load in that case.  Don't ask me why.

prwolfe:
To see the actual testing for this see https://github.com/prwolfe/Trilinos/pull/13  

This was needed due to the fact that this PR uses changes in the PullRequestDriver.sh script and therefore cannot bootstrap in the normal PR test environment. It was long because the interaction on ride in particular turned out to be difficult.

PR #6880, opened 2020-02-20
Ifpack2 Factory

cgcgcg:
@trilinos/ifpack2 @trilinos/muelu 

## Motivation
Add a method to check whether a preconditioner is supported by Ifpack2, then use it in MueLu

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_Factory_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Factory_def.hpp
@@ -43,6 +43,7 @@
 #ifndef IFPACK2_DETAILS_FACTORY_DEF_HPP
 #define IFPACK2_DETAILS_FACTORY_DEF_HPP
 
+#include "Ifpack2_Factory.hpp"
 #include "Ifpack2_Details_OneLevelFactory.hpp"
 #include "Ifpack2_AdditiveSchwarz.hpp"
 #if defined(HAVE_IFPACK2_EXPERIMENTAL) && defined(HAVE_IFPACK2_SUPPORTGRAPH)
@@ -64,13 +65,7 @@ create (const std::string& precType,
   RCP<prec_type> prec;
 
   // precTypeUpper is the upper-case version of precType.
-  std::string precTypeUpper (precType);
-  if (precTypeUpper.size () > 0) {
-    std::locale locale;
-    for (size_t k = 0; k < precTypeUpper.size (); ++k) {
-      precTypeUpper[k] = std::toupper<char> (precTypeUpper[k], locale);
-    }
-  }
+  std::string precTypeUpper = canonicalize(precType);
 
   if (precTypeUpper == "SCHWARZ") {
     // Discussion related to Bug 5987: The line of code below will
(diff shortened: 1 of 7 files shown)


packages/ifpack2/src/Ifpack2_Details_OneLevelFactory_def.hpp
```diff
@@ -224,6 +224,113 @@ OneLevelFactory<MatrixType>::create (const std::string& precType,
...
 
+template<class MatrixType>
+bool
+OneLevelFactory<MatrixType>::isSupported (const std::string& precType) const
+{
+  // precTypeUpper is the upper-case version of precType.
+  std::string precTypeUpper (precType);
+  if (precTypeUpper.size () > 0) {
+    std::locale locale;
+    for (size_t k = 0; k < precTypeUpper.size (); ++k) {
```
mhoemmen: This conversion to upper case appears twice, so it would make sense to factor it out into a "canonicalize" function.

packages/ifpack2/src/Ifpack2_Details_OneLevelFactory_def.hpp
```diff
@@ -224,6 +224,113 @@ OneLevelFactory<MatrixType>::create (const std::string& precType,
...
+      precTypeUpper[k] = std::toupper<char> (precTypeUpper[k], locale);
+    }
+  }
+
+  if (precTypeUpper == "CHEBYSHEV") {
```
mhoemmen: Why not just define an array of the permitted names, like this?
```cpp
const char permittedNames[] = {
  "CHEBYSHEV", "DENSE", "LAPACK", "AMESOS2", /* ... etc. ... */ };
const size_t numPermittedNames = /* whatever it is */;
const auto end = permittedNames + numPermittedNames;
auto it = std::find(precTypeUpper, permittedNames, end);
return it != end;
```
Just because it was there before doesn't mean it's the best way to write the code (even if I wrote it before ;-) ).

packages/ifpack2/src/Ifpack2_Details_Factory_def.hpp
```diff
@@ -64,13 +65,7 @@ create (const std::string& precType,
...
 
   // precTypeUpper is the upper-case version of precType.
-  std::string precTypeUpper (precType);
-  if (precTypeUpper.size () > 0) {
-    std::locale locale;
-    for (size_t k = 0; k < precTypeUpper.size (); ++k) {
-      precTypeUpper[k] = std::toupper<char> (precTypeUpper[k], locale);
-    }
-  }
+  std::string precTypeUpper = canonicalize(precType);
```
mhoemmen: Bless you : - D

packages/ifpack2/src/Ifpack2_Details_Factory_def.hpp
```diff
@@ -189,6 +184,31 @@ create (const std::string& precType,
...
+template<class SC, class LO, class GO, class NT>
+bool
+Factory<SC, LO, GO, NT>::
+isSupported (const std::string& precType)
+{
+  // precTypeUpper is the upper-case version of precType.
+  std::string precTypeUpper = canonicalize(precType);
+
+  std::vector<std::string> supportedNames = {
+    "SCHWARZ",
```
mhoemmen: Huh, [apparently it's OK](https://godbolt.org/z/Ld3iz6) to have the comma there even if `"SUPPORTGRAPH"` is `#ifdef`'d out.

packages/ifpack2/src/Ifpack2_Details_Factory_def.hpp
```diff
@@ -189,6 +184,31 @@ create (const std::string& precType,
...
+#if defined(HAVE_IFPACK2_EXPERIMENTAL) && defined(HAVE_IFPACK2_SUPPORTGRAPH)
+    "SUPPORTGRAPH"
+#endif
+  };
+  auto it = std::find(std::begin(supportedNames), std::end(supportedNames), precTypeUpper);
+
+  if (it != std::end(supportedNames)) {
```
mhoemmen: Isn't this much better? : - D  Thanks for revising the code, btw!

cgcgcg:
@mhoemmen Thanks for reviewing!

Issue #6796, opened 2020-02-10
TpetraCore_cudaMpi_CanaryTest_MPI_2 tests needs to check Behavior, not just CMake option (ATDM Trilinos 'vortex' builds)


mhoemmen:
## Bug Report
@trilinos/tpetra 

### Description

The `cudaMpi_CanaryTest` test in `tpetra/core/test/Comm` builds and runs based only on the `Tpetra_ASSUME_CUDA_AWARE_MPI` CMake option.  However, the CMake option only sets Tpetra's _default_ behavior.  Users (and the Dashboard tests) can always override that behavior by setting the `TPETRA_ASSUME_CUDA_AWARE_MPI` environment variable.

### Steps to Reproduce

See instructions for building and running tests on ATS-2 machine 'vortex':

* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md#ats-2

1. Successfully finish building TpetraCore's tests with CUDA enabled and the CMake option `Tpetra_ASSUME_CUDA_AWARE_MPI` explicitly disabled (set to `OFF`).  (It often gets enabled by default, since Tpetra attempts to query `ompi_info` if it exists.)

2. Set the environment variable `TPETRA_ASSUME_CUDA_AWARE_MPI=1`, and run `cudaMPI_CanaryTest`.  It should fail, since the test doesn't check `Tpetra::Details::Behavior`.

Thanks to @bartlettroscoe and @e10harvey for pointing this out : - ) .

bartlettroscoe:
@mhoemmen,

To make this concrete, as shown in [this query](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-01-20&end=2020-02-18&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg&field2=testname&compare2=61&value2=TpetraCore_cudaMpi_CanaryTest_MPI_2&field3=site&compare3=61&value3=vortex), the test:

* `TpetraCore_cudaMpi_CanaryTest_MPI_2`

is failing in the builds:

* `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg`
[...]

PR #6881, opened 2020-02-20
tpetra:  For #6796, adding check of environment variable to test.

### LINKED ISSUE #6873: Tpetra: Remove Details::assumeMpiIsCudaAware in favor of Behavior::assumeMpiIsCudaAware
## Bug Report
@trilinos/tpetra 

### Description

Remove `Tpetra::Details::assumeMpiIsCudaAware` (and its header file) in favor of `Tpetra::Details::Behavior::assumeMpiIsCudaAware`.  See [discussion here](https://github.com/trilinos/Trilinos/issues/6796#issuecomment-588571853).  The two functions are redundant, and Behavior defines the default behavior that we want.

kddevin:
@trilinos/tpetra 

## Motivation

Test cudaMpi_CanaryTest.cpp was intended to be built and run only when Tpetra_ASSUME_CUDA_AWARE_MPI was true.  Its build was included/excluded via CMakeLists.txt via a CMake variable.  

However, new ATS-2 tests are building with Tpetra_ASSUME_CUDA_AWARE_MPI=ON, but running with the environment variable TPETRA_ASSUME_CUDA_AWARE_MPI=OFF.  Thus, this test failed, as it was never meant to be run with TPETRA_ASSUME_CUDA_AWARE_MPI=OFF.

This PR changes the test to check the environment variable via Tpetra::Details::Behavior::assumeMpiIsCudaAware.  It also removes duplicated function Tpetra::Details::assumeMpiIsCudaAware; I don't know why Tpetra had the same functionality implemented twice, but now it won't.

It is dangerous to have CMake variables and environment variables controlling the same behavior, as inconsistent builds and runs can result as in #6796 .

## Related Issues

* Closes #6873 #6796 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback

@bartlettroscoe let me know if you see further test problems.

## Testing

Tested on doom (which defaults to Tpetra_ASSUME_CUDA_AWARE_MPI=OFF) and waterman (which defaults to Tpetra_ASSUME_CUDA_AWARE_MPI=ON) both with and without the environment variable TPETRA_ASSUME_CUDA_AWARE_MPI set opposite to the default.


bartlettroscoe:
> @bartlettroscoe let me know if you see further test problems.

@kddevin, following up is the problem.  That is actually a non-trivial task when there are O(100) tracked tests and O(30) issue trackers.  I am asking for some help to automate the task of checking that tests are fixed and then updating associated GitHub issues.

kddevin:
@trilinos/framework The autotester is reporting test failure on this PR, but CDash is not showing failed tests (only build warnings in ML, AztecOO, and similar).  Can you point me to the reason the autotest is failing?  Thanks.

prwolfe:
Looks like I missed some of the proxy settings on Friday. These are failing due to a networking error Jim reported in #6886 - I should be able to fix this in just a few minutes.

The reason CDash looks clean is that the jobs that do fail fail well before anything is reported to CDash.

PR #6884, opened 2020-02-21
tpetra:  #6883 Removing behavior test with default values controlled by CMake options

### LINKED ISSUE #6883: TpetraCore_Behavior_Default_MPI_4 fails in new 'ats2' CUDA builds: Setting both environment variables and CMake options for behaviors causes tests to fail
## Bug Report
@trilinos/tpetra 

Many Tpetra options can be controlled by both CMake options and environment variables.
The most troublesome pair recently is CMake option -DTpetra_ASSUME_CUDA_AWARE_MPI and environment variable TPETRA_ASSUME_CUDA_AWARE_MPI.  Compiling with one value and setting the environment variable to the opposite value causes tests to fail.  See, for example, #6796 .

This issue will deal with test TpetraCore_Behavior_Default.  See record of its failures here:
https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-01-22&end=2020-02-20&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static&field2=testname&compare2=61&value2=TpetraCore_Behavior_Default_MPI_4&field3=site&compare3=61&value3=vortex

These failures were caused by changes to the ATS-2 testing in which a single build is used with two different values of environment variable TPETRA_ASSUME_CUDA_AWARE_MPI.  One of the two value conflicts with the default value, causing this test to fail.

Since we will now have tests with environment variables differing from configuration variables' values, we cannot rely on the configuration variables values anywhere in CMakeLists.txt or source code when we do testing.  

The solution for this test is to remove the check that assumeMpiIsCudaAware returns the default value, as it does not -- it returns the environment variable.

The longer term solution probably is to remove the CMake option for all variables that can be controlled by environment variables.

### Steps to Reproduce
See link above for test configuration.

kddevin:
@trilinos/tpetra 
## Motivation
This PR removes comparisons of default behavior values with acquired ones for behaviors that are controlled by both CMake options and environment variables.

Trilinos testing is now setting environment variables explicitly, and those settings may differ from the configure-time values of the variables.  Thus, tests that compare the run-time settings with the configure-time settings no longer make sense and may fail.  #6883  #6796 

## Related Issues

* Closes #6883 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #6796 #6881 
* Part of 
* Composed of 

## Stakeholder Feedback

@bartlettroscoe Let me know if you continue to see problems with this test.

## Testing

cee-lan with TPETRA_ASSUME_CUDA_AWARE_MPI = 0 and 1, and TPETRA_DEBUG = 0 and 1
gcc 7.3.0, openmpi 1.10.1
This test does not use CUDA, so running on cee-lan is sufficient.

### DIFF
--- a/packages/tpetra/core/test/Behavior/Behavior_Default.cpp
+++ b/packages/tpetra/core/test/Behavior/Behavior_Default.cpp
@@ -61,25 +61,9 @@ namespace {
 
 TEUCHOS_UNIT_TEST(Behavior, Default)
 {
-#ifdef HAVE_TPETRA_DEBUG
-  bool debug_default = true;
-#else
-  bool debug_default = false;
-#endif
-  bool dbg = Tpetra::Details::Behavior::debug();
-  TEUCHOS_TEST_ASSERT(dbg==debug_default, out, success);
-
   bool verbose_default = false;
   bool verb = Tpetra::Details::Behavior::verbose();
   TEUCHOS_TEST_ASSERT(verb==verbose_default, out, success);
-
-#ifdef TPETRA_ASSUME_CUDA_AWARE_MPI
-  bool cuda_aware_mpi_default = true;
-#else
-  bool cuda_aware_mpi_default = false;
-#endif
-  bool cuda_aware_mpi = Tpetra::Details::Behavior::assumeMpiIsCudaAware();
-  TEUCHOS_TEST_ASSERT(cuda_aware_mpi==cuda_aware_mpi_default, out, success);
 }
 
 TEUCHOS_UNIT_TEST(Behavior, verbosePrintCountThreshold) {
--- a/packages/tpetra/core/test/Behavior/Behavior_Default.cpp
+++ b/packages/tpetra/core/test/Behavior/Behavior_Default.cpp
@@ -64,6 +64,18 @@ TEUCHOS_UNIT_TEST(Behavior, Default)
   bool verbose_default = false;
   bool verb = Tpetra::Details::Behavior::verbose();
   TEUCHOS_TEST_ASSERT(verb==verbose_default, out, success);
+
+  // Print current values of other behaviors. 
+  // Can't test against default since these behaviors may be 
+  // changed by environment variables (in which case, test against
+  // default fails)
+  std::cout << "\n        Cuda-aware MPI?  " 
+            << Tpetra::Details::Behavior::assumeMpiIsCudaAware()
+            << "\n";
+
+  std::cout << "\n        Tpetra Debug?  "
+            << Tpetra::Details::Behavior::debug()
+            << "\n";
 }
 
 TEUCHOS_UNIT_TEST(Behavior, verbosePrintCountThreshold) {


packages/tpetra/core/test/Behavior/Behavior_Default.cpp
```diff
@@ -61,25 +61,9 @@ namespace {
...
   bool verbose_default = false;
   bool verb = Tpetra::Details::Behavior::verbose();
   TEUCHOS_TEST_ASSERT(verb==verbose_default, out, success);
-
-#ifdef TPETRA_ASSUME_CUDA_AWARE_MPI
-  bool cuda_aware_mpi_default = true;
-#else
-  bool cuda_aware_mpi_default = false;
-#endif
-  bool cuda_aware_mpi = Tpetra::Details::Behavior::assumeMpiIsCudaAware();
```
mhoemmen: Will we add back the test for this function at some point?  If we end up getting rid of the CMake option, then we still need some default value.

kddevin:
> I get why you don't want to test for the default value, but why not just call the function and print the value? Part of a unit test is making sure that the function call compiles -- that you can call it with zero arguments and that it returns something convertible to `bool`. If we delete the function call altogether, then we lose that aspect of testing.

@mhoemmen I added the print statements as requested.  Please re-review.  Thanks.

mhoemmen:
@trilinos/framework One of the tests failed in an interesting way:
```
[tr-host-0.novalocal:20042] opal_os_dirpath_create: Error: Unable to create the sub-directory (/tmp/openmpi-sessions-trilinos@tr-host-0_0/17131) of (/tmp/openmpi-sessions-trilinos@tr-host-0_0/17131/0/0), mkdir failed [1]
[tr-host-0.novalocal:20042] [[17131,0],0] ORTE_ERROR_LOG: Error in file util/session_dir.c at line 107
[tr-host-0.novalocal:20042] [[17131,0],0] ORTE_ERROR_LOG: Error in file util/session_dir.c at line 402
[tr-host-0.novalocal:20042] [[17131,0],0] ORTE_ERROR_LOG: Error in file ess_hnp_module.c at line 638
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
...
```

prwolfe:
> @trilinos/framework One of the tests failed in an interesting way:
> 
> ```
> [tr-host-0.novalocal:20042] opal_os_dirpath_create: Error: Unable to create the sub-directory (/tmp/openmpi-sessions-trilinos@tr-host-0_0/17131) of (/tmp/openmpi-sessions-trilinos@tr-host-0_0/17131/0/0), mkdir failed [1]
> [tr-host-0.novalocal:20042] [[17131,0],0] ORTE_ERROR_LOG: Error in file util/session_dir.c at line 107
> [tr-host-0.novalocal:20042] [[17131,0],0] ORTE_ERROR_LOG: Error in file util/session_dir.c at line 402
> [tr-host-0.novalocal:20042] [[17131,0],0] ORTE_ERROR_LOG: Error in file ess_hnp_module.c at line 638
> --------------------------------------------------------------------------
> It looks like orte_init failed for some reason; your parallel process is
> likely to abort.  There are many reasons that a parallel process can
> ...
> ```

Looks like the initial error was that it could not create a temporary directory. Space and permissions are ok, and it runs fine for me this morning.  Not sure what might have caused this.

mhoemmen:
@prwolfe Cool, just wanted to give a heads-up : - )

PR #6885, opened 2020-02-21
Archive the cloud initialization scripts

prwolfe:
These scripts are used when starting up a new
Pull request or test node in the SRN cloud env.
I don't really like putting them here, but I'm
not seeing anything better at the moment.

## Motivation
I am trying to make sure some other people can change and re-work the cloud environment and that means these must be accessible.

## Testing
I have been using these for all re-builds in the cloud systems for a while, there is no additional testing at this point.


prwolfe:
After discussion at todays stand-up we will be looking at other repos for this.

PR #6913, opened 2020-02-27
StackedTimer XML output for Watchr

brian-kelley:
@trilinos/teuchos 
@trilinos/tpetra 
@trilinos/muelu 

## Motivation

In order to use Watchr (the SEMS performance plugin for Jenkins) for performance testing, we need the Teuchos::StackedTimer to output XML files in a specific format. This PR adds functions ``reportXML()`` and ``reportWatchrXML()`` to StackedTimer. ``reportWatchrXML()`` is the one that's intended for performance tests - in the directory ``$WATCHR_PERF_DIR`` (environment variable), it creates an XML file with a name prefix, and a current datestamp. This makes it easy to add XML output to perf tests. It also makes it very easy to set up the performance testing on a given machine:
```
git clone $url_to_data/PerformanceData
export WATCHR_PERF_DIR=cwd/PerformanceData/$machine
# build develop branch Trilinos, with test category PERFORMANCE
# run ctest - the performance tests create files in $WATCHR_PERF_DIR
cd $WATCHR_PERF_DIR
git add .
git commit -m "Testing for $machine on $date"
git push
```

Here is a sample XML file (TpetraCG4_2020_02_27.xml, produced by calling the pseudocode ``reportWatchrXML("TpetraCG" + comm->size(), comm);``:
```
<?xml version="1.0"?>
<performance-report date="2020-02-27T18:00:08" name="nightly_run_2020_02_27" time-units="seconds">
    <timing name="CG: global" value="3.17592">
        <timing name="CG: axpby" value="0.372165"/>
        <timing name="CG: spmv" value="2.53646"/>
        <timing name="CG: dot" value="0.226998"/>
        <timing name="Remainder" value="0.0402957"/>
    </timing>
</performance-report>
```
It's a tree of nested timers with indentation just like report().

## Stakeholder Feedback

This supports the effort for Tpetra and MueLu to do performance testing.
## Testing

Added testing for ``reportXML()`` in the existing StackedTimer_Basic unit test. Note that ``reportWatchrXML()`` is a thin wrapper for ``reportXML()``, that just gets the current time, reads the environment variable and opens the file for writing. By separating out ``reportXML()`` and testing that, the unit test doesn't do disk writes or depend on any environment variables. Also made the test check that both ``report()`` and ``reportXML()`` produce identical output each time they're called (so they don't have any permanent side effects on the StackedTimer object).

### DIFF
--- a/packages/muelu/test/scaling/MatrixMatrixMultiply.cpp
+++ b/packages/muelu/test/scaling/MatrixMatrixMultiply.cpp
@@ -209,6 +209,10 @@ int main_(Teuchos::CommandLineProcessor &clp, Xpetra::UnderlyingLib &lib, int ar
     options.print_warnings = false;
     timer->report(std::cout, comm, options);
 
+    auto xmlOut = stacked_timer->reportWatchrXML(std::string("MueLu_MatrixMatrix") + std::to_string(comm->getSize()), comm);
+    if(xmlOut.length())
+      std::cout << "\nAlso created Watchr performance report " << xmlOut << '\n';
+
     if (comm->getRank() == 0)
       std::cout << "End Result: TEST PASSED";
 
--- a/packages/muelu/test/scaling/MatrixMatrixMultiply.cpp
+++ b/packages/muelu/test/scaling/MatrixMatrixMultiply.cpp
@@ -209,7 +209,7 @@ int main_(Teuchos::CommandLineProcessor &clp, Xpetra::UnderlyingLib &lib, int ar
     options.print_warnings = false;
     timer->report(std::cout, comm, options);
 
-    auto xmlOut = stacked_timer->reportWatchrXML(std::string("MueLu_MatrixMatrix") + std::to_string(comm->getSize()), comm);
+    auto xmlOut = timer->reportWatchrXML(std::string("MueLu_MatrixMatrix") + std::to_string(comm->getSize()), comm);
     if(xmlOut.length())
       std::cout << "\nAlso created Watchr performance report " << xmlOut << '\n';
 
--- a/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
+++ b/packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
@@ -576,6 +576,9 @@ StackedTimer::printLevelXML (std::string prefix, int print_level, std::ostream &
   // computeColumnWidthsForAlignment() or the alignments will be
   // incorrect if the user requests aligned output!
 
+  //Adding an extra indent level, since the <performance-report> is at indent 0
+  int indent = 4 * (print_level + 1);
+
   double total_time = 0.0;
 
   for (int i=0; i<flat_names_.size(); ++i ) {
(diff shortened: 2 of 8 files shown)


brian-kelley:
BTW, added missing copyright header to StackedTimer hpp and cpp.

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -566,6 +604,56 @@ StackedTimer::printLevel (std::string prefix, int print_level, std::ostream &os,
   return total_time;
 }
 
+double
+StackedTimer::printLevelXML (std::string prefix, int print_level, std::ostream &os, std::vector<bool> &printed, double parent_time)
+{
+  // NOTE: If you change the outputting format or logic in this
+  // function, you must make a corresponding change to the function
+  // computeColumnWidthsForAlignment() or the alignments will be
+  // incorrect if the user requests aligned output!
```
rppawlo: This is the third place where we replicate this fragile logic. All three functions should change this comment to list all functions that need to be changed if any one of them change.

brian-kelley: @rppawlo Hmm, I think I should have deleted this comment (obviously, I just copy-pasted printLevel()). printLevelXML() indents according to the current level, but it doesn't support aligning columns.

rppawlo: ok. Sounds good @brian-kelley !

brian-kelley: @rppawlo I had to fix something else, so I deleted this comment.

packages/muelu/test/scaling/MatrixMatrixMultiply.cpp
```diff
@@ -209,6 +209,10 @@ int main_(Teuchos::CommandLineProcessor &clp, Xpetra::UnderlyingLib &lib, int ar
     options.print_warnings = false;
     timer->report(std::cout, comm, options);
 
+    auto xmlOut = timer->reportWatchrXML(std::string("MueLu_MatrixMatrix") + std::to_string(comm->getSize()), comm);
```
mhoemmen: Why would the XML string have zero length?  Is that some sort of indicator that you didn't output Watchr XML?

mhoemmen: Also, this interface looks like a function, but I'm guessing that it has side effects (e.g., output to a file).  Why not just have it return `void` in that case?

EDIT: I can tell from below that it returns the filename.  You could make that clear in the test:
```cpp
if (comm->getRank() == 0) {
  std::cout << "Printed Watchr performance report to file \"" << xmlOut << "\"" << std::endl;
}
```

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -600,6 +683,64 @@ StackedTimer::report(std::ostream &os, Teuchos::RCP<const Teuchos::Comm<int> > c
...
+    os << "<?xml version=\"1.0\"?>\n";
+    os << "<performance-report date=\"" << timestamp << "\" name=\"nightly_run_" << datestamp << "\" time-units=\"seconds\">\n";
+    printLevelXML("", 0, os, printed, 0.0);
+    os << "</performance-report>\n";
+  }
+}
+
+std::string
+StackedTimer::reportWatchrXML(const std::string& name, Teuchos::RCP<const Teuchos::Comm<int> > comm) {
+  char* rawWatchrDir = getenv("WATCHR_PERF_DIR");
```
mhoemmen: PLEASE cast this to `const char*`.  It's UB to modify the result, and we don't want to give other developers ideas ;-) .

brian-kelley: Good idea, done.

packages/muelu/test/scaling/MatrixMatrixMultiply.cpp
brian-kelley: @mhoemmen Could you clarify what you mean by the test? The idea was, user code only needs to test ``xmlOut.length() != 0`` because only ranks that actually wrote a file return a non-empty string. ``if(comm->rank() == 0 && xmlOut.length() != 0)`` is certainly correct, and even though the rank test is redundant, it might make things clearer. Would you prefer I add that?

brian-kelley: Here's the doxygen for that function:
```
  /**
   * Dump all the data from all the MPI ranks to <code>$WATCHR_PERF_DIR/name_$DATESTAMP.xml</code>,
   *  if the environment variable WATCHR_PERF_DIR is defined and non-empty (otherwise, do nothing).
   *  DATESTAMP is calculated from the current UTC time, in the format YYYY_MM_DD.
   * @param [in] name - Name of performance test
   * @param [in] comm - Teuchos comm pointer
   * @return If on rank 0 and output was produced, the complete output filename. Otherwise the empty string.
   */
  std::string reportWatchrXML(const std::string& name, Teuchos::RCP<const Teuchos::Comm<int> > comm);
```

packages/teuchos/comm/src/Teuchos_StackedTimer.cpp
```diff
@@ -701,7 +701,7 @@ StackedTimer::reportXML(std::ostream &os, const std::string& datestamp, const st
 
 std::string
 StackedTimer::reportWatchrXML(const std::string& name, Teuchos::RCP<const Teuchos::Comm<int> > comm) {
-  char* rawWatchrDir = getenv("WATCHR_PERF_DIR");
+  const char* rawWatchrDir = getenv("WATCHR_PERF_DIR");
```
mhoemmen: Thank you :-D

PR #6917, opened 2020-02-27
Tempus fix output for constant time step

### LINKED ISSUE #6916: Tempus:  Fix Double Output at the end of Time Integration
## Bug Report
@trilinos/tempus 

### Description
With constant time stepping, the output occurs at the next to last time step AND last time step.  It should only output at the last time step.

krcb:
@trilinos/tempus 

## Motivation

Fixes issue with incorrect output times for constant timestep using a list.


krcb:
@ccober6 Here is the change.

PR #6918, opened 2020-02-28
Make work with /bin/sh and not just bash on 'vortex'

bartlettroscoe:
See https://cee-gitlab.sandia.gov/EM-Plasma/BuildScripts/issues/230

## How was this tested?

I did not test this on 'vortex' for the use case reported above but Scot Swan did and said it worked.

I tested this on my SNL RHEL6 machine with:

```
$ bash cmake/std/atdm/load-env.sh default
This script 'cmake/std/atdm/load-env.sh' is being called.  Instead, it must be sourced!

Hostname 'crf450.srn.sandia.gov' matches known ATDM host 'sems-rhel6' and system 'sems-rhel6'
Using SEMS RHEL6 compiler stack GNU-7.2.0 to build DEBUG code with Kokkos node type SERIAL
```


cmake/std/atdm/load-env.sh
```diff
@@ -5,9 +5,8 @@
 ################################################################################
 
 # Assert this script is sourced, not run!
-called=$_
-if [ "$called" == "$0" ] ; then
-  echo "This script '$0' is being called.  Instead, it must be sourced!"
+if [ "${BASH_SOURCE[0]}" == "${0}" ] ; then
```
cgcgcg: That's a bash variable, no? So how would this work with other shells?

bartlettroscoe: > That's a bash variable, no? So how would this work with other shells?

@cgcgcg, Scot Swan says this fixed his use case.  I could not confirm it.  This script was never meant to be used by any shell other than bash but if it fixes that EMPIRE use case, so be it.

Can you approve this so it can be merged so EMPIRE can have this?

bartlettroscoe:
Wow, two PR testing iterations failed in a row due to a problem with the Intel license server.  Third times a charm?

cmake/std/atdm/load-env.sh
```diff
@@ -5,9 +5,8 @@
...
+  echo "This script '${0}' is being called.  Instead, it must be sourced!"
```
jmgate: In Scot’s patch, the curly braces weren’t there around the 0.  Not sure if that changes things.

bartlettroscoe: Scot's patch was:

```
-if [ "$called" == "$0" ] ; then
+if [ "${BASH_SOURCE[0]}" == "${0}" ] ; then
   echo "This script '$0' is being called.  Instead, it must be sourced!"
   exit 1
 fi
```

I just added the `{}` around `0` in the `echo` statement to be consistent.

bartlettroscoe: @jmgate, can you approve?

bartlettroscoe:
Shoot, this never got merged due to problems with the PR tester.  Adding `AT: RETEST` to try to get this to get merged ...

PR #6919, opened 2020-02-28
shylubasker,amesos2: updates for kokkos v3 release

ndellingwood:
@trilinos/shylu 

## Motivation
Update ShyLUBasker

## Testing
Tested locally that code compiles and runs


ndellingwood:
@srajama1 or @hkthorn if you get a chance for review, this will be ready for AUTOMERGE when testing completes.

ndellingwood:
Intel build failed to configure due to license issue, adding RETEST label

ndellingwood:
@trilinos/framework who do we contact regarding issues with the intel license server?

packages/amesos2/src/Amesos2_ShyLUBasker.cpp
```diff
@@ -47,331 +47,22 @@
 #include "Amesos2_ShyLUBasker_decl.hpp"
 #include "Amesos2_ShyLUBasker_def.hpp"
 #include "Amesos2_ExplicitInstantiationHelpers.hpp"
+#include "TpetraCore_ETIHelperMacros.h"
```
srajama1: ShyLUBasker should not depend on Tpetra, just Kokkos right ?

ndellingwood: Correct, ShyLUBasker does not depend on Tpetra, but Amesos2 depends on Tpetra so we are fine with the include here. The ETIHelperMacros clean up the code and help better guard against ETI issues with non-default types enabled.

mhoemmen: https://github.com/trilinos/Trilinos/blob/cfccfc9ffc274248d8284dc74607ee3e795f16c0/packages/amesos2/cmake/Dependencies.cmake#L5

srajama1: Thanks, @ndellingwood ! I just didn't notice the filename.

PR #6920, opened 2020-02-28
Amesos2: Upgrade Cholmod for GPU

MicheldeMessieres:
Adds GPU capability to Cholmod solver.
Refactor Cholmod solver to use Kokkos Views.
Fix ETI for Cholmod.
Fix Complex for Cholmod.
Epetra now working for Cholmod.
Add long long GO test for Cholmod to match Trilinos defaults.
Add new Kokkos Adapter tests for Cholmod.
Setup/clean up Cholmod float but Cholmod does not support it yet.

@trilinos/amesos2

## Testing

Amesos2_Cholmod_Solver_Test

### DIFF
--- a/packages/amesos2/src/Amesos2_Cholmod_TypeMap.hpp
+++ b/packages/amesos2/src/Amesos2_Cholmod_TypeMap.hpp
@@ -41,16 +41,13 @@
 //
 // @HEADER
 
-
 /**
-   \file   Amesos2_PardisoMKL_TypeMap.hpp
+   \file   Amesos2_Cholmod_TypeMap.hpp
    \author John Doe <jd@sandia.gov>
    \date
 
-   \brief Provides definition of PardisoMKL types as well as
-          conversions and type traits.  For the purpose of
-          demonstration, we assume that PardisoMKL has defined its own
-          complex data-types called `complex' and `doublecomplex'.
+   \brief Provides definition of Cholmod types as well as
+          conversions and type traits.
 */
 
 #ifndef AMESOS2_CHOLMOD_TYPEMAP_HPP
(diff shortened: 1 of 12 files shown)


srajama1:
@ndellingwood @iyamazaki will take the first pass at this PR.

packages/amesos2/src/Amesos2_Cholmod_TypeMap.hpp
```diff
@@ -68,92 +65,13 @@
...
-
-
 namespace Amesos2 {
 
   // forward declaration due to circular reference
   template <class, class> class Cholmod;
 
   template <>
-  struct TypeMap<Cholmod,float>
+  struct TypeMap<Cholmod,float> // Cholmod does not support float yet
```
ndellingwood: @MicheldeMessieres what happens if someone tries to call this solver with float enabled, will it throw an error with a warning, silently convert to double, or silently fail?
Similar question for complex_float cases.

ndellingwood:
@MicheldeMessieres I don't think cholmod will be tested by PR testing (KLU2 is enabled as the default solver), can you post your configuration and how you tested? Thanks!

packages/amesos2/src/Amesos2_Cholmod_TypeMap.hpp
MicheldeMessieres: This will give a run time error that float is not supported.

MicheldeMessieres:
A config for testing on white:

cmake \
  -GNinja \
  -DAmesos2_ENABLE_TESTS=ON -DTrilinos_ENABLE_Amesos2=ON \
  -DAmesos2_ENABLE_Cholmod=ON \
  -DTPL_ENABLE_Cholmod=ON \
  -DCholmod_LIBRARY_DIRS="/ascldap/users/medemes/SuiteSparse/lib" \
  -DCholmod_LIBRARY_NAMES="cholmod,camd,ccolamd,colamd,suitesparseconfig,amd" \
  -DCholmod_INCLUDE_DIRS="/ascldap/users/medemes/SuiteSparse/include" \
  -DTPL_ENABLE_METIS=ON \
  -DTPL_ENABLE_MPI=ON \
  -DTrilinos_ENABLE_Epetra=ON \
  -DTrilinos_ENABLE_EpetraExt=ON \
  -DAmesos2_ENABLE_Experimental=ON \
  ../trilinospath

This would use the white installed Metis. SuiteSparse can install its own metis so alternative is to set the Metis paths for that. If SuiteSparse has metis.h and you are using white's metis, then there can be conflicts. Cholmod solver won't exhibit a problem but Tacho solver (or the upcoming cuSolver solver) will crash on the metis sort, for example.

For installing SuiteSparse on white, I set ENV as follows:
export BLAS=“-L/home/projects/ppc64le-pwr8/netlib/3.8.0/gcc/7.2.0/lib -lblas -lgfortran -lgomp”
export LAPACK=“-L/home/projects/ppc64le-pwr8/netlib/3.8.0/gcc/7.2.0/lib -llapack -lgfortran -lgomp”
export MY_METIS_LIB=“/home/projects/ppc64le-pwr8/metis/5.0.1/gcc/7.2.0/lib/libmetis.a”
export MY_METIS_INC=“/home/projects/ppc64le-pwr8/metis/5.0.1/gcc/7.2.0/include”

Then make config can check the status before make.

Note trying to set things like MY_METIS_LIB in the make line directly, instead of by export first, as indicated in the docs, didn't seem to work.

PR #6925, opened 2020-02-28
Intrepid2: fixed #6924 (incorrect OPERATOR_D3 output for C1 Hex basis).

### LINKED ISSUE #6924: Intrepid2: Wrong values for OPERATOR_D3 in Basis_HGRAD_HEX_C1_FEM
## Bug Report
@trilinos/intrepid2 

### Description
When evaluating the C1 hexahedral basis with OPERATOR_D3, the basis incorrectly fills the View with zeros.  There should be some non-zero values (they turn out to be either 1/8 or -1/8), corresponding to taking one derivative in each spatial dimension.  (The "Cn" basis with n=1 gets the right values here.)

CamelliaDPG:
Also added:
- BasisEquivalenceTests (one of which demonstrated the issue in #6924)
- location accessor features in ViewIterator
- rank-agnostic view comparison feature for testing (in Intrepid2_TestUtils)

Also fixed the test in
   Discretization/Basis/HGRAD_HEX_C1_FEM/test_01.hpp
which enforced the incorrect zero values for OPERATOR_D3.

@trilinos/intrepid2

## Motivation
As part of the ongoing hierarchical basis development work (now adding simplices), I implemented some basis equivalence tests.  The way these work is that for two bases that span the same space, we compute a conversion matrix that allows one to convert values computed by one basis to those that should be generated by the other.  We compute this conversion matrix using `OPERATOR_VALUE`, and then we use it to compute expected values for other operators.  Effectively, these tests allow us to check that a pair of bases span the same space, and are consistent in terms of their derivative computations.

I added a handful of tests that check this equivalence, both between hierarchical and nodal bases.  One of these, comparing the `C1` hexahedral basis to the `Cn` hexahedral basis with `n=1`, failed for `OPERATOR_D3`.  Inspecting, I determined that the `C1` basis was incorrect.

This PR introduces the new tests, as well as the fix for the `C1` basis.

### DIFF
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_HEX_C1_FEM.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_HEX_C1_FEM.hpp
@@ -144,6 +144,7 @@ namespace Intrepid2 {
           }
           case OPERATOR_GRAD :
           case OPERATOR_D2 :
+          case OPERATOR_D3 :
           case OPERATOR_MAX : {
             auto       output = Kokkos::subview( _outputValues, Kokkos::ALL(), pt, Kokkos::ALL() );
             const auto input  = Kokkos::subview( _inputPoints,                 pt, Kokkos::ALL() );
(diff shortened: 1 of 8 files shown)


mperego:
> It looks good! @ndellingwood Thanks for fixing this and adding the equivalence test.

Sorry.. meant to say @CamelliaDPG not @ndellingwood ..

CamelliaDPG:
@mperego Looks like I have a couple of sign comparison issues here ([link](https://testing.sandia.gov/cdash/viewBuildError.php?buildid=6497860)); I'll fix these on Monday.

Thanks for the fast approval!

CamelliaDPG:
Add WIP label to prevent the auto tester from testing.

CamelliaDPG:
@mperego, I believe I've now fixed all the errors that prevented building with the test configurations, so I've removed the AT:WIP label.  Would you please click the Approve button again?

CamelliaDPG:
@mperego, the latest failures were a tolerance that I needed to relax.  I've now adjusted that tolerance.  Would you please approve again?

PR #6926, opened 2020-02-28
Tpetra: address #3238, fix warnings

### LINKED ISSUE #3238: Tpetra: Replace (implementation of) getEntryOnHost with Kokkos::deep_copy
`Tpetra::Details::getEntryOnHost` returns the value of an entry of a 1-D `Kokkos::View`.  `Kokkos::deep_copy` now has an overload that copies a View to a value, so we can either replace `getEntryOnHost`, or simplify its implementation by calling that overload of `Kokkos::deep_copy`.

@trilinos/tpetra

brian-kelley:
* #3238: implement ``Tpetra::Details::getEntryOnHost`` using ``Kokkos::deep_copy`` from an arbitrary
  0-D view to a host scalar.
* Fix lots of ``calling __host__ function from __host__ __device__ function``
  warnings, due to ``Tpetra::merge2`` being declared as
  ``KOKKOS_INLINE_FUNCTION``, but being passed STL iterators. merge2 is
  only used in a few places in Trilinos and none of them are in device
  kernels, so this PR removes the ``KOKKOS_INLINE_FUNCTION`` marking.
* Fix shadow warning in a MultiVector test

@trilinos/tpetra 

## Motivation

- #3238 is an opportunity to remove redundant code
- fixes dozens of warnings on CUDA builds
- fixes one -Wshadow warning that should show up in any build

## Related Issues

* Closes #3238
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing

getEntryOnHost and merge2 each have their own unit tests, in core/test/Utils.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Util.hpp
+++ b/packages/tpetra/core/src/Tpetra_Util.hpp
@@ -629,7 +629,7 @@ namespace Tpetra {
   ///         val.begin (), val.end ());
   /// \endcode
   template<class IT1, class IT2>
-  KOKKOS_INLINE_FUNCTION void
+  void
   merge2 (IT1& indResultOut, IT2& valResultOut,
           IT1 indBeg, IT1 indEnd,
           IT2 valBeg, IT2 /* valEnd */)
--- a/packages/tpetra/core/src/Tpetra_Details_getEntryOnHost.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_getEntryOnHost.hpp
@@ -53,73 +53,6 @@
 
 namespace Tpetra {
 namespace Details {
-namespace Impl {
-
-template<class ViewType,
-         class IndexType = typename ViewType::size_type,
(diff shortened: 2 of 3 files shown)


packages/tpetra/core/src/Tpetra_Util.hpp
```diff
@@ -629,7 +629,7 @@ namespace Tpetra {
   ///         val.begin (), val.end ());
   /// \endcode
   template<class IT1, class IT2>
-  KOKKOS_INLINE_FUNCTION void
+  void
```
mhoemmen: Huh, I remember this causing warnings, but had thought that we were actually calling `merge2` in a GPU kernel somewhere.  Guess not! : - )

brian-kelley: I was also sure there was a good reason for it, but I checked every call site.

PR #6834, opened 2020-02-14
Xpetra: PR to close #6082 and #6765

kliegeois:
@trilinos/xpetra 
@trilinos/muelu 

## Motivation

The motivation behind this PR is to resolve issues #6765 and #6082.

Issue #6082 is resolved by correcting the typo.

Issue #6765 is resolved by computing the static size required to store the merged blocked crs matrix.
To do so, the implementation relies on the member function getNumEntriesInLocalRow which was not fully implemented yet.
The provided implementation of `getNumEntriesInLocalRow` relies on `getMapIndexForGID` to access the block id of a global id.

Therefore, this PR provides:
- An implementation of `getNumEntriesInLocalRow` and `getNumEntriesInGlobalRow` which supports both Thyra and Xpetra mode.
- The computation of the required memory which has to be allocated to store the merged blocked crs matrix.

## Testing

This branch has been tested on the Blake system.

PR #6931, opened 2020-03-02
Muelu: PR to add an interface aggregation strategy (Aggregation algorithm for Lagrange multipliers)

kliegeois:
@trilinos/muelu

## Motivation

The motivation behind this PR is to add an interface aggregation strategy to construct aggregates on boundaries of physical volumes.
This is based on the work of @tawiesn (algorithm 10 and figure 7.2 of Wiesner, T. (2015). Flexible aggregation-based algebraic multigrid methods for contact and flow problems. PhD thesis, Technische Universitat München) and is required to use a full multigrid preconditioner for mesh-tying and contact problems discretized with Mortar FEM.

Capture d'écran 2020-03-02 16 09 49

In the proposed implementation, a `std::map<LO,LO>` is used to map the primal and dual variables (this mapping is illustrated by the dashed lines in figure 7.2 between the black diamond-shaped nodes and the white dots). This map has to be given on the finest level and is then coarsened and propagated to the next levels automatically.

This PR includes a test which uses the blocked status test of Belos with a blocked Xpetra matrix as requested by @mhoemmen in PR #6834 .

## Testing

This branch has been tested on a system of the University of Liège with debian 9 and gcc.


packages/muelu/test/meshtying/MeshTyingBlocked_SimpleSmoother.cpp
```diff
@@ -0,0 +1,323 @@
...
+  blockedMatrix->fillComplete();
+
+  // Create the preconditioner
+  std::string xmlFile = "myXML.xml";
+
+  ParameterListInterpreter mueLuFactory(xmlFile, *comm);
+  RCP<Hierarchy> H = mueLuFactory.CreateHierarchy();
+  H->GetLevel(0)->Set("A", rcp_dynamic_cast<xpetra_matrix_type>(blockedMatrix));
+
+  H->GetLevel(0)->template Set<RCP<std::map<GO, GO>>>("Lagr2Dof", rcpFromRef(myLagr2Dof));
```
lucbv: This is not really the standard interface to pass parameters to MueLu, more of a hack-ish way to do so.
I would recommend actually dropping the map on the parameter list.

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_TentativePFactory_def.hpp
```diff
@@ -966,6 +966,9 @@ namespace MueLu {
 
     const size_t numElements = rowElements.size();
 
+    if (colElements.size() < rowElements.size())
```
lucbv: Maybe use the variable `numElements` defined above?

packages/muelu/src/Misc/MueLu_InterfaceMappingTransferFactory_def.hpp
```diff
@@ -0,0 +1,37 @@
...
+{
+    fineLevel.DeclareInput("CoarseLagr2Dof", NoFactory::get(), this);
+}
+
+template <class LocalOrdinal, class GlobalOrdinal, class Node>
+void InterfaceMappingTransferFactory<LocalOrdinal, GlobalOrdinal, Node>::Build(Level &fineLevel, Level &coarseLevel) const
+{
+    Monitor m(*this, "Interface Mapping transfer factory");
+    RCP<std::map<GlobalOrdinal, GlobalOrdinal>> coarseLagr2Dof = fineLevel.Get<RCP<std::map<GlobalOrdinal, GlobalOrdinal>>>("CoarseLagr2Dof", NoFactory::get());
+    coarseLevel.Set<RCP<std::map<GlobalOrdinal, GlobalOrdinal>>>("Lagr2Dof", coarseLagr2Dof, NoFactory::get());
```
lucbv: Instead of `NoFactory::get()` you need to pass `this` so that the factory management works inside MueLu, here you are relying on being lucky to get/set the right data from/on the level.

mayrmt:
@kliegeois I'm highly interested in this, since we're in the process of migrating @tawiesn's in-house implementation from his thesis back then to a new implementation. So, I'll be a user of this in the near future.

kliegeois:
@lucbv thanks a lot for your review! I will update the PR with the requested changes.
@mayrmt that is great, hopefully this will be useful for you!

packages/muelu/test/meshtying/MeshTyingBlocked_SimpleSmoother.cpp
tawiesn: I guess, the reason why @kliegeois did it that way is, that "Lagr2Dof" is information that needs to be provided on the finest level by the user (like "A"). One could probably put the dof map in the parameter list but probably not when using xml files (or json files), right? If you want to use the parameter list to provide the finest level information, you would need to read the xml file into a parameter list object and then add the "Lagr2Dof" map to the parameter list using C++ code. Or has this changed in between?

tawiesn: @kliegeois I'm not quite sure whether "Lagr2Dof" is a very obvious name for the map you need. It is the mapping of the Lagrange node ids to dof ids, right? In between I'm tending to use longer but clearer variable names in order to produce "cleaner code" that is easier to understand and maintain for other people.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_decl.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+  //! @name Build methods.
+  //@{
+
+  /*! @brief Build aggregates. */
+  void Build(Level &currentLevel) const;
+
+  //@}
+
+private:
```
tawiesn: Please remove private statement (if empty)

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_def.hpp
```diff
@@ -0,0 +1,165 @@
...
+  return validParamList;
+}
+
+template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
+void InterfaceAggregationFactory<Scalar, LocalOrdinal, GlobalOrdinal, Node>::DeclareInput(Level &currentLevel) const
+{
+  Input(currentLevel, "A");
+  Input(currentLevel, "Aggregates");
+
+  currentLevel.DeclareInput("Lagr2Dof", NoFactory::get(), this);
```
tawiesn: This is a simple and maybe "hacky" way to get this information. Fine for me, but maybe in between there is a smarter way to do so (e.g., getting this information from a parameter list on the finest level). Not that i'm aware of it though.

tawiesn: If other people want to use this factory, they have to be aware how to do so. Maybe one should add a comment or point to the unit test or add a small example in the tutorial.

packages/muelu/src/Misc/MueLu_InterfaceMappingTransferFactory_decl.hpp
```diff
@@ -0,0 +1,80 @@
...
+  template <class LocalOrdinal = DefaultLocalOrdinal,
+            class GlobalOrdinal = DefaultGlobalOrdinal,
+            class Node = DefaultNode>
+class InterfaceMappingTransferFactory : public TwoLevelFactoryBase
+{
+#undef MUELU_INTERFACEMAPPINGTRANSFERFACTORY_SHORT
+#include "MueLu_UseShortNamesOrdinal.hpp"
+public:
+    //! Constructor.
+    InterfaceMappingTransferFactory();
```
tawiesn: A general comment: one could start adding keywords like "explicit" or...

packages/muelu/src/Misc/MueLu_InterfaceMappingTransferFactory_decl.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    //! Destructor.
+    virtual ~InterfaceMappingTransferFactory() {}
+
+    RCP<const ParameterList> GetValidParameterList() const;
```
tawiesn: ... or keywords like "override" where it makes sense.

packages/muelu/src/Utils/MueLu_UtilitiesBase_decl.hpp
```diff
@@ -343,7 +343,6 @@ namespace MueLu {
     */
     static RCP<Vector> GetMatrixOverlappedDeletedRowsum(const Matrix& A) {
       RCP<const Map> rowMap = A.getRowMap(), colMap = A.getColMap();
-      using STS = typename Teuchos::ScalarTraits<Scalar>;
```
tawiesn: Not sure, where this function is used and whether this is necessary.

packages/muelu/test/meshtying/MeshTyingBlocked_SimpleSmoother.cpp
```diff
@@ -0,0 +1,323 @@
...
+#include <Xpetra_CrsMatrixWrap.hpp>
+#include <Xpetra_MultiVectorFactory.hpp>
+#include <Xpetra_VectorFactory.hpp>
+#include <Xpetra_Parameters.hpp>
+
+// MueLu
+#include <MueLu_ConfigDefs.hpp>
+#include <MueLu_Memory.hpp>
+#include <MueLu_Hierarchy.hpp>
+#include <MueLu_CoupledAggregationFactory.hpp>
```
tawiesn: I really doubt that you need all those includes. In order to improve compile speed, reduce them to the absolute minimum

packages/muelu/test/meshtying/MeshTyingBlocked_SimpleSmoother.cpp
kliegeois: @tawiesn both of your comments are right.
For the first one, as you mentioned, it would be difficult to write the map in xml files especially because this map is typically not known when writing the xml files and is an output of the computation of the matrices (with Moertel for instance).  However, it would definitely be possible to first read the xml files and add the map to the parameter list in C++.
For the second one, the map is the mapping of the Lagrange node ids to dof ids if there is one dof per mesh node and one Lagrange multipliers per interface node. Otherwise, it is the mapping of the Lagrange node ids to mesh node ids. 
If you prefer, we can use a map which would be the mapping between the Lagrange dof ids and the primal dof ids, however, this map would be larger.
I can definitely use a larger and clearer variable name here.

tawiesn: @kliegeois Ok, i see. Storing the mapping of the Lagrange node ids to dof ids is fine. But you have those (reasonable) assumptions (e.g., one Lagrange multiplier per interface node). Some of them might not be immediately obvious for other people not so familiar with the code. Maybe you can add some comments about this in the code or description of the classes.

kliegeois: @tawiesn there are three assumptions: the same number of dof per node (primal variable), the same number of dof per interface node (dual variable), and the fact that all dof related to a particular node (primal or dual) are in a same aggregate. The code, as it is, works when more than one Lagrange multiplier per interface node, for instance when solving contact problem with friction. I will clarify that in comments. However, it might be better to remove all those assumptions using a map which is the mapping between the dual dof ids and the primal dof ids. I do not have a strong feeling on the best option.

mayrmt: @kliegois @tawiesn For the case of equal number of primal and dual DOFs (as for friction), I like the idea of just providing a node-to-node mapping. Aggregates are formed out of nodes, so nodes are the information that you want. Providing dog maps just to convert them manually back to node maps is not super straightforward in my opinion. 

A clarification in the documentation should suffice.

tawiesn: Yes, from an aggregation perspective, using node maps makes more sense. Converting dof maps to node maps would just introduce additional overhead.

kliegeois:
Thanks all of you for your comments. 

I tried to improve the implementation, there is still the issue of propagating the map from level to level without using `NoFactory::get()`. I will discuss that tomorrow with @lucbv .

I updated the test to have 3 levels instead of 2.
It raised two issues that I had solved in another branch:

- `SetFixedBlockSize` fails if the matrix has not the same block size for the range and domain map which can be the case for off-diagonal matrices of a block matrix. In the case of the example, the block 00 has a block size of 3 and the block 11 has a block size of 1. This leads to off-diagonal blocks with different block sizes for the domain and range map.
- Not initializing `coarseX_` to zero leads to nan when using SIMPLE with more than 2 levels with the test.

kliegeois:
@lucbv I add the retest label just to launch again the tests even if some changes are still required.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_def.hpp
```diff
@@ -0,0 +1,191 @@
...
+  else
+  {
+    GlobalOrdinal indexBase = aggRangeMap->getIndexBase();
+    auto comm = aggRangeMap->getComm();
+    std::vector<GlobalOrdinal> myDualNodes = {};
+
+    for (size_t i = 0; i < aggRangeMap->getNodeNumElements(); i += nDOFPerDualNode)
+      myDualNodes.push_back((aggRangeMap->getGlobalElement(i) - indexBase) / nDOFPerDualNode + indexBase);
+
+    aggVertexMap = rcp(new const Xpetra::TpetraMap<LocalOrdinal, GlobalOrdinal, Node>(globalNumDualNodes, myDualNodes, indexBase, comm));
```
mayrmt: At the beginning of this function, you typedef `Xpetra::Map` (which can be an Epetra or Tpetra map if I'm not mistaken), but here you explicitly create an `Xpetra::TpetraMap`. Is there a specific reason for that? What about running this code with Epetra?

packages/muelu/test/meshtying/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+    ${PACKAGE_NAME}_ENABLE_Amesos2)
+
+  TRIBITS_COPY_FILES_TO_BINARY_DIR(MeshTying_cpdata
+    SOURCE_FILES Q_mm.txt G_mm.txt GT_mm.txt C_mm.txt f_mm.txt Lagr2Dof.txt
+    )
+
+  TRIBITS_ADD_EXECUTABLE_AND_TEST(
+    MeshTyingBlocked_SimpleSmoother
+    SOURCES MeshTyingBlocked_SimpleSmoother.cpp
+    NUM_MPI_PROCS 1
```
mayrmt: Is the new interface aggregation currently limited to serial or just the test?

packages/muelu/test/meshtying/myXML.xml
```diff
@@ -0,0 +1,227 @@
...
+      <Parameter name="factory"                             type="string" value="InterfaceMappingTransferFactory"/>
+    </ParameterList>
+
+    <ParameterList name="myInterfaceAggs2">
+      <Parameter name="factory" type="string" value="InterfaceAggregationFactory"/>
+      <Parameter name="Aggregates" type="string" value="myAggFact1"/> <!-- use aggregates from primal block! -->
+      <Parameter name="number of DOFs per dual node" type="int" value="1"/>
+      <Parameter name="Interface Mapping Transfer Factory" type="string" value="mySpecialMappingFact"/>
+    </ParameterList>
+
```
tawiesn: In order to deal with such a cyclic variable dependency between two factories, you can extend the factory definition in the xml file after the factories have been defined.

Add a new factory block here, that looks like
>    < ParameterList name="mySpecialMappingFactDeps" >
>      < Parameter name="dependency for" type="string" value="mySpecialMappingFact" / >
>      < Parameter name="CoarseDualNodeID2PrimalNodeID" type="string" value="myInterfaceAggs2" / >
>    < / ParameterList>

tawiesn: The name "mySpecialMappingFactDeps" has no meaning. You need a new and unique block name for each xml block. But this name gives you a hint that you want to extend the definition of the mySpecialMappingFact block from above. The parameter "dependency for" is the keyword. The value contains the "factory name" that you want to extend (in our case "mySpecialMappingFact"). And then you just add the missing parameter dependencies. In this case you declare the myInterfaceAggs2 factory to be the generating factory for the "Coarse..." variable. You couldn't do it in the block "mySpecialMappingFact" above as the factory name "myInterfaceAggs2" was not defined, yet. But now you can do it after the "myInterfaceAggs2" block is defined.

tawiesn: Please also have a look at
https://github.com/trilinos/Trilinos/blob/master/packages/muelu/src/Interface/MueLu_ParameterListInterpreter_def.hpp
particularly line 2115 ff.
There you find a more detailed description of how to handle cyclic problems. I guess, if you grep through the examples looking for "dependency for" you might (hopefully) find some examples, too.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_def.hpp
kliegeois: You are right! Thanks!

packages/muelu/test/meshtying/CMakeLists.txt
kliegeois: The test is limited to serial at this point.
I will investigate its behavior with MPI but it is currently failing.

kliegeois:
Thanks a lot @tawiesn for your inputs!
I think that it is working correctly now.

The test is working in serial, could we consider accepting this PR as it is and that I introduce a second PR later to address the test with MPI?

I would need this PR to be accepted for other developments.

kliegeois:
Do you have an idea of why PanzerDofMgr_tUniqueGlobalIndexerUtilities_MPI_2 fails with the gcc 4.8.4 build?

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_def.hpp
```diff
@@ -0,0 +1,165 @@
...
+    size_t n_dof_per_lag = (aggDomainMap->getGlobalNumElements()) / lagr2dof->size();
+    LocalOrdinal numVertices = lagr2dof->size();
+    GlobalOrdinal indexBase = aggDomainMap->getIndexBase();
+    auto comm = aggDomainMap->getComm();
+    std::vector<GlobalOrdinal> myVertices = {};
+
+    for (size_t i = 0; i < aggDomainMap->getNodeNumElements(); i += n_dof_per_lag)
+      myVertices.push_back((aggDomainMap->getGlobalElement(i) - indexBase) / n_dof_per_lag + indexBase);
+
+    aggVertexMap = rcp(new const Xpetra::TpetraMap<LocalOrdinal, GlobalOrdinal, Node>(numVertices, myVertices, indexBase, comm));
```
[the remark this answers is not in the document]
kliegeois: This has been resolved. 
I do not know why github still shows this older version.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_decl.hpp
```diff
@@ -0,0 +1,107 @@
...
+//
+// @HEADER
+#ifndef MUELU_INTERFACEAGGREGATIONFACTORY_DECL_HPP_
+#define MUELU_INTERFACEAGGREGATIONFACTORY_DECL_HPP_
+
+#include "MueLu_ConfigDefs.hpp"
+#include "MueLu_SingleLevelFactoryBase.hpp"
+
+#include "MueLu_Level_fwd.hpp"
+#include "MueLu_Aggregates_fwd.hpp"
```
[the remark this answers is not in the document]
kliegeois: I think that you are right, I can remove those two includes.

kliegeois:
@tawiesn wrote:
> Is the test running with the myXml.xml provided here? I could not see the diff to the previous version. If yes, it's fine for me. Thanks for working on this!

Yes, the test was passing since my last commit. The test that blocked the PR was another test which had no links with the changes of this PR and which failed on other PRs too.

Thanks a lot for your review!

kliegeois:
@lucbv would you agree to review again this PR as your requested change blocks the merge?

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_decl.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+#undef MUELU_INTERFACEAGGREGATIONFACTORY_SHORT
+#include "MueLu_UseShortNamesScalar.hpp"
+
+public:
+  //! @name Constructors/Destructors.
+  //@{
+
+  //! Constructor.
+  explicit InterfaceAggregationFactory();
```
lucbv: I do not think that the explicit keyword makes sense since there are no arguments passed to the constructor.
On the other hand you could (should?) go with the `default` qualifier:
```
InterfaceAggregationFactory() = default;
```
this allows you to remove the implementation of the constructor in the `_def.hpp` file.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_decl.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+  //! Destructor.
+  virtual ~InterfaceAggregationFactory() {}
```
lucbv: I do not think that we have any plans on having a child class implement an override of that destructor so I am not sure why it would be declared virtual? I see we do it a lot in MueLu so that's fine but I believe it is not really correct...

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_def.hpp
```diff
@@ -0,0 +1,196 @@
...
+  }
+}
+
+template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
+void InterfaceAggregationFactory<Scalar, LocalOrdinal, GlobalOrdinal, Node>::Build(Level &currentLevel) const
+{
+  typedef Xpetra::Map<LocalOrdinal, GlobalOrdinal, Node> Map;
+  typedef Aggregates<LocalOrdinal, GlobalOrdinal, Node> Aggregates;
+  typedef std::map<LocalOrdinal, LocalOrdinal> Dual2Primal_type;
+  typedef Xpetra::MapFactory<LocalOrdinal, GlobalOrdinal, Node> MapFactory;
```
lucbv: consider switching `typedef` to `using`, it makes @mhoemmen happy ; )

lucbv: Also I think these are already defined in the `UseShortNamesOrdinal` header, I think you should check and try to remove these as they might be redefinitions...

lucbv: Finally if you get tired of typing `LocalOrdinal` you could just switch to `LO`...

packages/muelu/src/Misc/MueLu_InterfaceMappingTransferFactory_decl.hpp
```diff
@@ -0,0 +1,80 @@
...
+  template <class LocalOrdinal = DefaultLocalOrdinal,
+            class GlobalOrdinal = DefaultGlobalOrdinal,
+            class Node = DefaultNode>
+class InterfaceMappingTransferFactory : public TwoLevelFactoryBase
+{
+#undef MUELU_INTERFACEMAPPINGTRANSFERFACTORY_SHORT
+#include "MueLu_UseShortNamesOrdinal.hpp"
+public:
+    //! Constructor.
+    explicit InterfaceMappingTransferFactory();
```
lucbv: again no need for the `explicit` qualifier but rather use the `default` one.

packages/muelu/src/Misc/MueLu_InterfaceMappingTransferFactory_def.hpp
```diff
@@ -0,0 +1,38 @@
+#ifndef MUELU_INTERFACEMAPPINGTRANSFERFACTORY_DEF_HPP_
+#define MUELU_INTERFACEMAPPINGTRANSFERFACTORY_DEF_HPP_
+
+#include "MueLu_InterfaceMappingTransferFactory_decl.hpp"
+
+namespace MueLu
+{
+
+template <class LocalOrdinal, class GlobalOrdinal, class Node>
+InterfaceMappingTransferFactory<LocalOrdinal, GlobalOrdinal, Node>::InterfaceMappingTransferFactory()
```
lucbv: Can be removed if the qualifier `default` is used in the declaration of the constructor.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_def.hpp
kliegeois: Concerning `UseShortNamesOrdinal`, it was what I thought too but it is not working if I remove one of those `typedef`. I will, however, use `using` instead of `typedef`! Thanks!

kliegeois:
@lucbv wrote:
> Some more small changes but it looks pretty good overall!
> Thanks

Thanks a lot for your review!
I just updated the files as suggested (except concerning `UseShortNamesOrdinal` as explained in my comment). 
Those changes have not impacted the tests on my computer.

mayrmt:
@kliegeois You have added a test with 1 DOF per dual node. Have you used this also with multiple DOFs per dual node? If I do so, the dual node aggregates `aggVertexMap` in the `InterfaceAggregationFactory` does not contain all dual nodes. 

I will look into this for our application, where we always have 2 or 3 DOFs per dual node in 2D or 3D problems, respectively. I'm just wondering, if I can modify the multiple-DOFs-per-node case freely or if you're using this with multiple DOFs per dual node somewhere.

Issue #6932, opened 2020-03-02
MueLu:  Build error in MueLu_BelosSmoother


vbrunini:
## Bug Report
@trilinos/muelu

### Description
We're seeing a build error in all of our integration builds (gcc, intel, and nvidia).

`Trilinos/packages/muelu/src/Smoothers/MueLu_BelosSmoother_decl.hpp:72:10: fatal error: BelosXpetraAdapter.hpp: No such file or directory`

### Steps to Reproduce
Configure script:
```
cmake \
  -G "Ninja" \
  -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
  -D Trilinos_EXTRA_LINK_FLAGS:STRING="" \
  -D CMAKE_BUILD_TYPE:STRING=release \
 -D Trilinos_ENABLE_OpenMP:BOOL=OFF \
 -D TPL_ENABLE_CUDA:BOOL=OFF \
 -DCMAKE_C_FLAGS_RELEASE_OVERRIDE:STRING="-DNDEBUG" \
 -DCMAKE_CXX_FLAGS_RELEASE_OVERRIDE:STRING="-DNDEBUG" \
 -DCMAKE_Fortran_FLAGS_RELEASE_OVERRIDE:STRING="-DNDEBUG" \
 -DCMAKE_C_FLAGS:STRING="-O2 -finline-functions -ftemplate-depth-128 -w " \
 -DCMAKE_CXX_FLAGS:STRING="-O2 -finline-functions -ftemplate-depth-128 -w " \
 -DCMAKE_Fortran_FLAGS:STRING="-O2 -finline-functions -ftemplate-depth-128 -w  " \
 -DHAVE_GCC_ABI_DEMANGLE:BOOL=ON \
 -DTrilinos_WARNINGS_AS_ERRORS_FLAGS:STRING="" \
 -DTrilinos_SHOW_DEPRECATED_WARNINGS:BOOL=OFF \
 -DTrilinos_ASSERT_MISSING_PACKAGES:BOOL=OFF \
 -DTrilinos_ENABLE_COMPLEX_DOUBLE:BOOL=ON \
 -DTrilinos_ENABLE_FLOAT:BOOL=ON \
 -DTrilinos_ENABLE_LINEAR_SOLVER_FACTORY_REGISTRATION:BOOL=ON \
 -DTrilinos_ENABLE_EXPLICIT_INSTANTIATION:BOOL=ON \
[...]

PR #6939, opened 2020-03-02
MueLu: Require Belos-Xpetra adapter when using Belos in MueLu

cgcgcg:
@trilinos/muelu 

## Motivation
Takes care of #6932

### DIFF
--- a/packages/muelu/adapters/CMakeLists.txt
+++ b/packages/muelu/adapters/CMakeLists.txt
@@ -20,6 +20,12 @@ ENDIF()
 #
 IF (${PACKAGE_NAME}_ENABLE_Belos)
 
+  ASSERT_DEFINED(Belos_ENABLE_Xpetra)
+
+  IF (NOT Belos_ENABLE_Xpetra)
+    MESSAGE(FATAL_ERROR "MueLu requires the Belos Xpetra adapters in order to work with Belos. Please set \"Belos_ENABLE_Xpetra:BOOL=ON\"")
+  ENDIF ()
+
   INCLUDE_DIRECTORIES(${CMAKE_CURRENT_SOURCE_DIR}/belos)
 
   APPEND_SET(HEADERS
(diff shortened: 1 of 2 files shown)


packages/muelu/adapters/CMakeLists.txt
```diff
@@ -20,6 +20,12 @@ ENDIF()
 #
 IF (${PACKAGE_NAME}_ENABLE_Belos)
 
+  ASSERT_DEFINED(Belos_ENABLE_Xpetra)
+
+  IF (NOT Belos_ENABLE_Xpetra)
+    MESSAGE(FATAL_ERROR "MueLu requires the Belos Xpetra adapters in order to work with Belos. Please set \"Belos_ENABLE_Xpetra:BOOL=ON\"")
```
mhoemmen: @trilinos/muelu I don't like this -- see e.g., #6932.  This should either be on by default, or you shouldn't need it.  Now MueLu is forcing all of Belos to build yet again.  @vbrunini @rrdrake

cgcgcg: @mhoemmen Ok, I'll take another stab at fixing this.

mhoemmen: @cgcgcg Sorry to be a pest -- I just think it would be better to avoid the Xpetra instantiation of Belos if possible.  If you can't help it -- e.g., if you need Xpetra's special blocked operators to work -- then that's OK.

cgcgcg: @mhoemmen Have a look at PR #6952. (I put the automerge on it without thinking, so it already got merged.)

mhoemmen: @cgcgcg Thanks for letting me know!  It looks like PR #6952 just deleted the offending implementation -- is that right?  I'm totally OK with that btw : - ) .

Issue #6957, opened 2020-03-04
KokkosKernels:  Compilation errors with TPL_ENABLE_MKL=ON


etphipp:
@trilinos/kokkos-kernels 

Kokkos-Kernels does not compile when enabling the Trilinos option TPL_ENABLE_MKL=ON.  

There are two issues.  First the file src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp generates errors such as
```
/home/etphipp/Trilinos/Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp:212:9: error: 
      unknown type name 'sparse_index_base_t'; did you mean
      'KokkosBatched::sparse_index_base_t'?
        sparse_index_base_t c_indexing;
        ^~~~~~~~~~~~~~~~~~~
        KokkosBatched::sparse_index_base_t
/home/software/intel/compilers_and_libraries_2019.3.199/linux/mkl/include/mkl_spblas.h:658:7: note: 
      'KokkosBatched::sparse_index_base_t' declared here
    } sparse_index_base_t;

/home/etphipp/Trilinos/Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp:216:13: error: 
      use of undeclared identifier 'SPARSE_STATUS_SUCCESS'; did you mean
      'KokkosBatched::SPARSE_STATUS_SUCCESS'?
        if (SPARSE_STATUS_SUCCESS !=
            ^~~~~~~~~~~~~~~~~~~~~
            KokkosBatched::SPARSE_STATUS_SUCCESS
/home/software/intel/compilers_and_libraries_2019.3.199/linux/mkl/include/mkl_spblas.h:624:9: note: 
      'KokkosBatched::SPARSE_STATUS_SUCCESS' declared here
        SPARSE_STATUS_SUCCESS           = 0,    /* the operation was suc...
```
Looking through that file, I thought I was seeing these because I had KokkosKernels_ENABLE_TPL_MKL=OFF (which is the default, regardless of TPL_ENABLE_MKL), but even turning that on doesn't fix it.

Second, files such as src/batched/KokkosBatched_Gemm_Serial_Impl.hpp and src/batched/KokkosBatched_Trsm_Serial_Impl.hpp also include MKL-specific specializations that generate errors if deprecated code in Kokkos is turned off, e.g.,
```
[...]

PR #6959, opened 2020-03-04
KokkosKernels - replace dimension with extent

kyungjoo-kim:
## Motivation

As reported in #6957, mkl header location in kokkos batched util file is problematic. It should go above kokkosbatched name space. 

This PR also replaces dimension with extent. 

A separate PR for kokkoskernels will be submitted.

### DIFF
--- a/packages/kokkos-kernels/src/batched/KokkosBatched_Util.hpp
+++ b/packages/kokkos-kernels/src/batched/KokkosBatched_Util.hpp
@@ -24,9 +24,7 @@
 
 #include "KokkosKernels_config.h"
 
-namespace KokkosBatched {
-
-  // TPL macros
+// TPL macros
 #if defined (KOKKOSKERNELS_ENABLE_TPL_MKL) 
 #define __KOKKOSBATCHED_INTEL_MKL__ 1
 #include "mkl_version.h"
@@ -37,6 +35,8 @@ namespace KokkosBatched {
   //#include "mkl_types.h"
 #endif
 #endif
+
+namespace KokkosBatched {
 
 #if defined (KOKKOSKERNELS_ENABLE_TPL_CUBLAS)
 #define __KOKKOSKERNELS_NVIDIA_CUBLAS__ 1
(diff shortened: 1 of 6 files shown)


ndellingwood:
@kyungjoo-kim please add matching PR to kokkos-kernels so this isn't clobbered by the next promotion

ndellingwood:
kokkos/kokkos-kernels#635

packages/kokkos-kernels/src/batched/KokkosBatched_Util.hpp
```diff
@@ -24,9 +24,7 @@
 
 #include "KokkosKernels_config.h"
 
-namespace KokkosBatched {
-
-  // TPL macros
+// TPL macros
 #if defined (KOKKOSKERNELS_ENABLE_TPL_MKL) 
 #define __KOKKOSBATCHED_INTEL_MKL__ 1
 #include "mkl_version.h"
```
etphipp: I'm not sure this fully resolves the issues if KokkosKernels is built with TPL_ENABLE_MKL=ON but KokkosKernels_ENABLE_TPL_MKL=OFF (which is the default).  In particular I don't see how it prevents the MKL  instantiations in KokkosKernels from being compiled.  Did you try a build in this configuration?

kyungjoo-kim: The problem in the spgemm is that the code somehow sees mkl included in my code (with namespace KokkosBatched - this is my mistake). With MKL=ON or OFF, the mkl files should be included outside without a name space and the compiler does not confuse with the index object with kokkos batched namespace.

kyungjoo-kim: Also, I was able to compile the code in trilinos. BTW, @ndellingwood don't we have a nightly test for TPLs ?

ndellingwood: > @ndellingwood don't we have a nightly test for TPLs

Not yet but we will this week, I put in the PR last week to update testing scripts that more easily allows TPLs to be enabled on the systems we run nightlies.

PR #6960, opened 2020-03-04
MueLu ParameterListInterpreter test: run single-threaded

cgcgcg:
@trilinos/muelu 

## Motivation
Run the ParameterListInterpreter test always using a single thread. This hopefully takes care of all non-deterministic results we're seeing.


mhoemmen:
This sounds a bit like "threads make the test fail, so don't run threads."

FYI @rrdrake @vbrunini @micahahoward

PR #6965, opened 2020-03-05
Tacho levelset implementation

kyungjoo-kim:
## Motivation
For efficient sparse factorization and solve, it is better to use level set scheduling as kokkos tasking does not allow user-defined team size yet (it is fixed as 32 and it is too small). This PR also include re-organization of files for ETI.

## Testing
Testing CUDA and OPENMP on kokkos-dev-2


kyungjoo-kim:
@srajama1 I do not plan to interface to zoltan2. I believe that zoltan and zoltan2 provide partition and ordering capabilities together with MPI. If this code is MPI parallel, I would consider to use zoltan2 but I don't think that it is good to interface zoltan2 as this code is primarily a node parallel code.

srajama1:
> @srajama1 I do not plan to interface to zoltan2. I believe that zoltan and zoltan2 provide partition and ordering capabilities together with MPI. If this code is MPI parallel, I would consider to use zoltan2 but I don't think that it is good to interface zoltan2 as this code is primarily a node parallel code.

@kyungjoo-kim Zoltan2 supports both node parallel and mpi parallel. From the Trilinos perspective, it is good to reuse packages others write, avoid interfacing to the same software in multiple packages which causes issues later.

kddevin:
Having a single interface is especially helpful when TPLs update ad change their APIs.  Updates in Trilinos need to be done only once, rather than in dozens of places.  As you know, I am pushing for similar practices in @trilinos/tpetra, so that we do not call cuda libraries directly but, rather, call them through KokkosKernels. (I know tpetra does not have complete success yet, but we are working with KokkosKernels to get there.)

If you need anything specific in @trilinos/zoltan2, please let me know.

PR #6967, opened 2020-03-05
Amesos2 add cuSolver solver

MicheldeMessieres:
Amesos2 add cuSolver solver

@trilinos/amesos2

## Motivation

Setup cuSolver capability for Amesos2.

## Testing

Amesos2_cuSOLVER_Solver_Test


packages/amesos2/src/Amesos2_Util.hpp
```diff
@@ -66,11 +66,15 @@
...
 #include "Amesos2_TypeDecl.hpp"
 #include "Amesos2_Meta.hpp"
+#include "Amesos2_Kokkos_View_Copy_Assign.hpp"
 
 #ifdef HAVE_AMESOS2_EPETRA
 #include <Epetra_Map.h>
 #endif
 
+#ifdef HAVE_AMESOS2_METIS
+#include "metis.h" // to discuss, remove from header?
```
mhoemmen: Please do -- including TPL headers in Trilinos headers risks breaking applications' builds (e.g., if they use extern declarations instead of headers, or if the versions don't quite line up).

packages/amesos2/src/Amesos2_Util.hpp
```diff
@@ -1023,6 +1027,166 @@ namespace Amesos2 {
...
+        // we'll permute matrix on device to a new set of arrays
+        row_ptr_view_t new_row_ptr(
+          Kokkos::ViewAllocateWithoutInitializing("new_row_ptr"), row_ptr.size());
+        cols_view_t new_cols(
+          Kokkos::ViewAllocateWithoutInitializing("new_cols"), cols.size() - new_nnz/2);
+        values_view_t new_values(
+          Kokkos::ViewAllocateWithoutInitializing("new_values"), values.size() - new_nnz/2);
+
+        // permute row indices
+        Kokkos::RangePolicy<exec_space_t> policy_row(0, row_ptr.size());
```
ndellingwood: @MicheldeMessieres I think a fence may need to be added before the `parallel_scan` as a precaution to prevent issues when UVM is enabled to ensure data is properly synced (unless all device Views are explicitly templated on `Kokkos::CudaSpace` when Cuda is enabled).

ndellingwood:
@MicheldeMessieres I don't think cusparse will be tested by PR testing (KLU2 is enabled as the default solver), can you post your configuration and how you tested? Thanks!

packages/amesos2/src/Amesos2_Util.hpp
MicheldeMessieres: Discussed plan was to have this use zoltan2 ordering instead of Metis.h. Will need to discuss best way to make that work with the new kokkos adapter.

For the cuSolver headers, those are also in the include stack, via Amesos2_cuSOLVER_FunctionMap.hpp which is currently accessed directly by the decl file Amesos2_cuSOLVER_decl.hpp. The other Amesos2 solvers also include this way for TPLs so the plan is to not make any changes for that right now.

packages/amesos2/src/Amesos2_Util.hpp
MicheldeMessieres: Currently cuSolver will work explicitly on CudaSpace or Host but not UVM. The Kokkos adapter tests will send all 3 types (Host, CudaUVMSpace, CudaSpace) as inputs but the solver itself is never using UVM views. For Cholmod the views are always Host. For Tacho, the Views are still UVMSpace due to the TaskScheduler issues we had. Tacho doesn't use this ordering path and we plan to make it also just CudaSpace once the other issues are fixed. So perhaps the fence is not needed presently. Not sure if we want to start anticipating CUDA_LAUNCH_BLOCKING off formatting yet.

MicheldeMessieres:
This is a config used to test this on white:

cmake \
  -GNinja \
  -DAmesos2_ENABLE_TESTS=ON \
  -DTrilinos_ENABLE_Amesos2=ON \
  -DTPL_ENABLE_MPI=ON \
  -DAmesos2_ENABLE_METIS=ON \
  -DTPL_ENABLE_METIS=ON \
  -DTrilinos_ENABLE_Epetra=ON \
  -DTrilinos_ENABLE_EpetraExt=ON \
  -DTPL_ENABLE_CUSOLVER=ON \
  ../<trilinos>

Also tested float and complex on.

Note turning METIS off will give a run-time error that ordering is not supported. The param supports reordering as an option so you can run it without metis reordering, with the loss of performance.  This will be changing anyways as we will migrate to zoltan2 ordering shortly.

MicheldeMessieres:
@ndellingwood I rebased to make sure the prior Cholmod PR was cleanly merged with this one. Then removed name spacing on cuSolver includes since Kokkos may include those already and it created build conflicts (Stokhos cdash errors on prior auto test). If Amesos2 wants to namespace cuSolver as it does other TPLs like superlu, I'll need to discuss with you how to resolve that. If this is ok, then the PR should be ready for another round of auto testing.

MicheldeMessieres:
@ndellingwood Just pinging on this. If we can review this update I can open up the superlu PR once this is in. Thanks.

Issue #6968, opened 2020-03-05
PR Testing: Use C++14 standard for GCC 8.3 build


jwillenbring:
## Enhancement
@trilinos/framework

We are going to add -std=c++14 to the GCC 8.3 build (and remove -std=c++11). We will for the time being also keep the GCC 4.8.4 build, which will continue to insure compatibility with C++11. While some packages are looking to use C++14 (and C++17) features, this will at least move us in the right direction and ensure that customers will be able to build Trilinos with c++14.

This incremental step is consistent with discussions at this week's Trilinos Product Leaders meeting and is related to #6260.

jwillenbring:
The PR has been merged, so I am closing this.

jwillenbring:
As noted in #6260 this fix may not have done what was expected. I am going to reopen.

bartlettroscoe:
CC: @crtrott, @mwglass 

@jwillenbring, the issue of C++14 PR builds for Trilinos just came up again today in the ATDM PI meeting.  Is there is a PR build that has C++14 support enabled?  Do we need to merge PR #6814 before this gets addressed?  NOTE: I think you can just set `-D Trilinos_CXX_FLAGS="-std=c++14` for now until we get PR #6814 (and the related Kokkos and TriBITS PRs) merged.

bartlettroscoe:
CC: @trilinos/framework, @ibaned 

> NOTE: I think you can just set `-D Trilinos_CXX_FLAGS="-std=c++14"` for now until we get PR #6814 (and the related Kokkos and TriBITS PRs) merged.

I checked into that and that trick no longer works as we have given up all handling of C++ options to Kokkos.

@crtrott,

As discussed at the ATDM PI Meeting today, can we put in a short term hack to the Kokkos version in Trilinos to allow us to add `-std=c++14` to the commandline instead of `-std=c++11` to allow the setup of a C++14 PR ASAP?  Perhaps we  set a temporary (undocumented) cache var like:

```
-D Kokkos_USE_CXX_STD=14
```

that would result in the Kokkos CMake logic putting in `-std=c++14` instead of `-std=c++11`?  We might even be able to hack that into the file:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/ProjectCompilerPostConfig.cmake

(by pulling out `-std=c++11` and putting in `-std=c++14`)?

[...]

PR #6969, opened 2020-03-05
Move the GCC 8.3 PR build to C++14

jwillenbring:
@trilinos/<teamName>

## Motivation

We want to move the GCC 8.3 PR build to C++14 as described in #6968. If PR testing for this PR passes, we can do that, otherwise we will have results showing what issues have to be resolved.

In addition, this PR represents an incremental step towards #6260.


mhoemmen:
@rrdrake @micahahoward @sebrowne @vbrunini FYI

PR #6975, opened 2020-03-06
shylubasker: fix transpose solve, add test

ndellingwood:
@trilinos/shylu 

## Motivation
Fix shylubasker for transpose solves

## Testing
Added simple test to check non-transpose and transpose solves, verified correctness

### DIFF
--- a/packages/shylu/shylu_node/basker/src/shylubasker_def.hpp
+++ b/packages/shylu/shylu_node/basker/src/shylubasker_def.hpp
@@ -369,12 +369,9 @@ namespace BaskerNS
       }
       else
       {
-        //Will transpose and put in A using little extra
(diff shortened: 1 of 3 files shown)


ndellingwood:
Adding @iyamazaki also, welcome to Trilinos developers!

ndellingwood:
Added forgotten new test file (missed it when adding "updated" files)

ndellingwood:
> What about when the options are changed on the solver object using the top-level "use transpose" option, not the solver-specific one?

@hkthorn by top-level are you referring to calls from Amesos2? I'm going to address that in a separate PR, this just fixes transpose solves when building ShyLUBasker as a solver outside of Amesos2.

PR #6978, opened 2020-03-06
KokkosKernels - Impl::is_same gone.

kyungjoo-kim:
## Motivation
Trilinos kokkoskernels is not compiled because Kokkos::Impl::is_same gone. I am not sure how this is missing from 3.0 promotion testing.

## Testing
I found this problem on kokkos-dev-2 and confirms that this fixes the problem. A matching PR to kokkoskernels will be submitted soon.


kyungjoo-kim:
I also found that kokkoskernels already has std::is_same now. Maybe I should discard this PR.

PR #6982, opened 2020-03-09
(DO NOT MERGE YET) Tpetra::CrsMatrix: Use cuSPARSE for non-transposed sparse matrix-vector multiply with Scalar=float or double

mhoemmen:
@trilinos/tpetra 

Please don't merge this yet; I just would like to see if it passes PR tests.

## Motivation

Local sparse matrix-vector multiply has poor load balance on GPUs for matrices that have a few rows with many entries, and many rows with few entries.  cuSPARSE's merge path algorithm promises to help with that.  This PR offers some steps towards making that work.

## Stakeholder Feedback

Aria (see e.g., @vbrunini ) very much needs this to be done this FY.

## Testing

All but the last two commits pass Tpetra's tests in a CUDA 9.2 + GCC 7.2 + OpenMPI build.  For the last two commits, I actually added a test in Tpetra for a function that wasn't tested before.


mhoemmen:
Goofy build error that Clang didn't catch (weird), due to a merge.  Retesting.

cgcgcg:
@lucbv

brian-kelley:
@mhoemmen FYI, if cuSPARSE is enabled, then each KokkosKernels CrsMatrix is automatically associated with a cusparse handle and matrix descriptor. KK also has cuSparse spmv support in TPLs. So all we'd have to do to support the on-node part is provide an interface to select the algorithm (which is currently always ``CUSPARSE_CSRMV_ALG1``).

mhoemmen:
@brian-kelley wrote:
> FYI, if cuSPARSE is enabled, then each KokkosKernels CrsMatrix is automatically associated with a cusparse handle and matrix descriptor. KK also has cuSparse spmv support in TPLs. So all we'd have to do to support the on-node part is provide an interface to select the algorithm (which is currently always `CUSPARSE_CSRMV_ALG1`).

Hi Brian!  Thanks for pointing this out!  I'm aware of this -- I had a chat with @lucbv about it last week.  There are a few reasons why I did things the way I did them:

1. Tpetra will need to change anyway, in order to decide which cuSPARSE algorithm to use.  See e.g., `CrsGraph::nodeMinNumRowEntries_` and `setMinMaxNumberOfEntriesPerRow` in this PR.  I'm not quite done with that yet, but you'll see where this is going : - ) .

2. Tpetra will need to store additional data, like the kokkos-kernels handle and a copy of the row offsets array as an `int`.  (`Tpetra::CrsMatrix`'s `KokkosSparse::CrsMatrix` currently uses a different unsigned type as the row offsets type.  This is not compatible with cuSPARSE.  However, we can't change that in CrsMatrix, because it's a public typedef and thus exposed to applications.)

3. It's critical not to expose downstream application code to TPL headers. Doing so has caused Sierra build errors in the past.  I took special care not to do this here.

mhoemmen:
I had a nice CUDA build going, but the evening's SPARC tests seem to be taking all its memory now, so I can't run anything at the moment.  My last two commits (just pushed) reset the PR tests.

mhoemmen:
My last commit (just pushed) makes `Tpetra::CrsMatrix::apply` use cuSPARSE's merge path algorithm for `Scalar=float` or `Scalar=double`, and `mode=Teuchos::NO_TRANS`.  Please don't merge this quite yet; the intent is to see whether that algorithm can actually make it through Trilinos' tests.

mhoemmen:
Hm, testing.sandia.gov is timing out.

mhoemmen:
@trilinos/framework I didn't see any of those builds listed above on the Dashboard.  Weird.  Anyway, I'll try a retest.

mhoemmen:
Ooo, hold on, I need to disable the `Scalar=float` specializations if `HAVE_TPETRA_INST_FLOAT` is not defined.

mhoemmen:
@vbrunini @lucbv Weird, cuSPARSE appears to get the wrong answer if the matrix is empty.  I added a special case for that and got Tpetra's tests to pass.

mhoemmen:
MueLu appears to create a `Tpetra::CrsMatrix` in a way that bypasses my local operator's `fillComplete` call.  I was able to manifest these issues with the `MueLu_SimpleTpetra_MPI_4` test.  I'm remarking this as WIP, and I'll push more debug checking and verbose output code.

mhoemmen:
@vbrunini says that with CUDA 10.1, everything works if you tell cuSPARSE to use the "naïve" algorithm.

kyungjoo-kim:
@mhoemmen At the end, the merge based algorithm does not solve the problem. Does cusparse work better than the current spmv impl (whatever is used in the develop branch) in Tpetra ?

vbrunini:
@kyungjoo-kim Are you saying you tested the cusparse merge based algorithm on a matrix with some dense rows similar to the one that motivated this and have confirmed that it also performs badly?

kyungjoo-kim:
@vbrunini I did not test it. Oh... I misinterpreted @mhoemmen 's comment. I thought that you told him after you tested both naive and merge algorithms. It looks like that the merge algorithm does not work in your test.

vbrunini:
Ok, got it. I did actually test both algorithms. The naive algorithm runs the test correctly, I did not compare performance with the current kokkos-kernels spmv algorithm. The merge algorithm runs the first time step successfully and appears to perform significantly better there, but then we start seeing correctness issues on future time steps. At the moment it is unclear if that is from an issue in cusparse itself, or in the tpetra interface to it in this pull request.

srajama1:
@lucbv : FYI .. 
@srajama1

mhoemmen:
Superseded by PR https://github.com/trilinos/Trilinos/pull/7021 .

PR #6983, opened 2020-03-09
Bring compareTestOutput up to python 2.7 or above

prwolfe:
This was failing on python 3 testing in prep
for that becoming the default in RHEL 8.
This now runs in 2.7 and 3.6

@trilinos/ml 
@trilinos/framework 

## Motivation
Several of the constructs in this script were built for much older python that is not in use and those are no longer available in python 3. As that will be the default starting in RHEL 8 we are working to run with that set now rather than at the last minute.

## Related Issues

* Blocks #6878 

## Testing
This was tested separately on my blade and as part of testing for #6878 in my branch as that PR has bootstrapping issues.

## Additional Information

### DIFF
--- a/commonTools/test/utilities/compareTestOutput
+++ b/commonTools/test/utilities/compareTestOutput
@@ -16,8 +16,8 @@ import getopt
 
 
 version = sys.version_info[0] + sys.version_info[1]/10.0
-if version < 2.7:
-    sys.exit("python 2.7 or greater required!")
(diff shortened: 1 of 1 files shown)


jwillenbring:
@jhux2 The @trilinos/framework wanted to offer a chance to look at this PR prior to applying it. Please either approve, ask someone else to, or indicate that we should just move forward.

Thanks.

Issue #6987, opened 2020-03-09
Tpetra: Map/Directory min GID uninitialized in some circumstances


tjfulle:
## Bug Report
@trilinos/tpetra

### Description
`Tpetra::Map::myMinGID_` is the minimum GID held by a map on its owning process.  For a map that has no entries on a process, its value is left uninitialized.

The `Tpetra::Details::Directory::allMinGIDs_` array holds the minimum GID on each participating process.  This array is used, eg, in the directory's `getEntriesImpl` method.  For ranks whose maps are empty `allMinGIDs_[rank]` is uninitialized and any work using it triggers undefined behavior.

@rppawlo discovered this in a simple test where in a map was empty on every process *except* the last.  Use that map to build a `Tpetra::MultiVector` and then import that `MultiVector` to another `MultiVector` with a locally replicated map.  The result is failure to import the values from the original vector to the locally replicated vector (because of comparisons with uninitialized values in `Directory::getEntriesImpl`.  A simple reproducer is copied below.

A potential solution is to initialize `myMinGID_ = Invalid;` and check for `Invalid` in `getEntriesImpl` as part of its algorithm.

Any other ideas @mhoemmen ?

@rppawlo does the above adequately describe the issue?

Reproducer:

```c++
TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL(Directory, AllMinGIDs, SC, LO, GO)
{
  using map_type = Tpetra::Map<LO, GO>;
  using vector_type = Tpetra::MultiVector<SC, LO, GO>;
  using import_type = Tpetra::Import<LO, GO>;

  auto comm = getDefaultComm();
  const auto my_rank = comm->getRank();
  const auto num_procs = comm->getSize();

  const GO index_base = 0;
  const size_t num_vecs = 1;

  Teuchos::RCP<vector_type> vec1;
  Teuchos::RCP<map_type> map1;
  {
    const Tpetra::global_size_t num_global_elements = 5;
    const size_t num_local_elements = (my_rank == num_procs - 1) ? 5 : 0;
    map1 = Teuchos::rcp(new map_type(num_global_elements, num_local_elements, index_base, comm));
    vec1 = Teuchos::rcp(new vector_type(map1, num_vecs, true));
    if (my_rank == 0) vec1->putScalar(Teuchos::as<SC>(5.0));
  }

  Teuchos::RCP<vector_type> vec2;
  Teuchos::RCP<const map_type> map2;
  {
[...]

PR #6992, opened 2020-03-10
Tpetra: Fixes Issue #6987

tjfulle:
This PR addresses Issue #6987 

Rather than reimplement the broken search algorithm, this PR simply starts the search from the first initialized GID.


tjfulle:
We discussed implementing a binary search instead of the pseudo-linear search already implemented.  I suppose that could still be done, but starting the linear search from the first initialized GID fixes the current issues.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
```diff
@@ -431,10 +431,25 @@ namespace Tpetra {
...
+        [](GO const gid) { return gid != std::numeric_limits<GO>::max(); }
+      );
+#ifdef HAVE_TPETRA_DEBUG
+      TEUCHOS_TEST_FOR_EXCEPTION(
+        it == allMinGIDs_.end()-1,
+        std::logic_error,
+        "Every min GID in the map is std::numeric_limits<GO>::max()!"
+      );
+#endif // HAVE_TPETRA_DEBUG
+      const GO i0 = std::distance(allMinGIDs_.begin(), it);
```
mhoemmen: You might want to make that an explicit cast, though it should be harmless as long as GO is signed.  Remember that Albany uses DTK 2.0 and thus uses `GO=unsigned long`.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
```diff
@@ -431,10 +431,25 @@ namespace Tpetra {
...
       LookupStatus res = AllIDsPresent;
 
+      // Find the first initialized GID for search below
+      auto it = std::find_if(
+        allMinGIDs_.begin(),
+        allMinGIDs_.end()-1,
+        [](GO const gid) { return gid != std::numeric_limits<GO>::max(); }
+      );
+#ifdef HAVE_TPETRA_DEBUG
+      TEUCHOS_TEST_FOR_EXCEPTION(
```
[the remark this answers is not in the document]
tjfulle: No, it shouldn't be an error.  Debugging for this specific case that should be cleaned out.

mhoemmen:
@tjfulle Would you like to merge this for now, or fix the possible signed/unsigned issue?

tjfulle:
@mhoemmen I just pushed an update, I'll merge when testing completes.

rppawlo:
Thanks @tjfulle ! I'll run the new test once this gets pushed.

tjfulle:
one last update @mhoemmen.  i check to see if the map is empty and skip the search if it is.  can you check the logic?

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
```diff
@@ -431,10 +431,28 @@ namespace Tpetra {
 
       RCP<const Teuchos::Comm<int> > comm = map.getComm ();
       const int numProcs = comm->getSize ();
-      const global_size_t nOverP = map.getGlobalNumElements () / numProcs;
       const LO LINVALID = Teuchos::OrdinalTraits<LO>::invalid();
       LookupStatus res = AllIDsPresent;
 
+      // Find the first initialized GID for search below
+      auto it = std::find_if(
```
mhoemmen: This should be OK, since line 303 of `Tpetra_DirectoryImpl_def.hpp` always initializes `allMinGIDs_` to size `numProcs + 1`.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
```diff
@@ -431,10 +431,28 @@ namespace Tpetra {
...
+        allMinGIDs_.begin(),
+        allMinGIDs_.end()-1,
+        [](GO const gid) { return gid != std::numeric_limits<GO>::max(); }
+      );
+
+      if (it == allMinGIDs_.end()-1) {
+        // No entries have been assigned on this process
```
mhoemmen: Not just on this process -- on any process, right?  This should be what happens when you have an empty Map that you created with the contiguous nonuniform constructor.

More importantly, we may want to check this with an actual empty Map, created with Map's default constructor.  It sets `minMyGID_` to invalid, not the max.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
```diff
@@ -431,10 +431,28 @@ namespace Tpetra {
...
+        res = (globalIDs.size() > 0) ? IDNotPresent : AllIDsPresent;
+        std::fill(nodeIDs.begin(), nodeIDs.end(), -1);
+        if (computeLIDs)
+          std::fill(localIDs.begin(), localIDs.end(), LINVALID);
+        return res;
+      }
+
+      const int i0 = as<GO>(std::distance(allMinGIDs_.begin(), it));
```
mhoemmen: Woah, you have a triple cast here -- iterator distance, to `GO`, to `int`.  You could just assign to `int` directly (with a `static_cast`), without even the `as`, since `allMinGIDs.size() == numProcs + 1` (and `MPI_Comm_size` returns an `int`).  This is harmless, but it could result in an unsigned-to-signed assignment warning, if GO is unsigned.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
```diff
@@ -431,10 +431,28 @@ namespace Tpetra {
...
+      const global_size_t nOverP = map.getGlobalNumElements () / (numProcs - i0);
```
mhoemmen: Consider casting `numProcs - i0` to `global_size_t` first.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
```diff
@@ -431,10 +431,28 @@ namespace Tpetra {
...
+
       // Map is distributed but contiguous.
       typename ArrayView<int>::iterator procIter = nodeIDs.begin();
```
mhoemmen: `auto` should work fine here.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
```diff
@@ -452,7 +470,7 @@ namespace Tpetra {
           const GO one = as<GO> (1);
           const GO two = as<GO> (2);
           const GO nOverP_GID = as<GO> (nOverP);
-          const GO lowerBound = GID / std::max(nOverP_GID, one) + two;
+          const GO lowerBound = i0 + GID / std::max(nOverP_GID, one) + two;
```
mhoemmen: Why did Tpetra add `two` here?  (I actually don't know; I just found this here and fixed the casting mistakes long ago.)  It seems like this change could make the `while` loop below break early, claiming that it couldn't find the GID.  We could just delete the `two`; the loop should still work fine.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
tjfulle: ya, typo.

packages/tpetra/core/src/Tpetra_DirectoryImpl_def.hpp
tjfulle: i don't know why the two is there either.  it is also in the Epetra version.

tjfulle: when i changed the search to use a bisection it broke a number of other Tpetra tests.  so, I just left the search as is, only changing the lower bound of where the search could start.

tjfulle: So, I removed the `two`, but a Albany labeled Tpetra test broke.  I didn't look in to it further.

PR #7001, opened 2020-03-12
shylubasker: fix shylubasker transpose solve correctness in amesos2

ndellingwood:
Added example to test through Amesos2

## Testing
Tested in local build to confirm changes address transpose solve correctness issues when calling shylubasker through amesos2

### DIFF
--- a/packages/amesos2/example/SimpleSolveTranspose.cpp
+++ b/packages/amesos2/example/SimpleSolveTranspose.cpp
@@ -1,0 +1,212 @@
+// @HEADER
+//
+// ***********************************************************************
+//
+//           Amesos2: Templated Direct Sparse Solver Package
+//                  Copyright 2011 Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the Corporation nor the names of the
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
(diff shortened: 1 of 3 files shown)


ndellingwood:
@hkthorn this is one more small PR to fix correctness of transpose solves when calling shylubasker through amesos2. This PR does not include a couple other issues you've pointed out that need addressing (I wanted to fix transpose solve correctness first with the solver), a follow on PR will address the inconsistency of setting transpose mode in Amesos2, and remove need to create different solver instances for basker in cases when a user will want to compute forward and transpose solves.

packages/amesos2/example/SimpleSolveTranspose.cpp
```diff
@@ -0,0 +1,218 @@
...
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact Michael A. Heroux (maherou@sandia.gov)
+//
+// ***********************************************************************
+//
+// @HEADER
+
+/**
```
srajama1: Cut and paste, may be ?

packages/amesos2/example/SimpleSolveTranspose.cpp
```diff
@@ -0,0 +1,218 @@
...
+  typedef Tpetra::CrsMatrix<Scalar,LO,GO> MAT;
+  typedef Tpetra::MultiVector<Scalar,LO,GO> MV;
+
+  using Tpetra::global_size_t;
+  using Teuchos::tuple;
+  using Teuchos::RCP;
+  using Teuchos::rcp;
+
+
+  // Before we do anything, check that SuperLU is enabled
```
srajama1: Another cut and paste ?

packages/amesos2/example/SimpleSolveTranspose.cpp
ndellingwood: Yes, cleaned up in newest commit I just pushed, good catch.

packages/amesos2/example/SimpleSolveTranspose.cpp
ndellingwood: Yes, just cleaned this up in recent commit and cleaned up the commit history.

ndellingwood:
PR failed because I pushed changes while the autotester was running, next run through should be clean.

ndellingwood:
Needed to namespace a use of "endl" in the example I added to amesos2, not sure why it built fine for me locally. Commit with fix pushed, retesting.

ndellingwood:
Thanks @hkthorn !

PR #7003, opened 2020-03-12
Tempus: ERK - test stage number accessor

sconde:
@trilinos/tempus 

## Motivation

- add test for stage number accessor
- move `getStageX` to `StepperRKBase`

## Testing

ERK test stage number accessor

### DIFF
--- a/packages/tempus/src/Tempus_StepperDIRK_impl.hpp
+++ b/packages/tempus/src/Tempus_StepperDIRK_impl.hpp
@@ -31,7 +31,7 @@ void StepperDIRK<Scalar>::setupDefault()
   this->setUseEmbedded(        this->getUseEmbeddedDefault());
   this->setZeroInitialGuess(   false);
 
-  this->setStageNumber(0);
+  this->setStageNumber(-1);
 
   stepperObserver_ = Teuchos::rcp(new StepperRKObserverComposite<Scalar>());
   this->setDefaultSolver();
@@ -55,7 +55,7 @@ void StepperDIRK<Scalar>::setup(
   this->setUseEmbedded(        useEmbedded);
   this->setZeroInitialGuess(   zeroInitialGuess);
 
-  this->setStageNumber(0);
+  this->setStageNumber(-1);
 
   stepperObserver_ = Teuchos::rcp(new StepperRKObserverComposite<Scalar>());
   this->setObserver(obs);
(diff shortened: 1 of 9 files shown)


packages/tempus/src/Tempus_StepperDIRK_impl.hpp
```diff
@@ -55,7 +55,7 @@ void StepperDIRK<Scalar>::setup(
   this->setUseEmbedded(        useEmbedded);
   this->setZeroInitialGuess(   zeroInitialGuess);
 
-  this->setStageNumber(0);
+  this->setStageNumber(-1);
```
krcb: This is a good idea. Is there any scheme where -1 would be valid? We could use std::invalid as an alternative here.

ccober6:
@sconde Should we set the stage number to -1 at the end of takeStep() to clearly indicate that we are outside the stage loop?

packages/tempus/src/Tempus_StepperRKBase.hpp
```diff
@@ -21,13 +24,16 @@ class StepperRKBase
 
 public:
 
-  virtual int getStageNumber() { return stageNumber_; }
+  virtual int getStageNumber() const { return stageNumber_; }
 
   virtual void setStageNumber(int s) { stageNumber_ = s; }
 
+  virtual Teuchos::RCP<Thyra::VectorBase<Scalar> > getStageX() {return stageX_;}
```
krcb: I like moving this down to the base class. Can we move it further down so that we can unify getting this across the steppers? The default would be to either do nothing or throw an exception so that the user know that they are using this incorrectly.

sconde:
> Should we set the stage number to -1 at the end of takeStep() to clearly indicate that we are outside the stage loop?

I reset the stageNumber to -1 to indicate that the stepper is out of "takeStep". Did I miss something?

krcb:
@sconde I left a couple of comments. I think these are useful/good improvements as it stands, so I've approved.

packages/tempus/src/Tempus_StepperDIRK_impl.hpp
sconde: All RK schemes are multistage ( s >= 1 ) methods. And we can consider BDF schemes a single stage ( s = 1 ), multi-step schemes ( k >= 1 ). So to my knowledge, I don't think any scheme can have a valid s = -1.

ccober6:
@sconde Sorry I miss read it.  It looks good!  I see that @krcb already approved.

packages/tempus/src/Tempus_StepperRKBase.hpp
ccober6: @krcb I would not move RK specific functions and data into the Stepper class.  I hear you about wanting to unify across steppers but I think it will cause more confusion for other non-RK steppers.  It seems the driver to unify is the new AppActions, but they will be specific to each stepper, so unifying would not gain much.

PR #7012, opened 2020-03-13
WIP: Thrya: Fix some unreachable code with debug builds

bartlettroscoe:
## Next Action Status

These warnings are generated in a SPARC build against Trilinos using cuda+gnu on the CEE ASCIC GPU machines.  Next: Wait to add support for CUDA builds on the CEE ASCIC GPU machines as pat of the 'cee-rhel6' ATDM Trilinos configuration (see [ATDV-272](https://sems-atlassian-srn.sandia.gov/browse/ATDV-272)) and then reproduce and find a better patch to silence warnings without impacting code maintenance ...

## Description

Made by SPARC developers to silence warnings impacting SPARC builds.  The change was to change switch statement functions like:

```
inline
std::string Thyra::toString(ModelEvaluatorBase::EInArgsMembers arg)
{
  switch(arg) {
    case ModelEvaluatorBase::IN_ARG_x_dot_dot:
      return "IN_ARG_x_dot_dot";
    <...>
    case ModelEvaluatorBase::IN_ARG_stage_number:
      return "IN_ARG_stage_number";
#ifdef TEUCHOS_DEBUG
    default:
      TEUCHOS_TEST_FOR_EXCEPT(true);
#else
  }
  return ""; // Will never be executed!
}
```

and replaces them with:

```
inline
std::string Thyra::toString(ModelEvaluatorBase::EInArgsMembers arg)
{
  switch(arg) {
    case ModelEvaluatorBase::IN_ARG_x_dot_dot:
      return "IN_ARG_x_dot_dot";
    <...>
    case ModelEvaluatorBase::IN_ARG_stage_number:
      return "IN_ARG_stage_number";
    default:
#ifdef TEUCHOS_DEBUG
      TEUCHOS_TEST_FOR_EXCEPT(true);
#else
      return ""; // Will never be executed!
#endif
  }
}
```

The problem with the latter is that the compiler will no longer issue a waning if all of the cases are not covered because of that `default` clause.  That makes the code much more fragile to  continued maintenance as it will silence these compiler warnings.  But in a debug-mode build, we want to have a `default` clause that throws to avoid undefined behavior.

@micahahoward, what specific compile and build does this trigger warnings for?  It would be good to know that so we can find a solution that does not add a `default` block in the non-debug-mode build.

## Steps to Reproduce

Waiting for to add CUDA support for ATDM Trilinos 'cee-rhel6' configuration ...

### DIFF
--- a/packages/thyra/core/src/interfaces/nonlinear/model_evaluator/fundamental/Thyra_ModelEvaluatorBase_decl.hpp
+++ b/packages/thyra/core/src/interfaces/nonlinear/model_evaluator/fundamental/Thyra_ModelEvaluatorBase_decl.hpp
@@ -1408,12 +1408,13 @@ std::string Thyra::toString(ModelEvaluatorBase::EInArgsMembers arg)
       return "IN_ARG_step_size";
     case ModelEvaluatorBase::IN_ARG_stage_number:
       return "IN_ARG_stage_number";
-#ifdef TEUCHOS_DEBUG
     default:
+#ifdef TEUCHOS_DEBUG
       TEUCHOS_TEST_FOR_EXCEPT(true);
+#else
+      return ""; // Will never be executed!
 #endif
   }
-  return ""; // Will never be executed!
 }
 
 
@@ -1435,12 +1436,13 @@ std::string Thyra::toString(ModelEvaluatorBase::EOutArgsMembers arg)
       return "OUT_ARG_W_prec";
     case ModelEvaluatorBase::OUT_ARG_f_poly:
       return "OUT_ARG_f_poly";
-#ifdef TEUCHOS_DEBUG
     default:
+#ifdef TEUCHOS_DEBUG
       TEUCHOS_TEST_FOR_EXCEPT(true);
+#else
+      return ""; // Will never be executed!
 #endif
   }
-  return ""; // Will never be executed!
 }
 
 
@@ -1454,12 +1456,13 @@ std::string Thyra::toString(
       return "DERIV_MV_BY_COL";
     case ModelEvaluatorBase::DERIV_TRANS_MV_BY_ROW:
       return "DERIV_TRANS_MV_BY_ROW";
-#ifdef TEUCHOS_DEBUG
     default:
+#ifdef TEUCHOS_DEBUG
       TEUCHOS_TEST_FOR_EXCEPT(true);
+#else
+      return ""; // Should never execute this!
 #endif
   }
-  return ""; // Should never execute this!
 }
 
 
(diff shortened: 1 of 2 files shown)


bartlettroscoe:
It would be good to find another way to silence these warnings because we don’t want to add the ‘default’ block in a non-debug build because if there is a missing case, we want the compiler to warn us about this.  The issue is that different compilers will generate different warnings “unreachable code” and “not all paths may return a value” (which is why these functions had an unreachable return because that silenced warnings for those compilers).

bartlettroscoe:
@micahahoward,

Looking over the SPARC Trilinos Integration builds since 2020-02-01 with any build warnings shown in this [SPARC CDash query](https://sems-cdash-srn.sandia.gov/sparc/index.php?project=SPARC&begin=2020-02-01&end=2020-03-13&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Trilinos%20Integration&field2=configureerrors&compare2=41&value2=0&field3=builderrors&compare3=41&value3=0&field4=buildwarnings&compare4=43&value4=0) I can't find any warnings coming from Thyra header files.

What build were you seeing these warnings coming from?

bartlettroscoe:
Micah reported that these warnings were getting generated in a CUDA build of SPARC pulling in these header files.  I will try to reproduce those CUDA/nvcc build warnings with a SPARC build on 'vortex'.

Also, I will take off the `AT: WIP` label to let the PR tester test this to see what types of warnings this generates in other compilers, if any.

bartlettroscoe:
FYI: There are no Thyra warnings coming from the PR builds:

* https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-7012-test&field2=buildstarttime&compare2=84&value2=NOW&field3=subprojects&compare3=93&value3=Thyra

bartlettroscoe:
@micahahoward,

Specifically, what machine and CUDA build configuration did you see these warnings?  I could not generate any Thyra warnings when building SPARC on ATS-2 'vortex' using `cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24`.  I need to be able to reproduce these build warnings (without this patch) so that I can fix the warnings in a way that does not impact maintainability of the code.

micahahoward:
@bartlettroscoe, ASCICGPU systems, GCC/CUDA debug builds

bartlettroscoe:
> ASCICGPU systems, GCC/CUDA debug builds

@micahahoward, we don't have our CEE ATDM Trilinos builds supporting a CUDA build just yet.  Therefore, I will address this once we get support for that added as per [ATDV-272](https://sems-atlassian-srn.sandia.gov/browse/ATDV-272).

In the meantime, the SPARC Trilinos 'master' branch has workaround with this current patch which is just fine.

Thanks!

bartlettroscoe:
Added `AT: WIP` to try to get this to stop sending out emails everyday.

e10harvey (approved):

bartlettroscoe:
@e10harvey, why did you approve this PR?  It is marked "WIP" and see [above](https://github.com/trilinos/Trilinos/pull/7012#issue-387839657) "Next Action Status". 

This PR should not be merged but a better solution should be implemented that makes the warnings go away but we need to reproduce the warnings first (and now we can with the ATDM Trilinos 'cee-rhel7' CUDA build configuration).

e10harvey:
The changes look OK to me.

PR #7015, opened 2020-03-13
Anasazi:  remove deprecated use of Node object

kddevin:
@trilinos/anasazi 

## Motivation
Anasazi's TpetraTsqrTest used Node object in a deprecated way.  This PR removes that usage.

However, TpetraTsqrTest is not used **anywhere**.  
@mhoemmen @hkthorn  Should I remove that file?

## Stakeholder Feedback
NA

## Testing
Anasazi and Tpetra's TSQR build correctly on my mac with
```
module load sems-gcc/6.4.0
```
All tests pass.

The code changes in TsqrTpetraTest.hpp are not tested, however, as the file is not used anywhere.  Perhaps the file should be deleted.

### DIFF
--- a/packages/anasazi/tpetra/src/TsqrTpetraTest.hpp
+++ b/packages/anasazi/tpetra/src/TsqrTpetraTest.hpp
@@ -64,7 +64,6 @@ namespace TSQR {
         typedef S scalar_type;
         typedef LO local_ordinal_type;
         typedef GO global_ordinal_type;
-        typedef Node node_type;
 
         typedef typename TSQR::ScalarTraits< S >::magnitude_type magnitude_type;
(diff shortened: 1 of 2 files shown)


kddevin:
Thanks for the review, @mhoemmen 
No, this test does not build at all.  File TsqrTpetraTest.hpp is not used anywhere.  Should I remove it?

mhoemmen:
@kddevin wrote:
> No, this test does not build at all. File TsqrTpetraTest.hpp is not used anywhere. Should I remove it?

I would say yes, though it would be better to ask folks @hkthorn, @jennloe, and @iyamazaki.

hkthorn:
@kddevin Sure, remove it.

PR #7021, opened 2020-03-15
Tpetra::CrsMatrix: Use cuSPARSE's merge path SpMV algorithm where possible

mhoemmen:
@trilinos/tpetra 

This supersedes PR https://github.com/trilinos/Trilinos/pull/6982 .

Please don't merge this yet; I just would like to see if it passes PR tests.

## Motivation

Local sparse matrix-vector multiply has poor load balance on GPUs for matrices that have a few rows with many entries, and many rows with few entries.  cuSPARSE's merge path algorithm promises to help with that.  This PR offers some steps towards making that work.

## Stakeholder Feedback

Aria (see e.g., @vbrunini ) very much needs this to be done this FY.

## Testing

On a GPU machine.  I actually added tests in Tpetra for existing functionality.


lucbv:
@mhoemmen 
I am commenting so I can follow this PR, as you know I am quite interested in this : )

mhoemmen:
I haven't gotten Trilinos' develop branch to pass most of MueLu's tests in a CUDA build, so I'm not sure what wizardry @trilinos/muelu uses to make it work.  I get the same nonsense that y'all see below, in random places that have no apparent relation to my changes:
```
  what():  cudaDeviceSynchronize() error( cudaErrorAssert): device-side assert triggered .../Trilinos/packages/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:138
```

mhoemmen:
@lucbv wrote:
> I am commenting so I can follow this PR, as you know I am quite interested in this : )

Do you see the MueLu test failures that I've been seeing?  My sense is that cuSPARSE is buggy and corrupting memory somewhere.  @vbrunini wasn't able to make `cuda-memcheck` work on Aria, but perhaps it could work on a simpler test.

I wish I could say it was my cuSPARSE wrappers, but those tests pass just fine, even when I change the matrix's values.  However, they use small (though nonsymmetric) matrices.  Have you had good results when running cuSPARSE's merge path algorithm with large matrices?

lucbv:
We have not tried to integrate with Trilinos yet but will likely try soon. We will at least see if the default spam algorithm creates problems or not?
I'll keep you posted if I see anything strange in cusparse or MueLu, but with the default Kokkos-kernels implementation MueLu seems fine.

mhoemmen:
@lucbv wrote:
> but with the default Kokkos-kernels implementation MueLu seems fine.

I'm not sure what that means -- kokkos-kernels is not using cuSPARSE, right?

vbrunini:
@mhoemmen Can you try switching this branch to always use the naive cusparse algorithm and having it re-run the PR testing?

mhoemmen:
@vbrunini wrote:
> Can you try switching this branch to always use the naive cusparse algorithm and having it re-run the PR testing?

I'll give that a try!

mhoemmen:
@vbrunini Only one, totally unrelated (since upstream), KokkosCore unit test failure.  Let's retest.

vbrunini:
So that pretty strongly suggests this is a cuSparse bug with the merge path algorithm right? Can we try cuda 10.2 or get some nvidia assistance with this?

mhoemmen:
@vbrunini wrote:
> So that pretty strongly suggests this is a cuSparse bug with the merge path algorithm right?

I'd like to try some more matrices first, but I think you're right.

> Can we try cuda 10.2 or get some nvidia assistance with this?

Do we have any systems with CUDA 10.2 installed, or at least a big enough 10.1 revision that we can try the new interface?

mhoemmen:
btw I have a minor build error fix for CUDA 10.2 ready, but I'm unable to link because of cuBLAS link errors (that are likely unrelated to these changes).

lucbv:
@mhoemmen both Sierra (Lassen) and Summit provide the appropriate Cuda version to test the new interface in cusparse. They might be good platform to test things.

lucbv:
Regarding the Cuda header the issue appears if you use cuda/10.1.168, this was fixed with cuda/10.1.243 which is the version I use to test the new interface on Lassen

mhoemmen:
@kddevin I need to fix the CUDA > 10.1 build error first, and test with the new interface.  I've been having troubles with the CUDA 10.2 module on my system, so I'm trying now with as new of a CUDA 10.1 module as I can find, per @lucbv's instructions.

mhoemmen:
I'm testing my fixes with the new (CUDA >= 10.1) interface now, using CUDA 10.1.243.  They build correctly -- I'm running Tpetra's tests now.

mhoemmen:
Weirdly, in my CUDA 10.1 build, with either the old or new cuSPARSE interface, I get a crash in `TpetraExt_MatrixMatrix_def.hpp`, near the end of `mult_A_B_newmatrix`.  I will add verbose output to trace the issue.

mhoemmen:
It's definitely crashing in `KokkosSparse::Experimental::spgemm_numeric`, in the CUDA specialization of `mult_A_B_newmatrix_kernel_wrapper` that lives in `TpetraExt_MatrixMatrix_Cuda.hpp`.

@jhux2 @csiefer2 @brian-kelley @kyungjoo-kim does this not normally run in the PR test builds?

brian-kelley:
@mhoemmen You're now seeing this on a Volta machine, right? The CUDA version of SPGEMM does run in PR builds, but on CUDA 9.2 and Kepler.

There are a few places inside SPGEMM where Volta and non-Volta take different paths, but KokkosKernels does regular testing on Volta.

mhoemmen:
@brian-kelley wrote:
> You're now seeing this on a Volta machine, right?

Yes, it's ascicgpu22, a machine with 4 Volta GPUs.  I'm just wondering why my changes make the one Tpetra test fail on my machine, but not in the PR test builds.  It's weird, and I'd like to understand it before going to NVIDIA with what we think is a merge path bug.

brian-kelley:
@mhoemmen So the product from this SPGEMM is the matrix that's passed into merge path SPMV later?

If the cuSPARSE TPL is enabled and the size_type/lno_t are the same (int32 or int64), KokkosKernels actually uses SPGEMM_CUSPARSE as the default algorithm. But Tpetra seems to use SPGEMM_KK_MEMORY by default, so I assume that's what you're getting. When I run the MatrixMatrix unit tests on my machine with CUSPARSE enabled, Tpetra still uses KK_MEMORY.

I'm not sure if you have a debug build handy but if you want to know for certain, stick a breakpoint at KokkosSparse_spgemm_cuSPARSE_impl.hpp:190. If that breakpoint is hit, KokkosKernels is calling CUSPARSE SPGEMM. Otherwise, I don't know what's going on - even this case isn't covered by PR testing, volta + Tpetra testing is done nightly on Waterman.

mhoemmen:
@brian-kelley btw I reconfigured on this machine with `KOKKOS_ARCH_VOLTA70=ON` instead of PASCAL60, and the test passed (!).  I'll keep looking.

@trilinos/tpetra @vbrunini I just pushed commits that make the Tpetra tests pass with the "new" interface, and use the new interface with CUDA 10.1 or greater (not 10.2 or greater as before).

brian-kelley:
@mhoemmen Oh, interesting. In principle code for pascal should run on volta, but in practice who knows, we don't test it.

Anyway, Tpetra is definitely calling SPGEMM_KK_MEMORY.

mhoemmen:
My MueLu tests pass now with the "naïve" algorithm.  I'll try now with merge path.

mhoemmen:
If this passes @vbrunini 's tests and improves solve time (preliminary results suggest yes to both), then I plan to change this so that it only uses cuSPARSE with CUDA 10.1 or greater.  This means that the PR tests won't actually exercise cuSPARSE, since the PR tests' CUDA build uses CUDA 9 (see #7047).

@lucbv take note -- it looks like it's not even worth supporting the old interface.

mhoemmen:
@vbrunini reports that this passes tests and improves `CrsMatrix::apply` time dramatically.

We'll probably need to change the fused residual in `Tpetra::Details::residual` (used by Jacobi in `Ifpack2::Relaxation`) to defer to `CrsMatrix::apply` if doing so would be faster (because cuSPARSE's merge path algorithm would be faster).  This will require a heuristic for deciding which option is faster.  It must be a globally consistent heuristic (over all MPI processes in the matrix's communicator), since the fused residual kernel uses different MPI communication.

Victor suggests using the ratio (max number of entries in a row) / (min number of entries in a row) as the heuristic.  This PR makes CrsGraph compute the min number of entries per row _locally_, but CrsGraph doesn't compute the global min yet.  We could accomplish that in `CrsGraph::computeGlobalConstants` without adding another `MPI_Allreduce`, if we add another entry to the `globalMaxNumRowEntries_` all-reduce, use signed integers instead of unsigned, and do the `min(x) == -max(-x)` trick.

Another heuristic would be simply the maximum number of entries globally in any row.  CrsGraph already computes that.

mhoemmen:
Trilinos' tests pass for me, if I use cuSPARSE's merge path algorithm unconditionally:
```
100% tests passed, 0 tests failed out of 1548

Subproject Time Summary:
Amesos         =  24.95 sec*proc (13 tests)
Amesos2        = 112.33 sec*proc (6 tests)
Anasazi        = 587.33 sec*proc (77 tests)
AztecOO        =  24.41 sec*proc (17 tests)
Belos          = 2004.04 sec*proc (106 tests)
Epetra         =  71.51 sec*proc (63 tests)
EpetraExt      =  21.49 sec*proc (10 tests)
Ifpack         =  78.55 sec*proc (48 tests)
Ifpack2        = 973.03 sec*proc (44 tests)
Kokkos         = 226.68 sec*proc (28 tests)
ML             =  63.65 sec*proc (34 tests)
MueLu          = 2643.68 sec*proc (64 tests)
NOX            = 314.15 sec*proc (106 tests)
Sacado         = 384.47 sec*proc (302 tests)
Stratimikos    =  29.30 sec*proc (39 tests)
Teko           = 610.12 sec*proc (19 tests)
Teuchos        = 176.96 sec*proc (139 tests)
Thyra          = 110.57 sec*proc (82 tests)
Tpetra         = 4497.19 sec*proc (239 tests)
Zoltan2        = 2980.35 sec*proc (113 tests)
```

kyungjoo-kim:
@mhoemmen Good news that cusparse works well.

lucbv:
@mhoemmen that's indeed good news, I will look at the details of how you assert that the inputs passed to cusparse are valid according to cusparse's assumptions.
This result probably means that I need to work on my handle idea soon so that we can allow the users (or Tpetra) to choose the algorithms used by cusparse.

mhoemmen:
@kyungjoo-kim wrote:
> @mhoemmen Good news that cusparse works well.

Thank you for suggesting the merge path algorithm!  I'm very happy to let vendors do at least a little work for us ; - ) .

mhoemmen:
# Summary of current status:

I consider this PR safe to merge.  Ifpack2 needs a bit of optimization; see below.

## Features of this PR

  - @vbrunini observed "dramatic" speedups for `Tpetra::CrsMatrix::apply`.
  - PR uses cuSPARSE's SpMV only for CUDA 10.1 or greater, thus avoiding the CUDA 9.2 issues we observed.
  - PR only uses cuSPARSE for `Scalar={double, float}`, thus avoiding alignment issues for CUDA's native complex types vs. `Kokkos::complex` and `std::complex`.
  - PR uses the merge path algorithm for the non-transpose case, and the standard ("naïve") algorithm otherwise.
  - The cuSPARSE wrapper correctly handles the special cases of zero dimensions and/or zero entries in the sparse matrix.
  - The "local operator with resumeFill / fillComplete" architecture can easily be extended to support other third-party implementations, such as MKL's inspector-executor interface.  It insulates `Tpetra::CrsMatrix` from the implementation details of specific third-party libraries.
  - PR fully shields downstream code from third-party library includes.  This avoids future issues like https://github.com/trilinos/Trilinos/issues/1194 .
  - Default cuSPARSE handle singleton's lifetime automatically ends with `Kokkos::finalize`.  PR's interface lets users create their own handles from `Kokkos::Cuda` instances.

## Work left to do

### Ifpack2 Jacobi

`Ifpack2::Relaxation`'s Jacobi implementation bypasses kokkos-kernels' sparse matrix-vector multiply, and uses a fused `Tpetra::Details::residual` kernel.  @vbrunini found the same load-balancing issues with this kernel, and that not using the fused kernel (and thus deferring to cuSPARSE for the mat-vec) fixes the problem.  This suggests introducing a heuristic that `Tpetra::Details::residual` can use for whether to fuse the kernel.

Victor proposed using the ratio (max number of entries in a row) / (min number of entries in a row).  `Tpetra::CrsGraph` currently calculates both quantities _locally_, not globally.  However, this may need to be a global decision, because the fused residual kernel may do MPI communication differently.  We could change `CrsGraph::computeGlobalConstants` to compute the global min as well as the global max.  We could even do this without adding an all-reduce, if we change the all-reduce to use signed integers instead of unsigned `global_size_t`, and use the same `min(x) == -max(-x)` trick that `Tpetra::Map` uses.

Another heuristic is just the maximum number of entries in a row.  CrsGraph already computes that.  Long rows probably indicate that the kernels will take a long time, so kernel fusion is less important.

mhoemmen:
@kddevin Per your request, I pushed these changes to a branch in trilinos/Trilinos with the same name, `Tpetra-LocalOperator-CuSparse-4`.  It's not possible to edit a PR to change the source branch, so if you want to pull from that branch, you'll need to make a new PR.  It currently is the same as this PR's source branch, so I'm leaving this PR open.

kddevin:
closed; superceded by #7569

PR #7023, opened 2020-03-16
Remove CTEST_NIGHTLY_START_TIME is it does nothing with git

bartlettroscoe:
As explained in the documentation:

  https://cmake.org/cmake/help/v3.13/manual/ctest.1.html?highlight=nightlystarttime#ctest-update-step

the CTest setting `NighltyStartTime` (global var `CTEST_NIGHTLY_START_TIME`) only
determines how central VCS like CVS and SVN checkout the code.  It does
nothing for DVCS like git.  Therefore, it is better to just take this out.

### DIFF
--- a/CTestConfig.cmake
+++ b/CTestConfig.cmake
@@ -1,8 +1,5 @@
 INCLUDE(SetDefaultAndFromEnv)
 
-# Must match what is in CDash project 'Trilinos'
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
@william76, can you review and approve this?  I think this is the right thing to do.

PR #7048, opened 2020-03-19
Use/test Orientation with Hierarchical basis

mperego:
@trilinos/intrepid2 

## Motivation
This PR enable the use of Orientation tools with the new nodal and hierarchical basis functions on Quads and Hexas.

## Testing
Two tests have been added to exercise this capability.


mperego:
@CamelliaDPG All good so far. Tested HGRAD, HDIV and HCURL for quads. You can start having a look at it if you like.

mperego:
@CamelliaDPG I think this is mostly done. There is one issue, the basis functions 
`DerivedNodalBasisFamilyModified<>::HGRAD_QUAD`
and
`HierarchicalBasisFamily<>::HGRAD_QUAD`
and similarly those for (`HGRAD_HEX`, `HCURL_QUAD`,...) have the same name (the string you get with `basis.getName()`).
This is a problem because the Orientation tools perform a lazy instantiation of the orientation mappings based on the name of the classes. So if you first compute the Orientation for `DerivedNodalBasisFamilyModified<>::HGRAD_QUAD` and then for `HierarchicalBasisFamily<>::HGRAD_QUAD`, the Orientation mappings won't be recomputed for the second basis because it has the same name as the first, and you would get a wrong result.

Anyway I think it is better that different basis functions have different names, but I'm not sure of a clean way to do so. Any ideas?

CamelliaDPG:
> Anyway I think it is better that different basis functions have different names, but I'm not sure of a clean way to do so. Any ideas?

@mperego, I agree that we should ensure that the different basis functions have different names.  I think the cleanest way to do so is to:
- assign a name to each of the component basis classes (the HGRAD and HVOL on the line classes); I'm pretty sure I neglected this for the hierarchical versions.
- make the tensor basis, direct sum basis classes return a name that is composed from their components.  (This should be determined and stored at construction time, I think, since the return type of `getName()` is a char pointer.)

Does/should this issue block this PR?  Or is it reasonable to check this in, and then fix the issue soon?  I can certainly implement the fix pretty soon.

mperego:
> > Anyway I think it is better that different basis functions have different names, but I'm not sure of a clean way to do so. Any ideas?
> 
> @mperego, I agree that we should ensure that the different basis functions have different names. I think the cleanest way to do so is to:
> 
>     * assign a name to each of the component basis classes (the HGRAD and HVOL on the line classes); I'm pretty sure I neglected this for the hierarchical versions.
> 
>     * make the tensor basis, direct sum basis classes return a name that is composed from their components.  (This should be determined and stored at construction time, I think, since the return type of `getName()` is a char pointer.)
> 
> 
> Does/should this issue block this PR? Or is it reasonable to check this in, and then fix the issue soon? I can certainly implement the fix pretty soon.

@CamelliaDPG Sounds like a good plan. No I don't think that this PR should be put on hold because of that.

kyungjoo-kim:
If you follow the Demkowicz approach, the orientation tools is not necessary.

CamelliaDPG:
@kyungjoo-kim, that's true, but for simplicity of implementation as well as certain performance considerations, we are not supporting Demkowicz's approach to orientations.  Our implementations are of Demkowicz's basis functions with the default orientation.

packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefMatrixData.hpp
```diff
@@ -56,36 +56,53 @@
...
     auto tagToOrdinal = basis->getAllDofOrdinal();
 
     //
     // High order HGRAD Elements
     //
 
-    if (name == "Intrepid2_HGRAD_QUAD_Cn_FEM") {
+    //std::cout << "Name: " << name << " " << order << std::endl;
+
+    if (name.find("HGRAD_QUAD") != std::string::npos) {
```
CamelliaDPG: @mperego Could you instead check getBaseCellTopology() and getFunctionSpace()?  I think this would be cleaner than the string comparison, and it also might be that my proposed string manipulations to define the names for tensor bases would end up including "HGRAD_QUAD" as a substring for say, the HGRAD_HEX basis.

mperego: Sure, I can. I didn't realize that those methods were implemented also for the old basis functions

packages/intrepid2/unit-test/Orientation/Serial/test_orientation_QUAD_newBasis.cpp
```diff
@@ -0,0 +1,62 @@
...
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact Kyungjoo Kim  (kyukim@sandia.gov), or
+//                    Mauro Perego  (mperego@sandia.gov)
+//
+// ************************************************************************
+// @HEADER
+
+/** \file test_01.cpp
+    \brief  Test for checking orientation tools for quadrialateral elements.
```
CamelliaDPG: @mperego, please edit this to make clear which bases are being tested (the new "derived" basis implementations).

packages/intrepid2/unit-test/Orientation/test_orientation_HEX_newBasis.hpp
```diff
@@ -0,0 +1,763 @@
...
+//
+// Questions? Contact Kyungjoo Kim  (kyukim@sandia.gov), or
+//                    Mauro Perego  (mperego@sandia.gov)
+//
+// ************************************************************************
+// @HEADER
+
+
+/** \file
+    \brief  Test for checking orientation tools for Hexahedral elements
```
CamelliaDPG: Similarly, please indicate which bases are being tested here.

packages/intrepid2/unit-test/Orientation/test_orientation_HEX_newBasis.hpp
```diff
@@ -0,0 +1,763 @@
...
+#include "Intrepid2_OrientationTools.hpp"
+#include "Intrepid2_HGRAD_LINE_C1_FEM.hpp"
+#include "Intrepid2_HGRAD_QUAD_C1_FEM.hpp"
+#include "Intrepid2_HGRAD_HEX_C1_FEM.hpp"
+#include "Intrepid2_HGRAD_HEX_Cn_FEM.hpp"
+#include "Intrepid2_HCURL_HEX_In_FEM.hpp"
+#include "Intrepid2_HCURL_QUAD_In_FEM.hpp"
+#include "Intrepid2_HVOL_LINE_Cn_FEM.hpp"
+#include "Intrepid2_HVOL_QUAD_Cn_FEM.hpp"
+#include "Intrepid2_PointTools.hpp"
```
CamelliaDPG: I don't think you need any of these basis includes, and I think the code would be clearer if you deleted them.  Instead of these, you probably should include:
```
#include "Intrepid2_HierarchicalBasisFamily.hpp"
#include "Intrepid2_NodalBasisFamily.hpp"
```
which define the bases you are testing.

packages/intrepid2/unit-test/Orientation/test_orientation_QUAD_newBasis.hpp
```diff
@@ -0,0 +1,709 @@
...
+//
+// Questions? Contact Kyungjoo Kim  (kyukim@sandia.gov), or
+//                    Mauro Perego  (mperego@sandia.gov)
+//
+// ************************************************************************
+// @HEADER
+
+
+/** \file
+    \brief  Test for checking orientation tools for quadrialateral elements
```
CamelliaDPG: Again, please edit to clarify which bases are being tested (and also please fix the typo in "quadrilateral" here).

packages/intrepid2/unit-test/Orientation/test_orientation_QUAD_newBasis.hpp
```diff
@@ -0,0 +1,709 @@
...
+#define INTREPID2_TEST_FOR_DEBUG_ABORT_OVERRIDE_TO_CONTINUE
+#endif
+
+#include "Intrepid2_Orientation.hpp"
+#include "Intrepid2_OrientationTools.hpp"
+#include "Intrepid2_HGRAD_LINE_C1_FEM.hpp"
+#include "Intrepid2_HGRAD_QUAD_C1_FEM.hpp"
+#include "Intrepid2_HCURL_QUAD_In_FEM.hpp"
+#include "Intrepid2_HVOL_LINE_Cn_FEM.hpp"
+#include "Intrepid2_HVOL_QUAD_Cn_FEM.hpp"
```
CamelliaDPG: Again, these includes are for bases that are not being tested here.  Instead include the following:

```
#include "Intrepid2_HierarchicalBasisFamily.hpp"
#include "Intrepid2_NodalBasisFamily.hpp"
```

mperego:
@CamelliaDPG I implemented all the changes requested. Can you have a look at the cases and approve? Note that I rebased the changes.

ARRASTACADELA (approved):

Issue #7067, opened 2020-03-22
Stokhos:  Errors while compiling with ensemble reduction


kliegeois:
## Bug Report
@trilinos/stokhos 

### Description
If the develop branch is configured and compiled with ensemble reduction, compiling errors occur:

```
[ 76%] Building CXX object packages/stokhos/src/CMakeFiles/stokhos_muelu_mp_8_openmp.dir/MueLu_BlackBoxPFactory_MP_Vector_8_OpenMP.cpp.o
In file included from /home/kim/dev/Trilinos_ER/Trilinos/packages/teuchos/numerics/src/Teuchos_SerialQRDenseSolver.hpp:54:0,
                 from /home/kim/dev/Trilinos_ER/Trilinos/packages/stokhos/src/sacado/kokkos/vector/Teuchos_SerialQRDenseSolver_MP_Vector.hpp:49,
                 from /home/kim/dev/Trilinos_ER/Trilinos/packages/stokhos/src/sacado/kokkos/vector/Stokhos_Sacado_Kokkos_MP_Vector.hpp:65,
                 from /home/kim/dev/Trilinos_ER/Trilinos/packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_MP_Vector.hpp:49,
                 from /home/kim/dev/Trilinos_ER/TrilinosERB/packages/stokhos/src/Stokhos_Tpetra_ETI_Helpers_MP_Vector_8_OpenMP.hpp:43,
                 from /home/kim/dev/Trilinos_ER/TrilinosERB/packages/stokhos/src/MueLu_BlackBoxPFactory_MP_Vector_8_OpenMP.cpp:51:
/home/kim/dev/Trilinos_ER/Trilinos/packages/teuchos/numerics/src/Teuchos_SerialDenseSolver.hpp: In instantiation of ‘int Teuchos::SerialDenseSolver<OrdinalType, ScalarType>::applyRefinement() [with OrdinalType = int; ScalarType = Sacado::MP::Vector<Stokhos::StaticFixedStorage<int, double, 8, Kokkos::OpenMP> >]’:
[...]

PR #7068, opened 2020-03-23
Stokhos: maintain ER build

kliegeois:
@trilinos/stokhos

## Motivation

This PR addresses the issue #7067.

The changes to the file `packages/stokhos/src/sacado/kokkos/vector/Teuchos_LAPACK_MP_Vector.hpp` address the compilation errors listed in the issue.

The change to the `cmake/std/PullRequestLinuxGCC4.8.4TestingSettings.cmake` has the purpose of testing the ensemble reduction in at least one PR build to maintain the Ensemble reduction in future release. 

## Testing

These changes have been tested both with and without ensemble reduction.

### DIFF
--- a/cmake/std/PullRequestLinuxGCC4.8.4TestingSettings.cmake
+++ b/cmake/std/PullRequestLinuxGCC4.8.4TestingSettings.cmake
@@ -22,6 +22,9 @@ set(Tpetra_INST_SERIAL ON CACHE BOOL "Set by default for PR testing")
 # Disable just one Teko sub-unit test that fails with openmpi 1.10 (#2712)
 set (Teko_DISABLE_LSCSTABALIZED_TPETRA_ALPAH_INV_D ON CACHE BOOL "Temporarily disabled in PR testing")
 
+# Enable the ensemble reduction in this build to maintain Stokhos with ensemble reduction (#7067)
+set (Stokhos_ENABLE_Ensemble_Reduction ON CACHE BOOL "Enable reduction across ensemble components")
+
 # Disable three ShyLu_DD tests - see #2691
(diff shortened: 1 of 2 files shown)


cmake/std/PullRequestLinuxGCC4.8.4TestingSettings.cmake
```diff
@@ -22,6 +22,9 @@ set(Tpetra_INST_SERIAL ON CACHE BOOL "Set by default for PR testing")
 # Disable just one Teko sub-unit test that fails with openmpi 1.10 (#2712)
 set (Teko_DISABLE_LSCSTABALIZED_TPETRA_ALPAH_INV_D ON CACHE BOOL "Temporarily disabled in PR testing")
 
+# Enable the ensemble reduction in this build to maintain Stokhos with ensemble reduction (#7067)
+set (Stokhos_ENABLE_Ensemble_Reduction ON CACHE BOOL "Enable reduction across ensemble components")
```
etphipp: Do we want to put this change in a PR for one of the more recent GCC versions instead (e.g., 4.9, 7.2 or 8.3)?  I suspect 4.8.4 will go away in the not too distant future.  Comments @trilinos/framework?

jwillenbring: The default compiler for RHEL 7 is 4.8.5. That is the biggest factor in keeping the build for now. I cannot accurately predict how long the build will be around, but its removal is not currently imminent.

etphipp: OK, for now I think we will just go ahead and add the change to the 4.8.4 PR builds.

PR #7069, opened 2020-03-23
Fix CUDA_LAUNCH_BLOCKING for Tpetra parallel tests on white

MicheldeMessieres:
@mhoemmen @kddevin 

These changes are sufficient to get the Tpetra test suite passing on white with CUDA_LAUNCH_BLOCKING unset. (except for the trivial case of initialization tests which currently fail because they check CUDA_LAUNCH_BLOCKING).

White was down for a while so I started checking on a local machine running a GTX960 card. That revealed a few new issues which I will open as a separate PR to keep things organized.

For TpetraExt_TripleMatrixMultiply_def.hpp I added 4 fences, but also removed 1. I don't think we need the removed fence. If we do need it, then we may need it for all 4 spots in that file.

For KokkosBlas::dot and KokkosBlas::gemm fences, we likely need more fences for other KokkosBlas calls. This was discussed in [https://github.com/kokkos/kokkos-kernels/pull/660](url). Until the plan is resolved, I only changed those spots I specifically encountered as failures on white. 

@trilinos/tpetra 

## Motivation

Resolve missing fences when CUDA_LAUNCH_BLOCKING is unset.

## Testing

These issues were discovered by running the Tpetra test suite on white with CUDA_LAUNCH_BLOCKING=0.

### DIFF
--- a/packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
@@ -492,7 +492,6 @@ namespace Tpetra {
       // Run through all the hash table lookups once and for all
       lo_view_t targetMapToOrigRow(Kokkos::ViewAllocateWithoutInitializing("targetMapToOrigRow"),Aview.colMap->getNodeNumElements());
       lo_view_t targetMapToImportRow(Kokkos::ViewAllocateWithoutInitializing("targetMapToImportRow"),Aview.colMap->getNodeNumElements());
-      Kokkos::fence();
       Kokkos::parallel_for("Tpetra::mult_R_A_P_newmatrix::construct_tables",range_type(Aview.colMap->getMinLocalIndex(), Aview.colMap->getMaxLocalIndex()+1),KOKKOS_LAMBDA(const LO i) {
           GO aidx = Acolmap_local.getGlobalElement(i);
           LO P_LID = Prowmap_local.getLocalElement(aidx);
@@ -505,6 +504,7 @@ namespace Tpetra {
             targetMapToImportRow(i) = I_LID;
           }
         });
+      Kokkos::fence();
 
       // Call the actual kernel.  We'll rely on partial template specialization to call the correct one ---
       // Either the straight-up Tpetra code (SerialNode) or the KokkosKernels one (other NGP node types)
@@ -606,6 +606,7 @@ namespace Tpetra {
 
           }
         });
+      Kokkos::fence();
 
       // Call the actual kernel.  We'll rely on partial template specialization to call the correct one ---
       // Either the straight-up Tpetra code (SerialNode) or the KokkosKernels one (other NGP node types)
@@ -771,6 +772,7 @@ namespace Tpetra {
             targetMapToImportRow(i) = I_LID;
           }
         });
+      Kokkos::fence();
 
       // Call the actual kernel.  We'll rely on partial template specialization to call the correct one ---
       // Either the straight-up Tpetra code (SerialNode) or the KokkosKernels one (other NGP node types)
@@ -870,6 +872,7 @@ namespace Tpetra {
 
           }
         });
+      Kokkos::fence();
 
       // Call the actual kernel.  We'll rely on partial template specialization to call the correct one ---
       // Either the straight-up Tpetra code (SerialNode) or the KokkosKernels one (other NGP node types)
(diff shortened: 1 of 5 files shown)


packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
```diff
@@ -492,7 +492,6 @@ namespace Tpetra {
       // Run through all the hash table lookups once and for all
       lo_view_t targetMapToOrigRow(Kokkos::ViewAllocateWithoutInitializing("targetMapToOrigRow"),Aview.colMap->getNodeNumElements());
       lo_view_t targetMapToImportRow(Kokkos::ViewAllocateWithoutInitializing("targetMapToImportRow"),Aview.colMap->getNodeNumElements());
-      Kokkos::fence();
       Kokkos::parallel_for("Tpetra::mult_R_A_P_newmatrix::construct_tables",range_type(Aview.colMap->getMinLocalIndex(), Aview.colMap->getMaxLocalIndex()+1),KOKKOS_LAMBDA(const LO i) {
```
mhoemmen: @csiefer2 @brian-kelley  Fun fact: these loop bounds work even if the calling process has no indices, but only because `getMaxLocalIndex()` returns `OrdinalTraits<LO>::invalid()` which happens to be `-1` for signed or `numeric_limits<LO>::max()` for unsigned.

packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
```diff
@@ -505,6 +504,7 @@ namespace Tpetra {
             targetMapToImportRow(i) = I_LID;
           }
         });
+      Kokkos::fence();
```
mhoemmen: I would have preferred putting the fences as close to the incorrect code as possible, with a giant comment saying "FIXME the code below is broken."  However, it's already great just to have these fences, and the Tpetra team can always grep for them and try to purge them later.

packages/tpetra/core/src/Tpetra_Map_def.hpp
```diff
@@ -1645,6 +1645,7 @@ namespace Tpetra {
         Kokkos::RangePolicy<LO, typename DT::execution_space>
           range (static_cast<LO> (0), static_cast<LO> (lgMap.size ()));
         Kokkos::parallel_for (range, *this);
+        Kokkos::fence();
```
mhoemmen: Again, it would be nicer to put the fence as close to the erroneous code as possible.  The code here is fine; the code that follows (probably that uses the local-to-global Map on host) is wrong.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3921,6 +3921,7 @@ namespace Tpetra {
       ProfilingRegion regionGemm ("Tpetra::MV::multiply-call-gemm");
       KokkosBlas::gemm (&ctransA, &ctransB, alpha_IST, A_sub, B_sub,
                         beta_local, C_sub);
+      Kokkos::fence();
```
mhoemmen: What's actually broken after this?  Is it the code immediately below, like the `::Tpetra::deep_copy` or `reduce`, or is it code not in this function?  My guess is that it's the `reduce`, in which case the `fence` should go right above the `reduce` call (that would be nicer, since it would skip the fence in some cases).  If it's Belos itself, then we should fence in the `Belos::MultiVecTraits` specialization.

MicheldeMessieres:
@mhoemmen I'll follow up and look at the issues you mentioned. I was in the habit of putting fences right after the parallel_for that seeds the issue but I understand you are saying it would be best to push the fence as late as possible to guarantee it's used minimally.

mhoemmen:
@MicheldeMessieres wrote:
> ... but I understand you are saying it would be best to push the fence as late as possible to guarantee it's used minimally.

It's more about identifying code that requires a fence.  If the fences happen early, then it will take more work in the future to identify and fix whatever code needs the fence.  Furthermore, if you put the fences later, then it might be easier to find future regressions (changes to code that break with `CUDA_LAUNCH_BLOCKING` unset).  On the other hand, if you've already done as much work as you have time to do, then by all means place those fences where you know best : - ) .

MicheldeMessieres:
Thanks - I wasn't really appreciating that none of these fences are expected to be permanent. I think then that sync to the required space should eventually be executing all the required fences so Tpetra would not usually do it explicitly.

mhoemmen:
@MicheldeMessieres wrote:
> I wasn't really appreciating that none of these fences are expected to be permanent.

There might be some that need to be permanent -- e.g., if an MPI implementation can handle CUDA buffers but is not aware of GPU streams, then Tpetra code calling MPI sends needs to know that the device pack kernel is done before issuing a send.

The idea is first to make Trilinos work with `CUDA_LAUNCH_BLOCKING` unset, and then to remove unnecessary fences.

Issue #7072, opened 2020-03-24
Teuchos: Encapsulate Kokkos profiling region + Teuchos::TimeMonitor


mhoemmen:
## Enhancement
@trilinos/teuchos @rppawlo 

### Summary

We want an interface that abstracts over Kokkos profiling regions, `Teuchos::TimeMonitor`, stack timers, and all sorts of timing things.  We want this for the following reasons:

  1. (Seriously, do I need to explain why encapsulating N different timer libraries in one interface is a good idea?)
  2. Kokkos profiling regions currently incur overhead due to extra fences, so Trilinos wants to be able to turn off using them even if the user has loaded a profiling library at run time.
  3. We've seen stack timers incur overhead for large numbers of timers and/or deep stacks, so users might want to turn them off, perhaps even at run time.

### Desired features

  - The interface should live in Teuchos, so that downstream solver packages like MueLu that do not require Kokkos can use it, and so that Kokkos is not forced to be MPI-aware.
  - It should be able to distinguish recursive calls (e.g., for different levels of a multigrid hierarchy in MueLu), and in general should have features like those of `Teuchos::StackedTimer`.

### What's annoying about current interfaces

  - There are too many of them (`Teuchos::Time`, `Teuchos::TimeMonitor`, `Teuchos::StackedTimer`, Kokkos profiling regions, whatever MueLu's thing is, ...).
  - They all look and behave differently.
  - Developers don't use all of them consistently, so users see different timers, depending on which interface(s) they enable.
  - Some interfaces (e.g., `Teuchos::Time`) have an option to use other interfaces (e.g., Kokkos profiling regions), so unwitting developers may be "double-timing" some regions.
  - Developers have a habit of selectively enabling some timers at compile time by sprinkling macros like `HAVE_TPETRA_MMM_TIMINGS` all over the code.  Whatever is inside those macros doesn't get tested and might not even build.  Users don't have a way to enable those timers without manually changing the code and rebuilding.
  - Developers have a habit of doing ugly things like managing scope guards by pointer instead of by value.  This leads to issues like https://github.com/trilinos/Trilinos/issues/2804 .

### What we have now

[...]

PR #7073, opened 2020-03-24
Tpetra: Adding environment variable to allow Teuchos::Timers in Tpetra::ProfilingRegion

csiefer2:
Does what you expect as applied to the MueLu Driver; See #7072
Tests: RHEL6


mhoemmen:
btw, if you turn on both options, this might actually double-count some regions, since `Teuchos::Time` uses Kokkos' profiling regions by default.

PR #7074, opened 2020-03-24
Tpetra CUDA_LAUNCH_BLOCKING unset fixes

MicheldeMessieres:
These are additional fixes identified on a machine with an Nvidia GTX960 card. These errors did not exhibit on white. So this PR extends #7069.

@trilinos/tpetra

## Motivation

Make Tpetra run with CUDA_LAUNCH_BLOCKING unset.

## Related Issues
#7069 

## Testing

Tpetra tests - these CUDA_LAUNCH_BLOCKING issues were tracked down on a system with a GTX960 card.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
+++ b/packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
@@ -323,15 +323,6 @@ namespace Tpetra {
              range_type range (execSpace, 0, lclNumRows);
 
              Kokkos::parallel_for (kernelLabel, range, g);
-
-             // This fence is currently needed for the following specific situation:
-             // For transform Y to X:
-             //  Y.putScalar()    // acts on device
-             //  Y.sync_host()    // now need_sync_host() and need_sync_device() are false
-             //  transform (on device)
-             //  Y.sync_host()    // no modifications so no fence - this usually will be a fence
-             //  read Y           // crashes
-             Kokkos::fence();
            },
            readOnly (input).on (memSpace).at (execSpace),
            writeOnly (output).on (memSpace).at (execSpace));
@@ -367,7 +358,6 @@ namespace Tpetra {
             range_type range (execSpace, 0, lclNumRows);
 
             Kokkos::parallel_for (kernelLabel, range, g);
-            Kokkos::fence(); // see note in above transform_vec_notSameObject
           },
           readOnly (input).on (memSpace).at (execSpace),
           writeOnly (output).on (memSpace).at (execSpace));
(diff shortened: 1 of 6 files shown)


MicheldeMessieres:
@mhoemmen Realize you are probably slammed, but if you can comment on the fence in Tpetra_transform_MultiVector.hpp that would be helpful. I'm not sure we expect/want a fence here. The code comment explains the specific sequence where the problem occurs.

packages/tpetra/core/src/Tpetra_transform_MultiVector.hpp
```diff
@@ -323,6 +323,15 @@ namespace Tpetra {
              range_type range (execSpace, 0, lclNumRows);
 
              Kokkos::parallel_for (kernelLabel, range, g);
+
+             // This fence is currently needed for the following specific situation:
+             // For transform Y to X:
+             //  Y.putScalar()    // acts on device
+             //  Y.sync_host()    // now need_sync_host() and need_sync_device() are false
```
mhoemmen: Wait, if `putScalar` operates on device data, then `need_sync_host()` should be `true` after calling `putScalar`.  If not, then that's a bug in `putScalar`.

MicheldeMessieres: @mhoemmen It is need_sync_host() true after the putScalar. But then the user calls sync_host() before calling transform, at which point I verified need_sync_host() and need_sync_device() are now both false, which I think is correct. But then it seems that in this case there is no way for sync state information to provide the fence since Y never actually changes and is already synced to device.

Probably user would normally not do this before the transform and would normally not care to read Y anyways since it's not modified. But not sure how to make sure that one case is logically covered without adding some additional internal logic.

mhoemmen: @MicheldeMessieres wrote:
> But not sure how to make sure that one case is logically covered without adding some additional internal logic.

It seems like the problem is that `Tpetra::MultiVector::sync_host` doesn't fence.  (This is because it does nothing but call `Kokkos::DualView::sync_host`, and that doesn't fence.)  Why not just have `Tpetra::MultiVector::sync_host` do a fence?

It seems like the goal should be for developers to avoid calling `sync_host`, since that implies they have code that needs to run on host.  Thus, I'm OK with making `sync_host` always fence.  @trilinos/tpetra, what do you all think?

MicheldeMessieres:
@kddevin @csiefer2 Added a fence to sync_host as suggested.

Note I found an additional failure on the GTX960 machine for parallel tests TpetraCore_SubmapExportTests_MPI_2 and TpetraCore_SubmapImportTests_MPI_2. I'll open solutions for that as a separate PR.

PR #7077, opened 2020-03-26
MueLu UncoupledAggregationFactory_kokkos: Accept more params

cgcgcg:
@trilinos/muelu 

## Motivation
Make sure that "aggregation: phase3 avoid singletons" and"aggregation: error on nodes with no on-rank neighbors" get set and passed to phase 3.


cgcgcg:
@brian-kelley Thanks for approving!

PR #7084, opened 2020-03-27
Teuchos Stacked timer histogram fix

cgcgcg:
@trilinos/teuchos 

## Motivation
Fix the stacked timer histogram in the case when timers are not used on all ranks.


cgcgcg:
@jhux2 I believe it may have never worked correctly.

Issue #7046, opened 2020-03-18
Set MueLu_ENABLE_Kokkos_Refactor_Use_By_Default=ON in at least on Trilinos PR build


bartlettroscoe:
The ATDM Trilinos configuration sets `MueLu_ENABLE_Kokkos_Refactor_Use_By_Default=ON` while none of the Trilinos PR builds do.  That allows failures like occurred in #7041 to slip into the ATDM Trilinos testing.

It would be nice if at least on of the Trilinos PR builds set  `MueLu_ENABLE_Kokkos_Refactor_Use_By_Default=ON`.  Alternatively, it might be nice to take one of the Trilinos PR builds and replace it with the ATDM Trilinos configuration to ensure that that configuration is always tested in PR testing.  (But that has the downside of every change to the cmake/std/atdm/ directory triggering a full build and testing of all packages in Trilinos.  Therefore, I am not sure I would vote for that but perhaps it is worth it.)

bartlettroscoe:
CC: @cgcgcg 

FYI: Another bunch of failures cased by not having  `MueLu_ENABLE_Kokkos_Refactor_Use_By_Default=ON` set in at least one PR build (see #7088).

@trilinos/framework, can we pick on of the PR builds to turn on `MueLu_ENABLE_Kokkos_Refactor_Use_By_Default=ON`?  Can someone (e.g. me) just pick one of the PR builds and set this?  For example, can I just turn this on in the CUDA PR build?

bartlettroscoe:
@trilinos/muelu, as this option is being used by all of the ATDM APPs, is there a reason why this can't be made the default for MueLu?

jwillenbring:
@bartlettroscoe 

> @trilinos/framework, can we pick on of the PR builds to turn on MueLu_ENABLE_Kokkos_Refactor_Use_By_Default=ON? Can someone (e.g. me) just pick one of the PR builds and set this? For example, can I just turn this on in the CUDA PR build?

At first this sounded like something experimental to me, but upon closer inspection, it looks like this is what ATDM relies on, and possibly also pointing to functionality that Kokkos currently supports. If that is the case, if the CUDA build is the logical one to modify, go ahead and ask for a Framework team member review (me and/or others) and we can do this.

cgcgcg:
> @trilinos/muelu, as this option is being used by all of the ATDM APPs, is there a reason why this can't be made the default for MueLu?

@bartlettroscoe FYI #7139

[...]

Issue #7090, opened 2020-03-28
Test MueLu_ParameterListInterpreterTpetra_MPI_1 regularly randomly timing out in 'waterman' CUDA opt builds starting 2020-01-29


bartlettroscoe:
CC: @trilinos/muelu, @srajama1 (Trilinos Linear Solvers Product Lead)

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label "impacting: tests"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

Having this test run on its own GPU does not help (see PR #6840).  This test is disabled in all ATDM Trilinos 'waterman' CUDA builds in PR #7407 and has been merged to 'atdm-nightly' in commit 4804b08.  Next: Waiting for confirmation on CDash that this test is disabled in ATDM Trilinos CUDA builds starting testing day 2020-05-21 ...

## Description

As shown in [this query](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2019-10-01&end=2020-03-28&filtercount=5&showfilters=1&filtercombine=and&field1=groupname&compare1=62&value1=Experimental&field2=buildname&compare2=65&value2=Trilinos-atdm-waterman&field3=testname&compare3=65&value3=MueLu_ParameterListInterpreterTpetra_MPI_&field4=status&compare4=61&value4=failed&field5=details&compare5=63&value5=Timeout) the test:

* `MueLu_ParameterListInterpreterTpetra_MPI_1`

in the builds:

* `Trilinos-atdm-waterman-cuda-9.2-opt`
* `Trilinos-atdm-waterman_cuda-9.2_fpic_static_opt`
* `Trilinos-atdm-waterman_cuda-9.2_shared_opt`

are fairly regularly randomly timing out starting on testing day <YYYY-MM-DD>.

The last 30 days of history are summarized in this table:

[...]

PR #7091, opened 2020-03-28
Set MueLu_ENABLE_Kokkos_Refactor_Use_By_Default=ON (#7046)

bartlettroscoe:
Not having this set in at least one PR build has allowed a mass number of test
failures to slip through into ATDM Trilinos testing.  See #7046.


bartlettroscoe:
I suspect this will fail in PR testing due to #7090 (which is the whole point).

At least one PR build should protect ATDM and the CUDA build seems like a good candidate.

@cgcgcg, can you please approve?

cgcgcg:
@bartlettroscoe Will do, but let's make sure it fails first.

PR #7097, opened 2020-03-30
Automatic snapshot commit from seacas at 500c3cef0

gsjaardema:

At commit:

commit 500c3cef066a7d7c498afe6168e3ee70da3331f3
Date:    Mon Mar 30 11:46:32 2020 -0600
Summary: IOSS: cgns - fix flush logic

* CGNS: Better support for flexible naming of BC nodes
* CGNS: Fix logic used for flushing of CGNS files
* CGNS: Better message when proc_count exceeds cell count
* CGNS: Remove some explicit template specification

* C++ -- Remove TOPTR() macro which was used when both pre and post
  C++11 compilers; use .data() now that C++11 is minimum version
  supported
* NEM_SPREAD: Check for valid element global ids
* EXODUS:
   * use calloc instead of malloc so memory zeroed
   * remove extraneous file

@trilinos/seacas


packages/seacas/libraries/exodus/test/testwt-results.c
```diff
@@ -93,8 +93,8 @@ int main(int argc, char **argv)
   num_node_sets = 0;
   num_side_sets = 0;
 
-  error = ex_put_init(exoid, title, num_dim, num_nodes, num_elem, num_elem_blk, num_node_sets,
```
bartlettroscoe: No longer checking error code?  Never checked the error code?

gsjaardema: It was assigning, but never checking the error code before. Removed the assignment to avoid "unused variable warning"

The check in this case is later when it verifies that correct values exist in the file...

PR #7103, opened 2020-04-01
MueLu: new experimental smooth randvec coalescedrop factory

rstumin:
Tests: Linux


packages/muelu/src/Graph/MatrixTransformation/MueLu_SmooVecCoalesceDropFactory_def.hpp
```diff
@@ -0,0 +1,556 @@
...
+#include <Xpetra_IO.hpp>
+
+#include <algorithm>
+#include <cstdlib>
+#include <string>
+
+// If defined, read environment variables.
+// Should be removed once we are confident that this works.
+// #define DJS_READ_ENV_VARIABLES
+
```
csiefer2, line 90: Don't think you need this

PR #7116, opened 2020-04-02
Intrepid2: add support for hierarchical H(grad) bases on triangle and tetrahedron.

CamelliaDPG:
Continuing the work implementing hierarchical bases, this PR adds support for H(grad) bases on triangles and tetrahedra.  (These are the bases that EMPIRE expects to use for limiters soon.)

There are two types of tests included here:
1. Basis equivalence tests: these confirm that both the DG and CG variants of the bases are consistent with the existing nodal bases in Intrepid2.
2. Sub-basis inclusion tests: these confirm that the hierarchical bases do in fact nest.

In addition to those tests, an offline verification has been performed to confirm that this implementation agrees with the equivalent bases in ESEAS.


CamelliaDPG:
Closed because I meant to do PR relative to develop, not master.

PR #7128, opened 2020-04-04
WIP: Stokhos: add an optimized GEMV for the ensemble GMRES without ensemble reduction

kliegeois:
@trilinos/stokhos 

## Motivation

This PR adds a GEMV for the ensemble types in Stokhos.
This implementation has been optimized with the aim of reaching maximal memory bandwidth on CPU during the Arnoldi process occurring in the ensemble GMRES without the ensemble reduction.

The implementation relies on a tiling strategy to reuse data from cache and assume that the matrix is stored with a left layout (the implementation works with a right layout but is suboptimal in that case).

## Testing

This PR branch has been tested on the Blake system without the ensemble reduction.

### DIFF
--- a/packages/stokhos/CMakeLists.txt
+++ b/packages/stokhos/CMakeLists.txt
@@ -30,6 +30,23 @@ TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_Ensemble_Reduction
   "Enable reduction across across ensemble components."
   OFF )
 
+IF(NOT HAVE_STOKHOS_ENSEMBLE_REDUCT AND NOT Stokhos_ENABLE_CUDA)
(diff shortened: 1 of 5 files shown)


etphipp:
FYI, the autotester is down, and there is not ETA on when it will come back.  Otherwise this looks good.  I'll approve it when the autotester comes back up and the code compiles clean.

kliegeois:
Thanks @etphipp for your comment!

kliegeois:
@etphipp one test fails and one file does not compile.
I am looking into that.

kliegeois:
OK this is very strange.
The tests are currently failing because the norm explicit residual is NaN.
However, the implicit residual does converge.
Therefore, it seems that the Hessenberg matrix is correct.

What is very strange is that I am unable to reproduce the issue on Blake or any other system right now. On Blake the tests are passing.

I will continue to try to reproduce this issue to debug the problem.

In the meantime, this PR can be labeled as WIP.

PR #7131, opened 2020-04-06
MueLu: modifying region coarse grid setup

### LINKED ISSUE #7129: MueLu: improving region coarse grid smoother
## Enhancement
@trilinos/muelu 

The coarse composite operator does not need to be computed when a coarse grid smoother is used instead of a coarse grid direct solve. To improve performance we need to modify the logic around the coarse grid solver setup. Also adding an interface to use a different smoother on the coarse grid would allow to perform a stronger coarse grid solve compared to regular level smoothers.

lucbv:
@trilinos/muelu 

## Motivation
1. no longer computing the coarse composite operator when a coarse grid smoother is used and not throwing on "coarse solver type = smoother".
2. adding a command line parameter to pass specific option to the coarse grid smoother since it might require more iterations than a regular level smoother.

## Related Issues

* Closes #7129
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing
Added new input deck for coarse grid smoother and tested on local machine.

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
@@ -731,6 +731,8 @@ void createRegionHierarchy(const int maxRegPerProc,
     levelList.set<Teuchos::Array<RCP<Vector> > >("solution", regSol, "Cached solution vector");
   }
 
+  // std::cout << mapComp->getComm()->getRank() << " | MakeCoarseLevelMaps ..." << std::endl;
+
   tmLocal = Teuchos::null;
   tmLocal = rcp(new TimeMonitor(*TimeMonitor::getNewTimer("createRegionHierarchy: MakeCoarseLevel")));
   MakeCoarseLevelMaps(maxRegPerGID,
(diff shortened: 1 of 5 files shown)


packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
```diff
@@ -795,17 +757,71 @@ void createRegionHierarchy(const int maxRegPerProc,
...
-  if (coarseSolverType == "smoother") noCoarseSmoothing = 0;
+  // std::cout << mapComp->getComm()->getRank() << " | Setup smoothers ..." << std::endl;
 
   tmLocal = Teuchos::null;
   tmLocal = rcp(new TimeMonitor(*TimeMonitor::getNewTimer("createRegionHierarchy: SmootherSetup")));
-  for(int levelIdx = 0; levelIdx < numLevels-noCoarseSmoothing; ++levelIdx) {
+  // Only set the smoother up to the last but one level
+  // if we want to use a smoother on the coarse level
+  // we will handle that separately with "coarse solver type"
+  for(int levelIdx = 0; levelIdx < numLevels - 1; ++levelIdx) {
```
mayrmt, line 767: @lucbv Shouldn't the index for "all levels except the last one" have an upper bound
```suggestion
  for(int levelIdx = 0; levelIdx < numLevels - 2; ++levelIdx) {
```
?

lucbv: That's a fair question, I will double check that and get back to you.
At the moment that would mean doing an extra setup on the coarsest level so probably not too bad but since it's easy we should fix it if it's wrong!

lucbv:
@mayrmt the loops is correct, it will stop at `levelIdx = numLevels - 2` as it is the last integer that verifies `levelIdx < numLevels`. I did push an extra commit as I had forgotten to checkin the new xml needed for coarse grid smoother example.

packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
mayrmt: You're right. Stupid me!

mayrmt:
Autotester seems to be running again, so let's give it another try. Setting the "AT: RETEST" label.

PR #7134, opened 2020-04-06
Piro: TempusSolver refactor and cleanup

ikalash:
Some improvements to Piro package: 

- Piro::TempsSolver has been refactored into helper class called Piro::TransientSolver.  Transient sensitivities will go in the latter class, similar to Piro::NOXSolver and Piro::SteadyStateSolver.  
- Obsolete template arguments that used to be needed for Albany have been removed from various Piro classes.


ikalash:
@mperego : could you please re-approve?  I have to make a slight change to fix some -Werror=reorder issues that showed up in the auto-tests.

PR #7146, opened 2020-04-09
DO NOT MERGE Tpetra: Communication/Computation overlap

cgcgcg:
@trilinos/tpetra 

## Motivation
This is a hacked together attempt of splitting Tpetra's communication routines to allow for communication/computation overlap. THIS IS NOWHERE READY TO BE MERGED.

In `DistObject::doTransferNew` I split the `doPostAndWaits` call into `doPosts` and `doWaits`. After `doPosts`, I return a so-called `Tpetra::TransferRequest` object to the user. The `process` method of this object kicks off `doWaits` and `unpackAndCombine`.

The functionality of `doTransfer` (and hence `doImport` and `doExport`) has not been changed, it simply calls `process` on the `TransferRequest` right away. I added `startImport` and `startExport` methods that return the `TransferRequest` object instead to the user. I made the necessary changes to use it in `MueLu::RefMaxwell`.


jhux2:
I didn't know about draft PRs, nice!

PR #7148, opened 2020-04-09
MueLu: make Galeri a required test dependency

jhux2:
Otherwise the set of built tests is pretty much useless.

## Stakeholder Feedback
Useful primarily for MueLu developers.

## Testing
Does not affect any source code.

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/7148)


cgcgcg:
@jhux2 I don't understand why, but it seems that this did not do what we thought it would:
https://testing.sandia.gov/cdash/index.php?subproject=MueLu&project=Trilinos
Seems that MueLu now thinks that it doesn't have Galeri, and hence tons of tests don't get run.

cgcgcg:
Oh wait, required dependency means we shouldn't check for Galeri anymore cause it's always enabled.

PR #7151, opened 2020-04-10
ATDM: Add new 'van-tx2' env to support ASTRA systems

bartlettroscoe:
* Added new 'van1-tx2' env to support ASTRA systems for SPARC, EMPIRE and other ATDM APPs
* Other small tweaks

## How was this tested?

A lot of it.  See subtasks of [CDOFA-27](https://sems-atlassian-srn.sandia.gov/browse/CDOFA-27) and especially [CDOFA-1113](https://sems-atlassian-srn.sandia.gov/browse/CDOFA-113).


cmake/ctest/drivers/atdm/van1-tx2/local-driver.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash -l
```
bartlettroscoe, line 1: @e10harvey, I just realized this file is not getting used.  We should likely remove it to avoid confusion.

e10harvey: I don't think this file should be removed. local-driver.sh is used in two places: https://github.com/trilinos/Trilinos/pull/7151/files#diff-81f96a09d6a0988a70a0432048a23567R9 and https://github.com/trilinos/Trilinos/pull/7151/files#diff-602f89090dbaf9b43a16cf4d01a61b35R9.

cmake/std/atdm/van1-tx2/environment.sh
```diff
@@ -0,0 +1,153 @@
...
+else
+  echo
+  echo "***"
+  echo "*** ERROR: COMPILER=$ATDM_CONFIG_COMPILER is not supported!"
+  echo "***"
+  return
+fi
+
+# Common modules for all builds
+module load ninja
```
e10harvey, line 72: @bartlettroscoe: Should this be `module load ninja/1.8.2` to guard against something changing under the hood going forward?

cmake/std/atdm/utils/get_known_system_info.sh
```diff
@@ -31,6 +31,9 @@ fi
 source ${ATDM_CONFIG_SCRIPT_DIR}/utils/get_system_info_utils.sh
 
 realHostname=`hostname`
+if [[ "${ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE}" ]] ; then
```
e10harvey, line 34: Please remove lines 34-36 or add a warning under the branch here, such as: `echo: "WARNING: realHostname=$realHostname overriden to value of ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE which is $ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE in $PWD/get_known_system_info.sh"`.

cmake/std/atdm/load-env.sh
```diff
@@ -56,7 +56,7 @@ fi
 source ${ATDM_CONFIG_SCRIPT_DIR}/utils/get_system_info.sh
 
 if [[ $ATDM_CONFIG_SYSTEM_NAME == "" ]] ; then
-  echo "Error, could not determine a system configuration, aborting env loading script!"
+  echo "Error, could not determine a system configuration for hostname='$realHostname', aborting env loading script!"
```
e10harvey, line 59: :+1: Nice! This is very helpful!

cmake/std/atdm/README.md
```diff
@@ -1162,6 +1163,89 @@ $ ./checkin-test-atdm.sh <buildname0> <buildname1> ... \
...
+  -DTrilinos_ENABLE_Tpetra=ON \
+  $TRILINOS_DIR
+
+$ srun -N 1 make NP=20  # We have the entire compute node to ourselves!
+
+$ ctest -j4
+```
+
+The advantage of the latter approach is that one just waits once for a node
+allocationand then one can immediately run fast parallel builds on the compute
```
e10harvey, line 1217: :+1: This documentation looks great! Small typo here: "allocationand" -> "allocation and"

cmake/ctest/drivers/atdm/van1-tx2/local-driver.sh
bartlettroscoe: @e10harvey, no, you are correct.  I added this file :-)  I was thinking I was looking in the directory cmake/std/atdm/van1-tx2/.

cmake/ctest/drivers/atdm/van1-tx2/local-driver-on-allocation.sh
```diff
@@ -0,0 +1,51 @@
...
+echo ""
+echo "  Running ${JOB_NAME}"
+echo "  in salloc allocation from node '$(hostname)'"
+echo ""
+echo "======================================================================"
+echo
+
+# Must use same site name for build and test results so they match up on
+# CDash!
+export CTEST_SITE=${ATDM_CONFIG_CDASH_HOSTNAME}
```
e10harvey, line 21: :+1:

cmake/ctest/drivers/atdm/van1-tx2/local-driver.sh
```diff
@@ -0,0 +1,32 @@
...
+source $WORKSPACE/Trilinos/cmake/std/atdm/load-env.sh $JOB_NAME
+echo
+
+if [[ "${ATDM_CONFIG_WCID_ACCOUNT}" == "" ]] ; then
+  export ATDM_CONFIG_WCID_ACCOUNT=${ATDM_CONFIG_WCID_ACCOUNT_DEFAULT}
+fi
+
+set -x
+
+salloc -N 1 --time=${SALLOC_CTEST_LIMIT_MINUTES} -p short,batch \
```
e10harvey, line 21: :+1:

bartlettroscoe:
@e10harvey, scratch my stupid comment above.  I think this is ready to merge.

(FYI: I did a diff of the rebased and cleaned up branch and it was identical to the branch before this so it showed this was done perfectly.)

cmake/std/atdm/utils/get_known_system_info.sh
bartlettroscoe: @e10harvey, that is just to drive testing (initially manual testing but in the future, automated unit testing).  To make that clear, we could change the name to `ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE_FOR_UNIT_TESTING`?   And we should add a comment to describes the purpose of this variable as well.  Is that enough?

We can also add the echo "Warning" output too if you want.  But if someone sets a variable called `ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE_FOR_UNIT_TESTING` and is surprised about the behavior of this, I don't feel badly for them.

cmake/std/atdm/van1-tx2/environment.sh
bartlettroscoe: That is worth considering.  On the other hand, if we get an updated ninja automatically, I would not complain either.  I think it is important that we try to be testing the default envs on these machines since I suspect that is what users depend on.  Really we need to be using what the ATDM APPs are using so really this should be their choice.   How about we ask them?  Do they want us to be testing with the most recent Fortran-capable ninja on that system or do they want us to be testing against a fixed version of ninja?

cmake/std/atdm/utils/get_known_system_info.sh
e10harvey: Both the current variable name and ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE_FOR_UNIT_TESTING are OK. Documenting this variable would be excellent. Yes, please either add the warning or remove those lines.

bartlettroscoe: I resolved this in the commit 02303b7f0c.  I tested this on 'stria' with:

```
Hostname 'stria-login1' matches known ATDM host 'stria-login1' and system 'van1-tx2'

$ export ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE_FOR_UNIT_TESTING=stria-something

***
*** WARNING: realHostname=stria-login1 overriden to value of
*** ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE_FOR_UNIT_TESTING='stria-something'
*** in <trilinos-dir>/cmake/std/atdm/utils/get_known_system_info.sh.
*** This variable should only be set for unit testing purposes!
***

Hostname 'stria-something' matches known ATDM host 'stria-something' and system 'van1-tx2'

$ unset ATDM_CONFIG_GET_KNOW_SYSTEM_INFO_REAL_HOSTNAME_OVERRIDE_FOR_UNIT_TESTING

Hostname 'stria-login1' matches known ATDM host 'stria-login1' and system 'van1-tx2'
```

That looks correct to me.

NOTE: You can't use `$PWD/get_known_system_info.sh` because the current directory is not `<trilinos-dir>/cmake/std/atdm/utils` when this script is run.  I also decided not to put in the full path since it makes it really long.  People can figure out how to find that if they need to.

PR #7156, opened 2020-04-11
Tpetra Distributor: add method to get reverse distributor

cgcgcg:
@trilinos/tpetra 

## Motivation
Allows to set parameters on reverse distributor

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Distributor.cpp
+++ b/packages/tpetra/core/src/Tpetra_Distributor.cpp
@@ -385,6 +385,12 @@ namespace Tpetra {
     }
   }
 
+  Teuchos::RCP<Distributor>
+  Distributor::
+  getReverseDistributor() {
+    return reverseDistributor_;
+  }
+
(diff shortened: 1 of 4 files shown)


lucbv:
I think I would like it better if you introduce a method like `setReverseDistributorParams` so as not to expose the object itself directly.

jhux2:
Is this related to PR #7146?

cgcgcg:
@jhux2 No, but working on #7146 I noticed that no parameters were set on the reverse distributor used for the doExport using an importer in RefMaxwell. This change seems desirable independent of #7146.

kddevin:
@cgcgcg Would you please add a comment on how you tested this change?  Did you add a Tpetra test that exercises the method?  If not, it will not get exercised in PR testing.
We need this information for ASC's software quality assessments; we are trying to not approve PRs until test information is added.  Thanks!

cgcgcg:
@kddevin Added a testing section to the description, will add a unit test in Tpetra for it.

kddevin:
Thanks, @cgcgcg!

PR #7160, opened 2020-04-12
Amesos2: Refactor superlu to use Kokkos views

MicheldeMessieres:
@trilinos/amesos2

## Motivation

Solvers are being refactored so they use Kokkos views and can handle inputs from different memory spaces.

## Testing

Amesos2_SuperLU_Solver_Test


packages/amesos2/src/Amesos2_Superlu_FunctionMap.hpp
```diff
@@ -79,6 +79,19 @@ namespace SLU {
 #include "slu_util.h"
 #include "superlu_enum_consts.h"
 
+  #ifdef HAVE_TEUCHOS_COMPLEX
```
srajama1, line 82: This is strange, I don't know why these were working before need to be added here. Can you please explain ?

packages/amesos2/src/Amesos2_Superlu_FunctionMap.hpp
```diff
@@ -724,11 +737,8 @@ namespace Amesos2 {
 
 #ifdef HAVE_TEUCHOS_COMPLEX
 
-  /* The specializations for Teuchos::as<> for SLU::complex and
-   * SLU::doublecomplex are provided in Amesos2_Superlu_Type.hpp
-   */
   template <>
-  struct FunctionMap<Superlu,SLU::C::complex>
+  struct FunctionMap<Superlu, Kokkos::complex<float>>
```
srajama1, line 741: Are you saying SuperLU can take Kokkos::complex directly, there was some problem with it long time ago, that is why we we needed its inbuilt complex. I don't remember what it is. If it works w/ kokkos::complex then that is fine to skip the old one.

packages/amesos2/src/Amesos2_Superlu_FunctionMap.hpp
```diff
@@ -805,10 +815,11 @@ namespace Amesos2 {
     }
 
     static void create_CompCol_Matrix(SLU::SuperMatrix* A, int m, int n, int nnz,
-				      SLU::C::complex* nzval, int* rowind, int* colptr,
+				      void * nzval, int* rowind, int* colptr,
```
srajama1, line 818: Did SuperLU interface change ? Then do we have support older SuperLU ?

packages/amesos2/src/Amesos2_Superlu_FunctionMap.hpp
```diff
@@ -821,10 +832,11 @@ namespace Amesos2 {
     }
 
     static void create_Dense_Matrix(SLU::SuperMatrix* X, int m, int n,
-				    SLU::C::complex* x, int ldx, SLU::Stype_t stype,
+				    void * x, int ldx, SLU::Stype_t stype,
```
srajama1, line 835: Same, did the interface change ?

packages/amesos2/src/Amesos2_Superlu_TypeMap.hpp
```diff
@@ -79,294 +74,9 @@ extern "C" {
 
 #undef __SUPERLU_SUPERMATRIX
 #include "supermatrix.h"	// for Dtype_t declaration
-
-#ifdef HAVE_TEUCHOS_COMPLEX
```
srajama1, line 83: The TypeMap is to go from one type to another type. FunctionMap to map functions. Looks like this was moved to FunctionMap.

packages/amesos2/src/Amesos2_TpetraMultiVecAdapter_def.hpp
```diff
@@ -564,7 +566,36 @@ namespace Amesos2 {
         mv_->doImport (source_mv, *importer_, Tpetra::REPLACE);
       }
       else {
-        TEUCHOS_TEST_FOR_EXCEPTION(true, std::runtime_error, "Kokkos adapter CONTIGUOUS_AND_ROOTED not implemented for put1dData_kokkos_view.");
+        multivec_t redist_mv (srcMap, num_vecs); // unused for ROOTED case
```
srajama1, line 569: @ndellingwood Can you check this part as you did this sometime ago, I think.

packages/amesos2/src/Amesos2_Superlu_FunctionMap.hpp
MicheldeMessieres: I moved the includes from Type_Map to FunctionMap since they were not needed in TypeMap anymore. However maybe the changes in TypeMap I did are not correct. We used to be able to build the solver templated over the native SLU type. But now that we are using views I didn't think we would ever want a view of that type. I'm not sure that can even work with Kokkos unless we added support in a lot of places. This is a backwards compatibility issue though. 

Would user ever want to build the solver directly on the native SLU type? They can't do that anymore.

Would we ever want to have a view of data which was the native SLU type?

If what I did for TypeMap is ok, then a lot of the other questions follow from that.

packages/amesos2/src/Amesos2_Superlu_FunctionMap.hpp
MicheldeMessieres: This seemed fine but I did't know about the prior issues. So maybe I am not seeing a problem that is going to come up for some build.

packages/amesos2/src/Amesos2_Superlu_FunctionMap.hpp
MicheldeMessieres: It did not change but I thought this might be the easiest place to store the reinterpret_cast step. I could restore the native type here and have the solver cast the ptr before calling the FunctionMap. Resolving this is probably related to the 1st comment and possible problems with how I did TypeMap changes.

packages/amesos2/src/Amesos2_Superlu_TypeMap.hpp
MicheldeMessieres: Same questions as the 1st comment - if the changes I did in TypeMap are bad then I'll be reverting some stuff. But if the new view solvers will ever store data of native types I will need to discuss more how we will do that.

packages/amesos2/src/Amesos2_Superlu_FunctionMap.hpp
MicheldeMessieres: Switched back to manual copy to an array of native SLU type, so no more use of reinterpret cast.

MicheldeMessieres:
This is rolled into #7302 so we decided to just do one PR and close this one.

Issue #7164, opened 2020-04-14
MueLu: How to pass primal/dual mapping to InterfaceMappingTransferFactory?


mayrmt:
## Question
@trilinos/muelu @kliegeois 

I'm trying to understand how I can pass the information on the mapping of dual nodes to primal nodes for the `InterfaceMappingTransferFactory` introduced in PR #6931 for our application. I'm looking at the example in `packages/muelu/test/meshtying/MeshTyingBlocked_SimpleSmoother.cpp`.

As in the example, I want to read the preconditioner configuration from an xml-file. In contrast to the example, I want to create the preonditioner via `MueLu::CreateXpetraPreconditioner()`.

The necessary information is set on the parameter list as shown here: https://github.com/trilinos/Trilinos/blob/29bf4e7151cf8f0810b173a3291097a3322a9745/packages/muelu/test/meshtying/MeshTyingBlocked_SimpleSmoother.cpp#L211-L215

As far as I understand, this hard-codes the requires the sublist being named `myInterfaceAggs2`. However, the name of this list is defined in the xml-file as a user-specified string, so there's no guarantee that the user uses that particular name in every xml-file.

@kliegeois How can I pass the primal/dual node mapping to MueLu in a way, where I do not rely on particular sublist names in the MueLu xml-file?

kliegeois:
Basically, what we need is the factory `InterfaceAggregationFactory` to have access to `DualNodeID2PrimalNodeID - level 0`.

In the example, as you said, I hard coded the `InterfaceAggregationFactory` to be named as `myInterfaceAggs2`.

I see two approaches not to hard code the name:
- Give the name of the `InterfaceAggregationFactory` using a variable, possibly defined at the beginning of the xml file.
- Scan the xml file, and keep the name of the (last) `InterfaceAggregationFactory`.

I think that the last approach would be better as it does not rely on the user to add other information (the name of the factory).
The only drawback would be that this prevents us from using 2 `InterfaceAggregationFactory` but I do not see any reason for doing that right now.

mayrmt:
@kliegeois I like the second option better as well. There might be another: maybe this data also can be passed in via the `"user data"` sublist, similar to the fine-level nullspace and coordinate information.

@lucbv What do you think? Is this the intended use case of the `"user data"` sublist?

lucbv:
@mayrmt 
[...]

PR #7165, opened 2020-04-14
MueLu: add documentation to InterfaceAggregationFactory

mayrmt:
@trilinos/muelu @kliegeois 

## Motivation

I'm working on using the `InterfaceAggregationFactory` in an application code. This PR is supposed to improve documentation of this (and associated) factories to make it easier for future users.

## Related Issues

* Follows PR #6931 

## Stakeholder Feedback

The `InterfaceAggregationFactory` has just been added recently. I'm not aware of its use in application codes so far other than by its author @kliegeois.

## Testing

This just adds documentation, so I rebuilt and double-checked Doxygen locally.

## Additional Information

@kliegeois Please have a look, if I described everything correctly. I'm happy to adapt it, if I got something wrong. Thanks in advance!


packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_decl.hpp
```diff
@@ -56,6 +56,51 @@ namespace MueLu
 /*!
     @class InterfaceAggregationFactory class.
     @brief Factory for building aggregates for Lagrange multipliers.
+
+    ## Context, assumptions, and use cases ##
+
+    This factory is intended to be used for saddle-point systems of surface-coupled problems,
+    where constraints are enforced with Lagrange multipliers.
+    In addition to the primal unknowns, Lagrange multipliers are considered as dual unknowns.
+    The presenence of Lagrange mulitpliers make this a prima/dual problem
```
kliegeois, line 65: > The presenence of Lagrange mulitpliers make this a prima/dual problem

Should be replaced by:
The presence of Lagrange multipliers make this a primal/dual problem.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_decl.hpp
```diff
@@ -56,6 +56,51 @@ namespace MueLu
...
+    A                       | Factory | null |   | * | * | Generating factory of the matrix A
+    Aggregates              | Factory | null |   | * | * | Generating factory of the aggregates (of type "Aggregates" produced, e.g., by the UncoupledAggregationFactory)
+    DualNodeID2PrimalNodeID | Factory | null |   | * | * | Generating factory of the fine dual-to-primal node mapping
+
+    The * in the @c master.xml column denotes that the parameter is defined in the @c master.xml file.<br>
+    The * in the @c validated column means that the parameter is declared in the list of valid input parameters (see InterfaceAggregationFactory::GetValidParameters).<br>
+    The * in the @c requested column states that the data is requested as input with all dependencies (see InterfaceAggregationFactory::DeclareInput).
+
+    The mapping of dual-to-primal node IDs, \c DualNodeID2PrimalNodeID, is of data type \c std::map<LocalOrdinal,LocalOrdinal>.
+    The 'key' refers to the global ID of the dual node, while the 'value' represents the global ID of its primal counterpart.
```
kliegeois, line 94: The IDs are actually stored as local ID and not global ID.
Therefore, the line:

> The 'key' refers to the global ID of the dual node, while the 'value' represents the global ID of its primal counterpart.

Should be replaced by:

The 'key' refers to the local ID of the dual node, while the 'value' represents the local ID of its primal counterpart.

We can argue on the use of a `std::map` and consider a `std::vector`. However, Moertel (the Trilinos package used to compute the Mortar matrices) relies on a map currently.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_def.hpp
```diff
@@ -71,6 +71,10 @@ RCP<const ParameterList> InterfaceAggregationFactory<Scalar, LocalOrdinal, Globa
     DualNodeID2PrimalNodeID represents the mapping between the Dual Node IDs to the Primal Node IDs.
     This map makes one assumption: the number of dof per dual node is a constant and has to be set in 
     the parameter list as "number of DOFs per dual node".
+
+    Layout of the std::map<key, value>:
+    - key: dual node ID
```
kliegeois, line 76: We can potentially stress one more time that it is a local ID.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_decl.hpp
mayrmt: I'm fine with `std::map` if this enhances compatibility wit `Moertel`.

packages/muelu/src/Misc/MueLu_InterfaceAggregationFactory_def.hpp
mayrmt: Sure!

mayrmt:
@kliegeois Thanks for you comments and fixes! I updated the branch. Please have another look.

mayrmt:
@kliegeois Since the dual-to-primal node mapping uses local IDs, shall I rename the parameter list entry `"DualNodeID2PrimalNodeID"` to `"DualNodeLID2PrimalNodeLID"` to reflect the local ID character? I would address this along with #7164, if ok with you.

PR #7167, opened 2020-04-14
MueLu: Adding CM/RCM support to MueLu::Utilities_kokkos

csiefer2:
Wraps @brian-kelley's RCM code in Kokkos-Kernels

Tests: RHEL 6

### DIFF
--- a/packages/muelu/src/Utils/MueLu_Utilities_kokkos_def.hpp
+++ b/packages/muelu/src/Utils/MueLu_Utilities_kokkos_def.hpp
@@ -46,6 +46,7 @@
 #ifndef MUELU_UTILITIES_KOKKOS_DEF_HPP
 #define MUELU_UTILITIES_KOKKOS_DEF_HPP
 
+#include <algorithm>
 #include <Teuchos_DefaultComm.hpp>
 #include <Teuchos_ParameterList.hpp>
 
@@ -103,7 +104,22 @@
 
 #include <MueLu_Utilities_kokkos_decl.hpp>
 
+#include <KokkosKernels_Handle.hpp>
+#include <KokkosSparse_partitioning_impl.hpp>
+
+
 namespace MueLu {
+
+  template<class ViewType, class execution_space = typename ViewType::execution_space>
+  void kokkos_reverse(ViewType v) {    
+    //    typedef typename Space::execution_space execution_space;
+    int N = (int)v.extent(0);
+    Kokkos::parallel_for("MueLu::Kokkos_reverse::reverse",
+                         Kokkos::RangePolicy<execution_space>(0,N/2),
+                         KOKKOS_LAMBDA(const int & i){
+                           std::swap(v[i],v[N-1-1]);
+                         });    
+  }
 
   template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
   Teuchos::ArrayRCP<Scalar> Utilities_kokkos<Scalar, LocalOrdinal, GlobalOrdinal, Node>::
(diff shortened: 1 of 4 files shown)


brian-kelley:
@csiefer2 @lucbv It works, but it's slow on GPU. I do know how I would write a much faster BFS but I need to finish the cluster GS rewrite first.

Also, the RCM object can do ``rcm()`` for reverse order, or ``cuthill_mckee()`` for forward. So you don't have to call ``rcm()`` and then reverse it in your own code.

brian-kelley:
I'm really itching to do a fast BFS because it could also easily generalize to single-source shortest path, widest path, coarsening and possibly the sparse triangular solves in KK :)

packages/muelu/src/Utils/MueLu_Utilities_kokkos_def.hpp
```diff
@@ -636,6 +652,70 @@ namespace MueLu {
...
+    return retval;
+  }
+  
+  template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
+  Teuchos::RCP<Xpetra::Vector<LocalOrdinal,LocalOrdinal,GlobalOrdinal,Node> > CuthillMcKee(Xpetra::Matrix<Scalar,LocalOrdinal,GlobalOrdinal,Node> &Op) {
+    using local_matrix_type = typename Xpetra::Matrix<Scalar,LocalOrdinal,GlobalOrdinal,Node>::local_matrix_type;
+    using local_graph_type  = typename local_matrix_type::staticcrsgraph_type;
+    using device            = typename local_graph_type::device_type;
+
+    Teuchos::RCP<Xpetra::Vector<LocalOrdinal,LocalOrdinal,GlobalOrdinal,Node> > ordering = MueLu::ReverseCuthillMcKee(Op);
```
brian-kelley, line 687: Here, could do
```
    rcm_t rcm(localMatrix.numRows(), localMatrix.graph.row_map, localMatrix.graph.entries);
    lno_nnz_view_t cmOrder = rcm.cuthill_mckee(); 
```
and then don't need to reverse it.

csiefer2: Ooh!  I missed that!  Thanks for the pointer!

csiefer2:
> 
> 
> @csiefer2 @lucbv It works, but it's slow on GPU. I do know how I would write a much faster BFS but I need to finish the cluster GS rewrite first.


Yeah, we don't really need performance at current.  But when we do, I'll be sure to let you know!

packages/muelu/src/Utils/MueLu_Utilities_kokkos_def.hpp
```diff
@@ -103,8 +104,23 @@
 
 #include <MueLu_Utilities_kokkos_decl.hpp>
 
+#include <KokkosKernels_Handle.hpp>
+#include <KokkosSparse_partitioning_impl.hpp>
+
+
 namespace MueLu {
 
+  template<class ViewType, class execution_space = typename ViewType::execution_space>
```
brian-kelley, line 113: Oh I missed this. We could just delete this now, and also you're swapping ``i`` and ``N-2``, not ``N-1-i``.

csiefer2: Will delete on next PR (when I integrate this goop into the Pairwise aggregation.

jhux2:
> I'm really itching to do a fast BFS because it could also easily generalize to single-source shortest path, widest path, coarsening and possibly the sparse triangular solves in KK :)

It could also be used in Zoltan2!

kddevin:
@egboman @srajama1 @csiefer2 

Are you adding new ordering capability to @trilinos/muelu ?  
Zoltan2 has an ordering interface.  What are your needs?

lucbv:
@kddevin 
Chris and I are working on an aggregation algorithms that is sensitive to local ordering.
Here we are just creating interfaces to call the cm and rcm from kokkos-kernels on `KokkosSparse::CrsMatrix`.
I think that for these local operations it is probably appropriate to call kk directly instead of Zoltan2?

PR #7174, opened 2020-04-15
KokkosKernels: Add Jacobi-fused SpGEMM

seheracer:
@trilinos/kokkos-kernels 

## Motivation

This is a matching PR for https://github.com/kokkos/kokkos-kernels/pull/684. 

This PR adds support for Jacobi-fused SpGEMM in kokkos-kernels in the form of 
C = (I - w D^-1 A)B.

**Inputs:**
A: nxn sparse matrix with nonzero diagonal entries.
B: nxm sparse matrix
w: scalar
D^-1: nx1 2D view.

For the people who are familiar with kokkos-kernels SpGEMM:

- Jacobi-SpGEMM uses the existing spgemm_symbolic as is, because the nonzero patterns of A*B and (I - w D^-1 A)*B are the same as long as all diagonal entries of A have nonzero values.

- Most of the newly-added implementation files, whose names contain spgemm_jacobi, are updated versions of their spgemm_numeric counterparts. These updates incorporate w and D^-1 to the computation properly.

## Testing

Passes the cm_test_all_sandia spot-check. Output is given in https://github.com/kokkos/kokkos-kernels/pull/684

### DIFF
--- a/packages/kokkos-kernels/perf_test/sparse/CMakeLists.txt
+++ b/packages/kokkos-kernels/perf_test/sparse/CMakeLists.txt
@@ -15,6 +15,11 @@ KOKKOSKERNELS_ADD_EXECUTABLE(
   sparse_spgemm
   SOURCES KokkosSparse_spgemm.cpp
   )
+
+KOKKOSKERNELS_ADD_EXECUTABLE(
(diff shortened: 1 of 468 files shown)


seheracer:
Not needed anymore.

PR #7179, opened 2020-04-15
DO NOT MERGE Tpetra: Batched Apply Prototype

csiefer2:
A (correct but not exhaustively tested) implementation of Tpetra  "batchedApply" allowing you to trigger a group of applys that need the same importer (and input vectors).  This is designed with @cgcgcg's RefMaxwell work in mind.

TESTS: RHEL6

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Apply_Helpers.hpp
+++ b/packages/tpetra/core/src/Tpetra_Apply_Helpers.hpp
@@ -1,0 +1,148 @@
+// @HEADER
+// ***********************************************************************
+//
+//          Tpetra: Templated Linear Algebra Services Package
+//                 Copyright (2008) Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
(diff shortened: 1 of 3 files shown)


packages/tpetra/core/src/Tpetra_Apply_Helpers.hpp
```diff
@@ -0,0 +1,148 @@
...
+        X_colMap = rcp_const_cast<const MV>(X_colMapNonConst);
+      }
+
+      for(size_type i=0; i<N; i++) {
+        RCP<const export_type> exporter = Matrices[i]->getGraph()->getExporter();
+
+        // Temporary MV for doExport (if needed),
+        RCP<MV> Y_rowMap = Matrices[i]->getRowMapMultiVector(*Y[i]);
+        if (!exporter.is_null()) {
+          Matrices[i]->localApply(*X_colMap, *Y_rowMap, Teuchos::NO_TRANS, ZERO, ZERO);
```
cgcgcg, line 111: Doesn't `ZERO, ZERO` result in y=0\*y+0\*A\*x ?

csiefer2: oops

PR #7183, opened 2020-04-16
fix installation of generated ETI headers

jjwilke:
@trilinos/kokkos-kernels 

## Motivation
Need to install ETI headers that preserve tree structure. Currently installed in wrong place.


crtrott:
@trilinos/framework what do we need to do to make this actually run?

prwolfe:
> @trilinos/framework what do we need to do to make this actually run?

not sure I understand. The builds are running now. There was a wait while the previous PR (7179) completed.

crtrott:
Yeah I thought the PR Tester will immediately schedule stuff, but it will just take a while until all in-queue stuff is done. So the PR-Tester will only react when it comes back from a currently running test?

prwolfe:
Yes - we don't have infinite capacity so we have to wait to have the slots. Our scheduling method at the moment is not particularly efficient either.

PR #7189, opened 2020-04-16
ATDM Config:  Match 'arm' Without Version

jmgate:
The van1-tx2 environment currently allows the matching of 'arm-' plus a
version number to determine the compiler.  This allows one to simply
specify 'arm' as the compiler instead of needing a version as well.


jmgate:
@bartlettroscoe, since whipping up a function to do smarter parsing of the `JOB_NAME` is a more  substantial change, I figured it'd be better to hold that for a separate PR.

bartlettroscoe:
> @bartlettroscoe, since whipping up a function to do smarter parsing of the JOB_NAME is a more substantial change, I figured it'd be better to hold that for a separate PR.

@jmgate, that is fine.  I think it just occurred to me to how to write a poor-man's unit test harness for bash that should be good enough for the small amount of bash code we need to test.

PR #7196, opened 2020-04-17
CONFIG: Need to specify use of parallel netcdf

gsjaardema:
@trilinos/seacas 

## Motivation

The NetCDF and HDF5 libraries that are used in the ATDM configuration scripts are both
compiled with the parallel capability enabled; however, this capability is not being specified
to the CMake system, so it is unable to correctly enable the specific parts of the SEACAS
package that take advantage of the parallel NetCDF and HDF5 capabilities -- namely the 
auto-decomposition and auto-join (or 1->N and N->1) options.

This can cause confusion since the internal configuration output of the libraries show that 
the parallel is enabled and everything looks correct, but the files needed to make use of
the capability are not added to the build by CMake.


bartlettroscoe:
@gsjaardema, just to make sure, but did you test this with any of the existing supported ATDM Trilinos builds?  Do we know if this will work with, for example, the SEMS-built TPLs?  Does at least one of the Trilinos PR builds also set this?  If not, we should add this to one of the Trilinos PR build configurations as well in this PR so that we can be sure that the SEMS-built TPLs support this.  Or a lot of our ATDM Trilinos nightly builds and EMPIRE builds are going to be broken.

gsjaardema:
The EMPIRE build using SEMS TPLs was the build that originally started the investigation that led to this.  So, yes it does work with SEMS. It has also been tested with non-SEMS. Tried with a couple SEMS builds.

bathmatt:
I've tested it, I think we merge.  I'll approve if you want

bathmatt:
@jmgate this would be good to get into our pipeline if the tetra bug takes a while

gsjaardema:
@bathmatt I think this is ready to merge.

jmgate:
@bartlettroscoe, should we merge?

PR #7197, opened 2020-04-17
CONFIG: Update NetCDF and HDF5 TPL version

gsjaardema:
@trilinos/seacas 

## Motivation

The current versions of NetCDF (4.4.1) and HDF5 (1.8.12) are relatively old and there have
been significant improvements in capability, performance, scalability, and quality since that 
time.

For parallel scalability, there are configurations where the current TPL versions provide 2 to 3 orders of magnitude better performance for parallell IO.

## Testing

My guess is that the build for this will fail on one or more platforms / compilers / openmpi combinations, but that will provide good feedback on where the sems builds need to be provided or where there are issues.

Note that for HDF5-1.10.x versions, it is recommended that openmpi-4.0.X be used instead of openmpi-1.10.x.  I'm not sure what the timeframe for moving away from openmpi-1.10.x is, but hopefully it is soon...


jjellio:
@gsjaardema Just FYI, 

I've been using these versions on LLNL's ATS2 testbed rzansel  for a while, with no issues.  This version of HDF5 actually passes its unit tests on Power9 (1.8 and below would fail long double conversions)

bartlettroscoe:
@gsjaardema, looks like this PR has some build errors.

gsjaardema:
I think this is ready for a review and potential merging.  Other than the pull request testing, I have also tested on attaway (cts1) and ride and mutrino. The only library version updates are with the sems systems; the other systems are primarily cleaning up the dependencies to eliminate boost (which was never a dependency) and to make it more clear about where pnetcdf is coming from. 

I could split this into a couple PR -- one to update SEMS versions and one to do the dependency cleanup.  

It would also be good to work on updating library versions the ATDM systems.  For example, mutrino could use the system-supplied HDF5 and NetCDF libraries potentially...

Issue #7209, opened 2020-04-20
ShyLU:  ShyLU_NodeTacho_Tacho_TestOpenMPDoubleTaskScheduler_MPI_1 randomly segfaulting in GCC 4.8.4 PR build


bartlettroscoe:
## Bug Report
@trilinos/shylu, @trilinos/framework 

### Description

As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-01-01&end=2020-04-20&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Pull%20Request&field2=testname&compare2=61&value2=ShyLU_NodeTacho_Tacho_TestOpenMPDoubleTaskScheduler_MPI_1&field3=status&compare3=61&value3=failed&field4=testoutput&compare4=95&value4=signal%2011%20(Segmentation%20fault)) is very randomly failing in the Trilinos GCC 4.8.4 PR build twice:

Site | Build Name | Test Name | Status | Time | Proc Time | Details | Build Time | Processors
-- | -- | -- | -- | -- | -- | -- | -- | --
tr-test-7.novalocal | PR-7202-test-Trilinos_pullrequest_gcc_4.8.4-6361 | ShyLU_NodeTacho_Tacho_TestOpenMPDoubleTaskScheduler_MPI_1 | Failed | 280ms | 280ms | Completed (Failed) | 2020-04-19T21:18:22 MDT | 1
tr-test-7.novalocal | PR-7023-test-Trilinos_pullrequest_gcc_4.8.4-6064 | ShyLU_NodeTacho_Tacho_TestOpenMPDoubleTaskScheduler_MPI_1 | Failed | 220ms | 220ms | Completed (Failed) | 2020-03-16T09:17:47 MDT | 1

showing the segfault:

```
[----------] 5 tests from DenseByBlocks
[ RUN      ] DenseByBlocks.ldl
[       OK ] DenseByBlocks.ldl (0 ms)
[ RUN      ] DenseByBlocks.chol
[       OK ] DenseByBlocks.chol (1 ms)
[ RUN      ] DenseByBlocks.gemm
[       OK ] DenseByBlocks.gemm (4 ms)
[ RUN      ] DenseByBlocks.herk
[       OK ] DenseByBlocks.herk (2 ms)
[ RUN      ] DenseByBlocks.trsm
--------------------------------------------------------------------------
[...]

PR #7213, opened 2020-04-20
Disabling Tacho test with kokkos task scheduler instance.

kyungjoo-kim:
## Motivation

GCC 4.8.4 fails on the tests with kokkos task scheduler while the other tests using the same code but different task scheduler instances pass the PR tests. This could be related with compiler issues or the kokkos task scheduler or Tacho. Since Tacho mainly use task scheduler multiple implementation and the test passes, disabling test of task scheduler does not degrade the testing capability of Tacho. Thus, without further digging the problem, disabling the test with GCC 4.8.4 would be the solution for moving forward.

## Related Issues

#7209 

## Testing

This is about disabling a part of PR tests.


kyungjoo-kim:
Ah... This also includes the code exporting upper triangular factors to std vector containers. @crdohrm

PR #7225, opened 2020-04-21
Panzer: set virtual cells to have appropriate normals and rotation matrices

CamelliaDPG:
## Motivation
The rule is supposed to be that virtual cells have always have face ordinal 0 being the one that abuts the real cell, and that these have normals that are the negative of the normal on the real cell at the same point.  This commit makes the panzer Workset adhere to this rule.

@trilinos/panzer

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_Workset.cpp
+++ b/packages/panzer/disc-fe/src/Panzer_Workset.cpp
@@ -95,7 +95,6 @@ void WorksetDetails::setupNeeds(Teuchos::RCP<const shards::CellTopology> cell_to
                                 const Kokkos::View<double***,PHX::Device> & cell_vertices,
                                 const panzer::WorksetNeeds & needs)
 {
-
   const size_t num_cells = cell_vertices.extent(0);
   const size_t num_vertices_per_cell = cell_vertices.extent(1);
   const size_t num_dims_per_vertex = cell_vertices.extent(2);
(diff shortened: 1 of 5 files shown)


packages/panzer/disc-fe/src/Panzer_Workset.cpp
```diff
@@ -88,10 +88,11 @@ WorksetDetails::setup(const panzer::LocalMeshPartition & partition,
   fc->setup(partition);
   _face_connectivity = fc;
 
-  setupNeeds(partition.cell_topology, partition.cell_vertices, needs);
+  setupNeeds(partition, partition.cell_topology, partition.cell_vertices, needs);
 }
 
-void WorksetDetails::setupNeeds(Teuchos::RCP<const shards::CellTopology> cell_topology,
+void WorksetDetails::setupNeeds(const panzer::LocalMeshPartition &mesh_info,
```
seamill, line 94: We're going to need to do this without passing the mesh partition into the setupNeeds call. I believe this call is used by a lot of other codes, and other codes do not generate LocalMeshInfo objects.

The information you are using in the LocalMeshInfo object already exists in the Workset in the SubcellConnectivity object.

CamelliaDPG: @seamill, thanks for the note.  I've revised the PR accordingly.  I also fixed the failing test (check_rotation_matrices was enforcing the old, incorrect behavior on virtual cells).

CamelliaDPG:
@seamill, with the latest changes, the tests pass.  Would you please approve when you get a chance?

packages/panzer/disc-fe/src/Panzer_Workset.cpp
```diff
@@ -144,6 +143,73 @@ void WorksetDetails::setupNeeds(Teuchos::RCP<const shards::CellTopology> cell_to
...
+            {
+              for(int dim2=0; dim2<3; ++dim2)
+              {
+                iv->surface_rotation_matrices(virtual_cell,point_ordinal,dim1,dim2) = 0;
+              }
+            }
+          }
+        }
+      }
+    }
```
seamill, line 212: Could we move this chunk of code to a separate call? I have a branch that rewrites all of the workset IntegrationValues/BasisValues initializations and it would be nice to make this a modular interface outside of the workset.

CamelliaDPG: I have no objection to putting this code in a separate method at some point, but I don't fully understand your use case.  Wouldn't it make more sense for you to move this in your branch, after this is merged, since you know what you want here?  I'd like to get this merged in as soon as possible, since I have an EMPIRE MR that's blocked by this.

CamelliaDPG:
Thanks, @seamill!

CamelliaDPG:
@seamill Do you understand what's going on with the "Pre-Merge Inspection" here?  It says:
```
Pending — NOT INSPECTED YET - NOT REVIEWED
```
Does that require human intervention?

seamill:
Good question. @rppawlo Do panzer reviews require your/panzer approval?

PR #7233, opened 2020-04-22
Tpetra: unit test to reproduce failure after kokkos 3.1 promotion

rppawlo:
@trilinos/tpetra

### DIFF
--- a/packages/kokkos/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/packages/kokkos/core/src/impl/Kokkos_ViewMapping.hpp
@@ -1286,7 +1286,7 @@ struct ViewOffset<
   /* Span of the range space */
   KOKKOS_INLINE_FUNCTION
   constexpr size_type span() const {
-    return m_stride * m_dim.N1 * m_dim.N2 * m_dim.N3 * m_dim.N4 * m_dim.N5 *
+    return (m_dim.N0>size_type(0)?m_stride:size_type(0)) * m_dim.N1 * m_dim.N2 * m_dim.N3 * m_dim.N4 * m_dim.N5 *
            m_dim.N6 * m_dim.N7;
   }
 
@@ -1882,7 +1882,7 @@ struct ViewOffset<
 
   /* Span of the range space */
   KOKKOS_INLINE_FUNCTION
-  constexpr size_type span() const { return m_dim.N0 * m_stride; }
+  constexpr size_type span() const { return size() > size_type(0) ? m_dim.N0 * m_stride : size_type(0) ; }
 
   KOKKOS_INLINE_FUNCTION constexpr bool span_is_contiguous() const {
     return m_stride == m_dim.N7 * m_dim.N6 * m_dim.N5 * m_dim.N4 * m_dim.N3 *
--- a/packages/tpetra/core/test/BugTests/BugTests.cpp
+++ b/packages/tpetra/core/test/BugTests/BugTests.cpp
@@ -42,6 +42,7 @@
 */
 
 #include "Teuchos_UnitTestHarness.hpp"
+#include "Teuchos_Comm.hpp"
 #include "Tpetra_Core.hpp"
 #include "Tpetra_MatrixIO.hpp"
 #include "Tpetra_Map.hpp"
--- a/packages/tpetra/core/test/BugTests/BugTests.cpp
+++ b/packages/tpetra/core/test/BugTests/BugTests.cpp
@@ -174,7 +174,7 @@ namespace {
   {
     // test bug where map test checks that maps are the same object, instead of checking that maps are equivalent
 
-    using std::endl;    
+    using std::endl;
     // mfh 06 Aug 2017: There was no obvious reason why this test
     // required LO=int and GO=int, nor did the original Bug 5129 bug
     // report depend on specific LO or GO types, so I relaxed this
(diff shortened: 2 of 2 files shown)


rppawlo:
This test only fails on cuda.

crtrott:
I have been looking into this. It seems to me that someone is trying to do a sync on an out of range subview. The failure happens in the new prefetch capability for UVM views, But the pointer for the sync was never allocated and it seems out of range for the size. 

Note the following things: how this MV::DualView is allocated at 0x3effc0000a80 of size 16, then the whole ptr output is around the prefetch op and it goes through. Then it tries later to do the same for a pointer 0x3effc0000a90 which is actually the end pointer of that first allocation (and I checked that address was never allocated anywhere).  I wonder if it has something to do with the rest of this where the permuteToLIDs is 0 size? I think if this guy would be zero length it would actually work (i.e. not get through to that code point) and you could get there by taking a subview(a,make_pair(a.extent(0),a.extent(0)) which would be a 0 length view with .data being the endpointer of the original allocation. But if it did subview(a,make_pair(a.extent(0),a.extent(0)+2) it would be illegal.

```
KokkosP: Allocate<CudaUVM> name: MV::DualView pointer: 0x3effc0000a80 size: 16
KokkosP: Executing parallel-for kernel on device 0 with unique execution identifier 22
KokkosP: timer
KokkosP:   Kokkos::View::initialization
ptr A: 0x3effc0000a80 16
ptr B: 0x3effc0000a80
ptr C: 0x3effc0000a80
ptr D: 0x3effc0000a80
ptr E: 0x3effc0000a80
KokkosP: Execution of kernel 22 is completed.
KokkosP: Allocate<Host> name: DualView::modified_flags pointer: 0x3efffaa09e00 size: 8
KokkosP: Executing parallel-for kernel on device 0 with unique execution identifier 23
KokkosP: timer
KokkosP:   Kokkos::View::initialization
KokkosP: Execution of kernel 23 is completed.
KokkosP: Allocate<Host> name: permuteToLIDs pointer: 0x3efffaa0b5c0 size: 0
KokkosP: Allocate<Host> name: permuteFromLIDs pointer: 0x10023ee2d40 size: 0
KokkosP: Allocate<Host> name: exportLIDs pointer: 0x3efffaa0cc00 size: 0
KokkosP: Allocate<Host> name: permuteToLIDs pointer: 0x3efffaa0d900 size: 0
KokkosP: Allocate<Host> name: permuteFromLIDs pointer: 0x3efffaa0da40 size: 0
KokkosP: Allocate<Host> name: permuteFromLIDs pointer: 0x3efffaa0db40 size: 0
KokkosP: Executing parallel-for kernel on device 33554432 with unique execution identifier 24
KokkosP: timer
KokkosP:   fill
KokkosP: Execution of kernel 24 is completed.
ptr A: 0x3effc0000a90 16
ptr B: 0x3effc0000a90
```

crtrott:
ah and note that this would have slipped through prior to 3.1 since the sync would be a true no-op (source and dest are invalid, but they are the same). Only the prefetch now means that syncs of dual view of with invalid uvm ptrs now do something …

crtrott:
Ok I found the problem: if you get a 2D (or higher) subview of a layout left view with the first dimension using a `make_pair(i,i)` the stride of that first dimension is retained and you end up with an span calculation giving back the wrong number. In the rare case where the `i` is actually a.extent(0) and at the same time the second dimension is limited to `make_pair(a.extent(1)-1,a.extent(1))` you end up with the end pointer as the new data pointer and a span of larger than 0 which in combination will make the cudaPtrAttribute call fail. My suspicion is that this is actually a very 

standalone code showing the problem. The second subview has a span which is not 0 …

```c++
#include<Kokkos_Core.hpp>

template<class View>
void print_view(View a) {
  printf("%s %p %i %i %i %i %i\n",a.label().c_str(),a.data(),a.extent_int(0),a.extent_int(1),(int)a.stride(0),(int)a.stride(1),(int)a.span());
}
int main(int argc, char* argv[]) {
  Kokkos::initialize(argc,argv);
  {
     int N = (argc>1) ? atoi(argv[1]) : 1000000;
     int M = (argc>2) ? atoi(argv[2]) : 10000;
     using view_t = Kokkos::View<double**,Kokkos::LayoutLeft>;
     view_t a32("A32",3,2);
     printf("EndPtr: %p %p %i\n",a32.data(),a32.data()+a32.span(),(int)a32.span());
     print_view(a32);
     view_t a02("A02",0,2);
     print_view(a02);
     view_t a30("A30",3,0);
     print_view(a30);
     view_t sa32_b(a32,std::make_pair(0,2),std::make_pair(0,0));
     print_view(sa32_b);
     view_t sa32_a(a32,std::make_pair(3,3),std::make_pair(1,2));
     print_view(sa32_a);

  }
  Kokkos::finalize();
}
```
This bug existed at least as far back as Kokkos 2.7 - possibly it was always there. 

It also affects layout right when using a pair of length zero for the last argument, with all other extents being non-zero. I pushed a fix.

crtrott:
@rppawlo thanks for the reproducer. This was really hitting a corner case of a corner case ...

crtrott:
Pushed the fix for the new test now building in Serial due to use of Teuchos::MpiComm without including TeuchosComm ...

crtrott:
Disabled the test when MPI is not enabled.

rppawlo:
> @rppawlo thanks for the reproducer. This was really hitting a corner case of a corner case ...

Thanks for working on this @crtrott !

rppawlo:
@kddevin - Since I put up the PR, I can't approve @crtrott 's changes. Looks like the PR tests passed. Can you review/approve this?

PR #7238, opened 2020-04-22
Warning dup st

bathmatt:
Line 87 has a ST typedef, so I removed one.


bathmatt:
haven't tested but commit and let god (auto-tester) figure it out

PR #7239, opened 2020-04-22
ATDM Config:  Don't get-platform on Vortex

jmgate:
@bartlettroscoe, @e10harvey, recently the vortex admins installed `/etc/profile.d/snlvariables.sh` on vortex to define `SNLSYSTEM`, `SNLCLUSTER`, etc.  Somehow that broke the `load_matching_env.sh` script that gets installed with Trilinos.  It winds up throwing
```
ERROR: Unrecognized cluster vortex
```
and running our scripts with `set -e` means we die.  It looks like the error's coming from `/projects/sems/modulefiles/utils/get-platform`.  This PR avoids the call to `get-platform` on vortex as you did with astra so we can avoid that error and be about our merry way.

### DIFF
--- a/cmake/std/atdm/utils/get_known_system_info.sh
+++ b/cmake/std/atdm/utils/get_known_system_info.sh
@@ -161,7 +161,8 @@ if [[ "${SEMS_PLATFORM}" == "rhel6-x86_64" ]] ; then
 elif [[ "${SEMS_PLATFORM}" == "rhel7-x86_64" ]] ; then
   systemNameTypeMatchedList+=(sems-rhel7)
   systemNameTypeMatchedListHostNames[sems-rhel7]=sems-rhel7
-elif [[ "${SNLSYSTEM}" == "astra" ]] ; then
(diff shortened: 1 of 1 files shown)


jmgate:
Testing manually on vortex now.  It looks like everything's fine.

bartlettroscoe:
I tested on the SEMS RHEL6 machine 'crf450'

```
[rabartl@crf450 Trilinos (dont-get-platform-on-vortex)]$ . cmake/std/atdm/load-env.sh default
Hostname 'crf450.srn.sandia.gov' matches known ATDM host 'crf450.srn.sandia.gov' and system 'sems-rhel6'
Using SEMS RHEL6 compiler stack GNU-7.2.0 to build DEBUG code with Kokkos node type SERIAL
```

and 'stria':

```
[rabartl@stria-login1 Trilinos (dont-get-platform-on-vortex)]$ . cmake/std/atdm/load-env.sh default
Hostname 'stria-login1' matches known ATDM host 'stria-login1' and system 'van1-tx2'
```

Looks like it is working correctly.

Issue #6697, opened 2020-01-30
Why are Trilinos PR builds enabling Secondary Tested Code?


bartlettroscoe:
## Question
@trilinos/framework, @rppawlo, @trilinos/phalanx, @kddevin (Trilinos Data Services lead), @srajama1 (Trilinos Linear Solvers Lead), @

I just noticed that the Trilinos PR builds (or at least some of them are enabling secondary tested code as shown by, for example, the broken gcc_7.2.0 PR build shown [here](https://testing.sandia.gov/cdash/viewConfigure.php?buildid=6300498) which shows:

```
Enabling all SE packages that are not currently disabled because of Trilinos_ENABLE_ALL_PACKAGES=ON (Trilinos_ENABLE_SECONDARY_TESTED_CODE=ON) ...
```

That results in the enable of all of the secondary tested packages:

* `Final set of enabled packages:  TrilinosFrameworkTests Gtest Kokkos Teuchos KokkosKernels RTOp Sacado MiniTensor Epetra Zoltan Shards Triutils EpetraExt Tpetra TrilinosSS Domi Thyra Xpetra Isorropia Pliris AztecOO Galeri Amesos Pamgen Zoltan2 Ifpack ML Belos ShyLU_Node Amesos2 SEACAS Komplex Anasazi Ifpack2 Stratimikos FEI Teko Intrepid Intrepid2 STK Percept Phalanx NOX Moertel MueLu ShyLU_DD ShyLU Rythmos Tempus Stokhos ROL Piro Panzer TrilinosCouplings Pike 55`

and only the non-enabled packages:

* `Final set of non-enabled packages:  TriKota PyTrilinos NewPackage 3`

But the problem is that the ctest -S driver is **not** listing all 55 of these packages as CDash subjects as shown here:

* https://testing.sandia.gov/cdash/index.php?project=Trilinos&parentid=6300498

It only shows 53 of these 55 packages.  In particular, it is missing the 3 packages:

* TrilinosFrameworkTest
* Percept
* Moertel

One consequence of this is that it is listing Moertel build failures as Phalax build failures.

But the real question is, why is `Trilinos_ENABLE_SECONDARY_TESTED_CODE=ON`?  By definition, no important SNL customer is using non Primary-Tested packagse so why are Secondary Tested Packages getting enabled.

NOTE: The only Primary Tested Package that is not also a Secondary Tested package other than PyTrilinos and TriKota is Moertel.

bartlettroscoe:
[...]

PR #7241, opened 2020-04-22
Compadre: Initial snapshot intro Trilinos

### LINKED ISSUE #6934: Compadre:  Add New Package to Trilinos
This issue is to track the insertion of [Compadre](https://github.com/SNLComputation/compadre) into Trilinos, and is a part of the overall issue [Compadre#164](https://github.com/SNLComputation/compadre/issues/164).

CC:  @mperego, @kuberry

kuberry:
## Motivation
Package information found at https://github.com/SNLComputation/compadre/issues/165.  Closes #6934.

## Testing
This adds a new package, which is covered by CTests.  These new tests pass when building within Trilinos.


jmgate:
@jwillenbring, here are [some examples](https://github.com/trilinos/Trilinos/pull/7241#issuecomment-618473604) of infrastructure failures.  Looks like failures in communicating with GitHub.

jmgate:
@bartlettroscoe, I could use some help interpreting the CDash results for this PR.  Compadre only depends on Kokkos, so I would've expected this PR to only run the ~30 tests in Compadre.  It looks like thousands were run, though.

bartlettroscoe:
> @bartlettroscoe, I could use some help interpreting the CDash results for this PR. Compadre only depends on Kokkos, so I would've expected this PR to only run the ~30 tests in Compadre. It looks like thousands were run, though.

@jmgate, you touched a project level file:

* https://github.com/trilinos/Trilinos/pull/7241/files#diff-2f96afb76ad4b4cacebc8213b4b6486f

and that triggers the enabling of all Trilinos packages.

That is the only safe thing to do.

mperego:
@kuberry @jmgate: Are Compadre tests enabled? I could not find them looking at the CDash results.

bartlettroscoe:
> @kuberry @jmgate: Are Compadre tests enabled? I could not find them looking at the CDash results.

@mperego,

Since Compadre  is currently listed as an 'EX' package as shown here:

* https://github.com/trilinos/Trilinos/pull/7241/files#diff-2f96afb76ad4b4cacebc8213b4b6486fR112

it will not be tested in any of the Trilinos PR builds.  Instead, someone will need to test this offline before pushes to ensure that this is working on the desired platforms.

mperego:
> Since Compadre is currently listed as an 'EX' package as shown here:
> 
>     * https://github.com/trilinos/Trilinos/pull/7241/files#diff-2f96afb76ad4b4cacebc8213b4b6486fR112
> 
> 
> it will not be tested in any of the Trilinos PR builds. Instead, someone will need to test this offline before pushes to ensure that this is working on the desired platforms.

Thanks @bartlettroscoe I'd prefer to not mark it as experimental and have it tested. @jmgate @kuberry are you fine with that?

(I think experimental packages should be tested. Labeling a package as "Experimental" is a warning for users/developers  that the package is not  mature and that they should expect it to change significantly. I'll bring this up to the next Trilinos Product Leads meeting)

jmgate:
@kuberry, can you push an update that switches `EX` to `ST`?

kuberry:
Sure I can. I've made the change. I'm running cmake + builds + ctest and then I'll push.

kuberry:
@jmgate, just pushed the change. Looks like it needs reviewed again.

bartlettroscoe:
> (I think experimental packages should be tested. Labeling a package as "Experimental" is a warning for users/developers that the package is not mature and that they should expect it to change significantly. I'll bring this up to the next Trilinos Product Leads meeting)

@mperego, note that 'PT', 'ST', and 'EX' is a test category, not a statement of maturity (see  [Test Test Category](https://tribits.org/doc/TribitsDevelopersGuide.html#test-test-category)).  So, by definition, we should never add automated testing something labeled with **test category** 'EX'.

The problem is trying conflating "test category" and "maturity" of a package.  Since Trilinos never adopted the proposal for package maturity levels (see  ["TriBITS Lifecycle Model", 2012](https://bartlettroscoe.github.io/publications/TribitsLifecycleModel_eScience_2012.pdf)), there is no way to communicate with users (or other developers) about the expected maturity of our software.

packages/compadre/CMakeLists.txt
```diff
@@ -0,0 +1,617 @@
...
+
+  #KOKKOS-KERNELS
+  bob_input(KokkosKernels_PREFIX "" PATH "Path to KokkosKernels install")
+  if (KokkosKernels_PREFIX)
+    if(NOT IS_ABSOLUTE ${KokkosKernels_PREFIX})
+      set(KokkosKernels_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/${KokkosKernels_PREFIX}")
+    endif()
+    set(CMAKE_PREFIX_PATH ${KokkosKernels_PREFIX} ${CMAKE_PREFIX_PATH})
+  endif()
+  set(KOKKOS_IN_TRILINOS OFF)
```
lucbv, line 179: @kuberry will compadre be distributed outside of Trilinos or only inside Trilinos?
If it is distributed only within Trilinos then this CMake logic needs to be cleaned-up.

packages/compadre/CMakeLists.txt
```diff
@@ -0,0 +1,617 @@
...
+  elseif (Compadre_USE_MKL)
+    set(Compadre_USE_CUDA OFF)
+  endif() #if Compadre_USE_CUDA, all others already set to OFF
+  # if all off (default), then set Compadre_USE_LAPACK to ON
+  if ((NOT(Compadre_USE_LAPACK) AND NOT(Compadre_USE_MKL)) AND NOT(Compadre_USE_CUDA))
+    set(Compadre_USE_LAPACK ON)
+  endif()
+
+  # ensure TPL_ENABLE_*** configured consistent with Compadre_USE_***
+  if(Compadre_USE_LAPACK AND NOT(TPL_ENABLE_LAPACK))
```
lucbv, line 70: I would argue that these should trigger a `FATAL ERROR` instead of a message. It is faster to explicitly disable compadre at configure time to fix this issue than to wait for a linking error after Trilinos has been built.

packages/compadre/CMakeLists.txt
mperego: @lucbv Compadre will be distributed also outside Trilinos
https://github.com/SNLComputation/compadre

mperego:
@kuberry I still don't see Compadre tests. I'm fine merging this as is, but I would try to get it tested as soon as possible.

bartlettroscoe:
> @kuberry I still don't see Compadre tests. I'm fine merging this as is, but I would try to get it tested as soon as possible.

@mperego, they are listed under Ifpack such as you can see here:

* https://testing.sandia.gov/cdash/viewTest.php?onlypassed&buildid=6867782&filtercount=1&showfilters=1&field1=testname&compare1=65&value1=Compadre_

This is because the Trilinos PR system does not use TRIBITS_CTEST_DRIVER(); it is using a static list of Trilinos packages form 2 years ago.

mperego:
> > @kuberry I still don't see Compadre tests. I'm fine merging this as is, but I would try to get it tested as soon as possible.
> 
> @mperego, they are listed under Ifpack such as you can see here:
> 
>     * https://testing.sandia.gov/cdash/viewTest.php?onlypassed&buildid=6867782&filtercount=1&showfilters=1&field1=testname&compare1=65&value1=Compadre_
> 
> 
> This is because the Trilinos PR system does not use TRIBITS_CTEST_DRIVER(); it is using a static list of Trilinos packages form 2 years ago.

@bartlettroscoe I see, thanks. Should we change that static of Trilinos packages then?

mperego:
> > (I think experimental packages should be tested. Labeling a package as "Experimental" is a warning for users/developers that the package is not mature and that they should expect it to change significantly. I'll bring this up to the next Trilinos Product Leads meeting)
> 
> @mperego, note that 'PT', 'ST', and 'EX' is a test category, not a statement of maturity (see [Test Test Category](https://tribits.org/doc/TribitsDevelopersGuide.html#test-test-category)). So, by definition, we should never add automated testing something labeled with **test category** 'EX'.
> 
> The problem is trying conflating "test category" and "maturity" of a package. Since Trilinos never adopted the proposal for package maturity levels (see ["TriBITS Lifecycle Model", 2012](https://bartlettroscoe.github.io/publications/TribitsLifecycleModel_eScience_2012.pdf)), there is no way to communicate with users (or other developers) about the expected maturity of our software.

I see, thanks for the explanation.

bartlettroscoe:
> @bartlettroscoe I see, thanks. Should we change that static of Trilinos packages then?

@mperego, I am going to help the @trilinos/framework team to update their non-TriBITS ctest -S driver to keep this list up to date while working #6697 for them.  I might be able to get to to that later today.

NOTE: The issue in #6697 is that there are Trilinos Subpackages being used by ATDM customers that are **not** getting tested in Trilinos PR testing.

packages/compadre/CMakeLists.txt
```diff
@@ -0,0 +1,617 @@
...
+# Called at the top of every CMakeLists.txt file
+macro(include_tribits_build)
+  if (${PROJECT_NAME}_TRIBITS_DIR)
+    include("${CMAKE_CURRENT_SOURCE_DIR}/CMakeLists.tribits.cmake")
+    return()
+  endif()
+endmacro()
+
+# cmake helper functions
+include(${CMAKE_CURRENT_LIST_DIR}/cmake/bob.cmake)
```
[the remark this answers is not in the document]
kuberry: Hi @lucbv , it wasn't meant to be a place holder. It is Dan Ibanez's cmake helper package ( https://github.com/ibaned/cmake_bob ).

packages/compadre/CMakeLists.txt
kuberry: @lucbv, I only disable the package if the user doesn't enable the TPL. If I add a FATAL_ERROR, cmake will fail. Is this the desired behavior? I thought if a packages TPLs can't be satisfied, then the package should just be turned off?

lucbv: I do not know if Trilinos has a set rule for that?
I just feel that if you explicitly enable a package at configure time and it then gets disabled you might end up compiling Trilinos (sometime a slow process...) and only realize that there is an issue if you link against an executable.
I would rather bug the user and have him/her fix the configuration line right away.

mperego: @bartlettroscoe can you please weigh in on this? Thanks!

kuberry: The Dependencies.cmake requirements already are conditional on Compadre_USE_LAPACK, Compadre_USE_MKL, or Compadre_USE_CUDA. Each sets a different combination of required TPLs (the ones being checked against here). Theoretically, it isn't possible for the user to hit this case. This is just a double check that required TPLs are present.

kuberry: I'm still interested to know what is the desired behavior for a Trilinos package, but I think there is no circumstance under which this could get called.

mperego: Oh right, if cmake has a required dependency on LAPACK then I think we are good. @lucbv are you fine with this?

bartlettroscoe: > @bartlettroscoe can you please weigh in on this? Thanks!

Looking at the Dependencies.cmake file, I am not sure this this is consistent with how the rest of Trilinos deals with BLAS and LAPACK (and MKL).

Looking at:

```
$ find packages/ -name Dependencies.cmake -exec grep -l MKL {} \;
packages/amesos/cmake/Dependencies.cmake
packages/amesos2/cmake/Dependencies.cmake
packages/muelu/cmake/Dependencies.cmake
packages/kokkos-kernels/cmake/Dependencies.cmake
packages/shylu/shylu_node/basker/cmake/Dependencies.cmake
packages/shylu/shylu_node/tacho/cmake/Dependencies.cmake
packages/shylu/shylu_node/hts/cmake/Dependencies.cmake
packages/shylu/shylu_dd/bddc/cmake/Dependencies.cmake
packages/stokhos/cmake/Dependencies.cmake
```

it does not look like MKL is being used as a replacement for BLAS and LAPACK (and it can't if you think about it).

I did not have anything to do with the definition of the 'MKL' TPL or its usage in any package.  I just know that I have never seen it enabled in any of the projects I have worked with.

Looking at a few examples, it looks like it is mostly used for testing and not as an implementation for BLAS.

I would just stick with the BLAS and LAPACK TPLs and only define a dependency on MKL if you are using functions in MKL that go beyond BLAS and LAPACK.

Note that there is no such thing as "FindMKL.cmake" as shown at:

* https://cmake.org/cmake/help/v3.17/manual/cmake-modules.7.html

Looks like someone attempted to contribute a MKLConfig.cmake file to Intel:

* https://software.intel.com/en-us/forums/intel-math-kernel-library/topic/850824

but I don't think that went anywhere.

kuberry: Can the MKL be left in this PR as an option (it has to be explicitly set to ON, and it is not currently used [future use])?

@bartlettroscoe if a required TPL for a package wasn't found, would it be better to throw a fatal error or to turn off the package?

bartlettroscoe: > @bartlettroscoe if a required TPL for a package wasn't found, would it be better to throw a fatal error or to turn off the package?

@kuberry, if a required TPL can't be found, it is a fatal configure error.

My advice is to just have the simpler Dependencies.cmake file:

```
if (TPL_ENABLE_CUDA)
  set(EXTRA_REQUIRED_TPLS CUSOLVER CUBLAS)
endif()

tribits_package_define_dependencies(
  LIB_REQUIRED_PACKAGES
    KokkosCore KokkosContainers KokkosAlgorithms
  LIB_REQUIRED_TPLS
    LAPACK BLAS ${EXTRA_REQUIRED_TPLS}
  LIB_OPTIONAL_TPLS
    MPI CUDA
  )
```

Just leave it to TriBITS and the user configuring Trilinos on providing BLAS and LAPACK through MKL.

Don't bother with the MKL TPL unless you are using functionality in MKL beyond just what is in BLAS and LAPACK.

It is possible that way that the file compadre/cmake/Dependencies.cmake is currently defined, it might actually be broken when BLAS, LAPACK, and MKL are all enabled.  Has anyone actually tested this with the BLAS, LAPACK,and MKL enabled?

kuberry: @bartlettroscoe I removed all MKL logic since the package doesn't currently use it. This package either uses LAPACK+BLAS or CUBLAS+CUSOLVER. It doesn't use LAPACK routines if using CUSOLVER. I think what I have in Dependencies.cmake now should take care of this.

@lucbv I removed the double check for if TPLs were enabled since they must be (they are required TPLs in Dependencies.cmake). Effectively, since the TPLs are listed as required, if they are not enabled/found then the configure will fail (what you mentioned as the desired behavior).

I think this addresses everyone's comments/input. Thank you both for the input.

PR #7248, opened 2020-04-23
ifpack2: Reduce memory usage of Ifpack2::IlukGraph::initialize().

vbrunini:
@trilinos/ifpack2

In particular for the case where there are a small number of significantly
more dense rows. Rather than using the max number of entries from any
row as the heuristic for every row instead use the per-row values.
This reduces the memory use of a particular test case from ~1.2GB
down to ~350MB (which is consistent with the equivalent Aztec solver
memory use).

@trilinos/<teamName>

### DIFF
--- a/packages/ifpack2/src/Ifpack2_IlukGraph.hpp
+++ b/packages/ifpack2/src/Ifpack2_IlukGraph.hpp
@@ -527,9 +527,15 @@ void IlukGraph<GraphType>::initialize()
           Kokkos::RangePolicy<execution_space>(0, NumMyRows),
           KOKKOS_LAMBDA(const int i)
           {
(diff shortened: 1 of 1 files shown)


kddevin:
@vbrunini Thanks for the change.  Can you add a note about how you tested the change?  We are trying to capture testing info to satisfy ASC software quality requirements.  Thank you.

vbrunini:
Sure, I've already ran this change through the whole Aria test suite, and was planning to rely on the Trilinos pull request testing to run the Trilinos test suite.

For ongoing regression testing of this within Trilinos I'd suggest taking the test case with a few dense rows that @tjfulle has been working on and creating an RILUK Ifpack2 preconditioner for it.

kddevin:
Thanks, @vbrunini

vbrunini:
@csiefer2 Can you review the latest commit? Thanks!

PR #7249, opened 2020-04-23
7236 app action interface for subcyling stepper

asteyer:

### DIFF
--- a/packages/tempus/src/Tempus_StepperSubcycling_impl.hpp
+++ b/packages/tempus/src/Tempus_StepperSubcycling_impl.hpp
@@ -359,12 +359,6 @@ void StepperSubcycling<Scalar>::setAppAction(
   }
   this->isInitialized_ = false;
 }
-  /*
-template<class Scalar>
-  Teuchos::RCP<StepperSubcyclingAppAction<Scalar> >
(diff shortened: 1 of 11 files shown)


packages/tempus/src/Tempus_StepperSubcycling_impl.hpp
```diff
@@ -98,7 +104,32 @@ StepperSubcycling<Scalar>::StepperSubcycling(
...
+  std::string ICConsistency,
+  bool ICConsistencyCheck,
+  const Teuchos::RCP<StepperSubcyclingAppAction<Scalar> >& stepperSCAppAction)
+{
+  this->setStepperType(        "Subcycling");
+  this->setUseFSAL(            useFSAL);
+  this->setICConsistency(      ICConsistency);
+  this->setICConsistencyCheck( ICConsistencyCheck);
 
+  this->setObserver();
```
ccober6, line 123: Need an ifdef Tempus_HIDE_DEPRECATED_CODE around this observer call.

asteyer: Made this change.

asteyer: I also added the AppAction unit tests.

PR #7252, opened 2020-04-24
SEACAS: lib::fmt - fix shadow variable warnings

gsjaardema:
@trilinos/seacas 

## Motivation

Eliminate several shadowed variable warnings from the lib::fmt library used by the IOSS and SEACAS suplib_cpp library.

### DIFF
--- a/packages/seacas/libraries/ioss/src/fmt/chrono.h
+++ b/packages/seacas/libraries/ioss/src/fmt/chrono.h
@@ -515,8 +515,8 @@ namespace internal {
 
     Rep hour12() const
     {
-      Rep hour = static_cast<Rep>(mod((s.count() / 3600), 12));
-      return hour <= 0 ? 12 : hour;
+      Rep phour = static_cast<Rep>(mod((s.count() / 3600), 12));
+      return phour <= 0 ? 12 : phour;
     }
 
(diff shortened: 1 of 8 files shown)


gsjaardema:
@bartlettroscoe Yes, should have done a 'p' and maybe a 'l' or 'my' prefix, but this is an older version of the library so the patch won't be pushed to the developer and will be overwritten when a new version is pushed to Trilinos.

bartlettroscoe:
> @bartlettroscoe Yes, should have done a 'p' and maybe a 'l' or 'my' prefix, but this is an older version of the library so the patch won't be pushed to the developer and will be overwritten when a new version is pushed to Trilinos.

@gsjaardema, if this is just going to be overwritten, they why bother fixing this locally in Trilinos?  Is the developer going to fix these in the external version of the software?

gsjaardema:
@bartlettroscoe Yes, I have fixed the newer version with a PR to the developer.  I fixed it in Trilinos since it will probably stay there awhile until I figure out why the newer version causes a problem with Sparc that is difficult to reproduce.  This way, I don't have any (or fewer) SEACAS-related warnings in the Trilinos build. Since it only took about half hour to fix, I decided it was worthwhile and makes it easier to find new problems in Trilinos SEACAS builds instead of wading through a few thousand lines of template or other warnings.

PR #7254, opened 2020-04-24
MueLu Phase2a:  Include root node in aggregate

### LINKED ISSUE #6330: MueLu: Treatment of "root node" in Phase 2a aggregation
## Question
@trilinos/muelu 

Phase 2a loops over unaggregated nodes and checks whether a large enough aggregate can be formed around them. If so, it forms an aggregate that does not include the "root node". Why is it not included?

cgcgcg:
@trilinos/muelu 

## Motivation
Phase 2a aggregation searches for root nodes and unaggregated nodes around them to form a new aggregate. Currently, the new aggregate does not include the root node itself, which seems weird. 

In practice, the root node seems to be added in the expansion phase to its aggregate.
Adding the root node directly in Phase 2 will therefore result in the same aggregates in a lot of cases. 

Closes #6330


packages/muelu/src/Graph/UncoupledAggregation/MueLu_AggregationPhase2aAlgorithm_def.hpp
```diff
@@ -122,13 +122,17 @@ namespace MueLu {
...
+        vertex2AggId[rootCandidate] = aggIndex;
+        procWinner  [rootCandidate] = myRank;
         for (size_t k = 0; k < aggSize; k++) {
           aggStat     [aggList[k]] = AGGREGATED;
           vertex2AggId[aggList[k]] = aggIndex;
           procWinner  [aggList[k]] = myRank;
         }
 
-        numNonAggregatedNodes -= aggSize;
+        numNonAggregatedNodes -= aggSize+1;
```
jhux2, line 135: Shouldn't the root count towards the aggregate size?

PR #7255, opened 2020-04-24
Galeri matrix constructor Teuchos timers

GeoffDanielson:
@trilinos/galeri 

## Motivation

This cleans up the various timers implemented in the galeri matrix constructors, enabling apples-to-apples comparisons when epetra and tpetra are swapped out underneath it.

## Testing

This was tested using the muelu scaling driver for behavior.


_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-midnight
```
cgcgcg, line 1: I don't think that you want to commit that.

GeoffDanielson: You're right.

PR #7258, opened 2020-04-24
7188 app action interface for operator splitstepper

asteyer:
@trilinos/<ttempus>


packages/tempus/src/Tempus_StepperOperatorSplit_decl.hpp
```diff
@@ -58,6 +61,19 @@ class StepperOperatorSplit : virtual public Tempus::Stepper<Scalar>
     int order,
     int orderMin,
     int orderMax);
+#endif
+
+  /// Constructor                                                                                               
+  StepperOperatorSplit(
+    std::vector<Teuchos::RCP<const Thyra::ModelEvaluator<Scalar> > > appModels,
+    std::vector<Teuchos::RCP<Stepper<Scalar> > > subStepperList,
+    const Teuchos::RCP<StepperOperatorSplitAppAction<Scalar> >& stepperOSAppAction,
```
ccober6, line 70: To be consistent with the other steppers, we should probably make the stepperOSAppAction the last argument.  I did this because I have run into issues in the past were the compiler cannot tell what is wrapped inside the RCP, so it does not see the difference between the two argument signatures.

packages/tempus/src/Tempus_StepperOperatorSplit_impl.hpp
```diff
@@ -59,6 +62,43 @@ StepperOperatorSplit<Scalar>::StepperOperatorSplit(
...
+  this->setUseFSAL(            useFSAL);
+  this->setICConsistency(      ICConsistency);
+  this->setICConsistencyCheck( ICConsistencyCheck);
+
+  this->setSubStepperList(subStepperList);
+  this->setOrder   (order);
+  this->setOrderMin(orderMin);
+  this->setOrderMax(orderMax);
+
+  this->setObserver();
```
ccober6, line 100: Need a TEMPUS_HIDE_DREPRECATED_CODE around the setObserver.

packages/tempus/src/Tempus_StepperOperatorSplit_decl.hpp
asteyer: I fixed this issue!

packages/tempus/src/Tempus_StepperOperatorSplit_impl.hpp
asteyer: Fixed.

PR #7259, opened 2020-04-24
MueLu RefMaxwell: Changes to rebalancing

cgcgcg:
@trilinos/muelu

## Motivation
* Allow to use `RebalanceHeuristicFactory` with a map instead of a matrix.
* Use it in RefMaxwell.

### DIFF
--- a/packages/muelu/doc/UsersGuide/masterList.xml
+++ b/packages/muelu/doc/UsersGuide/masterList.xml
@@ -787,7 +787,7 @@
       problem.  The reverse algorithm uses it in the numerator.</description>
       <comment-ML>parameter not existing in ML</comment-ML>
     </parameter>
-    
+
     <parameter>
       <name>aggregate qualities: zero threshold</name>
       <type>double</type>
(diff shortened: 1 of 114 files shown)


packages/muelu/doc/UsersGuide/masterList.xml
```diff
@@ -1196,6 +1196,13 @@
       <description-ML>Minimum level to run partitioner. \muelu does not rebalance levels finer than this one.</description-ML>
     </parameter>
 
+    <parameter>
+      <name>repartition: use map</name>
+      <type>bool</type>
+      <default>false</default>
+      <description>Instead of using a matrix, use a map to determine rebalancing heuristics. Obviously, this cannot evaluate nonzero imbalance.</description>
```
jhux2, line 1203: You could also use a weighted vector W = spones(A)*ones(size(A,1),1).

cgcgcg: True, that'd be a good addition. I won't be using it in RefMaxwell though, since we really don't have a matrix yet at this point, and I'm not sure if we can determine the nnz/row upfront.

PR #7262, opened 2020-04-25
ShyLU/FROSch: Collection of different improvements for FROSch

searhein:
@trilinos/shylu 

## Motivation

This PR contains a bunch of improvements for FROSch, including:
+ Improvements of the output of FROSch.
+ Fix for the coarse basis for the `IPOUHarmonicCoarseOperator`: So far, in certain cases, the coarse basis was scaled in an inconsistent way within the orthogonalization. This resulted slightly wrong coarse basis. This was fixed by detecting and resolving linear dependencies without orthogonalization of the coarse basis.
+ Adding Thyra support to the `SubdomainSolver` class.
+ Some smaller fixes related to the deprecation of dynamic profiling in Tpetra.


srajama1:
Silly question: Why do you need Thyra support for subdomain solvers ? I am assuming you are thinking something like multigrid for subdomain solvers ? Otherwise you could just depend on Amesos2/Ifpack2 .. May be you wanted one interface for both of them ?

searhein:
So far, I used it for playing around with the computation of the extensions in the basis functions. Here, I used Belos + a preconditioner, and having Thyra seemed to be the easier way for playing around with the iterative solver and different preconditioners.

However, also gives me the option to use multigrid for the subdomains.

srajama1:
BTW, why WIP label ? Do you expect more changes ?

searhein:
Thanks, for the quick reviews.

I put WIP to prevent further auto-testing. Is there a different way to do so?

PR #7267, opened 2020-04-27
SEACAS: Fix enabling of ParMETIS; clean up

gsjaardema:
@trilinos/seacas 

## Stakeholder Feedback

SEACAS was using the wrong symbol to detect whether ParMETIS library was enabled.  Fixed to use the `TPL_ENABLE_ParMETIS` symbol and also cleaned up the places the resulting define was needed in the SEACAS code.  Removed a couple places where `parmetis.h` was included, but not used.

Verified that the functionality provided by ParMETIS was enabled in the resulting build.


packages/seacas/libraries/ioss/src/CMakeLists.txt
```diff
@@ -43,8 +43,7 @@ IF (NOT ${PROJECT_NAME}_ENABLE_Zoltan)
   ADD_DEFINITIONS(-DNO_ZOLTAN_SUPPORT)
 ENDIF()
 
-IF (${PROJECT_NAME}_ENABLE_ParMETIS)
-ELSE()
+IF (NOT TPL_ENABLE_ParMETIS)
```
bartlettroscoe, line 46: @gsjaardema, as described here:

* https://tribits.org/doc/TribitsDevelopersGuide.html#how-to-add-a-new-tribits-tpl-dependency

really this should be looking at `${PACKAGE_NAME}_ENABLE_ParMETIS` stead of `TPL_ENABLE_ParMETIS`.  That allows the user to control what package enable what TPLs in case there is a problem with some package's usage of a TPL.  For example, we have used this in the past to control what packages turn on support for HDF5 (since EpetraExt HDF5 support has been problematic in the past.)

gsjaardema: OK, I thought on Friday you said to use `TPL_ENABLE_ParMETIS`.  Will change to use `${PACKAGE_NAME}_ENABLE_ParMETIS`

gsjaardema: Regarding Tests, I am also redoing this in the SEACAS repository and adding tests there.  Wanted to get this going in Trilinos so @bathmatt could test.

bartlettroscoe: > OK, I thought on Friday you said to use `TPL_ENABLE_ParMETIS`. Will change to use `${PACKAGE_NAME}_ENABLE_ParMETIS`.

@gsjaardema, setting `-DTPL_ENABLE_ParMETIS=ON` is how you turn on the ParMETIS TPL at configure time to look for ParMETIS and then turn on optional support ParMETIS in the various (SE) Packages the list ParMETIS as an optional TPL.  See:

* https://tribits.org/doc/TribitsDevelopersGuide.html#tribits-dependency-handling-behaviors

and especially:

* https://tribits.org/doc/TribitsDevelopersGuide.html#support-for-optional-se-package-tpl-is-enabled-by-default
* https://tribits.org/doc/TribitsDevelopersGuide.html#support-for-optional-se-package-tpl-can-be-explicitly-disabled

Sorry for the confusion about this.

This will be a little less confusing once TPLs and Packages are handled more consistently as per tribitspub/tribits#63.

gsjaardema:
This will fail once it is built on an ATDM / non-sems machine due to the IOSS requirement that the real and integer IDXTYPE be set to 64-bits.  On attaway at least, the ParMETIS library is using 32-bit integer and real.  I will need to figure out how to handle this before this is merged...

packages/seacas/libraries/ioss/src/CMakeLists.txt
bartlettroscoe, line 46: So you decided not to change this to `${PACKAGE_NAME}_ENABLE_ParMETIS`?

gsjaardema: No, it wasn't working so I went back to `TPL_ENABLE_ParMETIS` until I could figure out why the `${PACKAGE_NAME}` version wasn't working...

bartlettroscoe: > No, it wasn't working so I went back to `TPL_ENABLE_ParMETIS` until I could figure out why the `${PACKAGE_NAME}` version wasn't working...

Okay, we can look into this later to see what is up with that.

Issue #5939, opened 2019-09-19
Set up unit testing infrastructure for ATDM Trilinos bash-based scripts


bartlettroscoe:
## Description:

The ATDM Trilinos system composed of bash scripts is complex enough that we need to start adding automated testing for the bash functions and scripts.

## Tasks:

- [X] Find a bash unit test harness and experiment with one ... Selected https://github.com/kward/shunit2/ ... PR #7202 
- [X] Create a Triinos TriBITS package called TrilinosATDMConfigTests ... PR #7202 
- [X] Write unit tests for atdm/utils/set_build_options.sh ... PR #7202
- [x] Write unit tests for atdm/utils/get_system_info.sh (both supported and custom systems).
- [ ] Write unit tests for atdm/load-env.sh (both known supported and custom systems).
- [ ] Run a basic local configure, build, and install for some faster building and running Trilinos packages (e.g. TeuchosCore) using atdm/ctest-s-local-test-driver.sh for 'sems-rhel7' env (only if that env is available) and ensure that the installation happens and has the right parts and they work correctly (i.e. `<prefix>/load_matching_env.sh`). (This can run in PR testing for one more more of the Trilinos PR builds since these run in RHEH7 machines that have the SEMS env.)

bartlettroscoe:
This looks pretty close to what we need:

* https://github.com/kward/shunit2/

bartlettroscoe:
CC: @e10harvey 

It just occurred to me how we might be able to do poor-man's unit testing with bash code without needing an elaborate unit test harness (and therefore another dependency).  I think I am going to experiment with that quickly.

bartlettroscoe:
I am looking at:

* https://github.com/kward/shunit2

some and a few things I see that I don't like:

* There a quite a few files.  It is not all that small (but they are trying to support a bunch of different shells)
* It supports reporting line numbers but not fine names in the error messages
* Does not print out the names of vars being checked.
* Does not print out the functions being called and checked.
* It does not have any tests of its own (how can I know the unit test harness actually works?)

I am going to download this tool and play with this a little and see what I can learn but I think I want to build my own small little unit test harness in bash that will have:

* Add asserts that will print the vars being checked and the functions being called.
* All asserts will support file name and line number reporting.
* It will report the total number of tests run and the number of total asserts and passing asserts (so I can grep for that in CTest).
* Some unit testing of itself (raw bash tests).

bartlettroscoe:
[...]

PR #7269, opened 2020-04-27
ATDM/ats1: Add initial ats1 env (ATDV-271)

e10harvey:
- Added new 'ats1' env to support customers.

# How was this tested?
- Extensively. See atdv-271.


cmake/std/atdm/ats1/all_supported_builds.sh
```diff
@@ -0,0 +1,13 @@
...
+
+export ATDM_CONFIG_ALL_SUPPORTED_BUILDS=(
+  ats1-hsw_intel-19.0.4_mpich-7.7.6_openmp_static_opt
+  ats1-hsw_intel-19.0.4_mpich-7.7.6_openmp_static_dbg
+  ats1-knl_intel-19.0.4_mpich-7.7.6_openmp_static_opt
+  ats1-knl_intel-19.0.4_mpich-7.7.6_openmp_static_dbg
+  ats1-hsw_intel-18.0.5_mpich-7.7.6_openmp_static_opt
+  ats1-hsw_intel-18.0.5_mpich-7.7.6_openmp_static_dbg
+  ats1-knl_intel-18.0.5_mpich-7.7.6_openmp_static_opt
+  ats1-knl_intel-18.0.5_mpich-7.7.6_openmp_static_dbg
```
bartlettroscoe, line 12: This list should be the actual set of builds we are running in nightly testing.  There is no way we can run this many builds currently with the current jenkins-srn.sandia.gov slave (with max 2 executors).  I think we are going to need to trim down significantly or we are going to need to switch to running these with a massively parallel cron driver.  (I actually have that a driver script that can do that in the repo https://gitlab.sandia.gov/atdm/atdm-trilinos-nightly-builds.  I used that for the driver on 'stria' until they got the Jenkins slave up.)

cmake/std/atdm/README.md
```diff
@@ -763,8 +766,9 @@ For example, to configure, build and run the tests for say `MueLu` on
 
 ```
 $ cd <some_build_dir>/
+$ export ATDM_CONFIG_REGISTER_CUSTOM_CONFIG_DIR=$TRILINOS_DIR/cmake/std/atdm/mutrino
```
bartlettroscoe, line 769: We can do better than that.  I will add a commit for this.

cmake/std/atdm/README.md
```diff
@@ -763,8 +766,9 @@ For example, to configure, build and run the tests for say `MueLu` on
...
 
-$ source $TRILINOS_DIR/cmake/std/atdm/load-env.sh intel-opt-openmp
+$ source $TRILINOS_DIR/cmake/std/atdm/load-env.sh mutrino-intel-opt-openmp
```
bartlettroscoe, line 771: But yes, the user will need to add 'mutrino' to the build name.

cmake/std/atdm/README.md
```diff
@@ -1177,7 +1181,7 @@ $ cd <some_build_dir>/
 # List available environments
 $ source $TRILINOS_DIR/cmake/std/atdm/load-env.sh help
 
-# Load hsw env and configure on the login node
+# Load arm env and configure on the login node
```
bartlettroscoe, line 1184: Nice catch!

cmake/std/atdm/README.md
```diff
@@ -1244,6 +1248,65 @@ fail.)
...
+One can also run tests in the background on a compute node using sbatch:
+```
+echo '#!/bin/bash' > do-test.sh
+echo 'ctest -j1' >> do-test.sh
+$ sbatch --output=do-test.out -N1 --time=2:00:00 -J ats1-do-test --account=<YOUR_WCID> do-test.sh
+```
+
+To drop to a compute node for an interactive session:
+```
+$ salloc -N1 --account=<YOUR_WCID>
```
bartlettroscoe, line 1289: I don't think you need the account for 'mutrino', do you?

cmake/std/atdm/utils/get_known_system_info.sh
```diff
@@ -106,9 +107,9 @@ elif [[ $realHostname == "white"* ]] ; then
 elif [[ $realHostname == "ride"* ]] ; then
   hostnameMatch=ride
   hostnameMatchSystemName=ride
-elif [[ $realHostname == "mutrino"* ]] ; then
+elif [[ $realHostname == "mutrino"* || $HOST == "mutrino"* ]] ; then
   hostnameMatch=mutrino
-  hostnameMatchSystemName=mutrino
+  hostnameMatchSystemName=ats1
```
bartlettroscoe, line 112: I will add a commit to move this logic out of this if-block and handle it below to add support for both the 'ats1' and 'mutrino' envs on the 'mutrino' cluster.

cmake/std/atdm/ats1/custom_builds.sh
```diff
@@ -0,0 +1,41 @@
...
+# customer request.
+elif atdm_match_any_buildname_keyword \
+       intel-18.0.5-mpich-7.7.6 \
+       intel-18.0.5_mpich-7.7.6 \
+       intel-18.0.5 \
+       intel-18 \
+       intel \
+  ; then
+  export ATDM_CONFIG_COMPILER=INTEL-18.0.5_MPICH-7.7.6
+else
```
bartlettroscoe, line 29: So this is a bit strange but the 'default' compiler is 'intel-19' but the default 'intel' compiler is 'intel-18'?  I think we should just make 'intel-18' the 'default' compiler as well to be consistent.  Do you agree?

cmake/std/atdm/ats1/custom_builds.sh
```diff
@@ -0,0 +1,41 @@
...
+  echo
+  echo "***"
+  echo "*** ERROR: A supported compiler was not selected for 'ats1' env - $ATDM_CONFIG_BUILD_NAME"
+  echo "***"
+  echo "*** Supported compilers include:"
+  echo "***"
+  echo "****  intel-19.0.4-mpich-7.7.6                  (intel-19, default)"
+  echo "****  intel-18.0.5-mpich-7.7.6                  (intel-18, intel)"
```
bartlettroscoe, line 37: Seems like 'intel-18' should be both the 'default' compiler and the default 'intel' complier, no?

cmake/std/atdm/ats1/environment.sh
```diff
@@ -0,0 +1,200 @@
...
+  export OMP_PLACES=threads
+  export OMP_PROC_BIND=spread
+  # Ensure that no more than 8 tasks, per srun command, are launched.
+  export ATDM_CONFIG_MPI_PRE_FLAGS="--mpi=pmi2;--ntasks-per-node=8"
+  # KNL nodes have 272 virtual cores.
+  # Allow now more than 34 virtual cores per task.
+  export ATDM_CONFIG_MPI_POST_FLAGS="--hint=nomultithread;-c 4"
+  export ATDM_CONFIG_CTEST_PARALLEL_LEVEL=1
+  # Assume we are building on a compute node
+  export ATDM_CONFIG_BUILD_COUNT=272
```
bartlettroscoe, line 59: The compute nodes on 'mutrino' can really handle 272 parallel object-file builds and links?

cmake/std/atdm/ats1/environment.sh
```diff
@@ -0,0 +1,200 @@
...
+  node_arch="hsw"
+  module unload craype-mic-knl
+  module load craype-haswell
+  # HSW nodes have 64 virtual cores.
+  # Allow no more than 8 virtual cores per task.
+  export ATDM_CONFIG_MPI_POST_FLAGS="-c 4"
+  # If we have 1 MPI rank per srun command and 1 cpu per task, we can run up to 32 2-threaded tests "in parallel" on virtual cores.
+  export ATDM_CONFIG_CTEST_PARALLEL_LEVEL=1
+  # Assume we are building on a compute node
+  export ATDM_CONFIG_BUILD_COUNT=64
```
bartlettroscoe, line 44: Why is the build parallel level set to 64 here but 272 for the `KNL` block below?  I think that might require some explaining.

cmake/std/atdm/README.md
```diff
@@ -1244,6 +1248,65 @@ fail.)
...
+```
+
+To use the `ctest-s-local-test-driver.sh` script, one must set one's WCID
+account using:
+
+```
+$ export ATDM_CONFIG_WCID_ACCOUNT=<YOUR_WCID>
```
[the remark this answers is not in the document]
e10harvey: No.

cmake/std/atdm/common/define_run_on_slurm_compute_node_func.sh
```diff
@@ -70,8 +78,8 @@ function atdm_run_script_on_compute_node {
   echo
   echo "Running '$script_to_run' using sbatch in the background ..."
   set -x
-  sbatch --output=$output_file --wait -N1 --time=${timeout} \
-    -J $ATDM_CONFIG_BUILD_NAME --account=${account} ${script_to_run} &
```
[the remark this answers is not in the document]
e10harvey: Good catch. --account should not be removed.

cmake/std/atdm/ats1/environment.sh
```diff
@@ -0,0 +1,200 @@
...
+elif [[ "$ATDM_CONFIG_KOKKOS_ARCH" == "KNL" ]]; then
+  node_arch="knl"
+  module unload craype-haswell
```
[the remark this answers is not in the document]
e10harvey: Nothing.

cmake/std/atdm/ats1/environment.sh
```diff
@@ -0,0 +1,200 @@
...
+export ATDM_CONFIG_MPI_PRE_FLAGS="--mpi=pmi2;--ntasks-per-node;36"
+
+# Common sparc tpl path values
+sparc_tpl_prefix_path="/usr/projects/sparc/tpls"
+system_name="ats1"
+node_arch=""
+
+if [[ "$ATDM_CONFIG_KOKKOS_ARCH" == "HSW" ]]; then
+  node_arch="hsw"
+  module unload craype-mic-knl
```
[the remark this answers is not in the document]
e10harvey: Nothing.

cmake/std/atdm/common/define_run_on_slurm_compute_node_func.sh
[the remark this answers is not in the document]
e10harvey: We can simplu add `--account=${account}` back, this argument will not cause any issues on ats1.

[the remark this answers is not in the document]
bartlettroscoe: Does SLURM simply ignore the `--account` argument on 'mutrino'?  Can you pass in an invalid account and it will be ignored?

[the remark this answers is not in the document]
e10harvey: That's what I've observed.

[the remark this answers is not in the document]
bartlettroscoe: We might want to make that explicit then by setting :

```
export ATDM_CONFIG_WCID_ACCOUNT=IGNORED
```

in the file atdm/ats1/enviornment.sh

Then when someone sees `sbatch ... --account=IGNORED ...` being run, they will know something is up and will not think that the `--account` argument is actually being used.

bartlettroscoe:
FYI: In commit 30fc6fda0ba I made it so that the old 'mutirno' env works by just putting 'mutrino' in the build name.  I tested this on the cluster 'mutirno' with:

```
Hostname 'mutrino' matches known ATDM host 'mutrino' and system 'ats1'
Using ats1 compiler stack INTEL-18.0.5_MPICH-7.7.6 to build DEBUG code with Kokkos node type SERIAL and KOKKOS_ARCH=HSW

$ . cmake/std/atdm/load-env.sh mutrino-default
Hostname 'mutrino' matches known ATDM host 'mutrino' and system 'mutrino'
Setting compiler and build options for build-name 'mutrino-default'
Using mutrino compiler stack INTEL to build DEBUG code with Kokkos node type SERIAL and KOKKOS_ARCH=HSW

$ . cmake/std/atdm/load-env.sh ats1-default
Hostname 'mutrino' matches known ATDM host 'mutrino' and system 'ats1'
Setting compiler and build options for build-name 'ats1-default'
Using ats1 compiler stack INTEL-18.0.5_MPICH-7.7.6 to build DEBUG code with Kokkos node type SERIAL and KOKKOS_ARCH=HSW
```

The next task in #5939 will be to add unit tests for `atdm/utils/get[_known]_system_info.sh` and refactor all of it to use the same logic (to be more flexible and maintainable).

jmgate:
Does this mean if we have `mutrino` in the build name, it'll load the old environment that doesn't have the Intel 18 option, and if we want the new environment, we'll need to replace `mutrino` with `ats1` in the build name?

bartlettroscoe:
> Does this mean if we have `mutrino` in the build name, it'll load the old environment that doesn't have the Intel 18 option, and if we want the new environment, we'll need to replace `mutrino` with `ats1` in the build name?

@jmgate, yes.  Or just leave the env name out of the build name and it will pick the new 'ats1' env by default.  Hopefully that is pretty logical. 

The old 'mutrino' env will hopefully go away very soon but we are still struggling to get the test suite to run with the new 'ats1' env.

bartlettroscoe:
@e10harvey, I am on 'mutrino' just trying to build and run Kokkos and Teuchos tests with:

```
$ cd /lscratch1/rabartl/Trilinos.base/BUILDS/ATS1/CTEST-S/

$ env Trilinos_PACKAGES=Kokkos,Teuchos ./ctest-s-local-test-driver.sh all

***
*** ./ctest-s-local-test-driver.sh
***

ATDM_TRILINOS_DIR = '/lscratch1/rabartl/Trilinos.base/Trilinos'

Load some env to get python, cmake, etc ...

Hostname 'mutrino' matches known ATDM host 'mutrino' and system 'ats1'
Using ats1 compiler stack INTEL-18.0.5_MPICH-7.7.6 to build DEBUG code with Kokkos node type SERIAL and KOKKOS_ARCH=HSW

Running builds:
    ats1-hsw_intel-19.0.4_mpich-7.7.6_openmp_static_opt
    ats1-hsw_intel-19.0.4_mpich-7.7.6_openmp_static_dbg
    ats1-knl_intel-19.0.4_mpich-7.7.6_openmp_static_opt
    ats1-knl_intel-19.0.4_mpich-7.7.6_openmp_static_dbg
    ats1-hsw_intel-18.0.5_mpich-7.7.6_openmp_static_opt
    ats1-hsw_intel-18.0.5_mpich-7.7.6_openmp_static_dbg
    ats1-knl_intel-18.0.5_mpich-7.7.6_openmp_static_opt
    ats1-knl_intel-18.0.5_mpich-7.7.6_openmp_static_dbg

Tue Apr 28 11:10:09 MDT 2020

Running Jenkins driver Trilinos-atdm-ats1-hsw_intel-19.0.4_mpich-7.7.6_openmp_static_opt.sh ...

    See log file Trilinos-atdm-ats1-hsw_intel-19.0.4_mpich-7.7.6_openmp_static_opt/smart-jenkins-driver.out

???
```

and so far it is stuck just trying to run the Kokkos tests and is showing:

```
Running tests (parallel level 1) ...


Test project /lscratch1/rabartl/Trilinos.base/BUILDS/ATS1/CTEST-S/Trilinos-atdm-ats1-hsw_intel-19.0.4_mpich-7.7.6_openmp_static_opt/SRC_AND_BUILD/BUILD
        Start   1: KokkosCore_UnitTest_OpenMP_MPI_1
  1/167 Test   #1: KokkosCore_UnitTest_OpenMP_MPI_1 ....................................***Timeout 600.12 sec
        Start   2: KokkosCore_UnitTest_OpenMPInterOp_MPI_1
  2/167 Test   #2: KokkosCore_UnitTest_OpenMPInterOp_MPI_1 .............................***Timeout 600.10 sec
        Start   3: KokkosCore_UnitTest_Default_MPI_1
  3/167 Test   #3: KokkosCore_UnitTest_Default_MPI_1 ...................................***Timeout 600.11 sec
        Start   4: KokkosCore_UnitTest_PushFinalizeHook_MPI_1
```

Any idea what the problem might be?

e10harvey:
> Any idea what the problem might be?

@bartlettroscoe: No, I am not seeing these timeouts. Did you follow the instructions in the README.md? Did you specify the host and run the `hostname` smoke test?

bartlettroscoe:
> @bartlettroscoe: No, I am not seeing these timeouts. Did you follow the instructions in the README.md? Did you specify the host and run the hostname smoke test?

@e10harvey, I just logged onto 'mutrino', pulled the branch, and then ran:

```
$ env Trilinos_PACKAGES=Kokkos,Teuchos ./ctest-s-local-test-driver.sh all
```

which is basically equivalent to what the Jenkins driver (our a cron driver) would do.

The ctest -S driver has to be able to start from the login node and take care of the details of allocating resources and doing the configure and build.

e10harvey:
> which is basically equivalent to what the Jenkins driver (our a cron driver) would do.
> 
> The ctest -S driver has to be able to start from the login node and take care of the details of allocating resources and doing the configure and build.

Got it. Seems like we're seeing these issues with the mutrino env as well so we'll have to go around and compile a list from the `hostname` smoke test and iterate.

bartlettroscoe:
> Seems like we're seeing these issues with the mutrino env as well so we'll have to go around and compile a list from the hostname smoke test and iterate.

@e10harvey,

If you look at the [mutrino builds over the last 2 months](https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&begin=2020-03-01&end=2020-04-27&filtercount=1&showfilters=1&field1=buildname&compare1=65&value1=Trilinos-atdm-mutrino) you can see we started seeing missing test results randomly but fairly regularly starting on 2020-04-16.  We should look at the corresponding jenkins-srn.sandia.gov jobs to see if the days that we did not get test results were on the same (bad) nodes on 'mutrino'.

jjellio:
@e10harvey @bartlettroscoe I've been using mutrino's ENV a fair amount.

I changed the modules underpinning it to SPARC's (to get netcdf 4.7 among other things).

The single change above requires some modifications to the LIB env variables (netcdf chooses to put some libraries in lib64 vs lib)

I've also modified it to allow shared builds (which is super convenient for build size/speed + being able to LD_PRELOAD symbol overrides)

To do shared builds, you need to do something like:
```
if [ "$LINKAGE" == "shared" ]; then
  export ATDM_CONFIG_CMAKE_SKIP_INSTALL_RPATH=OFF
  export CRAYPE_LINK_TYPE=dynamic
  # how to set this ??
  emp_extra_cmake+=('-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON')
fi
```
Getting CMake to do the Rpath on Cray seems like voodoo, I'm guessing some CMake logic is telling CMake the system doesn't support shared linking (it does, you just need a compiler flag or ENV as I set above).

There is also a typo in ATDMDevSetttings.cmake 
It may make sense to review `ATDMDevSetttings.cmake`in its entirety for places where `$ENV{}` need to be. I hit it for this one case, where I noticed it wouldn't override the link options: `Trilinos_EXTRA_LINK_FLAGS`
```
diff --git a/cmake/std/atdm/ATDMDevEnvSettings.cmake b/cmake/std/atdm/ATDMDevEnvSettings.cmake
index feb6c55..6c11bad 100644
--- a/cmake/std/atdm/ATDMDevEnvSettings.cmake
+++ b/cmake/std/atdm/ATDMDevEnvSettings.cmake
@@ -246,7 +246,7 @@ ENDIF()
 
 # Other various options
 ATDM_SET_CACHE(CTEST_BUILD_FLAGS "-j$ENV{ATDM_CONFIG_BUILD_COUNT}" CACHE STRING)
-ATDM_SET_CACHE(Trilinos_EXTRA_LINK_FLAGS "${ATDM_EXTRA_LINK_FLAGS}" CACHE STRING)
+ATDM_SET_CACHE(Trilinos_EXTRA_LINK_FLAGS "$ENV{ATDM_EXTRA_LINK_FLAGS}" CACHE STRING)
 ATDM_SET_CACHE(CMAKE_JOB_POOL_LINK "${ATDM_CMAKE_JOB_POOL_LINK}" CACHE STRING)
 ATDM_SET_CACHE(Trilinos_PARALLEL_COMPILE_JOBS_LIMIT "${ATDM_PARALLEL_COMPILE_JOBS_LIMIT}"
   CACHE STRING)
```

bartlettroscoe:
> I've been using mutrino's ENV a fair amount.

@jjellio, which ENV? The old ATDM Trilinos 'mutrino' env or the new ATDM Trilinos 'ats1' env?

> To do shared builds, you need to do something like:

@jjellio, can you please post a new PR with your ideas of what needs to change and then we can tweak things there?

jjellio:
mutrino's ENV. This was before any ATS1 thing was added

bartlettroscoe:
> The single change above requires some modifications to the LIB env variables (netcdf chooses to put some libraries in lib64 vs lib)

@jjellio, what "single change above"?

jjellio:
Wasn't specific:

```
# Load the modules (can't purge)
-module load devpack/20180124/cray/7.6.2/intel/17.0.4
-module load gcc/4.9.3
+module load sparc-dev/intel-19.0.4_mpich-7.7.6_hsw
+# when using sparc, they put netcdf into lib64 not lib... ugh
+## module load devpack/20180124/cray/7.6.2/intel/17.0.4
+## module load gcc/4.9.3
 module load cmake/3.14.6

# sparc will set their installed Trilinos' bin/ into the path... this means you will use their SEACAS
# not what you just built (this is dangerous for unit testing)
# remove the path where decomp sits .... then after *we* build trilinos we will find it in *our* bin
+path_remove $(dirname $(which decomp))
 
-export ATDM_CONFIG_NETCDF_LIBS="-L${NETCDF_ROOT}/lib;${NETCDF_ROOT}/lib/libnetcdf.a;${PNETCDF_ROOT}/lib/libpnetcdf.a;${ATDM_CONFIG_HDF5_LIBS};-lm"
+export ATDM_CONFIG_NETCDF_LIBS="-L${BOOST_ROOT}/lib;-L${NETCDF_ROOT}/lib64;-L${PNETCDF_ROOT}/lib;-L${HDF5_ROOT}/lib;${BOOST_ROOT}/lib/libboost_program_options.a;${BOOST_ROOT}/lib/libboost_system.a;${NETCDF_ROOT}/lib64/libnetcdf.a;${PNETCDF_ROOT}/lib/libpnetcdf.a;${HDF5_ROOT}/lib/libhdf5_hl.a;${HDF5_ROOT}/lib/libhdf5.a;-lz;-ldl;-lm"
```

In the above netcdf line, I've changed to`lib64`, but this isn't a robust solution.
A better approach would be to set things like:
```-L${NETCDF_ROOT}/lib;-L${NETCDF_ROOT}/lib64;-lnetcdf;-lpnetcdf```

That would allow it to pickup either, or add logic above to change the library for netcdf based on where it is found.

OR, (the ultimate solution)
don't specify direct paths, reorganize ATDM_CONFIG_ to accept the library directories and another to accept the library names (the way CMake usually works). Then, Cmake will find the proper libs (shared or static) at configure time.

jjellio:
see also the note about sparc's installed Trilinos' bin being in the path, so you pickup installed trilinos binaries, rather than what you built. This is a problem w/EMPIRE (and anything that does installed testing), as they may end up with a different exodus than what they built.

bartlettroscoe:
> `+module load sparc-dev/intel-19.0.4_mpich-7.7.6_hsw`

@jjellio, we can't use the `sparc-dev` modules in our ATDM Trilinos configuration process because EMPIRE and GEMMA developers can't access those modules.

jjellio:
I wasn't really suggesting anyone should.  They are simply the path of least resistance for me to get updated TPLs.

The sticking point is that for  whatever reason, some things are getting lumped into `/tpl_root/lib64` when you expect `/lib`

It's also worth keeping in mine the addition an installed Trilinos to your path. This has implications for CMake and unit testing.

Issue #7275, opened 2020-04-28
Configure errors in all 'ats2', 'sems-rhel6' and 'tlcc2' with BOOST_ROOT starting 2020-04-28


bartlettroscoe:
CC: @gsjaardema 

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label "impacting: tests"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&date=2020-04-28&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-&field2=configureerrors&compare2=43&value2=0) the there are configure errors in the builds:

* `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg`
* `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg_cuda-aware-mpi`
* `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt`
* `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_opt_cuda-aware-mpi`
* `Trilinos-atdm-ats2-gnu-7.3.1-spmpi-2019.06.24_serial_static_dbg`
* `Trilinos-atdm-ats2-gnu-7.3.1-spmpi-2019.06.24_serial_static_opt`
* `Trilinos-atdm-sems-rhel6-gnu-7.2.0-openmp-complex-shared-release-debugN`
* `Trilinos-atdm-sems-rhel6-gnu-7.2.0-openmp-debug	`
[...]

PR #7276, opened 2020-04-28
ATDM: Add back removed BOOST_ROOT env vars (#7275)

bartlettroscoe:
I did not test these but these envs were all broken (due to PR #7197 as reported in #7275) so adding this back can't hurt.


gsjaardema:
Looks good.  I had just pushed the same changes and saw you had already done this, so aborting my PR

PR #7279, opened 2020-04-28
CONFIG: Fix pnetcdf and netcdf setting broke in #7197

gsjaardema:
@trilinos/seacas 

## Motivation

This should fix the NetCDF-related (or PnetCDF) failures seen on some shared-library builds. 
Note that the SEMS installs have shared/static NetCDF and HDF5 libraries, but for some reason, they use a static-only libpnetcdf.a.  This PR fixes the specification of the libpnetcdf library to always be .a in a shared or non-shared build.

It also correctly sets the NETCDF_ROOT CMake variable which is needed by the FindNetCDF.cmake package find module.

### DIFF
--- a/cmake/std/atdm/sems-rhel6/environment.sh
+++ b/cmake/std/atdm/sems-rhel6/environment.sh
@@ -195,7 +195,7 @@ fi
 
 export ATDM_CONFIG_HDF5_LIBS="${HDF5_ROOT}/lib/libhdf5_hl.${ATDM_CONFIG_TPL_LIB_EXT};${HDF5_ROOT}/lib/libhdf5.${ATDM_CONFIG_TPL_LIB_EXT};${ZLIB_ROOT}/lib/libz.${ATDM_CONFIG_TPL_LIB_EXT};-ldl"
 
-export ATDM_CONFIG_NETCDF_LIBS="-L${NETCDF_ROOT}/lib;${NETCDF_ROOT}/lib/libnetcdf.${ATDM_CONFIG_TPL_LIB_EXT};${PNETCDF_ROOT}/lib/libpnetcdf.${ATDM_CONFIG_TPL_LIB_EXT};${ATDM_CONFIG_HDF5_LIBS};-lcurl"
+export ATDM_CONFIG_NETCDF_LIBS="-L${NETCDF_ROOT}/lib;${NETCDF_ROOT}/lib/libnetcdf.${ATDM_CONFIG_TPL_LIB_EXT};${PNETCDF_ROOT}/lib/libpnetcdf.a;${ATDM_CONFIG_HDF5_LIBS};-lcurl"
 
 # NOTE: SEMS does not provide a *.a files for PNetCDF so we can't use them in
 # a shared lib build :-(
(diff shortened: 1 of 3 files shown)


cmake/std/atdm/sems-rhel6/environment.sh
```diff
@@ -195,7 +195,7 @@ fi
 
 export ATDM_CONFIG_HDF5_LIBS="${HDF5_ROOT}/lib/libhdf5_hl.${ATDM_CONFIG_TPL_LIB_EXT};${HDF5_ROOT}/lib/libhdf5.${ATDM_CONFIG_TPL_LIB_EXT};${ZLIB_ROOT}/lib/libz.${ATDM_CONFIG_TPL_LIB_EXT};-ldl"
 
-export ATDM_CONFIG_NETCDF_LIBS="-L${NETCDF_ROOT}/lib;${NETCDF_ROOT}/lib/libnetcdf.${ATDM_CONFIG_TPL_LIB_EXT};${PNETCDF_ROOT}/lib/libpnetcdf.${ATDM_CONFIG_TPL_LIB_EXT};${ATDM_CONFIG_HDF5_LIBS};-lcurl"
+export ATDM_CONFIG_NETCDF_LIBS="-L${NETCDF_ROOT}/lib;${NETCDF_ROOT}/lib/libnetcdf.${ATDM_CONFIG_TPL_LIB_EXT};${PNETCDF_ROOT}/lib/libpnetcdf.a;${ATDM_CONFIG_HDF5_LIBS};-lcurl"
```
bartlettroscoe, line 198: @gsjaardema, if one is setting the cache var `NETCDF_ROOT` does that mean that this list of libs `ATDM_CONFIG_NETCDF_LIBS` will get ignored?  Is not this line:

* https://github.com/trilinos/Trilinos/blob/cbbac67c79ef6b112ad14da3da8e96c318c09cf8/cmake/std/atdm/ATDMDevEnvSettings.cmake#L432

gsjaardema: I think we need to go through the `FindTPLNetCDF.cmake` and `FindNetCDF.cmake` files sometime and also the logic in the *.sh and *.cmake configuration files.  It is somewhat confusing to someone who is not working with them all the time.  It looks like the line you have highlighted should do the correct thing, but it didn't in this case, (but seemed to in other cases for #7197).  

While working on #7197, none of the builds that I tested went into the `FindNetCDF.cmake`, but for some reason this one was going into that file and the variables it was looking for were not set.  Setting `NETCDF_ROOT` fixed it so it knew where to look once it was inside `FindNetCDF.cmake`, but I'm not sure why it went in there...

gsjaardema: It might be better to put the setting of `NETCDF_ROOT` inside the non-sparc else() block...

bartlettroscoe: > It might be better to put the setting of NETCDF_ROOT inside the non-sparc else() block...

It will not help.  If `ATDM_CONFIG_NETCDF_LIBS` is set in the env, then that will result in the CMake cache var `TPL_Netcdf_LIBRARIES` getting set and that will ignore anything found by `find_package(NetCDF)`.  (That is the purpose of the TriBITS var`TPL_<TPL_NAME>_LIBRARIES` variable; to provide a a hard override of any auto find logic.)

PR #7283, opened 2020-04-29
MueLu Zoltan2Interface: Handle blockSize>1

cgcgcg:
@trilinos/muelu 

## Motivation
Block size > 1 currently segfaults. With this change, we force blocks to all be on the same partition, ignoring the Zoltan2 solution for all but one rows per block.

### DIFF
--- a/packages/muelu/src/Rebalancing/MueLu_Zoltan2Interface_def.hpp
+++ b/packages/muelu/src/Rebalancing/MueLu_Zoltan2Interface_def.hpp
@@ -256,8 +256,9 @@ namespace MueLu {
 
       const typename InputAdapterType::part_t * parts = problem->getSolution().getPartListView();
 
-      for (GO i = 0; i < numElements; i++) {
-        int partNum = parts[i];
+      // For blkSize > 1, ignore solution for every row but the first ones in a block.
+      for (GO i = 0; i < numElements/blkSize; i++) {
+        int partNum = parts[i*blkSize];
 
         for (LO j = 0; j < blkSize; j++)
           decompEntries[i*blkSize + j] = partNum;


cgcgcg:
@trilinos/framework 
```
    Error: A license for Comp-CL is not available now (-97,121).
    
    A connection to the license server could not be made.  You should
    make sure that your license daemon process is running: both an
    lmgrd process and an INTEL process should be running
    if your license limits you to a specified number of licenses in use
    at a time.  Also, check to see if the wrong port@host or the wrong
    license file is being used, or if the port or hostname in the license
    file has changed.
    
    License file(s) used were (in this order):
        1.  Trusted Storage
    **  2.  /projects/sems/install/rhel6-x86_64/sems/compiler/intel/17.0.1/base/Licenses/intel-Linux-SRN.lic
    **  3.  /projects/sems/install/rhel6-x86_64/sems/compiler/intel/17.0.1/base/compilers_and_libraries_2017.1.132/linux/bin/intel64/../../Licenses
    **  4.  /ascldap/users/trilinos/Licenses
    **  5.  /opt/intel/licenses
    **  6.  /Users/Shared/Library/Application Support/Intel/Licenses
    **  7.  /projects/sems/install/rhel6-x86_64/sems/compiler/intel/17.0.1/base/compilers_and_libraries_2017.1.132/linux/bin/intel64/*.lic
    
    Please refer http://software.intel.com/sites/support/ for more information..
    
    icc: error #10052: could not checkout FLEXlm license
```

jhux2:
Are you confident that the additional ignored rows don’t affect the behavior of the rebalancing algorithm?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cgcgcg:
@jhux2 The rebalancing algorithm sees the same problem as before, what has changed is how we use the solution. The block size > 1 case never worked, so the alternative would have been to interpret it a block size = 1 matrix. You could imagine cases where this is quite far from the Zoltan2 solution. It probably would be better to get some option in Zoltan2 to achieve this behavior. @trilinos/zoltan2

kddevin:
@egboman and @kddevin discussed at today's zoltan meeting, to respond to an email from @rstumin 

Maybe we should all talk to understand what you are trying.

PR #7290, opened 2020-04-30
Teuchos: add timestamp option to verbose stacked timer

rppawlo:
Adding a timestamp option to the StackedTimer when verbose mode is enabled. The user can enable and export the number of levels to output with:

```
export TEUCHOS_ENABLE_VERBOSE_TIMESTAMP_LEVELS=2
```

This is going to be used for LDMS performance monitoring on HPC machines.

## Motivation

Performance monitoring for milestone runs

## Stakeholder Feedback

Planning to be used in ATDM L1 Milestone runs.
 
## Testing

New unit test was added to stacked_timer.cpp.


rppawlo:
@trilinos/framework - this PR is seeing repeated autotester internal failures. Could you help determine the issue?

rppawlo:
@etphipp - would you mind reviewing again? gcc 4 was missing a function from c++11 so had to add a workaround.

PR #7297, opened 2020-05-01
ATDM: Skip changing group and perms on nonowning install base dir (ATDV-241)

bartlettroscoe:
This avoids a nasty little install-time error due to two different user accounts trying to install under the same base directory as described in https://github.com/TriBITSPub/TriBITS/issues/314#issuecomment-622440398.  This should 100% complete [ATDV-241](https://sems-atlassian-srn.sandia.gov/browse/ATDV-241).

## How was this tested?

With very strong TriBITS tests. See https://github.com/TriBITSPub/TriBITS/commit/4b335ea765597db4f0bf3e77f7e0c90891635ef4.


bartlettroscoe:
I also put in a fix for `ATDM_CONFIG_TRIL_CMAKE_INSTALL_PREFIX` in 2da3908 that was breaking the 'cee-rhel6' install.

PR #7306, opened 2020-05-04
ifpack2: Reduce ILUT memory usage.

vbrunini:
By making sure to free L & U prior to recomputing them.

@trilinos/ifpack2

## Motivation
This brings ifpack2 ILUT memory usage in line with aztec on several Aria test cases, saving ~30% high-water mark.


vbrunini:
@csiefer2 @jhux2

PR #7315, opened 2020-05-05
tpetra: Include multivector length in deciding where kernels should run.

vbrunini:
Instead of just the current sync state. For long multivectors it is
faster to run on the device even if the data is currently on the host.
This reduces GPU solve times by ~30% on the Aria milestone test.

@trilinos/tpetra

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -329,9 +329,23 @@ namespace { // (anonymous)
     }
   }
 
+  template <class impl_scalar_type, class buffer_device_type>
+  bool
+  runKernelOnHost ( Kokkos::DualView<impl_scalar_type*, buffer_device_type> imports )
+  {
+    if (! imports.need_sync_device ()) {
+      return false; // most up-to-date on device
+    }
+    else { // most up-to-date on host
+      constexpr size_t localLengthThreshold = 10000;
+      return imports.extent(0) <= localLengthThreshold;
+    }
+  }
+
+
   template <class SC, class LO, class GO, class NT>
   bool
-  multiVectorRunNormOnHost (const ::Tpetra::MultiVector<SC, LO, GO, NT>& X)
+  runKernelOnHost (const ::Tpetra::MultiVector<SC, LO, GO, NT>& X)
   {
     if (! X.need_sync_device ()) {
       return false; // most up-to-date on device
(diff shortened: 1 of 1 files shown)


vbrunini:
@kddevin @tjfulle

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -329,9 +329,23 @@ namespace { // (anonymous)
     }
   }
 
+  template <class impl_scalar_type, class buffer_device_type>
+  bool
+  runKernelOnHost ( Kokkos::DualView<impl_scalar_type*, buffer_device_type> imports )
```
lucbv, line 334: Allowing the user to choose the localLengthThreshold maybe using a parameter list would be better for performance. It is not obvious that all GPUs will have the same threshold.
Also allowing to restore the current behavior using a second parameter seem useful for debugging and for folks who have tests that rely on gold files and such.

vbrunini: I agree that some way to control the threshold would be nice, but I don't think there's a reasonable way to do it with a parameter list. MultiVector doesn't get a parameter list provided anywhere right now, and it would be very inconvenient as a user to have to find everywhere we construct MultiVectors and provide a parameter list to change this.

Maybe another function in Tpetra::Details::Behavior that looks for an environment variable? Ultimately this is the kind of thing I'd like the Kokkos auto-tuning feature that is being worked on to control, but I think that's a ways off still.

kddevin:
@vbrunini Thanks for this contribution. @trilinos/tpetra will review.
Can you add a note about how you tested the changes (e.g., Trilinos Tpetra tests run on doom)?
We are trying to track this info to satisfy ASC software quality assessments.  Thanks.

kddevin:
Hi, @lucbv .  I see you added the "results impacting" flag.  Have you seen changes in answers due to this change?  Please comment. Thanks!

lucbv:
@kddevin  You can notice small changes due to summation being performed in different orders on host vs device. This can actually be a problem especially when using float.
In general I doubt this will be a big problem though.

kddevin:
@GeoffDanielson will add an environment variable to replace the hard-coded value.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
GeoffDanielson: I'm working on making this happen via Tpetra::Details::Behavior right now.

vbrunini:
Superceded by #7387

PR #7320, opened 2020-05-06
ATDM/ats1: Prioritize builds and disable running tests (ATDV-347)

e10harvey:
## How was this tested?
Please see [cdash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&begin=2020-05-07&end=2020-05-06&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=ats1).

### DIFF
--- a/cmake/ctest/drivers/atdm/ats1/local-driver-on-allocation.sh
+++ b/cmake/ctest/drivers/atdm/ats1/local-driver-on-allocation.sh
@@ -1,3 +1,7 @@
 #!/bin/bash
+if [[ "${CTEST_DO_TEST}" == "" ]] ; then
+  export CTEST_DO_TEST=OFF
+fi
+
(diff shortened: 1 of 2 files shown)


bartlettroscoe:
@e10harvey, I am wondering if the KNL parallel build level specified at:

https://github.com/trilinos/Trilinos/blob/bf072433803d58c0866a7b52e6c7ed2adabe5bf3/cmake/std/atdm/ats1/environment.sh#L61

or even for HSW at:

https://github.com/trilinos/Trilinos/blob/bf072433803d58c0866a7b52e6c7ed2adabe5bf3/cmake/std/atdm/ats1/environment.sh#L46

may be too high. 

Have we ever experimented with building from scratch on the compute nodes with these parallel build levels?  I fear that we have really only been testing rebuilds locally (where most of all of the object files are already built).

PR #7324, opened 2020-05-06
Fix CUDA defaulted function warnings (#7314)

### LINKED ISSUE #7314: All CUDA builds: many warnings about "host/device annotation ignored on explicitly defaulted fn"
## Question

On any CUDA build, in several packages, I see a lot of warnings like this:
```
/ascldap/users/bmkelle/PerfTesting/Trilinos/packages/sacado/src/KokkosExp_View_Fad_Contiguous.hpp(1030): warning: __host__ annotation is
ignored on a function("~ViewMapping") that is explicitly defaulted on its first declaration
```

The lines causing these look like ``KOKKOS_INLINE_FUNCTION type foo() = default;``. My question is, does nvcc automatically generate device versions of explicitly defaulted functions? If it does, we should try to remove the KOKKOS_INLINE_FUNCTION from these to stop the warnings.

brian-kelley:
Fix warnings in several packages that happen in CUDA 9/10 builds,
by using ``KOKKOS_DEFAULTED_FUNCTION`` instead of ``KOKKOS_INLINE_FUNCTION``
for explicitly defaulted functions (default constructors, copy constructors, destructors, and ``operator=``).

``KOKKOS_DEFAULTED_FUNCTION`` is defined depending on the CUDA version (< 10.0, it's the same as ``KOKKOS_INLINE_FUNCTION``,  but for >= 10.0, it's empty). This is because starting in CUDA 10, the compiler can generate ``__device__`` versions of defaulted functions as needed, and warns if ``__host__`` and/or ``__device__`` annotations are added explicitly. CUDA 9 still requires the annotations, but also warns if they are added, so Kokkos adds flags to repress the warnings in that case.

@alanw0 There are some changes to STK, so those would need to be applied to the real STK repo.

@trilinos/tpetra 
@trilinos/stokhos 
@trilinos/stk 
@trilinos/intrepid2 
@trilinos/trilinoscouplings 

## Motivation

Just fixes lots of warnings, for example:
```
warning: __host__ annotation is ignored on a function("~ViewMapping")
that is explicitly defaulted on its first declaration
```

## Related Issues

* Closes #7314
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing

Did an all-packages build on RHEL7, CUDA 10.1.

### DIFF
--- a/packages/intrepid2/src/Orientation/Intrepid2_Orientation.hpp
+++ b/packages/intrepid2/src/Orientation/Intrepid2_Orientation.hpp
@@ -117,7 +117,7 @@ namespace Intrepid2 {
     KOKKOS_INLINE_FUNCTION
     Orientation();
 
-    KOKKOS_INLINE_FUNCTION
+    KOKKOS_DEFAULTED_FUNCTION
     Orientation(const Orientation &b) = default;
 
     KOKKOS_INLINE_FUNCTION
--- a/packages/intrepid2/src/Shared/Intrepid2_ArrayToolsDefContractions.hpp
+++ b/packages/intrepid2/src/Shared/Intrepid2_ArrayToolsDefContractions.hpp
@@ -140,7 +140,7 @@ namespace Intrepid2 {
               const bool sumInto_) 
         : _outputFields(outputFields_), _inputData(inputData_), _inputFields(inputFields_), _sumInto(sumInto_) {}
 
-      KOKKOS_INLINE_FUNCTION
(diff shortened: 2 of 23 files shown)


etphipp:
Stokhos and Sacado changes look good to me.  Thanks @brian-kelley!

brian-kelley:
@prwolfe There were still tons of warnings in my local build too, but I think this PR squashes all the "host/device annotation on explicitly defaulted function" ones.

Issue #7331, opened 2020-05-07
SEACAS build error in ATDM Trilinos debug builds starting 2020-05-07


bartlettroscoe:
CC: @trilinos/seacas, @kddevin (Trilinos Data Services Product Lead), @gsjaardema 

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label "impacting: tests"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&date=2020-05-07&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-&field2=builderrors&compare2=43&value2=0&field3=subprojects&compare3=93&value3=SEACAS) there is a build error in SEACAS in the ATDM Trilinos 'debug' builds:

* `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg`
* `Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg_cuda-aware-mpi`
* `Trilinos-atdm-ats2-gnu-7.3.1-spmpi-2019.06.24_serial_static_dbg`
* `Trilinos-atdm-cts1-intel-19.0.5_openmpi-4.0.1_openmp_static_dbg`
* `Trilinos-atdm-sems-rhel6-gnu-7.2.0-openmp-debug`
* `Trilinos-atdm-sems-rhel7-intel-18.0.5-openmp-shared-debug`
* `Trilinos-atdm-tlcc2-intel-debug-openmp`
* `Trilinos-atdm-waterman-cuda-9.2-debug`
* `Trilinos-atdm-white-ride-cuda-9.2-gnu-7.2.0-debug`
[...]

Issue #5184, opened 2019-05-15
Need to change one of the Trilinos PR builds to a pure "debug" build


bartlettroscoe:
@trilinos/framework 

Currently, it seems that all of the Trilinos PR builds use `CMAKE_BUILD_TYPE=RELEASE`.  This allowed a defect to get added to the code that broke all of the ATDM debug builds (added in PR #5153).

To ensure this never happens again, we need at least one of the Trilinos PR builds to set `CMAKE_BUILD_TYPE=DEBUG`.

bartlettroscoe:
Got bit by this again (see #7331 and #7333).

@trilinos/framework, can we please make this happen?  Can I put in a PR build to switch one of the faster-building builds to be a full debug build?

PR #7333, opened 2020-05-07
SEACAS: Fix debug build

gsjaardema:
Assert in Conjoin was accessing a variable that had been removed in a recent refactor.  Change assert to use correct value.

Addresses #7331

@trilinos/seacas

### DIFF
--- a/packages/seacas/applications/conjoin/Conjoin.C
+++ b/packages/seacas/applications/conjoin/Conjoin.C
@@ -2866,7 +2866,7 @@ namespace {
     // copy values to master nodeset value information
     for (size_t j = 0; j < entity_count; j++) {
       size_t global_loc = local_set.elemOrderMap[j];
-      SMART_ASSERT(global_loc < glob_entity_count);
+      SMART_ASSERT(global_loc < global_values.size());
       global_values[global_loc] = values[j];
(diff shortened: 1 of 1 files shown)


gsjaardema:
That is what I was wondering also.  It should have failed on all debug builds.

prwolfe:
> That is what I was wondering also. It should have failed on all debug builds.

We don't do a debug build. The wait time is enough of an issue as things stand ;-)

gsjaardema:
That answers that question.  Would only fail on a debug build.

bartlettroscoe:
> We don't do a debug build. The wait time is enough of an issue as things stand ;-)

That is not really acceptable for protecting customers.  One of the faster builds needs to be converted to a full debug builds and then perhaps just run the `BASIC` test suite.

bartlettroscoe:
Note: #5184

bartlettroscoe:
For some reason, the PR tester never even started testing this PR in the 17 hours since it was opened.  What is the deal with that?

bartlettroscoe:
FYI: I manually deployed this to the 'atdm-nightly' branch so it will run tomorrow in the ATDM Trilinos nightly builds.

PR #7361, opened 2020-05-13
Merge in updated TriBITS snapshot (ATDV-303)

bartlettroscoe:
This is to make sure that going forward, we don't have any merge conflicts due to the separate commits in #6814 and TriBITSPub/TriBITS#300.

There is no diff in this PR but it is important to merge this so that future changes to these files don't cause a diff.


bartlettroscoe:
Thanks @rppawlo!

Issue #7329, opened 2020-05-07
SEACAS: Ioss::Field::INFORMATION appears to have been removed


ikalash:
Did the Ioss::Field::INFORMATION attribute in SEACAS go away?  We are getting failures in the Albany nightlies saying this is not recognized: 

https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=1482

Was the attribute replaced by something else?

@trilinos/seacas

ikalash:
I verified that indeed INFORMATION is no longer there in Ioss_Field.h.  It seems I can replace it with COMMUNICATION and still have all Albany tests pass.  Here is where the attribute is used in Albany: https://github.com/SNLComputation/Albany/blob/43aaef979b702933ccb1bea5616c6da166104ef1/src/disc/stk/Albany_GenericSTKFieldContainer_Def.hpp#L52 .  @gsjaardema , please let me know if the change I made doesn't make sense.

gsjaardema:
The `MESH_REDUCTION` would be a better replacement.  I'm not sure why I didn't just make `INFORMATION` an alias for `MESH_REDUCTION` so I wouldnt break anything...

ikalash:
Ah ok, this is good to know.  I'll change it in Albany - thanks!

PR #7362, opened 2020-05-13
IOSS: Fix backward compatibility

gsjaardema:
The INFORMATION field type was changed to `MESH_REDUCTION`, but that
 broke some existing users of IOSS.  Add back `INFORMATION` as alias
 of `MESH_REDUCTION`

@trilinos/seacas


gsjaardema:
Exposed by #7329.  Removes need for that workaround.

gsjaardema:
Nice to have a quick easy experience every once in a while.

Issue #4657, opened 2019-03-19
Belos: Make ICGS default orthogonalizer for other GMRES-like solvers


mhoemmen:
@trilinos/belos @hkthorn 

PR #4650 fixes #994, which changes `Belos::PseudoBlockGmresSolMgr`'s default orthogonalizer from DGKS to ICGS.  We also need to change the default orthogonalizers of the other GMRES-like solvers in Belos, such as HybridGmres and GCRODR (recycling GMRES), so that they match `PseudoBlockGmresSolMgr`'s default.

See comment here: 
https://github.com/trilinos/Trilinos/pull/4650#issuecomment-474438001

## Related Issues

* Follows #994, #4650

PR #7366, opened 2020-05-14
Belos: Change default orthogonalization to ICGS[2]

jennloe:
@trilinos/belos 

## Motivation
This PR makes ICGS[2] the default orthogonalization scheme for all solvers, to be consistent with the default behavior in `PseudoBlockGmres` and in AztecOO.  

I also removed some unused header files in `RCGSolMgr.hpp` related to orthogonalization. 

## Related Issues
This PR closes issue #4657. (See also PR #4650.)


jhux2:
@spdomin @sayerhs This could potentially affect Nalu and Nalu-Wind.

jennloe:
This PR has uncovered a bug in the ICGS and IMGS orthogonalization for a few obscure test cases.  @hkthorn is working to fix.  It will be a few more days before this is merged.

iyamazaki:
At least my Nalu-wind runs use ICGS. I am not sure if this is the only place we set up Belos, but it looks like LinearSolverConfig sets it to be ICGS?

sayerhs:
Thanks @jhux2 . As @iyamazaki mentions, nalu has been using ICGS as the default 

https://github.com/Exawind/nalu-wind/blob/master/src/LinearSolverConfig.C#L71-L72

jennloe:
Bug should be fixed now. See PR #7392

spdomin:
@jhux2, thank you for looking out for Nalu-based codes - much appreciated!

As noted above, this has been our default for a while as our intent was to benchmark Belos vs AztecOO.

Do you suggest that we remove the setting or expose this to the user?

jhux2:
> @jhux2, thank you for looking out for Nalu-based codes - much appreciated!
> 
> As noted above, this has been our default for a while as our intent was to benchmark Belos vs AztecOO.
> 
> Do you suggest that we remove the setting or expose this to the user?

I would think only a power user would want to toggle this.

PR #7368, opened 2020-05-14
Seacas parallel netcdf detect

gsjaardema:
@trilinos/seacas 

## Motivation

Clean up the determination of whether to build a parallel-aware exodus.  This requires two conditions:
* Parallel (MPI) build
* NetCDF library is parallel enabled.

It is possible to have a parallel (mph) build and a serial NetCDF build. In this case, the parallel-aware exodus is not built which disables the auto-decomp and auto-join capabilities (1->N and N->1) of exodus and only enables the file-per-processor mode.

The PR also unifies the testing of some variables


packages/seacas/libraries/ioss/src/main/CMakeLists.txt
```diff
@@ -160,7 +160,7 @@ IF (SEACASIoss_ENABLE_THREADSAFE)
     )
 
   SET(IOSHELL_ARG --in_type generated 10x10x10+shell:xXyYzZ+sideset:xXyY+times:2+variables:element,2,nodal,3,sideset,4)
-IF ($(TPL_ENABLE_MPI))
+IF (TPL_ENABLE_MPI)
```
bartlettroscoe, line 163: Yes, that is the correct way

packages/seacas/CMakeLists.txt
```diff
@@ -17,6 +17,30 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
...
+         file(STRINGS "${meta_path}/netcdf_meta.h" netcdf_par_string REGEX "NC_HAS_PARALLEL ")
+         string(REGEX MATCH "[01]" netcdf_par_val "${netcdf_par_string}")
+         if (netcdf_par_val EQUAL 1)
+            set(TPL_Netcdf_PARALLEL True CACHE BOOL "Is NetCDF library parallel enabled")
+         endif()
+      endif()
+      if (NOT TPL_Netcdf_PARALLEL)  
+         set(TPL_Netcdf_PARALLEL False CACHE BOOL "Is NetCDF library parallel enabled")
+      endif()
+   endif()
```
bartlettroscoe, line 40: I am not sure about this logic.  I think it could have bad behavior.  I think this fails if the variable `TPL_Netcdf_PARALLEL` is set to `FALSE` or `OFF` before it is called.  For example, if it is already set in the cache as `FALSE` or `OFF` before calling this code, then this code will **not** change the value of `TPL_Netcdf_PARALLEL`.

What are your assumptions about the nature of the variable `TPL_Netcdf_PARALLEL` before this code is called?

gsjaardema: This maybe should be `if (NOT DEFINED TPL_Netcdf_PARALLEL)` although there is also some argument that could be made that the value should always be set based on the setting in `netcdf_meta.h` since if the code includes that file, it may get conflicting values?

We can discuss on Skype if that works better

bartlettroscoe: > there is also some argument that could be made that the value should always be set based on the setting in netcdf_meta.h since if the code includes that file

Then the logic for setting `TPL_Netcdf_PARALLEL` should be in the file `FindTPLNetcdf.cmake` and not in this SEACAS file.  And we need to be careful about cache vs non-cache vars.

> We can discuss on Skype if that works better

A 5-min talk, I promise.

packages/seacas/CMakeLists.txt
```diff
@@ -17,6 +17,30 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
 
 set(${PROJECT_NAME}_C_Standard_DEFAULT c99)
 
+if (TPL_ENABLE_MPI)
+   if (NOT TPL_Netcdf_PARALLEL)
```
bartlettroscoe, line 21:
```
IF (NOT "${TPL_Netcdf_PARALLEL}" STREQUAL "")
   ...
   set(TPL_Netcdf_PARALLEL True CACHE INTERNAL "Is NetCDF library parallel enabled")
   ...
ENDIF()
```

packages/seacas/CMakeLists.txt
bartlettroscoe: As we discussed, please move this logic to the end if `FindTPLNetcdf.cmake` and change it to:

IF (NOT "${TPL_Netcdf_PARALLEL}" STREQUAL "")
   ...
   set(TPL_Netcdf_PARALLEL True CACHE INTERNAL "Is NetCDF library parallel enabled")
   ...
ENDIF()

You will also want to change the other `TPL_Netcdf_PARALLEL` set statement in `FindTPLNetcdf.cmake` type to `INTERNAL` instead of `BOOL` since the user should not be allowed to set it.

bartlettroscoe:
Once this merges to 'develop', we will use `git format-patch` and `git am` to get this merged over on the TriBITS side.

PR #7380, opened 2020-05-15
SEACAS: Fix parallel build with serial NetCDF issues

### LINKED ISSUE #7379: Seacas: new seacas build error in Albany nightlies
We have a new seacas build error in some of our Trilinos nightlies: 
```
Trilinos/packages/seacas/libraries/ioss/src/exodus/Ioex_IOFactory.C:121: undefined reference to `Ioex::ParallelDatabaseIO::ParallelDatabaseIO(Ioss::Region*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Ioss::DatabaseUsage, ompi_communicator_t*, Ioss::PropertyManager const&)'
```
https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=1803

I'm guessing it crept in with PR https://github.com/trilinos/Trilinos/pull/7368 ?  @gsjaardema , can you please have a look?

@trilinos/seacas

gsjaardema:
Use the correct symbol to detect a parallel (MPI) build but a serial NetCDF library.

Should fix #7379

@trilinos/seacas

### DIFF
--- a/packages/seacas/libraries/exodus/include/exodusII.h
+++ b/packages/seacas/libraries/exodus/include/exodusII.h
@@ -50,7 +50,6 @@
 #include "netcdf_meta.h"
 #if NC_HAS_PARALLEL
 #ifndef PARALLEL_AWARE_EXODUS
-#define PARALLEL_AWARE_EXODUS
 #endif
 #endif
 #endif
--- a/packages/seacas/libraries/ioss/src/exodus/Ioex_IOFactory.C
+++ b/packages/seacas/libraries/ioss/src/exodus/Ioex_IOFactory.C
@@ -33,7 +33,7 @@
 #include <exodus/Ioex_IOFactory.h> // for Ioex IOFactory
 #include <exodus/Ioex_DatabaseIO.h> // for Ioex DatabaseIO
 
-#if defined(SEACAS_HAVE_MPI)
+#if defined(PARALLEL_AWARE_EXODUS)
 #include <exodus/Ioex_ParallelDatabaseIO.h> // for Ioex ParallelDatabaseIO
 #endif
 #include <tokenize.h>
@@ -54,7 +54,7 @@ namespace Ioss {
   class DatabaseIO;
 } // namespace Ioss
 
-#if defined(SEACAS_HAVE_MPI)
+#if defined(PARALLEL_AWARE_EXODUS)
 namespace {
   std::string check_decomposition_property(const Ioss::PropertyManager &properties,
                                            Ioss::DatabaseUsage          db_usage);
@@ -76,7 +76,7 @@ namespace Ioex {
     Ioss::IOFactory::alias("exodus", "exodusii");
     Ioss::IOFactory::alias("exodus", "exodusII");
     Ioss::IOFactory::alias("exodus", "genesis");
-#if defined(SEACAS_HAVE_MPI)
+#if defined(PARALLEL_AWARE_EXODUS)
     Ioss::IOFactory::alias("exodus", "dof_exodus");
     Ioss::IOFactory::alias("exodus", "dof");
 #endif
@@ -86,7 +86,7 @@ namespace Ioex {
                                        MPI_Comm                     communicator,
                                        const Ioss::PropertyManager &properties) const
   {
-#if defined(SEACAS_HAVE_MPI)
+#if defined(PARALLEL_AWARE_EXODUS)
     // The "exodus" and "parallel_exodus" databases can both be accessed
     // from this factory.  The "parallel_exodus" is returned only if the following
     // are true:
@@ -142,7 +142,7 @@ namespace Ioex {
   }
 } // namespace Ioex
 
-#if defined(SEACAS_HAVE_MPI)
+#if defined(PARALLEL_AWARE_EXODUS)
 namespace {
   std::string check_decomposition_property(const Ioss::PropertyManager &properties,
                                            Ioss::DatabaseUsage          db_usage)


ikalash:
I should add, I think a unique feature of the builds where we're seeing the seacas errors (https://sems-cdash-son.sandia.gov/cdash/index.php?project=Albany) is that they have -D TPL_Netcdf_PARALLEL:BOOL=OFF.

gsjaardema:
I haven't tried the Albany build yet, but when I apply this patch and just build seacas (io_info), with serial NetCDF and parallel build, everything works correctly.  Will try to get some time to try on CEE lan with Albany.

bartlettroscoe:
> I should add, I think a unique feature of the builds where we're seeing the seacas errors (https://sems-cdash-son.sandia.gov/cdash/index.php?project=Albany) is that they have -D TPL_Netcdf_PARALLEL:BOOL=OFF.

@ikalash, @gsjaardema can comment me, but I believe as of the recently merged PRs, that the user can no longer sets the value of `TPL_Netcdf_PARALLEL`.  It will be set automatically by the updated FindTPLNetcdf.cmake module as shown in:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/tribits/common_tpls/FindTPLNetcdf.cmake#L78
* https://github.com/trilinos/Trilinos/blob/develop/cmake/tribits/common_tpls/FindTPLNetcdf.cmake#L123
* https://github.com/trilinos/Trilinos/blob/develop/cmake/tribits/common_tpls/FindTPLNetcdf.cmake#L128

So any value you try to set for `TPL_Netcdf_PARALLEL` will be overwritten in the cache.

bartlettroscoe:
@gsjaardema, is the FindTPLNetcdf.cmake module exactly as it appears in Trilinos 'develop' as shown here:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/tribits/common_tpls/FindTPLNetcdf.cmake

exactly what should be merged back to TriBITS 'master'?

ikalash:
> @ikalash, @gsjaardema can correct me, but I believe as of the recently merged PRs, that the user can no longer sets the value of `TPL_Netcdf_PARALLEL`. It will be set automatically by the updated FindTPLNetcdf.cmake module as shown in:
> 
> * https://github.com/trilinos/Trilinos/blob/develop/cmake/tribits/common_tpls/FindTPLNetcdf.cmake#L78
> * https://github.com/trilinos/Trilinos/blob/develop/cmake/tribits/common_tpls/FindTPLNetcdf.cmake#L123
> * https://github.com/trilinos/Trilinos/blob/develop/cmake/tribits/common_tpls/FindTPLNetcdf.cmake#L128
> 
> So any value you try to set for `TPL_Netcdf_PARALLEL` will be overwritten in the cache.

I think this is not happening b/c I see different behavior in the same build when I set TPL_Netcdf_PARALLEL=ON vs OFF.

bartlettroscoe:
> I think this is not happening b/c I see different behavior in the same build when I set TPL_Netcdf_PARALLEL=ON vs OFF.

@ikalash, what is the most recent exact version of Trilinos you are seeing different behavior?  Can you reproduce this with the SEMS TPLs, for example, so that we can reproduce this?

ikalash:
Yes,. The modules to load are here: https://github.com/SNLComputation/Albany/blob/master/doc/nightlyTestHarness/mockba/modules_mockba.sh and the configure script here: https://github.com/SNLComputation/Albany/blob/master/doc/nightlyTestHarness/mockba/do-cmake-trilinos-mpi-mockba . You may have to change the blas/lapack paths, but other than that, the script should work out of the box on a RHEL7 machine with sems modules.

gsjaardema:
The problem is that Albany is using a parallel-aware NetCDF build (`module load sems-netcdf/4.4.1/exo_parallel`).  The CMake configuration specifies that NetCDF is not parallel (`-D TPL_Netcdf_PARALLEL:BOOL=OFF`)

This is a conflict since the Exodus library parses a NetCDF include file which detects that the NetCDF library is parallel, but the CMake configuration says the opposite.

Will revist this PR to make sure that we have consistency...

gsjaardema:
Which is the preferred behavior if the `TPL_Netcdf_PARALLEL` setting does not match the NetCDF library build:

- Set the variable based on the library configuration and just continue with the configuration
- Same as above, write a STATUS saying value was changed
- Same as above, but issue a WARNING
- Same as above, but issue an ERROR

ikalash:
So in the case of the scrips I sent, I agree that netcdf parallel can be enables because I’m loading a module that has it.  On so,e of our fedora machines, this is not possible because there’s not netcdf.  @lxmota can give you more Info on those builds.  Basically we do need to be able to disable netcdf at compile time for builds where there is no parallel netcdf available

bartlettroscoe:
>  Basically we do need to be able to disable netcdf at compile time for builds where there is no parallel netcdf available.

@ikalash, then can't you just disable Netcdf on these systems with `-DTPL_ENABLE_Netcdf=OFF` and just leave `TPL_Netcdf_PARALLEL` alone?  Why ever set  `TPL_Netcdf_PARALLEL`?

gsjaardema:
You can disable NetCDF with the current situation.  However, you will not get any exodus output at all in that case.  I think what I am saying is that you should not set the 'TPL_Netcdf_PARALLEL' and let CMake decide the correct setting.  You will get parallel or serial NetCDF set automatically and that should be correct.  For the case where there is not NetCDF, you use the 'TPL_Enable_Netcdf' setting.

gsjaardema:
If possible, it would be good to change to use the more current NetCDF SEMS modules (4.7.3) and also select the 'base' module if you don't want Parallel NetCDF or the 'parallel' module if you do want Parallel NetCDF.   We can talk in a different forum later to explain this better what you are trying to do.

ikalash:
@bartlettroscoe : we cannot turn off Netcdf b/c our runs require Netcdf. Just, some machines don't have parallel netcdf, and we need to be able to tell the code that.  In Albany, the only thing parallel netcdf is needed for is to decompose a mesh inside the code.  Only a small subset of tests use that capability, and they are turned off if parallel netcdf is not enabled.  @lxmota can provide more information, as his machines are the ones w/o parallel netcdf.  I believe when he tried removing the option to enable/disable parallel netcdf, he got an error that netcdf_par.h was not found.  @lxmota can expand on this.

lxmota:
Removing `TPL_Netcdf_PARALLEL` eliminated the compilation errors from a missing `netcdf_par.h`, but now we get many link errors like this in some of our Albany test beds:

```
/usr/bin/ld: ../exodus/libIoex.so.12.17: undefined reference to `Ioex::ParallelDatabaseIO::ParallelDatabaseIO(Ioss::Region*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Ioss::DatabaseUsage, ompi_communicator_t*, Ioss::PropertyManager const&)'
```

This is from trying to compile Trilinos, before going into configuring and compiling Albany.

ikalash:
@lxmota: Greg's PR should resolve those errors.

glhenni:
It looks like this failed due to a timeout, not something related to the code changes. Perhaps due to some hardware issues over the weekend? Can someone explain why it hasn't been resubmitted? Does someone have to manually resubmit? 

This is holding up our Trilinos fork update for charon and I'm starting to get concerned.

rppawlo:
@glhenni - there was a scheduled power outage that started friday morning. The cuda testing machine just came back up today. There is a long que of PRs waiting to be retested. This just PR restarted  a few minutes ago and should be merged today.

PR #7390, opened 2020-05-18
amesos2: Fix build ifdef protection of Amesos2_cuSolver.hpp include.

vbrunini:
Fixes build errors when using an installed Trilinos that defined
HAVE_AMESOS2_CUSOLVER but not HAVE_AMESOS2_CUSPARSE.

@srajama1 @lucbv

### DIFF
--- a/packages/amesos2/src/Amesos2_Factory.hpp
+++ b/packages/amesos2/src/Amesos2_Factory.hpp
@@ -136,7 +136,7 @@
 #include "Amesos2_Cholmod.hpp"
 #endif
 
-#ifdef HAVE_AMESOS2_CUSOLVER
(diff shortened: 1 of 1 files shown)


lucbv:
@srajama1 do you know if the auto-tester builds are turning on TPLs like cublas, cusparse and cusolver?
At one point it will probably make sense to test these things if customers rely on them...

srajama1:
> Thanks @vbrunini ! CUSOLVER was built on CUSPARSE. May be we just need to automatically enable CUSPARSE when CUSOLVER is enabled. This is still a good fix for now.

This should be easy to add in a CUDA build. @trilinos/framework : Can we enable CUSPARSE and CUSOLVER in a CUDA build (not all of them).

MicheldeMessieres:
@vbrunini @srajama1 I'm going to go ahead and merge this. Assuming that is ok.

PR #7391, opened 2020-05-18
Epetra: zero out nonlocal matrix during GlobalAssemble, plus override Print method

### LINKED ISSUE #7349: Epetra:  No way to zero out non local matrix in Epetra_FECrsMatrix
## Enhancement
@trilinos/epetra 

When someone uses a matrix multiple times, before starting the fill, it is common to zero-out the matrix. However, when using the FECrsMatrix version with nonlocal matrix enabled, there is no way to ensure the nonlocal entries are wiped before starting the fill. If the user does _not_ request a nonlocal matrix, then things are fine, since the nonlocal entries are wiped out at every call to GlobalAssemble.

Possible solutions:
1) make PutScalar in Epetra_CrsMatrix virtual, and override it in Epetra_FECrsMatrix to also zero out the nonlocal matrix (if present). Pro: cleaner. Con: need to _slightly_ change interface of base class (make the method virtual).
2) call PutScalar(0) on the nonlocal matrix (if present) at the end of GlobalAssemble. Pro: makes behavior similar to the case where nonlocal matrix is not used. Con: can't think of any. Maybe the user does not want to zero out the nonlocal matrix?

bartgol:
@trilinos/epetra 

## Motivation

This PR addresses issue #7349 . The behavior of GlobalAssemble in the cases where nonlocal matrix is used was slightly different from the case where it wasn't used. In particular, the nonlocal matrix was not zeroed out at the end. If the matrix was then reused later on, old entries were then going to pollute the matrix. Note that calling PutScalar(0.0) on the matrix has no effect on nonlocal entries, since there's no overridde inside the FECrs matrix class.

## Related Issues

* Closes #7349 

## Testing

There are no tests for FECrs matrix with more than 1 rank. I am not sure why that is, but the changes have no effect in case of single rank. Is it ok to run the FECrsMatrix test with 2 ranks?
 
## Additional Information
I also added an override of the Print method, which also prints nonlocal rows. This helped me identifying what the issue was in my application. Otherwise, there was no way for me to access the nonlocal rows information, and make sure it was consistent with what I expected. This print method, made me realize it wasn't.

### DIFF
--- a/packages/epetra/src/Epetra_FECrsMatrix.cpp
+++ b/packages/epetra/src/Epetra_FECrsMatrix.cpp
@@ -1009,29 +1009,32 @@ int Epetra_FECrsMatrix::GlobalAssemble(const Epetra_Map& domain_map,
       const Epetra_CrsGraph& graph = tempMat_->Graph();
       Epetra_CrsGraph& nonconst_graph = const_cast<Epetra_CrsGraph&>(graph);
       nonconst_graph.SetIndicesAreGlobal(true);
-	}
-	}
-      //Now we need to call FillComplete on our temp matrix. We need to
-      //pass a DomainMap and RangeMap, which are not the same as the RowMap
-      //and ColMap that we constructed the matrix with.
-      EPETRA_CHK_ERR(tempMat_->FillComplete(domain_map, range_map));
-
-    if (exporter_ == NULL)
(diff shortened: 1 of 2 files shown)


bartgol:
@cgcgcg requesting changes prevents the AT from launching the tests.

PR #7397, opened 2020-05-19
Tempus: Add TimeEvent Object

### LINKED ISSUE #7396: Tempus:  Add TimeEvent Object
## Enhancement
@trilinos/tempus 

Add the ability to trigger temporal events, e.g., 

1. Solution output
2. Diagnostic output
3. Restart
4. Screen dump
5. Debug
6. In-situ visualization
7. User-specified (i.e., observer)
8. Others?

Ability to specify temporal “events” via
- Time and/or time step index
- Start and stop times and/or indices
- Interval and/or number of “events”
- List of times and/or indices

Specify when to trigger “event”
 - Exactly at the specified time, i.e., adjust timestep to hit output time
 - If within timestep, output the following timestep, i.e., do not adjust timestep.

Additional details are on the slides from TimeEvent meeting (Ask @ccober6 for a copy.).

ccober6:
Add the ability to trigger temporal events, e.g.,

    Solution output
    Diagnostic output
    Restart
    Screen dump
    Debug
    In-situ visualization
    User-specified (i.e., observer)
    Others?

Ability to specify temporal “events” via

    Time and/or time step index
    Start and stop times and/or indices
    Interval and/or number of “events”
    List of times and/or indices

Specify when to trigger “event”

    Exactly at the specified time, i.e., adjust timestep to hit output time
    If within timestep, output the following timestep, i.e., do not adjust timestep.

Added unit tests to cover, the options.

@trilinos/tempus 

## Related Issues

* Closes #7396 

## Testing

All tests pass on my Mac.

### DIFF
--- a/packages/tempus/src/Tempus_TimeEventBase.hpp
+++ b/packages/tempus/src/Tempus_TimeEventBase.hpp
@@ -1,0 +1,118 @@
+// @HEADER
+// ****************************************************************************
+//                Tempus: Copyright (2017) Sandia Corporation
+//
+// Distributed under BSD 3-clause license (See accompanying file Copyright.txt)
+// ****************************************************************************
+// @HEADER
+
+#ifndef Tempus_TimeEventBase_decl_hpp
+#define Tempus_TimeEventBase_decl_hpp
+
+// Teuchos
+#include "Teuchos_Time.hpp"
+#include "Teuchos_VerboseObject.hpp"
+
+
+namespace Tempus {
+
+
+/** \brief This class defines time events which can be used to "trigger" an action.
+ *  Time events are points in time and/or timestep index where an
+ *  an action should occur, such as, solution output (mesh and solution)
+ *  diagnostic output, restart, screen dump, in-situ visualization,
+ *  user-specified, or any other action.
+ *
+ *  This class will store a collection time events, so that an object
+ *  may quiry it and take appropriate action.  Time events (time and
+ *  timestep index) can be specified via
+ *    - start, stop and stride
+ *    - list of events
+ *
+ */
+template<class Scalar>
+class TimeEventBase
+{
+public:
+
+  /// Constructor
+  TimeEventBase()
+    : name_("TimeEventBase"),
+      defaultTime_ (-std::numeric_limits<Scalar>::max()*1.0e-16),
+      defaultTol_  ( std::numeric_limits<Scalar>::min()),
+      defaultIndex_( std::numeric_limits<int>::min())
+  {}
+
+  /// Destructor
+  virtual ~TimeEventBase() {}
+
+  /// \name Basic methods
+  //@{
+    /// Test if time is near a TimeEvent (within tolerance).
+    virtual bool isTime(Scalar time) const
+    { return false; }
+
+    virtual Scalar getAbsTol() const
+    { return defaultTol_; }
+
+    /// How much time until the next event. Negative indicating the last event is in the past.
+    virtual Scalar timeToNextEvent(Scalar time) const
+    { return defaultTime_; }
+
+    /// Time of the next event. Negative indicating the last event is in the past.
+    virtual Scalar timeOfNextEvent(Scalar time) const
+    { return defaultTime_; }
+
+    /// Test if an event occurs within the time range.
+    virtual bool eventInRange(Scalar time1, Scalar time2) const
+    { return false; }
+
+    /// Test if index is a time event.
+    virtual bool isIndex(int index) const
+    { return false; }
+
+    /// How many indices until the next event. Negative indicating the last event is in the past.
+    virtual int indexToNextEvent(int index) const
+    { return defaultIndex_; }
+
+    /// Index of the next event. Negative indicating the last event is in the past.
+    virtual int indexOfNextEvent(int index) const
+    { return defaultIndex_; }
+
+    /// Test if an event occurs within the time range.
+    virtual bool eventInRangeIndex(int index1, int index2) const
+    { return false; }
+
+    /// Describe member data.
+    virtual void describe() const
+    {
+      Teuchos::RCP<Teuchos::FancyOStream> out =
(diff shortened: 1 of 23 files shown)


packages/tempus/src/Tempus_TimeEventBase.hpp
```diff
@@ -0,0 +1,118 @@
...
+    virtual bool eventInRangeIndex(int index1, int index2) const
+    { return false; }
+
+    /// Describe member data.
+    virtual void describe() const
+    {
+      Teuchos::RCP<Teuchos::FancyOStream> out =
+        Teuchos::VerboseObjectBase::getDefaultOStream();
+      *out << "TimeEventBase name = " << getName() << std::endl;
+    }
```
sconde, line 92: Does this make it a describable? I don't see it inheriting from `Teuchos::Describable`.

packages/tempus/src/Tempus_TimeEventComposite.hpp
```diff
@@ -0,0 +1,207 @@
...
+    /// Test if time is near a TimeEvent (within tolerance).
+    virtual bool isTime(Scalar time) const
+    {
+      bool is_time = false;
+      for(auto& e : timeEvents_) {
+        if (e->isTime(time)) {
+          is_time = true;
+          break;
+        }
+      }
```
sconde, line 49: This is only checking if `any` and not `all`?

sconde:
Should all time event index/time be an unsigned int/scalar?

packages/tempus/src/Tempus_TimeEventRangeIndex_decl.hpp
```diff
@@ -0,0 +1,91 @@
...
+    virtual void setNumEvents();
+  //@}
+
+
+protected:
+
+  int start_;
+  int stop_;
+  int stride_;
+  int numEvents_;
```
sconde, line 84: unsigned int?

packages/tempus/src/Tempus_TimeEventBase.hpp
ccober6: I did not make it inherit from Teuchos::Describable, but used the same name.  I do not like describe having any arguments.

packages/tempus/src/Tempus_TimeEventComposite.hpp
ccober6: Yeah, for the composite it should only need to match one of the TimeEvents to return true.  At some point it might be nice to know which TimeEvent matches, but for now we just need that it matches at least one TimeEvent.

packages/tempus/src/Tempus_TimeEventRangeIndex_decl.hpp
ccober6: start_ and stop_ could be signed, though stride_ and numEvents_ could be unsigned.  I will make the change for stride_ and numEvents_.

ccober6:
> Should all time event index/time be an unsigned int/scalar?

I have gone with allowing negative index/time values.  I did not see a reason to not allow them, as at some point someone may need to simulate -t_i to zero, and then simulate zero to t_f.

Thoughts?  Do you have a reason not to allow negative index/time?

packages/tempus/src/Tempus_TimeEventRangeIndex_decl.hpp
ccober6: Oh actually, we may want to allow a negative stride_.  For adjoint solves, the user may want to "begin" at stop_ and "end" at start_ with a negative stride_, but numEvents_ would still need to be non-negative.  Thoughts?

packages/tempus/unit_test/Tempus_UnitTest_TimeEvent.cpp
```diff
@@ -0,0 +1,1080 @@
...
+namespace Tempus_Unit_Test {
+
+using Teuchos::RCP;
+using Teuchos::rcp;
+using Teuchos::rcp_const_cast;
+using Teuchos::rcp_dynamic_cast;
+
+
+// ************************************************************
+// ************************************************************
```
sconde, line 36: can we start adding a more descriptive comment on what is being tested?

packages/tempus/unit_test/Tempus_UnitTest_TimeEvent.cpp
```diff
@@ -0,0 +1,1080 @@
...
+  te->addIndex(3);
+  TEST_COMPARE(te->getIndexList().size(), ==, 5);
+  te->addIndex(1);
+  TEST_COMPARE(te->getIndexList().size(), ==, 6);
+}
+
+
+
+// ************************************************************
+// ************************************************************
```
sconde, line 767: Also to be consistent with other tests, lets add the macros to enable individual tests. See ERK test.

packages/tempus/unit_test/Tempus_UnitTest_TimeEvent.cpp
ccober6: Yeah, I agree.  I just got tired of making unit tests.  I will go through break up these tests into smaller focused tests with a more descriptive comment/name.

packages/tempus/unit_test/Tempus_UnitTest_TimeEvent.cpp
ccober6: So the macro defines I think you are talking about, I have been slowly removing them.  They maybe all gone now.  I did this because Roger mentioned that we can get the same functionality via the "--test", e.g., `Tempus_UnitTest_TimeEvent.exe --test="Construction"`.  The only downside here is having better and more unique test names.  See above comment.

sconde: learned something new today

ccober6:
@sconde Ok, I just rebased with your suggested changes.  Can you re-review and approve?

Issue #965, opened 2017-01-03
Belos solver constructors ignore orthomanager paramerlist options


jjellio:
Currently, solvers that need an orthogonalizer call an orthomanager's constructor with only a label argument. (called from the SolMgr's setParameter routine) This allows one to change the type of orthogonalization, but never configure it. There are no methods for setting or obtaining the solver's orthogonalizer through the SolMgr interface, hence there is no way to control orthogonalization parameters even though they are supported.

A solution consistent with Trilinos' current packages would be to allow solvers to accept an optional parameter sublist called 'Orthogonalziation' that would contain optional parameters for the underlying orthogonalizer. Belos SolMgr implementations would then need to pass this list (if it exists) to the orthomanager constructors.

This is a bit more nuanced than simply adding the above logic, because ortho parameters have leaked into certain SolMgr implementations (GMRES/BlockGMRES, etc..) A small refactor would be required, forcing the solMgr to query the orthomanager instance for necessary parameters at the solver level. This will result in some modifications to certain SolMgr constructors.

@trilinos/belos

mhoemmen:
Belos::PseudoBlockGmresSolMgr takes an "Orthogonalization Constant" parameter.  However, it only delivers that parameter to its OrthoManager if the OrthoManager is "DGKS".

We have nested ParameterLists now (didn't 10+ years ago, if I remember right), so we could add an "Orthogonalization Parameters" sublist that goes straight into the OrthoManager.  @hkthorn , would that make sense?

hkthorn:
Yes, that is because the "Orthogonalization Constant" is only pertinent to DGKS, it is used to determine whether another step of classical Gram Schmidt is necessary.  The other methods have a static number of iterations they perform (ICGS, IMGS perform 2).  Each solver should tell you which orthogonalization routine they are using, which is stored in the internal parameter list that can be accessed using getCurrentParameters().  I see that the solvers are not storing that information in the internal parameter list ... hmm that used to be stored in there.  That can be fixed easily.

I'm not sure what the nested parameter list would be for, are the orthogonalization routines going to be more parametrized than they are now?

I've been watching the traffic, is there a logical issue where an orthogonalization type is being requested through the parameter list and ignored?  If it is just the case that the "Orthogonalization Constant" is ignored, except in the case of DGKS, that can be understood now.

What am I missing?

Thanks,
Heidi

--

Heidi K. Thornquist
Electrical Models & Simulation
Sandia National Laboratories
[...]

PR #7400, opened 2020-05-20
Belos: Have OrthoManager inherit from Teuchos::ParameterListAcceptorDefaultBase

jennloe:
This PR makes the Belos::OrthoManager class inherit from Teuchos::ParameterListAcceptorDefaultBase instead of each of the OrthoManager subclasses inheriting from it individually.  This step will allow the Belos solvers to set parameters in any instance of OrthoManager.  

@trilinos/belos 

## Testing
Expanded Epetra OthoManager test to use the get and set functions.


jhux2:
@jennloe Is this intended to address #965?  Just curious.

jennloe:
@jhux2 Yes, this is one of the first steps in that direction.  Several more pieces to go before it will all be working.

etphipp:
Is @trilinos/framework aware of the ride issues?  It looks like the tests were moved to the pascal nodes, but the build configuration wasn't changed to use the appropriate Kokkos-Cuda arch flag for pascal instead of kepler.

jhux2:
> Is @trilinos/framework aware of the ride issues? It looks like the tests were moved to the pascal nodes, but the build configuration wasn't changed to use the appropriate Kokkos-Cuda arch flag for pascal instead of kepler.

Yes, see #7498.

Issue #7274, opened 2020-04-28
Amesos2: suspected memory leak in classic basker


maartenarnst:
## Bug Report
@trilinos/amesos2 
@srajama1

### Description
I use Amesos2's classic basker as the linear solver in a Newton-type nonlinear solver.  I notice that as the Newton-type nonlinear solver goes through more and more iterations and classic basker is thus called more and more times to solve a linear system, the memory used by the program keeps increasing.  Ultimately, my code starts swapping and gets stuck.  The issue disappears when I use 'klu2' in place of 'basker.'

I suspect a memory leak related to the allocation of two arrays in dynamic memory, which do not appear to be freed later in the code.  This issue may have something to do with changes that appear to have been made in classic basker's memory management a while ago.  The code freeing these arrays was commented out, but it appears that it was not replaced with new code.

Specifically, in the 'develop' branch, in amesos2/src/basker/basker_def.hpp, on lines 253 and 255, two arrays 'tptr = new ... ' and 'X = new ... ' are allocated in dynamic memory.  On lines 603 and 604, there are now comments '//BASKERFREE(X)' and '//BASKERFREE(tptr)' but these lines do not appear to have been replaced with delete calls.

I replaced on line 603 '//BASKERFREE(X)' with 'delete [] X;' and on line 604 '//BASKERFREE(tptr)' with 'delete [] color;'  I used 'delete [] color;' instead of 'delete [] tptr;' because the basker code copies the pointer 'tptr' to the pointer 'color' and modifies 'tptr'.  In my code, the issue with the increasing memory consumption is solved after making this change.

I have not run any code with valgrind or any other tool to diagnose the memory use.

I remain available for any further information that may be helpful.  Thank you very much in advance!

jhux2:
@trilinos/amesos2

ndellingwood:
PR #7404 issued with the desired fix (along with some unit test fixes). Thanks @maartenarnst for reporting and testing the fix beforehand!

srajama1:
@maartenarnst : Sorry for missing this issue. The github volume has become too much to manage. This should be a pretty easy to fix. I am not sure why these got commented out. Let me check.

ndellingwood:
@srajama1 I put in a PR with the fix #7404, ready for review

srajama1:
@ndellingwood : Thanks for the PR ! I just approved it

@maartenarnst : Thanks for the fix ! This is just an embarrassing bug and I worry how it went without getting caught for years.

[...]

PR #2218, opened 2018-02-06
Feature shylubasker interface

ndellingwood:
Work toward adding ShyLUBasker interface to Amesos2; both versions
of Basker can now be enabled simultaneously

Add explicit Amesos2::ShyLUBasker interface and remove experimental workaround
requiring use of flag SHYLU_NODEBASKER to CMake flags

Use of ShyLUBasker in Amesos2 will require the following:
1. Trilinos_ENABLE_ShyLU_NodeBasker:BOOL=ON
2. Amesos2_ENABLE_ShyLU_NodeBasker:BOOL=ON
3. CMAKE_CXX_FLAGS:STRING="-DSHYLU_NODEBASKER" is no longer needed

Use of Basker in Amesos2 requires no changes.

@trilinos/amesos2 

## Description

Add explicit Amesos2::ShyLUBasker interface and remove experimental workaround
requiring use of flag SHYLU_NODEBASKER to CMake flags

Use of ShyLUBasker in Amesos2 will require the following:
1. Trilinos_ENABLE_ShyLU_NodeBasker:BOOL=ON
2. Amesos2_ENABLE_ShyLU_NodeBasker:BOOL=ON
3. CMAKE_CXX_FLAGS:STRING="-DSHYLU_NODEBASKER" is no longer needed

Use of Basker in Amesos2 requires no changes.

Amesos2::Basker: Renamed Basker class to BaskerClassic

Amesos2::Basker: Internal amesos2/src/basker code modified so that Basker class renamed to BaskerClassic, changes unseen by Amesos2 users. 

## Motivation and Context

CMake updates to ShyLU Basker address issue #2075 

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #2075 
* Part of 
* Composed of 

## How Has This Been Tested?

Tested locally with gcc/5.4, scotch 6.0.3, OpenMP, with each combo of MPI ON/OFF and Complex ON/OFF. 
Auto-tester will add testing coverage. 

## Checklist

- [x] My commit messages mention the appropriate GitHub issue numbers.
- [x] My code follows the code style of the affected package(s).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the [code contribution guidelines](../blob/master/CONTRIBUTING.md) for this project.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] No new compiler warnings were introduced.
- [x] These changes break backwards compatibility.

[...]

PR #7404, opened 2020-05-21
Issue 7274

ndellingwood:
@trilinos/amesos2 

## Motivation
Address issue #7274, fix tests and examples not tested by default

## Testing

Tested locally on Mac laptop, confirmed tests pass with changes.


ndellingwood:
@srajama1 
The commented delete lines occurred in PR in SHA 688b435e581 but I don't see an associated PR, so that must have occurred prior to Github pull requests.

The last PR they were touched was PR #2218, I had run a `sed` script that changed all the `FREE` calls `BASKERFREE` in basker directory, though the lines had been commented out prior to that in the SHA above.

srajama1:
It is very strange that this bug didn't get caught so long. Thanks for fixing it and the cleanup !

Issue #7122, opened 2020-04-03
Random tests failures on 'ats2' 'vortex' with no STDOUT/STDERR and return code 255


bartlettroscoe:
CC: @e10harvey 

I have been noticing random test failures on the 'ats2' 'vortex' builds where the test shows no output and the return value is 255.  For example, must today in the build:

* ` Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg_cuda-aware-mpi`

we saw the test failures:

Site | Build Name | Test Name | Status | Details | Consec­utive Non-pass Days | Non-pass Last 30 Days | Pass Last 30 Days | Issue Tracker
-- | -- | -- | -- | -- | -- | -- | -- | --
vortex | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_­static_­dbg_­cuda-aware-mpi | Sacado_­trad_­sfc_­example_­MPI_­1 | Failed | Completed   (Failed) | 1 | 3 | 20 |  
vortex | Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_­static_­dbg_­cuda-aware-mpi | Sacado_­tradoptest_­34_­EQA_­MPI_­1 | Failed | Completed   (Failed) | 1 | 3 | 20 |  

These show output like:

```
BEFORE: jsrun  '-p' '1' '--rs_per_socket' '4' '/vscratch1/jenkins/vortex-slave/workspace/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg/SRC_AND_BUILD/BUILD/packages/sacado/example/Sacado_trad_sfc_example.exe' '-v'
AFTER: export TPETRA_ASSUME_CUDA_AWARE_MPI=1; jsrun  '-M -disable_gpu_hooks' '-p' '1' '--rs_per_socket' '4' '/vscratch1/jenkins/vortex-slave/workspace/Trilinos-atdm-ats2-cuda-10.1.243-gnu-7.3.1-spmpi-2019.06.24_static_dbg/SRC_AND_BUILD/BUILD/packages/sacado/example/Sacado_trad_sfc_example.exe' '-v'
out_file=2c5e914b8db237218486ae3661bf193a.out
```

and

```
[...]

PR #7406, opened 2020-05-21
ATDM/ats2: Disable hack around error message

e10harvey:
This retry logic did not generated unique .out file names since some "Command Lines" are exactly the same but post to different ctest "Test" names. To make the retry logic work, we need the ctest "Test" name in the environment in order to generate a uniqe .out file name. This change disables the retry logic and may result in the mass Errors we were trying to work around. The retry code has been commented out but left there in case we need it again in the future.

## How was this tested?
See atdv-351.

### DIFF
--- a/cmake/std/atdm/ats2/trilinos_jsrun
+++ b/cmake/std/atdm/ats2/trilinos_jsrun
@@ -40,21 +40,22 @@ function debug_print {
 }
 
 ################################################################################
-# @brief: Hack around JSM server error "Error: Remote JSM server is not
-# responding"
+# @brief: Evaluate the jsrun command and return the jsrun return value.
 # This function assumes that "args" has been initialized
-# @return void
+# @return the jsrun return value
 ################################################################################
 function evaluate_jsrun_command {
   local jsrun_ret
-  local out_file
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
@e10harvey, I think we will need to do some pretty significant offline testing before we merge a PR lke this because it is possible it could cause mass test failures.  But I have seen strange random errors in some of the tests that might be explained by the temp `*.out` file not being unique so it will be good to test this out and get it merged.  At the same time, I would also like to consider adding logic to highlight the exit code 255 failures outlined in #7122.  (For those changes, I will create a PR against this PR branch for you to review).

Thanks again!

bartlettroscoe:
FYI: Will be merged as part of larger PR #7427.

bartlettroscoe:
@e10harvey, see, GitHub automatically shows this PR as merged once PR #7427 was merged!

bartlettroscoe:
And look the label `AT: STALE` is still applied, even through the PR got merged.  This seems like hole in the autotester use cases.  This is kind of confusing to anyone looking at these GitHub merged PRs.

PR #7410, opened 2020-05-21
Sacado:  Make the new Fad design the default.

etphipp:
Let the mayhem begin!

### DIFF
--- a/packages/sacado/CMakeLists.txt
+++ b/packages/sacado/CMakeLists.txt
@@ -28,7 +28,7 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
   ${PACKAGE_NAME}_NEW_FAD_DESIGN_IS_DEFAULT
   SACADO_NEW_FAD_DESIGN_IS_DEFAULT
   "Make the new Fad design the default, replacing the old one."
-  FALSE
+  ON
   )
 
 TRIBITS_ADD_OPTION_AND_DEFINE(
--- a/packages/trilinoscouplings/examples/scaling/TrilinosCouplings_EpetraIntrepidPoissonExample.cpp
+++ b/packages/trilinoscouplings/examples/scaling/TrilinosCouplings_EpetraIntrepidPoissonExample.cpp
@@ -1433,7 +1433,7 @@ sourceTerm (Scalar& x, Scalar& y, Scalar& z)
 {
   Scalar u;
   Scalar grad_u[3];
-  Scalar flux[3];
+  Scalar flux[3] = {0.0, 0.0, 0.0};
   Scalar material[3][3];
   Scalar f = 0.;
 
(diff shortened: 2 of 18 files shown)


mayrmt:
@etphipp @rppawlo I'm wondering: Is there a place, where I can find more details about the "new design"? Is it necessary for users to know about it or is this just an internal design change?

etphipp:
> @etphipp @rppawlo I'm wondering: Is there a place, where I can find more details about the "new design"? Is it necessary for users to know about it or is this just an internal design change?

It's just an internal design change.  The user API does not change.  The major benefit of it is it separates the expression template implementation from the Fad classes allowing for easier customization and specialization.  It will be most useful for optimizing nested Fad and Fad/Simd use-cases.

etphipp:
So all of these failures are build errors in TrilinosCouplings arising from uninitialized variable use warnings turned into errors.  They all appear to stem from the same pattern that is repeated in numerous examples:
```
  Scalar flux[3];
  Scalar material[3][3];
  Scalar f = 0.;

  // Indicate the independent variables
  x.diff(0,3);
  y.diff(1,3);
  z.diff(2,3);

  // Get exact solution and its gradient
  u = exactSolution(x, y, z);
  exactSolutionGrad(grad_u, x, y, z);

  // Get material tensor
  materialTensor<Scalar>(material, x, y, z);

  // Compute total flux = (A.grad u)
  for(int i = 0; i < 3; i++){

    // Add diffusive flux
    for(int j = 0; j < 3; j++){
      flux[i] += material[i][j]*grad_u[j];
    }
  }
```
The code is called with `Scalar = Sacado::Fad::SFad<double,3>`, and one can easily see that `flux` is used uninitialized, which is clearly an error.  The only reason it worked before was the old design for SFad always zero'd out the derivative array in the default constructor, and the new Fad design does not do this (this is only true for SFad though.  SLFad and DFad will zero the derivative array during the first resize that happens in the += operation because while they don't initialize the derivative array in the default constructor, they do initialize the derivative array size to zero in the default constructor and new derivative components are always zero'd out when size increases).

Part of me wants to just fix these examples in TrilinosCouplings, because they are wrong, and would not work correctly with any other scalar type (e.g., double).  But part of me also thinks I should make the behavior of the default SFad constructor in the new design consistent with the old.

Thoughts?

@trilinos/trilinoscouplings

etphipp:
I went ahead and fixed the TrilinosCouplings examples, because I don't think it is reasonable to expect the code to work the way it was written.  I haven't decided if I will change the behavior of SFad to be consistent with the old design.  The pro is it keeps SFad consistent with old behavior, and also consistent with DFad and SLFad.  The con is it unnecessarily reduces performance.

rppawlo:
My preference is to do the right thing for performance.

etphipp:
Thanks @rppawlo!

rppawlo:
@etphipp - can we merge?

PR #7416, opened 2020-05-23
ROL: changes to Thyra Adapters for SimOpt interface

mperego:
@trilinos/rol 

## Motivation
Improve performance by carefully avoiding re-computation of the objective, residual and their gradients when the state and/or control variables are not updated.
improved output messages (mainly for debugging purposes).

## Testing
Changes are tested in Piro test (collection of tests)  `Piro_AnalysisDriverTpetra`.

### DIFF
--- a/packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
+++ b/packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
@@ -52,6 +52,7 @@
 #include "ROL_Constraint_SimOpt.hpp"
 
 #include "Tpetra_CrsMatrix.hpp"
+#include "Teuchos_VerbosityLevel.hpp"
 
 using namespace ROL;
 
@@ -61,16 +62,22 @@ class ThyraProductME_Constraint_SimOpt : public Constraint_SimOpt<Real> {
 
 public:
 
-  ThyraProductME_Constraint_SimOpt(Thyra::ModelEvaluatorDefaultBase<double>& thyra_model_, int g_index_, const std::vector<int>& p_indices_,Teuchos::RCP<Teuchos::ParameterList> params_ = Teuchos::null, bool recompute = false) :
-    thyra_model(thyra_model_), g_index(g_index_), p_indices(p_indices_), params(params_) {
+  ThyraProductME_Constraint_SimOpt(Thyra::ModelEvaluatorDefaultBase<double>& thyra_model_, int g_index_, const std::vector<int>& p_indices_,
+    Teuchos::RCP<Teuchos::ParameterList> params_ = Teuchos::null, Teuchos::EVerbosityLevel verbLevel= Teuchos::VERB_HIGH) :
+        thyra_model(thyra_model_), g_index(g_index_), p_indices(p_indices_), params(params_),
+        out(Teuchos::VerboseObjectBase::getDefaultOStream()),
+        verbosityLevel(verbLevel){
     thyra_solver = Teuchos::null;
-    updateValue = updateJacobian1 = true;
-    value_ = 0;
+    computeValue = computeJacobian1 = solveConstraint = true;
     num_responses = -1;
-    x_ptr = Teuchos::null;
-    grad_ptr = Teuchos::null;
-    if(params != Teuchos::null)
+    value_ptr_ = Teuchos::null;
+    rol_u_ptr = Teuchos::null;
+    rol_z_ptr = Teuchos::null;
+    jac1 = Teuchos::null;
+    if(params != Teuchos::null) {
       params->set<int>("Optimizer Iteration Number", -1);
+      params->set<Teuchos::RCP<Vector<Real> > >("Optimization Variable", Teuchos::null);
+    }
   };
 
   void setExternalSolver(Teuchos::RCP<Thyra::ModelEvaluatorDefaultBase<double>> thyra_solver_) {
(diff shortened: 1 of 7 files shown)


packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
```diff
@@ -61,16 +62,22 @@ class ThyraProductME_Constraint_SimOpt : public Constraint_SimOpt<Real> {
 
 public:
 
-  ThyraProductME_Constraint_SimOpt(Thyra::ModelEvaluatorDefaultBase<double>& thyra_model_, int g_index_, const std::vector<int>& p_indices_,Teuchos::RCP<Teuchos::ParameterList> params_ = Teuchos::null, bool recompute = false) :
-    thyra_model(thyra_model_), g_index(g_index_), p_indices(p_indices_), params(params_) {
+  ThyraProductME_Constraint_SimOpt(Thyra::ModelEvaluatorDefaultBase<double>& thyra_model_, int g_index_, const std::vector<int>& p_indices_,
+    Teuchos::RCP<Teuchos::ParameterList> params_ = Teuchos::null, Teuchos::EVerbosityLevel verbLevel= Teuchos::VERB_HIGH) :
```
rppawlo, line 66: Should the default verbosity level be high?

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
```diff
@@ -331,14 +397,15 @@ class ThyraProductME_Constraint_SimOpt : public Constraint_SimOpt<Real> {
...
+    if(computeJacobian1)
     {
       outArgs.set_W_op(lop);
       thyra_model.evalModel(inArgs, outArgs);
       outArgs.set_W_op(Teuchos::null);
       inArgs.set_x(Teuchos::null);
       jac1 = lop;
-      updateJacobian1 = false;
+
+      computeJacobian1 = false;
```
rppawlo, line 408: The use of computeJacobian1 looks like it is introducing "state" into the model evaluator. Is that the case? It is something we have tried to avoid in the design.

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
```diff
@@ -654,54 +768,104 @@ class ThyraProductME_Constraint_SimOpt : public Constraint_SimOpt<Real> {
...
     }
+
     if(params != Teuchos::null)
       params->set<int>("Optimizer Iteration Number", iter);
   }
 
   /** \brief Update constraint functions with respect to Opt variable.
                 x is the optimization variable,
-                flag = true if optimization variable is changed,
+                flag = ??,
```
rppawlo, line 801: Does this doxygen need to be updated?

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
```diff
@@ -654,54 +768,104 @@ class ThyraProductME_Constraint_SimOpt : public Constraint_SimOpt<Real> {
...
   }
 
+  bool z_hasChanged(const Vector<Real> &rol_z) const {
+    bool changed = true;
+    if (Teuchos::nonnull(rol_z_ptr)) {
+      auto diff = rol_z.clone();
+      diff->set(*rol_z_ptr);
+      diff->axpy( -1.0, rol_z );
+      Real norm = diff->norm();
+      changed = (norm == 0) ? false : true;
```
rppawlo, line 837: Should this check have a tolerance due to roundoff?

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
```diff
@@ -654,54 +768,104 @@ class ThyraProductME_Constraint_SimOpt : public Constraint_SimOpt<Real> {
...
+  }
+
+  bool u_hasChanged(const Vector<Real> &rol_u) const {
+    bool changed = true;
+    if (Teuchos::nonnull(rol_u_ptr)) {
+      auto diff = rol_u.clone();
+      diff->set(*rol_u_ptr);
+      diff->axpy( -1.0, rol_u );
+      Real norm = diff->norm();
+      changed = (norm == 0) ? false : true;
```
rppawlo, line 849: Same here.

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Objective.hpp
```diff
@@ -164,28 +194,58 @@ class ThyraProductME_Objective : public Objective<Real> {
...
       }
     }
   }
+  
+  bool x_hasChanged(const Vector<Real> &rol_x) {
+    bool changed = true;
+    if (Teuchos::nonnull(rol_x_ptr)) {
+      rol_x_ptr->axpy( -1.0, rol_x );
+      Real norm = rol_x_ptr->norm();
+      changed = (norm == 0) ? false : true;
```
rppawlo, line 228: same tolerance question

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Objective_SimOpt.hpp
```diff
@@ -182,31 +242,86 @@ class ThyraProductME_Objective_SimOpt : public Objective_SimOpt<Real> {
...
   }
 
+  bool z_hasChanged(const Vector<Real> &rol_z) const {
+    bool changed = true;
+    if (Teuchos::nonnull(rol_z_ptr)) {
+      auto diff = rol_z.clone();
+      diff->set(*rol_z_ptr);
+      diff->axpy( -1.0, rol_z );
+      Real norm = diff->norm();
+      changed = (norm == 0) ? false : true;
```
rppawlo, line 293: same tolerance question

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Objective_SimOpt.hpp
```diff
@@ -182,31 +242,86 @@ class ThyraProductME_Objective_SimOpt : public Objective_SimOpt<Real> {
...
+  }
+
+  bool u_hasChanged(const Vector<Real> &rol_u) const {
+    bool changed = true;
+    if (Teuchos::nonnull(rol_u_ptr)) {
+      auto diff = rol_u.clone();
+      diff->set(*rol_u_ptr);
+      diff->axpy( -1.0, rol_u );
+      Real norm = diff->norm();
+      changed = (norm == 0) ? false : true;
```
rppawlo, line 305: same tolerance question

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
mperego: I was debating about this. I thought that when one tries it out, without knowing all the options, it's good to have more info in output. For real applications, one can change the verbose level as preferred. Anyway, it's not going to affect performance, it's just many print outs (I'm not printing out, e.g. the entries of a vector). I can change if you disagree with this.

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
mperego: Unfortunately yes. I don't see an easy way to overcome this. Alternatively I could write this info in the parameter list provided by the application.. but it seems even less clean to me.
Note that if we used something like an observer object provided by, e.g., Piro, then we would need ROL to depend on Piro.. which is bad. The alternative is to move the Thyra adapters into Piro. Which could be a better solution. 
I know that ROL is going to refactor the way the update function works. Not sure that will help (@dridzal ).

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
mperego: Sorry.. at some point I thought that the flag was signalling whether the optimization variable is changed or not. However, it has a more subtle meaning and it does not appear to be consistent.
ROL team is going to completely refactor the udate function, that's why I left the `??` rather then trying to understand this better. Anyway, I'm going to work more on this as we add support for second derivatives.

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
mperego: No.. I really want the two field to be exactly the same, in which case `diff` is a vector of zeros and  norm will be zero with no roundoff errors.

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
mperego: same as above

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Objective.hpp
mperego: same as above

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Objective_SimOpt.hpp
mperego: same as above

packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Objective_SimOpt.hpp
mperego: same as above

PR #7430, opened 2020-05-26
Belos: Avoid re-allocation in PseudoBlockCG

cgcgcg:
@trilinos/belos 

## Motivation
PseudoBlockCG re-allocates its state vectors in every solve. This PR fixes that. The same pattern can probably be found in other solvers, but before changing this everywhere, I'd like some opinion from @trilinos/belos.

### DIFF
--- a/packages/belos/src/BelosCGIter.hpp
+++ b/packages/belos/src/BelosCGIter.hpp
@@ -364,6 +364,16 @@ class CGIter : virtual public CGIteration<ScalarType,MV,OP> {
   void CGIter<ScalarType,MV,OP>::initializeCG(CGIterationState<ScalarType,MV>& newstate)
   {
     // Initialize the state storage if it isn't already.
+    if (!newstate.R.is_null() && MVT::GetNumberVecs(*newstate.R) == 1 &&
+        !newstate.Z.is_null() && MVT::GetNumberVecs(*newstate.Z) == 1 &&
+        !newstate.P.is_null() && MVT::GetNumberVecs(*newstate.P) == 1 &&
+        !newstate.AP.is_null() && MVT::GetNumberVecs(*newstate.AP) == 1) {
+      R_ = Teuchos::rcp_const_cast<MV>(newstate.R);
+      Z_ = Teuchos::rcp_const_cast<MV>(newstate.Z);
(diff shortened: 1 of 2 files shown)


jhux2:
@jennloe @hkthorn

jhux2:
@srajama1

jennloe:
Slowly digging through the pieces of this-
1. What is the context of calling solve() multiple times?  A new rhs each time?
2.  Why all the 'if numVecs==1' in the initialize function?
3. If someone changed the problem to have a matrix A with a different number of rows than the previous problem, does this break?  I suspect so.  R0 will be the right size, but the other retained vectors would not be.

cgcgcg:
@jennloe 
1. Yes, time-stepping with a constant step size.
2. Just making sure that the number of vectors hasn't changed. I believe this is the single rhs case, other CG iterations should have the general case.
3. Hm, Belos does allow changing the linear problem. This doesn't seem to be tested anywhere though, since this definitely should break then.

jennloe:
2. Isn't that case taken care of with `if (R_ == Teuchos::null || MVT::GetNumberVecs(*R_)!=blockSize_) {  \\Re-allocate storage} ` in `BlockCGIter::setStateSize`?  I guess the way you have written it saves a little work.
3. No, I don't think changing the problem to a matrix of different size is tested anywhere.  Perhaps it should be.  I wonder what else would break...

cgcgcg:
2. The BlockCG case seems to work a tiny bit differently from the PseudoBlockCG case. There is no such check there:
https://github.com/trilinos/Trilinos/blob/442a24ba155050971e7cba30fa2c22fdda9ebce9/packages/belos/src/BelosCGIter.hpp#L337-L346
3. Well, question is whether any sane person would attempt it. Changing the matrix entries makes sense, but changing the dimensions?

hkthorn:
@cgcgcg Because the solve methods recreate the iteration every time that solve is called, then the state is recreated.  For single-vector iterations saving the state between solves makes sense.  For multiple-RHS or block methods that have deflation, this logic is a little more complicated.  Thankfully, there are specializations for single-vector iterations in Belos, so we should be able to avoid the reallocations for those versions of the solvers, as long as we do it safely.  Thanks!

PR #7439, opened 2020-05-27
Update snapshot of Compadre

kuberry:
@trilinos/compadre 

Added divergence-free basis defined at points other than just the target site.
Previously the divergence-free basis was only hard-coded for order up to 4 and in a way that only allowed evaluations at the target sites.

Moved much of the scalar basis and divergence-free basis definitions out of the assembly routine for filling in the P matrices. Increases encapsulation.


mperego:
@kuberry thanks. I'll look at it. Can you please briefly describe the changes in the PR description and in the commit message? (feel free to amend the commit message and force push).

kuberry:
Those Cuda tests are strange. Indicates that it can't build the Lapack test or GMLS_Host test because it can't link to certain cuda libraries related to dense matrix solution for SVD. All of the other executables require exactly the same libraries and all built correctly.

kyungjoo-kim:
@kuberry If you look at the command line of the failed build,
```
 "/home/projects/ppc64le-pwr8-nvidia/openmpi/2.1.2/gcc/7.2.0/cuda/9.2.88/bin/mpicxx" "-pedantic" "-Wall" "-Wno-long-long" "-Wwrite-strings" "-expt-extended-lambda" "-lineinfo" "-arch=sm_37" "-Xcudafe" "--diag_suppress=esa_on_defaulted_function_ignored" "-O3" "-DNDEBUG" "packages/compadre/examples/CMakeFiles/Compadre_LAPACK_Test.dir/UnitTest_ThreadedLapack.cpp.o" "-o" "packages/compadre/examples/Compadre_LAPACK_Test.exe" "packages/compadre/src/libcompadre.a" "packages/kokkos/algorithms/src/libkokkosalgorithms.a" "packages/kokkos/containers/src/libkokkoscontainers.a" "packages/kokkos/core/src/libkokkoscore.a" "-ldl" "/home/projects/ppc64le-pwr8-nvidia/cuda/9.2.88/lib64/libcudart.so" "/home/projects/ppc64le-pwr8-nvidia/cuda/9.2.88/lib64/libcublas.so" "/home/projects/ppc64le-pwr8-nvidia/cuda/9.2.88/lib64/libcufft.so" "-L/home/projects/ppc64le-pwr8/netlib/3.8.0/gcc/7.2.0/lib" "-llapack" "-lgfortran" "-lgomp" "-lblas" "-lgfortran" "-lgomp" "-lblas" "-lm"
```
it does not include cuSolver. I think that you need to add dependence of the cuSolver. For many cases, cuSolver also requires cuSparse. Maybe it is safe to add those two TPL dependencies.

kuberry:
Thanks for catching that @kyungjoo-kim. It seemed strange, since these are normally in my dependencies. This is the first growing pain of not keeping my own repo up-to-date with something I committed to Trilinos (Dependencies.cmake was changed, but not in my repo, so when I snapshotted I got an older version).

kuberry:
Hmmm, 

terminate called after throwing an instance of 'std::logic_error'
  what():  /home/trilinos/jenkins/ride/workspace/trilinos-folder/Trilinos_pullrequest_cuda_9.2/Trilinos/packages/compadre/src/Compadre_PointCloudSearch.hpp:156: FAIL:
(std::is_same<typename view_type::memory_space, Kokkos::HostSpace>::value) && "Views passed to PointCloudSearch at construction should reside on the host."

And the view being passed is a:
Kokkos::View<double**>::HostMirror source_coords = Kokkos::create_mirror_view(source_coords_device);

So a ::HostMirror is violating this condition. I've not seen this on any other platform (all Cuda) before.

kyungjoo-kim:
What is the call stack ?

kyungjoo-kim:
ah.... also you need to remove the assert. there should be kokkos assert something like host accessible. I do not recall now.

kuberry:
Thank you for the explanation @kyungjoo-kim. I think with the advice you've given I'll be able to navigate getting the right changes in. I had generally been using the DefaultExecutionSpace::memory_space with the intention of it being Kokkos::CudaSpace. Definitely a good idea to be specific (and I will be). And yes, it wasn't written to use UVM space.

kuberry:
All issues with the asserts for memory space are fixed (switched to checking if accessible from host) and all tests now pass, with the exception of a few timeouts on Cuda.

mperego:
> All issues with the asserts for memory space are fixed (switched to checking if accessible from host) and all tests now pass, with the exception of a few timeouts on Cuda.

Unfortunately this is becoming a recurrent issue... are the tests timing out new?

mperego:
@kuberry  @jmgate I really don't like that the commit message is not informative. Is it possible to modify the automatic snapshot tool so that one can write a meaningful message? or so that it grabs the commits's messages from the standalone repo?

kuberry:
@mperego It simply grabs the last commit in the compadre repo.  As far as the timeouts, these are not new tests. They didn't timeout before because Cuda wasn't tested when the package was first inserted. None of those assertions were added since the initial commit of the package

kuberry:
While the snapshot creates the message in some way (it can't write a summary for us), we can do a git commit --amend after snapshotting and then put a summary in there.

kuberry:
I think leaving the snapshotter as it is would be the best way to go and just updating the commit message to be a summary of changes before asking for a PR.

jmgate:
Yes, that'd be my recommendation.

kuberry:
Do you want the summary in the log for this PR or just going forward?

mperego:
Amending the commit is fine with me. Yeah if you do not mind, once you are done with all the changes you could amend the commit with e.g. the summary you put in the PR message. Thanks!

kuberry:
@mperego Added doxygen documentation for the bases files that were added. Also, refactored the doxygen so CMake no longer configures Doxyfile.in->Doxyfile. Now there is just a Doxyfile and it has it's version populated when ./doc/build_docs is called.

PR #7443, opened 2020-05-28
Teuchos/Kokkos: Fixes for some potential upcoming changes to Kokkos

DavidPoliakoff:
@trilinos/teuchos

## Motivation

I've been adding a lot of functionality to Kokkos, some of which has effects in Trilinos. Because the scope of Tools is expanding, we're moving the namespace from "Profiling" to "Tools." In order to make that not a problem, we reexpose everything that used to be in "Profiling" as an alias in namespace "Kokkos::Profiling" to the corresponding entry in "Tools."

99% of the time that's enough. The exception is that in Teuchos StackedTimer, those functions in Profiling are declared "extern" in the Profiling namespace. This _also_ is fine, unless somebody has included the actual Kokkos header in which those are defined in namespace Kokkos::Tools. This happens to be the case somewhere in MueLu, they apparently include Kokkos and _then_ include the header I modified.

This PR fixes that problem by checking for the header in which these functions are defined having been included. If it has, these definitions are omitted. Also, this doesn't cause any harm if the namespace doesn't move, in that case you just avoid declaring something extern that doesn't need to be clared.

To the best of my knowledge this isn't related to any other work within Trilinos, but will be necessary to handle the changes from kokkos/kokkos#2422

## Stakeholder Feedback

This came out of conversations between @vbrunini and myself

## Testing

I'm not 100% sure on this, I'd like advice from somebody in Trilinos. I'd be astonished if we didn't have tests for StackedTimer, but I don't know my way around the testing system. If we cover StackedTimer in tests, this condition is satisfied

### DIFF
--- a/packages/teuchos/comm/src/Teuchos_StackedTimer.hpp
+++ b/packages/teuchos/comm/src/Teuchos_StackedTimer.hpp
@@ -57,12 +57,14 @@
 #include <iostream>
 
 #if defined(HAVE_TEUCHOS_KOKKOS_PROFILING) && defined(HAVE_TEUCHOSCORE_KOKKOSCORE)
+#ifndef KOKKOS_IMPL_KOKKOS_PROFILING_HPP
 namespace Kokkos {
 namespace Profiling {
 extern void pushRegion (const std::string&);
 extern void popRegion ();
 } // namespace Profiling
 } // namespace Kokkos
+#endif
 #endif
 
 


DavidPoliakoff:
So I'm not able to request reviews, but somebody on the Teuchos team would likely be appropriate

rppawlo:
@DavidPoliakoff - we have pretty extensive testing for stacked timer. This looks fine to me. I will run a test against empire later today.

DavidPoliakoff:
Don't merge this, it doesn't fully fix the problem

DavidPoliakoff:
Update: I fixed the Kokkos side of the issue, merging this side should now fully fix the problem

rppawlo:
@DavidPoliakoff - is this still needed? If so it passed all tests and is ready to push.

DavidPoliakoff:
@rppawlo, actually with what I've done Kokkos-side, we shouldn't need this, closing

Issue #2630, opened 2018-04-24
Tpetra: Deprecate & remove get*NumDiags & is{Lower,Upper}Triangular


mhoemmen:
@trilinos/tpetra 

Deprecate and remove the following functions from RowGraph, RowMatrix, CrsGraph, and CrsMatrix:
  - `isLowerTriangular`
  - `isUpperTriangular`
  - `getGlobalNumDiags` 
  - `getNodeNumDiags`

## Progress

We've finished deprecating the functions.  We have work-arounds in Ifpack2 and Tpetra, for getting the functionality without triggering the deprecation warnings.  We've also wrapped the deprecated functions in `#ifdef TPETRA_ENABLE_DEPRECATED_CODE ... #endif`.  The only thing remaining is to remove the deprecated functions.  This must wait until the next Trilinos major release.

## Motivation and Context

The first call of `CrsGraph::fillComplete` for a given graph instance, invokes `CrsGraph::computeLocalConstants`.  In order to compute what `getNodeNumDiags`, `getGlobalNumDiags`, and `is{Lower,Upper}Triangular()` need, `computeLocalConstants` must read all the entries in the matrix.  This takes significant time, as @alanw0 reported in #2565.  It's only rarely that downstream code needs to know this information.

Tpetra developers deem `nodeMaxNumRowEntries_` necessary for ML/MueLu "getrow" implementations, to preallocate buffers.  Tpetra developers are not sure if `getGlobalMaxNumRowEntries` is actually necessary.

## Related Issues
* Related to #2565

csiefer2:
Claiming MueLu/Xpetra

csiefer2:
I have Xpetra & MueLu purged on my machine and the tests all work  

@kddevin Can you take care of this one (and push separately)?  

zoltan2/test/unit/models/GraphModel.cpp

kddevin:
@csiefer2 Zoltan2 is done.  See PR #2634 waiting for your review.

csiefer2:
#2635 awaiting review & merge

mhoemmen:
@kddevin I jumped the gun a bit and merged #2634 for you :)

mhoemmen:
@csiefer2 I reviewed & approved #2635; looks good!

mhoemmen:
@csiefer2 I'm weirdly not able to push the merge button on #2635 despite it having been tested.

mhoemmen:
[...]

PR #7446, opened 2020-05-28
Tpetra: Review fences and add comments

MicheldeMessieres:
This PR is not intended for submission in this form.
Comments have been added for discussion purposes.

Note some fences have been tentatively removed and the current state passes with CUDA_LAUNCH_BLOCKING=0. However these fence removals need further discussion/consideration since passing tests don't prove anything regarding fence requirements.

@trilinos/tpetra 

## Motivation

To refactor and remove fences when possible.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
@@ -1111,6 +1111,13 @@ public:
     const Scalar one_minus_omega = Teuchos::ScalarTraits<Scalar>::one()-omega;
     const Scalar     minus_omega = -omega;
 
+    // FENCE REVIEW - CONFIRMED FAILURE
+    //   Testing: This code is exercised by unit tests.
+    //   White:   Fails Ifpack2Relaxation_double_int_longlong_TestLowerTriangularBlockCrsMatrix_UnitTest
+    //            Ifpack2Relaxation_double_int_longlong_TestUpperTriangularBlockCrsMatrix_UnitTest
+    //   Plan:
+    //   Notes:
+
     Kokkos::fence();
 
     if (numVecs == 1) {


packages/tpetra/core/src/Tpetra_Export_def.hpp
```diff
@@ -243,6 +243,13 @@ namespace Tpetra {
     LO numPermutes = 0;
     LO numExports = 0;
 
+    // FENCE REVIEW - CONFIRMED FAILURE
+    //   Testing: This code is exercised by unit tests.
+    //   GTX960:  Fails with CUDA_LAUNCH_BLOCKING=0 and fence removed. TpetraCore_SubmapExportTests_MPI_2
+    //   White:   Not checked since failure already confirmed on GTX960.
+    //   Plan:    Potentially refactor UVM access below and remove fence.
+    //   Notes:
+
```
csiefer2, line 252: In setupSamePermuteExport() : It isn't to me clear why this needs a fence.  I need to look at this in more detail.

packages/tpetra/core/src/Tpetra_Export_def.hpp
```diff
@@ -517,6 +524,13 @@ namespace Tpetra {
     host_remote_lids_type remoteLIDs
       (view_alloc_no_init ("remoteLIDs"), numRemoteIDs);
 
+    // FENCE REVIEW - CONFIRMED FAILURE
+    //   Testing: This code is exercised by unit tests.
+    //   GTX960:  Fails with CUDA_LAUNCH_BLOCKING=0 and fence removed. TpetraCore_SubmapExportTests_MPI_2
+    //   White:   Not checked since failure already confirmed on GTX960.
+    //   Plan:    Potentially refactor UVM access below and remove fence.
+    //   Notes:
+
```
csiefer2, line 533: In setupRemote(): It isn't clear why a fence needs to be right here.  MPI gets called before this (on a host array, so that probably doesn't need fencing.

packages/tpetra/core/src/Tpetra_Import_def.hpp
```diff
@@ -357,6 +357,13 @@ namespace Tpetra {
         Array<int> newRemotePIDs(indexIntoRemotePIDs-cnt);
         cnt = 0;
 
+        // FENCE REVIEW - NOT TESTED
+        //   Testing: This code is NOT exercised by unit tests.
+        //   GTX960:  N/A
+        //   White:   N/A
+        //   Plan:    Potentially add unit testing. Then potentialy remove fence.
+        //   Notes:
+
```
csiefer2, line 366: In Import():  Doesn't look like we should need this fence...

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
```diff
@@ -125,12 +155,27 @@ reallocDualViewIfNeeded (Kokkos::DualView<ValueType*, DeviceType>& dv,
...
-      execution_space().fence ();
+
+      // FENCE REVIEW - NOW PASSING
+      //   Testing: This code is exercised by unit tests.
+      //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   Plan:    Remove fence. But not sure yet why fence was placed.
+      //   Notes:
+
+      // execution_space().fence ();
```
kyungjoo-kim, line 178: This fence is not needed because it is just subview changing meta data only.

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
```diff
@@ -125,12 +155,27 @@ reallocDualViewIfNeeded (Kokkos::DualView<ValueType*, DeviceType>& dv,
...
       if (needFenceBeforeRealloc) {
-        execution_space().fence ();
+        // FENCE REVIEW - NOW PASSING
+        //   Testing: This code is exercised by unit tests.
+        //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   Plan:    Remove fence. But not sure yet why fence was placed.
+        //   Notes:
+
+        // execution_space().fence ();
```
kyungjoo-kim, line 165: Since needFenceBeforeRealloc is passed as an input argument, we should respect the intention. Fence is necessary.

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
```diff
@@ -100,22 +100,52 @@ reallocDualViewIfNeeded (Kokkos::DualView<ValueType*, DeviceType>& dv,
...
-      execution_space().fence ();
+
+      // FENCE REVIEW - NOW PASSING
+      //   Testing: This code is exercised by unit tests.
+      //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   Plan:    Remove fence. But not sure yet why fence was placed.
+      //   Notes:
+
+      // execution_space().fence ();
```
kyungjoo-kim, line 148: This fence is not necessary as the code does not allocate memory but simply decrease reference counting if the memory is managed.

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
```diff
@@ -100,22 +100,52 @@ reallocDualViewIfNeeded (Kokkos::DualView<ValueType*, DeviceType>& dv,
...
       if (needFenceBeforeRealloc) {
-        execution_space().fence ();
+        // FENCE REVIEW - NOW PASSING
+        //   Testing: This code is exercised by unit tests.
+        //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   Plan:    Remove fence. But not sure yet why fence was placed.
+        //   Notes:
+
+        // execution_space().fence ();
```
kyungjoo-kim, line 136: Since needFenceBeforeRealloc is passed as an input argument, we should respect the intention. Fence is necessary.

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
```diff
@@ -100,22 +100,52 @@ reallocDualViewIfNeeded (Kokkos::DualView<ValueType*, DeviceType>& dv,
...
-    execution_space().fence ();
+
+    // FENCE REVIEW - NOW PASSING
+    //   Testing: This code is exercised by unit tests.
+    //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+    //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+    //   Plan:    Remove fence. But not sure yet why fence was placed.
+    //   Notes:
+
+    // execution_space().fence ();
```
kyungjoo-kim, line 123: This fence is not necessary. Kokkos view allocation and view fill value functor put fence with execution space. It is safe to touch.

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
```diff
@@ -100,22 +100,52 @@ reallocDualViewIfNeeded (Kokkos::DualView<ValueType*, DeviceType>& dv,
...
     if (needFenceBeforeRealloc) {
-      execution_space().fence ();
+      // FENCE REVIEW - NOW PASSING
+      //   Testing: This code is exercised by unit tests.
+      //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   Plan:    Remove fence. But not sure yet why fence was placed.
+      //   Notes:
+
+      // execution_space().fence ();
```
kyungjoo-kim, line 110: Since needFenceBeforeRealloc is passed as an input argument, we should respect the intention. Fence is necessary.

packages/tpetra/core/src/Tpetra_replaceDiagonalCrsMatrix_def.hpp
```diff
@@ -90,6 +90,14 @@ replaceDiagonalCrsMatrix (CrsMatrix<SC, LO, GO, NT>& matrix,
...
   }
 
+  // FENCE REVIEW - NOW PASSING
+  //   Testing: This code is exercised by unit tests.
+  //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+  //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+  //   Plan:    Probably keep fence or refactor.
+  //   Notes:   colMap.getLocalElement should fail for UVM use of glMap_.
+  //            Need to verify if that path is tested.
+
```
kyungjoo-kim, line 100: This fence is necessary for UVM. Views used in the row map and colmap can use UVM and they are accessed in the following routine. So, we need to make sure that the values are available for touching in host.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -814,8 +814,16 @@ add (const Scalar& alpha,
...
+    // FENCE REVIEW - NOW PASSING
+    //   Testing: This code is exercised by unit tests.
+    //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+    //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+    //   Plan:    Remove fence.
+    //   Notes:   Confirm if this is still on host or was the sort fixed.
+
     //But now, have to sort on host (using UVM)
-    exec_space().fence();
+    // exec_space().fence();
```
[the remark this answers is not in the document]
brian-kelley: @csiefer2 Correct, it's gone (PR #7464)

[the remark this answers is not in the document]
brian-kelley: I could have pushed it to Michel's branch instead, but it fixes a KokkosKernels bug for the zero-row case, and improves performance for add().

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -4415,10 +4415,18 @@ namespace Tpetra {
...
+        //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   Plan:    Remove this fence.
+        //   Notes:   determineLocalTriangularStructure appears to be properly
+        //            setup as device.
+
         // Make sure that the GPU can see any updates made on host.
         // This code only reads the local graph, so we don't need a
         // fence afterwards.
-        execution_space().fence ();
+        // execution_space().fence ();
```
kyungjoo-kim, line 4429: @kddevin In my opinion, we need this fence even though it passes tests. Map uses kokkos view with relying on UVM (it does not have a dual view context). So, the values can be touched somewhere and we need to make sure the updates should be done before we use it. 

Passing unit tests means that we so far do not have a case that other functions touch the UVM views in the map before this function is used. However, it does not mean that this use case is right.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -4437,10 +4445,15 @@ namespace Tpetra {
...
-      execution_space().fence ();
+
+      // FENCE REVIEW - NOW PASSING
+      //   Testing: This code is exercised by unit tests. TpetraCore_CrsGraph_UnitTests0_MPI_4
+      //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   Plan:    Remove this fence.
+      //   Notes:   maxDifference looks properly setup as device.
+
+      // execution_space().fence ();
```
kyungjoo-kim, line 4456: @kddevin This one is a bit different from the above. Conservatively thinking, we also need the fence here as lclGraph_.row_map is on UVM and it can be still updated. In practice, the local graph construction should be done before this is called. This routine is computeLocalConstants. If we want a better code, we need a flag stating that the local graph is completed and safe to use it without fence. 

For now, I recommend to put the fence. Defining the state of local object can be improvements in the code.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -4129,6 +4129,15 @@ namespace Tpetra {
...
     else {
+      // FENCE REVIEW - NOT TESTED
+      //   Testing: This code is NOT exercised by unit tests.
+      //   GTX960:  N/A
+      //   White:   N/A
+      //   Plan:    Potentially add unit testing for this code.
+      //            Potentially remove this fence but following code appears
+      //            to need refactoring first.
+      //   Notes:
+
```
kyungjoo-kim, line 4140: @kddevin This fence is necessary. rowValues are from local crs matrix (kokkos crs matrix) with device type where the device currently has UVM. Without UVM, this routine is not correct as line 4150 access the values from the host code. 

Again, if we can define the state of the local object, we can skip the fence.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -4141,6 +4150,16 @@ namespace Tpetra {
...
+      // FENCE REVIEW - NOT TESTED
+      //   Testing: This code is NOT exercised by unit tests.
+      //   GTX960:  N/A
+      //   White:   N/A
+      //   Plan:    Potentially add unit testing for this code.
+      //            Potentially remove this fence but prior code appears
+      //            to need refactoring.
+      //   Notes:
+
```
kyungjoo-kim, line 4162: @kddevin This can be a design choice. We need to decide a API principle. In this code, we put fence in entering the code and also put the fence exiting the code. We should choose either of them. If we can make sure all Tpetra code assume that the code is non-blocking, we can leave the exiting as open-end. If we assume that all Tpetra code is blocking and synchronize when exiting, we do not need to put fence at the beginning of the API. 

We need to apply some consistency rules. According to the design principle, we can remove either of them.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -4208,6 +4227,15 @@ namespace Tpetra {
...
     else {
+      // FENCE REVIEW - NOT TESTED
+      //   Testing: This code is NOT exercised by unit tests.
+      //   GTX960:  N/A
+      //   White:   N/A
+      //   Plan:    Potentially add unit testing for this code.
+      //            Potentially remove this fence but following code appears
+      //            to need refactoring.
+      //   Notes:
+
```
kyungjoo-kim, line 4238: @kddevin Same as above. We need this fence or the fence at the end of the routine. This can be a discussion topic.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -4221,6 +4249,16 @@ namespace Tpetra {
...
+      // FENCE REVIEW - NOT TESTED
+      //   Testing: This code is NOT exercised by unit tests.
+      //   GTX960:  N/A
+      //   White:   N/A
+      //   Plan:    Potentially add unit testing for this code.
+      //            Potentially remove this fence but prior code appears
+      //            to need refactoring.
+      //   Notes:
+
```
kyungjoo-kim, line 4261: Same as above. We need this fence or the one in the above if we can establish a design rule.

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
```diff
@@ -1142,6 +1142,15 @@ init (const keys_type& keys,
...
 
+  // FENCE REVIEW - NOW PASSING
+  //   Testing: This code is exercised by unit tests.
+  //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+  //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+  //   Plan:    Remove this fence if we refactor code below to avoid UVM access.
+  //   Notes:   Need to check debug path as well.
+  //            Seems the passes are false oks - this code could fail.
+  //            For example see Mark's comments: 'This assumes UVM'
+
```
kyungjoo-kim, line 1153: @kddevin The code has two routes. One is parallel build of the hash table. The other is serial build from host. So, fence is only required when it uses parallel build. Otherwise, mirroring would apply fence inside when it uses UVM. 

@crtrott To make it sure, does deep copy on UVM apply fence inside ?

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
```diff
@@ -1196,6 +1205,15 @@ init (const keys_type& keys,
...
```
kyungjoo-kim, line 1216: @kddevin Before this fence, it calls computeOffsetsFromCounts, which uses parallel scan. If scan does not apply fence, then we need the fence. 

@crtrott Do we apply fence at the end of parallel scan ? This scan does not request the reduction (the last scanned value)

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
kddevin: @csiefer2 @jhux2 @kyungjoo-kim 
I agree with @kyungjoo's request for a design rule, and with the decision we made last week.  In the case of CrsMatrix leftScale and rightScale, however, I question whether we actually need the branch that contains the fences.

This branch is executed if the CrsMatrix does not have a localMatrix yet.  A CrsMatrix has a localMatrix if it has ever been fillCompleted OR if it was constructed using a localMatrix.  Thus, this branch gets executed only for a non-fillComplete matrix constructed with something other than a localMatrix (e.g., with insertGlobalValues).  As @MicheldeMessieres ' experiments revealed, we do not test the case of leftScaling or rightScaling a never-fillCompleted matrix.  

Rather than remove one fence and add tests to cover this branch, I wonder if we should support leftScale and rightScale only for fillComplete matrices.   Are there use-cases in Ifpack2, MueLu or Tpetra's MatrixMatrix routines that rely on scaling non-fillComplete matrices?  

I appreciate your thoughts.  In the meantime, I'll fire up a test with leftScale and rightScale functional only for fillComplete matrices, and see if anything downstream breaks.  Thanks!

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -2019,6 +2035,13 @@ void mult_A_B_reuse(
       }
     });
 
+  // FENCE REVIEW - CONFIRMED FAILURE
+  //   Testing: This code is exercised by unit tests.
+  //   GTX960:  Failed with CUDA_LAUNCH_BLOCKING=0 and fence removed. TpetraCore_MatrixMatrix_UnitTests_MPI_4
+  //   White:   Not checked since failure already confirmed on GTX960.
+  //   Plan:    Keep fence or decide if we can refactor.
+  //   Notes:
+
```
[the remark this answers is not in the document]
MicheldeMessieres: Resolved with PR #7510.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -2325,6 +2348,13 @@ void jacobi_A_B_newmatrix(
       }
     });
 
+  // FENCE REVIEW - NOW PASSING
+  //   Testing: This code is exercised by unit tests.
+  //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+  //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed
+  //   Plan:    Keep fence or decide if we can refactor.
+  //   Notes:   This fence is likely necessary - pass was probable false ok.
+
```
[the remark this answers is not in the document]
MicheldeMessieres: Resolved with PR #7510.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
```diff
@@ -2671,6 +2701,13 @@ void jacobi_A_B_reuse(
   MM = Teuchos::null;
 #endif
 
+  // FENCE REVIEW - CONFIRMED FAILURE
+  //   Testing: This code is exercised by unit tests.
+  //   GTX960:  Failed with CUDA_LAUNCH_BLOCKING=0 and fence removed. TpetraCore_MatrixMatrix_UnitTests_MPI_4
+  //   White:   Not checked since failure already confirmed on GTX960.
+  //   Plan:
+  //   Notes:
+
```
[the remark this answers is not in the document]
MicheldeMessieres: Resolved with PR #7510.

packages/tpetra/core/src/Tpetra_Details_packCrsMatrix_def.hpp
```diff
@@ -796,12 +796,34 @@ packCrsMatrix (const CrsMatrix<ST, LO, GO, NT>& sourceMatrix,
   }
 
   if (num_export_lids == 0) {
```
tjfulle, line 798: I believe all the fences in this file can safely be removed.

packages/tpetra/core/src/Tpetra_Details_packCrsGraph_def.hpp
```diff
@@ -715,12 +715,34 @@ packCrsGraph
   }
```
tjfulle, line 715: I believe all the fences in this file can safely be removed.

packages/tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp
```diff
@@ -148,6 +148,17 @@ outOfBounds (const IntegerType x, const IntegerType exclusiveUpperBound)
...
+      // FENCE REVIEW - CONFIRMED FAILURE
+      //   Testing: This code is exercised by unit tests.
+      //   GTX960:  Passed (confirmed) with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   White:   Failed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+          // TpetraCore_CrsMatrix_ReplaceDomainMapAndImporter_MPI_4
+          // TpetraCore_ImportExport2_UnitTests_MPI_4
+          // TpetraCore_MatrixMatrix_UnitTests_MPI_4
+      //   Plan:    Need to determine downstream failure point. Fix and remove this fence.
+      //   Notes:
+
```
tjfulle, line 161: On `ride` with `TPETRA_DEBUG=ON` and `TPETRA_DEBUG=OFF` I did not see failures in these three tests with the fence is removed.

tjfulle: Oh, wait, didn't set `CUDA_LAUNCH_BLOCKING=0`, I'll rerun.

tjfulle: Even with `CUDA_LAUNCH_BLOCKING=0`, these still pass.  Others fail, though.

packages/tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp
```diff
@@ -231,6 +242,16 @@ outOfBounds (const IntegerType x, const IntegerType exclusiveUpperBound)
...
+      // FENCE REVIEW - CONFIRMED FAILURE
+      //   Testing: This code is exercised by unit tests.
+      //   GTX960:  Passed (confirmed) with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+      //   White:   Failed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+          // TpetraCore_CrsMatrix_ReplaceDomainMapAndImporter_MPI_4
+          // TpetraCore_ImportExport2_UnitTests_MPI_4
+      //   Plan:    Need to determine downstream failure point. Fix and remove this fence.
+      //   Notes:
+
```
tjfulle, line 254: I don't see the same failure on ride with this fence removed

tjfulle: with and without `CUDA_LAUNCH_BLOCKING=0`

MicheldeMessieres: @tjfulle Thanks. I'll verify these on white and post the exact config and setup I was using. Something got broken with white's libs since this morning so many be delayed a bit.

tjfulle: I do get other failures with `CUDA_LAUNCH_BLOCKING=0`.  I think unrelated.  I'll bring them up in today's Tpetra meeting.

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
kddevin: On Kokkos' slack, @crtrott answered "no" regarding Kokkos' applying a fence at the end of this parallel scan.

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
kddevin: On Kokkos' slack, @crtrott says "yes it does if you don't hand it a space argument"

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
kyungjoo-kim: Then this fence is necessary as we need to make sure that the offset view completes before the view is used in the next functor.

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
kyungjoo-kim: We need to leave a note that fence is required when 3 argument deep copy is used in the above. It looks like that we are changing deep copy from 2 argument interface to 3 argument interface. 

Also, it is a bit odd and could be problematic if deep copy behaves differently with uvm and without uvm for the 3 argument deep copy.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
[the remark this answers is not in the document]
MicheldeMessieres: Resolved with #7464

packages/tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp
MicheldeMessieres: I confirmed the failures on white for removing this fence. However note the 2nd fence you asked about below I can not replicate the failures. So that may have been an error in my recording.

Logged into nodes like this: `bsub -Is -n 16 -q rhel7G bash`

Configure:
```
cmake \
  -DTpetra_ENABLE_TESTS=ON -DTrilinos_ENABLE_Tpetra=ON \
  ../trilinos
```

For this current branch state.
Running with CUDA_LAUNCH_BLOCKING=1

	 52 - TpetraCore_Core_initialize_where_tpetra_initializes_mpi_and_user_initializes_kokkos_MPI_2 (Failed)

Running with CUDA_LAUNCH_BLOCKING=0

	 44 - TpetraCore_Core_initialize_where_user_initializes_mpi_MPI_4 (Failed)
	 45 - TpetraCore_Core_ScopeGuard_where_user_initializes_mpi_MPI_4 (Failed)
	 51 - TpetraCore_Core_ScopeGuard_where_user_initializes_kokkos_MPI_1 (Failed)
	 53 - TpetraCore_Core_ScopeGuard_where_tpetra_initializes_mpi_and_user_initializes_kokkos_MPI_2 (Failed)

Remove this fence with CUDA_LAUNCH_BLOCKING=0:

	 44 - TpetraCore_Core_initialize_where_user_initializes_mpi_MPI_4 (Failed)
	 45 - TpetraCore_Core_ScopeGuard_where_user_initializes_mpi_MPI_4 (Failed)
	 51 - TpetraCore_Core_ScopeGuard_where_user_initializes_kokkos_MPI_1 (Failed)
	 53 - TpetraCore_Core_ScopeGuard_where_tpetra_initializes_mpi_and_user_initializes_kokkos_MPI_2 (Failed)
	 81 - TpetraCore_CrsMatrix_ReplaceDomainMapAndImporter_MPI_4 (Failed)
	130 - TpetraCore_ImportExport2_UnitTests_MPI_4 (Failed)
	165 - TpetraCore_MatrixMatrix_UnitTests_MPI_4 (Failed)

So that confirmed the listed failures and also that shows the extra fails we discussed in meeting which I have been ignoring for the moment. Soon I'll be rebasing this branch to the latest develop to incorporate a bunch of relevant PRs and I'll update the comments.

packages/tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp
MicheldeMessieres: This one I could not replicate the fails. I will update the branch and comments when I rebase it to develop. Thanks.

packages/tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp
kyungjoo-kim: This fence is in the pack function. Examining other pack functions, no fence is applied at the end. Following the design rule (apply fence as late as possible and tpetra api leave it open-end), the fence should be called from the caller of pack function in my opinion.

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
MicheldeMessieres: Removed #7545

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
MicheldeMessieres: Kept and added comment. #7545

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
MicheldeMessieres: Removed #7545

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
MicheldeMessieres: Removed #7545

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
MicheldeMessieres: Removed #7545

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
MicheldeMessieres: Kept and added comments #7545

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
MicheldeMessieres: Kept and added comments #7545

packages/tpetra/core/src/Tpetra_replaceDiagonalCrsMatrix_def.hpp
MicheldeMessieres: Kept and added comments #7546

packages/tpetra/core/src/Tpetra_DistObject_def.hpp
```diff
@@ -963,7 +963,15 @@ namespace Tpetra {
...
-            DES().fence (); // just in case UVM doesn't behave right
+
+            // FENCE REVIEW - NOW PASSING
+            //   Testing: This code is exercised by unit tests.
+            //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+            //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+            //   Plan:    Remove this fence.
+            //   Notes:   Comment suggests this fence was not shown to be required.
+
+            // DES().fence (); // just in case UVM doesn't behave right
```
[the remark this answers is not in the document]
MicheldeMessieres: Removed #7545

packages/tpetra/core/src/Tpetra_DistObject_def.hpp
```diff
@@ -999,7 +1007,14 @@ namespace Tpetra {
...
             }
-            DES().fence (); // just in case UVM doesn't behave right
+
+            // FENCE REVIEW - NOW PASSING
+            //   Testing: This code is exercised by unit tests.
+            //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+            //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+            //   Plan:    Remove fence
+            //   Notes:   Comment suggests this fence was not shown to be required.
+            // DES().fence (); // just in case UVM doesn't behave right
```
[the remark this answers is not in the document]
MicheldeMessieres: Removed #7545

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
MicheldeMessieres: Updated comments #7546

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
MicheldeMessieres: Updated comments #7546

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
MicheldeMessieres: Updated comments #7546

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
MicheldeMessieres: Updated comments #7546

packages/tpetra/core/src/Tpetra_Details_packCrsMatrix_def.hpp
MicheldeMessieres: Removed #7545

packages/tpetra/core/src/Tpetra_Details_packCrsGraph_def.hpp
MicheldeMessieres: Removed #7545

packages/tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp
MicheldeMessieres: Updated the comments to reflect this.

packages/tpetra/core/src/Tpetra_Details_reallocDualViewIfNeeded.hpp
MicheldeMessieres: Kept and added comments #7546

packages/tpetra/core/src/Tpetra_idot.hpp
```diff
@@ -287,6 +299,14 @@ lclDotRaw (typename ::Tpetra::MultiVector<SC, LO, GO, NT>::dot_type* const resul
...
         auto result_0d = subview (result, 0);
         KokkosBlas::dot (result_0d, X_lcl_1d, Y_lcl_1d);
+
+        // FENCE REVIEW - CONFIRMED FAILURE
+        //   Testing: This code is exercised by unit tests.
+        //   GTX960:  Failed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   White:   Not checked since failure already confirmed on GTX960.
+        //   Plan:    May consider downstream failure point and move fence.
+        //   Notes:
+
```
jhux2, line 309: I've confirmed that if this fence and the one below are commented out, and CUDA_LAUNCH_BLOCKING=0, these tests fail on vortex:
```
	 39 - TpetraCore_idot_MPI_4 (Failed)
	 44 - TpetraCore_Core_initialize_where_user_initializes_mpi_MPI_4 (Failed)
	 45 - TpetraCore_Core_ScopeGuard_where_user_initializes_mpi_MPI_4 (Failed)
```

packages/tpetra/core/src/Tpetra_idot.hpp
```diff
@@ -295,6 +315,14 @@ lclDotRaw (typename ::Tpetra::MultiVector<SC, LO, GO, NT>::dot_type* const resul
...
                                  pair_type (0, Y_numVecs));
         KokkosBlas::dot (result, X_lcl_2d, Y_lcl_2d);
+
+        // FENCE REVIEW - CONFIRMED FAILURE
+        //   Testing: This code is exercised by unit tests.
+        //   GTX960:  Failed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   White:   Not checked since failure already confirmed on GTX960.
+        //   Plan:    May consider downstream failure point and move fence.
+        //   Notes:
+
```
jhux2, line 325: See my comment above.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
MicheldeMessieres: Removed with PR 7560

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
MicheldeMessieres: Removed with PR 7560

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
MicheldeMessieres: Removed with PR 7560

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
MicheldeMessieres: Removed with PR 7560

packages/tpetra/core/src/Tpetra_idot.hpp
kyungjoo-kim: I think that this is a problem from KokkosBlast::dot. I believe that the dot uses parallel reduce and it should wait until the result is fully updated even though it uses a device result view.

MicheldeMessieres: Just a reminder we had this discussion regarding dot:
[https://github.com/kokkos/kokkos-kernels/pull/660](https://github.com/kokkos/kokkos-kernels/pull/660)

MicheldeMessieres: Note that only the TpetraCore_idot_MPI_4 failure is specifically related to this fence.

kyungjoo-kim: Oh...... I have a short-term memory and already said the opposite in that issue.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
MicheldeMessieres: @kyungjoo-kim @kddevin I was looking into the possibility of adding a flag for this. I think we are talking about a bool bNeedFence in Tpetra_Graph_decl.hpp which is reset when needed (whenever we create the static graph). If computeLocalConstants is called again and the graph has not changed, we can skip the fence, similar to DualView need_sync_host behaviors. However it looks like there is no current test where adding such a flag will make a difference compared to just leaving the fence. The graph is always recreated before computeLocalConstants is called again. So I wanted to confirm if this is the idea you had in mind.

I did note 2 places where computeLocalConstants is used in Tpetra but not tested by a unit test.
    Call of computeLocalConstants in Tpetra_CrsMatrix_def.hpp.    line 4857
    Call of computeLocalConstants in graph fillComplete in Tpetra_CrsGraph_def.hpp.    line 3628

packages/tpetra/core/src/Tpetra_idot.hpp
jhux2: Based on @brian-kelley's suggestion in today's Tpetra meeting, I put fences just before the calls to `ireduce` in `Tpetra_idot.hpp`.  (Brian said that there needed to be a sync before calling MPI.)   With these fences, the `idot` test passes.

MicheldeMessieres: @jhux2 Thanks. Let me know if you want me to PR that change.

brian-kelley: @jhux2 OK, good that it passes. I found this old discussion on KokkosKernels about whether KokkosBlas::dot should fence after the case of reducing to a 1-D view: https://github.com/kokkos/kokkos-kernels/pull/660 . The concensus there was that the caller of KokkosBlas::dot should fence if they're going to read the UVM result on host.

I think that if we could replace the fence you just added with a deep_copy instead, that makes the intent of the code clearer. You can deep_copy from a UVM view to its host mirror (itself) and Kokkos will just fence the execution space once. It would be identical performance but makes the intent clearer than a fence.

Something else is a little funky in the lclDotRaw():
```
typedef View<dot_type*, device_type> result_dev_view_type;
typedef typename result_dev_view_type::HostMirror result_host_view_type;

// later:
  result_dev_view_type result;
  result_host_view_type result_h;
  if (resultOnDevice) {
    result = result_dev_view_type (resultRaw, numVecs);
    if (runOnHost) {
      // create_mirror_view doesn't promise to create a deep copy, so we
      // can't rely on this case to avoid an extra buffer if the input
      // communicator is an intercomm.
      result_h = ::Kokkos::create_mirror_view (result);
    }   
    // If running on device, not host, then we don't use result_h.
  }
  else {
    result_h = result_host_view_type (resultRaw, numVecs);
    if (! runOnHost) { // running on device, not host
      // We have to do a separate allocation here, since resultRaw is
      // not accessible from device.
      result = result_dev_view_type ("result", numVecs);
    }   
    // If running on host, not device, then we don't use result.
  }
```
BTW, I'm pretty sure we should have ``Kokkos::MemoryTraits<Kokkos::Unmanaged>`` in the typedef result_dev_view_type, and use ``View<dot_type*, HostSpace, MemoryTraits<Unmanaged>`` for ``result_host_view_type``. We're just trying to write the result to a C-style array that either lives on device or host. I asked Christian T about what it's doing now on kokkos slack.

brian-kelley: After looking some more, I think there's no avoiding the fence before ireduce/iallreduce. The fences inside lclDotRaw can go.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
MicheldeMessieres: @kyungjoo-kim @kddevin I may be figuring this out. I will plan to open a PR to clarify any questions I have.

packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
```diff
@@ -1215,6 +1223,14 @@ namespace Tpetra {
...
       // Necessary until following UVM host accesses are changed - for example Crowptr
       // Also probably needed in mult_R_A_P_newmatrix_kernel_wrapper - did not demonstrate this in test failure yet
+
+      // FENCE REVIEW - CONFIRMED FAILURE
+      //   Testing: This code is exercised by unit tests.
+      //   GTX960:  Fails with CUDA_LAUNCH_BLOCKING=0 and fence removed. TpetraCore_MatrixMatrix_UnitTests_MPI_4
+      //   White:   Not checked since failure already confirmed on GTX960.
+      //   Plan:    Decide if we want to refactor and remove.
+      //   Notes:
+
```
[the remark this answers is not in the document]
csiefer2: This should have been resolved by my previous re-arranging of fences...

packages/tpetra/core/src/Tpetra_idot.hpp
jhux2: @brian-kelley So do you think a fence is better than a deep_copy?

brian-kelley: @jhux2 My thought was that since there's only one view that needs to be synchronized for access on host, a deep_copy is nicer than just a fence because a fence in the code doesn't tell you why it's there. A deep_copy from UVM to itself is internally just a fence, but the intent of accessing the data on host is clear without needing comments.

brian-kelley: But now, I'm not sure that's a good idea because if we compute on device and communicate from device, but have CUDA_LAUNCH_BLOCKING off, then you still need to fence before the communication, but the current code won't go into the if for if communication is needed. So I think just forget about this and keep what you had (one fence before the communication)

kyungjoo-kim: Is it really necessary to write lclDotRaw with this much of complexity ? With several deep copy (mostly mirroring), removing or adding one fence does not seem to affect overall performance except for a lucky case that can avoid all deep copies in the code.

brian-kelley: @kyungjoo-kim Yes, I'm overthinking it :)

Let's just fence the execution space before the communication and call it done.

Later if we want we can simplify things a bit: lclDotRaw might as well always put the result on host, there's no performance benefit at all to the "resultOnDevice" stuff. If it's computed on device, it needs to get copied to host either by deep copy or UVM.

jhux2: @brian-kelley Thank you for the explanation.  @MicheldeMessieres Let's put fences before each ireduce/iallreduce.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3951,6 +3951,13 @@ namespace Tpetra {
...
     if (Case2) {
+      // FENCE REVIEW - CONFIRMED FAILURE
+      //   Testing: This code is exercised by unit tests.
+      //   GTX960:  Fails with CUDA_LAUNCH_BLOCKING=0 and fence removed. TpetraCore_MultiVector_UnitTests_MPI_4
+      //   White:   Not checked since failure already confirmed on GTX960.
+      //   Plan:    Investigate the reduce - determine failure point - probably we need the fence for a host conversion.
+      //   Notes:
+
       Kokkos::fence();
       this->reduce ();
```
jhux2, line 3962: This fence can be moved later to just before the call to `allReduceView` [here](https://github.com/Tech-XCorp/Trilinos/blob/0f3cca46a411134d95ee3e5438e87053ceeca675/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp#L4059) and [here](https://github.com/Tech-XCorp/Trilinos/blob/0f3cca46a411134d95ee3e5438e87053ceeca675/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp#L4064).

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -4415,11 +4415,14 @@ namespace Tpetra {
...
-        // updates should be done before we use it.
-        execution_space().fence ();
+        // FENCE REVIEW - NOW PASSING
+        //   Testing: This code is exercised by unit tests.
+        //   GTX960:  Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   White:   Passed with CUDA_LAUNCH_BLOCKING=0 and fence removed.
+        //   Plan:    Remove this fence.
+        //   Notes:   determineLocalTriangularStructure is device so we don't
+        //            need to fence here.
+        // execution_space().fence ();
```
kddevin, line 4425: Regarding this fence:  I propose removing this entire branch of code and all the fields that it computes.  This branch computes constants lowerTriangular_ , upperTriangular_, and nodeNumDiags_.  I do not see anyplace in Trilinos where these are used after they are set. And I don't see an accessor that would let a user get these values.  Also, #2630 says they are inconsistent.
@trilinos/ifpack2 people:  do you need lowerTriangular_ , upperTriangular_  and nodeNumDiags_?

jhux2: @kddevin Maybe I'm missing it, but I don't see any accessor method for {lower|upper}Triangular_ or nodeNumDiags_.  As far as I can see, none of the three are used in Ifpack2.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -5855,6 +5881,13 @@ namespace Tpetra {
        "exportLIDs.size() = " << numExportLIDs << " != numPacketsPerLID.size()"
        " = " << numPacketsPerLID.size () << ".");
 
+    // FENCE REVIEW - NOT TESTED
+    //   Testing: This code is NOT exercised by unit tests.
+    //   GTX960:  N/A
+    //   White:   N/A
+    //   Plan:    Track if graph is safe and fence based on that.
+    //   Notes:   Need fence for getLocalViewRawConst and getGlobalViewRawConst
+
```
kddevin, line 5890: This case is exercised when packAndPrepare is called on a RowGraph, rather than on a CrsGraph.   I assume this use case is needed for Ifpack2; is that correct, @trilinos/ifpack2?

In this case, packAndPrepare calls DistObject's pack() method, which calls packFillActive(), which has these fences.
Method packFillActive() is needed only because of the old Teuchos::ArrayView<> interface to the pack() method; otherwise, I think packFillActiveNew() could be called.

If we need this capability, we should add a unit test for it.
Should we change the interface to DistObject::pack() to use Kokkos::Views instead of Teuchos::ArrayView()?  Or would that create too much work in Ifpack2?

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -1111,6 +1111,13 @@ class GetLocalDiagCopy {
     const Scalar one_minus_omega = Teuchos::ScalarTraits<Scalar>::one()-omega;
     const Scalar     minus_omega = -omega;
 
+    // FENCE REVIEW - CONFIRMED FAILURE
+    //   Testing: This code is exercised by unit tests.
+    //   White:   Fails Ifpack2Relaxation_double_int_longlong_TestLowerTriangularBlockCrsMatrix_UnitTest
+    //            Ifpack2Relaxation_double_int_longlong_TestUpperTriangularBlockCrsMatrix_UnitTest
+    //   Plan:
+    //   Notes:
+
```
kddevin, line 1120: @brian-kelley  Will this code (function localGaussSeidel) be going away with your refactor of Ifpack2 and Tpetra?  If so, we can leave the fence here until you do your refactor, which I assume will remove localGaussSeidel from Tpetra.   If not, I will look more carefully at the fence.

brian-kelley: @kddevin Yes, that will be going away.

PR #7448, opened 2020-05-28
Trilinos: Fix ATDM Kokkos arch

### LINKED ISSUE #7247: Framework: ATDM configuration should update the way Kokkos arch(es) are set
## Bug Report
@trilinos/framework 

### Description
What went wrong?  What should have happened?  Do you have an idea what might
fix things?  How has this bug affected you?  What are you trying to accomplish?
What specifically must we demonstrate to call this work complete?

The ATDM environment scripts use the deprecated syntax ``-DKOKKOS_ARCH=XYZ``. After Kokkos 3.0, this should be ``-DKokkos_ARCH_XYZ=ON``. If there are multiple architectures enabled (like CPU and GPU) there should be multiple arch enable flags.

### Steps to Reproduce
Example on Stria:

```
source $TRILINOS_SRC/cmake/std/atdm/load-env.sh stria-arm-release
cmake DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=$TRILINOS_SRC/cmake/std/atdm/ATDMDevEnv.cmake ...... $TRILINOS_SRC

# This prints out warning:

CMake Warning at packages/kokkos/cmake/kokkos_functions.cmake:33 (MESSAGE):
  Given deprecated option list KOKKOS_ARCH.  This must now be given as
  separate -D options, which assuming you spelled options correctly would be:

    -DKokkos_ARCH_ARMV8-TX2=ON

  Remove CMakeCache.txt and re-run.  For a list of valid options, refer to
  BUILD.md or even look at CMakeCache.txt (before deleting it).
  packages/kokkos/cmake/kokkos_arch.cmake:25 (KOKKOS_DEPRECATED_LIST)
  packages/kokkos/cmake/kokkos_tribits.cmake:181 (INCLUDE)
  packages/kokkos/CMakeLists.txt:156 (KOKKOS_SETUP_BUILD_ENVIRONMENT)
```

jjellio:
Currently, the ATDMDevSettings.cmake script sets `KOKKOS_ARCH=name`, but this doesn't do anything with Kokkos 3.0.  The script needs to set `Kokkos_ARCH_name=ON`

@bartlettroscoe 

## Testing
I've used this to build ATDM EMPIRE.  (but it isn't obvious I copied my forked Trilinos source to Github trilinos correctly - it needs testing to vet this works on another ATDM build)

Fixes #7247


jjellio:
crap it's on master.

jjellio:
uncrapped it to develop.

cmake/std/atdm/ATDMDevEnvSettings.cmake
```diff
@@ -293,7 +293,16 @@ ATDM_SET_CACHE(Kokkos_ENABLE_CXX11_DISPATCH_LAMBDA ON CACHE BOOL)
 ATDM_SET_CACHE(Kokkos_ENABLE_CUDA_LAMBDA "${ATDM_USE_CUDA}" CACHE BOOL)
 ATDM_SET_CACHE(Kokkos_ENABLE_DEBUG_BOUNDS_CHECK "${Trilinos_ENABLE_DEBUG}" CACHE BOOL)
 ATDM_SET_CACHE(Kokkos_ENABLE_DEBUG "${Trilinos_ENABLE_DEBUG}" CACHE BOOL)
-ATDM_SET_CACHE(KOKKOS_ARCH "$ENV{ATDM_CONFIG_KOKKOS_ARCH}" CACHE STRING)
+
+# Update Kokkos arch, expected ATDM_CONFIG_KOKKOS_ARCH=arch1;arch2;arch3
+# set the modern Kokkos arch, e.g., KOKKOS_ARCH=arch1 => Kokkos_ARCH_arch1=ON
+set(kokkos_arch_list "$ENV{ATDM_CONFIG_KOKKOS_ARCH}")
+foreach(kokkos_arch_loop_var ${kokkos_arch_list} )
+  ATDM_SET_ENABLE(Kokkos_ARCH_${kokkos_arch_loop_var} ON)
```
brian-kelley, line 301: @jjellio What's the difference between ATDM_SET_ENABLE and ATDM_SET_CACHE? Will this work if you rerun cmake with a different ``$ATDM_CONFIG_KOKKOS_ARCH``?

jjellio: Don't know. If I use SET_CACHE, it doesn't print the nice little message 'setting foo-bar'. They use set_enable later.

I've had this patch for a while, It's needed on ARM since they configure your BLAS library's architecture from the compiler's architecture flags.  I saw your bug report and decided to commit it.

It'd be nice if the cmake gurus can explain the way it should be done.

jjellio: It's also not obvious if I need to create the list. I had originally tried without the list variable, but it didn't work. It's just really annoying when this doesn't work.

bartlettroscoe: > What's the difference between ATDM_SET_ENABLE and ATDM_SET_CACHE? Will this work if you rerun cmake with a different $ATDM_CONFIG_KOKKOS_ARCH?

@jjellio, they are defined here:

https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/utils/ATDMDevEnvUtils.cmake

jjellio:
> @jjellio, that is not a CMake list, just a string separated with comas. See, for example:
> 
> https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/ats2/environment.sh#L35
> 
> which shows:
> 
> ```
> export ATDM_CONFIG_KOKKOS_ARCH=Power9,Volta70
> ```
> 
> It is an easy fix. With your permission, I will push the fixing commit to your branch.

Please do.  FYI, I was using semicolons in my settings (It felt intuitive, but a bash array would have felt better)

I saw the definitions, I just don't know which should or should not be used (both are used in that file), and I like for it print setting 'XYZ'

TODO: Kokkos needs to throw an error if an ARCH is passed in that is not supported. In the ARM environment, the spelling has changed, but Kokkos will default to just ignoring it (annoying).

brian-kelley:
@jjellio To catch invalid ``-DKokkos_ARCH_XYZ=ON``, kokkos would need to pattern match over the names of all existing CMake variables, which I don't think is possible. That was a benefit of the old syntax...

jjellio:
> @jjellio To catch invalid `-DKokkos_ARCH_XYZ=ON`, kokkos would need to pattern match over the names of all existing CMake variables, which I don't think is possible. That was a benefit of the old syntax...

Old syntax is superior. Kokkos could just use the loop type structure above in their own code. Seems easier/less error prone (since they can trivially validate the user inputs). I'm happy to open an issue with Kokkos on this, this seems like a pain point for users going forward.

bartlettroscoe:
Almost there.  Kokkos now likes the arch names to be upper case.

cmake/std/atdm/ATDMDevEnvSettings.cmake
jjellio: Thanks Ross, the comments are helpful.  I have a massacred ATDMDevSettings.cmake that allows for link time optimizations. (it requires twiddling CMAKE_AR and its default output).  I'd like to commit that (separately of course) - understanding the different ways to set variables will hopefully help me get that right the first time.

bartlettroscoe:
@jjellio, as of the last commit b806ffb, I think I have it.  I will post my manual testing on a few platforms as evidence.  But please feel free to test on your own.

bartlettroscoe:
I tested this on the 'sems-rhel6' env that does not define a KOKKOS_ARCH, on 'ats1' that defines KOKKOS_ARCH with just one element and on 'ats2' which has KOKKOS_ARCH with two values.

I think that is enough testing.

Okay to merge?

----

**A)** For 'sems-rhel6' env on 'crf450' I tested this with:

```
$ ./checkin-test-atdm.sh gnu-openmp-dbg-pt --enable-packages=Kokkos --configure
```

the 'configure.out' file showed:

```
-- Architectures:
<empty>
```

**B)** Testing this out on 'mutirno' with the 'ats1' env:

```
$ cd /lscratch1/rabartl/Trilinos.base/BUILDS/ATS1/CHECKIN/

$ ./checkin-test-atdm.sh ats1-hsw-intel-19 ats1-knl-intel-18 --enable-packages=Kokkos --configure
```

The `ats1-hsw-intel-19/configure.out` file showed:

```
-- Setting default Kokkos_ARCH_HSW=ON

...

Processing enabled package: Kokkos (Core, Containers, Algorithms, Tests, Examples)
...
-- Architectures:
--  HSW
```

The `ats1-knl-intel-18/configure.out` file showed:

```
-- Setting default Kokkos_ARCH_KNL=ON

...

Processing enabled package: Kokkos (Core, Containers, Algorithms, Tests, Examples)
...
-- Architectures:
--  KNL
```

**C)** Testing 'vortex' with the 'ats2' env:

```
$ cd /vscratch1/rabartl/Trilinos.base/BUILDS/VORTEX/CHECKIN/

$ ./checkin-test-atdm.sh ats2-cuda-opt --enable-packages=Kokkos --configure
```

and the generated configure.out file showed:

```
-- Setting default Kokkos_ARCH_POWER9=ON
-- Setting default Kokkos_ARCH_VOLTA70=ON

...

Processing enabled package: Kokkos (Core, Containers, Algorithms, Tests, Examples)
...
-- Architectures:
--  POWER9
--  VOLTA70
```

jjellio:
Awesome.  Now I can start filling Kokkos bugs if they aren't setting the right arch flags (they don't on Power9 I believe).

bartlettroscoe:
> Currently, the ATDMDevSettings.cmake script sets KOKKOS_ARCH=name, but this doesn't do anything with Kokkos 3.0

@jjellio, I think Kokkos 3.0 still reads `KOKKOS_ARCH=name` but it just warns you to stop using that.  For example, the warning shown [here](https://testing-dev.sandia.gov/cdash/build/5486664/configure) shows:

```
Processing enabled package: Kokkos (Core, Containers, Algorithms)
...
CMake Warning at packages/kokkos/cmake/kokkos_functions.cmake:33 (MESSAGE):
  Given deprecated option list KOKKOS_ARCH.  This must now be given as
  separate -D options, which assuming you spelled options correctly would be:

    -DKokkos_ARCH_KNL=ON

  Remove CMakeCache.txt and re-run.  For a list of valid options, refer to
  BUILD.md or even look at CMakeCache.txt (before deleting it).
  packages/kokkos/cmake/kokkos_arch.cmake:25 (KOKKOS_DEPRECATED_LIST)
  packages/kokkos/cmake/kokkos_tribits.cmake:177 (INCLUDE)
  packages/kokkos/CMakeLists.txt:156 (KOKKOS_SETUP_BUILD_ENVIRONMENT)

...
-- Architectures:
--  KNL
```

So it is still reading `KOKKOS_ARCH=name`.

bartlettroscoe:
> Old syntax is superior. Kokkos could just use the loop type structure above in their own code. Seems easier/less error prone (since they can trivially validate the user inputs). I'm happy to open an issue with Kokkos on this, this seems like a pain point for users going forward.

@jjellio and @brian-kelley, I suspect their morivation for switching to individual `DKokkos_ARCH_<ARCH_NAME>` cache vars was to allow documenting these in the cache.  Documenting all of the KOKKOS_ARCH options in a list is hard to do as a cache var.

Issue #7445, opened 2020-05-28
Sacado: changes broke Albany build


ikalash:
It looks like recent changes to Sacado broke all the nightly Albany builds: https://sems-cdash-son.sandia.gov/cdash/index.php?project=Albany&date=2020-05-28&filtercount=1&showfilters=1&field1=groupname&compare1=61&value1=Nightly .  In particular, it looks like setUpdateValue has gone away: 
```
/.../repos/Albany/src/evaluators/gather/PHAL_GatherScalarNodalParameter_Def.hpp(188): error: class "Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::StaticStorage<RealType, 100>>" has no member "setUpdateValue" .  
```
https://sems-cdash-son.sandia.gov/cdash/viewBuildError.php?buildid=2186

Is there an alternate routine we can use that this one was replaced with?

@trilinos/sacado 

@etphipp

etphipp:
Sorry about this @ikalash.  I guess I removed this functionality in the new Fad design.  This functionality had very limited utility, it was only useful for differentiating linear problems and would avoid computation of the residual in the derivative computation (which was really only useful for intrusive PCE with Stokhos).  I can add a dummy operation back to the Sacado Fad types if you want.  The best solution though would be just to remove this call from Albany.

ikalash:
Thanks @etphipp .  If I remove these calls there are additional problems, unfortunately: 
```
[  4%] Building CXX object src/CMakeFiles/albanyLib.dir/evaluators/utility/PHAL_FieldFrobeniusNorm.cpp.o
In file included from /home/ikalash/nightlyCDash/repos/Albany/src/PHAL_AlbanyTraits.hpp:21,
                 from /home/ikalash/nightlyCDash/repos/Albany/src/evaluators/utility/PHAL_FieldFrobeniusNorm.cpp:7:
/home/ikalash/nightlyCDash/repos/Albany/src/Albany_SacadoTypes.hpp: In instantiation of ‘struct Albany::ScalarConversionHelper<const double, Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::ViewStorage<const double, 0, 1, Sacado::Fad::Exp::GeneralFad<Sacado::Fad::Exp::DynamicStorage<double, double> > > > >’:
[...]

Issue #3438, opened 2018-09-13
KokkosKernals_sparse_* tests are failing on ATDM cuda builds


fryeguy52:
CC: @trilinos/kokkos-kernels , @kddevin  (Trilinos Data Services Product Lead)
 
## Next Action Status

PR #3442 merged on 9/14/2018 did not appear to fix any of the unit tests.  Next: Determine next action to fix or revert merge #3442 disable these tests (depending if this is a blocker for EMPIRE and/or SPARC or not) on Friday morning 9/17/2018?

## Description

As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2018-09-12&filtercombine=and&filtercombine=and&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=Trilinos-atdm-&field2=testname&compare2=63&value2=KokkosKernels_sparse_&field3=status&compare3=62&value3=Passed) the tests:
* KokkosKernels_sparse_serial_MPI_1 
* KokkosKernels_sparse_cuda_MPI_1 

are failing in all the cuda builds on white, ride, hansen, and waterman:

* Trilinos-atdm-waterman-gnu-debug-openmp
* Trilinos-atdm-waterman-cuda-9.2-debug
* Trilinos-atdm-waterman-cuda-9.2-opt
* Trilinos-atdm-white-ride-cuda-9.2-opt
* Trilinos-atdm-white-ride-cuda-9.2-debug-pt
* Trilinos-atdm-white-ride-cuda-9.2-debug
* Trilinos-atdm-hansen-shiller-cuda-8.0-opt
* Trilinos-atdm-hansen-shiller-cuda-9.0-debug
* Trilinos-atdm-hansen-shiller-cuda-9.0-opt

[...]

PR #7449, opened 2020-05-28
Sacado:  Fix some Albany build issues with the new design

etphipp:
For issue #7445.  The new design broke the Albany build.  Two issues (so
far):

* setUpdateValue() disappeared.  I went ahead and re-added it as a
  no-op.
* ScalarType<ViewFad<const T,...> >::type is no longer const T.  I fixed
  fixed this by adding another partial specialization of ScalarType for
  ViewFad.  I chose to go this route instead of adding
  const_scalar_type, non_const_scalar_type to all of the Fad classes
  because ViewFad is the only Fad type that can have a const
  value/scalar type.

### DIFF
--- a/packages/sacado/src/new_design/Sacado_Fad_Exp_GeneralFad.hpp
+++ b/packages/sacado/src/new_design/Sacado_Fad_Exp_GeneralFad.hpp
@@ -136,6 +136,24 @@ namespace Sacado {
         this->fastAccessDx(ith) = value_type(1.);
       }
 
+      //! Set whether this Fad object should update values
+      /*! Retained for backward compatibility.
+       */
+      KOKKOS_INLINE_FUNCTION
+      void setUpdateValue(bool update_val) {}
+
+      //! Return whether this Fad object has an updated value
+      /*! Retained for backward compatibility.
+       */
+      KOKKOS_INLINE_FUNCTION
+      bool updateValue() const { return true; }
+
+      //! Cache values
+      /*! Retained for backward compatibility.
+       */
+      KOKKOS_INLINE_FUNCTION
+      void cache() const {}
+
       //! Returns whether two Fad objects have the same values
       template <typename S>
       KOKKOS_INLINE_FUNCTION
--- a/packages/sacado/src/new_design/Sacado_Fad_Exp_ViewFad.hpp
+++ b/packages/sacado/src/new_design/Sacado_Fad_Exp_ViewFad.hpp
@@ -124,6 +124,25 @@ namespace Sacado {
     typedef U type;
   };
 
+  //! Specialization of %ScalarType to ViewFad types
+  /*!
+   * This specialization overrides the one for GeneralFad to handle const
+   * value types so that resulting scalar type is still const.
+   */
+  template <typename ValueT, unsigned Size, unsigned Stride, typename Base>
+  struct ScalarType< Fad::Exp::ViewFad<ValueT,Size,Stride,Base> > {
+    typedef typename ScalarType<ValueT>::type type;
+  };
+
+  /*!
+   * This specialization overrides the one for GeneralFad to handle const
+   * value types so that resulting value type is still const.
+   */
+  template <typename ValueT, unsigned Size, unsigned Stride, typename Base>
+  struct ValueType< Fad::Exp::ViewFad<ValueT,Size,Stride,Base> > {
+    typedef ValueT type;
+  };
+
 } // namespace Sacado
 
 #endif // SACADO_FAD_EXP_VIEWFAD_HPP


ikalash:
Seems there is a Jenkins error causing failures with the AutoTester.  Is there a way to circumvent these?  @trilinos/framework

etphipp:
There also appears to be an Intrepid2 test (Intrepid2_unit-test_Discretization_Integration_Cuda_Test_01_MPI_1) that keeps timing out on ride.  I did a build myself and it passed just fine.  Also, I don't think this test even uses Sacado (but maybe someone from @trilinos/intrepid2 can verify).

mperego:
> There also appears to be an Intrepid2 test [(Intrepid2_unit-test_Discretization_Integration_Cuda_Test_01_MPI_1)](https://testing.sandia.gov/cdash/testDetails.php?test=108868920&build=7144974) that keeps timing out on ride. I did a build myself and it passed just fine. Also, I don't think this test even uses Sacado (but maybe someone from @trilinos/intrepid2 can verify).

I'm sorry you hit by the ride/Intrepid2 curse. That is one of the few tests that are perfectly fine by themselves but they timeout in the PR Cuda build on ride. It seems to be due to machine load and the fact that tests timings change a lot with ctest -jN (see e.g. discussion here #7201). @prwolfe looked into it but as of now there does not seem to be a solution. Here are other PR affected: #7442 #3438 #7443 #7434.

mperego:
> > There also appears to be an Intrepid2 test [(Intrepid2_unit-test_Discretization_Integration_Cuda_Test_01_MPI_1)](https://testing.sandia.gov/cdash/testDetails.php?test=108868920&build=7144974) that keeps timing out on ride. I did a build myself and it passed just fine. Also, I don't think this test even uses Sacado (but maybe someone from @trilinos/intrepid2 can verify).
> 
> I'm sorry you got hit by the ride/Intrepid2 curse. That is one of the few tests that are perfectly fine by themselves but they timeout in the PR Cuda build on ride. It seems to be due to machine load and the fact that tests timings change a lot with ctest -jN (see e.g. discussion here #7201). @prwolfe looked into it but as of now there does not seem to be a solution. Here are other PR affected: #7442 #3438 #7443 #7434.

and yes, that test does not use Sacado.. and it has been there since years..

prwolfe:
> Seems there is a Jenkins error causing failures with the AutoTester. Is there a way to circumvent these? @trilinos/framework

I'm afraid not. Jenkins is on our critical path for autotesting.

ikalash:
So how do we get this merged in if we have no control over the jenkins failure and it’s not caused by this PR?

prwolfe:
> So how do we get this merged in if we have no control over the jenkins failure and it’s not caused by this PR?

The CEE team has been notified of the failure.  I have not yet heard where they are at with the issue.

ikalash:
Ok.  I guess I will stay tuned.

ikalash:
Seems the auto-tester is back to normal - my Piro PR just went through.  Can we restart the tests for this PR?  @trilinos/framework

etphipp:
I removed and readded the "AT: RETEST" label, so maybe the autotester will start again.

etphipp:
A different intrepid2 test timed out this time.  I guess we just keep rerunning the autotester until we get lucky?

etphipp:
The `Intrepid2_unit-test_Discretization_Integration_Cuda_Test_01_MPI_1` timed out again.  At this rate, this PR is never going to be merged.  What if we tried making the Intrepid2 Cuda tests run one at a time by adding `RUN_SERIAL`to their `TRIBITS_ADD_EXECUTABLE_AND_TEST` calls?  I did this years ago for several Sacado tests to improve their throughput (see, e.g., [here](https://github.com/trilinos/Trilinos/blob/442a24ba155050971e7cba30fa2c22fdda9ebce9/packages/sacado/test/UnitTests/CMakeLists.txt#L218)).  Thoughts @trilinos/intrepid2?

mperego:
> 
> 
> The `Intrepid2_unit-test_Discretization_Integration_Cuda_Test_01_MPI_1` timed out again. At this rate, this PR is never going to be merged. What if we tried making the Intrepid2 Cuda tests run one at a time by adding `RUN_SERIAL`to their `TRIBITS_ADD_EXECUTABLE_AND_TEST` calls? I did this years ago for several Sacado tests to improve their throughput (see, e.g., [here](https://github.com/trilinos/Trilinos/blob/442a24ba155050971e7cba30fa2c22fdda9ebce9/packages/sacado/test/UnitTests/CMakeLists.txt#L218)). Thoughts @trilinos/intrepid2?

@etphipp, we are looking into that option here #7476.

etphipp:
Yay!  Looks like setting RUN_SERIAL in the intrepid2 and muelu tests worked.  Thanks @mperego  and @brian-kelley!

ikalash:
Finally!!!  I was starting to think this PR would never go in!

bartlettroscoe:
> Yay! Looks like setting RUN_SERIAL in the intrepid2 and muelu tests worked. Thanks @mperego and @brian-kelley!

@etphipp, no you just got lucky I think.  See https://github.com/trilinos/Trilinos/issues/3276#issuecomment-643546942.

Issue #7451, opened 2020-05-29
Belos: FPEs in Albany/LCM tests


ikalash:
A number of FPEs started showing up in some Albany/LCM tests on 5/20.

The changes are deep within Belos and appear related to changes made to Belos not long ago (b/w 5/6-5/20 - last successful run was on 5/6, and the tests did not run b/w 5/6-5/20 unfortunately).  The output from the failures can be found here:
https://sems-cdash-son.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=2177

No changes were made to Albany/LCM when these issues started, so they are due to changes to Trilinos.  Can someone please have a look?

@trilinos/belos 

Tagging @lxmota , the owner of Albany/LCM.

cgcgcg:
@ikalash EMPIRE saw some FPEs with PR #7366, but we're not sure whether that was just an issue in EMPIRE exposed by the change from DKGS to ICGS in Belos. Any chance you could revert that PR an check if that fixes things?

ikalash:
Indeed reverting that PR makes the FPEs go away.

jennloe:
@hkthorn and I are looking into this.

jhux2:
@bathmatt 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cgcgcg:
@ikalash Can this issue be closed?

ikalash:
@cgcgcg : I don't know.  The owner of the machines (@lxmota) where these failures were happening is on vacation this week and there are other errors plaguing those builds.  Can you please keep this open until we've had a chance to verify that the failures are indeed resolved (next week)?

cgcgcg:
@ikalash Of course!

ikalash:
Thanks for understanding.  Sorry about the delay!

lxmota:
These FPE errors have been resolved.

PR #7454, opened 2020-05-29
Fixes calls to normalize zero vectors in ICGS/IMGS

hkthorn:
@trilinos/belos 

## Motivation
The logic that checks if the norm of a vector is zero before normalizing
it needs to be inserted to avoid FPEs and bad solver behavior.  This conditional 
existed in DGKS, which is why this did not come up as a problem before the 
change to making ICGS[2] the default.

## Related Issues

#7451 

## Testing

The OrthoManager test has been amended to include a check for this
functionality, so that it does not come up again for customers.


ikalash:
I've verified that this PR does indeed fix the FPEs in Albany/LCM.  Thanks @hkthorn !  We do have some exodiff diff's still, but that could be due to having tolerances that are too tight for exodiff.  Will discuss with @lxmota .

lxmota:
Merge, so we can pull and test?

ikalash:
I just merged it into develop, yes.

PR #7456, opened 2020-05-30
Piro: bug fix required for transient forward sensitivities to work in Albany and extension of observer to observe dx/dp

ikalash:
This PR contains a small bug fix required for transient forward sensitivities to work correctly in Albany.

It also contains some extensions and corresponding tests to allow observation of transient sensitivities dx/dp.


ikalash:
Thanks for approving, Mauro.  There is a test that executes the code that was causing the fail in Piro.  The problem is it executes it in a different way - xdotdot is not null in the test whereas it is when called from Albany.  This is why I only caught the problem when running from Albany.  I actually spent some time trying to force the test to have the same behavior as Albany but was unsuccessful.  I may look at it again later - I am working on other additions to Piro (will be a separate PR) involving observers, which are relevant to this.  I'd like to get this in first, so that I can push the Albany fwd sensitivity code/tests.

ikalash:
@trilinos/framework : what is the holdup on getting this merged in?  This PR seems to be stuck in one of the preinspection steps.

ikalash:
It seems the pre-merge inspection step has hung.  Can I restart it somehow?  @trilinos/framework

ikalash:
@mperego : could you please re-review this?  I ended up adding my dx/dp observer extensions to this PR since I finished it, rather than waiting to do that in a separate PR.  I added tests to Piro to test the new observer routines.

ikalash:
Sorry about doing the merge.  I'll remove the auto-merge label as you suggest.

PR #7457, opened 2020-05-30
Zoltan2: Add Sphynx build infrastructure

seheracer:
@trilinos/zoltan2 @egboman @kddevin 

## Motivation

This PR splits Zoltan2 into two subpackages: Core and Sphynx. Core consists of the src, test, and example directories of the original Zoltan2 package, whereas Sphynx is empty at the moment and will have the source code after the SNL copyright is complete. The subpackage design follows the Tpetra subpackage design. Zoltan2_ENABLE_TESTS:BOOL=ON enables the tests in subpackages, which refers to the tests of Core at the moment. Sphynx is not enabled by default, one should use Trilinos_ENABLE_Zoltan2Sphynx:BOOL=ON to enable it. Sphynx depends on Anasazi and Belos (unlike Core dependencies), so enabling Sphynx enables Anasazi and Belos.

## Stakeholder Feedback

No changes to current users. The only change is that libzoltan2.a is now placed in `${BUILD_DIR}/packages/zoltan2/core/src/` and previously it was in `${BUILD_DIR}/packages/zoltan2/src/`.


seheracer:
ParMETIS and Scotch tests are failing because of the changes in the CMake files. For example, enabling TPL_ENABLE_ParMETIS does not turn HAVE_ZOLTAN2_PARMETIS on. I will find a way to fix this.

seheracer:
Closed due to the fact that this subpackage design will not resolve the circular dependency problem we will face when we add MueLu to the required depencencies of Sphynx.

Issue #7112, opened 2020-04-02
Add RUN_SERIAL test property to many ATDM Trilinos tests on CUDA builds to address timeouts


bartlettroscoe:
I need to add option `<fullTestName>_SET_RUN_SERIAL=[ON|OFF]` to `TRIBITS_ADD[ADVANCED_]TEST()` to add the CTest `RUN_SERIAL` test property and use to mark many CUDA tests in Trilinos GitHub issues #7090, #6805, #6804, #6801

But first, we need to finish up and merge #6840 to first get the work spread out over the GPUs to see if that resolves some of these timeouts.

bartlettroscoe:
The issues with timeouts in the ATDM Trilinos builds was resolved by using the new ctest GPU allocation approach in #6840 so we don't need to add support for `<fullTestName>_RUN_SERIAL` for that ATDM Trilinos builds.  But we may need to add it for the Trilinos PR builds (see #7464).

bartlettroscoe:
NOTE: Using the CTest resource allocation approach does not help on 'ride' for some reason so we are still getting quite a few timeouts there.  Therefore, I will go ahead and implement this option and apply `RUN_SREIAL` to several of the tests on 'ride' to avoid those timeouts.

bartlettroscoe:
The updates from PR #8063 seem to be working well setting the `RUN_SERIAL` property.  See evidence in the PR #8063 (which was merged to 'atdm-nightly' a few days ago).  Also see no timeouts in the build ??? the last 3 days as shown here:

* https://testing.sandia.gov/cdash/index.php?project=Trilinos&begin=2020-09-01&end=2020-09-23&filtercount=1&showfilters=1&field1=buildname&compare1=61&value1=Trilinos-atdm-white-ride-gnu-7.2.0-openmp-debug

bartlettroscoe:
[...]

PR #7464, opened 2020-06-02
Tpetra: remove fence in add()

brian-kelley:
Remove fence, by moving the CRS row sorting to device.
This should also grant a big speedup, but I didn't measure it.
Fix bug in kokkos-kernels causing divide-by-zero in row sorting
(will copy fix to kokkos-kernels repo).

Fix sign compare warning in MatrixMatrix_Cuda.

@trilinos/tpetra 

## Motivation

Removes one of the Kokkos execution space fences in TpetraExt_MatrixMatrix_def.hpp. One step towards #7446 .

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of #7446 
* Composed of 

## Stakeholder Feedback

Part of @MicheldeMessieres 's work to eliminate need for CUDA_LAUNCH_BLOCKING and also to remove unnecessary fences.
## Testing

- ran all Tpetra tests with launch blocking on
- ran just the MatrixMatrix unit tests with launch blocking off

### DIFF
--- a/packages/kokkos-kernels/src/common/KokkosKernels_SparseUtils.hpp
+++ b/packages/kokkos-kernels/src/common/KokkosKernels_SparseUtils.hpp
@@ -989,7 +989,9 @@ void sort_crs_matrix(const rowmap_t& rowmap, const entries_t& entries, const val
     //TODO (probably important for performnce): add thread-level sort also, and use that
     //for small avg degree. But this works for now.
     int teamSize = 1;
-    lno_t avgDeg = (entries.extent(0) + numRows - 1) / numRows;
+    lno_t avgDeg = 0;
+    if(numRows)
+      avgDeg = (entries.extent(0) + numRows - 1) / numRows;
     while(teamSize * 2 * 2 <= avgDeg)
     {
       teamSize *= 2;
@@ -1041,7 +1043,9 @@ void sort_crs_graph(const rowmap_t& rowmap, const entries_t& entries)
     //TODO (probably important for performnce): add thread-level sort also, and use that
     //for small avg degree. But this works for now.
     int teamSize = 1;
-    lno_t avgDeg = (entries.extent(0) + numRows - 1) / numRows;
+    lno_t avgDeg = 0;
+    if(numRows)
+      avgDeg = (entries.extent(0) + numRows - 1) / numRows;
     while(teamSize * 2 * 2 <= avgDeg)
     {
       teamSize *= 2;
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_Cuda.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_Cuda.hpp
@@ -653,7 +653,7 @@ void KernelWrappers2<Scalar,LocalOrdinal,GlobalOrdinal,Kokkos::Compat::KokkosCud
     Teuchos::Array<Scalar> diagonal(diagLength);
     diags.get1dCopy(diagonal());
 
-    for(LocalOrdinal i = 0; i < diagLength; ++i) {
+    for(size_t i = 0; i < diagLength; ++i) {
       TEUCHOS_TEST_FOR_EXCEPTION(diagonal[i] == Teuchos::ScalarTraits<Scalar>::zero(), 
 				 std::runtime_error, 
 				 "Matrix A has a zero/missing diagonal: " << diagonal[i] << std::endl <<
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp
@@ -813,10 +813,7 @@ add (const Scalar& alpha,
                               col_inds_array, global_col_inds_array,
                               typename map_type::local_map_type>
         (localColinds, globalColinds, CcolMap->getLocalMap()));
-    //TODO: use KokkosKernels batched sort on device as soon as it's available
-    //But now, have to sort on host (using UVM)
-    exec_space().fence();
-    Tpetra::Import_Util::sortCrsEntries(rowptrs, localColinds, vals);
+    KokkosKernels::Impl::sort_crs_matrix<exec_space, row_ptrs_array, col_inds_array, values_array>(rowptrs, localColinds, vals);
     C.setAllValues(rowptrs, localColinds, vals);
     C.fillComplete(CDomainMap, CRangeMap, params);
     if(!doFillComplete)


brian-kelley:
Hmm, TpetraCore_gemm_m_eq_2_MPI_1 and Intrepid2_unit-test_Discretization_Integration_Cuda_Test_01_MPI_1 are consistently timing out on RIDE. This is a release build, but Kokkos_ENABLE_DEBUG is ON. @bartlettroscoe Would it make sense to shrink or disable these tests if Kokkos_ENABLE_DEBUG?

bartlettroscoe:
> Would it make sense to shrink or disable these tests if Kokkos_ENABLE_DEBUG?

@brian-kelley, likely a good idea.

NOTE: The new ctest resource allocation stuff has worked very well on 'waterman' and 'vortex' but not on 'ride' for some reason.  See https://github.com/trilinos/Trilinos/pull/6840#issuecomment-631641270.

kddevin:
Is Kokkos_ENABLE_DEBUG on in every build (RELEASE and DEBUG)?

brian-kelley:
In a similar build on my workstation, the Tpetra test takes 17 seconds and the Intrepid2 test takes 0.8 seconds. I guess the only explanation is that the RIDE testing is trying to run multiple tests at the same time on the same GPU. Neither of these even uses MPI (although something appears to be wrong with MPI on RIDE, since this gets printed):
```
ibv_exp_query_device: invalid comp_mask !!! (comp_mask = 0x1003d7b5740 valid_mask = 0x1)
[ride11][[64127,1],2][btl_openib_component.c:1648:init_one_device] error obtaining device attributes for mlx5_0 errno says Invalid argument
WARNING: There was an error initializing an OpenFabrics device.

  Local host:   ride11
  Local device: mlx5_0
[ride11][[64127,1],0][btl_openib_proc.c:330:mca_btl_openib_proc_get_locked] 0: error exit from mca_btl_openib_proc_create
[ride11][[64127,1],3][btl_openib_proc.c:330:mca_btl_openib_proc_get_locked] 3: error exit from mca_btl_openib_proc_create
[ride11][[64127,1],1][btl_openib_proc.c:330:mca_btl_openib_proc_get_locked] 1: error exit from mca_btl_openib_proc_create
[ride11][[64127,1],3][btl_openib_proc.c:330:mca_btl_openib_proc_get_locked] 3: error exit from mca_btl_openib_proc_create
[ride11][[64127,1],1][btl_openib_proc.c:330:mca_btl_openib_proc_get_locked] 1: error exit from mca_btl_openib_proc_create
[ride11][[64127,1],0][btl_openib_proc.c:330:mca_btl_openib_proc_get_locked] 0: error exit from mca_btl_openib_proc_create
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ride11 and rank 0!
```

bartlettroscoe:
> The only explanation I can think of is that the RIDE testing is trying to run multiple tests at the same time on the same GPU.

Yes, that is what is happening. For most tests, doubling up on the same GPU is not so bad.  It is just that some tests, like these `TpetraCore_gemm` tests.  I did some more detailed analysis and timing studies of these tests as part of the testing of PR #6840 on multiple GPU platforms.

It would be great if Trilinos did not have to rely on a system like 'ride' for CUDA PR testing.  (But they were not able to find a system after a year so provided them a configuration for 'ride' and here we are.)

brian-kelley:
@bartlettroscoe OK. Has anyone tried to use CUDA MPS (multi-process service) to share a GPU across tests? It would probably be a lot of work to set up but it would cut down most of the overhead from the sharing.

brian-kelley:
Anyway, I don't want to disable those tests since they clearly aren't taking too long under normal circumstances.

bartlettroscoe:
The Trilinos CUDA PR build could reduce the parallel level from whatever it is now to a lower level?  I don't know where this is set in the Trilinos PR tester.

kddevin:
@trilinos/framework @jwillenbring Please see @bartlettroscoe 's comment above on reducing the level of parallelism in CUDA PR testing.

bartlettroscoe:
> The Trilinos CUDA PR build could reduce the parallel level from whatever it is now to a lower level?

From some of my experiments, running with a lower `ctest -j<N>` level will reduce individual test timeouts but it may significantly increase the wall clock time to run the test suite.  That is the tradeoff.

The other option is to set the `RUN_SERIAL` ctest property on some of these tests and force ctest to run them one-at-at-time.  There is no TriBITS option for this but one could be added.

brian-kelley:
@bartlettroscoe Doesn't TriBITS have RUN_SERIAL? e.g.
```
    TRIBITS_ADD_TEST(
      Performance-CGSolve
      NAME Performance_StrongScaling_CGSolve
      ARGS "--size=200"
      COMM mpi
      NUM_MPI_PROCS 1
      STANDARD_PASS_OUTPUT
      RUN_SERIAL
      CATEGORIES PERFORMANCE
    )
```

brian-kelley:
I totally forgot about RUN_SERIAL even though I used it for the performance tests. I just opened #7476 to add it to these tests.

bartlettroscoe:
> Doesn't TriBITS have RUN_SERIAL? e.g.

@brian-kelley, yes, but we don't want to hard-code that in the CMakeList.txt file for all systems.  instead, we want the ability to set that on a system-by-system basis by setting:

```
set(<full_test_name>_RUN_SERIAL ON CACHE BOOL ... )
```

in the the system-specific configuration file, in this case:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/PullRequestLinuxCuda9.2TestingSettings.cmake

For example, using the new ctest GPU allocation method, we get perfect scalability going from `ctest -j1` to `ctest -j4` for these `TpetraCore_gemm` tests on the 'ats2' system 'waterman' (see https://github.com/trilinos/Trilinos/pull/6840#issuecomment-628732349).   It would be as shame (and a waste of hardware) to run a bunch of tests one-at-at-time on a system where we can run them just file 4-at-at-time.  We only want to set the `RUN_SERIAL` property for systems where the ctest GPU allocation method does not work well (which, unfortunately, is 'ride').  Make sense?

See #7112

bartlettroscoe:
> I totally forgot about RUN_SERIAL even though I used it for the performance tests. I just opened #7476 to add it to these tests.

For now that is okay, but on systems like 'ats2' and 'waterman' where we can have ctest spread out these tests to the different GPUs, that will be leaving GPUs sitting idle.

bartlettroscoe:
> Has anyone tried to use CUDA MPS (multi-process service) to share a GPU across tests? It would probably be a lot of work to set up but it would cut down most of the overhead from the sharing.

@brian-kelley, there has been talk of that in the past but no one has ever gone through the trouble to set this up.

prwolfe:
> > I totally forgot about RUN_SERIAL even though I used it for the performance tests. I just opened #7476 to add it to these tests.
> 
> For now that is okay, but on systems like 'ats2' and 'waterman' where we can have ctest spread out these tests to the different GPUs, that will be leaving GPUs sitting idle.

A few notes:
1) I am unable to recreate the errors when running by hand. My suspicion is ordering differences as test does look at past runs to set order.
2) not all failures print the MPI error you are noting.
3) I searched all test failures in PR testing on ride for the past week or so (https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-05-29&end=2020-06-03&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=cuda&field2=buildname&compare2=63&value2=pull&field3=status&compare3=61&value3=Failed) and this test is the top contender, but there are a 12 tests in that list. 
4) we are running on th Firestone nodes (1 k80 GPU) and I have a PR (#7473) to change that to the Garrison nodes. It will have to be force pushed after approval as it hits these same test failures. 
5) I have tested reducing the -j<N> going in to ctest with no real improvement (I went as low as 6 before the slowdown started to be highly noticeable. 
7) we are looking at moving to vortex if we can find the TPL support to do it.

I like the idea of setting them to run serial in this PR testing as soon as we can. I will continue to work the other angles as well. Note that everything on RIDE (builds, tests, etc.) is slower since the certificate errors over the weekend, so running everything in serial was not on my agenda.

Paul

kddevin:
@brian-kelley Thank you for this fix and performance enhancement.
Do we have any timing numbers demonstrating the old vs new performance?

brian-kelley:
@kddevin 
Here's a 1-rank run on two synthetic 1M row, 1M column matrices, each with 27 entries per row. These timings were very consistent from run to run. This was timed directly with Kokkos timers.
```
Before: 0.0136, after: 0.00433 (3.1x speedup)
```

I also tried measuring it in StackedTimer and running the MiniEM performance tests. It turns out this code path runs only once during the entire problem, on a small coarsened matrix, so the sorting is almost negligible:
```
4 ranks: before 2.69e-4, after 7.72e-5 (3.5x speedup)
16 ranks: before 4.92e-4, after 2.39e-4 (2.1x speedup)
```

Even though MiniEM hardly spends any time doing add(), I think other apps like Drekar do.

kddevin:
3x -- Very nice!  Thank you for documenting the speed-up for us, @brian-kelley

Issue #7462, opened 2020-06-02
Belos:  BiCGSTAB with more than 1 vector in the RHS


helebloch:
## Bug Report
@trilinos/\<Belos\>

### Description
I'm trying to solve a non symmetric matrix using the BiCGSTAB solver. The RHS is a multivector, one column of this mutivector is constant to 0, the other columns are constant to a non-zero value. The initial guess is set to 0. When I'm trying to solve the system, I got the following error:
```
Belos::MultiVecTraits::Assign(A, mv): Input multivector 'A' has 2 columns, but output multivector 'mv' has only 1 columns.
```
However, if all the columns in the RHS are non-zero, the solver manages to converge. The error appears with different matrices, and seems independent to the number of MPI processes.

[Here is an example of a code that reproduces the error. ](https://github.com/trilinos/Trilinos/files/4718218/main.txt)

Thanks in advance!

### Steps to Reproduce
1. SHA1:  55a75997332636a28afc9db1aee4ae46fe8d93e7 and 7ed20167dccbcb2e50f4bd943df9e4cf97bd0a0d

cgcgcg:
@trilinos/belos

amklinv-nnl:
I was able to reproduce the issue.  It appears to be crashing on the line
`MVT::Assign(*lp_->getInitResVec(), *Rhat_);`
in
`void BiCGStabIter<ScalarType,MV,OP>::initializeBiCGStab(BiCGStabIterationState<ScalarType,MV>& newstate)`
because the initial residual has two vectors, but Rhat only has 1.  I'll see what I can do.

hkthorn:
@amklinv-nnl  I'm sure you already know this, but this happens when the first RHS converges, but the second is not converged yet.  So, there is an issue with reseting the solver after the deflation of the converged RHS.

amklinv-nnl:
Yup.  I should have a PR for you in a little bit.  (Trilinos doesn't build so fast on my laptop.)

hkthorn:
Cool, thanks!

amklinv:
Okay, that fixes @helebloch's code for me.  Granted, I didn't run the Belos test suite (because I'm not sure my laptop can handle it). It might be worth adding a test that uses multiple right hand sides.

hkthorn:
Thanks @amklinv-nnl !  I have a multiple RHS test that I concocted to reproduce the error.  I will commit it as a follow up PR.

PR #7472, opened 2020-06-03
Adds tests to Belos BiCGStab for multiple RHS solves

hkthorn:
@trilinos/belos 

## Motivation

Adds tests to Belos BiCGStab for multiple RHS solves
With respect to issue #7462, multiple RHS solves would cause a
premature exit of the solver if one RHS converged before the second
one.  The deflation of the RHS was not correctly handled by the
iteration, which as been fixed.

## Testing

This PR provides tests to ensure that the fix for #7462 is correct and does not reoccur.


amklinv:
Is this test guaranteed to have vectors converging at different iterations?

PR #7473, opened 2020-06-03
PullRequestLinuxCudaDriver.sh: change from Firestone to Garrison nodes

prwolfe:
I am not sure this is out actual issue and will monitor
to see what results we get.

@trilinos/framework 

## Motivation
Unreproducable timeouts on the ride platform

### DIFF
--- a/cmake/std/PullRequestLinuxCudaDriver.sh
+++ b/cmake/std/PullRequestLinuxCudaDriver.sh
@@ -14,10 +14,11 @@ fi
 set -x
 
 #TODO: review appropriate job size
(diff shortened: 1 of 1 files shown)


cgcgcg:
@prwolfe Is it possible that this autotester failure https://github.com/trilinos/Trilinos/pull/7493#issuecomment-640971589 is caused by this? Looks like all tests fail, since compiled for wrong architecture.

prwolfe:
Yes – that was my mistake as I forgot to change the build. I reverted and will be testing the build changes in a fork.

My apologies.

PR #7476, opened 2020-06-03
Tpetra,Intrepid2: Adding RUN_SERIAL to tests

brian-kelley:
These were timing out on RIDE PR testing. Even with Kokkos_ENABLE_DEBUG=ON, the Tpetra tests should take about 17 seconds each and the Intrepid2 test 0.8 seconds. So this won't increase the total time by much.

@trilinos/tpetra 
@trilinos/intrepid2 

## Motivation

See test timeouts/discussion at #7464


mperego:
@brian-kelley Thanks for doing this, but I think we are going around the issue instead of addressing it. If you turn this off you might have other tests randomly timing out. Also we probably want to test their execution on Cuda. It really seems to be an issue with work load and using ctest -jN. @trilinos/framework .

brian-kelley:
@mperego I agree this is just a bandaid over the problem, but it will only be temporary until we can put e.g.
```
set (TpetraCore_gemm_m_eq_2_MPI_1_RUN_SERIAL ON CACHE BOOL "Temporary serialization for CUDA PR testing to avoid timeouts")
```
in ``cmake/std/PullRequestLinuxCuda9.2TestingSettings.cmake``. If we let the autotester run on this PR, we'll at least see if any tests still time out.

mperego:
> 
> 
> @mperego I agree this is just a bandaid over the problem, but it will only be temporary until we can put e.g.
> 
> ```
> set (TpetraCore_gemm_m_eq_2_MPI_1_RUN_SERIAL ON CACHE BOOL "Temporary serialization for CUDA PR testing to avoid timeouts")
> ```
> 
> in `cmake/std/PullRequestLinuxCuda9.2TestingSettings.cmake`. If we let the autotester run on this PR, we'll at least see if any tests still time out.

I see, thanks. B.t.w. we could probably run serially these tests :
https://github.com/trilinos/Trilinos/blob/develop/packages/intrepid2/unit-test/Projection/Cuda/CMakeLists.txt
They take long time, and timeouts increased significantly after I pushed those tests.

brian-kelley:
@mperego Good call, 3 out of 4 of the Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection tests took longer than 8 minutes:
```
Intrepid2 | Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection_HEX_MPI_1 | Passed | 8m 21s 680ms | 8m 21s 680ms | Intrepid2 | Summary | 1
Intrepid2 | Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection_QUAD_MPI_1 | Passed | 9m 4s 440ms | 9m 4s 440ms | Intrepid2 | Summary | 1
Intrepid2 | Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection_TET_MPI_1 | Passed | 1m 17s | 1m 17s | Intrepid2 | Summary | 1
Intrepid2 | Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection_TRI_MPI_1 | Passed | 8m 26s 700ms | 8m 26s 700ms | Intrepid2 | Summary | 1
```

mperego:
Yeah, the HEX test is the really expensive one, followed by the TET one. The timings you see for the other two are really a consequence of running them in parallel.

brian-kelley:
@mperego Should I only do HEX and TET in serial?

mperego:
> 
> 
> @mperego Should I only do HEX and TET in serial?

Yes, thanks.

brian-kelley:
@mperego You were right, running certain tests in serial just moved the timeout to a different test (MueLu_UnitTestsIntrepid2Tpetra_MPI_1). So we probably have to reduce the ``-j`` value for RIDE, rather than trying to play whack a mole with RUN_SERIAL on specific tests that may time out.

mperego:
@brian-kelley 
Even when run serially, the tests timings change significantly:
`Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection_TET_MPI_1` took 7m 29s 280ms 	the first time and 5m 14s 880ms.

I think I also was wrong.. the QUAD and TRI tests took also a long time, while they took less than 30s when I run it on ride serially. Maybe we should also run them serially, if we want to continue with this approach,.. and the `MueLu_UnitTestsIntrepid2Tpetra_MPI_1`, that test also has been timing out a lot lately.

etphipp:
So is there a timeframe where this is expected to be resolved?  Looking at all of the comments for this PR and #7464, it sure seems like there isn't.    Some ideas:
* Can we make timeouts non-failures?
* The latest version of CTest has this option that could help:
```
--repeat <mode>:<n>

    Run tests repeatedly based on the given <mode> up to <n> times. The modes are:

    until-fail
        Require each test to run <n> times without failing in order to pass. This is useful in finding sporadic failures in test cases.

    until-pass
        Allow each test to run up to <n> times in order to pass. Repeats tests if they fail for any reason. This is useful in tolerating sporadic failures in test cases.

    after-timeout
        Allow each test to run up to <n> times in order to pass. Repeats tests only if they timeout. This is useful in tolerating sporadic timeouts in test cases on busy machines.
```
Otherwise, the only options I see are to start force-pushing PR's that are only being held up by timeouts, or shutting down PR testing on Ride altogether.

mperego:
@brian-kelley, I would propose to use the RUN_SERIAL option also for
```
Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection_QUAD_MPI_1`
Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection_TRI_MPI_1
MueLu_UnitTestsIntrepid2Tpetra_MPI_1
```
This is of course a temporary solution but I think it will help the several PRs currently on hold.

@etphipp suggestion makes sense to me, but it would probably need to be discussed longer, and we have been on hold since several days now.

brian-kelley:
@mperego OK, I added those RUN_SERIALs.

@prwolfe has opened #7473 to move RIDE's PR testing to the Pascal nodes (rhel7F -> rhel7G), which have 2 GPUs each, and they are individually about twice as fast. Hopefully that will be a permanent fix for these timeouts, and if this gets merged first, we can just roll this back after #7473.

mperego:
> 
> 
> @mperego OK, I added those RUN_SERIALs.
> 
> @prwolfe has opened #7473 to move RIDE's PR testing to the Pascal nodes (rhel7F -> rhel7G), which have 2 GPUs each, and they are individually about twice as fast. Hopefully that will be a permanent fix for these timeouts, and if this gets merged first, we can just roll this back after #7473.

It seems you added RUN_SERIAL to  
MueLu_UnitTestsIntrepid2Tpetra_MPI_4
but not
MueLu_UnitTestsIntrepid2Tpetra_MPI_1

brian-kelley:
@mperego Oops, fixed. I left it on the 4-rank as well, just in case it's slow too.

@trilinos/muelu There's probably something that can be done to the Intrepid2+Tpetra+MueLu unit tests to speed them up, since 8m36s is spent in just 2 of the unit tests: ``GenerateRepresentativeBasisNodes_HEX_Equispaced`` and ``GenerateRepresentativeBasisNodes_QUAD_Spectral``.

mperego:
Thanks @brian-kelley, fingers crossed.
I still don't understand why tests that take less than 30s when I run them on ride, they take 9m, even when run with RUN_SERIAL.

mperego:
Good. Cuda build is clean. :-)
B.t.w. this time
`Intrepid2_unit-test_Projection_Cuda_Test_InterpolationProjection_TET_MPI_1` took 2m30s.. this is 3 times faster than the worts timing (all with RUN_SERIAL). Ride really seems unreliable.

mperego:
@brian-kelley can we merge this?

brian-kelley:
@mperego Sure, it should unblock the PRs like #7464

Issue #7175, opened 2020-04-15
MueLu: reduce communication in regionMG's sumInterfaceValues()


mayrmt:
## Enhancement
@trilinos/muelu 
@lucbv @rstumin @pohm01 

### Current Status

To sum values at region interfaces, we currently perform two rounds of communication:
1. Export from quasiRegion to composite format 
1. Export from composite to quasiRegion format (with `CombineMode Xpetra::ADD`)

Time measurements have identified these operations as bottlenecks in our V-cycle.

### Planned Actions

We intend to reduce this to a single communication action. So far, we plan to completely stay in region format and to collect off-process entries in a local buffer. Then, we will perform summation locally. For communication/gathering of entries, we intend to use some `Import()`/`Export()` mechanism. Summation needs to be implemented by ourselves.

### Additional Information

In region format, duplicated interface nodes reside at the same spatial location, but have unique global IDs (and also their DOFs have unique global IDs). So, an `Import()`/`Export()` should work just fine on these unique maps. 

Figuring out, which entries need to be summed up, is a task to be done by ourselves. We have this information already stored somewhere, as we rely on it during node duplication, i.e. when generation the region maps out of the quasiRegion maps.

lucbv:
OK, I think this has been achieved for the MatVec with the new implementation that we just merged with PR #7499 
@mayrmt do we have other "well known" spot in the vCycle where this approach can be reused to improve performance further?

mayrmt:
@lucbv Related to this, we could look into using `KokkosKernels` to exploit grid structuredness or at least to work with the raw data arrays when applyting the prolongator/restrictor. (There's no communication involved, so we don't need the "fast MatVec" there.)

lucbv:
What we could do is add a flag to fast MatVec if true do local operation only, otherwise do the whole thing: local + communication + add contribution from neighbors.
The flag can be defaulted to false so that we only need to pass the true value in the few places where it makes sense.

mayrmt:
@lucbv As discussed today, I’ve switched the level transfers to use the fast MatVec with appropriate flags. Tests are passing.  Will do some more testing and will create a PR probably tomorrow. 

I also want to look into another tiny issue, where we potentially copy large amounts of data instead of just paying a pointer to it.

mayrmt:
@lucbv To pass the `regionInterfaceLIDs` needed `ApplyMatVec` through the code, we store and extract them from a `ParameterList` as follows:
```
[...]

PR #7478, opened 2020-06-04
MueLu: region matvec, see issue #7175

lucbv:
@trilinos/muelu

## Motivation
This should accelerate significantly the matvec costs associated with the region format as it now performs only one neighbor communication instead of tow.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of #7175
* Composed of 

## Stakeholder Feedback
This work is part of an ASCR project with @rstumin and @lucbv 

## Testing
Local tests have been performed and so far all are passing except for elasticity problems since this PR only addresses the scalar equation case. A follow-up PR will tackle the multiple dof per node case, in the meantime, a runtime parameter can be selected to use the "slow path" for matvec that works with system of PDEs.

## Additional Information
Work remaining before this PR is complete:
1. add capabilities for multiple dofs per node
2. modify the matvec implementation to expose the Kokkos-kernel call which will allow to easily write higher performance matvec algorithm based on structured grid assumptions
3. create a matvec specific driver to assess the performance of various kernels implementations based on the structure of the grid

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionHierarchy_def.hpp
@@ -135,7 +135,11 @@ void MakeCoarseLevelMaps(const int maxRegPerGID,
                          Array<std::vector<RCP<Xpetra::Map<LocalOrdinal, GlobalOrdinal, Node> > > >& quasiRegRowMaps,
                          Array<std::vector<RCP<Xpetra::Map<LocalOrdinal, GlobalOrdinal, Node> > > >& quasiRegColMaps,
                          Array<RCP<Xpetra::Map<LocalOrdinal, GlobalOrdinal, Node> > >& compRowMaps,
-                         Array<std::vector<RCP<Xpetra::Import<LocalOrdinal, GlobalOrdinal, Node> > > >& regRowImporters) {
+                         Array<std::vector<RCP<Xpetra::Import<LocalOrdinal, GlobalOrdinal, Node> > > >& regRowImporters,
+                         Array<Teuchos::RCP<Xpetra::MultiVector<GlobalOrdinal, LocalOrdinal, GlobalOrdinal, Node> > >& interfaceGIDs,
+                         Array<Teuchos::RCP<Xpetra::MultiVector<LocalOrdinal, LocalOrdinal, GlobalOrdinal, Node> > >& regionsPerGIDWithGhosts,
+                         Array<Teuchos::Array<LocalOrdinal> >& regionMatVecLIDs,
+                         Array<Teuchos::RCP<Xpetra::Import<LocalOrdinal, GlobalOrdinal, Node> > >& regionInterfaceImporter) {
 
 #include "Xpetra_UseShortNames.hpp"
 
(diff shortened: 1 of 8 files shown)


mayrmt:
@lucbv I have completed the first steps towards multiple DOFs per node. Here's what I did:
- create unit test for FastMatVec using an "Elasticity2D" matrix
- adapt indexing in in `createRegionData()`, when computing the LIDs and GIDs of interface DOFs

I've verified the outcome of `createRegionData()` with a 7x7 nodes/4 regions mesh w/ 2 DOFs per node as sketched out here: 
IMG_3020_small
List of LIDs/GIDs are computed as expected.

When proceeding to call `SetupMatVec()`,  the lines
```c++
  Teuchos::Array<GO> regionMatVecGIDs;
  Array<ArrayRCP<const LO> > regionsPerGIDWithGhostsData(maxRegPerGID);
  Array<ArrayRCP<const GO> > interfaceGIDsData(maxRegPerGID);
  for(LO regionIdx = 0; regionIdx < maxRegPerGID; ++regionIdx) {
    regionsPerGIDWithGhostsData[regionIdx] = regionsPerGIDWithGhosts->getData(regionIdx);
    interfaceGIDsData[regionIdx] = interfaceGIDsPerGrp[0]->getData(regionIdx); /* That's the line causing trouble */
    for(LO idx = 0; idx < static_cast<LO>(regionsPerGIDWithGhostsData[regionIdx].size()); ++idx) {
      if((regionsPerGIDWithGhostsData[regionIdx][idx] != -1)
         && (regionsPerGIDWithGhostsData[regionIdx][idx] != myRank)) {
        regionMatVecLIDs.push_back(idx);
        regionMatVecGIDs.push_back(interfaceGIDsData[regionIdx][idx]);
      }
    }
  }
```
(Troubling line marked with comment)
results in this error:
```
 
  /ssd/codes/mayrmt_trilinos/src-trilinos/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp:2917:
  
  Throw number = 1
  
  Throw test that evaluated to true: (static_cast<size_t> (hostView_j.extent (0)) < static_cast<size_t> (dataAsArcp.size ()))
  
  Tpetra::MultiVector<long long,int,long long,Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>>::getData: hostView_j.extent(0) = 32 < dataAsArcp.size() = 64.  Please report this bug to the Tpetra developers.
```
According to the code in `packages/tpetra/core/src/Tpetra_MultiVector_def.hpp:2917:`, this error occurs during the `getData()` call.

Any ideas?

mayrmt:
Update: Above's error is independent of the choice of `<Scalar,LocalOrdinal,GlobalOrdinal>` template parameters. It occurs for both `<int,int,int>` and `<long long, int, long long>`.

lucbv:
@mayrmt will get a look at this tomorrow, hopefully this can be fixed easily?

mayrmt:
@lucbv I haven't pushed my changes yet. Since @rstumin might use this branch for performance testing, I don't want to break it.

Just let me know, if you need my current status of the code and how I should provide it.

PR #7482, opened 2020-06-04
Tacho cuda performance improvement + clean up.

kyungjoo-kim:
## Motivation

This includes two types of improvements
- Replacing fence with space.fence where it is applicable. Tacho use multiple cuda streams and there are some cases that I can fence individual streams where previously device synchronization is used. 
- Invert diagonals. To improve the solve phase, I attempted again to invert diagonals. Previsouly, when I first attemptted it, it does not effectively reduce the solve time. One reason I suspect is that ETI is the difference from my first trial. 

Clean up
- Delete old files that includes my previous development work (mostly trials).


kyungjoo-kim:
The error is not related to this PR and attempts to retest.

PR #7484, opened 2020-06-05
ATDM: cts1: Extend slrum test timeout from 1:20:00 to 3:00:00 (ATDV-357)

bartlettroscoe:
The slurm jobs running the tests all timed out with at timeout of 1:20:00.
Something might be wrong with this configuration that would cause such a large
increase in the test runtimes. But we can deal with that later.


bartlettroscoe:
This did not fix anything.  The fix is in PR #7468 

I reverted this change on the 'atdm-nightly' branch with commit https://github.com/trilinos/Trilinos/commit/4383eb748eaa3124bafb7854b15ebc82985d694b that I just pushed to the 'atdm-nightly-manual-updates' branch.

Closing this PR without merging.

PR #7487, opened 2020-06-06
Xpetra: fix binary reader

jhux2:
Tpetra requires a static graph.  To do this with file reading,
my solution is to make two passes through the file, the first pass
to figure out the sparsity pattern, the second to populate the matrix.
Because the input file is binary, hopefully this isn't too expensive.

@trilinos/xpetra

## Motivation

This is primarily for developers.

## Testing

Tested locally.

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/7487)


jhux2:
> Sure and I guess we do not plan on doing this for complex types so double is fine.
> Same goes with float I assume?

The reader never supported complex. I'm just making it work again.

PR #7490, opened 2020-06-06
Turn off STKBalance at configure time for Intel PR build if Tpetra-In…

alanw0:
…t is activated.

@trilinos/<teamName>


prwolfe:
> Looks like timeouts for the CUDA build. Might help if the @trilinos/framework turns down the ctest parallel level for the CUDA build on 'ride'.
> 
> NOTE: The new ctest GPU alllocation method (see PR #6840) did not seem to help runtimes on 'ride' at all :-(

I have been spending far to much time with ride of late and am not seeing good solutions. I have tried lower -j levels, but no joy. I was thinking of moving to vortex - don't you have some atom jobs there? My first naive attempt complained that it couldn't link C++ and fortran together.

PR #7496, opened 2020-06-09
Revert "PullRequestLinuxCudaDriver.sh: change from Firestone to Garrison nodes"

prwolfe:
Reverts trilinos/Trilinos#7473

I missed the required build changes. With a 9 hour cycle time and the need to set up testing on a fork this will take a bit to fix.


cgcgcg:
@prwolfe Thanks!

PR #7499, opened 2020-06-09
MueLu: fast region MatVec with multiple DOFs per node

mayrmt:
@trilinos/muelu 

## Motivation

This PR extends the fast MatVec with only one round of communication during the Apply phase to the case of multiple DOFs per node (as e.g. in elasticity problems).

## Related Issues

* Part of #7175 
* Follows #7478 

## Stakeholder Feedback

This work is part of an ASCR project with @rstumin and @lucbv.

## Testing
We've added a unit test and did manual testing and comparisons to the old MatVec implementation.

## Additional Information

A sketch of a 2D mesh and expected LID/GID numberings for testing are given in the predecessor of this PR in [this comment](https://github.com/trilinos/Trilinos/pull/7478#issuecomment-640070394).


mayrmt:
Setting the WIP label, as there are still some problems to be resolved...

mayrmt:
@lucbv @rstumin @pohm01 I ran some tests on 2D Elasticity and compared the "slow" and the "fast" MatVec. Results seem to be independent of the MatVec implementation. When @lucbv has added the missing interface communication for 3D problems, we can check 3D Laplace and 3D Elasticity.

lucbv:
@mayrmt my last commit fixed the 3D unit-tests.
I still want to run some more tests as the unit-test only use 1 and 4 MPI ranks.
I would like to bring that up to 27 and see that example work properly, it would be a more clear progress.

lucbv:
Also as a note, we have two regression tests:
1. Star2D_Tpetra_GS
2. Linear_Brick3D_Tpetra
both are passing now but we should add at least one regression test for elasticity.
Before this commit Linear_Brick3D was not passing which I thought was doing elasticity but it's really not : (

lucbv:
Alright did a quick rebase after doing some checks, this looks good now.

mayrmt:
@lucbv I did some more cleaning (debug output, consistent naming of tests) and performed tests with the driver and 3D elasticity. Old and new MatVec yield same results.

Feel free to review, approve, and merge.

mayrmt:
@lucbv I've created an Elasticity3D regression test.

lucbv:
@mayrmt if you feel confident I can approve.
I also did not find any problems with my own tests...
Let us see what Ray thinks of it when he runs this on Cori

PR #7186, opened 2020-04-16
Tpetra: batchedApply() prototype

csiefer2:
Tests: RHEL6

PR #7500, opened 2020-06-09
Adds example/test for multi-matrix solves using pseudo-block GMRES

hkthorn:
@trilinos/belos 

## Motivation

The pseudo-block solvers have the ability to solve a set of linear
systems simultaneously.  This has to be defined by the linear problem,
which will manage the active set of linear problems being solved and
ensure the correct operator is applied in the iterative solver.

Some of the methods in the Belos::LinearProblem class needed to be
virtual to enable this.  The example/test in the MultiMatrixSolve
directory of the Epetra interface shows how to create a new
LinearProblem class and solve an unpreconditioned set of linear systems.

## Testing

Tested using GCC9 in serial on OSX

 The example/test in the MultiMatrixSolve directory of the Epetra interface shows how to create a new LinearProblem class and solve an unpreconditioned set of linear systems.


jennloe:
Should the MultiShiftLinearProblem be moved to belos/src?

hkthorn:
It's an example of how to create a different linear problem, or set of related linear problems, and have Belos solve it.  By having it in the example/test directories it exhibits the extensibility of Belos for particular use cases.  If it was in the src directory, it would be confusing because users might think it is a core capability of Belos.

jhux2:
@hkthorn Would it make sense to have a corresponding Tpetra example?

hkthorn:
@jhux2 Good point, I will issue a separate pull request with such an example.

kddevin:
Just curious:  What is meant by "multi-matrix solve"?
@csiefer2 recently implemented a batched Matrix apply in Tpetra #7186 ; is it relevant to this case?

hkthorn:
@kddevin There is a preconceived notion that Belos solvers can only perform a simultaneous solve for a single matrix and multiple right-hand sides.  I fished an idea out of my head, which is in this PR, to illustrate using Belos pseudo-block solvers to perform linear solves for a set of linear systems, A_i*x_i = b_i, not just A*x_i = b_i for i=1,...,N.  Pseudo-block solvers and their rudimentary deflation techniques can enable linear solves for a set of linear systems, where the application of the operator and preconditioner are aggregated (or maybe batched).  This could be beneficial for users trying to solve a set of related linear systems, like shifted systems that result from frequency sweeps.  This idea has come up with Eiger/Gemma developers and in conversations with Eric Phipps, so it felt like an example of this capability could be useful.  I am not aware of the batched Matrix apply to which you speak.  @csiefer2 is there any synergy with this PR?

kddevin:
The batched apply communicates the input multivector X once, then applies a series of (identically mapped) matrices A_i to it yielding different output vectors Y_i = A_i X
It saves the redundant communication of X for each A_i X.

PR #7502, opened 2020-06-09
MueLu: fix compile error

jhux2:
Only manifests if no direct solver is enabled

@trilinos/muelu 

## Motivation

Fixes a compile error that you would only see if no direct solver is enabled.

## Testing

Confirmed on Summit with gcc 7.4.0 that this patch compiles.

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/7502)


lucbv:
Sneaky little bug!

PR #7506, opened 2020-06-09
Remove vvt from path

### LINKED ISSUE #7504: Remove Certain sparc-dev Modules
## Bug Report
@bartlettroscoe, @e10harvey

### Description
After the transition to the `sparc-dev` modules, sourcing the `load_matching_env.sh` that Trilinos drops as part of the install winds up loading SPARC modules that have nothing to do with Trilinos.  The concrete example I ran into was vvtest, which caused problems for us because it's the wrong version for EMPIRE.  We've already fixed things on our end to get rid of SPARC's vvtest and load our own, but it'd be nice to make sure the Trilinos configuration scripts aren't loading extraneous stuff that's not needed by Trilinos.

### Steps to Reproduce
```bash
$ ssh eclipse.sandia.gov
$ cd /projects/empire/installs/eclipse/INTEL-18.0.2_OPENMPI-4.0.3-RELEASE-OPENMP-STATIC/trilinos/latest
$ source load_matching_env.sh
Hostname 'eclipse-login6' matches known ATDM host 'eclipse-login6' and system 'cts1'
Setting compiler and build options for build-name 'EMPIRE-deploy-Trilinos-eclipse-intel-opt-openmp'
Using cts1 toss3 compiler stack INTEL-18.0.2_OPENMPI-4.0.3 to build RELEASE code with Kokkos node type OPENMP
This version of Open MPI 4.0.3 is temporary and was built with Intel/18.0 compilers and cudatoolkit/10.1


The following have been reloaded with a version change:
  1) intel/19.0 => intel/18.0.2.199
$ which vvtest
/projects/sparc/tools/vvt/vvtest
```

e10harvey:
## How was this tested?
```bash
$ source cmake/std/atdm/load-env.sh intel
Hostname <snip> matches known ATDM host <snip> and system 'cts1'
Setting compiler and build options for build-name 'intel'
Using cts1 toss3 compiler stack INTEL-18.0.2_OPENMPI-4.0.3 to build DEBUG code with Kokkos node type SERIAL
This version of Open MPI 4.0.3 is temporary and was built with Intel/18.0 compilers and cudatoolkit/10.1

$ which vvtest
/usr/bin/which: no vvtest in <snip>
```

Fixes #7504.


jmgate:
Removing automerge to await @bartlettroscoe's feedback.

e10harvey:
> The `atdm_remove_from_path` function looks fragile in bash, compared to the more comprehensive parsing of `PATH` that you could do with, e.g., Python. Nothing stands out as wrong; I'm just a little concerned if I poked it in the wrong way, I'd get bit.

@jmgate: Can you elaborate?

`sed` is POSIX compliant and I have switched `echo` to `printf`.

bartlettroscoe:
> The `atdm_remove_from_path` function looks fragile in bash

The purpose of this function is to change a bach var so it would seem very reasonable to implement the entire thing in bash.

cmake/std/atdm/utils/atdm_config_helper_funcs.sh
```diff
@@ -65,3 +65,14 @@ function get_sparc_dev_module_name() {
...
+#
+# Remove the "substring" from the PATH var.
+# @param substring: the string to remove.
+#
+function atdm_remove_from_path() {
+  local substring=$(printf "%s" "$1" | sed 's/\//\\\//g')
+  local path=$(printf "%s" "$PATH" | sed "s/$substring://g")
+  path=$(printf "%s" "$PATH" | sed "s/:$substring//g")
+  export PATH=$path
+}
```
bartlettroscoe, line 78: It might be more straightforward and safe to split this on ':' into a bash array, then loop over removing entries that match the input list, and putting it back together again?

bartlettroscoe: I wrote some very similar bash code in:

* https://github.com/trilinos/Trilinos/blob/master/commonTools/framework/get-changed-trilinos-packages-helpers.sh

I think if you replace 'comma' with 'colon' in these names and `,` with `:`, I think you have the same algorithm.  And thinking of this more, we could have:

```
atdm_remove_paths_from_path_dir PATH <path1> <path2> ...
```

And we can write some unit tests for that using the shunit2 framework.

Issue #7507, opened 2020-06-09
Muelu:  failed to compile with complex. 


kyungjoo-kim:
## Bug Report
@trilinos/muelu 

@crdohrm attempts to compile Trilinos for Sierra code. Compiling fails with the following error message.
```
ded-virtual -O2 -w --relocatable-device-code=true  -expt-extended-lambda -arch=sm_70  -DNDEBUG   -std=c++11 -o CMakeFiles/muelu-adapters.dir/ExplicitInstantiation/MueLu_RefMaxwell.cpp.o -c /vscratch1/crdohrm/Sierra/codeTI/objs/tpls/Trilinos/f600745ecce0ca83/packages/muelu/adapters/ExplicitInstantiation/MueLu_RefMaxwell.cpp
/vscratch1/crdohrm/Sierra/codeTI/TPLs_src/Trilinos/packages/kokkos/core/src/Cuda/Kokkos_Cuda_Parallel.hpp(609): error: calling a constexpr __host__ function("std::complex<double> ::complex") from a __device__ function("Kokkos::Impl::ParallelFor< ::,  ::Kokkos::RangePolicy<int,  ::Kokkos::Cuda > ,  ::Kokkos::Cuda> ::operator () const") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
 
/vscratch1/crdohrm/Sierra/codeTI/TPLs_src/Trilinos/packages/kokkos/core/src/Cuda/Kokkos_Cuda_Parallel.hpp(609): error: identifier "std::complex<double> ::complex" is undefined in device code 
```
The error message says that it requires --expt-relaxed-constexpr. However, the problem is somehow muelu instanciates with std::complex on cuda intead of using kokkos complex.

### Steps to Reproduce
1. SHA1: current master branch.
2. Configure script:  it enables complex on cuda and also enable relocatable device code.

jhux2:
@kyungjoo-kim @crdohrm Thanks for reporting.  Would it be possible to get a minimal cmake configure line?

lucbv:
@kyungjoo-kim trilinos does not need relocatable device flags so I am not sure why it is enabled. If I remember correctly though, there should be an ATDM build that enables it, so it should not create problems?
The complex issue I am sorry to say is not so surprising...

cgcgcg:
@kyungjoo-kim Did the patch I sent you do anything?

jhux2:
[...]

PR #7511, opened 2020-06-10
MueLu RefMaxwell: attempt fixing bug with cuda+complex

cgcgcg:
@trilinos/muelu 

## Motivation
Attempted fix for #7507


cgcgcg:
Let's see if this fixes the issue, at least the autotester seems to not dislike it..

PR #7512, opened 2020-06-10
Tpetra: Hierarchical unpack

tjfulle:
Adds hierarchical unpacking to `Tpetra::CrsMatrix::unpackAndCombine`.
@trilinos/tpetra

## Motivation
See #6603.  When a matrix has a few long rows, the unbalance between the rows leads to sever slowdown when unpacking data.  This PR adds hierarchical unpacking to matrix rows.  Rows are unpacked by teams of chunks and chunks are unpacked entrywise (in parallel).

## Related Issues
#6603

## Stakeholder Feedback
Lead to a 10x speed up in the unpack region of some simulations.

## Testing
All Tpetra tests pass.

## Interested parties:
@vbrunini


packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
```diff
@@ -399,6 +506,188 @@ struct UnpackCrsMatrixAndCombineFunctor {
...
+        }
+      },
+      errors
+    );
+
+    team_member.team_barrier();
+    tokens.release(token);
+
+    if (errors > 0) {
+      dst = Kokkos::make_pair(4, lid_no); // invalid combine mode
```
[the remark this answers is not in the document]
brian-kelley: Right, this is not safe, dst refers to the same location across the entire team. Using that single will fix it.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_PackUnpack.cpp
```diff
@@ -208,7 +208,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(CrsMatrix, PackThenUnpackAndCombine, SC, LO, G
   out << "Building second matrix" << endl;
   auto graph = A->getCrsGraph();
   RCP<crs_matrix_type> B (new crs_matrix_type (graph));
-  B->setAllToScalar(SC {});
+  B->setAllToScalar(SC{-1.});
```
kyungjoo-kim, line 211: Any particular reason to set scalars with -1 ?

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
```diff
@@ -172,6 +173,99 @@ unpackRow(const typename PackTraits<GO>::output_array_type& gids_out,
...
+///
+/// \tparam ST The type of the numerical entries of the matrix.
+///   (You can use real-valued or complex-valued types here, unlike
+///   in Epetra, where the scalar type is always \c double.)
+/// \tparam LO The type of local indices.  See the
+///   documentation of Map for requirements.
+/// \tparam GO The type of global indices.  See the
+///   documentation of Map for requirements.
+template<class ST, class LO, class GO, class ExecutionSpace>
+KOKKOS_FUNCTION int
```
kyungjoo-kim, line 186: Why do you use KOKKOS_FUNCTION instead of KOKKOS_INLINE_FUNCTION ? This is defined in a header file. If I remember correctly, KOKKOS_FUNCTION is meant for the case that we use relocatable device code (a device code is separately compiled in a separate object file from kernel launching code). @crtrott Could you make it clear about the kokkos_function usage ? I might be wrong.

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
```diff
@@ -340,7 +447,7 @@ struct UnpackCrsMatrixAndCombineFunctor {
     // is an integer in [0, tokens.size()).  It is used to grab a unique (to
     // this thread) subview of the scratch arrays.
     const size_type token = tokens.acquire();
-    const size_t a = static_cast<size_t>(token) * max_num_ent;
+    const size_t a = static_cast<size_t>(token) * batch_size;
```
kyungjoo-kim, line 450: So, now the flat unpack also use batch size as a parameter. What would happen if we run this flat unpack code where the batch size is 1. It will have some functional overhead but mapping of threads to workitems are almost same as the hierarchical version. Just wondering. I vote for removing flat unpacking.

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
```diff
@@ -399,6 +506,188 @@ struct UnpackCrsMatrixAndCombineFunctor {
...
+      dst = Kokkos::make_pair(4, lid_no); // unpack error
+      tokens.release(token);
+      return;
+    }
+
+    int errors = 0;
+    constexpr bool matrix_has_sorted_rows = true; // see #6282
+    Kokkos::parallel_reduce(
+      Kokkos::TeamThreadRange(team_member, num_entries_in_batch),
+      KOKKOS_LAMBDA(const LO& j, int& loc_errors)
```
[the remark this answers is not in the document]
kyungjoo-kim: Using a global atomic would be a good idea. If the code cannot recover from the error, we can still give the detailed information by printing and abort the code where the error happens.

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
```diff
@@ -529,6 +818,53 @@ compute_total_num_entries (
...
+  return static_cast<size_t>(num_ent_LO);
+}
+
+/// \brief Compute the index and batch number associated with each batch
+///
+/// batch_info(i, 0) is the local index of the ith batch
+/// batch_info(i, 1) is the local batch number of the ith batch
+template<class View1, class View2>
+KOKKOS_INLINE_FUNCTION
+bool
```
kyungjoo-kim, line 848: This looke like a host only function. Then, you can replace KOKKOS_INLINE_FUNCTION with inline.

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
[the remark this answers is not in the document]
tjfulle: @brian-kelley @kyungjoo-kim what does a global atomic variable look like?  I've never used one.

packages/tpetra/core/test/CrsMatrix/CrsMatrix_PackUnpack.cpp
tjfulle: only for easier debugging.

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
tjfulle: my mistake,

tjfulle:
Flat unpacking is gone.  Need to put in more reasonable defaults and global error counters.

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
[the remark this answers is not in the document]
kyungjoo-kim: You can probably do this way,
```
using error_type = int;
/// this error code is allocated on device and use atomic attribute
Kokkos::View<error_type,device_type,Kokkos::Atomic> err("error");
Kokkos::parallel_reduce(team_policy, KOKKOS_LAMBDA(const member_type &member) {
  Kokkos::parallel_for(team_range or team_vector_range, [&](const int &i) {
    // assume do something returns a error code.
    err += doSomething(i);
    // or you can just use Kokkos::atomic_fetch_and_add(err.data(), 
    const auto local_error = doSomething(i);
    Kokkos::atomic_fetch_and_add(err.data(), local_error);
  });
});
```

tjfulle:
@kyungjoo-kim - just finished up some more work.  can you please review again?

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
```diff
@@ -215,190 +212,204 @@ struct UnpackCrsMatrixAndCombineFunctor {
...
+          Kokkos::atomic_compare_exchange_strong(error_code.data(), 0, 31);
+        }
+      }
+    );
 
-    // Combine the values according to the combine_mode
-    const LO* const lids_raw = const_cast<const LO*> (lids_out.data ());
-    const ST* const vals_raw = const_cast<const ST*> (vals_out.data ());
-    LO num_modified = 0;
+    team_member.team_barrier();
```
kyungjoo-kim, line 401: The error code records the first error. Thus, after team barrier, you might want to check the error code and return immediately (another thread may already report an error). After this kernels is finished with a non-zero error code, a user can find the first problem without waiting for finishing this kernel.   

and result in early termination o

kyungjoo-kim: I am not sure how much checking the error code every time costs. Maybe checking the error code every time does not make sense. When error happens, who care if the code tells you that "you are wrong" earlier or not. I think that it can go as it is.

tjfulle:
@kyungjoo-kim - can you re-review the last push?  it's a 2 line addition to avoid having too large team size on non-cuda execution spaces.

PR #7529, opened 2020-06-15
Intrepid2: ensure that derived basis classes have unique names

### LINKED ISSUE #7071: Intrepid2: Some Bases Don't Have Unique Names
## Bug Report
@trilinos/intrepid2

### Description
As discussed in #7048, some of the new bases in Intrepid2 don't override the base class's `getName()` method.  Since some Intrepid2 methods rely on a unique name for caching of basis-related values, this can result in incorrect computations.

This issue is therefore to add in those `getName()` implementations.  For the new "derived" basis implementations this can be done in a fairly general way, as follows:
1. Each component basis (line H^1, line H(vol)) should implement a descriptive name.
2. TensorBasis should implement a name that depends on the names of its tensorial components, something along the lines of `basis1Name x basis2Name`.
3. DirectSumBasis should behave similarly, with something along the lines of `basis1Name + basis2Name`.
4. HCurl, HDiv bases should do something similar, with something like `HCurl(HGRADBasisName,HVOLBasisName)`.

CamelliaDPG:
@trilinos/intrepid2 

## Motivation
Some Intrepid2 components rely on basis names to track, for example, basis values in a map keyed on the basis name.  These can break if distinct bases return the same name.  This PR adds unique basis names for some recently added basis classes.

Along related lines, prior to this PR certain Intrepid2 components attempted to parse the basis name to determine e.g. function space and topology.  There are already direct methods provided by the basis that can be queried for these things, and the basis names added by this PR do not follow the naming convention assumed by these parsing strategies (tensor bases in this PR have names derived from their components).  In this PR, the direct methods are invoked, and the parsing strategy dropped.

## Related
Fixes #7071.

### DIFF
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_DerivedBasis_HCURL_HEX.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_DerivedBasis_HCURL_HEX.hpp
@@ -411,16 +411,6 @@ namespace Intrepid2
      */
     Basis_Derived_HCURL_HEX(int polyOrder) : Basis_Derived_HCURL_HEX(polyOrder, polyOrder, polyOrder) {}
 
-    /** \brief  Returns basis name
-
-        \return the name of the basis
-    */
-    virtual
-    const char*
-    getName() const {
-      return "Intrepid2_DerivedBasis_HCURL_HEX";
-    }
-
     /** \brief True if orientation is required
     */
     virtual bool requireOrientation() const {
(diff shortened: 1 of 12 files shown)


CamelliaDPG:
Oops!  Had `master` as the target.  Changed to `develop` just now.

mperego:
> Oops! Had `master` as the target. Changed to `develop` just now.

Ah, OK, that's why there were changes in the cmake folder

CamelliaDPG:
The test failures were build failures, due to an extra copy of the `getName()` implementation that probably resulted from merging in from develop.

mperego:
@CamelliaDPG these commits should fix the tests failing.

CamelliaDPG:
> @CamelliaDPG these commits should fix the tests failing.

Thanks, @mperego!  I just pushed something that should fix the "unused variables" warnings for some compilers.

CamelliaDPG:
@mperego, looks like the tests passed. Would you please approve again?

PR #7559, opened 2020-06-19
Tempus: Adjustments to the AppAction Interface

### LINKED ISSUE #7549: Tempus:  Adjustments to the AppAction Interface
## Enhancement
@trilinos/tempus 

The StepperRKModifierX interface additionally needs the stage number, and for BEFORE_EXPLICIT_EVAL, the solution passed should the stage value, instead of
the working state.

What specifically must we demonstrate to call this work complete?

Show that we can use the stage number and we are getting the stage value.

ccober6:
@trilinos/tempus 

## Motivation
The RK ModifierX interface needed the stage number, and in the BEFORE_EXPLICIT_EVAL case the stageX solution needed to be passed in for EMPIRE.

 * Moved the RK Tableau in to RKBase, along with related
   accessors, to allow access from the base class.
    - IMEX steppers return the explicit tableau from getTableau().
 * Moved getStageNumber() to RKBase.
 * Add stageNumber to the StepperRKModifierXBase interface.
 * Added RK stepper unit tests for stageNumber and stageX.
   The stageX tests were a little difficult to develop, because
   in explicit RK steppers stageX and workingState->getX are the
   same.  So there is no stageX to test outside the stepper.
   Therefore direct comparisons against numerical values was used.

Miscellaneous
 * Fixed some doxygen comments.

## Related Issues

* Closes #7549 

## Stakeholder Feedback
N/A

## Testing
Added unit tests for the stage number access and for the stageX values.


ccober6:
@asteyer @sconde @rppawlo Sorry to pester about this, but can someone review this for me?  Thanks!

packages/tempus/src/Tempus_StepperRKBase.hpp
```diff
@@ -30,8 +31,15 @@ class StepperRKBase : virtual public Tempus::Stepper<Scalar>
 
 public:
 
-  virtual int getStageNumber() const { return stageNumber_; }
+  virtual Teuchos::RCP<const RKButcherTableau<Scalar> > getTableau() const
+  { return tableau_; }
+
+  virtual Scalar getOrder()    const{return getTableau()->order();}
+  virtual Scalar getOrderMin() const{return getTableau()->orderMin();}
+  virtual Scalar getOrderMax() const{return getTableau()->orderMax();}
```
sconde, line 39: type should be `int`

packages/tempus/src/Tempus_StepperIMEX_RK_Partition_decl.hpp
```diff
@@ -417,7 +421,8 @@ class StepperIMEX_RK_Partition : virtual public Tempus::StepperImplicit<Scalar>,
     virtual OrderODE getOrderODE()   const {return FIRST_ORDER_ODE;}
   //@}
 
-  Teuchos::RCP<Thyra::VectorBase<Scalar> >& getStageZ() {return stageZ_;};
+  Teuchos::RCP<Thyra::VectorBase<Scalar> > getStageX() {return stageZ_;}
+  Teuchos::RCP<Thyra::VectorBase<Scalar> > getStageZ() {return stageZ_;};
```
sconde, line 425: Can we doxygen on this?

packages/tempus/src/Tempus_StepperIMEX_RK_decl.hpp
```diff
@@ -326,6 +326,10 @@ class StepperIMEX_RK : virtual public Tempus::StepperImplicit<Scalar>,
 
   /// \name Basic stepper methods
   //@{
+    /// Returns the explicit tableau!
+    virtual Teuchos::RCP<const RKButcherTableau<Scalar> > getTableau() const
+    { return explicitTableau_; }
+
```
sconde, line 332: This is only getting the ERK tableau. And what about the DIRK tableau?

packages/tempus/src/Tempus_StepperRKBase.hpp
```diff
@@ -30,8 +31,15 @@ class StepperRKBase : virtual public Tempus::Stepper<Scalar>
...
+  virtual int getNumberOfStages() const {return getTableau()->numStages();}
 
+  virtual int getStageNumber() const { return stageNumber_; }
   virtual void setStageNumber(int s) { stageNumber_ = s; }
 
   virtual Teuchos::RCP<Thyra::VectorBase<Scalar> > getStageX() {return stageX_;}
```
sconde, line 45: Should `getStageZ` be moved to the base class as well? and application could simply test if its null to determine if stepper is partitioned or not?

packages/tempus/src/Tempus_StepperRKBase.hpp
ccober6: @sconde  I have debated this.  Are there time integrators with fractional order?

packages/tempus/src/Tempus_StepperIMEX_RK_Partition_decl.hpp
ccober6: Yeah, sounds like I good idea.  I mentioned it in the commit, but something more permanent would be a good idea.

packages/tempus/src/Tempus_StepperIMEX_RK_decl.hpp
ccober6: Yeah, I had to decide what would getTableau() do, since it is in the base class and only returns a single tableau.  Would you like to have a getExplicitTableau() and a getImplicitTableau()?

packages/tempus/src/Tempus_StepperRKBase.hpp
ccober6: I have leaned towards having the get and set functions be in the same class as the member data definition, and would not unnecessarily move the member data definition to a base class.  Does that seem reasonable?

packages/tempus/src/Tempus_StepperIMEX_RK_decl.hpp
sconde: If we can have the base class take a `bool explicit = true`, I think we can keep a single method.

packages/tempus/src/Tempus_StepperRKBase.hpp
sconde: I'm ok with that. Thanks

packages/tempus/src/Tempus_StepperRKBase.hpp
sconde: Not by designed. But somes schemes are known to have slighter higher convergence rate in the asymptotic regime for certain classes of problems.

If the idea is to potentially test for these in the future, then `Scalar` is fine by me :)

sconde:
@ccober6 please let me know when I should review this again.

packages/tempus/src/Tempus_StepperRKBase.hpp
ccober6: I will get this in the next PR.

packages/tempus/src/Tempus_StepperIMEX_RK_Partition_decl.hpp
ccober6: Done!

packages/tempus/src/Tempus_StepperIMEX_RK_decl.hpp
ccober6: Discussed with @sconde and going with getExplicitTableau() and getImplicitTableau() for IMEX and IMEX Partition.

ccober6:
@sconde Did a force push, so it should be ready for review.  Thanks for the help!

ccober6:
@sconde I made some additional fixes and restarted the testing.  Can you review this for me? Sorry for the multiple reviews, but hopefully we can get this one through and I will start another issue.  Thanks!

ccober6:
@sconde I have failed to push my recent changes.  Give me a few minutes to get it pushed.

ccober6:
@sconde OK already for you to review.  Thanks!

PR #7563, opened 2020-06-20
Tpetra: Make host version of map hash

MicheldeMessieres:
Give map a host hash and remove fences as discussed in #7446 

@kddevin Code in clone method was also updated but it is deprecated and not tested.

@trilinos/tpetra 

## Motivation

Fences are being evaluated and removed from Tpetra when possible.

## Testing

Tpetra unit tests. Checked white Cuda parallel build and Mac parallel/serial builds.
Also tested these changes with #7446  to verify fence removal works for CUDA_LAUNCH_BLOCKING=0 after these changes.

### DIFF
--- a/packages/ifpack2/example/RelaxationWithEquilibration.cpp
+++ b/packages/ifpack2/example/RelaxationWithEquilibration.cpp
@@ -102,12 +102,8 @@ tpetraToEpetraMap (const Tpetra::Map<LO, GO, NT>& map_t,
     }
   }
   else {
-    auto myGblInds_d = map_t.getMyGlobalIndices ();
-    typename decltype(myGblInds_d)::HostMirror::non_const_type
-      myGblInds_h ("myGblInds_h", myGblInds_d.extent (0));
-    Kokkos::deep_copy (myGblInds_h, myGblInds_d);
-    Kokkos::fence ();
-    return Epetra_Map (gblNumInds, lclNumInds, myGblInds_h.data (), indexBase, comm_e);
+    auto myGblInds = map_t.getMyGlobalIndices (); // returns host data
+    return Epetra_Map (gblNumInds, lclNumInds, myGblInds.data (), indexBase, comm_e);
   }
 }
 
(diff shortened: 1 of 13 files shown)


csiefer2:
The existing lgMap_ view CudaUVMSpace yes?  Can we get it only in CudaSpace to avoid increasing the host-side memory use over the existing implementation?

MicheldeMessieres:
@csiefer2 Thanks. I will make that change. Also that reminds me for non-Cuda builds, I should change the host version to just point to the device version, so no extra copy.

MicheldeMessieres:
@csiefer2 Can you please clarify if you meant lgMap_, the local global mapping (not modified in this PR)? I think you meant glMap_, the FixedHashTable which I added a host copy of. The current implementation of FixedHashTable assumes UVM so if we did want to template it on CudaSpace we would need to refactor that. I expect we would want to do that as a separate PR, perhaps a precursor to this one. It looks like FixedHashTable still needs a full kokkos implementation. Before proceeding I wanted to confirm what you had in mind.

MicheldeMessieres:
@csiefer2 I added a minor change so the non-coda build won't generate duplicate copies of the views.

csiefer2:
Games can be played to get CudaSpace allocations, for when we really want thank instead of CudaUVMSpace w/o having to muck with templates, e.g.:

```
 typedef typename std::conditional<std::is_same<node_memory_space,Kokkos::CudaUVMSpace>::value,
                                        Kokkos::CudaSpace,
                                        node_memory_space>::type memory_space;

**Comment by MicheldeMessieres** [comment]
@kddevin @csiefer2 I was going to add the changes to make lgMap_ CudaSpace as well to this PR along with the other things we discussed - I think we decided to put them all together. Also this PR needs a fix so it will build on Serial. I added the WIP tag until I fix that.

**Comment by MicheldeMessieres** [comment]
@kddevin @csiefer2 Summarizing some of the previous e-mail threads: I believe there are two questions to resolve before continuing:

(1) Should getMyGlobalIndices() return a pure host form? I think this may be the correct approach. Previously it was UVM.

(2) Should we change this map constructor to take a pure CudaSpace for indexList? I think we may decide to just leave this as it was, taking UVM:
```
Map (const global_size_t numGlobalElements,
      const Kokkos::View<const global_ordinal_type*, device_type>& indexList,
      const global_ordinal_type indexBase,
      const Teuchos::RCP<const Teuchos::Comm<int> >& comm);
```

MicheldeMessieres:
@kddevin Just pinging my update to fix Xpetra needs a new review to test.

PR #7569, opened 2020-06-22
Tpetra: integrate cuSPARSE merge path spmv

brian-kelley:
- Copy over necessary changes from KK PR https://github.com/kokkos/kokkos-kernels/pull/725 to Trilinos
- Change ``CrsMatrix::apply()`` to use merge path if matrix is sufficiently
imbalanced. Behavior is selected independently on each node:
```
if maxNodeNumRowEntries - avgNodeNumRowEntries >= $TPETRA_LONG_ROW_MIN_NUM_ENTRIES
  then use merge path
else
  use default spmv
```
This heuristic is designed to only enable merge path if there is actually imbalance. So if the average nonzeros per row is high but the rows are balanced, the default SPMV can still be used.

Additionally, by default this will only be used if doing SpMV on a single vector,
not multivector.  If $TPETRA_MULTIVECTOR_USE_MERGE_PATH == ON, then
merge path SPMV will be applied to one column at a time. This is because
cuSPARSE doesn't yet support merge path on multivectors.

@trilinos/tpetra 
@trilinos/kokkos-kernels 

## Motivation

Merge path SPMV in cuSPARSE has better performance than default SPMV on matrices with extreme entries per row imbalance, for example an FE matrix with a few "supernode" rows dense and all others following a usual sparse pattern.

## Stakeholder Feedback

Feature requested by @vbrunini .
## Testing

Added a test "ImbalancedRowMatrix" to CrsMatrix_UnitTests. This exercises both the rank-1 and rank-2 behavior of apply, on a matrix that triggers the imbalanace condition for using merge path. I verified using printouts that doing ``export TPETRA_MULTIVECTOR_USE_MERGE_PATH=ON`` does actually change the behavior for the multivector case.


vbrunini:
There is also Tpetra_Details_residual.hpp that has a fused residual kernel instead of an spmv + multivector update call. The merge patch spmv + multivector update is much faster for these imbalanced matrices than the fused kernel.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -5442,7 +5449,14 @@ namespace Tpetra {
          std::runtime_error, "X and Y may not alias one another.");
     }
 
-    lclMatrix_->apply (X_lcl, Y_lcl, mode, alpha, beta);
+    LocalOrdinal nrows = getNodeNumRows();
+    LocalOrdinal maxRowImbalance = 0;
+    if(nrows != 0)
+      maxRowImbalance = getNodeMaxNumRowEntries() - (getNodeNumEntries() / getNodeNumRows());
```
kddevin, line 5455: Reuse nrows instead of getNodeNumRows()?

packages/tpetra/core/src/Tpetra_LocalCrsMatrixOperator_decl.hpp
```diff
@@ -92,6 +92,16 @@ namespace Tpetra {
            const mv_scalar_type alpha,
            const mv_scalar_type beta) const override;
 
+    void
```
kddevin, line 95: A comment describing what this method is and why it is here would be good.

packages/tpetra/core/src/Tpetra_LocalCrsMatrixOperator_def.hpp
```diff
@@ -99,6 +100,65 @@ apply (Kokkos::View<const mv_scalar_type**, array_layout,
...
+  TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC
+    (X.data () == Y.data () && X.data () != nullptr,
+     std::runtime_error, "X and Y may not alias one another.");
+#endif // HAVE_TPETRA_DEBUG
+
+  const auto op = transpose ?
+    (conjugate ? KokkosSparse::ConjugateTranspose :
+     KokkosSparse::Transpose) : KokkosSparse::NoTranspose;
+  //Select the merge path algorithm (used if available, otherwise has no effect)
+  //TODO BMK: If/when KokkosKernels gets its own SPMV implementation for imbalanced rows,
```
kddevin, line 136: Nice comments.  Thanks

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
brian-kelley: Good idea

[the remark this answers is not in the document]
brian-kelley: - I'm not sure if one of these (integer vs. proportion) is necessarily better than the other. One contrived example where absolute integer is better: all rows have 1 entry, except for one row that has 4 entries. It would be faster to use normal SPMV, but the imbalance proportion is 4x so that would probably trigger the imbalanced code path.
- I agree that the "excess of average" is not what the Behavior function name suggests. The reason I did that is to reuse what's already there from Tim's work, and that in the target use case (large, sparse matrix with a handful of dense rows) the average is much smaller than the max row length, so the difference between "excess of average" and just "long row length" is small. I realize now that Tim does TeamPolicy unpacking all the time, so I'm the only one using ``Behavior::longRowMinNumEntries()``. I think I'll change this like you suggest.
- Integer division should be OK here: IDIV has a max latency of 95 cycles, which is 3e-8s at 3 GHz. So about 0.1% of a 30us kernel launch (it's not nothing, but I would take this over precomputing it in computeLocalConstants() and then reading it each time).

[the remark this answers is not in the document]
brian-kelley: @kddevin OK, I renamed that to ``Behavior::rowImbalanceThreshold()``/``TPETRA_ROW_IMBALANCE_THRESHOLD``. The exact meaning is documented here:
```
  /// \brief Threshold for deciding if a local matrix is "imbalanced" in
  /// the number of entries per row. The threshold is compared against
  /// the difference between maximum row length and average row length.
  ///
  /// The threshold is measured in max number of entries in excess of the
  /// average (it is not a proportion between max and average).
  ///
  /// If the "imbalance" of a local matrix is greater than this threshold,
  /// a different algorithm may be used for some operations like 
  /// sparse matrix-vector multiply, packAndPrepare, and
  /// unpackAndCombine.  You may control this at run time via the
  /// <tt>TPETRA_ROW_IMBALANCE_THRESHOLD</tt> environment variable.
```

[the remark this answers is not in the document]
kddevin: I should have been more specific about the integer division.
I was not thinking about its performance.  Rather, I was thinking about its truncation.  From my other work, I am used to thinking of imbalance as a percentage, and we've been burnt in the past by integer division's truncation.  I can't quickly think of a case here where it would be a problem, but I thought I'd ask the question.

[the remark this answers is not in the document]
brian-kelley: @kddevin Ah, sorry. I don't think truncation will hurt us here, because it can only change the mean entries per row by either 0 or 1. So the real threshold is somewhere between ``rowImbalanceThreshold() - 1`` (maximum truncation) and ``rowImbalanceThreshold()`` (no truncation). The mean could round down to 0, but the "imbalance" will never be negative.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -5442,7 +5449,14 @@ namespace Tpetra {
          std::runtime_error, "X and Y may not alias one another.");
     }
 
-    lclMatrix_->apply (X_lcl, Y_lcl, mode, alpha, beta);
+    LocalOrdinal nrows = getNodeNumRows();
+    LocalOrdinal maxRowImbalance = 0;
+    if(nrows != 0)
+      maxRowImbalance = getNodeMaxNumRowEntries() - (getNodeNumEntries() / nrows);
+    if(size_t(maxRowImbalance) >= Tpetra::Details::Behavior::rowImbalanceThreshold())
```
kyungjoo-kim, line 5456: A question which may not be very important. Here, to apply merge based algorithm we check load imbalance ratio to detect long rows. This means that we still use the so-called standard spmv for a balanced matrix of long rows. What I would like to point is that the merge based spmv does not only resolve the workload imbalance but it also creates more concurrency. For the balanced long  rows, it can still chop the long rows into multiple pieces (I am not 100% sure; but, the merge algorithm should be able to do so). Thus, more concurrent threads can work efficiently. On the other hand, the current kokkos spmv distributes threads over multiple rows (team-thread) and row reduction (thread-vector). This might be suboptimal choice for the balanced long row matrices. Even though we can increase the team size, it might not work well and under-utilize the device for a collection of extremely long rows.

brian-kelley: @kyungjoo-kim That's a good point. Now I'm thinking a simple "long rows" cutoff like before is what we want:
```
if getNodeMaxNumRowEntries() >= threshold:
  do_merge_path()
else
  do_normal_spmv()
```
Then, only if all rows are shorter than the threshold will normal SPMV be used. So the threshold would be chosen to be a little higher than the maximum degree seen in a regular sparse matrix.

brian-kelley: @kyungjoo-kim @kddevin  I did some experiments on V100 to test this case in KokkosKernels. What I found was that although KokkosKernels SPMV becomes ~2x slower for the few rows but many entries case, cuSPARSE's default algorithm actually stays slightly faster than merge path. It seems that merge path really is only beneficial for high imbalance.

Example that generates a fully dense matrix that's 100x100k.
```
[bmkelle@kokkos-dev-2 sparse]$ ./sparse_spmv_merge_wide --compare -numRows 100 -numCols 100000
Test parameters:
  - loop:           100
  - compare:        1
  - numRows:        100
  - numCols:        100000

Matrix statistics:
  - average nnz per row: 100000
cuSPARSE Merge alg    ---  min: 1.14594e-06 max: 2.27052e-06 avg: 1.16123e-06
cuSPARSE Default alg  ---  min: 1.08963e-06 max: 1.15345e-06 avg: 1.10374e-06
Kokkos Native alg     ---  min: 1.99581e-06 max: 2.0303e-06 avg: 2.00509e-06
```
More extreme: 10x1 million. 10 is way less than the number of SM's in the device, so no way KokkosKernels is getting good occupancy even at 1 team per row):
```
[bmkelle@kokkos-dev-2 sparse]$ ./sparse_spmv_merge_wide --compare -numRows 10 -numCols 1000000
Test parameters:
  - loop:           100
  - compare:        1
  - numRows:        10
  - numCols:        1000000

Matrix statistics:
  - average nnz per row: 1000000
cuSPARSE Merge alg    ---  min: 1.04111e-05 max: 2.22215e-05 avg: 1.06343e-05
cuSPARSE Default alg  ---  min: 1.04109e-05 max: 1.04592e-05 avg: 1.04225e-05
Kokkos Native alg     ---  min: 2.03733e-05 max: 2.04348e-05 avg: 2.03899e-05
```
It seems that cuSPARSE's default algorithm can make more than 1 block work on each row, unlike KokkosKernels. Although KokkosKernels isn't that bad even in this extreme case.

I think the conclusion to draw from this is that the current "imbalance" metric is a good choice, since whenever it's not getting triggered, the balance will be good enough for cuSPARSE's default to do well.

kyungjoo-kim: Let me re-iterate abit. For the first example, it is dense gemv with long rows. A = (100x100k). Gemv flops is 2*100*100k/1e9 = 0.02 GFlop. This is done with 1.0e-6 second, which means 20 Tflops. Am I right ? This seems to be way faster than it could be in gemv. In terms of bandwidth, 100*100k*8 bytes = 0.08 GB are loaded. Within 1.0e-6 second, it gives 80 TB/s. Either I calculate something really stupid or the measurement does not seem to be right.

brian-kelley: @kyungjoo-kim Wow, you're totally right. For this experiment I made a copy of KokkosSparse_spmv_merge.cpp, in perf_test/sparse, and it seems to divide the times by 1000 at the end:
Measuring (this looks OK):
```
        for(int iterIdx = 0; iterIdx < loop; ++iterIdx) {
          Kokkos::Timer timer;
          // KokkosSparse::spmv(controls, "N", alpha, test_matrix, x, beta, y);
          KokkosSparse::Impl::spmv_beta<matrix_type, values_type, values_type, 1>(controls, "N", alpha, test_matrix, x, beta, y); 
          Kokkos::fence();
          double time = timer.seconds();
          avg_time += time;
          if(time>max_time) max_time = time;
          if(time<min_time) min_time = time;
        }   
```
Reporting:
```
        std::cout << "Kokkos Native alg     ---  min: " << min_time / 1000
                  << " max: " << max_time / 1000
                  << " avg: " << avg_time / (loop*1000) << std::endl;
```
I don't know why it does this. It's still valid to compare times against each other because they're all scaled by the same factor, but I will put in a PR to fix that.

kyungjoo-kim: Now it makes sense. So, the number is 20 Gflop/s and 80 Gflop/s. This evaluates some aspects of the code. I agree that cusparse spmv select a different code path than a usual cases. Comparing the codes between nvidia spmv and nvidia merge spmv, it shows almost no difference. Actually, we do not know how nvidia spmv works. It may internally choose merge spmv or some specialized algorithm (changing the spmv into gemv) for this type of matrices. I generally assume that you test simulation a case that those long row column indices are contiguous. Am I right ? Are column indices sorted ? Is sorted column indices requirement for cusparse spmv ? I am just thinking if there is way that we can avoid the specialized code path in this evaluation or not.  

Since two algorithms perform almost same, i think you can choose any selection criteria whatever you like.

brian-kelley: @kyungjoo-kim The answers to your questions about the 10 x million and 10 x 100k matrices are yes (contiguous) yes (sorted), no (sorting required by cuSPARSE). If I modify the 10 x million matrix to randomly shuffle column indices within each row, I get this:
```
[bmkelle@kokkos-dev-2 sparse]$ ./sparse_spmv_merge_wide --compare -numRows 10 -numCols 1000000
Test parameters:
  - loop:           100
  - compare:        1
  - numRows:        10
  - numCols:        1000000

Matrix statistics:
  - average nnz per row: 1000000
cuSPARSE Merge alg    ---  min: 0.0129282 max: 0.0260412 avg: 0.013524
cuSPARSE Default alg  ---  min: 0.0119601 max: 0.0144351 avg: 0.0131062
Kokkos Native alg     ---  min: 0.0271301 max: 0.0445138 avg: 0.0300604
```
With the rows sorted and contiguous:
```
[bmkelle@kokkos-dev-2 sparse]$ ./sparse_spmv_merge_wide --compare -numRows 10 -numCols 1000000
Test parameters:
  - loop:           100
  - compare:        1
  - numRows:        10
  - numCols:        1000000

Matrix statistics:
  - average nnz per row: 1000000
cuSPARSE Merge alg    ---  min: 0.0104084 max: 0.0222391 avg: 0.010676
cuSPARSE Default alg  ---  min: 0.010407 max: 0.0114375 avg: 0.0104316
Kokkos Native alg     ---  min: 0.020373 max: 0.0204352 avg: 0.0203861
```
I'm kind of surprised the difference between shuffled and sorted is so small, only like 30-50%. But, the mergepath vs. default comparison doesn't change.

kyungjoo-kim: Thank you for the comparison. Usually cusparse takes longer for the first iteration. The higher max time of the cusparse merge algorithm is probably due to the initial start-up overhead. Also, we see very close performance in the average time. With 80GB/s bandwidth utilization (the actual bandwidth utilization is probably higher than that), the GPU device is not fully utilized. So, judging algorithms with this 10x1M and 100x100k problems may not be right unless this problem size is meaningful for a certain sandia applications (even so, I would advise to use dense gemv for this case). Anyway I  think  that finding a practical use case for this type of problems would be difficult.

brian-kelley: @kyungjoo-kim I agree, I don't know of a use case for this shape of matrix. I was just trying to answer the question of whether merge path was better for the "balanced but long rows" case, where KokkosKernels (with at least 1 team per row) can't get good occupancy.

lucbv: @brian-kelley sorry for the misleading choice of units in my example, I probably thought it was useful at some point but definitely did not document it in anyways... I think when I worked on the SpMV the usual difficulty is to load the X vector efficiently and that leads to poor occupancy at least in the kk-native algorithm.
Regarding longer vs shorter rows, we usually try to handle that with the `vector_size` we use for the inner most reduction. This seems to be done similarly in cusparse at least according to the papers I read about it.

brian-kelley:
OK, I think I fixed the last build error.

prwolfe:
I notice this pushed build times on ride to more than double what it was (about 1h24m to 4h30m) Any idea why that is so long?

PR #7570, opened 2020-06-22
This just casts an unsigned to signed to avoid a compiler assert in a…

prwolfe:
… debug build.

Not sure how this was missed this long - found
when setting up a debug PR build.

@trilinos/pamgen 

## Motivation
This is in preparation for a debug gcc 7.2.0 build

## Testing
Built against a gcc 7.2.0 debug build with -Wall

### DIFF
--- a/packages/pamgen/rtcompiler/Registrar.C
+++ b/packages/pamgen/rtcompiler/Registrar.C
@@ -556,7 +556,7 @@ double Readline::execute(Value** args)
 
   string line;
   if (getline(curr_file, line)) {
-    assert(args[1]->getSize() > line.size()); // buffer needs to be big enough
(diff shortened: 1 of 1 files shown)


prwolfe:
Adding reviewers from recent committers. If there are other candidates feel free to add them.

packages/pamgen/rtcompiler/Registrar.C
```diff
@@ -556,7 +556,7 @@ double Readline::execute(Value** args)
 
   string line;
   if (getline(curr_file, line)) {
-    assert(args[1]->getSize() > line.size()); // buffer needs to be big enough
+     assert(args[1]->getSize() > int(line.size())); // buffer needs to be big enough
```
rppawlo, line 559: It loos like getSize() returns a "long" argument, not int.

PR #7572, opened 2020-06-23
Tpetra: new constructor for FECrsGraph accepting owned+shared domain map

### LINKED ISSUE #7455: Tpetra: FECrsGraph wrong default domain map?
## Question
@trilinos/Tpetra

I am not sure if this is a bug, so I'm posting it under 'questions', but I have the feeling it is either wrong or counterintuitive.

One can create a FECrsGraph passing only ownedRowMap and ownedPlusSharedRowMap (the domain and column maps being defaulted to null). In this case, FECrsGraph sets the domain map to the ownedRowMap, even if the graph is in FE_ACTIVE_OWNED_PLUS_SHARED state.

This cause the (overlapped) graph to potentially (and, in practice, I do observe it) have a domain map which is not locally fitted to the column map. This is usually not the case in a crs graph, and took me a while to figure out it's not the case for the FECrsGraph (when in owned+shared mode).

So the question is probably two faced:

1) In a CrsGraph, is it usually desired (and safe to assume) that colMap_->isLocallyFitted(domainMap_) returns true? My assumption was "yes", cause it helps in import/export operations. At the very least, isn't it desirable (and even safe to assume) in the case where the column map is internally generated by Tpetra (rather than passed in by the user)?
2) Assuming the answer to 1 is "yes" (or 'no', but 'yes' to the 'at least' clause), is there a reason not to keep this behavior for the overlapped FECrsGraph?

Note: assuming 1=yes, 2=no, we could restore this symmetry by simply setting domaniMap_ to ownedPlusSharedRowMap in case the input domain map is null.

bartgol:
@trilinos/tpetra

## Motivation

This PR adds a set of constructors to FECrsGraph, which accept a domain map for the owned+shared graph. Before this PR, the owned+shared graph and the owned graph were both created with the same _domain_ map, which, by default, was taken to be the owned row map. This caused the owned+shared dof map in the owned+shared graph to not be locally fitted to the graph's column map, which can cause issues in FE codes assembly that partition the mesh by elements. In such codes, during FE assembly kernels, one may only have the owned+shared dofs map available (coming from a discretization or connectivity object), and not the col map of the matrix. When converting the GIDs of the dofs in the element to local ordinals (in order to assembly the local matrix) one _needs_ the local ids in the owned+shared map to be _the same_ as they will be in the col map of the matrix. Having the owned+shared graph storing the owned+shared dofs as its domain map ensures that owned+shared map is locally fitted to the col map, so that the local ids obtained when converting the dofs GIDs to local ordinal are the same (when using the owned+shared map or the col map of the matrix).
 

## Related Issues

* Closes #7455 

## Testing

The new constructors cannot produce the same graph as the existing ones (otherwise there'd be no point in this PR). However, the resulting graphs should be "locally (and globally) equivalent", meaning that they should store the same row/range/domain map and the same set of GID in each row as a graph built with the existing constructors. The only difference is that colum (global) indices may appear in different order, as well as have different local ids. So I added two sets of tests:

- A test that checks that FECrsGraph built with the new constructor is "locally" equivalent to a good old CrsGraph that is globalAssemble-d.
- A test for MatrixMatrix multiplication, that checks that A*B is "locally equivalent" regardless of whether A,B are FECrsMatrix (with graph built with the new constructors) or CrsMatrix (globalAssemble-d).

### DIFF
--- a/packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
@@ -119,6 +119,8 @@ namespace Tpetra {
   ///
   ///       Notice that if your mesh is partitioned by node (N) OR you provide a valid colMap
   ///       to the graph's constructor, then using V1 is likely ok for you.
+  ///
+  ///       For more details, see GitHub issue #7455
 
   template <class LocalOrdinal,
             class GlobalOrdinal,
--- a/packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
@@ -73,6 +73,52 @@ namespace Tpetra {
   /// This class has an interface like that of Epetra_CrsGraph, but
   /// also allows insertion of data into nonowned rows, much like
   /// Epetra_FECrsGraph.
+  ///
+  /// Note on some versions of the constructors:
+  ///       consturctors appear in two fashions: one that accepts only one (optional) domainMap
+  ///       (let's call it V1) and one that accepts two domainMaps (let's call it V2).
+  ///       In V1, the input domainMap will be used for both owned and owned+shared graphs.
+  ///       In this case, you *should* provide a map that works for the owned graph (since
+  ///       that's the graph you will use in mat-vec and mat-mat produts, which will be checked).
+  ///       The domain map for the owned+shared graph is *usually* only needed to build the colMap
+  ///       of the owned+shared graph.
+  ///
+  ///       Whether you need V2 or are ok with V1 depends on what your "owned+shared" maps contain.
+  ///
+  ///       If you partition a mesh by node (N), then your owned+shared maps should contain all the
+  ///       nodes that are in the patch of one of the owned nodes, which is precisely what
+  ///       the colMap of the owned+shared graph should contain.
+  ///       On the other hand, if you partition the mesh by cell (C), then your owned+shared maps 
+  ///       should contain all nodes belonging to one of the owned elements. Such map does *NOT*
+  ///       necessarily contain all the GID of all the nodes connected to one of the owned nodes
+  ///       (which is what the colMap should contain). In this case, you are not likely to have
+  ///       a colMap available, and you'll rely on Tpetra to build one for you once the graph
+  ///       is filled.
+  ///
+  ///       If you do not provide a valid owned+shared colMap, then Tpetra creates one for you,
+  ///       which is used for both the owned+shared and owned graphs. When doing this, Tpetra
+  ///       guarantees that the domain map of the owned+shared graph is "locally fitted" to its
+  ///       colMap (see Tpetra::Map documentation for the meaning of 'locally fitted').
+  ///       If you use V1, then the domain map is the owned dofs only, which means that the
+  ///       non-owned GIDs will appear in the colMap in an order that is out of your control.
+  ///
+  ///       Now, if you partition by Cell (C), then you don't have a colMap when you create
+  ///       the graph, but only a owned+shared one. If you use version V1 of the consturctor,
+  ///       the Local ID of the shared dofs in the graph's colMap is likely to NOT be the same
+  ///       as in your owned+shared map, since the colMap creation only guarantees that the input
+  ///       domainMap (which should contian owned dofs only) is going to be locally fitted in the
+  ///       colMap. All other GIDs will appear in an order that you can't control.
+  ///
+  ///       On the other hand, if you use V2, then you are guaranteed that the owend+shared domain
+  ///       map you provide will be locally fitted to the graph colMap.
+  ///
+  ///       This difference is important when you do local assembly with a FECrsMatrix: if the Local IDs
+  ///       of the dofs in your owned+shared maps is THE SAME as in the graph's col map, then you can do
+  ///       assembly using local ids. Otherwise, you need to have access to the GIDs of the dofs,
+  ///       extract the colMap from the matrix, and convert GIDs into LIDs ACCORDINGLY TO THE COL MAP.
+  ///
+  ///       Notice that if your mesh is partitioned by node (N) OR you provide a valid colMap
+  ///       to the graph's constructor, then using V1 is likely ok for you.
 
   template <class LocalOrdinal,
             class GlobalOrdinal,
@@ -128,20 +174,20 @@ namespace Tpetra {
     /// \param ownedPlusSharedToOwnedimporter [in] Optional importer between the ownedMap and ownedPlusSharedMap
     ///   This will be calculated by FECrsGraph if it is not provided
     ///
-    /// \param ownedDomainMap [in] Optional domain map for both owned and owned+shared graphs. If this is not provided,
-    ////  then ownedMap will be used in the call to endFill()
+    /// \param domainMap [in] Optional domain map for both owned and owned+shared graphs. If this is null,
+    ////  then ownedMap will be used as domain map for both graphs.
     ///
     /// \param ownedRangeMap [in] Optional range map for the owned graph.  If this is not provided, then ownedMap
-    ///   will be used for the ownedDomainMap in the call to endFill()
+    ///   will be used as range map for the owned graph.
     ///
-    /// \param params [in/out] Optional list of parameters.  If not
+    /// \param params [in/out] Optional list of parameters.  If not 
     ///   null, any missing parameters will be filled in with their
     ///   default values.
     FECrsGraph(const Teuchos::RCP<const map_type> & ownedRowMap,
                const Teuchos::RCP<const map_type> & ownedPlusSharedRowMap,
                const size_t maxNumEntriesPerRow,
                const Teuchos::RCP<const import_type> & ownedPlusSharedToOwnedimporter = Teuchos::null,
-               const Teuchos::RCP<const map_type> & ownedDomainMap = Teuchos::null,
+               const Teuchos::RCP<const map_type> & domainMap = Teuchos::null,
                const Teuchos::RCP<const map_type> & ownedRangeMap = Teuchos::null,
                const Teuchos::RCP<Teuchos::ParameterList>& params = Teuchos::null);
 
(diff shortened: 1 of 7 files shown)


ikalash:
@csiefer2 @trilinos/tpetra

mperego:
@bartgol looks good. Just a warning treated as error:
```
/scratch/trilinos/jenkins/tr-test-0/workspace/trilinos-folder/Trilinos_pullrequest_gcc_7.2.0/Trilinos/packages/tpetra/core/test/FECrsGraph/FECrsGraph_UnitTests.cpp:175:23: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     for (size_t j=0; j<numEntries; ++j) {
                      ~^~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

bartgol:
Thanks for the explanation @mperego ! I didn't know (remember?) that the AT treated warnings as errors. Should be fixed now.

Quick question: the AT triggers also tests in downstream packages that use Tpetra, right? I just want to make sure that my mods didn't mess up things for customers that already used FECrs stuff (it shouldn't have changed anything regarding the behavior of existing ctors, but you know how it goes...).

mperego:
> Quick question: the AT triggers also tests in downstream packages that use Tpetra, right?

Yes.

packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
```diff
@@ -141,10 +141,45 @@ namespace Tpetra {
                const Teuchos::RCP<const map_type> & ownedPlusSharedRowMap,
                const size_t maxNumEntriesPerRow,
                const Teuchos::RCP<const import_type> & ownedPlusSharedToOwnedimporter = Teuchos::null,
-               const Teuchos::RCP<const map_type> & domainMap = Teuchos::null,
-               const Teuchos::RCP<const map_type> & rangeMap = Teuchos::null,
+               const Teuchos::RCP<const map_type> & ownedDomainMap = Teuchos::null,
+               const Teuchos::RCP<const map_type> & ownedRangeMap = Teuchos::null,
                const Teuchos::RCP<Teuchos::ParameterList>& params = Teuchos::null);
 
+    /// \brief Constructorfor globally-indexed assembly specifying a single upper bound for the
```
kddevin, line 148: Please add documentation saying why one would use this constructor over the existing constructors.

packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
```diff
@@ -172,8 +207,43 @@ namespace Tpetra {
                 const Teuchos::RCP<const map_type> & ownedPlusSharedRowMap,
                 const Kokkos::DualView<const size_t*, execution_space>& numEntPerRow,
                 const Teuchos::RCP<const import_type> & ownedPlusSharedToOwnedimporter = Teuchos::null,
-                const Teuchos::RCP<const map_type> & domainMap = Teuchos::null,
-                const Teuchos::RCP<const map_type> & rangeMap = Teuchos::null,
+                const Teuchos::RCP<const map_type> & ownedDomainMap = Teuchos::null,
+                const Teuchos::RCP<const map_type> & ownedRangeMap = Teuchos::null,
+                const Teuchos::RCP<Teuchos::ParameterList>& params = Teuchos::null);
+
+    /// \brief Constructor for globally-indexed assembly specifying a (possibly different) upper
```
kddevin, line 214: Please add documentation saying why one would use this constructor over the existing constructors.

packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
```diff
@@ -207,10 +277,79 @@ namespace Tpetra {
                const Teuchos::RCP<const map_type> & ownedPlusSharedColMap,
                const size_t maxNumEntriesPerRow,
                const Teuchos::RCP<const import_type> & ownedPlusSharedToOwnedimporter = Teuchos::null,
-               const Teuchos::RCP<const map_type> & domainMap = Teuchos::null,
-               const Teuchos::RCP<const map_type> & rangeMap = Teuchos::null,
+               const Teuchos::RCP<const map_type> & ownedDomainMap = Teuchos::null,
+               const Teuchos::RCP<const map_type> & ownedRangeMap = Teuchos::null,
                const Teuchos::RCP<Teuchos::ParameterList>& params = Teuchos::null);
 
+    /// \brief Constructor for locally-indexed assembly specifying a single upper bound for the
```
kddevin, line 284: Please add documentation saying why one would use this constructor over the existing constructors.

packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
```diff
@@ -42,11 +42,13 @@
 #ifndef TPETRA_FECRSGRAPH_DEF_HPP
 #define TPETRA_FECRSGRAPH_DEF_HPP
 
-#include <type_traits>
-#include "Tpetra_CrsGraph.hpp"
+#include "Tpetra_FECrsGraph.hpp"
 #include "Tpetra_Details_Behavior.hpp"
 #include "Tpetra_Details_getEntryOnHost.hpp"
+#include "Tpetra_Details_makeColMap.hpp"
```
kddevin, line 48: Is this include needed?  It appears the code that calls it is commented out below.

packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
```diff
@@ -308,6 +379,42 @@ void FECrsGraph<LocalOrdinal, GlobalOrdinal, Node>::beginFill() {
 
 }
 
+// template<class LocalOrdinal, class GlobalOrdinal, class Node>
```
kddevin, line 382: This function is added to the file, but completely commented out.  Is it needed?

packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
```diff
@@ -128,11 +128,11 @@ namespace Tpetra {
     /// \param ownedPlusSharedToOwnedimporter [in] Optional importer between the ownedMap and ownedPlusSharedMap
     ///   This will be calculated by FECrsGraph if it is not provided
     ///
-    /// \param domainMap [in] Optional domainMap for the owned graph.  If this is not provided, then ownedMap
-    ///   will be used for the domainMap in the call to endFill()
+    /// \param ownedDomainMap [in] Optional domain map for both owned and owned+shared graphs. If this is not provided,
```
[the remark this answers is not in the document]
bartgol: Ah, good catch. I was updating the comments to stress the fact that that map was for the owned graph only, but I must have messed up with copy/paste from another ctor. I will go over the comments again, fix this, and make sure there are no other mistakes.

packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
```diff
@@ -42,11 +42,13 @@
 #ifndef TPETRA_FECRSGRAPH_DEF_HPP
 #define TPETRA_FECRSGRAPH_DEF_HPP
 
-#include <type_traits>
-#include "Tpetra_CrsGraph.hpp"
+#include "Tpetra_FECrsGraph.hpp"
```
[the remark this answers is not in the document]
bartgol: It is not. I just thought the above one was not the one intended. I'm ok removing the include of FECrsGraph, but then, what's the point of including CrsGraph either? Should I remove that too?

packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
```diff
@@ -166,24 +243,21 @@ void FECrsGraph<LocalOrdinal, GlobalOrdinal, Node>::setup(const Teuchos::RCP<con
...
      
    }
-
-   // For FECrsGraph, we do all the aliasing AFTER import.  All we need here is a constructor
-   if(ownedPlusSharedColMap.is_null())
-     inactiveCrsGraph_ = Teuchos::rcp(new crs_graph_type(ownedRowMap,0,StaticProfile,params));
-   else
-     inactiveCrsGraph_ = Teuchos::rcp(new crs_graph_type(ownedRowMap,ownedPlusSharedColMap,0,StaticProfile,params));
+ } else {
+   // We don't need the importer. Setting it to null helps detecting we are in a non-FE case later
```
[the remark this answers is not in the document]
bartgol: I'll improve the comment. Non-FE is indeed a poor word choice. I was trying to refer to the case where there's no need for an inactive graph, since owned and owned+shared row maps coincide.

packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
bartgol: Will do.

packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
bartgol: Ah, good catch. I went through a few impl cycles, where at some point I tried to build a col map, but then backed off. This was supposed to be removed.

packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
bartgol: Ouch, same as above. I'll fix all this sloppy nonsense.

packages/tpetra/core/src/Tpetra_FECrsGraph_decl.hpp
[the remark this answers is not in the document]
bartgol: Ah, sorry, this is the already existing ctor. Indeed, the old ctor was using this map for both owned and owned+shared. I will change the input var name so that it does not imply owned or owned+shared.

bartgol:
@kddevin I think I addressed your concerns, but please, let me know if something is still off/unclear/misleading.

I ended up adding a relatively long paragraph in the class documentation at the top, explainind the difference (and reason of being) between the consturctor that accept a owned+shared domain map and those that don't. It might have come out a bit long, but I didn't know how to clearly explain the issue in two lines, and also I figured "better a bit verbose than a bit cryptic".

bartgol:
> Can you also add a comment with the github issue number motivating the additions?

Do you mean inside the code?

bartgol:
> > Can you also add a comment with the github issue number motivating the additions?
> 
> Do you mean inside the code?

I think that's what you meant, so I went ahead and put a reference in the comment in the FECrsGraph header.

mperego:
@bartgol I guess the autotester won't run until @kddevin will approve your changes. 
@jwillenbring can you please add @bartgol to the trilinos team? He is  a Sandian. He is  not actively working on trilinos but he is a developer of Albany and he occasionally has patches for bugs or new features (like this one) for trilinos.

jwillenbring:
> @jwillenbring can you please add @bartgol to the trilinos team? 

I sent an invite.

bartgol:
> > @jwillenbring can you please add @bartgol to the trilinos team?
> 
> I sent an invite.

Thanks!

mperego:
@kddevin , I think @bartgol addressed all of your comments. We'll merge the branch by COB today if there are no objections.

PR #7587, opened 2020-06-25
Phalanx: fix shared mem check and support corner case

rppawlo:
## Motivation
A shared memory safety check was failing in an app. Needed to fix the check for a specific use case.

While fixing that, also noticed a corner case that was incorrect. Fixed and added new tests.

## Stakeholder Feedback

Fix was verified in EMPIRE. Able to run problem to completion with shared memory enabled. EF-787

## Testing

All covered under new tests.

### DIFF
--- a/packages/phalanx/src/Phalanx_DAG_Manager.hpp
+++ b/packages/phalanx/src/Phalanx_DAG_Manager.hpp
@@ -41,7 +41,6 @@
 // ************************************************************************
 // @HEADER
 
-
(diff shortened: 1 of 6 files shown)


rppawlo:
@etphipp - can you reapprove. Had to fix a compiler warning.

etphipp:
> @etphipp - can you reapprove. Had to fix a compiler warning.

Done :)

Issue #7357, opened 2020-05-13
Tempus:  add initialize() and setObserver() routines to IntegratorAdjointSensitivity class


ikalash:
This is a feature request from Albany.  In order to add adjoint transient sensitivities to Piro/Albany given the current design (intended to work in the same way for the no sensitivity, forward sensitivity, and adjoint sensitivity case), the Tempus::IntegratorAdjointSensitivity class needs the following routines to be implemented:
- ```void initialize()```
- ```void setObserver(Teuchos::RCP< IntegratorObserver< Scalar > > obs)```

@etphipp , can you please add these to Tempus and let me know when they are there?  Hopefully it is not too difficult to do so.

@trilinos/tempus

etphipp:
I added a PR (#7592) that adds these functions as requested.  No tempus tests use these functions, so I guess we will just wait and see how they are used in piro to see if further changes are needed.  There are also many other IntegratorBasic functions that could be added as well, but I did not do so.

ikalash:
Routines have been added - thanks @etphipp . Closing.

PR #7592, opened 2020-06-26
Tempus:  Add setObserver(), initialize() funcs for issue #7357

etphipp:
This adds the setObserver() and initialize() functions to the adjoint
sensitivity integrator as requested in issue #7357.  These just call the
corresponding functions on the forward and adjoint integrators.  Note it
is is unclear how observes should work with adjoint sensitivities, so I
did not add the observer to the adjoint integrator.  That can easily be
changed if required.


etphipp:
> Looks good, Eric. I verified that if I uncomment the calls to the new routines in Piro_TempusIntegrator_Def.hpp, the code compiles just fine and all Piro tests pass.

Great!  I assume none of the Piro tests actually use the adjoint sensitivity integrator yet though, correct?

ikalash:
Right they don't.  I need these routines in order to add that capability.

PR #7596, opened 2020-06-27
Panzer: Fixing evaluation type conditionals in equation set gather registration.

crockatt:
@trilinos/panzer

## Motivation

#7337 introduced the capability to disable evaluation types at runtime. However, there seemed to be some errors in panzer::PhysicsBlock::buildAndRegisterGatherAndOrientationEvaluators. This should hopefully fix that. Seems to be working to disable tangent evaluation types in drekar.


crockatt:
@rppawlo Here is the pull request I mentioned at the meeting today.

PR #7597, opened 2020-06-28
Stk patch simd 0628

alanw0:
@trilinos/<teamName>


alanw0:
No, not against master...

Issue #7532, opened 2020-06-15
Panzer: Build times for Cuda are very long


prwolfe:
## Enhancement
@trilinos/panzer 

While attempting to sort out the issues on ride I looked to see what was causing the up to 9 hour build times. Turns out two files account for most of this time.

packages/panzer/disc-fe/src/CMakeFiles/panzer-disc-fe.dir/Panzer_PointValues2.cpp.o
packages/panzer/disc-fe/src/CMakeFiles/panzer-disc-fe.dir/Panzer_IntegrationValues2.cpp.o

Each of these takes over 6.5 hours to compile both for the Forrester and Garrison nodes. They also consume a good bit of memory during that compile, but the machine has enough to handle that. The rest of Trilinos finishes around 1.5 hours and these two run until the library they are in finally links, then dependencies finish up.

bartlettroscoe:
CC: @rppawlo 

@prwolfe, we are we not seeing anything like this in the ATDM Trilinos builds on 'ride' with Panzer as shown here yesterday:

* https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&date=2020-06-21&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-white-ride-cuda&field2=buildname&compare2=64&value2=-exp

?

Actually, we can go back a few days to last Thursday's results here:

* https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&date=2020-06-18&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-white-ride-cuda&field2=buildname&compare2=64&value2=-exp

that triggered a rebuild of Panzer files (since there were changes to Tpetra) and it showed:

## ATDM

| Site | Build Name | Update | Update Time | Conf Err | Conf Warn | Conf Time | Build Err | Build Warn | Build Time | Test Not Run | Test Fail | Test Pass | Test Time | Test Proc Time | Start Test Time | Labels |
| -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- |
[...]

PR #7600, opened 2020-06-29
The compile times for two Panzer files went up to over 6 hours.

prwolfe:
This turns off one feature that allows these in about 24 minutes. Please remove
when issue #7532 is resolved.

@trilinos/framework 
@trilinos/panzer 

## Motivation
See #7532 

## Related Issues
* Part of #7532

## Testing
I built this by hand on ride and saw the shortened build times expected.


prwolfe:
> Thank you for tracking this down. Looks good.

No - it was @etphipp that tracked it down and deserves the thanks. And I did run a few tests, but the PR tester will run them all so we will see that soon.

PR #7607, opened 2020-06-30
I commented out an assert to get debug tests to run

### LINKED ISSUE #7582: ROL:  Debug test failure in ROL_example_tempus_example_parabolic_modeleval_forward-adjoint_MPI_1
## Bug Report
@trilinos/rol 

### Description
Debug build using gcc 7.2.0 was used and the test ROL_example_tempus_example_parabolic_modeleval_forward-adjoint_MPI_1 failed with

ROL_example_tempus_example_parabolic_modeleval_forward-adjoint.exe: /scratch/prwolfe/Trilinos/packages/rol/example/tempus/example_parabolic_modeleval_forward-adjoint.hpp:542: void ParabolicModelWopForward<Real>::applyImpl(Thyra::EOpTransp, const Thyra::MultiVectorBase<Scalar>&, const Teuchos::Ptr<Thyra::MultiVectorBase<Scalar> >&, Real, Real) const [with Real = double]: Assertion `opSupportedImpl(M_trans)' failed.

We are attempting to set up this build for PR use so this needs to work - I can turn it off temporarily if it turns out to represent a bigger underlying issue.

### Steps to Reproduce
SHA1:  debug will work
Configure script:  Use cmake/std/PullRequestLinuxGCC7.2.0TestingSettings.cmake and add
* set (CMAKE_BUILD_TYPE DEBUG CACHE STRING "" )
* set (Trilinos_ENABLE_DEBUG ON CACHE STRING "")

prwolfe:
It looks to me like this assert is not always
valid as the test at line 529 will return false
if the trans requested is not Thyra::NOTRANS.
The test does in fact call it with Thyra::TRANS.

## Motivation
This is part of setting up debug builds for the PR testing system

## Related Issues

* Closes #7582 
* Blocks #7581 

## Testing
This was run in a gcc 7.2 build using DEBUG settings from PR 7581

### DIFF
--- a/packages/rol/example/tempus/example_parabolic_modeleval_forward-adjoint.hpp
+++ b/packages/rol/example/tempus/example_parabolic_modeleval_forward-adjoint.hpp
@@ -539,7 +539,7 @@ protected:
                  const Teuchos::Ptr<Thyra::MultiVectorBase<Real> >& Y,
                  const Real a,
                  const Real b) const {
-    assert(opSupportedImpl(M_trans));
+    // assert(opSupportedImpl(M_trans));
     ROL::Ptr<const Thyra::VectorBase<Real>> x = X.col(0);
     ROL::Ptr<Thyra::VectorBase<Real>> y = Y->col(0);
     ROL::Ptr<Thyra::VectorBase<Real>> Mx = y->clone_v();


prwolfe:
@dridzal - is there a problem with this approach?

jwillenbring:
CC: @ccober6 due to this being Tempus related
CC: @mperego due to associated product lead

Do either of you have an issue with this change? @prwolfe explains his reasoning above. It seems the assert is contradicting.

If we don't hear anything on this by August 12, we are going to approve and merge this to get the debug build working.

jwillenbring:
Thanks @dridzal !

mperego:
> CC: @ccober6 due to this being Tempus related
> CC: @mperego due to associated product lead
> 
> Do either of you have an issue with this change? @prwolfe explains his reasoning above. It seems the assert is contradicting.
> 
> If we don't hear anything on this by August 12, we are going to approve and merge this to get the debug build working.@

@jwillenbring  FYI, tempus is in Roger's product. Anyway glad you got this problem sorted out.

prwolfe:
> This change seems fine to me for now. A slight refactor of the capability will be needed in the future.

Thanks!

PR #7614, opened 2020-07-01
IOSS: cgns - better hueristic for detecting periodic cgns structured …

gsjaardema:
@trilinos/seacas 

## Motivation

The basic problem is that IOSS has to determine whether a “ZoneGridConnectivity” (ZGC) exists because of a normal block-to-block connection or whether it exists due to a processor decomposition.  This is “easy” except for one case – a periodic block in which the block connects to itself.
 
We’ve done a couple “fixes” that will handle this for new models, (added a “Decomp” descriptor to the ZGC) and added some heuristics that work in case this descriptor doesn’t exist (some models have ZGC that are created from processor decomposition with specific naming conventions).  However, there are some models which have existed for a while which don’t have either of these cues…
 
This model is one of those – it has a ZGC named “1to1ConnectionA1” and it does show that the connection is across processors (it is on proc 0, but the donor zone is “blk-1_proc-1”.  But, there is no way to know for sure whether this is blk-1 being split across ranks or whether it is split in another location and this is a ZGC that existed in the original serial model and was updated since now blk-1 is split and the portion of blk-1 it is contacting is now on proc-1…
 
One option is to check whether any of the ZGC in the entire model have the “specific naming convention” or the “Decomp” descriptor and if they don’t then fall back to the previous method and assume that the model isn’t periodic.  This “works” in the majority of cases since I think that most periodic models are newer.
 
This categorization of periodic vs decomp ZGC is needed to correctly determine the “global” IJK range of a block and the local IJK offset of the decompose


packages/seacas/libraries/ioss/src/cgns/Iocgns_DatabaseIO.C
```diff
@@ -80,6 +80,28 @@ namespace {
...
   }
 
+  bool has_decomp_name_kluge(int cgns_file_ptr, int base, int zone, int zgc_idx)
+  {
+    // Check the `zgc_idx`-th ZGC node to see if the name matches the
+    // form described in the `name_is_decomp` function below.  We want to
+    // see if there are *any* names that match this form and if so, we can
+    // use the kluge; otherwise we can't and need to rely on other hueristics.
+    char                    connectname[CGNS_MAX_NAME_LENGTH + 1];
+    char                    donorname[CGNS_MAX_NAME_LENGTH + 1];
```
bartlettroscoe, line 90: Is there no way to use std::string so that we can avoid possible buffer over-runs?  is there strong checking for these max sizes to make sure that never happens?

gsjaardema: No, these get passed in to a C-API TPL which we don't have control over.  There is good checking in the TPL to avoid buffer over-runs.

PR #7619, opened 2020-07-02
Adelus: Remove redundant copy in solve + Host pinned mem for MPI_Irecv/MPI_Send + Add UVM specialization for cuBLAS copy

vqd8a:
@trilinos/adelus

## Motivation

Some minor updates addressing GEMMA's requests:
- Remove a redundant copy in the solve() when host pinned memory is used. 
- Use host pinned memory for MPI_Irecv/MPI_Send (Spectrum MPI) in the solve() to avoid Kokkos view's reference count not reaching zero when the solve function ends.
- Add UVM specialization for TPL cuBLAS copy.


packages/adelus/src/Adelus_defines.h
```diff
@@ -74,7 +74,7 @@
 #elif defined(DREAL)
 
   #define ADELUS_DATA_TYPE double
-  #undef COMPLEX
+  #undef ADELUS_COMPLEX
```
srajama1, line 77: Should use Kokkos Complex, but that is fine for this PR.

vqd8a: Thanks, @srajama1 !

vqd8a:
> 
> 
> Trying to approve again.

Thanks, Siva @srajama1

PR #7621, opened 2020-07-02
TrilinosCouplings: Fixing stk test

csiefer2:
TESTS: RHEL7

### DIFF
--- a/packages/trilinoscouplings/examples/scaling/example_Poisson_stk.cpp
+++ b/packages/trilinoscouplings/examples/scaling/example_Poisson_stk.cpp
@@ -650,43 +650,6 @@ int main(int argc, char *argv[]) {
   }
 
 
-
-  /**********************************************************************************/
-  /************************** DIRICHLET BC SETUP ************************************/
-  /**********************************************************************************/
-
-  // Vector for use in applying BCs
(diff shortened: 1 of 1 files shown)


packages/trilinoscouplings/examples/scaling/example_Poisson_stk.cpp
```diff
@@ -922,27 +890,35 @@ int main(int argc, char *argv[]) {
...
+  std::vector<int> ownedBoundaryNodes; ownedBoundaryNodes.reserve(bcNodes.size());
+  // Loop over boundary nodes
+  for (unsigned i = 0; i < bcNodes.size(); i++) {
+    int bcNodeId = bulkData.identifier(bcNodes[i]);
+    int lid = globalMapG.LID((int) bcNodeId -1);
+    if(lid != -1) {
+      ownedBoundaryNodes.push_back(lid);
+      
+      // get coordinates for this node
+      entity_type bcnode = bulkData.get_entity(NODE_RANK,bcNodeId);
```
alanw0, line 906: It looks like 'bcnode' is the same as 'bcNodes[i]' above, so you don't need this bulkdata.get_entity by Global-ID, which is a map find.

PR #7634, opened 2020-07-04
Refactor idot

brian-kelley:
@trilinos/tpetra 

Refactor idot. Main things are factoring out duplicated code, avoiding redundant View allocation/initialization/copies, and ensuring correctness for non-CUDA aware MPI. 140 net lines of code removed.

## Motivation

Originally started with just reviewing Kokkos fences with #7446 (minimal fences but correctness without CUDA_LAUNCH_BLOCKING). Then I thought I could refactor to make it way shorter and simpler, but that mostly didn't pan out because of the UVM + MPI potential issue. But there's still a net removal of code, slight improvements to performance and a guarantee of correctness with CudaSpace or CudaUVMSpace and non-CUDA aware MPI. I don't think an error here has been observed, but @jjellio has told me that UVM and non-CUDA aware MPI can cause unpredictable failures on some machines.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of #7446 
* Composed of 

## Testing

Added another idot test (same executable) if CUDA is enabled, which sets ``ENVIRONMENT TPETRA_ASSUME_CUDA_AWARE_MPI=OFF`` in order to disable the assumption of CUDA-aware MPI. This exercises the other (new) code path in idot. Ran Belos and Tpetra tests on kokkos-dev2 (CUDA 10 + CUDA-aware MPI) with CUDA_LAUNCH_BLOCKING off. Every test that calls idot passed (Belos and Tpetra are the only packages which call it).

### DIFF
--- a/packages/tpetra/core/src/Tpetra_idot.hpp
+++ b/packages/tpetra/core/src/Tpetra_idot.hpp
@@ -189,7 +189,6 @@ void blockingDotImpl(
   using dev_mem_space = typename result_dev_view_type::memory_space;
   using mirror_mem_space = typename result_mirror_view_type::memory_space;
   using unmanaged_result_dev_view_type = Kokkos::View<dot_type*, dev_mem_space, Kokkos::MemoryTraits<Kokkos::Unmanaged>>;
-  using unmanaged_result_mirror_view_type = Kokkos::View<dot_type*, mirror_mem_space, Kokkos::MemoryTraits<Kokkos::Unmanaged>>;
   using unmanaged_result_host_view_type = Kokkos::View<dot_type*, Kokkos::HostSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged>>;
   const size_t numVecs = globalResult.extent(0);
   //Logic to compute the local dot is no different than when CUDA-aware MPI.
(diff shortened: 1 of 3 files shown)


MicheldeMessieres:
@brian-kelley I confirmed the second fence was working as expected (line 300) for CUDA_LAUNCH_BLOCKING=0. The 1st fence (line 262) doesn't currently have a test.

brian-kelley:
@MicheldeMessieres Oh, I see. Since CudaUVMSpace is both the device and host memory space for DualView, it's impossible to take the "run on host" branch with it. That means I can safely remove some stuff.

brian-kelley:
@MicheldeMessieres To be more specific, the "run on host" branch can only be taken if X lives in exactly Kokkos::HostSpace. That means no need to check for CUDA-aware MPI, and no need for that fence.

brian-kelley:
@MicheldeMessieres OK, this is ready for review now. There is one fence in the file now, and it does get exercised by testing (but only with CUDA-aware MPI). I ran the 4-rank idot test on a system with CUDA-aware and no launch blocking, and it took the correct code path and passed.

PR #7659, opened 2020-07-10
Tpetra: Update MultiVector fences

MicheldeMessieres:
@trilinos/tpetra 

## Motivation

@jhux2 As discussed in #7446 I have moved the fence to allReduceView. I think we may only need the fence for the 1st allReduceView where need_sync_host is true. If the data was modified on host it should already be fenced regarding UVM access. I verified via testing.

## Testing
Cuda build - confirmed using PR #7446 setup.
<!---

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -3951,7 +3951,6 @@ namespace Tpetra {
 
     // If Case 2 then sum up *this and distribute it to all processes.
     if (Case2) {
-      Kokkos::fence();
       this->reduce ();
     }
   }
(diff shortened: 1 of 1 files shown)


cgcgcg:
@MicheldeMessieres In your testing, have you seen performance benefits from this?

MicheldeMessieres:
@cgcgcg I haven't done any performance testing for the CUDA_LAUNCH_BLOCKING=0 and fence change work. Were you thinking of performance in general or specific to this PR? This PR probably won't make a different in performance since we're just moving the fence. The general approach has been to fix code so we can remove fences or move fences to the last point they are needed.

cgcgcg:
My understanding is that the fence for allreduce is delayed, so this could lead to slightly better inner products, no?

MicheldeMessieres:
@jhux Just pinging on this. I wanted to get your confirmation before merging.

PR #7666, opened 2020-07-13
Tpetra: add safety check to matrix market reader

mayrmt:
@trilinos/tpetra 

## Motivation

This PR equips the MatrixMarket reader with more descriptive error messages in case of non-existing/non-open files to be read.

Check, if the file to be read could actually be opened. Throw exception, if not. 

This throws an exception right away, instead of causing downstream errors when trying to read a non-openend/non-existing file.

## Stakeholder Feedback

While I added these tests for debugging of some block methods in MueLu, this is rather general capability and is not associated with any particular project.

## Testing

`ctest` with Tpetra tests enabled succeeds.


mayrmt:
@trilinos/tpetra Auto-requesting code review from the tpetra team did not work. Hence, I manually requested some reviewers.

PR #7668, opened 2020-07-14
Tpetra: Add state tracking for graph to avoid extra fencing.

MicheldeMessieres:
@trilinos/tpetra 
@kddevin @kyungjoo-kim 
For fences in Tpetra_CrsGraph.hpp we discussed the possibility of tracking the state of the graph to validate whether it needed fencing. This PR is a draft of how that might look so opening this so we can discuss if this is what we had in mind.

Having set this up, we can probably reduce the number of fence calls to about 1/2, as measured by running all unit tests. However I'm wondering if this extra complexity is worth it. It seems it would be better to just do the simple thing and always call  Kokkos::fence() before the various UVM host access places.

Note I removed the fence before maxDifference which is on device so I think that fence is not actually needed regardless of our plan for CrsGraph.


MicheldeMessieres:
@kddevin This may be fine to merge as is but I will wait for confirmation from you since I know there were ongoing discussions about the graph fences.

PR #7672, opened 2020-07-15
MueLu: region restriction operator factory

lucbv:
This factory makes assumptions about the problem to allow for
faster algorithms to compute the restriction operator.
Implemented all the geometric cases independently to
fill the restriction matrix. This should allow for
good performance, hopefully faster than what smoothed
aggregation does on a Laplace 3D problem

@trilinos/muelu 

## Motivation
The goal is to improve the setup cost in the context of region multigrid.
This factory combines coalesce drop/amalgamation, aggregation and
transfer operator factory into one. This should decrease the cost compared
to the current approach.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing
A unit-test is added to check the correctness of the new transfer operator.


packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_decl.hpp
```diff
@@ -0,0 +1,104 @@
...
+#define MUELU_REGIONRFACTORY_DECL_HPP
+
+#include "MueLu_TwoLevelFactoryBase.hpp"
+#include "MueLu_RegionRFactory_fwd.hpp"
+
+namespace MueLu {
+
+  /*!
+    @class RegionRFactory class
+    @brief Factory that builds a restriction operator for region multigrid
```
mayrmt, line 57:
```suggestion
    @brief Factory that builds a restriction operator for region multigrid for a 3D Laplace problem
```

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_def.hpp
```diff
@@ -0,0 +1,1015 @@
...
+    RCP<ParameterList> validParamList = rcp(new ParameterList());
+
+// #define SET_VALID_ENTRY(name) validParamList->setEntry(name, MasterList::getEntry(name))
+//     SET_VALID_ENTRY("region: build coarse coordinates");
+// #undef  SET_VALID_ENTRY
+
+    validParamList->set<RCP<const FactoryBase> >("A",                            Teuchos::null,
+                                                 "Generating factory of the matrix A");
+    validParamList->set<RCP<const FactoryBase> >("numDimensions",                Teuchos::null,
+                                                 "Number of spacial dimensions in the problem.");
```
mayrmt, line 73:
```suggestion
                                                 "Number of spatial dimensions in the problem.");
```

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_def.hpp
```diff
@@ -0,0 +1,1015 @@
...
+    validParamList->set<RCP<const FactoryBase> >("lNodesPerDim",                 Teuchos::null,
+                                                 "Number of nodes per spatial dimension on the fine grid.");
```
mayrmt, line 75:
```suggestion
                                                 "Number of local nodes per spatial dimension on the fine grid.");
```

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_def.hpp
```diff
@@ -0,0 +1,1015 @@
...
+  void RegionRFactory<Scalar, LocalOrdinal, GlobalOrdinal, Node>::
+  DeclareInput(Level& fineLevel, Level& /* coarseLevel */) const {
+    const ParameterList& pL = GetParameterList();
+
+    Input(fineLevel, "A");
+    Input(fineLevel, "numDimensions");
+    Input(fineLevel, "lNodesPerDim");
+    Input(fineLevel, "Nullspace");
+
+    if(pL.get<bool>("interp: build coarse coordinates")) {
```
mayrmt, line 95: Where is this parameter defined as valid? `GetValidParameterList()` has some commented code along the lines of `SET_VALID_ENTRY("region: build coarse coordinates");`, but it's commented and uses a slightly different name for the parameter list entry.

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_def.hpp
```diff
@@ -0,0 +1,1015 @@
...
+
+    // Let us create R and pass it down to the
+    // appropriate specialization and see what we
+    // get back!
+    RCP<Matrix> R;
+
+    if(numDimensions == 1) {
+      throw std::runtime_error("RegionRFactory no implemented for 1D case yet.");
+    } else if(numDimensions == 2) {
+      throw std::runtime_error("RegionRFactory no implemented for 1D case yet.");
```
mayrmt, line 147:
```suggestion
      throw std::runtime_error("RegionRFactory no implemented for 2D case yet.");
```

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_def.hpp
```diff
@@ -0,0 +1,1015 @@
...
+
+    // Based on lCoarseNodesPerDim and lFineNodesPerDim
+    // we can compute numRows, numCols and NNZ for R
+    const LO numRows = lCoarseNodesPerDim[0]*lCoarseNodesPerDim[1]*lCoarseNodesPerDim[2];
+    const LO numCols = lFineNodesPerDim[0]*lFineNodesPerDim[1]*lFineNodesPerDim[2];
+
+    // Let us set some parameter that will be useful
+    // while constructing R
+
+    // Length of LOerpolation stencils based on geometry
```
mayrmt, line 193:
```suggestion
    // Length of interpolation stencils based on geometry
```

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_def.hpp
lucbv: Yeah at this point my test does not implement the request/release mechanism properly and having that parameter enabled would lead to a runtime error being thrown by MueLu. While this is understandable it makes it complicated to develop and debug the algorithm. I will look into refining the test and turning back on the coarse coordinates calculation.

mayrmt:
@lucbv Looks good to me. Please let me know, if I can be of any assistance to proceed with this PR.

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_def.hpp
lucbv: So after further thinking, since this factory focuses on linear interpolation it always requires coordinates to be present so I removed completely that boolean parameter.

lucbv:
@mayrmt it's going fine at the moment, I am creating an xml file that can run with Laplace 3D discretizations to assess the convergence rate of the propose operator and also compare the residuals we get with those from `GeometricInterpolationP`.
Unless there are a lots of changes needed to work with this restriction operator instead of a prolongation operator, this should be completed this morning and a commit that includes the new xml will be pushed.

lucbv:
@pohm01 @mayrmt @rstumin 
This is the last update needed, it has already been tested so if someone feels like approving this we can probably merge it this afternoon and let @pohm01 try this on Cori.

PR #7675, opened 2020-07-15
shylu/frosch: missing include statement

jthies:
@trilinos/<teamName>

### DIFF
--- a/packages/shylu/shylu_dd/frosch/src/Tools/FROSch_SubdomainSolver_decl.hpp
+++ b/packages/shylu/shylu_dd/frosch/src/Tools/FROSch_SubdomainSolver_decl.hpp
@@ -100,6 +100,7 @@
 #endif
 
 #ifdef HAVE_SHYLU_DDFROSCH_THYRA
+#include "Stratimikos_DefaultLinearSolverBuilder.hpp"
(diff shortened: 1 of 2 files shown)


searhein:
@jthies Thanks for adding this include. I am wondering if this is related to some change I did to the `SubdomainSolver`. Does this also work if you more the header to `FROSch_SubdomainSolver_decl.hpp`?

jthies:
yes, it could be moved to the decl-file, but I think it belongs in the def-file because that's where it's used. You probably didn't run into the issue because your code included the Strtimikos header before including the ShyLU-header

jthies:
I’m getting mails hat I should add the AT: AUTOMERGE label, but I don’t see how I can add a label here

 

Von: Alexander Heinlein [mailto:notifications@github.com] 
Gesendet: Mittwoch, 15. Juli 2020 13:21
An: trilinos/Trilinos
Cc: Thies, Jonas; Mention
Betreff: Re: [trilinos/Trilinos] shylu/frosch: missing include statement (#7675)

 

@searhein approved this pull request.

@jthies <https://github.com/jthies>  Thank you!

chochmuth:
@jthies at the top right below Reviewers and Assignees you can add labels- I just added it automerge label

PR #7683, opened 2020-07-16
TrilinosCouplings: Adding pamgen-to-exodus converter

csiefer2:
TESTS: RHEL7


gsjaardema:
Note that the `io_shell` program in the IOSS library can use an input type of `pamgen` and will output an exodus file.  So, in essence it is a pamgen to exodus translator...

csiefer2:
Cool!

PR #7685, opened 2020-07-17
Ifpack2 - fix for a race condition. 

kyungjoo-kim:
## Motivation
 
SPARC reports a weird NAN problem. It turns out that there is a race condition occurring in a certain runtime condition, which is not covered from unit tests. This fixes the problem.

## Stakeholder Feedback
@tcfisher 

## Testing

@csiefer2 confirms that it does not go into the NAN spiral.


ndellingwood:
@kyungjoo-kim let me know when the kokkos-kernels PR is ready

kyungjoo-kim:
@ndellingwood I will submit a matching PR after I confirm that this fix the SPARC issue.  Thanks.

PR #7686, opened 2020-07-17
ifpack2: silence warning when mpi is not enabled

ndellingwood:
Changes to be committed:
	modified:   Ifpack2_BlockTriDiContainer_impl.hpp

@trilinos/ifpack2 

## Motivation

Silence unused var warnings when MPI is not enabled


ndellingwood:
@kyungjoo-kim this just moves the MPI if-guards to enclose variables only used when MPI is enabled, silencing unused variable warnings.

PR #7707, opened 2020-07-22
Tempus ssp coeff

sconde:
@trilinos/tempus 

## Motivation

Add the TVD/SSP coefficient of methods. The coefficients are computed numerically for unknown methods.

John asked for this for Drekar to use the information.

### DIFF
--- a/packages/tempus/src/Tempus_RKButcherTableau.hpp
+++ b/packages/tempus/src/Tempus_RKButcherTableau.hpp
@@ -146,6 +146,11 @@ class RKButcherTableau :
     virtual bool isDIRK() const { return isDIRK_; }
     /** \brief Return true if the RK method has embedded capabilities */
     virtual bool isEmbedded() const { return isEmbedded_; }
+    /** \brief Return true if the RK method is TVD */
+    virtual bool isTVD() const { return isTVD_; }
+    /** \brief Return TVD coefficient of RK method */
+    virtual Scalar getTVDCoeff() const { return tvdCoeff_; }
+
 
     /* \brief Redefined from Teuchos::Describable */
     //@{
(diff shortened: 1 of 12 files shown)


packages/tempus/src/Tempus_RKButcherTableau.hpp
```diff
@@ -240,7 +247,9 @@ class RKButcherTableau :
     bool isImplicit_;
     bool isDIRK_;
     bool isEmbedded_;
+    bool isTVD_ = false;
     Teuchos::SerialDenseVector<int,Scalar> bstar_;
+    Scalar tvdCoeff_;
```
ccober6, line 252: We should probably just have isTVD_ and tvdCoeff_ stored in a single location.  I would put it here, and remove it in StepperDIRK and StepperExplicitRK.

packages/tempus/src/Tempus_StepperDIRK_decl.hpp
```diff
@@ -234,6 +234,8 @@ class StepperDIRK : virtual public Tempus::StepperImplicit<Scalar>,
     virtual void setUseEmbedded(bool a) { useEmbedded_ = a; }
     virtual bool getUseEmbedded() const { return useEmbedded_; }
     virtual bool getUseEmbeddedDefault() const { return false; }
+    virtual void setTVDCoeff(const Scalar a) { tvdCoeff_ = a; }
+    virtual void setTVD(bool a) { isTVD_ = a; }
```
ccober6, line 238: Now that we have StepperRKBase, these set functions might be better there instead of duplicating them here and ExplicitRK (and IMEX?).

packages/tempus/src/Tempus_RKButcherTableau.hpp
sconde: ok.

packages/tempus/src/Tempus_StepperDIRK_decl.hpp
sconde: ok

sconde:
@ccober6 I think I've updated the PR for the requested changes

PR #7714, opened 2020-07-23
Tpetra: enforce CUDA_LAUNCH_BLOCKING on old arches

brian-kelley:
Verify that if CUDA architecture is < 6.0 (Pascal), $CUDA_LAUNCH_BLOCKING=1.
Check this in Tpetra::initialize().

@trilinos/tpetra 

## Motivation

Before the Pascal architecture, UVM didn't migrate pages between host and device on demand. Instead, all live UVM views  were migrated to the device before each kernel, and to host after. As @MicheldeMessieres figured out, this violates our assumptions about Kokkos::DualView. For example:
```
//Assume device is Cuda/CudaUVMSpace
View a;
View b;
//assume a, b have both been accessed most recently on host
parallel_for(something_with_a);
b(0) = 5;
```
Even though only a was accessed on device, b can no longer be directly accessed on host without a fence. With Pascal and later, the above works fine since only a is actually moved to device.

Rather than require code to fence explicitly, we will just continue to require CUDA_LAUNCH_BLOCKING=1 for older architectures. This PR verifies this setting when Tpetra is initialized.

## Testing

Tested this on a Pascal-based machine, and a Kepler-based machine (with an extra printout of the "arch" value) and it worked as expected. If I manually unset CUDA_LAUNCH_BLOCKING, the new check threw an exception and aborted the program.


jhux2:
I hate to ask ... but is there a test for this?

MicheldeMessieres:
@brian-kelley We'll need to include:
  #include "TpetraCore_config.h"
Otherwise HAVE_TPETRACORE_CUDA is not defined.

At least it won't work on this white config using the Kepler35 node:
```
cmake \
  -DTpetra_ENABLE_TESTS=ON -DTrilinos_ENABLE_Tpetra=ON \
  ../trilinos2
```

brian-kelley:
@MicheldeMessieres Thanks, fixed.

@jhux2 There isn't now. I just thought about it and figured out how this can be tested.

brian-kelley:
@jhux2 There's a test now, test/Core/check_launch_blocking, that tries to do Tpetra::initialize without launch blocking. I just ran it on my workstation (pascal) and RIDE (kepler).

brian-kelley:
@kddevin Sorry, I just made one last change, can you re-re-review? The verbose output on Kepler looks like this now, before there was a ton of MPI errors because MPI_Finalize wasn't getting called. (I only call it now if MPI is enabled, and it was initialized)
```
17: Environment variables: 
17:  CUDA_LAUNCH_BLOCKING=0
17:  CTEST_KOKKOS_DEVICE_TYPE=gpus
17: Test timeout computed to be: 10000000
17: Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
17:                                   without setting CUDA_LAUNCH_BLOCKING=1.
17:                                   The code must call Cuda().fence() after each kernel
17:                                   or will likely crash when accessing data on the host.
17: Tpetra::initialize threw exception because CUDA_LAUNCH_BLOCKING required but not set.
17: Tpetra::initialize threw exception because CUDA_LAUNCH_BLOCKING required but not set.
17: Tpetra::initialize threw exception because CUDA_LAUNCH_BLOCKING required but not set.
17: Tpetra::initialize threw exception because CUDA_LAUNCH_BLOCKING required but not set.
17: End Result: TEST PASSED
17: End Result: TEST PASSED
17: End Result: TEST PASSED
17: End Result: TEST PASSED
17/17 Test #17: TpetraCore_check_launch_blocking_MPI_4 ......................................................   Passed    0.97 sec
```

PR #7728, opened 2020-07-26
Belos CG improvements

cgcgcg:
@trilinos/belos

## Motivation
In Single Reduce CG:
* Remove one kernel call via memory layout.
* Fix iter counts so that they match standard Block CG.
* Save one vector.

Adds option "Fold Convergence Detection Into Allreduce" to standard (Pseudo)Block CG.

### DIFF
--- a/packages/belos/src/BelosCGIter.hpp
+++ b/packages/belos/src/BelosCGIter.hpp
@@ -53,6 +53,7 @@
 #include "BelosLinearProblem.hpp"
 #include "BelosOutputManager.hpp"
 #include "BelosStatusTest.hpp"
+#include "BelosStatusTestGenResNorm.hpp"
 #include "BelosOperatorTraits.hpp"
 #include "BelosMultiVecTraits.hpp"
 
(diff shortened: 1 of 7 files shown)


hkthorn:
@cgcgcg What do you mean by "Remove one kernel call via memory layout" ?

cgcgcg:
@hkthorn Before, we had two `MvAddMv` calls with single vectors:
https://github.com/trilinos/Trilinos/blob/fd4172f906200be292ee5cdf91749bff095069dd/packages/belos/src/BelosCGSingleRedIter.hpp#L626-L634
By slightly changing the memory layout, we only have one call with multivectors with two columns each:
https://github.com/trilinos/Trilinos/blob/66e9815aff8153261dcd2957ca045b3330772187/packages/belos/src/BelosCGSingleRedIter.hpp#L637-L641
The hope is that this is actually cheaper on GPUs..

packages/belos/src/BelosCGIter.hpp
```diff
@@ -176,7 +176,13 @@ class CGIter : virtual public CGIteration<ScalarType,MV,OP> {
 
   //! Get the norms of the residuals native to the solver.
   //! \return A std::vector of length blockSize containing the native residuals.
-  Teuchos::RCP<const MV> getNativeResiduals( std::vector<MagnitudeType> * /* norms */ ) const { return R_; }
+  Teuchos::RCP<const MV> getNativeResiduals( std::vector<MagnitudeType> * norms ) const {
+    if (foldConvergenceDetectionIntoAllreduce_) {
```
hkthorn, line 180: This is conditional on the type of norm being requested, see the same method in CGSingleRedIter.  Is there a reason it should be different here?

cgcgcg: Good catch!

hkthorn:
@cgcgcg Thanks for making that change.  I did not realize that in order to check that the norm is compatible with the native norm, the residual norm class needs to be passed in to the iteration object separate from the status test.  I'm not a fan of that design, but it has already been established with the single reduction CG iteration.

PR #7736, opened 2020-07-28
Sacado:  Convert CppUnit tests to GTest and fix several issues with the pow() operation

etphipp:
This PR replaces the entire CppUnit testsuite in Sacado with GTest.  Unfortunately, the version of GTest in Trilinos doesn't include all of the needed functionality, so I added my own version of GTest for Sacado (this is the full GTest package from Github).  

This PR also fixes several issues with correct differentiation of the pow() operation.


kliegeois:
Thanks for the PR!
I am testing it since this morning, there are still some issues with nested FAD types.
I hope to be able to propose modifications today.

etphipp:
Added the WIP label because even though the AT passes, there are still installation issues to be sorted out.

etphipp:
I've given up on updating the Trilinos GTest package with the full googletest package.  I can't figure out how to make the installed CMake files work correctly.  So I restored the previous version there and moved the googletest version I need back to a test-only library in Sacado.

PR #7743, opened 2020-07-29
MueLu: More cut drop stuff

csiefer2:
TESTS: RHEL7


rstumin:
what the heck is the stuff about changing the coordinates? Something about pairing (0,1) and (1,0) or something like that.

rstumin:
I guess it is just in the unit test, but it is a little confusing.

PR #7752, opened 2020-07-30
Amesos2: Stokhos: Fix Basker TypeMap to not use static for dtype

MicheldeMessieres:
@trilinos/amesos2 @trilinos/stokhos 

## Motivation

Resolve inconsistency in Basker TypeMap. This change is potentially a backwards compatibility issue if users have customized the TypeMap (similar to how Stokhos changes it).

@ndellingwood This is the change I mentioned earlier. I think we will want to do this. This is not going to fix anything, as the change is just to make things conventionally consistent. I believe the original use of static may have been an artifact of copying from the SuperLU solvers.

### DIFF
--- a/packages/amesos2/src/Amesos2_Basker_decl.hpp
+++ b/packages/amesos2/src/Amesos2_Basker_decl.hpp
@@ -99,7 +99,7 @@ public:
   // TODO: Would like to change dtype to be a regular type, not static.
   // Seems nothing was using dtype before anyways but Stokhos would break so
   // will address that as a separate PR.
-  typedef decltype(type_map::dtype)                            basker_dtype;
+  typedef typename type_map::dtype                             basker_dtype;
 
   typedef FunctionMap<Amesos2::Basker,basker_type>             function_map;
 
--- a/packages/amesos2/src/Amesos2_Basker_decl.hpp
+++ b/packages/amesos2/src/Amesos2_Basker_decl.hpp
@@ -96,9 +96,6 @@ public:
 
   typedef typename type_map::type                               basker_type;
 
-  // TODO: Would like to change dtype to be a regular type, not static.
-  // Seems nothing was using dtype before anyways but Stokhos would break so
-  // will address that as a separate PR.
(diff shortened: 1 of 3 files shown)


packages/amesos2/src/Amesos2_Basker_decl.hpp
```diff
@@ -99,7 +99,7 @@ class Basker : public SolverCore<Amesos2::Basker, Matrix, Vector>
   // TODO: Would like to change dtype to be a regular type, not static.
```
ndellingwood, line 99: I think this comment can now be removed

MicheldeMessieres: @ndellingwood Thanks I removed this - if you can please review again.

MicheldeMessieres:
@ndellingwood Ran Basker unit tests on Amesos2 and also Stokhos tests on my Mac parallel build. A build sufficient to demonstrate the compile error if the Stokhos part was not changed. I did not run the white Cuda build yet but can do that later this evening.

packages/stokhos/src/sacado/kokkos/vector/amesos2/Amesos2_Basker_MP_Vector.hpp
```diff
@@ -68,7 +68,7 @@ namespace Amesos2 {
   // Enable MP::Vector as a valid Scalar type for Basker
   template <class ST>
   struct TypeMap< Basker,Sacado::MP::Vector<ST> > {
-    static Sacado::MP::Vector<ST> dtype;
+    typedef Sacado::MP::Vector<ST> dtype;
```
ndellingwood, line 71: @hkthorn I wanted to check for use of Basker (classic) in Xyce, do you maintain a custom TypeMap struct (similar to this location in Stokhos) and will this change to remove the static type from Basker pose any issues?

MicheldeMessieres:
@ndellingwood I ran Amesos2 Basker and Stokhos tests on Cuda White as well. I'll wait to merge until you confirm it's good to proceed.

PR #7753, opened 2020-07-30
Ifpack2: Resolve CUDA_LAUNCH_BLOCKING=0

MicheldeMessieres:
@trilinos/ifpack2

## Motivation

Adds fencing so that Ifpack2 unit tests will pass on white Cuda builds with CUDA_LAUNCH_BLOCKING=0.

Note this is tested stacked on top of ongoing work for Tpetra so this will not work stand alone with CUDA_LAUNCH_BLOCKING=0. I suggest we accept the addition of these fences and then address removal of individual fences with separate PRs that refactor as needed.

## Testing

Tested on Cuda white builds stacked on ongoing Tpetra work to fix fences. Also tested on Mac parallel. This PR just adds fences and sync_host calls.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_MultiVectorLocalGatherScatter.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_MultiVectorLocalGatherScatter.hpp
@@ -110,6 +110,7 @@ public:
         X_out_j[i] = X_in_j[i_perm];
       }
     }
+    X_out.modify_host();
   }
 
   //Gather blocks (contiguous groups of blockSize rows)
@@ -136,6 +137,7 @@ public:
         }
       }
     }
+    X_out.modify_host();
   }
 
   void
@@ -155,6 +157,7 @@ public:
         X_in_j[i_perm] = X_out_j[i];
       }
     }
+    X_out.modify_host();
   }
 
   void
@@ -178,6 +181,7 @@ public:
         }
       }
     }
+    X_out.modify_host();
   }
 
   /******************/
--- a/packages/ifpack2/src/Ifpack2_Details_MultiVectorLocalGatherScatter.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_MultiVectorLocalGatherScatter.hpp
@@ -101,6 +101,7 @@ public:
     using Teuchos::ArrayRCP;
     const size_t numRows = X_out.getLocalLength ();
     const size_t numVecs = X_in.getNumVectors ();
+    Kokkos::fence();
     for (size_t j = 0; j < numVecs; ++j) {
       ArrayRCP<const InScalar> X_in_j = X_in.getData(j);
       ArrayRCP<OutScalar> X_out_j = X_out.getDataNonConst(j);
@@ -124,6 +125,7 @@ public:
     using Teuchos::ArrayRCP;
     const size_t numBlocks = X_out.getLocalLength() / blockSize;
     const size_t numVecs = X_in.getNumVectors ();
+    Kokkos::fence();
     for (size_t j = 0; j < numVecs; ++j) {
       ArrayRCP<const InScalar> X_in_j = X_in.getData(j);
       ArrayRCP<OutScalar> X_out_j = X_out.getDataNonConst(j);
@@ -144,6 +146,7 @@ public:
     using Teuchos::ArrayRCP;
     const size_t numRows = X_out.getLocalLength();
     const size_t numVecs = X_in.getNumVectors();
+    Kokkos::fence();
     for (size_t j = 0; j < numVecs; ++j) {
       ArrayRCP<InScalar> X_in_j = X_in.getDataNonConst(j);
       ArrayRCP<const OutScalar> X_out_j = X_out.getData(j);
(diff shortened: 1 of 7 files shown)


packages/ifpack2/src/Ifpack2_Details_MultiVectorLocalGatherScatter.hpp
```diff
@@ -101,6 +101,7 @@ class MultiVectorLocalGatherScatter {
     using Teuchos::ArrayRCP;
     const size_t numRows = X_out.getLocalLength ();
     const size_t numVecs = X_in.getNumVectors ();
+    Kokkos::fence();
     for (size_t j = 0; j < numVecs; ++j) {
       ArrayRCP<const InScalar> X_in_j = X_in.getData(j);
       ArrayRCP<OutScalar> X_out_j = X_out.getDataNonConst(j);
```
brian-kelley, line 107: Should there be a ``X_out.modify_host()`` at the end of each of these?

I'm also in favor of moving these to device, or even better executing it wherever X_in is most up to date like idot does.

MicheldeMessieres: I updated with modify_host calls for the MV functions (understanding we will probably refactor anyways).

MicheldeMessieres:
@brian-kelley Do you want to do refactor to device in this PR? I suggest we may consider to accept this PR in the current form to get Ifpack2 to a CUDA_LAUNCH_BLOCKING=0 working state. Then address the kernels as a separate PR. I have started a prototype for that which I can open. I think this will give us a better work flow since we'll be modifying the code from a working state.

brian-kelley:
@MicheldeMessieres Yes, I can take care of this later. This PR should be merged now.

MicheldeMessieres:
@brian-kelley Thanks. I can also work on this, if @kddevin agrees. I think we can discuss this in today's Tpetra meeting.
I did some preliminary checks and noted that only the View to View methods are being tested on the basic unit tests. There might be some additional things I need to turn on to test the others.

MicheldeMessieres:
@brian-kelley Also if you end up doing the refactoring I had done a bit of work just to reindex the loops into a single kernel.  I copied those methods below which may be helpful. This doesn't work yet on Cuda because I arbitrarily chose to use the out exec space which won't work if the input is a pure host and the out was a CudaUVMSpace. I expect we'll want to check what can access what and make decisions. Then as you said, the MV cases have the additional complexity that they may be 'ready' on host or device and we can decide based on that. But this code may be a useful starting point for whoever does the refactoring. We may need to get some of the MV loops in testable states first.

```
  /******************/
  /* View <==> View */
  /******************/
  template<typename InView, typename OutView>
  void gatherViewToView(OutView X_out,
                        const InView X_in,
                        const Teuchos::ArrayView<const LO> perm) const
  {
    //note: j is col, i is row
    Kokkos::parallel_for(
      Kokkos::RangePolicy<typename OutView::execution_space, size_t> (0, X_out.size()),
      KOKKOS_LAMBDA (size_t n) {
        size_t j = n / X_out.extent(0);
        size_t i = n % X_out.extent(0);
        const LO i_perm = perm[i];
        X_out(i, j) = X_in(i_perm, j);
    });
  }

  template<typename InView, typename OutView>
  void scatterViewToView(InView X_in,
                         const OutView X_out,
                         const Teuchos::ArrayView<const LO> perm) const
  {
    Kokkos::parallel_for(
      Kokkos::RangePolicy<typename OutView::execution_space, size_t> (0, X_out.size()),
      KOKKOS_LAMBDA (size_t n) {
        size_t j = n / X_out.extent(0);
        size_t i = n % X_out.extent(0);
        const LO i_perm = perm[i];
        X_in(i_perm, j) = X_out(i, j);
    });
  }

  template<typename InView, typename OutView>
  void gatherViewToViewBlock(OutView X_out,
                             const InView X_in,
                             const Teuchos::ArrayView<const LO> perm,
                             LO blockSize) const
  {
    //note: j is col, i is row
    size_t numBlocks = X_out.extent(0) / blockSize;
    Kokkos::parallel_for(
      Kokkos::RangePolicy<typename OutView::execution_space, size_t> (0, X_out.size()),
      KOKKOS_LAMBDA (size_t n) {
        size_t j = n / X_out.extent(0);
        size_t q = n % X_out.extent(0);
        size_t i = q / blockSize;
        size_t k = q % blockSize;
        const LO i_perm = perm[i];
        X_out(i * blockSize + k, j) = X_in(i_perm * blockSize + k, j);
    });
  }

  template<typename InView, typename OutView>
  void scatterViewToViewBlock(InView X_in,
                              const OutView X_out,
                              const Teuchos::ArrayView<const LO> perm,
                              LO blockSize) const
  {
    //note: j is col, i is row
    Kokkos::parallel_for(
      Kokkos::RangePolicy<typename OutView::execution_space, size_t> (0, X_out.size()),
      KOKKOS_LAMBDA (size_t n) {
        size_t j = n / X_out.extent(0);
        size_t q = n % X_out.extent(0);
        size_t i = q / blockSize;
        size_t k = q % blockSize;
        const LO i_perm = perm[i];
        X_in(i_perm * blockSize + k, j) = X_out(i * blockSize + k, j);
    });
  }
```

PR #7759, opened 2020-07-31
tpetra:  Fixing issue #7758:  export with CombineMode ADD for MultiVector 

### LINKED ISSUE #7758: Tpetra:  MultiVector Export / Import errors for Combine Mode ADD
## Bug Report
@trilinos/tpetra 

### Description

Tpetra Vector's Import/Export with CombineMode ADD works correctly if the source map is a superset of the target map.  But when the source map is not a superset of the target map, the ADD operation does not produce correct results.

For example, say the target map is linear block (default Trilinos Map) and the source map is Cyclic (like dealing out cards). 
In this case, existing values in the target vector are not zeroed out before the add operation, so received values are added to the existing values.  If, instead, the source map is overlapped, all existing values are overwritten by the source values before add operations are performed.  

I argue the behavior of the export should not differ depending on whether or not the source map is a superset of the target map.

This issue is related to #7745, in which @jhux2 said the export should not include any existing values in the vector during the export ADD. 

This issue likely doesn't affect many applications.  The superset source map is a very common pattern.  Alternatively, if the application zeros out the target vector before the export / import, the results will match.

### Steps to Reproduce
I'll add a test to Tpetra tests shortly.  Stay tuned.

kddevin:
@trilinos/tpetra @trilinos/epetra 

## Motivation

MultiVector export with CombineMode ADD behaves differently depending on whether or not the target map is a superset of the source map.  See #7758 

The first push to this PR adds a reproducer test.  The test builds vectors with (1) default Trilinos map, (2) cyclic map, (3) overlap map, and (4) superset map.  It exports from a vector with one of these maps to a vector with the default map.  Following the report in #7758, the result of the export is incorrect in export (2)->(1) and (3)->(1).

Subsequent pushes added initialization of the target vector to zero when the number of "sames" and "permutes" is less than the number of entries in the target map.  The most important file to review is Tpetra_MultiVector_def.hpp.

Epetra had the same incorrect behavior.  I pushed changes and an equivalent test for Epetra as well.

## Related Issues

* Closes  #7758

## Testing

Reproducer test tpetra/core/test/MultiVector/Bug7758.cpp and epetra/test/MultiVector/Bug7758.cpp test four scenarios, two of which represent common Trilinos usage and two of which reproduce the error in #7758.
Run on Mac OS X, clang.


kddevin:
@jhux2 @brian-kelley Would you please re-review this PR?  I fixed a few compiler warnings (signed vs unsigned).  Thanks!

kddevin:
I just found a corner case that wasn't covered by this PR.   When an exporter has numSames + numPermutes = 0, copyAndPermute doesn't get called.  But we need it to be called to zero-out the entries that are to be overwritten.  Otherwise, we would need distObject to zero out the entries (but it would need another method that inheriting classes would implement to do that).  Instead, I am allowing copyAndPermute to be called even when numSames+numPermutes=0.
Back to WIP.  Update coming shortly...

PR #7761, opened 2020-08-01
MueLu: Avoid some global communication in setup

cgcgcg:
@trilinos/muelu 

## Motivation
Avoid some Allreduces in setup.


packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_TentativePFactory_kokkos_def.hpp
```diff
@@ -554,10 +554,10 @@ namespace MueLu {
     Set(coarseLevel, "Nullspace", coarseNullspace);
     Set(coarseLevel, "P",         Ptentative);
 
-    if (IsPrint(Statistics1)) {
+    if (IsPrint(Statistics2)) {
```
cgcgcg, line 557: This now matches the non-kokkos version.

PR #7764, opened 2020-08-03
Tacho - condensed graph analysis with group varialbes

kyungjoo-kim:
## Motivation

This PR adds new analysis capability using a condensed graph and group varialbes to improve performance of the analysis phase. It uses condensed graph and evaporate the symbolic analysis with graoup varialbles. This feature is requested from @crdohrm.


kyungjoo-kim:
Warnings are type mismatch when ordinal type = int , size_type = size_t whiel I usually set ordinal type = int, size type = int for testing as they are the combination used by the application.

PR #7767, opened 2020-08-03
C++14 build fixes for Sierra

vbrunini:
@trilinos/tpetra @trilinos/ifpack2 @trilinos/tacho

## Motivation
These changes fix compiler errors when enabling -std=c++14 with Intel 19 and nvcc.

### DIFF
--- a/packages/shylu/shylu_node/tacho/src/impl/Tacho_Chol.hpp
+++ b/packages/shylu/shylu_node/tacho/src/impl/Tacho_Chol.hpp
@@ -50,7 +50,8 @@ namespace Tacho {
         const int ierr = Chol<ArgUplo,ArgAlgo>
           ::invoke(member, _A);
 
-        Kokkos::single(Kokkos::PerTeam(member), [&, ierr]() {
+        Kokkos::single(Kokkos::PerTeam(member), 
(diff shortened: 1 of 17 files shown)


vbrunini:
The lambda capture ones are definitely a compiler bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83204). I'll go through and add comments for those.

The added specialization for Tpetra::withLocalAccess I'm less clear on whether it is a compiler bug or actual language spec issue.

kddevin:
@vbrunini Thank you!  
For the ASC SQ assessors:  can you please add a short comment on how you tested?  Thanks.

crtrott:
Looks good to me. The capture clause things are a clear compiler bug, but the workaround is reasonably harmless. The intel 19 specialization issue is weirder, but if it passes everything I think this is good enough.

vbrunini:
> @vbrunini Thank you!
> For the ASC SQ assessors: can you please add a short comment on how you tested? Thanks.

Built & tested intel-19 and cuda-10/gcc-7.2 builds with C++14 enabled locally. Beyond that I'm relying on the PR testing process.

kddevin:
Thanks, @vbrunini 
Merging now.

crtrott:
we may wanna switch at least one of the CUDA builds to C++14 in the PR testing

PR #7769, opened 2020-08-03
Stratimikos: add bicgstab to belos via stratimikos options

rstumin:
adds bicgstab to Stratimikos/belos


rstumin:
Tested using Stratimikos_simple_stratimikos_example.exe using M22.mtx and replacing GMRES part of belos.ifpack.xml with 
a BiCGStab option.

PR #7771, opened 2020-08-03
MueLu: region w-cycle and symmetric Gauss-Seidel

pohm01:
@trilinos/muelu 

## Motivation
Adds the options for a W-cycle and Symmetric Gauss-Seidel smoother for region multigrid.

### DIFF
--- a/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
+++ b/packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
@@ -82,7 +82,8 @@ std::map<std::string, int> getListOfValidSmootherTypes()
   smootherTypes.insert(std::pair<std::string, int>("None",      0));
   smootherTypes.insert(std::pair<std::string, int>("Jacobi",    1));
   smootherTypes.insert(std::pair<std::string, int>("Gauss",     2));
-  smootherTypes.insert(std::pair<std::string, int>("Chebyshev", 3));
+  smootherTypes.insert(std::pair<std::string, int>("SymmetricGauss",     3));
+  smootherTypes.insert(std::pair<std::string, int>("Chebyshev", 4));
 
   return smootherTypes;
 }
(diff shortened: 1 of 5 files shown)


packages/muelu/research/mmayr/composite_to_regions/src/SetupRegionSmoothers_def.hpp
```diff
@@ -260,6 +262,25 @@ void GSIterate(RCP<Teuchos::ParameterList> smootherParams,
       OneregX[k] += ldelta[k];
     }
     zeroInitGuess = false;
+
+    if( sgs ){
+      for (size_t k = numRows; k--;) {
```
mayrmt, line 267: Shouldn't this be
```suggestion
      for (size_t k = numRows - 1; k >= 0; k--) {
```
to start looping at the last row and then move backward to the first row?

pohm01: I'm not sure why, but `for (size_t k = numRows - 1; k >= 0; k--)` does not work. It is some sort of c++ thing that I don't fully understand.
I did check that `for (size_t k = numRows; k--;)` correctly iterates from last row to first row (starts at `numRows-1` and ends at `0`).

mayrmt: I was just curious. If it works, then I'm fine with it.

brian-kelley: @pohm01 @mayrmt size_t is unsigned, so if it's 0 and you decrement it, it wraps around to 2^64 - 1. So k >= 0 is true no matter what k is. You could use signed counter, or loop k from 0 to numRows and then use ``numRows - k - 1`` in the loop.

mayrmt:
@pohm01 Feel free to merge this, whenever you think this is ready.

PR #7773, opened 2020-08-03
Tpetra: optionally disable ownership check in FECrsGraph

### LINKED ISSUE #7697: Tpetra: should FECrsGraph node ownership debug check be relaxed?
## Question
@trilinos/tpetra 

Currently, there's a check performed by FECrsGraph during the call to `doOwnedPlusSharedToOwned`, which, if not successful, prints the error message:

```
"if you own an element (in the finite element sense) you "
"must also own one of the attached nodes.  This assumption has been violated in "
"your matrix fill."
```

However, this check seems a bit restrictive for folks that receive meshes partitioned by other tools. If you consider an unstructured triangular mesh, a partitioning will likely generate a zig-zagged border. It is not unlikely that one element might have all nodes belonging to the other process.

Yes, one could probably work out some rebalancing routine, to adjust these scenarios. However, it might take some work, and in some cases it might require diving into some complex data structures (e.g., stk meshes). Is there a reason why Tpetra should not like the idea of a rank owning an element without owning any of its nodes? After all, Tpetra does not even know what a mesh is, so why should it care about who owns (mesh) elements? Upon assembly, the matrix is exactly the same (modulo possibly some changes in the column map LIDs), regardless of who owned the element to start with.

bartgol:
The user is now allowed to disable the check that ensures that for each
owned element, there is at least one owned node. This check is helpful
to keep the column map optimally sized, but for some application it might
be difficult to rearrange the mesh ownership, to the point that it might
be easier to accept a small performance hit during mat-vec product (due
to import of not needed gids).

Fixes #7697.

@trilinos/tpetra 

## Motivation

The FECrsGraph class checks (in debug mode) that there are no owned elements for which none of the nodes is owned, since it means we will import unnecessary entries during mat-vec products.
However, some apps (in this case, Albany) may be dealing with meshes coming from TPLs, and/or might have a hard time redistributing mesh entities, so that this does not occur. While we want them to fix their meshes (to avoid perf loss), it might be hard for them (at least in the short term) to do so, perhaps since their mesh structs are quite involved, and there are only a handful of such "bad" elements, so they might be willing to accept the (small) perf hit rather than rearranging the mesh.

Also, fixing the way the check is carried out, by performing a reduction over the graph's comm, so that if one rank fails, all the ranks in the comm group are notified, and are not left hanging (as it is done already in several other checks in Tpetra).

## Related Issues

* Closes 7697

## Testing

This PR should _not_ affect existing code, since the default behavior is to still perform the check. The client Albany is however using this feature, and _does_ have meshes with this kind of "bad" elements. Tests are passing in Albany using this feature (they were not before).

### DIFF
--- a/packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
@@ -286,11 +286,20 @@ void FECrsGraph<LocalOrdinal, GlobalOrdinal, Node>::doOwnedPlusSharedToOwned(con
           flag[indices[j]] = true;
       }
       
-      bool lclSuccess = true;
+      bool success = true;
       for(size_t i=0; i<(size_t)flag.size(); i++)
-        if(!flag[i]) {lclSuccess=false; break;}
+        if(!flag[i]) {success=false; break;}
+
+      // Perform a reduction over the input comm, so that ranks with success=true won't be hanging.
+      // Note: this only ensures things don't hang 
+      int lclSuccess = success ? 1 : 0;
+      int gblSuccess = lclSuccess;
+      auto comm = this->getComm();
+      if (!comm.is_null()) {
+        Teuchos::reduceAll<int,int>(*comm, Teuchos::REDUCE_MIN, 1, &lclSuccess, &gblSuccess);
+      }
       TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC
-        (lclSuccess ==false,
+        (gblSuccess == 0,
          std::invalid_argument, "if you own an element (in the finite element sense) you "
          "must also own one of the attached nodes.  This assumption has been violated in "
          "your matrix fill.");
--- a/packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
@@ -269,8 +269,13 @@ void FECrsGraph<LocalOrdinal, GlobalOrdinal, Node>::doOwnedPlusSharedToOwned(con
     crs_graph_type::fillComplete(this->domainMap_,this->getRowMap());
 
     // In debug mode, we check to make sure the "if you own an element, you own at least one of its nodes"
+    // However, we give the user the ability to bypass this check. This can be useful when dealing with mesh
+    // that are generated and partitioned by a TPL, and/or where it is so complicated to rearrange the
+    // dofs ownership that the customer app might be willing to accept a small performance hit.
     const bool debug = ::Tpetra::Details::Behavior::debug ();
-    if (debug) {
+    const bool enforceAtLeastOneOwnedColIndexPerRow =
+        this->getMyNonconstParamList().is_null() ? true : this->getMyNonconstParamList()->get("Enforce At Least One Owned Col Index Per Row",true);
+    if (debug && enforceAtLeastOneOwnedColIndexPerRow) {
       Teuchos::RCP<const map_type> colmap = this->getColMap();
       Teuchos::Array<bool> flag(colmap->getNodeNumElements(),false);
       Teuchos::Array<LocalOrdinal> indices(this->getNodeMaxNumRowEntries());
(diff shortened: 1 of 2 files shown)


bartgol:
Correction: tests are still not passing in Albany due to the new parameter not being found in the list of valid parameters (despite FECrs overriding the `getValidParameters` method). Currently investigating the issue.

bartgol:
The code in FECrsGraph is actually ok, the issue is the way we used it from Albany. That raised a very linked question though, which should be addressed before deciding whether this PR is ok. @csiefer2 , I'd like to hear your opinion on this.

The issue was this:

In Albany, I was creating a ParameterList, in which I set the new option for FECrsGraph ('Enforce At Least One Owned Col Index Per Row'), and I set it to false in order to bypass the test. I overridden the `getValidParameters` method in FECrsGraph, hoping it would get my new parameter accepted during the input pl check. However, that was *not* the case. Here's why. The input pl is passed to the CrsGraph c-tor, which calls `resumeFill(params)`; inside resumeFill, CrsGraph checks that the input params are valid. However, _since we are still in the constructor_, we do NOT get virtual dispatch for the call to `getValidParameters`, which ends up calling `CrsGraph::getValidParameters`, instead of `FECrsGraph::getValidParameters`.

The solution on the Albany side was simple: do NOT pass the parameter list to the FECrsGraph constructor; instead, call `setParameterList` _right after_ construction (or, at least, before `endFill`).

However, this makes me wonder whether FECrsGraph ctor should _not_ pass the input PL to CrsGraph's ctor, and instead set the parameter list inside the body of FECrsGraph's ctor. After all, in CrsGraph's ctor, the PL is only used to call `setParameterList`, but its content is _not_ used at construction time, so we can set it after CrsGraph's ctor has completed.

packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
```diff
@@ -281,11 +286,20 @@ void FECrsGraph<LocalOrdinal, GlobalOrdinal, Node>::doOwnedPlusSharedToOwned(con
           flag[indices[j]] = true;
       }
       
-      bool lclSuccess = true;
+      bool success = true;
       for(size_t i=0; i<(size_t)flag.size(); i++)
-        if(!flag[i]) {lclSuccess=false; break;}
+        if(!flag[i]) {success=false; break;}
```
mperego, line 291: Why don't we use a counter instead ofe the boolean `success` and count the number of entries with flag[i]==true? In this way when we error out we can say how many nodes in the column map are not attached to any row.

bartgol: I can do that.

packages/tpetra/core/src/Tpetra_FECrsGraph_def.hpp
```diff
@@ -281,11 +286,20 @@ void FECrsGraph<LocalOrdinal, GlobalOrdinal, Node>::doOwnedPlusSharedToOwned(con
...
+      // Note: this only ensures things don't hang 
+      int lclSuccess = success ? 1 : 0;
+      int gblSuccess = lclSuccess;
+      auto comm = this->getComm();
+      if (!comm.is_null()) {
+        Teuchos::reduceAll<int,int>(*comm, Teuchos::REDUCE_MIN, 1, &lclSuccess, &gblSuccess);
+      }
       TEUCHOS_TEST_FOR_EXCEPTION_CLASS_FUNC
-        (lclSuccess ==false,
+        (gblSuccess == 0,
```
mperego, line 302: It's better to indicate that it is possible to disable this check.

bartgol: I'm not sure what you mean. This check is wrapped inside an if statement, and there's a comment right above the if statement, that explains that we are giving the option to bypass the check.

mperego: > I'm not sure what you mean. This check is wrapped inside an if statement, and there's a comment right above the if statement, that explains that we are giving the option to bypass the check.

I meant in the error message. It's true that one can figure that out looking at the code. Anyway, it's not a big deal.

bartgol: Ah, I see. Done.

mperego:
@trilinos/framework something's wrong with autotester here. I tried to restart it but nothing happened.

cgcgcg:
@mperego That's because `AT:WIP` is still set..

mperego:
> 
> 
> @mperego That's because `AT:WIP` is still set..

Oh gosh.. I didn't notice that. Sorry for pinging you and for blaming the autotester, and thanks for the quick reply.

PR #7782, opened 2020-08-05
Add Debug build into config map

prwolfe:
@trilinos/framework 

## Motivation
Found a missing config entry during testing

## Testing
This was found when running the Debug job from Jenkins by hand. I will replicate that directly and note the results here.


prwolfe:
Can I get a review on this when someone gets a chance. It seems to be solid now.

PR #7783, opened 2020-08-05
FEI patch for astra from Mike glass.

prwolfe:
From his email -

I don't know if any other ASC apps are using the FEI package in Trilinos but we (Sierra) have come across a compiler issue on stria related to this package. We're using the ARMHPC 20.1 toolset. The following is our work around. Can this get incorporated into the Trilinos repo?

## Motivation
This is required for the 20.1 toolset on astra - see above

## Testing
This came in from the Sierra project and was tested via their work. Currently none of the dashboard lines on astra build FEI, so any additional testing will need to be done by hand or fed added.


prwolfe:
@alanw0 - can you look at this one? I expect it's ok but you should review/

Issue #7801, opened 2020-08-07
STK:  Broke clang


jrood-nrel:
## Bug Report
@trilinos/stk

### Description
I'm seeing any build I have of Trilinos with clang failing with:
```
packages/stk/stk_mesh/stk_mesh/base/CreateFaces.cpp:133:46: error: implicit instantiation of undefined template
```

jhux2:
@alanw0 @trilinos/stk

alanw0:
We'll take a look and see what happened.

jvo1012:
We just pushed in a change that we believe will fix this error.  Please let us know otherwise.

jrood-nrel:
I think this fixed Clang 10, but I still see errors with Clang 7.0.1. I'm not exactly sure what the issue is but you can see the errors I'm seeing here https://my.cdash.org/build/1866607/notes :
```
packages/stk/stk_unit_test_utils/stk_unit_test_utils/stk_mesh_fixtures/QuadFixture.cpp:243:31: error: implicit instantiation of undefined template 'std::__1::array<double, 3>' std::array<double, 3> temp;
```

jrood-nrel:
Actually I just tried this myself using Clang 10 and I'm seeing this issue, so there is definitely still a problem.

jhux2:
> We just pushed in a change that we believe will fix this error. Please let us know otherwise.

Can you link the PR to this issue?

jrood-nrel:
I was able to build with clang on my laptop just now. I updated trilinos for the nalu-wind tests, so I will see if this is fixed tomorrow.

jrood-nrel:
Looks like this has been fixed.

PR #7804, opened 2020-08-07
STK: fix clang build

jvo1012:
@trilinos/<teamName>


alanw0:
This is expected to fix #7801.

PR #7805, opened 2020-08-07
Zoltan html documentation changes

kddevin:
@trilinos/zoltan 

## Motivation
Sandia is mucking with its websites and certificates.  This PR attempts to accommodate their changes, including changing http to https wherever possible, and changing www.cs.sandia.gov to cs.sandia.gov.

While I was at it, I updated some affiliations and other out-of-date info (e.g., filing bugs in bugzilla).

## Testing

NA. No code changes.  I did look at the HTML and click a lot of links.


egboman (approved):
Looks reasonable, but I did not actually download the html and look at the formatted version. I suggest to update the authors (add Seher, move some folks to "past/former").

Erik

Issue #7807, opened 2020-08-08
ifpack:  memory leak in Ifpack_Hypre


ecoon:
## Bug Report
@trilinos/ifpack

### Description

Normal usage of Ifpack_Hypre leaks memory.

### Steps to Reproduce
1. SHA1:  688a924
2. Configure script:  (built through cmake, standard configure of Epetra stack)
3. Configure log:  (if needed I can go searching...)
4. Build log:  (same as configure)
5. Input deck:  [attach here]
[solver_pc_boomeramg.cc.txt](https://github.com/trilinos/Trilinos/files/5045269/solver_pc_boomeramg.cc.txt)

See attached unittest. 

6. Compile and run valgrind:

`valgrind --leak-check=full solver_pc_boomeramg &> valgrind_out.txt`

7. Run log:
[valgrind_out.txt](https://github.com/trilinos/Trilinos/files/5045272/valgrind_out.txt)

### Why is it broken?

Effectively the issue is the following path through the code:

1. Use Hypre as a Preconditioner `SolverOrPreconditioner == Preconditioner`
2. Supply no Solver.  Ifpack_Hypre defaults to `PCG`.
 https://github.com/trilinos/Trilinos/blob/master/packages/ifpack/src/Ifpack_Hypre.cpp#L73
3.  Despite not being asked to supply a Solver, Ifpack calls CreateSolver, which since there is a default value of a real solver (e.g. not NULL or comparable), Ifpack creates the object.  This allocates memory (in Hypre)
  https://github.com/trilinos/Trilinos/blob/master/packages/ifpack/src/Ifpack_Hypre.cpp#L166
4. When `Compute()` is called, `SolverSetupPtr_` is _not_ called, because there it correctly checks and sees that a solver was not requested.  As a result, the flag  `IsSolverSetup_[0] == false`
https://github.com/trilinos/Trilinos/blob/master/packages/ifpack/src/Ifpack_Hypre.cpp#L279
5. Since `IsSolverSetup_[0] == false`, `Destroy()` does _not_ call `DestroySolverPtr_`: 
https://github.com/trilinos/Trilinos/blob/master/packages/ifpack/src/Ifpack_Hypre.cpp#L142

Not calling `DestroySolverPtr_` leaks the memory.  Note, I have not confirmed that DestroySolverPtr_ can be called on a created but not setup Hypre solver, but one would hope so.

[...]

PR #7808, opened 2020-08-08
ifpack: Issue/7807 ifpack hypre memory leak

ecoon:
@trilinos/ifpack

## Motivation

See #7807 

## Testing

I believe there are no Ifpack_Hypre tests currently.  In the issue, I included source for a test, and we maintain a set of tests for this capability in Amanzi because we use it often.

## Additional Information

valgrind clean!


ecoon:
Ummm, it looks like I screwed this up somehow.  Maybe because I based the branch I created initially off of master initially instead of develop?  Help please?  This should consist of a single commit with changes to two files, and that commit ( 2e0c851 ) could be cherry-picked onto something else if needed.

ecoon:
closed due to contributor error...

PR #7809, opened 2020-08-08
ifpack: fixes missing memory leak in Ifpack_Hypre

### LINKED ISSUE #7807: ifpack:  memory leak in Ifpack_Hypre

ecoon:
@trilinos/ifpack

Apologies for the last pull request which was "dirty" since the commit was based on master.  This is on develop now, so should include only one commit...

## Motivation

Closes #7807 

## Testing

I believe there are no Ifpack_Hypre tests currently.  In the issue, I included source for a test, and we maintain a set of tests for this capability in Amanzi because we use it often.

## Additional Information

valgrind clean...

### DIFF
--- a/packages/ifpack/src/Ifpack_Hypre.cpp
+++ b/packages/ifpack/src/Ifpack_Hypre.cpp
@@ -74,10 +74,10 @@ Ifpack_Hypre::Ifpack_Hypre(Epetra_RowMatrix* A):
   PrecondType_(Euclid),
   UsePreconditioner_(false)
 {
-  IsSolverSetup_ = new bool[1];
-  IsPrecondSetup_ = new bool[1];
-  IsSolverSetup_[0] = false;
-  IsPrecondSetup_[0] = false;
+  IsSolverCreated_ = new bool[1];
+  IsPrecondCreated_ = new bool[1];
+  IsSolverCreated_[0] = false;
+  IsPrecondCreated_[0] = false;
   MPI_Comm comm = GetMpiComm();
   // Check that RowMap and RangeMap are the same.  While this could handle the
   // case where RowMap and RangeMap are permutations, other Ifpack PCs don't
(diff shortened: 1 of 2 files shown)


cgcgcg:
@trilinos/ifpack

csiefer2:
@cgcgcg and I have a branch with a bunch of related fixes.  I'll need to check that against this.

ecoon:
@csiefer2 Is this being pulled into that branch then?  Would be nice to have this fixed as we currently have to patch it downstream.

cgcgcg:
@ecoon It's on my list of things to do...

cgcgcg:
Taken care of via #8060

PR #7811, opened 2020-08-09
Zoltan2 Sphynx: Add options for initial guess and full/partial ortho

cgcgcg:
@trilinos/zoltan2 

## Motivation
Add option for a blockwise constant initial guess and full/partial orthogonalization.

### DIFF
--- a/packages/zoltan2/sphynx/src/Zoltan2_Sphynx.hpp
+++ b/packages/zoltan2/sphynx/src/Zoltan2_Sphynx.hpp
@@ -517,6 +517,7 @@ namespace Zoltan2 {
     bool isHermitian = true;
     bool relLockTol = false;
     bool lock = false;
+    bool useFullOrtho = true;
 
     // Information to output in a verbose run
     int numfailed = 0;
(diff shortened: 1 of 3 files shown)


packages/zoltan2/sphynx/src/Zoltan2_Sphynx.hpp
```diff
@@ -558,14 +563,41 @@ namespace Zoltan2 {
     anasaziParams.set("Orthogonalization", ortho);
     anasaziParams.set("Use Locking", lock);
     anasaziParams.set("Relative Locking Tolerance", relLockTol);
+    anasaziParams.set("Full Ortho", useFullOrtho);
 
+    typedef typename Adapter::user_t graph_type;
+    typedef typename graph_type::global_ordinal_type GO;
```
seheracer, line 569: Global ordinal is already defined as gno_t in line 105.  Do we need GO?

seheracer:
Thanks, @cgcgcg! Changes look good for now 👍

packages/zoltan2/sphynx/src/Zoltan2_Sphynx.hpp
seheracer: Resolved, thanks!

cgcgcg: No, we don't :+1:

PR #7812, opened 2020-08-09
SEACAS IOSS: Add environment variable to set common_nodes to 2

seheracer:
@trilinos/seacas @cgcgcg @egboman 

## Motivation

This PR adds the option of using common_nodes=2, which will be used when constructing the dual graph in ParMETIS_V3_PartMeshKway.  When common_nodes=2, elements which share a common edge are connected in the dual graph. This will increase the number of connections in the dual graph (unless the elements are triangular), so captures the total communication volume more correctly in the case of an edge-based discretization.

I obtained up to 10% improvement in the solve time when I used common_nodes=2 on a tetrahedral mesh.

The matching PR in the seacas repo: https://github.com/gsjaardema/seacas/pull/184

### DIFF
--- a/packages/seacas/libraries/ioss/src/Ioss_Decomposition.C
+++ b/packages/seacas/libraries/ioss/src/Ioss_Decomposition.C
@@ -135,6 +135,12 @@ namespace {
     Ioss::ParallelUtils par_util(comm);
     common_nodes = par_util.global_minmax(common_nodes, Ioss::ParallelUtils::DO_MIN);
 
+    // Overwrite common_nodes if user wants to set it to 2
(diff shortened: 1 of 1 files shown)


gsjaardema:
I think the idea for this is good, but I would prefer a different implementation that uses the existing `IOSS_PROPERTIES` environement variable method for setting properties.  Maybe the property `PARMETIS_COMMON_NODE_COUNT` so the user would do something like:
```
export IOSS_PROPERTIES="PARMETIS_COMMON_NODE_COUNT=2:..."
```
I will modify the code to fit this recommendation and see if it still solves the same problem.  Hopefully sometime today...

cgcgcg:
@gsjaardema I see that you already implemented this change in seacas :+1:  Is it possible to get the change into Trilinos?

PR #7816, opened 2020-08-10
IOSS: Allow use of property to set parmetis kway common_node count

gsjaardema:
@trilinos/seacas


gsjaardema:
This is an alternative implementation of #7812

cgcgcg:
@gsjaardema Thank you!

seheracer:
Thanks, @gsjaardema!

PR #7820, opened 2020-08-10
Ifpack. Removing "virtual inline" functions to make it standard compliant

### LINKED ISSUE #7802: Ifpack:  Removing "virtual inline" functions to make it standard compliant
## Enhancement
@trilinos/ifpack

There are virtual inline functions declared in ifpack. They are defined in a different translation unit, which makes it a non-standard C++.
gcc gives a warning, nvc++ gives an error in a pedantic mode. Removing 'inline' fixes the issue.

alenacpp:
@trilinos/ifpack

## Motivation
There are virtual inline functions declared in ifpack. They are defined in a different translation unit, which makes it a non-standard C++.
gcc gives a warning, nvc++ gives an error in a pedantic mode. Removing 'inline' fixes the issue.

## Related Issues

* Closes #7802 

## Testing
This is a change in the existing Ifpack code. Should be covered by Ifpack tests.


cgcgcg:
@trilinos/ifpack

jhux2:
@alenacpp Thank you for the contribution.

PR #7830, opened 2020-08-12
MueLu: reorganize directory structure of region MG code

### LINKED ISSUE #7824: MueLu: reorganize regionMG code
## Enhancement
@trilinos/muelu 

The regionMG code started out as purely experimental code in `packages/muelu/research/mmayr/composite_to_regions/` and relied on the interplay of some C++ executable with some Matlab scripts. Meanwhile, a pure C++ implementation is available in `packages/muelu/research/regionMG/`.

The first Matlab-based version still compiles, but has not been used in more than a year. All developments are only carried out in the new C++ version. To prepare for a more mature implementation of the regionMG framework, @lucbv and I have discussed some ideas on how to restructure the current source code directories. This issue is intended to agree on a new directory structure, which will then be implemented.

## Proposal

What about this new directory structure:
```
packages/muelu/research/regionMG/
    |__doc/
    |__examples/
    |__src/
```
where
- the `src/` folder will contain the `SetupRegion*.hpp` files currently living in `packages/muelu/research/mmayr/composte_to_regions/`
- the `examples/` folder will contain the driver `Driver_Structured_Regions.cpp` and the associated `xml`-files

Then, 
- the existing directory `packages/muelu/research/mmayr/hhg/` will be deleted
- the remaining files in `packages/muelu/research/mmayr/composte_to_regions/` can be removed
- finally, we can probably remove the entire `packages/muelu/research/mmayr/` directory

mayrmt:
@trilinos/muelu 
@lucbv @pohm01 @rstumin 

## Motivation

This PR aims at cleaning up the region MG source code and its directory structure in order to separate useful and more mature files from old, unused, and experimental files.

The core capabilities of the current implementation of region MG are now moved into their own directory to separate them from very experimental code from the early days of region MG. This early code is deleted, since it does not offer any other than purely historic value and actually sometimes hinders development.

## Related Issues

* Closes #7824 

## Stakeholder Feedback

This work is part of an ASCR project with @rstumin and @lucbv.

## Testing

All existing tests have been migrated along with the source code.

As reported in #7832, a single unit test has been already failing in my machine before these changes, though it passes on other machines.


lucbv:
@mayrmt I am OK with these changes but I somehow would like to keep an `example` folder.
In my mind while we do perform a lot of tests using the current driver it is also used to perform at scale runs and such activities which I would consider more like examples than test...
But I agree that separating the test and example utility of the current driver makes sense, any idea how we can reflect that cleanly in our tree structure?
I am wondering if we can have examples and tests rely on the same source code from the driver and simply have CMake build the executable in both locations?
I have also been thinking about writing a new example (test) using `percept` but the idea is not fully mature yet...

mayrmt:
@lucbv I agree that having an `examples/` folder would be nice. Let's merge this for now. We can revisit that topic in the future, once we know, if there's another example to be added.

PR #7836, opened 2020-08-13
Patch an issue that seems unique to certain clang builds.

alanw0:
Also remove a debug-print statement that crept in recently.

@trilinos/<teamName>


spdomin:
@jvo1012, I was hoping that we would have a successful Mac build/test result before this was approved:) We will hope for the best and I will report my Mac results on the main support ticket. Thanks.

PR #7841, opened 2020-08-14
Tempus: Bug in RK Forward Euler with FSAL

### LINKED ISSUE #7835: Tempus: Bug in RK Forward Euler with FSAL
## Bug Report
@trilinos/tempus 

### Description
While debugging an issue with Albany, it turns out that for useFSAL=true and with one stage, Explicit RK does not correctly calculate the update for the next time step, e.g., evalModel() will never be called.

### Steps to Reproduce
Step up an 'RK Forward Euler' stepper, set useFSAL=true, and try to take a few steps.  The solution will not change.

ccober6:
@trilinos/tempus 

## Motivation
While debugging an issue with Albany, it turns out that for
useFSAL=true and with one stage, Explicit RK does not correctly
calculate the update for the next time step, i.e., evalModel() will
never be called.

 * Changed the default useFSAL to true for RK Forward Euler.  This
   matches Forward Euler stepper.
 * Added xDot to Dahlquist test problem for unit testing.
   - New constructor allows construction with or without xDot
   - Added xDot to nominalValues() and getExactSolution().
 * Added unit test to cover this bug.
 * BDF2 used RK Forward Euler as a startup stepper. Switched it
   to RK Explicit Midpoint, which is second order.  This allowed BDF2
   to remain second order, but the overall error decreased.  With the
   new default useFSAL=true, the workingState->getXDot() was updated
   by RK Forward Euler which interfers with the BDF2.

## Related Issues

* Closes #7835 

## Stakeholder Feedback
@ikalash is the stakeholder.

## Testing
Added a unit test to cover this bug.

### DIFF
--- a/packages/tempus/src/Tempus_StepperExplicitRK_impl.hpp
+++ b/packages/tempus/src/Tempus_StepperExplicitRK_impl.hpp
@@ -363,6 +363,17 @@ void StepperExplicitRK<Scalar>::takeStep(
       }
     }
 
+    if (this->getUseFSAL()) {
+      if (numStages == 1) {
+        const Scalar ts = time + dt;
+        auto p = Teuchos::rcp(new ExplicitODEParameters<Scalar>(dt));
+        // Evaluate xDot = f(x,t).
+        this->evaluateExplicitODE(stageXDot_[0], workingState->getX(), ts, p);
+      }
+      if (workingState->getXDot() != Teuchos::null)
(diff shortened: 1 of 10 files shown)


ccober6:
@sconde  Can you review this?  Thanks!

ccober6:
@ikalash Tagging for FYI.

sconde:
If it's a single stage explicit RK scheme, why not just default to using Forward Euler?

PR #7843, opened 2020-08-15
Intrepid2: vectorNorm fencing

MicheldeMessieres:
I'm not sure we'd want to do things this way, but opening the PR for discussion.

We need to fence the Serial vectorNorm for UVM. However the Serial vectorNorm is also used as the core of the kernel form, when we would not want to fence. We could resolve as in this PR, or fence every call of Serial vectorNorm which I don't think we'd want, or perhaps design this a different way.

Potentially this is a backwards compatibility issue if external code is using Serial vectorNorm in device kernels.

@trilinos/intrepid2 

## Motivation

To get Intrepid2 passing all unit tests with CUDA_LAUNCH_BLOCKING=0 in preparation for further work to analyze core fences in panzer.

## Testing

White Cuda Intrepid2 test suite.

### DIFF
--- a/packages/intrepid2/src/Cell/Intrepid2_CellToolsDefPhysToRef.hpp
+++ b/packages/intrepid2/src/Cell/Intrepid2_CellToolsDefPhysToRef.hpp
@@ -181,7 +181,10 @@ namespace Intrepid2 {
 
       // Average L2 error for a multiple sets of physical points: error is rank-2 (C,P) array 
       rst::vectorNorm(errorCellwise, errorPointwise, NORM_ONE);
-      const auto errorTotal = rst::Serial::vectorNorm(errorCellwise, NORM_ONE);
+
+      auto errorCellwise_h = Kokkos::create_mirror_view(errorCellwise);
+      Kokkos::deep_copy(errorCellwise_h, errorCellwise);
+      const auto errorTotal = rst::Serial::vectorNorm(errorCellwise_h, NORM_ONE);
     
       // Stopping criterion:
       if (errorTotal < tol) 
--- a/packages/intrepid2/unit-test/Discretization/Basis/HGRAD_HEX_C1_FEM/test_02.hpp
+++ b/packages/intrepid2/unit-test/Discretization/Basis/HGRAD_HEX_C1_FEM/test_02.hpp
@@ -609,7 +609,9 @@ namespace Intrepid2 {
             
                 rst::subtract(interpolant, exact_solution);
 
-                ValueType relNorm = rst::Serial::vectorNorm(interpolant, NORM_TWO);
+                auto interpolant_h = Kokkos::create_mirror_view(interpolant);
+                Kokkos::deep_copy(interpolant_h, interpolant);
+                ValueType relNorm = rst::Serial::vectorNorm(interpolant_h, NORM_TWO);
                               
                 *outStream << "\nRelative norm-2 error between exact solution polynomial of order ("
                            << x_order << ", " << y_order << ", " << z_order
--- a/packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
+++ b/packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
@@ -365,9 +365,11 @@ namespace Intrepid2 {
           art::contractFieldFieldScalar(out1_c_l_r, in_c_l_p, in_c_r_p);
           art::contractFieldFieldScalar(out2_c_l_r, in_c_l_p, in_c_r_p);
           rst::subtract(out1_c_l_r, out2_c_l_r);
-          if (rst::Serial::vectorNorm(out1_c_l_r, NORM_ONE) > tol) {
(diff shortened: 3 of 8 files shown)


mperego:
@kyungjoo-kim do you mind having a look at this? Thanks!

packages/intrepid2/src/Shared/Intrepid2_RealSpaceTools.hpp
```diff
@@ -93,7 +93,7 @@ namespace Intrepid2 {
       KOKKOS_INLINE_FUNCTION
       static inVecValueType
       vectorNorm( const Kokkos::DynRankView<inVecValueType,inVecProperties...> inVec, 
-                  const ENorm normType );
+                  const ENorm normType, bool bFence = true );
```
kyungjoo-kim, line 96: This is Kokkos inline function which can be used inside of a functor. We cannot put a fence here.

packages/intrepid2/src/Shared/Intrepid2_RealSpaceToolsDef.hpp
```diff
@@ -85,6 +85,11 @@ namespace Intrepid2 {
     const ordinal_type kend = inVec.extent(2);
     const ordinal_type lend = inVec.extent(3);
     const ordinal_type mend = inVec.extent(4);
+
+    if(bFence) {
+     Kokkos::fence();
+    }
```
kyungjoo-kim, line 91: fence here is not allowed.

MicheldeMessieres:
@kyungjoo-kim There are a bunch of intrepid2 tests which use it directly as a host call.

For example Shared/ArrayTools/test_01.hpp has:

```
          if (rst::Serial::vectorNorm(out1_c_l_r, NORM_ONE) > tol) {
            *outStream << "\n\nINCORRECT contractFieldFieldScalar (1): check COMP_CPP vs. COMP_BLAS; "
                       << " diff-1norm = " << rst::Serial::vectorNorm(out1_c_l_r, NORM_ONE) << "\n\n";
            errorFlag = -1000;
          }
```
So the bFence was just to make sure the fence was only used when it was placed in the functor. I could also fix this by manually fencing all the unit tests where they call vectorNorm directly. Or maybe the unit tests should to be using it in this way.

mperego:
Also here:
https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Cell/Intrepid2_CellToolsDefPhysToRef.hpp#L180

https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Discretization/FunctionSpaceTools/Intrepid2_FunctionSpaceToolsDef.hpp#L510

kyungjoo-kim:
Oh..... The test is wrong. The view should be mirrored into the host before it calculate the norm and check the value.

mperego:
@kyungjoo-kim never mind my comment.. those are not using the serial version

MicheldeMessieres:
@kyungjoo-kim Got it. Thanks. I will remove the internal fence stuff and fix the tests locally to mirror to host first.

mperego:
@MicheldeMessieres @kyungjoo-kim OK, so we need to fix several tests. @MicheldeMessieres are you up for it, or would you like the @trilinos/intrepid2 to fix them?

MicheldeMessieres:
@mperego I will take care of it.

kyungjoo-kim:
Thanks. Those tests are a kind of direct translation of the old code a few years ago. At that time, our Kokkos-fu was so weak.

MicheldeMessieres:
@mperego @kyungjoo-kim I've pushed an update which resolves all the Serial::vectorNorm calls to be on host. Also fixed a few device reads that were directly connected. There are probably more places throughout Trilinos tests that we are missing which are already protected by a fence somewhere. I expect we won't track all of those down until we run these tests on pure CudaSpace. But this will pass with CUDA_LAUNCH_BLOCKING=0.

Also I fixed some minor typos in output debug statements in test_02.hpp for ArrayTools. There were 2 spots printing data_c_p which should have been data_c_1.

kyungjoo-kim:
@kddevin  @jwillenbring I have one question about testing Trilinos without cuda launch blocking. Are we going to have a PR test without cuda launch blocking so that we no longer introduce this type of errors after @MicheldeMessieres fixed all ?

mperego:
@kyungjoo-kim yes, I think that's the plan.

mperego:
> @mperego @kyungjoo-kim I've pushed an update which resolves all the Serial::vectorNorm calls to be on host. Also fixed a few device reads that were directly connected. There are probably more places throughout Trilinos tests that we are missing which are already protected by a fence somewhere. I expect we won't track all of those down until we run these tests on pure CudaSpace. But this will pass with CUDA_LAUNCH_BLOCKING=0.
> 
> Also I fixed some minor typos in output debug statements in test_02.hpp for ArrayTools. There were 2 spots printing data_c_p which should have been data_c_1.

@MicheldeMessieres Thanks Michel for fixing that!

PR #7844, opened 2020-08-15
Phalanx: Fix test for CUDA_LAUNCH_BLOCKING=0

MicheldeMessieres:
@trilinos/phalanx 

## Motivation

Fixing packages to run with CUDA_LAUNCH_BLOCKING=0

## Testing

White Cuda build Phalanx unit tests with CUDA_LAUNCH_BLOCKING=0

### DIFF
--- a/packages/phalanx/test/Kokkos/tKokkos.cpp
+++ b/packages/phalanx/test/Kokkos/tKokkos.cpp
@@ -608,6 +608,7 @@ namespace phalanx_test {
 	  });
 	auto host_f = Kokkos::create_mirror_view(f);
 	Kokkos::deep_copy(host_f,f);
+	Kokkos::fence();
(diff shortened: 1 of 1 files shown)


MicheldeMessieres:
@rppawlo Thanks. Yes, a discussion would be great - I was going to ask you if the fence you mentioned was separate from this test issue. If you let me know which test (tests) can be related I could try running them a bunch of times. Usually fence failures are consistent but there have been cases which triggered rarely. I've seen a failure rate as low as 1 in 2000 on white. We've definitely had several spots which clearly needed a fence but just didn't ever fail the test.

MicheldeMessieres:
Cancelling this PR. These fences were not required and will be fixed via [https://github.com/kokkos/kokkos/pull/3244](url)

PR #7869, opened 2020-08-19
Tempus: just some minor bug fixes

sconde:
@trilinos/tempus 

## Motivation

Fix a minor bug/typo.

Also log the number of steps and step failure.

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorObserverBasic_impl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorObserverBasic_impl.hpp
@@ -109,6 +109,9 @@ observeEndIntegrator(const Integrator<Scalar>& integrator){
        << "  Total runtime = " << runtime << " sec = "
        << runtime/60.0 << " min\n"
        << std::asctime(std::localtime(&end))
+       << "\nNumber of Accepted Steps = " << integrator.getSolutionHistory()->getCurrentState()->getIndex()
+       << "\nNumber of Failures = " << integrator.getSolutionHistory()->getCurrentState()->getMetaData()->getNRunningFailures()
+       << "\n"
        << exitStatus << "\n"
        << std::endl;
 }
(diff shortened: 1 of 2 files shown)


packages/tempus/src/Tempus_IntegratorObserverBasic_impl.hpp
```diff
@@ -109,6 +109,9 @@ observeEndIntegrator(const Integrator<Scalar>& integrator){
        << "  Total runtime = " << runtime << " sec = "
        << runtime/60.0 << " min\n"
        << std::asctime(std::localtime(&end))
+       << "\niSteps = " << integrator.getSolutionHistory()->getCurrentState()->getIndex()
+       << "\nnFail = " << integrator.getSolutionHistory()->getCurrentState()->getMetaData()->getNRunningFailures()
+       << "\n"
```
ccober6, line 114: Can we be more descriptive?
Number of steps taken =
Number of failures        =

Also getIndex returns the state's index, and not the number of steps taken.  For restart, this will not be correct.

packages/tempus/src/Tempus_TimeStepControlStrategyPID.hpp
```diff
@@ -110,7 +110,7 @@ class TimeStepControlStrategyPID
      if (workingState->getSolutionStatus() == Status::PASSED or
          workingState->getSolutionStatus() == Status::WORKING) {
         if(lastStepRejected_){
-           dt = std::min(dt, workingState->getTimestep());
+           dt = std::min(dt, workingState->getTimeStep());
```
ccober6, line 113: Hmm... How did this even compile?

packages/tempus/src/Tempus_IntegratorObserverBasic_impl.hpp
sconde: > Also getIndex returns the state's index, and not the number of steps taken. For restart, this will not be correct.

Do we currently keep the track of the number of steps taken?

ccober6: Actually I do not think we do.  I would suggest a loop counter on the while loop in IntegratorBasic::advanceTime.  This would also be available for Subcycling.

sconde: I guess I should change it to `Number of Accepted Steps Taken` since that is what I'm really using the index for here. 

But with restart, we will have to make sure the starting index counter is correct.

sconde:
@ccober6 is this ready for review?

PR #7871, opened 2020-08-20
Panzer: Add fences for BasisValues2

MicheldeMessieres:
@rppawlo If we accept this PR, then I can proceed to get Panzer passing tests with CUDA_LAUNCH_BLOCKING=0. I'lll follow up with a PR which tags all fences in Panzer just for discussion purposes. Then we can return and fix these loops after discussion.

If we want to fix these loops now I'll need to discuss. This is clearly not a final solution. I expect we may do a single parallel_for kernel over the full range of all loops. Ifpack2 has some scatter methods with similar issues.

@trilinos/panzer 

## Motivation

Get Panzer passing for CUDA_LAUNCH_BLOCKING=0 prior to analyzing and resolving fences in panzer which will be a second phase of the project.

## Testing

Cuda white panzer DiscFE tests.

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_BasisValues2_impl.hpp
+++ b/packages/panzer/disc-fe/src/Panzer_BasisValues2_impl.hpp
@@ -555,7 +555,7 @@ evaluateValues_HGrad(const PHX::MDField<Scalar,Cell,IP,Dim> & cub_points,
 
     if(compute_derivatives){
         fst::HGRADtransformGRAD(cell_grad_basis.get_view(),cell_jac_inv.get_view(),cell_grad_basis_ref.get_view());
-        typename PHX::Device().fence();\
+        typename PHX::Device().fence();
         for(int b=0;b<num_basis;++b)
           for(int p=0;p<num_points;++p)
             for(int d=0;d<num_dim;++d)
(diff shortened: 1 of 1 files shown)


packages/panzer/disc-fe/src/Panzer_BasisValues2_impl.hpp
```diff
@@ -544,12 +548,14 @@ evaluateValues_HGrad(const PHX::MDField<Scalar,Cell,IP,Dim> & cub_points,
...
 
     fst::HGRADtransformVALUE(cell_basis_scalar.get_view(),cell_basis_ref_scalar.get_view());
+    typename PHX::Device().fence();
     for(int b=0;b<num_basis;++b)
       for(int p=0;p<num_points;++p)
         basis_scalar(cell,b,p)=cell_basis_scalar(0,b,p);
 
     if(compute_derivatives){
         fst::HGRADtransformGRAD(cell_grad_basis.get_view(),cell_jac_inv.get_view(),cell_grad_basis_ref.get_view());
+        typename PHX::Device().fence();\
```
rppawlo, line 558: Can you remove the extra backslash at the end of this line?

packages/panzer/disc-fe/src/Panzer_BasisValues2_impl.hpp
```diff
@@ -1183,7 +1216,7 @@ applyOrientations(const std::vector<Intrepid2::Orientation> & orientations,
   for (size_t i=0; i < drv_orts.size(); ++i)
     host_drv_orts(i) = orientations[i];
   Kokkos::deep_copy(drv_orts,host_drv_orts);
-  typename PHX::Device().fence();
+  typename PHX::Device().fence(); // TODO Remove this fence once DynRankView has all been fixed - pending soon
```
rppawlo, line 1219: Since the sacado change for DynRankView went in, can we remove this now?

MicheldeMessieres:
@rppawlo Thanks - I have fixed the extra slash and removed the fence.

PR #7896, opened 2020-08-24
Framework: PR Driver Script Change Detection

william76:
@trilinos/framework 

@prwolfe @jwillenbring Since this is messing with the top level .sh driver script, let's chat about this one before we merge it.  Manual testing looks ok but we'll want to merge this when we know we'll be around to monitor it in case we need to revert. 😉 

## Motivation
The current set of scripts that drive the PR testing have a known issue that can cause problems when either of these two scripts are modified:
1. `cmake/std/PullRequestLinuxDriver.sh` - Top level script that is called by Jenkins in a PR test.
2. `cmake/std/PullRequestLinuxDriverMerge.py` - Python script that handles merging the incoming branch into the target branch.

When either of these two scripts are modified, the incoming changes do not actually get tested by the PR system since the scripts themselves are loaded into memory prior to the incoming changes being merged in. This can cause the PR system to become broken when we need to update these scripts in a manner that requires a Framework team member to use special access to bypass the PR system to revert a change. Given that we really really really do not want to do that except in extraordinary circumstances (and even then, I treat this like using nuclear launch keys in the movies -- I will not do it without agreement from another Framework team member).

This modification adds a check step for scripts (1) and (2) to identify if either is changed during the _merge_ step. If so, the PR should re-launch the `PullRequestLinuxDriver.sh` script, which at that point should not have changes in the merge call.

### Prototype Bash Script
I developed a prototype bash script to test the general process and ran it on an ascic build server as well as on ride:
```bash
#!/usr/bin/env bash

# Get the full path to the current script
script_name=`basename $0`
script_path=$(dirname $(readlink -f $0))
script_file="${script_path}/${script_name:?}"
echo -e "script_file = ${script_file}"

# Get MD5 checksum of the existing file
sig_old=$(md5sum ${script_file} | cut -d' ' -f1)
echo -e "sig_old = ${sig_old}"


# DO THE MERGE HERE
echo -e ">>> Execute merge operation"
sleep 10

# Get MD5 checksum of the file post-merge
sig_new=$(md5sum ${script_file} | cut -d' ' -f1)
echo -e "sig_new = ${sig_new}"

# if file is different, re-run it and exit, otherwise continue.
if [ "${sig_old}" != "${sig_new}" ] || [ "foo" != "foo" ]
then
    echo -e "CHANGED!"
    ${script_file:?}
    exit $?
fi

echo -e ">>> Driver script unchanged."
echo -e ">>>"
echo -e ">>> Execute post-merge operations"
```

### Manual execution of PR jobs on Jenkins
Once this PR is created, I will test the following:
1. Manual run with a 'change' detected in the PullRequestDriver.sh script [Jenkins Console Output][2]. This shows a detected change in the md5sum for just the PullRequestDriver.sh script and the driver.sh script is re-launched.
2. Manual run with a 'change' detected in the PullRequestDriverMerge.py script added [Jenkins Console Output][3]. This one shows a change in the md5sum for both scripts.

[CDash Output should be here][1] 

[1]: https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=PR-7896
[2]: https://ascic-jenkins.sandia.gov/job/trilinos-folder/job/Trilinos_pullrequest_python_3/3245/console
[3]: https://ascic-jenkins.sandia.gov/job/trilinos-folder/job/Trilinos_pullrequest_python_3/3246/console

### DIFF
--- a/cmake/std/PullRequestLinuxDriver.sh
+++ b/cmake/std/PullRequestLinuxDriver.sh
@@ -143,5 +143,3 @@ echo -e "PRDriver> "
 ${test_cmd}
 exit $?
 
-
-
--- a/cmake/std/PullRequestLinuxDriver.sh
+++ b/cmake/std/PullRequestLinuxDriver.sh
@@ -143,3 +143,5 @@ echo -e "PRDriver> "
 ${test_cmd}
 exit $?
 
+
+
--- a/cmake/std/PullRequestLinuxDriver.sh
+++ b/cmake/std/PullRequestLinuxDriver.sh
@@ -1,7 +1,44 @@
 #!/usr/bin/env bash
 # set -x  # echo commands
 
-#    make  sure we have a newer version of git/python
+function get_scriptname() {
+    # Get the full path to the current script
+    local script_name=`basename $0`
+    local script_path=$(dirname $(readlink -f $0))
+    local script_file="${script_path}/${script_name:?}"
+    echo "${script_file}"
+}
+
+function get_scriptpath() {
+    # Get the full path to the current script
+    local script_name=`basename $0`
+    local script_path=$(dirname $(readlink -f $0))
+    echo "${script_path}"
+}
+
+# Get the md5sum of a filename.
+# param1: filename
+# returns: md5sum of the file.
+function get_md5sum() {
+    local filename=${1:?}
+    local sig=$(md5sum ${filename:?} | cut -d' ' -f1)
+    echo "${sig:?}"
+}
+
+echo -e "PRDRiver> ================================================="
+echo -e "PRDriver> ="
+echo -e "PRDriver> = PullRequestLinuxDriver.sh"
+echo -e "PRDriver> ="
+echo -e "PRDriver> ================================================="
+
+# Set up Sandia PROXY environment vars
+export https_proxy=http://wwwproxy.sandia.gov:80
+export http_proxy=http://wwwproxy.sandia.gov:80
+export no_proxy='localhost,localnets,127.0.0.1,169.254.0.0/16,forge.sandia.gov'
+
+
+# Load the right version of Git / Python based on a regex 
+# match to the Jenkins job name.
 cuda_regex=".*(_cuda_).*"
 ride_regex=".*(ride).*"
 if [[ ${JOB_BASE_NAME:?} =~ ${cuda_regex} ]]; then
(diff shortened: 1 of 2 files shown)


william76:
Pulled the WIP flag off to let the Autotester run. 

**Note**: Autotester results for this test will not be representative of the actual changes in this PR since I'm touching the `PullRequestLinuxDriver.sh` script which is cached before the Merge script is executed.

william76:
This update prepends `PRDriver> ` to the console output for print commands that are executed by `PullRequestLinuxDriver.sh` -- this makes it easier to scan through the console log in Jenkins to see output that's generated by the script itself.

Example Output from the manual test run:
```
00:01:20.488 PRDriver> Merge completed successfully.
00:01:20.488 PRDriver> 
00:01:20.496 PRDriver> Old md5 checksum a67d301c16fd06f0f44056458f048e26 for /scratch/trilinos/jenkins/tr-test-4/workspace/trilinos-folder/Trilinos_pullrequest_python_3/Trilinos/cmake/std/PullRequestLinuxDriver.sh
00:01:20.496 PRDriver> New md5 checksum d2d021cd50ed420a4d58ef2732c3bf1b for /scratch/trilinos/jenkins/tr-test-4/workspace/trilinos-folder/Trilinos_pullrequest_python_3/Trilinos/cmake/std/PullRequestLinuxDriver.sh
00:01:20.496 PRDriver> 
00:01:20.501 PRDriver> Old md5 checksum 1316630fa23a64eb5510997cc9963da7 for /scratch/trilinos/jenkins/tr-test-4/workspace/trilinos-folder/Trilinos_pullrequest_python_3/Trilinos/cmake/std/PullRequestLinuxDriverMerge.py
00:01:20.501 PRDriver> New md5 checksum 478ab629d41f3bb2bcc7e14a92666abe for /scratch/trilinos/jenkins/tr-test-4/workspace/trilinos-folder/Trilinos_pullrequest_python_3/Trilinos/cmake/std/PullRequestLinuxDriverMerge.py
00:01:20.501 PRDriver> 
00:01:20.501 PRDriver> Driver or Merge script change detected. Re-launching PR Driver
00:01:20.501 PRDriver> 
00:01:20.508 PRDRiver> =================================================
00:01:20.508 PRDriver> =
00:01:20.508 PRDriver> = PullRequestLinuxDriver.sh
00:01:20.508 PRDriver> =
00:01:20.508 PRDriver> =================================================
00:01:20.656 PRDriver> REPO_ROOT : /scratch/trilinos/jenkins/tr-test-4/workspace/trilinos-folder/Trilinos_pullrequest_python_3/Trilinos
00:01:20.668 PRDriver> 
00:01:20.668 PRDriver> Execute Merge Command: /scratch/trilinos/jenkins/tr-test-4/workspace/trilinos-folder/Trilinos_pullrequest_python_3/Trilinos/cmake/std/PullRequestLinuxDriverMerge.py https://github.com/william76/Trilinos PRDriverChangeDetection https://github.com/william76/Trilinos PRDriverChangeDetectionTest 18a03bfc6d4d0a6c83ce740526ef7993a2ccfa4f /scratch/trilinos/jenkins/tr-test-4/workspace/trilinos-folder/Trilinos_pullrequest_python_3
00:01:20.668 PRDriver> 
```

cmake/std/PullRequestLinuxDriver.sh
```diff
@@ -1,7 +1,44 @@
 #!/usr/bin/env bash
 # set -x  # echo commands
 
-#    make  sure we have a newer version of git/python
+function get_scriptname() {
+    # Get the full path to the current script
+    local script_name=`basename $0`
```
prwolfe, line 6: Perhaps just realpath $0 ?

cmake/std/PullRequestLinuxDriver.sh
```diff
@@ -19,40 +56,90 @@ else
...
-                                            ${BUILD_NUMBER:?} \
-                                            ${WORKSPACE:?}
+echo -e "PRDriver> "
+echo -e "PRDriver> Execute Merge Command: ${merge_cmd:?}" 
+echo -e "PRDriver> "
+${merge_cmd:?}
+err=$?
+if [ $err != 0 ]; then
+    echo -e "PRDriver> An error occurred during merge"
+    exit $err
```
prwolfe, line 95: Much nicer error message - thanks!

cmake/std/PullRequestLinuxDriverMerge.py
```diff
@@ -167,3 +167,5 @@ def run():
         exit(0)
     else:
         exit(1)
+
```
prwolfe, line 170: Extra whitespace

cmake/std/PullRequestLinuxDriver.sh
william76: I think I've run into compatibility issues on some platforms with `realpath` which is why I use `readlink -f` in other places. I can look into it, but if it's just cosmetic, I'd rather not make the change in this PR because the manual testing is a pain to run.

cmake/std/PullRequestLinuxDriverMerge.py
william76: Yes. I tend to prefer to add a couple extra empty lines to the bottom of scripts. It doesn't affect the results but gives me a small buffer at the end that's a little easier to see the end of code. Especially if I print out the source using `cat` that it gives a couple extra lines before the next prompt starts, etc.  Purely just a style thing that I tend to do.

william76:
@prwolfe 
> This looks nice, Did you get a run through the attester or just the one by hand?

Yes. The PR itself ran the attester but those are irrelevant since they'd use the version of these scripts on the target branch.
I linked to the console log in 2x Jenkins tests in the PR description above where I ran the Python3 PR job.  I kicked it off manually to use my branch. (it's actually trickier than just picking two SHA1's on the same branch, so I had to make a branch off my branch and use that as the 'target' and then add a dummy commit to my branch so there'd be a diff and use that as the source SHA1.

prwolfe:
> Yes. The PR itself ran the attester

First - DYAC
Second - I was hoping you set it up so that this branch was the target - sounds like you have covered the issue even if not through the autotester.

Thanks!

PR #7912, opened 2020-08-26
MueLu: Update RegionRFactory unit test to use coordinate maps

GrahamBenHarper:
@trilinos/muelu 

## Motivation
Previously the unit test assumed the coordinate map is the same as the row map. That is only true in the case of one DOF per node. This change is necessary to support multiple DOFs per node on the RegionRFactory unit tests.

## Testing
I reran all MueLu unit tests on Skybridge and they pass.


mayrmt (approved):
LGTM

PR #7914, opened 2020-08-26
Sacado: Fence review

MicheldeMessieres:
@etphipp For Tpetra we removed fences after kernels and placed them before the actual UVM reads. Removing these fences will still pass all unit tests with CUDA_LAUNCH_BLOCKING=0. These are the only core fences in Sacado. I am not sure if we want to remove them.

@trilinos/sacado 

## Motivation

Removing fences from core code.

## Testing

Sacado tests on white Cuda


MicheldeMessieres:
@etphipp I should have mentioned we do have some core fences for the Sacado DynRankView which I recently added to be consistent with the Kokkos version. However I don't expect we will be changing those anytime soon.

PR #7915, opened 2020-08-26
Phalanx: Add fence review

MicheldeMessieres:
@rppawlo This PR tags internal fences of phalanx. There is currently just one which is experimental. Closing the PR as it is currently just for reference.

@trilinos/phalanx

## Motivation

Remove core fences.

## Testing

Phalanx and Panzer tests on Cuda white build

### DIFF
--- a/packages/phalanx/src/Phalanx_CreateDeviceEvaluator.hpp
+++ b/packages/phalanx/src/Phalanx_CreateDeviceEvaluator.hpp
@@ -24,7 +24,6 @@ namespace PHX {
                          {
                            new (e) DeviceEvaluatorType(evaluator_to_clone_on_device);
                          });
-    ExecutionSpace().fence();
     return e;
   }
 


packages/phalanx/src/Phalanx_CreateDeviceEvaluator.hpp
```diff
@@ -24,7 +24,6 @@ namespace PHX {
                          {
                            new (e) DeviceEvaluatorType(evaluator_to_clone_on_device);
                          });
-    ExecutionSpace().fence();
```
rppawlo, line 27: @MicheldeMessieres - this is experimental code that would not be caught by standard testing. It needs some special configure flags to enable along with relocatable device code. Can we hold off on removing until I can do some more extensive testing?

MicheldeMessieres: @rppawlo Yes - sorry - I 'tested' for this to see if it was used, and then all tests passed and forgot what I was doing ... we can just treat this as a 'fence review' to tag internal fences. Fine with me if you want to just close this for now.

MicheldeMessieres:
@rppawlo I fixed the initial comments to be accurate.

rppawlo:
Thanks @MicheldeMessieres ! We can leave open or close - whatever you prefer.

PR #7916, opened 2020-08-26
Tempus: set Rel/Abs to currentstate for optimal initdt

sconde:
@trilinos/tempus 

## Motivation

Set tolerance to workingstate

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
@@ -413,6 +413,12 @@ void IntegratorBasic<Scalar>::startIntegrator()
     integratorStatus_ = Status::FAILED;
     return;
   }
+  
+  //set the Abs/Rel tolerance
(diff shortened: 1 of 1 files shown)


packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
```diff
@@ -413,6 +413,12 @@ void IntegratorBasic<Scalar>::startIntegrator()
     integratorStatus_ = Status::FAILED;
     return;
   }
+  
+  //set the Abs/Rel tolerance
+  auto ws = solutionHistory_->getCurrentState();
+  ws->setTolRel(timeStepControl_->getMaxRelError());
+  ws->setTolAbs(timeStepControl_->getMaxAbsError());
+
```
ccober6, line 421: Looks good.  You may want to name the variable cs instead of ws to avoid confusion.

sconde:
@ccober6 I made the change. Can I get a review again please

PR #7921, opened 2020-08-26
kokkos: Fix DynamicView reporting correct sizes and labels

ndellingwood:
Patch matching kokkos/kokkos#3315 by @crtrott
Reported by @bathmatt

@trilinos/kokkos


bathmatt:
I ran this and it fixed my bug. thanks!

PR #7924, opened 2020-08-27
Framework: Add a test commit for testing the PR driver update

william76:
@trilinos/framework 

Test PR (will not merge) with some changes for testing out the updated PR Driver script changes from #7896


william76:
@jwillenbring @prwolfe looks like this passed so the update to the driver script seems ok.  I'll go ahead and close this PR.

PR #7926, opened 2020-08-27
MueLu: Root node filtering

csiefer2:
This is mostly to see which tests break.  No serious reviews yet please.

TESTS: RHEL7

### DIFF
--- a/packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
+++ b/packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
@@ -195,7 +195,7 @@ namespace MueLu {
       ////////////////////////////////////////////////////
       // Remove this bit once we are confident that cut-based dropping works.
 #ifdef HAVE_MUELU_DEBUG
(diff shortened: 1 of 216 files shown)


lucbv (dismissed):
Yep this is fine

packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
```diff
@@ -195,7 +195,7 @@ namespace MueLu {
       ////////////////////////////////////////////////////
       // Remove this bit once we are confident that cut-based dropping works.
 #ifdef HAVE_MUELU_DEBUG
-      int distanceLaplacianCutVerbose = 0;
+   #   int distanceLaplacianCutVerbose = 0;
```
lucbv, line 198: Does this line of code compile? It looks like a `#` symbol instead of `//`

csiefer2: Probably doesn't compile.   Fixed.

jhux2 (approved):
seems ok

PR #7946, opened 2020-08-31
Tpetra: Convert deep_copy to 3-arg and add temporary comments.

MicheldeMessieres:
Note: This PR passed on white's Pascal node with CUDA_LAUNCH_BLOCKING off. However it needs further review. Specifically any deep_copy which potentially has HostSpace on one side and CudaSpace or CudaUVMSpace on the other side can be a failure point since nothing will be protecting access to the host data after the async deep_copy.

Since the tests are all passing it's likely that many of these are fine but the logic of each needs to be checked. 

The PR also adds some comments to identify the kind of deep_copy:
```
DEVICE - means CudaSpace, CudaUVMSpace, or HostSpace - depends on the setup
HOSTMIRROR - means CudaUVMSpace or HostSpace - depending on what was mirrored.
HOST - always means HostSpace
```

@trilinos/tpetra

### DIFF
--- a/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
@@ -681,6 +681,8 @@ public:
     localError_ (new bool (false)),
     errs_ (new Teuchos::RCP<std::ostringstream> ()) // ptr to a null ptr
   {
+    using execution_space = typename Node::execution_space;
+
     TEUCHOS_TEST_FOR_EXCEPTION(
       ! graph_.isSorted (), std::invalid_argument, "Tpetra::"
       "BlockCrsMatrix constructor: The input CrsGraph does not have sorted "
@@ -715,7 +717,9 @@ public:
 
       row_map_type ptr_d = graph.getLocalGraph ().row_map;
       nc_host_row_map_type ptr_h_nc = Kokkos::create_mirror_view (ptr_d);
-      Kokkos::deep_copy (ptr_h_nc, ptr_d);
+
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy (execution_space(), ptr_h_nc, ptr_d);
       ptrHost_ = ptr_h_nc;
     }
     {
@@ -724,7 +728,8 @@ public:
 
       entries_type ind_d = graph.getLocalGraph ().entries;
       nc_host_entries_type ind_h_nc = Kokkos::create_mirror_view (ind_d);
-      Kokkos::deep_copy (ind_h_nc, ind_d);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy (execution_space(), ind_h_nc, ind_d);
       indHost_ = ind_h_nc;
     }
 
@@ -751,6 +756,8 @@ public:
     localError_ (new bool (false)),
     errs_ (new Teuchos::RCP<std::ostringstream> ()) // ptr to a null ptr
   {
+    using execution_space = typename Node::execution_space;
+
     TEUCHOS_TEST_FOR_EXCEPTION(
       ! graph_.isSorted (), std::invalid_argument, "Tpetra::"
       "BlockCrsMatrix constructor: The input CrsGraph does not have sorted "
@@ -781,7 +788,8 @@ public:
 
       row_map_type ptr_d = graph.getLocalGraph ().row_map;
       nc_host_row_map_type ptr_h_nc = Kokkos::create_mirror_view (ptr_d);
-      Kokkos::deep_copy (ptr_h_nc, ptr_d);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy (execution_space(), ptr_h_nc, ptr_d);
       ptrHost_ = ptr_h_nc;
     }
     {
@@ -790,7 +798,8 @@ public:
 
       entries_type ind_d = graph.getLocalGraph ().entries;
       nc_host_entries_type ind_h_nc = Kokkos::create_mirror_view (ind_d);
-      Kokkos::deep_copy (ind_h_nc, ind_d);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy (execution_space(), ind_h_nc, ind_d);
       indHost_ = ind_h_nc;
     }
 
@@ -968,6 +977,7 @@ public:
          prefix << "The matrix's values need sync on both device and host.");
 #endif // HAVE_TPETRA_DEBUG
       this->modify_host ();
+      // DEEP_COPY REVIEW - NOT TESTED
       Kokkos::deep_copy (getValuesHost (), alpha);
     }
     else {
@@ -976,7 +986,9 @@ public:
       // device.  Also, prefer modifying on device if neither side is
       // marked as modified.
       this->modify_device ();
-      Kokkos::deep_copy (this->getValuesDevice (), alpha);
+      // DEEP_COPY REVIEW - VALUE-TO-DEVICE
+      using execution_space = typename device_type::execution_space;
+      Kokkos::deep_copy (execution_space(), this->getValuesDevice (), alpha);
     }
   }
 
@@ -3755,7 +3767,9 @@ public:
 
     // The code below works on host, so use a host View.
     auto diagOffsetsHost = Kokkos::create_mirror_view (diagOffsets);
-    Kokkos::deep_copy (diagOffsetsHost, diagOffsets);
+    // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+    using execution_space = typename device_type::execution_space;
+    Kokkos::deep_copy (execution_space(), diagOffsetsHost, diagOffsets);
     // We're filling diag on host for now.
     diag.template modify<typename decltype (diagOffsetsHost)::memory_space> ();
 
--- a/packages/tpetra/core/src/Tpetra_BlockMultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_BlockMultiVector_def.hpp
@@ -305,7 +305,9 @@ replaceLocalValuesImpl (const LO localRowIndex,
   auto X_dst = getLocalBlock (localRowIndex, colIndex);
   typename const_little_vec_type::HostMirror::const_type X_src (reinterpret_cast<const impl_scalar_type*> (vals),
                                                                 getBlockSize ());
-  Kokkos::deep_copy (X_dst, X_src);
+  // DEEP_COPY REVIEW - HOSTMIRROR-TO-DEVICE
+  using execution_space = typename device_type::execution_space;
+  Kokkos::deep_copy (execution_space(), X_dst, X_src);
 }
 
 
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -382,7 +382,9 @@ namespace Tpetra {
     in_view_type numAllocPerRowIn (numEntPerRow.getRawPtr (), lclNumRows);
     nc_view_type numAllocPerRowOut ("Tpetra::CrsGraph::numAllocPerRow",
                                     lclNumRows);
-    Kokkos::deep_copy (numAllocPerRowOut, numAllocPerRowIn);
+    // DEEP_COPY REVIEW - HOST-TO-HOSTMIRROR
+    using execution_space = typename nc_view_type::execution_space;
+    Kokkos::deep_copy (execution_space(), numAllocPerRowOut, numAllocPerRowIn);
     k_numAllocPerRow_ = numAllocPerRowOut;
 
     resumeFill (params);
@@ -518,7 +520,9 @@ namespace Tpetra {
     in_view_type numAllocPerRowIn (numEntPerRow.getRawPtr (), lclNumRows);
     nc_view_type numAllocPerRowOut ("Tpetra::CrsGraph::numAllocPerRow",
                                     lclNumRows);
-    Kokkos::deep_copy (numAllocPerRowOut, numAllocPerRowIn);
+    // DEEP_COPY REVIEW - HOST-TO-HOSTMIRROR
+    using execution_space = typename nc_view_type::execution_space;
+    Kokkos::deep_copy (execution_space(), numAllocPerRowOut, numAllocPerRowIn);
     k_numAllocPerRow_ = numAllocPerRowOut;
 
     resumeFill (params);
@@ -1249,7 +1253,10 @@ namespace Tpetra {
         std::cerr << os.str();
       }
       row_ent_type numRowEnt (ViewAllocateWithoutInitializing (label), numRows);
-      Kokkos::deep_copy (numRowEnt, static_cast<size_t> (0)); // fill w/ 0s
+      // DEEP_COPY REVIEW - VALUE-TO-HOSTMIRROR
+      using execution_space = typename device_type::execution_space;
+      Kokkos::deep_copy (execution_space(), numRowEnt, static_cast<size_t> (0)); // fill w/ 0s
+      Kokkos::fence(); // TODO: Need to understand downstream failure points and move this fence.
       this->k_numRowEntries_ = numRowEnt; // "commit" our allocation
     }
 
@@ -2380,7 +2387,8 @@ namespace Tpetra {
       // of create_mirror_view might actually be a new allocation.
       // This helps with debugging when there are two memory spaces.
       typename row_map_type::HostMirror ptr_h = create_mirror_view (k_rowPtrs_);
-      Kokkos::deep_copy (ptr_h, k_rowPtrs_);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy (execution_space(), ptr_h, k_rowPtrs_);
       if (debug_) {
         TEUCHOS_TEST_FOR_EXCEPTION
           (ptr_h.extent (0) != k_rowPtrs_.extent (0), std::logic_error,
@@ -2404,6 +2412,7 @@ namespace Tpetra {
       ret_view_type ptr_d (ViewAllocateWithoutInitializing ("ptr"), size);
       ::Tpetra::Details::copyOffsets (ptr_d, k_rowPtrs_);
       typename ret_view_type::HostMirror ptr_h = create_mirror_view (ptr_d);
+      // DEEP_COPY REVIEW - NOT TESTED
       Kokkos::deep_copy (ptr_h, ptr_d);
       ptr_st = Kokkos::Compat::persistingView (ptr_h);
     }
@@ -2420,6 +2429,7 @@ namespace Tpetra {
 
     // If size_t == row_offset_type, return a persisting host view of
     // k_rowPtrs_.  Otherwise, return a size_t host copy of k_rowPtrs_.
+    Kokkos::fence(); // Make sure unfenced 3-arg deep_copy is done before using result of persistingView
     ArrayRCP<const size_t> retval =
       Kokkos::Impl::if_c<same,
         ArrayRCP<const row_offset_type>,
@@ -3011,7 +3021,10 @@ namespace Tpetra {
       // an assignment of View<row_offset_type*, ...> to View<size_t*,
       // ...> unless size_t == row_offset_type.
       input_view_type ptr_decoy (rowPointers.getRawPtr (), size); // never used
-      Kokkos::deep_copy (Kokkos::Impl::if_c<same,
+      // DEEP_COPY REVIEW - HOST-TO-DEVICE
+      using execution_space = typename device_type::execution_space;
+      Kokkos::deep_copy (execution_space(),
+                         Kokkos::Impl::if_c<same,
                            nc_row_map_type,
                            input_view_type>::select (ptr_rot, ptr_decoy),
                          ptr_in);
@@ -3032,6 +3045,7 @@ namespace Tpetra {
         // execution space would avoid the double copy.
         //
         View<size_t*, layout_type ,execution_space > ptr_st ("Tpetra::CrsGraph::ptr", size);
+        // DEEP_COPY REVIEW - NOT TESTED
         Kokkos::deep_copy (ptr_st, ptr_in);
         // Copy on device (casting from size_t to row_offset_type,
         // with bounds checking if necessary) to ptr_rot.  This
@@ -5135,7 +5149,9 @@ namespace Tpetra {
     row_ptrs_type row_ptrs_beg(
       view_alloc("row_ptrs_beg", WithoutInitializing),
       k_rowPtrs_.extent(0));
-    Kokkos::deep_copy(row_ptrs_beg, k_rowPtrs_);
+
+    // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
+    Kokkos::deep_copy(execution_space(), row_ptrs_beg, k_rowPtrs_);
 
     const size_t N = row_ptrs_beg.extent(0) == 0 ? size_t(0) :
       size_t(row_ptrs_beg.extent(0) - 1);
@@ -5148,6 +5164,9 @@ namespace Tpetra {
       view_alloc("row_ptrs_end", WithoutInitializing), N);
 
     const bool refill_num_row_entries = k_numRowEntries_.extent(0) != 0;
+
+    execution_space().fence(); // we need above deep_copy to be done
+
     if (refill_num_row_entries) { // Case 1: Unpacked storage
       // We can't assume correct *this capture until C++17, and it's
       // likely more efficient just to capture what we need anyway.
@@ -5701,6 +5720,7 @@ namespace Tpetra {
         Kokkos::MemoryUnmanaged> exports_a_h (exports_a.getRawPtr (), newSize);
       exports.clear_sync_state ();
       exports.modify_host ();
+      // DEEP_COPY REVIEW - NOT TESTED
       Kokkos::deep_copy (exports.view_host (), exports_a_h);
     }
     // packCrsGraphNew requires a valid localGraph.
@@ -6609,7 +6629,7 @@ namespace Tpetra {
           }
         } // whether lclColInd is a valid local column index
       } // for each local row
-
+      // DEEP_COPY REVIEW - NOT TESTED
       Kokkos::deep_copy (offsets, offsets_h);
     } // whether the graph is fill complete
 
@@ -6746,6 +6766,7 @@ namespace Tpetra {
       copyBackIfNeeded (const host_offsets_type& hostOffsets,
                         const device_offsets_type& deviceOffsets)
       {
+        // DEEP_COPY REVIEW - NOT TESTED
         Kokkos::deep_copy (hostOffsets, deviceOffsets);
       }
     };
@@ -7633,8 +7654,12 @@ namespace Tpetra {
     {
       typename local_graph_type::row_map_type::const_type::HostMirror k_rowPtrs_host_this = Kokkos::create_mirror_view(this->k_rowPtrs_);
       typename local_graph_type::row_map_type::const_type::HostMirror k_rowPtrs_host_graph= Kokkos::create_mirror_view(graph.k_rowPtrs_);
-      Kokkos::deep_copy(k_rowPtrs_host_this, this->k_rowPtrs_);
-      Kokkos::deep_copy(k_rowPtrs_host_graph, graph.k_rowPtrs_);
+
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy(execution_space(), k_rowPtrs_host_this, this->k_rowPtrs_);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy(execution_space(), k_rowPtrs_host_graph, graph.k_rowPtrs_);
+      Kokkos::fence(); // UVM below
       for(size_t i=0; output && i<k_rowPtrs_host_this.extent(0); i++)
         output = k_rowPtrs_host_this(i) == k_rowPtrs_host_graph(i) ? output : false;
     }
@@ -7645,8 +7670,11 @@ namespace Tpetra {
     {
       typename local_graph_type::entries_type::non_const_type::HostMirror k_lclInds1D_host_this = Kokkos::create_mirror_view(this->k_lclInds1D_);
       typename local_graph_type::entries_type::non_const_type::HostMirror k_lclInds1D_host_graph= Kokkos::create_mirror_view(graph.k_lclInds1D_);
-      Kokkos::deep_copy(k_lclInds1D_host_this, this->k_lclInds1D_);
-      Kokkos::deep_copy(k_lclInds1D_host_graph, graph.k_lclInds1D_);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy(execution_space(), k_lclInds1D_host_this, this->k_lclInds1D_);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy(execution_space(), k_lclInds1D_host_graph, graph.k_lclInds1D_);
+      Kokkos::fence(); // UVM below
       for(size_t i=0; output && i < k_lclInds1D_host_this.extent(0); i++)
         output = k_lclInds1D_host_this(i) == k_lclInds1D_host_graph(i) ? output : false;
     }
@@ -7657,8 +7685,11 @@ namespace Tpetra {
     {
       typename t_GlobalOrdinal_1D::HostMirror k_gblInds1D_host_this  = Kokkos::create_mirror_view(this->k_gblInds1D_);
       typename t_GlobalOrdinal_1D::HostMirror k_gblInds1D_host_graph = Kokkos::create_mirror_view(graph.k_gblInds1D_);
+      // DEEP_COPY REVIEW - NOT TESTED
       Kokkos::deep_copy(k_gblInds1D_host_this, this->k_gblInds1D_);
+      // DEEP_COPY REVIEW - NOT TESTED
       Kokkos::deep_copy(k_gblInds1D_host_graph, graph.k_gblInds1D_);
+      Kokkos::fence(); // UVM below
       for(size_t i=0; output && i<k_gblInds1D_host_this.extent(0); i++)
         output = k_gblInds1D_host_this(i) == k_gblInds1D_host_graph(i) ? output : false;
     }
@@ -7675,8 +7706,11 @@ namespace Tpetra {
     {
       typename local_graph_type::row_map_type::HostMirror lclGraph_rowmap_host_this  = Kokkos::create_mirror_view(this->lclGraph_.row_map);
       typename local_graph_type::row_map_type::HostMirror lclGraph_rowmap_host_graph = Kokkos::create_mirror_view(graph.lclGraph_.row_map);
-      Kokkos::deep_copy(lclGraph_rowmap_host_this, this->lclGraph_.row_map);
-      Kokkos::deep_copy(lclGraph_rowmap_host_graph, graph.lclGraph_.row_map);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy(execution_space(), lclGraph_rowmap_host_this, this->lclGraph_.row_map);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy(execution_space(), lclGraph_rowmap_host_graph, graph.lclGraph_.row_map);
+      Kokkos::fence(); // UVM below
       for(size_t i=0; output && i<lclGraph_rowmap_host_this.extent(0); i++)
         output = lclGraph_rowmap_host_this(i) == lclGraph_rowmap_host_graph(i) ? output : false;
     }
@@ -7686,8 +7720,11 @@ namespace Tpetra {
     {
       typename local_graph_type::entries_type::HostMirror lclGraph_entries_host_this = Kokkos::create_mirror_view(this->lclGraph_.entries);
       typename local_graph_type::entries_type::HostMirror lclGraph_entries_host_graph = Kokkos::create_mirror_view(graph.lclGraph_.entries);
-      Kokkos::deep_copy(lclGraph_entries_host_this, this->lclGraph_.entries);
-      Kokkos::deep_copy(lclGraph_entries_host_graph, graph.lclGraph_.entries);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy(execution_space(), lclGraph_entries_host_this, this->lclGraph_.entries);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
+      Kokkos::deep_copy(execution_space(), lclGraph_entries_host_graph, graph.lclGraph_.entries);
+      Kokkos::fence();  // UVM below
       for(size_t i=0; output && i<lclGraph_entries_host_this.extent(0); i++)
         output = lclGraph_entries_host_this(i) == lclGraph_entries_host_graph(i) ? output : false;
     }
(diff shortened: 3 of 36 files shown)


packages/tpetra/core/src/Tpetra_Details_EquilibrationInfo.hpp
```diff
@@ -138,21 +138,27 @@ struct EquilibrationInfo {
   void
   assign (const EquilibrationInfo<ScalarType, SrcDeviceType>& src)
   {
+  std::terminate();
```
csiefer2, line 141: Don't think we want to keep the std::terminate

packages/tpetra/core/src/Tpetra_Details_EquilibrationInfo.hpp
```diff
@@ -138,21 +138,27 @@ struct EquilibrationInfo {
...
+    // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
     Kokkos::deep_copy (rowNorms, src.rowNorms);
```
csiefer2, line 143: These guys are probably device-to-device in the tests, but the code is written to allow the device_type and the SrcDeviceType to be different.

Ugh.

Admittedly, we might not *need* this functionality, so if that's true we could rip it the SrcDeviceType stuff.

I vote for deferring a decision on this guy until we get feedback from the Kokkos team on Geoff's test.
I want to know what 3-arg is supposed to do if the views are in different memory spaces.  If it doesn't cost us much to keep the functionality, we can keep it.  Otherwise we can yank it.

packages/tpetra/core/src/Tpetra_Details_createMirrorView.hpp
```diff
@@ -148,6 +148,7 @@ class CreateMirrorViewFromUnmanagedHostArray<ValueType, OutputDeviceType, true,
     else {
       // No need to initialize, if we're going to copy into it anyway.
       outView_nc = nc_output_view_type (view_alloc (std::string (label), WithoutInitializing), inSize);
+      // DEEP_COPY REVIEW - HOST-TO-DEVICE
```
csiefer2, line 151: The createMirrorView file is a way to create host views of (potentially) device data or *not* do it depending on the underlying architecture (which is handles via template partial specialization).  This stuff gets used in tests frequently, but also is littered all over the core classes.

Not sure it is the cleanest way to do this, but this is what we've got.

csiefer2: Given the way there are two sets of device types InputDeviceType and OutputDevice type, I'm not 100% sure the device to host annotations are correct here.

packages/tpetra/core/src/Tpetra_Details_fill.hpp
```diff
@@ -83,6 +83,7 @@ class Fill {
   {
     static_assert (std::is_integral<IndexType>::value,
                    "IndexType must be a built-in integer type.");
+    // DEEP_COPY REVIEW - NOT TESTED
```
csiefer2, line 86: As far as I can tell, this code is only used by MultiVector for implementing putScalar.

The only value this guy adds, I think is filling only certain columns of a MultiVector which is bigger.  I'm not sure KokkosBlas can do that (@brian-kelley Can it?).

This should all be scalar-to-device copies.


I'm sorely tempted to recommend rewriting this file (and/or dumping the functionality into MultiVector).

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
```diff
@@ -627,12 +627,16 @@ makeColMap (Teuchos::RCP<const Tpetra::Map<LO, GO, NT>>& colMap,
   //Pull down the sizes
   GO numLocalColGIDs = 0;
   GO numRemoteColGIDs = 0;
+  // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
csiefer2, line 630: These are deep_copy's of device views of scalars to host.

I'm at a loss for why we do this rather than use the parallel scan's return value capability.

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
```diff
@@ -627,12 +627,16 @@ makeColMap (Teuchos::RCP<const Tpetra::Map<LO, GO, NT>>& colMap,
...
   Kokkos::deep_copy(numLocalColGIDs, numLocals);
+  // DEEP_COPY REVIEW - DEVICE-TO-HOST
   Kokkos::deep_copy(numRemoteColGIDs, numRemotes);
   //Pull down the remote lists
   auto localsHost = Kokkos::create_mirror_view(localGIDView);
   auto remotesHost = Kokkos::create_mirror_view(remoteGIDView);
+  // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
csiefer2, line 637: Yup.  This is a real, device to host copy of an array.  This goop could be replaced by a create_mirror_view_and_copy.  Though I we did that, we'd need to add a sync semantic check for it to Geoff's tests.

packages/tpetra/core/src/Tpetra_computeRowAndColumnOneNorms_def.hpp
```diff
@@ -127,6 +127,8 @@ computeLocalRowScaledColumnNorms_RowMatrix (EquilibrationInfo<typename Kokkos::A
   using mag_type = typename KAT::mag_type;
 
   auto rowNorms_h = Kokkos::create_mirror_view (result.rowNorms);
+
+  // DEEP_COPY REVIEW - NOT TESTED
```
csiefer2, line 131: This is a working, but poor performing implementation which uses applies a function pointer to each row of the matrix via a getRow.

If we cared about the performance of the thing we'd just rewrite this all in the Kokkos-verse.

But as it is, those are device-to-host / host-to-device deep copy calls (in that order).

csiefer2 (commented):
Comments added

packages/tpetra/core/src/Tpetra_Details_EquilibrationInfo.hpp
MicheldeMessieres: fixed - BTW I won't plan to use this PR for any submission, just for discussion. I'll update the PR as I merge other PR's so the DEEP_COPY REVIEW comments will only exist for unprocessed deep_copy calls.

packages/tpetra/core/src/Tpetra_Details_createMirrorView.hpp
MicheldeMessieres: I've been commenting things as device-to-device if they could be device-to-device but not necessarily always. For host-to-device to device-to-host I tried to do that only when host was specific but sometimes it was hard to tell. I think we're currently planning to make all these cases run same device execution space so it may not be critical I got the annotations always correct. I am using them as a way to break down the project into smaller PRs which may help when issues come up.

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -715,6 +715,7 @@ class GetLocalDiagCopy {
 
       row_map_type ptr_d = graph.getLocalGraph ().row_map;
       nc_host_row_map_type ptr_h_nc = Kokkos::create_mirror_view (ptr_d);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kyungjoo-kim, line 718: Fine. KK

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -724,6 +725,7 @@ class GetLocalDiagCopy {
 
       entries_type ind_d = graph.getLocalGraph ().entries;
       nc_host_entries_type ind_h_nc = Kokkos::create_mirror_view (ind_d);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kyungjoo-kim, line 728: Fine. KK

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -781,6 +783,7 @@ class GetLocalDiagCopy {
 
       row_map_type ptr_d = graph.getLocalGraph ().row_map;
       nc_host_row_map_type ptr_h_nc = Kokkos::create_mirror_view (ptr_d);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kyungjoo-kim, line 786: Fine. KK

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -790,6 +793,7 @@ class GetLocalDiagCopy {
 
       entries_type ind_d = graph.getLocalGraph ().entries;
       nc_host_entries_type ind_h_nc = Kokkos::create_mirror_view (ind_d);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kyungjoo-kim, line 796: Fine. KK

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -968,6 +972,7 @@ class GetLocalDiagCopy {
          prefix << "The matrix's values need sync on both device and host.");
 #endif // HAVE_TPETRA_DEBUG
       this->modify_host ();
+      // DEEP_COPY REVIEW - NOT TESTED
```
kyungjoo-kim, line 975: The method is setAllToScalar. The majority of use cases in my opinion is to set the matrix zero. Otherwise, setting all block crs matrix with a scalar value does not make sense to me. Then, why do we need to write this complicated routine ? We just call deep copy of scalar value for both host and device and set the dual view is synced. This is probably easier.

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -976,6 +981,7 @@ class GetLocalDiagCopy {
       // device.  Also, prefer modifying on device if neither side is
       // marked as modified.
       this->modify_device ();
+      // DEEP_COPY REVIEW - VALUE-TO-DEVICE
```
kyungjoo-kim, line 984: See the above comment.

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -3754,6 +3760,7 @@ class GetLocalDiagCopy {
 
     // The code below works on host, so use a host View.
     auto diagOffsetsHost = Kokkos::create_mirror_view (diagOffsets);
+    // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kyungjoo-kim, line 3763: This method is getLocalDiagCopy. I think that this code can be trivially done on device. For this moment, the code requires the device-to-host deep copy because it uses host method of the diag vector. The code does not seem to be efficient and it is serial. What I mean is that by extracting diagonals on device, this deep copy is not really necessary. Also the code check host_sync only with the debug mode. With a dual view context, why is this users' responsibility ? It might also be necessary to discuss why we need to follow the last touch rule on a gpu system.

packages/tpetra/core/src/Tpetra_BlockMultiVector_def.hpp
```diff
@@ -305,6 +305,7 @@ replaceLocalValuesImpl (const LO localRowIndex,
   auto X_dst = getLocalBlock (localRowIndex, colIndex);
   typename const_little_vec_type::HostMirror::const_type X_src (reinterpret_cast<const impl_scalar_type*> (vals),
                                                                 getBlockSize ());
+  // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
```
kyungjoo-kim, line 308: This is not device to device. It is host to device. In the above code, it uses HostMirror. 

This interface is not going to be efficient either. It provides an access to a single block in a block multi vector. If the block multi vector is defined on a gpu, then this causes frequent memory transfer for a single block. Since this interface defines on host, we should collect all modifications of small blocks. After all modifications are done, we should sync host view to device view. The syntax is probably....
```
block_multivector.beginModificationOnHost();
for (int i=0;i<many;++i) {
  block_multivector.replaceLocalValues(rowIdx, colIdx, valsFromHost);
}
block_multivector.endModificationOnHost();
```

packages/tpetra/core/src/Tpetra_Details_copyConvert.hpp
```diff
@@ -279,10 +279,13 @@ namespace { // (anonymous)
         // it's cheaper to allocate.  Hopefully users aren't doing
         // aliased copies in a tight loop.
         auto src_copy = Kokkos::create_mirror (Kokkos::HostSpace (), src);
+        // DEEP_COPY REVIEW - NOT TESTED
```
kyungjoo-kim, line 282: In my opinion, this logic is probably wrong. The intention of this copy code is a safe utility function to avoid possible pointer aliasing. However, this logic does not solve the issue anyway. First, the aliasing only consider that the view is contiguous. If two views are subviewed and strided. they can be copied without a trouble. Second, this logic creates a mirror view to host for temporal memory. This probably work on a gpu but it is more expensive. It can just use device memory instead. Also, if the default space is host, then this logic is wrong as the mirror view construction is just a softcopy and src_copy is still overlapping with dst. I do not recommend to use this function if it is used somewhere.

packages/tpetra/core/src/Tpetra_Details_copyConvert.hpp
```diff
@@ -279,10 +279,13 @@ namespace { // (anonymous)
...
         Kokkos::deep_copy (src_copy, src);
+        // DEEP_COPY REVIEW - NOT TESTED
         Kokkos::deep_copy (dst, src_copy);
       }
       else { // no aliasing
+        // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
```
kyungjoo-kim, line 288: This deep copy itself is fine but I do not recommend to use this copyConvert

packages/tpetra/core/src/Tpetra_Details_copyConvert.hpp
```diff
@@ -328,6 +331,7 @@ namespace { // (anonymous)
       using output_memory_space = typename OutputViewType::memory_space;
       auto src_outputSpaceCopy =
         Kokkos::create_mirror_view (output_memory_space (), src);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kyungjoo-kim, line 334: In my opinion, this copyConvert can be directly replaced by kokkkos deep_copy.

packages/tpetra/core/src/Tpetra_createDeepCopy_CrsMatrix_def.hpp
```diff
@@ -28,17 +28,23 @@ localDeepCopyFillCompleteCrsMatrix (const CrsMatrix<SC, LO, GO, NT>& A)
   using inds_type = typename local_graph_type::entries_type;
   inds_type ind (view_alloc ("ind", WithoutInitializing),
                  A_lcl.graph.entries.extent (0));
+
+  // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
```
kyungjoo-kim, line 32: Fine. See the last comment in this file. KK

packages/tpetra/core/src/Tpetra_createDeepCopy_CrsMatrix_def.hpp
```diff
@@ -28,17 +28,23 @@ localDeepCopyFillCompleteCrsMatrix (const CrsMatrix<SC, LO, GO, NT>& A)
...
   Kokkos::deep_copy (ind, A_lcl.graph.entries);
 
   using offsets_type =
     typename local_graph_type::row_map_type::non_const_type;
   offsets_type ptr (view_alloc ("ptr", WithoutInitializing),
                     A_lcl.graph.row_map.extent (0));
+
+  // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
```
kyungjoo-kim, line 40: Fine. See the last comment in this file. KK

packages/tpetra/core/src/Tpetra_createDeepCopy_CrsMatrix_def.hpp
```diff
@@ -28,17 +28,23 @@ localDeepCopyFillCompleteCrsMatrix (const CrsMatrix<SC, LO, GO, NT>& A)
...
   Kokkos::deep_copy (ptr, A_lcl.graph.row_map);
 
   using values_type = typename local_matrix_type::values_type;
   values_type val (view_alloc ("val", WithoutInitializing),
                    A_lcl.values.extent (0));
+
+  // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
```
kyungjoo-kim, line 47: Instead of allocating - deep copy three times, we may want to allocate 3 views first and call 3 deep copies. In this way, probably the code does not have to use a system call and it can run without blocking from the system call.

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
```diff
@@ -147,6 +147,7 @@ lclNormImpl (const RV& normsOut,
 
   if (lclNumRows == 0) {
     const mag_type zeroMag = Kokkos::ArithTraits<mag_type>::zero ();
+    // DEEP_COPY REVIEW - VALUE-TO-DEVICE
```
GeoffDanielson, line 150: Untested, used by tpetra_multivector.  The obvious answer would to be to use multiVector to establish its own norm Impl then the identity (kokkos::arithtraits<T>::zero()) wouldn't be necessary.

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
```diff
@@ -252,6 +253,7 @@ gblNormImpl (const RV& normsOut,
     // MPI doesn't allow aliasing of arguments, so we have to make
     // a copy of the local sum.
     RV lclNorms ("MV::normImpl lcl", numVecs);
+    // DEEP_COPY REVIEW - HOST-TO-DEVICE
```
GeoffDanielson, line 256: Also untested anywhere, and similarly only used in multivector.

Not sure whether this was intended to be used in a wider context than this or not.

packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
```diff
@@ -1635,6 +1636,7 @@ namespace Tpetra {
       // std::complex conversions, but those two implementations
       // should generally be bitwise compatible.
       // CT: no this can't possible work .....
+      // DEEP_COPY REVIEW - NOT TESTED
```
tasmith4, line 1639: device to device, not tested, surrounding comments suggest we should remove

packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
```diff
@@ -1764,6 +1767,7 @@ namespace Tpetra {
       // Sacado and Stokhos packages are likely to care about this use
       // case.  It could also come up with Kokkos::complex ->
       // std::complex conversion.
+      // DEEP_COPY REVIEW - NOT TESTED
```
tasmith4, line 1770: device to device, not tested, surrounding comments suggest we should remove

packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
```diff
@@ -1854,6 +1859,7 @@ namespace Tpetra {
       // Sacado and Stokhos packages are likely to care about this use
       // case.  This could also come up with Kokkos::complex ->
       // std::complex conversion.
+      // DEEP_COPY REVIEW - NOT TESTED
```
tasmith4, line 1862: device to device, not tested, surrounding comments suggest we should remove

packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
```diff
@@ -1937,6 +1944,7 @@ namespace Tpetra {
...
```
tasmith4, line 1947: device to device, not tested, surrounding comments suggest we should remove

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -1065,6 +1068,7 @@ namespace Tpetra {
 
           auto tgt_j = Kokkos::subview (tgt_h, rows, tgtCol);
           auto src_j = Kokkos::subview (src_h, rows, srcCol);
+          // DEEP_COPY REVIEW - HOST-TO-HOST
```
tasmith4, line 1071: this is tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -1079,6 +1083,7 @@ namespace Tpetra {
 
           auto tgt_j = Kokkos::subview (tgt_d, rows, tgtCol);
           auto src_j = Kokkos::subview (src_d, rows, srcCol);
+          // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
```
tasmith4, line 1086: this is tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -1634,10 +1639,12 @@ namespace Tpetra {
       whichVecs = Kokkos::DualView<size_t*, device_type> ("whichVecs", numVecs);
       if (unpackOnHost) {
         whichVecs.modify_host ();
+        // DEEP_COPY REVIEW - NOT TESTED FOR CUDA BUILD
```
tasmith4, line 1642: host to host, not tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -1634,10 +1639,12 @@ namespace Tpetra {
...
         Kokkos::deep_copy (whichVecs.view_host (), whichVecsIn);
       }
       else {
         whichVecs.modify_device ();
+        // DEEP_COPY REVIEW - HOST-TO-DEVICE
```
tasmith4, line 1647: host to device, not tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -1865,6 +1872,7 @@ namespace Tpetra {
           // output buffers, so we have to make a copy of the local
           // sum.
           typename RV::non_const_type lclDots (Kokkos::ViewAllocateWithoutInitializing ("tmp"), numVecs);
+          // DEEP_COPY REVIEW - NOT TESTED
```
tasmith4, line 1875: this is almost certainly copying between the same memory space, but it's not obvious what that space is.
not tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -2203,6 +2211,7 @@ namespace Tpetra {
                    lclSums.data (), meansOut.data ());
       }
       else {
+        // DEEP_COPY REVIEW - NOT TESTED
```
tasmith4, line 2214: host to host, not tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -2232,6 +2241,7 @@ namespace Tpetra {
                    lclSums.data (), meansOut.data ());
       }
       else {
+        // DEEP_COPY REVIEW - HOST-TO-HOST - NOT TESTED FOR MPI BUILD
```
tasmith4, line 2244: not tested in CUDA build either

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3519,9 +3531,11 @@ namespace Tpetra {
 
     if (this->isConstantStride ()) {
       if (useHostVersion) {
+        // DEEP_COPY REVIEW - HOST-TO-DEVICE
```
tasmith4, line 3534: host to host, tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3519,9 +3531,11 @@ namespace Tpetra {
...
         Kokkos::deep_copy (A_view, srcView_host);
       }
       else {
+        // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
```
tasmith4, line 3538: device to host, tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3532,10 +3546,12 @@ namespace Tpetra {
 
         if (useHostVersion) {
           auto srcColView_host = Kokkos::subview (srcView_host, rowRange, srcCol);
+          // DEEP_COPY REVIEW - NOT TESTED
```
tasmith4, line 3549: host to host, not tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3532,10 +3546,12 @@ namespace Tpetra {
...
           Kokkos::deep_copy (dstColView, srcColView_host);
         }
         else {
           auto srcColView_dev = Kokkos::subview (srcView_dev, rowRange, srcCol);
+          // DEEP_COPY REVIEW - NOT TESTED
```
tasmith4, line 3554: device to host, not tested

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -4352,6 +4368,7 @@ namespace Tpetra {
           // temporary host copy and print that.
           auto X_dev = getLocalViewDevice ();
           auto X_host_copy = Kokkos::create_mirror_view (X_dev);
+          // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
tasmith4, line 4371: tested

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7634,7 +7644,9 @@ namespace Tpetra {
     {
       typename local_graph_type::row_map_type::const_type::HostMirror k_rowPtrs_host_this = Kokkos::create_mirror_view(this->k_rowPtrs_);
       typename local_graph_type::row_map_type::const_type::HostMirror k_rowPtrs_host_graph= Kokkos::create_mirror_view(graph.k_rowPtrs_);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 7647: OK; note that doxygen for function isIdenticalTo is incorrect.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7634,7 +7644,9 @@ namespace Tpetra {
...
       Kokkos::deep_copy(k_rowPtrs_host_this, this->k_rowPtrs_);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 7649: OK; note that doxygen for function isIdenticalTo is incorrect.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7646,7 +7658,9 @@ namespace Tpetra {
     {
       typename local_graph_type::entries_type::non_const_type::HostMirror k_lclInds1D_host_this = Kokkos::create_mirror_view(this->k_lclInds1D_);
       typename local_graph_type::entries_type::non_const_type::HostMirror k_lclInds1D_host_graph= Kokkos::create_mirror_view(graph.k_lclInds1D_);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 7661: OK; note that doxygen for function isIdenticalTo is incorrect.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7646,7 +7658,9 @@ namespace Tpetra {
...
       Kokkos::deep_copy(k_lclInds1D_host_this, this->k_lclInds1D_);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 7663: OK; note that doxygen for function isIdenticalTo is incorrect.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7658,7 +7672,9 @@ namespace Tpetra {
     {
       typename t_GlobalOrdinal_1D::HostMirror k_gblInds1D_host_this  = Kokkos::create_mirror_view(this->k_gblInds1D_);
       typename t_GlobalOrdinal_1D::HostMirror k_gblInds1D_host_graph = Kokkos::create_mirror_view(graph.k_gblInds1D_);
+      // DEEP_COPY REVIEW - NOT TESTED
```
kddevin, line 7675: I'm not sure what condition would exercise this branch.
This routine is used only for comparisons in swap test.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7658,7 +7672,9 @@ namespace Tpetra {
...
       Kokkos::deep_copy(k_gblInds1D_host_this, this->k_gblInds1D_);
+      // DEEP_COPY REVIEW - NOT TESTED
```
kddevin, line 7677: I'm not sure what condition would exercise this branch.
This routine is used only for comparisons in swap test.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7676,7 +7692,9 @@ namespace Tpetra {
     {
       typename local_graph_type::row_map_type::HostMirror lclGraph_rowmap_host_this  = Kokkos::create_mirror_view(this->lclGraph_.row_map);
       typename local_graph_type::row_map_type::HostMirror lclGraph_rowmap_host_graph = Kokkos::create_mirror_view(graph.lclGraph_.row_map);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 7695: OK

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7676,7 +7692,9 @@ namespace Tpetra {
...
       Kokkos::deep_copy(lclGraph_rowmap_host_this, this->lclGraph_.row_map);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 7697: OK

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7687,7 +7705,9 @@ namespace Tpetra {
     {
       typename local_graph_type::entries_type::HostMirror lclGraph_entries_host_this = Kokkos::create_mirror_view(this->lclGraph_.entries);
       typename local_graph_type::entries_type::HostMirror lclGraph_entries_host_graph = Kokkos::create_mirror_view(graph.lclGraph_.entries);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 7708: OK

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7687,7 +7705,9 @@ namespace Tpetra {
...
       Kokkos::deep_copy(lclGraph_entries_host_this, this->lclGraph_.entries);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 7710: OK

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7698,7 +7718,9 @@ namespace Tpetra {
     {
       typename local_graph_type::row_block_type::HostMirror lclGraph_rbo_host_this = Kokkos::create_mirror_view(this->lclGraph_.row_block_offsets);
       typename local_graph_type::row_block_type::HostMirror lclGraph_rbo_host_graph = Kokkos::create_mirror_view(graph.lclGraph_.row_block_offsets);
+      // DEEP_COPY REVIEW - NOT TESTED
```
kddevin, line 7721: I'm not sure what condition would exercise this branch.
This routine is used only for comparisons in swap test.

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -7698,7 +7718,9 @@ namespace Tpetra {
...
       Kokkos::deep_copy(lclGraph_rbo_host_this, this->lclGraph_.row_block_offsets);
+      // DEEP_COPY REVIEW - NOT TESTED
```
kddevin, line 7723: I'm not sure what condition would exercise this branch.
This routine is used only for comparisons in swap test.

packages/tpetra/core/src/Tpetra_BlockMultiVector_def.hpp
MicheldeMessieres: Fixed comment to be host-to-device. Thx.

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -7156,6 +7160,7 @@ namespace Tpetra {
       typedef Kokkos::Device<HES, HostSpace> host_device_type;
       Kokkos::View<const char*, host_device_type>
         exports_a_kv (exports_a.getRawPtr (), newAllocSize);
+      // DEEP_COPY REVIEW - NOT TESTED
```
kddevin, line 7163: Yes, this case is exporting/importing into a CrsMatrix from a RowMatrix that cannot be cast to CrsMatrix.
It is untested in Tpetra and unexercised in Trilinos.

packages/tpetra/core/src/Tpetra_Details_computeOffsets.hpp
```diff
@@ -294,6 +294,7 @@ computeOffsetsFromCounts (const ExecutionSpace& execSpace,
             countsAPS(Kokkos::ViewAllocateWithoutInitializing("APS_counts"),
                       numCounts);
 
+    // DEEP_COPY REVIEW - NOT TESTED
```
kddevin, line 297: The code in this ifdef block was added to help debug / workaround the so-called "UVM bug" that turned out to be a Kokkos::parallel_scan bug.  The ifdef block should be removed.  I will do it.

packages/tpetra/core/src/Tpetra_Details_computeOffsets.hpp
```diff
@@ -302,6 +303,7 @@ computeOffsetsFromCounts (const ExecutionSpace& execSpace,
 
     total = ptrAPS(numCounts);
 
+    // DEEP_COPY REVIEW - NOT TESTED
```
kddevin, line 306: The code in this ifdef block was added to help debug / workaround the so-called "UVM bug" that turned out to be a Kokkos::parallel_scan bug.  The ifdef block should be removed.  I will do it.

packages/tpetra/core/src/Tpetra_Util.hpp
```diff
@@ -926,10 +926,12 @@ namespace Tpetra {
       Kokkos::DualView<T*, DT> x_out (label, len);
       if (leaveOnHost) {
         x_out.modify_host ();
+        // DEEP_COPY REVIEW - NOT TESTED FOR CUDA BUILD
```
kddevin, line 929: We should change a multivector communication test to use a big vector (e.g., 1M entries).  
I think this code will be exercised with a big-enough vector.

packages/tpetra/core/src/Tpetra_Util.hpp
```diff
@@ -926,10 +926,12 @@ namespace Tpetra {
...
         Kokkos::deep_copy (x_out.view_host (), x_in);
       }
       else {
         x_out.template modify<DMS> ();
+        // DEEP_COPY REVIEW - HOST-TO-DEVICE
```
kddevin, line 934: OK

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
kddevin: @MicheldeMessieres Both arguments of this deep_copy have type RV, so the copy should be HOST-TO-HOST or DEVICE-TO-DEVICE, right?   

Also, following the same call chain as for lclNormImpl above, I suspect this code is tested.

packages/tpetra/core/src/Tpetra_Details_packCrsGraph_def.hpp
```diff
@@ -229,6 +229,7 @@ class NumPacketsAndOffsetsFunctor{
   //! Host function for getting the error.
   int getError () const {
     auto error_h = Kokkos::create_mirror_view (error_);
+    // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 232: OK; deep_copy of one error code back to host.

packages/tpetra/core/src/Tpetra_Details_packCrsGraph_def.hpp
```diff
@@ -835,6 +836,8 @@ packCrsGraph (const CrsGraph<LO, GO, NT>& sourceGraph,
   View<size_t*, HostSpace, MemoryUnmanaged>
     num_packets_per_lid_h (numPacketsPerLID.getRawPtr (),
                            numPacketsPerLID.size ());
+
+  // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 840: ok

packages/tpetra/core/src/Tpetra_Details_packCrsGraph_def.hpp
```diff
@@ -1004,6 +1008,7 @@ packCrsGraphWithOwningPIDs
   // have to copy them back to host.
   View<size_t*, HostSpace, MemoryUnmanaged> num_packets_per_lid_h
     (numPacketsPerLID.getRawPtr (), numPacketsPerLID.size ());
+  // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 1011: ok

packages/tpetra/core/src/Tpetra_Details_packCrsGraph_def.hpp
```diff
@@ -848,6 +851,7 @@ packCrsGraph (const CrsGraph<LO, GO, NT>& sourceGraph,
   }
   View<packet_type*, HostSpace, MemoryUnmanaged>
     exports_h (exports.getRawPtr (), exports.size ());
+  // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
kddevin, line 854: ok

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -382,6 +382,7 @@ namespace Tpetra {
     in_view_type numAllocPerRowIn (numEntPerRow.getRawPtr (), lclNumRows);
     nc_view_type numAllocPerRowOut ("Tpetra::CrsGraph::numAllocPerRow",
                                     lclNumRows);
+    // DEEP_COPY REVIEW - HOST-TO-DEVICE
```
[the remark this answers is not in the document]
MicheldeMessieres: Fixed comment

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -518,6 +519,7 @@ namespace Tpetra {
...
```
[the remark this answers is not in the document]
MicheldeMessieres: Fixed comment

packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -1249,6 +1251,7 @@ namespace Tpetra {
         std::cerr << os.str();
       }
       row_ent_type numRowEnt (ViewAllocateWithoutInitializing (label), numRows);
+      // DEEP_COPY REVIEW - VALUE-TO-DEVICE
```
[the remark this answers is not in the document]
MicheldeMessieres: Fixed comment

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
MicheldeMessieres: Tested in TpetraCore_MultiVector_UnitTests_MPI_4

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
MicheldeMessieres: I confirmed a bunch of Tpetra unit tests exercise this deep_copy.

MicheldeMessieres: Fixed the comment to device-to-device. I'm calling it 'device' even when it really could be host or device. I'm calling things 'host' only if they are always host.

packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
```diff
@@ -1733,6 +1735,7 @@ namespace Tpetra {
       using host_norms_view_type = Kokkos::View<mag_type*, Kokkos::HostSpace>;
       host_norms_view_type h_norms ("Tpetra::MV::h_norms", norms.extent (0));
       this->norm1 (h_norms);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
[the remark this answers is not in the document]
MicheldeMessieres: Fixed comment - host-to-device

packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
```diff
@@ -1908,6 +1914,7 @@ namespace Tpetra {
       using host_norms_view_type = Kokkos::View<mag_type*, Kokkos::HostSpace>;
       host_norms_view_type h_norms ("Tpetra::MV::h_norms", norms.extent (0));
       this->normInf (h_norms);
+      // DEEP_COPY REVIEW - DEVICE-TO-HOST
```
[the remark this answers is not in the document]
MicheldeMessieres: Fixed comment - host-to-device

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
MicheldeMessieres: Fixed comment - device-to-host

csiefer2:
Updating to the tpetraDeepCopy branch on this fork.  Will close this PR once the new guy is ready

PR #7964, opened 2020-09-02
Tpetra: Fix initialize tests for Pascal node on ride/white

MicheldeMessieres:
@kddevin @csiefer2 I'd like to get the test suite all passing on white for the Pascal node with CUDA_LAUNCH_BLOCKING=0. The last thing remaining is to fix these Tpetra initialization tests. The first thing to resolve is that some don't pass on Pascal nodes for white, unrelated to the CUDA_LAUNCH_BLOCKING=0 issue.

The issue is that MPI is not setup before Kokkos::initialize in some cases.
I have been fixing other tests which have the same problem.

This PR is a working solution where I added a GlobalMPISession before the Kokkos::initialize. But then the GlobalMPISession is the part that actually calls MPI_Finalize, not say Tpetra:ScopeGuard.

I'm not sure we want to make changes like this though. Perhaps some of these tests have patterns which we don't really want to support anyways. Or perhaps we want to find a way so Kokkos can be 'updated' even if MPI inits after the Kokkos::initialize call. I hope this PR can be a discussion starter for this.

I also don't fully understand the underlying issue here. In Tpetra copyConvert.cpp we have this comment and code:
```
  // Initialize MPI (if enabled) before initializing Kokkos.  This
  // lets MPI control things like pinning processes to sockets.
  Teuchos::GlobalMPISession mpiSession (&argc, &argv);
  Kokkos::initialize (argc, argv);
```
So that I presume is why these are segfaulting when they are not setup in the right order. I could spend time to look into those details further.

@trilinos/tpetra 

## Motivation

Get the initialization tpetra tests to work on white's Pascal nodes.

## Testing

Cuda white pascal unit tests for tpetra

### DIFF
--- a/packages/tpetra/core/test/Core/ScopeGuard_tpetra_inits_mpi_user_inits_kokkos.cpp
+++ b/packages/tpetra/core/test/Core/ScopeGuard_tpetra_inits_mpi_user_inits_kokkos.cpp
@@ -89,6 +89,10 @@ void testMain (bool& success, int argc, char* argv[])
       "before Kokkos::initialize was called." << endl;
     return;
   }
+
+  { // GlobalMPISession is in local scope so we can check MPI_Finalize
+
+  Teuchos::GlobalMPISession mpiSession(&argc, &argv); // before Kokkos::initialize
   Kokkos::initialize (argc, argv);
   if (! Kokkos::is_initialized ()) {
     success = false;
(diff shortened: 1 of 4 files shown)


csiefer2 (approved):
As per the Kokkos team, initializing MPI first is a "best practice:"

Issue #7966, opened 2020-09-02
Tpetra: Fix initialization tests to properly capture errors from Kokkos::initialize


MicheldeMessieres:
## Enhancement

@kddevin @csiefer2 This perhaps can be considered more of a bug since tests are not working properly, but it only impacts the tests themselves. The Tpetra initialization tests in packages/tpetra/core/test/Core capture std::cerr warning statements from Kokkos::initialize and are supposed to fail if the output is not empty. However for multiple ranks, the capture may be empty for some ranks and then the test will output text as FAIL PASS PASS PASS and the PASS gets picked up as sufficient to pass the test. The output should be FAIL FAIL FAIL FAIL as all ranks are writing the error to std::cerr.

I'm not sure we want to check just the root rank since the ranks may not all have the same error. It's not clear to me yet why it doesn't work. I have tried moving the capture so it always happens after MPI_Init is setup but that didn't seem to be a factor. I can investigate this further.

@trilinos/tpetra

kddevin:
These are truly awful tests -- checking for output from a snapshotted TPL -- ugh.
See #3453 and #6748.  
It appears that in #6748, @ndellingwood disabled the Kokkos warnings for these tests so that the tests always pass.   But you are seeing them fail, right?  Weird.

MicheldeMessieres:
@kddevin To clarify I wasn't seeing extra failures because of this issue. I was seeing passing when the tests should fail. I think all of the initialization tests should currently fail with launch blocking off but currently only some of them do.

kddevin:
@MicheldeMessieres I will look at these tests so that we can move ahead with a CUDA_LAUNCH_BLOCKING=0 nightly test.  Do all the tests in packages/tpetra/core/test/Core pass when they should fail?   Or just a subset of them?
Thanks.

MicheldeMessieres:
@kddevin For the latest develop all should pass now on Pascal with CUDA_LAUNCH_BLOCKING=1.
With CUDA_LAUNCH_BLOCKING=0 I have the following fails:
```
	 47 - TpetraCore_Core_initialize_where_user_initializes_mpi_MPI_4 (Failed)
	 48 - TpetraCore_Core_ScopeGuard_where_user_initializes_mpi_MPI_4 (Failed)
```
I know some were passing that should fail due to std::cerr being empty on just some ranks. I didn't get to the bottom of why these were behaving differently.

PR #7967, opened 2020-09-02
Tpetra: Pass initialization tests with CUDA_LAUNCH_BLOCKING off

MicheldeMessieres:
Since packages are getting updated to work with CUDA_LAUNCH_BLOCKING=0
and Tpetra is completed, this allows the unit test suite to all pass for Tpetra.

The tests were checking for CUDA_LAUNCH_BLOCKING off and failing based on std::cerr output.
This keeps the warning but changes it to std::cout so the Tpetra test won't detect it.

I expect we'll want to update this warning again when Trilinos is all working with launch blocking off. Perhaps warn that user code needs fencing if they have it off and warn that they might want to turn it off for efficient Trilinos code, if they have it on.

Note we'll also need #7964 to get the tests passing on white Pascal nodes with blocking off.

@trilinos/tpetra 

## Motivation

Get Tpetra all passing for CUDA_LAUNCH_BLOCKING=0

## Testing

Cuda white Pascal node - Tpetra Core tests


csiefer2 (commented):
My thought is to leave this test failing until we're convinced the stack works without launch blocking.

But @kddevin might have other ideas.

kddevin:
@MicheldeMessieres  I see that you only are changing std::cerr to std::cout.  Is there a problem with writing to std::cerr?  Does TRIBITS / Ctest trigger a failure in that case? 

This code is actually in Kokkos, not Tpetra.  We would have to get any changes approved by the Kokkos team.  They may or may not have a standard for writing warnings to std::cerr; I don't know.

MicheldeMessieres:
@kddevin I was not paying attention to the fact this was in kokkos - so not the right plan.
The issue is that the Tpetra tests captures the std::cerr stream and will fail if it's not empty. That's not really working anyways (see #7966). So if we want to get these tests passing now maybe we just don't want to check std::cerr from kokkos. There are other warnings though that Tpetra could catch from kokkos.

PR #7971, opened 2020-09-03
Tpetra: Convert deep_copy to output view type exec space to use 3-arg

MicheldeMessieres:
@csiefer2 These are cases where I get the execution space from the output view. I am not certain if this is always what we want. In some of these cases we could have device-to-host. Would we potentially want to queue device-to-host deep_copy on the device execution space in some cases and would we want user to control that?

@trilinos/tpetra

## Motivation

Convert deep_copy to 3-arg forms

## Testing

Cuda on white pascal node - Tpetra unit tests


csiefer2 (approved):

csiefer2:
That looks great!

MicheldeMessieres:
Will reopen this in a different form.

PR #7984, opened 2020-09-04
MueLu: Use SerialNode for Epetra vs. Tpetra test

brian-kelley:
@trilinos/tpetra 

## Motivation

Always use SerialNode when comparing Tpetra vs. Epetra SpMV performance. This is to make a fair comparison, when CudaNode is enabled for Tpetra. Currently will just affect the performance testing build on Vortex, since Stria and Eclipse were already using SerialNode only.

## Testing

N/A


lucbv (approved):
That's fine

PR #7387, opened 2020-05-15
Tpetra:  Include multivector length in deciding where kernels should run

GeoffDanielson:
@trilinos/tpetra 

## Motivation
Implements @vbrunini 's changes in #7315 via Tpetra::Details::Behavior environment variables.  Defaults to his value of 10000, but allows anyone to change the threshold via the TPETRA_VECTOR_DEVICE_THRESHOLD environment variable.

## Related Issues

* Closes #7315 

## Testing
Added a couple of long multivector tests.  Existing and new unit tests pass on vesper and eclipse.

PR #7985, opened 2020-09-04
Adjust runKernelOnHost threshold

GeoffDanielson:
@trilinos/tpetra 

## Motivation
This adjustment is based on experimental observations of kernel behaviors on vortex.  The performance inflection point between running on host and device is about 22,000.

## Related Issues

* Follows https://github.com/trilinos/Trilinos/pull/7387

## Testing
This touches nothing that isn't already unit-tested.

Experimental data, using thresholds of 0 and 1,000,000,000:
xoverpoint


kddevin (changes requested):
There's still a hard-coded value of 10000 in Tpetra_Multivector_def.hpp.  Along with this change, can you please change the hard-coded value to use the Behavior?

GeoffDanielson:
> There's still a hard-coded value of 10000 in Tpetra_Multivector_def.hpp. Along with this change, can you please change the hard-coded value to use the Behavior?

Fixed the other instance.

kddevin (approved):

PR #7975, opened 2020-09-03
Tpetra: Convert deep_copy to 3-arg

MicheldeMessieres:
@trilinos/tpetra 

## Motivation

Convert deep_copy to 3-arg

## Testing

Cuda white Pascal node launch blocking off

PR #7986, opened 2020-09-04
Stokhos: Add 3-arg deep_copy specializations.

MicheldeMessieres:
@csiefer2 @kddevin Just FYI

@etphipp Adds 3-arg deep_copy specializations for Stokhos. This is needed for current work to refactor Tpetra to use 3-arg deep_copy. (#7975 #7971 for example).

2 questions for this:

(1) I duplicated a chunk of code which is not desirable. I can fix this but wanted to get some rough guidance from you first how you'd like this organized. If we accept this PR I could work on fixing that and proceed with Tpetra work in parallel.

(2) I see some other 3-arg deep_copy overloads for Stokhos don't use the ExecSpace.  For the new methods I added, I just changed all internal deep_copy calls to 3-arg using the ExecSpace. I wasn't sure if there might be reasons not to do this.

@trilinos/stokhos 

## Motivation

Refactor Tpetra deep_copy calls to use 3-arg deep_copy.

## Testing

Mac serial build and cuda white parallel build.

### DIFF
--- a/packages/sacado/src/KokkosExp_View_Fad.hpp
+++ b/packages/sacado/src/KokkosExp_View_Fad.hpp
@@ -456,6 +456,50 @@ void deep_copy( const View<DT,DP...> & dst ,
   Kokkos::deep_copy( dst_array , src_array );
 }
 
+/* Specialize for deep copy of FAD */
+template< class ExecSpace, class DT , class ... DP , class ST , class ... SP >
+inline
+void deep_copy( const ExecSpace &,
+                const View<DT,DP...> & dst ,
+                const View<ST,SP...> & src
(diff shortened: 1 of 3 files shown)


etphipp (approved):
Looks good to me.  If it would be possible to eventually consolidate the implementations of the versions that take the execution_space argument and those that don't, that would be ideal.

MicheldeMessieres:
@etphipp I opened #7988 to keep track of the need to fix the duplication - I have that currently assigned to myself. I need to think on how we design this without making it too complicated. The 2-arg specialization won't have an exec space to pass to the shared method.

etphipp:
Great, thanks!

PR #7991, opened 2020-09-09
Tempus tscs pid patch

sconde:
@trilinos/tempus

### DIFF
--- a/packages/tempus/src/Tempus_TimeStepControlStrategyPID.hpp
+++ b/packages/tempus/src/Tempus_TimeStepControlStrategyPID.hpp
@@ -69,7 +69,7 @@ public:
      */
 
      Scalar beta = 1.0;
-     const std::string strategy_name = this->tscsPL_->get<std::string>("Name");
+     const std::string strategy_name = this->tscsPL_->template get<std::string>("Name");
      Scalar k1 = Teuchos::as<Scalar>(-k1_ / order);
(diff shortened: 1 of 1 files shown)


sconde:
@jmgate can we get this patch merged?

packages/tempus/src/Tempus_TimeStepControlStrategyPID.hpp
```diff
@@ -139,6 +138,7 @@ class TimeStepControlStrategyPID
      }
      tscsPL_->validateParametersAndSetDefaults(*this->getValidParameters());
 
+     strategy_name_ = tscsPL_->get<std::string>("Name");
```
seamill, line 141: @sconde Why don't we need the "template" keyword here? I don't see a difference between this and what we had.

jmgate (approved):

seamill (approved):

seamill:
@jmgate It looks like an unrelated panzer error. Can we hotfix this fix into the EMPIRE fork for now?

jmgate:
Hotfixed into EM-Plasma/Trilinos.  Why are we seeing a Panzer failure here, though?

seamill:
Given how today is going, I'm just going to say its just another glitch in the matrix. We can try testing again in a few hours and if it fails again we will look into it. Our change should not affect anything on the Panzer side, only the downstream codes like Drekar/EMPIRE.

PR #7994, opened 2020-09-09
Muelu: regionMG add rebalance before coarse AMG iterations

pohm01:
@trilinos/muelu 

## Motivation

Following structured coarsening in the region format, each rank may only own a few dofs. Here we rebalance the coarse grid operator before constructing the coarser grid AMG hierarchy in an attempt to reduce communication.

Also, fixes the call to iterate on the coarse AMG hierarchy.

## Testing

Tests have been added to the regionMG tests that utilize this rebalancing. 

## Additional Information
The number of partitions for the coarse gird repartitioning is an input (default is 1).

When more than one iteration of the coarse grid AMG iteration are used (so deviating from a V-cycle structure, and approaching the AMG as a coarse level solve), the convergence of rebalancing before building the coarse AMG is nearly identical to the converge of not rebalancing (convergence is not identical since different partitions result in different AMG iterations).

### DIFF
--- a/packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
+++ b/packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
@@ -66,7 +66,7 @@
 //#include <MueLu_RebalanceAcFactory_decl.hpp>
 //#include <MueLu_RepartitionHeuristicFactory.hpp>
 #include <MueLu_RepartitionFactory.hpp>
-#include <MueLu_ZoltanInterface.hpp>
+#include <MueLu_Zoltan2Interface.hpp>
 
 #include "SetupRegionVector_def.hpp"
 #include "SetupRegionMatrix_def.hpp"
@@ -479,9 +479,7 @@ void RebalanceCoarseCompositeOperator(const int maxRegPerProc,
   fos->setOutputToRootOnly(0);
   *fos << "Rebalancing coarse composite operator." << std::endl;
 
-  RCP<const Map> map = coarseCompOp->getMap();
-
-  RCP<Xpetra::Vector<GlobalOrdinal,LocalOrdinal,GlobalOrdinal,Node> > decomposition = Xpetra::VectorFactory<GlobalOrdinal,LocalOrdinal,GlobalOrdinal,Node>::Build(map, true);
+  //RCP<const Map> map = coarseCompOp->getMap();
 
   const int numPartitions = rebalanceNumPartitions;
 
@@ -492,7 +490,7 @@ void RebalanceCoarseCompositeOperator(const int maxRegPerProc,
   RCP<FactoryManagerBase> factoryHandler = rcp(new FactoryManager());
   level.SetFactoryManager(factoryHandler);
 
-  RCP<ZoltanInterface> zoltan = rcp(new ZoltanInterface());
+  RCP<Zoltan2Interface> zoltan = rcp(new Zoltan2Interface());
 
   level.Set<RCP<Matrix> >     ("A",                    coarseCompOp);
   level.Set<RCP<MultiVector> >("Coordinates",          compCoarseCoordinates);
@@ -513,7 +511,7 @@ void RebalanceCoarseCompositeOperator(const int maxRegPerProc,
   level.Get("Importer", rebalanceImporter, repart.get());
 
   ParameterList XpetraList;
-  XpetraList.set("Restrict Communicator",false);
+  XpetraList.set("Restrict Communicator", true);
   XpetraList.set("Timer Label","MueLu::RebalanceAc-for-coarseAMG");
 
   // Build rebalanced coarse composite operator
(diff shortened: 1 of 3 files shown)


packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
```diff
@@ -1176,13 +1270,28 @@ void vCycle(const int l, ///< ID of current level
...
+
+          RCP<Vector> rebalancedRhs = VectorFactory::Build(rebalanceImporter->getTargetMap());
+          RCP<Vector> rebalancedX = VectorFactory::Build(rebalanceImporter->getTargetMap());
+          rebalancedRhs->doImport(*compRhs, *rebalanceImporter, Xpetra::INSERT);
+          rebalancedX->doImport(*compX, *rebalanceImporter, Xpetra::INSERT);
+
+          amgHierarchy->Iterate(*rebalancedRhs, *rebalancedX, 1, true);
+
+          compRhs->doExport(*rebalancedRhs, *rebalanceImporter, Xpetra::INSERT);
+          compX->doExport(*rebalancedX, *rebalanceImporter, Xpetra::INSERT);
```
pohm01, line 1293: @lucbv How do I want to do this? Is the export OK or do I want to build the correct `import` and do a `doImport`

lucbv: In my mind the above is not correct, I have asked Chris about it, let us see what he says...

lucbv: Hum, looking carefully at the `doExport` documentation in Tpetra I see that with an importer as input it performs a reverse import operation so that above actually makes more sense to me know!

cgcgcg: Does `rebalancedX` actually need to be communicated? Or are its values just ignored? Calling `Iterate` with last arg `true` seems to suggest that you don't need this.

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
```diff
@@ -455,6 +460,72 @@ MakeCompositeDirectSolver(RCP<Xpetra::Matrix<Scalar, LocalOrdinal, GlobalOrdinal
...
+
+  // Build
+  level.Request("Importer", repart.get());
+  repart->Build(level);
+
+  // Build importer for rebalancing
+  level.Get("Importer", rebalanceImporter, repart.get());
+
+  ParameterList XpetraList;
+  XpetraList.set("Restrict Communicator",false);
```
cgcgcg, line 516: Is there a reason for not restricting?

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
```diff
@@ -455,6 +460,72 @@ MakeCompositeDirectSolver(RCP<Xpetra::Matrix<Scalar, LocalOrdinal, GlobalOrdinal
...
+  const int numPartitions = rebalanceNumPartitions;
+
+  // We build a fake level
+  Level level;
+  level.SetLevelID(1);
+
+  RCP<FactoryManagerBase> factoryHandler = rcp(new FactoryManager());
+  level.SetFactoryManager(factoryHandler);
+
+  RCP<ZoltanInterface> zoltan = rcp(new ZoltanInterface());
```
cgcgcg, line 495: Does this work with Tpetra?

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
lucbv: Actually we really should restrict since a bunch of ranks will have no data (at least in our paper test case)

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
pohm01: It does, unless I have tricked myself into not actually using Tpetra. But should change this to zoltan2 right?

lucbv: I be believe that Zoltan might have predated Epetra and it could rely only on plain old data pointers... but Zoltan2 would make more sense.

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
```diff
@@ -1283,13 +1282,17 @@ void vCycle(const int l, ///< ID of current level
           RCP<const Import> rebalanceImporter = coarseSolverData->get<RCP<const Import> >("rebalanceImporter");
 
           RCP<Vector> rebalancedRhs = VectorFactory::Build(rebalanceImporter->getTargetMap());
-          RCP<Vector> rebalancedX = VectorFactory::Build(rebalanceImporter->getTargetMap());
+          RCP<Vector> rebalancedX = VectorFactory::Build(rebalanceImporter->getTargetMap(), true);
```
cgcgcg, line 1285: You could cache these vectors.

lucbv: That's true, I don't think we need to do that now if the performance we see on Cori is decent.
The issue with not caching things well probably stems a little bit from the fact that we are not using a "level" like container anywhere in this code and that leads to the catastrophic interface call with 40+ input parameters...
I feel it would be nice if we can add a comment here to remember that issue and tackle it when we refactor the code and introduce a container to simplify this type of things a lot.

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
```diff
@@ -63,6 +63,11 @@
 #include <MueLu_CreateXpetraPreconditioner.hpp>
 #include <MueLu_Utilities.hpp>
 
+//#include <MueLu_RebalanceAcFactory_decl.hpp>
+//#include <MueLu_RepartitionHeuristicFactory.hpp>
```
lucbv, line 67: It might be nice to remove the above comments if we do need the headers?

lucbv (changes requested):
Just a couple of small modifications please : )

lucbv (approved):
lgtm

PR #7997, opened 2020-09-10
Zoltan2 Sphynx: Set default values for parameters

seheracer:
@trilinos/zoltan2 

## Motivation

This PR adds a function which determines the default values for some parameters in Sphynx: preconditioner type, eigenvalue problem type, tolerance, and initial guess.
 

## Testing

Tests in the local CUDA builds are passing.


egboman (approved):
This is useful to get good partitioning results.

PR #7998, opened 2020-09-10
Tpetra: Add a new reader interface to get the distribution object

seheracer:
@kddevin 

## Motivation

This PR adds a new interface for readSparseFile() to get the distribution object as an output parameter. This change is needed to access the chunkCuts computed in the LowerTriangularBlock distribution type.

## Testing

I tested this using a triangle counting driver.


kddevin (changes requested):
Hi, @seheracer 
Thanks for the submission.  I'd rather not replicate the entire function, as having two nearly identical versions will be difficult to maintain.  Can we do something like this instead?

Keep your version of the function (with the Distribution object returned).
Remove the old version of the function (without the Distribution object returned).
Add a new method with the old interface that calls your new version:
```
static Teuchos::RCP<sparse_matrix_type>
readSparseFile(
  const std::string &filename,    // MatrixMarket file to read
  const Teuchos::RCP<const Teuchos::Comm<int> > &comm,  
  const Teuchos::ParameterList &params
)
{
  Teuchos::RCP<Distribution<global_ordinal_type,scalar_type> > dist;
  return readSparseFile(filename, comm, params, dist);
}

Will this work?

seheracer:
> Hi, @seheracer
> Thanks for the submission. I'd rather not replicate the entire function, as having two nearly identical versions will be difficult to maintain. Can we do something like this instead?
> 
> Keep your version of the function (with the Distribution object returned).
> Remove the old version of the function (without the Distribution object returned).
> Add a new method with the old interface that calls your new version:
> 
> ```
> static Teuchos::RCP<sparse_matrix_type>
> readSparseFile(
>   const std::string &filename,    // MatrixMarket file to read
>   const Teuchos::RCP<const Teuchos::Comm<int> > &comm,  
>   const Teuchos::ParameterList &params
> )
> {
>   Teuchos::RCP<Distribution<global_ordinal_type,scalar_type> > dist;
>   return readSparseFile(filename, comm, params, dist);
> }
> ```
> 
> Will this work?

I like this approach. I am currently testing this.

seheracer:
@kddevin I pushed the updated version. It looks much better!

kddevin (approved):

PR #8000, opened 2020-09-10
Tempus: app action for newmark explicit a form

### LINKED ISSUE #7892: Tempus:  AppAction for Newmark Explicit A Form
## Enhancement
@trilinos/tempus 

Add AddAction interface for Newmark Explicit A Form

sconde:
- add unit_test for modifier and modifierX

@trilinos/tempus 

## Motivation

Add AppAction interface for Newmark Explicit A Form

### DIFF
--- a/packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
+++ b/packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
@@ -11,6 +11,7 @@
 
 #include "Teuchos_VerboseObjectParameterListHelpers.hpp"
 #include "Thyra_VectorStdOps.hpp"
+#include "Tempus_StepperNewmarkExplicitAFormModifierDefault.hpp"
 
 //#define DEBUG_OUTPUT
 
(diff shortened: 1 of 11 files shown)


packages/tempus/src/Tempus_StepperNewmarkExplicitAForm_impl.hpp
```diff
@@ -276,6 +307,12 @@ void StepperNewmarkExplicitAForm<Scalar>::takeStep(
     predictDisplacement(*d_new, *d_old, *v_old, *a_old, dt);
     predictVelocity(*v_new, *v_old, *a_old, dt);
 
+    stepperNewmarkExpAppAction_->execute(solutionHistory, thisStepper,
+        StepperNewmarkExplicitAFormAppAction<Scalar>::ACTION_LOCATION::BEFORE_SOLVE);
+
+    stepperNewmarkExpAppAction_->execute(solutionHistory, thisStepper,
+        StepperNewmarkExplicitAFormAppAction<Scalar>::ACTION_LOCATION::AFTER_SOLVE);
+
```
ccober6, line 315: Let's go with BEFORE_EXPLICIT_EVAL and AFTER_EXPLICIT_EVAL around the this->evaluateExplicitODE(a_new, d_new, v_new, time_old, p) call. There is no solve in the explicit Newmark.

sconde: I tried to stay consistent with ERK. There's no solve there either but similar call is made. What do you think?

ccober6: So we made the AppAction consistent across ERK, DIRK and IMEX, because we can "derive" ERK and DIRK from IMEX, so to allow RKAppActions to be usable between all the RK methods, they all have all the RKAppAction locations.  However to do that across all steppers seemed excessive and possibly confusing.  So the other steppers have more tailored AppActions and likely have AppAction locations that others do not.  If anything this stepper, Newmark explicit, would look closer to ForwardEuler, which has BEGIN_STEP, BEFORE_EXPLICIT_EVAL, and END_STEP.  I thought AFTER_EXPLICIT_EVAL might be useful because of the correctVelocity and getUseFSAL block.  Does this seem reasonable?

sconde: That makes sense. Thank you.

ccober6 (approved):
Looks good!

PR #8006, opened 2020-09-11
MueLu: Ifpack2: implement lumped diagonal scaling

jhux2:
This implements a matrix diagonal scaling suggested by Ray Tuminaro.
The default is still to use the inverse of the point diagonal.
This commit adds an option to use as the diagonal the row sums
of absolute values:

   d_ii = sum(abs(A(ii,:)))

The reciprocal of this "lumped" diagonal  is used in the power method
used to estimate the maximum eigenvalue of D \ A and in the Chebyshev smoother.

@trilinos/muelu 

## Motivation

This implements an idea suggested by Ray Tuminaro to use as the diagonal the row sum of abs(A).

## Testing

---
This change is [Reviewable](https://reviewable.io/reviews/trilinos/trilinos/8006)


rstumin:
Since we have an option to cache ... or reuse the eigenvalue ... does this mess up anything when we do prolongator smoothing? In reading the code, I'm not totally sure what happens in the prolongator smoother  if you decide to use this lumped diagonal stuff for Chebyshev. We definitely don't want to use an eigenvalue computation made with the lumped diagonal ... but then use it with the actual diagonal . Has this been tested?

jhux2:
> Since we have an option to cache ... or reuse the eigenvalue ... does this mess up anything when we do prolongator smoothing? In reading the code, I'm not totally sure what happens in the prolongator smoother if you decide to use this lumped diagonal stuff for Chebyshev. We definitely don't want to use an eigenvalue computation made with the lumped diagonal ... but then use it with the actual diagonal . Has this been tested?

@rstumin I've only run unsmoothed AMG so far.  With the lumping enabled, the 32-MPI rank McAlister problem that previously didn't converge is now solvable.  You're right that the eigenvalue will be cached.    After I pushed this, I realized that SA-AMG will probably still use the point diagonal.  Let me look into whether the lumped diagonal can be also be used in prolongator smoothing.   I'm guessing that this will be easy in some cases, but not, for example, with the fused Jacobi kernel.

rstumin:
@jhux2 At the very least ... if we cannot use the lumped diagonal in the prolongator smoother ... we definitely don't want to cache the eigenvalue. However, when there is no filtering done on the matrix ... then I think we only have one copy of the matrix ... which means that there can only be one eigenvalue? Is this right? In the filtering case, you can at least have two different eigenvalues.

jhux2:
@rstumin There's now a single option that will enable the absrowsum diagonal for both the smoother (if it's Chebyshev) and the prolongator smoothing.  Please have a look.  The tests will probably croak because I haven't update the gold files yet.  I'll porbably wait until #8068 is merged to make your life easier....

rstumin (dismissed):
@jhux2 This looks good. I was just thinking that someday this could be generalized to allow for a block diagonal option when we have PDE systems. The key word is 'someday', certainly not today.

jhux2:
@rstumin Please have another look.

packages/muelu/src/Utils/MueLu_UtilitiesBase_decl.hpp
```diff
@@ -510,22 +510,47 @@ namespace MueLu {
...
+      Scalar lambda = PowerMethod(A, diagInvVec, niters, tolerance, verbose, seed);
+      return lambda;
+    }
 
     /*! @brief Power method.
 
     @param A matrix
-    @param scaleByDiag if true, estimate the largest eigenvalue of \f$ D^; A \f$.
+    @param invDiag reciprocal of matrix diagonal
     @param niters maximum number of iterations
```
rstumin, line 546: @jhux2 It looks like the parameter is now called diagInvVec .... while the above @param comment refers to invDiag

packages/muelu/src/Utils/MueLu_Utilities_kokkos_decl.hpp
```diff
@@ -172,7 +172,7 @@ namespace MueLu {
 
     NOTE -- it's assumed that A has been fillComplete'd.
     */
-    static RCP<Vector> GetMatrixDiagonalInverse(const Matrix& A, Magnitude tol = TST::eps()*100); // FIXME
+    static RCP<Vector> GetMatrixDiagonalInverse(const Matrix& A, Magnitude tol = TST::eps()*100, const bool doLumped = false); // FIXME
```
rstumin, line 175: @jhux2 I'm not a big fan of the term "Lumped" or any reference to Lump when describing the rowSumAbs thing. I don't have a great alternative other than saying rowSumAbs. Lumping I feel refers to the moving of off diagonal entries (so zeroing out some entries) and grouping them with others (usually to the diagonal). This is what it means at least in the finite element community. The idea here is to produce a sparsified matrix with a similar effect as the original matrix. This is not really the same thing as taking the row sum of abs(a_ij). In this case, we are looking to scale the matrix (which can be thought of as a super cheap approximation of the matrix inverse). Normally, we use the diagonal and invert it ... but in this case we use the rowSumAbs thing.

rstumin (dismissed):
@jhux2 The code looks pretty good to me. I made two comments (outside of this review). One is about generally not liking the word 'lumped' in the context of what you are doing. The other is a minor thing that at move points in the power method, we use invDiag ... but there are a couple of spots where the parameter is instead diagInvVec. I'm going to go ahead and approve this, but you might consider these minor naming changes.

jhux2:
> @jhux2 The code looks pretty good to me. I made two comments (outside of this review). One is about generally not liking the word 'lumped' in the context of what you are doing. The other is a minor thing that at move points in the power method, we use invDiag ... but there are a couple of spots where the parameter is instead diagInvVec. I'm going to go ahead and approve this, but you might consider these minor naming changes.

I fixed the comment.  Regarding the use of "lumping", that's what the name of the static utility method was.  I've stayed away from using that in the documentation or in `SaPFactory`.  In both those places, I use some variant of "rowsumabs".

rstumin:
@jhux2  Just discovered that when we do normal smoothed aggregation ... there is code in GetInverse() to basically jam in the diagonal is really small. However, there is no equivalent possibility when we do absRowSum ... there is just a check to put in a zero if the absRowSum thing is identically 0 ... then we don't take a reciprocal and just stick a zero in the inv diag. However, if the number is really small, we will take the reciprocal. This is kind of bad when all entries are dropped in the filter matrix. So in this case, the diagonal is the only value in the filter matrix ... and this number will likely be real small (as it normally has a zero row sum). As  max(Dinv A) \approx 1 when using absRowSum ... then the prolongator smoothing step for this row looks like  1 - (4/3) 1 =  -1/3. That is, we get a negative entry in P. Here, the first '1' in the expression comes from Ptent (without QR). The second '1' comes from  Dinv A being 1 when there is just one nonzero in a row of filtered A.

jhux2:
@rstumin wrote:
> Just discovered that when we do normal smoothed aggregation ... there is code in GetInverse() to basically jam a zero in the inverse diag if the diagonal is really small. However, there is no equivalent possibility when we do absRowSum ... there is just a check to put in a zero if the absRowSum thing is identically 0 ... then we don't take a reciprocal and just stick a zero in the inv diag. However, if the number is really small, we will take the reciprocal. This is kind of bad when all entries are dropped in the filter matrix. So in this case, the diagonal is the only value in the filter matrix ... and this number will likely be real small (as it normally has a zero row sum). As max(Dinv A) \approx 1 when using absRowSum ... then the prolongator smoothing step for this row looks like 1 - (4/3) 1 = -1/3. That is, we get a negative entry in P. Here, the first '1' in the expression comes from Ptent (without QR). The second '1' comes from Dinv A being 1 when there is just one nonzero in a row of filtered A.

I've propagated the use of tolerance and replacement tolerance.  The default behavior for the rowsumabs option is to replace a "too-small" diagonal with zero.

csiefer2 (approved):

PR #8011, opened 2020-09-11
Xpetra: Guard Kokkos fences for Tpetra only

### LINKED ISSUE #8010: Xpetra: Kokkos::fence with Epetra
## Bug Report
@trilinos/xpetra 

### Description
Build failure on an Epetra-only build:
https://testing.sandia.gov/cdash/viewBuildError.php?buildid=7850071
The call to `Kokkos::fence` needs to be guarded.

MicheldeMessieres:
@trilinos/xpetra 

## Motivation

Resolve #8010 
guard Kokkos::fence for Epetra only build

## Testing

Mac Tpetra and Epetra-only builds.

### DIFF
--- a/packages/xpetra/test/CrsMatrix/CrsMatrix_UnitTests.cpp
+++ b/packages/xpetra/test/CrsMatrix/CrsMatrix_UnitTests.cpp
@@ -295,7 +295,10 @@ namespace {
 
     A->leftScale(*s);
 
+#ifdef HAVE_XPETRA_TPETRA
     Kokkos::fence();
+#endif
+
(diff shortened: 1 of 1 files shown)


cgcgcg (approved):
This will do the trick.
(It will still call `fence` when built with Tpetra, but running with Epetra, but that's hopefully a no-op.)

MicheldeMessieres:
@cgcgcg Host build Kokkos::fence is no operation. I don't expect to add any more of these fences to Xpetra and it's just for the test - so probably not worth making a higher level system unless things get more complicated.

PR #8012, opened 2020-09-11
Framework: PR Driver Overhaul and Cleanup

william76:
@trilinos/framework 

## WORK IN PROGRESS
This is still a work in progress but it's to a place where it is running on the current cloud nodes that we have. This got more kludgy than I'd have liked it to because our cloud nodes do not have the same environment as the usual CEE nodes that we run on. This has especially caused headaches w/rt to trying to support both Python2 and Python3.

I'd recommend us getting that sorted out and ditching Python2 since it's not supported and is at end-of-life as of Jan 1, 2020.

This PR is to enable some testing while I also do some cleanup passes (especially if we can remove Python2 dependencies) as well as code reviews within the Framework team.

## Motivation
Some cleanup / overhaul work on the PR driver script(s) to break things up into a more modular set of tools rather than a monolithic file. This is related to installation testing development since installation testing is a multi-step thing. This lays the groundwork for us to move towards that while also improving the testing coverage of our Python infrastructure.

## Detailed Description of Changes
TBD

## Related Issues
* Related to # 7 3 8 3  (not tagging it yet with this PR)

## Testing
lots.

## Additional Information
This is a pretty extensive overhaul of the PR driver, but it will get us to a better place to support new builds, etc.


william76:
git + rebase wrecked this branch pretty bad today... I burned an hour trying to figure out where the right commit was to try and revert but didn't find a good resolution for reverting to it without possibly deleting stuff that may contain updates I want.  I'll peel off another branch from the last known 'ok' commit and try and rebuild.

Issue #8008, opened 2020-09-11
ATDM: Make custom configuration the default when no other known configurations match


bartlettroscoe:
CC: @ndellingwood, @jmgate, @e10harvey 

It was requested in https://github.com/trilinos/Trilinos/pull/7749#issuecomment-690729110 to make it so that when using a custom ATDM Trilinos configuration on a system that does not support any other configuration (like 'weaver') that one can leave the custom system name out of the build name.  For example, currently on 'weaver' if you register the custom configuration:

```
$ cd Trilinos/
$ export ATDM_CONFIG_REGISTER_CUSTOM_CONFIG_DIR=$PWD/cmake/std/atdm/contributed/weaver
```

then you have to include the custom system name 'weaver' in the build name as:

```
$ . cmake/std/atdm/load-env.sh weaver-default
Selecting custom system configuration 'weaver'
Setting compiler and build options for build-name 'weaver-default'
Using weaver compiler stack GNU-7.2.0-OPENMPI-2.1.2 to build DEBUG code with Kokkos node type SERIAL and KOKKOS_ARCH=Power9
```

If you leave the name 'weaver' out of the build name then you get:

```
$ . cmake/std/atdm/load-env.sh default
Error, could not determine a system configuration for hostname='weaver11', aborting env loading script!
```

Instead, it would be desirable to use the registered custom configuration by default if no other known configurations were supported on the system.  So in the case of 'weaver', we want this:

```
$ . cmake/std/atdm/load-env.sh default
Selecting custom system configuration 'weaver'
Setting compiler and build options for build-name 'default'
Using weaver compiler stack GNU-7.2.0-OPENMPI-2.1.2 to build DEBUG code with Kokkos node type SERIAL and KOKKOS_ARCH=Power9
```

bartlettroscoe:
@ndellingwood, FYI, the PR that adds this is #8013.  Just need someone to review and approve to get merged.

bartlettroscoe:
@ndellingwood, PR #8013 just merged.  You should now be able to leave `wevear` out of the build name on 'weaver'.  I just pulled the tip of 'develop' (commit 008ee502fb9e8b844cbf8a758051d55fd87216da) and tested this with:

```
$ cd Trilinos/

$ git log-short -1
008ee50 "Merge Pull Request #8013 from bartlettroscoe/Trilinos/8008-atdm-custom-config-default"
Author: trilinos-autotester <trilinos@sandia.gov>
[...]

PR #8013, opened 2020-09-11
ATDM: Use registered custom configuration if no known configuratin matches (#8008)

bartlettroscoe:
This implements the use case described in #8008 and adds strong unit tests for `atdm/utils/get_system_info.sh` to boot.  I did not feel confident adding this feature without adding these new strong unit tests for existing behavior.

## How was this tested?

There is a strong unit test for this and I tested this on 'weaver' with:

```
$ export ATDM_CONFIG_REGISTER_CUSTOM_CONFIG_DIR=$PWD/cmake/std/atdm/contributed/weaver

Selecting custom system configuration 'weaver'
Using weaver compiler stack GNU-7.2.0-OPENMPI-2.1.2 to build DEBUG code with Kokkos node type SERIAL and KOKKOS_ARCH=Power9
```

Note above that I did **not** put the custom system name `weaver` in the build name `default` passed to `atdm/load-env.sh`.


cmake/ctest/drivers/atdm/README.md
```diff
@@ -498,24 +498,34 @@ The following `<system_name>` sub-directories exist (in alphabetical order):
...
 ATDM_KNOWN_SYSTEM_NAMES_LIST=(
   ...
   )
 ```
 
-Second, if the system selection is done by matching to the `hostname`, then
-add a new `elif` statement for that set of machines.  Note that more than one
-`hostname` machine may map to the same `<new_system_name>` (e.g. both `white`
-and `ride` machines map to the system env `ride`).
+Them, if the system selection is done by matching to the `hostname`, add a new
```
e10harvey, line 525: Them -> Then

e10harvey (approved):

PR #8018, opened 2020-09-14
Ifpack2 develop2

kyungjoo-kim:
## Motivation

This is to improve communication packing and unpacking of Block Tridiag container on CUDA using multiple streams.

TESTS: SPARC scaling, vortex, 32GPUs.  No negative performance effect.


csiefer2 (approved):
Approved.  No negative effect on a sparc scaling test on 32 GPUs on vortex.

PR #8021, opened 2020-09-14
Belos: Register SolverFactory for ScalarType = float

jennloe:
This PR modifies solver factory registration routines so that the solver factory can compile and run with scalarType = float.  

@trilinos/belos 

## Testing

We copy existing solver factory tests to also run with scalarType = float.


packages/belos/src/BelosSolverFactory_Belos.hpp
```diff
@@ -62,6 +62,14 @@ class BelosSolverFactory : public Impl::SolverFactoryParent<double,MultiVec<doub
     };
 };
 
+class BelosFloatSolverFactory : public Impl::SolverFactoryParent<float,MultiVec<float>,Operator<float>>
```
cgcgcg, line 65: Does this need to be guarded? The complex stuff has a `#ifdef HAVE_TEUCHOS_COMPLEX`.

jennloe: Possibly, but I am not sure what to guard it with.  When I wrapped the Factory.cpp test with `#ifdef Trilinos_ENABLE_FLOAT `, that test did not build, even though I had enabled that option.

jennloe: @hkthorn ??

hkthorn: I would say that "float" does not need to be guarded.  There are basic tests of LAPACK/BLAS/SDM in Teuchos that are based on float and are not guarded.  If the linear algebra stack supports float, then it should be fine.  Tpetra is controlling which scalar types it is enabling, like float, but this is not the only linear algebra that Belos can work with.

jennloe:
@cgcgcg This is failing on a MueLu test.  Can you look?

cgcgcg (approved):

PR #8026, opened 2020-09-14
ATDM: Add more tests, fix small typo

bartlettroscoe:
I forgot to push this commit that added some more tests (for failure cases).  I also fixed the typo found by @e10harvey noted in https://github.com/trilinos/Trilinos/pull/8013#pullrequestreview-487758447.

## How was this tested?

The new unit tests are the tests :-)


e10harvey (approved):
Thanks, Ross! We may want to use`${_ASSERT_EQUALS_}` everywhere for line numbers going forward.

bartlettroscoe:
> Thanks, Ross! We may want to use `${_ASSERT_EQUALS_}` everywhere for line numbers going forward.

@e10harvey, yes.  The trick is that you need to double quote the arguments with `'"<arg>"' `.

PR #8031, opened 2020-09-15
tpetra: workaround gcc/7.2.0 bug with c++14

ndellingwood:
Add work-around in unit test
tpetra/core/test/CrsMatrix/Equilibration.cpp
to resolve internal compiler errors of type
`internal compiler error: in maybe_undo_parenthesized_ref, at cp/semantics.c:1705`

@trilinos/tpetra 

## Motivation
Workaround compilation error with cuda/9.2+gcc/7.2 with c++14 support enabled

## Related Issues
This is an internal gcc compiler error that pops up with c++14 support, see for example
https://github.com/kokkos/kokkos-kernels/issues/349

## Testing
Tested with a contributed atdm script I am working (essentially the same as sems-rhel7 env instead of kokkos-dev-2)

```bash
export TRILINOS_DIR=${HOME}/Trilinos
export ATDM_CONFIG_REGISTER_CUSTOM_CONFIG_DIR=$TRILINOS_DIR/cmake/std/atdm/contributed/kokkos-dev-2
source ~/Trilinos/cmake/std/atdm/load-env.sh kokkos-dev-2-cuda-opt
#source ~/Trilinos/cmake/std/atdm/load-env.sh Trilinos-atdm-sems-rhel7-cuda-opt
 
cmake  \
  -GNinja  \
  -DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake  \
  -DTrilinos_ENABLE_TESTS=ON \
  -DTrilinos_ENABLE_Kokkos=ON \
  -DTrilinos_ENABLE_TpetraCore=ON \
  -DCMAKE_CXX_STANDARD=14 \
  $TRILINOS_DIR
```

### DIFF
--- a/packages/tpetra/core/test/CrsMatrix/Equilibration.cpp
+++ b/packages/tpetra/core/test/CrsMatrix/Equilibration.cpp
@@ -1395,11 +1395,13 @@ makeMatrixTestWithExplicitZeroDiag (Teuchos::FancyOStream& out,
     Kokkos::deep_copy (ptr_h, A_lcl.graph.row_map);
     auto ind_h = Kokkos::create_mirror_view (A_lcl.graph.entries);
     Kokkos::deep_copy (ind_h, A_lcl.graph.entries);
+    // Work-around gcc/7.2.0 bug
+    const int myRank_ = myRank;
 
     for (LO lclRow = 0; lclRow < lclNumRows; ++lclRow) {
       using size_type = typename decltype (A_lcl.graph)::size_type;
       for (size_type k = ptr_h[lclRow]; k < ptr_h[lclRow+1]; ++k) {
-        const val_type expectedUnscaledVal = myRank == 1 ? 0.0 : 1.0;
+        const val_type expectedUnscaledVal = myRank_ == 1 ? 0.0 : 1.0;
         const mag_type rowNorm = gblRowNorms[lclRow];
         const mag_type scalingFactor = assumeSymmetric ?
           KAM::sqrt (rowNorm) : rowNorm;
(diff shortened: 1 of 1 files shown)


kddevin (approved):
Whoa! This is twisted and sick.  (Assign a const int to a const int and use the new copy.)  
How did you find this solution?  
Is there some pattern we should avoid in the future to avoid this problem?
I don't really like the way the test uses the lambda function -- seems needlessly complicated -- but I'm grateful that you took the time to fix the problem for us, @ndellingwood .  Thank you!

Do you have time to add the PR number #8031 to the workaround comments in the code?  Always nice to have that cross-reference.

ndellingwood:
@kddevin I pushed an update for the comment to include link to this PR.

Here's the bug report for the gcc/7.2 issue: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=882855

Since it's an internal compiler error, I couldn't say this happens consistently enough to prescribe a pattern for avoidance, though I've seen it show up elsewhere in the past and was familiar. Sometimes when a const variable is captured by a lambda and used within a loop this occurs, and the workaround is to create a local copy of the const var in the lambda outside of the loop. It is pretty gross.

ndellingwood:
@kddevin looks like an updated review is required after my push to update the comments, though it looks like the autotester is restarting without the RETEST label.

kddevin (approved):

kddevin:
Thanks, @ndellingwood !  You saved us a lot of time hunting that down.

PR #7979, opened 2020-09-03
MueLu: Extend RegionRFactory to allow multiple DOFs per node

GrahamBenHarper:
@trilinos/muelu 

## Motivation
This allows for multiple variable equations, such as elasticity, to be solved.

## Testing
I reran the region unit tests on skyridge with a debug build and they pass.

## Additional Information
@lucbv I'd like you to look at the loops for a moment. The current access pattern definitely isn't optimal because the outer loop is `l`.
What I currently have looks like
```
      LO rowIdx = 0, columnOffset = 0, entryOffset = 0;
      for(LO l = 0; l < blkSize; ++l) {
        row_map_h(rowIdx + 1 + l) = entryOffset + cornerStencilLength*blkSize + l;
        for(LO k = 0; k < 3; ++k) {
          for(LO j = 0; j < 3; ++j) {
            for(LO i = 0; i < 3; ++i) {
              entries_h(entryOffset + (k*9 + j*3 + i)*blkSize + l) = columnOffset
                + (k*lFineNodesPerDim[1]*lFineNodesPerDim[0] + j*lFineNodesPerDim[0] + i)*blkSize + l;
              values_h(entryOffset + (k*9 + j*3 + i)*blkSize + l)  = coeffs[k + 2]*coeffs[j + 2]*coeffs[i + 2];
	    }
          }
        }
        for(int dim = 0; dim <numDimensions; ++dim) {
          coarseCoordData[dim][rowIdx + l] = fineCoordData[dim][columnOffset + l];
        }
      }
```
however, I'm wondering if the best approach to make sure the access patterns are optimal would look more like this:
```
      LO rowIdx = 0, columnOffset = 0, entryOffset = 0;
      for(LO k = 0; k < 3; ++k) {
        for(LO j = 0; j < 3; ++j) {
          for(LO i = 0; i < 3; ++i) {
            for(LO l = 0; l < blkSize; ++l) {
              entries_h(entryOffset + (k*9 + j*3 + i)*blkSize + l) = columnOffset
                + (k*lFineNodesPerDim[1]*lFineNodesPerDim[0] + j*lFineNodesPerDim[0] + i)*blkSize + l;
              values_h(entryOffset + (k*9 + j*3 + i)*blkSize + l)  = coeffs[k + 2]*coeffs[j + 2]*coeffs[i + 2];
	    }
          }
        }
      }
      for(LO l = 0; l < blkSize; ++l) {
[...]

PR #8038, opened 2020-09-16
MueLu: Improve RegionRFactory and Tests

GrahamBenHarper:
@trilinos/muelu 

## Motivation
#7979 accidentally broke some MueLu experimental code. This fixes a majority of the issues, but the linear solve for the region Elasticity3D problem still seems to act strange. You can see this when watching the verbose test output.

There may still be some functionality related to these tests that forgets the number of DOFs per node internally. Setting the Xpetra block size for R and P when they are computed in RegionRFactory helped it remember what A's block size should be on the next level, but this other strange behavior is still being investigated.

## Testing

This time I made sure to run the experimental tests in addition to the unit tests on a debug build, and the tests are not failing, which is an improvement.

### DIFF
--- a/packages/muelu/research/regionMG/test/structured/CMakeLists.txt
+++ b/packages/muelu/research/regionMG/test/structured/CMakeLists.txt
@@ -23,6 +23,7 @@ IF (${PACKAGE_NAME}_ENABLE_Tpetra AND ${PACKAGE_NAME}_ENABLE_Amesos2)
       structured_1dof-complex.xml
       structured_linear_1dof.xml
       structured_linear_R_1dof.xml
+      structured_linear_R_3dof.xml
       structured_linear_3dof.xml
       structured_unstructured_1dof.xml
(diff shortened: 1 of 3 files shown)


GrahamBenHarper:
Running all tests (instead of just the ones I thought were relevant) shows the following, which I'll have to look at closer to see if I caused any issues:
```
	 15 - MueLu_Structured_Unstructured_Region_Example_Poisson3D_27pt_stencil_MPI_4 (Timeout)
	 40 - MueLu_Structured_Unstructured_Laplace3D_Tpetra_MPI_4 (Timeout)
	 44 - MueLu_Structured_Region_Linear_Elasticity3D_coarseRepartition_repartitioning_Tpetra_MPI_4 (Failed)
```

lucbv (approved):
lgtm, I see that you are setting the dofs in R & P a bit differently than the usual way.
I don't think that it is a problem though

PR #8040, opened 2020-09-16
Tpetra:  remove Kokkos::parallel_scan workaround implemented for 2019 parallel_scan bug

kddevin:
@trilinos/tpetra 

## Motivation

This code existed as a workaround to a Kokkos::parallel_scan bug (the so-called "UVM bug" of 2019).
The bug is fixed so this workaround is no longer needed.

## Testing

cee-lan, gcc 7.3


jhux2 (approved):
yes!

Issue #7216, opened 2020-04-21
Amesos2:  Improve handling of refactor options


MicheldeMessieres:
## Enhancement
@trilinos/amesos2

Amesos2 solvers will test a refactor option (--refactor) which modifies the matrix.

This is working for the current test matrices but the modification is too severe for large matrices and they will no longer be positive definite. Code applied is:
`    for( size_t i = 0; i < l_fst_row_nnz; ++i ){
      values[i] = values[i] * values[i];
    }`
It would be preferable that any valid matrix can be replaced for the test matrix and still run refactor. The Cholmod solver is one case at least which has issues.

MicheldeMessieres:
Also SuperLUDist is failing refactor.

PR #8043, opened 2020-09-17
Amesos2: SuperLUDist fixes

MicheldeMessieres:
Use superludist config to get int_t size 32 or 64.
Fix superlu dist test to run for long long go.
Remove --refactor option which needs to be resolved and fails.
Modify SuperLuDist to use int_t so it works for size 32 or 64.

@ndellingwood I disabled the --refactor option. This is failing right now but not resolved yet. There was a prior error in the refactor setup causing it to test an all 0 vector. I fixed that a few months ago at which point SuperLUDist would have started failing. If it's ok to disable this fo now I can investigate as a separate issue. There are some other refactor issues in #7216 and I added a comment there on this.

@srajama1 FYI - These are the fixes from the recent user issue regarding SuperLUDist with size 64.

@trilinos/amesos2 

## Motivation

Resolve Amesos2 SuperLUFDist so it works with size 64 int_t.

## Testing

Amesos2 SuperLUDist test on mac parallel build.

### DIFF
--- a/packages/amesos2/src/Amesos2_Superludist_FunctionMap.hpp
+++ b/packages/amesos2/src/Amesos2_Superludist_FunctionMap.hpp
@@ -259,7 +259,7 @@ namespace Amesos2 {
      * This form operates on a SuperMatrix having the NRformat_loc
      */
     static void gsequ_loc(SLUD::SuperMatrix* A, double* r, double* c, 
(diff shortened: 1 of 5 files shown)


ndellingwood (approved):
Thanks @MicheldeMessieres !

PR #8045, opened 2020-09-17
Zoltan2: Refactor directory to use Teuchos comm

MicheldeMessieres:
@kddevin These are the conversions to Teuchos comm we discussed a while back. Most of this is just straight forward conversion from the MPI version to the Teuchos version.

Teuchos didn't have logical or (LOR) for reduce and I switched to bitwise (BOR). We're using int as 0 or 1 to represent bool logic in this case. I'm not sure it matters so much which we use but I could easily add LOR to Teuchos.

The only complication was Teuchos does not have waitAny. What I did as a short term fix was serialize that segment to just call wait on each request in order (would be some loss of performance) and left a note describing how we can convert this back to waitAny when available. I started looking at what Teuchos needs and it should be pretty easy to set up since we can just copy the waitAll method as a pattern for the setup. But there is a complication since the Teuchos request holds the MPI_Request internally. For waitAll, Teuchos releases all of them. For waitAny we want to release just the one that responds (but don't know this until after waitAny). Maybe we release all and restore all except the one we get. I tried a bit and had some bug with it so decided to leave it for a separate PR.

What I'm proposing is we accept the loss of performance for now so this PR doesn't spread too much. Then I or someone could work on adding the Teuchos::waitAny and restore that as a separate PR.

Other minor fixes: I switched to Tpetra::ScopeGuard and also cleaned up a hack I had in place so it runs on serial. I've preserved the serial tests, which we don't really need, just because it really helps with debugging. But it's in a much cleaner setup now.

There are some untested code blocks remaining and I cleaned up the throw comments I had left for that.

@trilinos/zoltan2  

## Motivation

Clean up comm commands to use Teuchos instead of raw MPI calls.

## Testing

Mac parallel and serial builds - Cuda on white - zoltan2 tests

### DIFF
--- a/packages/zoltan2/core/src/directory/Zoltan2_Directory.hpp
+++ b/packages/zoltan2/core/src/directory/Zoltan2_Directory.hpp
@@ -48,6 +48,7 @@
 #define ZOLTAN2_DIRECTORY_H_
 
 #include <Teuchos_DefaultComm.hpp> // currently using Teuchos comm throughout
+#include <Teuchos_CommHelpers.hpp>
 
 #ifndef HAVE_MPI
 // support mpi serial - directory currently has a mix of Teuchos mpi commands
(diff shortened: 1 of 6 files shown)


kddevin (approved):
LTGM  Thanks, @MicheldeMessieres

kddevin (approved):

PR #8046, opened 2020-09-17
Tpetra - change the logic in set all to scalar

kyungjoo-kim:
## Motivation

While reviewing deep copy in tpetra, I thought that this method does not have to be too complicated as previously. I also would like to discuss a couple of things

1. Do we really need to follow the last touch rule in dual view ? 
I believe that the dual view is for convenience to keep the data both places and communicate more easily. I do not think that using the host view of the dual view is for a performance reason. We probably may want to use devices as much as possible. However, the last touch rule makes the code complicated and it can result in using inefficient host kernels. As seen in #7985, we can get better performance when we select where to execute the kernel considering the problem size or other aspects. I also think that there is a reason that tpetra chose to follow the last touch rule which I do not know. How to you think @trilinos/tpetra ? 

This PR initialize both host and device views with a given scalar alpha. With async device kernels launch, both initialization can be done simultaneously and I am expecting that it does not cause overhead compared to the case that initialization happens either host or device.

## Related Issues

#7946 

## Testing

There is Tpetra unit test for this method and I changed the test with modified work scenario. This code passes on cuda.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
@@ -954,30 +954,27 @@ public:
   BlockCrsMatrix<Scalar, LO, GO, Node>::
   setAllToScalar (const Scalar& alpha)
   {
-#ifdef HAVE_TPETRA_DEBUG
-    const char prefix[] = "Tpetra::BlockCrsMatrix::setAllToScalar: ";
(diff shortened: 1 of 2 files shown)


kddevin (approved):
LGTM.  Thanks, Kyungjoo.
I don't know the history of doing updates on most recently updated.  I agree that the choice should be made for performance.  These changes make sense to me.

PR #8053, opened 2020-09-17
Minor tweak to van1-tx2 environment

dc-snl:
I would like to push a minor tweak to `cmake/std/atdm/van1-tx2/environment.sh` to load python/3.6.8-arm which provides a newer version of numpy that EMPIRE requires.  Previously, python/3.6.8 got loaded.


jmgate:
@e10harvey, was there a reason the environment grabbed `python/3.6.8-gnu7` by default instead of the arm version?

e10harvey:
> @e10harvey, was there a reason the environment grabbed `python/3.6.8-gnu7` by default instead of the arm version?

IIRC, the arm version wasn't available when this change was made. Thanks for bringing it to my attention.

jmgate:
Ah, gotcha.  Perfect.  I'm going to approve and auto-merge then.

jmgate (approved):

PR #8079, opened 2020-09-22
Lower case diff

bathmatt:
Was seeing an issue on restart with different cases 

@gdsjaar can you review?


gsjaardema (commented):
IOSS-related changes look good.  Will try to document this better to avoid long debugging sessions...

Issue #8048, opened 2020-09-17
Phalanx:  Build Errors with C++14 builds


ZUUL42:
## Bug Report
@trilinos/phalanx 

### Description
In updating the CUDA 9.2.88 build to C++14, I'm getting build errors from Phalanx.

We're trying to update these builds to C++14 before 09/28 when @trilinos/kokkos is going to merge in a snapshot that requires C++14.
Ties to #8016 [#6260](https://github.com/trilinos/Trilinos/issues/6260#issuecomment-672256473).

### Steps to Reproduce
[CDash showing failing builds](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=cuda9c14&field2=buildstamp&compare2=63&value2=Experimental&field3=buildstarttime&compare3=83&value3=2020%2F09%2F15)

rppawlo:
@ZUUL42 - the two phalanx errors on the intel build look like they are actually kokkos issues. The nox and panzer build errors are exactly the same. For example, here are two of the errors:

```
In file included from /scratch/trilinos/jenkins/ascic158/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/Trilinos/packages/teuchos/core/src/Teuchos_ScalarTraits.hpp(59),
                 from /scratch/trilinos/jenkins/ascic158/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/Trilinos/packages/teuchos/core/src/Teuchos_TestingHelpers.hpp(52),
                 from /scratch/trilinos/jenkins/ascic158/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/Trilinos/packages/teuchos/core/src/Teuchos_LocalTestingHelpers.hpp(56),
                 from /scratch/trilinos/jenkins/ascic158/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/Trilinos/packages/teuchos/core/src/Teuchos_UnitTestHarness.hpp(54),
                 from /scratch/trilinos/jenkins/ascic158/workspace/trilinos-folder/Trilinos_pullrequest_intel_17.0.1/Trilinos/packages/phalanx/example/FiniteElementAssembly/auxiliary_tests/Example_FEM_Workset_AuxTest.cpp(50):
[...]

PR #8086, opened 2020-09-23
Phalanx: fix for gcc 5/6 lambda bug on cuda

rppawlo:
## Motivation

c++14 builds are broken with intel that uses gcc 5/6 due to lambda capture bug.

Closes https://github.com/trilinos/Trilinos/issues/8048

## Testing

c++14 PR testing will cover this.


etphipp (approved):

etphipp:
Just out of curiosity, is this the same lambda-capture issue with const variables we have seen for a while now, or something else?  It sure would be nice to drop support for gcc 5/6.

rppawlo:
Yes - it's the same one we've been tracking for the last few years. Somehow enabling c++14 triggered it in a few lamdas that didn't see this error previously.

PR #8098, opened 2020-09-24
Seacas fix symmetric communication

gsjaardema:
@trilinos/seacas 

## Motivation

Fixes error in last seacas snapshot which resulted in non-symmetric ZGC (communication interfaces) between zones.  This fixes the issue and adds a `validate_symmetric_communication` function to the `cgns_decomp` program which will catch this in the future.  Verified that the validate function correctly found the 4 bad interfaces prior to the fix and found no problems after the fix.

### DIFF
--- a/packages/seacas/libraries/ioss/src/cgns/Iocgns_StructuredZoneData.C
+++ b/packages/seacas/libraries/ioss/src/cgns/Iocgns_StructuredZoneData.C
@@ -25,7 +25,7 @@ namespace {
 
   bool overlaps(const Range &a, const Range &b)
   {
-    return a.begin() <= b.end() && b.begin() <= a.end();
+    return a.m_beg <= b.m_end && b.m_beg <= a.m_end;
   }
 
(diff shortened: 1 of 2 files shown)


gsjaardema:
@bartlettroscoe @prwolfe Could one of you please review this.

Thanks.

bartlettroscoe (commented):
@gsjaardema, are there no tests that were impacted by this?  If there was a bug that this commit fixes, why were not tests added or updated to reflect the correct behavior?

Sorry I did not review this earlier (was off Friday and have been putting out fires this morning).

gsjaardema:
@bartlettroscoe No existing tests were affected; however, when I added the changes in the cgns_decomp.C file, then tests failed pre-change and passed post-change, so there are now tests that will fail if the behavior regresses.

I am also adding additional testing which will be committed later this week, but wanted to get this to the SPARC developers.

bartlettroscoe (approved):
Thanks for the testing explanation @gsjaardema.  Just an FYI, but in the future, you might consider testing in the "Testing" section of the PR to address questions like this.

PR #8099, opened 2020-09-24
Framework: PR Driver Overhaul (WIP)

william76:
This is a work in progress. I'll update description + documentation once I sort out some PR testing errors and am ready for code reviews.


william76:
I believe the current testing failure is because the test branch has no profile set up for the `gcc_7.2.0_debug` test.
This round has some additional cleanup so I want to get this test round through first and will work on adding a gcc 7.2.0 debug profile next.

PR #8102, opened 2020-09-25
Muelu: some changes in preparation of a new InterfaceAggregationFactory

mayrmt:
@trilinos/muelu 

## Motivation

While working on extending some `Blocked*` capabilities and the `InterfaceAggregationFactory`, I made some minor changes/cleanups/fixes/new tests along the way.

## Testing

Existing tests pass. Some new tests have been added.


packages/galeri/src-xpetra/Galeri_Elasticity3DProblem.hpp
```diff
@@ -460,11 +460,13 @@ namespace Galeri {
                                         hy = TST::magnitude(stretch[1]),
                                         hz = TST::magnitude(stretch[2]);
 
-      GO myPID = this->Map_->getComm()->getRank();
-      GO const & negOne=-1;
-      GO mx = this->list_.get("mx", negOne), my = this->list_.get("my", negOne), mz = this->list_.get("mz", negOne);
+      const GO myPID = this->Map_->getComm()->getRank();
+      const GO one = Teuchos::ScalarTraits<GO>::one();
+      const GO mx = this->list_.get("mx", one);
```
cgcgcg, line 465: In some of the Galeri problems, -1 is used to ask for same number in every dimension. Would a similar logic make sense here>

mayrmt: @cgcgcg This change now follows the implementation in https://github.com/trilinos/Trilinos/blob/94400d92f136eecc1fac88c9616f43fdb1eadf60/packages/galeri/src-xpetra/Galeri_Elasticity2DProblem.hpp#L423-424.

I would not set `-1` here, since the mentioned default mechanism is not present here. Next step is to compute `const GO mxy = mx * my;`, which could get negative, resulting in possibly weird effects in the subsequent modulo operations. 

I'm happy to revert this change, if we are not sure, that this is an improvement. Let me know, what you think.

cgcgcg: Right, my question is simply if it would make sense to add this mechanism here. Please do whatever you think is best / works best for you.

mayrmt: @cgcgcg Ok, now I understand. I'll look into it. Marking this PR as WIP meanwhile.

cgcgcg: I think you might need to do something like `mx=my=mz=pow(comm->getSize(),1/3);` and some correction so that `mx*my*mz==comm->getSize()`.

mayrmt: @cgcgcg I reverted that change. Better now mess with something, that's currently working, when I don't can clearly improve it.

cgcgcg (approved):

PR #8104, opened 2020-09-25
Tpetra: Allows FixedHashTable to build on CudaSpace

MicheldeMessieres:
@kddevin @csiefer2 Resolves FixedHashTable so it can construct on CudaSpace. Previously it assumed UVM. This is the repair we discussed earlier and it cleans up the setup in Map.

The next step would be to write a parallel kernel version of the search for the first non-continuous key. We may not want to merge this PR until that is resolved since my change can impact performance. But I think it's ok to treat that as a separate issue because even though I added a copy in FixedHashTable, I was able to remove the extra copy in Map.

I could start working on a parallel kernel version. I am wondering if a simple parallel for loop with atomic comparison checks for non-contiguous neighbors might actually do unexpectedly well, based on prior experience with Multijagged kernels.  That would be much simpler but I'm not sure yet exactly what kind of distributions we are dealing with.

Note there was another potential UVM read I did not fix because it's in a block for buildInParallel false, which would never be the case currently for Cuda. I changed those comments to reflect this. 

@trilinos/tpetra 

## Motivation

Map currently builds FixedHashTable on CudaUVMSpace because FixedHashTable had UVM dependencies. This change allows Map to directly work with CudaSpace and avoid an extra copy of the hash.

## Testing

white Cuda build

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
@@ -1088,6 +1088,9 @@ init (const keys_type& keys,
   // Count the number of "buckets" per offsets array (ptr) entry.
   //
 
+  // Will only create the mirror for buildInParallel false - but then use it in two places
+  typename keys_type::HostMirror theKeysHost;
+
   // The Kokkos kernel uses atomic update instructions to count the
   // number of "buckets" per offsets array (ptr) entry.  Atomic
   // updates incur overhead, even in the sequential case.  The Kokkos
@@ -1113,17 +1116,17 @@ init (const keys_type& keys,
     }
   }
   else {
-    // Access to counts is not necessarily contiguous, but is
-    // irregular and likely touches all pages of the array.  Thus, it
-    // probably makes sense to use a host copy explicitly, rather than
-    // assume UVM.
+    theKeysHost = Kokkos::create_mirror_view(theKeys);
+    Kokkos::deep_copy(theKeysHost, theKeys);
+
     auto countsHost = Kokkos::create_mirror_view (counts);
     // FIXME (mfh 28 Mar 2016) Does create_mirror_view zero-fill?
     Kokkos::deep_copy (countsHost, static_cast<offset_type> (0));
 
+    Kokkos::deep_copy(theKeysHost, theKeys);
     for (offset_type k = 0; k < theNumKeys; ++k) {
       using key_type = typename keys_type::non_const_value_type;
-      const key_type key = theKeys[k];
+      const key_type key = theKeysHost[k];
 
       using hash_value_type = typename hash_type::result_type;
       const hash_value_type hashVal = hash_type::hashFunc (key, size);
@@ -1219,9 +1222,15 @@ init (const keys_type& keys,
     Kokkos::parallel_reduce (range_type (0, theNumKeys), functor, result);
   }
   else {
+    auto counts_h = Kokkos::create_mirror_view(counts);
+    Kokkos::deep_copy(counts_h, counts);
+    auto ptr_h = Kokkos::create_mirror_view(ptr);
+    Kokkos::deep_copy(ptr_h, ptr);
+    auto val_h = Kokkos::create_mirror_view(val);
+    Kokkos::deep_copy(val_h, val);
     for (offset_type k = 0; k < theNumKeys; ++k) {
       typedef typename hash_type::result_type hash_value_type;
-      const KeyType key = theKeys[k];
+      const KeyType key = theKeysHost[k];
       if (key > result.maxKey_) {
         result.maxKey_ = key;
       }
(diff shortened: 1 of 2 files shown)


packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
```diff
@@ -1236,16 +1233,15 @@ init (const keys_type& keys,
 
       // Return the old count; decrement afterwards.
       //
-      // NOTE (mfh 28 Mar 2016) This assumes UVM.  It might make more
-      // sense to use a host mirror here, due to the access pattern.
+      // Assumes UVM but this code should not execute for Cuda since buildInParallel is true
```
kddevin, line 1236: correction:  buildInParallel is false

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
```diff
@@ -1236,16 +1233,15 @@ init (const keys_type& keys,
...
       const offset_type count = counts[hashVal];
       --counts[hashVal];
       if (count == 0) {
         result.success_ = false; // FAILURE!
         break;
       }
       else {
-        // NOTE (mfh 28 Mar 2016) This assumes UVM.
+        // Assumes UVM but this code should not execute for Cuda since buildInParallel is true
```
kddevin, line 1244: correction:  buildInParallel is false

kddevin (approved):
Thanks, @MicheldeMessieres 
I'll approve to get the code testing; the requested changes are minor comment fixes.

I'm a little worried about the worthBuildingFixedHashTableInParallel logic.  If someone were to change the condition as the comments suggest (i.e., say we need 3000-way parallelism to be "worth" it), the assumed UVM code currently guarded by "buildInParallel==false" might be run, leading to invalid accesses.

I'm wondering if the buildInParallel logic is really worthwhile.  Is there *that* much overhead in the parallel kernels when run in serial?  We could eliminate a lot of code (including these worrisome UVM assumptions) if we always did buildInParallel.  Do you have a sense of how important this special logic is?

A parallel kernel for the search (a la Zoltan2's MJ) is a good next step.  Thank you!

MicheldeMessieres:
@kddevin  I don't understand the conditions where it's running serial and those performance differences would be critical. Otherwise I'd agree having one path is ideal.

I could fix the code chunk which uses UVM if buildInParallel is false. I think we just need mirrors and copies to and from host. I wasn't sure if that would be a good idea to do now if it's not actually being used. Please let me know if you'd prefer I'd do this now.

I will fix the comments and resolve Xpetra, which had a problem with this PR.

MicheldeMessieres:
@kddevin I see now the Xpetra was just ranks mangling output and the test was ok - also I see you already requested AT RETEST. Thanks.

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
MicheldeMessieres: Fixed

packages/tpetra/core/src/Tpetra_Details_FixedHashTable_def.hpp
MicheldeMessieres: Fixed

MicheldeMessieres:
@kddevin It seems this PR puts that Xpetra test in a sweet spot for the output failure. I can fix that next week. I'll add the fix to this PR.

kddevin:
I confirmed that the buildInParallel is worthwhile.  With Serial node on a cee lan workstation, running with the Kokkos parallel loops and atomics (buildInParallel = true) increased runtime by about 33% compared to the special-case serial logic (buildInParallel = false).   So it is worth keeping the special-case serial logic.

MicheldeMessieres:
@kddevin Thanks. Do you want me to the fix the code for buildInParallel=false to work for Cuda now? It will add mirrors and copies which should not impact the current performance at all since that won't ever run for CUDA and the copies won't do do anything. If someone did change it, I expect they might want to evaluate the performance at that point depending on the conditions. I could leave a comment regarding this.

MicheldeMessieres:
@kddevin I fixed the Xpetra test fail due to mangled output (#8117). This PR should pass now but will need a new review since I updated the errors in the comments.

If we decide we want to fix the code which is not UVM safe (and not currently used for Cuda builds) I could add that to this PR or make a new PR today.

kddevin:
Thanks, @michel.  
Yes, let's fix the UVM assumption rather than get burned by it later.  If you have time to do it, that would be great.  Thanks!

MicheldeMessieres:
@kddevin This is ready for review. I confirmed with buildInParallel false on white's Cuda build. This adds several host mirror copies which should have no performance impact since it currently only runs for non-Cuda builds and the copies will do nothing.

I didn't have time to work on a kernel for finding the first non-contiguous key index. I suggest a simple parallel_for with atomic comparisons may do surprisingly well on modern cards and perhaps do much better than a reduce algorithm. If we change the buildInParallel rules we will likely want to implement some kernels for all of these sections.

kddevin (approved):
LGTM.   Thanks, @MicheldeMessieres

PR #8113, opened 2020-09-28
Tempus: Identify usage of FSAL

### LINKED ISSUE #8105: Tempus: Identify usage of FSAL
## Enhancement
@trilinos/tempus 

Identify a Stepper as able to use FSAL (useFSAL=true), not use FSAL (useFSAL=false), or both.

ccober6:
* Identify a Stepper as able to use FSAL (useFSAL=true), not use
   FSAL (useFSAL=false), or both, through the set functions
   setUseFSALTrueOnly(), setUseFSALFalseOnly(), or setUseFSAL(),
   respectively.
 * There are just two Steppers that only use useFSAL=true, Trapezoidal
   and NewmarkImplicitAForm.  This is primarily due to wanting to only
   have one ODE evaluation per time step.
 * Most Steppers have useFSAL=false as they do not meet the
   requirements for FSAL.  If the application tries to set
   it to true, a warning is given but will remain useFSAL=false.
 * Currently we have 6 Steppers that can set useFSAL to either
   true or false:
    - ForwardEuler
    - NewmarkExplicitAForm
    - RK Forward Euler
    - Bogacki-Shampine
    - EDIRK 2-Stage Theta
    - RK Trapezoidal Rule

   There are two advantages for useFSAL=true:
    - x and xDot are at the same time level at the end of the timestep
    - Save an evaulation of the ODE
   However this does require the ICs to be consistent, i.e., x and
   xDot satisfy xDot_0 = f(x_0,t=0).  By default these Steppers have
   useFSAL=true.  This is why these Steppers have the IC consistency
   turned on by default (ICConsistency = "Consistent").
 * OperatorSplit should, in general, set useFSAL=false as each operator
   will change xDot, but there can be situations where useFSAL=true could
   be useful.  So it can be set to either true or false.
 * Removed getUseFSALDefault(), getICConsistencyDefault() and
   getICConsistencyCheckDefault().  Defaults are now set explicitly
   in the Stepper constructors.
 * Fixed a few situations where FSAL should not be used if the previous
   time step failed.
 * Changed and added tests to explicitly test (previous testing was
   "hidden").
   - useFSAL = true and false
   - ICConsistency = "Consistent" and ICConsistencyCheck = true

Miscellaneous
 * Removed tabs from files.

@trilinos/tempus 

## Motivation
Better identify each Steppers FSAL capabilities.

* Closes #8105 

## Stakeholder Feedback
This was a request from @ikalash. 

## Testing
All current tests passes, and additional tests added to cover FSAL settings.

### DIFF
--- a/packages/tempus/src/Tempus_Integrator.hpp
+++ b/packages/tempus/src/Tempus_Integrator.hpp
@@ -79,7 +79,7 @@ public:
     /// Get the stepper
     virtual Teuchos::RCP<Stepper<Scalar> > getStepper() const = 0;
     /// Return a copy of the Tempus ParameterList
-    virtual Teuchos::RCP<Teuchos::ParameterList> getTempusParameterList()	= 0;
+    virtual Teuchos::RCP<Teuchos::ParameterList> getTempusParameterList() = 0;
     virtual void setTempusParameterList(Teuchos::RCP<Teuchos::ParameterList> pl) = 0;
     /// Returns the SolutionHistory for this Integrator
     virtual Teuchos::RCP<const SolutionHistory<Scalar> > getSolutionHistory() const = 0;
--- a/packages/tempus/src/Tempus_StepperBDF2_decl.hpp
+++ b/packages/tempus/src/Tempus_StepperBDF2_decl.hpp
@@ -43,7 +43,7 @@ namespace Tempus {
  *                -  \frac{\tau_n}{\tau_n + \tau_{n-1}}
  *                   \left[ \frac{x_{n-1}-x_{n-2}}{\tau_{n-1}}\right], \f$
  *
- *  The First-Same-As-Last (FSAL) principle is not needed BDF2.
+ *  The First-Same-As-Last (FSAL) principle is not needed for BDF2.
  *  The default is to set useFSAL=false, however useFSAL=true will also work
  *  but have no affect (i.e., no-op).
  *
@@ -136,7 +136,6 @@ public:
       {return isExplicit() and isImplicit();}
     virtual bool isOneStepMethod()   const {return false;}
     virtual bool isMultiStepMethod() const {return !isOneStepMethod();}
-
     virtual OrderODE getOrderODE()   const {return FIRST_ORDER_ODE;}
   //@}
 
@@ -151,7 +150,6 @@ public:
   virtual void computeStartUp(
     const Teuchos::RCP<SolutionHistory<Scalar> >& solutionHistory);
 
-  virtual bool getICConsistencyCheckDefault() const { return false; }
   Teuchos::RCP<const Teuchos::ParameterList> getValidParameters() const;
 
   /// \name Overridden from Teuchos::Describable
--- a/packages/tempus/src/Tempus_StepperBDF2_impl.hpp
+++ b/packages/tempus/src/Tempus_StepperBDF2_impl.hpp
@@ -26,9 +26,9 @@ template<class Scalar>
 StepperBDF2<Scalar>::StepperBDF2()
 {
   this->setStepperType(        "BDF2");
-  this->setUseFSAL(            this->getUseFSALDefault());
-  this->setICConsistency(      this->getICConsistencyDefault());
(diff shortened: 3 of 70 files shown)


ccober6:
@ikalash @michael-a-hansen @sconde Theses changes could effect Albany, SPARC, Drekar and EMPIRE.  Before committing them could you check to see if there is any issues with it.  I will check EMPIRE.  I will turn on the autotester to get results and WIP to make sure it does not get through until we are ready.

ccober6:
@sconde you can hold off on the review as there is a failure I can not reproduce at the moment on my Mac.  I will likely need to commit a fix and then get it reviewed... :)

sconde (approved):

ccober6:
So EMPIRE is ready for this PR.

@ikalash @michael-a-hansen @sconde Have you had a chance to test this?

sconde:
Looks good on my end.

michael-a-hansen:
Looks good on the SPARC side. Thanks for the notification!

ccober6:
@michael-a-hansen Thanks for the confirmation that everything is good on the SPARC side!

ccober6:
@ikalash Sorry to pester, but will you have a chance to check this branch?  Thanks.

sconde:
So I build Piro and ran the tests during my review. My understanding is that Albany uses Tempus through Piro.

ikalash:
@ccober6 : Yes, absolutely.  I was interested in the PR when I saw it opened per our earlier discussion about FSAL, but I was on vacation at the time.  I am still getting caught up, but I will look at this and test Albany this week.

ccober6:
@ikalash Oh, sorry.  Hope you had a great vacation!  I guess it is a down side to communicating through the PR; we do not see any out of office notices.  Thanks for the update, and let us know when you have tested this.  Thanks.

ikalash:
I just looked at this PR and tested Albany.  Everything is clean in Albany.  It looks good except I have a qualm with one thing: 

> * Most Steppers have useFSAL=false as they do not meet the
>   requirements for FSAL.  If the application tries to set
>   it to true, a warning is given but will remain useFSAL=false.

We talked about this already @ccober6 , I know.  I would say the right thing to do is if a user tries to set useFSAL=false and the stepper does not support that, the code should throw not print a warning.  The reasoning is as follows.  If a user is setting useFSAL=false, it is likely intentional, as in general users will just use whatever is default, not knowing what the option means.  A lot of users don't really look at warnings in their output, so they may think useFSAL=false is being used when it is not.  Just my two cents.  Perhaps other Tempus folks can weigh in, e.g., @sconde ?

ccober6:
Yeah, I know we talked about this, and I have been on the fence in situations like this.  I believe you would suggest instead of a warning message, you should suggest error'ing out.  Most users have suggested in situations like this, they want the code to continue unless it is a fatal error.  The use case goes ... My job has been sitting in the queue days/weeks and it error'ed out on input with a value that does not effect the solution.  They do not want this to happen.

So for the case of useFSAL, whether you use useFSAL=true or useFSAL=false, the solution will be exactly the same, so long as the ICs are consistent (i.e., x and xDot satisfy xDot_0 = f(x_0,t_0) or f(x_0,xDot_0,t_0)).  The two main advantages of useFSAL=true are
 1. Can save a function evaluation each timestep.
 1. x and xDot will be at the same time level at the end of the timestep.
The first is a computational savings, and the latter may effect diagnostics (so there is some question if this effects the application's usage of the solution). 

Another piece I tried to factor in here is some applications additional do not want even warning/informational messages printed from TPLs.  I could not quite go that far.  

Hopefully, this explains the current reasoning.  Let us know if you have any suggestions.

ikalash:
@ccober6 : I see.  I guess I was thinking of the Albany use case where one of the options doesn't work properly with our BC implementation, but that is on us.

ccober6:
Yeah, getting back in the mind frame of your situation, where the BCs are being "set" within the ModelEvaluator and is dependent on what the currentState->getXDot() is, can cause issues.  I think you are right that Tempus will not have this knowledge, and the applications need to cover that situation.

PR #8116, opened 2020-09-28
Tempus: update documentation for TSC

### LINKED ISSUE #2228: Tempus: update TSC documention for TSCS ordering issue
Using Time Step Control Strategy (TSCS) allows for applications to define their very own time step control strategy and add it to a vector of strategies and the Time Step Control will iterate and call each strategy. Essentially, this could be an `and` case if each strategies does a `min` with the step size `dt = metaData->getDt()` and a new computed step size `dt_new` before updating the step size with `metaData->setDt( DT )`.

If the application wanted an `or` case, they can just appropriately update the `metaData->setDt(dt)`. The TSC, after calling all the TSCS, gets the updated `dt` by calling `metaData->getDt()`.

Note the ordering in the `Time Step Control Strategy List` (is important). A final TSCS that is called last can get just negate all previous TSCS updates.

The documentation needs to be updated to make this clear.

@trilinos/tempus

sconde:
@trilinos/tempus 

## Motivation

Update documentation

## Related Issues

* Closes #2228

### DIFF
--- a/packages/tempus/src/Tempus_TimeStepControlStrategyComposite.hpp
+++ b/packages/tempus/src/Tempus_TimeStepControlStrategyComposite.hpp
@@ -17,8 +17,23 @@ namespace Tempus {
 
 template<class Scalar> class TimeStepControl;
 
-/** \brief StepControlStrategy class for TimeStepControl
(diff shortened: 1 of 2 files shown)


ccober6 (approved):
Looks good!

PR #8122, opened 2020-09-29
Teuchos: Add Git SHA as Watchr metadata

brian-kelley:
This will show up in the mouseover for each datapoint, making it much
easier to match performance changes with code changes.

This is used by setting environment variable ``TRILINOS_GIT_SHA`` to `` `git rev-parse HEAD` `` in the testing scripts.
This is optional - if the variable is not set, the XML files just won't contain this metadata.

@trilinos/teuchos

## Motivation

This will increase the usability of the Jenkins performance dashboards. Right now, if there's a big swing in some measurement, you have to manually look at the date/time the build ran and find it in the git history. With this change, the chart will already have that info (e.g. for bisecting, datapoints N-1 and N will give the start and end).

## Stakeholder Feedback

This idea came from Tpetra meetings where we were trying to figure out which code changes caused big performance changes.
## Testing

Tested by doing a PERFORMANCE category build and run on my workstation, with and without setting TRILINOS_GIT_SHA. Elliott Ridgway (who just added this feature to Watchr) confirmed that the metadata line looks correct. This won't start showing up on the dashboard until Jenkins updates its plugins.


kddevin (approved):
Awesome!  Thanks, @brian-kelley

PR #8123, opened 2020-09-29
Intrepid2 - uvm removal on test

kyungjoo-kim:
## Motivation

We do not like uvm. This is a first test modification from using Kokkos::Cuda to Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>. There are lots of places to change the tests where they assume uvm. This seems to take more time. 

## Testing

Tested on weaver.

### DIFF
--- a/packages/intrepid2/unit-test/Shared/ArrayTools/Cuda/test_02.cpp
+++ b/packages/intrepid2/unit-test/Shared/ArrayTools/Cuda/test_02.cpp
@@ -54,7 +54,8 @@ int main(int argc, char *argv[]) {
   const bool verbose = (argc-1) > 0;
   Kokkos::initialize();
 
-  const int r_val = Intrepid2::Test::ArrayTools_Test02<double,Kokkos::Cuda>(verbose);
+  const int r_val = Intrepid2::Test::ArrayTools_Test02
+    <double,Kokkos::Cuda/* Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace>*/ >(verbose);
   
   Kokkos::finalize();
   return r_val;
--- a/packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
+++ b/packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
@@ -75,7 +75,7 @@ namespace Intrepid2 {
       };                                                                \
     }
     
-    template<typename ValueType, typename DeviceSpaceType>
+    template<typename ValueType, typename DeviceType>
     int ArrayTools_Test01(const bool verbose) {
       
       typedef ValueType value_type;
@@ -91,11 +91,11 @@ namespace Intrepid2 {
       Teuchos::oblackholestream oldFormatState;
       oldFormatState.copyfmt(std::cout);
 
-      typedef typename
-        Kokkos::Impl::is_space<DeviceSpaceType>::host_mirror_space::execution_space HostSpaceType ;
+      using DeviceExecSpaceType = typename DeviceType::execution_space;
+      using HostExecSpaceType = Kokkos::DefaultExecutionSpace;
 
-      *outStream << "DeviceSpace::  "; DeviceSpaceType::print_configuration(std::cout, false);
-      *outStream << "HostSpace::    ";   HostSpaceType::print_configuration(std::cout, false);
+      *outStream << "DeviceSpace::  "; DeviceExecSpaceType::print_configuration(std::cout, false);
+      *outStream << "HostSpace::    ";   HostExecSpaceType::print_configuration(std::cout, false);
 
       *outStream      \
         << "===============================================================================\n" \
@@ -112,9 +112,9 @@ namespace Intrepid2 {
         << "|                                                                             |\n" \
         << "===============================================================================\n";      
       
-      typedef RealSpaceTools<DeviceSpaceType> rst;
-      typedef ArrayTools<DeviceSpaceType> art; 
-      typedef Kokkos::DynRankView<value_type,DeviceSpaceType> DynRankView;
+      typedef RealSpaceTools<DeviceType> rst;
+      typedef ArrayTools<DeviceType> art; 
+      typedef Kokkos::DynRankView<value_type,DeviceType> DynRankView;
 
 #define ConstructWithLabel(obj, ...) obj(#obj, __VA_ARGS__)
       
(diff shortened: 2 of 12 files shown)


packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
```diff
@@ -91,11 +91,11 @@ namespace Intrepid2 {
       Teuchos::oblackholestream oldFormatState;
       oldFormatState.copyfmt(std::cout);
 
-      typedef typename
-        Kokkos::Impl::is_space<DeviceSpaceType>::host_mirror_space::execution_space HostSpaceType ;
+      using DeviceExecSpaceType = typename DeviceType::execution_space;
+      using HostExecSpaceType = Kokkos::DefaultExecutionSpace;
```
CamelliaDPG, line 95: @kyungjoo-kim, isn't the device the default exec space for a Cuda build?  It seems confusing to define a typedef that seems to say that this is the HostExecSpaceType...

packages/intrepid2/unit-test/Shared/ArrayTools/test_02.hpp
```diff
@@ -91,11 +91,11 @@ namespace Intrepid2 {
       Teuchos::oblackholestream oldFormatState;
       oldFormatState.copyfmt(std::cout);
 
-      typedef typename
-        Kokkos::Impl::is_space<DeviceSpaceType>::host_mirror_space::execution_space HostSpaceType ;
+      using DeviceExecSpaceType = typename DeviceType::execution_space;
+      using HostExecSpaceType = Kokkos::DefaultExecutionSpace;
```
CamelliaDPG, line 95: Same thing here with the HostExecSpaceType.

packages/intrepid2/unit-test/Shared/ArrayTools/test_03.hpp
```diff
@@ -91,11 +91,11 @@ namespace Intrepid2 {
...
 
-      typedef typename
-        Kokkos::Impl::is_space<DeviceSpaceType>::host_mirror_space::execution_space HostSpaceType ;
+      using DeviceExecSpaceType = typename DeviceType::execution_space;
+      using HostExecSpaceType = Kokkos::DefaultExecutionSpace;
 
-      *outStream << "DeviceSpace::  "; DeviceSpaceType::print_configuration(std::cout, false);
-      *outStream << "HostSpace::    ";   HostSpaceType::print_configuration(std::cout, false);
+      *outStream << "DeviceSpace::  "; DeviceExecSpaceType::print_configuration(std::cout, false);
+      *outStream << "HostSpace::    ";   HostExecSpaceType::print_configuration(std::cout, false);
```
CamelliaDPG, line 98: Here, too.

packages/intrepid2/unit-test/Shared/ArrayTools/Cuda/test_02.cpp
```diff
@@ -54,7 +54,8 @@ int main(int argc, char *argv[]) {
   const bool verbose = (argc-1) > 0;
   Kokkos::initialize();
 
-  const int r_val = Intrepid2::Test::ArrayTools_Test02<double,Kokkos::Cuda>(verbose);
+  const int r_val = Intrepid2::Test::ArrayTools_Test02
+    <double,Kokkos::Cuda/* Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace>*/ >(verbose);
```
CamelliaDPG, line 58: @kyungjoo-kim, why do you do something different here than in Cuda/test_01.cpp?  Is it possible to be consistent with that as well as test_03.cpp?

CamelliaDPG (changes requested):
@kyungjoo-kim, aside from those few small changes I suggested, this looks good to me.

packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
kyungjoo-kim: The default execution space has uvm memory as its default memory space. To test the code without the uvm, we need ``Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace>`` and execution space is derived fromthe device type.

packages/intrepid2/unit-test/Shared/ArrayTools/Cuda/test_02.cpp
kyungjoo-kim: I planned to work all three tests and there are so many things to fix. I was tired of changing things. So, I just fix test 01. I am going to separate this work into multiple PRs. Once I found a pattern, I may ask you and mauro to help convert the tests.

packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
kyungjoo-kim: Ah.... you are right. It should be DefaultHostExecutionSpace. I misunderstand your comments.

kyungjoo-kim:
@CamelliaDPG All fixed.

packages/intrepid2/unit-test/Shared/ArrayTools/Cuda/test_02.cpp
CamelliaDPG: Got it.  Would you please add a comment here indicating that the commented-out bit is what we intend to do once test_02 is fixed?

packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
CamelliaDPG: Looking a bit more closely at this, you don't actually use the HostExecSpaceType anywhere except to print the configuration.  I would suggest getting rid of this typedef, since it suggests (to me anyway) that we'll be using it somewhere in the code below.  I would also change the text in the configuration output to say simply "DefaultExecutionSpace" instead of saying anything about the host.  Presumably the configuration output will include the fact that UVM is or is not enabled.

packages/intrepid2/unit-test/Shared/ArrayTools/test_01.hpp
```diff
@@ -91,11 +91,11 @@ namespace Intrepid2 {
       Teuchos::oblackholestream oldFormatState;
       oldFormatState.copyfmt(std::cout);
 
-      typedef typename
-        Kokkos::Impl::is_space<DeviceSpaceType>::host_mirror_space::execution_space HostSpaceType ;
+      using DeviceExecSpaceType = typename DeviceType::execution_space;
+      using HostExecSpaceType = Kokkos::DefaultHostExecutionSpace;
```
CamelliaDPG, line 95: Oh, now I understand what you were saying (I hadn't seen your latest).  Yes, this is fine now.

CamelliaDPG (approved):
@kyungjoo-kim Looks good to me now.  Thanks!

PR #8126, opened 2020-09-30
LOCA: adding tpetra householder bordered solver

rppawlo:
## Motivation

Adding Tpetra version of Householder Bordered solver to NOX/LOCA for Charon2. 

## Testing

Multiple unit/system tests were added for this capability in the nox/test/tpetra directory. Tests are:
- tTpetra_1DFEM_ME_UnitTests.cpp
- tTpetra_ConstraintModelEvaluator.cpp
- tTpetra_HouseholderBorderedSolve.cpp


packages/nox/src-loca/src-thyra/LOCA_Thyra_Group.C
```diff
@@ -289,10 +317,11 @@ LOCA::Thyra::Group::computeDfDpMulti(const std::vector<int>& paramIDs,
   in_args_.set_x(x_vec_->getThyraRCPVector().assert_not_null());
   if (in_args_.supports(::Thyra::ModelEvaluatorBase::IN_ARG_x_dot))
     in_args_.set_x_dot(x_dot_vec);
-  in_args_.set_p(param_index, param_thyra_vec);
+  for (size_t i=0; i < param_index.size(); ++i)
+    in_args_.set_p(param_index[i], param_thyra_vec[i]);
   if (!isValidF)
     out_args_.set_f(f.getThyraRCPVector().assert_not_null());
-  out_args_.set_DfDp(param_index, deriv);
+  out_args_.set_DfDp(param_index[0], deriv);
```
etphipp, line 324: I think this block of code isn't quite right in the case where we have multiple parameter vectors.  In that case I think we would need to make a DfDp out-arg for each parameter vector, not just `param_index[0]`.  However I don't think this code is ever called because of the `implement_dfdp = false` above.  I'm not sure if we want to change that or not.  If we leave it as-is, LOCA will always do finite differencing for the df/dp, even if the model evaluator supports computing df/dp.

etphipp (approved):
Thanks for doing this Roger!

packages/nox/src-loca/src-thyra/LOCA_Thyra_Group.C
rppawlo: @etphipp - yes - that's exactly it. This block of code has been disabled for quite some time now. I had to make the change above to get it to compile since param_index was changed from int to std::vector<int>. I did not spend any time trying to fix it yet. If it's ok, I'd like to leave this as is for now and revisit if we see convergence issues in Charon.

etphipp: Sounds good, thanks!

rppawlo:
@etphipp - can you reapprove? Tests were failing due to RCP circular dependency check at the end of the run. Only change was to add this line to the end of the Householder unit test to get it passing:
```
LOCA::destroyGlobalData(global_data);
```

etphipp (approved):

Issue #8131, opened 2020-09-30
MueLu: region multigrid code has unguarded use of Zoltan2


jhux2:
## Bug Report
@trilinos/muelu 

### Description
As seen in https://testing.sandia.gov/cdash/viewBuildError.php?buildid=7988718, the region multigrid code is failing to compile in serial due to unguarded use of Zoltan2, which requires MPI.

jhux2:
@pohm01 Thanks, your patch has fixed the compile error.  There are a few test failures (see https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=8001795).  In any case, I'll close this issue as fixed.

@lucbv @GrahamBenHarper

PR #8133, opened 2020-10-01
MueLu: Fix to guard region coarse rebalance in serial

pohm01:
Also a fix to rebalancing the coords for the coarse level solve that was causing some regionR tests to not work with rebalancing.

@trilinos/muelu 

## Motivation
Addresses Issue #8131 

## Testing
It compiles in serial now.


cgcgcg:
Related question: How is the number of coarse ranks picked? I would have expected to see a call to `RepartitionHeuristicFactory` somewhere.

pohm01:
> 
> 
> Related question: How is the number of coarse ranks picked? I would have expected to see a call to `RepartitionHeuristicFactory` somewhere.

Currently it is set manually in the driver. Using the heuristic factory is something I plan to add later depending on the performance benefits of the coarse level rebalancing.

jhux2 (approved):
Lgtm.

Issue #7383, opened 2020-05-15
Framework:  Trilinos Installation Testing


william76:
@trilinos/framework 

This is a top-level tracking issue for Trilinos installation testing. This topic has been posted a few times before but it keeps getting down prioritized... but we also haven't really ever broken the overall task into sub-tasks that may be easier to work on and track.

Tasks
=====
This is an initial broad-brush list of tasks that I think we probably need for this. It will surely evolve over time.

- [ ] [Add installation testing support to CMake](#Add-Installation-Testing-Support-to-CMake)
- [ ] [Modify PR driver scripts](#Modify-PR-Driver-Scripts)
    - [ ] Enable "Install Trilinos" sequence.
    - [ ] Enable "Test Installed Trilinos" sequence.
- [ ] [Develop Jenkins jobs to Perform Installation Tests](#Develop-Jenkins-jobs-to-Perform-Installation-Tests)
- [ ] [Verify Proper CDash Reporting](#Verify-Proper-CDash-Reporting)
- [ ] [Add Packages to Installation Testing](#Add-Packages-to-Installation-Testing)

Details
=======

Add Installation Testing Support to CMake
------------------------------------------
Most of the infrastructure should already be in place. See https://github.com/TriBITSPub/TriBITS/tree/5027-enable-installation-testing for more information on this. With luck, we just need to verify that installation testing works w/in TriBiTS.

We may need to re-learn what we need to do to use this and there may be some modifications necessary in the CMakeLists.txt files in Trilinos packages to make everything work properly.

Modify PR Driver Scripts
------------------------
There are two sequences that we will generally need to support:

### Install Trilinos
1. Configure without tests
2. Build Trilinos
3. Install Trilinos

### Test Installed Trilinos
1. Configure with tests against installed Trilinos
2. Build Tests
3. Run Tests

Develop Jenkins jobs to Perform Installation Tests
----------------------------------------------------
We'll need to create the necessary Jenkins infrastructure to run these test 
sequences. Without getting too far down into the weeds, I see two main ways we might implement this on our Jenkins server...

### Single Jenkins Job w/ 2 Stages
This may be the simplest one to do from a Jenkins perspective since everything
is in a single self-contained Jenkins job. 

### Two Jenkins Jobs, triggered in sequence
This will be more complicated because:
1. We can't guarantee that two jobs would run on the same node in the build farm using a `trilinos-any` tag. 
[...]

Issue #5027, opened 2019-04-29
Framework: Installation Testing


william76:
## Enhancement
@trilinos/framework 
@bartlettroscoe 
@jwillenbring 

As part of the roadmap towards implementing a backwards compatibility test to the Trilinos testing suite we need to first get Installation Testing up and running.

This will require some updates to [TriBiTS][1] to support some of the installation testing capabilities that we will need.  @bartlettroscoe can fill in more details on what we're needing to do there.

[1]: https://tribits.org/

bartlettroscoe:
@william76, I should have time to look into this later this week.

bartlettroscoe:
FYI: Getting bit by not having installation testing as part of PR testing (i.e. #5098).

I will try to work on the TriBITS stuff for this this week.

bartlettroscoe:
@william76 and @jwillenbring,

So it turns out the installation testing feature in TriBITS still works!  The updated branch:

* https://github.com/TriBITSPub/TriBITS/tree/5027-enable-installation-testing

as of commit TriBITSPub/TriBITS@58ed6f999c9dd915d03e433d6b767b6cabcaff3a has the working test and draft documentation.

@william76, I will post a PR for this in the TriBITS repo.  After that, can you please test this this branch out with Trilinos and review the test and documentation?  We can go over this when I get back from my trip.

bartlettroscoe:
@william76, I  invited you to be a TriBITS repo contributor.  Once you accept that, you should be able to review and approve the PR TriBITSPub/TriBITS#289.

You can then test Trilinos with this by cloning the TriBITS GitHub repo under your local Trilinos git repo and add:

```
   -DTrilinos_TRIBITS_DIR:STRING=TriBITS/tribits
```

to your `cmake` configure line.

Once you confirm it works as it should, I will clean up that TriBITS branch, merge it to the TriBITS 'master' branch and then snapshot to Trilinos and create a Trilinos PR to update TriBITS in Trilinos.

bartlettroscoe:
@william76,

Any chance to review and approve https://github.com/TriBITSPub/TriBITS/pull/289 yet? See [above](https://github.com/trilinos/Trilinos/issues/5027#issuecomment-495730314).

bartlettroscoe:
[...]

PR #8142, opened 2020-10-03
Framework: PR test driver rework

william76:
@trilinos/framework 

## Motivation
Rework on the Python script that drives PR tests.  The main things this rework does are:

- Modularization pass of the existing script to break things up into packages that can be self-contained for easier maintenance.
- Moves configuration specs into a .ini file instead of in-line in code.
- Provides more flexibility for future build types
- This is also a preliminary step towards installation testing because that kind of test relies on the ability to do something different from a standard test.

## Related Issues
* Precedes #7383
* Part of #7383
* Related to #5027

## Testing
- Implements just under 100 tests via pytest for nightly testing.
- I tested this out on another PR + manual Jenkins runs.  See #8099 for the latest test PR.  Since that one got pretty busy I am starting a fresh PR since it passed a run.  

## Additional Information
Since this branch passed a PR run on #8099, it's probably time to do a cleanup pass and start thinking code review with the @trilinos/framework team.


cmake/std/PullRequestLinuxDriver.sh
```diff
@@ -25,6 +25,97 @@ function get_md5sum() {
...
+# Get pip
+# - @param1 python_exe - the python executable to install PIP for
+function get_pip() {
+    local python_exe=${1:?}
+
+    echo -e "--- Python: ${python_exe:?}"
+
+    # fetch get-pip.py
+    echo -e "--- curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py"
+    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
```
jmgate, line 38: Do you want to use a pattern like this to avoid potential glitches in the future where someone edits the command but not what gets `echo`ed?
```bash
cmd="insert command here"
echo -e "--- ${cmd}"
eval "${cmd}"
```

cmake/std/PullRequestLinuxDriver.sh
```diff
@@ -25,6 +25,97 @@ function get_md5sum() {
...
+
+# Load the right version of Git / Python based on a regex
+# match to the Jenkins job name.
+function bootstrap_modules() {
+    echo -e "PRDriver> ---------------------------------------"
+    echo -e "PRDriver> Bootstrap environment modules Start"
+    echo -e "PRDriver> ---------------------------------------"
+
+    cuda_regex=".*(_cuda_).*"
+    ride_regex=".*(ride).*"
```
jmgate, line 80: Do you want to change the ride job names to include some kind of delimiters around `ride` for increased robustness?

cmake/std/PullRequestLinuxDriver.sh
```diff
@@ -123,23 +198,44 @@ echo -e "PRDriver> "
...
+    --workspaceDir=${WORKSPACE:?}
+    --github_pr_number=${PULLREQUESTNUM:?}
+    --job_number=${BUILD_NUMBER:?}
+    --req-mem-per-core=3.0
+    --max-cores-allowed=29
+    --num-concurrent-tests=4
+    --mode=${mode}
+    #--mode=installation
+    --config="Trilinos/cmake/std/pr_config/pullrequest.ini"
+    #--dry-run
```
jmgate, line 225: Your command line options have multiple styles:  `--fooBar`, `--foo_bar`, and `--foo-bar`.  I think the third is the preference in the community.

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
-    TODO: wcm: Could it be possible to offload environMap and moduleMap into a
-          JSON or Yaml file?  (Yaml allows comments which is more helpful
-          than JSON IMO.)
+    Returns:
+        Namespace object containing the arguments from the command line.
     """
+    parser    = argparse.ArgumentParser(description='Parse the repo and build information', add_help=False)
+    required  = parser.add_argument_group('Required Arguments')
+    optional  = parser.add_argument_group('Optional Arguments')
+    cwd       = os.getcwd()
```
jmgate, line 59: Wherever possible, I'd recommend using `pathlib`'s `Path` in place of the equivalent functionality from `os`.

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
-    TODO: Needs explanation.
-          - How is environMap structured?
-          - How is moduleMap structured?
-    TODO: wcm: Could it be possible to offload environMap and moduleMap into a
-          JSON or Yaml file?  (Yaml allows comments which is more helpful
-          than JSON IMO.)
+    Returns:
+        Namespace object containing the arguments from the command line.
     """
+    parser    = argparse.ArgumentParser(description='Parse the repo and build information', add_help=False)
```
jmgate, line 56: Just curious, but why do you `add_help=False` and then add it back in below?

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
+                          dest="job_number",
+                          action='store',
+                          help='The Jenkins build number',
+                          required=True)
+
+    optional.add_argument('--config',
+                          dest='configfile',
+                          action='store',
+                          default=os.path.join(cwd, "pr_config/pullrequest.ini"),
+                          help="Configuration file. Default=%(default)s",
```
jmgate, line 116: What does the configuration file do?  If I'm a user of this script, this help doesn't tell me enough.

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
+                          required=False)
+
+    optional.add_argument('--workspaceDir',
+                          dest="workspaceDir",
+                          action='store',
+                          default=default_workspace,
+                          help="The local workspace directory that Jenkins set up." +
+                               " Default={}".format(default_workspace),
```
jmgate, line 124: Can we move to Python 3.6+ and use f-strings?

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
+                          required=False)
+
+    optional.add_argument('--packageEnables',
+                          dest="package_enables",
+                          action="store",
+                          default=default_package_enables,
+                          help="Custom packageEnables.cmake override." +
```
jmgate, line 131: More detail would be good here too.

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
+                               " Default={}".format(default_package_enables))
+
+    optional.add_argument('--subprojects_file',
+                          dest="subprojects_file",
+                          action="store",
+                          default=default_subprojects_file,
+                          help="Custom subprojects file." +
+                               " Default={}".format(default_subprojects_file))
```
jmgate, line 139: Not sure what this means.

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
+                          action='store',
+                          default=12,
+                          help="Max cores allowed, if >= 0 we will use the # of " +
+                               "detected cores on the system. Default = %(default)s")
+
+    optional.add_argument('--num-concurrent-tests',
+                          dest='num_concurrent_tests',
+                          action='store',
+                          default=-1,
+                          help="Number of concurrent tests to run. Default = %(default)s")
```
jmgate, line 166: What does -1 mean as the default.

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
+
+    optional.add_argument("--dry-run",
+                          dest="dry_run",
+                          action="store_true",
+                          default=False,
+                          help="Enable dry-run mode. Runs script but don't execute the build(s) Default = %(default)s")
```
jmgate, line 172: Run instead of runs?

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
+    print("|")
+    print("| - TRILINOS_TARGET_REPO   : {targetRepo}".format(**vars(arguments)))
+    print("| - TRILINOS_TARGET_BRANCH : {targetBranch}".format(**vars(arguments)))
+    print("|")
+    print("| - PULLREQUESTNUM         : {github_pr_number}".format(**vars(arguments)))
+    print("| - BUILD_NUMBER           : {job_number}".format(**vars(arguments)))
+    print("| - PACKAGE_ENABLES        : {package_enables}".format(**vars(arguments)))
+    print("|")
+    print("| - DRY_RUN                : {dry_run}".format(**vars(arguments)))
+    print("+" + "-"*78 + "+")
```
jmgate, line 220: Should the printing be pulled out into a separate routine?

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
-    -  master_merge_YYYYMMDD_HHMMSS
+    # Banner
+    print("+" + "="*78 + "+")
+    print("|")
+    print("|   T R I L I N O S   P U L L R E Q U E S T   D R I V E R   S T A R T")
+    print("|")
+    if args.dry_run:
+        print("|                D R Y   R U N   M O D E   E N A B L E D")
+        print("|")
+    print("+" + "="*78 + "+")
```
jmgate, line 238: Not sure if you want this here, but I like doing things like
```python
print(start_char + banner_text.center(width, fill_char) + end_char)
```

cmake/std/PullRequestLinuxDriverTest.py
```diff
@@ -42,706 +40,221 @@
...
-    """
-    output = "PR-{github_pr_number}-test-{job_base_name}-{job_number}".format(**vars(arguments))
-    return output
-
 
+if __name__ == "__main__":
+    args = parse_args()
+    status = main(args)
+    print("Done.")
+    sys.exit(status)
```
jmgate, line 258: Do you need anything for catching/handling exceptions?  You may not, but we usually have a `try`/`catch` block in `main`.

cmake/std/PullRequestLinuxPython2.cmake
```diff
@@ -14,7 +14,26 @@
...
 
-set(PYTHON_EXECUTABLE /projects/sierra/linux_rh7/install/Python/2.7.15/bin/python CACHE FILEPATH "Set by default for PR testing")
+#set(PYTHON_EXECUTABLE /projects/sierra/linux_rh7/install/Python/2.7.15/bin/python CACHE FILEPATH "Set by default for PR testing")
+#set(PYTHON_EXECUTABLE /projects/sierra/linux_rh7/install/Python/2.7.5/bin/python CACHE FILEPATH "Set by default for PR testing")
+set(PYTHON_PIP_EXECUTABLE "pip2" CACHE STRING "Set by default for PR testing")
+
+
+set(PYTHON_EXECUTABLE_SEARCH_PATHS
+    /projects/sierra/linux_rh7/install/Python/2.7.15
+    /projects/sierra/linux_rh7/install/Python/2.7.5
```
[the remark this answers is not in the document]
prwolfe: Long story - basically because testing using the sems installs is difficult (no mock package.) We are moving towards python3 only which will help remove this issue. Keeping both complicates a few other things as well.

cmake/std/pr_config/pullrequest.ini
```diff
@@ -0,0 +1,366 @@
...
+module-load sems-zlib: 1.2.8/base
+module-load sems-hdf5: 1.10.6/base
+module-load sems-netcdf: 4.7.3/base
+module-load sems-metis: 5.1.0/base
+module-load sems-superlu: 4.3/base
+module-load sems-cmake: 3.10.3
+use ATDM-ENV:
+module-load atdm-ninja_fortran: 1.7.2
+setenv SEMS_FORCE_LOCAL_COMPILER_VERSION: 4.9.3
+setenv OMP_NUM_THREADS: 2
```
jmgate, line 240: I don't understand this in relation to `Trilinos_pullrequest_gcc_7.2.0` above.  Shouldn't it just be using that one and making whatever changes are necessary?

cmake/std/pr_config/pullrequest.ini
```diff
@@ -0,0 +1,366 @@
...
+
+
+[Trilinos_pullrequest_clang_7.0.1]
+use SEMS-ENV:
+use ATDM-ENV:
+module-load sems-gcc: 5.3.0
+module-load sems-clang: 7.0.1
+use SEMS-DEFAULT:
+use ATDM-DEFAULT:
+module-load sems-cmake: 3.12.2
```
jmgate, line 279: Why do we use so many different CMake versions?

cmake/std/pr_config/pullrequest.ini
```diff
@@ -0,0 +1,366 @@
...
+module-load sems-hdf5: 1.10.6/parallel
+module-load sems-netcdf: 4.7.3/parallel
+module-load sems-parmetis: 4.0.3/parallel
+module-load sems-scotch: 6.0.3/nopthread_64bit_parallel
+module-load sems-superlu: 4.3/base
+module-load sems-cmake: 3.17.1
+module-load sems-ninja_fortran: 1.10.0
+
+setenv SEMS_FORCE_LOCAL_COMPILER_VERSION: 5.3.0
+setenv OMP_NUM_THREADS: 2
```
jmgate, line 315: Should this be using `SEMS-DEFAULT` and then modifying it?

cmake/std/pr_python_test_driver.sh
```diff
@@ -0,0 +1,100 @@
...
+    pip_args=(
+        --use-feature=2020-resolver
+        configparser
+        mock
+        pytest
+        pytest-cov
+    )
+    echo -e "--- ${pip_exe:?} install --user ${pip_args[@]}"
+    ${pip_exe:?} install --user ${pip_args[@]}
+}
```
jmgate, line 59: I think I already saw these two functions elsewhere.  Should they be pulled out into some other file that's sourced instead?

cmake/std/pr_python_test_driver.sh
```diff
@@ -0,0 +1,100 @@
...
+echo "|"
+echo "|                   P Y T H O N   T E S T   D R I V E R"
+echo "|"
+echo "+================================================================================+"
+echo ""
+echo "+--------------------------------------+"
+echo "|   P A C K A G E   S E T U P"
+echo "+--------------------------------------+"
+: "${1:?Param 1 ERROR: Python Executable not set or empty}"
+: "${2:?Param 2 ERROR: Pip Executable not set or empty}"
```
jmgate, line 73: I've never seen this pattern before.  What's happening here?

cmake/std/PullRequestLinuxDriver.sh
william76: @jmgate I'll have a look. Usually in Bash what I like to do is something like this:

```bash
cmd="the_executable"
args=(
    arg1
    arg2
    arg3
)
echo -e "--- ${cmd} ${args[@]}"
eval ${cmd} ${args[@]}
```

I like the associative arrays because if you want to disable one of the arguments you can just comment out that line in the array.  When people use continuation lines for long commands like this:

```bash
mycommand \
    arg1 \
    arg2 \
    arg3
```
if you want to turn off arg2, you can't just comment out that line because it will cause an error in bash.

In the past, I haven't typically launched other scripts via `eval`... is there any advantage to using that versus just launching the command?

cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
```diff
@@ -0,0 +1,587 @@
...
+        arg_package_enables_file: Path to the `packageEnables.cmake` file.
+        arg_workspace_dir: Path to the workspace (this would be the ${WORKSPACE}
+            variable set by Jenkins.)
+        arg_pr_config_file: The config.ini file that specifies the configuration to load.
+        arg_pr_jenkins_job_name: The Jenkins Job Name.
+        subprojects_file:
+        working_directory_ctest: Gen. working dir where TFW_testing_single_configure_prototype
+            is executed from.
+        trilinos_pr_env: Environment variable info specific to Trilinos Jobs.
+        config_data:
```
jmgate, line 49: Missing descriptions for `subprojects_file` and `config_data`?

cmake/std/PullRequestLinuxDriver.sh
william76: I will note though, in this case I was just cutting and pasting from the info I got from Python's website.  We're going to be removing the Python2 dependence on our PR driver scripts either as part of this update or right after we get this in... Since the 'get-pip.py' script need is mostly for Python2 support since our Python3 implementations actually have pip3 packaged with them this little function will be able to be removed.  

Many things get easier once we can ditch python2 for PR testing purposes... it's end-of-life was Jan 1, 2020 anyhow so pretty much everyone ought to be phasing out Python2 use anyhow ;)

william76: I've made this change in my working copy, it'll be in with the next push.

cmake/std/PullRequestLinuxDriver.sh
william76: This particular code is generally cut and paste in from the old scripts... I pretty much left this stuff as is.  I'm thinking we can do a cleanup pass on PullRequestLinuxDriver.sh after we remove the Python2 dependency and get the updated PR script in.

But I do agree that we do need to do some pass at some point to handle system-specific issues more cleanly. Currently stuff like this is spread around in a few places.

cmake/std/PullRequestLinuxDriver.sh
william76: Yep. 

I generally prefer using dashes like `--foo-bar` ... I will look at this with some cleanup work.

I think I kept these as is because some of it was there to match what was in the original script that I ported over from.  It was easier to do just keeping things the same.

Now that it's working in the PR system, it'll be easier to do a search/replace ;)

I'll add this to my list of todo items.

cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
```diff
@@ -0,0 +1,587 @@
...
+    def concurrency_build(self):
+        """
+        Concurrency to use for building Trilinos
+
+        This is equvalent to running `make -j <concurrency_build>` from the command line.
+        """
+        if self._concurrency_build is None:
+            si = sysinfo.SysInfo()
+
+            self._concurrency_build = si.compute_num_usable_cores(req_mem_gb_per_core = self.arg_req_mem_per_core,
```
jmgate, line 282: Should there be spaces here:  `req_mem_gb_per_core = self.arg_req_mem_per_core`?

cmake/std/PullRequestLinuxDriverTest.py
william76: Looks like this is Python3 only? 

We can look into this later... once our dependence on Python2 is gone.

```
[wcmclen@tr-test-4 ~]$ python
Python 2.7.9 (default, Feb 14 2020, 08:39:29)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pathlib
  File "<stdin>", line 1, in <module>
ImportError: No module named pathlib
```

cmake/std/PullRequestLinuxDriverTest.py
william76: It's the path to the configuration .ini file that contains the environment specifications mapped out to PR jobs. 

I've updated the description.

cmake/std/PullRequestLinuxDriverTest.py
william76: Not until SEMS provides a Python 3.6+

cmake/std/PullRequestLinuxDriverTest.py
william76: It's an option normal users probably should never touch.  the subprojects file is a generated file that gets passed into the `ctest` call that ultimately drives everything.

cmake/std/PullRequestLinuxDriverTest.py
william76: I updated the documentation.

This is equivalent to `ctest -j <num_concurrent_tests>` ... if the value is >= 0 then it's considered an override and that value is used. Otherwise, it will compute the parameter via `num_cores / max_test_parallelism`

cmake/std/PullRequestLinuxDriverTest.py
william76: /shrug semantics.

cmake/std/PullRequestLinuxDriverTest.py
william76: I don't think it really adds any value in this particular spot...  There are places where things woof and I am cool with the traceback just ending the application at that spot.  Generally that means something pretty bad went wrong and I don't really need another layer of traceback since I'll still have it anyhow. All a try...catch in this spot would be useful for is as a pass-through for an error. /shrug

cmake/std/PullRequestLinuxDriverTest.py
william76: In this particular case I don't know that it matters either way.  Putting it into a function might make Uncle Bob happy but the purpose of this block is to print out the arguments that were just parsed.  In this case, not ALL the arguments but the ones that came in from the PR job's parameters itself.

I may expand this to dump out what the values of all the parameters are but I think it can stay here.

cmake/std/README.md
```diff
@@ -1,6 +1,6 @@
 README
 ------
-
+- 2020-10-03 -
```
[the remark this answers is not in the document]
william76: meh, I'll just delete the file.  I'm the only one who ever touched it.  If we want to do something better in the future we can add one later.

cmake/std/pr_config/pullrequest.ini
william76: Nope.

The section names == the Jenkins Job name.  In this case, there are Jenkins jobs named: 
- `Trilinos_pullrequest_gcc_7.2.0`
- `Trilinos_pullrequest_gcc_7.2.0_serial`
- `Trilinos_pullrequest_gcc_7.2.0_debug`

so there must be a section for each or things will fail.  In this case, 7.2.0 and debug are identical so that one just loads in the same stuff from 7.2.0, but the _serial_ one must load the serial version of several packages. It also omits openmpi, etc.

cmake/std/pr_config/pullrequest.ini
william76: I have no idea, but my configurations are set up to match what the current PR test system uses. 

Most likely, it's because the different builds were configured at different times when the newest CMake was different. It might be that we need to use different versions of CMake to have coverage in the PR system that things work with different versions of CMake.

cmake/std/pr_config/pullrequest.ini
william76: It could. But I didn't :p

cmake/std/pr_python_test_driver.sh
william76: Yeah, I think it's cut and pasted into two places.

Generally I do have a `common.bash` script where I keep a single copy of functions like these... I'm not heavily optimizing the bash scripts in this PR... most of the legacy bash in Trilinos doesn't really use functions, it's generally just all in-line.  

I think a cleanup pass on stuff like this is worth doing in a follow-on PR.

cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
william76: To me, this:
```python
            self._concurrency_build = si.compute_num_usable_cores(req_mem_gb_per_core = self.arg_req_mem_per_core,
                                                                  max_cores_allowed   = self.max_cores_allowed)
```

is easier to read than:

```python
            self._concurrency_build = si.compute_num_usable_cores(req_mem_gb_per_core=self.arg_req_mem_per_core,
                                                                  max_cores_allowed=self.max_cores_allowed)
```

I like whitespace and I can not lie...

cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
william76: I was working on this yesterday and had to quit for the day.

It is documented in the `config_data` property.

cmake/std/pr_python_test_driver.sh
william76: If the parameter ${1} or ${2} don't exist, then an error is triggered with the message.

for example:
```bash
#!/usr/bin/env bash

: "${1:?Param 1 is missing or empty}"

echo "Param 1 = ${1}"
```

without the parameter:
```
$ ~/param_check.sh
/Users/wcmclen/param_check.sh: line 3: 1: Param 1 is missing or empty
```

with the parameter
```
$ ~/param_check.sh FOO
Param 1 = FOO
```

Think of using this construct as a way to validate expected parameters at the start of a script... kind of a 'design by contract' construct.

cmake/std/PullRequestLinuxPython2.cmake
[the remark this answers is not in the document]
william76: This ^

Also, if SEMS would get around to taking care of that request I filed a long time ago and get us a Python >= 3.6 it'd be super nice.  I think they had some reason why they were going to put that off but I don't recall what it was, but not having access to anything 3.6 or greater means that some python modules will fail to install if they are using f-strings. :/ 

Feel free to file additional tickets against them with the request :p

cmake/std/PullRequestLinuxDriverTest.py
prwolfe: 2 reasons to create a function
- the code is shared between two routines (not applicable here)
- for organization - i.e. to make it more readable. This is where I and Bob Martin part ways. This is small enough to leave as is in my opinion.

cmake/std/pr_config/pullrequest.ini
prwolfe: It's the timing thing I believe

cmake/std/PullRequestLinuxDriverTest.py
jmgate: I need a "shaking my fist" emoji so I can shake my fist at Python 2.

cmake/std/pr_python_test_driver.sh
jmgate: Oh, nice.  Yeah I'd never seen this method of validating parameters in bash.

prwolfe: Can I get that shaking fist emoji for bash please?

cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
jmgate: > I like whitespace and I can not lie...

Oh, I heartily agree.  I thought the spaces were a requirement, but maybe that's just a PEP8 thing.

cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
```diff
@@ -0,0 +1,592 @@
...
+
+        return 0
+
+
+    def validate_branch_constraints(self):
+        """
+        Verify that the source branch is allowed.
+
+        For the `master` branch, we only allow the source branch to be
+        a protected branch named with the scheme `master_merge_YYYYMMDD_HHMMSS`
```
jmgate, line 475: Just a heads-up:  This might be a point of fragility we want to deal with at some point in the future.  Maybe you also want to validate the PR author is legit before trying to merge something into `master`.

cmake/std/trilinosprhelpers/TrilinosPRConfigurationInstallation.py
```diff
@@ -0,0 +1,78 @@
...
+#from . import sysinfo
+#from . import jenkinsenv
+
+from . import TrilinosPRConfigurationBase
+
+
+
+class TrilinosPRConfigurationInstallation(TrilinosPRConfigurationBase):
+    """
+    Implements Standard mode Trilinos Pull Request Driver
```
jmgate, line 20: Installation mode?

jmgate: BTW, does this differ from the `Standard` one, or not yet?

william76: Installation Testing... it will work differently from a standard test.  It's stubbed in for now.

cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
william76: @jmgate branches that match the rule 'master_merge_*' are protected branches so that limits who can actually create one, this is how we'd prevent just anybody from creating a branch to try and merge into the master branch.

william76:
@jmgate 
I put in a pass to make the command line parameters more consistent in formatting and renamed some to be a little more descriptive even in the parameter name.  I checked in that update plus a fix to the driver .sh script which calls it so those things should work.

I will need to go through the code and rename a number of things for consistency... I'll try and get to that tomorrow :D 

Thanks for being quite brave and digging into this code without really any explanation into what is going on with it ;)  We haven't even done a real review on it yet with the Framework team so I appreciate the feedback.

cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
prwolfe: pep8 hates whitespace - I never got that...

cmake/std/common.bash
```diff
@@ -0,0 +1,83 @@
...
+    local script_path=$(dirname $(readlink -f $0))
+    local script_file="${script_path}/${script_name:?}"
+    echo "${script_file}"
+}
+
+
+# Gets the path to the current script (full path)
+function get_scriptpath() {
+    # Get the full path to the current script
+    local script_name=`basename $0`
```
jmgate, line 20: Don't think you need this line.

william76: `get_scriptname` and `get_scriptpath` were both in a different script, since I've added a `common.bash` script that holds some of these functions I moved them into it.

They're probably not necessary since other scripts are finding their locations and loading in the common.bash script via this:
```bash
SCRIPTFILE=$(realpath $BASH_SOURCE)
SCRIPTPATH=$(dirname $SCRIPTFILE)
source ${SCRIPTPATH:?}/common.bash
```

so we may not really need these functions any more but I don't think it hurts to hang onto them.  I'm not going to remove them for now.

I'm not going whole hog to optimize and clean up the bash stuff in this PR, I think that's an issue in itself.

cmake/std/trilinosprhelpers/jenkinsenv/JenkinsEnv.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+    @property
+    def git_url(self):
+        return self.get_envvar_str("GIT_URL")
+
+
+    @property
+    def git_branch(self):
+        return self.get_envvar_str("GIT_BRANCH")
```
jmgate, line 108: Do these suckers need `error_if_missing=False` as well?

william76: I'd have to dig into this but for now I think returning None rather than throwing is ok on these.  I don't actually think most of these are used but I can have a look.

cmake/std/trilinosprhelpers/setenvironment/ModuleHelper.py
```diff
@@ -0,0 +1,94 @@
...
+    print("WARNING: The environment variable 'MODULESHOME' was not found.")        # pragma: no cover
+    print("         ModuleHelper may not be able to locate modules.")              # pragma: no cover
+
+try:
+
+    from env_modules_python import module
+
+except ImportError:
+
+    def module(*args):
```
jmgate, line 31: At some point this routine/file could use more complete documentation.  This is the first thing I've run into in this PR where I'm not really sure what it's doing.

cmake/std/common.bash
jmgate: No problem.  I'm just asking questions as I run into them.  Filing the feedback away for later is fine with me.

cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
```diff
@@ -0,0 +1,447 @@
...
+
+        If we set filtered_keys_only to True then we ONLY print out envvars that
+        matched envvar_filter.
+
+        Arguments:
+            envvar_filter (list): a list of keys to print out the value.
+                all envvar values are printed if omitted.
+                Default: None
+            filtered_keys_only (bool): If true, we only display filtered keys.
+                If false, we display the keys of all keys.
```
jmgate, line 233: Swap the first `keys` in this line for `values`?

cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
```diff
@@ -0,0 +1,447 @@
...
+                        break
+            else:
+                filtered_keys_only = False
+
+            if filtered_keys_only == False or matched_key:
+                print("--- {}".format(k), end="")
+                if envvar_filter is None:
+                    print(" = {}".format(v), end="")
+                elif matched_key:
+                    print(" = {}".format(v), end="")
```
jmgate, line 266: I don't understand this if-block.  Looks like it's doing the same thing in either case.

cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
```diff
@@ -0,0 +1,447 @@
...
+
+
+    def _load_configuration(self):
+        """
+        Loads the environment data from the provided `profile` and saves the
+        information to the `actions` property.
+
+        This is a recursive operation (recurses on the `use <profile>:` entries
+        in the configuration file). We use a DFS recursion strategy with back-edge
+        checking to prevent cyclic recursion. We do not
```
jmgate, line 340: We do not&hellip;?

cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
```diff
@@ -0,0 +1,447 @@
...
+
+    def _load_configuration_r(self, config, profile, actions=None, processed_secs=None):
+        """
+        Recursive handler for _load_configuration.
+        Recursion is called when the operation is a 'use' operation in the key.
+
+        Raises:
+            KeyError: If `profile` section is not contained in `config`
+            Exception: If a section isn't loaded for a reason other than a KeyError.
+        """
```
jmgate, line 366: This routine is a little tough to follow.  I'm sure it makes good sense to you, as you're intimately familiar with the `*.ini` file syntax and what all that means, but as someone new coming up to speed on it, it's hard to tell what the various keys/values really mean wherever they're used.

cmake/std/trilinosprhelpers/setenvironment/test/test_ModuleHelper.py
```diff
@@ -0,0 +1,150 @@
...
+    Specialization of popen mock that will return with success.
+    """
+    def __init__(self, cmd, stdout=None, stderr=None):
+        super(mock_popen_status_ok, self).__init__(cmd,stdout,stderr)
+
+
+
+class mock_popen_status_error(mock_popen):
+    """
+    Specialization of popen mock that will return with success.
```
jmgate, line 63: Error instead of success?

cmake/std/trilinosprhelpers/setenvironment/test/test_ModuleHelper.py
```diff
@@ -0,0 +1,150 @@
...
+        print("mock_popen> communicate()")
+        stdout = b"stdout=1"
+        stderr = b"ERROR: Something wrong happened."
+        self.returncode = 1
+        return (stdout,stderr)
+
+
+class mock_popen_status_error_mlstatus(mock_popen):
+    """
+    Specialization of popen mock that will return with success.
```
jmgate, line 78: Update docstring.

cmake/std/trilinosprhelpers/setenvironment/test/test_ModuleHelper.py
```diff
@@ -0,0 +1,150 @@
...
+    def test_module_swap_status_ok(self, arg_popen):
+        r = ModuleHelper.module("swap", "sems-gcc/4.8.4", "sems-gcc/7.3.0")
+        print("result = {}".format(r))
+        self.assertEqual(0, r)
+
+
+    #@patch('subprocess.Popen', side_effect=mock_popen_status_ok)
+    def test_module_unload_status_ok(self):
+        with patch('subprocess.Popen', side_effect=mock_popen_status_ok):
+            r = ModuleHelper.module("unload", "sems-gcc/4.8.4")
```
jmgate, line 125: What's the difference between the `@patch` decorator and the `with patch(...):` pattern?

cmake/std/trilinosprhelpers/setenvironment/test/test_ModuleHelper.py
```diff
@@ -0,0 +1,150 @@
...
+        print("result = {}".format(r))
+        self.assertEqual(0, r)
+
+
+    def test_module_load_args_as_list(self):
+        with patch('subprocess.Popen', side_effect=mock_popen_status_ok):
+            r = ModuleHelper.module(["unload", "sems-gcc/4.8.4"])
+        print("result = {}".format(r))
+        self.assertEqual(0, r)
```
jmgate, line 134: Not sure the test is doing what the name suggests, but maybe I'm just reading it wrong.

cmake/std/trilinosprhelpers/setenvironment/test/test_SetEnvironment.py
```diff
@@ -0,0 +1,520 @@
...
+                ['load', 'sems-gcc/7.3.0'],
+                ['load', 'sems-boost/1.63.0/base'],
+                ['load', 'sems-cmake/3.10.3'],
+                ['load', 'sems-python/3.5.2']
+            ],
+            'setenv': [   {'key': 'OMP_NUM_THREADS', 'value': '2'},
+                          {'key': 'CC', 'value': 'gcc'},
+                          {'key': 'CXX', 'value': 'g++'},
+                          {'key': 'FC', 'value': 'gfortran'},
+                          {'key': 'TEST_ENVVAR_001', 'value': 'TEST_ENVVAR_001_VALUE'}
```
jmgate, line 95: I'm kind of surprised the `setenv` thing builds up a list of dicts instead of just a dict.

cmake/std/trilinosprhelpers/setenvironment/test/test_SetEnvironment.py
```diff
@@ -0,0 +1,520 @@
...
+        for opt in options:
+            print("---- {}".format(opt))
+        for k,v in setEnv.config.items("SAMPLE_MAP"):
+            print("---- {} : {}".format(k,v))
+
+
+    def test_SetEnvironment_pretty_print_envvars(self):
+        """
+
+        """
```
jmgate, line 433: Docstring?

cmake/std/trilinosprhelpers/setenvironment/test/test_SetEnvironment.py
```diff
@@ -0,0 +1,520 @@
...
+
+        # cleanup
+        del os.environ["SETENVIRONMENT_TEST_ENVVAR_001"]
+        del os.environ["SETENVIRONMENT_HIDDEN_ENVVAR_001"]
+
+
+    def _setEnv_test(self, filename, profile, truth=None, module_fail=False):
+        """
+        Test the instantiation of a SetEnvironment class. Loads the
+        ini file and applys its settings.
```
jmgate, line 476: applies

jmgate: Oh, or maybe `apply()s`

cmake/std/trilinosprhelpers/sysinfo/SysInfo.py
```diff
@@ -0,0 +1,186 @@
...
+    compile. We should select the # of cores we need in a parallel make to
+    ensure we don't violate the high-water mark.
+
+    One constraint we have on this class is that there are two ways to get the
+    memory avilable on a system.
+        1. The preferred way is to use the ``psutil`` module in Python.
+        2. If ``psutil`` isn't available, we attempt to use a fall-back
+           by reading in the ``/proc/meminfo`` file, which should exist
+           on *nix systems.
+        3. If ``psutil`` and ``/proc/meminfo`` isn't found then we will
```
jmgate, line 54: aren't

cmake/std/trilinosprhelpers/sysinfo/SysInfo.py
```diff
@@ -0,0 +1,186 @@
...
+
+    @have_psutil.setter
+    def have_psutil(self, value):
+        """
+        Setter for the 'have_psutil' property.
+
+        Args:
+            value (bool): A boolean value that indicates whether or not the
+                          algorithm will use psutil or not. Note: even if
+                          psutil is availble, setting this explicitly to False
```
jmgate, line 87: available

cmake/std/trilinosprhelpers/test/test_TrilinosPRConfigurationBase.py
```diff
@@ -0,0 +1,712 @@
...
+    print("    : Command is `{}`".format(cmd))                                 # pragma: no cover
+    raise subprocess.CalledProcessError(1, cmd)                                # pragma: no cover
+
+
+def mock_modulehelper_module_ok(*args, **kwargs):
+    """
+    Mock replacement for a ModuleHelper.module() call that succeeds
+    """
+    cmd = ", ".join(["'{}'".format(x) for x in args])
+    print("MOCK> module({})    QAPLA'!".format(cmd))                           # pragma: no cover
```
jmgate, line 112: Klingon?

cmake/std/trilinosprhelpers/test/test_TrilinosPRConfigurationBase.py
```diff
@@ -0,0 +1,712 @@
...
+        int 0 for PASS
+    """
+    params = ", ".join(["{}={}".format(k,v) for k,v in kwargs.items()])         # pragma: no cover
+    print("MOCK> SetEnvironment.apply({})".format(params))                      # pragma: no cover
+    return 0                                                                    # pragma: no cover
+
+
+def mock_se_apply_fail(*args, **kwargs):
+    """
+    Mock for SetEnvironment.apply() call that would pass (i.e., return 1)
```
jmgate, line 147: fail

william76: @jmgate this is not a useful comment, please elaborate.

cmake/std/trilinosprhelpers/test/test_TrilinosPRConfigurationBase.py
william76: Success!

cmake/std/trilinosprhelpers/setenvironment/test/test_SetEnvironment.py
william76: a list of dicts that have identical keys.

Note that each dict in this has the keys "key" and "value" ... when I iterate over this, it looks for the "key" and "Value" properties.

cmake/std/trilinosprhelpers/setenvironment/test/test_ModuleHelper.py
william76: They're different flavors of the same thing.

cmake/std/trilinosprhelpers/setenvironment/ModuleHelper.py
william76: This is addressed in the docstring at the top of this file.

I added a comment that explains at the point of the exception, but it's kind of redundant to what is stated at the top of the file.  Essentially, we try and import module() from env_modules_python... if that import fails then we roll our own version of that function.

cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
william76: nope, it is correct but not well written.

If `filtered_keys_only` == false then we will display the keys of all envvars.  if filtered_keys_only is true then we completely ignore printing any envvars that don't have a match.

cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
william76: the default argument to envvar_filter is None.  so you can call `pretty_print_envvars()` with no parameters and it should work.

This if/elif block could have also been written:

```python
if envvar_filter is None or matched_key:
    print(...)
```

but I broke it out into an if / elif to make it easier to do something different if I want to change it later.  This is a habit actually from my work as a compiler developer back in the day.  I tend to unroll things a bit more since compilers will do the fusion for you, but it is often easier to understand code that doesn't overly complicate the expression or if people try and put too much into a loop.  

In terms of performance, there's little difference between fusing this into a single if statement vs. having the if..elif.

The general operation to this function is something like:
```
for each item in the environment vars:
    for each item in the list of filters:
        if the envvar name contains the filter item:
            set matched_key = true and break
    print the envvar if it should be printed and its value if that should be printed
```

cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
william76: This is the internal recursive call... this is the part that actually does the parsing of the .ini file.  

I updated the docstring a little and cleaned up a few things. Turns out the `config` parameter isn't necessary as a parameter since it's the same as the property `self.config`, so that simplifies things a bit.  I renamed the profile parameter to be `config_section` to better describe that this corresponds to a section in the .ini file `[section name]` in a config file.

We could get into the details of this how what this does but it's probably not easy to do in a PR comment.

cmake/std/trilinosprhelpers/setenvironment/test/test_ModuleHelper.py
william76:
```python
module( ["unload", "sems-gcc/4.8.4"] )
#       ^                          ^ - list.
```

cmake/std/trilinosprhelpers/test/test_TrilinosPRConfigurationBase.py
jmgate: Sorry, switch the word `pass` to `fail` in the docstring.

cmake/std/trilinosprhelpers/setenvironment/test/test_SetEnvironment.py
jmgate: Oh, gotcha.

cmake/std/trilinosprhelpers/setenvironment/test/test_ModuleHelper.py
jmgate: Oh, gotcha.  The module subcommand (load vs unload) doesn't matter because it's testing passing the args in as a list.  👍

jmgate:
K, I'm done with comments.  Thanks for humoring me as I came up to speed.  Again, this is a *huge* step forward for us.

jmgate:
Something to handle in a future update:  Currently when printing out the parameters, we see in the Jenkins console log
```
00:00:25.705 | - PACKAGE_ENABLES        : ../packageEnables.cmake
```

It'd probably be good to turn that into an absolute path instead so the user isn't left wondering which `packageEnables.cmake` is being used.  Similarly, later on we see
```
00:00:25.705 --- subprojects_file         = ../package_subproject_list.cmake
```

jmgate (approved):

prwolfe (approved):

william76:
@prwolfe interesting... still says can't merge. Any ideas on what Github is missing?

prwolfe:
> @prwolfe interesting... still says can't merge. Any ideas on what Github is missing?

I think the autotester has to catch up and verify everything is done. Are there unresolved conversations - I don't see that, but sometimes that keeps them from merging.

prwolfe:
I did some checking - the autotester instance 2 that should pick this up is waiting on resources on ride. Until that fails on timeout or runs everything is waiting.

allevin:
@william76  I think @prwolfe is correct.  The AT needs time to cycle through and process this PR to verify its been reviewed and should then set the pre-merge inspection flag.   Patience my young padawan!

william76:
> I did some checking - the autotester instance 2 that should pick this up is waiting on resources on ride. Until that fails on timeout or runs everything is waiting.

@prwolfe Looks like instance2 is still waiting... going on just under 6 hours now.  Assuming it doesn't wrap up soon, I'll probably wait and merge it tonight or tomorrow morning early.

prwolfe:
Looks like all nodes on ride are running jobs, so this is likely to not return soon.

william76:
@prwolfe @jwillenbring @ZUUL42 I'll merge this in today during the kiddo's nap... probably around 12:30pm.  Once it's in I'll put in a toy PR that just touches a file in MueLu or something to kick off a test.

prwolfe:
Sounds good – I will watch for it!

william76:
@prwolfe merging this in now...

william76:
@jmgate FYI - merged this in today.  We'll keep an eye on things today to make sure all works properly.

PR #8144, opened 2020-10-05
MueLu: Correct logic in RegionRFactory and update MakeRegionMatrices to not forget block size

GrahamBenHarper:
~~Please do not merge this yet. I want to squash the commits down, but I don't want to do so until I see how it turns out.~~
Edit: everything has been squashed.
@trilinos/muelu 

There was a flaw in the logic in `RegionRFactory` that was the source of the issue. For a restrictor, the block size should be the outer loop. In addition, `MakeRegionMatrices` was throwing away the block size of the matrix A from which it was created, which caused some region tests to look like they were correctly preparing the 3 DOF per node version, when in fact the matrices in `regionGrpMats` were not treated as having 3 DOFs per node.

## Testing
I noticed a significant improvement in the iterations to solve the related regression tests. ~~I haven't run other tests yet because I want to see how this impacts the repartitioning, in particular.~~
Edit: all MueLu experimental tests pass.

~~@pohm01 can you please test your repartitioning examples based off this branch? I see an error when I try to run the repartitioning example you sent, but I also see the same error when I change the input xml from the `linear_R` flavor to the `linear` flavor.~~
Edit: I confirmed with Peter, and everything looks good when combined with his PR now.

### DIFF
--- a/packages/muelu/research/regionMG/src/SetupRegionMatrix_def.hpp
+++ b/packages/muelu/research/regionMG/src/SetupRegionMatrix_def.hpp
@@ -598,6 +598,8 @@ void MakeRegionMatrices(const RCP<Xpetra::CrsMatrixWrap<Scalar, LocalOrdinal, Gl
 
     regCorrection[j] = VectorFactory::Build(revisedRowMapPerGrp[j], true);
     regionGrpMats[j]->apply(*regNsp[j], *regCorrection[j]);
+    regionGrpMats[j]->SetFixedBlockSize(AComp->GetFixedBlockSize());
+
   }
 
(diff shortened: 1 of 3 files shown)


GrahamBenHarper:
Related to #7979 and #8038.

GrahamBenHarper:
I just confirmed with @pohm01 that this PR combined with #8133 does fix the region+repartitioning examples.

GrahamBenHarper:
I've verified that all MueLu tests, including the experimental ones, pass on my workstation now.

pohm01 (approved):
ok.

cgcgcg:
Ha, it's finally running on a different node!

Issue #8148, opened 2020-10-05
Intrepid2: avoid copies of DFad during unit test comparisons


CamelliaDPG:
## Bug Report
@trilinos/intrepid2 

### Description
See discussion under #8147.  The unit test:
``
AnalyticPolynomialsMatch_Sacado_Fad_DFadType_Sacado_Fad_DFadType_HierarchicalNodalComparisons_UnitTest
``
takes 400 of the 407 seconds that the test ` Intrepid2_unit-test_Discretization_Basis_HierarchicalBases_Hierarchical_Basis_Tests_MPI_1` takes to run.  A bit of investigation shows that at least part of the story is that during the floating point comparison part of the test, there are unnecessary copies happening.  This issue is to eliminate those extra copies.

PR #8149, opened 2020-10-05
Intrepid2: modifications to avoid unnecessary copies of Fad types during test comparisons.

CamelliaDPG:
Intrepid2: modifications to avoid unnecessary copies of Fad types during test comparisons.

See #8148.


mperego (approved):
Looks good!

PR #8150, opened 2020-10-05
MueLu: regionMG disable rebalance tests in serial build

pohm01:
@trilinos/muelu 

## Motivation

Fixes failing regionMG tests by removing them from serial builds. The removed tests were for rebalancing and thus not as relevant to a serial build.

## Testing
The failing tests no longer appear in serial builds on my machine, but still appear and pass on non-serial builds.


packages/muelu/research/regionMG/test/structured/CMakeLists.txt
```diff
@@ -144,6 +145,7 @@ IF (${PACKAGE_NAME}_ENABLE_Tpetra AND ${PACKAGE_NAME}_ENABLE_Amesos2)
     COMM serial mpi
```
cgcgcg, line 145: Wouldn't it be enough to remove `serial`?

pohm01: Is that how this works? Because that does seem like it would be better.

cgcgcg: Yes, in combination with `NUM_MPI_PROCS 4` this should do the trick

pohm01: I will test that, then change this PR to be that if it works.

pohm01: I did it.

cgcgcg (approved):

PR #8114, opened 2020-09-28
Framework/ifpack2/kokkos-kernels: Set standard to C++14 for CUDA 9.2 PR build

ZUUL42:
@trilinos/framework
@trilinos/ifpack2
@trilinos/kokkos-kernels

## Motivation

This will add the needed flags to set the standard to C++14 for the CUDA 9.2.88 build.
It also fixes a few build errors that come up in ifpack2 and kokkos-kernels.

Associated with #6260
Closes #8016

## Stakeholder Feedback

This work moves to a configuration that is c++14 ready making the PR testing scheme ready for Kokkos 3.3 to be merged in.
## Testing

Tested by running the CUDA 9.2.88 PR build job manually, merging this branch into develop.
Results: [CDash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=3&showfilters=1&filtercombine=and&field1=site&compare1=61&value1=ride12&field2=buildname&compare2=61&value2=PR-cuda9c14fix-test-Trilinos_pullrequest_cuda_9.2-5629&field3=buildstamp&compare3=61&value3=20200928-0300-Experimental)

Issue #8071, opened 2020-09-22
PR Testing: Create a new GCC 7.2.0 C++14 Serial PR build


ZUUL42:
## Enhancement
@trilinos/framework 

We are looking to add -std=c++14 to a GCC 7.2.0 build. While some packages are using C++14 (and C++17) features, this will move us in the right direction and ensure that customers will be able to build Trilinos with C++14.
Moving to C++14 will also require us to turn off testing for the 4.9.3 Serial build, so this GCC 7.2.0 build will also be a serial build.

This incremental step is consistent with discussions at the Trilinos Product Leaders meeting and is related to #6260.

ZUUL42:
Initial testing has 20 build errors.
[CDash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=3&showfilters=1&filtercombine=and&field1=site&compare1=61&value1=ascic158&field2=buildname&compare2=61&value2=PR-gcc72c14serial-test-Trilinos_pullrequest_gcc_7.2.0-5688&field3=buildstamp&compare3=61&value3=20200920-0048-Experimental)
All are WError related, either unused-but-set-variable, unused-variable, or one unused-label
Affected packages include @trilinos/nox @trilinos/epetraext @trilinos/ML @trilinos/stokhos @trilinos/anasazi @trilinos/muelu @trilinos/pamgen @trilinos/trilinoscouplings 

Since these warnings are covered in the GCC 8.3.0 build, unless anyone has any objections, I will turn off these warnings for this build. `-Wno-unused-but-set-variable -Wno-unused-variable -Wno-unused-label` Then we've got a clean build for C++14.
[CDash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=3&showfilters=1&filtercombine=and&field1=site&compare1=61&value1=ascic158&field2=buildname&compare2=61&value2=PR-gcc72c14serial-Wno-test-Trilinos_pullrequest_gcc_7.2.0-5695&field3=buildstamp&compare3=61&value3=20200922-0808-Experimental)

jhux2:
> Since these warnings are covered in the GCC 8.3.0 build, unless anyone has any objections, I will turn off these warnings for this build.

@ZUUL42 Would you post a link to the GCC 8.3.0 build?  I can't seem to find it in the list of MueLu dashboard results.

lucbv:
[...]

PR #8152, opened 2020-10-06
C++14 clean up

ZUUL42:
@trilinos/framework 

A little clean up after the C++14 changes:

- The `CMAKE_CXX_FLAGS "-std=c++14` are uneseccary and [can cause issues](https://github.com/trilinos/Trilinos/pull/8114#discussion_r496124370), so they are removed.
- `CMAKE_CXX_STANDARD "14"` is moved to a standardized place on top of all cmake files.
- The order of `-DTRILINOS_HIDE_DEPRECATED_HEADER_WARNINGS -Werror` [was causing warnings](https://github.com/trilinos/Trilinos/issues/8071#issuecomment-697743665), so it has been reversed in all applicable locations.
- Set C++14 for the Clang 7 build.

### DIFF
--- a/cmake/std/PullRequestLinuxClang10.0.0TestingSettings.cmake
+++ b/cmake/std/PullRequestLinuxClang10.0.0TestingSettings.cmake
@@ -6,6 +6,8 @@
 
 # Usage: cmake -C PullRequestLinuxClang10.0.0TestingSettings.cmake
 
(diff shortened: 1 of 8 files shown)


jwillenbring (approved):
Looks good. Thank you for the clean up pass and the solid documented explanation.

PR #8154, opened 2020-10-06
Intrepid2: prototype faster Fad comparisons for tests

CamelliaDPG:
@trilinos/intrepid2

## Motivation
Following on #8148, this PR prototypes a new mechanism for performing error measurements, shifting more of the burden to the device.  In the Fad test we're working with, this reduced the runtime on ride by a factor of about 6.  (#8148 itself had reduced the runtime by about a factor of 5, so this represents a cumulative 30x speedup.)

The ultimate idea is to revise the testFloatingEquality() methods that we define in TestUtils to take advantage of this new mechanism, thereby speeding up all tests that take advantage of those.

### DIFF
--- a/packages/intrepid2/src/Shared/Intrepid2_TestUtils.hpp
+++ b/packages/intrepid2/src/Shared/Intrepid2_TestUtils.hpp
@@ -74,17 +74,18 @@ namespace Intrepid2
   //! Adapted from Teuchos::relErr(); for use in code that may be executed on device.
   template <class Scalar1, class Scalar2>
   KOKKOS_INLINE_FUNCTION
-  typename Teuchos::ScalarTraits< typename std::common_type<Scalar1,Scalar2>::type >::magnitudeType
-  relErr( const Scalar1 &s1, const Scalar2 &s2, const typename Teuchos::ScalarTraits< typename std::common_type<Scalar1,Scalar2>::type >::magnitudeType &smallNumber )
+  bool
+  relErrMeetsTol( const Scalar1 &s1, const Scalar2 &s2, const typename Teuchos::ScalarTraits< typename std::common_type<Scalar1,Scalar2>::type >::magnitudeType &smallNumber, const double &tol )
   {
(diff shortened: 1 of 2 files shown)


CamelliaDPG:
@mperego, please look carefully at what I've done in TestUtils, considering what it means for Fad types.  I *think* this is equivalent to the existing Fad comparison, which uses the traits `magnitude()`.  I couldn't use that here, because those traits methods aren't defined on device.

mperego (approved):
This look good to me. It's a huge performance improvement for tests, thanks!

I've a couple of comments for future use:
1. Now `TEUCHOS_TEST_FLOATING_EQUALITY` supports FAD types
2. Instead of checking for small values, I think it would be better to use a check with both relative and absolute tolerances, something like this:
`| x_ref - x_comp | < abs_tol + x_ref * rel_tol`
This is pretty standard and works well for large numbers (controlled by rel_tol) and small numbers (controlled by abs_tol). I intend to add this function to Teuchos. Maybe we can do that while restructuring tests to become UVM free.

CamelliaDPG:
Thanks, @mperego!  

> 1. Now `TEUCHOS_TEST_FLOATING_EQUALITY` supports FAD types
The trouble with `TEUCHOS_TEST_FLOATING_EQUALITY` is that it must execute on host.  I think Teuchos is written to be Kokkos-agnostic, so I'm not sure that there is any straightforward way to decorate even the lowest-level Teuchos implementations with `KOKKOS_INLINE_FUNCTION`.  And with ride in particular, host execution is quite slow for Fad types.

> 2. Instead of checking for small values, I think it would be better to use a check with both relative and absolute tolerances, something like this:
>    `| x_ref - x_comp | < abs_tol + x_ref * rel_tol`
>    This is pretty standard and works well for large numbers (controlled by rel_tol) and small numbers (controlled by abs_tol). I intend to add this function to Teuchos. Maybe we can do that while restructuring tests to become UVM free.

That sounds good to me.  I can use that in the View-level `testFloatingEquality()` implementations in TestUtils.

PR #8155, opened 2020-10-06
MueLu: Add CMake include directories for regionMG

GrahamBenHarper:
This is for discussion purposes: We'd like to use the regionMG source with an upcoming TrilinosCouplings example, however, simply including the required headers does not work.
```
No such file or directory
 #include "SetupRegionUtilities.hpp"
```
Including the full path also does not work.
```
No such file or directory
 #include "muelu/research/regionMG/src/SetupRegionUtilities.hpp"
```
I suppose this is because `research` is treated as an examples directory by Tribits (line 461 of `muelu/CMakeLists.txt`), but I'm not positive how these things work internally:
```
  TRIBITS_ADD_EXAMPLE_DIRECTORIES(research)
```
This is one proposed fix, and I've verified it works on my end, but I'm hoping to have a discussion here to see if others have suggestions that may be more suitable.


GrahamBenHarper:
I got some feedback from @bartlettroscoe this morning (thank you!), and it seems that placing all of `muelu/research` in a Tribits examples directory causes the headers to not be found externally, which would cause issues with any potential applications in the future that use the regionMG code. Therefore, I'll propose a slightly different approach:
 1. Change `muelu/research` from using `TRIBITS_ADD_EXAMPLE_DIRECTORIES` to `ADD_SUBDIRECTORY`
 2. Change each individual's research folder to `TRIBITS_ADD_EXAMPLE_DIRECTORIES` except for `muelu/research/regionMG`, which will still use the `ADD_SUBDIRECTORY`, but it will remain guarded by the experimental flags of course.

I'll do some testing on my end to confirm this proposed change doesn't break anything before I push them.

bartlettroscoe:
Does this work?  I guess PR testing will not really test this since the experimental flag will not be set.

lucbv (approved):
This looks good to me, I like that both the `test` and `example` subdirectories in regionMG are both treated as `EXAMPLE_DIRECTORIES`

PR #8161, opened 2020-10-08
Phalanx: switch from device to memory and execution spaces

rppawlo:
Phalanx no longer uses the default memory space from
Kokkos::Cuda. This allows for Phalanx to be UVM free even if users
enable UVM as the default space for cuda at configure time.

@trilinos/<teamName>

## Motivation

Phalanx no longer uses the default memory space form the Kokkos::Cuda object. This default was set at configure time and could be toggled on Cuda nodes between UVM and non-UVM allocations. This change allows phalanx to execute UVM-free regardless of whether UVM was configured as the default space for Cuda.

## Testing

Tested on cuda builds with both UVM enabled and disabled at configure time. This change touches just about every file in phalanx and all tests exercide this.


rppawlo:
@etphipp - I'll ping you when it is ready. I am trying to make the changes backwards compatible with all the apps using phalanx, but turns out I am not quite there yet.

lucbv:
@rppawlo this seems to be in contradiction with what is being done in PR #8247 which promotes using `device` instead of `execution_space` and `memory_space`.
I think maybe there should be a technical discussion on what is the right thing to do in Trilinos and document it somewhere so that we do not end-up with an inconsistent code that will confuse new developers and users.

rppawlo:
> @rppawlo this seems to be in contradiction with what is being done in PR #8247 which promotes using `device` instead of `execution_space` and `memory_space`.

@lucbv - I think this is actually doing the correct thing. The "device" being swapped is PHX::Device which is currently a typedef to the execution space. PHX::Device is being converted to Kokkos::Device. But Kokkos:Device doesn't implement the full execution space API, thus the changes to exec and memory as needed. This is required to make progress on UVM removal and is documented in the UVM removal plan hosted internally at https://snl-wiki.sandia.gov/display/TRIL/UVM+Removal. This is the path that @crtrott recommended. Should we have a quick teams meeting this week to discuss?

lucbv:
@rppawlo 
I assumed that `PHX::device`, `PHX::execution_space` and `PHX::memory_space` are fairly similar to their kokkos counterpart?
Also I am wondering if the changes are still allowing users to have `CudaUVMSpace` or do they need to first change the memory space of their objects before passing them to Phalanx? (Admittedly I don't know what kind of input Phalanx uses...)

The MueLu team stand-up is this morning, I will discuss this in our team, maybe I am not aware that information was already shared with us regarding how to remove UVM. Also I cannot access that wiki-page, I'm guessing I need to be invited to be able to see it.

rppawlo:
Prior to this PR, the PHX::Device was not a Kokkos::Device but an execution space. Phalanx was converted to kokkos early on before the execution and memory spaces were separate objects in kokkos. Kokkos::Device is just two typedefs - one for exec space and one for mem space. It doesn't support the full API for an execution space, so all the work in this pr is to update the use of PHX::Device since it is no longer an exec space. 

The UVM conversion path can be handled in a few ways and it is really up to the package. MueLu and the linear solver stack are keying off the KOKKOS UVM configure flag which sets the default memory space on the CUDA execution space. Phalanx is forcing use of device type explicitly. The choice is affected by timelines, backwards compatibility constraints, and developer constraints. This change will remove UVM completely from phalanx. It does not impact interoperability with the linear solver stack. I have not pushed this PR yet because all application codes will need to be updated and we want to have PRs ready in all those codes before pulling the trigger.

lucbv:
That's fine with me, at the moment I am trying to understand what can be done in MueLu based on what others are already doing. Thanks for explaining your approach vs what might happen in linear solvers, it does make things more clear : )

PR #8162, opened 2020-10-08
Geminga: Add CUDA build with UVM disabled

cgcgcg:
@trilinos/muelu 

## Motivation
Add a build with UVM disabled. This build currently fails all over the place..

### DIFF
--- a/cmake/ctest/drivers/geminga/CMakeLists.txt
+++ b/cmake/ctest/drivers/geminga/CMakeLists.txt
@@ -7,7 +7,7 @@ execute_process(COMMAND date +%A OUTPUT_VARIABLE DAYOFWEEK)
 # parallel valgrind tests run every night except Sunday
 # coverage tests run on Saturday night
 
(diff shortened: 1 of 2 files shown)


jhux2 (commented):
It looks this would run multiple times a week ... is that useful right now?

cgcgcg:
Does it hurt? It's nice to see it on the dashboard every day. And the build is probably quite quick anyway :-P

jhux2:
> Does it hurt? It's nice to see it on the dashboard every day. And the build is probably quite quick anyway :-P

Fair enough!

jhux2 (approved):
lgtm

lucbv:
I think it's a good idea too, I am guessing it fails right-away in Tpetra or something like that at the moment?
The effort to remove UVM should get started soon, it will be helpful to have a regular build that tests that : )

lucbv (approved):
Good idea : )

cgcgcg:
Yep. It fails first in Tpetra, but that can be "fixed" in a way that it compiles. It the fails in Amesos2, Belos and Xpetra. Most of these issues don't seem terrible, but that's obviously only results in compiling, but not necessarily working code.

lucbv:
I guess we will take it one step as a time, first compile it, then make it work x)

PR #8169, opened 2020-10-09
Tpetra: Separate the file read from matrix creation in the new reader

seheracer:
@kddevin 

## Motivation

This PR separates the file read from matrix creation in MatrixMarket_TpetraNew.hpp. This separation is needed to minimize the code duplication when the binary reader is added. 

After this PR is merged, I will push another PR that adds `readBinary` function, which is an alternative to `readMatrixMarket`.

## Testing

The example `TpetraCore_NewReaderExample.exe` and my triangle counting driver work correctly.

### DIFF
--- a/packages/tpetra/core/inout/MatrixMarket_TpetraNew.hpp
+++ b/packages/tpetra/core/inout/MatrixMarket_TpetraNew.hpp
@@ -137,37 +137,21 @@ buildDistribution(
   return Teuchos::rcp<basedist_t>(retval);
 }
 
-public:
-
-// This is the default interface.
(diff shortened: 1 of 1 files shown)


kddevin (commented):
@seheracer Do all the tests in tpetra/core/test/inout also still pass?

seheracer:
> @seheracer Do all the tests in tpetra/core/test/inout also still pass?

Yes, all the Tpetra tests are passing.

kddevin (approved):
A good start.  
Eventually, the Tpetra reader should be refactored / renamed to use the more generic reader, but that change isn't needed now.

PR #8170, opened 2020-10-09
Panzer: add option to shift mesh gids above 32-bit int limit for testing

rppawlo:
Added an option to the inline quad mesh generator to shift the element
and node gids in stk to all be greater than the 32-bit int limit. This
allows small meshes to be used to test certain code paths "at scale"
for ordinal type issues.

@trilinos/panzer 

## Motivation

Help debug an issue an application is seeing for large problems.

## Testing

Changed the main_driver/energy steady state problem to exercise this capability.

### DIFF
--- a/packages/panzer/adapters-stk/example/main_driver/energy-ss-blocked.xml
+++ b/packages/panzer/adapters-stk/example/main_driver/energy-ss-blocked.xml
@@ -19,6 +19,7 @@
                 <Parameter name="Y0" type="double" value="0.0" />
                 <Parameter name="Xf" type="double" value="1.0" />
                 <Parameter name="Yf" type="double" value="1.0" />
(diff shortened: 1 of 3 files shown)


etphipp (approved):

dridzal:
@rppawlo , I'm wondering if the GIDs should be shifted by two times (or more, say 10 times) the number you used, to prevent any side effects from the **unsigned** int limit.  Just to be safe.  If someone somehow managed to consistently use unsigned int in this context, by design or by mistake, then the current shift would be insufficient to uncover mistakes.

rppawlo:
@dridzal - yes - there is a mismatch. Trilinos does all of it's testing with signed global ordinals. I don't know if unsigned even works anymore. stk hard codes the EntityId to uint64_t. We should probably have a check in panzer to make sure that the gids from stk don't pass the Tpetra global ordinal limit.

PR #8175, opened 2020-10-09
MueLu: Don't throw when rebalancing is enabled in serial run, just warn

cgcgcg:
@trilinos/muelu 

## Motivation
Instead of throwing when rebalancing is enabled in a serial run, just warn about it.
This allows running XMLs that have rebalancing enabled in serial.


jhux2 (approved):
Seems reasonable.

PR #8182, opened 2020-10-12
Tpetra: Allow configuration without UVM

cgcgcg:
@trilinos/tpetra 

## Motivation
Allow to configure Tpetra without UVM by setting `Kokkos_ENABLE_CUDA_UVM` and `Tpetra_ENABLE_NO_UVM`.
This is only so that we can start working on removing UVM. Building without UVM does currently not work.


tasmith4:
@cgcgcg on an initial glance I'd prefer to configure with `-DTpetra_ENABLE_UVM=OFF` vs `-DTpetra_ENABLE_NO_UVM=ON`, although I'm open to a persuasive argument otherwise ...

If your PR doesn't already do this, I also think we should warn at configure time when UVM is disabled that Tpetra is not guaranteed to work.

cgcgcg:
@tasmith4 How about this?

tasmith4:
@cgcgcg I like it, thanks!

kddevin (approved):

Issue #8177, opened 2020-10-10
cmake build with fetchcontent


Char-Aznable:
## Question
@trilinos/framework
Is there a way to build some of the packages as dependency of my app using CMake's fetchcontent function? The motivation is to avoid the complication of managing different versions preinstalled libraries so that the users of my app don't have to know how to build and install Trilinos

Char-Aznable:
@trilinos/framework

jhux2:
@trilinos/framework @bartlettroscoe

bartlettroscoe:
@Char-Aznable, I am not familiar with the [CMake FetchContent modulel](https://cmake.org/cmake/help/v3.18/module/FetchContent.html).  I don't know anyone really using [CMake ExternalProject modulel](https://cmake.org/cmake/help/v3.18/module/ExternalProject.html#module:ExternalProject) to build Trilinos.

Seems like the powers that be are pushing [Spack](https://spack.io/) but don't know if the Trilinos package there would meet your needs or not.

Char-Aznable:
@bartlettroscoe FetchContent allows the user to set up dependency on a fetched source at config time (as opposed to build time like ExternalProject). `Kokkos` is already set up to be used with fetchcontent and I have been using Kokkos this way to allow the user to configure building my app with a host or device Kokkos parallel backend as needed. This allows the user to focus on using the dependent app without having to know how to build Kokkos with different backends. 

`Spack` is great but it takes a while to get familiar with. Practically speaking, not every user  wants to invest their time to learn how to use `Spack`. `Spack` also has its shortcoming in terms of how it propagates the compiler flags, e.g., it will rebuild the entire chain of software if you want to just add a debug flag to building Trilinos. `Spack` is also very slow on NFS.

It's not too complicated to set up to use fetchcontent. I found this minimal patch to allow Trilinos to be built from fetchcontent:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 97cc402c6ac..105ea02c7ea 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.10.0 FATAL_ERROR)
 
 # Must set the project name as a variable at very beginning before including anything else
 # We set the project name in a separate file so CTest scripts can use it.
-INCLUDE(${CMAKE_SOURCE_DIR}/ProjectName.cmake)
[...]

PR #8184, opened 2020-10-12
Use CMAKE_CURRENT_LIST_DIR instead of CMAKE_SOURCE_DIR or CMAKE_CURRENT_SOURCE_DIR

Char-Aznable:
@trilinos/framework
@bartlettroscoe

## Motivation

This allows Trilinos to be added as a dependent cmake fetchcontent at config time.
See https://github.com/trilinos/Trilinos/issues/8177 for a detailed discussion

## Related Issues
* Follows https://github.com/trilinos/Trilinos/issues/8177


bartlettroscoe (approved):
This should always be correct in every use case that I can think of so I don't see why not to approve.

Issue #8173, opened 2020-10-09
MueLu: Have single proc tests run in MPI and non-MPI builds


bartlettroscoe:
## Enhancement
@cgcgcg 

Following the conversation in PR #8115, it would be good for MueLu single proc tests that ran in non-MPI (i.e. serial) builds to run in non-MPI and MPI builds.   This way, the set of tests that run in a non-MPI build will be a strict subset of the tests that run in an MPI build (and the MPI build will run a supersets of tests so just doing an MPI build should provide pretty good confidence that a non-MPI build will also pass all the MueLu tests).

The change in PR #8115 had the side-effect of removing several single proc tests in MueLu in a non-MPI build due to a change in backward compatibility addressing the issue #8110 the ripple effect that had as described in #8115.

bartlettroscoe:
As discussed in #8115, I am going to write the function `MUELU_ADD_SERIAL_AND_MPI_TEST()` that will result in single-proc and multi-proc tests in an MPI build and just a single-proc test in a non-MPI build.  This will restore all of single-proc tests for a non-MPI build and will also add the single-proc tests for the MPI build (so the tests run in an MPI build are a strict superset of the tests run in an non-MPI build).

jhux2:
@bartlettroscoe Thanks for fixing this.  Btw, I think other packages that MueLu would want to do this, i.e.., build with MPI but run single-process tests.  So I think it might be better to give the macro a more general name, i.e., `TRILINOS_ADD_SERIAL_AND_TEST`.

jhux2:
@trilinos/muelu

bartlettroscoe:
>  Btw, I think other packages that MueLu would want to do this, i.e.., build with MPI but run single-process tests.

@cgcgcg, let's start with MueLu and then we can move this to a central location in Trilinos for re-use.

Note that other packages like Ifpack2 expected just the opposite.  They were defining tests like:

```
TRIBITS_ADD_TEST(
  tif_belos
  NAME Jacobi_hb_belos
  ARGS "--xml_file=test_Jacobi_nos1_hb.xml"
  COMM serial mpi
  NUM_MPI_PROCS 1
  STANDARD_PASS_OUTPUT
)

TRIBITS_ADD_TEST(
  tif_belos
  NAME Jacobi_hb_belos
  ARGS "--xml_file=test_2_Jacobi_nos1_hb.xml"
  COMM serial mpi NUM_MPI_PROCS 2
[...]

PR #8115, opened 2020-09-28
Switch to add_test(NAME ... COMMAND ...) and remove duplicate tests (#8110)

bartlettroscoe:
This allows using generator expressions with test properties (see #8110).  This also no longer allows duplicate tests.

I had to remove a duplicate MueLu test.

In addition to bringing in TriBITS changes for this from https://github.com/TriBITSPub/TriBITS/pull/333 and https://github.com/TriBITSPub/TriBITS/pull/334, this  also brings in TriBITS changes from https://github.com/bartlettroscoe/TriBITS/commit/2850245d94de4dc12a7a3971876bb2e823d26810.

## How was this tested?

This passed the TriBITS test suite (see https://github.com/TriBITSPub/TriBITS/commit/6a3c841be15b7e091ece6cbce31910583adaf7d6) and I tested a Trilinos configure with:

```
$ ./checkin-test-sems.sh --enable-all-packages=on --allow-no-pull --configure
```

which yielded:

```
PASSED (NOT READY TO PUSH): Trilinos: crf450.srn.sandia.gov

Mon Sep 28 13:07:50 MDT 2020

Enabled Packages: 
Enabled all Packages

Build test results:
-------------------
0) MPI_RELEASE_DEBUG_SHARED_PT_OPENMP => passed: configure-only passed => Not ready to push! (2.96 min)

*** Commits for repo :

0) MPI_RELEASE_DEBUG_SHARED_PT_OPENMP Results:
----------------------------------------------

  passed: Trilinos/MPI_RELEASE_DEBUG_SHARED_PT_OPENMP: configure-only passed
  
  Mon Sep 28 13:07:50 MDT 2020
  
  Enabled Packages: 
  Enabled all Packages
  Hostname: crf450.srn.sandia.gov
  Source Dir: /home/rabartl/Trilinos.base/Trilinos/cmake/tribits/ci_support/../../..
  Build Dir: /home/rabartl/Trilinos.base/BUILDS/CHECKIN/MPI_RELEASE_DEBUG_SHARED_PT_OPENMP
  
[...]

PR #8189, opened 2020-10-13
Automatic snapshot commit from seacas at 90e52d1

gsjaardema:

At commit:

commit 90e52d17b523906420408d144657000112010097
Date:    Tue Oct 13 09:43:45 2020 -0600
Summary: IOSS: glob - update comment [ci skip]

* Add nodemap/elemmap increment command to help test files with large node/element ids
* Add 'globbing' support to io_modify
* Add 'version()' and _VERSION to aprepro so can change execution based on version
* Fix exodiff handling of partial mappings
* Update to catch-2.13.1
* Better error output for bad exodus file open/create
* Fix CGNS zgc handling of overlap with reversals
* Fix IOSS/exodus-par handling of proc-files with no nodes/elements

@trilinos/<teamName>

### DIFF
--- a/packages/seacas/libraries/ioss/src/utest/CMakeLists.txt
+++ b/packages/seacas/libraries/ioss/src/utest/CMakeLists.txt
@@ -29,8 +29,7 @@ TRIBITS_ADD_EXECUTABLE(
 TRIBITS_ADD_TEST(
         Utst_blob_write_read
 	NAME Utst_blob_write_read
-	NUM_MPI_PROCS ${NPROCS}
-	COMM mpi serial
+	NUM_MPI_PROCS 1
 	XHOSTTYPE Windows
   )
 
--- a/packages/seacas/libraries/ioss/src/utest/CMakeLists.txt
+++ b/packages/seacas/libraries/ioss/src/utest/CMakeLists.txt
@@ -29,7 +29,9 @@ TRIBITS_ADD_EXECUTABLE(
 TRIBITS_ADD_TEST(
         Utst_blob_write_read
 	NAME Utst_blob_write_read
-	NUM_MPI_PROCS 1
+	NUM_MPI_PROCS ${NPROCS}
+	COMM mpi serial
+	XHOSTTYPE Windows
   )
 
 TRIBITS_ADD_EXECUTABLE(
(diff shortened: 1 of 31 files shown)


packages/seacas/libraries/ioss/src/utest/CMakeLists.txt
```diff
@@ -29,7 +29,9 @@ TRIBITS_ADD_EXECUTABLE(
 TRIBITS_ADD_TEST(
         Utst_blob_write_read
 	NAME Utst_blob_write_read
-	NUM_MPI_PROCS 1
+	NUM_MPI_PROCS ${NPROCS}
+	COMM mpi serial
```
bartlettroscoe, line 33: @gsjaardema, note, new behavior of TriBITS as for PR #8115 is to skip a test from the test suite non-MPI (i.e. "serial") builds if `NUM_MPI_PROCS` is greater than 1.  If you want to run a 1-proc version of this test an a multi-proc version, you should add those as two separate tests, like we did with MueLu in PR #8173.  (NOTE: If there is interest in using that approach, that function could be added to TriBITS proper so SEACS could reuse it.  But if this pattern in rare in SEACAS, it is not worth it.  Just define two separate tests.)

bartlettroscoe (dismissed):
I did not review all of the code, just the CMake code.  Note that this change looks like a test may disappear from a serial (non-MPI) build (see my comment below).

Otherwise, this is a standard SEACAS snapshot into Trilinos so approval is about automatic.

packages/seacas/libraries/ioss/src/utest/CMakeLists.txt
gsjaardema: Actually, I think that this change was not wanted since for some reason this test fails in Trilinos builds on certain platforms, but doesn't fail in SEACAS...

packages/seacas/libraries/ioss/src/utest/CMakeLists.txt
```diff
@@ -29,8 +29,7 @@ TRIBITS_ADD_EXECUTABLE(
 TRIBITS_ADD_TEST(
         Utst_blob_write_read
 	NAME Utst_blob_write_read
-	NUM_MPI_PROCS ${NPROCS}
-	COMM mpi serial
+	NUM_MPI_PROCS 1
```
bartlettroscoe, line 32: @gsjaardema, FYI, leaving out the COMM argument is the same as `COMM serial mpi`.

bartlettroscoe (approved):
See my last review

PR #8193, opened 2020-10-13
MueLu: regionMG added use of rebalance heuristic factory

pohm01:
To determine number of ranks for coarse rebalancing. Can still specify manually as well.
Also timers for region coarse level rebalance.

@trilinos/muelu 

## Motivation
No longer have to manually specify number of ranks for coarse level rebalancing for regionMG

## Testing
Covered by the rebalancing tests in regionMG.

### DIFF
--- a/packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
+++ b/packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
@@ -65,6 +65,7 @@
 
 #if defined(HAVE_MUELU_ZOLTAN2) && defined(HAVE_MPI)
 #include <MueLu_RepartitionFactory.hpp>
+#include <MueLu_RepartitionHeuristicFactory.hpp>
 #include <MueLu_Zoltan2Interface.hpp>
 #endif
 
@@ -475,12 +476,13 @@ void RebalanceCoarseCompositeOperator(const int maxRegPerProc,
 {
 #include "MueLu_UseShortNames.hpp"
   using CoordType = typename Teuchos::ScalarTraits<Scalar>::coordinateType;
+  using Teuchos::TimeMonitor;
+
+  RCP<TimeMonitor> tm = rcp(new TimeMonitor(*TimeMonitor::getNewTimer("ReblanceCoarseCompositeOperator: ")));
 
   RCP<Teuchos::FancyOStream> fos = Teuchos::fancyOStream(Teuchos::rcpFromRef(std::cout));
   fos->setOutputToRootOnly(0);
   *fos << "Rebalancing coarse composite operator." << std::endl;
-
-  //RCP<const Map> map = coarseCompOp->getMap();
 
   const int numPartitions = rebalanceNumPartitions;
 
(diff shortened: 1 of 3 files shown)


packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
```diff
@@ -475,39 +476,48 @@ void RebalanceCoarseCompositeOperator(const int maxRegPerProc,
...
   *fos << "Rebalancing coarse composite operator." << std::endl;
 
-  //RCP<const Map> map = coarseCompOp->getMap();
-
   const int numPartitions = rebalanceNumPartitions;
 
   // We build a fake level
   Level level;
-  level.SetLevelID(1);
+  level.SetLevelID(3);
```
cgcgcg, line 491: Huh? Why?

GrahamBenHarper, line 491: Is there any significance of level id 1 vs level id 3 since it's a fake level?

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
```diff
@@ -475,39 +476,48 @@ void RebalanceCoarseCompositeOperator(const int maxRegPerProc,
...
   level.Set<RCP<MultiVector> >("Coordinates",          compCoarseCoordinates);
-  level.Set<int>              ("number of partitions", numPartitions);
 
   RCP<RepartitionFactory> repart = rcp(new RepartitionFactory());
   Teuchos::ParameterList paramList;
   paramList.set("repartition: remap parts",       false);
+  if( numPartitions > 0 ){ // If number of coarse rebalance partitions was provided by the user.
+    level.Set<int>              ("number of partitions", numPartitions);
+  } else {
+    RCP<RepartitionHeuristicFactory> repartHeuristic = rcp(new RepartitionHeuristicFactory());
```
cgcgcg, line 509: Don't you need to pass in parameters somewhere?

GrahamBenHarper (approved):
I just have a question about the fake level ids, but other than that it looks great.

I like the idea of using a heuristic if we request rebalancing but don't specify the number of partitions.

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
pohm01: How do I do that? Or which parameter list would that be?

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
GrahamBenHarper: ah I didn't even see your review while I was writing mine :stuck_out_tongue: I had the same question.

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
cgcgcg: Did you check that you can actually set the parameters that the heuristic factory takes? Cause otherwise, you'll just be running the defaults all the time.

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
pohm01: It is to start past the default repartition start level.

cgcgcg: That's a bit hacky, but ok. Maybe check that this doesn't appear anywhere in the output, since that could become quite confusing..

packages/muelu/research/regionMG/src/SetupRegionHierarchy_def.hpp
cgcgcg: You could have a look at how the `ParameterListInterpreter` handles the parameters.

GrahamBenHarper (dismissed):
lgtm!

GrahamBenHarper (approved):

PR #8204, opened 2020-10-15
8200 Trilinos with templated stack on Windows machines

jrobcary:
@trilinos/kokkos-kernels @trilinos/intrepid2 @trilinos/zoltan2

## Motivation
The goal of this pull request is to fix some compilation failures on Windows.  This is a work in progress, as the build on Windows is not yet working.  I was encouraged to submit this anyway to get some feedback as to whether I am going in the right direction.

## Testing

Is there a way to know which test corresponds to some source file?  In any case, the changes are all in #if blocks, so not likely to affect any tests run on Linux.

### DIFF
--- a/packages/amesos2/src/Amesos2_Superlu_def.hpp
+++ b/packages/amesos2/src/Amesos2_Superlu_def.hpp
@@ -752,6 +752,7 @@ Superlu<Matrix,Vector>::getValidParameters_impl() const
 
     // ILU parameters
 
+#if 0
     setStringToIntegralParameter<SLU::rowperm_t>("RowPerm", "LargeDiag",
             "Type of row permutation strategy to use",
             tuple<string>("NOROWPERM","LargeDiag","MY_PERMR"),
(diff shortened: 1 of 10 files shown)


kddevin (commented):
The zoltan2 changes are OK with me.  I cannot review the other packages.

Since KokkosKernels is a snap-shot into Trilinos, KokkosKernels changes might have to be coordinated with the Kokkos github repo.

packages/amesos2/src/Amesos2_Superlu_def.hpp
```diff
@@ -752,6 +752,7 @@ Superlu<Matrix,Vector>::getValidParameters_impl() const
 
     // ILU parameters
 
+#if 0
```
srajama1, line 755: This "if 0" worries me. We have to support older SuperLU and I am not sure this has been removed everywhere.

srajama1 (commented):
Please see my comment on Amesos2. Kokkos Kernels changes look ok to me, but we have no way to test it.

Can we get a corresponding PR on the kokkoskernel  repo as what lives in Trilinos is a snapshot of that ?

srajama1:
Here is the repo for Kokkos Kernels https://github.com/kokkos/kokkos-kernels

packages/kokkos/core/src/impl/Kokkos_Atomic_Compare_Exchange_Strong.hpp
```diff
@@ -328,11 +328,13 @@ atomic_compare_exchange(volatile T* const, const Kokkos::Impl::identity_t<T>,
 }
 #endif
 
+#ifndef _WIN32
```
crtrott, line 331: What is the config here and why do we need this change? We are currently testing Kokkos with Visual Studio during PR testing and it passes. I'd prefer if we could fix this in Kokkos first and make sure we support the config.

packages/kokkos/core/src/impl/Kokkos_ChaseLev.hpp
```diff
@@ -190,7 +190,8 @@ struct ChaseLevDeque {
       return_value = *static_cast<T*>(a[b]);  // relaxed load
       if (t == b) {
         /* single last element in the queue. */
-#ifdef _WIN32
+// #ifdef _WIN32
+#if 0 // clang-cl says atomic_compare_exchange_strong is ambiguous.
```
crtrott, line 194: Again we are testing with visual studio during PR testing. I suppose we might not test clang-cl but the non-clang version of VS compiler but that one is a compiler we support. Does this change break that?

crtrott (changes requested):
We need to introduce the Kokkos changes first in Kokkos proper, and I am not confident they are kosher. We are testing Visual Studio as part of our PR testing and it works. My suspicion is that these changes here come from using the clang frontend in VS. Generally we need more information. I.e. what is the configuration here (the visual studio configuration) etc. .

kddevin:
Despite the name of this PR, no Tpetra files were changed by the PR.  I will change the name to something more appropriate.
@ccober6 @jwillenbring  Perhaps this PR has some info relevant to today's SART discussion.

jrobcary:
This project is on hold while dealing with other priorities.  Will get 
back to it over the next few months.

PR #8208, opened 2020-10-16
ROL: solve issues related to falling checks when HAVE_ROL_DEBUG is defined

kliegeois:
@trilinos/rol 
@mperego 

## Motivation

The initial goal of this PR was to configure the definition of the HAVE_ROL_DEBUG macro using cmake.
After that, when running ctest with HAVE_ROL_DEBUG defined, some issues in Piro_PerformAnalysis with Thyra adapters have been observed.

It has been observed that, in at least some optimization algorithm implementations, the update function is not always called before the computation of the Hessian or the gradient when either x (called u in ROL) or p (called z in ROL) has been updated.
There were two options to overcome this:
1. To revisit ROL functions and add update if required,
2. To add a check at the beginning of the Thyra adapters to test if x or p has changed, if it is the case, the values are updated.

It is the latest solution which is implemented in this PR.

## Testing

I tested the branch on Blake and I tested Albany tests built against this branch.

### DIFF
--- a/packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
+++ b/packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
@@ -524,7 +524,14 @@ public:
     TEUCHOS_ASSERT(Teuchos::nonnull(lows_factory));
     Teuchos::RCP< Thyra::LinearOpBase<double> > lop;
 
-    if(computeJacobian1)
+    bool explicitlyTransposMatrix = false;
+    if(params != Teuchos::null) {
+      explicitlyTransposMatrix = params->get("Enable Explicit Matrix Transpose", false);
+      if(explicitlyTransposMatrix)
+        params->set("Compute Transposed Jacobian", true);
+    }
+
+    if(computeJacobian1 || explicitlyTransposMatrix)
       lop = thyra_model.create_W_op();
     else {
       if(verbosityLevel >= Teuchos::VERB_HIGH)
@@ -544,14 +551,14 @@ public:
     }
     const Teuchos::RCP<Thyra::LinearOpWithSolveBase<Real> > jacobian = lows_factory->createOp();
 
-    if(computeJacobian1)
+    if(computeJacobian1 || explicitlyTransposMatrix)
     {
       outArgs.set_W_op(lop);
       thyra_model.evalModel(inArgs, outArgs);
       outArgs.set_W_op(Teuchos::null);
       jac1 = lop;
 
-      computeJacobian1 = false;
+      computeJacobian1 = explicitlyTransposMatrix;
     }
 
     if (Teuchos::nonnull(prec_factory))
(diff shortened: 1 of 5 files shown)


packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Constraint_SimOpt.hpp
```diff
@@ -89,11 +89,9 @@ class ThyraProductME_Constraint_SimOpt : public Constraint_SimOpt<Real> {
   }
 
   void value(Vector<Real> &c, const Vector<Real> &u, const Vector<Real> &z, Real &tol) {
-#ifdef  HAVE_ROL_DEBUG
-    //u and z should be updated in the update functions before calling value
-    TEUCHOS_ASSERT(!u_hasChanged(u));
-    TEUCHOS_ASSERT(!z_hasChanged(z));
-#endif
+
```
mperego, line 92: I thought more about this, and I would prefer to keep the check only in debug, as before, and update where needed in ROL. Do you mind trying to do that? Hopefully you don't have to do that in many places. Same thing in the other functions/classes. The policy is: "if a function changes u or z, then it should call the update functions".

mperego (changes requested):

mperego:
@kliegeois I think I fixed the issue with the update, modifying a function in ROL. Can you try it out?
I also added the modifications I had mentioned.

mperego (approved):

kliegeois:
@mperego , I have just tested your modifications and they worked both for Trilinos and Albany tests. Thanks a lot!
Reading your commit messages, it seems that you wanted to rebase the commits before merging them. Can I let you do so?

mperego:
> 
> 
> @mperego , I have just tested your modifications and they worked both for Trilinos and Albany tests. Thanks a lot!
> Reading your commit messages, it seems that you wanted to rebase the commits before merging them. Can I let you do so?

Great. Sure no problem.

mperego:
@dridzal @dpkouri Can you have a quick look at the changes in ROL? The option ROL_HAVE_DEBUG has been defined, and we added a couple of calls to the objective `update` function in the Trust Region `update` function.

kliegeois:
> > @mperego , I have just tested your modifications and they worked both for Trilinos and Albany tests. Thanks a lot!
> > Reading your commit messages, it seems that you wanted to rebase the commits before merging them. Can I let you do so?
> 
> Great. Sure no problem.

@mperego, thanks!

mperego:
@kliegeois I talked to Drew. He didn't like one of the changes in ROL. So for now I decided to disable the check for the Hessian. In fact we do not store any information at the moment to compute the AD Hessian, so there is no need to call update. For the FD Hessian we already call the needed update functions. We'll look more carefully into this once ROL pushes the changes to the update function. Let me know if this is OK with you.

kliegeois:
@mperego thanks for your comment. It is definitely fine for me. It is true that, when computing the Hessian-vector products with AD, the update is not required.

mperego (approved):

PR #8212, opened 2020-10-16
Tpetra: Bring KokkosWrapperNodes from Teuchos to TpetraClassic

tasmith4:
@trilinos/teuchos @trilinos/tpetra 

## Motivation
The wrapper nodes in `Kokkos::Compat` seem intended only for use in Tpetra, so it makes more sense for them to live in Tpetra.

## Testing
Expect existing tests to continue to pass.


kddevin (approved):
Thanks, @tasmith4

PR #8232, opened 2020-10-20
Fixes BiCGStab algorithm so that it detects breakdown

hkthorn:
@trilinos/belos 

## Motivation

There are three coefficient calculations in this implementation
of the BiCGStab algorithm that need to be checked to see if the
denominator is less than the sfmin for that scalar type.  These
checks have been added to the algorithm and logic has been added
to the solver manager to check for breakdown if the iteration
exits before convergence or the maximum number of iterations.
There is now a test that solves an inner-outer linear system
with diagonal matrices that checks for one type of breakdown
(the good kind).  The solver has been checked for the other types
of breakdown.

## Testing

This has been tested in serial and parallel using GCC 4.9.2 and OpenMPI 1.10


jhux2 (commented):
@hkthorn Thanks for adding these guards.  Is there a Tpetra test?  I see what looks like an Epetra-specific one.

hkthorn:
@jhux2 In general it has been observed that the Tpetra tests are lacking for all the solvers.  I would like to resolve that in a separate issue, if that's ok?

jhux2 (approved):
@hkthorn Looks good to me, thank you!

PR #8226, opened 2020-10-19
Allow activating kokkos TPLS in Tribits, fix libdl issues

jjwilke:
@trilinos/kokkos

## Motivation

Standalone Kokkos and Tribits do not currently have compatible TPL systems.
This attempts to bride the gap between them and adds the first "shared TPL: libdl
It also fixes with profiling tests that are now able to be activated in Trilinos.

## Testing

All unit tests in Kokkos will have profiling/libdl activated on them now.

PR #8233, opened 2020-10-21
allow activating kokkos TPLS in Tribits, fix libdl issues

jjwilke:
Clone of PR #8226 to reset build system


jjwilke:
@trilinos/framework @bartlettroscoe 

I'm struggling to understand the behavior of the autotester. It appears that anytime I push a build system fix, the changes do not appear in the next run of the autotester.  Is the autotester not doing a clean build? Or is there a way to force it to do a clean build?

jwillenbring:
@jjwilke 
> Or is there a way to force it to do a clean build?

Yes, there is a way. I will force a clean build. We have also run into some situations where pushing additional changes to a branch has caused issues too. @prwolfe may know more about the details of these situations. For now, I will see if a clean build helps.

jwillenbring:
I ran the reset job and set the retest flag.

bartlettroscoe:
@jjwilke, why is a build-from-scratch required?  Why does not a rebuild work?

It will be very unfortunate if the Kokkos build system break the ability to cleanly rebuild Trilinos.  (That will massively tax the PR and nightly builds if they have to build from scratch each time.)

jjwilke:
A clean rebuild should not be required.  If I make build system changes outside the autotester, the bugs go away even without a clean rebuild. Every time I push a change to CMake files, the autotester keeps showing the same build system bugs even though they've been fixed. I really don't understand the behavior.

I expect there will be cases, though, where a build system mistake leaves the cache in a "erroneous" state and needs to be cleared out.

jjwilke:
Another note: is there a way to see the actual bash scripts (or similar) that are actually executing? I only see the CDash results, which doesn't really give me enough information to replicate what's going on.  I really want to see the module list and the exact CMake command executed.

jjwilke:
@bartlettroscoe Yeah, so it did it, again on ascic143. I am getting the error that was occurring on commit cc5eb06df2f7d2b8529d785679828c2c711dce49. This error is fixed in a13091be998250822e73f29863d498f0a0ea95f5  - which I know because the line numbers in the CMake backtrace match the old commit, not the new commit.  It's like it's ignoring the new commit.

bartlettroscoe:
> It's like it's ignoring the new commit.

@jjwilke, someone on the @trilinos/framework team will need to verify that.  The problem from the TriBITS side is that the configure will only report the top commit that is created locally and it does not show the ancestors.  You just have to take the autotester's word for what the two ancestor commits are that get merged locally.

prwolfe:
> @bartlettroscoe Yeah, so it did it again on ascic143. I am getting the error that was occurring on commit [cc5eb06](https://github.com/trilinos/Trilinos/commit/cc5eb06df2f7d2b8529d785679828c2c711dce49). This error is fixed in [a13091b](https://github.com/trilinos/Trilinos/commit/a13091be998250822e73f29863d498f0a0ea95f5) - which I know because the line numbers in the CMake backtrace on CDash match the old commit, not the new commit. It's like it's ignoring the new commit.

As far as I can tell none of the last set of jobs ran on ascic143 - I think you may be viewing old results. Try looking at https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-8233-test&field2=buildstarttime&compare2=83&value2=Oct%2021%2021%3A27 for just the last build. 

That has only a ride test failure KokkosCore_ProfilingTestLibraryLoad_MPI_1

jjwilke:
Yup, that's the issue. So weird. I was sorting the CDash tests by most recent, but the old tests were still on the first page. I didn't even look at the test time column.

Okay, this restores sanity. I might be an idiot who can't sort by time - but I'm a sane idiot now.

ndellingwood (approved):

ndellingwood:
Thanks @jjwilke !! Merging in now, should add reference to this PR in matching Kokkos PR.

jjwilke:
Yeah, sorry, let's go ahead and merge this. I'll make sure these changes make it into kokkos master/devel before the next promotion.

PR #8236, opened 2020-10-21
Geminga: Test more packages in no-UVM build

cgcgcg:
@trilinos/muelu


lucbv (approved):
Yeah, I think having these would cover our needs and ATDM a bit better : )
So we should have a more detailed view of how things are in Tpetra with this?

cgcgcg:
As of last night, Tpetra builds in this build. We now can look at the build errors from Xpetra & MueLu. Tpetra tests are failing all over the place, see #8209.

jhux2:
@cgcgcg I don't remember the name, but there's an option to control whether to do incremental builds or to do a clean build with package XYZ disabled if XYZ won't build.  The former option might be a little more appropriate here, since I don't think we want to just avoid packages that won't build with UVM disabled.

Issue #8238, opened 2020-10-21
Amesos2: Build fails with UVM off


cgcgcg:
## Bug Report
@trilinos/amesos2

### Description
Disabling UVM makes builds with Amesos2 fail:
https://testing.sandia.gov/cdash/viewBuildError.php?buildid=8102668
It appears that 
https://github.com/trilinos/Trilinos/blob/9c8e898171290e3fcc7fdb08379d55141ac4d6ae/packages/amesos2/src/Amesos2_Kokkos_Impl.hpp#L116-L133
instantiates the same solver twice.
Commenting out 
https://github.com/trilinos/Trilinos/blob/9c8e898171290e3fcc7fdb08379d55141ac4d6ae/packages/amesos2/src/Amesos2_Kokkos_Impl.hpp#L123
makes the build pass.

cgcgcg:
FYI @tasmith4

PR #8239, opened 2020-10-21
Amesos2: Do not instantiate twice when UVM off

cgcgcg:
@trilinos/amesos2

## Motivation
When UVM is disabled we don't need to instantiate twice.

#8238


ndellingwood (dismissed):
LGTM, thanks @cgcgcg

ndellingwood (approved):

PR #8242, opened 2020-10-22
MueLu: Print stack traces

cgcgcg:
@trilinos/muelu 

## Motivation
Add printing of stack traces to MueLu test ETI. (Requires TPL BinUtils.)

Use it in the no-UVM nightly build to get spots with inaccessible memory spaces.


GrahamBenHarper (approved):
lgtm!

PR #8243, opened 2020-10-22
Teuchos: Add option to print stacktrace to CommandLineProcessor

cgcgcg:
@trilinos/teuchos 

## Motivation
Add an option to the `CommandLineProcessor` to print a stacktrace when a segfault is encountered.


bartlettroscoe (commented):
@cgcgcg, I know a lot of code in this set is not unit tested (because it was written pre-2007) but can we add some type of test that checks that this actually works?  I think there is already some stacktrack-related test code in Teuchos that you could look and just copy and paste most of it.

Issue #1088, opened 2017-02-23
Tpetra::MultiVector: Don't use CudaUVMSpace for internal comm buffers


mhoemmen:
@trilinos/tpetra 

Blocked by: #1571, #1602, #1658

MPI on Pascal, not K80, appears to be slow when accessing UVM-allocated device buffers, but not standard CUDA device buffers.  This causes trouble with Tpetra::Distributor sending and receiving packed data, as measured by @ambrad .  It also probably causes trouble with dot product and norm.

@crtrott and I wrote an MPI + Kokkos benchmark to measure this independently of Tpetra.

Possible work-around: Don't use CudaUVMSpace for internal comm buffers in DistObject methods packAndPrepareNew and unpackAndCombineNew.  Define a typedef pack_dev_memory_space which is CudaSpace when execution_space is Cuda.

mhoemmen:
Inside Tpetra::DistObject: 
```
#ifdef KOKKOS_HAVE_CUDA
typedef typename std::conditional<std::is_same<typename device_type::execution_space, Kokkos::Cuda>::value, Kokkos::CudaSpace, typename device_type::memory_space>::type pack_memory_space;
#else
typedef typename device_type::memory_space pack_memory_space;
#endif // KOKKOS_HAVE_CUDA
```

crtrott:
I am on it, was thinking I can get that in one batch through with the testing for KokkosKernels.

mhoemmen:
> I am on it, was thinking I can get that in one batch through with the testing for KokkosKernels.

one step at a time :) we don't know if this really the issue yet, also we may have to do the same for dot products and i haven't tested whether this _builds_ yet ;-P

crtrott:
Digging around inside Tpetra_Import_Util2.hpp

crtrott:
Looks like it is compiling :-)

crtrott:
And it only broke the CrsMatrix UnitTests and the BlockCrsMatrix UnitTests, but those are not important right ;-)?

crtrott:
This probably means the sync logic is bad inside pack and unpack ...

mhoemmen:
> This probably means the sync logic is bad inside pack and unpack ...

:( note that we've never tested Tpetra on CUDA not with UVM.  On the other hand, pack and unpack for CrsGraph and CrsMatrix should use the old interface that takes Teuchos::Array*.

mhoemmen:
[...]

PR #8247, opened 2020-10-23
Tpetra: Use device_type instead of execution_space for Kokkos types

tasmith4:
@trilinos/tpetra 

## Motivation

Replace `execution_space` with `device_type` in template arguments for Kokkos types.  This is needed for FakeCuda to build, but seems to be the correct behavior no matter what.  Thanks to @brian-kelley who did most of the actual work here.

## Testing

This is a refactoring, so no new tests have been added.


packages/tpetra/core/src/Tpetra_Details_Transfer_decl.hpp
```diff
@@ -75,7 +75,7 @@ class Transfer : public Teuchos::Describable {
 private:
   using execution_space = typename NT::device_type::execution_space;
   using memory_space =
-    ::Tpetra::Details::DefaultTypes::comm_buffer_memory_space<execution_space>;
+    ::Tpetra::Details::DefaultTypes::comm_buffer_memory_space<typename NT::device_type>;
```
lucbv, line 78: Here adding `using device_type = typename NT::device_type` would help promote its use instead of `execution_space`.

packages/tpetra/core/src/Tpetra_ImportExportData_decl.hpp
```diff
@@ -148,7 +148,7 @@ namespace Tpetra {
 
     using execution_space = typename Node::device_type::execution_space;
     using memory_space =
-      ::Tpetra::Details::DefaultTypes::comm_buffer_memory_space<execution_space>;
+      ::Tpetra::Details::DefaultTypes::comm_buffer_memory_space<typename Node::device_type>;
```
lucbv, line 151: Same as above

lucbv (commented):
This looks all good to me but I feel it's more appropriate for @brian-kelley or @kddevin to approve this PR since it uses the automerge flag

packages/tpetra/core/src/Tpetra_Details_Transfer_decl.hpp
tasmith4: @lucbv if you look at the line right below, we already have a `using device_type`, in this case, the `typename NT::device_type` is just a temporary to setup the real device type we want to be used.

packages/tpetra/core/src/Tpetra_ImportExportData_decl.hpp
tasmith4: @lucbv same as above

packages/tpetra/core/src/Tpetra_Details_Transfer_decl.hpp
lucbv: OK but I am a bit curious, what is the different between `typename NT::device_type` and `Kokkos::Device<execution_space, memory_space>`?
I simply don't know so in my mind I thought the type declarations would actually look like this:
```
using device_type = typename NT::device_type;
using execution_space = typename device_type::execution_space;
using memory_space = ::Tpetra::Details::DefaultTypes::comm_buffer_memory_space<device_type>;
```
But I am guessing that Tpetra is doing something special in `::Tpetra::Details::DefaultTypes::comm_buffer_memory_space` that changes the `memory_space` used in `NT::device_type`?

tasmith4: @lucbv see `core/src/Tpetra_Details_DefaultTypes.hpp`, lines 105-117, and #1088.  I wasn't around for this history, but apparently there are performance issues with using CudaUVMSpace to create buffers for communication, so when we are using a Cuda node, this code forces the local `device_type` to use CudaSpace regardless of the default memory space associated with the Cuda execution space.

tasmith4:
@brian-kelley @kddevin are there any changes you'd like to see with this?

PR #8249, opened 2020-10-24
Tpetra: Add binary file read function to the new reader

seheracer:
@kddevin 

## Motivation

This PR adds a binary file read function to the new Tpetra reader. 

Expected format:
`numVertices numEdges source_1 target_1 source_2 target_2 source_3 target_3 ... source_numEdges target_numEdges
`
All entries are expected to have `unsigned int` type except for `numEdges`, which is expected to have `unsigned long long` type.      

## Related PRs
* Follows: https://github.com/trilinos/Trilinos/pull/8169 

## Testing

TpetraCore_NewReaderExample.exe and my triangle counting driver are working correctly.


kddevin (approved):
Thanks, @seheracer 
MatrixMarket format is well defined.  
But we will need documentation for the format of the binary file that this code reads.
It would be OK to add the documentation with the "readBinary" function in the source code.

seheracer:
> Thanks, @seheracer
> MatrixMarket format is well defined.
> But we will need documentation for the format of the binary file that this code reads.
> It would be OK to add the documentation with the "readBinary" function in the source code.

This is a good idea. I will add documentation in the source code.

PR #8252, opened 2020-10-24
Stk snapshot 20201024

alanw0:
@trilinos/<teamName>


alanw0:
Ugh, don't have the cmake support right yet. But scripts on the sierra side are getting closer to providing a pre-push check for correctness.

PR #8254, opened 2020-10-26
Teuchos:  changed declaration of size_ to size_type for ArrayView<const T> (as it is for ArrayView<T>)

kddevin:
@trilinos/teuchos  

Private member size_ of ArrayView of T  has type size_type.
The same member of ArrayView of const T had type int.  Overflows resulted.  
This PR changes size_ to have type size_type for both specializations of ArrayView.

## Motivation

int overflows for big problems

## Testing

Tested on mac with clang.

### DIFF
--- a/packages/teuchos/core/src/Teuchos_ArrayViewDecl.hpp
+++ b/packages/teuchos/core/src/Teuchos_ArrayViewDecl.hpp
@@ -515,7 +515,7 @@ public:
 
 private:
   const T* ptr_;
-  int size_;
+  size_type size_;
 #ifdef HAVE_TEUCHOS_ARRAY_BOUNDSCHECK
   ArrayRCP<const T> arcp_;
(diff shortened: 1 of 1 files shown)


kddevin:
@ibogle

bartlettroscoe (approved):
Yes, this looks like this is the right change to make.  Not sure why we did not get any compiler warnings for this mismatch up to this point.

@kddevin, does this fix a specific issue that some Trilinos developer or customer was having with this code?  I did not see an issue tracker mentioned so I was not sure. 

Anyway, thanks for posting!

kddevin:
Thanks for the review, @bartlettroscoe .  My intern ran into the problem with an application that he was running.  He didn't file an issue because by the time he narrowed the problem down to Teuchos (rather than his code), the solution was simple and obvious.

PR #8255, opened 2020-10-26
MueLu: Add threshold to diagonal fix

cgcgcg:
@trilinos/muelu 

## Motivation
Add parameter 'rap: fix zero diagonals threshold'. Use it in RefMaxwell to fix diagonals that are smaller or equal than eps and not outright zero.


jhux2 (commented):
@cgcgcg Since you're mucking with this anyway, I would recommend adding an additional option, "replacement value", similar to what is done in diagonal lumping.

jhux2 (dismissed):
lgtm

jhux2 (dismissed):
shazam!

mayrmt (approved):

PR #6748, opened 2020-02-04
tpetra: disable kokkos warnings in initialize tests

ndellingwood:
This uses kokkos env variables to disable warnings in a subset of tpetra
unit tests to prevent false failures.

@trilinos/tpetra 

## Motivation
This change is necessary to prevent false test failures in tests that catch warning output that may be unreliable.
In particular, older Macs may be prone to this following changes in Kokkos kokkos/kokkos#2477

## Stakeholder Feedback
These changes were discussed with @mhoemmen and @kddevin via email.

## Testing
Locally tested on my Mac laptop, I confirm that these previously failing tests now pass:

```
The following tests FAILED:
  364 - TpetraCore_Core_initialize_where_tpetra_initializes_kokkos (Failed)
  365 - TpetraCore_Core_ScopeGuard_where_tpetra_initializes_kokkos (Failed)
  366 - TpetraCore_Core_initialize_where_user_initializes_kokkos (Failed)
  367 - TpetraCore_Core_ScopeGuard_where_user_initializes_kokkos (Failed)
  427 - TpetraCore_Map_no_warn_on_pre_finalize_destruction_issue_2372 (Failed)
```

Issue #3453, opened 2018-09-17
Tpetra:  unreliable test condition in ScopeGuard tests


kddevin:
@trilinos/tpetra @trilinos/kokkos 

## Current Behavior

The following tests fail on white for a CUDA build:
         31 - TpetraCore_Core_initialize_where_user_initializes_mpi_MPI_4 (Failed)
         32 - TpetraCore_Core_ScopeGuard_where_user_initializes_mpi_MPI_4 (Failed)
         35 - TpetraCore_Core_initialize_where_tpetra_initializes_kokkos_MPI_1 (Failed)
         36 - TpetraCore_Core_ScopeGuard_where_tpetra_initializes_kokkos_MPI_1 (Failed)
         37 - TpetraCore_Core_initialize_where_user_initializes_kokkos_MPI_1 (Failed)
         38 - TpetraCore_Core_ScopeGuard_where_user_initializes_kokkos_MPI_1 (Failed)
         39 - TpetraCore_Core_initialize_where_tpetra_initializes_mpi_and_user_initializes_kokkos_MPI_2 (Failed)
         40 - TpetraCore_Core_ScopeGuard_where_tpetra_initializes_mpi_and_user_initializes_kokkos_MPI_2 (Failed)

These tests rely on Kokkos not writing to std::cerr during Kokkos::initialize.  However, for reasons unrelated to proper/improper use of Kokkos::initialize, Kokkos may write to std::cerr.  E.g.,  

"Captured output: Kokkos::Cuda::initialize WARNING: running kernels compiled for compute capability 3.5 on device with compute capability 3.7 , this will likely reduce potential performance."

In this case, all the initialization took place correctly (so the test should have passed), but Kokkos issued a warning to std::cerr (so the test failed).

Since Tpetra cannot control what Kokkos writes to std::cerr, this condition is not a reliable way to determine whether these tests pass or fail.

A side note:  The test goes on to say "Captured output is empty!" when they should say "Captured output is NOT empty!"  The incorrect message is confusing, but secondary to the unreliable test condition.

## Steps to Reproduce

On white:

module purge
module load openmpi/2.1.2/gcc/7.2.0/cuda/9.2.88
module load cmake/3.9.6
module load openblas/0.2.20/gcc/7.2.0
module load boost/1.65.1/gcc/7.2.0
module load cuda/9.2.88
module load netcdf-exo/4.4.1.1/openmpi/2.1.2/gcc/7.2.0/cuda/9.0.176
[...]

PR #8256, opened 2020-10-26
Tpetra:  remove check of Kokkos warning output in Tpetra tests

### LINKED ISSUE #7966: Tpetra: Fix initialization tests to properly capture errors from Kokkos::initialize

kddevin:
@trilinos/tpetra 

Addressing #7966, #3453, #6748.

These tests exercises various combinations of MPI, Kokkos and Tpetra  initialization.  They checked the combinations for correctness (e.g., that Tpetra::ScopeGuard initialized TPLs if needed and finalized them upon destruction).  All good!

But the tests also captured warnings from Kokkos::initialize, and failed if Kokkos produced any warnings.  The problem is, we can't control what warnings Kokkos chooses to issue, and some Kokkos warnings had nothing to do with whether or not these tests worked correctly.  In particular, this check caused problems with CUDA_LAUNCH_BLOCKING=0 testing (see #7966).

I removed the check for output from Kokkos::initialize.  
All other checks (various combinations of initialized/finalized for MPI and Kokkos) in these tests remain.

## Motivation

These tests were holding up CUDA_LAUNCH_BLOCKING=0 testing.
Also, having a check of stderr from a TPL is bad form; we don't control what warnings Kokkos chooses to print.

## Related Issues

* Closes #7966, #3453
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #6748
* Part of 
* Composed of 

## Testing

Tested on Ride with and without CUDA_LAUNCH_BLOCKING (cuda 9.2, pascal, debug)
Also tested on mac with clang.

### DIFF
--- a/packages/tpetra/core/test/Core/ScopeGuard_tpetra_inits_kokkos.cpp
+++ b/packages/tpetra/core/test/Core/ScopeGuard_tpetra_inits_kokkos.cpp
@@ -3,10 +3,6 @@
 #include "Tpetra_Core.hpp"
 #include "Kokkos_Core.hpp"
 
-namespace { // (anonymous)
-
-// NOTE TO TEST AUTHORS: The code that calls this function captures
-// std::cerr, so don't write to std::cerr on purpose in this function.
 void testMain (bool& success, int argc, char* argv[])
 {
   using std::cout;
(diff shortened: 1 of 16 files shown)


tasmith4:
@kddevin as per your [comment](https://github.com/trilinos/Trilinos/issues/3453#issuecomment-458750851) on #3453, should a change to re-enable these tests for CUDA PR builds be part of this PR, or has/will that be a separate PR?

kddevin:
Good suggestions, @tasmith4 
I will re-enable those tests.

kddevin:
@tasmith4  
Actually, those tests must have been reactivated at some point in the past.  I cannot find any disabled tests in the specified file.   
I checked the PR test output for CUDA, and those tests ran.  
https://testing.sandia.gov/cdash/viewTest.php?onlypassed&buildid=8134433
So with the passing tests, this PR is ready for review, please.  Thanks.
@trilinos/tpetra

tasmith4 (approved):

PR #8262, opened 2020-10-27
MueLu: Fix Aggregates when UVM=off

cgcgcg:
@trilinos/muelu 

Motivation: UVM removal
See #8265.


lucbv (approved):
Sure, that makes sense!

jhux2 (approved):
lgtm

PR #8263, opened 2020-10-27
Fix fei issue for clang compiler, fix provided by Mike Glass.

### LINKED ISSUE #8241: FEI:  Segfaults with clang compiler
## Bug Report
@trilinos/\<fei\>

### Description
For Sierra, we are now building/testing with clang compilers (v10) since there are now many compilers moving toward the clang/llvm framework (Cray, Intel, ARMHPC, IBM, etc). We had previously identified a bug in

EqnCommMgr::mirrorProcEqns() in packages/fei/base/fei_EqnCommMgr.cpp

with the ARMHPC compilers on stria. A workaround was put in place to de-optimize that function when compiling with the ARMHPC compilers. With our clang-10 testing, we now see that this issue is present here too. We would like to have this fixed to work with an optimized build with clang compilers.

alanw0:
@trilinos/<teamName>

## Related Issues

* Closes #8241

### DIFF
--- a/packages/fei/base/fei_EqnCommMgr.cpp
+++ b/packages/fei/base/fei_EqnCommMgr.cpp
@@ -741,9 +741,9 @@ void EqnCommMgr::exchangeSoln()
 }
 
 //------------------------------------------------------------------------------
-// This works around an issue with the ARMHPC 20.1 compiler
+// This works around an issue with the clang and ARMHPC compiler
 // Needs to be revisited with later versions
-#ifdef __ARM_HPC_COMPILER_VERSION__
+#if defined (__clang__) && ! defined(__INTEL_CLANG_COMPILER)
 __attribute__((optnone))
 #endif
 int EqnCommMgr::mirrorProcEqns(ProcEqns& inProcEqns, ProcEqns& outProcEqns)


alanw0:
This is the fix for issue #8241

packages/fei/base/fei_EqnCommMgr.cpp
```diff
@@ -741,9 +741,9 @@ void EqnCommMgr::exchangeSoln()
 }
 
 //------------------------------------------------------------------------------
-// This works around an issue with the ARMHPC 20.1 compiler
+// This works around an issue with the clang and ARMHPC compiler
 // Needs to be revisited with later versions
-#ifdef __ARM_HPC_COMPILER_VERSION__
+#if defined (__clang__) && ! defined(__INTEL_CLANG_COMPILER)
```
tasmith4, line 746: would we need to have if defined ARM or if defined clang?  Or is straight replacement ok?

tasmith4: @mwglass

mwglass: The ARMHPC compiler defines __clang__ so we don't need the __ARM_HPC_COMPILER_VERSION__ check any longer. The previous check was to just restrict this workaround for the ARMHPC compiler but now this looks to be an issue with the generic clang compiler too but does not affect the Intel next gen compiler.

tasmith4 (approved):

PR #8269, opened 2020-10-27
Reduce clang warning from Teuchos

### LINKED ISSUE #8210: Teuchos:  Clang 10 compiler warning
## Bug Report
@trilinos/teuchos
@shardest

### Description
Using clang-10 to build Salinas from the master Sierra branch, we are getting 3095 compiler warnings related to the file Teuchos_as.hpp. This is against whatever Trilinos is currently in Sierra – not Trilinos integration, although it appears that this file has not changed in some time. There are several hundred more warnings that I have not examined; they may or may not have anything to do with Trilinos. The following is a typical example (starting at Teuchos_as.hpp:1392):

```c++
  static unsigned long long safeConvert (const double t) {
    const unsigned long long minVal = 0; // unsigned, so min value is 0.                          
    const unsigned long long maxVal = std::numeric_limits<unsigned long long>::max ();

    TEUCHOS_TEST_FOR_EXCEPTION(
      t < minVal || t > maxVal,
      std::range_error,
      "Teuchos::ValueTypeConversionTraits<unsigned long long, double>::safeConvert: "
      "Input double t = " << t << " is out of the valid range [" << minVal
      << ", " << maxVal << "] for conversion to unsigned long long.");

    return static_cast<unsigned long long> (t);
  }
```

```
Teuchos_as.hpp:1396:55: warning: implicit conversion from 'const unsigned long long' to 'double' changes value from 18446744073709551615 to 18446744073709551616 [-Wimplicit-int-float-conversion]
```

It appears to be implicitly casting maxVal to a double for the comparison against t, which is bad. I would think we’d want both values to be explicitly the type with larger range before doing a comparison. There are many related functions in this file with the same issue. I’m surprised that there isn’t yet some kind of STL function that does this?

prwolfe:
This is to resolve issue #8210

@trilinos/teuchos 

## Motivation
The clang compiler issues a significant number of warnings due to the implicit casts in these files. This makes those casts explicit to handle this warnings

## Related Issues

* Closes #8210 

## Testing
Teuchos was built against GCC7.2 and Clang 10 and all tests passed

### DIFF
--- a/packages/teuchos/core/src/Teuchos_ScalarTraits.hpp
+++ b/packages/teuchos/core/src/Teuchos_ScalarTraits.hpp
@@ -649,7 +649,7 @@ struct ScalarTraits<float>
     random();
 #endif
   }
-  static inline float random() { float rnd = (float) std::rand() / RAND_MAX; return (-1.0f + 2.0f * rnd); }
+  static inline float random() { float rnd = (float) std::rand() / static_cast<float>(RAND_MAX); return (-1.0f + 2.0f * rnd); }
   static inline std::string name() { return "float"; }
   static inline float squareroot(float x)
     {
--- a/packages/teuchos/core/src/Teuchos_as.hpp
+++ b/packages/teuchos/core/src/Teuchos_as.hpp
@@ -1339,7 +1339,7 @@ public:
     const unsigned long maxVal = std::numeric_limits<unsigned long>::max ();
 
     TEUCHOS_TEST_FOR_EXCEPTION(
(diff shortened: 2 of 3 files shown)


bartlettroscoe:
Should I review this code so it can get merged?

prwolfe:
I asked for Dan to do so as he is the author of the last few commits. I know Mike wants it relatively soon though.

ibaned (approved):
I'm definitely not the real owner of this code but the changes look reasonable enough for me to sign off on.

PR #8274, opened 2020-10-28
Muelu: remove code duplicates in BlockedCoarseMapFactory

mayrmt:

## Motivation

Although `BlockedCoarseMapFactory` derives from `CoarseMapFactory`, they have shared a lot of duplicated code. This PR refactors `CoarseMapFactory` to reduce code duplication. Now, the only difference between both factories is the mechanism to compute the GID offset.

The actual coarsening is implemented in `CoarseMapFactory`. For blocked systems, `BlockedCoarseMapFactory` computes the GID offset and then uses the base class implementation (`CoarseMapFactory`) to computed the coarsened map.

Further changes: some cleanups in various MueLu and Xpetra files.

## Testing

Existing tests pass. 

## Additional Information

`CoarseMapFactory` allows for a parameter `"Domain GID offsets"`. This is prone to errors and requires a-priori knowledge about the number and size of levels. It's rather recommended to use `BlockedCoarseMapFactory` to auto-compute the GID offset. Hence, I'd like to deprecate/remove `"Domain GID offsets"` from `CoarseMapFactory`. It is not used in any test/example in the Trilinos repository, but I don't know about application codes. Since I'm not sure how to do so safely, I'm open for suggestions.


packages/muelu/src/Transfers/BlockedTransfers/MueLu_BlockedCoarseMapFactory_decl.hpp
```diff
@@ -64,60 +58,77 @@
...
 #undef MUELU_BLOCKEDCOARSEMAPFACTORY_SHORT
 #include "MueLu_UseShortNames.hpp"
 
 public:
 
-  //! @name Constructors/Destructors.
-  //@{
-
-  //! Constructor
-  BlockedCoarseMapFactory();
```
lucbv, line 90: So do you want to remove all constructor/destructors?
I could understand the constructor removal if you only to construct these with Build factories but I feel keeping the destructor around is a good idea...
At this point the class will simply define the same default constructors implicitly, I'm not sure why we should remove the explicit ones.

lucbv (approved):
Looks OK

packages/muelu/src/Transfers/BlockedTransfers/MueLu_BlockedCoarseMapFactory_decl.hpp
mayrmt: This was just about removing the default con-/destructor, that is auto-generated anyway.

PR #8282, opened 2020-10-29
Framework: PR Configuration fix for Clang 7.0.1 build

william76:
The clang 7.0.1 build loads `sems-cmake/3.10.3` from the SEMS-DEFAULT section but then it tries to load
`sems-cmake/3.12.2` which will fail because a `sems-cmake` module is already loaded.

I added a `module-remove sems-cmake:` rule prior to loading the changed version. This should fix that issue.

@trilinos/framework 
@prwolfe

### DIFF
--- a/cmake/std/pr_config/pullrequest.ini
+++ b/cmake/std/pr_config/pullrequest.ini
@@ -274,6 +274,7 @@ module-load sems-gcc: 5.3.0
 module-load sems-clang: 7.0.1
 use SEMS-DEFAULT:
 use ATDM-DEFAULT:
+module-remove sems-cmake:
 module-load sems-cmake: 3.12.2
 setenv SEMS_FORCE_LOCAL_COMPILER_VERSION: 5.3.0
 


william76:
I found this by having a look at the console log in Jenkins for the clang 7.0.1 tests... since it dies before we hit the CTest call, there was nothing yet reported to CDash.

If we had some way to catch these or report them up to the PR comments it'd be handy, but I suspect that won't be around until the next WFT tool comes out.

prwolfe:
Any reason not to update all of the cmake modules and make them consistent?

prwolfe (approved):

william76:
> Any reason not to update all of the cmake modules and make them consistent?

Not really, I think it's something we can easily do... provided we aren't keeping some older versions enabled to add an extra compatibility test?  Also, I want to see some clean master-merge operations getting through before I start tweaking things too much.

I have some preliminary work on removing Python2 dependencies started but we'll need to take the Python2 test out of the autotester's list :D

PR #8285, opened 2020-10-30
Percept: Fix install paths for headers

bricar:
@trilinos/percept

## Motivation
The current install of percept does not put the header files in their correct directories. This fix does this using the top level CMakeLists.txt file.


rppawlo (approved):

GrahamBenHarper:
@rppawlo ping

PR #8294, opened 2020-11-02
Added "TRIBITS Pthread" to PTHREAD CMake configuration in Kokkos.

kuberry:
@trilinos/kokkos

## Motivation

PARSED_TRIBITS does not have Pthread registered, so it is not possible to build Trilinos with TPL_ENABLE_Pthread:BOOL=ON. This PR adds "TRIBITS Pthread" to the PTHREAD section in kokkos_tpls.cmake.

## NOTE: These lines are different between kokkos standalone and kokkos in Trilinos. For reference, you can check out the CUDA lines in kokkos_tpls.cmake 

```
KOKKOS_TPL_OPTION(CUDA    ${Kokkos_ENABLE_CUDA} TRIBITS CUDA)
``` 

which does not have 'TRIBITS CUDA' in the kokkos standalone repo.

Addresses issue [https://github.com/kokkos/kokkos/issues/3555](https://github.com/kokkos/kokkos/issues/3555)


kuberry:
Was informed this should go into kokkos/develop rather than directly into Trilinos.

PR #8298, opened 2020-11-02
MueLu: Fix spread lumping when using BuildNew() version of filtered matrix

rstumin:
Auto-PR for SHA 02fe812

### DIFF
--- a/packages/muelu/src/Misc/MueLu_FilteredAFactory_def.hpp
+++ b/packages/muelu/src/Misc/MueLu_FilteredAFactory_def.hpp
@@ -212,6 +212,7 @@ namespace MueLu {
        Xpetra::IO<SC,LO,GO,NO>::Write("A.dat", *A);
        RCP<Matrix> origFilteredA = MatrixFactory::Build(A->getRowMap(), A->getColMap(), A->getNodeMaxNumRowEntries());
        BuildNew(*A, *G, lumping, dirichlet_threshold,*origFilteredA);
+       if (use_spread_lumping) ExperimentalLumping(*A, *origFilteredA, DdomAllowGrowthRate, DdomCap);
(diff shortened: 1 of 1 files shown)


cgcgcg:
@trilinos/framework @william76 Any idea why the autotester doesn't want to start?

jwillenbring:
@cgcgcg We had an issue with a node and an effort to keep that node idle inadvertently made it so the GCC 8.3 test could not run anywhere. This issue should be corrected. I applied the retest label.

cgcgcg:
@jwillenbring Great, thanks!

packages/muelu/src/Misc/MueLu_FilteredAFactory_def.hpp
```diff
@@ -573,7 +576,7 @@ namespace MueLu {
  	if (row >= (LO)numRows) continue;
 	A.getLocalRowView(row, indsA, valsA);
 	for(LO k=0; k<(LO)indsA.size(); k++) {
-	  if(indsA[k] < (LO)numRows) {
+	  if ( (indsA[k] < (LO)numRows) && (valsA[k] != 0.0) ) {
```
cgcgcg, line 579: Instead of checking for actual zeros, should this have a threshold?

rstumin: The hard zero is ok ... but I should do the scalar traits stuff ... that I forgot about. The zero tests whether this is a really a nonzero or not. It is actually somewhat arbitrary (like it changes the math ... but it is not really sure what is better numerically) and does not cause the code to bomb either way. On the other hand, I should be using the scalar traits swill ... and zero() ... and stuff like that. I'll make that change .. hopefully in a few minutes.

cgcgcg (approved):

PR #8303, opened 2020-11-03
Tpetra: CrsGraph: stop trying to read row pointers on host

### LINKED ISSUE #8286: Tpetra:  Invalid access of k_rowPtrs_ inside CrsGraph
## Bug Report
@trilinos/tpetra 

### Description
When running the CrsGraph unit tests, Kokkos errors with "attempt to access inaccessible memory space" from line 1590 in `TpetraCore_CrsGraph_def.hpp`.  This is an attempt to index into `this->k_rowPtrs_` which is a view templated on `device_type`.

### Steps to Reproduce
1. Configure Tpetra for Cuda with `-DKokkos_ENABLE_CUDA_UVM=OFF` and `-DTpetra_ENABLE_CUDA_UVM=OFF`
2. Build
3. Run `ctest -R "TpetraCore_CrsGraph_UnitTests0_MPI_4"`

tasmith4:
@trilinos/tpetra @cgcgcg @lucbv 

## Motivation
CrsGraph was reading the row pointers on host, which leads to segfaults when CudaSpace is the device memory space.

## Related Issues

* Closes #8286 

## Testing
Tests still pass locally in a UVM build.  The particular segfault no longer occurs in a non-UVM build.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -1577,8 +1577,9 @@ namespace Tpetra {
         ret.allocSize = 0;
       }
       else {
-        ret.offset1D  = this->k_rowPtrs_(myRow);
-        ret.allocSize = this->k_rowPtrs_(myRow+1) - this->k_rowPtrs_(myRow);
+        auto lclRowPtrs = ::Tpetra::Details::getEntriesOnHost(this->k_rowPtrs_, myRow, 2);
+        ret.offset1D  = lclRowPtrs(0);
+        ret.allocSize = lclRowPtrs(1) - lclRowPtrs(0);
       }
 
       ret.numEntries = (this->k_numRowEntries_.extent (0) == 0) ?
@@ -1633,8 +1634,9 @@ namespace Tpetra {
         ret.allocSize = 0;
       }
       else {
-        ret.offset1D  = this->k_rowPtrs_(myRow);
-        ret.allocSize = this->k_rowPtrs_(myRow+1) - this->k_rowPtrs_(myRow);
+        auto lclRowPtrs = ::Tpetra::Details::getEntriesOnHost(this->k_rowPtrs_, myRow, 2);
+        ret.offset1D  = lclRowPtrs(0);
+        ret.allocSize = lclRowPtrs(1) - lclRowPtrs(0);
       }
 
       ret.numEntries = (this->k_numRowEntries_.extent (0) == 0) ?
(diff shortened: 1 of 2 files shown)


packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
```diff
@@ -3084,8 +3086,9 @@ namespace Tpetra {
         // We have to iterate through the row offsets anyway, so we
         // might as well check whether all rows' bounds are the same.
         bool allRowsReallySame = false;
+        auto k_rowPtrs_h = ::Tpetra::Details::getEntriesOnHost(this->k_rowPtrs_, 0, numRows);
```
tasmith4, line 3089: @kddevin or @brian-kelley -- I'm copying the whole array here because we know we're going to walk it anyway.  Are there any guarantees about when k_rowPtrs_ changes that allow us to save some performance by moving this copy out of this function to a place called less frequently?

brian-kelley: Yes, ``k_rowPtrs_`` should never change in between a fillComplete() and resumeFill(), since it can only change when entries are inserted or deleted from a row. ``lclGraph.row_map`` is const-valued to enforce this.

tasmith4: so is there a problem with making a member view `k_rowPtrs_h_` that gets copied over during `fillComplete`?

What other functions might need to be called when the matrix is not fill-complete?  e.g. could we get rid of the `getEntryOnHost` and `getEntriesOnHost` calls throughout, or do some functions need to access the row pointers before fill-complete?

brian-kelley: @tasmith4 I think that's a good strategy - before fillComplete(), just use ``k_rowPtrs_h_`` so that ``getRowInfo()`` is very fast, and then copy it to device ``k_rowPtrs_`` during fillComplete when the local graph/matrix is being constructed. Then, ``k_rowPtrs_h_`` is always up-to-date whether fill complete or not.

tasmith4: @brian-kelley so you're saying that all the performance-critical device things we need `k_rowPtrs_` to live on device for actually happen after fill-complete, so we don't need to be on device before then?

brian-kelley: @tasmith4 Right

brian-kelley: If you end up getting runtime errors from ``k_rowPtrs_`` being accessed on device before fillComplete, then you might have to make it a DualView or manually deep copy, but I don't know of anywhere that happens

tasmith4: @brian-kelley do you know the difference between `fillComplete` and `expertStaticFillComplete`?

brian-kelley: I didn't know off the top of my head but:
3675     // Note: We don't need to do the following things which are normally done in fillComplete:
3676     // allocateIndices, globalAssemble, makeColMap, makeIndicesLocal, sortAndMergeAllIndices

brian-kelley (approved):
Looks good to me (correct), future TODO is to store k_rowPtrs_h_ persistently while not fill-complete so that getRowInfo() is fast during assembly.

brian-kelley (changes requested):
Like we talked about on Friday, take care of #8345 before merging.

tasmith4:
No longer relevant

PR #8307, opened 2020-11-03
Tpetra: fix typo in test instantiation

brian-kelley:
Replace TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL with TEUCHOS_UNIT_TEST_TEMPLATE_4_INSTANT
at the bottom of Vector_offsetViewCtor.cpp. the ...DECL macro is already called for this test above, at the bottom it needs to be instantiated with the ...INSTANT macro. Note that this file is not actually compiled (@mhoemmen had commented out it in CMakeLists.txt) but it's tripping up the automatic LO/GO removal script.

@trilinos/tpetra 

## Motivation

Just allows the LO removal script to run on this file. It doesn't change any behavior or affect testing, since this file is never compiled.

### DIFF
--- a/packages/tpetra/core/test/MultiVector/Vector_offsetViewCtor.cpp
+++ b/packages/tpetra/core/test/MultiVector/Vector_offsetViewCtor.cpp
@@ -190,7 +190,7 @@ namespace { // (anonymous)
 //
 
 #define UNIT_TEST_GROUP( SCALAR, LO, GO, NODE ) \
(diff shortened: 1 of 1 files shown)


brian-kelley:
I did try uncommenting the build of this test in CMakeLists, but there are other build errors. At some point we should either fix this file or just delete it.

tasmith4 (approved):

tasmith4:
@brian-kelley if one doesn't exist already, can you create a bug report to document that we need to circle back and fix this file or delete it?

brian-kelley:
Also opened #8309 to deal with this file

mhoemmen:
Dang, y'all still working on this?

PR #8288, opened 2020-10-30
Framework: Remove Python2 Dependencies in PR Driver Scripts

william76:
@trilinos/framework 

## Motivation
Python 2.x is beyond its end-of-life, so it's time to remove its support from our PR driver scripts.

This should only affect the PR drivers which are run by the Autotester.

## Testing Information
Local testing:
```
0 $ python3 -m pytest
======================================== test session starts ========================================
platform darwin -- Python 3.8.0, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
rootdir: /Users/wcmclen/dev/trilinos/trilinos-dev-framework/source/Trilinos/cmake/std
plugins: metadata-1.10.0, cov-2.10.0, xdist-2.0.0, sphinx-0.3.1, forked-1.3.0, osxnotify-0.1.7
collected 98 items

trilinosprhelpers/gitutility/unittests/test_GitUtility.py ....                                [  4%]
trilinosprhelpers/jenkinsenv/unittests/test_EnvvarHelper.py ....                              [  8%]
trilinosprhelpers/jenkinsenv/unittests/test_JenkinsEnv.py ....................                [ 28%]
trilinosprhelpers/jenkinsenv/unittests/test_TrilinosJenkinsEnv.py .........                   [ 37%]
trilinosprhelpers/setenvironment/unittests/test_ModuleHelper.py .......                       [ 44%]
trilinosprhelpers/setenvironment/unittests/test_SetEnvironment.py .....................       [ 66%]
trilinosprhelpers/sysinfo/unittests/test_SysInfo.py ....                                      [ 70%]
trilinosprhelpers/unittests/test_TrilinosPRConfigurationBase.py ........................      [ 94%]
trilinosprhelpers/unittests/test_TrilinosPRConfigurationInstallation.py ..                    [ 96%]
trilinosprhelpers/unittests/test_TrilinosPRConfigurationStandard.py ...                       [100%]

======================================== 98 passed in 0.92s =========================================
```

We can hold off on merging this until sometime next week after we've had a chance to chat in the stand-up meeting.

I'd also like to see that Tempus error go away from the Master Merge jobs.

PR #8322, opened 2020-11-07
Framework: Remove Python 2 use from PR test driver

william76:
@trilinos/framework 

Remaking RP #8288 - the CDash logs were getting a little confusing with the reporting. This PR should clean things up a bit.

The GCC 8.3.0 build seems to be having some odd SEACAS error in the other PR -- will have to investigate why it's happening if it continues.

## Motivation
Python 2.x is beyond its end-of-life, so it's time to remove its support from our PR driver scripts.

This should only affect the PR drivers which are run by the Autotester.


william76:
FYI @jmgate This is the PR I have currently testing out the Python 2.x removal.

william76:
@prwolfe this one is passing now with upping the CMake version to the SEMS 3.12.2 one in the gcc 8.3.0 test.

I enabled the automerge flag, if you're ok with merging in feel free to approve... or we can do a review pass first if you want.

FYI @jwillenbring @ZUUL42 looks like it's the gcc 8.3.0 test that requires some CMake higher than 3.10.x to work... I changed it to 3.12 and it's passing now.

prwolfe:
Will - do we understand what the newer python provides for the gcc 8.3 build? This looks good, but unexplained items always bug me.

prwolfe (approved):

william76:
> 
> 
> Will - do we understand what the newer python provides for the gcc 8.3 build? This looks good, but unexplained items always bug me.

Not really... the gcc 8.3.0 build always failed if we used CMake 3.10.3.  This was the one that @ZUUL42 mentioned yesterday that wouldn't work with 3.10.x so he had raised it in the old setup to the 3.17.x version.  In this PR it looks like 3.12.x is sufficient but the error was funny and I'm not sure what was going on there.  It was that 1 build warning that triggered a warning-as-error I think down in SEACAS.  No tests failed, just that one build issue.

@ZUUL42 Henry, correct me if I got what you said yesterday incorrect...

jmgate:
Does this mean we're Python 3.6+ from now on?  If so, that's good&mdash;just making sure I understand the PR.

PR #8327, opened 2020-11-10
WIP: Tpetra kokkos counter regression test

GeoffDanielson:
@trilinos/tpetra 
@trilinos/kokkos 

## Motivation

This is a regression test to count the deep copies performed when using Kokkos Views in tpetra.  It needs eyes from Kokkos team members to ensure that the number of times the cuda functions are called by kokkos are correct.

## Testing

This is mostly a testing change - the only thing that might be reusable down the line is the LD_PRELOAD stuff for intercepting cuda calls, but even that should mostly be in a testing context.


crtrott:
There is neither an explicit nor an implicit guarantee that these numbers will not change for whatever reasons. I think it is generally not a great idea to write tests in downstream project which assert a specific way of us implementing capabilities. So generally speaking: I would advise against this, and we will not consider it a-priori a breaking change on our side if this test breaks. Specifically we may change stuff like using async copies instead of synchronous memcpies replace stream synchronize with other synchining mechanisms, change launch mechanism, require additional memcpies for various backend reasons, change the way functors are copied which may introduce additional memcpies or symbol copies etc. 

We could MAYBE add a test like that in Kokkos proper, and adjust this based on what we do. But these numbers will for example change if tools are loaded (profiling, tuning or debugging tools). So this will be at least reasonable brittle in CI environments which may start to use tools hooking into the Kokkos callbacks.

kddevin:
Thanks, @crtrott.  Part of the motivation behind this test is knowing what the Kokkos code is doing and discovering when its behavior changes.  The first version of this code, for example, helped us understand differences between deep_copy with two and three arguments.  When Kokkos' behavior changes -- as it certainly is allowed to -- it is helpful if we know about it.  This test will help us discover changes without having to stay up-to-date on Kokkos' slack.
I can see that such a test might be more appropriate in Kokkos than Tpetra, but then it would be up to the Kokkos developers to let us know when there are behavior changes to which we should adapt.

crtrott:
I mean it is your call in the end. I think it might be better to do timing tests which check for asynchronicity, than intercepting backend calls. Some of those backend calls don't tell you everything you need to know with respect to asynchronicity. For example certain operations in the default stream will block everything else, while from the call perspective completely identical calls with non-default streams will not. But I do get the motivation as long as you are ok to fix the test when these counts change.

DavidPoliakoff:
I haven't gotten a reply over email, so I'm checking here. What would this give you that fence profiling wouldn't, if fence profiling is the primary goal? If every time Kokkos calls cudaXSynchronize, we have a tool event that says "I'm beginning/ending a fence?" That also has portability benefits, we can do the same for HIP/SYCL without needing to generate these interceptor libraries.

I'm asking because I've written these interceptor tools, and they always are good ideas for two months, and on the third month somebody compiles libcuda to make some of its calls use the GOT and your results break, or a vendor changes their library structure so a call crosses a library boundary and your numbers change, or something gets statically linked...

If we could give you these results with instrumentation and callbacks, you'd be a lot less likely to bump up against the sheer goofiness of shared libraries. My only other recommendation is that this shouldn't be a test that happens in a default build of Trilinos unless you want users who run into the above mentioned problems to file confusing issues about the failing test

kddevin (approved):
I like having a canary in the coalmine.  Thanks, @GeoffDanielson

kyungjoo-kim:
@GeoffDanielson Is this good to be merged ?

PR #8329, opened 2020-11-10
Minor fix to includes.

alanw0:
stk-mesh headers should be included with a relative path in order to
make sure that cmake support can be kept consistent.
stk cmake support will soon be improved and '-I' paths will be cleaned
up.

@trilinos/<teamName>


rppawlo (approved):
Thanks @alanw0 !

PR #8331, opened 2020-11-11
Xpetra: remove obsolete CMake commands

mayrmt:
@trilinos/xpetra 

## Motivation

We don't have to exclude the directory `src/test/BlockedCrsOperator`, since it does not exist.

## Stakeholder Feedback

n/a.

## Testing

Build and existing tests pass.


lucbv (approved):
lgtm

PR #8333, opened 2020-11-11
WIP: ETI on as default

prwolfe:
@trilinos/framework 

## Motivation
This will change the default for ETI from OFF to ON. I realize this is also changing the default in tributes, which may not be wanted, so I am asking @bartlettroscoe to weigh in on that.

## Testing
This has been used extensively in PR testing for some months. The specific change was tested using gcc 8.3.0 directly.

### DIFF
--- a/cmake/tribits/core/package_arch/TribitsGlobalMacros.cmake
+++ b/cmake/tribits/core/package_arch/TribitsGlobalMacros.cmake
@@ -482,7 +482,7 @@ MACRO(TRIBITS_DEFINE_GLOBAL_OPTIONS_AND_DEFINE_EXTRA_REPOS)
     "Print out when all of the various files get processed."
     )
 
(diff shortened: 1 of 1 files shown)


bartlettroscoe (commented):
@prwolfe, if this change is to be made in TriBITS, it should be made in the TriBITS repo and then have a new TriBITS snapshot.  Can you submit a PR to the https://github.com/TriBITSPub/TriBITS repo for this change?

bartlettroscoe:
But as for making ETI turned on by default for Trilinos, that is a big thumbs up from me.  And users will thank you as well.

prwolfe:
> But as for making ETI turned on by default for Trilinos, that is a big thumbs up from me. And users will thank you as well.

Agreed, but I was not sure if Tribits was the place or the top level CMakeLists.txt in Trilinos. Since you are good with this I will move this PR over to the tribits repo and close this one afterward.

prwolfe:
This has been moved to the TriBITS red as https://github.com/TriBITSPub/TriBITS/pull/336

PR #8335, opened 2020-11-11
Tpetra: NewReader: Fix bug in chunk computation

seheracer:
@kddevin 

This PR fixes a bug in chunk computation.


kddevin (approved):
Thanks, @seheracer

PR #8337, opened 2020-11-12
Create Intel 19.0.5 Build

### LINKED ISSUE #6294: Create Intel 19.0.5 PR Testing Build
## Enhancement
@trilinos/framework 

Look into creating a PR Testing build that uses Intel 19.0.5.
Currently we have an Intel 17.0.1 PR testing build.
Intel 19 would likely replace the 17 build.

ZUUL42:
@trilinos/framework 

## Motivation

This will put an Intel 19.0.5 build in place.

Closes #6294 

## Testing

[CDash results](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=PR-intel19ZoltanScotchOFF-test-Trilinos_pullrequest_intel_19.0.5-104&field2=buildstamp&compare2=63&value2=Experimental) from building these changes with the Jenkins job that will run it.

## Additional Information

While this does turn off a few tests, it will get an Intel 19 build out there and a new issue will be created to follow up on that work.

### DIFF
--- a/cmake/std/pr_config/pullrequest.ini
+++ b/cmake/std/pr_config/pullrequest.ini
@@ -96,7 +96,8 @@ Trilinos_pullrequest_gcc_7.2.0_serial: PullRequestLinuxGCC7.2.0SerialTestingSett
 Trilinos_pullrequest_gcc_8.3.0:        PullRequestLinuxGCC8.3.0TestingSettings.cmake
 
 # Intel Builds
-Trilinos_pullrequest_intel_17.0.1:     PullRequestLinuxIntelTestingSettings.cmake
+Trilinos_pullrequest_intel_17.0.1:     PullRequestLinuxIntel17.0.1TestingSettings.cmake
+Trilinos_pullrequest_intel_19.0.5:     PullRequestLinuxIntel19.0.5TestingSettings.cmake
 
 # Python PR Driver Builds
 #Trilinos_pullrequest_python_2:         PullRequestLinuxPython2.cmake
@@ -274,6 +275,19 @@ use SEMS-ENV:
 use ATDM-ENV:
 module-load sems-gcc: 5.3.0
 module-load sems-intel: 17.0.1
+module-load sems-mpich: 3.2
+use SEMS-DEFAULT:
+module-remove sems-openmpi:
+use ATDM-DEFAULT:
+setenv SEMS_FORCE_LOCAL_COMPILER_VERSION: 4.9.3
+
+
+
+[Trilinos_pullrequest_intel_19.0.5]
+use SEMS-ENV:
+use ATDM-ENV:
+module-load sems-gcc: 6.1.0
+module-load sems-intel: 19.0.5
 module-load sems-mpich: 3.2
 use SEMS-DEFAULT:
 module-remove sems-openmpi:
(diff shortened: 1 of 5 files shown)


cmake/std/pr_config/pullrequest.ini
```diff
@@ -267,6 +268,19 @@ setenv SEMS_FORCE_LOCAL_COMPILER_VERSION: 4.9.3
...
+[Trilinos_pullrequest_intel_19.0.5]
+use SEMS-ENV:
+use ATDM-ENV:
+module-load sems-gcc: 6.1.0
+module-load sems-intel: 19.0.5
+module-load sems-mpich: 3.2
+use SEMS-DEFAULT:
+module-remove sems-openmpi:
+use ATDM-DEFAULT:
+setenv SEMS_FORCE_LOCAL_COMPILER_VERSION: 4.9.3
```
jwillenbring, line 280: Why are we forcing the local compiler version to 4.9.3? Why not something newer? 4.9.3 doesn't support C++14.

cmake/std/trilinosprhelpers/unittests/trilinos_pr_test.ini
```diff
@@ -222,6 +223,18 @@ setenv SEMS_FORCE_LOCAL_COMPILER_VERSION: 4.9.3
...
 
+[Trilinos_pullrequest_intel_19.0.5]
+use SEMS-ENV:
+use ATDM-ENV:
+module-load sems-gcc: 6.1.0
+module-load sems-intel: 19.0.5
+module-load sems-mpich: 3.2
+use SEMS-DEFAULT:
+use ATDM-DEFAULT:
+setenv SEMS_FORCE_LOCAL_COMPILER_VERSION: 4.9.3
```
jwillenbring, line 234: Same question as above. If we are using 6.1.0, why wouldn't we use that for the local version also?

prwolfe: If I recall, this is supposed to change what the sems modules expect, but does not. I would have to experiment.

cmake/std/pr_config/pullrequest.ini
ZUUL42: @jwillenbring I was following the pattern in most all the other builds. 4.9.3 is used for force local compiler in GCC 7.2.0, 7.2 debug, 7.2 serial, 8.3.0 & Intel 17.0.1. The only ones that push higher are the Clang builds.
But going to 5.x for C++14 support or higher to match the build makes complete sense to me.

cmake/std/trilinosprhelpers/unittests/trilinos_pr_test.ini
ZUUL42: Ya, I've not seen it make a difference when putting together a build.
But I never dug further into it.
I can set them to the associated used GCC version and we can worry about the details in another issue.

jwillenbring: If we set it to the associated GCC, I wonder if we need it at all. It would seem the purpose of such an option might be to specify a compiler version different than the version of the compiler being used. For example, there are times when you have a set of TPLs built with a different version of a compiler that is link-compatible with the compiler you are using, but is not the same compiler. Could we just take this out?

ZUUL42: That's sort of what I was wondering.
I can remove them all and see how it tests out.

jwillenbring (approved):
This looks good now. Thanks.

PR #8342, opened 2020-11-13
Ifpack2/MueLu: Add timers

jewatkins:
@trilinos/ifpack2 @trilinos/muelu

## Motivation
I'd like to collect more timing/profiling data from muelu and ifpack2 for Albany Land Ice performance tests. This adds timers to the muelu ifpack2 smoother and the ifpack2 BlockRelaxation class. The timers I added to the BlockRelaxation class follow the same convention used in the Relaxation class.

I also removed some quotation marks from muelu timers because they were causing issues with the kokkos chrome tracing tool.

@mperego

### DIFF
--- a/packages/muelu/src/Smoothers/MueLu_Ifpack2Smoother_def.hpp
+++ b/packages/muelu/src/Smoothers/MueLu_Ifpack2Smoother_def.hpp
@@ -736,8 +736,19 @@ namespace MueLu {
       supportInitialGuess = true;
 
     } else if (type_ == "RELAXATION"       ||
+               type_ == "BLOCK_RELAXATION" ||
+               type_ == "BLOCK RELAXATION" ||
                type_ == "BLOCKRELAXATION"  ||
-               type_ == "BANDEDRELAXATION" ||
+               // Banded
+               type_ == "BANDED_RELAXATION" ||
+               type_ == "BANDED RELAXATION" ||
+               type_ == "BANDEDRELAXATION"  ||
+               // Tridiagonal
+               type_ == "TRIDI_RELAXATION"       ||
+               type_ == "TRIDI RELAXATION"       ||
+               type_ == "TRIDIRELAXATION"        ||
+               type_ == "TRIDIAGONAL_RELAXATION" ||
+               type_ == "TRIDIAGONAL RELAXATION" ||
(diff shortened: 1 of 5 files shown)


packages/muelu/src/Smoothers/MueLu_Ifpack2Smoother_def.hpp
```diff
@@ -765,7 +768,14 @@ namespace MueLu {
       prec_->apply(tpB, tpX);
     } else {
       typedef Teuchos::ScalarTraits<Scalar> TST;
-      RCP<MultiVector> Residual   = Utilities::Residual(*A_, X, B);
```
cgcgcg, line 768: @jewatkins We probably never want to be in this branch if it can be avoided. I am assuming that you're trying to use 'BLOCK RELAXATION'. Could you add another case like 
https://github.com/trilinos/Trilinos/blob/b49ef596e42047c862965ca0cad48b6d16439a63/packages/muelu/src/Smoothers/MueLu_Ifpack2Smoother_def.hpp#L738-L741
but for `BLOCK RELAXATION`?

jewatkins: Great, yeah let me try that.

jewatkins: Works thanks

packages/muelu/src/Smoothers/MueLu_Ifpack2Smoother_def.hpp
```diff
@@ -732,7 +735,10 @@ namespace MueLu {
       SetPrecParameters(paramList);
       supportInitialGuess = true;
 
-    } else if (type_ == "RELAXATION") {
+    } else if (type_ == "RELAXATION"       ||
+               type_ == "BLOCKRELAXATION"  ||
+               type_ == "BANDEDRELAXATION" ||
+               type_ == "TRIDIAGONALRELAXATION") {
```
cgcgcg, line 741: We probably want the gazillion other values of `type_` here as well.

cgcgcg (approved):
Approve, so the AT can do its thing

cgcgcg (approved):
LGTM

PR #8344, opened 2020-11-13
Windows: remove obsolete winmath.h

sethrj:
This TriBITS-distributed file seems to have been necessary at some point in the distant past for successful Windows builds, but it now causes build errors:
```
    D:\nfr\scale\TriBITS\tribits\win_interface\include\winmath.h(155): error: function "tgammal" has alread

      inline long double tgammal(long double z){
```

Without the winmath includes, Intrepid successfully builds on VS2015 (with an older Trilinos version) and VS2017.

@bartlettroscoe 

## Motivation

Building on Windows with newer versions of Visual Studio.

### DIFF
--- a/packages/intrepid/src/Shared/Intrepid_BurkardtRulesDef.hpp
+++ b/packages/intrepid/src/Shared/Intrepid_BurkardtRulesDef.hpp
@@ -48,10 +48,6 @@
     \author Created by D. Kouri and D. Ridzal.
  */
 
-#ifdef _MSC_VER
-#include "winmath.h"
-#endif
-
 //# include <cstdlib>
 //# include <iomanip>
 //# include <iostream>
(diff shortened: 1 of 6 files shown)


bartlettroscoe:
@sethrj, looks fine to me but I think that at least one @trilinos/intrepid and @trilinos/intrepid2 developer should approve this before it gets merged.

But I will go ahead and approve this so that the PR builds can fire off.

bartlettroscoe (approved):
Approving so that the PR builds will fire off so this can be tested.  Otherwise, some @trilinos/intrepid and @trilinos/intrepid2 developer should approve before this gets merged.

sethrj:
Thanks @bartlettroscoe ! It looks like the original `winmath.h` includes are actually attributed to @billhoffman in a4dbd013f25ea4a80a6d782e26c20e8a61db8889 11 years ago.

bartlettroscoe (approved):
Given that this is a very simple change that only impacts Windows builds (for which there are no automated builds of Trilinos right now posting to CDash) and given that no  @trilinos/intrepid  or @trilinos/intrepi2 developer has objected to this PR, I will go ahead and approve and add the `AT: AUTOMERGE` label to get this merged.

PR #8355, opened 2020-11-16
Trilinos:  Remove patch file

kddevin:
@trilinos/shylu  @searhein 

## Motivation

There was a patch file committed at the top level of Trillions.  (It blocks my autocomplete to the packages directory.)
I suspect it was added by accident, so this PR deletes it.
If it really is meant to be in the repository, we can discard this PR.  But I'd be curious to know why it belongs here.

Thanks!

## Testing

Not needed; this PR removes only a patch file.


srajama1 (approved):
Sorry for the trouble ! Yes, removing it is fine.

PR #8358, opened 2020-11-16
zoltan2:  fix edge cnt overflow in GraphModel

kddevin:
@ibogle detected an overflow in the edge counter in GraphModel.
ecnt type should be offset_t instead of lno_t.

@trilinos/zoltan2 

## Motivation

We want to work with large irregular graphs.

## Testing

Tested on cee-lan


egboman (approved):
Minor fix, looks fine!

PR #8373, opened 2020-11-19
Tempus remove duplicate tscs

sconde:
@trilinos/tempus 

## Motivation

Removing the duplicate file since it is not called anywhere. `IntegralControler` is what is being used and not `PID`.


ccober6 (approved):
Looks good!

Issue #8073, opened 2020-09-22
Documentation:  Can the bootstrap JavaScript be updated to >v3.4.0


sebrowne:
## Question
@trilinos/framework

There are reported security vulnerabilities in the bootstrap version (JavaScript code) contained inside of Trilinos (which appear to be related to generating documentation).  Can this be upgraded to >=3.4.0?  That was the listed mitigation for the issues when I had it scanned last.  (Also I didn't know the best team to put, but framework seemed almost barely relevant?)

sebrowne:
@bartlettroscoe

bartlettroscoe:
@sebrowne, what code is this?

bartlettroscoe:
CC: @trilinos/framework, @jwillenbring

sebrowne:
These two in particular, I believe:

./doc/DocumentingParameterLists/webfiles/bootstrap/js/bootstrap.js
./doc/DocumentingParameterLists/webfiles/bootstrap/js/bootstrap.min.js

bartlettroscoe:
People who have touched this:

```
$ cd Trilinos

$ git shortlog -ns doc/DocumentingParameterLists/
    17  William F. Spotz
     2  Brent M. Perschbacher
     2  Dena Vigil
     1  Peter Bosler
     1  Roscoe A. Bartlett
     1  Sidafa Conde
     1  William McLendon
```

This stuff was being used to display ParameterList information at some point.   The main contributed to this (Bill Spotz) is no longer at SNL. No idea if that is stuff still working or if anyone is using any of this.

I  wonder if this needs to even be in Trilinos proper at this point and could not just be pulled out into its own little git repo and cloned under Trilinos and managed with gitdist if someone wanted to use this stuff?

jwillenbring:
This functionality is still used as part of the documentation generation process. There are issues with that process when we move to C++14 on Monday, but it is currently used nightly.

jwillenbring:
I discovered some of the files that were used no longer exist in current versions. It is going to take a little work to see if the new version works. I don't recall the specifics for what this does, and there is no testing I am aware of.

sebrowne:
That's a very acceptable solution to me, thank you!

bartlettroscoe:
> Therefore, I propose that we simply remove these two files from the source tree. 

If someone needs to add back updated files, someone can do that in the future.

jwillenbring:
Since #8376 has been merged into develop, I am closing this issue.

PR #8376, opened 2020-11-19
Remove javascript files

jwillenbring:
@trilinos/framework 

Removing files due to possible security issues. See  issue #8073.

-->
@trilinos/<teamName>

## Motivation

@sebrowne reported issues with these files. See #8073 

## Stakeholder Feedback

@sebrowne indicated in #8073 that removing these files is an acceptable solution for him.

## Testing

I ran the parameter list documentation build on webserver without the files and it passed and all associated tests also passed. If we run into issues later, we could look at upgrading to a newer version of bootstrap as is noted as a possibility in #8073. However, these files do not exist in the newer versions, so the functionality either moved or was removed.


jwillenbring:
@ZUUL42 Could you review this PR to remove some js files?

ZUUL42 (approved):
Looks good

Issue #8306, opened 2020-11-03
Framework:  Automating Issue / Pull Request Maintenance and Cleanup


william76:
Description
------------
Trilinos currently has over 1100 open issues and over 40 open pull requests. We have some open issues that date back to 2015 with no activity since 2017, and PR's that have also had no activity in over a year. This becomes difficult to manage from the framework side because it's hard to tell what issues have been abandoned and which ones should be kept around for some reason such as planning, documentation, etc.

This issue is to facilitate some discussion on my proposal to set up a [GitHub Action][1] which will automatically mark issues stale and close them out after long term inactivity.

Proposal
---------
I'd like to set up a GitHub Action to flag issues and pull requests as 'stale' and then close them based on the following rules:
1. Issues or Pull Requests with _no_ activity for 12 months will get a new flag: `FLAGGED FOR AUTO CLOSE`
2. If the Issue or PR has had no new activity after being flagged in (1) after another month, then the issue or PR is closed by the bot.
3. When being flagged or closed the bot can add a useful message to the comments explaining why it's being autoclosed.
4. Issues or PR's that we wish to be exempt from the bot can be given a label that would exempt it from this check.
    1. `NO AUTOCLOSE` - This could be a generic flag to exempt items from the bot.
    1. `DOCUMENTATION` - This could be something to track Issues that are used for tracking documentation.
    1. `PROJECT PLANNING` - This could be added for issues used for long term project planning type things.
    1. `CLOSED DUE TO INACTIVITY` - A label to apply to issues and PR's that were autoclosed by the bot (makes them more searchable).

The flag names here are mostly just a quick stab at what we might be able to do.  The main page for the particular runner we're looking at [is here][4]. It has quite a few control options. We can set a custom message that it will add to the comments for Issues or PR's when it's flagging them and when its closing them.  We can also add in a flag that would be applied to issues that are closed via the bot which might be useful if we need to dig through what was autoclosed.

I think we can iterate on what Labels we might want to use to exempt things from the bot and how long we might want to give before marking stuff for removal.  I had thought 1 year to flagging and then an additional month before closing since we do have things that stick around long term in Trilinos so enforcing a policy that people should touch issues and PR's at least once a year if they wish to keep them around isn't a huge ask. Adding in the ability to _explicitly_ exempt items from the bot also benefits us because those issues would become more easily searched.

[...]

PR #8383, opened 2020-11-23
Framework: Add GitHub Actions script to close stale issues and PR's

william76:
@trilinos/framework 

## Motivation
See #8306 for more information about this work.

This PR pulls in the initial github runner script that will mark issues and PR's as stale after 365 days of inactivity and then close them after another 30 days from being marked for closure.

## Testing
I've been testing this out in a test repository at https://github.com/william76/workflowTest/

## Additional Information
This is in **debug mode** (`debug-only: true`) currently so it should not actually affect any issues or PR's. We should get this in place so we can watch what it does in our repository before we enable it.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -4,9 +4,6 @@
 #
 name: "Close stale issues"
 on:
-  # For debugging: toggle action by starring/unstarring the repo
-  watch:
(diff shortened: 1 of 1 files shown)


william76:
FYI @jwillenbring @prwolfe

Issue #8387, opened 2020-11-24
Framework:  Some build configurations may not work with CMake 3.10.x


william76:
## Bug Report
@trilinos/framework 

### Description
When I was working on the PR driver code recently I had noticed that Trilinos' CMake indicates a minimum required version of 10.0.0 so I set all the builds to use CMake 10.0.3, which is the only 10.x version provided by SEMS modules. This resulted in a failure in the GCC 8.3.0 build.

In a discussion of this during our Framework stand-up meeting @ZUUL42 mentioned that he had run into this too and had set the CMake for that to the highest one SEMS provides (3.17.x) and it worked.

I changed the version in our configuration to use the 3.12.x SEMS one and that worked so I pushed that on up.

This brings up an issue though.  **If** we have CMake set to require 3.10.0 as the minimum version but we have a build that does not work with CMake 3.10.x then either:
1. The minimum required version in the top-level CMakeLists.txt file is wrong and should be changed to a higher version.
2. There is a bug somewhere in our CMake scripts that has broken our minimum version requirement and needs to be fixed to bring our CMake back in-line with the stated minimum requirement of 3.10.0.

I'm filing this issue so we can track it. I think we need some further experimentation to confirm this and then we should decide what to do re: minimum requirement for CMake.

### FYI:
@jwillenbring 
@prwolfe 
@ZUUL42 
@crtrott 
@bartlettroscoe 
@mmwolf

bartlettroscoe:
> When I was working on the PR driver code recently I had noticed that Trilinos' CMake indicates a minimum required version of 10.0.0 so I set all the builds to use CMake 10.0.3, which is the only 10.x version provided by SEMS modules.

@william76, you mean "CMake 3.10.0" and not "CMake 10.0.3", correct?

> This resulted in a failure in the GCC 8.3.0 build.

What was the nature of this failure?  Can you provide more details on that?

william76:
> @william76, you mean "CMake 3.10.0" and not "CMake 10.0.3", correct?
I mean that in our CMakeLists.txt file we set the minimum required version of CMake to be 3.10.0 and that the closest to that we have available is the sems-cmake/3.10.3.

> What was the nature of this failure? Can you provide more details on that?
I don't remember it. It was buried in the initial PR that we had to get that stuff in place.  The PR itself had about 200 comments on it from the review that @jmgate did for us.

[...]

PR #8392, opened 2020-11-25
Framework: Setting CMake specification to 3.10.3 for PR builds

william76:
@trilinos/framework 

## Motivation
We have some PR builds that seem to fail if we use CMake 3.10.x, which violates the assertion in the top-level CMakeLists.txt file that the minimum required CMake is a 3.10.x variant.

The intent of this PR is to set the specifications in our builds to use CMake 3.10.3 (SEMS) and to trigger some builds so that we can see the errors and discuss them in issue #8387 

## Related Issues
* Part of #8387 

## Additional Information
This PR shouldn't be merged, once I get a test set I'll close it out and we can link to it from the issue as needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,3 +105,5 @@ ENDIF()
 IF(${PROJECT_NAME}_ENABLE_YouCompleteMe)
   INCLUDE(CodeCompletion)
 ENDIF()
+
+# a simple text change to trigger testing. DO NOT MERGE!
--- a/cmake/std/pr_config/pullrequest.ini
+++ b/cmake/std/pr_config/pullrequest.ini
@@ -244,7 +244,7 @@ module-load sems-hdf5: 1.10.6/base
 module-load sems-netcdf: 4.7.3/base
 module-load sems-metis: 5.1.0/base
 module-load sems-superlu: 4.3/base
-module-load sems-cmake: 3.12.2
+module-load sems-cmake: 3.10.3
 use ATDM-ENV:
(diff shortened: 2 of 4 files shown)


william76:
@prwolfe any idea why the autotester didn't run this through a run?  It looks like it's wanting a pre-test approval which doesn't make sense since I'm on the Trilinos team.

prwolfe:
There are no autotester comments which is suspicious to begin with. Lookat the the Jenkins server all jobs are failing with

07:49:10 fatal: unable to access 'https://prwolfe:rJpyCM7tesqsjFA6QUsq@gitlab-ex.sandia.gov/SEMS/sems-workflow-autotesting.git/': Failed connect to gitlab-ex.sandia.gov:443; Connection timed out

so we have an issue with gitlab-ex. Which is odd because I can log into it just fine.

Paul

prwolfe:
This turned out to be a proxy setting change. Again. They are all running now, but I'm sure it will take some time to catch up.

prwolfe:
@william76 - it looks like this is failing due to a ninja installation that does not support fortran. I looked at our configure *ini and found that clang is the only build using "module load sems-ninja_fortran/1.10.1" with the exception go gcc 8.3.0 which then does "module load atdm-ninja_fortran/1.7.2". I would be checking that ninja installation to see if it contains the patches from kitware at https://github.com/Kitware/ninja/tree/features-for-fortran#readme

Paul

william76:
Closing this out. It is superseded by #8521

PR #8412, opened 2020-12-03
MueLu: Set Max Eig Value for Prolongator Smoother

rstumin:
Auto-PR for SHA 2be9185


mayrmt:
@rstumin Can this new parameter be included into testing?

rstumin:
> @rstumin Can this new parameter be included into testing?

@mayrmt @jhux2  I suppose. This option makes most sense in conjunction with the 'sa: use rowsumabs diagonal scaling' option ... but whoever put in that more significant option didn't make a test for it

csiefer2 (approved):

PR #8414, opened 2020-12-03
Remove Trilinos warnings when building with gcc 7.2.0

kliegeois:
@trilinos/anasazi 
@mperego
@ikalash  

## Motivation

This PR removes 3 warnings which are treated as errors by the gcc 7.2.0 Debug build of the Trilinos autotester.

## Testing

This branch has been tested on Blake using a Debug build with gcc 7.2.0.


mperego (approved):
Looks good Kim! Thanks for doing all this work.
I'm not a fan of unsigned int (nor size_t) but that's more personal taste!

mperego:
> Looks good Kim! Thanks for doing all this work.
> I'm not a fan of unsigned int (nor size_t) but that's more personal taste!

B.t.w. My comment was misplaced. It was related to the related Albany PR, not this one. Ops..

PR #8415, opened 2020-12-03
Panzer: gcc 10 compiler fix

### LINKED ISSUE #8407: Panzer:  Mismatch in return type causing link failures with GNU 10.x
## Bug Report
@trilinos/\<Panzer\>

Attention to: @rppawlo , @eric-c-cyr 

### Description
When compiling the code with the GNU 10.x compiler and `-std=c++11`, or latter probably, a mismatch in return types is causing an unresolved symbol failure during the link.

### Steps to Reproduce
Compile Charon with the GNU 10.x compiler. 

The problem is that for GNU 10.x anything that has a `std::string` in it gets an additional tag, `abi:cxx11` added to it in the symbol table of the resulting object file when compiled. You can read about why that happens around the web. It allowed some non-backward compatible changes when the standard was finalized or some such. Anyway, it's clearly a bug as it relies on the erroneous fact that
```
std::vector<int>::size_type = std::vector<std::string>::size_type
```
That's absolutely true and that's allowed it to work up until now. But GNU 10 is adding some extra cruft to the symbol due to the `std::string` and so, in effect, the statement above is no longer true and the compiler is generating two versions of `getIntegrationRuleIndex()`, one with a `abi:cxx11` added and one without and that causes a link failure because only one is defined. 

Here is a possible fix, but I didn't care to follow the logic down deep enough to find out what type is actually needed for the `::size_type`. Is it really a `std::vector<std::string>` size or a `std::vector<int>` size?

And then again, maybe the right solution is to get rid of the two argument function altogether and make us change Charon up for the three argument function?

Anyway, diff for possible fix
```diff
diff --git a/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.cpp b/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.cpp
index 421e48594d..78c33f48b0 100644
--- a/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.cpp
+++ b/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.cpp
@@ -210,7 +210,7 @@ namespace panzer {
     WorksetDetailsAccessor wda;
     return getBasisIndex(basis_name, workset, wda);
   }
-  std::vector<int>::size_type
+  std::vector<std::string>::size_type
   getIntegrationRuleIndex(int ir_degree, const panzer::Workset& workset) {
     WorksetDetailsAccessor wda;
     return getIntegrationRuleIndex(ir_degree, workset, wda);
diff --git a/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.hpp b/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.hpp
index 771413fa82..69ed34a5ac 100644
--- a/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.hpp
+++ b/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.hpp
@@ -66,7 +66,7 @@ namespace panzer {
   std::vector<std::string>::size_type 
   getBasisIndex(std::string basis_name, const panzer::Workset& workset, WorksetDetailsAccessor& wda);
 
-  std::vector<int>::size_type
+  std::vector<std::string>::size_type
   getIntegrationRuleIndex(int ir_degree, const panzer::Workset& workset, WorksetDetailsAccessor& wda);
 
   void printWorkset(std::ostream& os, const panzer::Workset & workset, WorksetDetailsAccessor& wda);
@@ -74,7 +74,7 @@ namespace panzer {
   // Temporarily provide non-wda versions so that Charon continues to build and work.
   std::vector<std::string>::size_type getPureBasisIndex(std::string basis_name, const panzer::Workset& workset);
   std::vector<std::string>::size_type getBasisIndex(std::string basis_name, const panzer::Workset& workset);
-  std::vector<int>::size_type getIntegrationRuleIndex(int ir_degree, const panzer::Workset& workset);
+  std::vector<std::string>::size_type getIntegrationRuleIndex(int ir_degree, const panzer::Workset& workset);
   void printWorkset(std::ostream& os, const panzer::Workset & workset);
 }
```

rppawlo:
## Motivation

fixes #8407 

closes #8407 

## Stakeholder Feedback

Verified fixes Charon builds with gcc 10.

## Testing

All panzer tests use this feature.

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.cpp
+++ b/packages/panzer/disc-fe/src/Panzer_Workset_Utilities.cpp
@@ -93,7 +93,7 @@ namespace panzer {
     return std::distance(wda(workset).basis_names->begin(), basis);
   }
 
-  std::vector<std::string>::size_type
+  std::vector<int>::size_type
   getIntegrationRuleIndex(int ir_degree, const panzer::Workset& workset, WorksetDetailsAccessor& wda)
   {
     std::vector<int>::iterator ir;


etphipp (approved):
Interesting.  What was the compiler error?

rppawlo:
> Interesting. What was the compiler error?

During the link it could not find the function for getIntegrationRuleIndex(...).

PR #8416, opened 2020-12-03
Zoltan2 Fixed case of Teuchos parameter

kddevin:
@trilinos/zoltan2 
Fixed case of parameter.

## Motivation

Sigh.  Teuchos parameters are still case sensitive, after all these years.

## Testing

Tested on mac with clang


egboman (approved):
A one character bug fix! Unbeatable record 👍

PR #8411, opened 2020-12-02
Zoltan2:  first implementation of Tpetra CrsMatrix colorer

kddevin:
@trilinos/zoltan2   #8356 @etphipp @mperego 

## Motivation

This PR provides a first implementation of Zoltan2's Tpetra CrsMatrix colorer.
It is an intermediate step toward the final product.  It is suitable for friendly application use.

Currently, the code defaults to using distance-2 coloring and Zoltan's graph symmetrization (TRANSPOSE).

Remains to be done:
-  make partial distance-2 coloring with bipartite symmetrization work correctly; currently, tests with non-contiguous maps fail this case
-  incorporate Zoltan2's Kokkos-enabled coloring (distance-2)
-  add partial-distance-2 coloring to Zoltan2.

Current Teuchos parameters for using the colorer:
-  "symmetric" can give a hint to the colorer that the matrix is symmetric, avoiding symmetrization; default = false
-  "matrixType" specifies whether the matrix is a Jacobian or Hessian; right now, this is really a mathy code name for "symmetric," as Jacobians are assumed to be nonsymmetric while Hessians are symmetric.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of #8356 
* Composed of 

## Testing

New test zoltan2/test/core/TpetraCrsColorer.cpp demonstrates usage and exercises various cases (symmetric, nonsymmetric inputs, distributed and non-distributed matrices, etc.).
Tested on mac with clang.

Issue #8356, opened 2020-11-16
Zoltan2:  Add utility to compute seed matrix for automatic differentiation based on graph coloring


kddevin:
## Enhancement
@trilinos/zoltan2 @trilinos/tpetra

Add a utility to compute the seed matrix for automatic differentiation.  
Given a Tpetra CrsMatrix, compute the Seed matrix in a Tpetra::MultiVector.
Use partial distance-2 coloring when available; use distance-1 coloring of modified matrix if not.

@egboman @etphipps

kddevin:
I added the colorer classes to Zoltan2; for now, they are on branch zoltan2_8356

An example using the colorer is here: https://github.com/trilinos/Trilinos/blob/zoltan2_8356/packages/zoltan2/test/core/TpetraCrsColorer/TpetraCrsColorer.cpp .  The interface is a little different from Eric's original design.  I moved the choice of coloring algorithm into the class, rather than having the user decide whether to use a Zoltan or Zoltan2 colorer object.  (The user can still specify that choice through a parameter, but the code will assign a default algorithm.)  

A few questions for @etphipp :

*    In the Colorer interface, about which functions does an application care?  Is the application interested in getting the actual colors, the seed matrix, and/or the compressed matrix? 
*    Is the reconstructMatrix operation useful only for testing, or would applications want it, too?
*    Is having computeSeedMatrixFitted in the public interface useful?  I plan to make just one public interface, computeSeedMatrix.  I would move the test for isLocallyFitted to computeSeedMatrix and do the import there only if needed.  Then there would be fewer chances that the users pick the wrong interface.  (Similar plan for reconstructMatrix.)
 *   Would users ever use the same Colorer object for more than one coloring (say, a second coloring with different parameters)?  If not, I will eliminate method computeColoring and move its work to the constructor.

A few questions for @egboman :

*    Does the logic for selecting the coloring method look correct? https://github.com/trilinos/Trilinos/blob/9881c99b085a1b8cab6c7da54a9728f03d2e2a1d/packages/zoltan2/core/src/TpetraCrsColorer/Zoltan2_TpetraCrsColorer.hpp#L207
 *   Do the parameters look sufficient to handle the cases (matrix type, symmetrization, etc.) that you care about?  What other parameters do we need?
*    What other tests are needed?  I know we still need to exercise the different libraries (zoltan vs zoltan2); those tests will be more relevant when Ian's work is available.  #8363 

Thanks, @etphipp and @egboman

[...]

PR #8417, opened 2020-12-03
Zoltan2:  fix map used for transpose_graph in TpetraCrsColorer

kddevin:
@trilinos/zoltan2 
#8356 #8411 #8416 

## Motivation

The TpetraCrsColorer should work with Domain and Range maps that are not the default Trilinos contiguous map.
This PR re-enables the tests that use cyclic Domain and Range maps.

This PR changes the Range map used by the transpose_graph when bipartite symmetrization is used.
Normally, for a transpose graph, we'd use
-  original matrix's RangeMap as the DomainMap, and
-  original matrix's DomainMap as the RangeMap.

Moreover, we're not doing SpMV with the transpose graph, so
you might think we wouldn't care about RangeMap and DomainMap.
But we DO care, because exportAndFillCompleteCrsGraph uses the
shared (overlapped) transpose matrix's RangeMap as the RowMap of the
transpose matrix, while the Zoltan callbacks are assuming the
transpose matrix's RowMap is the same as the original matrix's.
So this PR uses the original matrix's RowMap as the RangeMap.

Teuchos parameter 
-  symmetrize = true tells the Zoltan colorer to use Zoltan's TRANSPOSE symmetrization and distance-2 coloring; 
-  symmetrize = false tells Zoltan to use bipartite symmetrization and partial-distance-2 coloring.  

RIght now, the default is symmetrize=true.   Let me know if you want me to change the default, or eliminate this parameter all together.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of  #8356 #8411 #8416 
* Composed of 

## Testing

Tested on Mac with clang.
Re-enabled test cases using cyclic maps for Range and Domain maps.

### DIFF
--- a/packages/zoltan2/core/src/TpetraCrsColorer/Zoltan2_TpetraCrsColorerUtils.hpp
+++ b/packages/zoltan2/core/src/TpetraCrsColorer/Zoltan2_TpetraCrsColorerUtils.hpp
@@ -129,9 +129,20 @@ compute_transpose_graph(const Tpetra::CrsGraph<LO, GO, NO> &graph)
 
   // Build (possibly overlapped) transpose graph using original graph's
   // column map as the new row map, and vice versa
+  //
+  // Normally, for a transpose graph, we'd use
+  //     original matrix's RangeMap as the DomainMap, and
+  //     original matrix's DomainMap as the RangeMap.
+  // Moreover, we're not doing SpMV with the transpose graph, so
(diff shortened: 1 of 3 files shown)


egboman (approved):
Good catch and fix, @kddevin !
I agree to have transpose=true for now, but really we should make false the default later after the bipartite option has been debugged and tested.

PR #8427, opened 2020-12-05
Tpetra: Fix modify/sync issues in MultiVector and unit tests

### LINKED ISSUE #8361: Tpetra:  MultiVector failing norm tests without UVM
## Bug Report
@trilinos/tpetra 

### Description
Without UVM, some MultiVector unit tests that compute norms are generating incorrect values:
* 7 OrthoDot
* 17 ScaleAndAssign

After discussing as a team, I believe we have high confidence that the KokkosBlas calls are doing the right thing, both numerically and with the memory spaces they use.

tasmith4:
* Change behavior of replaceLocalValue/sumIntoLocalValue to call modify_host and sync_host internally, which feels safer for users.
* Remove a modify_device call inside multiVectorSingleColumnDot (since a dot product shouldn't change the values of the input vectors).
* Also fixes some const inconsistencies and updates some debug ifdefs to use the Behavior class instead.

@trilinos/tpetra 

## Related Issues

* Closes #8361 
* Related to #8209 

## Testing
MultiVector unit tests now passing in a non-UVM build on ascicgpu24.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
@@ -842,7 +842,7 @@ namespace Tpetra {
     void
     replaceGlobalValue (const GlobalOrdinal gblRow,
                         const size_t col,
-                        const impl_scalar_type& value) const;
+                        const impl_scalar_type& value);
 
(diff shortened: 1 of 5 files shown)


brian-kelley (approved):
Good changes.

Issue #8431, opened 2020-12-07
Piro: introduced a required dependency on ROL.


rppawlo:
## Bug Report
@trilinos/piro

### Description
Both Charon and EMPIRE are reporting failures due to new ROL requirement in Piro. 

```
00:03:35.172   105 - PanzerAdaptersSTK_solver_MPI_4 (Failed)
00:03:35.172   166 - PanzerAdaptersSTK_main_driver_energy-ss (Failed)
00:03:35.172   168 - PanzerAdaptersSTK_main_driver_energy-ss-blocked (Failed)
00:03:35.172   170 - PanzerAdaptersSTK_main_driver_energy-ss-blocked-tp (Failed)
00:03:35.172   171 - PanzerAdaptersSTK_main_driver_energy-periodic-32bit-limit (Failed)
00:03:35.172   172 - PanzerAdaptersSTK_main_driver_energy-neumann (Failed)
00:03:35.172   175 - PanzerAdaptersSTK_periodic_wedge (Failed)
 
00:01:49.758  *********** Caught Exception: Begin Error Report ***********
00:01:49.758  /scratch/empire-jenkins/trilinos-jobs/clang-opt-serial-static/src/Trilinos/packages/piro/src/Piro_SteadyStateSolver_Def.hpp:1256:
00:01:49.758  
00:01:49.758  Throw number = 1
00:01:49.758  
00:01:49.758  Throw test that evaluated to true: true
00:01:49.758  
00:01:49.758  
00:01:49.758  Piro::SteadyStateSolver::evalReducedHessian():  This fuction requires ROL.
00:01:49.758  
00:01:49.758  ************ Caught Exception: End Error Report ************
 ```

Neither code uses ROL or wants to build it. Can piro be cleaned up for this case?

@mperego @kliegeois @etphipp - we believe this was due to recent changes in Hessian support.

@jmgate @glhenni 

### Steps to Reproduce
1. SHA1:  [insert here]
1. Configure script:  [attach here]
1. Configure log:  [attach here]
1. Build log:  [attach here]
1. Input deck:  [attach here]
1. Do this.
1. Do that.
1. Shake fist angrily at computer.
1. Run log:  [attach here]

etphipp:
@kliegeois, @mperego  Did we introduce this dependency intentionally?

kliegeois:
We introduced a capability in Piro which requires ROL.
I will fix that tomorrow (Belgian time) and create a PR. 
This should be an easy fix.

[...]

PR #8432, opened 2020-12-07
Piro: call ReducedHessian only when required by the outArgs

mperego:
@trilinos/piro 

## Motivation
Colses #8431


kliegeois (approved):
This looks good to me. Thanks for your quick fix!

Issue #8429, opened 2020-12-05
Can't reproduce non-CUDA PR builds on SNL COE RHEL7 machines with just the SEMS modules


bartlettroscoe:
@trilinos/framework,

FYI: It looks like you can no longer reproduce the non-CUDA PR builds on an SNL COE RHEL7 machine using just the SEMS modules.  Instead, it appears that you must be on a CEE LAN RHEL7 machine as now the Python being used in 3.6.10 which is using a Sierra module.  See https://github.com/trilinos/Trilinos/pull/7508#issuecomment-739318388

How are developers supposed to reproduce PR builds if they don't have a CEE LAN account and can't get one (i.e. if you can't get an SRN account)?

Also note that standard CEE LAN RHELl7 machines are not installed with BLAS and LAPACK and therefore you can't reproduce any of the non-CUDA PR builds on those machines either.  See https://github.com/trilinos/Trilinos/pull/7508#issuecomment-739073829.

bartlettroscoe:
@trilinos/framework,

Even on a CEE LAN RHEL7 machine, you can't reproduce the non-CUDA PR builds.  It selects the wrong Python compared to the version shown in PR testing (for example [here](https://testing.sandia.gov/cdash/build/8338228/configure) for the 'gcc_7.2.0_debug') which shows:

```
-- Found PythonInterp: /projects/sierra/linux_rh7/install/Python/3.6.10/bin/python (found suitable version "3.6.10", minimum required is "2.6") 
-- PYTHON_EXECUTABLE='/projects/sierra/linux_rh7/install/Python/3.6.10/bin/python'
```

But when I try to reproduce this locally using Trilinos 'develop' as of commit fb03550 on my CEE LAN RHEL7 machine (just updated to RHEL7 from RHEL6 last month) using the directory setup:

```
$ cd /scratch/rabartl/Trilinos.base/BUILDS/PR/gcc_7.2.0_debug/

$ ls -l load-env.sh 
lrwxrwxrwx 1 rabartl rabartl 87 Dec  4 12:10 load-env.sh -> /scratch/rabartl/Trilinos.base/Trilinos/cmake/std/sems/PullRequestGCC7.2.0TestingEnv.sh

$ cat do-configure
TRILINOS_DIR=/scratch/rabartl/Trilinos.base/Trilinos
rm -r CMake*
cmake \
-G Ninja \
[...]

PR #8442, opened 2020-12-09
Issue 8429

william76:
@trilinos/framework 
@bartlettroscoe 

## Motivation
With the updates to the PR drivers, the scripts to [reproduce PR errors][1] have become outdated. Especially where Python has been concerned. The main problem it appears from #8429 is that these scripts encounter a problem trying to load the right version of Python which conflicts with the work we put in to remove the use of Python 2.x from our PR scripts.

One problem is that loading the SEMS boost module also loads the `sems-python/2.7.x` module as a side-effect, so when we tried to load a Python 3.x module we got conflicts. 

I went in and modified the non CUDA scripts to perform a `module purge` in the beginning as well as a `module unload sems-python` before we load`sems-python/3.5.2`.  

## Related Issues
* Related to #8429 

## Testing
Ran tests on a RHEL COE system using a simple script:
```bash
#!/usr/bin/env bash

TRILINOS_DIR=${HOME:?}/dev/trilinos-dev/source/Trilinos

build_name="Trilinos_pullrequest_gcc_8.3.0"
env_file="PullRequestGCC8.3.0TestingEnv.sh"
cmake_file="PullRequestLinuxGCC8.3.0TestingSettings.cmake"

echo ""
echo "$ source ${TRILINOS_DIR}/cmake/std/sems/${env_file}"
echo ""
source ${TRILINOS_DIR}/cmake/std/sems/${env_file}

module list

rm -rf build-${build_name}
mkdir build-${build_name}
cd build-${build_name}

cmake_opts=(
    -C ${TRILINOS_DIR}/cmake/std/${cmake_file}
    -DTrilinos_ENABLE_Teuchos=ON
    ${TRILINOS_DIR}/.
    )

cmake ${cmake_opts[@]}

make -j 20
```
for these builds:
- [x] `Trilinos_pullrequest_gcc_8.3.0`
- [x] `Trilinos_pullrequest_clang_10.0.0`

They are still running but I'll go ahead and get this PR in so we can get a review from @bartlettroscoe and iterate over the fix.

[1]: https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors

### DIFF
--- a/cmake/std/sems/PullRequestClang10.0.0TestingEnv.sh
+++ b/cmake/std/sems/PullRequestClang10.0.0TestingEnv.sh
@@ -9,13 +9,13 @@
 # $ module purge
 # or Trilinos/cmake/unload_sems_dev_env.sh
 
+module purge
+
 source /projects/sems/modulefiles/utils/sems-modules-init.sh
 
-module load sems-git/2.10.1
 module load sems-gcc/5.3.0
 module load sems-clang/10.0.0
 module load sems-openmpi/1.10.1
-module load sems-python/2.7.9
 module load sems-boost/1.69.0/base
 module load sems-zlib/1.2.8/base
 module load sems-hdf5/1.10.6/parallel
@@ -23,8 +23,14 @@ module load sems-netcdf/4.7.3/parallel
 module load sems-parmetis/4.0.3/parallel
 module load sems-scotch/6.0.3/nopthread_64bit_parallel
 module load sems-superlu/4.3/base
+
 module load sems-cmake/3.17.1
 module load sems-ninja_fortran/1.10.0
+
+module load sems-git/2.10.1
+
+module unload sems-python
+module load sems-python/3.5.2
 
 # add the OpenMP environment variable we need
 export OMP_NUM_THREADS=2
--- a/cmake/std/sems/PullRequestClang7.0.1TestingEnv.sh
+++ b/cmake/std/sems/PullRequestClang7.0.1TestingEnv.sh
@@ -9,13 +9,14 @@
 # $ module purge
 # or Trilinos/cmake/unload_sems_dev_env.sh
 
+module purge
+
 source /projects/sems/modulefiles/utils/sems-modules-init.sh
 
 module load sems-git/2.10.1
 module load sems-gcc/5.3.0
 module load sems-clang/7.0.1
 module load sems-openmpi/1.10.1
-module load sems-python/2.7.9
 module load sems-boost/1.63.0/base
 module load sems-zlib/1.2.8/base
 module load sems-hdf5/1.10.6/parallel
@@ -23,7 +24,6 @@ module load sems-netcdf/4.7.3/parallel
 module load sems-parmetis/4.0.3/parallel
 module load sems-scotch/6.0.3/nopthread_64bit_parallel
 module load sems-superlu/4.3/base
-module load sems-cmake/3.12.2
 
 # Load the SEMS CMake Module
 # - One of the SEMS modules will load CMake 3.4.x also,
@@ -40,6 +40,9 @@ module load sems-cmake/3.12.2
 module load atdm-env
 #module load atdm-cmake/3.10.1
 module load atdm-ninja_fortran/1.7.2
+
+module unload sems-python
+module load sems-python/3.5.2
 
 # add the OpenMP environment variable we need
 export OMP_NUM_THREADS=2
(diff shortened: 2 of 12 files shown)


prwolfe (approved):

bartlettroscoe (commented):
@william76, this looks okay but does this now mean that Trilinos is no longer supporting Python 2.7? It looks like all of the PR builds are now using Python 3.5?  If that is a requirement for Trilinos to use Python 3.5+, then that should be codified as described in:

* https://tribits.org/doc/TribitsDevelopersGuide.html#python-support

So Trilinos should set `PythonInterp_FIND_VERSION=3.5` if that is indeed the minimum supported version of Python.

If not, at least one PR build needs to run the inner Trilinos CMake configure, build, and test with Python 2.7.

bartlettroscoe:
@william76, for the purposes of reporting and browsing, it is good to give a PR a better name name than "Issue 8429".  When you look at an issue/pr query, that name is meaningless.

Instead, how about a name like:

* "Change PR builds to use sems-python/3.5.2 (#8429)"

?

william76:
> @william76, this looks okay but does this now mean that Trilinos is no longer supporting Python 2.7? It looks like all of the PR builds are now using Python 3.5? If that is a requirement for Trilinos to use Python 3.5+, then that should be codified as described in:
> 
>     * https://tribits.org/doc/TribitsDevelopersGuide.html#python-support
> 
> 
> So Trilinos should set `PythonInterp_FIND_VERSION=3.5` if that is indeed the minimum supported version of Python.
> 
> If not, at least one PR build needs to run the inner Trilinos CMake configure, build, and test with Python 2.7.

There is a bit of a nuance with the Python requirement -- we're primarily just addressing the python requirements of the PR driver infrastructure itself w/rt to the 3.x requiremet, not so much the Trilinos project as a whole (we _should_ move everything up to a Python 3 requirement but that's probably a conversation we should have with @jwillenbring and @prwolfe to look into what would be necessary for enforcing a Trilinos-wide requirement on Python 3.

I had hoped we wouldn't need to mess with the Python requirements for the replication scripts since we don't generally run the PR driver itself when replicating builds. TBH, the whole replication process needs to be revamped -- possibly part of the ATDM migration work? 

Python hit end of life on Jan 1, 2020 so it's time to move away from Python 2.x. (https://www.python.org/doc/sunset-python-2/) 

@bartlettroscoe I do not have the ability to edit the TribitsDevelopersGuide...

bartlettroscoe:
> I do not have the ability to edit the TribitsDevelopersGuide...

@william76, the link I gave at:

* https://tribits.org/doc/TribitsDevelopersGuide.html#python-support

just describes how TriBITS projects can specify their dependence on Python.  If you see typos, then you can post a PR to the TriBITS repo.  The source is at:

* https://github.com/TriBITSPub/TriBITS/tree/master/tribits/doc/developers_guide

Instructions on how to build are at:

* https://github.com/TriBITSPub/TriBITS/blob/master/tribits/doc/developers_guide/README

bartlettroscoe:
CC: @rppawlo, @william76, @trilinos/framework 

Looking at the configure output for the last passing PR iteration shown [here](https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-8442-test&field2=buildstarttime&compare2=84&value2=NOW) shows the Python versions:

* clang_10.0.0: /projects/sems/install/rhel7-x86_64/sems/compiler/python/2.7.9/bin/python
* cuda_9.2: /home/projects/ppc64le/python/3.7.3/bin/python
* gcc_7.2.0_debug: /projects/sierra/linux_rh7/install/Python/3.6.10/bin/python
* gcc_7.2.0_serial: /projects/sems/install/rhel7-x86_64/sems/compiler/python/2.7.9/bin/python
* gcc_8.3.0: /projects/sems/install/rhel7-x86_64/sems/compiler/python/2.7.9/bin/python
* intel_17.0.1: /projects/sierra/linux_rh7/install/Python/3.6.10/bin/python
* python_3: /projects/sierra/linux_rh7/install/Python/3.6.10/bin/python3

So Trilinos is getting tested with Python 2.7l.9 and various Python 2 and 3.x.

However, it looks like non-SEMS modules are getting used with several builds:

*  /projects/sierra/linux_rh7/install/Python/3.6.10/bin/python

So can we fix that as well?

PR #8450, opened 2020-12-10
Turn CUDA_LAUNCH_BLOCKING OFF for EMPIRE

jmgate:
## Motivation
EMPIRE just went through a lot of work to be able to build with `CUDA_LAUNCH_BLOCKING` turned off.  Now we want to make sure we don't accidentally regress, so we're turning it off for all EMPIRE users/developers/CI services.


jmgate:
@e10harvey, @bartlettroscoe, is this the right way to do this in the right place?

e10harvey:
> @e10harvey, @bartlettroscoe, is this the right way to do this in the right place?

@jmgate: IIRC, `CUDA_LAUNCH_BLOCKING` is an environment variable not a CMake cache variable. I think you could unset this here in CMake via `unset(ENV{CUDA_LAUNCH_BLOCKING})`. @bartlettroscoe, Is that an acceptable approach?

jmgate:
Thanks, @e10harvey.  PR updated.

cmake/std/atdm/apps/empire/EMPIRETrilinosEnables.cmake
```diff
@@ -12,6 +12,9 @@ ATDM_SET_CACHE(Gtest_SKIP_INSTALL TRUE CACHE BOOL)
 
 ATDM_SET_CACHE(Trilinos_ENABLE_Gtest FALSE CACHE BOOL)
 
+# Always set CUDA_LAUNCH_BLOCKING off for EMPIRE.
+unset(ENV{CUDA_LAUNCH_BLOCKING})
```
bartlettroscoe, line 16: I don't thinks this actually does what you want, does it?  This only unsets the env var `CUDA_LAUNCH_BLOCKING` when configuring Trilinos.  This will not impact running of EMPIRE tests.

bartlettroscoe (changes requested):
CC: @rppawlo, @bathmatt 

@jmgate, did you actually test this? I don't think there is any way this can work they way you want.

There are a few options to fixing this for EMPIRE.  Can we create a new GitHub issue to do a requirements analysis and discuss there the different options for keeping the Trilinos CUDA tests passing but still unset this var when running EMPIRE tests?

cmake/std/atdm/apps/empire/EMPIRETrilinosEnables.cmake
jmgate: Ah, so changing this CMake fragment file won't make the change propagate to the `load_matching_env.sh` that gets installed?  That was the assumption I was working under.

> did you actually test this? 

No, I just created the PR to start the conversation.

> Can we create a new GitHub issue&hellip;?

Let's just keep the conversation here.  If this is going to be a substantial amount of work to make the change on the Trilinos side, we'll just make it on the BuildScripts side instead in the near term.

bartlettroscoe: @jmgate, I think an issue would be better to separate it from different implementation options in different PRs.  If there is any requirements analysis that needs to happen, then that should happen in an issue.

jmgate:
We wound up implementing this change on our end instead.

bartlettroscoe:
Related to:
* SEPW-215

PR #8453, opened 2020-12-10
Intrepid: fix bug in face tag initialization for high-order H(div) Triangle elements.

CamelliaDPG:
Note: this regards Intrepid, *not* Intrepid2.  Since there is no intrepid team as such, sending to intrepid2.

Fixes a bug in face tag initialization for high-order H(div) Triangle elements; some face tag entries were left uninitialized.  This was exhibiting as a hang during tag initialization for a Camellia user, when attempting to use a degree-3 H(div) triangle basis.

@trilinos/intrepid2


mperego (approved):
It looks good. Thanks for fixing it.

PR #8456, opened 2020-12-10
Teko use default ordinal types

### LINKED ISSUE #8455: Teko:  Doesn't build against certain GO/LO types from Tpetra
## Bug Report
@trilinos/teko 

### Description
Teko has a fairly narrowly defined set of LO/GO types it can use from Tpetra. This is problematic, switch it to use the default.

### Steps to Reproduce
1. SHA1:  [insert here]
1. Configure script:  [attach here]
1. Configure log:  [attach here]
1. Build log:  [attach here]
1. Input deck:  [attach here]
1. Do this.
1. Do that.
1. Shake fist angrily at computer.
1. Run log:  [attach here]

eric-c-cyr:
Resolves #8455 

@trilinos/teko


eric-c-cyr:
@rppawlo Could you look at this @vbrunini asked me to change how the LO/GO's were set in Teko.

rppawlo (approved):
Looks good to me!

rppawlo:
Only comment is that this Config file pulls in the Map headers which is quite a bit to parse when you might only want the ordinals. To avoid parsing these large files in all of your library objects, you could switch to pulling in a much smaller file - here's what I did in panzer:
```
#include "Tpetra_Details_DefaultTypes.hpp" // for ordinal types
...
using GlobalOrdinal = Tpetra::Details::DefaultTypes::global_ordinal_type;
using LocalOrdinal = int;
```
Mark had planned to pull the ordinal types out of this "Details" file and put it into a separate interface file to support this. This isn't required as I see the Map was already being pulled into the config file before this commit. Just wanted to mention.

PR #8459, opened 2020-12-11
Tpetra: Remove some MultiVector unit tests with an anti-pattern

tasmith4:
@trilinos/tpetra
 
## Motivation

The MultiVector reduce_strided unit tests test various methods for performing the reduce operation on a MultiVector.  Some tests were directly calling reduce on a MultiVector with strided memory in the underlying DualView.  This worked with UVM because UVM memory is accessible from both host and device.  When the Cuda memory is distinct from host in the DualView, doing a deep_copy on strided memory is an error, so these tests were wrong and have been removed.

The correct pattern is to first copy the MultiVector, then do the reduce on the copy (which has contiguous memory).  Other tests in this file which demonstrate this pattern have been kept.

## Related Issues

* Related to #8209 

## Testing

This PR is a cleanup on the test suite itself.

### DIFF
--- a/packages/tpetra/core/test/MultiVector/MV_reduce_strided.cpp
+++ b/packages/tpetra/core/test/MultiVector/MV_reduce_strided.cpp
@@ -286,78 +286,6 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL( MultiVector, reduce_strided, Scalar, LocalOrd
   TEST_ASSERT( Z1_Z3_equal );
 
   {
-    // Make sure Z4 has a stride greater than its number of rows.
-    const size_t Z4_stride = static_cast<size_t> (lclNumRows + 31);
-    dual_view_type Z4_dv_extra ("Z4", Z4_stride, numCols);
-    auto Z4_dv = Kokkos::subview (Z4_dv_extra,
-                                  pair_type (0, lclNumRows),
-                                  pair_type (0, numCols));
-    TEST_ASSERT( LO (Z4_dv.extent (0) == lclNumRows ) );
-    TEST_ASSERT( LO (Z4_dv.extent (1) == numCols ) );
-    TEST_ASSERT( LO (Z4_dv.d_view.extent (0) == lclNumRows ) );
-    TEST_ASSERT( LO (Z4_dv.d_view.extent (1) == numCols ) );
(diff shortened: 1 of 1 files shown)


tasmith4:
Looked into this some more, and I now believe there are actually two issues:
* **(Z6 test -- fails at all proc counts)** The test calls a sync on a noncontiguous sub-view before calling reduce().  This is illegal because Kokkos does not allow a deep_copy with both a layout and memory space change.  As noted above, this test should almost certainly be deleted.
* **(Z4 test -- fails for proc counts > 1)** The test calls reduce() on a MultiVector with a noncontiguous view.  In Tpetra_Details_allReduceView.hpp, at lines 213-214, the code realizes the view is noncontiguous, so makes temporary buffers.  However, the input view is a device view (see Tpetra_MultiVector_def.hpp, line 4128), so when allReduceView deep copies the noncontiguous input view into the contiguous temp buffer at line 215, it is attempting to copy across both a layout and memory space change.

Some ideas for solving the Z4 test are:
1. Assert that the user should not be calling reduce() on a noncontiguous view and delete the test.
2. Allocate the temporary contiguous buffer in the same memory space as the noncontiguous view.  May still need to rework the logic in allReduceView to add another deep_copy across memory spaces when the MPI is not Cuda-aware.

@trilinos/tpetra, any thoughts on the best path forward here?

kddevin:
@jhux2 @csiefer2  @rppawlo Can you comment on use-cases in the solvers or analysis tools for non-contiguous views?   Would removing the capability for reduce() on non-contiguous views cause troubles for solvers and other users?

rppawlo:
@kddevin - is this the reduce() method that is only valid for locally replicated vectors? If so, I don't know of anything on the nonlinear solver side. cc'ing others that might know better: @etphipp @egphill @dridzal @hkthorn .

etphipp:
> @kddevin - is this the reduce() method that is only valid for locally replicated vectors? If so, I don't know of anything on the nonlinear solver side. cc'ing others that might know better: @etphipp @egphill @dridzal @hkthorn .

I don't know of any use-cases for this.  Honestly I didn't even know that you could do this with Tpetra vectors.

csiefer2:
I never use non-contiguous views, so...

csiefer2 (approved):

tasmith4:
Superseded by #8821

PR #8465, opened 2020-12-14
Panzer: fix race condition with launch blocking off in GED accessor.

rppawlo:
## Motivation

The empire sync with Trilinos failed with launch blocking off due to a single panzer test for the GED accessor. Turned out to be a race condition that needed fencing.

## Stakeholder Feedback

Fixes EMPIRE Trilinos update.

## Testing

Covered by the test: ```PanzerDiscFE_LinearObjFactory_Tests_MPI_2```


bathmatt (approved):
Not sure if both are needed, but they don't hurt..

rppawlo:
> Not sure if both are needed, but they don't hurt..

For the unit test, the one after putScalar was the culprit. The other probably isn't needed but wanted to be safe. It's is one I'll review when UVM removal gets underway.

bathmatt:
@rppawlo how long does the autotester take these days?

rppawlo:
> @rppawlo how long does the autotester take these days?

I've seen it take up to 6 hours for low level changes since it has to build and test many downstream packages. Panzer is at the top level so rebuilds are very fast - it can happen in under an hour. This PR took about an hour and is now merged.

Issue #7045, opened 2020-03-18
Tpetra:  Add MatrixMarket reader that uses 2D Cartesian matrix distribution


kddevin:
## Enhancement
@trilinos/tpetra 

Tpetra's MatrixMarket reader assumes a 1D (i.e., row-wise) matrix distribution.  @ndellingwood would like a reader that allows 2D Cartesian matrix distribution.  

The current MatrixMarket reader in Tpetra reads the matrix entirely on processor 0, constructs a CrsMatrix on processor 0, distributes relevant entries to each process, and constructs a distributed CrsMatrix with those entries.

For the 2D reader, I propose a different approach.  Processor 0 reads the entries in batches from the MatrixMarket file and broadcasts them to all processes.  Each processor keeps the entries that are assigned to it.  After all entries are broadcast, a distributed CrsMatrix will be constructed.  This algorithm would work for 1D distributions as well, and may be more scalable wrt memory use than the existing code.

mhoemmen:
We've proposed before the idea that if you know the row Map, Process 0 could send batches of entries directly to the process that owns them.  @aprokop looked into that too, I think.  This would work for any distribution; for 2-D distributions, each receiving process would need to reject entries not in its column Map ("column Map filtering"), so you would end up sending and receiving more data than necessary.  Alternately, Process 0 could use both the row Map and the column Map to decide where to send entries.

To figure out what process owns what entry, Process 0 could do a Directory lookup (`Map::getRemoteIndexList`).  However, that's an expensive collective if the Map is noncontiguous.  For 2-D distributions, both row and column Maps are usually noncontiguous.  (The only overlapping Map you can build with one of the contiguous constructors is a locally replicated Map.)  Alternately, you could start by distributing data according to the Directory Maps, then doing a big redistribution at the end.

kddevin:
The challenge (I learned in my stand-alone reader) is that the rowMap and colMap describing the Cartesian distribution are not necessarily the rowMap and colMap that we want to use in the CrsMatrix.  So we use some sort of rowMap or colMap to determine which nonzeros a processor keeps based on the Cartesian layout, then build a new rowMap (and let fillComplete compute a new colMap) based on the actual nonzeros that were kept by each processor.

mhoemmen:
[...]

PR #8472, opened 2020-12-14
Tpetra:  MatrixMarket reader that supports 2D matrix distributions

kddevin:
@trilinos/tpetra  @seheracer @ndellingwood 

## Motivation

The existing Tpetra matrix market reader reads the entire matrix onto rank 0, fillCompletes it, and then exports it to other processors.  This approach is neither memory nor runtime scalable.  The existing approach also supports only Tpetra's default 1D block (row-based) distribution.

This new reader is more scalable in terms of memory usage and performance.  Rank 0 reads chunks of the matrix market file, and broadcasts them to all ranks.  Then each rank inserts the nonzeros that it should own into a Tpetra matrix.  fillComplete is called after all nonzeros have been read and broadcast.

This reader also supports multiple different distributions of the matrix: 1D block (row-wise linear), 1D random (random assignment of rows to ranks), 2D block, 2D random, LowerTriangularBlock (a special layout requested by @seheracer  for triangle counting algorithms) and MMFile (users specify the owner of each nonzero as the value in the MatrixMarket file).

The new reader is faster and more scalable than the old reader.  Here are runtimes for an RMAT matrix with 17M rows and 151M nonzeros:

compare

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of  #7045
* Composed of 

## Testing

Tested on Mac and Summit.
New test  TpetraCore_MatrixMarket_Tpetra_CrsMatrix_Dist added to exercise many options  

## Work remaining
Some work remains, but merging to develop now helps @ndellingwood 
- Replace std::map data structures with two-level hash table, with hopes of speeding execution for extremely large files and further reducing memory use.
- Add more tests of existing capabilities (symmetrizing, require diagonal entries, pattern matrices, MMFile option).
- Deprecate existing read_sparse_file interface, preferring one that takes a parameter list instead of an argument list of parameters.
- Remove existing limited implementation, instead calling this new implementation (part of deprecation effort).


packages/tpetra/core/inout/mmio_Tpetra.c
```diff
@@ -0,0 +1,382 @@
+/*
```
jhux2, line 1: @seheracer I couldn't find any info on what license governs this software.  Any idea what it is?

packages/tpetra/core/inout/mmio_Tpetra.h
```diff
@@ -0,0 +1,127 @@
+/*
```
jhux2, line 1: @seheracer I couldn't find any info on what license governs this software. Any idea what it is?

jhux2 (commented):
@seheracer Thanks for implementing this.  One thing that isn't clear (to me) is what the binary format is.  Is it just the standard MM format but translated to binary?

packages/tpetra/core/inout/MatrixMarket_TpetraNew.hpp
```diff
@@ -0,0 +1,986 @@
...
+//   while(std::getline(in, line)) {
+//     std::stringstream ss2(line);
+//     ss2 >> edge[0] >> edge[1];
+//     out.write((char *)edge, sizeof(ord_type)*2);
+//
+//    }
+//   in.close();
+//   out.close();
+//
+//////////////////////////////////////////////////////////////////////////////
```
seheracer, line 496: @jhux2 @kddevin These comment lines describe the binary format and a code snippet to convert a MM file into this format. It only supports unweighted graphs for now. So, there are two basic differences between MM and this format:
1) Header information consists of numVertices and numEdges, so this format can not support a rectangular matrix.
2) Values are not supported either, that is, each nonzero entry only consists of the source and destination vertices (i.e., rowId and colId).

jhux2: @seheracer So does this mean that the binary format only supports graphs right now, not numeric matrices?  Just want to be sure I understand.

seheracer: Correct. Numeric matrices are not supported by this format yet.

kddevin:
Thanks, @brian-kelley and @jhux2, for taking time for a walk-through of this code.  Here are my notes; let me know if I missed something that we discussed:

Before accepting the PR:
- add comments about what LowerTriangularBlock distribution looks like (including an ASCII picture as in 2D distribution case)
- move the bug fixes for replaceDiagonalCrsMatrix into a separate PR
- make sure the tests use a random vector x in the SpMV (to increase chances of catching errors)
- make clear that new reader assumes square matrices; add error check 

Before deprecating existing readSparseFile capabilities:
- add capability to read skew symmetric and Hermetian matrices (as in Teuchos:
```
143 typedef Teuchos::ScalarTraits<value_type> STS;
144 const value_type Aji = skew_ ?
145 value_type(-(conjugate_ ? STS::conjugate(Aij) : Aij)) :
146 (conjugate_ ? STS::conjugate(Aij) : Aij);
in Teuchos_MatrixMarket_SymmetrizingAdder.hpp
```
- add interface that does 1D distribution using user-provided maps
- support rectangular matrices and their range/domain vector distributions
- increase testing of all options in new reader (diagonal, pattern, skew symmetric, etc.)

Future extensions:
- matrix market writer that works with 2D distributions
- full matrix market support with binary reader (values, rectangular, etc.)

Outstanding questions:
-  Is LowerTriangularBlock distribution general enough to stay in Tpetra?  Or should there be a way for users to write their own distribution to be used by readSparseFile?
- What distribution should be used when users provide one file per processor containing the processor's nonzeros?  Should we have an "identity" distribution, that answers Mine=true for all nonzeros?  Is information (e.g., chunk boundaries) still needed from a distribution in this case and, if so, how are they computed?

kddevin:
@brian-kelley @jhux2  Thank you again for meeting to review this PR.

I have completed the four items needed to push the PR to develop.
- I added comments about what LowerTriangularBlock distribution looks like (including an ASCII picture as in 2D distribution case):  https://github.com/trilinos/Trilinos/pull/8472/commits/d199ac12b80d6ebfe8f430a9c35da4d9cd044234 and https://github.com/trilinos/Trilinos/pull/8472/commits/1e65185b91bf3e7901328ff76a4f4af60b80f5bf
- I moved the bug fixes for replaceDiagonalCrsMatrix into a separate PR #8600  and added a test that exercised the 2D distribution for replaceDiagonalCrsMatrix
- I made sure the tests of this new reader use a random vector x in the SpMV (to increase chances of catching errors):  https://github.com/trilinos/Trilinos/pull/8472/commits/3642da3d7ff2c3c03cffb9bac95d51fcbba4acae
- I made clear that, for now, the new reader assumes square matrices by adding an error check:  https://github.com/trilinos/Trilinos/pull/8472/commits/ceedf2c5dc3ee9e2222a3d6b47ab864be3784659

Also, to make the SpMV tests pass with the **sorted** LowerTriangularBlock, it was necessary to add a permutation matrix to the DistributionLowerTriangularBlock class.  When sorting is used, the LowerTriangularBlock decomposition stores a permuted matrix, not the original MatrixMarket matrix.  If the vectors assume the MatrixMarket numbering (as they likely would), the SpMV operator needs to permute the vectors to get the same results as if the original MatrixMarket matrix were stored.  The permutation matrix is used in the LowerTriangularBlockOperator.
https://github.com/trilinos/Trilinos/pull/8472/commits/1b41e16f5e83d33f1c6c9772292bf414f5313761 and
https://github.com/trilinos/Trilinos/pull/8472/commits/1e65185b91bf3e7901328ff76a4f4af60b80f5bf 

I admit that this level of algorithm-specific detail with the LowerTriangularBlock distribution argues for moving it to @seheracer 's application rather than keeping it in Tpetra.  I'd like to defer the move to a later PR if possible.

With these changes (and assuming the tests pass), are you OK with approving this PR?

@seheracer @ndellingwood

seheracer:
@kddevin Thanks for the PR! I am okay with approving the PR.

> I admit that this level of algorithm-specific detail with the LowerTriangularBlock distribution argues for moving it to @seheracer 's application rather than keeping it in Tpetra. I'd like to defer the move to a later PR if possible.

Can I assume that we will start working on the PR which moves the DistributionLowerTriangularBlock class to my triangle counting code after the Friday session?

seheracer (approved):

jhux2 (approved):
@kddevin With your changes, I approve!

brian-kelley (approved):
Looks good, this is a great addition.

PR #8473, opened 2020-12-14
Framework: Bugfix for cycle detection PR configurations

william76:
@trilinos/framework 

## Motivation
The cycle-detection logic in the PR driver `setenvironment` module had a minor bug that resulted in an overly aggressive 'cycle prevention' mode.

For example, section dependencies like this would be blocked:
```
A -> B -> C -> B
```

but we also blocked this:
```
A -> B -> C
A -> B
```

This fixes the issue by popping off the sections from the 'seen list' as we pop out of scope during the recursive module that scans through the configurations.

This generally didn't affect any of our configurations but I could see a scenerio where we do something like this:

```ini
[MODULE-MANUAL-CONFIG]
setenv ENVVAR1: foo
setenv ENVVAR2: bar
setenv ENVVAR3: baz

[SEMS-DEFAULT]
use MODULE-MANUAL-CONFIG:

[SOME_BUILD_CONFIG]
use SEMS-DEFAULT:
use MODULE-MANUAL-CONFIG:
```

Generally, I think we can avoid stuff like this by the order in which things are done -- but as configurations get more complicated we probably wouldn't want to avoid having the explicit call to MODULE-MANUAL-CONFIG in the `SOME_BUILD_CONFIG` section be ignored because it had already been seen in the default config -- especially if there is some module that has a side effect which might smash something.

Provenance
------------
This came out from some of the Python work... for various reasons we need to use the sierra python compiler for our cloud nodes due to some errors we had with the SEMS Python 3.5.x compiler in the PR environment. Internally, the sems-boost module also loads a sems-python 2.x which we need to fix.

It's easily fixable by just setting our Sierra Python envvars after we load sems-boost... but it's good to make sure the cycle detection routine works properly to eliminate future gotchas.

FYI @jmgate @prwolfe

### DIFF
--- a/cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
+++ b/cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
@@ -162,6 +162,9 @@ class SetEnvironment(object):
             IndexError: Invalid number of parameters for 'module' command
         """
         status = 0
+
+        self.actions["module-op"]
+
         print("Module Operations")
         print("+---------------+")
         for m in self.actions["module-op"]:
@@ -206,9 +209,10 @@ class SetEnvironment(object):
             k = envvar['key']
             v = envvar['value']
             print("[envvar] export {}={}".format(k,v))
-            v = self._expand_envvars_in_string(v)
-            os.environ[k] = v
-            print("[envvar] export {}={} (actual)".format(k,v))
+            v2 = self._expand_envvars_in_string(v)
+            os.environ[k] = v2
+            if v != v2:
+                print("[envvar] export {}={} (expanded)".format(k,v2))
         for k in self.actions["unsetenv"]:
             del os.environ[k]
             print("[envvar] unset {}".format(k))
@@ -288,6 +292,8 @@ class SetEnvironment(object):
         Raises:
             IndexError: Invalid number of parameters for 'module' command
         """
+        self.actions["module-op"]
+
         print("")
         print("Module Operations")
         print("+---------------+")
@@ -360,7 +366,12 @@ class SetEnvironment(object):
                          "module-list": {}     # Keys = modules already set to 'load'
                         }
 
+        print("Processing Configuration .ini")
+        print("+---------------------------+")
+
         self.actions = self._load_configuration_r(self.profile, actions=self.actions)
+
+        print("")
 
         return self.actions
 
@@ -403,15 +414,11 @@ class SetEnvironment(object):
         if processed_secs is None:
             processed_secs = {}
 
-        # Exit if we've already processed this section (recursion breaker)
-        if config_section in processed_secs.keys():
-            print("WARNING: Cyclic dependencies encountered when loading {} ... {}".format(self.profile, config_section))
-            return actions
-
+        print("Section: {}".format(config_section))
         processed_secs[config_section] = True
 
         for op2,value in sec.items():
-            #print("> k=`{}` v=`{}`".format(k,v))
+            #print(": op2=`{}` value=`{}`".format(op2,value))
             value  = str(value.strip('"'))
             oplist = op2.split()
             op1    = str(oplist[0])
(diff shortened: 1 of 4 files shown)


cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
```diff
@@ -403,15 +414,11 @@ def _load_configuration_r(self, config_section, actions=None, processed_secs=Non
...
-        if config_section in processed_secs.keys():
-            print("WARNING: Cyclic dependencies encountered when loading {} ... {}".format(self.profile, config_section))
-            return actions
-
+        print("Section: {}".format(config_section))
         processed_secs[config_section] = True
 
         for op2,value in sec.items():
-            #print("> k=`{}` v=`{}`".format(k,v))
+            #print(": op2=`{}` value=`{}`".format(op2,value))
```
jmgate, line 421: Do you want to remove this commented out line?

cmake/std/trilinosprhelpers/setenvironment/unittests/setenvironment_config.ini
```diff
@@ -152,3 +154,50 @@ module-unload gcc:
...
+
+#
+# Sections to test cycle breaking
+# - Test by trying to process Sec_D, we should get blocked
+#   when Sec_A tries to load Sec_C
+# - Final result should have envvar Sec_C == '99'
+[Sec_A]:
+setenv Sec_A: 1
+setenv Sec_C: 99
+use Sec_C:
```
jmgate, line 167: Do you want the environment variable names to be the same as the section names?  Might be a source of confusion in the future.

cmake/std/trilinosprhelpers/setenvironment/unittests/test_SetEnvironment.py
```diff
@@ -583,7 +583,110 @@ def test_SetEnvironment_module_returns_NoneType(self):
...
+        """
+        Test loading TEST_MODULE_FAIL
+        - Tests loading a nonexistent module
+        - Module load should fail but we don't throw an exception.
+        - apply() should return a nonzero value.
+        """
+        filename    = self._filename
+        profile     = "TEST_MODULE_FAIL"
+        module_fail = True
+        throw_on_fail = False
```
jmgate, line 648: Do you want the `=`s aligned?  PEP8 says "no", but I know how you love your whitespace 😁

cmake/std/trilinosprhelpers/setenvironment/unittests/test_SetEnvironment.py
```diff
@@ -583,7 +583,110 @@ def test_SetEnvironment_module_returns_NoneType(self):
...
+        """
+        Test loading TEST_MODULE_FAIL
+        - Tests loading a nonexistent module
+        - Enables `throw_on_error` which should cause an Exception to be
+          thrown when the module command fails.
+        """
+        filename    = self._filename
+        profile     = "TEST_MODULE_FAIL"
+        module_fail = True
+        throw_on_fail = True
```
jmgate, line 673: Same here.

jmgate (approved):

jmgate:
Canceling automerge in case you wanted to change anything given the comments above.  I'm fine with it either way.

cmake/std/trilinosprhelpers/setenvironment/unittests/setenvironment_config.ini
william76: it doesnt matter

cmake/std/trilinosprhelpers/setenvironment/unittests/test_SetEnvironment.py
william76: PEP8 can be stupid sometimes.

cmake/std/trilinosprhelpers/setenvironment/unittests/test_SetEnvironment.py
william76: same as earlier. Strict PEP8 is nice in many places, but it can also make things hard to read.

cmake/std/trilinosprhelpers/setenvironment/SetEnvironment.py
william76: no. It's useful for debugging stuff... easier to just uncomment when I'm tracking down bugs and want to trace what its' doing.

cmake/std/trilinosprhelpers/setenvironment/unittests/test_SetEnvironment.py
jmgate: Oh I agree.  I was also trying to point out that the `=` signs aren’t all vertically aligned, which I thought is what you wanted.

PR #8499, opened 2020-12-18
Tpetra:  Add missing syncs/modifies in BlockMultiVector unit tests

tasmith4:
@trilinos/tpetra 

## Motivation
These tests operate on primarily on host, but without explicit sync/modify calls, data added on host using non-const Views handed out from the BlockMultivector was being overwritten.  Also, in the Import test, without X being marked modify_host after modifications were made to a host view of X's data, the correct data was not brought over to Y.

## Related Issues

* Related to #8209

## Testing
Now passing in a non-UVM build on an x86/V100 platform (ascicgpu).

### DIFF
--- a/packages/tpetra/core/test/Block/BlockMultiVector.cpp
+++ b/packages/tpetra/core/test/Block/BlockMultiVector.cpp
@@ -429,16 +429,6 @@ namespace {
     TEST_ASSERT( X_overlap.data () != NULL );
     TEST_EQUALITY_CONST( static_cast<size_t> (X_overlap.extent (0)), static_cast<size_t> (blockSize) );
 
-    // {
-    //   std::ostringstream os;
-    //   os << "Proc " << myRank
-    //      << ": X_overlap.data() = " << X_overlap.data ()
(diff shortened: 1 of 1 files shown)


brian-kelley (approved):
LGTM.

One thing I'm curious about, I see you did sync_host on a newly constructed BMV on line 266. Is that necessary to avoid overwriting host modifications?

For the constructor which just takes a map and makes a new DualView, both d_view and h_view are intialized to zeros and so are both modified flags. On the other hand, if the modified_flags(1) == 0, then this sync is a no-op so maybe it doesn't matter.

tasmith4:
@brian-kelley MultiVector and friends are almost always marked modified on device (need sync to host) after construction, and I confirmed that is also the case in these tests.  Since we need to modify on host, we need an intervening sync_host to guarantee the initial values are copied to host memory (and to avoid a Kokkos error when we follow up with a modify_host after modifying host data).

In this case, the initial values might also be correct on host, but several of the MV constructors do save time by only initializing on device, so it's not guaranteed, and in any case, we should respect the modified on device flag for consistency.

brian-kelley:
@tasmith4 OK, got it, thanks for figuring all this out.

PR #8513, opened 2020-12-22
MueLu: fix rebalance error message and update documentation

mayrmt:
@trilinos/muelu 

## Motivation

- The error check about the size of the range/domain map extractors before and after rebalancing now prints the number of sub maps before and after rebalancing (instead of the pointer of the `rebRangeMapExtractor`/`rebDomainMapExtractor`).
- Documentation of input and output of `RebalanceBlockAcFactory` has been added.
- Rely on auto-generation of default constructor and destructor.

## Stakeholder Feedback

- Having the correct debug information in a error message helps to track down the error.
- Documentation of the factory on the factory's doxygen html site is useful when using this factory.

## Testing

All MueLu tests pass locally on my workstation.


mayrmt:
@trilinos/muelu This PR is awaiting review. Thanks!

cgcgcg (approved):

PR #8518, opened 2020-12-23
MueLu: Mixed precision RefMaxwell

cgcgcg:

## Motivation
- Ifpack2Smoother: Only call `SetPrecParameters` when they actually change
- RAPFactory: Fix repair threshold parameter when Scalar != double
- RefMaxwell: When requested, use average edge length for nullspace normalization
- Xpetra: Add an operator that wraps an operator with ScalarTraits<Scalar>::halfPrecision into a Scalar one.
- Allow running RefMaxwell in half precision through Stratimikos interface

Let's see what the AT thinks of this..


lucbv (approved):
I guess assuming that `Tpetra::deep_copy` does the right thing under the hood it should be fine?

jhux2:
@cgcgcg What's the story with all the negative lines of code in the Thyra interface code?

cgcgcg:
@jhux2 I took the opportunity to refactor the Thyra interface, since I was already putting in the cast from double to float. No functional changes on that part, but less messy and more maintainable code

PR #8530, opened 2021-01-04
Sacado: Enable dynamic linking of host code using Sacado in Cuda buil…

vbrunini:
…ds with RDC.

Some Sierra tests that include user plugins that are dynamically linked
previously failed to compile in Cuda builds, reporting that the
global_sacado_cuda_memory_pool_on_device symbol was undefined. The
user plugin code is host only, and we don't enable the Sacado memory
pool in Sierra anyway. Checking whether or not the memory pool is
enabled in one additional ifdef resolves the undefined symbol for
the shared library creation.

@trilinos/Sacado
@etphipp 

## Testing
Tested in Sierra clang & Cuda builds.

### DIFF
--- a/packages/sacado/src/Sacado_DynamicArrayTraits.cpp
+++ b/packages/sacado/src/Sacado_DynamicArrayTraits.cpp
@@ -37,7 +37,7 @@ namespace Sacado {
 }
 #endif
 
-#if defined(HAVE_SACADO_KOKKOSCORE) && defined(SACADO_KOKKOS_USE_MEMORY_POOL) && !defined(SACADO_DISABLE_CUDA_IN_KOKKOS) && defined(KOKKOS_ENABLE_CUDA) && defined(__CUDACC__)
+#if defined(HAVE_SACADO_KOKKOSCORE) && !defined(SACADO_DISABLE_CUDA_IN_KOKKOS) && defined(KOKKOS_ENABLE_CUDA) && defined(__CUDACC__)
 namespace Sacado {
   namespace Impl {
     const Kokkos::MemoryPool<Kokkos::Cuda>* global_sacado_cuda_memory_pool_host = 0;
--- a/packages/sacado/src/Sacado_DynamicArrayTraits.cpp
+++ b/packages/sacado/src/Sacado_DynamicArrayTraits.cpp
@@ -37,7 +37,7 @@ namespace Sacado {
 }
 #endif
 
-#if defined(HAVE_SACADO_KOKKOSCORE) && !defined(SACADO_DISABLE_CUDA_IN_KOKKOS) && defined(KOKKOS_ENABLE_CUDA) && defined(__CUDACC__)
+#if defined(HAVE_SACADO_KOKKOSCORE) && defined(SACADO_KOKKOS_USE_MEMORY_POOL) && !defined(SACADO_DISABLE_CUDA_IN_KOKKOS) && defined(KOKKOS_ENABLE_CUDA) && defined(__CUDACC__)
 namespace Sacado {
   namespace Impl {
     const Kokkos::MemoryPool<Kokkos::Cuda>* global_sacado_cuda_memory_pool_host = 0;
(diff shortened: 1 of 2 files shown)


etphipp:
This seems reasonable.  Wouldn't you also need to make a corresponding change in Sacado_DynamicArrayTraits.cpp:
```
#if defined(HAVE_SACADO_KOKKOSCORE) && !defined(SACADO_DISABLE_CUDA_IN_KOKKOS) && defined(KOKKOS_ENABLE_CUDA) && defined(__CUDACC__)
namespace Sacado {
  namespace Impl {
    const Kokkos::MemoryPool<Kokkos::Cuda>* global_sacado_cuda_memory_pool_host = 0;
    const Kokkos::MemoryPool<Kokkos::Cuda>* global_sacado_cuda_memory_pool_device = 0;
#ifdef KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE
    __device__ const Kokkos::MemoryPool<Kokkos::Cuda>* global_sacado_cuda_memory_pool_on_device = 0;
#endif
  }
}
#endif
```

vbrunini:
Without that corresponding change it does build successfully and I think you just end up with unused symbols in the library, but yeah doing that makes sense. I'll add it.

etphipp (approved):
Looks good to me.

vbrunini:
Is the pull request testing working again? Is there something we need to do to make it re-run?

etphipp:
> Is the pull request testing working again? Is there something we need to do to make it re-run?

Just add the AT:RETEST label, which I just did (and also added AT:AUTOMERGE to merge it).

etphipp:
So the test failures reminded me again how this was setup.  Basically the memory pool symbol was always created and the `SACADO_KOKKOS_USE_MEMORY_POOL` #define just controlled whether it was used in Sacado's memory management routines.  It was done this way to make it easy for the user to turn it on/off in different translation units.  To fix these errors I guess we would need to wrap any use of the memory pool with that #define.  

That begs the question though as to why this change was needed in the first place.  In the past we have verified that Sacado builds and links correctly with RDC enabled and resolved some specific issues with these same symbols (although I don't recall if there is any regular testing of Trilinos with RDC).  Does that no longer work?  Is Sierra perhaps not linking against the Sacado library?

vbrunini:
I wish it was that easy :) Here's what causes the issue:
1) Build Trilinos static libraries for Cuda with RDC enabled.
2) Build Sierra static libraries & executables for Cuda with RDC enabled and link to Trilinos.
3) Build a shared library user plugin that links against Sierra & Trilinos. We build this with nvcc_wrapper, but it's all host-only code in the user plugin.

Steps 1&2 work without issue, Step 3 reports the undefined symbol errors for the memory pool on the device.

etphipp:
OK, that seems like it is some kind of bug in the compiler toolchain.  But regardless, I am sure we can work around this.  How about we add a configure option that enables/disables the memory pool and would guard this code, but then leave `SACADO_KOKKOS_USE_MEMORY_POOL` to determine whether it is actually used (if it is enabled)?  Then Sierra could just disable it.  If Sierra actually wanted to use the memory pool, I'm not sure what you could do.  And I guess to not impact others already using the memory pool, it would be enabled by default.  How does that sound?  @rppawlo, do you have any thoughts?

vbrunini:
That works for me (that's actually what I thought SACADO_KOKKOS_USE_MEMORY_POOL was for and I was trying to do in this PR). I'm still a little confused by the build errors on the PR since I don't see where the pool is referenced outside of something protected by SACADO_KOKKOS_USE_MEMORY_POOL now, but maybe I'm just failing at grep.

etphipp:
I think what is happening is the library is compiled with SACADO_KOKKOS_USE_MEMORY_POOL not defined, so the symbol isn't created, but then the memory pool is then used in a few tests since SACADO_KOKKOS_USE_MEMORY_POOL is defined in those tests.  Now I'm thinking the best thing is to not add the guard in Sacado_DynamicArrayTraits.cpp like I asked you to do before.  I believe that would fix this issue (the symbol is always created, just not necessarily used).  Would you mind just removing that second commit and I think it will fix this?

etphipp (approved):

Issue #8537, opened 2021-01-05
Intrepid2_unit-test_Shared_RealSpaceTools_Cuda_Test_01_MPI_1 failing in ATDM Trilinos CUDA builds starting before 2020-11-05


bartlettroscoe:
CC: @trilinos/intrepid2, @mperego (Trilinos Descretizations Product Lead), @e10harvey

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label "impacting: tests"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

[...]

PR #8555, opened 2021-01-08
Intrepid2: fix the atdm failure

kyungjoo-kim:
@trilinos/<teamName>


kyungjoo-kim:
This mixes the partial solutions and now I realize why I stopped working on the uvm removal that time. I create a separate PR for just fixing #8537

Issue #8554, opened 2021-01-08
Intrepid2: Warnings....


kyungjoo-kim:
## Enhancement
@trilinos/intrepid2

@mperego, I think that you use celltools in a wrong way. A device level code is included in the following projection tools. It also causes host device warnings. This needs to be fixed.

```
/ascldap/users/kyukim/Work/lib/trilinos/kyukim/packages/intrepid2/src/Projection/Intrepid2_ProjectionToolsDefL2.hpp(389): warning: calling a __host__ function("Intrepid2::CellTools< ::Kokkos::Cuda> ::mapToReferenceSubcell<double,  ::Kokkos::LayoutStride,  ::Kokkos::Device< ::Kokkos::Cuda,  ::Kokkos::CudaUVMSpace> ,  ::Kokkos::MemoryTraits<(unsigned int)0u>  , double,  ::Kokkos::LayoutStride,  ::Kokkos::Device< ::Kokkos::Cuda,  ::Kokkos::CudaUVMSpace> ,  ::Kokkos::MemoryTraits<(unsigned int)0u>  > ") from a __host__ __device__ function(" const") is not allowed

/ascldap/users/kyukim/Work/lib/trilinos/kyukim/packages/intrepid2/src/Projection/Intrepid2_ProjectionToolsDefL2.hpp(409): warning: calling a __host__ function("Intrepid2::CellTools< ::Kokkos::Cuda> ::mapToReferenceSubcell<double,  ::Kokkos::LayoutStride,  ::Kokkos::Device< ::Kokkos::Cuda,  ::Kokkos::CudaUVMSpace> ,  ::Kokkos::MemoryTraits<(unsigned int)0u>  , double,  ::Kokkos::LayoutStride,  ::Kokkos::Device< ::Kokkos::Cuda,  ::Kokkos::CudaUVMSpace> ,  ::Kokkos::MemoryTraits<(unsigned int)0u>  > ") from a __host__ __device__ function(" const") is not allowed
```

@CamelliaDPG There is warning in your code.

Currently the variable is used for compiling on host but it is not used with cuda. I am not sure how this code can pass the PR test but I see the following warnings. Would you please fix the warnings ? I compile the code with "Intrepid2_ENABLE_DEBUG"
```
/ascldap/users/kyukim/Work/lib/trilinos/kyukim/packages/intrepid2/src/Shared/Intrepid2_TensorViewIterator.hpp(140): warning: variable "contracting" was set but never used
```

[...]

PR #8558, opened 2021-01-09
Intrepid2: fix issue with L2 projection to run w/out CUDA UVM

mperego:
@trilinos/intrepid2 

## Motivation

Part of #8554.


CamelliaDPG (approved):
@mperego, looks good!

PR #8542, opened 2021-01-06
8541 Panzer: Lazy evaluation of intrepid components

seamill:
@trilinos/panzer

PR #8568, opened 2021-01-12
Layout swap

bathmatt:
You can now compile with layout right left or what is in the code now.  Swapped most of the Kokkos::view to PHX::view, 

going to test on cuda, does the MR tester test cuda?


packages/shylu/shylu_node/tacho/src/impl/Tacho_Solver_Impl.hpp
```diff
@@ -13,7 +13,8 @@ namespace Tacho {
   template<typename VT, typename ST>
   Solver<VT,ST>
   ::Solver()
-    : _m(0), _nnz(0),
+    : _transpose(0), _mode(0), _order_connected_graph_separately(0),
+      _m(0), _nnz(0),
       _ap(), _h_ap(), _aj(), _h_aj(),
```
rppawlo, line 18: @bathmatt - there is a shylu/tacho file change mixed into the panzer changes. Did you intend to commit this? If so maybe @kyungjoo-kim can review these two lines? They look fine to me.

rppawlo (approved):
@bathmatt - looks good to me. I think this could result in non-backwards compatible changes since layouts are changed and might break assignment operators. I'd like to test against Charon and Drekar to be safe before merging. Did you have to make any changes to empire for these changes? Also - @seamill has a very large change to worksets that I am reviewing today (https://github.com/trilinos/Trilinos/pull/8542). That commit might conflict or might introduce new raw kokkos views. Probably need to test that as well.

packages/shylu/shylu_node/tacho/src/impl/Tacho_Solver_Impl.hpp
kyungjoo-kim: This is not matt's modification I believe. Those changes seem to be a side effect from merging my PR to develop. Merging this as it is okay to me.

PR #8570, opened 2021-01-12
Intrepid2: resolve warnings under CUDA.

CamelliaDPG:
Intrepid2: resolve warnings under CUDA.

As part of the warning resolution, fixed some issues with the safety of using CellGeometry objects on device under CUDA.

@trilinos/intrepid2 

## Motivation
Some warnings issued during CUDA debug builds (see #8554) indicated that lambda captures, etc. of CellGeometry as implemented could lead to indeterminate behavior on device, due to host code being called during construction and destruction of the device copy of the objects.  This PR resolves that by isolating the host-only "members" of CellGeometry in a separate, static class.

## Testing
Tested by the CellGeometry group of unit tests in Intrepid2, among other unit tests.


mperego (approved):
It looks good to me. Thanks for fixing this quickly.

PR #8573, opened 2021-01-13
Add single-reduce MGS & CGS2 options into GmresSingleReduce

iyamazaki:
@trilinos/belos 

## Motivation
This PR adds the single-reduce MGS & CGS2 options into GmresSingleReduce under Belos "Native Tpetra" solvers

## Testing
Added new options into ctest under belos/tpetra/Native, and ran ctest on blake

### DIFF
--- a/packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSingleReduce.hpp
+++ b/packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSingleReduce.hpp
@@ -79,6 +79,9 @@ private:
                                    dense_matrix_type& H,
                                    dense_matrix_type& WORK) const
   {
+    Teuchos::RCP< Teuchos::Time > orthTimer = Teuchos::TimeMonitor::getNewCounter ("GmresSingleReduce::LowSynch::Ortho");
+    Teuchos::TimeMonitor OrthTimer (*orthTimer);
+
     int rank = 0;
     if (input.orthoType == "CGS") {
       // default, one-synch CGS, optionally with reortho
@@ -105,8 +108,8 @@ private:
     const mag_type eps  = STS::eps ();
     const mag_type tolOrtho = static_cast<mag_type> (10.0) * eps;
 
-    Teuchos::RCP< Teuchos::Time > dotsTimer = Teuchos::TimeMonitor::getNewCounter ("MixedGmres::LowSynch::dot-prod");
-    Teuchos::RCP< Teuchos::Time > projTimer = Teuchos::TimeMonitor::getNewCounter ("MixedGmres::LowSynch::project");
+    Teuchos::RCP< Teuchos::Time > dotsTimer = Teuchos::TimeMonitor::getNewCounter ("GmresSingleReduce::LowSynch::dot-prod");
+    Teuchos::RCP< Teuchos::Time > projTimer = Teuchos::TimeMonitor::getNewCounter ("GmresSingleReduce::LowSynch::project");
 
     int rank = 1;
     // ----------------------------------------------------------
@@ -300,6 +303,7 @@ private:
     return rank;
   }
 
+
   SolverOutput<SC>
   solveOneVec (Teuchos::FancyOStream* outPtr,
                vec_type& X, // in X/out X
@@ -313,6 +317,10 @@ private:
     const SC zero = STS::zero ();
     const SC one  = STS::one ();
     const bool computeRitzValues = input.computeRitzValues;
+
+
+    // timers
+    Teuchos::RCP< Teuchos::Time > spmvTimer = Teuchos::TimeMonitor::getNewCounter ("GmresSingleReduce::Sparse Mat-Vec");
 
     SolverOutput<SC> output {};
     // initialize output parameters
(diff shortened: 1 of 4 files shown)


mhoemmen:
@egboman Given that I'm no longer a Trilinos developer, could we talk a bit offline about the context?  I'll send you an e-mail from my (new) work account (hint: it has all the letters of my last name in it).

mhoemmen:
@egboman Did you consider including iterative solvers experts @jennloe or @hkthorn , presumably still your colleagues, in this conversation?

jhux2 (dismissed):
@iyamazaki As far as I can tell, this seems like a good addition, but it would be better to get @hkthorn or @jennloe to approve before merging.  In the test, I couldn't tell if there's a check on the number of iterations.  Would it make sense to compare to standard GMRES?  (Maybe that's already happening and I don't know the code well enough.)

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSingleReduce.hpp
```diff
@@ -324,7 +723,7 @@ class GmresSingleReduce : public Gmres<SC, MV, OP> {
       else if (output.numIters < input.maxNumIters) {
         // Restart, only if max inner-iteration was reached.
         // Otherwise continue the inner-iteration.
-        if (iter >= restart) {
+        if (iter >= restart || H(iter,iter-1) == zero) { // done with restart cycyle, or probably lost ortho
```
jennloe, line 726: Should there be a tolerance here?  How likely is it that H(i,i-1) = 0 exactly?  Also, should it be H(iter, iter+1)?? 
If we restart early, it might also be nice to give a warning to the user.

packages/belos/tpetra/test/Native/CMakeLists.txt
```diff
@@ -92,7 +92,31 @@ TRIBITS_ADD_TEST(
 TRIBITS_ADD_TEST(
   Tpetra_Native_solvers_nonsymm_matrix
   NAME Tpetra_Native_GMRES_SINGLE_REDUCE_nonsymm_matrix
-  ARGS "--solver=\"TPETRA GMRES SINGLE REDUCE\" --restartLength=10"
+  ARGS "--solver=\"TPETRA GMRES SINGLE REDUCE\" --restartLength=10 --noRitzValues"
```
jennloe, line 95: Does there exist a test in this file that does not have the --noRitzValues option?  Perhaps we should still test that case?

packages/belos/tpetra/test/Native/nonsymm_matrix.cpp
```diff
@@ -232,12 +235,15 @@ testSolver (Teuchos::FancyOStream& out,
   params->set ("Verbosity", verbose ? 1 : 0);
   params->set ("Maximum Iterations", commandLineOptions.maxNumIters);
   params->set ("Num Blocks", commandLineOptions.restartLength);
+  params->set ("Orthogonalization", commandLineOptions.orthoType);
   if (solverName == "TPETRA GMRES S-STEP") {
     params->set ("Step Size", commandLineOptions.stepSize);
-    params->set ("Compute Ritz Values", commandLineOptions.computeRitzValues);
```
jennloe, line 237: Why did we remove this option?

jennloe (commented):
I have skimmed through this and left a few comments.  Still might be a good idea to get @hkthorn or even @cgcgcg to look at this; I am not very familiar with the internal working of Belos orthogonalization routines.

packages/belos/tpetra/test/Native/nonsymm_matrix.cpp
iyamazaki: I don't think we removed it. We just moved it to [here](https://github.com/trilinos/Trilinos/blob/c27e31e8397f5bdf4d7861be118f01e23f522faf/packages/belos/tpetra/test/Native/nonsymm_matrix.cpp#L245).

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSingleReduce.hpp
iyamazaki: I believe we check the tolerance inside the orthogonalization kernel, and set H(i,i-1) to be zero if failed (e.g., [here](https://github.com/trilinos/Trilinos/blob/c27e31e8397f5bdf4d7861be118f01e23f522faf/packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSingleReduce.hpp#L426) But, we probably should take the tolerance as an input parameter..

iyamazaki:
I think we only check if it has converged within maxAllowedNumIters, but I can look into comparing with the standard Gmres. Thank you, @jhux2 and @jennloe !!

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSingleReduce.hpp
jennloe: Okay.  That sounds reasonable.  I still think we should warn the user (print a message) if orthog fails and we restart early.

iyamazaki: Thank you, @jennloe . I added a couple of print warnings for the non-positive scaling factors!!

packages/belos/tpetra/test/Native/CMakeLists.txt
jennloe: @iyamazaki What about this one?  Do we need to add a test without the --noRitzValues flag?

packages/belos/tpetra/src/solvers/Belos_Tpetra_GmresSingleReduce.hpp
jennloe: Looks good!

packages/belos/tpetra/test/Native/CMakeLists.txt
iyamazaki: We don't usually need Newton basis for single-reduce GMRES, but you are right @jennloe  we should test that option. I added it to the test.

jennloe (approved):

PR #8576, opened 2021-01-14
Zoltan2: Add quotient algorithm and communication graph model

seheracer:
<!---
Be sure to select `develop` as the `base` branch against which to create this
pull request.  Only pull requests against `develop` will undergo Trilinos'
automated testing.  Pull requests against `master` will be ignored.

Provide a general summary of your changes in the Title above.  If this pull
request pertains to a particular package in Trilinos, it's worthwhile to start
the title with "PackageName:  ".

Note that anything between these delimiters is a comment that will not appear
in the pull request description once created. Most areas in this message are

Please make sure to mark:
* Reviewers
* Assignees
* Labels

-->
@trilinos/zoltan2
## Motivation
<!--- 
Why is this change required?  What problem does it solve? Please link to a github 
issue that describes the problem/issue/bug this PR solves.
-->
This PR adds two major contributions in the following files: 

1) model/Zoltan2_CommGraphModel.hpp
CommGraphModel (communication graph model) creates a graph representing the communication topology of the MPI ranks for a given XpetraCrsGraphAdapter object. If there are n MPI ranks in the given communicator, then this model contains n vertices so that each vertex represents an MPI rank. If rank i sends a message to rank j (during the mat-vec on the matrix corresponding to the given adapter), then there is a directed edge from vertex i to vertex j in the graph. The weight of the edge is the number of nonzeros that cause that message. The weight of vertex i is the number of nonzeros                                                                                                                                                                                                       
currently residing at rank i. Since the communication graph is too small, we migrate it into a subset of ranks,  which we call activeRanks. nActiveRanks_ denotes the number of active ranks and is computed as n/threshold_. For now, this migration is mandatory but we can make it optional (by setting a parameter and migrated_ flag). The threshold_ value can also                                                                                                                                                                                                      
be parameterized.  

2) algorithm/partition/Zoltan2_AlgQuotient.hpp:
This algorithm partitions the CommGraphModel and transfers the solution (which is only stored on active ranks of CommGraphModel) to all MPI ranks. For now, it uses ParMETIS to partition the graph. Support for other partitioners can be added if needed.



<!---
If applicable, let us know how this merge request is related to any other open
issues or pull requests:

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 
-->


## Stakeholder Feedback
<!--- 
If a github issue includes feedback from the relevant stakeholder(s), please link it.  
If the stakeholder(s) communicated that feedback through a different medium, please note that you did so.
-->

## Testing
<!---
Please confirm that any classes or functions in the Trilinos library that this PR touches are 
exercised by at least one test in Trilinos.  Please specify which test that is.  For untestable 
changes (e.g. changes to the nightly testing system) or changes to Trilinos tests, please say "N/A".

-->
All tests are passing on CUDA builds on Summit (with and without ParMETIS enabled). 

<!--- 
## Additional Information
 -->

### DIFF
--- a/packages/muelu/src/Rebalancing/MueLu_Zoltan2Interface_def.hpp
+++ b/packages/muelu/src/Rebalancing/MueLu_Zoltan2Interface_def.hpp
@@ -256,14 +256,25 @@ namespace MueLu {
 
       const typename InputAdapterType::part_t * parts = problem->getSolution().getPartListView();
 
-      // For blkSize > 1, ignore solution for every row but the first ones in a block.
-      for (GO i = 0; i < numElements/blkSize; i++) {
-        int partNum = parts[i*blkSize];
-
-        for (LO j = 0; j < blkSize; j++)
-          decompEntries[i*blkSize + j] = partNum;
+      if(algo == "quotient") {
+	// Quotient algorithm outputs a partition vector of size #MPIranks,
+	// where each rank gets a single entry locally, i.e., parts[0]. This entry 
+	// denotes which MPI rank in the coarse system that the current MPI rank
+	// should migrate to as a whole.
+	int partNum = parts[0];
+	for (GO i = 0; i < numElements; i++) {
+	  decompEntries[i] = partNum;
+	}
       }
+      else {
+	// For blkSize > 1, ignore solution for every row but the first ones in a block.
+	for (GO i = 0; i < numElements/blkSize; i++) {
+	  int partNum = parts[i*blkSize];
 
+	  for (LO j = 0; j < blkSize; j++)
+	    decompEntries[i*blkSize + j] = partNum;
+	}
+      }
       Set(level, "Partition", decomposition);
     }
   }
--- a/packages/zoltan2/core/src/algorithms/partition/Zoltan2_AlgParMETIS.hpp
+++ b/packages/zoltan2/core/src/algorithms/partition/Zoltan2_AlgParMETIS.hpp
@@ -61,13 +61,13 @@
 // but Zoltan2 not built with ParMETIS.
 
 namespace Zoltan2 {
-template <typename Adapter>
+template <typename Adapter, typename Model=GraphModel<typename Adapter::base_adapter_t>>
 class AlgParMETIS : public Algorithm<Adapter>
 {
 public:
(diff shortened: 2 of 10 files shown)


cgcgcg (approved):
MueLu changes look good to me

egboman (approved):
The Zoltan2 code looks fine. (I did not check MueLu.)
As Seher mentioned, any graph partitioner can be used to partition the quotient graph. There is really no need to require ParMetis. A future extension is to allow PHG.

kddevin (commented):
Interface changes requested:
- Return partListView of quotient algorithm does not match that of other 
  algorithms.

- Currently returning:

  quotient:  one part assignment per MPI rank
    on each processor, Teuchos::ArrayView(1) with part(0) = 
    single value in [0,nparts)

  everything else:  one part assignment per matrix row
    Teuchos::ArrayView(nrows) with part(i) = some value in [0,nparts)

- We should return:

  quotient:  one part assignment per matrix row; all rows assigned to same
             part (until we implement chunking version)
    on each processor, Teuchos::ArrayView(nrows) with part(i) = 
    const A in [0,nparts) for every row i (same A for every row)

  everything else:  one part assignment per matrix row
    Teuchos::ArrayView(nrows) with part(i) = some value in [0,nparts)

- For quotient method, modifying partListView to return part assignment per 
  matrix row allows
    common user interface for all methods;
    matrix redistribution in the partitioning1.cpp test will work for 
      better testing; and
    with quotient, can also check that part(i) = A for all i

Algorithm changes requested:

-  To accomplish change above, might need to 
   have quotient algorithm take Adapter as input (as in AlgPuLP), 
   construct its CommGraphModel itself from adapter, 
   and then use GIDs from adapter to setParts with result.

- Make threshold_ a parameter so that for testing, we can set it 2 for some of
  the PR tests (runs on 4, but migrates to 2)

- Remove untested option to not migrate.
  User can run on 4 and migrate to 4 if he doesn't want to migrate
  Test this case

- Remove code that disables distributeInput for quotient method.

Tests:

-  test both distributeInput=false, true 
   (--no-distribute, --distribute (default))

-  set threshold to run on 4, migrate to 2

-  set threshold to run on 4, migrate to 4

PR #8584, opened 2021-01-15
Piro: add HessianScaledThyraVector to ROL adapters and use it in Piro_PerformAnalysis

kliegeois:
@mperego
@trilinos/rol 
@bartlettroscoe 

## Motivation

This PR adds the capability to use a dot product which uses the Hessian w.r.t a parameter while using ROL with Thyra vectors.

The modifications included in this PR are:
- Thyra:
  - `packages/thyra/core/src/support/nonlinear/model_evaluator/client_support/Thyra_ModelEvaluatorDelegatorBase.hpp` has been modified to enable to call the member function `create_hess_g_pp` of the underlying model.
- ROL:
  - `packages/rol/adapters/thyra/src/vector/ROL_HessianScaledThyraVector.hpp` has been added. This header defines the Thyra vector which the dot product is scaled by the Hessian.
  - `packages/rol/adapters/thyra/src/function/ROL_ThyraProductME_Objective_SimOpt.hpp` has been updated to add the capability to evaluate the Hessian w.r.t a parameter.
- Piro:
  - `packages/piro/src/Piro_PerformAnalysis.cpp` has been updated to add the option to use the Hessian dot product (this option is disabled by default).
  - `packages/piro/test/*` the test `packages/piro/test/Main_AnalysisDriver_Tpetra.cpp` has been updated to test the new option. A second mock model has been added; this mock model, while using the Hessian dot product, converges in 1 iteration as theoretically expected.

Edit:
An extra option has been added to remove the mean of the right-hand side before applying the inverse of the Hessian.
This is useful to apply pseudo-inverse of the Hessian matrix for cases where its kernel is   where  is the vector which all entries are 1.

## Testing

The new capability is tested in `packages/piro/test/Main_AnalysisDriver_Tpetra.cpp` 
The previous and updated tests passed on Blake.

### DIFF
--- a/packages/piro/src/Piro_PerformAnalysis.cpp
+++ b/packages/piro/src/Piro_PerformAnalysis.cpp
@@ -488,6 +488,7 @@ Piro::PerformROLAnalysis(
   }
 
   bool useFullSpace = rolParams.get("Full Space",false);
+  bool useHessianDotProduct = rolParams.get("Hessian Dot Product",false);
 
   *out << "\nROL options:" << std::endl;
   rolParams.sublist("ROL Options").print(*out);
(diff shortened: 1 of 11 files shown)


bartlettroscoe (commented):
From the Thyra side, the addition of the function forwarding for `create_hess_g_pp()` is exactly what a delegator should do.

I would approve the PR but  ROL and PIRO developers should also review and sign-off as well.

kliegeois:
Thanks @bartlettroscoe for your review!

kliegeois:
@mperego I updated the PR with the `Remove Mean Of The Right-hand Side`; I will update the description of the PR.

mperego (approved):
Looks good! Thanks!

bartlettroscoe:
Related to:
* SEPW-213

PR #8594, opened 2021-01-16
Added a reset call

bathmatt:
Added a reset call

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_HierarchicParallelism.hpp
+++ b/packages/panzer/disc-fe/src/Panzer_HierarchicParallelism.hpp
@@ -68,6 +68,9 @@ namespace panzer {
 		       const int& vector_size,
 		       const int& fad_vector_size);
 
+    /// \brief Reset the sizes to default
+    void resetSizes() { use_auto_team_size_=true;}
+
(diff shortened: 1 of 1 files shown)


rppawlo (approved):

bathmatt:
Why can't I merge?  ANy ideas?  Roger can you merge @rppawlo

bathmatt:
@jwillenbring do you know why I can't merge??  Roger approved it...  can you use your magic powers and hit button to merge?

jwillenbring:
@bathmatt The autotester has not run again and verified the review from @rppawlo. I don't have the merge button available either. We have a backlog because of network problems and load issues, so it may take a little while for the autotester to verify the review.

bathmatt:
> @bathmatt The autotester has not run again and verified the review from @rppawlo. I don't have the merge button available either. We have a backlog because of network problems and load issues, so it may take a little while for the autotester to verify the review.

OK, i added a retest, not sure why since nothing has changed it needs a retest.  Does @rppawlo need to reapprove since it is the non-check-mark-one

jwillenbring:
@bathmatt I took the retest flag off. It doesn't need to retest, it just needs to see the review on the next run. I left the auto merge flags on however, as after the review is verified, it will auto merge.

PR #8521, opened 2020-12-23
Framework: Preparation for CMake minimum required version update

william76:
@trilinos/framework
This is a Work in Progress.

## Motivation
This is part of Issue #8520 to prepare the way for upgrading the minimum CMake required version to 17.0.0
* Part of #8520

## Testing
The PR updates will be tested through the usual PR driver.

The updates to the manual replication scripts will need to be tested manually.

This is due by January 29th, 2021.

### Update and Test PR Replication Scripts
These will need to be manually tested on an appropriate system.  See [Reproducing PR Testing Errors][1] for detailed instructions.
- [x] PullRequestClang10.0.0TestingEnv.sh
- ~~PullRequestClang7.0.1TestingEnv.sh~~
- ~~PullRequestClang9.0.0TestingEnv.sh~~
- [x] PullRequestCuda10.1.105TestingEnv.sh
- [x] PullRequestCuda9.2TestingEnv.sh
-  ~~PullRequestGCC4.8.4TestingEnv.sh~~
- ~~PullRequestGCC4.9.3TestingEnv.sh~~
- ~~PullRequestGCC4.9.3TestingEnvSERIAL.sh~~
- ~~PullRequestGCC7.2.0SerialTestingEnv.sh~~
- [x] PullRequestGCC7.2.0TestingEnv.sh
- [x] PullRequestGCC8.3.0TestingEnv.sh
- [x] PullRequestIntel17.0.1TestingEnv.sh
- [x] PullRequestIntel19.0.5TestingEnv.sh
- ~~PullRequestPython3TestingEnv.sh~~

Update: marked out .sh files that we don't use anymore for PR's.

[1]: https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors

Issue #8520, opened 2020-12-23
Framework:  Prepare PR Jobs for CMake minimum version update


william76:
@trilinos/framework

Description
-----------
As part of #8401, we need to update all the PR test scripts to load CMake >= 3.17.0 in preparation for the minimum-required version switchover that's coming on 2021-01-30.

- This issue is part of #8401

Tasks
-----
We should update the version of CMake that is used by our PR and Test replication jobs to set to the minimum version of CMake over 17.0
See the list in #8401 for the master list.

| Platform(s)           | Available Module                          |
|-----------------------|-------------------------------------------|
| RHEL7 COE + SEMS      | sems-cmake/3.17.1                         |
| Ride/White            | devpacks load 3.12, where is a newer one?<BR> `/home/atdm-devops-admin/tools/ride/cmake-3.17.2/bin/cmake` |
| RHEL7 COE Cloud Nodes | sems-cmake/3.17.1                         |
| Vortex                | cmake/3.18.0                              |

### Update Jenkins Pull Request Jobs
This should be in the `cmake/std/pr_config/pullrequest.ini`
- [x] `Trilinos_pullrequest_clang_10.0.0`
- [x] `Trilinos_pullrequest_clang_7.0.1`
- [x] `Trilinos_pullrequest_clang_9.0.0`
- [x] `Trilinos_pullrequest_cuda_9.2`  _(cmake loaded via devpack)_
- [x] `Trilinos_pullrequest_cuda_10.1.105` _(cmake loaded via devpacks)_
-  ~~`Trilinos_pullrequest_cuda_10.1.243`~~ _(not in pullrequest.ini)_
- [x] `Trilinos_pullrequest_gcc_4.8.4`
- [x] `Trilinos_pullrequest_gcc_4.9.3_SERIAL`
- [x] `Trilinos_pullrequest_gcc_7.2.0`
- [x] `Trilinos_pullrequest_gcc_7.2.0_debug`
- [x] `Trilinos_pullrequest_gcc_7.2.0_serial`
- [x] `Trilinos_pullrequest_gcc_8.3.0`
- [x] `Trilinos_pullrequest_intel_17.0.1`
- [x] `Trilinos_pullrequest_intel_19.0.5`
- [x] `Trilinos_pullrequest_python_3`

### Update PR Replication Scripts
These are in `cmake/std/sems/`
- [x] PullRequestClang10.0.0TestingEnv.sh
[...]

Issue #8401, opened 2020-12-01
Framework: Increase the minimum CMake version requirement


william76:
## Enhancement
There was a discussion today in the developers' meeting about updating the minimum-required version of CMake to something newer than 3.10.0.  I think the general consensus was that we'd like to update this to the most recent version of CMake that's supported on platforms we care about, noting that anything over 3.13.x is a pretty big step and 3.14.x or newer would be ideal.

## CMake Availability
Let's fill this chart in so we can get an idea of what platforms we care about:

| Platform | CMake Version(s) Available |
| -------- | ------------------------ |
| Vanguard 1 (Astra) | 3.12.2, 3.14.5, 3.17.1 |
| Chama | 3.12.2, 3.17.1, 3.18.1 |
| Mutrino | 3.12.3, 3.14.6, 3.16.2, 3.17.2, 3.18.1 |
| RHEL7 (OS default) | 2.8.12.2 |
| CEE RHEL7 | 3.12.2, 3.17.1, |
| ATS-1 (Trinity) |  3.14.2, 3.14.6, 3.16.2, 3.17.2 |
| ATS-2 (Sierra) | 3.12.1, 3.14.5, 3.16.8, 3.18.0 |
| Vortex | 3.12.1, 3.14.6, 3.16.2, 3.17.2 |
| CTS-1 | 3.12.2, 3.17.1, 3.18.1 |
| Ride/White | 3.12.3, 3.17.2 (in atdm-devops-admin?) 
| Weaver | 3.12.3, 3.18.0 |
| others? | ?? |

| Environment Modules | CMake Version(s) Available |
| -------- | ------------------------ |
| SEMS modules | 3.11.1, 3.12.2, 3.17.1 |
| SIERRA modules | 3.12.2, 3.12.3 |

This is related to #8387 and #8392 which I'm using to identify the PR builds that fail when using CMake 3.10.x.

ATTN: @trilinos/framework @ccober6 @crtrott @kddevin @bartlettroscoe 
_I just tagged folks who spoke up during the meeting today when we were speaking about this, feel free to unfollow this issue if you're not interested or tag others who I may have forgotten to @mention_

## Tasks

* [x] Determine options for version of CMake to upgrade minimum to.
* [x] Select the target for upgrading CMake version to and a timeline ... CMake 3.17.1 by end of Jan. 2021!
* [ ] Change the minimum version of CMake in Trilinos and TriBITS to CMake 3.17.1

[...]

PR #8602, opened 2021-01-18
Framework: Expand the envvars that get fully printed out in PR's

william76:
@trilinos/framework

## Motivation
The autotester driver does some filtering on whether or not it prints out the _value_ of an envvar in the logs because some envvars have a lot of content that's not relevant. Generally it prints out the envvar names, but not the value unless we include a partial string match to the envvar name in `TrilinosPRConfigurationBase.py`.

We missed adding envvars such as `CXX`, `CC`, `FC`, etc. to the list of envvars we should fully print out in the logs.

This PR addresses this by adding the flags `CC`, `CXX`, `F77`, `F90`, `FC`, `MODULESHOME` so that their values will be printed out to the console during a PR run in the future.

- Related to #8401
- Related to #8520
- Related to #8521

### DIFF
--- a/cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
+++ b/cmake/std/trilinosprhelpers/TrilinosPRConfigurationBase.py
@@ -587,7 +587,13 @@ class TrilinosPRConfigurationBase(object):
             "PROXY",
             "PATH",
             "OMP_",
-            "WORKSPACE"
+            "WORKSPACE",
+            "CC",
+            "CXX",
+            "F77",
+            "F90",
+            "FC",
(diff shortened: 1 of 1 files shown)


prwolfe (approved):
Looks good.

william76:
@prwolfe thanks!

It's still weird to me why that CMake 3.17.1 based Intel build is failing in CMake on the Fortran check.

I logged into ascic-158 and built a F90 hello-world using that compiler and the same modules as the job loads in and it works just fine.

Once this merges, I'll check retest on #8602 and see if it's still failing. Perhaps that error we saw last week / over the weekend was more anomalous?

PR #8606, opened 2021-01-19
ATDM:  Update CMake on Chama to 3.19.1

jmgate:
## Motivation
EMPIRE ran into a problem deploying on Chama, and the culprit looks like
it's using an older CMake version.  When the environment for Chama was
created, 3.12 was the best available, but now we can update to 3.19, so
we'll go ahead and do that.

## Testing
Testing now.

### DIFF
--- a/cmake/std/atdm/common/toss3/environment_tlcc2.sh
+++ b/cmake/std/atdm/common/toss3/environment_tlcc2.sh
@@ -24,7 +24,7 @@ module purge &> /dev/null
 module purge
 . /projects/sems/modulefiles/utils/sems-modules-init.sh
 module load sems-env
-module load cmake/3.12.2
+module load sems-cmake/3.19.1
 module load sems-ninja_fortran/1.8.2
 
 if [ "$ATDM_CONFIG_NODE_TYPE" == "OPENMP" ] ; then


e10harvey (commented):
Looks good to me. Thanks, @jmgate!

How is this being tested?

jmgate:
I'm building/installing my own Trilinos on chama now, and will build EMPIRE against it.  If the EMPIRE build gets past the configure stage, I think we're good to go.

That said, @e10harvey, do you know if this might negatively impact SPARC or GEMMA?  @sebrowne, I doubt it, but does this affect you guys?

eric-c-cyr (dismissed):
Assuming it doesn't break other codes, this is fine with me.

e10harvey:
> That said, @e10harvey, do you know if this might negatively impact SPARC or GEMMA?

I do not. On closer inspection, you actually want to update `cts1empire/environment_new.sh`.

e10harvey (changes requested):
Please revert the change to `environment_tlcc2.sh` and update `cts1empire/environment_new.sh` instead.

jmgate:
Whoops, sorry about that.  Update pushed.  If testing passes, I'll squash before merging.

jmgate:
And since we're only touching the `cts1empire` environment now, we don't need to worry about buy-in from others.

jmgate:
Wound up resetting to the initial commit.  @e10harvey, running on chama with `intel-opt-openmp-static` uses the `tlcc2` environment, so it looks like the initial change is what's needed.  I've made it past the configure stage on the EMPIRE side, so I think we're good to merge here.

e10harvey:
> Wound up resetting to the initial commit. @e10harvey, running on chama with `intel-opt-openmp-static` uses the `tlcc2` environment, so it looks like the initial change is what's needed. I've made it past the configure stage on the EMPIRE side, so I think we're good to merge here.

Sorry about that. Glad you got it sorted out.

e10harvey (approved):

PR #8608, opened 2021-01-19
Teuchos TimeMonitor: Print more info on overlapping timers

cgcgcg:
@trilinos/teuchos 

## Motivation
Print the exception message coming out of the stacked timers, since it says which two timers are overlapping. This makes debugging a bit easier.


rppawlo (approved):

bartlettroscoe (approved):
Looks good to me.  Just wish there was an automated test that triggered this exception and verified it showed what it should be showing.

PR #8610, opened 2021-01-20
Tpetra:  store permutation matrix (rather than its transpose) for LowerTriangularBlock distribution

kddevin:
## Motivation
@trilinos/tpetra 
The sorted LowerTriangularBlock distribution stores a column permutation matrix that is used to correctly apply the matrix in the LowerTriangularBlockOperator.  This PR corrects use of the permutation matrix in the LowerTriangularBlockOperator.  It also corrects the operation y = by + aAx in LowerTriangualBlockOperator; scalar a was previously handled incorrectly.

## Stakeholder Feedback

@ndellingwood 

## Testing

Passes the sorted LowerTriangularBlock tests in Tpetra/core/test/inout on mac.
These tests exercise the distribution and the operator.
Tests are enhanced to check that result of apply is correctly permuted for sorted LowerTriangularBlock.
Also, tests are enhanced to use apply y = by + aAx (rather than just y = Ax).

### DIFF
--- a/packages/tpetra/core/inout/Tpetra_DistributionLowerTriangularBlock.hpp
+++ b/packages/tpetra/core/inout/Tpetra_DistributionLowerTriangularBlock.hpp
@@ -427,7 +427,7 @@ private:
   // in decreasing sort order of number of nonzeros per row in full matrix
   bool sortByDegree;
 
-  // Permutation matrix built only when sortByDegree = true;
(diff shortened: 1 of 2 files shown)


seheracer (approved):
LGTM, thanks!

kddevin:
@seheracer I made the tests better and fixed another problem with the operator; would you please re-review?  Thanks.

seheracer (approved):

PR #8622, opened 2021-01-22
MueLu: Fix BGS and Blocked Jacobi logic when InitialGuessIsZero

pohm01:
@trilinos/muelu 

## Motivation
The logic with `InitialGuessIsZero` int BlockedGaussSeidel and BlockedJacobi smoothers resulted in the use of a solution from the previous V-cycle as an initial guess instead of an actual zero initial guess.

This also fixes a bug where BlockedJacobi would update the residual after each block, resulting in a Gauss-Seidel-like method.

## Related Issues
Appears to resolve convergence issues discussed in:
rppawlo/DrekarBase#505

### DIFF
--- a/packages/muelu/src/Smoothers/BlockedSmoothers/MueLu_BlockedGaussSeidelSmoother_def.hpp
+++ b/packages/muelu/src/Smoothers/BlockedSmoothers/MueLu_BlockedGaussSeidelSmoother_def.hpp
@@ -303,12 +303,9 @@ namespace MueLu {
     Scalar omega = pL.get<Scalar>("Damping factor");
 
 
-    // Clear solution from previos V cycle if still stored
-    Scalar normX = rcpX->getVector(0)->norm2();
-    if(InitialGuessIsZero==true && normX > 0){
+    // Clear solution from previos V cycles in case it is still stored
+    if( InitialGuessIsZero==true )
       rcpX->putScalar(Teuchos::ScalarTraits<Scalar>::zero());
-      // Is there a better way to do this?
-    }
 
 
     // outer Richardson loop
--- a/packages/muelu/src/Smoothers/BlockedSmoothers/MueLu_BlockedGaussSeidelSmoother_def.hpp
+++ b/packages/muelu/src/Smoothers/BlockedSmoothers/MueLu_BlockedGaussSeidelSmoother_def.hpp
@@ -302,6 +302,15 @@ namespace MueLu {
     LocalOrdinal nSweeps = pL.get<LocalOrdinal>("Sweeps");
     Scalar omega = pL.get<Scalar>("Damping factor");
 
+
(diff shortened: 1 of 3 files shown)


packages/muelu/src/Smoothers/BlockedSmoothers/MueLu_BlockedGaussSeidelSmoother_def.hpp
```diff
@@ -302,6 +302,15 @@ namespace MueLu {
     LocalOrdinal nSweeps = pL.get<LocalOrdinal>("Sweeps");
     Scalar omega = pL.get<Scalar>("Damping factor");
 
+
+    // Clear solution from previos V cycle if still stored
+    Scalar normX = rcpX->getVector(0)->norm2();
+    if(InitialGuessIsZero==true && normX > 0){
+      rcpX->putScalar(Teuchos::ScalarTraits<Scalar>::zero());
+      // Is there a better way to do this?
+    }
```
mayrmt, line 311: @pohm01 What about this?
```c++
    if (InitialGuessIsZero)
      rcpX->putScalar(Teuchos::ScalarTraits<Scalar>::zero());
```
Instead of computing the norm (which involves some local operations and some communication), why not just set all entries to zero. The norm seems not to be required for anything else.

Alternatively, you could compute the norm only in case of `InitialGuessIsZero == true):
```c++
    if (InitialGuessIsZero) {
      Scalar normX = rcpX->getVector(0)->norm2();
      if (normX > 0)
        rcpX->putScalar(Teuchos::ScalarTraits<Scalar>::zero());
    }
```

These are just ideas and suggestions. While I like the approach without ever computing the norm the most, of course I leave the decision up to you.

mayrmt (changes requested):
@pohm01 I posted some ideas about the implementation. 

Could you please cover this fix with a unit test? You can find unit existing tests for `BlockedJacobi` and `BlockedGaussSeidel` in `packages/muelu/test/unit_tests/Smoothers/BlockedSmooher.cpp`. If each smoother is tested once with and once without a zero initial guess, then changes of this PR should be covered by tests. Thanks!

packages/muelu/src/Smoothers/BlockedSmoothers/MueLu_BlockedGaussSeidelSmoother_def.hpp
pohm01: There actually were already tests for each smoother with and without a zero initial guess. I've added a third that does a zero initial guess without zeroing the solution vector `X` that is passed in.

pohm01:
@mayrmt There actually were already tests for each smoother with and without a zero initial guess. I've added a third that does a zero initial guess without zeroing the solution vector `X` that is passed in.

mayrmt (approved):
@pohm01 Thanks for adding a test!

PR #8624, opened 2021-01-22
.github/workflows: Automate Build Triage management

e10harvey:
Automate moving cards to the In Progress column of the Build Triage project when an issue is assigned to an entity.

### DIFF
--- a/.github/workflows/assigned-issues-in-progress.yml
+++ b/.github/workflows/assigned-issues-in-progress.yml
@@ -1,0 +1,15 @@
+name: Move assigned issues into In Progress
+
+on:
+  issues:
+    types: [assigned]
+
+jobs:
+  automate-project-columns:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bartlettroscoe (changes requested):
From a  Agile/Lean/Kanban perspective, I am not sure that an issue being "assigned" means it is "in progress".  From a Kanban perspective, an issue being "assigned" should be a pre-requisite for the issue to be moved to "in progress" but nothing more.  From an Kanban perspective, you want to minimize the items "in progress" to what is actually being worked and we want to keep the number of "in progress" minimized so that we focus on getting "in progress" items "done".

So from my perspective, I don't think this is a good change.  But you might argue that if an issue is "assigned" then at least it should be in the "Selected" column.  So if you made that change, I think I would approve this.  (But it is not my place to approve a change to how the Trilinos GitHub project behaves.)

Issue #8591, opened 2021-01-16
Tpetra: track MultiVector DualView refactoring


brian-kelley:
## Work Tracking Issue

@trilinos/tpetra 

Branch for working on MultiVector DualView management refactor: TpetraDualViewRefactor.

Updated 2-3-21:
I replaced the new functions we just added (e.g. getLocalViewDeviceConst, getLocalViewHostNonConst) with permament tagged versions:
```
//New interface:
#include "Tpetra_Access.hpp"
//...
MultiVector::getLocalView<Device>(Tpetra::Access::ReadOnly());
MultiVector::getLocalView<Device>(Tpetra::Access::ReadWrite());
MultiVector::getLocalView<Device>(Tpetra::Access::WriteOnly());
```
Non-templated versions like ``getLocalViewHost(tag)`` too.
All getLocalView calls in MultiVector and core/test/MultiVector have been replaced with this new interface, but this caused new failures in tests.

MultiVector
- [x] TpetraCore_MultiVector_UnitTests_MPI_4
- [x] TpetraCore_Transform_MPI_4
- [x] TpetraCore_Bug7758_MPI_4
- [x] TpetraCore_Bug7745_MPI_4
- [x] TpetraCore_MultiVector_get2dView_MPI_1
- [x] TpetraCore_MultiVector_TwoArgRandomize_MPI_1
- [x] TpetraCore_Issue364_MPI_1
- [x] TpetraCore_MultiVector_MicroBenchmark_MPI_1 //works on 4 ranks, doesn't work on 1
- [x] TpetraCore_deep_copy_SerialDenseMatrix_to_MultiVector_MPI_1
- [x] TpetraCore_deep_copy_MultiVector_to_SerialDenseMatrix_MPI_1
- [x] TpetraCore_MV_reduce_strided_MPI_4 //Brian
- [x] TpetraCore_aliased_deep_copy_MPI_4

FEMultiVector
- [ ] TpetraCore_FEMultiVector_UnitTests_MPI_4. // Can't fix until CrsMatrix is fixed
- [x] TpetraCore_FEMultiVector_Fix3101_MPI_4

BlockMultiVector
- [x] BlockMultiVector_MPI_4 
- [x] BlockMultiVector2_MPI_4
- [x] BlockMap_MPI_4.

Import/Export
- [x] Issue3968

brian-kelley:
List of tests in core/test/MultiVector that fail on this branch:

Unit tests:
[...]

PR #8631, opened 2021-01-23
tpetra:  removed use of device_type in BlockView test

kddevin:
@trilinos/tpetra 

## Motivation

The BlockView.cpp test exercises LAPACK functions provided by Tpetra (GETF2, GETRS, etc.).

It had used Kokkos::Views declared using the device_type, but in truth, the tests run on host.  The device_type declaration was misleading (fake news) and didn't work without UVM.

This PR changes the view declarations to use HostSpace memory.

This test probably *should* exercise the Tpetra LAPACK implementations on device, since the intent of the functions likely is to be applied with parallelism over many small blocks.  Such an extension would be a nice addition to this test, but it is not provided in this PR.   This PR checks only the correctness of the functions for a single block.

Fixed while working on #8591 

## Testing

CEE-LAN GPU machine, CUDA build, with and without UVM

### DIFF
--- a/packages/tpetra/core/test/Block/BlockView.cpp
+++ b/packages/tpetra/core/test/Block/BlockView.cpp
@@ -43,7 +43,6 @@
 
 #include "Tpetra_TestingUtilities.hpp"
 #include "Tpetra_BlockView.hpp"
-#include "Tpetra_Vector.hpp"
 #include "Teuchos_Array.hpp"
 #include "Teuchos_BLAS.hpp"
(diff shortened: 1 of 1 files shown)


kddevin:
@trilinos/framework  The following test for this PR is failing:  TrilinosFrameworkTests_PullRequestLinuxDriverTest_UnitTests
I don't know what this test does.  Is its failure a known problem?   Thanks!

GeoffDanielson (approved):
Looks good to me.

Issue #8397, opened 2020-12-01
Framework:  Create build to protect UVM off


jwillenbring:
## Enhancement
@trilinos/framework 

Once either #7047 or #8129 has been completed, we need to create another PR build that turns UVM off. We need to start with the subset that can be built cleanly with UVM off and add packages as they are cleaned up.

csiefer2:
I suspect you can start with Teuchos :)

rppawlo:
Phalanx is ready if you want to use that as a test case.

jwillenbring:
With the #8482 PR submitted which will satisfy #7047, I want to lay out a proposed plan for completing this issue. We can iterate on this as it makes sense. As we complete the steps in the plan we decide upon, I want to track the completion of each step in this ticket. Here is my proposed plan:

1) Make changes to turn UVM off to a branch and fire off a test build. Paste a link to the results in this ticket.

1) Once we have a build reflective of the overall state, set up a nightly specialized build on the dashboard with all packages (at least those relevant to the build) enabled. Leave this build running until the end of the effort to establish a fully clean launch blocking=off build, including having a PR build that protects that functionality.

1) Establish a new PR build for UVM OFF that enables only those packages working at 100%. As @rppawlo says above, we can start with just Phalanx, but before we can call this done, we need to get all additional packages that are clean at that time enabled. We can add to this build as additional packages are cleaned up. However, when this build is fully in place, with all currently clean packages enabled, this ticket as defined is done.

@trilinos/framework Let's discuss this plan of action.

ZUUL42:
I've setup [a branch](https://github.com/trilinos/Trilinos/compare/develop...ZUUL42:uvm) for the CUDA 10.1.105 build that turns on UVM for Kokkos & Tpetra.
Running it through the Jenkins job, I get this. [CDash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&parentid=8435101)
I haven't disabled any packages for this yet.

ZUUL42:
After turning off 4 packages, @trilinos/panzer , @trilinos/stokhos, @trilinos/intrepid2, & @trilinos/TrilinosCouplings, it now builds without error.
[...]

PR #8633, opened 2021-01-25
Create UVM = OFF build

ZUUL42:
@trilinos/framework 
CC: @kddevin 

## Motivation

This creates a UVM = OFF specific build using CUDA 10.1.105.

4 packages were turned off, panzer, stokhos, intrepid2, & trilinoscouplings in order to get it to build cleanly.
A significant number of tests had to be turned off in order to get this build to test cleanly. However, this will give us a place to start.

As package teams make modification to get tests to pass, they can be turned back on by removing the associated line in:
[`cmake\std\PullRequestLinuxCuda10.1.105uvmTestingSettings.cmake`](https://github.com/trilinos/Trilinos/blob/develop/cmake/std/PullRequestLinuxCuda10.1.105uvmTestingSettings.cmake)
Teams can do this, just request a review from any of the Framework team and the test will be reenabled for this build.

Associated Issue: #8397

## Testing

This has been tested manually with the new Cuda 10.1.105 uvm job. See #8397.
It shouldn't affect other builds.

The next step will be to add it to the dev->master build set.


csiefer2:
Is there a reason to disable the failing tests?  Why not just leave them on and have them fail?

jwillenbring:
> Is there a reason to disable the failing tests (since they number in the hundreds)? Why not just leave them on and have them fail?

We have been asked to set up a PR build that protects what is passing and add to this over time. There cannot be any failing tests in a PR build.

csiefer2:
> We have been asked to set up a PR build that protects what is passing and add to this over time. There cannot be any failing tests in a PR build.

Ah, it didn't say up top that it was for a PR build!  That makes much more sense!

As a warning:  People are not going to remember to re-enable tests when they get fixed.  Periodic reminders to package leads would be advised.

ZUUL42:
@csiefer2 I didn't note it explicitly. All PR builds as well as dev->master builds must build cleanly and pass all tests. Otherwise the PR build merge to develop and/or the merge to master won't happen. This will also prevent any backsliding of progress. As a dev->master build, Framework and anyone else watching the dev->master builds will be alerted when anything fails.
Then teams can turn on tests as they make modifications and I can periodically go in and run tests to see if additional tests are passing.
If we promote this to a PR build, it will prevent any changes that break a test to be submitted.

While I admit it is a _lot_ of tests disabled, it is a place to start and hold ground from here as packages advance Trilinos to be UVM free.

csiefer2:
@ZUUL42 I suspect what you'll actually have is packages updating nothing until all of their tests are passing.  Less work for the developer, but still useful.

cmake/std/sems/PullRequestCuda10.1.105uvmTestingEnv.sh
```diff
@@ -0,0 +1,20 @@
...
+module load git/2.10.1
+module load devpack/20190404/openmpi/4.0.1/gcc/7.2.0/cuda/10.1.105
+module swap openblas/0.3.4/gcc/7.4.0 netlib/3.8.0/gcc/7.2.0
+#export OMPI_CXX=`which g++`
+current_dir=`dirname $BASH_SOURCE`
+Trilinos_dir=`realpath $current_dir/../../..`
+export OMPI_CXX=${Trilinos_dir}/packages/kokkos/bin/nvcc_wrapper
+export OMPI_CC=`which gcc`
+export OMPI_FC=`which gfortran`
+export CUDA_LAUNCH_BLOCKING=1
```
jwillenbring, line 15: Why is launch blocking turned on?

cmake/std/trilinosprhelpers/unittests/trilinos_pr_test.ini
```diff
@@ -306,6 +307,21 @@ unsetenv PATH_TMP:
...
 
 
+[Trilinos_pullrequest_cuda_10.1.105_uvm]
+module-load git: 2.10.1
+module-load devpack: 20190404/openmpi/4.0.1/gcc/7.2.0/cuda/10.1.105
+module-swap openblas/0.3.4/gcc/7.4.0: netlib/3.8.0/gcc/7.2.0
+setenv OMPI_CXX: ${WORKSPACE}/Trilinos/packages/kokkos/bin/nvcc_wrapper
+setenv OMPI_CC: ${CC}
+setenv OMPI_FC: ${FC}
+setenv CUDA_LAUNCH_BLOCKING: 1
```
jwillenbring, line 317: Why is launch blocking on?

ZUUL42: I started with a copy of the current CUDA 10.1.105 build. We can turn it off.

csiefer2:
On second thought, using this as a PR test might not work with the Tpetra team's approach to UVM removal --- we're contemplating intentionally completely breaking downstream builds because we're going to change interfaces needed to support use w/o UVM.  This is not set in stone, but you should really talk about it with @kddevin.

jwillenbring:
@ZUUL42 I think what happened here is like just the fact that there are now conflicts between this branch and the current develop. If you resolve those, testing may resume as normal.

jwillenbring (approved):
This looks good.

PR #8637, opened 2021-01-25
Automatic snapshot commit from seacas at 729e904fa5

gsjaardema:
Origin repo remote tracking branch: 'origin/master'
Origin repo remote repo URL: 'origin = https://github.com/gsjaardema/seacas'

At commit:

commit 729e904fa5ce3905da73a9856c17f30f1be5ffab
Author:  Greg Sjaardema <gsjaardema@gmail.com>
Date:    Fri Jan 22 07:46:47 2021 -0700
Summary: Merge pull request #210 from ibaned/hdf5-targets

* EPU: Enum class instead of enum in many places
* EPU: Support for `szip` compression in addition to `zlib`
* EPU: Support for `netcdf5` output
* EPU: Assembly support
* EPU: Now uses `original_global_id_map` if present.
* EXODIFF: Better diagnostics in mapping failures
* EXODIFF: Edge block and face block support
* EXODIFF: Warning fixes
* NAS2EXO: New application
* NEMSLICE: integer size issues
* SLICE: Fix partial input/output issue
* APREPRO: Eliminate potential duplicat symbol issues in getline
* EXODUS: SZIP support if present in netCDF
* EXODUS: Assembly names can be read via ex_get_names
* EXODUS: fix some tests on netCDF symbols (if vs ifdef)
* IOSS: Avoid some symbol clashes
* IOSS: Fix parallel exodus append (needs netCDF patch though)
* IOSS: Szip support
* IOSS: enum class instead of enum several places
* IOSS: integer size issue in transform
* IOSS: Use hash classes instead of std::unorderd_map to speed up some areas
* IOSS: avoid symbol clash in getline
* IOSS: windows build fixes
* IOSS: add option to output a sideset with boundary faces to io_shell
* PYTHON: fixes for assemblies
* FMT: New version
* CATCH2: new version
* HASH: New versions
* Eliminate shadow warnings

@trilinos/<teamName>

PR #8643, opened 2021-01-26
Correct proxy pattern for pip

prwolfe:
was http://www... now needs http://user:nopass@www...

@trilinos/framework 

## Motivation
A change in the proxy setting from SNL require adding a bit to the pip fetch line


jwillenbring (approved):
Looks good. Thanks for investigating the issue.

prwolfe:
> Looks good. Thanks for investigating the issue.

Thanks to @e10harvey for pointing me in the right direction!

e10harvey (approved):
Thanks, @prwolfe !

jwillenbring (approved):
Sounds good as discussed at Stand-up.

cgcgcg:
@trilinos/framework Can this PR be prioritized? Cause all other PRs fail until this one goes in, right?

cgcgcg:
@prwolfe It appears that error is still there, but now it's simply ignored. Is this safe?

prwolfe:
Safe?

This error is not checking the code, it’s plumbing. We are still looking at the issue and will fix it as soon as we can, but holding up everyone while I get there is not a good anser. The regex I used will avoid test failures but not catch the PIP errors. So it’s a small risk for a short term.

Paul

prwolfe:
> @prwolfe It appears that error is still there, but now it's simply ignored. Is this safe?

I want to know what the intel 17 build failure is about ?

jhux2:
@gsjaardema might know?  The file that fails seems to change.


https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=PR-&field2=buildname&compare2=63&value2=test-Trilinos_pullrequest_intel_17.0.1&field3=buildstarttime&compare3=83&value3=NOW-14days

cgcgcg:
In particular, is it possible that this is a residual from a previous test of #8637?

prwolfe:
> In particular, is it possible that this is a testing residual from a previous test of #8637?

I think so and I set the flag for future testing to clean the build space.

@trilinos/framework - can I get a second person from the team to look at this and evaluate if I should force push it to get PR's moving? Honestly I can't see how that build failure could come from this code. If this was not blocking the entire PR system I would say no to a force, but at this point perhaps??

jwillenbring:
> > In particular, is it possible that this is a testing residual from a previous test of #8637?
> 
> I think so and I set the flag for future testing to clean the build space.
> 
> @trilinos/framework - can I get a second person from the team to look at this and evaluate if I should force push it to get PR's moving? Honestly I can't see how that build failure could come from this code. If this was not blocking the entire PR system I would say no to a force, but at this point perhaps??

@prwolfe Given the extreme situation we are facing, if this currently running testing attempt fails, I think we can force push as you suggest. This is solely to fix the proxy so other tests don't fail.

prwolfe:
> @prwolfe Given the extreme situation we are facing, if this currently running testing attempt fails, I think we can force push as you suggest. This is solely to fix the proxy so other tests don't fail.

Done

cgcgcg:
Thanks @prwolfe!

PR #8644, opened 2021-01-26
Intrepid2: Added CMake flag to run the "MonolithicExecutable" Intrepid2 tests in serial (as opposed to MPI), to avoid some timeouts that can occur on some CUDA platforms.

CamelliaDPG:
Intrepid2: Added CMake flag to run the "MonolithicExecutable" Intrepid2 tests in serial (as opposed to MPI), to avoid some timeouts that can occur on some CUDA platforms.

@trilinos/intrepid2 

## Motivation
As discussed in #8516, the new `Intrepid2_unit-test_MonolithicExecutable_Intrepid2_Tests_MPI_1`, which encompasses many unit tests, can time out on some CUDA test platforms.  This PR adds a flag to run this test in serial (as opposed to having multiple tests running simultaneously) to ensure that this test is not competing with others for CUDA resources.

## Testing
N/A.

PR #8645, opened 2021-01-26
atdm: Set run serial monolithic executable test 2

e10harvey:
Update the nightly tests to run in serial. Related to #8644.


e10harvey:
Wrong branches. Let me try again.

PR #8646, opened 2021-01-27
Zoltan:  new serialization capability for #8476

### LINKED ISSUE #8476: Zoltan:  Provide methods to transfer zoltan struct information between communicators
## Enhancement
@trilinos/zoltan 

In E3SM, the coupler component that interpolates values from one component's mesh (e.g., ocean) to another component's mesh (e.g., atmosphere) operates more efficiently if the two components have similar partitions.  With Zoltan_LB_Point_PP_Assign, we can infer one partition (e.g., atmosphere's) from another RCB, RIB, or HSFC partition (e.g., ocean's).  But since the ocean and atmosphere components may run on distinct sets of processors, we need a way to transfer Zoltan information from one MPI communicator to another.

Through discussions with @iulian787 and @vijaysm (designers of the new E3SM coupler), we decided that, rather than have Zoltan transfer the data between communicators, we would provide routines to serialize the zoltan struct into a buffer and restore it from a buffer.  That way, the E3SM coupler can send the information to many different components (e.g., atmosphere, land, etc.).   We decided to send the entire Zoltan struct rather than only the cut information to provide some data protection within the zoltan struct and to allow Zoltan_LB_Point_PP_Assign to work on the receiving communicator.

kddevin:
@trilinos/zoltan   

## Motivation

The iMOAB coupler would like to be able to use an RCB decomposition computed on one communication (e.g., processors doing the ocean simulation) to do geometric search (Point_Assign, Box_Assign) on another communicator (e.g., processors doing the atmosphere simulation).    To use the RCB cuts in a different communicator, Zoltan methods were needed to serialize the Zoltan_Struct data (including the RCB cuts) so they can be sent to another communicator.

This PR adds three interface functions:  Zoltan_Serialize_Size, Zoltan_Serialize, and Zoltan_Deserialize, that pack relevant Zoltan_Struct data into an appropriately sized communication buffer and then unpack it on the receiving processors.  Partitioning-method specific functions provide seralization for the specific data structures (e.g., RCB cuts) of the method.

For now, full capability is available for RCB only. Capability can be added for other methods (RIB, HSFC, PHG) on an as-needed basis, using the example of RCB.

## Related Issues

* Closes  #8476 

## Stakeholder Feedback

@iulian787 @vijaysm

## Testing

New tests copyZZ.c and copyZZ.cpp test the new interface in C and C++.  The C test exercises more capabilities than the C++ test, which exercises only the interface.   Both tests perform RCB partitioning on a subset of processors, serialize the result, broadcast it to MPI_COMM_WORLD, deserialize it, and test it by doing Point_Assign operations on the received data.

These tests pass on my mac with clang; no warnings, no errors.


egboman (approved):
This is a feature I had not thought of, but I agree it is useful to some applications (like climate). The structure of the code looks good but I did not check every line. Thanks, Karen.

vijaysm:
Thanks for working on the update Karen. We will verify the functionality
with the MOAB testbed.

PR #8611, opened 2021-01-20
Tpetra: Add per-process binary reader 

seheracer:
<!---
Be sure to select `develop` as the `base` branch against which to create this
pull request.  Only pull requests against `develop` will undergo Trilinos'
automated testing.  Pull requests against `master` will be ignored.

Provide a general summary of your changes in the Title above.  If this pull
request pertains to a particular package in Trilinos, it's worthwhile to start
the title with "PackageName:  ".

Note that anything between these delimiters is a comment that will not appear
in the pull request description once created. Most areas in this message are
commented out and can be easily added by removing the comment delimiters.

Please make sure to mark:
* Reviewers
* Assignees
* Labels

Replace <teamName> below with the appropriate Trilinos package/team name.
-->
@trilinos/tpetra @kddevin 

## Motivation
<!--- 
Why is this change required?  What problem does it solve? Please link to a github 
issue that describes the problem/issue/bug this PR solves.
-->

This PR adds the per-process binary reader into the new Tpetra reader. The details are as follows (included in the code as a comment as well):

```
//////////////////////////////////////////////////////////////////////////////                                                                                                                                     
//////////// PER-PROCESS BINARY FILE READER & RESPECTIVE FORMAT //////////////                                                                                                                                     
//   (This format is used by the upcoming readPerProcessBinary function)    //                                                                                                                                     
//////////////////////////////////////////////////////////////////////////////                                                                                                                                     
//                                                                                                                                                                                                                 
//                                                                                                                                                                                                                 
// FILE FORMAT:                                                                                                                                                                                                    
//     globalNumRows globalNumCols localNumNnzs row_1 col_1 ... row_n col_n,                                                                                                                                       
//  where n = localNumNnzs                                                                                                                                                                                         
//                                                                                                                                                                                                                 
//                                                                                                                                                                                                                 
// ASSUMPTIONS:                                                                                                                                                                                                    
//  - The nonzeros should be sorted by their row indices within each file.                                                                                                                                         
//  - Nonzeros have global row and column indices.                                                                                                                                                                 
//  - The user specifies the base <filename>, where rank i reads <filename>.i.cooBin.                                                                                                                              
//                                                                                                                                                                                                                 
//                                                                                                                                                                                                                 
// TYPES:                                                                                                                                                                                                          
//  localNumNnzs:      unsigned long long                                                                                                                                                                          
//  everything else:   unsigned int                                                                                                                                                                                
//                                                                                                                                                                                                                 
//                                                                                                                                                                                                                 
[...]

PR #8653, opened 2021-01-28
Tpetra: Add tests for binary readers (global and per-process)

seheracer:
<!---
Be sure to select `develop` as the `base` branch against which to create this
pull request.  Only pull requests against `develop` will undergo Trilinos'
automated testing.  Pull requests against `master` will be ignored.

Provide a general summary of your changes in the Title above.  If this pull
request pertains to a particular package in Trilinos, it's worthwhile to start
the title with "PackageName:  ".

Note that anything between these delimiters is a comment that will not appear
in the pull request description once created. Most areas in this message are

Please make sure to mark:
* Reviewers
* Assignees
* Labels

-->
@trilinos/tpetra

## Motivation
<!--- 
Why is this change required?  What problem does it solve? Please link to a github 
issue that describes the problem/issue/bug this PR solves.
-->
This PR adds tests for the binary readers (global and per-process readers), which have recently been added to Tpetra. 
The binary readers are located in tpetra/core/inout/MatrixMarket_TpetraNew.hpp:
Global reader: https://github.com/trilinos/Trilinos/blob/5e739265d71aa3b41b5185d5f422760a5d78ef95/packages/tpetra/core/inout/MatrixMarket_TpetraNew.hpp#L508
Per-process reader: https://github.com/trilinos/Trilinos/blob/5e739265d71aa3b41b5185d5f422760a5d78ef95/packages/tpetra/core/inout/MatrixMarket_TpetraNew.hpp#L768

The new test program is in tpetra/core/test/inout/MatrixMarket_Tpetra_CrsMatrix_Dist_Binary.cpp
It tests the readers with the following distributions: 1D, 1DRandom, 2D, 2DRandom, 2D_npRow, 2D_npCol, 2D_npCol_npRow, LowerTriangularBlock. This program is very similar to the one in MatrixMarket_Tpetra_CrsMatrix_Dist.cpp. The constructor reads the matrix from a matrix market (mtx) file using the standard MatrixMarket reader and creates A_baseline. With a randomly created input vector, x_baseline, SpMV is performed on A_baseline and the output vector is stored in y_baseline for future comparisons.                                                                                                                                                                                                    

For each tested distribution, we read the input matrix (mtx) file again and create a matrix called AmatWriter with the desired distribution. We use AmatWriter to write the nonzeros in a global binary file or multiple per-process binary files. Then we read the binary file(s) and create AmatTest with the same distribution as AmatWriter. We perfom SpMV on AmatTest with x_baseline and compare the output vector with y_baseline.                                                                                                                                                                                                                

IMPORTANT: 
1) Since the current binary formats do not support any numeric values, the binary readers set all numeric values to one. Therefore, we also set all numeric values to one in this program.                                                                                                                                                                                                             
2) Since LowerTriangularBlock distribution drops the nonzeros in the upper triangular part, its SpMV output will be different than y_baseline (which considers all nonzeros in the given mtx file). Therefore, we compare the SpMV outputs of AmatWriter and AmatTest for LowerTriangularBlock distribution.    

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows https://github.com/trilinos/Trilinos/pull/8472 https://github.com/trilinos/Trilinos/pull/8611 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback
<!--- 
If a github issue includes feedback from the relevant stakeholder(s), please link it.  
If the stakeholder(s) communicated that feedback through a different medium, please note that you did so.
-->

## Testing
<!---
Please confirm that any classes or functions in the Trilinos library that this PR touches are 
exercised by at least one test in Trilinos.  Please specify which test that is.  For untestable 
changes (e.g. changes to the nightly testing system) or changes to Trilinos tests, please say "N/A".

-->
Tests are passing on my serial build (solo) and Cuda build (white).  

<!--- 
## Additional Information
 -->


kddevin (changes requested):
@seheracer walked me through the code.  The tests are good and I am happy to have them; thank you!  

Please make the following changes:

- The LowerTriangularBlockOperator apply performs by+aAx, where A is the original matrix, not the LowerTriangularBlock matrix.  So comparisons to yout_baseline should be OK in the LTB tests.
- Global binary reading should always use a binary file equivalent to the original Matrix Market file (i.e., with the nonzeros original order) rather than with a binary file ordered according to the tested distribution.  We agreed that @seheracer would write a global binary file in the test constructor and use that file in the global binary reader tests.  The current scheme for writing per-process binary files is correct and will be retained.
- With the above changes, the tests with sorted LTB can be restored for the global binary reader.

Longer term (and absolutely not part of this PR), we should consider an Identity distribution for use with the per-process reader.

kddevin:
@trilinos/muelu we could use some help.  Several muelu tests are failing here.  I expect the failures are unrelated to this PR.  Are you seeing these failures with other PRs?  Could the new cmake be causing problems?  Thanks.
```
sed: can't read kokkos/Output/operator_solve_1_np1_tpetra.gold_filtered: No such file or directory
```

cgcgcg:
Hm, that's pretty odd. I don't think that this PR should affect the existence of our gold files, yet it seems to be the only one affected by this issue. But it seems that the failures have started before the new CMake version was used, right?

seheracer:
Disabling the newly-added binary tests worked, so, those newly-added tests are somehow corrupting the other tests :( @kddevin

packages/tpetra/core/test/inout/MatrixMarket_Tpetra_CrsMatrix_Dist_Binary.cpp
```diff
@@ -0,0 +1,822 @@
...
+  {
+    params.set("verbose", true);
+    params.set("callFillComplete", true);
+    params.set("useTimers", true);
+
+    if(perProcess)
+      runTestLTBPerProcess(testname, params);
+    else
+      runTestLTBGlobal(testname, params);
+  }
```
kddevin, line 686: Function runTestLTB needs to return the error code from runTestPerProcess or runTestGlobal.

packages/tpetra/core/test/inout/MatrixMarket_Tpetra_CrsMatrix_Dist_Binary.cpp
```diff
@@ -0,0 +1,822 @@
...
+  {
+    params.set("verbose", true);
+    params.set("callFillComplete", true);
+    params.set("useTimers", true);
+
+    if(perProcess)
+      runTestPerProcess(testname, params);
+    else
+      runTestGlobal(testname, params);
+  }
```
kddevin, line 613: Function runTest needs to return the error code from runTestPerProcess or runTestGlobal.

kddevin (changes requested):
@seheracer I checked out your branch and found a problem when running the tests.  See the comments below about returning the error count from runTest and runTestLTB.
In my clone, I re-enabled the tests by uncommenting the CMakeLists.txt entries.
Without the error codes, the binary tests failed on my mac.
When I modify runTest and runTestLTB to return the error codes, the binary tests succeed on my mac.

packages/tpetra/core/test/inout/MatrixMarket_Tpetra_CrsMatrix_Dist_Binary.cpp
seheracer: Added the missing returns and enabled the tests back.

packages/tpetra/core/test/inout/MatrixMarket_Tpetra_CrsMatrix_Dist_Binary.cpp
seheracer: Added the missing returns and enabled the tests back. Thanks for the catch! I hope this error was causing the failures.

kddevin (approved):
Tests are now passing -- yea!
Good to merge.

PR #8654, opened 2021-01-28
use Kokkos parallel-for for the local restriction

iyamazaki:
@trilinos/shylu

## Motivation
This PR adds an option to use Kokkos parallel-for for the local restriction (to avoid moving data between CPU/GPU)

## Testing
I ran ctest shylu-dd/frosch on Vortex


srajama1 (dismissed):
Thanks @iyamazaki !

srajama1 (approved):
I am assuming this new check for library name is needed because both Tpetra and Epetra are enabled. Looks good. Thanks @iyamazaki !

iyamazaki:
Thank you @searhein, and @srajama1, that's right. I added the check in case both Epetra and Tpetra were enabled. I hope that makes sense..

searhein:
@iyamazaki Looks good to me. Thanks!

PR #8665, opened 2021-02-01
Add environment variable for intel configure with intel 19

prwolfe:
A previous PR had set this up for intel 17 and
subsequent testing showed a problem with intel 19
as well. This also adds the environment variables
to the evn.sh scripts for reproducing builds.

@trilinos/framework 

# Motivation
This is blocking develop->master promotions and will be needed in general.


jwillenbring (approved):
These changes look good.

PR #8667, opened 2021-02-01
Intrepid2: fix namespaces in orientation tools

rppawlo:
## Motivation

Tried to pull these files into empire, but had compile failures. Unit tests probably are declaring "using namespace Intrepid2" to work around this.


mperego (approved):
Looks good. Thanks @rppawlo fro fixing this!

PR #8674, opened 2021-02-02
KokkosKernels: patch in supernodal fix

brian-kelley:
@trilinos/kokkos-kernels 

## Motivation

Fixes build with devices other than Serial and KokkosKernels supernodal SpTRSV enabled. Uses explicit exec+mem space pair to match correct ETI versions of SpMV, while before would get an undefined reference link time error.

## Stakeholder Feedback

Issue reported by @iyamazaki 
## Testing

Just a build change, but SpTRSV is tested in KokkosKernels

### DIFF
--- a/packages/kokkos-kernels/src/sparse/KokkosSparse_sptrsv_handle.hpp
+++ b/packages/kokkos-kernels/src/sparse/KokkosSparse_sptrsv_handle.hpp
@@ -222,11 +222,11 @@ public:
   using integer_view_t = Kokkos::View<int*, supercols_memory_space>;
   using integer_view_host_t = Kokkos::View<int*, supercols_host_memory_space>;
 
-  using workspace_t = typename Kokkos::View<scalar_t*, supercols_memory_space>;
+  using workspace_t = typename Kokkos::View<scalar_t*, Kokkos::Device<execution_space, supercols_memory_space>>;
 
   //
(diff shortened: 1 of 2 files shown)


iyamazaki:
Thank you @brian-kelley ! This fixed my builds (OpenMP on White and Cuda on Summit), with supernodal enabled. 

One question: there seems to be another work_view_t in [UpperTriTranSupernodalFunctor](https://github.com/trilinos/Trilinos/blob/e07ef9d8feeee9fc116e1bb6548186017da6b938/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_sptrsv_solve_impl.hpp#L1031). Should that have execution_space, just to be consistent?

brian-kelley:
@iyamazaki Hmm, good catch. I did not get an error from that one, so I'm going through again and seeing if I can simply use memory_space for the workspace rather than the full Device<exec, mem>.

brian-kelley:
@iyamazaki I'm opening a new PR to make that type consistent. In general the workspace_t does need to have the full Device spelled out (the build broke when I changed them to ``View<scalar*, memory_space>``).

iyamazaki (approved):
Thank you, @brian-kelley , again!!

brian-kelley:
@srajama1 Can one of you review this for kokkos-kernels? It is just patching in https://github.com/kokkos/kokkos-kernels/pull/884 and https://github.com/kokkos/kokkos-kernels/pull/885 which Nathan looked over.

srajama1 (approved):
Thanks @brian-kelley !

Issue #8675, opened 2021-02-02
Framework:  Update ATDM builds for Minimum CMake requirement


jwillenbring:
## Bug Report
@trilinos/framework 

### Description

At the SART meeting today, @rppawlo asked if the ATDM scripts had been updated to reflect the new minimum CMake version requirement (3.17). A grep of the Trilinos/cmake/std/atdm directory showed:

cee-rhel7/environment.sh:module load sems-cmake/3.12.2
common/toss3/environment_new.sh:module load cmake/3.12.2
contributed/blake/environment.sh:  module swap cmake/3.9.0 cmake/3.12.3
contributed/blake/environment.sh:  module swap cmake/3.9.0 cmake/3.12.3
contributed/kokkos-dev-2/environment.sh:module load sems-cmake/3.12.2
contributed/spack-rhel/environment.sh:module load spack-cmake/3.14.5
contributed/weaver/environment.sh:module unload cmake/3.6.2
contributed/weaver/environment.sh:module load cmake/3.12.3
cts1empire/environment_new.sh:module load cmake/3.12.2
van1-tx2/environment.sh:module load cmake/3.12.2

I am not sure if we should just quickly update these, as we know they will fail otherwise, or if we should more carefully make sure things pass with the change.

jwillenbring:
In the ATDM builds, I see 3 intel builds that are suffering from the same missing link library issue we previously resolved for the PR intel build.

https://testing.sandia.gov/cdash/build/8645899/configure

https://testing.sandia.gov/cdash/build/8645823/configure

https://testing.sandia.gov/cdash/build/8645791/configure

These are the only configure failures I see so far due to the CMake minimum version change.

jwillenbring:
Under specialized 

cts1empire/environment_new.sh:module load cmake/3.12.2
van1-tx2/environment.sh:module load cmake/3.12.2

Also failed.

@e10harvey said he would address these issues within the next couple days.

e10harvey:
> In the ATDM builds, I see 3 intel builds that are suffering from the same missing link library issue we previously resolved for the PR intel build.

I will address this later today.

jwillenbring:
@e10harvey I think this can be closed. I am going to do so. If you see a reason it should be reopened, please do so.

PR #8676, opened 2021-02-02
ATDM rhel7 cmake 3.19

jwillenbring:
@trilinos/framework 

## Motivation

Trilinos now requires CMake 3.17 or greater

Part of #8675 

This updates the rhel7 cee environment to use a supported version of CMake.


jwillenbring:
> Any chance we can switch to 3.19.1 instead? There's some feature EMPIRE utilizes that requires 3.18 or higher on some machines, so I'd be tempted to go to the latest everywhere, unless someone has a specific problem with 3.19.

I updated the PR to 3.19.1.

jmgate (approved):

PR #8679, opened 2021-02-02
Intrepid2: Allow CubatureTensor in more dimensions

CamelliaDPG:
@trilinos/intrepid2 

## Motivation
By adding a constructor that allows a tensorial extension of an existing CubatureTensor object, we allow the use of CubatureTensor in dimensions beyond 3.  This has applications for space-time finite elements, as well as for Boltzmann models (which may be up to 6D).

Additionally, this PR includes a default constructor for BasisValues, which was missing.

## Testing
Both CubatureTensor and BasisValues are tested by existing unit tests, as well as one new one that explicitly tests the new default constructor for BasisValues.


mperego (approved):
Looks good. Thanks!

PR #8680, opened 2021-02-02
SEACAS: Fix non C++-11 construct

gsjaardema:
Remove a non C++-11 construct from the code.

@trilinos/seacas


prwolfe:
@jwillenbring - this response was faster than me trying to set up an issue. I will test this file against my PR on vortex to be sure, but I think Greg has this handled.

prwolfe (approved):
This resolved the compile failure and it looks like there is a series of link errors now. I will look as that may be system specific, Thanks Greg!

Issue #8620, opened 2021-01-21
Zoltan2:  interface and testing changes needed for Quotient algorithm


kddevin:
## Bug Report
@trilinos/zoltan2 

### Description
See comments in late review of #8576 

Interface changes requested:
- Return partListView of quotient algorithm does not match that of other 
  algorithms.

- Currently returning:

  quotient:  one part assignment per MPI rank
    on each processor, Teuchos::ArrayView(1) with part(0) = 
    single value in [0,nparts)

  everything else:  one part assignment per matrix row
    Teuchos::ArrayView(nrows) with part(i) = some value in [0,nparts)

- We should return:

  quotient:  one part assignment per matrix row; all rows assigned to same
             part (until we implement chunking version)
    on each processor, Teuchos::ArrayView(nrows) with part(i) = 
    const A in [0,nparts) for every row i (same A for every row)

  everything else:  one part assignment per matrix row
    Teuchos::ArrayView(nrows) with part(i) = some value in [0,nparts)

- For quotient method, modifying partListView to return part assignment per 
  matrix row allows
    common user interface for all methods;
    matrix redistribution in the partitioning1.cpp test will work for 
      better testing; and
    with quotient, can also check that part(i) = A for all i

Algorithm changes requested:

-  To accomplish change above, might need to 
   have quotient algorithm take Adapter as input (as in AlgPuLP), 
   construct its CommGraphModel itself from adapter, 
   and then use GIDs from adapter to setParts with result.

- Make threshold_ a parameter so that for testing, we can set it 2 for some of
  the PR tests (runs on 4, but migrates to 2)

- Remove untested option to not migrate.
  User can run on 4 and migrate to 4 if he doesn't want to migrate
  Test this case

- Remove code that disables distributeInput for quotient method.

Tests:

-  test both distributeInput=false, true 
   (--no-distribute, --distribute (default))

-  set threshold to run on 4, migrate to 2

-  set threshold to run on 4, migrate to 4

### Steps to Reproduce

Not applicable

egboman:
Fixed by PR #8701.

PR #8701, opened 2021-02-04
Zoltan2: Update the Quotient algorithm interface and add more tests

seheracer:
@trilinos/zoltan2 

## Motivation

This PR changes the Quotient Algorithm interface and testing as follows:
1) It makes the AlgQuotient class take InputAdapter as input instead of CommGraphModel and creates the CommGraphModel internally in the AlgQuotient class. It removes the creation of commGraphModel from PartitioningProblem class.
2) It extends the small quotient partitioning solution to the matrix size so that each local row has a separate entry in the resulting partitioning vector.  
3) It removes the flag migrated_ and its getter function because the migration is always performed. It also removes the untested else branch for the non-migrated case. 
4) It allows user to set the threshold_ parameter.
5) It adds new tests for the Quotient algorithm with different threshold values (1 and 2) and the distribute option being true and false. 
6) It adds a new check for the quotient tests where each entry of the local partitioning vector needs to be the same.
7) It updates the MueLu's Zoltan2 interface so that the Quotient algorithm is treated the same way as other graph based Zoltan2 algorithms are treated. 

## Related Issues

* Closes https://github.com/trilinos/Trilinos/issues/8620
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing

Local tests on a Cuda build on Summit are passing.

### DIFF
--- a/packages/muelu/src/Rebalancing/MueLu_Zoltan2Interface_def.hpp
+++ b/packages/muelu/src/Rebalancing/MueLu_Zoltan2Interface_def.hpp
@@ -256,25 +256,14 @@ namespace MueLu {
 
       const typename InputAdapterType::part_t * parts = problem->getSolution().getPartListView();
 
-      if(algo == "quotient") {
-	// Quotient algorithm outputs a partition vector of size #MPIranks,
-	// where each rank gets a single entry locally, i.e., parts[0]. This entry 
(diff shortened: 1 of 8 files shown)


cgcgcg (approved):
MueLu changes look good

kddevin (approved):
OK to merge if it passes PR testing

PR #8709, opened 2021-02-08
Fixing issue seen in #8705.

jwillenbring:
Moving to new CMake version for UVM dev->master build

As seen in the dev->master testing results for #8705, the uvm free dev->master build is using an old version of CMake no longer supported. This PR upgrades to the same version used for the UVM enabled variant of the build.


jwillenbring:
@ZUUL42 plans to submit a broader PR than what this covers, so I am going to close this.

PR #8711, opened 2021-02-08
Tpetra 2D reader:  add test for excluding diagonals

kddevin:
@trilinos/tpetra 

## Motivation

In #7045 , we added a 2D matrix market reader.  It has an option to exclude or require matrix diagonal entries.  This PR adds tests for excluding diagonals with all matrix distributions supported.
Also changed option "ignore" to "exclude"; it sounds better.

## Stakeholder Feedback

@ndellingwood 

## Testing

Tested on mac with clang


seheracer (approved):
Looks good to me.

PR #8718, opened 2021-02-09
Panzer: Cleanup of worksets after changes to orientations

seamill:
…Values2. Also a bit of cleanup.

@trilinos/panzer


seamill:
@rppawlo Found the issue I was having with empire (my fault). I ran these changes by empire and it works fine. Could your review?

rppawlo (approved):

PR #8719, opened 2021-02-09
Teuchos ParameterList: Add option to toggle printing of default values

cgcgcg:
@trilinos/teuchos 

## Motivation
Add a flag to `ParamterList::PrintOptions` that toggles the printing of default options.
The default behavior has not changed: defaults are printed.

### DIFF
--- a/packages/teuchos/parameterlist/test/ParameterList/ParameterList_UnitTests.cpp
+++ b/packages/teuchos/parameterlist/test/ParameterList/ParameterList_UnitTests.cpp
@@ -1096,6 +1096,39 @@ TEUCHOS_UNIT_TEST( ParameterList, attachValidatorRecursively ) {
 // TODO: test printing of modifiers
 // valid_pl.print(std::cout, ParameterList::PrintOptions().showDoc(true).indent(2).showTypes(true));
 
+TEUCHOS_UNIT_TEST( ParameterList, print ) {
+  ParameterList valid_pl("valid_pl");
+  valid_pl.set("a", 0.);
+
+  {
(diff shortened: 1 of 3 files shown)


packages/teuchos/parameterlist/test/ParameterList/ParameterList_UnitTests.cpp
```diff
@@ -1096,6 +1096,39 @@ TEUCHOS_UNIT_TEST( ParameterList, attachValidatorRecursively ) {
...
+    std::cout << ss.str();
+    TEST_ASSERT(ss.str().size() > 0);
+
+    ss.str("");
+    printOptions.showDefault(false);
+    pl.print(ss, printOptions);
+    std::cout << ss.str();
+    TEST_ASSERT(ss.str().size() == 0);
+  }
+}
```
bartlettroscoe, line 1130: Okay, this test seems to show the new behavior.  If a parameter is not set at the default value, then it will be printed regardless of the value of `showDefault`.  But if all of the parameters are at the default values, then nothing gets printed if `showDefault=true`.

bartlettroscoe (approved):
Looks good and is a nice addition.  I can see a use case where the user would only want to see parameters not set at their default values.

And I think the new unit tests are enough to protect the basics of this.

bartlettroscoe:
@cgcgcg, thanks for the PR!  Go ahead and set `AT: AUTOMERGE` if you want.

cgcgcg:
@bartlettroscoe Thanks for the quick review!

Issue #8310, opened 2020-11-03
Intrepid2 - UVM free worksheet.


kyungjoo-kim:
This issue is for tracking work in progress for UVM removal
===========================================

# UVM Free Tools
- [x] CellTools - mauro done
- [x] Discretization - kyungjoo, mauro, nate done
   - [x] Basis - kyungjoo done
     - [x] Hierarchical components - completed by @CamelliaDPG
   - [x] FunctionSpaceTools - mauro done
   - [x] Integration: kyungjoo done
- [x] Orientation: kyungjoo done
- [x] Projection - mauro done
- [x] Shared - kyungjoo, nate done
   - [x] ArrayTools: kyungjoo done
   - [x] RealSpaceTools: kyungjoo done
   - [x] Tensor Things - completed by @CamelliaDPG

# How to

1. We have cuda version cpp files of unit tests. Replace the template parameter from Kokkos::Cuda to Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace>. Then, it does not follow the default device type in trilinos. 

2.  Compile trilinos with DEBUG build. You will see a bunch of errors. Please fix one by one.

3. Easy. Done. 

# Potential Problems and Solutions

1. Strided view cannot be created with mirroring and deep copy. We either need to double deep copy or not to allow the strided view as input when it needs mirroring.

2. Basis functions dofCoords construction is not clean. We need to change as follows
From
```
   Kokkos::DynRankView<typename ScalarViewType::value_type,typename SpT::array_layout,Kokkos::HostSpace>
         dofCoords("dofCoordsHost", this->basisCardinality_,this->basisCellTopology_.getDimension());
    dofCoords(0,0) = -1.0;
    dofCoords(1,0) =  1.0;
 
    this->dofCoords_ = Kokkos::create_mirror_view(typename SpT::memory_space(), dofCoords);
    Kokkos::deep_copy(this->dofCoords_, dofCoords);
```
To
```
    this->dofCoords_ = Kokkos::DynRankView<typename ScalarViewType::value_type,SpT>                                                                                                                                                           
      ("dofCoordsHost", this->basisCardinality_,this->basisCellTopology_.getDimension());                                                                                                                                                     
    const auto dofCoords_host = Kokkos::create_mirror_view(Kokkos::HostSpace(), this->dofCoords_);                                                                                                                                            
    dofCoords_host(0,0) = -1.0;                                                                                                                                                                                                               
[...]

PR #8745, opened 2021-02-12
Use DeviceType in Basis subclasses, etc.

CamelliaDPG:
As part of the work described in #8310, this PR replaces uses of ExecutionSpace template arguments with Device arguments, in the new structured integration support classes:
- BasisValues
- CellGeometry
- Data
- IntegrationTools
- ProjectedGeometry
- TransformedVectorData
- VectorData

In addition, it performs the same conversion for a bunch of Basis subclasses:
    - HGRAD C2 - hex, quad, tet, tri, wedge
    - HGRAD Cn - hex, quad, tet, tri, line
    - HCURL In - hex, quad, tet, tri
    - HDIV In - hex, quad, tet, tri
    - HVOL Cn - line, hex, quad, tet, tri

It also includes a commented-out static_assert in the base Basis class that asserts that the template argument is a Kokkos::Device.  This can be used during development to find places where the conversion has not been made.
@trilinos/intrepid2

## Testing
The changed code does get exercised in many ways by the tests.  However, this PR does *not* include any new CUDA-specific tests against UVM-independence, as described in #8310.

### DIFF
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_QUAD_C1_FEMDef.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_QUAD_C1_FEMDef.hpp
@@ -151,7 +151,7 @@ namespace Intrepid2 {
       }
     }
 
-    template<typename SpT,
+    template<typename DT,
              typename outputValueValueType, class ...outputValueProperties,
              typename inputPointValueType,  class ...inputPointProperties>
     void
@@ -161,7 +161,7 @@ namespace Intrepid2 {
                const EOperator operatorType )  {
       typedef          Kokkos::DynRankView<outputValueValueType,outputValueProperties...>         outputValueViewType;
       typedef          Kokkos::DynRankView<inputPointValueType, inputPointProperties...>          inputPointViewType;
-      typedef typename ExecSpace<typename inputPointViewType::execution_space,SpT>::ExecSpaceType ExecSpaceType;
+      typedef typename ExecSpace<typename inputPointViewType::execution_space,typename DT::execution_space>::ExecSpaceType ExecSpaceType;
 
       // Number of evaluation points = dim 0 of inputPoints
       const auto loopSize = inputPoints.extent(0);
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_QUAD_C1_FEMDef.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_QUAD_C1_FEMDef.hpp
@@ -151,7 +151,7 @@ namespace Intrepid2 {
       }
     }
 
-    template<typename DT,
+    template<typename SpT,
              typename outputValueValueType, class ...outputValueProperties,
              typename inputPointValueType,  class ...inputPointProperties>
     void
@@ -161,7 +161,7 @@ namespace Intrepid2 {
                const EOperator operatorType )  {
       typedef          Kokkos::DynRankView<outputValueValueType,outputValueProperties...>         outputValueViewType;
       typedef          Kokkos::DynRankView<inputPointValueType, inputPointProperties...>          inputPointViewType;
-      typedef typename ExecSpace<typename inputPointViewType::execution_space,typename DT::execution_space>::ExecSpaceType ExecSpaceType;
+      typedef typename ExecSpace<typename inputPointViewType::execution_space,SpT>::ExecSpaceType ExecSpaceType;
 
       // Number of evaluation points = dim 0 of inputPoints
       const auto loopSize = inputPoints.extent(0);
--- a/packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
+++ b/packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
@@ -79,6 +79,7 @@ namespace Intrepid2
   template<class PointScalar, int spaceDim, typename DeviceType = Kokkos::DefaultExecutionSpace::device_type>
   class CellGeometry
   {
+    static_assert(Kokkos::is_device<DeviceType>::value, "Third template argument to CellGeometry must be a Kokkos::Device");
     public:
     /*! Supported types for CellGeometry */
     enum CellGeometryType {
--- a/packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
+++ b/packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
@@ -76,7 +76,7 @@ namespace Intrepid2
       
    \note Conceptually, it would make sense to use class inheritance and have different member data for each type of geometry supported.  We instead glom all the options together into one multi-modal class; this is basically to avoid certain difficulties with vtables under CUDA.
    */
-  template<class PointScalar, int spaceDim, typename ExecSpaceType = Kokkos::DefaultExecutionSpace>
+  template<class PointScalar, int spaceDim, typename DeviceType = Kokkos::DefaultExecutionSpace::device_type>
   class CellGeometry
   {
     public:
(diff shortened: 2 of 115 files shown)


kyungjoo-kim:
@CamelliaDPG You should use Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace> for testing cuda instead of using Kokkos::DefaultExecutionSpace::device_type. The default execution space still gives you a pair of cuda and uvm space. Also you need to change this in the testing files as well. Those parts are missing.

kyungjoo-kim (changes requested):
It needs to use Kokko::Device<Kokkos::Cuda,Kokkos::CudaDevice> in testing intead of using default execution space and its device type. I also recommend not to use the default template parameter of Kokkos::DefaultExecutionSpace::device_type. If you change source files with device type, also make relevant changes in the tests. Each Intrepid2 test has separate instanciation for Serial, OpenMP and Cuda. Please see the other test. Please change the the cpp test file in the Cuda directory as explained the above and verify it works without UVM.

packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_QUAD_C1_FEMDef.hpp
```diff
@@ -161,7 +161,7 @@ namespace Intrepid2 {
                const EOperator operatorType )  {
       typedef          Kokkos::DynRankView<outputValueValueType,outputValueProperties...>         outputValueViewType;
       typedef          Kokkos::DynRankView<inputPointValueType, inputPointProperties...>          inputPointViewType;
-      typedef typename ExecSpace<typename inputPointViewType::execution_space,typename DT::execution_space>::ExecSpaceType ExecSpaceType;
+      typedef typename ExecSpace<typename inputPointViewType::execution_space,SpT>::ExecSpaceType ExecSpaceType;
```
kyungjoo-kim, line 164: @CamelliaDPG You overwrote what I changed in the previous code. As all Intrepid2 class is planned to be with device type, this SpT is not right. It looks like that you run a sed script but I do not think that it is safe.

kyungjoo-kim (changes requested):

kyungjoo-kim:
@CamelliaDPG This does not seem to be properly tested without uvm and some parts of the code is overwritten incorrectly. Could we change the code incrementally from basis ?

CamelliaDPG:
> @CamelliaDPG You should use Kokkos::DeviceKokkos::Cuda,Kokkos::CudaSpace for testing cuda instead of using Kokkos::DefaultExecutionSpace::device_type.  The default execution space still gives you a pair of cuda and uvm space. Also you need to change this in the testing files as well. Those parts are missing.

@kyungjoo-kim, I am trying to proceed incrementally, as you suggested.  I made changes in the source files, leaving the testing changes for another PR, as mentioned in this PR's description.

packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_QUAD_C1_FEMDef.hpp
CamelliaDPG: @kyungjoo-kim, sorry about that; I've now restored what you had.  I didn't intend to overwrite your changes.  I changed course in terms of the way I was handling the Impl:: classes midway through.  I had thought that we could make the DeviceType change in the outward-facing classes and still use ExecutionSpace for the Impl, but perhaps that's incorrect.  (I was working quickly, but not using scripts.)

CamelliaDPG: > @CamelliaDPG This does not seem to be properly tested without uvm and some parts of the code is overwritten incorrectly. Could we change the code incrementally from basis ?

@kyungjoo-kim, my basic approach here was to make changes in the various structured integration classes.  This led to needing to make changes in many other places just to get things building; it's close to as incremental as I could make it, starting from where I did.  Part of my effort to keep it incremental, especially in light of your concerns about slowing down tests, was not to include the UVM-free tests; those can clearly be done in an incremental way, one at a time.  I did quite a bit of work here, and would be sorry to throw it away, as you seem to be suggesting.

So, changes that should be made to this PR:
- I can change the Basis `Impl` classes to use DeviceType.
- You also recommended against using a default template argument of `Kokkos::DefaultExecutionSpace::device_type`, but you didn't say why you recommend that, and I have some reasons for wanting it.  Would you explain what your concern is?
- Any other changes to the source (as opposed to the tests) that you think need to be made?

kyungjoo-kim:
It probably does not work in that incremental way (src -> test). We should change one class at at time; its source code and test. A PR can include modification for multiple classes but each class should be tested.

CamelliaDPG:
> It probably does not work in that incremental way (src -> test). We should change one class at at time; its source code and test. A PR can include modification for multiple classes but each class should be tested.

Would you please explain what it is you think will not work?  Is it that you think the tests will fail, or is your concern something else?

packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_QUAD_C1_FEMDef.hpp
kyungjoo-kim: - To test UVM free, we are using non-default device type. By defaulting the template argument, some code path and test can use a different device including uvm memory space. 
- I do not think it is good to merge partially completed code wihtout unit testing. For many classes included in this PR, they have unit tests. Testing the unit tests without UVM is the necessary condition for this work as its objective is removal of UVM. 
- I see that this PR is big but it is "not" tested.

kyungjoo-kim:
> > It probably does not work in that incremental way (src -> test). We should change one class at at time; its source code and test. A PR can include modification for multiple classes but each class should be tested.
> 
> Would you please explain what it is you think will not work? Is it that you think the tests will fail, or is your concern something else?

We cannot predict potential problems. To remove UVM, all view should be declared with device type where the device include the cude space for cuda execution space. This type of error shows up when we instanciate the code with the device type. We are going to change the code again since humans are not better than a compiler figuring out the mismatched types.

If we can work a class by class, why don't you work that way ? We can separate the work items in a more clean way ?

kyungjoo-kim:
> @kyungjoo-kim, I am trying to proceed incrementally, as you suggested. I made changes in the source files, leaving the testing changes for another PR, as mentioned in this PR's description.

We have a different interpretation of the "incremental" way. In the #8310 descripttion, I ask to replace ``Kokkos::Cuda`` with ``Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace>`` and fix problems when we instanciate the tests. I mean a class by class by "incremental".

packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_QUAD_C1_FEMDef.hpp
CamelliaDPG: @kyungjoo-kim, thank you for clarifying.

I understand why you recommend not to default the template argument; I prefer the default because it's the case most users will want, and including it allows user code to be more succinct.

I will defer to your rules regarding testing.  I thought that we could regard this PR as not introducing a regression supposing the tests passed (since the tests would cover existing functionality), and it counts as progress, since it puts us in a position to write the UVM-free tests.  But I can see that's not your interpretation.

CamelliaDPG:
> > @kyungjoo-kim, I am trying to proceed incrementally, as you suggested. I made changes in the source files, leaving the testing changes for another PR, as mentioned in this PR's description.
> 
> We have a different interpretation of the "incremental" way. In the #8310 descripttion, I ask to replace `Kokkos::Cuda` with `Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace>` and fix problems when we instanciate the tests. I mean a class by class by "incremental".

Yes, and I started by trying to use DeviceType in my new classes, and because of various dependencies, I had to change many classes.  But if I had started with individual Basis classes, I probably would have had a different result.  I will close this PR.

Issue #8129, opened 2020-09-30
Trilinos Framework:  Create Dev->Master build on Vortex


prwolfe:
## Enhancement
@trilinos/framework 

request #7047 is being moved to run on vortex.

Current status:

A set of environment and config files has been created (the branches at https://github.com/prwolfe/Trilinos/tree/vortex_build)

Builds with c++14 at -j 44 in about 4 hours with all packages.

tests disabled in the cuda 9.2 build are still disabled
there were 38 test failures in 50 minutes

```
Total Test time (real) = 2958.00 sec

The following tests FAILED:
	608 - Zoltan_ch_simple_zoltan_parallel (Failed)
	611 - Zoltan_hg_simple_zoltan_parallel (Failed)
	1382 - SEACASAprepro_lib_aprepro_lib_unit_test (Failed)
	1383 - SEACASAprepro_lib_aprepro_lib_array_test (Failed)
	1385 - SEACASAprepro_aprepro_unit_test (Failed)
	1386 - SEACASAprepro_aprepro_array_test (Failed)
	1387 - SEACASAprepro_aprepro_command_line_vars_test (Failed)
	1388 - SEACASAprepro_aprepro_command_line_include_test (Failed)
	1389 - SEACASAprepro_aprepro_test_dump_reread (Failed)
	2920 - ROL_example_PDE-OPT_ginzburg-landau_example_02_MPI_4 (Failed)
	2935 - Piro_AnalysisDriverTpetra_MPI_4 (Failed)
	3080 - PanzerAdaptersSTK_CurlLaplacianExample-ConvTest-Quad-Order-2 (Failed)
	3081 - PanzerAdaptersSTK_CurlLaplacianExample-ConvTest-Quad-Order-3 (Failed)
	3082 - PanzerAdaptersSTK_CurlLaplacianMultiblockExample-ConvTest-Quad-Order-1 (Failed)
	3084 - PanzerAdaptersSTK_MixedCurlLaplacianExample-ConvTest-Quad-Order-1 (Failed)
	3085 - PanzerAdaptersSTK_MixedCurlLaplacianExample-ConvTest-Quad-Order-2 (Failed)
	3086 - PanzerAdaptersSTK_MixedCurlLaplacianExample-ConvTest-Quad-Order-3 (Failed)
	3087 - PanzerAdaptersSTK_MixedCurlLaplacianMultiblockExample-ConvTest-Quad-Order-1 (Failed)
[...]

PR #8755, opened 2021-02-15
Vortex build

prwolfe:
@trilinos/framework 

## Motivation
This will be used to set up a new build on vortex using gcc 8.3.1 and Cuda

## Related Issues

* Part of #8129

## Testing
Since this does not have a direct test, I set up a Jenkins job that did the inverse (merged develop into the branch) and that is shown at https://ascic-jenkins.sandia.gov/job/trilinos-folder/job/Trilinos_pullrequest_cuda_10.1.243

## Additional Information
* This will be added to the develop->master testing in a separate PR
* Test fixes/exclusions need to be set before we can put this on the master merge testing


prwolfe:
I have a successful build on the dashboard (see https://testing.sandia.gov/cdash/index.php?project=Trilinos&parentid=8717880 )

I would like to merge this and run regular builds against a change that forces everything to build (cmake changes probably) and use that to drive test failures down enough to add it to the dev->master layer. As such I am removing the WIP label

cmake/std/PullRequestLinuxDriver.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-SCRIPTFILE=$(realpath $BASH_SOURCE)
+SCRIPTFILE=$(realpath ${WORKSPACE:?}/Trilinos/cmake/std/PullRequestLinuxDriver.sh)
```
jwillenbring, line 2: I don't understand this change.

jwillenbring (approved):
Approved.

PR #8760, opened 2021-02-15
Turn on PIC for all ATDM builds

sebrowne:
EMPIRE was doing this (effectively), now SPARC wants it as well.  I
changed to the CMake method for PIC, it seemed to work when I tested it
with SPARC on CEE and ATS2 (the flag isn't alwauys -fPIC, also this
applies it to other languages).

@trilinos/framework


sebrowne:
@bartlettroscoe @jmgate 

I used this approach (CMAKE_POSITION_INDEPENDENT_CODE) in the SPARC configure scripts, and it seemed to work, without having to manually manipulate the flags.  But I'll defer to your knowledge Ross in particular if there are any issues with it.  I turned it on for everything (defaulted, still based on the ATDM configure option), and got rid of the EMPIRE-specific code that did it too.

I'm working on getting this tested now, if it works on CTS1, I'd like to push it.

jmgate:
Looks fine to me, but I'll defer to @bartlettroscoe or @e10harvey.

bartlettroscoe (changes requested):
Need to likely add a `no-fpic` option and update the documentation at:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md#fpic

as well.

Otherwise, the `fpic` keyword option just needs to be removed and `-fPIC` should just be hard-coded in the compiler flags.  But I suspect that it would be best to provide and option to disable setting `-fPIC` in some situations so I suspect that adding a `no-fpic` keyword with the default being changed as per in this PR is the right thing to do.

sebrowne:
Okay, I'll take care of it.  Also I made a syntax error discovered upon testing, but the meat of the change appears to work once I fixed that.

bartlettroscoe:
> Okay, I'll take care of it. Also I made a syntax error discovered upon testing, but the meat of the change appears to work once I fixed that.

 Thanks @sebrowne.   And I suspect that setting `CMAKE_POSITION_INDEPENDENT_CODE` is far better than directly setting `-fPIC`.  I was not aware that option even existed in CMake.    Looks like that option has been part of CMake since at least CMake 3.10.

cmake/std/atdm/README.md
```diff
@@ -214,7 +214,9 @@ is not `CLANG`, then these options are ignored.)
 **`<fpic>`:** The following `<build-name>` keyword will result in `-fPIC`
 being added to `CMAKE_CXX_FLAGS`:
 
-* `fpic`: Add `-fPIC` to `CMAKE_CXX_FLAGS`
+* `fpic`: Turn on `CMAKE_POSITION_INDEPENDENT_CODE`
```
bartlettroscoe, line 217: Can we mention that `fpic` is the default option on all platforms?  Otherwise, the documentation is not clear here.

cmake/std/atdm/utils/set_build_options.sh
```diff
@@ -227,6 +227,8 @@ fi
 # Use -fPIC or not
 if atdm_match_buildname_keyword fpic; then
   export ATDM_CONFIG_FPIC=ON
+elif atdm_match_buildname_keyword no-fpic; then
```
bartlettroscoe: This needs to be moved before the match for 'fpic' or it will not ever match 'no-fpic'.

sebrowne:
I just discovered it in the last week or so, and when I tested it seemed to work, so I figured it was better that CMake maintain the list of flags instead of me :D

bartlettroscoe (changes requested):
@sebrowne, looks good.  The last thing is to document that 'fpic' is the default in all builds.  Then this looks good to me.

bartlettroscoe:
@sebrowne and @jmgate, I just remembered, the unit tests for this will need to be updated as well.  See:

*  https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/test/unit_tests/set_build_options_unit_tests.sh

(The PR testing will catch this.)

bartlettroscoe (approved):
Looks good to me.  Thanks @sebrowne!

cmake/std/atdm/test/unit_tests/set_build_options_unit_tests.sh
```diff
@@ -403,6 +403,14 @@ testFPIC() {
...
   ${_ASSERT_EQUALS_} ${ATDM_CONFIG_FPIC} ON
 
+  ATDM_CONFIG_BUILD_NAME=default-no-fpic-after
+  . ${ATDM_CONFIG_SCRIPT_DIR}/utils/set_build_options.sh
+  ${_ASSERT_EQUALS_} ${ATDM_CONFIG_FPIC} OFF
+
+  ATDM_CONFIG_BUILD_NAME=default_no-fpic-after
+  . ${ATDM_CONFIG_SCRIPT_DIR}/utils/set_build_options.sh
+  ${_ASSERT_EQUALS_} ${ATDM_CONFIG_FPIC} OFF
+
```
e10harvey, line 413: Missing tests for `fpic`?

cmake/std/atdm/utils/set_build_options.sh
```diff
@@ -70,7 +70,7 @@ export ATDM_CONFIG_USE_CUDA=OFF
 export ATDM_CONFIG_USE_PTHREADS=OFF
 export ATDM_CONFIG_CUDA_RDC=OFF
 export ATDM_CONFIG_ADDRESS_SANITIZER=OFF
-export ATDM_CONFIG_FPIC=OFF
+export ATDM_CONFIG_FPIC=ON
```
e10harvey, line 73: Corresponding unset missing but has been added since this merged.

e10harvey (commented):
Thanks, @sebrowne !

PR #8766, opened 2021-02-16
Intrepid2 develop

kyungjoo-kim:
## Motivation

UVM removal on HGRAD C2 basis functions

## Related Issues

#8310 

## Testing

Intrepid2 unit tests are passed on Weaver with ``Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace>``

### DIFF
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_HEX_C2_FEM.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_HGRAD_HEX_C2_FEM.hpp
@@ -157,7 +157,7 @@ namespace Intrepid2 {
         
       };
       
-      template<typename ExecSpaceType, 
+      template<typename DeviceType, 
                typename outputValueValueType, class ...outputValueProperties,
                typename inputPointValueType,  class ...inputPointProperties>
       static void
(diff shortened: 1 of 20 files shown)


CamelliaDPG:
@kyungjoo-kim, I have a basic question about the template arguments and their usage.  We are changing the name of the template argument in Basis and its subclasses from ExecutionSpace to DeviceType, but we are still allowing Basis instantiations using ExecutionSpace.  Is that correct?  It's a bit confusing that DeviceType might be a DeviceType, and might be an ExecutionSpace.  I was thinking that it would be good at the base class level to accept either, but then to define in the base class both a DeviceType and an ExecutionSpace.  The logic would be something like:

```
template<class DeviceTypeOrExecutionSpace,
  typename outputValueType,
  typename pointValueType>
class Basis
{
public:
  using DeviceType = std::conditional<Kokkos::is_device<DeviceTypeOrExecutionSpace>,  DeviceTypeOrExecutionSpace, typename DeviceTypeOrExecutionSpace::device_type>;
  using ExecutionSpace = typename DeviceType::execution_space;
};
```

Subclasses could then base their local typedefs on the base class's definitions of DeviceType and ExecutionSpace.  Something like:

```
template<class DeviceTypeOrExecutionSpace,
  typename outputValueType,
  typename pointValueType>
class MyBasis : public Basis<DeviceTypeOrExecutionSpace, outputValueType, pointValueType>
{
public:
  using BasisSuper = Basis<Device,outputValueType,pointValueType>;
  using DeviceType      = typename BasisSuper::DeviceType;
  using ExecutionSpace  = typename BasisSuper::ExecutionSpace;
};
```

How does that sound to you?

kyungjoo-kim:
All kokkos execution spaces are also a device type. We do not need to specialize it for an execution space. As long as we use device type consistently, it will work with an execution space.

CamelliaDPG:
> All kokkos execution spaces are also a device type. We do not need to specialize it for an execution space. As long as we use device type consistently, it will work with an execution space.

That's not entirely true.  `Kokkos::is_device<Kokkos::Serial>::value` is false, for example, while `Kokkos::is_execution_space<Kokkos::Serial>::value` is true.

kyungjoo-kim:
The device type is higher level abstraction. Internally, if we need an execution space, then we just greb it using ``device_type::execution_space``. This also works when we use the device type with an execution space ``given_exec_space::execution_space``. So, we do not need separate code path depending on exec space or device type.

CamelliaDPG:
> The device type is higher level abstraction. Internally, if we need an execution space, then we just greb it using `device_type::execution_space`. This also works when we use the device type with an execution space `given_exec_space::execution_space`. So, we do not need separate code path depending on exec space or device type.

My point is that the thing that is passed in to Basis as an argument is not necessarily a Kokkos::Device: certainly at the moment, we have many Basis subclasses that are templated on a non-Device ExecutionSpace.  And users may have their own Basis subclasses that do the same.  We have a number of Basis subclasses that right now depend on the base class's definition of ExecutionSpace, and going forward will depend on its definition of DeviceType.  I think it's cleaner to have clarity in terms of template argument names on what acceptable arguments are (ExecutionSpace or DeviceType).  When we convert some of those subclasses, we will need the base class to provide a DeviceType typedef.  It seems better to me that this is guaranteed to actually be a Device, as it would be in what I'm proposing here.

kyungjoo-kim:
We do not need such a logic. The current code accepts an execution space. Incremental change from execution space to device space means that we want to make the code more general to accept device type so that it can be tested with cuda space. Again, users' code with execution spaces still works.

CamelliaDPG:
> We do not need such a logic. The current code accepts an execution space. Incremental change from execution space to device space means that we want to make the code more general to accept device type so that it can be tested with cuda space. Again, users' code with execution spaces still works.

I'm not saying we need it.  I'm saying it would be better if we had it.  Could we please add it?

mperego:
> > We do not need such a logic. The current code accepts an execution space. Incremental change from execution space to device space means that we want to make the code more general to accept device type so that it can be tested with cuda space. Again, users' code with execution spaces still works.
> 
> I'm not saying we need it. I'm saying it would be better if we had it. Could we please add it?

I can see both view points. Let's have a quick offline discussion so that we can decide how to proceed.

kyungjoo-kim:
I am against adding the logic. It is not necessary.

mperego:
> > > We do not need such a logic. The current code accepts an execution space. Incremental change from execution space to device space means that we want to make the code more general to accept device type so that it can be tested with cuda space. Again, users' code with execution spaces still works.
> > 
> > 
> > I'm not saying we need it. I'm saying it would be better if we had it. Could we please add it?
> 
> I can see both view points. Let's have a quick offline discussion so that we can decide how to proceed.

@kyungjoo-kim @CamelliaDPG can we meet in 15m? I see an opening on our calendar.

kyungjoo-kim:
if you are fine with this pr, please approve it so that I can merge and submit another pr using the my work branch.

CamelliaDPG:
> > > > We do not need such a logic. The current code accepts an execution space. Incremental change from execution space to device space means that we want to make the code more general to accept device type so that it can be tested with cuda space. Again, users' code with execution spaces still works.
> > > 
> > > 
> > > I'm not saying we need it. I'm saying it would be better if we had it. Could we please add it?
> > 
> > 
> > I can see both view points. Let's have a quick offline discussion so that we can decide how to proceed.
> 
> @kyungjoo-kim @CamelliaDPG can we meet in 15m? I see an opening on our calendar.

Works for me.

mperego:
> > > > > We do not need such a logic. The current code accepts an execution space. Incremental change from execution space to device space means that we want to make the code more general to accept device type so that it can be tested with cuda space. Again, users' code with execution spaces still works.
> > > > 
> > > > 
> > > > I'm not saying we need it. I'm saying it would be better if we had it. Could we please add it?
> > > 
> > > 
> > > I can see both view points. Let's have a quick offline discussion so that we can decide how to proceed.
> > 
> > 
> > @kyungjoo-kim @CamelliaDPG can we meet in 15m? I see an opening on our calendar.
> 
> Works for me.

@kyungjoo-kim can you meet now? Otherwise I sent an invite for tomorrow.

kyungjoo-kim:
host the meeting

CamelliaDPG (approved):
Works for me!

PR #8767, opened 2021-02-16
Intrepid2 develop

kyungjoo-kim:
## Motivation

uvm removal on hcurl I1 and hvol c0. this also include some hgrad cn functions. 

## Related Issues

#8310 

## Testing

Pass unit tests on weaver.


mperego (approved):
Looks good!
For future PRs I would avoid using `DT` and always use `DeviceType`. We might end up changing that anyway, but for now let's stick with `DeviceType`.

kyungjoo-kim:
I mostly use DT in the definition file and DeviceType in the header interface file.  When I do some mechanical changes, that way is a bit easier.

PR #8772, opened 2021-02-17
Intrepid2 develop

kyungjoo-kim:
## Motivation

Fix typo of Devicetype to DeviceType.
Add HostBasis function interface. This is needed for orientation.

orientation tools are refactored. 

## Related Issues

#8310 

## Testing

tested on weaver

### DIFF
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
@@ -898,6 +898,13 @@ using BasisPtr = Teuchos::RCP<Basis<DeviceType,OutputType,PointType> >;
                                     ">>> ERROR (Basis::getSubCellRefBasis): this method is not supported or should be overridden accordingly by derived classes.");
     }
 
+    /// create a basis object living on host
+    virtual BasisPtr<Kokkos::HostSpace::device_type, OutputValueType, PointValueType>
+    getHostBasis() {
+      INTREPID2_TEST_FOR_EXCEPTION( true, std::logic_error,
+                                    ">>> ERROR (Basis::getSubCellRefBasis): this method is not supported or should be overridden accordingly by derived classes.");
+    }
+
   }; // class Basis
 
   //--------------------------------------------------------------------------------------------//
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
@@ -899,8 +899,8 @@ using BasisPtr = Teuchos::RCP<Basis<DeviceType,OutputType,PointType> >;
     }
 
     /// create a basis object living on host
-    virtual BasisPtr<Kokkos::HostSpace::device_type, OutputValueType, PointValueType>
-    getHostBasis() {
+    virtual BasisPtr<typename Kokkos::HostSpace::device_type, OutputValueType, PointValueType>
+    getHostBasis() const {
       INTREPID2_TEST_FOR_EXCEPTION( true, std::logic_error,
                                     ">>> ERROR (Basis::getSubCellRefBasis): this method is not supported or should be overridden accordingly by derived classes.");
     }
--- a/packages/intrepid2/src/Orientation/Intrepid2_OrientationTools.hpp
+++ b/packages/intrepid2/src/Orientation/Intrepid2_OrientationTools.hpp
@@ -98,9 +98,8 @@
 #include "Intrepid2_HDIV_TET_I1_FEM.hpp"
 #include "Intrepid2_HDIV_WEDGE_I1_FEM.hpp"
 
-/// KK: these cannot be included now
-//#include "Intrepid2_HierarchicalBasisFamily.hpp"
-//#include "Intrepid2_NodalBasisFamily.hpp"
+#include "Intrepid2_HierarchicalBasisFamily.hpp"
+#include "Intrepid2_NodalBasisFamily.hpp"
 
 #include "Teuchos_LAPACK.hpp"
 
@@ -449,7 +448,7 @@ namespace Intrepid2 {
              typename elemNodeValueType, class ...elemNodeProperties>
     inline 
     static void
-    getOrientation(/* */ Kokkos::DynRankView<elemOrtValueType,elemOrtProperties...> elemOrts,
+    getOrientation(Kokkos::DynRankView<elemOrtValueType,elemOrtProperties...> elemOrts,
                    const Kokkos::DynRankView<elemNodeValueType,elemNodeProperties...> elemNodes,
                    const shards::CellTopology cellTopo);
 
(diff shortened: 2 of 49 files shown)


packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
```diff
@@ -898,6 +898,13 @@ using BasisPtr = Teuchos::RCP<Basis<DeviceType,OutputType,PointType> >;
                                     ">>> ERROR (Basis::getSubCellRefBasis): this method is not supported or should be overridden accordingly by derived classes.");
     }
 
+    /// create a basis object living on host
+    virtual BasisPtr<typename Kokkos::HostSpace::device_type, OutputValueType, PointValueType>
+    getHostBasis() const {
```
mperego, line 903: This is something I wanted since some time, but it might be hard to implement  for hierarchical basis.

packages/intrepid2/src/Orientation/Intrepid2_OrientationTools.hpp
```diff
@@ -369,104 +370,105 @@ namespace Intrepid2 {
...
+        \param  elemOrts      [out]  - cell orientations
+        \param  elemNodes      [in]  - node coordinates
+        \param  cellTopo       [in]  - shards cell topology
     */
     template<typename elemOrtValueType, class ...elemOrtProperties,
              typename elemNodeValueType, class ...elemNodeProperties>
     inline 
     static void
-    getOrientation(      Kokkos::DynRankView<elemOrtValueType,elemOrtProperties...> elemOrts,
+    getOrientation(/* */ Kokkos::DynRankView<elemOrtValueType,elemOrtProperties...> elemOrts,
```
mperego, line 452: I remember Kara saying that `/* */` messes up with Doxygen. Have you checked whether this didn't disrupt Doxygen?

packages/intrepid2/src/Orientation/Intrepid2_OrientationTools.hpp
```diff
@@ -98,8 +98,9 @@
 #include "Intrepid2_HDIV_TET_I1_FEM.hpp"
 #include "Intrepid2_HDIV_WEDGE_I1_FEM.hpp"
 
-#include "Intrepid2_HierarchicalBasisFamily.hpp"
-#include "Intrepid2_NodalBasisFamily.hpp"
+/// KK: these cannot be included now
+//#include "Intrepid2_HierarchicalBasisFamily.hpp"
```
mperego, line 102: @kyungjoo-kim we cannot do this. We are not in such a rush. Roger is working on including projections in Empire. There are tests using orientations and projections for hierarchical basis

packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HCURL.hpp
```diff
@@ -210,48 +203,47 @@ getCoeffMatrix_HCURL(OutputViewType &output,
   // on the subcell spanning the same space as the bases \phi_j
 
   // Tangents t_j
-  ScalarViewType subcellTangents("subcellTangents", numSubcellBasis, subcellDim);
+  Kokkos::DynRankView<value_type,host_device_type> subcellTangents("subcellTangents", numSubcellBasis, subcellDim);
```
mperego, line 206: We should be able to make Orientation run on device without much effort. What issues did you encounter?

packages/intrepid2/unit-test/Orientation/Serial/CMakeLists.txt
```diff
@@ -71,14 +71,14 @@ TRIBITS_ADD_EXECUTABLE_AND_TEST(
...
   
-TRIBITS_ADD_EXECUTABLE_AND_TEST(
-  Test_Orientation_QUAD_newBasis
-  SOURCES test_orientation_QUAD_newBasis.cpp
-  ARGS PrintItAll
-  NUM_MPI_PROCS 1
-  PASS_REGULAR_EXPRESSION "TEST PASSED"
-  ADD_DIR_TO_NAME
-  ) 
+# TRIBITS_ADD_EXECUTABLE_AND_TEST(
```
mperego, line 74: I'm against disabling tests

mperego (changes requested):
I would expect that with a little more effort we should be able have orientations run without UVM. 
I don't think it is a good strategy to disable orientation tests for hierarchical basis. We are not in such a rush.

packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HCURL.hpp
kyungjoo-kim: This is initialization. There is no gain from exporting this to device.

packages/intrepid2/src/Orientation/Intrepid2_OrientationTools.hpp
kyungjoo-kim: I can include it but I would not test it.

packages/intrepid2/src/Orientation/Intrepid2_OrientationTools.hpp
kyungjoo-kim: done.

packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
kyungjoo-kim: This is necessary in my opinion. We have frequent situation that a basis is templated. For computing reference (master) element operations, we probably want the host version.

packages/intrepid2/src/Orientation/Intrepid2_OrientationTools.hpp
kyungjoo-kim: the headers are included but it does not effect anything. I don't understand why this is included in orientation tools if they are not used.

packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
kyungjoo-kim: @CamelliaDPG @rppawlo I need this function in the hierarchical components as well to use it in the orientation tools.

packages/intrepid2/src/Orientation/Intrepid2_OrientationTools.hpp
mperego: You are probably right. Maybe it was there before we started using `getReferenceSubCell` and it's no longer needed now.

packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HCURL.hpp
mperego: What problem do you encounter without UVM, if you try to run this on device?

packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
CamelliaDPG: @kyungjoo-kim @mperego Noted.  Yes, it will be slightly more complex for the tensor bases, but I think it won't be bad.  We will just need to get host versions of the component bases, and construct the host version of the composite basis using those.

kyungjoo-kim:
@CamelliaDPG @mperego @rppawlo 

This PR is blocked by the issue of not having getHostBasis interface from the derived basis class. I tried to add the interface in the code. However, it is not that simple as some basis family functions use default exec space. The construction becomes tricky. I will leave this to @CamelliaDPG. This PR cannot be merged until then. Or I am going to submit another PR that excludes orientation tools.

kyungjoo-kim:
close this for now and open #8774

PR #8779, opened 2021-02-18
Panzer: Add edge/face blocks to STK_ExodusReaderFactory

tkordenbrock:
@trilinos/panzer 

## Motivation
The edge/face block features of STK-IO were added to the inline mesh factories but not to ExodusReaderFactory.  This PR adds the option to import or create edge/face blocks.  ParameterList options were added to individually control the use of edge/face blocks.  The options default to disabled to prevent widespread exodiff test failures.  The options are named:
* `Create Edge Blocks: <true|false>`
* `Create Face Blocks: <true|false>`

## Testing
Test cases were added to tExodusReaderFactory.


rppawlo (approved):
Thanks @tkordenbrock !

tkordenbrock:
@rppawlo Thanks!

PR #8783, opened 2021-02-19
Shylu: use scoped enum to encode NullSpaceType

mayrmt:
@trilinos/shylu 
@searhein 

## Motivation

As @searhein and I discovered today, some places don't use the `enum`, but an `int` to encode the type of nullspace. By introduction of an `enum class NullSpaceType`, one is forced to use the scoped `enum` instead of the plain `int`.

This also adapts all calls to `BuildNullSpace()`.

## Testing

All existing tests pass.


searhein (approved):
@mayrmt Thanks for this change!

Issue #872, opened 2016-11-21
Tpetra: Finish tools to determine OpenMP and MPI binding / affinities


mhoemmen:
@trilinos/tpetra 
Assigned to: @jjellio
CC: @tjfulle 

James has tools, but needs to clean them for POSIX compatibility.
Affinity checking is done.
OpenMP is not done yet.  May need to getenv to figure this out.

nmhamster:
@trilinos/tpetra what does this entail? Something like `OMP_DISPLAY_ENV=TRUE` might be sufficient for finding which binding domains are being used. It would be good to use the standard tools/environment variables if possible.

mhoemmen:
@jjellio says "Yes."  He uses VERBOSE for that variable.  aprun from Cray dumps all info from all procs to Proc 0.  It confuses information from the different processes.  Will need to figure that out.

jjellio:
@nmhamster I use the environment variable currently (OMP_DISPLAY_ENV=verbose). There are a few issues with it:
1) Some process launchers funnel all outputs to rank 0, and the OMP ENV stuff does not use the host name (or some other identifier) to make it clear which display corresponds to which rank.

2) The "environment" variables in OMP_DISPLAY_ENV, are not system environment variables, so you can't do something like getenv("FOO"), unless FOO was explicitly set prior to launch. E.g., you can programmatically access the variables if they are using their default values

I currently have a tool that figures out 'local ranks', e.g., MPI processes that share the same hardware. With this information, I then want to add the OMP information so I can record the state of each MPI process and reason about whether the environment looks correct. Currently, the tool pulls affinity information using the pthreads API (pthread_getaffinity_np). The idea is to then combine the OMP info, with affinity info, with hwloc-ls, and automate the discovery of bad mappings. For now, I simply want to capture the OMP info at the process level, I have the other stuff already complete. I can do the actual analysis using Python.

nmhamster:
@jjellio any chance we can do this without hwloc? It would make it easier on some machines if we could just directly hit NUMA calls.

jjellio:
I don't need hwloc actually.  The pthread api allows you to get good affinity info, but I am not sure how well it is supported by Cray/Intel threads. On Bowman/Shepard, this tool reports exactly the same affinity info as KMP_AFFINITY=verbose or OMP_DISPLAY_ENV verbose.

[...]

PR #8784, opened 2021-02-19
Patch in KokkosKernels #872

### LINKED ISSUE #8727: Kokkos: TeamPolicy<Cuda> team size is too large
## Bug Report
@trilinos/kokkos
I will update the issue with other packages if we're able to narrow it down.

### Description
A couple of performance tests are failing in Albany in our GPU runs on weaver:
https://sems-cdash-son.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=12122

Here is the error message:
```
 Kokkos::TeamPolicy< Cuda > the team size is too large. Team size x vector length must be smaller than 1024.
```
and the relevant issue in Albany is here: https://github.com/SNLComputation/Albany/issues/661

I've narrowed down the issue to where Albany master does not work with kokkos-promotion-3.3: https://github.com/trilinos/Trilinos/commit/e4f6782142446ec47250b7d0f9b8c125f1e59bdd
but it works with the previous commit: https://github.com/trilinos/Trilinos/commit/7b95b7c396a45f4df4d4f15a7cc6fc08357d9229

Albany doesn't currently use TeamPolicy so my guess is that it is a trilinos package causing the issue. Unfortunately, I wasn't able to narrow down the package or kernel with kokkos tools. @ndellingwood do you have any ideas of what might have caused this or maybe you know of a way I could narrow it down?

### Steps to Reproduce
1. SHA1:  https://github.com/trilinos/Trilinos/commit/e4f6782142446ec47250b7d0f9b8c125f1e59bdd
1. Weaver modules: https://github.com/SNLComputation/Albany/blob/master/doc/dashboards/weaver.sandia.gov/weaver_modules_cuda.sh
1. Trilinos configure script: https://github.com/SNLComputation/Albany/blob/master/doc/dashboards/weaver.sandia.gov/do-cmake-weaver-trilinos
1. Albany configure script: can provide if needed
1. Input deck:  can provide if needed

brian-kelley:
(fix #8727, TeamPolicy team size too large in sort_crs_*)
Adds the KokkosKernels unit test that replicated this issue.

@trilinos/kokkos-kernels 

## Motivation

Fixes bug in sort_crs_matrix/sort_crs_graph for an input graph with a high average degree. It was constructing a TeamPolicy with the desired number of threads per team, just to ask Kokkos for the recommended team size. Instead, this policy should just use 1 thread per team.

## Related Issues

* Closes #8727 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of https://github.com/kokkos/kokkos-kernels/pull/872

## Stakeholder Feedback

This bug was affecting Exawind.
## Testing

This PR includes a unit test in KokkosKernels which calls sort_crs_graph and sort_crs_matrix with a matrix with very long rows, which replicated the bug before the fix was applied.

### DIFF
--- a/packages/kokkos-kernels/src/common/KokkosKernels_SparseUtils.hpp
+++ b/packages/kokkos-kernels/src/common/KokkosKernels_SparseUtils.hpp
@@ -1042,9 +1042,11 @@ void sort_crs_matrix(const rowmap_t& rowmap, const entries_t& entries, const val
   using lno_t = typename entries_t::non_const_value_type;
   using team_pol = Kokkos::TeamPolicy<execution_space>;
   bool useRadix = !kk_is_gpu_exec_space<execution_space>();
+  lno_t numRows = rowmap.extent(0) ? rowmap.extent(0) - 1 : 0;
+  if(numRows == 0)
+    return;
   SortCrsMatrixFunctor<execution_space, rowmap_t, entries_t, values_t>
     funct(useRadix, rowmap, entries, values);
-  lno_t numRows = rowmap.extent(0) ? rowmap.extent(0) - 1 : 0;
   if(useRadix)
   {
     Kokkos::parallel_for("sort_crs_matrix", Kokkos::RangePolicy<execution_space>(0, numRows), funct);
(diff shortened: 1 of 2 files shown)


srajama1 (approved):
Thanks @brian-kelley !

PR #8786, opened 2021-02-19
Fix to NOX/LOCA

kuberry:
NOX:NodeType not specified in type being cast to.
Fixes inability to compile using clang. 

NOX:NodeType not being specified was producing a cast-to type with Kokkos::Serial, while specifying it produces a cast-to type with Kokkos::Threads.

@trilinos/nox


rppawlo (approved):

kuberry:
@trilinos/rol 

A change was made to Trilinos that should not effect ROL, however ROL is preventing merging due to a test timing out. 

https://testing.sandia.gov/cdash/test/57171336

Is this test timing set appropriately?

PR #8806, opened 2021-02-23
Piro: solve a bug when more than 1 response is used

kliegeois:
@trilinos/piro
@mperego

## Motivation

This PR is intended to solve a bug which arises when more than 1 response is used.

## Testing

The commit has been tested on Blake.


mperego (approved):
Looks good! Thanks  @kliegeois for fixing this.

PR #8807, opened 2021-02-23
Tempus: set App nonlinear solver PL if available

sconde:
Tempus was only using the default solver PL for steppers like
`RK Backward Euler`. This PR will check if the user has provided a
solver PL and make sure to override the default solver PL

@trilinos/tempus


ccober6:
@sconde Nice catch!  Can you add a unit test to cover this bug in Tempus_UnitTest_DIRK_BackwardEuler.cpp?

ccober6 (approved):
Looks good!

PR #8808, opened 2021-02-23
Tempus output to root only

sconde:
@trilinos/tempus 

## Motivation

When in parallel, the warnings are printed by each rank. On small number processor count, this is not an issue. However, on large number of processor counts, this fills up the log file ( space issue ) and burdens the user.

So to fix it, define the output stream to be a Teuchos fanceOSstream and set output to root only to true.

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorAdjointSensitivity_impl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorAdjointSensitivity_impl.hpp
@@ -377,6 +377,7 @@ describe(
   const Teuchos::EVerbosityLevel verbLevel) const
 {
   out << description() << "::describe" << std::endl;
+  out.setOutputToRootOnly(0);
   state_integrator_->describe(out, verbLevel);
   adjoint_integrator_->describe(out, verbLevel);
 }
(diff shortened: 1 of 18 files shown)


sconde:
@crockatt This PR should help with warnings filling up the log file

ccober6:
@sconde I do not think we use processor dependent output anywhere in Tempus, so I would add this to all the usages of FancyOStream (I came up with ~56 files that should be checked).  Can you make sure we are using setOutputToRootOnly(0) everywhere we can?  Thanks, Curt

sconde:
I think I've caught all of the `FancyOStream` instances

ccober6:
@sconde Should we worry about the uses of FancyOStream that are past in?  The tough one here is the application may pass a FancyOStream, and then Tempus changes it?  In talking with Roger about it, there is probably not an unset setOutputToRootOnly.  So if we are worried about this, we probably need to make a local copy and then set setOutputToRootOnly.  Thoughts?

sconde:
I think that's a valid point. I will explore it

ccober6:
@sconde Actually Roger suggested grabbing the stream from the incoming FancyOStream, and construct a local FancyOStream from that and set setOutputToRootOnly on it.

sconde:
yes, that is what I'm looking at now.

ccober6 (approved):
Looks good!

PR #8811, opened 2021-02-23
Implement getHostBasis() in newer Basis subclasses

CamelliaDPG:
@trilinos/intrepid2 

## Motivation
This is part of #8310.

Intrepid2: added getHostBasis() implementations for several Basis subclasses, along with support for DeviceType template arguments in the same.

Specifically, the following classes now support getHostBasis and DeviceType template arguments:
- DerivedBasis_HCURL_HEX
- DerivedBasis_HCURL_QUAD
- DerivedBasis_HDIV_HEX
- DerivedBasis_HDIV_QUAD
- DerivedBasis_HGRAD_HEX
- DerivedBasis_HGRAD_QUAD
- DerivedBasis_HVOL_HEX
- DerivedBasis_HVOL_QUAD
- HGRAD_LINE_Cn_FEM
- HVOL_LINE_Cn_FEM
- IntegratedLegendreBasis_HGRAD_LINE
- LegendreBasis_HVOL_LINE

Other changes:
- added typedef Intrepid2::DefaultTestDeviceType to TestUtils; this is UVM-free on CUDA, and Kokkos::DefaultExecutionSpace::device_type otherwise.
- Made use of DefaultTestDeviceType as well as getHostBasis in the BasisValues test group (in MonolithicExecutable)
- revised template arguments of Basis_DirectSumBasis, Basis_TensorBasis, andBasis_TensorBasis3; these now depend only on the base Basis class
- added typedef for BasisBase in several Basis subclasses
- added typedef for HostBasis in several Basis subclasses (specifically, those defined on the line which are used in DerivedBasis instantiations)

## Testing
The modified tests within `MonolithicExecutable` exercise getHostBasis() on all modified bases.  These also test the modified bases with the UVM-free DeviceType when run under CUDA.

### DIFF
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
@@ -75,6 +75,11 @@ class Basis;
 template <typename DeviceType = void, typename OutputType = double, typename PointType = double>
 using BasisPtr = Teuchos::RCP<Basis<DeviceType,OutputType,PointType> >;
 
+/** \brief Pointer to a Basis whose device type is on the host (Kokkos::HostSpace::device_type), allowing host access to input and output views, and ensuring host execution of basis evaluation.
+ */
+template <typename OutputType = double, typename PointType = double>
+using HostBasisPtr = BasisPtr<typename Kokkos::HostSpace::device_type, OutputType, PointType>;
+
   /** \class  Intrepid2::Basis
       \brief  An abstract base class that defines interface for concrete basis implementations for
               Finite Element (FEM) and Finite Volume/Finite Difference (FVD) discrete spaces.
@@ -902,7 +907,7 @@ using BasisPtr = Teuchos::RCP<Basis<DeviceType,OutputType,PointType> >;
     
        \return Pointer to the new Basis object.
     */
-    virtual BasisPtr<typename Kokkos::HostSpace::device_type, OutputValueType, PointValueType>
+    virtual HostBasisPtr<OutputValueType, PointValueType>
     getHostBasis() const {
       INTREPID2_TEST_FOR_EXCEPTION( true, std::logic_error,
                                     ">>> ERROR (Basis::getHostBasis): this method is not supported or should be overridden accordingly by derived classes.");
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
@@ -898,10 +898,10 @@ using BasisPtr = Teuchos::RCP<Basis<DeviceType,OutputType,PointType> >;
                                     ">>> ERROR (Basis::getSubCellRefBasis): this method is not supported or should be overridden accordingly by derived classes.");
     }
 
-    /** \brief creates and returns a basis object allocated on host.
-     
-        \return pointer to a basis allocated on host.
-     */
+    /** \brief Creates and returns a Basis object whose DeviceType template argument is Kokkos::HostSpace::device_type, but is otherwise identical to this.
+    
+       \return Pointer to the new Basis object.
+    */
     virtual BasisPtr<typename Kokkos::HostSpace::device_type, OutputValueType, PointValueType>
     getHostBasis() const {
       INTREPID2_TEST_FOR_EXCEPTION( true, std::logic_error,
--- a/packages/intrepid2/src/Discretization/Basis/Intrepid2_DerivedBasis_HCURL_HEX.hpp
+++ b/packages/intrepid2/src/Discretization/Basis/Intrepid2_DerivedBasis_HCURL_HEX.hpp
@@ -540,11 +540,11 @@ namespace Intrepid2
   
   template<class HGRAD_LINE, class HVOL_LINE>
   class Basis_Derived_HCURL_Family1_Family2_HEX
-  : public Basis_DirectSumBasis <typename HGRAD_LINE::ExecutionSpace, typename HGRAD_LINE::OutputValueType, typename HGRAD_LINE::PointValueType>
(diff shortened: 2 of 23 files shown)


CamelliaDPG:
@mperego @kyungjoo-kim, FYI, the test failures appear to be issues with the CI tester; it's not getting to the point of actually building, from the looks of it.

mperego:
> @mperego @kyungjoo-kim, FYI, the test failures appear to be issues with the CI tester; it's not getting to the point of actually building, from the looks of it.

@CamelliaDPG yeah, I've seen that in another PR today.

mperego (approved):
Looks great to me. Neat implmentation of getHostBasis.

CamelliaDPG:
> Looks great to me. Neat implmentation of getHostBasis.

Thanks, @mperego!

packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
```diff
@@ -75,6 +75,11 @@ class Basis;
 template <typename DeviceType = void, typename OutputType = double, typename PointType = double>
 using BasisPtr = Teuchos::RCP<Basis<DeviceType,OutputType,PointType> >;
 
+/** \brief Pointer to a Basis whose device type is on the host (Kokkos::HostSpace::device_type), allowing host access to input and output views, and ensuring host execution of basis evaluation.
+ */
+template <typename OutputType = double, typename PointType = double>
+using HostBasisPtr = BasisPtr<typename Kokkos::HostSpace::device_type, OutputType, PointType>;
```
kyungjoo-kim, line 81: Why do you do defaulting for the output type and point type ? This should match to this output type and point type. It does not need to default with double.

packages/intrepid2/src/Discretization/Basis/Intrepid2_DerivedBasis_HCURL_HEX.hpp
```diff
@@ -631,7 +627,7 @@ namespace Intrepid2
         \param [in] subCellOrd - position of the subCell among of the subCells having the same dimension
         \return pointer to the subCell basis of dimension subCellDim and position subCellOrd
      */
-    BasisPtr<ExecutionSpace, OutputValueType, PointValueType>
+    Teuchos::RCP<BasisBase>
```
kyungjoo-kim, line 630: Is this change meaningful ? This means that the ``Teuchos::RCP<BasisBase>`` should be the same as ``BasisPtr``. That just means that the BasisBase should be defined and inherited from the Basis class. I am not sure how this improves things in your implementation.

packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
CamelliaDPG: @kyungjoo-kim, I followed the pattern of the BasisPtr typedef that is immediately above this typedef.  It's true it doesn't need to default to double, but I think it's good to be consistent between the two typedefs.

packages/intrepid2/src/Discretization/Basis/Intrepid2_DerivedBasis_HCURL_HEX.hpp
CamelliaDPG: Yes, it is meaningful.  BasisBase may have as its first template argument a DeviceType that is distinct from the ExecutionSpace.  If so, then `BasisPtr<ExecutionSpace, OutputValueType, PointValueType>` does not match the return type in the base class.

packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
kyungjoo-kim: This is internal typedef of the basis. Defaulting the template arguments here and in the above does not make sense to me.

packages/intrepid2/src/Discretization/Basis/Intrepid2_DerivedBasis_HCURL_HEX.hpp
kyungjoo-kim: Does this work in the virtual override context if you change ?

packages/intrepid2/src/Discretization/Basis/Intrepid2_Basis.hpp
CamelliaDPG: @kyungjoo-kim, ah, I see the confusion.  Actually, these typedefs are not internal to the Basis class.

packages/intrepid2/src/Discretization/Basis/Intrepid2_DerivedBasis_HCURL_HEX.hpp
CamelliaDPG: I'm not entirely sure what you are asking.  If you are asking whether things compile if I don't make this change, the answer is no, they do not compile.

kyungjoo-kim: I mean that both returns teuchos rcp and the compiler may consider this class is overriding function. If they consider this function is different, the compiler may redirect the function to its super class interface. Previously, it probably does not compile because it uses execution space template argument instead of device argument. As you mentioned, it compiles but does it override the function correctly when the object is casted to its super class (in other words, does this make sure that it has the same function signature to override) ? Unit test will tell this.

CamelliaDPG: @kyungjoo-kim, the method is marked as override.  If the compiler considered this to be a different function from that in the base class, it would refuse to compile it.

kyungjoo-kim (approved):
I think that it looks fine.

CamelliaDPG:
> I think that it looks fine.

Thanks, @kyungjoo-kim!

kyungjoo-kim:
@CamelliaDPG Please merge this PR.

PR #8817, opened 2021-02-25
Zoltan : minor changes to make compliant with non-c99 compilers

### LINKED ISSUE #8816: Zoltan:  non-compliant build without c99
## Bug Report
@trilinos/zoltan

### Description
@iulian787 reports
```
And I did not use -std=c99 or something like that, that could have fixed the error

"../../zoltan/src/params/free_params.c: In function ‘Zoltan_Deserialize_Params’:
../../zoltan/src/params/free_params.c:132:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
   for (int i = 0; i < nParam; i++) {
   ^
../../zoltan/src/params/free_params.c:132:3: note: use option -std=c99 or -std=gnu99 to compile your code
make[2]: *** [free_params.o] Error 1
```

We try to keep zoltan compliant with older C compilers for zoltan-only builds.

kddevin:
@trilinos/zoltan 

## Motivation

Minor changes to make non-c99 compilers happy.

## Related Issues

* Closes  #8816

## Stakeholder Feedback

@iulian787

## Testing

Tested on mac with clang

### DIFF
--- a/packages/zoltan/src/params/free_params.c
+++ b/packages/zoltan/src/params/free_params.c
@@ -123,13 +123,14 @@ int Zoltan_Deserialize_Params(struct Zoltan_Struct *to, char **buf)
 {
   /* Serialize the parameters */
   char *bufptr = *buf;
+  int i;
 
   /* Unpack number of parameters */
   int nParam = *((int *)bufptr);
   bufptr += sizeof(int);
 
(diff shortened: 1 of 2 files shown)


egboman (approved):
Very minor. Looks good.

PR #8820, opened 2021-02-27
ROL: Fix rol reduced constraint simopt

dougshidong:
@trilinos/rol

## Motivation

Constraints use adjointHessian products, and therefore, the use of applyAdjointHessian_12 and applyAdjointHessian_21 were switched out within the ROL_Reduced_Constraint_SimOpt class.

This error must be due to copying from ROL_Reduced_Objective_SimOpt, where Objective_SimOpt::hessVec_12 and hessVec_21 are the non-transposed versions.

## Testing

N/A

### DIFF
--- a/packages/rol/src/function/simopt/ROL_Reduced_Constraint_SimOpt.hpp
+++ b/packages/rol/src/function/simopt/ROL_Reduced_Constraint_SimOpt.hpp
@@ -146,7 +146,7 @@ private:
   void solve_adjoint_sensitivity(const Vector<Real> &w, const Vector<Real> &v, const Vector<Real> &z, Real &tol) {
     // Evaluate full hessVec in the direction (s,v)
     conVal_->applyAdjointHessian_11(*dualstate_,w,*state_sens_,*state_,z,tol);
-    conVal_->applyAdjointHessian_12(*dualstate1_,w,v,*state_,z,tol);
+    conVal_->applyAdjointHessian_21(*dualstate1_,w,v,*state_,z,tol);
     dualstate_->plus(*dualstate1_);
     // Apply adjoint Hessian of constraint
     conRed_->applyAdjointHessian_11(*dualstate1_,*adjoint_,*state_sens_,*state_,z,tol);
@@ -311,7 +311,7 @@ public:
       solve_adjoint_sensitivity(w,v,z,tol);
       // Build hessVec
       conRed_->applyAdjointJacobian_2(ahwv,*adjoint_sens_,*state_,z,tol);
-      conVal_->applyAdjointHessian_21(*dualcontrol_,w,*state_sens_,*state_,z,tol);
+      conVal_->applyAdjointHessian_12(*dualcontrol_,w,*state_sens_,*state_,z,tol);
       ahwv.plus(*dualcontrol_);
       conVal_->applyAdjointHessian_22(*dualcontrol_,w,v,*state_,z,tol);
       ahwv.plus(*dualcontrol_);


mperego:
@dougshidong you need to create the PR against Trilinos:develop, not Trilinos:master.

dougshidong:
Ooops. Fixed it. Thanks!

dridzal:
@dougshidong , thanks, we are looking into this.  @dpkouri

dridzal:
@dougshidong , I don't think this is a bug, this is expected behavior.  I've discussed it with @dpkouri .
We are asking for the Hessian of

  `L(u,z) =  < p,  c(u,z) >,`

so the gradient is

  `L_u(u,z) = c_u(u,z)^* p`
  `L_z(u,z) = c_z(u,z)^* p.`

The 12 Hessian (differentiate `L_u`) would be

  `d_z (L_u(u,z)) = d_z(c_u(u,z)^* p)`

and the 21 Hessian (differentiate `L_z`) would be

  `d_u (L_z(u,z)) = d_u(c_z(u,z)^* p).`

I believe the implementation corresponds to this interpretation.

dougshidong:
> @dougshidong , I don't think this is a bug, this is expected behavior. I've discussed it with @dpkouri .
> We are asking for the Hessian of
> 
> `L(u,z) = < p, c(u,z) >,`
> 
> so the gradient is
> 
> `L_u(u,z) = c_u(u,z)^* p`
> `L_z(u,z) = c_z(u,z)^* p.`
> 
> The 12 Hessian (differentiate `L_u`) would be
> 
> `d_z (L_u(u,z)) = d_z(c_u(u,z)^* p)`
> 
> and the 21 Hessian (differentiate `L_z`) would be
> 
> `d_u (L_z(u,z)) = d_u(c_z(u,z)^* p).`
> 
> I believe the implementation corresponds to this interpretation.

Yes, the 12 Hessian and 21 Hessian are defined as you described. However, the available constraint's Hessian operators are **transposed**. Therefore, when using `constraint->applyAdjointHessian_21`, we're really applying the equivalent of `objective->hessVec_12`.

I think that the best way to look at it is to see the difference in those 2 functions that basically do the exact same, but are defined for Objective and Constraint:
https://github.com/trilinos/Trilinos/blob/8c4644eefa3e949932db21c0c5e8987656a055c3/packages/rol/src/function/simopt/ROL_Reduced_Constraint_SimOpt.hpp#L146-L159
https://github.com/trilinos/Trilinos/blob/8c4644eefa3e949932db21c0c5e8987656a055c3/packages/rol/src/function/simopt/ROL_Reduced_Objective_SimOpt.hpp#L145-L158

Note that I found this bug by debugging through gdb, and the application of `constraint->applyAdjointHessian_21` onto the input vector was flagged as having non-matching sizes.

dridzal:
@dougshidong , I was referring to the constraint's Hessian.  Note this is the **adjoint** Hessian, so it should be transposed.

You mentioned non-matching sizes.  Can you provide some details of how you set up your operators?  Are you using different primal and dual vector representations, or are you using the defaults?

dougshidong:
So I have been using my Hessian operators with other functions, such as the Reduced_Objective_SimOpt and it all worked out fine. I am using the default vector representation. 

I'm pretty sure that my objective (Objective_SimOpt being aerodynamic lift) and PDE constraints (Constraint_SimOpt being Euler equations) are correctly set up, since I have successfully used ROL for unconstrained reduced-space optimization through Reduced_Objective_SimOpt. I am now adding a new constraint (aerodynamic drag), and want make it a reduced constraint through the ROL_Reduced_Constraint_SimOpt class, which I think is problematic.

I am still rather convinced that `constraint->applyAdjointHessian_12` should be `constraint->applyAdjointHessian_21`. Looking at the math, we should be vmult-ing Hessian_12 (aka AdjointHessian_21) onto the vector.  Looking at Eq. 27 of our recent paper, 

https://arxiv.org/pdf/2011.13461.pdf

we do want the Hessian of the Lagrangian L_{u \alpha}, where u are the simulation variables, and \alpha the control variables, aka Hessian_12, which corresponds to AdjointHessian_21.

I'd be happy to discuss this in more details.

dridzal:
@dougshidong , the Reduced_Constraint_SimOpt code that you pointed out is currently untested, so it is definitely possible that we reversed 12 and 21.  I didn't realize you were trying to use a state constraint --thanks for clarifying this in your previous comment.  I'll discuss this more with @dpkouri , and we'll let you know.

dougshidong:
Issue still ongoing.

dridzal:
@dpkouri :  Could you please examine the fix by @dougshidong ?

PR #8821, opened 2021-03-01
Tpetra MultiVector and BlockMultiVector refactor to remove UVM requirement

kddevin:
@trilinos/tpetra @trilinos/ifpack2 

## Motivation

Trilinos is removing the requirement to use UVM in Cuda builds.
This PR refactors Tpetra's MultiVector and BlockMultiVector to remove UVM dependence.
The PR also contains updates much of Ifpack2 to use the new interfaces.

## Related Issues

* Part of #8591 

## Testing

Ran PR tests with UVM with and without CUDA on ride and cee-lan machines

Also tested without UVM (Kokkos_ENABLE_CUDA_UVM=OFF, Tpetra_ENABLE_CUDA_UVM=OFF); all MultiVector-related tests pass.
Updated SEMS configuration file for no-UVM testing to enable Tpetra testing and indicated which tests we still expect to fail without UVM.  @trilinos/framework


packages/kokkos/core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -242,6 +242,11 @@ void *CudaUVMSpace::impl_allocate(
     const char *arg_label, const size_t arg_alloc_size,
     const size_t arg_logical_size,
     const Kokkos::Tools::SpaceHandle arg_handle) const {
+#ifndef KOKKOS_ENABLE_CUDA_UVM
```
kddevin, line 245: Do we want to push these changes into the develop branch?

packages/tpetra/core/src/Tpetra_BlockMultiVector_decl.hpp
```diff
@@ -462,20 +455,20 @@ class BlockMultiVector :
   ///   MultiVector as modified in the space in which you modified
   ///   it, by calling the modify() method with the appropriate
   ///   template parameter.
-  template<class TargetMemorySpace>
-  void sync () {
-    mv_.template sync<typename TargetMemorySpace::memory_space> ();
-  }
+  // template<class TargetMemorySpace>
```
kddevin, line 458: These functions are commented out.  Since they were part of the public interface, they should, instead, be deprecated.

packages/kokkos/core/src/Cuda/Kokkos_CudaSpace.cpp
brian-kelley: @kddevin No, we definitely don't want that to go into develop. I just rolled it back for this PR.

kyungjoo-kim:
The last commit I made is to include fence in the ``getLocalView(ReadWrite)``. I tested this branch on white and was able to reproduce the failure. With the fence in the getLocaView, I can pass all NOX tests. I am wondering how this works with PR testing. We can discuss performance aspect or the souce of the problem later if this resolves other test issues.

kyungjoo-kim:
As the PR test passes with the fence on getLocalView, this could be 1) dual sync does not apply fence with uvm, or 2) another mystery of double fencing. @crtrott Does dual view sync put fence when UVM is used ?

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -3715,6 +3715,9 @@ namespace Tpetra {
       }
       throw std::runtime_error(msg);
     }
+#if defined(KOKKOS_ENABLE_CUDA)
+    Kokkos::fence();
+#endif
     owningView_.sync_host();
```
rppawlo, line 3721: @kyungjoo-kim - why ifdef for CUDA here? This seems like this will make porting to new node types similar to cuda painful to debug. I would prefer to see a fence here without the ifdefs or let's figure out if there is a fix on the nox side. Is nox doing something unexpected or unsupported?

kyungjoo-kim: @rppawlo This is just testing and temporary. We still do not exactly know what causes the tests failing. With refactoring the multivector, we remove explicit sync/modify interface from users and downstream packages. The getLocalView interface takes care of the multi memory interface. Without UVM, the dual view sync gives synchronization point (I do not test myself for other packages but tpetra tests pass with this refactoring). With UVM, the dual view syntax should sync the device although it does not perform data transfer. In this case, the above fence won't be necessary. However, some tests in nox, teko and muelu as specified in the previous PR tests fail without the above explicit fence. This is a kind of double fencing as the next dual view sync should put a fence when it is UVM. We need to investigate more to understand the dual view logic better.

kddevin:
@kyungjoo-kim Thanks; it is interesting to see that the fence fixed the tests. 
All of the failing tests used Belos, I think.  I am going to look at the Belos MultiVector adapter to see whether it or something that it uses in Tpetra needs a fence, allowing us to remove the fences from getLocalView*.

kddevin:
@rppawlo @kyungjoo-kim  In 340a16c , I replaced the fences from Tpetra::MultiVector::getLocalView*, replacing them with more surgical fences in Thyra.   @rppawlo, I would be grateful for your review.
Thanks to @hkthorn for the guided tour of Thyra adapters.  I pushed a similar fence to Belos' Tpetra adapter MvTimesMatAddMv.  We can consider whether a fence is also needed after the multiply in MvTransMv if problems arise.

@trilinos/tpetra  if someone approves this PR, we can merge it to develop if all the tests pass.  Thanks.

kyungjoo-kim:
@kddevin The fence issue is 1) it happens on Pascal GPU 2) it does not happen on Volta GPU (Timothy told me that Mulelu was not failed when he tested). I am not quite convinced that the above changes can fix the problem. With UVM, kokkos dual view sync apply ``execution_space().fence`` while ``Kokkos::fence()`` is cudaDeviceSynchronize. What I suspect is UVM sync on the Pascal needs the device synchronization instead of the execution space fence (cudaStreamSynchronize). I might be wrong. @crtrott Please fix me if I do say wrong. If that is the root cause, then the above fix can be a fix but may not be the right one.

packages/belos/tpetra/src/BelosMultiVecTraits_Tpetra.hpp
```diff
@@ -409,6 +409,12 @@ namespace Belos {
         mv.multiply (Teuchos::NO_TRANS, Teuchos::NO_TRANS,
                      alpha, A, B_mv, beta);
       }
+      Kokkos::fence();  // Belos with Thyra's MvTimesMatAddMv allowed failures
+                        // when fence was not applied after mv.multiply; 
+                        // adding the fence fixed the tests in Thyra.  
+                        // Out of an abundance of caution (and with blessing 
+                        // from @hkthorn), we add the fence here as well.  
+                        // #8821 KDD
```
kyungjoo-kim, line 417: Even if this passes the PR test, I do not understand why the fence is required here. The Tpetra kernels is launched and if getLocalView interface is properly used, it will do lazy sync when it is necessary whe the getLocalView is used next.

kyungjoo-kim:
@kddevin @crtrott 

I tested both ``Kokkos::fence`` and ``Kokkos::Cuda().fence`` at the same location before the getLocalView. The static fence passes (the PR test also shows the same test results). The second one (instance fence) fails the same set test (I just looked at the nox tests). This sounds like that the static fence is required for uvm sync instead of execution instance fence.  

I tested this on Volta. So, it is not just pascal gpu problem.

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
crtrott: are you guys actually using multiple streams anywhere? Just to make sure. I think I convinced myself that UVM DualView sync should fence globally. But it should only make a difference if you actually were using multiple streams

DavidPoliakoff: Just to verify: this is the case when the DualView is UVM? It looks like it happens any time CUDA is snabled, which sounds too broad

kyungjoo-kim: @crtrott I thought that UVM itself uses multiple streams.

kyungjoo-kim: @crtrott BTW, I use multiple streams in the ifpack2 block tridiagonal solver. Ifpack2 tests are passed and I am not sure if it is related to the previous PR test failures on the other packages.

brian-kelley:
All tests in Tpetra, Belos, Ifpack2, Teko and MueLu now pass for UVM without launch blocking. The extra fences I added to DualView sync before are gone. The idot/iallreduce refactor is done and those pass for CudaSpace.

brian-kelley:
@tasmith4 @kddevin FYI I'm working on the ifpack2 failures (the build errors were easy but now sparse BlockRelaxation is failing).

brian-kelley (dismissed):

brian-kelley (approved):

PR #8822, opened 2021-03-01
Fix for ghost orientations.

seamill:
@trilinos/panzer

### DIFF
--- a/packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.cpp
+++ b/packages/panzer/adapters-stk/src/Panzer_STK_LocalMeshUtilities.cpp
@@ -51,12 +51,13 @@
 #include "Panzer_LocalPartitioningUtilities.hpp"
 #include "Panzer_FaceToElement.hpp"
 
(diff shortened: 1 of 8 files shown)


seamill:
@rppawlo Could you review this? Most of it is copying functions from panzer_stk to panzer namespace to ensure ghost cell indexing is consistent across the code.

rppawlo (dismissed):
Thanks @seamill ! I don't see a unit test to reproduce the issue in empire. Is this something you could quickly add?

seamill:
@rppawlo Maybe, I'm still not sure how to drive that error (it really should have segfaulted at some point). I can put in a parallel_apply_orientations test that makes sure we apply orientations to the ghost cells though. That will at least exercise the capability.

seamill:
@rppawlo I added the parallel test.

rppawlo (dismissed):
Thanks @seamill !

seamill:
@rppawlo Oop, missed a warning. One more review please.

rppawlo (approved):

PR #8823, opened 2021-03-01
Teuchos core: Reorder noreturn and DLL export

### LINKED ISSUE #8818: Teuchos: C++11 noreturn attribute error in MSVS
## Bug Report
@trilinos/teuchos @ibaned 

### Description
The positioning of the C+11 noreturn attribute introduced in packages/teuchos/core/src/Teuchos_TestForException.hpp in commit 9c3771bd seems to cause trouble in MSVS 2017 and 2019.
```
TEUCHOSCORE_LIB_DLL_EXPORT [[noreturn]] void TestForTermination_terminate(const std::string &msg);
```

Error: 
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C3829	standard attribute 'noreturn' may only be applied to functions	teuchoscore	c:\local\briadam\dakota\source\packages\external\trilinos\packages\teuchos\core\src\Teuchos_TestForException.hpp	79	
```

I've so far only observed this when building with Dakota + Teuchos, but the essence is that when doing a 64-bit shared library (DLL) build on Windows, MSVS seems to insist that the __declspec(dllexport) follow the [[noreturn]]. I can't find any MSDN docs that specify allowable interactions between the C++11 standard attributes and MS-specific extensions, but I can say that reordering so the attribute precedes the MS __declspec() works:

```
[[noreturn]] TEUCHOSCORE_LIB_DLL_EXPORT void TestForTermination_terminate(const std::string &msg);
```
This can also be demonstrated in a little test program in MSVS that has these signatures:
```
[[noreturn]] __declspec(dllexport) void works() { std::exit(1); }
 __declspec(dllexport) [[noreturn]] void fails() { std::exit(1); }

```
Unless the standard offers guidance, I'd suggest just reordering so the noreturn appears first. I haven't yet verified life is good on other compilers, but should have info soon as I'm making that change in Dakota's Teuchos snapshot.

briadam:
Reorder so [[noreturn]] precedes TEUCHOSCORE_LIB_DLL_EXPORT in
exception testers to restore DLL builds on MSVS 2015 and 2019.

Fixes #8818

@trilinos/teuchos
@ibaned 

## Related Issues
* Closes  #8818 

## Testing
Problem manifests on Windows MSVS, DLL builds only, so likely not covered by current PR testing. All other builds enabling Teuchos should hit the edited line of code, though.

### DIFF
--- a/packages/teuchos/core/src/Teuchos_TestForException.hpp
+++ b/packages/teuchos/core/src/Teuchos_TestForException.hpp
@@ -76,7 +76,7 @@ TEUCHOSCORE_LIB_DLL_EXPORT void TestForException_setEnableStacktrace(bool enable
 TEUCHOSCORE_LIB_DLL_EXPORT bool TestForException_getEnableStacktrace();
 
 /** \brief Prints the message to std::cerr and calls std::terminate. */
-TEUCHOSCORE_LIB_DLL_EXPORT [[noreturn]] void TestForTermination_terminate(const std::string &msg);
+[[noreturn]] TEUCHOSCORE_LIB_DLL_EXPORT void TestForTermination_terminate(const std::string &msg);
 
 
 } // namespace Teuchos


ibaned (approved):

briadam:
Looks like we're good! I don't have permission to add the AT: AUTOMERGE label AFAICT, so will leave to maintainers to close this one out.

PR #8774, opened 2021-02-17
Intrepid2 - add host basis function interface

kyungjoo-kim:
## Motivation

Add host basis function interface. This is required for orientation tools. I also expect that this can be useful for other purpose as well. 

## Related Issues

#8310 
#8772 - This PR is closed as it depends on the get host basis interface of all.

## Testing

tested on weaver.

PR #8824, opened 2021-03-01
Intrepid2 - uvm removal on orientation tools

kyungjoo-kim:
## Motivation

#8310. With completing getHostBasis interface (#8811, #8774) for all base functions, this orientation tools will work wtihout uvm.

## Testing

Tested on Weaver.


packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HCURL.hpp
```diff
@@ -210,48 +203,47 @@ getCoeffMatrix_HCURL(OutputViewType &output,
...
   // restrict \xi_j (and corresponding t_j) to points internal to the HCURL basis
-  PointViewType refPtsSubcell("refPtsSubcell", ndofSubcell, subcellDim);
-  ScalarViewType refSubcellTangents("subcellTangents", ndofSubcell, subcellDim);
-  auto tagToOrdinal = Kokkos::create_mirror_view_and_copy(typename ExecutionSpace::memory_space(), basisPtr->getAllDofOrdinal());
-  Kokkos::parallel_for( Kokkos::RangePolicy<ExecutionSpace>(0, ndofSubcell),
-      KOKKOS_LAMBDA (const int i){
+  Kokkos::DynRankView<value_type,host_device_type> refPtsSubcell("refPtsSubcell", ndofSubcell, subcellDim);
+  Kokkos::DynRankView<value_type,host_device_type> refSubcellTangents("subcellTangents", ndofSubcell, subcellDim);
+  auto tagToOrdinal = basisPtr->getAllDofOrdinal();
+  for (ordinal_type i=0;i<ndofSubcell;++i) {
```
CamelliaDPG, line 232: @kyungjoo-kim, is there any reason not to keep a parallel_for here, with `Kokkos::RangePolicy<typename host_device_type::ExecutionSpace>`?  (The advantage is that OpenMP would get to do this in parallel.)

packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HGRAD.hpp
```diff
@@ -201,64 +194,59 @@ getCoeffMatrix_HGRAD(OutputViewType &output,
...
-  Kokkos::parallel_for( Kokkos::RangePolicy<ExecutionSpace>(0, ndofSubcell),
-        KOKKOS_LAMBDA (const int i){
+  Kokkos::DynRankView<value_type,Kokkos::LayoutLeft,host_device_type>
+    PsiMat("PsiMat", ndofSubcell, ndofSubcell),
+    PhiMat("PhiMat", ndofSubcell, ndofSubcell);
+  
+  auto cellTagToOrdinal = cellBasis.getAllDofOrdinal();
+  auto subcellTagToOrdinal = subcellBasis.getAllDofOrdinal();
+
+  for (ordinal_type i=0;i<ndofSubcell;++i) {
```
CamelliaDPG, line 204: Here again with the for v. parallel_for.

packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HCURL.hpp
kyungjoo-kim: I do not think that it can gain performance from parallel operation even for the openmp case.  
1. the workload is quite small as the operation is just filling the data for a single matrix. 
2. in my opinion, sharing the work with multiple threads just cause more overhead as it can share the data on different caches losing locality.

packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HGRAD.hpp
kyungjoo-kim: See the above comment.

packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HCURL.hpp
CamelliaDPG: @kyungjoo-kim, OK; thanks for clarifying.  That sounds good to me.

CamelliaDPG (approved):
@kyungjoo-kim, these changes look good to me.  My only other comment is to make sure you call @mperego's attention to the comments you left for him in the code.

mperego:
>   /// KK: mauro, this is an expensive construction for high order elements  and we repeat this for all possible orientation combinations we need to address this later

@kyungjoo-kim how do you plan to address this?

kyungjoo-kim:
> > /// KK: mauro, this is an expensive construction for high order elements  and we repeat this for all possible orientation combinations we need to address this later
> 
> @kyungjoo-kim how do you plan to address this?

I don't know. I may use unordered map to construct a set of subbasis. However, it is still initialization part. So we can just leave it as it is. I leave the comment because we need to first look at this part if someone complains about that the initialization phase is expensive.

PR #8830, opened 2021-03-02
Tpetra:  remove UVM dependence from test

kddevin:
@trilinos/tpetra 

## Motivation

When UVM dependence is removed, one cannot call function FixedHashTable.get(key) from host for FixedHashTables that are build on device.  This PR replaces the host accesses with device accesses.

#8209 

## Testing

Tested on cee-lan gpus without UVM.

### DIFF
--- a/packages/tpetra/core/test/HashTable/FixedHashTableTest.cpp
+++ b/packages/tpetra/core/test/HashTable/FixedHashTableTest.cpp
@@ -176,32 +176,30 @@ namespace { // (anonymous)
          << numKeys << " != vals.extent(0) = " << vals.extent (0)
          << ".");
 
-      auto keys_h = Kokkos::create_mirror_view (keys);
-      Kokkos::deep_copy (keys_h, keys);
-      auto vals_h = Kokkos::create_mirror_view (vals);
-      Kokkos::deep_copy (vals_h, vals);
       size_type badCount = 0;
-      for (size_type i = 0; i < keys.size (); ++i) {
-        const KeyType key = keys_h(i);
-        const ValueType expectedVal = vals_h(i);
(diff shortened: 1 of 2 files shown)


packages/tpetra/core/test/HashTable/FixedHashTableTest.cpp
```diff
@@ -933,20 +930,11 @@ namespace { // (anonymous)
...
-      //
-      // Kokkos::CudaSpace::access_error attempt to execute Cuda
-      //   function from non-Cuda space
-      // Traceback functionality not available
-      if (! std::is_same<typename DeviceType::memory_space, Kokkos::CudaUVMSpace>::value) {
-        out << "Testing copy constructor to CudaUVMSpace" << endl;
-        TestCopyCtor<KeyType, ValueType, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaUVMSpace>,
+      if (! std::is_same<typename DeviceType::memory_space, Kokkos::CudaSpace>::value) {
+        out << "Testing copy constructor to CudaSpace" << endl;
+        TestCopyCtor<KeyType, ValueType, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>,
```
tasmith4, line 935: With this and the other similar changes, is there a reason to not just use the default memory space for Kokkos::Cuda?  That way we can test with UVM or non-UVM, depending on the build.

kddevin:
@tasmith4 FixedHashTable actually does NOT use UVM, even for UVM=ON builds.  It is strictly Cuda or Host in CUDA builds.  In this way, we remove use of UVM in Tpetra::Map.

tasmith4 (approved):

PR #8832, opened 2021-03-02
Tempus: Remove ParameterList from SolutionHistory

ccober6:
* SolutionHistory is no longer inherits from ParameterListAcceptor.
 * Moved SolutionHistory constructor with ParameterList to a
   nonmember constructor, createSolutionHistoryPL(pl).
 * Added function to set the history, setHistory().
 * Simplified the internal management of "Storage Type".
   - Can set with enum StorageType or a string.
   - Removed ParameterList validator for enum storageType.
 * Added methods to initialize SolutionHistory similar to
   other Tempus objects.
 * Added checks to addState to ensure new state's index does not
   match an index already in the SolutionHistory. This can
   be turned off.
 * Fixed bug in addWorkingState, where state was added to
   history (chronologically) and then the TIME was changed.
 * Added unit tests.

@trilinos/tempus

### DIFF
--- a/packages/tempus/src/Tempus_SolutionHistory_decl.hpp
+++ b/packages/tempus/src/Tempus_SolutionHistory_decl.hpp
@@ -246,10 +246,16 @@ public:
     /// Get the maximum storage of this history
     int getStorageLimit() const {return storageLimit_;}
 
+    /// Set the storage type via enum.
     void setStorageType(StorageType st);
+
+    /// Get the enum storage type.
     StorageType getStorageType() const {return storageType_;}
 
+    /// Set the storage type via string.
     void setStorageTypeString(std::string st);
+
+    /// Set the string storage type.
     std::string getStorageTypeString() const;
 
     /// Return the current minimum time of the SolutionStates
@@ -283,8 +289,11 @@ public:
     Teuchos::RCP<SolutionState<Scalar> > getStateTimeIndex(int index, bool warn = true) const;
   //@}
 
-    Teuchos::RCP<const Teuchos::ParameterList> getValidParameters() const;
-    Teuchos::RCP<Teuchos::ParameterList> getNonconstParameterList();
+  /// Return a valid ParameterList with current settings.
+  Teuchos::RCP<const Teuchos::ParameterList> getValidParameters() const;
+
+  /// Return a valid non-const ParameterList with current settings.
+  Teuchos::RCP<Teuchos::ParameterList> getNonconstParameterList();
 
   /// \name Overridden from Teuchos::Describable
   //@{
@@ -303,11 +312,19 @@ public:
     Teuchos::RCP<Interpolator<Scalar> > unSetInterpolator();
   //@}
 
+  /// Print information on States in the SolutionHistory.
   void printHistory(std::string verb="low") const;
 
+  /** \brief Initialize SolutionHistory
+   *
+   *  This function will check if all member data is initialized
+   *  and is consistent.  This function does not make member data
+   *  consistent, but just checks it.  This ensures it is inexpensive.
+   */
   void initialize() const;
+
+  /// Return if SolutionHistory is initialized.
   bool isInitialized() { return isInitialized_; }
-  void checkInitialized();
 
 
 protected:
--- a/packages/tempus/src/Tempus_SolutionHistory_decl.hpp
+++ b/packages/tempus/src/Tempus_SolutionHistory_decl.hpp
@@ -117,9 +117,7 @@ enum StorageType {
 template<class Scalar>
 class SolutionHistory
   : virtual public Teuchos::Describable,
-    virtual public Teuchos::VerboseObject<SolutionHistory<Scalar> >,
-    virtual public Teuchos::ParameterListAcceptor
-//, virtual public InterpolatorAcceptingObjectBase<Scalar>
+    virtual public Teuchos::VerboseObject<SolutionHistory<Scalar> >
 {
 public:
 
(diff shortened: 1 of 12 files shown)


packages/tempus/src/Tempus_SolutionHistory_decl.hpp
```diff
@@ -245,35 +246,45 @@ class SolutionHistory
     /// Get the maximum storage of this history
     int getStorageLimit() const {return storageLimit_;}
 
-    void setStorageType(StorageType st) {storageType_ = st;}
-    StorageType getStorageType() {return storageType_;}
+    void setStorageType(StorageType st);
```
sconde, line 249: please add space and doxygen to these

packages/tempus/src/Tempus_SolutionHistory_decl.hpp
```diff
@@ -245,35 +246,45 @@ class SolutionHistory
...
+    StorageType getStorageType() const {return storageType_;}
+
+    void setStorageTypeString(std::string st);
```
sconde, line 252: same here

packages/tempus/src/Tempus_SolutionHistory_decl.hpp
```diff
@@ -245,35 +246,45 @@ class SolutionHistory
...
+     *
+     *  If not available return Teuchos::null;
+     */
+    Teuchos::RCP<SolutionState<Scalar> > getStateTimeIndex(int index, bool warn = true) const;
   //@}
 
-  /// \name Overridden from Teuchos::ParameterListAcceptor
-  //@{
-    void setParameterList(const Teuchos::RCP<Teuchos::ParameterList> & pl);
     Teuchos::RCP<const Teuchos::ParameterList> getValidParameters() const;
```
sconde, line 286: here too

packages/tempus/test/Integrator/Tempus_IntegratorBasic_ref.xml
```diff
@@ -8,6 +8,7 @@
       <Parameter docString="" id="4" isDefault="true" isUsed="true" name="Storage Limit" type="int" value="2"/>
       <ParameterList name="Interpolator">
         <Parameter name="Interpolator Type" type="string" value="Lagrange"/>
+        <Parameter name="Order" type="int" value="0"/>
```
sconde, line 11: The order is now required?

packages/tempus/src/Tempus_SolutionHistory_decl.hpp
ccober6: Done!

packages/tempus/src/Tempus_SolutionHistory_decl.hpp
ccober6: Done!

packages/tempus/test/Integrator/Tempus_IntegratorBasic_ref.xml
ccober6: No, but to check if the defaults have been set correctly, I needed to have it set in the reference ParameterList.

ccober6:
@sconde Do you have anything else I should look at?  Otherwise I will commit the previous changes and get the final review approval.  :)

sconde:
Nothing more. When you push the changes, I will approve the PR

ccober6:
@sconde FYI, I just pushed those changes, so you can approve again when you are ready.  Thanks!

ccober6:
@sconde Just a quick reminder on approving this.  Thanks, Curt

ccober6:
@rppawlo Can you also review this one?  Thanks!

packages/tempus/src/Tempus_SolutionHistory_decl.hpp
```diff
@@ -292,19 +312,33 @@ class SolutionHistory
...
+  /** \brief Initialize SolutionHistory
+   *
+   *  This function will check if all member data is initialized
+   *  and is consistent.  This function does not make member data
+   *  consistent, but just checks it.  This ensures it is inexpensive.
+   */
+  void initialize() const;
+
+  /// Return if SolutionHistory is initialized.
+  bool isInitialized() { return isInitialized_; }
```
rppawlo, line 327: Should this be a const function?

packages/tempus/src/Tempus_SolutionHistory_decl.hpp
```diff
@@ -292,19 +312,33 @@ class SolutionHistory
...
   std::string                               name_;
-  Teuchos::RCP<Teuchos::ParameterList>      shPL_;
   Teuchos::RCP<std::vector<Teuchos::RCP<SolutionState<Scalar> > > > history_;
   Teuchos::RCP<Interpolator<Scalar> >       interpolator_;
   StorageType                               storageType_;
   int                                       storageLimit_;
 
   Teuchos::RCP<SolutionState<Scalar> > workingState_; ///< The state being worked on
+
+  mutable bool isInitialized_;     ///< Bool if SolutionHistory is initialized.
```
rppawlo, line 340: Why do you need mutable? Shouldn't the state of this flag should only change in non-const functions that change the state of the object?

rppawlo (approved):

packages/tempus/src/Tempus_SolutionHistory_decl.hpp
ccober6: So here was the thought.  Please let me know if you have a better way to do this.

The initialize() function is const, so that the object is not changed during initialize() (e.g., no expensive operations are done: memory allocation, etc.).  However we need to set isInitialized_ = true, if everything is initialized.  Thus I made isInitialized_ mutable.  Thoughts?

PR #8839, opened 2021-03-03
Correct setenv left in bash scripts

prwolfe:
These were copied from the ini file without interpretation
which of course is incorrect.

@trilinos/framework 

## Motivation
Fails in Production - reported by Ross.


bartlettroscoe (approved):
That should work.

bartlettroscoe:
Related to SEPW-203

Issue #8797, opened 2021-02-22
SEACAS: Failing tests in candidate PR build for ats2


prwolfe:
## Bug Report
@trilinos/seacas 

### Description
The tests

```
SEACASAprepro_aprepro_array_test
SEACASAprepro_aprepro_command_line_include_test
SEACASAprepro_aprepro_command_line_vars_test
SEACASAprepro_aprepro_test_dump_reread
SEACASAprepro_aprepro_unit_test
SEACASAprepro_lib_aprepro_lib_array_test
SEACASAprepro_lib_aprepro_lib_unit_test
```

Are failing in the candidate build for ats2 on vortex. They seem to be differences with gold files mainly.

These need to run clean in oder to complete #8129 

### Steps to Reproduce
SHA1: develop
Use the PR environment from cmake/std/sems/PullRequestCuda10.1.243TestingEnv.sh and instructions from https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors

prwolfe:
The combination of these two PR's resolved this issue. Thanks @gsjaardema !

PR #8841, opened 2021-03-04
Fix an error introduced in PR 8829

prwolfe:
Basically the set command was missed on this line
which leaves cmake very confused.

@trilinos/framework 

## Related Issues

* Follows #8829
* Related to #8797


gsjaardema (approved):

prwolfe:
Thanks @gdsjaar!

PR #8842, opened 2021-03-04
Only add -lcurl if NetCDF has DAP enabled.

dc-snl:
Some machines (e.g. Vortex) do not have libcurl.a. Since `-lcurl` automatically gets added to TPL_Netcdf_LIBRARIES, Trilinos will fail to build unless someone adds a dummy library (e.g. `ar cr libcurl.a`) somewhere in LIBRARY_PATH. We should only add curl as a dependency if `nc-config` reports that NetCDF was configured with DAP2 or DAP4 support (the NetCDF modules don't have DAP support on Vortex).

In short, this will fix out problem building Trilinos on Vortex.

@e10harvey @jmgate


jmgate:
This looks fine to me, but we'll want @e10harvey to review and approve before merging in.  @e10harvey, if you'd prefer to delegate to someone else, feel free.

e10harvey (approved):
LGTM. Thanks, @dc-snl!

e10harvey:
@dc-snl: Should the target branch be develop?

e10harvey:
@ZUUL42, FYI.

e10harvey (approved):

dc-snl:
Oops, lemme rebase real quick.

dc-snl:
Okay, rebased. Good catch. Thanks, @e10harvey

e10harvey (dismissed):
@dc-snl: I should mention, since this change touches all builds, we may need a follow-up PR or a revert if this causes builds to crash on the machines we care about.

dc-snl:
@e10harvey 

Sounds good to me. I am counting on `nc-config` being somewhere in `PATH`. That may or may not be a good assumption.

e10harvey:
> Sounds good to me. I am counting on `nc-config` being somewhere in `PATH`. That may or may not be a good assumption.

Before merging, I would at least test this out on ats2 via:
```
mkdir Trilinos-pr8842
cd Trilinos-pr8842
ln -s /path/to/Trilinos/cmake/std/atdm/ctest-s-local-test-driver.sh
nohup env ./ctest-s-local-test-driver.sh all &>ctest-s-local-test-driver.out &
```

dc-snl:
> > Sounds good to me. I am counting on `nc-config` being somewhere in `PATH`. That may or may not be a good assumption.
> 
> Before merging, I would at least test this out on ats2 via:
> 
> ```
> mkdir Trilinos-pr8842
> cd Trilinos-pr8842
> ln -s /path/to/Trilinos/cmake/std/atdm/ctest-s-local-test-driver.sh
> nohup env ./ctest-s-local-test-driver.sh all &>ctest-s-local-test-driver.out &
> ```

Will do. I've already tested this on Vortex (DAP disabled) and a machine with SEMS (DAP enabled).

cmake/tribits/common_tpls/FindTPLNetcdf.cmake
```diff
@@ -85,10 +85,16 @@ IF (Netcdf_ALLOW_PREFIND)
...
   ENDIF()
 ELSE()
-  # Curl library is only required if DAP is enabled; should detect inside FindNetCDF.cmake,
-  # but that is not being called... SEMS has DAP enabled; many HPC systems don't, but they
-  # override these settings...
-  SET(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} curl)
+  # Curl library is only required if DAP is enabled; should detect inside
+  # FindNetCDF.cmake, but that is not being called... SEMS has DAP enabled;
+  # many HPC systems don't, but they override these settings...
+  EXECUTE_PROCESS(COMMAND "nc-config --has-dap2"
```
e10harvey, line 91: @dc-snl: Perhaps check for existence of nc-config. If it doesn't exist, `SET(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} curl)`.

dc-snl: Good idea. Testing this out now.

dc-snl: Finally got a chance to test this out and push. Seems to work on SEMS and Vortex. Testing on Astra now.

dc-snl: Hang on. It doesn't work on Vortex. I'll look into this tomorrow.

dc-snl: Ah, okay. Forgot to wipe cmake. So building Trilinos on Vortex, Stria, and SEMS seems to work.

Running `ctest-s-local-test-driver.sh` on Vortex now.

jmgate (approved):
I'm going to approve because I'd like to see the autotester run on this.  Don't merge it yet.

e10harvey (approved):
Thanks, @dc-snl. I see the ats2 results [here](https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&date=2021-03-06&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=ats2&field2=groupname&compare2=61&value2=Experimental). In reviewing the mass failures (900+ failing tests) I see that this is not the first time we've seen these as shown by this [query](https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-10-01&end=2021-03-06&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-&field2=status&compare2=62&value2=passed&field3=testoutput&compare3=95&value3=ptssup_mkcltsock_afunix). I am not sure why `Trilinos-atdm-ats2-gnu-7.3.1-spmpi-rolling_serial_static_{opt,dbg}` were not run by `ctest-s-local-test-driver.sh all` but the cuda builds that were run test these same build options.

jmgate:
@bartlettroscoe, we've been banging our heads against the wall trying to figure out why this change didn't work.  We're still running into issues on vortex for opt builds when trying to build EMPIRE against our latest Trilinos deploy.  At this point, I'm wondering if something is going sideways within TriBITS, but that's where we need your expertise.  Can you take a look?

Also it looks like this `FindTPLNetcdf.cmake` might've been modified within Trilinos prior to this PR and then that change wasn't merged back into TriBITS itself.  Not sure if that was intentional or an oversight.

bartlettroscoe:
> we've been banging our heads against the wall trying to figure out why this change didn't work

@jmgate, what do you mean by "work"?   What behavior are you seeing and what behavior are you expecting?

I am confused as to what use case on 'vortex' this PR is trying to address. Is this for the current supported ATDM Trilinos 'ats2' configuration or a new custom ATDM Trilinos configuration that is being developed for 'vortex'?  If it is for the former, this PR should not change that 'ats2' configuration in any way (I will explain in my review that I will post).

> At this point, I'm wondering if something is going sideways within TriBITS

There are no defects in TriBITS.  It behaves as coded.  There is only defects in TriBITS documentation or violations of common sense :-)

Related to my epic SEPW-123

cmake/std/atdm/ATDMDevEnvSettings.cmake
```diff
@@ -524,7 +524,7 @@ ELSEIF (TPL_ENABLE_Netcdf )
       SET(PNETCDF_ROOT "$ENV{PNETCDF_ROOT}")
     ENDIF()
     ASSERT_DEFINED(ENV{ZLIB_ROOT})
-    ATDM_SET_CACHE(TPL_Netcdf_LIBRARIES "$ENV{NETCDF_ROOT}/lib/libnetcdf.${ATDM_TPL_LIB_EXT};${PNETCDF_ROOT}/lib/libpnetcdf.a;${HDF5_ROOT}/lib/libhdf5_hl.${ATDM_TPL_LIB_EXT};${HDF5_ROOT}/lib/libhdf5.${ATDM_TPL_LIB_EXT};$ENV{ZLIB_ROOT}/lib/libz.${ATDM_TPL_LIB_EXT};-ldl;-lcurl" CACHE FILEPATH)
+    ATDM_SET_CACHE(TPL_Netcdf_LIBRARIES "$ENV{NETCDF_ROOT}/lib/libnetcdf.${ATDM_TPL_LIB_EXT};${PNETCDF_ROOT}/lib/libpnetcdf.a;${HDF5_ROOT}/lib/libhdf5_hl.${ATDM_TPL_LIB_EXT};${HDF5_ROOT}/lib/libhdf5.${ATDM_TPL_LIB_EXT};$ENV{ZLIB_ROOT}/lib/libz.${ATDM_TPL_LIB_EXT};-ldl" CACHE FILEPATH)
```
bartlettroscoe, line 527: Note that if `ATDM_USE_SPARC_TPL_FIND_SETTINGS` is true, then this statement is never even called!  That is the case for the ATDM Trilinos 'ats2' configuration so this change does not impact the ATDM Trilinos 'ats2' configuration at all.

bartlettroscoe (commented):
I am confused as to what use case on 'vortex' this PR is trying to address. Is this for the current supported 'ats2' configuration or a new custom ATDM Trilinos configuration that is being developed for 'vortex'?  If it is for the former, these changes involve code that does not even get run for that 'ats2' configuration.  If it is for the latter and the `TPL_Netcdf_LIBRARIES` var is getting set, then the code changed in the file `FindTPLNetcdf.cmake` will not have any effect.

First, note that the ATDM Trilinos configuration for 'ats2' does **not** set `TPL_Netcdf_LIBRARIES` but instead sets `ATDM_USE_SPARC_TPL_FIND_SETTINGS` to true and allows running `FIND_PACKAGE(Netcdf ...)` as you can see, for example, for the `Trilinos-atdm-ats2-gnu-7.3.1-spmpi-rolling_serial_static_dbg` build here:

*  https://testing.sandia.gov/cdash/build/8878646/configure

which shows:

```
Processing enabled TPL: Netcdf (enabled explicitly, disable with -DTPL_ENABLE_Netcdf=OFF)
-- Using FIND_PACKAGE(Netcdf ...) ...
```

You would **not** see `Using FIND_PACKAGE(Netcdf ...)` if the configuration had set `TPL_Netcdf_LIBRARIES` (which the 'sems-rhel7' configurations do).

So the code being changed here for setting `TPL_<TPLNAME>_LIBRARIES` does not impact the 'ats2' configuration in any way that I can see.

Also, even if this was for an ATDM Trilinos configuration other than the 'ats2' configuration that sets `TPL_Netcdf_LIBRARIES` that executed that line of code in the file `ATDMDevEnvSettings.cmake`, the code added to in the `FindTPLNetcdf.cmake` file would have not effect as explained in the comment below.

Therefore, I would suggest you discuss issues with Netcdf configuration with @gsjaardema and how it interacts with this custom [FindNetCDF.cmake](https://github.com/dc-snl/Trilinos/blob/develop/cmake/tribits/common_tpls/find_modules/FindNetCDF.cmake) that he maintains for SEACAS.

But if this is for some other ATDM Trilinos configuration other than the existing supported 'ats2' configuration that is trying to directly set `TPL_Netcdf_LIBRARIES` , then that is a different story.

But in that case, this code still will not work as described in comments below.

dc-snl:
> > we've been banging our heads against the wall trying to figure out why this change didn't work
> 
> @jmgate, what do you mean by "work"? What behavior are you seeing and what behavior are you expecting?
In short, we would like to link against curl only when `nc-config --has-dap4` or `nc-config --has-dap2` prints `yes`; otherwise, when they both print `no`, we don't want to link against curl. 

> 
> I am confused as to what use case on 'vortex' this PR is trying to address. Is this for the current supported ATDM Trilinos 'ats2' configuration or a new custom ATDM Trilinos configuration that is being developed for 'vortex'? If it is for the former, this PR should not change that 'ats2' configuration in any way (I will explain in my review that I will post).

Sounds good. In short, the change will fix an issue where the compiler (either CUDA or GCC) fails to find curl. Since none of the NetCDF modules seem to even have DAP enabled, we shouldn't need curl. This means that a simple fix (in theory) should be to not link against curl when it isn't needed.

> 
> > At this point, I'm wondering if something is going sideways within TriBITS
> 
> There are no defects in TriBITS. It behaves as coded. There is only defects in TriBITS documentation or violations of common sense :-)
> 
> Related to my epic SEPW-123

Thanks,
--David

cmake/std/atdm/ATDMDevEnvSettings.cmake
dc-snl: Ah, I think EMPIRE explicitly unsets `ATDM_USE_SPARC_TPL_FIND_SETTINGS` after sourcing the ATDM env script.

bartlettroscoe:
>  In short, the change will fix an issue where the compiler (either CUDA or GCC) fails to find curl. Since none of the NetCDF modules seem to even have DAP enabled, we shouldn't need curl. This means that a simple fix (in theory) should be to not link against curl when it isn't needed.

@dc-snl, just curious, but what how did you verify that these changes fixed things on 'vortex'?  What ATDM Trilinos configuration are you using that these changes 'fixed'?  What are the reproducibility instructions?

dc-snl:
> > In short, the change will fix an issue where the compiler (either CUDA or GCC) fails to find curl. Since none of the NetCDF modules seem to even have DAP enabled, we shouldn't need curl. This means that a simple fix (in theory) should be to not link against curl when it isn't needed.
> 
> @dc-snl, just curious, but what how did you verify that these changes fixed things on 'vortex'? What ATDM Trilinos configuration are you using that these changes 'fixed'? What are the reproducibility instructions?

Good question. I mistakenly assumed that, if the `-lcurl` linker error went away, the problem was fixed. What's odd is that the error did go away but the CMake cache shows that `-lcurl` still makes it into TPL_Netcdf_LIBRARIES.

The configuration would be `cuda-debug-Volta70-static-rdc` (based on my understanding of the nomenclature). The only variation is that we unset `ATDM_USE_SPARC_TPL_FIND_SETTINGS`

The steps to reproduce should be something like this:
```
source /path/to/Trilinos/cmake/std/atdm/load-⁠env.sh cuda-⁠debug-⁠rdc
unset ATDM_CONFIG_ENABLE_SPARC_SETTINGS
cmake \
    -⁠D Trilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake,cmake/std/atdm/apps/empire/EMPIRETrilinosEnables.cmake \
    -⁠D TPL_ENABLE_CUSPARSE:BOOL=OFF \
    -⁠G Ninja \
    /path/to/Trilinos
ninja
```

It's been a while since I've looked at this. I'll need to verify that this actually reproduces the problem. By the time I get back to vortex tomorrow, my reproducer should be finished running.

cmake/std/atdm/ATDMDevEnvSettings.cmake
bartlettroscoe: @dc-snl, changing the value of `ATDM_USE_SPARC_TPL_FIND_SETTINGS` and tweaking other options makes this no longer an ATDM Trilinos 'ats2' configuration.  What that means is that the automated builds of the 'ats2' configuration being posted daily to [the Trilinos CDash site](https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercount=1&showfilters=1&field1=buildname&compare1=65&value1=Trilinos-atdm-ats2_) no longer are testing the configurations that are being used by customer APPs.  If each APP team is going to have its own custom Trilinos configuration tweaks, then we are just going back to the days where each APP maintains their own Trilinos configuration.  If that is the case, then these automated builds should be taken down and the APP teams should go back to maintaining their own configurations (which is where this seems to be going anyway).

jjellio:
There are inconsistencies in the ATDMDevSettings.cmake.  NetCDF is one of them. (boost is another)

From my experience, NetCDF doesn't respect libraries if the user passes them in. (Ross points this out in code comment above). Atleast in my builds, it always chooses to call CMake's 'find' for NetCDF (because I have the sparc setting enabled). That should be resolving this reported issue - the gripe is that CMake's find assumes it can execute the binaries (netcdf or hdf5 tools) - that usually breaks on cross compile systems. Another inconsistency with CMake's find, is that it typically prefers shared over static - you can control that, but if the game is that ATDMDEvSettings is the gate keeper for consistent builds, you really don't want to add cruft that dodges that.


IMO, we should refactor ATDMDevSettings, and perhaps enforce a consistent syntax for the core libraries. E.g., if the user sets `ATDM_SETTINGS_FOO_LIBRARIES`, then always respect that (and set TPL_FOO_LIBRARIES).  Similar semantics for INCLUDE. Then have something very clear, `ATDM_SETTINGS_FOO_USE_FIND` which supports the optional `ATDM_SETTINGS_FOO_LIBRARY_NAMES`.  Define some precedence, e.g., 1.  `ATDM_SETTINGS_FOO_LIBRARIES`, 2. `ATDM_SETTINGS_FOO_LIBRARY_NAMES` (implies find), 3. `ATDM_SETTINGS_FOO_USE_FIND`- doesn't require (2).

I am more than happy to work in the direction of the prior paragraph. 

It's also worth reinforcing - Apps should use the `ATDMDevSettings` + `environment.sh` - I am working with some vendors to have Trilinos (and core apps) working, and we are using the DevSettings method for consistent builds.


TLDR: I wrote this mostly as an aside... I seem to build ATDM trilinos + EMPIRE/SPARC on machines before they stabilize to where CI happens. Having consistent semantics in DevSettings would make working with SPACK/pkg-config easier (where we can hand off lists of libraries)... but only when you know you've got them correct!

bartlettroscoe:
> NetCDF doesn't respect libraries if the user passes them in.

@jjellio, I believe that if the user sets `TPL_Netcdf_LIBRARIES`, the those libraries get used without modification.  That is a TriBITS thing and has nothing to do with NetCDF in particular.

> always chooses to call CMake's 'find' for NetCDF 

NOTE, it is not really "CMake's find", but instead it uses the custom FindNetcdf.cmake module that @gsjaardema maintains for SEACAS.

> It's also worth reinforcing - Apps should use the ATDMDevSettings + environment.sh

Yes, any changes to that is asking for trouble.  All it takes on one small issue with a "tweak" and you can eat up hours (even days) of wasted time tracking down problems.

bartlettroscoe:
> Ross, I am referring directly to setting `ATDM_CONFIG_NETCDF_LIBRARIES="/.../bla.s;/a../.../as.a...."`.  Currently, if SPARC settings are on, cmake will `find` netcdf. 

@jjellio, there are always going to be situations where setting one var will cause one or more other vars to get ignored.  That is a problem with CMake in general actually (and CMake will not warn you about if the cache vars are set internally instead of on the command-line).  We should likely put in warnings for these cases so this type of behavior is not too surprising.  (But it will take a bit of work to identify these cases and generate good warning messages.)

bartlettroscoe:
@dc-snl, @jmgate, @e10harvey 

FYI: To help explain this interaction between `FindTPLNetcdf.cmake` and `FindNetCDF.cmake`, see:

* https://tribits.org/doc/TribitsDevelopersGuide.html#how-to-use-find-package-for-a-tribits-tpl

NOTE: To maintain needed backward compatibility this behavior will **not** change with the TriBITS refactoring TriBITSPub/TriBITS#299.  To do so would break a lot of CMake configure scripts that people have.

PR #8852, opened 2021-03-08
Intrepid2: UVM-free MonolothicExecutable Tests

CamelliaDPG:
@trilinos/intrepid2 

## Motivation
This PR adds DeviceType compatibility to the newer Intrepid2 container types, including:
- BasisValues
- CellGeometry
- Data
- ProjectedGeometry
- TensorData
- TensorPoints
- TransformedVectorData
- VectorData

In addition, the new DeviceType compatibility is exercised in all the tests in MonolothicExecutable, replacing earlier use of `Kokkos::DefaultExecutionSpace` with the `DefaultTestDeviceType` defined in `Intrepid2_TestUtils.hpp`, which resolves to a UVM-free DeviceType on CUDA, and the Kokkos default DeviceType on other platforms.  Thus, the tests in MonolothicExecutable now run UVM-free on CUDA.

Along the way, a couple other notable changes were necessary:
- Added a `DeviceType` typedef to the Basis base class.  This is defined as whatever the Basis was instantiated with (so it may be a non-device-type ExecutionSpace); this is useful when setting the types for BasisValues, etc., in derived classes.
- Because some CellTools is not yet fully compatible with DeviceType, but the tests in MonolithicExecutable are using DeviceType-templated containers for methods in CellTools, we need to allow CellTools to be templated on a DeviceType; to avoid cascading to all of CellTools, we add an ExecSpaceType typedef defined in terms of the DeviceType; this is used for all types in CellTools methods besides the ones required by the present PR.  This is inelegant, and a temporary solution until we complete the CellTools conversion.  I can confirm that it does build and pass tests under CUDA on vortex.

## Related Issues
- Part of #8310.

## Testing
This PR includes extensive tests.  Additionally, tests have been manually verified to pass on vortex (CUDA release build) and on Intel serial debug (under MacOS).

### DIFF
--- a/packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
+++ b/packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
@@ -76,7 +76,7 @@ namespace Intrepid2
       
    \note Conceptually, it would make sense to use class inheritance and have different member data for each type of geometry supported.  We instead glom all the options together into one multi-modal class; this is basically to avoid certain difficulties with vtables under CUDA.
    */
-  template<class PointScalar, int spaceDim, typename ExecSpaceType = Kokkos::DefaultExecutionSpace>
+  template<class PointScalar, int spaceDim, typename DeviceType = Kokkos::DefaultExecutionSpace>
   class CellGeometry
   {
     public:
@@ -122,7 +122,7 @@ namespace Intrepid2
        \param [in] endCell - the first cell ordinal for which the Jacobian will not be evaluated (used to define worksets smaller than the whole CellGeometry); use -1 to indicate that all cells from startCell on should be evaluated.
        \return a Data object appropriately sized to accomodate the specified Jacobian values.
     */
-    Data<PointScalar,ExecSpaceType> allocateJacobianDataPrivate(const TensorPoints<PointScalar,ExecSpaceType> &points, const int &pointsPerCell, const int startCell, const int endCell) const;
+    Data<PointScalar,DeviceType> allocateJacobianDataPrivate(const TensorPoints<PointScalar,DeviceType> &points, const int &pointsPerCell, const int startCell, const int endCell) const;
     
     /** \brief  Notionally-private method that provides a common interface for multiple public-facing setJacobianData() methods.  (Marked as public due to compiler constraints.)
        \param [out] jacobianData - a container, allocated by allocateJacobianData(), into which the evaluated Jacobians will be placed.
@@ -132,14 +132,14 @@ namespace Intrepid2
        \param [in] startCell - the first cell ordinal for which the Jacobian will be evaluated (used to define worksets smaller than the whole CellGeometry).
        \param [in] endCell - the first cell ordinal for which the Jacobian will not be evaluated (used to define worksets smaller than the whole CellGeometry); use -1 to indicate that all cells from startCell on should be evaluated.
     */
-    void setJacobianDataPrivate(Data<PointScalar,ExecSpaceType> &jacobianData, const TensorPoints<PointScalar,ExecSpaceType> &points, const int &pointsPerCell,
-                                const Data<PointScalar,ExecSpaceType> &refData, const int startCell, const int endCell) const;
+    void setJacobianDataPrivate(Data<PointScalar,DeviceType> &jacobianData, const TensorPoints<PointScalar,DeviceType> &points, const int &pointsPerCell,
+                                const Data<PointScalar,DeviceType> &refData, const int startCell, const int endCell) const;
   protected:
     HypercubeNodeOrdering nodeOrdering_;
     CellGeometryType    cellGeometryType_;
     SubdivisionStrategy subdivisionStrategy_ = NO_SUBDIVISION;
     bool affine_; // if true, each cell has constant Jacobian across the cell
-    Data<Orientation, ExecSpaceType> orientations_; // for grid types, this could have either a single entry or one matching numCellsPerGridCell().  For other types, it has as many entries as there are cells.
+    Data<Orientation, DeviceType> orientations_; // for grid types, this could have either a single entry or one matching numCellsPerGridCell().  For other types, it has as many entries as there are cells.
     
     // uniform grid data -- used for UNIFORM_GRID type
     Kokkos::Array<PointScalar,spaceDim> origin_;         // point specifying a corner of the mesh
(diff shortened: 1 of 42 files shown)


packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
```diff
@@ -76,7 +76,7 @@ namespace Intrepid2
       
    \note Conceptually, it would make sense to use class inheritance and have different member data for each type of geometry supported.  We instead glom all the options together into one multi-modal class; this is basically to avoid certain difficulties with vtables under CUDA.
    */
-  template<class PointScalar, int spaceDim, typename ExecSpaceType = Kokkos::DefaultExecutionSpace>
+  template<class PointScalar, int spaceDim, typename DeviceType = Kokkos::DefaultExecutionSpace>
```
kyungjoo-kim, line 79: I prefer not to default the device type as we do not want any accidental use of UVM when the default space is CUDA.

packages/intrepid2/src/Cell/Intrepid2_CellTools.hpp
```diff
@@ -102,8 +102,9 @@ namespace Intrepid2 {
   */
 
 
-  template<typename ExecSpaceType>
+  template<typename DeviceType>
```
kyungjoo-kim, line 105: Let's not change this if we do not fix the core part. This ``CellTools`` are not tested with actual ``Device<Cuda,CudaSpace>``.

packages/intrepid2/src/Shared/Intrepid2_Data.hpp
```diff
@@ -168,20 +168,20 @@ namespace Intrepid2 {
...
-    Kokkos::View<DataScalar*****,   ExecSpaceType> data5_;  // the rank 5 data that is explicitly stored
-    Kokkos::View<DataScalar******,  ExecSpaceType> data6_;  // the rank 6 data that is explicitly stored
-    Kokkos::View<DataScalar*******, ExecSpaceType> data7_;  // the rank 7 data that is explicitly stored
+    Kokkos::View<DataScalar*,       DeviceType> data1_;  // the rank 1 data that is explicitly stored
+    Kokkos::View<DataScalar**,      DeviceType> data2_;  // the rank 2 data that is explicitly stored
+    Kokkos::View<DataScalar***,     DeviceType> data3_;  // the rank 3 data that is explicitly stored
+    Kokkos::View<DataScalar****,    DeviceType> data4_;  // the rank 4 data that is explicitly stored
+    Kokkos::View<DataScalar*****,   DeviceType> data5_;  // the rank 5 data that is explicitly stored
+    Kokkos::View<DataScalar******,  DeviceType> data6_;  // the rank 6 data that is explicitly stored
+    Kokkos::View<DataScalar*******, DeviceType> data7_;  // the rank 7 data that is explicitly stored
```
kyungjoo-kim, line 184: This is not really related comment for this PR. Why do you use this hand version of  DynRankView ?

packages/intrepid2/src/Shared/Intrepid2_TransformedVectorData.hpp
```diff
@@ -60,22 +60,22 @@ namespace Intrepid2 {
  
  TransformedVectorData provides a View-like interface of rank 4, with shape (C,F,P,D).  When the corresponding accessor is used, the transformed value is determined from corresponding reference space values and the transformation.
 */
-  template<class Scalar, typename ExecSpaceType = Kokkos::DefaultExecutionSpace>
+  template<class Scalar, typename DeviceType = Kokkos::DefaultExecutionSpace>
```
kyungjoo-kim, line 63: I do not think that defaulting the device is not a good idea.

packages/intrepid2/src/Shared/Intrepid2_VectorData.hpp
```diff
@@ -61,11 +61,11 @@ namespace Intrepid2 {
  
  Typically, HDIV and HCURL bases have multiple families corresponding to the nonzero structure of the vector components.  VectorData therefore supports multiple families.  (Gradients of H^1 are expressed as a single family.)
 */
-  template<class Scalar, typename ExecSpaceType = Kokkos::DefaultExecutionSpace>
+  template<class Scalar, typename DeviceType = Kokkos::DefaultExecutionSpace>
```
kyungjoo-kim, line 64: same here

kyungjoo-kim (commented):
For now, CellTools need to be instanciated with execution space with UVM. Otherwise, it does not work. Can you improve this PR more surgically ? I am not sure if this is possible. I just do not like that cell tools conversion is incomplete for now. Overall, I looks fine.

kyungjoo-kim:
Okay I read your introduction in this PR. I agree that cell tools is a big guy. As long as your test can pass with CudaSpace device type, I am okay. I still like not to use default value for device type.

CamelliaDPG:
> Okay I read your introduction in this PR. I agree that cell tools is a big guy. As long as your test can pass with CudaSpace device type, I am okay. 

Sounds good; thanks.

> I still like not to use default value for device type.

I know that's your preference, and I understand the argument you've made about this in the past: by forcing specification, we can ensure that we don't have hidden/accidental use of DefaultExecutionSpace somewhere in the test code.  My counterargument is that 90% of the time (if not more), users will want the DefaultExecutionSpace, and it's nicer for them if they don't have to type it, and clutter up their code with the specification.  And so long as we are somewhat diligent in using the DeviceTypes in our tests, my recent experience is that the compiler will catch accidental usage of DefaultExecutionSpace, in the form of complaints about incompatible View assignment.

packages/intrepid2/src/Shared/Intrepid2_Data.hpp
CamelliaDPG: This allows us to use View instead of DynRankView in places where the performance difference might matter.  I instituted this while optimizing for CUDA performance in the sum factorization work.  I don't have hard numbers to show that this matters, but at the time anyway I suspected that it did matter.

mperego (dismissed):
It looks good. As discussed offline we can defer the removal of default template for DeviceType.

CamelliaDPG:
FYI: gcc failures just now due to an unused variable warning, just fixed.  @mperego, when you get a chance, would you please re-approve?

CamelliaDPG:
Now it's the Intel compiler that's complaining (incorrectly) about an unused variable.  I've eliminated the variable in my latest commit; that should get us past the warning.

mperego:
@CamelliaDPG warnings are a pain. Have you addressed this?

> In static member function ‘static void Intrepid2::CellTools<DeviceType>::setJacobian(Kokkos::DynRankView<DT, DP ...>, WorksetType, BasisGradientsType, int, int)’: /scratch/trilinos/jenkins/ascic141/workspace/trilinos-folder/Trilinos_pullrequest_gcc_8.3.0/Trilinos/packages/intrepid2Cell/Intrepid2_CellToolsDefJacobian.hpp:798:66: error: declaration of ‘using ExecSpaceType = typename DeviceType::execution_space’ shadows a previous local [-Werror=shadow]      using ExecSpaceType    = typename DeviceType::execution_space;

CamelliaDPG:
> @CamelliaDPG warnings are a pain. Have you addressed this?
> 
> > In static member function ‘static void Intrepid2::CellTools::setJacobian(Kokkos::DynRankView<DT, DP ...>, WorksetType, BasisGradientsType, int, int)’: /scratch/trilinos/jenkins/ascic141/workspace/trilinos-folder/Trilinos_pullrequest_gcc_8.3.0/Trilinos/packages/intrepid2Cell/Intrepid2_CellToolsDefJacobian.hpp:798:66: error: declaration of ‘using ExecSpaceType = typename DeviceType::execution_space’ shadows a previous local [-Werror=shadow]      using ExecSpaceType    = typename DeviceType::execution_space;

@mperego, I hadn't noticed that one.  Should be fixed in what I pushed just now.

CamelliaDPG:
A handful further warnings from Intel, gcc.  Hopefully fixed now.

mperego (approved):

PR #8855, opened 2021-03-09
Belos+MueLu: reuse stuff

cgcgcg:
@trilinos/belos @trilinos/muelu 

## Motivation
- Save state vectors in Belos CG solves for later reuse
- RefMaxwell: label allocated multivectors


lucbv (dismissed):
The MueLu changes are harmless enough so fine with me.
Maybe you can get Ichi to look at the Belos side?

cgcgcg:
@lucbv Let's see how much the AT complains first ;-)

lucbv:
Honestly I don't think the MueLu side would lead to any problems but if it does I'll definitely check to see why labels can be so finicky :p

hkthorn:
@cgcgcg I can review the changes since I'm knowledgeable about this code / issue.  Thx!

cgcgcg:
@hkthorn I might need to track down a couple more MVs that are getting reallocated for every new solve.

hkthorn:
@cgcgcg There is a temporary vector that is created if right preconditioning is used.  You may want to allocate a work vector and put it in the state data structure.  Also, we need to make sure that this approach works with the adaptive block size infrastructure, but I think that is tested by the Epetra tests.

hkthorn (commented):
I can see that is can be removed safely in setLSIndex because it is set later in the same method.  However, I'm not sure why it should be removed from setCurrLS, as that changes the state of the solver to address convergence or deflation.

cgcgcg:
@hkthorn Ugh, yeah, that's possible. Let's see how miserable the AT is. I commented it out, since it actually triggered a mysterious deallocation of a MV.

hkthorn:
I wouldn't use the AT to determine if this change is ok.  In most cases, the curX_ and curB_ are clone views of the X_ and B_ multi-vectors.  Are you looking to eliminate clones that are views too?

PR #8859, opened 2021-03-09
KokkosKernels: std::

searhein:
@trilinos/kokkos-kernels 

This PR just fixed a very small bug. I just changed `std:cout` to `std::cout` in two files.

### DIFF
--- a/packages/kokkos-kernels/src/graph/KokkosGraph_Distance1ColorHandle.hpp
+++ b/packages/kokkos-kernels/src/graph/KokkosGraph_Distance1ColorHandle.hpp
@@ -240,7 +240,7 @@ private:
     {
       this->coloring_algorithm_type = COLORING_SERIAL;
 #ifdef VERBOSE
-      std:cout << "Serial Execution Space, Default Algorithm: COLORING_SERIAL\n";
+      std::cout << "Serial Execution Space, Default Algorithm: COLORING_SERIAL\n";
 #endif
     }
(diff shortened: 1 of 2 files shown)


srajama1 (approved):
@searhein Thank you ! I assume you are enabling verbose.

@brian-kelley This might need a matching in KK git repo for D2 coloring.

searhein:
@srajama1 The problem actually occurred when compiling our finite element code, which links to Trilinos. The reason why this occurred is is that the macro name `VERBOSE` is very generic, and therefore, it was also defined somewhere in our code.

Maybe, you also want to change it to `KOKKOS_VERBOSE`?

brian-kelley:
@srajama1 Yes, kokkos-kernels develop also needs these changes, in both D1 and D2 coloring. I like the idea of using a more specific macro for this than "VERBOSE", however I don't want to step on Kokkos core by using "KOKKOS_VERBOSE", and we can't use "KOKKOSKERNELS_VERBOSE" because that is a runtime value in the SpGEMM handle.

I want to do this in a different PR, but we should unify all the 'verbose' branches under a runtime option in the KokkosKernels handle. That option already exists, it can be accessed via ``set_verbose()`` and ``get_verbose()``, but only SpGEMM is using it now. Everything else uses ``#ifdef VERBOSE``.

I copied the fixes here into https://github.com/kokkos/kokkos-kernels/pull/910 .

Issue #8846, opened 2021-03-07
MueLu: missing parameter in aggregation


mayrmt:
## Bug Report
@trilinos/muelu 

### Description

When running `ctest -VV` on a current version of `develop` (0f69cc4a68000d6e4003faf00887e23791f4de06), this error occurs in `MueLu_Structured_Unstructured_Region_Example_Poisson3D_27pt_stencil_MPI_4`:

```
33: Test command: /usr/bin/mpiexec "--oversubscribe" "-np" "4" "/hdd/codes/mayrmt_trilinos/build_muelu/packages/muelu/research/regionMG/example/MueLu_RegionExampleDriver.exe" "--linAlgebra=Tpetra" "--xml=structured_unstructured_1dof.xml" "--matrixType=Brick3D" "--nx=28" "--ny=28" "--nz=28" "--smootherIts=2" "--unstructured={1,2}"
33: Test timeout computed to be: 1500
33: ========================================================
33: Xpetra::Parameters
33:  Linear algebra library: Tpetra
33: Galeri::Xpetra::Parameters<long>
33:  Matrix type: Brick3D
33:  Problem size: 21952
33: Processor subdomains in x direction: 2
33: Processor subdomains in y direction: 2
33: Processor subdomains in z direction: 1
33: ========================================================
33: Galeri complete.
33: ========================================================
33: nodesPerDim: {15, 14, 28}
33: nodesPerDim: {14, 15, 28}
33: Clearing old data (if any)
33: 
33: ******* WARNING *******
33: Hierarchy::ReplaceCoordinateMap: matrix and coordinates maps are same, skipping...
33: MueLu::Amesos2Smoother: using "Klu"
33: Using default factory (SmootherFactory[17] ) for building 'CoarseSolver'.
33: Using default factory (AmalgamationFactory[20] ) for building 'UnAmalgamationInfo'.
33: Using default factory (CoarseMapFactory[21] ) for building 'CoarseMap'.
33: Level 0
33: Level 1
33: Build (MueLu::TentativePFactory)
33: Build (MueLu::HybridAggregationFactory)
33: Build (MueLu::CoalesceDropFactory)
33: Build (MueLu::AmalgamationFactory)
33: AmalagamationFactory::Build(): found fullblocksize=1 and stridedblocksize=1 from strided maps. offset=0
33: lightweight wrap = 1
33: algorithm = "classical" classical algorithm = "default": threshold = 0, blocksize = 1
33: 
[...]

PR #8860, opened 2021-03-09
MueLu: Add phase2a agg factor to param list for hybrid aggregation

pohm01:
@trilinos/muelu

## Motivation
Adds missing aggregation parameter to parameter list for Hybrid aggregation.
Solves #8846 

## Testing
Tests pass on my local machine.


mayrmt (approved):
@pohm01 Thanks for this fix! Looks good to me.

lucbv (approved):
Thanks for looking into this Peter

Issue #8498, opened 2020-12-17
Dropping support for Makefile.export.* files with TriBITS/Trilinos refactoring to support modern CMake features


bartlettroscoe:
CC: @trilinos/framework, @trilinos/trilinos-product-owners 

**Blocked By:** #8401 

As part of the imminent TriBITS refactoring described in TriBITSPub/TriBITS#299 to use modern CMake target-based features will that will be enabled by the upgrade of the minimum required CMake (see #8401) be that the current implementation for generating export makefiles (i.e.  [`Makefile.export.*` files](https://docs.trilinos.org/files/TrilinosBuildReference.html#generating-export-files)) will be broken and it may be difficult to bring it back.

We discussed this topic and the topic of dropping support for `Makefile.export.*` files at the Trilinos Product Leaders Meeting on 12/16/2020.  The action items that came out of that were:

* [x] Add an example raw CMake project in `Trilinos/demos/` that uses `find_package(Trilinos)` and uses some functionality in Trilinos ... **See [demos/simpleBuildAgainstTrilinos](https://github.com/trilinos/Trilinos/tree/develop/demos/simpleBuildAgainstTrilinos)**
* [x] Create a plan for customers to transition to no `Makefile.export.Trilinos` files. (I.e., either switch to CMake or manually get the list of libraries you need such as with `CMAKE_EXPORT_COMPILE_COMMANDS`) ... **Plan is for customers to use CMake or manually maintain their link lines (like Sierra does)**
* [x] Craft and send out an email to various mail lists announcing dropping support for `Makefile.export.Trilinos` and get feedback from internal and external customers ... **Sent out emails on 5/19/2021 stating support will be dropped on 6/10/2021**
* [x] Send out reminder email about dropping support for `Makefile.export.Trilinos` on 6/10/2021.
* [x] Update documentation on [trilinos.github.io related to building against Trilinos](https://trilinos.github.io/documentation.html) (update [Find_Trilinos.txt](https://github.com/trilinos/trilinos.github.io/blob/master/pdfs/Finding_Trilinos.txt) and remove [Export_Makefile.txt](https://trilinos.github.io/pdfs/Export_Makefile.txt)) ... **See PR https://github.com/trilinos/trilinos.github.io/pull/49 ...**
* [x] Drop support for `Makefile.export.Trilinos` from TriBITS and Trilinos 'develop' on 6/10/2021 ... **See PR #9284**
[...]

PR #8867, opened 2021-03-10
Update instructions/examples for building against installed Trilinos and update related doc files (#8498)

bartlettroscoe:
This addresses the first task of #8498.

This add the example demos/simpleBuildAgainstTrilinos and provides a link to it in the INSTALL.rst.  I also updated the top-level README.md to point to the correct local INSTALL.rst file.  I updated the INSTALL.rst file some too.

For details, please review the commit messages for this PR and patches in those commits.

To view the rendered updated files, see:

* [Trilinos/README.md](https://github.com/bartlettroscoe/Trilinos/blob/8498-update-buildAgainstTrilinos/README.md)
* [Trilinos/INSTALL.rst](https://github.com/bartlettroscoe/Trilinos/blob/8498-update-buildAgainstTrilinos/INSTALL.rst)
* [Trilinos/demos/simpleBuildAgainstTrilinos/README.md](https://github.com/bartlettroscoe/Trilinos/blob/8498-update-buildAgainstTrilinos/demos/simpleBuildAgainstTrilinos/README.md)

Try the links between these various documents to make sure they work.

## How was this tested?

I followed the updated instructions against an ATDM Trilinos install using clang 10.0 with the sems-rhel7 env.  It was very easy (after the build error was fixed)


rppawlo:
This is a much more complex example than I was hoping for. Source and examples are split across multiple directories and the key info is not all in a single CMakeLists.txt file. We should probably keep this as a more complex example. Maybe we could add a simpler example that compacts this down into a single main.cpp file that uses teuchos and a single CMakeLists.txt file with no cmake function/macro definitions. Something with the very minimum to get a new user started. Maybe the only complexity in handling whether mpi is enabled. Thoughts?

bartlettroscoe:
> Thoughts?

This was originally Andy's example.  Should we keep this or gut it and make it a single `*.cmake` and `CMakeLists.txt` file?  Or should we create a new simpler example?  Does not matter to me either way.

rppawlo:
I would lean toward dropping the more complex example. It uses epetra which we would like to discourage since all effort is going into tpetra. One thing we should keep is the printing of all the available trilinos variables - users then don't have to dig into the config file.

bartlettroscoe:
>  It uses epetra which we would like to discourage since all effort is going into tpetra.

@rppawlo, seems like it would be good to have the example link against Tpetra because that also requires Kokkos and makes for a more non-trivial link line.   But we would take out the conditional logic of using Tpetra.  What about that?

bartlettroscoe:
> Is there a reason to use tpetra either? 

@rppawlo, for a more complex link line that suggests the utility of `find_package()`.  The Tpetra code is pretty minimal.  Just create a map then a vector and print the norm1().  That is a simple way to ensure that MPI is actually working and makes this a simple smoke test for installations of Trilinos as well.

What do other people think?

rppawlo:
I see. Sounds good!

bartlettroscoe:
> I see. Sounds good!

Okay, I will post a commit that adds `demos/simpleBuildAgainstTrilinos` and then remove the `demos/buildAgainstTrilinos` example in a follow-up commit (and fix all of the links).

bartlettroscoe:
@rppawlo, can you please take a look at the new `demos/simpleBuildAgainstTrilinos` example I just added in commit https://github.com/trilinos/Trilinos/pull/8867/commits/29661d4f5e17ce09ee2cc1326db59a2ea6a4a484.

I will update the links to this example next and then we can consider deleting the old  `demos/buildAgainstTrilinos` example if that is the desire.  (We can also replace Epetra with Tpetra in that example and modernize it some if people see some value in that.)

bartlettroscoe:
FYI: This PR branch is now updated with the new [demos/simpleBuildAgainstTrilinos](https://github.com/bartlettroscoe/Trilinos/blob/8498-update-buildAgainstTrilinos/demos/simpleBuildAgainstTrilinos) example and I removed the old `demos/buildAgainstTrilinos`.  (We can bring back and update that more complex example if desired.)

This PR/branch is ready for more review.

rppawlo (approved):

bartlettroscoe:
I will go ahead an merge this.  If improvements are needed/desired, then we can do that in a follow-up PR.

Issue #8695, opened 2021-02-04
ATDM:  Disabled Package for EMPIRE Still On


jmgate:
Moving the conversation from email to here:

From @jmgate:

> @e10harvey, do you have any idea why builds on vortex would be ignoring [this change to Trilinos](https://github.com/trilinos/Trilinos/pull/8493) we pushed a week or so back to disable Tempus for EMPIRE.  Tempus looks to be disabled in other builds, e.g., [Deploy-develop-ascic-clang-opt-serial-static](https://do.sandia.gov/empire-jenkins/view/Deploy/job/Deploy-develop-ascic-clang-opt-serial-static/74/consoleFull) shows
>  
> 00:00:18.440  **Explicitly disabled packages** on input (by user or by default):  TrilinosFrameworkTests Gtest MiniTensor Domi Isorropia Pliris ShyLU_Node Komplex FEI TriKota Compadre Moertel ShyLU_DD ShyLU **Tempus** Stokhos ROL PyTrilinos NewPackage TrilinosCouplings Pike 21
>  
> But on vortex ([Deploy-develop-vortex-gnu-opt-openmp-static](https://do.sandia.gov/empire-jenkins/view/Deploy/job/Deploy-develop-vortex-gnu-opt-openmp-static/62/consoleFull)) we’re seeing
>  
> 00:01:22.708  -- **Setting Trilinos_ENABLE_Tempus=ON** because Piro has an optional dependence on Tempus
>  
> Any idea what might be going on?  Completely wiping everything out and starting from scratch didn’t help.

Then the response from @e10harvey:

> @jmgate, thank you for the detailed email. We are looking into a service desk or ticketing system to address requests like this going forward. For the time being, a GitHub issue is preferred over email.
>  
> From [Deploy-develop-vortex-gnu-opt-openmp-static](https://do.sandia.gov/empire-jenkins/view/Deploy/job/Deploy-develop-vortex-gnu-opt-openmp-static/62/consoleFull):
> 16:11:54  Explicitly disabled packages on input (by user or by default):  TrilinosFrameworkTests Gtest MiniTensor Domi Isorropia Pliris Komplex FEI TriKota Compadre Moertel ShyLU_DD ShyLU Stokhos PyTrilinos NewPackage TrilinosCouplings Pike 18
>  
[...]

PR #8882, opened 2021-03-15
Enable Zoltan ULLONG_IDS for all ATDM apps

jmgate:
@trilinos/framework 

## Motivation
From Nicole Slattengren on the EMPIRE DevOps Mattermost channel:
> Did you change the Trilinos configuration in the last couple months to stop using `-D Zoltan_ENABLE_ULLONG_IDS:Bool=ON`?  I thought I was able to use vt's ZoltanLB from EMPIRE with the EMPIRE Trilinos build on Stria in mid-January but this time it's complaining.

In #8695, we determined we needed to `unset ATDM_ENABLE_SPARC_SETTINGS` to prevent the ATDM build scripts from turning Tempus on after we'd turned it off.  This `Zoltan_ENABLE_ULLONG_IDS` setting was part of the SPARC settings, so this PR pulls it out and sets it for any ATDM applications.


e10harvey:
@glhenni: Will this impact GEMMA?

glhenni:
> @glhenni: Will this impact GEMMA?

In theory it shouldn't, since it was getting set via the SPARC setting prior to this anyway. But as with most things Trilinos, we won't know until we try it.

e10harvey (approved):

PR #8884, opened 2021-03-15
Frosch: use paralllel-for in compute phase

iyamazaki:
@trilinos/shylu 

## Motivation
This PR aims to improve the GPU performance of FROSch's compute phase by using parallel-for in three of the loops.

## Testing

Ran ctest under shylu_dd/frosch/test on Vortex (gnu-release and cuda-release)


searhein (dismissed):
@iyamazaki Thanks very much for the changes! I didn't realize anything odd, and it seems to work on all the test.

searhein (approved):
@iyamazaki Thanks for the changes. All tests seemed to run smoothly, and the changes also look good to me.

PR #8885, opened 2021-03-15
Piro: use nominal parameter values by default in NOXSolver

kliegeois:
@trilinos/piro
@mperego 

## Motivation

Before this PR, the `Piro::NOXSolver` was the only child class of `Piro::SteadyStateSolver` which was not using the nominal parameter values by default.

## Testing

This branch has been tested on Blake successfully.


mperego:
@kliegeois This looks good to me. @rppawlo I don't think this can create issues with Drekar or Empire, but I wanted to double check with you.

mperego:
@rppawlo we are going to merge this by th end of the day if we don't hear any objection from you.

rppawlo:
@mperego - I won't be able to test this today but it looks fine to me. If it passed the panzer PR tests, it should not cause an issue in Drekar. Thanks for checking!

mperego (approved):

PR #8907, opened 2021-03-22
Adding unique identifier to partitioned worksets

seamill:
@trilinos/panzer 

## Motivation

Found bug relating to uniquely identifying worksets constructed using the partitioned workset path.


seamill:
@rppawlo Could you review this? I ran into this issue while adding the new BCs to empire.

PR #8908, opened 2021-03-22
Upgrade stale action version

william76:
@trilinos/framework 

## Motivation
Update the version of the auto-closer **action** script to v3.0.18 from v3.0.14.  This new version should include the fixes on the issue I filed with them about the tool failing to remove the "AUTOCLOSED" label when a user re-opens an issue that had been auto-closed.

I am also bumping the rate limit parameter from 50 to 55 here to see if we can bump up the # of issues processed in our monthly cycle. I'd like to periodically increase this to see if we can get up past 100 issues/mo that we can process.  Note: there's not a 1:1 correlation between this threshold and the actual number of issues processed.  For example, this month we have 86 that are in-flight using the completely arbitrary value of 50 for the app's internal rate-limiter.  

It would be nice if they could pin that somehow to a counter that limits to the # of ISSUES + PR's that are evaluated... or I'd be happy even with something like being able to tell it to "process up to 100 issues and 10 PR's". Perhaps I'll put in an issue asking for this sometime. 😁 

@jwillenbring @prwolfe


william76:
ugh... so it pulled in lots of commits here. I think I need to fix something with my branch. I'll fix that and redo this one.

PR #8917, opened 2021-03-24
Intrepid2: Work on CellTools to make it UVM free

mperego:
@trilinos/intrepid2

## Motivation

This PR contributes to making Intrepid2 not reliant on UVM. #8310 
Here we focus on the `CellTools` class. 
We introduce three new classes `RefSubcellParametrization`, `RefCellNodes` and `RefCellCenter` that stores static views containing the subcell parametrizations, the cell nodes coordinates and the cell center coordinates, for supported topologies.
These classes provide getters and setters, a typical use is:
```auto edgeParametrization = Intrepid2::RefSubcellParametrization<DeviceType>::get(1, topologyKey);```
The view `edgeParametrization` lives in `DeviceTyp::memory_space`.

The data stred in `RefSubcellParametrization` and , `RefCellNodes`  were previously stored in the  `CellTools` class, which has now been simplified. The class `Impl::CellTools` has been simplified as well, and it now provides only Kokkos inline funcitons.

The `CellTools` tests are can now run without UVM, using the `DeviceType`  `Kokkos::Device<Kokkos::Cuda,Kokkos::CudaSpace>`.

### CODE REVIEW

**Comment by mperego** [comment]
@kyungjoo-kim  I was thinking we can move the `Impl::CellTools::Serial` struct into the `CellToolls::Serial`, what do you think?
Also, now `CellTools` is complitely stateless, and since we are deducing the `DeviceType` from the input arguments of the member functions, I think we could remove the `DeviceType` template from the `CellTools` class.

**Comment by kyungjoo-kim** [comment]
> @kyungjoo-kim I was thinking we can move the `Impl::CellTools::Serial` struct into the `CellToolls::Serial`, what do you think?

I did not take look at details yet. It could be okay. The primary reason to hide the interface in Impl namespace is because we do not want to spread them out much and we want to have freedom to change the interface where expert-level users use the interface with their own risk. Without the impl namespace, it implies that we are going to provide some sort of backward compatibilities.   

> Also, now `CellTools` is complitely stateless, and since we are deducing the `DeviceType` from the input arguments of the member functions, I think we could remove the `DeviceType` template from the `CellTools` class.

I don't think that it is good idea to remove device template type from the tool class. You are right that we can make it state-less. However, which input argument should we follow if multiple input arguments have different device types ? Our interface allows this as individual view is templated. We just impose a constraint that it must be dynrank view. Otherwise, the runtime polymorphic behavior is not possible. The device type of the tool class defines where we store the local reference data (integration and cell topology data) and it indicates the primary execution space. So, if the input views have multiple memory space, we do deep copy to the device type of the tool class and execute it there. This decision is made to avoid the problems that we encounter a few years ago. Multiple memory and execution space possibiities increase the code complexity a lot e.g., for host, deep copy to host and execute on host; for device, deep copy to device and execute. This workflow does not consider the amount of actual data movements and computations. If the processing data size is large, the code may perform better on device with deep copy even if the data lives on host. Maybe the execution space that we choose from an input argument among others can be different and incurs excessive data movements of other inputs. 

CellTools is a bit exceptional as it has several routines that just computes for a single cell only. As now we store the reference data on both host and device, using relavent data based on the input/output view memory space is probably the right solution to avoid unnecessary communications. With these merits, I still think that we need to keep the device templates to simplify the logic for choosing an execution space.

**Comment by mperego** [comment]
> > @kyungjoo-kim I was thinking we can move the `Impl::CellTools::Serial` struct into the `CellToolls::Serial`, what do you think?
> 
> I did not take look at details yet. It could be okay. The primary reason to hide the interface in Impl namespace is because we do not want to spread them out much and we want to have freedom to change the interface where expert-level users use the interface with their own risk. Without the impl namespace, it implies that we are going to provide some sort of backward compatibilities.
> 
> > Also, now `CellTools` is complitely stateless, and since we are deducing the `DeviceType` from the input arguments of the member functions, I think we could remove the `DeviceType` template from the `CellTools` class.
> 
> I don't think that it is good idea to remove device template type from the tool class. You are right that we can make it state-less. However, which input argument should we follow if multiple input arguments have different device types ? Our interface allows this as individual view is templated. We just impose a constraint that it must be dynrank view. Otherwise, the runtime polymorphic behavior is not possible. The device type of the tool class defines where we store the local reference data (integration and cell topology data) and it indicates the primary execution space. So, if the input views have multiple memory space, we do deep copy to the device type of the tool class and execute it there. This decision is made to avoid the problems that we encounter a few years ago. Multiple memory and execution space possibiities increase the code complexity a lot e.g., for host, deep copy to host and execute on host; for device, deep copy to device and execute. This workflow does not consider the amount of actual data movements and computations. If the processing data size is large, the code may perform better on device with deep copy even if the data lives on host. Maybe the execution space that we choose from an input argument among others can be different and incurs excessive data movements of other inputs.
> 
> CellTools is a bit exceptional as it has several routines that just computes for a single cell only. As now we store the reference data on both host and device, using relavent data based on the input/output view memory space is probably the right solution to avoid unnecessary communications. With these merits, I still think that we need to keep the device templates to simplify the logic for choosing an execution space.

@kyungjoo-kim I see your point. I have worked under the assumpition that when there are mutiple input views, they use compatible memory spaces, so that I coud deduce the device_type from any of them (often the output view) and use it in the function. It can be hard to guarantee that the functions work with incompatible views, when not using UVM. Anyway, have a look at what I have done and then we will discuss more about this. I'm happy to revert some of the changes, and we should state in the function description any assumption we are making.

**Comment by mperego** [comment]
@kyungjoo-kim @CamelliaDPG  b.t.w. the reason why we cannot Kokkos inline the getters of the subcell parametrizations and cell nodes, is that a static view cannot be called by device. So, I think the best approach is to first get the view on host, e.g. 
`auto edgeParametrization = Intrepid2::RefSubcellParametrization<DeviceType>::get(edgeId, topologyKey);`
and then use `edgeParametrization` , which is not static, in  a kernel. This is the approach we now use in `Imp::Celltools::Serial` functions to compute tangents and normals. These functions have in input the edge or face  parametrizations (instead of the cell topology).

**Comment by CamelliaDPG** [comment]
> @kyungjoo-kim @CamelliaDPG b.t.w. the reason why we cannot Kokkos inline the getters of the subcell parametrizations and cell nodes, is that a static view cannot be called by device. So, I think the best approach is to first get the view on host, e.g.
> `auto edgeParametrization = Intrepid2::RefSubcellParametrization<DeviceType>::get(1, topologyKey);`
> and then use `edgeParametrization` , which is not static, in a kernel. This is the approach we now use in `Imp::Celltools::Serial` functions to compute tangents and normals. These functions have in input the edge or face parametrizations (instead of the cell topology).

@mperego, that makes sense, and the pattern you're following is a good one.  Thanks for explaining!

**Comment by mperego** [comment]
Another design question. At the moment, the fisrt time a getter is called, the parametrizations of all topologies are set. This is probably OK because it is done only once. Another approach would be to set each parametrization separately, the first time it is called.

**Comment by CamelliaDPG** [comment]
> Another design question. At the moment, the fisrt time a getter is called, the parametrizations of all topologies are set. This is probably OK because it is done only once. Another approach would be to set each parametrization separately, the first time it is called.

I think probably either way is fine; there aren't that many topologies, and the parametrizations aren't that expensive (in memory or space).

File: packages/intrepid2/src/Cell/Intrepid2_CellToolsDefControlVolume.hpp [hunk 1]
```diff
@@ -386,6 +386,7 @@ namespace Intrepid2 {
     INTREPID2_TEST_FOR_EXCEPTION( cellCoords.extent(2) != primaryCell.getDimension(), std::invalid_argument,
                                   ">>> ERROR (Intrepid2::CellTools::getSubcvCoords): cell coords dimension(2) does not match to the dimension of the cell." );
 #endif
+    using deviceType = typename decltype(subcvCoords)::device_type;
```
CamelliaDPG, line 389: @mperego, I know you and @kyungjoo-kim are already discussing this.  I do think that we need to get some clarity around the meaning of the CellTools DeviceType template argument and what we require of the arguments.  My feeling is that the rule should be that all view arguments should have memory spaces that are accessible from DeviceType::memory_space.  We could add `static_assert`s to that effect.

CamelliaDPG: (If you agree with that, I would drop this typedef, and just use `DeviceType` throughout.

mperego: Yes, I can do that, as long as we agree on it. It might be good to go over this at our meeting on Monday.

kyungjoo-kim: deviceType sounds like a typo of DeviceType. Someone may confuse. what about saying local_device_type ? or device_type ?

packages/intrepid2/src/Cell/Intrepid2_CellToolsDefControlVolume.hpp
```diff
@@ -394,37 +395,33 @@ namespace Intrepid2 {
...
-    typedef Kokkos::View<ordinal_type**,Kokkos::LayoutRight,SpT>             mapViewType;
-
-    typedef typename ExecSpace<typename subcvCoordViewType::execution_space,SpT>::ExecSpaceType ExecutionSpace;
+    using subcvCoordViewType = decltype(subcvCoords);
+    using cellCoordViewType = decltype(cellCoords);
+    using mapViewType = decltype(edgeMap);
 
     const auto loopSize = numCells;
-    Kokkos::RangePolicy<ExecutionSpace,Kokkos::Schedule<Kokkos::Static> > policy(0, loopSize);
+    Kokkos::RangePolicy<typename deviceType::execution_space,Kokkos::Schedule<Kokkos::Static> > policy(0, loopSize);
```
kyungjoo-kim, line 424: using view execution space means that you assume that this is light weight computation. this still assumes that input views are accessible but it does not check its accessibilities.

kyungjoo-kim (approved):

mperego:
@rppawlo @sconde  @TimWildey  @egphill  @glhenni  In this PR we made some changes that could potentially impact your applications. In particular we now enforce that the views that are arguments of `CellTools<DeviceType>` functions have memory spaces that are compatible with `DeviceType`.  This will unlikely affect the applications at the moment since they are probably using Cuda UVM.  
Also we changed the signature of a couple of `CellTools` functions, although we don't expect  the application to use them directly. 
Let me know if you would like to check whether your application builds against this branch before we merge it. If I don't hear back from you, we might merge this tomorrow.

sconde:
testing against Drekar now....

sconde:
looks good on our end

mperego:
> looks good on our end

@sconde thanks for checking!

PR #8924, opened 2021-03-25
Belos: Update Gmres polynomial so degrees match in all versions

jennloe:
This is a very minor fix so that all three versions (Gmres, Arnoldi, and Roots) of the Gmres polynomial solver/preconditioner have consistent degree implementations.  Previously, the Gmres version was 'off by one'.  The GMRES polynomial would apply "max-degree" to p(A), while the other versions let "max-degree" indicate the degree of Ap(A).  Now all three versions do the latter.  

This code is tested in Belos_Tpetra_HybridGMRES_hb_test.exe.


cgcgcg (dismissed):
LGTM

cgcgcg (approved):

Issue #8796, opened 2021-02-22
Adelus: Failing tests on candidate Dev->Master build


prwolfe:
## Bug Report
@trilinos/Adelus

### Description
The candidate build for ats2 develop->master is failing the addles_vector_random_MPI[1-4] tests. All of these fail with errors like 

```
[vortex17:142199] *** Process received signal ***
[vortex17:142199] Signal: Segmentation fault (11)
[vortex17:142199] Signal code: Invalid permissions (2)
[vortex17:142199] Failing at address: 0x7ffef3737680
```

These need to be resolved in order to complete #8129 

### Steps to Reproduce
SHA1:  develop
Use the PR environment from cmake/std/sems/PullRequestCuda10.1.243TestingEnv.sh and instructions from https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors

jwillenbring:
@vqd8a Can you comment on this issue?

vqd8a:
@jwillenbring @prwolfe Sorry for the delay. I would appreciate if you can add a link to the failing tests. It looks to me that these tests were run without CUDA-aware MPI on vortex.

prwolfe:
> @jwillenbring @prwolfe Sorry for the delay. I would appreciate if you can add a link to the failing tests. It looks to me that these tests were run without CUDA-aware MPI on vortex.

The last one is at https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=8914772 - this job runs every 12 hours and you can find that series of jobs at https://testing.sandia.gov/cdash/index.php?project=Trilinos&begin=2021-03-22&end=now&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=PR-cuda10.1.243-test-Trilinos_pullrequest_cuda_10.1.243

vqd8a:
@prwolfe 
The Adelus tests assume CUDA-aware MPI when tested with CUDA while in the test output I see ```TPETRA_ASSUME_CUDA_AWARE_MPI=0```.  
```
WARNING, you have not set TPETRA_ASSUME_CUDA_AWARE_MPI=0 or 1, defaulting to TPETRA_ASSUME_CUDA_AWARE_MPI=0
[...]

PR #8931, opened 2021-03-29
Move the TPETRA_ASSUME_CUDA_AWARE_MPI to the environment

prwolfe:
It looks like the trilinos_jsrun is not getting
CMake variables settings (of course) so I am moving
this into the environemnt settings.

@trilinos/framework 

## Related Issues

* Part of #8796


vqd8a (approved):
@prwolfe Thanks for the fixes.

PR #8934, opened 2021-03-29
atdm/ats2: Only run Adelus tests with TPETRA_ASSUME_CUDA_AWARE_MPI=1

e10harvey:
- Skip Adelus tests when TPETRA_ASSUME_CUDA_AWARE_MPI is 0.
  Related to #8477.

## How was this tested?
On ats2 via:
```
nohup env Trilinos_PACKAGES=Kokkos,Adelus ./ctest-s-local-test-driver.sh all &>ctest-s-local-test-driver.out &
```

Test results posted [here](https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&date=2021-03-29&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=ats2&field2=groupname&compare2=63&value2=Experimental).

### DIFF
--- a/cmake/ctest/drivers/atdm/ats2/local-driver.sh
+++ b/cmake/ctest/drivers/atdm/ats2/local-driver.sh
@@ -32,6 +32,8 @@ fi
 
 # Allow default setting for TPETRA_ASSUME_CUDA_AWARE_MPI=0 in trilinos_jsrun
 unset TPETRA_ASSUME_CUDA_AWARE_MPI
+atdm_config_ctest_regex_old="$ATDM_CONFIG_CTEST_REGEX"
+export ATDM_CONFIG_CTEST_REGEX="$ATDM_CONFIG_CTEST_REGEX -E Adelus*"
 
 echo
 echo "======================================================================="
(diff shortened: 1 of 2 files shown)


prwolfe:
Why would this apply for Adelus only?

e10harvey:
> Why would this apply for Adelus only?

I spoke with the adelus developers and they expect their tests to only work with `TPETRA_ASSUME_CUDA_AWARE_MPI=1`. Perhaps this needs to be addressed in the Adelus source code?

jjellio:
@e10harvey I poked their code when Ross saw failures - I believe their code doesn't exclusively use Tpetra for communication, so Tpetra's cuda-aware variable doesn't provide coverage.

I don't know what the answer to this is - there is no `Trilinos assume cuda aware`. Maybe there could be an attribute in the Cmake/Ctest functions so developers could annotate if they will give MPI buffers that map to `memory spaces` beyond `host` (technically, even unified memory/UVM requires that MPI be device aware (on ATS2) ). At some point we'll have yet-another device, and MPI will either support buffers on the device or not. It's kinda a mess (But thank's to Tpetra atleast that chunk of code is pretty flexible w.r.t. MPI constraints)

e10harvey:
@prwolfe: I spoke with the Adelus developers about this again; they need '-M -gpu' passed to jsrun and decided that the best approach is to disable the single rank mpi test. So, this is ready to merge.

e10harvey:
@prwolfe: Ready to merge?

prwolfe (approved):

PR #8939, opened 2021-03-30
Belos: LHS/RHS vector dump

cgcgcg:
@trilinos/belos

## Motivation
Add the capability to dump LHS/RHS vectors in MatrixMarket format to file.


hkthorn:
@cgcgcg Why is this change of behavior needed?  Isn't it easier to add this interface for outputting these vectors in the Stratimikos interface to Belos?

cgcgcg:
@hkthorn Yes, I could add this only to the `apply` method in the Stratimikos interface if you prefer. I figured it'd be nice to have regardless of how the problem is created.

hkthorn:
@cgcgcg Thanks for the information.  I think this functionality is best left to the interfaces for the linear solvers, like Stratimikos.

cgcgcg:
@hkthorn Sounds good. I'll move it there then.

Issue #8540, opened 2021-01-06
Zoltan2_Sphynx_PolyPrec_MPI_4 randomly failing in -sems-rhel7-intel-18.0.5-openmp-shared debug builds starting on or before 2020-09-02


bartlettroscoe:
CC: @trilinos/zoltan2, @kddevin (Trilinos Data Services Product Lead), @e10harvey 

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label "impacting: tests"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

[...]

PR #8953, opened 2021-04-01
Belos: Add error message to avoid divide by zero crash. 

jennloe:
This PR is in response to issue #8540. This outputs an error message to the user in the event that one of the GMRES polynomial roots is near zero and gives instructions to reduce the polynomial degree or shift the matrix.  I have verified that this error check does catch the issue presented with the singular Laplacian in Zoltan2.

### DIFF
--- a/packages/belos/src/BelosGmresPolyOp.hpp
+++ b/packages/belos/src/BelosGmresPolyOp.hpp
@@ -711,12 +711,12 @@ namespace Belos {
 
     // Set index for sort function, verify roots are non-zero,
     // and sort Harmonic Ritz Values:
-    const MagnitudeType tol = (MagnitudeType) 10 * Teuchos::ScalarTraits<ScalarType>::eps();
+    const MagnitudeType tol = 10.0 * Teuchos::ScalarTraits<MagnitudeType>::eps();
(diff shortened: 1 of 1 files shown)


packages/belos/src/BelosGmresPolyOp.hpp
```diff
@@ -709,10 +709,14 @@ namespace Belos {
       std::cout << "GEEV solve : info = " << info << std::endl;
     }
 
-    // Set index for sort function and sort Harmonic Ritz Values:
+    // Set index for sort function, verify roots are non-zero,
+    // and sort Harmonic Ritz Values:
+    const MagnitudeType tol = (MagnitudeType) 10 * Teuchos::ScalarTraits<ScalarType>::eps();
```
cgcgcg, line 714: Why not just `Teuchos::ScalarTraits<MagnitudeType>::eps()`?

jennloe: Because that's how it is in the BelosMVOPTester.hpp I copied from?  :P  I dunno.  I'll go change it.

cgcgcg (approved):

PR #8956, opened 2021-04-02
Ifpack2: add new options for two-stage GS

iyamazaki:
@trilinos/ifpack2 

## Motivation
This PR adds new options for two-stage GS (inner damping factor, outer sweeps, & compact form):
- copies Kokkos-kernels [PR921](https://github.com/kokkos/kokkos-kernels/pull/921) and
- exposes these new options from Ifpack2

## Testing

Ran ctest under ifpack2 with PullRequestLinuxCuda10.1.105Testing on Ride


jhux2 (commented):
@iyamazaki  Thank you for doing this.  I have one suggestion.  Related methods like SGS and Jacobi use `relaxation: sweeps` to control the number of iterations.  In my opinion, it would be more consistent (and perhaps less confusing) to use `relaxation: sweeps` instead of `relaxation: outer sweeps`, especially because the option for inner sweeps is already labeled to differentiate it from the outer sweeps option.

jhux2:
@trilinos/ifpack2 @csiefer2 @lucbv

iyamazaki:
Thank you, @jhux2 . I was wondering the same.  
- It looks like Ifpack2 has one `relaxation: sweeps` for ifpack2, and there is no parameter to control the number of "local" sweeps (e.g., it hard-codes one sweep for SGS MT for Kokkos-kernel).
- In this PR, two-stage has three nested loops: one for Ifpack2 `relaxation: sweeps`, and two for Kokkos-kernels `relaxation: outer sweeps` and `relaxation: inner sweeps`.

I was wondering, instead, if I should
- hard-code "outer sweep" to one for two-stage GS (which is done before this PR), or
- add a parameter to set the number of local sweeps for the all the relaxation schemes (with default one local sweep)

jhux2:
@iyamazaki You are correct, Ifpack2's other methods only allow changing `relaxation: sweeps` (the number of global iterations), and that the number of local iterations is always one.

Maybe I've misunderstood -- does "outer sweeps" correspond to the number of local iterations?

iyamazaki:
Yes, for two-stage, both "outer sweeps" and "inner sweeps" are passed to the Kokkos kernels, so "outer sweeps" is the number of local iterations for the other methods (which is always one)..

jhux2:
> for two-stage, both "outer sweeps" and "inner sweeps" are passed to the Kokkos kernels, so "outer sweeps" is the number of local iterations for the other methods (which is always one)..

@iyamazaki  I don't feel super strongly about this.  The most important thing is that whatever you choose to do, the options should be well-documented.

jhux2 (approved):
The Ifpack2 changes lgtm.  I haven't reviewed the KokkosKernels changes, but presumably they've been reviewed already b/c this is a hot-patch from the KokkosKernels repo.

jhux2 (dismissed):
Lgtm.  Thanks for rebasing!

packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
```diff
@@ -362,12 +372,23 @@ void Relaxation<MatrixType>::setParametersImpl (Teuchos::ParameterList& pl)
 
   Teuchos::ArrayRCP<local_ordinal_type> localSmoothingIndices = pl.get<Teuchos::ArrayRCP<local_ordinal_type> >("relaxation: local smoothing indices");
 
+  // for Two-stage Gauss-Seidel
+  if (pl.isType<double>("relaxation: inner damping factor")) {
```
cgcgcg, line 376: Maybe change this to
```
if (!std::is_same<double, ST>::value && pl.isType<double>("relaxation: inner damping factor")) {
```
and move it before
```
const scalar_type innerDampingFactor = STS::one ();
pl->set ("relaxation: inner damping factor", innerDampingFactor, "Damping factor for the inner sweep of two-stage GS");
```
I'd hope that it would help avoid having to change the MueLu gold files. Cause MueLu prints "relaxation: inner damping factor" since it's not default anymore.

iyamazaki: Thank you, @cgcgcg. Something like this? Now, I see "[default]" after the print.

jhux2: @cgcgcg Thanks.  The main issue was that the gold files for the ML parameter list conversion tests.

cgcgcg: @jhux2 Unless I messed up, all MueLu interface tests should be ignoring default parameters. Before @iyamazaki change, "relaxation: inner damping factor" wasn't default, so it was getting checked. And I assume the gold files for this PR were updated without ML. With this change, you shouldn't have to update the gold files at all.

cgcgcg: @iyamazaki Wait, you got [default] even though you didn't move those line before `pl.validateParametersAndSetDefaults (* getValidParameters ())`?

iyamazaki: Hi, @cgcgcg. "[default]" wasn't there before, but now I see them.

cgcgcg: Ok, let's see what the AT says. I'm hoping it will fail, since we now don't need the updated gold files.

cgcgcg (approved):

packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
iyamazaki: Should I revert all the gold files?

cgcgcg: @iyamazaki Looks like I messed up somewhere. I'll try to fix it, and then update your branch.

cgcgcg: I fixed the MueLu issue in PR #9011. Once that is merged, this PR won't need any gold file changes. Just tested that locally. In order to avoid merge conflicts, I'd outright nuke commit feea4d8712955835a42fa124108541995274b002 and force push.

iyamazaki: Thank you, @cgcgcg. I just nuked that gold files.

jhux2 (approved):
lgtm

cgcgcg (approved):

iyamazaki:
Thank you, @cgcgcg!!

PR #8960, opened 2021-04-02
tpetra:  fixes for MatrixMatrix products

kddevin:
@trilinos/tpetra 

Much of the CUDA mat-mat stuff is running on host.
This PR adds create_mirror_view_and_copy for device views needed on host
It would be better to use KokkosKernels.

@csiefer2 @lucbv or @brian-kelley  would you please take a look?  For full effect, you have to compare the branch against develop.  This PR compares only against tpetraCrsRefactor; there are other changes to these files (already in tpetraCrsRefactor) that you'll see if you compare with develop.  

If it looks OK to you, please feel free to merge into tpetraCrsRefactor.  If it doesn't look OK, let me know what's needed.

An alternate implementation might convert the views to host in the calls to these functions, rather than in these functions themselves.    Let me know if that makes more sense.

@trilinos/tpetra 

## Testing

MatrixMAtrix_UnitTests now pass with UVM=OFF.

### DIFF
--- a/packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
@@ -890,10 +890,10 @@ namespace Tpetra {
     void KernelWrappers3<Scalar,LocalOrdinal,GlobalOrdinal,Node,LocalOrdinalViewType>::mult_R_A_P_newmatrix_kernel_wrapper(CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Rview,
                                                                                                                            CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Aview,
                                                                                                                            CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Pview,
-                                                                                                                           const LocalOrdinalViewType & Acol2Prow,
-                                                                                                                           const LocalOrdinalViewType & Acol2PIrow,
-                                                                                                                           const LocalOrdinalViewType & Pcol2Accol,
-                                                                                                                           const LocalOrdinalViewType & PIcol2Accol,
+                                                                                                                           const LocalOrdinalViewType & Acol2Prow_dev,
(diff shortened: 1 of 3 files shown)


packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
```diff
@@ -890,10 +890,10 @@ namespace Tpetra {
...
                                                                                                                            CrsMatrixStruct<Scalar, LocalOrdinal, GlobalOrdinal, Node>& Pview,
-                                                                                                                           const LocalOrdinalViewType & Acol2Prow,
-                                                                                                                           const LocalOrdinalViewType & Acol2PIrow,
-                                                                                                                           const LocalOrdinalViewType & Pcol2Accol,
-                                                                                                                           const LocalOrdinalViewType & PIcol2Accol,
+                                                                                                                           const LocalOrdinalViewType & Acol2Prow_dev,
```
csiefer2, line 893: Is there not an instantiation for the triple-product in the Cuda file?

packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
```diff
@@ -933,6 +933,16 @@ namespace Tpetra {
       size_t n = Accolmap->getNodeNumElements();
       size_t p_max_nnz_per_row = Pview.origMatrix->getNodeMaxNumRowEntries();
 
+      // Routine runs on host; have to put arguments on host, too
```
csiefer2, line 936: See above.

packages/tpetra/core/ext/TpetraExt_TripleMatrixMultiply_def.hpp
```diff
@@ -1184,6 +1194,16 @@ namespace Tpetra {
...
```
csiefer2, line 1197: See above.

csiefer2 (approved):
Nevermind the comments.  There is no such monster, which weird.

This code needs a refactor, but that is beyond the scope of what we need to do now.  So this is fine.

PR #8963, opened 2021-04-02
Fix unclosed <tt> in tpetra lesson04 documentation

cwpearson:
@trilinos/tpetra

## Motivation
This change closes an unclosed `<tt>` tag in the Tpetra lesson 04 documentation.
The problem can be seen at the bottom of https://docs.trilinos.org/dev/packages/tpetra/doc/html/Tpetra_Lesson04.html

## Testing
N/A


kddevin (approved):
Thanks, @cwpearson

PR #8968, opened 2021-04-04
ShyLU/HTS: Fix incorrect placement of omp flush.

### LINKED ISSUE #8485: Test ShyLU_NodeHTS_hts_test_1 randomly failing in all ASTRA 'van1-tx2' builds
CC: @trilinos/shylu, @srajama1 (Trilinos Linear Solvers Product Lead), @e10harvey 

<Checklist>
<???: Add label "client: ATDM">
<???: Add label "ATDM Sev: Blocker" (by default but could be other "ATDM Sev: XXX")>
<???: Add label "type: bug"?>
<???: Add label "impacting: tests"?>
<???: Add label for affected packages (e.g. "pkg: MueLu", "pkg: Tpetra", "pkg: Kokkos", etc.)>
<???: Add label "PA: ???Project Area???" (e.g. "PA: Linear Solvers", "PA: Data Services")>
<???: Add milestone "Initial cleanup of new ATDM ..." or "Keep promoted ATDM ...">
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-11-17&end=2020-12-16&filtercount=6&showfilters=1&filtercombine=and&field1=groupname&compare1=62&value1=Experimental&field2=buildname&compare2=65&value2=Trilinos-atdm-van1-tx2&field3=testname&compare3=61&value3=ShyLU_NodeHTS_hts_test_1&field4=status&compare4=62&value4=passed&field5=testoutput&compare5=96&value5=srun%3A%20error.*launch%20failed%3A%20Error%20configuring%20interconnect&field6=testoutput&compare6=95&value6=HTS%20Test%3A%20Failed) (click "Shown Matching Output" in upper right) the tests:

* `ShyLU_NodeHTS_hts_test_1`

in the builds:

* `Trilinos-atdm-van1-tx2_arm-20.0_openmpi-4.0.2_openmp_static_dbg`
* `Trilinos-atdm-van1-tx2_arm-20.0_openmpi-4.0.2_openmp_static_opt`
* `Trilinos-atdm-van1-tx2_arm-20.1_openmpi-4.0.3_openmp_static_dbg`
* `Trilinos-atdm-van1-tx2_arm-20.1_openmpi-4.0.3_openmp_static_opt`

when these 'van1-tx2' (i.e. ASTRA) builds first got set up.

The only error output from this test as shown [here](https://testing.sandia.gov/cdash/test/46352014) shows:

```
================================================================================

TEST_0

Running: "/opt/atse/mpi/openmpi4-arm/4.0.2/bin/mpirun" "--bind-to" "none" "-np" "1" "/lustre/jenkins/stria/workspace/Trilinos-atdm-van1-tx2_arm-20.0_openmpi-4.0.2_openmp_static_dbg/SRC_AND_BUILD/BUILD/packages/shylu/shylu_node/hts/test/hts_test.exe"

--------------------------------------------------------------------------------

HTS Test: Failed

--------------------------------------------------------------------------------

TEST_0: Return code = 0
TEST_0: Pass criteria = Match REGEX {Passed} [FAILED]
TEST_0: Result = FAILED

================================================================================
```

giving zero clues as to why the test is failing.

## Current Status on CDash

Run the [above query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2020-11-17&end=2020-12-16&filtercount=6&showfilters=1&filtercombine=and&field1=groupname&compare1=62&value1=Experimental&field2=buildname&compare2=65&value2=Trilinos-atdm-van1-tx2&field3=testname&compare3=61&value3=ShyLU_NodeHTS_hts_test_1&field4=status&compare4=62&value4=passed&field5=testoutput&compare5=96&value5=srun%3A%20error.*launch%20failed%3A%20Error%20configuring%20interconnect&field6=testoutput&compare6=95&value6=HTS%20Test%3A%20Failed) adjusting the "Begin" and "End" dates to match today any other date range or just click "CURRENT" in the top bar to see results for the current testing day.

## Steps to Reproduce

One should be able to reproduce this failure as described in:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md

and the system-specific instructions at:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md#specific-instructions-for-each-system

Just log into any of the associated machines and copy and paste the full CDash build name `<build-name>` listed above and run commands like:

```
$ cd <some_build_dir>/

$ source $TRILINOS_DIR/cmake/std/atdm/load-env.sh <build-name>

$ cmake \
 -GNinja \
 -DTrilinos_ENABLE_TESTS=ON -DTrilinos_ENABLE_<package-name>=ON \
 $TRILINOS_DIR

$ make NP=16

$ <command-to-run-on-compute-node> ctest -j4
```

where `<package-name>` is any package that you want to enabled to reproduce build and/or test results.

Again, for exact system-specific details on what commands to run to build and run tests, see:

* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md#specific-instructions-for-each-system

And if you can't figure out what commands to run to produce the issue given the above-referenced documentation, please post a comment here and we will give you the exact minimal commands to reproduce the failures.

ambrad:
Running hts_test on a new architecture revealed a bug: one omp flush line was
incorrectly placed, the one after s.n1Axpy. Fix this; the flush release should
be before the s_done lock is set, not after.

This achitecture is quite sensitive to omp flush, so I also took the opportunity
to rearrange placement of some others.

Also fix two cases of using int instead of size.

Closes #8485.

@trilinos/shylu

### DIFF
--- a/packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
+++ b/packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
@@ -1870,7 +1870,7 @@ SerialBlock::reinit_numeric_dense (const CrsMatrix& A) {
          j < irip1; ++j) {
       const Int lcol = A.jc[j] - c0_;
       if (lcol >= nc_) break;
-      const Int k = nc_*lrow + lcol;
+      const Size k = nc_*lrow + lcol;
       d_[k] = A.d[j];
     }
   }
@@ -3800,7 +3800,8 @@ inline void SerialBlock_n1Axpy_dense (
   const Int nrhs, Sclr* y, const Int ldy)
 {
   for (Int g = 0; ; ) {
-    for (Int i = 0, k = 0; i < nr; ++i) {
+    size_t k = 0;
+    for (Int i = 0; i < nr; ++i) {
       Sclr a = 0;
       for (Int j = 0; j < nc; ++j, ++k) a += d[k]*x[j];
       y[i] -= a;
@@ -4044,41 +4045,47 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
   const Int nthreads = t.nthreads();
   if (nthreads == 1) {
     t.solve(x, ldx, x, ldx, nrhs);
+#ifdef _OPENMP
+#   pragma omp flush
+#endif
     *t_barrier = step;
   } else {
     // Solve T wrk_ = x.
     t.solve(x, ldx, wrk_.data(), t.get_n(), nrhs);
-    { // Wait for the block row MVPs to finish.
-      const Int done = (step << 1);
-      inv_tri_done[tid] = done;
 #ifdef _OPENMP
 #   pragma omp flush
 #endif
+    { // Wait for the block row MVPs to finish.
+      const Int done = (step << 1);
+      inv_tri_done[tid] = done;
       for (Int i = 0; i < nthreads; ++i)
         while (inv_tri_done[i] < done) ;
     }
+#ifdef _OPENMP
+#   pragma omp flush
+#endif
     // Copy wrk_ to x.
     const Int row_start = t.block_row_start(tid), nr = t.block_nr(tid);
     for (Int irhs = 0; irhs < nrhs; ++irhs)
       memcpy(x + irhs*ldx + row_start,
              wrk_.data() + irhs*t.get_n() + row_start,
              nr*sizeof(Sclr));
+#ifdef _OPENMP
+#   pragma omp flush
+#endif
     { // Wait for the memcpy's to finish.
       const Int done = (step << 1) + 1;
       inv_tri_done[tid] = done;
-#ifdef _OPENMP
-#     pragma omp flush
-#endif
       //todo Not every thread necessarily needs this on-diag tri's solution, but
       // our dep graph doesn't encode that yet.
       for (Int i = 0; i < nthreads; ++i)
         while (inv_tri_done[i] < done) ;
     }
-    if (tid == 0)
-      *t_barrier = step;
 #ifdef _OPENMP
 #   pragma omp flush
 #endif
+    if (tid == 0)
+      *t_barrier = step;
   }
 }
 
@@ -4103,7 +4110,8 @@ RecursiveTri::solve (const Sclr* b, Sclr* x, const Int ldx,
   const p2p_Done done_symbol = nd_.done_symbol;
   { const OnDiagTri& t = nd_.t[0];
     if (tid < t.nthreads())
-      ondiag_solve(t, x, ldx, nrhs, tid, 0, t_barrier, inv_tri_done); }
+      ondiag_solve(t, x, ldx, nrhs, tid, 0, t_barrier, inv_tri_done);
+    }
   if ( ! nd_.os.empty()) {
     os += nd_.t[0].get_n();
     x_osi = x + nd_.os[0];
@@ -4119,17 +4127,14 @@ RecursiveTri::solve (const Sclr* b, Sclr* x, const Int ldx,
     if ( ! s.empty() && (s.parallel() || tid == 0)) {
       rbwait(s_done, s_ids, s_idx, i, done_symbol);
       s.n1Axpy(x_osi, ldx, nrhs, x_os, ldx, tid);
-      s_done[rb_p2p_sub2ind(i, tid)] = done_symbol;
 #ifdef _OPENMP
 #     pragma omp flush
 #endif
+      s_done[rb_p2p_sub2ind(i, tid)] = done_symbol;
     }
     if (tid < t.nthreads()) {
       rbwait(s_done, t_ids, t_idx, i, done_symbol);
       ondiag_solve(t, x_os, ldx, nrhs, tid, i+1, t_barrier, inv_tri_done);
-#ifdef _OPENMP
-#     pragma omp flush
-#endif
     }
     os += t.get_n();
     x_osi = x + nd_.os[i+1];
--- a/packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
+++ b/packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
@@ -3989,6 +3989,7 @@ LevelSetTri::solve (const Sclr* b, Sclr* x, const Int ldx,
           while (*done != p2p_done_value) ;
         }
       }
+      // Flush (acquire) the global to the local view of x.
 #ifdef _OPENMP
 #     pragma omp flush
 #endif
@@ -4001,6 +4002,7 @@ LevelSetTri::solve (const Sclr* b, Sclr* x, const Int ldx,
           a -= x[jc[j]] * d[j];
         x[r] = a * d[j++];
       }
+      // Flush (release) the local to the global view of x.
 #ifdef _OPENMP
 #     pragma omp flush
 #endif
--- a/packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
+++ b/packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
@@ -3989,7 +3989,8 @@ LevelSetTri::solve (const Sclr* b, Sclr* x, const Int ldx,
           while (*done != p2p_done_value) ;
         }
       }
-      // Flush (acquire) the global to the local view of x.
+      // Flush (acquire) the global to the local view of x. With OpenMP 5.0, we
+      // can specify acquire, release, or both, but for now we don't have 5.0.
 #ifdef _OPENMP
 #     pragma omp flush
 #endif
@@ -4025,7 +4026,13 @@ inline void Impl<Int, Size, Sclr>::
 rbwait (volatile p2p_Done* const s_done, const Size* s_ids,
         const Int* const s_idx, const Int i, const p2p_Done done_symbol) {
   const Int si = s_idx[i], si1 = s_idx[i+1];
-  if (si == si1) return;
+  if (si == si1) {
+    // acquire
+#ifdef _OPENMP
+#   pragma omp flush
+#endif
+    return;
+  }
   const Size* id = s_ids + si;
   const Size* const idn = s_ids + si1;
   while (id != idn) {
@@ -4033,7 +4040,7 @@ rbwait (volatile p2p_Done* const s_done, const Size* s_ids,
     while (*d != done_symbol) ;
     ++id;
   }
-  // Make sure x is updated.
+  // acquire
 #ifdef _OPENMP
 # pragma omp flush
 #endif
@@ -4047,6 +4054,7 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
   const Int nthreads = t.nthreads();
   if (nthreads == 1) {
     t.solve(x, ldx, x, ldx, nrhs);
+    // release
 #ifdef _OPENMP
 #   pragma omp flush
 #endif
@@ -4054,6 +4062,7 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
   } else {
     // Solve T wrk_ = x.
     t.solve(x, ldx, wrk_.data(), t.get_n(), nrhs);
+    // release
 #ifdef _OPENMP
 #   pragma omp flush
 #endif
@@ -4063,6 +4072,7 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
       for (Int i = 0; i < nthreads; ++i)
         while (inv_tri_done[i] < done) ;
     }
+    // acquire
 #ifdef _OPENMP
 #   pragma omp flush
 #endif
@@ -4072,6 +4082,7 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
       memcpy(x + irhs*ldx + row_start,
              wrk_.data() + irhs*t.get_n() + row_start,
              nr*sizeof(Sclr));
+    // release
 #ifdef _OPENMP
 #   pragma omp flush
 #endif
@@ -4083,9 +4094,6 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
       for (Int i = 0; i < nthreads; ++i)
         while (inv_tri_done[i] < done) ;
     }
-#ifdef _OPENMP
-#   pragma omp flush
-#endif
     if (tid == 0)
       *t_barrier = step;
   }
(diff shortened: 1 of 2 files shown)


srajama1 (dismissed):
Thank you @ambrad !

ambrad:
@srajama1 this PR fails the ride integration test. I think none of the ride nightlies turn on HTS, so we haven't seen this before. I'll look into this next I have some time, maybe next weekend, and add a commit to this PR that resolves the Ifpack2/HTS test failure on ride. Until then, I've labeled this PR as WIP.

ambrad:
@srajama1 I'm unable to reproduce the ride error in the integration autotesting. I ran
```
cmake \
    -C PullRequestLinuxCuda10.1.105TestingSettings.cmake \
    -C packageEnables.cmake \
    -DTrilinos_ENABLE_TESTS:BOOL=ON \
    -DTrilinos_PARALLEL_LINK_JOBS_LIMIT:STRING=2 \
    -G "Ninja" \
    ../Trilinos
```
using the two *.cmake files from the [CDash data](https://testing.sandia.gov/cdash/viewFiles.php?buildid=8958930) for this PR and the ride environment, written out explicitly so I know exactly what is happening,
```
module load git/2.10.1
module load devpack/20190404/openmpi/4.0.1/gcc/7.2.0/cuda/10.1.105
module swap openblas/0.3.4/gcc/7.4.0 netlib/3.8.0/gcc/7.2.0
export OMPI_CXX=/ascldap/users/ambradl/tril/Trilinos/packages/kokkos/bin/nvcc_wrapper
export OMPI_CC=`which gcc`
export OMPI_FC=`which gfortran`
export CUDA_LAUNCH_BLOCKING=1
export CUDA_MANAGED_FORCE_DEVICE_ALLOC=1
export PATH=/home/atdm-devops-admin/tools/ride/cmake-3.17.2/bin/:/ascldap/users/rabartl/install/white-ride/ninja-1.8.2/bin:$PATH
```
Then I ran the tests ten times, and I got 100% passes all ten times. The CDash failure above can't be affected by this PR because _OPENMP isn't defined in that build; except for two trivial variable type changes, this PR modifies only code guarded by _OPENMP. Thus, I recommend manually merging this PR to resolve issue 8485.

ambrad:
@srajama1 never mind. I took a guess that
```
export CUDA_LAUNCH_BLOCKING=1
export CUDA_MANAGED_FORCE_DEVICE_ALLOC=1
```
is not set in the above build, and removing these reproduces the failure. I'll push a fix in a moment.

ambrad:
@srajama1 or @trilinos/shylu This PR now passes integration testing, but Siva's approval is stale. Would someone re-approve it so it can be merged? Thanks.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -4044,41 +4045,47 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
   const Int nthreads = t.nthreads();
   if (nthreads == 1) {
     t.solve(x, ldx, x, ldx, nrhs);
+#ifdef _OPENMP
+#   pragma omp flush
+#endif
     *t_barrier = step;
```
e10harvey, line 4051: This is the thread-wide lock acquisition?

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -4044,41 +4045,47 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
   const Int nthreads = t.nthreads();
   if (nthreads == 1) {
     t.solve(x, ldx, x, ldx, nrhs);
+#ifdef _OPENMP
```
e10harvey, line 4048: Can you add a comment explaining why this is needed before the lock acquisition?

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -1870,7 +1870,7 @@ SerialBlock::reinit_numeric_dense (const CrsMatrix& A) {
          j < irip1; ++j) {
       const Int lcol = A.jc[j] - c0_;
       if (lcol >= nc_) break;
-      const Int k = nc_*lrow + lcol;
+      const Size k = nc_*lrow + lcol;
```
e10harvey, line 1873: Changed to Size to avoid overflow?

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -3800,7 +3800,8 @@ inline void SerialBlock_n1Axpy_dense (
   const Int nrhs, Sclr* y, const Int ldy)
 {
   for (Int g = 0; ; ) {
-    for (Int i = 0, k = 0; i < nr; ++i) {
+    size_t k = 0;
```
e10harvey, line 3803: Changes to size_t to avoid overflow?

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -4044,41 +4045,47 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
...
 #ifdef _OPENMP
 #   pragma omp flush
 #endif
+    { // Wait for the block row MVPs to finish.
+      const Int done = (step << 1);
+      inv_tri_done[tid] = done;
       for (Int i = 0; i < nthreads; ++i)
         while (inv_tri_done[i] < done) ;
     }
+#ifdef _OPENMP
```
e10harvey, line 4064: Needed for updating the current thread's view of memory before reading `_wrk.data()` and `t.get_n()`?

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -4044,41 +4045,47 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
...
       memcpy(x + irhs*ldx + row_start,
              wrk_.data() + irhs*t.get_n() + row_start,
              nr*sizeof(Sclr));
+#ifdef _OPENMP
+#   pragma omp flush
+#endif
     { // Wait for the memcpy's to finish.
       const Int done = (step << 1) + 1;
       inv_tri_done[tid] = done;
-#ifdef _OPENMP
```
e10harvey, line 4069: Why was this moved above?

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -4044,41 +4045,47 @@ ondiag_solve (const OnDiagTri& t, Sclr* x, const Int ldx, const Int nrhs,
...
   if (nthreads == 1) {
     t.solve(x, ldx, x, ldx, nrhs);
+#ifdef _OPENMP
+#   pragma omp flush
+#endif
     *t_barrier = step;
   } else {
     // Solve T wrk_ = x.
     t.solve(x, ldx, wrk_.data(), t.get_n(), nrhs);
-    { // Wait for the block row MVPs to finish.
```
e10harvey, line 4051: Why was this moved down?

e10harvey (commented):
Thanks, @ambrad ! Please see comments and questions.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
ambrad: Yes.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
ambrad: It's not a lock but rather a sync of the thread's local view of memory with the global memory.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
ambrad: One has to sync the local view of memory with the global view. Prior to this ARM architecture, this code ran only on Intel CPUs, and Intel CPUs tend to enforce memory consistency much more strongly than the OpenMP memory model requires. Thus, I never had a testbed to really get the flush placement tested robustly. ARM is very nifty in that it seems to take advantage of the OpenMP memory model quite efficiently, so it's great for flushing out any flush bugs.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
ambrad: Yes.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
ambrad: Sync of memory has to occur before the lock acquistion.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
ambrad: I added comments re: acquire and release flushes to the first two instances of each, to explain the general pattern.

ambrad:
@e10harvey if you're interested in this topic, see [this document](https://www.openmp.org/spec-html/5.0/openmpse4.html#x17-160001.4) for a description of the OpenMP memory model and a subsection of it for a description of what flush does. Note this is OpenMP 5.0, which permits a clause specifying release, acquire, or both in a flush, but Trilinos doesn't use OpenMP 5.0 yet, so the flushes in HTS are still bare.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
e10harvey: Ah, to ensure that the store is written to global memory for other threads to read.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -4119,17 +4129,14 @@ RecursiveTri::solve (const Sclr* b, Sclr* x, const Int ldx,
     if ( ! s.empty() && (s.parallel() || tid == 0)) {
       rbwait(s_done, s_ids, s_idx, i, done_symbol);
       s.n1Axpy(x_osi, ldx, nrhs, x_os, ldx, tid);
-      s_done[rb_p2p_sub2ind(i, tid)] = done_symbol;
 #ifdef _OPENMP
 #     pragma omp flush
 #endif
+      s_done[rb_p2p_sub2ind(i, tid)] = done_symbol;
```
e10harvey, line 4135: This store not getting flushed to global memory in some processes?

e10harvey (commented):
Thanks for the documentation link. One more question.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
ambrad: The `volatile` keyword leads to an implied flush of that variable. It's equivalent to the `list` clause in the OpenMP 5.0 `flush` construct, with the volatile variables in the list.

e10harvey (approved):

ambrad:
@e10harvey, after discussing these flushes with you, I spotted I think an opportunity to move one to reduce the number of flushes in certain cases by a few, so I'm going to try that out. Hence I put this back as AT:WIP.

ambrad:
@e10harvey thanks for the discussion. I added a commit that documents release vs acquire flushes.

While pairing the acquire and release flushes, I realized there was one flush at the end of ondiag_solve that was not quite correct, in the sense that it pairs correctly but is associated with the wrong solve operation. It was acting in place of a missing acquire flush in rbwait in the early-exit branch of rbwait. It was also running more times than it should because of the incorrect solve phase association. So this commit also moves the acquire flush from the end of ondiag_solve to rbwait.

packages/shylu/shylu_node/hts/src/shylu_hts_impl_def.hpp
```diff
@@ -3989,7 +3989,8 @@ LevelSetTri::solve (const Sclr* b, Sclr* x, const Int ldx,
           while (*done != p2p_done_value) ;
         }
       }
-      // Flush (acquire) the global to the local view of x.
+      // Flush (acquire) the global to the local view of x. With OpenMP 5.0, we
```
e10harvey, line 3992: This is great; thanks for adding it.

e10harvey (approved):

e10harvey:
> @e10harvey thanks for the discussion. I added a commit that documents release vs acquire flushes.
> 
> While pairing the acquire and release flushes, I realized there was one flush at the end of ondiag_solve that was not quite correct, in the sense that it pairs correctly but is associated with the wrong solve operation. It was acting in place of a missing acquire flush in rbwait in the early-exit branch of rbwait. It was also running more times than it should because of the incorrect solve phase association. So this commit also moves the acquire flush from the end of ondiag_solve to rbwait.

@ambrad Thinking about this more now I am wondering if atomics or another synchronization mechanism could be used to simplify this flush logic. Thanks again for looking into this.

ambrad:
> Thinking about this more now I am wondering if atomics or another synchronization mechanism could be used to simplify this flush logic.

@e10harvey it's an interesting question. First, it's important to distinguish between operation syncs (sequencing computer instructions) and memory syncs (making a thread's local view of memory consistent with the true global memory). Second, particularly for operation syncs, one must choose from among a range of methods of varying costs.

The costliest sync method is to end a parallel region; doing so syncs both operations and memory and also releases the threads unless a certain "hot thread" setting is true.

Second costliest is a `barrier` within a region, either explicitly called or implicit in another OpenMP construct like `for`; this syncs both ops and memory but keeps the threads active. You see barriers a lot in code.

`atomic` and `critical` regions also do both, but they are useful only (1) when the data set being sync'ed against is small, essentially one datum per lock, or (2) the code is on GPU, where atomics are extremely fast. (But the GPU has other challenges.) In the case of a solver, the data set is large, so atomics on CPU would be slow.

After these come sync methods that act separately on memory and operations and within a subset of threads. This is where `flush` comes in. Any sync construct that acts _just_ to sync operations almost certainly requires a flush to update the thread's view of global memory or to flush its writes to global memory. These methods are good for the case where large chunks of memory get read and written between syncs.

HTS is a type of solver that uses light-weight thread-to-thread (or sometimes called point-to-point) operation syncs. Hence it must use flushes, too, since there is no light-weight thread-to-thread operation sync construct that also syncs the thread's view of memory.

If you're interested in some nitty-gritty details, see, e.g., these [two](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.727.516&rep=rep1&type=pdf) [papers](https://epubs.siam.org/doi/pdf/10.1137/1.9781611974690.ch2).

PR #8973, opened 2021-04-05
Tpetra: add missing fence in WrappedDualView

tasmith4:
@trilinos/tpetra 

When we get the host view of a WrappedDualView with OverwriteAll, a device kernel is potentially still changing values on the device view, but we don't care because we're overwriting them anyway.  In UVM though, we need to wait to overwrite on host until device is done writing, because otherwise the things we write on host could be overwritten by the device kernel that hasn't finished yet.

### Testing

The need for this change was exposed by an existing WrappedDualView test.


kddevin (approved):
We will review with the team on Tuesday.

Issue #8982, opened 2021-04-06
ShyLU:  Unit tests appear to depend on Gtest (without depending on it)


jjellio:
## Bug Report
@trilinos/shylu 

### Description
I'm seeing a build failure with ATDMDevSettings + EMPIRE's cmake options file. (EMPIRE disables Gtest)

  >> 21835    /tmp/coe0201/spack-stage/spack-stage-atdm-trilinos-develop-m7i5m3n7t54wtfuxzq4papu3scxbezkt/spack-src/packages/shylu/shylu_node/tacho/unit-test/Tacho_TestSerial_double.cpp:1:10: fatal error: 'gtest/gtest.h' file not found
     21836    #include <gtest/gtest.h>

Maybe Tacho needs to require Gtest for this unit test?

```
cmake \
-DTrilinos_ENABLE_TESTS:BOOL=ON
-DTrilinos_ENABLE_EXAMPLES:BOOL=ON
-DTrilinos_ENABLE_ALL_PACKAGES:BOOL=ON
-DTrilinos_TRACE_ADD_TEST:BOOL=ON
-DTrilinos_ENABLE_CONFIGURE_TIMING:BOOL=ON
-DTrilinos_HOSTNAME:STRING=redwood
-DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake;cmake/std/atdm/apps/empire/EMPIRETrilinosEnables.cmake
```
Curiously, the CMake output shows:
```
-- Setting ShyLU_NodeTacho_ENABLE_TESTS=OFF because ShyLU_NodeTacho has a required test/example dependence on disabled package Gtest
-- Setting ShyLU_NodeTacho_ENABLE_EXAMPLES=OFF because ShyLU_NodeTacho has a required test/example dependence on disabled package Gtest
```

kyungjoo-kim:
i can handle it. probably by the end of this week.

jjellio:
I can post more if you want - the build is a little experimental since it's on an NDA machine.

I don't really understand why it's happening... seems the tests should be disabled!

Thanks


```
Processing enabled package: ShyLU_Node (Tacho, Tests, Examples)
-- ShyLU_NodeTacho_Tacho_TestUtilTaskScheduler_MPI_1: Added test (BASIC, NUM_MPI_PROCS=1, PROCESSORS=1)!
[...]

PR #8990, opened 2021-04-08
Shylu - disable tests and examples explicitly when gtest is not available

kyungjoo-kim:
## Motivation

In #8982  @jjellio found a case that atdm build script for empire and sparc does not build correctly wihtout satisfying the dependency on gtest. The problematic tests are supposed to be disabled but they are not with the script. When I just use cmake script enalbing the Tacho, I could not reproduce the same issue. This PR includes work around for the problem by excluding the problematic tests and examples explicitly when gtest is not enabled. 

## Testing

I confirmed the configuration on a testbed.


jjellio (approved):

jjellio:
This fixed build errors where I was enabling all packages, then using ATDM+EMPIRE to disable (among other things) Gtest.

This exposed that Tacho/BDDC needs Gtest, wasn't guarded against building tests/examples when it is enabled.

These patches wrap the tests and examples in guards for Gtest - and fixed my crazy build.

jjellio:
Ok - approved and set AUTOMERGE - thanks for the help

Issue #8798, opened 2021-02-22
Zoltan: Failing tests on candidate Dev->Master build


prwolfe:
## Bug Report
@trilinos/zoltan 

### Description
The tests

```
Zoltan_ch_simple_zoltan_parallel
Zoltan_hg_simple_zoltan_parallel
```
Are failing in the candidate build for ats2 on vortex. They seem to be differences with gold files mainly.

These need to run clean in oder to complete #8129

### Steps to Reproduce
SHA1: develop
Use the PR environment from cmake/std/sems/PullRequestCuda10.1.243TestingEnv.sh and instructions from https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors

jwillenbring:
@kddevin Is there someone who could look at this, or is it safe to simply disable these tests for this build?

kddevin:
@jwillenbring @prwolfe forgive me for the delay.  Can you add a link to the test so that I can peek at the Zoltan output?
I know you gave steps to reproduce, but if I can leverage your runs, I can make progress faster.  Thanks!

prwolfe:
> @jwillenbring @prwolfe forgive me for the delay. Can you add a link to the test so that I can peek at the Zoltan output?
> I know you gave steps to reproduce, but if I can leverage your runs, I can make progress faster. Thanks!

The series is at https://testing.sandia.gov/cdash/index.php?project=Trilinos&begin=2021-03-28&end=now&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=PR-cuda10.1.243-test-Trilinos_pullrequest_cuda_10.1.243 and the last Zoltan tests are at https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=8914829

These run every 12 hours.

kddevin:
@prwolfe @jwillenbring 
I have tracked this failure down to a difference in vortex's MPI implementation (Spectrum) from the implementation used on other test platforms (OpenMPI).

```
  struct {
    float val;
    int rank;
  } local, global;
  local.val = 1.0;
  local.rank = me;
  MPI_Allreduce(&local, &global, 1, MPI_FLOAT_INT, MPI_MINLOC, MPI_COMM_WORLD);
  printf("%d MINLOC value %f on rank %d\n", me, global.val, global.rank);
```

When I run this code on vortex, it says the minimum value is on rank 3.
When I run it with OpenMPI, it says the minimum value is on rank 0.
[...]

PR #8995, opened 2021-04-09
Zoltan:  add test to ensure that MPI MINLOC works correctly for gold-standard testing

kddevin:
@trilinos/zoltan @trilinos/framework 

## Motivation

Spectrum MPI appears to incorrectly implement the MPI standard.  Zoltan's gold-standard testing relies on having a correct MPI implementation.  This test detects the incorrect implementation of MPI that caused tests to fail for #8798.

Zoltan's gold-standard answers were generated with MPI that returns the
lowest numbered rank in MPI_Allreduce with MPI_MINLOC when the compared
values are equal on several processors.  

The MPI standard indicates that MPI_MINLOC should return the lowest
numbered rank; see, for example, 
https://www.mpi-forum.org/docs/mpi-1.1/mpi-11-html/node79.html

Spectrum MPI incorrectly returns the largest rank with the minimum value.

Some Zoltan PHG tests will fail if the lowest rank is not returned. #8798 
Zoltan_PHG_CoarsePartition tries several coarse partitions and selects
the best wrt hyperedge cut.  When two processors compute equally good
but different coarse partitions, the choice of a different coarse
partition leads to a different final partition.  The final partition is
correct, but it doesn't match the gold-standard answers and, thus, causes
Zoltan tests to fail.  

This test detects the condition needed in MINLOC to match the gold-standard
answers.  If this test fails, 
* Zoltan will still operate correctly and produce correct partitions
* But the Zoltan tests may report failures

Until SpectrumMPI is fixed, this test will have to be disabled in vortex PR testing (along with other Zoltan PHG tests).

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to #8798 
* Part of 
* Composed of 

## Testing

Tested on Mac with OpenMPI -- the new test passes
Tested on Vortex with Spectrum MPI -- the new test fails


prwolfe (approved):
Nice detective work - thanks for adding this test for the spectrumMpi issue as well as sending them you bug report!

PR #9007, opened 2021-04-13
MueLu: Enable reuse of Ifpack2 smoothers

cgcgcg:
@trilinos/muelu 

## Motivation
Enable reuse of some Ifpack2 preconditioners. This is only the plumbing on the MueLu side. Ifpack2 might still require some work.


GrahamBenHarper (approved):
nice! lgtm

PR #9008, opened 2021-04-13
Amesos2: Fixes mismatched new/delete and memory leaks in reused solver objects

hkthorn:
Integrating the templated Basker solver directly into Xyce to perform
custom linear solves for harmonic balance (HB) analysis, some memory
issues were noticed.  First there is a mismatch in the new and delete
used to create and destroy the pinv object, respectively.  If the same
solver object is reused, the L and U factors are leaked.  Also some
internal workspace is not properly cleaned up.  Due to clarity the
internal workspace objects have been refactored and the L and U
factors are deleted if the solver object is called to perform a
numeric factorization when one already exists.

This addresses the memory issues that have been observed through
valgrind.

@trilinos/amesos2 

## Motivation

Integrating the templated Basker solver directly into Xyce to perform
custom linear solves for harmonic balance (HB) analysis, some memory
issues were noticed.  First there is a mismatch in the new and delete
used to create and destroy the pinv object, respectively.  If the same
solver object is reused, the L and U factors are leaked.  Also some
internal workspace is not properly cleaned up.  Due to clarity the
internal workspace objects have been refactored and the L and U
factors are deleted if the solver object is called to perform a
numeric factorization when one already exists.

This addresses the memory issues that have been observed through
valgrind.

## Testing

This was tested on OSX / GNU 10.x compilers.  It has been verified on ascic180 with Intel 19.x compilers.


packages/amesos2/src/basker/basker_def.hpp
```diff
@@ -247,17 +255,18 @@ namespace BaskerClassicNS{
...
 
     /*Creating working space*/
-    Int *tptr;
+    Int *color, *pattern, *stack; 
     Entry *X;
-    //tptr = (Int *)   BASKERCALLOC( (ncol)+(4*nrow), sizeof(Int));
-    tptr = new Int[(ncol)+(4*nrow)]();
+    color = new Int[ncol]();
+    pattern = new Int[nrow]();
+    stack = new Int[2*nrow]();
```
ndellingwood, line 262: Thanks @hkthorn! Changes look reasonable to me, @srajama1 let us know if there's any unexpected consequence for splitting the workspace tptr allocation into the separate explicit allocations for color, pattern and stack (instead of previous behavior of later offsetting those pointers into the larger contiguous memory allocation of tptr)

ndellingwood (approved):

hkthorn:
@ndellingwood Could you review again, I updated my fork, but nothing changed in the code.  It was just a merge commit.

ndellingwood (approved):

PR #9018, opened 2021-04-15
Tempus: Replace Logical Operators

### LINKED ISSUE #9017: Tempus:  Replace Logical Operators
## Enhancement
@trilinos/tempus 

The ROL team has received requests from Windows users to remove the alternative logical operators 'and' and 'or' in favor of && and ||.  The C++ standard includes 'and' and 'or', of course.  However, the MS compiler only supports them in the -permissive mode, which apparently isn't allowed in many companies, due to the quality control niceties that the non-permissive (standard) mode provides.

Trilinos is not supporting Windows and do not have even platforms to test on.  This change should be straight forward, but there is not any mechanism to prevent regression.  Microsoft needs to support the c++ standard!

ccober6:
The ROL team has received requests from Windows users to remove the
alternative logical operators 'and' and 'or' in favor of && and ||.
The C++ standard includes 'and' and 'or', of course. However, the
MS compiler only supports them in the -permissive mode, which
apparently isn't allowed in many companies, due to the quality
control niceties that the non-permissive (standard) mode provides.

Trilinos is not supporting Windows and do not have even platforms
to test on. This change should be straight forward, but there is
not any mechanism to prevent regression. Microsoft needs to support
the c++ standard!

Not sure all instances were found and there is no compiler flag
to throw on this.

@trilinos/tempus 
@trilinos/rol

## Motivation
See above

## Related Issues
* Closes #9017 

## Stakeholder Feedback
Not at this time.

## Testing
Passes all tests on Mac.

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
@@ -391,8 +391,8 @@ bool IntegratorBasic<Scalar>::advanceTime()
     startIntegrator();
     integratorObserver_->observeStartIntegrator(*this);
 
-    while (integratorStatus_ == WORKING and
-        timeStepControl_->timeInRange (solutionHistory_->getCurrentTime()) and
+    while (integratorStatus_ == WORKING &&
+        timeStepControl_->timeInRange (solutionHistory_->getCurrentTime()) &&
         timeStepControl_->indexInRange(solutionHistory_->getCurrentIndex())){
 
       stepperTimer_->reset();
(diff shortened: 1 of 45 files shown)


dridzal (approved):
This looks good.  Of course, there is no way for us to check if this is all of them.  I'll lean on ROL's Windows users to give us feedback.

ccober6:
@dridzal Ok, it is merged.  Let us know if we missed any!

dridzal:
@ccober6 thanks!  Will do.  We'll run the scripts that @gregvw wrote, on Tempus, ROL and a few other packages.  We'll also get input from our Windows users early next week.

PR #9024, opened 2021-04-16
MueLu: fix agg export for multiple dofs per node

pohm01:
@trilinos/muelu 

## Motivation
Aggregation export gives garbage when used for problems with multiple dofs per node (like elasticity). To fix this, ghost coordinates are no longer added when aggregates don't cross processors (no need for ghosts), which fixes the problems when there are multiple dofs per node.

## Testing
As far as I know, aggregation export is not tested.
Looking at the output in paraview, the aggregates for elasticity now resemble aggregates instead of a crazy blob of things. Scalar equations look the same as before (except for the "point cloud" which is also better now because it now gives aggregate labels to the nodes on processor interfaces).

## Additional Information
The map that is used to find ghost coordinates in the aggregation export might be wrong, in which case this just circumvents the problem instead of actually fixing it.

### DIFF
--- a/packages/muelu/src/Utils/MueLu_AggregationExportFactory_def.hpp
+++ b/packages/muelu/src/Utils/MueLu_AggregationExportFactory_def.hpp
@@ -186,7 +186,8 @@ namespace MueLu {
       dims_ = coords->getNumVectors();  //2D or 3D?
       if(numProcs > 1)
       {
-        {
+        if (aggregates->AggregatesCrossProcessors())
+        { // Do we want to use the map from aggregates here instead of the map from A? Using the map from A seems to be problematic with multiple dofs per node
           RCP<Import> coordImporter = Xpetra::ImportFactory<LocalOrdinal, GlobalOrdinal, Node>::Build(coords->getMap(), Amat->getColMap());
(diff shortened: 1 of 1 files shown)


mayrmt:
@pohm Could we test this by maybe comparing test output with some gold-file-test-output, so that we can at least detect, if the output changes over time?

jhux2 (commented):
@pohm01 Thanks for looking at this.  How big is the matrix you're testing against?  I would like to establish a small test to make sure this fix "sticks".

pohm01:
> @pohm01 Thanks for looking at this. How big is the matrix you're testing against? I would like to establish a small test to make sure this fix "sticks".

@jhux2 I've been looking at a 10x10x10 mesh, which seems like a reasonable test size.

jhux2:
@pohm01 Are you planning to add a test with this PR, or separately?  If the latter, I will approve so this merges.

pohm01:
@jhux2 I was working on a test for this PR, but I suppose that doing it separately might be more efficient as I am still figuring out the tests.

jhux2 (approved):
lgtm

PR #9032, opened 2021-04-19
Tpetra: Remove some MultiVector that were checking modification state

tasmith4:
@trilinos/tpetra

## Motivation
We don't see a use case to check the modification state if we're going to use access tags that enforce correct modify/sync behavior internally anyway.


brian-kelley:
@tasmith4 I think you have to change the body of ``MultiVector::runKernelOnHost()``, since it relies on these functions.

packages/tpetra/core/src/Tpetra_BlockMultiVector_decl.hpp
```diff
@@ -474,26 +474,25 @@ class BlockMultiVector :
     mv_.sync_device();
   }
 
-#endif // TPETRA_ENABLE_DEPRECATED_CODE
-
   //! Whether this object needs synchronization to the given memory space.
   template<class TargetMemorySpace>
+  //TPETRA_DEPRECATED
```
brian-kelley, line 479: Is this commenting out intentional, to avoid spamming app developers with warnings?

tasmith4: that's correct.  Geoff did it already for the rest of the things we deprecated on Friday, I believe

kddevin (approved):

Issue #8930, opened 2021-03-29
Tpetra:  Track CrsGraph/CrsMatrix refactoring for UVM removal


kddevin:
## Tracking UVM-requirement removal from CrsGraph and CrsMatrix
@trilinos/tpetra 

Code is in branch **trilinos/tpetraCrsRefactor**

This issue tracks progress on removal of UVM from Tpetra's CrsGraph and CrsMatrix.

For information on Tpetra's MultiVector and BlockMultivector, see #8591 
Other related (and perhaps outdated) issues include #8426, #8345, #8305, #8286, #8209, #8151 

Currently failing Tpetra tests with UVM=OFF and DEPRECATED_CODE=ON

BlockCrsMatrix not yet addressed:

- [ ] 29:TpetraCore_BlockCrsMatrix_MPI_4    
- [ ] 268:TpetraCore_BlockCrsPerfTest_MPI_4

Under investigation

- [ ] 38:TpetraCore_iallreduce_MPI_4
- [ ] 42:TpetraCore_idot_MPI_4

Need merge of https://github.com/kokkos/kokkos-kernels/issues/917 into Trilinos:

- [x] 81:TpetraCore_CrsMatrix_UnitTests_Swap_MPI_2
- [x] 120:TpetraCore_FECrsMatrix_UnitTests_MPI_4
- [x] 211:TpetraCore_MatrixMatrix_UnitTests_MPI_4

Need refactor of FEMAssembly example; should not impact Tpetra use by solvers:

- [ ] 263:TpetraCore_FEMAssembly_InsertGlobalIndicesFESP_MPI_4
- [ ] 264:TpetraCore_FEMAssembly_InsertGlobalIndicesFESPKokkos_MPI_4
- [ ] 265:TpetraCore_FEMAssembly_TotalElementLoopSP_MPI_4
- [ ] 266:TpetraCore_FEMAssembly_TotalElementLoopSPKokkos_MPI_4

Still to do:

- [x] Builds with DEPRECATED_CODE=OFF
- [x] Handling of complex (std::complex vs Kokkos::complex)
- [ ] UVM=ON testing of Trilinos stack
- [ ] Clean up "KDDKDD UVM" notes still in code (esp wrapping in ArrayView in Tpetra_CrsMatrix_def.hpp

PR #9038, opened 2021-04-20
Tpetra:  UVM Removal branch with CrsGraph, CrsMatrix and Multivector

kddevin:
@trilinos/tpetra 

## Motivation

#8930 #8821 et al.
PR submitted to enable PR testing; please do not merge yet.

## Testing

Extensively tested with CUDA, UVM=ON/OFF, DEPRECATED_CODE=ON/OFF on ascicgpu machines.


srajama1:
Amesos2 and Shylu changes looks pretty harmless to me. @ndellin might want to check the Amesos2 changes once. Especially, I am surprised that it is always hostview call, I thought there will be a device view call for cuSolver for example. May be that is because we maintain a copy ?

csiefer2:
Complex is evil.  Looking at Ifpack2...

csiefer2:
Should compile now w/o UVM.

csiefer2:
@trilinos/framework  Well, the GCC 8.3.0 test failures can't be reproduced via the instructions.  Lovely.

kddevin:
I am fixing the failing tests now.

packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp
```diff
@@ -131,11 +132,20 @@ class WrappedDualView {
...
+        deviceView.data() != nullptr && deviceView.use_count() == 0,
+        std::invalid_argument,
+        "Tpetra::Details::WrappedDualView: cannot construct with a device view that\n"
+        "does not own its memory (i.e. constructed with a raw pointer and dimensions)\n"
+        "because the WrappedDualView needs to assume ownership of the memory.");
+    //If the provided view is default-constructed (null, 0 extent, 0 use count),
+    //leave the host mirror default-constructed as well in order to have a matching use count of 0.
     HostViewType hostView;
-    if (deviceView.data() != nullptr) {
+    if(deviceView.use_count() != 0)
```
tasmith4, line 144: Thanks for fixing this, Brian!  Can you also add unit tests to cover this logic so we don't forget why it's here and take it out later?  Also can you add a unit test ensuring the exception will be thrown under the right conditions?

brian-kelley: @tasmith4 I just added these. For the first one, if you construct a normal Kokkos::View with a label and zero extent, it never ends up with ``data() == nullptr``. But with ``View<Sacado::MP::Vector>``, it does - there is a lot of custom View code in there that I don't fully understand. So that test had to go in Stokhos.

csiefer2 (approved):

csiefer2:
@william76 's terrible diagram strikes!  Adding retest

csiefer2:
Wow.  Those GCC 8.3.0 test fails are weird.  Let's hope that is just the machine...

csiefer2 (approved):

kyungjoo-kim (approved):

PR #9046, opened 2021-04-23
Adjust default parameters of GmresSingleReduce

iyamazaki:
@trilinos/belos 

## Motivation
This PR adjusts the default parameters for GmresSingleReduce in belos/tpetra (to make it more robust/efficient):
- Previously, single-reduce CGS with Newton basis was the default
- Now, single-reduce MGS with delayed normalization is the default, without Newton basis


cgcgcg (approved):
Approving, so that the autotester can run.

Issue #9070, opened 2021-04-28
Kokkos:  Nalu diffs due to recent commit


wjhorne:
## Bug Report
@trilinos/kokkos 

### Description
Nalu ( @spdomin ) is currently producing diffs due to recent changes in Trilinos. The output of a bisect is shown below indicating the culprit is commit  [86c737da8545db97f51ae9c77197552e130cbe87] by @ndellingwood . 

### Steps to Reproduce
1. Fetch commit:  [86c737da8545db97f51ae9c77197552e130cbe87] 
2. Build with Configure script: 
[do-configTrilinos_release.TXT](https://github.com/trilinos/Trilinos/files/6393410/do-configTrilinos_release.TXT)
3. Build most current Nalu version and link to Trilinos
4. Run ctest 

### Bisection output

git bisect start
bad: [354848357b8ae65aed2e8e97c15c0e014897bb82] Merge Pull Request #9059 from trilinos/Trilinos/master_merge_20210427_000552
good: [464028f45c1fc1ff10d5c59e622aeb64574e3a96] make "delayed normalization" default for single-reduce MGS
good: [ec0f9a38a6b2c18e1ab96343a47bc99d8d278740] Merge Pull Request #8338 from trilinos/Trilinos/master_merge_20201112_000618
good: [d1f9a8cde3a21e9a1761d1c303b4f668d9dbe45b] Merge Pull Request #8639 from trilinos/Trilinos/master_merge_20210126_000547
good: [f23f4d3b010aeda367bf91bb941b1f1ae182618d] Merge Pull Request #9012 from trilinos/Trilinos/master_merge_20210415_000556
good: [cb6ccbee88da03f9d346b4efddf35f9471fea989] Merge Pull Request #9048 from trilinos/Trilinos/master_merge_20210424_000555
good: [8482d7991ae7f879806fa2ddd4efd52091268b16] Snapshot of kokkos.git from commit 04b8196e0e3bfc4cee4047dbbbb13fc227730fe8
bad: [ad38bd074a74b0abdc5fbcb3f5e6ed93c7eb1ceb] Merge pull request #9054 from trilinos/kokkos-promotion
good: [f8f33cbb111dbf50a0724a0eac970b2bb7e3bea9] Merge pull request #9046 from iyamazaki/peeks-2
[...]

PR #9074, opened 2021-04-28
KokkosKernels: Fix bug in Serial specialization of spmv

crtrott:
Will only hit spmvs with a beta of exactly -1 using the Serial backend.

@trilinos/kokkos-kernels 

Bug reported by some apps after the last release. Could be the reason for: #9070


brian-kelley (approved):

rppawlo:
@crtrott - That fixed it. Thanks!

crtrott:
@trilinos/framework any idea what the python failure there was?

PR #9086, opened 2021-05-03
Tpetra: Allow enabling SYCL backend in Tpetra

masterleinad:
@trilinos/tpetra

## Motivation
This pull request allows enabling Kokkos' SYCL backend in `Tpetra` to allow `Trilinos` running on Intel GPUs.

Currently, it needs https://github.com/kokkos/kokkos/pull/4012, https://github.com/kokkos/kokkos/pull/4007, https://github.com/kokkos/kokkos/pull/3983, https://github.com/kokkos/kokkos/pull/3998 and https://github.com/kokkos/kokkos-kernels/pull/959 (all relevant changes also contained in this pull request) to build `Trilinos` with `Tpetra` enabled and running the unit tests. At the moment,
- `ImbalancedRowMatrix`,
- `assembleElement`, and
- `MatMat`

are failing and have been disabled using preprocessor macros and marking the relevant places with `FIXME_SYCL`.

## Testing
Tested on a CPU and on a Intel GPU passing all tests (with some disabled as described above).


packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
```diff
@@ -284,7 +284,7 @@ struct UnpackCrsMatrixAndCombineFunctor {
 
     if (expected_num_bytes > num_bytes)
     {
-      printf(
+      KOKKOS_IMPL_DO_NOT_USE_PRINTF(
```
kddevin, line 287: I prefer not to use KOKKOS_IMPL details in Tpetra.  Is there a problem with printf and SYCL?  What does KOKKOS_IMPL_DO_NOT_USE_PRINTF do?  How would one know to use this macro rather than printf?

packages/tpetra/core/src/Tpetra_Details_StaticView.cpp
```diff
@@ -265,6 +292,60 @@ resize (Kokkos::Experimental::HIPHostPinnedSpace /* space */,
 
 #endif // KOKKOS_ENABLE_HIP
 
+#ifdef KOKKOS_ENABLE_SYCL
```
kddevin, line 295: Tpetra has different implementations for different architectures.  Should these implementation actually be in Kokkos, rather than in Tpetra?

packages/tpetra/core/src/Tpetra_Details_StaticView.cpp
```diff
@@ -123,6 +123,33 @@ void finalize_hip_host_pinned_memory ()
 }
 #endif // KOKKOS_ENABLE_HIP
 
+#ifdef KOKKOS_ENABLE_SYCL
```
kddevin, line 126: Tpetra has different implementations for different architectures.  Should these implementations actually be in Kokkos rather than Tpetra?

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
```diff
@@ -471,7 +471,8 @@ int main (int argc, char *argv[])
       {
         TimeMonitor timerMultiVectorFill(*TimeMonitor::getNewTimer("4) MultiVectorFill"));
 
-        auto value = X->getLocalView<typename exec_space::memory_space>(Tpetra::Access::OverwriteAll);
+        using device_type = typename tpetra_multivector_type::node_type::device_type;
+        auto value = X->getLocalView<device_type>(Tpetra::Access::OverwriteAll);
```
kddevin, line 475: OK, but prefer use of getLocalViewDevice(Tpetra::Access::OverwriteAll).

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
```diff
@@ -238,7 +238,7 @@ int main (int argc, char *argv[])
     // uses a different device. An easy check is device configuration for
     // 4 mpi nodes as the typical number of GPUs per node is 4.
     if (commTest->getRank () < 4) {
-      exec_space::print_configuration (std::cout, false);
+      exec_space{}.print_configuration (std::cout, false);
```
kddevin, line 241: What is this change doing?  I don't understand the curly braces.

kddevin (changes requested):
Thanks, @masterleinad , for these changes.
I looked at the Tpetra changes only, and request a few changes or explanations.
It would be good for @brian-kelley  to review as well.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
brian-kelley: It's constructing an exec_space instance like (), since print_configuration is supposed to be non-static

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
masterleinad: Oh, I should have pointed this out directly. `SYCL` doesn't support `printf` on the device. In `Kokkos` we are defining a workaround to make this work regardless with the intention of only using it internally (such that it could be removed at any point). The alternative is to just guard these places with `#ifndef KOKKOS_ENABLE_SYCL` or similar.

packages/tpetra/core/src/Tpetra_Details_StaticView.cpp
masterleinad: I was just repeating what was done above for `Cuda` and `HIP`. If you want this to be refactored, it should be done in a separate pull request IMHO.

packages/tpetra/core/src/Tpetra_Details_StaticView.cpp
masterleinad: Also, this is the same as for `Cuda` and `HIP`.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
crtrott: Its the default constructor. I.e. call the member function instead of the static function. The previous thing was actually wrong, just happens to work with most old backends but I think we technically only allow member functions for that. 

Curly braces is just another mode of construction in C++, one which is generally now preferred over ().

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
crtrott: yes there is a problem with printf in SYCL. Namely that it isn't supported currently and you have to call sycl::printf or something like that ... 
We are talking to various compiler folks to get that changed, and thus didn't want to add an official portability functionality like KOKKOS_PRINTF yet. But without changing it this guy won't compile/work with SYCL right now.

packages/tpetra/core/src/Tpetra_Details_StaticView.cpp
crtrott: technically you probably want to refactor all this in a way that you just support Kokkos::DefaultExecutionSpace and Kokkos::DefaultHostExecutionSpace (we talked about that in the Tpetra meeting). This is not something which should be in Kokkos, I mean what you create here are Tpetra data structures not Kokkos's. You could template them or as I said just create them for the default execution spaces, which then are the correct ones based on the Kokkos configurations you build with.

crtrott:
Btw. we are looking into making all the cahnges necessary for Kokkos/KokkosKernels part of the 3.4.1 patch release. So we wouldn't necessarily want you to go ahead and merge this before that is out and in Trilinos, and which point the changes are more or less confined to Tpetra.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp
masterleinad: See [`3221242` (#9086)](https://github.com/trilinos/Trilinos/pull/9086/commits/3221242de7962a85ffddda5c3dc0c0c66f655540).

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
```diff
@@ -132,6 +135,14 @@ namespace DefaultTypes {
   };
 #endif
 
+#ifdef KOKKOS_ENABLE_SYCL
+  template<>
+  struct CommBufferMemorySpace<Kokkos::Experimental::SYCLSharedUSMSpace>
```
brian-kelley, line 140: This should actually be specialized on SYCL the execution space. The HIPHostPinnedSpace above is wrong, it should be HIP.

brian-kelley (changes requested):
The only issue I see is the CommBufferMemorySpace thing is templated on execution space so the specialization should use SYCL, not the SYCLSharedUSMSpace. HIP is wrong above, I found that yesterday and will fix in another PR.

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
kddevin: Probably doing #ifndef KOKKOS_ENABLE_SYCL around the printfs is better than using a Kokkos implementation detail.  Plus, it documents why we are doing something different from printf -- that is, that SYCL alone doesn't like it.  When Kokkos adds an official portable KOKKOS_PRINTF, we will adopt it.

kddevin (changes requested):
Thanks for your explanations.  I resolved most of my conversations.
I'd still like to see KOKKOS_IMPL_DO_NOT_USE_PRINTF replaced with #ifndef KOKKOS_ENABLE_SYCL.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
masterleinad: Fixed in [`68f57e3` (#9086)](https://github.com/trilinos/Trilinos/pull/9086/commits/68f57e35ee3f00bae7a4de6fd9639c84754e8bc0).

packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
masterleinad: Fixed in [`45fefc7` (#9086)](https://github.com/trilinos/Trilinos/pull/9086/commits/45fefc7f17a789cc870b1b1e1525157ccf6126df).

brian-kelley (dismissed):
Thanks for making that change.

kddevin (approved):

packages/tpetra/CMakeLists.txt
```diff
@@ -529,6 +529,29 @@ ELSE () # NOT Tpetra_INST_HIP
     MESSAGE(STATUS "NOTE: Kokkos::HIP is ON (the CMake option Kokkos_ENABLE_HIP is ON), but the corresponding Tpetra Node type is disabled.  If you want to enable instantiation and use of Kokkos::HIP in Tpetra, please also set the CMake option Tpetra_INST_HIP:BOOL=ON.  If you use the Kokkos::HIP version of Tpetra without doing this, you will get link errors!")
   ENDIF ()
 ENDIF () # Tpetra_INST_HIP
+
+# Kokkos::SYCL (Kokkos::Compat::KokkosSYCLWrapperNode)
+GLOBAL_SET(HAVE_TPETRA_INST_SYCL_DEFAULT OFF)
```
lucbv, line 534: It would make sense to move this statement further up around line 365 and you might want to add an `ASSERT_DEFINED(Kokkos_ENABLE_SYCL)`

lucbv (commented):
I just had a small comment regarding the top CMakeList.txt in tpetra, it's nothing big but we should try to be uniform if possible.

packages/tpetra/CMakeLists.txt
masterleinad: See [`a2f73a1` (#9086)](https://github.com/trilinos/Trilinos/pull/9086/commits/a2f73a1a96516642b58fd60078cc4d74f08678c0).

masterleinad:
Since Kokkos 3.4.1 has been merged, this pull request is ready for (re-)review.

lucbv:
@masterleinad good point, where do you want to build/test this, the auto-tester is not really geared for that new backend...

masterleinad:
> @masterleinad good point, where do you want to build/test this, the auto-tester is not really geared for that new backend...

SYCL support can be tested on a CPU or on an `Nvidia` GPU (like we do in `Kokkos`). I am happy to help providing a configuration or help with setting that up in a different way if that helps.

brian-kelley (approved):

lucbv (approved):
Look good to me, thanks for taking the time to get this in Tpetra!

masterleinad:
I rebased to resolve the merge conflicts and fixed the resulting problems (making sure that all the tests pass again on Intel GPUs).

brian-kelley (approved):

kddevin (approved):
Thanks, @masterleinad

lucbv:
Thanks @masterleinad for taking care of this, @crtrott and I are also OK with the state of the PR so I am merging it now.

PR #9090, opened 2021-05-04
Tpetra: enable and fix tests for HIP

brian-kelley:
@trilinos/tpetra 

## Motivation

Enables the tests that used to crash AMD GPU nodes, and get all tests passing for rocm 4.2. I think the workarounds to do this aren't too hacky and can be permanent. For example, replacing the Tpetra local CRS transpose (which was still hanging on the GPU) with a call to KokkosKernels transpose.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows #8938 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback

Getting Tpetra ready for Frontier/El Capitan.
## Additional Information

One thing we will want to do eventually is refactor the TpetraExt MatrixMatrix kernels file that specialize for different backends - this adds one for HIP that is basically a copy-paste of the Cuda file. We could probably just have one for all GPUs and one for threads/OpenMP. There are also very implementation-specific options you can pass like "cuda: algorithm" and "cuda: team work size" for SpGEMM. I didn't replace these with "hip: ...", but they could probably be deprecated and replaced with more portable names in the future.


brian-kelley:
Note: the build where everything passes has ``-D Tpetra_INST_COMPLEX_DOUBLE:BOOL=OFF`` and ``-D Tpetra_ENABLE_EXAMPLES:BOOL=OFF``. There are 258 total tests that all pass. Complex support is waiting on https://github.com/kokkos/kokkos/issues/3974 and possibly follow up work in KokkosKernels.

crtrott:
Can we add the KokkosKernels change to the 3.4.1 release for that?

brian-kelley:
@crtrott Sure. I could also work around the issue with just Tpetra changes, should I do that?

crtrott:
Talk to Luc and Siva, I am fine either way.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_HIP.hpp
```diff
@@ -0,0 +1,781 @@
...
+
+#ifndef TPETRA_MATRIXMATRIX_HIP_DEF_HPP
+#define TPETRA_MATRIXMATRIX_HIP_DEF_HPP
+
+#ifdef HAVE_TPETRA_INST_HIP
+namespace Tpetra {
+namespace MMdetails {
+
+/*********************************************************************************************************/
+// MMM KernelWrappers for Partial Specialization to HIP
```
kddevin, line 50: Can you say (in the comments) what changed from other specializations to accommodate HIP?
(I don't know this code well enough to spot differences or errors.)
Longer term (not this PR), it might be good to refactor this code and all its specializations; I noticed a lot of duplication when doing UVM removal.

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
```diff
@@ -590,8 +587,8 @@ makeColMap (Teuchos::RCP<const Tpetra::Map<LO, GO, NT>>& colMap,
   using device_t = typename NT::device_type;
   using exec_space = typename device_t::execution_space;
   using memory_space = typename device_t::memory_space;
-  using bitset_t = Kokkos::Bitset<device_t>;
-  using const_bitset_t = Kokkos::ConstBitset<device_t>;
+  using bitset_t = Kokkos::Bitset<exec_space>;
```
kddevin, line 590: What determines whether to use device_t or exec_space?

packages/tpetra/core/src/Tpetra_RowMatrixTransposer_def.hpp
```diff
@@ -183,81 +184,9 @@ createTransposeLocal (const Teuchos::RCP<Teuchos::ParameterList>& params)
...
-        LO* cols_beg = t_cols.data () + beg;
-        IST* vals_beg = t_vals.data () + beg;
-        shellSortKeysAndValues (cols_beg, vals_beg, len);
-      });
-  }
-
-  local_matrix_type lclTransposeMatrix ("transpose", lclNumCols,
-                                        lclNumRows, nnz,
-                                        t_vals, t_offsets, t_cols);
+  local_matrix_type lclTransposeMatrix = KokkosKernels::Impl::transpose_matrix(lclMatrix);
```
kddevin, line 187: Good!  Definitely should use KokkosKernels wherever possible.
Are we certain that this code is exercised in a test?

packages/tpetra/core/test/CrsMatrix/CMakeLists.txt
```diff
@@ -88,17 +88,15 @@ TRIBITS_ADD_EXECUTABLE_AND_TEST(
...
-  TRIBITS_ADD_EXECUTABLE_AND_TEST(
-    CrsMatrix_2DRandomDist
-    SOURCES
-      CrsMatrix_2DRandomDist.cpp
-    COMM serial mpi
-    PASS_REGULAR_EXPRESSION "PASS"
-    FAIL_REGULAR_EXPRESSION "FAIL"
-    )
-ENDIF()
+# This test causes AMD MI25 to permanently crash (until reboot)
```
kddevin, line 91: Do we still need this comment if we are enabling the test?  
Do we know why this test crashed the AMD?

packages/tpetra/core/example/Finite-Element-Assembly/fem_assembly_typedefs.hpp
```diff
@@ -71,18 +72,18 @@ using fe_multivector_type = Tpetra::FEMultiVector<Scalar>;
 
 
 using global_ordinal_view_type =
-  Kokkos::View<global_ordinal_type*, execution_space>;
+  Kokkos::View<global_ordinal_type*, device_type>;
```
kddevin, line 75: What determines whether to use execution_space or device_type?

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
```diff
@@ -496,14 +496,12 @@ makeColMap (Teuchos::RCP<const Tpetra::Map<LO, GO, NT> >& colMap,
   return errCode;
 }
 
-template<typename GO, typename device_t>
+template<typename GOView, typename bitset_t>
```
kddevin, line 499: These changes are fine. 
Were they needed for HIP, or are they just cleaner code?  If needed for HIP, why?

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrsMeshDatabase.hpp
```diff
@@ -314,7 +316,7 @@ namespace BlockCrsTest {
     StructuredProcGrid _grid;
     StructuredBlockPart _owned;
 
-    typedef Kokkos::View<GO*,exec_space> global_ordinal_view_type;
+    typedef Kokkos::View<GO*,mem_space> global_ordinal_view_type;
```
kddevin, line 319: What determines whether to use exec_space vs mem_space?  Why use mem_space here and not device_type as in other code changes?

kddevin (changes requested):
Thanks, @brian-kelley .  I looked at the Tpetra code.  I requested only some changes to comments.  I asked some questions regarding device types and spaces mostly for my own education.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_HIP.hpp
brian-kelley: Nothing at all changed, it's just a copy-paste and then replace "KokkosCudaWrapperNode" with "KokkosHIPWrapperNode". It is a lot of duplication, but that's kind of unavoidable when the different backends are selected by template specialization. We can fix the duplication later with enable_if or making these specialized functions just thin wrappers over generic algorithms.

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
brian-kelley: device_t is an execution space/memory space pair. exec_space is just the execution space, so this bitset_t is now allocating into the execution space's default memory space (in this case of HIP, HIPSpace). This is required because most atomics in HIPHostPinnedSpace (like the bitwise OR needed by Bitset) cause the GPU to hang.

packages/tpetra/core/src/Tpetra_RowMatrixTransposer_def.hpp
brian-kelley: Hmm, I thought it was tested in KokkosKernels but it's not. It is used by the spgemm test, but never tested directly.

However, Tpetra RowMatrixTransposer is tested so this code is covered.

packages/tpetra/core/test/CrsMatrix/CMakeLists.txt
brian-kelley: No, good catch, that comment doesn't apply to rocm 4.2 and later.

packages/tpetra/core/example/Finite-Element-Assembly/fem_assembly_typedefs.hpp
brian-kelley: Here, this needs to be device_type in general so that it gets both execution space and memory space from the Node type. In the case of HIP, this means it's allocated in HIPHostPinnedSpace so that it's host accessible. Important since this test still relies on UVM. It also means that when we don't need UVM anymore, there's only one place to switch to HIPSpace (the KokkosHIPWrapperNode typedef).

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
brian-kelley: Cleaner code. It was easier to just template on the bitset type than reconstruct it with a typedef in the functor.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrsMeshDatabase.hpp
brian-kelley: Yes, this should be device_type for consistency with the other changes. mem_space and device_type are equivalent here but it's good to be consistent.

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
brian-kelley: BTW, this bitset is only used locally in this function, and only in kernels executing on the exec_space. So this won't cause any issues with type mismatches on other backends.

brian-kelley:
@crtrott Luc and I agreed it's fine to wait for 3.4.1. So I will just remove the KokkosKernels changes here.

lucbv (commented):
Just a few comments but nothing major, thanks for cleaning things up and making sure the tests are all passing!

packages/tpetra/core/test/CrsMatrix/CMakeLists.txt
crtrott: Yes, this is a bug in the AMD software stack prior to ROCM 4.2, ROCM 4.2 is available at ORNL, and will be publicly released in the next couple weeks. We will shift minimum version of ROCM to 4.2 immediately with Kokkos since it fixes that really bad bug which has bothering us more than a year.

packages/tpetra/core/src/Tpetra_RowMatrixTransposer_def.hpp
```diff
@@ -183,81 +184,9 @@ createTransposeLocal (const Teuchos::RCP<Teuchos::ParameterList>& params)
...
+  if (sort)
+    KokkosKernels::Impl::sort_crs_matrix(lclTransposeMatrix);
```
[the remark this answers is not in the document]
brian-kelley: The changes that moved sorting out of Impl haven't merged yet. I guess they might make it into 3.4.1 - if they do I'll change this. Otherwise it's not urgent - at the next promotion, this will start throwing deprecated warnings and we can fix it.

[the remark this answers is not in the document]
brian-kelley: I didn't really think about performance here. Tpetra and KokkosKernels had basically the same transpose algorithm, and the KokkosKernels sorting should generally be faster, but I will check it.

jjellio:
It will probably be tomorrow before I can spin off the build to test this PR on Redwood.

brian-kelley:
@jjellio Since I just reverted the necessary KokkosKernels changes, this PR won't quite work currently. If you wait for the 3.4.1 minor kokkos promotion it will work, or I can give you a patch if you want

lucbv:
@brian-kelley can you add a small change in `tpetra/CMakeList.txt`, can you move line 512 which reads:
```
GLOBAL_SET(HAVE_TPETRA_INST_HIP_DEFAULT off)
```
to line 365 where similar definitions are made for the other backends.
We should also probably add `ASSERT_DEFINED(Kokkos_ENABLE_HIP)` around line 369 and maybe move the assertion for `Kokkos_ENABLE_CUDA` next to it for uniformity.

lucbv:
Also I believe the `HAVE_TPETRA_INST_HIP_DEFAULT off` should be replaced with `HAVE_TPETRA_INST_HIP_DEFAULT OFF`

brian-kelley:
@lucbv Kokkos_ENABLE_CUDA was checked at the top of the file because a lot of CUDA-related logic came right after it, so I just moved the other assert_defined(Kokkos_ENABLE_XYZ) lines to there.

lucbv:
Totally fine, it seemed that the file had some structure so we might as well keep it.
Thanks for the change!

packages/tpetra/core/src/Tpetra_Details_makeColMap_def.hpp
masterleinad: It still feels a little weird to me to specify an execution space instead of a memory space. I would prefer the latter.

masterleinad (commented):
This touches a lot of places similarly to https://github.com/trilinos/Trilinos/pull/9086 :slightly_smiling_face: Hopefully, this will give enough merge conflicts so that we are forced not to forget to sync some of the restructurings that are done here or in https://github.com/trilinos/Trilinos/pull/9086.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
```diff
@@ -126,7 +126,7 @@ namespace DefaultTypes {
 
 #ifdef KOKKOS_ENABLE_HIP
   template<>
-  struct CommBufferMemorySpace<Kokkos::Experimental::HIPHostPinnedSpace>
+  struct CommBufferMemorySpace<Kokkos::Experimental::HIP>
   {
     using type = Kokkos::Experimental::HIPHostPinnedSpace;
```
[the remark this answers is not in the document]
brian-kelley: It should be HIPHostPinned for now, until we know that HIP-aware MPI support is good on the testbeds. In the long term, you're right, it will be HIPSpace (like CudaSpace is used for Cuda).

brian-kelley:
@masterleinad The idea with ``Bitset<exec_space>`` is just "give me a bitset that I can use in a kernel executing here". Since bitset doesn't work in HIPHostPinnedSpace (https://github.com/kokkos/kokkos/issues/3996), this is the only way I could think of that doesn't need a special case for any backend.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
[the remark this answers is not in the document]
masterleinad: OK. I thought GPU-aware MPI is handled somewhere else and by default host buffers are used. Am I wrong here? In that case I should also change to an host-accessible memory space in #9086.

masterleinad:
> The idea with `Bitset<exec_space>` is just "give me a bitset that I can use in a kernel executing here". Since bitset doesn't work in HIPHostPinnedSpace ([kokkos/kokkos#3996](https://github.com/kokkos/kokkos/issues/3996)), this is the only way I could think of that doesn't need a special case for any backend.

Oh, I was thinking about passing `execution_space::memory_space` instead of `just `execution_space` but I don't feel strongly about it.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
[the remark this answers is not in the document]
brian-kelley: @masterleinad Yeah, SYCL should use a host-accessible space for now too. Right now there's only logic throughout the code for CUDA-aware MPI specifically. There will be more work ahead to make it general across HIP/SYCL/OpenMPTarget.

brian-kelley:
@masterleinad I like ``execution_space::memory_space``, I just changed it to that.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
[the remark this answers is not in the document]
masterleinad: OK. I switched to `SYCLSharedUSMSpace` in [`d69c5f4` (#9086)](https://github.com/trilinos/Trilinos/pull/9086/commits/d69c5f457896113d48a385b4b16b70be620c983f).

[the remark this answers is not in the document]
jjellio: I'm not sure Pinned memory (for comm) is the right for testing out HIP w/tpetra.  That will trigger non-standard code paths - we want to exercise HIPSpace primarily - and for  Comm, then set `TPETRA_ASSUME_CUDA_AWARE_MPI=0` (or however it's spelled) - that's a better path imo.  This would mimic how Cuda was R&D'd - Tpetra *never* used UVM w/MPI - Mark was adamant about that.

Furthermore, app teams started messing around with Pinned, because they thought it would improve performance (it generally didn't - and their motivation was based on a lack of understanding of the MPI runtime (i.e., we had cuda-aware MPI ON, which is inherently slower).

When we had the discussion about the FECrs error - I wasn't aware that the default comm buffer was being put as Pinned... We should determine if the error remains if Tpetra is not cuda-aware and the comm buffer is device memory.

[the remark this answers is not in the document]
brian-kelley: @jjellio OK, that's a good idea. I'll see how much is needed to put in HIPSpace (maybe not much, ``Behavior::assumeMpiIsCudaAware()`` is only called in a dozen places or so).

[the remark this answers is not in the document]
brian-kelley: @masterleinad Did you see James's comment here? It applies to the SYCL PR as well. I thought there would be issues with MPI CUDA-aware codepaths being taken, but actually Tpetra decides at configure time that since CUDA is not enabled, MPI is not CUDA-aware. It's kind of doing the right thing for the wrong reason in HIP. But it will work until "cuda-aware" options are replaced with portable  "device-aware MPI". I switched CommBufferMemorySpace to HIPSpace and all tests still passed with no other changes.

packages/tpetra/core/src/Tpetra_RowMatrixTransposer_def.hpp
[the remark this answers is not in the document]
brian-kelley: OK, this didn't make it into 3.4.1, so it can just wait until the next feature release I guess.

packages/tpetra/core/src/Tpetra_Details_DefaultTypes.hpp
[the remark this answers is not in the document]
masterleinad: OK. I wasn't quite sure if there is more needed to make it work in an expected way apart from switching the memory space again. Anyway, I reverted the corresponding commit, see [`8b59e2f` (#9086)](https://github.com/trilinos/Trilinos/pull/9086/commits/8b59e2f89e13d63f18dd278df5e53bfc72b11bc9).

brian-kelley:
This is ready for a new review. I rebased it today (after Kokkos 3.4.1) and retested. I think I made all the requested changes, aside from moving ``KokkosKernels::Impl::transpose_matrix`` out of ``Impl::``, and adding a local test for it. But those aren't blockers for this PR (Tpetra's transpose is tested).

lucbv (approved):
From my perspective this looks fine now, let us wait for @kddevin and @crtrott to weigh in on this PR.

kddevin (dismissed):
Thanks, @brian-kelley

brian-kelley:
This is now rebased on top of the SYCL PR #9086, so I think it's ready for final review and merge.

lucbv (approved):
Thanks @brian-kelley the rebase looks good, let's wait for the auto-tester to do its job

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_HIP.hpp
```diff
@@ -0,0 +1,781 @@
...
+
+  // Node-specific code
+  using Teuchos::RCP;
+
+  // Options
+  //int team_work_size = 16;  // Defaults to 16 as per Deveci 12/7/16 - csiefer // unreferenced
+  std::string myalg("KK");
+  if(!params.is_null()) {
+    if(params->isParameter("cuda: jacobi algorithm"))
+      myalg = params->get("cuda: jacobi algorithm",myalg);
```
masterleinad, line 430:
```suggestion
    if(params->isParameter("hip: jacobi algorithm"))
      myalg = params->get("hip: jacobi algorithm",myalg);
```

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_HIP.hpp
```diff
@@ -0,0 +1,781 @@
...
+  scalar_view_t valuesC;
+
+  // Options
+  int team_work_size = 16; 
+  std::string myalg("SPGEMM_KK_MEMORY");
+  if(!params.is_null()) {
+    if(params->isParameter("cuda: algorithm"))
+      myalg = params->get("cuda: algorithm",myalg);
+    if(params->isParameter("cuda: team work size"))
+      team_work_size = params->get("cuda: team work size",team_work_size);
```
masterleinad, line 722:
```suggestion
    if(params->isParameter("hip: algorithm"))
      myalg = params->get("hip: algorithm",myalg);
    if(params->isParameter("hip: team work size"))
      team_work_size = params->get("hip: team work size",team_work_size);
```

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_HIP.hpp
```diff
@@ -0,0 +1,781 @@
...
+  //typedef typename graph_t::row_map_type::const_type lno_view_t_const;
+
+  // Options
+  int team_work_size = 16;  // Defaults to 16 as per Deveci 12/7/16 - csiefer
+  std::string myalg("SPGEMM_KK_MEMORY");
+  if(!params.is_null()) {
+    if(params->isParameter("cuda: algorithm"))
+      myalg = params->get("cuda: algorithm",myalg);
+    if(params->isParameter("cuda: team work size"))
+      team_work_size = params->get("cuda: team work size",team_work_size);
```
masterleinad, line 173:
```suggestion
    if(params->isParameter("hip: algorithm"))
      myalg = params->get("hip: algorithm",myalg);
    if(params->isParameter("hip: team work size"))
      team_work_size = params->get("hip: team work size",team_work_size);
```

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_HIP.hpp
brian-kelley: @masterleinad Good suggestion, I changed all these lines in one commit.

masterleinad (dismissed):

kddevin (approved):
Still LGTM

PR #9092, opened 2021-05-04
Set non-cache vars for CMAKE_<LANG>_FLAGS with user override (#8449)

### LINKED ISSUE #8449: ATDM Config: Change how compiler flags are handled
CC: @rppawlo, @trilinos/framework 

## The Problem

The current ATDM Trilinos configuration setup does not let the user pass in compiler flags using cache vars `CMAKE_CXX_FLAGS` without overriding the few special flags that the ATDM Trilinos configuration is setting like `--remove-duplicate-link-files` for CUDA+RDC builds and `-fPIC` for `fpic` builds (see https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md#build-name-keywords).  The current implementation for setting these variables is shown in:

https://github.com/trilinos/Trilinos/blob/e29f0be643de0d55a202263ef1da8d4bc9c8001c/cmake/std/atdm/ATDMDevEnvSettings.cmake#L227..L236

which currently shows:

```
ATDM_SET_CACHE(CMAKE_C_FLAGS
  "$ENV{ATDM_CONFIG_C_FLAGS} $ENV{EXTRA_C_FLAGS}" CACHE STRING)
ATDM_SET_CACHE(CMAKE_CXX_FLAGS
  "$ENV{ATDM_CONFIG_CXX_FLAGS}${EXTRA_EXTRA_CXX_FLAGS} $ENV{EXTRA_CXX_FLAGS}" CACHE STRING)
ATDM_SET_CACHE(CMAKE_Fortran_FLAGS
  "$ENV{ATDM_CONFIG_Fortran_FLAGS} $ENV{EXTRA_Fortran_FLAGS}" CACHE STRING)
# NOTE: That above "EXTRA_<LANG>_FLAGS" env vars are left in for backward
# compatibility with the EMPIRE configuration and are not directly set in any
# offical ATDM Trilinos configuration setup.  Only the
# "ATDM_CONFIG_<LANG>_FLAGS" env vars are set.
```

This leads to confusing behavior like resulted in https://github.com/TriBITSPub/TriBITS/issues/341 and [ATDV-415](https://sems-atlassian-srn.sandia.gov/browse/ATDV-415).

From reading the section "14.3. Compiler And Linker Variables" in the book "Professional CMake: A Practical Guide (5th edition)", it seems that the most acceptable way to set these vars is to set local non-cache vars and either prepend or append to the cache vars.

## Suggested Change

Given that I doubt that any EMPIRE developers are still using the env vars `EXTRA_<LANG>_FLAGS`, we should just go with the approach of prepending to the user-set cache vars using project-scope non-cache vars like:

```
SET(CMAKE_CXX_FLAGS
  "$ENV{ATDM_CONFIG_CXX_FLAGS}${EXTRA_EXTRA_CXX_FLAGS} ${CMAKE_CXX_FLAGS}")
```

That way, when the user configures with:

```
-DCMAKE_CXX_FLAGS="<flag1> <flag2> ..."
```

then those flags would come after (can therefore override) any flags being set internally but will not wipe out important flags like `--remove-duplicate-link-files` and `-fPIC`.

## Tasks:

* [ ] Create branch '8449-atdm-compliler-flags-fix'
* [ ] Implement change described above
* [ ] Add section to `atdm/README.md` on how compiler flags work and how to extend and override them
* [ ] Do local testing to ensure that set flags are working (and test adding new flags with `-DCMAKE_CXX_FLAGS=<extra_flags>`)
* [ ] Post PR for branch '8449-atdm-compliler-flags-fix' and get it merged

bartlettroscoe:
This fixes #8449 

This should not change behavior of any of the existing ATDM Trilinos nighty builds.  It only allows users to set their own flags (as per #8449 and [CII Best Practices build_standard_variables](https://bestpractices.coreinfrastructure.org/en/criteria#1.build_standard_variables)).

## How was this tested?

On the machine 'ceerws1113' using the 'cee-rhel7' system, I did:

```
$ cd /scratch/rabartl/Trilinos.base/BUILDS/ATDM/CEE-RHEL7/CHECKIN/

$ ./checkin-test-atdm-cee-rhel7.sh cee-rhel7-clang-asan-dbg \
  --enable-packages=Kokkos,Teuchos,Epetra,Zoltan --configure

$ cd cee-rhel7-clang-asan-dbg/

$ . load-env.sh 
Hostname 'ceerws1113' matches known ATDM host 'ceerws1113' and system 'cee-rhel7'
Setting compiler and build options for build-name 'cee-rhel7-clang-asan-dbg'
Using CEE RHEL7 compiler stack CLANG-9.0.1_OPENMPI-4.0.3 to build DEBUG code with Kokkos node type SERIAL

$ rm -r CMake* 

$ time ./do-configure -DCMAKE_C_FLAGS="-O3" -DCMAKE_CXX_FLAGS="-O4" -DCMAKE_Fortran_FLAGS="-O5" \
   &> configure.out

$ grep -B 1 "[-]O3" build.ninja | head -n 2
  DEP_FILE = packages/zoltan/src/CMakeFiles/zoltan.dir/all/all_allo.c.o.d
  FLAGS = -O3 -g -fPIC

$ grep -B 1 "[-]O4" build.ninja | head -n 2
  DEP_FILE = packages/kokkos/core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_CPUDiscovery.cpp.o.d
  FLAGS = -Wall -Wextra -g -O1 -fsanitize=address -fno-omit-frame-pointer -O4   -g -fPIC -std=c++14

$ grep -B 1 "[-]O5" build.ninja | head -n 2
  DYNDEP_INTERMEDIATE_FILE = packages/epetra/src/CMakeFiles/epetra.dir/Epetra_dcrsmm.f.o.ddi
  FLAGS = -O5 -g -O0 -Jpackages/epetra/src -fPIC
```

What this test shows are that the flags set with:

```
  -DCMAKE_C_FLAGS="-O3" \
  -DCMAKE_CXX_FLAGS="-O4" \
  -DCMAKE_Fortran_FLAGS="-O5" \
```

are correctly being applied to C files, C++ files, and Fortran files, respectively.

It also shows that the flags set with `-DCMAKE_CXX_FLAGS="-O4"` are correctly added **after** the internally-set C++ flags:

```
-Wall -Wextra -g -O1 -fsanitize=address -fno-omit-frame-pointer
```

which are set in the file `Trilinos/cmake/std/atdm/cee-rhel7/environment.sh`.

It is interesting to note the trialing C++ options:

```
  -g -fPIC -std=c++14
```

come from the following sources:

* `-g`: Comes from `CMAKE_CXX_FLAGS_DEBUG`
* `-fPIC`: Comes from setting `CMAKE_POSITION_INDEPENDENT_CODE=ON` in `Trilinos/cmake/std/atdm/ATDMDevEnvSettings.cmake`.
* `-std=c++14`: Comes from setting `CMAKE_CXX_STANDARD=14` in `Trilinos/CMakeLists.txt`

So it is interesting to note that the compiler flags that CMake sets itself can't be overridden with `CMAKE_<LANG>_FLAGS`.   I really don't like that since you are trusting that CMake will always do the right thing.  This is the type of thing that leads people to post-modifying `build.ninja` files after configure completes in some cases :-(


cmake/std/atdm/ATDMDevEnvSettings.cmake
```diff
@@ -222,16 +222,12 @@ IF (ATDM_FPIC)
...
-ATDM_SET_CACHE(CMAKE_Fortran_FLAGS
-  "$ENV{ATDM_CONFIG_Fortran_FLAGS} $ENV{EXTRA_Fortran_FLAGS}" CACHE STRING)
-# NOTE: That above "EXTRA_<LANG>_FLAGS" env vars are left in for backward
-# compatibility with the EMPIRE configuration and are not directly set in any
-# offical ATDM Trilinos configuration setup.  Only the
-# "ATDM_CONFIG_<LANG>_FLAGS" env vars are set.
+SET(CMAKE_C_FLAGS
+  "$ENV{ATDM_CONFIG_C_FLAGS} ${CMAKE_C_FLAGS}")
+SET(CMAKE_CXX_FLAGS
+  "$ENV{ATDM_CONFIG_CXX_FLAGS}${EXTRA_EXTRA_CXX_FLAGS} ${CMAKE_CXX_FLAGS}")
```
rppawlo, line 228: do you need a space in between the first and second arguments on this line?

bartlettroscoe: > do you need a space in between the first and second arguments on this line?

@rppawlo, not in this case given how that var is created above this.  But to be super safe, I guess it might be a good idea to put in a space.

I will push a commit to add that space to be safe.

bartlettroscoe: > I will push a commit to add that space to be safe.

@rppawlo, I just pushed https://github.com/trilinos/Trilinos/pull/9092/commits/3dc3b5019280cd90604ed6d5c199b14e31baa3f7

rppawlo (approved):
Thanks for taking care of this @bartlettroscoe !

PR #9097, opened 2021-05-05
Trilinos: ATDM Dev Settings TPL tweaks and add HIP

jjellio:
@trilinos/framework @bartlettroscoe @e10harvey 

This patch does two things that I need to use HIP w/ATDM.
1.  It fixes ParMETIS include/library paths.  There is an implicit assumption that METIS lives inside PARMETIS's directory. This isn't the case with my builds and the current code does not respect the ATDM_CONFIG variables to set the INCLUDE and LIBRARY dirs. This patch address that, using the existing path that is set as the `default` when trying to read `$ENV{ATDM_CONFIG_...}`.  It then sets the cache var to the result of that (either the default or the ENV).
    - I applied the same logic to METIS.
    -  This allows setting `ATDM_CONFIG_PARMETIS_INCLUDE or LIBRARY_DIRS` (which in my cases contains parmetis path and a separate metis path)
    - This keeps the default value what is was (if the ATDM_CONFIG variables aren't set)
2. I added HIP support (mirroring CUDA)
    -  We lack the definitions of a 'tpl' for HIP and ROCSparse (the analog of cusparse). 
    -   We need to add TPL definitions and have that use `HIP_ROOT`
    - I haven't added the TPLs, because noone is using them (yet) - but they need to be added

This is tested on the Cray El Cap testbed.

It would make sense to have this tested using the ATDM builds here - to make sure the ENV files don't have a bogus METIS or PARMETIS INCLUDE or LIBRARY variable set (and they've been happily depending on ROOT/include ROOT/lib to be correct, which would have masked a broken/old setting on those variables)

This patch is necessary for supporting ATDM builds via SPACK.

@bartlettroscoe - I was sitting on this to see if I found more things I needed to change... but since there have been 2 edits to DevSettings recently, I'd rather push this now (and not risk having a conflict if more changes get made)

@kddevin @srajama1 : I did not 'define' the TPLs backing HIP. I did add the code to enable them (to keep parity with Cuda) - but they won't do anything =)  It seems Kokkos Kernels should probably handle this. 

I also did not set a `hip` LAMDA flag ... What's the plan for that?


brian-kelley (approved):

cmake/std/atdm/ATDMDevEnvSettings.cmake
```diff
@@ -422,8 +432,15 @@ ENDIF()
...
 IF (TPL_ENABLE_METIS)
   ASSERT_DEFINED(ENV{METIS_ROOT})
-  ATDM_SET_CACHE(METIS_INCLUDE_DIRS "$ENV{METIS_ROOT}/include" CACHE FILEPATH)
-  ATDM_SET_CACHE(METIS_LIBRARY_DIRS "$ENV{METIS_ROOT}/lib" CACHE FILEPATH)
+  # allow the user to set metis include and library dirs from the ENV
+  # first set the from the ENV and use (what was here) as the default
+  # then set the cache variable to that.
+  ATDM_SET_ATDM_VAR_FROM_ENV_AND_DEFAULT(METIS_INCLUDE_DIRS
+					 "$ENV{METIS_ROOT}/include")
+  ATDM_SET_CACHE(METIS_INCLUDE_DIRS "${ATDM_METIS_INCLUDE_DIRS}" CACHE FILEPATH)
```
e10harvey, line 440: Can you revert this and use `export METIS_ROOT=/path/to/metis` instead?

e10harvey (changes requested):
Thanks, @jjellio ! I just have one question below and `unset_atdm_config_vars_environment.sh` needs to be updated to unset the new environment variables introduced in these changes.

Lastly, would you please test these changes via:
`nohup env ./ctest-s-local-test-driver.sh all &>ctest-s-local-test-driver.out &`

and post the cdash link?

cmake/std/atdm/ATDMDevEnvSettings.cmake
jjellio: There are no new ENVs here.  These variables *should* be supported for consistency. 

This is what causes headaches when a package installs into `lib64` vs `lib` - let us set the correct paths and not depend on `ROOT/lib and ROOT/include` to be the required directories for includes/libs - because it isn't universal.

I can't really see METIS ever needing more than one directory, but who knows.

brian-kelley:
@jjellio Lambdas work with HIP without any extra flags.

jjellio:
@brian-kelley

Brian, the question is really does the library need to declare. I think many people assume Lambda+cuda always works. If Trilinos stuff is going to universally make that assumption, then it seems OK to not have it.  I haven't put much  thought into it. It's something to think about and make sure coders have the proper tools to write portable stuff (I don't think Trilinos even tests for Cuda w/out lambda.)

brian-kelley:
@jjellio OK, I see what you mean. I can say that no code in Trilinos does this:
```
#ifdef KOKKOS_ENABLE_CUDA_LAMBDA
//lambda codepath
#else
//fallback path
#endif
```
aside from a few tests in Kokkos and KokkosKernels that use lambdas. Other than that, no package ever references the lambda. If Tpetra_ENABLE_CUDA, Tpetra checks that "Kokkos_ENABLE_CUDA_LAMBDA" is ON at configure time.

bartlettroscoe (changes requested):
This preserves backward compatibility so I think this is safe to merge without a lot of testing.  The only required change that I see is to add the new vars `PARMETIS_INCLUDE_DIRS` and `PARMETIS_LIBRARY_DIRS` to [cmake/std/atdm/utils/unset_atdm_config_vars_environment.sh](https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/utils/unset_atdm_config_vars_environment.sh).  (See **REQUIRED** below.)

But given that HIP is a new Kokkos backend very much like CUDA and currently the ATDM Trilinos system handles CUDA support as part of the compiler name (containing `cuda`) in the build-name,  we should ask how HIP support should be added.  Should this be a `hip` keyword added to the build-name and therefore parsed in  [cmake/std/atdm/utils/set_build_options.sh](https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/utils/set_build_options.sh) and documented?

* https://github.com/trilinos/Trilinos/blob/develop/cmake/std/atdm/README.md#kokkos_thread

?

If not, then how do does the HIP backend get handled?

cmake/std/atdm/ATDMDevEnvSettings.cmake
jjellio: @e10harvey @bartlettroscoe 
I see what you mean in terms of "new" variables.

In my context, I've been skipping all of the `load-env` magic because none of this stuff is really supported.  So, what I have is effectively an ENV that can drive ATDMDevSettings (without needing the framework you all have).

w.r.t. to Evan's comments - I'm looking forward a little bit. Ross has seen the monster I am creating.  It effectively marries `spack` and `ATDMDevSettings`. In that context, I **generate** the env based on what Spack has done. 

It follows then, that I have auto-generated (correct) `INCLUDE_DIRS`, `LIBRARY_DIRS`, and `..._LIBS` (as prescribed by what spack determines the dependencies to be. 

With that in mind, it is natural to define the ATDM_CONFIG api as:
`ATDM_CONFIG_TPL_{INCLUDE_DIRS,LIBRARAY_DIRS,LIBS,LIBRARY_NAMES}`
Which would allow the ENV to be written in a such a way that you could `find_library` if you wanted or hardcode `LIBS` if you want that.

In practice, I am setting `ATDM_CONFIG_foo_INCLUDE_DIRS` and `ATDM_CONFIG_foo_LIBS` - then I can spell out precisely shared or static link lines.

It's a different use case than what SNL is doing right now - but is is kinda a necessary thing if you expect ATDMDevSettings to work on platforms that aren't really well defined (noone can write an ENV file until they sort out the system).

See another comment below (in reply to ross)

cmake/std/atdm/ATDMDevEnvSettings.cmake
```diff
@@ -438,8 +455,16 @@ ENDIF()
 ATDM_SET_ENABLE(TPL_ENABLE_ParMETIS ${TPL_ENABLE_ParMETIS_DEFAULT})
 IF (TPL_ENABLE_ParMETIS)
   ASSERT_DEFINED(ENV{PARMETIS_ROOT})
-  ATDM_SET_CACHE(ParMETIS_INCLUDE_DIRS "$ENV{PARMETIS_ROOT}/include" CACHE FILEPATH)
-  ATDM_SET_CACHE(ParMETIS_LIBRARY_DIRS "$ENV{PARMETIS_ROOT}/lib" CACHE FILEPATH)
+
+  # if the naming in ATDM_CONFIG matched cmake variables, this is alot cleaner
+  ATDM_SET_ATDM_VAR_FROM_ENV_AND_DEFAULT(PARMETIS_INCLUDE_DIRS
+					 "$ENV{PARMETIS_ROOT}/include")
```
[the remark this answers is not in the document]
jjellio: @e10harvey @bartlettroscoe (consider this)

I'm going to digress a bit in this post - but it's totally worth you and Evan thinking about.

Ross, I've been using this snipplet to unset ATDM stuff for while.  What do you think replacing the bulk of that file with
```
for atdm_var in $(compgen -A variable | grep '^ATDM_CONFIG'); do 
  # for example only
  echo "Before: $atdm_var=${!atdm_var}"; 

  # the real deal
  unset $atdm_var;

  # for example only
  echo "After: $atdm_var=${!atdm_var}";
done

# the variables are actually undefined afterwards... but I print the name, so `ha=`
Before: ATDM_CONFIG_ha=1
After: ATDM_CONFIG_ha=
Before: ATDM_CONFIG_he=1
After: ATDM_CONFIG_he=
Before: ATDM_CONFIG_lol=3
After: ATDM_CONFIG_lol=
```

In the context of ATDMDevSettings, you could even write this to create a cache VAR for every `ATDM_CONFIG_<foo...>`

I've toyed with this (and it works wonderfully).

In a refactor, you'd move all of the non-prefixed `ATDM_CONFIG` variables to be `ATDM_CONFIG_<name>`. This would allow an `atdm author` to arbitrarily inject necessary variables.

The logic in ATDMDevSettings then gets cleaner. You effectively check for variables that are set (and define some policy/precedence for whether `find` is used vs raw `TPL_FOO_LIBRARIES` is used.

The code below would allow you to disable BOOST, by doing `ATDM_CONFIG_TPL_ENABLE_BOOST=OFF`
A little polish to recognize  `types`, but this would really simplify the soup.

```
function(load_atdm_config_variables)
# run a basic bash command grab all variable **names** in the ENV
# restrict to starting with ATDM_CONFIG_
# strip the ATDM_CONFIG_ prefix
# then package the result into list
#
# Note, these are variables NAMES, not values
execute_process(COMMAND bash "-c" "compgen -A variable"
                COMMAND grep "^ATDM_CONFIG"
                COMMAND sed "s/^ATDM_CONFIG_//"
                COMMAND tr  "[:space:]" ";"
                RESULT_VARIABLE rc
                OUTPUT_VARIABLE thebomb)

message( "rc == ${rc}" )
message( "${thebomb}" )
set(atdm_names ${thebomb})
message(${atdm_names})
foreach(e ${thebomb})
  message("ATDM_CONFIG_${e} = $ENV{ATDM_CONFIG_${e}} ")
  set(${e} $ENV{ATDM_CONFIG_${e}} CACHE STRING "user ATDM_CONFIG variable" FORCE)
endforeach()

endfunction()
```


Effectively what I am proposing is that ATDMDev take `ownership` of all ENVs prefixed `ATDM_CONFIG` - that seems perfectly reasonable to me - Then, in an `environment.sh` file, we **only** operate on `ATDM_CONFIG` prefixed variables.

This would avoid name clashes... e.g., `MPICC` or really anything.

The top-most snipplet shows how to en-masse remove ATDM_CONFIG variables, and the bottom shows how Cmake could be made to create variables from it (you could twiddle it to be CACHE - or use `macro` - and I think you could set a file local variable instead (probably preferred!).

PR #9117, opened 2021-05-07
Tpetra:  adding syncs of import buffers in MultiVector

kddevin:
@trilinos/tpetra 

## Motivation

In https://github.com/Exawind/nalu-wind/issues/858, @rcknaus reported export errors when TPETRA_ASSUME_CUDA_AWARE_MPI=0 after #8821 was merged.
Similar errors were exposed by tpetra/core/test/CrsMatrix/Bug8794.cpp.

In Bug8794.cpp, many multivector elements are imported, due to the dense rows in the test matrix.
With TPETRA_ASSUME_CUDA_AWARE_MPI=0, DistObject decides to do communication on host.
With long rows, the number of imports is large, so MultiVector decides to unpack on device.
However, the sync of the import buffer between the communication and unpack was missing.
This PR adds the sync.

I still need to investigate what changed in #8821 to cause this error.  We should also set up nightly tests with TPETRA_ASSUME_CUDA_AWARE_MPI=0; Trilinos PR testing has TPETRA_ASSUME_CUDA_AWARE_MPI=1 (which, for Bug8794.cpp, meant communication and unpacking were done on device, so no problems).  Adding unit test reproducer (separate from Bug8794.cpp) would be nice as well.

## Stakeholder Feedback

https://github.com/Exawind/nalu-wind/issues/858
@rcknaus
See https://github.com/Exawind/nalu-wind/issues/858#issuecomment-834808823

## Testing

Tested on ascicgpu with TPETRA_ASSUME_CUDA_AWARE_MPI=0 and 1.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_DistObject_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_DistObject_def.hpp
@@ -1101,7 +1101,7 @@ namespace Tpetra {
                  numImportPacketsPerLID_av);
             }
           }
-          else { // pack on device
+          else { // comm on device
             Kokkos::fence(); // for UVM
             this->imports_.modify_device ();
             if (revOp == DoReverse) {
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -336,7 +336,8 @@ namespace { // (anonymous)
     if (! imports.need_sync_device ()) {
       return false; // most up-to-date on device
     }
-    else { // most up-to-date on host
+    else { // most up-to-date on host, 
+           // but if large enough, worth running on device anyway
       size_t localLengthThreshold = 
              Tpetra::Details::Behavior::multivectorKernelLocationThreshold();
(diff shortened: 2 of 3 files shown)


brian-kelley (approved):

jjellio:
@kddevin I think the testing should be running *both* Cuda-aware ON and OFF.
E.g.,
https://testing.sandia.gov/cdash/test/70227930 (has it off)


I can't navigate CDash for the life of me - if I search for an ATDM build with the Tpetra flag = 0, I get
https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-&field2=testoutput&compare2=95&value2=TPETRA_ASSUME_CUDA_AWARE_MPI%3D0

To avoid a full rebuild, the code is built once, but the testing runs twice (as I understand it).

I pushed pretty hard for this functionality, because EMPIRE has used Cuda-aware = OFF in several instances...

Hope that's somewhat helpful. (if it didn't help flag the issue you're having it seems it has failed to do its job though!)

PR #9129, opened 2021-05-12
Tpetra MultiVector: Remove unnecessary check from isSameSize

cgcgcg:
@trilinos/tpetra 

## Motivation
Remove unnecessary check from `MV::isSameSize`. This avoids a potential memcopy to host.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -4684,16 +4684,6 @@ namespace Tpetra {
     if ((l1!=l2) || (this->getNumVectors() != vec.getNumVectors())) {
       return false;
     }
-    if (l1==0) {
-      return true;
-    }
-
-    auto v1 = this->getLocalViewHost(Access::ReadOnly);
-    auto v2 = vec.getLocalViewHost(Access::ReadOnly);
-    if (PackTraits<ST>::packValueCount (v1(0,0)) !=
-        PackTraits<ST>::packValueCount (v2(0,0))) {
-      return false;
-    }
 
     return true;
   }


kddevin (approved):
Looks good to me.  The compiler should prevent multivectors with different scalar types from being compared in this function.  
I'll add the UVM Removal tag, as this issue was more costly with UVM=OFF.
Thanks @cgcgcg

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -4684,16 +4684,6 @@ namespace Tpetra {
...
     }
-    if (l1==0) {
-      return true;
-    }
-
-    auto v1 = this->getLocalViewHost(Access::ReadOnly);
-    auto v2 = vec.getLocalViewHost(Access::ReadOnly);
-    if (PackTraits<ST>::packValueCount (v1(0,0)) !=
-        PackTraits<ST>::packValueCount (v2(0,0))) {
-      return false;
```
kyungjoo-kim, line 4695: @kddevin @cgcgcg This PR is already merged but I have one curiosity. The type check is not really meaningful if the value is not dynamically sized. Does this condition require for Sacado DFAD, where its packing size is determined only in runtime ?

rppawlo: @etphipp

kyungjoo-kim: @etphipp Do you happen to know that Mark put this condition for SACADO ?

kddevin: @kyungjoo-kim  good question.  I was expecting the compiler to enforce that v1(0,0) and v2(0,0) would have the same size, since this and vec are MultiVector<Scalar>.
@etphipp ?

etphipp: My guess is it was for stokhos, actually which does have a dynamically sized type (we currently don't run Sacado through Tpetra, but if we did, this would also be relevant to DFad).

etphipp: I'm not aware of any use-case that would call this function with differently sized types, however.

etphipp: If everything passes with it removed, I don't see why you shouldn't merge it.

kyungjoo-kim: This is already merged. ; ).

kyungjoo-kim: Maybe it might be better if we put the check with debug ifdef guard just in case.

cgcgcg: I'll let you guys decide what's best here. I just didn't like the fact that a host copy was triggered for a type check.

PR #9131, opened 2021-05-13
MueLu: update blocked coord transfer fact

pohm01:
@trilinos/muelu 

## Motivation
BlockedCoordinatesTransferFactory was not properly setup when called from an xml input deck. Also updated to work for multiple dofs per node (per block). This now allows for blocked operators to use coordinate dependent factories, such as zoltan2 rebalancing.

This is related to work being done for MHD using Drekar to allow for rebalancing of the operators on the coarse levels:
https://github.com/rppawlo/DrekarBase/issues/546

## Testing
To be added.

### DIFF
--- a/packages/muelu/src/Misc/MueLu_BlockedCoordinatesTransferFactory_def.hpp
+++ b/packages/muelu/src/Misc/MueLu_BlockedCoordinatesTransferFactory_def.hpp
@@ -95,7 +95,7 @@ namespace MueLu {
     GetOStream(Runtime0) << "Transferring (blocked) coordinates" << std::endl;
 
     const size_t numSubFactories = NumFactories();
-    //std::vector<RCP<const Map> > subBlockMaps(numSubFactories);
+    std::vector<RCP<const Map> > subBlockMaps(numSubFactories);
     std::vector<RCP<dMV> > subBlockCoords(numSubFactories);
 
     if (coarseLevel.IsAvailable("Coordinates", this)) {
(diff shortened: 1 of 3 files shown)


packages/muelu/src/Misc/MueLu_BlockedCoordinatesTransferFactory_def.hpp
```diff
@@ -109,14 +109,51 @@ namespace MueLu {
...
+      LO blkSize = 0;
+      LO otherBlkSize;
+      for(size_t i=0; i<numSubFactories; i++) {
+        if (rcp_dynamic_cast<const StridedMap>(coarseMap->getMap(i, thyraMode)) != Teuchos::null){
+          otherBlkSize = rcp_dynamic_cast<const StridedMap>(coarseMap->getMap(i, thyraMode))->getFixedBlockSize();
+        } else {
+          otherBlkSize = 1;
+        }
+        if( i>0 && blkSize/i != otherBlkSize)
+          GetOStream(Warnings1) << "BlockedCoordinatesTransferFactory: Subblocks have different stride. Might cause problems..." << std::endl;
```
jhux2, line 136: Should this be an error instead of a warning?

packages/muelu/src/Misc/MueLu_BlockedCoordinatesTransferFactory_def.hpp
```diff
@@ -109,14 +109,51 @@ namespace MueLu {
...
+        blkSize += otherBlkSize;
```
jhux2, line 137: Why is this added?

jhux2 (commented):
Could you please add something in the motivation section, so that future us will know this is related to your MHD work?

packages/muelu/src/Misc/MueLu_BlockedCoordinatesTransferFactory_def.hpp
pohm01: Was originally thinking of allowing block sizes to be different. I have changed it to be an error to avoid unexpected behavior.

packages/muelu/src/Misc/MueLu_BlockedCoordinatesTransferFactory_def.hpp
pohm01: I was playing around with allowing different block sizes. Couldn't quite figure it out, so now block sizes need to be the same between different larger blocks.

packages/muelu/test/unit_tests/BlockedRepartition.cpp
```diff
@@ -938,14 +941,331 @@ namespace MueLuTests {
...
+
+  /**********************************************************************************/
+  /* CREATE INITIAL MATRIX                                                          */
+  /**********************************************************************************/
+  RCP<const Map> bigMap;
+  RCP<const Map> map1;
+  RCP<const Map> map2;
+  GO numElements = 400;
+  GO numElements1 = 200;
+  GO numElements2 = 200;
```
jhux2, line 966: @pohm01 Are the blocks allowed to have different numbers of elements?  If so, would the test be more general and/or exercise more parts of the blocked transfer factory?  It's fine if the answer is "no", I'm just wondering.

jhux2 (dismissed):
Lgtm -- had one question, see my comment.

packages/muelu/test/unit_tests/BlockedRepartition.cpp
pohm01: Currently there is only support for one set of coordinates that is shared between the blocks, so the blocks need the same number of elements and co-located dofs/nodes.

jhux2 (approved):
lgtm

PR #9132, opened 2021-05-13
Skip view initializations

cgcgcg:
@trilinos/tpetra @trilinos/ifpack2 @trilinos/muelu 

## Motivation
- Skip unneeded view initializations.
- Tpetra global norm: use in-place reduction, similar to https://github.com/trilinos/Trilinos/blob/5242864373cc43d2fd8bedfd775c236daddeaf2c/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp#L1962-L1982

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
@@ -1043,7 +1043,7 @@ makeInverseDiagonal (const row_matrix_type& A, const bool useDiagOffsets) const
       *out_ << "Reusing pre-existing vector for diagonal extraction" << std::endl;
     D_rowMap = Teuchos::rcp_const_cast<V>(D_);
   } else {
-    D_rowMap = Teuchos::rcp(new V (A.getGraph ()->getRowMap ()));
+    D_rowMap = Teuchos::rcp(new V (A.getGraph ()->getRowMap (), /*zeroOut=*/false));
     if (debug_)
       *out_ << "Allocated new vector for diagonal extraction" << std::endl;
   }
@@ -1571,10 +1571,9 @@ makeTempMultiVector (const MV& B)
   // null, but also if the number of columns match, since some multi-RHS
   // solvers (e.g., Belos) may call apply() with different numbers of columns.
 
-  //W must be initialized to zero when it is used as a multigrid smoother.
   const size_t B_numVecs = B.getNumVectors ();
   if (W_.is_null () || W_->getNumVectors () != B_numVecs) {
-    W_ = Teuchos::rcp (new MV (B.getMap (), B_numVecs, true));
+    W_ = Teuchos::rcp (new MV (B.getMap (), B_numVecs, false));
   }
   return W_;
 }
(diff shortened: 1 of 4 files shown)


jhux2 (approved):
lgtm

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
```diff
@@ -1561,10 +1561,9 @@ makeTempMultiVector (const MV& B)
   // null, but also if the number of columns match, since some multi-RHS
   // solvers (e.g., Belos) may call apply() with different numbers of columns.
 
-  //W must be initialized to zero when it is used as a multigrid smoother.
```
jhux2, line 1564: So this statement is simply not correct?

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
```diff
@@ -248,18 +248,27 @@ gblNormImpl (const RV& normsOut,
...
+      if (whichNorm == NORM_INF) {
+        reduceAll<int, mag_type> (*comm, REDUCE_MAX, nv, lclSum, gblSum);
+      } else {
+        reduceAll<int, mag_type> (*comm, REDUCE_SUM, nv, lclSum, gblSum);
+      }
     } else {
-      reduceAll<int, mag_type> (*comm, REDUCE_SUM, nv, lclSum, gblSum);
+      mag_type* const gblSum = normsOut.data ();
+      if (whichNorm == NORM_INF) {
+        reduceAll<int, mag_type> (*comm, REDUCE_MAX, nv, gblSum, gblSum);
```
kddevin, line 268: Are all MPIs OK with aliasing the input and output?  I thought aliasing in MPI_Allreduce was bad.

kddevin (dismissed):
Questions about testing and capabilities (likely @trilinos/tpetra 's problem) and argument aliasing in MPI_Allreduce.

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
cgcgcg: This is essentially just copied over from the dot product. So I am assuming that this works.

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
cgcgcg: I don't think so. Looking at these lines
https://github.com/trilinos/Trilinos/blob/5242864373cc43d2fd8bedfd775c236daddeaf2c/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp#L1296-L1321
Either `W_` is written as `W := (1/theta)*D_inv*(B-A*X)`, or `W := (1/theta)*D_inv*B`.

But otherwise, we also should have had to zero the vector on every apply, and not just when we first allocate it.

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
cgcgcg: @kddevin Can I merge this, or did you want to have another look?

cgcgcg: ping @kddevin

cgcgcg: @kddevin Could you have another look?

jhux2 (approved):
lgtm

PR #9133, opened 2021-05-13
Tpetra: Skip unpackAndCombine

cgcgcg:
@trilinos/tpetra 

Allow to skip `unpackAndCombine` in the following (common) situation:
- import of a MV
- MV has single column
- from a 1-to-1 map into an overlapping map
- CombineMode == INSERT
- off-rank entries are grouped by rank
- memory spaces for comm and data buffers match
- no additional syncs are required, i.e. copyAndPermute and communication are both either on host or on device

If these conditions hold, then the `imports_` view can be aliased to part of the MV's data view, and `unpackAndCombine` can be skipped.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_Transfer_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_Transfer_decl.hpp
@@ -201,6 +201,13 @@ public:
   /// Maps in the way that you expect.
   bool isLocallyComplete () const;
 
+
+  void detectRemoteExportLIDsContiguous() const;
+
+  bool areRemoteLIDsContiguous() const;
+
+  bool areExportLIDsContiguous() const;
+
   /// \brief Describe this object in a human-readable way to the given
   ///   output stream.
   ///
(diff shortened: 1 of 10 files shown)


csiefer2 (approved):

tasmith4 (changes requested):
@cgcgcg these changes will not conflict with other work I have in progress to refactor Distributor.  However, the next steps for that work will be to adjust how DistObject uses Distributor so that asynchronous transfers using the same Import/Export object do not block each other.  Part of that work will probably involve some refactoring of the functions you've touched here.

I'm concerned that during this process I could accidentally break your code, since it seems the new test only covers the "happy path" case and doesn't test a wide range of corner case scenarios.  If you're confident these tests will fail if I accidentally break something while shifting things around, I'm happy to approve.  Otherwise could you add more tests until you're confident this code is protected?  I'd hate to break your optimization because I missed a subtle detail and the tests didn't catch my mistake.

cgcgcg:
@tasmith4 I'm not sure this is so easy to break. I guess the critical bit is DistObject. Even after shuffling things around, we still need to call `reallocImportsIfNeeded` before we post and sends/recvs. In that function, the aliasing occurs. The unit test checks that we have the correct aliasing / no aliasing behavior. Most of the code outside of DistObject is to determine whether we are allowed to alias.

tasmith4 (approved):
Just so you know, the refactoring will include breaking up Import into two pieces, and may include breaking up doPosts, so things inside Import will be reshuffled, and reallocImportsIfNeeded may not end up being called in the same way.

But like I said, this is really about me not wanting to accidentally break your code, so if you're confident, I'm ok with this submission.

PR #9135, opened 2021-05-14
Teko DiagonallyScaledPreconditioner test: Relax tolerance for CUDA

cgcgcg:
@trilinos/teko 

## Motivation
Relax tolerance for test, so that it passes on CUDA.

Should help avoid failures as seen in #9134 
https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=9177412


jhux2 (approved):
lgtm

Issue #1995, opened 2017-11-16
Remove use of HAVE_COMPLEX macro


mhoemmen:
Anasazi, Belos, and Stratimikos all use the `HAVE_COMPLEX` macro.  This is defined (unconditionally!) in `Teuchos_ConfigDefs.hpp`, where a comment marks it as a backwards compatibility measure.  This macro pollutes the global namespace; let's get rid of it.

@trilinos/anasazi @trilinos/belos @trilinos/teuchos  

## Motivation and Context

It's bad to pollute the global namespace with macros, especially if those macros have general, non-namespaced names that could easily collide with other projects' macros.  Furthermore, the macro is unconditionally defined, so it is not meaningful.

hkthorn:
The HAVE_COMPLEX ifdef has been removed from Anasazi and Belos, so it can be removed from Teuchos.

PR #9141, opened 2021-05-17
Anasazi: Removes HAVE_COMPLEX from complex-valued tests

hkthorn:
@trilinos/anasazi 

## Motivation

The HAVE_COMPLEX macro is a deprecated conditional from the days when
the complex<> class had non-uniform support across compilers. It can,
and should, be removed, as mentioned in Github issue #1995

## Testing

OSX GCC10.x serial


jennloe (approved):
Looks good to me.

Issue #9100, opened 2021-05-05
ATDM/van1-tx2: Deprecate and remove old toolchains


e10harvey:
CC: @trilinos/framework

Now that the openmpi 4.0.5 toolchain has been in use for several weeks, older toolchains should be marked for deprecation.

Related to https://github.com/trilinos/Trilinos/pull/9001.

e10harvey:
@jmgate Is `ARM-20.0_OPENMPI-4.0.2` still in use?

e10harvey:
@sebrowne is sparc using any toolchains other than `ARM-20.1_OPENMPI-4.0.5`?

rppawlo:
We have a TPL in empire that is hitting a bug in 4.0.2 that has been fixed in 4.0.5. Could we change the default to 4.0.5 even if we have to keep around the old toolchain?

e10harvey:
> Could we change the default to 4.0.5 even if we have to keep around the old toolchain?

Yes.

jmgate:
> @jmgate Is empire still using ARM-20.0_OPENMPI-4.0.2?

If it's still the default when matching `arm-20` in the build name, then yes.

rppawlo:
@e10harvey and @jmgate - Empire is building and passing fine with these changes except for one thing. The environment is loading python 2 instead of python 3 and the empire testing needs python 3. A simple module swap fixed it. The python issue could be unrelated to this commit, but perhaps it would be worth making that change as part of this commit? What do you think?  @bartlettroscoe - was there a reason stria modules load python 2?

bartlettroscoe:
> @bartlettroscoe - was there a reason stria modules load python 2?

@rppawlo, I don't remember.  The git log might say something about this but I doubt it.

e10harvey:
> What do you think?

I don't think so. The 4.0.5 environment isn't loading a python module.

e10harvey:
> I don't think so. The 4.0.5 environment isn't loading a python module.

I stand corrected. The python/2 module is loaded from the `arm-20.1_openmpi-4.0.5` module; since this works for other customers, let's leave the swap in empire.

jmgate:
And there should be nothing to do on the EMPIRE end.  If folks are using our BuildScripts, they should automatically get the Python we support in their environment.

sebrowne:
FWIW, we're pursuing changing the default to python3 (slowly but surely, and really maybe not that slowly)

SPARC is only using the mentioned toolchain

[...]

PR #9143, opened 2021-05-17
van1-tx2: Add deprecation notice

e10harvey:
CC: @sebrowne, @jwillenbring

### On van1-tx2:
```bash
$ source cmake/std/atdm/load-env.sh default
Hostname 'van1-tx2' matches known ATDM host 'van1-tx2' and system 'van1-tx2'
Using ARM ATSE compiler stack ARM-20.1_OPENMPI-4.0.5 to build DEBUG code with Kokkos node type SERIAL

$ source cmake/std/atdm/load-env.sh arm-20.0
Hostname 'van1-tx2' matches known ATDM host 'van1-tx2' and system 'van1-tx2'
Setting compiler and build options for build-name 'arm-20.0'
***
*** DEPRECATION NOTICE:
****   WARNING: "ARM-20.0_OPENMPI-4.0.2" will no longer be supported on 07-15-2021.
***
```

Related to #9100.

### DIFF
--- a/cmake/std/atdm/van1-tx2/environment.sh
+++ b/cmake/std/atdm/van1-tx2/environment.sh
@@ -193,6 +193,14 @@ export ATMD_CONFIG_MPI_USE_COMPILER_WRAPPERS=ON
 
 export ATDM_CONFIG_WCID_ACCOUNT_DEFAULT=fy150090
 
+# Print deprecation warning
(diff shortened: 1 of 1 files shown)


jmgate (approved):
LGTM.

jmgate (approved):
Approving again.

PR #9144, opened 2021-05-17
MueLu: piecwise constant semiCoarsening

rstumin:
Auto-PR for SHA b4ddb2c

### DIFF
--- a/packages/muelu/doc/UsersGuide/masterList.xml
+++ b/packages/muelu/doc/UsersGuide/masterList.xml
@@ -1058,6 +1058,15 @@
     </parameter>
 
     <parameter>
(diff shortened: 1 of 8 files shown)


GrahamBenHarper (commented):
Are we planning to test this anywhere?

rstumin:
@GrahamBenHarper  I've tested this locally but held back from putting in a real test. I was hoping to do a second PR with another capability that will work with this one and then add a test for this with that new capability. Not sure if this would be viewed as bad or not.

GrahamBenHarper (approved):
@rstumin that works with me. Let's see how the AT is feeling today... :stuck_out_tongue:

PR #9163, opened 2021-05-21
Belos:  Adds more unpreconditioned Tpetra tests

hkthorn:
@trilinos/belos 

## Motivation

This commit translates more of the Epetra-based tests to Tpetra
linear algebra.

1)  The Tpetra solver factory test was extended to include all of
the linear solvers that are accessible through the Belos linear
solver factory.  This resulted in the detection of a bug in the
test that was causing solver failures, which has been addressed.

2)  Adds diagonal test to MINRES and TFQMR solvers for Epetra.
The extension of the Tpetra solver factory test illustrated a
problem in the MINRES solver for diagonal/identity matrices,
which has been addressed in this commit.

3)  Adds BiCGStab, RCG and TFQMR test to Tpetra-based linear
algebra.  Testing for these solvers has been absent before the
extention to the Tpetra solver factory and the addition of these
tests.

## Testing

OSX GCC10.x serial


packages/belos/src/BelosMinresIter.hpp
```diff
@@ -592,8 +594,8 @@ class MinresIter : virtual public MinresIteration<ScalarType,MV,OP> {
 
       // Update x:
       // x = x + phi*w;
-      MVT::MvAddMv( one, *cur_soln_vec, phi, *W_, *cur_soln_vec );
-      lp_->updateSolution();
+      //MVT::MvAddMv( one, *cur_soln_vec, phi, *W_, *cur_soln_vec );
```
cgcgcg, line 597: remove commented line, add comment that `updateSolution` does this?

packages/belos/epetra/test/MINRES/test_minres_hb.cpp
```diff
@@ -310,11 +323,6 @@ main (int argc, char *argv[])
         }
       }
     }
-
-#   ifdef BELOS_TEUCHOS_TIME_MONITOR
-    Teuchos::TimeMonitor::summarize (verbOut);
-#   endif // BELOS_TEUCHOS_TIME_MONITOR
-
```
cgcgcg, line 317: Why?

packages/belos/tpetra/test/LinearSolverFactory/SolverFactory.cpp
```diff
@@ -186,8 +187,14 @@ testSolver (Teuchos::FancyOStream& out,
 
   out << "Apply solver to \"solve\" AX=B for X.  Belos already has solver "
     "tests; the point is to check that solve() doesn't throw." << endl;
-  X->putScalar (STS::zero ());
-  solver->solve ();
+  try {
+    solver->solve ();
+  } catch (std::exception& e) {
+    out << "*** FAILED: Belos::SolverFactory::solve threw an exception: "
```
cgcgcg, line 193: `SolverManager::solve`?

packages/belos/src/BelosMinresIter.hpp
```diff
@@ -403,12 +403,13 @@ class MinresIter : virtual public MinresIteration<ScalarType,MV,OP> {
     // Create convenience variables for zero, one.
     const ScalarType one = SCT::one();
     const MagnitudeType zero = SMT::zero();
+    const MagnitudeType m_zero = SMT::zero();
```
jennloe, line 406: How is the m_zero here different from zero?

jennloe (approved):
Needs tiny change in BelosMinresIter.hpp on line 405, but this can be fixed in a separate commit.  I don't think that makes any practical difference.

jennloe:
As a side note- It would be nice to add a Tpetra test or two that uses a Matrix Market file.  When I play with them, I always get annoyed that our Tpetra tests default to only .hb files, and then I have to put in the Matrix Market reader manually.  :)

packages/belos/src/BelosMinresIter.hpp
hkthorn: It's a zero for the magnitude type.  It's easier to identify the difference between the scalar type zero and magnitude type zero by using the "m_" prefix.  It is not used consistently throughout the code, but it helps with clarity.

packages/belos/src/BelosMinresIter.hpp
hkthorn: Sure, I can add that when I fix the solver for right-preconditioning.

packages/belos/tpetra/test/LinearSolverFactory/SolverFactory.cpp
hkthorn: Added a try/catch loop around the solve because the test was not catching solver failures.  This was useful for debugging the test and solvers that were having issues with identity matrices.

cgcgcg (approved):

packages/belos/src/BelosMinresIter.hpp
jennloe: Did you notice that the 'zero' above here is also a magnitude type?

PR #9167, opened 2021-05-24
Tempus: Doxygen Not Creating Equations

ccober6:
The images for the latex equations are not being created correctly.
These images are correctly generated on Mac and RHEL. Convert the
doxygen build to a build that has the Trilinos look and feel (e.g.,
panzer). This includes rebuilding from scratch with every build.

@trilinos/tempus 

## Motivation
Get doxygen workig correctly on the Trilinos.org site.

## Stakeholder Feedback
@krcb noted this and informed us.  Will check with him if this corre3cts this issue.

## Testing
Changes worked on Mac.


rppawlo (approved):

krcb:
@ccober6 This appears to still be broken on the Trilinos.org documentation

sconde:
Yes, it is still not rendering the equations

ccober6:
Yes, I have been working on it in my spare time.  :)  I believe the basic problem is github.io is not setting up latex correctly on the server, so doxygen is not building the equations.  Although I have been working with the sysadmins at github.io (via Jim Willenbring), I am having to basically debug each step in the failure myself (on a remote system that I can only access via webpages) and tell them what the problem is.  They fix that, and only that, even though they can see the next failure.  :(  NOT HAPPY ABOUT THE SUPPORT.  So I have not gotten to the next step in the failure, and I do not think their "fix" is propagating to the cron job that generates the equations.

The other packages have old equation images laying around so that is why it appears they are still working.

PR #9172, opened 2021-05-25
Belos:  Fixes MINRES solver to perform correctly for right-hand preconditioning

hkthorn:
@trilinos/belos 

## Motivation

The MINRES solver had been written to work with left preconditioning,
but doesn't work with right-preconditioning or preconditioning on both
sides.  That has been corrected with this commit, using the same 
preconditioning strategy as CG.  A test including preconditioning on 
either left or right-side has been added for Epetra.

Tpetra tests for all solvers, that includes preconditioners, will be
added in a future commit.

## Testing

Test added to Epetra/test/MINRES directory
OSX GCC10.x Serial testing used to confirm passing

### DIFF
--- a/packages/belos/epetra/test/MINRES/test_pminres_hb.cpp
+++ b/packages/belos/epetra/test/MINRES/test_pminres_hb.cpp
@@ -1,0 +1,255 @@
+//@HEADER
+// ************************************************************************
+//
+//                 Belos: Block Linear Solvers Package
+//                  Copyright 2004 Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the Corporation nor the names of the
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL SANDIA CORPORATION OR THE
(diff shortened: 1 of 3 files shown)


hkthorn:
@jennloe @cgcgcg Could you take a look at this merge request?  I fixed right-preconditioning for MINRES and addressed some of your comment in the previous commit.  Thanks!

packages/belos/epetra/test/MINRES/test_pminres_hb.cpp
```diff
@@ -0,0 +1,255 @@
...
+    bool set = problem->setProblem();
+    if (set == false) {
+      if (proc_verbose)
+        std::cout << std::endl << "ERROR:  Belos::LinearProblem failed to set up correctly!" << std::endl;
+      return -1;
+    }
+
+    // Create an iterative solver manager.
+    RCP< Belos::SolverManager<double,MV,OP> > solver
+      = rcp( new Belos::MinresSolMgr<double,MV,OP>(problem, rcp(&belosList,false)) );
```
jennloe, line 203: There are lots of lines like this in the solver tests. Cleaner way is to use rcpFromRef.

packages/belos/src/BelosMinresIter.hpp
```diff
@@ -513,10 +518,6 @@ class MinresIter : virtual public MinresIteration<ScalarType,MV,OP> {
       // Apply operator.
       lp_->applyOp (*V, *Y_);
 
-      // Apply shift
-      if (shift != zero)
-	MVT::MvAddMv (one, *Y_, -shift, *V, *Y_);
-
```
jennloe, line 519: Why did we remove this?

packages/belos/src/BelosMinresIter.hpp
```diff
@@ -594,8 +602,8 @@ class MinresIter : virtual public MinresIteration<ScalarType,MV,OP> {
 
       // Update x:
       // x = x + phi*w;
-      //MVT::MvAddMv( one, *cur_soln_vec, phi, *W_, *cur_soln_vec );
-      lp_->updateSolution( W_, true, phi );
```
jennloe, line 598: I thought Christian's comment meant to just delete the comment on line 597??

jennloe (approved):
A few minor things to improve in the next update, but overall this is good.

packages/belos/src/BelosMinresIter.hpp
hkthorn: Using shifts in this solver was never completed.  I'm removing dead code to avoid confusion.

packages/belos/src/BelosMinresIter.hpp
hkthorn: So, the MINRES solver updates the solution like the CG solver does, so the linear problem just needs to be aware of the fact that the solution has been updated manually.  I originally thought that it would be better to pass the update to the updateSolution call and let it handle unraveling the preconditioning, but that was wrong.  The code has been returned to the previous state.

packages/belos/epetra/test/MINRES/test_pminres_hb.cpp
hkthorn: Sure, this driver was copied from the preconditioned CG tests, so it's an artifact of copying older code.  I agree there are cleaner ways of handling this.

PR #9174, opened 2021-05-25
KokkosKernels/Ifpack2: MTGS long row kernel (

brian-kelley:
Patch in https://github.com/kokkos/kokkos-kernels/pull/993, and add support in Ifpack2.
``relaxation: long row threshold`` controls the minimum nonzeros a row needs to be considered long.

Wait for the KokkosKernels PR to be merged before merging this, in case changes are needed.

@trilinos/ifpack2 

## Motivation

See https://github.com/kokkos/kokkos-kernels/issues/923 : MTGS was previously very slow when there was a lot of row imbalance.

## Stakeholder Feedback

Needed by Sierra.  See performance numbers here: https://github.com/trilinos/Trilinos/pull/9174#issuecomment-848381721.
## Testing

Added an Ifpack2 Relaxation unit test and the KokkosKernels gauss_seidel test to exercise this.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
@@ -327,6 +327,9 @@ Relaxation<MatrixType>::getValidParameters () const
     const int cluster_size = 1;
     pl->set("relaxation: mtgs cluster size", cluster_size);
 
+    const int long_row_threshold = 0;
+    pl->set("relaxation: long row threshold", long_row_threshold);
+
     validParams_ = rcp_const_cast<const ParameterList> (pl);
   }
   return validParams_;
@@ -367,6 +370,9 @@ void Relaxation<MatrixType>::setParametersImpl (Teuchos::ParameterList& pl)
   int cluster_size = 1;
   if(pl.isParameter ("relaxation: mtgs cluster size")) //optional parameter
     cluster_size = pl.get<int> ("relaxation: mtgs cluster size");
+  int long_row_threshold = 0;
+  if(pl.isParameter ("relaxation: long row threshold")) //optional parameter
+    long_row_threshold = pl.get<int> ("relaxation: long row threshold");
 
   Teuchos::ArrayRCP<local_ordinal_type> localSmoothingIndices = pl.get<Teuchos::ArrayRCP<local_ordinal_type> >("relaxation: local smoothing indices");
 
@@ -396,6 +402,7 @@ void Relaxation<MatrixType>::setParametersImpl (Teuchos::ParameterList& pl)
   fixTinyDiagEntries_    = fixTinyDiagEntries;
   checkDiagEntries_      = checkDiagEntries;
   clusterSize_           = cluster_size;
+  longRowThreshold_      = long_row_threshold;
   is_matrix_structurally_symmetric_ = is_matrix_structurally_symmetric;
   ifpack2_dump_matrix_ = ifpack2_dump_matrix;
   localSmoothingIndices_ = localSmoothingIndices;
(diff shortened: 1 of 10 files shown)


jhux2 (approved):
@brian-kelley Thank you for making this available and the unit test.  This lgtm.

jhux2:
Here are some initial performance numbers -- they look great!

MTGS without long rows option (min/avg/max)
```
Belos: BiCGStabSolMgr total solve time              7.626 (1)        7.631 (1)        7.635 (1)        7.631 (1)
Belos: PseudoBlockCGSolMgr total solve time         6.015 (1)        6.025 (1)        6.033 (1)        6.025 (1)
Belos: PseudoBlockGmresSolMgr total solve time      6.064 (1)        6.109 (1)        6.125 (1)        6.109 (1)
```

MTGS with long rows option (min/avg/max)

```
Belos: BiCGStabSolMgr total solve time              1.121 (1)        1.16 (1)         1.178 (1)        1.16 (1)
Belos: PseudoBlockCGSolMgr total solve time         0.9129 (1)       0.9221 (1)       0.9305 (1)       0.9221 (1)
Belos: PseudoBlockGmresSolMgr total solve time      0.9765 (1)       1.019 (1)        1.035 (1)        1.019 (1)
```

jhux2 (dismissed):
lgtm

srajama1 (commented):
@brian-kelley : I have only one comment. The default value of zero worries me. That means all existing code with any matrix pattern will use the long row option unless they set the long row option explicitly to some larger number. Is this the intent? I would have done the default to be something like 30, beyond which it will be considered long rows. Users can override that then if needed.

brian-kelley:
@srajama1 Only positive values enable the long row kernel at all. For 0 and negative, it's just the old algorithm for all rows.

ndellingwood:
@brian-kelley to prevent the kokkos-kernels changes being clobbered please hold off on merge until #9158 goes in (it has passed tests and ready, should be merged in tomorrow (Wed))

brian-kelley:
@jhux2 I just checked in some new changes that should increase performance a little:
- using a separate stream was actually slowing it down on V100, so that's gone
- now chooses the number of teams based on the actual length of the long rows, not fixing it at 16 per row

Other than that, I think this is ready. I will just wait for Kokkos promotion 3.4.1 and then rebase this on develop.

ndellingwood:
@brian-kelley #9158 is merged

packages/ifpack2/src/Ifpack2_Relaxation_def.hpp
```diff
@@ -726,8 +733,10 @@ void Relaxation<MatrixType>::initialize ()
       if (mtKernelHandle_->get_gs_handle () == nullptr) {
         if (PrecType_ == Details::GS2 || PrecType_ == Details::SGS2)
           mtKernelHandle_->create_gs_handle (KokkosSparse::GS_TWOSTAGE);
-        else if(this->clusterSize_ == 1)
+        else if(this->clusterSize_ == 1) {
           mtKernelHandle_->create_gs_handle ();
+          mtKernelHandle_->get_point_gs_handle()->set_long_row_threshold(longRowThreshold_);
```
jhux2, line 738: @brian-kelley Does this mean the long row threshold is incompatible or meaningless for cluster size > 1?  Do you think should there be a check above to see if both are set?  Otherwise I believe the threshold is silently ignored.

brian-kelley: @jhux2 Yes, it's incompatible with cluster GS. I agree there should be a check.

jhux2:
@brian-kelley wrote:
> I just checked in some new changes that should increase performance a little:
>
>   * using a separate stream was actually slowing it down on V100, so that's gone
>   * now chooses the number of teams based on the actual length of the long rows, not fixing it at 16 per row


Here are results based on your latest changes.  The times are slightly faster now.
```
Belos: BiCGStabSolMgr total solve time           0.9509 (1)       0.9696 (1)       0.977 (1)        0.9696 (1)
Belos: PseudoBlockCGSolMgr total solve time      0.9458 (1)       0.9998 (1)       1.037 (1)        0.9998 (1)
Belos: PseudoBlockGmresSolMgr total solve time   0.9677 (1)       0.971 (1)        0.9744 (1)       0.971 (1)
```

jhux2 (commented):
@brian-kelley Looks good to me.

jhux2 (approved):

srajama1 (approved):
Thanks @brian-kelley !

Issue #5729, opened 2019-08-16
Framework: Create Build for ETI OFF


ZUUL42:
## Enhancement
@trilinos/framework

Create build that sets disables explicit instantiation (ETI).
This may first be tried in dev->master.

mhoemmen:
btw Sierra now enables ETI by default (as of Tue 13 Aug 2019), though this would still be helpful as long as we permit users to disable ETI.

jwillenbring:
With changing the Trilinos default to ETI ON, we will need to add a build with ETI off, or we will lose the ability to build with ETI off. @kddevin mentioned this recently as well. I suggest one of the dev->master builds. I propose the Clang 9 build, but I am open to any other suggestions.

mhoemmen:
Best practice with an ETI OFF test would be to enable "weird" types, just to make sure that downstream dependency inversion for factories is working correctly.

ZUUL42:
@mhoemmen what are you referring to exactly with "weird types"?

ZUUL42:
I've tested a few things with ETI OFF using GCC 7.2.0 and clang 9.0.0. There are a significant amount of errors for both.
To minimize the number coming from GCC, I turned off WError, but I'm still getting 50+ build errors.
[GCC 7.2.0 ETI OFF CDash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=eti_off_werror_off-test-Trilinos_pullrequest_gcc_7.2.0-5807&field2=buildstamp&compare2=63&value2=Experimental)
[clang 9.0.0 ETI OFF CDash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=eti_off-test-Trilinos_pullrequest_clang_9.0.0-1412&field2=buildstamp&compare2=63&value2=Experimental)

It seems we're primarily running into issues in @trilinos/tempus with a few problems in @trilinos/piro and @trilinos/rol and some odd issue in @trilinos/muelu, @trilinos/shylu, @trilinos/ifpack2, and @trilinos/xpetra. Though there are likely more hiding behind the 50 error limiter.
Could the teams take a look and let me know if there is something I can change in my configuration or if there are changes we're going to need to make in the packages. Thanks

jhux2:
[...]

PR #9183, opened 2021-05-27
KokkosKernels: fix name lookup issue in non-ETI build

brian-kelley:
@trilinos/kokkos-kernels 

## Motivation

See discussion in #5729. This fixes some of the build errors.

## Testing

Doesn't change code semantics, just makes the namespaces more explicit.


srajama1 (approved):
Thanks @brian-kelley !

PR #9189, opened 2021-05-28
TriBITS: treat compiler flags verbatim for installation

masterleinad:
## Motivation
Fixes the same issue as https://github.com/trilinos/Trilinos/pull/9179 but this time properly such that quoted compiler flags are treated correctly when exporting them in a Trilinos installation. As opposed to https://github.com/trilinos/Trilinos/pull/9179, no changes are needed with respect to the latest Kokkos release 3.4.1.

We introduced the Kokkos change for being able to Tpetra with SYCL support, see https://github.com/trilinos/Trilinos/pull/9086. Since the JIT compile-mode triggered some internal compiler errors, I needed to use ahead-of-time compilation which needs `"-device skl"` as compiler flag. In the mean time, we fixed the JIT problem in Kokkos, see https://github.com/kokkos/kokkos/pull/4013, so we could theoretically live without the fix for quoted compiler arguments but it turns out that running the test suite in JIT mode is extremely slow due to the huge amount of kernels that need to be compiled at runtime. Hence, the ahead-of-time mode still seems preferable and we should fix the issue with quotes in compiler arguments anyway.

The changes to Kokkos in this pull request are not really necessary but make sure that we only introduce quotes for compiler arguments that need it. I'm happy to discuss if we would just want to drop it and revert https://github.com/trilinos/Trilinos/pull/9179 instead.

## Testing
I tested this by compiling and running the testsuite using also the changes in https://github.com/trilinos/Trilinos/pull/9086 with SYCL and also the demo app in https://github.com/trilinos/Trilinos/tree/develop/demos/simpleBuildAgainstTrilinos (which uncovered the problem in the first place).


crtrott:
Ready to test.

crtrott (commented):
I think this looks good and properly fixes the issue. But @bartlettroscoe probably needs to ok this, and put the same change into Tribits proper?

bartlettroscoe:
I will need to look over this carefully and this is likely going to be impacted by TriBITSPub/TriBITS#299 (that I am actually for real allowed to pull the trigger on finally).

Note that there is zero automated installation testing for Trilinos so having the PR builds and pass means nothing for these changes.  Also, TriBITS only has very superficial testing for this file (I actually did not write any of that initial code in TriBITS so I am adding tests as that code gets changed).  So I may need some help putting together a TriBITS test case that demonstrates the problem and then shows that this patch fixes the problem.

@masterleinad, can you provide a clear example that demonstrates the problem?  I can then create a TriBITS test case based on that.

masterleinad:
> @masterleinad, can you provide a clear example that demonstrates the problem? I can then create a TriBITS test case based on that.

The problem is reproducible whenever compiler arguments contain quotes. 
Between Kokkos Release 3.4.1 and #9179 we were putting all compiler arguments in quotes. As an example the compiler flags
```
"-fsycl" "-fsycl-unnamed-lambda" "-fsycl-targets=spir64_gen-unknown-unknown-sycldevice" "-Xsycl-target-backend" "-device skl"  -O3 -DNDEBUG
```
result in
```
SET(Trilinos_CXX_COMPILER_FLAGS " "-fsycl" "-fsycl-unnamed-lambda" "-fsycl-targets=spir64_gen-unknown-unknown-sycldevice" "-Xsycl-target-backend" "-device skl"  -O3 -DNDEBUG")
```
if building `Kokkos` with `SYCL` support. On the command line, this gets then translated to
```
;-fsycl" "-fsycl-unnamed-lambda" "-fsycl-targets=spir64_gen-unknown-unknown-sycldevice" "-Xsycl-target-backend" "-device;skl"  -O3 -DNDEBUG" 
```
with error messages like
```
/bin/sh: -fsycl -fsycl-unnamed-lambda -fsycl-targets=spir64_gen-unknown-unknown-sycldevice -Xsycl-target-backend -device: command not found
```
when trying to compile https://github.com/trilinos/Trilinos/tree/develop/demos/simpleBuildAgainstTrilinos?rgh-link-date=2021-05-28T18%3A28%3A44Z.

bartlettroscoe:
@masterleinad, how does `CMAKE_CXX_FLAGS` get into that state?  How would `CMAKE_CXX_FLAGS` be specified on the command-line to `cmake` to configure Trilinos to get into this state?

masterleinad:
> @masterleinad, how does `CMAKE_CXX_FLAGS` get into that state? How would `CMAKE_CXX_FLAGS` be specified on the command-line to `cmake` to configure Trilinos to get into this state?

@bartlettroscoe As I was saying before:
> Between Kokkos Release 3.4.1 and #9179 we were putting all compiler arguments in quotes.

Any flag that is added by `Kokkos` will be in quotes between Kokkos Release 3.4.1 and #9179. I've picked `SYCL` since that is the only case I know of so far that absolutely needs the quotes (and thus this case is not fixed by #9179).
An example different from building with `SYCL` support is building with `OpenMP` support (`Kokkos_ENABLE_OPENMP=ON`) which adds `"-fopenmp"` (between the two versions mentioned above) for `g++`.

bartlettroscoe:
> An example different from building with `SYCL` support is building with `OpenMP` support (`Kokkos_ENABLE_OPENMP=ON`) which adds `"-fopenmp"` (between the two versions mentioned above) for `g++`.

@masterleinad, what are the exact reproducability instructions for producing the error with an OpenMP build of Trilinos?  Can you do it with just a version of Trilinos 'develop'?  Do you need an updated version of Kokkos instead of what is under Trilinos/packages/kokkos?  I am looking for exact versions and commands to reproduce just so I am sure that I that I am seeing what you are seeing.

masterleinad:
> @masterleinad, what are the exact reproducability instructions for producing the error with an OpenMP build of Trilinos? Can you do it with just a version of Trilinos 'develop'? Do you need an updated version of Kokkos instead of what is under Trilinos/packages/kokkos? I am looking for exact versions and commands to reproduce just so I am sure that I that I am seeing what you are seeing.

I can use 06b49ace0301b804baaa8a43741c1db3bb80ccf5 (which is between Kokkos Release 3.4.1 and #9179) with `gcc-7.5.0` and 
```
cmake -DTrilinos_ENABLE_Tpetra=ON -DKokkos_ENABLE_OPENMP=ON ${TRILINOS_SOURCE}
```
to see
```
SET(Trilinos_CXX_COMPILER_FLAGS " "-fopenmp"  -O3 -DNDEBUG")
```
in `trilinos-install/lib/cmake/Trilinos/TrilinosConfig.cmake` after installing.
With the modifications here (excluding the ones to `Kokkos`), I would get
```
SET(Trilinos_CXX_COMPILER_FLAGS [[ "-fopenmp"  -O3 -DNDEBUG]])
```

bartlettroscoe:
@masterleinad

> ```
> SET(Trilinos_CXX_COMPILER_FLAGS [[ "-fopenmp"  -O3 -DNDEBUG]])
> ```

Okay, I think understand what is happening here.  This double bracket syntax `set( <varName> [[ <args> ]] )` is documented in at:

* https://cmake.org/cmake/help/v3.17/manual/cmake-language.7.html#bracket-argument

where it says:

> A bracket argument is always given to the command invocation as exactly one argument.

@jjwilke said that with modern CMake you want to handle compiler arguments as lists of arguments, not as a single string of arguments (see "Change variables to lists instead of strings" in https://github.com/TriBITSPub/TriBITS/issues/299#issuecomment-582381964).  But the current TriBITS and Trilinos CMake build systems (and everyone's Trilinos CMake configure scripts for that last 13 years) have been using a single string to handle compile flags and this PR is just trying to make updated Kokkos CMake build system that handles compiler options as a list of arguments work with the current Trilinos and TriBITS CMake build systems that handle those as a single string.  Correct?

I have created https://github.com/TriBITSPub/TriBITS/issues/375 to try to probe the issue of handling compiler and linker flags and lists vs. single strings more deeply.

In the meantime, I will work to create a reproducer on the TriBITS side and accept this patch on the TriBITS side and them merge updated TriBITS into this PR branch to get this merged to Trilinos 'develop'.

masterleinad:
> @jjwilke said that with modern CMake you want to handle compiler arguments as lists of arguments, not as a single string of arguments (see "Change variables to lists instead of strings" in [TriBITSPub/TriBITS#299 (comment)](https://github.com/TriBITSPub/TriBITS/issues/299#issuecomment-582381964)). But the current TriBITS and Trilinos CMake build systems (and everyone's Trilinos CMake configure scripts for that last 13 years) have been using a single string to handle compile flags and this PR is just trying to make updated Kokkos CMake build system that handles compiler options as a list of arguments work with the current Trilinos and TriBITS CMake build systems that handle those as a single string. Correct?

Almost. The full translation from the "modern" flags `Kokkos` is using to the `Trilinos` ones is https://github.com/kokkos/kokkos/blob/e4831441fb213bd29485327b02cf1b54149a58a7/CMakeLists.txt#L207-L242. In particular, the semicolon-separated list needs to be converted to a whitespace-separated list first (and this is pre-existing since at least Kokkos 3.0.00). In this process, we lose quotation marks needed for flags that contain whitespace so I try to detect them in the `Kokkos` part of this pull request to only add quotation marks if needed but adding them for all flags also seems to work fine. Thus, without any changes to `TriBITS` the flags are set correctly when building `Trilinos` or running the test suite.

The `TriBITS` part of this pull request makes sure that the exported flags are exactly the same as the ones used internally (by treating `${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_${CMAKE_BUILD_TYPE}}` as one string using double-bracket syntax just as intended by `Trilinos` and its users). This worked fine before already if the flags don't contain quotation marks.

bartlettroscoe:
CC: @masterleinad, @crtrott 

There is a detailed discussion of how compiler options/flags should really be handled correctly in a modern CMake project and in TriBITS consistent with how Kokkos and Trilinos wants to use them in TriBITSPub/TriBITS#375.  The working of that story and TriBITSPub/TriBITS#299 will involve some refactoring in TriBITS but mostly the refactoring of the CMake connection between Kokkos and Trilinos to just use `INTERFACE_COMPILE_OPTIONS` target properties and propagate those options with `target_link_libraries()`.  Therefore, we will be able to strip out a lot of the CMake glue code for the Kokkos/Trilinos specific interface.

I will go ahead and pull off the TriBITS patch and apply it to TriBITS 'master' to ensure it passes the TriBITS test suite and then we can merge this PR on the Trilinos side as-is.  This PR is a nice backwards-compatible fix for this problem until we can address TriBITSPub/TriBITS#375 and TriBITSPub/TriBITS#299.

bartlettroscoe (approved):
I squashed the TriBITS commits in this branch to one commit, used `git format-patch` and `git am` to apply that to a TriBITS branch and then I tested with the TriBITS test suite and the existing test suite passed.  I created TriBITS PR https://github.com/TriBITSPub/TriBITS/pull/376 for that branch and merged to TriBITS 'master'.  In confirmed that the updated TriBITS snapshot  into Trilinos merges cleanly into this Trilinos branch so all is good.

The better fix will be TriBITSPub/TriBITS#299 and TriBITSPub/TriBITS#375.

bartlettroscoe:
@trilinos/framework, why has PR testing on this PR not started yet?  I approved this PR 14 hours ago?

bartlettroscoe:
CC: @crtrott, @masterleinad,

I think there may be something wrong with the Trilinos PR autotester. See https://github.com/trilinos/Trilinos/issues/3276#issuecomment-856018127.

prwolfe:
Looks like there was a force push after the approval. That changes the SHA and so the autotester sees the code as unreviewed. Not sure why GitHub thinks it is currently reviewed?

crtrott (approved):

crtrott:
Really I didn't see this?

prwolfe:
> Really I didn't see this?

It was hard to find as the comments seem to be in the right order. The fact that the SHA the autotester was seeing as approved did not match the one in githubs list of commits was the main clue. Again, not sure how this happened.

Paul

PR #9194, opened 2021-05-30
Teuchos and Kokkos: adding support for long double ScalarType

ikalash:
This capability is needed for Alegra and currently supports a build with ETI off.

### DIFF
--- a/packages/kokkos/algorithms/src/Kokkos_Random.hpp
+++ b/packages/kokkos/algorithms/src/Kokkos_Random.hpp
@@ -494,6 +494,7 @@ struct rand<Generator, double> {
   }
 };
 
+#ifndef KOKKOS_ENABLE_CUDA
 template <class Generator>
 struct rand<Generator, long double> {
   KOKKOS_INLINE_FUNCTION
@@ -509,6 +510,7 @@ struct rand<Generator, long double> {
     return gen.ldrand(start, end);
   }
 };
+#endif
 
 template <class Generator>
 struct rand<Generator, Kokkos::complex<float> > {
@@ -863,6 +865,7 @@ class Random_XorShift64 {
     return drand(end - start) + start;
   }
 
+#ifndef KOKKOS_ENABLE_CUDA
   KOKKOS_INLINE_FUNCTION
   long double ldrand() { return urand64() / static_cast<long double>(MAX_URAND64); }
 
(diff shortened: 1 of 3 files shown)


csiefer2 (approved):

ikalash:
@trilinos/framework 

It appears the GPU builds for this PR are failing in the auto test tests (https://testing.sandia.gov/cdash/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-9194-test&field2=buildstarttime&compare2=84&value2=NOW ).  I don't really see how my changes could have caused these even though I made a minor addition to Kokkos.  Is this error due to an existing problem in Kokkos?

crtrott:
No this error is because in GPU builds long double doesn't generally work. Also please issue changes to Kokkos first to Kokkos itself. Changing this in Trilinos means you are not on a release version and thus get into a semi unsupported state :-). Also the next Kokkos update to trilinos would wipe this change out.

crtrott (changes requested):
a) we need this in Kokkos proper first.
b) this is highly problematic since most GPU backends do not support long double, and I am not sure we can make this work without some special casing.
c) is it acceptable that you do not cover the full range of long double values? (urand64 does not produce enough values for long double value range)

crtrott:
The actual error here is likely an artifact of only changing kokkos-algorithms, and there is a bug in the cmake if you try to only enable kokkos-algorithms tests inside of Trilinos. I guess if you touch Kokkos_Macros.hpp (add a newline or so) the testing would likely go forward. Note you didn't add tests for this though.

ikalash:
@crtrott : I see.  I have added ifdef's so the new code is not built in a CUDA build.  I know that I will need to do a separate PR in Kokkos.  I wanted to put it here too to verify that everything builds/compiles with Trilinos.  I will submit a separate Kokkos PR once this is merged into Trilinos.

@csiefer2 : can you please re-review the changes I just pushed?

ikalash:
> a) we need this in Kokkos proper first.
> b) this is problematic since most GPU backends do not support long double, and I am not sure this works this way without some special casing.
> c) is it acceptable that you do not cover the full range of long double values? (urand64 does not produce enough values for long double value range)

@crtrott : I just put an ifdef to disable the long double in a CUDA build.  Do you not like this solution?

PR #9208, opened 2021-06-02
Zoltan2: (Coloring) Fixed D1 performance bug, added timing argument to coloring code

ibogle:
@trilinos/zoltan2 

## Motivation

This fixes a  performance bug in the distance-1 coloring in Zoltan2, as well as adding the capability to report accurate times without any other output to all coloring implementations.

## Testing

These changes are covered by existing tests.

### DIFF
--- a/packages/zoltan2/core/src/algorithms/color/Zoltan2_AlgHybridD1.hpp
+++ b/packages/zoltan2/core/src/algorithms/color/Zoltan2_AlgHybridD1.hpp
@@ -232,11 +232,13 @@ class AlgDistance1 : public Algorithm<Adapter>
           }
         }
       },recoloringSize(0));
+      Kokkos::fence();
       Kokkos::parallel_for(n_local, KOKKOS_LAMBDA(const int& i){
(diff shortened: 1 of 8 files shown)


packages/zoltan2/core/src/algorithms/color/Zoltan2_AlgHybridD1.hpp
```diff
@@ -232,11 +232,13 @@ class AlgDistance1 : public Algorithm<Adapter>
           }
         }
       },recoloringSize(0));
+      Kokkos::fence();
       Kokkos::parallel_for(n_local, KOKKOS_LAMBDA(const int& i){
```
kddevin, line 236: Does this parallel_for need to specify ExecutionSpace?
Similarly for other methods (D2, D1-2GL, PD2).

kddevin (changes requested):
Add execution space to all parallel_fors in methods templated on ExecutionSpace

kddevin (approved):

PR #9211, opened 2021-06-02
MueLu: PlaneSmoothing via Stratimikos and recursive AMG

rstumin:
Auto-PR for SHA 4937549

### DIFF
--- a/packages/muelu/test/scaling/recurMG.xml
+++ b/packages/muelu/test/scaling/recurMG.xml
@@ -1,0 +1,140 @@
+<ParameterList name="MueLu">
+
+  <Parameter name="verbosity" type="string" value="high"/>
+
+  <Parameter name="max levels" type="int" value="4"/>
+  <Parameter name="coarse: max size" type="int" value="150"/>
+  <Parameter        name="number of equations"                  type="int"      value="3"/>
+
+  <Parameter name="multigrid algorithm" type="string" value="sa"/>
+  <Parameter name="sa: damping factor"  type="double" value="1.33333"/>
+
+   <!-- THE BIG DATA DUMP -->
+    <ParameterList    name="export data">
+       <Parameter  name="P"                             type="string"   value="{1234}"/>
+       <Parameter  name="A"                             type="string"   value="{12345}"/>
+       <Parameter  name="Coordinates"                   type="string"   value="{30}"/>
+    </ParameterList>
+
+   <!-- 
+  <Parameter name="coarse: type" type="string" value="RELAXATION"/>
+  <ParameterList name="coarse: params">
+    <Parameter name="relaxation: type"  type="string" value="Jacobi"/>
+    <Parameter name="relaxation: sweeps" type="int" value="1"/>
+    <Parameter name="relaxation: damping factor" type="double" value="0.00000001"/>
+  </ParameterList>
+        -->
+
+  <Parameter name="smoother: pre or post" type="string" value="post"/>
+
+  <Parameter name="smoother: pre type" type="string" value="LINESMOOTHING_BANDEDRELAXATION"/>
+  <ParameterList name="smoother: pre params">
+    <Parameter name="relaxation: type"  type="string" value="Jacobi"/>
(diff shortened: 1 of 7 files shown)


rstumin:
@cgcgcg  By the way, I fixed a little bug in the Stratimikos smoother that corresponds to the case when InitialGuessIsZero is false. In particular, the revised code builds a new thyraCor MultiVector (instead of wrapping the X vector) and a residual is explicitly computed.  One other funny thing about this new option is that there is no entry in the MueLu manual. In particular, the option is kicked off by adding a parameter to the stratimikos smoother, <Parameter name="smoother: recurMgOnFilteredA" type="bool" value="true"/>. There is a test and an example xml file in the test directory, but it feels funny not having something in the manual. Not sure if I should be doing documenting this in some other way ...

packages/muelu/test/scaling/recurMG.xml
```diff
@@ -0,0 +1,140 @@
...
+
+  <Parameter name="multigrid algorithm" type="string" value="sa"/>
+  <Parameter name="sa: damping factor"  type="double" value="1.33333"/>
+
+   <!-- THE BIG DATA DUMP -->
+    <ParameterList    name="export data">
+       <Parameter  name="P"                             type="string"   value="{1234}"/>
+       <Parameter  name="A"                             type="string"   value="{12345}"/>
+       <Parameter  name="Coordinates"                   type="string"   value="{30}"/>
+    </ParameterList>
```
cgcgcg, line 17: Does this even do anything?

packages/muelu/test/scaling/recurMG.xml
```diff
@@ -0,0 +1,140 @@
...
+    <Parameter        name="number of equations"                  type="int"      value="1"/>
+    <Parameter        name="max levels"                           type="int"      value="3"/>
+  <Parameter name="aggregation: type" type="string" value="uncoupled"/>
+  <Parameter name="aggregation: min agg size" type="int" value="2"/>
+  <Parameter name="aggregation: max agg size" type="int" value="299"/>
+  <Parameter name="aggregation: drop tol" type="double" value="0.000000000001"/>
+    <ParameterList    name="export data">
+       <Parameter  name="P"                             type="string"   value="{1234}"/>
+       <Parameter  name="A"                             type="string"   value="{1234}"/>
+    </ParameterList>
```
cgcgcg, line 101: Same question

packages/muelu/test/scaling/recurMG.xml
rstumin: I suppose I should remove it.  This is my trick to "do nothing" when I want to keep something around ... and am too lazy to comment it out.  I'll get rid of it.

packages/muelu/test/scaling/recurMG.xml
rstumin: Ditto

rstumin:
Put the "enableMueLu" command now in an ifdef which is generally off. If one tries to execute the new option "smoother: recurMGOnFilteredA", then a throw occurs with a handy error message indicating a define that needs to be set as well as a script 'patchLinkForRecurMG' that might be needed to avoid link errors. This script should be run after running cmake. There is a test (that is normally off as well) in trilinos/packages/muelu/test/scaling with a sample xml file. The error message also indicates how to turn this test on when running cmake. One funny thing is that to use this option, one follows the standard procedure for a stratimikos smoother but includes the line        <Parameter name="smoother: recurMgOnFilteredA" type="bool" value="true"/> within the smoother: params sublist.  At present there is nothing in the MueLu manual actually describing this.

cgcgcg (approved):

rstumin:
@cgcgcg would you mind taking another look. The revised 'ifdef' version is now included and it seems to be doing reasonably well in testing.

rstumin:
@cgcgcg  ... I didn't see that you already approved. Ugh ... should I unassign you now?

cgcgcg:
@rstumin No worries. Just let the autotester do its thing.

PR #9213, opened 2021-06-03
Panzer: fix doxygen tag file

rppawlo:
Thanks to Curt for finding this!

## Motivation
doxygen file had a cut and paste error for path to phalanx tags


ccober6 (approved):
Looks good!

PR #9219, opened 2021-06-04
MueLu: Avoid CMake warning about source file extensions

cgcgcg:
@trilinos/muelu 

## Motivation
CMake >= 3.20 complains about missing extensions for source files https://cmake.org/cmake/help/git-stage/policy/CMP0115.html


jhux2 (approved):
lgtm

Issue #8761, opened 2021-02-15
Belos:  Failing tests without UVM


iyamazaki:
## Enhancement
@trilinos/belos

This issue is to keep track of failing tests in Belos/Tpetra when UVM is turned off

This is how I configured Trilinos on Vortex:

> cmake \
\
-DCMAKE_INSTALL_PREFIX:FILEPATH=../install/trilinos-uvm \
-DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake \
-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
\
-DCMAKE_BUILD_TYPE:STRING=RELEASE \
-DTrilinos_ENABLE_EXPLICIT_INSTANTIATION:BOOL=ON \
-DTrilinos_ENABLE_ALL_PACKAGES:BOOL=OFF \
\
-D TPL_ENABLE_MPI=ON \
-D Trilinos_ENABLE_ALL_OPTIONAL_PACKAGES:BOOL=OFF \
-D Teuchos_ENABLE_COMPLEX:BOOL=ON \
-D Trilinos_ENABLE_COMPLEX:BOOL=ON \
-D Trilinos_ENABLE_OpenMP:BOOL=ON \
-D Trilinos_ENABLE_EXAMPLES:BOOL=ON \
-D Trilinos_ENABLE_TESTS:BOOL=ON \
-D Trilinos_ENABLE_FLOAT:BOOL=ON \
\
-D Trilinos_ENABLE_Belos:BOOL=ON \
  -D Belos_ENABLE_Kokkos:BOOL=ON \
  -D Belos_ENABLE_TSQR:BOOL=ON \
\
-D Trilinos_ENABLE_Amesos2:BOOL=ON \
-D TPL_ENABLE_SuperLU:BOOL=ON \
  -D SuperLU_INCLUDE_DIRS:PATH=${superlu_dir}/SRC \
  -D SuperLU_LIBRARY_DIRS:PATH=${superlu_dir}/lib \
\
-D Trilinos_ENABLE_AztecOO:BOOL=ON \
-D Trilinos_ENABLE_Teuchos:BOOL=ON \
-D Trilinos_ENABLE_Ifpack:BOOL=ON \
-D Trilinos_ENABLE_Epetra:BOOL=ON \
-D Trilinos_ENABLE_EpetraExt:BOOL=ON \
[...]

PR #9220, opened 2021-06-04
Belos & Amesos2: changes following Tpetra_UVM_Removal branch

iyamazaki:
@trilinos/tpetra, @trilinos/belos, and @trilinos/amesos2 

## Motivation
This PR contains changes needed for Belos & Amesos2, following up the recent changes in Tpetra to remove UVM dependencies [(PR#9038)](https://github.com/trilinos/Trilinos/pull/9038)

## Related Issues

Closes [Issue#8761](https://github.com/trilinos/Trilinos/issues/8761)

## Testing
Tested on Vortex with/without UVM. For Amesos2, we tested with KLU2 and SuperLU, using MPI enabled.


jennloe:
@hkthorn ??

csiefer2:
As a Tpetroid, the Belos changes look fine to me!

lucbv (approved):
For some reason the auto-tester is not testing this so I am approving hoping this will make the auto-tester run. The AT:AUTO-MERGE is not used so this should not cause a problem

hkthorn:
@csiefer2 I added you so you can approve this merge request.

jhux2:
@lucbv Ross has already reported more AT woes on the usual issue.

csiefer2:
@jhux2 I really need to write that autoretester...

csiefer2 (approved):

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -131,35 +136,47 @@ namespace Amesos2 {
      * \exception std::runtime_error Thrown if there is an error while extracting
      * row values from the underlying matrix.
      */
+    #ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
ndellingwood, line 139: @iyamazaki thanks for the work on this PR, I have a couple questions on some Amesos2 changes:

1. There are several locations where routines like this `getCrs` are guarded based on enabled deprecated code but there is not an else branch, why does these need to be removed? I'm not involved with the Tpetra updates, it's not clear to me why/if this is necessary to coincide with those changes or just considered clean up
2. Can you confirm which solvers were enabled in testing, and was this PR tested with/without MPI? Some of the remnant non-Kokkos view routines may be necessary to support matrix inputs with non-contiguous GIDs, which are only tested with MPI enabled. If you could confirm the following KLU2, Basker and SuperLU tests ran and passed with these changes, that would help me feel comfortable with the removals (I think only KLU2 is enabled in PR testing):
https://github.com/trilinos/Trilinos/blob/a6d3274e2f75876e7d682e34f92b12dded96f014/packages/amesos2/test/solvers/Basker_UnitTests.cpp#L389
https://github.com/trilinos/Trilinos/blob/a6d3274e2f75876e7d682e34f92b12dded96f014/packages/amesos2/test/solvers/KLU2_UnitTests.cpp#L419
https://github.com/trilinos/Trilinos/blob/a6d3274e2f75876e7d682e34f92b12dded96f014/packages/amesos2/test/solvers/Superlu_UnitTests.cpp#L374
https://github.com/trilinos/Trilinos/blob/a6d3274e2f75876e7d682e34f92b12dded96f014/packages/amesos2/test/solvers/ShyLUBasker_UnitTests.cpp#L425

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -131,35 +136,47 @@ namespace Amesos2 {
...
+                            EStorage_Ordering ordering=ARBITRARY,
+                            EDistribution distribution=ROOTED) const; // This was placed as last argument to preserve API
 
 
     /**
      * Convenience overload for the getCrs function that uses an enum
      * to describe some of the most basic distributions that could be
      * desired.
      */
+    #ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
ndellingwood, line 164: Ditto

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -189,34 +206,46 @@ namespace Amesos2 {
      * \exception std::runtime_error Thrown if there is an error while extracting
      * row values from the underlying matrix.
      */
+    #ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
ndellingwood, line 209: Ditto

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -189,34 +206,46 @@ namespace Amesos2 {
...
+                            const Teuchos::Ptr<const Tpetra::Map<local_ordinal_t,global_ordinal_t,node_t> > colmap,
+                            EStorage_Ordering ordering=ARBITRARY,
+                            EDistribution distribution=ROOTED) const; // This was placed as last argument to preserve API
 
     /**
      * Convenience overload for the getCcs function that uses an enum
      * to describe some of the most basic distributions that could be
      * desired.
      */
+    #ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
ndellingwood, line 233: Ditto

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -297,23 +328,25 @@ namespace Amesos2 {
 
   private:
 
+    #ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
ndellingwood, line 331: Ditto

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -325,6 +358,7 @@ namespace Amesos2 {
          EStorage_Ordering ordering,
          no_special_impl nsi) const;
 
+    #ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
ndellingwood, line 361: Ditto

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -353,23 +388,25 @@ namespace Amesos2 {
         EStorage_Ordering ordering,
         row_access ra) const;
 
+    #ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
ndellingwood, line 391: Ditto

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
```diff
@@ -381,6 +418,7 @@ namespace Amesos2 {
          EStorage_Ordering ordering,
          no_special_impl nsi) const;
 
+    #ifdef TPETRA_ENABLE_DEPRECATED_CODE
```
ndellingwood, line 421: Ditto

packages/amesos2/src/Amesos2_MatrixAdapter_decl.hpp
csiefer2: This is an *excellent* point.  The path I took in Ifpack2 is to use the deprecated Tpetra interfaces *only* when I was inheriting from a Tpetra object which had deprecated interfaces.  Since Amesos2 is not doing this, those interfaces can probably go.

iyamazaki: Thank you for the review, @ndellingwood!! The most of the changes for Amesos2 are because Tpetra is switching to return the local Crs as Kokkos::views instead of Teuchos::ArrayViews. There are public slides that summarize the MV changes, but I am not sure if there are slides about Crs changes. I can ask @kddevin, if they would help.

I have been testing with KLU2 and SuperLU, but not with Baskers, using MPI. I updated the motivation above, and the [Issue#8761](https://github.com/trilinos/Trilinos/issues/8761) with my Amesos2 configure.

iyamazaki: Thank you for the inputs, @csiefer2. For now, we kept the old interfaces, which use the deprecated Tpetra interfaces, as deprecated codes because I was not sure if any solvers or other codes use these old adapter interfaces. So, if they do, these solvers need to enable the Tpetra's deprecated code. But, I can just remove them, if that is better!

brian-kelley:
@iyamazaki It looks like Amesos2_KLU2_def.hpp does this around line 520:
```
Util::get_ccs_helper_kokkos_view<
        MatrixAdapter<Matrix>,host_value_type_array,host_ordinal_type_array,host_ordinal_type_array>
        ::do_get(this->matrixA_.ptr(), host_nzvals_view_, host_rows_view_, host_col_ptr_view_,
            nnz_ret, ROOTED, ARBITRARY, this->rowIndexBase_);
```
but Matrix is an Epetra_CrsMatrix when the error happens, which explains why GO was int. I think ``get_cxs_helper_kokkos_view`` needs a separate codepath for Epetra objects.

iyamazaki:
Thank you, @brian-kelley. That was very helpful. I think I forgot fixing the changes I made to the MatrixTraits for Epetra. I am still building to reproduce the errors, but will try fixing it.

jennloe:
@trilinos/framework Could we please give @iyamazaki write access?  Looks like he is a member of the Belos team, but not Trilinos / Developers.

csiefer2 (approved):

csiefer2:
> 
> 
> @trilinos/framework Could we please give @iyamazaki write access? Looks like he is a member of the Belos team, but not Trilinos / Developers.

Taken care of.

csiefer2 (dismissed):

csiefer2 (approved):

packages/amesos2/src/Amesos2_TpetraMultiVecAdapter_def.hpp
```diff
@@ -208,10 +202,10 @@ namespace Amesos2 {
           const size_t lclNumRows = redist_mv.getLocalLength();
           for (size_t j = 0; j < redist_mv.getNumVectors(); ++j) {
             auto av_j = av(lda*j, lclNumRows);
-            auto X_lcl_j_2d = redist_mv.getLocalViewHost();
+            auto X_lcl_j_2d = redist_mv.getLocalViewHost(Tpetra::Access::ReadOnly);
```
brian-kelley, line 205: I don't think ``X_lcl_j_2d`` is necessary here since you already have ``contig_local_view_2d`` above.

packages/amesos2/src/Amesos2_TpetraMultiVecAdapter_def.hpp
```diff
@@ -208,10 +202,10 @@ namespace Amesos2 {
...
             auto X_lcl_j_1d = Kokkos::subview (X_lcl_j_2d, Kokkos::ALL (), j);
 
-            using val_type = typename decltype( X_lcl_j_1d )::value_type;
+            using val_type = typename std::remove_const<typename decltype( X_lcl_j_1d )::value_type>::type;
```
brian-kelley, line 208: Could you use ``typename multivec_t::impl_scalar_type`` instead for readability?

packages/amesos2/src/Amesos2_TpetraMultiVecAdapter_def.hpp
```diff
@@ -480,17 +470,17 @@ namespace Amesos2 {
           const size_t lclNumRows = redist_mv.getLocalLength();
           for (size_t j = 0; j < redist_mv.getNumVectors(); ++j) {
             auto av_j = new_data(lda*j, lclNumRows);
-            auto X_lcl_j_2d = redist_mv.getLocalViewHost();
+            auto X_lcl_j_2d = redist_mv.getLocalViewHost(Tpetra::Access::ReadOnly);
             auto X_lcl_j_1d = Kokkos::subview (X_lcl_j_2d, Kokkos::ALL (), j);
 
-            using val_type = typename decltype( X_lcl_j_1d )::value_type;
+            using val_type = typename std::remove_const<typename decltype( X_lcl_j_1d )::value_type>::type;
```
brian-kelley, line 476: Same as above - could use multivec_t::impl_scalar_type

brian-kelley (changes requested):

ndellingwood (approved):
Amesos2 changes look fine (a few remnant commented lines can be cleaned up later) and testing passes, thanks for the hard work @iyamazaki !

iyamazaki:
Thank you for the comments, @brian-kelley. I have the changes, but I am thinking to push them as a separate PR (since auto-tester seems to be struggling right now). Or, I can create an issue with a list of changes if @ndellingwood also has some inputs. What do you think?

brian-kelley (approved):

brian-kelley:
@iyamazaki That sounds fine. I just approved this, but my comments from the first review are still up there.

william76:
@prwolfe @allevin 
I got asked about this PR... the approval boxes got checked yesterday around 8am after the testing passed but the pre-merge inspection flag seems to still be unset.

Is it possible that there are just a bunch of jobs in this tester's swim lane and we need to wait a bit more for it to cycle around or could something else be blocking it?

@csiefer2 Paul or Aaron may have more insight into the delay than I do... but in the past my experience has generally been that the tester needs to cycle around and do its check of the PR to set the final flags.  Still, you approved this about 26 hours ago which seems long... but builds also take a long time too so if there's other stuff pending.... 🤷🏻‍♂️

prwolfe:
@william76 - we are currently struggling with the change in the proxy system and it's implications when run through python. Basically I can get the autotester to talk to GitHub or to talk with Jenkins, but not both as we need.

srajama1 (commented):
Not sure why Amesos2 is still showing as locked, approving to see if that fixes it.

PR #9239, opened 2021-06-07
EXODUS: Fix the fortran units tests to avoid grepping stdout

gsjaardema:
@trilinos/seacas 

## Motivation

This eliminates the grepping of stdout and stderr for known text and instead checks the status output of all executions and ncdumps and diffs.  This will be much more robust since it doesn't fail if there is extraneous output written to stdout/stderr by mpiexec or jsrun or some other extra junk.


ccober6 (approved):
Looks good!

PR #9240, opened 2021-06-07
Tpetra - fix failing test in block crs

kyungjoo-kim:
@trilinos/tpetra

## Motivation

A BlockCrs test is failing with deprecated code OFF, UVM OFF. The problem is to use const version of accessor when the test is modifying the values of the matrix. This PR changes the accessor from const to non const.

## Testing

Tested on kokkos-dev-2.


csiefer2 (approved):

brian-kelley (approved):

kddevin (approved):
Thanks, @kyungjoo-kim

PR #9258, opened 2021-06-09
Tpetra warnings

bathmatt:
I've removed unused typedefs that were throwing warnings in gcc-9, they were bugging me


rppawlo (approved):
Thanks @bathmatt !

PR #9264, opened 2021-06-10
Adds more Ifpack2 preconditioned tests for different Belos solvers

hkthorn:
@trilinos/ifpack2 @trilinos/belos

## Motivation

The solvers tested in Ifpack2 have been limited to CG, GMRES, TFQMR
and their variants.  To achieve better code coverage for Tpetra, these
tests have been extended to include more Belos solvers, such as
GCRODR, MINRES, and BiCGStab.  This required:

- The testing has also been extended so that a "side" can be selected
for left or right preconditioning.  At this time, there is no 2-sided
preconditioning in testing, and this can be an extension for later.
- The solvers are now obtained using the solver factory that is
available for Tpetra, replacing the list of conditional statements
and simplifying the driver for these tests. That required a name change
for most of the xml files, as the names used in the previous conditional
statements did not conform to the naming used by the solver factory.

## Testing

This purpose of this pull request is to increase code coverage between Belos/Ifpack2/Tpetra.


cgcgcg (approved):
LGTM

PR #9266, opened 2021-06-10
Tpetra: Enable more SYCL tests

masterleinad:
@trilinos/tpetra

## Motivation
These tests were only added for the HIP backend in #9090 so I missed them in #9086.

## Testing
Tested on Intel GPUs enabling `Tpetra`.

## Additional Information
Some tests were disabled because they require querying the amount of free GPU memory which is not part of the `SYCL` specifications.
Also, recursive function calls are not allowed in `SYCL` kernels, so I replaced `heapify` with an iterative implementation.

### DIFF
--- a/packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrsMeshDatabase.hpp
+++ b/packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrsMeshDatabase.hpp
@@ -117,16 +117,18 @@ namespace BlockCrsTest {
   template<typename T1, typename T2>
   KOKKOS_INLINE_FUNCTION
   static void heapify(T1 *v, T2 n, T2 i) {
-    T2 largest = i;
-    T2 l = 2*i + 1;
-    T2 r = 2*i + 2;
+    while (true) {
+      T2 largest = i;
+      T2 l = 2*i + 1;
+      T2 r = 2*i + 2;
 
-    if (l < n && v[l] > v[largest]) largest = l;
-    if (r < n && v[r] > v[largest]) largest = r;
-    if (largest != i) {
+      if (l < n && v[l] > v[largest]) largest = l;
+      if (r < n && v[r] > v[largest]) largest = r;
+      if (largest == i)
+        break;
       // swap
       T1 tmp = v[i]; v[i] = v[largest]; v[largest] = tmp;
-      heapify(v, n, largest);
+      i = largest;
     }
   }
 
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_SYCL.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_SYCL.hpp
@@ -1,0 +1,826 @@
+// @HEADER
+// ***********************************************************************
+//
+//          Tpetra: Templated Linear Algebra Services Package
+//                 Copyright (2008) Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the Corporation nor the names of the
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL SANDIA CORPORATION OR THE
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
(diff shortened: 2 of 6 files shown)


packages/tpetra/core/test/ImportExport2/ImportExport2_UnitTests.cpp
```diff
@@ -2845,11 +2845,20 @@ TEUCHOS_UNIT_TEST_TEMPLATE_2_DECL( Import_Util,GetTwoTransferOwnershipVector, LO
...
 
-#define UNIT_TEST_GROUP_SC_LO_GO( SC, LO, GO )                   \
+#define UNIT_TEST_GROUP_SC_LO_GO_COMMON( SC, LO, GO )                   \
   TEUCHOS_UNIT_TEST_TEMPLATE_3_INSTANT( CrsMatrixImportExport, doImport, LO, GO, SC ) \
   TEUCHOS_UNIT_TEST_TEMPLATE_3_INSTANT( FusedImportExport, doImport, LO, GO, SC ) \
-  TEUCHOS_UNIT_TEST_TEMPLATE_3_INSTANT( Import_Util, UnpackAndCombineWithOwningPIDs, LO, GO, SC ) \
-  TEUCHOS_UNIT_TEST_TEMPLATE_3_INSTANT( FusedImportExport, MueLuStyle, LO, GO, SC )
+  TEUCHOS_UNIT_TEST_TEMPLATE_3_INSTANT( Import_Util, UnpackAndCombineWithOwningPIDs, LO, GO, SC )
+
+// FIXME_SYCL requires quering free device memory
```
kddevin, line 2853: Please add better comments here.  What is this code doing, and what needs to be fixed?  Why is the _COMMON macro needed?  Does the MueLuStyle test not work with SYCL, and if so, why not?   Better details in the code for FIXME issues will help the future fixers know what needs to be done.  Thanks.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_SYCL.hpp
```diff
@@ -0,0 +1,826 @@
+// @HEADER
```
kddevin, line 1: Sigh.  Yet another copy of this MatrixMatrix code.  @trilinos/tpetra has to fix this duplication.  How does this SYCL version differ from other versions?  We'd like to capture this info so that we can someday merge all these duplicate files.  Thanks.

Note, I am assuming that this file is merely a copy of other MatrixMatrix files; I have not reviewed the code.  If there are significant differences, one of @trilinos/tpetra 's MatrixMatrix experts should review.

packages/tpetra/core/test/MatrixMatrix/MatrixMatrix_UnitTests.cpp
```diff
@@ -2222,6 +2221,16 @@ TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(Tpetra_MatMatAdd, locally_unsorted, SC, LO, GO
   TEUCHOS_UNIT_TEST_TEMPLATE_4_INSTANT(Tpetra_MatMatAdd, different_col_maps, SC, LO, GO, NT) \
   TEUCHOS_UNIT_TEST_TEMPLATE_4_INSTANT(Tpetra_MatMatAdd, different_index_base, SC, LO, GO, NT)
 
+// FIXME_SYCL requires quering free device memory
```
kddevin, line 2224: Ditto to above.

packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrsMeshDatabase.hpp
```diff
@@ -117,16 +117,18 @@ namespace BlockCrsTest {
...
   KOKKOS_INLINE_FUNCTION
   static void heapify(T1 *v, T2 n, T2 i) {
-    T2 largest = i;
-    T2 l = 2*i + 1;
-    T2 r = 2*i + 2;
-
-    if (l < n && v[l] > v[largest]) largest = l;
-    if (r < n && v[r] > v[largest]) largest = r;
-    if (largest != i) {
+    while (true) {
```
kddevin, line 120: Interesting that recursion is not allowed.

kddevin (changes requested):
Thanks for the work on this PR.  My comments mostly request more info -- better comments in the code or explanations in the PR.  Thanks.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_SYCL.hpp
masterleinad: Yes, this is a copy of the other `MatrixMatrix` files replacing the execution/memory space with the ones corresponding to `SYCL`.
I would think that it's not too difficult to avoid the code duplication for these files but that's out-of-scope of this pull request IMHO.

packages/tpetra/core/test/ImportExport2/ImportExport2_UnitTests.cpp
masterleinad: `FusedImportExport` tries to query for free device memory somewhere in Kokkos Kernels and this is not supported in the SYCL specifications.
To preserve the definition of `UNIT_TEST_GROUP_SC_LO_GO`, I introduced `UNIT_TEST_GROUP_SC_LO_GO_COMMON` to only disable the test mentioned above for `SYCL`.

masterleinad:
@kddevin I improved the in-code comments in the places you pointed out.

packages/tpetra/core/ext/TpetraExt_MatrixMatrix_SYCL.hpp
kddevin: I agree; this duplication should be handled by @trilinos/tpetra

kddevin (dismissed):
Thanks for the beefed up comments.  Is there a KokkosKernels issue regarding the SYCL incompatibility?  If so, please cite the number here and this PR in the KokkosKernels issue so that someone remembers to re-enable the tests when the problem is removed from KokkosKernels.

masterleinad:
https://github.com/kokkos/kokkos-kernels/issues/1062 is related to the disabled tests.

masterleinad:
@kddevin @brian-kelley Ping. Is there anything you want me to change here?

brian-kelley (approved):
@masterleinad No, this looks good to me. Thanks!

kddevin (approved):

masterleinad:
Looks like the `trilinos-autotester` was successful.

masterleinad:
Can we merge this now?

masterleinad:
@kddevin @brian-kelley It seems that this PR needs to be retested before it can get merged. Maybe, also set the `AT: AUTOMERGE` label?

PR #9268, opened 2021-06-10
Zoltan2:  small coloring cleanup

kddevin:
@trilinos/zoltan2 

## Motivation

This PR enables tests that were needlessly disabled when Galeri is not built; the tests run without Galeri.
This PR also makes colorMethod names in coloring1.cpp test consistent with the actual coloring method names.

## Testing

tested on cee-lan with gcc


egboman (approved):
Good change. Thanks, Karen.

PR #9271, opened 2021-06-10
Ifpack2: fix refcount related errors

brian-kelley:
- RILUK: move some host views into smaller scope that doesn't overlap with
scope of device views.
- BlockTriDi: make A values view unmanaged so that holding it doesn't
increment use count. This code is still responsible for managing syncs.

@trilinos/tpetra 

## Motivation

Make RILUK, BlockTriDiContainer obey the new Tpetra DualView semantics (don't hold host and device view simultaneously). Fixes tests for UVM off: ``Ifpack2_RILUK_HTS_hb_belos``, ``Ifpack2_BlockTriDiContainerUnitAndPerfTests``

### DIFF
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -1471,8 +1471,7 @@ namespace Ifpack2 {
       using impl_type = ImplType<MatrixType>;
       using local_ordinal_type_1d_view = typename impl_type::local_ordinal_type_1d_view;
       using size_type_1d_view = typename impl_type::size_type_1d_view;
-      using impl_scalar_type_1d_view_tpetra = typename impl_type::impl_scalar_type_1d_view_tpetra;
-
+      using impl_scalar_type_1d_view_tpetra = Unmanaged<typename impl_type::impl_scalar_type_1d_view_tpetra>;
       // rowptr points to the start of each row of A_colindsub.
       size_type_1d_view rowptr, rowptr_remote;
       // Indices into A's rows giving the blocks to extract. rowptr(i) points to
@@ -3784,6 +3783,13 @@ namespace Ifpack2 {
                                   "The seq method for applyInverseJacobi, " <<
                                   "which in any case is for developer use only, " <<
                                   "does not support norm-based termination.");
+      const bool device_accessible_from_host = Kokkos::SpaceAccessibility<
+        Kokkos::DefaultHostExecutionSpace, node_memory_space>::accessible;
+      TEUCHOS_TEST_FOR_EXCEPTION(is_seq_method_requested && !device_accessible_from_host,
+                                 std::invalid_argument,
+                                 "The seq method for applyInverseJacobi, " <<
+                                 "which in any case is for developer use only, " <<
+                                 "only supports memory spaces accessible from host.");
 
       // if workspace is needed more, resize it
       const size_type work_span_required = num_blockrows*num_vectors*blocksize;
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_def.hpp
@@ -98,13 +98,19 @@ namespace Ifpack2 {
     impl_->tpetra_importer = Teuchos::null;
     impl_->async_importer  = Teuchos::null;
     
-    if (importer.is_null()) // there is no given importer, then create one
-      if (useSeqMethod) 
+    if (useSeqMethod)
+    {
+      if (importer.is_null()) // there is no given importer, then create one
         impl_->tpetra_importer = BlockTriDiContainerDetails::createBlockCrsTpetraImporter<MatrixType>(impl_->A);
-      else 
-        impl_->async_importer = BlockTriDiContainerDetails::createBlockCrsAsyncImporter<MatrixType>(impl_->A);
-    else 
-      impl_->tpetra_importer = importer; // if there is a given importer, use it
+      else
+        impl_->tpetra_importer = importer; // if there is a given importer, use it
+    }
+    else
+    {
+      //Leave tpetra_importer null even if user provided an importer.
+      //It is not used in the performant codepath (!useSeqMethod)
+      impl_->async_importer = BlockTriDiContainerDetails::createBlockCrsAsyncImporter<MatrixType>(impl_->A);
+    }
 
     // as a result, there are 
     // 1) tpetra_importer is     null , async_importer is     null (no need for importer)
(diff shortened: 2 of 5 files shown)


packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -1471,7 +1471,12 @@ namespace Ifpack2 {
       using impl_type = ImplType<MatrixType>;
       using local_ordinal_type_1d_view = typename impl_type::local_ordinal_type_1d_view;
       using size_type_1d_view = typename impl_type::size_type_1d_view;
-      using impl_scalar_type_1d_view_tpetra = typename impl_type::impl_scalar_type_1d_view_tpetra;
+      using scalar_1d_managed_view_tpetra = typename impl_type::impl_scalar_type_1d_view_tpetra;
+      using impl_scalar_type_1d_view_tpetra = Kokkos::View<
+        typename scalar_1d_managed_view_tpetra::data_type,
+        typename scalar_1d_managed_view_tpetra::array_layout,
+        typename scalar_1d_managed_view_tpetra::device_type,
+        Kokkos::MemoryTraits<Kokkos::Unmanaged>>;
```
kyungjoo-kim, line 1479: I recommend to use 
```
using impl_scalar_type_1d_view_tpetra = typename impl_type::impl_scalar_type_1d_view_tpetra;
using impl_scalar_type_1d_view_tpetr_unmanaged = Unmanaged<impl_scalar_Type_1d_view_tpetra>;
```

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3812,9 +3817,11 @@ namespace Ifpack2 {
...
       vector_type_3d_view pmv(work.data(), num_blockrows, blocksize, num_vectors);
       const auto XX = X.getLocalViewDevice(Tpetra::Access::ReadOnly);
-      const auto YY = Y.getLocalViewDevice(Tpetra::Access::ReadWrite);
-      const auto ZZ = Z.getLocalViewDevice(Tpetra::Access::ReadWrite);
-      if (is_y_zero) Kokkos::deep_copy(YY, zero);
+      if (is_y_zero)
+      {
+        const auto YY = Y.getLocalViewDevice(Tpetra::Access::OverwriteAll);
+        Kokkos::deep_copy(YY, zero);
+      }
```
kyungjoo-kim, line 3824: I recommend to leave YY and ZZ as they are. I guessed that you changed this because the new tpetra code triggers host/device alive error. The following sequence of the code should run entirely on device. If there is something that prevents it, it means that we do unnecessary host view copy. It will cause a performance problem. We probably need to figure out where a local host view becomes necessary. We can talk offline about this.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3842,11 +3849,14 @@ namespace Ifpack2 {
 
               // y := x - R y
               Z.doImport(Y, *tpetra_importer, Tpetra::REPLACE);
+              const auto YY = Y.getLocalViewDevice(Tpetra::Access::ReadWrite);
+              const auto ZZ = Z.getLocalViewDevice(Tpetra::Access::ReadWrite);
```
kyungjoo-kim, line 3853: Here as well. If we need to this way to remove host/device alive error, it means that we use a host view somewhere in this code or tpetra code. If possible, we need to remove that case.

kyungjoo-kim (changes requested):
My main concern is that applications may see performance degrade with this change.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3842,11 +3849,14 @@ namespace Ifpack2 {
 
               // y := x - R y
               Z.doImport(Y, *tpetra_importer, Tpetra::REPLACE);
```
kyungjoo-kim, line 3851: Is doImport the trouble guy using a host local view ? At my glance, I do not use Tpetra routine in this function except for that. If so, the code path is for testing only. We can probably create a workaround and keep the code as it is. For instance, we can enable the sequential path in this and testing code when UVM is enabled only.

brian-kelley: @kyungjoo-kim Yes it is - doImport gets a ReadWrite hostview of Z and that was causing the exception.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
brian-kelley: That's great, I didn't see that at the top.

brian-kelley:
@kyungjoo-kim I added this check:
```
      const bool device_accessible_from_host = Kokkos::SpaceAccessibility<
        Kokkos::DefaultHostExecutionSpace, node_memory_space>::accessible;
      TEUCHOS_TEST_FOR_EXCEPTION(is_seq_method_requested && !device_accessible_from_host,
                                 std::invalid_argument,
                                 "The seq method for applyInverseJacobi, " <<
                                 "which in any case is for developer use only, " <<
                                 "only supports memory spaces accessible from host.");
```
and in the test, disabled ``seq_method == true`` for cases without UVM semantics. But this path still runs, because if the user provides a non-null importer to the container, that is used as the ``tpetra_importer``:
```
//in BlockTriDiContainer::initInternal()
101     if (importer.is_null()) // there is no given importer, then create one
102       if (useSeqMethod)                    
103         impl_->tpetra_importer = BlockTriDiContainerDetails::createBlockCrsTpetraImporter<MatrixType>(impl_->A);
104       else      
105         impl_->async_importer = BlockTriDiContainerDetails::createBlockCrsAsyncImporter<MatrixType>(impl_->A);
106     else        
107     {           
108       std::cout << "** Using user-provided importer.\n";
109       impl_->tpetra_importer = importer; // if there is a given importer, use it
110     }
```
Should this just ignore the user's importer if they didn't construct BlockTriDiContainer with ``useSeqMethod == true``? That bool isn't passed down to ImplObject. The ImplObject just looks at the importer:
```
const bool is_seq_method_requested = !tpetra_importer.is_null();
```

brian-kelley:
I made that change and the tests pass again for CudaSpace and CudaUVMSpace.

kyungjoo-kim:
@brian-kelley I think that we can just show a warning message and ignore the given importer from users for that case. 

The reason that I want to leave the code as it is much is because I experienced unexpected performance slow-down when the binary size changes. We probably do not want to change the performant case in order to handle hardly used and develop only sequential case.

However, it is still weird to me. The custom importer in the BlockTridiag does not need to use host view while Tpetra version needs to use a host view. In such a case, Tpetra version would be difficult to beat this custom import.

kyungjoo-kim (approved):

PR #9279, opened 2021-06-11
Tpetra: Use HIPSpace for HIPWrapperNode

brian-kelley:
Use HIPSpace, not HIPHostPinnedSpace. This required only copying a few changes
from Cuda to HIP in the matrix-matrix kernels. All tests still pass with ROCm 4.2.

@trilinos/tpetra 

## Motivation

Using HIPHostPinnedSpace was a temporary way to get UVM-like semantics with HIP, at the cost of much worse performance. Now that UVM removal is done within Tpetra, this no longer needed.


lucbv (approved):
This looks, fine the change of space for the HIP backend is important and the name changes to make device and host types and variables more obvious seems wise.

PR #9299, opened 2021-06-17
Tpetra: Fix compiling and running with SYCL and MPI

masterleinad:
@trilinos/tpetra

## Motivation
The changes in this pull request are necessary to pass the tests when building with `Tpetra` and `MPI` enabled using the `SYCL` backend in `Kokkos`. In particular, there were some changes necessary to run the PerformanceCGTest benchmark. This pull request also adds the corresponding changes for `HIP`.
It's debatable if we just want to allow using `SYCLSharedUSMSpace` for MPI communication or not, but I decided to mirror `CUDA` instead of `HIP` here since `SYCLSharedUSMSpace` corresponds to `SYCLSharedUSMSpace` whereas `HIPHostPinnedSpace` is somewhat different.

## Testing
I checked that the tests run correctly on Intel GPUs.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Distributor.hpp
+++ b/packages/tpetra/core/src/Tpetra_Distributor.hpp
@@ -2122,6 +2122,15 @@ namespace Tpetra {
        "See Trilinos GitHub issue #1088.");
 #endif // KOKKOS_ENABLE_CUDA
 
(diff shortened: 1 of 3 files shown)


brian-kelley (approved):

lucbv:
@brian-kelley any reason why we are not merging this?
Eventually it will have conflicts if we do not do it.

PR #9302, opened 2021-06-17
Tpetra: remove some deprecated stuff

brian-kelley:
- EpetraRowMatrix: don't call ArrayView version of getLocalRowCopy, and un-deprecate ExtractMyRowCopy since it can just call the Kokkos::View version.
- RowMatrixTransposer: don't call getLocalMatrix

@trilinos/tpetra 

## Motivation

Fix build of Tpetra (library+tests) with deprecated code off.

## Stakeholder Feedback

@jennloe ran into the build error of RowMatrixTransposer in a build of Albany. Our nightly deprecated-off build on ascicgpu031 has been hitting it too, but I think we all ignored that :P


iyamazaki (approved):
Thank you, @brian-kelley. They look good to me.

tasmith4 (approved):
Thanks @brian-kelley!

PR #9305, opened 2021-06-18
Amesos2: add new options for SuperLU

iyamazaki:
## Motivation

This PR is to add a few new options for SpTRSV with SuperLU, and also an option to call METIS before SuperLU


cgcgcg:
@trilinos/framework This PR had master set as base and was running the dev->master autotester. I don't think that's intended.

iyamazaki:
Thank you, @cgcgcg..

william76:
> @trilinos/framework This PR had master set as base and was running the dev->master autotester. I don't think that's intended.

Correct. Merging to `master` is restricted.

ndellingwood (approved):
thanks @iyamazaki !

PR #9310, opened 2021-06-18
Xpetra/MueLu testing: change proxy settings

cgcgcg:

## Motivation
Attempt to use the system proxy settings instead of hard-coding a proxy.


jhux2 (approved):
Lgtm.  Let's hope that's set in the environment.

cgcgcg:
@jhux2 It should be set. I already made the required change on geminga and pointed it to the new proxy. Let's see which machines keep posting...

PR #9315, opened 2021-06-19
Gracefully exist with 0 when build_stats.csv file is empty (#9314)

### LINKED ISSUE #9314: Build stats fails for PRs that don't build anything
## Bug Report
@bartlettroscoe @jjellio 

### Description
The build stats script seems to fail for PR builds that don't compile anything:
https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=9376784

bartlettroscoe:
Fixes #9314 (allows PRs that have no `*.timing` files to pass)

## How was this tested?

I manually verified the failure reported in #9314 and after the fix I manually verified that this fixes the test `TrilinosBuildStats_Results`.  The tick is that you need to delete all of the timing files with:

```
$ find . -name "*.timing" -exec rm {} \;
```

and then run the test `TrilinosBuildStats_Results`.  That results in an empty `build_stats.csv` file (with no column headers).

It seems harmless the gracefully return 0 and printing `No build statistics to summarize!` to be clear what has happened.


cgcgcg (approved):
LGTM

PR #9322, opened 2021-06-21
Xpetra/Zoltan2/Muelu HIP Support

crtrott:
@trilinos/tpetra @trilinos/zoltan2 @trilinos/xpetra @trilinos/muelu

This makes xpetra, Zoltan2 and MueLu compile with HIP. 
You need to add the cmake configure option:  Tpetra_ENABLE_HIPHostPinned:BOOL=ON though (the equivalent to the UVM setting). 

@trilinos/tpetra team: you may want to consider naming this option differently, make Tpetra_ENABLE_IMPL_HIPHostPinned for example or so. This is a temporary necessity until all the downstream packages get rid of UVM requirements. 

@trilinos/xpetra : All tests are passing so this looks pretty decent.

@trilinos/zoltan2 : some Coloring1 test mucks up the GPUs they are only rebooting now. We need to also if possible track the root cause down, but likely its what we found before, for example atomics on HIPHostPinnedSpace. Until we don't need UVM anymore we may just want to disable those tests when HIP is on.

@trilinos/muelu : I haven't yet added the corresponding #ifdef HIP stuff in the tests themselves where they have ifdef CUDA. 

@lucbv can you take a look at the muelu stuff and/or the disabling of the zoltan tests which break the GPUs?


crtrott:
Oh I don't quite know what the stuff with the Kokkos cmake warning is about. It was an error, but I made a warning for now.

crtrott:
Note many of the changes are related to using a proper Kokkos::Device instead of execution or memory space. This is since with the HIPHostPinned node type from Tpetra View<T*,HIP> is not assignable from View<T*,TpetraObject::device_type>. Since HIP implies the device Device<HIP,HIPSpace> while tpetra with that new option to simulate UVM like behavior uses Device<HIP,HIPHostPinnedSpace>

crtrott:
FYI: Running Zoltan2 with just Coloring excluded passes: 

```
bash-4.2$ ctest -R Zoltan2_ -E Zoltan2_Coloring
....
100% tests passed, 0 tests failed out of 118

Label Time Summary:
Zoltan2    = 1072.13 sec*proc (118 tests)

Total Test time (real) = 274.99 sec
```

csiefer2:
I'm really not a fan of adding more template parameters.

kddevin (changes requested):
Mostly just questions; one typo.  Thanks in advance for the Kokkos education.

jhux2:
@lucbv reports good progress on HIP support without requiring an extra template parameter.  He's leveraging @bathmatt's UVM removal in Ifpack2, Xpetra, MueLu, and Zoltan2, and has been using this PR as inspiration.

jhux2:
I've marked this PR as WP in favor of @lucbv's PRs.

lucbv:
This has been superseded by PR #9466 #9488 and #9580 I think we can close this.
Feel free to re-open if needed.

PR #9338, opened 2021-06-24
Muelu tpetra dep

bathmatt:
@trilinos/muelu 
this is a lot of changes but have muelu working with the new tpetra interface and have many UVM issues fixed.  

This doesn't fix all of the UVM issues but it fixes a lot so I want to get this going and merged once the zoltan2 one gets fixes.

### DIFF
--- a/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
+++ b/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
@@ -228,8 +228,8 @@ namespace MueLu {
     }
     else
       globalColsToZero = myColsToZero;
-    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(globalColsToZero->getDeviceLocalView(Xpetra::Access::ReadWrite),dirichletDomain);
-    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(myColsToZero->getDeviceLocalView(Xpetra::Access::ReadWrite),dirichletCols);
+    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(globalColsToZero->getDeviceLocalView(Xpetra::Access::ReadOnly),dirichletDomain);
+    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(myColsToZero->getDeviceLocalView(Xpetra::Access::ReadOnly),dirichletCols);
   }
 
 #endif
--- a/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
+++ b/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
@@ -204,8 +204,8 @@ namespace MueLu {
     TEUCHOS_ASSERT(dirichletDomain.extent(0) == domMap->getNodeNumElements());
     RCP<Xpetra::Vector<Scalar,LocalOrdinal,GlobalOrdinal,Node> > myColsToZero = Xpetra::VectorFactory<Scalar,LocalOrdinal,GlobalOrdinal,Node>::Build(colMap, /*zeroOut=*/true);
     // Find all local column indices that are in Dirichlet rows, record in myColsToZero as 1.0
-    auto myColsToZeroView = myColsToZero->getDeviceLocalView();
-    auto localMatrix = A.getLocalMatrix();
+    auto myColsToZeroView = myColsToZero->getDeviceLocalView(Xpetra::Access::ReadWrite);
+    auto localMatrix = A.getLocalMatrixDevice();
     Kokkos::parallel_for("MueLu:RefMaxwell::DetectDirichletCols", range_type(0,rowMap->getNodeNumElements()),
                          KOKKOS_LAMBDA(const LocalOrdinal row) {
                            if (dirichletRows(row)) {
@@ -228,8 +228,8 @@ namespace MueLu {
     }
     else
       globalColsToZero = myColsToZero;
-    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(globalColsToZero->getDeviceLocalView(),dirichletDomain);
-    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(myColsToZero->getDeviceLocalView(),dirichletCols);
+    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(globalColsToZero->getDeviceLocalView(Xpetra::Access::ReadWrite),dirichletDomain);
+    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(myColsToZero->getDeviceLocalView(Xpetra::Access::ReadWrite),dirichletCols);
   }
 
 #endif
@@ -1696,7 +1696,7 @@ namespace MueLu {
       typedef typename KCRS::values_type::non_const_type scalar_view_t;
 
       // Get data out of P_nodal_imported and D0.
-      auto localD0 = D0_Matrix_->getLocalMatrix();
+      auto localD0 = D0_Matrix_->getLocalMatrixHost();
 
       // Which algorithm should we use for the construction of the special prolongator?
       // Option "mat-mat":
@@ -1706,7 +1706,7 @@ namespace MueLu {
 
       if (skipFirstLevel_) {
 
-        auto localP = P_nodal_imported->getLocalMatrix();
+        auto localP = P_nodal_imported->getLocalMatrixHost();
 
         if (algo == "mat-mat") {
           RCP<Matrix> D0_P_nodal = MatrixFactory::Build(SM_Matrix_->getRowMap());
@@ -1717,7 +1717,7 @@ namespace MueLu {
 #endif
 
           // Get data out of D0*P.
-          auto localD0P = D0_P_nodal->getLocalMatrix();
+          auto localD0P = D0_P_nodal->getLocalMatrixHost();
 
           // Create the matrix object
           RCP<Map> blockColMap    = Xpetra::MapFactory<LO,GO,NO>::Build(P_nodal_imported->getColMap(), dim);
@@ -1743,7 +1743,7 @@ namespace MueLu {
                                  }
                                });
 
-          auto localNullspace = Nullspace_->getDeviceLocalView();
+          auto localNullspace = Nullspace_->getDeviceLocalView(Xpetra::Access::ReadOnly);
 
           // enter values
           if (D0_Matrix_->getNodeMaxNumRowEntries()>2) {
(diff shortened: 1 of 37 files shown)


jhux2 (approved):
Approving to let the AT run.

bathmatt:
@cgcgcg @jhux2  this is failing here

```
#13 0x00000000119960e4 in MueLu::NotayAggregationFactory<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaUVMSpace> >::localSpGEMM (this=0x40a6cbd0, A=..., B=..., 
    matrixLabel=..., C=...) at ../packages/muelu/src/Graph/PairwiseAggregation/MueLu_NotayAggregationFactory_def.hpp:913

```

It is calling into Kokkos Kernels here

```
    // Symbolic multiplication
    KokkosSparse::Experimental::spgemm_symbolic(&kh, A.numRows(),
                                                B.numRows(), B.numCols(),
                                                A.graph.row_map, A.graph.entries, false,
                                                B.graph.row_map, B.graph.entries, false,
                                                rowPtrC);

```

What kind of checks can i do outside of KK to get more info on why this is failing?  

I'm stuck a bit on how to proceed on this bug.

cgcgcg:
@lucbv

bathmatt:
new info, cuda-memcheck gets the test to pass, without cuda-memcheck it fails,

tested cuda-launch-blocking it fails in the same way

lucbv:
It seems to fail because of an access out of bounds, it is strange that this would happen because of the refactor for UVM removal though.
I guess turning on Kokkos' bound checking makes sense otherwise all other options are painful and will revolve around printf...
I could take a look at it but not before tomorrow.

bathmatt:
Really weird thing is that without UVM it passes as well...  I have a theory I'm testing but I don't believe this is something I caused, but has been floating around for a while.

lucbv:
Yes, that would not be too surprising, this Factory is fairly experimental and is not used in production so it could have issues that have not been discovered yet...

bathmatt:
Running bisect on my changes, will have a commit which triggered this error

bathmatt:
It was this commit that caused the error, I'm going to look a bit more into this

d0cf4a57ddc99aa7ac7657754ca377537e833491

bathmatt:
IT was this commit d0cf4a57ddc99aa7ac7657754ca377537e833491

can someone have a quick looksee to see if there is anything obvious,

bathmatt:
I think I've found it, it is the parallel for not doing what I thought it should, Let me fix my code,

lucbv:
You know the whole thing is pretty experimental still, we are not sure that we like this style of aggregation yet...
It could be left as is, without the parallel_for if it's a problem to fix it.

bathmatt:
OK, this test passes with and without UVM

packages/zoltan2/core/src/TpetraCrsColorer/Zoltan2_TpetraCrsColorer.hpp
```diff
@@ -323,11 +320,9 @@ TpetraCrsColorer<CrsMatrixType>::reconstructMatrixFitted(
...
       "  You must call the non-fitted version of this function.");
 
-  W.sync_device();
-
-  auto W_view_dev = W.getLocalViewDevice();
-  auto local_matrix = mat.getLocalMatrix();
-  auto local_graph = graph->getLocalGraph();
+  auto W_view_dev = W.getLocalViewDevice(Tpetra::Access::ReadOnly);
+  auto local_matrix = mat.getLocalMatrixDevice();
+  auto local_graph = graph->getLocalGraphDevice();
```
lucbv, line 325: Technically `local_matrix.graph` is cheaper as it will not recreate the local graph

lucbv (commented):
Overall this looks good but I think a few things can be improved for future maintenance.

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
```diff
@@ -204,8 +204,8 @@ namespace MueLu {
     TEUCHOS_ASSERT(dirichletDomain.extent(0) == domMap->getNodeNumElements());
     RCP<Xpetra::Vector<Scalar,LocalOrdinal,GlobalOrdinal,Node> > myColsToZero = Xpetra::VectorFactory<Scalar,LocalOrdinal,GlobalOrdinal,Node>::Build(colMap, /*zeroOut=*/true);
     // Find all local column indices that are in Dirichlet rows, record in myColsToZero as 1.0
-    auto myColsToZeroView = myColsToZero->getDeviceLocalView();
-    auto localMatrix = A.getLocalMatrix();
+    auto myColsToZeroView = myColsToZero->getDeviceLocalView(Xpetra::Access::ReadWrite);
```
cgcgcg, line 207: Wouldn't write-only be enough here?

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
```diff
@@ -228,8 +228,8 @@ namespace MueLu {
     }
     else
       globalColsToZero = myColsToZero;
-    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(globalColsToZero->getDeviceLocalView(),dirichletDomain);
-    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(myColsToZero->getDeviceLocalView(),dirichletCols);
+    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(globalColsToZero->getDeviceLocalView(Xpetra::Access::ReadWrite),dirichletDomain);
+    FindNonZeros<Scalar,LocalOrdinal,GlobalOrdinal,Node>(myColsToZero->getDeviceLocalView(Xpetra::Access::ReadWrite),dirichletCols);
```
cgcgcg, line 232: Isn't read-only enough?

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
```diff
@@ -1786,8 +1786,8 @@ namespace MueLu {
 
         NullspaceH_ = MultiVectorFactory::Build(P11_->getDomainMap(), dim);
 
-        auto localNullspace_nodal = Nullspace_nodal->getDeviceLocalView();
-        auto localNullspaceH = NullspaceH_->getDeviceLocalView();
+        auto localNullspace_nodal = Nullspace_nodal->getDeviceLocalView(Xpetra::Access::ReadOnly);
+        auto localNullspaceH = NullspaceH_->getDeviceLocalView(Xpetra::Access::ReadWrite);
```
cgcgcg, line 1790: Write-only?

packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_kokkos_def.hpp
```diff
@@ -596,11 +597,11 @@ namespace MueLu {
           {
             SubFactoryMonitor m2(*this, "MainLoop", currentLevel);
 
-            auto ghostedDiagView = ghostedDiag->getDeviceLocalView();
+            auto ghostedDiagView = ghostedDiag->getDeviceLocalView(Xpetra::Access::ReadWrite);
 
             CoalesceDrop_Kokkos_Details::ClassicalDropFunctor<LO, decltype(ghostedDiagView)> dropFunctor(ghostedDiagView, threshold);
             CoalesceDrop_Kokkos_Details::ScalarFunctor<typename ATS::val_type, LO, local_matrix_type, decltype(bndNodes), decltype(dropFunctor)>
-                scalarFunctor(kokkosMatrix, bndNodes, dropFunctor, rows, colsAux, valsAux, reuseGraph, lumping, threshold);
+	      scalarFunctor(A->getLocalMatrixDevice(), bndNodes, dropFunctor, rows, colsAux, valsAux, reuseGraph, lumping, threshold);
```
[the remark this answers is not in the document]
bathmatt: So, the problem is that this code goes between requiring device and host copies of the matrix, you can't get host copies when the device one is active therefore I have to free kokkosMatrix and regrab the matrix.  I'd be interested in other ways to do this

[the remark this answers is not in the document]
bathmatt: @lucbv I reduced the scope of the kokkosMatrix reset and reinitialize, the ones below are gone with the latest patch

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
bathmatt: I was not sure if all values are written or if some are just assumed to remain 0, can you confirm that 0s are written?

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
bathmatt: Fixed

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
bathmatt: Same as below, i am not sure that myColsToZero are 100% overwritten, can you confirm that it is.  I don't believe we can guarantee that initial values are 0

packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_kokkos_def.hpp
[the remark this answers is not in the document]
lucbv: Thanks for having a look!

bathmatt:
crap, fat fingered something, I'll push a fix, 

@jhux2 thoughts on @lucbv comment above?

For the zoltan fixes I'm going to push a zoltan2 only branch soon and then address these in this MR.

I don't want to rock that boat with the approval process.

packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
cgcgcg: Everything should be initialized to 0 when the MV is created here: https://github.com/trilinos/Trilinos/blob/d8c96d1ea19b5e60c1c6bc1b12c311f8b29b0ca0/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp#L1787
Entries that aren't modified are supposed to stay zero.

bathmatt:
@lucbv @cgcgcg I've addressed most the issues, and I've rebased off origin/develop.  there are some questions in the PR in the code segments to finish this up.  I think we can open another issue to address @lucbv comment and create a device version of GetDiagonalValues and use that so we wouldn't need to scope the kokkosMatrix life after this merges. 

Thoughts?  We can include it with this MR, but I want to make sure this doesn't go stale.

lucbv (approved):
@bathmatt thanks for looking into our questions.
I am fine with splitting things in multiple PRs although we might want to keep track of some of these future changes in an issue otherwise we might forget...

cgcgcg (approved):
Let's merge this. We can polish details in another PR.

PR #9346, opened 2021-06-25
enabled xpetra tests

bathmatt:
@trilinos/xpetra


lucbv:
For those who care to double check, one can that the Xpetra tests are enabled and passing in the `uvm_off` build.
Thanks Matt! I will go ahead and merge this : )

PR #9347, opened 2021-06-25
MueLu/Ifpack: Aggregate Block Smoothing Singleton Handling

csiefer2:
Adding singleton handling which Ifpack (not Ifpack2) didn't have

Issue: n/a (non-ASC)
Stakeholder feedback: n/a (non-ASC)
TESTS: RHEL7


jhux2 (approved):
lgtm

PR #9350, opened 2021-06-25
MueLu: Aggregate smoothing for >1 PDE per node

csiefer2:
Issue: n/a (non-ASC)
Stakeholder Feeback: n/a (non-ASC)
TESTS: New test added!


jhux2:
Why is it necessary to unamalgamate?

csiefer2:
> 
> 
> Why is it necessary to unamalgamate?

Because Ifpack2/Ifpack need a vector of length #dofs which contains the part id for each dof.  Aggregates are on the amalgamated object (# nodes), so we must unamalgamate to get back to dofs.  Make sense?

jhux2:
> > Why is it necessary to unamalgamate?
> 
> Because Ifpack2/Ifpack need a vector of length #dofs which contains the part id for each dof. Aggregates are on the amalgamated object (# nodes), so we must unamalgamate to get back to dofs. Make sense?

I suspected it was something like this.  There might be more elegant transfer mechanisms (e.g., using transpose of the tentative prolongator), but those are going to be more expensive.  Well, maybe not, once @cgcgcg's SpMV short-circuit code merges fully.

jhux2 (approved):
My name is @jhux2, and I approve this PR.

PR #9353, opened 2021-06-28
MueLu: Create cuBLAS and cuSPARSE handles up front

cgcgcg:
@trilinos/muelu 

## Motivation
Create Kokkos-Kernels TPL handles up front, so that they don't mess up solve timings.

### DIFF
--- a/packages/muelu/test/unit_tests/MueLu_Test_ETI.hpp
+++ b/packages/muelu/test/unit_tests/MueLu_Test_ETI.hpp
@@ -63,6 +63,11 @@
 #include <TpetraCore_config.h>
 #endif
 
+#ifdef HAVE_MUELU_KOKKOSKERNELS
+#include <KokkosKernels_config.h>
+#include <KokkosKernels_Controls.hpp>
+#endif
+
 #ifndef MUELU_AUTOMATIC_TEST_ETI_NAME
 #error "The macro MUELU_AUTOMATIC_TEST_ETI_NAME was not defined"
 #endif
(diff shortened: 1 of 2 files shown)


GrahamBenHarper (dismissed):
lgtm

packages/muelu/test/unit_tests/MueLu_Test_ETI.hpp
```diff
@@ -95,6 +100,18 @@ bool Automatic_Test_ETI(int argc, char *argv[]) {
...
+  // up timings.
+#ifdef HAVE_MUELU_KOKKOSKERNELS
+  KokkosKernels::Experimental::Controls controls;
+# ifdef KOKKOSKERNELS_ENABLE_TPL_CUBLAS
+  controls.getCublasHandle();
+# endif
+# ifdef KOKKOSKERNELS_ENABLE_TPL_CUSPARSE
+  controls.getCusparseHandle();
+# endif
+#endif
```
lucbv, line 114: You could consider adding a fence here for good measure, I believe that cuSPARSE is doing some async memory touches in the handle.

lucbv (approved):
This is fine

packages/muelu/test/unit_tests/MueLu_Test_ETI.hpp
cgcgcg: Ok, will do. In my runs it seemed to not matter, but maybe I'm hitting an early fence somewhere else..

jhux2 (approved):
lgtm, thanks!

PR #9368, opened 2021-07-01
MueLu: Disable three region tests for Cuda which have discrepancies

GrahamBenHarper:
@trilinos/muelu 
There are three tests which fail on Cuda builds due to minor discrepancies. 

The first test converges in 10 iterations instead of 11, the second test converges in 40 iterations instead of 42, and the final test has a small deviation in residual history.

## Motivation
Cleaning up the CDash for Cuda builds

## Testing
The tests in question no longer appear on a Cuda build on Geminga


pohm01:
Aren't these tests failing on every experimental muelu build, not just the cuda ones?

GrahamBenHarper:
@pohm01 I could have sworn we had a passing serial experimental build... whoops

PR #9373, opened 2021-07-02
MueLu RefMaxwell: Change default behavior for applying BCs to A_nodal

cgcgcg:
@trilinos/muelu 

## Motivation
Fixes an ALEGRA test, and does not seem to adversely affect any MiniEM or EMPIRE results. 

@skennon10


jhux2 (approved):
Sehr gut.

PR #9376, opened 2021-07-02
MueLu: Update structured region Brick3D gold test to match gold file on Trappist

GrahamBenHarper:
@trilinos/muelu

There's a bit of variation in the region tests across our machines, and it would be nice if at least one machine that submits to CDash doesn't have failures on for experimental tests so that we can tell if any actual bugs are introduced. I think trappist is the most appropriate choice for this.

Now, these tests have large enough differences between machines that making them pass across all machines would require raising the tolerance significantly. For example, on geminga, the residual history is 
```
# num procs = 4
# iteration | res-norm (scaled=1)
#
0	1.0000000000000000e+00
1	1.0743249665883020e-02
2	3.9893894075918067e-04
3	2.3168729086291099e-05
4	1.5098630455396873e-06
5	1.0492899146983019e-07
6	7.6018323570933677e-09
7	5.6574718296441223e-10
8	4.2853133775821826e-11
9	3.2848887501149060e-12
10	2.5391812685604059e-13
```

Let's have a more detailed discussion about fixing the discrepancies in residual histories across machines in an issue instead while we try to turn the CDash more green.


pohm01 (approved):
lgtm.

PR #9384, opened 2021-07-05
Make zoltan2 UVM free

bathmatt:
@trilinos/zoltan2 

This is the ongoing work get zoltan2 more UVM friendly. I'll use this to run the tests more broadly as we get closer, but this has about 3/4 of zoltan2 passing w/o UVM currently


cgcgcg:
@bathmatt The test failure isn't your fault. No idea how this got past the autotester, but #9386 should fix it.

bathmatt:
Think I have that unit test passing, my local build didn't have intrepid2 enabled so that test was empty. 

Might be ready to review.  @kddevin @cgcgcg @csiefer2  can someone look over both the muelu and zoltan changes?

bathmatt:
I don't believe those build errors as they look totally unrelated to what I've done, going to try to double check but thinking of just restarting the PR tester

On the good side, cuda w/o uvm is working for muelu :)

bathmatt:
K I'm freezing development on this branch, it has all of muelu passing and 13 failing tests in zoltan2, I will start to address those once we get this MR done.

bathmatt:
@rppawlo this is the muelu stuff, wonder what shit will hit the fan when we test with panzer as that is probably more coverage than the unit tests

packages/muelu/src/Smoothers/MueLu_Amesos2Smoother_def.hpp
```diff
@@ -202,6 +196,12 @@ namespace MueLu {
       SC normalization = Nullspace->getVector(0)->norm2();
       normalization = Teuchos::ScalarTraits<Scalar>::one()/(normalization*normalization);
 
+      ArrayView<const SC> nullspace, nullspaceImp;
+      nullspaceRCP = Nullspace->getData(0);
+      nullspace = nullspaceRCP();
+      nullspaceImpRCP = NullspaceImp->getData(0);
+      nullspaceImp = nullspaceImpRCP();
+
```
cgcgcg, line 204: What does this change do? Looks like the code was just moved around?

bathmatt: It did just move, however, it no longer has a double cache of the data, it is used on line 196 above for the norm

cgcgcg (approved):
MueLu changes look good

jhux2:
@kddevin Could you please take a look at this PR?

packages/zoltan2/core/src/input/Zoltan2_MatrixAdapter.hpp
```diff
@@ -175,11 +175,11 @@ template <typename User, typename UserCoord=User>
       \param colIds on return will point to the global column Ids for
          the non-zeros for each row.
    */
-  virtual void getCRSView(const offset_t *&offsets, const gno_t *&colIds) const
+  virtual void getCRSView(ArrayRCP<const offset_t> &offsets, ArrayRCP<const gno_t> &colIds) const
```
kddevin, line 178: This change is a non-backward-compatible interface change -- one that will likely be noticed by no one.  I am willing to accept it, but I am curious why it was needed?  Ditto for other methods in this adapter.

packages/zoltan2/core/src/models/Zoltan2_CoordinateModel.hpp
```diff
@@ -356,6 +350,25 @@ void CoordinateModel<Adapter>::sharedConstructor(
...
+  kokkos_xyz_ =   Kokkos::View<scalar_t **, Kokkos::LayoutLeft, typename node_t::device_type>("kokkos_xyz", kokkos_xyz.extent(0), kokkos_xyz.extent(1));
+  Kokkos::deep_copy(kokkos_xyz_, kokkos_xyz);
+
+  if(userNumWeights_ > 0) {
+    Kokkos::View<scalar_t **, typename node_t::device_type> kokkos_weights;
+    ia->getWeightsKokkosView(kokkos_weights);
+    kokkos_weights_ = Kokkos::View<scalar_t **, typename node_t::device_type>("kokkos_weights_",kokkos_weights.extent(0), kokkos_weights.extent(1));
+    Kokkos::deep_copy(kokkos_weights_, kokkos_weights);
+  }
+
```
kddevin, line 371: The deep copies are OK for now.  

When we circle back to Zoltan2, I'd like to try avoiding the copies by not saving the pointers in the class at all.  Instead, we could rewrite getCoordinatesKokkos to call the input adapter functions and return the device views without storing them in the class (e.g., 
```
void getCoordinatesKokkos(Ids, xyz, wgts) {
Ids=inputAdapter->getIdsKokkosView(); 
xyz = inputAdapter->getCoordinatesKokkosView();
if (use_weights) wgts = inputAdapter->getWeightsKokkosView();
}
```
 Method getCoordinatesKokkos() would be called by the Zoltan2 function, which would release the pointers before the application needed them again.

bathmatt: I looked at that in the initial cut, however, some of the versions do a bit of computation and other stuff to get the coords.  Because of this I thought the safest was to cache with a copy.  I really didn't like this for the reasons you mentinoed above.

packages/zoltan2/core/src/input/Zoltan2_MatrixAdapter.hpp
bathmatt: Pointers were going out of scope when I got them to be just in time.  You'd get the pointer and then the thing the pointer pointed at would go out of scope and it would be invalid.

kddevin (approved):

PR #9390, opened 2021-07-06
Tpetra: add test and fix for #9160

### LINKED ISSUE #9160: Tpetra: MultiVector row-wise subview constructor produces wrong num columns
## Bug Report
@trilinos/tpetra 

### Description
The (X, subMap, rowOffset) constructor for MultiVector should always construct an MV with the same number of columns as X, but there is a case where it may end up with more columns than X:
- ``X.isConstantStride() == true``
- X is a column-wise subview of some other MV (``X.origView_.extent(1) != X.view_.extent(1)``)
- rowOffset == 0

For example, Y is a 2-column MV, X is Y.getVector(0), and we're trying to construct Z as a subview of X's rows. The bug causes Z to end up with 2 columns when it should have 1.

brian-kelley:
@trilinos/tpetra 

## Related Issues

* Closes #9160
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback

Originally reported by Sierra although it wasn't a blocker for them. 
## Testing

There were already lots of unit tests for offset view MVs, just none that took offset views of column subviews. So I added a test for that in MultiVector_UnitTests.
## Additional Information

Mark had left this comment in the offset view constructor:
```
    // FIXME (mfh 29 Sep 2016) If we just use X.view_ here, it breaks
    // Ifpack2's Gauss-Seidel implementation (which assumes the
    // ability to create domain Map views of column Map MultiVectors,
    // and then get the original column Map MultiVector out again).
    // If we just use X.origView_ here, it breaks the fix for #46.
    // The test for rowOffset == 0 is a hack that makes both tests
    // pass, but doesn't actually fix the more general issue.  In
    // particular, the right way to fix Gauss-Seidel would be to fix
    // #385; that would make "getting the original column Map
    // MultiVector out again" unnecessary.
```
However, Ifpack2's Gauss-Seidel no longer does what's described here. It keeps an X_colMap that's used for the local apply, and uses a domainMap offset view of that (using the code changed here). This "rowOffset == 0 hack" was really the cause of 9160, so I just used ``X.view_`` in the offset view. ``X.view_.extent(1)`` has the correct num vectors for the offset view if constant stride.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -971,8 +971,8 @@ namespace Tpetra {
     // Check whether the source object is a MultiVector.  If not, then
     // we can't even compare sizes, so it's definitely not OK to
     // Import or Export from it.
-    typedef MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node> this_type;
-    const this_type* src = dynamic_cast<const this_type*> (&sourceObj);
+    typedef MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node> MV;
+    const MV* src = dynamic_cast<const MV*> (&sourceObj);
     if (src == nullptr) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -971,8 +971,8 @@ namespace Tpetra {
     // Check whether the source object is a MultiVector.  If not, then
     // we can't even compare sizes, so it's definitely not OK to
     // Import or Export from it.
-    typedef MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node> this_type;
-    const this_type* src = dynamic_cast<const this_type*> (&sourceObj);
+    typedef MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node> MV;
+    const MV* src = dynamic_cast<const MV*> (&sourceObj);
```
tasmith4, line 975: can we do this more simply with std::is_base_of or std::is_convertible?

tasmith4 (changes requested):

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
brian-kelley: @tasmith4 I don't think so, since those both test for conditions on types at compile-time, and ``sourceObj`` could be a reference to anything in that inheritance tree at runtime.

These two changed lines weren't meant to be semantic changes and have nothing to do with the MV subview issue - I just wanted to get rid of warnings about ``this_type`` shadowing a member typedef.

tasmith4: got it, thanks.  and sorry to have missed seeing your response.

tasmith4 (approved):

kddevin (approved):

PR #9392, opened 2021-07-06
Tpetra wrapped dual view microbenchmark

GeoffDanielson:
@trilinos/tpetra 
## Motivation

Adds a wrappedDualView microbenchmark to the unit test, comparing times using the wrappedDualView.getHostView() to getting a view out of a Kokkos dualview.

## Testing

This is an additional test.


tasmith4 (changes requested):
I'd love to get the microbench tests split out into a separate file before pushing this for real.  I also think the warmup calls could go in the same scope as the timing loop, as long as they still occur before the timer starts.

GeoffDanielson:
> I'd love to get the microbench tests split out into a separate file before pushing this for real. I also think the warmup calls could go in the same scope as the timing loop, as long as they still occur before the timer starts.

Yeah, I can see that being useful.  I think I've addressed it.

csiefer2 (approved):

tasmith4 (approved):

Issue #9200, opened 2021-06-01
Ifpack2: Improving Tests


csiefer2:
We should have better tests.

PR #9405, opened 2021-07-08
Ifpack2: Improving BlockTriDi Unit test

csiefer2:
Issues:#9200
Stakeholder feedback: n/a
Tests: A new one!

Turns out BlockTriDi will woof all over the place if the matrix you hand it isn't at least TriDiagonal.  Tests to check this and issue an error message (rather than segfault) to come in a later PR.


cgcgcg (approved):

lucbv (approved):
Makes sense to create a tridiag graph to test the block-tridiag preconditioner

csiefer2:
> 
> 
> Makes sense to create a tridiag graph to test the block-tridiag preconditioner

Yes.  Yes it does.

PR #9406, opened 2021-07-08
Sacado fix non cuda aware mpi

etphipp:
Fixes possible seg-faults in the Fad Kokkos Comm tests on CUDA when not using CUDA-aware MPI and UVM is disabled.  This is a kludge and should really be addressed at a higher level in Trilinos/Teuchos.


rppawlo (approved):
Thanks @etphipp !

PR #9034, opened 2021-04-20
Galeri: Fix issue in boundary conditions

cgcgcg:
@trilinos/galeri 

## Motivation
Fix boundary conditions for Laplace3D and Brick3D.

PR #9421, opened 2021-07-12
Muelu: Fix structured region unit tests

GrahamBenHarper:
@trilinos/muelu @lucbv 

## Motivation
I confirmed by CDash-informed bisecting that the region unit tests started failing due to commit SHA `cc0db334462fc79a2b1e47e44903ce1f7c49696b`, after #9034 was merged. I believe it changed the values of the hard-coded matrix that we use in region multigrid in the case of the Laplace3D 4 MPI ranks example. I don't have an explanation as to why this does not occur with 1 MPI rank, however.

## Testing

The test `MueLu_UnitTestsRegion_MPI_4` previously failed on my machine, vaporwave, and now it does not.


lucbv (approved):
This looks fine and as long as the unit test is fixed and matches with galeri I think we are good!
Thank for looking into it @GrahamBenHarper

PR #9424, opened 2021-07-13
Panzer: remove UVM use from IntegrationValues

rppawlo:
@trilinos/panzer

## Testing

All IntegrationValues2 tests updated to run without UVM.

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
+++ b/packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
@@ -58,6 +58,9 @@
 #include "Panzer_Traits.hpp"
 #include "Panzer_SubcellConnectivity.hpp"
 
+// For device member functions to avoid CUDA RDC in unit testing
+#include "Panzer_IntegrationValues2_impl.hpp"
+
 // ***********************************************************
 // * Specializations of setupArrays() for different array types
 // ***********************************************************
@@ -372,12 +375,12 @@ convertNormalToRotationMatrix(const Scalar normal[3], Scalar transverse[3], Scal
     binormal[2] = 0.;
   }
 }
-  
+  /*  
 template <typename Scalar>
 void IntegrationValues2<Scalar>::
 swapQuadraturePoints(int cell,
                      int a,
-                     int b)
+                     int b) const
 {
   const int new_cell_point = a;
   const int old_cell_point = b;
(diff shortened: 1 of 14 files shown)


packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
```diff
@@ -846,41 +774,51 @@ generateSurfaceCubatureValues(const PHX::MDField<Scalar,Cell,NODE,Dim>& in_node_
...
+    Kokkos::parallel_for("covarient metric tensor",policy,KOKKOS_LAMBDA (const int cell,const int ip) {
+      norm_contravarient_k(cell,ip) = 0.0;
+      for (size_type i = 0; i < contravarient_k.extent(2); ++i) {
+        for (size_type j = 0; j < contravarient_k.extent(3); ++j) {
+          norm_contravarient_k(cell,ip) += contravarient_k(cell,ip,i,j) * contravarient_k(cell,ip,i,j);
         }
       }
-      norm_contravarient(cell,ip) = std::sqrt(norm_contravarient(cell,ip));
-    }
+      norm_contravarient_k(cell,ip) = std::sqrt(norm_contravarient_k(cell,ip));
```
seamill, line 819: Can you call std::sqrt from within a kokkos functor?

packages/panzer/disc-fe/src/Panzer_IntegrationValues2.hpp
```diff
@@ -149,34 +174,29 @@ namespace panzer {
...
-  protected:
 
+    /// This should be a private method, but using lambdas on cuda forces this to be public.
+    void evaluateRemainingValues(const PHX::MDField<Scalar,Cell,NODE,Dim> & in_node_coordinates, const int in_num_cells);
+
+    /// This should be a private method, but using lambdas on cuda forces this to be public.
+    void getCubatureCV(const PHX::MDField<Scalar,Cell,NODE,Dim> & in_node_coordinates, const int in_num_cells);
+
+    /// This should be a private method, but using lambdas on cuda forces this to be public.
+    void generateSurfaceCubatureValues(const PHX::MDField<Scalar,Cell,NODE,Dim> & in_node_coordinates, const int in_num_cells,const SubcellConnectivity & face_connectivity);
```
seamill, line 187: Should we make these anonymous/non-member calls within the cpp file to prevent them from being public?

packages/panzer/disc-fe/src/Panzer_IntegrationValues2_impl.hpp
```diff
@@ -0,0 +1,145 @@
...
+   standalone device member function, put it in this file (for unit
+   testing).
+ */
+namespace panzer {
+
+template<typename Scalar,
+         typename T0,typename T1,typename T2,typename T3,
+         typename T4,typename T5,typename T6,typename T7>
+void
+swapQuadraturePoints(int cell,
```
seamill, line 18: Do you need to declare this as a kokkos inline function?

packages/panzer/disc-fe/src/Panzer_IntegrationValues2_impl.hpp
```diff
@@ -0,0 +1,145 @@
...
+      hold = surface_rotation_matrices(cell,new_cell_point,dim,dim2);
+      surface_rotation_matrices(cell,new_cell_point,dim,dim2) = surface_rotation_matrices(cell,old_cell_point,dim,dim2);
+      surface_rotation_matrices(cell,old_cell_point,dim,dim2) = hold;
+    }
+  }
+}
+
+template <typename Scalar>
+void IntegrationValues2<Scalar>::
+convertNormalToRotationMatrix(const Scalar normal[3], Scalar transverse[3], Scalar binormal[3])
```
seamill, line 93: It doesn't look like this needs to belong to the IntegrationValues2 class. Should we remove it from the class?

packages/panzer/disc-fe/src/Panzer_SubcellConnectivity.hpp
```diff
@@ -69,15 +69,17 @@ class SubcellConnectivity
    * \return Number of subcells associated with the cells
    */
   KOKKOS_INLINE_FUNCTION
-  int numSubcells() const {return _num_subcells;}
+  int numSubcells() const {return _num_subcells_device(0);}
+  int numSubcellsHost() const {return _num_subcells;}
```
seamill, line 73: I think (double check) you can just return `subcell_to_cells_adj.extent(0)-1`. That would mean we don't need a host version of the call.

packages/panzer/disc-fe/src/Panzer_SubcellConnectivity.hpp
```diff
@@ -69,15 +69,17 @@ class SubcellConnectivity
...
 
   /**
    * \brief Gives number of cells in connectivity
    *
    * \return Number of subcells associated with the cells
    */
   KOKKOS_INLINE_FUNCTION
-  int numCells() const {return _num_cells;}
+  int numCells() const {return _num_cells_device(0);}
+  int numCellsHost() const {return _num_cells;}
```
seamill, line 82: Same here: `_cell_to_subcells_adj.extent(0)-1`

seamill (dismissed):
Looks fine to me. I think my comments are more questions than corrections. Most of the fixme comments are things I can handle during the lazy evaluation refactor.

packages/panzer/disc-fe/src/Panzer_IntegrationValues2_impl.hpp
seamill: It doesn't look like this is the main declaration, so maybe not.

rppawlo: only in the declaration

packages/panzer/disc-fe/src/Panzer_IntegrationValues2_impl.hpp
rppawlo: I can do that.

packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
rppawlo: good catch. This actually works on cuda, but the preferred way that we have been using during atdm is to call transcedentals is without the std namespace. non-namespaced functions have a special version provided by cuda. I talked to Christian this morning and it looks like kokkos is now providing a math function routines for everything in the standard due to SYCL having a different namespace for it's version of the functions. So I switched everything to the Kokkos provided versions as part of this PR.

packages/panzer/disc-fe/src/Panzer_IntegrationValues2_impl.hpp
rppawlo: done.

packages/panzer/disc-fe/src/Panzer_SubcellConnectivity.hpp
rppawlo: done

packages/panzer/disc-fe/src/Panzer_SubcellConnectivity.hpp
rppawlo: done

packages/panzer/disc-fe/src/Panzer_IntegrationValues2.hpp
rppawlo: @seamill - that's a neat idea, but I would lean toward keeping them like this. If we make them stand alone, then we have to pass in 10+ more class members (all kokkos views) as arguments to each of the function calls. That seems like a big chance to accidentally mix up views. I'd prefer to leave them like this and hope that cuda fixes their compiler to allow this in the future. Is that ok?

seamill: Of course, that is fine. I've been running into this issue in other codes, and I'm unsure what the common practice is. As long as they are properly labeled (and they are) I see no issue.

rppawlo:
@seamill - can you reapprove?

seamill (approved):

PR #9415, opened 2021-07-12
Piro: changes related to the option of explicitly transposing the Jacobian

mperego:
- passing directly the Piro parameterList to ThyraROL objects, to simplify code
- enbabling the option of explicitly transposing the matrix when computing the reduced Hessian

@trilinos/piro

## Motivation
Missing support for explicitly transposing the matrix when computing the reduced Hessian.

## Testing
relevant Piro and Albany tests pass

PR #9427, opened 2021-07-14
Piro: enable "Explicit Matrix Transpose" option when computing adjoint senstivies

mperego:
This is related to #9415.
Enables the "Explicit Matrix Transpose" option for computing adjoint senstivies with the old code that does not rely on ROL.
At some point we will get rid of the old code, but for now we want to maintain it functional.

@trilinos/piro


kliegeois (approved):
Looks good to me!
Thanks.

PR #9438, opened 2021-07-15
WIP: Change PR system to use new LoadEnv and GenConfig code

prwolfe:
This removes the old setenvironment code and adds
the segmented, multilevel *ini files. Finally it
adds a file to contain the pullrequest parameters
previously in the single *ini file and which do
not apply in general and adds code to parse that
information.

@trilinos/framework 

## Motivation
The LoadEnv and GenConfig tools will be the default for end users and for the PR system as well. These changes will allow our scripts to extract environment and config information while still using our cmake scripts to report to CDash in the same way we already do.

## Testing
The unit tests for the PR scripts all pass, but some cleanup is still underway.


prwolfe:
This is a starting point for our team programming session later.

Issue #9425, opened 2021-07-13
Ifpack2:  BlockTriDiContainer test fails for Intel 19.1.3 optimized build


jfike:
## Bug Report
@trilinos/ifpack2

### Description
The test Ifpack2_BlockTriDiContainerUnitAndPerfTests_MPI_4 
fails for an optimized build using Intel 19.1.3.

But the test passes for a debug build.

When building with Intel 19.0.3, the test passes for both debug and optimized.

csiefer2:
@kyungjoo-kim

kyungjoo-kim:
@jfike Can I see the output of the test ? As the failure is from changing intel library build, the arithmetic can be slightly changed. In this case, I can just adjust the numeric threshold with more margin.

jfike:
There is a segmentation fault, so I ran it using gdb to get a backtrace.  Here is the output and backtrace.

<I> nranks 1 ni 10 nj 10 nk 10 bs 5 nrhs 1 isplit 1 jsplit 1
Kokkos Version: 3.4.1
Compiler:
KOKKOS_COMPILER_INTEL: 1910
Architecture:
Default Device: N6Kokkos6SerialE
KOKKOS_ENABLE_ISA_KNC: no
KOKKOS_ENABLE_ISA_POWERPCLE: no
KOKKOS_ENABLE_ISA_X86_64: no
Atomics:
KOKKOS_ENABLE_GNU_ATOMICS: no
KOKKOS_ENABLE_INTEL_ATOMICS: no
KOKKOS_ENABLE_WINDOWS_ATOMICS: no
Vectorization:
KOKKOS_ENABLE_PRAGMA_IVDEP: yes
KOKKOS_ENABLE_PRAGMA_LOOPCOUNT: yes
KOKKOS_ENABLE_PRAGMA_SIMD: no
KOKKOS_ENABLE_PRAGMA_UNROLL: yes
KOKKOS_ENABLE_PRAGMA_VECTOR: yes
Memory:
KOKKOS_ENABLE_HBWSPACE: no
KOKKOS_ENABLE_INTEL_MM_ALLOC: no
KOKKOS_ENABLE_POSIX_MEMALIGN: no
Options:
KOKKOS_ENABLE_ASM: yes
KOKKOS_ENABLE_CXX14: yes
KOKKOS_ENABLE_CXX17: no
KOKKOS_ENABLE_CXX20: no
[...]

PR #9439, opened 2021-07-15
Ifpack2 - separate jacobi setup from tridiag setup

kyungjoo-kim:
@trilinos/ifpack2 

## Motivation

#9425 reports that the ifpack block tridiagonal solver fails with Intel 19.1.3 while other compilers are fine. 

I reproduced the problem using a CEE machine. The problem is gone when I add a print statement. GDB cannot pin-point the place where the segmentation fault happens. So, this solution is based on my guess work. I suspected that the ternary operator access a pointer which is not supposed to be accessed by the condition statement. This PR removes all ternary operators and separate the setup into two cases. This resolve the issue reported in #9425.

## Related Issues

#9425 

## Testing
Tested with Intel 19.1.3

### DIFF
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -1231,35 +1231,66 @@ namespace Ifpack2 {
       part2rowidx0(0) = 0;
       part2packrowidx0(0) = 0;
       local_ordinal_type pack_nrows = 0;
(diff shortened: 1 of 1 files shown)


kyungjoo-kim:
@jfike Would you try this PR branch ? In my testing , this resolves the issue.

jfike:
I verified that this fixes the BlockTriDiContainer test and it also fixes the test failures in our application code.

kyungjoo-kim:
Could I ask the review ? It is confirmed that the issue is resolved in the application.

csiefer2 (approved):

PR #9441, opened 2021-07-16
Zoltan2 no UVM testing

bathmatt:
This is to trigger the auto-tester as we get close to being UVM free

The bulk of these commits are part of another PR so don't worry about reviewing yet


bathmatt:
we have one bug with a view layout issue on CPU, but all the zoltan2 tests pass without UVM so progress

bathmatt:
@kddevin this has the code we reviewed yesterday and some other code changes, if you could review we can get zoltan2 protected for UVM free operations.   It looks like the PR tester will go through

lucbv:
@bathmatt I see that you are turning on the testing with no-uvm for Zoltan2.
I guess once this merge I should be able to move ahead and enable HIP support here : )
Let's see what @kddevin thinks, I might be able do have it done before the weekend!

kddevin (approved):
I approve this PR, as it is needed to prevent regression of UVM removal efforts.
I discussed with @bathmatt and @JacobDomagala that these changes defeat the intent of the KokkosAdapter and mj_backwardcompatible adapter.  In both cases, the intent was to provide adapters that return ONLY Kokkos views or ONLY pointer-based views, allowing the Adapter base class to convert between them.  This approach proved infeasible with UVM removal and will be revisited in #9452.  As #9452 is addressed, the DualViews added in this PR can and should be removed.

PR #9443, opened 2021-07-16
PyTrilinos: Get CMAKE_BINARY_DIR directly from CMake

cgcgcg:
@trilinos/pytrilinos 

## Motivation
Ninja only write dummy Makefiles, so reading it in does not work.


kliegeois (approved):
It looks good to me.
The tests passed but none of those tests actually use the modified file.

PR #9444, opened 2021-07-16
WIP Tempus:  Add support for specifying the adjoint through a 2nd ME

etphipp:
Very few Trilinos solvers/preconditioners support solving transposed
linear systems, therefore we often have to resort to explicitly
transposing the matrix or assembling a transposed matrix directly for
adjoint sensitivities.  To better support this case, I extended the
Tempus adjoint integrator implementations to accept an additional
"adjoint" model evaluator for computing the adjoint W/W_op.  This is
necessary because there is currently no adjoint outArg that can be
requested.  I also included an adapter for the case of an implicitly
computed transpose, allowing for a single implementation.  I also
included a prototype Tpetra implementation that explicitly forms the
transpose, but it isn't tested.  Finally I modified the SinCosModel to
provide an explicit adjoint ME, and modified one of the tests to use it.

In a subsequent commit, the Thyra and Tpetra files will be moved to their
proper locations.

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorAdjointSensitivity_decl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorAdjointSensitivity_decl.hpp
@@ -39,6 +39,12 @@ namespace Tempus {
  *
  * Because of limitations on the steppers, the implementation currently assumes
  * df/dxdot is a constant matrix.
+ *
+ * To extract the final solution x(T) and sensitivity dg/dp(T) one should use
+ * the getX() and getDgDp() methods, which return these quantities directly.
+ * One can also extract this data for all times from the solution history,
+ * however the data is stored in Thyra product vectors which requires
+ * knowledge of the internal implementation.
  */
 template<class Scalar>
 class IntegratorAdjointSensitivity :
--- a/packages/tempus/src/Tempus_IntegratorAdjointSensitivity_decl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorAdjointSensitivity_decl.hpp
@@ -76,6 +76,24 @@ public:
    *    <li> "Mass Matrix Is Identity" (default: false) Whether the mass matrix
    *         is the identity matrix, in which some computations can be skipped.
    * </ul>
+   *
+   * To support use-cases with explicitly computed adjoint operators, the
+   * constructor takes an additional model evaluator for computing the adjoint
+   * W/W_op.  It is assumed the operator returned by this model evaluator is
+   * the adjoint, and so will not be transposed.  It is also assumed this
+   * model evaluator accepts the same inArgs as the forward model, however it
+   * only requires supporting the adjoint W/W_op outArgs.
+   */
+  IntegratorAdjointSensitivity(
+    Teuchos::RCP<Teuchos::ParameterList>                pList,
+    const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& model,
+    const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& adjoint_model);
+
+  /*! \brief Version of the constructor taking a single model evaluator. */
+  /*!
+   * This version takes a single model evaluator for the case when the adjoint
+   * is implicitly determined from the forward operator by the (conjugate)
+   * transpose.
    */
   IntegratorAdjointSensitivity(
     Teuchos::RCP<Teuchos::ParameterList>                pList,
@@ -173,6 +191,7 @@ protected:
   Teuchos::RCP<AdjointAuxSensitivityModelEvaluator<Scalar> >
   createAdjointModel(
     const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& model,
+    const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& adjoint_model,
     const Teuchos::RCP<Teuchos::ParameterList>& inputPL);
 
   void buildSolutionHistory(
@@ -180,7 +199,8 @@ protected:
     const Teuchos::RCP<const SolutionHistory<Scalar> >& adjoint_solution_history);
 
   Teuchos::RCP<Thyra::ModelEvaluator<Scalar> > model_;
-  Teuchos::RCP<AdjointAuxSensitivityModelEvaluator<Scalar> > adjoint_model_;
+  Teuchos::RCP<Thyra::ModelEvaluator<Scalar> > adjoint_model_;
+  Teuchos::RCP<AdjointAuxSensitivityModelEvaluator<Scalar> > adjoint_aux_model_;
   Teuchos::RCP<IntegratorBasicOld<Scalar> > state_integrator_;
   Teuchos::RCP<IntegratorBasicOld<Scalar> > adjoint_integrator_;
   Teuchos::RCP<SolutionHistory<Scalar> > solutionHistory_;
@@ -200,6 +220,14 @@ Teuchos::RCP<IntegratorAdjointSensitivity<Scalar> >
 integratorAdjointSensitivity(
   Teuchos::RCP<Teuchos::ParameterList>                pList,
   const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& model);
+
+/// Nonmember constructor
+template<class Scalar>
+Teuchos::RCP<IntegratorAdjointSensitivity<Scalar> >
+integratorAdjointSensitivity(
+  Teuchos::RCP<Teuchos::ParameterList>                pList,
+  const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& model,
+  const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& adjoint_model);
 
 /// Nonmember constructor
 template<class Scalar>
--- a/packages/tempus/src/Thyra_ImplicitAdjointModelEvaluator.hpp
+++ b/packages/tempus/src/Thyra_ImplicitAdjointModelEvaluator.hpp
@@ -1,0 +1,119 @@
+// @HEADER
+// ****************************************************************************
+//                Tempus: Copyright (2017) Sandia Corporation
+//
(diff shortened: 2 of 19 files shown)


etphipp:
The approach here is not the one I originally started with.  I started with trying to create a new ModelEvaluator interface for creating the adjoint W/W_op, but that didn't work because there is no way for the model's evalModel() implementation to know whether the supplied W/W_op should be the forward or adjoint operator.  So unfortunately I think the only two viable approaches for a general adjoint interface are the approach here using a 2nd ModelEvaluator, or extending OutArgs to have an explicit adjoint.  Due to difficulty in extending OutArgs, I chose the former.

etphipp:
And just to be clear, I do intend to move the files to their proper locations in Thyra, once we are satisfied with the design.

ikalash:
> The approach here is not the one I originally started with. I started with trying to create a new ModelEvaluator interface for creating the adjoint W/W_op, but that didn't work because there is no way for the model's evalModel() implementation to know whether the supplied W/W_op should be the forward or adjoint operator. So unfortunately I think the only two viable approaches for a general adjoint interface are the approach here using a 2nd ModelEvaluator, or extending OutArgs to have an explicit adjoint. Due to difficulty in extending OutArgs, I chose the former.

I was going to ask about this - I haven't looked at the PR in detail yet but it appears you didn't take the approach of implementing a ```createW_op_adjoint``` method.  I'll need to look at the code you added and the example.  @mperego do you think your hack where you provide an adjoint W_op based on a flag in Piro instead of creating a second ME would work with this code?

etphipp:
> I was going to ask about this - I haven't looked at the PR in detail yet but it appears you didn't take the approach of implementing a `createW_op_adjoint` method. I'll need to look at the code you added and the example. @mperego do you think your hack where you provide an adjoint W_op based on a flag in Piro instead of creating a second ME would work with this code?

I tried the approach of implementing a create_W_op_adjoint() first, but it doesn't work.  The problem is there is no way for evalModel() to know if the W_op it gets is for the forward or adjoint problem.

etphipp:
The advantage of the 2-ME approach is it is very flexible.  I think it would be straightforward to implement the needed 2nd ME in Piro that uses a flag for Albany just like you are already doing.  There is also a generic Tpetra implementation that transposes the matrix.  My only concern about it is I'm not sure how it would work with the linear solver.

mperego:
> There is also a generic Tpetra implementation that transposes the matrix. My only concern about it is I'm not sure how it would work with the linear solver.

Why are you worried about the linear solver? What does this generic implementation do, does it create a new transposed matrix given a matrix?

mperego (dismissed):
@etphipp It looks good to me. thanks!

etphipp:
> Why are you worried about the linear solver? What does this generic implementation do, does it create a new transposed matrix given a matrix?

Yes, it does exactly that.  I was worried about it primarily because I couldn't recall exactly how the linear solver gets created, and whether it would have the forward or transposed matrix (i.e., via create_W()).  After looking at the Albany implementation, I see it uses get_W_factory() instead of create_W(), and the factory takes the W_op to create the solver.  So it should be fine.

packages/tempus/src/Tempus_IntegratorAdjointSensitivity_decl.hpp
```diff
@@ -76,6 +76,24 @@ class IntegratorAdjointSensitivity :
...
+   * only requires supporting the adjoint W/W_op outArgs.
+   */
+  IntegratorAdjointSensitivity(
+    Teuchos::RCP<Teuchos::ParameterList>                pList,
+    const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& model,
+    const Teuchos::RCP<Thyra::ModelEvaluator<Scalar> >& adjoint_model);
+
+  /*! \brief Version of the constructor taking a single model evaluator. */
+  /*!
+   * This version takes a single model evaluator for the case when the adjoint
```
ikalash, line 94: Is the first part of this comment still accurate?  It looks like the IntegratorAdjointSensitivity now takes 2 MEs, one for the forward model, one for the adjoint?

packages/tempus/src/Thyra_ImplicitAdjointModelEvaluator.hpp
```diff
@@ -0,0 +1,119 @@
+// @HEADER
+// ****************************************************************************
```
ikalash, line 2: Is there a reason these adapters cannot be tested on the SinCos test case you are testing the new capabilities on?  I'm not requiring you to do this, just asking out of my own interest.  If it's possible, I can try it / add a test on the Piro side.

packages/tempus/test/BackwardEuler/Tempus_BackwardEuler_ASA.cpp
```diff
@@ -64,9 +64,12 @@ TEUCHOS_UNIT_TEST(BackwardEuler, SinCos_ASA)
       getParametersFromXmlFile("Tempus_BackwardEuler_SinCos.xml");
```
ikalash, line 64: Should the explicit case be tested too?  I am not requiring to do it in Tempus - I could try it in Piro, since I will be adding tests there anyway.

ikalash (approved):
Sorry for the delay in approving this.  It looks good - thanks for doing this so fast! I asked a few questions as I was going through the changes.  These are more for me to understand things, not necessarily things that need to be changed (esp. with respect to the tests).  I will be adding tests in Piro to ensure I understand the usage of the new capabilities and towards putting in Piro modifications for Albany, so I can add some of the missing testing there.

packages/tempus/src/Tempus_IntegratorAdjointSensitivity_decl.hpp
ikalash: I see there is another constructor that takes a single ME - maybe the comment is intended to go there?

etphipp: @ikalash I don't understand your comment.  The comments for the first constructor, which takes two ME's, describes what the 2nd ME is for.  The 2nd constructor, with a single ME, has the correct comment, I believe.

packages/tempus/src/Thyra_ImplicitAdjointModelEvaluator.hpp
etphipp: The ImplicitAdjointModelEvaluator is tested in almost every adjoint sensitivity test, since they all just supply a single ME (in which case Tempus creates this ImplicitAdjointModelEvaluator for the 2nd one).  I added an adjoint ME for the SinCos model and modified one adjoint sensitivity test to use it explicitly.  So both paths are tested.

packages/tempus/test/BackwardEuler/Tempus_BackwardEuler_ASA.cpp
etphipp: See above.  It is tested in one case using an explicit adjoint ME for the SinCos model.  What I cannot test is the Tpetra adapter, because Tempus has no Tpetra tests.  I think testing it in Piro would be the best place.

etphipp:
To finish this off I will resolve the autotester errors (just a -Werror thing) and move the Thyra and Tpetra files to the correct packages.

packages/tempus/src/Thyra_ImplicitAdjointModelEvaluator.hpp
ikalash: Ah I see.  Makes sense, thanks for clarifying.

packages/tempus/test/BackwardEuler/Tempus_BackwardEuler_ASA.cpp
ikalash: Ok, yes, makes sense.

packages/tempus/src/Tempus_IntegratorAdjointSensitivity_decl.hpp
ikalash: You're right.  I got thrown off b/c github only shows you a snippet of the code.

ikalash (approved):
Looks good, thanks again!

ikalash:
> To finish this off I will resolve the autotester errors (just a -Werror thing) and move the Thyra and Tpetra files to the correct packages.

I thought you were planning to do the move of the files in a later PR after we've prototyped the code in Piro/Albany?  Either way is fine, just asking.

etphipp:
> I thought you were planning to do the move of the files in a later PR after we've prototyped the code in Piro/Albany? Either way is fine, just asking.

I could do that too.  I feel weird committing Tpetra code to Tempus though.  That's the only file that really needs to be moved.  Would moving it to Piro make sense?

etphipp:
I went ahead and move the Tpetra explicit adjoint ME to the Thyra/Tpetra adapters where it belongs.  Assuming it comes back from the autotester clean, we should be good to go.

@ikalash do you want me to merge this for your prototyping in Piro, or do you want to add your Piro commits to this PR?

ikalash:
> I went ahead and move the Tpetra explicit adjoint ME to the Thyra/Tpetra adapters where it belongs. Assuming it comes back from the autotester clean, we should be good to go.
> 
> @ikalash do you want me to merge this for your prototyping in Piro, or do you want to add your Piro commits to this PR?

Thanks for doing this!  No, lets get this into develop.  My branch has a bunch of other changes which are not ready to merge yet w/o which things won't work in develop.  Therefore I prefer to have your changes in develop so that I can merge them into my branch and play with them there.

etphipp:
Don't approve this PR yet.  I still need to fix the W solve issue for the initial conditions for the non-identity mass matrix.

ikalash:
> Don't approve this PR yet. I still need to fix the W solve issue for the initial conditions for the non-identity mass matrix.

Thanks for the heads up.  Please LMK when it's ready for me to look at / approve.

ikalash (approved):
Looks good, thanks!

mperego:
Thanks @etphipp !

PR #9459, opened 2021-07-20
Panzer: Allow for multiple types of mesh rebalancing in ExodusReaderFactory

GrahamBenHarper:
@trilinos/panzer 
@rppawlo 

## Motivation

This follows our discussion in the Drekar meeting today. This is an initial PR to change the framework to allow for more options when it comes to rebalancing. A following PR will add an option to the list to rebalance the mesh according to Percept family trees if the information is available.

## Testing

Testing it on my end now so I won't set automerge, but also the autotester should catch any issues, and I designed the code to make sure it's backwards compatible.


rppawlo (approved):

rppawlo:
@GrahamBenHarper - the test failures come from a missing initialization value. You will need to initialize the value of rebalancing_ in all of the constructors for the Exodus reader class.

GrahamBenHarper:
Thanks @rppawlo! I think I got it now. Re-approve?

rppawlo (approved):

PR #9461, opened 2021-07-20
Tempus: change integrator pl validation depth to 1

### LINKED ISSUE #9446: Tempus:  PL validation
## Bug Report
@trilinos/tempus 

### Description

When it comes time to valid `Time Step Control Strategy` here
https://github.com/trilinos/Trilinos/blob/a0059347574d7f5491d9f26fec7318fe9b544df3/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp#L628

and the application, e.g. Drekar, has supplied an additional sublist definining its own strategy, this validation fails.

Also, why is this validation at the end? It almost serves no purpose since the objects have already been constructors and initialized. Moreover, `integratorPL` and `stepperPL` are local so the validation and defaults that are set are immidiately lost right after the validation is complete within this non-member constructor.

sconde:
Change the PL validation depth to 1 from the default value of 1000. This
is ensure the 'Time Step Control Strategy' sublist is not validated
because Tempus has no way of validating the application sublist TSCS.
The idea is to leave the validation to the each individual TSCS class.

@trilinos/tempus 

## Motivation

When an application code (e.g. Drekar) provides a sublist defining its own TSCS, the Integrator PL validation fails since the additional sublist is not found in the default PL. This PR changes the depth of the PL validation to 1 from the default value of 1000. This will limit the `Time Step Control Strategy` sublist from being validated. 

The sublist validation from the `Time Step Control Strategy` is now left to the TSCS class to validate the PL. This means each application will be responsible validating their own strategies.

## Related Issues

* Closes #9446

### DIFF
--- a/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
+++ b/packages/tempus/src/Tempus_IntegratorBasic_impl.hpp
@@ -625,7 +625,7 @@ Teuchos::RCP<IntegratorBasic<Scalar> > createIntegratorBasic(
   // Validate the Integrator ParameterList
   auto vIntegratorName = validPL->template get<std::string>("Integrator Name");
   auto vIntegratorPL = Teuchos::sublist(validPL, vIntegratorName, true);
-  integratorPL->validateParametersAndSetDefaults(*vIntegratorPL);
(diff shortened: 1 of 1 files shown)


ccober6 (approved):
Looks good!

Issue #9251, opened 2021-06-08
Tpetra:  non-blocking Import/Export


tasmith4:
## Enhancement
@trilinos/tpetra

Provide an API for non-blocking imports and exports on DistObjects.  The new API will have three general types of functions for both imports and exports.  Using import as an example:
* beginImport -- packs the data and initiates the asynchronous send
* endImport -- waits for the asynchronous request to complete and unpacks the data
* isImportReady -- check if the data is ready to unpack

@cgcgcg @vbrunini

tasmith4:
Yes, that was done a while back.  Looks like I tried to autoclose it in #10300, not sure why that didn't work ...

GrahamBenHarper:
@tasmith4 I noticed autocloses no longer work in Trilinos. I'm not sure what changed, but it's been this way for somewhere between months and a year

sebrowne:
I show that this has the DO_NOT_AUTOCLOSE label on it.  Can you link any others that should have been closed but were not?

tasmith4:
@sebrowne I think we thought that DO_NOT_AUTOCLOSE would keep the bot from closing it, but would still allow it to be closed with keywords in the PR.  Is that not true -- i.e. does DO_NOT_AUTOCLOSE also block closing on merge?

If so, is there a way to keep the "one-year bot" from closing it, without blocking closing on merge with keywords?

PR #9463, opened 2021-07-21
Tpetra:  Split Distributor into Plan/Actor

tasmith4:
@trilinos/tpetra 

## Motivation
Current implementation of Distributor doesn't support using the same Distributor for multiple asynchronous operations.  This PR splits the Distributor implementation into a Plan, which does the expensive one-time work to figure out which procs to send to and receive from, and an Actor, which actually performs the communication.

This split enables multiple actors to use the same plan, so that many asynchronous operations (e.g. DistObject::beginImport) can run simultaneously using the same plan without the need to recompute the plan for each.

## Related Issues

* Part of the solution for #9251 

## Testing
This is a refactoring of the implementation details of Distributor -- all Distributor tests still pass.  The new Actor and Plan are considered implementation details of Tpetra, and should not be used directly by other packages.


packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp
```diff
@@ -0,0 +1,796 @@
...
+#include "Teuchos_Comm.hpp"
+#include "Teuchos_RCP.hpp"
+#include "Teuchos_Time.hpp"
+
+namespace Tpetra {
+namespace Details {
+
+template <class View1, class View2>
+constexpr bool areKokkosViews = Kokkos::Impl::is_view<View1>::value && Kokkos::Impl::is_view<View2>::value;
+
```
kddevin, line 56: Need comments documenting the class and its methods.

packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp
```diff
@@ -0,0 +1,796 @@
...
+  static_assert(areKokkosViews<ExpView, ImpView>,
+      "Data arrays for DistributorActor::doPostsAndWaits must be Kokkos::Views");
+  doPosts(plan, exports, numPackets, imports);
+  doWaits(plan);
+}
+
+template <class ExpView, class ImpView>
+void DistributorActor::doPostsAndWaits(const DistributorPlan& plan,
+                                       const ExpView& exports,
+                                       const Teuchos::ArrayView<const size_t>& numExportPacketsPerLID,
```
kddevin, line 129: Eventually (not this PR), should all arguments be Kokkos views, or is mix of Kokkos views and Teuchos::ArrayView OK?

packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp
```diff
@@ -0,0 +1,796 @@
...
+      size_t p = i + procIndex;
+      if (p > (numBlocks - 1)) {
+        p -= numBlocks;
+      }
+
+      if (plan.getProcsTo()[p] != myRank) {
+        exports_view_type tmpSend = subview_offset(
+            exports, plan.getStartsTo()[p]*numPackets, plan.getLengthsTo()[p]*numPackets);
+
+        if (sendType == Details::DISTRIBUTOR_SEND) {
```
kddevin, line 328: Eventually (not this PR), we should decide whether to remove all these different send types.  Are they needed?

packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
```diff
@@ -0,0 +1,975 @@
...
+
+#include "Teuchos_StandardParameterEntryValidators.hpp"
+#include "Tpetra_Util.hpp"
+#include <numeric>
+
+namespace Tpetra {
+namespace Details {
+
+std::string
+DistributorSendTypeEnumToString (EDistributorSendType sendType)
```
kddevin, line 49: Eventually (not this PR), we should decide whether all these send types are really needed.

packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
```diff
@@ -0,0 +1,975 @@
...
+  //
+  // Teuchos::Comm requires that the data for a send are contiguous
+  // in a send buffer.  Therefore, if the data in the send buffer
+  // for doPosts() are not contiguous, they will need to be copied
+  // into a contiguous buffer.  The user has specified this
+  // noncontiguous pattern and we can't do anything about it.
+  // However, if they do not provide an efficient pattern, we will
+  // warn them if one of the following compile-time options has been
+  // set:
+  //   * HAVE_TPETRA_THROW_EFFICIENCY_WARNINGS
```
kddevin, line 156: Here's an untested feature.  I didn't know it existed.
Eventually (not this PR), we should either make the feature easier to enable (e.g., via environment variable) or get rid of it.

packages/tpetra/core/src/Tpetra_Distributor.cpp
```diff
@@ -46,49 +46,14 @@
...
-      }
-    }
-  } // namespace Details
+  // We set default values of Distributor's Boolean parameters here,
+  // in this one place.  That way, if we want to change the default
+  // value of a parameter, we don't have to search the whole file to
+  // ensure a consistent setting.
+  namespace {
+    // Default value of the "Debug" parameter.
+    const bool tpetraDistributorDebugDefault = false;
```
kddevin, line 55: Do we not need default value for barrierBetween_default or useDistinctTags_default any more?  They were set in old code.

Ah, I see below -- they moved to Details; never mind.

packages/tpetra/core/src/Tpetra_Distributor.hpp
```diff
@@ -1142,380 +905,9 @@ namespace Tpetra {
            size_t numPackets,
```
kddevin, line 905: Eventually (not this PR), consider deprecating ArrayRCP interfaces

kddevin (approved):
Thanks for your work and your patience with my review, @tasmith4 
I think the code is fine.  It would be nice to have some documentation for the methods/fields of the new DistributorActor and DistributorPlan classes; the files don't have only limited documentation saying how the classes should be used.  This absence is less critical since they are in Details, but documentation will help future Tpetra people.
I added other comments of things that, eventually, the Tpetra team should address, but they are out of scope for this PR.

Assuming most of the logic was copied from the old Distributor class, I didn't do a careful review of all of the logic.  If you want me to look at something in particular, please let me know.

packages/tpetra/core/src/Tpetra_Distributor.hpp
tasmith4: This is pretty high up my list to do, but it turns out the enable_if's make this not as easy as it seems.

packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
tasmith4: agreed, will add this to my list of things to investigate.

packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp
tasmith4: good point.  perhaps we can discuss at a team meeting in the near future.  I'm not really sure what to think right now.

packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp
tasmith4: agreed.  added to my list for future work

packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp
tasmith4: I think there's some opportunity for refactoring some of the private variables/methods so it's a little more obvious what they do just by reading their names.  I plan to do that and then add additional doxygen documentation as needed.

tasmith4:
@kddevin thanks for the review.  There's only one logic change, and it doesn't affect current usage.  Distributor::createFromRecvs was templated on an ordinal type, because in addition to setting up the proc ids to communicate, it would also do communication to return the GIDs (type GlobalOrdinal) associated with the sending proc ids.  This information was not used by Distributor, but was apparently provided as a convenience to calling routines in Tpetra.

It's my contention that this is assigning too many responsibilities to createFromRecvs, and that GID communication should be done separately in the class that needs it, so I left the GID communication in Distributor::createFromRecvs, but removed it from DistributorPlan::createFromRecvs.  Because I left that logic in Distributor, it won't affect current usage of Distributor that depends on this feature (as is confirmed by the passing unit tests).  We can talk about removing it from Distributor at a future date if that seems appropriate.

This is also consistent with our previous team discussions where we agreed that a TransferPlan object would be responsible for GID-level plans, and that the DistributorPlan is only responsible for the proc ID level plan.

PR #9468, opened 2021-07-22
EPU: Fix nodal variable output after memory optimization

gsjaardema:
@trilinos/seacas


rppawlo (approved):

rppawlo:
@gsjaardema - Thanks for this! The EMPIRE and Charon dashboards have a ton of exodiff failures this morning. Simulations run to completion but exodus files have bad values (-1, 0 or +1) for many of the variables. Verified that this PR fixes the empire test suite.

PR #9474, opened 2021-07-22
Panzer: check overlapping sidesets

rppawlo:
## Motivation

Application needed to prevent overlapping sidesets in a certain problem. This adds a check using panzer::STK_Interface.

## Stakeholder Feedback

Works for EMPIRE!

## Testing

new test added in panzer/adapters-stk/test/sideset_overlap


etphipp (approved):

alanw0 (commented):
Hi Roger, I just happened to notice this pull-request.

I'm curious that you're checking overlap of nodes of a sideset rather than the sides (faces in 3D, edges in 2D) of a sideset. Two sidesets could have distinct sets of sides, but still share connected nodes.

Secondly, I don't know if this check is ever likely to be a performance hotspot, but the overlap check could be done using bucket-ids instead of entity-ids, and would be up to 500 times faster depending on how many entities there are per bucket. (Default is 512). I would be happy to give you some rough code for that if you think performance might be an issue.

rppawlo:
@alanw0 - thanks for the comments! This is a special coupling where the algorithm won't work if any nodes in the sidesets are connected. The sidesets are coupling to other codes and must be completely separate.  At this time, the use case is for a 2D mesh and the check is only done once during setup. They don't think performance will be critical. However, for my own knowledge, and in case they scale to 3D in the future, I would really appreciate it if you could send me the faster code using bucket ids.

alanw0:
no problem. It's almost the same as your code, just substituting buckets for entities. It uses the idea that in stk-mesh a bucket is a grouping of entities (an entity is a general term for node, edge, face, element) which have the same rank (node in this case), topology (trivial for nodes), and part-membership. Furthermore, each entity is contained in one and only one bucket. So if the sideset nodes are distinct, then the buckets for each sideset should be distinct also. i.e., if any node were connected to both sidesets, then that node's bucket would be in the set of buckets for both sidesets. Buckets essentially represent all of the non-empty intersections of the parts of the mesh.
Here's the code:

std::vector<unsigned> bktIds_a;
std::vector<unsigned> bktIds_b;

//get the bucket-ids for selector_a:
const stk::mesh::BucketVector& bkts_a = mesh.getBulkData()->get_buckets(mesh.getNodeRank(), selector_a);
bktIds_a.reserve(bkts_a.size());
for(const stk::mesh::Bucket* bkt : bkts_a) {
  bktIds_a.push_back(bkt->bucket_id());
}

//get the bucket-ids for selector_b:
const stk::mesh::BucketVector& bkts_b = mesh.getBulkData()->get_buckets(mesh.getNodeRank(), selector_b);
bktIds_b.reserve(bkts_b.size());
for(const stk::mesh::Bucket* bkt : bkts_b) {
  bktIds_b.push_back(bkt->bucket_id());
}

//then code the same as yours for searching one list against the other:
for(unsigned a : bktIds_a) {
  if (std::binary_search(bktIds_b.begin(), bktIds_b.end(), a)) {
    has_overlap = 1;
    break;
  }
}

final note: the reason you get a vector of buckets for a given entity-rank and selector, is because a bucket has a maximum capacity (512 by default) so if there are more than 512 entities that can be described by the same rank and selector, then they have to be split into multiple buckets. bucket-ids are zero-based and unique within each entity-rank.

PR #9477, opened 2021-07-23
MueLu: add greedy Dirichlet option

pohm01:
Add and option to treat a node as Dirichlet if any of the DOFs on that node are Dirichlet (the default behavior is that nodes are marked as Dirichlet only if all DOFs on the node are Dirichlet).

@trilinos/muelu 

## Motivation
Observed convergence benefits in some Drekar simulations rppawlo/DrekarBase#546. Further investigation could be useful.

### DIFF
--- a/packages/muelu/doc/UsersGuide/masterList.xml
+++ b/packages/muelu/doc/UsersGuide/masterList.xml
@@ -609,6 +609,14 @@
     </parameter>
 
     <parameter>
+      <name>aggregation: greedy Dirichlet</name>
+      <type>bool</type>
+      <default>false</default>
+      <description>Force the node to be Dirichlet if any DOFs on the node are Dirichlet (default is nodes are Dirichlet only if all DOFs on node are Dirichlet).</description>
+      <comment-ML>parameter not existing in ML</comment-ML>
+    </parameter>
+
+    <parameter>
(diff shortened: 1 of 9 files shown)


mayrmt (commented):
@pohm01 I like this new feature! Looks good, but could you please add a test for the new feature?

packages/muelu/doc/UsersGuide/masterList.xml
```diff
@@ -608,6 +608,14 @@
       <comment-ML>parameter not existing in ML</comment-ML>
     </parameter>
 
+    <parameter>
+      <name>aggregation: greedy Dirichlet</name>
+      <type>bool</type>
+      <default>false</default>
+      <description>Force the node to be Dirichlet if any DOFs on the node are Dirichlet (default is nodes are Dirichlet only if all DOFs on node are Dirichlet).</description>
```
jhux2, line 615: Nitpick -- someone might confuse the use of node here with compute node, when what's really meant is aggregate.

jhux2 (changes requested):
@mayrmt is right, could you please add a unit test?

jhux2 (approved):
lgtm, thanks!

cgcgcg (approved):

PR #9478, opened 2021-07-23
Tempus: FSA - update the documentation

### LINKED ISSUE #7358: Tempus: request for better documentation of getX(), getXDot() and getXDotDot() in SolutionState class with sensitivities
I have discovered that the getX(), getXDot() and getXDotDot() routines in the Tempus sensitivity integrator classes work differently than the routines with the same names in the SolutionState class.  For forward sensitivities, if one calls integrator->getX(), this will return the current solution, x (w/o the sensitivities).  However, if one calls solution_state->getX(), this will return a Block MV where the first block contains the solution x, and the remaining blocks contain the forward sensitivities dx/dp.  It would be beneficial to document this difference in behavior in the Tempus doxygen to prevent confusion by users.  

@etphipp 
@trilinos/tempus

sconde:
@trilinos/tempus 

## Related Issues

* Closes #7358


ccober6 (approved):
Looks good!

PR #9479, opened 2021-07-23
Panzer: Add tests to cover mesh rebalancing options

GrahamBenHarper:
@trilinos/panzer 
@rppawlo 

## Motivation
I'm working on Percept-based mesh rebalancing for use with multigrid for Drekar (i.e. next PR will add the "percept" option for the "Rebalancing" parameter), but I'd like to add new tests first so that I can make sure this is the right direction for testing. I'm setting this test to use only 4 MPI ranks because rebalancing really doesn't make sense on 1 MPI rank, and later I will need 4 MPI ranks to make sure it's behaving as I'd like. I'm using Pamgen meshes for the tests because it doesn't make sense to regularly refine an inline mesh, but I cannot open an exodus file in emacs.

## Testing
I've tested this on my machine, vaporwave, and it runs and passes.

### DIFF
--- a/packages/panzer/adapters-stk/test/stk_interface_test/CMakeLists.txt
+++ b/packages/panzer/adapters-stk/test/stk_interface_test/CMakeLists.txt
@@ -170,6 +170,13 @@ IF(${PACKAGE_NAME}_ENABLE_Pamgen)
     )
 
   TRIBITS_ADD_EXECUTABLE_AND_TEST(
+    tSTKRebalancing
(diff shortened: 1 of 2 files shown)


rppawlo (approved):
Looks good to me!

GrahamBenHarper:
@rppawlo Reapprove?

rppawlo (approved):

GrahamBenHarper:
AT stability issues I suppose?
```
g++: fatal error: Killed signal terminated program cc1plus
compilation terminated.
```

PR #9483, opened 2021-07-24
Panzer no uvm tester

bathmatt:
@trilinos/panzer I am putting this up to enable panzer testing so we can track how many tests fail with Panzer as we go through the work to fix UVM.  As we fix UVM issues in develop we will merge in things and then we can push this branch once it is clean.


bathmatt:
@rppawlo you want to review this?  I commented out the tests that were commented out and they should pass...  I can delete those lines or leave them commented out.  My thought is someone did that for a reason and if they need to be switched back off it is easy.  But we should get this PR through the system

rppawlo:
Sure. I was holding off until #9682 went through first, but I guess that won't matter until we try it on hip backend?

rppawlo (approved):

bathmatt:
> Sure. I was holding off until #9682 went through first, but I guess that won't matter until we try it on hip backend?

I forgot all about that guy ;)

PR #9486, opened 2021-07-26
Fix some warnings

cgcgcg:
@trilinos/muelu 

## Motivation
- Fix some warnings in MueLu, Galeri, Xpetra, Thyra.
- Remove `PauseForDebugger` from MueLu. Teuchos already provides `pause-for-debugging`.


lucbv (approved):
in general I think this is fine but you might want to double check that no one is using the functions being removed?

cgcgcg:
@lucbv I removed it from our tests, and grepping doesn't show any other use within Trilinos.

PR #9494, opened 2021-07-27
Fixed include ordering errors, create_mirror_view now works for panze…

bathmatt:
Fixed it so you can create_mirror_views without UVM enabled. 
@trilinos/phalanx


bathmatt:
@rppawlo this should fix up things, can you do a quick review, 
Thanks

rppawlo (approved):
Thanks @bathmatt !

PR #9495, opened 2021-07-27
Have space correct for get orientations

bathmatt:
This fixes a view space so you can compile miniEM without UVM on for cuda, there are a lot of bugs still in the code so it won't run/pass but this is a small step for that end.


bathmatt:
@rppawlo another one to get teh code to compile since orientations don't take device views.

rppawlo (approved):

PR #9501, opened 2021-07-28
Piro: work on the Hessain based implementation of ROL dot product

mperego:
Generalization of  the parameter list for defining Hessian based dot product.
Now the parameter list looks like:

```
  Matrix Based Dot Product:
    Matrix Type: Hessian Of Response
    Matrix Types:
      Hessian Of Response:
        Response Index: 0
        Block Diagonal Solver:
          Block 0:
            Linear Solver Type: Belos
```

At the moment `Matrix Type` can either be `Identity` (usual l2 dot product) or the `Hessian Of Response`.
The user also has to provide the Index of the Reponse used to compute the Hessian dot product

@trilinos/piro

## Motivation
This makes the input file easier to read and allows to future extension to different matrix based dot product.

## Testing

Modified Piro test input file to exercise the new syntax.

### DIFF
--- a/packages/piro/src/Piro_PerformAnalysis.cpp
+++ b/packages/piro/src/Piro_PerformAnalysis.cpp
@@ -493,8 +493,6 @@ Piro::PerformROLAnalysis(
   }
 
   bool useFullSpace = rolParams.get("Full Space",false);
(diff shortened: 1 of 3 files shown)


kliegeois (approved):
Looks good to me, thanks!

PR #8235, opened 2020-10-21
Gtest: Update to version 1.10.0

tasmith4:
@trilinos/framework @trilinos/stk @jhux2 

## Motivation
Update Gtest to 1.10.0 to match the version in Sierra for consistency with STK.  This also adds functionality desired by the ExaWind project.  This PR only updates the fused-source googletest, and does not include googlemock, so may not completely address the use cases @etphipp and @rppawlo raised in #7698 and #4421.

## Related Issues

* Closes #8001 
* Related to #7698, #4421 

## Testing
This package is part of the testing suite, so is not directly tested and no new tests are added.  The Trilinos tests are expected to continue to pass.

Issue #7698, opened 2020-07-21
TrilinosFramework:  Sacado needs an updated GTest


etphipp:
@trilinos/framework 
@trilinos/sacado 

I have been converting a bunch of unit tests in Sacado that previously used CppUnit to GTest.  The main driver for this is you had to have the CppUnit TPL installed which almost no one has, so these tests were rarely run, and certainly were never run as part of the PR tester nor nightly testing.  However the version of GTest included in Trilinos/commonTools/gtest is not sufficient.  In particular, Sacado needs the "type-parameterized test" capability since almost all of these tests are templated on the Sacado AD type.  However the version in Trilinos appears to not support this capability.  It is unclear to me if this is because that capability didn't exist in GTest at the time it was added, or if only a subset of the GTest functionality was included in Trilinos (in the stock GTest on github, this functionality appears in a separate header that Trilinos does not have).

Can the version of GTest in Trilinos be updated to include this functionality?  Or should I just commit my own version of GTest in Sacado?

etphipp:
No response, so I guess I'll just add my own version of GTest in Sacado.

prwolfe:
Please do not duplicate this.

The version in use is under commonTools and I expect if you replace that and run everyones tests the results will get approved. Not sure how this got missed.

etphipp:
OK, I can try moving the version I added in Sacado to commonTools, replacing what is there.  My only concern is this version is not header-only like the one currently there.  Will that cause problems for downstream customers (e.g., Sierra)?  FYI, the relevant PR for this is #7736.

etphipp:
I was unable to get Trilinos' installed CMake files to work correctly with the new version of GTest I added in commonTools.  Basically the issue was Trilinos' CMake targets file wasn't pulling in the targets file from the nested googletest library.  The only way I could see to fix it would be to write a custom CMake targets file for the Trilinos GTest package, which I don't know how to do.  So I ended up just putting the version of GTest I need in Sacado as a test-only library.  If someone else wants to pick this up and can solve the CMake targets issue, I would be happy remove the version in Sacado.

PR #9514, opened 2021-08-02
Make GTest only a TPL

### LINKED ISSUE #4421: Framework: new version of gtest with gmock and gtest_main
@trilinos/framework - is there any chance someone could upgrade the gtest version in Trilinos to the most recent release and make sure to include gmock and gtest_main? Since gtest is installed as part of Trilinos, some external apps are interested in leveraging this.

@jwillenbring @bathmatt

### LINKED ISSUE #8001: Question about gtest
## Question
@trilinos/framework @bartlettroscoe 

The version of gtest in Trilinos is old, and project such as Exawind would benefit from having a newer version.

Why is gtest part of Trilinos instead of being a TPL?
Which packages rely on gtest?

As an FYI, I believe @alanw0 is looking at upgrading gtest.

ibaned:
@trilinos/framework

## Motivation

This pull request solves the problem that Trilinos includes copy-pasted outdated copies of GTest, which it installs (I know you can disable this but it also disables the install of an STK package we need), which then causes conflicts in a larger application build that obtains GTest as a proper package from its actual repository here on GitHub. It also solves a secondary problem that there are multiple copies of GTest in Trilinos because the first copy-pasted version is too out-of-date.

The problem is solved by removing the Package `Gtest` and the TPL `gtest` and replacing them both with a TPL `GTest` (the capitalization matches the exported CMake package that the latest version of GTest installs). All packages that depended on either the old package or old TPL now depend on the new TPL. This was tested with the latest version of GTest and packages in Trilinos seemed to build and execute their tests fine. Note that the optional/required nature of the dependency was not changed, so this new TPL is now a required test TPL for several packages.

Closes #8001
Closes #4421
Related to #7698
Related to #8235

## Stakeholder Feedback

Being solicited here!

## Testing

A build of Trilinos was done with most packages that depend on `Gtest` or `gtest` enabled, and tests executed successfully.

I've also put in a request for SEMS to maintain a GTest package.


jhux2:
@alanw0 @tasmith4

tasmith4:
@ibaned I agree in principle that GTest is more appropriate as a TPL than as a package.  One concern from the STK side, is that we develop all STK code against Sierra's GTest TPL, which is locked in at the older version currently snapshotted into Trilinos.  It's possible that with updates to GTest, a code that attempts to build STK against the latest GTest will fail to build if features have been deprecated and/or removed (especially given GTest's current "live at head" philosophy).  Worse, it's unlikely to be obvious to the user that the solution is in fact to roll their GTest back to v1.10.0 (or whatever the Sierra GTest happens to be at the time), and it's not clear to me how to effectively communicate that workaround (assuming that this rollback would even work for them in the first place -- possibly they're not in control of their TPL installs, or possibly they have code that depends on the same new feature that breaks STK).

I'm open to hearing your thoughts and suggested solutions for this problem, as I do agree it makes much more sense for GTest to be a TPL.

Also, can you share which STK package fails to build when GTest is not available?  Our intent with our build system was that no production code packages should rely on GTest, so that if GTest is missing, the tests are the only thing that won't build.

ibaned:
Thanks for looking at this @tasmith4 ! The package I was referring to is `STKUnit_test_utils`, which does not install itself if the built-in GTest doesn't install. This is by design in the CMake logic for the package, although I don't know why.

As for the version mismatch concern, I built the STK packages that use GTest with the latest version from Google today and didn't run into any issues. I wouldn't make decisions based on a fear of issues happening in the future since it is such a popular package and seems to be very committed to backward compatibility.

ibaned:
Just for my own records, here is a list of packages whose dependency files were changed:
```
Intrepid
MiniTensor
Percept
ROL
Sacado
SEACAS
ShyLU
STK
```

kyungjoo-kim:
@ibaned Do we also need to remove Kokkos internal gtest ? I see that the current PR does not include this but the motivation sounds like to do this as well.   

https://github.com/trilinos/Trilinos/tree/master/packages/kokkos/tpls/gtest/gtest
https://github.com/trilinos/Trilinos/tree/master/packages/kokkos-kernels/tpls/gtest/gtest


> This pull request solves the problem that Trilinos includes copy-pasted outdated copies of GTest

ibaned:
@lxmota for MiniTensor
@dridzal for ROL
@etphipp for Sacado
@gsjaardema for SEACAS

ibaned:
@kyungjoo-kim in principle I think this would be good to do, however it increases the political difficulty of this change (a lot), and my personal issue will be resolved without removing them.

rppawlo:
@ibaned - thanks for doing this! We've needed gtest sorted out for a while. I have a question. If gtest becomes a pure TPL, could we be at risk of losing testing without knowing? Right now if I configure with Trilinos_ENABLE_TESTS=ON, those packages will automatically enable testing. If gtest is a TPL only and the TPL is not installed/enabled on a platform, will all testing for these packages be silently disabled due to missing TPL dependencies? If the configure fails saying that it can't enable testing, that's fine. But if we expect all testing enabled yet silently lose testing for packages relying on gtest, that could be very bad, giving us false positives. Can someone verify the behavior?

Also, @bartlettroscoe is currently doing a refactor of TriBITS to treat packages and TPLs using the same infrastructure. This means in the future we can redirect any trilinos package in the dependency chain to an installed TPL version. In this case, we could leave a copy of gtest internally as long as we can always redirect to a TPL. Not sure how far out this capability is.

crtrott:
The Kokkos gtest never gets installed, so you shouldn't be able to see it externally.

ibaned:
@rppawlo I just confirmed that if you don't have GTest then CMake configuration will fail. This is because GTest is a *required* test dependency for these packages, so TriBITS will fail to configure if tests are enabled for the package and GTest is not present.

As for the future TriBITS TPL/Package equivalence, I don't think it's as applicable once you realize that GTest should never appear in Trilinos in the form of a package, so since it only ever needs to be a TPL it won't benefit from being able to be a TPL or a package.

rppawlo:
> As for the future TriBITS TPL/Package equivalence, I don't think it's as applicable once you realize that GTest should never appear in Trilinos in the form of a package, so since it only ever needs to be a TPL it won't benefit from being able to be a TPL or a package.

The benefit is one less package to download and install to build trilinos, along with avoiding having to figure out a correct version. Not a big deal. I'm fine with removing. Just need to make sure sems/atdm have a gtest module available on all our testing platforms (PR, nightly and ATDM machines).

NOTE: ROL, STK, Percept and SEACAS are all snapshotted into trilinos, so these changes could be wiped unless they are backported to the external repos as well.

tasmith4:
@ibaned thanks, since you're pretty confident GTest is committed to backwards compatibility I think it makes sense to go ahead with this change and not worry about compatibility issues unless they arise.

Since `STKUnit_test_utils` is a library supporting unit tests, it shouldn't be needed unless you're trying to build and run the STK unit tests.  However we are aware of a dependency problem where stk_search seems to require stk_unit_test_utils, and are working on a fix for that in our current open snapshot PR.  Not sure whether or not this is what you're hitting, but it should be resolved soon regardless.

ibaned:
@tasmith4 sounds good! the reason we're hitting the STK utils issue is that we're building another piece of Sierra (NGS (Morph)) against a build of Trilinos and it wants to use those STK utilities for its own unit tests.

gsjaardema:
@ibaned SEACAS changes look good.  I've made those changes in my SEACAS/github repository also, so they will not get overwritten on the next seacas snapshot.  Thanks for doing this.

ibaned:
Thank you @gsjaardema !

ibaned:
@wadeburgess confirmed that SEMS will ship GTest as a TPL module with their `V1.3` modules slated for mid-August.

rppawlo:
@glhenni - removing gtest from trilinos means the CDE environment will need to add support for gtest.
@nmhamster - since sems is not mounted on the atdm testbeds, looks like we will need gtest built on all these platforms as well. A quick look at vortex didn't turn up the library in the modules.

glhenni:
> @glhenni - removing gtest from trilinos means the CDE environment will need to add support for gtest.
> @nmhamster - since sems is not mounted on the atdm testbeds, looks like we will need gtest built on all these platforms as well. A quick look at vortex didn't turn up the library in the modules.

We don't use gtest in charon but gemma might. I'll have to look into it. Is it required by STK and that's the issue?

ibaned:
@rppawlo @nmhamster the SEMS request process has room to request testbed installs, so if you can provide a list of relevant testbeds I can comment on them in the request.

@glhenni GTest is not required to just install and use Trilinos, but it will be required to build Trilinos tests and run its tests to confirm it is working correctly. This is typically not something that applications like Gemma need to do, but if you do then this change is relevant.

nmhamster:
@ibaned - please provide the list of testbeds and what versions you want installed.

rppawlo:
It might be faster to list the machines we don't need it on :)  Start with all the testbeds that the trilinos atdm builds cover. That would be all of the builds in the Clean, Nightly, ATDM, PrimaryATDM, SecondaryATDM, and SparcATDM tracks listed on the trilinos cdash site:

https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project

Also, probably need the new machines supporting HIP (and SYCL?).

This PR will need to be updated by framework team to switch the builds scripts to load the new modules and add the TPL paths. @trilinos/framework 

 @jwillenbring - what other machines do we have testing on?

ibaned:
Some testbeds to include:
- [ ] vortex
- [ ] ride
- [ ] mutrino
- [ ] stria

bartlettroscoe:
I think this will break SPARC.  Better check with @sebrowne.

bartlettroscoe:
FYI: I don't have access to most of the ATDM systems anymore.  The maintenance of all of this has been handed over to the @trilinos/framework team.

bartlettroscoe:
You can stop the install of gtest in Trilinos by setting:

   -DGtest_SKIP_INSTALL:BOOL=TRUE

If you look in the CMakeCache.txt file, you will see that var defined there.

Perhaps we should just make that default to TRUE instead of FALSE?  (NOTE, the default is FALSE).  Note that you will break SPARC if you pull GTest out of Trilinos.  (If someone can help them to use their own copy of gtest, then this problem goes away.)

As far as I know, there is no reason to remove gtest from Trilinos just because you don’t want to install it.  It is harmless to have a copy of gtest in Trilinos since it is only used to test Trilinos packages.  It never gets used by any libraries in Trilinos that get installed (as far as I know).  That, for example, is how STK packages are enabled, tested, and installed in Trilinos but EMPIRE can still use them even though EMPIRE has its own version of GTest.

bartlettroscoe:
Okay, going over all of this PR I see the problem is with the subpackage `STKUnit_test_utils` which has libs that can be installed which do depend on the internal `Gtest` package.  That is unfortunate.

> Also, @bartlettroscoe is currently doing a refactor of TriBITS to treat packages and TPLs using the same infrastructure. This means in the future we can redirect any trilinos package in the dependency chain to an installed TPL version. In this case, we could leave a copy of gtest internally as long as we can always redirect to a TPL. Not sure how far out this capability is.

I am getting closer to this (you can track this in https://github.com/TriBITSPub/TriBITS/issues/367).  My idea for TPLs like this that you always need with Trilinos is to provide a separate optional git repo (called something like TrilinosCoreTPLs) that you can clone under Trilinos and then choose to build these (or any subset of these) as TriBITS packages along with Trilinos.  Otherwise, you need to point to existing installs of all of these as already installed external packages (i.e. TPLs).  That gives you the best of both worlds without having to resort to the full complexity of something like Spack.

ibaned:
This PR now has the fallback-to-the-package approach implemented. One important difference from before is that the Sacado copy of gtest is still gone, so Sacado needs to enable the TPL in order to build its tests. I did confirm that Sacado's tests build and run when the TPL is enabled.

packages/stk/stk_ngp_test/cmake/Dependencies.cmake
```diff
@@ -1,8 +1,16 @@
-SET(LIB_REQUIRED_DEP_PACKAGES Gtest Kokkos KokkosCore KokkosContainers)
+IF(TPL_ENABLE_GTest)
+  set(MAYBE_GTEST_TPL GTest)
+  set(MAYBE_GTEST_PACKAGE)
+ELSE()
+  set(MAYBE_GTEST_TPL)
+  set(MAYBE_GTEST_PACKAGE Gtest)
+ENDIF()
+
+SET(LIB_REQUIRED_DEP_PACKAGES Kokkos KokkosCore KokkosContainers)
```
tasmith4, line 9: Not sure if it's appropriate to take GTest out of the LIB_REQUIRED_DEP_PACKAGES.  Although stk_ngp_test is only used for testing, I'd be surprised if the library itself built without Gtest.

ibaned: Good catch @tasmith4! Not intentional, definitely a mistake on my part

tasmith4: Thanks for fixing @ibaned!

sebrowne:
Closing due to no traffic or immediate planned work.  Discussion is available in #8001 (and tracking internally via https://sems-atlassian-son.sandia.gov/jira/browse/TRILINOS-266), and this can always be re-opened.

Keeping open PRs low improves the performance of the AutoTester.

PR #9515, opened 2021-08-02
fixed node normals

bathmatt:
made node normals agree with and without UVM,


rppawlo (approved):

bathmatt:
WIth UVM on I'm seeing weird failures in both develop and my branch with
diffs, not sure why,  this is to get the PR tester to run mostly, If it
passes I'll assume we can merge and there is something wacked in my build
like stale files or running in debug mode which I'm doing.  It looks like a
numbering difference.

bathmatt:
@rppawlo what are your thoughts on this.

I get a test failure both with develop and this branch with UVM on.  It diffs in an odd way.  

However, it has the exact same diff in my branch with UVM off so I feel my changes are good.  

Should I merge this and then try to track down the diffs.  THe logic in the test is "we ran this and believe the results so we will hard code those results into the test" with stuff like 

if element == 14 and node == 4 then result should be 1.  The problem is there is no element 14 on the sideset, element 12 has the right values...

bathmatt:
BTW the tests will pass, cuda+UVM just passed

PR #9517, opened 2021-08-03
Teuchos: Fix TimeMonitor in MPI builds if MPI not initialized. 

jennloe:
@trilinos/teuchos 

## Motivation
This fix is similar to PR #7803.  For builds with MPI enabled but MPI has not been initialized.  Allows user to call Teuchos::TimeMonitor functions without crashing.  The previous version was crashing at the call to `getComm` before reaching the `#ifdef HAVE_MPI` logic.

### DIFF
--- a/packages/teuchos/comm/src/Teuchos_TimeMonitor.cpp
+++ b/packages/teuchos/comm/src/Teuchos_TimeMonitor.cpp
@@ -878,7 +878,7 @@ namespace Teuchos {
       // this should be a "serial" (no MPI, one "process")
       // communicator.
 
-      Teuchos::RCP<const Teuchos::Comm<OrdinalType> > comm;
+      Teuchos::RCP<const Teuchos::Comm<int> > comm;
 
   #ifdef HAVE_MPI
       int mpiStarted = 0;
--- a/packages/teuchos/comm/src/Teuchos_TimeMonitor.cpp
+++ b/packages/teuchos/comm/src/Teuchos_TimeMonitor.cpp
@@ -42,6 +42,7 @@
 #include "Teuchos_TimeMonitor.hpp"
 #include "Teuchos_CommHelpers.hpp"
 #include "Teuchos_DefaultComm.hpp"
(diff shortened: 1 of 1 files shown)


cgcgcg:
I am not sure about this. Now, `getDefaultComm` returns something that is not the same as `DefaultComm<int>::getComm`, so that's not default.
I think I'd rather have a check in `DefaultComm<int>::getComm` that `MPI_WORLD` is initialized, and otherwise throw. If the user doesn't want MPI, then disable it. But maybe I don't understand the use case here.

jennloe:
Hmmm.... The use case is that I am running Belos with Kokkos backend, and I want to run Belos with Tpetra backend (plus MPI) from the same build.  And the Belos-Kokkos code was crashing because MPI hadn't been initialized when "summarize" was called.

jennloe:
From what I can tell, this `getDefaultComm` is only called when the Timer needs to write output immediately.  It doesn't appear to be called when the timer is created, and the timer doesn't seem to keep a copy of the communicator.

cgcgcg:
Why not add an MPI_Init to your driver when it is built against Trilinos with MPI?

bartlettroscoe:
Every application that uses MPI should initialize MPI.

jhux2:
@jennloe wrote:
> Hmmm.... The use case is that I am running Belos with Kokkos backend, and I want to run Belos with Tpetra backend (plus MPI) from the same build. And the Belos-Kokkos code was crashing because MPI hadn't been initialized when "summarize" was called.

@jennloe I tend to agree with @cgcgcg and @bartlettroscoe.  If your driver is built using MPI, then it should initialize it first thing.

jennloe:
Okay, that's reasonable.  I'll go through and add some ifdefs to initialize MPI.  It just feels a little odd because there is nothing at all in my driver using MPI.  But I guess it's also odd to build code with MPI when it isn't needed. :P  It's just so I can make sure Tpetra and Kokkos are running the same functions underneath the hood.

PR #9524, opened 2021-08-05
Intrepid2: eliminate extra zero view allocations in Data construction

### LINKED ISSUE #9523: Intrepid2: Eliminate Extraneous "zeroView_" allocations during Data construction
## Enhancement
@trilinos/intrepid2

The cost of Data construction is dominated in most cases by an allocation of a DynRankView with a single zero entry in it.  This is used by Data to allow returns of references to zero in a way that is consistent with returns of references to data that is explicitly stored by the Data object.  There is, however, no reason that each Data object must have its own copy of the zero View.

The cost of this can be significant; after converting Camellia to use the new-style Intrepid2 Basis allocation mechanisms, as well as Data in other places, profiling suggests that approximately 20% of the runtime of the tests is spent in this particular allocation.

I have a draft implementation that constructs the appropriate DynRankView statically; on Data construction, this sets the zeroView_ member to the statically allocated DynRankView.  I have not yet tested this at all (waiting for it to build on my laptop at the moment).  It will be important to confirm that this works under CUDA; I'm a little vague on CUDA's rules for static allocations -- I think what I've done will work, but can't be entirely sure without testing it.

CamelliaDPG:
@trilinos/intrepid2

## Motivation
See Issue #9523.

## Related Issues
Closes #9523.

## Testing
Various Intrepid2 unit tests exercise Data construction.

### DIFF
--- a/packages/intrepid2/src/Shared/Intrepid2_Data.hpp
+++ b/packages/intrepid2/src/Shared/Intrepid2_Data.hpp
@@ -166,9 +166,14 @@ public:
   static ScalarView<DataScalar,DeviceType> zeroView()
   {
     static ScalarView<DataScalar,DeviceType> zeroView = ScalarView<DataScalar,DeviceType>("zero",1);
-    Kokkos::push_finalize_hook( [=] {
-      zeroView = ScalarView<DataScalar,DeviceType>();
-    });
+    static bool havePushedFinalizeHook = false;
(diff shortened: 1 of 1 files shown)


mperego (approved):
Looks good to me!

CamelliaDPG:
Thanks, @mperego!

PR #9530, opened 2021-08-06
MueLu: Adding complete hierarchy Factory graph dump capability

csiefer2:
If you set:
```
    <Parameter name="debug: graph level" type="int" value="-1"/>
```

you now get *all* the levels.

Tests: n/a (debugging capability)

Issue: n/a (non-ASC)

Stakeholder feedback: n/a (developer-only feature).


cgcgcg (commented):
Do we need to update the description of the master list parameter?

csiefer2:
Done

cgcgcg (approved):

Issue #9534, opened 2021-08-07
MueLu: Factory Graph Output Upgrades


key6oardWarrior:
## Enhancement
@trilinos/packages/muelu/utils/FacotryGraph

We need a python tool to combine files into a nested dot file. Currently there is no way of accomplishing this in python, so this tool will allow us to not have to do task by hand. For implementation the code will follow three rules:

1) Some nodes will appear on multiple levels. Those nodes should only appear on the first (lowest number) subgraph in which they exist (and the arrows should route between subgraphs).

2) "NoFactory" is a special case node and exists on every subgraph. Unlike other duplicates, that should just remain on each individual subgraph.

3) Nodes with no edges can be removed.

The lack of this feature makes it harder for anyone who wants to see all factories and how they are connected in one file. So someone will have to do this task by hand which will take a long time and that person may make mistakes.

To call this work complete we must demonstrate that all .dot files are merged into one .dot file based on the three rules listed and that a .png file can also be created for easy viewing.

cgcgcg:
Closed by #9536

PR #9536, opened 2021-08-07
Merge level graphs

key6oardWarrior:
@trilinos/<teamName>

## Motivation
We need a python tool to combine files into a nested dot file. Issue: https://github.com/trilinos/Trilinos/issues/9534

## Testing
N/A


key6oardWarrior:
@cgcgcg @csiefer2 Are susposed to be reviewers, but Github did not let me add them for some reason. There was never an option to add anyone as a reviewer.

cgcgcg (approved):
Kick off the autotester, but don't merge yet

cgcgcg:
@trilinos/muelu

packages/muelu/utils/MergeLevelGraphs/mergeLevelGraphs.py
```diff
@@ -0,0 +1,320 @@
...
+			while os.path.exists(imageFile):
+				ans = input(f"Are you sure you want to overwrite this file: {imageFile}\n Enter \"y\" to pick a different file name: ")
+				if ans.strip().lower() == "y":
+					ans = input("Enter file path: ")
+					imageFile = ans
+				else:
+					break
+		else:
+			graphFile = os.path.join(PATH, "output.dot")
+			imageFile = os.path.join(PATH, "output.png")
```
cgcgcg, line 282: The output is saved to the location of the script. That seems non-standard and leads to unintuitive behavior. Rather use the location from where the script is called.

cgcgcg (changes requested):
The script fails when called on dot files that are in a different folder, since the file name containing a relative path is used for the cluster names. In the output png, the clusters have no labels.

cgcgcg (approved):

Issue #9522, opened 2021-08-04
Belos: issue causing segfault when output requested


skennon10:
This line sets "Output Style" to `verbosity_` instead of `outputStyle_`:
https://github.com/trilinos/Trilinos/blob/53ecba4c5e6b7b962967e5acdb6337d17b5002c7/packages/belos/src/BelosPseudoBlockGmresSolMgr.hpp#L711
For my case using this solver, `verbosity_ = 41`, and  it causes a segfault here since `StatusTestOutputFactory::create` has no "default" clause in the switch statement and returns a null RCP: 
https://github.com/trilinos/Trilinos/blob/53ecba4c5e6b7b962967e5acdb6337d17b5002c7/packages/belos/src/BelosPseudoBlockGmresSolMgr.hpp#L1184

jennloe:
I'll look at this.

jennloe:
Closed by PR #9539.

PR #9539, opened 2021-08-09
Belos: Fix Output Style Crash, Issue #9522

jennloe:
@trilinos/belos 

This PR addresses issue #9522 where the PseudoBlockGmresSolMgr was crashing due to an attempt to change the Output Style.  The Output Factory was getting invalid input.  I have fixed the pseudoblock GMRES solver manager to pass in the correct input.  I have also added a 'default' case to the output factory so that it cannot crash due to an invalid output style. 

I checked the other Belos solver managers, and none of them seem to have the same issue of passing invalid input to the Output Factory.  

Note: An alternative (possibly more 'correct') fix for this issue would have been to change the Output Factory to accept a `Belos::OutputType` instead of `int`.  Since this change would require updating all the solver managers, I did not make that change at this time.


hkthorn (approved):
LGTM.  Thanks!

PR #9545, opened 2021-08-10
Panzer: remove uvm from point values

rppawlo:
@trilinos/<teamName>

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_PointValues2_impl.hpp
+++ b/packages/panzer/disc-fe/src/Panzer_PointValues2_impl.hpp
@@ -53,6 +53,39 @@
 
 namespace panzer {
 
+  
+
+  template<typename Scalar,typename CoordinateArray>
+  struct CopyNodeCoords {
+    using size_type = typename panzer::PointValues2<Scalar>::size_type;
+    const CoordinateArray source_;
+    PHX::MDField<Scalar,Cell,NODE,Dim> target_;
+    CopyNodeCoords(const CoordinateArray in_source,
+                   PHX::MDField<Scalar,Cell,NODE,Dim> in_target)
+      : source_(in_source),target_(in_target) {}
+
+    KOKKOS_INLINE_FUNCTION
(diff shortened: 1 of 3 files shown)


packages/panzer/disc-fe/src/Panzer_PointValues2_impl.hpp
```diff
@@ -53,6 +53,39 @@
...
+    const CoordinateArray source_;
+    PHX::MDField<Scalar,Cell,NODE,Dim> target_;
+    CopyNodeCoords(const CoordinateArray in_source,
+                   PHX::MDField<Scalar,Cell,NODE,Dim> in_target)
+      : source_(in_source),target_(in_target) {}
+
+    KOKKOS_INLINE_FUNCTION
+    void operator() (const size_type cell,const size_type node,const size_type dim) const {
+      target_(cell,node,dim) = source_(cell,node,dim);
+    }
```
bathmatt, line 70: why not use a deep copy?  THere is no transformation is there?

bathmatt:
@rppawlo I'm confused it looks like this is just a copy evaluator, can it just be a deep copy?  

the test changes look fine

packages/panzer/disc-fe/src/Panzer_PointValues2_impl.hpp
rppawlo: Since source_ is a class template parameter, it can be an MDField or a raw kokkos view. To use deep_copy, we would need to call get_view() on the mdfield to use deep_copy. However we can't use get_view() if source is a raw kokkos view. This is similar to the problem we discussed yesterday with Jacob. To do it, we'd have to add a little template metaprogramming function to return the underlying view. This just seem easier and clearer for users. I can go that route if you prefer.

bathmatt (approved):

prwolfe:
@rppawlo - this is failing on a python test due to something I did. I am fixing it now and will add a retest flag in a bit. My apologies for the inconvenience.

rppawlo:
Thanks @prwolfe !

PR #9546, opened 2021-08-10
panzer: fixed gather coords test for UVM removal

bathmatt:
Removed UVM from a couple evaluators and a test.


rppawlo (dismissed):

bathmatt:
@rppawlo you can't reduce a bool so I needed to modify this PR, it needs your approval again, sorry, worked on GPU but not CPU.

rppawlo (approved):
@bathmatt - looks good. If the lambda capture is using an mdfield, can you pull out the kokkos view and use that in the lambda to eliminate the host from device compiler warnings?

PR #9548, opened 2021-08-10
9547 panzer default lazy basis values from workset

### LINKED ISSUE #9547: Panzer:  Make lazy evaluation of BasisValues2 default for Worksets
## Enhancement
@trilinos/panzer

What feature/addition is needed?  How does this differ from what we currently
have?  Are there any alternative solutions you've considered?  Do you have
suggestions on implementation?  How has the lack of this feature affected you? 
What are you trying to accomplish?  What specifically must we demonstrate to
call this work complete?

seamill:
@trilinos/<teamName>

### DIFF
--- a/packages/panzer/adapters-stk/test/panzer_workset_builder/apply_orientations.cpp
+++ b/packages/panzer/adapters-stk/test/panzer_workset_builder/apply_orientations.cpp
@@ -166,7 +166,8 @@ testApplyOrientations(const bool by_container,
   }
 
   // Grab the basis values vector for the given integration points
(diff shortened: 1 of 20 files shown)


seamill:
@rppawlo Could you take a look at this? I wasn't able to get lazy BV2 objects to be default, but this does allow the code to create them. Panzer and Drekar are able to use the lazy-as-default changes, but empire's em stuff breaks. I'll have to make those changes once this branch in.

rppawlo (approved):
@seamill - looks good to me! cc'ing @bathmatt and @JacobDomagala - they are working on UVM removal from worksets builders. I don't think this should conflict but just in case...

PR #9550, opened 2021-08-10
Remove UVM from PanzerAdaptersSTK (partially)

JacobDomagala:
Fixed tests:
- tExodusFaceBlock
- tExodusEdgeBlock
- PanzerAdaptersSTK_initial_condition_builder
- tSTK_IO

### DIFF
--- a/packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_CustomMeshFactory.cpp
+++ b/packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_CustomMeshFactory.cpp
@@ -353,7 +353,7 @@ namespace panzer_stk {
         for (std::size_t k=0;k<dim_1;++k) {
           const auto loc = q*dim_1 + k;
           charge_density_by_local_ids(q,k) = ChargeDensity_[loc];
-          charge_density_by_local_ids(q,k) = ElectricPotential_[loc];
+          electric_potential_by_local_ids(q,k) = ElectricPotential_[loc];
         }
       }
 
--- a/packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_CustomMeshFactory.cpp
+++ b/packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_CustomMeshFactory.cpp
@@ -246,7 +246,7 @@ namespace panzer_stk {
 
     // build the elements
     std::vector<std::string> block_ids;
-    mesh.getElementBlockNames(block_ids);  
+    mesh.getElementBlockNames(block_ids);
 
     for (int i=0;i<NumElementsPerProc_;++i) {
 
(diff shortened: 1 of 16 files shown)


packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_CustomMeshFactory.cpp
```diff
@@ -326,50 +326,47 @@ namespace panzer_stk {
...
-             itr=elements.begin();itr!=elements.end();++itr) {
-        int q = mesh.elementGlobalId(*itr) - OffsetToGlobalElementIDs_;
-        for (int k=0;k<8;++k) {
-          int loc = q*8 + k;
-          charge_density_by_local_ids.push_back(ChargeDensity_[loc]);
-          electric_potential_by_local_ids.push_back(ElectricPotential_[loc]);
+
+      Kokkos::View<double**, Kokkos::HostSpace> charge_density_by_local_ids("charge_density_by_local_ids", n_elements, dim_1);
+      Kokkos::View<double**, Kokkos::HostSpace> electric_potential_by_local_ids("electric_potential_by_local_ids", n_elements, dim_1);
+
```
bathmatt, line 348: @rppawlo do you know if we need a layout here?  I don't think so but not 100%

packages/panzer/adapters-stk/test/stk_interface_test/tSTK_IO.cpp
```diff
@@ -397,31 +397,42 @@ void assignBlock(FieldContainer & block,FieldContainer & vertices, double val)
...
-      }
-   }
+   Kokkos::parallel_for(
+       cellCnt, KOKKOS_LAMBDA(size_t cell) {
+          for (std::size_t node = 0; node < nodeCnt; node++) {
+             block(cell, node) = val;
+          }
+       }
+   );
 }
```
bathmatt, line 410: deep_copy???

bathmatt (approved):
let the tester run, don't merge

alanw0 (commented):
We should have a discussion at some point, possibly. I'm not familiar exactly with what you're doing, but it seems like you might be copying stk field data into kokkos views in order to use the data on GPU device... I wonder if it would be convenient or useful for you to consider stk's ngp-mesh and ngp-field which allow you to manage that kind of thing using an API with modify-on-host/device and sync-to-host/device methods.

packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_CustomMeshFactory.cpp
rppawlo: I don't think so. The special layouts are only needed for FAD scalar types.

bathmatt:
This fails to build on cuda for me, will look into it and maybe push a patch or just get the issue back

packages/panzer/disc-fe/src/Panzer_Workset_Builder_impl.hpp
```diff
@@ -222,55 +222,64 @@ panzer::buildBCWorkset(const WorksetNeeds & needs,
...
-        }
+      Kokkos::parallel_for(
+          vertex_coordinates.extent(1), KOKKOS_LAMBDA(std::size_t vertex)
+          {
+            const auto extent = Teuchos::as<std::size_t>(vertex_coordinates.extent(2));
+
+            for (std::size_t dim = 0; dim < extent; ++dim)
+            {
+              coords_view(cell, vertex, dim) = vertex_coordinates(side.second[cell].first, vertex, dim);
+            }
```
bathmatt, line 265: Line 264 doesn't work for me and I'm amazed it did for you, but I can't compile.  @JacobDomagala what version of nvcc are you using?  

Anyway you can't have std::pair or std::vector so side.second[] has to be replaced with some view structure or the copy needs to happen on host.  I think the latter is better but,

rppawlo:
@alanw0 - yes - I have been planning to talk to you about a transition to STK-NGP. Maybe we can meet in the next week or two? I'll ping you outside of github.

packages/panzer/disc-fe/src/Panzer_Workset_Builder_impl.hpp
```diff
@@ -222,55 +222,64 @@ panzer::buildBCWorkset(const WorksetNeeds & needs,
...
 
-    PHX::View<int*> cell_local_ids_k = PHX::View<int*>("Workset:cell_local_ids",worksets[side->first].cell_local_ids.size());
-    for(std::size_t i=0;i<worksets[side->first].cell_local_ids.size();i++)
-      cell_local_ids_k(i) = worksets[side->first].cell_local_ids[i];
-    worksets[side->first].cell_local_ids_k = cell_local_ids_k;
+    auto cell_local_ids_k = PHX::View<int*>("Workset:cell_local_ids", worksets[side.first].cell_local_ids.size());
+    Kokkos::parallel_for(
+        worksets[side.first].cell_local_ids.size(), KOKKOS_LAMBDA(std::size_t i) {
+          cell_local_ids_k(i) = worksets.at(side.first).cell_local_ids[i];
+        }
```
bathmatt, line 274: Again here.

bathmatt (changes requested):

packages/panzer/disc-fe/src/Panzer_Workset_Builder_impl.hpp
JacobDomagala: I'm using nvcc 11.4

bathmatt (commented):
This is good to go, @rppawlo any comments?

bathmatt (approved):

rppawlo (approved):
Looks good to me! Thanks!

rppawlo:
Just built with these changes. There are a ton of new warnings on cuda for calling a host function from device code. We might need another PR to address. Suspect it is pulling in objects like sideset_info and mesh_info. Might just need to pull the views out of these objects so lambda capture only pulls in device specific objects.

PR #9561, opened 2021-08-12
Phalanx: add some utilities for view of views

rppawlo:
@trilinos/<teamName>

## Motivation

Some utilities to catch frequent mistakes with view-of-views

## Testing

New tests added: phalanx/test/ViewfViews/tPhalanxViewOfViews.cpp


bathmatt (approved):
THis looks great, like the recursive template.

PR #9562, opened 2021-08-13
add InstallSymLink CMake helper

ibaned:
@trilinos/seacas

## Motivation

SEACAS uses this module to create a backwards compatible name for Exodus when installing the static version of the Exodus library. In the SEACAS repository this file exists under `cmake/modules`, this PR adds the file to `cmake/` where it can be found  and used. Without this change, SEACAS fails to compile when static Exodus is enabled.

## Testing

One build with static Exodus was tested and worked.


ibaned:
Is PR testing broken? this PR literally just adds a CMake file not used normally

PR #9567, opened 2021-08-14
Tpetra: Remove Distributor from pack/unpack interfaces (WIP)

tasmith4:
@trilinos/tpetra 

## Motivation
This PR removes the Distributor argument that was passed through to all the pack and unpack routines.  This argument was enforced by both the DistObject and Packable interfaces, but was not used in any Tpetra subclass of either of these interfaces.

This argument was complicating the split of Distributor to Actor/Plan within DistObject::begin/endImport/Export, since the new structure does not contain a Distributor object at any point.  Also, since the argument was unused everywhere, it was forcing unnecessary boilerplate on every class that wanted to inherit from DistObject or Packable.

In general, it seems this argument should not be needed, since other arguments to these routines pass the import/export arrays along with sizing information (so there should be no need to query the Distributor).  A comment in the test file CrsGraph_UnpackIntoStaticGraph.cpp suggests the reason for putting Distributor in the interface is to allow some optimized packing order based on some detailed information about index permutations that Distributor holds.  Since Tpetra never takes advantage of this option, it doesn't seem worth it to have the argument hanging around doing nothing.  It seems better from a design perspective to try to figure out an alternative way of getting the permutation information to the pack routines if this optimization became important in future.

Currently WIP:  Using the autotester to identify Trilinos packages outside Tpetra that might be implementing DistObject or Packable (believe this is unlikely based on cursory repo search).

## Testing
This is a refactor, that continues to pass PR and nightly testing.

### DIFF
--- a/packages/tpetra/core/inout/Tpetra_Details_CooMatrix.hpp
+++ b/packages/tpetra/core/inout/Tpetra_Details_CooMatrix.hpp
@@ -1209,8 +1209,7 @@ protected:
      buffer_device_type>& exports,
    Kokkos::DualView<size_t*,
      buffer_device_type> numPacketsPerLID,
-   size_t& constantNumPackets,
-   ::Tpetra::Distributor& /* distor */)
+   size_t& constantNumPackets)
   {
     using Teuchos::Comm;
     using Teuchos::RCP;
@@ -1406,7 +1405,6 @@ protected:
    Kokkos::DualView<size_t*,
      buffer_device_type> numPacketsPerLID,
    const size_t /* constantNumPackets */,
-   ::Tpetra::Distributor& /* distor */,
    const ::Tpetra::CombineMode /* combineMode */)
   {
     using Teuchos::Comm;
--- a/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_decl.hpp
@@ -740,8 +740,7 @@ protected:
      buffer_device_type>& exports,
    Kokkos::DualView<size_t*,
      buffer_device_type> numPacketsPerLID,
-   size_t& constantNumPackets,
-   Distributor& /* distor */) override;
+   size_t& constantNumPackets) override;
 
   virtual void
   unpackAndCombine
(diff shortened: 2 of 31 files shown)


william76:
@prwolfe @e10harvey @jwillenbring 

Hey Guys, the Python3 build here is broken -- It looks like some of the new GenConfig stuff has made into the regular PR testing pool rather than staying cooped up in Dev->Master.

Here's the important snippet from the console output (see last PR output from the autotester in the comments above):
```
+--------------------------------------------------------------------+
|   E N V I R O N M E N T   S E T   U P   S T A R T
+--------------------------------------------------------------------+
Unhandled option found in section RHEL7_SIERRA-PYTHON-3.6.10 or one of its dependent sections.
The following key(s) does not match any known operation handlers:
    setenv MANPATH
    setenv PATH_TMP
    setenv PATH

Unhandled option found in section rhel7_sems-gnu-7.2.0-python-3.6.10-serial or one of its dependent sections.
The following key(s) does not match any known operation handlers:
    setenv MANPATH
    setenv PATH_TMP
    setenv PATH
Valid operations are:
   evvar-append
   evvar-assert-not-empty
   evvar-find-in-path
```
It looks like the .ini file has a couple of sections that specify `setenv` as **operations** (i.e., `setenv MANPATH` for example in the error message). 

I'm on kiddo duty today so I'm not logged into work so I can't double check this but I believe this is an error...  There is no `setenv` operation defined in `SetEnvironment` -- What you're looking for is `envvar-set` instead. I can't get on to try and fix this but this should be a pretty easy fix I think.

Also... evidence of the "new" validation checks we got into the new framework last week doing their job ;)

FYI @csiefer2

prwolfe:
> It looks like some of the new GenConfig stuff has made into the regular PR testing pool rather than staying cooped up in Dev->Master.

This one is supposed to be on the PR->develop side.

Joshua pushed to the GenConfig and LoadEnv repos about half an hour ago.

william76:
> This one is supposed to be on the PR->develop side.

@prwolfe that .ini file will need to be fixed or the job should be rolled back or no PR's will make it through since the validation check is a file-level thing I think for GenConfig :grin: I don't know what Josh's github name is, can't tag him.

prwolfe:
> > This one is supposed to be on the PR->develop side.
> 
> @prwolfe that .ini file will need to be fixed or the job should be rolled back or no PR's will make it through since the validation check is a file-level thing I think for GenConfig 😁 I don't know what Josh's github name is, can't tag him.

@e10harvey is on this. I agree

e10harvey:
@csiefer2: the python-3 config will not pass until https://github.com/trilinos/Trilinos/pull/9610 merges.

PR #9569, opened 2021-08-16
phalanx: Mdfield cast

bathmatt:
This merge allows one to deep copy from a MDField without having to get the view out

### DIFF
--- a/packages/phalanx/test/Field/Field.cpp
+++ b/packages/phalanx/test/Field/Field.cpp
@@ -353,19 +353,6 @@ TEUCHOS_UNIT_TEST(field, all)
       c_f6.setFieldData(PHX::KokkosViewFactory<double,typename PHX::DevLayout<double>::type,PHX::Device>::buildView(c_f6.fieldTag()));
       c_f7.setFieldData(PHX::KokkosViewFactory<double,typename PHX::DevLayout<double>::type,PHX::Device>::buildView(c_f7.fieldTag()));
     }
-
-    // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    // Kokkos cast DynRankView accessor
-    {
-      auto kva = a.get_view();
-      Kokkos::deep_copy(kva, a.get_view());
-
-      auto kva_s = a.get_static_view();
-      Kokkos::deep_copy(kva_s, a.get_view());
-
-      auto kvc = c.get_view();
-      Kokkos::deep_copy(kvc, c.get_static_view());
-    }
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     // Kokkos static View accessors
     {
--- a/packages/phalanx/test/Field/Field.cpp
+++ b/packages/phalanx/test/Field/Field.cpp
@@ -355,6 +355,18 @@ TEUCHOS_UNIT_TEST(field, all)
     }
 
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // Kokkos cast DynRankView accessor
+    {
+      auto kva = a.get_view();
+      Kokkos::deep_copy(kva, a.get_view());
+
+      auto kva_s = a.get_static_view();
+      Kokkos::deep_copy(kva_s, a.get_view());
+
+      auto kvc = c.get_view();
+      Kokkos::deep_copy(kvc, c.get_static_view());
+    }
+    // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     // Kokkos static View accessors
     {
       // non-const view
--- a/packages/phalanx/test/Field/MDField_Compiletime.cpp
+++ b/packages/phalanx/test/Field/MDField_Compiletime.cpp
@@ -418,6 +418,18 @@ TEUCHOS_UNIT_TEST(mdfield, CompileTimeChecked)
     }
 
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // Kokkos cast DynRankView accessor
+    {
+      auto kva = a.get_view();
+      Kokkos::deep_copy(kva, a.get_view());
+
+      auto kva_s = a.get_static_view();
+      Kokkos::deep_copy(kva_s, a.get_view());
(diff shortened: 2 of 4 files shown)


bathmatt:
@rppawlo why is there two pushes on the MR?  I don't get git sometimes..  
IF I rebase with origin/develop I only see one on my end.  
@JacobDomagala just tagging you

rppawlo:
Not sure. That should not be there. The view of views commit didn't go in until late this morning. I wonder if you rebased before that commit went into trilinos? Maybe check that origin/develop has the view of view commit in there before rebasing again?

bathmatt:
> Not sure. That should not be there. The view of views commit didn't go in until late this morning. I wonder if you rebased before that commit went into trilinos? Maybe check that origin/develop has the view of view commit in there before rebasing again?

Here is what I see if I do a git diff
```
csuadmins-MBP:Trilinos LOFTUSER$ git diff origin/develop
diff --git a/packages/phalanx/src/Phalanx_MDField.hpp b/packages/phalanx/src/Phalanx_MDField.hpp
index 5d0fb1a0e8d..2af0f2a8201 100644
--- a/packages/phalanx/src/Phalanx_MDField.hpp
+++ b/packages/phalanx/src/Phalanx_MDField.hpp
@@ -459,6 +459,9 @@ namespace PHX {
         dims[i] = static_cast<iType>(m_view.extent(i));  // dangerous
     }
 
+    KOKKOS_FORCEINLINE_FUNCTION
+    operator Kokkos::DynRankView<Scalar,typename PHX::DevLayout<Scalar>::type,PHX::Device> () const { return get_view();}
+    
     KOKKOS_FORCEINLINE_FUNCTION
     Kokkos::DynRankView<Scalar,typename PHX::DevLayout<Scalar>::type,PHX::Device> get_view()
     {return m_view;}
diff --git a/packages/phalanx/test/Field/Field.cpp b/packages/phalanx/test/Field/Field.cpp
index 6b4633d7748..74d5fd34566 100644
--- a/packages/phalanx/test/Field/Field.cpp
+++ b/packages/phalanx/test/Field/Field.cpp
@@ -354,6 +354,18 @@ TEUCHOS_UNIT_TEST(field, all)
       c_f7.setFieldData(PHX::KokkosViewFactory<double,typename PHX::DevLayout<double>::type,PHX::Device>::buildView(c_f7.fieldTag()));
     }
 
+    // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    // Kokkos cast DynRankView accessor
+    {
+      auto kva = a.get_view();
+      Kokkos::deep_copy(kva, a.get_view());
+
+      auto kva_s = a.get_static_view();
+      Kokkos::deep_copy(kva_s, a.get_view());
+
+      auto kvc = c.get_view();
+      Kokkos::deep_copy(kvc, c.get_static_view());
+    }
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     // Kokkos static View accessors
     {

```

Only my changes?  as much as I like git, i hate it much more sometimes :)

packages/phalanx/test/Field/Field.cpp
```diff
@@ -354,6 +354,18 @@ TEUCHOS_UNIT_TEST(field, all)
...
+      auto kva = a.get_view();
+      Kokkos::deep_copy(kva, a.get_view());
+
+      auto kva_s = a.get_static_view();
+      Kokkos::deep_copy(kva_s, a.get_view());
+
+      auto kvc = c.get_view();
+      Kokkos::deep_copy(kvc, c.get_static_view());
+    }
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
rppawlo, line 369: @bathmatt - Looks like this test code was added to the wrong test. There are two versions of field objects: "PHX::MDField" implemented in Phalanx_MDField.hpp and a "PHX::Field" implemented in Phalanx_Field.hpp. For MDField, you would have to add the test code to two test files MDField_Runtime.cpp and MDField_Compiletime.cpp. The runtime and compile time versions mirror each other but use different code paths underneath depending on whether the underlying kokkos view has a static or dynamic rank. The Field.cpp file is for testing the PHX::Field object. In case you are wondering, the MDField was the original object and allows for the extent information from structs (MDField<double,CELL,IP,DIM>), while the Field object takes an integer for the rank (Field<double,3>). Fields was requested by the empire fluids team.

bathmatt: OK, Ill add it there and remove my other test.  Sorry, not up to speed on the testing here.

bathmatt:
and like that, the bad commit is gone,

packages/phalanx/test/Field/MDField_Compiletime.cpp
```diff
@@ -417,6 +417,18 @@ TEUCHOS_UNIT_TEST(mdfield, CompileTimeChecked)
...
+      auto kva = a.get_view();
+      Kokkos::deep_copy(kva, a.get_view());
+
+      auto kva_s = a.get_static_view();
+      Kokkos::deep_copy(kva_s, a.get_view());
+
+      auto kvc = c.get_view();
+      Kokkos::deep_copy(kvc, c.get_static_view());
+    }
     // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
rppawlo, line 432: A silly question - the tests are doing a deep_copy to itself. Are we sure that kokkos doesn't look at the pointer and skip the operations knowing they are the same memory? Just wondering if these tests should use a host mirror to make sure a copy really takes place?

bathmatt: That is correct, before that code would have compile time errored out.  This is saying that the compiler can now do the cast and call the deep copy function.  What happens inside is not part of the test.  I can do a deep copy with a const and check that const,

bathmatt: Crap, it isn't doing what I want it to, let me dig more

bathmatt: probably too hopped up on cold meds to be doing this :). I will look into what I was thinking but it wasn't testing any of the new code...

bathmatt:
@rppawlo ok, my nyquill fulled code spree is finally over, i think this is ready to review

rppawlo (approved):
Looks good @bathmatt ! Thanks!

PR #9585, opened 2021-08-18
SEACAS: Aprepro - do not diff stderr in tests

gsjaardema:
Replaced the last test that was relying on a robust diff of stderr to no longer do that.  Many HPC platforms were polluting the stderr output and that resulted in non-robust testing.  The test no longer relies on a diff of stderr and instead just checks the contents of files that apepro itself writes directly.

@trilinos/seacas


prwolfe (approved):
Nice

PR #9586, opened 2021-08-18
MueLu: Don't require Epetra for ML defaults

cgcgcg:
@trilinos/muelu 

## Motivation
Using the defaults from ML shouldn't require Epetra.


cgcgcg:
@skennon10

lucbv (approved):
I think that's fine, ML does not depend on `EPETRA` and MueLu might well want to use the ML default in a Tpetra only build

PR #9594, opened 2021-08-19
MueLu: Bug out when mode!=NO_TRANS

### LINKED ISSUE #9590: MueLu:  Trans mode is not supported but no error is thrown when requested
## Bug Report
@trilinos/muelu 

### Description
MueLu, when used as preconditioner, does not support the TRANS mode. However, no error is thrown when TRANS mode is requested, which typically leads to very poor convergence since the preconditioner is built for the original matrix rather than the transposed one.

An error should be thrown when the TRANS mode is requested.
In case it's useful, here is how other solvers handle this:
https://github.com/trilinos/Trilinos/blob/master/packages/ifpack2/src/Ifpack2_Details_Amesos2Wrapper_def.hpp#L452
https://github.com/trilinos/Trilinos/blob/master/packages/shylu/shylu_dd/frosch/src/SolverInterfaces/FROSch_Amesos2SolverTpetra_def.hpp#L102

### Steps to Reproduce
I don't have a unit test to reproduce this, but once you have a fix, I can verify it using Albany tests, which use Stratimikos.

cgcgcg:
@trilinos/muelu 

## Motivation
Try to fix #9590


lucbv (dismissed):
lgtm

jhux2 (approved):
lgtm

PR #9603, opened 2021-08-23
panzer: Main driver1 uvm free for tpetra blocked 

bathmatt:
@trilinos/panzer

## Motivation
as part of the UVM removal we got the main driver running to completion for the blocked tpetra stack.


rppawlo (approved):

bathmatt:
@trilinos/framework I'm seeing failures in one of your tests where I only touched panzer code, is there a problem currently or is there something I did?

bathmatt:
OK, good to know it isn't my changes so I don't need to track them down.

e10harvey:
The python-3 config should be fixed after #9610 merges.

PR #9619, opened 2021-08-26
Framework: Reset boostrap code path

prwolfe:
@trilinos/framework 

## Motivation
It's confusing to have the old paths in the logs and that path will be going away at some point

## Testing
Unit test pass and I ran one instance by hand with these changes to trigger the sha1 change (had to be in target of course.)


prwolfe:
> > > Should `cmake/std/PullRequestLinuxDriver.sh` be removed as part of this PR?
> > 
> > 
> > Not until we stop using it. Most of the PR->develop jobs are still using this.
> 
> @prwolfe I approved but disabled `AT:Automerge` for now. Just to double check and parrot what you state here. This won't break any of the existing PR's because the ones using the existing framework will still use the framework in `cmake/std`... this only fixes issue in the new `packages/framework` location where we were doing merge stuff against `cmake/std/` rather than `packages/framework`... If I'm following this right feel free to throw the automerge back on.

This was my understanding. I know the e-mail was long, think of it as my process notes while I was figuring this out ;->

Issue #9583, opened 2021-08-18
Tpetra/Belos: multiply with locally empty multivectors produces incorrect results


pohm01:
## Bug Report
@trilinos/tpetra @trilinos/belos

### Description
Problematic uninitialized values propagated in reduce call when multiply done using multivectors that are locally empty on a process.

If you do the following Tpetra::MultiVector multiply call

`C->multiply(trans, no_trans, alpha, A, B, beta)`

where `beta=0` and and `alpha=1` (i.e. C = A^T * B) and memory is distributed to trigger Case 2 in the multiply routine (Case 2: C(local) = A^T(distr) * B  (distr)). A problem arises when `A` and `B` have `getLocalLength()=0` on a process. When this occurs the local computation (in this case KokkosBlas::gemm) does not do anything to the local `C` since the local vectors have zero length. Since `C` is allocated but the values are not initialized, it is left filled with random noise. So when `C->reduce()` is called the uninitialized values are included.

This currently manifests itself in Belos during orthogonalization. Specifically `packages/belos/src/BelosICGSOrthoManager.hpp` line `1252` (`MatOrthoManager<ScalarType,MV,OP>::innerProd(*Q[i],X,MX,*C[i]);`).

### Steps to Reproduce
Currently happens in a Drekar simulation.
Here is a PR with small test that reproduces the error in Tpetra: #9593

jennloe:
@GrahamBenHarper Was this the issue you mentioned to me a few days ago?

GrahamBenHarper:
@jennloe yep!

jennloe:
The definition of `MatOrthoManager::innerProd()` is at `BelosMatOrthoManager.hpp` line 115-145.  This is probably failing in line 144: ` MVT::MvTransMv(SCT::one(),*P,*Q,Z);`.  

Looking at `BelosMultiVecTraits_Tpetra.hpp`, the `MvTransMV` (line 446) calls the following:
```
475       MV C_mv = makeStaticLocalMultiVector (A, numRowsC, numColsC);
476       // Filling with zero should be unnecessary, in theory, but not
477       // in practice, alas (Issue_3235 test fails).
478       C_mv.putScalar (ZERO);
479       C_mv.multiply (Teuchos::CONJ_TRANS, Teuchos::NO_TRANS, alpha, A, B, ZERO);
480       Tpetra::deep_copy (C, C_mv);
```

[...]

PR #9635, opened 2021-08-30
DO NOT MERGE:  kokkos-kernels:  commented out early exit that caused #9583

kddevin:
See also https://github.com/kokkos/kokkos-kernels/issues/1088
See #9583

Don't merge until we check with @trilinos/kokkos-kernels 
But I'd like to see whether the tests run.  @pohm01 's reproducer in #9583 passes with this change.

@lucbv @srajama1 

@trilinos/<teamName>


kddevin:
The correct fix for #9583 is in https://github.com/kokkos/kokkos-kernels/pull/1091 , which is now merged into Trilinos.
A Tpetra unit test checking the condition is in #9819.  
This PR should be closed and the branch removed.

PR #9641, opened 2021-09-01
Fix various windows build issues

Keno:
@trilinos/epetra @trilinos/teuchos @trilinos/ifpack

## Motivation

Fixes a number of minor issues when building for windows (in particular a MinGW cross compile from linux - though most of the issues are applicable to MSVC also). Details of each particular change are in the commit message. With these changes, my specific configuration of Trilinos builds fully for Windows.

## Testing

Automated build testing for all platforms I have access to was done in https://github.com/JuliaPackaging/Yggdrasil/pull/3569. No functional changes expected.

### DIFF
--- a/cmake/tribits/core/package_arch/TribitsConstants.cmake
+++ b/cmake/tribits/core/package_arch/TribitsConstants.cmake
@@ -96,11 +96,11 @@ set(${PROJECT_NAME}_PACKAGE_DEPS_FILES_DIR "cmake/dependencies")
 
 # Other stuff
 
-if(WIN32)
+if(${CMAKE_HOST_SYSTEM_NAME} STREQUAL "Windows")
   #Apparently FIND_PROGRAM looks for an exact match of the file name.
   #So even though "git clone ..." is valid to use on windows we need to give the
   #full name of the command we want to run.
-  set(GIT_NAME git.cmd)
-else(WIN32)
-  set(GIT_NAME git)
-endif(WIN32)
+  SET(GIT_NAME git.cmd)
+else()
+  SET(GIT_NAME git)
+endif()
--- a/packages/epetra/src/Epetra_C_wrappers.h
+++ b/packages/epetra/src/Epetra_C_wrappers.h
@@ -55,8 +55,8 @@ typedef long long  * EPETRA_LONG_LONG;
 
 #ifdef EPETRA_ADDRESS64BIT
 
-typedef long int EPETRA_OBJECT_PTR;
-typedef long int & EPETRA_OBJECT_REF;
+typedef long long int EPETRA_OBJECT_PTR;
+typedef long long int & EPETRA_OBJECT_REF;
 
 #else
 
--- a/packages/ifpack/src/Ifpack_SerialTriDiSolver.h
+++ b/packages/ifpack/src/Ifpack_SerialTriDiSolver.h
@@ -127,7 +127,7 @@ Examples using Ifpack_SerialTriDiSolver can be found in the Epetra test director
 */
 
 //=========================================================================
-class EPETRA_LIB_DLL_EXPORT Ifpack_SerialTriDiSolver :
+class Ifpack_SerialTriDiSolver :
   public Epetra_CompObject, public Epetra_BLAS,
   public Epetra_Object    {
   public:
--- a/packages/teuchos/numerics/src/Teuchos_BLAS.cpp
+++ b/packages/teuchos/numerics/src/Teuchos_BLAS.cpp
@@ -113,11 +113,11 @@ namespace Teuchos {
 //resolving them when linking dlls.
 #ifdef _MSC_VER
 #  ifdef HAVE_TEUCHOS_COMPLEX
-     template BLAS<long int, std::complex<float> >;
-     template BLAS<long int, std::complex<double> >;
+     template class BLAS<long int, std::complex<float> >;
+     template class BLAS<long int, std::complex<double> >;
 #  endif
-     template BLAS<long int, float>;
-     template BLAS<long int, double>;
+     template class BLAS<long int, float>;
+     template class BLAS<long int, double>;
 #endif
 
   // *************************** BLAS<int,float> DEFINITIONS ******************************
--- a/packages/teuchos/parameterlist/src/Teuchos_ParameterListAcceptorDefaultBase.hpp
+++ b/packages/teuchos/parameterlist/src/Teuchos_ParameterListAcceptorDefaultBase.hpp
@@ -58,7 +58,7 @@ namespace Teuchos {
  * using the non-virtual protected members <tt>setMyParamList()</tt> and
  * <tt>getMyParamList()</tt>.
  */
-class ParameterListAcceptorDefaultBase : virtual public ParameterListAcceptor {
+class TEUCHOSPARAMETERLIST_LIB_DLL_EXPORT ParameterListAcceptorDefaultBase : virtual public ParameterListAcceptor {
 public:
 
   /** \name Overridden from ParameterListAcceptor */
--- a/packages/epetra/src/Epetra_Import_Util.h
+++ b/packages/epetra/src/Epetra_Import_Util.h
@@ -67,7 +67,8 @@ namespace Epetra_Import_Util {
 
    \warning This method is intended for expert developer use only, and should never be called by user code.
  */
-int PackAndPrepareWithOwningPIDs(const Epetra_CrsMatrix & SourceMatrix,
+int EPETRA_LIB_DLL_EXPORT PackAndPrepareWithOwningPIDs(
+				 const Epetra_CrsMatrix & SourceMatrix,
 				 int NumExportIDs,
 				 int * ExportLIDs,
 				 int & LenExports,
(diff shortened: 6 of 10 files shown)


jhux2:
@trilinos/epetra @trilinos/teuchos @trilinos/ifpack

jhux2:
I haven't reviewed, but I will approve to let the AT run.  **DO NOT MERGE** until reviewed.

jhux2 (approved):
Approving to let the autotester run.

cmake/tribits/core/package_arch/TribitsConstants.cmake
```diff
@@ -96,11 +96,11 @@ SET(${PROJECT_NAME}_PACKAGE_DEPS_FILES_DIR "cmake/dependencies")
 
 # Other stuff
 
-IF(WIN32)
+IF(${CMAKE_HOST_SYSTEM_NAME} STREQUAL "Windows")
```
bartlettroscoe, line 99: What happens on Cygwin?

Keno: CMAKE_HOST_SYSTEM_NAME is CYGWIN on cygwin. Up to date versions of CMake do not define WIN32 on cygwin unless cross compiling using mingw (which is the correct behavior, because cygwin proper is POSIX not WIN32). As a result this change should not affect behavior on cygwin (and in particular, if cygwin git is installed it should work here). The one caveat there is that older versions of CMake do incorrectly define WIN32 on cygwin. I believe the behavior as implemented is correct, because building on cygwin, you do want to use the cygwin git build if at all possible to pick up any cygwin specific representations (e.g. symbolic links). It is possible that invoking the WIN32 git would work also, but it's be a bit of a Frankenstein thing to do.

packages/teuchos/parameterlist/src/Teuchos_ParameterListAcceptorDefaultBase.hpp
```diff
@@ -58,7 +58,7 @@ namespace Teuchos {
  * using the non-virtual protected members <tt>setMyParamList()</tt> and
  * <tt>getMyParamList()</tt>.
  */
-class ParameterListAcceptorDefaultBase : virtual public ParameterListAcceptor {
+class TEUCHOSPARAMETERLIST_LIB_DLL_EXPORT ParameterListAcceptorDefaultBase : virtual public ParameterListAcceptor {
```
bartlettroscoe, line 61: Okay, we are just adding an export.  I can't see how this can break an existing windows build.

cmake/tribits/core/package_arch/TribitsConstants.cmake
```diff
@@ -96,11 +96,11 @@ SET(${PROJECT_NAME}_PACKAGE_DEPS_FILES_DIR "cmake/dependencies")
...
   #Apparently FIND_PROGRAM looks for an exact match of the file name.
   #So even though "git clone ..." is valid to use on windows we need to give the
   #full name of the command we want to run.
   SET(GIT_NAME git.cmd)
-ELSE(WIN32)
+ELSE()
   SET(GIT_NAME git)
-ENDIF(WIN32)
+ENDIF()
```
bartlettroscoe, line 106: Okay, we are just fixing the logic for modern CMake versions on Windows and this will still use 'git' for cygwin.

packages/teuchos/numerics/src/Teuchos_BLAS.cpp
```diff
@@ -113,11 +113,11 @@ namespace Teuchos {
...
 #  ifdef HAVE_TEUCHOS_COMPLEX
-     template BLAS<long int, std::complex<float> >;
-     template BLAS<long int, std::complex<double> >;
+     template class BLAS<long int, std::complex<float> >;
+     template class BLAS<long int, std::complex<double> >;
 #  endif
-     template BLAS<long int, float>;
-     template BLAS<long int, double>;
+     template class BLAS<long int, float>;
+     template class BLAS<long int, double>;
```
bartlettroscoe, line 120: This is inside of an ifdef for `_MSC_VER` so Trilinos PR testing will **not** test these changes.  But this looks like correct C++ code (and I am surprised that it compiled the way it was).

packages/epetra/src/Epetra_C_wrappers.h
```diff
@@ -55,8 +55,8 @@ typedef long long  * EPETRA_LONG_LONG;
 
 #ifdef EPETRA_ADDRESS64BIT
 
-typedef long int EPETRA_OBJECT_PTR;
-typedef long int & EPETRA_OBJECT_REF;
+typedef long long int EPETRA_OBJECT_PTR;
+typedef long long int & EPETRA_OBJECT_REF;
```
bartlettroscoe, line 59: This seems reasonable.  C++11 added official support for `long long` int and it is guaranteed to be at least 64 bits.  So I would think this is the right change to make to achieve that.

packages/ifpack/src/Ifpack_SerialTriDiSolver.h
```diff
@@ -127,7 +127,7 @@ Examples using Ifpack_SerialTriDiSolver can be found in the Epetra test director
 */
 
 //=========================================================================
-class EPETRA_LIB_DLL_EXPORT Ifpack_SerialTriDiSolver :
```
bartlettroscoe, line 130: Will removing this export not break some customer's usage of this class from the DLL?  Why is this export being removed?  Why not leave this export?

bartlettroscoe (dismissed):
I am good with the TriBITS and Teuchos changes (which is what I am approving here).   I personally think that the change to `Epetra_C_wrappers.h` is okay as well that changes to `long long` for 64 bit integers (but will let some @trilinos/epetra person approve that).

All of the rest of the changes all appear to be just changes to DLL exports.  All of the changes are adding exports except for one case where an export is removed (for the class `Ifpack_SerialTriDiSolver` in the file `Ifpack_SerialTriDiSolver.h`).  Except for that one export removal,  I can't see how adding additional exports can break anything.

Others will need to review and approve the remaining changes for Epetra and Ifpack but @trilinos/ifpack developers should look carefully at the export being removed in the file `Ifpack_SerialTriDiSolver.h` and make sure that is okay.

NOTE: After this PR is merged, I will cherry-pick off the TriBITS change and apply it back to the TriBITS repo and snasphot TriBITS back to Trilinos to address any possible future merge conflicts.

packages/ifpack/src/Ifpack_SerialTriDiSolver.h
Keno: As described in the commit message, the export indicates epetra, but this class is in ifpack - as a result this was never exported in the first place and the only effect it has is breaking the build of ifpack on windows. It is certainly possible that this should have an ifpack export, but I felt this to be the more conservative change.

bartlettroscoe: Okay, I see that in commit c4d2092687991bfc657b0440ab78d2652a39bec4.  I should have looked at that.

With that, exports are only being added in this PR so I don't see any reason not to merge this PR.

bartlettroscoe:
@Keno, if no-one else approves this in the next few days, ping back here with a comment and CC those of us that commented here already.

bartlettroscoe:
@trilinos/framework, are there any automated builds of Trilinos for Windows that might be impacted (for better or worse) by this PR?

jhux2:
@jrobcary Does this PR affect your work?

jrobcary:
It looks innocuous, so no objection here.  We have stayed with what we got working.

bartlettroscoe:
@trilinos/framework, @trilinos/epetra, @trilinos/ifpack, @trilinos/ifpack, there is now a conflict here (likely with PR #9663 that was just merged).  Unless there are objections, I propose we resolve this conflict, run another round of PR testing and merge.

jhux2:
> there is now a conflict here (likely with PR #9663 that was just merged). Unless there are objections, I propose we resolve this conflict, run another round of PR testing and merge.

@bartlettroscoe I'm in favor of this.

kddevin (approved):
The Epetra changes look harmless.  I don't know what EPETRA_LIB_DLL_EXPORT does, and it certainly clutters the code, but the changes here match existing use of EPETRA_LIB_DLL_EXPORT in functions and classes in Epetra.

jhux2:
@Keno If you are able to resolve the conflicts, then this will be ready to merge.

Keno:
> @Keno If you are able to resolve the conflicts, then this will be ready to merge.

Done

Keno:
> I don't know what EPETRA_LIB_DLL_EXPORT does

It marks DLL visibility for windows. Classes without this attribute will not be visible from downstream dlls. Other platforms have a similar concept, but with the opposite default.

jhux2 (approved):
lgtm

bartlettroscoe:
FYI: I pulled off the commit patch from commit 5dee9af974810141b7f1ea54a21ee3e6998f380e and applied to the TriBITS repo in https://github.com/TriBITSPub/TriBITS/pull/422 (along with fixing the command case).  This will create a merge conflict on the next TriBITS snapshot back to Trilinos but it will be trivial to resolve.

PR #9651, opened 2021-09-02
Intrepid2: fix shadowed DeviceType declarations in function scope

japlews:
@trilinos/Intrepid2
@mperego @kyungjoo-kim

## Motivation

Shadowed declaration of `DeviceType` in class scope; there are several other declarations in function scope which seem to be no longer needed.

## Stakeholder Feedback

N/A

## Testing

N/A


CamelliaDPG (approved):
Looks good to me.  Thanks!

PR #9652, opened 2021-09-02
Fix PanzerAdaptersSTK_STK_ResponseLibraryTest2_MPI_2 with UVM off

JacobDomagala:


packages/panzer/disc-fe/src/responses/Panzer_ResponseScatterEvaluator_Functional.hpp
```diff
@@ -159,13 +162,15 @@ void FunctionalScatter<LO,GO>::scatterDerivative(const PHX::MDField<const panzer
     for(std::size_t b=0;b<ugis_.size();b++) {
       int start = blockOffsets[b];
 
-      LIDs = ugis_[b]->getElementLIDs(cellLocalId); 
+      LIDs = ugis_[b]->getElementLIDs(cellLocalId);
+      auto LIDs_h = Kokkos::create_mirror_view(LIDs);
+      Kokkos::deep_copy(LIDs_h, LIDs);
```
bathmatt, line 168: Can you use the same trick you did in the blocked epetra and move the LIDs_h outside the worksetCellIndex loop?  This will replace 1000s of deep copies with just 1 at the start

bathmatt (approved):

PR #9654, opened 2021-09-02
Panzer: Stops virtual orientations from being applied

seamill:
… cells

@trilinos/panzer


seamill:
@rppawlo Could you review? It appears to be backward compatible with empire, and shouldn't affect panzer/charon.

rppawlo (approved):

PR #9661, opened 2021-09-03
Tpetra:  reduce false alarms from test Bug7745.cpp

kddevin:
@trilinos/tpetra 

## Motivation

The test Bug7745.cpp has more than normal false-alarm failures.  It spewed a lot of output, which would sometimes mess up the output that the Teuchos test harness seeks to determine pass/fail.
This PR reduces the amount of output produced by the test, with the hope that reducing the output will reduce the false alarms.

## Testing

Tested on ascicgpu033


tasmith4 (approved):
This looks great to me.  If this doesn't fix it maybe instead of hardcoding VERB_LOW we can use an environment variable to turn it on (Tpetra has one, right?) so it's always off unless the user actually asks for it, presumably because they're debugging the test.

PR #9677, opened 2021-09-09
found missing deep copy

bathmatt:
Fixed missing deep copy if conditional was not set to true


rppawlo (approved):
Thanks @bathmatt ! This one is my fault! Sorry.

bathmatt:
@rppawlo how do we resolve this, assume a merge with develop once it gets the right fix in?

rppawlo:
The intel compiler died with:
```
icpc: error #10106: Fatal error in /projects/sems/install/rhel7-x86_64/sems/compiler/intel/17.0.1/base/compilers_and_libraries_2017.1.132/linux/bin/intel64/mcpcom, terminated by kill signal
```
Think this is a system hiccup. It failed on a file we did not touch and there were muelu and stokhos failures as well. Hoping a retest will go through...

rppawlo:
Once this goes in, I think we can enable panzer in the no UVM build! Thanks for all your hard work on this!

PR #9682, opened 2021-09-10
panzer: Create mirror fixes

bathmatt:
found some places that needed views to be unmanaged


rppawlo (approved):

bathmatt:
Odd, it failed to build with UVM enabled, even though we build without UVM enabled.  @rppawlo I can't do much about this right now, sorry

rppawlo (approved):

Issue #9690, opened 2021-09-13
MueLu:  Indexing bug


skennon10:
## Bug Report
@trilinos/\<MueLu\>

### Description
In this line: https://github.com/trilinos/Trilinos/blob/42120c4935a39eac1bb069115b96c2fea8d62ff4/packages/muelu/src/Interface/MueLu_MLParameterListInterpreter_def.hpp#L620
it should be 
```
coordPtr[i][j]
```
As is, it is putting only x-coords into the coordinates array.
@cgcgcg

cgcgcg:
@skennon10 This should be fixed in develop.

Issue #9686, opened 2021-09-11
Framework: autotester CUDA instance failing in all PRs


jhux2:
## Bug Report
@trilinos/framework @ccober6 

### Description
As can be seen in the AT results for PRs #9682, #9532, and probably others, results for the cuda builds are not appearing.  Everything else is green.  It looks like this started sometime late Thursday or early Friday.  In the respective PRs, in the message that testing failed, there are messages like this:

```

LSB_EXIT_PRE_ABORT = 99
GIT_COMMIT_1 = 6b18fbd791032f935925d2a3ea1c94ca51e173fa
NODE_OWNER = ceeadmin
LSB_JOBEXIT_STAT = 0
PWD = /home/trilinos/jenkins/ride/workspace/trilinos-folder/Trilinos_pullrequest_cuda_10.1.105
_LMFILES_ = /home/projects/ppc64le/modulefiles/git/2.10.1:/home/projects/ppc64le/modulefiles/python/3.7.3
LSB_INTERACTIVE = Y
GIT_COMMIT_2 = 83804082aada11571975f7d9db53d353e5d9d690
HUDSON_URL = https://ascic-jenkins.sandia.gov/
JOB_OWNER_EMAIL = HRSwant@sandia.gov
LANG = en_US.UTF-8
LSB_RES_GET_FANOUT_INFO = Y
MODULEPATH = /usr/share/Modules/modulefiles:/etc/modulefiles:/home/projects/sparc/modules/ride:/projects/sierra/common_modules/ride:/home/projects/ppc64le/modulefiles:/home/projects/ppc64le-pwr8/modulefiles:/home/projects/ppc64le-pwr8-nvidia/modulefiles
JOB_NAME = trilinos-folder/Trilinos_pullrequest_cuda_10.1.105
LSB_CHKFILENAME = /ascldap/users/trilinos/.lsbatch/1631297091.17416
LOADEDMODULES = git/2.10.1:python/3.7.3
LSB_DJOB_HOSTFILE = /ascldap/users/trilinos/.lsbatch/1631297091.17416.hostfile
BUILD_DISPLAY_NAME = #2230
BUILD_CAUSE = MANUALTRIGGER
JENKINS_URL = https://ascic-jenkins.sandia.gov/
BUILD_ID = 2230
LSF_JOB_TIMESTAMP_VALUE = 1631297091
TRILINOS_SOURCE_REPO = https://github.com/trilinos/Trilinos
[...]

PR #9691, opened 2021-09-13
MueLu/Xpetra stuff

cgcgcg:
@trilinos/muelu 

## Motivation
- Don't warn about mismatch of matrix block sizes, unless the user actually set a block size on the input matrix.
- Fix a bug in MLParameterListInterpreter #9690.
- Expose some additional args of `GetMatrixDiagonal`.


cgcgcg:
Blocked by #9686.

jhux2 (dismissed):
lgtm

cgcgcg:
FYI @mperego

mperego:
> FYI @mperego

Thanks!! And sorry I forgot to send you the error message.

mayrmt:
@cgcgcg Do you see the possibility to cover the bug reported in #9690 with a test to prevent it in the future?

cgcgcg:
@mayrmt I moderately improved the MueLu-ML tests by passing coordinates and nullspace through `MLParameterListInterpreter`. Unfortunately though, all our interface tests are using a 1D operator, so that particular bug would not have been caught.

mperego:
@cgcgcg should we test this with Albany before it is merged, or are you relatively confident that this is not going to break Albany?

cgcgcg:
@mperego I don't expect it to break Albany, but that doesn't mean it won't. However, I don't think that is is any more likely than any other PR to cause issues.

jhux2 (approved):
Zehr gut

PR #9695, opened 2021-09-14
MueLu: Tag isolated nodes as Dirichlet

cgcgcg:
@trilinos/muelu 

## Motivation
When `CoalesceDrop` is dropping quite aggressively, it can happen that only the diagonal element of a row remains. In that case, this row should also be tagged as a Dirichlet boundary.

### DIFF
--- a/packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
+++ b/packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
@@ -635,6 +635,13 @@ namespace MueLu {
 
           columns.resize(realnnz);
           numTotal = A->getNodeNumEntries();
+
+          // If the only element remaining after filtering is diagonal, mark node as boundary
+          for (LO row = 0; row < Teuchos::as<LO>(A->getRowMap()->getNodeNumElements()); ++row) {
+            if (rows[row+1]- rows[row] <= 1)
+              boundaryNodes[row] = true;
+          }
+
           RCP<GraphBase> graph = rcp(new LWGraph(rows, columns, A->getRowMap(), A->getColMap(), "thresholded graph of A"));
           graph->SetBoundaryNodeMap(boundaryNodes);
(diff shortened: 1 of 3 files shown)


packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
```diff
@@ -635,6 +635,13 @@ namespace MueLu {
 
           columns.resize(realnnz);
           numTotal = A->getNodeNumEntries();
+
+          // If the only element remaining after filtering is diagonal, mark node as boundary
+          for (LO row = 0; row < Teuchos::as<LO>(A->getRowMap()->getNodeNumElements()); ++row) {
+            if (rows[row+1]- rows[row] <= 1)
+              boundaryNodes[row] = true;
```
lucbv, line 642: Marking nodes as boundary when they are not boundary nodes seems wrong, I would rather have them tagged as `IGNORED` for aggregation, actually this is something that is achieved by running the `IsolatedNodeAggregationAlgorithm`.

lucbv (changes requested):
Boundary condition is not the right state for such nodes, they should be marked as isolated or ignored instead.

packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
cgcgcg: I agree that that would be cleaner. It looks like `IsolatedNodeAggregationAlgorithm` has been inactive since 2015 https://github.com/trilinos/Trilinos/commit/7e892ecb58513a6e6f9017f1e277915713a8f008#diff-4abf20d14b91dda5281d2427a0075a78ed80ca9a39342e1362fd3ccba6f741ef, and instead this functionality has been put into `CoalesceDrop`. Not sure why.

lucbv (approved):
Sure, we discussed this issue at the stand-up and decided to tackle the Isolated vs. Dirichlet in another PR.

PR #9697, opened 2021-09-14
Adding preprocessor directives to disable Hessian interface in thyra ME.

crockatt:
@trilinos/thyra @rppawlo @mperego 

## Motivation

Certain simulations we are running in drekar use on the order of 100 responses and parameters each. In these cases, some of the arrays allocated for the Hessian support in thyra OutArgs contain on the order of 1e+6 elements. Even though we are not using any of the Hessian derivative objects for these simulations (the RCP arrays contain all null RCPs), there is a non-negligible cost (memory and runtime) associated with allocating, filling, and deallocating these arrays for each OutArgs object.

This pull request defines a preprocessor flag and corresponding cmake variable to allow us to disable the Hessian objects in the thyra ME interface during compilation to circumvent the performance issues we've been seeing. The default settings preserve the current behavior (compile with the Hessian objects included).

## Related Issues

* Closes rppawlo/DrekarBase#442


mperego (approved):
Looks good to me. Thanks for figuring this out.
You could add a comment, e.g. where you introduce the flag in the CMakeList file, about why this is needed.

mperego (approved):

crockatt:
Thanks for reviewing this @mperego. Can you do the merge now? I do not have permissions to do it myself.

PR #9700, opened 2021-09-14
Adding flag to NOX line search solver to enable catching of exceptions thrown by linear solvers

crockatt:
@trilinos/nox @rppawlo 

## Motivation

A number of packages inside the linear/nonlinear solver stack will throw an exception if a bad state is encountered. For time-dependent simulations that we are running in drekar, it is often possible to recover from such errors by reducing the timestep size and trying again. This pull request adds a boolean option `"Catch Throws During Solve"` for the line-search based solver in NOX that will:
- Catch an exception that is thrown during the call to `solve`.
- Print out a warning containing the text string defined within the exception that is thrown.
- Mark and return a solver status of `Failed`, so that the time integrator can make adjustments as needed.

## Related Issues

* Closes rppawlo/DrekarBase#375

## Testing

Brief testing in drekar gives the expected behavior with and without the flag present. I am not sure if there is a good place to put a permanent test in NOX. Not sure if you have any suggestions @rppawlo or if you think a test is even needed here.

### DIFF
--- a/packages/nox/src/NOX_Solver_LineSearchBased.C
+++ b/packages/nox/src/NOX_Solver_LineSearchBased.C
@@ -95,6 +95,11 @@ void NOX::Solver::LineSearchBased::init()
   directionPtr = NOX::Direction::
     buildDirection(globalDataPtr, paramsPtr->sublist("Direction"));
 
+  if ( paramsPtr->isType<bool>( "Catch Throws During Solve" ) )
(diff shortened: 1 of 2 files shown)


rppawlo (approved):
Thanks @crockatt ! Looks good to me. Let's get this into the repo to support drekar. Next week I'll be working on some nox items and can add some code to test this.

PR #9704, opened 2021-09-15
Update panzer Tpetra (blocked and non-blocked) ScatterResidual evaluators to handle arbitrary numbers of LIDs

crockatt:
@trilinos/panzer @rppawlo 

## Motivation

The existing implementations of the Jacobian specializations of the ScatterResidual evaluators for Tpetra in panzer make use of temporary arrays with a fixed size of 256 elements. When running problems with large numbers of equations, it is possible to exceed this limit. This pull request modifies the implementation of these evaluators to use Kokkos::Views as scratch space, in order to support arbitrary system sizes.

## Related Issues

* Closes rppawlo/DrekarBase#379

## Testing

Existing tests are passing on a serial CPU build. The performance impact on CUDA builds has not yet been evaluated.


rppawlo (approved):
@crockatt - looks good to me! Thanks fro doing this! Before we commit, I would like to run some timings on cuda to see how this impacts scatter performance. I will hit approve to get the autotester to check the code, but please hold off committing if it passes.

rppawlo:
No impact on cuda performance for blocked tpetra scatter! Merging.

PR #9709, opened 2021-09-16
Panzer: Lazy evaluate integration values 2 from workset

seamill:
@trilinos/panzer


seamill:
@rppawlo I think this is ready to go. I get some exodus errors in empire, but panzer passes just fine.

packages/panzer/disc-fe/src/Panzer_IntegrationRule.cpp
```diff
@@ -74,23 +74,29 @@ IntegrationRule(const panzer::CellData& cell_data, const std::string & in_cv_typ
...
   // TODO: These isSide conditions shouldn't be required... I'm missing something...
   if(in_cv_type == "volume"){
-    if(cell_data.isSide()){
-      IntegrationDescriptor::setup(75, IntegrationDescriptor::CV_VOLUME,cell_data.side());
-    } else {
+    TEUCHOS_TEST_FOR_EXCEPT_MSG(cell_data.isSide(),
+                                "IntegrationRule::IntegrationRule : Control Volume 'volume' type requested, but CellData is setup for sides.");
+//    if(cell_data.isSide()){
+//      IntegrationDescriptor::setup(75, IntegrationDescriptor::CV_VOLUME,cell_data.side());
+//    } else {
```
rppawlo, line 81: This block and a few below are commented out. Have we lost capability here or is it just not needed. Should these be removed?

rppawlo (dismissed):

rppawlo:
@seamill - once the PR testing passes, but before we merge, I would like to run this through charon just to be safe with the changes to the CV support. Is this ok?

seamill:
@rppawlo Please do test Charon. I think the changes to commented on above are the only non-backward compatible change. We had some conflicting methods for creating CV side vs CV boundary integration rules. I made the appropriate changes in Panzer, but I'm unsure if Charon is building its own IntegrationRules.

packages/panzer/disc-fe/src/Panzer_IntegrationRule.cpp
seamill: Done

rppawlo:
@seamill - 67 of 168 are failing in Charon.
```
p=1 | *********** Caught Exception: Begin Error Report ***********
p=1 | ../Trilinos/packages/panzer/disc-fe/src/Panzer_IntegrationRule.cpp:81:
p=1 |
p=1 | Throw number = 1
p=1 |
p=1 | Throw test that evaluated to true: not cell_data.isSide()
p=1 |
p=1 | IntegrationRule::IntegrationRule : Control Volume 'side' type requested, but CellData is not setup for sides.
p=1 | ************ Caught Exception: End Error Report ************
```
will need your help on this.

seamill:
@rppawlo I think this was just me making a bad assumption. I pushed a fix that should get you past this error. Could you try again?

PR #9718, opened 2021-09-20
ML2MueLu translation of aux aggregation

### LINKED ISSUE #9715: MueLu:  typo in Maxwell3D.cpp
## Bug Report
@trilinos/\<muelu\>

### Description
https://github.com/trilinos/Trilinos/blob/7e09778fc8d832628e9cdd3c07f5b8f7a28fc9d1/packages/muelu/test/maxwell/Maxwell3D.cpp#L734
Should be `MueLu-Maxwell1`.

@csiefer2

cgcgcg:
@trilinos/muelu 

## Motivation
- Translate aux aggregation parameters from ML to MueLu.
- Fix #9715

### DIFF
--- a/packages/muelu/test/interface/default/MLParameterListInterpreter/MLaux.xml
+++ b/packages/muelu/test/interface/default/MLParameterListInterpreter/MLaux.xml
@@ -4,5 +4,7 @@
   <Parameter name="coarse: max size"            type="int"      value="100"/>
   <Parameter name="repartition: enable"         type="int"      value="1"/>
   <Parameter name="aggregation: aux: enable"    type="bool"     value="true"/>
+  <!-- Set threshold to something weird, since aux: threshold should be used. -->
+  <Parameter name="aggregation: threshold"      type="double"   value="-1e15"/>
   <Parameter name="aggregation: aux: threshold" type="double"   value="1e-2"/>
 </ParameterList>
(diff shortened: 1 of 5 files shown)


packages/muelu/test/interface/default/MLParameterListInterpreter/MLaux.xml
```diff
@@ -4,5 +4,7 @@
   <Parameter name="coarse: max size"            type="int"      value="100"/>
   <Parameter name="repartition: enable"         type="int"      value="1"/>
   <Parameter name="aggregation: aux: enable"    type="bool"     value="true"/>
+  <!-- Set threshold to something weird, since aux: threshold should be used. -->
+  <Parameter name="aggregation: threshold"      type="double"   value="-1e15"/>
```
jhux2, line 8: Does the XML translator parse this correctly?

cgcgcg: With the change to ML2MueLuParameterTranslator, it should see that aux aggregation is enabled, and then overwrite "aggregation: threshold" with "aggregation: aux: threshold" before translating that to "aggregation: drop tol". What happened before is that "sa: use filtered matrix" was always on, and it would use "aggregation: threshold" for the drop tol.

jhux2 (dismissed):
lgtm

jhux2 (approved):
lgtm

PR #9722, opened 2021-09-21
Tempus: add default flag to delay initialization

sconde:
Edge case: when app is only using its own TSCS in a composite, TSC
fails initialization because the app has not had the chance to add its
TSCS to the compose vector yet. So this `runInitialize` (default: true)
allow the app to delay the initialization until it has populated the
composite vector and explicitly calls the initialization.

@trilinos/tempus 

## Motivation

In Drekar when only using `TimeScale` TSCS, Tempus tries initializing TSC but the vector size is zero since Drekar has not had the chance to add the App's own TSCS. This is not the case if the `Time Step Control Strategy List` contains at least one TSCS that Tempus can add to the composite vector. But if all the strategies are unique to the App, the composite vector size is zero and fails the assertion and the initialization.

### DIFF
--- a/packages/tempus/test/ExplicitRK/Tempus_ExplicitRKTest.cpp
+++ b/packages/tempus/test/ExplicitRK/Tempus_ExplicitRKTest.cpp
@@ -80,10 +80,20 @@ TEUCHOS_UNIT_TEST(ExplicitRK, ParameterList)
       tempusPL->sublist("Demo Stepper").set("Stepper Type", RKMethods[m]);
     }
 
-    // Test constructor IntegratorBasic(tempusPL, model)
+    // Test constructor IntegratorBasic(tempusPL, model, runInitialize)
     {
       RCP<Tempus::IntegratorBasic<double> > integrator =
-        Tempus::createIntegratorBasic<double>(tempusPL, model);
+        Tempus::createIntegratorBasic<double>(tempusPL, model,false);
+
+      // check initialization
+      {
+        // TSC should not be initialized
+        TEST_ASSERT(!integrator->getNonConstTimeStepControl()->isInitialized());
+        // now initialize everything (TSC should also be initilized)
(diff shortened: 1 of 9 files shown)


ccober6 (changes requested):
This looks good!  Can you add a unit test to cover this new feature?  Thanks!

packages/tempus/test/ExplicitRK/Tempus_ExplicitRKTest.cpp
```diff
@@ -80,10 +80,20 @@ TEUCHOS_UNIT_TEST(ExplicitRK, ParameterList)
       tempusPL->sublist("Demo Stepper").set("Stepper Type", RKMethods[m]);
     }
 
-    // Test constructor IntegratorBasic(tempusPL, model)
+    // Test constructor IntegratorBasic(tempusPL, model, runInitialize)
     {
       RCP<Tempus::IntegratorBasic<double> > integrator =
-        Tempus::createIntegratorBasic<double>(tempusPL, model);
+        Tempus::createIntegratorBasic<double>(tempusPL, model,false);
```
sconde, line 86: @ccober6 test added

rppawlo (approved):

sconde:
None of the failures are due to the changes in Tempus.

rppawlo:
> None of the failures are due to the changes in Tempus.

@sconde - maybe it is my system/build? I ran drekar against trilinos develop and had 7 failures without the tempus changes, but when I merged in these changes I got 8 failures. The new failure was the plasma oscillation problem.

sconde:
> The new failure was the plasma oscillation problem.

This test has always been failing in Drekar. 


> None of the failures are due to the changes in Tempus.

I'm refering to the  PR's autotester failing. I'm going to try building with ROL on locally to see.

ccober6 (approved):

PR #9723, opened 2021-09-21
Fix RCP initialization in NOX thyra group to avoid null pointer dereference

crockatt:
@trilinos/nox @rppawlo 

## Motivation

Certain tests in drekar have run into an issue when right scaling weights are used that results in a null pointer dereference and segfault. This has been traced back to `scaled_x_vec_` not being initialized in all cases. This pull request resolves the issue by making sure `scaled_x_vec_` gets allocated whenever `right_weight_vec_` is not null.

## Related Issues

* Closes rppawlo/DrekarBase#583


rppawlo (approved):
Thanks @crockatt !

PR #9739, opened 2021-09-24
Phalanx: add conversion utility

rppawlo:
@trilinos/phalanx 

## Motivation

Calling a TPL that does not respect const correctness. Need utility to remove const from kokkos views.

Also added stacked timer to an example in panzer.

## Stakeholder Feedback

Fixes issue.

## Testing

tested in phalanx/test/Field/NonConstDynRankViewFromConstMDField.cpp.


seamill (approved):
Looks good to me. Thanks!

PR #9747, opened 2021-09-27
MueLu: Make Phase2a do the same as ML

cgcgcg:
@trilinos/muelu 

## Motivation
Match ML's Phase2a:
https://github.com/trilinos/Trilinos/blob/e0adf23dfd61bca77cb13bc4c13f5432e188fd52/packages/ml/src/Coarsen/ml_agg_MIS.c#L2674-L2700


GrahamBenHarper (commented):
I know we've talked about it in standups a little bit, but did we address how it affects other existing codes that use MueLu? I worry a little bit about this changing the results for other codes that use MueLu, but I recall this was only for pretty specific situations.

cgcgcg:
I tried to track down how the phase changed with the merged PR #7254 and this PR.
- Originally, the root node was not included in the tentative aggregate and also not counted as a neighbor. The condition for acceptance as a new aggregate was `aggSize>factor*numNeighbors`.
- With PR7254, if `"aggregation: phase2a include root"=true` (default), the root node is included in the tentative aggregate, so `aggSize` is +1 compared to what it used to be, but the root wasn't counted as a neighbor. The acceptance condition was changed to `aggSize-1>factor*numNeighbors`. So the criterion didn't change, only the fact that the root node would be included in the new aggregate. When `"aggregation: phase2a include root"=false`, the behavior is exactly the same as before. (Nalu uses that.)
- With this PR and `"aggregation: phase2a include root"=true`, the root node is counted as a neighbor, and the criterion is changed to `aggSize>factor*numNeighbors`. This matches ML. Again, `"aggregation: phase2a include root"=false` still matches the original behavior.

jhux2 (approved):
lgtm!

jhux2:
@spdomin fyi

PR #9748, opened 2021-09-27
Xpetra: Epetra changes

cgcgcg:
@trilinos/xpetra 

## Motivation
- Add `randomize` method with limits to Xpetra's MV.
- Don't throw from `Xpetra::EpetraMultiVectorT::getDeviceLocalView`, but return a host view, add `setAllValues`.
- `EpetraCrsMatrixT` was caching its local matrix, since the rowptr had the wrong type and hence needed to be converted. I ran into some issue where the cache wasn't invalidated on reuse, so I removed the caching entirely.

### DIFF
--- a/packages/xpetra/src/MultiVector/Xpetra_MultiVector_def.hpp
+++ b/packages/xpetra/src/MultiVector/Xpetra_MultiVector_def.hpp
@@ -89,6 +89,26 @@ Xpetra_randomize()
     }
 }
 
+template<class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
+void
+MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>::
+Xpetra_randomize(const Scalar& minVal, const Scalar& maxVal)
+{
+    typedef Teuchos::ScalarTraits<Scalar> SCT;
+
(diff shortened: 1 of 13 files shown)


packages/xpetra/src/MultiVector/Xpetra_MultiVector_def.hpp
```diff
@@ -89,6 +89,26 @@ Xpetra_randomize()
...
+
+    const size_t numVectors = getNumVectors();
+    for(size_t i = 0; i < numVectors; i++)
+    {
+        Teuchos::ArrayRCP<Scalar> datai = getDataNonConst(i);
+
+        const size_t myLength = getLocalLength();
+        for(size_t j = 0; j < myLength; j++)
+        {
+          datai[ j ] = 0.5*(maxVal-minVal)*SCT::random()+0.5*(maxVal+minVal);
```
jhux2, line 107: Is this really what you want?  See https://github.com/trilinos/Trilinos/blob/7527310aa8d57d922fca98bb3c0ea42a1ecf53bb/packages/ml/src/Utils/ml_utils.c#L1825.

jhux2 (dismissed):
lgtm

jhux2 (approved):
lgtm

Issue #9737, opened 2021-09-24
Tempus: transient adjoint sensitivities not working correctly when 'Initial Time' is non-zero


ikalash:
Problem can be reproduced by setting ‘Initial Time: 0.1’ in Tempus_BDF2_SinCos_SA.xml in packages/tempus/test/BDF2, and then running ./Tempus_BDF2_ASA.exe.  The following error is encountered:

```
p=0: *** Caught standard std::exception of type 'std::logic_error' :

  /home/ikalash/transientAdjointSensitivities/Trilinos/packages/tempus/src/Tempus_SolutionHistory_impl.hpp:185:
  
  Throw number = 1
  
  Throw test that evaluated to true: !(minTime() <= time && time <= maxTime())
  
  Error - SolutionHistory::findState() Requested time out of range!
          [Min, Max] = [0, 0]
          time = 0.9
  
 [FAILED]  (8.719e-03 sec) BDF2_SinCos_ASA_UnitTest
Location: /home/ikalash/transientAdjointSensitivities/Trilinos/packages/tempus/test/BDF2/Tempus_BDF2_ASA.cpp:46

```
@etphipp is aware of the problem and is looking at it.  I am opening the issue for documentation.

etphipp:
@ikalash This should be resolved now.

ikalash:
Thanks @etphipp .  yes I saw the PR get merged yesterday.  I am speaking in a conference this week and haven’t had a chance to try this out yet, but I will soon (either today or tomorrow).  Thanks!

ikalash:
I just tested this in Piro and Albany and I'm still getting an error, but a different one:

```

p=0: *** Caught standard std::exception of type 'std::logic_error' :

 /nightlyAlbanyTests/Results/Trilinos/packages/tempus/src/Tempus_SolutionHistory_impl.hpp:185:
 
 Throw number = 1
 
 Throw test that evaluated to true: !(minTime() <= time && time <= maxTime())
 
 Error - SolutionHistory::findState() Requested time out of range!
         [Min, Max] = [0.0001, 0.01]
         time = 0.0001
```
[...]

PR #9761, opened 2021-09-29
Tempus:  Fix two more adjoint sensitivity problems when t0 != 0

etphipp:
For issue #9737:
  * Add a fudge factor in findState() to allow for round-off
  * Reset timeFinal in the adjoint model evaluators after the forward integration, because it may not match the final time requested.


ccober6 (approved):
Looks good!

PR #9762, opened 2021-09-29
Framework: packages/framework: add --ctest-drop-site

e10harvey:
@trilinos/framework

## Motivation

Support posting to different cdash servers.

## Testing

Trilinos auto-tester.

The framework ctest driver scripts will need to be updated after this merges.


william76:
@e10harvey looks like the tests are failing on something with the python3 test of this.

Also, for PR's and Issues on Trilinos, the convention is generally to prefix these titles with`<packagename>:`... so this PR for example should be prefixed with "Framework:".

william76 (approved):

Issue #9615, opened 2021-08-25
Zoltan2:  printf warnings


tasmith4:
## Bug Report
@trilinos/zoltan2 

### Description
On several of the Sierra nightly platforms, we notice warnings similar to the ones below, from files needed by stk_balance.  It's not currently a big deal for the STK team, but we wanted you to be aware.  We're assuming the issue is in Zoltan2 code, but if something needs to be fixed in STK code instead, please let us know.

```
/sierra/build/linux_rh7/nightly/ats2_nvidia_10_surrogate_master/objs/tpls/spack/spack/spack_path_placeholder/spack_path_placehol/linux-rhel7-x86_64/gcc-7.2.0/trilinos-2021.08.09-qbsfj6zdn463nu77oykper5dvgj7ns66/include/Zoltan2_AlgHybridD1.hpp: In instantiation of 'void Zoltan2::AlgDistance1&lt;Adapter&gt;::hybridGMB(size_t, const Teuchos::ArrayView&lt;const typename Adapter::lno_t&gt;&, const Teuchos::ArrayView&lt;const typename Adapter::offset_t&gt;&, const Teuchos::RCP&lt;Tpetra::FEMultiVector&lt;int, typename Adapter::lno_t, typename Adapter::gno_t, Kokkos::Compat::KokkosDeviceWrapperNode&lt;Kokkos::Cuda&gt; &gt; &gt;&, const Teuchos::ArrayView&lt;const typename Adapter::gno_t&gt;&, const Teuchos::ArrayView&lt;const int&gt;&, const Teuchos::ArrayView&lt;const int&gt;&, Teuchos::RCP&lt;const Tpetra::Map&lt;typename Adapter::lno_t, typename Adapter::gno_t, Kokkos::Compat::KokkosDeviceWrapperNode&lt;Kokkos::Cuda&gt; &gt; &gt;, const std::unordered_map&lt;typename Adapter::lno_t, std::vector&lt;int&gt; &gt;&) [with Adapter = StkMeshZoltanAdapter; size_t = long unsigned int; typename Adapter::lno_t = int; typename Adapter::offset_t = int; typename Adapter::gno_t = long int; typename Adapter::gno_t = long int; typename Adapter::gno_t = long int; typename Adapter::lno_t = int; typename Adapter::lno_t = int]':
/tmp/sierra/tmpxft_00016c15_00000000-5_balance.cudafe1.stub.c:27:966:   required from here
[...]

PR #9764, opened 2021-09-30
Zoltan2: Removed some Zoltan2 debug print statements (Issue #9615)

djglaze:
that are causing build errors in Sierra.  These statements are printing
several things, including some IDs that are configurable as either
32-bit or 64-bit unsigned integers.  The format specifiers assume one
size and Sierra uses the other, leading to build warnings that we promote
to errors on many platforms.

@trilinos/<teamName>

### DIFF
--- a/packages/zoltan2/core/src/algorithms/color/Zoltan2_AlgHybrid2GL.hpp
+++ b/packages/zoltan2/core/src/algorithms/color/Zoltan2_AlgHybrid2GL.hpp
@@ -1640,7 +1640,6 @@ class AlgTwoGhostLayer : public Algorithm<Adapter> {
           }
         }      
         
-        if(comm->getRank() == 0) printf("did %d rounds of distributed coloring\n", distributedRounds);
         uint64_t totalVertsPerRound[numStatisticRecordingRounds];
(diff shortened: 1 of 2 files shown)


kddevin:
@ibogle  Would you please look at this PR?  Do you still need these print statements for running experiments for your dissertation?  We can accept the PR if you don't need the print statements.  If you do need them, we should #ifdef them out or reformat them in C++ so that they do not produce compiler warnings / errors.  Thanks for your response.

@djglaze Thanks for submitting this PR.

kddevin (commented):
I am waiting for a response from @ibogle.  Please do not merge until I hear back from him.  Thanks.

kddevin (changes requested):
Actually, these print statements are in code whose main purpose is to compute and report statistics.  Computing the statistics and then not reporting them doesn't make sense.  Let me take a stab at rewriting in C++ format to avoid the compiler warnings.

kddevin:
@djglaze  I am closing this PR without merging.  I have replaced it with PR #9769.  Please let me know whether it meets your needs.  Thanks!

@ibogle no need to review.

PR #9769, opened 2021-10-01
Zoltan2:  remove compiler warnings by using std::cout<< instead of printf

kddevin:
@trilinos/zoltan2

## Motivation

#9615 reports compilation warnings due to mismatched printf formats and arguments.
This PR replaces printf with std::cout<<, allowing C++ to determine the print format.

This PR replaces PR #9764; rather than remove the warning-generating print statements, this PR replaces them with equivalent C++ statements.

## Stakeholder Feedback

@djglaze @tasmith4 

## Testing

Tested on ascicgpu with serial node, sems gcc compilers.
Tested with GO = int and GO = long long.


djglaze (approved):
Ah, okay.  I didn't know that these were supposed to be here.  This looks great.  Thanks!

kddevin:
Thanks, @djglaze .  Let me know whether or not these changes resolve the STK compiler problems.

Issue #746, opened 2016-10-28
override warnings in panzer


bathmatt:
@rppawlo @eric-c-cyr Do we want to fix these by adding overrides to all the model evaluator calls

In file included from /Users/mbetten/Trilinos/mini-PIC/src/main.cpp:69:
/Users/mbetten/installs/install-for-panzer/clang-DEBUG/include/Panzer_ModelEvaluator.hpp:131:68: warning: 'get_W_factory' overrides a member
      function but is not marked 'override' [-Winconsistent-missing-override]
  Teuchos::RCP<const Thyra::LinearOpWithSolveFactoryBase<Scalar> > get_W_factory() const;

jmgate:
Did a quick search on this warning and it sounds like <a href="clang.llvm.org/extra/clang-tidy">clang-tidy</a> might be of some help in automating the fix, though I've never used it.

rppawlo:
We should definitely do this. What compiler version is giving this warning?

bathmatt:
clang on the mac,

s1000670:include mbetten$ clang++  --version

Apple LLVM version 7.3.0 (clang-703.0.31)

Target: x86_64-apple-darwin15.6.0

bathmatt:
So, I looked at this a bit, it seems that we can automate the fixing with
clang-tidy.  There are some issues, but we need to make a cmake output file
and use it as input to clang-tidy.  I don't want to have everyone looking
into this, so should we assign it to one of us>

bathmatt:
ok, it looks like it's sorta working, let me fix this

On Mon, Oct 31, 2016 at 10:16 AM, Matt Bettencourt <dr.bettencourt@gmail.com
> wrote:

> So, I looked at this a bit, it seems that we can automate the fixing with
> clang-tidy.  There are some issues, but we need to make a cmake output file
> and use it as input to clang-tidy.  I don't want to have everyone looking
> into this, so should we assign it to one of us>
>
> On Mon, Oct 31, 2016 at 9:14 AM, Matt Bettencourt <

bathmatt:
So, the auto-fix doesn't seem to do its thing for override.  Not sure why.


dir-evaluators (clang) 119 $ clang-tidy -fix-errors
Panzer_Integrator_BasisTimesVector.cpp
-checks='clang-analyzer-*,-clang-analyzer-alpha*,cppcoreguidelines-*,modernize-use-override'
-p ~/builds/build_for_panzer/clang/tidy/

8 warnings generated.

[...]

PR #9777, opened 2021-10-04
Tempus:  Zero out initial guess for adjoint initial conditions solve

etphipp:
This avoids NaNs in debug builds for linear solvers that don't zero out
or overwrite the initial guess.  Thanks to @ikalash for figuring this
out.  For Albany issue [#746](https://github.com/sandialabs/Albany/issues/746).


ccober6 (approved):
LGTM!

ikalash:
Thanks @etphipp !  I am testing Albany right now.  Will report back if there are any problems.

PR #9778, opened 2021-10-05
Tpetra: (WIP) change matrix modification logic

### LINKED ISSUE #9726: Tpetra: change logic in Tpetra::CrsMatrix::CombineGlobalValues
## Enhancement
@trilinos/tpetra 

At the 2021-Sept-21 Tpetra developer meeting, it was decided to modify the logic in Tpetra::CrsMatrix::CombineGlobalValues().  Instead of depending on

```
if (isStaticGraph ()) {
...
} else {
...
}
```

the appropriate method, e.g., `sumIntoGlobalValues`, should be unconditionally called, and its return value checked to determine success or failure.

This is to resolve #9720.

jhux2:
## Do not merge yet.

@trilinos/tpetra @csiefer2 

## Motivation
The logic for CRS matrix row insertion was based on the notion of static/dynamic graph.  

## Related Issues

Closes #9726.

## Testing
I'm relying on existing unit tests in Tpetra, Ifpack2, and MueLu.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -1567,7 +1567,6 @@ namespace Ifpack2 {
       IFPACK2_BLOCKTRIDICONTAINER_TIMER("BlockTriDi::SymbolicPhase");
 
       using impl_type = ImplType<MatrixType>;
-      using node_memory_space = typename impl_type::node_memory_space;
       using host_execution_space = typename impl_type::host_execution_space;
 
       using local_ordinal_type = typename impl_type::local_ordinal_type;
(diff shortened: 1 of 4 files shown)


kyungjoo-kim (approved):
Looks fine to me. Just a curiosity. Why do we return invalid return code from replaceGlobalValues instead of throwing the exception inside of the function ? If we cannot expect proper error handling with the invalid return value, it might be better to throw an exception in the replaceGlobalValues function instead of requiring the caller function check its return value and throw an exception.

william76:
@jhux2 this had passed testing but went stale because it was never merged.

Is this still a valid PR?

jhux2:
> @jhux2 this had passed testing but went stale because it was never merged.
> 
> Is this still a valid PR?

@william76 Yep, this is still valid.

PR #9779, opened 2021-10-05
TrilinosCouplings: Adding support for rotated diffusion

csiefer2:
Extending example


cgcgcg:
Would it make sense to run this in a test?

csiefer2:
> 
> 
> Would it make sense to run this in a test?

Sure it would.  But I don't see this as something we necessarily need to test regularly.

cgcgcg (approved):

PR #9784, opened 2021-10-06
adding "start owned" parameter to FECrsMatrix to start

### LINKED ISSUE #9638: Tpetra: FECrsMatrix ctor sets active=owned+shared, and fillState=closed, regardless of graph status.
## Enhancement
@trilinos/tpetra 

When building a FECrsMatrix from a FE graph, the active matrix is set to be the owned+shared one, and the fillState is set to closed. I find this confusing and possibly wrong, for two reasons:

1. fillState should match the active matrix. So if active=owned+shared, fillState should be "open", otherwise one between "modify" and "closed". 
2. if one builds a FE mat from a pre-assembled graph, they might not need an assembly phase. Creating the matrix in owned+shared mode, even if the input graph is closed, forces to do a pointless global assembly.

Additionally, the matrix might be filled inside a routine, which likely will start with `beginAssembly` or `beginModify` (depending on the type of filling). The routine should work the same, regardless of whether the matrix was newly created, or simply reused.

To sum it up, I would expect the FE matrix to satisfy the following:

- if input graph in ctor is closed, active=owned, and fillState=closed
- if input graph in ctor is open, active=owned+shared, and fillState=open
- at any given time, if active=owned+shared, fillState=open, and if active=owned, fillState=closed or fillState=modify

tjfulle:
closes #9638 

@bartgol how is this for a way to start the matrix in "owned" mode?

```c++
params->set<bool>("start owned", true);
 FECrsMatrix m(graph, params);
```


packages/tpetra/core/src/Tpetra_FECrsMatrix_def.hpp
```diff
@@ -71,7 +71,17 @@ FECrsMatrix(const Teuchos::RCP<const fe_crs_graph_type>& graph,
      ( *graph->activeCrsGraph_!= fe_crs_graph_type::FE_ACTIVE_OWNED,std::runtime_error,
       "Input graph must be in FE_ACTIVE_OWNED mode when this constructor is called.");
 
-  activeCrsMatrix_     = Teuchos::rcp(new FEWhichActive(FE_ACTIVE_OWNED_PLUS_SHARED));
+  bool start_owned = false;
+  if (! params.is_null ()) {
+    if (params->isParameter ("start owned")) {
```
bartgol, line 76: Minor, but, since you are willing to set `start_owned` as default (see line below), then you might as well not check whether `start owned` is a parameter or not...

bartgol (approved):
Looks good. I had one nano-comment, which is likely just subjective taste. Feel free to ignore.

PR #9786, opened 2021-10-07
Panzer: Make sure that panzer::ScatterResidual_BlockedTpetra checks for LOCir_GlobalEvaluationData for Residual evaluation.

crockatt:
@trilinos/panzer @rppawlo 

## Motivation

Incorporate additional checks to support response vectors into ScatterResidual_BlockedTpetra for the Residual evaluation type. The Jacobian evaluation already does this, and the non-blocked version also does this for both Residual and Jacobian evaluations currently.


rppawlo (approved):
Thanks @crockatt !

william76:
@crockatt In future PR's to Trilinos please prefix the title of your PR's with the `package_name:`.  It looks like this one is a panzer focused one so I went ahead and added the prefix :)

crockatt:
@rppawlo Do you know what went wrong here? It looks like there was some failure of the testing system itself? Is there a way to make it redo the tests?

rppawlo:
@crockatt - just added the retest label

Issue #9788, opened 2021-10-07
Ifpack: Compile Time Error with ifpack-hypre interface


keitat:
## Bug Report
@trilinos/ifpack 

### Description
After the change in the data type definitions in Hyper package at LLNL, we see compile time error at Ifpack's hype interface.   Ifpack is a 1st-gen package, and I believe that nobody is using this capability/interface.  Should we remove this or still change the source to support the new versions of Hypre?  

### Steps to Reproduce
See https://github.com/xsdk-project/xsdk-issues/issues/157#issuecomment-938003490

cgcgcg:
I think we can just comment out that particular call.

csiefer2:
Not in Trilinios, you can't.  We use that.

cgcgcg:
Should be fixed. Please reopen if this is still an issue.

PR #9792, opened 2021-10-07
Ifpack: Comment out deprecated Hypre function call

cgcgcg:
@trilinos/ifpack 

## Motivation
Issue #9788


keitat:
> @keitat Is this the only change? If so, we could simply issue another PR to merge this into the release branch.

It is only the error reported by Satish (xsdk-0.7.0 build)

PR #9794, opened 2021-10-08
Epetra/Tpetra/Zoltan: Some small Hash and Sort improvements

bavier:
@trilinos/epetra
@trilinos/tpetra
@trilinos/zoltan

## Motivation

This PR attempts to upstream some small performance improvements to Epetra and Tpetra sort utilities and a hash-map in Zoltan.

## Testing
The routines this PR touches are exercised by many of the unit tests of the affected packages.

### DIFF
--- a/packages/epetra/src/Epetra_Util.cpp
+++ b/packages/epetra/src/Epetra_Util.cpp
@@ -120,9 +120,13 @@ void Epetra_Util::Sort(bool SortAscending, int NumKeys, T * Keys,
 
   int n = NumKeys;
   T * const list = Keys;
-  int m = n/2;
+  static int ciura[] = {1, 4, 10, 23, 57, 152, 301, 701, 1577, 3548, 7983,
+                        17961, 40412, 90927, 204585, 460316, 1035711, 2330349,
+                        5243285, 11797391, 26544129, 59724290, 134379652};
+  int c = sizeof(ciura) / sizeof(int);
 
-  while (m > 0) {
+  while (c-- > 0) {
+    int m = ciura[c];
     int max = n - m;
     for (int j=0; j<max; j++)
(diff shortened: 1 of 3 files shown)


jhux2:
@trilinos/epetra @trilinos/zoltan @trilinos/tpetra

kddevin (changes requested):
@bavier  Thanks for the contributions.
Can you give us some performance numbers involving these changes?   
I can find Ciura's gap sequence up to 1750 online; how did you generate the rest of the numbers?
What happens if someone tries to sort 135M values.  While doing so may not be advised with shell sort, there are no guards to prevent it in the new implementation.
Is there a better starting value for c?  Many uses of sort2 are sorting short lists, e.g., local column indices for a given row.

bavier:
@kddevin This excerpt from one of the commit messages:

> E.g. trilinoscouplings/examples/scaling/Poisson.cpp run with -n128
-N16 and a 200x200x200 mesh with scotch-based load-balancing went from
95.7s runtime to 86.9s (6% speedup overall, 8.8% spent in
Epetra_Util::Sort to 5.2%).

As far as how the sequence was derived: at the time I developed this patch (at least 7 years ago) I believe I had a performance test harness where for those past 2000 or so I tested several values in the area around what a linear recurrence would produce.  But I can't find that test harness, and I don't recall how much having the larger gap numbers made on the final performance of very large arrays.  Ultimately, I recall that any gap sequence was far better than the `n/2` sequence, but I would need to do another analysis to convince myself again that the larger gap elements were truly beneficial.

Each gap value is used as a the stride for a sublist sort.  There's no "greatest length" that can be handled.  For shorter lists, the larger gap values are skipped via the `j<max` condition.

PR #9828, opened 2021-10-19
MueLu: Simplify passing precomputed objects

cgcgcg:
@trilinos/muelu 

## Motivation
- Allow to pass filenames in the "user data".
- Set generating factory for matrices P, R, etc when read from "user data".
- Be a bit more verbose when an object is not available on a level.
- Add option to write out solution to MueLu driver.

### DIFF
--- a/packages/muelu/test/interface/default/Output/operator_solve_5_np1_epetra.gold
+++ b/packages/muelu/test/interface/default/Output/operator_solve_5_np1_epetra.gold
@@ -6,31 +6,6 @@ smoother ->
  relaxation: sweeps = 1   [unused]
  relaxation: damping factor = 1   [unused]
 Level 1
-Prolongator smoothing (MueLu::SaPFactory)
-Matrix filtering (MueLu::FilteredAFactory)
-Build (MueLu::CoalesceDropFactory)
-Build (MueLu::AmalgamationFactory)
-[empty list]
-algorithm = "classical" classical algorithm = "default": threshold = 0, blocksize = 1
-lightweight wrap = 1
-Filtered matrix is not being constructed as no filtering is being done
-Build (MueLu::TentativePFactory)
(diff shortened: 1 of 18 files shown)


csiefer2 (approved):

packages/muelu/src/MueCentral/MueLu_HierarchyUtils_def.hpp
```diff
@@ -98,15 +98,52 @@ namespace MueLu {
...
+              // We might also want to read maps here.
+              mat = Xpetra::IO<Scalar,LocalOrdinal,GlobalOrdinal,Node>::Read(Teuchos::getValue<std::string>(levelListEntry->second), lib, comm);
+            else
+              mat = Teuchos::getValue<RCP<Matrix > > (levelListEntry->second);
+            level->Set(name, mat, NoFactory::get());
             M->SetFactory(name, NoFactory::getRCP()); // TAW: not sure about this: be aware that this affects all levels
                                                       //      However, A is accessible through NoFactory anyway, so it should
                                                       //      be fine here.
           }
           else if(name == "P" || name == "R" || name == "K" || name == "M") {
```
jhux2, line 133: What are "K" and "M"?

cgcgcg: I'm assuming something like mass and stiffness, but no idea where in MueLu that's actually used.

packages/muelu/src/MueCentral/MueLu_HierarchyUtils_def.hpp
```diff
@@ -282,6 +337,7 @@ namespace MueLu {
               throw std::runtime_error("Invalid non-serializable data on list");
           }
         }
+        // level->print(std::cout, MueLu::Debug);
```
mayrmt, line 340:
```suggestion
```

PR #9851, opened 2021-10-22
MueLu: update absrowsum option

jhux2:
Update the absrowsum capability. The tolerance that was passed in was immediately being set to zero, which is an obvious bug. I've also added an option (suggested by @rstumin) that sets the tolerance to a scaled average of the absolute values of the (local) diagonal.

~~This is marked WIP until there's a unit test.~~ Done.

@trilinos/<teamName>


jhux2:
> I'm not sure what is going on ... but I just tried this branch on the spe10 problem and the results are pretty horrible. Basically, I'm not getting convergence with this version of absrowsum. When I look at the AMG hierarchy summary, it is definitely different from what we had before ... and some of the eigenvalue estimates have moved around ... but it doesn't look horrible ... except when you run gmres ... and it does not converge.

@rstumin I've made the changes you requested ... with the exception of diagVals[i] == 1.  The spe5 system is not converging for me, but at least we can see if this code compiles cleanly everywhere.

rstumin:
@jhux2 Did you verify that the true false stuff is being set properly for the automatic diagonal avg case in SaPFactory and the Ifpack2 smoother code?

rstumin:
@jhux2 Can you summarize the status? Have you checked that all parameters are actually making it into the appropriate factories ... via Get?

rstumin (approved):
looks good. I've also run spe10 tests, which give reasonable numbers

PR #9897, opened 2021-11-03
Belos: fix issue with MultiVector modify flags

brian-kelley:
Belos's diagonal matrix apply was using Tpetra's ``MultiVector::getLocalView<device_type>(...)``
but this was incorrectly setting the "modified on host" flag when the device is
Cuda/CudaUVMSpace. The resulting view's execution space is Cuda, so it should definitely be considered a device view even though (being UVM) it's accessible from host. Instead, using ``MultiVector::getLocalViewDevice(...)`` correctly sets the
"modified on device" flag. This fixes 2 Belos test failures for when UVM
enabled in Tpetra, but this doesn't fix the behavior of ``getLocalView<device_type>`` for UVM - that can be a separate issue.

@trilinos/tpetra
@trilinos/belos 

## Motivation

In a build with ``Tpetra_ENABLE_CUDA_UVM=ON``, these two tests were failing:
```
Belos_Tpetra_Native_CG_PIPELINE_diagonal_matrix_MPI_4
Belos_Tpetra_Native_CG_SINGLE_REDUCE_diagonal_matrix_MPI_4
```

## Testing

This was really an issue in the test itself, not in tested code.

### DIFF
--- a/packages/belos/tpetra/test/Native/diagonal_matrix.cpp
+++ b/packages/belos/tpetra/test/Native/diagonal_matrix.cpp
@@ -109,9 +109,9 @@ public:
       auto X_j = X.getVector (j);
       auto Y_j = Y.getVectorNonConst (j);
 
-      auto X_j_lcl_2d = X_j->template getLocalView<device_type> (Tpetra::Access::ReadOnly);
+      auto X_j_lcl_2d = X_j->getLocalViewDevice (Tpetra::Access::ReadOnly);
       auto X_j_lcl = Kokkos::subview (X_j_lcl_2d, Kokkos::ALL (), 0);
-      auto Y_j_lcl_2d = Y_j->template getLocalView<device_type> (Tpetra::Access::ReadWrite);
+      auto Y_j_lcl_2d = Y_j->getLocalViewDevice (Tpetra::Access::ReadWrite);
       auto Y_j_lcl = Kokkos::subview (Y_j_lcl_2d, Kokkos::ALL (), 0);
 
       KokkosBlas::mult (static_cast<ISC> (beta), Y_j_lcl,


iyamazaki (approved):
Very nice catch. It looks good to me. Thank you, @brian-kelley. Any idea why this do not show up in any other tests?

kddevin (approved):
Good catch, @brian-kelley .  Subtle.  Thank you

brian-kelley:
@iyamazaki When we did the UVM removal, we generally preferred to use ``getLocalViewHost``/``getLocalViewDevice`` instead of the templated ``getLocalView<device_t>``. I think that's the only reason.

I do have a new, simple test in Tpetra that replicates this issue. I think it's going to be a simple long-term fix - from this code in WrappedDualView, it's clear why this was happening:

```
  template<class TargetDeviceType>
  typename std::remove_reference<decltype(std::declval<DualViewType>().template view<TargetDeviceType>())>::type::const_type
  getView (Access::ReadOnlyStruct s DEBUG_UVM_REMOVAL_ARGUMENT) const {
    bool returnDevice = std::is_same<
    {
      auto tmp = dualView.template view<TargetDeviceType>();
      if (tmp == this->dualView.view_host()) returnDevice = false;
    }
...
```
So for UVM, the host and device views are identical meaning this ``tmp == this->dualView.view_host()`` is always going to be true, and ``returnDevice`` is going to be false. That is what determines which modify flag gets set.

iyamazaki:
Very nice, @brian-kelley. I was actually wondering if you now have any idea why, e.g., `Belos_Tpetra_Native_CG_PIPELINE_diagonal_matrix_MPI_4` was passing in the PR tests (with UVM and CUDA_BLOCKING OFF)?

PR #9902, opened 2021-11-05
Tpetra: DistObject::transferArrived()

### LINKED ISSUE #9251: Tpetra:  non-blocking Import/Export

tasmith4:
@trilinos/tpetra @vbrunini 

## Motivation
Provides a transferArrived function for DistObject, that will return true once the data from transfer started by beginImport etc. has been communicated appropriately and is ready for the unpack and combine step performed by endImport etc.  This completes support for the API specified in #9251.

## Related Issues

* Closes #9251

## Testing
This PR includes unit tests for the new transferArrived function.  There are no specific tests that transferArrived returns false for CrsMatrix when beginImport is only called on one rank, because CrsMatrix uses the "variable number of packets per LID" code path, which requires a blocking communication across all ranks to establish the number of packets each rank should expect to receive.  However, the conditions under which transferArrived returns false are still tested through the MultiVector.

The test "MultiVectorGroup" illustrates the use case where a user simultaneously kicks off several import operations on multiple MultiVectors and allows them to complete asynchronously.

### DIFF
--- a/packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
+++ b/packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
@@ -833,10 +833,13 @@ namespace {
     { }
 
     void operator()(DistObjectRCP source, DistObjectRCP target) const {
-      Import<LO, GO> importer(source->getMap(), target->getMap());
-      int myRank = target->getMap()->getComm()->getRank();
+      auto comm = target->getMap()->getComm();
+      int myRank = comm->getRank();
+      int numProcs = comm->getSize();
 
-      if (myRank == 0) {
+      Import<LO, GO> importer(source->getMap(), target->getMap());
+
+      if (numProcs > 1 && myRank == 0) {
         target->beginImport(*source, importer, INSERT);
         TEST_ASSERT(!target->transferArrived());
       }
@@ -868,10 +871,13 @@ namespace {
     { }
 
     void operator()(DistObjectRCP source, DistObjectRCP target) const {
-      Export<LO, GO> exporter(source->getMap(), target->getMap());
-      int myRank = target->getMap()->getComm()->getRank();
+      auto comm = target->getMap()->getComm();
+      int myRank = comm->getRank();
+      int numProcs = comm->getSize();
 
-      if (myRank == 0) {
+      Export<LO, GO> exporter(source->getMap(), target->getMap());
+
+      if (numProcs > 1 && myRank == 0) {
         target->beginExport(*source, exporter, INSERT);
         TEST_ASSERT(!target->transferArrived());
       }
--- a/packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
+++ b/packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
@@ -743,9 +743,11 @@ namespace {
 
     void operator()(DistObjectRCP source, DistObjectRCP target) const {
       Import<LO, GO> importer(source->getMap(), target->getMap());
+
       target->beginImport(*source, importer, INSERT);
       while (!target->transferArrived());
       target->endImport(*source, importer, INSERT);
+
       TEST_ASSERT(target->transferArrived());
     }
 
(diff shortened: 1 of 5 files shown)


packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
```diff
@@ -693,6 +780,246 @@ namespace {
...
+    CheckTransferArrivedForwardImport(FancyOStream& o, bool& s)
+      : out(o),
+        success(s)
+    { }
+
+    void operator()(DistObjectRCP source, DistObjectRCP target) const {
+      Import<LO, GO> importer(source->getMap(), target->getMap());
+
+      target->beginImport(*source, importer, INSERT);
+      while (!target->transferArrived());
```
kddevin, line 835: Is this while loop calling transferArrived needed? 
Is the intent of this test to determine whether transferArrived is true after endImport/endExport?

packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
```diff
@@ -693,6 +780,246 @@ namespace {
...
+    void operator()(DistObjectRCP source, DistObjectRCP target) const {
+      TEST_ASSERT(target->transferArrived());
+    }
+
+  private:
+    FancyOStream& out;
+    bool& success;
+  };
+
+  TEUCHOS_UNIT_TEST_TEMPLATE_3_DECL( TransferArrived, MultiVector_defaultTrue, Scalar, LO, GO )
```
kddevin, line 803: For each test, having a one-sentence comment describing the feature being exercised would be helpful.  I spent a bit of time trying to figure out why this test didn't actually do an import/export, before realizing that it was just checking that transferArrived is true when an import/export hasn't been initiated.    So for all tests, a comment like the following would be helpful:
```
// Test that transferArrived is true when no multivector transfer has been initiated 
```
or 
```
// Test that multivector transfer works correctly with target multivector completely on rank 0
```

packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
```diff
@@ -693,6 +780,246 @@ namespace {
...
+    CheckTransferArrivedForwardExport(FancyOStream& o, bool& s)
+      : out(o),
+        success(s)
+    { }
+
+    void operator()(DistObjectRCP source, DistObjectRCP target) const {
+      Export<LO, GO> exporter(source->getMap(), target->getMap());
+
+      target->beginExport(*source, exporter, INSERT);
+      while (!target->transferArrived());
```
kddevin, line 880: ditto

packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
```diff
@@ -693,6 +780,246 @@ namespace {
...
+        targets[i]->beginImport(*sources[i], importer, INSERT);
+      }
+
+      unsigned completedImports = 0;
+      std::vector<bool> completedImport(false, sources.size());
+      while (completedImports < completedImport.size()) {
+        for (unsigned i=0; i<sources.size(); i++) {
+          if (completedImport[i]) continue;
+          if (targets[i]->transferArrived()) {
+            targets[i]->endImport(*sources[i], importer, INSERT);
```
kddevin, line 1005: Use case for @vbrunini -- nice!

kddevin (approved):

packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
tasmith4: I tend to view a need for comments as an indication that I've failed to write code that's clear enough to stand on its own (as is obviously true in this case) -- so my first preference is to rewrite the code to be more clear and obvious about its intent.  Is there another name for this functor or for the test/fixtures that would make the intent of the test more clear to you?  Or should I just admit failure and add the comments?

packages/tpetra/core/test/ImportExport2/AsyncTransfer_UnitTests.cpp
tasmith4: it's not strictly needed for the workflow -- the endImport will wait until everything has arrived anyway, and the main point of the test is that transferArrived is still true after endImport/Export.

The loop does represent a subtle assert that transferArrived does in fact return true at some point after beginImport and before endImport.  i.e., we want to TEST_ASSERT(target->transferArrived()) after the begin import (kind of the opposite check as the one in the "transfer not arrived" test), but we can't call the assert directly because we don't know exactly when things arrived.  So we loop as a way of checking that it does at some point become true.

Since there's not an explicit assert here, the fact that this loop actually represents an assert is probably too subtle, but I wasn't sure how else to do it.  I'm also open to an argument for just dropping it if it's too confusing.  Thoughts?

tasmith4:
Thanks for the helpful review @kddevin.  I've entered some responses to your comments, once we agree on the best path forward I'll get these changes in.

kddevin (dismissed):

kddevin (approved):

PR #9904, opened 2021-11-05
Tpetra: remove obsolete configure option

tasmith4:
@trilinos/tpetra 

## Motivation
During the UVM removal process, a configure option was added to allow developers to build Tpetra without UVM, even though it (UVM-off) was not guaranteed to work.  Now that it works, this option is obsolete and redundant -- users and developers can build with/without UVM by changing the value of Kokkos_ENABLE_CUDA_UVM.

## Testing
N/A -- this is a build system change.


jhux2 (approved):
lgtm

PR #9914, opened 2021-11-10
MueLu: Enable Hiptmair translation from ML

cgcgcg:
@trilinos/muelu 

## Motivation
- Read Hiptmair settings from ML parameter lists.
- In MueLu RefMaxwell, use the smoother factory to create Hiptmair.


GrahamBenHarper (approved):
Looks good to me. Are there any input decks that need to be updated to set the appropriate values for hiptmair?

cgcgcg:
@GrahamBenHarper I don't think so, but I guess the AT will tell us.

GrahamBenHarper:
If the AT is alive :eyes:

PR #9917, opened 2021-11-10
Sacado local deep copy

etphipp:
A Sacado customer requested the ability to use `Kokkos::Experimental::local_deep_copy()` with View's holding Fad scalar types.  This requires overloading two forms of `local_deep_copy_contiguous()` that don't work for Sacado view's because `value_type`  (a.k.a., a Fad) is not assignable to `*pointer_type` (a.k.a. a double).  All of the other cases do appear to work however.  This PR adds the necessary overloads, and also changes the signature of the Kokkos versions to only be valid for "normal" views where `specialize == void`.  The Kokkos changes will also have to be pushed to the Kokkos repo.


ndellingwood:
@etphipp for the kokkos changes we'd like to get those merged to the main repo first then we'll snapshot to Trilinos for the use case here, would you like me to submit a PR to kokkos or would you prefer to do so?

etphipp:
@ndellingwood If you don't mind doing it, that would be great.  Thanks!

ndellingwood:
@etphipp the kokkos changes are submitted kokkos/kokkos#4520, after that PR is reviewed+merged we'll cherry-pick to master and snapshot to Trilinos along with a couple other updates

etphipp:
Thanks @ndellingwood !

tmranse:
@etphipp I'm not sure the Kokkos change alone resolved this issue -- since there is no fallback for specialized view types, I still get an error saying that there is no available instantiation of `local_deep_copy_contiguous` . Or am I missing something here?

etphipp:
I think this PR got dropped while waiting for the needed changes in Kokkos.  @ndellingwood, did the required changes ever get put into Kokkos?

ndellingwood:
Looks like https://github.com/kokkos/kokkos/pull/4520 contained the changes and came into Trilinos with #9958 , so the Sacado changes should be good to pursue separately from the Kokkos changes in this PR

etphipp:
I updated the branch, fixed the merge conflicts, and got the code building again.  We'll see if it passes the autotester.

etphipp:
@ndellingwood would you mind reviewing this PR?

ndellingwood (commented):
Hi @etphipp , I looked over the changes and the added `local_deep_copy_contiguous` routines seem reasonable, but to make sure I understand the changes, can you comment on my summary here?

* The new unit tests call `local_deep_copy` (not `local_deep_copy_contiguous`), which does not require overloads in Sacado
* This PR adds overloads for `local_deep_copy_contiguous` for specialized (non-void) View types (i.e. Fad, FadContinguous, etc)
* The call to `local_deep_copy` is a routine in Kokkos that calls out to `local_deep_copy_contiguous` 
*  SFINAE helps the compiler select the appropriate overload, where the Sacado overloads provided here are the non-void-specialized counterparts to the void-specialized impls in Kokkos at https://github.com/kokkos/kokkos/blob/2f5723bd6c0717c301b217fba43f7c32864cb37b/core/src/Kokkos_CopyViews.hpp#L2160-L2179 

Let me know if I'm following, or please correct me so I understand properly

etphipp:
Yes, I think that is all correct.

ndellingwood (approved):

PR #9923, opened 2021-11-11
Tpetra: Xpetra: fix matrix reader

jhux2:
@trilinos/tpetra @trilinos/xpetra 
## Motivation
This fixes two bugs in matrix reading:
1) The Tpetra matrix reader was allocating space based on a scalar max nnz per row, which could cause an out-of-memory error if one of the rows is (nearly) dense.
2) The Xpetra binary reader was calculating the number of nonzeros per row incorrectly.

## Stakeholder Feedback

Used primarily by developers.

## Testing

Tested locally.

### DIFF
--- a/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
+++ b/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
@@ -3609,7 +3609,6 @@ namespace Tpetra {
         ArrayRCP<size_t> rowPtr;
         ArrayRCP<global_ordinal_type> colInd;
         ArrayRCP<scalar_type> values;
-        size_t maxNumEntriesPerRow = 0;
 
         // Proc 0 first merges duplicate entries, and then converts
         // the coordinate-format matrix data to CSR.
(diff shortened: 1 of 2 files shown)


cgcgcg:
@jhux2 I added a `ReadLocal` method that reads a different file on each MPI rank, and fixed some small issues in the binary, single file case.

jhux2:
@cgcgcg Cool.  So now a matrix can be read in binary either entirely on rank 0 and then exported, or on a per-rank basis?

cgcgcg:
@jhux2 Correct. And same for MatrixMarket.

jhux2:
> Correct. And same for MatrixMarket.

Excellent.  Lgtm!

cgcgcg (dismissed):

packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
```diff
@@ -3840,18 +3840,36 @@ namespace Tpetra {
           values = null;
         } // if myRank == rootRank
 
+        //FIXME this is no longer needed
         broadcast<int,size_t> (*pComm, 0, &maxNumEntriesPerRow);
```
mayrmt, line 3844: The comment says, this is not need anymore. Can it be removed?

mayrmt (commented):
Any chance, that one could include the new binary reader into testing?

packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
jhux2: Fixed with commit 4a60479.

mayrmt (approved):

PR #9929, opened 2021-11-14
Ifpack2: rebaselining long double test

ikalash:
Something in Trilinos changed that changed slightly the matrices / vectors defined in this problem.  The problem signed up in the ST = long double build in the Trilinos tests: https://sems-cdash-son.sandia.gov/cdash/test/1838303.  Verified that precision of result given new matrix/vector is correct.


ikalash:
@trilinos/framework : it looks like there are some test failures in the PR testing that have nothing to do with this PR 

https://testing.sandia.gov/cdash/viewBuildError.php?buildid=10240241 

What should I do in this case?

csiefer2 (approved):

ikalash:
@trilinos/framework : I don't understand why this PR is not going through.  Is the autotester still broken?

PR #9935, opened 2021-11-16
Piro: correcting logic

ikalash:
This PR corrects some logic that was incorrectly attempting to create an adjoint ME even when the adjoint ME passed in (from e.g., Albany) was null.  

@lxmota , can you please review, since this affects LCM?


lxmota (approved):
Looks good to me

PR #9941, opened 2021-11-17
MueLu: convert nullspace dimension error to warning

mayrmt:
@trilinos/muelu 

## Motivation

In many cases, the dimension of the nullspace is >= the number of unknonws per fine level mesh node, e.g. for solid mechanics of Navier-Stokes. However, there are scenarios, where the dimension of the nullspace can be smaller than the number of unknowns per fine level node, e.g. for beam finite elements (I'll mention a particular example below for illustration.) To allow MueLu to be used for such structural
models, we must be able to run this wihtout throwing an error.

As it still makes sense to perform this check for standard cases, the
check of the nullspace dimension stays in place, but the user can
suppress it by providing a particular option in the `HierarchyManager`.

By default, we still perform the check to help users to get their
nullspace correct.

**Beam FEM example:**

For torsion-free Kirchhoff-Love beam formulations, one ends up with 6 unkonwns per node (3 displacements, 3 "changes" in the tangent vector), but only 5 nullspace vectors (3 translations, 2 rotations around axes orthogonal to the beam centerline). A third rotation is missing since these beams do not possess a rigid body mode for the rotation around their own axis/centerline.

**Minor changes:**

- move some comment, such that they are now processed by Doxygen
- remove trailing whitespaces in various places

## Stakeholder Feedback

n/a

## Testing

Existing tests pass.

When suppressing the check, this is the new warning message in action:
```
******* WARNING *******
User-provided nullspace has fewer vectors (5) than number of PDE equations (6). It depends on the PDE, if this is a problem or not.
```

## Additional Information

I have discussed this offline with @rstumin and we both believe that this change is in line with multigrid theory. Since it depends on the type of PDE to be solved how the number of nullspace vectors and number of unknowns per fine level mesh node are related to each other, we came to the conclusion that a warning instead of an error is reasonable to also raise awareness about this phenomenon.

/cc @Max-pert


jhux2:
@mayrmt It seems that the scenario where the nullspace dimension is legitimately smaller than the number of DOFs is (much?) less common.

What do you think about, for example, an override option that the user would have to explicitly set to avoid this error?

mayrmt:
@jhux2 I like this idea! Maybe something like "nullspace: override dimension check"?

jhux2:
> I like this idea! Maybe something like "nullspace: override dimension check"?

Is the option primarily for smoothed aggregation?  If so, what do you think about "sa: override nullspace dimension check"?  I only say this because there doesn't seem to be any other options starting with "nullspace:".

I don't feel strongly btw, so whatever you decide is fine by me!

rstumin:
@mayrmt @jhux2 By the way  .... my current PR does add an option that begins with "nullspace: " ...

mayrmt:
Marking as "AT: WIP" until I find time for further changes.

packages/muelu/src/Interface/MueLu_HierarchyManager.hpp
```diff
@@ -86,20 +86,20 @@ namespace MueLu {
...
+    //! Constructor
     HierarchyManager(int numDesiredLevel = MasterList::getDefault<int>("max levels")) :
-        numDesiredLevel_        (numDesiredLevel),
-        maxCoarseSize_          (MasterList::getDefault<int>("coarse: max size")),
-        verbosity_              (Medium),
-        doPRrebalance_          (MasterList::getDefault<bool>("repartition: rebalance P and R")),
-        implicitTranspose_      (MasterList::getDefault<bool>("transpose: use implicit")),
-        fuseProlongationAndUpdate_ (MasterList::getDefault<bool>("fuse prolongation and update")),
-        sizeOfMultiVectors_     (MasterList::getDefault<int>("number of vectors")),
-        // -2 = no output, -1 = all levels
```
cgcgcg, line 98: What was wrong with that comment?

mayrmt: It's not wrong, but I have move the documentation to the definition of the class member. Then, it can be picked up by Doxygen.

cgcgcg (approved):

PR #9945, opened 2021-11-17
Tpetra: fix modify flags of WDV::getView<T>

brian-kelley:
Use compile-time logic to figure out whether WrappedDualView::getView<T>
will return a host or device view. Add test to make sure the correct
modify flags based on the device type T. #9897 fixed one example of incorrect modify flags, but this is a permanent fix for the general issue.

The general issue was, for a MultiVector with the device and host being the same (e.g. UVM), doing
``mv.template getLocalView<DeviceType>(ReadWrite)`` would set the host modify flag. But now, plugging in the device consistently sets the device modify flag.

@trilinos/tpetra 

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows #9897
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing

Added a test in MultiVector/LocalViewTests to replicate the previous issue and make sure the modify flags are set correctly.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp
@@ -146,8 +146,8 @@ public:
 
   // This is an expert-only constructor
   // For WrappedDualView to manage synchronizations correctly,
-  // it must have an DualView which is not a subview to due the 
-  // sync's on.  This is what origDualV is for.  In this case, 
+  // it must have an DualView which is not a subview to due the
+  // sync's on.  This is what origDualV is for.  In this case,
   // dualV is a subview of origDualV.
   WrappedDualView(DualViewType dualV,DualViewType origDualV)
     : originalDualView(origDualV),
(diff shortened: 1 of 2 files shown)


kddevin (dismissed):

brian-kelley:
@kddevin Could you have another look? I had to change one of the tests for when both OpenMP and Serial are enabled. Because of the way mirror views are defined, you can end up with <Serial,HostSpace> being the 'device' type and <OpenMP,HostSpace> being the 'host' type. When you ask for a 'host' view using getView<T>, it ends up setting the device view as modified. I believe this is a Kokkos issue, but this PR's changes are still valid.

brian-kelley:
@kddevin OK, this is ready for a new review. I tested it on Cuda+UVM, Serial+OpenMP (which was failing autotester before), and HIP.

kddevin (approved):
Thanks, @brian-kelley

william76:
@brian-kelley @kddevin @csiefer2  
FYI, it looks like this passed its tests a couple days ago :) 
Automerge wasn't enabled so maybe you're waiting on a final check or timing to merge... but just in case I thought I'd send ya a FYI 😀

PR #9948, opened 2021-11-17
Tpetra: Update Tpetra_TestBlockCrs

tcclevenger:
Make updates to Tpetra_TestBlockCrs example.

Changes:
- Fix Host call of getLocalBlockDeviceNonConst() to getLocalBlockHostNonConst()
- Add assert to ensure BlockCrs and Crs matvec yield same results (up to a tolerance)

@trilinos/tpetra 

## Motivation

Found that BlockCrs::apply() gave a diff to Crs::aaply() when doing some benchmarking.

## Testing

Added an assert within the example which requires both BlockCrs and Crs apply to give the same results (up to a tol).


jhux2:
@tcclevenger This line looks a little strange to me:

https://github.com/trilinos/Trilinos/blob/6c69f9eeacef7487e14ad61f4983e80f273a13d8/packages/tpetra/core/example/BlockCrs/Tpetra_TestBlockCrs.cpp#L408

because of the `rowptr_host` on the RHS.

tcclevenger:
This fix doesn't work. The solution is being rethought, will open a new PR when it is found.

PR #9953, opened 2021-11-18
MueLu: calculate null space rotations capability

rstumin:
Auto-PR for SHA c9c7498


packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_NullspaceFactory_def.hpp
```diff
@@ -93,6 +98,13 @@ namespace MueLu {
     if (currentLevel.IsAvailable(nspName, NoFactory::get()) == false && currentLevel.GetLevelID() == 0)
       Input(currentLevel, "A");
 
+    if( currentLevel.GetLevelID() == 0 &&
+        currentLevel.IsAvailable("Coordinates", NoFactory::get()) &&     // we have coordinates (provided by user app)
+        pL.get<bool>("nullspace: calculate rotations") ) {               // and we want to calculate rotation modes
+      calculateRotations_ = true;
+      Input(currentLevel, "Coordinates");
```
csiefer2, line 105: Are we sure this Input declaration is correct?

csiefer2 (approved):

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_NullspaceFactory_def.hpp
rstumin: @csiefer2 it has definitely been tested. Is there something funny or strange about it that makes you concerned?

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_NullspaceFactory_def.hpp
```diff
@@ -114,6 +126,30 @@ namespace MueLu {
...
+
+      xvals  = Coords->getData(0);
+      cx     = Coords->getVector(0)->meanValue();
+      if (Coords->getNumVectors() > 1) {
+        yvals = Coords->getData(1);
+        cy    = Coords->getVector(1)->meanValue();
+      }
+      if (Coords->getNumVectors() > 2) {
+        zvals = Coords->getData(2);
+        cz    = Coords->getVector(1)->meanValue();
```
mayrmt, line 149: Index `1` refers to y-direction, so maybe this is correct for the z-direction:
```suggestion
        cz    = Coords->getVector(2)->meanValue();
```

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_NullspaceFactory_def.hpp
```diff
@@ -146,8 +190,47 @@ namespace MueLu {
...
+          // when creating ptent should take care of scaling issues ... but leaving it just in case.
+           if ( (int) nullspaceDim > numPDEs ) {
+            Teuchos::Array<typename Teuchos::ScalarTraits<Scalar>::magnitudeType> norms2(nullspaceDim);
+            nullspace->norm2(norms2);
+            Teuchos::Array<Scalar> norms2scalar(nullspaceDim);
+            for (int i = 0; i < nullspaceDim; i++)
+              norms2scalar[i] = norms2[0] / norms2[i];
+            nullspace->scale(norms2scalar);
+          }
+          */
```
mayrmt, line 230:
```suggestion
```

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_NullspaceFactory_def.hpp
```diff
@@ -178,9 +261,48 @@ namespace MueLu {
...
+        if ( (int) nullspaceDim > numPDEs ) {
+           /* xy rotation */
+           ArrayRCP<Scalar> nsValues = part->getDataNonConst(numPDEs);
+           int numBlocks = nsValues.size() / numPDEs;
+           TEUCHOS_TEST_FOR_EXCEPTION(numBlocks != (int) xvals.size(), Exceptions::RuntimeError, "MueLu::NullspaceFactory::fillNullspaceVector(): number of coordinates does not match  ndofs/numPDEs.");
+           for (int j=0; j< numBlocks; ++j) {
+             nsValues[j*numPDEs + 0] = -(yvals[j]-cy);
+             nsValues[j*numPDEs + 1] =  (xvals[j]-cx);
+           }
+         }
```
mayrmt, line 273: That looks like a duplication from the function `Build()`. What's the difference between this code and the one above? Can this duplicate be avoided by moving this to a private function, that is then called by both `Build()` and `fillNullspaceVector()`.

This comment also applies to the other two rotations.

packages/muelu/test/scaling/MatrixLoad.hpp
```diff
@@ -156,6 +156,26 @@ void MatrixLoad(Teuchos::RCP<const Teuchos::Comm<int> > &comm,  Xpetra::Underlyi
       Galeri::Xpetra::BuildProblem<SC,LO,GO,Map,CrsMatrixWrap,MultiVector>(galeriParameters.GetMatrixType(), map, galeriList);
     A = Pr->BuildMatrix();
     nullspace = Pr->BuildNullspace();
+    //  The coordinates used by Galeri here might not match the coordinates thta come into this function.
+    //  In particular, they might correspond to different strretch factors. To fix this, we overwrite the
```
mayrmt, line 160:
```suggestion
    //  The coordinates used by Galeri here might not match the coordinates that come into this function.
    //  In particular, they might correspond to different stretch factors. To fix this, we overwrite the 
```

mayrmt (changes requested):
@rstumin Requesting changes, as there seems to be an index problem, when computing the z-component of the centroid. Also some typos.

Regarding testing, the new nullspace computation is now executed by a test, but is the correctness of the result checked anywhere? This could/should be done via a unit test for the `NullspaceFactory`.

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_NullspaceFactory_def.hpp
rstumin: Hmmm ... that does seem messed up. It's funny how my own personal tests passed. I actually did one with the ice sheet code that compares with the null space computed by the application with the one using muelu. Here, I got identical residual histories ... but this test was only 2D so it didn't test the 3rd dimension. The second test was with the MueLu_Driver comparing again the null space computed by MueLu and the one computed by Galeri. Again, this test gives identical residual histories. There is a chance that the centroid calculation only improves the conditioning of the basis ... but not the actual residual histories if numerical conditioning of the basis is not an issue.

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_NullspaceFactory_def.hpp
rstumin: This redundancy was already in the code ... and I just continued it. It basically corresponds to whether or not we have a blockmap or a map. I'll look into whether this can be done in a single function.

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_NullspaceFactory_def.hpp
rstumin: The best test is to run the code twice .... with Galeri rotations and MueLu rotations ... and see that the residuals are identical. I wasn't sure if this was easy to do with the Driver? Do you know?

mayrmt: I don't know how to do this in the driver. However, by comparing the residual histories, you do not directly test the nullspace, only indirecltly.

For directly testing the correctness of the nullspace, you would need to add a unit test, e.g. in `packages/muelu/test/unit_tests/NullspaceFactory.cpp`. Unfortunately, there isn't a single unit test for the `NullspaceFactory` so far, so you would need to create the `packages/muelu/test/unit_tests/NullspaceFactory.cpp` file and add your test (and maybe some other tests as well). There, you could compare the generated nullspace to a hand-coded nullspace, that you know to be correct.

rstumin:
Added unit test and resolved issues

rstumin:
@mayrmt Matthias, are things ok now. I tried to resolve most of the issues that you raised ... including a hand computed nullspace unit test

packages/muelu/test/unit_tests/NullspaceFactory.cpp
```diff
@@ -0,0 +1,149 @@
...
+    Teuchos::ParameterList nspaceList = *(nspace->GetValidParameterList());
+    nspaceList.set("nullspace: calculate rotations", true);
+    nspace->SetParameterList(nspaceList);
+    fineLevel.Request(*nspace);
+    nspace->Build(fineLevel);
+    RCP<MultiVector> nullSpace = fineLevel.Get<RCP<MultiVector> >("Nullspace",nspace.get());
+
+    ArrayRCP<real_type> xvals = coordinates->getDataNonConst(0); 
+    ArrayRCP<real_type> yvals = coordinates->getDataNonConst(1); 
+    ArrayRCP<real_type> zvals = coordinates->getDataNonConst(2);
```
mayrmt, line 116: This could be const:
```suggestion
    ArrayRCP<real_type> xvals = coordinates->getData(0); 
    ArrayRCP<real_type> yvals = coordinates->getData(1); 
    ArrayRCP<real_type> zvals = coordinates->getData(2); 
```

mayrmt (approved):
@rstumin Looks good!

PR #9954, opened 2021-11-19
MueLu: add kokkos version of new SaP row constraint. Unit test for new SaP row constraint.

pohm01:
@trilinos/muelu 

## Motivation
New row constraint did not have a unit test or a kokkos version.

## Testing
This contains unit tests for the new SaP row constraint function.


mayrmt:
@pohm01 What's the difference between `SatisfyPConstraints()` and `new SatisfyPConstraints()`? This does not become obvious from the naming, as the prefix `new` is not very informative. Can you think of a more descriptive name?

When adding a new version, what will happen to the old one? Is it still required or are the plans to deprecate it now or at some later point?

pohm01:
> @pohm01 What's the difference between SatisfyPConstraints() and new SatisfyPConstraints()? This does not become obvious from the naming, as the prefix new is not very informative. Can you think of a more descriptive name?
> When adding a new version, what will happen to the old one? Is it still required or are the plans to deprecate it now or at some later point?

@rstumin may be better equipped to answer these questions. I was mostly just trying my hand at writing up a kokkos version.

rstumin:
@mayrmt @pohm01 newSatisfyPConstraints() should probably be changed to optimalSatisfyPConstraintsForScalarPDEs() while SatisfyPConstraints() should probably be called suboptimalSatisfyPConstraintsForPDESystems(). That would be a more descriptive name. At some point the 'optimal' version could be upgraded to handle PDESystems .... but it is not a top priority because we are still just assessing how useful this capability actually is

rstumin:
@pohm01 could you take care of this renaming.

jhux2 (commented):
There are a number of for loops inside the functor.  Can those be further converted to Kokkos?

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_SaPFactory_kokkos_def.hpp
```diff
@@ -360,6 +360,207 @@ struct constraintKernel {
...
+       {
+         SC aBigNumber;
+         SC rowSumDeviation, temp, delta;
+         SC closestToLeftBoundDist, closestToRghtBoundDist;
+         LO closestToLeftBound, closestToRghtBound;
+         bool   flipped;
+
+         SC leftBound = zero;
+         SC rghtBound = one;
+         if ((Kokkos::ArithTraits<SC>::real(rsumTarget) >= Kokkos::ArithTraits<SC>::real(leftBound*((SC)nnz))) &&
```
lucbv, line 408: You might want to add the following alias above:
```c++
using KAT = Kokkos::ArithTraits<SC>
```

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_SaPFactory_kokkos_def.hpp
```diff
@@ -360,6 +360,207 @@ struct constraintKernel {
...
+
+           flipped    = false;
+           // compute aBigNumber to handle some corner cases where we need
+           // something large so that an if statement will be false
+           aBigNumber = Kokkos::ArithTraits<SC>::zero();
+           for (auto entryIdx = rowPtr(rowIdx); entryIdx < rowPtr(rowIdx + 1); entryIdx++){
+             if ( Kokkos::ArithTraits<SC>::magnitude( values(entryIdx) ) > Kokkos::ArithTraits<SC>::magnitude(aBigNumber)) 
+               aBigNumber = Kokkos::ArithTraits<SC>::magnitude( values(entryIdx) );
+           }
+           aBigNumber = aBigNumber+ (Kokkos::ArithTraits<SC>::magnitude(leftBound) + Kokkos::ArithTraits<SC>::magnitude(rghtBound))*((SC)100.0);
```
lucbv, line 419: Instead of ((SC)100.0) you should do `100*KAT::one()` that way you will likely not get any conversion failures

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_SaPFactory_kokkos_def.hpp
```diff
@@ -360,6 +360,207 @@ struct constraintKernel {
...
+           // the code that follow after this if statement assumes that rowSumDeviation is positive. If this
+           // is not the case, flip the signs of everything so that rowSumDeviation is now positive. 
+           // Later we will flip the data back to its original form.
+           if (Kokkos::ArithTraits<SC>::real(rowSumDeviation) < Kokkos::ArithTraits<SC>::real(Kokkos::ArithTraits<SC>::zero())) {
+             flipped = true;
+             temp = leftBound; leftBound = -rghtBound; rghtBound = temp; 
+         
+             /* flip sign of origSorted and reverse ordering so that the negative version is sorted */
+         
+             if ((nnz%2) == 1) origSorted(rowIdx, (nnz/2)  ) =  -origSorted(rowIdx, (nnz/2)  );
```
lucbv, line 480: The above is pretty bad for GPU performance... you can do bit shifting instead if you feel brave.

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_SaPFactory_kokkos_def.hpp
```diff
@@ -360,6 +360,207 @@ struct constraintKernel {
...
+  
+           // initial fixedSorted so bounds are satisfied and interiors correspond to origSorted
+         
+           for (LO i =                  0; i < closestToLeftBound; i++) fixedSorted(rowIdx, i) = leftBound;
+           for (LO i = closestToLeftBound; i < closestToRghtBound; i++) fixedSorted(rowIdx, i) = origSorted(rowIdx, i);
+           for (LO i = closestToRghtBound; i < (LO) nnz;           i++) fixedSorted(rowIdx, i) = rghtBound;
+         
+           while ((Kokkos::ArithTraits<SC>::magnitude(rowSumDeviation) > Kokkos::ArithTraits<SC>::magnitude(((SC)1.e-10)*rsumTarget))){ // && ( (closestToLeftBound < nEntries ) || (closestToRghtBound < nEntries))) {
+            if (closestToRghtBound !=  closestToLeftBound)
+                 delta = rowSumDeviation/ (SC)(closestToRghtBound -  closestToLeftBound);
```
lucbv, line 507: Use a static_cast<SC> instead of (SC) please

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_SaPFactory_kokkos_def.hpp
```diff
@@ -375,6 +576,21 @@ struct constraintKernel {
...
+  void SaPFactory_kokkos<Scalar,LocalOrdinal,GlobalOrdinal,Kokkos::Compat::KokkosDeviceWrapperNode<DeviceType>>::optimalSatisfyPConstraintsForScalarPDEs( RCP<Matrix>& P) const {
+
+    using Device = typename Matrix::local_matrix_type::device_type;
+    LO nPDEs = 1;//A->GetFixedBlockSize();
+
+    using local_mat_type = typename Matrix::local_matrix_type;
+    // what is the difference between getGlobalMaxNumRowEntries and getNodeMaxNumRowEntries?
+    optimalSatisfyConstraintsForScalarPDEsKernel<local_mat_type> myKernel(nPDEs,P->getGlobalMaxNumRowEntries(),P->getLocalMatrixDevice() );
+    //optimalSatisfyConstraintsForScalarPDEsKernel<local_mat_type> myKernel(nPDEs,P->getNodeMaxNumRowEntries(),P->getLocalMatrixDevice() );
+    Kokkos::parallel_for("enforce constraint",Kokkos::RangePolicy<typename Device::execution_space>(0, P->getRowMap()->getNodeNumElements() ),
```
lucbv, line 589: I think `P->getNodeNumRows()` would be more explicit for future maintenance

lucbv (changes requested):
Some small changes requested, mostly you should define a short hand for Kokkos::ArithTraits<SC> and avoid casting with () operator, prefer the c++ static_cast<>() instead as the compiler will clearly tell you what is going wrong if casting fails.
Additionally you might want to define `SC_one` and `SC_zero`

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_SaPFactory_kokkos_def.hpp
pohm01: Done, thanks.

packages/muelu/src/Transfers/Smoothed-Aggregation/MueLu_SaPFactory_kokkos_def.hpp
jhux2: @lucbv Is this something that @pohm01 could do later in a separate PR?

pohm01: Added a comment noting that this is bad for GPU performance, but I feel switching to bit shifting is beyond the scope of this PR. This can be addressed later if this function begins to see heavy use.

jhux2 (approved):
lgtm.  @pohm01 has addressed all of @lucbv's comments except the bit shifting.

PR #9964, opened 2021-11-22
Amesos2::SuperLU fix with option to use Metis

iyamazaki:
@trilinos/amesos2 

## Motivation
This PR fixes the option to use METIS with SuperLU (sequential).
- skip calling get-perm-c when use-METIS option is selected
- check for empty graph of A+A^T used to call METIS before freeing it


ndellingwood (approved):

srajama1 (approved):
Thanks @iyamazaki

Issue #9886, opened 2021-10-29
Tpetra:  FECrsGraph Unit Tests pass but output says some processor FAILED


kddevin:
## Question
@trilinos/tpetra 

FECrsGraph Unit Tests pass, but in their output, there are several instances of "FAILED on at least one process!"  See output below.

Should the tests pass despite the output, or are is the test passing by mistake?
- If the tests' passing is correct, the output be changed to something like, "FAILED on at least one process as expected."
- If the output about FAILED tests is correct, the TriBITS or test logic needs to be changed to catch the failure.  (And, then, eventually, the failure should be fixed as well.)

@tjfulle can you take a look please?  Thank you.

 ```
 mpirun -np 4 TpetraCore_FECrsGraph_UnitTests.exe 
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascicgpu030 and rank 0!
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascicgpu030 and rank 2!
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascicgpu030 and rank 1!

***
*** Unit test suite ...
***


Sorting tests by group name then by the order they were added ... (time = 1.5e-05)

Running unit tests ...

Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascicgpu030 and rank 3!
0. FECrsGraph_int_longlong_Kokkos_Compat_KokkosSerialWrapperNode_Diagonal_UnitTest ... [Passed] (0.00373 sec)
1. FECrsGraph_int_longlong_Kokkos_Compat_KokkosSerialWrapperNode_Diagonal_LocalIndex_UnitTest ... [Passed] (0.000921 sec)
2. FECrsGraph_int_longlong_Kokkos_Compat_KokkosSerialWrapperNode_Assemble1D_UnitTest ... [Passed] (0.0108 sec)
3. FECrsGraph_int_longlong_Kokkos_Compat_KokkosSerialWrapperNode_Assemble1D_LocalIndex_UnitTest ... [Passed] (0.00953 sec)
Succeeded on all processes!
Succeeded on all processes!
FAILED on at least one process!
Succeeded on all processes!
Succeeded on all processes!
FAILED on at least one process!
Succeeded on all processes!
Succeeded on all processes!
FAILED on at least one process!
[...]

PR #9965, opened 2021-11-22
Tpetra: Fixed two errors in Tpetra::FECrsMatrix test

tjfulle:
1. Check for number of rows in graph was checking number of rows against number of entries
2. Assignment of success from one test was being overwritten by further assignments of success
   so that if an evaluation failed and the next succeeded the test would report as succeeding
   overall.

@trilinos/tpetra 

Closes: #9886


tjfulle:
@kddevin or @csiefer2 - do you mind approving this PR?

csiefer2 (approved):

PR #9967, opened 2021-11-23
Framework: Migrate Autotester TPLs

ZUUL42:
@trilinos/framework 

## Motivation

The current TPLs used for the autotester are moving locations.
This code change will move to that archived set of TPLs to allow for more time to move to the new set of spack-cm released TPLs.


jwillenbring (approved):
Looks good. Thanks.

PR #9981, opened 2021-12-01
Piro: some fixes / additions relevant to Schwarz coupling in Albany-LCM.

ikalash:
In trying to understand why Explicit-Implicit and Explicit-Explicit Schwarz coupling does not work anymore in Albany-LCM, I came across some things that can be improved within Piro.  This PR implements those fixes/additions.


lxmota (approved):
Looks good to me

Issue #9500, opened 2021-07-28
Zoltan2:  TpetraCrsColorer with noncontiguous maps


kliegeois:
## Bug Report
@trilinos/zoltan2 @mperego 
### Description
A bug has been observed in Albany when reconstructing a Hessian matrix using the `Zoltan2::TpetraCrsColorer` when the maps of the graph of the matrix are not contiguous (i.e. the global IDs included in the maps are not contiguous).

For the observed occurrences, the `Zoltan2::TpetraCrsColorer` was set to be used with the symmetry option and the reconstructed matrix was not symmetrical. Most of the entries were fine (more than 90% of the entries were correct) but few rows related to the last global IDs were not correct.

As an example, the following figure illustrates the correct Hessian of a response in Albany:

![Figure_1](https://user-images.githubusercontent.com/15634348/127350033-bf03292a-f7e4-4c8f-9d9b-f3928d80ffed.png)

And this second figure illustrates the difference between the correct Hessian and the Hessian computed using `Zoltan2::TpetraCrsColorer` with the symmetry option set to true and noncontiguous maps (the entries of the matrix have been reordered for comparison purpose with the contiguous case):

![Figure_5](https://user-images.githubusercontent.com/15634348/127350083-abf5eba4-617a-4a09-aa26-899e85976217.png)

The problem has been overcome in Albany using contiguous maps but it will be great if the `Zoltan2::TpetraCrsColorer` can support the noncontiguous cases.

kddevin:
Appears to be a bug in Zoltan2's TpetraCrsColorer's symmetrization of the matrix before coloring.  I'm on it.

kliegeois:
Thanks a lot!

mperego:
Thanks @kddevin !

kddevin:
@kliegeois  Does #9982 fix the problem in Albany?  If so, please feel free to close this issue.

mperego:
@kddevin The test in Albany still fails for non contiguous ids. The matrix computed with coloring is not symmetric. Not sure if the issue is in Zoltan2, I would need to dig into it. @kliegeois is on vacation.

kddevin:
@mperego Thanks; I'll take another look.

kddevin:
@mperego can you point me to the Albany code that is calling the colorer?  I'd like to see the calling sequence.  Thanks.

mperego:
@kddevin here is the code calling the coloring
[...]

PR #9982, opened 2021-12-01
Zoltan2:  fix indexing for non-contiguous row maps in TpetraCrsColorer's bipartite symmetrization

kddevin:
@trilinos/zoltan2

## Motivation

In #9500, @kliegeois reported incorrect results from TpetraCrsColorer when using matrices in non-contiguous row maps.
This PR fixes the problem.

The bipartite symmetrization used the number of global entries in the row map as the first index of the column vertices.  However, for a non-contiguous map with the largest index greater than the global number of entries, the number of global entries is incorrect.  Using the maximum global ID instead solves the problem, at least for the new test Bug9500.cpp.

#9500 

## Stakeholder Feedback

@kliegeois 

## Testing

New test added:  Bug9500.cpp -- builds row maps with GIDs that are multiples of a specified "multiple"; does coloring for matrices with contiguous (multiple=1) and non-contiguous (multiple=2, 5) row maps.  
Test failed for multiple > 1; this PR allows the test to run with all multiples.
Run on ascicgpu with SEMS modules (serial node, CPU only)

### DIFF
--- a/packages/zoltan2/core/src/TpetraCrsColorer/Zoltan2_TpetraCrsColorer_Zoltan.hpp
+++ b/packages/zoltan2/core/src/TpetraCrsColorer/Zoltan2_TpetraCrsColorer_Zoltan.hpp
@@ -270,7 +270,8 @@ ZoltanCrsColorer<CrsMatrixType>::computeColoring(
   // Do coloring of columns with Zoltan -- we can request colors for
   // columns we don't own
   const size_t num_local_cols  = this->graph->getNodeNumCols();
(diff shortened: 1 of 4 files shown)


egboman (approved):
Good catch. Thanks, @kddevin !

rsivascience (approved):
Thanks @kddevin !

srajama1 (approved):
Approving with right github account.

egboman (approved):
Approving again. Thanks, @kddevin !

PR #9985, opened 2021-12-03
Stokhos: Remove power of two

ndellingwood:
@trilinos/stokhos 

## Motivation
Kokkos is removing Kokkos::Impl::power_of_two, this update is for compatibility with kokkos/kokkos#4578

## Testing
Autotester

### DIFF
--- a/packages/stokhos/src/kokkos/Stokhos_CrsProductTensor.hpp
+++ b/packages/stokhos/src/kokkos/Stokhos_CrsProductTensor.hpp
@@ -817,7 +817,7 @@ public:
                       const size_type thread_rank ) const
   {
     enum { work_align = 64 / sizeof(VectorValue) };
-    enum { work_shift = Kokkos::Impl::power_of_two< work_align >::value };
+    enum { work_shift = Kokkos::Impl::integral_power_of_two( work_align ) };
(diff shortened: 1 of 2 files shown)


ndellingwood:
Not sure why the change in stokhos triggered review request from so many other packages...

cgcgcg:
@ndellingwood Because the PR was initially against master, and hence included changes in lots of other packages.

etphipp (approved):
Looks good to me.

dalg24 (changes requested):
No this is equally as bad to use that function as it is defined in the `Kokkos::Impl::` namespace.
I recommend just moving the `power_of_two` definition into Stokhos.

william76:
@ndellingwood please don't forget to add the project/package prefix to your PR titles. In this case, you missed adding "Stokhos:".

etphipp (approved):

etphipp:
So what are we doing with this PR.  It is fine with me as-is, but there were objections.

ndellingwood:
@etphipp I can put in a different PR to add the power_of_two compile time routines to Stokhos instead of the changes here to avoid the Impl routines.

ndellingwood:
Closing, replaced by #10044

PR #9989, opened 2021-12-06
Framework: Spack environment

fryeguy52:
@trilinos/framework 

## Motivation
This is the begging of work on the trilinos spack package

see: TRILFRAME-199

### DIFF
--- a/cmake/std/spack-environment/README
+++ b/cmake/std/spack-environment/README
@@ -1,0 +1,6 @@
+This is a very simple spack environment to build trilinos.  It is built off of the v0.16.2 tag in the spack repository.
+
+1) install spack at v0.16.2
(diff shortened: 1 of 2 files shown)


william76:
@fryeguy52 The convention in Trilinos for issues and Pull Requests is to format the title as: `<TeamName>: <title>`

For Framework, these should be prefixed as `Framework: <title>`

e10harvey (changes requested):
Thanks, @fryeguy52 ! Please change `root: ...` to `root: $HOME/trilinos-spack-test-install`.

Will `install_missing_compilers: true` attempt to load the gcc/10.1.10 module?

jwillenbring:
@fryeguy52 did you see the changes suggested by @e10harvey ?

PR #9993, opened 2021-12-06
MueLu: add option to control whether dropping can lead to additional Dirichlet rows

jhux2:
This commit adds an option `aggregation: dropping may create Dirichlet` to control whether filtering during aggregation can create new Dirichlet rows.  The default value is `true`.

@trilinos/muelu

## Motivation
For certain problems (SPE10) with very small/jumpy coefficients, solver convergence is much better if dropping during aggregation isn't allowed to lead to new Dirichlet rows.

## Stakeholder Feedback
Requested by @rstumin.

## Testing
This PR includes a unit test, and @rstumin has confirmed that when this option is set to `false`, certain SPE10 linear systems can be solved effectively.

### DIFF
--- a/packages/muelu/doc/UsersGuide/masterList.xml
+++ b/packages/muelu/doc/UsersGuide/masterList.xml
@@ -748,6 +748,16 @@
       <comment-ML>parameter not existing in ML</comment-ML>
     </parameter>
 
+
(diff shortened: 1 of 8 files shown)


rstumin (approved):
Looks good.

PR #9999, opened 2021-12-08
Zoltan:  move files to enable github to render web pages

kddevin:
@trilinos/zoltan 

## Motivation

Sandia closed down www.cs.sandia.gov (and, thus, Zoltan's web pages) without notifying anyone.  Fortunately,  @victoriasmith and @wadeburgess let me know about the changes.  I was told the best option was to move the pages, so I am moving them to github.io as directed by @wadeburgess .  

To use github.io, it appears that the HTML files need to be in a directory named "docs".  I couldn't get symbolic links to work, so I am just moving the files.  

## Testing

I tested that the github.io builds work from the https://github.com/sandialabs/Zoltan/ repo from which Zoltan distributions are done.  This commit mimics the changes I made in that repo.  Future updates to that repo should, thus, be straightforward, as the directory structures will match.


rsivascience (approved):
Thanks @kddevin !

srajama1 (approved):
With the right github account.

egboman (approved):
Looks good. Thanks, @kddevin !

Issue #10008, opened 2021-12-10
Tpetra: importAndFillComplete fails for one-to-many distributions


kddevin:
## Bug Report
@trilinos/tpetra @trilinos/zoltan2 @csiefer2 @tjfulle 

### Description
While debugging #10001, I'm trying to let Zoltan2 balance and migrate a matrix that is entirely on rank 0 initially; the resulting matrix should be spread across processors.

Below is a standalone code showing how I am trying to do the migration.   
@csiefer2  @tjfulle  Is my code using importAndFillComplete correctly?   
 It throws an error in lowCommunicationMakeColMapAndReindex, which is called from importAndFillComplete
```
 p=0: *** Caught standard std::exception of type 'std::invalid_argument' 
  /ascldap/users/kddevin/code/Trilinos.develop/packages/tpetra/core/src/Tpetra_Import_Util2.hpp:793:
  Throw number = 1
  Throw test that evaluated to true: PID == -1
  lowCommunicationMakeColMapAndReindex: Cannot figure out if PID is owned.
 [FAILED]  (0.0033 sec) CrsMatrix_double_int_longlong_Kokkos_Compat_KokkosSerialWrapperNode_Bug10001_UnitTest
```

Here's the code:
```  
  typedef Tpetra::global_size_t GST;
  const GST dummy = Teuchos::OrdinalTraits<GST>::invalid();

  typedef Tpetra::Map<LO,GO,NT> map_t;
  typedef Tpetra::Import<LO,GO,NT> import_t;
  typedef Tpetra::CrsMatrix<SC,LO,GO,NT> matrix_t;

  // Map with everything on rank zero
  const GO nRows = 25;
  const int nPerRow = 5;
  Teuchos::RCP<const map_t> mapOne =
           rcp(new map_t(dummy, (me == 0 ? nRows : 0), 0, comm));
    
  // Matrix using mapOne -- all data on rank 0
  matrix_t matrixOne(mapOne, nPerRow);

  if (me == 0) {
    Teuchos::Array<GO> gids(nPerRow);
    Teuchos::Array<SC> values(nPerRow);
    for (GO i = 0; i < nRows; i++) {
      // five-point stencil
      int nnz = 0;
      gids[nnz++] = i;
      if (((i-1) >= 0) && (((i-1)%nPerRow) < (i%nPerRow))) gids[nnz++] = i-1; // west nbor
[...]

PR #10011, opened 2021-12-13
Tpetra:  Reproducer for #10008

kddevin:
@trilinos/tpetra 

## Motivation

This PR has a reproducer for #10008.

The test creates a matrix with all entries on one processor.  It attempts to importAndFillComplete it to a matrix on all processors.

The test will throw and error on one rank and hang on the others.

The error is thrown from lowCommunicationMakeColMapAndReindex.

The error is reproducible on ascicgpu machines with SerialNode and SEMS gcc7.2 modules.


tjfulle:
I'm building and running the reproducer right now.  Will update #10008 when more info is learned.

tjfulle:
@kddevin and @csiefer2 - the changes in this PR now pass @kddevin's tests.  Waiting to see if the changes have side effects in the PR testing...

csiefer2 (approved):

kddevin:
@tjfulle do you have time to fix the warnings-as-errors compilation failures?

kddevin:
@csiefer2  The tests finally passed.  Whew!  Would you please re-review?  Thank you!

kddevin:
@trilinos/tpetra. This PR has passed but needs review.  Please help. Thanks.

tasmith4:
@kddevin what is the change in CrsMatrix_def?  Is that the fix for the problem this test reproduces?

tjfulle:
@tasmith4 - the matrix packing routine puts -1 in the array of TargetPids for local entries (rather than its own pid). We believe it is a hold over from Epetra to ensure proper ordering of pids when sorted (local entries first).  This change is to replace the -1 with the actual pid before calling other routines (that expect an actual pid)

Issue #9743, opened 2021-09-24
MueLu: improve factory unit testing


jhux2:
## Enhancement
@trilinos/Muelu @csiefer2 

This is a checklist of MueLu classes identified as having less than 75% code coverage (per analysis by @JacobDomagala on 2021-Sept-09).

`muelu/adapters/stratimikos`
 - [ ] Thyra_MueLuPreconditionerFactory
 - [ ] Thyra_XpetraLinearOp

`muelu/adapters/tpetra`
 - [ ] **MueLu_TpetraOperator**

`muelu/adapters/xpetra`
 - [ ]   MueLu_Maxwell_Utils
 - [ ]   MueLu_RefMaxwell
 - [ ]   MueLu_Maxwell1

`muelu/src/Graph/Containers`
 - [ ] **MueLu_LWGraph_kokkos_def**
 - [ ] **MueLu_LWGraph**
 - [ ] MueLu_Zoltan2GraphAdapter

`muelu/src/Graph/HybridAggregation`
 - [ ] MueLu_HybridAggregationFactory (*must enable MueLu research code)

`muelu/src/Graph/MatrixTransformation`
 - [ ] MueLu_VariableDofLaplacianFactory
 - [ ] **Muelu_AmalgamationInfo**
 - [ ] **MueLu_CoalesceDropFactory**

`muelu/src/Graph/StructuredAggregation`
 - [ ] MueLu_IndexManager

`muelu/src/Graph/StructuredAggregation/coupled`
 - [ ] MueLu_GlobalLexicographicIndexManager

`muelu/src/Graph/StructuredAggregation/uncoupled`
 - [ ] MueLu_UncoupledIndexManager

`muelu/src/Graph/UncoupledAggregation`
 - [ ] **MueLu_UncoupledAggregationFactory_kokkos**
 - [ ] **MueLu_UncoupledAggregationFactory**
 - [ ] **MueLu_PreserveDirichletAggregationAlgorithm_kokkos**
 - [ ] **MueLu_AggregationPhase3Algorithm**

`muelu/src/Graph/UserAggregation`
 - [ ] MueLu_UserAggregationFactory

`muelu/src/Interface`
 - [ ] MueLu_AdaptiveSaMLParameterListInterpreter
 - [ ] **MueLu_HierarchyManager**
 - [ ] MueLu_ML2MueLuParameterTranslator

`muelu/src/Interface/FacadeClasses`
 - [ ]   MueLu_Facade_Simple
[...]

PR #10013, opened 2021-12-13
WIP Added code that improves coverage but failed with cdash

mperrinel:
Related to #9743

This code has been extracted from the PR #9832 to split it in a way where #9832 passed automatic tests and this one fails with CDASH and need more fix


GrahamBenHarper:
bump?

Issue #10016, opened 2021-12-14
Test Ifpack2_unit_tests_MPI_4 unit tests randomly failing in  many ATDM and PR builds since at least 2021-08-30


bartlettroscoe:
CC: @trilinos/ifpack2, @<triage-contact\> (Trilinos <product-area-name\> Triage Contact (or "Current ATDM contact"))

<Checklist>
<???: Add label "type: bug"?>
<???: Add label "impacting: [tests, build, performance]"?>
<???: Add label "pkg: [from TestName]"?>
<???: Add label "PA: [see https://sems-atlassian-son.sandia.gov/confluence/display/TRIL/Triaging+contacts]?>
<???: Add label "Primary Build"?>
<???: Add label "Secondary Build"?>
<???: Add project "Build Triage"?>
<???: Once GitHub Issue is created, add entries for tests to TrilinosATDMStatus/*.csv files?>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&begin=2021-08-01&end=2021-12-14&filtercount=3&showfilters=1&filtercombine=and&field1=testname&compare1=61&value1=Ifpack2_unit_tests_MPI_4&field2=status&compare2=61&value2=failed&field3=testoutput&compare3=97&value3=FAILED.*Ifpack2Chebyshev_double_int_longlong_Test0_UnitTest) (click "Shown Matching Output" in upper right) the tests:

* `Ifpack2_unit_tests_MPI_4`

in the builds:

* `PR-9483-test-Trilinos_pullrequest_clang_10.0.0-3559`
* `PR-9483-test-Trilinos_pullrequest_gcc_7.2.0_debug-3527`
* `PR-9483-test-Trilinos_pullrequest_gcc_7.2.0_debug-3591`
* `PR-9627-test-Trilinos_pullrequest_cuda_10.1.105-2132`
* `PR-9627-test-Trilinos_pullrequest_cuda_10.1.105_uvm_off-1129`
* `PR-9660-test-Trilinos_pullrequest_gcc_7.2.0_debug-3528`
* `PR-9660-test-Trilinos_pullrequest_gcc_7.2.0_debug-3538`
* `PR-9676-test-Trilinos_pullrequest_clang_10.0.0-3585`
* `PR-9691-test-Trilinos_pullrequest_clang_10.0.0-3641`
[...]

PR #10017, opened 2021-12-14
Ifpack2: Bump up tolerance in Chebyshev unit test

cgcgcg:
@trilinos/ifpack2 

## Motivation
Increase tolerance in unit test so that it passes on all systems, see #10016.


bartlettroscoe (approved):
That should do it.  Thanks for addressing this quickly @cgcgcg!

jhux2:
@cgcgcg Thanks for patching this.  Fyi, @rstumin had a suggestion for an analytic Chebyshev test -- I no longer recall the details.

PR #10018, opened 2021-12-14
MueLu : fix NullspaceFactor unit test

iyamazaki:
@trilinos/muelu 

## Motivation

This PR puts getDataNonConst in a separate scope to avoid "Cannot access data on device while a host view is alive" in NullSpaceFactor unit test.

## Testing

I've tested on an ASCICGPU machine without UVM.


jhux2 (approved):
lgtm

Issue #2163, opened 2018-01-16
Tempus commits on 1/10 causing slow-downs in Albany dynamic problems


ikalash:
Continuing the discussion of the following Albany issue https://github.com/gahansen/Albany/issues/243 which is really a Trilinos issue.  

Some recent Tempus commits have slowed down substantially (5x) several Albany dynamic problems that utilize Tempus.  Here are some sample timings for the Schwarz_Alternating_Dynamics_Cubes test in Albany along with Trilinos commit tags:

6ab30362499394c06a019be52ee6a243a379dd35 : 19.14 sec 
1d0aae2911cd77a48965cd0afe9057b15fc560a8 : 104.95 sec

This means that the commit 1d0aae2911cd77a48965cd0afe9057b15fc560a8 (Tempus: move integrator output to observer) started this problem.

Here is some sample verbose output from the Albany CDash:

1/10 run: http://cdash.sandia.gov/CDash-2-3-0/testDetails.php?test=3387873&build=65282
1/11 run (slower one): http://cdash.sandia.gov/CDash-2-3-0/testDetails.php?test=3390813&build=65335

@trilinos/tempus

ikalash:
I have dug into this a little bit and the problem has to do with there being many more observer calls.  I have put print statements in Piro_ObserverToTempusIntegrationObserverAdapter_Def.hpp, observeEndIntegrator routine, to provide an example.  Before the slow-downs, there are 836 calls to this routine; after, there are 175978 calls (for the serial alternating Schwarz dynamic cubes problem in Albany which we are debugging).  Here is some sample output:

Before slowdown:
```
Time integration complete.

IKT observeEndIntegrator
IKT observeTimeStep
IKT observing solution time = 0.05
T final actual: 5.000e-02

F) Check the solution to the forward problem ...
```

After slowdown:
```
Time integration complete.

IKT observeEndIntegrator
IKT observeTimeStep
IKT observing solution time = 0.05
IKT observeEndIntegrator
IKT observeTimeStep
IKT observing solution time = 0.05
T final actual: 5.000e-02

F) Check the solution to the forward problem ...

```
For some reason, it gets called twice now after the problem is solved.

sconde:
This was fixed by #2189

PR #10025, opened 2021-12-15
Tempus: Remove IntegratorBasicOld

### LINKED ISSUE #10023: Tempus:  Remove IntegratorBasicOld
## Enhancement
@trilinos/tempus 

Remove the IntegratorBasicOld class.  It was used to allow transition to the new IntegratorBasic that removed internal usage of ParameterLists and a few other improvements.  The last derived classes using IntegratorBasicOld have been move to inherit from the new IntegratorBasic, so it is time to deprecate IntegratorBasicOld.

ccober6:
@trilinos/tempus 

## Motivation
Remove the IntegratorBasicOld class. It was used to allow transition
to the new IntegratorBasic that removed internal usage of ParameterLists
and a few other improvements. The last derived classes using
IntegratorBasicOld have been move to inherit from the new
IntegratorBasic, so it is time to deprecate IntegratorBasicOld.

## Related Issues
* Closes #10023

### DIFF
--- a/packages/piro/src/Piro_TempusIntegrator_Def.hpp
+++ b/packages/piro/src/Piro_TempusIntegrator_Def.hpp
@@ -53,16 +53,16 @@
 #include <iostream>
 
 template <typename Scalar>
-Piro::TempusIntegrator<Scalar>::TempusIntegrator(Teuchos::RCP< Teuchos::ParameterList > pList, 
+Piro::TempusIntegrator<Scalar>::TempusIntegrator(Teuchos::RCP< Teuchos::ParameterList > pList,
    const Teuchos::RCP< Thyra::ModelEvaluator< Scalar > > &model,
-   const SENS_METHOD sens_method) : 
+   const SENS_METHOD sens_method) :
    out_(Teuchos::VerboseObjectBase::getDefaultOStream())
 {
   if (sens_method == NONE) {
     //no sensitivities
-    basicIntegrator_ = Tempus::integratorBasic<Scalar>(pList, model);
-    fwdSensIntegrator_ = Teuchos::null; 
-    adjSensIntegrator_ = Teuchos::null; 
+    basicIntegrator_ = Tempus::createIntegratorBasic<Scalar>(pList, model);
+    fwdSensIntegrator_ = Teuchos::null;
+    adjSensIntegrator_ = Teuchos::null;
   }
   else if (sens_method == FORWARD) {
     //forward sensitivities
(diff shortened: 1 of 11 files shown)


packages/piro/src/Piro_TempusIntegrator_Def.hpp
```diff
@@ -107,192 +107,188 @@ Piro::TempusIntegrator<Scalar>::TempusIntegrator(Teuchos::RCP< Teuchos::Paramete
...
 }
 
 template <typename Scalar>
-void 
+void
 Piro::TempusIntegrator<Scalar>::clearObservers()
 {
   if (basicIntegrator_ != Teuchos::null) {
-    basicIntegrator_->getObserver()->clearObservers();
+    basicIntegrator_->setObserver();
```
sconde, line 250: Is this going to change the timing in Albany?

ccober6: Hi @sconde!  Not I sure I understand.  Just switching to IntegratorBasic which holds the base class IntegratorObserver instead of a composite.  You can still give it a composite.  The IntegratorBasic just does not know about composite functionality.

ccober6:
@ikalash This is the PR I mentioned that removes IntegratorBasicOld and may affect Albany.  If you can take I look and let us know if it causes any problems.  Thanks, Curt

packages/piro/src/Piro_TempusIntegrator_Def.hpp
sconde: I'm worried about this issue: https://github.com/trilinos/Trilinos/issues/2163

ccober6: Gotcha.  I do not think this will be a problem here.

In IntegratorBasicOld, integratorObserver_ is an IntegratorObserverComposite and the setObserver(obs) function would **_add_** observers, obs, to integratorObserver_.   This means Piro needed to clearObservers() to make sure duplicates were not added.

In IntegratorBasic, integratorObserver_ is an IntegratorObserver and the setObserver(obs) function just sets the observer, obs, to integratorObserver_, while setObserver() (w/o args) sets integratorObserver_ to IntegratorObserverBasic.  So Piro needs to make sure it clears its IntegratorObserverComposite (which it appears they do), and then just set it on the IntegratorBasic.

ikalash:
@ccober6 : sure I'll test Albany against the relevant branch.

ikalash:
I tested Albany and Albany-LCM and everything looks good.

ccober6:
@etphipp Would you mind reviewing this?  I touched several sensitivity integrators, and would like you to see if I missed anything.  Thanks, Curt.

ccober6:
@trilinos/framework  It looks like the autotester is not grabbing the current version of this branch to test.  Is there something I can do, or do you need to do something?  Thanks!

jwillenbring:
@ccober6 It looks like you may have force pushed some new commits? When this happens the testing sometimes breaks like this. The best thing would probably be to close this PR and just submit a new one with your current branch.

ccober6:
Yes, I did do a forced push because I needed to rebase to resolve some conflicts.  I will go ahead and close this and open a new PR.  Thanks!

PR #10027, opened 2021-12-15
MueLu: Fix failing CuthillMcKee tests

JacobDomagala:
This "fix" removes the problematic checks. 

See https://github.com/trilinos/Trilinos/issues/10026 for more information

### DIFF
--- a/packages/muelu/test/unit_tests_kokkos/Utilities_kokkos.cpp
+++ b/packages/muelu/test/unit_tests_kokkos/Utilities_kokkos.cpp
@@ -89,35 +89,22 @@ namespace MueLuTests
     out << "version: " << MueLu::Version() << std::endl;
 
     // Build the problem
-    RCP<Matrix> A = MueLuTests::TestHelpers_kokkos::TestFactory<Scalar, LocalOrdinal, GlobalOrdinal, Node>::Build1DPoisson(2001);
+    RCP<Matrix> A = MueLuTests::TestHelpers_kokkos::TestFactory<Scalar, LocalOrdinal, GlobalOrdinal, Node>::Build1DPoisson(100);
 
     // CM Test
     {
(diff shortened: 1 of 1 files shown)


JacobDomagala:
@jhux2 This should get rid of failures on CI

brian-kelley:
@JacobDomagala If you would still like to sanity check the actual contents of ``ordering`` and not just the length, you could check that it contains every number from ``0`` to ``A->getNodeNumRows()-1`` exactly once (meaning it's a valid permutation).

cgcgcg:
Does it make sense to check whether the RCM is idempotent?

jhux2:
@brian-kelley @cgcgcg RCM was producing a strange ordering on one rank.  So for the moment, I suggested disabling a check of the ordering itself.

brian-kelley:
@jhux2 Are you referring to the last rank where the ordering was reversed compared to the others? I just realized that it happens because the BFS starts at the vertex with minimum degree. But degree is computed using the rowptrs, and so it counts remote edges. On rank 3, local vertex 0 is connected to local vertex 9 from rank 2 and so it counts as degree 2.

I'm not sure whether this should be considered an issue or not. If it ignored remote edges, both endpoints would have the minimum degree 1 so RCM could still start at either one arbitrarily. Also, filtering remote edges would cost some extra time.

jhux2:
Yes, that's exactly the issue.  @JacobDomagala did not see any problems in his local build.

cgcgcg:
@brian-kelley It's a valid RCM ordering, right? I guess always can flip the ordering and get the other valid one. So this sounds fine to me.

brian-kelley:
@cgcgcg Yes, it is a correct RCM. The question is whether the KokkosKernels RCM should be changed so that ranks with identical local subdomains end up with identical orderings (by ignoring off-process edges). I'm leaning towards not changing it, since if there is a parallel impl in the future it would probably be nondeterministic anyway.

cgcgcg:
@jhux2 @JacobDomagala Can this be merged?

jhux2 (approved):
lgtm

PR #9860, opened 2021-10-26
Tpetra:  reproducer for #9856

kddevin:
@trilinos/tpetra 

## Motivation

This PR has a reproducer test for #9856.
This test cannot be merged until https://github.com/kokkos/kokkos/issues/4461 is fixed and merged into Trilinos.
Exercises the Kokkos bug through Tpetra and KokkosBlas.

## Stakeholder Feedback

@jennloe 

## Testing

New test that developed as trying to reproduce #9856.
Reproduces #9856 on ascicgpu machines with UVM=OFF

Issue #1204, opened 2017-04-03
Query nonlinear solver state from Thyra::ModelEvaluator


ibaned:
This is actually a high-level design issue for Albany, but I'm opening it as a Trilinos issue to get feedback from Trilinos developers.

In Albany, we interface with non-linear solvers like NOX through a `Thyra::Model Evaluator` derived class. Most of the interesting interaction happens through the `evalModelImpl` method.

The NOX algorithm goes roughly like this:

```cpp
while(1) {
  residual = ask_for_residual(x);
  if (is_converged (residual)) break;
  jacobian = ask_for_jacobian(x);
  step = solve_linear_system (jacobian, residual);
  x += step;
}
```

The two pseudo-code ask_for calls become separate calls to `evalModelImpl`, both times `inArgs` has `x`, once `outArgs` has `W_op` and the second time `outArgs` has `f`.

The first problem is that we are starting to consider various modifications of the residual which are based on jacobian information. One example of this is scaling linear system rows based on diagonal values. Issue gahansen/Albany#55 illustrates well why this failed: we don't know which residual queries correspond to which jacobian queries. Even better for us would be if we were informed when the solution is update between one step and the next. From this we could infer associated queries.

While row scaling is something that NOX  could implement, we are also considering similar modifications that are based on Dirichlet boundary condition information, which NOX can't access natively.

A second, related problem is that it may make sense to use the non-modified residual for convergence checking and the modified residual for linear system solution. The reason is to avoid artificially changing the tolerance as a consequence of scaling. This would require some mechanism for updating the residual between the two steps.

As I think about this, I could envision a NOX PrePostOperator could handle the DBC operations too,  if it could be defined in Albany and then given to NOX. This would have to go through Piro, though maybe would not require different use of Thyra...

@lxmota @agsalin @trilinos/thyra @trilinos/nox @gahansen @bartgol

ibaned:
@bgranzow

lxmota:
Another situation where this arises is in our implementation of the Schwarz alternating method, where multiple subdomains are coupled. The global solution is obtained through a Schwarz loop that for each iteration requires a NOX nonlinear solve for each of the subdomains. The convergence criterion for the Schwarz loop requires knowledge of the solution before and after the NOX solve for each subdomain, but this is not available at that level.

rppawlo:
[...]

PR #10028, opened 2021-12-15
Patch in Kokkos Kernels #1204 to fix #9856

### LINKED ISSUE #9856: KokkosBlas:  limited vector size in norm calculation?
## Bug Report
@trilinos/tpetra 

### Description
@jennloe  reports

I have this Trilinos (Belos + Tpetra) code which is crashing, and I don't understand why

```
76   typedef double                            ScalarType;
77   typedef int                               OT;
78   typedef Tpetra::MultiVector<ScalarType>   MV;
79   typedef Tpetra::MultiVector<ScalarType>::mag_type   MT;
...
132   RCP<Tpetra::MultiVector<ScalarType>> X1 = rcp( new Tpetra::MultiVector<ScalarType>(map, 1008) );
133   X1->randomize();
134   RCP<Tpetra::MultiVector<ScalarType>> XCopy = rcp(new Tpetra::MultiVector<ScalarType>(*X1, Teuchos::Copy)); //Deep copy of X1.
....
154   // DEBUG: //Verify coefficients:
155   MV coeffs_mv = makeStaticLocalMultiVector (*X1, 1008, 1008);
156   std::cout << "Starting deep copy" << std::endl;
157   Tpetra::deep_copy(coeffs_mv, *coeffMat);
158   std::cout << "Finished deep copy" << std::endl;
159   XCopy->multiply(Teuchos::NO_TRANS, Teuchos::NO_TRANS, 1.0, *X1, coeffs_mv, -1.0);
160   std::cout << "Passed multiply." << std::endl;
161   std::vector<MT> norms(1008);
162   Teuchos::ArrayView<MT> normView(norms);
163   std::cout << "created norms array" << std::endl;
164   XCopy->norm2(normView);
165   std::cout << "Here are the QR-Orig norms.  Should be zero." << std::endl;
166   std::cout << normView << std::endl << std::endl;
```

And the output is:

```
$ ./Belos_TpetraOrthogCoeffs.exe
Starting deep copy
Finished deep copy
Passed multiply.
created norms array
  what():  cudaStreamSynchronize(m_stream) error( cudaErrorIllegalAddress): an illegal memory access was encountered /ascldap/sers/jloe/Trilinos/packages/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:314
Aborted (core dumped)
```

At first glance, the problem appears to be a mismatch between the norms argument (on host) and the vector argument (on device)

### Steps to Reproduce
Reproduced with simple test program on ascicgpu with no UVM.

brian-kelley:
Patching in https://github.com/kokkos/kokkos-kernels/pull/1204 :

Remove runtime-sized array reductions from BLAS-1 MV functions. Where
possible, use the same approach as dot-based GEMM (have multiple teams
work on a reduction, and use atomics to contribute to the final result).
If that is not possible, just process each column one at a time. Both
approaches give better performance than array reductions, and also work
for any number of columns (before, both Cuda and HIP had limits, see #9856).

I checked that the tests to replicate the issue (from #9860), pass with these changes.

@trilinos/kokkos-kernels 
@trilinos/tpetra 

## Testing

All the changes are within Kokkos Kernels. KokkosKernels_blas_* has unit tests for each of these functions.

### DIFF
--- a/packages/anasazi/tpetra/test/MVOPTester/MultiVecTraitsTest2.cpp
+++ b/packages/anasazi/tpetra/test/MVOPTester/MultiVecTraitsTest2.cpp
@@ -570,7 +570,7 @@ namespace {
       out << "]" << endl;
 
       for (size_t j = 0; j < static_cast<size_t> (ind.size ()); ++j) {
-        TEST_EQUALITY( B_view_norms[j], normsB1[ind[j]] );
+        TEST_FLOATING_EQUALITY( B_view_norms[j], normsB1[ind[j]], STS::eps() );
       }
 
       out << "Check that modifying B_view modifies the corresponding columns "
--- a/packages/anasazi/tpetra/test/MVOPTester/MultiVecTraitsTest2.cpp
+++ b/packages/anasazi/tpetra/test/MVOPTester/MultiVecTraitsTest2.cpp
@@ -135,7 +135,7 @@ namespace {
       out << "Expected 1-norm of X_" << j << ": " << X_norms[j] << endl;
       out << "Actual 1-norm of X_" << j << ": " << norms[j] << endl;
       TEST_ASSERT( X_norms[j] != STN::zero () );
-      TEST_ASSERT( norms[j] == X_norms[j] );
+      TEST_FLOATING_EQUALITY( norms[j], X_norms[j], STS::eps() );
     }
 
     out << "Create MultiVector Y" << endl;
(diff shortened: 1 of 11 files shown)


ndellingwood (approved):
These changes previously reviewed and tested in kokkos/kokkos-kernels#1204, thanks @brian-kelley

hkthorn (commented):
Thanks for the thorough explanation of the issue in this test.  I agree with the argument for putting a tolerance on the norm comparisons.  I'm wondering if that tolerance should be hard-coded or based on the scalar type being used in the test?

brian-kelley:
@hkthorn Good catch. I was thinking that ``double`` was always the default scalar type for MultiVector, but it can also be float. I'll fix it.

brian-kelley:
@hkthorn OK, now it is using ``Teuchos::ScalarTraits<S>::eps()``.

hkthorn (dismissed):
LGTM

cgcgcg (approved):

PR #10030, opened 2021-12-16
Amesos2 : SuperLU_MT adjusts 'nprocs' default parameter

iyamazaki:
@trilinos/amesos2 

## Motivation

This PR likes to adjust the default parameter value for `nprocs` such that it is the current number of threads, e.g., OMP_NUM_THREADS (the default value was always one).


ndellingwood (approved):
Thanks @iyamazaki

Issue #8441, opened 2020-12-08
Ifpack2: ok to add coloring algorithm as MTGS/MTSGS parameter?


brian-kelley:
## Question
@trilinos/ifpack2 
@jhux2 especially

KokkosKernels has a couple of different distance-1 coloring algorithms suitable for GPUs: VBBIT and EB. These are used by Relaxation MTGS/MTSGS. Currently there is no way to choose the algorithm, and the KokkosKernels default for GPUs is EB. Would it make sense to add this a parameter, like "relaxation: mtgs coloring algorithm"?

Here are the tradeoffs:
VBBIT: faster setup and apply. Coloring is faster, and fewer colors used so fewer kernel launches during apply().
EB: more colors, so slightly better preconditioner.

The Ifpack2/Belos tests actually fail with VBBIT because the iterations increase by 1 or 2. If algorithm was a parameter, I would make VBBIT the default but use EB for these tests.

cgcgcg:
I think that this is a pretty reasonable addition. Would it be possible to spell out the differences between the coloring options and their impact on performance in the parameter description (as you did above)? I personally don't have a problem with changing the default.

jhux2:
@brian-kelley I agree, this sounds very reasonable to add.  Thank you.  Feel free to forge ahead!   @csiefer2

csiefer2:
Yup!  No argument here!

PR #10032, opened 2021-12-16
Add coloring algo as option for Ifpack2 MTGS/MTSGS

brian-kelley:
(Address #8441). This is also to avoid random failures of Ifpack2/Belos tests that check the number of iterations.
Use deterministic coloring (VBDBIT) for these tests, so they should now pass every time.

This pulls over the necessary changes from https://github.com/kokkos/kokkos-kernels/pull/1199 , and adds the extra Ifpack2 logic to add the option "relaxation: mtgs coloring algorithm". I added a little explanation of it and the possible choices to the user guide PDF.

@trilinos/ifpack2
@trilinos/kokkos-kernels 

## Motivation

Mainly needed to stop random failures in nightly testing. But also useful to allow users to choose VBBIT over EB for faster setup (see #8441).

## Testing

Tested in the Ifpack2 unit tests for Relaxation, and in ifpack2/test/belos.


ndellingwood (dismissed):
@brian-kelley changes look reasonable to me, please also add a link to the corresponding kokkos-kernels PR with the kokkos-kernels changes for tracking with the next release, thanks!

brian-kelley:
@ndellingwood Done, I added that above (it was https://github.com/kokkos/kokkos-kernels/pull/1199).

srajama1 (dismissed):
Thanks @brian-kelley ! This is a nice addition.

brian-kelley:
@srajama1 @ndellingwood Could one of you re-review? I just added right-hand side vectors in MatrixMarket files to some Ifpack2/Belos problems (instead of generating randomly) so that those solves are deterministic.

PR #10038, opened 2021-12-17
Framework: Python3 build: Fix/move to SEMS-archive TPLs

ZUUL42:
@trilinos/framework 

## Motivation

When the rest of the builds were moved to use sems-archive TPLs, one was missed because it uses a different PullRequestLinuxDriver.sh.
Why it does, that can be addressed at a later time.

## Testing

In testing I realized this PR is going to require we sneak it past the relevant Python3 test. These changes are used before the branch is merged in.
So, to test, I setup [another branch based on this one with an inconsequential comment](https://github.com/ZUUL42/Trilinos/compare/python3_fix...ZUUL42:python3_fix_test) to use at the source and this branch as the target. That builds fine, see [Jenkins](https://ascic-jenkins.sandia.gov/job/trilinos-folder/job/python-3/853/console) & [CDash](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=test-rhel7_sems-gnu-7.2.0-anaconda3-serial_debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_pr-framework&field2=buildstarttime&compare2=84&value2=NOW) (see Experimental "python3_fix" compared to pre 12/16 builds)

@jwillenbring, normally I wouldn't want to sneak something past testing without going over it with the team, but in this case, since it's a quiet Friday and it's only the Python3 test, I think it best to do now instead of during the busy last week before break.
There will be another sneak required for Master Merge, but we can discuss that next week.

### DIFF
--- a/packages/framework/pr_tools/PullRequestLinuxDriver.sh
+++ b/packages/framework/pr_tools/PullRequestLinuxDriver.sh
@@ -35,10 +35,10 @@ function bootstrap_modules() {
         fi
     else
         execute_command_checked "module load apps/anaconda3.7"
-        source /projects/sems/modulefiles/utils/sems-modules-init.sh
-        execute_command_checked "module unload sems-git"
-        execute_command_checked "module unload sems-python"
-        execute_command_checked "module load sems-git/2.10.1"
+        source /projects/sems/modulefiles/utils/sems-archive-modules-init.sh
+        execute_command_checked "module unload sems-archive-git"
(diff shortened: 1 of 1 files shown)


e10harvey (approved):
Thank you, @ZUUL42 !

PR #10045, opened 2021-12-20
Testing on Geminga: Use new SEMS modules

cgcgcg:
@trilinos/muelu 

## Motivation
Currently, all testing on Geminga is broken due to the new SEMS modules. This PR attempts to fix this. I built on geminga using these modules. GCC 5.3.0 is no longer available, so I switched to gcc 8.3.0. I also merged the environment for regular host builds and the CUDA builds.


cmake/ctest/drivers/geminga/cron_driver.sh
```diff
@@ -35,17 +35,24 @@ export TDD_HTTPS_PROXY=$https_proxy
...
-module load sems-gcc/5.3.0
-module load sems-openmpi/1.10.1
-module load sems-superlu/4.3/base
-module load sems-git/2.10.1
+# module load sems-cmake/3.17.1
+# module load sems-gcc/5.3.0
+# module load sems-openmpi/1.10.1
+# module load sems-superlu/4.3/base
+# module load sems-git/2.10.1
+module load muelu-gcc
```
GrahamBenHarper, line 55: since when was there a muelu-gcc module? :open_mouth:

GrahamBenHarper (approved):

cmake/ctest/drivers/geminga/cron_driver.sh
cgcgcg: since I added it in this PR ;-)

PR #10054, opened 2021-12-30
MueLu: build restriction with semicoarsening

rstumin:
Auto-PR for SHA 2b22ec7


rstumin:
@csiefer2 Not sure who is best to review/approve this. I assigned to you. There is one funky thing done here. I created a new factory  MueLu_RfromP_Or_TransP. This is actually a small modification of the usual TransP factory with one small difference. It looks to see if the prolongator factory has set an operator RfromP on the level. If so, it takes this operator instead of transposing P ... otherwise it transposes P. I'm using this in conjunction with the toggle factory ... so on some levels semicoarsening is used while on other levels SA is used. The semicoarsening levels produce an R while the SA ones don't ... so this allows one to either grab an existing R or not depending on whether it exists. Let me know if you have questions. By the way, I've only used this with Tobias-style input (which is what Prospect uses).

csiefer2 (approved):

PR #10061, opened 2022-01-04
Sacado:  Add overloaded function scalarValue() for extracting scalars.

etphipp:
Sacado has always provided ScalarValue<T>::eval() for extracting the
scalar value of any Fad type, but being a class template specialization
isn't always the easiest to call (since it requires explicitly
specifying the type).  This commit adds a template function to make
it easier to call, leveraging the compilers template deduction
mechanisms.  Also added a test for it.


rppawlo (approved):
Thanks @etphipp !

PR #10064, opened 2022-01-05
MueLu: update ssh-agent utility script

jhux2:
Update a MueLu utility that checks for a running ssh-agent.

Support python 2 and 3.  Fix a corner-case bug.

@trilinos/muelu

### DIFF
--- a/packages/muelu/utils/misc/setSshEnv.python
+++ b/packages/muelu/utils/misc/setSshEnv.python
@@ -1,7 +1,13 @@
 #!/usr/bin/env python
+from __future__ import print_function
 import os
 import pwd
-import commands
+import sys
+# support major versions 2 and 3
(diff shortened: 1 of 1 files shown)


packages/muelu/utils/misc/setSshEnv.python
```diff
@@ -62,10 +68,10 @@ for agent in agentList:
         #print "checking key=>"+key+"<"
         if keyFingerprint in key:
           keyFound=1
-          print envCmd + agent
+          print (envCmd + myagent)
```
cgcgcg, line 71: You might need to add `from __future__ import print_function` to the top of the script.

jhux2: done -- python is weird

cgcgcg (approved):

PR #10067, opened 2022-01-07
Testing: sync all the ssh-agent scripts

jhux2:
@trilinos/muelu 

## Motivation

Sync the various copies.


cgcgcg:
Would it make sense to delete all but one of them?

jhux2:
Yes, eventually.  The script would need to updated to accept a key's hash, and all the crontabs then need to be updated to the new syntax.

I actually wonder if using github tokens might make this whole thing easier to maintain, and avoid the dependence on the person who owns the crontab.

cgcgcg (approved):

PR #10068, opened 2022-01-08
Make cron driver Python3 compatible

cgcgcg:
@trilinos/framework 

## Motivation
The cron driver fails since SEMS switched to Python3. Additional changes in TriBITS might be required to get our testing back up, e.g.
https://github.com/trilinos/Trilinos/blob/5a67b70fb3816e1cea26554c650b4dc7ca26747d/cmake/tribits/dashboard_driver/tdd_driver.py#L204-L206


bartlettroscoe (approved):
Sorry, this is not under automated testing.  See https://github.com/TriBITSPub/TriBITS/issues/309#issuecomment-1009026033.

Issue #10080, opened 2022-01-13
Ifpack2: Build fails due to const_cast


nrsharan:
## Bug Report
@trilinos/ifpack2

### Description
While trying to build trilinos/ifpack2, I am met with the following error:

```
In file included from /home/sramesh/dev/myTrilinos/buildTrilinos/packages/ifpack2/src/Ifpack2_Details_FastILU_Base_Serial.cpp:50:
/home/sramesh/dev/myTrilinos/src/Trilinos-trilinos-release-13-2-0/packages/ifpack2/src/Ifpack2_Details_FastILU_Base_def.hpp: In instantiation of ‘void Ifpack2::Details::FastILU_Base<Scalar, LocalOrdinal, GlobalOrdinal, Node>::apply(const Tpetra::MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>&, Tpetra::MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>&, Teuchos::ETransp, Scalar, Scalar) const [with Scalar = double; LocalOrdinal = int; GlobalOrdinal = long long int; Node = Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial>]’:
/home/sramesh/dev/myTrilinos/buildTrilinos/packages/ifpack2/src/Ifpack2_Details_FastILU_Base_Serial.cpp:61:1:   required from here
/home/sramesh/dev/myTrilinos/src/Trilinos-trilinos-release-13-2-0/packages/ifpack2/src/Ifpack2_Details_FastILU_Base_def.hpp:122:56: error: passing ‘const Tpetra::MultiVector<double, int, long long int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >’ as ‘this’ argument discards qualifiers [-fpermissive]
  122 |     auto x2d = X.template getLocalView<execution_space>(Tpetra::Access::ReadWrite);
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/sramesh/dev/myTrilinos/src/Trilinos-trilinos-release-13-2-0/packages/tpetra/core/src/Tpetra_Vector_decl.hpp:50,
                 from /home/sramesh/dev/myTrilinos/buildTrilinos/packages/tpetra/core/src/Tpetra_Vector.hpp:1,
[...]

PR #10084, opened 2022-01-14
Ifpack2 FastILU: Fix MV access

cgcgcg:
@trilinos/ifpack2 @trilinos/tpetra 

## Motivation
I couldn't reproduce #10080 on my system, but the multivector access looks wrong.


jhux2 (approved):
This looks correct to me.

PR #10092, opened 2022-01-17
Tpetra: remove unused device_type typedef in functor

brian-kelley:
In future release of Kokkos this will be checked against the execution
space of the policy, and this was erroring out (policy is on host, but
typedef was for device). See https://github.com/kokkos/kokkos/issues/4677.

@trilinos/tpetra 
@trilinos/kokkos


ndellingwood (approved):
Thanks @brian-kelley

PR #10093, opened 2022-01-17
Changed test regex for a test

bbraunj:
@trilinos/<teamName>

## Motivation

The test regex did not capture the passing status of this test: https://trilinos-cdash.sandia.gov/test/362987


bbraunj:
@e10harvey It's not letting me add a reviewer, but could you review?

kddevin (commented):
Thanks for looking at the Tpetra tests.
This change isn't needed.
The STANDARD_PASS_OUTPUT was printed in the example:  "TEST PASSED"
But output from other ranks interfered with rank 0's printing of "TEST PASSED"; rank 0's "TEST PASSED" was interrupted by other rank's output, so the output file contained "TES" and "T PASSED"
Is this test preventing you from merging other PRs?  If so, we should look at reducing the amount of output from other ranks, or changing how that output is printed to prevent the interruptions.

e10harvey:
> Is this test preventing you from merging other PRs? If so, we should look at reducing the amount of output from other ranks, or changing how that output is printed to prevent the interruptions.

@bbraunj: Thank you very much.

@kddevin: Thank you for looking at this; it sounds like this is an intermittent failure. Yes, this test (in the gcc 8.3.0 build) caused dev2master to fail to fail today. [See CDash](https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&date=2022-01-17&filtercount=1&showfilters=1&field1=site&compare1=63&value1=).

bbraunj:
Sounds like this is something we can close for now. Thanks for taking a look @kddevin!

Issue #10043, opened 2021-12-20
Zoltan2: Test failure for RDC build


jwillenbring:
## Bug Report
@trilinos/zoltan2 

CC: @kddevin 
CC: @trilinos/framework 

### Description

The RDC build had one test failure the last time it ran: Zoltan2_GraphModel_MPI_4. I wasn't able to find exactly why this failed. I would appreciate any insight as to if this is a failure that needs to be fixed, or if this was a one-off issue perhaps.

https://testing.sandia.gov/cdash/test/109158159

e10harvey:
There is also one other RDC failure in teuchos (@trilinos/teuchos): https://trilinos-cdash.sandia.gov/test/104443.

bartlettroscoe:
The test `TeuchosComm_UnitTestHarness_Parallel_UnitTests_MPI_4` does not use CUDA at all so it is doubtful that failures of that test are realted to CUDA RDC issues.  My guess is that this is the system failures reported in #8759.  Please take Teuchos off of this as #8759 covers the issues (which is not actually a Teuchos problem).

kddevin:
The Zoltan2 failure is due only to mis-aligned output from many processors.  PR #10039 is already submitted to try to fix the problem.  The PR needs a review (anyone can review it, please) and it needs to get through PR testing.  I am hopeful that it will fix the problem. @e10harvey

kddevin:
I suspect this is a one-off error.  The test passed, but the output got mangled so that the test harness couldn't find the "PASS" keyword.
If this problem persists, let me know and we'll reduce the amount of output.
@jwillenbring @e10harvey

PR #10101, opened 2022-01-18
Tpetra - giving a choice of little block layout between layout left and layout right

kyungjoo-kim:
@trilinos/tpetra 

## Motivation

In block crs, we currently use layout right (row major) as its default little block layout. Since the little block is an internal object, changing the layout between layout right and layout left should be possible unless users directly access the pointer assuming either of the layout. This PR provides configure option to choose between layout left and layout right for the little block. In my opinion, using the layout left would be more performant in terms of vectorization on both CPUs and GPUs as the little block mat-vec operation consists of vector scaling operations instead of horizontal sum (vector reduce).  

Ifpack2 block tri diag container code directly accesses the little block pointer. Necessary changes are made to respect Tpetra little block layout. 

## Testing

Currently, the layout left is enabled by giving a CFLAGS ``-D TPETRA_ENABLE_BLOCKCRS_LITTLEBLOCK_LAYOUTLEFT`` for evaluation purpose (not a cmake option). I confirmed that layout left passes the PR test.


kyungjoo-kim:
Still testing to see if the code change pass the pr test.

kyungjoo-kim:
@tcclevenger This is the PR that I am expecting some performance benefits for larger block sizes (in particular on CPUs).

kyungjoo-kim:
weird pr test system error ``Error: A license for Comp-CL is not available now ``. Put retest label

kyungjoo-kim:
weird test failure is reported from the latest pr test. the tester looks for "TEST PASSED" and the output of a bug7745 test shows "T PASSED". I am not sure if this is just one time hiccup or not and just re-run the test.

kyungjoo-kim:
@trilinos/tpetra The failure from [TpetraCore_CrsMatrix_Bug7745_MPI_4](https://testing.sandia.gov/cdash/test/115424865) seems to be repeatable. From the output of the test, it passes but it does not seem to print "TEST PASSED". Instead, it prints "T PASSED" at the end. Or the output of the test does not include all verbose output of the test ? Any idea @trilinos/framework ? 

https://testing.sandia.gov/cdash/test/115424865

e10harvey:
> https://github.com/orgs/trilinos/teams/tpetra The failure from [TpetraCore_CrsMatrix_Bug7745_MPI_4](https://testing.sandia.gov/cdash/test/115424865) seems to be repeatable. From the output of the test, it passes but it does not seem to print "TEST PASSED". Instead, it prints "T PASSED" at the end.

Related to https://github.com/trilinos/Trilinos/issues/10043#issuecomment-1015812299.

kyungjoo-kim:
@e10harvey the Cuda failure is not reported in the cdash and it seems to be related with auto-tester.

e10harvey:
> @e10harvey the Cuda failure is not reported in the cdash and it seems to be related with auto-tester.

@kyungjoo-kim: We are working on a fix. See https://github.com/trilinos/Trilinos/issues/10167. The failure stopped the script before cdash submission.

kyungjoo-kim:
@e10harvey Upon the request, I close and reopen a new PR.

PR #10102, opened 2022-01-18
UMR as optional external package in Panzer

bricar:
* UMR and SGM added as optional, external to PackagesList.cmake
* make panzer/adapters-stk optionally depend on UMR
* add UMR unit test within panzer/adapters-stk
* test does UMR on mesh+geometry (2block_cylinders_30deg.g)
* refactor STK_ExodusReaderFactory to support >2 mesh types
* these are: Exodus, Exodus Refinement, Pamgen
* unit test checks final element count after one refinement

@trilinos/panzer

## Motivation
First step to integrate uniform mesh refinement (UMR) library with EMPIRE (via panzer). Uses Sandia Geometry Modeler (SGM) to project new boundary nodes to CAD geometry. 

## Testing
packages/panzer/adapters-stk/test/stk_interface_test/tExodusReaderFactory.cpp


rppawlo (approved):
Looks good to me @bricar ! Thanks! This will be a non-backwards compatible change for the Exodus reader ctor. We may have to update some apps for this. I'll run a quick check on empire. I think Charon and Drekar use the default parameter value so it is ok for them. Albany (@mperego ) might want to test as well.

bricar:
Thanks for reviewing @rppawlo !  Let me know if there are any issues with the non-backwards compatible change for the Exodus reader ctor and I can address.

rppawlo (approved):

PR #10108, opened 2022-01-19
packages/framework:

e10harvey:
- Limit -j to 10 for rdc

@trilinos/framework

## Motivation

For a non incremental build, this reduces the dev2master cuda 10.1.243 rdc build time from 34 hours to 32 hours.

## Testing

This was tested via a jenkins project which runs the cuda 10.1.243 rdc build.


jwillenbring (approved):
Thank you for doing this.

PR #10110, opened 2022-01-20
Piro: add routines to disable initial acceleration solve in Trap. Rule stepper

ikalash:
This PR adds some routines to Piro that enable one to disable the initial acceleration solve within the TrapezoidalRule stepper.  This capability is of interest to Albany-LCM.


lxmota (approved):
Looks good to me

ikalash:
@ccober6 : yes, I know.  This is a separate implementation of a time-stepper for a second-order-in-time ODE (the Trapezoidal Rule) found in Piro.  We are using it for some Arctic erosion problems b/c it has the capability to work with adapted meshes (which we have), whereas Tempus does not, or at least did not when we were first scoping out the implementation a couple years ago.

ccober6:
@ikalash  Ah.  So you have transiently adapted/changing meshes?  Yeah, Tempus does not handle that currently.

ikalash:
Yes, that's right.  @lxmota would be able to provide more information on the adaptation.  Basically we kill off elements once an erosion criterion is triggered.

PR #10115, opened 2022-01-20
Phalanx: sacado memory pool example

rppawlo:
## Motivation

Adding example for sacado memory pool use. dfad temporaries in kernels usually call new/delete from device to allocate derivative arrays. HIP does not support this. Need to switch to memory pool or use extra views for temporaries. This provides an exemplar for the application teams to follow.

## Stakeholder Feedback

N/A

## Testing

new unit test added in PhalanxSacadoMemoryPool.cpp


rppawlo:
Not going to add this capability. After discussions with the kokkos team, they are no longer supporting the memory pool object that this capability is built on. It does not work for HIP backends. Our work around when using dfads was to allocate the temporaries in views on host.

PR #10124, opened 2022-01-25
Geminga: Disable AMGX build

cgcgcg:
@trilinos/muelu 

## Motivation
Disable the AMGX build, since it seems to crash the machine, and the no-UVM build, since it's no longer needed.


jhux2 (dismissed):
lgtm

cgcgcg:
Geminga hasn't crashed in a couple of days, maybe this magically fixed itself?

jhux2 (approved):
lgtm

PR #10125, opened 2022-01-25
SEACAS:  Use pragma once instead of include guards.

gsjaardema:
@trilinos/seacas 

## Motivation

Separated out this change into a single PR to avoid the next seacas snapshot being too full of miscellaneous changes to be able to review cleanly.  This PR is only pragma changes.

### DIFF
--- a/packages/seacas/libraries/ioss/src/Ioss_Assembly.h
+++ b/packages/seacas/libraries/ioss/src/Ioss_Assembly.h
@@ -1,11 +1,10 @@
-// Copyright(C) 1999-2020 National Technology & Engineering Solutions
+// Copyright(C) 1999-2020, 2022 National Technology & Engineering Solutions
 // of Sandia, LLC (NTESS).  Under the terms of Contract DE-NA0003525 with
 // NTESS, the U.S. Government retains certain rights in this software.
 //
 // See packages/seacas/LICENSE for details
 
-#ifndef IOSS_Ioss_Assembly_h
-#define IOSS_Ioss_Assembly_h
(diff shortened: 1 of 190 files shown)


gsjaardema:
@prwolfe @alanw0  Would it be possible for one of you to give a quick review of this.  Minor changes -- just remove include guards and replace with `#pragma once`.  There will be more changes coming later, but this makes later snapshots smaller and easier to review.

This one finally made it through the regression tester...

alanw0 (approved):

gsjaardema:
@alanw0 Thanks.

alanw0:
> @alanw0 Thanks.

No problem. I'll be interested to see how that goes. We've considered using pragma once in stk, but haven't yet. After some reading, we weren't convinced that it was standard, maybe not guaranteed to be portable, etc. Although I know pretty much all compilers (or pre processors) are supporting it. It does seem cleaner than include-guards.

gsjaardema:
@jwillenbring It looked like Autotester finished all tests correctly, but it is now retesting.  Not sure what is going on, but I know you've been having some issues with the autotester...

jwillenbring:
@gsjaardema My mistake. I put a retest flag on it because I didn't see this had completed the testing since we fixed it late last night. I got the flag in there in the time between when the tests passed and the autotester swept past this PR again to see the tests had passed. In any case, the testing went again and this was automerged. Sorry for the confusion.

gsjaardema:
@jwillenbring No problem. Just checking.  Thanks for explanation.

PR #10128, opened 2022-01-26
Tempus:  Allow separating forward and adjoint psuedo-transient evaluations

etphipp:
Added flags to control whether the forward and/or adjoint integrations
are done in advanceTime() for the pseudo-transient adjoint integrator.
Using the update() methods of ROL-2.0, this eliminates the duplicate
forward evaluation when ROL calls value() and gradient() for the
objective function with the same set of parametes, when using the
pseudo-transient adjoint integrator to compute gradients.


ccober6 (approved):
LGTM

PR #10135, opened 2022-01-27
Teuchos: replace curly braces with parens in Watchr

brian-kelley:
Unescaped curly braces are allowed in XML strings, but Watchr has a bug
with parsing them. Replace with parentheses as quick workaround.

Tested on my RHEL machine.

@trilinos/teuchos


kddevin (approved):
Thanks, @brian-kelley, for this fix.  Maybe WATCHR will fix this bug someday, but it will be nice to have a work-around.

Issue #10136, opened 2022-01-28
Ifpack2: RILUK has multiple inefficiencies when calling Kokkos-Kernel's SPILUK


jhux2:
## Enhancement
@trilinos/ifpack2 @srajama1 @csiefer2 @brian-kelley @jennloe @vqd8a 

Problems identified so far:

- There are two rounds of matrix copies, one in initialize() and one in compute.
- The triangular factors are copied back to host.

I'm unsure if https://github.com/kokkos/kokkos-kernels/issues/48, referenced in Ifpack2's triangular solve
https://github.com/trilinos/Trilinos/blob/a66bb9c6fa4df2b13557a9d2b90d9c66119c0c98/packages/ifpack2/src/Ifpack2_LocalSparseTriangularSolver_def.hpp#L758 is still valid.

jhux2:
<img width="1603" alt="Screen Shot 2022-01-27 at 5 08 21 PM" src="https://user-images.githubusercontent.com/12970120/151469258-f6982b95-13d3-4b0d-b89a-6edd40490bca.png">

jhux2:
@vbrunini

PR #10137, opened 2022-01-28
Ifpack2: [WIP] remove unnecessary matrix copy

jhux2:
Cache local CRS matrix created in initialize(), rather than recreating it in compute().
Add additional timers for Additive Schwarz and RILUK.

## Motivation

Partially addresses #10136.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_RILUK_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_RILUK_def.hpp
@@ -503,6 +503,8 @@ void RILUK<MatrixType>::initialize ()
     // we just copy the input matrix if it's not a CrsMatrix.
 
     if (this->isKokkosKernelsSpiluk_) {
+      Teuchos::Time timer2("RILUK::initialize::create_spiluk_handle");
+      Teuchos::TimeMonitor timeMon_kkhandle(timer2);
       this->KernelHandle_ = Teuchos::rcp (new kk_handle_type ());
       KernelHandle_->create_spiluk_handle( KokkosSparse::Experimental::SPILUKAlgorithm::SEQLVLSCHD_TP1, 
                                            A_local_->getNodeNumRows(),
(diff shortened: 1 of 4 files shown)


cgcgcg (approved):
lgtm

packages/ifpack2/src/Ifpack2_RILUK_def.hpp
```diff
@@ -903,9 +932,9 @@ void RILUK<MatrixType>::compute ()
   }
   else {
     {//Make sure values in A is picked up even in case of pattern reuse
-      RCP<const crs_matrix_type> A_local_crs =
-        rcp_dynamic_cast<const crs_matrix_type> (A_local_);
-      if (A_local_crs.is_null ()) {
+      if (A_local_crs_.is_null ()) {
```
vqd8a, line 935: @jhux2 I think this might not work when ```rcp_dynamic_cast<const crs_matrix_type> (A_local_)``` is ```null``` and ```compute()``` is called multiple times (i.e. pattern reuse). A's values might not be updated in this case.

vqd8a: A quick fix could be
```
RCP<const crs_matrix_type> tmp_ptr =  rcp_dynamic_cast<const crs_matrix_type> (A_local_);
if (tmp_ptr.is_null ()) {
.....
}
```

jhux2: @vqd8a I believe the suggested case will always fail, because `A_local_` is itself of type `LocalFilter` and only guaranteed to meet the `RowMatrix` interface.

Thus, we'd be back to the case of copying the matrix twice into CRS format.

Can you explain what you mean by pattern reuse?

vqd8a: @jhux2 Yes, you are right. ```A_local_crs_``` is not ```null``` after ```initialize()```. So when the values of A are updated, the second ```compute()``` will not re-compute ```A_local_crs_``` from A because ```A_local_crs_``` is not ```null```, thus the triangular factors won't change in this case.

Sorry, my suggested fix above is wrong. I think we should check whether compute() has been called or not
```
if (numCompute_>0) {
  A_local_crs_ = rcp_dynamic_cast<const crs_matrix_type> (A_local_);
  if (A_local_crs_.is_null ()) {
        Teuchos::Time timer2("RILUK::initialize::copy_into_crs");
        Teuchos::TimeMonitor timeMon_copy2Crs(timer2);
        local_ordinal_type numRows = A_local_->getNodeNumRows();
        ...
  }
}
```
What do you think?

jhux2: Given the typical application use case below, this needs more consideration.  @vqd8a Your suggestion would work, but I believe would result in frequent matrix copying.

```
for each time:
  for each newton iteration:

    if first_step or mesh_modified: # mesh_modified happens rarely
        preconditioner.initialize()
   
    assemble_matrix_values()
 
    preconditioner.compute()
    linear_solve()
```

vqd8a: @jhux2 Yes, it looks to me we cannot avoid these copies in subsequent ```compute()``` calls

vbrunini: Where does that leave us? The profiling data we've got is pretty clear that redoing these copies on the host every `compute()` call is not viable. Maybe two possible paths?
1. Get the copy to happen on the device at least. I still don't love the additional memory requirement from this, but it at least means not having to change kokkos-kernels.
2. Make it possible for the kokkos-kernels ILU to accept some kind of per-row modification operation to apply so that things like the local filtering & reordering can be done on the fly inside the factorization kernel similar to the old Serial algorithm.

PR #10138, opened 2022-01-28
zoltan:  better link to downloads for stand-alone zoltan

kddevin:
@trilinos/zoltan

## Motivation

The previous link to Zoltan downloads took users to Sandia's github page, which required people to find the release link.
With this PR, users will instead go directly to the release page -- a better experience.

## Testing

NA; documentation only.
I did test that the link was correct.


egboman (approved):
Nice change. Thanks, @kddevin .

PR #10141, opened 2022-01-29
10140 panzer out of bounds error in li ds

### LINKED ISSUE #10140: Panzer: out of bounds error in LIDs
## Bug Report
@trilinos/panzer

### Description
There's a use case of `getElementLIDs(PHX::View<const int*> cellIds, ArrayT lids)` (I saw this come up specifically with the blocked Tpetra gather solution evaluator) that result in a Kokkos out of bounds error. The function fills a local LIDs array with values from the global LIDs, and within the Kokkos functor, the original code loops over the second dimension with
```
       for(int i=0;i<static_cast<int>(local_lids.extent(1));i++)
         local_lids(cell,i) = global_lids(cellIds(cell),i);
``` 
In the corner case, a local LIDs array is passed in that is padded in the second dimension (i.e. `local_lids.extent(1) > global_lids.extent(1)`). There is a comment in `Panzer_GatherSolution_BlockedTpetra_impl.hpp` explaining this:
```
  // We will use one workset lid view for all fields, but has to be
  // sized big enough to hold the largest elementBlockGIDCount in the
  // ProductVector.
```
The problem is that when `local_lids.extent(1) > global_lids.extent(1)`, the above loop goes out of bounds on `global_lids`.

For proper functioning, we should never have `local_lids.extent(1) < global_lids.extent(1)`, so a simple fix is for the loop to run to `global_lids.extent(1)`.

### Steps to Reproduce
Exposed by a test in Drekar run with `KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK=ON`

egphill86:
@trilinos/panzer

## Motivation

Closes #10140

## Stakeholder Feedback

Fixes issue in Drekar 

## Testing

Almost all tests/examples in adapters-stk subpackage exercise this change.

### DIFF
--- a/packages/panzer/dof-mgr/src/Panzer_GlobalIndexer.hpp
+++ b/packages/panzer/dof-mgr/src/Panzer_GlobalIndexer.hpp
@@ -295,7 +295,7 @@ public:
      KOKKOS_INLINE_FUNCTION
      void operator()(const int cell) const
      {
-       for(int i=0;i<static_cast<int>(local_lids.extent(1));i++) 
+       for(int i=0;i<static_cast<int>(global_lids.extent(1));i++) 
          local_lids(cell,i) = global_lids(cellIds(cell),i);
      }
      


rppawlo (approved):
Thanks @egphill86 !

Issue #10139, opened 2022-01-28
Zoltan:  Invalid MPI message tag for intelmpi-2021


nate945:
## Bug Report
@trilinos/zoltan

### Description
We are currently upgrading Sierra from IntelMPI-2019 to IntelMPI-2021.  As part of this we hit an issue in Zoltan around invalid MPI message tags.  It appears when using IntelMPI-2021 the maximum allowable MPI message tag is 524287.  Zoltan is using a tag of up to 1021981, which crashes the Sierra applications.  The MPI standard appears to state that the maximum tag required to be supported is only 32767.  So I don’t think there is any MPI bug here, IntelMPI-2021 just seems to be more restrictive on allowable tags than any of the previous versions.

The problem is at: 
   packages/zoltan/src/matrix/matrix_distribute.c   Line ~ 222

     int msg_tag = 1021982;

When the MPI_IRecv is posted with that tag later on the application will immediately abort with an MPI error.

### Steps to Reproduce
1. SHA1:  [insert here]
1. Configure script:  [attach here]
1. Configure log:  [attach here]
1. Build log:  [attach here]
1. Input deck:  [attach here]
1. Do this.
1. Do that.
1. Shake fist angrily at computer.
1. Run log:  [attach here]

ccober6:
@nate945 FYI, I have found on an Intel oneAPI site the following [link](https://community.intel.com/t5/Intel-oneAPI-HPC-Toolkit/Intel-MPI-maximum-tag-value-is-much-smaller-with-MLX-provider/td-p/1241310), which shows for Intel MPI the maximum tag value has been decreasing, since 2018.  :(

Also the standard states, "..., the tag upper bound value must be _at least_ 32767.", so `msg_tag` should not be less than 32767?

nate945:
@ccober6 I think for future-proofing the msg_tag used by libraries should be 32767 or less because the standard only guarantees that tag values up 0->32767 must be supported.  

It is also possible to run-time query what the maximum message tag is with:
MPI_Comm_get_attr(MPI_COMM_WORLD, MPI_TAG_UB, &tag_ub, &flag);

I observed the number returned there seems to vary wildly on our different platforms.  Most seem to support max int, intelmpi-2021 being the current lowest at ~500,000

ccober6:
I may be misreading it, but the standard reads, 

> Tag Values
Tag values range from 0 to the value returned for MPI_TAG_UB, inclusive. These values are
[...]

PR #10143, opened 2022-01-31
zoltan:  the MPI standard says message tags should have numbers <= 32767

kddevin:
@trilinos/zoltan 

## Motivation

See the title and discussion in #10139

## Stakeholder Feedback

Thanks to @nate945 for reporting and suggesting the fix.


egboman (approved):
A minimal bug fix - just one number. Smallest PR ever!

PR #10148, opened 2022-02-01
Fixes Householder transform so that it is correct for complex numbers

### LINKED ISSUE #10065: Belos:  Using the Belos::BlockGmresSolMgr to solve complex matrix equations, the explicit residuals do not converge
## Question
@trilinos/\<Belos\>

I try to use Belos to solve a simple complex matrix equation A*X=B. Matrix A is a 14th-order complex matrix with all zero imaginary parts. Matrix B on the right hand side of the equation is a complex matrix with 14 rows and 2 columns. I wrote a custom "Tpetra::Operator" subclass to implement the A*X multiplication operation. Set the parameter "Block Size" to the number of columns of matrix B to achieve simultaneous solution.

In the actual test, it is found that when all the imaginary parts of the complex matrix are all zero, the explicit residual and the implicit residual will converge at the same time, but when the imaginary part of the complex matrix B is set to a number that is not zero, such as All are set to 1. When solving, it is found that only the implicit residual converges but the explicit residual does not, and the explicit residual does not decrease with the increase of the number of iteration steps.

Why does this happen? Can you give me some tips for improvement? thanks.

A_matrix：（RowMajor）
(1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (1,0)  (1,0)  (1,0)  (1,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (1,0)  (1,0)  (1,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (7,0)  (1,0)  (1,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (7,0)  (8,0)  (1,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (7,0)  (8,0)  (9,0)  (1,0)   (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (7,0)  (8,0)  (9,0)  (10,0)  (1,0)   (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (7,0)  (8,0)  (9,0)  (10,0)  (11,0)  (1,0)   (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (7,0)  (8,0)  (9,0)  (10,0)  (11,0)  (12,0)  (1,0)   (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (7,0)  (8,0)  (9,0)  (10,0)  (11,0)  (12,0)  (13,0)  (1,0)  
(1,0)  (2,0)  (3,0)  (4,0)  (5,0)  (6,0)  (7,0)  (8,0)  (9,0)  (10,0)  (11,0)  (12,0)  (13,0)  (14,0)  
B_：（ColMajor）
(0,1)  (1,1)  (2,1)  (3,1)  (4,1)  (5,1)  (6,1)  (7,1)  (8,1)  (9,1)  (10,1)  (11,1)  (12,1)  (13,1)  
(1,1)  (2,1)  (3,1)  (4,1)  (5,1)  (6,1)  (7,1)  (8,1)  (9,1)  (10,1)  (11,1)  (12,1)  (13,1)  (14,1)

hkthorn:
@trilinos/belos 

## Motivation

It was reported in issue #10065 that the block Gmres solver would
claim convergence using implicit residuals, yet actual residuals
did not indicate convergence.  This was observed for complex-valued
linear systems and multiple right-hand sides.  

For complex-valued linear systems with multiple right-hand sides,
the code path for the block Gmres solver will go through a
Householder transform that reduces the block Hessenberg matrix to
a triangular matrix.  The implementation of the Householder transform,
implemented directly in the solver (as opposed to through LAPACK)
was not correct for complex arithmetic.  It has been rewritten to
be simpler to understand and correct.
    
An Epetra block Gmres test was improved to enable multiple right-hand
sides, which was used to check that double-precision arithmetic was
performing correctly.  The block Gmres test using basic linear algebra
implementations that reproduces the issue is being committed to ensure
the issue is fixed.  Furthermore, block Krylov subspace debugging
information which was also used to verify this issue is fixed is also
being included in this commit and is hidden by the Debug and OrthoDetails
verbosity flags.  Commented-out code to perform this subspace debugging
has been removed from the pseudo-block Gmres solver, as it is not being
used right now.

## Testing

OSX, gcc 10.x (serial), clang (parallel)


jennloe:
I'll try to look at this tomorrow.  Adding Ichi as another potential reviewer.

packages/belos/src/BelosBlockGmresIter.hpp
```diff
@@ -852,7 +859,7 @@ class BlockGmresIter : virtual public GmresIteration<ScalarType,MV,OP> {
           sigma = blas.DOT( blockSize_, &(*R_)(curDim+j+1,curDim+j),
                             1, &(*z_)(curDim+j+1,i), 1);
           sigma += (*z_)(curDim+j,i);
-          sigma *= beta[curDim+j];
+          sigma *= SCT::conjugate(beta[curDim+j]);
           blas.AXPY(blockSize_, ScalarType(-sigma), &(*R_)(curDim+j+1,curDim+j),
                     1, &(*z_)(curDim+j+1,i), 1);
           (*z_)(curDim+j,i) -= sigma;
```
iyamazaki, line 865: I was wondering why `-sigma` needs to be type-casted to ScalarTyppe.

iyamazaki (approved):
This looks good to me. Thank you for making time for the fix @hkthorn. I was curious why not using LAPACK (which may have more safe-guards?). Is it just more flexible in this way?

packages/belos/src/BelosBlockGmresIter.hpp
hkthorn: @iyamazaki Belos has limited the use of LAPACK across the solvers due to the lack of templated implementations in Teuchos::LAPACK.  For users who want to use ScalarTypes other than the standard float,double,complex<float>, complex<double>, this can limit the number of available iterative solvers.  So, yes, it is just more flexible this way.

hkthorn: @iyamazaki I believe the -sigma type-cast was required by a reported compiler warning.

iyamazaki: Ah, that makes sense. Thank you!!

hkthorn:
@jennloe That's a great question.  The OrthoManager is tested for Epetra in double precision arithmetic and Tpetra for both double precision and complex precision arithmetic (if complex is enabled).  We can add an OrthoManager test for the simple templated linear algebra implementation in belos/test to cover all our bases.

jennloe:
@hkthorn And all of ICGS, IMGS, and DGKS are tested in both real and complex arithmetic?

hkthorn:
@jennloe Yes, both the Epetra and Tpetra OrthoManager tests exercise ICGS, IMGS, and DGKS for double precision arithmetic.  Tpetra exercises ICGS, IMGS, and DGKS for complex<double>.  Is there any plan for TSQR to be an orthogonalization option?  There may be cause for a rank-revealing in multiple RHS solves.

iyamazaki:
I think TSQR is already there. I think it uses Householder just for normalization (I think it uses Gram-Schmidt for projection, but I need to double-check). It would be nice if we could use different schemes for projection and normalization.

hkthorn:
@iyamazaki Sure, TSQR is in the code, but is it enabled by default and regression tested?

iyamazaki:
Ah. No and no, I think. We need to enable it from Tpetra. There are tests for it in Tpetra and Belos, and it passed last time I tried. But I don't think it is tested regularly.

rppawlo:
@trilinos/framework - there seems to be something wrong with the autotester. This PR failed to load the modules:
```
ModuleCmd_Load.c(213):ERROR:105: Unable to locate a modulefile for 'sems-archive-netcdf/4.7.3/parallel'

```

e10harvey:
> @trilinos/framework - there seems to be something wrong with the autotester. This PR failed to load the modules:
> 
> ```
> ModuleCmd_Load.c(213):ERROR:105: Unable to locate a modulefile for 'sems-archive-netcdf/4.7.3/parallel'
> ```

@rppawlo Thanks for reporting this. This should be fixed now.

CC: @jwillenbring

jennloe:
@trilinos/framework This is still failing due to sems modules that won't load. 
@e10harvey @jwillenbring

e10harvey:
> @trilinos/framework This is still failing due to sems modules that won't load.
> @e10harvey @jwillenbring

@jennloe: Thank you for reporting this. We have identified a bug in code that syncs modules to systems where PR testing runs. We are investigating the bug and will report back when we know more.

Related to https://github.com/trilinos/Trilinos/issues/10167.

CC: @rppawlo, @jwillenbring

ZUUL42:
I have a fix in so the correct TPLs are being synced to the cloud nodes again. They should be good to go again.

Issue #10154, opened 2022-02-02
Teuchos:  zero-sized MPI Allreduce fails


kddevin:
## Bug Report
@trilinos/teuchos 

### Description
@nate945 reports:

-    Teuchos::reduceAll is a wrapper around MPI_Allreduce
 -   In this function MPI_Allreduce is run on a custom data type of size some number of bytes that is passed in
 -   In one empty-matrix corner case the number of bytes==0 going into this routine.  This calls the MPI_Allreduce with a zero-size MPI data type which crashes MPI with a divide by zero floating point exception, probably somewhere in the guts of the MPI call it is dividing some data array length by the data array size or something.
-    Just returning from the function if bytes==0 fixes the problem (shown in attached patch).  I think the return should always be safe as best case scenario if bytes==0 the MPI_Allreduce is just an expensive no-op as there is no data being operated on.

PR #10155, opened 2022-02-02
teuchos recommended by @nate945 for #10154

kddevin:
@trilinos/teuchos

## Motivation

New compiler and MPI version doesn't work with zero-sized Allreduce.
See #10154 

## Stakeholder Feedback

@nate945, thank you for contributing this fix

## Testing

Compiles on mac with clang


ccober6 (approved):
LGTM!

PR #10182, opened 2022-02-11
Zoltan2:  remove use of deprecated doPostsAndWaits as required by #10164

kddevin:
@trilinos/zoltan2 

## Motivation

#10164 deprecated Teuchos::ArrayView interfaces to the distributor, favoring Kokkos::View interfaces.
This change affected several parts of the MultiJagged algorithm, as MJ uses Distributors to migrate data to subsets of processors.

In most cases in this PR, I removed creation of ArrayViews from Kokkos::Views.  I ensured that all views passed to the Distributor were HostSpace views, as UVM views are disallowed.  In many cases, fewer copies are needed with the new interface, as we avoid copies from/to ArrayRCPs and host views.

## Testing

Tested on ascicgpu machines:
serial node with and without Tpetra deprecated code
cude node with and without UVM


tasmith4 (approved):
Looks good to me.  I don't love the idea of calling the distributor in a loop (GeometricGenerator.hpp:2362), but I don't have a better idea ... :)

PR #10184, opened 2022-02-14
Zoltan2: Move creation of Models from Problems to Algorithms (during solve/color method)

JacobDomagala:


packages/zoltan2/core/src/input/Zoltan2_Adapter.hpp
```diff
@@ -243,11 +243,31 @@ template <typename User>
...
+  void generateWeightFileOnly(const char* fileprefix,
                               const Teuchos::Comm<int> &comm) const;
 
 };
 
+template <typename User>
+class AdapterWithCoords : public BaseAdapter<User>
+{
+public:
+  virtual void getCoordinatesKokkosView(
```
kddevin, line 255: It probably would make sense to put all getCoordinates* methods into this class (rather than only the Kokkos-based methods)?

packages/zoltan2/core/src/input/Zoltan2_Adapter.hpp
```diff
@@ -243,11 +243,31 @@ template <typename User>
...
   // BaseAdapter does not know about edge info; it writes
   // only the Chaco header and vertex weights (if applicable).
-  void generateWeightFileOnly(const char* fileprefix, 
+  void generateWeightFileOnly(const char* fileprefix,
                               const Teuchos::Comm<int> &comm) const;
 
 };
 
+template <typename User>
+class AdapterWithCoords : public BaseAdapter<User>
```
kddevin, line 252: Is this class used only to reduce code duplication in the adapters?

packages/zoltan2/core/src/input/Zoltan2_Adapter.hpp
```diff
@@ -243,11 +243,31 @@ template <typename User>
...
+  virtual void getCoordinatesKokkosView(
+    Kokkos::View<typename BaseAdapter<User>::scalar_t **, Kokkos::LayoutLeft, typename BaseAdapter<User>::node_t::device_type> &elements) const = 0;
+};
+
+// Forward declare
+template <typename User>
+class VectorAdapter;
+
+template <typename User, typename UserCoord=User>
+class WrapperForCoords : public BaseAdapter<User>
```
kddevin, line 264: Perhaps "AdapterWithCoordsWrapper" would be a clearer name.
Will this inheritance work if the user does NOT have register coordinates with, say, a GraphAdapter?  That is, say a user wants to provide only Graph information, without coordinate information.  Will this code still work?

egboman (approved):
This PR refactors the Zoltan2 design to be more GPU-friendly. Before, the models were created in the Problems and their data structures had to persist. This was problematic as there might be pointers to both host and device data, which would not work without UVM. Now, each Algorithm take an adapter as input and create a Model only as needed and do not hold onto pointers longer than necessary. These changes have been discussed in many meetings over several months. Nice work!

JacobDomagala:
@egboman The missing label is **AT: PRE-TEST INSPECTED**

JacobDomagala:
I don't have access to CDash results, as they're hosted on Sandia server. I can assume that there were some warnings (with -werror flag). I removed the warnings in the latest commit.

egboman:
There were 25 errors, same for all compilers, all running with MPI. However, not all MPI tests failed.
I attach the errors in csv format.

[test-export-1.csv](https://github.com/trilinos/Trilinos/files/9730418/test-export-1.csv)

JacobDomagala:
@egboman Ok let's rerun the testers

egboman:
@JacobDomagala I changed the AT labels so think the tests will run again. Please update if needed.

JacobDomagala:
@egboman I think the `Pre-Test Inspection` has to be added again

egboman:
Feel free to add more labels if needed.

JacobDomagala:
@egboman  I think you need to again approve the changes before it gets merged

egboman (approved):
Minor updates. I approve again.

Issue #9799, opened 2021-10-11
Tpetra:  finalize interface changes and enable deprecated warnings


kddevin:
## Enhancement
@trilinos/tpetra 

Several Tpetra interfaces changed due to removal of the UVM requirement from Tpetra.  To be user-friendly during Tpetra-stack conversion to the new interfaces, we suppressed the deprecation warnings.  @trilinos/tpetra will review the remaining interfaces and decide which need to be deprecated as well.  We will enable deprecation warnings as we go along.

- [x] Deprecations due to UVM removal
- [x] CrsMatrix constructors that take profileType as argument  #9865
- [x] Rsend and Ssend options in Distributor #9852 #9873
- [x] Return packed views always in getNodePackedIndices, getNodeRowPtrs, getAllValues #9885 
- [x] ArrayView interfaces to Distributor #10164  
- [x] methods getNode* are now getLocal* #10189 
- [ ] Graph and Matrix accessors now use access tags #10188

tasmith4:
@csiefer2 guessing this can be closed now since #10552 is merged?

csiefer2:
Sure.

PR #10188, opened 2022-02-14
Tpetra:  make CRS accessors consistent with UVM-free interface and fixed matrix constructor

kddevin:
@trilinos/tpetra 

## Motivation

As part of #9799, the Tpetra team is deprecating some interfaces that rely on Teuchos::ArrayRCP, favoring Kokkos::Views instead.  These changes are consistent with changes to getLocalView required for UVM-free execution.

This PR deprecates getAllValues, getNodePackedIndices, getNodeRowPtrs, and getValuesView, replacing them with getLocalRowPtrsHost/Device, getLocalIndicesHost/Device, and getLocalValuesHost/Device(AccessTag).

Changes are propagated back to downstream packages.  Xpetra interfaces were not changed, but Xpetra calls the new functions.

Also fixed CrsMatrix copy constructor that took CopyOrView argument.  Without UVM, this constructor as implemented would not have worked, as it initialized the copy with a device view from the source matrix, throwing of the use count for the source matrix.  

## Testing

Tested on ascicgpu machines with CUDA and with sems gcc.
Built affected downstream packages and ran many of their tests.

Added new unit test exercising the new functions.


packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
```diff
@@ -1969,12 +1969,15 @@ namespace MueLu {
      auto sym = rcp(new Tpetra::CrsGraphTransposer<LocalOrdinal,GlobalOrdinal,Node>(tpGraph));
      auto tpGraphSym = sym->symmetrize();
 
-     auto rowsSym = tpGraphSym->getNodeRowPtrs();
+     auto colIndsSym =        // temporary; better fix would be change to LWGraph constructor
```
tasmith4, line 1972: maybe the comment should say FIXME, so if we ever get bored we can grep for it along with Mark's other similar comments?

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
```diff
@@ -1494,19 +1494,33 @@ namespace Tpetra {
...
-    /// \brief Get a host view of the row offsets.
-    ///
-    /// \note Please prefer getLocalGraph() to get the row offsets.
+    /// \brief Get a host view of the packed row offsets.
     ///
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
     /// This may return either a copy or a view of the row offsets.
     /// In either case, it will <i>always</i> live in host memory,
     /// never in (CUDA) device memory.
     Teuchos::ArrayRCP<const size_t> getNodeRowPtrs () const;
```
tasmith4, line 1503: does this need a TPETRA_DEPRECATED marking?

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
```diff
@@ -1323,7 +1309,9 @@ namespace Tpetra {
       Kokkos::Compat::persistingView (valuesPacked_wdv.getHostView(Access::ReadOnly));
     values = Teuchos::arcp_reinterpret_cast<const Scalar> (vals);
   }
+#endif // TPETRA_ENABLE_DEPRECATED_CODE
 
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
   template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
```
tasmith4, line 1315: does this also need TPETRA_DEPRECATED?

packages/tpetra/core/test/CrsMatrix/CMakeLists.txt
```diff
@@ -47,6 +47,15 @@ TRIBITS_ADD_EXECUTABLE_AND_TEST(
   STANDARD_PASS_OUTPUT
   )
 
+TRIBITS_ADD_EXECUTABLE_AND_TEST(
+  CrsMatrix_UnitTests5
```
tasmith4, line 51: Is there anything special about why we group these particular tests together, that would let us give a more descriptive name than "5"?

tasmith4 (approved):
thanks @kddevin, minor comments only, shouldn't block this if we need to get it in before the planned minor release tomorrow.

packages/tpetra/core/test/CrsMatrix/CMakeLists.txt
kddevin: Renamed CrsMatrix_UnitTests_LocalAccessors

packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp
kddevin: Good catch.  Added.

packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
kddevin: Good catch.  Added.

packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_def.hpp
kddevin: Done.

rppawlo (approved):
Thanks for updating panzer for these changes @kddevin !

tasmith4 (dismissed):
Thanks @kddevin!

kddevin:
@tasmith4 @rppawlo  This PR has passed testing.  Would you please re-review?  Thank you!

tasmith4 (approved):

rppawlo (approved):

packages/ifpack2/src/Ifpack2_RILUK_def.hpp
```diff
@@ -926,9 +926,10 @@ void RILUK<MatrixType>::compute ()
         A_local_crs_nc->fillComplete (A_local_->getDomainMap (), A_local_->getRangeMap ());
         A_local_crs = rcp_const_cast<const crs_matrix_type> (A_local_crs_nc);
       }
-      A_local_rowmap_  = A_local_crs->getLocalMatrixDevice().graph.row_map;
-      A_local_entries_ = A_local_crs->getLocalMatrixDevice().graph.entries;
-      A_local_values_  = A_local_crs->getLocalValuesView();
+      auto lclMtx = A_local_crs->getLocalMatrixDevice();
+      A_local_rowmap_  = lclMtx.graph.row_map;
+      A_local_entries_ = lclMtx.graph.entries;
+      A_local_values_  = lclMtx.values;
```
jhux2, line 932: @kddevin Thank you for these changes.  For my own education, was this change made in order to reduce data transfer, or was the old approach deprecated and/or result in errors?

kddevin: @jhux2   CrsMatrix used to store a pointer to a pre-constructed KokkosSparse::CrsMatrix, and getLocalMatrix() returned that pointer.  After the UVM changes, CrsMatrix no longer stores a KokkosSparse::CrsMatrix but, rather, builds one on demand when getLocalMatrixDevice/Host is called.  So this change reduces the number of times the KokkosSparse::CrsMatrix is constructed.  The old method would still work, however.  Another alternative would be
```
A_local_rowmap_ = A_local_crs->getLocalRowPtrsDevice();
A_local_entries = A_local_crs->getLocalIndicesDevice();
A_local_values_ = A_local_crs->getLocalValuesDevice(Tpetra::Access::ReadWrite);
```
Note that if A_local_values_ is actually used as read-only in this Ifpack2 case, using the latter pattern with Tpetra::Access::ReadOnly might reduce data transfer, as the modified flag would not be set.

Are you seeing any problems?

PR #10189, opened 2022-02-15
Tpetra:  deprecate functions named getNode*, preferring more consistent getLocal*

kddevin:
@trilinos/tpetra 

## Motivation

Tpetra has many functions getLocal* (e.g., getLocalViewHost, getLocalRowView, getLocalMatrixHost).  Some older functions are named getNode* (e.g., getNodeNumRows, getNodeNumElements).  It is awkward to use both conventions, and clumsy for users.  This PR changes most fiunctions with getNode* to getLocal*.   (Note that #10188 changes a few of these functions as well, so they are excluded from this PR.)

Who wants to review 264 files?  
In truth, many files are touched, only with name changes.
Significant changes are in CrsGraph_decl/def, CrsMatrix_decl/def, BlockCrsMatrix_decl/def and Map_decl/def, in which old versions are deprecated.  RowMatrix and RowGraph now contain default implementations of the old getNode* functions.  So there are only ~10 files requiring serious examination.

Changes are not yet propagated downstream.  

## Testing

ascicgpu machines with and without Tpetra deprecated code enabled.
Tpetra tests pass with and without deprecated code.


csiefer2 (approved):
Wow.  Query replace like a champ.

kddevin:
@e10harvey @ZUUL42 @jwillenbring  @trilinos/framework 

Please, I need some help getting this PR through testing.  The compilation errors I am getting do not appear to be related to the code; rather, they are 
```
icpc: error #10106: Fatal error in /projects/sems/install/rhel7-x86_64/sems/compiler/intel/17.0.1/base/compilers_and_libraries_2017.1.132/linux/bin/intel64/mcpcom, terminated by kill signal
```
which I assume is a problem with the system.  What can be done to get the PR through?  Thanks.

ZUUL42:
@kddevin, that seems to be coming from the SEMS Intel 17.0.1 TPL.
As far as I know we aren't seeing this widespread in many other Intel 17 PR test builds, but I did find one example in #9929. I am looking into a reason. We may be looking at some sort of intermittent issue. In the meantime, I'd just retest it.

PR #10205, opened 2022-02-17
Tpetra: add transferArrived function to DistObject

tasmith4:
@trilinos/tpetra @vbrunini 

## Motivation
Provides a transferArrived function for DistObject, that will return true once the data from transfer started by beginImport etc. has been communicated appropriately and is ready for the unpack and combine step performed by endImport etc.

This is a repost of #9902 to attempt to work around autotester issues.

## Related Issues

* Related to #9251

## Testing
This PR includes unit tests for the new transferArrived function.  There are no specific tests that transferArrived returns false for CrsMatrix when beginImport is only called on one rank, because CrsMatrix uses the "variable number of packets per LID" code path, which requires a blocking communication across all ranks to establish the number of packets each rank should expect to receive.  However, the conditions under which transferArrived returns false are still tested through the MultiVector.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_DistributorActor.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_DistributorActor.cpp
@@ -70,6 +70,14 @@ namespace Details {
     }
   }
 
+  bool DistributorActor::isReady() const {
+    bool result = true;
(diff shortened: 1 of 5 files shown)


kddevin (approved):
Thanks, @tasmith4

PR #10213, opened 2022-02-18
Update 3-Level ParameterLists

roeverf:
@trilinos/shylu

The parameterlists for the 3-level tests were outdated for the recent changes to FROSch, therefore these have been updated in this PR.


searhein (approved):
@roeverf Thanks for fixing!

PR #10223, opened 2022-02-21
MueLu: enable user-supplied eigen-value for SaP

jhux2:
If the user __supplies__ an eigenvalue estimate for Chebyshev smoothing, cache it for later use in prolongator smoothing. (There is already an option to force a new eigenvalue estimate in the SaPFactory.)

This is a bit of corner case, since MueLu already has the capability to cache an eigenvalue estimate __calculated__ by Chebyshev for later use in the SaPFactory.

@trilinos/<teamName>

## Testing

Tested locally on ascicgpu031.


cgcgcg (approved):
Seems reasonable.

Issue #10228, opened 2022-02-22
Stokhos:  Test failures with clang/10, serial build


ndellingwood:
## Bug Report
@trilinos/stokhos 

### Description
The following Stokhos tests failed when testing a Serial, non-MPI build with clang/10:

```
	1070 - Stokhos_SacadoTraitsUnitTest (Subprocess aborted)
	1072 - Stokhos_SacadoPCEUnitTest (Subprocess aborted)
	1073 - Stokhos_SacadoETPCEUnitTest (Subprocess aborted)
	1074 - Stokhos_SacadoPCESerializationTests (Subprocess aborted)
	1075 - Stokhos_SacadoPCECommTests (Subprocess aborted)
	1092 - Stokhos_sacado_example (Subprocess aborted)
```

This was tested using the clang/10 module on the Blake testbed, which has Intel Skylake architecture

### Steps to Reproduce
1. SHA1:  55d7185956289a91f9f75d4510b1a734ed4efd22
2. Environment:
```
module load cmake/3.22.2 clang/10.0.1 openblas/0.3.13/gcc/10.2.0
# clang/10.0.1 auto-loads gcc/10.2.0
```
3. Configuration:
```
cmake \
 -D CMAKE_INSTALL_PREFIX="${TRILINOS_INSTALL_DIR}" \
 -D CMAKE_CXX_STANDARD="14" \
 -D CMAKE_CXX_FLAGS="-g" \
 -D CMAKE_BUILD_TYPE=DEBUG \
\
 -D TPL_ENABLE_MPI=OFF \
\
 -D TPL_ENABLE_BLAS:STRING=ON \
  -D BLAS_LIBRARY_DIRS:FILEPATH=${BLAS_ROOT}/lib \
  -D BLAS_LIBRARY_NAMES:STRING="openblas" \
 -D TPL_ENABLE_LAPACK:STRING=ON \
  -D LAPACK_INCLUDE_DIRS:FILEPATH="${LAPACK_ROOT}/include" \
  -D LAPACK_LIBRARY_DIRS:FILEPATH=${LAPACK_ROOT}/lib \
  -D LAPACK_LIBRARY_NAMES:STRING="openblas" \
\
 -D Trilinos_ENABLE_TESTS=ON \
 -D Trilinos_ENABLE_EXAMPLES=ON \
 -D Trilinos_ENABLE_COMPLEX=ON \
\
 -D Trilinos_ENABLE_Ifpack2=ON \
[...]

PR #10231, opened 2022-02-22
Stokhos:  Fix static initialization

etphipp:
A couple of Stokhos classes have static data members that internally have RCP's
in them.  On blake with the clang compiler, these static data members were
being initialized before some static data used internally within RCP, leading
to exceptions and failed tests.  This removes the static data to workaround
this problem.

For issue #10228.


ndellingwood (approved):
Thanks for fast fix @etphipp !

etphipp:
No worries!  This is very old code, so I have no idea why this came up now.

Issue #9639, opened 2021-08-31
Amesos2:  Tpetra Deprecated function 


JulienLoiseau:
## Question
@trilinos/Amesos2

I am posting this as a question but it might be a bug. 

The function: `get_crs_helper` is declared _tpetra-deprecated_ here: 
https://github.com/trilinos/Trilinos/blob/240c56b57f731146724178341d1a4f87cf590905/packages/amesos2/src/Amesos2_Util.hpp#L965

It is used in a non _tpetra-deprecated_ function `loadA_impl` here:  
https://github.com/trilinos/Trilinos/blob/240c56b57f731146724178341d1a4f87cf590905/packages/amesos2/src/Amesos2_Superludist_def.hpp#L847

Should the function `loadA_impl` be deprecated or the function `get_crs_helper` be replaced by the new `get_crs_helper_kokkos_view`?

Thank you.

cgcgcg:
@trilinos/amesos2

ndellingwood:
Thanks for reporting, `loadA_impl` should not be deprecated, while `get_crs_helper` should likely be replaced by `get_crs_helper_kokkos_view` as you have determined. I didn't catch this while reviewing some recent code updates, thanks for the catch. If you have changes already made locally please feel free to issue a PR and add me as reviewer, otherwise I can put in a PR for the update.

JulienLoiseau:
I don't have the fix, I just compiled with the deprecated flag for now.

etphipp:
FYI, this causes the build of SPARC against Trilinos-develop to fail, since they build with Tpetra deprecated code disabled, so it would be nice if this could be fixed soon.

etphipp:
And I guess there is apparently no automated testing with Tpetra deprecated code disabled, which there probably should be.

rppawlo:
> And I guess there is apparently no automated testing with Tpetra deprecated code disabled, which there probably should be.

@etphipp - We did have this testing in place in PR testing at one point. When Tpetra deprecated the functions, the corresponding changes were not made in all other trilinos packages, so we had to disable that testing option. The required changes to other packages were not trivial in some cases so it was left to individual packages to decide how to best update. @ccober6 - can we add this to the product leaders agenda?

ccober6:
Yeah, I added with #9712.

iyamazaki:
[...]

PR #10232, opened 2022-02-22
Amesos2, Superludist : fixing compilation errors with Tpetra deprecatated code off

iyamazaki:
@trilinos/amesos2 

## Motivation

This PR tries fixing the compilation errors with Amesos2/SuperLU_DIST when Tpetra's deprecated codes are turned off.

## Related Issues

This may fix https://github.com/trilinos/Trilinos/issues/9639.

## Testing

Tested on Weaver with SuperLU_DIST 6.00 and Tpetra deprecated code off.

### DIFF
--- a/packages/amesos2/src/Amesos2_Superludist_decl.hpp
+++ b/packages/amesos2/src/Amesos2_Superludist_decl.hpp
@@ -117,6 +117,10 @@ public:
 
   typedef FunctionMap<Amesos2::Superludist,slu_type>           function_map;
 
+  typedef Kokkos::DefaultHostExecutionSpace HostExecSpaceType;
+  typedef Kokkos::View<SLUD::int_t*, HostExecSpaceType>   host_size_type_array;
+  typedef Kokkos::View<SLUD::int_t*, HostExecSpaceType>   host_ordinal_type_array;
+  typedef Kokkos::View<slu_type*,    HostExecSpaceType>   host_value_type_array;
 
   /// \name Constructor/Destructor methods
   //@{
@@ -308,11 +312,11 @@ private:
 
   // The following Arrays are persisting storage arrays for A, X, and B
   /// Stores the values of the nonzero entries for SuperLU_DIST
-  Teuchos::Array<slu_type> nzvals_;
+  host_value_type_array nzvals_view_;
   /// Stores the row indices of the nonzero entries
-  Teuchos::Array<SLUD::int_t> colind_;
+  host_ordinal_type_array colind_view_;
   /// Stores the location in \c Ai_ and Aval_ that starts row j
-  Teuchos::Array<SLUD::int_t> rowptr_;
+  host_size_type_array rowptr_view_;
   /// 1D store for B values
   mutable Teuchos::Array<slu_type> bvals_;
   /// 1D store for X values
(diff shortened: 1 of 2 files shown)


ndellingwood (commented):
https://github.com/trilinos/Trilinos/pull/10232/files#diff-0a8bf616832dd789cc5141b1be485931c13053f878184be7a688f29bd96946f6R530

Should packages/amesos2/src/Amesos2_Superludist_def.hpp line 533 (old) typedef also be converted to `get_1d_copy_helper_kokkos_view`? Assuming so, the `do_get` call would need an additional boolean as first argument

packages/amesos2/src/Amesos2_Superludist_decl.hpp
```diff
@@ -308,11 +312,11 @@ class Superludist : public SolverCore<Amesos2::Superludist, Matrix, Vector>
...
-  Teuchos::Array<slu_type> nzvals_;
+  host_value_type_array nzvals_view_;
   /// Stores the row indices of the nonzero entries
-  Teuchos::Array<SLUD::int_t> colind_;
+  host_ordinal_type_array colind_view_;
   /// Stores the location in \c Ai_ and Aval_ that starts row j
-  Teuchos::Array<SLUD::int_t> rowptr_;
+  host_size_type_array rowptr_view_;
   /// 1D store for B values
   mutable Teuchos::Array<slu_type> bvals_;
```
ndellingwood, line 321: Don't the bvals_ and xvals_ need to be converted to the Kokkos::View equivalents as well?

iyamazaki: It looks like we still have this interface for MultiVector. For MultiVector, Tpetra is copying the numerical values from input to output vectors, while for CrsMatrix, it is doing a shallow-copy and needs to use Kokkos::View? We could double-check with team Tpetra. I'll just tag @brian-kelley, for now..

brian-kelley: @iyamazaki If you replaced bvals and xvals with Kokkos views as well (with the correct type), then the Amesos2 utility ``get_1d_copy_helper::do_get`` would also be able to shallow copy. Right now, with them being Arrays, it is always a deep copy. But it's hard to tell at a glance if shallow copying will cause issues with Tpetra's view reference counting.

iyamazaki: Thank you, @brian-kelley & @ndellingwood. I am thinking to keep what we have for now (it sounds safer and probably behaves similar to what we used to have with deprecated codes). What do you think?

brian-kelley: I think it's fine to leave it as is. Copying a vector is probably really cheap compared to the Superlu solve anyway.

ndellingwood: That's fine then to leave as is for now. Will need the `get_1d_copy_helper_kokkos_view::do_get` (also requires an extra boolean as first arg) version when updating to use Kokkos::Views for bvals_ and xvals_

ndellingwood (approved):

Issue #9289, opened 2021-06-15
Tpetra:  Change offset_type to local_ordinal_type


kddevin:
## Enhancement
@trilinos/tpetra 

Currently, Tpetra's offset_type (for indexing into column indices and values array) is size_t.  CuSparse, however, requires offset_type and local_ordinal_type to match.  Thus, Tpetra has to create and store a copy of the offsets array with type local_ordinal_type to use CuSparse (see #9276).  This extra storage has size numLocalRows+1.

@trilinos/tpetra should consider changing offset_type to local_ordinal_type.

Advantages:
- extra storage and extra code path not needed for CuSparse
- casting from size_t to local_ordinal_type no longer needed. 

Disadvantages:
- limits number of nonzeros in an MPI rank to std::numeric_values<local_ordinal_type>::max() nonzeros, even though local_ordinal_type is intended to index rows and columns, not nonzeros 

As an aside, there is no testing of the case mentioned in the disadvantages; that is, no test exercises more than 2B nonzeros in a rank.  So we don't know whether that capability would actually work with the solvers.

A second aside:  if needed, we could add and test local_ordinal_type = long long to accommodate large local problems.

@csiefer2 @rppawlo @srajama1 @iyamazaki  Do you see other advantages / disadvantages of such a change?

csiefer2:
Ripping out all of the casts and the manual uses of size_t will be time consuming --- size_t explicitly shows up 2244 times in MueLu alone.  Not all of those need to be replaced, though.

brian-kelley:
@csiefer2 I grepped through muelu and the vast majority of size_t uses won't need to change. It'll only be the loops that look like this which need to change:
```
for(size_t j = rowptrs(row); j < rowptrs(row + 1); j++)
{
//do something with entries(j)
}
```
and MueLu looks to be using the local graph/matrix typedef for offsets (at least sometimes), for example NotayAggregation does this:
```
using size_type         = typename local_graph_type::size_type;
...
    for(int rowIdx = 0; rowIdx < numRows; ++rowIdx) {
      for(size_type entryIdx = origRowPtr_h(rowIdx); entryIdx < origRowPtr_h(rowIdx + 1); ++entryIdx) {
...
```

PR #10233, opened 2022-02-23
Tpetra:  first step toward #9289 -- DO NOT MERGE YET

kddevin:
@trilinos/tpetra 

## Motivation

This PR is a first step toward #9289.

I added a new default type: Tpetra::Details::DefaultTypes::offset_type
For now, it is set to size_t, to maintain backward compatibility with current code.  
Every place in Tpetra that creates offset arrays now uses this default type instead of size_t.

To implement #9289, change this default offset_type to local_ordinal_type.

Caveats:
-  deprecated function getAllValues (and, perhaps, other deprecated functions) has not been retrofitted to use the offset_type, as we cannot cast views of size_t to be viest of local_ordinal_type.  Rather than retrofit deprecated functions, changing to local_ordinal_type can be delayed until deprecated code is removed.
- using Tpetra::Details::DefaultTypes::offset_type in the Tpetra API exposes a Details member -- bad form.  We should either use local_ordinal_type or add an offset_type typedef to RowGraph so that users don't have to type "Details::"
- Details::DefaultTypes::offset_type is easy to grep and replace with a better interface

Unfortunately, changing the offset_type to local_ordinal_type didn't really speed up SpMV on SerialNode as hoped.  Its benefit should be most seen when using CUSparse as an additional offset array will not be needed.  See #9289

These changes were more work than I expected, so I didn't want to just throw them away after my timing experiment.  @trilinos/tpetra can decide how to proceed with them.

## Stakeholder Feedback

@brian-kelley 
## Testing

Tested with SerialNode on ascicgpu machines, with offset_type = local_ordinal_type and offset_type = size_t.

The Tpetra tests build and pass in both cases -- yea!
With offset_type = local_ordinal_type, many compiler warnings about comparing signed and unsigned are emitted.  Some code cleanup will be needed.

### DIFF
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_decl.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_decl.hpp
@@ -567,7 +567,7 @@ struct AddKernels
   typedef typename map_type::local_map_type local_map_type;
   typedef typename Kokkos::View<GlobalOrdinal*, device_type> global_col_inds_array;
   typedef Kokkos::RangePolicy<execution_space> range_type;
-  typedef KokkosKernels::Experimental::KokkosKernelsHandle<size_t, LocalOrdinal, impl_scalar_type,
+  typedef KokkosKernels::Experimental::KokkosKernelsHandle<Details::DefaultTypes::offset_type, LocalOrdinal, impl_scalar_type,
               execution_space, memory_space, memory_space> KKH;
 
   /// \brief Given two matrices in CRS format, return their sum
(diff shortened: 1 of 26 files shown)


kddevin:
@csiefer2 @brian-kelley  
This PR is a first step toward #9289.  I made the change to see whether having offset type = int would speed up the small matrices for Xyce.  It didn't, but it didn't hurt anything, either.  As you can see, it did propagate through the code quite a bit, so perhaps this PR will be useful to you if you decide to change the type for cusparse.

ccober6:
@csiefer2 should this be closed?

PR #10245, opened 2022-02-25
Tpetra:  fixed used of deprecated code

kddevin:
@trilinos/tpetra 

## Motivation

PR #10189 merged after #10188.  This PR removes deprecations in #10189 from the new test in #10188.  

It also deprecates some code in Details namespace that will not be needed once deprecated code is removed.  Details::localRowOffsets was called only by other deprecated code.

## Testing

Compiled on ascicgpu machines, sems gcc7.2 modules, with and without Tpetra deprecated code enabled.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_localRowOffsets_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_localRowOffsets_def.hpp
@@ -187,10 +187,6 @@ template LocalRowOffsetsResult<NT> \
 localRowOffsets (const RowGraph<LO, GO, NT>& A); \
 }
 
-#else  // !TPETRA_ENABLE_DEPRECATED_CODE
-
-#define TPETRA_DETAILS_LOCALROWOFFSETS_INSTANT(LO, GO, NT)
-
 #endif // TPETRA_ENABLE_DEPRECATED_CODE
 
 #endif // TPETRA_DETAILS_LOCALROWOFFSETS_DEF_HPP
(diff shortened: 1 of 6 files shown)


packages/tpetra/core/src/Tpetra_Details_localRowOffsets_def.hpp
```diff
@@ -182,4 +187,10 @@ template LocalRowOffsetsResult<NT> \
 localRowOffsets (const RowGraph<LO, GO, NT>& A); \
 }
 
+#else  // !TPETRA_ENABLE_DEPRECATED_CODE
```
tasmith4, line 190: why do we need this else block?  In particular, it seems to imply that some non-deprecated code depends on this definition, which would seem to conflict with the comment at line 55 that this whole file can be deleted when deprecated code is removed.

kddevin: @tasmith4 This is a great question.  I couldn't find the answer last night, but I think I found it today.  (I have better grep skills this morning.)

The explicit instantiation system kept failing with deprecated code turned off.  It appeared to be looking for the definition of macro TPETRA_DETAILS_LOCALROWOFFSETS_INSTANT even though we don't need any of the code in the file anymore.  Having an empty definition of the macro allowed compilation with deprecated code off.  

The problem was that the explicit instantiation system continued to generate Tpetra_Details_localRowOffsets_INT_LONG_LONG_SERIAL.cpp even though all the code is commented out.

Oddly, a similar fix wasn't needed for Tpetra_Details_localDeepCopyRowMatrix.def.  With better grep skills this morning, I found the CMake magic to disable the explicit instantiation.  I'll make the change this morning and update the PR for review.

Thanks for pushing me to look at this problem again.

kddevin:
@tasmith4  I pushed the correct fix.  I dislike the ETI system -- simple code changes touch too many parts of the code.

PR #10263, opened 2022-03-01
MueLu: region MG solver

lucbv:
Adding a new interface to solver the region problem, this interface accepts A, X and B in composite format and still
needs the hierarchy in region format. With such an interface hooking the preconditioner with Belos should be a lot easier
although we need to be careful how the composite vectors are transfered to region format.

Next step will be adding a CG solver using composite vectors and compostie A but preconditioned with region MG.

@trilinos/muelu


mayrmt:
@lucbv I pushed some changes. In particular, I believe that the update of `beta_new` has been missing. Also zeroing out `Z` before entering the V-cycle seems the right thing to do. Let me know, how these changes affect your tests.

lucbv:
@mayrmt @pohm01 
I pushed my small changes to make things a little clearer.
Please let me know if you have any comments? Otherwise let's merge this puppy!

packages/muelu/research/regionMG/src/SolveRegionHierarchy_def.hpp
```diff
@@ -624,7 +626,7 @@ void solveCompositeProblemPCG(const double tol, const bool scaleResidualHist, co
     {
       normRes = Res->norm2();
 
-      if(cycle == 0) { normResIni = normRes; }// out << "NormResIni = " << normResIni << std::endl;}
+      // if(cycle == 0) { normResIni = normRes; }// out << "NormResIni = " << normResIni << std::endl;}
```
mayrmt, line 629: I'd suggest to actually delete that line.
```suggestion
```

packages/muelu/research/regionMG/src/SolveRegionHierarchy_def.hpp
```diff
@@ -309,6 +309,61 @@ void vCycle(const int l, ///< ID of current level
...
+  RCP<Vector> regInterfaceScalings = level0->Get<RCP<Vector> >("regInterfaceScalings");
+  RCP<Matrix> regMat               = level0->Get<RCP<Matrix> >("A");
+  RCP<const Map> revisedRowMap     = regMat->getRowMap();
+
+  // Compute region vectors for B and X
+  RCP<Vector> quasiRegB;
+  RCP<Vector> regB;
+  compositeToRegional(B, quasiRegB, regB,
+                      revisedRowMap, rowImport);
+  // scaleInterfaceDOFs(regB, regInterfaceScalings, true);
```
mayrmt, line 350:
```suggestion
```

packages/muelu/research/regionMG/test/structured/Driver_Structured_Regions.cpp
```diff
@@ -430,7 +431,7 @@ int main_(Teuchos::CommandLineProcessor &clp, Xpetra::UnderlyingLib& lib, int ar
     procsPerDim[2] = galeriList.get<GO>("mz");
   }
 
-  const LO numLocalCompositeNodes = lNodesPerDim[0]*lNodesPerDim[1]*lNodesPerDim[2];
+  // const LO numLocalCompositeNodes = lNodesPerDim[0]*lNodesPerDim[1]*lNodesPerDim[2];
```
mayrmt, line 434:
```suggestion
```

packages/muelu/research/regionMG/test/structured/Driver_Structured_Regions.cpp
```diff
@@ -597,8 +598,10 @@ int main_(Teuchos::CommandLineProcessor &clp, Xpetra::UnderlyingLib& lib, int ar
                      revisedRowMap, revisedColMap,
                      rowImport, quasiRegionMats, regionMats);
 
-  // We don't need the composite operator on the fine level anymore. Free it!
-  A = Teuchos::null;
+  // Actually for now we are keeping it so we can implement
+  // iterative solvers using composite operator and vectors.
+  // // We don't need the composite operator on the fine level anymore. Free it!
+  // A = Teuchos::null;
```
mayrmt, line 604:
```suggestion
```

mayrmt (changes requested):
@lucbv Just some minor typos and mainly my request to remove commented lines. Otherwise, we will wonder in the future, why these are commented and if they might be needed again (which they do not).

mayrmt:
@lucbv I took the liberty to add some tests for the new solver options. Please let me know, what you think.

mayrmt:
@pohm01 @rstumin Please have a look!

rstumin:
@mayrmt @lucbv Just a quick question about the numLevels parameter to the vCycleAdapter? Is there anyway this can be eliminated by deducing something from the passed in region hierarchy? I also feel that vCycleAdapter might not be a good name ... if someone can use this to actually do a Wcycle? Maybe something like MGcycleAdapter or MGIterateAdapter would be better?

mayrmt:
@rstumin I did rename the adapter function to `RegionMgCycleAdapter()`. I'll look into the `numlevels` parameter.

rstumin:
@mayrmt @lucbv I'm a bit confused on the function vCycle(). Does it also do Wcycles? I see the cycleCount loop which makes me think that it does do W cycles ... but then I see that something is excluded on the finest grid??? The standard way to incorporate V and W cycles would be to not have the smoothing in the cycleCount loop. Instead, you have  1) presmooth, 2) for (cycle=0; cycle < cycleCount; cycle++) { recursiveCall() applied to coarse problem } 3) postsmooth . Is there a reason we did it the way it is done now with the smoothers inside the cycle count loop?

mayrmt:
@rstumin You're right. The `vCycle()` function can also perform W-cycles. Hence, I have renamed it to `MgCycle()` and have also updated its documentation and names of its internal `Teuchos::TimeMonitor` objects.

rstumin:
@mayrmt @lucbv Just to be explicit on this .. is the W-cycle a true W-cycle? Will it do 2 solves of the level 1 equation to correct the level 0 solve?

mayrmt:
@rstumin I looked into the W-cycle and I think that it performs two solves on level 1 to correct level 0.

rstumin:
@mayrmt @lucbv Are you sure about the W cycle. Looking at line 79 of SolveRegionHierarchy_def.hpp, I see
if (cycleType == "W" && levelID > 0) // W cycle and not on finest level
    cycleCount = 2;
which I think means that cycleCount is 1 when levelID = 0 ... which I think means that there is only one coarse solve at this level? The problem with setting cycleCount to 2 here is that it will also apply the smoothing twice on the finest level, which is not what we want. I hate to be stubborn ... but this doesn't seem right. If it is right, there should be a comment explaining it? As I indicated earlier, the more standard thing would have been to have the cycleCount loop only applied to the coarse grid correction and not the smoothers.

pohm01:
@lucbv @mayrmt I think @rstumin is correct here, and I overlooked the extra smoothing steps when I originally implemented the W.

mayrmt:
@rstumin Great eye for the details! Let's figure this out and fix it.

The extra smoothing steps on level 0 could be avoided by adding some more logic to the implementation, right? @pohm01, could you maybe look into this?

mayrmt:
As there seem to be some more open questions, I have removed the AUTOMERGE label.

mayrmt:
@rstumin @lucbv @pohm01 I did some more cleanup and have added a regression test for the W-cycle. Please let me know what you think.

mayrmt:
@lucbv @rstumin @pohm01 Csn you have a look and review this PR? I think it's time to move forward on this.

rstumin:
@mayrmt Sorry for the delay. I think that we now do not want the levelID check in line 79 anymore. 'if (cycleType == "W" && levelID > 0) // W cycle and not on finest level'. I think it is supposed to be    ' if (cycleType == "W" ) // W cycle ' . I suspect that previously it was there to avoid extra smoothing on the finest level ... but now that we don't have the fine level smoothing relying on cycleCount ... we don't need this anymore.

rstumin:
@mayrmt One other little comment is that if a direct solver is used on the coarsest level ... I think we will now solve it twice with the W cycle .... so we might want to not set cycleCount to 2 if we are on the 2nd to last level.

pohm01:
@rstumin I've fixed the W cycle to only do one cycle on the second to last level when the last level is a direct solve.

mayrmt:
@rstumin Please have a (hopefully) final look and approve, if ok.

rstumin (dismissed):
This looks good to me. The only thing that I might consider is either renaming or putting comments associated with solveRegionProblem(). We now have 3 solve procedures ... which might be confusing to users/developers. The solvers with CG and Richardson seem pretty clear ... but solveRegionProblem() might be not so clear. If solveRegionProblem() is really just our old way of doing things .... maybe we should put the word 'Old' in its name or have some comments saying this. Anyway, I don't think this is a big deal ... so I approve.

mayrmt:
@rstumin I have renamed the function and have added comments to clarify the difference between the three solver routines. Please re-approve, if this is ok with you.

mayrmt:
Is this 

> Cannot Merge Pull Request# 10263: Github Mergeability = blocked; Git Mergeable = blocked

an issue of the autoteseter or is there an actual git merge conflict?

mayrmt:
@cgcgcg Thanks!

PR #10283, opened 2022-03-04
Update muemex/mlmex library links to new CMake syntax

brian-kelley:
(that has the PUBLIC/PRIVATE/INTERFACE keyword before the library name)

This is because elsewhere Tribits uses the new syntax with these target executables,
and CMake doesn't let you mix old and new.

@trilinos/muelu 
@trilinos/ml


jhux2 (approved):
lgtm

dridzal:
Thanks!

PR #10284, opened 2022-03-04
Fix build of MortarSurfaceCoupling with complex

brian-kelley:
@trilinos/muelu 

## Motivation

While doing something unrelated, I noticed this test had a build failure when complex is enabled. I fixed that, and then made these tests only run with float/double, since they need to read real-valued MatrixMarket files.


jhux2 (approved):
ltgm.  Thanks, @brian-kelley!

brian-kelley:
@jhux2 Thanks for reviewing. I just added another commit so that these tests actually run if both real and complex scalars are enabled (in the first commit, the test was a no-op in that case).

jhux2 (approved):
Even better!

PR #10253, opened 2022-02-26
 Kokkos + KokkosKernels Promotion To 3.6.0

ndellingwood:
@trilinos/kokkos 
@trilinos/kokkos-kernels 

## Motivation
Release 3.6.0 of Kokkos Ecosystem

## Testing
Autotester plus regular nightly testing

PR #10289, opened 2022-03-07
Add fallthrough comment so Sierra compiles with warnings as errors

tvoskui:
@trilinos/kokkos

## Motivation
So Sierra can include pod_hash and build without errors when we have "warnings are errors" on.

### DIFF
--- a/packages/kokkos/containers/src/impl/Kokkos_Functional_impl.hpp
+++ b/packages/kokkos/containers/src/impl/Kokkos_Functional_impl.hpp
@@ -107,7 +107,9 @@ uint32_t MurmurHash3_x86_32(const void* key, int len, uint32_t seed) {
 
   switch (len & 3) {
     case 3: k1 ^= tail[2] << 16;
+      // fall through
     case 2: k1 ^= tail[1] << 8;
+      // fall through
     case 1:
       k1 ^= tail[0];
       k1 *= c1;


jhux2:
@tvoskui If it hasn't already been, this should be applied against Kokkos proper, as well.  @trilinos/kokkos

tvoskui:
> @tvoskui If it hasn't already been, this should be applied against Kokkos proper, as well. @trilinos/kokkos

Yep, there is a PR in the Kokkos repo to fix this too.

jhux2:
@tvoskui https://github.com/kokkos/kokkos/pull/4847 and this one differ substantially, am I missing something?

tvoskui:
> @tvoskui [kokkos/kokkos#4847](https://github.com/kokkos/kokkos/pull/4847) and this one differ substantially, am I missing something?

There was some iteration on that PR - originally it was just adding the fall through comments but the reviewers requested a non-switch form instead and now can't agree on whether that's what they want or not. Jan is handling that one so we can just wait to see what form it ends up in. I don't know when/how the upstream Kokkos changes are integrated into Trilinos - if they just override these than it shouldn't be a problem to use the "quick fix" here with the comments and then override it later with whatever form they end up with in Kokkos. 

I added a patch in Spack in Sierra to fix this so I'll let you decide what to do with the PR.

ndellingwood:
After iteration on the kokkos side it looks like this issue was resolved with PR kokkos/kokkos#4852
Will cherry-pick those changes to the 3.6 release candidate for snapshot into Trilinos, which will be part of PR #10253

Issue #10113, opened 2022-01-20
Trilinos failing to compile on Summit (gcc 10.2.0 cuda 11.3.1)


PaulMullowney:
SHA: dd53a1e0f
Using gcc/10.2.0 and cuda/11.3.1

[ 80%] Building CXX object packages/stk/stk_expreval/stk_expreval/CMakeFiles/stk_expreval.dir/Function.cpp.o
/gpfs/alpine/cfd116/scratch/mullowne/Trilinos/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h: In instantiation of 'const std::vector<double>& Iotm::text_mesh::Coordinates<EntityId, Topology>::operator[](EntityId) const [with EntityId = long int; Topology = Iotm::TopologyMapEntry]':
/gpfs/alpine/cfd116/scratch/mullowne/Trilinos/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C:370:59:   required from here
/gpfs/alpine/cfd116/scratch/mullowne/Trilinos/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h:567:13: error: '__T0' was not declared in this scope; did you mean '__y0'?
  567 |         auto it(m_nodalCoords.find(nodeId));
      |         ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |         __y0

Has anyone seen this before?
-Paul

alanw0:
This build output seems to be a little garbled, as stk_expreval doesn't include nor have anything to do with Iotm_TextMeshUtils.h.
However, perhaps @tokusanya can take a look at the TextMesh issue.

jhux2:
@PaulMullowney I'm able to compile Trilinos develop https://github.com/trilinos/Trilinos/commit/a5a5bf9a6f118b6bfc886e697e6c1a10c355b68f (with the patch https://github.com/kokkos/kokkos/pull/4673 applied) on Summit with gcc 10.2.0 and CUDA 11.3.1.

PaulMullowney:
With a serial build:

[ 31%] Building CXX object packages/seacas/libraries/ioss/src/text_mesh/CMakeFiles/Iotm.dir/Iotm_TextMesh.C.o
[...]

PR #10290, opened 2022-03-07
SEACAS: STK: compile fixes under cuda 11.4.2 gcc 10.2

jhux2:
@trilinos/seacas
@alanw0 @tasmith4 

## Motivation
Fixes compilation error on Summit.  See #10113.


alanw0 (approved):
This looks fine to me.
We'll need to separately put this stk change in the sierra repo which is the source of stk-trilinos snapshots in general.

gsjaardema (approved):

Issue #10308, opened 2022-03-09
Tpetra:  Distributor::doPosts needs unique tags


tasmith4:
## Bug Report
@trilinos/tpetra 

### Description
`Distributor::doPosts` launches a number of MPI send/receive requests, and can be called asynchronously (i.e., multiple doPosts calls from different Distributors can be active at the same time).  Since doPosts uses hard-coded tags that are constant for each code path, there is potential for messages to become mixed up (and in fact we do notice strange failures with asynchronous doPosts calls, see #10300).

Based on the discussion in #10296, we need to implement a strategy for using unique tags across separate invocations of doPosts.

tasmith4:
As demonstrated in #10300, we actually don't need unique tags, almost certainly due to the ordering guarantees in the MPI standard.

PR #10300, opened 2022-03-08
Tpetra:  Added a test for multiple asynchronous transfers

tasmith4:
@trilinos/tpetra 

## Motivation
Test that we can do asynchronous transfers across many MultiVectors at the same time (use case for #9251).

## Related issues
* Closes #9251 

## Testing
Now working on ascicgpu in the following builds:  CPU, Cuda UVM, Cuda no-UVM

PR #10310, opened 2022-03-09
Tpetra: Added a test for multiple asynchronous transfers with cyclic maps

tasmith4:
@trilinos/tpetra 

## Motivation
@kddevin and I noticed that when !importsAreAliased() and CUDA_AWARE_MPI is defined, we get the wrong answer in multiple asynchronous transfers.  In many-to-one transfer, imported items from other proc appear to not be inserted in the target multivector, while copies on target proc work.  This test does multiple asynchronous transfers with a cyclic source map, which has !importsAreAliased().

## Related Issues

* Is blocked by #10308, #10300

## Testing
Currently failing in a Cuda UVM build on ascicgpu.


tasmith4:
combining with #10300

PR #10320, opened 2022-03-11
Fix 10294

### LINKED ISSUE #10294: Intrepid2 test fails without UVM
## Bug Report
@trilinos/intrepid2 

Intrepid2_unit-test_MonolithicExecutable_Intrepid2_Tests.exe test fails
```
96. Legendre_DerivativesMatchJacobi_UnitTest ... [Passed] (0.0382 sec)
97. ProjectedGeometry_CircularGeometryConvergesInP_SingleCell_UnitTest ... error for p = 1: 1.14159
Kokkos::View ERROR: attempt to access inaccessible memory spaceAborted (core dumped)
```

### Description

When ``Kokkos_ENABLE_CUDA_UVM=OFF``, the test fails with attempting invalid memory space.

CamelliaDPG:
@trilinos/intrepid2

## Motivation
See #10294.  Some of the ProjectionTools code relied on UVM, some of it in subtle ways: certain `create_mirror_view_and_copy()` calls were failing here with an error indicating that a temporary allocation would be required to do the copy.  In this PR, I replace those with a two-stage copy: device subview to device view, followed by device view to host.

Closes #10294.

## Testing
Have confirmed that all MonolithicExecutable tests pass under CUDA when built without KokkosKernels and with `Kokkos_ENABLE_CUDA_UVM=OFF` (tried this on Weaver).

### DIFF
--- a/packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
+++ b/packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
@@ -790,6 +790,9 @@ public:
         elemDofCopy(i) = elemDof(i);
       });
       auto serialElemDof = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), elemDofCopy);
+      
+      Kokkos::DynRankView<typename ViewType2::value_type, HostDeviceType> A_host("A0_host", elemMat.extent(1),elemMat.extent(2));
+      auto A_device = Kokkos::create_mirror_view(typename DeviceType::memory_space(), A_host);
 
       if(matrixIndependentOfCell_) {
         ViewType2 elemRhsTrans("transRhs", elemRhs.extent(1), elemRhs.extent(0));
@@ -798,8 +801,6 @@ public:
 
         Kokkos::View<valueType**,Kokkos::LayoutLeft,HostDeviceType> serialElemRhs("serialElemRhs", n+m, numCells);
 
-        Kokkos::DynRankView<typename ViewType2::value_type, HostDeviceType> A_host("A0_host", elemMat.extent(1),elemMat.extent(2));
-        auto A_device = Kokkos::create_mirror_view(typename DeviceType::memory_space(), A_host);
         Kokkos::deep_copy(A_device, Kokkos::subview(elemMat, 0, Kokkos::ALL(), Kokkos::ALL()));
         Kokkos::deep_copy(A_host, A_device);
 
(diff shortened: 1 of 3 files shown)


CamelliaDPG:
(Turned off automerge because I will want to do a squash and merge; I have some extraneous commits here.)

packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
```diff
@@ -780,15 +780,27 @@ class ProjectionTools {
       Teuchos::LAPACK<ordinal_type,valueType> lapack_;
       ordinal_type numCells = basisCoeffs.extent(0);
 
+      // due to the particular type of elemDof (possibly due either to layout or subviews having been
+      // taken), the usual pattern of create_mirror_view_and_copy() does not work "out of the box".
+      // So we do a manual copy on device, followed by the create_mirror_view_and_copy().
+      Kokkos::View<ordinal_type*> elemDofCopy("elemDofCopy", elemDof.extent(0));
```
kyungjoo-kim, line 786: We do not allocate a view with default space. All Intrepid2 is templated with device type and the view allocation should respect the device type.

packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
```diff
@@ -780,15 +780,27 @@ class ProjectionTools {
...
+      Kokkos::parallel_for ("Intrepid2::ProjectionTools - copy elemDof",
+                            Kokkos::RangePolicy<ExecSpaceType, int> (0, elemDof.extent(0)),
+      KOKKOS_LAMBDA (const unsigned &i) {
+        elemDofCopy(i) = elemDof(i);
+      });
+      auto serialElemDof = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), elemDofCopy);
+      
+      Kokkos::DynRankView<typename ViewType2::value_type, HostDeviceType> A_host("A0_host", elemMat.extent(1),elemMat.extent(2));
+      auto A_device = Kokkos::create_mirror_view(typename DeviceType::memory_space(), A_host);
+
```
kyungjoo-kim, line 796: I don't understand why you need this copy assignment.

packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
```diff
@@ -780,15 +780,27 @@ class ProjectionTools {
...
       if(matrixIndependentOfCell_) {
         ViewType2 elemRhsTrans("transRhs", elemRhs.extent(1), elemRhs.extent(0));
         Kokkos::View<valueType**,Kokkos::LayoutLeft,HostDeviceType>
         pivVec("pivVec", m+n + std::max(m+n, numCells), 1);
 
         Kokkos::View<valueType**,Kokkos::LayoutLeft,HostDeviceType> serialElemRhs("serialElemRhs", n+m, numCells);
 
-        Kokkos::DynRankView<typename ViewType2::value_type, HostDeviceType> A_host("A0_host", elemMat.extent(1),elemMat.extent(2));
-        auto A_device = Kokkos::create_mirror_view(typename DeviceType::memory_space(), A_host);
```
kyungjoo-kim, line 791: When the layout does not match here, use the derived layout from device type. We don't need to run the manual copy.

kyungjoo-kim: Also when we create the A_host, it can be created appropriately as a mirror view having the same memory layout of elemMat

packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
```diff
@@ -816,20 +828,27 @@ class ProjectionTools {
...
-              basisCoeffs_);
+          Kokkos::deep_copy(A_device, Kokkos::subview(elemMat, ic, Kokkos::ALL(), Kokkos::ALL()));
+          Kokkos::deep_copy(A_host, A_device);
+          auto A = A_host;
+          Kokkos::deep_copy(b_device, Kokkos::subview(elemRhs, ic, Kokkos::ALL()));
+          Kokkos::deep_copy(b_host, b_device);
+          auto b = b_host;
+          auto basisCoeffsSubview = Kokkos::subview(basisCoeffs, ic, Kokkos::ALL());
+          Kokkos::deep_copy(basisCoeffsSubviewCopy_device, basisCoeffsSubview);
+          Kokkos::deep_copy(basisCoeffsSubviewCopy_host, basisCoeffsSubviewCopy_device);
```
kyungjoo-kim, line 851: Do we really need multiple deep copies in the loop of Cells ? Too much overhead and consider a different refactoring approach.

kyungjoo-kim: @mperego This might impose some inherent issue in designing the projection tools. Would you check and address the UVM issue on the projection tools from the class design ?

kyungjoo-kim (changes requested):
Even if this resolves the issue, I cannot agree that the fix is performant. There should be a different way to address the uvm issue.

packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
CamelliaDPG: @kyungjoo-kim, I'm not sure I follow.  Could you give a code snippet?

CamelliaDPG:
> Even if this resolves the issue, I cannot agree that the fix is performant. There should be a different way to address the uvm issue.

Could be; this is what I was able to come up with through trial and error.  This particular code path is only run when built without Kokkos Kernels; I was assuming that this is basically there for compatibility, not particularly expected to be performant.

Please let me know if you have some suggestion for how you'd prefer to address this.

mperego:
I'll look at it early next week. 
Do we really need to maintain the searial code that does not depend on Kokkos Kernels? Or should we simply make the dependence on Kokkos Kernels not optional?

CamelliaDPG:
> I'll look at it early next week.
> Do we really need to maintain the searial code that does not depend on Kokkos Kernels? Or should we simply make the dependence on Kokkos Kernels not optional?

I would be in favor of requiring Kokkos Kernels; I'm thinking about another approach for sum factorization that might benefit from Kokkos Kernels, and it would help me if I could assume the dependency.

kyungjoo-kim:
I would like to keep the kokkos kernels library as an optional package. I think that you encounter this issue while you try to fix the monolithic unit test. I will check a better solution for the projection tools and you can fix the other uvm issue on the example with enabling kokkos kernels in your test.

CamelliaDPG:
Closing in favor of PR #10333.

PR #10325, opened 2022-03-14
packages/framework: Add LaunchDriver.py

e10harvey:
@trilinos/framework

## Motivation
Create a system-agnostic proxy script based on GenConfig dependencies for launching pull request drivers.

## Related Issues
TRILFRAME-262

* Closes TRILFRAME-264

## Testing
This was tested manually by running gcc-7.2.0 and cuda-10.1.243.

## TODOs
- [x] Add unit and integration tests


e10harvey:
@ZUUL42, @jwillenbring, @fryeguy52: this is ready for a review.

jwillenbring (approved):
This seems fine to me. I am not a Python expert.

packages/framework/pr_tools/PullRequestLinuxDriver.sh
```diff
@@ -4,37 +4,31 @@ SCRIPTPATH=$(dirname $SCRIPTFILE)
...
-            echo -e "Job is CUDA node is vortex"
-            execute_command_checked "module load git/2.20.0"
-            execute_command_checked "module load python/3.7.2"
-            get_python_packages pip3
-            envvar_set_or_create PYTHON_EXE python3
-        else
-            message_std "PRDriver> " "ERROR: Unable to find matching environment for CUDA job not on Ride."
-            exit -1
-        fi
+    if [[ ${NODE_NAME:?} =~ ${vortex_regex} || ${on_ats2} == "1" ]]; then
```
ZUUL42, line 19: Does ats2 not use the SEMS NFS?

e10harvey: @fryeguy52 do you know?

ZUUL42: That's the only thing that catches my eye, but I don't really know anything about ats2. I just noticed it would not see the source /projects/sems/modulefiles/utils/sems-archive-modules-init.sh line in the else block. (soon to be non-archive) I only know of ride, vortex, and now weaver to do that, but ats2 could just as easily be in that same situation.

Looking a bit deeper into it, I'm seeing ats2 builds being cuda or using compiler versions that aren't sems or sems-archive, so they must use some other source. So it should be fine. Just wanted to check.

ZUUL42 (approved):
Was curious about ats2 ifs, but after looking into it, I think that's fine.

PR #10328, opened 2022-03-14
Xpetra/MueLu: binary matrix reading

cgcgcg:

## Motivation
- Xpetra: Add a unit test for reading of matrices with missing rows
- MueLu: Fix MM->binary converted for use case with missing rows


packages/muelu/utils/matrix/ascii2binary.cpp
```diff
@@ -47,27 +47,23 @@ int main(int argc, char* argv[]) {
...
-        int rownnz = inds.size();
-        ofs.write(reinterpret_cast<char*>(&row),    sizeof(row));
-        ofs.write(reinterpret_cast<char*>(&rownnz), sizeof(rownnz));
-        for (int k = 0; k < rownnz; k++) ofs.write(reinterpret_cast<char*>(&inds[0] + k), sizeof(inds[k]));
-        for (int k = 0; k < rownnz; k++) ofs.write(reinterpret_cast<char*>(&vals[0] + k), sizeof(vals[k]));
-
+    for (int row = 0; row < m; row++) {
+      inds.resize(0);
+      vals.resize(0);
+      while (row == i) {
```
jhux2, line 53: Does this assume the matrix is stored with rows altogether and in strictly ascending order?

cgcgcg: Yes, it does. But I don't think that that assumption has changed from before.

jhux2: The original code assumes all the values in a row are contiguously stored, but I think the rows can be in any order.  If I understand the logic of the original code, if a new row is encountered, the check `while (row == i ...` is false, but `row==i+1` isn't necessarily true.

cgcgcg: Ah, bugger. Yes, that's correct. I guess in that case, the simplest way around all these restrictions is to read the matrix using the Tpetra MM reader, sort it, and then dump to binary.

jhux2: What do you think of this hacky idea.  Keep a boolean array of length num_rows.  As a row is processed, mark its entry true.  After all rows have been processed, for all rows with false entries, add a "-1" entry in the binary file.

cgcgcg: yup, that'll work.

jhux2 (approved):
Lgtm, thanks for fixing this!

packages/xpetra/test/IO/IO_UnitTests.cpp
```diff
@@ -0,0 +1,180 @@
...
+namespace {
+
+  TEUCHOS_UNIT_TEST_TEMPLATE_6_DECL( IO, MMMissingRows, M, MA, Scalar, LO, GO, Node )
+  {
+
+    // get a comm and node
+    Teuchos::RCP<const Teuchos::Comm<int> > comm = Xpetra::DefaultPlatform::getDefaultPlatform().getComm();
+    TEUCHOS_ASSERT_EQUALITY(comm->getSize(), 1);
+
+    if (Teuchos::ScalarTraits<Scalar>::isComplex)
```
jhux2, line 63: I approve!

jhux2 (approved):
lgtm

PR #10344, opened 2022-03-17
PyTrilinos: Fix Tpetra bindings when compiled with -DTpetra_INST_INT_…

Sbte:
…LONG_LONG:BOOL=OFF.

@trilinos/pytrilinos
@trilinos/tpetra

## Motivation

Importing PyTrilinos fails with
```
ImportError: /mnt/fvm-env/install/lib/libpytrilinos.so.13: undefined symbol: _ZNK6Tpetra3MapIixN6Kokkos6Compat23KokkosDeviceWrapperNodeINS1_6SerialENS1_9HostSpaceEEEE19getLocalElementListEv 
``` 
without this patch. 

I don't know if this is the right fix, so I would appreciate any feedback. I don't use Tpetra with PyTrilinos, and I can't compile the tests, so I have no way of actually testing this, or maybe other configurations with other `-DTpetra_INST_*` configurations that also need fixing.

## Testing

I've been unable to compile the PyTrilinos tests:
```
[ 98%] Generating testutil.py
/bin/sh: 1: ../util/copyWithCMakeSubstitutions.py: not found
make[2]: *** [packages/PyTrilinos/test/CMakeFiles/PyTrilinos_Test_testutil.dir/build.make:73: packages/PyTrilinos/test/testutil.py] Error 127
make[1]: *** [CMakeFiles/Makefile2:16564: packages/PyTrilinos/test/CMakeFiles/PyTrilinos_Test_testutil.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```
But that does not seem to be a problem with the patch.

### DIFF
--- a/packages/PyTrilinos/cmake/PyTrilinos_config.h.in
+++ b/packages/PyTrilinos/cmake/PyTrilinos_config.h.in
@@ -123,7 +123,12 @@
 
 /* PyTrilinos ordinal types */
 
+#ifdef HAVE_TPETRA_INST_INT_LONG_LONG
 #define PYTRILINOS_GLOBAL_ORD long long
+#else
+#define PYTRILINOS_GLOBAL_ORD int
+#endif
+
 #define PYTRILINOS_LOCAL_ORD  int
 
 /**********************************************************************/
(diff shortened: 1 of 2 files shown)


jhux2:
@trilinos/pytrilinos @trilinos/tpetra

pyt-viper:
I will take a look at this Monday

pyt-viper (approved):
Tpetra support in PyTrilinos is still experimental, so these changed files are only accessed if you are willing to be a part of the experiment.  The native Python integer type is `long long` so it is nice to use that if it is available.  These protections appear to do it correctly.

jhux2:
@pyt-viper Thanks for your review.

PR #10354, opened 2022-03-22
Zoltan2: SphynxProblem inherits now from PartitioningProblem

mperrinel:


cgcgcg:
@trilinos/zoltan2

egboman (approved):
This looks good. The key feature is that the SphynxProblem now inherits from the PartitioningProblem, which simplifies things both for users and developers. Parameter validation had to change a bit as the Sphynx parameters are only known to the Zoltan2Sphynx package, which remains an optional package separate from Zoltan2Core.
Thanks, @mperrinel !

PR #10357, opened 2022-03-22
Panzer: Use HGRAD_*_Cn instead of HGrad_*_C2

cgcgcg:
@trilinos/panzer @trilinos/intrepid2 

## Motivation
HGRAD_*_C2 is missing some method implementations, and Intrepid2 would like to deprecate the quadratic elements in favor of n-th order elements.


mperego (approved):
Thanks! Looks good to me.

mperego:
> One concern. When I refactor the code long time ago, I remembered that a Sandia application uses (assumed) the basis points for the C2 elements. I don't remember that they use through Panzer or not. The ordering of high order element (n=2) basis points is different from one used in C2. Some visualization code may also have some issues if the code assumes the local basis ordering as used in C2 instead of C(n=2).

Yes. @rppawlo checked that Drekar (the only code we are aware of that uses quadratic elements) works fine.

rppawlo (approved):

PR #10370, opened 2022-03-25
Ifpack2: add nnz for ILU factor descriptions

jhux2:
@trilinos/ifpack2

## Motivation

Users may want to know the number of nonzeros in the triangular factors.

## Stakeholder Feedback

N/A

## Testing

I verified that the upper/lower nonzero counts were correct for a small two-process MueLu test problem.


packages/ifpack2/src/Ifpack2_LocalSparseTriangularSolver_def.hpp
```diff
@@ -47,6 +47,7 @@
 #include "Tpetra_Core.hpp"
 #include "Teuchos_StandardParameterEntryValidators.hpp"
 #include "Tpetra_Details_determineLocalTriangularStructure.hpp"
+#include "MatrixMarket_Tpetra.hpp"
```
cgcgcg, line 50: Why?

jhux2: fixede

cgcgcg (dismissed):

cgcgcg (approved):

PR #10374, opened 2022-03-27
Panzer: use HGrad_*_Cn basis with n=2, instead of HGrad_*_C2.

mperego:
Continuing work of PR #10357, and have Panzer use HGrad Cn basis instead of C2 basis.
In contrast to Cn basis, C2 basis dofs are ordered as the nodes of Shards quadratic topology geometry and it's best to limit their use ONLY for mappings for quadratic meshes.

@trilinos/panzer

### DIFF
--- a/packages/panzer/adapters-stk/test/stk_connmngr/tCubeHexMeshDOFManager.cpp
+++ b/packages/panzer/adapters-stk/test/stk_connmngr/tCubeHexMeshDOFManager.cpp
@@ -55,7 +55,6 @@
 #include "Panzer_STKConnManager.hpp"
 
 #include "Intrepid2_HGRAD_HEX_C1_FEM.hpp"
-#include "Intrepid2_HGRAD_HEX_C2_FEM.hpp"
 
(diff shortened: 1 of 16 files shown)


mperego:
@rppawlo I tried to replace C2 with Cn in [tFieldAggPattern.cpp ](https://github.com/trilinos/Trilinos/blob/master/packages/panzer/dof-mgr/test/dofmngr_test/tFieldAggPattern.cpp) but I didn't understand it well and didn't want to change the hardcoded values. Is the `GeometricAggFieldPattern` supposed to work with any basis? Also I found this comment that's a bit concerning:
https://github.com/trilinos/Trilinos/blob/7bc4eb697c4ccd9fab1742661da3260a1069d41a/packages/panzer/dof-mgr/test/dofmngr_test/tFieldAggPattern.cpp#L459

rppawlo:
@mperego - GeomAggregateFieldPattern must work for all bases. Maybe we can have a quick discussion about this today?  Commit #10357 has broken many empire tests this weekend.

Issue #10191, opened 2022-02-15
Trilinos 13.4 release status


jwillenbring:
@trilinos/developers 

We are currently waiting on the 13.4 minor release until we resolve the issues associated with PRs that are not yet in develop.

bartlettroscoe:
@jwillenbring, any update on the ETA for when the Trilinos 13.4 release branch might get cut?  I am wanting on that before I can merge PR #9978 which we know will break backward compatibility (and will straight-up break some builds and require users to take specific actions to resolve the issues).

BTW: the release notes documentation for this change in backward compatibility and how users can deal with it is described in an updated Trilinos/RELEASE_NOTES item [here](https://github.com/bartlettroscoe/Trilinos/blob/3281be9b5af148f16d9f507519d3e06b916b675c/RELEASE_NOTES#L7-L73).

If any Trilinos developers would be willing to look at that documentation and provide feedback in that PR, it would be appreciated.

jwillenbring:
@bartlettroscoe The PRs that we were waiting on got in, but then last week there was a flurry of PRs to fix issues with things that were deprecated in the first set of PRs. We are going to reevaluate this week. There are a couple failing MueLu tests yet, but I am not yet certain if those will be fixed for the minor release.

cgcgcg:
@jwillenbring I was only aware of one failing MueLu test. Can you point me to a listing?

jwillenbring:
@cgcgcg I don't have a list, but I asked @csiefer2 who said that you have a PR up that might fix the issues.

cgcgcg:
NVM then :-) I already talked to @csiefer2.

bartlettroscoe:
Any updates on this?   We are still waiting to merge PR #9978 till after the Trilinos 13.4 release branch is created off of 'master'.  I would love to get that merged before more TriBITS changes get applied (that will also break backward compatibility in very rare cases).

jwillenbring:
@ccober6 
@rppawlo 
@srajama1 
@mperego 
@csiefer2 
@trilinos/framework 

I went through the errors in the deprecated code off build today:

https://testing.sandia.gov/cdash/viewBuildError.php?buildid=10764652

Here are some representative failures from each package I saw issues with:

stokhos:
[...]

PR #10379, opened 2022-03-29
ROL: fix calls to Tpetra deprecated code and disable PDE-OPT/0ld tests

mperego:
Note: I only tried to fix calls to Tpetra deprecated code in the most straightforward way. There is likely room for improving performance. Maybe this can be addressed when refactoring the code to be UVM free.

@trilinos/rol

## Motivation

Related to #10191


dridzal (approved):
The changes look reasonable.  The PDE-OPT directory 0ld no longer requires building and testing, and it can be archived.  The sysbuilder and PinT capabilities are not widely used at the moment, but it is important to convert them because of future needs.  Thanks @mperego !

mperego:
> 

B.t.w. why is it called `0ld` instead of `Old`?

PR #10385, opened 2022-03-30
Enables code for multi-threaded block RHS solves based on Trilinos_ENABLE_OpenMP

hkthorn:
@trilinos/shylu 

## Motivation
This code has been placed behind a non-standard ifdef KLU_ENABLE_OPENMP that was only known
and used by Xyce when configuring the hybrid-hybrid ShyLU solver.  In an effort to more
easily enable and evaluate this solver for customer circuits, a transition is being made
to enable this code using a more standardized approach, which uses Trilinos_ENABLE_OpenMP.
The TrilinosSS package now has a package-specific ifdef that is enabled whenever Trilinos
if built with OpenMP and enables this code.  It will allow for easier integration of the
hybrid-hybrid ShyLU solver.

The motivation for doing this is supported by the discussion in the Jira issue:

https://sems-atlassian-son.sandia.gov/jira/browse/TRILINOS-203

## Stakeholder Feedback

This has been tested by the Xyce stakeholder to ensure it simplifies the configure script that enables the hybrid-hybrid ShyLU solver.  

## Testing

OSX GNU v11 w/ and w/o OpenMP threading

### DIFF
--- a/packages/common/auxiliarySoftware/SuiteSparse/src/CMakeLists.txt
+++ b/packages/common/auxiliarySoftware/SuiteSparse/src/CMakeLists.txt
@@ -1,6 +1,25 @@
+######################################################################
+## Define the header and source files (and directories)
+######################################################################
+
 SET(HEADERS "")
 SET(SOURCES "")
 SET(IMPORTEDLIBS "")
+
+######################################################################
+## Package-specific configuration options
+######################################################################
+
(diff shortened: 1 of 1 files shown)


srajama1 (dismissed):
I have discussed this with @hkthorn ! Thanks Heidi for these changes.

Issue #10390, opened 2022-03-31
Tpetra: Test Failure


csiefer2:
Kokkos integration test is building wonky.  And also having some issues running.

As reported by @jennloe

csiefer2:
Fixed

PR #10391, opened 2022-03-31
Tpetra: Fixing #10390

csiefer2:
Issue: #10390 
Tests: Fixing it!
Stakeholder feedback:  Communication from @jennloe


packages/tpetra/core/test/KokkosIntegration/ApiTest.cpp
```diff
@@ -77,12 +77,12 @@ void ApiTest::incr(std::string key) {
 }
 
 void ApiTest::printAll() {
-  fprintf (stderr, "**************** cuda call analysis ****************\n"); 
-  fprintf (stderr, "call                                           times\n");
+  //  fprintf (stderr, "**************** cuda call analysis ****************\n");
```
tasmith4, line 80: we can just delete these if we don't need them any more, right?

packages/tpetra/core/test/KokkosIntegration/KokkosUVM.cpp
```diff
@@ -135,7 +135,7 @@ void reportTimings(Teuchos::RCP<Teuchos::StackedTimer> timer) {
 }
 
 int main(int argc, char *argv[]) {
-  Kokkos::ScopeGuard kokkos(argc, argv);
+  Kokkos::initialize(argc, argv);
```
tasmith4, line 138: why doesn't Kokkos::ScopeGuard work anymore?

packages/tpetra/core/test/KokkosIntegration/KokkosDeepcopy.cpp
```diff
@@ -51,18 +54,22 @@
...
+  void run(int argc, char *argv[])  {
     log("Tpetra Kokkos DeepCopy Regression test start");
 
     std::string cudaSync("cudaDeviceSynchronize");
     std::string cudaMemcpy("cudaMemcpy");
     std::string cudaMemcpyAsync("cudaMemcpyAsync");
     ApiTest *counter = ApiTest::getInstance();
     //initialize
+#ifdef HAVE_TPETRACORE_MPI
     MPI_Init(&argc, &argv);
```
tasmith4, line 71: don't we have a Tpetra::ScopeGuard that can take care of some of this logic for us?  But also, shouldn't this initialize/finalize stuff get called in main instead?

packages/tpetra/core/test/KokkosIntegration/KokkosDeepcopy.cpp
```diff
@@ -240,17 +248,28 @@ class DeepCopyTester {
 
 private:
   bool isConsistent;
-  void log(const std::string& msg) {
+  void log(const std::string& msg, ApiTest *counter=0) {
```
tasmith4, line 251: I'd strongly prefer to use nullptr as the default

tasmith4 (changes requested):

packages/tpetra/core/test/KokkosIntegration/KokkosUVM.cpp
csiefer2: It probably does, actually.  I've just never used it before and removed it trying to debug something else.

packages/tpetra/core/test/KokkosIntegration/KokkosDeepcopy.cpp
csiefer2: Yeah, all of this initialize / finalize is an artifact of the way this was written.

packages/tpetra/core/test/KokkosIntegration/KokkosDeepcopy.cpp
csiefer2: Sure, why not.

Issue #10389, opened 2022-03-31
Stokhos: compilation error when linking with Intel One API MKL version >= 2021


maartenarnst:
## Enhancement
@trilinos/stokhos

A compilation error occurs when Stokhos is compiled with Trilinos linked with Intel One API MKL version >= 2021.  The compilation error comes from the file ```packages/stokhos/src/Stokhos_SDMUtils.hpp```.  This file declares a function DGEQP3_F77 that MKL is also declaring and defining.  However, starting from Intel One API MKL version >= 2021, MKL is including a noexcept specifier, which Stokhos doesn't.

We have implemented the following patch:

```
From 2e79cdacdd876a7a864b7f44a669220212084712 Mon Sep 17 00:00:00 2001
From: Maarten Arnst <maarten.arnst@uliege.be>
Date: Fri, 18 Mar 2022 15:38:45 +0100
Subject: [PATCH] Patch solves compilation error observed when linking with
 Intel One API MKL version >= 2021.

---
 packages/stokhos/src/Stokhos_SDMUtils.hpp | 14 +++++++++++++-
 1 file changed, 13 insertions(+), 1 deletion(-)

diff --git a/packages/stokhos/src/Stokhos_SDMUtils.hpp b/packages/stokhos/src/Stokhos_SDMUtils.hpp
index d93319320aa..9dcc5858554 100644
--- a/packages/stokhos/src/Stokhos_SDMUtils.hpp
+++ b/packages/stokhos/src/Stokhos_SDMUtils.hpp
@@ -50,9 +50,21 @@
 #include <ostream>
 
 #define DGEQP3_F77  F77_BLAS_MANGLE(dgeqp3,DGEQP3)
+
 extern "C" {
+#if defined(HAVE_STOKHOS_MKL)
+    #include "mkl_version.h"
+    #if __INTEL_MKL__ >= 2021
+        #define MKL_NO_EXCEPT noexcept
+    #else
+    	#define MKL_NO_EXCEPT
+    #endif
+#else 
+    #define MKL_NO_EXCEPT
+#endif
+
 void DGEQP3_F77(const int*, const int*, double*, const int*, int*,
-                double*, double*, const int*, int*);
+            double*, double*, const int*, int*) MKL_NO_EXCEPT;
 }
 
 #include "Stokhos_ConfigDefs.h"
-- 
2.25.1
```

jhux2:
@trilinos/stokhos

etphipp:
[...]

PR #10393, opened 2022-03-31
Stokhos: Fixing #10389

maartenarnst:
@trilinos/stokhos

This patch solves #10389: Stokhos: compilation error when linking with Intel One API MKL version >= 2021.


maartenarnst:
Hi @etphipp!  Should we do something to try to move this PR forward through the testing?

etphipp:
Let's see what the autotester returns.  The issues we were seeing before should be fixed now.

PR #10397, opened 2022-04-01
Ifpack2: add StackedTimer/Watchr to Ifpack2_tif_belos

brian-kelley:
Adds command-line options ``--with_stacked_timer`` (default: without)
and ``--problem_name="XYZ"`` to set the top-level timer name
(this will also be the top-level Watchr plot title).

Default behavior is unchanged, although I did change some existing
timer names for clarity: "init" -> "Init preconditioner" and
"precond" -> "Compute preconditioner".

I had to refactor things a bit to make StackedTimer include the precond setup, but exclude the I/O (which often dominates the total time). I combined the two functions that were in build_problem.hpp, and then moved the few lines for precond creation into main().

@trilinos/ifpack2 

## Motivation

SierraTF would like nightly performance testing of some problems, and this driver is perfect for doing that, it just needed the Watchr support.

### DIFF
--- a/packages/ifpack2/test/belos/CMakeLists.txt
+++ b/packages/ifpack2/test/belos/CMakeLists.txt
@@ -7,6 +7,7 @@ ADD_SUBDIRECTORY(AdditiveSchwarzRILUK)
 IF (Tpetra_INST_DOUBLE)
   TRIBITS_ADD_EXECUTABLE(
     tif_belos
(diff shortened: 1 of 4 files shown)


jhux2 (approved):
lgtm, thanks @brian-kelley

jhux2:
Wow, this passed on the first try!

brian-kelley:
@jhux2 I know, couldn't belive it :P

PR #10403, opened 2022-04-04
MiniEM: Add higher order solver

cgcgcg:
@trilinos/panzer 

## Motivation
Add a multigrid solver for higher order discretizations to MiniEM.


cgcgcg:
Replaced by #10530

PR #10415, opened 2022-04-06
Tpetra: Use numVectors==1 specialization for local parts of norms

cgcgcg:
@trilinos/tpetra 

## Motivation
Use the KokkosKernels specialization for 1-vector MVs for the local parts of norms.

For a run on 32 GPUs, I see 

**Before**
```
Belos::MVT::MvNorm: 1.4864 - 3.84499% [3604] {min=1.47393, max=1.49435, std dev=0.00603344} <1, 4, 1, 1, 3, 3, 7, 3, 1, 8>
```
**After**
```
Belos::MVT::MvNorm: 0.19496 - 0.527566% [3604] {min=0.177702, max=0.205975, std dev=0.00839741} <2, 2, 2, 1, 5, 5, 0, 2, 9, 4>
```

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
@@ -151,18 +151,37 @@ lclNormImpl (const RV& normsOut,
   }
   else { // lclNumRows != 0
     if (constantStride) {
-      if (whichNorm == NORM_INF) {
-        KokkosBlas::nrminf (normsOut, X);
-      }
-      else if (whichNorm == NORM_ONE) {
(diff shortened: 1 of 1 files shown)


brian-kelley (approved):
@cgcgcg KokkosKernels is supposed to have this path built in (so you can always call the rank-2 nrm functions and get the best performance), but I accidentally removed those paths while fixing a different issue. I put them back in https://github.com/kokkos/kokkos-kernels/pull/1354, so they are fixed in KokkosKernels develop. I'm just waiting for the 3.6.0 promotion to merge so I can patch it in.

But until then, this is a good fix.

PR #10419, opened 2022-04-06
Intrepid2: Support Full Exact Sequence in Structured Integration

CamelliaDPG:
@trilinos/intrepid2 

## Motivation
This PR extends previous work in "Structured Integration", a term that generalizes sum factorization algorithms for finite element assembly.  The main thing that was missing for full support of the exact sequence was pullbacks in FunctionSpaceTools for spaces other than H(grad).  This PR adds those, as well as implementations of some sample formulations -- corresponding to the natural norms in each function space -- in both unit tests and performance tests.  There are some changes to the new data structures as well as to the new integration kernels that were required for this full support.

StructuredIntegrationTests has gotten an upgrade; for the assembly comparison tests (of which there are now five distinct formulations assembled: Poisson plus the natural norms for each space), we now use test templates.  This simplifies declarations,  makes it easier to see what cases are covered, and eliminates a fair amount of code redundancy.

Notably, we observe worse performance for standard Intrepid2 integration under Cuda on Weaver than we previously saw on Ride (which is no longer available).  It is not clear to me what the cause is, but I have verified that Trilinos develop exhibits similar behavior; I have seen no evidence that this PR is in any way implicated in the performance degradation (which would be unlikely in any case, since this PR makes no changes to standard integration kernels; its target is the new structured integration kernels).

(A document with full performance results is available on request.)

## Testing
The new features introduced here are thoroughly tested.

### DIFF
--- a/packages/intrepid2/assembly-examples/GRADGRADStandardAssembly.hpp
+++ b/packages/intrepid2/assembly-examples/GRADGRADStandardAssembly.hpp
@@ -10,7 +10,11 @@
 
 #include "JacobianFlopEstimate.hpp"
 
-//! version that uses the classic, generic Intrepid2 paths;
+/** \file   GRADGRADStandardAssembly.hpp
+    \brief  Locally assembles a Poisson matrix -- an array of shape (C,F,F), with formulation (grad e_i, grad e_j), using standard Intrepid2 methods; these do not algorithmically exploit geometric structure.
+ */
+
+//! Version that uses the classic, generic Intrepid2 paths.
 template<class Scalar, class BasisFamily, class PointScalar, int spaceDim, typename DeviceType>
 Intrepid2::ScalarView<Scalar,DeviceType> performStandardQuadratureGRADGRAD(Intrepid2::CellGeometry<PointScalar, spaceDim, DeviceType> &geometry,
                                                                               const int &polyOrder, int worksetSize,
--- a/packages/intrepid2/assembly-examples/GRADGRADStandardAssembly.hpp
+++ b/packages/intrepid2/assembly-examples/GRADGRADStandardAssembly.hpp
@@ -39,7 +39,7 @@ Intrepid2::ScalarView<Scalar,DeviceType> performStandardQuadratureGRADGRAD(Intre
 
   shards::CellTopology cellTopo = geometry.cellTopology();
   
-  auto basis = getBasis< NodalBasisFamily<DeviceType> >(cellTopo, fs, polyOrder);
+  auto basis = getBasis< BasisFamily >(cellTopo, fs, polyOrder);
   
   int numFields = basis->getCardinality();
   int numCells = geometry.numCells();
--- a/packages/intrepid2/assembly-examples/GRADGRADStandardAssembly.hpp
+++ b/packages/intrepid2/assembly-examples/GRADGRADStandardAssembly.hpp
@@ -1,0 +1,147 @@
+//
+//  GRADGRADStandardAssembly.hpp
+//  Trilinos
+//
+//  Created by Roberts, Nathan V on 7/2/21.
+//
+
+#ifndef Intrepid2_GRADGRADStandardAssembly_hpp
+#define Intrepid2_GRADGRADStandardAssembly_hpp
+
+#include "JacobianFlopEstimate.hpp"
+
+//! version that uses the classic, generic Intrepid2 paths;
+template<class Scalar, class BasisFamily, class PointScalar, int spaceDim, typename DeviceType>
+Intrepid2::ScalarView<Scalar,DeviceType> performStandardQuadratureGRADGRAD(Intrepid2::CellGeometry<PointScalar, spaceDim, DeviceType> &geometry,
+                                                                              const int &polyOrder, int worksetSize,
+                                                                              double &transformIntegrateFlopCount, double &jacobianCellMeasureFlopCount)
+{
+  using ExecutionSpace = typename DeviceType::execution_space;
+  int numVertices = 1;
+  for (int d=0; d<spaceDim; d++)
+  {
+    numVertices *= 2;
+  }
+  
+  auto jacobianAndCellMeasureTimer = Teuchos::TimeMonitor::getNewTimer("Jacobians");
+  auto fstIntegrateCall = Teuchos::TimeMonitor::getNewTimer("transform + integrate()");
+  auto initialSetupTimer = Teuchos::TimeMonitor::getNewTimer("Initial Setup");
+  initialSetupTimer->start();
+  
+  using CellTools = Intrepid2::CellTools<DeviceType>;
+  using FunctionSpaceTools = Intrepid2::FunctionSpaceTools<DeviceType>;
+  
+  using namespace Intrepid2;
+  
+  using namespace std;
+  // dimensions of the returned view are (C,F,F)
+  auto fs = FUNCTION_SPACE_HGRAD;
+
+  shards::CellTopology cellTopo = geometry.cellTopology();
+  
+  auto basis = getBasis< NodalBasisFamily<DeviceType> >(cellTopo, fs, polyOrder);
+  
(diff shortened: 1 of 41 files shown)


packages/intrepid2/assembly-examples/GRADGRADStandardAssembly.hpp
```diff
@@ -0,0 +1,147 @@
...
+//
+//  Created by Roberts, Nathan V on 7/2/21.
+//
+
+#ifndef Intrepid2_GRADGRADStandardAssembly_hpp
+#define Intrepid2_GRADGRADStandardAssembly_hpp
+
+#include "JacobianFlopEstimate.hpp"
+
+//! version that uses the classic, generic Intrepid2 paths;
```
mperego, line 13: Maybe here, and also in the other assembly files, you can add a sentence describing what you are assembling

mperego (approved):
This looks good to me. Thanks @CamelliaDPG !

If I understand correctly in the assembly examples you are not accounting for orientation which is a bit weird because you are assembling CG finite element terms. Would it be hard to add the orientation mappings?

It's great that you are obtaining such a huge speedup on GPUs.
As for the degradation of the "standard" Intrepid2 performance, it would be good to open an issue.

CamelliaDPG:
Sounds good; will do.

On Apr 19, 2022, at 5:41 PM, mperego ***@***.******@***.***>> wrote:



@mperego commented on this pull request.

________________________________

In packages/intrepid2/assembly-examples/GRADGRADStandardAssembly.hpp<https://gcc02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Ftrilinos%2FTrilinos%2Fpull%2F10419%23discussion_r853601924&data=05%7C01%7Cnvrober%40sandia.gov%7Cf345f5146a1a4abcfb5008da225e24fd%7C7ccb5a20a303498cb0c129007381b574%7C1%7C0%7C637860085007153673%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=0f7u7CYDWk0QPV8itAh6ELgH%2Fm9SKZJ54M5%2FQ8UyNt0%3D&reserved=0>:

> @@ -0,0 +1,147 @@
+//
+//  GRADGRADStandardAssembly.hpp
+//  Trilinos
+//
+//  Created by Roberts, Nathan V on 7/2/21.
+//
+
+#ifndef Intrepid2_GRADGRADStandardAssembly_hpp
+#define Intrepid2_GRADGRADStandardAssembly_hpp
+
+#include "JacobianFlopEstimate.hpp"
+
+//! version that uses the classic, generic Intrepid2 paths;


Maybe here, and also in the other assembly files, you can add a sentence describing what you are assembling

CamelliaDPG:
Thanks, @mperego!  What I do here is purely local assembly: I produce an array of shape (C,F,F): cell, (local) field, (local) field.  I think orientations would naturally be applied at the point of global assembly, when you are mapping from (C,F,F) to (GF,GF) in for    a (sparse) global matrix indexed by global field ordinals.  Right?

IIRC, I did include orientation support in the CellGeometry object, so it should not be hard to apply orientations.  But I haven't yet thought through this too carefully (at least not lately — I probably did think about it carefully a year ago).

Yes, agreed on opening an issue for "standard" Intrepid2 performance.

On Apr 19, 2022, at 5:47 PM, mperego ***@***.******@***.***>> wrote:



@mperego approved this pull request.

This looks good to me. Thanks @CamelliaDPG<https://gcc02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FCamelliaDPG&data=05%7C01%7Cnvrober%40sandia.gov%7C676b0a7bfd864715a08008da225eeb68%7C7ccb5a20a303498cb0c129007381b574%7C1%7C0%7C637860088316582944%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=8IjWKAT2%2B%2BKs3KlolwBmUehrpWIdco4DYxcK7cmlu%2Bg%3D&reserved=0> !

If I understand correctly in the assembly examples you are not accounting for orientation which is a bit weird because you are assembling CG finite element terms. Would it be hard to add the orientation mappings?

It's great that you are obtaining such a huge speedup on GPUs.
As for the degradation of the "standard" Intrepid2 performance, it would be good to open an issue.

mperego:
> Thanks, [@mperego](https://github.com/mperego)!  What I do here is purely local assembly: I produce an array of shape (C,F,F): cell, (local) field, (local) field.  I think orientations would naturally be applied at the point of global assembly, when you are mapping from (C,F,F) to (GF,GF) in for    a (sparse) global matrix indexed by global field ordinals.  Right?

I think that in principle for simple terms like that you could apply the orientation after you build the local matrix, but that's not what we do, we apply orientations before we transform the basis from reference to physical elements. Also for nonlinear terms it doesn't seem possible to do that.

mperego:
> > Thanks, [@mperego](https://github.com/mperego)!  What I do here is purely local assembly: I produce an array of shape (C,F,F): cell, (local) field, (local) field.  I think orientations would naturally be applied at the point of global assembly, when you are mapping from (C,F,F) to (GF,GF) in for    a (sparse) global matrix indexed by global field ordinals.  Right?
> 
> I think that in principle for simple terms like that you could apply the orientation after you build the local matrix, but that's not what we do, we apply orientations before we transform the basis from reference to physical elements. Also for nonlinear terms it doesn't seem possible to do that.

I guess one could apply the orientation matrix to the "unknown" basis coefficients and to the row of the matrix. This would be equivalent to modify the gather and scatter methods in Panzer, and would avoid performing the mapping of the basis functions at each quadrature point, and for each basis function evaluation (value, gradient, divergence, etc.). Our usual orientation mappings are permutation. If so, in Panzer we could simply reorder the local maps of the DOF manager. However, in general, they don't have to be permutations. Anyway, this is something we should take into account, not in this PR.

CamelliaDPG:
> 



> > > Thanks, [@mperego](https://github.com/mperego)!  What I do here is purely local assembly: I produce an array of shape (C,F,F): cell, (local) field, (local) field.  I think orientations would naturally be applied at the point of global assembly, when you are mapping from (C,F,F) to (GF,GF) in for    a (sparse) global matrix indexed by global field ordinals.  Right?
> > 
> > 
> > I think that in principle for simple terms like that you could apply the orientation after you build the local matrix, but that's not what we do, we apply orientations before we transform the basis from reference to physical elements. Also for nonlinear terms it doesn't seem possible to do that.
> 
> I guess one could apply the transposed orientation matrix to the basis coefficients of FE fields and left-multiply the local matrix by the orientation matrix. This would be equivalent to modify the gather and scatter methods in Panzer, and would avoid performing the mapping of the basis functions at each quadrature point, and for each basis function evaluation (value, gradient, divergence, etc.). Our usual orientation mappings are permutations. If so, in Panzer we could simply implement orientations by reordering the local maps provided by the DOF manager. However, in general, they don't have to be permutations. Anyway, this is something we should take into account, not in this PR.

Yes, let's talk further about this sometime soon.  I do think you can do this in a perfectly general way at the global assembly stage -- I do it this way in Camellia, and I think I was imitating Demkowicz's code in doing so.  But I also think that there's a pretty natural way to add orientations into the new data structures; specifically, the orientation for a tensor-product geometry will decompose into orientations on the tensorial components.  One idea would be to include an optional Orientation array in the "transform" methods, and incorporate into the TransformedBasisValues structure included in this PR.  The assembly could then apply the orientations inline, probably without too much change required in the core kernels.

GrahamBenHarper (approved):
:)

PR #10423, opened 2022-04-07
cmake/std/PullRequestLinuxDriver.sh: Use sems-python

e10harvey:
@trilinos/framework
## Motivation

PR testing is failing on cloud nodes due to an absolute path to python3. This PR removes the path and relies on the sems-python instead.

## Testing

Tested by running this change through the testing infrastructure.

### DIFF
--- a/cmake/std/PullRequestLinuxDriver.sh
+++ b/cmake/std/PullRequestLinuxDriver.sh
@@ -34,26 +34,12 @@ function bootstrap_modules() {
             exit -1
         fi
     else
-        source /projects/sems/modulefiles/utils/sems-archive-modules-init.sh
-        module unload sems-archive-git
-        module unload sems-archive-python
-        module load sems-archive-git/2.10.1
-
-#        module load sems-python/3.5.2      # Currently not on cloud nodes
-#        #pip3 install --user configparser
-#        get_python_packages pip3
-#        export PYTHON_EXE=python3
-
-#         envvar_set_or_create     PYTHONHOME /projects/sierra/linux_rh7/install/Python/3.6.3
-         envvar_set_or_create     PYTHONHOME /projects/sierra/linux_rh7/install/Python/3.6.10
-#        #envvar_set_or_create     PYTHONPATH ${HOME}/.local/lib/python3.6/site-packages
-#        #envvar_append_or_create  PYTHONPATH ${PYTHONHOME:?}/lib/python3.6/site-packages
-#        unset PYTHONHOME
-         unset PYTHONPATH
-         envvar_prepend_or_create PATH       ${PYTHONHOME:?}/bin
-         envvar_set_or_create     PYTHON_EXE ${PYTHONHOME:?}/bin/python3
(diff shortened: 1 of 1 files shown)


ZUUL42 (approved):

ndellingwood (approved):

e10harvey:
@jwillenbring, @ZUUL42: Please consider force-merging this PR. On 4/7/22 I verified that these changes worked on both the cloud node runners and non-cloud node runners.

jwillenbring:
@e10harvey is this the chicken and the egg problem then at this point?

e10harvey:
> @e10harvey is this the chicken and the egg problem then at this point?

@jwillenbring: Yessir.

ZUUL42:
@e10harvey, I would think this would test fine. Isn't PullRequestLinuxDriver.sh merged before it's used.
It's really only if a file is used before it's merged that we should need to force merge or temporary lift tests.
Do you happen to have any of the job numbers from that test you mentioned on 4/7? I scanned through the GCC 8.3.0 jobs that day and didn't find any jobs run on 4/7 that weren't kicked off by the entity account.

cmake/std/PullRequestLinuxDriver.sh
```diff
@@ -34,26 +34,12 @@ function bootstrap_modules() {
...
-         envvar_set_or_create     PYTHONHOME /projects/sierra/linux_rh7/install/Python/3.6.10
-#        #envvar_set_or_create     PYTHONPATH ${HOME}/.local/lib/python3.6/site-packages
-#        #envvar_append_or_create  PYTHONPATH ${PYTHONHOME:?}/lib/python3.6/site-packages
-#        unset PYTHONHOME
-         unset PYTHONPATH
-         envvar_prepend_or_create PATH       ${PYTHONHOME:?}/bin
-         envvar_set_or_create     PYTHON_EXE ${PYTHONHOME:?}/bin/python3
-         #export PYTHONHOME=/projects/sierra/linux_rh7/install/Python/3.6.3
-         #export PYTHONPATH=${HOME}/.local/lib/python3.6/site-packages:${PYTHONHOME:?}/lib/python3.6/site-packages
+	source /projects/sems/modulefiles/utils/sems-modules-init.sh
```
ZUUL42, line 37: Will this allow the builds still on sems-archive to get to their TPLs?

e10harvey: To the best of my knowledge, all the builds have both sems-archive and "sems" available.

jwillenbring:
I left the following comment on this merge commit:

This needs to be merged before the test jobs will be reliably assigned to nodes where they can run properly, so we are force pushing this. Currently if any jobs are assigned to cloud nodes, the module loads will fail. This is bringing down too many tests. The only other way to resolve this would be to modify all of the test jobs to not run on cloud nodes, which would create resource contention for the PRs, and not test any of the possible unlikely corner cases that could require additional modifications to this PR.

Issue #10433, opened 2022-04-13
Tpetra: Finish off Kokkos::deep_copy 2-arg to 3-arg conversion where appropriate


csiefer2:
For a successor PR to #7946

PR #10434, opened 2022-04-13
Tpetra: Move from 2-arg to 3-arg Kokkos::deep_copy where appropriate

csiefer2:
Issue: #10433
Tests: Cuda w/ and w/o UVM
Stakeholder feedback: n/a

This is a successor to #7946 which we will close when this one merges.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_get1DConstView.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_get1DConstView.hpp
@@ -1,0 +1,117 @@
+// @HEADER
+// ***********************************************************************
+//
+//          Tpetra: Templated Linear Algebra Services Package
+//                 Copyright (2008) Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the Corporation nor the names of the
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL SANDIA CORPORATION OR THE
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact Michael A. Heroux (maherou@sandia.gov)
+//
+// ************************************************************************
+// @HEADER
+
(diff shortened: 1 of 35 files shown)


kyungjoo-kim:
@csiefer2 @crtrott One question. One suggestion. Currently, it uses an execution space instance without any stream information. So, in such case, does Kokkos provide something like ``Kokkos::deep_copy_async(A, B)`` that is equivalent to ``Kokkos::deep_copy(exec_space(), A, B)`` ? The key point of using execution space instance is that it does not synchronize and the deep copy async is more clear for readability.

csiefer2:
> @csiefer2 @crtrott One question. One suggestion. Currently, it uses an execution space instance without any stream information. So, in such case, does Kokkos provide something like `Kokkos::deep_copy_async(A, B)` that is equivalent to `Kokkos::deep_copy(exec_space(), A, B)` ? The key point of using execution space instance is that it does not synchronize and the deep copy async is more clear for readability.

Two points:
- Nothing currently works with streams, so all of this will need to be overhauled (again) for that anyway.
- IIRC, 3-arg deep_copy does does a cudaStreamSynchronize, so "async" here would be syntactically different.

kyungjoo-kim:
Seeing Kokkos_CudaSpace.cpp file, providing cuda execution space instance, it does not sync. So, the syntax of using 3 arg version is equivalent to deep copy async. 
```
void DeepCopyCuda(void *dst, const void *src, size_t n) {
  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMemcpy(dst, src, n, cudaMemcpyDefault));
}

void DeepCopyAsyncCuda(const Cuda &instance, void *dst, const void *src,
                       size_t n) {
  KOKKOS_IMPL_CUDA_SAFE_CALL(
      cudaMemcpyAsync(dst, src, n, cudaMemcpyDefault, instance.cuda_stream()));
}

void DeepCopyAsyncCuda(void *dst, const void *src, size_t n) {
  cudaStream_t s = cuda_get_deep_copy_stream();
  KOKKOS_IMPL_CUDA_SAFE_CALL(
      cudaMemcpyAsync(dst, src, n, cudaMemcpyDefault, s));
  Impl::cuda_stream_synchronize(
      s,
      Kokkos::Tools::Experimental::SpecialSynchronizationCases::
          DeepCopyResourceSynchronization,
      "Kokkos::Impl::DeepCopyAsyncCuda: Deep Copy Stream Sync");
}
```

crtrott:
The three arg version with a cuda instance does not sync, it just enqueues the copy.

csiefer2:
> The three arg version with a cuda instance does not sync, it just enqueues the copy.

Well, I stand corrected then.

packages/tpetra/core/src/Tpetra_Details_get1DConstView.hpp
```diff
@@ -0,0 +1,117 @@
...
+#include "Tpetra_ConfigDefs.hpp"
+#include "Tpetra_Util.hpp"
+#include "Kokkos_DualView.hpp"
+#include <Teuchos_Array.hpp>
+#include <utility>
+
+namespace Tpetra {
+namespace Details {
+
+// mfh 28 Apr 2016: Sometimes we have a raw host array, and we need
```
tasmith4, line 59: I'm confused why a file with a comment from Mark from 2016 shows up as completely new in the diff.  Is this something we had deleted that's being re-added incorrectly?

tasmith4 (changes requested):
There are a number of unqualified calls to `Kokkos::fence()` added in this PR.  I think we generally prefer `execution_space().fence()` in most instances to avoid fencing other execution spaces unnecessarily (this has bitten us before).  Are we sure that every new `Kokkos::fence()` absolutely must fence all execution spaces?

packages/tpetra/core/src/Tpetra_Details_get1DConstView.hpp
csiefer2: Probably

cgcgcg (approved):

tasmith4 (approved):

rppawlo:
@csiefer2 - this commit has broken panzer. It snuck through PR testing because CUDA_LAUNCH_BLOCKING was accidentally enabled in PR testing.  Might need some help fixing.

@jwillenbring @trilinos/framework - we had disabled CUDA_LAUNCH_BLOCKING in PR testing but it looks like it is now active in the vortex build. At least it is enabled in PullRequestCuda10.1.243TestingEnv.sh if that is what we are using with PR testing. Reverting this commit fixes the panzer tests.

rppawlo:
The failing Panzer tests are:
```
The following tests FAILED:
	135 - PanzerAdaptersSTK_check_sorted_face_quads_MPI_1 (Failed)
	154 - PanzerAdaptersSTK_LocalMesh_ParallelPeriodicTests_MPI_2 (Failed)
	178 - PanzerAdaptersSTK_PoissonInterfaceTpetra_2d_MPI_4 (Failed)
	180 - PanzerAdaptersSTK_PoissonInterfaceTpetra_3d_MPI_4 (Failed)
Errors while running CTest
```

csiefer2:
@rppawlo What configuration is failing?

csiefer2:
Also, I would have _never_ seen this unless I was searching through my email for your name for a completely unrelated reason.

rppawlo:
@csiefer2 = good to know. I will send direct email next time. Vortex is the issue. The cuda PR environment is setting CUDA_LAUNCH_BLOCKING=1 by accident. I think @e10harvey made a change to some of the internal PR testing scripts, but you will have to manually unset that env variable before running if you are using the scripts in Trilinos/cmake/std/sems. Those still have the variable set.

rppawlo:
You should be able to reproduce on any cuda build though. I have reproduced on the ascicgpu machines.

Issue #10334, opened 2022-03-15
Memory corruption in the AZ_pgmres function


AndrewUshakov:
## Bug Report
@trilinos/aztecoo/

### Description

 The first parameter of the function

```double *AZ_manage_memory(unsigned int input_size, int action, int type, char *name, int *status)```

defined in az_util.c:851, has type `unsigned int` (4 bytes), while it is called with `unsigned long` (8 bytes) in the az_gmres.c:240

```vblock = AZ_manage_memory((kspace+1)*aligned_N_total*sizeof(double),AZ_ALLOC, AZ_SYS+az_iterate_id,label, &i);```

where kspace - Krylov space size and aligned_N_total - number of unknowns. So if expression

```(kspace+1)*aligned_N_total*sizeof(double)```

does not fit in 4 bytes, size of really allocated memory is smaller than requested, and crash occurs in the AZ_MSR_matvec_mult or AZ_VBR_matvec_mult finally called from az_gmres.c:535. 

### How to reproduce:

1. Github commit fc799bdb (09/03/2022) was used.

2. My test program is:

```
#include <stdio.h>
#include <az_aztec.h>
#include <stdlib.h>
#include <assert.h>

int main(int argc, char** argv)
{
    FILE*   f;
    int     numProcUnknowns;
    double* wAC;
    double* bAC;
    int     options[AZ_OPTIONS_SIZE];
    double  params[AZ_PARAMS_SIZE];
    int     bindxLength;
    int*    bindx;
    double* val;
    int     data_org[AZ_COMM_SIZE];
    double  status[AZ_STATUS_SIZE];
    int     proc_config[AZ_PROC_SIZE];
    
    assert(argc == 2);
    printf(argv[1]);
    printf("\n");
    f = fopen(argv[1], "rb");
    
    if (!f)
        return 1;
    
    fread(&numProcUnknowns, sizeof(int), 1, f);
    printf("numProcUnknowns = %d\n", numProcUnknowns);
    
    wAC = calloc(numProcUnknowns, sizeof(double));
    assert(wAC != NULL);
    fread(wAC, sizeof(double), numProcUnknowns, f);
    
    bAC = calloc(numProcUnknowns, sizeof(double));
    assert(bAC != NULL);
    fread(bAC, sizeof(double), numProcUnknowns, f);
    
    fread(options, sizeof(int), AZ_OPTIONS_SIZE, f);
    
    fread(params, sizeof(double), AZ_PARAMS_SIZE, f);

[...]

PR #10436, opened 2022-04-13
aztecoo: Fix memory corruption in the AZ_pgmres function

AndrewUshakov:
## aztecoo: Fix memory corruption in the AZ_pgmres function

@trilinos/aztecoo

Summary of changes:

    * Correct signatures of the functions AZ_manage_memory, AZ_allocate and AZ_realloc;
    * Correct type of the size field of the mem_ptr and widget structs;
    * Correct types of some local variables;
    * Correct conversions of/to integer types.

## Motivation

Issue #10334

## Testing

N/A

### DIFF
--- a/packages/aztecoo/src/az_aztec.h
+++ b/packages/aztecoo/src/az_aztec.h
@@ -728,7 +728,7 @@ void PREFIX AZ_SLAIC1_F77(int * , int *, float *, float *, float *, float *,
 
   extern int AZ_adjust_N_nz_to_fit_memory(int N, int , int);
 
-  extern char *AZ_allocate(unsigned int iii);
+  extern char *AZ_allocate(size_t iii);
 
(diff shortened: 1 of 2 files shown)


AndrewUshakov:
As I understand, issue still exist and influences on the GOMA software. Please review this pull request!

jhux2:
Let's run this through PR testing.

jhux2:
@AndrewUshakov Is it possible to test this without requiring a large input matrix?

AndrewUshakov:
@jhux2: This pull request corrects an error caused by an incorrect (too short) length of the parameter used to reserve the memory needed for the computation. If the matrix is small, there is no error even without using the discussed pool request. On the other hand, without big matrix it is impossible to verify that this patch fully fixes the problem.

I can only suggest generating it for testing.

rstumin (approved):
This looks good.  I don't have any great ideas on a new large matrix test for this, so I think it is good as is.

PR #10444, opened 2022-04-14
belos: add missing parameter to getValidParameters in GmresPolySolMgr

tmranse:
@trilinos/belos
@jennloe 

## Motivation
In trying to use the `BelosGmresPolySolMgr` object in Sierra/Aria, it was uncovered that the "Outer Solver Param" argument was not in the set of valid arguments provided. This was causing the object to throw when trying to give non-default outer solver parameters,

## Testing
Tested by running this change through the testing infrastructure.


hkthorn (approved):
LGTM.

jennloe (approved):

PR #10447, opened 2022-04-15
Tpetra: Deprecate Distributor's useDistinctTags option

tasmith4:
@trilinos/tpetra 

## Motivation

Since the default is to use distinct tags for each code path,
this deprecation effectively is taking away the option for users
to ask Distributor to always use the same tag.

The more robust strategy of giving one unique tag per DistributorActor
should make useDistinctTags unnecessary.  Deprecating the option to
pass it in the parameter list lets us change it as needed later.

## Related Issues

* Related to #10308

## Testing
Manually tested a Tpetra build with deprecated code off on GCC 7.2.0, and did not see any failing tests.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
@@ -890,7 +890,9 @@ void DistributorPlan::setParameterList(const Teuchos::RCP<Teuchos::ParameterList
 #endif
     const Details::EDistributorSendType sendType =
       getIntegralValue<Details::EDistributorSendType> (*plist, "Send type");
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
     const bool useDistinctTags = plist->get<bool> ("Use distinct tags");
+#endif
 
 #ifdef TPETRA_ENABLE_DEPRECATED_CODE
     {
@@ -929,8 +931,8 @@ void DistributorPlan::setParameterList(const Teuchos::RCP<Teuchos::ParameterList
 #endif
 #ifdef TPETRA_ENABLE_DEPRECATED_CODE
     barrierBetweenRecvSend_ = barrierBetween;
-#endif
     useDistinctTags_ = useDistinctTags;
+#endif
 
     // ParameterListAcceptor semantics require pointer identity of the
     // sublist passed to setParameterList(), so we save the pointer.
@@ -963,8 +965,8 @@ DistributorPlan::getValidParameters() const
 
 #ifdef TPETRA_ENABLE_DEPRECATED_CODE
   const bool barrierBetween = Details::barrierBetween_default;
-#endif
   const bool useDistinctTags = Details::useDistinctTags_default;
+#endif
 
   Array<std::string> sendTypes = distributorSendTypes ();
   const std::string defaultSendType ("Send");
@@ -991,9 +993,11 @@ DistributorPlan::getValidParameters() const
   setStringToIntegralParameter<Details::EDistributorSendType> ("Send type",
       defaultSendType, "When using MPI, the variant of send to use in "
       "do[Reverse]Posts()", sendTypes(), sendTypeEnums(), plist.getRawPtr());
+#ifdef TPETRA_ENABLE_DEPRECATED_CODE
   plist->set ("Use distinct tags", useDistinctTags, "Whether to use distinct "
       "MPI message tags for different code paths.  Highly recommended"
       " to avoid message collisions.");
+#endif
   plist->set ("Timer Label","","Label for Time Monitor output");
 
 #ifdef TPETRA_ENABLE_DEPRECATED_CODE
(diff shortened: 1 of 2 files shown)


csiefer2 (approved):

mperego:
We cannot check whether these changes are OK since no PR builds turn off deprecated code. @tasmith4, have you tried running all Tpetra dependent packages tests in Trilinos with deprecated code off?  @jwillenbring  should we delay merging this PR until deprecated code is turned off in one of the PR builds? Removing the Automerge flag for now. Sorry for the intrusion but we have been chasing issues related to Tpetra deprecated code for months.

tasmith4:
@mperego while I did not run all Tpetra dependent packages tests in Trilinos with deprecated code off, I did grep the entire Trilinos project for explicit usages of this option and did not find any, so I would be extremely surprised if any tests failed due to this deprecation.  In addition, as I noted in the PR text above, we are moving in the direction of *more* robust tag assignment, not less.  By removing this option we're effectively removing the ability for users to ask for Distributor to use the same tag for all communication.  Since we've already observed that using the same tag for all communications has caused errors in Tpetra, we don't believe this is a good option to maintain.  It's also difficult to see how forcing users to use different tags for different rounds of MPI communication could actively cause errors in previously working code.

Are you aware of specific packages that may be affected by this?  Alternatively, are you aware of potential scenarios where code that is correct with multiple MPI calls sharing the same (comm, rank, tag) tuple would become incorrect if the tags were forced to be distinct?

tasmith4:
I'm very sympathetic to the pain of the UVM removal and associated Tpetra deprecations, as I've had to work on retrofitting packages that depend on Tpetra for this stuff as well.  I also agree that it could make sense for the Trilinos leadership team to have a broader discussion about the Trilinos deprecation strategy and how to handle merging PRs with deprecations (although I'll note that from my experience on Sierra, there are downsides to the strategy of blocking merges before all dependent deprecations are resolved).

However, I'm hoping these broader concerns don't need to block this PR, which I believe is extremely unlikely to cause issues for dependent packages, especially in Trilinos, and which is itself blocking other work in Distributor that is needed to fix the currently broken asynchronous Import/Export capability, which several Tpetra customers are interested in.

mperego:
@tasmith4 I'm not questioning at all the benefits of the PR. I just want to be on the safe side and I worry that these changes are not seen by the PR testing. If you are positive that no Trillions package will break because of this change when deprecated code is turn off then I won't interfere with this.

tasmith4:
@mperego I don't have any reason to believe these changes would break any Trilinos package.

mperego:
> although I'll note that from my experience on Sierra, there are downsides to the strategy of blocking merges before all dependent deprecations are resolved

what was your experience?

tasmith4:
> what was your experience?

Sierra turns off deprecated code in the main development builds (maybe in all builds), and both STK (which I work on) and all the Sierra apps are in the same base repo.  So if we deprecate anything in STK, we can't push it until we fix all the usages in all the app code, or the build will be broken.

This ensures no issues with apps using deprecated code, which is a plus.  But it effectively removes any internal deprecation cycle (the only reason we even bother to mark things deprecated is for the people that use STK through Trilinos).  Although deprecation cycles are annoying, I think they do serve a useful purpose, by allowing two versions of the code to coexist side-by-side for a period of time and giving users time to switch their usage, instead of forcing all usages to be fixed in one massive commit that's difficult to review, takes a lot of time to prepare, and creates a large potential for merge conflicts.  In addition, these commits often touch many packages, so can be very hard to get through the merge process -- if there's an issue in one, it holds up everyone, instead of allowing each package to fix issues separately.

Because the Sierra teams have generally good communication with each other, these issues can be and are regularly overcome.  But it isn't all rainbows and unicorns. :)  The bottom line for me is that I think deprecations in low-level packages with many dependents are challenging no matter what, because of all the people using the API, especially when it's something very fundamental like what Tpetra did.  I tend to lean towards allowing packages to push separately with deprecation cycles, since I think it's always good when you can break up PRs into smaller, decoupled chunks.  But I don't think any merge strategy can ultimately save us from all the pain associated with deprecations in low-level packages.

mperego:
@tasmith4 Sorry for getting back to you only now. I see your point and I agree there is not perfect solution. 
I prefer to force consistency in Trilinos (that is any Trilinos package should not use any deprecated code at any time) for mainly two reasons. The first, is that this is what Trilinos customers would expect. I'm an Albany developer and I don't want to see any deprecation warning that is not related to Albany and that I cannot fix by modifying Albany only. The second is that it's way more efficient if the developers that deprecate the code take care of fixing all the calls to that code in Trilinos, because they know what to do and they can do it one sitting instead of having n developers fixing their own packages. I understand that in some cases the fixes to avoid calling deprecated functions are not trivial and need the help form other package developers. In these cases one can always use a temporary cmake option (internal to Trilinos) to ifdef out a function that needs to be deprecated and work with other package developers to remove all the calls to that function, and only then, the function can be deprecated.

tasmith4:
@mperego thanks for the follow up -- yeah, you make some good counter-points, which just goes to show why it's a hard problem with no "right answer".  Probably whichever factors each person perceives as most important will dictate their preference.

PR #10458, opened 2022-04-25
Ifpack2, FastILU : add an option to use KokkosKernels::sptrsv

iyamazaki:
@trilinos/ifpack2, @trilinos/shylu 

## Motivation

This PR adds an option to use KokkosKernels::sptrsv with FastILU

## Testing

This new option has been added to ctest in ifpack2/belos

## Additional Information

It also tries removing `Arguments are being passed in but not used.  UNPARSED_ARGUMENTS =` warning in CMakeLists.txt of ifpack2/test/belos with fastilu

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_FastILU_Base_decl.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_FastILU_Base_decl.hpp
@@ -200,6 +200,7 @@ template<class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
     {
       Params() {}
       Params(const Teuchos::ParameterList& pL, std::string precType);
+      bool standard_sptrsv;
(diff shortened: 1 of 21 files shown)


ndellingwood (approved):
@iyamazaki changes lgtm. I'm not familiar with the fast ilu implementation, but from what I can tell two matrices are stored, one each for non-transpose and transpose solves, in which case this all looks fine and defaults to cuSPARSE when the tpl is enabled in Cuda builds. Thanks Ichi!

ndellingwood:
@iyamazaki once #10462 merges the symbolic phase for sptrsv won't be so terribly, horribly slow (those changes didn't make it in for 3.6)

iyamazaki:
Thank you @ndellingwood !!

Issue #10463, opened 2022-04-26
Tpetra:  Nalu regression testing fails due to recent Trilinos changes


wjhorne:
## Bug Report
@trilinos/tpetra

### Description
Nalu's  (@spdomin) nightly regression test suite currently fails due to recent changes to Trilinos. A git bisect yields the output below. From what we can tell there are new, extremely small diffs in residuals due to the changes in the commit below. This could be due to something as small as an order of operations change or similar. Could we get some verification that the changes are correct so that we can re-bless our tests accordingly?  

72314b8136aa3980e339d186be6d4bc8242c7506 is the first bad commit
commit 72314b8136aa3980e339d186be6d4bc8242c7506
Author: Christian Glusa <caglusa@sandia.gov>
Date:   Wed Apr 6 12:43:21 2022 -0600

    Tpetra: Use numVectors==1 specialization for local parts of norms

:040000 040000 475b4cef23ffb399e71a1e3749d5ae193b1a5ba1 d9be457f73757da6aa0ee8d1480b96750f5b0dc2 M packages

### Steps to Reproduce
1. Checkout commit shown above 
2. Link resulting build to latest Nalu build

cgcgcg:
@trilinos/tpetra

cgcgcg:
We switched from using a multivector kernel of the norm to a single vector kernel when a multivector only has one vector. I'm not sure why this would give a slightly different result. @brian-kelley

brian-kelley:
@cgcgcg It would change the associativity of the additions. The single vector kernel does whatever Kokkos::parallel_reduce does, but the multivector kernel breaks each column into blocks and atomically adds the partial sums into the result. 

However, I'm surprised this would cause a Nalu diff now because before I merged in #10028 on Jan. 13, which accidentally took out this single-vector special case from the multivector kernel in KokkosKernels. But that didn't cause a regression, as far as I know.

So basically the behavior of Tpetra's norms with 1 column was:

pre-Jan 13: parallel_reduce
Jan 13 - Apr 6: multivector kernel, blocks combined with atomic_add (slow for 1 column)
Apr 6 - now: parallel_reduce again

BTW, since Kokkos 3.6 was merged today, I'm working on a permanent PR to restore the numVecs=1 special path in Kokkos Kernels. This way Tpetra doesn't need to have the special path. But it won't have an effect on this regression.

cgcgcg:
@brian-kelley Could this have been the issue in January? #10086

[...]

Issue #1354, opened 2017-05-23
Amesos2 test failures on dashboard


ndellingwood:
The following tests failed 05/23/2017 with build configuration `Linux-GCC-4.8.3-SERIAL_DEBUG_DEV_MueLu_ExtraTypes`

```
Amesos2_Epetra_MultiVector_Adapter_UnitTests_MPI_4
Amesos2_Epetra_RowMatrix_Adapter_UnitTests_MPI_4
Amesos2_Tpetra_CrsMatrix_Adapter_UnitTests_MPI_4
```

More specifically:
The following tests FAILED:
    3. MultiVecAdapter_Copy_Map_UnitTest ... 

The following tests FAILED:
    2. CrsMatrixAdapter_CRS_Serial_UnitTest ... 
    4. CrsMatrixAdapter_CRS_Map_UnitTest ... 

The following tests FAILED:
    4. CrsMatrixAdapter_ComplexDouble_int_int_CRS_Map_UnitTest ... 
    10. CrsMatrixAdapter_double_int_int_CRS_Map_UnitTest ...

ndellingwood:
Closing, duplicate #1172; fixed with sha1 bab8b66393e97c632d9530b8fb38d69bf6abdc2b

PR #10465, opened 2022-04-26
KokkosKernels/Tpetra: Restore numVecs=1 BLAS path

brian-kelley:
This patches in Kokkos Kernels PR #1354, which restores the special
path in BLAS-1 MV reduction-like kernels for when there is only one
column. This path just calls the single-column kernel, which is a simple
parallel_reduce. This path existed before but I accidentally removed it.
This also reverts #10415 since Tpetra won't need to intercept the special path
if Kokkos Kernels is taking care of it.

@trilinos/kokkos-kernels 
@trilinos/tpetra 

## Motivation

For multivectors (rank-2 views) with a single column, the single-column kernels use a simple parallel_reduce and get much better performance than the general multivector kernel. This restores the single-column path for all the BLAS-1 reduction-like kernels.

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows #10415
* Precedes 
* Related to #10463
* Part of 
* Composed of 

## Testing

This expands KokkosKernel's BLAS-1 tests - now the rank-2 interface is tested with Nx1 vectors, and the weighted 2-norm kernels ``nrm2w`` and ``nrm2w_squared`` are tested (they weren't before).

### DIFF
--- a/packages/kokkos-kernels/src/blas/KokkosBlas1_nrm2w.hpp
+++ b/packages/kokkos-kernels/src/blas/KokkosBlas1_nrm2w.hpp
@@ -76,7 +76,8 @@ nrm2w(const XVector& x, const XVector& w) {
       typename XVector::device_type, Kokkos::MemoryTraits<Kokkos::Unmanaged> >
       XVector_Internal;
 
-  typedef Kokkos::View<mag_type, Kokkos::LayoutLeft, Kokkos::HostSpace,
+  typedef Kokkos::View<mag_type, typename XVector_Internal::array_layout,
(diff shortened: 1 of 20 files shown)


cgcgcg (approved):
LGTM. Wasn't the Kokkos-Kernels change supposed to be part of the promotion?

brian-kelley:
@cgcgcg No, it was merged into Kokkos Kernels develop a few days too late to be included in the promotion.

ndellingwood:
kokkos/kokkos-kernels#1354

Issue #10428, opened 2022-04-11
Trilinos: Autotester Down


csiefer2:
And more than just #10420 and #10412

Nothing is passing.  Also reported via trilinos-help

e10harvey:
There are 3 known recent problems:
- [x] Infrastructure nodes did not have the required python version. Fixed via https://github.com/trilinos/Trilinos/pull/10423.
- [ ] Krino unit tests failing during dynamic loading. Pending fix via #10429.
- [ ] All PRs blocked due to system issues with CUDA node. Ticket files with system-team, pending resolution.

csiefer2:
Thanks for the update!

csiefer2:
Intel & GCC 7 tests are failing across the board. @e10harvey @jwillenbring

jwillenbring:
There was a tmp space that was full. I think those failures should not occur for future builds.

csiefer2:
Cuda configures are failing _again_


```
CMake Error at /projects/sparc/tools/cmake/3.19.4/ppc64le/share/cmake-3.19/Modules/CMakeTestCXXCompiler.cmake:59 (message):
  The C++ compiler

    "/usr/tce/packages/spectrum-mpi/spectrum-mpi-rolling-release-gcc-8.3.1/bin/mpicxx"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /vscratch1/trilinos/jenkins/vortex-trilinos/workspace/trilinos-folder/_cuda_10.1.243@2/pull_request_test/CMakeFiles/CMakeTmp
    
    Run Build Command(s):/projects/atdm_devops/vortex/ninja-fortran-1.8.2/ninja cmTC_d7c9a && [1/2] Building CXX object CMakeFiles/cmTC_d7c9a.dir/testCXXCompiler.cxx.o
    FAILED: CMakeFiles/cmTC_d7c9a.dir/testCXXCompiler.cxx.o 
    /usr/tce/packages/spectrum-mpi/spectrum-mpi-rolling-release-gcc-8.3.1/bin/mpicxx    -o CMakeFiles/cmTC_d7c9a.dir/testCXXCompiler.cxx.o -c testCXXCompiler.cxx
    nvcc fatal   : Could not open output file '/tmp/trilinos/tmpxft_0001fa1f_00000000'
    ninja: build stopped: subcommand failed.
```

rppawlo:
[...]

PR #10469, opened 2022-04-27
packages/framework: Always create tmp dir for nvcc

e10harvey:
@trilinos/framework

## Motivation

On some ats2 node, /tmp/$USER does not exist which makes nvcc sad. See https://github.com/trilinos/Trilinos/issues/10428#issuecomment-1109956415.

Related to #10428.

## Testing

This was tested manually by running mkdir on ats2.


ZUUL42 (approved):

e10harvey:
@jwillenbring: Should this be force merged given that it is affecting many PR and MM builds?

jwillenbring:
> @jwillenbring: Should this be force merged given that it is affecting many PR and MM builds?

@e10harvey Let's see how it comes back since it is in the middle of testing anyway. It has been testing for a long time, so should report back somewhat soon.

Issue #10342, opened 2022-03-17
Sacado:  Build failures with cuda 11.4


fryeguy52:
## Bug Report
@trilinos/sacado
@jwillenbring 
@e10harvey 

### Description
building the trilinos cuda PR configuration with cuda 11.4 has build failures in sacado

see this [build on cdash](https://testing.sandia.gov/cdash/viewBuildError.php?buildid=10736459)

I am seeing he same failures when run on the sems gpu machine and on the weaver dev queue

### Steps to Reproduce
`sems-son-rhel7-gpu-01` or `sems-srn-rhel7-gpu-01`

#### Environment
```
export TRILINOS_DIR=~/Trilinos
source /projects/sems/modulefiles/utils/sems-modules-init.sh
module purge

module load sems-dev
module load sems-dev-gcc/10.1.0
module load sems-dev-openmpi/4.0.5-cuda-11.4.2
module load sems-dev-cuda/11.4.2

module load sems-dev-cgns/4.2.0
module load sems-dev-netcdf-c/4.8.1
module load sems-dev-netcdf-fortran/4.5.3
module load sems-dev-parmetis/4.0.3
module load sems-dev-hdf5/1.10.7
module load sems-dev-netcdf-cxx/4.2
module load sems-dev-parallel-netcdf/1.12.2
module load sems-dev-superlu-dist/7.1.1
module load sems-dev-boost/1.70.0
module load sems-dev-openblas/0.3.18
module load sems-dev-scotch/6.0.3
module load sems-dev-yaml-cpp/0.6.2
module load sems-dev-metis/5.1.0
module load sems-dev-zlib/1.2.11
module load sems-dev-superlu/5.3.0
module load sems-cmake/3.21.1
module load sems-ninja/1.10.1

################################                                                                                                                                                                                                               

export OMPI_CXX=${TRILINOS_DIR}/packages/kokkos/bin/nvcc_wrapper
export BLAS_ROOT=$OPENBLAS_ROOT
```
#### configure
```
cmake \
  -C "${TRILINOS_DIR}/cmake/std/PullRequestLinuxCuda11.4.2uvmOffTestingSettings.cmake" \
  -DTrilinos_ENABLE_Sacado=ON \
  -D Trilinos_ENABLE_ALL_FORWARD_DEP_PACKAGES=OFF \
  -DTrilinos_ENABLE_TESTS:BOOL=ON \
  ${TRILINOS_DIR}
```
#### Build log
```
[...]

PR #10471, opened 2022-04-27
Sacado:  Fix Cuda 11 error about variables being undefined in device code

etphipp:
This fixes a build error with Cuda 11 in Sacado about certain variables being
undefined in device code.  The cause is taking a reference to certain
static constexpr variables, which according to this:

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#const-variables

is not allowed in device code.  Passing by value appears to resolve it.

For issue #10342


ndellingwood (approved):
Thanks @etphipp for the update and tracking down the official NVIDIA documentation!

PR #10474, opened 2022-04-27
tacho: updates to drop using namespace Kokkos in tests

ndellingwood:
Resolve namespacing issues with Kokkos for post-3.6.00 release
See issue kokkos/kokkos#4817

@trilinos/shylu @kyungjoo-kim 

## Motivation

Resolve compilation errors with namespacing issues with kokkos' develop branch (post version 3.6.00)

## Testing

Locally tested with intel/18 build with Trilinos as well as source-override to test with Kokkos' develop branch

### DIFF
--- a/packages/shylu/shylu_node/tacho/unit-test/Tacho_TestDenseLinearAlgebra.hpp
+++ b/packages/shylu/shylu_node/tacho/unit-test/Tacho_TestDenseLinearAlgebra.hpp
@@ -17,8 +17,6 @@
 #include "Tacho_Lapack_External.hpp"
 
 using namespace Tacho;
-using std::abs;
-using Kokkos::abs;
 
 typedef Kokkos::DualView<ValueType**,Kokkos::LayoutLeft,DeviceType> matrix_type;
 
@@ -112,7 +110,7 @@ TEST( DenseLinearAlgebra, team_gemm_nn ) {
   const MagnitudeType eps = std::numeric_limits<MagnitudeType>::epsilon() * 1000;
   for (int i=0;i<m;++i)
     for (int j=0;j<n;++j) 
-      EXPECT_NEAR(::abs(C1.h_view(i,j)), ::abs(C2.h_view(i,j)), eps);
+      EXPECT_NEAR(abs(C1.h_view(i,j)), abs(C2.h_view(i,j)), eps);
   TEST_END;
 }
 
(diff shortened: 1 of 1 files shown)


kyungjoo-kim:
I do not understand why this becomes a problem. putting the both std and kokkos abs versions is to handle different complex types when they are instantiated with std complex and kokkos complex. why does this become a problem while it used to be okay previously ? in my opinion, the namespace usage should be orthogonal to kokkos change.

ndellingwood:
@kyungjoo-kim the issue comes up after some updates to kokkos to promote the mathematical functions out of the experimental namespace, some compilers have problems resolving the choice of function, this is the type of output I see:

```
18:35:41 /home/jenkins/blake-new/workspace/KokkosEco_Trilinos_Blake_Intel181163_Serial_Cplx/Trilinos/packages/shylu/shylu_node/tacho/unit-test/Tacho_TestDenseLinearAlgebra.hpp(115): error: more than one instance of overloaded function "abs" matches the argument list:
18:35:41             function "std::abs(double)"
18:35:41             function "Kokkos::abs(double)"
18:35:41             argument types are: (double)
18:35:41         EXPECT_NEAR(::abs(C1.h_view(i,j)), ::abs(C2.h_view(i,j)), eps);
...
```

Here is more details tracking the issue https://github.com/kokkos/kokkos/issues/4817 , I'll also add @dalg24 to the issue to add more clarification if needed about the kokkos changes

ndellingwood:
> in my opinion, the namespace usage should be orthogonal to kokkos change

@kyungjoo-kim I'm not disagreeing, but some compilers (intel for certain) have problems resolving between the kokkos and std namespaced routines

kyungjoo-kim:
@ndellingwood please make sure that it is tested for both std complex and kokkos complex. since you remove both std and kokkos abs using statement, it will be likely that the built-in abs function is used, which might be okay for double and float. currently the test is not instantiated with std complex. I just leave the code that way after I manually tested for the both std complex and kokkos complex. the std complex can be fed from applications even if I do not use it for testing.

kyungjoo-kim (approved):

ndellingwood:
Thanks @kyungjoo-kim , I enabled `Trilinos_ENABLE_COMPLEX_DOUBLE=ON` in testing with Trilinos + kokkos develop branch (which had triggered the errors above), is that sufficient for testing?

kyungjoo-kim:
@ndellingwood it should be fine.

ndellingwood:
@dalg24 it's been awhile since I peaked at the kokkos mathematical functions, but under-the-hood something like `Kokkos::abs` falls through to call `std::abs`, in cases of `std::complex` types but may have an overload for `Kokkos::complex` types, or am I misremembering or misunderstanding?

@kyungjoo-kim if my understanding above is correct, is that behavior sufficient for what you need tested?

dalg24:
> @dalg24 it's been awhile since I peaked at the kokkos mathematical functions, but under-the-hood something like `Kokkos::abs` falls through to call `std::abs`, in cases of `std::complex` types but may have an overload for `Kokkos::complex` types, or am I misremembering or misunderstanding?
> 
> @kyungjoo-kim if my understanding above is correct, is that behavior sufficient for what you need tested?

Correct

PR #10479, opened 2022-04-28
IOSS: cgns - remove lambdas to see if fixes stria issue

gsjaardema:
@trilinos/seacas 

## Motivation

Tests are failing on stria build.  Unable to reproduce the issue, but it looks like they are related to the use of lambdas to eliminate some duplicated code.  Removed the lambdas to see if this can fix the failing stria seacas tests.


gsjaardema:
@alanw0 @tokusanya @prwolfe 
This is a test to see if it fixes the failures on Stria that seem to be related to the use of lambdas.  Will be cleaned up if it fixes the issue.  Can't reproduce myself, so seeing if this will fix once committed.

alanw0 (approved):

PR #10488, opened 2022-05-02
Tpetra: Moving selected deep_copy calls back to 2-arg for Panzer

csiefer2:
Issue: #10433
Tests: CUDA
Stakeholder Feedback: Via email

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -1133,7 +1133,6 @@ namespace Tpetra {
    const Kokkos::DualView<const local_ordinal_type*, buffer_device_type>& permuteFromLIDs,
    const CombineMode CM)
   {
-    using execution_space = typename device_type::execution_space;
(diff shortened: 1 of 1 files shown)


rppawlo (approved):
Thanks @csiefer2 !

brian-kelley:
@csiefer2 What actually went wrong with these 3-arg deep copies? The only ways I can think of for a 3-arg and 2-arg to be different behavior is if multiple streams are in use, or if ``execution_space`` was the wrong type (like maybe it was fencing ``Kokkos::Serial()``, when it instead needed to fence ``Kokkos::Cuda()``). But looking at this code and grepping for stream usage, neither seems to be the case here.

brian-kelley:
Maybe downstream is using ``MultiVector<SC, LO, GO, KokkosSerialWrappeNode>``, but the default Kokkos execution space is still being used in its doImport somewhere?

csiefer2:
Panzer tests failed and when Roger and I started reverting to the 2-arg deep_copy those ones are the ones that had to switch back to make the test pass.

PR #10489, opened 2022-05-02
Seacas fix bad cgns symbol detect

gsjaardema:
@trilinos/seacas 

## Motivation

The symbol specifying whether a CGNS library build is parallel or not has moved locations from 3.4.0 (or earlier) to 4.X.  Code in Iocgns_DatabaseIO.C was relying on the new location. Fixed so that it will find the definition for all versions.


gsjaardema:
Note that this reverts the lambda changes from the previous PR #10479 since that was not the issue this time...  I was finally able to reproduce the issue locally and figured out true cause.

rppawlo (approved):

Issue #9957, opened 2021-11-19
MueLu: CoalesceDropFactory vs CoalesceDropFactory_kokkos


mperrinel:
## Bug Report
@trilinos/@muelu:CoalesceDropFactory.cpp>

### Description
CoalesceDropFactory has a parameter "aggregation: greedy Dirichlet" but that is not the case for CoalesceDropFactory_kokkos. 
For the kokkos version, we need to use the "aggregation: preserve Dirichlet points" parameter of UncoupledAggregationFactory_kokkos.

jhux2:
@pohm01

pohm01:
I'm not sure I fully understand. The parameter "aggregation: greedy Dirichlet" and "aggregation: preserve Dirichlet points" do different things. The "aggregation: preserve Dirichlet points" dictates how Dirichlet points are treated for aggregation, where as "aggregation: greedy Dirichlet" is for which nodes are to be considered Dirichlet.

I can look into adding an "aggregation: greedy Dirichlet" option for CoalesceDropFactory_kokkos if that is what we need.

jhux2:
@pohm01 I think @mperrinel is asking whether it's possible to get the option `aggregation: greedy Dirichlet` option for CoalesceDropFactory_kokkos.

PR #10502, opened 2022-05-06
MueLu add greedy dirichlet option for kokkos CoalesceDropFactory

pohm01:
@trilinos/muelu

## Motivation
This addresses #9957 
It also makes Dirichlet boundary detection consistent with the scalar case as well as the non-kokkos version, through the use of `Utilities_kokkos::DetectDirichletRows` to detect Dirichlet nodes.

## Testing
This adds a unit test for greedy dirichlet.


packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_kokkos_def.hpp
```diff
@@ -312,74 +312,86 @@ namespace MueLu {
 
     // build the dof-based column map containing the local dof ids belonging to blkSize rows in matrix
     // the DofIds may not be sorted.
```
jhux2, line 314: I guess this comment is no longer valid?

jhux2 (approved):
lgtm

packages/muelu/src/Graph/MatrixTransformation/MueLu_CoalesceDropFactory_kokkos_def.hpp
pohm01: Yep, forgot to update the comments. Have updated now

jhux2 (approved):
lgtm

PR #9978, opened 2021-11-29
Fix and reapply: TriBITS: Pull in partial refactoring to modern CMake targets (TriBITSPub/TriBITS#299) (#9894, #9972, #9973)

bartlettroscoe:
This PR reverts the revert commit from PR #9977 and therefore reapplies the merge of PR #9894.

This PR branch addresses the issues reported in #9972 and #9973 and listed in  TriBITSPub/TriBITS#433 and offers the chance for developers to try out these changes in their own configurations so we can get this fixed.

In addition to reverting the revert commit from PR #9977, this also pulls in the additional TriBITS PRs (listed most recent to least recent):

* TriBITSPub/TriBITS#459
* TriBITSPub/TriBITS#456
* TriBITSPub/TriBITS#453
* TriBITSPub/TriBITS#452
* TriBITSPub/TriBITS#450
* TriBITSPub/TriBITS#449
* TriBITSPub/TriBITS#447
* TriBITSPub/TriBITS#446
* TriBITSPub/TriBITS#445
* TriBITSPub/TriBITS#444
* TriBITSPub/TriBITS#441
* TriBITSPub/TriBITS#440
* TriBITSPub/TriBITS#439
* TriBITSPub/TriBITS#438
* TriBITSPub/TriBITS#436
* TriBITSPub/TriBITS#435
* TriBITSPub/TriBITS#432

This PR also addresses issues with several application codes.

**NOTE:** This PR breaks backward compatibility of the usage of `<Package>Config.cmake` files from Trilinos.  (See release note for Trilinos 14.0 [here](https://github.com/bartlettroscoe/Trilinos/blob/3281be9b5af148f16d9f507519d3e06b916b675c/RELEASE_NOTES#L7-L73))

## Tasks:

* [x] Reproduce and resolve issues with Albany (#9972 and [internal tribitsrefactoring#1](https://cee-gitlab.sandia.gov/rabartl/tribitsrefactoring/-/issues/1))
* [x] Bring back DEPRECATED non-namespace IMPORTED targets (see https://github.com/TriBITSPub/TriBITS/issues/299#issuecomment-1011558738).
* [x] Reproduce and resolve issues with Nalu (#9973 and [internal tribitsrefactoring#2](https://cee-gitlab.sandia.gov/rabartl/tribitsrefactoring/-/issues/2))
[...]

PR #10507, opened 2022-05-08
Stratimikos: Factory for scalar types != double

cgcgcg:
@trilinos/stratimikos @trilinos/muelu @trilinos/trilinoscouplings @trilinos/panzer 

## Motivation
Stratimikos currently only exposes the `DefaultLinearSolverBuilder` that creates solvers and preconditioners for double operators.

This PR adds `Stratimikos::LinearSolverBuilder<Scalar>` and uses it in MueLu. This allows to remove a lot of special code from MueLu. It seems to be working fine for `Scalar = double`, `float` and `complex<double>`.


bartlettroscoe:
@cgcgcg how is this tested for Scalar tests other than `double`?

cgcgcg:
Currently not at all. I'll have to see if I can even get it to pass the AT builds before I add tests.

bartlettroscoe:
> Currently not at all. I'll have to see if I can even get it to pass the AT builds before I add tests.

@cgcgcg, sounds reasonable. To make that more clear, you might consider prefixing the summary with `WIP: ` to let others know that still PR is still being worked on and not ready to be merged (in case someone approves it and all of the tests pass).

It may be fairly tractable to take the existing testing infrastructure that is currently hard-coded to `double` and template it on the Scalar type to test `float` for example.

Just curious, but what Scalar types are you looking to support, other than just `double` and `float`?  The changes look like it may also be trying instantiate `std::complex<>` scalar types as well (and therefore differentiating `ScalarType` and `MagnitudeType`).

cgcgcg:
I primarily want `float` in order to do some multiprecision work. It would be nice if we could enable other scalar types as well, as that would avoid having workarounds in code that is templated on scalar. For example, we have some MueLu test drivers that either use Belos directly or through the Stratimikos interface. I think I got the issues with complex fixed, but not sure yet about the Sacado types. I don't really want to support that.

bartlettroscoe:
> I think I got the issues with complex fixed, but not sure yet about the Sacado types. I don't really want to support that.

@cgcgcg, are people really wanting to create preconditioners and do linear solves with Sacado differentiation types?  If you are doing differentiation, linear solves (to high enough tolerance) are a linear function between input and output so you should not need (nor should you want) to actually differentiate through the linear solver.  You just do solves involving deltas from the RHSs and you solve for deltas in the solution.

cgcgcg:
MueLu supports it, so I don't want to break it..

etphipp:
Currently we don't support Sacado AD types in any templated solvers.  We do however support Stokhos UQ types and being able to propagate those types through Stratimikos would be useful (and even different types for the linear solver and preconditioner).

cgcgcg:
@etphipp What options do I need to enabled those? I tried enabling Sacado and Stokhos, but it seems that wasn't enough to reproduce the build errors.

etphipp:
As long as you enable Sacado, Tpetra and Muelu, that should do it.  It looks like the linking errors are due to missing ETI for Thyra on Stokhos scalar types.  I think ETI for Thyra is on by default, but I'm not sure.  ETI for Thyra in Stokhos would have to be added.  Also, we have not propagated Stokhos scalar types through Thyra before at all, so this could be a really deep rabbit hole.  You could also consider just disabling instantiation of the Muelu-Stratimikos stuff in Stokhos (in which case you would have to provide stubs to fix the linking errors that would likely create, unless there is no code in Muelu itself that actually calls it).

cgcgcg:
Ah, I just didn't enable the correct tests :-) I can now reproduce the error.

bartlettroscoe:
>  ETI for Thyra in Stokhos would have to be added.

There are standard ETI macros used in Thyra that would need to be extended or replaced to include explicit instantiations on Stokhos scalar types.  It should be a straightforward process to do that.  (But I worry about an explosion in build times with such complex instantiations.  We need to do more to combat build times in Trilinos involving templated and non-templated code and there are well-proven strategies for doing this.)

etphipp:
The approach I would recommend is calling the macros in new cpp files in Stokhos.  That way they don't increase the build times for Thyra (plus its probably the only feasible way given the current package dependency structure).  We typically generate one instantiation per file, and hope for lots of parallelism in the build to reduce compile times.

bartlettroscoe:
> The approach I would recommend is calling the macros in new cpp files in Stokhos.

That is a different ETI model than what is currently used for Thyra but it is doable.  You would have to duplicate a bunch of the existing `<file_base_name>.cpp` files that contain the explicit instantiations in Thyra and then replace the instantization macros on those files.  It might even be better and more maintainable long term to automatically generate those  `*.cpp` explicit instantiation files for Stokhos.  I would have to study all of the explicit instantiation `*.cpp` files in Thyra and see if that would work.

bartlettroscoe:
@cgcgcg, I will look into this in some detail later today or tomorrow.

packages/stratimikos/test/CMakeLists.txt
```diff
@@ -169,33 +169,72 @@ IF (${PROJECT_NAME}_ENABLE_Tpetra)
...
+  ENDIF()
+
+  IF (Tpetra_INST_COMPLEX_FLOAT AND Teuchos_ENABLE_COMPLEX)
+
+    TRIBITS_ADD_TEST(
+      GaleriDriver
+      NAME "Galeri_complex_float_Jacobi"
+      ARGS "--scalarType=\"complex<float>\" --xml=stratimikos_jacobi.xml"
+      NUM_MPI_PROCS 4
+      )
```
bartlettroscoe, line 235: FYI, there are no versions of this `std::complex<float>` test running in PR testing as (not) shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercount=3&showfilters=1&filtercombine=and&field1=testname&compare1=65&value1=Stratimikos_&field2=testname&compare2=63&value2=Galeri_complex&field3=buildstarttime&compare3=84&value3=now).  It only shows the `std:complex<double>` version.

packages/stratimikos/test/CMakeLists.txt
```diff
@@ -169,33 +169,72 @@ IF (${PROJECT_NAME}_ENABLE_Tpetra)
...
+
+  ASSERT_DEFINED(Teuchos_ENABLE_COMPLEX)
+  IF (Tpetra_INST_COMPLEX_DOUBLE AND Teuchos_ENABLE_COMPLEX)
+
+    TRIBITS_ADD_TEST(
+      GaleriDriver
+      NAME "Galeri_complex_double_Jacobi"
+      ARGS "--scalarType=\"complex<double>\" --xml=stratimikos_jacobi.xml"
+      NUM_MPI_PROCS 4
+      )
```
bartlettroscoe, line 224: We do see this `std::complex<double>` version of the test running in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercount=3&showfilters=1&filtercombine=and&field1=testname&compare1=65&value1=Stratimikos_&field2=testname&compare2=63&value2=Galeri_complex&field3=buildstarttime&compare3=84&value3=now) showing:

<html>
<body>


Site | Build Name | Test Name | Status | Time | Proc Time | Details | Build Time | Processors
-- | -- | -- | -- | -- | -- | -- | -- | --
ascic158 | PR-10507-test-Trilinos_pullrequest_gcc_7.2.0_debug-5607 | Stratimikos_Galeri_complex_double_Jacobi_MPI_4 | Passed | 1s 190ms | 4s 760ms | Completed | 2022-05-10T18:21:36 MDT | 4
ascic166 | PR-10507-test-Trilinos_pullrequest_gcc_8.3.0-7593 | Stratimikos_Galeri_complex_double_Jacobi_MPI_4 | Passed | 300ms | 1s 200ms | Completed | 2022-05-10T18:21:45 MDT | 4


</body>
</html>

bartlettroscoe (approved):
@cgcgcg, this looks good.  Thanks for taking the time to put together this PR!  (It is pretty interesting how minimal the changes to Stratimikos were to allow this.) 

It is also pretty interesting that any changes to Belos at all were needed to support this.  This suggest there has been no testing with these Belos solvers up till now with `std::complex<>` types.

I looked over almost every line of code changed and it looks reasonable.  There are `std::complex<double>` versions of the Stratimikos test driver running shown in [this query](https://testing.sandia.gov/cdash/queryTests.php?project=Trilinos&filtercount=3&showfilters=1&filtercombine=and&field1=testname&compare1=65&value1=Stratimikos_&field2=testname&compare2=63&value2=Galeri_complex&field3=buildstarttime&compare3=84&value3=now) showing:

<html>
<body>


Site | Build Name | Test Name | Status | Time | Proc Time | Details | Build Time | Processors
-- | -- | -- | -- | -- | -- | -- | -- | --
ascic158 | PR-10507-test-Trilinos_pullrequest_gcc_7.2.0_debug-5607 | Stratimikos_Galeri_complex_double_Jacobi_MPI_4 | Passed | 1s 190ms | 4s 760ms | Completed | 2022-05-10T18:21:36 MDT | 4
ascic166 | PR-10507-test-Trilinos_pullrequest_gcc_8.3.0-7593 | Stratimikos_Galeri_complex_double_Jacobi_MPI_4 | Passed | 300ms | 1s 200ms | Completed | 2022-05-10T18:21:45 MDT | 4


</body>
</html>

I can't really see what is going on with these tests because the test output is being cut off at 1024 bytes.  But I will assume these are testing something.

But there are no versions of the tests running with  `std::complex<float>` so just note that those tests have never been run in PR testing.  (Did you run them locally?)

My one concern is that this seems to be breaking backward compatibility in that the free template function `Stratimikos::enableMueLu()` now requires a new `Scalar` argument at the beginning.  Any customer code that is using Stratimikos with MueLu will then break.  All of the calls of this function in Trilinos have been changed in this PR (and there are only a small  handful of these) but that will not fix downstream customer code.

Given that non-backwards compatible changes to Trilinos 'develop' should be allowed starting Thursday, May 19 (see https://github.com/trilinos/Trilinos/pull/9978#issuecomment-1127907122), then you might wait to until then to merge this PR.

packages/panzer/adapters-stk/src/Panzer_STK_SetupLOWSFactory.cpp
```diff
@@ -181,12 +181,12 @@ namespace {
 
     #ifdef PANZER_HAVE_MUELU
     {
-      Stratimikos::enableMueLu<int,panzer::GlobalOrdinal,panzer::TpetraNodeType>(linearSolverBuilder,"MueLu");
-      Stratimikos::enableMueLuRefMaxwell<int,panzer::GlobalOrdinal,panzer::TpetraNodeType>(linearSolverBuilder,"MueLuRefMaxwell");
+      Stratimikos::enableMueLu<double,int,panzer::GlobalOrdinal,panzer::TpetraNodeType>(linearSolverBuilder,"MueLu");
```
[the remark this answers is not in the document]
cgcgcg: Will do!

packages/stratimikos/test/CMakeLists.txt
cgcgcg: Yep, `complex<float>` isn't enabled in any builds, as far as I know.

cgcgcg:
@bartlettroscoe Thank you for doing such a thorough review!

bartlettroscoe (approved):
Re-approving after conflict resolution

PR #10516, opened 2022-05-11
Intrepid2 CellGeometry node-based constructor may take (C,N,D) containers for indexing

GrahamBenHarper:
@trilinos/intrepid2 
@CamelliaDPG 

## Motivation
The `Intrepid2::CellGeometry` class allows `cellToNodes_` to be unallocated. In which case, `nodes_` is expected to be a `(C,N,D)` container for mesh nodes. One of the ways such a `(C,N,D)` container is commonly obtained is through `panzer_stk::STK_Interface.getElementVertices()`, so this also makes it slightly easier to pair with a `panzer_stk::STK_Interface` mesh.

The node-based constructor for straight-edge geometry previously checked the extents of `cellToNodes` to set the number of `numCells_` and `numNodesPerCell_` and vertices per cell. This updates it to check the allocation of `cellToNodes` first and then set `numCells_` and `numNodesPerCell_` using either the `cellToNodes` or `nodes`, accordingly.

## Testing
I tested this in an external code I'm using and my assembled matrix stayed the same after I switched it to use sum factorization with a `CellGeometry` constructed from the `STK_Interface` nodes.

I added a *very minimal* test to check if this constructor works when passed a `(C,N,D)` nodes container for the unit hexahedron with tensor ordering, but I didn't build it with tests turned on since I used it for linking against external codes, so I'm really just hoping to throw it at the autotester for testing. :crossed_fingers:

### DIFF
--- a/packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
+++ b/packages/intrepid2/src/Cell/Intrepid2_CellGeometry.hpp
@@ -169,10 +169,10 @@ namespace Intrepid2
                  SubdivisionStrategy subdivisionStrategy = NO_SUBDIVISION,
                  HypercubeNodeOrdering nodeOrdering = HYPERCUBE_NODE_ORDER_TENSOR);
     
-    /** \brief  Node-based constructor for straight-edged geometry.
+    /** \brief  Node-based constructor for straight-edged geometry. 
         \param [in] cellTopo - the cell topology for all cells.
         \param [in] cellToNodes - (C,LN) container specifying the global index of the local node.
-        \param [in] nodes - (GN,D) container specifying the coordinate weight for the global node in the specified dimension.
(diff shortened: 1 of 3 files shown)


CamelliaDPG (approved):
@GrahamBenHarper This looks great to me.  Thanks!

cgcgcg:
These look like actual build errors.

GrahamBenHarper:
Ahhh I was feeling confident in my ability to add a basic test without building it. It seems like I was close. I'll do my local build with tests turned on and fix this :stuck_out_tongue:

GrahamBenHarper:
Okay it should be fixed now! Here's what I'm seeing in Intrepid2:
```
181: 55. CellGeometry_NodeBased_Constructor_UnitTest ... [Passed] (0.000125 sec)
100% tests passed, 0 tests failed out of 183
```
That was using a typical non-cuda build... so we'll see if the cuda AT build passes, but I think it should...

CamelliaDPG (approved):
Looks good to me!

mperego:
thanks @GrahamBenHarper!

PR #10522, opened 2022-05-13
ShyLU_node, FastILU : fixing performance issue

iyamazaki:
@trilinos/shylu 

## Motivation

This PR contains changes for fixing performance of FastILU

- changes to improves the compute time
- option to run SpTRSV on host for solve

## Testing

Modify the tester to add the change.

## Additional Information

The same changes should be made for FastILDL and FastIC


jhux2 (changes requested):
@iyamazaki Thank you for this enhancement!  Could you please add these options to the Ifpack2 user's manual?

iyamazaki:
@jhux2 That is a good idea. Can you tell me how I do that?

jhux2:
> @jhux2 That is a good idea. Can you tell me how I do that?

The LaTeX source is in `ifpack2/doc/UsersGuide`.

iyamazaki:
Thank you, @jhux2. I'll update the document.

By the way, I update ctest to include new options, and locally test (help from @vqd8a). But, I don't think this solver is enabled in any regular tests (PR or nightly)?

jhux2:
> By the way, I update ctest to include new options, and locally test (help from @vqd8a). But, I don't think this solver is enabled in any regular tests (PR or nightly)?

I'm not sure.  But it's fairly straightforward to add specialized nightly tests.  MueLu and Tpetra already have their own tests in the Experimental track.

packages/shylu/shylu_node/fastilu/src/shylu_fastilu.hpp
```diff
@@ -482,47 +506,75 @@ class FastILUPrec
...
         void numericILU()
         {
-            aVal = ScalarArray("aVal", aColIdx.extent(0));
-            aVal_ = Kokkos::create_mirror(aVal);
+            const Scalar zero = STS::zero();
+            #ifdef FASTILU_TIMER
+            Kokkos::Timer Timer;
+            #endif
+            //Sort each row of A by ColIdx
+            KokkosKernels::sort_crs_matrix<ExecSpace, OrdinalArray, OrdinalArray, ScalarArray>(aRowMapIn, aColIdxIn, aValIn);
```
srajama1, line 522: Can this be in symbolic?

iyamazaki: @srajama1, we internally sort the "graph" in symbolic, but when the user inputs the "matrix" for numeric, the matrix is not sorted? We could insert the numerical value, looking for ColIdx match, but sorting the matrix seems to go faster.

srajama1: Understood. Thanks! Typically the structure should not change for numeric, so the position should be the same, but this is ok.

vqd8a (approved):
The changes look good to me. Thanks @iyamazaki for fixing the performance issue. Tested with Aria matrices. Compute time is significantly faster with these changes.

vqd8a:
@srajama1 Is this okay to merge?

iyamazaki:
Thank you everyone for the reviews!! Just a note for @srajama1, I meant: we internally store the sorted graph and matrix, but our current symbolic doesn't save the "mapping" from the non-sorted input ColIdx to the internal sorted ColIdx. That is why we are sorting again the input matrix in numeric. Not sure if it is clearer but if sort-time becomes significant, I'll look into storing the "map" in symbolic.

srajama1:
@iyamazaki That is reasonable. I will approve it.

srajama1 (approved):

PR #10541, opened 2022-05-18
Fix issue with Clang gpu compilation.

### LINKED ISSUE #10535: Belos:  Clang 14 (with gpus) doesn't like taking a reference to cout in a class variable
## Bug Report
@trilinos/belos 

### Description
I've been trying a build of Trilinos that uses Clang 14.x with a gpu target.

This type of error is replicated in many of the solvers.
e.g.,
https://github.com/trilinos/Trilinos/blob/0bd35b1d0298b79ab6e18619e61fa364c84a6267/packages/belos/src/BelosBlockGmresSolMgr.hpp#L340

```
packages/belos/src/BelosBlockGmresSolMgr.hpp:340:37: error: dynamic initialization is not supported for __device__, __constant__, __shared__, and __managed__ variables.
    static constexpr std::ostream * outputStream_default_ = &std::cout;
                                    ^                       ~~~~~~~~~~
```

I'm not sure what the best approach is to mitigate this.  @brian-kelley any thoughts on a good replacement?

I don't think Sandia has a testbed with Clang 14.x to reproduce this on.

jjellio:
Addresses #10535
@trilinos/belos 

    Fix issue with Clang gpu compilation.
    
    Clang does not like dynamic initialization of a class variable when
    using a gpu/device target. It may be possible to address this in
    other ways; namely adding an explicit initialization when the class ETI
    is processed.  Given that most (every?) solver replicated this default
    pattern, and all use std::cout as the default, this change removes the
    explicit variable with the default variable (which was const anyway).
    
    This potentialy may causes issues with Windows, which was already using
    a workaround to capture a pointer to cout.
    ```
    packages/belos/src/BelosBlockGmresSolMgr.hpp:340:37:
    error: dynamic initialization is not supported for __device__,
           __constant__, __shared__, and __managed__ variables.
        static constexpr std::ostream * outputStream_default_ = &std::cout;
                                        ^                       ~~~~~~~~~~
    ```
Closes #10535 

## Testing
This seems to resolve the issue on the system I am testing on (configures, compiles, and a minimal program runs).

It would be nice to vet this doesn't break Windows builds.

### DIFF
--- a/packages/belos/src/BelosBiCGStabSolMgr.hpp
+++ b/packages/belos/src/BelosBiCGStabSolMgr.hpp
@@ -272,13 +272,6 @@ namespace Belos {
     static constexpr int defQuorum_default_ = 1;
     static constexpr const char * resScale_default_ = "Norm of Initial Residual";
     static constexpr const char * label_default_ = "Belos";
-// https://stackoverflow.com/questions/24398102/constexpr-and-initialization-of-a-static-const-void-pointer-with-reinterpret-cas
-#if defined(_WIN32) && defined(__clang__)
-    static constexpr std::ostream * outputStream_default_ =
-       __builtin_constant_p(reinterpret_cast<const std::ostream*>(&std::cout));
-#else
-    static constexpr std::ostream * outputStream_default_ = &std::cout;
-#endif
 
     // Current solver values.
     MagnitudeType convtol_,achievedTol_;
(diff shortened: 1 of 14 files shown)


brian-kelley (approved):

jhux2:
> It would be nice to vet this doesn't break Windows builds.

Agreed, but since Trilinos has no Windows PR testing or nightly testing, I wouldn't let this hold up your PR.

PR #10545, opened 2022-05-19
Teuchos: use reinterpret_cast instead of invalid implicit cast for co…

kuberry:
@trilinos/teuchos 

## Motivation

Function `LAPACK<int, std::complex<float> >::LARND` in Teuchos_LAPACK.cpp relies on an implicit cast from the return type of `CLARND_F77` or `ZLARND_F77` to `std::complex<float>` and `std::complex<double>`, respectively.

These implicit casts are not in the C++ standards and fail with apple-clang on mac. Usually a user wouldn't see this because the functions are guarded by

```C
#ifdef HAVE_TEUCHOS_LAPACKLARND
```

which is never satisfied by `lapack` provided through XCode.

However, using openblas for the Lapack distribution will evoke this problem.

This PR addresses this through a `reinterpret_cast`.

### DIFF
--- a/packages/teuchos/numerics/src/Teuchos_LAPACK.cpp
+++ b/packages/teuchos/numerics/src/Teuchos_LAPACK.cpp
@@ -115,6 +115,12 @@ namespace Teuchos
 {
   // BEGIN INT, FLOAT SPECIALIZATION IMPLEMENTATION //
 
+  std::complex<float> convert_Fortran_complex_to_CXX_complex(_Complex float val)
+  {
+    return reinterpret_cast<std::complex<float>&>(val);
+    // NOTE: The above is guaranteed to be okay given the C99 and C++11 standards
+  }
+
   void LAPACK<int, float>::PTTRF(const int& n, float* d, float* e, int* info) const
   { SPTTRF_F77(&n,d,e,info); }
 
@@ -523,7 +529,11 @@ namespace Teuchos
 
   // BEGIN INT, DOUBLE SPECIALIZATION IMPLEMENTATION //
 
-
+  std::complex<double> convert_Fortran_complex_to_CXX_complex(_Complex double val)
+  {
+    return reinterpret_cast<std::complex<double>&>(val);
+    // NOTE: The above is guaranteed to be okay given the C99 and C++11 standards
+  }
 
   void LAPACK<int, double>::PTTRF(const int& n, double* d, double* e, int* info) const
   { DPTTRF_F77(&n,d,e,info); }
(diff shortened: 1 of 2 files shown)


bartlettroscoe:
@kuberry, how do you know for sure that a reinterpret cast is will be save and correct in all cases?  Is this not implementation defined by the LAPACK implementation? Note Item 92 "Avoid using reinterpet_cast" from the book:

* Sutter, Herb and Andrei Alexandrescu. ["C++ Coding Standards"](https://bartlettroscoe.github.io/reading-list/#c++_coding_standards_2005). Addison Wesley, 2005

kuberry:
In general, I agree. However, in this case, please consult the section "Array-oriented access" at https://en.cppreference.com/w/cpp/numeric/complex for why this should be okay. Also, https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf section 26.4.

bartlettroscoe:
Actually, I would feel a lot better about this if someone added and used the template function:

```
template <typename S>
std::complex<S> covert_C_complex_to_Cxx_complex(S _Complex val)
{
  return reinterpret_cast<std::complex<S>>(val);
}
// NOTE: The above is guaranteed to be okay given the C99 and C++11 standards!
```

for all of these conversions.

kuberry:
I tried using the templated version, but for `S _Complex`,  S was catching the whole `double _Complex` or whatever type was provided so I specified it as a float and double version.

These two functions are the only ones that return complex numbers in Teuchos_LAPACK.cpp. All of the others have void output with intended output of the function passed by reference.

bartlettroscoe (dismissed):
Okay, this looks great.  It is safe (because it will only convert from `_Complex S` to `std::complex<S>` and that conversion is guaranteed to be safe according to the C++11 standard) and should be 100% portable because it is standard C++11.

kuberry:
Most recent commit moves the definition of `convert_Fortran_complex_to_CXX_complex` from `Teuchos_LAPACK.hpp` to `Teuchos_LAPACK.cpp` so that we don't get duplicate symbols during linking.

bartlettroscoe (commented):
These could have been inlined to avoided duplicate link symbols.  But having definitions in cpp files is fine in cases like this where performance is not an issue.

bartlettroscoe (approved):
As per my comment [above](https://github.com/trilinos/Trilinos/pull/10545#pullrequestreview-980534618), I reapprove this for the same reasons as last time (but this time without duplicate link symbols.

bartlettroscoe:
BTW, this will likely fail the intel-19.0.5 PR build because changing Teuchos will likely trigger the enable of ShyLU_Node and therefore trigger the build error described in #10549.

PR #10550, opened 2022-05-20
Phalanx: Array of Views acceptance test.

rppawlo:
## Motivation

Another use case for v-of-v construction to protect. In this case the outer view is a Kokkos:Array

## Testing

New test added to ViewOfViews tests


rppawlo:
@etphipp - could you reapprove this?

Issue #10549, opened 2022-05-20
ShyLY: Tacho_TestSerial_double.cpp build error in intel-19.0.5 PR builds since 5/18/2022


bartlettroscoe:
## Bug Report

@trilinos/framework, @trilinos/shylu 

### Description

As shown in [this query](https://testing.sandia.gov/cdash/index.php?project=Trilinos&begin=2022-05-18&end=2022-05-20&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=test-rhel7_sems-intel-19.0.5-mpich-3.2-serial_release-debug_static_no-kokkos-arch_no-asan_no-complex_fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables&field2=builderrors&compare2=43&value2=0&field3=buildstarttime&compare3=84&value3=NOW) showing:

![image](https://user-images.githubusercontent.com/1959736/169534272-1159a0fd-1e2a-4b2c-9825-dd94f0965394.png)

the file:

```
packages/shylu/shylu_node/tacho/unit-test/Tacho_TestSerial_double.cpp
```

is failing to build with the build error (for example, as shown [here](https://testing.sandia.gov/cdash/viewBuildError.php?buildid=11012660)):

```
FAILED: packages/shylu/shylu_node/tacho/unit-test/CMakeFiles/Tacho_TestSerialDouble.dir/Tacho_TestSerial_double.cpp.o 
[...]

PR #10560, opened 2022-05-23
Framework: Update Intel 19.0.5 spack-cm PR build

ZUUL42:
@trilinos/framework 

## Motivation

This PR is intended to address some issues that presented themselves when the build was initially released in the PR test set.
Much of that discussion can be found in #10549, but it will also require the fixes that may come from that. It also prompted me to reverted the PR test set to Intel 17 until we can insure this new Intel 19.0.5 can be introduced smoothly.

This also updates the PullRequestIntel19.0.5TestingEnv.sh file to follow what the PR build loads (SEMS spack-cm TPLs), opposed to what the MM build loads (SEMS-archive TPLs). Eventually this build will be migrated to MM as well.

## Testing

[CDash query](https://testing.sandia.gov/cdash/index.php?project=Trilinos&display=project&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=intel-19.0.5&field2=buildstamp&compare2=63&value2=Experimental)

## Additional Information

This is not the same build as the current Master Merge build.
The MM build uses SEMS-archive TPLs and has Scotch disabled. This build uses SEMS spack-cm TPLs and has Scotch reenabled.

### DIFF
--- a/cmake/std/PullRequestLinuxIntel19.0.5TestingSettings.cmake
+++ b/cmake/std/PullRequestLinuxIntel19.0.5TestingSettings.cmake
@@ -7,67 +7,6 @@
 # Usage: cmake -C PullRequestLinuxIntel19.0.5TestingSettings.cmake
 
 set (CMAKE_CXX_STANDARD "14" CACHE STRING "Set C++ standard to C++14")
-#Failing tests under C++14
-#Remove line if test has been fixed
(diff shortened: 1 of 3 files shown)


ZUUL42:
PR no longer needed with changes moving such content.
Content will be useful for migrating to the new system though.
But this can be closed so it isn't unnecessarily merged.

PR #10564, opened 2022-05-24
Panzer: added tpetra tests

mperrinel:


rppawlo (changes requested):
Looks good @mperrinel ! The only change we would need is the option to build for epetra and tpetra tests at the same time. Thanks!

PR #10571, opened 2022-05-25
MueLu LowPrecisionFactory: Fix for other combos of enables Scalar types

cgcgcg:
Enabling double, float and complex<double>, but not complex<float> resulted in linking issues.


jhux2 (approved):
lgtm

PR #10572, opened 2022-05-25
MiniEM: Fix lambda capture issues in evaluators

cgcgcg:
@trilinos/panzer 

## Motivation
Fix EMPIRE's Trilinos promotion.


rppawlo (approved):
Thanks @cgcgcg !

PR #10573, opened 2022-05-25
Convert Panzer to use MeshBuilder instead of constructing BulkData.

alanw0:
@trilinos/<teamName>

### DIFF
--- a/packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_Interface.cpp
+++ b/packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_Interface.cpp
@@ -52,6 +52,7 @@
 #include <stk_mesh/base/Selector.hpp>
 #include <stk_mesh/base/GetEntities.hpp>
 #include <stk_mesh/base/GetBuckets.hpp>
+#include <stk_mesh/base/MeshBuilder.hpp>
 #include <stk_mesh/base/CreateAdjacentEntities.hpp>
 
 // #include <stk_rebalance/Rebalance.hpp>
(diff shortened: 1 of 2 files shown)


rppawlo (approved):
Thanks @alanw0 !

mperego:
@alanw0 what's the motivation for this? I'm wondering if we should do something similar in Albany as well.

mperego:
> > @alanw0 what's the motivation for this? I'm wondering if we should do something similar in Albany as well.
> 
> The motivation is to insulate calling codes from specializations/variations of the BulkData class that we intend to produce. Examples of this include a version of BulkData that doesn't create "upward" mesh connectivities/adjacencies. This would save significant memory for algorithms that don't need those upward connections. Another example would include a version of BulkData that is optimized for certain use-cases such as massive structural mesh modification, like mesh generation, etc., without damaging the performance of existing "bread-and-butter" use-cases like striding through field data for an explicit-dynamics app that isn't modifying the mesh much. This MeshBuilder "factory" approach is intended to insulate calling codes from knowing about child classes, or composition-based configuration changes we make behind the scenes. Calling code will still obtain and use a "BulkData" but don't need to know how it was constructed.
> 
> It does indeed also need to be done in Albany. I sent an email to Irina Tezaur and Alejandro Mota about this but I didn't know that I should have also sent it to you. I will send it to you as it included a couple of powerpoint slides that may be helpful. Feel free to send more questions and concerns.

Thanks @alanw0 for the explanation and the email. I'm sure Irina would have forwarded this to the Albany team soon, but feel free to send Albany related info to me as well.

alanw0:
@jwillenbring could you again help me with finding the cdash result for the cuda build that apparently is failing here?
What is the recipe for finding that so I don't have to keep bugging you? :-)

alanw0:
Gents I don't know why this won't pass the cuda build... but in the meantime I shouldn't have pushed this on the stk-snapshot branch, the stk team has scripts that need to use that branch for other stuff. So I'm going to close this pull-request and re-open it from another branch shortly...

jwillenbring:
@alanw0 Sorry for the slow reply. We are working through multiple issues with our cuda testing machine. There was an upgrade that caused a few hiccups.

PR #10584, opened 2022-05-29
ShyLU_FastILU : fix a bug for non-symmetric matrix

iyamazaki:
@trilinos/shylu 

## Motivation

This PR 
- fixes a bug in FastILU for a non-symmetric matrix.
- also adds an (default) option to call TRSV with unit diagonals (which may be faster, especially when the matrix is very sparse)


srajama1 (approved):
Thanks @iyamazaki !

srajama1 (approved):
Thanks @iyamazaki !

vqd8a (approved):
Thanks @iyamazaki

PR #10590, opened 2022-06-02
Convert Panzer to use MeshBuilder instead of constructing BulkData.

alanw0:
@trilinos/panzer 

## Motivation

This is the conversion that was previously pushed on the stk-snapshot branch, I have now closed that other pull-request and opened this one with the same change and also rebased on today's develop branch.


rppawlo:
@alanw0 - the cuda PR build on vortex was broken last week. Some TPLs the PR tester uses were updated and permissions changed. That was fixed but the new cmake TPL was not configured to allow for submissions to cdash, so that is being worked on now. We are waiting on a PR to get the testing back up and running. @e10harvey has more details on this.

alanw0:
Thanks for the info @rppawlo , we can re-test this after that gets straightened out.

PR #10601, opened 2022-06-06
packages/framework/ini-files: Added

e10harvey:
@trilinos/framework

## Motivation

Decouple ini files from GenConfig and LoadEnv.

## Testing

Tested via Trilinos_PR_testproject and the AutoTester.

### DIFF
--- a/packages/framework/get_dependencies.sh
+++ b/packages/framework/get_dependencies.sh
@@ -12,15 +12,21 @@ git submodule update --init --recursive ${script_dir}/GenConfig
 
 # Update ini file submodules
 git submodule update --init --remote ${script_dir}/srn-ini-files || true
+git submodule update --init --remote ${script_dir}/son-ini-files || true
 
 # Point to srn ini files if they exist
 if [[ ! -z "$(ls ./srn-ini-files)" && "$ini_file_option" == "--srn" ]]; then
     cd GenConfig/deps/LoadEnv/ini_files
     ln -sf ${script_dir}/srn-ini-files/trilinos/framework/environment-specs.ini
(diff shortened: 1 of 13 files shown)


packages/framework/get_dependencies.sh
```diff
@@ -0,0 +1,39 @@
...
+git submodule update --init --remote ${script_dir}/srn-ini-files || true
+git submodule update --init --remote ${script_dir}/son-ini-files || true
+
+# Point to srn ini files if they exist
+if [[ ! -z "$(ls ./srn-ini-files)" && "$ini_file_option" == "--srn" ]]; then
+    cd GenConfig/deps/LoadEnv/ini_files
+    ln -sf ${script_dir}/srn-ini-files/trilinos/framework/environment-specs.ini
+    ln -sf ${script_dir}/srn-ini-files/trilinos/framework/supported-systems.ini
+    ln -sf ${script_dir}/ini-files/supported-envs.ini
+elif [[ ! -z "$(ls ./son-ini-files)" && "$ini_file_option" == "--son" ]]; then
```
jwillenbring, line 23: Why is this an elif and else instead of 3 separate if statements? Wouldn't a user potentially want all of these ini files pulled in?

e10harvey: @jwillenbring: Only one set of ini files may be used at a time.

e10harvey: The user can pull in all the ini files but only point to a single set at a time.

jwillenbring (approved):
After resolving my inquiry, I think this looks fine.

PR #10603, opened 2022-06-07
FROSch: Propagate changes for Scalar!=double to FROSch

cgcgcg:
@trilinos/shylu 

## Motivation
Enable use of Scalar types != double in FROSch by using the Stratimikos interfaces for general scalar type.

## Testing
In my local testing I changed one of the tests that is hard-coded to double to float, and it still passed. I also had to disable Epetra for that. Otherwise I was getting build errors.


iyamazaki (approved):
@cgcgcg. This looks great to me (with all the changes we had in `frosch-half` branch, but now with your Stratimikos changes). Pulling this in (with some conflicts), I managed to build both of our usual FROSch test with `float`, and `frosch-half` without commenting out thyra!! Thank you.

cgcgcg:
Yes, with this change `DefaultLinearSolverBuilder = LinearSolverBuilder<double>`. So in order to use this for other scalar types, `LinearSolverBuilder` has to be used.

PR #10406, opened 2022-04-04
Tacho structure symmetric LU solver for CPU and GPU

kyungjoo-kim:
## Motivation

Two things are considered for this refactoring: 
- AMD build
  - Adding HIP enabling macro and determine the code path for HIP
- LU factorization
  - This is work-in-progress.
- Test without tasking infrastructure
  - On GPUs, we do not want to use async tasking but the infrastructure is still required 
  - AMD does not have the infrastructure. For further future work, we first remove the dependency on tasking.

## Related Issues

* Is blocked by 
  #10383 - Trilinos common package gtest does not work when CUDA is used. This can triggers to fail the PR test.
  #10366 - Kokkos OpenMP tasking is not activated when CUDA RDC is not ON. For now, I have work-around and this is not urgent.

## Stakeholder Feedback

@crdohrm For the incremental development and testing purpose, you can use this branch or wait until it is merged.

## Testing

The code relies on Trilinos gtest package and it does not have use correct the cuda SM flag. This may triggers to fail the PR test as well. For my own test, I override the flag and it passes the unit tests.

PR #10609, opened 2022-06-08
packages/tacho: Specify namespace of abs

e10harvey:
@trilinos/shylu

## Motivation

This build error is blocking master merge.

## Testing

Ran generated mpicxx command and verified that this change fixes the intel-19 master merge build.

### DIFF
--- a/packages/shylu/shylu_node/tacho/unit-test/Tacho_TestDenseLinearAlgebra.hpp
+++ b/packages/shylu/shylu_node/tacho/unit-test/Tacho_TestDenseLinearAlgebra.hpp
@@ -110,7 +110,7 @@ TEST( DenseLinearAlgebra, team_gemm_nn ) {
   const MagnitudeType eps = std::numeric_limits<MagnitudeType>::epsilon() * 1000;
   for (int i=0;i<m;++i)
     for (int j=0;j<n;++j) 
-      EXPECT_NEAR(abs(C1.h_view(i,j)), abs(C2.h_view(i,j)), eps);
+      EXPECT_NEAR(std::abs(C1.h_view(i,j)), std::abs(C2.h_view(i,j)), eps);
   TEST_END;
 }
 
@@ -157,7 +157,7 @@ TEST( DenseLinearAlgebra, team_gemm_nt ) {
   const MagnitudeType eps = std::numeric_limits<MagnitudeType>::epsilon() * 1000;
   for (int i=0;i<m;++i)
     for (int j=0;j<n;++j) 
-      EXPECT_NEAR(abs(C1.h_view(i,j)), abs(C2.h_view(i,j)), eps);
+      EXPECT_NEAR(std::abs(C1.h_view(i,j)), std::abs(C2.h_view(i,j)), eps);
   TEST_END;
 }
 
@@ -197,7 +197,7 @@ TEST( DenseLinearAlgebra, team_gemm_nc ) {
   const MagnitudeType eps = std::numeric_limits<MagnitudeType>::epsilon() * 1000;
   for (int i=0;i<m;++i)
     for (int j=0;j<n;++j) 
-      EXPECT_NEAR(abs(C1.h_view(i,j)), abs(C2.h_view(i,j)), eps);
+      EXPECT_NEAR(std::abs(C1.h_view(i,j)), std::abs(C2.h_view(i,j)), eps);
   TEST_END;
 }
 
@@ -237,7 +237,7 @@ TEST( DenseLinearAlgebra, team_gemm_tn ) {
   const MagnitudeType eps = std::numeric_limits<MagnitudeType>::epsilon() * 1000;
   for (int i=0;i<m;++i)
     for (int j=0;j<n;++j) 
-      EXPECT_NEAR(abs(C1.h_view(i,j)), abs(C2.h_view(i,j)), eps);
+      EXPECT_NEAR(std::abs(C1.h_view(i,j)), std::abs(C2.h_view(i,j)), eps);
   TEST_END;
 }
 
@@ -281,7 +281,7 @@ TEST( DenseLinearAlgebra, team_gemm_tt ) {
   const MagnitudeType eps = std::numeric_limits<MagnitudeType>::epsilon() * 1000;
   for (int i=0;i<m;++i)
     for (int j=0;j<n;++j) 
-      EXPECT_NEAR(abs(C1.h_view(i,j)), abs(C2.h_view(i,j)), eps);
+      EXPECT_NEAR(std::abs(C1.h_view(i,j)), std::abs(C2.h_view(i,j)), eps);
   TEST_END;
 }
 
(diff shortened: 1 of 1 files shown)


kyungjoo-kim:
Putting std might not be right. The same code is used for Kokkos::complex type and std does not take Kokkos. @ndellingwood resolved the namespace issue in #10474. With the PR I thought that the abs works for both std and kokkos complex types. @ndellingwood Could you remind what was the issue that time ?

kyungjoo-kim (changes requested):
We need to talk with @ndellingwood.

kyungjoo-kim:
The way I tested different scalar types with different template instantiation. The previous ``::abs`` with using statement of both abs functions selects the appropriate name space and based on your explanation. Your change to ``abs`` was working fine as it passes the PR test. Why do we encounter this now ? What is the difference on the test bed ? If the current code is legal, then is the problem rooted from the compiler and other environment ? Then, we might need a separate issue. 

I checked the testing code. The code does not test with ``std::complex``. Does is ``std::abs`` working ? One work-around might be to put ``Kokkos::abs`` instead.

kyungjoo-kim:
The #10406 include modification of testing with eti. Can we merge that PR first ?

ndellingwood:
> @ndellingwood: From the [test status comment](https://github.com/trilinos/Trilinos/pull/10602#issuecomment-1149522226), please find the PR reproducer instructions and CDash link

@e10harvey I went to the link to look at the intel/19 issues but all build errors are reported from Zoltan2Sphynx (6), though following the link didn't show any errors https://trilinos-cdash.sandia.gov/viewBuildError.php?buildid=167920

e10harvey:
> @e10harvey I went to the link to look at the intel/19 issues but all build errors are reported from Zoltan2Sphynx (6), though following the link didn't show any errors https://trilinos-cdash.sandia.gov/viewBuildError.php?buildid=167920

@ndellingwood: Yes, there are 2 sets of errors:
1. Zoltan2Sphynx (due to an internal compiler error)
2. The build errors due to multiple overloads of `abs` that this PR aims to resolve

For some reason, there is a 500 internal server error when the automation attempts to submit the `abs` build errors for the intel-19 build results. Interestingly, the same code and machines run all other builds and submit build results successfully; nothing in the intel-19 environment stands out either.

Related to: TRILINOSHD-91

CC: @trilinos/framework

e10harvey:
@ndellingwood, @kyungjoo-kim: I see that this my `std::abs` change here breaks the PR builds when the scalar type is `Kokkos::complex`; how should I proceed to unblock the intel-19 master merge build originally reported via:
https://github.com/trilinos/Trilinos/pull/10486#issuecomment-1116083746

CC: @jwillenbring

ndellingwood:
@e10harvey if a source code modification is needed I believe you can replace the `std` namespace with `Kokkos` which would handle the case for `Kokkos::complex`.
I tried to test the intel/19 build on a system with sems compiler (kokkos-dev-2) but it was not recognized (maybe this requires the "public" .ini file capability when it is available). The reason I tried testing was my previous experience with this issue was that the failure with abs occurred with intel/19 and gcc/6.1 loaded, but did not occur when loading gcc/7.2.
The error message posted in a prior message was very strange because the `std::abs` option that accepts `double` types as input was not listed as an available option to the compiler. The combo of bizarre error, and disappearance of error with a newer gcc I think indicates possible toolchain issue?

@kyungjoo-kim you've indicated in another post that calling `abs` from the Kokkos namespace would be an acceptable change if necessary, but I wanted to add you to this discussion so you are in the loop

e10harvey:
> @e10harvey if a source code modification is needed I believe you can replace the `std` namespace with `Kokkos` which would handle the case for `Kokkos::complex`.

I tried Kokkos::abs as well but intel-19 still would not build properly. @trilinos/shylu, please take it from here.

kyungjoo-kim:
[this commit](https://github.com/trilinos/Trilinos/pull/10406/commits/05dc4bfb63d3c508e6de18240114bc44827ff216) includes the explicit abs usage. This PR can be merged first but I personally like to merge #10406 PR first and it will resolve the namespace issue as well.

e10harvey:
> I tried to test the intel/19 build on a system with sems compiler (kokkos-dev-2) but it was not recognized (maybe this requires the "public" .ini file capability when it is available). The reason I tried testing was my previous experience with this issue was that the failure with abs occurred with intel/19 and gcc/6.1 loaded, but did not occur when loading gcc/7.2.
> The error message posted in a prior message was very strange because the `std::abs` option that accepts `double` types as input was not listed as an available option to the compiler. The combo of bizarre error, and disappearance of error with a newer gcc I think indicates possible toolchain issue?

@ndellingwood: Please follow the instructions noted in the [test status comment](https://github.com/trilinos/Trilinos/pull/10602#issuecomment-1149522226). If you encounter any issues (such as a toolchain issue), please open a ticket with trilinos-help.

kyungjoo-kim:
> > @e10harvey if a source code modification is needed I believe you can replace the `std` namespace with `Kokkos` which would handle the case for `Kokkos::complex`.
> 
> I tried Kokkos::abs as well but intel-19 still would not build properly. @trilinos/shylu, please take it from here.

One of my curiosity is why this namespace issue does not pop up for the first time when the kokkos promotion branch is merged to Trilinos even though it gets through the same PR tests. What is difference of the current test environments from when kokkos promotion branch is merged ?

e10harvey:
> > > @e10harvey if a source code modification is needed I believe you can replace the `std` namespace with `Kokkos` which would handle the case for `Kokkos::complex`.
> > 
> > 
> > I tried Kokkos::abs as well but intel-19 still would not build properly. @trilinos/shylu, please take it from here.
> 
> One of my curiosity is why this namespace issue does not pop up for the first time when the kokkos promotion branch is merged to Trilinos. I thought that it gets through the same PR tests. What is difference of the current test environments from when kokkos promotion branch is merged ?

@kyungjoo-kim: Trilinos PR testing does not run with intel-19, only the more comprehensive master merge testing does. This supports @ndellingwood suspicion of a toolchain problem.

ndellingwood:
@e10harvey just to follow up, it sounds like @kyungjoo-kim preference to resolve the issue as part of PR #10406 , are you okay with that path forward for resolution for the intel/19 failures?

e10harvey:
> @e10harvey just to follow up, it sounds like @kyungjoo-kim preference to resolve the issue as part of PR #10406 , are you okay with that path forward for resolution for the intel/19 failures?

Certainly. Thank you for your help with this.

CC: @kyungjoo-kim, @jwillenbring

e10harvey:
@kyungjoo-kim: The trilinos product leaders would like to prioritize the fix proposed in this PR before we merge #10406 in an effort to get `develop` merged into `master`; would you please open a separate PR in place of #10609 with the required fix?

kyungjoo-kim:
I would just disable the test temporary not to duplicate work in #10406.

PR #10615, opened 2022-06-13
Panzer: Fixing orientations in lazy evaluation path

eric-c-cyr:
The orientations were not properly applied when different worksets
were used on a processor. This was because the orientations vector
was not subindexed to the workset cells. So always the same
cells on the processor were used for defining orientations. This
resulted in failed interpolation schemes.

@trilinos/<teamName>

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_BasisValues2.hpp
+++ b/packages/panzer/disc-fe/src/Panzer_BasisValues2.hpp
@@ -288,7 +288,7 @@ namespace panzer {
     int num_orientations_cells_;
 
     // Orientations object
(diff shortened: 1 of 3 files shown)


eric-c-cyr:
@rppawlo this is the bug fix for the multi-block/orientation issue we were seeing. It was as you suggested the element blocks were not properly iterated. The precise cause is that the orientations array was not properly sub indexed so that only the first elements on the processor were used regardless of the workset. This restructuring fixes that. Note that the basis values now have access to _only_ the orientations they need.

rppawlo (approved):
Thanks @eric-c-cyr !

PR #10616, opened 2022-06-13
Allow working with GenConfig repos with gitdist

bartlettroscoe:
This allows using the tool gitdist to view and work with the GenConfig repos. This makes it easy to see what is going on with these various repos.

NOTE: Having this .gitdist.default file does not impact the git submodule manipulation as defined in .gitsubmodules or
packages/framework/get_dependencies.sh at all.  All it does it provide another easy way to view and work with these repos.

### DIFF
--- a/.gitdist.default
+++ b/.gitdist.default
@@ -6,6 +6,14 @@ packages/ForTrilinos
 package/optika
 packages/mesquite
 packages/xSDKTrilinos
+packages/framework/GenConfig
+packages/framework/GenConfig/deps/SetEnvironment
+packages/framework/GenConfig/deps/DetermineSystem
+packages/framework/GenConfig/deps/KeywordParser
+packages/framework/GenConfig/deps/LoadEnv
+packages/framework/GenConfig/deps/SetProgramOptions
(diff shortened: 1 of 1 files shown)


.gitdist.default
```diff
@@ -6,6 +6,14 @@ packages/ForTrilinos
 package/optika
 packages/mesquite
 packages/xSDKTrilinos
+packages/framework/GenConfig
+packages/framework/GenConfig/deps/SetEnvironment
+packages/framework/GenConfig/deps/DetermineSystem
+packages/framework/GenConfig/deps/KeywordParser
+packages/framework/GenConfig/deps/LoadEnv
+packages/framework/GenConfig/deps/SetProgramOptions
```
e10harvey, line 14:
```suggestion
packages/framework/GenConfig/deps/SetProgramOptions
packages/framework/GenConfig/deps/ConfigParserEnhanced
```

e10harvey (approved):
Thanks, @bartlettroscoe !

e10harvey (approved):

PR #10625, opened 2022-06-16
Tacho - temporary disable test

kyungjoo-kim:
## Motivation

Temporary fix for #10609. A more permanent fix is ins #10406.


e10harvey:
Thank you, @kyungjoo-kim. I'd rather leave the test enabled and merge #10406; can you please point me to the commit(s) in #10406 that addresses this the `abs` overload issue reported here: https://github.com/trilinos/Trilinos/pull/10486#issuecomment-1116083746?

e10harvey (changes requested):
Please pull in the subset of commits from #10406 that address the `abs` overload issue that's blocking master merge.

PR #10628, opened 2022-06-16
Intrepid2 triangle hierarchical bases

CamelliaDPG:
@trilinos/intrepid2

## Motivation
This PR fills in missing H(curl), H(div), and H(vol) hierarchical bases on the triangle.  It also brings the ordering of the existing H^1 basis on the triangle into agreement with the ESEAS implementation of the same basis, placing face basis functions of like polynomial degree together.

Forthcoming PRs will similarly add missing hierarchical bases on the tetrahedron, wedge, and pyramid.

## Testing
Some basic "basis equivalence" tests, which establish that the hierarchical bases span the same space as existing nodal bases, are included here.  Additionally, I have done offline comparison with the ESEAS implementation of each basis, testing up to 8th order, with excellent agreement.

### DIFF
--- a/packages/teuchos/core/src/Teuchos_RCPDecl.hpp
+++ b/packages/teuchos/core/src/Teuchos_RCPDecl.hpp
@@ -1189,6 +1189,18 @@ embeddedObjDeallocArrayDelete(const Embedded &embedded, EPrePostDestruction preP
 template<class T> inline
 RCP<T> rcp(T* p, bool owns_mem = true);
 
+/**
+ * Allocates and constructs an object of type \c T 
+ * passing @p args to its constructor, and returns an object of type
+ * @ref Teuchos::RCP that owns and stores a pointer to it.
+ */
+template <typename T, typename ... Args>
+inline auto make_rcp(Args&& ... args)
+{
+    return Teuchos::rcp(
+        new T(std::forward<Args>(args)...)
+    );
+}
 
 /** \brief Initialize from a raw pointer with a deallocation policy.
  *
--- a/packages/teuchos/core/test/MemoryManagement/RCP_UnitTests.cpp
+++ b/packages/teuchos/core/test/MemoryManagement/RCP_UnitTests.cpp
@@ -333,6 +333,26 @@ TEUCHOS_UNIT_TEST( RCP, rcpFromUndefRef )
   TEST_ASSERT(nonnull(a_rcp));
 }
 
+/**
+ * @test Test @ref Teuchos::make_rcp without constructor argument.
+ */
+TEUCHOS_UNIT_TEST( RCP, make_rcp_no_constructor_arg )
+{
+  Teuchos::RCP<A> a_rcp = Teuchos::make_rcp<A>();
+  TEST_ASSERT(Teuchos::nonnull(a_rcp));
+}
+
+/**
+ * @test Test @ref Teuchos::make_rcp with constructor arguments.
+ */
+TEUCHOS_UNIT_TEST( RCP, make_rcp )
+{
+  Teuchos::RCP<A> a_rcp = Teuchos::make_rcp<A>(1,2);
+  TEST_ASSERT(Teuchos::nonnull(a_rcp));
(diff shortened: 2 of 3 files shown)


mperego (commented):
It looks good to me!
Can you try uncommenting these lines, so that we make sure that the new basis provides all the methods needed for orientations/projections:
https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/unit-test/Projection/test_convergence_TRI.hpp#L628
https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/unit-test/Projection/test_convergence_TRI.hpp#L886
https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/unit-test/Projection/test_convergence_TRI.hpp#L1148

CamelliaDPG:
> It looks good to me!
> Can you try uncommenting these lines, so that we make sure that the new basis provides all the methods needed for orientations/projections:
> https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/unit-test/Projection/test_convergence_TRI.hpp#L628
> https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/unit-test/Projection/test_convergence_TRI.hpp#L886
> https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/unit-test/Projection/test_convergence_TRI.hpp#L1148

Thanks for looking this over, @mperego!  I tried uncommenting those lines and running the test locally (have not pushed the change yet).  I get an exception thrown:
```
[Intrepid2] Error in file /Users/nvrober/secondary-fork-Trilinos/packages/intrepid2/src/Orientation/Intrepid2_OrientationToolsDefCoeffMatrix_HCURL.hpp, line 151
            Test that evaluated to true: !subcellBasisIsHVOL
            >>> ERROR (Intrepid::OrientationTools::getCoeffMatrix_HCURL): subcellBasis function space (1d) is not consistent to cellBasis. 
```
The code around that exception is this:
```
      // edge hcurl is hgrad with gauss legendre points
      switch (subcellDim) {
      case 1: {
        //TODO: Hex, QUAD, TET and TRI element should have the same 1d basis
        if (cellIsHex || cellIsQuad) {
          INTREPID2_TEST_FOR_EXCEPTION( !(subcellBasisIsHGRAD||subcellBasisIsHVOL),
              std::logic_error,
              ">>> ERROR (Intrepid::OrientationTools::getCoeffMatrix_HCURL): "
              "subcellBasis function space (1d) is not consistent to cellBasis.");
        } else if (cellIsTet || cellIsTri) {
          INTREPID2_TEST_FOR_EXCEPTION( !subcellBasisIsHVOL,
              std::logic_error,
              ">>> ERROR (Intrepid::OrientationTools::getCoeffMatrix_HCURL): "
              "subcellBasis function space (1d) is not consistent to cellBasis.");
        }
        break;
      }
```
In the debugger, I see that `subcellBasisIsHGRAD` is true, so if simplices were subject to the same test (`(subcellBasisIsHGRAD||subcellBasisIsHVOL)`, rather than just `subcellBasisIsHVOL`) as hypercubes, we'd get past this assertion.  But I'm a bit vague on what's going on here, both how the subcellBasis flags are getting set, and what they might determine in the code that follows.

Could you please advise?

mperego:
I think the restriction of HCURL on edges should be an HVOL basis. Is it possible that `IntegratedLegendreBasis_HGRAD_LINE` is in fact an HVOL basis, that is, it does not have DoFs associated to the edge end points?

CamelliaDPG:
> I think the restriction of HCURL on edges should be an HVOL basis. Is it possible that `IntegratedLegendreBasis_HGRAD_LINE` is in fact an HVOL basis, that is, it does not have DoFs associated to the edge end points?

Yes, you're right.  `getSubCellRefBasis()` was incorrectly implemented to return an HGRAD basis (a copy-and-paste error: I copied from the HGRAD triangle basis, and never edited).  With that corrected the test now passes for me (locally).  Will push the changes in a moment.

CamelliaDPG:
Oops! Did a destructive rebase.  Will restore from backup and fix shortly.

PR #10632, opened 2022-06-17
FastILU : reshuffle symbolic

iyamazaki:
@trilinos/shylu 

## Motivation

This PR re-organizes symbolic phase of FastILU to reduce the setup time. It also adjust some default parameters.


vqd8a (approved):
Thanks @iyamazaki !

PR #10644, opened 2022-06-21
Ifpack2: general block smoothing for MueLu

rstumin:
Auto-PR for SHA bf7f618


csiefer2 (approved):
Thanks for walking me through your changes!

Issue #10642, opened 2022-06-21
Teuchos:  `make_rcp`


romintomasetti:
## Enhancement
@trilinos/teuchos 

Oftentimes in our code, we create a `Teuchos::RCP` as follows:
```c++
auto ptr = Teuchos::rcp(new MyClass(... some args ...));
```

Similarly to `std::make_shared`, it would be nice to have:
```c++
/**
 * Allocates and constructs an object of type \c T 
 * passing @p args to its constructor, and returns an object of type
 * @ref Teuchos::RCP that owns and stores a pointer to it.
 */
template <typename T, typename ... Args>
inline auto make_rcp(Args&& ... args)
{
    return Teuchos::rcp(
        new T(std::forward<Args>(args)...)
    );
}
```

such that one could write:
```c++
auto ptr = Teuchos::make_rcp<MyClass>(... some args ...);
```

I could make a PR to add the above code in `Teuchos_RCP.hpp`. What do you think ?

Thanks !

bartlettroscoe:
NOTE: This is a duplicate of #1016 that was closed due to in-activity.

If you want to supply a PR with some tests that adds this, please do and I will review.

romintomasetti:
@bartlettroscoe I think this issue can be safely closed, as the proposed PR was accepted.

bartlettroscoe:
With the merge of #10647, this can be closed.

PR #10647, opened 2022-06-22
Teuchos : make_rcp (#10642)

romintomasetti:
This PR addresses:
* https://github.com/trilinos/Trilinos/issues/10642

It introduces `Teuchos::make_rcp`, similar to `std::make_shared`.


bartlettroscoe (approved):
Looks great.  And thanks for adding some unit tests!

As a later optimization, we can refactor Teuchos::RCP to bundle the storage for the RCP node and the object into a single allocation, like `std::shared_ptr` does with `std::make_shared`

bartlettroscoe:
Thanks for your contribution @romintomasetti!

PR #10651, opened 2022-06-22
Tpetra/KokkosKernels: Fix #10612 (HIP + complex build)

### LINKED ISSUE #10612: Tpetra/Kokkos-kernels: compilation error in HIP build when enabling complex double
## Bug Report
@trilinos/tpetra @trilinos/kokkos-kernels

### Description
What went wrong?  

We are compiling Trilinos with Rocm 5.1.  We are seeing compilation errors when we enable complex doubles (`Trilinos_ENABLE_COMPLEX_DOUBLE`=`ON`).   

The compilation errors look like:

```
~/Trilinos/packages/kokkos-kernels/src/Kokkos_ArithTraits.hpp:1290:12: error: no matching function for call to 'isinf'
    return isinf(real(x)) || isinf(imag(x));
           ^~~~~
~/Trilinos/packages/tpetra/core/src/Tpetra_computeRowAndColumnOneNorms_def.hpp:193:18: note: in instantiation of member function 'Kokkos::Details::ArithTraits<std::complex<double>>::isInf' requested here
        if (KAT::isInf (matrixVal)) {
                 ^
~/Trilinos/packages/tpetra/core/src/Tpetra_computeRowAndColumnOneNorms_def.hpp:675:12: note: in instantiation of function template specialization 'Tpetra::Details::computeLocalRowOneNorms_RowMatrix<std::complex<double>, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>>' requested here
     return computeLocalRowOneNorms_RowMatrix (A);
            ^
 ~/Trilinos/packages/tpetra/core/src/Tpetra_computeRowAndColumnOneNorms_def.hpp:992:26: note: in instantiation of function template specialization 'Tpetra::Details::computeLocalRowOneNorms<std::complex<double>, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace>>' requested here
   auto result = Details::computeLocalRowOneNorms (A);
                          ^
 /opt/rocm-5.1.0/llvm/lib/clang/14.0.0/include/__clang_hip_cmath.h:509:23: note: candidate template ignored: substitution failure [with __T = double]: no type named 'type' in '__hip_enable_if<false, bool>'
 __HIP_OVERLOAD1(bool, isinf)
 ~~~~~~~~~~~~~~~~~~~~~~^~~~~~
 /opt/rocm-5.1.0/llvm/lib/clang/14.0.0/include/__clang_hip_cmath.h:449:7: note: expanded from macro '__HIP_OVERLOAD1'
       __fn(__T __x) {                                                          \
       ^
 /opt/rocm-5.1.0/llvm/lib/clang/14.0.0/include/__clang_hip_cmath.h:93:31: note: candidate function not viable: call to __device__ function from __host__ function
 __DEVICE__ __CONSTEXPR__ bool isinf(float __x) { return ::__isinff(__x); }
                               ^
 /opt/rocm-5.1.0/llvm/lib/clang/14.0.0/include/__clang_hip_cmath.h:94:31: note: candidate function not viable: call to __device__ function from __host__ function
 __DEVICE__ __CONSTEXPR__ bool isinf(double __x) { return ::__isinf(__x); }
                               ^
 In file included from ~/Trilinos/build-with-GNU-ROCm/packages/tpetra/core/src/Tpetra_computeRowAndColumnOneNorms_COMPLEX_DOUBLE_INT_LONG_LONG_SERIAL.cpp:68:
 In file included from ~/Trilinos/packages/tpetra/core/src/Tpetra_computeRowAndColumnOneNorms_decl.hpp:49:
```

Do you have an idea what might fix things?  

The compilation errors arise in instantiations related to `Tpetra_computeRowAndColumnOneNorms_decl.hpp` and the corresponding `_def.hpp` file.  The functions `forEachLocalRowMatrixRow` and `computeLocalRowAndColumnOneNorms_RowMatrix` in these files do computations on host mirrors of views.  When complex doubles are enabled, they call `Kokkos_ArithTraits.hpp` 's specialisations for the `isNan` and `isInf` functions for `std::complex<double>`.  Around line 1290 in `Kokkos_ArithTraits.hpp`, we can see that these specialisations want to use `std::isnan` and `std::isinf`.  However, these uses are guarded by the macro `KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST`.  This macro is defined in `Kokkos_Macros.hpp` and appears to be essentially a proxy for `__HIP_DEVICE_COMPILE__` not being defined.  What appears to be happening is that this macro `KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST` is not defined in the pass that `hipcc` makes over this file when the compilation error arises.  It seems `__HIP_DEVICE_COMPILE__` is defined even though `hipcc` is compiling host code.

One way we found to reproduce these compilation errors in a simple example is to try to compile the following code with `hipcc` (with Trilinos's includes and linker flags):

```cpp
#include <iostream>
#include <complex>

#include "Kokkos_Core.hpp"

double isInf(std::complex<double> x) {
    #if defined(KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST)
        using std::isinf;
    #endif
    return isinf(real(x));
}
int main() {};
``` 

The compilation fails with similar compilation error messages.   It appears that `hipcc` passes over the code twice, once with `__HIP_DEVICE_COMPILE__` defined and once with `__HIP_DEVICE_COMPILE__` not defined.  It appears to be the pass with `__HIP_DEVICE_COMPILE__` defined that causes the problem.  As in the case of the tpetra compilation error, it is not clear to me why `hipcc` wants to pass over this file, seemingly only including host code, with `__HIP_DEVICE_COMPILE__` defined.

It may also be worthwhile to mention that it is not completely clear to me why the macro `KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST` is used in `Kokkos_ArithTraits.hpp` 's specialisations for the `isNan` and `isInf` functions for `std::complex<double>`.  These functions are not marked as Kokkos functions, which appears to suggest that they are compiled only as host code, so it is not completely clear why the guard by the macro is needed.  Removing the guard by the macro might be one way to avoid the compilation errors.  To know whether such a fix would be a good fix, it would probably be important to understand better why the guard by the macro is needed.

brian-kelley:
@trilinos/tpetra @trilinos/kokkos-kernels 

## Motivation

Fixes build errors when using the HIP backend and complex scalars.

## Related Issues

* Closes #10612 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of

### DIFF
--- a/packages/kokkos-kernels/src/Kokkos_ArithTraits.hpp
+++ b/packages/kokkos-kernels/src/Kokkos_ArithTraits.hpp
@@ -1284,10 +1284,7 @@ class ArithTraits<std::complex<RealFloatType> > {
   }
 #else
   static bool isInf(const std::complex<RealFloatType>& x) {
-#ifdef KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST
-    using std::isinf;
-#endif
-    return isinf(real(x)) || isinf(imag(x));
+    return Kokkos::Experimental::isinf(real(x)) || Kokkos::Experimental::isinf(imag(x));
   }
 #endif
 #ifdef KOKKOS_ENABLE_SYCL
@@ -1307,10 +1304,7 @@ class ArithTraits<std::complex<RealFloatType> > {
   }
 #else
   static bool isNan(const std::complex<RealFloatType>& x) {
-#ifdef KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST
-    using std::isnan;
-#endif
-    return isnan(real(x)) || isnan(imag(x));
+    return Kokkos::Experimental::isnan(real(x)) || Kokkos::Experimental::isnan(imag(x));
   }
 #endif
   static mag_type abs(const std::complex<RealFloatType>& x) {
--- a/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_HIP.hpp
+++ b/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_HIP.hpp
@@ -154,7 +154,7 @@ void KernelWrappers<Scalar,LocalOrdinal,GlobalOrdinal,Kokkos::Compat::KokkosHIPW
 
   // Lots and lots of typedefs
   using Teuchos::RCP;
-  typedef typename Tpetra::CrsMatrix<Scalar,LocalOrdinal,GlobalOrdinal,Node>::local_matrix_type KCRS;
+  typedef typename Tpetra::CrsMatrix<Scalar,LocalOrdinal,GlobalOrdinal,Node>::local_matrix_device_type KCRS;
   typedef typename KCRS::device_type device_t;
   typedef typename KCRS::StaticCrsGraphType graph_t;
   typedef typename graph_t::row_map_type::non_const_type lno_view_t;


ndellingwood (approved):
Thanks @brian-kelley !

Cross-referencing your patch to kokkos-kernels 3.6.01 release, kokkos/kokkos-kernels#1447

PR #10658, opened 2022-06-23
Resolve stk build issues

CamelliaDPG:
@trilinos/stk

## Motivation
On Apple clang builds, I get the following error when building stk’s `ReportHandler.cpp`:

```
/opt/homebrew/include/boost/stacktrace/detail/collect_unwind.ipp:33:2: error: "Boost.Stacktrace requires `_Unwind_Backtrace` function. Define `_GNU_SOURCE` macro or `BOOST_STACKTRACE_GNU_SOURCE_NOT_REQUIRED` if _Unwind_Backtrace is available without `_GNU_SOURCE`."
#error "Boost.Stacktrace requires `_Unwind_Backtrace` function. Define `_GNU_SOURCE` macro or `BOOST_STACKTRACE_GNU_SOURCE_NOT_REQUIRED` if _Unwind_Backtrace is available without `_GNU_SOURCE`."
```

I’ve found I can resolve this error by defining _GNU_SOURCE prior to the include of boost/stacktrace.hpp (and then undef’ing it immediately afterward).

I also get an error when building `CouplingVersions.cpp`; it appears to be missing an include for `<array>`.

This PR is to fix both those issues.


CamelliaDPG:
> @CamelliaDPG thanks for letting us know. We actually have a fix for the missing include in #10657 already.
> 
> With respect to boost stacktrace, we don't believe it's correct to define _GNU_SOURCE directly in the source file, since that should be defined by the compiler. We'd recommend defining _GNU_SOURCE or BOOST_STACKTRACE_GNU_SOURCE_NOT_REQUIRED through your build script if you know for sure the necessary functions are available in your build environment.
> 
> Alternatively, you can define STK_NO_BOOST_STACKTRACE to completely turn off boost stacktrace and avoid these errors.

Thanks @tasmith4!  I'll try STK_NO_BOOST_STACKTRACE; was not aware of that option.

CamelliaDPG:
For the record (and my own reference), what I ended up doing to resolve the build issues here was to set the Trilinos CMake option `-D TPL_ENABLE_Boost=OFF`.  Probably a blunt instrument, but it worked for my purposes.

alanw0:
That's a valid approach. We've worked hard to remove boost from stk. It's not completely gone, but at least the remaining boost dependency is optional.

PR #10664, opened 2022-06-24
Zoltan2: fix inverse permutation logic

brian-kelley:
Set ``haveInverse_`` to true after computing the reverse permutation of an OrderingSolution. This looks like it might have been a copy-paste bug, since the functions computePerm and computeInverse are identical except for having "perm" and "inverse" swapped.

@trilinos/zoltan2 

## Motivation

``haveInverse_`` should be set to true after computing the reverse permutation. ``havePerm_`` is already true since it's in an ``if(havePerm_)`` branch.


srajama1 (approved):
Thanks @brian-kelley !

PR #10671, opened 2022-06-24
Panzer: Add a communicator parameter to FaceToElement methods

### LINKED ISSUE #10669: Panzer: FaceToElement::initialize() assumes MPI_COMM_WORLD
## Bug Report
@trilinos/panzer 

### Description
panzer::FaceToElement::initialize() uses MPI_COMM_WORLD even if the panzer::ConnectionManager was created with a sub-communicator.

tkordenbrock:
FaceToElement::initialize(ConnectionManager) was hardcoded to use MPI_COMM_WORLD.
 - Added a constructor and initialize() method to the FaceToElement class that
   has a communicator parameter.
 - For backward compatibility, the old constructor and initialize() method now
   pass MPI_COMM_WORLD as a parameter to new methods.
Added initialize() tests for default, explicit MPI_COMM_WORLD and sub-communicator.

@trilinos/panzer 

## Related Issues

* Closes #10669 

## Testing
A unit test was added to test initialization with a default and explicit communicators.


packages/panzer/disc-fe/src/Panzer_FaceToElement.hpp
```diff
@@ -78,10 +78,19 @@ class FaceToElement {
 
   FaceToElement(panzer::ConnManager & conn);
 
-  /** Build the mapping from a mesh topology.
+  FaceToElement(panzer::ConnManager & conn,
+                const Teuchos::RCP<const Teuchos::Comm<int>> comm);
+
+  /** Build the mapping from a mesh topology using MPI_COMM_WORLD.
     */
   void initialize(panzer::ConnManager & conn);
```
eric-c-cyr, line 86: Are you maintaining this for backwards compatiblity?

eric-c-cyr (approved):
Thanks for the change @tkordenbrock

packages/panzer/disc-fe/src/Panzer_FaceToElement.hpp
eric-c-cyr: Lets do that, just to keep things clean.

tkordenbrock: OK.  Since it looks like the autotester is backed up, I will deprecate it now.

eric-c-cyr (approved):
Great

PR #10675, opened 2022-06-25
Tpetra:  Use a consistent MPI tag for DistributorActor

tasmith4:
@trilinos/tpetra 

## Motivation
Clean up MPI tagging in DistributorActor, and remove last vestiges of the deprecated useDistinctTags option.  Enable a "unique tag per Actor" strategy to be easily implemented if necessary.  Clarify what the actual tag values are and how to set them.

## Testing
Existing Tpetra tests that use Distributor directly and indirectly continue to pass.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_DistributorActor.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_DistributorActor.cpp
@@ -42,14 +42,17 @@
 namespace Tpetra {
 namespace Details {
 
-  DistributorActor::DistributorActor() {
+  DistributorActor::DistributorActor()
+    : mpiTag_(getNewMpiTag())
+  {
 #ifdef HAVE_TPETRA_DISTRIBUTOR_TIMINGS
     makeTimers();
 #endif // HAVE_TPETRA_DISTRIBUTOR_TIMINGS
   }
 
   DistributorActor::DistributorActor(const DistributorActor& otherActor)
-    : requests_(otherActor.requests_)
+    : mpiTag_(otherActor.mpiTag_),
+      requests_(otherActor.requests_)
   {
 #ifdef HAVE_TPETRA_DISTRIBUTOR_TIMINGS
     makeTimers();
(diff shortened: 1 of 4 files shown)


packages/tpetra/core/src/Tpetra_Details_DistributorActor.cpp
```diff
@@ -42,14 +42,17 @@
 namespace Tpetra {
 namespace Details {
 
-  DistributorActor::DistributorActor() {
+  DistributorActor::DistributorActor()
+    : mpiTag_(1)
```
cwpearson, line 46: I think the choice of `1` here is arbitrary, if so we could toss in a comment saying so

cwpearson:
This looks generally fine, I added a couple comments with very small suggestions

packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
```diff
@@ -643,9 +636,7 @@ void DistributorPlan::computeReceives()
   const int myRank = comm_->getRank();
   const int numProcs = comm_->getSize();
 
-  // MPI tag for nonblocking receives and blocking sends in this method.
-  const int pathTag = 2;
-  const int tag = getTag(pathTag);
+  const int mpiTag = 0;
```
cwpearson, line 639: I think the choice of `0` is arbitrary, if so let's add a comment saying so

packages/tpetra/core/src/Tpetra_Details_DistributorActor.cpp
tasmith4: @cwpearson I'm strongly opposed to comments unless they're absolutely necessary, I'd much prefer to express myself clearly in code.  What if instead we added `mpiTagDefault = 1` in an anonymous namespace, then use that variable to initialize `DistributorActor::mpiTag_`?  Would that make it clear enough to you that this value was arbitrary?

cwpearson (approved):

csiefer2 (approved):

PR #10698, opened 2022-07-01
Belos: Removes unused Failure classes inheriting from Belos::Error

hkthorn:
@trilinos/belos 

## Motivation

There are many Failure classes that are defined in the iteration
classes and solver managers that are not used.  Many of these are
the result of general code copying during the development of new
solvers.  However, if they are not being used, they should be
removed to improve the general clarity of the code.

## Testing

Mac OSX GCC11 serial and CLANG parallel


cgcgcg (approved):
Some of the exceptions were used, but I'm not sure anyone was actually trying to catch them.

Issue #9655, opened 2021-09-02
Tpetra:  remove unneeded operations from fillComplete


kddevin:
## Enhancement
@trilinos/tpetra

Tpetra's fillComplete does unnecessary work in cases where the graph is not modified.  These operations could be avoided.

Right now, to change the values in a matrix, it is recommended (and in some cases, enforced) that users do 
```
A.resumeFill();
change the values
A.fillComplete();
```

If only the values are changed, fillComplete does not need to recompute any graph metrics, nor check the structure of the graph, etc.  It really needs to do only globalAssemble.    But it does a lot of graph checking and recomputation of metrics.  This slows down operations that should be fast, such as scale and setAllToScalar.  

We could remove the requirement to resumeFill for these operations (as proposed in #9640), but the same performance hit would be felt when users are polite enough to call resumeFill / fillComplete when they use replace*Values or grab the values views and modify the values themselves.  And it won't help users who have the resumefill/fillComplete around scale or setAllToScalar because of the past checks.  (I'll argue that #9640 should still be done.)

Here's an example using a matrix from Xyce.  It does 100 calls to setAllToScalar.  The times for 100 calls are reported.  The fillComplete adds 33% overhead to the total time to setAllToScalar on one processor.  Since Epetra doesn't require the resumeFill and fillComplete, it doesn't add the same overhead.

one processor
```
CRS Init Epetra set all to scalar      7.581 (100)       
CRS Init Tpetra resumeFill             0.000124 (100)          
CRS Init Tpetra setAllToScalar call    7.585 (100)          
CRS Init Tpetra fillComplete           2.576 (100)          
CRS Init Tpetra Total Time            10.16 (100) 
```

four processors
```
CRS Init Epetra set all to scalar                           1.775 (100)       1.951 (100)        2.11 (100)         0.01951 (100)
CRS Init Tpetra resumeFill                                  5.96e-05 (100)    8.613e-05 (100)    0.0001209 (100)    8.613e-07 (100)
CRS Init Tpetra setAllToScalar call                         1.832 (100)       1.932 (100)        2.015 (100)        0.01932 (100)
CRS Init Tpetra FillComplete                                0.9604 (100)      4.217 (100)        5.979 (100)        0.04217 (100)
CRS Init Tpetra Total Time                                  2.94 (100)        6.149 (100)        7.881 (100)        0.06149 (100)
```

I propose that we add a flag modifiedGraph_ to CrsGraph that is  
- set to false in resumeFill;
[...]

PR #10706, opened 2022-07-05
Ifpack2: Speedup additive schwarz, fix #9606

### LINKED ISSUE #9606: Ifpack2: RILUK with overlap iteration counts increase with repeated solves
@jhux2 

Observed on a test problem that with repeated reuse of an Ifpack2 RILUK preconditioner with "schwarz: overlap level" set to 1 led to a gradual increase in the linear solver iteration counts over time. If we restarted the simulation partway through the solver iteration counts immediately dropped back down. Running with "schwarz: overlap level" set to 0 the iteration counts also remain roughly constant for the full simulation. It appears there is some internal state associated with overlap that is not being reset between uses of the preconditioner and causing problems with repeated use. I can provide access to files to reproduce to an appropriate developer.

brian-kelley:
@trilinos/ifpack2 

## Motivation

- Fix #9606: make ``AdditiveSchwarz::compute()`` re-import the overlapping part of the matrix (halo) so that the inner preconditioner can see the new values.
- Fuse LocalFilter, ReorderFilter and SingletonFilter into a single class called "Details::AdditiveSchwarzFilter" that constructs the filtered matrix on device. This benefits AdditiveSchwarz with KokkosKernels ILUK because KokkosKernels needs this filtered matrix to be constructed explicitly (it can't use the Tpetra::RowMatrix interface). 
- Make OverlappingRowMatrix use only the matrix on device; avoid host-device transfers.

## Stakeholder Feedback

Required for Sierra milestone
## Testing

OverlappingRowMatrix and AdditiveSchwarz are tested in Ifpack2.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_AdditiveSchwarzFilter_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_AdditiveSchwarzFilter_def.hpp
@@ -292,7 +292,7 @@ setup(const Teuchos::RCP<const row_matrix_type>& A_unfiltered,
   //Create a serial comm and the map for the final filtered CrsMatrix (each process uses its own local map)
 #ifdef HAVE_IFPACK2_DEBUG
   TEUCHOS_TEST_FOR_EXCEPTION(
-    ! mapPairsAreFitted (A_unfiltered_), std::invalid_argument, "Ifpack2::LocalFilter: "
+    ! mapPairsAreFitted (*A_unfiltered_), std::invalid_argument, "Ifpack2::LocalFilter: "
     "A's Map pairs are not fitted to each other on Process "
     << A_->getRowMap ()->getComm ()->getRank () << " of the input matrix's "
     "communicator.  "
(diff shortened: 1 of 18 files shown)


brian-kelley:
@csiefer2 Like we talked about at today's Tpetra meeting, I just added a debug check in ``CrsGraph::setAllIndices`` to check that columns are in bounds.

csiefer2 (dismissed):

srajama1 (approved):
@brian-kelley Thanks!

As you know, I would like Ifpack2 inner preconditioner to change to directly use Kokkos Crs matrices as opposed to Tpetra/Xpetra matrices. The use cases for distributed inner preconditioners are very theoretical and we might be able to extract more performance if we skip one more layer of abstraction. However, no need to worry about it in this PR yet as it is a big redesign.

brian-kelley:
@vbrunini It's not posting to CDash so I don't know what's failing. I opened a helpdesk ticket about it. But I fixed the actual issues with my changes from the last time it posted, so it might just be random failures now.

jhux2:
I don't see failures in the cdash reports.  Can someone post where we might find the results?

brian-kelley:
@jhux2 The results aren't being posted to CDash correctly, so I opened a ticket about that. Framework gave me the Jenkins output directly. There was a real build error in my changes that didn't happen on my compiler, but I just pushed a fix.

brian-kelley:
@jhux2 @csiefer2 Could you give this a fresh review? I don't think there have been major changes since you each reviewed before.

jhux2:
> Could you give this a fresh review? I don't think there have been major changes since you each reviewed before.

@brian-kelley I'm looking at this now.

packages/ifpack2/src/Ifpack2_Details_AdditiveSchwarzFilter_def.hpp
```diff
@@ -0,0 +1,769 @@
...
+  localMap_ = rcp (new map_type (numLocalRows, 0, localComm, Tpetra::GloballyDistributed));
+  //Create the inner filtered matrix.
+  auto crsParams = rcp(new Teuchos::ParameterList);
+  crsParams->template set<bool>("sorted", true);
+  //NOTE: this is the fastest way to construct A_ - it's created as fillComplete,
+  //and no communication needs to be done since localMap_ uses a local comm.
+  //It does need to copy the whole local matrix to host when DualViews are constructed
+  //(only an issue with non-UVM GPU backends) but this is just unavoidable when creating a Tpetra::CrsMatrix.
+  //It also needs to compute local constants (maxNumRowEntries) but this should be a
+  //cheap operation relative to what this constructor already did.
```
jhux2, line 320: My understanding:   This copy-to-host only happens once, at construction.  The matrix graph is fixed, and if matrix values are later updated, no copy to host occurs.

jhux2 (approved):
lgtm -- have one question that I think I know the answer to

packages/ifpack2/src/Ifpack2_Details_AdditiveSchwarzFilter_def.hpp
brian-kelley: @jhux2 Unfortunately it also copies when I do setAllValues, since that re-constructs the WrappedDualViews in the matrix and graph. I don't think there's any other legal way to replace all the values without doing resumeFill/fillComplete, which I wanted to avoid.

jhux2: Oh, that's too bad.  Perhaps fodder for discussion at a future Tpetra meeting.

brian-kelley: In future, if Karen's suggestion #9655 gets implemented, that would make this really easy because the WrappedDualView will handle it automatically. I already get a ReadWrite view of the matrix values, so I can just change them and then be done. No copying to host or fillComplete needed.

PR #10707, opened 2022-07-05
MueLu: Toggle/SemiCoarsening Additions

rstumin:
Auto-PR for SHA 801ffe0


packages/muelu/src/Transfers/SemiCoarsen/MueLu_SemiCoarsenPFactory_decl.hpp
```diff
@@ -145,7 +145,7 @@ namespace MueLu {
     LO MakeSemiCoarsenP(LO const Ntotal, LO const nz, LO const CoarsenRate, LO const LayerId[],
                                   LO const VertLineId[], LO const DofsPerNode, RCP<Matrix>& Amat,
                                   RCP<Matrix>& P, RCP<const Map>& coarseMap, 
-                                  const RCP<MultiVector> fineNullspace, RCP<MultiVector>& coarseNullspace, RCP<Matrix>& R, bool buildRestriction) const;
+                                  const RCP<MultiVector> fineNullspace, RCP<MultiVector>& coarseNullspace, RCP<Matrix>& R, bool buildRestriction, bool doLinear) const;
```
GrahamBenHarper, line 148: this might be a silly nitpick, should we have something more like `int piecewiseInterpolationDegree` instead? I only mention that in case there's some sick world where we decide to do quadratic, cubic, etc., so we won't have to change this interface again.

packages/muelu/src/Transfers/Petrov-Galerkin-SA/MueLu_PgPFactory_def.hpp
```diff
@@ -214,6 +214,14 @@ namespace MueLu {
       // prolongation factory is in prolongation mode
       Set(coarseLevel, "P", P_smoothed);
 
+      // RfromPFactory used to indicate to TogglePFactory that a factory
+      // capable  or producing R can be invoked later. TogglePFactory
+      // replaces dummy value with an index into it's array of prolongators
+      // pointing to the correct prolongator factory. This is later used by
+      // RfromP_Or_TransP to invoke the prolongatorfactory in RestrictionMode
+      int dummy = 7;
+      Set(coarseLevel,"RfromPfactory", dummy);
```
GrahamBenHarper, line 223: remind me; what does `dummy` do?

GrahamBenHarper (approved):
Looks fine to me. There was only one comment I had about the use of `doLinear` in an interface. It's fine to use inside the function, but might not be very extensible in an interface.

packages/muelu/src/Transfers/Petrov-Galerkin-SA/MueLu_PgPFactory_def.hpp
rstumin: > remind me; what does `dummy` do?
So this is a little bad but the value of dummy is not used. The toggle factor will just check if "RfromPfactory" is set by someone and then toggle factory will then reset the value associated with RfromPfactory to a value indicating which prolongator within the list of toggle prolongators corresponds to PgP Factory.  There should be comments within the toggle factory.

packages/muelu/src/Transfers/SemiCoarsen/MueLu_SemiCoarsenPFactory_decl.hpp
rstumin: This might be a possibility, but I guess we would have to use something like -1 to indicate that this type of geometric interpolation is not being used

GrahamBenHarper: That's fair. It's not a huge deal either way. I've found myself on the receiving end of backwards-incompatible changes in Trilinos quite a bit, so that's the only reason I bring it up, although I'm pretty optimistic that nobody would use MueLu internal functions like this one.

GrahamBenHarper (approved):

PR #10708, opened 2022-07-05
ShyLU FastILU : add option to call Metis for pre-ordering

iyamazaki:
@trilinos/shylu @trilinos/ifpack2 

## Motivation

This PR adds an option to call METIS for pre-ordering before calling FastILU (which often gives a better SpTRSV performance on a GPU).

## Additional Information
It may be better to call METIS in additive Shwarz setup so that it will be available to other solvers.


srajama1 (dismissed):
Thanks @iyamazaki !

Ideally, Ifpack2 should call Zoltan2 instead of calling Metis directly. However, approving this now for the milestone. We can do the clean up later.

srajama1 (approved):
Thanks @iyamazaki !

vqd8a (approved):
Thanks @iyamazaki . LGTM

PR #10709, opened 2022-07-05
Ifpack2: fix build+test with Ifpack2_ENABLE_DEBUG

brian-kelley:
Fix the CMake logic for the Ifpack2 debug option, and the option for QD TPL in Ifpack2.
The cache variables that control cmakedefines in config headers
put the package name in all caps (``$PACKAGE_NAME_UC`` in tribits).

Then fix 2 issues when ``Ifpack2_ENABLE_DEBUG=ON``:

- Bug in LocalFilter debug checks - the map isLocallyFitted checks
were backwards. If ``map2.isLocallyFitted(map1)``, map1 is a subset of map2,
not the other way around. For example, ``colMap.isLocallyFitted(domainMap)`` is
typically true but ``domainMap.isLocallyFitted(colMap)`` is not.
Added a comment to this function in Tpetra::Map to make this easier to understand in future.

- Fix a simple build error in a debug snippet of TriDiContainer.

@trilinos/ifpack2

### DIFF
--- a/packages/ifpack2/CMakeLists.txt
+++ b/packages/ifpack2/CMakeLists.txt
@@ -56,7 +56,7 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
 
 TRIBITS_ADD_OPTION_AND_DEFINE(
   ${PACKAGE_NAME}_ENABLE_DEBUG
-  HAVE_${PACKAGE_NAME}_DEBUG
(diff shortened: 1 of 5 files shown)


csiefer2 (dismissed):

jhux2 (dismissed):
lgtm, thanks!

csiefer2 (approved):

Issue #10715, opened 2022-07-06
Various:  Rename macros named KOKKOS_FOO


ndellingwood:
## Enhancement

Various Trilinos packages name macros of the type KOKKOS_FOO. The KOKKOS_ prefix should be changed to distinguish from macros owned by Kokkos, aligning with kokkos/kokkos#4965

As of SHA d3282210b5c703d124b292ed787d32747ae74c62

```
git grep 'define KOKKOS_' origin/develop | egrep -v 'packages/kokkos(-kernels)?/'
origin/develop:packages/minitensor/src/MiniTensor_Definitions.h:#define KOKKOS_INLINE_FUNCTION  inline
origin/develop:packages/phalanx/src/Phalanx_KokkosViewFactoryFunctor.hpp:#define KOKKOS_VIEW_FACTORY_FUNCTOR_HPP
origin/develop:packages/sacado/src/KokkosExp_View_Fad.hpp:#define KOKKOS_EXPERIMENTAL_VIEW_SACADO_FAD_HPP
origin/develop:packages/sacado/src/KokkosExp_View_Fad.hpp:#define KOKKOS_VIEW_DATA_ANALYSIS_SACADO_FAD( NS ) \
origin/develop:packages/sacado/src/KokkosExp_View_Fad_Contiguous.hpp:#define KOKKOS_EXPERIMENTAL_VIEW_SACADO_FAD_CONTIGUOUS_HPP
origin/develop:packages/sacado/src/Kokkos_DynRankView_Fad.hpp:#define KOKKOS_DYN_RANK_VIEW_SACADO_FAD_HPP
origin/develop:packages/sacado/src/Kokkos_DynRankView_Fad_Contiguous.hpp:#define KOKKOS_DYN_RANK_VIEW_SACADO_FAD_CONTIGUOUS_HPP
origin/develop:packages/sacado/src/Kokkos_LayoutContiguous.hpp:#define KOKKOS_LAYOUT_CONTIGUOUS_HPP
origin/develop:packages/sacado/src/Kokkos_LayoutNatural.hpp:#define KOKKOS_LAYOUT_NATURAL_HPP
origin/develop:packages/sacado/src/Kokkos_ViewFactory.hpp:#define KOKKOS_VIEW_FACTORY_HPP
origin/develop:packages/sacado/src/Kokkos_View_Fad.hpp:#define KOKKOS_VIEW_FAD_HPP
[...]

PR #10717, opened 2022-07-07
Phalanx: rename include guard that used KOKKOS_ prefix

rppawlo:
## Motivation

Fix phalanx include guard that was prefixed with "KOKKOS_" as mentioned in #10715 

## Testing

This code change is tested by almost all unit tests in phalanx. It is used by the FieldManager class.


ndellingwood (approved):
Thanks @rppawlo !

PR #10722, opened 2022-07-07
Panzer MiniEM: Fix input file for performance tests

cgcgcg:
@trilinos/panzer

## Motivation
This should fix failures in the performance runs with MiniEM.


cgcgcg:
FYI @brian-kelley

PR #10697, opened 2022-07-01
Ifpack2/Tpetra: Fix some warnings on Cuda

brian-kelley:
- Don't use Teuchos::OrdinalTraits::invalid() in device code
- Fix definition of impl_scalar_type in RBILUK. impl_scalar_type != scalar_type. This was causing std::complex to be used in Tpetra::GEMM which is device code.
- Don't use KOKKOS_LAMBDA for a TeamThread loop body

@trilinos/tpetra 
@trilinos/ifpack2

PR #10723, opened 2022-07-07
Tpetra: address Cuda 10 compiler bug

### LINKED ISSUE #10724: Tpetra: Cuda10 Compiler Bug
As discovered by @GrahamBenHarper.  Need to revert perfectly valid code which tickled the bug.

brian-kelley:
Change TeamThread lambda to use capture-by-value [=] explicitly. Semantically all Kokkos nested loop bodies must capture by value, but we often just make sure captures aren't modified and use [&] to avoid extra copies. In this case, using [&] triggered an NVCC internal compiler error that evaded PR testing. See nightly build error in https://testing.sandia.gov/cdash/viewBuildError.php?buildid=11137806

@trilinos/tpetra 

## Motivation

Compiler bug in #10724

## Related Issues

* Closes #10724
* Blocks 
* Is blocked by 
* Follows #10697 (changes in this PR triggered the internal compiler error) 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback

@csiefer2 says: "As the stakeholder, I approve this fix"

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_unpackCrsMatrixAndCombine_def.hpp
@@ -345,9 +345,11 @@ struct UnpackCrsMatrixAndCombineFunctor {
     }
 
     constexpr bool matrix_has_sorted_rows = true; // see #6282
+    //Note BMK 6-22: this lambda must use capture-by-value [=] and not capture-by-ref [&].
+    //By ref triggers compiler bug in CUDA 10.
     Kokkos::parallel_for(
       Kokkos::TeamThreadRange(team_member, num_entries_in_batch),
-      [&](const LO& j)
+      [=](const LO& j)
       {
         size_t distance = 0;
 


csiefer2 (approved):

rppawlo (approved):

GrahamBenHarper:
Thank you @csiefer2 and @brian-kelley!

brian-kelley:
@kyungjoo-kim In this instance, yes it's an irregular compiler issue. But in general, Kokkos does require that the lambda body be written as if it were const and capture-by-value, meaning you can't modify captured variables. You can still put [&] to make sure the captures aren't copied, but that's just for performance. I found where Christian talks about this in the lecture series: https://youtu.be/s9ecpeWRePs?t=2955

kyungjoo-kim:
@tasmith4 KOKKOS_LAMBDA is equivalent to ``[=] __device__ __host__``, I believe. In this PR, ``[=]``  is used inside of the functor code for regular lambda functions used in team-thread-vector ranges.

PR #10736, opened 2022-07-11
FastILU : metis-chec & skip-sort options

iyamazaki:
@trilinos/shylu 

## Motivation

This PR adds a check to call Metis only with non-empty local matrix, for FastILU, and add an option to skip sorting matrix if the matrix has been already sorted.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_FastILU_Base_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_FastILU_Base_def.hpp
@@ -116,13 +116,15 @@ apply (const Tpetra::MultiVector<Scalar,LocalOrdinal,GlobalOrdinal,Node> &X,
   }
   //zero out applyTime_ now, because the calls to applyLocalPrec() will add to it
   applyTime_ = 0;
-  int nvecs = X.getNumVectors();
+  int  nvecs = X.getNumVectors();
+  auto nrowsX = X.getLocalLength();
+  auto nrowsY = Y.getLocalLength();
   if(nvecs == 1)
   {
     auto x2d = X.getLocalViewDevice(Tpetra::Access::ReadOnly);
     auto y2d = Y.getLocalViewDevice(Tpetra::Access::ReadWrite);
-    ImplScalarArray x1d (const_cast<ImplScalar*>(x2d.data()), x1d.extent(0));
-    ImplScalarArray y1d (const_cast<ImplScalar*>(y2d.data()), y1d.extent(0));
+    ImplScalarArray x1d (const_cast<ImplScalar*>(x2d.data()), nrowsX);
+    ImplScalarArray y1d (const_cast<ImplScalar*>(y2d.data()), nrowsY);
 
     applyLocalPrec(x1d, y1d);
   }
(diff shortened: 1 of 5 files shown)


packages/shylu/shylu_node/fastilu/src/shylu_fastilu.hpp
```diff
@@ -687,8 +688,10 @@ class FastILUPrec
               ExecSpace().fence();
             }
             //Sort each row of A by ColIdx
-            KokkosKernels::sort_crs_matrix<ExecSpace, OrdinalArray, OrdinalArray, ScalarArray>(aRowMapIn, aColIdxIn, aValIn);
-            ExecSpace().fence();
+            if (!skipSortMatrix) {
+              KokkosKernels::sort_crs_matrix<ExecSpace, OrdinalArray, OrdinalArray, ScalarArray>(aRowMapIn, aColIdxIn, aValIn);
```
brian-kelley, line 692: You could also skip sorting the matrix in symbolic above, around line 532.

brian-kelley (changes requested):

packages/ifpack2/src/Ifpack2_Details_FastILU_Base_def.hpp
```diff
@@ -179,64 +181,65 @@ initialize()
...
-          if (metis_colidx(k) != i) {
-            metis_colidx(nnz) = metis_colidx(k);
-            nnz++;
+    if (nrows > 0) {
+      // reorder will convert both graph and perm/iperm to the internal METIS integer type
+      metis_perm_  = MetisArrayHost(Kokkos::ViewAllocateWithoutInitializing("metis_perm"),  nrows);
+      metis_iperm_ = MetisArrayHost(Kokkos::ViewAllocateWithoutInitializing("metis_iperm"), nrows);
+
+      // copy ColInds to host
+      auto localColIndsHost_ = Kokkos::create_mirror(localColInds_);
```
brian-kelley, line 191:
```suggestion
      auto localColIndsHost_ = Kokkos::create_mirror_view(Kokkos::ViewAllocateWithoutInitializing("localColIndsHost_"), localColInds_);
```
create_mirror always allocates a new view, but create_mirror_view will only allocate if localColinds_ is not already host accessible.

packages/shylu/shylu_node/fastilu/src/shylu_fastilu.hpp
brian-kelley: Never mind, that is not sorting the input A but rather the A with fill-in, so I think that one can't be skipped.

This sort also can't be skipped if you're applying the Metis column permutation to ``aColIdxIn``, even if ``skipSortMatrix == true``.

iyamazaki: @brian-kelley. Good point. I was checking this in Ifpack2, but will move it into FastILU. I'll also look at applying Metis ordering in Ifpack2 (as you suggested, but might be in another PR) so that we can skip this sort with Metis also..

packages/ifpack2/src/Ifpack2_Details_FastILU_Base_def.hpp
iyamazaki: I'll go through the code, thank you @brian-kelley!!

packages/ifpack2/src/Ifpack2_Details_Filu_def.hpp
```diff
@@ -101,7 +101,8 @@ initLocalPrec()
 {
   auto nRows = this->mat_->getLocalNumRows();
   auto& p = this->params_;
-  localPrec_ = Teuchos::rcp(new LocalFILU(this->localRowPtrs_, this->localColInds_, this->localValues_, nRows, p.sptrsv_algo,
+  bool skipSortMatrix = !p.use_metis;
```
brian-kelley, line 104: I'm realizing now that this will sometimes skip sorting when it's still required. If the input matrix ``mat_`` is not a Tpetra::CrsMatrix and metis is off, skipSortMatrix will be true but we don't really know that the matrix is sorted.

```suggestion
  bool skipSortMatrix = !rcp_dynamic_cast<const TCrsMatrix>(mat_).is_null() && rcp_dynamic_cast<const TCrsMatrix>(mat_)->getCrsGraph()->isSorted() && !p.use_metis;
```

brian-kelley: You could also put the dynamic cast on its own line.

brian-kelley (changes requested):

brian-kelley (approved):

vqd8a (approved):
Thanks @iyamazaki

PR #10739, opened 2022-07-11
Ifpack2 Relaxation: Don't check for bit-by-bit identity

cgcgcg:
@trilinos/ifpack2 

## Motivation
Relax a bit-by-bit error check. Should fix a failure in the chama nightly:
https://testing.sandia.gov/cdash/test/136817886


jhux2 (approved):
lgtm

PR #10752, opened 2022-07-12
FastILU: skip sorting Ut if using Fast trisolve

brian-kelley:
Make SwapDiag functor support L/Ut matrices with unsorted rows. They
are still written to take O(1) time per row if the matrices are sorted,
though. Then skip sorting Ut if using Fast solver.

@trilinos/shylu 

## Motivation

``sort_crs_matrix(Ut)`` can be an expensive operation - in one problem, it takes 30% of ``FastILUPrec::compute()`` time. But this can be skipped if the triangular solve doesn't require Ut to have sorted rows. Only one small functor had to be changed to remove this requirement from the Fast solver. 

## Testing

Tested by running Ifpack2+Belos driver on a matrix from SuiteSparse. These changes did not change the iteration count. Parameter list:
```
<ParameterList name="test_params">
  <Parameter name="mm_file"     type="string" value="/ascldap/users/bmkelle/MatrixCollection/cfd1.mtx"/>

  <Parameter name="solver_type" type="string" value="Block Gmres"/>
  <ParameterList name="Belos">
    <Parameter name="Num Blocks" type="int" value="40"/>
    <Parameter name="Verbosity" type="int" value="33"/>
    <Parameter name="Output Style" type="int" value="1"/>
    <Parameter name="Output Frequency" type="int" value="1"/>
  </ParameterList>

  <Parameter name="Ifpack2::Preconditioner" type="string" value="SCHWARZ"/>
  <ParameterList name="Ifpack2">
    <Parameter name="schwarz: overlap level" type="int" value="0"/>
    <Parameter name="schwarz: combine mode" type="string" value="ZERO"/>
    <Parameter name="schwarz: use reordering" type="bool" value="false"/>
    <Parameter name="subdomain solver name" type="string" value="FAST_ILU"/>
    <ParameterList name="subdomain solver parameters">
      <Parameter name="metis" type="bool" value="true"/>
      <Parameter name="sweeps" type="int" value="10"/>
      <Parameter name="damping factor" type="double" value="0.2"/>
      <Parameter name="triangular solve type" type="string" value="Fast"/>
      <Parameter name="level" type="int" value="3"/>
    </ParameterList>
  </ParameterList>
</ParameterList>
```

### DIFF
--- a/packages/shylu/shylu_node/fastilu/src/shylu_fastilu.hpp
+++ b/packages/shylu/shylu_node/fastilu/src/shylu_fastilu.hpp
@@ -1201,14 +1201,16 @@ class FastILUPrec
             {}
 
             // just calling SwapDiagTag
-            FastILUPrec_Functor(ScalarArray  lVal_, OrdinalArray  lRowMap_,
-                                ScalarArray utVal_, OrdinalArray utRowMap_,
+            FastILUPrec_Functor(const SwapDiagTag&, ScalarArray  lVal_, OrdinalArray  lRowMap_, OrdinalArray lColIdx_,
+                                ScalarArray utVal_, OrdinalArray utRowMap_, OrdinalArray utColIdx_,
                                 ScalarArray diagElems_) :
             diagElems (diagElems_),
             lVal (lVal_),
             lRowMap (lRowMap_),
+            lColIdx (lColIdx_),
             utVal (utVal_),
-            utRowMap (utRowMap_)
+            utRowMap (utRowMap_),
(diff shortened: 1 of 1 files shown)


iyamazaki (approved):
Thank you, @brian-kelley!!

iyamazaki:
I was assuming Kokkos-Kernels [1463](https://github.com/kokkos/kokkos-kernels/pull/1463) to get patched in before this get merged? Maybe, I should have waited to approve..

PR #10758, opened 2022-07-14
Stokhos: support for custom types in `Sacado::MP::Vector`

romintomasetti:
@trilinos/Stokhos
@etphipp

## What's new in this PR ?

Stokhos is now able to perform basic mathematical operations with ensembles of `std::complex` and `Kokkos::Complex`.

The main fix brought by this PR is the addition of `using std::matop` in the macros of `packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector_ops.hpp`. This brings a behavior that is similar to `packages/sacado/src/Sacado_Fad_Ops.hpp` for instance.

Moreover, the test `packages/stokhos/test/UnitTest/Stokhos_SacadoMPVectorUnitTest.cpp` has been adapted to ensure that `Kokkos::Complex` now works with ensembles.

This PR was already discussed and reviewed in [uliegecsm/Trilinos/1.](https://github.com/uliegecsm/Trilinos/pull/1)

## Side quests

As a side quest, I also added a `Stokhos_UnitBaseTestLib` (static) that contains the `int main(...)` entrypoint. Only the tests that were strictly following the sequence `MPI - Kokkos::init - Test - Kokkos::finalize - return` now link to this *test only* library.

## Notes

This PR needed:
* https://github.com/kokkos/kokkos/pull/5009.

### DIFF
--- a/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector_ops.hpp
+++ b/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector_ops.hpp
@@ -115,7 +115,7 @@ namespace Sacado {
 }
 */
 
-#define MP_UNARYOP_MACRO(OPNAME,OP,OPER)                                \
+#define MP_UNARYOP_MACRO(OPNAME,OP,OPER,USING_OP)                       \
 namespace Sacado {                                                      \
   namespace MP {                                                        \
                                                                         \
@@ -150,22 +150,26 @@ namespace Sacado {                                                      \
                                                                         \
       KOKKOS_INLINE_FUNCTION                                            \
       value_type val() const {                                          \
+        USING_OP                                                        \
         return OPER(expr.val());                                        \
       }                                                                 \
                                                                         \
       KOKKOS_INLINE_FUNCTION                                            \
       value_type coeff(int i) const {                                   \
+        USING_OP                                                        \
         return OPER(expr.coeff(i));                                     \
       }                                                                 \
                                                                         \
       KOKKOS_INLINE_FUNCTION                                            \
       value_type fastAccessCoeff(int i) const {                         \
+        USING_OP                                                        \
         return OPER(expr.fastAccessCoeff(i));                           \
       }                                                                 \
                                                                         \
       template <int i>                                                  \
       KOKKOS_INLINE_FUNCTION                                            \
       value_type getCoeff() const {                                     \
+        USING_OP                                                        \
         return OPER(expr.template getCoeff<i>());                       \
       }                                                                 \
                                                                         \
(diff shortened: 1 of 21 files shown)


etphipp:
@ndellingwood This PR includes a change in Kokkos.  How should that be handled?

ndellingwood:
@etphipp the Kokkos changes should be separated out and submitted as a PR to the kokkos repo for review by the developers [kokkos/kokkos/pulls](https://github.com/kokkos/kokkos/pulls). After merge to the kokkos' repo it will be snapshot into Trilinos for the next release (3.7.00 will be coming soon); make sure to indicate it is needed as part of 3.7.00 as the window for PRs is closing soon

romintomasetti:
@ndellingwood @etphipp The fix in Kokkos was already submitted to Kokkos more than a month ago in https://github.com/kokkos/kokkos/pull/5009...

ndellingwood:
@etphipp the Kokkos changes haven't made it into a release yet so they won't be snapshot into Trilinos until 3.7.00

Adding @crtrott @dalg24 for discussion if approval is needed for the kokkos changes here before the 3.7.00 release (merged into kokkos develop branch as PR kokkos/kokkos#5009 pointed out by @romintomasetti)

etphipp:
I guess my question is, since the changes are already in kokkos-develop, can we not just merge this PR including its kokkos changes, which will then be overwritten at the next snapshot?  Otherwise we will have to wait until the next snapshot to merge this PR.

ndellingwood:
> I guess my question is, since the changes are already in kokkos-develop, can we not just merge this PR including its kokkos changes, which will then be overwritten at the next snapshot?

@etphipp we typically haven't allowed this with kokkos (minus emergency bug fix patches) and prefer to keep any changes to kokkos make it into Trilinos via releases, but adding @dalg24 and @crtrott in case there is urgency here to break from convention

ndellingwood:
This fell off my radar, the kokkos 3.7.00 release has been integrated into Trilinos. It looks like this PR may need rebase on Trilinos develop, but is no longer blocked on the Kokkos changes. Thank you for patience with the delay in the release process

ndellingwood:
@trilinos/stokhos

etphipp:
@romintomasetti Can you resolve the merge conflicts?

romintomasetti:
@ndellingwood No problem! I know how release processes are difficult ;)
@etphipp I just rebased the branch on trilinos/Trilinos@develop :rocket: If you agree with the changes, I will go over it one last time as well ;)

packages/stokhos/test/UnitTest/Stokhos_UnitTestHelpers.hpp
```diff
@@ -142,10 +148,10 @@ namespace Stokhos {
 
     // Compare elements
     for( int i = 0; i < n; ++i ) {
-      ValueType err = std::abs(a1.coeff(i) - a2.coeff(i));
+      ValueType err = KAT::abs(a1.coeff(i) - a2.coeff(i));
```
dalg24, line 151: Why did you change?  `Kokkos::ArithTraits` is being replaced by Kokkos math functions and numeric traits.
This function is not annotated with `KOKKOS_FUNCTION` so `std::abs` is just fine.  If there was an ambiguity, then cast the arguments as necessary.

packages/stokhos/test/UnitTest/Stokhos_UnitTestHelpers.hpp
```diff
@@ -142,10 +148,10 @@ namespace Stokhos {
...
       ValueType tol =
-        abs_tol + rel_tol*std::max(std::abs(a1.fastAccessCoeff(i)),
-                                   std::abs(a2.fastAccessCoeff(i)));
+        abs_tol + rel_tol*std::max(KAT::abs(a1.fastAccessCoeff(i)),
+                                   KAT::abs(a2.fastAccessCoeff(i)));
```
dalg24, line 154: Same comment here, avoid `Kokkos::ArithTraits`

packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector_ops.hpp
```diff
@@ -150,22 +150,26 @@ namespace Sacado {                                                      \
                                                                         \
       KOKKOS_INLINE_FUNCTION                                            \
       value_type val() const {                                          \
+        USING_OP                                                        \
         return OPER(expr.val());                                        \
```
dalg24, line 154: Use Kokkos mathematical functions, don't roll your own.
```suggestion
        return Kokkos::OPER(expr.val());                                        \
```

packages/stokhos/test/UnitTest/Stokhos_SacadoMPVectorUnitTest.cpp
```diff
@@ -47,7 +47,9 @@
 #include "Stokhos_Sacado_Kokkos_MP_Vector.hpp"
 #include "Stokhos_UnitTestHelpers.hpp"
 
-#include <Kokkos_Core.hpp>
+#include "Kokkos_Core.hpp"
+#include "Kokkos_ArithTraits.hpp"
```
dalg24, line 51: Avoid `Kokkos::ArithTraits`

romintomasetti:
@dalg24 Thanks for the review!

Unfortunately, we cannot use `Kokkos::OPER` as you suggest. Indeed, the code must support `double`, `std::complex<double>` and `Kokkos::complex<double>`.

Using `Kokkos::OPER` is generally fine for `double` and `Kokkos::complex<double>` types because:
* for `double` it uses functions from `Kokkos_MathematicalFunctions.hpp`
* for `Kokkos::complex<double>` it uses the ones from `Kokkos_Complex.hpp`

However, it will not compile for `std::complex<double>` as functions in `Kokkos_MathematicalFunctions.hpp` are injected in `Kokkos` namespace only for standard floating point types.

dalg24:
> Unfortunately, we cannot use Kokkos::OPER as you suggest. Indeed, the code must support double, std::complex<double> and Kokkos::complex<double>.

I see.  It is a bit awkward though because it is not technically legal to use `std::complex` in a `KOKKOS_FUNCTION`. It happens to work for a number of combinations of Kokkos backends/options enabled and given toolchain but we can't promise that it compiles warning free and execute properly across all combinations that we currently support.  This may not be an issue in your use case.

Make sure you don't introduce usage of the arithmetic traits.  These are actually defined in KokkosKernels and we would like to deprecate them in the end.

romintomasetti:
> > Unfortunately, we cannot use Kokkos::OPER as you suggest. Indeed, the code must support double, std::complex and Kokkos::complex.
> 
> I see. It is a bit awkward though because it is not technically legal to use `std::complex` in a `KOKKOS_FUNCTION`. It happens to work for a number of combinations of Kokkos backends/options enabled and given toolchain but we can't promise that it compiles warning free and execute properly across all combinations that we currently support. This may not be an issue in your use case.
> 
> Make sure you don't introduce usage of the arithmetic traits. These are actually defined in KokkosKernels and we would like to deprecate them in the end.

1. We had already wrapped the tests for `std::complex<double>` in a `#if !defined(KOKKOS_ENABLE_CUDA) && !defined(KOKKOS_ENABLE_HIP)` region.
2. I now removed the `KokkosArithTraits`.

etphipp:
The changes look good to me.  @dalg24 are your concerns addressed?

dalg24:
> The changes look good to me. @dalg24 are your concerns addressed?

Yes this works.  (I do not have permissions to mark conversations as resolved.)

etphipp (dismissed):

romintomasetti:
Sorry, I also pushed the changes in
* #11033

because otherwise I couldn't compile on our infrastructure.
I just removed that mistake from the branch.
For me, it's ready to go :flight_departure:

etphipp (approved):

maartenarnst:
Hi @etphipp, is it necessary to put the label ‘ AT: PRE-TEST INSPECTED’ label once more, as the autotester bot seems to want us to do?

etphipp:
> Hi @etphipp, is it necessary to put the label ‘ AT: PRE-TEST INSPECTED’ label once more, as the autotester bot seems to want us to do?

Done.  I also added auto-merge, so it should be merged once the autotester thinks it is ready.

romintomasetti:
Hi @etphipp and @dalg24 :wave: 
It seems this PR is still blocked by your bot ;)
I think we need to remove the `trilinos/kokkos-kernels` reviewer. It suppose it was added because the first version of this PR was adding the `log10` function for `Kokkos::complex`. As it no longer touches the `kokkos-kernels` snapshot, maybe it can safely be removed?

etphipp:
@trilinos/framework This PR seems to be stuck.  The autotester says is hasn't been inspected, but the "AT: PRE-TEST INSPECTED" label has already been added.  I tried adding "AT: RETEST" and that didn't seem to help either.

rppawlo:
PR testing is down, see:
https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-225

etphipp:
> PR testing is down, see: https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-225

Thanks for the heads-up @rppawlo!  Unfortunately I don't appear to have permission to view that issue.

PR #10761, opened 2022-07-14
FastILU: skip sorting U^T if using serial trsv

brian-kelley:
@trilinos/shylu 

## Motivation

With https://github.com/kokkos/kokkos-kernels/pull/1463, KokkosSparse::trsv (called StandardHost in FastILU) does not require the input matrix to have sorted rows. So skip sorting U^T when using StandardHost.


jhux2 (approved):
lgtm

brian-kelley:
Now that KokkosKernels 3.7 is in Trilinos, this should be ready to go after a fresh round of testing.

PR #10762, opened 2022-07-14
Add C++17 standard for Clang 10.0.0 PR build

srbdev:
## Motivation

Converting Clang 10.0.0 PR build to use C++17 standard


srbdev:
@jwillenbring I found this
```
//from .ini configuration
CMAKE_CXX_STANDARD:STRING=17
```
from the job's `CMakeCache.txt`

PR #10763, opened 2022-07-14
Add C++17 standard for GCC 7.2.0 and GCC 7.2.0 serial

srbdev:
## Motivation

Converting GCC 7.2.0 and GCC 7.2.0 serial PR builds to use C++17 standard


srbdev:
Checked and the string
```
//from .ini configuration
CMAKE_CXX_STANDARD:STRING=17
```
in the `CMakeCache.txt` appears as well for these two jobs.

jwillenbring (approved):
Looks good. Thanks for verifying the CMakeCache output.

PR #10764, opened 2022-07-14
Add C++17 standard for Python 3 PR build

srbdev:
## Motivation

Converting Python 3 PR build to use C++17 standard


jwillenbring (approved):
Looks good. Thanks for verifying the testing.

PR #10770, opened 2022-07-18
Ifpack2: importing rowmap option

vqd8a:
@trilinos/ifpack2 

## Motivation

This PR adds an option to import a row map from a given matrix market file. It is required for testing for Sierra milestone.


jhux2 (approved):
lgtm

PR #10777, opened 2022-07-19
Ifpack2: fix Schwarz+FastILU sort skipping

brian-kelley:
The AdditiveSchwarzFilter gets passed in directly as the FastILU input matrix, so a dynamic cast to CrsMatrix isn't what we want. Use getCrsMatrix helper instead to extract the filtered&sorted CrsMatrix when using AdditiveSchwarzFilter.

Now, the sort skipping path added in #10736 is taken correctly when calling AdditiveSchwarz with FastILU as inner solver.

@trilinos/ifpack2 
@vbrunini 

## Testing

Tested by running Ifpack2+Belos driver with this configuration, and checking which path is taken:
```
<ParameterList name="test_params">
  <Parameter name="mm_file"     type="string" value="/ascldap/users/bmkelle/MatrixCollection/Laplace3D_100.mtx"/>

  <Parameter name="solver_type" type="string" value="Block Gmres"/>
  <ParameterList name="Belos">
    <Parameter name="Num Blocks" type="int" value="40"/>
    <Parameter name="Verbosity" type="int" value="33"/>
    <Parameter name="Output Style" type="int" value="1"/>
    <Parameter name="Output Frequency" type="int" value="1"/>
  </ParameterList>

  <Parameter name="Ifpack2::Preconditioner" type="string" value="SCHWARZ"/>
  <ParameterList name="Ifpack2">
    <Parameter name="schwarz: overlap level" type="int" value="0"/>
    <Parameter name="schwarz: combine mode" type="string" value="ZERO"/>
    <Parameter name="schwarz: use reordering" type="bool" value="true"/>
    <Parameter name="subdomain solver name" type="string" value="FAST_ILU"/>
    <ParameterList name="subdomain solver parameters">
      <Parameter name="metis" type="bool" value="false"/>
      <Parameter name="sweeps" type="int" value="10"/>
      <Parameter name="damping factor" type="double" value="0.2"/>
      <Parameter name="triangular solve type" type="string" value="StandardHost"/>
      <Parameter name="level" type="int" value="3"/>
    </ParameterList>
  </ParameterList>
</ParameterList>
```

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_Filu_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Filu_def.hpp
@@ -47,6 +47,7 @@
 
 #include "Ifpack2_Details_Filu_decl.hpp"
 #include "Ifpack2_Details_CrsArrays.hpp"
+#include "Ifpack2_Details_getCrsMatrix.hpp"
 #include <Kokkos_Timer.hpp>
 #include <shylu_fastilu.hpp>
 
@@ -102,7 +103,7 @@ initLocalPrec()
   typedef Tpetra::CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> TCrsMatrix;
   auto nRows = this->mat_->getLocalNumRows();
   auto& p = this->params_;
-  auto matCrs = Teuchos::rcp_dynamic_cast<const TCrsMatrix>(this->mat_);
+  auto matCrs = Ifpack2::Details::getCrsMatrix(this->mat_);
 
   bool skipSortMatrix = matCrs && matCrs->getCrsGraph()->isSorted() &&
                        !p.use_metis;
--- a/packages/ifpack2/src/Ifpack2_Details_Filu_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Filu_def.hpp
@@ -105,7 +105,7 @@ initLocalPrec()
   auto& p = this->params_;
   auto matCrs = Ifpack2::Details::getCrsMatrix(this->mat_);
 
-  bool skipSortMatrix = matCrs && matCrs->getCrsGraph()->isSorted() &&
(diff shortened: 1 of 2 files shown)


jhux2 (approved):
lgtm

packages/ifpack2/src/Ifpack2_Details_Filu_def.hpp
```diff
@@ -102,7 +103,7 @@ initLocalPrec()
   typedef Tpetra::CrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> TCrsMatrix;
   auto nRows = this->mat_->getLocalNumRows();
   auto& p = this->params_;
-  auto matCrs = Teuchos::rcp_dynamic_cast<const TCrsMatrix>(this->mat_);
+  auto matCrs = Ifpack2::Details::getCrsMatrix(this->mat_);
 
   bool skipSortMatrix = matCrs && matCrs->getCrsGraph()->isSorted() &&
```
vqd8a, line 108: @brian-kelley Can ```matCrs``` be changed to ```matCrs->is_null()``` here?

vqd8a (approved):
@brian-kelley Thanks for the fix.

packages/ifpack2/src/Ifpack2_Details_Filu_def.hpp
brian-kelley: Yes, I just made that change.

vqd8a (dismissed):
Thanks @brian-kelley

jhux2 (approved):
lgtm

jhux2 (approved):

jwillenbring:
This PR can be force pushed. Since there is an automerge flag on it, I am going to do so. Here is my reasoning:

The last couple PR test runs have been successful except for the new CUDA 11 build, which experienced random failures. The CUDA 11 build passed in a previous PR test run, after the the most recent commits were pushed to this PR.

PR #10785, opened 2022-07-20
Change to C++17 standard for new CUDA 11 build

srbdev:
## Motivation

Update to C++17 standard for the new CUDA 11 build.


jwillenbring:
It seems this build is OK to switch to the C++17 standard. The only failures were the known random failures, and the build was configured in such a way that all the tests were running for all PRs for the new build for the first few days.

fryeguy52 (approved):

PR #10786, opened 2022-07-20
Revert "Add C++17 standard for GCC 7.2.0 and GCC 7.2.0 serial"

jwillenbring:
Reverts trilinos/Trilinos#10763

This appears to possibly be causing issues for PR #10777. We are considering temporarily reverting this commit until the issues can be dealt with.


brian-kelley:
@jwillenbring Thanks, I agree that reverting this for now is a good idea but I'm looking into the failures. The one build error in MueLu was easy to deal with, but now there are errors about RCPs still being alive when tests exit. So I'm trying to replicate those, and then I'll put another PR up with those two fixes.

srbdev (approved):

PR #10788, opened 2022-07-21
cuda 11.4.2 don't run all packages in PR

fryeguy52:
Will not enable all packages on all PR in cuda 11 build
@trilinos/framework


jwillenbring (approved):
Looks good. This allows us to have a UVM on build with CUDA 11 also.

PR #10805, opened 2022-07-28
MueLu RefMaxwell: Fix donut bug

cgcgcg:
@trilinos/muelu 

## Motivation
When the two RefMaxwell subproblems are too large to be rebalanced onto subcommunicators, this needs to be handled to avoid a crash.

### DIFF
--- a/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
+++ b/packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
@@ -500,110 +500,118 @@ namespace MueLu {
 
           // }
 
-          // if (doRebalancing && !reuse) {
-          // rebalance AH
-          RCP<Teuchos::TimeMonitor> tm = getTimer("MueLu RefMaxwell: Rebalance AH");
+          if ((numProcsAH < 0) || (numProcsA22 < 0) || (numProcsAH + numProcsA22 > numProcs)) {
+            GetOStream(Warnings0) << "RefMaxwell::compute(): Disabling rebalancing of subsolves, since partition heuristic resulted "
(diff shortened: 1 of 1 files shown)


packages/muelu/adapters/xpetra/MueLu_RefMaxwell_def.hpp
```diff
@@ -500,110 +500,119 @@ namespace MueLu {
...
+              NullspaceH_ = coarseLevel.Get< RCP<MultiVector> >("Nullspace", newP.get());
+
+            AH_AP_reuse_data_ = Teuchos::null;
+            AH_RAP_reuse_data_ = Teuchos::null;
+
+            if (!disable_addon_ && enable_reuse_) {
+              // Rebalance the addon for next setup
+              RCP<const Import> ImporterH = coarseLevel.Get< RCP<const Import> >("Importer", repartFactory.get());
+              RCP<const Map> targetMap = ImporterH->getTargetMap();
+              SM_Matrix_->getDomainMap()->getComm()->barrier();
```
jhux2, line 611: Why is this necessary?

cgcgcg: We need to make sure that the addon term lives on the same set of MPI ranks as the coarse (1,1) problem. If we do not want to do reuse this isn't an issue, since the matrix for the coarse (1,1) problem will be set up on the global communicator before getting rebalanced to a sub-comm (with addon added it). If we do want reuse, then we have to update that matrix, but the addon part stays the same. We update the coarse matrix on the sub-communicator directly, but we then also need the addon there.

jhux2 (approved):
lgtm

jhux2:
:mind

PR #10815, opened 2022-08-01
MueLu_Utilities_kokkos: GetMatrixDiagonalInverse Kokkos impl for all GO

cgcgcg:
@trilinos/muelu 

## Motivation
- Utilities_kokkos: The double, int, int version was always using the host implementation.
- Update input decks for Stratimikos examples to use deterministic aggregation.


GrahamBenHarper (approved):
Looks good!

I want to double-checking that I'm reading this right. You're changing `GetMatrixDiagonalInverse` so that it's not templated on the same parameters that `UtilitiesBase` is templated on, but instead doing what is most efficient depending on the template parameters of the input matrix, right?

cgcgcg:
No, I changed the specialization for double,int,int,Node to call the same code as Scalar,LocalOrdinal,GlobalOrdinal,Node.

jhux2 (approved):
lgtm, mea culpa

PR #10817, opened 2022-08-01
MueLu: integrate MIS-2 based aggregation

brian-kelley:
In UncoupledAggregationFactory_kokkos, add two MIS-2 based aggregation schemes. Enable with "aggregation: coloring algorithm" = "mis2 coarsening" or "mis2 aggregation". "coarsening" is more aggressive (fewer, larger aggregates), and works the same as ViennaCL's AMG. "aggregation" is less aggressive and more similar to MueLu's serial agg (it creates secondary aggregates similar to Phase2A/2B).

Both are deterministic, and both require a structurally symmetric local matrix (like all D2 coloring based algorithms, so any user of UncoupledAggregationFactory_kokkos was already dealing with this requirement).

Add unit tests for both algorithms.

@trilinos/muelu

## Motivation

Adds a faster and higher quality aggregation scheme for Kokkos refactor. Also deterministic, unlike the existing D2 coloring algorithms that can run on device.

## Testing

Added unit tests, and verified that they are exercising the new code paths and producing valid output.

### DIFF
--- a/packages/muelu/src/Graph/UncoupledAggregation/MueLu_UncoupledAggregationFactory_kokkos_def.hpp
+++ b/packages/muelu/src/Graph/UncoupledAggregation/MueLu_UncoupledAggregationFactory_kokkos_def.hpp
@@ -76,6 +76,7 @@
 
 #include "KokkosGraph_Distance2ColorHandle.hpp"
 #include "KokkosGraph_Distance2Color.hpp"
+#include "KokkosGraph_MIS2.hpp"
 
 namespace MueLu {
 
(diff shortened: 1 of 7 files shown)


packages/muelu/src/Graph/UncoupledAggregation/MueLu_UncoupledAggregationFactory_kokkos_def.hpp
```diff
@@ -234,14 +235,56 @@ namespace MueLu {
...
     }
 
-
     const RCP<const Teuchos::Comm<int> > comm = graph->GetComm();
     GO numGlobalRows = 0;
     if (IsPrint(Statistics1))
       MueLu_sumAll(comm, as<GO>(numRows), numGlobalRows);
 
+    LO numNonAggregatedNodes = numRows;
+    std::string aggAlgo = pL.get<std::string>("aggregation: coloring algorithm");
```
jhux2, line 244: @brian-kelley Thank you for this enhancement!  I don't see this option in muelu/doc/masterlist.xml, which it would need to be in order to parsed from MueLu's ParameterListInterpreter.    If you add the option to `masterlist.xml, then run `update_params.sh` from that directory, it will generate documentation and add it to the parser.

jhux2 (changes requested):

jhux2 (dismissed):
I stand corrected.  `aggregation: coloring algorithm` is in the masterlist.  I'm not sure why I didn't see it the first time.
```
(~/.../packages/muelu) egrep -rl "aggregation: coloring algorithm" *
doc/UsersGuide/masterList.xml
doc/UsersGuide/options_aggregation.tex
doc/UsersGuide/paramlist.tex
doc/UsersGuide/paramlist_hidden.tex
src/Graph/UncoupledAggregation/MueLu_UncoupledAggregationFactory_kokkos_def.hpp
src/Interface/MueLu_ParameterListInterpreter_def.hpp
src/MueCentral/MueLu_MasterList.cpp
src/Transfers/Classical/MueLu_ClassicalMapFactory_def.hpp
src/Transfers/Classical/MueLu_ClassicalPFactory_def.hpp
```

brian-kelley:
@jhux2 
> I stand corrected. `aggregation: coloring algorithm` is in the masterlist. I'm not sure why I didn't see it the first time.

The parameter's name, type and default are documented there now, but is there a different place to document the new choices for its value? I actually forgot to think about documentation for this (shocker!). I know in Ifpack2 the user guide is not auto-generated from XML, and many parameters actually list all possible values and explain what they do. Don't know if MueLu has something similar.

jhux2:
> is there a different place to document the new choices for its value?

The manual is the right place to document the options.  I'd suggest a table with the options and a description of each.  There are examples for some other options, e.g., `multigrid algorithm` on page 40 refers to Table 5.6.  The table reference is in `masterlist.xml`, and the table itself is in `options.tex`.

brian-kelley:
@jhux2 I added a new table to the guide with the different options. It looks like this when built:

image

jhux2:
> I added a new table to the guide with the different options:

@brian-kelley This is great, thank you!

jhux2 (approved):
lgtm

GrahamBenHarper:
> @jhux2 I added a new table to the guide with the different options:

Oh wow that table is awesome! Thanks so much for doing that @brian-kelley!

PR #10818, opened 2022-08-01
MueLu add regionRFactory kokkosed version

pohm01:
@trilinos/muelu 

## Motivation
Adds a kokkos version of RegionRFactory for structured transfer operator construction.

## Testing
Three unit tests have been added to cover this.


mayrmt:
Maybe a stupid question: Since Kokkos is required anyways to build MueLu, why do we add new files for Kokkos and keep old non-Kokkos code as well? This essentially doubles the amount of code to maintain and bears the danger of having two implementations of the same idea, that provide different capabilities, if a new feature is added only to one of them.

GrahamBenHarper:
> why do we add new files for Kokkos and keep old non-Kokkos code as well?

I never quite understood this either. My best guess is some of the algorithms require fundamental changes in order to perform well on device. Since not all of MueLu was written with Kokkos in mind, it was easier to just make Kokkos versions when something has official Kokkos support? Also, there's the issue of how do we let users know if Kokkos is supported for a particular algorithm. I could be totally wrong here.

I think one of the ways we could mitigate this is by creating some sort of document that shows which algorithms support Kokkos and which algorithms don't, and we slowly try to merge things until we eventually reach 100% Kokkos support, but that might take more effort than we have time for.

jhux2:
MueLu still supports Epetra for a while longer.   The split happened when it wasn't possible to always have Kokkos enabled.

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_kokkos_def.hpp
```diff
@@ -0,0 +1,1199 @@
...
+    // we can compute numRows, numCols and NNZ for R
+    const LO numRows = blkSize*lCoarseNodesPerDim[0]*lCoarseNodesPerDim[1]*lCoarseNodesPerDim[2];
+    const LO numCols = blkSize*lFineNodesPerDim[0]*lFineNodesPerDim[1]*lFineNodesPerDim[2];
+
+    // Create the coarse coordinates multivector
+    // so we can fill it on the fly while computing
+    // the restriction operator
+    RCP<Map> rowMap = MapFactory::Build(A->getRowMap()->lib(),
+                                        Teuchos::OrdinalTraits<GO>::invalid(),
+                                        numRows,
```
lucbv, line 244: It this correct? `numRows` is `numCoarseCoordinates*blkSize`, I think you might want `numCoarseCoordinates` here instead?

lucbv (commented):
In general it is okay, this should work fine, except for the coordinate rowmap which looks wrong.
In terms of performance I have doubts. I think you can merge this if you want but I would run some tests on a GPU just to see if that makes sense though.

packages/muelu/src/Transfers/GeneralGeometric/MueLu_RegionRFactory_kokkos_def.hpp
GrahamBenHarper: That looks like something I introduced... :sweat_smile: 
This is used toward the bottom for the construction of R, however, and I think the use of `rowMap` is correct in that instance. I think `coarseCoordinates` is likely accidentally bigger than intended, but nothing else is impacted?

lucbv (dismissed):
Thanks Peter, this looks fine now!

lucbv (approved):
lgtm

PR #10819, opened 2022-08-01
ShyLU FastILU: simple optimizations

brian-kelley:
Remove unneeded Kokkos fences, view initializations. To make sure kernels are still run in proper sequence without the fences, specify the execution space on every RangePolicy (instead of using default space). Also replace a sort_crs_matrix with sort_crs_graph, since in symbolic the matrix values haven't been initialized yet.

Still keep the fences after the top-level functions that users can call directly (initialize, compute, apply) so that initTime/computeTime/applyTime remain accurate. Also keep fences before other timing measurements (but these are guarded by ``#ifdef FASTILU_TIMER``).

@trilinos/shylu 

## Motivation

Performance improvements

## Stakeholder Feedback

This could benefit performance on Sierra milestone problems.
## Testing

Tested by running from the Ifpack2+Belos driver: RCM + FastILU + Host trsv, and METIS + FastILU + FastSptrsv. Ran on Cuda, Cuda+UVM, and Serial. The Serial ran clean under Valgrind, so no uninitialized memory issues.

### DIFF
--- a/packages/shylu/shylu_node/fastilu/src/shylu_fastilu.hpp
+++ b/packages/shylu/shylu_node/fastilu/src/shylu_fastilu.hpp
@@ -440,6 +440,7 @@ class FastILUPrec
         //according to the level of fill
         void symbolicILU()
         {
+            using WithoutInit = Kokkos::ViewAllocateWithoutInitializing;
             #ifdef FASTILU_INIT_TIMER
             Kokkos::Timer timer;
             #endif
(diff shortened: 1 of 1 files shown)


iyamazaki (approved):
Thank you, @brian-kelley. These changes look good to me.

brian-kelley:
@srajama1 @ndellingwood This needs a review by a ShyLU code owner, could one of you please look at it? Thanks!

jennloe:
Or perhaps @egboman @ambrad or @kyungjoo-kim could leave a ShyLu review?

PR #10821, opened 2022-08-02
Adelus: Add factor and solve interfaces + Use handle + Improve backsolve performance

vqd8a:
@trilinos/adelus 

## Motivation

This PR adds new features to enable Adelus to work in the Gemma Schur-complement PCA preconditioner code, including:
- Add factorization interface and solve interface. These are useful for applications that do not have RHS at the time of factorization and would like to solve different RHS with a pre-factorized matrix.
- Allow ```Adelus::Factor``` and ```Adelus::Solve``` to run on GPU and with multiple RHS vectors
- Add ```Adelus::AdelusHandle``` class to allow applications to initialize a handle to the MPI communicator. Adelus can now take any MPI communicator instead of using MPI_COMM_WORLD. The handle also helps remove the use of global variables in the Adelus code
- Improve performance of backsolve algorithm in Adelus
- Update tests, example and README with these new features

## Testing

Tests were performed on Lassen and Weaver, and were confirmed by @jdkotul .


vqd8a:
@srajama1 I just want to give you some updates. These are some new features for Adelus that I *recently* worked on.

jdkotul (approved):
I have gone through the code changes with Vinh and they are appropriate.  The code has been tested on a CPU and GPU platforms and finished successfully.  This was also tested integrated with GEMMA.

srajama1:
@vqd8a Thanks for letting me know! Looks like a great set of changes.

PR #10831, opened 2022-08-04
Amesos2 : support non-standard scalar type for IR

iyamazaki:
@trilinos/amesos2 

## Motivation

This PR tries to fix the errors when building Amesos2 IR with a non-standard scalar type.


ndellingwood (approved):
thanks @iyamazaki !

PR #10829, opened 2022-08-03
MueLu: Enabling BlockCrs through the hierarchy

csiefer2:
Issue:#10748
Tests:Two new ones (which test to make sure we really get BlockCrs matrices in the hierarchy).
Stakeholder feedback: n/a

Thanks to @jhux2 for the design review!

Uses placeholder transpose and spgemm.  Those will get replaced with the real deal at the next KokkosKernels update.

PR #10833, opened 2022-08-04
MueLu: Add MultiVectorAggregationTransferFactory

GrahamBenHarper:
@trilinos/muelu 

This adds a factory for transferring MultiVectors across levels using only aggregates (i.e. not depending on the matrix R). This is a hybrid of `CoordinateTransferFactory` and `MultiVectorTransferFactory`.

### Motivation
Hacking quantities into `CoordinateTransferFactory` works decently, but it's not pretty, and it would be nice to eventually expand this to include multiple different options for transfers. Currently this averages over aggregates, just like `CoordinateTransferFactory`, but doing direct injection from aggregate roots like @rstumin and I have talked about would be a very simple modification here. This is beneficial for multiple ongoing projects, including transferring emin constraints between levels and some trivial matrix-free grid transfers.

### Testing
I added three unit tests and confirmed they work on the usual MueLu experimental build with int/longlong, and it also works on multiple MPI ranks.


jhux2:
git squash?

packages/muelu/src/Misc/MueLu_MultiVectorAggregationTransferFactory_def.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+    // coarseMap is being used to set up the domain map of tentative P, and therefore, the row map of Ac
+    // Therefore, if we amalgamate coarseMap, logical nodes in the coordinates vector would correspond to
+    // logical blocks in the matrix
+
+    ArrayView<const GO> elementAList = coarseMap->getLocalElementList();
+
+    LO                  blkSize      = 1;
+    if (rcp_dynamic_cast<const StridedMap>(coarseMap) != Teuchos::null)
+      blkSize = rcp_dynamic_cast<const StridedMap>(coarseMap)->getFixedBlockSize();
```
jhux2, line 122: Be careful, see PR #10829.  In particular, see [this comment](https://github.com/trilinos/Trilinos/pull/10829/files/44da6788e571d635ba45b751911d65cb0147ac7c#r939453218).  @csiefer2 Could this be a problem?

packages/muelu/src/Misc/MueLu_MultiVectorAggregationTransferFactory_def.hpp
```diff
@@ -0,0 +1,177 @@
...
+    for (LO i = 0; i < Teuchos::as<LO>(numElements); i++)
+      elementList[i] = (elementAList[i*blkSize]-indexBase)/blkSize + indexBase;
+
+    RCP<const Map>   uniqueMap      = fineVector->getMap();
+    RCP<const Map>   coarseMVMap = MapFactory        ::Build(coarseMap->lib(), Teuchos::OrdinalTraits<Xpetra::global_size_t>::invalid(), elementList, indexBase, coarseMap->getComm());
+    coarseVector   = Xpetra::MultiVectorFactory<SC,LO,GO,NO>::Build(coarseMVMap, fineVector->getNumVectors());
+
+    // Create overlapped fine vector to reduce global communication
+    RCP<MultiVector> ghostedVector = fineVector;
+    if (aggregates->AggregatesCrossProcessors()) {
```
jhux2, line 138: We don't generally support aggregates crossing processors (the aggregation is very expensive), but maybe I don't understand your use case well enough.

packages/muelu/test/unit_tests/MultiVectorAggregationTransferFactory.cpp
```diff
@@ -0,0 +1,241 @@
...
+    fineLevel.Set("A",A);
+
+    RCP<UncoupledAggregationFactory> aggFact = rcp(new UncoupledAggregationFactory());
+
+    RCP<MultiVector> fineOnes = MultiVectorFactory::Build(A->getRowMap(),1);
+    fineOnes->putScalar(1.0);
+    fineLevel.Set("onesVector",fineOnes);
+
+    RCP<FactoryManager> M = rcp(new FactoryManager());
+    M->SetKokkosRefactor(false);
```
jhux2, line 114: So there's no Kokkkos version?

packages/muelu/test/unit_tests/MultiVectorAggregationTransferFactory.cpp
```diff
@@ -0,0 +1,241 @@
...
+    coarseLevel.Request("R",RFact.get());
+    coarseLevel.Request("P",TentativePFact.get());
+    */
+
+    /*
+       RCP<MultiVector> coarseOnes = coarseLevel.Get<RCP<MultiVector> >("onesVector",mvtf.get());
+       Teuchos::Array<Teuchos::ScalarTraits<SC>::magnitudeType> vn(1);
+       coarseOnes->norm2(vn);
+
+       TEST_FLOATING_EQUALITY(vn[0]*vn[0],((SC)fineOnes->getGlobalLength()),1e-12);
```
jhux2, line 229: This test doesn't check the result?

jhux2 (commented):
See my comments.

packages/muelu/src/Misc/MueLu_MultiVectorAggregationTransferFactory_decl.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+  public:
+    //! @name Constructors/Destructors.
+    //@{
+
+    //! Constructor.
+    MultiVectorAggregationTransferFactory() { }
+
+    //! Destructor.
+    virtual ~MultiVectorAggregationTransferFactory() { }
```
mayrmt, line 109: Could be removes, as these are just default implementations.

packages/muelu/src/Misc/MueLu_MultiVectorAggregationTransferFactory_decl.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    RCP<const ParameterList> GetValidParameterList() const;
```
mayrmt, line 111: Technically, this is not a constructor/destructor. Hence, the doxygen grouping is not exact here.

packages/muelu/src/Misc/MueLu_MultiVectorAggregationTransferFactory_decl.hpp
```diff
@@ -0,0 +1,142 @@
...
+    This is separate from MultiVectorTransferFactory and CoordinatesTransferFactory
+    as it transfers non-coordinate quantities based only on aggregate information.
+
+  ## Input/output of MultiVectorAggregationTransferFactory ##
+
+  ### User parameters of MultiVectorAggregationTransferFactory ###
+  Parameter | type | default | master.xml | validated | requested | description
+  ----------|------|---------|:----------:|:---------:|:---------:|------------
+  | Vector name| string | "undefined" |   | * | (*) | Name of vector
+  | Vector factory | Factory | null |   | * | (*) | Factory providing vector
```
mayrmt, line 74: `DeclarInput()` also declares `Aggregates` and `CoarseMap` as input. Shouldn't they show up in this table as well?

packages/muelu/src/Misc/MueLu_MultiVectorAggregationTransferFactory_def.hpp
```diff
@@ -0,0 +1,177 @@
...
+    RCP<MultiVector> coarseVector;
+    RCP<MultiVector> fineVector;
+
+    if (coarseLevel.IsAvailable(vectorName, this)) {
+      GetOStream(Runtime0) << "Reusing " << vectorName << std::endl;
+      return;
+    }
+
+    RCP<Aggregates> aggregates = Get< RCP<Aggregates> > (fineLevel, "Aggregates");
+    fineVector = fineLevel.Get<RCP<MultiVector> >(vectorName);
```
mayrmt, line 106: One could remove the declaration `RCP<MultiVector> fineVector` above, as it is not needed up to this point, and then directly declare and initialize this vectors as `const` as follows:
```suggestion
    RCP<const MultiVector> fineVector = fineLevel.Get<RCP<MultiVector> >(vectorName);
```

packages/muelu/src/Misc/MueLu_MultiVectorAggregationTransferFactory_def.hpp
GrahamBenHarper: Oh, good catch! I'll watch out for this

packages/muelu/src/Misc/MueLu_MultiVectorAggregationTransferFactory_def.hpp
GrahamBenHarper: I agree that we generally don't use/support aggregates that cross processors. This is really just a failsafe so that the algorithm doesn't do something unexpected. For example, in a matrix-free context, I absolutely do not want aggregates to cross processors.

packages/muelu/test/unit_tests/MultiVectorAggregationTransferFactory.cpp
GrahamBenHarper: This is probably just an artifact of the unit test that I used to construct this. You know, it's a hybrid of both CoordinateTransferFactory and MultiVectorTransferFactory, so it should "support" Kokkos, but since CoordinateTransferFactory transfers using ArrayRCP interfaces, that's the choice I went with. I don't think setting the Kokkos refactor on or off changes anything, but I'll double-check.

GrahamBenHarper:
Don't autoclose yet. We definitely need a better way to transfer information through the hierarchy than hacking random variables into `CoordinateTransferFactory` each time, but this might not be the right approach.

PR #10860, opened 2022-08-11
Tacho - change hip stream destruction order.

kyungjoo-kim:
## Motivation

hip stream is used when kokkos execution space instances are destroyed. so, the stream should be defalcated afterward. cuda stream does not have this problem even if the stream is deleted first. for consistency, cuda destruction order is also changed.

## Test

Tested on caraway03 with ROCM 5.2

### DIFF
--- a/packages/shylu/shylu_node/tacho/src/impl/Tacho_NumericTools_LevelSet.hpp
+++ b/packages/shylu/shylu_node/tacho/src/impl/Tacho_NumericTools_LevelSet.hpp
@@ -627,12 +627,9 @@ public:
 
   virtual ~NumericToolsLevelSet() {
 #if defined(KOKKOS_ENABLE_CUDA)
-    // destroy previously created streams
-    for (ordinal_type i = 0; i < _nstreams; ++i) {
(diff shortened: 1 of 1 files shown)


iyamazaki:
I will double-check, but this looks good to me. Are you filing an issue/question to Kokkos?

kyungjoo-kim:
No. you can post issues for further discussion about Kokkos execution space behavior consistency.

jjellio:
I won't be able to pull and test this on EAS3 today - but it's probably a viable work around. Just need to setup and issue with Kokkos and let them figure out the semantics they want (so it's consistent)

kyungjoo-kim:
The code is tested on caraway03 with ROCM 5.2.

kyungjoo-kim:
BTW, the order to delete streams and kokkos execution space instances should not matter. When a stream is destroyed, it should sync to finalize all submitted jobs. If kokkos execution space instance see that the stream is not valid, then they should not take any actions on the stream and necessary synchronization should happen when the stream is actually destroyed instead of when kokkos execution space instance is destroyed.

PR #10880, opened 2022-08-15
Panzer MiniEM: Adjust rebalancing targets for RefMaxwell sub-problems

cgcgcg:
@trilinos/panzer 

## Motivation
The change should fix a regression picked up by the performance tests.


cgcgcg:
@csiefer2 FYI

rppawlo (approved):

PR #10882, opened 2022-08-15
[WIP] Matrix-free MiniEM

cgcgcg:
@trilinos/panzer @trilinos/muelu @trilinos/thyra 

## Motivation
The goal is to add a matrix-free code path to MiniEM for higher order discretizations. So far, only the transfers between multigrid levels and the curl operator are actually matrix-free.


GrahamBenHarper:
Thanks @cgcgcg! This is perfect for what I need.

Issue #10839, opened 2022-08-08
Compadre:  GMLS Python Script, can't find file


DaveBeCoding:
## Bug Report
@trilinos/compadre 

### Description
GMLS testing executes test python script, and fails with:
```

Running: "/path/to/python/2.7.9/bin/python" "/path/to/build/packages/compadre/examples/GMLS_Staggered_Manifold.py" "--porder=3" "--grids=4" "--in-trilinos=True"

--------------------------------------------------------------------------------

['/path/to/build/packages/compadre/examples/Compadre_GMLS_Staggered_Manifold_Test.exe', '--p', '3', '--nt', '100', '--d', '3', '--ns', '2000', '--solver', 'QR', '--problem', 'MANIFOLD', '--kokkos-threads=4']
Traceback (most recent call last):
  File "/path/to/build/packages/compadre/examples/GMLS_Staggered_Manifold.py", line 45, in <module>
    output = subprocess.check_output(output_commands, stderr=devnull)
  File "/path/to/python/2.7.9/lib/python2.7/subprocess.py", line 566, in check_output
    process = Popen(stdout=PIPE, *popenargs, **kwargs)
  File "/path/to/python/2.7.9/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/path/to/python/2.7.9/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory



```

### Steps to Reproduce
1. See errors,  https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&display=project&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=PR-__UNKNOWN__-test-rhel7_sems-clang-10.0.0-openmpi-1.10.1-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables-532&field2=buildstamp&compare2=61&value2=20220727-1638-Experimental

2. Follow instructions in https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors  and use 
https://github.com/srbdev/Trilinos/commits/framework-pr-clang10-c%2B%2B17-test for step 5 "Reproducing PR testing errors"

kuberry:
[...]

PR #10891, opened 2022-08-16
commonTools: get-changes-trilinos-packages.sh

e10harvey:
- Update macro to use FORCE set

@trilinos/framework

## Motivation

While trying to work through #10839  @srbdev found several errors in https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors and updated that documentation. Another error that we found was during step 10 where `Trilinos_ENABLE_ALL_PACKAGES` was already set in the cmake cache, and so, is never overwritten by the value specified in `packageEnables.cmake`. After checking `packages/framework/ini-files`, it is clear that something else in the trilinos cmake is caching `Trilinos_ENABLE_ALL_PACKAGES`. This PR addresses this silent error during step 10 by using a "forceful" cmake cache  variable set.

## Stakeholder Feedback

@bartlettroscoe: Please let me know if there is a better way to address this reproducibility problem. 

## Testing

Tested this change locally. The PR AutoTester will also use this script here.


bartlettroscoe:
@e10harvey, first, I would like to suggest to not put put **any** package enable-related set() statements in the initial configure by Gen Config or the fragment file produced by Gen Config.  Put all package enable-related settings including `Trilinos_ENABLE_ALL_PACKAGES`, `Trilinos_ENABLE_ALL_FORWARD_DEP_PACKAGES` and `Trilinos_ENABLE_TESTS` in the file `packageEnablers.cmake`.    That solves the problem.

Otherwise, stop using this two-phase cmake configure process.  Always have gen-config.sh produce an CMake fragment file `GenConfigSettings.cmake` when loading the env and then configure **one time** with:

```
cmake -C packageEnables.cmake -C GenConfigSettings.cmake [option options] <tril-src-dir>
```

That is more efficient, more straightforward, and more like a developer would actually use CMake if given a choice.

In summary, the trick to dealing with overriding CMake cache var setting is to put your set() statements first.  That is, the first call:

```
set(<someCacheVar> <someValue> CACHE <type> "<doc>")
```

for a cache var name `<someCacheVar>` determines the default value for the cache var.  All later set() cache var statements are ignored.  This is how CMake is designed to work/

So, if you want the cache var settings in `<fragmentFile2>.cmake` to override those in `<fragmentFile1>.cmake` you list them first as:

```
cmake -C <fragmentFile2>.cmake -C <fragmentFile1>.cmake [option options] <tril-src-dir>
```

You want to never use `FORCE`, otherwise, any options the user passes in with `-D<cacheVar>=<userVal>` will get ignored and you never want that to be the case.  The number-one purpose of cache vars is to allow the user to override them on the command-line (or by providing their fragment file first on the command-line with `-C <userFragmentFile>.cmake`).

But if you are not convinced, then I propose to add a option `--use-set-force` to this Python tool to optionally add `FORCE` and then you can use that in the PR tester.  But for other usages as described above, you would not do this.

e10harvey:
> @e10harvey, first, I would like to suggest to not put put **any** package enable-related set() statements in the initial configure by Gen Config or the fragment file produced by Gen Config. Put all package enable-related settings including `Trilinos_ENABLE_ALL_PACKAGES`, `Trilinos_ENABLE_ALL_FORWARD_DEP_PACKAGES` and `Trilinos_ENABLE_TESTS` in the file `packageEnablers.cmake`. That solves the problem.

@bartlettroscoe: This does not meet the user requirements for gen config - we must allow package enable statements. Trilinos_ENABLE_ALL_PACKAGES is not set by gen config but instead, something else in the trilinos cmake infrastructure. 

> Otherwise, stop using this two-phase cmake configure process.

I agree. We will update the documentation and close this pull request.



> You want to never use `FORCE`, otherwise, any options the user passes in with `-D<cacheVar>=<userVal>` will get ignored and you never want that to be the case. The number-one purpose of cache vars is to allow the user to override them on the command-line (or by providing their fragment file first on the command-line with `-C <userFragmentFile>.cmake`).

+1

Thank you for your feedback.

bartlettroscoe:
>  This does not meet the user requirements for gen config - we must allow package enable statements. Trilinos_ENABLE_ALL_PACKAGES is not set by gen config but instead, something else in the trilinos cmake infrastructure.

What user asked for this?  I bet if you go back to that user and explained the downsides of setting package enable-related vars in the CMake fragment file, they would be fine in taking them out.  And if you can't identify that user, then that is not a real requirement.

e10harvey:
> > This does not meet the user requirements for gen config - we must allow package enable statements. Trilinos_ENABLE_ALL_PACKAGES is not set by gen config but instead, something else in the trilinos cmake infrastructure.
> 
> What user asked for this? I bet if you go back to that user and explained the downsides of setting package enable-related vars in the main GenConfig CMake fragment file, then they would be fine in taking them out. And if you can't identify that user, then that is not a real requirement.

Package enable vars are a Trilinos-specific problem related to how CMake cache variables work while GenConfig is a project-agnostic tool.

PR #10900, opened 2022-08-18
Ifpack2: add AVG Combine Mode option to Schwarz DD

rstumin:
Issue:  #10901
Tests: Added to MueLu
Stakeholder feedback: n/a

Auto-PR for SHA be3d516

Issue #10823, opened 2022-08-02
Framework: Some PR build errors showing up as strange 'cmake --build . --config Release -- -j29 -k 0' errors


bartlettroscoe:
## Bug Report

@trilinos/framework 

### Next Action Status

This is due to a defect in CTest introduced in CMake 3.18.  The fix for this is in [CMake 3.24.3](https://github.com/Kitware/CMake/releases/tag/v3.24.3) (released 2022-11-01) .  (See [SNL Kitware #209](https://gitlab.kitware.com/snl/project-1/-/issues/209)).  Next: Install CMake 3.24.3 everywhere and use with Trilinos PR builds ...

### Internal issues

* [TRILINOSHD-136](https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-136)
* [TRILINOSHD-228](https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-228)
* [SEMSHELPD-3419](https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/4/SEMSHELPD-3419)

### Description

As shown in [this query](https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&begin=2022-07-24&end=2022-08-02&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=rhel7_sems-cuda-11.4.2-sems-gnu-10.1.0-sems-openmpi-4.0.5_release_static_Volta70_no-asan_complex_no-fpic_mpi_pt_no-rdc_no-uvm_deprecated-on_no-package-enables&field2=builderrors&compare2=43&value2=0) showing:

![image](https://user-images.githubusercontent.com/1959736/182380974-30854095-667f-484f-9d6e-177d5e8cc246.png)

the new Trilinos Framework GenConfig build `rhel7_sems-cuda-11.4.2-sems-gnu-10.1.0-sems-openmpi-4.0.5_release_static_Volta70_no-asan_complex_no-fpic_mpi_pt_no-rdc_no-uvm_deprecated-on_no-package-enables` is failing with a build error reported in the `Zoltan2Sphynx` package showing:

```
"/projects/sems/install/rhel7-x86_64/sems/v2/utility/cmake/3.21.1/gcc/7.3.0/mxfpluq/bin/cmake" "--build" "." "--config" "Release" "--" "-j29" "-k" "0"
```

returning error code `1`.

[...]

PR #10606, opened 2022-06-08
MueLu: add static sparse approximate inverse algorithm

maxfirmbach:
@trilinos/muelu

## Motivation

Adding a static sparse approximate inverse algorithm (SPAI). Based on a static input sparsity
pattern, the inverse of a matrix A is calculated on the given pattern.

- enhance the inverse approximation factory by including the static SPAI algorithm
- currently uses the CrsGraph of given input matrix A as default choice, different pattern selection
  methods will be included at a later point

Especially for ill-conditioned and non-diagonal dominant systems the inverse approximation methods currently implemented are not sufficient for certain applications as they only use mostly diagonal information for the inverse approximation. With the SPAI algorithm more information of the inverse can be received which (hopefully) results in a better Schur complement approximation.

## Related Issues

* Related to #10437, #10484.

## Stakeholder Feedback

AMG for block matrices with schur complement

## Testing

- added unit testing for the static SPAI algorithm inside the inverse appoximation factory, checking for different matrix types and sizes

Joined work with @mayrmt

Issue #209, opened 2016-03-11
IFPACK2 Compile Failure - GCC 4.8.4 on Hansen Test System


nmhamster:
Reproducer:

`module load devpack/openmpi/1.10.0/gcc/4.8.4/cuda/7.5.18` (Hansen)

Configure Trilinos, Run `make` and output:

```
[ 90%] Building CXX object packages/ifpack2/src/CMakeFiles/ifpack2.dir/Ifpack2_AdditiveSchwarz_Serial.cpp.o
In file included from /home/sdhammo/git/trilinos-github-repo/build-devpack-gcc-4.8.4/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_Serial.cpp:50:0:
/home/sdhammo/git/trilinos-github-repo/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp: In instantiation of ‘void Ifpack2::AdditiveSchwarz<MatrixType, LocalInverseType>::setup() [with MatrixType = Tpetra::RowMatrix<double, int, long int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >; LocalInverseType = Ifpack2::Preconditioner<double, int, long int, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial> >]’:
/home/sdhammo/git/trilinos-github-repo/build-devpack-gcc-4.8.4/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_Serial.cpp:61:1:   required from here
/home/sdhammo/git/trilinos-github-repo/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp:1531:75: error: conversion from ‘Teuchos::ArrayRCP<long int>’ to non-scalar type ‘Teuchos::ArrayRCP<int>’ requested
       ArrayRCP<local_ordinal_type> perm = sol.getPermutationRCPConst (true);
                                                                           ^
/home/sdhammo/git/trilinos-github-repo/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp:1532:74: error: conversion from ‘Teuchos::ArrayRCP<long int>’ to non-scalar type ‘Teuchos::ArrayRCP<int>’ requested
       ArrayRCP<local_ordinal_type> revperm = sol.getPermutationRCPConst ();
                                                                          ^
[...]

PR #10902, opened 2022-08-19
MueLu: Add ComputeNodesInAggregate for Aggregates_kokkos

GrahamBenHarper:
@trilinos/muelu 

## Motivation
This addresses an inconsistency between `Aggregates` and `Aggregates_kokkos`. As an added bonus, this enables matrix-free restriction grid transfers. With how non-parallel the algorithm is, I think I can see why nobody did this before 🙂 

I have a couple bits of tidying to do before the final test, but I figured I'd open this now anyway. Mainly I'm really trying to figure out if the main couple loops can be turned into some more parallel-friendly algorithm. Otherwise I'll just copy everything to host and do the serial thing.

### DIFF
--- a/packages/muelu/src/Graph/Containers/MueLu_Aggregates_def.hpp
+++ b/packages/muelu/src/Graph/Containers/MueLu_Aggregates_def.hpp
@@ -191,12 +191,12 @@ namespace MueLu {
     for(LO i=0; i<numNodes; i++) {
       LO aggregate = vertex2AggId[i];
       if(aggregate !=INVALID) {
-	aggNodes[aggCurr[aggregate]] = i;
-	aggCurr[aggregate]++;
+        aggNodes[aggCurr[aggregate]] = i;
+        aggCurr[aggregate]++;
       }
       else {
-	unaggregated[currNumUnaggregated] = i;
-	currNumUnaggregated++;
+        unaggregated[currNumUnaggregated] = i;
+        currNumUnaggregated++;
       }
     }
     unaggregated.resize(currNumUnaggregated);
--- a/packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
+++ b/packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
@@ -180,6 +180,58 @@ namespace MueLu {
 
     return graph_;
   }
+  
+  template <class LocalOrdinal, class GlobalOrdinal, class DeviceType>
+  void 
+  Aggregates_kokkos<LocalOrdinal, GlobalOrdinal, Kokkos::Compat::KokkosDeviceWrapperNode<DeviceType> >::ComputeNodesInAggregate(LO_view & aggPtr, LO_view & aggNodes, LO_view & unaggregated) const {
+    LO numAggs  = GetNumAggregates();
+    LO numNodes = vertex2AggId_->getLocalLength();
+    auto vertex2AggId = vertex2AggId_->getDeviceLocalView(Xpetra::Access::ReadOnly);
+    typename aggregates_sizes_type::const_type aggSizes = ComputeAggregateSizes(true);
+    LO INVALID = Teuchos::OrdinalTraits<LO>::invalid();
+
+    aggPtr = LO_view("aggPtr",numAggs+1);
+    aggNodes = LO_view("aggNodes",numNodes);
+    LO_view aggCurr("agg curr",numAggs+1);
+
+    // Construct the "rowptr" and the counter
+    Kokkos::parallel_scan("MueLu:Aggregates:ComputeNodesInAggregate:scan", range_type(0,numAggs+1),
+      KOKKOS_LAMBDA(const LO aggIdx, LO& aggOffset, bool final_pass) {
+        LO count = 0;
+        if(aggIdx < numAggs)
+          count = aggSizes(aggIdx);
+        if(final_pass) {
+          aggPtr(aggIdx) = aggOffset;
+          aggCurr(aggIdx) = aggOffset;
+          if(aggIdx==numAggs)
+            aggCurr(numAggs) = 0; // use this for counting unaggregated nodes
+        }
+        aggOffset += count;
+      });
+
+    // Preallocate unaggregated to the correct size
+    LO numUnaggregated = 0;
+    Kokkos::parallel_reduce("MueLu:Aggregates:ComputeNodesInAggregate:unaggregatedSize", range_type(0,numNodes),
+      KOKKOS_LAMBDA(const LO nodeIdx, LO & count) { 
+        if(vertex2AggId(nodeIdx,0)==INVALID) 
+          count++;
+      }, numUnaggregated);
+    unaggregated = LO_view("unaggregated",numUnaggregated);
+
+    // Stick the nodes in each aggregate's spot
+    Kokkos::parallel_for("MueLu:Aggregates:ComputeNodesInAggregate:for", range_type(0,numNodes),
+      KOKKOS_LAMBDA(const LO nodeIdx) {
+        LO aggIdx = vertex2AggId(nodeIdx,0);
+        if(aggIdx != INVALID) {
+          // atomic postincrement aggCurr(aggIdx) each time
+          aggNodes(Kokkos::atomic_fetch_add(&aggCurr(aggIdx),1)) = nodeIdx;
+        } else {
+          // same, but using last entry of aggCurr for unaggregated nodes
+          unaggregated(Kokkos::atomic_fetch_add(&aggCurr(numAggs),1)) = nodeIdx;
+        }
+      });
+    
+  }
 
   template <class LocalOrdinal, class GlobalOrdinal, class DeviceType>
   std::string Aggregates_kokkos<LocalOrdinal, GlobalOrdinal, Kokkos::Compat::KokkosDeviceWrapperNode<DeviceType> >::description() const {
--- a/packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_decl.hpp
+++ b/packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_decl.hpp
@@ -116,6 +116,7 @@ namespace MueLu {
     using node_type           = Kokkos::Compat::KokkosDeviceWrapperNode<DeviceType>;
     using device_type         = DeviceType;
     using range_type          = Kokkos::RangePolicy<local_ordinal_type, execution_space>;
+    using LO_view             = Kokkos::View<local_ordinal_type*, device_type>;
 
     using aggregates_sizes_type = Kokkos::View<LocalOrdinal*, device_type>;
 
@@ -258,6 +259,12 @@ namespace MueLu {
     typename aggregates_sizes_type::const_type ComputeAggregateSizes(bool forceRecompute = false) const;
 
     local_graph_type GetGraph() const;
+
+    /*! @brief Generates a compressed list of nodes in each aggregate, where
+      the entries in aggNodes[aggPtr[i]] up to aggNodes[aggPtr[i+1]-1] contain the nodes in aggregate i.
+      unaggregated contains the list of nodes which are, for whatever reason, not aggregated (e.g. Dirichlet)
+     */
+    void ComputeNodesInAggregate(LO_view & aggPtr, LO_view & aggNodes, LO_view & unaggregated) const;
 
     //! @name Overridden from Teuchos::Describable
     //@{
--- a/packages/muelu/test/unit_tests_kokkos/Aggregates_kokkos.cpp
+++ b/packages/muelu/test/unit_tests_kokkos/Aggregates_kokkos.cpp
@@ -614,15 +614,15 @@ namespace MueLuTests {
     // Check ComputeNodesInAggregate
     typename Aggregates_kokkos::LO_view aggPtr, aggNodes, unaggregated;
     aggregates->ComputeNodesInAggregate(aggPtr, aggNodes, unaggregated);
-    TEST_EQUALITY(aggPtr.extent(0), numAggs+1);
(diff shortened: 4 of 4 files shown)


jhux2 (commented):
Lgtm

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
```diff
@@ -180,6 +180,60 @@ namespace MueLu {
...
+    LO numAggs  = GetNumAggregates();
+    LO numNodes = vertex2AggId_->getLocalLength();
+    auto vertex2AggId = vertex2AggId_->getDeviceLocalView(Xpetra::Access::ReadOnly);
+    typename aggregates_sizes_type::const_type aggSizes = ComputeAggregateSizes(true);
+    LO INVALID = Teuchos::OrdinalTraits<LO>::invalid();
+
+    LO currNumUnaggregated=0;
+
+    // Construct the "rowptr" and the counter
+    // This is completely egregious to perform on host, but doesn't seem parallelizeable; is there a way to improve this?
```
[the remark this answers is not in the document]
GrahamBenHarper: Ah shoot Github just ate my comment when I clicked away.

I think the hard part boils down to the fact that a CRS matrix has a `rowPtr`, but I don't have `aggPtr`. Instead, I have `aggSizes`, so if I parallelize on device over aggregates (which is probably the most logical choice), each aggregate needs to compute its own `aggPtr[i]=aggSizes[0]+aggSizes[1]+...+aggSizes[i]`. I feel like that would introduce some terrible thread balancing without doing some sort of smart divide and conquer strategy, but I don't know how I'd do it. Nonetheless, it could very well be faster than running on host in the first place, so it might be worthwhile.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_def.hpp
```diff
@@ -191,12 +191,12 @@ namespace MueLu {
     for(LO i=0; i<numNodes; i++) {
       LO aggregate = vertex2AggId[i];
       if(aggregate !=INVALID) {
-	aggNodes[aggCurr[aggregate]] = i;
-	aggCurr[aggregate]++;
+        aggNodes[aggCurr[aggregate]] = i;
+        aggCurr[aggregate]++;
```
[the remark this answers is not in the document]
GrahamBenHarper: I've been guilty of introducing tabs, so I feel like it's only appropriate :stuck_out_tongue:

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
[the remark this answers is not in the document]
brian-kelley: @GrahamBenHarper That's a scan pattern and it actually can be done in parallel. This will transform aggSizes to aggPtr (assuming aggSizes length is ``numAggs``, and aggPtr length is ``numAggs + 1``).

```
    Kokkos::parallel_scan(Kokkos::RangePolicy<Device>(0, numAggs + 1),
        KOKKOS_LAMBDA(local_ordinal_type i, local_ordinal_type& lsum, bool finalPass)
        {
          local_ordinal_type count = 0;
          if(i < numAggs)
            count = aggSizes(i);
          if(finalPass)
            aggPtr(i) = lsum;
          lsum += count;
        });
```
This version doesn't need aggPtr to be initialized beforehand. You could also do it in-place if you have the sizes start out in ``aggPtr[0...numAggs-1]``, with ``aggPtr[numAggs] == 0``. Just replace aggSizes above with aggPtr in that case.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
```diff
@@ -180,6 +180,60 @@ namespace MueLu {
...
+
+    Kokkos::resize(aggPtr_h,numAggs+1);
+    Kokkos::resize(aggNodes_h,numNodes);
+    Kokkos::resize(unaggregated_h,numNodes);
+    LO_view aggCurr("agg curr",numAggs+1);
+    typename LO_view::HostMirror aggCurr_h = Kokkos::create_mirror_view(aggCurr); // FIXME: not needed on device... Kokkos::View<LO*,host_type>?
+    Kokkos::deep_copy(aggCurr_h,aggCurr);
+
+    aggPtr_h(0) = 0; 
+    for(LO i=0; i<numAggs; i++) {
```
lucbv, line 213: In general, using `i` as index is a good way to get lost when you'll have more than on index in your code. How about `aggIdx` instead, it's pretty clear what that refers to.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
```diff
@@ -180,6 +180,60 @@ namespace MueLu {
...
+      aggPtr_h(i+1) = aggSizes_h(i) + aggPtr_h(i);
```
lucbv, line 214: So you are basically trying to build the equivalent of a CrsMatrix `row_ptr`? In that case I suggest reading up [this Tpetra lesson](https://docs.trilinos.org/dev/packages/tpetra/doc/html/Tpetra_Lesson07.html), especially look for parallel scan.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
```diff
@@ -180,6 +180,60 @@ namespace MueLu {
...
+    // This is completely egregious to perform on host, but doesn't seem parallelizeable; is there a way to improve this?
+    typename LO_view::HostMirror aggPtr_h = Kokkos::create_mirror_view(aggPtr);
+    typename LO_view::HostMirror aggNodes_h = Kokkos::create_mirror_view(aggNodes);
+    typename LO_view::HostMirror unaggregated_h = Kokkos::create_mirror_view(unaggregated);
+    typename Kokkos::View<LO**,device_type>::HostMirror vertex2AggId_h = Kokkos::create_mirror_view(vertex2AggId); // TODO: this is a dualview, so we can't do it this way
+    typename aggregates_sizes_type::HostMirror aggSizes_h = Kokkos::create_mirror_view(aggSizes);
+    Kokkos::deep_copy(vertex2AggId_h, vertex2AggId); // TODO: same here
+    Kokkos::deep_copy(aggSizes_h, aggSizes);
+
+    Kokkos::resize(aggPtr_h,numAggs+1);
```
lucbv, line 205: Are the view initialized somewhere else but with the wrong allocation size? I would rather we avoid using `Kokkos::resize` at all cost.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
```diff
@@ -180,6 +180,60 @@ namespace MueLu {
...
+    Kokkos::deep_copy(aggCurr_h,aggCurr);
+
+    aggPtr_h(0) = 0; 
+    for(LO i=0; i<numAggs; i++) {
+      aggPtr_h(i+1) = aggSizes_h(i) + aggPtr_h(i);
+      aggCurr_h(i) = aggPtr_h(i);
+    }
+
+    // Stick the nodes in each aggregate's spot
+    for(LO i=0; i<numNodes; i++) {
```
lucbv, line 219: Same as above, can you rename `i` as `nodeIdx`?
Assuming you us a `parallel_scan` above, this should become a `parallel_for`

lucbv (changes requested):
If we are not going to run things on device I would recommend only writing a wrapper that calls the non-kokkos version, that would require less maintenance and lead to less bugs down the road. But I think this is can be parallelized.

GrahamBenHarper:
@lucbv I've never used `parallel_scan` before. I'll check it out! I just pushed a first pass at attempting to do this on device properly, but I'll look into your feedback now. Thanks for looking at this!

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
GrahamBenHarper: Done!

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
GrahamBenHarper: I see now! I should have known there was something that computed exactly what I needed. Thanks!

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
GrahamBenHarper: This method modifies the input views and returns them to the user. I'm not going to trust the user to size things appropriately :slightly_smiling_face:

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
GrahamBenHarper: Done!

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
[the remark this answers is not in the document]
GrahamBenHarper: thanks so much @brian-kelley!

GrahamBenHarper:
There we go. I'm much happier now. I must have slept through the Kokkos tutorial parts where they covered `parallel_scan`, because this is the greatest thing ever :sweat_smile:

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
GrahamBenHarper: We just chatted about this. It's better to just throw away the user input than resize it. Done!

GrahamBenHarper:
Okay this is for real the last pass. This should be pretty close to the best performance we can get on device.

jhux2:
@GrahamBenHarper I noticed the dashboard failure is under `Zoltan2Sphyx`, but the error is clearly when building MueLu's unit tests.  I saw this same thing in #10900.  @bartlettroscoe Any idea why this might be?

GrahamBenHarper:
I believe I've mentioned this before somewhere. Every time I see a `Zoltan2Sphynx` error I just click past it because there's always a different error in the log. The parallel build doesn't die until it reaches `Zoltan2Sphynx`, however, which is why I think the build reports the failure in that location erroneously.

GrahamBenHarper:
Yep, you can definitely see this same false reporting on #10606. Here's the relevant [cdash entry](https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-10606-test&field2=buildstarttime&compare2=84&value2=NOW) which incorrectly reports the `Zoltan2Sphynx` error. It was happening before that PR as well, but I assumed people were clicking past it to see the actual build log.

bartlettroscoe:
> @GrahamBenHarper I noticed the dashboard failure is under `Zoltan2Sphyx`, but the error is clearly when building MueLu's unit tests. I saw this same thing in #10900. @bartlettroscoe Any idea why this might be?

@jhux2, this behavior was noted in #10823.  I am working with @zackgalbreath at Kitware to try to figure this out  the best we can (see [SNL Kitware #209](https://gitlab.kitware.com/snl/project-1/-/issues/209)).  There are several challenges.  First, we can't see the XML files getting generated on the Jenkins clients (see [TRILINOSHD-188)](https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-188)).  Also, it is very hard to get to the  XML files that are submitted to trilinos-cdash.sandia.gov.  (Kitware did not set up that CDash server and has limited access to it.)  I did not write the code run by the ctest -S script and I don't even know where that code lives.  All of this is owned by the Trilinos Framework team.  They own the Jenkins jobs, they own the CDash server, they own the code that is run on the Jenkins server.  I have no way to reproduce anything.  It is an extremely difficult situation to deal with.

csiefer2 (approved):

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
```diff
@@ -180,6 +180,58 @@ namespace MueLu {
...
+  void 
+  Aggregates_kokkos<LocalOrdinal, GlobalOrdinal, Kokkos::Compat::KokkosDeviceWrapperNode<DeviceType> >::ComputeNodesInAggregate(LO_view & aggPtr, LO_view & aggNodes, LO_view & unaggregated) const {
+    LO numAggs  = GetNumAggregates();
+    LO numNodes = vertex2AggId_->getLocalLength();
+    auto vertex2AggId = vertex2AggId_->getDeviceLocalView(Xpetra::Access::ReadOnly);
+    typename aggregates_sizes_type::const_type aggSizes = ComputeAggregateSizes(true);
+    LO INVALID = Teuchos::OrdinalTraits<LO>::invalid();
+
+    aggPtr = LO_view("aggPtr",numAggs+1);
+    aggNodes = LO_view("aggNodes",numNodes);
```
cgcgcg, line 194: Is `numNodes` the number of all aggregated nodes, or the number of all nodes?

GrahamBenHarper: The number of all nodes. This is the same way it's handled in `MueLu_Aggregates` from what I recall, meaning there's probably a small amount of padding at the end in the case where there are unaggregated nodes.

lucbv (approved):
Looks good now

PR #10904, opened 2022-08-19
Amesos2-IR & ShyLU-Basker fixes

iyamazaki:
@trilinos/amesos2, @trilinos/shylu 

## Motivation
- Fixes the iterative refinements in Amesos2
- Improve performance of ShyLU-Basker with MWM (by applying AMD ordering on block D)

## Testing
Tested using test matrices from Xyce

### DIFF
--- a/packages/shylu/shylu_node/basker/src/shylubasker_def.hpp
+++ b/packages/shylu/shylu_node/basker/src/shylubasker_def.hpp
@@ -925,10 +925,11 @@ namespace BaskerNS
           }
         }
 
-        #ifdef AMD_ON_D
+        #define SHYLU_BASKER_AMD_ON_D
+        #ifdef SHYLU_BASKER_AMD_ON_D
         // --------------------------------------------
         // reset the small D blocks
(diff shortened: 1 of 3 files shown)


packages/shylu/shylu_node/basker/src/shylubasker_def.hpp
```diff
@@ -1100,7 +1103,7 @@ namespace BaskerNS
         numeric_col_iperm_array(i) = i;
       }
 
-      #ifdef AMD_ON_D
+      #ifdef SHYLU_BASKER_AMD_ON_D
       if (btf_top_tabs_offset > 0) {
```
srajama1, line 1107: I thought this would be a runtime option chosen with a parameter. This is ok for now, but we should change it to runtime option so we have the path tested with two different tests (one with AMD and one without). With compile time options, this code is not going to be tested in our nightly tests.

srajama1 (approved):
Thanks @iyamazaki ! One comment below. I will approve it so this can be merged, but it would be ncie to change it to runtime option soon.

PR #10897, opened 2022-08-18
Improve TrilinosInstallTests to make more clear (#10896)

bartlettroscoe:
### Description

Based on the confusion by a Trilinos developer that creatd issue #10896, I have improved these installation tests (see the commit log message for details).

### How was this tested?

In addition to this getting tested in PR testing, I ran two cases locally:

**1) Build passes and all of the TrilinosInstallTests tests run and pass**:
  * => See results on CDash [here](???)

**2) Build fails (due to injected type in Tpetra_Map.cpp) and only the TrilinosInstallTests_doInstall test runs and fails and the others are skipped**:
  * => See results on CDash [here](???)

Issue #10896, opened 2022-08-17
TrilinosInstallTests Failing in PR


csiefer2:
These guys:

- TrilinosInstallTests_doInstall
- TrilinosInstallTests_find_package_Trilinos
- TrilinosInstallTests_simpleBuildAgainstTrilinos

See example:
https://trilinos-cdash.sandia.gov/viewTest.php?onlyfailed&buildid=392610

bartlettroscoe:
@csiefer2, I will try to reproduce and update the doInstall test to show the tail of the install command to see what is failing.  (Can't show the entire STDOUT since that is over 1 MB in size.)

bartlettroscoe:
> Awesome, thanks. That same build has Zoltan2Sphynx dying like this:
> 
> ninja: error: '/usr/lib64/liblapack.so', needed by 'packages/teuchos/numerics/src/libteuchosnumerics.so.13.5', missing and no known rule to make it

Well that is why then the install is failing.  If the build does not work, the install will fail as well  (because the install target is dependent on all of the output targets).  But it would still be nice to update the doInstall test to show at least some of those errors (and to warn users that if the build fails, these tests will fail as well).

> So I'm wondering if something system-related happened like the disk filling up again....

No.  Do a search of the GitHub issues (e.g. "ninja: error: '/usr/lib64/liblapack.so'") and you will see why.  I just reported this this morning.  (If you are no sure how to search github issues, I can provide some tips)

csiefer2:
@bartlettroscoe The install is failing on another build as well which doesn't have any build failures

https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-10827-test&field2=buildstarttime&compare2=84&value2=NOW

csiefer2:
Although it is on one of the cursed machines...

bartlettroscoe:
> @bartlettroscoe The install is failing on another build as well which doesn't have any build failures

@csiefer2, no, those clang-10.0.0  builds on the 'ascic0xy' machines also have the same build failures (because system BLAS and LAPACK are not being installed), they are just not being reported to CDash.  See explanation in https://github.com/trilinos/Trilinos/issues/10893#issuecomment-1218138724.

[...]

Issue #10858, opened 2022-08-10
EPIC: Issues with Trilinos PR testing 2022-08


bartlettroscoe:
There are a number of issues impacting Trilinos PR testing currently that are blocking many active PRs from passing the PR tester:

* #6861
* #10842
* #10836
* #10847 ... **Tests have been disabled for now**
* #10841 
* #10857 
* #10875
* #10885
* #10893
* #10898
* #10917
* #10931... **Pliris is disabled in ATS2 CUDA PR builds with merge of PR [#10930](https://github.com/trilinos/Trilinos/pull/10930)**

Also of interest:

* #10823 
* #10867
* #10840
* #10849 
* #10899

csiefer2:
https://github.com/trilinos/Trilinos/issues/10896

bartlettroscoe:
> #10896

As explained in https://github.com/trilinos/Trilinos/issues/10896#issuecomment-1218210333, there are not causes of PR failures, they are victims of a build failure that was also reported (or not as per https://github.com/trilinos/Trilinos/issues/10893#issuecomment-1218138724 :-( ).

bartlettroscoe:
Even with all of the fixes in the linked issues [above](), we still have a large log jam in PR testing.  For example, right now [this issues query](https://github.com/trilinos/Trilinos/pulls?q=is%3Aopen+is%3Apr+label%3A%22AT%3A+RETEST%22) shows there are 8 PRs that have the `AT: RETEST` label on them.  It looks like something is wrong with the PR tester is that only one set of PR build is running at one time.  The outer driver for the second set of PRs looks to be broken (see  https://do.sandia.gov/trilinos-ci/job/Trilinos_autotester_driver_inst_1/).  I will post a Trilinos HelpDesk issue.

bartlettroscoe:
FYI: I posted [TRILINOSHD-18](https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-187) to see if they can turn off auto-retest of failing PR builds.  That is often a waste of computing resources as the developer has not even had a chance to look over the failed PR build results to see if rerunning the PR builds is likely to fix anything.

bartlettroscoe:
As noted [above](https://github.com/trilinos/Trilinos/issues/10858#issuecomment-1220937087), only one of two Jenkins `Trilinos_autotester_driver_inst` instances is running and there seems to be logic built into assume that the other instance is running PR builds. For example:

[...]

PR #10911, opened 2022-08-21
Improve TrilinosInstallTests to make more clear (#10896)

bartlettroscoe:
Duplicate of #10897 but with an even-numbered PR ID so it will get run by the PR tester (see https://github.com/trilinos/Trilinos/issues/10858#issuecomment-1221454400) :-)


bartlettroscoe:
Shoot, this turned out to be an odd-numbered PR so it will not get tested.  Closing and will open a new PR.

PR #10926, opened 2022-08-24
SpMV Communication/Computation Overlap

cwpearson:
# This PR is a work in progress. Do not merge

Introduces optional communication-computation overlap for SpMV (CrsMatrix `apply`).
* Uses the four-array CRS, so Kokkos Kernels is not usable when enabled
* Bring over a modified Kokkos Kernels SpMV into LocalCrsMatrix
* Hidden behind the `TPETRA_OVERLAP_SPMV` environment variable
* Add `Tpetra_Spaces.hpp/.cpp` to manage multiple execution spaces inside Tpetra

@trilinos/tpetra

## Motivation

Improve SpMV performance, which is fundamental to many applications.

## Prerequisites
- [x] https://github.com/trilinos/Trilinos/pull/11703
- [x] https://github.com/trilinos/Trilinos/pull/11714
- [x] https://github.com/trilinos/Trilinos/pull/11734
- [x] https://github.com/trilinos/Trilinos/pull/11735
- [ ] https://github.com/trilinos/Trilinos/pull/11736
- [ ] export with execution space instance
- [ ] import with execution space instance

## Overlap Design

**Key:** to interpret diagrams below
```
import: import
p     : pack
u     : unpack
L(x)  : launch kernel x
f(x)  : fence exec space x
x->y  : y waits for x
on    : on-rank SpMV
off   : off-rank SpMV
spmv  : combined SpMV
```

In the diagrams below:
* time proceeds left-to-right
* the left column annotates where activity occurs, and for GPU streams gives a number for each
* The MPI implementation manages one or more of its own streams, which we cannot touch/modify without significant implementation-specific hacky stuff (thus, `(?)`).

**Current**
```
(1) Kokkos default stream |   |p|             |u| |------------------spmv------------------|
                         -|
(?) MPI internal stream   |         |---?--|
                         -|
CPU                       |L(p)|f(1)|import|f|L(u)|L(spmv)| next op
```

**Proposed**
```
(3) lo-prio Stream        |   |------------------on--------------|     
                         -|
(2) hi-prio Stream        |                |p|         |u|
                         -|
(1) Kokkos default stream |                                          |-off-|
                         -|
(?) MPI internal stream   |                    |---?--|
                         -|
CPU                       |1->3|L(on)|1->2|L(p)|import|L(u)|3->1|2->1|L(off)|

[overlap]                                  |<------------>|
```

In the proposed implementation, additional GPU streams `(2), (3)` is used to execute operations in parallel with those of existing Tpetra code.
It's possible that `(3)` could be done in the default stream, however sometimes the default stream has blocking semantics, so it safer to manage our own  to avoid any strange interaction between the MPI internal stream and the default stream.

Packing (`p`) must occur in a higher-priority stream, so it can preempt the on-rank SpMV, and the on-rank SpMV will not therefore block the MPI operations.
This allows MPI operations to happen concurrently with the on-rank SpMV.

The machinery for adding this Tpetra-managed spaces is in `Tpetra_Space.hpp`/`.cpp`.
This stream is created lazily when it is first requested.

## Spaces Design

Overlapping communication and computation requires managing additional Kokkos execution spaces.
The interface for this is implemented in `Tpetra_Spaces.hpp`/`.cpp` and `Tpetra_Details_SpaceManager.hpp`/`.cpp` in the `Tpetra::Details::Spaces` 

The lifetime of the space instances must be scoped to that of Kokkos. Therefore, the spaces cannot have a static lifetime (i.e. be cleaned up after `main` returns) since `Kokkos::finalize()` should be called before that. Tpetra will provide `RCP<Space>` to objects that request them. When all Tpetra objects go away, the underlying space instances will also be cleaned up.

To minimize the number of created space instances, Tpetra internally tracks **weak** `RCP<Space>`.
These weak RCP do not prevent the instances from being dtor'ed when all objects go away, but can be converted to strong spaces as long as at least one object out there still retains a strong RCP.

The API is roughly the following:

* `enum class Priority {low, medium, high};`
* A `SpaceUser` class that objects can inherit from
  * providing `RCP<Space> space_instance<Space, Priority>()` to access instances of different types and priorities
* `RCP<Space> space_instance<Space, Priority>(id);`: lazily creates space with `id` and the requested `Priority` and returns it.
* `RCP<Space> space_instance<Space>(id, priority);`: same as above, but `priority` at runtime instead of statically
* `exec_space_wait<S1, S2>(S1 waitee, S2 waiter)`: causes future work in `waiter` to be blocked until current work in `waitee` is done. *May* return before work in `waitee` is done. Naively implemented as `waitee.fence()` for most cases, with a more optimized version using `cudaStreamWaitEvent` for Kokkos::Cuda

`Priority` is as abstraction over any priority implemented by the underlying space (e.g., CUDA streams). Kokkos does not currently support priorities, so CUDA stream priorities are assigned when the stream is created and before it's handed over to Kokkos. Priority is meaningless for all other Kokkos execution space types.

`Spaces` will complain if a user tries to access more than `TPETRA_SPACES_ID_WARN_LIMIT` spaces, (default `16`), since this probably indicates a logic error somewhere. This is configurable just in case the user does something we don't anticipate.

## Initial Results

**CG, 300 iterations**
| Matrix | Config | Original | Intermediate | Speedup |
|-|-|-|-|-|
| EMPIRE | Vortex, 8n 32r, ~250k rows/gpu | 0.128s | 0.111 | 1.15x |
|miniFE| Vortex, 1n 4r, ~250k rows/gpu | 0.114s | 0.104s | 1.1x |
|miniFE| Vortex, 2n 8r, ~220k rows/gpu | 0.110s | 0.097s | 1.13x |

**SpMV, throughput**
| Matrix | Config | Original | Proposed | Speedup |
|-|-|-|-|-|
| EMPIRE | Vortex, 8n 32r | 273us | 218us | 1.25x |
| Bump_2911 | Vortex, 8n 32r | 455us | 449us | 1.01x | 
| Fault_639 | Vortex, 1n 4r | 253us | 210us | 1.20x |  

**SpMV, latency**
| Matrix | Config | Original | Proposed | Speedup |
|-|-|-|-|-|
| EMPIRE | Vortex, 8n 32r | 312us | 265 us | 1.18x |
| Bump_2911 | Vortex, 2n 8r | 486us | 475us | 1.02x |
| Fault_639 | Vortex, 1n 4r | 283us | 231us | 1.22x |

## To-do

- [x] catch up with develop 
- [x] priority interface for spaces
- [x] propagate exec_space up through localApply[On/Off]Rank interface
- [x] Behavior to control warnings for large numbers of `Space`s
- [x] Fast Tpetra::Spaces syncing for `Kokkos::Cuda`
- [x] Replace Kokkos Core stream priority hack with import execution space argument
- [x] enable by default so the new code-paths can be tested in the autotester
- [ ] rework changes to `test/TpetraUtils_MatrixGenerator.hpp`
- [ ] implementations besides non-transpose
- [ ] Rebase commits
- [ ] gather CPU performance impact
- [ ] Compare against non-overlapped on GPU with cuSparse
- [ ] Unit test to make sure `localApplyOffRank() + localApplyOnRank() = localApply()`
- [ ] disable by default on CPU
- [ ] warn when enabled on CPU
- [ ] cache `isLocallyFitted` in CrsMatrix if possible?
- [ ] Use `operator==` to compare execution space instances once Kokkos is updated


cmake/TPLs/FindTPLCUBLAS.cmake
```diff
@@ -59,6 +59,8 @@ ELSE()
   find_library(CUDA_cublas_LIBRARY
     cublas
     HINTS ${CUDA_TOOLKIT_ROOT_DIR}/lib
+          ENV TRILINOS_CUBLAS_ROOT_DIR
```
cgcgcg, line 62: Why is this needed?

cmake/TPLs/FindTPLCUSPARSE.cmake
```diff
@@ -54,20 +54,25 @@
...
-      HINTS ${CUDA_TOOLKIT_ROOT_DIR}/lib
-      )
-    IF(CUDA_cusparse_LIBRARY STREQUAL "CUDA_cusparse_LIBRARY-NOTFOUND") 
-      MESSAGE(FATAL_ERROR "\nCUSPARSE: could not find cuspasre library.")
-    ENDIF()
-  ENDIF(CMAKE_VERSION VERSION_LESS "2.8.8")
+
+# Finding CUDA may fail to find the cuSparse library for CMake < 2.8.8
+# or if cuSparse is not in the CUDA toolkit directory.
+# let the user give us a hint too
```
cgcgcg, line 63: Isn't there a CMake variable like `TPL_CUBLAS_LIBRARY_DIRS` or so?

packages/tpetra/core/src/Tpetra_DistObject_def.hpp
```diff
@@ -567,7 +583,7 @@ namespace Tpetra {
 
     // mfh 18 Oct 2017: Set TPETRA_VERBOSE to true for copious debug
     // output to std::cerr on every MPI process.  This is unwise for
-    // runs with large numbers of MPI processes.
+    // runs with large numbes of MPI pocesses.
```
cgcgcg, line 586: Huh?

cwpearson: must have `:%s/r//` by accident

cmake/TPLs/FindTPLCUBLAS.cmake
cwpearson: Originally branched off master and this was resolved differently in develop since then. I'll stick with the current develop approach.

packages/tpetra/core/src/Tpetra_Space.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
```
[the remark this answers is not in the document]
cwpearson: I wonder which compiler does not support it

PR #10929, opened 2022-08-24
Intrepid2: Update TensorData.setFirstComponentExtentInDimension0 to modify extents_[0]

GrahamBenHarper:
@trilinos/intrepid2 

In the sum factorization case where the remaining number of elements is smaller than the workset size, this routine is usually called to resize the data. This sets the member variable for the extents so that any subsequent queries to `extent(0)` will return a size that is consistent with the underlying data.


CamelliaDPG (approved):
lgtm.  Thanks, @GrahamBenHarper!

PR #10951, opened 2022-08-29
Geminga: Disable Intrepid2 tests on CUDA

cgcgcg:
@trilinos/muelu 

Geminga gets somehow stuck on them, and the only other way to fix appears to be a reboot.


csiefer2 (approved):

lucbv:
It might be worth mentioning to the Intrepid2 folks, this could be happening to other users?

PR #10971, opened 2022-08-31
3.7 promotion fixes - ifpack2/fastilu

brian-kelley:
- In FastILU, update headers and namespaces for KokkosKernels utility functions (sort, transpose)
- In Ifpack2, get rid of custom volatile join() function in parallel_reduce functor. This doesn't need a join() at all since it's a simple sum-reduce.

Verified that Ifpack2 tests built and passed with FastILU enabled


ndellingwood (approved):
Thanks @brian-kelley ! Approved and merging

PR #10909, opened 2022-08-20
Add pre- and post-filtering

maxfirmbach:
@trilinos/muelu

## Motivation

Use a threshold value to filter the given sparsity to explicitly exclude small entries.
Additionally also apply this procedure to the approximate inverse to reduce the computational
cost applying the operator.

## Related Issues
* Related to #10437, #10484 and #10606

## Stakeholder Feedback

AMG for block matrices with schur complement

## Testing

- added a unit test for pre- and post-filtering of sparse graph and approximated sparse inverse matrix, checking correct values and number of nonzeros

Joined work with @mayrmt

PR #10437, opened 2022-04-14
MueLu: split schur and approximate inverse calculation

maxfirmbach:
@trilinos/muelu

## Motivation

The schur complement calculation needs an approximate inverse of one matrix block to
do the respective calculation.

This commit includes the following steps:
- split the schur and inverse calculation
- add an inverse approximation factory
- adjust the testing

I tried to keep backwards compatibility, but this is not possible for all cases ... especially
were certain settings are chosen (e.g. fixing, lumping ...).

The split of those routines will help me with future implementations (e.g. sparse inverse
approximation methods) for the schur complement.

## Stakeholder Feedback

AMG for block matrices with schur complement

## Testing

- modified some existing tests of the schur complement factory to work with the new inverse approximation factory
- added two navier-stokes tests with explicit usage of the inverse approximation factory in the xml files

Joined work with @mayrmt

PR #10976, opened 2022-09-01
MueLu: Add algorithm to enrich sparse graphs based on static powers

maxfirmbach:
@trilinos/muelu

## Motivation

Add a static pattern determination algorithm enriching the given pattern by applying powers of itself to it.
This should produce better quality sparse approximate inverses for ill-conditioned problems. Especially in the
case of block diagonal matrices (even with different block sizes) a really good approximation can be archived.

## Related Issues
* Related to  #10437, #10484, #10606 and #10909

## Stakeholder Feedback

AMG for block matrices with schur complement

## Testing

- added unit tests for enriched sparse matrix graphs

Joined work with @mayrmt


packages/muelu/src/Misc/MueLu_InverseApproximationFactory_def.hpp
```diff
@@ -143,6 +160,60 @@ namespace MueLu {
     Set(currentLevel, "Ainv", Ainv);
   }
 
+  template <class Scalar, class LocalOrdinal, class GlobalOrdinal, class Node>
+  RCP<const Xpetra::CrsGraph<LocalOrdinal, GlobalOrdinal, Node>>
+  InverseApproximationFactory<Scalar, LocalOrdinal, GlobalOrdinal, Node>::GetStaticPattern(const RCP<const CrsGraph>& sparsityPattern, int power) const {
+
+    // gather missing rows from other procs to generate an overlapping map
+    const RCP<const Import> rowImport = ImportFactory::Build(sparsityPattern->getRowMap(), sparsityPattern->getColMap());
```
cgcgcg, line 168: Does sparsity pattern maybe already have an importer?

cgcgcg:
It might be easier to perform matrix-matrix products to obtain the graph.

maxfirmbach:
> It might be easier to perform matrix-matrix products to obtain the graph.

I also thought about this in the beginning, but for performance reasons I don't need the explicitly calculated values. On the other hand the matrix-matrix product is so optimized it might be faster anyway. Not sure about this ...

cgcgcg:
Yes, you don't need the values. But I also think that you don't want to maintain an implementation of something that's similar to a matrix-matrix product. Also, I don't think that such a method belongs in MueLu. It would be nice if Tpetra could provide the graph of a matrix power.

packages/muelu/src/Misc/MueLu_InverseApproximationFactory_def.hpp
maxfirmbach: I don't think so, at least I don't see were one has been initialized.

cgcgcg: `sparsityPattern` is the graph of A, right? So you could check that `sparsityPattern->getImporter()` returns something non-null.

maxfirmbach:
As our in-house code still uses Epetra I don't have access to Tpetra ... so an implementation in Xpetra would also be fine I guess.

packages/muelu/src/Misc/MueLu_InverseApproximationFactory_def.hpp
maxfirmbach: Yep, I checked, it returns null :sweat_smile: .

jhux2:
@maxfirmbach @mayrmt fyi, PR testing isn't working at the moment, see the pinned issue

packages/muelu/src/Misc/MueLu_InverseApproximationFactory_def.hpp
mayrmt: The `Xpetra::CrsGraph()` can return an importer via `getImporter()`. However, this is only implemented for the `Xpetra::TpetraCrsGraph`.

The `Xpetra::EpetraCrsGraph()` returns a hard-coded `Teuchos::null`. When looking at the implementation of the `Epetra_CrsGraph` in `Epetra`, it looks like the graph is capable of returning an importer. So, it might be just the incompleteness of `Xpetra` that comes into play here. (@maxfirmbach's, if you have tested this in your application code, then you have tested the `Epetra`- route through `Xpetra`.)

cgcgcg: @mayrmt No, it's implemented for Epetra as well. The two specializations for EpetraNode return the importer.

maxfirmbach:
This is my second take on this PR:

- moved the graph creation functionality out of `MueLu` to `Xpetra`
- simplified everything as proposed to now use a simple matrix-matrix product to indirectly get the graph of the result matrix, which should help with maintaining the method
-  also moved to a more simple loop instead of recursion

What do you think @cgcgcg  and @mayrmt ?

packages/xpetra/sup/Utils/Xpetra_GraphUtils.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+  /*!
+    \@brief Get the exponential of a CrsGraph
+
+    @param graph Graph to calculate the exponential of
+    @param levelFill Parameter to control the recursive calculation of the exponential
+
+    @return Exponential CrsGraph with increased fill-in
+  */
+  static RCP<const CrsGraph> getExponentialCrsGraph(const RCP<const CrsGraph>& graph, int levelFill, Teuchos::FancyOStream &out) {
```
cgcgcg, line 84: Is this correct? If I specify `levelFill=3` doesn't this give me `G^{4}`?

maxfirmbach: Correct, in this version the levelFill is not representing the graph exponent directly, which yields the result you described. This could be changed of course … as right now the adjacency matrix is multiplied by itself.

maxfirmbach: I changed the method to exactly produce what it states. So know levelFill=3 produces `G^{3}`, this should be more intuitive.

cgcgcg: Cool, thank you!

maxfirmbach:
Would someone mind posting the issue why the auto-tester failed? I assume it is related to the testing tolerances.

maxfirmbach:
Available now in our own code and `Teko`, thus closing.

PR #10981, opened 2022-09-01
Adelus: Fix to allow permute view value type rather than int

vqd8a:
@trilinos/adelus

## Motivation

This PR adds some changes to allow the permute view value type to be other than ```int``` type for the ```Adelus::Factor``` and ```Adelus::Solve``` interfaces

## Testing

Tests are performed on Lassen and Weaver, and are confirmed by @jdkotul .

### DIFF
--- a/packages/adelus/src/Adelus_factor.hpp
+++ b/packages/adelus/src/Adelus_factor.hpp
@@ -81,7 +81,7 @@ template<class HandleType,
          class ZDView,
          class ViewType1D,
          class ViewType2D,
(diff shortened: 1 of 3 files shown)


jdkotul (approved):
These changes allow users to use different types (views) in the execution of Adelus for the permutation vector.  This is only invoked when the user performs the factor and solve separately.   Adelus then casts these types(views) as needed by Adelus.   This is transparent to the user, but gives them more flexibility in their code.

Issue #10948, opened 2022-08-29
MueLu: Tests "MueLu_Structured_Region_Linear*" fail on geminga and enigma, deviation from gold file


jhux2:
## Bug Report
@trilinos/muelu 

### Description

Linux-sems-gcc-8.3.0-SEMS-OPENMPI-4.0.5_RELEASE_NO-EPETRA_NO-INT_COMPLEX_NO-SERIAL_OPENMP_EXPERIMENTAL 
https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=11222478

Linux-sems-gcc-8.3.0-SEMS-OPENMPI-4.0.5_RELEASE_NO-SERIAL_OPENMP_COMPLEX_EXPERIMENTAL 
https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=11223713

Linux-sems-gcc-8.3.0-SEMS-OPENMPI-4.0.5_RELEASE_NO-EPETRA_NO-INT_COMPLEX_EXPERIMENTAL 
https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=11223849

Linux-sems-gcc-8.3.0-SEMS-OPENMPI-4.0.5_RELEASE_NO-EPETRA_NO-SERIAL_OPENMP_EXPERIMENTAL 
https://testing.sandia.gov/cdash/viewTest.php?onlyfailed&buildid=11222696

Errors like the following:
```
Number of iterations: OK
Using relative error
Residual in iteration 0: OK
Using relative error
Residual in iteration 1: WRONG -- deviation from value in gold file.
End Result: TEST FAILED
```

GrahamBenHarper:
I believe this is related to #9377

pohm01:
After a little investigation, it seems that the initial starting vector is different is different on different builds.

Looking specifically at Linear_Brick3D_Tpetra_MPI_4_regression:
I have two builds on enigma, a serial build that passes and an openmp build that fails. The starting vector between the two builds is not the same. The interesting thing is that the first 128 entries on each rank are the same, and then the remaining 122 entries are different (1000 dofs, 250 on each rank).

So maybe something weird with the random vector generation.

pohm01:
As suggested by @cgcgcg, using the xpetra randomize() implementation fixes the differences in the random starting vectors.

Should I make a PR for the test driver and rerun the gold files to use the xpetra randomize?

mayrmt:
[...]

PR #10983, opened 2022-09-02
MueLu: change Region tests to use xpetra randomize to fix failing tests

pohm01:
@trilinos/muelu

## Motivation
#10948
The randomize() function that is currently being used (tpetra?) to initialize the starting vector is only consistent for the first 128 terms when compared between different build configurations. This is causing the region regression tests to fail.
Switching the randomize() function to be the xpetra implementation appears to fix this.

## Testing
This PR is to fix failing muelu tests.


jhux2 (approved):
@pohm01 Lgtm, thank you

GrahamBenHarper (approved):
I can't believe this was the solution this whole time. Thanks @pohm01 and @cgcgcg :sweat_smile:

mayrmt:
Should this be reported to the Tpetra team, maybe in a separate issue?

pohm01:
@cgcgcg is the tpetra implementation the default randomize() in xpetra? Or is it something else? I forget what you said.

brian-kelley:
@mayrmt I don't think there's a bug here - ``MultiVector::randomize`` uses the Kokkos parallel random generator, so it can't be expected to give the same results from run to run.

From https://docs.trilinos.org/dev/packages/tpetra/doc/html/classTpetra_1_1MultiVector.html:
```
Set all values in the multivector to pseudorandom numbers.

Note
    Do not expect repeatable results. 
    Behavior of this method may or may not depend on external use of the C library routines srand() and rand(). In particular, setting the seed there may not affect it here. 
```

Although... I will say the fact that the output isn't repeatable makes the function less useful. I ran into the same issue on some Ifpack2 tests - solve iteration counts were slightly higher only on some platforms because of a different random RHS, making the test fail. The solution was to save the RHS in a file to ensure it's always the same.

cgcgcg:
I think the Tpetra/Kokkos implementation wasn't producing the same values independent of node type. But that's not a bug, and there isn't any promise that this would be the case. The Xpetra implementation is purely serial, that's why it always gives the same values.

brian-kelley:
@pohm01 It sounds like this fixed the regression test failures, but keep in mind ``Xpetra::MultiVector::randomize(true)`` uses std::rand, which can have different implementations on different platforms/standard libraries. At least it's reproducible on a given platform from run to run, though.

lucbv:
We might want to discuss this at our next MueLu stand-up, in general we like the idea of being able to check that math algorithms are giving identical results from run to run, that's a classic check for correctness. We could use a non-random input vector I guess?

mayrmt (approved):

brian-kelley:
@lucbv One option is to have a deterministic random fill function that uses hashes of the seed and index, instead of the Kokkos RNG state pool. KokkosKernels has the xorshift* hash functions that would work for that, and it's even the same basic RNG that Kokkos random uses so the quality of the randomness would be about the same.

Issue #11019, opened 2022-09-14
MueLu: Randomly failing BlockCrs unit test


GrahamBenHarper:
## Bug Report
@trilinos/muelu 
@csiefer2 

Several PRs are blocked by a randomly failing MueLu BlockCrs unit test. Partial output is below:
```
575. ParameterListInterpreter_std_complex0double0_int_longlong_Kokkos_Compat_KokkosOpenMPWrapperNode_PointCrs_vs_BlockCrs_UnitTest ... WARNING: Utilities::Transpose(): Using inefficient placeholder algorithm for Transpose
WARNING: Using inefficient BlockCrs Multiply Placeholder
... 
 PointH->GetNumLevels() = 3 == BlockH->GetNumLevels() = 3 : passed
 norm = 0 < tol = 1.49012e-08 : passed
 norm = 0 < tol = 1.49012e-08 : passed
 norm = 0 < tol = 1.49012e-08 : passed
 norm = 0 < tol = 1.49012e-08 : passed
 
 p=0: *** Caught standard std::exception of type 'std::invalid_argument' :
 
  /scratch/trilinos/jenkins/ascic166/workspace/Trilinos_PR_gcc-8.3.0/Trilinos/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp:951:
  
  Throw number = 453
  
  Throw test that evaluated to true: domainMapsSame
  
  TpetraExt::MatrixMatrix::Add(): The domain Maps of Op(A) and Op(B) are not the same.
 [FAILED]  (0.0442 sec) ParameterListInterpreter_std_complex0double0_int_longlong_Kokkos_Compat_KokkosOpenMPWrapperNode_PointCrs_vs_BlockCrs_UnitTest
 Location: /scratch/trilinos/jenkins/ascic166/workspace/Trilinos_PR_gcc-8.3.0/Trilinos/packages/muelu/test/unit_tests/ParameterList/ParameterListInterpreter.cpp:184
```

Affected PRs: #10983 #10960

csiefer2:
@jhux2 is looking at this

csiefer2:
Caused by non-deterministic aggregation.  PR up.

PR #11022, opened 2022-09-15
MueLu: Fixing random test fail in #11019

csiefer2:
Issue: #11019
Tests: This
Stakeholder feedback:n/a

Root cause was a comparison test which used non-deterministic aggregation on OpenMP.  Oops.  We probably shouldn't have our default OpenMP aggregation algorithm be non-deterministic, but that's a discussion for another time.


GrahamBenHarper (approved):
yay nondeterministic aggregation on OpenMP builds :)

PR #11028, opened 2022-09-16
Panzer move tangent evaluation to device

rppawlo:
## Motivation

Update tangent code in Panzer to run on device instead of host.

### DIFF
--- a/packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
+++ b/packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
@@ -1135,7 +1135,8 @@ getWeightedMeasure(const bool cache,
       // Calculate measures (quadrature weights in physical space) for this side
       auto side_weighted_measure = Kokkos::DynRankView<Scalar,PHX::Device>("side_weighted_measure",num_evaluate_cells_,num_points_on_side);
       if(cell_dim == 1){
-        Kokkos::deep_copy(side_weighted_measure, side_cub_weights(0));
+        auto side_cub_weights_host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(),side_cub_weights);
+        Kokkos::deep_copy(side_weighted_measure, side_cub_weights_host(0));
       } else {
 
         // Copy from complete jacobian to side jacobian
--- a/packages/panzer/disc-fe/src/evaluators/Panzer_GatherTangents_impl.hpp
+++ b/packages/panzer/disc-fe/src/evaluators/Panzer_GatherTangents_impl.hpp
@@ -59,33 +59,33 @@ panzer::GatherTangents<EvalT, Traits>::
 GatherTangents(
   const Teuchos::ParameterList& p)
 { 
-  dof_name = (p.get< std::string >("DOF Name"));
+  dof_name_ = (p.get< std::string >("DOF Name"));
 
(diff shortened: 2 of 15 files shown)


etphipp (approved):

packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
```diff
@@ -1135,7 +1135,8 @@ getWeightedMeasure(const bool cache,
       // Calculate measures (quadrature weights in physical space) for this side
       auto side_weighted_measure = Kokkos::DynRankView<Scalar,PHX::Device>("side_weighted_measure",num_evaluate_cells_,num_points_on_side);
       if(cell_dim == 1){
-        Kokkos::deep_copy(side_weighted_measure, side_cub_weights(0));
+        auto side_cub_weights_host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(),side_cub_weights);
+        Kokkos::deep_copy(side_weighted_measure, side_cub_weights_host(0));
```
PhilMiller, line 1139: This feels like a terribly inefficient way to accomplish this. How about instead of allocating and copying to the host, just

`Kokkos::parallel_for(policy, [](int i, int j) { side_weighted_measure(i, j) = side_cub_weights(0); } );`

PhilMiller: Though if I'm commenting on initialization code (I can't really tell), then whatever

packages/panzer/disc-fe/src/evaluators/Panzer_GatherTangents_impl.hpp
```diff
@@ -96,11 +96,24 @@ void panzer::GatherTangents<EvalT, Traits>::
...
-  this->utils.setFieldData(pointValues.jac,fm);
-  const shards::CellTopology & parentCell = *basis->getCellTopology();
-  int edgeDim = 1;
-  edgeParam = Intrepid2::RefSubcellParametrization<PHX::Device>::get(edgeDim, parentCell.getKey());
+  auto orientations = d.orientations_;
+  orientations_ = Kokkos::View<Intrepid2::Orientation*>("orientations_",orientations->size());
+  auto orientations_host = Kokkos::create_mirror_view(orientations_);
+  for (size_t i=0; i < orientations->size(); ++i)
+    orientations_host(i) = (*orientations)[i];
+  Kokkos::deep_copy(orientations_,orientations_host);
```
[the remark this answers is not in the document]
rppawlo: oh - also - this is a one time setup cost.

packages/panzer/disc-fe/src/Panzer_IntegrationValues2.cpp
rppawlo: I will change that. Thanks!

crtrott: I wouldn't change it. What you do here with the deep_copy is more efficient.

packages/panzer/disc-fe/src/evaluators/Panzer_GatherTangents_impl.hpp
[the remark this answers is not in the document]
crtrott: We actually fixed a long time ago our deep_copy to fail if there is a padding mismatch. And unmanaged views by definition have to be un-padded.

PR #11051, opened 2022-09-22
MueLu: Update to fix MueLu linking with RDC enabled

seanofthemillers:
@trilinos/MueLu

## Motivation

These changes fix a linking issue when building MueLu with relocatable device code (RDC) for the ROCm/HIP build path (ROCm 5.2.3).

ROCm RDC compiles currently require all object files in each static lib to have unique names. These changes introduce a workaround, while the compiler is addressed.

## Testing

This change only affects compilation. Currently some tests are failing for a static+rdc build of Trilinos with HIP, but this appears to be a separate issue.


seanofthemillers:
@cgcgcg @jhux2 I'm not sure who to set for review on this issue. We were hearing that the Nalu-wind team were having issues building with HIP support for RDC, and these changes helped get past a linking issue, so I wanted to make sure they get upstreamed.

cgcgcg (approved):
I think this is ok. Seems we don't ETI the interface classes automatically using some script, but using the files you modified by hand. So they should not get overwritten on the next update.

PR #11053, opened 2022-09-22
MiniEM: Refactor and work on matrix-free

cgcgcg:
@trilinos/panzer 

## Motivation
- Panzer: Fix UVM dependence in GradBasisDotVector, add evaluator GradBasisDotTensorTimesVector
- MiniEM: Refactor code, allow to solve higher order problem with matrix-free grid transfers


packages/panzer/disc-fe/src/evaluators/Panzer_Integrator_GradBasisDotVector_impl.hpp
```diff
@@ -278,13 +270,13 @@ namespace panzer
       for (int qp(0); qp < numQP; ++qp)
       {
 	team.team_barrier();
-	tmp_(cell) = 1.0;
+        ScalarT fieldMultsTotal(1);
```
rppawlo, line 273: This will not work for AMD machines. If ScalarT is a dynamic fad type, it tries to allocate the derivative array on device while the kernel is running. This capability is not supported by AMD. We refactored all evaluators to not use "ScalarT tmp" but to allocate the views for tmp_ array before the kernel is launched.

rppawlo (commented):
@cgcgcg - can you test these changes to make sure they still run on HIP backend?

packages/panzer/disc-fe/src/evaluators/Panzer_Integrator_GradBasisDotVector_impl.hpp
cgcgcg: Hm. Guess I'll have to revisit that. I changed it and it circumvented some errors on Cuda.

rppawlo: That's weird. I'm routinely building the panzer test suite with uvm off for empire work and not seeing any failures. What kind of failures were you seeing?

cgcgcg: There might be some testing gaps? Otherwise I don't understand how `Integrator_GradBasisDotVector::postRegistrationSetup` worked without UVM.

cgcgcg: Is it possible that there is no testing with "Field Multipliers"? That would explain how `postRegistrationSetup` worked, and it could also explain how I was seeing issues in `operator()`.

rppawlo: That is possible. Let me take a look at the tests...

bartlettroscoe:
See

* #11072

packages/panzer/disc-fe/src/evaluators/Panzer_Integrator_GradBasisDotVector_impl.hpp
cgcgcg: @rppawlo I reverted the change to the `tmp_` array, and the matrix values change compared to a serial run (and the solver fails to converge). With the change, I only saw differences at the order of machine precision.

cgcgcg: @rppawlo  I built on HIP. With the change to `Integrator_GradBasisDotVector` the runs fail with an allocation error. I do get past that without the change, but the symptoms are like on Cuda. Now, it's obviously possible that something else in my PR breaks things in a non-trivial way. Is there a test in Panzer that I could easily extend with field multipliers?

rppawlo: @cgcgcg - there is a simple energy equation example in:
```
Trilinos/packages/panzer/adapters-stk/example/main_driver
```
To add the multiplier, you will need to edit the file:
```
Trilinos/packages/panzer/disc-fe/test/equation_set/user_app_EquationSet_Energy_impl.hpp
```
to add in the multiplier when GradBasisDotVector is constructed for the diffusion term. There are three paths through the kernel - no multipliers, one multiplier and more than one multiplier. We should probably add tests for all three paths at some point.

cgcgcg: @rppawlo I _think_ I got it now. After looking at some of the other evaluators, it seems that the `kokkosFieldMultipliers` inner view should be unmanaged. As far as I can tell, this change fixed my issues on Cuda. I'll rebuild on HIP as well to double check the changes of this PR.

rppawlo: Cool! Yes - inner views must be unmanaged. To make this easier to remember (and handle memory management issues), there are actually three implementations of ViewOfView objects for managing different use cases. Since you have it working, I would not change anything, but in the future, would recommend using the ViewOfViews3 object (see Phalanx_KokkosViewOfViews.hpp) as it makes the inner views unmanaged at runtime by using the pointer constructor - the template parameter is then not needed on the inner views. This works best for backwards compatibility.

cgcgcg: The Panzer tests are all passing on HIP (minus some that failed since the machine has NetCDF without parallel support). I will split out the Tpetra/Ifpack2/MueLu changes to reduce this PR to Panzer only changes.

packages/panzer/mini-em/example/BlockPrec/solverCG.xml
```diff
@@ -43,7 +43,7 @@
               <ParameterList name="Pseudo Block CG">
                 <Parameter name="Convergence Tolerance" type="double" value="1.0e-9"/>
                 <Parameter name="Orthogonalization" type="string" value="ICGS"/>
-                <Parameter name="Maximum Iterations" type="int" value="100"/>
+                <Parameter name="Maximum Iterations" type="int" value="2000"/>
```
rppawlo, line 46: With bumping the iterations significantly, does the test runtime have a corresponding increase?

packages/panzer/mini-em/example/BlockPrec/maxwell-pthOrder.xml
```diff
@@ -25,15 +25,15 @@
...
+        <Parameter name="Z Procs" type="int" value="-1" />
         <Parameter name="X Blocks" type="int" value="1" />
         <Parameter name="Y Blocks" type="int" value="1" />
         <Parameter name="Z Blocks" type="int" value="1" />
-        <Parameter name="X Elements" type="int" value="5" />
-        <Parameter name="Y Elements" type="int" value="5" />
-        <Parameter name="Z Elements" type="int" value="5" />
+        <Parameter name="X Elements" type="int" value="2" />
+        <Parameter name="Y Elements" type="int" value="2" />
+        <Parameter name="Z Elements" type="int" value="2" />
```
cgcgcg, line 36: See here

packages/panzer/mini-em/example/BlockPrec/solverCG.xml
cgcgcg: It does, but I also reduced the test size.

packages/panzer/mini-em/src/solvers/MiniEM_MatrixFreeInterpolationOp.cpp
```diff
@@ -0,0 +1,612 @@
...
+
+
+        for (size_t j = 0; j<numVectors; ++j)
+          li::getBasisCoeffs(Kokkos::subview(basisCoeffsLIOriented_d, Kokkos::ALL(), Kokkos::ALL(), j),
+                             Kokkos::subview(reducedValuesAtDofCoordsOriented_d, Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL(), j),
+                             ho_dofCoeffs_d);
+
+        auto owner_d = owner_d_;
+
+        // ToDo: Make this a functor.
```
rppawlo, line 370: can remove this comment

rppawlo (approved):
@cgcgcg - wow - that was a ton of work! Looks good to me!

cgcgcg:
@rppawlo Thank you for such a quick review! I removed the old comments in the code.

rppawlo (approved):

PR #11054, opened 2022-09-22
tacho: remove volatile qualifiers from join method

ndellingwood:
Compatibility update for disable of kokkos deprecated code

@trilinos/tacho @trilinos/shylu 

## Motivation
Successfully compile with Kokkos' deprecated code disable

## Stakeholder Feedback

NA

## Testing

Locally testing confirmed successful compilation with Kokkos' deprecated code disabled; AT for more testing

### DIFF
--- a/packages/shylu/shylu_node/tacho/src/impl/Tacho_SupernodeInfo.hpp
+++ b/packages/shylu/shylu_node/tacho/src/impl/Tacho_SupernodeInfo.hpp
@@ -52,7 +52,7 @@ struct SuperNodeInfoInitReducer {
     dst.nnz += src.nnz;
   }
 
-  KOKKOS_INLINE_FUNCTION void join(volatile value_type &dst, const volatile value_type &src) const {
+  KOKKOS_INLINE_FUNCTION void join(value_type &dst, const value_type &src) const {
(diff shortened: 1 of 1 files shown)


ndellingwood:
Not sure who is maintaining tacho, @srajama1 can you give this a review on the shylu side? Minimal change, just dropping the `volatile` quantifier from a join method in order to conform with current versions of kokkos when deprecated code is disabled?

ndellingwood:
@iyamazaki would you be able to give a review from the shylu side? This PR has a one line change to remove a volatile qualifier from a join method, a compatibility update needed to compile with kokkos' deprecated code disabled

iyamazaki:
Hi, @ndellingwood. Thank you for the fix!! This looks good, but I was wondering if this `join` function is still used anywhere? If not, I am wondering if we should just remove it?

iyamazaki (approved):
Ah, I now see how it is used.

ndellingwood:
@iyamazaki thanks! The join method is used with a functor in parallel_reduce calls. I had tried removing the overload with the `volatile` qualifier but that broke compilation (the overload was still needed for the `const` qualifier)

PR #11066, opened 2022-09-26
Intrepid2: rename basis function Intrepid2_HVOL_C0_FEM

mperego:
@trilinos/Intrepid2

## Motivation

Orientation tools require basis functions to have unique names.
`Intrepid2_HVOL_C0_FEM` basis function has the same name independently of the basis topology, which create issues in 2d `HCurl` and `HDiv` boundary projections. 
Fixing this by including the topology name into the basis name, e.g. 
`Intrepid2_HVOL_Line_2_C0_FEM`
`Intrepid2_HVOL_Quad_4_C0_FEM`
In this way basis functions with different topologies have different names, which is required by orientations.


CamelliaDPG (approved):
Looks good to me!

PR #11071, opened 2022-09-27
Ifpack2: fix performance bug in OverlappingRowMatrix

brian-kelley:
OverlappingRowMatrix needs to keep track of unique GIDs to add to the column map of the next overlap level. It was doing this with linear search in an unsorted array, for O(N^2) time. Instead, use unordered_set to keep track of unique GIDs to reduce this part to O(N).

@trilinos/ifpack2 

## Motivation

Major speedup in AdditiveSchwarz preconditioner initialize, when overlap level > 0.

Reduces OverlappingRowMatrix construction time in SierraTF abnormal energy surrogate problem from 159 s to 6.8 s. Overall AdditiveSchwarz init time speedup by 2.04x.

## Testing

Tested by Ifpack2 unit tests and Ifpack2_AdditiveSchwarz test.


jhux2 (approved):
lgtm, thanks, @brian-kelley!

PR #11082, opened 2022-09-30
stk: remove use of Kokkos::finalize_all

ndellingwood:
Compatibility update for disable of kokkos deprecated code Necessary for kokkos@4.0 release, see kokkos/kokkos#5134

@trilinos/stk 

## Motivation
Compatibility update for disable of kokkos deprecated code Necessary for kokkos@4.0 release, see kokkos/kokkos#5134

Since stk is snapshot to Trilinos this PR can be closed if changes can be included in the next snapshot (I'm unfamiliar with the process for PRs or similar to the code base)


alanw0 (approved):
This is fine. We have a commit being put into the sierra repository that replaces all stk usage of Kokkos::finalize_all() with Kokkos::finalize().
So this commit can go in, and when our other commit comes into trilinos with a snapshot, this change will still be made.

Issue #11085, opened 2022-09-30
NOX:  NOX is disabled in the rhel7 cuda-11.4.2-uvm-off PR build


bartlettroscoe:
## Bug Report
@trilinos/nox, @trilinos/framework

### Description
It appears that NOX is being disable in the Trilinos PR rhel7 cuda-11.4.2-uvm-off PR builds.  You can see this at:

https://github.com/trilinos/Trilinos/blob/eb0988078d8bb6f6ba43e56f5caa10dc7c2ffc6b/packages/framework/ini-files/config-specs.ini#L1625

and in PR builds showing up on CDash like [here](https://trilinos-cdash.sandia.gov/build/485967/configure) showing:

```
Explicitly disabled packages on input (by user or by default):   ... NOX ...  9
```

### Steps to Reproduce
See https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors

srbdev:
FYI I'm currently working on re-enabling some of these packages (and their relevant tests) for the rhel7 cuda-11.4.2-uvm-off PR builds. You can see the current progress on my [fork](https://github.com/srbdev/Trilinos/commits/framework-pr-cuda11-missing-packages). NOX is enabled again, and it built and passed tests successfully.

sebrowne:
@bartlettroscoe I believe all packages are now enabled in that PR build (see [here](https://trilinos-cdash.sandia.gov/build/658871/configure) for a sample configure output)

bartlettroscoe:
Yup, it looks like NOX is not being disabled in any of the PR builds (except the one Python-only build).  For example, for the [PR #11394 iteration](https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&date=2022-12-15&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=PR-11394-test-&field2=subprojects&compare2=93&value2=NOX) we see NOX tests running in 5 of the 6 PR builds (one is the Python only build that does not test any Trilinos packages).

PR #11057, opened 2022-09-23
Make changes under packages/framework/ trigger enable of all packages (#10881)

bartlettroscoe:
### Description:

These changes can break the PR build and/or tests in any Trilinos package so, to be safe, everything needs to be enabled when changing files under this directory (see #10881).

### Testing

I tested this locally by using the local commits:

```
$ git log-short --name-status HEAD --not github/develop 
f5ecd341e85 "REMOVE ME: Change a file under packages/framework/"
Author: Roscoe A. Bartlett <rabartl@sandia.gov>
Date:   Fri Sep 23 08:28:33 2022 -0600 (5 minutes ago)

M       packages/framework/get_dependencies.sh

0bc59ef04ee "Make changes under packages/framework/ trigger enable of all packages (#10881)"
Author: Roscoe A. Bartlett <rabartl@sandia.gov>
Date:   Fri Sep 23 08:24:00 2022 -0600 (10 minutes ago)

M       commonTools/framework/ProjectCiFileChangeLogic.py
M       commonTools/framework/ProjectCiFileChangeLogic_UnitTests.py

e034f024a7c "Minor typo and format fixes (#10881)"
Author: Roscoe A. Bartlett <rabartl@sandia.gov>
Date:   Fri Sep 23 08:16:58 2022 -0600 (17 minutes ago)

M       commonTools/framework/ProjectCiFileChangeLogic.py
M       commonTools/framework/get-changed-trilinos-packages.sh
```

And then I ran the updated command:

```
$ cd Trilinos/

$ ./commonTools/framework/get-changed-trilinos-packages.sh github/develop HEAD \
  enablePackages.cmake subprojectsList.cmake
```

which gave the file `enablePackages.cmake`:

```
MACRO(PR_ENABLE_BOOL  VAR_NAME  VAR_VAL)
  MESSAGE("-- Setting ${VAR_NAME} = ${VAR_VAL}")
  SET(${VAR_NAME} ${VAR_VAL} CACHE BOOL "Set in enablePackages.cmake")
ENDMACRO()

PR_ENABLE_BOOL(Trilinos_ENABLE_TrilinosFrameworkTests ON)
PR_ENABLE_BOOL(Trilinos_ENABLE_ALL_PACKAGES ON)
```

PR #11087, opened 2022-09-30
Re-enable STK, PIRO, NOX and Panzer in cuda 11 no-uvm builds

rppawlo:
@trilinos/<teamName>

## Motivation

STK, PIRO, NOX and Panzer are all disbled in the cuda 11 no-uvm PR builds. Not sure why. Reenabling to get a baseline to see what we need to fix.

### DIFF
--- a/packages/framework/ini-files/config-specs.ini
+++ b/packages/framework/ini-files/config-specs.ini
@@ -1622,11 +1622,6 @@ use PACKAGE-ENABLES|PR-FRAMEWORK
 [RHEL7_SEMS_CUDA_UVM_OFF_DISABLES]
 
 # Package Disables
-opt-set-cmake-var Trilinos_ENABLE_NOX BOOL FORCE : OFF
-opt-set-cmake-var Trilinos_ENABLE_PanzerAdaptersSTK BOOL FORCE : OFF
-opt-set-cmake-var Trilinos_ENABLE_PanzerMiniEM BOOL FORCE : OFF
-opt-set-cmake-var Trilinos_ENABLE_Piro BOOL FORCE : OFF
-opt-set-cmake-var Trilinos_ENABLE_STK BOOL FORCE : OFF
 opt-set-cmake-var Trilinos_ENABLE_ShyLU_DD BOOL FORCE : OFF
 opt-set-cmake-var Trilinos_ENABLE_ShyLU_NodeTacho BOOL FORCE : OFF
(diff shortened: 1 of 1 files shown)


cgcgcg:
@rppawlo Looks like further down in the file a couple of Piro and Panzer tests are being disabled. Would it make sense to enable them as well?

rppawlo:
Thanks for pointing that out @cgcgcg . I know some of those were disabled due to excessive runtimes early on in ATDM. They might be ok now that most code is correctly using cuda kernels. I will wait for the current round of CI testing to finish and then enable those as well.

bartlettroscoe:
@rppawlo, what this PR motivated by #11085?

Note that changing this file will (or at least should) trigger the enable of every Trilinos package and their tests (see #11057).  And that will run every build, not just this one CUDA build being impacted by this change.

FYI: I have developed a tool to make it easier to drive Trilinos PR builds using GenConfig locally (see [run_trilinos_pr_builds](https://gitlab-ex.sandia.gov/rabartl/run_trilinos_pr_builds/-/blob/master/README.md)). That way, you don't take up PR testing resources just to test out a single build.  For example, if you have access to an ascicgpu machine, you setup the Trilinos repo as:

```
$ cd <trilinos-src>
$ git clone git@gitlab-ex.sandia.gov:rabartl/run_trilinos_pr_builds.git
$ echo /run_trilinos_pr_builds/ >> .git/info/exclude
$ ./packages/framework/get_dependencies.sh --srn
```

then you run:

```
$ cd <some-base-dir>/

$ ln -s <trilinosSrcDir>/run_trilinos_pr_builds/run_trilinos_pr_builds.py .

$ ./run_trilinos_pr_builds.py \
  -b reenable-packages \
  -p STK,Piro,NOX \
  -j 32 -t 8 \
  cuda-11.4.2-uvm-off
```

and that will configure, build, and run tests locally (archiving everything in local log files) and post to CDash (and print the CDash URL).

And that will only enable the packages STK, Piro and NOX with their tests and the packages and the tests downstream from those (instead of everything).

Just an idea that might reduce the load in the PR testing system and provide faster results.

bartlettroscoe:
FYI: Without the PR #11057, this PR would not trigger the testing of anything.  The fact that this PR was posted with the expectation that all packages would get tested is confirmation that the change in #11057 was needed.

rppawlo:
This has been completed in other PRs. Closing.

PR #11090, opened 2022-09-30
Tpetra: Modifications to support GPU-aware MPI with HIP

seanofthemillers:
@trilinos/Tpetra

## Motivation

These changes do two things:
1. Adds some Kokkos::fence calls to the Tpetra distributor actor - it appears that deep copy operations handled by kokkos were not completing before MPI started using the data.
2. Modifies existing cmake code to use GPU-aware naming conventions instead of CUDA-aware.

Note that the cmake changes are backward compatible, so the previous cmake variable Tpetra_ASSUME_CUDA_AWARE_MPI is still valid.

## Testing

These changes were tested against ROCm 5.2.3, using OpenMPI 4.1.4 which was built using OpenUCX 1.13.1.


seanofthemillers:
@lucbv Do you happen to know who can review this? The cmake/naming changes are somewhat invasive.

jhux2:
@trilinos/tpetra @csiefer2

csiefer2:
@seanofthemillers We'll review this PR at the Tpetra Team Meeting on Tuesday!  Thanks for submitting!

seanofthemillers:
@csiefer2 Feel free to make modifications, I wasn't sure what to do about documentation, or updating any of the ATDM builds if that was necessary, so I am expecting some follow-on work.

csiefer2:
@seanofthemillers : @cwpearson and I will be going over this PR  We need to add code pathways for SYCL as well, so our current plan is to add that to your HIP-related changes.

packages/tpetra/CMakeLists.txt
```diff
@@ -260,21 +281,28 @@ ELSE ()
...
+  ${Tpetra_ASSUME_GPU_AWARE_MPI_DEFAULT})
+
+# This is for backward compatability
 TRIBITS_ADD_OPTION_AND_DEFINE(
   Tpetra_ASSUME_CUDA_AWARE_MPI
   TPETRA_ASSUME_CUDA_AWARE_MPI
   "Assume that the MPI implementation that Tpetra uses is CUDA aware.  \"CUDA aware\" means that MPI can use CUDA device allocations (the result of cudaMalloc) as send and receive buffers.  If MPI is CUDA aware, then Tpetra can give CUDA device allocations directly to MPI sends and receives.  MPI implementations might not support all MPI communication operations.  For example, different versions of OpenMPI support different subsets of MPI operations (see https://www.open-mpi.org/faq/?category=runcuda ).  Tpetra developers may safely assume that MPI_*[Ss]end and MPI_*[Rr]ecv work.  Anything more than that should be considered specific to the MPI implementation."
```
cwpearson, line 304: "CUDA aware" -> "GPU-aware"

cwpearson, line 304: "CUDA aware" -> "GPU-aware"

packages/tpetra/core/src/Tpetra_Details_allReduceView.hpp
```diff
@@ -78,12 +79,23 @@ allReduceRawContiguous (const OutputViewType& output,
     // DEEP_COPY REVIEW - This could be either DEVICE-TO-DEVICE or HOST-TO-HOST
     // Either way, MPI is called right afterwards, meaning we'd need a sync on device
```
cwpearson, line 80: Strike this line of the comment since the sync is added

cwpearson (changes requested):
In `Tpetra_Details_DistributorActor.hpp` the added `Kokkos::fence` before the sends is guarded with `Behavior::assumeMpiIsGpuAware`. I think what we really want to know is if the export views were produced by stuff in a GPU execution space.

Unfortunately we can't ask the view that, the best we can do is ask if the export is reachable from a GPU execution space, something like 
`Kokkos::SpaceAccessibility<Kokkos::HIP, typename ExpView::memory_space>::accessible || Kokkos::SpaceAccessibility<Kokkos::Cuda, typename ExpView::memory_space>::accessible` and if so, do a fence. 

Same thing in `Tpetra_Details_allReduceView.hpp` I think. @csiefer2 does that seem right?

packages/tpetra/CMakeLists.txt
cgcgcg: Isn't this for backward compatibility, as stated in the comment?

csiefer2:
I've reached the end of what I'm actually willing to push to your branch.  I'm going to push new branch based on this removing the fence changes, since that seems to be a "Works with CUDA doesn't work with HIP and we need to understand why sort  of thing.  We'll get the CUDA->GPU nomenclature stuff fixed and then revisit the fences.

seanofthemillers:
@csiefer2 Thanks for looking into this. I was confused by the requirement for the kokkos::fence()'s I had added. I had originally thought that the deep_copy operations had fences internally, but apparently for HIP, I needed some additional commands.

PR #11092, opened 2022-10-02
ci: add minimum GitHub token permissions for workflows

boahc077:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using [secure-workflows](https://github.com/step-security/secure-workflows).

The GitHub Actions workflow has a GITHUB_TOKEN with write access to multiple scopes. Here is an example of the permissions in one of the workflow runs:
https://github.com/trilinos/Trilinos/actions/runs/3168617479/jobs/5159937582#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflows:

- stale.yml
- tpetra_muelu_label_to_project.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- [GitHub recommends](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) defining minimum GITHUB_TOKEN permissions.
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>


bartlettroscoe:
CC: @trilinos/framework, @e10harvey

csiefer2 (dismissed):

jhux2:
fyi, this is an automated PR, one of many this company is pushing out

bartlettroscoe:
> Are these failures related to these changes? Looks like they are coming from Jenkins and the workflows changes shouldn't have any impact on Jenkins jobs.

I don't even see any results on CDash.  I am assuming a catastrophic error occurred in the infrastructure.  Actually, looking [above](https://github.com/trilinos/Trilinos/pull/11092#issuecomment-1268240998) it claims there are 11 PR builds being run, including builds `clang-7.0.1` and `clang-9.0.0` which makes no sense given there are only 8 PR builds for other PRs like https://github.com/trilinos/Trilinos/pull/11099#issuecomment-1269142489.  Seems like the autotester is having some problems.

cgcgcg:
It's a PR against master. We don't allow that, and the AT is not giving any useful error message.

bartlettroscoe:
> It's a PR against master. We don't allow that, and the AT is not giving any useful error message.

Then why did the autotester even run the builds against this PR?  When I make the mistake of setting the target branch to 'master' the autotester just ignores the PR.

bartlettroscoe:
Wait, there is a clue [above]() in that it shows:

> Status Flag 'Master Merge AutoTester' - Jenkins Testing: 1 or more Jobs FAILED

So it shows it is running the "Master Merge" set of PR builds.  This is confusing.

csiefer2 (approved):

Issue #10915, opened 2022-08-22
MueLu: adding matrix-free capabilities


GrahamBenHarper:
## Enhancement
@trilinos/muelu 

This is a tracking thread for matrix-free capabilities in MueLu
 - [ ] Matrix-free tentative grid transfers
 - [ ] Matrix-free tentative coarse operators
 - [ ] Extend support
 - [ ] Matrix-free smoothed aggregation
 - [ ] Matrix-free distance Laplacian

etc.

GrahamBenHarper:
begone bot

PR #11100, opened 2022-10-04
Tpetra, Ifpack2, MueLu: Components for matrix-free multigrid

cgcgcg:
@trilinos/tpetra @trilinos/ifpack2 @trilinos/muelu 

## Motivation
- Add optional diagonal to Tpetra::Operator
- Move RefMaxwell coarse solver
- Ifpack2 Chebyshev and Hiptmair for operators
- MueLu Stratimikos interface for operators
- MueLu RAP: Disable reuse on GPU, as it's slower
- MueLu: Instantiate tests on all enabled types
- Clean up MueLu output
- Silence some warnings


GrahamBenHarper (dismissed):
I personally understand what `FakeTpetraRowMatrix` means, but others may not. Would it be better to call it something like `TpetraOperatorAsRowMatrix`? I'm unsure if that suggestion is any better.

I'm excited to see some smoother interfaces generalized for operators too :) Do we know what would happen if a smoother which isn't compatible with operators were selected?

Also related to #10915

cgcgcg:
If we select a smoother that needs more than the provided interface, we throw.

mayrmt:
@GrahamBenHarper wrote

> I personally understand what FakeTpetraRowMatrix means, but others may not. Would it be better to call it something like TpetraOperatorAsRowMatrix? I'm unsure if that suggestion is any better.

I'm with @GrahamBenHarper on this one. `FakeTpetraRowMatrix` just indicates, that something weird is going on without telling me what it actually is. `TpetraOperatorAsRowMatrix` is much more expressive and descriptive.

GrahamBenHarper (approved):

GrahamBenHarper (dismissed):

csiefer2 (approved):

GrahamBenHarper (approved):

PR #11108, opened 2022-10-05
Seacas snapshot zellij oct 2022

gsjaardema:
@trilinos/seacas 

## Motivation

Fixes for Zellij.  Changes were made in SEACAS awhile ago, but are needed in Trilinos to make it easier for team to use.
* Add `offset` command to permit moving the output mesh to a specific location
* Fix output of parallel data to include all maps until clients can handle reduced nemesis data.


dridzal (approved):
This looks good.  I've already cherrypicked these changes from the Seacas repo into my local Trilinos builds, so they should work.

PR #11110, opened 2022-10-05
Piro: enable some tests in Epetra-free build

mperego:
All Piro tests were disabled in a Epetra-free build. 
This PR enables some of them, after implementing minor changes (e.g. use Belos instead of AztecOO) when needed.
In particular it enables Thyra-based NOX unit tests, ROL-based analysis tests and the Tempus-based tests that don't depend on Epetra-based model evaluators.

@trilinos/piro

## Testing

Tested both in a Epetra-free build and in a "standard" build.

### DIFF
--- a/packages/piro/CMakeLists.txt
+++ b/packages/piro/CMakeLists.txt
@@ -37,11 +37,7 @@ ENDIF()
 #
 
 ADD_SUBDIRECTORY(src)
-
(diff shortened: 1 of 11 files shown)


rppawlo (approved):

ikalash:
I'm OK to merge this in and check whether Albany is OK based on the nightlies.

PR #11113, opened 2022-10-05
MueLu: Adding parameters to control performance and tracing of AMG

seanofthemillers:
@trilinos/muelu

## Motivation

These changes address three things:
- MueLu: Adding parameter for aggregation in AMG to reuse coloring results from a previous iteration.
- Ifpack2: Adding parameter to use native SpMV/blas kernels instead of fused SpMV kernel for applying Chebyshev smoother (noticeable performance difference on GPUs with vendor libraries).
- Tpetra: Adding fix for Kokkos profiling regions, which can otherwise cause incorrect results with tracing tools.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
@@ -306,6 +306,7 @@ Chebyshev (Teuchos::RCP<const row_matrix_type> A) :
   assumeMatrixUnchanged_ (false),
   textbookAlgorithm_ (false),
   computeMaxResNorm_ (false),
+  ckUseNativeSpMV_(false),
   debug_ (false)
 {
   checkConstructorInput ();
@@ -337,6 +338,7 @@ Chebyshev (Teuchos::RCP<const row_matrix_type> A,
   assumeMatrixUnchanged_ (false),
   textbookAlgorithm_ (false),
   computeMaxResNorm_ (false),
+  ckUseNativeSpMV_(false),
   debug_ (false)
 {
   checkConstructorInput ();
@@ -382,6 +384,7 @@ setParameters (Teuchos::ParameterList& plist)
   const bool defaultAssumeMatrixUnchanged = false;
   const bool defaultTextbookAlgorithm = false;
   const bool defaultComputeMaxResNorm = false;
+  const bool defaultCkUseNativeSpMV = false;
   const bool defaultDebug = false;
 
   // We'll set the instance data transactionally, after all reads
@@ -403,6 +406,7 @@ setParameters (Teuchos::ParameterList& plist)
   bool assumeMatrixUnchanged = defaultAssumeMatrixUnchanged;
   bool textbookAlgorithm = defaultTextbookAlgorithm;
   bool computeMaxResNorm = defaultComputeMaxResNorm;
+  bool ckUseNativeSpMV = defaultCkUseNativeSpMV;
   bool debug = defaultDebug;
 
   // Fetch the parameters from the ParameterList.  Defer all
@@ -480,6 +484,10 @@ setParameters (Teuchos::ParameterList& plist)
     // would be the proper place to compute the range Map version of
     // userInvDiag.
   }
+
+  // Load the kernel fuse override from the parameter list
+  if (plist.isParameter ("chebyshev: use native spmv"))
+    ckUseNativeSpMV = plist.get("chebyshev: use native spmv", ckUseNativeSpMV);
 
   // Don't fill in defaults for the max or min eigenvalue, because
   // this class uses the existence of those parameters to determine
@@ -686,6 +694,7 @@ setParameters (Teuchos::ParameterList& plist)
   assumeMatrixUnchanged_ = assumeMatrixUnchanged;
   textbookAlgorithm_ = textbookAlgorithm;
   computeMaxResNorm_ = computeMaxResNorm;
+  ckUseNativeSpMV_ = ckUseNativeSpMV;
   debug_ = debug;
 
   if (debug_) {
@@ -1355,7 +1364,7 @@ ifpackApplyImpl (const op_type& A,
 
     if (ck_.is_null ()) {
       Teuchos::RCP<const op_type> A_op = A_;
-      ck_ = Teuchos::rcp (new ChebyshevKernel<op_type> (A_op));
+      ck_ = Teuchos::rcp (new ChebyshevKernel<op_type> (A_op, ckUseNativeSpMV_));
     }
     // W := (1/theta)*D_inv*(B-A*X) and X := X + W.
     // X := X + W
@@ -1374,7 +1383,7 @@ ifpackApplyImpl (const op_type& A,
 
   if (numIters > 1 && ck_.is_null ()) {
     Teuchos::RCP<const op_type> A_op = A_;
-    ck_ = Teuchos::rcp (new ChebyshevKernel<op_type> (A_op));
+    ck_ = Teuchos::rcp (new ChebyshevKernel<op_type> (A_op, ckUseNativeSpMV_));
   }
 
   // The rest of the iterations.
--- a/packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_decl.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_decl.hpp
@@ -87,7 +87,8 @@ private:
   using vector_type = Tpetra::Vector<SC, LO, GO, NT>;
 
 public:
-  ChebyshevKernel (const Teuchos::RCP<const operator_type>& A);
+  ChebyshevKernel (const Teuchos::RCP<const operator_type>& A,
+                   const bool useNativeSpMV=false);
 
   void
   setMatrix (const Teuchos::RCP<const operator_type>& A);
@@ -112,6 +113,10 @@ private:
   std::unique_ptr<multivector_type> V1_;
 
   Teuchos::RCP<vector_type> W_vec_, B_vec_, X_vec_;
+
+  // External override to not fuse operations into a single kernel
+  // And use native blas/SpMV operations
+  bool useNativeSpMV_;
 
   // Do the Import, if needed, and return the column Map version of X.
   vector_type&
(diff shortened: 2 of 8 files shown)


cgcgcg:
@seanofthemillers  Where is this actually reusing the coloring?

cgcgcg:
We might want to integrate the reuse of coloring in the existing reuse mechanism. I'm not sure right now if that works well with our current string option.

seanofthemillers:
> @seanofthemillers Where is this actually reusing the coloring?
Ha, I missed a copy/paste somewhere. I'll fix that it a moment.

cgcgcg:
@trilinos/muelu

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
```diff
@@ -481,6 +485,10 @@ setParameters (Teuchos::ParameterList& plist)
     // userInvDiag.
   }
 
+  // Load the kernel fuse override from the parameter list
+  if (plist.isParameter ("chebyshev: use native spmv"))
+    ckUseNativeSpMV = plist.get("chebyshev: use native spmv", ckUseNativeSpMV);
```
jhux2, line 490: This should also be added to the list of allowable Ifpack2 options, in `Ifpack2_Parameters.cpp`.

jhux2 (commented):
@seanofthemillers Thanks for the contribution.  See my comment.

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
seanofthemillers: Will do.

packages/muelu/src/Utils/MueLu_Monitor.cpp
```diff
@@ -4,6 +4,7 @@ int MueLu::FactoryMonitor::timerIdentifier_=0;
 namespace MueLu {
   PrintMonitor::PrintMonitor(const BaseClass& object, const std::string& msg, MsgType msgLevel) : object_(object) {
     tabbed = false;
+    Kokkos::Profiling::pushRegion(msg);
```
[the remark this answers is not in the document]
seanofthemillers: Ah, I wasn't seeing any MueLu tracing regions in my analysis, and couldn't find a flag to enable it, so I just dropped this into the Factory/SubMonitor via the PrintMonitor. I can remove this line since it feels a bit out of place. Is there a flag/env-var that enables kokkos tracing in MueLu?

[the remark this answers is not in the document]
cgcgcg: You could try `Teuchos_KOKKOS_PROFILING=ON`.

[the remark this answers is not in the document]
jhux2: @seanofthemillers I've been able to get traces before.  Do have the Kokkos profiling tools enabled?

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
seanofthemillers: Fixed.

packages/muelu/src/Utils/MueLu_Monitor.cpp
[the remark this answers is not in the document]
seanofthemillers: I have removed this line. I have the `Teuchos_KOKKOS_PROFILING` on and tested with the roctx tracer from the kokkos-tools repo, but was not able to get MueLu factories to show up in the trace (I had lots of Belos and Tpetra regions though). I'm probably missing something that enables kokkos profiling in Teuchos's timers - I'll look at that the next chance I get.

packages/muelu/src/Graph/UncoupledAggregation/MueLu_UncoupledAggregationFactory_kokkos_def.hpp
```diff
@@ -284,79 +304,109 @@ namespace MueLu {
...
+      bool build_coloring = true;
+      const bool reuse_coloring = pL.get<bool>("aggregation: reuse coloring");
+
+      // Existing coloring data is stored as user-data in the currentLevel
+      if(reuse_coloring and currentLevel.IsAvailable(coloring_data_name))
+        build_coloring = false;
+        
+      TEUCHOS_TEST_FOR_EXCEPTION(reuse_coloring && (not pL.get<bool>("aggregation: deterministic")), 
+                                 Exceptions::RuntimeError, 
+                                 "MueLu::UncoupledAggregationFactory::Build: Request to reuse coloring requires aggregation to be deterministic!");
```
cgcgcg, line 319: Why do we need deterministic aggregation for color reuse?

seanofthemillers:
@cgcgcg You mentioned that I should change the name from "aggregation: reuse coloring" = ON/OFF. Would "reuse: type" = "coloring" be preferable?

packages/muelu/src/Graph/UncoupledAggregation/MueLu_UncoupledAggregationFactory_kokkos_def.hpp
seanofthemillers: Yup, it appears that it will change the coloring each iteration, so we need it to be deterministic. I assume. I get segfaults if I allow coloring reuse with non-deterministic.

cgcgcg: I don't think so. We just need valid D2 coloring. The non-deterministic one should also work unless I'm missing something.

cgcgcg:
@seanofthemillers Something along those lines, but we need to double-check that this isn't orthogonal to "reuse: type".

packages/muelu/src/Graph/UncoupledAggregation/MueLu_UncoupledAggregationFactory_kokkos_def.hpp
seanofthemillers: Still tracking this down. Currently this error appears to be tied to the reuse of the level 1 aggregation algorithm `AggregationPhase1Algorithm_kokkos::BuildAggregatesRandom` (and likely others), which results in a `numNonAggragatedNodes<0` condition.

cgcgcg: Wait, numNonAggregatedNodes is actually negative?

seanofthemillers: I see. Is there an easy way to check if the graph has changed (without storing the previous graph)?

cgcgcg: Well, we are calling the aggregation because we want to generate a new graph. So I think the assumption should be that it changes.

seanofthemillers:
@jhux2 I would be happy to, but I'm curious, is there a way to check if these options are being used? As it stands, the solver will work fine if the option is ignored since this is more of a performance improvement. I'll get started on a test.

packages/muelu/src/Utils/MueLu_Monitor.cpp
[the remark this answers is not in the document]
seanofthemillers: Ah, figured it out. I needed to set the verbosity flag to "timings" in the preconditioner file in order to enable the Teuchos stacked timer in the TimeMonitor, which then triggers the kokkos profiling regions.

seanofthemillers:
@cgcgcg @jhux2 I have addressed your comments (I think) and have added a new test, though it is a little limited. I have used a tracing API to confirm that the flags are indeed being obeyed, but the flags become ignored at some point, the test will likely still pass. Could you give this another look?

cgcgcg:
@seanofthemillers We will. Just to better understand: what is the motivation for keeping aggregate information when you can keep tentative P instead?

seanofthemillers:
@cgcgcg Ah, this might just be me misunderstanding what tentative P is. For the problem I'm looking at, I am assuming that the values stored in matrix A are changing on each re-use, but sparsity of A is remaining the same. When I first looked through the factories, I saw that the aggregation was purely topology based, while tentative P contained values. Are the values in tentative P are just defined from the aggregation/topology or do they depend on the values in A? Ptent assembly seems to require the Nullspace of A, and has some QR factorization component (based on the trace), so I assumed it depended on A's values.

cgcgcg:
Does you use case involve changing the Nullspace? Tentative P does not use the values of A.

cgcgcg:
You could cherry-pick d1dc49e76675b83d104573a60eb3ddf42ee21d44 from PR #11100. (I'm having some issues getting that PR merged just yet.)

csiefer2:
@seanofthemillers What's the logic on the Tpetra changes?  Something that corrects for people enabling and disabling profiling within the life of a profiling region?

seanofthemillers:
@csiefer2 The Tpetra changes are there because the ProfilingRegion can enable "groups" of profiling regions. If a group is not enabled, then the ProfilingRegion class will still pop the top region in its destructor. These changes make it so it only pops a region if a region is active.

csiefer2:
@cgcgcg From my inspection, the seems to be passing now on geminga.  Do you concur?

csiefer2:
@seanofthemillers Not following.  The existing code only pushes/pops a region if Behavior::profilingRegionUseKokkosProfiling() is true.  Is there something wrong in the logic of that function?

seanofthemillers:
@csiefer2 Ah, take a look at the `timeit` variable as part of the `ProfilingRegion (const char name[], const char group[])` constructor. It will turn off the profiling region if the `group` is not enabled, but it will always pop the top region in the destructor independent of a group being enabled.

cgcgcg:
@seanofthemillers  I'd suggest to remove the MueLu changes and merge the Tpetra and Ifpack2 stuff.

seanofthemillers:
@cgcgcg Will do, I'll remove the MueLu changes.

seanofthemillers:
@cgcgcg MueLu changes have been removed.

cgcgcg (approved):

PR #11123, opened 2022-10-07
convert GCC 7.2.0 debug to to GCC 8.3.0

jmlapre:
@trilinos/framework 

## Motivation
Migrating to C++17 across all compilers.  Updating from gcc 7.2.0 to gcc 8.3.0 while we're doing so as gcc 7.2.0 is being retired.


e10harvey (changes requested):
In general, everything looks great.

This won't pass PR testing since we've removed the gcc 7.2.0 configuration settings here. Please change the target branch to https://github.com/e10harvey/Trilinos/tree/sprint17_cpp17.

e10harvey:
Closing in favor of https://github.com/e10harvey/Trilinos/pull/11

PR #11131, opened 2022-10-10
Teuchos: Fix parallel XML reader to correctly handle special characters

cgcgcg:
@trilinos/teuchos 

## Motivation
Fixes parallel reading XML input decks with special characters in them:
```
<Parameter name="sigma" type="string" value="if (x >= 0.0 &amp;&amp; y >= 0.0 &amp;&amp; x &lt;= 0.5 &amp;&amp; y &lt;= 0.5)"/>
```

### DIFF
--- a/packages/teuchos/parameterlist/src/Teuchos_XMLObjectImplem.cpp
+++ b/packages/teuchos/parameterlist/src/Teuchos_XMLObjectImplem.cpp
@@ -210,7 +210,7 @@ std::string XMLObjectImplem::toString() const
   std::string rtn;
   if (content_.length()==0 && children_.length()==0)
   {
-    rtn = terminatedHeader() + "\n";
+    rtn = terminatedHeader(true) + "\n";
   }
   else
   {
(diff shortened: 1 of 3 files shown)


jhux2 (approved):
lgtm

packages/teuchos/parameterlist/src/Teuchos_XMLObjectImplem.cpp
```diff
@@ -210,7 +210,7 @@ std::string XMLObjectImplem::toString() const
   std::string rtn;
   if (content_.length()==0 && children_.length()==0)
   {
-    rtn = terminatedHeader() + "\n";
+    rtn = terminatedHeader(true) + "\n";
```
bartlettroscoe, line 213: This makes this consistent with:

https://github.com/cgcgcg/Trilinos/blob/00ef95f4fd641f824502e499eca5c833fd354210/packages/teuchos/parameterlist/src/Teuchos_XMLObjectImplem.cpp#L249

so this seems logical.

Not sure why you would **not** want to terminate using strict XML syntax.  What is that use case for?

bartlettroscoe (approved):
Looks very logical to me.  And thanks for the test :-)

PR #11136, opened 2022-10-11
Xpetra&MueLu: Stop warnings about unused short names

cgcgcg:

## Motivation
Silence warnings about unused short names. Requires C++17, so cannot merge yet.


jhux2:
Oh, that'll be nice!

jhux2 (approved):
lgtm

csiefer2:
How did you even find out that feature exists?

GrahamBenHarper (approved):
How standard is this? Is this going to break some obscure compiler?

lucbv:
@GrahamBenHarper Trilinos will start requiring c++17 very soon so it better means it supports it too: https://en.cppreference.com/w/cpp/language/attributes

csiefer2:
It is in the C++17 standard, which we do require full support for...

PR #11138, opened 2022-10-12
Ctest/Drivers: Update to lightsaber

csiefer2:
Update to lightsaber driver since something decided to hate csh

### DIFF
--- a/cmake/ctest/drivers/lightsaber/cron_driver.sh
+++ b/cmake/ctest/drivers/lightsaber/cron_driver.sh
@@ -1,4 +1,4 @@
-#!/bin/csh
+#!/bin/bash
 
 echo
 echo "Starting nightly Trilinos development testing on lightsaber: `date`"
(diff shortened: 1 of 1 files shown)


cgcgcg (approved):

cmake/ctest/drivers/lightsaber/cron_driver.sh
```diff
@@ -8,36 +8,36 @@ echo
...
-setenv TDD_HTTP_PROXY "http://wwwproxy.sandia.gov:80"
-setenv TDD_HTTPS_PROXY "https://wwwproxy.sandia.gov:80"
-setenv http_proxy "http://wwwproxy.sandia.gov:80"
-setenv https_proxy "https://wwwproxy.sandia.gov:80"
-setenv TDD_FORCE_CMAKE_INSTALL 1
-setenv TDD_DEBUG_VERBOSE 1
+export TDD_HTTP_PROXY="http://wwwproxy.sandia.gov:80"
+export TDD_HTTPS_PROXY="https://wwwproxy.sandia.gov:80"
+export http_proxy="http://wwwproxy.sandia.gov:80"
+export https_proxy="https://wwwproxy.sandia.gov:80"
```
jhux2, line 26: `wwwproxy.sandia.gov` should be `proxy.sandia.gov`.

jhux2 (changes requested):

cmake/ctest/drivers/lightsaber/cron_driver.sh
csiefer2: You're right, but since we don't actually use external network connections, it doesn't matter.

jhux2: just delete?

jhux2 (approved):
I laughed, I cried

PR #11142, opened 2022-10-13
Intrepid2: Fix some shadow variable warnings

japlews:
@trilinos/intrepid2

## Motivation
Fix shadow warnings that show up in Sierra builds, caused by some new headers in Intrepid2

@CamelliaDPG @mperego


CamelliaDPG (approved):
Works for me.  Thanks, @japlews!

PR #11146, opened 2022-10-14
rol: remove volatile from join in ROL_TpetraMultiVector.hpp

ndellingwood:
Compatibility update following kokkos@3.7
See kokkos/kokkos#4077 kokkos/kokkos#5194

@trilinos/rol 

## Motivation
Compatibility update needed to compile with Kokkos' deprecated code disabled


dridzal (approved):
If this is no longer needed, then the change is fine.

PR #11155, opened 2022-10-17
MueLu: Update to fix additional MueLu linking

kuberry:
- Completes the renaming of files affected by commit 4f6e0e89

@trilinos/muelu 

## Motivation

PR #11051 introduced the following type of errors:
```make
make[1]: *** Waiting for unfinished jobs....
[ 91%] Linking CXX shared library libmuelu.dylib
Undefined symbols for architecture x86_64:
  "MueLu::FacadeClassBase<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >::FacadeClassBase()", referenced from:
      MueLu::FacadeSimple<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >::FacadeSimple() in ETI_MueLu_FacadeClassFactory.cpp.o
      MueLu::FacadeBGS2x2<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> >::FacadeBGS2x2() in ETI_MueLu_FacadeClassFactory.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

because it renamed half of the files "ETI_*" and changed the rule to expect "ETI_*" style naming, but left four files related to explicit template instantiation with the original naming convention.

This PR renames the additional four files and fixes this error.


jhux2:
@kuberry Thanks for reporting.

> How come we don't see that in our nightlies?

@cgcgcg This appears to be a MacOS build, so that's one big difference compared to our nightlies.  I don't find any tests that exercise the Facade class, so could the linker be pickier?

mayrmt:
I had some trouble with linking the Facade classes recently as well, though I don't recall the details. This was on a Ubuntu machine with gcc@9.3.

mperego:
@trilinos/muelu can you please inspect/approve this?

cgcgcg (approved):

PR #11157, opened 2022-10-18
Teko fix return of temporary address

kuberry:
@trilinos/teko 

## Motivation

const char * function returns address to temporary variable
```make
/trilinos/packages/teko/src/Epetra/Teko_EpetraHelpers.hpp:167:12: warning: returning address of local temporary object [-Wreturn-stack-address]
   {return ("zeroed( "+std::string(epetraOp_->Label())+" )").c_str(); }
```

This PR adds a `label_` std::string that is set at construction (all info needed is there at that time) and the c_str() is then returned for label. This is consistent with `/trilinos/packages/teko/src/Epetra/Teko_EpetraOperatorWrapper.hpp`


cgcgcg (approved):
No idea why we are returning the label as a `char*` in the first place, but this looks good to me.

Issue #10999, opened 2022-09-08
Framework: Mass gnu and clang PR test failures on new 'ascic0xyz' machines


csiefer2:
Link:
https://trilinos-cdash.sandia.gov/test/11636999

Sample error:
```
/scratch/trilinos/workspace/Trilinos_PR_gcc-7.2.0-debug@2/pull_request_test/packages/amesos/example/Amesos_a_trivial_mpi_test.exe: symbol lookup error: /projects/sems/install/rhel7-x86_64/sems/compiler/gcc/7.2.0/openmpi/1.10.1/lib/libmca_common_verbs.so.7: undefined symbol: ompi_common_verbs_usnic_register_fake_driver
```

Reported in [TRILINOSHD-203](https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-203).

csiefer2:
Nothing like watching over 2,000 tests fail...

bartlettroscoe:
Hum, I saw the same error when trying to use one of the GenConfig builds on an 'hpws' machine showing runtime errors like:

```
1: /fgs/rabartl/Trilinos.base/BUILDS/ATDM/CHECKIN/sems-rhel7-intel-18.0.5-openmp-shared-release-debug/packages/teuchos/core/test/Allocator/TeuchosCore_Allocator_UnitTest.exe: symbol lookup error: /projects/sems/install/rhel7-x86_64/sems/compiler/intel/18.0.5/openmpi/1.10.1/lib/libmca_common_verbs.so.7: undefined symbol: ompi_common_verbs_usnic_register_fake_drivers
```

See [TRILINOSHD-59](https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-59) and:

* https://github.com/trilinos/Trilinos/issues/10022#issuecomment-1239438691
* https://github.com/trilinos/Trilinos/issues/10836#issuecomment-1216017031

bartlettroscoe:
NOTE: This PR build is running on the machine [ascic0193](https://trilinos-cdash.sandia.gov/viewSite.php?siteid=96&project=1&currenttime=1662609600) which is one of the 'ascic0xy' machines that was supposed to have been taken out of the pool of PR testing machines (see #10893).  How did it get back on?

bartlettroscoe:
[...]

PR #11172, opened 2022-10-20
Teuchos: Add a constructor to Teuchos::RCP

kliegeois:
@trilinos/teuchos 

## Motivation

`Teuchos::RCP` is currently missing one constructor of the `std::shared_ptr` ([constructor (8) here](http://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr)) used for instance by pybind11 with multiple inheritance  https://github.com/pybind/pybind11/issues/1306 .

This PR adds the extra constructor.

### DIFF
--- a/packages/teuchos/core/src/Teuchos_RCPDecl.hpp
+++ b/packages/teuchos/core/src/Teuchos_RCPDecl.hpp
@@ -546,6 +546,11 @@ public:
   template<class T2>
   inline RCP(const RCP<T2>& r_ptr);
 
+  /** \brief Construct from another <tt>RCP<T2></tt> and from a raw pointer.
+   */
+  template<class T2>
+  inline RCP(const RCP<T2>& r_ptr, T* ptr);
+
   /** \brief Removes a reference to a dynamically allocated object and possibly deletes
    * the object if owned.
    *
--- a/packages/teuchos/core/src/Teuchos_RCPDecl.hpp
+++ b/packages/teuchos/core/src/Teuchos_RCPDecl.hpp
@@ -546,7 +546,14 @@ public:
   template<class T2>
   inline RCP(const RCP<T2>& r_ptr);
 
-  /** \brief Construct from another <tt>RCP<T2></tt> and from a raw pointer.
+  /** \brief Aliasing constructor: Construct using the ownership of a <tt>RCP<T2></tt> and from a raw pointer.
+   * 
+   * Constructs a <tt>RCP<T></tt> which shares ownership information with the initial 
+   * value of r_ptr, but holds an unrelated and unmanaged pointer ptr.
+   * 
+   * This constructor corresponds to the constructor 
+   * template< class Y > shared_ptr( const shared_ptr<Y>& r, element_type* ptr ) noexcept;
+   * of the std::shared_ptr.
    */
   template<class T2>
   inline RCP(const RCP<T2>& r_ptr, T* ptr);
--- a/packages/teuchos/core/src/Teuchos_RCP.hpp
+++ b/packages/teuchos/core/src/Teuchos_RCP.hpp
@@ -300,6 +300,15 @@ RCP<T>::RCP(const RCP<T2>& r_ptr)
 
 
 template<class T>
+template<class T2>
+inline
+RCP<T>::RCP(const RCP<T2>& r_ptr, T* ptr)
+  : ptr_(ptr), 
+    node_(r_ptr.access_private_node())
+{}
+
+
+template<class T>
 inline
 RCP<T>::~RCP()
 {}
--- a/packages/teuchos/core/test/MemoryManagement/RCP_UnitTests.cpp
+++ b/packages/teuchos/core/test/MemoryManagement/RCP_UnitTests.cpp
@@ -1116,6 +1116,23 @@ TEUCHOS_UNIT_TEST( RCP, Fix_createRCPWithBadDealloc )
 }
 
 
+/**
+ * @test Test the aliasing constructor.
+ */
+TEUCHOS_UNIT_TEST( RCP, aliasing_constructor )
+{
+  Teuchos::RCP<A> a_rcp = Teuchos::make_rcp<A>(1,2);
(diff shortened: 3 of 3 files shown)


cgcgcg:
Also add something to an RCP unit test?

bartlettroscoe (commented):
I would really like to see the use case for this and a unit test.   What is the assumed relationship between `r_ptr` and `ptr`?  When is this safe to use?  (I can think of several cases where this is not safe to use.)

bartlettroscoe:
> Regardless of the use case, this is a missing feature of the `Teuchos::RCP` compared to the `std::shared_ptr`.

Then so be it.  I am just curious how the memory management of a modern C++ code gets into a state where such a feature is required.

bartlettroscoe:
> In practice though, I think that this constructor is used for upcasting/downcasting.

Yikes!  That is what these functions are for:

* [`RCP< T 2> rcp_implicit_cast (const RCP< T1 > &p1)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1RCP.html#a61fc9f719ee13e5d961dcf47053883b4)
* [`RCP< T2 > 	rcp_static_cast (const RCP< T1 > &p1)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1RCP.html#a1c369381869737e67be8cab843b18076)
* [`RCP< T2 > 	rcp_const_cast (const RCP< T1 > &p1)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1RCP.html#a6931e6c2f74c9fa9467e5dc91c6cd983)
* [`RCP< T2 > 	rcp_dynamic_cast (const RCP< T1 > &p1, bool throw_on_fail=false)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1RCP.html#a1580c807c6df3fdbcb953687b4ff60c7)

Regular C++ client code should not be messing around with raw pointers.

bartlettroscoe:
@kliegeois, it has been a long time since I wrote this software but I forgot that these types of non-ideal cases are already supported through the non-member constructor functions:

* [`RCP< T > 	rcpWithEmbeddedObjPreDestroy (T *p, const Embedded &embedded, bool owns_mem=true)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1RCP.html#a02148aef172f7b2c0fb2de29cb01b91a)
* [`RCP< T > 	rcpWithEmbeddedObjPostDestroy (T *p, const Embedded &embedded, bool owns_mem=true)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1RCP.html#a8272eea6a27f1fce6c30fc5d568f2de7)
* [`RCP< T > 	rcpWithEmbeddedObj (T *p, const Embedded &embedded, bool owns_mem=true)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1RCP.html#a506210a123b6e1747d910f037065bd73)

See the section "5.9.4 Embedded objects" in [Teuchos C++ Memory Management Classes, Idioms, and Related Topics: The Complete Reference](https://docs.trilinos.org/dev/packages/teuchos/doc/html/TeuchosMemoryManagementSAND.pdf).

But I am fine with adding this new constructor to match `std::shared_ptr`.

kliegeois:
@bartlettroscoe thanks for all your comments! Yes, that's actually very similar. I will look how you wrote the unit tests for them to write a similar one for the constructor.

I am not a fan of having to add this constructor but sadly it is required if we want to use Pybind11 with rcp (and there are couple of projects that have interests in that). I will update the PR asap with the unit test.

kliegeois:
@bartlettroscoe the test passed on my workstation. If the test looks good to you, I will remove the WIP tag and put the retest one.

packages/teuchos/core/src/Teuchos_RCPDecl.hpp
```diff
@@ -546,6 +546,11 @@ class RCP {
   template<class T2>
   inline RCP(const RCP<T2>& r_ptr);
 
+  /** \brief Construct from another <tt>RCP<T2></tt> and from a raw pointer.
+   */
```
bartlettroscoe, line 550: It might be good to explain the assumptions about the lifetime of `ptr` in relation to the created `RCP` object or just mention the relationship to `std::shared_ptr`.

bartlettroscoe (changes requested):

bartlettroscoe:
@kliegeois, I clicked the review button too early but it would be good to update the comment a little as per [above](https://github.com/trilinos/Trilinos/pull/11172#discussion_r1001229788).  Other than, that this looks good to me.

kliegeois:
Thanks @bartlettroscoe for the review! I updated the documentation.

packages/teuchos/core/src/Teuchos_RCPDecl.hpp
```diff
@@ -546,6 +546,18 @@ class RCP {
...
 
+  /** \brief Aliasing constructor: Construct using the ownership of a <tt>RCP<T2></tt> and from a raw pointer.
+   * 
+   * Constructs a <tt>RCP<T></tt> which shares ownership information with the initial 
+   * value of r_ptr, but holds an unrelated and unmanaged pointer ptr.
+   * 
+   * This constructor corresponds to the constructor 
+   * template< class Y > shared_ptr( const shared_ptr<Y>& r, element_type* ptr ) noexcept;
+   * of the std::shared_ptr.
+   */
```
bartlettroscoe, line 557: Look great.  Thanks!

bartlettroscoe (approved):
With the updated documentation, this looks good to merge (once it passes the PR builds).

kliegeois:
Thanks @bartlettroscoe !

kliegeois:
@bartlettroscoe can you access the output of the auto tests ? Do you know how/why it failed ?

bartlettroscoe:
> @bartlettroscoe can you access the output of the auto tests ? Do you know how/why it failed ?

Yup, the same old thing:

* https://github.com/trilinos/Trilinos/issues/10999

Not sure this keeps happening.

Issue #11109, opened 2022-10-05
Framework: Trilinos PR  ats2 build using cuda-10.1.243 on 'vortex' is irrelevant, customers now using cuda-1.2.152


bartlettroscoe:
## Bug Report
@trilinos/framework, @rppawlo, @ccober6, @e10harvey 

### Description
The current Trilinos PR ats2 CUDA build which is using cuda-10.1.243 on 'vortex' is irrelevant because customers (e.g. SPARC) have been using cuda-11.2.152 on 'vortex' since April 26, 2022

Given, as shown in [this SPARC CDash query](https://sparc-cdash.sandia.gov/index.php?project=SPARC&begin=2022-01-01&end=2022-10-04&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=ats2-v100_cuda-11.2.152) and [this SPARC CDash query](https://sparc-cdash.sandia.gov/index.php?project=SPARC&begin=2022-01-01&end=2022-10-04&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=ats2-v100_cuda-10.1.243), SPARC switched to using cuda-11.2.152 on 2022-04-26 ...

Given the move to C++17 and requiring CUDA 11 and the long known problems on this system killing PR builds:

* #6861
* #7122 

...

Given that a Trilinos PR CUDA 11 build has been running for some time on 'ascicgpu' machines ...

Can we ask that this problematic Trilinos PR build be removed from Trilinos PR testing?  If you want to keep it for develop-to-master merge testing (and then perhaps just run as an informational build), then fine.  But please remove this from blocking the merges of PRs.

e10harvey:
CUDA 10 will be removed as part of the C++17 transition. CUDA 10 and CUDA 11 currently result in a different set of package enables. In order to preserve code coverage for the cuda toolchain, I suggest waiting until the C++17 transition is complete before disabling CUDA 10. That being said, we can certainly disable CUDA 10 now if the product leaders want that.

bartlettroscoe:
What important Trilinos customer currently needs CUDA 10 support from Trilinos?

e10harvey:
> What important Trilinos customer currently needs CUDA 10 support from Trilinos?

@ccober6 ?

From my perspective, we don't want to loose code coverage with the CUDA toolchain by disabling CUDA 10 before the C++17 transition.

rppawlo:
[...]

Issue #6861, opened 2020-02-19
Random mass test failures showing "Error initializing RM connection. Exiting" on 'vortex'


bartlettroscoe:
CC: @jjellio 

As shown in [this query](https://testing-dev.sandia.gov/cdash/index.php?project=Trilinos&begin=2020-02-16&end=2020-02-19&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=65&value1=Trilinos-atdm-ats2&field2=testsfailed&compare2=43&value2=500), we are getting random mass failures on test on 'vortex'.  As shown in that query, when it occurs in a given build, it impacts over a thousand tests and it is random about which build and which days it impacts.  When it occurs, the failures output like:

```
Error: Remote JSM server is not responding on host vortex5902-19-2020 03:31:02:827 114114 main: Error initializing RM connection. Exiting.
```

bartlettroscoe:
@jjellio, I am not sure how much detail we can put in this issue but at least this gets this on the board.

I am setting up to filter out tests that show this failure from the CDash summary emails in the filters:

* Promoted ATDM Trilinos builds: https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2020-02-19&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=ATDM&field2=buildname&compare2=65&value2=Trilinos-atdm-&field3=status&compare3=62&value3=passed&field4=testoutput&compare4=94&value4=Error%3A%20Remote%20JSM%20server%20is%20not%20responding%20on%20host%20vortex

* Specialized ATDM Trilinos Builds Cleanup: https://testing-dev.sandia.gov/cdash/queryTests.php?project=Trilinos&date=2020-02-18&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Specialized&field2=buildname&compare2=65&value2=Trilinos-atdm-&field3=status&compare3=62&value3=passed&field4=testoutput&compare4=94&value4=Error%3A%20Remote%20JSM%20server%20is%20not%20responding%20on%20host%20vortex

That will allow me to start cleaning up the failing tests in these builds but not allow these failures to flood out the the emails and make them worthless.

bartlettroscoe:
I just pushed the following `TrilinosATDMStatus` repo commit:

```
*** Base Git Repo: TrilinosATDMStatus
commit b3c45994e778b3d784498f20c8116c419f5a08ed
[...]

PR #11175, opened 2022-10-21
Improve the TriBITS Netcdf and MPI TPLs

ibaned:
@trilinos/tribits 

## Motivation

This is motivated by issues TriBITSPub/TriBITS#533 and TriBITSPub/TriBITS#534, which are respectively:
1. The desire for the Netcdf TriBITS TPL to load only the CMake config files installed by NetCDF and HDF5, using the most modern CMake approach to bringing in those packages (using the targets that those projects define)
2. The desire to be able to use a custom implementation of MPI (not a compile wrapper but actually an include dir and library) to supply the TriBITS MPI TPL

## Stakeholder Feedback

As both the TriBITS contributor and also the ASC IC PI stakeholder, I would really like the capabilities listed in Motivation.

## Testing

That's what this branch is for!

### DIFF
--- a/cmake/tribits/common_tpls/FindTPLHDF5.cmake
+++ b/cmake/tribits/common_tpls/FindTPLHDF5.cmake
@@ -4,17 +4,18 @@
 
 if (Netcdf_ALLOW_MODERN)
 
-  #set(minimum_modern_HDF5_version 4.7.4)
+  set(minimum_modern_HDF5_version 1.13.2)
   print_var(Netcdf_ALLOW_MODERN)
   message("-- Using find_package(HDF5 ${minimum_modern_HDF5_version} CONFIG) ...")
   find_package(HDF5  ${minimum_modern_HDF5_version}  CONFIG)
   if (HDF5_FOUND)
     message("-- Found HDF5_CONFIG=${HDF5_CONFIG}")
     message("-- Generating Netcdf::all_libs and NetcdfConfig.cmake")
+    message("-- HDF5_EXPORT_LIBRARIES=${HDF5_EXPORT_LIBRARIES}")
     tribits_extpkg_create_imported_all_libs_target_and_config_file(
       HDF5
       INNER_FIND_PACKAGE_NAME  HDF5
-      IMPORTED_TARGETS_FOR_ALL_LIBS   HDF5::HDF5)
+      IMPORTED_TARGETS_FOR_ALL_LIBS   ${HDF5_EXPORT_LIBRARIES})
   endif()
 
 endif()
--- a/cmake/tribits/common_tpls/FindTPLHDF5.cmake
+++ b/cmake/tribits/common_tpls/FindTPLHDF5.cmake
@@ -2,78 +2,97 @@
 # See associated tribits/Copyright.txt file for copyright and license! #
 ########################################################################
 
-#
-# First, set up the variables for the (backward-compatible) TriBITS way of
-# finding HDF5.  These are used in case find_package(HDF5 ...) is not called
-# or does not find HDF5.  Also, these variables need to be non-null in order
-# to trigger the right behavior in the function
-# tribits_tpl_find_include_dirs_and_libraries().
-#
+if (Netcdf_ALLOW_MODERN)
 
-set(REQUIRED_HEADERS hdf5.h)
-set(REQUIRED_LIBS_NAMES hdf5)
+  #set(minimum_modern_HDF5_version 4.7.4)
+  print_var(Netcdf_ALLOW_MODERN)
+  message("-- Using find_package(HDF5 ${minimum_modern_HDF5_version} CONFIG) ...")
+  find_package(HDF5  ${minimum_modern_HDF5_version}  CONFIG)
+  if (HDF5_FOUND)
+    message("-- Found HDF5_CONFIG=${HDF5_CONFIG}")
+    message("-- Generating Netcdf::all_libs and NetcdfConfig.cmake")
+    tribits_extpkg_create_imported_all_libs_target_and_config_file(
+      HDF5
+      INNER_FIND_PACKAGE_NAME  HDF5
+      IMPORTED_TARGETS_FOR_ALL_LIBS   HDF5::HDF5)
+  endif()
 
-if (HDF5_REQUIRE_FORTRAN)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_fortran)
 endif()
 
-if (TPL_ENABLE_MPI)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} z)
-endif()
+if (NOT TARGET HDF5::all_libs)
 
-if (TPL_ENABLE_Netcdf)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_hl)
-endif()
+  # First, set up the variables for the (backward-compatible) TriBITS way of
+  # finding HDF5.  These are used in case find_package(HDF5 ...) is not called
+  # or does not find HDF5.  Also, these variables need to be non-null in order
+  # to trigger the right behavior in the function
+  # tribits_tpl_find_include_dirs_and_libraries().
 
-#
-# Second, search for HDF5 components (if allowed) using the standard
-# find_package(HDF5 ...).
-#
-tribits_tpl_allow_pre_find_package(HDF5  HDF5_ALLOW_PREFIND)
-if (HDF5_ALLOW_PREFIND)
+  set(REQUIRED_HEADERS hdf5.h)
+  set(REQUIRED_LIBS_NAMES hdf5)
 
-  message("-- Using find_package(HDF5 ...) ...")
-
-  set(HDF5_COMPONENTS C)
   if (HDF5_REQUIRE_FORTRAN)
-    list(APPEND HDF5_COMPONENTS Fortran)
+    set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_fortran)
   endif()
 
   if (TPL_ENABLE_MPI)
-    set(HDF5_PREFER_PARALLEL TRUE)
+    set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} z)
   endif()
 
-  find_package(HDF5 COMPONENTS ${HDF5_COMPONENTS})
-
-  # Make sure that HDF5 is parallel.
-  if (TPL_ENABLE_MPI AND NOT HDF5_IS_PARALLEL)
-    message(FATAL_ERROR "Trilinos is configured for MPI, HDF5 is not.
-    Did CMake find the correct libraries?
-    Try setting HDF5_INCLUDE_DIRS and/or HDF5_LIBRARY_DIRS explicitly.
-    ")
+  if (TPL_ENABLE_Netcdf)
+    set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_hl)
   endif()
 
-  if (HDF5_FOUND)
-    # Tell TriBITS that we found HDF5 and there no need to look any further!
-    set(TPL_HDF5_INCLUDE_DIRS ${HDF5_INCLUDE_DIRS} CACHE PATH
-      "HDF5 include dirs")
-    set(TPL_HDF5_LIBRARIES ${HDF5_LIBRARIES} CACHE FILEPATH
-      "HDF5 libraries")
-    set(TPL_HDF5_LIBRARY_DIRS ${HDF5_LIBRARY_DIRS} CACHE PATH
-      "HDF5 library dirs")
+  #
+  # Second, search for HDF5 components (if allowed) using the standard
+  # find_package(HDF5 ...).
+  #
+  tribits_tpl_allow_pre_find_package(HDF5  HDF5_ALLOW_PREFIND)
+  if (HDF5_ALLOW_PREFIND)
+
+    message("-- Using find_package(HDF5 ...) ...")
+
+    set(HDF5_COMPONENTS C)
+    if (HDF5_REQUIRE_FORTRAN)
+      list(APPEND HDF5_COMPONENTS Fortran)
+    endif()
+
+    if (TPL_ENABLE_MPI)
+      set(HDF5_PREFER_PARALLEL TRUE)
+    endif()
+
+    find_package(HDF5 COMPONENTS ${HDF5_COMPONENTS})
+
+    # Make sure that HDF5 is parallel.
+    if (TPL_ENABLE_MPI AND NOT HDF5_IS_PARALLEL)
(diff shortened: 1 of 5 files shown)


cmake/tribits/common_tpls/FindTPLNetcdf.cmake
```diff
@@ -44,98 +44,158 @@
...
+  set(minimum_modern_netCDF_version 4.7.4)
+  message("-- Netcdf_ALLOW_MODERN=${Netcdf_ALLOW_MODERN}")
+  message("-- Using find_package(netCDF ${minimum_modern_netCDF_version} CONFIG) ...")
+  find_package(netCDF ${minimum_modern_netCDF_version} CONFIG)
+  if (netCDF_FOUND)
+    message("-- Found netCDF_CONFIG=${netCDF_CONFIG}")
+    message("-- Generating Netcdf::all_libs and NetcdfConfig.cmake")
+    # instead of using tribits_extpkg_create_imported_all_libs_target_and_config_file,
+    # we will do what it does ourselves because we need to bring in two
+    # inner find packages (netCDF and hdf5) because netCDF does not properly
```
bartlettroscoe, line 64: This is a really unfortunate chuck of duplication and would be a terrible precedent to set.  Might we consider having the TriBITS Netcdf TPL depend on the TriBITS HDF5 TPL and then have the Netcdf TPL depend on the HDF5 TPL as I tried to outline in https://github.com/TriBITSPub/TriBITS/issues/533#issuecomment-1285905951?  I think that would be very compact and you could essentially accomplish what this large amount of duplicated code is doing.  Is that something you would be willing to consider?  I can add a commit to this branch that attempts that (but I don't have a way to test it with `Netcdf_ALLOW_MODERN=ON`).

bartlettroscoe (commented):
I would like to see if we can accomplish this without mass duplication if what is implemented in TriBITS.  Please see  above comment.

bartlettroscoe:
@ibaned, for an easier way to run the Trilinos GenConfig PR builds locally, try out:

* https://gitlab-ex.sandia.gov/rabartl/run_trilinos_pr_builds

That one can debug these issues without having to wait in the Trilinos PR testing system (and bog down builds for other PRs that are ready to merge).

cmake/tribits/common_tpls/FindTPLNetcdf.cmake
ibaned: Yes, I'm fine with moving in that direction. If you'd be willing to add a commit to this branch that would be super helpful, and I could test the modern option on my end.

bartlettroscoe:
@ibaned, as of the commit [`4f18ccc` (#11175)](https://github.com/trilinos/Trilinos/pull/11175/commits/4f18ccca400f790cc4a10c18dde051f21703e792), this should be close to what you want (and with almost no duplication).  There are some adjustments that may be needed or should be made but this should maintain backward compatibility.

I can't fully test this because I don't have the updated versions of HDF5 and Netcdf installed on any machine that I have access to.  Please give this a test and let me know what happens (or give me access to a machine with the updated HDF5 and Netcdf installed so I can test this myself).

e10harvey:
@bartlettroscoe : I cancelled the jobs running on the problematic nodes reported in https://github.com/trilinos/Trilinos/issues/10999#issuecomment-1288163129. Please try adding a RETEST label.

CC: @DaveBeCoding

ibaned:
Thanks @bartlettroscoe ! I confirmed that the MPI part works for me, let me check the improvements to HDF5 dependence and report back

ibaned:
I've had to make a couple of changes for the HDF5 and Netcdf setup to work in modern mode, I'll commit these here when the build completes

ibaned:
Okay, with the latest changes in this branch this works in my use case with the MODERN option on and custom MPI!

cmake/tribits/core/std_tpls/FindTPLMPI.cmake
```diff
@@ -50,7 +50,9 @@ if(WIN32 AND TPL_ENABLE_MPI)
   global_set(TPL_MPI_LIBRARIES ${MPI_LIBRARIES})
 endif()
 
-tribits_tpl_find_include_dirs_and_libraries(MPI)
+tribits_tpl_find_include_dirs_and_libraries(
+    MPI
+    REQUIRED_HEADERS mpi.h)
```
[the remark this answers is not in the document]
ibaned: Ah sorry, I didn't see the question at the end of this comment, didn't mean to ignore it. I guess philosophically I think that in a full-modern-CMake world compiler wrappers should not be necessary and so I'm trying as hard as possible to do the whole build without them, however I may soon have to give in and implement them.

ibaned:
It looks like the PR build for this still thinks that MPI has REQUIRED_HEADERS even though you reverted that change? is there some mechanism for clearing the CMake cache of the PR build that we need to do?

e10harvey:
> It looks like the PR build for this still thinks that MPI has REQUIRED_HEADERS even though you reverted that change? is there some mechanism for clearing the CMake cache of the PR build that we need to do?

Hi @ibaned: Each PR build configures and builds from scratch. I see that this MPI build configured properly: https://trilinos-cdash.sandia.gov/build/545791/configure. Please try adding a RETEST...

e10harvey:
> > It looks like the PR build for this still thinks that MPI has REQUIRED_HEADERS even though you reverted that change? is there some mechanism for clearing the CMake cache of the PR build that we need to do?
> 
> Hi @ibaned: Each PR build configures and builds from scratch. I see that this MPI build configured properly: https://trilinos-cdash.sandia.gov/build/545791/configure. Please try adding a RETEST...

@ibaned: Note that I sorted by start time in [CDash Test Results for PR# 11175](https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-11175-test&field2=buildstarttime&compare2=84&value2=NOW) to find the most recent CDash results.

bartlettroscoe:
>  Please try adding a RETEST...

@ibaned and @e10harvey, I took off `AT: RETEST`, that is not going to work.  See https://github.com/trilinos/Trilinos/issues/10999#issuecomment-1288163129.

@e10harvey, how do these bad nodes (or at least nodes where the SEMS modules are broken) keep getting into the pool of Jenkins PR build machines?

cmake/tribits/core/std_tpls/FindTPLMPI.cmake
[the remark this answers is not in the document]
bartlettroscoe: @ibaned, you can use raw compilers with MPI.  See "Configuring to build using raw compilers and flags/libraries:" at:

* https://docs.trilinos.org/files/TrilinosBuildReference.html#configuring-with-mpi-support

This has been supported and used with Trilinos for at least 14 years.

bartlettroscoe:
> It should work now. I've added both of you to the related help desk ticket.

@e10harvey, would it be possible to comment in #10999 so other developers can see this as well?  The HelpDesk issues are closed by default to **everyone**.

cmake/tribits/common_tpls/FindTPLHDF5.cmake
```diff
@@ -4,17 +4,18 @@
 
 if (Netcdf_ALLOW_MODERN)
 
-  #set(minimum_modern_HDF5_version 4.7.4)
+  set(minimum_modern_HDF5_version 1.13.2)
```
bartlettroscoe, line 7: Thumbs up

cmake/tribits/common_tpls/FindTPLNetcdf.cmake
```diff
@@ -43,7 +43,7 @@ endif()
 
 set(Netcdf_ALLOW_MODERN FALSE CACHE BOOL "Allow finding Netcdf as a modern CMake config file with exported targets (and only this way)")
 
-if (Netcdf_ALLOW_MODERN  AND  TARGET HDF5::HDF5)
+if (Netcdf_ALLOW_MODERN)
```
bartlettroscoe, line 46: The danger of not checking to see if the modern HDF5 was found is that you can end up with a situation where the modern find_package(HDF5 CONFIG ...) commands fails and this find_package(netCDF CONFIG ...) command runs which will result in a configure-time error because it is missing the necessary HDF5 targets.

We need some type of check that the correct HDF5 module was found.  How can we do that?

bartlettroscoe (commented):
@ibaned, I think the if() logic in FindTPLNetcdf.cmake needs a tweak.  See comment above.

cmake/tribits/common_tpls/FindTPLNetcdf.cmake
bartlettroscoe: Sure, I agree.  Let's define a var called something like `HDF5_FOUND_MODERN_CONFIG_FILE=TRUE`.

cmake/tribits/common_tpls/FindTPLHDF5.cmake
```diff
@@ -16,10 +18,13 @@ if (Netcdf_ALLOW_MODERN)
       HDF5
       INNER_FIND_PACKAGE_NAME  HDF5
       IMPORTED_TARGETS_FOR_ALL_LIBS   ${HDF5_EXPORT_LIBRARIES})
+    set(INTERNAL_IS_MODERN TRUE)
   endif()
 
 endif()
 
+set(HDF5_FOUND_MODERN_CONFIG_FILE ${INTERNAL_IS_MODERN} CACHE BOOL "True if HDF5 was found by the modern method" FORCE)
```
bartlettroscoe, line 26: The only suggestion I would have to be to change this to `INTERNAL` as:

```
set(HDF5_FOUND_MODERN_CONFIG_FILE ${INTERNAL_IS_MODERN} CACHE INTERNAL "True if HDF5 was found by the modern method")
```

because we don't want this showing up as an option that a user might be able to change (like with the CMake GUI or with ccmake).

cmake/tribits/common_tpls/FindTPLHDF5.cmake
```diff
@@ -2,6 +2,8 @@
 # See associated tribits/Copyright.txt file for copyright and license! #
 ########################################################################
 
+set(INTERNAL_IS_MODERN FALSE)
```
bartlettroscoe, line 5: Might want to namespace this as `HDF5_INTERNAL_IS_MODERN` because these modules are included in base project scope.

bartlettroscoe (approved):
Just some minor suggestions but I think this would be okay to merge.  But if you wanted to make these suggested changes, that is okay too  :-)

ibaned:
@e10harvey a `vortex3` runner snuck in there and did this:
```
Error: error in ptssup_mkcltsock_afunix()
10-26-2022 00:23:23:819 121619 main: Error initializing RM connection. Exiting.
jsrun return value: 1
```

bartlettroscoe:
> @e10harvey a `vortex3` runner snuck in there and did this:
> 
> ```
> Error: error in ptssup_mkcltsock_afunix()
> 10-26-2022 00:23:23:819 121619 main: Error initializing RM connection. Exiting.
> jsrun return value: 1
> ```

@ibaned, this is a **long** known random error on this build.  See:

* #6861
* #7122
* #11109

bartlettroscoe (approved):
Thanks for those final improvements!

ibaned:
@bartlettroscoe I've made the next round of improvements you pointed out. Once the CI on this PR gets figured out, should I open a PR on TriBITS with these changes?

bartlettroscoe:
> @bartlettroscoe I've made the next round of improvements you pointed out. Once the CI on this PR gets figured out, should I open a PR on TriBITS with these changes?

@ibaned, I can take care of that.   (It is just careful usage of `git format-patch` and `git am`.)

cmake/tribits/common_tpls/FindTPLHDF5.cmake
```diff
@@ -2,78 +2,103 @@
...
 
-if (TPL_ENABLE_MPI)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} z)
-endif()
+if (Netcdf_ALLOW_MODERN)
+
+  set(minimum_modern_HDF5_version 1.13.2)
+  print_var(Netcdf_ALLOW_MODERN)
+  message("-- Using find_package(HDF5 ${minimum_modern_HDF5_version} CONFIG) ...")
+  find_package(HDF5  ${minimum_modern_HDF5_version}  CONFIG)
```
bartlettroscoe, line 12: @ibaned, @gsjaardema pointed out in https://github.com/sandialabs/seacas/pull/336#pullrequestreview-1157178993 that this should really be a minimum version, not a specific version of HDF5.  I don't know if `HDF5Config.cmake` supports version ranges, but if it does, I think this would be:

```
    find_package(HDF5 ${minimum_modern_HDF5_version}... CONFIG)
```

cmake/tribits/common_tpls/FindTPLNetcdf.cmake
```diff
@@ -37,105 +37,130 @@
...
-    set(TPL_Netcdf_LIBRARY_DIRS ${_hdf5_LIBRARY_SEARCH_DIRS} CACHE PATH
-      "${DOCSTR} library files")
-    set(TPL_Netcdf_LIBRARIES ${NetCDF_LIBRARIES} CACHE PATH
-      "List of semi-colon separated library names (not 'lib' or extension).")
-    set(TPL_Netcdf_INCLUDE_DIRS ${NetCDF_INCLUDE_DIRS} CACHE PATH
-      "${DOCSTR} header files.")
+  set(minimum_modern_netCDF_version 4.7.4)
+  print_var(Netcdf_ALLOW_MODERN)
+  message("-- Using find_package(netCDF ${minimum_modern_netCDF_version} CONFIG) ...")
+  find_package(netCDF ${minimum_modern_netCDF_version} CONFIG)
```
bartlettroscoe, line 51: @ibaned, as per [above](https://github.com/trilinos/Trilinos/pull/11175/files#r1006168735), if `netCDFConfig.cmake` supports version ranges, I think this should be:

```
    find_package(netCDF ${minimum_modern_netCDF_version}... CONFIG)
```

no?

bartlettroscoe:
FYI: The matching changes to the TriBITS repo are in:

* https://github.com/TriBITSPub/TriBITS/pull/535

I will merge that PR as soon as this PR merges (fingers crossed on PR testing).

cmake/tribits/common_tpls/FindTPLHDF5.cmake
ibaned: I think it was concluded in other discussion that the code as is does request this version or higher. That is certainly my intention anyway, that this work for later versions.

cmake/tribits/common_tpls/FindTPLNetcdf.cmake
ibaned: As above, it was concluded that the current code already means that, but if it doesn't we will change it to mean that

PR #11177, opened 2022-10-21
MueLu nightlies: Attempt to not pull extra repos

cgcgcg:
@trilinos/muelu 

## Motivation
Do not try to pull extra repos (MOOCHO, preCopyrightTrilinos, etc) since that fails anyway.

@csiefer2 I did not change the lightsaber config, since that might rely on Avatar being pulled.


jhux2 (approved):
lgtm

PR #11182, opened 2022-10-22
Add include guards for *_def.hpp file (#10507)

bartlettroscoe:
When this file was converted from a *.cpp file to a *_def.hpp file when Stratimikos::LinearSolverBuilder was templated in PR #10507, include guards did not get added to this file.


bartlettroscoe:
@jennloe, this adds the include guards that we noticed were missing.

cgcgcg (approved):
LGTM

PR #11183, opened 2022-10-22
Tacho : fix issue with LDLt

iyamazaki:
@trilinos/shylu 

## Motivation

This PR aims to fix issues with LDLt factorization

## Additional Information

Issues occurs when the same 2-by-2 pivots are picked for the consecutive pairs of rows.

### DIFF
--- a/packages/shylu/shylu_node/tacho/src/impl/Tacho_NumericTools_LevelSet.hpp
+++ b/packages/shylu/shylu_node/tacho/src/impl/Tacho_NumericTools_LevelSet.hpp
@@ -1465,6 +1465,9 @@ public:
       constexpr bool copy_to_l_buf(false);
       _info.copySparseToSuperpanels(copy_to_l_buf, _ap, _aj, _ax, _perm, _peri);
     }
+    if (_nstreams > 1) {
+      exec_space().fence(); // wait for copy
+    }
     stat.t_copy = timer.seconds();
 
     stat_level.n_kernel_launching = 0;
@@ -2706,6 +2709,9 @@ public:
       constexpr bool copy_to_l_buf(false);
       _info.copySparseToSuperpanels(copy_to_l_buf, _ap, _aj, _ax, _perm, _peri);
     }
+    if (_nstreams > 1) {
+      exec_space().fence(); // wait for copy
+    }
     stat.t_copy = timer.seconds();
 
     stat_level.n_kernel_launching = 0;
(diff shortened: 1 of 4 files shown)


crdohrm (approved):

packages/shylu/shylu_node/tacho/src/impl/Tacho_SupernodeInfo.hpp
```diff
@@ -335,7 +335,7 @@ template <typename ValueType, typename DeviceType> struct SupernodeInfo {
               {
                 const ordinal_type lbeg = ap(i), lend = ap(i + 1);
                 ordinal_type *first = &aj(lbeg);
```
brian-kelley, line 337: I guess the old code caused Kokkos bounds check errors? A cleaner way to get offset pointers without dereferencing might be:
```
ordinal_type *first = aj.data() + lbeg;
ordinal_type *last = aj.data() + lend;
```

packages/shylu/shylu_node/tacho/src/impl/Tacho_NumericTools_LevelSet.hpp
```diff
@@ -2719,15 +2725,17 @@ class NumericToolsLevelSet : public NumericToolsBase<ValueType, DeviceType> {
 #if (11000 > CUDA_VERSION)
```
brian-kelley, line 2725: Hmmm... Having ``CUDA_VERSION`` defined doesn't necessarily mean that ``exec_space`` is ``Kokkos::Cuda``. Unless there is already a compile-time check for this somewhere else...

packages/shylu/shylu_node/tacho/src/impl/Tacho_NumericTools_LevelSet.hpp
```diff
@@ -2719,15 +2725,17 @@ class NumericToolsLevelSet : public NumericToolsBase<ValueType, DeviceType> {
...
+      const ordinal_type team_size_update[2] = {16, 8},  vector_size_update[2] = {16, 32};
 #else
       /// cuda 11.1 and higher
       const ordinal_type team_size_factor[2] = {64, 64}, vector_size_factor[2] = {8, 4};
+      const ordinal_type team_size_update[2] = {16, 8},  vector_size_update[2] = {32, 32};
 #endif
 #else
       /// not cuda
       const ordinal_type team_size_factor[2] = {64, 64}, vector_size_factor[2] = {8, 4};
+      const ordinal_type team_size_update[2] = {16, 8},  vector_size_update[2] = {32, 32};
```
brian-kelley, line 2737: This would be safer for non-cuda devices where vector length 32 might be too many:
```
      cont int vmax = exec_space::vector_length_max();
      const ordinal_type team_size_update[2] = {16, 8},  vector_size_update[2] = {std::min(vmax, 32), std::min(vmax, 32)};
```
Could do the same for vector_size_factor as well.

packages/shylu/shylu_node/tacho/src/impl/Tacho_NumericTools_LevelSet.hpp
brian-kelley: And hardcoding team/vector sizes isn't necessarily going to be portable, but if this works on the supported hardware types then it's OK.

Note that a team/vector size that works in one build might not work in another build with debug, bounds checking etc. because those things can use more registers. If you want to guarantee that it's safe in all cases, you could do this (and then the analogous thing for policy_update)
```
//Create an initial dummy policy with team size 1, just to query the max team size
policy_factor = team_policy_factor(pcnt, 1, vector_size_factor[idx]);
int factor_teamsize_max = policy_factor.team_size_max(functor, Kokkos::ParallelForTag());
policy_factor = team_policy_factor(pcnt, std::min(factor_teamsize_max, team_size_factor[idx]), vector_size_factor[idx]);
```

packages/shylu/shylu_node/tacho/src/impl/Tacho_NumericTools_LevelSet.hpp
```diff
@@ -1491,11 +1494,13 @@ class NumericToolsLevelSet : public NumericToolsBase<ValueType, DeviceType> {
...
 #endif
 
-        functor_type functor(_info, _factorize_mode, _level_sids, _buf);
-
         team_policy_factor policy_factor(1, 1, 1);
         team_policy_update policy_update(1, 1, 1);
+        functor_type functor(_info, _factorize_mode, _level_sids, _buf);
 
+        // get max vector size
+        const ordinal_type factor_vmax = policy_factor.vector_length_max();
```
brian-kelley, line 1502: You could also just have one "vmax" - it will always be the same for both policy tags

brian-kelley (dismissed):

iyamazaki:
Thank you for all the comments, @brian-kelley!!

brian-kelley (approved):

Issue #7741, opened 2020-07-28
Sacado: Memory leak with nested reverse-forward type.


dougshidong:
## Bug Report
@trilinos/\<Sacado\>

### Description
Using the nested type `Sacado::Rad::ADvar< Sacado::Fad::DFad< double > >` will result in a some lost memory if `Gradcomp()` is called more than once using this nested type.

Note that the reverse mode templated with double `Sacado::Rad::ADvar< double >` does not have a memory leak, nor does `Sacado::Fad::DFad<double>`. It is specifially some memory within `DFad` that is not freed when used in conjunction with `ADvar`.

Not sure if there is a way to flush the memory. I have tried ADcontext::free_all() without success.

### Steps to Reproduce
1. Modify trad_dfad_example.cpp to perform the loop twice.
1. Run with valgrind --leak-check-full --track-origins=yes

or

1. Modify trad_dfad_example.cpp to perform the loop 999999999 times.
1. Watch RAM in CPU activity slowly blow up.

I have tried looping with trad_example.cpp and dfad_dfad_example.cpp and those do not cause a memory leak. It really is specifically `Sacado::Rad::ADvar<Sacado::Fad::DFad<double>>`.

```cpp
// $Id$ 
// $Source$ 
// @HEADER
// ***********************************************************************
// 
//                           Sacado Package
//                 Copyright (2006) Sandia Corporation
// 
// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
// the U.S. Government retains certain rights in this software.
// 
// This library is free software; you can redistribute it and/or modify
// it under the terms of the GNU Lesser General Public License as
// published by the Free Software Foundation; either version 2.1 of the
// License, or (at your option) any later version.
//  
// This library is distributed in the hope that it will be useful, but
// WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
// Lesser General Public License for more details.
//  
// You should have received a copy of the GNU Lesser General Public
// License along with this library; if not, write to the Free Software
// Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301
// USA
// Questions? Contact David M. Gay (dmgay@sandia.gov) or Eric T. Phipps
// (etphipp@sandia.gov).
// 
// ***********************************************************************
// @HEADER

// trad_dfad_example
//
//  usage: 
//     trad_dfad_example
//
[...]

PR #11190, opened 2022-10-26
Sacado:  Fix memory leaks in Rad when using DFad nested scalar type.

etphipp:
Rad does its own memory management, but doesn't know anything about it when its nested scalar type is also dynamically allocated, which was causing massive memory leaks when using, e.g., Rad< DFad<double> >.  The only way to fix this without a major overhaul in Rad is to ask the user to call a new function to zero out all of Rad's internally stored data, which as a side-effect will deallocate DFad derivative arrays.  Also added this new function to the trad-dfad example.

For issue #7741.

### DIFF
--- a/packages/sacado/example/trad_dfad_example.cpp
+++ b/packages/sacado/example/trad_dfad_example.cpp
@@ -138,6 +138,12 @@ int main(int argc, char **argv)
 	    << std::setw(w) << d2rdbda_ad << " (AD) Error = " << std::setw(w) 
 	    << d2rdadb - d2rdbda_ad << std::endl;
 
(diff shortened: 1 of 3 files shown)


dougshidong:
Should `free_all()` then call `zero_out()`?

etphipp:
> Should `free_all()` then call `zero_out()`?

I considered that, but chose not to do that for a few reasons:
* It has no trivial run-time cost and isn't necessary for most scalar types, only scalar types like DFad that involve dynamic memory allocation;
* `zero_out()` should be called if the gradient computation is inside a loop (whereas `free_all()` doesn't necessarily need to be called), so the user needs to know about it anyway and depending on how the user implements their loop, `freee_all()` may not need to call `zero_out()`.

PR #11191, opened 2022-10-26
Frosch : add an option to call Zoltan2 before local ILU

iyamazaki:
@trilinos/shylu 

## Motivation

This PR adds an option to call Zoltan2 before local ILU


iyamazaki:
Hi, @searhein.. No, this is just for ILU. Sparse direct solvers usually come with different options for the fill-reducing ordering but if such options are needed, just let me know.

searhein:
It could be interesting to have a node-level DD method for the local solve. Therefore, I though one could implement the METIS in a general way such that all node-level solvers can use it?

brian-kelley (approved):
Ifpack2 changes look good to me.

iyamazaki:
Thank you, @brian-kelley!!

iyamazaki:
@searhein. For this PR, I put Zoltan2 in Ifpack2 solver interface for local matrix ordering, not for partitioning. Maybe, we could discuss your use case?

searhein:
@iyamazaki Oh, I see. I just had a very quick look. So I mixed this up. Thanks!

searhein (approved):

Issue #11179, opened 2022-10-21
Trilinos develop SEACASNemesis_ne_ctest_wrap compilation error on Crusher


PaulMullowney:
Hash : c0b25c6908353f014eb10134cc4f68029fa23068

Q:  Is there a way to avoid compilation of SEACASNemesis_ne_ctest_wrap with a CMake flag?

Here's the compilation error:
````
     1087    [ 42%] Linking CXX executable SEACASNemesis_ne_ctest_wrap.exe
     1088    cd /gpfs/alpine/cfd116/proj-shared/mullowne/spack-manager-hip/environments/nalu-master-trilinos-develop_rocm-5.2.0/trilinos/spack-build-uhufbdx/packages/seacas/libraries/nemesis && /autofs/nccs-svm1_sw/crusher/spack-envs/base/opt/cray-sles15-zen3/clang-14.0.0-rocm5.1.0/cmake-3.22.2-iv5tnnzbqsgz6ebgrysktc55b5unqvjy/bin/cmake -E cma
             ke_link_script CMakeFiles/SEACASNemesis_ne_ctest_wrap.dir/link.txt --verbose=1
     1089    /opt/cray/pe/mpich/8.1.16/ofi/amd/5.0/bin/mpicxx -DSTK_NO_BOOST_STACKTRACE -DUSE_STK_SIMD_NONE -march=znver3 -mtune=znver3 -fgpu-rdc -DDESUL_HIP_RDC  -O3 -DNDEBUG CMakeFiles/SEACASNemesis_ne_ctest_wrap.dir/ne_ctest_wrap.c.o -o SEACASNemesis_ne_ctest_wrap.exe   -L/opt/cray/pe/mpich/8.1.16/ofi/cray/10.0/lib  -L/opt/cray/pe/mpich/8.1
             .16/gtl/lib  -L/opt/cray/pe/cce/14.0.2/cce/x86_64/lib  -L/opt/gcc/10.3.0/snos/lib64  -L/opt/cray/pe/gcc/10.3.0/snos/lib64  -L/opt/cray/pe/gcc/10.3.0/snos/lib/gcc/x86_64-suse-linux/10.3.0  -L/opt/cray/pe/cce/14.0.2/cce-clang/x86_64/lib  -Wl,-rpath,/gpfs/alpine/cfd116/proj-shared/mullowne/spack-manager-hip/spack/opt/spack/cray-sles1
[...]

PR #11210, opened 2022-10-28
SEACAS: Fix some cmake issues

gsjaardema:
@trilinos/seacas

## Motivation

This cleans up some CMake issues in the seacas project that were causing problems when an external application was using or building Trilinos.

Also includes some cleanup found while looking at CMake files in general while trying to fix the applications issue.


gsjaardema:
Fixes #11179

packages/seacas/applications/exo_format/cmake/Dependencies.cmake
```diff
@@ -1,9 +1,3 @@
-if(CMAKE_PROJECT_NAME STREQUAL "SEACAS")
-  TRIBITS_PACKAGE_DEFINE_DEPENDENCIES(
-    LIB_REQUIRED_PACKAGES Exodus
-  )
```
bartlettroscoe, line 4: This should not be impacting the Trilinos project.  Not sure why this had to be pulled out.

packages/seacas/libraries/nemesis/CMakeLists.txt
```diff
@@ -28,17 +28,13 @@ TRIBITS_ADD_LIBRARY(
   SOURCES ${SOURCES}
   )
 
-IF (${PACKAGE_NAME}_ENABLE_TESTS)
-  TRIBITS_ADD_TEST(
+IF (${CMAKE_PROJECT_NAME}_ENABLE_TESTS)
```
bartlettroscoe, line 31: Why was this changed?  The correct logic **is** `if (${PACKAGE_NAME}_ENABLE_TESTS)`.

This change violates the documented behavior of TriBITS projects (see [<Project>_ENABLE_TESTS only enables explicitly enabled package tests](https://tribitspub.github.io/TriBITS/users_guide/index.html#project-enable-tests-only-enables-explicitly-enabled-package-tests)) .  For example, if someone configures Trilinos with:

```
-D Trilinos_ENABLE_Panzer=ON \
-D Trilinos_ENABLE_TESTS=ON \
```

they are supposed to only get Panzer tests and examples turned on.  But with checking only `${CMAKE_PROJECT_NAME}_ENABLE_TESTS`, these tests would be enabled by surprise to everyone.

However, note that this if statement is meaningless because the the logic to enable/disable tests is baked into the functions `tribits_add_test()` and `tribits_add_executable_and_test()`.  See:

* https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-add-test

which says:

> The tests are only added if tests are enabled for the package (i.e. [`${PACKAGE_NAME}_ENABLE_TESTS`](https://tribitspub.github.io/TriBITS/users_guide/index.html#package-name-enable-tests))

(I need to change the last part of that sentence because it is no longer true that setting `${PARENT_PACKAGE_NAME}_ENABLE_TESTS=TRUE` will automatically enable the tests in all enabled subpackages.)

packages/seacas/scripts/cmake/Dependencies.cmake
```diff
@@ -1,9 +1,3 @@
-if(CMAKE_PROJECT_NAME STREQUAL "SEACAS")
-TRIBITS_PACKAGE_DEFINE_DEPENDENCIES(
-  LIB_REQUIRED_PACKAGES Exodus)
-)
```
bartlettroscoe, line 4: Again, why was this impacting the Trilinos build?

bartlettroscoe (changes requested):

bartlettroscoe:
@gsjaardema, I clicked "request changes" a bit too early but the only thing that really needs to change is to put back the if conditional the way it was [above](https://github.com/trilinos/Trilinos/pull/11210#discussion_r1008699821).

packages/seacas/applications/exo_format/cmake/Dependencies.cmake
gsjaardema: Found while looking for actual issue.  Did not need the if.  Dependency is always there.

packages/seacas/scripts/cmake/Dependencies.cmake
gsjaardema: Similar, wasn't impacting build, but was overly complicated so cleaned up.  I have updated main header comment on this PR to indicate that not all changes were directly related to Trilinos build.

packages/seacas/libraries/nemesis/CMakeLists.txt
gsjaardema: I have removed the `if` block to let `TRIBITS_ADD_EXECUTABLE_AND_TEST` take care of enable/disable.

Issue #10742, opened 2022-07-11
Framework: No complex or float PR or 'master' merge builds needed to support ASC Sierra Code


bartlettroscoe:
@trilinos/framework, @vbrunini, @jclause, @jhux2 

## Description

The ASC Sierra Code requires support for `complex<double>` and `float` with the configure arguments:

```
-D Trilinos_ENABLE_COMPLEX_DOUBLE=ON \
-D Trilinos_ENABLE_FLOAT=ON \
```

(see #10635 and #10720.)  Yet there appear to be no PR or 'master' promotion builds that enable either of these.

The documentation at:

* https://github.com/trilinos/Trilinos/wiki/Pull-Request-Testing-Interface

claims there are two PR builds that enable complex:

* `rhel7_cuda-10.1.105-gnu-7.2.0-spmpi-rolling_release_static_Volta70_Power9_no-asan_complex_fpic_mpi_pt_no-rdc_no-uvm_deprecated-on_no-package-enables`
* `rhel7_cuda-10.1.105-gnu-7.2.0-spmpi-rolling_release_static_Volta70_Power9_no-asan_complex_fpic_mpi_pt_no-rdc_uvm_deprecated-on_no-package-enables`

and two 'master' promotion builds that enable complex:

* `rhel7_cuda-10.1.105-gnu-7.2.0-spmpi-rolling_release_static_Volta70_Power9_no-asan_complex_fpic_mpi_pt_no-rdc_no-uvm_no-package-enables
* `rhel7_cuda-10.1.105-gnu-7.2.0-spmpi-rolling_release_static_Volta70_Power9_no-asan_complex_fpic_mpi_pt_no-rdc_uvm_deprecated-on_no-package-enables`

But those builds don't seem to be running from looking at recent PR testing iterations like https://github.com/trilinos/Trilinos/pull/10720#issuecomment-1179472450.

This is a problem because important customers like Sierra require building with `-D Trilinos_ENABLE_COMPLEX_DOUBLE=ON` (see #10635).

Also, there does not not appear to even be support for the GenConfig system for setting `-DTrilinos_ENABLE_FLOAT=ON` which is required by important customers like Sierra.    This is (not) shown by:

```
$ find packages/framework/ -type f -exec grep -nH ENABLE_FLOAT {} \; | wc -l
0
```

But there are lots of references to ENABLE_COMPLEX_DOUBLE:

```
[...]

Issue #535, opened 2016-08-02
Thyra/Ifpack2: segfault


nschloe:
When using Thyra with Ifpack2, `initializePrec` segfaults.

MWE:
```c++
#include <Stratimikos_DefaultLinearSolverBuilder.hpp>
#include <Teuchos_DefaultComm.hpp>
#include <Tpetra_CrsMatrix.hpp>
#include <Thyra_Ifpack2PreconditionerFactory.hpp>
#include <Thyra_TpetraThyraWrappers.hpp>

Teuchos::RCP<const Tpetra::CrsMatrix<double,int,int>>
create_matrix(
    const Teuchos::RCP<const Teuchos::Comm<int>> & comm
    )
{
  const Tpetra::global_size_t numGlobalElements = 100;

  const int indexBase = 0;
  auto map = Teuchos::rcp(new Tpetra::Map<int,int>(
        numGlobalElements,
        indexBase,
        comm
        ));

  const size_t numMyElements = map->getNodeNumElements();
  auto myGlobalElements = map->getNodeElementList();

  auto A = Teuchos::rcp(new Tpetra::CrsMatrix<double,int,int>(map, 1));

  for (size_t i = 0; i < numMyElements; i++) {
    A->insertGlobalValues(
        myGlobalElements[i],
        Teuchos::tuple(myGlobalElements[i]),
        Teuchos::tuple(1.0 / (myGlobalElements[i] + 1))
        );
  }

  A->fillComplete();
  return A;
}

int main(int argc, char *argv[]) {
  Teuchos::GlobalMPISession session(&argc, &argv, NULL);
  auto out = Teuchos::VerboseObjectBase::getDefaultOStream();

  const auto comm = Teuchos::DefaultComm<int>::getComm();

  const auto A = create_matrix(comm);

  auto b = Tpetra::Vector<double,int,int>(A->getRangeMap());
  b.putScalar(1.0);

  auto x = Tpetra::Vector<double,int,int>(A->getDomainMap());
  x.putScalar(0.0);

  Stratimikos::DefaultLinearSolverBuilder builder;
  auto p = Teuchos::rcp(new Teuchos::ParameterList());
  builder.setParameterList(p);

  auto lowsFactory = builder.createLinearSolveStrategy("");
  lowsFactory->setVerbLevel(Teuchos::VERB_LOW);

  const Tpetra::Operator<double,int,int> & opA = *A;
[...]

PR #11222, opened 2022-11-02
Stratimikos belos prec

jennloe:
@trilinos/stratimikos @trilinos/belos 

## Motivation
This PR provides an adapter so that Belos solvers (with Tpetra linear algebra) can be used as a preconditioner for other Belos solvers generated with Stratimikos. We also allow the Belos preconditioners to be generated in single precision. This allows combinations such as:
- Belos solver + Belos GMRES polynomial preconditioner
- FGMRES with Belos solver as preconditioner
- Mixed precision solvers such as GMRES-IR.  (Use single precision GMRES to precondition Fixed-Point iteration.)

@srajama1 @iyamazaki This finishes out my commitments for the multiprecision project. 

## Testing

There are three new tests as follows:
1. Belos GMRES + Belos GMRES polynomial preconditioner (all double precision)
2. Belos GMREs + single precision polynomial preconditioner
3. GMRES-IR (Fixed point iteration + single precision GMRES preconditioner.)


packages/stratimikos/adapters/belos/tpetra/README.md
```diff
@@ -0,0 +1,36 @@
+# Belos Tpetra Preconditioning in Stratimikos
+
+The adapters in this directory allow a (unpreconditioned) Belos solver using Tpetra linear algebra to be called from a Stratimikos input deck as a preconditioner to another Belos solver.
```
cgcgcg, line 3: What about the preconditioned case?

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,367 @@
...
+#include "Thyra_DefaultPreconditioner.hpp"
+#include "Thyra_TpetraLinearOp.hpp"
+#include "Thyra_TpetraThyraWrappers.hpp"
+
+#include "BelosTpetraOperator.hpp"
+#include "BelosConfigDefs.hpp"
+#include "BelosLinearProblem.hpp"
+#include "BelosTpetraAdapter.hpp"
+
+#include "Tpetra_RowMatrix.hpp"
```
cgcgcg, line 56: is that needed?

packages/stratimikos/test/CMakeLists.txt
```diff
@@ -187,40 +187,70 @@ IF (${PROJECT_NAME}_ENABLE_Tpetra AND ${PACKAGE_NAME}_ENABLE_ThyraTpetraAdapters
...
+
+  TRIBITS_ADD_TEST(
+    test_tpetra_belos
+    NAME test_tpetra_belos_gmres_ir
+    ARGS
+      "--input-file=stratimikos_GMRES_IR_ParameterList.xml --show-timer-summary"
+    COMM serial mpi
+    NUM_MPI_PROCS 4
+    )
+
```
iyamazaki, line 239: Do you need something like `Trilinos_ENABLE_FLOAT` to run these tests?

bartlettroscoe: That seems to be the case.  And you see these tests failing for builds that don't have float turned on [above](https://github.com/trilinos/Trilinos/pull/11222#issuecomment-1301375425).  (It seems that there are no PR builds that have float enabled so PR testing will not even run these tests if you guard them with `EXCLUDE_IF_NOT_TRUE ${PROJECT_NAME}_ENABLE_FLOAT` (see [`tribits_add_test()`](https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-add-test)).)

iyamazaki:
Thank you for pushing these to Trilinos, @jennloe. These look very nice!! I forgot if the GMRES-IR in Stratimikos (double-precision fixed-point with single-precision GMRES) is implemented same as your implementation with Kokkos-Kernels, or is there any numerical difference (e.g., the initial vector to the GMRES iteration is normalized in float)?

packages/stratimikos/test/CMakeLists.txt
bartlettroscoe: > Do any of the nightly or master merge test builds enable float?

No, not a one.  See:

* #10742 

This means this code you are writing is not going to be tested in any automated tests that are run by the Trilinos Framework team.  The person to discuss that with is @ccober6, who is the PO for the Framework team.

packages/stratimikos/adapters/belos/tpetra/README.md
jennloe: We are discussing this in person... need to meet with the rest of the multiprecision team.

packages/stratimikos/test/CMakeLists.txt
jennloe: Will discuss this further with the multiprecision team. For right now, @cgcgcg says we can have this tested in the MueLu nightlies. (@csiefer2 )

iyamazaki: Shouldn't we at least guard these tests with `*_ENABLE_FLOAT`?

bartlettroscoe: > Shouldn't we at least guard these tests with `*_ENABLE_FLOAT`?

That is the minimum requirement.

I think you can do this by adding the argument:

```
EXCLUDE_IF_NOT_TRUE  ${PROJECT_NAME}_ENABLE_FLOAT  Tpetra_INST_COMPLEX_FLOAT
```

to the above `TRIBITS_ADD_TEST()` calls.  Using the `EXCLUDE_IF_NOT_TRUE` argument results in a line being printed to CMake output that lists the function by describes why it was not added.  (When you use an `if ()` statement, you don't see anything for a test in the output.)

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
jennloe: Apparently not.

jennloe:
Thanks @iyamazaki ! 
We need to study the performance of this for sure.  At first glance, it is not so fast as our Kokkos Kernels version.

I'm not as sure about numerical differences... Yes, thinking about it, I think the GMRES starting vector is downcast to float and then is normalized in float. But I think that was the same in my Kokkos GMRES-IR driver.  See here: https://github.com/jennloe/Trilinos/blob/KokkosBelos/packages/belos/kokkos/src/GmresIRKokkos.cpp 

Any other numerical differences we should be watching for?

iyamazaki:
Thank you, @jennloe!! Sorry, I think you are right. In the mixed-precision iterative refinement, we solve for the correction in the lower precision (so GMRES scales the starting vector). I've encounter numerical issues in some cases, and using the working precision to normalize the starting vector solved the particular issues, but that may be a hack. :)

jennloe:
@iyamazaki Yes, unfortunately scaling the starting vec in double precision isn't exactly an option with the current Belos implementations.  Maybe we should keep this in mind for future modifications.

jennloe:
Note, the last commit pushed fixes corresponding to #535, just to link the issues.
The test for that issue had somehow been disabled long before this PR, and half precision additions to the IfPack2 Thyra adapter had broken the test in the meantime.  The test has been re-enabled, and a simpler workaround to address the issue is in place in both the IfPack2 and corresponding Belos-Prec adapters.

cgcgcg (approved):
LGTM. Outstanding issues are perfomance testing and preconditioning of the inner solver.

bartlettroscoe:
This is a large patch so I will try to review it commit-by-commit.

cgcgcg (dismissed):
Why is there an empty CMakeList.txt?

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,369 @@
...
+    Teuchos::RCP<BelosTpLinProb> belosLinProb = Teuchos::RCP<BelosTpLinProb>(new BelosTpLinProb());
+    belosLinProb->setOperator(tpetraFwdOp);
+    Teuchos::RCP<TpetraLinOp> belosOpRCP = Teuchos::RCP<TpetraLinOp>(new BelosTpOp(belosLinProb, packageParamList, solverType, true));
+    thyraPrecOp = Thyra::createLinearOp(belosOpRCP);
+  }
+  defaultPrec->initializeUnspecified(thyraPrecOp);
+
+  totalTimer.stop();
+  if (Teuchos::nonnull(out) && Teuchos::includesVerbLevel(verbLevel, Teuchos::VERB_LOW)) {
+    *out << "\nTotal time in Thyra::Ifpack2PreconditionerFactory::initializePrec(...) = " << totalTimer.totalElapsedTime() << " sec\n";
```
cgcgcg, line 247: Ifpack2 -> Belos

packages/stratimikos/adapters/CMakeLists.txt
```diff
@@ -36,4 +36,7 @@ ENDIF()
 ASSERT_DEFINED(${PACKAGE_NAME}_ENABLE_Belos)
 IF (${PACKAGE_NAME}_ENABLE_Belos)
   ADD_SUBDIRECTORY(belos)
+  IF (${PACKAGE_NAME}_ENABLE_ThyraTpetraAdapters)
+    ADD_SuBDIRECTORY(belos/tpetra)
```
GrahamBenHarper, line 40: I know CMake is resilient to upper/lowercase, but it definitely bothers my OCD side :stuck_out_tongue:

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,369 @@
...
+  if (paramList_.is_null ()) {
+    innerParamList = Teuchos::rcp(new Teuchos::ParameterList(*getValidParameters()));
+  }
+  else {
+    innerParamList = paramList_;
+  }
+
+  bool useHalfPrecision = false;
+  if (innerParamList->isParameter("half precision"))
+    useHalfPrecision = innerParamList->get<bool>("half precision");
```
bartlettroscoe, line 195: Nitpick: Why not be consistent with the line below this one and use:

```
useHalfPrecision = getParameter<bool>(*innerParamList, "half precision");
```

?

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,369 @@
...
+
+  const std::string solverType = Teuchos::getParameter<std::string>(*innerParamList, "BelosPrec Solver Type");
+  const Teuchos::RCP<Teuchos::ParameterList> packageParamList = 
+          Teuchos::rcp(new Teuchos::ParameterList(*Teuchos::sublist(innerParamList, "BelosPrec Solver Params")));
```
bartlettroscoe, line 199: Are you trying to create a copy of the sublist?  I don't see any reason why a copy should be created here.

If you are not trying to create a copy, just use:

```
const auto packageParamList = sublist(innerParamList, "BelosPrec Solver Params");
```

That idiom is that we want the input PL to get updated with default values if they are not already set.

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,369 @@
...
+  std::string solverTypeUpper (solverType);
+  if (solverTypeUpper.size () > 0) {
+    for (size_t k = 0; k < solverTypeUpper.size (); ++k) {
+      solverTypeUpper[k] = ::toupper(solverTypeUpper[k]);
+    }
+  }
+
+  // Create the initial preconditioner
+  if (Teuchos::nonnull(out) && Teuchos::includesVerbLevel(verbLevel, Teuchos::VERB_LOW)) {
+    *out << "\nCreating a new BelosTpetra::Preconditioner object...\n";
```
bartlettroscoe, line 211: Do you really want to print this at `Teuchos::VERB_LOW`?  Don't remember how this plays with other printing in the stack of objects.

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,369 @@
...
+    if (Teuchos::nonnull(out) && Teuchos::includesVerbLevel(verbLevel, Teuchos::VERB_LOW)) {
+      Teuchos::OSTab(out).o() << "> Creating half precision preconditioner\n";
+    }
+    const Teuchos::RCP<const MatrixType> tpetraFwdMatrix = Teuchos::rcp_dynamic_cast<const MatrixType>(tpetraFwdOp);
+    TEUCHOS_TEST_FOR_EXCEPT(Teuchos::is_null(tpetraFwdMatrix));
+    auto tpetraFwdMatrixHalf = tpetraFwdMatrix->template convert<half_scalar_type>();
+
+    Teuchos::RCP<BelosTpLinProbHalf> belosLinProbHalf = Teuchos::RCP<BelosTpLinProbHalf>(new BelosTpLinProbHalf());
+    belosLinProbHalf->setOperator(tpetraFwdMatrixHalf);
+    Teuchos::RCP<TpetraLinOpHalf> belosOpRCPHalf = Teuchos::RCP<TpetraLinOpHalf>(new BelosTpOpHalf(belosLinProbHalf, packageParamList, solverType, true));
```
bartlettroscoe, line 226: Replace this with:

```
const Teuchos::RCP<TpetraLinOpHalf> belosOpRCPHalf =
  rcp(new BelosTpOpHalf(belosLinProbHalf, packageParamList, solverType, true);
```

?

Does Belos::TpetraOperator not have any non-member constructors that directly return RCP objects?  (See "5.8.1 The non-member constructor function idiom" in [Teuchos C++ Memory Management Classes, Idioms, and Related Topics](https://docs.trilinos.org/dev/packages/teuchos/doc/html/TeuchosMemoryManagementSAND.pdf).)

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,369 @@
...
+    thyraPrecOp = Thyra::createLinearOp(wrappedOp);
+#else
+    if (Teuchos::nonnull(out)) {
+      *out << "\nSolver does not have correct precisions enabled to use half precision.\n";
+    }
+    TEUCHOS_TEST_FOR_EXCEPT(true);
+#endif
+  } else {
+    // Wrap concrete preconditioner
+    Teuchos::RCP<BelosTpLinProb> belosLinProb = Teuchos::RCP<BelosTpLinProb>(new BelosTpLinProb());
```
bartlettroscoe, line 238: Same here.  Why not:

```
const Teuchos::RCP<BelosTpLinProb> belosLinProb = rcp(new BelosTpLinProb());
```

?

Or, better yet, a non-member constructor function like:

```
const Teuchos::RCP<BelosTpLinProb> belosLinProb = BelosTpLinProb::create();
```

Would be nice to eliminate raw keyword `new` as much as possible.

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,369 @@
...
+
+
+template <typename MatrixType>
+void BelosTpetraPreconditionerFactory<MatrixType>::setParameterList(
+  Teuchos::RCP<Teuchos::ParameterList> const& paramList
+  )
+{
+  TEUCHOS_TEST_FOR_EXCEPT(Teuchos::is_null(paramList));
+
+  const Teuchos::RCP<const Teuchos::ParameterList> validParamList = this->getValidParameters();
```
bartlettroscoe, line 297: Could be:

```
const auto validParamList = this->getValidParameters();
```

bartlettroscoe (changes requested):
A few things should be addressed (also mentioned in comments below), IMHO:

1. Please try using [`Thyra::TpetraOperatorVectorExtraction`](https://docs.trilinos.org/dev/packages/thyra/doc/html/classThyra_1_1TpetraOperatorVectorExtraction.html) to get Tpetra objects out of Thyra adapters.  (And we can consider adding some more typdefs to make the usage code even more compact.)

2. Please use `TEUCHOS_TEST_FOR_EXCPETION(<test>, <except>, <message>)` instead of printing to `*out` and then calling `TEUCHOS_TEST_FOR_EXCEPT(true)`.  (See "GCG 17: Always send output to some general std::ostream object; Never send output directly to std::cout or std::cerr; Never print output with print(...) or printf(...)"  in [Thyra Coding and Documentation Guidelines](https://docs.trilinos.org/dev/packages/thyra/doc/html/ThyraCodingGuideLines.pdf).

3. Remove the empty file `packages/stratimikos/adapters/belos/tpetra/CMakeLists.txt`.

4. Related to previous, move directory `packages/stratimikos/adapters/belos/tpetra` to `packages/stratimikos/adapters/belos/src/tpetra`.  That would make more sense in the file `packages/stratimikos/adapters/belos/src/CMakeLists.txt`.

In addition to those, you might consider:

5. Consider tightening up RCP-related code with the usage of the non-member function `Teuchos::rcp()` (with `using Teuchos::rcp;` at the top and then using non-namespaced `rcp()` calls), or, better yet, define and use non-member constructor functions for the various classes (see below).

6. Consider the other misc suggestions given below.

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
jennloe: Going with the rcp solution for now.  We can discuss further improvements in the future.

jennloe: No, Belos::TpetraOperator does not currently have any non-member constructors.

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
jennloe: In the analogous Ifpack2 adapter (/ifpack2/adapters/thyra/Thyra_Ifpack2PreconditionerFactory_def.hpp) we definitely need a copy of that param list due to a potential for changes down the line.  The sublist function returns a const list. 

Right out of the box, the const sublist doesn't work for the BelosPrec adapter, either. I get a compiler error from the const-ness.  The first layer is in the BelosTpetraOperator class.  But even if I change that class to take a const param list, then it fails when I try to pass that list to SolverFactory.create, which also expects a non-cost list. (And then who knows what else down the line.)

I will leave this paramList copy as-is for now.

bartlettroscoe: @jennloe, you are supposed to write the default values of parameters not already set into the user's PL.  See:

* [`virtual void Teuchos::ParameterListAcceptor::setParameterList(const RCP< ParameterList > & 	paramList)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1ParameterListAcceptor.html#a8ff906ce725b1759c4c68dc9d4457f8d)

* [`void Teuchos::ParameterList::validateParametersAndSetDefaults(ParameterList const &validParamList, int const 	depth)`](https://docs.trilinos.org/dev/packages/teuchos/doc/html/classTeuchos_1_1ParameterList.html#a3408d34b88539d6f6024faa05856b91a)

I think there is a documentation problem here.

That is why the the above function `Teuchos::ParameterListAcceptor::setParameterList()` takes as **non-const** `ParameterList ` object.

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -0,0 +1,369 @@
...
+  Teuchos::RCP<LinearOpBase<scalar_type> > thyraPrecOp;
+
+  if (useHalfPrecision) {
+#ifdef THYRA_BELOS_PREC_ENABLE_HALF_PRECISION
+    if (Teuchos::nonnull(out) && Teuchos::includesVerbLevel(verbLevel, Teuchos::VERB_LOW)) {
+      Teuchos::OSTab(out).o() << "> Creating half precision preconditioner\n";
+    }
+    const Teuchos::RCP<const MatrixType> tpetraFwdMatrix = Teuchos::rcp_dynamic_cast<const MatrixType>(tpetraFwdOp);
+    TEUCHOS_TEST_FOR_EXCEPT(Teuchos::is_null(tpetraFwdMatrix));
+    auto tpetraFwdMatrixHalf = tpetraFwdMatrix->template convert<half_scalar_type>();
```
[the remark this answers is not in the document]
jennloe: That does look a bit nicer. However, I didn't write that utility, so I'm not going to touch it this round.

packages/stratimikos/adapters/belos/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
jennloe: Okay, @cgcgcg helped me clear up some confusion with the constness.  I think we've addressed the original issue.  Pushing updated version shortly.

jennloe:
Okay, @bartlettroscoe I'm letting the AT run again.  Think I addressed all your comments 1-4 and partially fixed 5 and 6.

packages/stratimikos/adapters/belos/src/tpetra/Thyra_BelosTpetraPreconditionerFactory_def.hpp
```diff
@@ -76,6 +76,7 @@
 
 namespace Thyra {
 
+using Teuchos::rcp;
```
bartlettroscoe, line 79: This needs to be removed and put into the individual function bodies where it is needed.  It is not safe to put a using declaration for a function in a header file like this.  See Appendix D.2 "Amendments for ’using’ declarations and directives" in [Thyra Coding and Documentation Guidelines (TCDG) Version 1.0](https://docs.trilinos.org/dev/packages/thyra/doc/html/ThyraCodingGuideLines.pdf) and  Item 59 "Don't write namespace usings in a header file before an #include" in ["C++ Coding Standards"](https://dl.acm.org/doi/abs/10.5555/1036281).

packages/ifpack2/adapters/thyra/Thyra_Ifpack2PreconditionerFactory_def.hpp
```diff
@@ -76,6 +76,7 @@
 
 namespace Thyra {
 
+using Teuchos::rcp;
```
bartlettroscoe, line 79: This needs to be moved into the function bodies that use this.  See other comment on this.

bartlettroscoe (changes requested):
@jennloe, thanks for considering my comments!

I have a few suggestions for improvements listed below but really one two requests for further changes:

1. Move the `using Teuchos::rcp;` from inside of the Thyra namespace and put it into the function bodies that use this function.

2. Add all of the template arguments to `TpetraOperatorVectorExtraction` and consider using a shortened typename in function body.

jennloe:
@bartlettroscoe I think that should do it now.  Thanks for the detailed comments!

bartlettroscoe (approved):
The last commit https://github.com/trilinos/Trilinos/commit/476d2a9f76ceabd1694738c68dc8b3661843e7c6 resolves all of the issues I raised.  Thanks for considering the feedback :-)

Looks good to me.

jennloe:
I was able to dismiss @cgcgcg 's review (knowing all the changes were addressed) to go ahead and merge it.

PR #11223, opened 2022-11-02
Ctest: Well, that was fun while it lasted

csiefer2:
Turns out Kokkos doesn't support SYCL CPU builds.  This is because it uses sub_groups in reductions and sub_groups are not supported by OneAPI 2022.3.0 on CPUs.


jhux2 (approved):
:(

GrahamBenHarper (approved):
rip

PR #11228, opened 2022-11-04
Muelu: add detailed output of aggregates

mayrmt:
@trilinos/muelu 

## Motivation

For debugging purposes, a new method `PrintAllNodesPerAggregate()` has been added to `MueLu::Aggregates`. It lists all aggregates as well as their nodes (either local or global node IDs).

## Stakeholder Feedback

Useful for debugging `MueLu::InterfaceAggregationFactory`.

## Testing

A unit test has been added.


packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
```diff
@@ -207,6 +207,14 @@ namespace MueLu {
      */
     void ComputeNodesInAggregate(Array<LO> & aggPtr, Array<LO> & aggNodes,Array<LO> & unaggregated) const;
 
+    /*! @brief List all aggregates and their nodes
+
+    @param out [in] Output stream to be used for printing
+
+    \warning Use this for debugging only, as this will generate a ton of screen output.
+    */
+    void PrintAllNodesPerAggregate(Teuchos::FancyOStream& out, bool globalNodeIds = true) const;
```
cgcgcg, line 216: document optional parameter

cgcgcg (changes requested):
Please also implement this for Aggregates_kokkos to avoid further divergence between the implementations.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_kokkos_def.hpp
```diff
@@ -260,6 +260,43 @@ namespace MueLu {
...
+
+    // Loop aggregates and their nodes and print them
+    const int myRank = vertex2AggId_->getMap()->getComm()->getRank();
+    LO numAggregates = GetNumAggregates();
+    const auto aggSizes = ComputeAggregateSizes(true);
+    TEUCHOS_ASSERT(numAggregates==aggSizes.size());
+
+    std::stringstream myStream;
+    myStream << "\n";
+    for (LO lAggId = Teuchos::OrdinalTraits<LO>::zero(); lAggId < numAggregates; ++lAggId)
```
mayrmt, line 282: @cgcgcg In the Kokkos version, I should probably do a `Kokkos::parallel_for` here, right? However, then the output might be scrambled, so how can I obtain deterministic output for the unit test?

GrahamBenHarper: Since this is a print routine, I'd `deep_copy` to host and do the printing on host since you can't print from device anyway

mayrmt: Ok, I have added my approach to this. Not sure, if its the best approach, though.

mayrmt:
I did a MPI+OpenMP build locally and now the aggregates are not deterministic anymore (even when setting `params.set<bool>("aggregation: deterministic", true);` during aggregate creation in the unit test using `gimmeUncoupledAggregates_kokkos()`). This makes it hard to test the output. Should we skip the test for the kokkos version? Any ideas how to fix this?

cgcgcg:
@mayrmt The way the Kokkos aggregate unit test is set up, the aggs aren't deterministic. Both the coloring and the aggregation algos need to be deterministic. Maybe you only switched over the aggregation? Have a look here: https://github.com/trilinos/Trilinos/blob/21f3f8ce046a8b4e5cce7b31480f31c20498c558/packages/muelu/src/Graph/UncoupledAggregation/MueLu_UncoupledAggregationFactory_kokkos_def.hpp#L307-L340

cgcgcg:
Oh, the issue isn't that the aggs are different, but that their numbering is different? I.e. Agg 0 can be agg 3 on the next run?

PR #11204, opened 2022-10-27
Master merge 20221027 175815

e10harvey:
@trilinos/<teamName>

PR #11231, opened 2022-11-04
Ifpack2: Make the Chebyshev power method consistent with itself

GrahamBenHarper:
@trilinos/ifpack2 

I was able to reproduce the Ifpack2 Cheby test failure reported in #11204 on my machine, and thanks to help from @cgcgcg after banging our heads together for a bit, we realized this is a strange combination of several things.
1. $D^{-1}A$ for this problem ( $A$ is in `calore1.mtx`) has two dominant eigenvalues, 12.5839 and -10.6639.
2. The power method uses a random initial guess with a default of 10 iterations.
3. The loop computing `lambdaMax` is computed using $\\|D^{-1}Ax\\|_2$, but the final computation after the loop instead uses $x^T D^{-1}A x$.

There are all sorts of strange things that can happen. For this particular example, the eigenvalue history when using `<Parameter name="debug" type="bool" value="true"/>` will look like it's converging to something in the range of 10-12 because it uses `norm2()` inside the power method iteration. However, the final estimate after breaking out of the loop is computed using $x^T A x$, which means if the vector still has a large component pointing in the direction of the negative eigenvalue, the final estimate in the history will suddenly be way off (anywhere between -10.6639 and 12.5839 basically).

This splits the power method into two routes: spectral radius (computing $\\|D^{-1}Ax\\|$) and dominant eigenvalue (computing $x^TD^{-1}Ax$), and adds it as an option to Chebyshev to test code coverage. Realistically, one shouldn't use `computeSpectralRadius=false` unless debugging a nasty matrix like the one that caused us so much trouble :)

### DIFF
--- a/packages/ifpack2/src/Ifpack2_PowerMethod.hpp
+++ b/packages/ifpack2/src/Ifpack2_PowerMethod.hpp
@@ -169,13 +169,13 @@ powerMethodWithInitGuess(const OperatorType& A,
   if(y.is_null())
     y = Teuchos::rcp(new V(A.getRangeMap()));
 
-  // GH 04 Nov 2022: The previous implementation of the power method was incorrect.
-  // It computed x->norm2() inside the loop, but returned x^T*Ax. This returned horribly
-  // incorrect answers in certain cases, such as the Cheby_belos test, which has two 
-  // largest eigenvalues of 12.5839 and -10.6639. The power method iteration history 
-  // would appear to converge to something between 10 and 12, but the result would 
-  // return something between -10 and 12 instead. These have now been split into two
-  // routes which will behave consistently to themselves.
+  // GH 04 Nov 2022: The previous implementation of the power method was inconsistent with itself.
+  // It computed x->norm2() inside the loop, but returned x^T*Ax.
+  // This returned horribly incorrect estimates for Chebyshev spectral radius in certain 
+  // cases, such as the Cheby_belos test, which has two dominant eigenvalues of 12.5839, -10.6639.
+  // In such case, the power method iteration history would appear to converge to something
+  // between 10 and 12, but the resulting spectral radius estimate was sometimes negative.
+  // These have now been split into two routes which behave consistently with themselves.
   if(computeSpectralRadius) {
     // In this route, the update is as follows:
     // y=A*x, x = Dinv*y, lambda = norm(x), x = x/lambda
--- a/packages/ifpack2/src/Ifpack2_PowerMethod.hpp
+++ b/packages/ifpack2/src/Ifpack2_PowerMethod.hpp
@@ -113,8 +113,9 @@ private:
 /// \param tolerance [in] The relative eigenvalue tolerance. (default: 1e-7)
 /// \param eigNormalizationFreq [in] The frequency of normalization. (default: 1)
 /// \param out [in] The stream to send verbose output to. (default: null)
-/// \param computeSpectralRadius [in] Compute and return \f$\|D^{-1}Ax\|_2\f$ 
-/// if true, compute and return \f$x^TD^{-1}Ax\f$ if false. (default: true)
+/// \param computeSpectralRadius [in] Compute the absolute value of the dominant
+/// eigenvalue of \f$D^{-1}A\f$ if true. Compute the dominant eigenvalue of 
+/// \f$D^{-1}A\f$ if false. (default: true)
 ///
 /// \return Estimate of the maximum eigenvalue of A*D_inv.
 template<class OperatorType, class V>
@@ -342,8 +343,9 @@ computeInitialGuessForPowerMethod (V& x, const bool nonnegativeRealParts)
 /// \param tolerance [in] The relative eigenvalue tolerance. (default: 1e-7)
 /// \param eigNormalizationFreq [in] The frequency of normalization. (default: 1)
 /// \param out [in] The stream to send verbose output to. (default: null)
-/// \param computeSpectralRadius [in] Compute and return \f$\|D^{-1}Ax\|_2\f$ 
-/// if true, compute and return \f$x^TD^{-1}Ax\f$ if false. (default: true)
+/// \param computeSpectralRadius [in] Compute the absolute value of the dominant
+/// eigenvalue of \f$D^{-1}A\f$ if true. Compute the dominant eigenvalue of 
+/// \f$D^{-1}A\f$ if false. (default: true)
 ///
 /// \return Estimate of the maximum eigenvalue of A*D_inv.
 template<class OperatorType, class V>
--- a/packages/ifpack2/src/Ifpack2_PowerMethod.hpp
+++ b/packages/ifpack2/src/Ifpack2_PowerMethod.hpp
@@ -86,7 +86,7 @@ public:
     typedef Kokkos::Details::ArithTraits<IST> STS;
     typedef Kokkos::Details::ArithTraits<typename STS::mag_type> STM;
 
-    if (STS::real (x_(i)) < STM::zero ()) {
+    if(STS::real (x_(i)) < STM::zero ()) {
       x_(i) = -x_(i);
     }
   }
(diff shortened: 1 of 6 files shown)


GrahamBenHarper:
okay @cgcgcg it's ready for review for real :)

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
```diff
@@ -113,18 +113,21 @@ class PositivizeVector {
 /// \param tolerance [in] The relative eigenvalue tolerance. (default: 1e-7)
 /// \param eigNormalizationFreq [in] The frequency of normalization. (default: 1)
 /// \param out [in] The stream to send verbose output to. (default: null)
+/// \param computeSpectralRadius [in] Compute and return \f$\|D^{-1}Ax\|_2\f$ 
+/// if true, compute and return \f$x^TD^{-1}Ax\f$ if false. (default: true)
```
cgcgcg, line 117: What is x?

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
```diff
@@ -152,70 +155,142 @@ powerMethodWithInitGuess (const OperatorType& A,
...
-
-  for (int iter = 0; iter < numIters-1; ++iter) {
-    if (verbose) {
-      *out << "  Iteration " << iter << endl;
+    y = Teuchos::rcp(new V(A.getRangeMap()));
+
+  // GH 04 Nov 2022: The previous implementation of the power method was incorrect.
+  // It computed x->norm2() inside the loop, but returned x^T*Ax. This returned horribly
+  // incorrect answers in certain cases, such as the Cheby_belos test, which has two 
+  // largest eigenvalues of 12.5839 and -10.6639. The power method iteration history
```
cgcgcg, line 175: The answers were correct, just not consistent with the debug output.

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
```diff
@@ -267,24 +342,27 @@ computeInitialGuessForPowerMethod (V& x, const bool nonnegativeRealParts)
 /// \param tolerance [in] The relative eigenvalue tolerance. (default: 1e-7)
 /// \param eigNormalizationFreq [in] The frequency of normalization. (default: 1)
 /// \param out [in] The stream to send verbose output to. (default: null)
+/// \param computeSpectralRadius [in] Compute and return \f$\|D^{-1}Ax\|_2\f$ 
+/// if true, compute and return \f$x^TD^{-1}Ax\f$ if false. (default: true)
```
cgcgcg, line 346: ditto

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
```diff
@@ -152,70 +155,142 @@ powerMethodWithInitGuess (const OperatorType& A,
...
-      *out << "  Iteration " << iter << endl;
+    y = Teuchos::rcp(new V(A.getRangeMap()));
+
+  // GH 04 Nov 2022: The previous implementation of the power method was incorrect.
+  // It computed x->norm2() inside the loop, but returned x^T*Ax. This returned horribly
+  // incorrect answers in certain cases, such as the Cheby_belos test, which has two 
+  // largest eigenvalues of 12.5839 and -10.6639. The power method iteration history 
+  // would appear to converge to something between 10 and 12, but the result would 
+  // return something between -10 and 12 instead. These have now been split into two
+  // routes which will behave consistently to themselves.
```
cgcgcg, line 178: I'm not sure this comment makes much sense without the actual matrix.

cgcgcg (changes requested):

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
GrahamBenHarper: It some cases it was reporting negative leading eigenvalues for a matrix with a leading positive eigenvalue, which you said yourself was definitely wrong :P

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
GrahamBenHarper: The parameter 5 rows above it

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
GrahamBenHarper: It references the test `Cheby_belos`, which has the matrix it refers to, right?

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
cgcgcg: No, it wasn't. It was the correct estimate given the prescribed number of iterations.

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
cgcgcg: Yep.

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
cgcgcg: nope, that's the initial vector.

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
GrahamBenHarper: Okay I see what you mean. I updated the comment accordingly.

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
GrahamBenHarper: ah got it

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
```diff
@@ -152,70 +156,142 @@ powerMethodWithInitGuess (const OperatorType& A,
...
-
-  for (int iter = 0; iter < numIters-1; ++iter) {
-    if (verbose) {
-      *out << "  Iteration " << iter << endl;
+    y = Teuchos::rcp(new V(A.getRangeMap()));
+
+  // GH 04 Nov 2022: The previous implementation of the power method was inconsistent with itself.
+  // It computed x->norm2() inside the loop, but returned x^T*Ax.
+  // This returned horribly incorrect estimates for Chebyshev spectral radius in certain 
+  // cases, such as the Cheby_belos test, which has two dominant eigenvalues of 12.5839, -10.6639.
```
cgcgcg, line 176: The estimate was correct given the number of iterations.

GrahamBenHarper: A negative estimate for a spectral radius is incorrect though?

cgcgcg: The method was computing an estimate for the dominant eigenvalue. After 10 its, it returned the correct (but arguably bad) estimate. After 30 its the estimate was better. The method itself was fine though, so I wouldn't say that something was "horribly incorrect", especially given that your implementation will return the exact same estimate when called with `computeSpectralRadius=true` after 10 iterations.

GrahamBenHarper: > given that your implementation will return the exact same estimate when called with computeSpectralRadius=true after 10 iterations.

Wait now I'm confused. `computeSpectralRadius=true` will definitely not return the same answer as it did previously for 10 iterations. `computeSpectralRadius=true` returns $\\|D^{-1}Ax\\|_2$ instead of the previous $x^TD^{-1}Ax$.

cgcgcg: Sorry, `computeSpectralRadius=false`

cgcgcg (dismissed):

packages/ifpack2/src/Ifpack2_PowerMethod.hpp
```diff
@@ -152,70 +156,142 @@ powerMethodWithInitGuess (const OperatorType& A,
...
+    }
+    for (int iter = 0; iter < numIters-1; ++iter) {
+      if(verbose) {
+        *out << "  Iteration " << iter << endl;
+      }
+      A.apply(*x, *y);
+      y->elementWiseMultiply(STS::one(), D_inv, *y, STS::zero());
+
+      if(((iter+1) % eigNormalizationFreq == 0) && (iter < numIters-2)) {
+        norm = x->dot(*y);
```
cgcgcg, line 249: @GrahamBenHarper While you're debugging, might as well change that variable name.

GrahamBenHarper: updated to `xDinvAx` :)

cgcgcg (approved):

GrahamBenHarper:
Thanks for the help and discussions and review @cgcgcg!

GrahamBenHarper:
@e10harvey FYI this will fix the random failures in Ifpack2

cgcgcg:
Thanks for implementing the fix @GrahamBenHarper !

PR #11248, opened 2022-11-09
tpetra: compares no. of packed values to packed row pointers

cwpearson:
@trilinos/tpetra

## Motivation
Previously, were comparing against unpacked row pointer array, which is filled w.r.t the allocated size of the matrix, not the number of entries actually used.

## Testing
The code is not covered by automatic testing, though testing locally with `TPETRA_DEBUG=1` will exercise it.


cgcgcg:
@cwpearson Since this bit of debug code is apparently not covered by any autotester or nightly buids: you ran the entire Tpetra test suite with `TPETRA_DEBUG=1 `and it passes now?

cwpearson:
I tried, then I got stuck in the issue fixed by pr 11254. I will see if I can complete the whole test suite now.

cwpearson:
One test continues to fail with this change, as it does in develop: https://github.com/trilinos/Trilinos/issues/11256

Issue #11250, opened 2022-11-09
tpetra: Can't take host view while device views are active


cwpearson:
@trilinos/tpetra 

When executing `ctest -R MatrixMarket_Operator -V` with `TPETRA_VERBOSE=1`, get output like

```
Test: Operator Matrix Market I/O, w/ Map with index base 0
170:   Creating the row Map
170:   Creating original matrix
170:  
170:  p=0: *** Caught standard std::exception of type 'std::runtime_error' :
170:  
170:   Tpetra::Details::WrappedDualView (name = Tpetra::CrsGraph column indices; host use_count = 2; device use_count = 3): Cannot access data on host while a device view is alive
170:  [FAILED]  (1.09 sec) OperatorOutput_double_int_longlong_Kokkos_Compat_KokkosCudaWrapperNode_IndexBase0_UnitTest
```

Accessing the host views here is invalid because the device view `indices_new` lives on.
https://github.com/trilinos/Trilinos/blob/14a80ebfde8f69dbbd55b24a42bf6d91cf9f2cb4/packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp#L379-L380

I guess the autotester does not have a GPU build with `TPETRA_VERBOSE=1` enabled.

jhux2:
> I guess the autotester does not have a GPU build with `TPETRA_VERBOSE=1` enabled.

almost certainly not

cwpearson:
To reproduce:
* `make -j`nproc` TpetraCore_MatrixMarket_Operator_Test`
* `export TPETRA_VERBOSE=1`
* `cd Trilinos`
* `ctest -R MatrixMarket_Operator -V`

jhux2:
> > I guess the autotester does not have a GPU build with `TPETRA_VERBOSE=1` enabled.
> 
> almost certainly not

That being said, we could certainly set up a _nightly_ build that enables `TPETRA_VERBOSE=1`.

cwpearson:
fixed in https://github.com/trilinos/Trilinos/pull/11251

PR #11251, opened 2022-11-09
tpetra: drop device views before requesting host in `TPETRA_VERBOSE=1`

cwpearson:
@trilinos/tpetra 

## Motivation
Can't getHostView from indices_wdv nor values_wdc while the device view used to create them lives on.

## Testing
`TPETRA_VERBOSE=1` on GPU execution is not covered by automatic tests.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
@@ -280,101 +280,103 @@ pad_crs_arrays(
         typename Indices::t_dev::non_const_value_type;
   using vals_value_type = typename Values::t_dev::non_const_value_type;
 
-  auto indices_old = indices_wdv.getDeviceView(Access::ReadOnly);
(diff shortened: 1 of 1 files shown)


cwpearson:
Attempts to fix https://github.com/trilinos/Trilinos/issues/11250

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -375,6 +375,10 @@ pad_crs_arrays(
   indices_wdv = Indices(indices_new);
   values_wdv = Values(values_new);
 
+  // drop device views before getting host in verbose block
+  indices_new = decltype(indices_new)();
+  values_new = decltype(values_new)();
```
cgcgcg, line 380: You could just wrap the usage of the two views into its own code block. But I guess this is also fine.

cgcgcg (approved):

Issue #11253, opened 2022-11-09
tpetra: `TpetraCore_Distributor_UnitTests_MPI_4` hangs when `TPETRA_DEBUG=1`


cwpearson:
## Question
@trilinos/tpetra

Vortex, 31509b0630525264c8ee689d88ac6e834acf0a1d

I believe this may be related to the following, but I'm not sure.

This test tries to send a message from proc i to proc i+1 to cause the distributor to throw:

https://github.com/trilinos/Trilinos/blob/31509b0630525264c8ee689d88ac6e834acf0a1d/packages/tpetra/core/test/Distributor/Distributor_UnitTests.cpp#L201

However, the `starts` array (actually, the number of sends to each proc...`starts`?) is length `comm_->getSize() + 1`, so actually this is not an immediate problem and is treated as a valid ID since it's greater than `0`.
https://github.com/trilinos/Trilinos/blob/31509b0630525264c8ee689d88ac6e834acf0a1d/packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp#L173

Actually, *any* id > comm_->getSize() will cause an out-of-bounds access into that starts array and cause a segfault instead of throwing (the test is easily modified to do this).

Furthermore, this getSize()+1 is never used except here:

https://github.com/trilinos/Trilinos/blob/31509b0630525264c8ee689d88ac6e834acf0a1d/packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp#L350-L351

where it appears to be an accident, or at least will yield a negative number for `lengthsTo_` which sounds wrong (since `starts` it's initialized to 0)

* What is the purpose of `starts` being the size it is?
* Is it reasonable to expect every process to throw if any process in this function receives an invalid ID?

cwpearson:
This commit https://github.com/trilinos/Trilinos/commit/868edb2767e29e0d8594e99e45121a3c0fbd7d96 removed the check for a bad process ID

cwpearson:
Fixed in pull/11254

PR #11254, opened 2022-11-09
tpetra: make distributor plan createFromSends throw on bad proc ID when `TPETRA_DEBUG=1`

cwpearson:
Fix for https://github.com/trilinos/Trilinos/issues/11253

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
@@ -40,6 +40,7 @@
 
 #include "Teuchos_StandardParameterEntryValidators.hpp"
 #include "Tpetra_Util.hpp"
+#include "Tpetra_Details_Behavior.hpp"
 #include <numeric>
 
 namespace Tpetra {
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/src/Tpetra_Details_DistributorPlan.cpp
```diff
@@ -175,10 +178,13 @@ size_t DistributorPlan::createFromSends(const Teuchos::ArrayView<const int>& exp
   // numActive is the number of sends that are not Null
   size_t numActive = 0;
   int needSendBuff = 0; // Boolean
-
+  int badID = -1; // only used in debug mode
```
tasmith4, line 181: If this is only used in debug mode, why not move this and the check inside the for loop down inside the `if (debug)`?  That would make it clear what's going on

cwpearson: good idea

tasmith4 (dismissed):

cwpearson:
@tasmith4 can you please re-review, I think my force push after I squashed invalidated it.

tasmith4 (approved):

PR #11260, opened 2022-11-11
Piro: refactor to use ROL version 2

mperego:
@trilinos/piro
@trilinos/rol

## Motivation
Switch to ROL version 2, as version 1 is now deprecated.
This PR also adds several improvements to the Piro::PerformAnalysis capability.
In particular we use ROL update function to determine what quantities need to be reused and to trigger observers (e.g. to save the solution).
Also, we provide the possibility to pass an adjoint model evaluator, similarly to what done for transient problems, to allow the user to provide its own implementation of the adjoint. This is typically useful when the solvers do not support transient solves, in this case the user can explicitly assemble the Jacobian transpose and provide it through the adjoint model.
We also improved the output, giving the user the possibility to fine tune the output verbosity.
Finally, we allow the user to set a recovery value for the objective function, to be used when state solves fail to converge. This forces ROL to reject the trial parameters that cause the solver failures.

 Please do not merge! These changes need to be synchronized with changes in Albany.

## Testing
Modified Piro tests and added new test to exercise new capabilities


mperego:
@dridzal I made some minor additions to ROL, hope they are fine with you. Also, I fixed one of the functions to check derivatives that were still calling the old update function.

mperego:
@trilinos/framework can you please advise on where I can reproduce build
`rhel7_sems-gnu-7.2.0-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables`
I tried on blake but I get an error when loading the environment. Thanks.

```
source $TRILINOS_SRC/packages/framework/GenConfig/gen-config.sh --cmake-fragment gen-config.cmake rhel7_sems-gnu-7.2.0-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables $TRILINOS_SRC
Using system 'rhel7' based on matching hostname 'blake.sandia.gov'.
Overriding system to 'rhel7' based on specification in build name 'rhel7_sems-gnu-7.2.0-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables'.
Matched environment name 'sems-gnu-7.2.0-serial' in build name 'rhel7_sems-gnu-7.2.0-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables'.
Matched complete configuration 'rhel7_sems-gnu-7.2.0-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables'
  for build name 'rhel7_sems-gnu-7.2.0-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables'.
* CMake fragment file written to: /home/mperego/trilinos-pr-build/gen-config.cmake

Using system 'rhel7' based on matching hostname 'blake.sandia.gov'.
Overriding system to 'rhel7' based on specification in build name 'rhel7_sems-gnu-7.2.0-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables'.
Matched environment name 'sems-gnu-7.2.0-serial' in build name 'rhel7_sems-gnu-7.2.0-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_no-mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables'.
  File "/home/mperego/trilinos/packages/framework/GenConfig/LoadEnv/load_env.py", line 547, in <module>
    main(sys.argv[1 :])
  File "/home/mperego/trilinos/packages/framework/GenConfig/LoadEnv/load_env.py", line 537, in main
    le.apply_env()
  File "/home/mperego/trilinos/packages/framework/GenConfig/LoadEnv/load_env.py", line 237, in apply_env
    rval = self.set_environment.apply(self.parsed_env_name)
  File "/home/mperego/trilinos/packages/framework/GenConfig/LoadEnv/setenvironment/SetEnvironment.py", line 323, in apply
    self.configparserenhanceddata[section]
  File "/home/mperego/trilinos/packages/framework/GenConfig/LoadEnv/configparserenhanced/ConfigParserEnhanced.py", line 1586, in __getitem__
    raise KeyError(key)
KeyError: 'rhel7_sems-gnu-7.2.0-serial'
```

kliegeois (commented):
I can't comment for the ROL modifications ( I leave that to @dridzal ); but the Piro ones look good to me. 
On top of the refactor, I like the addition of the extra errors and the changes to the cout that should ease the grep of the cout.

mperego:
Thanks for the review, @kliegeois !

dridzal:
The changes in the ROL files should be reviewed by @dpkouri ---please, take a look.

dpkouri (commented):
@mperego I am not sure I understand the purpose of "set_precomputed_state".  Can you explain this?  I believe that the state_ variable will be the previous computed state, which is typically a good initial guess for the "solve" function.  Is this not the case for your application?  In any case, couldn't the user's (Piro's) Constraint_SimOpt implementation manage the initial guess for the solve?  Why does it have to be passed to Reduced_Objective_SimOpt?

mperego:
> @mperego I am not sure I understand the purpose of "set_precomputed_state". Can you explain this? I believe that the state_ variable will be the previous computed state, which is typically a good initial guess for the "solve" function. Is this not the case for your application? In any case, couldn't the user's (Piro's) Constraint_SimOpt implementation manage the initial guess for the solve? Why does it have to be passed to Reduced_Objective_SimOpt?

@dpkouri we use that function when we want to compute the adjoint sensitivities (the gradients in a reduced-space framework). In that case we don't really want to solve an optimization problem, but simply compute the sensitivites with respect to a parameter once the state has already been computed. We could derive our version of the `Reduced_Objective_SimOpt` in Piro, but some of the methods we need to call are private in ROL class.

dpkouri:
@mperego Are you calling the "gradient" function in Reduced_Objective_SimOpt?  If so, the gradient will compute the state if it isn't already computed.

mperego:
@dpkouri yes, but we want to avoid that ROL calls again the state solver. Even if it's not particularly expensive because the the solution is already converged, it would be confusing/annoying for the user.

mperego:
@dpkouri The issue is that I cannot find a way to cleanly initialize `Reduced_Objective_SimOpt` with a given state. The constructors don't store the state in the `stateStore` and so the constraint solve gets called, which is what I want to avoid.
Initializing the `stateStore` outside of `Reduced_Objective_SimOpt` is also not possible because `getParameter`() function is protected, so I cannot use the `Reduced_Objective_SimOpt` constructors that accept a `stateStore`.

dpkouri:
@mperego By initializing Reduced_Objective_SimOpt with a pointer to a stateStore, you will have access to the state variable.  In particular, if you call "gradient" without solving the constraint equation, the gradient method with first solve the constraint equation, storing the solution in stateStore, and then compute the adjoint and gradient.  Since  you have a point to the stateStore, you can access the constraint solution.  As for accessing the samples through "getParameter()", if your problem is deterministic, then the only parameter is an empty std::vector.  If the problem is stochastic, then the parameters will correspond to the samples from a SampleGenerator that you pass in when constructing the stochastic objective function.

mperego:
@dpkouri OK, I didn't quit understand how `getParameter` works. So I was able to use the constructor and I reverted the changes in `Reduced_Objective_SimOpt`. Are you OK with the other changes in ROL?

dpkouri:
@mperego Have you verified that this solution satisfies your requirements?

mperego:
@dpkouri yeah, it's a bit more involved but it's fine. All Piro and Albany tests pass.

kliegeois (approved):
The modifications since my last review looks good, thanks!

PR #11276, opened 2022-11-15
Amesos2: Fixes for LAPACK adapter

cgcgcg:
@trilinos/amesos2 

## Motivation
LAPACK should be defined as a TPL dependency. With the LAPACK adapter enabled I got crashes. I copied some code over from the Klu2 adapter and that seems to fix it.


packages/amesos2/CMakeLists.txt
```diff
@@ -62,11 +62,6 @@ TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_TIMERS
 #
 
 
-TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_LAPACK
-  HAVE_${PACKAGE_NAME_UC}_LAPACK
-  "Enable the dense serial solver interface"
-  OFF )
```
GrahamBenHarper, line 68: @cgcgcg double-checking here. since you remove this and set LAPACK to an optional TPL, does that mean the macro name changed? I could be confusing this with our MueLu naming convention, but wouldn't it be something like `${PACKAGE_NAME}_HAVE_LAPACK`?

cgcgcg: The issue was that Amesos2 didn't declare LAPACK as a dependency. So tribits didn't set up these variables. Now it does, and the variable is called `Amesos2_ENABLE_LAPACK` and sets `HAVE_AMESOS2_LAPACK`. So this shouldn't break backwards compatibility unless I'm mistaken. This is the same naming scheme as in MueLu.

GrahamBenHarper: gotcha. that's what I was thinking of. that makes sense to me!

cgcgcg:
@srajama1 @ndellingwood @iyamazaki Could someone review this? Thank you!

ndellingwood (approved):

cgcgcg:
Thanks, @ndellingwood !

packages/amesos2/src/Amesos2_Lapack_def.hpp
```diff
@@ -289,16 +289,16 @@ namespace Amesos2 {
...
       ccs_helper::do_get(this->matrixA_.ptr(),
                          nzvals_view_, rowind_view_, colptr_view_,
-                         nnz_ret, ROOTED, SORTED_INDICES, 0);
+                         nnz_ret, ROOTED, ARBITRARY, this->rowIndexBase_);
     }
     else {
       ccs_helper::do_get(this->matrixA_.ptr(),
                          nzvals_view_, rowind_view_, colptr_view_,
-                         nnz_ret, CONTIGUOUS_AND_ROOTED, SORTED_INDICES, 0);
+                         nnz_ret, CONTIGUOUS_AND_ROOTED, ARBITRARY, this->rowIndexBase_);
```
PhilMiller, line 301: Totally aside from the changes in this PR, but I happened to be here looking:
* Based on the `if (is_contiguous_)` condition, is the use of `ROOTED` vs  `CONTIGUOUS_AND_ROOTED` reversed in the two calls?
* (unclear to me whether this matters) Does switching from `SORTED_INDICES` to `ARBITRARY` have negative performance implications, such that the code should conditionalize that for when it's necessary?

ndellingwood: @PhilMiller :

Re: point 1 - no, though the convention is a bit counter-intuitive (at least I found a bit confusing); e.g. CONTIGUOUS_AND_ROOTED means to hit the code path to make contiguous and rooted, which is the correct path when the GIDs are not contiguous

PhilMiller: That reversed convention, of the argument being a request rather than an affirmation, would perhaps be worth noting in a comment.

PhilMiller: Or, the enumerated values for that argument could be renamed to indicative that they're directives for what the code should do, not preconditions that it will rely upon

PR #11278, opened 2022-11-15
KokkosKernels: Fix #11033

### LINKED ISSUE #11033: Tpetra:  compilation errors in build with MKL enabled
@trilinos/tpetra
@trilinos/kokkos-kernels

### Description

What went wrong?  

We are seeing compilation errors in a Trilinos build with MKL enabled.  The compilation errors seem to follow from changes that were introduced in the recent kokkos snapshot in PR #10962.   

The error messages look like

```
In file included from ~/Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_mkl_impl.hpp:49,
from ~/Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_spgemm_numeric_spec.hpp:59,
from ~/Trilinos/packages/kokkos-kernels/src/sparse/KokkosSparse_spgemm_numeric.hpp:48,
from ~/home/costmo-user/Trilinos/packages/kokkos-kernels/src/sparse/KokkosSparse_spgemm.hpp:47,
from ~/Trilinos/packages/tpetra/core/ext/TpetraExt_MatrixMatrix_def.hpp:66,
from ~/Trilinos/packages/tpetra/core/ext/TpetraExt_MatrixMatrix.cpp:49:

~/Trilinos/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_mkl.hpp: 
In instantiation of 'class KokkosSparse::Impl::MKLSparseMatrix<long long int>':

~/Trilinos/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_mkl.hpp:126:58: error: static assertion failed: Scalar type used in MKLSparseMatrix<value_type> is NOT supported by MKL
    126 |   static_assert(mkl_is_supported_value_type<value_type>::value,
```

Do you have an idea what might fix things?  

It seems Tpetra wants to instantiate several classes using `int` and `long long int` as scalar type.  This appears to cause problems here in sparse GEMM calls when MKL is enabled.  It seems that kokkos-kernels is deciding whether or not to direct sparse GEMM calls to MKL by using runtime conditionals in `src/sparse/impl/KokkosSparse_spgemm_numeric_spec.hpp`.  But, PR #10962 changed how function calls are directed to MKL and introduced a compile-time assert to check that the scalar type is supported by MKL (which `int` and `long long int`, seemingly, are not).    

One way of solving this issue might be to replace the compile-time assert with runtime conditionals.  Changing `KokkosSparse_Utils_mkl.hpp ` as in the following patch solved the compilation errors in our build:

```
From d79bd6181a5449d4f558276165af4766f671349c Mon Sep 17 00:00:00 2001
From: Maarten Arnst <maarten.arnst@uliege.be>
Date: Sat, 17 Sep 2022 11:13:34 +0200
Subject: [PATCH] Change compile-time assert into run-time check in
 MKLSparseMatrix.

---
 .../src/sparse/KokkosSparse_Utils_mkl.hpp        | 16 ++++++++++------
 1 file changed, 10 insertions(+), 6 deletions(-)

diff --git a/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_mkl.hpp b/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_mkl.hpp
index b9eb3a9bd25..4abeeca88c1 100644
--- a/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_mkl.hpp
+++ b/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_mkl.hpp
@@ -123,16 +123,16 @@ template <typename value_type>
 class MKLSparseMatrix {
   sparse_matrix_t mtx;
 
-  static_assert(mkl_is_supported_value_type<value_type>::value,
-                "Scalar type used in MKLSparseMatrix<value_type> is NOT "
-                "supported by MKL");
-
  public:
   inline MKLSparseMatrix(sparse_matrix_t mtx_) : mtx(mtx_) {}
 
   // Constructs MKL sparse matrix from KK sparse views (m rows x n cols)
   inline MKLSparseMatrix(const MKL_INT num_rows, const MKL_INT num_cols,
-                         MKL_INT *xadj, MKL_INT *adj, value_type *values);
+                         MKL_INT *xadj, MKL_INT *adj, value_type *values) {
+    throw std::runtime_error(
+        "Scalar type used in MKLSparseMatrix<value_type> is NOT "
+                "supported by MKL");                       
+  }
 
   // Allows using MKLSparseMatrix directly in MKL calls
   inline operator sparse_matrix_t() const { return mtx; }
@@ -140,7 +140,11 @@ class MKLSparseMatrix {
   // Exports MKL sparse matrix contents into KK views
   inline void export_data(MKL_INT &num_rows, MKL_INT &num_cols,
                           MKL_INT *&rows_start, MKL_INT *&columns,
-                          value_type *&values);
+                          value_type *&values) {
+    throw std::runtime_error(
+        "Scalar type used in MKLSparseMatrix<value_type> is NOT "
+                "supported by MKL");   
+  }
 
   inline void destroy() {
     KOKKOSKERNELS_MKL_SAFE_CALL(mkl_sparse_destroy(mtx));
-- 
2.25.1

```

brian-kelley:
Do not emit compile-time error if attemping to use MKL utils with unsupported scalar types. These changes were suggested by @maartenarnst .

This can be considered a temporary fix because the proper TPL wrappers using unification layer will not instantiate MKL wrappers for unsupported types in the first place. In particular, this is coming for SpGEMM.

@trilinos/kokkos-kernels 

## Motivation

Replaces compile-time error with runtime exception. When using Tpetra::CrsMatrix with Scalar being something other than float/double/complex_float/complex_double (which is perfectly fine), KokkosKernels shouldn't be trying to instantiate any MKL utils/wrappers for those scalar types. But since it is, replacing with runtime error instead is a reasonable fix for now.

For the 4.0 release, the MKL SpGEMM wrappers will go through the unification layer and only get instantiated for supported type combinations, so these can become static_asserts once again.

### DIFF
--- a/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_mkl.hpp
+++ b/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_mkl.hpp
@@ -123,16 +123,16 @@ template <typename value_type>
 class MKLSparseMatrix {
   sparse_matrix_t mtx;
 
-  static_assert(mkl_is_supported_value_type<value_type>::value,
-                "Scalar type used in MKLSparseMatrix<value_type> is NOT "
-                "supported by MKL");
-
  public:
   inline MKLSparseMatrix(sparse_matrix_t mtx_) : mtx(mtx_) {}
 
   // Constructs MKL sparse matrix from KK sparse views (m rows x n cols)
   inline MKLSparseMatrix(const MKL_INT num_rows, const MKL_INT num_cols,
-                         MKL_INT *xadj, MKL_INT *adj, value_type *values);
+                         MKL_INT *xadj, MKL_INT *adj, value_type *values) {
+    throw std::runtime_error(
+        "Scalar type used in MKLSparseMatrix<value_type> is NOT "
+                "supported by MKL");
+  }
 
   // Allows using MKLSparseMatrix directly in MKL calls
   inline operator sparse_matrix_t() const { return mtx; }
@@ -140,7 +140,11 @@ class MKLSparseMatrix {
   // Exports MKL sparse matrix contents into KK views
   inline void export_data(MKL_INT &num_rows, MKL_INT &num_cols,
                           MKL_INT *&rows_start, MKL_INT *&columns,
-                          value_type *&values);
+                          value_type *&values) {
+    throw std::runtime_error(
+        "Scalar type used in MKLSparseMatrix<value_type> is NOT "
+                "supported by MKL");
+  }
 
   inline void destroy() {
     KOKKOSKERNELS_MKL_SAFE_CALL(mkl_sparse_destroy(mtx));


ndellingwood (approved):
lgtm, thanks @brian-kelley !

ndellingwood:
Cross-reference #11033

PR #11283, opened 2022-11-18
PIRO: fix bug with ostream

rppawlo:
@mperego 

## Motivation

Recent change to piro is causing compile failures on all empire builds. The error is in an ifdef that apparently is not exercised by PR builds but is used by empire.

## Testing

N/A


etphipp (approved):

mperego:
Thanks @rppawlo and sorry for the trouble.

PR #11290, opened 2022-11-18
Piro: use implicitly defined adjoint model evaluator for transient sensitivities

mperego:
This PR allows using Tempus implicitly defined adjoint model evaluator when the user provided adjoint model is null. 

@trilinos/piro

## Testing
Added test case to exercise use of implicitly defined adjoint model evaluator.


ikalash (approved):

mperego:
Thanks @ikalash !

Issue #11302, opened 2022-11-22
`tmpSendBuf` is equivalent to `tmpSend` already in scope


cwpearson:
@trilinos/tpetra 

https://github.com/trilinos/Trilinos/blob/7dfdae2518dc0a99df31ea677969b1df442cbdb1/packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp#L337

should be removed

cwpearson:
Now I'm concerned this was intentional due to some interaction with `persistingView` in the isend implementation:

https://github.com/trilinos/Trilinos/blob/7dfdae2518dc0a99df31ea677969b1df442cbdb1/packages/teuchos/kokkoscomm/src/Kokkos_TeuchosCommAdapters.hpp#L109

PR #11303, opened 2022-11-22
Remove redundant subview from `DistributorActor::doPosts`

cwpearson:
Fixes https://github.com/trilinos/Trilinos/issues/11302

`tmpSendBuf` is identical to `tmpSend` which is already in-scope, created a few lines above.


tasmith4 (approved):

cwpearson:
@csiefer2 we discussed drilling down on this in the Tpetra developer meeting but it looks like the tests are passing, how would you like to proceed?

Issue #11297, opened 2022-11-21
Stokhos:  Build errors with new CUDA 11 build


srbdev:
## Bug Report
@trilinos/stokhos 

### Description
There are build errors for (and related to) the Stokhos package for the `rhel7_sems-cuda-11.4.2-sems-gnu-10.1.0-sems-openmpi-4.0.5_release_static_Volta70_no-asan_complex_no-fpic_mpi_pt_no-rdc_no-uvm_deprecated-on_all` build. It is currently ran on the following fork https://github.com/e10harvey/Trilinos and branch `sprint17_cpp17`.

See the [CDash link](https://trilinos-cdash.sandia.gov/viewBuildError.php?buildid=616428) for additional details.

ndellingwood:
Based on the compilation error messages in the cdash link this is likely following merge of #11263 , @vqd8a can you help take a look?
@srbdev is Stokhos disabled in the cuda/11 PR testing? If so, that is likely how this slipped through

srbdev:
It used to be but not anymore. I've deleted all the explicitly removed packages from the build in order to catch as many of these errors as possible. But yes, that's the reason it slipped through in the past.

ndellingwood:
@vqd8a here is a compilation error snip from the first cuda/11.4.2 error showing in stokhos:
```
Error while building C++ object file " packages/stokhos/src/CMakeFiles/stokhos_ifpack2_mp_16_serial.dir/Ifpack2_LocalSparseTriangularSolver_MP_Vector_16_Serial.cpp.o" in target stokhos_ifpack2_mp_16_serial

trilinos/workspace/Trilinos_nightly_pipeline/Trilinos/packages/kokkos-kernels/src/sparse/KokkosSparse_Utils_cusparse.hpp(120): error: static assertion failed with "cuSparse TPL does not support scalar type"
          detected during:
            instantiation of "cudaDataType KokkosSparse::Impl::cuda_data_type_from<T>() [with T=Sacado::MP::Vector<SFS_int_double_16_DFN_CPU_default_local_ordinal_type_default_global_ordinal_type_Kokkos_Compat_KokkosSerialWrapperNode>]" 
trilinos/workspace/Trilinos_nightly_pipeline/Trilinos/packages/kokkos-kernels/src/sparse/impl/KokkosSparse_sptrsv_cuSPARSE_impl.hpp(119): here
[...]

PR #11308, opened 2022-11-25
Kokkos-Kernels:  Disable compile-time check on supported scalar types for cuSPARSE

etphipp:
This change removes the compile-time static-assert on valid scalar types for the cuSPARSE TPL.  The run-time check is still there, and this mimics the behavior for the interface to the older (CUDA-10) cuSPARSE interface.  Making this a compiler error makes it very difficult for packages such as Ifpack2 to support scalar types that cuSPARSE might not, since it can't even instantiate Kokkos-Kernels when cuSPARSE is eanbled and fixes several compiler errors when Ifpack2 is instantiated on Stokhos scalar types.

For issue #11297.


vqd8a (approved):
@etphipp Thanks for the change. LGTM.

PR #11311, opened 2022-11-27
Ifpack2: Call cuSPARSE SPTRSV if scalar type is supported by cuSPARSE

vqd8a:
@trilinos/ifpack2 

## Motivation

As per the suggestion of @etphipp in #11297, this PR adds another condition of scalar types supported by cuSPARSE for Ifpack2 calling cuSPARSE SPTRSV. Otherwise, Ifpack2 calls Kokkos Kernels's SPTRSV implementation.


etphipp (approved):
Looks good to me, thanks!

PR #11319, opened 2022-11-30
README: Update the README to point users to the Trilinos package owners wiki

GrahamBenHarper:
This is one of many things that may help external users reach package owners. It relies on the wiki page being as up-to-date as possible, which isn't much of a limitation, as it is easy to edit the wiki.

This helps external users mention the correct parties because external users may not mention teams for a project.


jhux2 (approved):
lgtm

PR #11320, opened 2022-11-30
Intrepid2: add ETI for quadrature classes

### LINKED ISSUE #10806: Intrepid2: Internal compiler errors with intel-2021.3 when using "-march=broadwell -mtune=broadwell"
## Bug Report
@trilinos/Intrepid2

### Description
When building with intel-2021.3 and "-march=broadwell -mtune=broadwell" we are seeing files that depend on Intrepid2 fail to compile with an internal compiler error:
```
internal error: 101003_1112
```
It looks like this can be reproduced just with the Intrepid2 tests enabled.

### Steps to Reproduce
1. SHA1:  245898d71ebfd26b2a2b4d21887fe1c9b72ca8f9
1. Configure script:
```
cmake \
  -GNinja \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_CXX_STANDARD=14 \
  -DCMAKE_C_COMPILER=/projects/sierra/linux_rh7/SDK/compilers/intel/IntelOneAPI-2021.3.0/mpi/2021.3.0/bin/mpicc \
  -DCMAKE_CXX_COMPILER=/projects/sierra/linux_rh7/SDK/compilers/intel/IntelOneAPI-2021.3.0/mpi/2021.3.0/bin/mpicxx \
  -DCMAKE_C_FLAGS:STRING="-O2 -ftemplate-depth-128 -finline-functions -w0 -march=broadwell -mtune=broadwell" \
  -DCMAKE_CXX_FLAGS:STRING="-O2 -ftemplate-depth-128 -finline-functions -w0 -march=broadwell -mtune=broadwell" \
  -DTrilinos_ENABLE_ALL_PACKAGES:BOOL=OFF \
  -DTrilinos_ENABLE_TESTS:BOOL=ON \
  -DTrilinos_ENABLE_Intrepid2:BOOL=ON \
  -DTrilinos_ENABLE_MueLu:BOOL=ON \
  ../
  ```
1. Configure log:  [attach here]
1. Build log:  [attach here]
1. Input deck:  [attach here]
1. Do this.
1. Do that.
1. Shake fist angrily at computer.
1. Run log:  [attach here]

mperego:
@trilinos/intrepid2

## Motivation

Introducing ETI for quadrature (cubature) classes to work around a bug with Intel compilers (see #10806).

Closes #10806.

## Testing

At the moment we are doing explicit instantiation using `double` for quadrature points and weights. We might want to specify Intrepid2 instantiation types in the future.


CamelliaDPG:
@mperego Thanks for doing this.  Just to confirm: it is possible to mix and match ETI with other (non-ETI) instantiations?  I.e., if someone wants to do Cubature with Fad types, that's still possible, right?

Maybe more precisely: how does the build system "know" whether to look for object code compiled in a .cpp file (as in the ETI for doubles, here) versus building new object code from the header (as would happen here for an instantiation of Cubature with Fad types)?

If we *can* mix and match in this way, then it seems very worthwhile to do ETI for doubles across the code; should get us considerable speedups for building.  The trick for Fad types is that some of those may not be known in advance, so we can't pre-build all of them.  But it might still be worthwhile to do ETI for particular, common Fad cases.

Issue #11324, opened 2022-12-01
Tpetra: `memcpy` in `DistributorActor::packOffset` assumes views of trivial types


cwpearson:
https://github.com/trilinos/Trilinos/blob/901236a9aa18ae9ec066fce5fc80d928b5ca0b5b/packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp#L165

There's no guarantee that `DstViewType::value_type` is trivially copy-assignable (e.g. `Sacado::MP::Vector<...>`). This will emit a warning in GCC 8+. Consider replacing with

```c++
auto first = src.data() + src_offset;
auto last = first + size;
auto d_first = dst.data() + dst_offset;
TEUCHOS_TEST_FOR_EXCEPTION(d_first >= first && d_first < last,
                    std::logic_error, "range overlap in packOffset");
std::copy(first, last, d_first); 
```

or possibly
```c++
Kokkos::Compat::deep_copy_offset(dst, src, dst_offset, src_offset, size);
```
keeping in mind that Kokkos mayn not yet handle this correctly: https://github.com/kokkos/kokkos/issues/2021

Issue #11296, opened 2022-11-21
Tempus_IMEX_RK_Partitioned_Staggered_FSA_Partitioned_IMEX_RK_1st_Order_MPI_1 randomly timing out in Trilinos PR builds


bartlettroscoe:
CC: @trilinos/tempus, @ccober6, @trilinos/framework 

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](https://trilinos-cdash.sandia.gov/queryTests.php?project=Trilinos&begin=2022-09-01&end=2022-11-21&filtercount=3&showfilters=1&filtercombine=and&field1=testname&compare1=61&value1=Tempus_IMEX_RK_Partitioned_Staggered_FSA_Partitioned_IMEX_RK_1st_Order_MPI_1&field2=status&compare2=61&value2=failed&field3=details&compare3=63&value3=timeout) (click "Shown Matching Output" in upper right) the tests:

* `Tempus_IMEX_RK_Partitioned_Staggered_FSA_Partitioned_IMEX_RK_1st_Order_MPI_1`

in the unique GenConfig builds:

* `rhel7_sems-gnu-7.2.0-openmpi-1.10.1-serial_debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables`
* `rhel7_sems-gnu-8.3.0-openmpi-1.10.1-serial_debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_all`

started failing on testing day 2022-10-28.

The specific set of CDash builds impacted where:

* `PR-11226-test-rhel7_sems-gnu-7.2.0-openmpi-1.10.1-serial_debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables-1442`
* `PR-Trilinos_nightly_gcc_8.3.0-debug-test-rhel7_sems-gnu-8.3.0-openmpi-1.10.1-serial_debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_all-154`
[...]

Issue #10194, opened 2022-02-16
Tpetra:  Distributor slow path really is slow


kddevin:
## Bug Report
@trilinos/tpetra @trilinos/zoltan2 @egboman @tasmith4 @seheracer @jhux2 

### Description

@seheracer reports slowdowns in Zoltan2's MultiJagged partitioner.  Slowdowns in Multijagged were also reported by @jhux2  in #9835.

I tracked the problem to Tpetra's slow-path distributor.  

Time spent in the Distributor increased roughly 50x after commit 2a8b7c437c2ce28898413bf013151addec939f77 relative to 76c7b6ac4461d3bbcc8108061df65b0bbdde8679 (determined by git bisect).  The difference is that the new code uses Kokkos::Views rather than ArrayRCPs.  Specifically, the new code uses deep_copy_offset.  In the slow path's tight loops, deep_copy_offset is creating too much overhead.

Running a Zoltan2 test from Sphynx with 2a8b7c437c2ce28898413bf013151addec939f77, we see the slow path distributor times:
```
Tpetra::Distributor: doPosts(3) KV                27.07 (546)       40.56 (546)       53.89 (546)       0.07428 (546)        
Tpetra::Distributor: doPosts(3): barrier KV       0 (0)             0 (0)             0 (0)             0 (0)                
Tpetra::Distributor: doPosts(3): recvs KV         0.002004 (546)    0.002272 (546)    0.002559 (546)    4.162e-06 (546)      
Tpetra::Distributor: doPosts(3): sends KV         27.07 (546)       40.55 (546)       53.89 (546)       0.07427 (546)        
Tpetra::Distributor: doPosts(3): sends KV FAST    0.383 (540)       0.4673 (537)      0.5496 (534)      0.0008702 (537)      
Tpetra::Distributor: doPosts(3): sends KV SLOW    26.68 (6)         40.08 (9)         53.4 (12)         4.453 (9)       
```

When I do a hack replacing deep_copy_offset with memcpy, the times reduce to times comparable to the ArrayRCP version of the code in 76c7b6ac4461d3bbcc8108061df65b0bbdde8679 (which uses std::copy)
```
Tpetra::Distributor: doPosts(3) KV                0.4761 (547)      0.5908 (547)      0.7083 (547)      0.00108 (547)        
Tpetra::Distributor: doPosts(3): barrier KV       0 (0)             0 (0)             0 (0)             0 (0)                
[...]

PR #11325, opened 2022-12-01
Tpetra: Replace memcpy with std::copy in case of non-trivial types

cwpearson:
@trilinos/tpetra 

Potential fix for https://github.com/trilinos/Trilinos/issues/11324

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp
@@ -41,6 +41,7 @@
 #define TPETRA_DETAILS_DISTRIBUTOR_ACTOR_HPP
 
 #include "Tpetra_Details_DistributorPlan.hpp"
+#include "Tpetra_TypeTraits.hpp"
 #include "Tpetra_Util.hpp"
 
 #include "Teuchos_Array.hpp"
@@ -50,6 +51,8 @@
 #include "Teuchos_Time.hpp"
 
 #include "Kokkos_TeuchosCommAdapters.hpp"
+
+#include <algorithm>
 
 #ifdef HAVE_TPETRA_MPI
 #include "mpi.h"
@@ -167,7 +170,19 @@ packOffset(const DstViewType& dst,
            const size_t src_offset,
            const size_t size)
 {
-  memcpy(dst.data()+dst_offset, src.data()+src_offset, size*sizeof(typename DstViewType::value_type));
+  using DstScalar = typename DstViewType::value_type;
+  using SrcScalar = typename SrcViewType::value_type;
+
+  if constexpr (Tpetra::okay_to_memcpy_v<DstScalar> && Tpetra::okay_to_memcpy_v<SrcScalar>) {
+    // these types may not actually be trivially-copyable according to c++, which may issue a warning.
+    // If they're not trivially-copyable, that means our users have opted in, so cast to void* to
+    // silence the warning
+    const void* srcp = reinterpret_cast<const void*>(src.data()+src_offset);
+    void* dstp = reinterpret_cast<void *>(dst.data()+dst_offset);
+    memcpy(dstp, srcp, size * sizeof(DstScalar));
+  } else {
+    std::copy(src.data() + src_offset, src.data() + src_offset + size, dst.data() + dst_offset);
+  }
 }
 
 template <typename DstViewType, typename SrcViewType>
--- a/packages/tpetra/core/src/Tpetra_TypeTraits.hpp
+++ b/packages/tpetra/core/src/Tpetra_TypeTraits.hpp
@@ -1,0 +1,92 @@
+// @HEADER
+// ***********************************************************************
+//
+//          Tpetra: Templated Linear Algebra Services Package
+//                 Copyright (2008) Sandia Corporation
+//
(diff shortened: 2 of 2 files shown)


tasmith4:
@cwpearson you might want to test this out for performance e.g. with use cases from #10194.  If this is the part of the code I'm thinking it is, we moved away from `Kokkos::Compat::deep_copy_offset` for performance reasons.

cwpearson:
Thanks for bringing this to my attention. I'm happy to avoid `Kokkos::deep_copy_offset` on the host-accessible spaces.

In my opinion we need an `std::copy` path for views of non-trivial things. `std::copy` can almost certainly be counted on to do the right thing for trivial types when optimization is enabled (i.e., just call `memmove` or `memcpy`).

The safest and highest maintenance burden would be to detect trivial types ourselves with `std::is_trivial` and call `memcpy` explicitly in that case.

tasmith4 (dismissed):
I'm approving, since this seems to be a reasonable solution that respects performance concerns.

However after reading your comment about MPI I'd personally come down on the side of not changing this at all.  This routine is explicitly used for packing an MPI send buffer, it should not be used for anything else.  If memcpy will do the wrong thing, I'd expect MPI to do the wrong thing also, so I don't see a need to move away from memcpy in this routine.

cwpearson:
I don't want to issue a warning here, so I'd rather just do what's obviously right wherever possible.

What sent me down this road is a warning like the following:

```
void* memcpy(void*, const void*, size_t)' writing to an object of type 'Kokkos::Impl::ViewMapping<Kokkos::ViewTraits<Sacado::MP::Vector<Stokhos::StaticFixedStorage<int, double, 16, Kokkos::Cuda> >*, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, Kokkos::MemoryTraits<0> >, Kokkos::Experimental::Impl::ViewMPVectorContiguous>::sacado_mp_vector_type' {aka 'class Sacado::MP::Vector<Stokhos::StaticFixedStorage<int, double, 16, Kokkos::Cuda> >'} with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
```

Basically, `Sacado::MP::Vector<...>` is not a [trivially copyable class](https://en.cppreference.com/w/cpp/language/classes#Trivially_copyable_class), but probably it actually is  actually trivially copyable and just unnecessarily defines one of the constructors or operators that disqualifies it. 

I'd advocate for some static asserts that prevent people from using Tpetra with types that are unsafe for MPI, but that's a broader issue.

GrahamBenHarper:
also blocked by #11296

cwpearson:
Yes, here's an example of the problem:

https://github.com/trilinos/Trilinos/blob/develop/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector.hpp#L430-L452

cwpearson:
I've made a change so that we will still call `memcpy` on two kinds of types:

1. Those that are trivially copyable
2. Those types `S` that do something like this:

```
struct S {
...
  using tpetra_memcpy_opt_in = std::true_type;
...
}; 

We could tell Sacado that if they're concerned about this, they need to add the above typedef to their types to explicitly opt-in.

**Review by brian-kelley — APPROVED** [review]

**Comment by jhux2** [comment]
There was a network problem earlier today, let's trying retesting.

File: packages/tpetra/core/src/Tpetra_Details_DistributorActor.hpp [hunk 1]
```diff
@@ -167,7 +170,19 @@ packOffset(const DstViewType& dst,
...
+  if constexpr (Tpetra::okay_to_memcpy_v<DstScalar> && Tpetra::okay_to_memcpy_v<SrcScalar>) {
+    // these types may not actually be trivially-copyable according to c++, which may issue a warning.
+    // If they're not trivially-copyable, that means our users have opted in, so cast to void* to
+    // silence the warning
+    const void* srcp = reinterpret_cast<const void*>(src.data()+src_offset);
+    void* dstp = reinterpret_cast<void *>(dst.data()+dst_offset);
+    memcpy(dstp, srcp, size * sizeof(DstScalar));
+  } else {
+    std::copy(src.data() + src_offset, src.data() + src_offset + size, dst.data() + dst_offset);
+  }
```
crtrott, line 185: Have you actually measured that there is a difference? I would be VERY surprised if std::copy doesn't go to memcpy already for trivial types.

packages/tpetra/core/src/Tpetra_TypeTraits.hpp
```diff
@@ -0,0 +1,92 @@
...
+template <typename T>
+struct okay_to_memcpy<
+    T, typename std::enable_if<std::is_same_v<typename T::tpetra_memcpy_opt_in,
+                                              std::true_type>>::type>
+    : public std::true_type {};
+
+template <typename T>
+constexpr bool okay_to_memcpy_v = okay_to_memcpy<T>::value;
+
+} // namespace Tpetra
```
crtrott, line 90: You really want to make this guy a public thing in Tpetra? Should this be in Details or wahtever the impl namespace is?

crtrott (changes requested):

crtrott:
lol ok just now seeing that this is super old ...

cwpearson:
Perhaps revisit this another time.

PR #11329, opened 2022-12-01
Remove CUDA_LAUNCH_BLOCKING from testing

sebrowne:
Not needed for testing, and confusing for people trying to ensure that CUDA_LAUNCH_BLOCKING is not set anymore.

User Support Ticket(s) or Story Referenced: TRILINOSHD-227

@trilinos/framework 

## Motivation
Real variables should not be used in capability testing, for pretty much this exact reason (Roger was concerned that it may have been used "for real").


rppawlo (approved):
Thanks @sebrowne !

e10harvey (approved):

PR #11334, opened 2022-12-02
Tpetra: cleanup: unused parameters in Tpetra_Details_Hash.hpp

cwpearson:
@trilinos/tpetra 

Remove some unused parameters in `Tpetra_Details_Hash.hpp`

Using any of these functions will already emit a compile-time error, so this just reduces spurious warnings.


jhux2 (approved):
lgtm, but why not just delete the names altogether?

cwpearson:
Just to emphasize that it's part of a consistent interface (it's just the non-`int` version of a template)

GrahamBenHarper:
blocked by #11296

PR #11335, opened 2022-12-02
Tpetra: cleanup: set-but-unused variable in Details_DistributorPlan.cpp

cwpearson:
@trilinos/tpetra

`index` is set but unused in two places in Tpetra_Details_DistributorPlan.cpp


jhux2 (approved):
lgtm

Issue #11331, opened 2022-12-01
Prepare for `ExecSpace::concurrency()` member function becoming non-static in Kokkos 


dalg24:
In anticipation of an upcoming change in Kokkos (see kokkos/kokkos#5655) here is a list of lines that need to be updated.
@ndellingwood would you please coordinate with the developers to make sure Trilinos does break when we implement the changes in Kokkos develop?
```
packages/amesos2/src/Amesos2_Superlumt_def.hpp:465:    int default_nprocs = Kokkos::DefaultHostExecutionSpace::concurrency();
packages/amesos2/src/Amesos2_Superlumt_def.hpp:524:      int default_nprocs = Kokkos::DefaultHostExecutionSpace::concurrency();
packages/muelu/src/Rebalancing/MueLu_RepartitionHeuristicFactory_def.hpp:135:      thread_per_mpi_rank = execution_space::concurrency();
packages/muelu/test/scaling/JacobiKernelDriver.cpp:480:    out<< "Kokkos::Compat::KokkosOpenMPWrapperNode::execution_space::concurrency() = "<<Kokkos::Compat::KokkosOpenMPWrapperNode::execution_space::concurrency()<<endl
packages/muelu/test/scaling/MatvecKernelDriver.cpp:959:    out<< "Kokkos::Compat::KokkosOpenMPWrapperNode::execution_space::concurrency() = "<<Kokkos::Compat::KokkosOpenMPWrapperNode::execution_space::concurrency()<<endl
packages/muelu/test/scaling/TwoMatrixMMKernelDriver.cpp:557:    out<< "Kokkos::Compat::KokkosOpenMPWrapperNode::execution_space::concurrency() = "<<Kokkos::Compat::KokkosOpenMPWrapperNode::execution_space::concurrency()<<endl
packages/sacado/test/performance/advection/advection_hierarchical_dfad.cpp:148:  size_t nkernels = ExecSpace::concurrency()*2;
packages/sacado/test/performance/advection_const_basis/advection_hierarchical_dfad.cpp:149:  size_t nkernels = ExecSpace::concurrency()*2;
packages/sacado/test/performance/fenl_assembly/TestAssembly.cpp:137:                << "Threads performance with " << Kokkos::Threads::concurrency()
packages/sacado/test/performance/fenl_assembly/TestAssembly.cpp:150:                << "OpenMP performance with " << Kokkos::OpenMP::concurrency()
[...]

PR #11348, opened 2022-12-05
Amesos2,Sacado: Update calls to concurrency calls to non-static member functions

ndellingwood:
@trilinos/amesos2 
@trilinos/sacado 

## Motivation

Update concurrency call to non-static usage
Address issue #11331 for compatibility with kokkos/kokkos#5655

## Testing

AT will handle the testing

### DIFF
--- a/packages/amesos2/src/Amesos2_Superlumt_def.hpp
+++ b/packages/amesos2/src/Amesos2_Superlumt_def.hpp
@@ -462,7 +462,7 @@ namespace Amesos2 {
 
     RCP<const Teuchos::ParameterList> valid_params = getValidParameters_impl();
 
-    int default_nprocs = Kokkos::DefaultHostExecutionSpace::concurrency();
+    int default_nprocs = Kokkos::DefaultHostExecutionSpace().concurrency();
     data_.options.nprocs = parameterList->get<int>("nprocs", default_nprocs);
 
     data_.options.trans = this->control_.useTranspose_ ? SLUMT::TRANS : SLUMT::NOTRANS;
@@ -521,7 +521,7 @@ namespace Amesos2 {
     if( is_null(valid_params) ){
       Teuchos::RCP<Teuchos::ParameterList> pl = Teuchos::parameterList();
 
-      int default_nprocs = Kokkos::DefaultHostExecutionSpace::concurrency();
+      int default_nprocs = Kokkos::DefaultHostExecutionSpace().concurrency();
       Teuchos::RCP<EnhancedNumberValidator<int> > nprocs_validator
         = Teuchos::rcp( new EnhancedNumberValidator<int>() );
       nprocs_validator->setMin(default_nprocs);
--- a/packages/sacado/test/performance/advection/advection_hierarchical_dfad.cpp
+++ b/packages/sacado/test/performance/advection/advection_hierarchical_dfad.cpp
@@ -145,7 +145,7 @@ double time_dfad_hierarchical_team(int ncells, int num_basis, int num_points,
   // Include a fudge factor of 1.2 since memory pool treats a block as full
   // once it reaches 80% capacity
   const size_t block_size = N*sizeof(double);
-  size_t nkernels = ExecSpace::concurrency()*2;
+  size_t nkernels = ExecSpace().concurrency()*2;
   if (is_cuda_space<ExecSpace>::value)
     nkernels /= 32;
   const size_t mem_pool_size = static_cast<size_t>(1.2*nkernels*block_size);
--- a/packages/sacado/test/performance/advection_const_basis/advection_hierarchical_dfad.cpp
+++ b/packages/sacado/test/performance/advection_const_basis/advection_hierarchical_dfad.cpp
@@ -146,7 +146,7 @@ double time_dfad_hierarchical_team(int ncells, int num_basis, int num_points,
   // Include a fudge factor of 1.2 since memory pool treats a block as full
   // once it reaches 80% capacity
   const size_t block_size = N*sizeof(double);
-  size_t nkernels = ExecSpace::concurrency()*2;
+  size_t nkernels = ExecSpace().concurrency()*2;
   if (is_cuda_space<ExecSpace>::value)
     nkernels /= 32;
   const size_t mem_pool_size = static_cast<size_t>(1.2*nkernels*block_size);
--- a/packages/sacado/test/performance/fenl_assembly/TestAssembly.cpp
+++ b/packages/sacado/test/performance/fenl_assembly/TestAssembly.cpp
@@ -134,7 +134,7 @@ int main(int argc, char *argv[])
       typedef Kokkos::Threads Device;
 
       std::cout << std::endl
-                << "Threads performance with " << Kokkos::Threads::concurrency()
+                << "Threads performance with " << Kokkos::Threads().concurrency()
                 << " threads:" << std::endl;
 
       performance_test_driver<Device>(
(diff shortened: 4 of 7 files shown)


etphipp (approved):
Sacado changes look good.  Thanks @ndellingwood!

iyamazaki (approved):
Thank you, @ndellingwood, for the Amesos2 changes!!

ndellingwood:
The failure with Trilinos_PR_cuda-11.4.2-uvm-off is test failure in ` PanzerAdaptersSTK_MixedCurlLaplacianExample-ConvTest-Tri-Order-1` but the reason is unclear, maybe a timeout?

This is the tail of the output:
```
...
TEST_1: Result = PASSED

================================================================================

TEST_2

Running: "/projects/sems/install/rhel7-x86_64/sems/tpl/openmpi/4.0.5/gcc/10.1.0/base/bxejupk/bin/mpiexec" "-np" "4" "-map-by" "socket:PE=4" "/scratch/trilinos/workspace/Trilinos_PR_cuda-11.4.2-uvm-off@2/pull_request_test/packages/panzer/adapters-stk/example/MixedCurlLaplacianExample/PanzerAdaptersSTK_MixedCurlLaplacianExample.exe" "--use-tpetra" "--use-twod" "--cell=Tri" "--x-elements=32" "--y-elements=32" "--z-elements=4" "--basis-order=1"

  Writing output to file "/scratch/trilinos/workspace/Trilinos_PR_cuda-11.4.2-uvm-off@2/pull_request_test/packages/panzer/adapters-stk/example/MixedCurlLaplacianExample/MPE-ConvTest-Tri-1-32"
```

No further output following the write to the output file statement.
There are various output messages in the test 1 output,
```
[ascicgpu050][[18736,1],1][btl_tcp.c:559:mca_btl_tcp_recv_blocking] recv(49) failed: Connection reset by peer (104)
```
but I'm assuming this is system output that is ignored since test_1 reported as PASSED


@trilinos/framework has the `PanzerAdaptersSTK_MixedCurlLaplacianExample-ConvTest-Tri-Order-1` test posed any spurious failures in other Cuda PR testing?

The changes in this PR should have had no impact on the Panzer test in a Cuda build - the Amesos2 change is for an OpenMP-based solver that is not part of PR testing, and the Sacado changes were only made to performance tests, no source code files used by other packages

ndellingwood:
Thanks for the update @sebrowne !

> Being able to re-run just this step (CUDA11) as opposed to the whole autotester would be another band-aid for the issue, but we can't do that at the moment.

This will be useful down the road :) In the meantime, would rerunning tests that fail with e.g. ctest option `--repeat until-pass:<n>` with n=2 or so be an option?


> I don't fully understand why the parallel level would be causing this, because only one test is run at a time with CUDA builds (though there may have been other tests using this same GPU from another autotester instance). If it is GPU load related, this does tie into work that I've got assigned for this sprint, and may cease to be a problem soon.

If other AT tests run on the same GPU this has potential for contention for resources that can have a pretty nasty impact on runtime performance (I've seen tests that typically run in e.g. 30 sec jump up to a 1500 sec timeout due to resource competition).
In case useful, here is a link of a comparison of potential impact on test runtime when running tests with parallelism v.s. individually https://github.com/trilinos/Trilinos/pull/10253#issuecomment-1085034993 ; that link referenced testing for `ctest -j 4` vs -j 1, not the same situation as described here, but may give a sense of possible impacts on runtime

sebrowne:
> If other AT tests run on the same GPU this has potential for contention for resources that can have a pretty nasty impact on runtime performance (I've seen tests that typically run in e.g. 30 sec jump up to a 1500 sec timeout due to resource competition).

Okay, if that's the case then that's most likely what is happening (I confirmed this morning that two autotester runs on the GPU PR machine were both running tests only on GPU 0, even though there are 4 GPUs on the machines.  This is something I want to get resolved ASAP and I'm going to attempt to get to it this sprint since the CUDA builds are the most expensive PR build right now.

There's a good chance that the j4 vs j1 table you linked to was caused by the exact same problem, only within a given ctest.  It can be tricky to tell individual test instances to use the correct number and indices of GPUs.

How do you want to proceed in the meantime?  If you want to disable the test, I can push up a change that does that and you can pull it in here if you want.  If you're okay retrying without the disable we can do that too.

ndellingwood:
> If you're okay retrying without the disable we can do that too.

@sebrowne I think this sounds best, I agree with your reasoning to hold off on disabling tests (if the same test shows up as a timeout in other PRs and is definitely problematic then I'd request to have it disabled). Thanks for the updates. Should I add the RETEST label, or will the PR automatically retest?

ndellingwood:
@sebrowne an extra mechanism that may help for controlling which GPU tests run on is to use the environment variable `KOKKOS_DEVICE_ID`. Not sure if you can use a build_number % num_gpus to help with assigning a built to a GPU?

rppawlo:
Some of the panzer tests are pretty gpu intensive and can really slow down if other tests are hitting the same gpus. To get ctest to honor multiple devices, see:

https://tribits.org/doc/TribitsBuildReference.html#spreading-out-and-limiting-tests-running-on-gpus

We have used this successfully in empire. In empire we found that overloading each gpu with three tests is the sweet spot. But maybe we don't want to overload to help this panzer test run faster. If this doesn't fix the issue, we could try setting the panzer test to run by itself in ctest:

https://tribits.org/doc/TribitsBuildReference.html#set-specific-tests-to-run-in-serial

I would try to avoid ever using```--repeat until-pass:<n>```. That would allow users to inject code with race conditions into develop. You would have to be absolutely sure it has nothing to do with your PR. Would prefer we not allow that path. If the suggestions above don't fix the issue, please disable the panzer test and I can take a look at speeding up the test on cuda. Maybe there is a kernel running on host that we missed.

sebrowne:
> https://tribits.org/doc/TribitsBuildReference.html#spreading-out-and-limiting-tests-running-on-gpus
Yes, this is exactly what I wanted to use, but I wanted to verify that it does what I think it does (hence my meeting with @bartlettroscoe later this week).  It appears to work pretty much as I expect compared to the vanilla CMake feature that I've used with one of my other codes.  My concern is getting the individual autotester instances to know about each other and use a different GPU (or more realistically, tell each instance which GPU to use, which is much easier since we know how many instances there are, so I could feasibly hardcode instance 0 to tell its test job to use GPU 0, 1 to use 1, etc.).

bartlettroscoe:
> My concern is getting the individual autotester instances to know about each other and use a different GPU (or more realistically, tell each instance which GPU to use, which is much easier since we know how many instances there are, so I could feasibly hardcode instance 0 to tell its test job to use GPU 0, 1 to use 1, etc.).

@sebrowne, that is a use case that CTest does not support.  You need a single CTest instance owning the entire node.  You can't use two or more CTest instances running on the same node and expecting them to not step on each other's toes when running on the GPUs associated with the node.

We can talk more tomorrow in case I am misunderstanding what you are suggesting.

sebrowne:
>@sebrowne, that is a use case that CTest does not support. You need a single CTest instance owning the entire node. You can't use two or more CTest instances running on the same node and expecting them to not step on each other's toes when running on the GPUs associated with the node.

Agreed.  It would have to take place somewhere above ctest and then telling it which GPUs it can farm jobs to (effectively splitting up the node with respect to the GPU resources).  This still has the disadvantage of not being able to share a GPU between two instances in such a way that the slot count is still respected (2 ranks from instance 1, 2 ranks from instance 2).  However, it would allow me to say "instance 1 use only GPU 0" and "instance 2 use only GPU 1", which is a whole lot better than what it's doing now (both instance 1 and instance 2 only use GPU 0!).

Issue #11349, opened 2022-12-05
Tpetra: `crsGraph::getLocalOffRankOffsets` silently produces garbage if `isFillComplete` is false


cwpearson:
## Bug Report
@trilinos/tpetra 

### Description

the off-rank offsets is a row-pointer array that points into a sorted CrsGraph to tell where in the row the non-zeros which contain off-rank values begin. It is computed lazily and cached.

This function returns the view by reference. The view allocated without initializing, set to the reference argument, and then then filled with correct values if `isFillComplete` is true. Otherwise...the caller may happily use an appropriately-sized view that was returned by reference, but was not actually filled with anything.

It would be good to error out noisily rather than continue on.

* [ ] Move `offsets = k_offRankOffsets_` into the loop branch where `k_offRankOffsets_` is filled.
* [ ] Add an `else` branch that throws.

PR #11350, opened 2022-12-05
Tpetra: throw when requesting off-rank offsets in non-fill-complete graph

cwpearson:
Attempt to fix https://github.com/trilinos/Trilinos/issues/11349


csiefer2 (approved):

cwpearson:
Why does CDash only show builds from yesterday?

Issue #11356, opened 2022-12-06
Check CXX_STANDARD >= 17 at configuration, error out if condition not met


ndellingwood:
## Enhancement
@trilinos/framework @trilinos/tribits 

I'm not sure which package or team to indicate for the request, so I added framework and tribits to start discussion.

With the updated requirement for c++17 with testing, this issue is to request setting c++17 as the default c++ standard in Trilinos and add a test to check at configuration to confirm this minimum has been met.

A quick check configuring with SHA 940ead2f01e87d7f96134e48d7248a8f2ba71b4c shows the current default is still c++14

The motivation for the request is that the update to c++17 for testing allows newer c++ features into the code base that requires c++17 as a minimum in order to compile (e.g. our eagerly anticipated usage of`if constexpr` , like that included with PR #11338)

When compiling with compilers without c++17 enabled (and that do not support c++17, e.g. cuda/10.1.105) the build could error out (in this case when Stokhos is enabled) though the output message does not make the reason clear:

```
/home/jenkins/weaver-new/workspace/KokkosEco_Trilinos_Weaver_CUDA101_opt/Trilinos/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector.hpp(272): error: expected a "("
```

(The above is not missing a "(" but makes use of `if constexpr` not recognized by the compiler https://github.com/trilinos/Trilinos/blob/3adb30811047b5690c8344679f5933b9962e267d/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector.hpp#L272)

glhenni:
I have a weird error when setting `CMAKE_CXX_STANDARD:STRING=17` and use the intel compiler (version below). exodiff (in the seacas package) hangs for a significant number of our regression tests, but strangely not all. I also see errors thrown by epu, another seacas utility. Am I the only one that sees this?

CDE/v3 modules on CEE/ASCIC, intel compiler + OpenMPI 4.1.2
```
% mpicxx --version
icpc (ICC) 2021.1.2 20201208
Copyright (C) 1985-2020 Intel Corporation.  All rights reserved.
```

sebrowne:
@trilinos/seacas

glhenni:
This may just be a problem in *epu*. It's possible the *exodiff* problem is due to the fact that a previous *epu* invocation for the same problem fails. At any rate, I compiled a debug version and there's an assertion failing:
```
[...]

PR #11357, opened 2022-12-06
Framework: Make C++17 the default

sebrowne:
Also remove support for C++14.

User Support Ticket(s) or Story Referenced:
https://github.com/trilinos/Trilinos/issues/11356

@trilinos/framework

## Motivation
C++17 is required for the upcoming Kokkos integration.

## Related Issues
#11356


sebrowne:
I didn't search exhaustively, but I think this does what @ndellingwood is asking for in #11356 

@ccober6 This is what came up earlier today.  It will break any customers who can't use C++17 effectively immediately (meaning they won't be able to integrate newer versions of Trilinos).

ndellingwood (approved):
Thanks @sebrowne ! Changes look good to me

sebrowne:
@rppawlo I know you asked about the timeline surrounding this yesterday

ndellingwood:
Just a note that without @sebrowne's change here that current builds with Stokhos enabled and an unspecified CMAKE_CXX_STANDARD (resulting in c++14 being enabled) or if explicitly requesting c++14 will fail to compile. A hack that may work around this is something like:

```c++
diff --git a/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector.hpp b/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector.hpp
index ac6543de688..9a55c9c2eea 100644
--- a/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector.hpp
+++ b/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector.hpp
@@ -269,6 +269,7 @@ namespace Sacado {
        */
       KOKKOS_INLINE_FUNCTION
       Vector(std::initializer_list<value_type> l) : s(l.size(), l.begin()) {
+#ifndef KOKKOS_ENABLE_CXX14
         if constexpr (Storage::is_static) {
           const auto         lsz = static_cast<ordinal_type>(l.size());
           const ordinal_type  sz = this->size();
@@ -282,6 +283,7 @@ namespace Sacado {
             }
           }
         }
+#endif
       }
 
       //! Destructor
```

rppawlo:
@sebrowne - you might also want to add 23 to the allowed standards while you are changing this file. It is supported in cmake since 3.20.

sebrowne:
Planning on merging this ~ 01-09-2023

ccober6 (approved):

alanw0:
@sebrowne we put a stk snapshot into develop yesterday which uses c++17 and it broke a bunch of trilinos builds. We were under the impression that c++17 was already legal based on emails which we might have mis-intrepreted?
I guess we should revert that snapshot.
When is c++17 expected to be legal in trilinos?

cgcgcg:
@alanw0 I don't think you need to revert if merging this PR fixes the issues.

sebrowne:
@alanw0 which builds failed?  Once this PR is merged, C++17 is mandatory (and automatic), so said builds will either break for other reasons that aren't your fault, or they will start passing with your changes because they're currently still set to C++14.  Either way, I don't think your changes are an issue.

cgcgcg:
https://testing.sandia.gov/cdash/index.php?project=Trilinos for the dashboard with the failing builds.

sebrowne:
Okay that's kind of what I thought.  No worries @alanw0, we can wait and see what happens once this PR is merged.

alanw0:
> Okay that's kind of what I thought. No worries @alanw0, we can wait and see what happens once this PR is merged.

To be clear, are you suggesting I wait on reverting the stk commit?

sebrowne:
Correct, I do NOT think you need to revert the commit

jhux2:
Since #11461 assumes c++17, shouldn't it have failed PR testing?

sebrowne:
@jhux2 No, because PR testing is explicitly enabling C++17 (this happened during the compiler upgrade in early November).  Since then, we've been in a state where C++17 is fully supported via PR testing, but was not actually required by TriBITS (that is, TriBITS would allow for C++17 to not be enabled, but C++17 features may have been present in the code, leading to exactly what happened here).  We were delaying requiring C++17 until we resolved @glhenni's reported issue in #11356, as well as by request of @rppawlo for EMPIRE to get further along towards C++17 support.

TL;DR: The possibility of this exact situation was known by myself and @ccober6 when we did the autotester change in November, and we accepted that possibility until we merged this PR.  Merging this PR should prevent this issue from now on by enabling C++17 by default.

jhux2:
@sebrowne Got it, thank you for the explanation!

PR #11360, opened 2022-12-07
Actions: Adding a title/label to mention utility for Ifpack2, MueLu

csiefer2:
Because github won't let non-team members do team mentions.

This is the "test" version of this utility, just targeting Ifpack2 and MueLu.  We can add more packages later, assuming it works.

### DIFF
--- a/.github/workflows/title_to_mention.yml
+++ b/.github/workflows/title_to_mention.yml
@@ -18,13 +18,16 @@ jobs:
 
     steps:
     - name: Mention MueLu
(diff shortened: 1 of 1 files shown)


GrahamBenHarper (approved):
It might be overkill once some of the other changes hit master, and turning it off will also be a pain if it doesn't work, but I'm happy to see if it can correctly mention the teams since it's only for a couple packages.

cgcgcg (approved):

PR #11364, opened 2022-12-08
Panzer q2 mesh support

reuterb:
@trilinos/panzer 

## Motivation
Panzer is looking to support higher order meshes for various customers.
As a first step, we are looking to get a lot of the machinery in place to do so, as well as ensure second order mesh support.


reuterb:
@rppawlo Putting this up as a WIP so we can look more closely at some of the changes I'm proposing. Leaving in a lot of incremental commits for now until we are sure we didn't break anything and decide the changes are reasonable. 

As an FYI, I was able to get the mesh topology to the intrepid bases (one of my big concerns from the other day) fairly easily.

packages/panzer/dof-mgr/src/Panzer_IntrepidFieldPattern.cpp
```diff
@@ -257,7 +263,13 @@ namespace panzer {
 
     if(numCells>0) {
       Intrepid2::CellTools<PHX::Device> cellTools;
-      cellTools.mapToPhysicalFrame(coords,localCoords,cellVertices,intrepidBasis_->getBaseCellTopology());
+      // For backwards compatability, allow the FE basis to supply the mesh cell topology (via the FEM base cell topo)
+      // If provided, use the mesh topology directly
+      if (meshCellTopology==Teuchos::null) {
+        cellTools.mapToPhysicalFrame(coords,localCoords,cellNodes,intrepidBasis_->getBaseCellTopology());
```
mperego, line 269: Why don't you pass directly the (HGRAD) basis to the `mapToPhysicalFrame` function?
https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Cell/Intrepid2_CellTools.hpp#L1021-L1024

In fact, when you pass the topology, intrepid2 has to build the corresponding HGRAD basis.

reuterb: Good question. That particular bit was there from before my time, but we should examine. Would you get different behavior passing in a higher Intrepid basis for any reason?

reuterb: Anyway, my thought process was ::

1) If the user has a first-order mesh, then cellNodes will just be cellVertices and to be compatible with computing the map, you'd want a first-order basis in that calculation.
2) If the user has a higher-order mesh, the cellNodes is larger and corresponds with the order of the mesh.

Essentially, in either case, we are using a basis which corresponds to the mesh order to create this map.
If no mesh topology is supplied, we are using the intrepidbasis to determine the mesh topology (which will be the lowest-order topology).

mperego: I see your reasoning. What is the `intrepidBasis`? Is it the basis used for mapping between the ref element and the physical element (e.g., for affine meshes it would be HGRAD_C1, for quadratic meshes it would be HGRAD_C2)? If so, it's better to directly pass the basis, which will also work for higher order (>2) meshes.

reuterb: In this case, `intrepidBasis_` is the FE basis, which at least theoretically could be any order and, generally, does not match the mesh order. As this snippet indicates, we are not explicitly constructing or carrying around an additional intrepid basis specifically for the mapping -- but rather constructing it within `mapToPhysicalFrame`.

If you see this as problematic, I can look into constructing a basis upstream specifically for this purpose.
I guess it may depend on if this function is called many time and a temporary basis keeps getting made over and over.
There may also be some utility to eliminating this if statement and always explicitly requiring the mesh topology (or a "mesh-order" intrepid basis, like you point out).

packages/panzer/disc-fe/src/Panzer_Workset_Builder_impl.hpp
```diff
@@ -111,15 +111,15 @@ panzer::buildWorksets(const WorksetNeeds & needs,
...
-	 RCP<panzer::BasisValues2<double> > bv2
+	       RCP<panzer::BasisValues2<double> > bv2
              = rcp(new panzer::BasisValues2<double>("",true,true));
-	 bv2->setupArrays(b_layout);
-	 i->bases.push_back(bv2);
+	       bv2->setupArrays(b_layout);
+	       i->bases.push_back(bv2);
 
-	 basis_names->push_back(b_layout->name());
+	       basis_names->push_back(b_layout->name());
```
rppawlo, line 122: Why was this indented?

packages/panzer/adapters-stk/example/PoissonExampleCurvilinear/compareWithGold.py
```diff
@@ -0,0 +1,64 @@
...
+      if match!=None:
+        q2_errors.append(float(match.group(1)))
+  with open(q1test) as f:
+    for line in f:
+      match = ErrorPat.match(line)
+      if match!=None:
+        q1_errors.append(float(match.group(1)))
+
+  for i in range(len(q2gold_errors)):
+    if q2gold_errors[i] != q2_errors[i]:
```
rppawlo, line 44: I think these comparisons of floating point numbers should use a tolerance. What is the expected accuracy?

rppawlo (changes requested):
@reuterb - looks good. Just have a few questions.

packages/panzer/adapters-stk/example/PoissonExampleCurvilinear/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+
```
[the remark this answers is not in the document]
reuterb: If we don't want to check the L2/H1 error, this can be easily incorporated into main_driver.
A better short term solution may be to modify PoissonExample to always do the area calculation, allow for a mesh to be supplied, but grab different reference solutions depending on the test case.
These examples probably require a bit of an overhaul at some point. We should discuss.

packages/panzer/disc-fe/src/Panzer_Workset_Builder_impl.hpp
reuterb: This is not showing up in my local copy... not sure what's going on.

packages/panzer/adapters-stk/example/PoissonExampleCurvilinear/compareWithGold.py
reuterb: This is a straight regression test -- I'll add a tolerance of 1e-14.

terwin:
@reuterb - It looks like there's an assumption of linear basis in [adapters-stk/src/Panzer_STK_IOClosureModel_Factory.cpp:161](https://github.com/reuterb/Trilinos/blob/0a346ffaac44f16f966291c64296de58b275b4d9/packages/panzer/adapters-stk/src/Panzer_STK_IOClosureModel_Factory.cpp#L161) that prevents output of nodal fields on Q2 meshes.

reuterb:
> @reuterb - It looks like there's an assumption of linear basis in [adapters-stk/src/Panzer_STK_IOClosureModel_Factory.cpp:161](https://github.com/reuterb/Trilinos/blob/0a346ffaac44f16f966291c64296de58b275b4d9/packages/panzer/adapters-stk/src/Panzer_STK_IOClosureModel_Factory.cpp#L161) that prevents output of nodal fields on Q2 meshes.

Hrm, I will take a closer look. So you are not able to output at all? On my end I was able to output to the Q2 nodes -- although this suggests the output I was seeing wasn't actually high order. Thanks for pointing this out @terwin.  The I/O is due for a bigger overhaul but we should at least get this case correct short-term.

terwin:
> > @reuterb - It looks like there's an assumption of linear basis in [adapters-stk/src/Panzer_STK_IOClosureModel_Factory.cpp:161](https://github.com/reuterb/Trilinos/blob/0a346ffaac44f16f966291c64296de58b275b4d9/packages/panzer/adapters-stk/src/Panzer_STK_IOClosureModel_Factory.cpp#L161) that prevents output of nodal fields on Q2 meshes.
> 
> Hrm, I will take a closer look. So you are not able to output at all? On my end I was able to output to the Q2 nodes -- although this suggests the output I was seeing wasn't actually high order. Thanks for pointing this out @terwin. The I/O is due for a bigger overhaul but we should at least get this case correct short-term.

We can output the nodal fields of the primary solution variables. It looks like the `ScatterFields` evaluator for those is created in [adapters-stk/src/responses/Panzer_STK_ResponseEvaluatorFactory_SolutionWriter_impl.hpp:169](https://github.com/reuterb/Trilinos/blob/0a346ffaac44f16f966291c64296de58b275b4d9/packages/panzer/adapters-stk/src/responses/Panzer_STK_ResponseEvaluatorFactory_SolutionWriter_impl.hpp#L169) using the correct basis. The issue arises when trying to output additional nodal fields, including time derivatives of the primary solution fields.

reuterb:
@terwin Was the error you were getting something like: `ERROR: The required field "MYFIELD:double:<C,B>(20,4)" does not have an evaluator.` ?
If not, let me know what error you were seeing.

Would you ever run higher order elements on a first order mesh? If so, STKIO responses might also fail in a similar manner.

terwin:
> @terwin Was the error you were getting something like: `ERROR: The required field "MYFIELD:double:<C,B>(20,4)" does not have an evaluator.` ? If not, let me know what error you were seeing.
> 
> Would you ever run higher order elements on a first order mesh? If so, STKIO responses might also fail in a similar manner.

Yes, that's the error I get. In your example, I _would_ have an evaluator for `MYFIELD:double:<C,B>(20,9)` (Quad9 basis).

We have been, and will continue to, run high order on linear meshes. I just checked, and I do get this error in that case as well. Apparently we had not tried to write a high order nodal field (STKIO response) until now.

reuterb:
> Yes, that's the error I get. In your example, I would have an evaluator for MYFIELD:double:<C,B>(20,9) (Quad9 basis).
>
> We have been, and will continue to, run high order on linear meshes. I just checked, and I do get this error in that case as well. Apparently we had not tried to write a high order nodal field (STKIO response) until now.

OK, this is consistent with my current understanding so that is good.

A plausible solution is constructing a basis consistent with the mesh order and projecting onto that before output. I need to assess how much of a lift that is versus an easier but less robust solution that could at least keep things moving before a more thorough IO overhaul.

reuterb:
@rppawlo Please review, thanks!

rppawlo (approved):

packages/panzer/adapters-stk/example/PoissonExample/compareWithGold_Curvilinear.py
```diff
@@ -0,0 +1,66 @@
...
+  q1test = sys.argv[4]
+
+  ErrorPat = re.compile('.*Error = (.*)$')
+
+  q2gold_errors = []
+  q1gold_errors = []
+  q2_errors = []
+  q1_errors = []
+
+  tol = 1e-14
```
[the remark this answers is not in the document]
reuterb: Good point, I will change.

reuterb:
@rppawlo Changed that one tolerance. Please re-approve.

rppawlo (dismissed):

reuterb:
@rppawlo Are we OK to go ahead and merge here?

reuterb:
> @reuterb - I don't see the commit to add the TEMPORARY define to the config file in this commit. Did you put that in a different PR? It should probably be in this branch before we merge.

Thanks, just committed something that should address this.

rppawlo (approved):

reuterb:
@rppawlo Please re-approve when you have a chance. I needed to merge the develop branch in and correct one thing in the test since the L2 projection stuff won the race.

rppawlo (approved):

Issue #11361, opened 2022-12-07
Framework: random failures blocking merge of PR


jhux2:
## Bug Report
@trilinos/framework 

### Description
There are a variety of random failures that are blocking the merge of PR #11341.  No of these are due to the PR itself, since it touches a Python utility that doesn't get exercised by the AT.

Reported in https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-235.

GrahamBenHarper:
I believe other PRs have recently run into the problem of running out of memory, but I didn't mark which ones since they were fixed with a retest. If I find more instances, I'll make sure to link them here as related.

GrahamBenHarper:
https://github.com/trilinos/Trilinos/pull/11330#issuecomment-1338009742 is another example of a job killed by (presumably) running out of memory. I found it by skimming a handful of PR builds which have exactly two errors as shown in [this cdash query](https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-&field2=buildstarttime&compare2=84&value2=NOW&field3=buildstarttime&compare3=83&value3=Nov%2001%2C%202022).

sebrowne:
The Intel build running out of memory is fixed (not really, but we turned the Intel build back off so a re-run will bypass the failure until we get the correct level of build parallelism figured out such that it won't overload the machine).

The MiniTensor test from earlier in that PR's testing history is fixed now.

The Domi test timed out and was on a CUDA build.  That is another thing we have immediate work planned for (hopefully by the end of the week) to avoid due to GPU oversubscription.

Framework empathizes with the difficulty that these nondeterministic results is causing, and we're working towards resolution on each of the issues as fast as we can.  I will update this once we get the CUDA changes in and the Intel build turned back on, if that's satisfactory.

jhux2:
no longer an issue

PR #11369, opened 2022-12-08
Automatically assign all GPUs on machine to tester

sebrowne:
Scan a machine for NVidia GPUs, and if found, assign them all to the ctest instance with 2 slots each.

This will be coupled with some rearrangement of the PR pool such that each GPU machine only runs one GPU PR at a time, and nothing else.

@trilinos/framework 

## Motivation
Want more robust test results, and faster ones too.  Seeing repeated flaky testing on the GPU PR runs and suspect it's due to multiple autotester instances using the same GPU.  I've monitored this and there can be as many as 8 MPI ranks communicating with GPU 0, but all of the others are always idle.

## Related Issues
#11361 

## Testing
Framework unit tests


jmlapre (dismissed):
Looks good to me.

sebrowne:
Now that each PR CUDA build gets exclusive access to a GPU machine, this should be safely testable.

sebrowne:
Closing in lieu of #11391

PR #11382, opened 2022-12-14
Increase CUDA PR build jobs to 96

sebrowne:
Each CUDA PR build now has exclusive use of a 48 core GPU machine with plenty of memory, so turn the parallelism up for all of them.

User Support Ticket(s) or Story Referenced: TRILFRAME-522

I don't really like doing this in a file that says "do it for every CUDA build".  I'd rather encode it into the autotester configs, but for now this seems to be where it's done.

@trilinos/framework 

## Motivation
Want faster CUDA PR turnaround (it and Intel are very long).


e10harvey (dismissed):

sebrowne:
Closing in lieu of https://github.com/trilinos/Trilinos/pull/11391

PR #10921, opened 2022-08-23
Teko: Make Epetra optional for Teko package

JacobDomagala:
There're lots of whitespace changes in this PR (removed trailing whitespaces), for easier reviewing I'd suggest turning whitespace changes off in GitHub.
image

PR #11383, opened 2022-12-14
Panzer: make epetra optional in stk adapters and mini-em

rppawlo:
## Motivation

make epetra stack optional for panzer

## Testing

N/A


rppawlo:
To actually be able to disable the epetra stack in panzer, we will also need #10921 to be merged for teko.

cgcgcg (approved):
Looks good to me. Thanks!

rppawlo:
@GrahamBenHarper - unfortunately, we need to keep the epetra stack in panzer for the next year or so. We have some important applications still relying on the epetra stack. Some panzer capabilities were never ported to the tpetra  stack. Once we get those implemented, we can drop epetra.

PR #11386, opened 2022-12-14
ccache preliminary support

jmlapre:
@trilinos/framework 

## Motivation
Trilinos is a massive codebase often building for hours at a time.  However, changes are often incremental and fairly redundant.  Adding [ccache](https://ccache.dev/) would drastically reduce the amount of time required per-build. This PR adds some minimal supporting changes and somehow fell off our internal boards unmerged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,32 @@ INCLUDE(${CMAKE_CURRENT_LIST_DIR}/ProjectName.cmake)
 # not in an include file :-(
 PROJECT(${PROJECT_NAME} NONE)
 
+find_program(CCACHE_PROGRAM ccache)
+if(CCACHE_PROGRAM)
+    # Set up wrapper scripts
+    set(C_LAUNCHER   "${CCACHE_PROGRAM}")
+    set(CXX_LAUNCHER "${CCACHE_PROGRAM}")
+    configure_file(launch-c.in   launch-c)
+    configure_file(launch-cxx.in launch-cxx)
+    execute_process(COMMAND chmod a+rx
+                     "${CMAKE_BINARY_DIR}/launch-c"
+                     "${CMAKE_BINARY_DIR}/launch-cxx"
+    )
+
+    if(CMAKE_GENERATOR STREQUAL "Xcode")
+        # Set Xcode project attributes to route compilation and linking
+        # through our scripts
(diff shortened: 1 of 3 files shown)


sebrowne:
Adding this to the configure command does effectively the same thing:
`-DCMAKE_C_COMPILER_LAUNCHER=ccache
-DCMAKE_CXX_COMPILER_LAUNCHER=ccache`

Is there  a reason to do it with wrappers instead?

jmlapre:
They add support for Xcode and also define `CCACHE_CPP2`.  But if we are OK with cutting Xcode then we could simply define those variables as well as `CCACHE_CPP2` and call it a day.

CMakeLists.txt
```diff
@@ -69,6 +69,35 @@ INCLUDE(${CMAKE_CURRENT_LIST_DIR}/ProjectName.cmake)
 # not in an include file :-(
 PROJECT(${PROJECT_NAME} NONE)
 
+if("${CCACHE_NODISABLE}")
```
sebrowne, line 72: I think this is more descriptive as `$ENV{CCACHE_NODISABLE}`

CMakeLists.txt
```diff
@@ -69,6 +69,35 @@ INCLUDE(${CMAKE_CURRENT_LIST_DIR}/ProjectName.cmake)
...
+        set(CMAKE_XCODE_ATTRIBUTE_CC         "${CMAKE_BINARY_DIR}/launch-c")
+        set(CMAKE_XCODE_ATTRIBUTE_CXX        "${CMAKE_BINARY_DIR}/launch-cxx")
+        set(CMAKE_XCODE_ATTRIBUTE_LD         "${CMAKE_BINARY_DIR}/launch-c")
+        set(CMAKE_XCODE_ATTRIBUTE_LDPLUSPLUS "${CMAKE_BINARY_DIR}/launch-cxx")
+    else()
+        # Support Unix Makefiles and Ninja
+        set(CMAKE_C_COMPILER_LAUNCHER   "${CMAKE_BINARY_DIR}/launch-c")
+        set(CMAKE_CXX_COMPILER_LAUNCHER "${CMAKE_BINARY_DIR}/launch-cxx")
+    endif()
+ endif()
```
sebrowne, line 98: If ccache was requested but not found, I'd like to get a message about it

sebrowne (changes requested):

sebrowne (approved):

PR #11387, opened 2022-12-14
WIP MueLu: Remove Epetra support

cgcgcg:

## Motivation
If merged, this PR would do the following:
- Remove `MueLu_ENABLE_Kokkos_Refactor` and instead always build the Kokkos version of factories. (The parameter `use kokkos refactor` is not affected by this.)
- Remove `Xpetra_ENABLE_Kokkos_Refactor` and make Tpetra a hard dependency for Xpetra. (Needed, as we rely on `Tpetra::LocalMap`.)
- Remove Epetra, Amesos, Ifpack, AztecOO from MueLu. Make Tpetra, Ifpack2 and Amesos2 hard dependencies.

The first item could probably be merged soon. The remainder requires a bit more discussion. But I want to see if this builds and runs across multiple configurations of the AT. 

What's still missing / to be discussed:
- Xpetra still has the optional Epetra dependency.
- There are lots of leftovers and mentions of Epetra in MueLu.
- A few MueLu tests don't have a Tpetra version.
- Muemex


packages/muelu/matlab/src/MueLu_MatlabUtils_decl.hpp
```diff
@@ -49,173 +49,5 @@
 
 #include "MueLu_ConfigDefs.hpp"
 
-#if !defined(HAVE_MUELU_MATLAB) || !defined(HAVE_MUELU_EPETRA) || !defined(HAVE_MUELU_TPETRA)
 #error "Muemex requires MATLAB, Epetra and Tpetra."
```
GrahamBenHarper, line 52: Hm. Does this break MueMex?

cgcgcg: Probably

csiefer2:
Do we still have algorithms that aren't in the Kokkos factory path?

cgcgcg:
Cut-drop has not Kokkos factory path. I'm sure there are others. I'm not removing non-Kokkos factories.

mayrmt:
@cgcgcg What's the timeline for this effort? Our code still uses Epetra and we are in the process of migrating to Tpetra. Just want to get a feeling, if we need to hurry or still have some time to complete the migration.

cgcgcg:
@mayrmt My guess is you still have some time. I don't think this will get merged anytime soon.

PR #11391, opened 2022-12-15
Framework: Speed up CUDA PR builds/tests

sebrowne:
Scan a machine for NVidia GPUs, and if found, assign them all to the ctest instance with 2 slots each.

This will be coupled with some rearrangement of the PR pool such that each GPU machine only runs one GPU PR at a time, and nothing else.

@trilinos/framework 

## Motivation
Want more robust test results, and faster ones too. Seeing repeated flaky testing on the GPU PR runs and suspect it's due to multiple autotester instances using the same GPU. I've monitored this and there can be as many as 8 MPI ranks communicating with GPU 0, but all of the others are always idle.

## Related Issues
#11361 
[TRILFRAME-522](https://sems-atlassian-son.sandia.gov/jira/browse/TRILFRAME-522)


sebrowne:
@trilinos/framework can somebody give me a review on this?

jmlapre (approved):
Looks good.

sebrowne:
Had two tests that do not look like resource issues to me.  One was a CUDA illegal memory access error (which I can't speak to, but I'm deeply suspicious declaring that the GPU was out of memory.  I have a cron job monitoring that which only gives me per-minute resolution, but I never saw it get above 24% on any one of the four GPUs).  The other one _appears_ to be a legitimate failure in `KokkosCore_UnitTest_CudaInterOpStreams_MPI_1` (see below)

[ RUN      ] cuda.raw_cuda_streams
Warning: environment variable 'KOKKOS_NUM_DEVICES' is deprecated. Use 'KOKKOS_MAP_DEVICE_ID_BY=mpi_rank' instead. Raised by Kokkos::initialize().
<snip>/packages/kokkos/core/unit_test/cuda/TestCuda_InterOp_Streams.cpp:73: Failure
Expected equality of these values:
  600
  sum
    Which is: 0
[  FAILED  ] cuda.raw_cuda_streams (473 ms)
[----------] 1 test from cuda (473 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (473 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] cuda.raw_cuda_streams


The good news is that this test run only took 47 minutes!!!!  Since it incorporates a framework change, it ran everything.  Normally that has been taking ~2hr 45min on other framework changes.

The build was still very slow at ~2hr 36min, so I'm going to investigate why even with 96 cores and 500GB of memory the machine built Trilinos so slowly.

ndellingwood:
@sebrowne the Kokkos failure is legitimate, for reference reported as kokkos/kokkos#5632 and resolved on kokkos develop branch with kokkos/kokkos#5634. The bug fix didn't make it into Trilinos for the 3.7.01 patch release, we will include with the 3.7.02 patch release.
If the test failure is impeding progress on your work I can create a patch directly to Trilinos, or if you can temporarily disable the test in the Cuda builds until 3.7.02 makes it into Trilinos?

sebrowne:
@ndellingwood this is very interesting.  It's not a big deal right now, because the whole point of this PR is to start assigning tests to the individual GPUs more efficiently (meaning that we don't now, so that's why we never saw the failure in the current PR testing).  So until this gets merged, the test will still pass for other PRs.  What is the anticipated timeline on getting 3.7.02 in?

sebrowne:
I'm also willing to entertain waiting until the 4.0 release, unless you're planning on a 3.7.02 anyway

sebrowne:
Waiting on #11600, then should be able to merge this.

sebrowne:
Retesting now that #11600 is merged

bartlettroscoe:
@sebrowne, just curious, but why is this PR writing its own test resource file when it is only two cache variables to set up the resource file:

* `Trilinos_CUDA_NUM_GPUS`
* `Trilinos_CUDA_SLOTS_PER_GPU`

Is there some limitation to the ctest resource file that TriBITS is already writing and using automatically here?  Otherwise, this seems like extra code that duplicates what is already present.

Otherwise, yes, you can't expect `ctest -S` to effectively run in multiple instances on the same node with shared GPUs and not stomp on the same GPUs by different `ctest -S` instances.  That would require and extension to CTest similar to what they implemented for core load limiting across an entire machine.  See `TEST_LOAD <threashold>` in:

* https://cmake.org/cmake/help/v3.23/command/ctest_test.html#command:ctest_test
 
But as far as I know, nothing like that exists for limiting across multiple CTest invocations based on the test resource file.  (But we could ask Kitware to implement that feature if we think it is important enough.)

Otherwise, I think the only way to address this is to only run one CUDA build on these machines with GPUs at one time.  (But you could run some non-CUDA builds to use some of the idle CPU cores.)

sebrowne:
@bartlettroscoe so the issue with multiple PR runs per machine has been resolved (outside of this PR).  I changed the Jenkins configurations to only run a single GPU build/test per GPU machine, which immediately helped the issue by overloading the GPUs to a lesser degree than before.  We have enough GPU machines to tolerate that for now.

With respect to why I didn't use the TriBITS capability, it's because I wanted it to automatically detect and spread across all GPUs on the machine.  The writing of the file is certainly duplicated.  Looking back at this, the issue was that I can't put executable code in config-specs.ini, only values, so I handled it in the PR scripts.  However, what I CAN do is instead of writing the file, feed the numerical value that I generate to TriBITS via the aforementioned variables.  That would reduce duplication and still allow me to not hardcode the number of GPUs in the configuration (for example, our Volta machines have 4 GPUs each, but the Ampere ones only have 2 each).

sebrowne:
Retesting to have a current CMakeCache.txt to investigate

packages/framework/pr_tools/trilinosprhelpers/TrilinosPRConfigurationStandard.py
```diff
@@ -83,7 +83,7 @@ def execute_test(self):
             gpu_indices = gpu_utils.list_nvidia_gpus()
             self.message(f"-- REMARK: Using {slots_per_gpu} slots per GPU")
             self.message(f"-- REMARK: Using GPUs {gpu_indices}")
-            cmd.append(f"-DEXTRA_CONFIGURE_ARGS:STRING=\"-DTrilinos_AUTOGENERATE_TEST_RESOURCE_FILE:BOOL=ON;-DTrilinos_CUDA_NUM_GPUS:STRING={len(gpu_indices)};-DTrilinos_CUDA_SLOTS_PER_GPU:STRING={slots_per_gpu}\"")
+            cmd.append(f"-DEXTRA_CONFIGURE_ARGS:STRING=-DTrilinos_AUTOGENERATE_TEST_RESOURCE_FILE:BOOL=ON; -DTrilinos_CUDA_NUM_GPUS:STRING={len(gpu_indices)}; -DTrilinos_CUDA_SLOTS_PER_GPU:STRING={slots_per_gpu}")
```
sebrowne, line 86: Excellent call.  I missed that `subprocess` is not being used in shell mode, which means that the list gets passed directly to it instead of concatenated into a string, and it handles the different arguments itself with no need for quoting.

ndellingwood:
`Trilinos_PR_intel-2021.3` test had two compilation errors:

`Intrepid2`:
[Error building packages/intrepid2/unit-test/Projection/CMakeFiles/Intrepid2_unit-test_Projection_test_convergence_HEX_Serial_DOUBLE.dir/test_convergence_HEX_Serial_DOUBLE.cpp.o](https://trilinos-cdash.sandia.gov/viewBuildError.php?buildid=1273724#)
```
compilation aborted for packages/intrepid2/unit-test/Projection/test_convergence_HEX_Serial_DOUBLE.cpp (code 1)
```
* This indicates the compiler ran out of memory

`TrilinosCouplings` error - fallout from the intrepid2 error above

I'm going to attempt to retest but worried the `icpc: error #10106: Fatal error` will occur again unless the amount of build parallelism is not throttled down. @trilinos/framework @sebrowne if this intel error keeps occurring, can the level of parallelism at the build stage be reduced for this PR build (that is, reduce 'N' in e.g. `make -j N`)?

ndellingwood:
Oops, I thought I was looking at results for a different PR. I deleted the post, but maybe it can be helpful (even though I thought this applied to a different PR), so I'll repost below:

`Trilinos_PR_intel-2021.3` test had two compilation errors:

`Intrepid2`:
[Error building packages/intrepid2/unit-test/Projection/CMakeFiles/Intrepid2_unit-test_Projection_test_convergence_HEX_Serial_DOUBLE.dir/test_convergence_HEX_Serial_DOUBLE.cpp.o](https://trilinos-cdash.sandia.gov/viewBuildError.php?buildid=1273724#)

compilation aborted for packages/intrepid2/unit-test/Projection/test_convergence_HEX_Serial_DOUBLE.cpp (code 1)

* This indicates the compiler ran out of memory

`TrilinosCouplings` error - fallout from the intrepid2 error above

~~I'm going to attempt to retest~~ (Edit: I won't mess with the workflow of this PR, sorry for the noise!) but worried the `icpc: error #10106: Fatal error` will occur again unless the amount of build parallelism is not throttled down. @trilinos/framework @sebrowne if this intel error keeps occurring, can the level of parallelism at the build stage be reduced for this PR build (that is, reduce 'N' in e.g. `make -j N`)?

achauphan:
Tests are able to utilize multiple GPUs now based on the number of them available on the machine they're running on with a current set max of 2 ranks per GPU. This last GPU test run took about 30mins! However, a few tests did fail...

@ndellingwood @alanw0 with the context of multiple GPUs being used, do you have any insight of why `Kokkos_CoreUnitTest_Cuda1_MPI_1` or `STKUnit_tests_stk_util_unit_tests_MPI_4` may be failing?

ndellingwood:
Also, the environment variable `KOKKOS_MAP_DEVICE_ID_BY=<mpi_rank>` may need to be set if not being used. If `KOKKOS_NUM_DEVICES` is used, consider replacing it with `KOKKOS_MAP_DEVICE_ID_BY` as `KOKKOS_NUM_DEVICES` was deprecated before the 4.0 release

alanw0:
I had Dong Hun (@ldh4) take a look at the stk-util unit-test errors. He thinks those errors are coming from the kokkos side. Perhaps they will be resolved by the same suggestions that @ndellingwood mentions above.

sebrowne:
@ndellingwood we're using a CTest resource file (that is being written by TriBITS) to do the GPU rank layout.  We're not intentionally setting `KOKKOS_NUM_DEVICES`.  I quickly grepped through the source code and it appears that there is a different mechanism for setting up the GPU layout if it is using CTest (see https://github.com/kokkos/kokkos/blob/71a9bcae52543bd065522bf3e41b5bfa467d8015/core/src/impl/Kokkos_Core.cpp#L227).  Should we need to do anything in addition to that?

ndellingwood:
> Should we need to do anything in addition to that?

@sebrowne I'm not certain. When the `Kokkos_CoreUnitTest_Cuda1_MPI_1` runs in this setup, I'm assuming it may overlap with other tests executing simultaneously on the same GPU? That can cause issues (though not predictably), if that is the case can you try marking the test to `RUN_SERIAL` so it does not run in parallel with other tests?

sebrowne:
> When the `Kokkos_CoreUnitTest_Cuda1_MPI_1` runs in this setup, I'm assuming it may overlap with other tests executing simultaneously on the same GPU? That can cause issues (though not predictably), if that is the case can you try marking the test to `RUN_SERIAL` so it does not run in parallel with other tests?

Yes we do that intentionally.  Currently we allow 2 tests per GPU, but we'd like to bump that as high as reasonably possible for throughput reasons.  @achauphan the RUN_SERIAL thing is a good idea to try.

Forgive my lack of experience with GPU programming, but is running multiple tests per GPU a valid idea?  If not, that will seriously hamper our test throughput (asking because I don't like the RUN_SERIAL idiom in general).

ndellingwood:
> is running multiple tests per GPU a valid idea? If not, that will seriously hamper our test throughput (asking because I don't like the RUN_SERIAL idiom in general).

@sebrowne I would say it could be a bit brittle, though not invalid. We all definitely appreciate the efforts to reduce test times while maintaining coverage, thanks for working on this! 👍  

I say brittle because changes that come in with a PR can lead to cases where a test that previously ran in seconds may begin to timeout (> 10 mins), though it is not really predictable when this will occur (this can occur with changes in any package, not specific to Kokkos). I'm not sure how ctest handles the partitioning of tests to GPUs under the hood, but I'm guessing for example adding a new test in a package could alter how tests are balanced/spread among the GPUs, and depending on which/how many packages are enabled may result in resource-sensitive tests running simultaneously on the same GPU causing issues for one another.

For these types of failures, they don't tend to reproduce when run individually, so one option to counter this would be to rerun failed tests (would `ctest -j1` be an option for a rerun?); the downside is if this occurs due to a timeout then there would be accumulated 10 min overhead to test time (per timeout). Another option is to mark the test as `RUN_SERIAL`, which allows you to avoid taking on timeout overhead, but it is a reactive strategy after observing a timeout in PRs often enough to notice.
Not great options (I'm sure there are others but these are two I can think of), but either case would be worth the reduced testing time using multiple GPUs 

> Tests are able to utilize multiple GPUs now based on the number of them available on the machine they're running on with a current set max of 2 ranks per GPU. This last GPU test run took about 30mins!

👍

achauphan:
I've been unable to reproduce the individual failing tests manually of the last two AT runs `Phalanx_tViewOfViews_MPI_1 ` on 11/22 and `Adelus_vector_random_npr3_rhs1_MPI_3` on 11/27. Going to try another retest to see a third result.

ndellingwood:
> I've been unable to reproduce the individual failing tests manually of the last two AT runs `Phalanx_tViewOfViews_MPI_1 ` on 11/22 and `Adelus_vector_random_npr3_rhs1_MPI_3` on 11/27. Going to try another retest to see a third result.

@achauphan these tests may be candidates for the `RUN_SERIAL` treatment as well

achauphan:
Updated CUDA PR build to utilize `KOKKOS_MAP_DEVICE_ID_BY=mpi_rank` instead of `KOKKOS_NUM_DEVICES`.

retesting

achauphan:
Reverted the addition of `KOKKOS_MAP_DEVICE_ID_BY=mpi_rank` environment variable for CUDA PR builds as we suspect the mpi rank mapping is interfering with ctest resource mapping with multiple gpus.

retesting

achauphan:
@trilinos/phalanx do you happen to know if there are any issues with the unit test `PhalanxViewOfViews_KokkosToolsDefaultStreamCheck_UnitTest ` that ya'll may know of if the test is being run on a GPU other than GPU 0? 

I've noticed that this test is on GPU 0, it passes (which happened once in this PR), and when it is assigned to any other GPU, this specific test fails.

rppawlo:
Hang on. I used the wrong build. The cuda one does have an issue. Let me take a look at this...

achauphan:
> @achauphan - I just posted an issue to the kokkos team about the phalanx test mentioned above. I can disable the part of the test that is failing temporarily until this issue is resolved if you would like.
> 
> [kokkos/kokkos-tools#224](https://github.com/kokkos/kokkos-tools/issues/224)

@rppawlo Thanks for looking further into this and confirming my suspicions! Yes, it would be great if you could disable the failing part of the test for now.

achauphan:
Still unable to manually reproduce a failing `Adelus_vector_random_npr3_rhs1_MPI_3` with multiple GPUs enabled.

From past AT runs, this test has been unstable with it passing on 12/12 and 11/28, and failing elsewhere for CUDA builds.

I've investigated whether this may be a similar issue to the failing Phalanx test above, where the test behaves differently depending on which GPUs the test is running on, but was not able to reproduce a failing result.

Going to disable this test for CUDA PR builds.

fryeguy52 (approved):

Issue #11377, opened 2022-12-12
Stokhos: Compilation hang with CUDA RDC


sebrowne:
## Bug Report
@trilinos/stokhos 

### Description
The following four files hang when compiling with relocatable device code.  I've confirmed that if I disable Stokhos, the build completes.
packages/stokhos/src/MueLu_SemiCoarsenPFactory_kokkos_UQ_PCE_Cuda.cpp
packages/stokhos/src/MueLu_ScaledNullspaceFactory_UQ_PCE_Cuda.cpp
packages/stokhos/src/MueLu_SaPFactory_kokkos_UQ_PCE_Cuda.cpp
packages/stokhos/src/MueLu_CoalesceDropFactory_kokkos_UQ_PCE_Cuda.cpp

### Steps to Reproduce
1. Use the GenConfig for sems-cuda-11.4.2-sems-gnu-10.1.0-sems-openmpi-4.0.5_release_static_Volta70_no-asan_complex_no-fpic_mpi_pt_rdc_uvm_deprecated-on_all
2. Build with ninja (suspect make will work as well)
3. After a while, look at  top, and it will show an indefinite hang on those four source files (I've let it get as far as 8hrs per file, so I'm comfortable calling that a hang)

Note that since it's a build issue, you can replicate it on a machine without GPUs.

If you need a smaller reproducer, I suspect that building with CUDA11/GCC10.1.0/OpenMPI4.0.5 with RDC enabled and Stokhos enabled will also reproduce, but I have not explicitly tested that yet.  Also not sure how much "smaller" that will be....

sebrowne:
@e10harvey

sebrowne:
I think that's an option to nvcc_wrapper.  It looks like CMake thinks your compiler is g++ directly as opposed to mpicxx or nvcc_wrapper.

etphipp:
I figured it out.  I am reusing the build directory from a previous Cuda genconfig build but forgot to delete the cmake cache.

etphipp:
I've gotten to where I can finally reproduce this.  I let it run for a couple days and it never completed.  The amount of memory used seems to be changing so I don't think it is actually hung, maybe it is stuck in an infinite loop or something.  Regardless, I'm not sure what can really be done except disable the PCE scalar type in this build with
```
-D Stokhos_ENABLE_PCE_Scalar_Type=OFF \
```
[...]

PR #11394, opened 2022-12-15
Framework/Stokhos: Disable Stokhos MueLu with RDC (and warn if enabled)

sebrowne:
Enabling MueLu in Stokhos when building with CUDA RDC causes apparently-infinite compile times.  Turn it off in our RDC builds.  Also add a warning if it's enabled, just to maybe prevent others from encountering this.

User Support Ticket(s) or Story Referenced: [TRILFRAME-517](https://sems-atlassian-son.sandia.gov/jira/browse/TRILFRAME-517)

@trilinos/framework @trilinos/stokhos

## Motivation
Want nightly RDC builds, but enabling MueLu in Stokhos was causing infinite compile times.

## Related Issues
#11377


etphipp (approved):
Looks good, thanks!

PR #11397, opened 2022-12-15
Framework: Enable shadow warnings everywhere we use -Wall

sebrowne:
SIERRA builds with shadowing warnings/errors and in the past has been hit with shadowing warnings from Trilinos.  Turn on shadowing errors in Trilinos to try and keep everything cleaner.

User Support Ticket(s) or Story Referenced: [TRILFRAME-376](https://sems-atlassian-son.sandia.gov/jira/browse/TRILFRAME-376)

@trilinos/framework

### DIFF
--- a/packages/framework/ini-files/config-specs.ini
+++ b/packages/framework/ini-files/config-specs.ini
@@ -1457,6 +1457,53 @@ opt-set-cmake-var Zoltan_ENABLE_Scotch BOOL FORCE : OFF
 [CUDA11-RUN-SERIAL-TESTS]
 opt-set-cmake-var KokkosKernels_batched_dla_cuda_MPI_1_SET_RUN_SERIAL BOOL FORCE : ON
 
+[GCC_PACKAGE_SPECIFIC_WARNING_FLAGS]
+opt-set-cmake-var Amesos_CXX_FLAGS STRING : ${CMAKE_CXX_FLAGS|CMAKE} -Wno-error=shadow
+opt-set-cmake-var Amesos2_CXX_FLAGS STRING : ${CMAKE_CXX_FLAGS|CMAKE} -Wno-error=shadow
+#opt-set-cmake-var Anasazi_CXX_FLAGS STRING : ${CMAKE_CXX_FLAGS|CMAKE} -Werror
+opt-set-cmake-var Belos_CXX_FLAGS STRING : ${CMAKE_CXX_FLAGS|CMAKE} -Wno-error=shadow
+opt-set-cmake-var Compadre_CXX_FLAGS STRING : ${CMAKE_CXX_FLAGS|CMAKE} -Wno-error=shadow
+opt-set-cmake-var Domi_CXX_FLAGS STRING : ${CMAKE_CXX_FLAGS|CMAKE} -Wno-error=shadow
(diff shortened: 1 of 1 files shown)


sebrowne:
Yikes that's a lot of shadowing warnings.  Will discuss the best path forwards with the team.  I'm loath to make them just warnings, but I'm also not super keen on the idea of fixing them all immediately.

sebrowne:
@rppawlo I have written down in https://sems-atlassian-son.sandia.gov/jira/browse/TRILINOSHD-57 that you were the one who requested this.  I do have the context from the Sierra side, so I know why we wanted to do this for Trilinos.  Is just enabling the warnings satisfactory?  Do we think they'll get fixed over time, or do we need to make a thrust now to attempt to get rid of all of the shadowing and then turn it on as an error?

rppawlo:
@sebrowne - I suspect that if you just add the warning it will be ignored. Many packages are in maintenance mode so people don't really watch the builds. If we want to clean this out for the internal applications, we should probably make a serious coordinated effort at removal. Assign issues to packages that are not clean. We can toggle packages one by one to keep them from regressing.

sebrowne:
@rppawlo *sigh* I agree.  So, I'll leave the warnings on globally, and add -Werror=shadow to all of the individual package CXX/C flags that will currently build.  Then we can see what we're really dealing with with respect to how many packages need fixes (obviously we hit the 50 max on CDash for warnings/errors in the autotester runs).  If any packages have just one or two and I feel comfortable fixing them, I may just go ahead and do that this week.

I started fixing them this morning, and it didn't really take too long to fix Tpetra and a few others, but I'm not really a C++ programmer (I say with great shame), so I could be breaking things without knowing it.

cgcgcg:
@sebrowne 
After these PRs go in, we should be able to switch on shadow warnings as errors for some additional packages.

Belos, Stratimikos: #11423
Ifpack2: #11418
Tempus: #11413
Teko, Tpetra: #11422

sebrowne:
@cgcgcg Thank you so much!  I did see the Ifpack2 one and am reacting accordingly in this PR (better to have them be errors as soon as possible).  I'll remove the Wno-error flag from the other packages, though I'm not 100% sure I'll have it done over the break (doubtful).  Still, thanks a lot for taking care of these!

jhux2:
@sebrowne Are there other github issues related to eliminating shadow warnings (related to slide 7 from today's SART meeting).

sebrowne:
@jhux2 I didn't make any yet, no.  If that's how we want to handle it I can do so.  It's a lot of packages, would it be better to make one per package, or is it acceptable to tag all of the packages in a single issue?

sebrowne:
Also note that the warnings WILL start showing up for all of the packages, it's just "warnings-as-errors" that can't be enabled yet for the packages who currently have warnings.

jhux2:
> I didn't make any yet, no. If that's how we want to handle it I can do so. It's a lot of packages, would it be better to make one per package, or is it acceptable to tag all of the packages in a single issue?

@sebrowne You're right, that would be a lot.  I'm agnostic as to how, but I'd find it helpful to have this tracked somewhere.

srbdev (approved):

sebrowne:
@jhux2 I created #11497, does that work for tracking it?

jhux2:
Yep, thanks!

PR #11399, opened 2022-12-16
Teuchos StackedTimer: Fix for repeated reports with different options

cgcgcg:
@trilinos/teuchos 

## Motivation
This fixes segfaults when repeatedly exercising stacked timer reports with different options. This can happen when timing summaries are printed to both terminal and Watchr XML.


rppawlo (approved):

GrahamBenHarper:
Thanks @cgcgcg!

PR #11409, opened 2022-12-19
Panzer: remove hard-coded 3d dim in ElemFieldPattern method

bartgol:
@trilinos/panzer 

## Motivation

The class `ElemFieldPattern` was hard-coding `3` in one of its method, as the value for element dimension. This PR changes the check to use `getDimension()`, so that it works for Elem fields in meshes of arbitrary dimension. The behavior is unchanged if the mesh dimension is 3.

## Stakeholder Feedback

In Albany, we are checking `getSubcellIndices(dim,0).size()` (for `dim=0,...,getDimensions()`), to determine if the field in the mesh is a NODE or ELEM field. With the current master, on (n<=2)-dimensional elements, ElemFieldPattern shows no dofs for all `dim`. Things would work correctly if we used a `FaceFieldPattern`, but it'smore concise to always use `ElemFieldPatter` for fields that have one value per cell, regardless of mesh dimension.

## Testing

Added a unit test to verify indices counts for basic patterns (Nodal,Edge,Face,Elem).

### DIFF
--- a/packages/panzer/dof-mgr/src/Panzer_ElemFieldPattern.cpp
+++ b/packages/panzer/dof-mgr/src/Panzer_ElemFieldPattern.cpp
@@ -71,7 +71,7 @@ int ElemFieldPattern::getSubcellCount(int dim) const
 
 const std::vector<int> & ElemFieldPattern::getSubcellIndices(int dim,int cellIndex) const
 {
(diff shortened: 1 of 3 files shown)


rppawlo (approved):
Thanks @bartgol !

PR #11416, opened 2022-12-21
Zoltan2:  Allow coloring interface to process only chunks of colors at a time

etphipp:
Previously the Zoltan2 coloring interface required allocating multivectors with numbers of columns equal to the number of colors so that reconstructMatrix() could reconstruct the whole matrix.  However, when the number of colors is large, the seed and compression multivectors can consume large amounts of memory (easily greater than the matrix itself).  This change allows computing the seed matrix and processing of colors in chunks of a user-chosen size, allowing the multivectors to only need a number of columns equal to the chunk size. This was a fairly simple change and is backwards compatible with processing all colors at once.  It does require the user to write a loop around the process of computing the seed matrix, computing the compression matrix (user-defined), and reconstructing the matrix incorporating whatever algorithm they choose for chunking up the colors.

@trilinos/zoltan2

### DIFF
--- a/packages/zoltan2/core/src/TpetraCrsColorer/Zoltan2_TpetraCrsColorer.hpp
+++ b/packages/zoltan2/core/src/TpetraCrsColorer/Zoltan2_TpetraCrsColorer.hpp
@@ -49,29 +49,29 @@ public:
   // Compute seed matrix
   template <typename MultiVectorType>
   void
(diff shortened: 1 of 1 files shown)


egboman (approved):
This is a very useful change (enhancement). Chunking the matrix coloring saves memory and does not affect other things. Thanks, @etphipp !
(Sorry for the delay, I forgot about this review.)

PR #11432, opened 2023-01-03
Adelus: Add original factorsolve interfaces back for calls within EIGER

vqd8a:
@trilinos/adelus

## Motivation

Since Adelus interfaces now require a handle, this PR adds back the original interfaces ```Adelus::GetDistribution``` and ```Adelus::FactorSolve_hostPtr()``` which act as wrappers calling the new interfaces. This will help EIGER to be able to call Adelus with its recent updates.

## Testing

Tests were performed by @jdkotul.


jdkotul (approved):
The code changes are consistent with the request concerning the interface to the electromagnetic code EIGER.   These changes allow the code to use the original interface to Adelus that was previously written.   These interfaces have been integrated and tested successfully with EIGER on a NVIDIA GPU system.

PR #11437, opened 2023-01-05
Enable ccache for SRN PR testing

jmlapre:
@trilinos/framework 

## Motivation

As anyone associated with Trilinos, I want the autotester to be faster.

Use ccache for PR testing.  Ideally, start with the CUDA build, since it takes the longest.

Store the cache at `/fgs/trilinos/ccache/cache`.

This is a follow-up to #11386.

https://sems-atlassian-son.sandia.gov/jira/browse/TRILFRAME-521

### DIFF
--- a/packages/framework/pr_tools/PullRequestLinuxDriver.sh
+++ b/packages/framework/pr_tools/PullRequestLinuxDriver.sh
@@ -8,6 +8,21 @@ source ${SCRIPTPATH:?}/common.bash
 on_weaver=$(echo "$@" | grep '\-\-on_weaver' &> /dev/null && echo "1")
 on_ats2=$(echo "$@" | grep '\-\-on_ats2' &> /dev/null && echo "1")
 
+# Configure ccache via environment variables
+function configure_ccache() {
+    print_banner "Configuring ccache"
+
+    # Identify the path to the trilinos repository root
+    REPO_ROOT=`readlink -f ${SCRIPTPATH:?}/../..`
+    test -d ${REPO_ROOT:?}/.git || REPO_ROOT=`readlink -f ${WORKSPACE:?}/Trilinos`
+
+    envvar_set_or_create CCACHE_NODISABLE true
+    envvar_set_or_create CCACHE_DIR '/fgs/trilinos/ccache/cache'
+    envvar_set_or_create CCACHE_BASEDIR "$REPO_ROOT"
+    envvar_set_or_create CCACHE_NOHARDLINK true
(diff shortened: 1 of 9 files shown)


sebrowne:
Should this also set CCACHE_NODISABLE to engage it in the CMake code from #11386?

jmlapre:
> Should this also set CCACHE_NODISABLE to engage it in the CMake code from #11386?

Yes, good catch.

packages/framework/pr_tools/PullRequestLinuxDriver.sh
```diff
@@ -8,6 +8,16 @@ source ${SCRIPTPATH:?}/common.bash
 on_weaver=$(echo "$@" | grep '\-\-on_weaver' &> /dev/null && echo "1")
 on_ats2=$(echo "$@" | grep '\-\-on_ats2' &> /dev/null && echo "1")
 
+# Configure ccache via environment variables
+function configure_ccache() {
+    print_banner "Configuring ccache"
+
+    envvar_set_or_create CCACHE_NODISABLE true
+    envvar_set_or_create CCACHE_BASEDIR '/fgs/trilinos/ccache/cache'
```
sebrowne, line 16: Sorry I missed this before, but I think what we want is this:
CCACHE_DIR '/fgs/trilinos/ccache/cache'
CCACHE_BASEDIR '/path/to/trilinos/source/code'

I imagine that second one is available as a variable in here somewhere.

sebrowne (changes requested):

sebrowne (approved):

jmlapre:
Unfortunately it seems to be a greater possibility than anticipated:
  
https://gitlab.kitware.com/cmake/cmake/-/issues/24345

I think ccache is likely to be a bigger win than ninja but we'll wait and see the results...

sebrowne:
> Unfortunately it seems to be a greater possibility than anticipated:
> 
> https://gitlab.kitware.com/cmake/cmake/-/issues/24345
> 
> I think ccache is likely to be a bigger win than ninja but we'll wait and see the results...

@jmlapre I think this only applies to coverage builds (.gcno files are only generated for coverage-profiling enabled builds as far as I know)

sebrowne:
Okay, I did some diagnosis on this today.  I think that what is happening is that Trilinos (nearly every source file) includes generated headers, which means it has -I directives that point into the BUILD directory.  For pretty obvious reasons, this makes it impossible to cache files on a per-object basis.  Will discuss at standup tomorrow.

sebrowne:
Closing since #11616 is now merged

PR #11442, opened 2023-01-09
SEACAS: Update embedded lib{fmt} version to 9.1.0

gsjaardema:
@trilinos/seacas

## Motivation

Update embedded lib{fmt} version to get better support for C++14, C++17, and beyond.  
Long-term solution is to support lib{fmt} as an external library instead of embedding it in code.


ccober6 (approved):
Assuming most of the changes are from update to version 9.1.0, so lgtm.

sebrowne (approved):

PhilMiller:
This may be a silly question, but why does SEACAS have *two* copies of libfmt?

gsjaardema:
> This may be a silly question, but why does SEACAS have _two_ copies of libfmt?

In the standalone version, there is only a single version and it is an external TPL.  In Trilinos (and originally in Seacas), we put two versions to avoid a dependency between suplib and ioss.  We could have added a single embedded fmt that both depended on, but was easier at the time to have two copies...

Once can get fmt as an external TPL for Trilinos, then can remove both copies which is how standalone seacas is working curretnly.

ccober6:
@gsjaardema, for gtest, there is a need to be able to link against an external copy but have an internal copy to test on platforms that might not have an external copy.  Do you see a similar situation for fmt?

If you have not yet, can you submit a story for the external fmt to the Trilinos Help Desk so we can get it the Framework backlog?

gsjaardema:
> @gsjaardema, for gtest, there is a need to be able to link against an external copy but have an internal copy to test on platforms that might not have an external copy. Do you see a similar situation for fmt?

If we remove the embedded versions, then we would not be able to build on platforms that did not have the external copy.  It would be a required dependency.

rppawlo:
I think this will a be a moot point for fmt in another year or so. Kokkos plans to require the new c++ standard 3-4 years after release. So I expect that next year we will be moving both kokkos and trilinos to require c++20. The fmt library was adopted into the c++20 standard. So we could keep these fmt copies around in the short term and drop them when kokkos and trilinos require the C++20 standard.

gsjaardema:
> I think this will a be a moot point for fmt in another year or so. Kokkos plans to require the new c++ standard 3-4 years after release. So I expect that next year we will be moving both kokkos and trilinos to require c++20. The fmt library was adopted into the c++20 standard. So we could keep these fmt copies around in the short term and drop them when kokkos and trilinos require the C++20 standard.

Yes, that is an option, but would also be nice to get rid of them so Trilinos snapshot is easier...  But it isn't that much of an issue, so may not be worth the effort of changing.

rppawlo:
@gsjaardema - this merge caused all gnu builds (gcc 7.2.0) in empire testing to fail last night with many errors. A couple are listed below. Do you know if this new version of fmt requires c++17? The empire test suite is still using c++14, so this might be the issue. I will look into it further.
```
00:08:40  gnu-debug-openmp-static/src/Trilinos/packages/seacas/libraries/ioss/src/private_copy_fmt/fmt/format.h:1241:8: error: uninitialized variable ‘buffer’ in ‘constexpr’ function
00:08:40     Char buffer[digits10<UInt>() + 1];
00:08:40          ^~~~~~
```
and
```
gnu-debug-openmp-static/src/Trilinos/packages/seacas/libraries/ioss/src/private_copy_fmt/fmt/format.h:1116:4: error: use of ‘fmt::v9::detail::count_digits(UInt)::<lambda(UInt)> [with int BITS = 4; UInt = unsigned int]’ before deduction of ‘auto’
00:08:40     return [](UInt m) {
00:08:40            ~~~~~~~~~~~~
00:08:40       int num_digits = 0;
00:08:40       ~~~~~~~~~~~~~~~~~~~
00:08:40       do {
00:08:40       ~~~~
00:08:40         ++num_digits;
00:08:40         ~~~~~~~~~~~~~
00:08:40       } while ((m >>= BITS) != 0);
00:08:40       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
00:08:40       return num_digits;
00:08:40       ~~~~~~~~~~~~~~~~~~
00:08:40     }(n);
00:08:40     ~^~~
```

gsjaardema:
No, C++17 is not required.   It has been used in Sierra and SEACAS standalone with C++11 and C++14 although those are using the non-header-only version.

gsjaardema:
The previous version of the code in your first error wasn't marked constexpr, so maybe removing the `FMT_CONSTEXPR`around line 1236 would fix that issue.  

The other error looks like the same code in the previous version vs this version, so I'm not sure what is happening there yet.

I can compile using gcc-7.5.0,  but there were some constexpr-related changes between 7.2.0 and 7.5.0...  Will see if I can reproduce.

PhilMiller:
Even if GCC 7.2 nominally supports C++17, the minimum requirement from Kokkos 4 is going to be 8.2.

I think that leans even harder against putting substantial effort into any sort of workarounds.

gsjaardema:
I was able to reproroduce this (or at least part of it) on the CEE Lan.  It does look like removing the `FMT_CONSTEXPR` at line 1104 fixes at least some of the problems.  Not sure how easy it would be to check that on your side.

It also looks like using current development version fixes the same issue.

gsjaardema:
> Even if GCC 7.2 nominally supports C++17, the minimum requirement from Kokkos 4 is going to be 8.2.
> 
> I think that leans even harder against putting substantial effort into any sort of workarounds.

Just to clarify, there is no C++17 requirement from lib{fmt}.

rppawlo:
It turns out that removing the same line in both copies of fmt fixes all the failures. If you don't mind I will put up a PR so that the trilinos sync of empire will continue to work while we upgrade our compilers. By the end of the month we should be updated and can revert the hack.

gsjaardema:
That sounds good to me.  Glad it worked.

rppawlo:
@gsjaardema - thanks for your help on this!

PR #11443, opened 2023-01-10
Adelus: Do not use `Kokkos::Impl::throw_runtime_exception`

dalg24:
It is not legal to use `Kokkos::Impl::throw_runtime_exception()`.
Throw `std::runtime_error` instead.

adelus @srajama1


jhux2:
@trilinos/adelus

vqd8a (approved):
Thanks @dalg24

dalg24:
Any chance to get this tested and merged in?

PR #11450, opened 2023-01-10
Shards: make Particle topology 0-dimensional

bartgol:
@trilinos/shards 

## Motivation

The fact that a Particle was 1-dimensional created issues when using it as a "cell" topology in Panzer's dof manager stack. Although no Finite Elements can be defined on a Particle, being able to use it allows to avoid treating special cases everywhere, and keep a uniform code.

## Stakeholder Feedback

Will considerably help Albany to switch to Panzer::DOFManager class(es).

## Testing

No additional tests required. Shards existing tests already cover Particle.

### DIFF
--- a/packages/shards/src/Shards_BasicTopologies.cpp
+++ b/packages/shards/src/Shards_BasicTopologies.cpp
@@ -400,14 +400,20 @@ struct Descriptor<
     };
 };
 
-template<>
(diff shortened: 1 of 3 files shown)


mperego (approved):
Looks good to me. Thanks!

PR #11460, opened 2023-01-11
CMake: Update Rocketman Avatar nightly, update Avatar repository location

GrahamBenHarper:
@cgcgcg 

This should hopefully clean up some of the Rocketman nightly dashboard.
Also, Avatar_repo used to point to Mark's gitlab, so I updated it to point to the sandialabs github repo.


cgcgcg (commented):
Don't you need to tell the ctest driver to actually pull and install avatar?

cgcgcg (approved):
Ah, gotcha. That's cool then.

Issue #11459, opened 2023-01-11
Tpetra: Deprecate all use of things Kokkos that aren't actually Kokkos


csiefer2:
See also #10715

But mostly the headers in tpetra/core/compat

jhux2:
🔥

spdomin:
Hi @csiefer2, is this new Nalu fail (on a variety of platforms) related to this issue,

/nightlyBuildAndTest/Nalu/include/TpetraLinearSystem.h:15:10: fatal error: 
      'Kokkos_DefaultNode.hpp' file not found
#include <Kokkos_DefaultNode.hpp>

- or, perhaps the Nalu codes did not see a notice to remove this header?

Best,

GrahamBenHarper:
@spdomin, I believe #11462 may be related to your issue if you are building without deprecated code. Are you able to get past the error with deprecated code turned on? i.e.
`-D Tpetra_ENABLE_DEPRECATED_CODE:BOOL=ON`

spdomin:
Removing the included header on my Nalu files fixed the issue. I do not believe that we add this deprecated line in our trilinos configuration.  I will look at the above ticket. Thanks.

csiefer2:
@spdomin At the Kokkos team's request, we deprecated that header, and others like it called Kokkos_*.hpp.

PR #11465, opened 2023-01-13
Tpetra-and-friends:  Deprecating tpetra/core/compat/Kokkos_* headers and removing their use

csiefer2:
Issues: #10715 and #11459
Tests: Run
Stakeholder Feedback: n/a


packages/xpetra/test/compile_tests/generate_tests.sh
```diff
@@ -23,9 +23,9 @@ not_a_templated_class=$(find ../.. -name "*Epetra*.hpp" -exec basename {} \;)
 not_a_templated_class="
 $not_a_templated_class
 
-Kokkos_ConfigDefs.hpp
-Kokkos_DefaultNode.hpp
-Kokkos_DefaultNode.hpp
+TpetraCore_config.h
+KokkosCompat_DefaultNode.hpp
+KokkosCompat_DefaultNode.hpp
```
cgcgcg, line 28: Duplication?

csiefer2: Evidently my sed/grep monkeys failed me ;P  Fixed.

cgcgcg (approved):

packages/tpetra/core/compat/KokkosCompat_DefaultNode.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+#ifndef DOXYGEN_SHOULD_SKIP_THIS
+namespace Teuchos {
+  // Dear users: This is just a forward declaration.
+  // Please skip over it.
+  class ParameterList;
+} // namespace Teuchos
+#endif // DOXYGEN_SHOULD_SKIP_THIS
+
+namespace KokkosClassic {
```
dalg24, line 58: Overall this PR is in the right direction.

Just making sure you understand that it would be better if you prefixed these symbols with `Tpetra`, `Tpetra_`, or nested it in the `Tpetra::` namespace because you technically do not own the `KokkosClassic` nor the `KokkosCompat` "brand".
The same apply to filenames and macros.
(It does not have to be `Tpetra` but it must be something you actually have ownership over.)

csiefer2: @dalg24   A bunch of that KokkosClassic / KokkosCompat nomenclature predates Kokkos as a standalone thing.  Yay for technical debt :)

Moving to something like something like Tpetra_KokkosCompat_* for filenames and Tpetra::KokkosClassic for namespaces, would be better aligned with Kokkos' goals, yes?

dalg24: Yes

PR #11466, opened 2023-01-13
zoltan2: Workaround c++20 deprecation of implicit this capture via [=].

vbrunini:
@trilinos/<teamName>


vbrunini:
@trilinos/zoltan2

ndellingwood (commented):
@vbrunini just a question (I'm not a zoltan2 dev), but can the capture just be changed to capture by reference `[&]` to avoid the if-else guards?

vbrunini:
> @vbrunini just a question (I'm not a zoltan2 dev), but can the capture just be changed to capture by reference `[&]` to avoid the if-else guards?

That would also prevent the deprecation warning yes. I have not read through the surrounding code to determine if the semantic change to capturing the variables other than "this" by reference would still be correct though.

jhux2:
@trilinos/zoltan2 This looks fine to me.  Adding a label to get it tested.

egboman:
Looks good to me. Thanks.

jhux2 (approved):
Approving (@egboman already signed off on this)

PR #11473, opened 2023-01-17
Tpetra: Fix wrong execution space in CrsGraph

cgcgcg:
@trilinos/tpetra 

## Motivation
Broken in #11422. Changing back to what it was while avoiding shadow warnings.


jhux2 (approved):
lgtm

PR #11474, opened 2023-01-17
Testing: geminga: suppress valgrind warnings about leaks in system perl library

jhux2:
## Motivation
Memory leak warnings in system libraries could make it more difficult to identify leaks in Trilinos itself.

## Stakeholder Feedback
N/A

## Testing
Will evaluate when the next nightly Valgrind testing runs.


GrahamBenHarper (approved):
The hero we need!

PR #11017, opened 2022-09-14
KokkosKernels: Fix #10950

brian-kelley:
(Patch for KK PR #1528). Fixes bug #10950 where uninitialized memory was used in MIS-2 aggregation phase3. This didn't break the validity of the aggregation, but it did break determinism and potentially decreased the quality (size balance) of the aggregates.

@trilinos/kokkos-kernels 
@trilinos/muelu 

## Motivation

Fixes a real bug, see description above

## Related Issues

* Closes #10950 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of https://github.com/kokkos/kokkos-kernels/pull/1528

## Testing

Bug showed up as Valgrind errors (conditional branch on uninitialized values) in the KokkosKernels graph unit tests and the MueLu Kokkos refactor unit tests.

PR #11476, opened 2023-01-17
KokkosKernels: fix MIS2 agg determinism, valgrind

brian-kelley:
Patch in KokkosKernels #1528 and #1530. 1528 fixed MIS2 aggregation being possibly non-deterministic, and fixed valgrind conditional on uninitialized value warnings. 1530 fixed valgrind warnings in EB D1 coloring.

@trilinos/kokkos-kernels 

## Motivation

Fixes https://github.com/kokkos/kokkos-kernels/issues/1646 .

The MIS2 changes are a re-application of #11017 . Fixes non-determinism (a real bug) and also valgrind uninitialized value warnings.
The D1 EB changes have been in KokkosKernels develop for a while, but never in Trilinos. There didn't seem to be an actual bug here, but patching this will help make the nightly geminga build with valgrind run cleanly.


jhux2 (approved):
lgtm.  Thanks @brian-kelley!

PR #11480, opened 2023-01-19
Amesos2: Add row permutation option to SuperLU_dist

### LINKED ISSUE #11469: Amesos2:  Row permutation in SuperLU_dist
Hello,

In the scope of switching our in-house multi-physics code from `Epetra` to `Tpetra` and thus from `Amesos` to `Amesos2`, we came across a problem in the `Amesos2` adapter of `SuperLU_dist`. In the source code it states, that row permutations are currently not available/allowed, due to a sequential bottleneck in the implementation.

Some parts of our code are depending on this feature, which is available in `Amesos`. For us it would be really helpful and in some parts also necessary to add row permutation as an user input (leaving the current hard coded option as default) to the `SuperLU_dist` interface in `Amesos2`.

A colleague and myself would contribute this feature.
Whats you're opinion on this? Was this never used by someone?

Might be interesting for: @mayrmt @vryy

Best regards,
Max

maxfirmbach:
@trilinos/amesos2

## Motivation

Currently the `SuperLU_dist` interface of `Amesos2` allows no row permutation as pre-ordering method. For certain type of matrices a row pre-ordering is a necessary step to be able to factorize the matrix without running into errors. This adds the `LargeDiag_MC64` row permutation option, even though it might be a sequential bottleneck in the algorithm.

## Stakeholder Feedback

Enhancing the `SuperLU_dist` interface of `Amesos2`

## Additional Information

Joint work with @vryy and @mayrmt 

See also #11469.

### DIFF
--- a/packages/amesos2/src/Amesos2_Superludist_decl.hpp
+++ b/packages/amesos2/src/Amesos2_Superludist_decl.hpp
@@ -144,6 +144,14 @@ public:
 private:
 
   /**
+   * \brief Compute the row permutation for option LargeDiag-MC64.
+   *
+   * SuperLU_DIST supports several forms of row permutations.  Refer
+   * to \ref slu_mt_options for the available \c RowPerm options.
+   */
+  void computeRowPermutationLargeDiagMC64(SLUD::SuperMatrix& GA);
+
+  /**
    * \brief Performs pre-ordering on the matrix to increase efficiency.
    *
    * SuperLU_DIST supports several forms of column permutations.  Refer
@@ -326,6 +334,7 @@ private:
   /// \c true if this processor is in SuperLU_DISTS's 2D process grid
   bool in_grid_;
   bool same_symbolic_;
+  bool force_symbfact_;
   mutable bool same_solve_struct_; // may be modified in solve_impl, but still `logically const'
 
   /// Maps rows of the matrix to processors in the SuperLU_DIST processor grid
--- a/packages/amesos2/src/Amesos2_Superludist_decl.hpp
+++ b/packages/amesos2/src/Amesos2_Superludist_decl.hpp
@@ -308,6 +308,7 @@ private:
     SLUD::DiagScale_t equed;    ///< Whether/what kind of equilibration to use/has been used
     bool rowequ, colequ;        ///< whether row/col equilibration has been applied to AC
     magnitude_type rowcnd, colcnd, amax;
+    int largediag_mc64_job;     // job id for LargeDiag_MC64 row permutation
   } data_;
 
   // The following Arrays are persisting storage arrays for A, X, and B
--- a/packages/amesos2/src/Amesos2_Superludist_def.hpp
+++ b/packages/amesos2/src/Amesos2_Superludist_def.hpp
@@ -327,13 +327,60 @@ namespace Amesos2 {
   int
   Superludist<Matrix,Vector>::preOrdering_impl()
   {
-    // We will always use the NATURAL row ordering to avoid the
-    // sequential bottleneck present when doing any other row
-    // ordering scheme from SuperLU_DIST
-    //
-    // Set perm_r to be the natural ordering
-    SLUD::int_t slu_rows_ub = Teuchos::as<SLUD::int_t>(this->globalNumRows_);
-    for( SLUD::int_t i = 0; i < slu_rows_ub; ++i ) data_.perm_r[i] = i;
+    if (data_.options.RowPerm == SLUD::NOROWPERM) {
+      SLUD::int_t slu_rows_ub = Teuchos::as<SLUD::int_t>(this->globalNumRows_);
+      for( SLUD::int_t i = 0; i < slu_rows_ub; ++i ) data_.perm_r[i] = i;
+    } else if (data_.options.RowPerm == SLUD::LargeDiag_MC64) {
+
+      double *R1, *C1;
+
+      /* Allocate storage for scaling factors. */
+
+      TEUCHOS_TEST_FOR_EXCEPTION( !(R1 = SLUD::D::doubleMalloc_dist(data_.A.nrow)),
+          std::runtime_error,
+          "SUPERLU_MALLOC fails for R1[]");
+
+      TEUCHOS_TEST_FOR_EXCEPTION( !(C1 = SLUD::D::doubleMalloc_dist(data_.A.ncol)),
+          std::runtime_error,
+          "SUPERLU_MALLOC fails for C1[]");
+
+      SLUD::SuperMatrix GA;      /* Global A in NC format */
+      SLUD::D::pdCompRow_loc_to_CompCol_global(true, &data_.A, &data_.grid, &GA);
+
+      SLUD::NCformat *GAstore = (SLUD::NCformat*) GA.Store;
+      SLUD::int_t* colptr = GAstore->colptr;
+      SLUD::int_t* rowind = GAstore->rowind;
+      SLUD::int_t nnz = GAstore->nnz;
+      double *a_GA = (double *) GAstore->nzval;
+
+      int iinfo, job = 5;
+      if ( !data_.grid.iam ) { /* Process 0 finds a row permutation */
+
+        iinfo = SLUD::D::dldperm_dist(job, data_.A.nrow, nnz, colptr, rowind, a_GA,
+                data_.perm_r.getRawPtr(), R1, C1);
+
+        MPI_Bcast( &iinfo, 1, mpi_int_t, 0, data_.grid.comm );
+        if ( iinfo == 0 ) {
+            MPI_Bcast( data_.perm_r.getRawPtr(), data_.A.nrow, mpi_int_t, 0, data_.grid.comm );
+            if ( job == 5 && data_.options.Equil ) {
+                MPI_Bcast( R1, data_.A.nrow, MPI_DOUBLE, 0, data_.grid.comm );
+                MPI_Bcast( C1, data_.A.ncol, MPI_DOUBLE, 0, data_.grid.comm );
+            }
+        }
+      } else {
+          MPI_Bcast( &iinfo, 1, mpi_int_t, 0, data_.grid.comm );
+          if ( iinfo == 0 ) {
+            MPI_Bcast( data_.perm_r.getRawPtr(), data_.A.nrow, mpi_int_t, 0, data_.grid.comm );
+            if ( job == 5 && data_.options.Equil ) {
+                MPI_Bcast( R1, data_.A.nrow, MPI_DOUBLE, 0, data_.grid.comm );
+                MPI_Bcast( C1, data_.A.ncol, MPI_DOUBLE, 0, data_.grid.comm );
+            }
+          }
+      }
+
+      SUPERLU_FREE(R1);
+      SUPERLU_FREE(C1);
+    }
 
     // loadA_impl();                    // Refresh matrix values
 
@@ -736,6 +783,13 @@ namespace Amesos2 {
     bool equil = parameterList->get<bool>("Equil", false);
     data_.options.Equil = equil ? SLUD::YES : SLUD::NO;
 
+    if( parameterList->isParameter("RowPerm") ){
+      RCP<const ParameterEntryValidator> rowperm_validator = valid_params->getEntry("RowPerm").validator();
+      parameterList->getEntry("RowPerm").setValidator(rowperm_validator);
+
+      data_.options.RowPerm = getIntegralValue<SLUD::rowperm_t>(*parameterList, "RowPerm");
+    }
+
     if( parameterList->isParameter("ColPerm") ){
       RCP<const ParameterEntryValidator> colperm_validator = valid_params->getEntry("ColPerm").validator();
       parameterList->getEntry("ColPerm").setValidator(colperm_validator);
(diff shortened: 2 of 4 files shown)


iyamazaki (commented):
Thank you for the contributions, @maxfirmbach and @vryy!! 

First of all, my understanding is that `dldperm_dist` requires the numerical values of the matrix. So I believe it should be called at the beginning of `numericFactorization_impl` instead of in `preOrdering_impl`, which is called at the beginning of `symbolicFactorization`?

Then, since it permutes the matrix (which I don't see in this PR?), we need to call `symbolicFactorization_impl` again? Have you tested this code on the matrix that needs `MC64`?

Also, `ldperm_dist` should be added to `Amesos2_Superludist_FunctionMap.hpp` so that it can be called with complex double.

maxfirmbach:
@iyamazaki With the proposed changes our in-house code works nicely and the so far failing tests run through. I guess it would be good to add some kind of test for the changes. 

As I understand it we just need to return `perm_r`, which contains the new row ordering, which is then used in the factorization, where the actual permutation takes place.

@vryy Please correct me if I'm wrong.

iyamazaki:
Thank you for testing, @maxfirmbach. 

I was asking:
- This is hard-coded for `double` so won't work for other precision, e.g., `complex double`?
- Do you need to manually apply some of the matrix preprocessing, e.g., where are the matrix scaling `R1` and `C1` applied?
- `MC64` needs numerical values so this needs to be called in `numericFactorization`
- Need to destroy `GA` to avoid memory-leak?

vryy:
> Thank you for testing, @maxfirmbach.
> 
> I was asking:
> 
> * This is hard-coded for `double` so won't work for other precision, e.g., `complex double`?
> * Do you need to manually apply some of the matrix preprocessing, e.g., where are the matrix scaling `R1` and `C1` applied?
> * `MC64` needs numerical values so this needs to be called in `numericFactorization`
> * Need to destroy `GA` to avoid memory-leak?

@iyamazaki your point 1, 2, 4 are totally valid and we will address it in the next commit. Furthermore, for point 2 if we use job=1, R1 and C1 are not necessary and can be removed. For point 3 (MC64) i did not quite get it since it is used to compute a row permutation vector and shall be called during preOrdering phase. It requires values of A therefore GA (clone in column format) is created.

iyamazaki:
Thank you again for the contributions, @vryy!! It looks like you are using `job=5`, which needs the numerical values, and also the matrix scaling `R` and `C` should be applied? 

Amesos2 has two setup phases, 1) symbolic, which only uses the graph structure, and then 2) numeric, which uses the numerical values. So if the graph stays the same, a user can call numerical factorization multiple time without doing symbolic factorization each time. If I look at `Amesos2_SolverCore_def.hpp`, I see `preOrdering` is called in symbolic. Since `MC64` with `job=5` needs numerical values, it needs to be moved to numericalFactorization, I think.

vryy:
@iyamazaki you are right, with job=5 then additional scaling for row permutation should be performed and it should be done in the numerical factorization. For job=1->4 the values are not needed hence row permutation vector can be computed in preOrdering(). For a clean implementation, we suggest to compute the row-perm scaling vector (for job=5) in preOrdering() and reuse it in numericFactorization()

packages/amesos2/src/Amesos2_Superludist_decl.hpp
```diff
@@ -308,6 +308,7 @@ class Superludist : public SolverCore<Amesos2::Superludist, Matrix, Vector>
     SLUD::DiagScale_t equed;    ///< Whether/what kind of equilibration to use/has been used
     bool rowequ, colequ;        ///< whether row/col equilibration has been applied to AC
     magnitude_type rowcnd, colcnd, amax;
+    int largediag_mc64_job;     // job id for LargeDiag_MC64 row permutation
```
mayrmt, line 311:
```suggestion
    int largediag_mc64_job;     ///< job id for LargeDiag_MC64 row permutation
```

iyamazaki:
Thank you, @vryy!!  I also think these are very useful options.  For job!=1, the user should be aware the row permutation is computed using the numerical values of the matrix (which may not be initializes, e.g., may be all zero).

I still have a couple of questions/comments about the matrix scaling, e.g., for job=5,
- We should take `exp` of `R1` and `C1` in `preOrdering` instead of in `numericalFactorization` (numericalFact may be called multiple times per preOrder)
- If we want to combine `job=5` and `Equil`, shouldn't we apply `R1` and `C1` before computing `R` and `C`? And, `R1` and `C1` are not applied unless `Equil` is on?

Which option do you actually need, for your runs? Another option may be to allow only some `job` for now (e.g., just job=1)?

By the way, we should use `Teuchos` for MPI, instead of using `MPI_DOUBLE`..

vryy:
> * We should take `exp` of `R1` and `C1` in `preOrdering` instead of in `numericalFactorization` (numericalFact may be called multiple times per preOrder)
that makes sense

> * If we want to combine `job=5` and `Equil`, shouldn't we apply `R1` and `C1` before computing `R` and `C`? And, `R1` and `C1` are not applied unless `Equil` is on?
I implemented it so based on observation of the solver driver provided along with Superlu_dist. I think either solution may have a performance implication but obviously not the accuracy. For the latter question, I understand that Equil is the control flag for scaling at all. Please correct me if I'm wrong.

> Which option do you actually need, for your runs? Another option may be to allow only some `job` for now (e.g., just job=1)?
Right now I'm using job=5. With job=1, Superlu_dist gave error since it needs to be linked with mc21ad and I yet to figure it out a solution. For job=1,2,3,4 Superlu_dist uses internal algorithm and they need no special treatment as job=5.

> By the way, we should use `Teuchos` for MPI, instead of using `MPI_DOUBLE`..
sure

vryy:
@iyamazaki, I saw that the build pipeline failed. Could you please share with us the build log and test lo?

iyamazaki:
A different test failed in the last two PR tests. The last one may be related to Issue #11530, but not sure about the one before (it might have ran out of time). Let me include @ndellingwood..

mayrmt:
@iyamazaki It is not possible to view test results from outside the SNL network. Could you please repost the log of the failing test, so that @vryy has a chance to fix it?

ndellingwood:
From cdash looks like the `NOX_1DfemAztecSolverPrecRecompute_MPI_4` test failed in the `Trilinos_PR_gcc-8.3.0-debug` build. I don't see it listed explicitly on #11530 , but it could be another instance of a spurious test failure (and the output message is inconsistent with failure)

Test result:
```
Test passed!
-------------------------------------------------------
a non-zero exit code.. Per user-direction, the job has been aborted.
-------------------------------------------------------
```

@rppawlo does the `NOX_1DfemAztecSolverPrecRecompute_MPI_4` test use any solver capabilities from the Amesos2::SuperLU_DIST solver? I'm assuming not since it references Aztec

Adding the RETEST label to get the PR retested and see if the failure occurs again

ndellingwood:
Most recent test had two failures in the `Trilinos_PR_gcc-8.3.0-debug ` build:

* [NOX_1DfemStratimikos_MPI_4](https://trilinos-cdash.sandia.gov/test/19865221) - spurious failing test reported in #11530 
*  `Amesos2_CrsMatrix_Adapter_Consistency_Tests_MPI_4` - even though it's an Amesos2 failure (and this PR is an update for Amesos2), the reported failure contradicts the test output:
```
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascic166 and rank 0!
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascic166 and rank 1!
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascic166 and rank 2!
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascic166 and rank 3!

***
*** Unit test suite ...
***


Sorting tests by group name then by the order they were added ... (time = 0.000321)

Running unit tests ...

0. CrsMatrixAdapter_Dimensions_UnitTest ... [Passed] (0.0239 sec)
1. CrsMatrixAdapter_CRS_Serial_UnitTest ... [Passed] (0.024 sec)
2. CrsMatrixAdapter_CRS_Replicated_UnitTest ... [Passed] (0.0217 sec)
3. CrsMatrixAdapter_CRS_Map_UnitTest ... [Passed] (0.0171 sec)

Total Time: 0.0877 sec

Summary: total = 4, run = 4, passed = 4, failed = 0

End Result: TEST PASSED
-------------------------------------------------------
a non-zero exit code.. Per user-direction, the job has been aborted.
-------------------------------------------------------
```

Not sure if the failure was fallout from the NOX failure maybe, or another spurious fail?

rppawlo:
@ndellingwood - the nox test is an epetra based test - does not use amesos2.

mayrmt:
@iyamazaki The autotester seems to be happy now. Can you have another look at the code and let us know, if this can be merged or if further changes are necessary? Thanks in advance.

iyamazaki:
Thank you so much for the contributions, @vryy, and also being so patient with all my questions.

> I implemented it so based on observation of the solver driver provided along with Superlu_dist. I think either solution may have a performance implication but obviously not the accuracy. 

Since `R1` is computed based on `A` in `preOrdering`, I would apply `R1` before computing `R` in `numericalFactorization`. To me, it does not make sense to compute both `R1` and `R` based on original `A`, e.g., if you have computed `R1` and `C1` with `job=5` (and the matrix values have not changed), and if we compute `R` and `C` after applying `R1` and `C1`, then they should be all ones, but for original `A`, they may not be? Similarly, it does not make much sense to apply `R1` after applying `R` to `A`?

> For the latter question, I understand that Equil is the control flag for scaling at all. Please correct me if I'm wrong.

Are you saying, we always compute `R` if we have computed `R1`? I think these two options can be applied, independently? 

I can also test, but can you tell me how much of these have been tested? I would run `SimpleSolve` with different `job` values, and with/without `Equil`, using a simple matrix (e.g., anti-diagonal matrix with a(i,n-i+1) = i), and make sure the matrix has expected values after all the permutation and scaling are applied (it should be identity)?

maxfirmbach:
I can also do another round of testing including the difficult examples were row permutation is necessary to solve the system.

vryy:
@iyamazaki you are actually right. We should compute `R1`, `C1` after computing and applying `R`, `C`. But note that `R1` and `C1` arise from row permutation for `job=5`. So matrix `A` has to be modified right at the `preOrdering` step. Does it violate the design of the linear solver? One solution is to create a copy of `A` at the `preOrdering` and then scale `A` again in `numericalFactorization`. The cost is one redundant copying and scaling.

For `Equil`, my understanding is if it's false, then no scaling at all is applied, neither `R`, `C` nor `R1`, `C1`. Therefore, for `job=5`, it will merely apply the row permutation without additional scaling. User should be aware about this behavior when using `SuperLU_Dist` solver, that they have to set `Equil=true` if they want `job=5`. We can take care to turn `Equil` on in the default setting, if it is not forced outside.

Right now we test with our physics problem of our in-house code. But I can also do unit-test with the anti-diagonal matrix as well. But we shall agree first between us on the problem above.

iyamazaki:
I am glad we are on the same page that the current way of applying`R1` and `C1` needs fixing. I think the main thing is to support your needs (though Amesos2 may end up not doing exactly what SuperLU_Dist does)? 

The simplest may be, in `numericalFact`,

1.  we first apply `R1` and `C1` if `job==5`, and
2.  then if `Equil==YES`, we compute and apply additional `R` and `C` scaling?

Can you then check if (`job=5` and `Equil=NO`) is sufficient for all your cases? If the current way of applying scaling in this PR is sufficient for your needs, I hope the above is good enough.

vryy:
@iyamazaki in the latest commit I compute and apply R, C (to a temporary matrix) before computing row permutation (job=5). In the `numericalFact`, R, C (computed from preOrdering) is applied to `A` if `Equil=YES`; R1, C1 will be applied subsequently if job=5. For my cases the `Equil` flag is not needed, it can work with pure row permutation. However, `Equil=YES` is required for the anti-diagonal matrix test (job=5).

packages/amesos2/src/Amesos2_Superludist_def.hpp
```diff
@@ -332,8 +332,36 @@ namespace Amesos2 {
...
+        SLUD::int_t info = 0;
+
+        // Compute scaling
+        data_.R.resize(this->globalNumRows_);
+        data_.C.resize(this->globalNumCols_);
+        function_map::gsequ_loc(&Acopy, data_.R.getRawPtr(), data_.C.getRawPtr(),
+                                &(data_.rowcnd), &(data_.colcnd), &(data_.amax), &info, &(data_.grid));
+
+        // std::cout << "R vector:" << data_.R << std::endl;
+        // std::cout << "C vector:" << data_.C << std::endl;
```
maxfirmbach, line 352: Can be removed right?

vryy: sure

mayrmt:
@iyamazaki Could you please have another look? Thanks.

iyamazaki:
Sorry for the delay. Thank you for the changes, they look good. I have one more comment. I also talked with @srajama1, but we really don't like the code that use numerical values in `symbolicFactorization`. 

So, I was wondering the following:

1. put your row-permutation code (e.g., [this](https://github.com/trilinos/Trilinos/blob/1c9c3f676b4a79d9788aee609bfb59ee37d45294/packages/amesos2/src/Amesos2_Superludist_def.hpp#L364-L412)) into a private member function, and call the function in `preOrdering`, but only for the row-permutation option without numerical value (`if (data_.largediag_mc64_job == 1 && data_.options.Equil == false)`), so also move all the `Equil` code from `preOrdering` back to `numericalFactorization_impl`.
2. if a user asked for row-permutation with numerical value, then skip `symbolicFactoctorization_impl` because we'll be calling it from `numericalFactorization_impl`
3. Now, add the row-permutation code in `numericalFactorization_impl`, after `Equil`, for the row-permutation that needs numerical values. If requested, then apply the matrix scaling and call the `symbolicFactorization` before continuing the numerical factorization.

I think this will fit the Amesos2 design best. Thank you again for the contributions, and taking time to integrate. I think this will be very useful.

vryy:
@iyamazaki the last two commits addressed your concern about row-permutation code that needs matrix values shall be put in `numericalFactorization`. Since the `preOrdering` can be skipped for `RowPerm == SLUD::LargeDiag_MC64`, I have to add a check in `SolverCore` to increment the counter only when the return code is `EXIT_SUCCESS`. Otherwise the internal `data_` will be deleted before it's allocated, leading to seg-fault error. I hope that these are fine.

iyamazaki:
Thank you, @vryy!! This looks great. I see some issues with scalar values, but also issues existed before this PR. So, I will have a small patch to fix those after this PR get merged. I can try this in a follow-up PR, but we could call `computeRowPermutationLargeDiagMC64` in symbolic if `LargeDiag_MC64` and `job==0`?

I like to run a couple of local tests, and then will approve!!

vryy:
> Thank you, @vryy!! This looks great. I see some issues with scalar values, but also issues existed before this PR. So, I will have a small patch to fix those after this PR get merged. I can try this in a follow-up PR, but we could call `computeRowPermutationLargeDiagMC64` in symbolic if `LargeDiag_MC64` and `job==0`?
> 
> I like to run a couple of local tests, and then will approve!!

I haven't checked with `job==0`, however with `LargeDiag_MC64`, it is restricted by `1 <= JOB <= 5.`, see [here](https://github.com/xiaoyeli/superlu_dist/blob/5c134712a0ab22775b1c801acaf42b05965a3215/SRC/dldperm_dist.c#L62).

iyamazaki (approved):
Sorry, I meant `job=1`, which does not use any numerical values, but I am now not sure if SuperLU_DIST supports it? In any case, these look good to me. Thank you so much for the changes, @vryy.

mayrmt:
@iyamazaki @vryy Thanks for iterating on this MR! Since it has been approved, I will merge now.

PR #11486, opened 2023-01-19
add congfig for clang 11 sems v2 build

fryeguy52:
@trilinos/framework 

## Motivation
Add clang-11 configuration to the ini PR/MM files that uses the sems v2 tpls.  This will greatly improve the usability of rhe ini files by reducing our dependence on two systems

## Testing
see these results on cdash which show builds passing all tests using the PR/MM infrastructure on nightly builds

https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Nightly&field2=buildstarttime&compare2=83&value2=2023-01-18&field3=buildstarttime&compare3=84&value3=2023-01-20&field4=buildname&compare4=63&value4=sems-v2

### DIFF
--- a/packages/framework/ini-files/config-specs.ini
+++ b/packages/framework/ini-files/config-specs.ini
@@ -1748,6 +1748,11 @@ use RHEL7_POST
 
 opt-set-cmake-var CMAKE_CXX_STANDARD STRING FORCE : 17
 
+[rhel7_sems-clang-11.0.1-openmpi-1.10.1-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_all]
+#uses sems-archive modules
+use rhel7_sems-clang-11.0.1-openmpi-1.10.1-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables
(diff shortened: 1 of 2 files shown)


DaveBeCoding (dismissed):

packages/framework/ini-files/config-specs.ini
```diff
@@ -1732,11 +1748,38 @@ use RHEL7_POST
 
 opt-set-cmake-var CMAKE_CXX_STANDARD STRING FORCE : 17
 
-[rhel7_sems-clang-11.0.1-openmpi-1.10.1-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_all]
```
sebrowne, line 1735: Getting rid of this will break the autotester if we don't make a simultaneous change in there I believe.  Could we leave it, then change the autotester, then delete it?

sebrowne (changes requested):

packages/framework/ini-files/config-specs.ini
fryeguy52: yes thanks for catching that

fryeguy52: @sebrowne done

sebrowne (approved):

Issue #11487, opened 2023-01-20
SEACAS: header "fmt/ostream.h" not found in Nalu and Albany builds


spdomin:
## Bug Report
In file included from /ascldap/users/spdomin/gitHubWork/scratch_build/install/gcc10.3.0/Trilinos_nightly_release/include/Ioss_StructuredBlock.h:16,
                 from /ascldap/users/spdomin/gitHubWork/scratch_build/install/gcc10.3.0/Trilinos_nightly_release/include/Ioss_SubSystem.h:35,
                 from /home/spdomin/gitHubWork/nightlyBuildAndTest/NaluNightly/src/InputOutputRealm.C:24:
/ascldap/users/spdomin/gitHubWork/scratch_build/install/gcc10.3.0/Trilinos_nightly_release/include/Ioss_ZoneConnectivity.h:14:10: fatal error: fmt/ostream.h: No such file or directory
   14 | #include <fmt/ostream.h>

Good:
NaluCFD/Nalu SHA1: 8959168ab0f53a39178694457d276d4b887bcf1a
Trilinos/develop SHA1: e9b4687e7a95c0626abd0bc64da33e42470895fe

Bad:
NaluCFD/Nalu SHA1: 8959168ab0f53a39178694457d276d4b887bcf1a
Trilinos/develop SHA1: 7ab1eeca884dbfff6eedfd906f7affa10c773b44

Perhaps @alanw0 could advise.

best,

alanw0:
Hmm, I'm not sure what might be going on here. @gsjaardema could this be caused by recent seacas changes?

jhux2:
Could be related to #11442.

jhux2:
Also affecting Albany.  @ikalash

alanw0:
A quick glance through InputOutputRealm.C seems to indicate that you are not using IOSS directly in that file, you are only using it through stk_io. Thus, you don't need to directly include an IOSS header.
Also glancing at Realm.C, it appears as though you are using Ioss::Property, so you do need to include Ioss_PropertyManager.h, but it doesn't look like Realm.C needs to include Ioss_SubSystem.h either.

spdomin:
Agreed. i will plan on removing the includes and wait to see how this story unfolds.

I seem to recall a tool that can be run to remove extra header includes. I should probably track this down and prune the 10 years of headers that appear via legacy.

jhux2:
[...]

PR #11489, opened 2023-01-23
SEACAS: Remove uses of fmt in include files

gsjaardema:
SEACAS uses the lib{fmt} library for output, but that library is not yet a Trilinos TPL, so external packages using Trilinos are not required to have the lib{fmt} include files available.

SEACAS in Trilinos uses lib{fmt} as a header-only library and there is an embedded version in SEACAS/Trilinos.

This commit removes all fmt-related includes from header files so applications using Trilinos do not need to have the fmt include files.

@trilinos/seacas


gsjaardema:
Should fix #11487

PR #11504, opened 2023-01-25
SEACAS/Stokhos: Some fixes for RDC enabled builds with HIP

seanofthemillers:
@trilinos/seacas
@trilinos/stokhos

## Motivation
These are a couple of quick fixes for a two build errors with HIP with RDC enabled.

To build with RDC, you need the flag:

```
-DSEACASExodus_UNIT_TEST_LINKER_LANGUAGE=C
```

We may want to set that as the default, but the code mentions that there are cases where a CXX linker is required for some of the exodus unit tests.


seanofthemillers:
@jjellio Here are a couple of fixes for the RDC builds with HIP. It looks like most of the other issues I came across have already been addressed.

jjellio:
I did this as my work around:

```
    "-DSEACASExodus_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASNemesis_CXX_FLAGS=-fno-gpu-rdc"
    #"-DSEACASIoss_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASChaco_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASAprepro_lib_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASSuplibC_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASSuplibCpp_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASAprepro_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASConjoin_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASEjoin_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASEpu_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASCpup_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASExodiff_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASExomatlab_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASExo_format_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASNas2exo_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASZellij_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASNemslice_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASNemspread_CXX_FLAGS=-fno-gpu-rdc"
    #"-DSEACASSlice_CXX_FLAGS=-fno-gpu-rdc"
    #"-DSEACAS_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASNemesis_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASExo_format_CXX_FLAGS=-fno-gpu-rdc"
    "-DSEACASExodiff_CXX_FLAGS=-fno-gpu-rdc"
```

This was cut from a build that enabled Fortran, which likely added more tools to the list.

psakievich:
@gdsjaar @bartlettroscoe Could we get a triage going on this PR? It is a high priority fix forExaWind  building with ROCM.

packages/seacas/libraries/nemesis/CMakeLists.txt
```diff
@@ -28,11 +28,14 @@ TRIBITS_ADD_LIBRARY(
   SOURCES ${SOURCES}
   )
 
+IF (${PACKAGE_NAME}_ENABLE_TESTS)
```
gsjaardema, line 31: I didn't think that this was needed if the `TRIBITS_ADD_EXECUTABLE_AND_TEST` function was used?

packages/seacas/CMakeLists.txt
```diff
@@ -31,6 +31,15 @@ ENDIF ()
 set(SEACASExodus_ENABLE_STATIC_DEFAULT "OFF")
 set(SEACASExodus_ENABLE_SHARED_DEFAULT "OFF")
 
+# Certain environments require either C or CXX linkers for these C unit tests
```
gsjaardema, line 34: The linker language is already specified in the Exodus unit tests in `packages/seacas/libraries/exodus/test/CMakeLists.txt`:
```
  FOREACH ( testName ${cbind_OLDTESTS} )
    TRIBITS_ADD_EXECUTABLE( ${testName} NOEXEPREFIX NOEXESUFFIX SOURCES ${testName}.c LINKER_LANGUAGE CXX)
    set_property(TARGET ${testName} PROPERTY C_STANDARD 99)
  ENDFOREACH ( testName )
```
Should this symbol be used there instead of what is already there, or should similar code just be added to the nemesis test CMakeLists.txt

packages/seacas/applications/exo_format/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ IF(NOT __LINUX_COMPILER_PGI)
     NOEXEPREFIX
     NOEXESUFFIX
     SOURCES ${SOURCES}
+    LINKER_LANGUAGE ${SEACASExodus_UNIT_TEST_LINKER_LANGUAGE}
```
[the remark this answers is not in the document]
seanofthemillers: Ah, good point. Is there another flag you would suggest, or should we just hard code this to 'C' for the linker language?

packages/seacas/libraries/nemesis/CMakeLists.txt
seanofthemillers: Do you know what the command is to turn off nemesis tests from the cmake config? Something like `SEACASNemesis_ENABLE_TESTS=OFF`? I wasn't able to narrow it down earlier, so I added this. I can remove it for this PR, since these changes will remove the need for the if.

seanofthemillers:
@jjellio To be honest, I'm in complete agreement. I am always a proponent of not adding obscure flags to build systems, especially if it may end up spanning as many SEACAS packages as you listed. If anyone asks the internal (AMD) JIRA ticket for the RDC compiler/linking fix is SWDEV-377156.

@psakievich It may be a better approach to inject James' cmake config fix (https://github.com/trilinos/Trilinos/pull/11504#issuecomment-1404258389) into your spack package. If you agree I will go ahead and close this PR. I can move the Stokhos changes to another issue.

packages/seacas/libraries/nemesis/CMakeLists.txt
bartlettroscoe: > I didn't think that this was needed if the `TRIBITS_ADD_EXECUTABLE_AND_TEST` function was used?

As documented at the bottom of the section:

* https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-add-executable-and-test

> Finally, the tests are only added if tests are enabled for the package (i.e. [${PACKAGE_NAME}_ENABLE_TESTS](https://tribitspub.github.io/TriBITS/users_guide/index.html#package-name-enable-tests) = ON) and other criteria are met. But the test executable will always be added if this function is called, regardless of the value of `${PACKAGE_NAME}_ENABLE_TESTS`. To avoid adding the test (or example) executable when `${PACKAGE_NAME}_ENABLE_TESTS=OFF`, put this command in a subdir under test/ or example/ and that subdir with [tribits_add_test_directories()](https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-add-test-directories) or [tribits_add_example_directories()](https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-add-example-directories), respectively.

bartlettroscoe:
> I don't understand using a C-linker for C++ apps. This just seems wrong. I think the problem is AMD's RDC tool chain is not correctly handling C++ codes that have no HIP. (adding -fno-gpu-rdc disables RDC , and these link perfectly fine with C++). This seems like gunk to work around AMD, is it really something that should be committed long term to a code base? (I don't think so). Also, if you do link via C - how is the C++ library getting added?

My experience is that you should always be linking with the C++ linker/compiler if there is any C++ libraries being linked in upstream.  Is there any harm to linking C and even Fortran programs using the C++ linker/compiler even if there are no C++ binaries upstream?

seanofthemillers:
Alright, I'm closing. Thanks @jjellio @bartlettroscoe @gsjaardema for looking at this.

gsjaardema:
> I don't understand using a C-linker for C++ apps. This just seems wrong. I think the problem is AMD's RDC tool chain is not correctly handling C++ codes that have no HIP. (adding -fno-gpu-rdc disables RDC , and these link perfectly fine with C++). This seems like gunk to work around AMD, is it really something that should be committed long term to a code base? (I don't think so). Also, if you do link via C - how is the C++ library getting added?

I don't know of any place that I am using a C-linker for C++ Apps.  There may be some places that we are using a C++ linker for C-apps, but not vice-versa.  If you find any (C-linker with C++ app), let me know and I will see if I can explain or fix.

packages/seacas/applications/exo_format/CMakeLists.txt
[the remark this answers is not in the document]
gsjaardema: I'm not sure why it needs the linker language specified, but maybe it isn't picking it up correctly, so just adding linker language c would be enought.  I have no problems with that.

packages/seacas/libraries/nemesis/CMakeLists.txt
gsjaardema: OK, I misunderstood this then.  I thought I used to have these all inside if blocks so they were only built if tests were enabled and that the `TRIBITS_ADD_EXECUTABLE_AND_TEST` took care of that, so I removed the if blocks.  I will put those back so test executables are only built if tests are enabled...

PaulMullowney: The origin of this problem is in Nalu-Wind. Our Cmake options are shown here:
https://github.com/sandialabs/spack-manager/blob/main/repos/exawind/packages/trilinos/package.py#L114
However, we still are forced to wrap the nemesis unit test in order to build, i.e.:
+IF (${PACKAGE_NAME}_ENABLE_TESTS)
 TRIBITS_ADD_EXECUTABLE_AND_TEST(
   ne_ctest_wrap
   NAME ne_ctest_wrap
   SOURCES ne_ctest_wrap.c
   NUM_MPI_PROCS 1
 )
+ENDIF()
Is there something wrong with the cmake options?

gsjaardema: > The origin of this problem is in Nalu-Wind. Our Cmake options are shown here: https://github.com/sandialabs/spack-manager/blob/main/repos/exawind/packages/trilinos/package.py#L114 However, we still are forced to wrap the nemesis unit test in order to build, i.e.: +IF (${PACKAGE_NAME}_ENABLE_TESTS) TRIBITS_ADD_EXECUTABLE_AND_TEST( ne_ctest_wrap NAME ne_ctest_wrap SOURCES ne_ctest_wrap.c NUM_MPI_PROCS 1 ) +ENDIF() Is there something wrong with our cmake options?

So is the problem that `ne_ctest_wrap` does not build (and perhaps adding the `LINKER_LANGUAGE` would fix it, or is there another reason that I am missing?

PaulMullowney: The test fails to build because the linker is messed up with rdc. My question is whether we should expect
-DSEACASNemesis_ENABLE_TESTS=OFF
to work. If yes, then I think those changes I show above are necessary.

bartlettroscoe: NOTE: If only a small number of specific test executables don't build, you can add configure-time options to disable defining those executables by setting:

```
-D <exeTargetName>_EXE_DISABLE=ON
```

(see [here](https://tribits.org/doc/TribitsBuildReference.html#disabling-specific-test-executable-builds)).

But if the intent was to not build the test executables if `<Package>_ENABLE_TESTS=OFF`, then wrapping the calls in if() statements or (better) in a directory added using [`tribits_add_test_directories()`](https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-add-test-directories).

PaulMullowney: That was the intent. 
https://github.com/trilinos/Trilinos/blob/master/packages/seacas/libraries/nemesis/CMakeLists.txt#L31
seems to wrap the test. 

However, it still gets compiled, despite having -DSEACASNemesis_ENABLE_TESTS=OFF, and then breaks because of the broken linker.

PR #11505, opened 2023-01-26
Intrepid2: remove extraneous Kokkos finalize

mperego:
Remove extraneous Kokkos finalize form an Intrepid2 tests. 
It will throw an error with Kokkos 4.

@trilinos/intrepid2


ndellingwood (approved):
lgtm!

PR #11507, opened 2023-01-26
SEACAS: Make getline termios symbols unique across uses

gsjaardema:
@trilinos/seacas 

## Motivation

Aprepro, suplib, and IOSS all contain an embedded getline function that defined a `new_termios` and `old_termios` global symbol which could cause conflicts on some platforms.  Changed each instance to a new name that is unique for each use.


rppawlo (approved):
Thanks @gsjaardema !

PR #11514, opened 2023-01-27
MueLu: read matrix and parallelize without requiring rowmap file

jhux2:
Enable MueLu's Driver to read a matrix from file and parallelize using a uniform distributed map.  This is useful when there's no row map file.

@trilinos/muelu 

## Stakeholder Feedback

N/A

## Testing

Tested locally.


GrahamBenHarper (dismissed):
Ah this is really nice to have. However, what are the odds that this generates a totally garbage map for, say, a matrix generated as the result of a Panzer discretization?

jhux2:
> Ah this is really nice to have. However, what are the odds that this generates a totally garbage map for, say, a matrix generated as the result of a Panzer discretization?

This is definitely warranty-free and doesn't handle PDEs > 1.  And yeah, it assumes the app has assemble the matrix in some sane fashion.

GrahamBenHarper (approved):

PR #11516, opened 2023-01-30
Panzer and Thyra: Fixed bug related to evaluation order of function arguments for mini-em

skaramati:
@csiefer2 

### Summary
This PR fixes an issue that is caused when running .... on a mixed cluster of x86 nodes and arm64 nodes. This is because the order of evaluation of arguments in a C++ expression is unspecified, and the same compiler but with different backends (for example, same gcc + version, but one targeting x86 and the other arm64) would behave differently.

### Detailed Description
I encountered a problem while running "Trilinos/packages/panzer/mini-em/example/BlockPrec/main.cpp" simultaneously on x86 and arm processors. The issue is an incompatibility error in vector size caused by the unspecified order of evaluation of arguments in a C++ function call. Since these arguments are function calls with MPI collective inside taking the range space vectors or domain space vectors as input, the different order of function argument evaluation on two hardware platforms causes MPI collective operation to use a range space vector on some processes while using a domain space vector on others, causing inconsistency.

### Fix
In this pull request, I have addressed the issue by first evaluating the function arguments in a predetermined order (before the function call) and then passing the evaluated values to the function. This ensures consistent behavior of the compiler/runtime across x86 and arm processors.


cgcgcg:
@skaramati I don't understand the change. What was the bug?

skaramati:
> @skaramati I don't understand the change. What was the bug?

@cgcgcg I updated the pull request with the description and what the issue is. Thanks.

cgcgcg:
Oh wow. I'm actually surprised that that's the only thing that fails under these conditions.

csiefer2:
@skaramati I'm looking into the test failure.  Can you give me permission to push the fix to your fork?

skaramati:
> @skaramati I'm looking into the test failure. Can you give me permission to push the fix to your fork (if needed_?

@csiefer2 Sure, I granted access to you in my repo.

csiefer2:
@skaramati Thanks! At the moment, I think the failures are random ones, unrelated to your code.  But the testing system will tell me one way or the other...

csiefer2 (approved):

csiefer2:
Four different random failures so far.  Yum :(

csiefer2:
@trilinos/framework No failing tests reporting by cdash the last time.  The Trilinos_PR_gcc-8.3.0-debug build listed as failed above did not post to cdash at all.

sebrowne:
@csiefer2 I just checked and we're getting hit with a module load error (very early in the process, so no reporting) from the SEMS modules.  I've made them aware of it, and hopefully they'll get back to me with an ETA for a fix soon.  I may disable the autotester until then to reduce confusion.

PR #11518, opened 2023-01-30
Tacho : fix issues with multiple streams on AMD GPU

iyamazaki:
@trilinos/shylu 

## Motivation

This PR aims to fix issues using multiple streams on an AMD GPU

- On an AMD GPU, we seem to require a separate handle for each stream.

## Testing

Tested on MI100.

## Additional Information

On an NVIDIA GPU, we still use one handle for all the streams.


ndellingwood:
The gcc/8.3.0 builds did not post to cdash in the most recent run, maybe that is the reason for the reported failure? Retesting

crdohrm (approved):

PR #11520, opened 2023-01-31
Piro: fix sensitivity computation when ROL is not enabled

mperego:
@trilinos/piro

## Motivation

Piro adjoint sensitivity computation was broken if ROL was not enabled. This PR fixes this.

This PR addresses https://github.com/sandialabs/Albany/issues/867.


kliegeois (approved):
It looks good to me, thanks!

PR #11521, opened 2023-01-31
MueLu: fix stack timer usage with reruns

### LINKED ISSUE #11522: MueLu: Driver fails in rerun mode with stack timers
## Bug Report
@trilinos/muelu 

### Description
When using the MueLu driver in "rerun" mode, i.e., multiple XML inputs in a single file, and with the option "--stack-timer", the driver fails with the error

```
 Base_Timer:stop Failed timer not running
```

The issue is that the Stack Timer's base timer is stopped after Run1 but never restarted.

### Steps to Reproduce

```
 ./MueLu_Driver.exe --xml=scaling-with-rerun.xml  --stacked-timer
```

jhux2:
In the case of multiple runs using the same matrix data, ensure the base timer is restarted.
Add timers for coordinate and RHS reading.

Fixes #11522.

@trilinos/muelu 
## Motivation

It's very handy to be able to gather stacked timer data from the MueLu driver when using its "rerun" capability
with large matrix data files.

## Stakeholder Feedback

N/A

## Testing

Verified bug locally and that this PR fixes it.

### DIFF
--- a/packages/muelu/test/scaling/Driver.cpp
+++ b/packages/muelu/test/scaling/Driver.cpp
@@ -454,6 +454,9 @@ MueLu::MueLu_AMGX_initialize_plugins();
         }
         if (runList.isParameter("solver")) solveType = runList.get<std::string>("solver");
         if (runList.isParameter("tol"))    tol       = runList.get<double>     ("tol");
+
+        if (useStackedTimer && runCount > 1)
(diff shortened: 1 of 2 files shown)


csiefer2 (approved):

GrahamBenHarper (approved):
I always like Mue driver enhancements 😊

PR #11523, opened 2023-01-31
Framework: Use default C++ standard for PR/MM testing

sebrowne:
@trilinos/framework

## Motivation
We want to use the default C++ standard for most testing, unless we're in the process of advancing standards.  In particular, some of the build configs were still set to C++14 which caused the nightly RDC build to fail.


sebrowne:
Seems to be a spurious failure of [FEI_fei_ubase_MPI_3](https://trilinos-cdash.sandia.gov/test/19153878)

Retesting.

sebrowne:
Concerning, the same test failed again.  Perhaps moving toolchains caused this failure.  I'll attempt to replicate.

jwillenbring (approved):
In addition to advancing standards, there may be situations where we want to test Trilinos against multiple standards, but generally I agree that in most cases we want to go with the default.

PR #11533, opened 2023-02-01
Teuchos: StackedTimer Watchr documentation, date override

brian-kelley:
@trilinos/teuchos 

## Motivation

- In doxygen for ``StackedTimer::reportWatchrXML()``, document the optional ``TRILINOS_GIT_SHA`` environment variable (just a comment change)
- Add a new variable ``WATCHR_BUILD_DATE`` to override the build date in the report filename and the XML header of the report. If unset, the date is just when the build happened (this is the current behavior). This lets us run on a Trilinos version from the past, and have the date on the report match the date of the version.

This is needed because some recent performance dashboard data was generated from incorrect builds and needs to be regenerated.

## Testing

Tested by running a performance test and outputting Watchr reports - once with the default date (today), and once with the date overridden. This created two separate reports with the correct names and contents.


rppawlo (approved):

brian-kelley:
@rppawlo Could you approve this again? I just squashed the 2 commits to 1 and fixed the author name on that commit. Thanks!

rppawlo (approved):

PR #11534, opened 2023-02-01
Fix .mailmap name

brian-kelley:
@trilinos/framework 

## Motivation

Change the spelling of my name in ``.mailmap``. This gets used for commit author, and overrides what I set in ``git config --global user.name``.


bartlettroscoe (approved):
@brian-kelley, this is your entry so you should be able to make this to your liking.

brian-kelley:
Thanks @bartlettroscoe !

PR #11540, opened 2023-02-03
Cmake: Remove Epetra stack from Rocketman testing

GrahamBenHarper:
@cgcgcg 

Rocketman nightlies are struggling to build Ifpack1 and Amesos1 due to errors related to SuperLU. These builds really don't need the Epetra stack, and I'd rather just see if the build succeeds without the Epetra stack as opposed to troubleshooting an issue that may potentially require me to rebuild SuperLU.


jhux2 (approved):
lgtm

PR #11557, opened 2023-02-10
Ifpack2: BlockTriDi StackedTimer/Watchr support

brian-kelley:
@trilinos/ifpack2 

## Motivation

Lets the Ifpack2_BlockTriDiagonalSolver be used to drive performance tests that are plotted in Watchr. Adds ``--withStackedTimer`` option to enable stacked timers, and ``--problemName="..."`` option to put a description of the matrix/problem in the Watchr plot title.

Also fix an issue in flat timer mode (``--withoutStackedTimer``) where the "Total" timer would always read 0, because it was never stopped.

## Testing

Tested by comparing new and old flat timer output, and then by running with stacked timer and outputting a Watchr report.

### DIFF
--- a/packages/ifpack2/example/BlockTriDi.cpp
+++ b/packages/ifpack2/example/BlockTriDi.cpp
@@ -11,13 +11,14 @@
 #include <Teuchos_CommandLineProcessor.hpp>
 #include <Teuchos_ParameterXMLFileReader.hpp>
 #include <Teuchos_TimeMonitor.hpp>
+#include "Teuchos_StackedTimer.hpp"
 #include <Teuchos_StandardCatchMacros.hpp>
 
 namespace { // (anonymous)
 
 // Values of command-line arguments.
 struct CmdLineArgs {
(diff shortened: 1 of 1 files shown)


packages/ifpack2/example/BlockTriDi.cpp
```diff
@@ -459,8 +498,8 @@ main (int argc, char* argv[])
   TEUCHOS_STANDARD_CATCH_STATEMENTS(verbose, std::cerr, success);
   */
 
-  
-  std::cout<<"End Result: TEST PASSED"<<std::endl;
+  comm->barrier();
```
brian-kelley, line 501: This stops the "TEST PASSED" text being printed 4 times, interleaved randomly with the timer output.

brian-kelley:
@csiefer2 For that SPARC problem, you'll want to add a ``--problemName=""`` flag describing what problem is being run (this can just be pushed, don't worry about making a PR)

csiefer2 (approved):

PR #11558, opened 2023-02-10
Shylu: disable gtest install

rppawlo:
@trilinos/shylu 

## Motivation

When Shylu's tests are enabled, it pulls in a copy of gtest. Gtest has a flag INSTALL_GTEST that defaults to ON. If the shylu tests are enabled, gtest will get installed along with trilinos. This can conflict with an application's version of gtest. We need to change this default to OFF. This use case is documented in the gtest and tells users to do this. I've just copied this change from sacado which is also pulling in gtest but was setting the default correctly.


rppawlo:
@iyamazaki - can you review this?

iyamazaki:
Thank you, @rppawlo!! This looks good to me, but I like to get @ndellingwood's input as well.

ndellingwood (approved):

PR #11569, opened 2023-02-16
MueLu: Add Matrix-Free 1D Laplacian Multigrid

GrahamBenHarper:
@trilinos/muelu 
@cgcgcg @pohm01

I'm moving pieces of my matrix-free multigrid code into MueLu incrementally so that collaborators can use it. `example_01` is a 1D Laplacian. This will make it easier for everybody to keep track of updates. This is intentionally not built at the moment, as some pieces later have dependencies on Intrepid2/STK/Panzer, and those dependencies will have to be managed carefully or simply not built.

I recently switched everything from Tpetra to Xpetra, so there are a couple cleanup actions needed in future PRs, but I'm ignoring that for now.

### DIFF
--- a/packages/muelu/research/graham/mf_example_01.cpp
+++ b/packages/muelu/research/graham/mf_example_01.cpp
@@ -1,0 +1,642 @@
+// @HEADER
+//
+// ***********************************************************************
+//
+//        MueLu: A package for multigrid based preconditioning
+//                  Copyright 2012 Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the Corporation nor the names of the
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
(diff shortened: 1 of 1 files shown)


packages/muelu/research/graham/mf_example_01.cpp
```diff
@@ -0,0 +1,642 @@
...
+#include "MueLu_Hierarchy.hpp"
+#include "MueLu_HierarchyManager.hpp"
+#include "MueLu_ParameterListInterpreter.hpp"
+#include "MueLu_AmalgamationFactory.hpp"
+#include "MueLu_CoordinatesTransferFactory.hpp"
+#include "MueLu_UncoupledAggregationFactory.hpp"
+#include "MueLu_AggregationExportFactory.hpp"
+#include "MueLu_Factory.hpp"
+
+#include "mpi.h"
```
cgcgcg, line 117: Why is that needed?

packages/muelu/research/graham/mf_example_01.cpp
```diff
@@ -0,0 +1,642 @@
...
+  typedef Kokkos::CudaSpace DeviceMem;
+  typedef Kokkos::Compat::KokkosCudaWrapperNode DeviceNode;
+#else
+  typedef Kokkos::Serial DeviceExec;
+  typedef Kokkos::HostSpace DeviceMem;
+  typedef Kokkos::Compat::KokkosSerialWrapperNode DeviceNode;
+#endif
+typedef Kokkos::Device<DeviceExec,DeviceMem> DeviceDevice;
+
+#define PI 3.141592653589793238463
```
cgcgcg, line 135: ?

packages/muelu/research/graham/mf_example_01.cpp
GrahamBenHarper: It probably isn't. This was built externally and included a header with roughly 120 separate include statements across all Trilinos packages, so I copy-pasted the list and then skimmed it and threw out everything that didn't look important.

packages/muelu/research/graham/mf_example_01.cpp
GrahamBenHarper: There's some nonsense there to have it pick up on custom-defined device/execution combos, but it should inherit from the Kokkos config now. `PI` is just a define since I can set an analytic sin(PI*x) solution, with 21 digits since I'm extra :)

cgcgcg:
We do. But there is no way we can make Panzer a dependency.

GrahamBenHarper:
Well, the 1D Laplace example is fine for MueLu because the operator doesn't rely on meshes or discretizations. It's just the typical [-1 2 -1] stencil. If I wanted to spend time messing with TriBITS/CMakeLists, I could make it so MueLu builds it as an example when experimental is enabled.

I'll ask the Panzer team if it's possible to turn other matrix-free apps with Panzer dependencies into Panzer examples.

cgcgcg (approved):

Issue #10355, opened 2022-03-22
Upgrade CMake to most current version possible for Trilinos 14.0


bartlettroscoe:
CC: @trilinos/framework, @keitat, @trilinos/kokkos, @trilinos/kokkos-kernels 

**Child Issues:**

* #10574

**Associated Internal Issues:**

* [SEPW-520](https://sems-atlassian-son.sandia.gov/jira/browse/SEPW-520): Get CMake 3.22 installed on all Trilinos platforms
* [TRILINOSHD-184](https://sems-atlassian-son.sandia.gov/jira/servicedesk/customer/portal/7/TRILINOSHD-184): Coordinate upgrading minimum version of CMake from 3.17 to 3.22

### Description

The [TriBITS refactoring work](https://github.com/TriBITSPub/TriBITS/projects/2) and efforts to [reduce the size of TriBITS to use raw CMake](https://github.com/TriBITSPub/TriBITS/issues/411) would be helped if TriBITS and Trilinos could upgrade the minimum required version of CMake from 3.17 to something much newer.  Ideally, we would upgrade to CMake 3.22 (released in November 18, 2021).

A sampling of the new CMake features we could exploit in TriBITS and Trilinos include:

* Use new imported targets generated by `FindHDF5.cmake` not available in CMake 3.17 to simplify and robustify the Trilinos `FindTPLHDF5.cmake` module (see [below](https://github.com/trilinos/Trilinos/issues/10355#issuecomment-1196843040)). (CMake 3.20) 
 * `cmake_langauge()` command: Better handle lists and function forwarding and allow for function pointers to simplify TriBITS implementation. (CMake 3.18)
 * Set ctest test Labels and Details fields at runtime (CMake 3.22).  (Allows custom details about failed tests.)
* `file(CHMOD_RECURSE …)` to replace [Linux/bash-specific commands used to fix directory permissions](https://tribits.org/doc/TribitsBuildReference.html#setting-install-ownership-and-permissions) on SNL systems used in Trilinos installation testing and SPARC and EMPIRE + Trilinos Integration testing.
* Support for `string(JSON)` for native JSON processing to allow more robust handling of data in input files (CMake 3.19)
* Improved HIP support may benefit Kokkos and Trilinos efforts on Frountier (CMake 3.21 for improved HIP support)
[...]

PR #11583, opened 2023-02-21
Update CMake min version from 3.17 to 3.23 (#10355) 

bartlettroscoe:
This is needed before the Trilinos 14.0 branch is created or we will be stuck with CMake 3.17 for at least another year.  See #10355.

All of the major systems at SNL and LLNL should already have CMake 3.23+.


bartlettroscoe:
@ccober6 and @sebrowne, does someone want to send out an announcement to Trilinos developers, users, and stakeholders?  (Past experience says almost no one will pull the branch and test.) Otherwise, I say we let this merge PR and see what happens.  If there is a critical issue, it is trivial to revert this PR and address the issue.

sebrowne (approved):

sebrowne:
I'm fine sending out an announcement, do you want to do it prior to merging this?  @ccober6 I can send you a draft and then send something out later today if that works.

ccober6:
@sebrowne , sounds good!  Go ahead and send me a draft announcement.

GrahamBenHarper:
FYI the build failure seems to be related to version incompatibilities of SuperLU and ML

bartlettroscoe:
CC: @trilinos/ml, @trilinos/framework, @sebrowne 

The non-CUDA PR builds shown [above](https://github.com/trilinos/Trilinos/pull/11583#issuecomment-1439250498) are all failing showing the compile errors:

```
/scratch/trilinos/jenkins/ascic142/workspace/Trilinos_PR_clang-11.0.1/Trilinos/packages/ml[Smoother/ml_csolve.c:17](https://github.com/trilinos/Trilinos//blob/master/Smoother/ml_csolve.c#L17):10: fatal error: 'dsp_defs.h' file not found
#include "dsp_defs.h"
         ^~~~~~~~~~~~
1 error generated.
```

and

```
/scratch/trilinos/jenkins/ascic142/workspace/Trilinos_PR_clang-11.0.1/Trilinos/packages/ml[Smoother/ml_superlu.c:19](https://github.com/trilinos/Trilinos//blob/master/Smoother/ml_superlu.c#L19):10: fatal error: 'dsp_defs.h' file not found
#include "dsp_defs.h"
         ^~~~~~~~~~~~
1 error generated.
```

This must be related to the change in a CMake policy by changing the minimum version CMake from 3.17 to 3.23 (because the version of CMake 3.24.4 did not change).  Looking over the policies for CMake 3.18 through 3.23 [here](https://cmake.org/cmake/help/v3.24/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-23), I don't see anything that jumps out.

It looks like someone will have to debug this.  Who will that be?

bartlettroscoe:
The header `dsp_defs.h` appears to be a Linux Kernel header?

* https://docs.huihoo.com/doxygen/linux/kernel/3.7/dsp__defs_8h_source.html

?

Cloning the SuperLU git repo on GitHub, the name `dsp_defs` only shows up in commits:

```
$ git log-short --name-status  -Sdsp_defs
b93b4e0 "update DOC/html doxygen files."
Author: Xiaoye Li <xsli@lbl.gov>
Date:   Fri Apr 8 18:53:17 2016 -0700 (7 years ago)

D       DOC/html/slu__ddefs_8h-source.html

851ee77 "Initial import"
Author: xiaoye <xiaoye@80d2f580-31aa-11e5-bf80-072c59723c4c>
Date:   Fri Jul 24 02:28:27 2015 +0000 (8 years ago)

A       DOC/html/slu__ddefs_8h-source.html
A       SRC/slu_ddefs.h
```

Why would ML be including a Linux Kernel header?

Anyway, I have the error reproduced on a local machine and I will take a look.

sebrowne:
@bartlettroscoe I have as well, I'm happy to debug this if you'd like.

bartlettroscoe:
@sebrowne, I have determined that if you change back to:

```
CMAKE_MINIMUM_REQUIRED(VERSION 3.17.0 FATAL_ERROR)
```

the define `SUPERLU` never gets defined in ML.

I verified this by adding the line:

```
$ git diff -- packages/ml/
diff --git a/packages/ml/src/Smoother/ml_smoother.c b/packages/ml/src/Smoother/ml_smoother.c
index 440a5f4b0c5..c0cf19678ba 100644
--- a/packages/ml/src/Smoother/ml_smoother.c
+++ b/packages/ml/src/Smoother/ml_smoother.c
@@ -180,6 +180,7 @@ int ML_permute_for_dgetrs_special(double *Z[], int Nblocks, int blocksize,
 /* ------------------------------------------------------------------------- */
 
 #if defined(SUPERLU)
+This will not compile!
 #include "dsp_defs.h"
 #include "util.h"
 #else
```

and the code still compiles.  That proves that the macro `SUPERLU` is not defined with:

```
CMAKE_MINIMUM_REQUIRED(VERSION 3.17.0 FATAL_ERROR)
```

but it does get defined with:

```
CMAKE_MINIMUM_REQUIRED(VERSION 3.23.0 FATAL_ERROR)
```

Don't now why this is the case but that is what is happening.

My guess is that the files `dsp_defs.h` and `util.h` existed in an old version of SuperLU but ML has not supported enabling SuperLU support for many years (perhaps to when CMake was first adopted for Trilinos back in 2008).

The solution is likely to just set:

```
#undef SUPERLU
```

in the ML code or just rip out all of the ifdefs `SUPERLU` since it has not been tested for many years.

ccober6:
@bartlettroscoe, probably should double check this with @trilinos/ml developers.

bartlettroscoe:
> @bartlettroscoe, probably should double check this with @trilinos/ml developers.

@csiefer2, @jhux2, what is the deal with the `SUPERLU` ifdefs in ML?  It looks like these have not been enabled for many years.

sebrowne:
@bartlettroscoe I show cmake_packages/ML/MLConfig.cmake as having set(HAVE_ML_SUPERLU "ON"), which is then propagating down into the configured headers.  Still trying to figure out how it got turned on.  Perhaps something to do with newer scoping rules in newer CMake?

bartlettroscoe:
> @bartlettroscoe I show cmake_packages/ML/MLConfig.cmake as having set(HAVE_ML_SUPERLU "ON"), which is then propagating down into the configured headers. Still trying to figure out how it got turned on. Perhaps something to do with newer scoping rules in newer CMake?

Yes, the ML configure is defining `HAVE_ML_SUPERLU` but I can't see how `SUPERLU` gets defined.  That is the mystery that needs to be solved.

sebrowne:
Okay, here's more:

(rhel7_sems-v2-gnu-8.3.0-serial) $ grep -r "ML_ENABLE_Super" *
CMakeCache.txt:ML_ENABLE_SuperLU:STRING=
CMakeCache.txt:ML_ENABLE_SuperLU1_0:BOOL=OFF
CMakeCache.txt:ML_ENABLE_SuperLU2_0:BOOL=OFF

Note the type of ML_ENABLE_SuperLU, and the value.  Still hunting this.

sebrowne:
> Yes, the ML configure is defining `HAVE_ML_SUPERLU` but I can't see how `SUPERLU` gets defined. That is the mystery that needs to be solved.

I believe it's in packages/ml/src/Include/ml_common.h

sebrowne:
Also, confirmed that HAVE_ML_SUPERLU (cmake variable) is defined and set to ON _prior_ to this snippet in packages/ml/CMakeLists.txt:

```
TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_SuperLU1_0
  HAVE_ML_SUPERLU
  "Enable the SUPERLU functionality."
  OFF )
```

bartlettroscoe:
Okay, I have verified that with:

```
CMAKE_MINIMUM_REQUIRED(VERSION 3.17.0 FATAL_ERROR)
```

you get the configured header with `HAVE_ML_SUPERLU` **not** enabled: 

```
$ grep "SUPERLU" packages/ml/src/ml_config.h
/* #undef HAVE_ML_SUPERLU */
/* #undef HAVE_ML_SUPERLU2_0 */
#define HAVE_ML_SUPERLU4_0
/* #undef HAVE_ML_SUPERLU_DIST */
```

but with:

```
CMAKE_MINIMUM_REQUIRED(VERSION 3.23.0 FATAL_ERROR)
```

you get the configured header with `HAVE_ML_SUPERLU` enabled: 

```
$ grep "SUPERLU" packages/ml/src/ml_config.h
#define HAVE_ML_SUPERLU
/* #undef HAVE_ML_SUPERLU2_0 */
#define HAVE_ML_SUPERLU4_0
/* #undef HAVE_ML_SUPERLU_DIST */
```

My guess is that this is related to the line of code in pakages/ml/CMakeLists.txt:

```
TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_SuperLU1_0
  HAVE_ML_SUPERLU
  "Enable the SUPERLU functionality."
  OFF )
```

But TriBITS is setting `HAVE_ML_SUPERLU` automatically as per:

* https://tribitspub.github.io/TriBITS/users_guide/index.html#how-to-add-a-new-tribits-external-package-tpl

My guess is that a new CMake policy about how CMake cache and non-cache vars are being handled is the cause of this.  My guess is the policy [CMP0126](https://cmake.org/cmake/help/v3.24/policy/CMP0126.html) introduced in CMake 3.21 is related to this.

My guess is that due to this, `HAVE_ML_SUPERLU` was never defined in the generated `ml_config.h` file until we tried updating the minimum version of CMake beyond 3.21.

sebrowne:
@bartlettroscoe I concur, that's exactly the policy I was just looking at.

We don't want this enabled for downstream customers, so I don't think fixing it view setting disables, etc. in the config-specs.ini file for the PR tester is the right approach, this is something that needs to be fixed in the build system one way or another.  Probably just forcibly disabling ML SuperLU support is the best approach, so I suppose with that we're back to waiting for input from developers who have more an idea of what is going on with ML.

bartlettroscoe:
@jhux2, @csiefer2,

Based on the above analysis, it is likely that ML support for SuperLU was likely **never** enabled in the CMake build system for Trilinos!

It would seem that the best approach for now is to disable support for SuperLU in ML so we can upgrade the min CMake version and then someone can get SuperLU support in ML working again in the future if there is a desire to do this.

Does that sound like a plan?

bartlettroscoe:
Some more info on this and policy [CMP0126](https://cmake.org/cmake/help/v3.24/policy/CMP0126.html) in this case ...

When TriBITS sets `HAVE_<PACKAGE_UC_NAME>_<OPT_DEP_NAME>` (i.e. `HAVE_ML_SUPERLU`) as a result of `${packageName}_ENABLE_${optDepPkg}` (i.e. `ML_ENABLE_SuperLU`), it does so as a project-level local variable as shown here:

https://github.com/TriBITSPub/TriBITS/blob/5470aab7408e074419127c26f2a2d0782e6f0289/tribits/core/package_arch/TribitsAdjustPackageEnables.cmake#L1122

But when the function `tribits_add_option_and_define()` sets the macro  `HAVE_ML_SUPERLU` in this case, it does so as  cache variable as shown at:

https://github.com/TriBITSPub/TriBITS/blob/5470aab7408e074419127c26f2a2d0782e6f0289/tribits/core/package_arch/TribitsAddOptionAndDefine.cmake#L83

it does so as a `cache( ... INTERNAL ...)` cache var.

For some reason, when the `ml_config.h` file gets configured (with `configure_file()`) with the `OLD` behavior it is using the value of the cache var `ML_ENABLE_SuperLU` (which was always `OFF`) but with the `NEW` behavior, it uses the value of the project-level local variable `ML_ENABLE_SuperLU` which is set to `ON` when `ML_ENABLE_SuperLU` is `ON`.

So the intent of the ML developers seems to have been that unless `ML_ENABLE_SuperLU1_0` is set to `ON` by the user, then SuperLU support should not be enabled.

Given this, I think this is how you address this:

1. Add a check in `tribits_add_option_and_define(<optionName> <defineName> ... )` that errors out if the variables `<optionName>` or `<defineName>` already exist.

2. Change the var set in `TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_ENABLE_SuperLU1_0  HAVE_ML_SUPERLU ...)` from `HAVE_ML_SUPERLU` to `HAVE_ML_SUPERLU1_0` and change it in `packages/ml/src/Include/ml_common.h`.

Doing "1" above would help to catch other potential mistakes like this and doing "2" would disable SuperLU support in ML according to the wishes of the ML developers when they wrote the `packages/ml/CMakeLists.txt` file.

The other quick fix is to just set the policy of [CMP0126](https://cmake.org/cmake/help/v3.24/policy/CMP0126.html) to `OLD` at the top of the file `packages/ml/CMakeLists.txt`.  The preserves the existing (buggy) behavior so we can move on.

bartlettroscoe:
FYI: I just pushed the commit [`3e8c0ae` (#11583)](https://github.com/trilinos/Trilinos/pull/11583/commits/3e8c0aedd36308825d7a56f6cf314bb0a048b9bc) that shows that a number of Trilinos packages are mistakenly trying to redefine TriBITS-defined vars.

Boy, handling vars in CMake is so tricky, especially when they change behavior like this.

rstumin:
@bartlettroscoe @sebrowne  We concur that it has probably been a long time since ML and superLU have worked together properly. From our perspective, it is fine to just disable this possibility in the build system.

bartlettroscoe:
> @bartlettroscoe @sebrowne We concur that it has probably been a long time since ML and superLU have worked together properly. From our perspective, it is fine to just disable this possibility in the build system.

@rstumin, @jhux2, @csiefer2, the commit that addresses this issue with ML is [`8959a3e` (#11583)](https://github.com/trilinos/Trilinos/pull/11583/commits/8959a3eef96c97a897a5a6e069c9d7d65be5c329).  The Git commit message for that commit should explain this.

bartlettroscoe:
@sebrowne, this PR should be ready to be approved and merged.  Please take a look over similar the fixes in the various packages flagged with the [updated tribits_add_option_and_define()](https://github.com/trilinos/Trilinos/pull/11583/commits/144ffdcebf12dbde52984b1316ccd5c10e8e137d) command.

bartlettroscoe:
> @sebrowne, this PR should be ready to be approved and merged. Please take a look over similar the fixes in the various packages flagged with the [updated tribits_add_option_and_define()](https://github.com/trilinos/Trilinos/pull/11583/commits/144ffdcebf12dbde52984b1316ccd5c10e8e137d) command.

Wait, I just realized there is one issue I need to address before this is merged.

sebrowne:
The approach makes sense to me, thanks for doing the legwork in the individual packages to get this fixed!  I'll approve this whenever you're ready.

sebrowne (approved):

PR #11584, opened 2023-02-21
Ifpack2: block tridi example category

brian-kelley:
Put Ifpack2 BlockTriDiagonalSolver example in both BASIC and PERFORMANCE categories so that it will be built in performance testing builds.

@trilinos/ifpack2 

## Motivation

This executable is needed in performance testing builds to run the SPARC problem.


jhux2 (approved):
lgtm

Issue #11497, opened 2023-01-24
Many: Fix shadowing warnings


sebrowne:
## Bug Report
The following packages have shadowing warnings in them, and as such cannot build with -Werror=shadow.  We currently set these to warnings only (-Wno-error=shadow) on a per-package basis for said packages, but would like to remove this once the warnings get fixed to prevent them from reoccurring and impacting downstream customers who use more aggressive warnings options.

Packages:
Amesos2 @trilinos/amesos2 
Belos @trilinos/belos 
Compadre @trilinos/compadre 
Galeri @trilinos/galeri 
Intrepid2 @trilinos/intrepid2 
NOX @trilinos/nox 
Pamgen @trilinos/pamgen 
Phalanx @trilinos/phalanx 
Piro
ROL @trilinos/rol 
Sacado @trilinos/sacado 
Shards @trilinos/shards 
ShyLU_Node @trilinos/shylu 
STK
Stokhos @trilinos/stokhos 
Stratimikos @trilinos/stratimikos 
TrilinosCouplings @trilinos/trilinoscouplings 
Zoltan @trilinos/zoltan 
 
Only when OpenMP is enabled:
Panzer @trilinos/panzer 
Percept @trilinos/percept 
Pliris @trilinos/pliris 
ShyLU_DD

These warnings can be seen in the GCC builds for PR testing.  See #11397 for more details/examples.

PR #11587, opened 2023-02-21
Framework: Enable shadowing errors in more fixed packages

sebrowne:
Comapdre fixed shadowing warnings, so enable them as errors now to prevent recurrences.

User Support Ticket(s) or Story Referenced: #11497

@trilinos/compadre 

## Motivation
Want to prevent future shadowing warnings.


jwillenbring (approved):
Making progress. Thanks.

PR #11599, opened 2023-02-24
STK: Snapshot 02-23-23 20:19 from Sierra 5.11.7-495-g36224bc1

ldh4:
@trilinos/<teamName>


alanw0 (approved):

packages/stk/stk_ngp_test/stk_ngp_test/GlobalReporter.cpp
```diff
@@ -24,11 +24,17 @@ ReporterBase*& getDeviceReporterOnHost()
...
-  #else
+  ))
+  KOKKOS_IF_ON_HOST((
     static ReporterBase* deviceReporterOnDevice = nullptr;
-  #endif
+  ))
+#else
+  static ReporterBase* deviceReporterOnDevice = nullptr;
+#endif
+
```
dalg24, line 37: Wait did that actually compile?  `KOKKOS_IF_ON_{HOST,DEVICE}` is adding a scope, meaning I expect `deviceReporterOnDevice` is out of scope by the time you reach the return statement.

ldh4: No you are right. It wouldn't compile. I am unsure how it slipped through. I'll patch that in and update this PR.

ldh4:
@bartlettroscoe Looking at build failures from CDash, It looks like the CI test is failing here because TPL_CDT is being enabled:
`-- Setting TPL_ENABLE_CDT=ON because STKMiddle_mesh has a required dependence on CDT`

which then leads to enable this
`-- Setting STKMiddle_mesh_ENABLE_CDT=ON since Trilinos_ENABLE_STKMiddle_mesh=ON AND TPL_ENABLE_CDT=ON`

But I thought TPL_ENABLE_CDT and everything that has a dependency on it should be hard disabled by default, based on your comment from here https://github.com/trilinos/Trilinos/pull/11608#discussion_r1119358768. Did I misunderstand this?

I'm also curious if this is because all STK submodules are explicitly enabled at the beginning of the configure step:
```
loading initial cache file /scratch/trilinos/workspace/Trilinos_PR_cuda-11.4.2-uvm-off/packageEnables.cmake
-- Setting Trilinos_ENABLE_STK = ON
...
-- Setting Trilinos_ENABLE_STKMiddle_mesh = ON
```
If so, is there a way for us to disable `Trilinos_ENABLE_STKMiddle_mesh` even when `Trilinos_ENABLE_STK = ON` if it finds that CDT files are not available?

packages/stk/cmake/Dependencies.cmake
```diff
@@ -11,6 +11,7 @@ SET(SUBPACKAGES_DIRS_CLASSIFICATIONS_OPTREQS
   IO              stk_io               PT  OPTIONAL
   Search          stk_search           PT  OPTIONAL
   Transfer        stk_transfer         PT  OPTIONAL
+  Middle_mesh     stk_middle_mesh      PT  OPTIONAL
```
bartlettroscoe, line 14: This needs to be declared 'EX' instead of 'PT'.  You can't have a non-`EX` package having a required dependency on an `EX` TPL.

bartlettroscoe:
> @bartlettroscoe Looking at build failures on CDash, It looks like builds are failing here because TPL_CDT is being enabled: `-- Setting TPL_ENABLE_CDT=ON because STKMiddle_mesh has a required dependence on CDT`

As described in:

* https://tribitspub.github.io/TriBITS/users_guide/index.html#ex-tpls-given-default-unset-enable-disable-state

the problem is that the subpackage `STKMiddle_mesh` is being declared `PT` instead of `EX` and that triggering the enable of the **required** dependency `CDT`.   The solution to this problem is to declare the subpackage `STKMiddle_mesh` as `EX`.  It makes no sense for a non-`EX` package to have a required dependency on an `EX` TPL. 

NOTE: The issue here is that TriBITS seems to be lacking the check that an `EX` TPL only has downstream `EX` packages that have required dependences on tis and it should generate an error in this case.  I am guessing that there are other such cases that are not caught as well like a non-`EX` package trying to have a required dependency on an upstream `EX` package.  That makes no sense either.  Those should be error conditions.  I can update TriBITS to checks for those cases and error out with a good error message.  (That should be a really easy check to put in.  The only hard part are adding the automated TriBITS tests.)

packages/stk/stk_middle_mesh/cmake/Dependencies.cmake
```diff
@@ -1,8 +1,8 @@
-SET(LIB_REQUIRED_DEP_PACKAGES STKSearch)
+SET(LIB_REQUIRED_DEP_PACKAGES STKUtil STKSearch)
 SET(LIB_OPTIONAL_DEP_PACKAGES)
 SET(TEST_REQUIRED_DEP_PACKAGES Gtest)
-SET(TEST_OPTIONAL_DEP_PACKAGES)
-SET(LIB_REQUIRED_DEP_TPLS)
+SET(TEST_OPTIONAL_DEP_PACKAGES STKMesh STKIO)
+SET(LIB_REQUIRED_DEP_TPLS CDT)
```
bartlettroscoe, line 5: Here is the problem.  If `CDT` really is a required dependency then `STKMiddle_Mesh` needs to be `EX	` as well.

alanw0 (approved):

PR #11608, opened 2023-02-27
Adding in CDT into the TPLsList

ldh4:
Added `CDT` into TPLsList and a corresponding `FindTPLCDT.cmake` file.

## Motivation

CDT is a newly added dependency for a STK module, STK_MiddleMesh.
This change is necessary for the latest STK snapshot: #11599.

### DIFF
--- a/TPLsList.cmake
+++ b/TPLsList.cmake
@@ -174,6 +174,7 @@ TRIBITS_REPOSITORY_DEFINE_TPLS(
   MAGMASparse     "cmake/TPLs/"    EX
   Check           "cmake/TPLs/"    EX
   SARMA           "cmake/TPLs/"    EX
+  CDT             "cmake/TPLs/"    EX
   )
 
 # NOTES:
--- a/cmake/TPLs/FindTPLCDT.cmake
+++ b/cmake/TPLs/FindTPLCDT.cmake
@@ -1,0 +1,59 @@
+# @HEADER
+# ************************************************************************
+#
+#            Trilinos: An Object-Oriented Solver Framework
+#                 Copyright (2001) Sandia Corporation
+#
+#
+# Copyright (2001) Sandia Corporation. Under the terms of Contract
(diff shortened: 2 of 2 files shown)


sebrowne (commented):
This seems fine to me conceptually, but I'll defer to @bartlettroscoe since he knows the mechanics better than I.

bartlettroscoe:
Is CDT really a header-only TPL?

ldh4:
@bartlettroscoe The way we currently have CDT set up in Sierra is to use it as a header-only tpl. That's why we didn't include `REQUIRED_LIBS_NAMES CDT` in `FindTPLCDT.cmake`.

TPLsList.cmake
```diff
@@ -174,6 +174,7 @@ TRIBITS_REPOSITORY_DEFINE_TPLS(
   MAGMASparse     "cmake/TPLs/"    EX
   Check           "cmake/TPLs/"    EX
   SARMA           "cmake/TPLs/"    EX
+  CDT             "cmake/TPLs/"    EX
```
bartlettroscoe, line 177: @ldh4, will this be required or optional TPL for STK?  With the TPL currently being declared `EX` it will be automatically hard disabled and will take out anything downstream that has a required dependency on it.

ldh4: This can be an optional TPL. It is only required for one of STK modules, and we are okay having that module disabled if the user doesn't manually enable CDT and provide its installation path.

bartlettroscoe (approved):
Okay, given that it is okay to have the CDT TPL hard disabled by default and have that take out the required downstream STK subpackage, this looks correct to me.

alanw0 (approved):
I concur. This is an optional dependency for stk, and one stk sub-module gets disabled if this CDT library is not present.
If Ross thinks these changes look good, I definitely agree, let's do it! :-)

PR #11602, opened 2023-02-25
Panzer MiniEM: Add Darcy problem

cgcgcg:
@trilinos/panzer 

## Motivation
Adds assembly and solution of a Darcy problem to MiniEM. This makes sense, as lots of setup and solver code can be reused. We can think about renaming MiniEM later.

PR #11484, opened 2023-01-19
teuchos(fix): add missing move constructors for Teuchos::any and Teuchos::ParameterEntry

romintomasetti:
@trilinos/Teuchos

## Motivation

Adding missing move semantics, see https://github.com/trilinos/Trilinos/issues/11481.

## Related Issues

* https://github.com/trilinos/Trilinos/issues/11481

## Testing

Tested for `OpenMP`, `Cuda 12`, and `ROCm 5.4`.

PR #11610, opened 2023-02-27
Xpetra,MueLu,ShyLU: Drop kokkos refactor

cgcgcg:

## Motivation
A while back, Xpetra and MueLu changed Kokkos to a required dependency. 
This PR removes `HAVE_XPETRA_KOKKOS_REFACTOR` and `HAVE_MUELU_KOKKOS_REFACTOR`.


GrahamBenHarper:
Does this still require us to drop the non-kokkos factories or merge them with their kokkos counterparts?

cgcgcg:
@GrahamBenHarper I wouldn't call it a requirement, but it would be good to do that at some point. Always building the Kokkos factories is one step in that direction.

GrahamBenHarper:
Right. I suppose this means certain factories wouldn't be used anymore once it merges?

GrahamBenHarper:
Thanks! I follow what this does now.

jhux2:
> No, it means that both non-Kokkos and Kokkos factories will always be compiled. Before you had to build with "MueLu_ENABLE_Kokkos_Refactor" to get that. At runtime, you can still select to either category of factories.

@cgcgcg Are you sure about that?  After Kokkos became a required dependency, I thought the Kokkos factories are now always compiled, i.e., `HAVE_MUELU_KOKKOS_REFACTOR` is true.

Whether the Kokkos factories are enabled at runtime is a different matter.  Serial node defaults to the non-Kokkos factories, but the Kokkos factory chain can be enabled by setting `use kokkos refactor` to `true`.

jhux2 (approved):
@cgcgcg wins the negative-lines-of-code PR of the week!

cgcgcg:
@trilinos/framework @sebrowne I'm not sure why this PR and #11602 aren't merging. Do they need to be retested after being switched from draft mode?

sebrowne:
I think the autotester just hasn't gotten back to them believe it or not.  It's currently testing #11484, so once it's done with that it will continue advancing through the even-numbered PRs until it gets to this one and (hopefully) merges it.

cgcgcg:
Sounds good. Thanks for the quick reply!

PR #11612, opened 2023-02-27
Xpetra: Set HAVE_XPETRA_EPETRAEXT=OFF when no Epetra support

bartlettroscoe:
CC: @cgcgcg 

## Description
    
This fixes an Albany Trilinos build configuration issue that resulted in a broken Trilinos build.  When you overide an optional package enable like Xpetra_ENABLE_EpetraExt=OFF that was ON, you also need to set the related macro name var HAVE_XPETRA_EPETRAEXT=OFF (which TriBITS automatically sets by default before processing the packages) which is used in the generation of the Package's <Package>_Config.h file.

I was able to reproduce the reported build error reported by @cgcgcg before this change and verified that it fixed the build of Xpetra after this change.


cgcgcg (approved):
Sweet! Thank you very much for looking into this!

bartlettroscoe:
@cgcgcg, could you please look over the updated TriBITS documentation in:

* https://github.com/TriBITSPub/TriBITS/pull/566

and see if it explains the issues resolved in this PR (#11612) okay?

PR #11618, opened 2023-02-28
Framework: Change CDash settings for nightly jobs

sebrowne:
I want nightly CDash jobs to work the way that CDash expects.  So, make the build name for nightly jobs be the same, and 
set the model to nightly.  For this, "nightly jobs" means "any job not in the track 'Pull Request'".

Long-term we probably do not want to use the PR framework for nightly builds (doesn't make a whole lot of sense), but since we do at the moment, this change helps us move towards more sensible output.

@trilinos/framework 

## Motivation
This will enable the historical graphs and other analysis of night-to-night runs on CDash.


prwolfe (approved):

jmlapre (approved):
Everything looks good.

PR #11620, opened 2023-03-01
add OpenMP to a recent clang build using the llvm OpenMP runtime (not GCC)

jmlapre:
At a testbed meeting, it was suggested that we add OpenMP to a recent clang build using the llvm OpenMP runtime (not GCC). This will help prepare Trilinos for future architectures.

The SEMS clang supports OpenMP, so you just need to enable OpenMP (how do you do that in general: Trilinos_ENABLE_OPENMP or Kokkos_ENABLE_OPENMP?)

@trilinos/framework 

https://sems-atlassian-son.sandia.gov/jira/browse/TRILFRAME-1

### DIFF
--- a/packages/framework/ini-files/config-specs.ini
+++ b/packages/framework/ini-files/config-specs.ini
@@ -1788,6 +1788,22 @@ use RHEL7_TEST_DISABLES|CLANG
 
 use RHEL7_POST
 
+[rhel7_sems-clang-11.0.1-openmpi-1.10.1-openmp_debug_static_no-kokkos-arch_asan_no-complex_fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables]
+use rhel7_sems-clang-11.0.1-openmpi-1.10.1-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables
+
+opt-set-cmake-var Ifpack2_LocalSparseTriangularSolver_DISABLE BOOL FORCE : ON
+opt-set-cmake-var Ifpack2_LocalSparseTriangularSolver2_DISABLE BOOL FORCE : ON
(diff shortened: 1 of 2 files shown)


packages/framework/ini-files/config-specs.ini
```diff
@@ -1788,6 +1788,22 @@ use RHEL7_TEST_DISABLES|CLANG
 
 use RHEL7_POST
 
+[rhel7_sems-clang-11.0.1-openmpi-1.10.1-openmp_debug_static_no-kokkos-arch_asan_no-complex_fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables]
+use rhel7_sems-clang-11.0.1-openmpi-1.10.1-serial_release-debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables
+
+opt-set-cmake-var Ifpack2_LocalSparseTriangularSolver_DISABLE BOOL FORCE : ON
```
sebrowne, line 1794: What's the reason behind these test disables?

jmlapre: The compilation (not tests) failed with the specified configuration.

sebrowne: Just the compilation of those respective tests?

jmlapre: One or more files required by those specific tests failed to compile.

sebrowne: Hmmm I'm torn.  If it doesn't cause catastrophic compile failures in other packages, I think it would be better to leave them enabled.  Meaning, we'll still have the rest of the tests run.  I'm fine with it either way though (I'm thinking along the line of "this way hides that there is a problem with the tests").

sebrowne (approved):

sebrowne (approved):

PR #11626, opened 2023-03-03
MueLu: Fix fscanf and system warnings

GrahamBenHarper:
@trilinos/muelu

I saw some warnings about `fscanf` unused returns and `system` unused returns originating from `MatrixLoad.hpp` and `AggExport.cpp` like the one below. I fixed them :)
```
/home/graham/Programming/cpp/Trilinos/Trilinos-source/packages/muelu/test/scaling/MatrixLoad.hpp:295:19: warning: ignoring return value of ‘int fscanf(FILE*, const char*, ...)’, declared with attribute warn_unused_result [-Wunused-result]
  295 |             fscanf(fp,"%d %d %d\n",&nBlks, &nRows, &nnzs);
      |             ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Note: While I placed asserts at the start of the file parsing in `MatrixLoad.hpp`, I did not continue by placing asserts inside the main loops. In principle, they should be inside the loop, but I would rather not sacrifice performance on loading user input, although I have no clue how much of an impact it would have on performance anyway. Feel free to discuss if you think it's worth adding an assert in the main loop as well.

### DIFF
--- a/packages/muelu/test/scaling/MatrixLoad.hpp
+++ b/packages/muelu/test/scaling/MatrixLoad.hpp
@@ -296,6 +296,7 @@ void readUserBlks(const std::string& userBlkFileName, const std::string& smoothe
           if (mueluList.sublist(smootherOrCoarse + ": params").sublist("subdomain solver parameters").get<std::string>("partitioner: type") == "user") {
 
             FILE   *fp;
+            int    retVal;
             int    nBlks, nRows, nnzs, ch, row, col;
             int    procId = comm->getRank();
             double val;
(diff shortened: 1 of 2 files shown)


packages/muelu/test/scaling/MatrixLoad.hpp
```diff
@@ -296,6 +296,7 @@ void readUserBlks(const std::string& userBlkFileName, const std::string& smoothe
           if (mueluList.sublist(smootherOrCoarse + ": params").sublist("subdomain solver parameters").get<std::string>("partitioner: type") == "user") {
 
             FILE   *fp;
+	    int    retVal;
```
cgcgcg, line 299: not sure if github is playing tricks on me here: indentation?

GrahamBenHarper: my emacs config might be messed up on this machine. how do I keep the default emacs from inserting tabs?

cgcgcg: I have `(setq-default indent-tabs-mode nil)` in my config.

GrahamBenHarper: okay it should be fixed now. thanks for that :)

GrahamBenHarper:
My first PR from 38000ft. Let's see if it testing passes while I'm still airborne

cgcgcg (approved):

PR #11629, opened 2023-03-04
MueLu: Suppress more system and mkstemp warnings, update affiliations

GrahamBenHarper:
@trilinos/muelu

I suppressed a couple more warnings related to `system` and `mkstemp` calls. I also updated Matthias' affiliation in the `DoxygenDocumentation` to reflect his correct affiliation.

The snow finally subsided in Detroit, but I'm still waiting to hopefully get on a plane soon. :)

### DIFF
--- a/packages/muelu/test/unit_tests/Hierarchy.cpp
+++ b/packages/muelu/test/unit_tests/Hierarchy.cpp
@@ -1592,11 +1592,13 @@ namespace MueLuTests {
     // Write matrices out, read fine A back in, and check that the read was ok
     // by using a matvec with a random vector.
     char t[] = "XXXXXX";
+    int filedescriptor = 0;
     if (comm->getRank() == 0)
-      mkstemp(t); //mkstemp() creates a temporary file. We use the name of that file as
-                  //the suffix for the various data files produced by Hierarchy::Write().
-                  //A better solution would be to write to a file stream, but this would
-                  //involve writing new interfaces to Epetra's file I/O capabilities.
+      filedescriptor = mkstemp(t); //mkstemp() creates a temporary file. We use the name of that file as
+                       //the suffix for the various data files produced by Hierarchy::Write().
+                       //A better solution would be to write to a file stream, but this would
+                       //involve writing new interfaces to Epetra's file I/O capabilities.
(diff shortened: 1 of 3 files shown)


packages/muelu/test/unit_tests/Hierarchy.cpp
```diff
@@ -1592,11 +1592,13 @@ namespace MueLuTests {
     // Write matrices out, read fine A back in, and check that the read was ok
     // by using a matvec with a random vector.
     char t[] = "XXXXXX";
+    int fd;
```
mayrmt, line 1595: You could rename to `filedescriptor`.... a bit more descriptive

packages/muelu/test/unit_tests/Hierarchy.cpp
```diff
@@ -1592,11 +1592,13 @@ namespace MueLuTests {
...
     if (comm->getRank() == 0)
-      mkstemp(t); //mkstemp() creates a temporary file. We use the name of that file as
-                  //the suffix for the various data files produced by Hierarchy::Write().
-                  //A better solution would be to write to a file stream, but this would
-                  //involve writing new interfaces to Epetra's file I/O capabilities.
+      fd = mkstemp(t); //mkstemp() creates a temporary file. We use the name of that file as
+                       //the suffix for the various data files produced by Hierarchy::Write().
+                       //A better solution would be to write to a file stream, but this would
+                       //involve writing new interfaces to Epetra's file I/O capabilities.
+    TEST_INEQUALITY(fd, -1); // if the file descriptor is -1, it failed
```
mayrmt, line 1601: `fd` is uninitialized on procs other than rank 0. Might this affect this test? Maybe initialize `fd` above to be on the safe side...

packages/muelu/test/unit_tests/Hierarchy.cpp
GrahamBenHarper: done!

packages/muelu/test/unit_tests/Hierarchy.cpp
GrahamBenHarper: Good point. It might introduce random unspecified behavior, which is the last thing we need in the testsuite... especially since the only failed test from the AT was a random failure

mayrmt (approved):

PR #11639, opened 2023-03-07
MueLu: Add option to build Graham's matrix-free

GrahamBenHarper:
@trilinos/muelu @pohm01 

This enables building of the matrix-free example in Graham's research folder when the following configure conditions are enabled:
```
Trilinos_ENABLE_Belos=ON \
MueLu_ENABLE_Experimental=ON \
MueLu_ENABLE_EXAMPLES=ON \
```

This also includes some minor tweaks to get it compiling and running without error. It'll need more tweaks later.


cgcgcg:
Why do we need a new option for that? Isn't that what `MueLu_ENABLE_Experimental` is for?

GrahamBenHarper:
@cgcgcg That's a good point. I guess in my mind I didn't want to try to support every possible configuration we use in experimental, but I can try just building it under experimental and see if the nightlies survive. It's also not totally clear to me that Belos is the only package I need to check for, but we'll see.

cgcgcg (approved):

PR #11644, opened 2023-03-09
Update versions in anticipation of 14.0 release

sebrowne:
@trilinos/framework 

## Motivation
Trilinos 14.0 is ready for release.  Need the versions to be ready for the tag.

PR #11645, opened 2023-03-09
Framework: Update versions to reflect 14.0 release

sebrowne:
@trilinos/framework 

## Motivation
Once Trilinos 14.0 is released, want develop/master to reflect that 14.0 was the most recent release.

## Related Issues
Wait for #11644 to be merged first.

### DIFF
--- a/Version.cmake
+++ b/Version.cmake
@@ -59,10 +59,10 @@
 # for release mode and set the version.
 #
 
-SET(Trilinos_VERSION 13.5)
-SET(Trilinos_MAJOR_VERSION 13)
-SET(Trilinos_MAJOR_MINOR_VERSION 130500)
-SET(Trilinos_VERSION_STRING "13.5 (Dev)")
+SET(Trilinos_VERSION 14.1)
+SET(Trilinos_MAJOR_VERSION 14)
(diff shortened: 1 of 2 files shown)


ccober6 (approved):
Looks reasonable.

ndellingwood (approved):
lgtm, thanks!

Version.cmake
```diff
@@ -59,10 +59,10 @@
 # for release mode and set the version.
 #
 
-SET(Trilinos_VERSION 13.5)
-SET(Trilinos_MAJOR_VERSION 13)
-SET(Trilinos_MAJOR_MINOR_VERSION 130500)
-SET(Trilinos_VERSION_STRING "13.5 (Dev)")
+SET(Trilinos_VERSION 14.1)
```
lucbv, line 62: This looks like a bad idea, what will be in develop after 14.0 is not really what will end up in 14.1
I would suggest labeling develop as 140099, the latest (and unlikely) patch number before 14.1

jwillenbring: See for example the recent list of release versions: https://github.com/trilinos/Trilinos/releases

PR #11651, opened 2023-03-09
MueLu: Remove CoarseMapFactory_kokkos

cgcgcg:
@trilinos/muelu 

## Motivation
The only difference (as far as I can tell) between `CoarseMapFactory` and `CoarseMapFactory_kokkos` is the dependence on `Aggregates` vs `Aggregates_kokkos`. Until we get rid of the former, we can use a single factory to handle both cases.

Also, I fixed the scope of the coloring timer in `UncoupledAggregationFactory_kokkos`.


GrahamBenHarper (dismissed):
Nice! Thanks for doing this.

GrahamBenHarper:
related to #11610

GrahamBenHarper (approved):

PR #11654, opened 2023-03-10
Tpetra sycl updates

lucbv:
@trilinos/tpetra 

## Motivation
Making some fixes to our code for proper runs with SYCL backend

## Related Issues

* Closes 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Stakeholder Feedback
ECP

## Testing
Tested on ANL machines with oneAPI compilers.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
@@ -224,11 +224,11 @@ pad_crs_arrays(
     auto row_ptr_end_h = create_mirror_view(
       hostSpace, row_ptr_end, verbose, prefix.get());
     // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
-    Kokkos::deep_copy(execution_space(), row_ptr_end_h, row_ptr_end);
+    Kokkos::deep_copy(exec_space_instance, row_ptr_end_h, row_ptr_end);
     auto row_ptr_beg_h = create_mirror_view(
       hostSpace, row_ptr_beg, verbose, prefix.get());
     // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
-    Kokkos::deep_copy(execution_space(), row_ptr_beg_h, row_ptr_beg);
+    Kokkos::deep_copy(exec_space_instance, row_ptr_beg_h, row_ptr_beg);
 
     // lbv 03/15/23: The execution space deep_copy does an asynchronous
     // copy so we really want to fence that space before touching the
--- a/packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
@@ -220,6 +220,7 @@ pad_crs_arrays(
   size_t increase = 0;
   {
     // Must do on host because padding uses std::map
+    execution_space exec_space_instance = execution_space();
     auto row_ptr_end_h = create_mirror_view(
       hostSpace, row_ptr_end, verbose, prefix.get());
     // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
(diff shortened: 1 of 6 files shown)


jennloe:
Per conversation with @csiefer2, would it be prudent to wrap these fences in some sort of "If SYCL enabled" construct so they can't add overhead for other backends?

packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
```diff
@@ -2564,6 +2564,13 @@ namespace Tpetra {
           KokkosBlas::sum (subview (lclSums, j), subview (X_lcl, ALL (), col));
         }
       }
+      // lbv 10 mar 2023: Kokkos Kernels provides non-blocking BLAS
+      // functions unless they explicitely return a value to Host.
+      // Here while the lclSums are on host, they are not a return
+      // value, therefore they might be avaible to us immediately.
+      // Adding a frnce here guarantees that we will have the lclSums
+      // ahead of the MPI reduction.
```
masterleinad, line 2572:
```suggestion
      // lbv 10 mar 2023: Kokkos Kernels provides non-blocking BLAS
      // functions unless they explicitly return a value to Host.
      // Here while the lclSums are on the host, they are not a return
      // value, therefore they might be available to us immediately.
      // Adding a fence here guarantees that we will have the lclSums
      // ahead of the MPI reduction.
```

masterleinad (commented):
@jennloe To me, it doesn't look like any of the fences are specific to the `SYCL` backend but rather bug fixes affecting correctness for all backends.

lucbv:
@jennloe these fences are necessary for all backends, we just happen to be lucky and not to see errors with CUDA.
I am not in favor of guarding them as this can lead us to chase bugs down the road that we can already fix.

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -229,6 +230,14 @@ pad_crs_arrays(
     // DEEP_COPY REVIEW - DEVICE-TO-HOSTMIRROR
     Kokkos::deep_copy(execution_space(), row_ptr_beg_h, row_ptr_beg);
```
crtrott, line 231: Us exec_space_instance here and the deep_copy above.

packages/tpetra/core/src/Tpetra_Details_crsUtils.hpp
```diff
@@ -229,6 +230,14 @@ pad_crs_arrays(
...
 
+    // lbv 03/15/23: The execution space deep_copy does an asynchronous
+    // copy so we really want to fence that space before touching the
+    // host data as it is not guarenteed to have arrived by the time we
+    // start the parallel_reduce below which might use a different
+    // execution space, see:
+    // https://kokkos.github.io/kokkos-core-wiki/API/core/view/deep_copy.html#semantics
+    exec_space_instance.fence();
```
crtrott, line 239: Note the reason why it may or may not cause issues on other backends is that on some backends allocations are blocking. But that is not guaranteed, so this fence needs to be here.

packages/tpetra/core/src/Tpetra_Details_normImpl.hpp
```diff
@@ -333,6 +334,15 @@ normImpl (MagnitudeType norms[],
...
   Impl::lclNormImpl (normsOut, X, numVecs, whichVecs,
                      isConstantStride, whichNorm);
+
+  // lbv 03/15/23: the data from the local norm calculation
+  // better really be available before communication happens
+  // so fencing to make sure the local computations have
+  // completed on device. We might want to make this an
+  // execution space fence down the road?
+  execution_space exec_space_instance = execution_space();
+  exec_space_instance.fence();
```
crtrott, line 344: Note you better be sure that the default instance is what is used in lclNormImpl. Also presumably that call is only asynchronous because it calls a parallel reduce with a Kokkos::View as output argument instead of a scalar?
parallel_reduce with a scalar output argument is blocking, with View output arguments its not.

crtrott (commented):
Use the instance above but otherwise these look correct unconditionally.

masterleinad:
Any chance that we can already get this merged and fix more issues in a later pull request?

lucbv:
Sure, I will rebase and remove the AT:WIP, that should be fine

lucbv:
@csiefer2 this could use a review

sebrowne:
@lucbv if you include this change: https://github.com/trilinos/Trilinos/compare/develop...sebrowne:Trilinos:develop

It should disable that ROL test for the specific problematic toolchains.

GrahamBenHarper:
I put retest back on since the problematic test should be turned off now

Issue #11655, opened 2023-03-10
Tpetra&KokkosKernels: Debug mode checks on SpGEMM fail


cgcgcg:
@trilinos/kokkos-kernels @trilinos/tpetra 

## Bug Report
In debug builds Kokkos Kernels recently added checks that the input matrices are sorted. Some TPLs require that, but currently Tpetra SpGEMM cannot use them.

When computing `A*B` on multiple ranks, Tpetra imports B into the colMap of A. The resulting Bmerged is not necessarily sorted and hence triggers the assert in Kokkos Kernels.

One such example is the test `PanzerMiniEM_MiniEM-BlockPrec_RefMaxwell_reuse_MPI_4`. I verified that in this case both A and B are sorted, but Bmerged is not.

jhux2:
> Some TPLs require that, but currently Tpetra SpGEMM cannot use them.

@cgcgcg By this, you mean that Tpetra's SpGEMM is unable to take advantage of sorted indices?

cgcgcg:
It seems that whenever Tpetra does an SpGEMM, no matter what TPLs are enabled, we end up with the KokkosKernels implementation (due to wrong type for offsets). Kokkos Kernels does not require sorted indices, only some TPLs do.

brian-kelley:
@jhux2 It's because Tpetra CrsMatrix uses offset = size_t for the local matrix type, but all 3 TPLs only work when offset = int (offset being the rowptrs element type).

I'm thinking of adding the code in Tpetra MatrixMatrix to convert the offsets to int explicitly so that the TPLs can be called (just for performance).

jhux2:
> It seems that whenever Tpetra does an SpGEMM, no matter what TPLs are enabled, we end up with the KokkosKernels implementation (due to wrong type for offsets).

> It's because Tpetra CrsMatrix uses offset = size_t for the local matrix type, but all 3 TPLs only work when offset = int (offset being the rowptrs element type).

@cgcgcg @brian-kelley Gotcha, makes sense.

cgcgcg:
@rppawlo Can you confirm that #11663 fixed this for EMPIRE?

cgcgcg:
@brian-kelley I guess we can leave the issue open to track that something should still happen on the Tpetra side.

PR #11663, opened 2023-03-11
KokkosKernels: temp. fix for #11655

brian-kelley:
(Temporary) In SpGEMM, disable the check that the input matrices are sorted if the KokkosKernels (non-TPL) implementation is going to be called. This impl does not actually require sorted inputs. Tpetra's MatrixMatrix::Multiply logic needs some nontrivial changes to have a sorted B, but that will be the long-term fix for #11655.

@trilinos/kokkos-kernels 
@trilinos/tpetra

## Motivation

With this change, all Panzer-MiniEM tests pass in an OpenMP+Debug build (develop branch had 6 failures).

### DIFF
--- a/packages/kokkos-kernels/sparse/src/KokkosSparse_spgemm_symbolic.hpp
+++ b/packages/kokkos-kernels/sparse/src/KokkosSparse_spgemm_symbolic.hpp
@@ -140,15 +140,26 @@ void spgemm_symbolic(KernelHandle *handle,
   // Verify that graphs A and B are sorted.
   // This test is designed to be as efficient as possible, but still skip
   // it in a release build.
+  //
+  // Temporary fix for Trilinos issue #11655: Only perform this check if a TPL
+  // is to be called. The KokkosKernels (non-TPL) implementation does not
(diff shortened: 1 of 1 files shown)


cgcgcg (approved):
Thanks @brian-kelley !

rppawlo (approved):

rppawlo:
Thanks @brian-kelley and @cgcgcg !

brian-kelley:
@ndellingwood Hmm... I was hoping this could just be a short-term change that could be reverted soon. It's really a workaround for a Tpetra issue that evaded testing, rather than a KokkosKernels issue.

When are you planning to put together the 4.0.1 release candidate? Depending on that I will either patch to KK, or go for the permanent fix in Tpetra and (at the same time) revert this PR.

ndellingwood:
> I was hoping this could just be a short-term change that could be reverted soon

@brian-kelley thanks for the update, I hadn't realized that. No ETA on a 4.0.01 patch yet, I'll check in the kokkos meeting this week. I'll check back in with you closer to snapshot time with 4.0.01, if I don't see a PR to Tpetra with the long-term fix we can coordinate the patch to eventual release candidate branch

Issue #11658, opened 2023-03-10
MueLu: Merge *_kokkos factories and their classical counterparts


cgcgcg:
@trilinos/muelu 

## Enhancement
We duplicated a lot of algorithms and code paths for use without and with Kokkos. Since MueLu effectively requires Kokkos now, we can start to remove these code duplications. Some removals are quite safe, but others will require a lot of work since the two code branches have diverged significantly.

- [x] Remove CMake variables `Xpetra_ENABLE_KOKKOS_REFACTOR` and `MueLu_ENABLE_KOKKOS_REFACTOR` #11610 
- [ ] Merge `UtilitiesKokkos` into `UtilitiesBase` and `Utilities` #11641 
 - [ ] Aggregates_kokkos    
 - [ ] AggregationStructuredAlgorithm_kokkos
 - [ ] AmalgamationFactory_kokkos
 - [ ] AmalgamationInfo_kokkos
 - [ ] CoalesceDropFactory_kokkos (nightmare?)
 - [x] CoarseMapFactory_kokkos #11651 
 - [ ] CoordinatesTransferFactory_kokkos (might be simple)
 - [ ] GeometricInterpolationPFactory_kokkos
 - [ ] IndexManager_kokkos
 - [ ] LWGraph_kokkos
 - [ ] LocalLWGraph_kokkos
 - [ ] MatrixFreeTentativePFactory_kokkos (rename)
 - [ ] MatrixFreeTentativeP_kokkos (rename)
 - [ ] MueLu_AggregationPhase1Algorithm_kokkos
 - [ ] MueLu_AggregationPhase2aAlgorithm_kokkos
 - [ ] MueLu_AggregationPhase2bAlgorithm_kokkos
 - [ ] MueLu_AggregationPhase3Algorithm_kokkos
 - [ ] MueLu_IsolatedNodeAggregationAlgorithm_kokkos
 - [ ] MueLu_OnePtAggregationAlgorithm_kokkos
 - [ ] MueLu_PreserveDirichletAggregationAlgorithm_kokkos
 - [ ] MueLu_UncoupledAggregationFactory_kokkos
 - [ ] NullspaceFactory_kokkos (only difficulty: move rotations to Kokkos)
 - [ ] RegionRFactory_kokkos
 - [ ] SaPFactory_kokkos
 - [ ] SemiCoarsenPFactory_kokkos
 - [ ] StructuredAggregationFactory_kokkos
 - [ ] TentativePFactory_kokkos
 
[...]

Issue #11660, opened 2023-03-11
MueLu: Build errors in matrix-free example


cgcgcg:
## Bug Report
```
/home/caglusa/Trilinos/packages/muelu/research/graham/mf_example_01.cpp:627:7: error: 'sleep' was not declared in this scope
       sleep(my_rank);
       ^~~~~
```

Also:

https://testing.sandia.gov/cdash/viewBuildError.php?buildid=11498666

GrahamBenHarper:
Hah, that's a really random error. I guess I need `<unistd.h>` included? I can also fix the inconsistent Teuchos::Comm vs Teuchos::MpiComm usage.

GrahamBenHarper:
Okay I'm having trouble reproducing this somehow. It even builds on Fugaku without issue, which is ironic since building on Fugaku has been the hardest for us. I'll see if I can reproduce the environment/config so I can fix all these at once instead of doing repeated incremental fixes.

cgcgcg:
No more build errors!

cgcgcg:
Spoke too soon. There are a bunch of hard coded template args that cause link errors.

GrahamBenHarper:
oh yeah, I assume things like hardcoded uses of double? link to the failures?

GrahamBenHarper:
Ohhh I see now. It's probably from
```
    using SC = double;
    using LO = int;
    using GO = long long;
    using NO = Tpetra::MultiVector<>::node_type;
```
With the testing framework in there, what do you think is an appropriate fixup for that?

cgcgcg:
Just use the template args:
https://github.com/trilinos/Trilinos/blob/07a951062df54e4d4a095c1358fcf1102d9243cc/packages/muelu/research/graham/mf_example_01.cpp#L441

GrahamBenHarper:
How about now? :)

cgcgcg:
Seems to build.

PR #11664, opened 2023-03-11
MueLu: Rename MatrixFree*_kokkos objects and files

GrahamBenHarper:
@trilinos/muelu @cgcgcg 

Renames the matrix-free objects away from the old _kokkos format. Part of #11658.


GrahamBenHarper:
FYI, git is going to claim I renamed files incorrectly and therefore replaced the lines inside them incorrectly. I did it correctly, but git is having a meltdown.
image

GrahamBenHarper:
Lumping in a fix for #11660

cgcgcg (approved):

Issue #11674, opened 2023-03-14
Teuchos:  Missing Includes for GCC13


hartsw:
## Bug Report
Missing include in teuchos and kokkos  @bartlettroscoe

### Description

The Fedora 38 beta release came out today, which includes GCC 13 (also prerelease).  To get my systems ready to move to this version I've started to test against it.   Trilinos fails to compile due to missing includes.

I generally use `spack` to build Trilinos, but I checked out the tag 13-4-0 to build (13-4-1 also exhibits the problem).

### Steps to Reproduce
1. SHA1:  13-4-0
1. Configure script:  `cmake -DTrilinos_ENABLE_ALL_PACKAGES=YES -DTPL_ENABLE_Netcdf=OFF -DBUILD_SHARED_LIBS=YES ..`
1. Configure log:  [attach here]
1. Build log:  [attach here]
1. Input deck:  N/A

Missing includes are for `std::uint32_t` and others, remedied by including `<cstdint>` in a couple files.

Git Diff:
```
diff --git a/packages/kokkos/core/src/impl/Kokkos_MemoryPool.cpp b/packages/kokkos/core/src/impl/Kokkos_MemoryPool.cpp
index 889d821bb1c..3ddc1d3515e 100644
--- a/packages/kokkos/core/src/impl/Kokkos_MemoryPool.cpp
+++ b/packages/kokkos/core/src/impl/Kokkos_MemoryPool.cpp
@@ -46,6 +46,7 @@
 
 #include <ostream>
 #include <sstream>
+#include <cstdint>
 
 //----------------------------------------------------------------------------
 //----------------------------------------------------------------------------
diff --git a/packages/teuchos/core/src/Teuchos_BigUIntDecl.hpp b/packages/teuchos/core/src/Teuchos_BigUIntDecl.hpp
index e82e8be9e9a..b41b0d035af 100644
--- a/packages/teuchos/core/src/Teuchos_BigUIntDecl.hpp
+++ b/packages/teuchos/core/src/Teuchos_BigUIntDecl.hpp
@@ -43,6 +43,7 @@
 #define TEUCHOS_BIG_UINT_DECL_HPP
 
 #include <iosfwd>
+#include <cstdint>
 
 /*! \file Teuchos_BigUIntDecl.hpp
     \brief Arbitrary-precision unsigned integer declaration.
```

bartlettroscoe:
CC: @sebrowne 

[...]

PR #11676, opened 2023-03-14
teuchos: Add missing <cstdint> include in teuchos for GCC 13

hartsw:
While testing Fedora 38 (beta) it was determined that teuchos was missing an include for `<cstdint>`.  This was added where required.

@trilinos/teuchos
## Motivation

While testing building Trilinos 13-0-1 with GCC 13 it was determined that there were missing include files.  The following adds the includes where needed for `develop` but probably doesn't fix the older versions.

Related to issue #11674


hartsw:
This change is fairly trivial.  I tested also made a patch for `13-4-0` and `13-4-1` which required another file to have the same header included, but it looks like that file has been removed in the meantime.

bartlettroscoe (approved):
Should be fine if it passes all of the PR builds.

Issue #10569, opened 2022-05-25
Tpetra: tests failing in OpenMP+Cuda and OpenMP+Rocm builds


romintomasetti:
## Bug Report
@trilinos/tpetra

### Description

When compiling `Trilinos` (*develop branch*) with `OpenMP`+`Cuda` or with `OpenMP`+`Rocm` (`HIP`), we find that *compilation succeeds*, but *some Tpetra tests fail*. 

Specifically, compiling `Trilinos` with:
* `Tpetra`
* `OpenMP`
* `ROCm` (4.5)

works.

We run our tests with
```json
"MPI_EXEC_POST_NUMPROCS_FLAGS"    : {"type" : "STRING", "value" : "--map-by;socket:PE=1;--bind-to;core"},
```

When we set `OMP_NUM_THREADS` to a value **larger than 1**, some `Tpetra` tests with more than one MPI process fail:
```cmake
The following tests FAILED:
25 - TpetraCore_BlockCrsMatrix_MPI_4 (Failed)
55 - TpetraCore_CrsGraph_UnitTests0_MPI_4 (Failed)
56 - TpetraCore_CrsGraph_UnitTests1_MPI_4 (Failed)
57 - TpetraCore_CrsGraph_UnitTests_Swap_MPI_2 (Failed)
58 - TpetraCore_CrsGraph_ReindexColumns_MPI_4 (Failed)
62 - TpetraCore_CrsGraph_getNumDiags_MPI_4 (Failed)
65 - TpetraCore_CrsGraph_StaticImportExport_MPI_4 (Failed)
68 - TpetraCore_CrsGraphTransposer_UnitTests_MPI_4 (Failed)
113 - TpetraCore_Distributor_CreateFromSendsAndRecvs_MPI_4 (Failed)
116 - TpetraCore_FECrsGraph_UnitTests_MPI_4 (Failed)
117 - TpetraCore_FECrsMatrix_UnitTests_MPI_4 (Failed)
124 - TpetraCore_ImportExport_ImportConstructExpert_MPI_4 (Failed)
129 - TpetraCore_UnpackLongRows_MPI_4 (Failed)
138 - TpetraCore_ImportExport2_UnitTests_ISend_MPI_4 (Failed)
195 - TpetraCore_MatrixMatrix_UnitTests_MPI_4 (Failed)
196 - TpetraCore_FECrs_MatrixMatrix_UnitTests_MPI_4 (Failed)
[...]

PR #11679, opened 2023-03-15
Tpetra: fix tests with both Cuda & OpenMP

brian-kelley:
Fix all Tpetra tests in a build with both Cuda and OpenMP nodes enabled.

Most issues were caused by non-thread-safe WrappedDualView functions being used in host-parallel kernels. Fixed this by disabling the WrappedDualView "sync path" parallel kernels. As a side effect this reduces a bit of overhead by skipping the reference-count checks.

In SpGEMM-Jacobi, the issues were just that the Kokkos default device was being used for temporary views, so it was trying to use e.g. CudaSpace views inside an OpenMP kernel. Fixed by using the Node's device type for those views.

@trilinos/tpetra 

## Motivation

Fixes some actual bugs - non-thread-safe WrappedDualView functions were being used inside parallel kernels.

## Stakeholder Feedback

See #10569 - getting this configuration to work was requested by @romintomasetti 

## Testing

Yes, all changes are tested (the original issue was because Tpetra tests were failing in this configuration).

### DIFF
--- a/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
@@ -2609,6 +2609,9 @@ public:
         const auto policy =
           policy_type(numExportLIDs, 1, 1)
           .set_scratch_size(0, Kokkos::PerTeam(sizeof(GO)*maxRowLength));
+        // The following parallel_for needs const access to the local values of src.
+        // (the local graph is also accessed on host, but this does not use WDVs).
+        getValuesHost();
         Kokkos::parallel_for
           (policy,
            [=](const typename policy_type::member_type &member) {
@@ -2877,6 +2880,9 @@ public:
       using host_scratch_space = typename host_exec::scratch_memory_space;
       
       using pair_type = Kokkos::pair<size_t, size_t>;
+
+      //The following parallel_for modifies values on host while unpacking.
+      getValuesHostNonConst();
       Kokkos::parallel_for
         ("Tpetra::BlockCrsMatrix::unpackAndCombine: unpack", policy,
          [=] (const typename policy_type::member_type& member) {
--- a/packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp
+++ b/packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp
@@ -119,7 +119,8 @@ sync_device(DualViewType dualView) { }
 
 }
 
-
+/// \brief A wrapper around Kokkos::DualView to safely manage data
+///        that might be replicated between host and device.
 template <typename DualViewType>
 class WrappedDualView {
 public:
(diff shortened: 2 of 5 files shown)


csiefer2 (approved):

packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp
```diff
@@ -597,9 +611,10 @@ class WrappedDualView {
     // check use counts and take care of sync/modify for the underlying DualView
     //
     // The logic is this:
-    // 1) For non-CUDA archtectures where there the host/device pointers are aliased
+    // 1) If WrappedDualView tracking is disabled, then never take the sync path.
+    // 2) For non-CUDA archtectures where there the host/device pointers are aliased
     // we don't need the "sync path."
-    // 2) For CUDA, we always need the "sync path" if we're using the CudaUVMSpace (we need to make sure
+    // 3) For CUDA, we always need the "sync path" if we're using the CudaUVMSpace (we need to make sure
```
romintomasetti, line 617: * For documentation purposes, might be good to move this documentation such that Doxygen can pick it as the description for `needsSyncPath`?
* I am not sure to understand why these sentences are related to `Cuda` only. Isn't the WDV also used for the `HIP` case?

packages/tpetra/core/src/Tpetra_Details_WrappedDualView.cpp
```diff
@@ -0,0 +1,60 @@
...
+void enableWDVTracking()
+{
+  if(wdvTrackingEnabled)
+    throw std::runtime_error("WrappedDualView refcount tracking is already enabled!");
+  wdvTrackingEnabled = true;
+}
+
+void disableWDVTracking()
+{
+  if(!wdvTrackingEnabled)
```
romintomasetti, line 54: I don't know how `Tpetra` positions w.r.t. task-based computing, but I guess such tricks based on global variable are dangerous from that view point?

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
```diff
@@ -2609,6 +2609,10 @@ class GetLocalDiagCopy {
         const auto policy =
           policy_type(numExportLIDs, 1, 1)
           .set_scratch_size(0, Kokkos::PerTeam(sizeof(GO)*maxRowLength));
+        // The following parallel_for needs const access to the local values of src.
+        // (the local graph is also accessed on host, but this does not use WDVs).
+        getValuesHost();
```
romintomasetti, line 2614: I guess the main goal of this is to call `WDV::modify_host`.

Instead of calling it implicitly through `getValuesHost`, wouldn't it be more readable to do so more explicitly?

packages/tpetra/core/src/Tpetra_Details_WrappedDualView.cpp
romintomasetti: It seems you are mimicking the `Kokkos::View` ref counting behavior. See for instance `packages/kokkos/core/src/Kokkos_Parallel.hpp`:

```c++
Kokkos::Impl::shared_allocation_tracking_disable();
Impl::ParallelFor<FunctorType, ExecPolicy> closure(functor, inner_policy);
Kokkos::Impl::shared_allocation_tracking_enable();
```
so when the view is copied on the device, it stops reference counting.

Note that in this case, the variable is **thread_local**:
```c++
static inline thread_local int t_tracking_enabled = 1;
```

romintomasetti (changes requested):
Hi @brian-kelley ! Thanks for the work on this :)

I have suggestions/questions from my side, but I'm not that of an expert in `WDV` like you are :smiley:

packages/tpetra/core/src/Tpetra_BlockCrsMatrix_def.hpp
brian-kelley: Basically yeah, but there are a few necessary things that these function do:
- Sync to the requested space (if data was modified in the other one)
- Mark modified in the requested space (unless read-only access)
- Check the reference counts to make sure a view in the other space isn't already alive

We want to do all of these things before turning off the tracking, so the WDV getView.... functions are the only way to do them all in one line.

packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp
brian-kelley: That is true, this does apply to any GPU-like (heterogeneous memory) backend. Kokkos also recently added the generalization of CudaUVMSpace, called "SharedSpace", for Cuda, HIP and SYCL.

packages/tpetra/core/src/Tpetra_Details_WrappedDualView.cpp
romintomasetti: I think I would be fine with the "global flag" approach as long as it is `thread_local`. The use case I see is:
- Thread 1 works on a problem with `Tpetra` object `OA`
- Thread 2 works on another problem with `Tpetra` object `OB`

These threads do uncorrelated things so the global flag should be *per-thread*.

I agree that having the flag *per-WDV* could be cumbersome in some cases (e.g. if several WDV are used in a parallel region).

As a side question, do you think using `execution_space::in_parallel` could be an alternative? I don't know what's the status of this `in_parallel` because it seems nobody uses that.
```

packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp
brian-kelley: @romintomasetti I updated the comments like you suggested. I can build the doxygen HTML locally and it looks fine now, but only if I mark the function "public:" (I guess doxygen will not run for private functions by default). But if we build docs for developers in the future, it will work here now.

packages/tpetra/core/src/Tpetra_Details_WrappedDualView.cpp
brian-kelley: I also wanted to say, to concurrently use Tpetra objects, the one approach that would work is to split the MPI comm. Say the global comm has 8 ranks, over 4 nodes. If you make sub-communicators for the even and odd ranks, you could then put ``OA`` on the evens and ``OB`` on the odds, and effectively get similar behavior without thread safety problems.

brian-kelley:
@csiefer2 This is ready to go, can you give it another review? Only new thing is WDV skips sync path in parallel kernels automatically - no more enable/disableWDVTracking.

PR #11685, opened 2023-03-16
Moves KokkosContainers from an optional to a required dependency

hkthorn:
@trilinos/stokhos 

## Motivation

We often build Trilinos for Xyce with “optional” packages and capabilities disabled, so we sometimes come across necessary headers that aren’t installed when they need to be.  I did come across an issue with that in Stokhos.

The Kokkos_StaticCrsGraph.hpp file is included in several files in stokhos/src/kokkos, which is in the KokkosContainers subpackage. That is an optional dependency in the Dependencies.cmake file, so those headers are not installed if optional dependencies are not included.  The result is a build failure in Xyce because, when Stokhos is enabled, that header is not found.  By moving KokkosContainers to a required dependency, this build failure goes away.

## Stakeholder Feedback

I have tested the changes against Xyce, using an installed version of Trilinos and the build issue is resolved.
## Testing

OSX GCC12 serial and parallel


hkthorn:
@etphipp @sebrowne Xyce will need this PR cherry picked onto the Trilinos 14.0 branch when it goes through.  What is the process for that?  Thanks!

etphipp (approved):
Thanks @hkthorn!

sebrowne:
@hkthorn this should be more or less the same as PR -> develop branch (PR -> trilinos-release-14-0 branch).  After that, if we need to do a 14.0.1 release, the Framework team will have to re-tag and re-release off of that branch.

For specific process, the steps would (could) be:
check out trilinos-release-14-0-branch
create and check out topic branch
cherry-pick this commit
push topic branch to fork
open PR from topic branch -> trilinos-release-14-0-branch
test and merge PR

PR #11686, opened 2023-03-16
Epetra: Throw if CrsGraph indices are negative

### LINKED ISSUE #10249: PackageName:  General Summary of the Enhancement
## Enhancement
@trilinos/\Epetra\>

When using more than MAX_INT local entries in Epetra_CrsGraph the code runs into a segfault 
when calling OptimizeStorage() as this results in negative local indices.

There are two places in MakeLocalIndices [1](https://github.com/trilinos/Trilinos/blob/7f8bbe22f1fb97e72857337ca696c3b69be450b6/packages/epetra/src/Epetra_CrsGraph.cpp#L1820) and [2](https://github.com/trilinos/Trilinos/blob/7f8bbe22f1fb97e72857337ca696c3b69be450b6/packages/epetra/src/Epetra_CrsGraph.cpp#L1851)
in which the resulting local index is checked for an error in Epetra_BlockMap::LID(gid) 
by comparing the result with -1.
Couldn't (result < 0) be checked instead to catch this error as well?

GrahamBenHarper:
@trilinos/epetra

Epetra's CrsGraph previously only checked if an LID was valid with `if(LID != -1)`, which did not cover the case where local indices overflow, as seen in #10249. This addresses it by instead defining `LID > -1` to be valid, so it will now throw on overflows.

Closes #10249.

### DIFF
--- a/packages/epetra/src/Epetra_CrsGraph.cpp
+++ b/packages/epetra/src/Epetra_CrsGraph.cpp
@@ -1817,9 +1817,9 @@ int Epetra_CrsGraph::MakeIndicesLocal(const Epetra_BlockMap& domainMap, const Ep
       for(int j = 0; j < NumIndices; j++) {
       int GID = ColIndices[j];
       int LID = colmap.LID(GID);
-      if(LID != -1)
+      if(LID > -1)
         ColIndices[j] = LID;
-      else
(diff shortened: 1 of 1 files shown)


jhux2 (approved):
lgtm

Issue #11704, opened 2023-03-21
Tpetra: CRS reader only fills row-pointer array up to row of final entry


cwpearson:
https://github.com/trilinos/Trilinos/blob/f79103208f046aabc1a5b4bb6ce534432b372183/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp#L3663-L3688

Set the first row-pointer entry: 
https://github.com/trilinos/Trilinos/blob/f79103208f046aabc1a5b4bb6ce534432b372183/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp#L3665

Set the last row-pointer entry:
https://github.com/trilinos/Trilinos/blob/f79103208f046aabc1a5b4bb6ce534432b372183/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp#L3688

Set row-pointer array for any row with a non-zero:
https://github.com/trilinos/Trilinos/blob/f79103208f046aabc1a5b4bb6ce534432b372183/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp#L3676-L3678

This skips setting all row-pointer array values between the row with the last non-zero, and the final row.

PR #11705, opened 2023-03-21
Tpetra: set all row-pointer entries in readSparse

cwpearson:
@trilinos/tpetra 

Instead of updating the row-pointer array only when we find a non-zero, use the previously-computed `numEntriesPerRow`. This prevents us from missing any rows after the last non-zero.

## Motivation
Resolve https://github.com/trilinos/Trilinos/issues/11704

### DIFF
--- a/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
+++ b/packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
@@ -3673,19 +3673,17 @@ namespace Tpetra {
                   << prvRow << ", at curPos = " << curPos << ".  Please report "
                   "this bug to the Tpetra developers.");
                 if (curRow > prvRow) {
-                  for (global_ordinal_type r = prvRow+1; r <= curRow; ++r) {
-                    rowPtr[r] = curPos;
-                  }
                   prvRow = curRow;
                 }
                 numEntriesPerRow[curRow]++;
                 colInd[curPos] = curEntry.colIndex();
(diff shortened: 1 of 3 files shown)


packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
```diff
@@ -3673,19 +3673,17 @@ namespace Tpetra {
                   << prvRow << ", at curPos = " << curPos << ".  Please report "
                   "this bug to the Tpetra developers.");
                 if (curRow > prvRow) {
-                  for (global_ordinal_type r = prvRow+1; r <= curRow; ++r) {
-                    rowPtr[r] = curPos;
-                  }
```
jhux2, line 3678: From #11704 
> This skips setting all row-pointer array values between the row with the last non-zero, and the final row.

@cwpearson Is there a test that will expose this error?

brian-kelley (dismissed):

packages/tpetra/core/inout/MatrixMarket_Tpetra.hpp
cwpearson: I don't think so, let me see if I can remedy that

cwpearson: I have added one

jhux2 (dismissed):
lgtm. @cwpearson Thanks for adding a test.

jhux2 (approved):
lgtm

PR #11706, opened 2023-03-22
Zoltan2: Update TpetraCrsColorer's Zoltan2 version

JacobDomagala:
@trilinos/zoltan2


egboman (approved):
This version enables TpetraCrsColorer to work with Zoltan2. I approve this now since there is a user that wants to try it out. 

It is not clear why the A+A' symmetrization is needed, but it works. In the future, we will investigate partial distance-2 (PD2) on the bipartite graph, which might give fewer colors.

etphipp:
@etphipp

PR #11708, opened 2023-03-22
Ifpack2: Fix tests for SYCL

masterleinad:
@trilinos/ifpack2

## Motivation
Running on Intel GPUs using SYCL.

## Testing
Running the `Ifpack2` unit tests with `Kokkos_ENABLE_SYCL=ON`.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -205,6 +205,8 @@ namespace Ifpack2 {
     template<> struct is_sycl<Kokkos::Experimental::SYCL> { enum : bool { value = true  }; };
 #endif
 
+    template<typename T> struct is_device                  { enum : bool { value = is_cuda<T>::value || is_hip<T>::value || is_sycl<T>::value }; };
+
     
     ///
     /// execution space instance
@@ -675,9 +677,7 @@ namespace Ifpack2 {
           const auto pid_send_value = pids.send[i];
           for (local_ordinal_type j=0,jend=epids.size();j<jend;++j)
             if (epids[j] == pid_send_value) lids_send_host[cnt++] = elids[j];
-#if !defined(__HIP_DEVICE_COMPILE__) && !defined(__CUDA_ARCH__) && !defined(__SYCL_DEVICE_ONLY__)
           TEUCHOS_ASSERT(static_cast<size_t>(cnt) == offset_host.send[i+1]);
-#endif
         }
         Kokkos::deep_copy(lids.send, lids_send_host);
       }
@@ -738,8 +738,7 @@ namespace Ifpack2 {
       }
 
       // ======================================================================
-      // Async version using cuda stream
-      // - cuda only with kokkos develop branch
+      // Async version using execution space instances
       // ======================================================================
 
 #if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP) || defined(KOKKOS_ENABLE_SYCL)
@@ -867,11 +866,11 @@ namespace Ifpack2 {
         waitall(reqs.send.size(), reqs.send.data());
 #endif // HAVE_IFPACK2_MPI
       }
-#endif //defined(KOKKOS_ENABLE_CUDA)
+#endif //defined(KOKKOS_ENABLE_CUDA|HIP|SYCL)
 
       // ======================================================================
-      // Generic version without using cuda stream
-      // - only difference between cuda and host architecture is on using team
+      // Generic version without using execution space instances
+      // - only difference between device and host architecture is on using team
       //   or range policies.
       // ======================================================================
       template<typename PackTag>
@@ -886,8 +885,7 @@ namespace Ifpack2 {
         const local_ordinal_type mv_blocksize = blocksize_*num_vectors;
         const local_ordinal_type idiff = iend_ - ibeg_;
         const auto abase = buffer_.data() + mv_blocksize*ibeg_;
-        if (is_cuda<execution_space>::value || is_hip<execution_space>::value) {
-#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP) || defined(KOKKOS_ENABLE_SYCL)
+        if constexpr (is_device<execution_space>::value) {
           using team_policy_type = Kokkos::TeamPolicy<execution_space>;
           local_ordinal_type vector_size(0);
           if      (blocksize_ <=  4) vector_size =  4;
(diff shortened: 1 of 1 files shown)


packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -675,9 +677,7 @@ namespace Ifpack2 {
           const auto pid_send_value = pids.send[i];
           for (local_ordinal_type j=0,jend=epids.size();j<jend;++j)
             if (epids[j] == pid_send_value) lids_send_host[cnt++] = elids[j];
-#if !defined(__HIP_DEVICE_COMPILE__) && !defined(__CUDA_ARCH__) && !defined(__SYCL_DEVICE_ONLY__)
           TEUCHOS_ASSERT(static_cast<size_t>(cnt) == offset_host.send[i+1]);
-#endif
```
masterleinad, line 680: This is a host-only function. Is there a good reason to still check the device compiler pass? AFAICT, it shouldn't matter.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3636,40 +3614,17 @@ namespace Ifpack2 {
         IFPACK2_BLOCKTRIDICONTAINER_TIMER("BlockTriDi::ComputeResidual::<SeqTag>");
 
         y = y_; b = b_; x = x_;
-        if (is_cuda<execution_space>::value) {
-#if defined(KOKKOS_ENABLE_CUDA)
+        if constexpr (is_device<execution_space>::value) {
```
kliegeois, line 3617: This part probably needs some modifications as the features are not 1-to-1 with the previous code.

The proposed modification calls `ComputeResidualVectorRecommendedCudaVectorSize` if the `is_device<execution_space>::value` is True. In order to be consistent with the previous implementation, we might need a function:

```
template<typename T> 
    static inline int ComputeResidualVectorRecommendedVectorSize(const int blksize,
                                                                     const int team_size) {
    if ( is_cuda<T>::value )
      return ComputeResidualVectorRecommendedCudaVectorSize(blksize, team_size);
    if ( is_hip<T>::value )
      return ComputeResidualVectorRecommendedHIPVectorSize(blksize, team_size);
    if ( is_sycl<T>::value )
      return ComputeResidualVectorRecommendedSYCLVectorSize(blksize, team_size);
    return -1;
    }
```

This comment holds for the other functions such as `ExtractAndFactorizeRecommendedCudaTeamSize`.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3686,36 +3641,17 @@ namespace Ifpack2 {
...
         } else {
           y_packed = y_packed_;
         }
 
-        if (is_cuda<execution_space>::value) {
-#if defined(KOKKOS_ENABLE_CUDA)
+        if constexpr(is_device<execution_space>::value) {
           const local_ordinal_type blocksize = blocksize_requested;
           const local_ordinal_type team_size = 8;
           const local_ordinal_type vector_size = ComputeResidualVectorRecommendedCudaVectorSize(blocksize, team_size);
```
kliegeois, line 3657: Same comment as above.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3903,81 +3757,7 @@ namespace Ifpack2 {
           default : BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL( 0);
           }
 #undef BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL
-#endif
-	} else if (is_hip<execution_space>::value) {
-#if defined(KOKKOS_ENABLE_HIP)
-          const local_ordinal_type blocksize = blocksize_requested;
-          const local_ordinal_type team_size = 8;
-          const local_ordinal_type vector_size = ComputeResidualVectorRecommendedHIPVectorSize(blocksize, team_size);
```
kliegeois, line 3911: Same as above.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3742,69 +3678,7 @@ namespace Ifpack2 {
...
-          case   9: BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL( 9);
-          case  10: BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL(10);
-          case  11: BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL(11);
-          case  16: BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL(16);
-          case  17: BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL(17);
-          case  18: BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL(18);
-          default : BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL( 0);
-          }
-#undef BLOCKTRIDICONTAINER_DETAILS_COMPUTERESIDUAL
-#endif
```
kliegeois, line 3803: Same as above.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -205,6 +205,8 @@ namespace Ifpack2 {
     template<> struct is_sycl<Kokkos::Experimental::SYCL> { enum : bool { value = true  }; };
 #endif
 
+    template<typename T> struct is_device                  { enum : bool { value = is_cuda<T>::value || is_hip<T>::value || is_sycl<T>::value }; };
```
kliegeois, line 208: I like this idea; this simplifies the code a lot, thanks!

kliegeois (changes requested):
Thanks @masterleinad for this PR!
I think that this PR goes into the right direction but requires a little extra work.
Right now, the PR is dropping a required feature: the ability to specify different team size and vector length for different execution space (see comments below).

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
masterleinad: Ah, my bad I missed that the actually call different functions. I followed your suggestion now.

masterleinad: `ExtractAndFactorizeRecommended*TeamSize` is actually only used in functions that are specific to be respective backend anyway.

kliegeois (approved):
Thanks @masterleinad !
This looks good to me.

lucbv (approved):
Looks good to me

PR #11709, opened 2023-03-22
Tpetra: Permanent fix for #11655

brian-kelley:
@trilinos/tpetra
@trilinos/kokkos-kernels 

## Motivation

Permanent fix for #11655, where Tpetra was passing unsorted matrices to KokkosKernels SpGEMM and triggering debug checks. Revert the workaround in KokkosKernels (#11663).

## Testing

@rppawlo With these changes, all MiniEM tests are still passing with OpenMP+Debug. I would be surprised if EMPIRE still runs into the issue, but could you please check (either before or after this merges)?

### DIFF
--- a/packages/kokkos-kernels/sparse/src/KokkosSparse_spgemm_symbolic.hpp
+++ b/packages/kokkos-kernels/sparse/src/KokkosSparse_spgemm_symbolic.hpp
@@ -140,26 +140,15 @@ void spgemm_symbolic(KernelHandle *handle,
   // Verify that graphs A and B are sorted.
   // This test is designed to be as efficient as possible, but still skip
   // it in a release build.
-  //
-  // Temporary fix for Trilinos issue #11655: Only perform this check if a TPL
-  // is to be called. The KokkosKernels (non-TPL) implementation does not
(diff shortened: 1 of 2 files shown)


jhux2 (approved):
lgtm

brian-kelley:
@jhux2 There isn't really a timeline for that.

@cgcgcg I'm realizing now that enforcing sortedness unconditionally was a mistake, it would be better to have both paths. Some TPLs require it and some don't, so for the ones that do we can call KK as a fallback (for the time being, anyway). Also, it should not be enforced if the Tpetra LTG algo is used.

I'm going to try to get this into KK for the 4.1 release. Until then, the workaround can stay.

cgcgcg:
Thanks, @brian-kelley! That makes sense to me.

ndellingwood:
@brian-kelley should the kokkos-kernels changes from https://github.com/trilinos/Trilinos/pull/11663 get cherry-picked to the release-candidate-4.0.01 branch to avoid clobber with the Trilinos #11663 patch (since the long-term update to kokkos-kernels will be part of later 4.1 release)?

PR #11714, opened 2023-03-23
Tpetra: Add `Tpetra::Details::Spaces`, `...::Slot`, and `...::User`

cwpearson:
@trilinos/tpetra

## Summary
Adds `Tpetra::Details::ExecutionSpaces` namespace with a variety of free functions to create and cache Kokkos execution space instances. In this PR, only `Kokkos::Cuda` execution space instances are special-cased for performance, but all backends are supported.

Adds `Tpetra::Details::ExecutionSpaces::User` which a Tpetra object can inherit from, which provides a couple utility functions to access execution space instances and scope those instances to the lifetime of the object.

Makes `Tpetra::CrsMatrix` a `Tpetra::Details::ExecutionSpaces::User` but since `Tpetra::CrsMatrix` is otherwise unchanged, has no impact on its behavior.

Adds a `.clang-format` file generated by clang-format-8 in the LLVM style (`clang-format-8 --style=LLVM --dump-config`) for a consistent format for future Tpetra code

## Motivation
Needed to manage execution space instances and priorities in communication-computation overlap: https://github.com/trilinos/Trilinos/pull/10926

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Details_ExecutionSpaces.cpp
+++ b/packages/tpetra/core/src/Tpetra_Details_ExecutionSpaces.cpp
@@ -1,0 +1,65 @@
+#include "Tpetra_Details_ExecutionSpaces.hpp"
+
+#include <sstream>
+#include <vector>
+
+namespace Tpetra {
+namespace Details {
+namespace Spaces {
+
+#if defined(KOKKOS_ENABLE_CUDA)
+// cuda has default stream priority 0
+CudaInfo::CudaInfo() : initialized_(false), mediumPrio_(0) {}
+#endif
+
+void lazy_init() {
+#if defined(KOKKOS_ENABLE_CUDA)
+  if (!cudaInfo.initialized_) {
+    cudaInfo.initialized_ = true;
+    CUDA_RUNTIME(cudaEventCreateWithFlags(&cudaInfo.execSpaceWaitEvent_,
+                                          cudaEventDisableTiming));
+    CUDA_RUNTIME(cudaDeviceGetStreamPriorityRange(&cudaInfo.lowPrio_,
+                                                  &cudaInfo.highPrio_));
+
+    // We expect
+    //   medium priority should be 0
+    //   lower numbers to be higher priorities
+    //   low is at least as good as medium
+    //   medium is at least as good as high
+    if (!(cudaInfo.lowPrio_ >= cudaInfo.mediumPrio_ &&
+          cudaInfo.mediumPrio_ >= cudaInfo.highPrio_)) {
+      std::stringstream ss;
+      ss << "CUDA stream priority does not follow assumptions."
+         << " low=" << cudaInfo.lowPrio_ << " medium=" << cudaInfo.mediumPrio_
+         << " high=" << cudaInfo.highPrio_
+         << " Please report this to the Tpetra developers.";
+      throw std::runtime_error(ss.str());
+    }
+  }
+#endif
+}
+
+/* -----------------------------
+    Space Management Singletons
+   -----------------------------*/
+
+#if defined(KOKKOS_ENABLE_CUDA)
+/*extern*/ InstanceLifetimeManager<Kokkos::Cuda> cudaSpaces;
+/*extern*/ CudaInfo cudaInfo;
+#endif
+#ifdef KOKKOS_ENABLE_SERIAL
+/*extern*/ InstanceLifetimeManager<Kokkos::Serial> serialSpaces;
+#endif
+#ifdef KOKKOS_ENABLE_OPENMP
+/*extern*/ InstanceLifetimeManager<Kokkos::OpenMP> openMPSpaces;
+#endif
+#ifdef KOKKOS_ENABLE_HIP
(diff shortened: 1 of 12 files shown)


packages/tpetra/core/src/Tpetra_Details_ExecutionSpaces.hpp
```diff
@@ -0,0 +1,476 @@
...
+ * i
+ */
+template <typename ExecSpace, Priority priority = Priority::medium,
+          IsSYCL<ExecSpace> = true>
+Teuchos::RCP<const ExecSpace> space_instance(int i = 0) {
+  return SYCLSpaces.space_instance<priority>(i);
+}
+#endif
+
+/*! \briefget a strong Teuchos::RCP to Tpetra-managed Kokkos execution space
```
maartenarnst, line 364:
```suggestion
/*! \brief get a strong Teuchos::RCP to Tpetra-managed Kokkos execution space
```

packages/tpetra/core/src/Tpetra_Details_ExecutionSpaces.cpp
```diff
@@ -0,0 +1,65 @@
...
+    //   lower numbers to be higher priorities
+    //   low is at least as good as medium
+    //   medium is at least as good as high
+    if (!(cudaInfo.lowPrio_ >= cudaInfo.mediumPrio_ &&
+          cudaInfo.mediumPrio_ >= cudaInfo.highPrio_)) {
+      std::stringstream ss;
+      ss << "CUDA stream priority does not follow assumptions."
+         << " low=" << cudaInfo.lowPrio_ << " medium=" << cudaInfo.mediumPrio_
+         << " high=" << cudaInfo.highPrio_
+         << " Please report this to the Tptera developers.";
```
maartenarnst, line 35:
```suggestion
         << " Please report this to the Tpetra developers.";
```

packages/tpetra/core/test/Spaces/SpaceUser.cpp
```diff
@@ -0,0 +1,142 @@
...
+    auto e3 =
+        space_instance<ExecSpace, Tpetra::Details::Spaces::Priority::high>();
+    auto e4 = space_instance<ExecSpace>(Tpetra::Details::Spaces::Priority::low);
+    auto e5 =
+        space_instance<ExecSpace>(Tpetra::Details::Spaces::Priority::medium);
+    auto e6 =
+        space_instance<ExecSpace>(Tpetra::Details::Spaces::Priority::high);
+  }
+
+  // request the same instance 1B times should be pretty quick
```
maartenarnst, line 73:
```suggestion
  // request the same instance B times should be pretty quick
```

packages/tpetra/core/src/Tpetra_Details_ExecutionSpaces.hpp
```diff
@@ -0,0 +1,476 @@
...
+  return ExecSpace();
+}
+
+/*! \brief Construct a Kokkos::Cuda execution space instance with the requested
+   priority
+
+    This creates a stream with the requested priority, and attaches it to a new
+   space
+*/
+#if defined(KOKKOS_ENABLE_CUDA)
```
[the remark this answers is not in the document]
cwpearson: I'd have to put everything except `return true` in the header file and add `extern`, and then add the entire thing as written here to the `.cpp` This seemed simpler

cwpearson:
There are two orthogonal issues here that produce execution-space type-specific code.

## Instance Cache

The basic problem this is trying to prevent is that every Tpetra object that wants an execution space instance creating their own. This will pollute any profiles and has a small (?) performance cost.

Fundamentally we then have to have a cache of instances for each enabled execution space type.

The only way I know how to implement a map-like thing in C++ that has a type as a key (not a value with a type, but the type itself), and a cache of values as a value, is by using a C++ identifier (e.g. `cudaSpaces`) as a key, and then the cache of those instances as a value. Then we can interact with that key-value store by specializing functions on the type of execution space instances.

If we are willing to forgo the cache, and allow two different Tpetra objects that request instances to always result in two different instances, then this part can be dramatically simplified.

## Special-Casing Exec Space Sync

Kokkos currently requires fencing the host when fencing exec space instances. Working around that requires instance-specific code as long as Kokkos doesn't support it natively. If it got merged into Kokkos this part could be removed.

packages/tpetra/core/src/Tpetra_Details_ExecutionSpaces.hpp
```diff
@@ -0,0 +1,476 @@
...
+                             const int line) {
+  if (err != cudaSuccess) {
+    std::stringstream ss;
+    ss << file << ":" << line << ": ";
+    ss << cudaGetErrorString(err);
+    throw std::runtime_error(ss.str());
+  }
+}
+#define CUDA_RUNTIME(x)                                                        \
+  Tpetra::Details::Spaces::success_or_throw((x), __FILE__, __LINE__)
```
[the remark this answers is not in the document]
cwpearson: How about `TPETRA_DETAILS_SPACES_CUDA_RUNTIME`? A bit clumsy, but the hope is this would eventually be a part of Kokkos and this code could be deleted.

PR #11715, opened 2023-03-23
Tpetra: early exit from CRS remove zeros

### LINKED ISSUE #11671: Tpetra: possible optimization when removing zeros
## Enhancement
This is a potential follow-up to #11604.  In `removeCrsMatrixZeros` there should be a check whether any zeros were actually removed.  If not, then the function can just return early.

brian-kelley:
@trilinos/tpetra 

## Motivation

In ``Tpetra::removeCrsMatrixZeros``, skip the fill-complete if there were no zeros to remove in the local matrices of all ranks (it cannot be skipped on only some ranks, because expertStaticFillComplete might do collectives, so all ranks must participate).

## Related Issues

* Closes #11671 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing

Tested in Tpetra: CrsMatrix_UnitTests4


jhux2 (approved):
Lgtm.  Thanks, @brian-kelley!

PR #11717, opened 2023-03-24
amesos2: Move declaration and specializations of slu_dist_mult for no…

### LINKED ISSUE #11716: Amesos2:  Superludist build broken when complex is disabled.
## Bug Report
@iyamazaki #11710 broke all of SPARC's Trilinos integration builds because we do not enable complex and the declarations of slu_dist_mult that were added are all inside an `#ifdef TEUCHOS_ENABLE_COMPLEX` but are used unconditionally.

vbrunini:
…n-complex types outside of ifdef.

Fixes #11716.

@trilinos/<teamName>

### DIFF
--- a/packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
+++ b/packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
@@ -111,27 +111,26 @@ extern "C" {
 
 } // end extern "C"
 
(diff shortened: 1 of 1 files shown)


ndellingwood (approved):

iyamazaki:
Thank you for the fix, @vbrunini.

PR #11724, opened 2023-03-24
Ifpack2: Add implementation of fourth-kind chebyshev polynomial smoothers

### LINKED ISSUE #11721: Ifpack2:  Add in Chebyshev polynomials of the fourth-kind
## Enhancement
@jhux2 @csiefer2 

Add an implementation of the fourth-kind Chebyshev polynomial smoothers from James Lottes's research: https://arxiv.org/pdf/2202.08830.pdf.
The basis idea is that, rather than targeting an interval $[\lambda_{min}, \lambda_{max}]$ to smooth on, one can instead omit the $\lambda_{min}$ parameter by directly optimizing the polynomial with respect to a given multigrid error bound. The scaled fourth-kind Chebyshev is the optimal polynomial with respect to a two-level error bound dating back to Hackbusch (see: S2, eq 4 from the Lottes preprint mentioned above).

One can further sharpen the error bound for the multi-level case, see Lemma 3.2 from Lottes. The fourth-kind Chebyshev polynomials form a convenient basis for the optimal polynomial with respect to this error bound. The tabulated $\beta_k$ coefficients are the result of this optimization problem, which I've included up to order 16.

I've found using either the fourth or "optimized" fourth-kind Chebyshev smoother outperforms Chebyshev polynomial smoothing of the 1st-kind (i.e., the standard Ifpack2 algorithm), especially at higher orders: https://arxiv.org/pdf/2210.03179.pdf.

MalachiTimothyPhillips:
@trilinos/ifpack2
@jhux2 @csiefer2

## Motivation
#11721 

Add in fourth-kind Chebyshev polynomial smoothing, which in certain cases may outperform the first-kind Chebyshev polynomial smoothers.

## Related Issues

* Closes #11721 

## Testing
Added test coverage through `packages/ifpack2/test/unit_tests/Ifpack2_UnitTestChebyshev2.cpp`

### DIFF
--- a/packages/ifpack2/doc/UsersGuide/options.tex
+++ b/packages/ifpack2/doc/UsersGuide/options.tex
@@ -376,7 +376,7 @@ The following parameters are used in the Chebyshev method:
     % {bool}
     % {\false}
     % {If true, use the textbook variant; otherwise, use the \ifpack{} variant.}
-\ccc{chebyshev: kind}
+\ccc{chebyshev: algorithm}
     {string}
     {"first"}
     {The algorithm for the Chebyshev polynomial. Currently supports
--- a/packages/ifpack2/doc/UsersGuide/options.tex
+++ b/packages/ifpack2/doc/UsersGuide/options.tex
@@ -376,6 +376,22 @@ The following parameters are used in the Chebyshev method:
     % {bool}
     % {\false}
     % {If true, use the textbook variant; otherwise, use the \ifpack{} variant.}
+\ccc{chebyshev: kind}
+    {string}
+    {"first"}
+    {The algorithm for the Chebyshev polynomial. Currently supports
+     the default \ifpack{} variant, otherwise known as the first-kind ("first") Chebyshev polynomials.
+     Chebyshev smoothing via the fourth-kind ("fourth") and optimized fourth-kind ("opt_fourth")
+     are also implemented as in ~\cite{Lottes2022}.
+     Note that these latter two algorithms do not require the ad-hoc minimum eigenvalue estimate
+     as the standard \ifpack{} algorithm ("first").
+     The textbook algorithm ("textbook") as described in ~\cite{Saad2003} is also available.
+     However, the "first", "fourth", and "opt_fourth" kinds are preferred.
+     Currently supported: "first", "fourth", "opt_fourth", and "textbook".
+     Generally, the "fourth" and "opt_fourth" provide better smoothing properties
+     when employing the Chebyshev polynomials as multigrid smoothers, especially at orders
+     greater than or equal to 3.
+    }
 \ccc{chebyshev: compute max residual norm}
     {bool}
     {\false}
--- a/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
@@ -650,11 +650,17 @@ setParameters (Teuchos::ParameterList& plist)
       "Ifpack2::Chebyshev: Ifpack2 only supports \"first\", \"textbook\", \"fourth\", and \"opt_fourth\", for \"chebyshev: kind\".");
   }
 
-#ifndef IFPACK2_ENABLE_DEPRECATED_CODE
-  if (plist.isParameter ("chebyshev: textbook algorithm")) {
-    const bool textbookAlgorithm = plist.get<bool> ("chebyshev: textbook algorithm");
-    if(textbookAlgorithm){
-      chebyshevKind = "textbook";
+#ifdef IFPACK2_ENABLE_DEPRECATED_CODE
+  // to preserve behavior with previous input decks, only read "chebyshev:textbook algorithm" setting
+  // if a user has not specified "chebyshev: kind"
+  if (!plist.isParameter ("chebyshev: kind")) {
+    if (plist.isParameter ("chebyshev: textbook algorithm")) {
+      const bool textbookAlgorithm = plist.get<bool> ("chebyshev: textbook algorithm");
+      if(textbookAlgorithm){
+        chebyshevKind = "textbook";
+      } else {
+        chebyshevKind = "first";
+      }
     }
   }
 #endif
--- a/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
@@ -1666,7 +1666,8 @@ description () const {
       << ", lambdaMax: " << lambdaMaxForApply_
       << ", alpha: " << eigRatioForApply_
       << ", lambdaMin: " << lambdaMinForApply_
-      << ", boost factor: " << boostFactor_;
+      << ", boost factor: " << boostFactor_
+      << ", algorithm: " << chebyshevAlgorithm_;
   if (!userInvDiag_.is_null())
     oss << ", diagonal: user-supplied";
   oss << "}";
--- a/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
@@ -1329,7 +1329,7 @@ fourthKindApplyImpl (const op_type& A,
                      const V& D_inv) const
 {
   // standard 4th kind Chebyshev smoother has \beta_i := 1
-  std::vector<ScalarType> betas(numIters, 1.0)
+  std::vector<ScalarType> betas(numIters, 1.0);
   if(useOptimalWeights_){
     betas = optimalWeightsImpl<ScalarType>(numIters);
   }
@@ -1337,6 +1337,7 @@ fourthKindApplyImpl (const op_type& A,
   const ST zero = Teuchos::as<ST> (0);
   const ST one = Teuchos::as<ST> (1);
   const ST mone = Teuchos::as<ST> (-1);
+  const ST invEig = 1.0 / (lambdaMax * boostFactor_);
 
   if (zeroStartingSolution_) {
     X.putScalar (zero);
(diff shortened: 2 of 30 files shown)


cgcgcg:
Before the change, we had 2 algorithm options, `textbook` or `!textbook`. Instead of adding another Boolean, wouldn't it make sense to add an option "chebyshev: algorithm"? We could deprecate "chebyshev: textbook algorithm".

MalachiTimothyPhillips:
> Before the change, we had 2 algorithm options, `textbook` or `!textbook`. Instead of adding another Boolean, wouldn't it make sense to add an option "chebyshev: algorithm"? We could deprecate "chebyshev: textbook algorithm".

Sounds good to me -- thanks!

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
```diff
@@ -1298,6 +1318,62 @@ textbookApplyImpl (const op_type& A,
...
+    Z.update(rScale, P, zScale);
+
+    // X = X + betas[i] * Z
+    X.update(betas[i], Z, one);
+
+    // R = R - A*Z
+    A.apply(Z, P);
+
+    R.update(mone, P, one);
+
```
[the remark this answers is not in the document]
MalachiTimothyPhillips: This is not necessarily the same as the residual -- it is for the standard 4th-kind Chebyshev smoother with $\beta_i := 1$. However, for the optimized kind with $\beta \neq 1$. See top paragraph p. 18 from https://arxiv.org/pdf/2202.08830.pdf for more details.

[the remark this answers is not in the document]
cgcgcg: This guy:
https://github.com/trilinos/Trilinos/blob/3741eaed0aa290e8f18be7ac171c079606ce631b/packages/ifpack2/src/Ifpack2_Details_ChebyshevKernel_decl.hpp#L66-L69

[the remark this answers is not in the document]
MalachiTimothyPhillips: @cgcgcg I noticed that both of those code paths exist -- is the one you linked to for GPU support? If so, these changes should be implemented in both spots, right?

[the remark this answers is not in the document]
cgcgcg: The one I linked is used by the "Ifpack" apply which is the default code path. If you can reuse that `W := alpha*D_inv*(B-A*X) + beta*W` primitive I'd use that. Both will run on GPU, but the other code path has some performance tweaks.

MalachiTimothyPhillips:
> Before the change, we had 2 algorithm options, `textbook` or `!textbook`. Instead of adding another Boolean, wouldn't it make sense to add an option "chebyshev: algorithm"? We could deprecate "chebyshev: textbook algorithm".

Wrapped in a `#ifndef IFPACK2_ENABLE_DEPRECATED_CODE` -- is that the correct way to deprecate the command line argument?

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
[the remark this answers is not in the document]
MalachiTimothyPhillips: I was able to re-write it to use the Chebyshev kernels, albeit with one additional vector addition type operation to correctly handle the optimized 4th-kind case ($\beta_i \neq 1$).

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
```diff
@@ -638,16 +639,35 @@ setParameters (Teuchos::ParameterList& plist)
...
+    TEUCHOS_TEST_FOR_EXCEPTION(
+      chebyshevKind != "first" &&
+      chebyshevKind != "textbook" &&
+      chebyshevKind != "fourth" &&
+      chebyshevKind != "opt_fourth",
+      std::invalid_argument,
+      "Ifpack2::Chebyshev: Ifpack2 only supports \"first\", \"textbook\", \"fourth\", and \"opt_fourth\", for \"chebyshev: kind\".");
+  }
+
+#ifndef IFPACK2_ENABLE_DEPRECATED_CODE
```
cgcgcg, line 653: Maybe I'm confused. Shouldn't this be `#ifdef`?

packages/ifpack2/doc/UsersGuide/options.tex
```diff
@@ -376,6 +376,22 @@ \section{Chebyshev}\label{s:Chebyshev}
     % {bool}
     % {\false}
     % {If true, use the textbook variant; otherwise, use the \ifpack{} variant.}
+\ccc{chebyshev: kind}
```
cgcgcg, line 379: Why "chebyshev: kind" instead of "chebyshev: algorithm"?

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
```diff
@@ -638,16 +639,35 @@ setParameters (Teuchos::ParameterList& plist)
...
   if (plist.isParameter ("chebyshev: textbook algorithm")) {
-    textbookAlgorithm = plist.get<bool> ("chebyshev: textbook algorithm");
+    const bool textbookAlgorithm = plist.get<bool> ("chebyshev: textbook algorithm");
```
cgcgcg, line 655: What about the case of "chebyshev: textbook algorithm" being in the parameter list and its value being `false`? I guess that should results in `chebyshevKind = "first"`?

cgcgcg (commented):
Thank you for making these changes!

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
MalachiTimothyPhillips: My bad -- I think you are right!

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
MalachiTimothyPhillips: Not necessarily -- what if a user sets both:

```
"chebyshev: kind", "fourth"
"chebyshev: textbook algorithm", "false" 
```

I would think that this would mean `chebyshevKind = "fourth"`.

At the same time, I should also explicitly guard against:
```
"chebyshev: kind", "fourth" // or any other setting that is not textbook
"chebyshev: textbook algorithm", "true" 
```

since it's not clear to a user which should take precedence.

```
"chebyshev: kind", "textbook"
"chebyshev: textbook algorithm", "true" 
```

should also be okay -- the user just specified more-or-less the same option twice.

However,

```
"chebyshev: kind", "textbook"
"chebyshev: textbook algorithm", "false" 
```

should throw an error -- how are we supposed to interpret that?

:face_exhaling: All of this additional logic for a boolean knob that will (hopefully) soon be deprecated.

packages/ifpack2/doc/UsersGuide/options.tex
cgcgcg: Hm. Yeah, I see. Before you make changes, let's see if @jhux2 or @csiefer2 have strong preferences. I don't think "kind" is very descriptive, but neither is "algorithm".

packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
cgcgcg: :-) 

I think what we really want is that someone with an old input deck does not see a change of behavior. They would not have "chebyshev: kind" in their input deck, and it seems that no default value would get injected prior to those lines of code. They might however have "chebyshev: textbook algorithm".

Once the user specifies "chebyshev: kind" I feel it's safe to ignore "chebyshev: textbook algorithm".

packages/ifpack2/doc/UsersGuide/options.tex
MalachiTimothyPhillips: Changed to "algorithm" -- this is a pretty easy find/replace if we want to change it before merging this PR.

jhux2 (commented):
@MalachiTimothyPhillips Thanks for the contribution, this looks good.  I have one request.  Could you add a line to the smoother description, such as

`<< "algorithm: " << chebyshevAlgorithm_`?

 This would print in the multigrid summary, and at a glance we'd know which Chebyshev algorithm is being used.

https://github.com/trilinos/Trilinos/blob/461c497db44b9632d22c15c43ac1364bd7c2ac65/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp#L1659-L1674

MalachiTimothyPhillips:
> @MalachiTimothyPhillips Thanks for the contribution, this looks good. I have one request. Could you add a line to the smoother description, such as
> 
> `<< "algorithm: " << chebyshevAlgorithm_`?
> 
> This would print in the multigrid summary, and at a glance we'd know which Chebyshev algorithm is being used.
> 
> https://github.com/trilinos/Trilinos/blob/461c497db44b9632d22c15c43ac1364bd7c2ac65/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp#L1659-L1674

Thanks -- added in most recent commit!

jhux2 (dismissed):
lgtm

MalachiTimothyPhillips:
> @MalachiTimothyPhillips The update to the Chebyshev output has caused diff's in some MueLu gold files.

Thanks for the detailed instructions! I was able to reproduce the errors using `ctest`, and after running `make rebase` the `ctest` output was clean.

MalachiTimothyPhillips:
According to cdash, it seems the CUDA failure is due to `ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_21` timing out -- is there anything I need to do on my end, or were we just unlucky? It looks like that test passed two sha's ago (https://github.com/trilinos/Trilinos/pull/11724#issuecomment-1503798050). Note that the two new shas pertain to outputting the `chebyshevAlgorithm` and resolving diffs in the gold files from that change. Thanks!

cgcgcg:
@MalachiTimothyPhillips Nothing to do on your end, just unlucky.

PR #11720, opened 2023-03-24
MueLu: Merge Kokkos and non-Kokkos AmalgamationInfo&Factory

cgcgcg:
@trilinos/muelu 

## Motivation
Merges Amalgamation / Amalgamation_kokkos and AmalgamationFactory / AmalgamationFactory_kokkos.

Closes #11719

PR #11726, opened 2023-03-24
MueLu: Merge Aggregates and Aggregates_kokkos

GrahamBenHarper:
@trilinos/muelu 
@cgcgcg 

Based on #11720 
Part of #11658

 ## Top-level summary of this PR:
 - Renamed all instances of `Aggregates_kokkos` to `Aggregates` (similar for caps case), replaced `MueLu_Aggregates` with `MueLu_Aggregates_kokkos`, deleting the original `MueLu_Aggregates`.
 - Added an ArrayRCP interface for `ComputeAggregateSizes`
 - Added an additional member variable `aggregateSizesHost_` (to track the device view `aggregateSizes_`)
 - Added an additional member variable `geoDataKokkos_` (to preserve functionality between `IndexManager` and `IndexManager_kokkos`)
 - Switched to using Kokkos views in `FilteredAFactory`
 - Cleanup of many test drivers which did not release the pointers of input data after construction
 - Massive overhaul of visualization objects, as they previously stored long-term pointers to coordinate data

I think that's all the key points...

## Testing
Testing looks good in serial and on Weaver with a non-UVM build.

**Note:** I highly recommend using the split view for the important files in this PR (e.g. `Aggregates` files), because the workflow for this was essentially "copy Aggregates_kokkos -> Aggregates, add necessary interfaces", so the diff unnecessarily report a lot of single-line differences that aren't relevant.
image


GrahamBenHarper:
Note, mainly to myself: we use `isRoot` pretty liberally, and the fact that `isRoot_` lives on device now will likely be a problem in a heterogeneous build. We'll see how bad it is when the AT reports back.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
```diff
@@ -114,85 +144,133 @@ namespace MueLu {
...
 
     /*! @brief Returns a nonconstant vector that maps local node IDs to local aggregates IDs.
 
         For local node ID i, the corresponding vector entry v[i] is the local aggregate id to which i belongs on the current processor.
     */
-    RCP<LOMultiVector> & GetVertex2AggIdNonConst()     { return vertex2AggId_;       }
+    RCP<LOMultiVector>& GetVertex2AggIdNonConst()     { return vertex2AggId_;       }
 
-    /*! @brief Returns nonconstant vector that maps local node IDs to owning processor IDs.
+    /*! @brief Returns nonconsant vector that maps local node IDs to owning processor IDs.
```
cgcgcg, line 247: "nonconsant"

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_decl.hpp
```diff
@@ -64,7 +64,7 @@
 
 #include "MueLu_AmalgamationInfo_fwd.hpp"
 #include "MueLu_Aggregates_fwd.hpp"
-#include "MueLu_Aggregates_kokkos_fwd.hpp"
+#include "MueLu_Aggregates_fwd.hpp"
```
cgcgcg, line 67: duplicate include

packages/muelu/test/viz/Viz.cpp
```diff
@@ -294,7 +294,7 @@ int main_(Teuchos::CommandLineProcessor &clp, Xpetra::UnderlyingLib& lib, int ar
       }
 
       // AggregationExportFactory can't work with
-      // Aggregates_kokkos (yet)
+      // Aggregates (yet)
```
cgcgcg, line 297: delete comment?

packages/muelu/test/unit_tests/Aggregates.cpp
cgcgcg, line 1: Don't delete the unit test. We want to make sure it still passes.

packages/muelu/src/Utils/ForwardDeclaration/MueLu_Aggregates_kokkos_fwd.hpp
cgcgcg, line 1: Why do we still have this header file?

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
```diff
@@ -114,85 +144,133 @@ namespace MueLu {
...
 
         For local node ID i, the corresponding vector entry v[i] is the owning processor ID.
     */
-    const RCP<LOVector> & GetProcWinner() const   { return procWinner_;         }
+    const RCP<LOVector>& GetProcWinner() const   { return procWinner_;         }
 
     //! Returns true if node with given local node id is marked to be a root node
-    bool IsRoot(LO i) const               { return isRoot_[i];          }
+    KOKKOS_INLINE_FUNCTION
+    bool IsRoot(LO i) const                      { return isRoot_(i);          }
```
cgcgcg, line 266: This is used on host by the classical routines and on device by Kokkos factories, so this will most likely crash.

packages/muelu/test/viz/Viz.cpp
GrahamBenHarper: what about the `mueluList.set("use kokkos refactor", false);` right below it?

packages/muelu/test/unit_tests/Aggregates.cpp
GrahamBenHarper: oh good point. I didn't realize I got rid of that

packages/muelu/test/viz/Viz.cpp
GrahamBenHarper: right. I'll wait to crack down on the weird uses of `"use kokkos refactor"` until we merge everything then :)

packages/muelu/src/Utils/ForwardDeclaration/MueLu_Aggregates_kokkos_fwd.hpp
GrahamBenHarper: got it :)

packages/muelu/test/unit_tests/CMakeLists.txt
```diff
@@ -23,7 +23,6 @@ IF (${PACKAGE_NAME}_ENABLE_Experimental)
 ENDIF()
 
 APPEND_SET(SOURCES
-  Aggregates.cpp
```
cgcgcg, line 26: We still want to run those unit tests

GrahamBenHarper: Thank you!

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
GrahamBenHarper: I'm fairly certain this can go on host for now, seeing as all the parts of MueLu that called `IsRoot` from device were commented out anyway. We'll revisit whether that should realistically live on host/device and how the calls to it should be handled later.

GrahamBenHarper:
Re: conversations with @lucbv and @cgcgcg 

To avoid a performance regression for the old routes, we'll explicitly use a host node template argument (`using HostNO = Tpetra::KokkosCompat::KokkosSerialWrapperNode`) for `Aggregates` in factories that would otherwise take a performance hit. That is, objects in the old route will use `Aggregates<LO,GO,HostNO>`. This should be easy because we ETI based on Tpetra, as we believe we always have a serial node instantiation.

cgcgcg:
I don't think that we always have a serial node. We always have a host node though.

GrahamBenHarper:
Everything looks good on a no-UVM build on Weaver now. There were a lot of parts of the testsuite which constructed the test case data and then kept a pointer hanging onto the data while MueLu ran, which caused plenty of issues :sweat_smile:

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
```diff
@@ -95,16 +102,39 @@ namespace MueLu {
...
-           class GlobalOrdinal = DefaultGlobalOrdinal,
-           class Node = DefaultNode>
-  class Aggregates : public BaseClass {
+  template <class LocalOrdinal, class GlobalOrdinal, class DeviceType>
+  class Aggregates<LocalOrdinal, GlobalOrdinal, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<DeviceType> > : public BaseClass {
+  public:
+    // For some reason we seem intent on having these declared before pulling the short names in
+    // I am not sure why but I will keep things as it is for now
+    // If you need to define a type that depend on a short name please do it further down after
+    // the header has been included!
```
cgcgcg, line 114: Remove

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
```diff
@@ -114,85 +144,131 @@ namespace MueLu {
      */
     Aggregates(const GraphBase & graph);
 
+    /*! @brief Standard constructor for Aggregates structure
+     *
+     * Standard constructor of aggregates takes a Graph object as parameter.
+     * Uses the graph.GetImportMap() to initialize the internal vector for mapping nodes to (local) aggregate ids as well as
```
cgcgcg, line 150: LWGraph_kokkos, not Graph

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_def.hpp
```diff
@@ -300,12 +271,6 @@ namespace MueLu {
     return ComputeUnamalgamatedImportDofMap(aggregates.GetMap());
   }
 
-  template <class LocalOrdinal, class GlobalOrdinal, class Node>
-  RCP<Xpetra::Map<LocalOrdinal, GlobalOrdinal, Node> > AmalgamationInfo<LocalOrdinal, GlobalOrdinal, Node>::
-  ComputeUnamalgamatedImportDofMap(const Aggregates_kokkos& aggregates) const {
-    return ComputeUnamalgamatedImportDofMap(aggregates.GetMap());
-  }
-
```
cgcgcg, line 308: If we no longer need the two code paths, please remove the trivial methods.

GrahamBenHarper: This is the only public interface, and it's used in factories like TentativePFactory. Should I make the other interface public?

packages/muelu/test/unit_tests_kokkos/Aggregates_kokkos.cpp
```diff
@@ -237,7 +237,7 @@ namespace MueLuTests {
     return numDiscontiguousAggregates;
   }
 
-  TEUCHOS_UNIT_TEST_TEMPLATE_4_DECL(Aggregates_kokkos, JustUncoupledAggregationFactory, Scalar, LocalOrdinal, GlobalOrdinal, Node)
```
cgcgcg, line 240: Maybe don't rename the unit tests?

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
```diff
@@ -46,18 +46,25 @@
...
-#include "MueLu_ConfigDefs.hpp"
 #include "MueLu_BaseClass.hpp"
-#include "MueLu_Aggregates_fwd.hpp"
 
-#include "MueLu_Graph_fwd.hpp"
+#include "MueLu_LWGraph_kokkos.hpp"
+#include "MueLu_Graph.hpp"
 #include "MueLu_GraphBase.hpp"
 #include "MueLu_IndexManager.hpp"
+#include "MueLu_IndexManager_kokkos.hpp"
```
cgcgcg, line 67: Do IndexManager and IndexManager_kokkos do the same thing? Is the IndexManager still used? If not, remove include.

cgcgcg: If they don't do the same thing, how does one use IndexManager?

GrahamBenHarper: They don't, unfortunately. They're totally different objects because `IndexManager` is an abstract class with I believe three implemented children, and `IndexManager_kokkos` is not derived from it due to concerns about vtables on GPUs when used for the structured multigrid

packages/muelu/src/Misc/MueLu_FilteredAFactory_def.hpp
```diff
@@ -501,16 +499,27 @@ namespace MueLu {
 
     // Lists of nodes in each aggregate
     struct {
-      Array<LO> ptr,nodes, unaggregated;
+      // GH: For now, a lot of these routines (e.g. singleton deletion) seem unnecessary to run on device.
```
cgcgcg, line 502: I'd say that they haven't been ported to device, not that that is unnecessary.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
cgcgcg: So this effectively kills off functionality offered by IndexManager then?

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_def.hpp
cgcgcg: No, just rename.

packages/muelu/src/Misc/MueLu_FilteredAFactory_def.hpp
GrahamBenHarper: Well, I think I chose the word `unnecessary` simply because things like singleton deletion seem like low-intensity operations that would likely be less efficient if run on a device.

packages/muelu/src/Utils/MueLu_AggregationExportFactory_decl.hpp
```diff
@@ -166,13 +169,15 @@ namespace MueLu {
...
-    mutable Teuchos::ArrayRCP<const typename Teuchos::ScalarTraits<Scalar>::coordinateType> zCoords_;
-    mutable Teuchos::ArrayRCP<const typename Teuchos::ScalarTraits<Scalar>::coordinateType> cx_; //coarse local coordinates
-    mutable Teuchos::ArrayRCP<const typename Teuchos::ScalarTraits<Scalar>::coordinateType> cy_;
-    mutable Teuchos::ArrayRCP<const typename Teuchos::ScalarTraits<Scalar>::coordinateType> cz_;
+    // struct AggData {
+    //   Teuchos::ArrayRCP<LocalOrdinal> vertex2AggId_;
+    //   Teuchos::ArrayRCP<LocalOrdinal> aggSizes_;
+    //   std::vector<bool> isRoot_;
+    // };
+
```
cgcgcg, line 177: Remove?

packages/muelu/src/Utils/MueLu_AggregationExportFactory_def.hpp
```diff
@@ -418,7 +411,7 @@ namespace MueLu {
       {
         if(vertex2AggId_[j] == i)
         {
-          Point p(xCoords_[j], yCoords_[j]);
+          Point p(xCoords[j], yCoords[j]);
```
cgcgcg, line 414: Why were these changes needed?

GrahamBenHarper: Because `AggregationExportFactory` stored long-term pointers to `coordinates->getData()` as members of the class, which caused issues when it comes time to transfer coordinates to the next level in MueLu. I changed it to instead store RCPs for the MultiVectors and only grab the data pointers when needed.

packages/muelu/src/Misc/MueLu_FilteredAFactory_def.hpp
cgcgcg: And so it's a good idea to copy everything to host? I disagree.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
GrahamBenHarper: This PR in particular simply allows storage of an RCP to either `IndexManager` or `IndexManager_kokkos` since previously they corresponded to `Aggregates` and `Aggregates_kokkos`, respectively. It doesn't change any of the functionality offered by either `IndexManager` class, and I think it would have appeared in testing if it removed any functionality.

packages/muelu/src/Misc/MueLu_FilteredAFactory_def.hpp
GrahamBenHarper: Well, until we set `FilteredAFactory` to run device kernels, we have to copy to host. But I can see how that would change the interpretation of the comment.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
cgcgcg: Ok, that was my question above. If you know that they do the same thing you can remove the non-Kokkos one.

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_def.hpp
GrahamBenHarper: Can you say more? I'm confused about the suggestion.

cgcgcg (commented):
I guess we'll see what the AT thinks. This PR makes the `isRoot` member variable an `ArrayRCP`. This needs to be fixed down the line in order to be able to use roots in the future. If `IndexManager` and `IndexManager_kokkos` have indeed the same functionality the former can be removed as this PR effectively eliminates its use.

packages/muelu/src/Graph/Containers/MueLu_Aggregates_decl.hpp
GrahamBenHarper: They don't do the same thing unfortunately, and merging them will probably be a slightly messy step in #11658.
We'll have to reconcile the following:
```
class IndexManager : public BaseClass
class GlobalLexicographicIndexManager : public IndexManager<LocalOrdinal, GlobalOrdinal, Node>
class LocalLexicographicIndexManager : public IndexManager<LocalOrdinal, GlobalOrdinal, Node>
class UncoupledIndexManager : public IndexManager<LocalOrdinal, GlobalOrdinal, Node>
class IndexManager_kokkos : public BaseClass
```

packages/muelu/src/Graph/MatrixTransformation/MueLu_AmalgamationInfo_def.hpp
cgcgcg: `AmalgamationInfo` had methods with signatures
```
void UnamalgamateAggregates(const Aggregates& aggregates, Teuchos::ArrayRCP<LocalOrdinal>& aggStart, Teuchos::ArrayRCP<GlobalOrdinal>& aggToRowMap) const;
void UnamalgamateAggregatesLO(const Aggregates& aggregates, Teuchos::ArrayRCP<LocalOrdinal>& aggStart, Teuchos::ArrayRCP<LO>& aggToRowMap) const;
void UnamalgamateAggregates(const Aggregates_kokkos& aggregates, Teuchos::ArrayRCP<LocalOrdinal>& aggStart, Teuchos::ArrayRCP<GlobalOrdinal>& aggToRowMap) const;
void UnamalgamateAggregatesLO(const Aggregates_kokkos& aggregates, Teuchos::ArrayRCP<LocalOrdinal>& aggStart, Teuchos::ArrayRCP<LO>& aggToRowMap) const;
void UnamalgamateAggregates(const Teuchos::RCP<const Map> &nodeMap, const RCP<LOVector> &procWinnerVec, const RCP<LOMultiVector> &vertex2AggIdVec, const GO numAggregates, Teuchos::ArrayRCP<LocalOrdinal>& aggStart,Teuchos::ArrayRCP<GlobalOrdinal>& aggToRowMap) const;
void UnamalgamateAggregatesLO(const Teuchos::RCP<const Map> &nodeMap, const RCP<LOVector> &procWinnerVec, const RCP<LOMultiVector> &vertex2AggIdVec, const GO numAggregates, Teuchos::ArrayRCP<LocalOrdinal>& aggStart, Teuchos::ArrayRCP<LO>& aggToRowMap) const
```
All that's needed now is:
```
void UnamalgamateAggregates(const Aggregates& aggregates, Teuchos::ArrayRCP<LocalOrdinal>& aggStart, Teuchos::ArrayRCP<GlobalOrdinal>& aggToRowMap) const;
void UnamalgamateAggregatesLO(const Aggregates& aggregates, Teuchos::ArrayRCP<LocalOrdinal>& aggStart, Teuchos::ArrayRCP<LO>& aggToRowMap) const;
```

GrahamBenHarper:
A single test failure only in `MueLu_UnitTestsTpetra_MPI_4`, but this time it's openmpi...
```
    19. Aggregates_double_int_longlong_Tpetra_KokkosCompat_KokkosOpenMPWrapperNode_GreedyDirichlet_UnitTest ... 
    45. Aggregates_std_complex0double0_int_longlong_Tpetra_KokkosCompat_KokkosOpenMPWrapperNode_GreedyDirichlet_UnitTest ... 
```

GrahamBenHarper:
The aggregation test was fixed! Cuda build experienced a random failure: `ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2`

cgcgcg (approved):
I think this is fine. As far as I can tell, this does not introduce performance regressions. Using an ArrayRCP for `isRoots` is sub-optimal, but it's not a regression compared to the previous state because the Kokkos code path simply doesn't set root information. This also means that a lot of the plotting routines do not work with the Kokkos code path for constructing aggregates.

GrahamBenHarper:
That's right. I'll open a follow-on issue after this merges about working on aggregate root node capabilities for GPU since those were previously commented out. I think the only mildly annoying part is we don't have `MV<bool>` for syncing the `isRoot_` data between host and device, and we'll also probably want to refactor/rewrite the whole `IsRoot()` function.

@jhux2 said he wants to take a look at this before COB, so I'll wait to merge until I see his feedback as well. :)

packages/muelu/src/Misc/MueLu_CoordinatesTransferFactory_def.hpp
```diff
@@ -204,16 +203,10 @@ namespace MueLu {
...
       bool aggregatesCrossProcessors;
-      if (IsType<RCP<Aggregates> >(fineLevel, "Aggregates")) {
-        aggregates = Get< RCP<Aggregates> > (fineLevel, "Aggregates");
-        aggregatesCrossProcessors = aggregates->AggregatesCrossProcessors();
-      } else {
-        aggregatesKokkos = Get<RCP<Aggregates_kokkos> >(fineLevel, "Aggregates");
-        aggregatesCrossProcessors = aggregatesKokkos->AggregatesCrossProcessors();
-      }
+      aggregates = Get<RCP<Aggregates> >(fineLevel, "Aggregates");
+      aggregatesCrossProcessors = aggregates->AggregatesCrossProcessors();
```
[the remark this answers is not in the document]
cgcgcg: Unfortunately no. BrickAgg creates aggs that cross.

packages/muelu/src/Misc/MueLu_CoordinatesTransferFactory_def.hpp
```diff
@@ -225,66 +218,35 @@ namespace MueLu {
         ghostedCoords->doImport(*fineCoords, *importer, Xpetra::INSERT);
       }
 
-      if (!aggregates.is_null()) {
```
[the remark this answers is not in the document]
jhux2: @GrahamBenHarper explained -- there's only one aggregate construct now.

packages/muelu/src/Misc/MueLu_CoordinatesTransferFactory_def.hpp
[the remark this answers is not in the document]
jhux2: Oh fooey.

packages/muelu/src/Misc/MueLu_CoordinatesTransferFactory_def.hpp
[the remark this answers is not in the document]
GrahamBenHarper: Previously we had host vs device routes for whether we were using `Aggregates` or `Aggregates_kokkos`. We checked if we were on host by checking whether `aggregates` or `aggregatesKokkos` was null. Now we just have one route, and the data lives in the view, so we just delete the non-Kokkos route.

packages/muelu/src/Misc/MueLu_CoordinatesTransferFactory_def.hpp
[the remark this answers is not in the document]
GrahamBenHarper: Good point! The aggregates object probably doesn't need to store that information anymore either then.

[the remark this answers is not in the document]
GrahamBenHarper: oh the other conversation didn't load for me. good to know!

jhux2 (approved):
@GrahamBenHarper Lgtm, thank you!

PR #11735, opened 2023-03-27
Tpetra: execution space instances in MultiVector packAndPrepare

cwpearson:
@trilinos/tpetra
@trilinos/stokhos 

## Motivation
Needed before execution-space instance enabled import/export for SpMV communication-computation overlap: https://github.com/trilinos/Trilinos/pull/10926

### DIFF
--- a/packages/stokhos/src/sacado/kokkos/pce/tpetra/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels_UQ_PCE.hpp
+++ b/packages/stokhos/src/sacado/kokkos/pce/tpetra/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels_UQ_PCE.hpp
@@ -95,9 +95,10 @@ namespace Details {
     static void pack(const DstView& dst,
                      const SrcView& src,
                      const IdxView& idx,
-                     size_t col) {
+                     size_t col,
+                     const execution_space &space) {
       Kokkos::parallel_for(
-        Kokkos::MPVectorWorkConfig<execution_space>( idx.size(), BlockSize ),
+        Kokkos::MPVectorWorkConfig<execution_space>( space, idx.size(), BlockSize ),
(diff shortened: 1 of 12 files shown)


etphipp (dismissed):
Stokhos changes look good.

etphipp:
Why were the execution space overloads of `packAndPrepare`, etc... added as overloads instead of using a new function name?  Now anyone who implements their own DistObject will get warnings about partially overridden functions since those overloads won't be visible in their implementation, unless they overload the execution space versions as well, or bring them into scope with a using statement.

cwpearson:
Nothing beyond two functions that do the same thing having the same name. I take your point about ergonomics - should I understand you'd prefer a different name rather than an overload? I will bring this up in the next Tpetra meeting.

etphipp:
I just mention it because it is a common source of warnings from compilers, because it is a common source of errors among code developers.  If someone implements their own DistObject (as I have in another code), and doesn't reimplement both overloaded versions, they will get a compiler warning as the unimplemented overload is hidden in the derived class and cannot be called.  For that reason, a common nugget of wisdom is to not overload virtual functions in the base class.

cwpearson:
Additional resources:

ctrl-F "-Woverloaded-virtual" here: https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html

Small example:
```
class Base {
public:
    virtual void overloaded(int) {};
    virtual void overloaded(int, float) {};

    virtual void normal(int) {};
};

class Derived: public Base {
public:
    void overloaded(int a,float b); // warning: 'virtual void Base::overloaded(int)' was hidden [-Woverloaded-virtual]
};

int main(void) {

    Derived d;

    d.normal(3); // works
    // d.overloaded(3) // doesn't work even though in Base (hidden)
    d.overloaded(3, 3.0f); // works
}
```

Despite `void overloaded(int)` being present in the base class, you can't call it from the derived class because it's hidden by the other declaration in the derived class.

PR #11749, opened 2023-03-29
MueLu: Set "aggregation: match ML phase2a" in ML2MueLu translation

cgcgcg:
@trilinos/muelu 

## Motivation
This seems reasonable, but I'll wait for @rstumin's feedback.

### DIFF
--- a/packages/muelu/src/Interface/MueLu_ML2MueLuParameterTranslator.cpp
+++ b/packages/muelu/src/Interface/MueLu_ML2MueLuParameterTranslator.cpp
@@ -396,6 +396,9 @@ namespace MueLu {
         adaptingParamList.remove(pname,false);
       }
 
+      // make sure that MueLu's phase2a matches ML's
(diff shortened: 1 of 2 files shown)


jhux2:
@rstumin @cgcgcg This does exactly what the name implies?

cgcgcg:
If I can trust my past self, then yes.

cgcgcg:
The option was added/modified in PR #9747.

I wrote up the story behind it in this comment:
https://github.com/trilinos/Trilinos/pull/9747#issuecomment-994086900

rstumin (approved):
looks good.

jwillenbring:
@cgcgcg @trilinos/framework Yes, this does seem odd. The AT only ran the one time, and the testing was stale, and was properly marked as stale, yet the merge was allowed to happen.

It almost seems like the AT would not automatically merge, but for some reason it did allow a manual merge. Not sure if there is some glitch with the timing of the review?

Issue #11756, opened 2023-04-03
Tpetra: External Kokkos Support


csiefer2:
We need to do this in a way that minimized the chance of user mistakes.

PR #11757, opened 2023-04-03
Tpetra: Adding version checking for external Kokkos

csiefer2:
Issue: #11756
Tests: Tested all use cases (internal, external ok, external not matching, external not matching with override)
Stakeholder feedback: See below

The logic here is that Tpetra is a sensible place to check Kokkos compatibility with Trilinos, since the vast majority of Trilinos users that use Kokkos also use Tpetra.  If there's a strong demand to put this somewhere at the top level, we can look into that, but that might be overkill.

### DIFF
--- a/packages/tpetra/CMakeLists.txt
+++ b/packages/tpetra/CMakeLists.txt
@@ -30,13 +30,13 @@ SET(Tpetra_SUPPORTED_KOKKOS_VERSION "4.0.0")
 # Option to allow developers to ignore incompatible Kokkos versions
 TRIBITS_ADD_OPTION_AND_DEFINE(
   Tpetra_IGNORE_KOKKOS_COMPATIBILITY
-  TPETRA_IGNORE_KOKKOS_COMPATIBILITY
+  ""
   "If true, disables the Kokkos version compatibility checking in Tpetra"
   OFF
 )
 # Kokkos version checking
 ASSERT_DEFINED (Tpetra_IGNORE_KOKKOS_COMPATIBILITY)
-IF(DEFINED TPL_ENABLE_Kokkos AND TPL_ENABLE_Kokkos)
+IF(TPL_ENABLE_Kokkos)
   # If we're using external Kokkos, we check the supported version, unless the cmake option says otherwise
   IF(DEFINED Kokkos_VERSION)
     IF(Kokkos_VERSION VERSION_EQUAL Tpetra_SUPPORTED_KOKKOS_VERSION)


packages/tpetra/CMakeLists.txt
```diff
@@ -21,6 +21,38 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
...
+# Supported Kokkos version in Trilinos
+# NOTE: When we snapshot Kokkos into Trilinos, we have to update these numbers to maintain
+# compatibility with external Kokkos
+SET(Tpetra_SUPPORTED_KOKKOS_VERSION "4.0.0")
+
+
+# Option to allow developers to ignore incompatible Kokkos versions
+TRIBITS_ADD_OPTION_AND_DEFINE(
+  Tpetra_IGNORE_KOKKOS_COMPATIBILITY
+  TPETRA_IGNORE_KOKKOS_COMPATIBILITY
```
bartlettroscoe, line 33: What is the variable `TPETRA_IGNORE_KOKKOS_COMPATIBILITY` used for?  That second argument is typically used for a macro define in a `<Package>_Config.h.in` file that gets configured.   (See the documentation for [`tribits_add_option_and_define()`](https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-add-option-and-define).)

In this case, I think you just want:

```
set(TPETRA_IGNORE_KOKKOS_COMPATIBILITY OFF CACHE BOOL
  "If true, disables the Kokkos version compatibility checking in Tpetra")
```

or:

```
  Tpetra_IGNORE_KOKKOS_COMPATIBILITY
  ""
  "If true, disables the Kokkos version compatibility checking in Tpetra"
  OFF
)
```

if you want to export this variable `Tpetra_IGNORE_KOKKOS_COMPATIBILITY` to the `<Pacakge>Config.cmake` file (See [`tribits_pkg_export_cache_var()`](https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-pkg-export-cache-var).)

packages/tpetra/CMakeLists.txt
```diff
@@ -21,6 +21,38 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
   ${Tpetra_ENABLE_DEPRECATED_CODE_DEFAULT}
 )
 
+# Supported Kokkos version in Trilinos
+# NOTE: When we snapshot Kokkos into Trilinos, we have to update these numbers to maintain
+# compatibility with external Kokkos
+SET(Tpetra_SUPPORTED_KOKKOS_VERSION "4.0.0")
```
bartlettroscoe, line 27: 👍

packages/tpetra/CMakeLists.txt
```diff
@@ -21,6 +21,38 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
...
+  TPETRA_IGNORE_KOKKOS_COMPATIBILITY
+  "If true, disables the Kokkos version compatibility checking in Tpetra"
+  OFF
+)
+# Kokkos version checking
+ASSERT_DEFINED (Tpetra_IGNORE_KOKKOS_COMPATIBILITY)
+IF(DEFINED TPL_ENABLE_Kokkos AND TPL_ENABLE_Kokkos)
+  # If we're using external Kokkos, we check the supported version, unless the cmake option says otherwise
+  IF(DEFINED Kokkos_VERSION)
+    IF(Kokkos_VERSION VERSION_EQUAL Tpetra_SUPPORTED_KOKKOS_VERSION)
```
bartlettroscoe, line 42: 👍

NOTE: In this case, I think you do need to have the wrapping `IF(DEFINED Kokkos_VERSION)` statement because otherwise an undefined or empty.

packages/tpetra/CMakeLists.txt
```diff
@@ -21,6 +21,38 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
...
+      MESSAGE(STATUS "Tpetra: Detected compatible external Kokkos version: ${Kokkos_VERSION}")
+    ELSEIF(Tpetra_IGNORE_KOKKOS_COMPATIBILITY)
+      MESSAGE(WARNING "Tpetra: Ignoring Kokkos compatibility between external Kokkos version (${Kokkos_VERSION}) and Trilinos supported version (${Tpetra_SUPPORTED_KOKKOS_VERSION}) at user request.  This may lead to problems in Trilinos.")
```
bartlettroscoe, line 45: 👍

I think a `message(WARNING ...)` is warranted in this case.

bartlettroscoe (changes requested):
@csiefer2, looks great.  The only thing that needs to be changed is to remove the definition of the unused var `TPETRA_IGNORE_KOKKOS_COMPATIBILITY` as described below.

packages/tpetra/CMakeLists.txt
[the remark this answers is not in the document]
bartlettroscoe: I think the snapshotting process could be updated to provide this info.  But the Kokkos subdirectory would need to be storing this info somehow and provide this info to downstream CMake customers.  When building from source in the same CMake project (e.g. using `add_subdirectory()`), the top-level Kokkos CMakeLists.txt file could define a cache variable `Kokkos_VERSION` (e.g. reading a file `KokkosVersion.cmake` that gets updated in the snapshotting process).  And  `Kokkos_VERSION` would need to be written into the `KokkosConfig.cmake` file as well, no matter how Kokkos was built (stand-alone or under TriBITS).

bartlettroscoe (approved):
@csiefer2, thanks for update.  This looks good to merge.

packages/tpetra/CMakeLists.txt
```diff
@@ -21,6 +21,38 @@ TRIBITS_ADD_OPTION_AND_DEFINE(
   ${Tpetra_ENABLE_DEPRECATED_CODE_DEFAULT}
 )
 
+# Supported Kokkos version in Trilinos
+# NOTE: When we snapshot Kokkos into Trilinos, we have to update these numbers to maintain
```
ndellingwood, line 25: Making a note here that we'll need to add a step during Kokkos releases to update the Kokkos version checked by Tpetra in the lines below, @crtrott @dalg24 @lucbv

PR #11761, opened 2023-04-04
Stokhos: Moves KokkosContainers from an optional to a required dependency

hkthorn:
The Kokkos_StaticCrsGraph.hpp file is included in several files in stokhos/src/kokkos, which is in the KokkosContainers subpackage. That is an optional dependency in the Dependencies.cmake file, so those headers are not installed if optional dependencies are not included.  The result is a build failure in Xyce because, when Stokhos is enabled, that header is not found.  By moving KokkosContainers to a required dependency, this build failure goes away.

@trilinos/<teamName>


hkthorn:
This is a cherry-pick of PR #11685 onto the Trilinos 14.0 branch, per @sebrowne instructions.

sebrowne (approved):

PR #11762, opened 2023-04-04
Intrepid2 :: Projection doesn't always need orientations

reuterb:
In my testing of the new wrappers introduced in #11698 I found that if `orts` is an empty array, the projection operation thinks `ncells = 0` and nothing occurs.
I'm proposing this simple fix which should allow for an empty `orts` array.
The use case is for nodal bases where one may not need to compute the orientations.

@trilinos/<teamName>

### DIFF
--- a/packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
+++ b/packages/intrepid2/src/Projection/Intrepid2_ProjectionTools.hpp
@@ -630,7 +630,7 @@ public:
     projStruct.createL2ProjectionStruct(dstBasis, srcBasis->getDegree());
 
     ordinal_type numPoints = projStruct.getNumTargetEvalPoints();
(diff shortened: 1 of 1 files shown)


mperego:
@reuterb I think that the projections don't work without orientations. While it's true that for some basis orientations are not needed, I think projections in general require them, so this fix won't solve the problem. I would not pass an empty array for orientations. If anything I would pass an array with zeros (0 means that the cell is oriented as the reference cell).

reuterb:
> @reuterb I think that the projections don't work without orientations. While it's true that for some basis orientations are not needed, I think projections in general require them, so this fix won't solve the problem. I would not pass an empty array for orientations. If anything I would pass an array with zeros (0 means that the cell is oriented as the reference cell).

OK, can do. Thanks.

PR #11774, opened 2023-04-07
MueLu: Guard repartition option in hierarchy write unit test by HAVE_MUELU_ZOLTAN2

### LINKED ISSUE #11773: MueLu: Guard hierarchy write unit test by HAVE_MUELU_ZOLTAN2
@trilinos/muelu 

If MueLu is built without any form of Zoltan support, the Hierarchy.cpp test on line 1546 will fail in the following way:
```
105. Hierarchy_double_int_longlong_Tpetra_KokkosCompat_KokkosOpenMPWrapperNode_Write_UnitTest ...
 ******* WARNING *******
 Zoltan2 interface is not available, trying to switch to Zoltan
 p=0: *** Caught standard std::exception of type 'MueLu::Exceptions::RuntimeError' :
  Zoltan interface is not available
 [FAILED]  (0.000484 sec) Hierarchy_double_int_longlong_Tpetra_KokkosCompat_KokkosOpenMPWrapperNode_Write_UnitTest
 Location: packages/muelu/test/unit_tests/Hierarchy.cpp:1546
```

I assume we could wrap the repartition settings on line 1570 by `#if defined (HAVE_MPI) && defined(HAVE_MUELU_ZOLTAN2)` instead of just `#ifdef HAVE_MPI`.

GrahamBenHarper:
Closes #11773.

@trilinos/muelu 
I have no clue if this will even be noticed by testing since I assume testing enables the majority of packages, but I think this is the best way to fix the test on builds without Zoltan2.

### DIFF
--- a/packages/muelu/test/unit_tests/Hierarchy.cpp
+++ b/packages/muelu/test/unit_tests/Hierarchy.cpp
@@ -1567,7 +1567,7 @@ namespace MueLuTests {
     RCP<RealValuedMultiVector> coordinates = Galeri::Xpetra::Utils::CreateCartesianCoordinates<SC,LO,GO,Map,RealValuedMultiVector>("1D", A->getRowMap(), galeriList);
 
     Teuchos::ParameterList paramList;
-#ifdef HAVE_MPI
+#if defined(HAVE_MPI) && defined(HAVE_MUELU_ZOLTAN2)
     paramList.set("repartition: enable", true);
     paramList.set("repartition: start level", 1);
     paramList.set("repartition: min rows per proc", 6);


mayrmt (approved):

cgcgcg:
I don't think that this is the correct fix. It's perfectly fine to stick rebalancing params into the PL, even when there is no Zoltan2. It's the job of the ParameterListInterpreter to realize that there is no rebalancing option available and ignore the rebalancing params. So that's were a fix needs to go.

cgcgcg:
Also, what happens when MueLu has Zoltan but not Zoltan2? I think this change should be reverted.

GrahamBenHarper:
Yeah, I can look into a different option. Sorry, I didn't think that would go through so quickly.

PR #11775, opened 2023-04-07
MueLu: Remove CGAL as a TPL

### LINKED ISSUE #11765: MueLu: Remove CGAL TPL
@trilinos/muelu 

While working on some of the visualization tools during our Kokkos merge, I realized that while we technically list CGAL as a TPL, there are large chunks of CGAL code surrounded by `#if 0`. I asked around, and we don't even test against CGAL anymore, so it seems appropriate to simply remove it.

GrahamBenHarper:
Closes #11765

@trilinos/muelu

I think it's appropriate to remove CGAL as a TPL from MueLu (and therefore, Trilinos), as large portions of our CGAL capabilities were guarded by `#if 0`, and the parts that were not macro'd out did not seem like they were tested. Additionally, we don't have any testing builds with CGAL enabled. I think this is the right thing to do unless somebody is willing to test/maintain/update those capabilities.


cgcgcg:
What's the downside of keeping CGAL? It seems that there is some code path in `MueLu::VisualizationHelpers` that uses it?

GrahamBenHarper:
Random test failure: `ROL_example_PDE-OPT_poisson_example_01_MPI_4`

The downside of keeping it is the "Alpha Hulls" option in `AggregationExportFactory` and `CoarseningVisualizationFactory` are the only things that require it, and the methods in `AggregationExportFactory` have been commented out since June 2016 with comments like `#if 0 // taw: does not compile with CGAL 4.8`. The VisualizationHelpers methods only exist to assist those particular factories, and it's not clear if the "CGAL Convex Hulls" in `CoarseningVisualizationFactory` do anything different from "Convex Hulls".

I'm not going to pretend that I completely understand CGAL and what it was used for, but the way it sits right now gives me the impression it hasn't been tested in many years, probably doesn't compile properly, and would take some effort to get working again. Keeping it in MueLu means we have some strange inbetween where we aren't really testing or supporting it, but we want people to use it. If somebody were insistent on keeping CGAL, it seems like most of the required information would be related to [alpha shapes](https://doc.cgal.org/latest/Alpha_shapes_3/index.html), but I think it would be better to look for Kokkos-based TPLs in the long term.

mayrmt:
The feature of visualizing aggregates is surely important. It helps in presenting results in publications, but can also be a useful debugging tool. As far as I know, it is currently quite tricky to use and fails at least in some cases, that our group is interested in.

I don't know, how CGAL is exactly related to that and if we could generate nice aggregate visualization without it. So, I cannot contribute to the decision. Just want to raise the point, that aggregate visualization is/would be a useful tool.

GrahamBenHarper:
@mayrmt I agree that aggregate visualization is important. We still have options for "convex hulls", so this doesn't remove much in terms of aggregate visualization. It simply removes the "alpha hulls" and "CGAL convex hulls", which seemed like they barely functioned and were no longer tested.

brian-kelley:
I have a really old branch where I was trying to improve agg visualization, especially performance and robustness of the 2D/3D convex hull. Should I try to clean that up and add it?

GrahamBenHarper:
@brian-kelley I remember there were some issues with robustness for aggregate visualization, but I don't remember what the main problems were... something about the convex hulls not being quite right?

Just as a heads-up, I did a little bit of refactoring to `AggregationExportFactory` in #11726 to force it to stop holding onto long-term `ArrayRCP`s of coordinate data since it causes issues when exporting aggregates from Cuda builds. I don't know how that would affect your branch.

brian-kelley:
@GrahamBenHarper IIRC the current impl has trouble when a hull face contains points other than its vertices, because whether those points are "inside" or "outside" the hull comes down to floating-point error. That would be fixed in this newer impl.

The reasons we added CGAL was to get a better and faster convex hull impl, as well as to support alpha hulls as a feature. But I don't know of anyone using the alpha hull option.

GrahamBenHarper:
@brian-kelley oh, that would be great! I think with how little CGAL was used and because the use case was so specific, it would make more sense to simply improve our existing code without relying on a TPL like you've done. I'm definitely interested if it's not too much trouble for you :)

packages/muelu/src/Utils/MueLu_AggregationExportFactory_def.hpp
```diff
@@ -315,15 +306,6 @@ namespace MueLu {
...
-        #ifdef HAVE_MUELU_CGAL
-        doAlphaHulls_(vertices, geomSizes);
-        #else
-        GetOStream(Warnings0) << "   Trilinos was not configured with CGAL so Alpha Hulls not available.\n   Using Convex Hulls instead." << std::endl;
-        doConvexHulls(vertices, geomSizes);
-        #endif
-      }
       else
       {
         GetOStream(Warnings0) << "   Unrecognized agg style.\n   Possible values are Point Cloud, Jacks, Jacks++, Convex Hulls and Alpha Hulls.\n   Defaulting to Point Cloud." << std::endl;
```
cgcgcg, line 311: "Alpha Hulls" is not longer valid, right?

GrahamBenHarper: Good catch!

cgcgcg (approved):

PR #11776, opened 2023-04-07
ML: Fix checking Windows platform for non-msvc toolchain

Biswa96:
This changes _MSC_VER with _WIN32 macro to check Windows platform.
    _MSC_VER is specific for MSVC toolchain and not defined in mingw
    based toolchain. So, this fixes the following compiler error with
    mingw clang.

    Trilinos/packages/ml/src/Utils/ml_utils.c:1965:9: error: call to undeclared function 'gethostname'
    Trilinos/packages/ml/src/Utils/ml_utils.c:1966:68: error: call to undeclared function 'getpid'
    Trilinos/packages/ml/src/Utils/ml_utils.c:1973:9: error: call to undeclared function 'sleep'

    Also remove the MSVC specific pragma directive. ws2_32 library is
    already mentioned in corresponding CMakeLists.txt file.

@trilinos/<teamName>

## Motivation

This fixes compiling with clang 16 in mingw toolchain.

### DIFF
--- a/packages/ml/src/Utils/ml_utils.c
+++ b/packages/ml/src/Utils/ml_utils.c
@@ -23,8 +23,7 @@ Modified by:      $Author$
 #include "ml_lapack.h"
 #include <time.h>
 #include "ml_viz_stats.h"
-#ifdef _MSC_VER
-#pragma comment(lib, "Ws2_32.lib")
+#ifdef _WIN32
 # include <Winsock2.h>
 # include <process.h>
 void sleep(int sec)


cgcgcg:
@rstumin @jhux2 @csiefer2 ?

jhux2 (approved):
We don't test or formally support building on Windows.  Presumably you've checked that this works.

Biswa96:
> We don't test or formally support building on Windows. Presumably you've checked that this works.

The macros in this pull request are explained in this official documentation https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros

Here is an analogy. Using `_MSC_VER` to check Windows is as wrong as using `__GNUC__` to check Linux.

Thank you.

PR #11777, opened 2023-04-07
MueLu: Update ParameterListInterpreter to disable repartitioning if Zoltan/Zoltan2 are unavailable

GrahamBenHarper:
@trilinos/muelu 

This is a more appropriate resolution for #11773/#11774. The `#ifdef` logic in `ParameterListInterpreter` was a little messy, as evidenced by an `#ifdef HAVE_MPI` nested inside an `#ifdef HAVE_MPI`, which I also removed. Now the ParameterListInterpreter will turn repartitioning off and print an appropriate warning if the user does not have both MPI and Zoltan/Zoltan2.

Therefore, I don't think I need to guard repartitioning in the hierarchy test with `HAVE_MPI` or `HAVE_MUELU_ZOLTAN2` since that logic now falls on the `ParameterListInterpreter`.

Previous behavior: If the user requests repartitioning and does not have MPI, it will print a warning and turn repartitioning off. If the user requests repartitioning and has MPI enabled but does not have Zoltan2 enabled, it will attempt to switch to Zoltan and then throw and crash if Zoltan isn't available either. 
New behavior: If the user requests repartitioning and has MPI enabled but does not have Zoltan or Zoltan2 enabled, it will print a warning about Zoltan/Zoltan2 and disable repartitioning, consistent with the current behavior when MPI is unavailable.

### DIFF
--- a/packages/muelu/src/Interface/MueLu_ParameterListInterpreter_def.hpp
+++ b/packages/muelu/src/Interface/MueLu_ParameterListInterpreter_def.hpp
@@ -1576,7 +1576,7 @@ namespace MueLu {
     MUELU_SET_VAR_2LIST(paramList, defaultList, "reuse: type", std::string, reuseType);
     MUELU_SET_VAR_2LIST(paramList, defaultList, "repartition: enable", bool, enableRepart);
     if (enableRepart) {
-#ifdef HAVE_MPI
+#if defined(HAVE_MPI) && (defined(HAVE_MUELU_ZOLTAN) || defined(HAVE_MUELU_ZOLTAN2)) // skip to the end, print warning, and turn off repartitioning if we don't have MPI and Zoltan/Zoltan2
       MUELU_SET_VAR_2LIST(paramList, defaultList, "repartition: use subcommunicators in place", bool, enableInPlace);
       // Short summary of the issue: RebalanceTransferFactory shares ownership
       // of "P" with SaPFactory, and therefore, changes the stored version.
@@ -1623,24 +1623,25 @@ namespace MueLu {
       TEUCHOS_TEST_FOR_EXCEPTION(partName != "zoltan" && partName != "zoltan2", Exceptions::InvalidArgument,
                                  "Invalid partitioner name: \"" << partName << "\". Valid options: \"zoltan\", \"zoltan2\"");
 
-#ifndef HAVE_MUELU_ZOLTAN
+# ifndef HAVE_MUELU_ZOLTAN
       bool switched = false;
       if (partName == "zoltan") {
         this->GetOStream(Warnings0) << "Zoltan interface is not available, trying to switch to Zoltan2" << std::endl;
         partName = "zoltan2";
         switched = true;
       }
-#else
-# ifndef HAVE_MUELU_ZOLTAN2
+# else
+#  ifndef HAVE_MUELU_ZOLTAN2
       bool switched = false;
-# endif
-#endif
-#ifndef HAVE_MUELU_ZOLTAN2
+#  endif // HAVE_MUELU_ZOLTAN2
+# endif // HAVE_MUELU_ZOLTAN
+
+# ifndef HAVE_MUELU_ZOLTAN2
       if (partName == "zoltan2" && !switched) {
         this->GetOStream(Warnings0) << "Zoltan2 interface is not available, trying to switch to Zoltan" << std::endl;
         partName = "zoltan";
       }
-#endif
+# endif // HAVE_MUELU_ZOLTAN2
 
       MUELU_SET_VAR_2LIST(paramList, defaultList, "repartition: node repartition level",int,nodeRepartitionLevel);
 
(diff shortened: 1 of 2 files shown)


packages/muelu/test/unit_tests/Hierarchy.cpp
```diff
@@ -1567,11 +1567,9 @@ namespace MueLuTests {
     RCP<RealValuedMultiVector> coordinates = Galeri::Xpetra::Utils::CreateCartesianCoordinates<SC,LO,GO,Map,RealValuedMultiVector>("1D", A->getRowMap(), galeriList);
 
     Teuchos::ParameterList paramList;
-#if defined(HAVE_MPI) && defined(HAVE_MUELU_ZOLTAN2)
```
cgcgcg, line 1570: Just double-checking: this used to be `#ifdef HAVE_MPI`. Doesn't this throw now when MPI=OFF?

packages/muelu/src/Interface/MueLu_ParameterListInterpreter_def.hpp
```diff
@@ -1623,24 +1623,24 @@ namespace MueLu {
       TEUCHOS_TEST_FOR_EXCEPTION(partName != "zoltan" && partName != "zoltan2", Exceptions::InvalidArgument,
                                  "Invalid partitioner name: \"" << partName << "\". Valid options: \"zoltan\", \"zoltan2\"");
 
-#ifndef HAVE_MUELU_ZOLTAN
```
cgcgcg, line 1626: Why the change of logic?

packages/muelu/test/unit_tests/Hierarchy.cpp
GrahamBenHarper: I double-checked, and currently we don't actually throw if MPI=OFF. We simply turn repartition off and print a warning.
```
#ifdef HAVE_MPI
...
#else
      paramList.set("repartition: enable",false);
      this->GetOStream(Warnings0) << "No repartitioning available for a serial run\n";
#endif
```

packages/muelu/src/Interface/MueLu_ParameterListInterpreter_def.hpp
GrahamBenHarper: whoops, good catch. I started typing out `#if !defined(HAVE_MUELU_ZOLTAN)`, and then backspaced it when I decided I didn't want to change all of them to `#if defined`, but replaced it incorrectly.

cgcgcg (dismissed):

cgcgcg (approved):

GrahamBenHarper:
The gold files are reporting that repartitioning was turned off. There must be some sort of logic error in there still.

GrahamBenHarper:
hah, I just realized I have that same style of typo `MUELU_HAVE_ZOLTAN` that I noticed in the other PR last week... that would explain things

jhux2:
> hah, I just realized I have that same style of typo `MUELU_HAVE_ZOLTAN` that I noticed in the other PR last week... that would explain things

Like this?

```
src/Operators/MueLu_RefMaxwell_def.hpp:#ifdef MUELU_HAVE_TPETRA
src/Graph/Containers/MueLu_Zoltan2GraphAdapter.hpp:#endif// MUELU_HAVE_ZOLTAN2
src/Smoothers/MueLu_BelosSmoother_def.hpp:#ifdef MUELU_HAVE_TPETRA
src/Smoothers/MueLu_BelosSmoother_def.hpp:#ifdef MUELU_HAVE_TPETRA
src/Smoothers/MueLu_BelosSmoother_def.hpp:#ifdef MUELU_HAVE_TPETRA
test/unit_tests/Repartition.cpp:#   if !defined(MUELU_HAVE_AMESOS) || !defined(MUELU_HAVE_IFPACK)
test/unit_tests/Repartition.cpp:#   if !defined(MUELU_HAVE_AMESOS2) || !defined(MUELU_HAVE_IFPACK2)
test/unit_tests/Repartition.cpp:#   if !defined(MUELU_HAVE_AMESOS) || !defined(MUELU_HAVE_IFPACK)
test/unit_tests/Repartition.cpp:#   if !defined(MUELU_HAVE_AMESOS2) || !defined(MUELU_HAVE_IFPACK2)
test/unit_tests/TentativePFactory.cpp:#   if !defined(MUELU_HAVE_IFPACK)
test/unit_tests/TentativePFactory.cpp:#   if !defined(MUELU_HAVE_IFPACK2)
test/unit_tests/GenericRFactory.cpp:#   if !defined(MUELU_HAVE_IFPACK)
test/unit_tests/GenericRFactory.cpp:#   if !defined(MUELU_HAVE_IFPACK2)
test/unit_tests/GenericRFactory.cpp:#   if !defined(MUELU_HAVE_IFPACK)
test/unit_tests/GenericRFactory.cpp:#   if !defined(MUELU_HAVE_IFPACK2)
```

PR #11546, opened 2023-02-07
Tweaks to Trilinos to get working with pre-installed Kokkos and KokkosKernels (with TriBITS updates) (#11545)

bartlettroscoe:
## Description

This PR contains changes to TriBITS and Trilinos for:

* #11545

This brings in the TriBITS changes as of the merge of the TriBITS PR:

* https://github.com/TriBITSPub/TriBITS/pull/560

and therefore, also includes the TriBITS-only PR:

* #11746

FYI: The matching PR without the TriBITS changes is:

* #11747

and that one may be more attractive to review file-by-file.  (But this PR #11546 can be reviewed commit by commit just fine).

## Testing

See comments below.

## Tasks

* [x] Run some manual builds/installs for building and installing Kokkos and KokkosKernels and building downstream packages based on these pre-installed packages
  * [x] Get a non-CUDA build configuration working (see [below](https://github.com/trilinos/Trilinos/pull/11546#issuecomment-1433410117))
  * [x] Get a CUDA build configuration working (see [below](https://github.com/trilinos/Trilinos/pull/11546#issuecomment-1437311239))
* [x] Wait for TriBITS PR https://github.com/TriBITSPub/TriBITS/pull/560 to be completed and merged
* [x] Do another round of testing (see [below](https://github.com/trilinos/Trilinos/pull/11546#issuecomment-1489238255))
* [x] Demo and present to Trilinos and Kokkos developers ... Meeting was had on 3/29/2023 and was recorded.
* [x] Get TriBITS changes in PR #11746 to merge first?
* [ ] Break off commits for Kokkos and apply to Kokkos GitHub repo

PR #11779, opened 2023-04-10
Kokkos: Avoid using TriBITS in Kokkos

masterleinad:
@trilinos/kokkos @bartlettroscoe 

## Motivation
Currently, we are essentially using three build systems in Kokkos:
- Makefile-based (only supported for using Kokkos inline, not standalone)
- raw CMake
- through Trilinos using TriBITS

We discussed that we want to try to lighten that burden by removing a full TriBITS option by only providing only necessary additions to the raw CMake build system. This pull request explores doing this.   

## Related Issues
* Related to https://github.com/TriBITSPub/TriBITS/issues/562 and https://github.com/trilinos/Trilinos/pull/11546
 
## Testing
Compiling Trilinos locally

## Additional Information
~~This is not fully working yet. I can compile `Trilinos` but we don't export `Kokkos` in a way usable for exporting other libraries depending on `Kokkos`.~~


bartlettroscoe:
@masterleinad, what about defining and running tests in a Trilinos build with internal Kokkos?  Would that be supported?

I will submit a sister PR to remove the usage of subpackages from the Kokkos TriBITS build and the rest of the Trilinos packages at the end of this week.

masterleinad:
> @masterleinad, what about defining and running tests in a Trilinos build with internal Kokkos? Would that be supported?

We can discuss if it's acceptable to temporarily disable all `Kokkos` tests in `Trilinos`. My preference would be to reenable them in a later pull request, though.

crtrott:
Agreed I think we start with just not having tests for Kokkos itself enabled in Trilinos and then go from there.

bartlettroscoe:
> Agreed I think we start with just not having tests for Kokkos itself enabled in Trilinos and then go from there.

From what I have seen a part of #11808 and https://github.com/kokkos/kokkos/pull/6104 I don't see why you would have to disable Kokkos tests with the native Kokkos CMake build system when building Kokkos under Trilinos. It already responds to the options `Kokkos_ENABLE_TESTS` and `Kokkos_ENABLE_EXAMPLES` which are standard TriBITS options.

The more challenging part will be generating a KokkosConfig.cmake file for the build dir and the install dir and such issues.  (And that has to be done or we break Trilinos for customers currently getting Trilinos from `find_package(Trilinos)`.)

masterleinad:
> The more challenging part will be generating a KokkosConfig.cmake file for the build dir and the install dir and such issues.

See https://github.com/kokkos/kokkos/pull/6078.

cmake/tribits/core/package_arch/TribitsPackageMacros.cmake
```diff
@@ -867,7 +867,7 @@ macro(tribits_package_postprocess)
   tribits_package_finalize_dependency_vars()
   tribits_package_postprocess_common()
 
-  if (${PACKAGE_NAME}_SOURCE_DIR STREQUAL ${PROJECT_NAME}_SOURCE_DIR)
+  if (${PACKAGE_NAME}_SOURCE_DIR STREQUAL ${CMAKE_PROJECT_NAME}_SOURCE_DIR)
```
masterleinad, line 870: @bartlettroscoe What do you think about this change? This allows using `project` in `Kokkos`.

bartlettroscoe: Yes, the change from `PROJECT_NAME` to `CMAKE_PROJECT_NAME` needs to be made throughout TriBITS to address this issue and allow calls to `project()` from individual packages.

In addition, TriBITS will need to change `${PROJECT_SOURCE_DIR}` and `${PROJECT_BINARY_DIR}` to  `${${CMAKE_PROJECT_NAME}_PROJECT_SOURCE_DIR}` and `${{${CMAKE_PROJECT_NAME}_PROJECT_BINARY_DIR}`, respectively, almost everywhere.

Would you mind creating a TriBITS GitHUb issue for this and point out the places where you had to make this change?

cmake/tribits/core/package_arch/TribitsGlobalMacros.cmake
```diff
@@ -2177,7 +2177,7 @@ macro(tribits_configure_enabled_packages)
         endif()
         if (NOT ${PACKAGE_NAME}_TRIBITS_PACKAGE_POSTPROCESS)
           tribits_report_invalid_tribits_usage(
```
bartlettroscoe, line 2179: I think I will be able to adjust the logic to avoid this check if the target `${PACKAGE_NAME}::all_libs` is already defined.

I will be creating a PR that will duplicate this Kokkos using case (i.e. using a raw CMake build system for a TriBITS-compliant internal package that does not call any TrBITS functions).  This is one of the issues I will hit with that.

masterleinad: Yeah, that would be good. I think we would rather define the target then calling `tribits_package_postprocess()`.

cmake/tribits/core/package_arch/TribitsPackageMacros.cmake
masterleinad: See https://github.com/TriBITSPub/TriBITS/issues/582.

bartlettroscoe:
CC: @jwillenbring, @sebrowne, @rppawlo, @ccober6, @jhux2 

FYI: If the native Kokkos CMake build system does not use `tribits_add_test()` (or have the equivalent behavior), then this this will break the Trilinos PR builds because of the need to be able surgically disable tests (and even individual GTest unit tests).  For example, see https://github.com/trilinos/Trilinos/issues/11940#issuecomment-1572907274.  (I.e. there is more to managing tests than just calling `add_test()`.)

So perhaps it is best to just disable Kokkos tests when building under Trilinos for the foreseeable future?

bartlettroscoe:
@masterleinad, 

> > So perhaps it is best to just disable Kokkos tests when building under Trilinos for the foreseeable future?
> 
> Yes, let me hardcode that.

Not being able to run Kokkos tests when someone is building Trilinos on a new platform is a bit sketchy.  Perhaps just disable the native Kokkos tests by default for all Trilinos builds but allow the user to enable them if they would like?  Issues like https://github.com/trilinos/Trilinos/issues/11940#issuecomment-1572907274 are mostly a problem for automated testing that can't tolerate random failures.

>  I would appreciate getting some CI running on this.

I removed the label `AT: WIP` and set the label `AT: PRE-TEST INSPECTED` so the autotester should run to test this.

I will work on the TriBITS test and changes for this use case.

bartlettroscoe:
@masterleinad, since I suspect you can't see the PR testing results linked to on CDash [above](https://github.com/trilinos/Trilinos/pull/11779#issuecomment-1573897025), all of the builds failed with the configure error:

```
Processing enabled top-level package: ShyLU_Node (HTS, Tacho, Tests, Examples)
-- ShyLU_NodeHTS_hts_test_1: Added test (BASIC, PROCESSORS=2)!
CMake Error at packages/shylu/shylu_node/tacho/CMakeLists.txt:12 (MESSAGE):
  Tacho can not be build with Pthreads as the Kokkos Host Backend.


-- Configuring incomplete, errors occurred!
```

bartlettroscoe:
> Can you give me an example configuration?

@masterleinad, it can be hard to reproduce these PR configurations if you are not on an SNL machine.  See:

* https://github.com/trilinos/Trilinos/wiki/Reproducing-PR-Testing-Errors

Any chance you have access to the SNL SRN as part of your Kokkos collaboration work?

Otherwise, I will give this a try once I have a TriBITS example and test set up.

masterleinad:
@bartlettroscoe So it looks like the configuration does `TPL_ENABLE_Pthread=ON` and we currently have https://github.com/trilinos/Trilinos/blob/c3b1a94753b3611e3f931a1a1b6e9f026b30ecb8/packages/kokkos/cmake/kokkos_tpls.cmake#L64
which should imply that the default for `Kokkos_ENABLE_THREADS` should be the same but this is not actually the place where the default is set but in https://github.com/trilinos/Trilinos/blob/c3b1a94753b3611e3f931a1a1b6e9f026b30ecb8/packages/kokkos/cmake/kokkos_enable_devices.cmake#L22 and we are missing something like
```diff
diff --git a/packages/kokkos/cmake/kokkos_enable_devices.cmake b/packages/kokkos/cmake/kokkos_enable_devices.cmake
index d4a7744eb59..0f7bad1e999 100644
--- a/packages/kokkos/cmake/kokkos_enable_devices.cmake
+++ b/packages/kokkos/cmake/kokkos_enable_devices.cmake
@@ -18,8 +18,11 @@ KOKKOS_CFG_DEPENDS(DEVICES NONE)
 # Put a check in just in case people are using this option
 KOKKOS_DEPRECATED_LIST(DEVICES ENABLE)
 
-
+IF(Trilinos_ENABLE_Kokkos AND TPL_ENABLE_Pthread)
+KOKKOS_DEVICE_OPTION(THREADS ON HOST "Whether to build C++ threads backend")
+ELSE()
 KOKKOS_DEVICE_OPTION(THREADS OFF HOST "Whether to build C++ threads backend")
+ENDIF()
```
with current `Trilinos`/`Kokkos` `develop` to set the default correctly. This pull request forces `Kokkos_ENABLE_THREADS` to have the same values as `TPL_ENABLE_PThread` if specified, thus fixing the issue differently. As we are seeing right now, this uncovers that some of the test configurations are invalid: `Tacho` can't be built with `THREADS`/`Pthread` support but apparently `TPL_ENABLE_Pthread=ON` is set which implies `ShyLU_NodeTacho_ENABLE_Pthread=ON` and this variable is entirely ignored in favor of `Kokkos_ENABLE_THREADS` which is currently always `OFF` if not set explicitly.

masterleinad:
This means the `Trilinos` CI doesn't test the `THREADS` backend unless `Kokkos_ENABLE_THREADS=ON` is set.

bartlettroscoe:
@masterleinad, technically speaking, the value of `Kokkos_ENABLE_Pthread` should determine threading support in `Kokkos`, not `TPL_ENABLE_Pthread` (see [10.7 How to add a new TriBITS external package/TPL dependency](https://tribitspub.github.io/TriBITS/users_guide/index.html#how-to-add-a-new-tribits-external-package-tpl-dependency)).  So Kokkos should set the default for the cache var `Kokkos_ENABLE_THREADS` from `${Kokkos_ENABLE_Pthread}` and **NOT** `${TPL_ENABLE_Pthread}`.  

SIDENOTE: That reminds me, we need to update that in the Kokkos CMake function `KOKKOS_TPL_OPTION()` at:

https://github.com/trilinos/Trilinos/blob/c3b1a94753b3611e3f931a1a1b6e9f026b30ecb8/packages/kokkos/cmake/kokkos_tpls.cmake#L14-L17

to change `TPL_ENABLE_${PARSED_TRIBITS}` to `Kokkos_ENABLE_${PARSED_TRIBITS}` to be technically correct.

NOTE: The name `Pthread` is historical going back before even Kokkos.  CMake makes it hard to deal with multiple variables representing the same thing.  And then there is the relationship between cache and non-cache vars that makes things even more complex  (but TriBITS actually exploits that the non-cache var overrides the cache var as described in [TriBITS auto-enables/disables done using non-cache local variables](https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-auto-enables-disables-done-using-non-cache-local-variables)).

We will work out these issues.

sebrowne:
Can this be closed in favor of #13423 ?

masterleinad:
> Can this be closed in favor of #13423 ?

Sure.

PR #11798, opened 2023-04-17
TRILFRAME-110 to resurrect ATDM ATS2 build

srbdev:
@trilinos/framework 

## Motivation
Resurrect previous ATS2 genconfig build, but with newer compiler/cuda/mpi. See TRILFRAME-110 for more information.

### DIFF
--- a/packages/framework/ini-files/config-specs.ini
+++ b/packages/framework/ini-files/config-specs.ini
@@ -950,6 +950,184 @@ opt-set-cmake-var Trilinos_PARALLEL_LINK_JOBS_LIMIT STRING FORCE : 16
 # 4. Tweaks corresponding to particular systems.
 #------------------------------------------------------------------------------
 
+#
+# ATS2
+#
+[ATS2]
+use COMMON
+opt-set-cmake-var MPI_EXEC STRING FORCE : ${TRILINOS_DIR|ENV}/cmake/std/sems/trilinos_jsrun
+opt-set-cmake-var CTEST_USE_LAUNCHERS BOOL FORCE : ON
+opt-set-cmake-var Trilinos_CUDA_NUM_GPUS STRING : ${KOKKOS_NUM_DEVICES|ENV}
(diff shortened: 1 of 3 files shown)


packages/framework/ini-files/config-specs.ini
```diff
@@ -950,6 +950,184 @@ opt-set-cmake-var Trilinos_PARALLEL_LINK_JOBS_LIMIT STRING FORCE : 16
...
+
+
+[ats2_cuda-11.2.152-gnu-8.3.1-spmpi-rolling_release_static_Volta70_Power9_no-asan_no-complex_no-fpic_mpi_pt_no-rdc_no-uvm_deprecated-on_all]
+use ATS2
+use ATS2-COMMON-OVERRIDES
+
+use NODE-TYPE|CUDA
+use BUILD-TYPE|RELEASE
+# Override settings from [BUILD-TYPE|RELEASE]
+opt-set-cmake-var Kokkos_ENABLE_DEBUG_BOUNDS_CHECK BOOL FORCE : ON
```
sebrowne, line 1111: I assume this was restored from the previous build config, but do we really want Kokkos debugging enabled?  If it's just restored from old build, then we can leave it.

sebrowne (approved):

srbdev:
See [highlighted line](https://github.com/trilinos/Trilinos/pull/11798/files#diff-483364839bc0ba0c4b4723b848b4134b29913fb569bf1ef3d23ffdbc43b67470R101) for context on this comment: 

"It doesn't look like `argparse` supports differentiating an argument set by default vs. the user so this will overrides the argument no matter what. I think there's a way using a different parser with a different set of arguments and the `parse_known_args` method but seems a bit of a hack. "

srbdev:
It looks like I need to push these changes to `develop` to test because the PR driver seems to be setting up the `exec` command before it merges my changes, hence the reason the testing wasn't being ran on a compute node.

PR #11802, opened 2023-04-19
Adelus: Fix GPU-aware MPI issue on HIP and add specialization for HIPManagedSpace

vqd8a:
@trilinos/adelus

## Motivation

This PR is to 
- Fix the issue of incorrect results when using GPU-aware MPI on Tioga. ```Kokkos::fence```s are added in the two functions ```back_solve_rhs_pipelined_comm``` and ```perm1_``` to make sure device kernels finish before calling MPI functions with device pointers
- Add specialization for HIPManagedSpace in BLAS copy in Adelus

## Testing

Tested the GPU-aware MPI issue on Tioga and got correct results. 
I will test HIPManagedSpace later when HIPManagedSpace is also supported in other BLAS functions in Kokkos Kernels.


jdkotul (approved):
Two of the functions were changed, by adding Kokkos:fences for explicit synchronization, to ensure that the MPI collectives do not give incorrect results which was identified when using the GPU aware MPI on tioga, which is an AMD CPU / GPU system).  Note this information was also given to Cray HPE so they could address this issue.

PR #11816, opened 2023-04-25
Actions: Cleanup to the automentioner

csiefer2:
This should dramatically cut down on the mentions.

In the current version, it would check the title *and* the label every time a label was added.  Now it will only check the title on open.

### DIFF
--- a/.github/workflows/title_to_mention.yml
+++ b/.github/workflows/title_to_mention.yml
@@ -19,14 +19,14 @@ jobs:
     steps:
     - name: Mention MueLu
       uses: peter-evans/create-or-update-comment@v2
-      if: contains(github.event.label.name, 'MueLu') || contains(github.event.issue.title, 'MueLu')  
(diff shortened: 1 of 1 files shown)


.github/workflows/title_to_mention.yml
```diff
@@ -19,14 +19,14 @@ jobs:
     steps:
     - name: Mention MueLu
       uses: peter-evans/create-or-update-comment@v2
-      if: contains(github.event.label.name, 'MueLu') || contains(github.event.issue.title, 'MueLu')  
+      if: (contains(github.event.action, 'labeled') && contains(github.event.label.name, 'MueLu')) || (contains(github.event.action, 'opened') && contains(github.event.issue.title, 'MueLu'))
```
jhux2, line 22: what does `labeled` do?

GrahamBenHarper (approved):
It'll still run in addition to the title if a MueLu label is added, but it won't repeat on new labels added now, right?

jhux2 (approved):

jhux2:
Why not have the script add the MueLu label if "MueLu" appears in the title?

PR #11819, opened 2023-04-26
Piro: Add ProductModelEvaluator for steady state problems and a test with 2 parameters.

kliegeois:
This PR adds a Piro::ProductModelEvaluator that combines together multiple parameters into one so that API that supports one parameter such as ROL can be used with multiple parameters.

This PR adds a test with 2 parameters.

@mperego

### DIFF
--- a/packages/piro/test/Main_AnalysisDriver_Tpetra.cpp
+++ b/packages/piro/test/Main_AnalysisDriver_Tpetra.cpp
@@ -45,6 +45,7 @@
 
 #include "MockModelEval_A_Tpetra.hpp"
 #include "MockModelEval_B_Tpetra.hpp"
+#include "MockModelEval_B_Tpetra_2_parameters.hpp"
 //#include "ObserveSolution_Epetra.hpp"
 
 #include "Piro_SolverFactory.hpp"
@@ -61,6 +62,7 @@
 #include "Piro_StratimikosUtils.hpp"
 #include "Thyra_DetachedVectorView.hpp"
 #include "Tpetra_Core.hpp"
+#include "Piro_ProductModelEval.hpp"
 
 #ifdef HAVE_PIRO_MUELU
 #include "Stratimikos_MueLuHelpers.hpp"
@@ -92,7 +94,7 @@ int main(int argc, char *argv[]) {
 
   Piro::SolverFactory solverFactory;
 
-  for (int iTest=0; iTest<7; iTest++) {
+  for (int iTest=0; iTest<8; iTest++) {
 
     if (doAll) {
       switch (iTest) {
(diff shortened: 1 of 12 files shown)


packages/piro/test/_input_Analysis_ROL_ReducedSpace_TrustRegion_HessianBasedDotProduct_2_parameters.xml
```diff
@@ -0,0 +1,329 @@
...
+      <ParameterList name="Dakota">
+        <Parameter name="Input File" type="string" value="dak.in" />
+        <Parameter name="Output File" type="string" value="dak.out" />
+        <Parameter name="Restart File" type="string" value="dak.res" />
+        <Parameter name="Error File" type="string" value="dak.err" />
+      </ParameterList>
+      <ParameterList name="ROL">
+        <Parameter name="Number Of Parameters" type="int" value="2" />
+        <Parameter name="Response Vector Index" type="int" value="0" />
+        <Parameter name="Parameter Vector Index 0" type="int" value="0" />
```
mperego, line 183: Can you change this so that `index[i] != i`?

mperego (commented):
Looks good! Thanks for doing these complex changes!

I've left a comment in the added test. Also, can you add the same test to the Thyra_Solve test, so that we test also sensitivities for non analysis problems? Along those lines, can you please build Trilinos without ROL and make sure that the Piro sensitivities work without ROL? Unfortunately that's not tested by Trilinos PR testing.

packages/piro/test/_input_Analysis_ROL_ReducedSpace_TrustRegion_HessianBasedDotProduct_2_parameters.xml
kliegeois: Done, it works (after changing the order of the expected results).

kliegeois:
Thanks @mperego !
I am currently trying without ROL.
For the extra test in `Main_ThyraSolver_Tpetra.cpp` can it be done in a different PR?
It will require some time as right now this test only uses one of the `MockModelEval` (and not the one I modified).

packages/piro/test/Main_AnalysisDriver_Tpetra.cpp
```diff
@@ -227,8 +227,8 @@ int main(int argc, char *argv[]) {
             p_exact[1] = 4;
           }
           if (mockModel=="MockModelEval_B_Tpetra_2_parameters") {
-            p_exact[0] = 6;
-            p_exact[1] = 4;
+            p_exact[0] = 4;
```
mperego, line 230: I think you should not change this, but below use `p_view(p_indices[0]) `instead of `p_view(0)`, etc.

mperego: > Thanks @mperego ! I am currently trying without ROL. For the extra test in `Main_ThyraSolver_Tpetra.cpp` can it be done in a different PR? It will require some time as right now this test only uses one of the `MockModelEval` (and not the one I modified).

sounds good.

kliegeois: > I think you should not change this, but below use `p_view(p_indices[0]) `instead of `p_view(0)`, etc.

@mperego I am not sure about that because most of the tests tested in this executable only have 1 parameter (with 2 entries) and therefore do not have `p_indices[1]`.

So something like:
```
double l2_diff = std::sqrt(std::pow(p_view(p_indices[0])-p_exact[0],2) + std::pow(p_view(p_indices[1])-p_exact[1],2));
```
does not work for the other tests (but works for the added test of this PR).

We can modify this verification check to have two different code paths though but I am not sure if it is the best.

mperego (approved):

packages/piro/test/Main_AnalysisDriver_Tpetra.cpp
mperego: OK, we can address this later.

PR #11823, opened 2023-04-26
Tpetra: fix #11820

### LINKED ISSUE #11820: Tpetra: matrix filling, fillComplete performance regression
### Description
Between Mar. 28 and Mar. 29 (versions 7969c and a2064), the ATS-2 performance dashboard shows graph/matrix filling and fillComplete() slowing down significantly, ~50% or more.

These are the sub-timers that jumped:
- FEAssembly
  - ElementLoop (Graph)
  - FillComplete (Graph)
- MueLu Setup-Solve
  - Matrix Build
    - Galeri: Laplace 3D fillComplete
    - Galeri: Laplace 3D generation

The migration of the Jenkins job does not appear to be related. The Mar. 29 run was on the old jenkins-srn, and no script changes had happened by that point. Mar. 30 was the first run on the new JaaS jenkins.

brian-kelley:
In #11679, the WDV "sync path" was disabled when in a host-parallel region. However, it turns out that ``Serial::in_parallel()`` always returns false, so ``Kokkos::parallel`` kernels running on Serial would still suffer the overhead of WDV tracking. This brings back the explicit enable/disable of WDV tracking. This means a few more lines of boilerplate, but it fixes perf regression #11820, and as a bonus makes the fillComplete in FEAssembly perf test 20% faster than before #11679.

@trilinos/tpetra 

## Motivation

Fix major performance regression

## Related Issues

* Closes #11820 
* Blocks 
* Is blocked by 
* Follows #11679 
* Precedes 
* Related to 
* Part of 
* Composed of

### DIFF
--- a/packages/tpetra/core/src/CMakeLists.txt
+++ b/packages/tpetra/core/src/CMakeLists.txt
@@ -937,4 +937,4 @@ SET_PROPERTY(
 # / from this directory, or to / from the 'impl' subdirectory.  That ensures
 # that running "make" will also rerun CMake in order to regenerate Makefiles.
 #
-# Here's another change, another, and another
+# Here's another change, another, and another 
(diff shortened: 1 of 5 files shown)


jhux2 (approved):
lgtm

PR #11830, opened 2023-04-28
MueLu: Move Tpetra from implicit to explicit dependency

### LINKED ISSUE #11821: MueLu: Hard dependency on Tpetra is implicit
## Bug Report
@trilinos/muelu 

Based on our dependencies list, we have Tpetra listed as an optional package
```cmake
SET(LIB_REQUIRED_DEP_PACKAGES Teuchos Xpetra KokkosCore KokkosContainers KokkosKernels)
SET(LIB_OPTIONAL_DEP_PACKAGES Amesos Amesos2 AvatarT Belos Epetra EpetraExt Teko
                              Ifpack Ifpack2 Intrepid2 ML Tpetra
                              Zoltan Zoltan2Core Stratimikos Thyra ThyraTpetraAdapters
                              Isorropia)
```

However, since we require Kokkos to be enabled now, we have logic that is confusing like
```cmake
IF (${PACKAGE_NAME}_ENABLE_KokkosCore AND NOT ${PACKAGE_NAME}_ENABLE_Tpetra)
  MESSAGE(FATAL_ERROR "MueLu requires that if Kokkos is enabled, Tpetra must also be enabled.")
ENDIF()
```

We should clean up this sort of logic and make these sorts of details a little more obvious.

GrahamBenHarper:
@trilinos/muelu

Ever since we required Kokkos in #9532, we've had an implicit dependency on Tpetra since enabling Kokkos requires Tpetra in MueLu. This PR moves our hard dependency on Tpetra from an implicit dependency to an explicit dependency, and adjusts all of our CMake logic to reflect that.

Note: We seem to have a `test/permutation` folder, which I think only runs when Epetra is enabled and Tpetra is not? I haven't looked into it closely, but I'd like some feedback on whether it should be totally deleted or not.

**Important note:** there was a piece of refmaxwell guarded by `MUELU_HAVE_TPETRA`, which is a nonexistent macro. I only noticed that because I entered the same typo. This PR may now change RefMaxwell behavior due to that.

Closes #11821.

# Testing
This configures and builds in an extremely generic build on my Ubuntu desktop. The more important thing that has not been verified yet: the number of MueLu tests run by the autotester should not change from previous PRs. If it changes, then the logic was not handled properly.


mayrmt:
@GrahamBenHarper `test/permutation` is intended to test the `MueLu:: PermutationFactory`, a piece of code by @tawiesn that (as far as I know) is not used. It originates from @tawiesn's thesis work, however at least our group does not rely on it anymore. In sum, I'd be ok with removal.

GrahamBenHarper:
Ah that's good to know. I may handle that in a separate PR. If the testing is guarded by an Epetra only config, I wonder if the class is also guarded by and Epetra only config. I'll have to look into it.

cgcgcg:
This is WIP, right?

GrahamBenHarper:
@cgcgcg Yes, I meant to make it more explicit that this is a first attempt while I see how everything looks

GrahamBenHarper:
From another conversation: we do in fact have `#define HAVE_MUELU_TPETRA` and 168 tests pass on an experimental Tpetra build. Now, I want to remove the `HAVE_MUELU_TPETRA` guards everywhere as part of this, but it does in fact build and pass on a more official build.

GrahamBenHarper:
A crude comparison against the testing results from Max's PR that merged yesterday shows the number of tests has not changed.
image

Now we'll see if I've handled all of the `HAVE_MUELU_TPETRA` logic properly...

GrahamBenHarper:
**Important note:** there was a piece of refmaxwell guarded by `MUELU_HAVE_TPETRA`, which is a nonexistent macro. I only noticed that because I entered the same typo. This PR may now change RefMaxwell behavior due to that.

PR #11842, opened 2023-05-02
MueLu: Clean ups for matrix-free example 01, fix Belos route

GrahamBenHarper:
@trilinos/muelu @pohm01 

Just some minor cleanups with the matrix-free driver that I had sitting around, like fixing the non-multigrid route and adding some more multigrid settings.

### DIFF
--- a/packages/muelu/research/graham/mf_example_01.cpp
+++ b/packages/muelu/research/graham/mf_example_01.cpp
@@ -60,6 +60,7 @@
 #include <stdio.h>
 #include <random>
 #include <unistd.h>
+#include <cmath>
 
 // Teuchos includes
 #include "Teuchos_GlobalMPISession.hpp"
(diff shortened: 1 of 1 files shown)


packages/muelu/research/graham/mf_example_01.cpp
```diff
@@ -528,85 +521,99 @@ int main_(Teuchos::CommandLineProcessor &clp, Xpetra::UnderlyingLib lib, int arg
       auto rhs_1d = Kokkos::subview (rhs_2d, Kokkos::ALL(), 0);
       SC x = x_left;
       for(size_t i=0; i<n_local; ++i) {
-        rhs_1d(i) = 4*h*h*PI*PI*sin(2*PI*x);
```
cgcgcg, line 531: Could use the math constants from Kokkos..

cgcgcg (approved):

packages/muelu/research/graham/mf_example_01.cpp
GrahamBenHarper: I think when I first started this, some of those constants were in the Kokkos experimental namespace. Are they in a different namespace now?

cgcgcg: https://kokkos.github.io/kokkos-core-wiki/API/core/numerics/mathematical-constants.html

GrahamBenHarper: Perfect, thanks! I'll queue that up for next time :)

Issue #11845, opened 2023-05-03
Ifpack2: build errors with SC=complex<float>


jhux2:
## Bug Report
Reported by @jennloe.

@trilinos/ifpack2 @GrahamBenHarper 

<details>

<summary>GPU compile error</summary>

```
packages/ifpack2/src/CMakeFiles/ifpack2.dir/Ifpack2_Details_Chebyshev_OpenMP.cpp.o -c /ascldap/users/jloe/Software/AscicGPUTrilinosDev/packages/ifpack2/src/Ifpack2_Details_Chebyshev_OpenMP.cpp
/ascldap/users/jloe/TrilinosDev/Trilinos/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp(1346): error: no operator "/" matches these operands
            operand types are: double / std::complex<float>
          detected during instantiation of "void Ifpack2::Details::Chebyshev<ScalarType, MV>::fourthKindApplyImpl(const Ifpack2::Details::Chebyshev<ScalarType, MV>::op_type &, const MV &, MV &, int, Ifpack2::Details::Chebyshev<ScalarType, MV>::ST, const Ifpack2::Details::Chebyshev<ScalarType, MV>::V &) [with ScalarType=std_complex0float0, MV=Tpetra::MultiVector<std_complex0float0, int, longlong, Tpetra_KokkosCompat_KokkosOpenMPWrapperNode>]" 
(1045): here

/ascldap/users/jloe/TrilinosDev/Trilinos/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp(1367): error: no operator "*" matches these operands
            operand types are: double * const std_complex0float0
          detected during instantiation of "void Ifpack2::Details::Chebyshev<ScalarType, MV>::fourthKindApplyImpl(const Ifpack2::Details::Chebyshev<ScalarType, MV>::op_type &, const MV &, MV &, int, Ifpack2::Details::Chebyshev<ScalarType, MV>::ST, const Ifpack2::Details::Chebyshev<ScalarType, MV>::V &) [with ScalarType=std_complex0float0, MV=Tpetra::MultiVector<std_complex0float0, int, longlong, Tpetra_KokkosCompat_KokkosOpenMPWrapperNode>]" 
(1045): here

/ascldap/users/jloe/TrilinosDev/Trilinos/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp(1375): error: no operator "*" matches these operands
[...]

PR #11849, opened 2023-05-03
Ifpack2: fix build with complex<float>

brian-kelley:
Use magnitude type, not double, since you can't directly multiply ``complex<float>`` by double

@trilinos/ifpack2 

## Motivation
#11845

### DIFF
--- a/packages/ifpack2/src/Ifpack2_DatabaseSchwarz_decl.hpp
+++ b/packages/ifpack2/src/Ifpack2_DatabaseSchwarz_decl.hpp
@@ -388,7 +388,7 @@ private:
   std::vector<int> DatabaseIndices_;
 
   /// Patch weights
-  std::vector<double> Weights_;
+  std::vector<magnitude_type> Weights_;
 
   /// Pivots used for LAPACK
   mutable Teuchos::Array<int> ipiv_;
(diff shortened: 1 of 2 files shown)


jhux2:
@brian-kelley Thanks.  I'm puzzled, though, as to why this wasn't caught in PR or nightly testing.

brian-kelley:
@jhux2 Skimming the build list on cdash, it looks we have builds that do complex<double> but none with complex<float>.

@trilinos/framework Is it possible to create a new build with float/complex<float> enabled, or add them to an existing build? If it's a non-CUDA build, it doesn't appear to increase the parallel build time that much.

jhux2:
> you can't directly multiply complex<float> by double

> Skimming the build list on cdash, it looks we have builds that do complex<double> but none with complex<float>.

Yes, I should have read your comment more carefully :P.

brian-kelley:
@jhux2 You probably read it correctly the first time, I didn't know stuff in angle brackets dissapears unless you make it ``code style``. I fixed it a minute ago

jennloe:
Note: I've confirmed that this PR fixes issue #11845.

jennloe:
Linking to issue #10742

Issue #11874, opened 2023-05-10
ROL: intermittent failures of `ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2`


cwpearson:
## Bug Report
@trilinos/rol 

### Description
Periodic failure of `ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2`

This test has periodically failed for one of my PRs, looking like this: [CDash](https://trilinos-cdash.sandia.gov/test/26329290)

```
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascicgpu050 and rank 0!
Proc 0/2
Sweeps = 1
Omega = 0.6666
Levels = 3
Teuchos::GlobalMPISession::GlobalMPISession(): started processor with name ascicgpu050 and rank 1!
Proc 1/2
NORM EXACT= 41104.9


START CG = 0.488699
CG Residual 0 = 0.0771334
CG Residual 1 = 0.00123099
CG Residual Reduction  = 0.00123099
START CG = 0.160014
CG Residual 0 = 0.0244437
CG Residual 1 = 0.000124952
CG Residual Reduction  = 0.000124952
[ascicgpu050:207404] *** An error occurred in MPI_Recv
[ascicgpu050:207404] *** reported by process [2251030529,0]
[ascicgpu050:207404] *** on communicator MPI COMMUNICATOR 4 SPLIT FROM 0
[ascicgpu050:207404] *** MPI_ERR_TRUNCATE: message truncated
[ascicgpu050:207404] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[ascicgpu050:207404] ***    and potentially your MPI job)
[ascicgpu050:207399] 1 more process has sent help message help-mpi-errors.txt / mpi_errors_are_fatal
[ascicgpu050:207399] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

and I have managed to reproduce it on my mac with ApplyClang 14.0.0 and OpenMPI 4.1.5 on `develop` (cd2c0a14).

### Steps to Reproduce

```
cmake -DTPL_ENABLE_MPI=ON -DTrilinos_ENABLE_ROL=ON -DROL_ENABLE_TESTS=ON

make

while true; do ctest -R ROL_example_PinT_parabolic-control_Augmented; done
```

```
...

Subproject Time Summary:
ROL    =   0.59 sec*proc (1 test)

Total Test time (real) =   0.73 sec
[...]

Issue #11621, opened 2023-03-02
ROL: ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2 test failed sporadically


sebrowne:
## Question
@trilinos/rol 
I saw ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2 fail twice last night in the newly-reenabled Intel 2021.3.0/OpenMPI 4.0.5 PR build.  I wanted to ask if there was any possibility if this was a code error, or if we should suspect an issue with the MPI installation/system load/other non-code issue?  I recognize that this is very difficult to answer definitively, so a guess is fine.  Just trying to decide how to handle it if it starts to fail more frequently.

Here are the two specific failure instances:
[ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2](https://trilinos-cdash.sandia.gov/test/20923276)
[ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2](https://trilinos-cdash.sandia.gov/test/20948962)

sebrowne:
I actually just saw it on #11614 as well, but this time the CUDA build (still OpenMPI 4.0.5 though): [ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2](https://trilinos-cdash.sandia.gov/test/21003042)

dridzal:
@sebrowne :  If the failure is random, I'm not sure what to do.  @eric-c-cyr , do you think this could be a code issue?

sebrowne:
I know this is a tall ask because it's not usually failing.  I only noticed it because it was the same test a couple of times.  Figured I would ask in case it was obviously related to any recent ROL work, but by no means do I think that's the only possibility here.  I'll see if I can reproduce it on my blade if I run the test in a loop (which is still not fun, but possibly may be debuggable, especially if I can build my own MPI and that solves the issue).

sebrowne:
Okay, I have an update!  Using this script:
```
#!/bin/bash

for i in {1..100}
do
    echo -e "\n#############\nITERATION ${i}\n#############"
    ctest -V --timeout 10 -R ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2 || exit 1
done
```
[...]

PR #11875, opened 2023-05-10
Framework: disable ROL test for PR testing

jhux2:
Disable ROL_example_PinT_parabolic-control_AugmentedSystem_test_MPI_2 in PR testing, with the blessing of the ROL developers.

Reported in issues #11621, #11874, and #11654.

@trilinos/framework @trilinos/rol @cwpearson


dridzal (approved):

jhux2:
@sebrowne Just curious, is there a process in place for recording this?

sebrowne:
@jhux2 there is not (hence the large number of comments in that file related to test disables).  Just so that everybody knows my general opinion, I'm very averse to disabling any tests for a couple of reasons.  One of which I think you're alluding to is that we don't really track which ones are off, and hence they can get disabled and forgotten, never to be of use again.  Only disabling them on specific toolchains is one mitigation for that, and in this case because it's a known problem with our internal toolchain installations, I think it makes sense.

PR #11878, opened 2023-05-11
Teuchos: sprintf -> snprintf in Teuchos_Utils.cpp

cwpearson:
@trilinos/teuchos 

`sprintf` is deprecated in macOS 13, and issues warnings like this during the build:

```
.../Trilinos/packages/teuchos/core/src/Teuchos_Utils.cpp:107:7: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
        std::sprintf(s, "%g", x);
```

`snprintf` has been available since c++11 and is a suitable replacement in this circumstance.

### DIFF
--- a/packages/teuchos/core/src/Teuchos_Utils.cpp
+++ b/packages/teuchos/core/src/Teuchos_Utils.cpp
@@ -83,28 +83,28 @@ std::string Utils::trimWhiteSpace( const std::string& str )
 std::string Utils::toString(const int& x)
 {
 	char s[100];
-	std::sprintf(s, "%d", x);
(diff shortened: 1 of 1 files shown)


bartlettroscoe (approved):
If it passes the builds and tests, then this looks good to me.

NOTE: This is very old code and I am not sure how much of this is actually used.  (It would be great if one of the Trilinos PR builds was a coverage build so we could see the coverage for lines of code that are changed in a PR.)

PR #11880, opened 2023-05-11
Tpetra: Mark MultiVector functions as override to silence warnings

cwpearson:
Mark various overridden MultiVector functions `override`, which will issue an error if the DistObject interface changes.
This also silences `-Winconsistent-override` warnings.


jhux2 (approved):
lgtm

PR #11882, opened 2023-05-11
Amesos: Cleanup memory leak in KLU test

hkthorn:
@trilinos/amesos 

## Motivation

The new dashboard for memory leaks in Trilinos has identified an issue in the Amesos tests.  This is being addressed with this commit.

https://trilinos-cdash.sandia.gov/index.php?project=Trilinos#!/%23DynamicAnalysis

## Testing

Mac OSX GCC12 / RHEL7 Intel 19.x


iyamazaki (approved):
@hkthorn. This looks good, thank you!! Just curious (probably not important for this simple test):
- We don't worry about Map/Matrix's creation failed.
- Should we delete Map/Matrix before the convergence check, or don't need to if it exits with convergence failure.

PR #11889, opened 2023-05-15
Ifpack2 BlockTriDi: updating example

lucbv:
Adding more logic to generate 3D matrices and have the block size as command line parameter. The driver still fails in some cases for unknown reasons at this time. One big issue at the moment is that only the y direction gets distributed across MPI ranks...

@trilinos/ifpack2 

## Motivation
This work allows us to reproduce more closely the use case for the BlockTriDiagSolver

## Testing
Tested on Frontier

### DIFF
--- a/packages/ifpack2/example/BlockTriDi.cpp
+++ b/packages/ifpack2/example/BlockTriDi.cpp
@@ -14,11 +14,14 @@
 #include "Teuchos_StackedTimer.hpp"
 #include <Teuchos_StandardCatchMacros.hpp>
 
+#include <fstream>
+#include <sstream>
+
 namespace { // (anonymous)
 
 // Values of command-line arguments.
 struct CmdLineArgs {
-  CmdLineArgs ():blockSize(-1),numIters(10),tol(1e-12),nx(172),lpp(10),useStackedTimer(false),overlapCommAndComp(false){}
+  CmdLineArgs ():blockSize(-1),numIters(10),tol(1e-12),nx(172),ny(-1),nz(-1),mx(1),my(1),mz(1),lpp(10),useStackedTimer(false),overlapCommAndComp(false){}
 
   std::string mapFilename;
   std::string matrixFilename;
@@ -30,6 +33,9 @@ struct CmdLineArgs {
   int nx;
   int ny;
   int nz;
+  int mx;
+  int my;
+  int mz;
   int lpp;
   bool useStackedTimer;
   bool overlapCommAndComp;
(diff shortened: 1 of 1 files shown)


jhux2 (changes requested):
Needs to merge to `develop`, not `master`.

lucbv:
Thanks @jhux2, I updated the base branch

lucbv:
Ideally a github action should be able to detect that and block the PR automatically...

jhux2 (approved):

jhux2:
@lucbv I was trying to dismiss my review and accidentally approved.  @kliegeois should review.

packages/ifpack2/example/BlockTriDi.cpp
```diff
@@ -278,23 +356,81 @@ main (int argc, char* argv[])
...
-    // line info (lpp lines per proc)
-    int line_length = std::max(1, args.nx  / args.lpp);
+    // line info (sublinesPerLine lines per proc along direction x)
+    if ( args.sublinesPerLine < 1 && args.sublinesPerLine != -1) {
+      std::string msg = "the value of sublinesPerLine = " + std::to_string(args.sublinesPerLine) + " is not supported";
+      throw std::runtime_error(msg);
+    }
+
+    int line_length = std::max(1, (int) std::ceil(args.nx  / args.sublinesPerLine));
+    int line_per_x_fiber = std::ceil(args.nx  / line_length);
```
jhux2, line 413: Please put a comment as to what an "x_fiber" is.

packages/ifpack2/example/BlockTriDi.cpp
```diff
@@ -128,50 +145,112 @@ Teuchos::RCP<Tpetra::BlockCrsMatrix<Scalar, LocalOrdinal, GlobalOrdinal, Node> >
...
-    if(lclRowInd > meshRowMap.getMinLocalIndex ()) {
-      lclColInds[0] = lclRowInd - 1;
-      bcrsmatrix->replaceLocalValues(lclRowInd, lclColInds.getRawPtr(), &offmatrix[0], 1);
+  int entryIdx = 0;
+  for(int rowIdx = 0; rowIdx < blocksize; ++rowIdx) {
+    for(int colIdx = 0; colIdx < blocksize; ++colIdx) {
+      entryIdx = rowIdx*blocksize + colIdx;
+      basematrix[entryIdx] = ((entryIdx % 2) == 0 ? one : two);
+    }
+    basematrix[rowIdx*blocksize + rowIdx] = three + 26 * one + blocksize * two;
```
jhux2, line 172: Since this an example, it would be good to have comment to explain this formula.

jhux2 (commented):
Couple comments.

packages/ifpack2/example/BlockTriDi.cpp
kliegeois: Does something like this is clear enough?

```
    // We compute the number of lines oriented along the x direction of the mesh.
    // This number is called line_per_x_fiber where a fiber refers to an initial
    // x line in the mesh before dividing it in sublines.
```

packages/ifpack2/example/BlockTriDi.cpp
kliegeois: Does a comment like this make sense to you?

```
    // We enforce that the block matrix is SPD by setting the diagonal value
    // to a value which is greater than the sum of the absolute value of all the
    // entries on the corresponding row knowing that the off diagonal blocks are
    // set to minus identity.
```

packages/ifpack2/example/BlockTriDi.cpp
jhux2: @kliegeois Yes, that explanation is clear, thank you.

packages/ifpack2/example/BlockTriDi.cpp
jhux2: @kliegeois Yes, this is also clear.

kliegeois:
Thanks for the review @jhux2 !

jhux2 (approved):
lgtm

kliegeois:
@jhux2 what was the error?

PR #11896, opened 2023-05-16
Tpetra,Shylu: Drop all cmake `ASSERT_DEFINED(Kokkos_ENABLE_{CUDA,OPENMP,...})`

### LINKED ISSUE #11895: Tpetra: compatibility with external Kokkos
## Enhancement
@csiefer2

This is related to enabling configuring Trilinos against a Kokkos standalone install.

Kokkos is kindly inquiring whether you would be willing to drop all the `ASSERT_DEFINED(Kokkos_ENABLE_{OPENMP,CUDA,...})` from your cmake files.
We support
```cmake
find_package(Kokkos REQUIRED)
if(Kokkos_ENABLE_CUDE)
  # ...
endif()
```
The `ASSERT_DEFINED` would force us to export `set(Kokkos_ENABLE_CUDA OFF)` when it was not enabled which we would rather not do.

dalg24:
@csiefer2 

## Motivation

This facilitates enabling a build of Trilinos against a standalone Kokkos.

## Related Issues

* Closes #11895 
* Related to kokkos/kokkos#6138


bartlettroscoe:
Otherwise, is this ready to be merged?

dalg24:
> The only danger here is that if Kokkos changes the names of these variables then any if() statement involving these will evaluate to false always. That is why packages use assert_defined() the make sure this does not happen.

We understand that.  The `Kokkos_ENABLE_<BACKEND>` variables are not going anywhere.  Even if we wanted it is virtually impossible for us to change by now.

> One problem is that the variables exposed to customers of Kokkos are different between when it is pulled in using `add_subdirectory(kokkos)` and when it is pulled in with `find_package(kokkos)`. I guess that just means that Trilinos will need some PR builds building Kokkos internally and others pulling in pre-installed Kokkos.

Which variables?  Certainly not the `Kokkos_ENABLE_<BACKEND>` ones.  This is something that is being relied on downstream everywhere, not an isolated use case.  The code in question will not silently break.

bartlettroscoe:
> Which variables? Certainly not the `Kokkos_ENABLE_<BACKEND>` ones. This is something that is being relied on downstream everywhere, not an isolated use case. The code in question will not silently break.

@dalg24, potentially any CMake cache variable defined under `kokkos/` using `add_subdirectory(kokkos)` could be used by a downstream Trilinos package since these variables are globally accessible.  (There is no such thing a a scoped CMake cache variable.)  The only way to have a scoped global-ish variable that I know if is to use properties on directories in CMake, see [here](https://stackoverflow.com/questions/10031953/how-to-set-the-global-variable-in-a-function-for-cmake).

Therefore, you can't assume that a Kokkos customer Cmake project will behave the same when using `add_subdirectory(kokkos)` vs. `find_package(kokkos).

PR #11897, opened 2023-05-16
Tpetra: SYCL build fixes

brian-kelley:
@trilinos/tpetra 

## Motivation

Some fixes for build errors with SYCL. Makes Tpetra lib and all tests build for INTEL_PVC arch.

## Testing

This is just about getting things to build. Most things seem to actually run, but not all. For example, TpetraCore_Bug10008_MPI_4 segfaults.


lucbv (approved):
Ugh, some of these things always slip through the cracks or just back in : (

PR #11904, opened 2023-05-22
Zoltan2:  Fix calculation of color offset for BlockCrs

etphipp:
This fixes the calculation of the offset for coloring BlockCrs when processing the colors in chunks.

@trilinos/zoltan2


egboman (approved):
Looks good. Thanks, @etphipp !

PR #11905, opened 2023-05-22
Ifpack2: Add fences for the BlockTriDi timers

kliegeois:
@trilinos/ifpack2 

The timers of the BlockTriDi solver were not accurate due to some missing fences.
This PR adds fences in the solver *only* if `HAVE_IFPACK2_BLOCKTRIDICONTAINER_TIMERS` is defined.

### DIFF
--- a/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
+++ b/packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
@@ -329,8 +329,10 @@ namespace Ifpack2 {
 
 #if defined(HAVE_IFPACK2_BLOCKTRIDICONTAINER_TIMERS)
 #define IFPACK2_BLOCKTRIDICONTAINER_TIMER(label) TEUCHOS_FUNC_TIME_MONITOR(label);
+#define IFPACK2_BLOCKTRIDICONTAINER_TIMER_FENCE(execution_space) execution_space().fence();
 #else
 #define IFPACK2_BLOCKTRIDICONTAINER_TIMER(label)
+#define IFPACK2_BLOCKTRIDICONTAINER_TIMER_FENCE(execution_space)
 #endif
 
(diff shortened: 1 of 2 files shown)


packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3896,6 +3914,7 @@ namespace Ifpack2 {
 # endif
         }
 #endif
+        IFPACK2_BLOCKTRIDICONTAINER_TIMER_FENCE(typename ImplType<MatrixType>::execution_space)
```
lucbv, line 3917: Probably not necessary just after an all reduce?

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -3934,6 +3953,7 @@ namespace Ifpack2 {
         } else {
           sweep_step_ = sweep_step_upper_bound_;
         }
+        IFPACK2_BLOCKTRIDICONTAINER_TIMER_FENCE(typename ImplType<MatrixType>::execution_space)
```
lucbv, line 3956: Looks like everything is serial here, do you worry that other things prior to this function call might be non-blocking? In that case I would suggest we control the flow in that location instead of here.

packages/ifpack2/src/Ifpack2_BlockTriDiContainer_impl.hpp
```diff
@@ -4130,6 +4150,7 @@ namespace Ifpack2 {
       //sqrt the norms for the caller's use.
       if (is_norm_manager_active) norm_manager.finalize();
 
+      IFPACK2_BLOCKTRIDICONTAINER_TIMER_FENCE(typename ImplType<MatrixType>::execution_space)
```
lucbv, line 4153: This is just after the end of a serial parallel for, this should not be required?

lucbv (changes requested):
I think there are too many fences, we should at least discuss why we need them before merging

Issue #11530, opened 2023-02-01
Belos/FEI/Ifpack2/NOX: Randomly Failing Tests on ascisc1** breaking PR builds


csiefer2:
SUMMARY: <summary-line> 2023-01-30

CC: @trilinos/<package-name\>, @<triage-contact\> (Trilinos <product-area-name\> Triage Contact)

<Checklist>
<???: Add label "type: bug"?>
<???: Add label "impacting: [tests, build, performance]"?>
<???: Add label "pkg: [from <TestName>]>
<???: Add label "PA: [see https://sems-atlassian-son.sandia.gov/confluence/display/TRIL/Triaging+contacts]?>

## Next Action Status

<status-and-or-first-action>

## Description

As shown in [this query](    
https://trilinos-cdash.sandia.gov/queryTests.php?project=Trilinos&begin=2023-01-29&end=2023-01-31&filtercount=4&showfilters=1&filtercombine=and&field1=groupname&compare1=63&value1=Pull%20Request&field2=status&compare2=63&value2=failed&field3=site&compare3=63&value3=ascic1&field4=block&field4count=4&field4field1=testname&field4compare1=61&field4value1=Ifpack2_SGSMT_compare_with_Jacobi_MPI_4&field4field2=testname&field4compare2=61&field4value2=NOX_1DfemStratimikos_MPI_4&field4field3=testname&field4compare3=61&field4value3=FEI_fei_ubase_MPI_3&field4field4=testname&field4compare4=61&field4value4=Belos_Tpetra_tfqmr_hb_2_MPI_4) (click "Shown Matching Output" in upper right) the tests:

* `Belos_Tpetra_tfqmr_hb_2_MPI_4`
* `FEI_fei_ubase_MPI_3`
* `Ifpack2_SGSMT_compare_with_Jacobi_MPI_4`
* `NOX_1DfemStratimikos_MPI_4`

in the unique GenConfig builds:

* `rhel7_sems-gnu-8.3.0-openmpi-1.10.7-openmp_release-debug_static_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables`
* `rhel7_sems-gnu-8.3.0-openmpi-1.10.7-serial_debug_shared_no-kokkos-arch_no-asan_no-complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables`

started failing on testing day 2023-01-30.

[...]

PR #11907, opened 2023-05-22
NOX: fix exit status of 1DfemStratimikos test

tkordenbrock:
`success` is initialized to `false` and only changed if `TEUCHOS_STANDARD_CATCH_STATEMENTS()` catches.  `success` should be set based on the solver status.  The other versions of the 1DfemStratimikos tests already do this.

@trilinos/nox

Discovered while investigating MPI exit hangs discussed in https://github.com/trilinos/Trilinos/issues/11530

Related to https://github.com/trilinos/Trilinos/issues/11530


rppawlo (approved):
Thanks @tkordenbrock !

Issue #11868, opened 2023-05-09
NOX: Don't just throw exceptions via 'throw "NOX error";'


bangerth:
NOX is a great nonlinear solver, but it reports errors via both nonzero return codes,  as well as by throwing exceptions. That's already awkward -- pick one or the other. But to make it worse, it throws exceptions as simply `char *` objects. See for example here: https://github.com/trilinos/Trilinos/blob/master/packages/nox/src/NOX_Direction_Newton.C#L328-L333

Throwing exceptions not derived from `std::exception` is generally poor style, because to catch such exception, you have to know its type. If it were derived from `std::exception`, you could just do `catch (const std::exception &e) { std::cout << e.what() << std::endl; }` to learn what the exception was about, which works for all classes derived from `std::exception`. But for everything else, you have to *know* the type thrown, and that type is generally not documented.

A better version of the `throwError()` function linked to above would be
```
void NOX::Direction::Newton::throwError(const std::string& functionName, const std::string& errorMsg)
{
    if (utils->isPrintType(Utils::Error))
      utils->err() << "NOX::Direction::Newton::" << functionName << " - " << errorMsg << std::endl;
    throw std::runtime_error("NOX Error");
}
```
which uses https://en.cppreference.com/w/cpp/error/runtime_error . Ideally, the exception would also include `functionName` and `errorMsg`. There are multiple versions of `throwError()` in the NOX code base, all of which should be changed in this way.

Just to illustrate how much these two design decisions above (return codes + exceptions, exceptions not derived from `std::exception`) make it difficult to properly deal with errors, here's the code in deal.II (to be proposed as a pull request soon) looks like:
```
    // Solve, then check whether an exception was thrown by one of the user
    // callback functions. If so, exit by rethrowing the exception that
    // we had previously saved. This also calls the destructors of all of
    // the member variables above, so we do not have to clean things up by hand.
    //
    // NOX has the annoying habit of reporting success through a return code,
    // but if a user callback function throws an exception (which we translate
    // into a return code of -1), then it throws its own exception. So we
    // have to check both the return code and also catch exceptions :-(
    try
      {
        const auto status = solver->solve();

        if (status == NOX::StatusTest::Converged)
          return solver->getNumIterations();
        else
          {
            // See if NOX aborted because we had thrown an exception in a user
            // callback:
            if (pending_exception)
              {
                std::exception_ptr this_exception = pending_exception;
                pending_exception                 = nullptr;

[...]

PR #11908, opened 2023-05-22
NOX: Avoid 'throw "..."' in favor of 'throw std::runtime_error("...")'

bangerth:
This patch addresses #11868, at least in part. In short, NOX has a habit of using code of the kind
```
  throw "NOX Error";
```
which is difficult to deal with because one has to know the type of the exception to be chosen. A better approach is to say
```
  throw std::runtime_error("NOX Error");
```
because now the exception is derived from `std::exception` and one can use the `virtual` function `std::exception::what()` to inquire about the error at the place where the exception is caught.

Of course, in reality this is only the first step. The error messages are not actually very informative here, because pretty much throughout all of NOX, we find this sort of pattern:
```
    utils.err() << "NOX::StatusTest::NormF::NormF - Unable to compute F"
        << std::endl;
    throw std::runtime_error("NOX Error");
```
What should *really* be happening is that the error message should be part of the exception, rather than go onto some sort of console. But that's for a separate patch.

@rppawlo @GrahamBenHarper FYI.


GrahamBenHarper:
Thank you @bangerth! It looks great to me, but I'll wait to make sure somebody from @trilinos/nox gives the final blessing :)

rppawlo (approved):
Thanks @bangerth ! Much appreciated! I will work on improving the messages.

PR #11912, opened 2023-05-24
tpetra(CrsMatrix): stop relying on raw pointers to support more complex scalar types

romintomasetti:
@trilinos/Tpetra

@csiefer2 @etphipp @brian-kelley @maartenarnst 

## Motivation

This PR:
1. ~removes `HAVE_SACADO_CXX11` (useless now that `c++17` is required)~ -> #11915
2. ~removes `HAVE_STOKHOS_CXX11` (same as `1.`)~ -> #11915
3. makes changes to `Tpetra::CrsMatrix` (see below for a more detailed description)

I'll appreciate any comment you have on the "refactoring" of `Tpetra::CrsMatrix`. Suggest changes, additional modifications; in short, feel free to take a moment to propose things while we are at it :smiley: Also, please note I'm waiting your first "high level thoughts" before going through a "self-introspection" of my own modifications.

### Changes to `Tpetra::CrsMatrix`

These changes were written while trying to make `Tpetra::CrsMatrix` work with a `Kokkos::View` of `Sacado::Fad::SFad<scalar_t, number-of-variables>`.

For such a specialized view, the value and its derivatives are stored in a rank-2 view, and `Kokkos::View<Sacado::Fad::SFad<double,Num>*>::pointer_type` returns the `double*` handle to this underlying storage.

This breaks the whole `CrsMatrix` class (at least methods I changed) that were relying on pointer arithmetic. For instance, `view.data() + 1` is not guaranteed to jump to the next `Sacado::Fad::SFad`.

To solve that problem, I introduced `Kokkos::Experimental::Begin`. I guess doing arithmetic on iterators is more future-proof (*e.g.* in `Tpetra::CrsMatrix::mergeRowIndicesAndValues`).

Therefore I needed to change some methods to take `Kokkos::View`s instead of `Teuchos::ArrayView`s. I think this change is desirable because it alleviates the need for a bunch of static casts or weird conversions to non-owning `Teuchos::ArrayView`. Just to quote some old comments in `Tpetra_CrsMatrix_def.hpp`:
```c++
        // KDDKDD UVM TEMPORARY:  refactor combineGlobalValues to take
        // KDDKDD UVM TEMPORARY:  Kokkos::View instead of ArrayView
        // KDDKDD UVM TEMPORARY:  For now, wrap the view in ArrayViews
        // KDDKDD UVM TEMPORARY:  Should be safe because we hold the KokkosViews
        rowIndsConstView = Teuchos::ArrayView<const GO> (  // BAD BAD BAD
                           rowIndsView.data(), rowIndsView.extent(0),
                           Teuchos::RCP_DISABLE_NODE_LOOKUP);
        rowValsConstView = Teuchos::ArrayView<const Scalar> (  // BAD BAD BAD
                           reinterpret_cast<const Scalar*>(rowValsView.data()), rowValsView.extent(0),
                           Teuchos::RCP_DISABLE_NODE_LOOKUP);
```

I'm more then open to discuss this change!


etphipp:
The Sacado/Stokhos changes look fine to me.  I would suggest splitting this PR into 2, one with just those changes and another for Tpetra, as the latter is likely to be controversial.

cwpearson:
@romintomasetti sorry for the delay. I've started a more thorough review but thought we could cover some higher-level stuff here.

if I'm understanding this correctly, the main pain point is in `CrsMatrix::mergeRowIndicesAndValues` where you've replaced the pointer arithmetic with iterators since the View's `pointer_type` 

https://kokkos.github.io/kokkos-core-wiki/API/core/view/view.html#_CPPv412pointer_type

is not necessarily a pointer to the `value_type`, but rather, the "scalar type", which I can't find documented, but 

https://kokkos.github.io/kokkos-core-wiki/API/core/view/view.html#_CPPv417scalar_array_type

calls out Sacado FAD explicitly as an edge case where the pointer type is a pointer to underlying storage (`double` in this case)

Most of the Tpetra part of the PR is mechanically propagating the consequences of wanting to use iterators in `CrsMatrix::mergeRowIndicesAndValues` instead of raw pointers. I'm broadly happy with how that looks.

Do you think it would be possible to reimplement `CrsMatrix::mergeRowIndicesAndValues` in terms of indices into the views instead of iterators (previously pointers), possibly replacing `sort2` with a Kokkos sort? Would that sidestep the whole issue, or is there another piece that I'm missing?

cwpearson:
We can ignore the `sort2` issue, it's more or less orthogonal.

PR #11913, opened 2023-05-24
Tpetra: fix UVM builds for recent execution space instance change

rppawlo:
Fixes failures in panzer unit tests for UVM=ON builds caused by PR #11734 

@trilinos/tpetra 

## Motivation

A couple of panzer unit tests were failing due to changes in #11734 

The issue was that there were two deep copies that were changed from using the two-argument version to the three-argument version. There is a big difference between the two. The two-arg fences automatically, while the three-arg does not. This broke UVM builds. This commit adds back the fences. Only the first deep_copy on line 1232 was needed to fix the panzer tests. But I'm pretty sure the second one (line 1260) needs the explicit fence as well (panzer testing was just not using this path). I'm not sure if we have unit tests that trigger the switch from host to device (switch testing from line 1232 to 1260). Maybe tpetra team can comment on this.

## Testing

Fixes panzer unit tests.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp
@@ -1229,6 +1229,7 @@ namespace Tpetra {
             // Copy src_j into tgt_j
             // DEEP_COPY REVIEW - HOSTMIRROR-TO-HOSTMIRROR
             Kokkos::deep_copy (space, tgt_j, src_j); 
+            space.fence();
           }
         }
       }
@@ -1256,6 +1257,7 @@ namespace Tpetra {
             // Copy src_j into tgt_j
             // DEEP_COPY REVIEW - DEVICE-TO-DEVICE
             Kokkos::deep_copy (space, tgt_j, src_j); 
+            space.fence();
           }
         }
       }


sebrowne:
> @sebrowne - The autotester for this PR seems to have issues similar to what happened to @jhux2 in #11748. The autotester reported that all gcc 8.3, cuda and intel builds as failed, but cdash shows that they all passed clean (except intel - those results did not post). Running a retest now.

Thanks @rppawlo, yes I agree it looks the same.  Everything is behaving so far today, so hopefully it was just a fluke yesterday and Tuesday....

rppawlo:
@cwpearson @csiefer2 @brian-kelley @jhux2 - would one of you mind reviewing this? It's blocking the empire sync. This just restores the original fencing behavior.

brian-kelley (approved):

brian-kelley:
@rppawlo This makes sense and is fine for now.

@cwpearson In the long term, I think we have to revisit the WrappedDualView design a bit to work with streams. Right now, all the automatic DualView syncs will globally fence. So if you want two streams to actually execute simultaneously, neither of them can call any ``WDV::getView`` functions.

PR #11914, opened 2023-05-25
Ifpack2: Speed up DatabaseSchwarz::Compute

GrahamBenHarper:
@trilinos/ifpack2 @cgcgcg

This speeds up the `Compute()` phase of `DatabaseSchwarz` by at least 8%-10% based on 1M DOF runs I was testing on Attaway thanks to a comment made by @rstumin earlier today.

This breaks out of the compression check loop if the accumulated difference already exceeds the tolerance, which helps immensely, especially as the patch size increases. There should be almost no performance hit because even the most basic branch prediction will handle this in an efficient manner.


GrahamBenHarper:
Testing immediately failed and there's no cdash entry? That's strange...

GrahamBenHarper:
@sebrowne is the autotester down?

jhux2 (approved):
sehr gut

PR #11915, opened 2023-05-25
sacado(cleaning): removing old CXX11 related macros

romintomasetti:
@trilinos/Sacado

@etphipp 

## Motivation

This PR is an extract from #11912 and is a simple cleaning of outdated `CXX11` stuff in Sacado (since C++17 is now a requirement).

## Related Issues

- #11912


packages/sacado/src/Sacado_ConfigDefs.h
```diff
@@ -119,14 +119,12 @@ Questions? Contact David M. Gay (dmgay@sandia.gov) or Eric T. Phipps
...
-#    if (__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 8) )
-#      define SACADO_ENABLE_NEW_DESIGN 1
-#    endif
-#  else
+#if defined(__GNUC__) && !defined(__clang__)
+#  if (__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 8) )
 #    define SACADO_ENABLE_NEW_DESIGN 1
 #  endif
+#else
+#  define SACADO_ENABLE_NEW_DESIGN 1
```
romintomasetti, line 127: @etphipp I guess this could also be removed, since C++17 requires a much more recent gcc version than 4.8 :smiley:

romintomasetti: And so the question would be: if so, should we also get rid of the "Sacado old design" ? :smiley:

romintomasetti: (of course that would be another PR)

packages/sacado/src/mpl/Sacado_mpl_is_convertible.hpp
```diff
@@ -32,56 +32,12 @@
 
 #include "Sacado_ConfigDefs.h"
 
-#ifdef HAVE_SACADO_CXX11
-
 #include <type_traits>
```
romintomasetti, line 35: @etphipp I guess the file can simply be deleted now :)

packages/sacado/src/Sacado_ConfigDefs.h
etphipp: This has been on the to-do list for a while.  I think some codes might still be using the old design though.

packages/sacado/src/mpl/Sacado_mpl_is_convertible.hpp
etphipp: True, but we would also need to make sure this isn't called anywhere

packages/sacado/src/Sacado_cmath.hpp
```diff
@@ -34,7 +34,7 @@
 #include "Sacado_ConfigDefs.h"
 
 // Define some math functions that aren't usually in cmath
-#if !( (defined(_GLIBCXX_USE_C99_MATH_TR1) && defined(__GXX_EXPERIMENTAL_CXX0X__)) || defined(HAVE_SACADO_CXX11) || defined(HAS_C99_TR1_CMATH) || defined(USER_DISABLE_SACADO_TR1_CMATH) )
+#if !( (defined(_GLIBCXX_USE_C99_MATH_TR1) && defined(__GXX_EXPERIMENTAL_CXX0X__)) || defined(HAS_C99_TR1_CMATH) || defined(USER_DISABLE_SACADO_TR1_CMATH) )
```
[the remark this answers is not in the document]
etphipp: I think we could delete this whole file.

etphipp (dismissed):

packages/sacado/src/Sacado_cmath.hpp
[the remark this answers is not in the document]
romintomasetti: I kept the file, because there is `if_then_else` that I don't want to touch now :smiley:

packages/sacado/src/Sacado_ConfigDefs.h
romintomasetti: Ok, not to be done now.

romintomasetti:
@etphipp I'm done ;) However, I pushed a new commit that dismissed your review :angel:

etphipp (dismissed):
Looks good, thanks!

romintomasetti:
@etphipp I guess we are missing `LABEL 'AT: PRE-TEST INSPECTED' NOT SET ` for triggering the pipeline?

etphipp:
I added the labels.

etphipp:
@romintomasetti not sure if you can see the CDash output, but there are numerous build errors about not being able to find `std::is_same`.  Looks like an include of `<typetraits>` is necessary.

romintomasetti:
@etphipp This is something I do all the time when working alone on a feature branch. I like having a clean git history.

See for instance https://www.caichinger.com/blog/2022/10/04/git_best_practices/.

In this case for instance, I didn't wanted to have a commit with the missing include. So I added the change, amended it and needed to force push.

Note that this method shouldn't impact the reviewer in general if you use the review functionalities of GitHub to mark files you reviewed and agreed. See for instance https://github.blog/2019-07-01-mark-files-as-viewed/ 😉

etphipp:
There is nothing wrong with adding a second commit.  I'm not seeing what changes you made to resolve the build errors.

romintomasetti:
@etphipp I think my big mistake here is that I forced pushed not only my modification but also a rebase on the current develop branch... My bad sorry!

My only relevant change is that I needed to include `type_traits` in `packages/sacado/src/mpl/Sacado_mpl_find.hpp` as you suggested :smile:

etphipp:
I think you should add `#include <typetraits>` to any file that had included `Sacado_mpl_is_same.hpp` previously to make sure `std::is_same<>` is available, as opposed to relying on it coming in through `Sacado_mpl_find.hpp`.  Some applications only include a subset of the Sacado headers and as it is now might cause problems for them.

romintomasetti:
@etphipp Indeed, and done :wink:

etphipp (approved):

PR #11918, opened 2023-05-25
Percept: Use Intrepid2 instead of Intrepid

mperego:
Percept depends on Intrepid which will be archived soon.
This PR contains the bare minimum changes to use Intrepid2 instead of Intrepid.
Note that at the moment all the Kokkos views are on Host. 
The `MDArray` used in Percept is now a `Kokkos:DynRankView`, instead of being a shards array. The Intrepid::FieldContainer objects have been replace by `Kokkos:DynRankView` objects.

I'm not sure that `IntrepidManager` class is still needed. While it is still true that Intrepid2 integrate functions can process one DOF field at a time, I think that using `Kokkos::subviews` would allow to select a given field without requiring deep copies. If so, removing IntrepidManager might be addressed in a subsequent PR. 

@trilinos/percept 

## Stakeholder Feedback
This branch will be tested by Brian Carnes form the Sierra team

## Testing
There are no Percept tests and little testing of Percept capabilities in Trilinos.
These changes will be tested in Sierra.


bricar (commented):
Glad to finally get this done! I like the idea of removing the IntrepidManager later.

rppawlo (approved):

PR #11919, opened 2023-05-25
Belos: fix return value on unit test

rppawlo:
@trilinos/belos 

## Motivation

While investigating MPI exit hangs discussed in https://github.com/trilinos/Trilinos/issues/11530 we found that some unit tests were declaring success (and passing the teuchos unit test harness  based on printing the string "TEST PASSED") yet still returning non-zero exit codes to mpi. The non-zero exit code is one possible path to triggering the mpi hangs.

success is initialized to false and only changed if the TEUCHOS_STANDARD_CATCH_STATEMENTS() catches. success should be set based on the solver status.

Related to https://github.com/trilinos/Trilinos/issues/11530


jhux2 (approved):
lgtm

PR #11925, opened 2023-05-30
 Xpetra: Don't use std::binary_function 

masterleinad:
@trilinos/xpetra

## Motivation
`std::binary_function` (https://en.cppreference.com/w/cpp/utility/functional/binary_function) has been deprecated since C++11 and removed in C++17. Recent compilers like `clang 16` don't include this class anymore in their STL implementation.
std::binary_function only provides aliases for `first_argument_type`, `second_argument_type`, and `result_type` which the classes in question already provide anyway.

## Related Issues

* Related to #11920

## Testing
Compiling Trilinos with Tpetra and clang-16.

### DIFF
--- a/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
+++ b/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp
@@ -150,7 +150,7 @@ namespace Tpetra {
   /// this and other reasons not to state <tt>using namespace
   /// std;</tt> in the global namespace.
   template<class Arg1, class Arg2>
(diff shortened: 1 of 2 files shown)


jhux2:
@trilinos/xpetra

sebrowne:
This got hit by the 404 error that is currently affecting the autotester (sporadically).  It will likely come back as failed, even if everything visible looks like a pass.  Recommend retesting once it returns if so.

cgcgcg:
I don't think we even use `firstArg` or `secondArg` anywhere. So we could probably just remove them.

PR #11939, opened 2023-06-01
Sacado, Stokhos: resolve issue #11890

### LINKED ISSUE #11890: Stokhos:  Tests will fail after https://github.com/kokkos/kokkos/pull/6008 for shared spaces (Cuda UVM at least)
## Bug Report

@etphipp

### Description

If one tries to bring the changes from
- https://github.com/kokkos/kokkos/pull/6008

in Trilinos, `Stokhos` tests will fail with `CUDAUVM` enabled. This is clear when looking at the following items:
* in the above-mentioned PR, there were fences in `Kokkos::View` constructor before and after allocation/initialization **only when Cuda UVM was enabled**. They are now removed in the Kokkos develop branch.
* The specialized `Kokkos::ViewMapping` for `Kokkos::Experimental::Impl::ViewMPVectorContiguous`-specialized `Kokkos::View` in `packages/stokhos/src/sacado/kokkos/vector/KokkosExp_View_MP_Vector_Contiguous.hpp` for instance, is a complete outdated re-write of the `Kokkos::ViewMapping` from `Kokkos`, and I guess it misses a fence somewhere to make the tests work.

I guess these specializations of `Kokkos::ViewMapping` for `Stokhos` types are outdated (it is clear when comparing `Kokkos::ViewMapping::allocate_shared` of the current version of `Kokkos` develop branch and the one of the specialization for `Stokhos` types).

I expect the same issue in `Sacado` (though I haven't checked that yet), and in fact any other package that would make specializations.

To me, this is an issue that goes beyond the scope of `Stokhos` and that we haven't foreseen in https://github.com/kokkos/kokkos/pull/6008 (@dalg24 @masterleinad). As soon as one needs a specialization for a custom value type of a specialized view, one needs to specialized so many things from Kokkos that it cannot be seriously maintained in the long term.

Taking the case of `Stokhos`, it seems only the allocation size (and surely a few other things I overlooked) would need to be specialized (a view of static fixed storage of size 8 of double would need allocations of `view size * SFS size` and the layout may change). Maybe inheriting from the default view mapping could be a solution, only some functions would need to be overridden.

I think this general "specializing a view mapping" issue should be tackled at the Kokkos level (to provided a simple way to say "hey my type is a little different from usual, here are the very few specializations it needs").

I think this problem is similar to
- #11873

in which deep copy and other things must also be re-written, and do no benefit of any novelty the Kokkos team would bring in Kokkos core because these rewrites in Trilinos subpackages are sometimes carved in stone for many years.

What do you think?

### Steps to Reproduce

1. Take the changes from the Kokkos PR above (or simply go to the Kokkos view constructor and delete the fences embedded in the `#ifdef KOKKOS_ENABLE_CUDAUVM`)
2. Run the Stokhos tests, at least `Stokhos_KokkosViewMPVectorUnitTest_Cuda_MPI_1` and `Stokhos_KokkosCrsMatrixMPVectorUnitTest_Cuda_MPI_1` will fail.
  Or simply run the following test (extracted from test `Kokkos_View_MP_SFS_LayoutLeft_DeepCopy_UnitTest:`)
```c++
TEST(KokkosView, Stokhos_SFS_DeepCopy)
{
    using scalar_t = Sacado::MP::Vector<Stokhos::StaticFixedStorage<int, double, STOKHOS_DEFAULT_ENSEMBLE_SIZE, execution_space>>;
    using view_t = Kokkos::View<scalar_t*, execution_space>;

    view_t device("device view", 1); // in here there is no fencing anymore

    // add a fence here to make it work
    // Kokkos::fence("hooray it works with this fence!");

    const auto& mirror = Kokkos::create_mirror_view(device);

    for(size_t irow = 0; irow < mirror.extent(0); ++irow) {
        for(size_t icol = 0; icol < STOKHOS_DEFAULT_ENSEMBLE_SIZE; ++icol) {
            // this line might run before the view value initialization kernel and make the test fail
            mirror(irow).fastAccessCoeff(icol) = irow * STOKHOS_DEFAULT_ENSEMBLE_SIZE + icol;
        }
    }

    Kokkos::deep_copy(device, mirror);

    const auto& mirror_check = Kokkos::create_mirror_view(device);
    Kokkos::deep_copy(mirror_check, device);

    for(size_t irow = 0; irow < mirror.extent(0); ++irow) {
        for(size_t icol = 0; icol < STOKHOS_DEFAULT_ENSEMBLE_SIZE; ++icol) {
            // will fail, most values are zeros instead of the expected value because the view value init. kernel ran after
            EXPECT_EQ(mirror_check(irow).fastAccessCoeff(icol), irow * STOKHOS_DEFAULT_ENSEMBLE_SIZE + icol) << "Wrong at (" << irow << "," << icol << ")";
        }
    }
}
```

ndellingwood:
@trilinos/sacado @trilinos/stokhos 

## Motivation
sacado, stokhos: add exec space instances to allocate_shared
    
Attempt to resolve #11890
    
Co-authored-by: Christian Trott <crtrott@sandia.gov>

Also referenced in kokkos/kokkos#6158

## Testing
These changes have been through a couple rounds of nightly testing and resolve the `Stokhos_KokkosViewMPVectorUnitTest_Cuda_MPI_1` and `Stokhos_KokkosCrsMatrixMPVectorUnitTest_Cuda_MPI_1` failures in Cuda builds with UVM enabled
AT will test more extensively

### DIFF
--- a/packages/sacado/src/KokkosExp_View_Fad.hpp
+++ b/packages/sacado/src/KokkosExp_View_Fad.hpp
@@ -1521,6 +1521,7 @@ public:
       m_impl_handle = handle_type( reinterpret_cast< pointer_type >( record->data() ) );
 
       if ( ctor_prop::initialize ) {
+        auto space = ((ViewCtorProp<void,execution_space> const &) prop).value;
         // Assume destruction is only required when construction is requested.
         // The ViewValueFunctor has both value construction and destruction operators.
 				if (execution_space_specified)
@@ -1537,6 +1538,7 @@ public:
 
         // Construct values
         record->m_destroy.construct_shared_allocation();
+        space.fence();
       }
     }
 
--- a/packages/sacado/src/KokkosExp_View_Fad.hpp
+++ b/packages/sacado/src/KokkosExp_View_Fad.hpp
@@ -1525,7 +1525,7 @@ public:
         // Assume destruction is only required when construction is requested.
         // The ViewValueFunctor has both value construction and destruction operators.
 				if (execution_space_specified)
-					record->m_destroy = functor_type( ( (ViewCtorProp<void,execution_space> const &) prop).value
+					record->m_destroy = functor_type( space
 							, (fad_value_type *) m_impl_handle
 							, m_array_offset.span()
 							, record->get_label()
--- a/packages/stokhos/src/sacado/kokkos/pce/KokkosExp_View_UQ_PCE_Contiguous.hpp
+++ b/packages/stokhos/src/sacado/kokkos/pce/KokkosExp_View_UQ_PCE_Contiguous.hpp
@@ -1405,7 +1405,7 @@ public:
       // Assume destruction is only required when construction is requested.
       // The ViewValueFunctor has both value construction and destruction operators.
       record->m_destroy = m_impl_handle.create_functor(
-        ( (ViewCtorProp<void,execution_space> const &) prop).value
+        space
         , ctor_prop::initialize
         , m_impl_offset.span()
         , m_sacado_size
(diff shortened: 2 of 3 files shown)


etphipp (approved):

etphipp:
Thanks @ndellingwood

PR #11941, opened 2023-06-02
STK: Field redeclaration with increased number of states, increases the number of states.

palmerb4:
@trilinos/stk
@alanw0 

## Motivation
Hi Alan,

STK currently forbids the redeclaration of a field with a different number of states. The attached modifications, on the other hand, alter STK such that the redeclaration of a field with a larger number of states increases the number of states. Similarly, the redeclaration of a field with fewer states leaves the current number of states intact. The motivation for this change can be seen from a simple example:

Consider the same paradigm as Nalu-wind, where different algorithms declare their required fields to isolate an algorithm's requirements with that algorithm. If, for example, one algorithm requires a "node_coordinates" field with two states (as is often needed for second-order time-stepping schemes), then every other algorithm that attempts to declare the "node_coordinates" field with one state will need to be modified, lest STK throw an error. If field states are to be treated as "the memory of the field," which consists of the last N values of said field, then increasing the number of states will have no impact on algorithms that require fewer states.

I will note that this change is incompatible with the view of states as being cyclic since increasing the number of states would cause algorithms that assume a certain cycle width to break. I leave it up to you to decide if field states rotating cyclically after calling update_field_data_states was a feature or merely an implementation detail about how shifting the states was performed. Otherwise, the attached modifications should have no impact on existing code.

## Testing
I added a test for this feature to UnitTestField.cpp, but I'll copy it here:
```
TEST(UnitTestField, testFieldStates)
{
  stk::ParallelMachine pm = MPI_COMM_SELF;

  typedef stk::mesh::Field<double> rank_one_field;
  const std::string name1("test_field_1");
  const std::string name2("test_field_2");
  const int spatial_dimension = 3;
  std::shared_ptr<stk::mesh::BulkData> bulkPtr = build_mesh(spatial_dimension, pm);
  stk::mesh::MetaData& meta_data = bulkPtr->mesh_meta_data();

  // Redeclaration with fewer number of states.
  rank_one_field* field1_v1 = &meta_data.declare_field<double>(stk::topology::ELEMENT_RANK, name1, 2);
  rank_one_field* field1_v2 = &meta_data.declare_field<double>(stk::topology::ELEMENT_RANK, name1, 1);
  EXPECT_EQ(field1_v1, field1_v2);

  // Redeclaration with greater number of states.
  rank_one_field* field2_v1 = &meta_data.declare_field<double>(stk::topology::ELEMENT_RANK, name2, 1);
  rank_one_field* field2_v2 = &meta_data.declare_field<double>(stk::topology::ELEMENT_RANK, name2, 2);
  EXPECT_EQ(field2_v1, field2_v2);
}
```

### DIFF
--- a/packages/stk/stk_mesh/stk_mesh/base/MetaData.hpp
+++ b/packages/stk/stk_mesh/stk_mesh/base/MetaData.hpp
@@ -880,7 +880,7 @@ field_type * MetaData::get_field(stk::mesh::EntityRank arg_entity_rank,
 
   Traits::assign_tags( tags );
 
-  FieldBase * const field = m_field_repo.get_field( arg_entity_rank, name , dt , Traits::Rank , tags , 0 );
+  FieldBase * const field = m_field_repo.get_field( arg_entity_rank, name , dt , Traits::Rank , tags , 1);
 
   STK_ThrowRequireMsg(field == nullptr || field->data_traits().type_info == dt.type_info || dt_void.type_info == dt.type_info,
                   "field " << field->name() << " has type " << field->data_traits().type_info.name() << " when expecting type " << dt.type_info.name());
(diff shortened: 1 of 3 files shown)


palmerb4:
Hi Dave, 

Thank you for the detailed reply and for thoughtfully considering my modification. I fully understand and agree with your assessment, as the current code supports algorithms that need a minimum number of states and those that need an exact number of states. My modification removes support for the latter. 

Even with this door closed, I believe I can support my desired functionality without directly modifying STK, through in a less elegant way.

Thanks for your time, 
Bryce

PR #11950, opened 2023-06-06
fastilu: Fix memory leak.

vbrunini:
@trilinos/shylu 

@iyamazaki 

## Motivation
Clang's leak sanitizer was reporting this memory leak in some Sierra tests.

## Testing
Tested a clang build with Ifpack2 FastILU enabled locally.


iyamazaki (approved):
Thank you for the fix, @vbrunini.

PR #11952, opened 2023-06-07
Ifpack2: Cache temporary vector in 4th kind Chebyshev

cgcgcg:
@trilinos/ifpack2 

## Motivation
Cache a vector instead of reallocating on every apply. Use 4th kind Chebyshev in MiniEM.


cgcgcg:
FYI @MalachiTimothyPhillips

jhux2 (approved):
lgtm

PR #11954, opened 2023-06-08
Stokhos:  allow explicit conversions of ensembles to integral types

etphipp:
Allows, e.g., static_cast<int>(x) where x is an ensemble, since this is something Kokkos-Kernels does.  Elsewhere in Trilinos we have relied on Teuchos::as<>, but Kokkos-Kernels can't rely on that.

@jhux2


kliegeois (approved):
It looks good to me, thanks!

PR #11964, opened 2023-06-12
Trilframe 378 complex float

srbdev:
@trilinos/framework 

## Motivation
The ASC Sierra Code requires support for complex<double> and float with the configure arguments:
```
-D Trilinos_ENABLE_COMPLEX_DOUBLE=ON \
-D Trilinos_ENABLE_FLOAT=ON \
```
Yet there appear to be no PR or `master` promotion builds that enable either of these.

For details, see:

- https://github.com/trilinos/Trilinos/issues/10742


cgcgcg:
Does `complex<float>` pass all the tests? I believe the other variants (`float`, `complex<double>`) are fine.

sebrowne:
@cgcgcg it actually doesn't build all of the way.  Some of the testers won't build, but I think everything that builds passes its tests (or at least most of it).

PR #11978, opened 2023-06-15
Ifpack2,MueLu,Stratimikos,Belos,Panzer: Timer options, Maxwell fixes

cgcgcg:
@trilinos/ifpack2 @trilinos/muelu @trilinos/stratimikos @trilinos/belos @trilinos/panzer 

## Motivation
- Panzer: Use a more appropriate solver input deck for MiniEM on HIP platforms.
- Panzer: Add an input deck for donut problems
- Ifpack2: Add option "timer for apply" that allows to enable/disable the apply timer. Default behavior is unchanged.
- MueLu: `Low` and `None` verbosity disable the MueLu timers. Also disable the Ifpack2 apply timers in that case.
- MueLu RefMaxwell: Avoid singleton aggregates in construction of special prolongator.
- MueLu RefMaxwell: Add an option to run sub-problems on partition of global communicator.
- Stratimikos, Belos: Add a CMake option to disable the timers in the Belos adapter. Default behavior is unchanged.
- Stratimikos: Fix double registration of Ifpack2 in test: https://testing.sandia.gov/cdash/test/162154837
- Stratimikos: Register Ifpack2 for Scalar!=double


packages/stratimikos/adapters/belos/src/BelosThyraAdapter.hpp
```diff
@@ -275,8 +278,9 @@ namespace Belos {
          const ScalarType beta, TMVB& mv )
     {
       using Teuchos::arrayView; using Teuchos::arcpFromArrayView;
-
+#ifdef HAVE_STRATIMIKOS_BELOS_TIMERS
       Teuchos::TimeMonitor tM(*Teuchos::TimeMonitor::getNewTimer(std::string("Belos::MVT::MvTimesMatAddMv")));
+#endif
```
bartlettroscoe, line 283: I am wondering if it might be more maintainable to define a macro for Stratimikos called something like `STRATIMIKOS_TIME_MONITOR()` which would replace this block of code with:

```
STRATIMIKOS_TIME_MONITOR("Belos::MVT::MvTimesMatAddMv");
```

In fact, I am kind of surprised that someone wrote lots of messy code like:

```
Teuchos::TimeMonitor tM(*Teuchos::TimeMonitor::getNewTimer(std::string("Belos::MVT::MvTimesMatAddMv")));
```

all over the place.

The definition of the macro would be:

```
#ifdef HAVE_STRATIMIKOS_BELOS_TIMERS
#  define STRATIMIKOS_TIME_MONITOR(NAME) \
    Teuchos::TimeMonitor tM(*Teuchos::TimeMonitor::getNewTimer(std::string(NAME)))
#else
#  define STRATIMIKOS_TIME_MONITOR(NAME)
#endif
```

bartlettroscoe (changes requested):
Could we define macros in each packages like `STRATIMIKOS_TIME_MONITOR()` suggested below for Stratimikos?  That would really clean this code up a good bit and make it more maintainable.

packages/stratimikos/src/Stratimikos_LinearSolverBuilder_def.hpp
```diff
@@ -541,6 +541,15 @@ void LinearSolverBuilder<Scalar>::initializeDefaults()
     );
 #endif
 
+#ifdef HAVE_STRATIMIKOS_IFPACK2
+  setPreconditioningStrategyFactory(
+    abstractFactoryStd<Thyra::PreconditionerFactoryBase<Scalar>,
+    Thyra::Ifpack2PreconditionerFactory<Tpetra::CrsMatrix<Scalar>>>(),
+    "Ifpack2", true
+    );
+#endif
```
bartlettroscoe, line 550: The updated code should allow for multiple registrations of the same preconditioner strategy without any errors.  (The latter registrations are ignored.)

cgcgcg: What updated code?

bartlettroscoe: https://github.com/trilinos/Trilinos/commit/f126ecf6df804c54f8a1b266ab8f148119cc1964

cgcgcg: Hm. That's odd. Cause there were crashes that got fixed by removing a duplicate registration: 
https://testing.sandia.gov/cdash/test/162173350

bartlettroscoe: Hum, that is not right.  How can I reproduce the problem you were seeing?

cgcgcg: The test fails in a couple of builds but I'm not quite sure what the common factor is:
https://testing.sandia.gov/cdash/testSummary.php?project=1&name=Stratimikos_Galeri_xpetra_double_Jacobi_Tpetra_MPI_4&date=2023-06-28
Let me debug this some more, maybe I missed something.

cgcgcg: @bartlettroscoe Have a look at f5197bf1c0a84d84c3804fa4277bf6b281ff790d. I think that was the issue.

packages/panzer/mini-em/example/BlockPrec/MiniEM_helpers.cpp
```diff
@@ -166,6 +166,10 @@ namespace mini_em {
 #ifdef KOKKOS_ENABLE_CUDA
             if (typeid(panzer::TpetraNodeType).name() == typeid(Tpetra::KokkosCompat::KokkosCudaWrapperNode).name())
               updateParams("solverMueLuRefMaxwellCuda.xml", lin_solver_pl, comm, out);
+#endif
+#ifdef KOKKOS_ENABLE_HIP
+            if (typeid(panzer::TpetraNodeType).name() == typeid(Tpetra::KokkosCompat::KokkosHIPWrapperNode).name())
+              updateParams("solverMueLuRefMaxwellCuda.xml", lin_solver_pl, comm, out);
```
jhux2, line 172: Is `solverMueLuRefMaxwellCuda.xml` the base parameter set?  The name threw me because of the `ifdef` guard.

cgcgcg: For the time being we can use the Cuda input deck for HIP as well. That's still better than using the CPU settings for rebalancing.

packages/muelu/src/Operators/MueLu_RefMaxwell_def.hpp
```diff
@@ -438,6 +438,16 @@ namespace MueLu {
...
 
+          if (parameterList_.get<bool>("refmaxwell: fill communicator", false)) {
+            int temp;
+            temp  = std::nearbyint(Teuchos::as<double>(numProcsAH  * numProcs) / Teuchos::as<double>(numProcsAH+numProcsA22));
+            numProcsA22 = std::nearbyint(Teuchos::as<double>(numProcsA22 * numProcs) / Teuchos::as<double>(numProcsAH+numProcsA22));
+            numProcsAH = temp;
+            if (numProcsAH+numProcsA22 == numProcs+1)
+              numProcsA22 -= 1;
+            TEUCHOS_ASSERT(numProcsAH + numProcsA22 == numProcs);
+          }
```
jhux2, line 449: Could you explain this?

jhux2 (commented):
One question.

jhux2 (approved):
lgtm

PR #11980, opened 2023-06-15
Tpetra: mark `BlockMultiVector::checkSizes` as `override`

cwpearson:
Fix for

```
.../Trilinos/packages/tpetra/core/src/Tpetra_BlockMultiVector_decl.hpp:551:16: warning: 'checkSizes' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
  virtual bool checkSizes (const Tpetra::SrcDistObject& source);
```


jhux2 (approved):
lgtm

PR #11989, opened 2023-06-20
Belos: Change argument order in MvDot in BelosBiCGStabIter.hpp

vqd8a:
@trilinos/belos 

## Motivation

This PR is to change the argument order when calling MvDot in
https://github.com/trilinos/Trilinos/blob/8e95d511dcb0d947c9a15e18506aadda7269762d/packages/belos/src/BelosBiCGStabIter.hpp#L596

For complex type, conjugate transpose should be applied to T, not S. It should changed to ```MVT::MvDot(*S,*T,tS);```

## Testing

Tested with Gemma complex matrices and Belos BiCGStab and got correct results.


hkthorn (approved):
LGTM

PR #12000, opened 2023-06-26
Tpetra: Some changes to build with MSVC

hartsw:
@trilinos/kokkos
@trilinos/teptra

## Motivation

I know that MSVC isn't in your test matrix, but the attached commit makes some simple changes such that  some extra packages can be compiled on MSVC for Windows.

As far as I know `__restrict` should be the same as `__restrict__` on GCC, so nothing should change.

## Testing

All packages should continue to build as before with your testing matrix.


jhux2:
@hartsw Thanks for the contribution.  Since Kokkos-Kernels is developed as a separate project, could you please split this PR and submit the Kokkos-Kernels change to https://github.com/kokkos/kokkos-kernels?

jhux2:
@trilinos/tpetra @trilinos/kokkos-kernels

hartsw:
@jhux2 I split out the kokkos-kernel stuff and removed it to post in a different PR.  Thanks!

jhux2 (approved):
lgtm

PR #12004, opened 2023-06-29
KokkosKernels: Patch to fix 11926

### LINKED ISSUE #11926: Tpetra: cusparse misalignment error in spmv on subvectors
## Bug Report
@tpetra, @kokkos-kernels, @csiefer2

### Description

We're using Anasazi's GeneralizedDavidson solver to solve an eigenproblem with an odd number of rows.   

On CPU, it works. However, in a cuda build with cusparse, the solver aborts with a "cudaErrorMisalignedAddress" error.

```
error( cudaErrorMisalignedAddress): misaligned address  ...

void KokkosSparse::Impl::spmv_cusparse<KokkosSparse::CrsMatrix<double const, int const, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, 
```

We tracked down the issue to an spmv call on the following line

- https://github.com/trilinos/Trilinos/blob/0a376ba29537c19384056f04d8925d7587bc3dc3/packages/anasazi/src/AnasaziGeneralizedDavidson.hpp#L1147

What appears to be happening is that Anasazi creates two multivectors with multiple columns (`d_V` and `d_AV`).  The first multivector stores certain vectors.  The second multivector serves to store the result of multiplying the matrix with these vectors. The issue appears to arise because Anasazi wants to do such spmv on subsets of the vectors (`V_new` and `AV_new`). In particular, we see the abort when it extracts from each multivector the second vector and then wants to do the spmv on those (`V_new` contains the second vector from `d_V` and `AV_new` contains the second vector from `d_AV`). It appears that when the eigenproblem has an odd number of rows, `V_new` and `AV_new` are not aligned in a way that cusparse expects. I wasn't able to find in the cusparse doc what exactly cusparse expects. But certain other cusparse functions expect alignement to 16 bytes, and, if that's the case here, then, if there is an odd number of rows, even if the multivector is aligned to 16 bytes, the second vector in it will be aligned to 8 bytes.

We're able to reproduce the issue using just functions from Tpetra (`getVectorNonConst` and `apply`). So even though the issue arises in a computation with Anasazi, it appears this is not really an Anasazi issue, but rather an issue concerning Tpetra's multivector alignment in connection with cusparse spmv.

We tried passing `Kokkos::AllowPadding` to the constructor of the (dual) view, but this didn't solve the problem.

We are unsure how to analyze/solve the issue further.

```
using scalar_t        = double;
using execution_space = Kokkos::DefaultExecutionSpace;

using local_ordinal_t  = Tpetra::Details::DefaultTypes::local_ordinal_type;
using global_ordinal_t = Tpetra::Details::DefaultTypes::global_ordinal_type;

using node = Tpetra::KokkosCompat::KokkosDeviceWrapperNode<execution_space>;

using multivector_t = Tpetra::MultiVector<scalar_t, local_ordinal_t, global_ordinal_t, node>;
using matrix_t      = Tpetra::CrsMatrix  <scalar_t, local_ordinal_t, global_ordinal_t, node>;
using map_t         = multivector_t::map_type;

TEST(SPMV, odd_number_of_rows)
{
    constexpr size_t num_rows = 5;
    constexpr size_t num_values_per_row = 3;

    auto contiguous_map = Teuchos::make_rcp<const map_t>(num_rows, 0, TEUCHOS_SERIAL_COMM);

    auto mat_A = Teuchos::make_rcp<matrix_t>(contiguous_map, num_values_per_row);

    mat_A->insertGlobalValues(
        0,
        Teuchos::tuple<global_ordinal_t>( 0 , 1 ),
        Teuchos::tuple<scalar_t        >( 1., 2.)
    );

    mat_A->fillComplete();

    auto jvec  = Teuchos::make_rcp<multivector_t>(contiguous_map, 2);
    auto kvec  = Teuchos::make_rcp<multivector_t>(contiguous_map, 2);

    mat_A->apply(*jvec->getVectorNonConst(0), *kvec->getVectorNonConst(0)); // works
    mat_A->apply(*jvec->getVectorNonConst(0), *kvec->getVectorNonConst(1)); // cusparse misaligned address for odd num_rows

    using view_t = Kokkos::View<scalar_t**, execution_space>;
    const auto lvec_values = view_t(Kokkos::view_alloc("values", Kokkos::AllowPadding), num_rows, 2);
    const auto lvec = Teuchos::make_rcp<multivector_t>(contiguous_map, lvec_values);

    mat_A->apply(*jvec->getVectorNonConst(0), *lvec->getVectorNonConst(1)); // cusparse misaligned address for odd num_rows
}

TEST(SPMV, even_number_of_rows)
{
    constexpr size_t num_rows = 6;
    constexpr size_t num_values_per_row = 3;

    auto contiguous_map = Teuchos::make_rcp<const map_t>(num_rows, 0, TEUCHOS_SERIAL_COMM);

    auto mat_A = Teuchos::make_rcp<matrix_t>(contiguous_map, num_values_per_row);

    mat_A->insertGlobalValues(
        0,
        Teuchos::tuple<global_ordinal_t>( 0 , 1 ),
        Teuchos::tuple<scalar_t        >( 1., 2.)
    );

    mat_A->fillComplete();

    auto jvec  = Teuchos::make_rcp<multivector_t>(contiguous_map, 2);
    auto kvec  = Teuchos::make_rcp<multivector_t>(contiguous_map, 2);

    mat_A->apply(*jvec->getVectorNonConst(0), *kvec->getVectorNonConst(0)); // works
    mat_A->apply(*jvec->getVectorNonConst(0), *kvec->getVectorNonConst(1)); // works for even number of rows

    using view_t = Kokkos::View<scalar_t**, execution_space>;
    const auto lvec_values = view_t(Kokkos::view_alloc("values", Kokkos::AllowPadding), num_rows, 2);
    const auto lvec = Teuchos::make_rcp<multivector_t>(contiguous_map, lvec_values);

    mat_A->apply(*jvec->getVectorNonConst(0), *lvec->getVectorNonConst(1)); // works for even number of rows
}
```

brian-kelley:
(Patch of https://github.com/kokkos/kokkos-kernels/pull/1889)

For cusparse 12 spmv, make sure y vector is aligned to 16 bytes. If not, call the native impl as a fallback.
@maartenarnst This should fix #11926.

@trilinos/kokkos-kernels 

## Motivation

Fixes ``cudaErrorMisalignedAddress`` errors when calling ``KokkosSparse::spmv`` with an output vector aligned to < 16 bytes (for example, a column subview of a multivector with odd number of rows). See the issue linked above.

## Stakeholder Feedback

Reported by @maartenarnst first, but everybody uses cusparse spmv and we want to support cusparse 12 correctly

### DIFF
--- a/packages/kokkos-kernels/sparse/src/KokkosSparse_spmv.hpp
+++ b/packages/kokkos-kernels/sparse/src/KokkosSparse_spmv.hpp
@@ -161,6 +161,12 @@ void spmv(KokkosKernels::Experimental::Controls controls, const char mode[],
     useFallback = useFallback || (mode[0] == Conjugate[0]);
 #endif
   }
+  // cuSPARSE 12 requires that the output (y) vector is 16-byte aligned for all
+  // scalar types
+#if defined(CUSPARSE_VER_MAJOR) && (CUSPARSE_VER_MAJOR == 12)
+  uintptr_t yptr = uintptr_t((void*)y.data());
+  if (yptr % 16 != 0) useFallback = true;
+#endif
 #endif
 
 #ifdef KOKKOSKERNELS_ENABLE_TPL_ROCSPARSE


packages/kokkos-kernels/sparse/src/KokkosSparse_spmv.hpp
```diff
@@ -161,6 +161,12 @@ void spmv(KokkosKernels::Experimental::Controls controls, const char mode[],
     useFallback = useFallback || (mode[0] == Conjugate[0]);
 #endif
   }
+  // cuSPARSE 12 requires that the output (y) vector is 16-byte aligned for all
+  // scalar types
+#if defined(CUSPARSE_VER_MAJOR) && (CUSPARSE_VER_MAJOR == 12)
```
romintomasetti, line 166: Should this be a `>=` comparison? Or do you expect that Cuda will revert this behavior at some point?

romintomasetti (commented):
I'm not convinced this is a good way to solve the problem of @maartenarnst. This fix would mean that the first vector (possibly aligned) will be sent to the backend version while others (possibly misaligned) will use the fallback. All this in a quite unpredictable manner...

packages/kokkos-kernels/sparse/src/KokkosSparse_spmv.hpp
brian-kelley: Since the need for alignment isn't documented in cuSPARSE 12, and this function has the same interface as cuSPARSE 11 (which doesn't seem to need the same alignment), I think there's a good chance that they'll do one of these things in a future version:
- go back to allowing ``sizeof(Scalar)`` alignment
- start documenting the need for 16-byte alignment

In the first case, we don't want to limit performance on future versions by falling back when we don't need to. And in the second, it would be easy to look this up and expand this condition to add version 13.

brian-kelley:
@romintomasetti I agree that this isn't ideal and makes the performance less predictable, but unfortunately I don't think there's an alternative now.

Once Kokkos deals with https://github.com/kokkos/kokkos/issues/2995, we can start padding all Tpetra multivectors to make sure each column is aligned, and then cuSPARSE will always be called. This issue is that ``create_mirror_view`` doesn't pad the mirror when the input is padded, so you can't actually deep_copy between them. This would break a ton of stuff in Trilinos if we tried to pad now.

And then more generally, we need ``KokkosSparse::spmv`` to work with vectors that are only aligned to ``alignof(Scalar)``. Even after we pad Tpetra::MultiVectors, the non-Tpetra users of spmv could still use <16 byte aligned vectors.

maartenarnst:
Hi @brian-kelley! Thanks for this fix! In the mean time, I have tested the fix with our code, and the fix indeed solves the issue. I've tested it with Cuda 12.1.1. Without the fix, the issue is still present. With the fix, the issue is no longer present.

I agree with @romintomasetti's concerns, and I also understand your replies. I don't have an idea of an other way of solving this. Other than that it may be worthwhile to bring the issue/question of the alignment to Nvidia's attention, and to bring the issue to the Kokkos team's attention to add to the motivations to solve kokkos/kokkos#2995.

packages/kokkos-kernels/sparse/src/KokkosSparse_spmv.hpp
romintomasetti: Ok, I agree with you. Maybe adding the test that @maartenarnst uses to Kokkos Kernels' test suite would be a nice addition to this PR, especially if it help you detect any behavioral change with CUDA 13 for instance :smile:

brian-kelley: Hi @romintomasetti , I just got an update from NVIDIA on this - starting in the next minor version (12.3), spmv will not require the 16-byte alignment, so we can go back to the old behavior. I made a note to narrow the version range for the workaround but it won't matter until that cusparse release (I'm leaving on vacation today)

PR #12005, opened 2023-06-29
Intrepid2: Pass non const value type to mapToPhysicalFrame.

maartenarnst:
@trilinos/intrepid2 @mperego

## Motivation

The `Intrepid2` function `mapToPhysicalFrame` allows to map coordinates of points in the reference domain (input) to coordinates in the physical domain (output). We would like to be able to pass in the input coordinates with a const value type. However, this is not currently possible because the value type is passed to a basis factory which expects a non const value type.  

This PR solves this issue by using `std::remove_const_t`.

This change was discussed with @CamelliaDPG at EUTug 2023.

### DIFF
--- a/packages/intrepid2/src/Cell/Intrepid2_CellTools.hpp
+++ b/packages/intrepid2/src/Cell/Intrepid2_CellTools.hpp
@@ -316,7 +316,8 @@ public:
                  const Kokkos::DynRankView<pointValueType,pointProperties...>       points,
                  const Kokkos::DynRankView<worksetCellValueType,worksetCellProperties...> worksetCell,
                  const shards::CellTopology cellTopo ) {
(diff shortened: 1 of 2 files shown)


mperego (approved):
Looks good to me. Thanks!

mperego:
@maartenarnst Have you consider using directly this function?
https://github.com/trilinos/Trilinos/blob/master/packages/intrepid2/src/Cell/Intrepid2_CellTools.hpp#L1033-L1036
If you already have the needed HGRAD basis you avoid having Intrepid2 create a new one.

maartenarnst:
Thanks, @mperego! And thanks for suggesting that function. We're using it in a case where the FE basis can be higher order, and can thus be different from the basis that's used for the geometric mapping. But even in this case, it may make sense to construct the degree 1 basis just once and then pass it to the function that you're suggesting.

PR #12009, opened 2023-06-29
Tpetra: fix #11959

### LINKED ISSUE #11959: Tpetra: Non-lazy H2D copy in graph construction
As discussed with @brian-kelley and @jhux2 

We should fix this.

brian-kelley:
Lazily copy row pointers from device to host, for both packed and unpacked.

@trilinos/tpetra 

## Motivation

In CrsGraph, this can avoid an eager host allocation and/or device-to-host copy. It also improves performance on CudaUVMSpace (or any other Kokkos SharedSpace) by using separate allocations for the host and device rowptrs. Since the data is immutable, this is better than having the same memory pages get passed back and forth at each access.

## Related Issues

* Closes #11959 
* Blocks 
* Is blocked by 
* Follows 
* Precedes 
* Related to 
* Part of 
* Composed of 

## Testing

<!--- 
## Additional Information
Anything else we need to know in evaluating this merge request

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_decl.hpp
@@ -2150,6 +2150,7 @@ public:
     row_ptrs_device_view_type rowPtrsPacked_dev_;
     mutable row_ptrs_host_view_type rowPtrsPacked_host_;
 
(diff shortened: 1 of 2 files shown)


brian-kelley:
This is blocked by a PR testing failure, need #12012 to fix it

brian-kelley:
Only autotest failure was fixed in #12012 , retesting

PR #12012, opened 2023-07-01
KokkosKernels: remove non-existant include dir

brian-kelley:
In CMake, don't export the binary directory kokkos-kernels/ode as an include dir. It doesn't exist since the ODE component has no auto-generated ETI files.

Patch of https://github.com/kokkos/kokkos-kernels/pull/1892.

This will fix errors in PR testing like this one at https://trilinos-cdash.sandia.gov/test/29606177 :
```
Found Tpetra!  Here are the details: 
   Tpetra_DIR = /scratch/trilinos/jenkins/ascic164/workspace/Trilinos_PR_intel-2021.3/pull_request_test/cmake_packages/Tpetra
-- Configuring done
CMake Error in CMakeLists.txt:
  Imported target "Tpetra::all_libs" includes non-existent path

    "/scratch/trilinos/jenkins/ascic164/workspace/Trilinos_PR_intel-2021.3/pull_request_test/packages/kokkos-kernels/ode"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.
```

@trilinos/kokkos-kernels 

## Testing

This fixes the test ``TrilinosInstallTests_simpleBuildAgainstTrilinos_by_package_build_tree``, but it would also fix a real app building against Trilinos with the same configuration.

### DIFF
--- a/packages/kokkos-kernels/ode/CMakeLists.txt
+++ b/packages/kokkos-kernels/ode/CMakeLists.txt
@@ -11,5 +11,8 @@ ENDIF()
 
 
 # Adding unit-tests
-KOKKOSKERNELS_INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR}/ode)
+# Note BMK: Since ODE has no auto-generated ETI files, this directory does not exist in a build without unit tests.
+# This causes configure errors when building an app against a Trilinos install, and the unit test build dir doesn't contain any headers that need to be found.
+# But uncomment the next line if ETI headers are ever added.
+# KOKKOSKERNELS_INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR}/ode)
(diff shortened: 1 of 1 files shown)


bartlettroscoe (approved):
I don't understand how earlier builds are not catching these errors.  I will a test that will call `find_package()` for every `<Package>Config.cmake` file in the build dir for every enabled package.

brian-kelley:
@bartlettroscoe I was confused too, but I think the recipe for this failure is, enabling TpetraCore or something downstream, but NOT enabling KokkosKernels directly. Because if KokkosKernels tests are built, this ``ode`` directory in the build will exist.

bartlettroscoe:
> @bartlettroscoe I think the recipe for this failure is, enabling TpetraCore or something downstream of KokkosKernels, but NOT enabling KokkosKernels itself. Because if KokkosKernels unit tests get built, this `ode` directory in the build will exist.

@brian-kelley, I think you are right.  That is the issue.  The KokkosKernels CMakeLists.txt files could use some cleanup to help address this an other issues I have run across.

PR #12013, opened 2023-07-02
Intrepid2: Avoid including `<Kokkos_Vector.hpp>` header

dalg24:
@mperego
@ndellingwood 

## Motivation

Kokkos is looking at deprecating `Kokkos::vector` and may want to emit warnings when that header is included.

Moreover, these header includes are unnecessary since `Kokkos::vector` is currently not actually used in Intrepid2 (nor anywhere else in Trilinos as far as I can tell).

* Related to kokkos/kokkos#6252

## Testing

N/A


mperego (approved):
Looks good to me. thanks!

mperego:
PR #12018 should have fixed the shadow warnings that popped up in the PR testing.

PR #12017, opened 2023-07-05
Phalanx: unit test for fad subviews

rppawlo:


rppawlo:
Closing. Tests were added to sacado. Don't need duplicates in phalanx.

PR #11811, opened 2023-04-24
Panzer :: Evaluator for intrepid2's local L2 projection

reuterb:
@trilinos/intrepid2 has recently added a wrapper which allows for local L2 projection of FE fields.
This PR includes a Panzer evaluator employing this functionality.

@trilinos/panzer

PR #12018, opened 2023-07-05
Intrepid2: fix ProjectField function

mperego:
@trilinos/intrepid2 

## Motivation
Fix Project field function so that it works also when the input views are subviews.
We were using view type of one of the function input view to define a new view, which can be problematic.

* Related to  #11811

## Testing
Modified test to replicate conditions that triggered the bug.


rppawlo (approved):

reuterb (dismissed):
Your fix is functionally identical to the one I've been using in my branch.
Sorry for the delay but we are getting hung up on some other issue unrelated to this particular change.

mperego:
> Your fix is functionally identical to the one I've been using in my branch. Sorry for the delay but we are getting hung up on some other issue unrelated to this particular change.

there's still something wrong when running on cuda. I'm trying to fix that.

mperego:
@rppawlo apparently we never used projections with subviews. Anyway now it's working, and we are testing with subviews. I also took the opportunity to add testing for Wedge elements, sorry if this makes the PR a bit less focused.

mperego:
@rppawlo I want to squash the commits, so please do not do set the `automerge` label.

mperego:
Thanks @rppawlo !

PR #12020, opened 2023-07-07
EXODUS: Fix for mpi-io with fixed-length path limit

gsjaardema:
@trilinos/seacas 

## Motivation

Some mpi-io implementations have a fixed length maximum of ~256 characters for the filename + path.  They sometimes prepend some other information, so there are cases where the maximum available is ~250 characters.  We are hitting close to those limits on some jenkins or other runs with long paths...


gsjaardema:
Not sure who is around today, but here is a quick small PR that could use a review...

ccober6 (approved):
LGTM

Issue #11214, opened 2022-11-01
MueLu: Improve Maxwell Support


csiefer2:
As per customer request.

PR #12021, opened 2023-07-07
ML: Adding sorting support for Maxwell1

csiefer2:
Part of the continuing ML/MueLu harmonization project.  Previously, you couldn't make the nodal aggregation hierarchy sort rows of the coarse A matrices.  This is problematic since Phase 2b aggregation is sensitive to row ordering.  This PR addresses that issue.

Issue: #11214 
Tests: n/a
Stakeholder feedback: n/a


jhux2 (approved):
lgtm

PR #12028, opened 2023-07-10
Framework: Remove need for source_branch in all PR scripts

sebrowne:
Won't have access to source_branch any more.

Switch to sha-based merging (not a big deal probably) to do the testing.

Remove the checks around what branches can/cannot be merged to master branch (github rules should take care of this automatically by only allowing the automated system to merge into master branch).

Also removed check for branch tip != SHA1, because that's not possible if we only use the SHA1.  The thing that cancels PR runs and makes you run a new one if it finds changes on the branch after a run starts should handle this issue.

User Support Ticket(s) or Story Referenced: TRILFRAME-573

@trilinos/framework 

## Motivation
New autotester is changing its interface a bit, and want to be ready for when it rolls out.

### DIFF
--- a/packages/framework/pr_tools/trilinosprhelpers/TrilinosPRConfigurationBase.py
+++ b/packages/framework/pr_tools/trilinosprhelpers/TrilinosPRConfigurationBase.py
@@ -649,36 +649,6 @@ class TrilinosPRConfigurationBase(object):
         return 0
 
 
-    def validate_branch_constraints(self):
-        """
-        Verify that the source branch is allowed.
-
-        For the `master` branch, we only allow the source branch to be
-        a protected branch named with the scheme `master_merge_YYYYMMDD_HHMMSS`
-        """
-        print("")
-        print("Validate target branch constraints:")
-        print("--- Target branch is '{}'".format(self.args.target_branch_name))
-
-        re_master_merge_source = "master_merge_[0-9]{8}_[0-9]{6}"
-        if "master" == self.args.target_branch_name:
-            print("--- Target branch is 'master'. Checking source branch constraints...")
-            if not re.match(re_master_merge_source, self.args.source_branch_name):
-                message  = "+" + "="*78 + "+\n"
-                message += "ERROR: Source branch is NOT trilinos/Trilinos::master_merge_YYYYMMDD_HHMMSS\n"
-                message += "       This violates Trilinos policy for pull requests into the master\n"
-                message += "       branch.\n"
-                message += "       Source branch provided is {}\n".format(self.args.source_branch_name)
-                message += "       Perhaps you forgot to set `develop` as the target in your PR?\n"
-                message += "+" + "="*78 + "+\n"
-                #print(message)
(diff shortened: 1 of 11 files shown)


sebrowne:
@prwolfe I added you as the original author of most of this to get your feeling on whether this is safe or not.  I can't see it as being a huge issue.  Sorry to bother, I'm sure you don't have many other things to do....

packages/framework/pr_tools/trilinosprhelpers/TrilinosPRConfigurationBase.py
```diff
@@ -649,46 +649,13 @@ def create_package_enables_file(self, dryrun=False):
         return 0
 
 
-    def validate_branch_constraints(self):
```
prwolfe, line 652: Is there a replacement for this safeguard?

prwolfe (commented):
Generally the code will work, but I am concerned about removal of guardrails here. Is there a replacement for this in the future? It's always good to avoid these things in advance.

packages/framework/pr_tools/trilinosprhelpers/TrilinosPRConfigurationBase.py
sebrowne: I'm not really that concerned about this one.  The repository permissions in GitHub should forbid anybody other than the autotester, you, and Jim W. at the moment.

sebrowne: And realistically I think I'm going to change them to be "only the autotester"

sebrowne: Okay, so I changed that.  Now that I'm thinking about it though, there may still be an issue with the autotester being able to merge if somebody uses AT: AUTOMERGE.  I'll see what I can come up with.  Really I want to set it up so that only the autotester can even _open_ PRs into master.

prwolfe: > Although, on second thought, what is the issue with direct merges to master? As long as it passes the correct status checks, I don't see why it should be an issue.

The makes sense. The main point is to assure that those check pass before a merge is allowed. Allows for a lot of upstream messiness.

sebrowne: Yes, the autotester system seems to always test changes into master with the correct set of builds (the way I'm reading the config files), so I think it's safe to remove this from a code quality standpoint.  Worst case, somebody decides to merge to the master branch, and the history isn't as clean as we want it.  However, I think it's reasonable to rely on code review to catch that particular issue.  Does that satisfy the concern of removing this check?

prwolfe: I would say so and if it becomes a problem address it as needed.

prwolfe (approved):

PR #12038, opened 2023-07-12
IOSS: Remove fmt dependency for Trilinos users

gsjaardema:
@trilinos/seacas 

## Motivation
Since Trilinos version of SEACAS has an embedded lib::fmt library, it should not be referenced as a TPL.  Remove the dependency so clients using Trilinos will build correctly.


sebrowne:
So in the future, is the plan for SEACAS to always have the internal fmt?

Long-term, wouldn't it be better to rely on the external one?  For example, if one was using spack, we could add the dependency on fmt easily, as opposed to having a possibly conflicting one within SEACAS?  Does my question make sense?

sebrowne:
> I would love to use an extern fmt in Trilinos.

You have _my_ full support, though how much realistic effect that will have is certainly not definitive.

gsjaardema:
> @sebrowne and @gsjaardema - I believe the fmt library has bee accepted in to the c++20 standard as std::format. We expect to require c++20 as the minimum standard in early FY25 in Trilinos based on Kokkos schedule. Instead of forcing users to build extra TPLs now (and spend developer effort in the transition), we could just wait until FY25 and then switch to the std library version (unless I am I missing something?)

That is a possibility; however, there are Trilinos clients that are themselves using lib::fmt currently and the embedded version in SEACAS can cause them issues if the two versions are out-of-sync.  I've tried to eliminate all uses of fmt in include files or the IOSS interface to minimize these types of issues

sebrowne:
I think Roger's comment makes sense.  It sucks for now, but it will only suck until FY25 (hopefully).  I can live with that (speaking as not the person who will be maintaining the conflict headaches).

rppawlo (approved):

PR #12049, opened 2023-07-13
IOSS: Fix bad changes from !compare to !=

gsjaardema:
@trilinos/seacas

## Motivation
Previous snapshot had a bad commit which had mistakenly changesd some !string1.compare(string2) into `string1 == string2`.  Fixed those and added two tests that will catch this in the future.


gsjaardema:
@tokusanya @alanw0 If either of you are around could you take a quick look at this PR.  Very small this time.  Thanks.

rppawlo (approved):

gsjaardema:
@rppawlo Thanks.

PR #12052, opened 2023-07-13
Sacado: Fix subview() for LayoutContiguous<LayoutLeft> views

etphipp:
Calling subview() with LayoutContiguous<LayoutLeft> can result in  invalid indexing when the returned view is LayoutStride.  This is because the implementation assumes the right-most index is the fad dimension for all layouts except LayoutLeft, including LayoutStride.  However, if the original view was LayoutLeft, this isn't correct.  I fixed this by interchanging the dimensions in the LayoutStride case.  I also added a test to verify it is correct now by checking the indexing in numerous subview cases.

Also fixed a few uses of deprecated Rank.

Thanks to @rppawlo  for reporting this.


rppawlo (approved):
Thanks @etphipp !

PR #12058, opened 2023-07-17
MueLu: Add clang-format workflow and run it on MueLu

GrahamBenHarper:
@trilinos/muelu 

Suggested by @csiefer2 in light of his recent PRs :)

This will almost certainly break any in-progress PRs in MueLu, so this is not necessarily up for serious consideration at the moment until we see if we really like how it formats things.

## How does this work?
I ran `clang-format` version 14.0.0 and 14.0.6 from the [llvm 14.0.0 binary release](https://github.com/llvm/llvm-project/releases/tag/llvmorg-14.0.0) and from our SEMS clang 14.0.6 module on MueLu. To do so, I placed the `.clang-format` from Kokkos at the top directory of MueLu, with a couple documented additions.

Other technical details: when `clang-format` is invoked from a directory, it searches for `.clang-format`, and moves up a directory every time it does not find the format file. Once it finds the format file, it uses that file. It errors out if it does not find a format file. Therefore, we placed it in the root directory of MueLu so that any invocation within MueLu will find the format file.

## Summary

The workflow runs `clang-format` version 14.0.0 on the `packages/muelu/src` directory on only the `.cpp` and `.hpp` files. If the formatting is correct, the workflow passes. If the formatting is not correct, the workflow will report a diff for the offending lines when you click "details". Here's a screenshot of the workflow dashboard I were to purposely mess up indentation in a file. Notice it suggests the correct indentation in green.
image

## Compatibility for Users
Since it might be complicated for users to try to setup clang-format on their own, the script in `packages/muelu/utils/run_clang_format.sh` provides helpful instructions if clang-format is not detected on a system. In particular, it points users to the [llvm 14.0.0 binary release](https://github.com/llvm/llvm-project/releases/tag/llvmorg-14.0.0) and instructs them to download it if they do not have clang-format installed. Additionally, the script checks to make sure the version is 14.0.0 since there are sometimes vast differences in behavior between clang-format versions. Otherwise, just like rebasing gold files, you simply need to run `packages/muelu/utils/run_clang_format.sh` to fix any code formatting reported by the workflow. 

I think a couple people from MueLu should try to follow the directions and see if it's difficult to fix code formatting or not.


GrahamBenHarper:
@csiefer2 actually probably not. I stole it from a different code. I should try the Kokkos version instead to be more consistent.

jhux2:
@GrahamBenHarper It looks like the formatter is enforcing an 72 column width.  Is that something that can be relaxed?

GrahamBenHarper:
@jhux2 yep there's definitely some gore in heavily indented regions due to the 72 column width. There should be a way to relax things like that. I'll also check what Kokkos does

GrahamBenHarper:
I switched over to what Kokkos uses, but it still looks pretty comparable to what I saw before. I'd have to dig deeper to see how to change the allowed column width

cgcgcg:
Ideally, you want to set a code style that only flags things that are currently clearly inconsistent.
Once we settle on a code style, can you add a nightly run with it so that we get bugged when we break things? In the long term, it would be nice to add something to the PR tester let's not make this framework's problem just yet.

mayrmt:
I value the fact, that there is a code formatter, much more than the individual formatting options. Of course, I might make sense to stick to a style, that is somehow common in the Trilinos/Kokkos universe, so Kokkos style sounds very reasonable. 

I agree with @cgcgcg: ultimately, one should not be allowed to check-in style-incompatible code.

GrahamBenHarper:
@cgcgcg @mayrmt I agree. I added a `run_clang_format.sh` script in our `utils` directory (I'm open to other directory suggestions by the way) which runs the formatter on all `*.cpp` and `*.hpp` files. Unfortunately, because of the way the `clang-format` works, I have to create a temporary file and then replace the original with the temp file. If I run `clang-format -style=file FILENAME > FILENAME` it wipes the file. However, it's very easy to run and it error checks to make sure it only runs from MueLu's top level.

@trilinos/muelu I set `ColumnLimit: 0`, which shows it had created about 25k extra lines. Most of those extra lines were created by placing line breaks in simple lines like this
```
      RCP<MultiVector> vector = MultiVectorFactory::build(map,cols);
```
so I think it makes more sense to ignore the column limit for now. I haven't looked at these new changes closer yet, so I still don't know if there are other options we want to play with. Let me know if you spot something else that might be possible improve.

cgcgcg:
@GrahamBenHarper Why does `-i` not work?

GrahamBenHarper:
I'm not a big fan of the single space for class access modifiers, but this screengrab does a decent job at capturing some of the different things the formatter does. Note it removes indents for the MueLu namespace, but I think that's generally fine since almost everything is guarded by the MueLu namespace.
image

GrahamBenHarper:
And I think this is where most of those extra 3k deletions came from
image

GrahamBenHarper:
This last commit allows up to 3 consecutive empty lines, puts access modifiers on the same line as the class, and splits constructor initializer lists again. Now it's almost exactly the same number of additions and deletions, which is a good sign that it's not horribly mangling things

GrahamBenHarper:
@cgcgcg @jhux2 @trilinos/muelu 
Okay, I finally have things fixed. Here's a summary of how things work.

## Summary

The workflow runs `clang-format` version 8.0.1 on the `packages/muelu/src` directory on only the `.cpp` and `.hpp` files. If the formatting is correct, the workflow passes. If the formatting is not correct, the workflow will report a diff for the offending lines when you click "details". Here's a screenshot of the workflow dashboard I were to purposely mess up indentation in a file. Notice it suggests the correct indentation in green.
image

## Compatibility for Users
Since it might be complicated for users to try to setup clang-format on their own, the script in `packages/muelu/utils/run_clang_format.sh` provides helpful instructions if clang-format is not detected on a system. In particular, it points users to the [llvm 8.0.1 binary release](https://github.com/llvm/llvm-project/releases/tag/llvmorg-8.0.1) and instructs them to download it if they do not have clang-format installed. Additionally, the script checks to make sure the version is 8.0.1 since there are sometimes vast differences in behavior between clang-format versions. Otherwise, just like rebasing gold files, you simply need to run `packages/muelu/utils/run_clang_format.sh` to fix any code formatting reported by the workflow. 

I think a couple people from MueLu should try to follow the directions and see if it's difficult to fix code formatting or not.

Out of all 69,000 changes, only one change produced by clang-format locally disagreed with the workflow. I fixed it by adding an empty line. The two versions were confused whether `}; // SemiCoarsenPFactory` should be aligned with the closing brace or the comment on the previous line. Otherwise, it seems unlikely to run into edge cases.

GrahamBenHarper:
Also pinging @sebrowne since a workflow should probably get some framework approval.

packages/muelu/src/.clang-format
cgcgcg, line 1: Move this up one level so that tests are covered as well?

.github/workflows/muelu_clang_format.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.16.2
+      with:
+        source: './packages/muelu/src'
```
cgcgcg, line 13: Run it directly from `muelu`?

.github/workflows/muelu_clang_format.yml
```diff
@@ -0,0 +1,18 @@
...
+        exclude: './lib'
+        extensions: 'cpp,hpp'
+        clangFormatVersion: 8
```
cgcgcg, line 16: Why this version instead of anything more recent?

.github/workflows/muelu_clang_format.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.16.2
+      with:
+        source: './packages/muelu/src'
+        exclude: './lib'
```
cgcgcg, line 14: What's that exclusion for?

.github/workflows/muelu_clang_format.yml
GrahamBenHarper: How?

.github/workflows/muelu_clang_format.yml
GrahamBenHarper: So I don't forget that's an option in case we need it. I can remove it if you'd like

.github/workflows/muelu_clang_format.yml
GrahamBenHarper: Because Kokkos uses clang-format 8.0.0 (this workflow actually provides 8.0.1 when version 8 is specified)

packages/muelu/src/.clang-format
GrahamBenHarper: Good point! We can add tests to this as well.

.github/workflows/muelu_clang_format.yml
GrahamBenHarper: Oh you mean `./packages/muelu` instead of `./packages/muelu/src`?

GrahamBenHarper:
Ah, it broke the masterlist :)

.github/workflows/muelu_clang_format.yml
GrahamBenHarper: Actually, we'll need to exclude the master list, so now I have a good use for it :)

cgcgcg: Why? The master list can be formatted correctly.

GrahamBenHarper: I think it would require syncing formatting between the xml and cpp for the MasterList in a weird way. Test results are below

<details>
<summary> Failing: MueLu_MasterList </summary>

```
abstract.tex matches
acknowledgment.tex matches
appendix.tex matches
definitions.tex matches
introduction.tex matches
license.tex matches
mloptions_aggregation.tex matches
mloptions_general.tex matches
mloptions_misc.tex matches
mloptions_multigrid.tex matches
mloptions_rebalancing.tex matches
mloptions_refmaxwell.tex matches
mloptions_reuse.tex matches
mloptions_smoothing_and_coarse.tex matches
mloptions.tex matches
mueluguide.tex matches
muemex.tex matches
multigrid.tex matches
options_aggregation.tex matches
options_general.tex matches
options_misc.tex matches
options_multigrid.tex matches
options_rebalancing.tex matches
options_refmaxwell.tex matches
options_reuse.tex matches
options_smoothing_and_coarse.tex matches
options.tex matches
paramlist_hidden.tex matches
paramlist.tex matches
performance.tex matches
references.tex matches
SANDdistribution.tex matches
starting.tex matches
62d61
< 
76d74
< 
96,100c94,107
<       std::stringstream temp1; temp1 << "\"" << "MGV" << "\"";
<       std::stringstream temp2; temp2 << "\"" << "MGV" << "\"";
<       if (value == temp1.str() ) { ss << "<Parameter name=\"cycle type\" type=\"string\" value=\"V\"/>"; }
<       else if (value == temp2.str()) { ss << "<Parameter name=\"cycle type\" type=\"string\" value=\"W\"/>"; }
<       else TEUCHOS_TEST_FOR_EXCEPTION(true, MueLu::Exceptions::RuntimeError, "MasterList::interpretParameterName, Line " << __LINE__ << ". "
---
>     std::stringstream temp1;
>     temp1 << "\""
>           << "MGV"
>           << "\"";
>     std::stringstream temp2;
>     temp2 << "\""
>           << "MGV"
>           << "\"";
>     if (value == temp1.str()) {
>       ss << "<Parameter name=\"cycle type\" type=\"string\" value=\"V\"/>";
>     } else if (value == temp2.str()) {
>       ss << "<Parameter name=\"cycle type\" type=\"string\" value=\"W\"/>";
>     } else
>       TEUCHOS_TEST_FOR_EXCEPTION(true, MueLu::Exceptions::RuntimeError, "MasterList::interpretParameterName, Line " << __LINE__ << ". "
107,108c114,121
<       std::stringstream temp; temp << "\"" << "1" << "\"";
<       if (value == temp.str() ) { ss << "<Parameter name=\"multigrid algorithm\" type=\"string\" value=\"pg\"/>"; return ss.str(); }
---
>     std::stringstream temp;
>     temp << "\""
>          << "1"
>          << "\"";
>     if (value == temp.str()) {
>       ss << "<Parameter name=\"multigrid algorithm\" type=\"string\" value=\"pg\"/>";
>       return ss.str();
>     }
112c125,128
<       std::stringstream temp1; temp1 << "\"" << "1" << "\"";
---
>     std::stringstream temp1;
>     temp1 << "\""
>           << "1"
>           << "\"";
121,171c137,336
< 
<     if (name == "output filename") { ss << "<Parameter name=\"output filename\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "number of equations") { ss << "<Parameter name=\"number of equations\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "max levels") { ss << "<Parameter name=\"max levels\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "W cycle start level") { ss << "<Parameter name=\"W cycle start level\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "coarse grid correction scaling factor") { ss << "<Parameter name=\"coarse grid correction scaling factor\" type=\"double\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "fuse prolongation and update") { ss << "<Parameter name=\"fuse prolongation and update\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "number of vectors") { ss << "<Parameter name=\"number of vectors\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "problem: symmetric") { ss << "<Parameter name=\"problem: symmetric\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "hierarchy label") { ss << "<Parameter name=\"hierarchy label\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "aggregation: drop tol") { ss << "<Parameter name=\"aggregation: drop tol\" type=\"double\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "print initial parameters") { ss << "<Parameter name=\"print initial parameters\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "print unused parameters") { ss << "<Parameter name=\"print unused parameters\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: damping factor") { ss << "<Parameter name=\"sa: damping factor\" type=\"double\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: use filtered matrix") { ss << "<Parameter name=\"sa: use filtered matrix\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: eigen-analysis type") { ss << "<Parameter name=\"sa: eigen-analysis type\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: eigenvalue estimate num iterations") { ss << "<Parameter name=\"sa: eigenvalue estimate num iterations\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: use rowsumabs diagonal scaling") { ss << "<Parameter name=\"sa: use rowsumabs diagonal scaling\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: enforce constraints") { ss << "<Parameter name=\"sa: enforce constraints\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: max eigenvalue") { ss << "<Parameter name=\"sa: max eigenvalue\" type=\"double\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: rowsumabs diagonal replacement tolerance") { ss << "<Parameter name=\"sa: rowsumabs diagonal replacement tolerance\" type=\"double\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: rowsumabs use automatic diagonal tolerance") { ss << "<Parameter name=\"sa: rowsumabs use automatic diagonal tolerance\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: rowsumabs diagonal replacement value") { ss << "<Parameter name=\"sa: rowsumabs diagonal replacement value\" type=\"double\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "sa: rowsumabs replace single entry row with zero") { ss << "<Parameter name=\"sa: rowsumabs replace single entry row with zero\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "replicate: npdes") { ss << "<Parameter name=\"replicate: npdes\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "combine: numBlks") { ss << "<Parameter name=\"combine: numBlks\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "pcoarsen: element") { ss << "<Parameter name=\"pcoarsen: element\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "pcoarsen: schedule") { ss << "<Parameter name=\"pcoarsen: schedule\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "pcoarsen: hi basis") { ss << "<Parameter name=\"pcoarsen: hi basis\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "pcoarsen: lo basis") { ss << "<Parameter name=\"pcoarsen: lo basis\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "smoother: neighborhood type") { ss << "<Parameter name=\"smoother: neighborhood type\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "tentative: calculate qr") { ss << "<Parameter name=\"tentative: calculate qr\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "tentative: constant column sums") { ss << "<Parameter name=\"tentative: constant column sums\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "repartition: enable") { ss << "<Parameter name=\"repartition: enable\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "repartition: start level") { ss << "<Parameter name=\"repartition: start level\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "repartition: use map") { ss << "<Parameter name=\"repartition: use map\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "repartition: use subcommunicators in place") { ss << "<Parameter name=\"repartition: use subcommunicators in place\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "repartition: node repartition level") { ss << "<Parameter name=\"repartition: node repartition level\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "repartition: node id") { ss << "<Parameter name=\"repartition: node id\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "repartition: min rows per proc") { ss << "<Parameter name=\"repartition: min rows per proc\" type=\"int\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "repartition: max imbalance") { ss << "<Parameter name=\"repartition: max imbalance\" type=\"double\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "use external multigrid package") { ss << "<Parameter name=\"use external multigrid package\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "maxwell1: dump matrices") { ss << "<Parameter name=\"maxwell1: dump matrices\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "refmaxwell: mode") { ss << "<Parameter name=\"refmaxwell: mode\" type=\"string\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "refmaxwell: disable addon") { ss << "<Parameter name=\"refmaxwell: disable addon\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "refmaxwell: use as preconditioner") { ss << "<Parameter name=\"refmaxwell: use as preconditioner\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "refmaxwell: dump matrices") { ss << "<Parameter name=\"refmaxwell: dump matrices\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "refmaxwell: subsolves on subcommunicators") { ss << "<Parameter name=\"refmaxwell: subsolves on subcommunicators\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "refmaxwell: enable reuse") { ss << "<Parameter name=\"refmaxwell: enable reuse\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "refmaxwell: skip first (1,1) level") { ss << "<Parameter name=\"refmaxwell: skip first (1,1) level\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
<     if (name == "refmaxwell: normalize nullspace") { ss << "<Parameter name=\"refmaxwell: normalize nullspace\" type=\"bool\" value=" << value << "/>"; return ss.str(); }      
---
>   if (name == "output filename") {
>     ss << "<Parameter name=\"output filename\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "number of equations") {
>     ss << "<Parameter name=\"number of equations\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "max levels") {
>     ss << "<Parameter name=\"max levels\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "W cycle start level") {
>     ss << "<Parameter name=\"W cycle start level\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "coarse grid correction scaling factor") {
>     ss << "<Parameter name=\"coarse grid correction scaling factor\" type=\"double\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "fuse prolongation and update") {
>     ss << "<Parameter name=\"fuse prolongation and update\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "number of vectors") {
>     ss << "<Parameter name=\"number of vectors\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "problem: symmetric") {
>     ss << "<Parameter name=\"problem: symmetric\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "hierarchy label") {
>     ss << "<Parameter name=\"hierarchy label\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "aggregation: drop tol") {
>     ss << "<Parameter name=\"aggregation: drop tol\" type=\"double\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "print initial parameters") {
>     ss << "<Parameter name=\"print initial parameters\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "print unused parameters") {
>     ss << "<Parameter name=\"print unused parameters\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: damping factor") {
>     ss << "<Parameter name=\"sa: damping factor\" type=\"double\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: use filtered matrix") {
>     ss << "<Parameter name=\"sa: use filtered matrix\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: eigen-analysis type") {
>     ss << "<Parameter name=\"sa: eigen-analysis type\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: eigenvalue estimate num iterations") {
>     ss << "<Parameter name=\"sa: eigenvalue estimate num iterations\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: use rowsumabs diagonal scaling") {
>     ss << "<Parameter name=\"sa: use rowsumabs diagonal scaling\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: enforce constraints") {
>     ss << "<Parameter name=\"sa: enforce constraints\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: max eigenvalue") {
>     ss << "<Parameter name=\"sa: max eigenvalue\" type=\"double\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: rowsumabs diagonal replacement tolerance") {
>     ss << "<Parameter name=\"sa: rowsumabs diagonal replacement tolerance\" type=\"double\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: rowsumabs use automatic diagonal tolerance") {
>     ss << "<Parameter name=\"sa: rowsumabs use automatic diagonal tolerance\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: rowsumabs diagonal replacement value") {
>     ss << "<Parameter name=\"sa: rowsumabs diagonal replacement value\" type=\"double\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "sa: rowsumabs replace single entry row with zero") {
>     ss << "<Parameter name=\"sa: rowsumabs replace single entry row with zero\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "replicate: npdes") {
>     ss << "<Parameter name=\"replicate: npdes\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "combine: numBlks") {
>     ss << "<Parameter name=\"combine: numBlks\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "pcoarsen: element") {
>     ss << "<Parameter name=\"pcoarsen: element\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "pcoarsen: schedule") {
>     ss << "<Parameter name=\"pcoarsen: schedule\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "pcoarsen: hi basis") {
>     ss << "<Parameter name=\"pcoarsen: hi basis\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "pcoarsen: lo basis") {
>     ss << "<Parameter name=\"pcoarsen: lo basis\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "smoother: neighborhood type") {
>     ss << "<Parameter name=\"smoother: neighborhood type\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "tentative: calculate qr") {
>     ss << "<Parameter name=\"tentative: calculate qr\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "tentative: constant column sums") {
>     ss << "<Parameter name=\"tentative: constant column sums\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "repartition: enable") {
>     ss << "<Parameter name=\"repartition: enable\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "repartition: start level") {
>     ss << "<Parameter name=\"repartition: start level\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "repartition: use map") {
>     ss << "<Parameter name=\"repartition: use map\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "repartition: use subcommunicators in place") {
>     ss << "<Parameter name=\"repartition: use subcommunicators in place\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "repartition: node repartition level") {
>     ss << "<Parameter name=\"repartition: node repartition level\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "repartition: node id") {
>     ss << "<Parameter name=\"repartition: node id\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "repartition: min rows per proc") {
>     ss << "<Parameter name=\"repartition: min rows per proc\" type=\"int\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "repartition: max imbalance") {
>     ss << "<Parameter name=\"repartition: max imbalance\" type=\"double\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "use external multigrid package") {
>     ss << "<Parameter name=\"use external multigrid package\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "maxwell1: dump matrices") {
>     ss << "<Parameter name=\"maxwell1: dump matrices\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "refmaxwell: mode") {
>     ss << "<Parameter name=\"refmaxwell: mode\" type=\"string\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "refmaxwell: disable addon") {
>     ss << "<Parameter name=\"refmaxwell: disable addon\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "refmaxwell: use as preconditioner") {
>     ss << "<Parameter name=\"refmaxwell: use as preconditioner\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "refmaxwell: dump matrices") {
>     ss << "<Parameter name=\"refmaxwell: dump matrices\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "refmaxwell: subsolves on subcommunicators") {
>     ss << "<Parameter name=\"refmaxwell: subsolves on subcommunicators\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "refmaxwell: enable reuse") {
>     ss << "<Parameter name=\"refmaxwell: enable reuse\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "refmaxwell: skip first (1,1) level") {
>     ss << "<Parameter name=\"refmaxwell: skip first (1,1) level\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
>   if (name == "refmaxwell: normalize nullspace") {
>     ss << "<Parameter name=\"refmaxwell: normalize nullspace\" type=\"bool\" value=" << value << "/>";
>     return ss.str();
>   }
395,398c560,561
< "</ParameterList>"
< ;
<   std::map<std::string,std::string> MasterList::DefaultProblemTypeLists_ = DefaultProblemStrings<std::string,std::string>
< ("Poisson-2D",
---
>     "</ParameterList>";
> std::map<std::string, std::string> MasterList::DefaultProblemTypeLists_ = DefaultProblemStrings<std::string, std::string>("Poisson-2D",
408,410c571
<     "</ParameterList>"
<   )
< ("Poisson-2D-complex",
---
>                                                                                                                           "</ParameterList>")("Poisson-2D-complex",
426,428c587
<     "</ParameterList>"
<   )
< ("Poisson-3D",
---
>                                                                                                                                               "</ParameterList>")("Poisson-3D",
438,440c597
<     "</ParameterList>"
<   )
< ("Poisson-3D-complex",
---
>                                                                                                                                                                   "</ParameterList>")("Poisson-3D-complex",
456,458c613
<     "</ParameterList>"
<   )
< ("Elasticity-2D",
---
>                                                                                                                                                                                       "</ParameterList>")("Elasticity-2D",
468,470c623
<     "</ParameterList>"
<   )
< ("Elasticity-2D-complex",
---
>                                                                                                                                                                                                           "</ParameterList>")("Elasticity-2D-complex",
486,488c639
<     "</ParameterList>"
<   )
< ("Elasticity-3D",
---
>                                                                                                                                                                                                                               "</ParameterList>")("Elasticity-3D",
498,500c649
<     "</ParameterList>"
<   )
< ("Elasticity-3D-complex",
---
>                                                                                                                                                                                                                                                   "</ParameterList>")("Elasticity-3D-complex",
516,518c665
<     "</ParameterList>"
<   )
< ("MHD",
---
>                                                                                                                                                                                                                                                                       "</ParameterList>")("MHD",
552,554c699
<     "</ParameterList>"
<   )
< ("ConvectionDiffusion",
---
>                                                                                                                                                                                                                                                                                           "</ParameterList>")("ConvectionDiffusion",
574,576c719
<     "</ParameterList>"
<   )
< ;
---
>                                                                                                                                                                                                                                                                                                               "</ParameterList>");
1007,1010c1150
<          ("refmaxwell: normalize nullspace","refmaxwell: normalize nullspace")
<       ;
< 
< }
---
>                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             ("refmaxwell: normalize nullspace", "refmaxwell: normalize nullspace");
1011a1152
> }  // namespace MueLu
MueLu_MasterList.cpp does not match
Do not edit MueLu_MasterList.cpp directly, but edit masterList.xml and run update_params.sh
a non-zero exit code. Per user-direction, the job has been aborted.
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[6952,1],0]
  Exit code:    1
```
</details>

cgcgcg: You could just take what clang format tells you and apply that to the template that is used to generate the master list.

GrahamBenHarper: I just tried it out, and it's easy enough to fix the header part of the template, but the part of the MasterList that is handled by
`xsltproc gen_interpreter.xsl masterList.xml >> $code_file` has a very specific format it generates, and it conflicts with what clang-format wants.

.github/workflows/muelu_clang_format.yml
cgcgcg: Does it really matter to have the same version as Kokkos? We should probably choose a version that is available on SEMS and from most modern distributions.

GrahamBenHarper: I mean, we're both in Trilinos, so I'd hope that things are at least slightly similar/compatible.

The main problem is that clang-format is totally dependent on the version. It's horribly incompatible with itself. For example, if I run clang-format version 14.0.6 provided by SEMS, it disagrees with this PR resulting in
`523 files changed, 3841 insertions(+), 3836 deletions(-)`
Also, we can't guarantee that SEMS will support the same version of clang forever. Once the `sems-clang/14.0.6` module is gone, we're stuck in the same boat of running an unsupported version of `clang-format` anyway. What would we do in that case? Reformat all of MueLu again to agree with the supported module du jour?

Maybe I'm being a bit cynical, but while clang-format is great, different versions produce vastly different results, which is the main reason why I think we should point to some official LLVM release and not rely on a module, even though it may be slightly more difficult.

cgcgcg: If different versions produce different output we should just pick the most recent version in the hope that we can stick to it for the longest time possible and that there will be fewer and fewer changes for future versions.

GrahamBenHarper: Okay. The latest version of LLVM that is released is 17.0.3, but the most recent version supported by SEMS TPLs is 14.0.6 (11.0.1 is also supported). However, the Github action only currently supports the following:
```
clang-format version 5.0.2 
clang-format version 6.0.1 
clang-format version 7.1.0 
clang-format version 8.0.1 
clang-format version 9.0.0
clang-format version 10.0.0
clang-format version 11.0.0
clang-format version 11.1.0
clang-format version 12.0.0
clang-format version 12.0.1
clang-format version 13.0.0
clang-format version 14.0.0
clang-format version 15.0.2
clang-format version 16.0.0
clang-format version 16.0.3
```
I could try to submit a PR to the workflow to add support for clang-format version 14.0.6 if there's a difference between 14.0.0 and 14.0.6 in MueLu.

cgcgcg: Let's table this and talk about it during the standup tomorrow.

GrahamBenHarper:
I spoke a little bit more with @dalg24 after TUG today, and he suggests moving to the most recent version we can support, like clang14 or 16. Since clang14 is the most recent module supported by the SEMS modules we use, it makes the most sense to try to use clang-format version 14. He said Kokkos will also likely upgrade to version 14 in the future to sync with us. He also pointed out that a lot of other things such as grepping for code patterns and building regular expressions become much easier when a clang format is enforced in a package.

There are some tools like [clang-format-diff](https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/clang-format-diff.py) which will analyze only the touched files for diffs, but I'm not sure at the moment how it would work with Trilinos. It seems like our MueLu-only workflow runs in a minute, which isn't too terrible, but might be something to improve.

Regardless of which code editor you use, there are [emacs plugins](https://github.com/SavchenkoValeriy/emacs-clang-format-plus) and [vim plugins](https://github.com/rhysd/vim-clang-format) which can make these formatting processes easier by simply applying a .clang-format file when you save files. This would also make it painless to edit files and still enforce some sort of standard.

Lastly, if you can't be bothered with any of that, you can simply copy-paste the reported diff from your PR's workflow into a `diff.txt` file in your terminal and run `git apply diff.txt` to fix the formatting, meaning it's still extremely easy :)

cgcgcg:
Just add the creation of a patch to the github action and upload it as an artifact. We talked about this on Monday.

GrahamBenHarper:
@cgcgcg oh I didn't hear that part about creating an artifact in standup; I'll look into that before we meet tomorrow!

.github/workflows/muelu_clang_format.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.16.2
+      with:
+        source: './packages/muelu'
+        exclude: './packages/muelu/src/MueCentral/MueLu_MasterList.cpp'
```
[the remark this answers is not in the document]
GrahamBenHarper: oh I didn't know that was possible! Does the position relative to the copyright matter, or can I place it after the copyright?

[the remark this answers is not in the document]
dalg24: It does not matter.  That said I would do after the copyright so that all your source files have the exact same first few lines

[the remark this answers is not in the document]
GrahamBenHarper: I agree! That's exactly why I asked :)

GrahamBenHarper:
It seems like the `PR-12058-test-rhel7_sems-cuda-11.4.2-sems-gnu-10.1.0-sems-openmpi-4.0.5_release_static_Volta70_no-asan_complex_no-fpic_mpi_pt_no-rdc_no-uvm_deprecated-on_no-package-enables-2712` on ascicgpu050 timed out?

sebrowne (commented):
Fine by me from an actions perspective.

.github/workflows/muelu_clang_format.yml
```diff
@@ -0,0 +1,29 @@
+name: Check MueLu clang-format
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
cgcgcg, line 10: Update to version v4 to avoid the warning?

GrahamBenHarper: let's do v3 since that's what the other workflows use for uniformity

cgcgcg (approved):
I'm cool with this.
@trilinos/muelu Speak now or forever hold your peace.

jhux2:
Can you give us mere mortals a few days to have a look?

mayrmt (approved):

GrahamBenHarper:
Sure @jhux2, please let me know if you have any questions about it!

lucbv:
Fine with me

jhux2 (approved):
i   aPprove of
   this P R .

cgcgcg:
For future reference:
Put 
```yaml
files: "packages/muelu/.*pp"
repos:
-   repo: https://github.com/pre-commit/mirrors-clang-format
    rev: v14.0.1
    hooks:
    - id: clang-format
      types_or: [c++, c, cuda]
``` 
in `.pre-commit-config.yaml` in your Trilinos source root.
Then
```
python3 -m pip install pre-commit
pre-commit install
```

PR #12071, opened 2023-07-20
SEACAS: decomp, nem_slice better handling of no-zoltan case

gsjaardema:
@trilinos/seacas

## Motivation
If Trilinos is built without zoltan enabled, then the default decomposition method in the decomp script is not valid. 
Made the default be `rib` if zoltan present and `inertial` if zoltan isn't available.  
Fixed nem_slice to report an error if the selected decomposition method is not available in the current build instead of just falling through and giving an incomprehensible error.


gsjaardema:
There are also some changes in the elb_loadbal.C file due to rerunning clang-format.

rppawlo (approved):

gsjaardema:
@rppawlo @alanw0 would one of you have a chance to re-review this PR.  Thanks.

rppawlo (approved):

PR #11009, opened 2022-09-10
Intrepid2 Tetrahedral Hierarchical Bases

CamelliaDPG:
@trilinos/intrepid2

## Motivation
This PR fills in missing H(curl), H(div), and H(vol) hierarchical bases on the tetrahedron.  This follows #10662, which did the same for the triangle.

Forthcoming PRs will similarly add missing hierarchical bases on the wedge, and pyramid.

## Testing
Some basic "basis equivalence" tests, which establish that the hierarchical bases span the same space as existing nodal bases, are included here.  Additionally, I have done offline comparison with the ESEAS implementation of each basis, testing up to 8th order, with excellent agreement.  (One exception is that for p=5,6,7,8, the curls of certain functions disagree with ESEAS.  Given that the values agree with ESEAS (establishing that our dof numbering agrees), and given that the basis equivalence tests pass for these orders (establishing that the curls and values are consistent in the new implementation), I believe that this is either an issue with ESEAS, or something wrong with the interface I have set up to ESEAS for the comparison.  I have made some offline notes on which degrees of freedom are implicated; if I do find a bug in ESEAS or "smoking gun" evidence of incorrect curls there, I will communicate that to the ESEAS developers.)

PR #12079, opened 2023-07-26
Intrepid2: Add H^1 Hierarchical Pyramid Basis

CamelliaDPG:
## Motivation
This PR adds an arbitrary-order, hierarchical basis for H(grad) on the pyramid.  This follows #11061, which added hierarchical bases for the wedge, and  #11009, which did so for the tetrahedron.  We will be adding a hierarchical L^2 basis for the pyramid soon after this, followed by H(div) and H(curl).

We also add support for generating equispaced point lattices on the pyramid.

## Testing
A basic "basis equivalence" test, which establishes that the new basis spans the same space as the existing low-order nodal basis, is included here.  Additionally, I have done offline comparison with the ESEAS implementation of this basis, testing up to 9th order, with excellent agreement.  I hope to include these tests with Intrepid2 soon; this has been prevented previously by the license for ESEAS, but they are changing the license to one that will allow inclusion.


mperego (approved):
@CamelliaDPG this looks good to me! We should add more testing though, including tests for orientations and projections. I'm fine to address this in a subsequent PR once you have implemented the HCURL and HDIV basis functions.

CamelliaDPG:
Thanks, @mperego!  Yes, agreed on the testing.  I realized just now that I should also have some tests against the assignment of basis functions to subcells: that's implemented, but almost completely untested.  (The "basis equivalence" test provides some tests of the correctness of the assignment, but the only basis we have to compare with is the lowest-order basis, so we're just checking the vertex dof assignment.  The offline ESEAS comparison tests confirm the correctness of the basis values and gradients, but do not do anything with the topological data.)  For all the other function spaces, we have no equivalent bases to compare against, so as I implement those, I'll add some tests for the subcell associations, including the H^1 basis here.

PR #12086, opened 2023-07-30
panzer: workset: fix cell_to_face initialization bug

tkordenbrock:
The cell_to_face device view was initialized to -1 then mirrored but not copied to host.  The host mirror was populated and then copied to device overwriting the initialized device view.  This works with UVM on but not with UVM off.  Now instead of initializing the device view, the host mirror is initilialized, populated and copied to device.

@trilinos/panzer


rppawlo (approved):
Thanks @tkordenbrock !

PR #12094, opened 2023-08-01
Framework: Post 14.4 release update

cabejackson:
@trilinos/framework 

## Motivation
Trilinos 14.4 release


sebrowne (changes requested):
You're going to want to set the new version on develop to 14.5 (and 140500).  Also wait until the release is done to merge this one.

ccober6:
> You're going to want to set the new version on develop to 14.5 (and 140500). Also wait until the release is done to merge this one.

If we do another minor release it would be 14.6.

sebrowne:
> > You're going to want to set the new version on develop to 14.5 (and 140500). Also wait until the release is done to merge this one.
> 
> If we do another minor release it would be 14.6.

Correct, but that variable appears to track the odd version immediately following the current release.  Meaning, if we just released 14.4, 'develop' would version as 14.5, and then update to 14.7 if we release a 14.6.

ccober6:
Gotcha.  Thanks for the clarification!

sebrowne (approved):
Wait until the release is completed to merge this PR

sebrowne:
This is the same SEACAS test that sporadically failed on some other toolchains for my PR earlier in the week.  There is no way that it is related to this PR.  @cabejackson you can disable it as part of this PR, which is unfortunate but acceptable.

cabejackson:
New PR can be found [here](https://github.com/trilinos/Trilinos/pull/12101)

PR #12100, opened 2023-08-02
MueLu: Add CMake logic to simplify installation of drivers

cgcgcg:
@trilinos/muelu 

## Motivation
The added piece of CMake allows to install an executable and additional files needed to run it. The relative path in the installation matches the relative path within the source tree.


packages/muelu/cmake/MueLuInstall.cmake
```diff
@@ -0,0 +1,59 @@
...
+    ""
+    ""
+    "ADDITIONAL_FILES"
+    ${ARGN}
+  )
+
+  # set the install directory to the current source folder, but relative to install prefix
+  set(INSTALL_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+  file(RELATIVE_PATH INSTALL_DIRECTORY ${CMAKE_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
+  set(INSTALL_DIRECTORY ${CMAKE_INSTALL_PREFIX}/${INSTALL_DIRECTORY})
```
bartlettroscoe, line 51: So this would install under:

```
<prefix>/packages/muelu/example/basic/
```

for example, in the call from:

```
<trilinosSrcDir>/packages/muelu/example/basic/CMakeLists.txt
```

?

Would in not be better to install these under:

```
<prefix>/examples/muelu/basic/
```

?

That would make it clear that the user will only find examples under `<prefix>/examples/`.  Otherwise, they might thing they will find packages under `<prefix>/packages/` (for whatever "packages" might mean to them).

This is just my opinion but it does impact the overall understandability of the Trilinos install tree so this should be reviewed by the Framework team and other product area leads as well.

packages/muelu/cmake/MueLuInstall.cmake
```diff
@@ -0,0 +1,59 @@
...
+
+  # install executable
+  INSTALL(TARGETS "${PACKAGE_NAME}_${TARGET}" COMPONENT ${PACKAGE_NAME} RUNTIME DESTINATION ${INSTALL_DIRECTORY})
```
bartlettroscoe, line 54: One thing to take note of is that if the user selects shared libs but not setting `RPATH`, then these installed exectuables will not run without setting `LD_LIBRARY_PATH`.

bartlettroscoe (commented):
@cgcgcg, this looks like it should work.

But I have a few questions:

1. Do we really want to be installing examples under `<prefix>/packages/muelu/...`?  Would it not be better to tell users right up front that these are all examples by using the subdir `<prefix>/examples/...`?  (See my comment below on different options).

2. What type of clues does the user have in running these installed example exectuables?  For instance, the install of the example executable `MueLu_Stratimikos` from `muelu/example/basic/CMakeLists.txt` must be run with the arguments `--xml=stratimikos_ParameterList2.xml --linAlgebra=Epetra` as shown [here](https://github.com/trilinos/Trilinos/blob/ee9acd882acc48b6219b1161b20251945b33f214/packages/muelu/example/basic/CMakeLists.txt#L94) but how would a user know that from just having the raw binary exectuable?  They could run it with `--help` and might be able to figure out the `--xml` arguments but how would they know that they also had to pass in the arguments `-linAlgebra=Epetra` or `--linAlgebra=Tpetra` without looking at that `muelu/example/basic/CMakeLists.txt `file (which they can't even see)?  And the `--help` option will return a **lot** of options as shown [here](https://github.com/trilinos/Trilinos/blob/ee9acd882acc48b6219b1161b20251945b33f214/packages/muelu/example/basic/Stratimikos.cpp#L114-L132).

3. How would this get tested to make sure these installed examples work correctly?  (It would be terrible to install examples and have them be broken because some files were missing or there was some other issue.)  If we can't test a feature we should likely not provide that feature (because it will break).

I think this PR should be reviewed by others as well to approve the directory structure for installing examples and also the issue of not being able to test the installed examples.

IMHO, it would be better to install CMake projects that call `find_package(MueLu)` and will build and run the examples (with CTest).  That approach can be tested in an automated way in the package `TrilinosInstallTests`.

packages/muelu/cmake/MueLuInstall.cmake
cgcgcg: That's a good point!

packages/muelu/cmake/MueLuInstall.cmake
cgcgcg: @bartlettroscoe I don't think I really care where it gets installed, as long as there is some sensible pattern to it. I like the current approach, as it makes it easy to figure out where the corresponding source lives.

cgcgcg:
@bartlettroscoe I like the installation of a small CMake project. However, I am not sure how this would interact with spack. Would we then run the install for this additional project from the Trilinos package? 

I agree, the installed code is currently not tested in its installed location, only within the build tree. So we could install incomplete examples.

packages/muelu/cmake/MueLuInstall.cmake
bartlettroscoe: Right, but the base level install directory structure should be agreed upon by Trilinos as a whole.  Just having a `<prefix>/packagess/` directory show up should be decided upon at the Trilinos level.  And there is also a common-look and feel issue if package's don't do things in a consistent way.

cgcgcg: Yep, fully agree.

bartlettroscoe:
CC: @jwillenbring 

> @bartlettroscoe I like the installation of a small CMake project. However, I am not sure how this would interact with spack. Would we then run the install for this additional project from the Trilinos package?

The Spack trilinos/package.py file can define the install test anyway it would like.  Take a look at:

https://github.com/spack/spack/blob/ff30efcebcd3e234d6e095a96059e7b23a9979d2/var/spack/repos/builtin/packages/trilinos/package.py#L929-L956

Clearly that is not currently working (because Makefile.export.Trilinos is not supported anymore) but you can see you can fill in that install step anyway you would like.

> I agree, the installed code is currently not tested in its installed location, only within the build tree. So we could install incomplete examples.

If you can't get it to build, at the very least you can install some simple shell script (or Python driver if you want to be portable on all platforms) that will run examples for the supported use cases.  That is what we did for the CASL project to test the installs.

bartlettroscoe:
@jwillenbring and @cgcgcg, there likely needs to be a higher level discussion about how Trilinos should be doing install testing with Spack.

PR #12109, opened 2023-08-07
Converted Panzer to the new STK simple_fields workflow

djglaze:
This essentially removes all usage of the template parameters beyond just the datatype in `stk::mesh::Field`s.  This makes setting the size of a `Field` a run-time configuration option through calls to `stk::mesh::put_field_on_mesh()` and keeps the datatype (double, int, etc.) a compile-time configuration option.  Functions like `stk::mesh::MetaData::declare_field()` and `stk::mesh::MetaData::get_field()` now just pass the datatype as a template parameter instead of the whole templated `Field` type itself.

The `stk::mesh::MetaData::use_simple_fields()` method is called early in `MetaData` construction to force the new behavior, to prevent any accidental regressions and to ensure that any auto-registered `Field`s (like coordinates) will be sized consistently by things like `stk::mesh::StkMeshIoBroker`.  Any calls to the legacy functions will trigger a run-time error.

There should be no user-visible behavior changes, and external applications that use `Panzer` should not need to make any changes as long as they consistently use `Panzer`'s typedefs to refer to higher-dimensional `Field`s like `panzer_stk::VectorFieldType`.  If external apps grab a `stk::mesh::MetaData` and register or manipulate their own `Field`s, there is a chance that they will encounter a new error.  The error message should guide them to the correct solution.

### DIFF
--- a/packages/panzer/adapters-stk/src/evaluators/Panzer_STK_GatherFields_impl.hpp
+++ b/packages/panzer/adapters-stk/src/evaluators/Panzer_STK_GatherFields_impl.hpp
@@ -90,7 +90,7 @@ postRegistrationSetup(typename Traits::SetupData /* d */,
   for (std::size_t fd = 0; fd < gatherFields_.size(); ++fd) {
     std::string fieldName = gatherFields_[fd].fieldTag().name();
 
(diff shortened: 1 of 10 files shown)


rppawlo (approved):
Thanks for the updates @djglaze !

PR #12111, opened 2023-08-07
Framework: Prohibit MPI_COMM_WORLD usages

sebrowne:
Disallow MPI_COMM_WORLD in new/changed code unless a specific comment identifier is added.

## Motivation

Helps clean up usages of MPI_COMM_WORLD.


sebrowne:
@ccober6 @jwillenbring I think this PR does what we were discussing with respect to MPI_COMM_WORLD

.github/workflows/probited-strings.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: echo '${{ steps.file_changes.outputs.files}}'
+
+      # This step uses the `run` command to execute a script that is stored in the repository at `script/rendered-content-link-checker.mjs` and passes all the parameters it needs to run.
+      - name: Find illegal MPI_COMM_WORLD strings
+        run: |
+          files=$(grep -l MPI_COMM_WORLD ${{steps.file_changes.outputs.files}} | grep packages/)
+          if [ -n "${files}" ]
+          then
+            echo -e "\n#########################################################################"
+            echo -e "The following changed files contain MPI_COMM_WORLD, which is not allowed:"
```
ccober6, line 49: Can you add "which is not allowed (xSDK policy M3):"?

bartlettroscoe:
FYI, this check must exclude the tests directories where you need to be setting MPI_COMM_WORLD to be able to run MPI-based tests.

ccober6:
@sebrowne and @bartlettroscoe, what is the easiest way for NGA to get their PR (https://github.com/NexGenAnalytics/Trilinos/pull/73) into the Trilinos repo?  New Trilinos PR?  Take over this one?

@stmcgovern, is there a way to filter out the tests and examples directories?  @sebrowne and @bartlettroscoe, do you have some insight here?

stmcgovern:
> @sebrowne and @bartlettroscoe, what is the easiest way for NGA to get their PR ([NexGenAnalytics#73](https://github.com/NexGenAnalytics/Trilinos/pull/73)) into the Trilinos repo? New Trilinos PR? Take over this one?
> 
> @stmcgovern, is there a way to filter out the tests and examples directories? @sebrowne and @bartlettroscoe, do you have some insight here?

@ccober6 this is done in https://github.com/NexGenAnalytics/Trilinos/pull/73. I suggest that the file https://github.com/NexGenAnalytics/Trilinos/pull/73/files#diff-6b3e92373443a8676159ed906dbfe6433db1de1be94c596cf49258957ec38538 replace that of the current yml in this PR.

bartlettroscoe:
> @sebrowne and @bartlettroscoe, what is the easiest way for NGA to get their PR ([NexGenAnalytics#73](https://github.com/NexGenAnalytics/Trilinos/pull/73)) into the Trilinos repo? New Trilinos PR? Take over this one?

I would say to post a PR for the branch in NexGenAnalytics#73 to this PR's branch [sebrowne-patch-3](https://github.com/trilinos/Trilinos/tree/sebrowne-patch-3) (which is called a [stacked PR](https://blog.logrocket.com/using-stacked-pull-requests-in-github/)) (and then merge it to this PR when ready).

> @stmcgovern, is there a way to filter out the tests and examples directories? @sebrowne and @bartlettroscoe, do you have some insight here?

I would suggest that if `/test/`, `/tests/`, `/unit-tests/`, /example/`, `/examples/` or common variations of these appear in the directory name, then skip the check. Also, for rare cases where `MPI_COMM_WORLD` is allowed to appear in library code, put in a mechanism to ignore either a block of code in a file or the entire file.  (For example, there are valid usages of `MPI_COMM_WORLD` in [Teuchos_GlobalMPISession.cpp](https://github.com/trilinos/Trilinos/blob/develop/packages/teuchos/core/src/Teuchos_GlobalMPISession.cpp) that need to be allowed.  (That is code that makes it easy to create an MPI executable program that is used for a lot of tests and examples.)

One proposal would be to look for special comments like `CHECK: ALLOW MPI_COMM_WORLD` on lines containing `MPI_COMM_WORLD` like

https://github.com/trilinos/Trilinos/blob/6a010be957bb58861fdb44e43032fab778f5d8b3/packages/teuchos/core/src/Teuchos_GlobalMPISession.cpp#L215

which would become something like:

```
    MPI_Abort(MPI_COMM_WORLD, MPI_ERR_UNKNOWN);  // CHECK: ALLOW MPI_COMM_WORLD
```

If the tool sees a line like that, it would allow `MPI_COMM_WORLD`.

Also, in my opinion, you would these checks to be runnable as tests so developers will see the violations **before** they post their PR.  For example, see [here](https://github.com/TriBITSPub/TriBITS/blob/master/test/codespell/CMakeLists.txt).

If the check is only run in a full PR build, that is not idea (IMHO).

sebrowne:
I don't have a strong opinion, other than somewhere that we can add other such pre-test scripts that are run as part of CI testing.  The proposed location sounds fine to me.  Adding it to this PR is also fine by me, then I (or they) can modify the yaml file to just call that script (I think is the idea?)

ccober6:
> I don't have a strong opinion, other than somewhere that we can add other such pre-test scripts that are run as part of CI testing. The proposed location sounds fine to me. Adding it to this PR is also fine by me, then I (or they) can modify the yaml file to just call that script (I think is the idea?)

@stmcgovern 

Ok, we will go with commonTools/test/utils for now.  Yes, NGA has added the call to the yaml file.

sebrowne:
I think this is ready to be merged if we're happy with the outcome.  I changed the name of the check to be 'codestyle', open to alternatives.

sebrowne:
Ready to merge this if we're happy with it

ccober6:
@sebrowne will this be immediately active after the merge?  If so, we should probably get out an announcement with a description of the new feature.

sebrowne:
I believe so.  It will either take effect once it's merged to develop or once that is merged to master.

bartlettroscoe:
> The workflow runs on pull request (without any branch specified, but that could be done). So it would run in develop when merged. Same for master when it gets there Cf: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request

Just a suggestion, but would we not also set up a CTest test so that the developer could get this test run on their enabled packages so they could see there is a problem before the PR tests run?

ccober6:
@sebrowne, would it be difficult to add what @bartlettroscoe suggested?

sebrowne:
> @sebrowne, would it be difficult to add what @bartlettroscoe suggested?

Not really.  It would mandate that you have git, and that you know the revisions.  It would work best on unstaged local changes.  But if we're going to do that, why even bother with the actions check?  I think it makes most sense as an actions check only, since it's intrinsically tied to the revision identifiers and thus to which branch merging is being requested.

If we're going to add it as a ctest test, I would recommend that it only applies to unstaged local changes to aid in developer productivity without adding undue complexity.

bartlettroscoe:
So the current tool only looks at changed lines in changed files?

https://github.com/trilinos/Trilinos/blob/561cb2e6da5b69878a865166b03d3bc0d5d94ab7/commonTools/test/utilities/check-mpi-comm-world-usage.py#L80-L82

To run that test as a local CTest test, you would need a reference commit to compare to.  That is not so easy to get in the general case.

One option is to just have the local CTest version of this run on all of the files in all of the enabled packages.  If that does not take super long it is likely no more expensive than many of the existing Trilinos tests.

JacobDomagala:
The python script can be used locally without specifying commits. By default it will compare your last commit with origin/develop. Uncommited changed should also be detected.

sebrowne:
That does assume that all work is targeting develop, which is probably the most frequent case.  Do we still want to add a ctest test that calls that script then?  I can't shake the feeling that that is not right.  Do people normally add tests for clang-format, include-what-you-use, etc?  Every software project I've been a part of does that through hook-based approaches or separate CI checks from the main tests.

bartlettroscoe:
> I can't shake the feeling that that is not right. Do people normally add tests for clang-format, include-what-you-use, etc?

Typically, it is a git commit hook to run clang-format checks and the such.

> Every software project I've been a part of does that through hook-based approaches or separate CI checks from the main tests.

Right, but Trilinos developers have objected to mandatory git commit hooks.

In my view, running these checks along with other tests with CTest is a reasonable compromise if developers are not willing to use git commit hooks (which need to be installed and maintained automatically if they are going to be used consistently).

bartlettroscoe:
> The python script can be used locally without specifying commits. By default it will compare your last commit with origin/develop. Uncommited changes should also be detected.

@stmcgovern, as long as you could specify a different reference version (I call the main Trilinos GitHub repo `github` instead of `origin` because  I don't lots of multi-repo development workflows), this would work fine.  So add and argument to the Python script (e.g. `--reference-version=<version-spec>`) and a CMake cache var that sets that for the CTest test (e.g. `Trilinos_GIT_UPSTREAM_REFERENCE_SPEC`) would do the trick and be about perfect I guess (because it is exactly the same check that would be done by the GHA job).

sebrowne:
> In my view, running these checks along with other tests with CTest is a reasonable compromise if developers are not willing to use git commit hooks (which need to be installed and maintained automatically if they are going to be used consistently).

Isn't the compromise that they have to wait on the PR check to run?  Meaning, if they don't want to implement user-side hooks, they do not get as immediate of feedback?

bartlettroscoe:
> > In my view, running these checks along with other tests with CTest is a reasonable compromise if developers are not willing to use git commit hooks (which need to be installed and maintained automatically if they are going to be used consistently).
> 
> Isn't the compromise that they have to wait on the PR check to run? Meaning, if they don't want to implement user-side hooks, they do not get as immediate of feedback? Meaning they are welcome to not use the hooks (don't make them mandatory), but that means feedback is delayed.

As a general rule of thumb, it should be easy for developers to reproduce locally any automated check enforced on their PR.   (They can do that Trilinos PR builds using GenConfig, for example, using some simple driver scripts.)  And that is in the best interest of Trilinos too.  We don't want Trilinos developers pushing small tweaks to GitHub just to run this check again (and tying up huge computational resources running all of the PR builds again).

If one expects developers to use Git hooks, then the project should provide reasonable infrastructure and guidance on how to install those checks.  (For example, Trilinos should provide suggested git commit hooks that developers can just install and update.)  Personally, I think running more expensive checks as part of the CTest test suite makes more sense.

Technically speaking, the Trilinos test suite could also be run as a commit hook but that would really hut productivity.

The one type of check that makes the most sense to check at commit time are checks for the commit  log message itself like:

```
$ cat .git/hooks/commit-msg
which codespell >> /dev/null
which_codespell_rtn=$?

if [[ "${which_codespell_rtn}" == "0" ]] ; then
  codespell $1
fi
```

I have said my piece.  Let's see what the feedback from Developers is like.  (It may be a while before someone inserts `MPI_COMM_WORLD` in their library code so this might not be a big deal.)

bartlettroscoe:
@stmcgovern and @sebrowne, one should note that the tool `check-mpi-comm-world-usage.py` has about 95% of the code needed to check for the occurrence of any undesirable identifier in the code base.  In the future, it would be easy to refactor this tool to read from a list of undesirable identifiers and flag for them (with the same ignore text `CHECK: ALLOW <some-word>`).

ndellingwood:
> @stmcgovern and @sebrowne, one should note that the tool `check-mpi-comm-world-usage.py` has about 95% of the code needed to check for the occurrence of any undesirable identifier in the code base.

Oh, that would be helpful to chase down if-guarded blocks of deprecated code slated for removal

ccober6:
Let's go with what we currently have to test out this new Github action.  We can return to the additional feature that @bartlettroscoe suggested at a later date.  This meets the current requirements and priority, and reassess the new feature.  Thanks, @ccober6 and @jwillenbring

ccober6 (approved):
lgtm

Issue #11743, opened 2023-03-28
Tpetra: Add memory diagnostics to performance tests


csiefer2:
As per stakeholder request.

PR #12113, opened 2023-08-07
Tpetra: Adding initial memory HWM performance test

csiefer2:
Issue: #11743
Tests: This is one
Stakeholder feedback: n/a

### DIFF
--- a/packages/tpetra/core/test/ImportExport/GDSW_Proxy_def.hpp
+++ b/packages/tpetra/core/test/ImportExport/GDSW_Proxy_def.hpp
@@ -527,10 +527,7 @@ void TpetraFunctions<Scalar,LocalOrdinal,GlobalOrdinal,Node>::
 importSquareMatrixFromImporter4(RCP<const CrsMatrix> inputMatrix, 
                                RCP<const Import> importer,
                                 RCP<CrsMatrix> & outputMatrix) {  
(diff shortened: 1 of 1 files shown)


packages/tpetra/core/test/ImportExport/ImportExport_GDSWStyle.cpp
```diff
@@ -0,0 +1,694 @@
...
+  RCP<const Comm<int> > comm = A.getRowMap ()->getComm ();
+  int globalEqual = 0;
+  reduceAll<int, int> (*comm, REDUCE_MIN, 1, &localEqual, &globalEqual);
+  return globalEqual == 1;
+}
+
+
+
+// return current memory usage in kilobytes
+size_t get_memory_usage_now()
```
cwpearson, line 356: Do we care about GPU memory? If so, maybe return a std::pair for system + gpu

cwpearson (commented):
I assume the particular problem under test was requested by a user, so I can't speak to that; my only concern is whether this should be augmented to track GPU memory as well.

cwpearson:
Any maybe a squash if you're feeling like it

cwpearson (approved):

ccober6:
@csiefer2 , yeah it is amazing that the optimized runs on my laptop in <10 seconds, and the debug takes >10 minutes when resources are loaded!

Issue #12106, opened 2023-08-04
Tpetra/MueLu: Deep Copy Diagnostics


csiefer2:
We need to justify when we do them.

PR #12114, opened 2023-08-07
Tpetra: Adding verbose mode for Kokkos::deep_copy timer injection

csiefer2:
Issue: #12106
Test: n/a
Stakeholder feedback: Long MueSlack discussions


jhux2 (approved):
yes!

packages/tpetra/core/src/Tpetra_Details_DeepCopyTeuchosTimerInjection.cpp
cwpearson, line 1: Not to be picky but it might be best if there is a `" "` between `"]"` and the added `extra_label`

cwpearson (approved):

PR #12126, opened 2023-08-14
Add option to test against develop for kokkos and kokkos-kernels packages

srbdev:
@trilinos/framework 

## Motivation
As a Kokkos/KokkosKernels developer, I want Trilinos developers to know about issues with integrating a newer version of these core packages as early as possible, to decrease the effort of a single person overseeing the integration.

As a Trilinos developer, I want the most current Kokkos/KokkosKernels as is feasible to enable new capability.

As a Trilinos customer, I want the most current version of all packages as soon as possible, particularly for use on advanced new platforms.

See TRILFRAME-571 for more details.

### DIFF
--- a/packages/framework/pr_tools/PullRequestLinuxDriver.sh
+++ b/packages/framework/pr_tools/PullRequestLinuxDriver.sh
@@ -93,67 +93,67 @@ sig_merge_old=$(get_md5sum ${REPO_ROOT:?}/packages/framework/pr_tools/PullReques
 if [[ ${kokkos_develop} == "1" ]]; then
     message_std "PRDriver> --kokkos-develop is set - setting kokkos and kokkos-kernels packages to current develop"
     ./$SCRIPTPATH/SetKokkosDevelop.sh
-fi
+else
+    print_banner "Merge Source into Target"
+    message_std "PRDriver> " "TRILINOS_SOURCE_SHA: ${TRILINOS_SOURCE_SHA:?}"
 
-print_banner "Merge Source into Target"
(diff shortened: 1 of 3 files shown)


packages/framework/pr_tools/PullRequestLinuxDriver.sh
```diff
@@ -8,6 +8,7 @@ source ${SCRIPTPATH:?}/common.bash
 on_weaver=$(echo "$@" | grep '\-\-on_weaver' &> /dev/null && echo "1")
 on_ats2=$(echo "$@" | grep '\-\-on_ats2' &> /dev/null && echo "1")
 bootstrap=$(echo "$@" | grep '\-\-\no\-bootstrap' &> /dev/null && echo "0" || echo "1")
+kokkos_develop=$(echo "$@" | grep '\-\-kokkos\-develop' &> /dev/null && echo "0" || echo "1")
```
sebrowne, line 11: I think this is backwards.  Note that the bootstrap variable is inverted from its option (--no-bootstrap)

sebrowne (changes requested):

packages/framework/pr_tools/PullRequestLinuxDriver.sh
srbdev: Good catch! 😅 

I pushed the change: re-organized the lines to group them by the way the flag is parsed and renamed to `on_kokkos_develop` to copy the convention of the other variables.

sebrowne (approved):

PR #12141, opened 2023-08-17
Intrepid2: Removing Experimental namespace used for projections (part1)

mperego:
@trilinos/intrepid2 
@trilinos/panzer

## Motivation
The projection tools are mature enough and we want to remove the Experimental namespace. 
Added cmake option` Intrepid2_KEEP_EXPERIMENTAL_NAMESPACE`. When set,
we use the legacy namespace, otherwise the Experimental namespace is removed. The option is ON by default.
This will give time to applications (EMPIRE) to update their code before we remove the namespace completely.
Some of the legacy projection functions will be removed together with the Experimental namespace.

This PR also include some modifications to the cell tools to:
1. allow the map ref cells in a physical space that is higher dimensional  (e.g. quadrilateral in a 3d physical space)
2. call mapToReferenceSubcell on a batch of cells, where the subcell ordinal is an array changing from cell to cell.
This is useful when working on a batch of sides.

### DIFF
--- a/packages/intrepid2/CMakeLists.txt
+++ b/packages/intrepid2/CMakeLists.txt
@@ -34,7 +34,7 @@ ENDIF()
 TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_KEEP_EXPERIMENTAL_NAMESPACE
   HAVE_INTREPID2_EXPERIMENTAL_NAMESPACE
   "Whether Intrepid2 keeps experimentatl namespace for projections."
-  OFF
+  ON
 )
 
 #
--- a/packages/intrepid2/CMakeLists.txt
+++ b/packages/intrepid2/CMakeLists.txt
@@ -30,6 +30,13 @@ ELSE()
     OFF)
 ENDIF()   
 
+# temporary flag
+TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_KEEP_EXPERIMENTAL_NAMESPACE
+  HAVE_INTREPID2_EXPERIMENTAL_NAMESPACE
+  "Whether Intrepid2 keeps experimentatl namespace for projections."
+  OFF
+)
+
 #
 # C) Add the libraries, tests, and examples
 #
--- a/packages/intrepid2/cmake/Intrepid2_config.h.in
+++ b/packages/intrepid2/cmake/Intrepid2_config.h.in
@@ -10,3 +10,6 @@
 /* Define if want to build with KokkosKernels enabled */
 #cmakedefine HAVE_INTREPID2_KOKKOSKERNELS
 
+/* Temporary flag */
+#cmakedefine HAVE_INTREPID2_EXPERIMENTAL_NAMESPACE
+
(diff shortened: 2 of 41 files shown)


packages/intrepid2/CMakeLists.txt
```diff
@@ -30,6 +30,13 @@ ELSE()
     OFF)
 ENDIF()   
 
+# temporary flag
+TRIBITS_ADD_OPTION_AND_DEFINE(${PACKAGE_NAME}_KEEP_EXPERIMENTAL_NAMESPACE
```
cgcgcg, line 34: Can this be handled with the existing deprecation mechanisms?

mperego: Well, we could, but I don't want to wait until the point release to remove the namespace. I just want to give Roger some time to make the changes in Empire.

CamelliaDPG:
@mperego, I'm working on the review, and have a couple high-level questions.

First, are there some functions here that you intend to remove when the Experimental namespace is removed?  E.g., it looks like in `Intrepid2_ProjectionToolsDefHGRAD.hpp`, the function `ProjectionTools::getHGradEvaluationPoints()` will no longer exist.  I think that makes sense, given the changes you made recently with quadrature points, but I just wanted to confirm that this is the intent -- it's not just that we're removing the namespace, but we're also taking the opportunity to remove some now-unnecessary functionality.

Second, if `Intrepid2_KEEP_EXPERIMENTAL_NAMESPACE` is off by default, do you also include here some testing of the case when it's on?  Could you point me to where that is?  (This may be quite obvious once I get to it, but I don't think I've seen it yet.)

CamelliaDPG (changes requested):
@mperego, the changes look good to me.  My only concern is the testing of the ON option for HAVE_INTREPID2_EXPERIMENTAL_NAMESPACE; it seems to me that if we're keeping the namespace around for the sake of EMPIRE, e.g., we'd better have testing of their use case.  I may have missed it, but it looks to me like the ON case is not tested.

I'm not sure how easy it would be to add testing of the ON case in addition to the OFF case, but I'd be fine with testing only the ON case; I'd suggest making ON the default until EMPIRE is ready to switch.

If I've missed the test, just let me know and I'll approve.

CamelliaDPG (approved):
Thanks, @mperego!

mperego:
> @mperego, the changes look good to me. My only concern is the testing of the ON option for HAVE_INTREPID2_EXPERIMENTAL_NAMESPACE; it seems to me that if we're keeping the namespace around for the sake of EMPIRE, e.g., we'd better have testing of their use case. I may have missed it, but it looks to me like the ON case is not tested.
> 
> I'm not sure how easy it would be to add testing of the ON case in addition to the OFF case, but I'd be fine with testing only the ON case; I'd suggest making ON the default until EMPIRE is ready to switch.
> 
> If I've missed the test, just let me know and I'll approve.

Good point. I've changed the default for the `HAVE_INTREPID2_EXPERIMENTAL_NAMESPACE` to ON. Also I've edited the PR description to explain that some legacy function will be removed together with the namespace.

PR #12142, opened 2023-08-17
DO NOT MERGE MueLu+Panzer: Darcy problems

cgcgcg:
@trilinos/muelu @trilinos/panzer 

## Motivation
- Add a multigrid solver for grad-div problems to MueLu by extending RefMaxwell.
- Assemble a Darcy problem in MiniEM first or higher order elements, reduce to grad-div, solve directly using new solver (1st order) or using p-multigrid.

DO NOT YET REVIEW OR MERGE: I want to see how many tests still fail.


cgcgcg:
@trilinos/framework Shouldn't the MPI_COMM_WORLD check run only for files that are changed by the PR?

sebrowne:
> @trilinos/framework Shouldn't the MPI_COMM_WORLD check run only for files that are changed by the PR?

That's the theory.  @JacobDomagala wrote the script that does the diff generation, he may have a better idea of what is happening.  It does think that you modified a ROL file that doesn't show up in the diff at the very least (there are others).

JacobDomagala:
> > @trilinos/framework Shouldn't the MPI_COMM_WORLD check run only for files that are changed by the PR?
> 
> That's the theory. @JacobDomagala wrote the script that does the diff generation, he may have a better idea of what is happening. It does think that you modified a ROL file that doesn't show up in the diff at the very least (there are others).

The python script compares origin/develop to user's most recent commit using:
`git diff -U0 --ignore-all-space <start_commit> <end_commit>`
I didn't take into account the case where user's branch is out of sync with target branch (Trilinos' develop). This is why you see files that differ between our branch and the develop. 

I've created https://github.com/trilinos/Trilinos/pull/12199 which should fix the sync issue

packages/panzer/mini-em/src/closures/MiniEM_PiecewiseConstant_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+  const double zl = zl_;
+  const double zr = zr_;
+
+  if (ir_dim == 3) {
+    Kokkos::MDRangePolicy<PHX::exec_space,Kokkos::Rank<2>> policy({0,0},{workset.num_cells,values.extent_int(1)});
+    Kokkos::parallel_for("panzer:PiecewiseConstant 3D",policy,KOKKOS_LAMBDA (const int cell,const int point) {
+
+        const ScalarT& x = tmp_coords(cell,point,0);
+        const ScalarT& y = tmp_coords(cell,point,1);
+        const ScalarT& z = tmp_coords(cell,point,2);
```
rppawlo, line 78: I would probably make these "auto" instead of ScalarT.

packages/panzer/mini-em/src/closures/MiniEM_PiecewiseConstant_impl.hpp
```diff
@@ -0,0 +1,106 @@
...
+          tmp_values(cell,point) = value0;
+        else
+          tmp_values(cell,point) = value1;
+      });
+  } else {
+    Kokkos::MDRangePolicy<PHX::exec_space,Kokkos::Rank<2>> policy({0,0},{workset.num_cells,values.extent_int(1)});
+    Kokkos::parallel_for("panzer:PiecewiseConstant 2D",policy,KOKKOS_LAMBDA (const int cell,const int point) {
+
+        const ScalarT& x = tmp_coords(cell,point,0);
+        const ScalarT& y = tmp_coords(cell,point,1);
```
rppawlo, line 92: same here

packages/trilinoscouplings/examples/scaling/example_Maxwell.cpp
```diff
@@ -403,7 +406,16 @@ int evalCurlCurlu(double & curlcurlu0,
 /**********************************************************************************/
 /******************************** MAIN ********************************************/
 /**********************************************************************************/
+int body(int argc, char *argv[]);
+
 int main(int argc, char *argv[]) {
+  Kokkos::initialize(argc,argv);
```
rppawlo, line 412: If you use MPI, then you should initialize kokkos immediately after initializing MPI. Either move the GlobalMpiSession here or move the kokkos::init into body and call after constructing the mpi session object. This is needed so that kokkos gets the process binding masks from MPI:

https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Initialization.html#initialization

packages/panzer/mini-em/src/closures/MiniEM_PiecewiseConstant_impl.hpp
cgcgcg: done

packages/panzer/mini-em/src/closures/MiniEM_PiecewiseConstant_impl.hpp
cgcgcg: done

packages/trilinoscouplings/examples/scaling/example_Maxwell.cpp
cgcgcg: Fixed for this test and the source of my pattern match change ;-)

rppawlo (dismissed):

rppawlo:
@cgcgcg - with these changes, will we have to update the input files for muelu refmaxwell use in empire?

cgcgcg:
The interface changes should be backward compatible. If EMPIRE's promotion gets stuck please ping me and I will take care of it. Down the line I'll switch EMPIRE over to the new parameter names.

bartlettroscoe:
FYI: The last PR test iteration [above](https://github.com/trilinos/Trilinos/pull/12142#issuecomment-1779739567) failed because of a timeout (see [Tempus_BDF2_MPI_1](https://trilinos-cdash.sandia.gov/test/38991651)) and also failed the [TrilinosInstallTests_reduced_tarball](https://trilinos-cdash.sandia.gov/test/38990846) test showing:

```
Final set of non-enabled top-level packages:  ML 1

...

TEST_2: Pass criteria = Match all REGEX {Final set of non-enabled top-level packages:  0} [FAILED]
```

For some reason, in this  build [rhel7_sems-cuda-11.4.2-sems-gnu-10.1.0-sems-openmpi-4.0.5_release_static_Volta70_no-asan_complex_no-fpic_mpi_pt_no-rdc_no-uvm_deprecated-on_no-package-enables](https://trilinos-cdash.sandia.gov/build/1218451), the package directory for ML is not being excluded from the tarball.  That is strange because the test `TrilinosInstallTests_reduced_tarball` is running and passing in the other builds and does not show this problem with ML.  The question is, why is ML not getting enabled in a CUDA build from the reduced source tree but but is from the non-CUDA builds?  Well, it seems that ML is being explicitly disabled in the CUDA build as shown [here](https://trilinos-cdash.sandia.gov/test/38990846) showing:

```
Explicitly disabled top-level packages on input (by user or by default):  ML 1
```

Is there some logic that is disabling ML for CUDA build by default?

sebrowne:
@bartlettroscoe we're disabling it in the `config-specs.ini` settings for CUDA builds.  I think I looked at a very similar issue a few weeks ago, and the issue is likely a CMake variable scoping/precedence issue.  I think if I add FORCE to the setting in `config-specs.ini` it will fix the issue (I tested it, but again has been a few days).  The "build what changed" script decided (correctly) to enable ML, even though we disable it in the config.  Forcing it in the config fixes it and leaves it disabled in the main (non-install test) build.

bartlettroscoe:
> @bartlettroscoe we're disabling it in the `config-specs.ini` settings for CUDA builds. I think I looked at a very similar issue a few weeks ago, and the issue is likely a CMake variable scoping/precedence issue. 

@sebrowne, okay, I see:

```
set(Trilinos_ENABLE_ML OFF CACHE BOOL "from .ini configuration")
```

in the file [generatedPRFragment.cmake](https://trilinos-cdash.sandia.gov/upload/533ceac9df08f71583a21c4691b6273f3a59993d/generatedPRFragment.cmake).  So yes, the problem is that `Trilinos_ENABLE_ML` is getting set to `ON` by the file [packageEnables.cmake](https://trilinos-cdash.sandia.gov/upload/5db2305164ab672fd29912dc6c59ddab0002b6bd/packageEnables.cmake) being passed in by `-C packageEnables.cmake`. ML should have never of been enabled in the outer CMake configure process.

> I think if I add FORCE to the setting in `config-specs.ini` it will fix the issue (I tested it, but again has been a few days). The "build what changed" script decided (correctly) to enable ML, even though we disable it in the config. Forcing it in the config fixes it and leaves it disabled in the main (non-install test) build.

Yes, I think if you change it to `FORCE` in the `*.ini` file, then that should fix this.  But it can be pretty hard to debug where these variables are being overridden like this.  Hopefully CMake would put the doc string "from .ini configuration" in the generated `CMakeCache.txt` file to give a clue about what is happening her.  (And it would help if the Trilinos CTest -S drive would post the generated CMakeCache.txt file to help debug issues.)

The other way to get around this is to prcoess `packageEnables.cmake` **after** `generatedPRFragment.cmake` so the cache vars set in `generatedPRFragment.cmake` would not get overridden by those in `packageEnables.cmake`.  But the user's cache vars passed in with `-D <cacheVar>=<value>` would override anything set in any of these files.  That gives the end user the ability to change settings (which FORCE destroys). 

But now we have a case where the reduced tarball testing happened to catch a defect in the setup of the CI testing process, so this is a good thing 👍

sebrowne:
@cgcgcg try changing lines 912-942 of `packages/framework/ini-files/config-specs.ini` (the blocks that say ALL-NO-EPETRA and NO-EPETRA) to have 'FORCE' before the colon.  For example:
`opt-set-cmake-var Trilinos_ENABLE_ML BOOL : OFF`
becomes
`opt-set-cmake-var Trilinos_ENABLE_ML BOOL FORCE : OFF`

sebrowne (dismissed):

cgcgcg:
Thanks @rppawlo !

cgcgcg:
Sure thing, that's why I added the label.

PR #12149, opened 2023-08-18
Framework: Create GitHub action check for detecting addition of `MPI_COMM_WORLD`

JacobDomagala:
Per discussion in https://github.com/trilinos/Trilinos/pull/12111

### DIFF
--- a/.github/workflows/detect-mpi-comm-world.yml
+++ b/.github/workflows/detect-mpi-comm-world.yml
@@ -1,0 +1,20 @@
+name: Check MPI_COMM_WORLD
+
+on:
+  pull_request:
+
+jobs:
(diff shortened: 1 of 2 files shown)


ccober6:
@sebrowne

sebrowne (approved):

sebrowne (commented):
Will this script detect itself?

stmcgovern:
> Will this script detect itself?

`  if file_name.endswith((".c", ".cpp", ".h", ".hpp")) and ...`
  
  so it shouldn't be looking in a yml. 

Thanks for approving @sebrowne . Thanks for making this @JacobDomagala . It looks good to me.

PR #12163, opened 2023-08-24
AztecOO: Don't use hardcoded `MPI_COMM_WORLD` in `AztecOO` code

JacobDomagala:
@trilinos/aztecoo

~~Introduce  `az_global_mpi_comm` global variable that defaults to `MPI_COMM_WORLD` and use it instead of hardcoded `MPI_COMM_WORLD`.~~

This PR is part of the initiative to phase out the use of `MPI_COMM_WORLD` in Trilinos.

EDIT: 
See https://github.com/trilinos/Trilinos/pull/12163#issuecomment-1737863375

### DIFF
--- a/packages/aztecoo/src/az_old_matvec_mult.c
+++ b/packages/aztecoo/src/az_old_matvec_mult.c
@@ -1,13 +1,13 @@
 /*
 //@HEADER
 // ***********************************************************************
-// 
-//        AztecOO: An Object-Oriented Aztec Linear Solver Package 
+//
+//        AztecOO: An Object-Oriented Aztec Linear Solver Package
 //                 Copyright (2002) Sandia Corporation
-// 
+//
 // Under terms of Contract DE-AC04-94AL85000, there is a non-exclusive
 // license for use of this work by or on behalf of the U.S. Government.
-// 
+//
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are
 // met:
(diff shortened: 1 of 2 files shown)


ccober6:
Hi @JacobDomagala.  The reason to remove hardcoded uses of MPI_COMM_WORLD is to allow the user to supply a communicator which is different than MPI_COMM_WORLD.  Of course the default can be MPI_COMM_WORLD, but I think there needs to be a way to supply the user's communicator.

JacobDomagala:
@csiefer2 @ccober6 I added setter and getter so user can set different communicator:
```c++
MPI_Comm newComm; 
MPI_Comm_split(/*...*/, &newComm)
AZ_initialize_global_comm(newComm)

/*
Use AztecOO code 
*/
```

jwillenbring:
> @csiefer2 @ccober6 I added setter and getter so user can set different communicator:
> 
> ```c++
> MPI_Comm newComm; 
> MPI_Comm_split(/*...*/, &newComm)
> AZ_initialize_global_comm(newComm)
> 
> /*
> Use AztecOO code 
> */
> ```

@csiefer2 Can you please comment on if these changes are sufficient? Thanks.

stmcgovern:
> This is still just hiding MPI_COMM_WORLD. I'm not in charge of this exercise, so I don't really get a vote, but I was under the impression that we're actually supposed to stop using MPI_COMM_WORLD in anything but tests.

The current status of the code using  `MPI_COMM_WORLD` directly in the MPI calls requires modification.
The proposed strategy is to allow the user to pass a communicator in ,`comm`,  otherwise default to `MPI_COMM_WORLD` as the global comm. I believe that this PR satisfies that approach. 

FYI @ccober6 @jwillenbring

cgcgcg:
How is this tested? Personally, I cannot tell by looking at the code if this actually runs correctly on a sub-communicator.

JacobDomagala:
Ok I just realized that AztecOO actually provides the Comm-friendly API for MPI wrappers:
https://github.com/trilinos/Trilinos/blob/develop/packages/aztecoo/src/md_wrap_mpi_c.c#L348-L571

As for the `az_old_matvec_mult` changes, from the looks of it it's deprecated function. 

I'll just add comments that allow the use there.

cgcgcg (approved):

PR #12165, opened 2023-08-24
ROL: SplitComm allow for user to pass custom communicator

JacobDomagala:
@trilinos/rol 

This PR is part of the initiative to phase out the use of MPI_COMM_WORLD in Trilinos.

### DIFF
--- a/packages/rol/src/oed/utilities/OED_SplitComm.hpp
+++ b/packages/rol/src/oed/utilities/OED_SplitComm.hpp
@@ -58,15 +58,15 @@ private:
   ROL::Ptr<Comm> comm_sample_;
 
 public:
-  SplitComm(const int m) {
+  SplitComm(const int m, Mpi_Comm global_comm = MPI_COMM_WORLD) {
     Ordinal rank, Ngroups, size;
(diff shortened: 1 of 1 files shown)


dridzal:
@JacobDomagala : This looks fine, and I can formally review it, however can you explain the initiative?  I see that you moved MPI_COMM_WORLD to be the default argument value for a potentially different global comm.  Is that sufficient to meet the requirements of the initiative?  CCing @dpkouri.

ccober6:
Hi @dridzal, this is part of our debt reduction effort, specifically trying to meet the [xSDK policy (M3)](https://github.com/xsdk-project/xsdk-community-policies/blob/master/package_policies/M3.md).  We are working with the assumption that MPI_COMM_WORLD is a perfectly good default, but should not be directly hard coded in MPI functions (with the possible exception of MPI_Abort).  We are trying to limit the scope of this modification, and still meet the policy.  Obviously there would need to be the correct usage with the new communicator, but it would be available.  Hope this helps.

ccober6:
@dridzal, @dpkouri, please review if you are ok with this.  Thanks, Curt

dpkouri (approved):

PR #12167, opened 2023-08-24
STK: Don't use `MPI_COMM_WORLD`

JacobDomagala:
@trilinos/stk

This PR is part of the initiative to phase out the use of `MPI_COMM_WORLD` in Trilinos.

### DIFF
--- a/packages/stk/stk_balance/stk_balance/setup/M2NParser.cpp
+++ b/packages/stk/stk_balance/stk_balance/setup/M2NParser.cpp
@@ -35,6 +35,7 @@
 #include "stk_util/command_line/CommandLineParserUtils.hpp"
 #include "stk_util/util/string_utils.hpp"
 #include "stk_balance/balanceUtils.hpp"
+#include "stk_util/parallel/GlobalComm.hpp"
 
 namespace stk {
 namespace balance {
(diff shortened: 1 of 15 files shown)


alanw0 (commented):
Thanks for the proposed changes.

stk is developed in a separate repository and periodically snapshotted into trilinos, so we will need to make changes there to avoid having these get lost at the next snapshot.

We agree with the general premise that MPI_COMM_WORLD should not be used except in main and in functions that are equivalent to main such as gtest unit-tests.

We will prefer to implement some of these changes slightly differently, such as adding a mechanism to pass a communicator from a scope that already has one, rather than creating a global-comm facility.

What time-frame are you hoping to have MPI_COMM_WORLD usage eliminated in trilinos?

ccober6:
@alanw0, this is some contract work we are having done.  They are hoping to complete this task by next week.  :)  If you can meet that timeline, it would be great, but not necessary.  Otherwise, would you be able to get this in before the end of the FY (~Sep. 22) and the next release (Trilinos 15.0)?

Another part of this effort is to have a non-blocking PR test to warn developers about adding back MPI_COMM_WORLD usage, as a few years ago we had a similar effort and since have had some regression.  :(

alanw0:
@ccober6 Thanks Curt, these changes don't look too difficult for stk, we can probably have them in stk/sierra by next week. Getting the snapshot back to trilinos is sometimes unpredictable. But beating the next release should not be a problem.

ccober6:
Great!  Thanks @alanw0!

jwillenbring:
@alanw0 Please confirm when you have made (the equivalent of) these changes to STK. We can close this PR at that time.

alanw0:
@jwillenbring we have made most of them, still a couple of stragglers we should have in the next week. I agree it's probably safe to go ahead and close this. We'll have a new stk snapshot coming into trilinos soon also.

alanw0:
> @alanw0 Have these changes been made in the Trilinos repo?

No not yet. I have the changes in a snapshot of stk that I'm trying to put into trilinos, but it hasn't made it through PR testing yet. We apparently have something wrong with one of our newer doc-test executables. I'll try to get that through soon.

alanw0:
@jwillenbring these stk changes did go into trilinos/develop mid-late last week.

PR #12172, opened 2023-08-28
Adelus: Add inline to get_seconds and showtime functions

vqd8a:
@trilinos/adelus

## Motivation

This PR adds inline to the two functions in Adelus: ```get_seconds``` and ```showtime``` to fix the errors of multiple definitions in Gemma

## Testing

The Gemma team confirmed this fixes the errors.


jdkotul (approved):
This eliminates the multiple definitions of a function and fixes a linking error.

PR #12179, opened 2023-08-29
Ifpack2:  Adds missing test dependency on Xpetra  

hkthorn:
@trilinos/ifpack2 

## Motivation

The Ifpack2 tests should have a required dependency on Xpetra.  There is
    an optional dependency on Xpetra for the library, but if someone uses a
    configuration like this:
    
    cmake \
    -DCMAKE_C_COMPILER=mpicc \
    -DCMAKE_CXX_COMPILER=mpicxx \
    -DCMAKE_Fortran_COMPILER=mpif77 \
    -DCMAKE_CXX_FLAGS="-O3 -fPIC" \
    -DCMAKE_C_FLAGS="-O3 -fPIC" \
    -D Trilinos_ENABLE_Tpetra:BOOL=ON \
    -D Trilinos_ENABLE_Ifpack2:BOOL=ON \
    -D Teuchos_ENABLE_COMPLEX:BOOL=OFF \
    -D Trilinos_ENABLE_COMPLEX:BOOL=OFF \
    -D Trilinos_ENABLE_OpenMP:BOOL=OFF \
    -D CMAKE_VERBOSE_MAKEFILE:BOOL=ON \
    -DTPL_ENABLE_MPI:BOOL=ON \
    -DTPL_ENABLE_BLAS=ON \
    -DTPL_ENABLE_LAPACK=ON \
    ../../
    
  Then Xpetra will not be enabled since optional dependencies are off,
  yet some tests will fail to build due to the absence of Xpetra.

## Stakeholder Feedback

After the change to the cmake dependencies file all the Ifpack2 tests will build.

## Testing

OSX GCC12 serial and parallel

### DIFF
--- a/packages/muelu/src/Graph/UncoupledAggregation/MueLu_AggregationPhase2aAlgorithm_def.hpp
+++ b/packages/muelu/src/Graph/UncoupledAggregation/MueLu_AggregationPhase2aAlgorithm_def.hpp
@@ -81,7 +81,40 @@ namespace MueLu {
     LO numLocalAggregated = numLocalNodes - numNonAggregatedNodes;
 
     const double aggFactor = params.get<double>("aggregation: phase2a agg factor");
-    double       factor    = as<double>(numLocalAggregated)/(numLocalNodes+1);
+    double       factor;
+
+
+    if (matchMLbehavior) {
+      // Note: ML uses global counts to set the factor
+      // Passing  # of nonaggregated nodes and # of nodes via aggStat
+      GO in_data[2] ={(GO)numNonAggregatedNodes,(GO) aggStat.size()};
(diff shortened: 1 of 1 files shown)


jhux2 (changes requested):
@hkthorn I grep'd through the tests, and there is logic in most to check for the presence of Xpetra.  The only error I could see is in `Ifpack2_UnitTestSingleProcessILUT.cpp`, where it looks likes Xpetra is used unconditionally.

I'd rather see the logic fixed than make Xpetra a hard dependency.

jhux2:
> It sounds like you would rather I change the testing logic to address the presence of Xpetra than add cmake dependencies?

Yes, please.

PR #12186, opened 2023-08-31
Tpetra: Making the MV random pool static

csiefer2:
Issue: #12106 
Tests: Should already be exercised.
Stakeholder feedback: n/a

Every time MV::randomize() gets called, it initializes a Kokkos::Random_XorShift64_Pool object, which involves some pretty beefy H2D transfers.  We'll try to make that static so it only happens on the first call and subsequent calls just advance the state.

Right now, each Chebyshev setup (at each level) in MueLu triggers the pool initialization.


jhux2 (approved):
lgtm

jhux2 (changes requested):
nah

PR #12187, opened 2023-08-31
Tpetra: Removing the incorrect ETI_DEVICES

csiefer2:
We don't use them and they're not consistent with the ETI_NODES

Issue: #12106
Tests: n/a
Stakeholder feedback: n/a

### DIFF
--- a/packages/tpetra/CMakeLists.txt
+++ b/packages/tpetra/CMakeLists.txt
@@ -111,18 +111,7 @@ INCLUDE(CMakeCuda.txt)
 # Note on (2) above: some of the types can only be enabled if certain
 # requirements are fulfilled.
 #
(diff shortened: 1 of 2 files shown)


jhux2 (approved):
lgtm

jhux2 (changes requested):
What about https://github.com/trilinos/Trilinos/blob/67e9032a6d0524d57e6e6bad55b2550a062f0c0a/packages/tpetra/core/guide/src/Obtaining/Building.rst?plain=1#L107?

sebrowne:
@csiefer2 checked it out, looks like one of those builds lost a connection to GitHub when it attempted to check out Trilinos.  That same one already passed this time.  Probably a random connection hiccup.

jhux2 (approved):
Consider me happy.

PR #12191, opened 2023-09-01
EpetraExt: Remove use of hardcoded MPI_COMM_WORLD

thearusable:
@trilinos/epetraext

### DIFF
--- a/packages/epetraext/src/inout/EpetraExt_HDF5.cpp
+++ b/packages/epetraext/src/inout/EpetraExt_HDF5.cpp
@@ -379,7 +379,7 @@ void EpetraExt::HDF5::Open(const std::string FileName, int AccessType)
   const Epetra_MpiComm* MpiComm ( dynamic_cast<const Epetra_MpiComm*> (&Comm_) );
 
   if (MpiComm == 0)
-    H5Pset_fapl_mpio(plist_id_, MPI_COMM_WORLD, MPI_INFO_NULL);
+    H5Pset_fapl_mpio(plist_id_, MPI_COMM_WORLD, MPI_INFO_NULL); // CHECK: ALLOW MPI_COMM_WORLD
   else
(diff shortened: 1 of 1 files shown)


cgcgcg:
@thearusable Can you give some more details on why this change is required? It seems that particular piece of code already allows to use a user specified communicator and only uses `MPI_COMM_WORLD` when nothing else is specified.

thearusable:
> @thearusable Can you give some more details on why this change is required? It seems that particular piece of code already allows to use a user specified communicator and only uses `MPI_COMM_WORLD` when nothing else is specified.

I thought that all occurrences of `MPI_COMM_WORLD` in the source files needs to be removed. 

If this one is indeed allowed then I can remove my changes and mark it with `// CHECK: ALLOW MPI_COMM_WORLD` for future CI runs.

stmcgovern:
@cgcgcg What's your preferred approach here? (related to https://github.com/trilinos/Trilinos/pull/12194 ) 

FYI @thearusable

cgcgcg:
I think this a pattern that is mirrored in several packages: only if no communicator is passed by the user then use `MPI_COMM_WORLD`. I think that this is a safe usage of `MPI_COMM_WORLD`.

stmcgovern:
@cgcgcg and @jwillenbring would you be able to recommend/assign a reviewer for this package please?

jwillenbring:
> I think this a pattern that is mirrored in several packages: only if no communicator is passed by the user then use `MPI_COMM_WORLD`. I think that this is a safe usage of `MPI_COMM_WORLD`.

This sounds similar then to #12194

thearusable:
@cgcgcg @jwillenbring - I removed my previous changes. Now there is only a comment which allows `MPI_COMM_WORLD` usage to pass the check.

@stmcgovern - FYI

cgcgcg (approved):

cgcgcg (approved):

PR #12192, opened 2023-09-01
Belos: Remove use of hardcoded MPI_COMM_WORLD

thearusable:
@trilinos/belos


hkthorn (approved):
LGTM

PR #12193, opened 2023-09-01
Ifpack: Remove use of hardcoded MPI_COMM_WORLD

thearusable:
@trilinos/ifpack

~~Introduce `Global_MPI_Comm` global variable that defaults to `MPI_COMM_WORLD` and use it instead of the hardcoded `MPI_COMM_WORLD`. This PR includes also method which allows for the change of the global communicator.~~

See comments:
- https://github.com/trilinos/Trilinos/pull/12193#discussion_r1339098111
- https://github.com/trilinos/Trilinos/pull/12193#discussion_r1339162430

### DIFF
--- a/packages/ifpack/src/euclid/guards_dh.h
+++ b/packages/ifpack/src/euclid/guards_dh.h
@@ -52,7 +52,7 @@
 
 #define  INITIALIZE_DH(argc, argv, help) \
             MPI_Init(&argc,&argv);  \
-            comm_dh = MPI_COMM_WORLD;    \
+            comm_dh = MPI_COMM_WORLD; /* CHECK: ALLOW MPI_COMM_WORLD */ \
             MPI_Errhandler_set(comm_dh, MPI_ERRORS_RETURN); \
             EuclidInitialize(argc, argv, help); \
             dh_StartFunc(__FUNC__, __FILE__, __LINE__, 1); \
--- a/packages/ifpack/src/euclid/mat_dh_private.c
+++ b/packages/ifpack/src/euclid/mat_dh_private.c
@@ -1166,7 +1166,7 @@ partition_and_distribute_metis_private (Mat_dh A, Mat_dh * Bout)
   /* broadcast number of rows to all processors */
   if (myid_dh == 0)
     m = A->m;
-  MPI_Bcast (&m, 1, MPI_INT, 0, MPI_COMM_WORLD);
+  MPI_Bcast (&m, 1, MPI_INT, 0, MPI_COMM_WORLD); // CHECK: ALLOW MPI_COMM_WORLD
 
   /* broadcast number of nonzeros in each row to all processors */
   rowLengths = (int *) MALLOC_DH (m * sizeof (int));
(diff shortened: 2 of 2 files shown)


stmcgovern:
@jwillenbring could you please suggest a reviewer here?

jwillenbring:
@srajama1 Could you review this PR or suggest someone who could?

cgcgcg:
How does this interact with ML that uses Ifpack? Shouldn't there be a call from ML to set the correct communicator?

jwillenbring:
> How does this interact with ML that uses Ifpack? Shouldn't there be a call from ML to set the correct communicator?

@cgcgcg I am not fully understanding your question. How is this working properly now? What changes are you suggesting more specifically?

cgcgcg:
I'm assuming the same changes were made so ML can run on a sub-comm. How is that propagated to Ifpack? If there already was a way to use a sub-comm in Ifpack, why do we need a new method of setting a sub-comm?

cgcgcg:
TBH, since the Epetra stack packages will go away in the near future, it would make more sense to me to just mark these as allowable use of `MPI_COMM_WORLD`. There has been no testing of actually running any of this on a subcommunicator.

packages/ifpack/src/euclid/guards_dh.h
```diff
@@ -44,15 +44,15 @@
...
+#include "Ifpack_GlobalComm.h"
 
 /*
    This file defines the INITIALIZE_DH and FINALIZE_DH macros
 */
 
 #define  INITIALIZE_DH(argc, argv, help) \
             MPI_Init(&argc,&argv);  \
-            comm_dh = MPI_COMM_WORLD;    \
+            comm_dh = ifpack_get_global_comm();    \
```
JacobDomagala, line 55: Since this macro initializes MPI, I don't think using `MPI_COMM_WORLD` is incorrect here

packages/ifpack/src/euclid/mat_dh_private.c
```diff
@@ -1166,7 +1167,7 @@ partition_and_distribute_metis_private (Mat_dh A, Mat_dh * Bout)
   /* broadcast number of rows to all processors */
   if (myid_dh == 0)
     m = A->m;
-  MPI_Bcast (&m, 1, MPI_INT, 0, MPI_COMM_WORLD);
+  MPI_Bcast (&m, 1, MPI_INT, 0, ifpack_get_global_comm());
```
JacobDomagala, line 1170: Both `partition_and_distribute_private` and `partition_and_distribute_metis_private` are only being used within `readMat_par` function which has the following [documentation](https://github.com/trilinos/Trilinos/blob/5aaae1ada6fe1ce777e671a0ff84fdc4f0779406/packages/ifpack/src/euclid/mat_dh_private.h#L87-L92):
```c++
/* Next function is primarily (?) for testing/development/debugging.
   P_0 reads and partitions the matrix, then distributes 
   amongst the other processors.
*/
  extern void readMat_par (Mat_dh * Aout, char *fileType, char *fileName,
			   int ignore);
```

I don't see `readMat_par` being used anywhere in Trilinos

JacobDomagala:
Given the above, are we fine with just allowing the `MPI_COMM_WORLD` usage here?

cgcgcg (approved):

PR #12194, opened 2023-09-01
Panzer: Remove use of hardcoded MPI_COMM_WORLD

thearusable:
@trilinos/panzer


packages/panzer/adapters-stk/src/stk_interface/Panzer_STK_QuadraticToLinearMeshFactory.hpp
```diff
@@ -60,7 +61,7 @@ class QuadraticToLinearMeshFactory : public STK_MeshFactory {
 public:
 
   QuadraticToLinearMeshFactory(const std::string& quadMeshFileName,
-                          stk::ParallelMachine mpi_comm = MPI_COMM_WORLD,
```
cgcgcg, line 63: What's the problem with `MPI_COMM_WORLD` as a default arg?

thearusable: I wanted to align the global and default communicator for the entire package. I thought that this would be a behavior expected by the users. 
When we use `initialize_global_comm` to set a communicator for the package then this communicator will be also used as default for the objects created in the package.

This change isn't necessary so I can remove it if you want.

stmcgovern: @cgcgcg Is your preference to have this change reverted? 

FYI @thearusable

rppawlo: I agree with @cgcgcg . The point of the effort at removing raw MPI_COMM_WORLD use is to make sure a library can run on a subset of MPI processes when given a communicator. Panzer meets this requirement as the function directly accepts a communicator. MPI_COMM_WORLD is only used as a default. All this PR is really doing is moving the use of comm world into a separate function. Nothing has changed functionally and you have only added more code to the package and more indirection (having to look up another function). The only reason it might be worth doing is if the automated script in PR testing that tries to block MPI_COMM_WORLD use is having trouble differentiating valid use cases from real issues (I would argue the tool should be fixed in that case or allow users to review and mark code as valid use).

jwillenbring: @thearusable Does @rppawlo 's argument apply to changes proposed or applied to other packages?

stmcgovern: Thanks @rppawlo . Re: The github action will ignore uses of MPI_COMM_WORLD when the comment `CHECK: ALLOW MPI_COMM_WORLD` is present on the line.

thearusable: @cgcgcg @rppawlo - I updated the PR, now the usage of `MPI_COMM_WORLD`as default parameter is allowed. 

@jwillenbring - No, only for one package for which the PR is not yet created. I will apply similar fix before creating it. 

@stmcgovern - FYI

stmcgovern: thanks @thearusable!

jwillenbring:
@rppawlo Please review the changes to this PR to see if they are acceptable. We are targeting completion of the entire body of work for this by the end of the month, so we are hoping to wrap up the individual items by Sept 20 or so if possible.

packages/panzer/core/src/Panzer_PauseToAttach.hpp
```diff
@@ -74,7 +75,7 @@ namespace panzer {
 
   inline void pauseToAttach()
   {
-    pauseToAttach(MPI_COMM_WORLD);
```
cgcgcg, line 77: I think this actually needs to be `MPI_COMM_WORLD` to make sense.

packages/panzer/disc-fe/src/Panzer_FaceToElement_impl.hpp
```diff
@@ -100,7 +101,7 @@ void
 FaceToElement<LocalOrdinal,GlobalOrdinal>::
 initialize(panzer::ConnManager & conn)
 {
-  Teuchos::RCP<const Teuchos::Comm<int>> comm_world(new Teuchos::MpiComm< int>(MPI_COMM_WORLD));
+  Teuchos::RCP<const Teuchos::Comm<int>> comm_world(new Teuchos::MpiComm< int>(panzer::get_global_comm()));
```
cgcgcg, line 104: This is not the correct fix. There is a method just above that takes a `comm` argument. Just allow `MPI_COMM_WORLD` here.

packages/panzer/core/src/Panzer_GlobalComm.hpp
cgcgcg, line 1: remove

stmcgovern:
Thanks @cgcgcg - you are right. @cwschilly will clean this up and add the allow checks on the valid `MPI_COMM_WORLD` usages

packages/panzer/core/src/Panzer_PauseToAttach.hpp
cwschilly: Fixed

packages/panzer/disc-fe/src/Panzer_FaceToElement_impl.hpp
cwschilly: Fixed

packages/panzer/disc-fe/src/Panzer_FaceToElement_impl.hpp
```diff
@@ -55,6 +55,7 @@
 #include "Panzer_EdgeFieldPattern.hpp"
 #include "Panzer_FaceFieldPattern.hpp"
 #include "Panzer_ElemFieldPattern.hpp"
+#include "Panzer_GlobalComm.hpp"
```
cgcgcg, line 58: remove?

cwschilly: Fixed; sorry about that

cgcgcg (commented):
@rppawlo This looks good to me now.

PR #12196, opened 2023-09-01
Shylu: FastILU fixes to handle large matrices

jgfouca:
1) Add "fill blocks for input" option. We don't want to pay for block filling if the input matrix already has the block filled.
2) Remove the duplicate Crs copy in FastILU_Base
3) Big re-write of fillLogicalBlocks. The original impl had bool view of size blocks_per_row * nrows, which will use too much memory for large matrices. Instead, count each teams active blocks with a scratch view per team, creating a row_map and col_ids for blocks. This was a bit ugly because I needed to also have a version that worked when the number of threads was less than the block size.

@trilinos/shylu 
@trilinos/tpetra 
@trilinos/ifpack2 

## Motivation
Shylu FastILU with blockCrs' enabled will now work much better.

## Testing
Did the same soak test for the new fillLogicalBlocks as I did for the original PR.

Did performance tests with a huge input matrix on both CPU and GPU:
```
Machine: mappy, CPU (2xEpyc), 128 threads
Shylu  init: 2.02572
 Total Init: 30.2242
Shylu compute: 0.204193
Shylu applies: 0.03-0.04 each
 
Machine: Weaver:
Shylu  init: 1.58161
 Total Init: 21.016
Shylu compute: 0.386065
Shylu applies: 0.0066 each
```


jhux2:
@trilinos/ifpack2

jgfouca:
Thanks, @jhux2 . I forgot to tag ifpack2.

packages/ifpack2/src/Ifpack2_Details_FastILU_Base_def.hpp
```diff
@@ -448,6 +451,7 @@ Params::getDefaults()
   p.blockSize = 1;      // # of rows / thread, for SpTRSV
   p.blockCrs = false;   // whether to use block CRS for fastILU
   p.blockCrsSize = 1;   // block size for block CRS
+  p.fillBlocks = false; // whether input matrix needs to be filled
```
srajama1, line 454: I believe the input matrix will be filled by default. Adding zeroes is just a convenience.

srajama1 (approved):
Ifpack2 / ShyLU changes are pretty simple. If @csiefer2 is ok with Tpetra changes, this is good to go.

packages/ifpack2/src/Ifpack2_Details_FastILU_Base_def.hpp
jgfouca: @srajama1 , yeah, filling with zeroes is off by default. It was useful for testing against unblocked CRSs.

sebrowne:
@jgfouca try merging/rebasing to get #12199, I think that's the one that addressed your issue (seen in other PRs)

jgfouca:
@sebrowne , done. Thanks

sebrowne:
Ah, I guess I didn't need to set the retest label.  It will see the new commit and retest anyway.  Trying to be helpful and I end up doing redundant work....

jhux2:
Fyi, not sure if this got a @trilinos/tpetra review.

PR #12203, opened 2023-09-04
MueLu: Don't use hardcoded `MPI_COMM_WORLD`

JacobDomagala:
@trilinos/muelu 


packages/muelu/src/Utils/MueLu_GlobalComm.hpp
```diff
@@ -0,0 +1,75 @@
...
+static std::mutex mpi_mutex;
+static MPI_Comm Global_MueLu_Comm = MPI_COMM_WORLD; // CHECK: ALLOW MPI_COMM_WORLD
+
+inline void InitializeGlobalComm(MPI_Comm comm) {
+  std::lock_guard<std::mutex> guard(mpi_mutex);
+  Global_MueLu_Comm = comm;
+}
+
+inline MPI_Comm GetGlobalComm() {
+  std::lock_guard<std::mutex> guard(mpi_mutex);
```
jhux2, line 68: Why is mutex required here?

cgcgcg (commented):
Wouldn't it make more sense to use the already existing wrappers in Teuchos? Or is there a reason why that doesn't work?

JacobDomagala:
> Wouldn't it make more sense to use the already existing wrappers in Teuchos? Or is there a reason why that doesn't work?

The idea is that user can use MueLu within custom communicator, i.e.:
```c++
MPI_Comm newComm; 
MPI_Comm_split(/*...*/, &newComm);
MueLu::InitializeGlobalComm(newComm);

/*
   Use MueLu's code 
*/
```

Teuchos provides a DefaultComm, which wraps `MPI_COMM_WORLD` (for MPI build) and doesn't allow for user to change it (unless I'm missing something)

cgcgcg:
MueLu can run on a user provided communicator, provided in the form of a Teuchos communicator, no?

packages/muelu/src/Utils/MueLu_Memory.cpp
```diff
@@ -121,9 +122,9 @@ namespace MueLu {
       double localTime = timer.totalElapsedTime();
 #ifdef HAVE_MPI
       int ntimers=1, root=0;
-      MPI_Reduce(&localTime,&maxTime,ntimers,MPI_DOUBLE,MPI_MAX,root,MPI_COMM_WORLD);
-      MPI_Reduce(&localTime,&minTime,ntimers,MPI_DOUBLE,MPI_MIN,root,MPI_COMM_WORLD);
-      MPI_Reduce(&localTime,&avgTime,ntimers,MPI_DOUBLE,MPI_SUM,root,MPI_COMM_WORLD);
+      MPI_Reduce(&localTime,&maxTime,ntimers,MPI_DOUBLE,MPI_MAX,root,MueLu::GetGlobalComm());
+      MPI_Reduce(&localTime,&minTime,ntimers,MPI_DOUBLE,MPI_MIN,root,MueLu::GetGlobalComm());
+      MPI_Reduce(&localTime,&avgTime,ntimers,MPI_DOUBLE,MPI_SUM,root,MueLu::GetGlobalComm());
 #else
```
cgcgcg, line 128: This should probably just use the provided comm instead of `MPI_COMM_WORLD`.

packages/muelu/adapters/amgx/MueLu_AMGXOperator_decl.hpp
```diff
@@ -241,7 +242,7 @@ namespace MueLu {
 #ifdef NEW_COMM
         AMGX_resources_create(&Resources_, Config_, &mpiComm, 1/* number of GPU devices utilized by this rank */, device);
 #else
-        AMGX_resources_create(&Resources_, Config_, MPI_COMM_WORLD, 1/* number of GPU devices utilized by this rank */, device);
+        AMGX_resources_create(&Resources_, Config_, MueLu::GetGlobalComm(), 1/* number of GPU devices utilized by this rank */, device);
 #endif
```
cgcgcg, line 246: @csiefer2 Would it make more sense to remove the second code path since it's not compiled anyway?

packages/muelu/src/MueCentral/MueLu_VerboseObject.cpp
```diff
@@ -140,7 +141,7 @@ namespace MueLu {
     int mpiStarted = 0; MPI_Initialized(&mpiStarted);
     if (mpiStarted)     {
       int procRank;
-      MPI_Comm_rank(MPI_COMM_WORLD, &procRank);
+      MPI_Comm_rank(MueLu::GetGlobalComm(), &procRank);
       fn = filename + "." + std::to_string(procRank);
```
cgcgcg, line 145: This doesn't really fix the issue but just moves the `MPI_COMM_WORLD` to a different header.

jhux2: It's unclear if the use of MPI_COMM_WORLD is a problem here.

lucbv: We could change the constructor of VerboseObject to accept a Teuchos::Comm object?

cgcgcg: Doesn't every single MueLu class derive from `VerboseObject`?

cgcgcg: - Is `numProcs_` actually used by anything?
- Could we skip the bit with `MPI_COMM_WORLD` and instead set `procRank_` once we have a communicator? We already do that in `CreateXpetraPreconditioner` anyway..

jhux2: `grep` shows that `numProcs_` isn't used anywhere.

> Could we skip the bit with MPI_COMM_WORLD and instead set procRank_ once we have a communicator? We already do that in CreateXpetraPreconditioner anyway..

I *think* this would be fine.

cgcgcg:
@JacobDomagala Do all the suggestions make sense?

packages/muelu/src/MueCentral/MueLu_VerboseObject.cpp
JacobDomagala: Do you want to set the `procRank` in every class that inherits from `BaseClass`? 
If you feel like the use of `MPI_COMM_WORLD` is justified here, we can just add `// CHECK: ALLOW MPI_COMM_WORLD` to those lines.

sebrowne:

packages/muelu/adapters/amgx/MueLu_AMGXOperator_decl.hpp
JacobDomagala: If we were to remove it or use `Teuchos::DefaultComm` here, then I'd just remove the newly added `MueLu::GlobalComm` stuff. 
What do you think?

packages/muelu/src/MueCentral/MueLu_VerboseObject.hpp
```diff
@@ -133,9 +133,7 @@ namespace MueLu {
   private:
     //! Verbose level specific to 'this'
     VerbLevel verbLevel_;
-    mutable
-      int procRank_;
-    int numProcs_;
+    static inline int procRank_ = -1;
```
JacobDomagala, line 136: Actually I don't think making it static is correct here. Wouldn't we want different objects to have different verbose ranks?

We could change `SetMueLuOFileStream` (the reason why I made this static) to either take rank as an additional input, or use `Teuchos::DefaultComm` to obtain current rank?

What do you think?

JacobDomagala: `SetMueLuOFileStream` is static so we can't use the non-static `procRank_` there. It used to get the rank using `MPI_COMM_WORLD` which I've changed (initially using the `MueLu::GlobalComm` and now to use the verbose rank (which enforces procRank_ to be static aswell)

cgcgcg: Ah. I think it might actually be good to use filenames that are labeled using the global communicator. We have spots where we reduce the problem to a sub-communicator. So some ranks could be rank k on the global communicator and rank j in the sub-communicator. And we probably want that output to end up in the same file.

jhux2: > Ah. I think it might actually be good to use filenames that are labeled using the global communicator. We have spots where we reduce the problem to a sub-communicator. So some ranks could be rank k on the global communicator and rank j in the sub-communicator. And we probably want that output to end up in the same file.

I agree.  Rank with respect to MPI_COMM_WORLD is the only thing that makes sense to me for post-run analysis.

packages/muelu/src/MueCentral/MueLu_VerboseObject.cpp
```diff
@@ -139,9 +121,9 @@ namespace MueLu {
 #ifdef HAVE_MPI
     int mpiStarted = 0; MPI_Initialized(&mpiStarted);
     if (mpiStarted)     {
-      int procRank;
-      MPI_Comm_rank(MPI_COMM_WORLD, &procRank);
-      fn = filename + "." + std::to_string(procRank);
+      MPI_Comm rank;
```
cgcgcg, line 124: Huh?

JacobDomagala: Woops! Fixed : )

cgcgcg (approved):

jwillenbring:
@cgcgcg Can this be merged?

cgcgcg:
We will run some additional tests outside of the autotester before we merge. Is there a deadline for this to be delivered?

cgcgcg (changes requested):
I tested the PR by running `mpiexec -n 4 ./PanzerMiniEM_BlockPrec.exe` from `packages/panzer/mini-em/example/BlockPrec/`. The output for some ranks is missing. So something isn't right yet.

JacobDomagala:
With the changes I made, miniEM gives a bit more verbose output (I don't have the idea why - I don't know `MueLu` that well) 

Let me know if you're fine with the changes made so far. 
Also, if you're able to quickly spot the change which made it extra verbose, I'd appreciate that, otherwise I'll continue the investigation.

JacobDomagala:
If you feel like using the global communicator is fine there, we can always ignore the use of `MPI_COMM_WORLD`

JacobDomagala:
Ok, after a bit of analysis: 

- Extra verbosity comes from rank 3 which is being used for Hierarchy22_ (it is using A22_ with subcommunuicator of size 1 -> only rank 3 is being part of it)

```
Partition distribution over cores (ownership is indicated by '+')
..      0:1
.+      2:3
```

This is the place in code where it happends (I added comments with my reasoning)
[MueLu_RefMaxWell_def.hpp](https://github.com/trilinos/Trilinos/blob/5aaae1ada6fe1ce777e671a0ff84fdc4f0779406/packages/muelu/src/Operators/MueLu_RefMaxwell_def.hpp#L817C1-L868C58)
```c++
        // temporarily set procRank_ to 0 on rank 3 (as it's root rank for A22_ operator)
        int oldRank = SetProcRankVerbose(A22_->getDomainMap()->getComm()->getRank()); 
        (...)
          if (!reuse) {
            (...)
            // Everything created here will have procRank_ of A22_
            Hierarchy22_ = MueLu::CreateXpetraPreconditioner(A22_, precList22_);
          } else {
            RCP<MueLu::Level> level0 = Hierarchy22_->GetLevel(0);
            level0->Set("A", A22_);
            Hierarchy22_->SetupRe();
          }
        }
        // Set the proc rank to old one (3 in that case)
        SetProcRankVerbose(oldRank);
```

@cgcgcg @jhux2  Let me know if my analysis is correct.

cgcgcg:
@JacobDomagala I think this is probably the correct spot. What happens there is that we moved an operator to a sub-communicator and then set up a multigrid hierarchy on that subcommunicator. For the construction on that subcommunicator, we called `setProcRankVerbose` with ranks wrt the subcommunicator.

jwillenbring:
@cgcgcg Do you know the status of this?

cgcgcg:
I do not. As it stands, this PR is not ready to be merged.

ccober6:
@cgcgcg, what would be the best path forward on this?  Further debugging, different approach, ...?

cgcgcg:
If there is nobody working on it we might as well close it.

JacobDomagala:
@cgcgcg I'm planning on working on it, been quite busy with other projects lately, sorry about that.

ccober6:
> If there is nobody working on it we might as well close it.

@cgcgcg, we need to fix this as it is not compliant with xSDK policy (M3) (https://sems-atlassian-son.sandia.gov/jira/browse/TRILINOS-256).  @JacobDomagala do you have a timeline on this?  I know this is outside the current contract, so I figured someone else might need to pick this up?

cgcgcg:
It's missing output when you run `mpiexec -n 4 ./PanzerMiniEM_BlockPrec.exe`.

JacobDomagala:
Yes I'm still working on it, I only had a chance to resolve the conflicts with recent changes to MueLu.

JacobDomagala:
@cgcgcg This PR is at the point where `miniEM` (on 4 ranks) prints all the information that it currently does on develop, plus the hierarchy setup (that we discussed above). I've emailed you the logs produced on my machine. 
128/128 MueLu tests pass for me as well. 

Let me know what you think.

PR #12204, opened 2023-09-05
Isorropia: Remove use of hardcoded MPI_COMM_WORLD

thearusable:
@trilinos/isorropia

Remove hardcoded `MPI_COMM_WORLD` and instead use the global communicator which was set in Zoltan package by using `zoltan_initialize_global_comm`.

Depends on https://github.com/trilinos/Trilinos/pull/12195


egboman (approved):
Looks good. Thanks.

PR #12213, opened 2023-09-06
Set variable Trilinos_MUST_FIND_ALL_TPL_LIBS to TRUE

### LINKED ISSUE #10345: Set Trilinos_MUST_FIND_ALL_TPL_LIBS=TRUE by default for Trilinos 15.0
@trilinos/framework

With Trilinos 14.0 coming up, this would be a good time to change the default for `Trilinos_MUST_FIND_ALL_TPL_LIBS` from `OFF` to `ON`.  (By setting `Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT` to `ON` as described [here](https://tribits.org/doc/TribitsUsersGuide.html#project-name-must-find-all-tpl-libs)).  It was turned off by default because it breaks backward compatibility but it causes problems for new users and new configurations (see, for example, https://github.com/trilinos/Trilinos/issues/10322#issuecomment-1071301714).

So as soon as the Trilinos 13.4 branch is created, we should make this change in the 'develop' branch.

srbdev:
@trilinos/framework 

## Motivation
See #10345 

## Related Issues

* Closes #10345


CMakeLists.txt
```diff
@@ -105,6 +105,9 @@ INSTALL_BUILD_STATS_SCRIPTS()
 # Install TriBITS so that other projects can use it
 include(SetupTribitsInstall)
 
+# See https://tribitspub.github.io/TriBITS/users_guide/index.html#project-name-must-find-all-tpl-libs
+SET(Trilinos_MUST_FIND_ALL_TPL_LIBS TRUE)
```
bartlettroscoe, line 109:
```suggestion
SET(Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT TRUE)
```

As per:

* https://tribitspub.github.io/TriBITS/users_guide/index.html#project-name-must-find-all-tpl-libs

There are several problems with setting an non-cache var like this when there is also a cache var of the same name.

bartlettroscoe (changes requested):
Also, a release note for this needs to be added to the file [RELEASE_NOTES](https://github.com/trilinos/Trilinos/blob/develop/RELEASE_NOTES).

Also, it seems we missed the release notes for 14.4 about building against external Kokkos.

srbdev:
@sebrowne see release notes under the Files changed tab.

bartlettroscoe (dismissed):
Looks good.  Thanks!

FYI: My guess is that this PR will cause a merge conflict with the other PR #12214 (once it adds release notes, for sure).  This is an example of where it might have made sense to  combine these two changes into one PR.  That avoids expected conflicts and it saves on computing resources to run the PR builds.

srbdev:
> Looks good. Thanks!
> 
> FYI: My guess is that this PR will cause a merge conflict with the other PR #12214 (once it adds release notes, for sure). This is an example of where it might have made sense to combine these two changes into one PR. That avoids expected conflicts and it saves on computing resources to run the PR builds.

I hadn't thought about updating the `RELEASE_NOTES` file for this until you mentioned it, so I realized there would a merge conflict between the two PRs only at that point.

sebrowne:
This branch MUST be rebased prior to merge given the 'develop' change (re-adds forbidden files).

sebrowne (changes requested):
Please rebase to eliminate extraneous commits.

srbdev:
Closing. See https://github.com/trilinos/Trilinos/pull/12258 instead.

PR #12214, opened 2023-09-06
Set non-cache variable Trilinos_USE_GNUINSTALLDIRS to ON

### LINKED ISSUE #12104: Set Trilinos_USE_GNUINSTALLDIRS=ON for Trilinos 15.0
CC: @sebrowne, @ccober6 

## Description

TriBITS has had the ability to use that paths selected by the standard CMake module [GNUInstallDirs.cmake](https://cmake.org/cmake/help/v3.23/module/GNUInstallDirs.html) for a **long** time (see [Setting the install prefix](https://docs.trilinos.org/files/TrilinosBuildReference.html#setting-the-install-prefix)).  But it is turned off in TriBITS by default (for the sake of backward compatibility) and it was never turned on in Trilinos by default (also to maintain backward compatibility).

Therefore, in the goal to move Trilinos and TriBITS to modern CMake ( see TriBITSPub/TriBITS#411), we should hard set the **non-cache** variable `Trilinos_USE_GNUINSTALLDIRS=ON` in the base Trilinos `CMakeLists.txt` file.  This will force every Trilinos user to absorb this change in Trilinos 15.0 (but most customers should never even notice).

Once Trilinos 15.0 comes out, then we can refactor TriBITS to just directly use the standard variables that come from the standard CMake module `include(GNUInstallDirs)`.

Easy as pie (at least for Trilinos and TriBITS).

srbdev:
@trilinos/framework 

## Motivation
See #12104 

## Related Issues

* Closes #12104


CMakeLists.txt
```diff
@@ -91,6 +91,9 @@ SET(${PROJECT_NAME}_ENABLE_CPACK_PACKAGING_DEFAULT ON)
 # Don't allow disabled subpackages to be excluded from tarball
 SET(${PROJECT_NAME}_EXCLUDE_DISABLED_SUBPACKAGES_FROM_DISTRIBUTION_DEFAULT FALSE)
 
+# See https://docs.trilinos.org/files/TrilinosBuildReference.html#setting-the-install-prefix
+SET(Trilinos_USE_GNUINSTALLDIRS ON)
```
bartlettroscoe, line 95:
```suggestion
SET(Trilinos_USE_GNUINSTALLDIRS_DEFAULT ON)
```

As per:

* https://tribitspub.github.io/TriBITS/users_guide/index.html#project-name-use-gnuinstalldirs

this has to set the non-cache var `Trilinos_USE_GNUINSTALLDIRS_DEFAULT` **not** `Trilinos_USE_GNUINSTALLDIRS`.

bartlettroscoe (changes requested):
Please switch to the suggestion and add a RELEASE_NOTES entry for this break in backward compatibility.

bartlettroscoe (changes requested):
As this breaks backward compatibility, this needs a release note like was added for PR #12213.

sebrowne:
This branch MUST be rebased prior to merge given the 'develop' change (re-adds forbidden files).

srbdev:
Closing. See https://github.com/trilinos/Trilinos/pull/12258 instead.

PR #12221, opened 2023-09-07
teuchos(cleaning): remove old type traits is_same

romintomasetti:
@trilinos/teuchos

## Motivation

Very small PR to remove `Teuchos::TypeTraits::is_same` (now that `Trilinos` is on `C++17`).

@bartlettroscoe


bartlettroscoe (approved):
@romintomasetti, thanks for putting this in!  This looks great.  This might break some Trilinos customer code if they were using this type traits class, but since the next Trilinos release 15.0 will be coming out next, a break in backward compatibility is fine.

PR #12230, opened 2023-09-08
Pamgen: Remove use of hardcoded MPI_COMM_WORLD

thearusable:
@trilinos/pamgen

### DIFF
--- a/packages/pamgen/extra_functions/pamgen_extras.cpp
+++ b/packages/pamgen/extra_functions/pamgen_extras.cpp
@@ -8,6 +8,7 @@
 
 #ifdef HAVE_MPI
 #include <mpi.h>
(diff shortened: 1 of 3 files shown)


jwillenbring:
@dmhensi If you would like to review this, please do, otherwise if we don't hear from you we will assume that someone else should review this. We are targeting completion of these PRs prior to the end of this month, so please reply by Sept 20 if you intend to review this. Thanks.

cwschilly:
This PR needs retesting due to autotester failure last week

jwillenbring:
@sebrowne Do you see why this failed?

jwillenbring (approved):
These changes are consistent with the approach used across Trilinos.

PR #12219, opened 2023-09-07
amesos2(cleaning): removing old type traits that are not supported by C++17

romintomasetti:
@trilinos/amesos2

## Motivation

I'm still working on that bug I mentioned in #12209.

I've removed old type traits that are now fully supported by C++17. Enjoy!

@iyamazaki

PR #12209, opened 2023-09-06
amesos2: solve style and performance warnings

romintomasetti:
@trilinos/amesos2

## Motivation

We're currently working on a bug of `KLU2`. I made these changes while working on it. Here they are, for free :smile:

It is basically some missing `override` and some `const std::string&`.

PR #12232, opened 2023-09-08
amesos2(KLU2): enhance standalone testing

romintomasetti:
@trilinos/amesos2

@iyamazaki @srajama1 

## Motivation

As mentioned in #12209 and #12219, @maartenarnst and myself are tracking a bug in KLU2 that only happens for `gcc-12` (and also `gcc-13`...) in `Release` mode (basically `-O3`).

The problem was that the function `KLU_lsolve` (in `packages/amesos2/src/KLU2/Source/klu2.hpp`) was simply optimized away by the compiler in that specific configuration, though it has clear **side effects/observable effects** when called from `KLU_solve2` in `packages/amesos2/src/KLU2/Source/klu2_solve.hpp`.

Adding something like the following `asm("")`, or adding a `printf` anywhere in `KLU_lsolve` would prevent the compiler from optimizing it away. So the first fix was something like this:
```c++
 #if defined(__GNUC__) && __GNUC__ >= 12
     asm("");
 #endif
```

However, we judged that fix ugly enough that we looked for something nicer and came up with adding a `default` to the `switch`:
```c++
default:
            throw std::domain_error("More than 4 right-hand sides is not supported.");
```
and that would also prevent `gcc-12` removing `KLU_lsolve`.

We really think it is a compiler bug. *It also happens with `gcc-13`*. Should we report a bug to them? Note that we were not able to come up with a minimal reproducer.


packages/amesos2/src/KLU2/Source/CMakeLists.txt
```diff
@@ -102,7 +102,7 @@ klu2_memory.hpp
   )
 
 #
-# C) Define the targets for package's example
+# C) Define the targets for package's test
 #
 
-TRIBITS_ADD_EXAMPLE_DIRECTORIES(examples)
+TRIBITS_ADD_TEST_DIRECTORIES(tests)
```
romintomasetti, line 108: Note for the reviewer(s): I promoted the example to being a "standalone test" of the `KLU2` sources.

packages/amesos2/src/KLU2/Source/tests/klu2_simple.cpp
```diff
@@ -0,0 +1,183 @@
+// @HEADER
```
romintomasetti, line 1: Also note for the reviewer(s): I used `git mv` bug there are too many changes so `git` considers it deleted and re-created.

packages/amesos2/src/KLU2/Include/klu2_internal.h
```diff
@@ -198,67 +198,6 @@ size_t KLU_kernel_factor            /* 0 if failure, size of LU if OK */
     KLU_common<Entry, Int> *Common  /* the control input/output structure */
 ) ;
 
-template <typename Entry, typename Int>
```
romintomasetti, line 201: Note for the reviewer(s): Here I deleted the forward declarations because there is no ETI for these. I guess in the very first version it was here a header file, and `klu2.hpp` was actually a `.cpp` file. And then it was changed to become templated.

sebrowne:

srajama1 (dismissed):
Thanks @romintomasetti and @maartenarnst !

maartenarnst:
Hi @srajama1. It seems some of the tests failed for this one. On our side, @romintomasetti and I don't have permissions to look into the CDash or ask for a retest. Would you have a moment to do one of those things to move this forward? Thanks.

iyamazaki:
@maartenarnst Thank you for the PR, and I am sorry for missing it. The errors seem to happen when building the simple test with the complex scalar type, and I could reproduce them locally, e.g.,

````
Trilinos/packages/amesos2/src/KLU2/Source/klu2_scale.hpp(166): error: no suitable conversion function from "std::complex<double>" to "double" exists
              KLU2_ABS (a, Az [p]) ;
          detected during:
            instantiation of "Int klu_scale(Int, Int, Int *, Int *, Entry *, double *, Int *, klu_common<Entry, Int> *) [with Entry=std::complex<double>, Int=int]" at line 151 of "Trilinos/packages/amesos2/src/KLU2/Source/klu2_factor.hpp"
            instantiation of "void factor2(Int *, Int *, Entry *, klu_symbolic<Entry, Int> *, klu_numeric<Entry, Int> *, klu_common<Entry, Int> *) [with Entry=std::complex<double>, Int=int]" at line 562 of "Trilinos/packages/amesos2/src/KLU2/Source/klu2_factor.hpp"
            instantiation of "klu_numeric<Entry, Int> *klu_factor(Int *, Int *, Entry *, klu_symbolic<Entry, Int> *, klu_common<Entry, Int> *) [with Entry=std::complex<double>, Int=int]" at line 147 of "Trilinos/packages/amesos2/src/KLU2/Source/tests/klu2_simple.cpp"
            instantiation of "void solve_and_check<T,D,tol_t>(tol_t) [with T=std::complex<double>, D=int, tol_t=double]" at line 180 of "Trilinos/packages/amesos2/src/KLU2/Source/tests/klu2_simple.cpp"
````

maartenarnst:
Hi @iyamazaki. Thanks for looking into it and letting us know about the error. It seems we weren't protecting the test for the case of the complex scalar type with the `HAVE_TEUCHOS_COMPLEX` macro. I've added it. Would you have a moment to set it up for a retest? Thanks in advance!

iyamazaki:
Thank you for the change @maartenarnst! I just want to double-check with @ndellingwood. I feel a bit strange that we have to protect Amesos2 tester with `HAVE_TEUCHOS_COMPLEX`?

maartenarnst:
Hi @iyamazaki.  

OK of course to check with @ndellingwood.

Just a few thoughts to try to help the discussion:

The use of `HAVE_TEUCHOS_COMPLEX` is not new. It was already in the file `KLU2/Source/Examples/klu2_simple.cpp` that this PR turns into a test. In this file in the current develop branch:

- https://github.com/trilinos/Trilinos/blob/95b79420e8a711fdd3eeb3da7dd648ba8dbc844f/packages/amesos2/src/KLU2/Source/examples/klu2_simple.cpp#L92-L109

If we don't use `HAVE_TEUCHOS_COMPLEX` to protect the test, we run into the test failures that we had seen. The reason seems to be that there are places in the main KLU2 code where `HAVE_TEUCHOS_COMPLEX` is used. It seems that if we remove this protection at one location, such as the example/test, we'll have to be consistent and remove it systematically.

- https://github.com/trilinos/Trilinos/blob/95b79420e8a711fdd3eeb3da7dd648ba8dbc844f/packages/amesos2/src/KLU2/Include/klu2_scalartraits.h#L91-L95

@romintomasetti authored this PR. When he and I talked this morning, he too expressed that he felt that it is strange to use the `HAVE_TEUCHOS_COMPLEX` in the test. I agree with both of you. It's just that if we change it, there may be quite a few other things that we'll have to change.

ndellingwood (approved):
Thanks for the contributions!
Yeah, adding the `HAVE_TEUCHOS_COMPLEX` to guard the test seems the easiest way to handle the existing guarding mentioned by @maartenarnst at e.g. [Trilinos/packages/amesos2/src/KLU2/Include/klu2_scalartraits.h](https://github.com/trilinos/Trilinos/blob/95b79420e8a711fdd3eeb3da7dd648ba8dbc844f/packages/amesos2/src/KLU2/Include/klu2_scalartraits.h#L91-L95)

Adding the pre-test-inspection and automerge label so we can get this in, thanks again

PR #12233, opened 2023-09-08
Extend Ifpack2 BlockTriDi example to enable repeating compute & solve…

vbrunini:
… multiple times.

To get more reliable timings. Also add a separate stacked timer for compute.

@trilinos/ifpack2
@kliegeois @srajama1


kliegeois (approved):
It looks good to me, thanks!

PR #12239, opened 2023-09-11
Stokhos: Add ensemble versions of `fmin` and `max`

maartenarnst:
@etphipp 

This is a small PR that adds ensemble versions of the functions `fmin` and `fmax`.


etphipp (approved):
Looks good, thanks @maartenarnst!

PR #12241, opened 2023-09-11
Belos: Update all *_diag.cpp tests to use tpetra

cwschilly:
@trilinos/belos 
@hkthorn 

## Motivation
This PR creates Tpetra versions of the following three (out of five) `*_diag.cpp` tests in Belos. 

- [x] `test_bl_gmres_multidiag.cpp`
- [x] `test_bicgstab_diag.cpp`
- [x] `test_bl_gmres_diag.cpp`
- [ ] `test_minres_diag.cpp`
- [ ] `test_tfqmr_diag.cpp`

The remaining two tests will be included in future PRs that convert the entire MINRES and TFQMR test folders.


cwschilly:
A few points regarding `test_bl_gmres_multidiag.cpp`:

1. While the name of the file suggests it uses the block GMRES to solve, the actual implementation (in the original Epetra version) used the pseudo-block GMRES solver. This is what I used in the Tpetra version. Should I change the name of the file(s) to reflect this (eg `test_pseudo_gmres_multidiag.cpp`)?
2. There was no existing `MultiMatrixSolve` folder in the `tpetra` directory, so I put this test in the existing `MultipleSolves` directory.
3. I created a duplicate "`BelosLinearMultiShiftProblem.hpp`" (currently in `epetra/test/MultiMatrixSolve`) and placed it in the `MultipleSolves` directory (`tpetra/test/MultiplesSolves`)

Please let me know if I should make any changes based on these items.

hkthorn:
> 1. While the name of the file suggests it uses the block GMRES to solve, the actual implementation (in the original Epetra version) used the pseudo-block GMRES solver. This is what I used in the Tpetra version. Should I change the name of the file(s) to reflect this (eg test_pseudo_gmres_multidiag.cpp)?

This solution strategy can only apply to the pseudo block GMRES solver since it requires the subspaces be independent and not block orthogonalized.  So, sure, you can change the name of the file to reflect this.

> 2. There was no existing MultiMatrixSolve folder in the tpetra directory, so I put this test in the existing MultipleSolves directory.

No, the MultiMatrixSolve concept is different than MultipleSolves (which uses the solver factory), so they should be in different directories.  

> 3. I created a duplicate "BelosLinearMultiShiftProblem.hpp" (currently in epetra/test/MultiMatrixSolve) and placed it in the MultipleSolves directory (tpetra/test/MultiplesSolves)

Copy that over the generic "BelosLinearMultiShiftProblem.hpp" to the new MultiMatrixSolve directory under tpetra/test.  This is a special implementation of a linear problem that illustrates/tests that a user can create a specialized LinearProblem for solving linear systems that have specific form.  Again, this is different than the solver factory test in tpetra/test/MultipleSolves.

cwschilly:
Thank you for your feedback @hkthorn , I have made the requested changes. Please let me know if there's anything else I should fix.

hkthorn (approved):
The 3 *_diag.cpp tests that are included in this PR LGTM!  Thanks!

PR #12242, opened 2023-09-11
Anasazi:  Fix OrthoManager tests for Tpetra and non-Petra

hkthorn:
@trilinos/anasazi 

## Motivation

The OrthoManager tests were effectively disabled due to the use of an ifdef 
that is undefined.  This has been fixed and the tests have been updated since 
they have been disabled for some time.

## Testing

OSX Clang serial and parallel


sebrowne:

hkthorn:
@sebrowne I've rebased the develop branch of my fork, but there are these extraneous commits.  The changes made in this PR are the correct ones.  There are no file changes outside of those from the original PR, but how do I get rid of these extraneous commits?  Will a squash work here?

sebrowne:
@hkthorn it looks like you merged trilinos:develop into your develop (at least the way I read GitHub).  I think for this to work the best (and to prevent issues going forwards), you should git rebase and then force push to your develop branch.

hkthorn:
@sebrowne Got it!  This PR is fixed now!

packages/anasazi/test/OrthoManager/cxx_main_complex.cpp
```diff
@@ -127,8 +127,7 @@ int main(int argc, char *argv[])
     }
 
     // below we will assume that sizeX1 > sizeX2
-    // this does not effect our testing, since we will test P_{X1,Y1} P_{X2,Y2} as well as P_{X2,Y2} P_{X1,Y1}
```
jennloe, line 130: Why delete this line?

jennloe (approved):

packages/anasazi/test/OrthoManager/cxx_main_complex.cpp
hkthorn: This is copied over from the generalized test.  There isn't Y1, Y2 basis for this test, so the comment is not correct.

PR #12244, opened 2023-09-11
Draft: SEACAS: Update embedded lib::fmt to 1.10.1

gsjaardema:
@trilinos/seacas 

## Motivation
The embedded version of lib::fmt currently used in SEACAS is 9.1.  Update to the current version that is being used externally in the SEACAS standalone build to  avoid differences in behavior.


sebrowne:

Issue #12048, opened 2023-07-13
ShyLU: ShyLU/ShyLUConfig.cmake:164 (include): could not find requested file: ../ShyLU_Node/ShyLU_NodeConfig.cmake


eugeneswalker:
## Bug Report
@srajama1 

### Description
Trilinos@13.4.1 built with `+shylu` using Spack contains a ShyLUConfig.cmake that wants to include a non-existent `ShyLU_NodeConfig.cmake`.

```
$> sed -n '164p' $(spack location -i trilinos+shylu)/lib/cmake/ShyLU/ShyLUConfig.cmake
include("${CMAKE_CURRENT_LIST_DIR}/../ShyLU_Node/ShyLU_NodeConfig.cmake")

$> file $(spack location -i trilinos+shylu)/lib/cmake/ShyLU/../ShyLU_Node/ShyLU_NodeConfig.cmake
...  (No such file or directory)

$> find $(spack location -i trilinos+shylu) -type f -name ShyLU_NodeConfig.cmake
$>
```

Here is the exact `trilinos@13.4.1 +shylu` that was installed:
```
$> spack find -lv trilinos+shylu
-- linux-ubuntu20.04-x86_64 / gcc@11.1.0 ------------------------
fwbxvsw trilinos@13.4.1~adelus~adios2+amesos+amesos2+anasazi+aztec~basker+belos+boost~chaco~complex~cuda~cuda_rdc~debug~dtk+epetra+epetraext~epetraextbtf~epetraextexperimental~epetraextgraphreorderings~exodus+explicit_template_instantiation~float+fortran~gtest~hdf5~hypre+ifpack+ifpack2+intrepid+intrepid2~ipo+isorropia+kokkos~mesquite+minitensor+ml+mpi+muelu~mumps+nox~openmp~panzer+phalanx+piro~python~rocm~rocm_rdc+rol+rythmos+sacado~scorec+shards+shared+shylu+stk+stokhos+stratimikos~strumpack~suite-sparse~superlu+superlu-dist+teko+tempus+thyra+tpetra+trilinoscouplings~wrapper~x11+zoltan+zoltan2 build_system=cmake build_type=Release cxxstd=14 generator=make gotype=long_long
==> 1 installed package
```

### Steps to Reproduce
Using `spack@develop` (`2dc76248d31` from `Fri Jul 7 23:27:32 2023 +0200`) install the following concrete spec.

[...]

PR #12245, opened 2023-09-12
shylu: make shylu_node lib optional

ndellingwood:
Address issue #12048
shylu_node solvers are optional requirements; by making shylu_node an optional requirement of shylu, we prevent missing cmake config files when using the Trilinos_ENABLE_Shylu option, but not explicitly enabling a solver from shylu_node

@trilinos/shylu 

## Testing
Checked a local install to confirm that use of `-DTrilinos_ENABLE_ShyLU=ON` did not result in an empty `<install-path>/lib/cmake/ShyLU_Node` directory (the directory is not present)


sebrowne:

ndellingwood:
Rebased

iyamazaki (approved):
@ndellingwood Thank you for the change, I am very sorry that I missed it.

PR #12258, opened 2023-09-13
Famework: Change defaults for Trilinos-specific CMake flags

### LINKED ISSUE #10345: Set Trilinos_MUST_FIND_ALL_TPL_LIBS=TRUE by default for Trilinos 15.0

### LINKED ISSUE #12104: Set Trilinos_USE_GNUINSTALLDIRS=ON for Trilinos 15.0

srbdev:
@trilinos/framework 

## Motivation
See #10345 and #12104

## Related Issues

* Closes #10345 
* Closes #12104

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,6 +1,23 @@
 
 ###############################################################################
 #                                                                             #
+# Trilinos Release 15.0 Release Notes                                         #
+#                                                                             #
+###############################################################################
+
+CMake
+
+  - Change the default for `Trilinos_MUST_FIND_ALL_TPL_LIBS` from `OFF` to
+    `ON`, by setting `Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT` to `ON`. It was
+    turned off by default because it breaks backward compatibility but it also
+    causes problems for new users and new configurations.
+
+    Users that do not want this new behavior can set
+    `-D Trilinos_MUST_FIND_ALL_TPL_LIBS=OFF`, which is backward compatible.
+
+
+###############################################################################
+#                                                                             #
 # Trilinos Release 14.2 Release Notes                                         #
 #                                                                             #
 ###############################################################################
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -16,7 +16,7 @@ CMake
     `-D Trilinos_MUST_FIND_ALL_TPL_LIBS=OFF`, which is backward compatible.
 
   - Change the default for `Trilinos_USE_GNUINSTALLDIRS` from `OFF` to `ON`,
-    by setting `Trilinos_USE_GNUINSTALLDIRS_DEFAULT` to `ON`, in the goal to
+    in the goal to
     move Trilinos and TriBITS to modern CMake.
 
     TriBITS has had the ability to use that paths selected by the standard
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -15,6 +15,20 @@ CMake
     Users that do not want this new behavior can set
     `-D Trilinos_MUST_FIND_ALL_TPL_LIBS=OFF`, which is backward compatible.
 
+  - Change the default for `Trilinos_USE_GNUINSTALLDIRS` from `OFF` to `ON`,
+    by setting `Trilinos_USE_GNUINSTALLDIRS_DEFAULT` to `ON`, in the goal to
+    move Trilinos and TriBITS to modern CMake.
+
+    TriBITS has had the ability to use that paths selected by the standard
+    CMake module `GNUInstallDirs.cmake` for a long time. But it is turned off
+    in TriBITS by default and was never turned on in Trilinos, both for the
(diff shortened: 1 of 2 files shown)


sebrowne (dismissed):

RELEASE_NOTES
```diff
@@ -1,4 +1,35 @@
...
+###############################################################################
+#                                                                             #
+# Trilinos Release 15.0 Release Notes                                         #
+#                                                                             #
+###############################################################################
+
+CMake
+
+  - Change the default for `Trilinos_MUST_FIND_ALL_TPL_LIBS` from `OFF` to
+    `ON`, by setting `Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT` to `ON`. It was
```
bartlettroscoe, line 11:
```suggestion
    `ON`. It was
```

You don't need to tell users how the default was changed.  That is an implementation detail they should not care about.

RELEASE_NOTES
```diff
@@ -1,4 +1,35 @@
...
+    turned off by default because it breaks backward compatibility but it also
+    causes problems for new users and new configurations.
+
+    Users that do not want this new behavior can set
+    `-D Trilinos_MUST_FIND_ALL_TPL_LIBS=OFF`, which is backward compatible.
+
+  - Change the default for `Trilinos_USE_GNUINSTALLDIRS` from `OFF` to `ON`,
+    by setting `Trilinos_USE_GNUINSTALLDIRS_DEFAULT` to `ON`, in the goal to
```
bartlettroscoe, line 19:
```suggestion
    in the goal to
```

Again, you don't need to tell the user how the default was changed.  That is an implementation detail.

bartlettroscoe (changes requested):
Just a few small changes.  You can remove some of this detail, IMHO.

bartlettroscoe:
FYI: I thought that I had published this review last week, but I had forgotten to click the final button. Sorry for the delay.

RELEASE_NOTES
```diff
@@ -1,4 +1,35 @@
...
+# Trilinos Release 15.0 Release Notes                                         #
+#                                                                             #
+###############################################################################
+
+CMake
+
+  - Change the default for `Trilinos_MUST_FIND_ALL_TPL_LIBS` from `OFF` to
+    `ON`. It was
+    turned off by default because it breaks backward compatibility but it also
+    causes problems for new users and new configurations.
```
bartlettroscoe, line 13: nit: no longer well formatted but no big deal.

RELEASE_NOTES
```diff
@@ -1,4 +1,35 @@
...
+
+    Users that do not want this new behavior can set
+    `-D Trilinos_MUST_FIND_ALL_TPL_LIBS=OFF`, which is backward compatible.
+
+  - Change the default for `Trilinos_USE_GNUINSTALLDIRS` from `OFF` to `ON`,
+    in the goal to
+    move Trilinos and TriBITS to modern CMake.
```
bartlettroscoe, line 20: nit: no longer well formatted but no big deal.

bartlettroscoe (approved):
Looks correct.

nit: formatting could be better since this is a raw file.  In the future, this file should be changed to a *.md file so it will look good when formatted by GitHub and be easier to read.

PR #12269, opened 2023-09-15
Belos: create Tpetra version of GCRODR example

cwschilly:
@trilinos/belos 
@hkthorn 

## Motivation
This PR creates a Tpetra version of the `GCRODREpetraExFile.cpp`.

`PrecGCRODREpetraExFile.cpp` is omitted, because updating from an Ifpack preconditioner to Ifpack2 created a circular dependency with Belos.


packages/belos/tpetra/example/GCRODR/Dummy
hkthorn, line 1: Please remove this file.  This is an artifact from autotools and is no longer needed by Trilinos.  Removing these from the result of Anasazi is in PR #12242 .

hkthorn (changes requested):
Please remove Dummy file.  Otherwise LGTM!

packages/belos/tpetra/example/GCRODR/Dummy
cwschilly: I removed it; thank you!

PR #12271, opened 2023-09-15
Remove std::cout printing from Zoltan2 rebalance call

tvoskui:
@egboman 

## Motivation
When using Zoltan2 to do runtime mesh rebalancing, there are a couple statements printing to std::cout every time we rebalance, which we want to avoid when running Sierra. This changes those to print to the env_ using the same logging level as similar outputs in that call stack.


egboman (approved):
Looks good. Thanks!

lucbv:
Looking at this I would recommend hitting the merge button, @tvoskui let me know if you want me to do that in the morning?

tvoskui:
> Looking at this I would recommend hitting the merge button, @tvoskui let me know if you want me to do that in the morning?

Yes, please do. It says I'm not authorized to merge it myself.

PR #12272, opened 2023-09-15
Remove guard that disables stk-balance if Tpetra Global-Ordinal=int

alanw0:
I tested and confirmed that stk-balance builds and runs with Global-Ordinal = int.

Technically I didn't confirm that stk-balance is ok with Global-Ordinal = unsigned. I attempted to build that way, but Tpetra won't allow it. Tpetra has compile-time checks for signed-ness and won't compile.
There is no *known* reason why stk-balance
wouldn't work with unsigned, if it were allowed
through Zoltan2/Tpetra.

@trilinos/tpetra

### DIFF
--- a/cmake/RepositoryDependenciesSetup.cmake
+++ b/cmake/RepositoryDependenciesSetup.cmake
@@ -12,11 +12,6 @@
 #   endif()
 # endif() 
 
-#########################################################################
-# STKBalance does not work with GO=INT or GO=UNSIGNED
-# Note and disable it if it was set as a dependence of STK
-# Error out if it was explicitly requested by user
-
 SET(KDD_INT_INT OFF)     # Current default
 SET(KDD_INT_UNSIGNED OFF)  # Current default
 SET(KDD_INT_LONG OFF)      # Current default
@@ -45,28 +40,6 @@ ENDIF()
 
 IF(NOT ("${Tpetra_INST_INT_LONG_LONG}" STREQUAL ""))
   SET(KDD_INT_LONG_LONG ${Tpetra_INST_INT_LONG_LONG})
-ENDIF()
-
-IF ((NOT ${KDD_INT_LONG}) AND (NOT ${KDD_INT_LONG_LONG}))
-  IF ("${${PROJECT_NAME}_ENABLE_STKBalance}" STREQUAL "")
-    # STKBalance may be enabled but only implicitly (as a dependence of STK);
-    # give a message but turn off STKBalance support
-    MESSAGE("NOTE:  int global indices are enabled in Trilinos. "
-            "Because STKBalance requires long or long long "
-            "global indices, STKBalance will be disabled.  "
-            "To make this warning go away, do not request "
-            "int global indices in Trilinos (that is,  do not "
-            "set Tpetra_INST_INT_INT=ON or "
-            "Tpetra_INST_INT_UNSIGNED=ON)." )
-    SET(${PROJECT_NAME}_ENABLE_STKBalance OFF)
-  ELSEIF (${PROJECT_NAME}_ENABLE_STKBalance)
-    # STKBalance was explicitly enabled by the user, so error out
-    MESSAGE(FATAL_ERROR 
-            "STKBalance requires long or long long global indices, "
-            "but Trilinos is using int indices "
-            "(likely via Tpetra_INST_INT_INT or Tpetra_INST_INT_UNSIGNED).  "
-            "Disable STKBalance or specify Tpetra_INST_INT_LONG_LONG.")
-  ENDIF()
 ENDIF()
 
 # Tpetra supports only one GO type at a time, and the default is long long.


bartlettroscoe (commented):
Looks reasonable to me.  The only risk here I can see is that this change will allow the enable of the `STKBalance` subpackage in configurations where it might be broken that are not being tested by the Trilinos PR builds.

FYI: It looks like there may be some PR builds might be defined that use `GlobalOrdinal=int` as shown by:

https://github.com/trilinos/Trilinos/blob/d19fc8111569a69d2d5ef45569deb37baeb29345/packages/framework/ini-files/config-specs.ini#L2115

However, I don't see that build:

* `rhel7_sems-intel-19.0.5-mpich-3.2-serial_release-debug_static_no-kokkos-arch_no-asan_no-complex_fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables`

showing up anywhere on CDash as **not** shown [here](https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&date=2023-09-01&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=rhel7_sems-intel-19.0.5-mpich-3.2-serial_release-debug_static_no-kokkos-arch_no-asan_no-complex_fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables).

I looked at all of the attached `generatedPRFragment.cmake` files for all of the builds for this PR [here](https://trilinos-cdash.sandia.gov/index.php?project=Trilinos&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=63&value1=PR-12272-test&field2=buildstarttime&compare2=84&value2=NOW) and I can't find `Tpetra_INST_INT_INT` being set in any of them.  Also, I don't see any "intel" builds being run as part of nightly testing.

Therefore, I don't think this change is being tested as part of Trilinos automated testing.  How much of a risk this represents to customers.  I will defer to @rppawlo and @ccober6 and others on that question.

alanw0:
Thanks @bartlettroscoe . Here's a little background for this change. The Sierra/Plato folks wanted to build trilinos with int for global-ordinal, but that automatically disabled stk-balance, which prevented them from doing what they wanted to do. So I manually built and tested trilinos/stk-balance with int, and confirmed that it worked. Thus, this change to remove that restriction.
I understand Ross' concern if there isn't automatic testing of the int configuration.

bartlettroscoe:
> Thanks @bartlettroscoe . Here's a little background for this change. The Sierra/Plato folks wanted to build trilinos with int for global-ordinal, but that automatically disabled stk-balance, which prevented them from doing what they wanted to do. So I manually built and tested trilinos/stk-balance with int, and confirmed that it worked. Thus, this change to remove that restriction. I understand Ross' concern if there isn't automatic testing of the int configuration.

@alanw0, I guess if the customer and STK developers are okay without automated testing for this use case, then that is okay with me.

alanw0:
> >  I understand Ross' concern if there isn't automatic testing of the int configuration.
> 
> @alanw0, I guess if the customer and STK developers are okay without automated testing for this use case, then that is okay with me.

I wasn't clear. I should have said "I share Ross' concern...".

My thinking for this change was that if tpetra/zoltan2 can be configured/built with int, then stk-balance should be included in that instead of being automatically turned off.
I didn't know that int is an un-tested trilinos capability until your comment above. 
Sierra always configures with long long, so the int configuration is not tested in Sierra.

alanw0:
@rppawlo and @ccober6 what are the options for moving this forward?

As discussed by Ross above, it would be ideal if there were automated PR testing for the global-ordinal=int case.
If that's not feasible to activate, I would still like to make this change. If Zoltan2 and Tpetra are able to be configured with int, then stk-balance should be able to also.
I can add int testing in my own environment if needed.

bartlettroscoe:
> As discussed by Ross above, it would be ideal if there were automated PR testing for the global-ordinal=int case.

@alanw0, I did not mean to suggest this PR should not be merged in the absence of testing.  I was just pointing out that it would be difficult for Trilinos to support this feature if there were no automated builds covering this use case.

rppawlo:
I don't think there will be any issues with this. Will bring up the GO=int testing item at the next leaders meeting. My bet is that it is tested in nightly, but not in PR testing.

@alanw0 - a quick question - did all boost use get removed from stk? I know you were almost there when we last talked, but not sure if that last bit was removed.

alanw0:
@rppawlo stk no longer has a mandatory dependence on boost.
There is still an optional dependence, such that if boost is available, then some stack-trace functionality is available.

bartlettroscoe:
> I don't think there will be any issues with this. Will bring up the GO=int testing item at the next leaders meeting. My bet is that it is tested in nightly, but not in PR testing.

I could not find evidence of that on the CDash site.

ccober6:
What the current status on this?

alanw0:
> What the current status on this?

I would like to merge it. It needs an approval from someone, and appears to need re-test. I can set the re-test label.

rppawlo (approved):

Issue #12273, opened 2023-09-15
Tpetra: Host/Device Shared Memory Prep


csiefer2:
This issue is to track work prepping for host/device shared memory spaces, like Grace-Hopper and the AMD MI300 (as well as CUDA UVM).

PR #12274, opened 2023-09-15
Tpetra: Fix to WrappedDualView which should (in theory) help support GH/MI300

csiefer2:
Issue:#12273
Tests: Tested CUDA with and without UVM
Stakeholder feedback: n/a


jhux2 (approved):
lgtm, after discussion with @csiefer2

PR #12284, opened 2023-09-19
Add some constexpr traits to Tpetra Node types

brian-kelley:
- ``Node::is_serial``: is it Kokkos::Serial?
- ``Node::is_cpu``: is it a CPU-like device? (default memspace is HostSpace)
- ``Node::is_gpu``: is it a GPU-like device? (default memspace is something other than HostSpace)

@trilinos/tpetra 

## Motivation

Possible solution to issue discussed at Tpetra meeting. Instead of having to compare the node or execution space to every possible case (with ifdefs to check if types like ``Kokkos::Serial`` even exist), this lets users check the Node for common properties more easily.

@csiefer2 comments: Basically, I only want to see the words "HIP" "CUDA" and "SYCL" in Trilinos code where it is absolutely necessary for performance.  We have a lot of boilerplate where this stuff is littered through Trilinos right now, which makes adding new backends way harder then it needs to be.

### DIFF
--- a/packages/tpetra/core/compat/Tpetra_KokkosCompat_ClassicNodeAPI_Wrapper.hpp
+++ b/packages/tpetra/core/compat/Tpetra_KokkosCompat_ClassicNodeAPI_Wrapper.hpp
@@ -43,6 +43,18 @@ public:
   /// release.  This Node type is safe to use.
   static constexpr bool classic = false;
 
+  //! Whether the ExecutionSpace is Kokkos::Serial.
+#ifdef KOKKOS_ENABLE_SERIAL
+  static constexpr bool is_serial = std::is_same_v<ExecutionSpace, Kokkos::Serial>;
+#else
+  static constexpr bool is_serial = false;
+#endif
+
+  //! Whether the ExecutionSpace is CPU-like (its default memory space is HostSpace)
+  static constexpr bool is_cpu = std::is_same_v<typename ExecutionSpace::memory_space, Kokkos::HostSpace>;
+  //! Whether the ExecutionSpace is GPU-like (its default memory space is not HostSpace)
+  static constexpr bool is_gpu = !is_cpu;
+
   KokkosDeviceWrapperNode (Teuchos::ParameterList& /* params */) = delete;
   KokkosDeviceWrapperNode () = delete;
 


brian-kelley:
@csiefer2 @jhux2 @cwpearson This is probably the way I would deal with the issue @cgcgcg brought up. We can talk about it at next week's meeting

lucbv (approved):
Looks fine

cgcgcg:
Spots we can use this in MueLu:
https://github.com/trilinos/Trilinos/blob/master/packages/muelu/test/interface/ParameterListInterpreter.cpp#L114
https://github.com/trilinos/Trilinos/blob/master/packages/muelu/test/interface/CreateOperator.cpp#L307
https://github.com/trilinos/Trilinos/blob/master/packages/muelu/src/Operators/MueLu_RefMaxwell_def.hpp#L162
https://github.com/trilinos/Trilinos/blob/master/packages/muelu/src/MueCentral/MueLu_FactoryManager_decl.hpp#L135
https://github.com/trilinos/Trilinos/blob/master/packages/muelu/src/MueCentral/MueLu_FactoryManager_decl.hpp#L161
https://github.com/trilinos/Trilinos/blob/master/packages/muelu/src/Interface/MueLu_MLParameterListInterpreter_def.hpp#L209
https://github.com/trilinos/Trilinos/blob/master/packages/muelu/src/Interface/MueLu_ParameterListInterpreter_def.hpp#L270

lucbv:
@cgcgcg do you know why we use checks like:
```
typeid(Node).name() == typeid(Tpetra::KokkosCompat::KokkosHIPWrapperNode).name()
```
instead of the simpler
```
std::is_same_v<Node, Tpetra::KokkosCompat::KokkosHIPWrapperNode>
```

cgcgcg:
@lucbv Because someone wrote that once and then it got copy-pasta'd all over the place.

brian-kelley:
@csiefer2 The first two things you mention exist. There are 3 std::string names you can get from a node: node, exec and mem:
```
KokkosCudaWrapperNode::name() // returns "Cuda/Wrapper"
KokkosCudaWrapperNode::execution_space::name() // returns "Cuda"
KokkosCudaWrapperNode::memory_space::name() // returns "CudaSpace" or "CudaUVMSpace"
```

For the comm buffer memspace, we have ``Tpetra::Details::DefaultTypes::comm_buffer_memory_space<Node>``

masterleinad:
Related to https://github.com/kokkos/kokkos/issues/6006.

packages/tpetra/core/compat/Tpetra_KokkosCompat_ClassicNodeAPI_Wrapper.hpp
```diff
@@ -43,6 +43,18 @@ class KokkosDeviceWrapperNode {
...
 
+  //! Whether the ExecutionSpace is Kokkos::Serial.
+#ifdef KOKKOS_ENABLE_SERIAL
+  static constexpr bool is_serial = std::is_same_v<ExecutionSpace, Kokkos::Serial>;
+#else
+  static constexpr bool is_serial = false;
+#endif
+
+  //! Whether the ExecutionSpace is CPU-like (its default memory space is HostSpace)
+  static constexpr bool is_cpu = std::is_same_v<typename ExecutionSpace::memory_space, Kokkos::HostSpace>;
```
lucbv, line 54: Careful, Kokkos actually also defines `HBWSpace` for things like KNL and SapphireRapids which is should lead to `is_cpu == true`.

brian-kelley: Good catch, I'll fix this in a follow up PR

Issue #512, opened 2016-07-21
Teuchos TimeMonitor summarize prints zeros


srajama1:
Teuhos TimeMonitor summarize() call returns zeros for all the timers from Amesos2. However calling solver->printSolution in Amesos2 shows the actual times and number of calls for the Amesos2 timers.

The behavior was first exhibited in ROL. I have modified the a simple example in Amesos2 to demonstrate this behavior as well. Just run the example with --print-solution to see the problem.

The example code is in
packages/amesos2/example/SimpleSolve_File.cpp

CCing interested parties  @bartlettroscoe  @mhoemmen @dridzal

mhoemmen:
It's been nearly 4 years so I'm closing this issue.  Please feel free to reopen if it's still a problem.

PR #12287, opened 2023-09-20
Framework: Trilframe 512 add son nightly cuda build

cabejackson:
@trilinos/framework

## Motivation

The PRLaunchDriver is using a rhel8 system and when it runs `module load sems-archive-git/2.10.1` the jenkins job errors out with `ERROR: Unable to locate a modulefile for 'sems-archive-git/2.10.1'`. These changes fix that error.

Related to [Ticket #512](https://sems-atlassian-son.sandia.gov/jira/browse/TRILFRAME-512)

### DIFF
--- a/packages/framework/pr_tools/PullRequestLinuxDriver.sh
+++ b/packages/framework/pr_tools/PullRequestLinuxDriver.sh
@@ -8,6 +8,7 @@ source ${SCRIPTPATH:?}/common.bash
 on_weaver=$(echo "$@" | grep '\-\-on_weaver' &> /dev/null && echo "1")
 on_ats2=$(echo "$@" | grep '\-\-on_ats2' &> /dev/null && echo "1")
 on_kokkos_develop=$(echo "$@" | grep '\-\-kokkos\-develop' &> /dev/null && echo "1")
+on_rhel8=$(echo "$@" | grep '\-\-on_rhel8' &> /dev/null && echo "1")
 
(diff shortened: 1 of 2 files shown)


packages/framework/pr_tools/PullRequestLinuxDriver.sh
```diff
@@ -47,6 +48,14 @@ function bootstrap_modules() {
         module load python/3.7.3
         get_python_packages pip3
 
+        module list
+    elif [[ ${on_rhel8} == "1" ]]; then
+        source /projects/sems/modulefiles/utils/sems-modules-init.sh
```
sebrowne, line 53: I don't think this is needed.  It is likely run elsewhere.  If it were needed here, I think it would be needed at line 44 as well.

I run this manually before running genconfig.  The automated process handles this for you though.

sebrowne (changes requested):

packages/framework/pr_tools/PullRequestLinuxDriver.sh
fryeguy52: We do need the `source ...` here.  Weaver and vortex do not need this because `MODULEPATH` is set automatically on those systems such that the modules we need are available.  Some of the non-pipline jenkins jobs run this source in their build step

sebrowne: Oh I missed it at line 61 (I thought I was looking at RHEL7 but was looking at Weaver).  Yes, we need it, I was wrong.

cabejackson:
Opened [new PR here](https://github.com/trilinos/Trilinos/pull/12308)

PR #12301, opened 2023-09-21
teko: Guard ProbingPreconditionerFactory by TEKO_HAVE_EPETRA

Keno:
@trilinos/teko

## Motivation
The TEKO_HAVE_EPETRA switch controls whether ProbingPreconditionerFactory gets compiled, but not all includes of this header were appropriately guarded by the preprocessor macro. Correct this to fix the compilation of teko with TEKO_HAVE_EPETRA disabled.

## Testing
Build testing in the `!TEKO_HAVE_EPETRA` case was performed locally. This patch should not affect any runtime behavior.

## Additional Information
Found while attempting to enable `Teko` in the Julia distribution of Trilinos at https://github.com/JuliaPackaging/Yggdrasil/pull/7386.

### DIFF
--- a/packages/teko/src/Teko_PreconditionerFactory.cpp
+++ b/packages/teko/src/Teko_PreconditionerFactory.cpp
@@ -61,8 +61,8 @@
 #include "Teko_DiagonallyScaledPreconditionerFactory.hpp"
 #ifdef TEKO_HAVE_EPETRA
 #include "Teko_DiagonalPreconditionerFactory.hpp"
-#endif
 #include "Teko_ProbingPreconditionerFactory.hpp"
+#endif
 #include "Teko_IdentityPreconditionerFactory.hpp"
 #include "NS/Teko_LSCPreconditionerFactory.hpp"
 #include "NS/Teko_SIMPLEPreconditionerFactory.hpp"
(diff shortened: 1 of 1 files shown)


packages/teko/tests/unit_tests/tProbingFactory.cpp
```diff
@@ -68,7 +68,9 @@
 #include "Teko_DiagonallyScaledPreconditionerFactory.hpp"
 #include "Teko_PreconditionerInverseFactory.hpp"
 #include "Teko_PreconditionerLinearOp.hpp"
+#ifdef TEKO_HAVE_EPETRA
 #include "Teko_ProbingPreconditionerFactory.hpp"
+#endif
```
rppawlo, line 73: Changes to this file are probably not needed. This entire file/test is removed by cmake if epetra is disabled in teko.

rppawlo (dismissed):
Thanks for submitting this @Keno !

packages/teko/tests/unit_tests/tProbingFactory.cpp
Keno: Good call, I'll remove this part of the change. The other file was the issue, I was just grepping for other instances of the same pattern and didn't catch the build system guard.

rppawlo:
@sebrowne - This PR came back as failed (every build type failed), but when I follow the link to the cdash, it shows a green dashboard - everything passed. Can you help me figure out what is going on?

sebrowne:
Autotester is down (CEE GitLab is down and it needs that).

https://github.com/trilinos/Trilinos/issues/12305

PR #12302, opened 2023-09-21
ML: Don't use MPI_COMM_WORLD in MLAPI_Workspace

JacobDomagala:


jwillenbring:
@jhux2 These are fairly basic changes to bring Trilinos into compatibility with the xSDK Community Policies. Does this look reasonable enough to approve?

jhux2 (approved):
@jwillenbring This is probably fine, and easy enough to back out if it causes problems downstream.

PR #12308, opened 2023-09-21
Framework: Trilframe 512 add son nightly cuda build module load fix

cabejackson:
@trilinos/framework

## Motivation

The PRLaunchDriver is using a rhel8 system and when it runs `module load sems-archive-git/2.10.1` the jenkins job errors out with `ERROR: Unable to locate a modulefile for 'sems-archive-git/2.10.1'`. Then the python3 change is just because the jenkins job runs into an `unable to find /usr/bin/python3 file` error without it. These changes fix those errors.

Related to [JIRA Ticket #512](https://sems-atlassian-son.sandia.gov/jira/browse/TRILFRAME-512)


sebrowne (dismissed):

sebrowne (approved):

sebrowne (commented):
I spoke with Joe about this this morning, I think the real issue is that we're using an old version of Python3 (want to use 3.6 or newer).  Try using sems-python instead of sems-archive python

sebrowne (approved):

PR #12312, opened 2023-09-22
Galeri: add Tpetra tests

cwschilly:
@trilinos/galeri 
@jhux2 

## Motivation
This PR updates Galeri tests to use Tpetra instead of Epetra. Tests and table/notes written by @thearusable 

Tests found under the `test`, `example*` directories.

| Folder          | Epetra Tests (Main file name)          | Conversion Status                        |
|-----------------|-------------------------------|---------------------------------|
| test/VerySimple | cxx_main.cpp                            | **Partial**                     |
| example             | Map.cpp                                      | **Full**                         |
| example             | CrsMatrix.cpp                             | **Partial**                        |
| example             | VbrMatrix.cpp                            | **NO**                          |
| example             | LinearProblem.cpp                    | **Partial**                          |

- `VbrMatrix.cpp` - This test can't be converted because it's using a `Epetra::VbrMatrix` which doesn't exists in `Tpetra` package. It was present at some point in the past but it was deprecated and removed from the package as stated in the Release Notes: https://github.com/NexGenAnalytics/Trilinos/blob/NGA-FY23-develop/packages/tpetra/ReleaseNotes.txt#L438
- `cxx_main.cpp` / `LinearProblem.cpp` - Those two have a creation of necessary objects for calculations done, but there are no linear solvers in `Galeri`/`Tpetra` package which could be used to solve the problem. I added a comment that those solvers can be found for example in `Amesos2` or `Ifpack2`.
- `CrsMatrix.cpp` - Original tests used a helper method which generated and printed a 2D stencil of the matrix. I didn't found any good replacement for that, and I didn't create my own. 

In the tests there are methods used from `Galeri::Xpetra` namespace so they will require `Xpetra` to be enabled.

### DIFF
--- a/packages/galeri/example-tpetra/CMakeLists.txt
+++ b/packages/galeri/example-tpetra/CMakeLists.txt
@@ -1,0 +1,53 @@
+IF(${PACKAGE_NAME}_ENABLE_Tpetra AND ${PACKAGE_NAME}_ENABLE_Xpetra)
+  IF (Tpetra_INST_INT_INT)
+
+    TRIBITS_ADD_EXECUTABLE_AND_TEST(
+      Map_Tpetra
+      SOURCES Map_Tpetra.cpp
+      NUM_MPI_PROCS 1
+      COMM serial mpi
+      )
+
+    TRIBITS_ADD_EXECUTABLE_AND_TEST(
+      CrsMatrix_Tpetra
+      SOURCES CrsMatrix_Tpetra.cpp
+      NUM_MPI_PROCS 1
+      COMM serial mpi
+      )
+
+    TRIBITS_ADD_EXECUTABLE_AND_TEST(
+      LinearProblem_Tpetra
+      SOURCES LinearProblem_Tpetra.cpp
+      NUM_MPI_PROCS 1
+      COMM serial mpi
+      )
+
+  ENDIF()
+
+  IF (Tpetra_INST_INT_LONG_LONG)
+
+    TRIBITS_ADD_EXECUTABLE_AND_TEST(
+      Map_Tpetra_LL
+      SOURCES Map_Tpetra.cpp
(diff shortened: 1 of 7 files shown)


jwillenbring:
@jhux2 would you please review this PR to see if you feel it can be merged?

packages/galeri/example-tpetra/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
+IF(${PACKAGE_NAME}_ENABLE_Tpetra AND ${PACKAGE_NAME}_ENABLE_Xpetra)
```
cgcgcg, line 1: Why is Xpetra needed?

packages/galeri/example-tpetra/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+  IF (Tpetra_INST_INT_INT)
```
cgcgcg, line 2: Why is that a requirement?

packages/galeri/example-tpetra/CrsMatrix_Tpetra.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+  using Teuchos::RCP;
+  using Teuchos::rcp;
+
+  typedef Tpetra::Map<LO, GO, Node> Tpetra_Map;
+  typedef Tpetra::CrsMatrix<Scalar, LO, GO, Node> Tpetra_CrsMatrix;
+
+#ifdef HAVE_MPI
+  MPI_Init(&argc, &argv);
+#endif
```
cgcgcg, line 70: Isn't that handled by Teuchos?

jhux2 (approved):
These are only tests/examples.  Approving.

PR #12322, opened 2023-09-25
Belos: add `Tpetra` MINRES tests

tlamonthezie:
@trilinos/belos
@hkthorn
 
## Motivation
This PR provides the following Tpetra MINRES tests:

* [x]  `test_minres_diag.cpp`
* [x]  `test_minres_hb.cpp`
* [x]  `test_minres_indefinite.cpp`
* [x]  `test_pminres_hb.cpp`


hkthorn (changes requested):
Since we don't have a definitive path forward for preconditioned solver tests, please remove the test_pminres_hb test.  I see that it's using Belos, instead of Ifpack, as a preconditioner.  A ubiquitous approach for preconditioned testing needs to be added to all solvers, so this isn't needed until that approach is determined.  Thanks!

tlamonthezie:
> Since we don't have a definitive path forward for preconditioned solver tests, please remove the test_pminres_hb test. I see that it's using Belos, instead of Ifpack, as a preconditioner. A ubiquitous approach for preconditioned testing needs to be added to all solvers, so this isn't needed until that approach is determined. Thanks!

Ok @hkthorn test_pminres_hb has been removed.

hkthorn (approved):
LGTM

PR #12324, opened 2023-09-25
Stokhos (fixes #12279): construct flattened vector by copy constructing WDV

### LINKED ISSUE #12279: Stokhos: Amesos2 solvers are not working for GPU backend with more than one MPI rank
## Bug Report

@etphipp

### Description

`Amesos2` solvers (at least `KLU2` and `Basker`) do work for ensemble types with several ranks **only if there is no device**.

That is, for ensemble types with `Cuda` or `HIP` **and** with more than one rank, trying to solve fails. On a single rank, solving with ensemble type works.

The problem is due to `Tpetra::Details::WrappedDualView` counting (the infamous `Cannot access data on host while a device view is alive` problem).

For instance, the problem pops up in `KLU2` at the end of `solve_impl` when doing the `do_put` (see below provided `gdb` backtrace). However, I think it pops up there but the problem must be looked for in some upstream code. Indeed, for non-ensemble type, we only have the `Amesos2` stuff. For ensemble type, `Stokhos` uses the `create_flat_vector_view` (and related) to "flatten" the system but if I get it correctly, `Amesos2` is then called without any further manipulation of the data.

I think (but not sure) that when running with 2 ranks, the problem only happens on the non-root rank. That might be an hint.
Also, the fact that the solvers work in the same conditions for non-ensemble types is another hint.

If you want to have a reproducer, I suggest to have a look at @maartenarnst 's reproducer in #11578 (https://github.com/trilinos/Trilinos/issues/11578#issuecomment-1476705395), and simply change from a `double` scalar type to `Sacado::MP::Vector<Stokhos::StaticFixedStorage<int, double, STOKHOS_DEFAULT_ENSEMBLE_SIZE, execution_space>>`. You should see the test fails with `WrappedDualView` throwing at you :wink:

### Steps to Reproduce
1. SHA1: 608f54c97a294ddbbaeb69f992402a62ddfb654e
1. Shake fist angrily at computer (done :heavy_check_mark: )

### GDB backtrace

Here is the relevant part of my `gdb` output when running a test with a command like `mpiexec -np 2 gdb -ex run -ex bt <executable-path>`. This test works for `double` and `std::complex<double>` but fails as mentioned above for ensemble types.

```bash
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[ RUN      ] TestFixtureWithExecutionSpace_execution_space_Timoshenko1DTest.TestAgainstExactSol
  what():  Tpetra::Details::WrappedDualView (name = MV::DualView; host use_count = 2; device use_count = 4): Cannot access data on host while a device view is alive

Thread 1 "exec_Timoshenko" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=140734698422272) at ./nptl/pthread_kill.c:44
44      ./nptl/pthread_kill.c: No such file or directory.
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140734698422272) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140734698422272) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140734698422272, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007fffd72b8476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007fffd729e7f3 in __GI_abort () at ./stdlib/abort.c:79
--Type <RET> for more, q to quit, c to continue without paging--#5  0x00007fffd7627b9e in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007fffd763320c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007fffd7633277 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007fffd76334d8 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#9  0x00007fffddf14ca1 in Tpetra::Details::WrappedDualView<Kokkos::DualView<double**, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, void> >::throwIfDeviceViewAlive (this=0x5555578f3880) at /home/costmo-user/Trilinos/packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp:684
#10 0x00007fffddeff54f in Tpetra::Details::WrappedDualView<Kokkos::DualView<double**, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaSpace>, void> >::getHostView (this=0x5555578f3880) at /home/costmo-user/Trilinos/packages/tpetra/core/src/Tpetra_Details_WrappedDualView.hpp:263
#11 0x00007fffddeddd5f in Tpetra::MultiVector<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >::getLocalViewHost (
    this=0x5555578f3720, s=...) at /home/costmo-user/Trilinos/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp:3980
#12 0x00007fffddef070a in Tpetra::MultiVector<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >::unpackAndCombine (
    this=0x5555578f3720, importLIDs=..., imports=..., constantNumPackets=1, CM=Tpetra::REPLACE, space=...)
    at /home/costmo-user/Trilinos/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp:2057
#13 0x00007fffdf5eb21e in Tpetra::DistObject<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >::doUnpackAndCombine
    (this=0x5555578f3720, remoteLIDs=..., constantNumPackets=1, CM=Tpetra::REPLACE, space=...)
    at /home/costmo-user/Trilinos/packages/tpetra/core/src/Tpetra_DistObject_def.hpp:1664
#14 0x00007fffdf5e8a31 in Tpetra::DistObject<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >::endTransfer (
    this=0x5555578f3720, src=..., transfer=..., modeString=0x7fffffffb5b0 "endImport (forward mode)", 
    revOp=Tpetra::DistObject<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >::DoForward, CM=Tpetra::REPLACE, 
    restrictedMode=false) at /home/costmo-user/Trilinos/packages/tpetra/core/src/Tpetra_DistObject_def.hpp:1339
#15 0x00007fffdf5e199d in Tpetra::DistObject<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >::endImport (
    this=0x5555578f3720, source=..., importer=..., CM=Tpetra::REPLACE, restrictedMode=false)
 at /home/costmo-user/Trilinos/packages/tpetra/core/src/Tpetra_DistObject_def.h--Type <RET> for more, q to quit, c to continue without paging--   pp:549
#16 0x00007fffdf5dff34 in Tpetra::DistObject<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >::doImport (
    this=0x5555578f3720, source=..., importer=..., CM=Tpetra::REPLACE, restrictedMode=false)
    at /home/costmo-user/Trilinos/packages/tpetra/core/src/Tpetra_DistObject_def.hpp:298
#17 0x00007fffbb3e2c35 in Amesos2::MultiVecAdapter<Tpetra::MultiVector<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> > >::put1dData_kokkos_view<Kokkos::View<double**, Kokkos::LayoutLeft, Kokkos::OpenMP> > (this=0x55555b6c55d0, kokkos_new_data=..., lda=0, source_map=..., 
    distribution=Amesos2::ROOTED) at /home/costmo-user/Trilinos/packages/amesos2/src/Amesos2_TpetraMultiVecAdapter_def.hpp:588
#18 0x00007fffbb3c9699 in Amesos2::Util::put_1d_data_helper_kokkos_view<Amesos2::MultiVecAdapter<Tpetra::MultiVector<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> > >, Kokkos::View<double**, Kokkos::LayoutLeft, Kokkos::OpenMP> >::do_put (mv=..., kokkos_data=..., ldx=0, 
    distribution_map=..., distribution=Amesos2::ROOTED) at /home/costmo-user/Trilinos/packages/amesos2/src/Amesos2_MultiVecAdapter_def.hpp:373
#19 0x00007fffbb3a6e3b in Amesos2::Util::put_1d_data_helper_kokkos_view<Amesos2::MultiVecAdapter<Tpetra::MultiVector<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> > >, Kokkos::View<double**, Kokkos::LayoutLeft, Kokkos::OpenMP> >::do_put (mv=..., kokkos_data=..., ldx=0, 
    distribution=Amesos2::ROOTED, indexBase=0) at /home/costmo-user/Trilinos/packages/amesos2/src/Amesos2_MultiVecAdapter_def.hpp:394
#20 0x00007fffbb37532e in Amesos2::KLU2<Tpetra::CrsMatrix<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >, Tpetra::MultiVector<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> > >::solve_impl (this=0x55555b6dcc90, X=..., B=...)
    at /home/costmo-user/Trilinos/packages/amesos2/src/Amesos2_KLU2_def.hpp:383
#21 0x00007fffbb525bf1 in Amesos2::SolverCore<Amesos2::KLU2, Tpetra::CrsMatrix<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >, Tpetra::MultiVector<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> > >::solve (this=0x55555b6dcc90, 
    X=..., B=...) at /home/costmo-user/Trilinos/packages/amesos2/src/Amesos2_SolverCore_def.hpp:226
#22 0x00007fffbb5259dc in Amesos2::SolverCore<Amesos2::KLU2, Tpetra::CrsMatrix<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> >, Tpetra::MultiVector<double, int, long long, Tpetra::KokkosCompat::KokkosDeviceWrapperNode<Kokkos::Cuda, Kokkos::CudaSpace> > >::solve (this=0x55555b6dcc90)
    at /home/costmo-user/Trilinos/packages/amesos2/src/Amesos2_SolverCore_def.hpp:175
```

@maartenarnst

maartenarnst:
Hi @etphipp,

In the mean time, @romintomasetti and I have looked more into the problem with running `Amesos2` solvers with ensembles on GPU with more than one MPI rank (issue #12279).

This PR is a proposed fix.

### Reproducing the problem using existing Stokhos tests

Running `Amesos2` solvers with ensembles is tested in `stokhos/test/UnitTest/Stokhos_TpetraCrsMatrixMPVectorUnitTest.hpp`. 

The existing pipeline typically runs fine, and issue #12279 doesn't show up. It seems the reason is that this test selects only one `Amesos2` solver, 
depending on which ones the user enabled and an order of preference. When the user activated `basker`, that's the one that `Stokhos` chooses.
For `basker`, the code goes into the specialisation of `basker` for ensembles -> no problem.

However, when `Amesos2` is compiled without `basker` but with e.g. `klu2`, then `Stokhos` doesn't provide a specialisation for ensembles, and it creates
a "flattened" system that it provides to `Amesos2`. In such a situation, the Stokhos test throws runtime errors when running on GPU with more than one MPI rank.

### The culprit   

We think the culprit is how `Stokhos` creates flattened vectors from vectors of ensembles. 

This is done in the function `create_flat_vector_view` in `stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_Utilities_MP_Vector.hpp`:

- https://github.com/trilinos/Trilinos/blob/fd93b55195def346c69c806bb2b0aab7b6070fc9/packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_Utilities_MP_Vector.hpp#L467-L471

Currently, this function works by extracting from the vector of ensembles a device view of the local data. This device view is flattened by using a "reshaping" conversion assignment operator.
Then the flattened vector is created by using the `Tpetra::MultiVector` constructor that takes the (flattened) device view of the local data and the (flattened) map.

We think there are two problems:

- We end up with a flattened vector for which, from the outset, the `use_count` for the device view is higher than the `use_count` for the host view. Indeed, while the flattened vector and the original
vector of ensembles share the same data allocation on device, the used `Tpetra::MultiVector` constructor creates a new data allocation on host. Due to the logic of the underlying wrapped dual view,
this difference in the use counts in turn implies that we can't do `getLocalViewHost` on the flattened vector later in the code.

- The flattened vector and the original vector of ensembles don't share the same dual view, hence they don't share the `modified_flags` and don't know about each other's sync state.

When running an `Amesos2` solver such as `klu2` with ensembles on GPU with more than one MPI rank, the code wants to gather the entire system on the root rank, solve it there,
and then redistribute the pieces of the solution again to the ranks. Some of these steps want to use `getLocalViewHost`, and some of these steps rely on the syncing mechanism of the underlying WDV.
Hence, the issues that we see in #12279.

### Proposed fix

The proposed fix is to create the flattened vector by using a "reshaping" conversion assignment operator/copy constructor for the entire underlying wrapped dual view. This way,
the flattened vector and the vector of ensembles share the same data allocations on host and device, share the use counts, and share the `modified_flags` (sync state).

The `Tpetra::MultiVector` class does not currently have a getter for the underlying WDV (it's a protected member), and the WDV does not currently have templated conversion 
assignment operator/copy constructor. So this PR adds those. Especially adding a getter for the underlying WDV could be controversial :) So we'll likely have to check 
with the `tpetra` team to get their input. But we do think we have a convincing use case here.

Note that the `Kokkos_DualView.hpp` is missing the templated assignment operator, but it has the templated copy constructor.

On the side, this PR also modifies a bit the logic in `stokhos/test/UnitTest/Stokhos_TpetraCrsMatrixMPVectorUnitTest.hpp` to ensure that all enabled Amesos2 solvers
are tested. And this PR removes some (seemingly) unused code from `stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_Utilities_MP_Vector.hpp`. In particular,
it seems the versions of `create_flat_vector_view` templated on `Node` can be expected to be unused because there are "more specialised" ones templated on `Device`. 

What do you think? Thanks in advance!

### DIFF
--- a/packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_Utilities_MP_Vector.hpp
+++ b/packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_Utilities_MP_Vector.hpp
@@ -241,268 +241,64 @@ namespace Stokhos {
     return flat_graph;
   }
 
-
   // Create a flattened vector by unrolling the MP::Vector scalar type.  The
   // returned vector is a view of the original
-  template <typename Storage, typename LocalOrdinal, typename GlobalOrdinal,
-            typename Node>
-  Teuchos::RCP< const Tpetra::MultiVector<typename Storage::value_type,
-                                          LocalOrdinal,GlobalOrdinal,Node> >
+  template <typename Storage, typename LocalOrdinal, typename GlobalOrdinal, typename Node>
+  Teuchos::RCP<Tpetra::MultiVector<typename Storage::value_type, LocalOrdinal, GlobalOrdinal, Node>>
   create_flat_vector_view(
-    const Tpetra::MultiVector<Sacado::MP::Vector<Storage>,
-                              LocalOrdinal,GlobalOrdinal,Node>& vec_const,
-    const Teuchos::RCP< const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> >& flat_map) {
-    using Teuchos::ArrayRCP;
-    using Teuchos::RCP;
-    using Teuchos::rcp;
+    const Tpetra::MultiVector<Sacado::MP::Vector<Storage>, LocalOrdinal, GlobalOrdinal, Node>& vec,
+    const Teuchos::RCP<const Tpetra::Map<LocalOrdinal, GlobalOrdinal, Node>>& flat_map
+  ) {
+    using BaseScalar = typename Storage::value_type;
+    using FlatVector = Tpetra::MultiVector<BaseScalar, LocalOrdinal, GlobalOrdinal, Node>;
 
-    typedef Sacado::MP::Vector<Storage> Scalar;
-    typedef typename Storage::value_type BaseScalar;
-    typedef Tpetra::MultiVector<Scalar,LocalOrdinal,GlobalOrdinal,Node> Vector;
-    typedef Tpetra::MultiVector<BaseScalar,LocalOrdinal,GlobalOrdinal,Node> FlatVector;
-
-    // MP size
-    const LocalOrdinal mp_size = Storage::static_size;
-
-    // Cast-away const since resulting vector is a view and we can't create
-    // a const-vector directly
-    Vector& vec = const_cast<Vector&>(vec_const);
-
-    // Get data
-    ArrayRCP<Scalar> vec_vals = vec.get1dViewNonConst();
-    const size_t vec_size = vec_vals.size();
-
-    // Create view of data
-    BaseScalar *flat_vec_ptr =
-      reinterpret_cast<BaseScalar*>(vec_vals.getRawPtr());
-    ArrayRCP<BaseScalar> flat_vec_vals =
-      Teuchos::arcp(flat_vec_ptr, 0, vec_size*mp_size, false);
+    // Create flattenend view using reshaping conversion copy constructor
+    typename FlatVector::wrapped_dual_view_type flat_vals(vec.getWrappedDualView());
 
     // Create flat vector
-    const size_t stride = vec.getStride();
-    const size_t flat_stride = stride * mp_size;
-    const size_t num_vecs = vec.getNumVectors();
-    RCP<FlatVector> flat_vec =
-      rcp(new FlatVector(flat_map, flat_vec_vals, flat_stride, num_vecs));
-
-      return flat_vec;
+    return Teuchos::make_rcp<FlatVector>(flat_map, flat_vals);
   }
 
-  // Create a flattened vector by unrolling the MP::Vector scalar type.  The
-  // returned vector is a view of the original.  This version creates the
-  // map if necessary
-  template <typename Storage, typename LocalOrdinal, typename GlobalOrdinal,
-            typename Node>
-  Teuchos::RCP< const Tpetra::MultiVector<typename Storage::value_type,
-                                          LocalOrdinal,GlobalOrdinal,Node> >
+  // This version creates the map if necessary
+  template <typename Storage, typename LocalOrdinal, typename GlobalOrdinal, typename Node>
+  Teuchos::RCP<Tpetra::MultiVector<typename Storage::value_type, LocalOrdinal, GlobalOrdinal, Node>>
   create_flat_vector_view(
-    const Tpetra::MultiVector<Sacado::MP::Vector<Storage>,
-                              LocalOrdinal,GlobalOrdinal,Node>& vec,
-    Teuchos::RCP< const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> >& flat_map) {
+    const Tpetra::MultiVector<Sacado::MP::Vector<Storage>, LocalOrdinal, GlobalOrdinal, Node>& vec,
+          Teuchos::RCP<const Tpetra::Map<LocalOrdinal, GlobalOrdinal, Node>>& flat_map
+  ) {
     if (flat_map == Teuchos::null) {
       const LocalOrdinal mp_size = Storage::static_size;
       flat_map = create_flat_map(*(vec.getMap()), mp_size);
     }
-    const Teuchos::RCP< const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> > const_flat_map = flat_map;
+    const Teuchos::RCP<const Tpetra::Map<LocalOrdinal, GlobalOrdinal, Node>> const_flat_map = flat_map;
     return create_flat_vector_view(vec, const_flat_map);
   }
 
-  // Create a flattened vector by unrolling the MP::Vector scalar type.  The
-  // returned vector is a view of the original
-  template <typename Storage, typename LocalOrdinal, typename GlobalOrdinal,
-            typename Node>
-  Teuchos::RCP< Tpetra::MultiVector<typename Storage::value_type,
-                                    LocalOrdinal,GlobalOrdinal,Node> >
+  template <typename Storage, typename LocalOrdinal, typename GlobalOrdinal, typename Node>
+  Teuchos::RCP<Tpetra::Vector<typename Storage::value_type, LocalOrdinal, GlobalOrdinal, Node>>
   create_flat_vector_view(
-    Tpetra::MultiVector<Sacado::MP::Vector<Storage>,
-                        LocalOrdinal,GlobalOrdinal,Node>& vec,
-    const Teuchos::RCP< const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> >& flat_map) {
-    using Teuchos::ArrayRCP;
-    using Teuchos::RCP;
-    using Teuchos::rcp;
-
-    typedef Sacado::MP::Vector<Storage> Scalar;
-    typedef typename Storage::value_type BaseScalar;
-    typedef Tpetra::MultiVector<BaseScalar,LocalOrdinal,GlobalOrdinal,Node> FlatVector;
-
-    // MP size
-    const LocalOrdinal mp_size = Storage::static_size;
-
-    // Get data
-    ArrayRCP<Scalar> vec_vals = vec.get1dViewNonConst();
-    const size_t vec_size = vec_vals.size();
-
-    // Create view of data
-    BaseScalar *flat_vec_ptr =
-      reinterpret_cast<BaseScalar*>(vec_vals.getRawPtr());
-    ArrayRCP<BaseScalar> flat_vec_vals =
-      Teuchos::arcp(flat_vec_ptr, 0, vec_size*mp_size, false);
-
-    // Create flat vector
-    const size_t stride = vec.getStride();
-    const size_t flat_stride = stride * mp_size;
-    const size_t num_vecs = vec.getNumVectors();
-    RCP<FlatVector> flat_vec =
-      rcp(new FlatVector(flat_map, flat_vec_vals, flat_stride, num_vecs));
-
-      return flat_vec;
+    const Tpetra::Vector<Sacado::MP::Vector<Storage>, LocalOrdinal, GlobalOrdinal, Node>& vec,
+    const Teuchos::RCP<const Tpetra::Map<LocalOrdinal, GlobalOrdinal, Node>>& flat_map
+  ) {
+    return create_flat_vector_view(
+      dynamic_cast<const Tpetra::MultiVector<Sacado::MP::Vector<Storage>, LocalOrdinal, GlobalOrdinal, Node>&>(vec),
+      flat_map
+    )->getVectorNonConst(0);
   }
 
-  // Create a flattened vector by unrolling the MP::Vector scalar type.  The
-  // returned vector is a view of the original.  This version creates the
-  // map if necessary
-  template <typename Storage, typename LocalOrdinal, typename GlobalOrdinal,
-            typename Node>
-  Teuchos::RCP< Tpetra::MultiVector<typename Storage::value_type,
-                                    LocalOrdinal,GlobalOrdinal,Node> >
+  template <typename Storage, typename LocalOrdinal, typename GlobalOrdinal, typename Node>
+  Teuchos::RCP<Tpetra::Vector<typename Storage::value_type, LocalOrdinal, GlobalOrdinal, Node>>
   create_flat_vector_view(
-    Tpetra::MultiVector<Sacado::MP::Vector<Storage>,
-                        LocalOrdinal,GlobalOrdinal,Node>& vec,
-    Teuchos::RCP< const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> >& flat_map) {
+    const Tpetra::Vector<Sacado::MP::Vector<Storage>, LocalOrdinal, GlobalOrdinal, Node>& vec,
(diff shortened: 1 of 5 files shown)


etphipp:
Thanks @maartenarnst for the PR and great explanation.  This seems reasonable to me.  I think another solution that wouldn't require changes to Tpetra would be to sync the newly created flattened vector to the host after construction, but that would create a new host-side allocation, as you described.  So I think this solution is better in the long run.  Can someone from @trilinos/tpetra take a look?

etphipp (dismissed):

maartenarnst:
Thanks, @etphipp ! Hi @brian-kelley or @csiefer2, would you have a moment to take a look at the tpetra changes? Thanks, in advance!

etphipp:
@trilinos/framework The autotester is failing tests called "TrilinosInstallTests_doInstall" and "TrilinosInstallTests_simpleBuildAgainstTrilinos_by_package_build_tree" with errors like:
```
CMake Error at /scratch/trilinos/jenkins/ascic164/workspace/Trilinos_PR_gcc-8.3.0/pull_request_test/packages/tpetra/core/src/cmake_install.cmake:46 (file):
  file INSTALL cannot find
  "/scratch/trilinos/jenkins/ascic164/workspace/Trilinos_PR_gcc-8.3.0/pull_request_test/packages/tpetra/core/src/libtpetra.a":
  No such file or directory.
  /scratch/trilinos/jenkins/ascic164/workspace/Trilinos_PR_gcc-8.3.0/pull_request_test/packages/tpetra/core/cmake_install.cmake:52 (include)
  /scratch/trilinos/jenkins/ascic164/workspace/Trilinos_PR_gcc-8.3.0/pull_request_test/packages/tpetra/cmake_install.cmake:52 (include)
  /scratch/trilinos/jenkins/ascic164/workspace/Trilinos_PR_gcc-8.3.0/pull_request_test/cmake_install.cmake:223 (include)
```
and
```
[ 50%] Building CXX object CMakeFiles/MyApp.dir/app.cpp.o
gmake[2]: *** No rule to make target `/scratch/trilinos/jenkins/ascic164/workspace/Trilinos_PR_gcc-8.3.0/pull_request_test/packages/tpetra/core/ext/libtpetraext.a', needed by `MyApp'.  Stop.
gmake[1]: *** [CMakeFiles/MyApp.dir/all] Error 2
gmake: *** [all] Error 2
```
Do you know what is going on here?

packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_Utilities_MP_Vector.hpp
```diff
@@ -241,269 +241,65 @@ namespace Stokhos {
...
-                   LocalOrdinal,GlobalOrdinal,Node>& vec,
-    const Teuchos::RCP< const Tpetra::Map<LocalOrdinal,GlobalOrdinal,Node> >& flat_map) {
-    Tpetra::MultiVector<Sacado::MP::Vector<Storage>,LocalOrdinal,GlobalOrdinal,Node>& mv = vec;
-    Teuchos::RCP< Tpetra::MultiVector<typename Storage::value_type,LocalOrdinal,GlobalOrdinal,Node> > flat_mv = create_flat_vector_view(mv, flat_map);
-    return flat_mv->getVectorNonConst(0);
+    const Tpetra::Vector<Sacado::MP::Vector<Storage>, LocalOrdinal, GlobalOrdinal, Node>& vec,
+    const Teuchos::RCP<const Tpetra::Map<LocalOrdinal, GlobalOrdinal, Node>>& flat_map
+  ) {
+    return create_flat_vector_view(
+      dynamic_cast<const Tpetra::MultiVector<Sacado::MP::Vector<Storage>, LocalOrdinal, GlobalOrdinal, Node>&>(vec),
```
brian-kelley, line 284: This could be a ``static_cast`` since Vector is a subclass of MultiVector

brian-kelley (dismissed):

packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_Utilities_MP_Vector.hpp
brian-kelley: But this is a very minor thing, if you're not making any other changes for the autotester then don't worry about it

brian-kelley:
@maartenarnst I agree that the changes make sense, and thanks for the detailed rationale!

sebrowne:
My guess is that they're caused by the build errors in Tpetra.  If Tpetra can't be built, then it can't be installed, which is what those tests do.  This is confirmed (at least, the intent of the author is confirmed) with the higher-up message:

`
TEST_0

Install enabled and built Trilinos packages (NOTE: This test will fail if the project has **any** build errors!)
`

etphipp:
@sebrowne, you are right, there are build errors.  Thanks for pointing this out.

@maartenarnst this is the error, can you fix it?
```
In file included from packages/stokhos/src/Tpetra_MultiVector_MP_Vector_16_Serial.cpp:48:
../Trilinos/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp:4020:4: error: extra ‘;’ [-Werror=pedantic]
   };
    ^
[CTest: warning matched] cc1plus: all warnings being treated as errors
```

packages/stokhos/src/sacado/kokkos/vector/tpetra/Stokhos_Tpetra_Utilities_MP_Vector.hpp
maartenarnst: Thanks, @brian-kelley ! I've just changed it into a `static_cast`.

maartenarnst:
Thanks, @etphipp for sending the error message! I've just removed the semi-colon. Would you have a moment to set it up for a retest?

brian-kelley (approved):

etphipp (approved):

Issue #10081, opened 2022-01-13
Set LINK_LIBRARIES_ONLY_TARGETS=TRUE by default and verify Trilinos is clean with CMake 3.23+


bartlettroscoe:
As discussed in https://github.com/TriBITSPub/TriBITS/issues/299#issuecomment-1012312724, CMake 3.23 will learn the option `LINK_LIBRARIES_ONLY_TARGETS` that if set to `TRUE`, will help catch errors when linking against undefined targets.  With the updated TriBITS, raw library names for external packages/TPLs are no longer directly passed to `target_link_libraries()` and therefore, unless people are directly calling `target_link_libraries()` against raw lib names in Trilinos CMakeLists.txt files, we should be able to set `LINK_LIBRARIES_ONLY_TARGETS` to true with Trilinos.

But before I felt confident about that, Trilinos would need to upgrade its minimum CMake version to CMake 3.23+ and all of the Trilinos PR and ATDM Trilinos builds would need to be tested with this option enabled with the upgraded CMake.

bartlettroscoe:
@sebrowne, now that Trilinos is requiring CMake 3.23+, we can pull the trigger on this one.  This will make the CMakeLists.txt files more robust and will be critical when refactoring to use raw library linking as listed in:

* https://github.com/TriBITSPub/TriBITS/issues/569

sebrowne:
So to clarify, this will turn "nonexistant" targets called with target_link_libaries(myTarget <TYPE> someBadTarget) into configure errors instead of doing "lld: error: -lsomeBadTarget not found" at link time?

bartlettroscoe:
> So to clarify, this will turn "nonexistant" targets called with `target_link_libaries(myTarget <TYPE> someBadTarget)` into configure errors instead of doing `lld: error: -lsomeBadTarget not found` at link time?

@sebrowne, yes.  From [LINK_LIBRARIES_ONLY_TARGETS](https://cmake.org/cmake/help/v3.23/prop_tgt/LINK_LIBRARIES_ONLY_TARGETS.html?highlight=link_libraries_only_targets):

> Enforce that link items that can be target names are actually existing targets.

Otherwise, if someone makes a mistake, you get a build-time error about a missing dependency (which is confusing for some people to figure out).  That does not happen much with current TriBITS due to auto-linking between dependent package libraries but if Trilinos moves to manual linking (see https://github.com/TriBITSPub/TriBITS/issues/569) then this will happen all of the time.

[...]

PR #12339, opened 2023-09-28
add SET(LINK_LIBRARIES_ONLY_TARGETS TRUE) to CMakeLists.txt

fryeguy52:
@trilinos/framework 

## Motivation
Addressees #10081 by adding `SET(LINK_LIBRARIES_ONLY_TARGETS TRUE)` to `CMakeLists.txt`

## Related Issues
#10081

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,6 +109,8 @@ include(SetupTribitsInstall)
 
 SET(Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT TRUE)
 
+SET(LINK_LIBRARIES_ONLY_TARGETS TRUE)
+
 IF(${PROJECT_NAME}_ENABLE_YouCompleteMe)
   INCLUDE(CodeCompletion)
 ENDIF()


sebrowne:
Add an entry detailing what we did to the RELEASE_NOTES, and link to the CMake documentation please.

CMakeLists.txt
```diff
@@ -109,6 +109,8 @@ include(SetupTribitsInstall)
 
 SET(Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT TRUE)
 
+SET(LINK_LIBRARIES_ONLY_TARGETS TRUE)
```
bartlettroscoe, line 112: I think setting this as a non-cache var is appropriate.  It may impact projects like Drekar that use a base Trilinos CMake project but it should be easy to for them to fix any.  I am curious to see if all of Trilinos actually configures with this turned on (since raw library names will no longer be accepted).

This should **not** cause any change in behavior for downstream Trilinos customers just pulling in Trilinos with `find_package(Trilinos ...)` or `find_package(<TrilinosPackage> ...)`.  Therefore, I am not sure it requires a release note.

sebrowne: Ah good point.  This affects _developers_, not customers.  and with how TriBITS automates things it may not even affect developers all that much.

bartlettroscoe (changes requested):
From reading:

* https://cmake.org/cmake/help/latest/prop_tgt/LINK_LIBRARIES_ONLY_TARGETS.html

`LINK_LIBRARIES_ONLY_TARGETS` is a target property, not a variable.  I think you need to set:

* https://cmake.org/cmake/help/latest/variable/CMAKE_LINK_LIBRARIES_ONLY_TARGETS.html#variable:CMAKE_LINK_LIBRARIES_ONLY_TARGETS

Please make sure and do a local build and test that this is catching cases like described in:

* https://cmake.org/cmake/help/latest/prop_tgt/LINK_LIBRARIES_ONLY_TARGETS.html

bartlettroscoe:
BTW, when I saw that the PR builds passed [above](https://github.com/trilinos/Trilinos/pull/12339#issuecomment-1740348405), I knew this option was not getting turned on by this PR because I know there are violations of this in Trilinos.  For example:

https://github.com/trilinos/Trilinos/blob/00cacb842123733882fd9078a73fa20a373b9931/packages/zoltan/src/CMakeLists.txt#L671-L683

That passing of of a raw `m` should trigger the error described in:

* https://cmake.org/cmake/help/latest/prop_tgt/LINK_LIBRARIES_ONLY_TARGETS.html

and that shows how to fix this.

CMakeLists.txt
fryeguy52, line 1: @sebrowne @bartlettroscoe setting this in the top level `CMakeLists.txt` does not result in this being set in the cache.  Do you have insight as to why this is? Where is the right place to make this change?

bartlettroscoe: Right, but the only reason to make it a cache variable is if you want to allow the user to change it.  If you don't want to allow the user to change the value of this variable, then using a project-level non-cache variable is the way to go.  (And individual packages can turn this off with a directory-scoped variable if they want.)

The relationship to cache and non-cache variables is a fundamental aspect of CMake (and also confuses many people because it is so different that other programming languages).

The way to know if this is doing what you think is to make sure that a violation is caught.

sebrowne:
@fryeguy52 the reason this appeared to not work was because the variable was being set after the tribits call, which means all of the libraries had already been created/configured.  Moving it up in the file fixes it.  I actually noticed another variable that was set too low as well, and fixed that as part of #12362.

Closing in favor of #12362 (Joe handed this off to me this morning).

CMakeLists.txt
sebrowne: Agreed.  I figured out why it wasn't working (see comment at the end and new PR).

PR #12345, opened 2023-09-29
Anasazi: create GeneralizedDavidson Tpetra tests

cwschilly:
@trilinos/anasazi 
@hkthorn 

## Motivation

This PR creates the following tests in the `tpetra/test/GeneralizedDavidson` directory:

- `cxx_main_nh.cpp`
- `cxx_main.cpp`
- `cxx_main_lap.cpp`
---

`cxx_main_nh.cpp` is modeled after the current test in the Epetra test directory (basically a one-to-one conversion, changing E's to T's)

`cxx_main.cpp` and `cxx_main_lap.cpp` are modeled after the tests of other solvers in the Tpetra test directory. Since the GeneralizedDavidson solver does not support complex types, `cxx_main.cpp` uses the same logic as the `cxx_main_complex.cpp` tests for the other solvers, but uses `ST=double` instead.


packages/anasazi/tpetra/test/GeneralizedDavidson/cxx_main.cpp
hkthorn, line 1: It is my understanding that the mhd1280b.cua file contains a complex-valued matrix stored as interleaved doubles, so it's not useful for this test.  Since the Generalized Davidson solver only works for real scalar types, this test should replicate the anasazi/epetra/test/GeneralizedDavidson/cxx_main.cpp test instead of the other cxx_main.cpp tests in the Tpetra directory.

hkthorn (changes requested):
Fix cxx_main.cpp test to replicate anasazi/epetra/test/GeneralizedDavidson/cxx_main.cpp test.

jwillenbring:
@hkthorn So you are saying that this test should be modeled after a different existing test?

PR #12347, opened 2023-09-29
MueLu_Amesos2Smoother: Fixes for projection and non-contiguous maps

cgcgcg:
@trilinos/muelu 

## Motivation
- Projection was incorrect for nullspaces of dim>1.
- Check for non-contiguous map was insufficient. 
  (I saw this for a small problem where MueLu only set up a direct solver, and the input matrix had a non-contiguous   rowmap.)


jhux2 (approved):
lgtm

GrahamBenHarper (approved):
lgtm

PR #12355, opened 2023-10-03
Sacado: Fix clang -Wdtor-name warning

maartenarnst:
@trilinos/sacado

@etphipp

## Motivation

When compiling with `clang`, we get the warning:

```
/opt/Trilinos/.../include/Sacado_Fad_VectorImp.hpp:88:66: error: ISO C++ requires the name after '::~' to be found in the same scope as the name before '::~' [clang-diagnostic-dtor-name]
   88 | Sacado::Fad::Vector< OrdinalType, Sacado::Fad::DVFad<ValueType> >::
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
      |                                                                  ::Vector
11957 warnings and 1 error generated.
``` 

This PR solves this issue by putting this piece of code explicitly in the namespace `Sacado::Fad`. 

Note that I've tried fixing it the other way around by writing something like 

```
~Vector<OrdinalType, Sacado::Fad::DVFad<ValueType>()
```

That works with `clang`, but then `gcc` complains. The proposed fix seems to work for both `clang` and `gcc`.

### DIFF
--- a/packages/sacado/src/Sacado_Fad_VectorImp.hpp
+++ b/packages/sacado/src/Sacado_Fad_VectorImp.hpp
@@ -84,12 +84,9 @@ Vector(const Sacado::Fad::Vector< OrdinalType, Sacado::Fad::DVFad<ValueType> >& 
   }
 }
 
-namespace Sacado::Fad
-{
-
 template <typename OrdinalType, typename ValueType>
-Vector< OrdinalType, DVFad<ValueType> >::
-~Vector()
+Sacado::Fad::Vector< OrdinalType, Sacado::Fad::DVFad<ValueType> >::
+~Vector< OrdinalType, Sacado::Fad::DVFad<ValueType> >()
 {
   // Here we must destroy the value and derivative arrays
   if (vec_.size() > 0) {
@@ -101,8 +98,6 @@ Vector< OrdinalType, DVFad<ValueType> >::
     }
   }
 }
-
-} // namespace Sacado::Fad
 
 template <typename OrdinalType, typename ValueType>
 Sacado::Fad::Vector< OrdinalType, Sacado::Fad::DVFad<ValueType> >&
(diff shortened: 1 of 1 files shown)


etphipp (approved):

maartenarnst:
Hi @etphipp. 

It seems one of the builds failed for this PR. Would you have a moment to take a look to have some info on the cause?

Perhaps just a word about this PR, which may seem a bit random :)

-> We compile our code with "warning as error". Because this issue is in a header file, it shows up as an error in our builds.

-> In our builds, it's actually `Intrepid2_OrientationTools.hpp` that we include, and this include leads, after a long sequence of includes, to including the Sacado header with this issue.

Thanks in advance!

etphipp:
Sorry, I somehow missed the failed autotester.  There are several build errors in the Intel build, all of the form
```
compilation aborted for ../Trilinos/packages/intrepid2/unit-test/MonolithicExecutable/AnalyticPolynomialsMatchTests.cpp (code 1)
```
They are all in Intrepid2 and Panzer.  It isn't clear to me how to resolve that.

maartenarnst:
Ok, thanks. It’s not very clear indeed. Should we try a retest to make sure there’s a problem?

etphipp:
Its a compiler problem, most likely a bug (which is not unusual for this compiler).  If you give me access to your branch I have two potential workarounds we can try.

maartenarnst:
OK. Thanks a lot. I've just added you to our fork of Trilinos. I think GitHub should have sent an invite.

etphipp:
Well, that didn't work.  Trying something else.

etphipp:
Well, I tried a couple of things, but nothing seems to work with the Intel compiler.  I'm not sure what else to do.  Are you getting the warning that motivated this PR when compiling with '-pedantic'?  All I can suggest is removing that flag.

etphipp (approved):

maartenarnst:
Thanks, Eric!

PR #12356, opened 2023-10-03
MueLu Amesos2Smoother: Only set "IsContiguous" to false for single rank

cgcgcg:
@trilinos/muelu 

## Motivation
On multiple ranks Amesos2 performs an import into a single rank first.


cgcgcg:
@ikalash Could you try this commit? If that doesn't work I'll just revert.

cgcgcg:
Ok, lemme write a unit test for this. Hopefully I can get this merged before tonight.

ikalash:
I just tested this and all Albany tests pass.  Thank you!

PR #12357, opened 2023-10-03
MueLu: Fixing two Maxwell1 bugs

csiefer2:
Issue: #11214
Stakeholder feedback: n/a
Tests: Already there

Fixing two bugs in Maxwell1.  The first was a map inconsistency issue which shows up in parallel.  This was a the-code-is-wrong bug.  The second was an inconsistency in the tie-break polarity vs. ML.  These are both fixed in this PR


rstumin (approved):
look good.

PR #12362, opened 2023-10-04
Framework: Enable CMAKE_LINK_LIBRARIES_ONLY_TARGETS

sebrowne:
Handed-off from #12339.  I also noticed that Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT was set too low in the top-level CMakeLists.txt, so I repaired that as well.

@trilinos/framework

## Motivation
Want to adopt CMake best practices, and `target_link_libraries(myProjectLib1 PUBLIC someLibNotInMyProject)` is not one of them, because it restricts CMake's ability to reason about missing link libraries at configure time.  Far preferable to error for this type of thing at configure time as opposed to build time.

## Related Issues
#10081

## Testing
This should immediately fail without fixing some bad 'm' linkages in zoltan, trilinosss, etc.


CMakeLists.txt
```diff
@@ -93,6 +93,10 @@ SET(${PROJECT_NAME}_EXCLUDE_DISABLED_SUBPACKAGES_FROM_DISTRIBUTION_DEFAULT FALSE
 
 SET(Trilinos_USE_GNUINSTALLDIRS_DEFAULT ON)
 
+SET(Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT TRUE)
+
+SET(CMAKE_LINK_LIBRARIES_ONLY_TARGETS)
```
bartlettroscoe, line 98:
```suggestion
SET(CMAKE_LINK_LIBRARIES_ONLY_TARGETS TRUE)
```
Fun fact, the command:

```
SET(CMAKE_LINK_LIBRARIES_ONLY_TARGETS)
```

actually **unsets** the variable `CMAKE_LINK_LIBRARIES_ONLY_TARGETS` if it is already set in the current scope (but does not unset a cache var of the same name).  See:

* https://cmake.org/cmake/help/latest/command/set.html#normal

which says:

> If no value is given, unset the variable. This is equivalent to [unset(<variable>)](https://cmake.org/cmake/help/latest/command/unset.html#command:unset).

Pretty crazy, right?

bartlettroscoe (changes requested):
Need to set a value for `CMAKE_LINK_LIBRARIES_ONLY_TARGETS`.

bartlettroscoe:
FYI: I set `AT: WIP` to avoid the autotester to run until addressing the above issue.

CMakeLists.txt
sebrowne: Would it make a difference if I promised that I typed ON there before and that obviously my computer messed it up, not me?  Fixed.

bartlettroscoe (approved):
That should do it.  Let's see what testing shows. It would also be nice to at least locally verify that setting `Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT=TRUE` results in an error if the user tries to list a library in `<TPLNAME>_LIBRARY_NAMES` that can't be found.

sebrowne:
Attempted to fix the same way I fixed this for SIERRA (which is incidentally exactly the example at https://cmake.org/cmake/help/latest/prop_tgt/LINK_LIBRARIES_ONLY_TARGETS.html).

Will test the MUST_FIND_ALL_TPL_LIBS locally tomorrow.

CMakeLists.txt
```diff
@@ -93,6 +93,13 @@ SET(${PROJECT_NAME}_EXCLUDE_DISABLED_SUBPACKAGES_FROM_DISTRIBUTION_DEFAULT FALSE
 
 SET(Trilinos_USE_GNUINSTALLDIRS_DEFAULT ON)
 
+SET(Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT TRUE)
+
+SET(CMAKE_LINK_LIBRARIES_ONLY_TARGETS ON)
+
+add_library(toolchain::m INTERFACE IMPORTED)
+set_property(TARGET toolchain::m PROPERTY IMPORTED_LIBNAME "m")
```
bartlettroscoe, line 101: My only concern is that this creates more clutter in the top CMakeLists.txt file and makes the packages less self-contained.  Why not just copy define two targets `Zoltan::m` and `SuiteSparse::m` and put them in the packages themselves?

bartlettroscoe (dismissed):
Okay to merge but I think could be improved (see below comment).

CMakeLists.txt
sebrowne: I'm willing to do that, since it keeps the packages more independently-buildable.  Stand by

bartlettroscoe (approved):
Looks great.  Thanks for moving the targets to the packages!  (One less thing we have to do to get any of the Trilinos packages to be independently buildable as their own CMake projects outside of Trilinos.)

PR #12364, opened 2023-10-04
Piro: Decouple MassSpringDamperModel from the rest

kliegeois:
@trilinos/piro
@mperego 

This PR decouples MassSpringDamperModel to be run with only one MPI rank.

### DIFF
--- a/packages/framework/ini-files/config-specs.ini
+++ b/packages/framework/ini-files/config-specs.ini
@@ -919,7 +919,6 @@ opt-set-cmake-var Trilinos_ENABLE_Intrepid BOOL : OFF
 opt-set-cmake-var Trilinos_ENABLE_Komplex BOOL : OFF
 opt-set-cmake-var Trilinos_ENABLE_Moertel BOOL : OFF
 opt-set-cmake-var Trilinos_ENABLE_ML BOOL : OFF
-opt-set-cmake-var Trilinos_ENABLE_Piro BOOL : OFF
 opt-set-cmake-var Trilinos_ENABLE_Rythmos BOOL : OFF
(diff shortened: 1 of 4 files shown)


rppawlo (approved):

sebrowne:
Yes.  It's failing because Piro is explicitly disabled for the CUDA builds (no Epetra packages), but the reduced tarball test is deciding that it should be included (since this PR changes Piro).  I'll look at the test real quick to see if there's a way to fix it.

sebrowne:
Wait, now that I say that, should Piro be disabled for non-epetra builds?  It looks like Piro's dependency on Epetra is optional.

mperego:
> Wait, now that I say that, should Piro be disabled for non-epetra builds? It looks like Piro's dependency on Epetra is optional.
I think Piro should be enabled in non-epetra builds.

sebrowne:
Oh darn it.  This was my fault.  When I supplied the list of packages to deprecate to the team earlier this year I typed 'Piro' when I meant 'Pike'.  Please delete any lines that say `opt-set-cmake-var Trilinos_ENABLE_Piro BOOL : OFF` from `packages/framework/ini-files/gen-config.ini` (lines 922 and 940 I believe) and add them to the PR, then this should work.

sebrowne:
Also I apologize, this will likely build/test everything since it touches the Framework directory.  Completely my fault here.

mperego:
> Also I apologize, this will likely build/test everything since it touches the Framework directory. Completely my fault here.

No worries, Sam. It's good we quickly figured out the issue.

kliegeois:
Thanks @sebrowne @csiefer2 @rppawlo and @mperego for the comments, fixes, and reviews!

Issue #12381, opened 2023-10-09
Ifpack2: compilation error


ikalash:
In the LCM nightlies, we are getting a failed Trilinos build with the following error: 

```
[ 84%] Building CXX object packages/anasazi/tpetra/src/CMakeFiles/anasazitpetra.dir/AnasaziTpetraAdapter.cpp.o
In file included from /home/lcm/LCM/trilinos-build-serial-gcc-release/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_Serial.cpp:50:
/home/lcm/LCM/Trilinos/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp: In member function ‘void Ifpack2::AdditiveSchwarz<MatrixType, LocalInverseType>::describe(Teuchos::FancyOStream&, Teuchos::EVerbosityLevel) const’:
/home/lcm/LCM/Trilinos/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp:1336:34: error: ‘rcp_dynamic_cast’ was not declared in this scope
           auto ifpack2_inverse = rcp_dynamic_cast< Ifpack2::Details::LinearSolver<scalar_type,local_ordinal_type,global_ordinal_type,node_type> > (Inverse_);
                                  ^~~~~~~~~~~~~~~~
/home/lcm/LCM/Trilinos/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_def.hpp:1336:34: note: suggested alternative:
In file included from /home/lcm/LCM/Trilinos/packages/teuchos/core/src/Teuchos_ArrayRCPDecl.hpp:47,
                 from /home/lcm/LCM/Trilinos/packages/teuchos/core/src/Teuchos_ArrayRCP.hpp:46,
                 from /home/lcm/LCM/Trilinos/packages/teuchos/core/src/Teuchos_Array.hpp:52,
                 from /home/lcm/LCM/Trilinos/packages/tpetra/core/src/Tpetra_ConfigDefs.hpp:78,
                 from /home/lcm/LCM/Trilinos/packages/ifpack2/src/Ifpack2_ConfigDefs.hpp:48,
                 from /home/lcm/LCM/trilinos-build-serial-gcc-release/packages/ifpack2/src/Ifpack2_AdditiveSchwarz_Serial.cpp:44:
[...]

PR #12386, opened 2023-10-10
Ifpack2: Hopefully fixing LCM Issue

csiefer2:
Issue: #12381
Stakeholder feedback: @ikalash please test!
Test: n/a

Well, I can't reproduce your issue directly.  But this looks like it should fix it.


ikalash (approved):
I tested this and it appears to fix the problem.  Thanks!

GrahamBenHarper (approved):
blessing also based on @ikalash's feedback

Issue #12376, opened 2023-10-07
Trilinos no longer finding Netcdf when Netcdf exists


ikalash:
The Albany nightlies started failing today due to Netcdf libs not being found even though they are there and are pointed to.  Here is how the error looks like: 

```
--   Searching for lib 'netcdf' ...
--     Found lib '/nightlyCDash/albany-tpls-gcc-11.1.1-openmpi-4.1.0/lib/libnetcdf.so'
-- Searching for a lib in the set "pnetcdf":
--   Searching for lib 'pnetcdf' ...
-- ERROR: Did not find a lib in the lib set "pnetcdf" for the TPL 'Netcdf'!
-- ERROR: Could not find the libraries for the TPL 'Netcdf'!
-- TIP: If the TPL 'Netcdf' is on your system then you can set:
     -DNetcdf_LIBRARY_DIRS='<dir0>;<dir1>;...'
   to point to the directories where these libraries may be found.
   Or, just set:
     -DTPL_Netcdf_LIBRARIES='<path-to-libs0>;<path-to-libs1>;...'
   to point to the full paths for the libraries which will
   bypass any search for libraries and these libraries will be used without
   question in the build.  (But this will result in a build-time error
   if not all of the necessary symbols are found.)
-- ERROR: Failed finding all of the parts of TPL 'Netcdf' (see above), Aborting!

-- TPL_Netcdf_PARALLEL is False
-- NOTE: The find module file for this failed TPL 'Netcdf' is:
     /nightlyAlbanyTests/Results/Trilinos/cmake/tribits/common_tpls/FindTPLNetcdf.cmake
   which is pointed to in the file:
     /nightlyAlbanyTests/Results/Trilinos/TPLsList.cmake

TIP: Even though the TPL 'Netcdf' was explicitly enabled in input,
it can be disabled with:
  -DTPL_ENABLE_Netcdf=OFF
which will disable it and will recursively disable all of the
downstream packages that have required dependencies on it.
When you reconfigure, just grep the cmake stdout for 'Netcdf'
and then follow the disables that occur as a result to see what impact
this TPL disable has on the configuration of Trilinos.

CMake Error at cmake/tribits/core/package_arch/TribitsProcessEnabledTpls.cmake:278 (message):
  ERROR: TPL_Netcdf_NOT_FOUND=TRUE, aborting!

```
[...]

PR #12387, opened 2023-10-10
Framework: Fix location of TriBITS default var

sebrowne:
Need default settings for TriBITS to be set prior to TRIBITS_PROJECT() call.

User Support Ticket(s) or Story Referenced: N/A

@trilinos/framework 

## Motivation
Want to ensure cleaner Trilinos configurations with respect to TPLs, and the previous change didn't get this setting turned on in the correct spot.

## Related Issues

This may re-introduce #12376

## Testing
Technically I think that #12376 tests and confirms this is working.


bartlettroscoe (approved):
Yup, it needs to go before `tribits_project()`.  I should have seen that on the first review.

sebrowne:
Intel compiler appears to have run out of memory.  Retesting.

PR #12392, opened 2023-10-11
Ifpack2: Adding BC-respecting initial guess for Chebyshev eigenvalue estimation

csiefer2:
Issue: #11214
Stakeholder feedback: Fixes the application
Tests: pass

### DIFF
--- a/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
+++ b/packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
@@ -195,7 +195,36 @@ struct LapackHelper{
                            Teuchos::ArrayRCP<typename Teuchos::ScalarTraits<ScalarType>::magnitudeType> offdiag) {
     throw std::runtime_error("LAPACK does not support the scalar type.");
   }
+
 };
+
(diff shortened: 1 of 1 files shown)


packages/ifpack2/src/Ifpack2_Details_Chebyshev_def.hpp
```diff
@@ -195,8 +195,37 @@ struct LapackHelper{
...
+
+
+  // Zero the stuff that where the diagonal is equal to one.  These are assumed to
+  // correspond to OAZ rows in the matrix.
+  size_t N = x.getLocalLength();
+  auto d_view = diagonal.template getLocalView<device_type>(Tpetra::Access::ReadOnly);
+  auto x_view = x.template getLocalView<device_type>(Tpetra::Access::ReadWrite);
+
+  auto ONE  = Teuchos::ScalarTraits<typename V::scalar_type>::one();
+  auto ZERO = Teuchos::ScalarTraits<typename V::scalar_type>::zero();
```
cgcgcg, line 218: I think you might need the impl scalar type here. Then again, we probably don't want to call Cheby for complex.

Issue #12333, opened 2023-09-27
Belos:  BelosTpetraTestFramework.hpp compilation fatal error: 'Trilinos_Util_iohb.h' file not found


ndellingwood:
## Bug Report
@trilinos/belos 

### Description
Compilation with Hip+rocm/5.2.0, Serial+gcc/11.3.0 fails with output of the form:
```
03:24:40 In file included from /home/jenkins/caraway-new/workspace/Trilinos_Caraway_Hip_Serial_Rocm5_2_0/Trilinos/packages/belos/tpetra/example/TFQMR/PseudoBlockTFQMRTpetraExFile.cpp:73:
03:24:40 /home/jenkins/caraway-new/workspace/Trilinos_Caraway_Hip_Serial_Rocm5_2_0/Trilinos/packages/belos/tpetra/src/BelosTpetraTestFramework.hpp:56:10: fatal error: 'Trilinos_Util_iohb.h' file not found
03:24:40 #include "Trilinos_Util_iohb.h"
```

I scanned over the failing file for changes the include line has been there since 2019, not sure what would trigger this kind of error...

This is a list of changes that possibly could have triggered the error

[#745 (Sep 26, 2023 10:12:00 PM)](https://jenkins-son.sandia.gov/job/Trilinos_Caraway_Hip_Serial_Rocm5_2_0/745/changes)

    Tpetra: fix uninitialized const ref complaints from clang 14 compiler — [Julia A. Plews](https://jenkins-son.sandia.gov/user/japlews/) / [detail](https://jenkins-son.sandia.gov/job/Trilinos_Caraway_Hip_Serial_Rocm5_2_0/745/changes#detail0)
    ShyLU: Make DebugManager take a user's comm in constructor — [domagala.jacob](https://jenkins-son.sandia.gov/user/domagala.jacob/) / [detail](https://jenkins-son.sandia.gov/job/Trilinos_Caraway_Hip_Serial_Rocm5_2_0/745/changes#detail1)
    Belos: add Tpetra GCRODR test — [antoine.meyer54](https://jenkins-son.sandia.gov/user/antoine.meyer54/) / [detail](https://jenkins-son.sandia.gov/job/Trilinos_Caraway_Hip_Serial_Rocm5_2_0/745/changes#detail2)
[...]

PR #12401, opened 2023-10-13
belos: guard BlockGmres tests that require Triutils

ndellingwood:
Resolve missing header error when Triutils is not enabled: "fatal error: Trilinos_Util_iohb.h: No such file or directory"

@trilinos/belos @hkthorn 

## Motivation

Resolve compilation errors when Triutils is not enabled:
```
Trilinos-kp/packages/belos/tpetra/src/BelosTpetraTestFramework.hpp:56:10: fatal error: Trilinos_Util_iohb.h: No such file or directory
   56 | #include "Trilinos_Util_iohb.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~
```

## Testing
Local testing, AT

### DIFF
--- a/packages/belos/tpetra/test/BlockGmres/CMakeLists.txt
+++ b/packages/belos/tpetra/test/BlockGmres/CMakeLists.txt
@@ -21,14 +21,6 @@ IF (${PACKAGE_NAME}_ENABLE_Triutils)
      )
 
   ENDIF()
-ENDIF()
-
-TRIBITS_ADD_EXECUTABLE_AND_TEST(
-  Tpetra_BlockGMRES_hb_test
-  SOURCES test_bl_gmres_hb.cpp
-  ARGS "--verbose"
-  COMM serial mpi
-  )
 
 TRIBITS_ADD_EXECUTABLE_AND_TEST(
   Tpetra_resolve_gmres_hb
@@ -46,6 +38,14 @@ TRIBITS_ADD_EXECUTABLE_AND_TEST(
   COMM serial mpi
   ARGS "--verbose --filename=orsirr1.hb"
   STANDARD_PASS_OUTPUT
+  )
+ENDIF()
+
+TRIBITS_ADD_EXECUTABLE_AND_TEST(
+  Tpetra_BlockGMRES_hb_test
+  SOURCES test_bl_gmres_hb.cpp
+  ARGS "--verbose"
+  COMM serial mpi
   )
 
 ASSERT_DEFINED(Belos_ENABLE_TSQR)


ndellingwood:
Should help resolve #12333

packages/belos/tpetra/test/BlockGmres/CMakeLists.txt
```diff
@@ -47,6 +39,14 @@ TRIBITS_ADD_EXECUTABLE_AND_TEST(
   ARGS "--verbose --filename=orsirr1.hb"
   STANDARD_PASS_OUTPUT
   )
+ENDIF()
```
lucbv, line 42: @ndellingwood shouldn't the test be declared before the `ENDIF()` statement if it requires Triutils to be ON?

lucbv (commented):
I'm not sure what this change does?

packages/belos/tpetra/test/BlockGmres/CMakeLists.txt
hkthorn: @ndellingwood I have the same question.  ;)

hkthorn (changes requested):
Correct logic in the CMakeFile to ensure test is guarded by Triutils ifdef.

packages/belos/tpetra/test/BlockGmres/CMakeLists.txt
ndellingwood: @hkthorn @lucbv the diff looks weird - the change I made was to move the `test_resolve_gmres_hb.cpp` and `test_pseudo_gmres_hb.cpp` into the `IF (${PACKAGE_NAME}_ENABLE_Triutils)` block of code - these were the tests triggering the error without Triutils.
The diff makes it look like I shifted the `ENDIF()` (corresponding to `IF (${PACKAGE_NAME}_ENABLE_Triutils)`) down 18 lines and shifted the `test_bl_gmres_hb.cpp` test out of the block. Effectively the same thing, but looks misleading from the way I made the actual changes.
Please view the full file for confirmation, the git diff obscures the intent of the changes. FWIW I tested this locally while working on the fix and it resolves the issue

ndellingwood:
> Correct logic in the CMakeFile to ensure test is guarded by Triutils ifdef.

@hkthorn @lucbv the diff looks weird - the change I made was to move the test_resolve_gmres_hb.cpp and test_pseudo_gmres_hb.cpp into the IF (${PACKAGE_NAME}_ENABLE_Triutils) block of code - these were the tests triggering the error without Triutils.
The diff makes it look like I shifted the ENDIF() (corresponding to IF (${PACKAGE_NAME}_ENABLE_Triutils)) down 18 lines and shifted the test_bl_gmres_hb.cpp test out of the block. Effectively the same thing, but looks misleading from the way I made the actual changes.
Please view the full file for confirmation, the git diff obscures the intent of the changes. FWIW I tested this locally while working on the fix and it resolves the issue

PR #12183, opened 2023-08-30
Anasazi: Added Randomized Eigensolver

Heatherms27:
@trilinos/anasazi

Requested Reviewers:
Jennifer Loe
Erik Boman

## Motivation

We added into Anasazi a new solver that creates a block of orthonormal random initial starting vectors before performing subspace iteration by applying a matrix A to the block a specified number of times. The user can also specify how often to orthonormalize the basis + check the residuals for convergence. 
The purpose of this solver is to run a cheap eigensolver that sacrifices accuracy for speed. It is currently being tested in conjunction to spectral graph partitioning in Sphynx under the Zoltan2 package.

## Testing

Directly testing in:
/Trilinos/packages/anasazi/tpetra/test/Randomized/rand_driver.cpp
with the executable
/Trilinos/build/packages/anasazi/tpetra/test/Randomized/Anasazi_Randomized_Solver_Driver.exe

The randomized solver has almost been added as an option into:
/Trilinos/packages/zoltan2/test/sphynx/Sphynx_Research_Driver.cpp
with the executable
Trilinos/build/packages/zoltan2/test/sphynx/Zoltan2_Sphynx_Research_Driver.exe

## Additional Information
Have also added more test files which are identical to pre-existing "cxx_main_doublefloat.cpp" tests , but allow for Matrix Market files as input.
/anasazi/tpetra/test/LOBPCG/cxx_main_test.cpp -> Anasazi_Tpetra_LOBPCG_test.exe
/anasazi/tpetra/test/BlockKrylovSchur/cxx_main_test.cpp -> Anasazi_Tpetra_BlockKrylovSchur_test.exe
/anasazi/tpetra/test/BlockDavidson/cxx_main_test.cpp -> Anasazi_Tpetra_BlockDavidson_test.exe

PR #12414, opened 2023-10-17
Anasazi: Add Randomized Eigensolver (Clean PR)

jennloe:
@trilinos/anasazi 

This is a clean version of PR #12183 which had some rebasing issues. I will close the other PR momentarily. 
The primary work for these changes was done by Heather Switzer (@Heatherms27)

## Motivation
We added into Anasazi a new solver that creates a block of orthonormal random initial starting vectors before performing subspace iteration by applying a matrix A to the block a specified number of times. The user can also specify how often to orthonormalize the basis + check the residuals for convergence.
The purpose of this solver is to run a cheap eigensolver that sacrifices accuracy for speed. It is currently being tested in conjunction to spectral graph partitioning in Sphynx under the Zoltan2 package.


egboman (approved):
Looks good. If Anasazi had a factory method, Zoltan2/Sphynx could have used that to select *any* eigensolver, but it doesn't.

sebrowne:
ctest appeared to time-out the Jenkins job for the GPU tests (the ones that aren't submitted), which meant that the results did not submit.  This is an unfortunate behavior in the current autotester.  Recommend AT: RETEST for this one.

PR #12419, opened 2023-10-18
KokkosKernels: controls: don't print a noisy warning

cwpearson:
Match a change made in kokkos-kernels upstream 86d8371fdb3d528095a86635b9add53a36bb516e

The next kokkos-kernels snapshot will include this change


ndellingwood (approved):
Change was part of kokkos/kokkos-kernels#1740

PR #12429, opened 2023-10-20
Tpetra: in TAFC iterate over domainMap efficiently

brian-kelley:
Iterate over domainMap efficiently on device using local map ``getGlobalElement(lid)``, instead of getting a host view of GIDs and copying that to device. This avoids a copy and is faster if the domain map is contiguous (where ``getGlobalElement()`` is just an addition).

@trilinos/tpetra 

## Testing

Covered by several CrsMatrix unit tests: Bug10008_2, Bug10008_3, Bug8447.


jhux2 (approved):
lgtm, thanks @brian-kelley

PR #12441, opened 2023-10-25
Framework: yml file to enable a gcc-openmpi container build that can run for PR

fryeguy52:
@trilinos/framework 

## Motivation
This will enable a containerized build through github actions. The idea is that it will be run for all PR

### DIFF
--- a/.github/workflows/PR-gcc-openmpi.yml
+++ b/.github/workflows/PR-gcc-openmpi.yml
@@ -35,6 +35,7 @@ jobs:
         working-directory: /
         run: |
           mkdir -p /home/Trilinos/src/Trilinos
+	  mkdir -p /home/Trilinos/build
       - name: clone trilinos
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/PR-gcc-openmpi.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: |
+          bash -l -c "./get_dependencies.sh --container"
+      - name: Generate PR cmake fragment
+        working-directory: /home/Trilinos/build/
+        run: |
+          bash -l -c "source /home/Trilinos/src/Trilinos/packages/framework/GenConfig/gen-config.sh rhel8_gcc-openmpi_debug_shared_no-kokkos-arch_no-asan_complex_no-fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-on_no-package-enables --cmake-fragment PR-configuration.cmake /home/Trilinos/src/Trilinos"
+      - name: Generate enable packages cmake fragment
+        working-directory: /home/Trilinos/build/
+        run: |
+          bash -l -c "TRILINOS_DIR=/home/Trilinos/srs/Trilinos /home/Trilinos/srs/Trilinos/commonTools/framework/get-changed-trilinos-packages.sh HEAD develop package-enables.cmake ctest-labels.cmake"
```
sebrowne, line 57: 'develop' will need to be dynamic, the merge target won't always be the develop branch.  This is fine for prototyping though.

sebrowne (approved):

sebrowne (commented):
Why are so many tests failing?  Can you get output from a couple of them to see if we need to change something in the container?

sebrowne (dismissed):
The failing tests are small enough now in magnitude that I'm okay with this.  Want to get those figured out pretty soon though.

packages/framework/ini-files/config-specs.ini
```diff
@@ -2848,13 +2849,15 @@ use PACKAGE-ENABLES|NO-PACKAGE-ENABLES
 
 use COMMON_SPACK_TPLS
 
-opt-set-cmake-var MPI_EXEC_PRE_NUMPROCS_FLAGS                            STRING       : --bind-to;none
```
sebrowne, line 2851: I'd like us to stop aligning columns like this for exactly this reason.  It generates a large diff when you really changed a couple of things.  Don't worry about it for this one, but for the future I do not think keeping the columns aligned is a good idea (https://stackoverflow.com/questions/101958/code-formatting-is-lining-up-similar-lines-ok agrees with me, though that's not exactly an iron-clad source).

sebrowne (approved):

sebrowne:
@cgcgcg thank you for catching that!  I will reach out to Eric and get that changed to PreTest-Inspectors

triple3567:
@cgcgcg should be fixed now. Anyone in the "pretest-inspectors" group can start builds by setting the "AT: PRE-TEST INSPECTED" Label, or by opening a new PR, or by committing changes to an existing PR.

cgcgcg:
@triple3567 Is there a way to trigger this not only by setting "AT: PRE-TEST INSPECTED" but also by approving a PR? I tried this on #12491 and before the label was set the action failed early on. Is that the expected behavior?

Issue #12391, opened 2023-10-10
Random test timeouts bringing down PR build & test iterations 2023


bartlettroscoe:
CC: @ccober6, @sebrowne, @trilinos/framework 

## Description

Anecdotal evidence seems to suggest that random test failures, including random timeouts, are bringing down PR testing iterations fairly regularly.  When this happens, all of the builds need to be run again from scratch, wasting testing computing resources, blocking   PR testing iterations for other PRs, and delaying the merge of PRs.

For example, [this query over the last two months](https://trilinos-cdash.sandia.gov/queryTests.php?project=Trilinos&begin=2023-08-10&end=2023-10-10&filtercount=3&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Pull%20Request&field2=status&compare2=61&value2=failed&field3=details&compare3=63&value3=Timeout) suggest that random test timeouts took out PR testing iterations for the following PRs:

* #12388 
* #12372 
* #12367
* #12281 
* #12259
* #12223 
* #12103

Note that the test timeouts for the PRs #12050 and #12297 shown in that query don't appear to be random.  Filtering out those PRs yields [this reduced query over the last two months](https://trilinos-cdash.sandia.gov/queryTests.php?project=Trilinos&begin=2023-08-10&end=2023-10-10&filtercount=5&showfilters=1&filtercombine=and&field1=groupname&compare1=61&value1=Pull%20Request&field2=status&compare2=61&value2=failed&field3=details&compare3=63&value3=Timeout&field4=buildname&compare4=64&value4=PR-12050&field5=buildname&compare5=64&value5=PR-12297) shows the 7 randomly failing tests:

<html>
<body>

Site | Build Name | Test Name | Status | Time | Proc Time | Details | Build Time | Processors
-- | -- | -- | -- | -- | -- | -- | -- | --
ascic164 | PR-12388-test-rhel7_sems-intel-2021.3-sems-openmpi-4.0.5_release-debug_shared_no-kokkos-arch_no-asan_no-complex_fpic_mpi_no-pt_no-rdc_no-uvm_deprecated-off_no-package-enables-1204 | ROL_example_PDE-OPT_ginzburg-landau_example_01_MPI_4 | Failed | 10m 130ms | 40m 520ms | Completed (Timeout) | 2023-10-10T12:08:21 MDT | 4
[...]

PR #12442, opened 2023-10-25
Framework: Relocate CXX standard setup

sebrowne:
Need to process CXX standard after a potential CONFIGURE_OPTIONS_FILE has been processed.  This manifested a small bug where the configure message about which CXX standard was being used showed 17 even though the entry in the Trilinos_CONFIGURE_OPTIONS_FILE read 20 (and in fact it did use 20).

## Motivation
Want diagnostic messages to align with the actual behavior.

## Testing
Tested using `-DTrilinos_CONFIGURE_OPTIONS_FILE` and it now behaves correctly.

### DIFF
--- a/cmake/CallbackSetupExtraOptions.cmake
+++ b/cmake/CallbackSetupExtraOptions.cmake
@@ -92,3 +92,6 @@ include("${Trilinos_SOURCE_DIR}/commonTools/build_stats/BuildStatsWrappers.cmake
 generate_build_stats_wrappers()
 remove_build_stats_file_on_configure()
 remove_build_stats_timing_files_on_fresh_configure()
+
+# Set up C++ language standard selection
+include(SetTrilinosCxxStandard)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,9 +96,6 @@ SET(Trilinos_USE_GNUINSTALLDIRS_DEFAULT ON)
 
 SET(Trilinos_MUST_FIND_ALL_TPL_LIBS_DEFAULT TRUE)
 
-# Set up C++ language standard selection
-include(SetTrilinosCxxStandard)
-
 # Some CMake and TriBiTS tweaks just for Trilinos
 include(TrilinosTweaks)
 


cmake/CallbackSetupExtraOptions.cmake
```diff
@@ -92,3 +92,6 @@ include("${Trilinos_SOURCE_DIR}/commonTools/build_stats/BuildStatsWrappers.cmake
 generate_build_stats_wrappers()
 remove_build_stats_file_on_configure()
 remove_build_stats_timing_files_on_fresh_configure()
+
+# Set up C++ language standard selection
+include(SetTrilinosCxxStandard)
```
bartlettroscoe, line 97: This should do it but I am not sure why stuff got placed outside of the macro `TRIBITS_REPOSITORY_SETUP_EXTRA_OPTIONS()` that gets defined in this file and called by TriBITS.  Semantically, there is not a big difference in behavior except that this code outside of the macro `TRIBITS_REPOSITORY_SETUP_EXTRA_OPTIONS()` in that file actually gets called before the code inside of the macro as you can see in:

https://github.com/trilinos/Trilinos/blob/3ff154a7d724400b53afa726d868408ede667e39/cmake/tribits/core/package_arch/TribitsPackagingSupport.cmake#L312-L314

But I see that I was the one who set the precedent for this by adding the build_stats stuff there in commit 3ea94177a4a38e847fb5a78fe0d9edf7d7ebd1d4 with no explanation for why that was done.

SIDENOTE: I don't remember why I decided to put things in a macro in these files because just including CMake code in a file is largely equivalent to calling a macro.

sebrowne:
Configure output indicates all is still well, C++17 is being used in the PR builds.

cmake/CallbackSetupExtraOptions.cmake
sebrowne: I'll do whichever you think is best.  I noticed that too, but decided to put it in (IMO) the simpler spot

bartlettroscoe: We can do this in a separate PR now that PR testing has already started.

bartlettroscoe:
FYI: Last PR testing iteration got taken out by random build errors in two configurations and a random test timeout reported in https://github.com/trilinos/Trilinos/issues/12391#issuecomment-1780249515.  The build errors showed:

```
```

and

```
g++: fatal error: Killed signal terminated program cc1plus
compilation terminated.
```

sebrowne:
Yes, almost certainly a load issue on the machine.  RETESTing.

Issue #12237, opened 2023-09-11
Tpetra:  compilation error with Cuda 12.2 and GCC 12.3


maartenarnst:
@brian-kelley @csiefer2 

We are compiling Trilinos with Cuda 12.2 and with GCC 12.3 as the host compiler.

We're seeing a compilation error for the file `Tpetra_CrsMatrix_def.hpp`:

```
INFO:root:#21 1477.5 ...build-with-GNU-Cuda-amd64/packages/tpetra/core/src/Tpetra_CrsMatrix_LONG_LONG_INT_LONG_LONG_SERIAL.cpp:74:16:   required from here
INFO:root:#21 1477.5 .../packages/tpetra/core/src/Tpetra_CrsMatrix_def.hpp:1689:139: error: no type named 'const_type' in 'cuda::std::__4::__is_primary_template<cuda::std::__4::iterator_traits<long long int, void> >'
INFO:root:#21 1477.5  1689 |         typename row_entries_type::const_type numRowEnt_h =
INFO:root:#21 1477.5       | 
```

It's quite mysterious because `row_entries_type ` seems to be a `Kokkos::View`, so the line should be fine.

I think I tracked it down to this commit:

- https://github.com/trilinos/Trilinos/commit/0abba21aeea1bbf57212f3af27552fb38794540b

which ultimately results in an include of `Kokkos_Sort.hpp` in `Tpetra_CrsMatrix_def.hpp`. It seems to be the include of `thrust/device_ptr.h` and `thrust/sort.h` from `Kokkos_Sort.hpp`  that ultimately causes the issue. I.e., if I compile using an older version of Trilinos and add those two thrust includes to `Tpetra_CrsMatrix_def.hpp`, I get the same error. 

This is just a bug report. I have no explanation for this compilation error. And no fix to propose.

jhux2:
@trilinos/tpetra

brian-kelley:
@csiefer2 Weaver has both of those

maartenarnst:
Hi @csiefer2. Thanks for following up.

We're building and testing in a docker container based on the cuda:12.2.0-devel-ubuntu22.04 image. It's x86, GCC 12.3, Cuda 12.2 and openmpi 4.1.2.

I'll try to run checks again tomorrow with our configuration, as well as with gcc 12.2 and cuda 12.0 that your are using. I'll keep you updated, and I'll also send more details.

Also tagging @romintomasetti.

bathmatt:
[...]

PR #12443, opened 2023-10-25
Tpetra: Working around GCC 12.3 + CUDA Compiler Bug #12237

csiefer2:
Issue: #12237
Stakeholder feedback: @maartenarnst can you check?
Tests: n/a

### DIFF
--- a/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
+++ b/packages/tpetra/core/src/Tpetra_CrsGraph_def.hpp
@@ -385,7 +385,7 @@ namespace Tpetra {
     resumeFill (params);
     checkInternalState ();
   }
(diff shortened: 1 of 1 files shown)


bathmatt:
It is the ones only with
-    typedef decltype (myGraph_->k_numRowEntries_) row_entries_type;
or
-    using row_entries_type = decltype (staticGraph_->k_numRowEntries_);

but only with row_entries_type

jhux2 (approved):
sure

bathmatt:
I confirm this compiles under cuda-12.3 which has the same bug as 12.2.  The compiler team is workgin a fix and hopefully this will make 12.4 and the WAR won't be needed anymore.  
M

PR #12447, opened 2023-10-25
Multiple packages: update name of Tpetra GPU-aware MPI option

brian-kelley:
@trilinos/framework 
@trilinos/tpetra 

## Motivation

Tpetra changed the names of its CUDA-aware MPI options to "GPU-aware" instead, to be more generic for HIP and SYCL. In particular, the old environment variable ``TPETRA_ASSUME_CUDA_AWARE_MPI=0|1`` was removed and has no effect now. The thing we overlooked was, the ``trilinos_jsrun`` script used this variable to auto-detect whether the ``-M -gpu`` (nranks > 1) or ``-M -disable_gpu_hooks`` (nranks == 1) should be passed to mpiexec.

This updates trilinos_jsrun to auto-detect using the correct environment variable ``TPETRA_ASSUME_GPU_AWARE_MPI`` now. If the old variable is set but the new one isn't, a warning is printed but the new one is set to the same value, so things should be able to run. This fixes the Vortex performance dashboard runs, for example.

Also went through other scripts that used the old option name and updated them.

## Testing

Tested by running some Vortex performance tests: once with ``TPETRA_ASSUME_GPU_AWARE_MPI=1``, and once with the old option ``TPETRA_ASSUME_CUDA_AWARE_MPI=1``. Both worked, and in the second case a warning was printed saying to switch to the other variable:
```
19: ** WARNING **, the environment variable TPETRA_ASSUME_CUDA_AWARE_MPI is no longer supported.
19:                Use TPETRA_ASSUME_GPU_AWARE_MPI instead.
19:                For now, will set TPETRA_ASSUME_GPU_AWARE_MPI=1 and continue.
19: BEFORE: jsrun  '-p' '1' '--rs_per_socket' '4' '/vscratch1/bmkelle/ATS2_Performance/build/packages/muelu/test/scaling/MueLu_SpMVPerformance.exe' '--node=serial' '--linAlgebra=Tpetra' '--nx=120' '--ny=120' '--nz=120' '--matrixType=Laplace3D' '--num-runs=1000'
19: AFTER: export TPETRA_ASSUME_GPU_AWARE_MPI=1; jsrun  '-M -disable_gpu_hooks' '-p' '1' '--rs_per_socket' '4' '/vscratch1/bmkelle/ATS2_Performance/build/packages/muelu/test/scaling/MueLu_SpMVPerformance.exe' '--node=serial' '--linAlgebra=Tpetra' '--nx=120' '--ny=120' '--nz=120' '--matrixType=Laplace3D' '--num-runs=1000'
```


sebrowne (commented):
I checked the code briefly and I feel it's fine to rename all instances of `Trilinos_CTEST_RUN_CUDA_AWARE_MPI`

packages/tpetra/scripts/PerformanceTesting/vortex/loadEnv.sh
```diff
@@ -0,0 +1,6 @@
+module load sparc-dev/cuda-11.2.152_gcc-8.3.1_spmpi-rolling
+export NVCC_WRAPPER_TMPDIR=/tmp
```
sebrowne, line 2: Recommend `/tmp/${USER}` instead (though also remember to mkdir that directory as well), we've seen problems in the past with the nvidia compiler's temp files having hash collisions between users depending on how heavily the machine is used.  That's actually exactly why I added the NVCC_WRAPPER_TMPDIR capability way back when.

sebrowne (dismissed):
Mostly a pattern-matching review from me.  Added a comment about `NVCC_WRAPPER_TMPDIR` value.

packages/tpetra/scripts/PerformanceTesting/vortex/loadEnv.sh
brian-kelley: @sebrowne That is a good point. I just tried keeping the ``/tmp/${USER}`` value from sparc-dev, and the only reason it failed is because ``/tmp/${USER}`` may not exist. Pretty sure that's the only reason I added this line.

But I can just ``mkdir -p /tmp/${USER}`` here instead. I will update this

sebrowne (approved):

Issue #12451, opened 2023-10-26
Tpetra: Simplify CMake / #ifdef logic


csiefer2:
Clean out the cruft.  Maybe simplify things.

PR #12452, opened 2023-10-26
Tpetra: Cleaning up CUDA cmake logic (part 1)

csiefer2:
Issue: #12451 
Stakeholder feedback: Verified supported versions on Kokkos slack
Tests: n/a

We don't support CUDA 7.5 any more, so we can clean up a lot of that logic.


jhux2 (approved):
lgtm

jhux2:
@sebrowne Is the AT acting up?  Latest results for this PR are missing data.

PR #12459, opened 2023-10-28
Intrepid2: fix namespace of Pyramid projection test

mperego:
@trilinos/intrepid2 

## Motivation

Fix to allow compilation with `Intrepid2_KEEP_EXPERIMENTAL_NAMESPACE:BOOL=OFF`, see #12141.


CamelliaDPG (approved):
Thanks for fixing this, @mperego!

PR #12460, opened 2023-10-30
MueLu: Removing superfluous output

csiefer2:
Issue: #11214 
Stakeholder feedback: Exposed by stakeholder test.  
Test: Exposed by stakeholder test.


cgcgcg (changes requested):
nope

PR #12462, opened 2023-10-30
tpetra(multivector): provide a space instance for routines invoking KokkosBlas under the hood

romintomasetti:
@trilinos/Tpetra

## Motivation

Any `Tpetra` routine that relies on `KokkosBlas` should be accepting a `space` instance.

This PR brings this for:
- `MultiVector::update`
- `MultiVector::elementWiseMultiply`

### DIFF
--- a/packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
+++ b/packages/tpetra/core/src/Tpetra_MultiVector_decl.hpp
@@ -1722,9 +1722,19 @@ namespace Tpetra {
     /// NaN entries.  It is legal for the input A to alias this
     /// MultiVector.
     void
-    update (const Scalar& alpha,
+    update (const execution_space& space,
+            const Scalar& alpha,
             const MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>& A,
             const Scalar& beta);
+
+    //! @overload
+    void
+    update (const Scalar& alpha,
+            const MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>& A,
+            const Scalar& beta)
+    {
+      this->update(execution_space{}, alpha, A, beta);
+    }
 
     /// \brief Update: <tt>this = gamma*this + alpha*A + beta*B</tt>.
     ///
@@ -1733,11 +1743,23 @@ namespace Tpetra {
     /// contains NaN entries.  It is legal for the inputs A or B to
     /// alias this MultiVector.
     void
-    update (const Scalar& alpha,
+    update (const execution_space& space,
+            const Scalar& alpha,
             const MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>& A,
             const Scalar& beta,
             const MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>& B,
             const Scalar& gamma);
+
+    //! @overload
+    void
+    update (const Scalar& alpha,
+            const MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>& A,
+            const Scalar& beta,
+            const MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>& B,
+            const Scalar& gamma)
+    {
+      this->update(execution_space{}, alpha, A, beta, B, gamma);
+    }
 
     /// \brief Compute the one-norm of each vector (column), storing
     ///   the result in a host view.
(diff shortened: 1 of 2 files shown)


romintomasetti:
@csiefer2 Ok. I'd be more than happy to participate!

Just to help me understand what you mean by serious thoughts. What would make `Tpetra::MultiVector::update` different from *e.g.* `Kokkos::sort` that has 2 versions, one taking a space instance and the other one not?

If we think of space instances as ways to allow the user to better orchestrate asynchronicity, priority, execution order or even "which GPU", I think such `Tpetra` function must follow the design `Kokkos` initiated.

Also note that for stakeholders not wanting to bother with this, the proposed change does not require any action on their side, nor that it changes the semantic meaning of their code.

romintomasetti:
@csiefer2 We've discussed this topic thoroughly this afternoon with @maartenarnst.

We observe that there are 2 types of methods in such `Tpetra` routines. And that both types can be made compliant with allowing a `space` instance to be passed, although we identify challenges.

### Methods that act on local data

This is typically the case of `Tpetra::MV::update` or `Tpetra::MV::elementWiseMultiply`. These methods act only on local device data, without any MPI or host-device synchronization other than those potentially incurred by `getLocalViewDevice`(*). Moreover, they are usually handing control to `KokkosBlas` routines. This shows that such methods should already accept a `space` instance. Moreover, allowing these methods to take a `space` instance does not make the code unsafe.

> (*) Note that if *e.g.* in `Tpetra::MV::update`, `getLocalViewDevice` leads to a host-device sync through `sync_device()`, hence it will be preceded by a `Kokkos::fence` and followed by another `Kokkos::fence`, see below for more details.

The second picture below shows what would happen if one uses `Tpetra::MV::update(space, ...)` followed by `Tpetra::CRS::apply(...)`.

* The code can still be safe if `Tpetra::CRS::apply` requires the host view before accessing device data, because the dual view will then say "Ok, but let's first make a sync with the device counterpart". Indeed, since none of the sync in `Tpetra` pass a `space` to the `Kokkos::DualView`, these syncs incur global fences with `Kokkos::fence` (through `Kokkos::deep_copy`, one `Kokkos::fence` before the copy, and one `Kokkos::fence` after the copy).
* Otherwise, if I launched the `Tpetra::MV::update(space, ...)` before calling `Tpetra::CRS::apply(...)` to which I cannot pass a space instance, and if the `Tpetra::CRS::apply(...)` code path does not require the host view (so no sync), I must as a user add a `space.fence("wait MV::update finishes before handing to another routine that I can't tell which space instance it uses")`. Only then I'm sure I'm on a 100% safe road :car: 

Note that the current PR leaves the current functionalities (and risks) unchanged (since calling `KokkosBlas::update(...)` forwards to `KokkosBlas::update(execution_space{}, ...)`).

### Methods that require MPI synchronization

This is typically the case of `Tpetra::CRSMatrix::apply` (it will import/export). Please refer to the first picture below. It depicts the following situation:
1. There are 2 MPI ranks.
2. On rank 0, there are 2 space instances. On rank 1, there is one space instance.
3. The 100% local kernel `CRS::update` runs on `space_1` on rank 0 and on `space_0` on rank 1.
4. Then, the user calls `CRS::apply` passing it the `space_0` on rank 0, and still uses `space_0` on rank 1.

In this use case, it is the user's responsibility to add the yellow link on rank 0. This link ensures that once `CRS::apply` is entered, the compute node `CRS::update` has finished modifying the data.

In the general case, such codes should be thought off as a **graph composed of nodes with dependencies**:
* a node is a kernel running on a specified space instance
* fences mark dependencies (*e.g.* `space_1.fence()`)

It is the user's duty to ensure that its dependencies graph is correct. It usually means adding necessary fences (as you'd do with `Tpetra::Details::Spaces::exec_space_wait` for instance).

This "graph thinking" allows to clearly identify implications for `Tpetra` routines. Indeed, since the user wanted for some reason to run `CRS::update` on `space_1` but then run `CRS::apply` on `space_0` and since these 2 will act on the same data, the user must add a `space_1.fence` before `CRS::apply`. Then, since the user required that `CRS::apply` runs on `space_0`, the challenge for `Tpetra` is to ensure all device kernels and host-device syncs are run using the provided `space_0` **or** any other space instance(s) as long as it is guaranteed that the dependencies graph is respected (think of the node `CRS::apply` as being itself a graph of inter-dependent nodes, see third image). A second challenge is that passing a `space` instance to `CRS::apply` also requires that *it cannot make any global fencing*.

## Summary

I think the key messages are:
1. For methods acting on local data as described above, I think the proposed PR is the best way to go. Note that it can still be modified in the future to use the object's space instance (part of the state) if no space instance is provided.
2. `Tpetra` calls like `CRS::apply` currently incur many global (implicit) fences and that can be easily modified. After all, thinking of `Tpetra` calls as a graph node (that itself can be a graph) makes the discussion easy to follow and suggests that it's feasible to make it more compliant with space instances, while keeping backward compatibility.
3. This discussion also points out that the wrapped dual view's `getLocal<Device/Host>View` should be allowed to take a space instance, to ensure that any potential host-device sync is placed correctly in the dependencies graph and can be done without requiring global fencing.

result

result

image

romintomasetti:
@csiefer2 Sorry! My mistake :smile: Moved to https://github.com/trilinos/Trilinos/issues/12463#issuecomment-1790385139 :flight_departure:

Issue #12427, opened 2023-10-20
Stokhos:  Stokhos_TpetraCrsMatrixUQPCEUnitTest_Cuda_MPI_4 unit test failure in Cuda build


ndellingwood:
## Bug Report
@trilinos/stokhos @etphipp 

### Description
The `Stokhos_TpetraCrsMatrixUQPCEUnitTest_Cuda_MPI_4` unit test fails in cuda/11.4.2 builds with the following output:
```
...
317: Cuda Runtime Configuration:
317: macro  KOKKOS_ENABLE_CUDA      : defined
317: macro  CUDA_VERSION          = 11040 = version 11.4
317: Kokkos::Cuda[ 0 ] Tesla V100-PCIE-16GB capability 7.0, Total Global Memory: 15.77 G, Shared Memory per Block: 48 K : Selected
317: Kokkos::Cuda[ 1 ] Tesla V100-PCIE-16GB capability 7.0, Total Global Memory: 15.77 G, Shared Memory per Block: 48 K
317:   Kokkos Version: 4.1.0
317: Compiler:
317:   KOKKOS_COMPILER_GNU: 850
317:   KOKKOS_COMPILER_NVCC: 1140
317: Architecture:
317:   CPU architecture: none
317:   Default Device: N6Kokkos4CudaE
317:   GPU architecture: VOLTA70
317:   platform: 64bit
317: Atomics:
317: Vectorization:
317:   KOKKOS_ENABLE_PRAGMA_IVDEP: no
317:   KOKKOS_ENABLE_PRAGMA_LOOPCOUNT: no
317:   KOKKOS_ENABLE_PRAGMA_UNROLL: no
317:   KOKKOS_ENABLE_PRAGMA_VECTOR: no
317: Memory:
317:   KOKKOS_ENABLE_HBWSPACE: no
317:   KOKKOS_ENABLE_INTEL_MM_ALLOC: no
317: Options:
317:   KOKKOS_ENABLE_ASM: yes
317:   KOKKOS_ENABLE_CXX17: yes
317:   KOKKOS_ENABLE_CXX20: no
317:   KOKKOS_ENABLE_CXX23: no
317:   KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK: no
317:   KOKKOS_ENABLE_HWLOC: no
317:   KOKKOS_ENABLE_LIBDL: yes
317:   KOKKOS_ENABLE_LIBRT: no
317: Host Serial Execution Space:
[...]

PR #12466, opened 2023-10-31
stokhos: revise use of Cuda impl routines from kokkos

ndellingwood:
Compatibility update needed for kokkos following kokkos/kokkos#6544
Address kokkos/kokkos#6561

Co-authored-by: Damien L-G <dalg24+github@gmail.com>

Thanks @dalg24 for suggested update!

@trilinos/stokhos @etphipp 

## Motivation

Resolve compilation errors with Cuda backend

## Testing

**Local test (stokhos only) with sems-cuda/11.4.2**:
```
...
98% tests passed, 1 tests failed out of 66

Label Time Summary:
Stokhos    = 132.16 sec*proc (66 tests)

Total Test time (real) =  74.11 sec

The following tests FAILED:
   45 - Stokhos_TpetraCrsMatrixUQPCEUnitTest_Cuda_MPI_4 (Failed)
Errors while running CTest
```

where Stokhos_TpetraCrsMatrixUQPCEUnitTest_Cuda_MPI_4 is known failure #12427 

**Configuration**:
```
module purge
module load sems-gcc/8.5.0 sems-cuda/11.4.2 sems-cmake sems-git sems-openmpi/4.1.4 sems-ninja
export OMPI_CXX=$KOKKOS_DIR/bin/nvcc_wrapper

# Configure Trilinos
cmake \
 -GNinja \
 -DCMAKE_INSTALL_PREFIX="${PWD}/install" \
 -DCMAKE_CXX_STANDARD="17" \
 -DTPL_ENABLE_MPI:BOOL=ON \
 -DTPL_ENABLE_BLAS:BOOL=ON \
 -DTPL_ENABLE_LAPACK:BOOL=ON \
 -DTPL_ENABLE_CUSPARSE:BOOL=ON \
 -DTrilinos_ENABLE_TESTS=OFF \
 -DTrilinos_ENABLE_ALL_PACKAGES=OFF \
 -DTrilinos_ENABLE_Kokkos=ON \
 -DKokkos_ENABLE_CUDA=ON \
 -DKokkos_ENABLE_CUDA_LAMBDA=ON \
 -DKokkos_ENABLE_CUDA_UVM=OFF \
 -DTrilinos_ENABLE_Stokhos=ON \
  -DStokhos_ENABLE_TESTS=ON \
 -DKokkos_SOURCE_DIR_OVERRIDE:STRING=kokkos \
 -DKokkosKernels_SOURCE_DIR_OVERRIDE:STRING=kokkos-kernels \
$TRILINOS_DIR
```

### DIFF
--- a/packages/framework/ini-files/config-specs.ini
+++ b/packages/framework/ini-files/config-specs.ini
@@ -360,9 +360,6 @@ opt-set-cmake-var Kokkos_CoreUnitTest_Serial1_EXTRA_ARGS STRING : --gtest_filter
 opt-set-cmake-var Kokkos_CoreUnitTest_Default_MPI_1_EXTRA_ARGS STRING : --gtest_filter=-*defaultdevicetype.shared_space
 opt-set-cmake-var Kokkos_CoreUnitTest_Default_EXTRA_ARGS STRING : --gtest_filter=-*defaultdevicetype.shared_space
 
-# Run test in serial to avoid timeouts
-opt-set-cmake-var PanzerMiniEM_Darcy_MueLu_order1_MPI_1_RUN_SERIAL BOOL FORCE : ON
-
 # Options from SEMSDevEnv.cmake
 opt-set-cmake-var Trilinos_ENABLE_PyTrilinos BOOL : OFF
 opt-set-cmake-var Boost_INCLUDE_DIRS PATH : ${SEMS_BOOST_INCLUDE_PATH|ENV}
--- a/packages/framework/ini-files/config-specs.ini
+++ b/packages/framework/ini-files/config-specs.ini
@@ -360,6 +360,9 @@ opt-set-cmake-var Kokkos_CoreUnitTest_Serial1_EXTRA_ARGS STRING : --gtest_filter
 opt-set-cmake-var Kokkos_CoreUnitTest_Default_MPI_1_EXTRA_ARGS STRING : --gtest_filter=-*defaultdevicetype.shared_space
 opt-set-cmake-var Kokkos_CoreUnitTest_Default_EXTRA_ARGS STRING : --gtest_filter=-*defaultdevicetype.shared_space
 
+# Run test in serial to avoid timeouts
+opt-set-cmake-var PanzerMiniEM_Darcy_MueLu_order1_MPI_1_RUN_SERIAL BOOL FORCE : ON
+
 # Options from SEMSDevEnv.cmake
 opt-set-cmake-var Trilinos_ENABLE_PyTrilinos BOOL : OFF
 opt-set-cmake-var Boost_INCLUDE_DIRS PATH : ${SEMS_BOOST_INCLUDE_PATH|ENV}
--- a/packages/panzer/mini-em/example/BlockPrec/CMakeLists.txt
+++ b/packages/panzer/mini-em/example/BlockPrec/CMakeLists.txt
@@ -207,6 +207,7 @@ TRIBITS_ADD_TEST(
    POSTFIX_AND_ARGS_0
    "order1" --solver=MueLu --numTimeSteps=1 --linAlgebra=Tpetra --inputFile=darcyTet.xml --x-elements=15  --y-elements=15 --z-elements=15
    NUM_MPI_PROCS 1
+   RUN_SERIAL
    )
 
 TRIBITS_ADD_TEST(
--- a/packages/stokhos/src/kokkos/Cuda/Stokhos_Cuda_BlockCrsMatrix.hpp
+++ b/packages/stokhos/src/kokkos/Cuda/Stokhos_Cuda_BlockCrsMatrix.hpp
@@ -113,13 +113,19 @@ public:
                      const block_vector_type & x ,
                      const block_vector_type & y )
   {
+    const int warp_size = Kokkos::Impl::CudaTraits::WarpSize;
+    auto const maxWarpCount = std::min<unsigned>(
+        execution_space().cuda_device_prop().maxThreadsPerBlock / warp_size,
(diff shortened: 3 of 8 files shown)


etphipp (approved):
Looks great, thanks!

ndellingwood:
The cuda build failed in a Panzer test due to timeout [PanzerMiniEM_Darcy_MueLu_order1_MPI_1](https://trilinos-cdash.sandia.gov/test/39614928)
@cgcgcg @etphipp does this PanzerMiniEM_Darcy_MueLu_order1_MPI_1 test use Stokhos?

ndellingwood:
I'm curious if the timeout occurs consistently with this change, retesting

cgcgcg:
I don't think Panzer has a dependency on Stokhos. But it does depend on Sacado. Could that have an impact? I only added the test in question a couple of days ago, so it could also be that it takes too long under load.

ndellingwood:
@cgcgcg Thanks for the quick reply! These changes would only impact Stokhos (and really should not result in any difference, they essentially move some impl stuff over from Kokkos). I suspect your guess that the test time may be impacted under load might be the case here. We'll see how it looks after a retest

ndellingwood:
The [PanzerMiniEM_Darcy_MueLu_order1_MPI_1](https://trilinos-cdash.sandia.gov/test/39661729) test timed out in the Cuda build again

ndellingwood:
There were two consecutive nightly builds with gcc/8.3.0 (no cuda) posting timeouts for the PanzerMiniEM_Darcy_MueLu_order1_MPI_1 test, independent of the changes in this PR. I'll test on the son sems gpu machine to see if I can reproduce the cuda failures in the PR and see if applying the RUN_SERIAL option to this test helps avoid the timeout

rppawlo:
@ndellingwood - if that doesn't work, then please just disable the test. Its new and obviously there is something that needs to be fixed.

ndellingwood:
@cgcgcg @rppawlo @etphipp I updated the frameworks ini config to set the `PanzerMiniEM_Darcy_MueLu_order1_MPI_1` to RUN_SERIAL to see how that goes with the AT. Since this occurred with cuda and serial gcc/8.3.0 builds I placed the setting in the most general location. My change marked the PR stale, could one of you reapprove?

Adding @trilinos/framework @sebrowne to the thread since I made an update to the config-specs.ini file, can you confirm if the change I made is acceptable or needs to be handled differently?

cgcgcg (dismissed):
I approve of running that test in serial. If that indeed fixes things I should go back and see why this particular one likes to hang.

rppawlo:
Thanks @ndellingwood ! I think we really want to set in this in the tribits_add_advanced_test macro for this test. Setting this in the .ini file means it only runs in serial for the autotester. Anyone not using genconfig (e.g. external users, devs with their own cmake scripts, and spack users) will probably see this issue in their local testing.

ndellingwood:
Adding the RUN_SERIAL got the test to passing, so that is useful data point testing that in the AT environemtn

> I think we really want to set in this in the tribits_add_advanced_test macro for this test

@rppawlo @cgcgcg should I merge this as is (this will get Cuda builds with kokkos core/kernels develop branches compiling again) and I'll follow on with a `tribits_add_advanced_test` update?

ndellingwood:
@rppawlo @cgcgcg I added the `RUN_SERIAL` option to the CMakeLists.txt file for the Darcy_MueLu test in b3d9fb62bf054e6e9527e92c82f51fc933a3d2fc, can you confirm and reapprove if this is sufficient for testing beyond the AT config option?

rppawlo (dismissed):
Thanks @ndellingwood ! Looks good to me. You could remove the change to the .ini file since it is now covered in the change to the cmake file.

sebrowne (changes requested):
As Roger stated, if we can omit an extra entry from `config-specs.ini` I'd like to do that for the sake of readability.  But yes, this change was correct if you wanted to change the setting in this file (FWIW I agree with Roger that this belongs in the `CMakeLists.txt`, so thanks!)

ndellingwood:
cb94104f99b3bd7e75c9582d2d76b038cce40124 contains the revert of the `config-specs.ini` file change; @rppawlo @sebrowne I'll need a reapproval please, thanks!

sebrowne (commented):
Approved from the perspective of it doesn't change Framework code any more.  Will defer to others for technical changes.

rppawlo (approved):

sebrowne:
> @sebrowne the PR-gcc-openmpi build failed during runner setup:
> 
> ```
> CMakeCache.txt
>  Error: Value cannot be null. (Parameter 'ContainerId')
> ```
> 
> Should I just retest?

It shouldn't require that to pass to merge the PR.  This is intentional while the new github-actions-based stuff is being worked.  If you see failures, feel free to tag framework or @fryeguy52 and @cabejackson so that they can take a look.

ndellingwood:
@trilinos/framework @fryeguy52 @cabejackson @sebrowne just to check this is intentional, it appears the AUTOMERGE label is disabled when the PR-gcc-openmpi job fails, e.g.
```
Cannot Merge Pull Request# 12466: Github Mergeability = unstable; Git Mergeable = True
```

The PR-gcc-openmpi failed earlier in this PR due apparent issues at the runner loading stage (some error with CMakeCache.txt)

sebrowne:
Hmmmm perhaps not intentional, but I'm glad that it did allow you to merge.  We will investigate if there is a way to get AUTOMERGE to work even while the new PR build fails.  Worst case, AUTOMERGE is broken until we transition.  Although, once we get a few little hiccups resolved, that build should start passing so that may fix it as well.

PR #12471, opened 2023-11-02
tpetra: guard printf use with sycl

ndellingwood:
@trilinos/tpetra @csiefer2 @brian-kelley @jhux2 

## Motivation

Resolve compilation error with SYCL backend (printf still broken)

### DIFF
--- a/packages/tpetra/core/src/Tpetra_Import_Util2.hpp
+++ b/packages/tpetra/core/src/Tpetra_Import_Util2.hpp
@@ -1311,10 +1311,6 @@ lowCommunicationMakeColMapAndReindex (
         const int PID = owningPIDs_view[j];
         auto outcome = RemoteGIDs_view_map.insert(GID, PID);
         if(outcome.success() && PID == -1) {
-// FIXME_SYCL Enable again once a SYCL conforming printf implementation is available.
-#ifndef KOKKOS_ENABLE_SYCL
-          printf("Cannot figure out if ID is owned.\n");
-#endif
           Kokkos::abort("Cannot figure out if ID is owned.\n");
         }
       }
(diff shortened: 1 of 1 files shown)


brian-kelley (approved):
Guess trilinos doesn't have ``Kokkos::printf`` yet :(

packages/tpetra/core/src/Tpetra_Import_Util2.hpp
```diff
@@ -1311,7 +1311,10 @@ lowCommunicationMakeColMapAndReindex (
         const int PID = owningPIDs_view[j];
         auto outcome = RemoteGIDs_view_map.insert(GID, PID);
         if(outcome.success() && PID == -1) {
+// FIXME_SYCL Enable again once a SYCL conforming printf implementation is available.
+#ifndef KOKKOS_ENABLE_SYCL
           printf("Cannot figure out if ID is owned.\n");
+#endif
```
masterleinad, line 1317: Why not just
```suggestion
```
?

masterleinad (commented):
After the next release, there will be `Kokkos::printf` as a portable version of `printf`.

masterleinad:
I hope there will be some `SYCL` CI soon.

packages/tpetra/core/src/Tpetra_Import_Util2.hpp
ndellingwood: Fair question, I assumed the printf was a useful diagnostic in the package. @brian-kelley @csiefer2 would you prefer that I delete the printf altogether?

brian-kelley: Fine with me to delete it, especially since the ``Kokkos::abort`` message is exactly the same.

ndellingwood:
Oop, merging Daniel's suggested change blew away the previous review, @brian-kelley can you reapprove when you get a chance?

brian-kelley (approved):

PR #12479, opened 2023-11-03
MueLu: Update coloring documentation

### LINKED ISSUE #12410: MueLu: Some coloring algorithms missing from user's guide
@trilinos/muelu 

As noted in #12403, "Vertex based" and "edge filtering" coloring algorithms are not documented in the user's guide Table 5.6 on page 36. We should document them. For example, it was totally unclear that "edge filtering" is actually "vertex based bit set" followed by a filtering step without reading a helpful comment in `UncoupledAggregationFactory_kokkos_def.hpp`.

GrahamBenHarper:
@trilinos/muelu 

We talked about eventually just pointing to the Kokkos-kernels documentation when we allow the options to pass straight through. Until then, we should still document these options correctly.

Closes #12410

### DIFF
--- a/packages/muelu/doc/UsersGuide/options.tex
+++ b/packages/muelu/doc/UsersGuide/options.tex
@@ -215,7 +215,10 @@ An example of line smoothing is provided in \texttt{packages/trilinoscouplings/e
                                     This means the coarse ratio is higher than in \verb!mis2 aggregation!.
                                     Faster than coloring-based aggregation.  Cannot control min/max aggregate size. & Yes & Yes \\
       \verb!serial! &               Computes a D2 coloring on host, and then does 4-phase aggregation on device. & No & Optional \\
-      \verb!vertex based bit set! & Computes a D2 coloring in parallel with conflicts resolved by neighbors-of-neighbors loop. Then does 4-phase aggregation on device. & Yes & No \\
+      \verb!default! &              Computes \verb!serial! on a serial device and \verb!net based bit set! on a parallel device. & Yes & Optional \\
+      \verb!vertex based! &         Computes a D2 coloring in parallel with conflicts resolved by neighbors-of-neighbors loop. Then does 4-phase aggregation on device. & Yes & No \\
+      \verb!vertex based bit set! & Computes the same as \verb!vertex based!, but with an optimization using a 32-bit integer instead of 32 bools to track forbidden colors. & Yes & No \\
+      \verb!edge filtering! &       Computes the same as \verb!vertex based bit set!, but with an optimization that can filter and skip some edges. & Yes & No \\
       \verb!net based bit set! &    Computes a D2 coloring in parallel with net-based algorithm, which is asymptotically faster than vertex-based (though not always faster in practice).
                                     Then does 4-phase aggregation on device. & Yes & No \\
       \bottomrule


cgcgcg:
Why not point to Kokkos Kernels now?

GrahamBenHarper:
@cgcgcg I don't think just linking to Kokkos-Kernels would fully address the relevant aspects to MueLu.
 1. We would need something that translates MueLu's options like "vertex based bit set" to Kokkos-Kernels' `KokkosGraph::COLORING_D2_VB_BIT` if we link to Kokkos-Kernels currently, making it more complicated for the user.
 2. We would still need a table or list of some sort describing which options we support currently, as we have to manually add the options in MueLu at the moment. Any changes in Kokkos-Kernels may not be reflected in MueLu immediately without a better interface--not that I'm expecting changes, but the documentation, regardless of who hosts it, should reflect our code's features as-is.

Once we have a full pass-through interface to Kokkos-Kernels, I'm more than happy to yank it all and simply point to Kokkos-Kernels, but there's still a lot of translating that we need to do for users in the meantime, and I don't see any issues on the Kokkos-Kernels repo indicating they care about an interface at the moment.

cgcgcg (approved):

packages/muelu/doc/UsersGuide/options.tex
```diff
@@ -215,7 +215,10 @@ \section{Aggregation options}
                                     This means the coarse ratio is higher than in \verb!mis2 aggregation!.
                                     Faster than coloring-based aggregation.  Cannot control min/max aggregate size. & Yes & Yes \\
       \verb!serial! &               Computes a D2 coloring on host, and then does 4-phase aggregation on device. & No & Optional \\
-      \verb!vertex based bit set! & Computes a D2 coloring in parallel with conflicts resolved by neighbors-of-neighbors loop. Then does 4-phase aggregation on device. & Yes & No \\
+      \verb!default! &              Computes the same as \verb!serial!. & No & Optional \\
```
cgcgcg, line 218: That's not correct, according to the Kokkos Kernels documentation:
https://github.com/kokkos/kokkos-kernels/wiki/D2-Graph-Coloring#available-algorithms

cgcgcg (changes requested):

packages/muelu/doc/UsersGuide/options.tex
GrahamBenHarper: Thanks for catching that. I had misaligned rows while cross-referencing our interface in `UncoupledAggregationFactory_kokkos` to the Kokkos-Kernels wiki :)

packages/muelu/doc/UsersGuide/options.tex
```diff
@@ -215,7 +215,10 @@ \section{Aggregation options}
                                     This means the coarse ratio is higher than in \verb!mis2 aggregation!.
                                     Faster than coloring-based aggregation.  Cannot control min/max aggregate size. & Yes & Yes \\
       \verb!serial! &               Computes a D2 coloring on host, and then does 4-phase aggregation on device. & No & Optional \\
-      \verb!vertex based bit set! & Computes a D2 coloring in parallel with conflicts resolved by neighbors-of-neighbors loop. Then does 4-phase aggregation on device. & Yes & No \\
+      \verb!default! &              Computes the same as \verb!net based bit set!. & No & Optional \\
```
cgcgcg, line 218: Maybe we're not looking at the same documentation, but this seems also wrong.

cgcgcg (changes requested):

packages/muelu/doc/UsersGuide/options.tex
GrahamBenHarper: I see it now. It's both, not one or the other

cgcgcg (dismissed):

cgcgcg (approved):

GrahamBenHarper:
Nothing was posted on the cdash. Maybe the AT died?

PR #12480, opened 2023-11-03
MueLu: Node traits and coarse solver factorization

### LINKED ISSUE #11843: MueLu: MUELU_HAVE_* preprocessor defines are incorrect
@trilinos/muelu 

As pointed out by @jhux2 in https://github.com/trilinos/Trilinos/pull/11777#issuecomment-1532166976, we check quite a few preprocessor defines of the form `MUELU_HAVE_*` which do not exist.
```
$ cd packages/muelu
$ grep -rn MUELU_HAVE
src/Graph/Containers/MueLu_Zoltan2GraphAdapter.hpp:457:#endif// MUELU_HAVE_ZOLTAN2
test/unit_tests/GenericRFactory.cpp:92:#   if !defined(MUELU_HAVE_IFPACK)
test/unit_tests/GenericRFactory.cpp:95:#   if !defined(MUELU_HAVE_IFPACK2)
test/unit_tests/GenericRFactory.cpp:227:#   if !defined(MUELU_HAVE_IFPACK)
test/unit_tests/GenericRFactory.cpp:230:#   if !defined(MUELU_HAVE_IFPACK2)
test/unit_tests/TentativePFactory.cpp:541:#   if !defined(MUELU_HAVE_IFPACK)
test/unit_tests/TentativePFactory.cpp:544:#   if !defined(MUELU_HAVE_IFPACK2)
test/unit_tests/Repartition.cpp:909:#   if !defined(MUELU_HAVE_AMESOS) || !defined(MUELU_HAVE_IFPACK)
test/unit_tests/Repartition.cpp:912:#   if !defined(MUELU_HAVE_AMESOS2) || !defined(MUELU_HAVE_IFPACK2)
test/unit_tests/Repartition.cpp:1022:#   if !defined(MUELU_HAVE_AMESOS) || !defined(MUELU_HAVE_IFPACK)
test/unit_tests/Repartition.cpp:1025:#   if !defined(MUELU_HAVE_AMESOS2) || !defined(MUELU_HAVE_IFPACK2)
```
Those are typos and anything of the form `MUELU_HAVE_*` is never true, so we should fix them.

cgcgcg:
@trilinos/muelu 

## Motivation
- Use new `Node` traits in two more spots.
- Call `numericFactorization` in MueLu's coarse solver setup.
- Interface test: rebase gold files and ignore smoother complexity because it depends on the type of coarse solver that's chosen. Before we triggered factorization in setup the coarse solver was simply ignored in the complexity computation.
- ReitzingerP test: skip construction of coarse solvers as the matrices can be singular.
- Change assert in block smoother test so that it passes with other scalar types.
- Replace "MUELU_HAVE_*" with "HAVE_MUELU_*". Closes #11843
- Fix PL bug in SaPFactory
- Add missing propagation of "Node Comm" to TentativeP_kokkos.

@ndellingwood : Before this change, we only called `setParameters` during the multigrid setup and `solve` during multigrid solve. We were seeing that the Amesos2 solver was constructed during our first solve. Can you confirm that with this change we should not be seeing factorization during the first `solve` call?

### DIFF
--- a/packages/muelu/src/Graph/Containers/MueLu_Zoltan2GraphAdapter.hpp
+++ b/packages/muelu/src/Graph/Containers/MueLu_Zoltan2GraphAdapter.hpp
@@ -454,6 +454,6 @@ template <typename User, typename UserCoord>
 }  //namespace MueLu
 
 
-#endif// MUELU_HAVE_ZOLTAN2
+#endif// HAVE_MUELU_ZOLTAN2
   
 #endif
--- a/packages/muelu/src/Operators/MueLu_Maxwell1_def.hpp
+++ b/packages/muelu/src/Operators/MueLu_Maxwell1_def.hpp
@@ -235,18 +235,7 @@ namespace MueLu {
 
 
     // Are we using Kokkos?
-# ifdef HAVE_MUELU_SERIAL
-    if (typeid(Node).name() == typeid(Tpetra::KokkosCompat::KokkosSerialWrapperNode).name())
-      useKokkos_ = false;
-# endif
-# ifdef HAVE_MUELU_OPENMP
-    if (typeid(Node).name() == typeid(Tpetra::KokkosCompat::KokkosOpenMPWrapperNode).name())
-      useKokkos_ = true;
-# endif
-# ifdef HAVE_MUELU_CUDA
-    if (typeid(Node).name() == typeid(Tpetra::KokkosCompat::KokkosCudaWrapperNode).name())
-      useKokkos_ = true;
-# endif
+    useKokkos_ = !Node::is_serial;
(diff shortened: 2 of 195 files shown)


ndellingwood:
> Can you confirm that with this change we should not be seeing factorization during the first `solve` call?

@cgcgcg if you're explicitly separating out the numeric factorization call from the solve, then your statement should be correct. Which solver(s) does this use?

cgcgcg:
This should be the common interface for all solvers we could use from Amesos2. In practice it's probably Klu2, SuperLU or (in very specific cases) LAPACK.

lucbv (approved):

PR #12483, opened 2023-11-04
ATDM: Update ats2 configuration to use compliant cuda and cmake versions

GrahamBenHarper:
@bartlettroscoe I still use the ATDM configures for ats2 machines like Vortex sometimes, but the script as-is only allows cuda 10, which is no longer supported. I've upgraded it to 11 by simply setting the string `ATDM_CONFIG_COMPILER=CUDA-11.2.152_GCC-8.3.1_SPMPI-ROLLING` since the sparc modules have a build that corresponds to this exactly.

Additionally, I removed the `module load cmake/3.18.0` line since the sparc modules pick up the correct version of cmake anyway.

I don't know what level of support or testing is required for this, but I get the feeling I'm one of the only people using this since it's so terribly out of date. I'm hoping something like this can be merged so I don't have to keep this stashed in my local Trilinos versions :)

## Testing
I've been building and testing MueLu with this on Vortex, and I can confirm this doesn't break anything.


bartlettroscoe (approved):
Looks good to me.  If it works for you then that is all that matters.

bartlettroscoe:
@GrahamBenHarper, looks like the unit tests need to be updated for this change.  The file that needs updated is:

* [cmake/std/atdm/ats2/custom_builds_unit_tests.sh](https://github.com/GrahamBenHarper/Trilinos/blob/atdm_vortex/cmake/std/atdm/ats2/custom_builds_unit_tests.sh)

That is what caused the last PR testing iteration to fail.

Note that you should be able to run those unit tests locally manually as well.

bartlettroscoe:
> The only reason I'm still using ATDM and not genconfig is because I learned ATDM first, and I had ATDM scripts for MueLu development and for the MrHyDE application. I imagine it isn't difficult to learn genconfig, but it's still on my todo list.

@GrahamBenHarper, if you do switch over to GenConfig, please let me know about the usability of these two systems, from the Trilinos developer standpoint.

bartlettroscoe (approved):
Thanks for fixing tests!

PR #12485, opened 2023-11-06
Tempus: Split Up BDF2 Tests

### LINKED ISSUE #12484: Tempus: Split Up BDF2 Tests
## Enhancement
BDF2 tests in debug mode were occasionally running too
long, when resources were loaded.  Splitting up these tests into
individual tests based on the governing equations (ModelEvaluator).

Moved the constructor tests to the unit tests.

ccober6:
BDF2 tests in debug mode were occassionally running too long, when resources were loaded.  Splitting up these tests into individual tests based on the governing equations (ModelEvaluator).

Moved the constructor tests to the unit tests.

@trilinos/tempus 

## Related Issues

* Closes #12484


ccober6:
@trilinos/framework   Hmm... not sure what the issue is here.

` Error: Value cannot be null. (Parameter 'ContainerId')`

Can you help?

rppawlo:
It looks like all the Tempus_Test_BDF2_*.cpp files repeat all the BDF tests in each file (e.g. The Tempus_Test_BDF2_SinCos.cpp contains CDR and vanderpol as well as SinCos). You should probably remove the tests that are not necessary for each particular file.

rppawlo (approved):

PR #12493, opened 2023-11-07
Amesos2 mumps bugfix - Missing comma in function call

modkin:
@trilinos/amesos2

## Motivation

Commit [cdb2ce](https://github.com/trilinos/Trilinos/commit/cdb2cecfde2688a7f6f947694eb02e3f18d50c5c) introduced a bug which appears in Amesos2 if build with MUMPS support. The `do_get` function call misses a comma.

## Testing

Apparently there are no tests for Amesos2 with MUMPS support. Otherwise this would not have passed in the first place.


cgcgcg:
Thanks for catching that! Yes, our testing has some gaps. I wish we would have picked that up in nightly testing somewhere..

jhux2 (approved):
lgtm, thank you for the contribution

ndellingwood (approved):

PR #12497, opened 2023-11-08
MueLu RefMaxwell: Fix issue in default parameterlist

cgcgcg:
@trilinos/muelu 

- Change a parameter value in RefMaxwell's default list to match previous behavior.
- Revert change to trigger coarse solver factorization during setup. EMPIRE and Albany reported issues.


rppawlo (dismissed):
Thanks @cgcgcg ! This gets us past most of the failures in empire.

PR #12512, opened 2023-11-10
Tempus: Fix DIRK getAlpha

### LINKED ISSUE #12511: Tempus: Fix DIRK getAlpha
## Bug Report
Fix and improve getAlpha(dt) function.  It would only return
alpha(0,0) instead of alpha(i,i).

ccober6:
Fix and improve getAlpha(dt) function.  It would only return alpha(0,0) instead of alpha(i,i).

@trilinos/tempus 

* Closes #12511


mperego (approved):
It looks good. Thanks!

PR #12518, opened 2023-11-10
Ifpack2 Hypre: Only instantiate interface for hypre enabled options

cgcgcg:
@trilinos/ifpack2 

## Motivation
Before this change, `Ifpack2::Hypre` was implemented the same no matter what template types were used. But Hypre gets compiled only for one combination of Scalar & Ordinal. This change should handle cases where we enable multiple combinations of types.

Also fixes the scripts that generate the header for the `HYPRE_*` functions. Every now and then that file was invalid C++.


GrahamBenHarper (commented):
How are we planning to test this?

cgcgcg:
We don't have Hypre in our nightlies or PR testing. I built it locally.

jhux2 (commented):
How does `HYPRE_REAL` get defined?

PR #12521, opened 2023-11-11
MueLu: Update update_params.sh to work for BSD and GNU sed

### LINKED ISSUE #12500: MueLu: MueLu_MasterList_MPI_1 failure with MacOS - local environment issue?
## Question
@trilinos/muelu @csiefer2 @cgcgcg @GrahamBenHarper 

In a build of Trilinos on my Mac (using MacPorts for compilers) I have failures with `MueLu_MasterList_MPI_1` as follows:

```
    Start 885: MueLu_MasterList_MPI_1
1/1 Test #885: MueLu_MasterList_MPI_1 ...........***Failed    4.32 sec
sed: 1: "MueLu_MasterList.cpp": invalid command code M
sed: 1: "MueLu_MasterList.cpp": invalid command code M
...
569c569
<             "<Parameter name=\"multigrid algorithm\" type=\"string\" value=\""pg"\"/>"
---
>             "<Parameter name=\"multigrid algorithm\" type=\"string\" value=\"pg\"/>"
MueLu_MasterList.cpp does not match
Do not edit MueLu_MasterList.cpp directly, but edit masterList.xml and run update_params.sh
a non-zero exit code. Per user-direction, the job has been aborted.
mpiexec detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[1775,1],0] 
  Exit code:    1


0% tests passed, 1 tests failed out of 1

Subproject Time Summary:
MueLu    =   4.32 sec*proc (1 test)

Total Test time (real) =  20.39 sec

The following tests FAILED:
  885 - MueLu_MasterList_MPI_1 (Failed)
```

Is my version of sed incompatible with the test?

From my sed man page:
```
...
HISTORY
     A sed command, written by L. E. McMahon, appeared in Version 7 AT&T UNIX.

AUTHORS
     Diomidis D. Spinellis <dds@FreeBSD.org>

BUGS
     Multibyte characters containing a byte with value 0x5C (ASCII ‘\’) may be incorrectly treated as line continuation characters in arguments to the
     “a”, “c” and “i” commands.  Multibyte characters cannot be used as delimiters with the “s” and “y” commands.

macOS 12.7                                                            June 10, 2020                                                            macOS 12.7
```

Thanks

GrahamBenHarper:
@trilinos/muelu

As pointed out by @ndellingwood in #12500, invoking `sed -i` requires an extension argument immediately following `-i` on Apple machines because they use the BSD version of sed.

Really it boils down to the fact that using `-i` is not portable in testing. In fact, our checks for it in our C++ MueLu source seem fragile as well, as somebody who installs GNU sed on an Apple machine will experience different test failures. I changed this particular instance to manually create and move a `.tmp` file since that's what `sed -i` is supposed to do internally.

Closes #12500.

### DIFF
--- a/packages/muelu/doc/UsersGuide/update_params.sh
+++ b/packages/muelu/doc/UsersGuide/update_params.sh
@@ -173,9 +173,13 @@ echo ';
 }
 ' >> $code_file
 
-# fix quotation
-sed -i '/<Parameter/ s/\\""/\\"/g' $code_file
-sed -i '/<Parameter/ s/"\\"/\\"/g' $code_file
+# fix quotation using sed
+# GH: similar to other instances in MueLu, we need to work around the GNU/BSD sed issue
+#     the moral of the story is -i is not portable for testing!
+sed '/<Parameter/ s/\\""/\\"/g' "$code_file" > "$code_file.tmp"
+mv "$code_file.tmp" "$code_file"
+sed '/<Parameter/ s/"\\"/\\"/g' "$code_file" > "$code_file.tmp"
+mv "$code_file.tmp" "$code_file"
 
 # generate LaTeX files (MueLu options and ML compatibility)
 SECTIONS=( "general" "smoothing_and_coarse" "aggregation" "misc" "multigrid" "rebalancing" "reuse" "refmaxwell" )


GrahamBenHarper:
Blocked by randomly failing ROL test `ROL_test_algorithm_TypeP_ProximalGradient_MPI_1`. @dridzal just FYI it looks like the tolerance might be a hair too tight

cgcgcg (approved):
If "-i" is not portable then we should remove it everywhere in MueLu.

GrahamBenHarper:
@cgcgcg I agree, but it may take slightly more work since they're system invocations from C++ source

PR #12524, opened 2023-11-13
Amesos2 : CUDA-enabled Superlu_DIST

iyamazaki:
@trilinos/amesos2 

## Motivation

This PR try fixing the issues encountered while linking with CUDA-enabled SuperLU_DIST.

## Additional Information

Hopeful fix for [xSDK issue](https://github.com/xsdk-project/xsdk-issues/issues/235), @cgcgcg.

### DIFF
--- a/packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
+++ b/packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
@@ -72,8 +72,6 @@
 
 namespace SLUD {
 
-extern "C" {
-
 #if SUPERLU_DIST_MAJOR_VERSION > 4
 // SuperLU_Dist before major version 5 does not contain the config file
 #include "superlu_dist_config.h" // provides define for size 32 or 64 int_t
@@ -107,9 +105,6 @@ extern "C" {
 #include "superlu_zdefs.h"     // double-precision complex definitions
   }
 #endif  // HAVE_TEUCHOS_COMPLEX
-
-
-} // end extern "C"
 
 // multiplication of SLUD types
 template <typename slu_scalar_t, typename slu_mag_t>
--- a/packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
+++ b/packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
@@ -69,6 +69,11 @@
 
 #include "Amesos2_TypeMap.hpp"
 
+#ifdef KOKKOS_ENABLE_CUDA
+  #include <cublas_v2.h>
+  #include <cuda_runtime_api.h>
+#endif
+
 
 namespace SLUD {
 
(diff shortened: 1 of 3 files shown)


packages/amesos2/test/adapters/Tpetra_CrsMatrix_Adapter_UnitTests.cpp
```diff
@@ -202,7 +199,7 @@ namespace {
     }
     eye->fillComplete();
     // Create using non-member function
-    RCP<ADAPT> adapter  = Amesos2::createMatrixAdapter<MAT>(eye);
```
ndellingwood, line 205: @iyamazaki does commenting this portion of code remove a portion of the test, or was this unused?

packages/amesos2/test/adapters/Tpetra_CrsMatrix_Adapter_UnitTests.cpp
```diff
@@ -435,9 +432,6 @@ namespace {
     RCP<const Comm<int> > comm = Tpetra::getDefaultComm();
     const size_t numprocs = comm->getSize();
     const size_t rank     = comm->getRank();
-    // create a Map for our matrix
-    global_size_t nrows = 6;
```
ndellingwood, line 439: @iyamazaki does commenting this portion of the test remove code coverage? Is this necessary to get the changes working with Cuda-enabled SLUDist?

packages/amesos2/test/adapters/Tpetra_CrsMatrix_Adapter_UnitTests.cpp
iyamazaki: @ndellingwood, I don't think this will remove the portion of the test, and the variable is just unused?

packages/amesos2/test/adapters/Tpetra_CrsMatrix_Adapter_UnitTests.cpp
iyamazaki: I don't think this will change the test coverage, @ndellingwood.

packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
```diff
@@ -69,10 +69,13 @@
 
 #include "Amesos2_TypeMap.hpp"
 
+#ifdef KOKKOS_ENABLE_CUDA
```
ndellingwood, line 72: Is there a way to detect if SLUDist was built with Cuda enabled and add that to this check?

iyamazaki: That will be nice, but I am not sure. Any idea?

cgcgcg:
Can this be merged?

packages/amesos2/src/Amesos2_Superludist_TypeMap.hpp
ndellingwood: Including the cuda headers makes me feel a bit uneasy, would probably be better to guard based on whether sludist was configuring and built with Cuda support but I won't hold up the PR

ndellingwood (approved):

